Issue #463, opened 2017-04-10
P4Runtime Serializer discards annotations on match key fields and action params


antoninbas:
Annotations on match key fields and action parameters don't show up in the generated file.
I'm attaching an example program to help reproduce. There is a `@p4runtime_id(999)` annotation on a match field and on an action param, and neither show in the output.
[p4_16_annotations.p4.txt](https://github.com/p4lang/p4c/files/911481/p4_16_annotations.p4.txt)

PR #491, opened 2017-04-17
Support annotations for actions params and match keys. (#463)

sethfowler:
This PR implements P4Runtime serializer support for annotations on action parameters and match keys.

I had to make KeyElement implement IAnnotated; this is just an oversight of the current code, since KeyElement already contains a list of annotations.

### DIFF
--- a/control-plane/p4RuntimeSerializer.cpp
+++ b/control-plane/p4RuntimeSerializer.cpp
@@ -233,6 +233,8 @@ struct MatchField {
     const cstring name;       // The fully qualified external name of this field.
     const MatchType type;     // The match algorithm - exact, ternary, range, etc.
     const uint32_t bitwidth;  // How wide this field is.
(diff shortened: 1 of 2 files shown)


control-plane/p4RuntimeSerializer.cpp
```diff
@@ -938,6 +940,7 @@ class P4RuntimeSerializer {
             auto paramName = controlPlaneName(actionParam);
             param->set_id(index++);
             param->set_name(paramName);
+            addAnnotations(param, annotations);
```
antoninbas: this is incorrect, it picks up the action's annotations, not the parameter's annotations

sethfowler: Ack, you're right. Fixing now.

sethfowler:
@antoninbas OK, it's fixed.

Issue #486, opened 2017-04-16
P4_16 program struct field name collision causes wrong bmv2 JSON file


jafingerhut:
See the sample P4_16 program and JSON file generated on this issue: https://github.com/p4lang/behavioral-model/issues/331

I do not know whether fixing this requires only a change in p4c-bm2-ss, or whether it would require changes in both that and bmv2.

antoninbas:
I had pushed a patch to bmv2 to check for name / id duplicates when appropriate (https://github.com/p4lang/behavioral-model/pull/334). However it broke the tests for this repo (see output below). I reverted my change temporarily, but it would be great if the bmv2 backend could be fixed.

```
FAIL: bmv2/testdata/p4_16_samples/action_selector_shared-bmv2.p4
================================================================
Calling target program-options parser
Duplicate actions with name 'drop' in action profile 'as'
Could not connect to thrift client on port 9795
Make sure the switch is running and that you have the right port
Check for  ../testdata/p4_16_samples/action_selector_shared-bmv2.stf
*** simple_switch died with return code 1
*** CLI process failed with exit code 1
*** Test failed
FAIL bmv2/testdata/p4_16_samples/action_selector_shared-bmv2.p4.test (exit status: 1)
FAIL: bmv2/testdata/p4_16_samples/x-bmv2.p4
===========================================
Calling target program-options parser
Duplicate objects of type 'header' with name 's'
Could not connect to thrift client on port 9822
Make sure the switch is running and that you have the right port
Check for  ../testdata/p4_16_samples/x-bmv2.stf
*** simple_switch died with return code 1
*** CLI process failed with exit code 1
*** Test failed
FAIL bmv2/testdata/p4_16_samples/x-bmv2.p4.test (exit status: 1)
FAIL: bmv2/testdata/p4_14_samples/queueing.p4
=============================================
Calling target program-options parser
Duplicate objects of type 'header type' with name 'queueing_metadata_t'
Could not connect to thrift client on port 9850
Make sure the switch is running and that you have the right port
Check for  ../testdata/p4_14_samples/queueing.stf
*** simple_switch died with return code 1
*** CLI process failed with exit code 1
*** Test failed
FAIL bmv2/testdata/p4_14_samples/queueing.p4.test (exit status: 1)
FAIL: bmv2/testdata/p4_14_samples/16-TwoReferences.p4
=====================================================
Calling target program-options parser
Duplicate objects of type 'control node' with name 'C'
[...]

PR #494, opened 2017-04-17
Fix for issue #493

mbudiu-vmw:


mbudiu-vmw:
unfortunately it is not easy to create a negative test for this using our default testing framework, since this test succeeds with p4test, and fails with p4c-bm2-ss, and our scripts always run both compilers.

backends/bmv2/midend.cpp
```diff
@@ -121,8 +121,11 @@ MidEnd::MidEnd(CompilerOptions& options) {
                 // nothing further to do
                 return nullptr;
             // We save the names of some control blocks for special processing later
-            if (main->getConstructorParameters()->size() != 6)
-                ::error("%1%: Expected 6 arguments for main package", main);
+            if (main->getConstructorParameters()->size() != 6) {
+                ::error("%1%: Expected 6 arguments for main package; are you using %2%?",
+                        main, P4V1::V1Model::instance.file.toString());
+                return nullptr;
```
ChrisDodd: Isn't this caught by typechecking?   The constructor call to main must match the declaration in v1model.p4...

mbudiu-vmw:
This fails when v1model was not used.
We have no easy way to check which model was used, this is an approximation of a check.

mbudiu-vmw:
Travis has picked-up a new version of BMv2 which is very strict, checking for issue #486, which won't be fixed very soon.

PR #503, opened 2017-04-20
Update and clean up README.md

sethfowler:
`README.md` needed some updating; it contained out-of-date information about compiler and platform support. While I was at it, I tried to clean it up and make it a bit easier to follow for new users.

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,11 +35,11 @@ The code is currently alpha quality.
 
 # Dependences
 
-We have tested the compiler on U*X systems (MacOS and Ubuntu).  The
-following tools are required to build and run the compiler and tests:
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -125,7 +152,7 @@ Installing on macOS:
   ```
   brew install doxygen graphviz
   ```
-  For Latex we recommend MacTeX (http://www.tug.org/mactex/).
```
cc10512: These are required only if you want to build documentation. And I actually disabled PDF generation (it was a fairly useless doc) and thus LaTeX is no longer required. And the same for Ubuntu.

sethfowler: Thanks, I'll update the PR.

PR #506, opened 2017-04-20
Fix PI submodule ref pointer

antoninbas:


antoninbas:
I exceptionally used admin privileges to merge this without waiting for the Travis job to complete since it would have blocked all builds.

sethfowler:
That was a good move. Thanks.

PR #508, opened 2017-04-22
add generic getAnnotation method to make it easier to inspect annotations

ChrisDodd:


ir/base.def
```diff
@@ -232,6 +232,8 @@ class Annotations {
 /// Implemented by all objects that can have annotations
 interface IAnnotated {
     virtual Annotations getAnnotations() const = 0;
+    Annotation getAnnotation(cstring name) const override {
```
mbudiu-vmw: I am wondering whether this should be called getSingleAnnotation - to make it clear that it expects a single annotation with this name.

ChrisDodd: While in theory you can have multiple annotations with the same name, it almost never happens, so the common case is a single annotation.   We should have a special function that returns a set of annotations to deal with the multiple case, but we don't as no existing code deals with it, instead just picking the first one.

PR #509, opened 2017-04-23
Partial implementation of adding filename, line, column info to JSON …

jafingerhut:
…output

This PR is not ready to go yet, but I wanted to see if anyone had time to consider whether such a change would be welcome in p4c-bm2-ss.  The basic idea is to add keys "filename", "line", and "column" to many of the objects in the bmv2 JSON configuration file that is written.  There are 2 possible purposes I can think of for this:

+ help people looking at the JSON file correlate the objects back to the original P4 source code
+ enable future enhancements to bmv2 where its logging output can include this information, for easier debugging of P4 programs by developers.

If this is considered interesting for future inclusion in p4c, then I am also looking for comments on ways to improve the code I have written.  These changes aren't likely to be what someone skilled in C++ would write.

### DIFF
--- a/backends/bmv2/jsonconverter.cpp
+++ b/backends/bmv2/jsonconverter.cpp
@@ -355,6 +355,7 @@ class ExpressionConverter : public Inspector {
                 v->append(0);
                 v->append(width);
                 map.emplace(expression, j);
+                // TBD jafinger - add line/col here?
                 return;
             }
         } else if (instance->is<P4::BuiltInMethod>()) {
@@ -369,6 +370,7 @@ class ExpressionConverter : public Inspector {
                 auto l = get(bim->appliedTo);
                 e->emplace("right", l);
                 map.emplace(expression, result);
+                // TBD jafinger - add line/col here?
                 return;
             }
         }
@@ -803,13 +805,34 @@ JsonConverter::JsonConverter(const CompilerOptions& options) :
         headerParameter(nullptr), userMetadataParameter(nullptr), stdMetadataParameter(nullptr)
 {}
 
+void addSrcInfoData(Util::JsonObject* result,
+                    const IR::Node* node) {
+    unsigned lineNumber, columnNumber;
+    cstring fName = node->srcInfo.toSourcePositionData(&lineNumber,
+                                                       &columnNumber);
+    if (fName == NULL) {
+        // Do not add anything to the JSON file for this, as this is
+        // likely a statement synthesized by the compiler, and either
+        // not easy, or it is impossible, to correlate it directly
+        // with anything in the user's P4 source code.
+    } else {
+        cstring sourceFrag = node->srcInfo.toSourceFragment2();
+        result->emplace("filename", fName);
+        result->emplace("line", lineNumber);
+        result->emplace("column", columnNumber);
+        result->emplace("expression_string", sourceFrag);
+    }
+}
+
 // return calculation name
 cstring JsonConverter::createCalculation(cstring algo, const IR::Expression* fields,
-                                         Util::JsonArray* calculations) {
+                                         Util::JsonArray* calculations,
+                                         const IR::Node* node) {
     cstring calcName = refMap->newName("calc_");
     auto calc = new Util::JsonObject();
     calc->emplace("name", calcName);
     calc->emplace("id", nextId("calculations"));
+    addSrcInfoData(calc, node);
     calc->emplace("algo", algo);
     if (!fields->is<IR::ListExpression>()) {
         // expand it into a list
@@ -836,6 +859,9 @@ JsonConverter::convertActionBody(const IR::Vector<IR::StatOrDecl>* body,
                                  Util::JsonArray* result, Util::JsonArray* fieldLists,
                                  Util::JsonArray* calculations, Util::JsonArray* learn_lists) {
     for (auto s : *body) {
+        // TBD jafinger - add line/col at all individual cases below,
+        // or perhaps it can be done as a common case above or below
+        // for all of them?
         if (!s->is<IR::Statement>()) {
             continue;
         } else if (auto block = s->to<IR::BlockStatement>()) {
@@ -847,6 +873,7 @@ JsonConverter::convertActionBody(const IR::Vector<IR::StatOrDecl>* body,
         } else if (s->is<IR::ExitStatement>()) {
             auto primitive = mkPrimitive("exit", result);
             (void)mkParameters(primitive);
+            addSrcInfoData(primitive, s);
             break;
         } else if (s->is<IR::AssignmentStatement>()) {
             const IR::Expression* l, *r;
@@ -862,6 +889,11 @@ JsonConverter::convertActionBody(const IR::Vector<IR::StatOrDecl>* body,
                 operation = "modify_field";
             auto primitive = mkPrimitive(operation, result);
             auto parameters = mkParameters(primitive);
+            // TBD jafinger - assignment statement location with this
+            // code only includes the '=' 'operator'.  It would be
+            // nice if it included the entire assignment statement.
+            // Easy to do?
+            addSrcInfoData(primitive, s);
             auto left = conv->convertLeftValue(l);
             parameters->append(left);
             bool convertBool = type->is<IR::Type_Boolean>();
@@ -903,6 +935,7 @@ JsonConverter::convertActionBody(const IR::Vector<IR::StatOrDecl>* body,
                 }
                 auto primitive = mkPrimitive(prim, result);
                 primitive->emplace("parameters", parameters);
+                addSrcInfoData(primitive, s);
                 continue;
             } else if (mi->is<P4::ExternMethod>()) {
                 auto em = mi->to<P4::ExternMethod>();
@@ -911,6 +944,7 @@ JsonConverter::convertActionBody(const IR::Vector<IR::StatOrDecl>* body,
                         BUG_CHECK(mc->arguments->size() == 1, "Expected 1 argument for %1%", mc);
                         auto primitive = mkPrimitive("count", result);
                         auto parameters = mkParameters(primitive);
+                        addSrcInfoData(primitive, s);
                         auto ctr = new Util::JsonObject();
                         ctr->emplace("type", "counter_array");
                         ctr->emplace("value", extVisibleName(em->object));
@@ -924,6 +958,7 @@ JsonConverter::convertActionBody(const IR::Vector<IR::StatOrDecl>* body,
                         BUG_CHECK(mc->arguments->size() == 2, "Expected 2 arguments for %1%", mc);
                         auto primitive = mkPrimitive("execute_meter", result);
                         auto parameters = mkParameters(primitive);
+                        addSrcInfoData(primitive, s);
                         auto mtr = new Util::JsonObject();
                         mtr->emplace("type", "meter_array");
                         mtr->emplace("value", extVisibleName(em->object));
@@ -943,6 +978,7 @@ JsonConverter::convertActionBody(const IR::Vector<IR::StatOrDecl>* body,
                     if (em->method->name == v1model.registers.read.name) {
                         auto primitive = mkPrimitive("register_read", result);
                         auto parameters = mkParameters(primitive);
+                        addSrcInfoData(primitive, s);
                         auto dest = conv->convert(mc->arguments->at(0));
                         parameters->append(dest);
                         parameters->append(reg);
@@ -952,6 +988,7 @@ JsonConverter::convertActionBody(const IR::Vector<IR::StatOrDecl>* body,
                     } else if (em->method->name == v1model.registers.write.name) {
                         auto primitive = mkPrimitive("register_write", result);
                         auto parameters = mkParameters(primitive);
+                        addSrcInfoData(primitive, s);
                         parameters->append(reg);
                         auto index = conv->convert(mc->arguments->at(0));
                         parameters->append(index);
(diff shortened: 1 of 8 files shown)


backends/bmv2/jsonconverter.cpp
```diff
@@ -803,13 +805,34 @@ JsonConverter::JsonConverter(const CompilerOptions& options) :
         headerParameter(nullptr), userMetadataParameter(nullptr), stdMetadataParameter(nullptr)
 {}
 
+void addSrcInfoData(Util::JsonObject* result,
```
cc10512: It would be helpful to take advantage of ```SourceFileLine``` (see lib/source_file.h:221) and add a toJson(Util::JsonObject *output) method. Then you can simply ask each node to print its SourceFileLine object.

lib/source_file.cpp
```diff
@@ -226,6 +226,36 @@ cstring InputSources::getSourceFragment(const SourceInfo &position) const {
     return result + toadd + marker + cstring::newline;
 }
 
+cstring InputSources::getSourceFragment2(const SourceInfo &position) const {
```
sethfowler: I haven't had a chance to really review this, but one initial comment is that this method needs a more descriptive name, and ideally a Doxygen comment describing what it does.

lib/source_file.cpp
```diff
@@ -226,6 +226,36 @@ cstring InputSources::getSourceFragment(const SourceInfo &position) const {
...
+            --end;
+        }
+        toadd = " ...";
+    } else {
+        end = position.getEnd().getColumnNumber();
+    }
+    len = end - start;
+
+    char substr[result_len+1];
+    memcpy(substr, &result[start], len);
```
sethfowler: It'd be better to use [this](https://github.com/p4lang/p4c/blob/master/lib/cstring.h#L147) instead of memcpy.

sethfowler:
Thanks for this patch, by the way! I agree that this could make the debugging experience in BMV2 much nicer.

lib/source_file.cpp
jafingerhut: This is exactly the kind of thing I was hoping to make better as a result of feedback.  Thanks for pointing out that existing method I could use.

backends/bmv2/jsonconverter.cpp
jafingerhut: Some comments claim that SourceFileLine contains column numbers, but the implementation does not.  Should that be added?

Also, are you suggesting that _every_ node have and print a SourceFileLine object?  I'm not sure that is what we want for nested expression trees, for example (((x+5) ^ (y-7)) - 1) should probably have source info once for the root node, not for every sub-node.

Also there are nodes synthesized during compilation that either shouldn't have, or might in general be difficult to create for them, source location information.  Are you suggesting that for such nodes the toJson() method should be a no-op?

lib/source_file.cpp
jafingerhut: Agreed for sure on the name.  It was a placeholder name when I wrote it.

Do you have a pointer to a Doxygen comment I could use as an example to pattern a new one after?  source_file.cpp and source_file.h don't seem to have any right now.

backends/bmv2/jsonconverter.cpp
mbudiu-vmw: The comment is on SourcePosition.
There are nodes synthesized which won't have source position.

backends/bmv2/jsonconverter.cpp
```diff
@@ -803,13 +805,60 @@ JsonConverter::JsonConverter(const CompilerOptions& options) :
...
+                    const IR::Node* node) {
+    unsigned lineNumber, columnNumber;
+    cstring fName = node->srcInfo.toSourcePositionData(&lineNumber,
+                                                       &columnNumber);
+    if (fName == NULL) {
```
mbudiu-vmw: The preferred way in C++ is to use nullptr instead of NULL.

lib/source_file.cpp
```diff
@@ -242,13 +265,29 @@ cstring SourceInfo::toSourceFragment() const {
     return InputSources::instance->getSourceFragment(*this);
 }
 
+cstring SourceInfo::toSourceFragment2() const {
```
mbudiu-vmw: This method should probably be called toBriefSourceFragment or something similar.

lib/source_file.cpp
```diff
@@ -242,13 +265,29 @@ cstring SourceInfo::toSourceFragment() const {
...
 cstring SourceInfo::toPositionString() const {
     if (!this->isValid())
         return "";
     SourceFileLine position = InputSources::instance->getSourceLine(this->start.getLineNumber());
     return position.toString();
 }
 
+cstring SourceInfo::toSourcePositionData(unsigned *outLineNumber,
+                                         unsigned *outColumnNumber) const {
+    SourceFileLine position = InputSources::instance->getSourceLine(this->start.getLineNumber());
```
mbudiu-vmw: This is the oldest file in the compiler source tree; in the meantime we gave up on using `this->`. 
I will submit a fix removing it from everywhere.

lib/source_file.cpp
```diff
@@ -226,6 +226,29 @@ cstring InputSources::getSourceFragment(const SourceInfo &position) const {
...
+
+    cstring result = this->getLine(position.getStart().getLineNumber());
+    unsigned int start = position.getStart().getColumnNumber();
+    unsigned int end;
+    cstring toadd = "";
+
+    // If the position spans multiple lines, truncate to just the first line
+    if (position.getEnd().getLineNumber() > position.getStart().getLineNumber()) {
+        // go to the end of the first line
+        end = strlen(result);
```
mbudiu-vmw: cstring has its own size() method.

backends/bmv2/jsonconverter.cpp
```diff
@@ -836,6 +885,9 @@ JsonConverter::convertActionBody(const IR::Vector<IR::StatOrDecl>* body,
                                  Util::JsonArray* result, Util::JsonArray* fieldLists,
                                  Util::JsonArray* calculations, Util::JsonArray* learn_lists) {
     for (auto s : *body) {
+        // TBD jafinger - add line/col at all individual cases below,
```
mbudiu-vmw: so far we have TODO, FIXME, and a few XXX.
Perhaps we should use TODO instead of TBD: makes it simpler to grep.

mbudiu-vmw: Also, the recommended syntax (by cpplint) is
`TODO(jafinger)`

mbudiu-vmw:
The travis job failed because of cpplint; please run `make cpplint` and fix the reported issue.

backends/bmv2/jsonconverter.cpp
jafingerhut: @cc10512 OK, thinking more about your suggestion here, and how precisely to do it.  The existing code in jsonconverter.cpp walks through a bunch of IR::Node structures, sometimes recursively, in a custom way for each type of Node object.  For each one of those that has a valid srcInfo field (type SourceInfo), I want not to print JSON, but to add some elements using the emplace() method to an in-memory JSON object that is being filled in, and will later be printed out after it is completely created.

You are suggesting that instead of all of these calls to addSrcInfoData() that I have created, each of them should instead be a call to a new IR::Node method that has a signature something like Node::addSourceInfoToJson(Util::JsonObject *jsonObj) that does the emplace() appropriate emplace() calls on that jsonObj?

If I do that, all of the addSrcInfoData() function calls that this patch adds would change to method calls to addSourceInfo().  Is that what you hope to see?

lib/source_file.cpp
jafingerhut: Understood.  My reason for writing it that way is that I copied, pasted, and edited this method from toPositionString() above it, and it seems to be working as desired.

backends/bmv2/jsonconverter.cpp
cc10512: Hi Andy,
Here are my thoughts -- not quite at the level of CPR though :) Note that in order to be a cleaner implementation, this might require a bit more work on your part, but I believe that in the long run it will adhere better to the design that we have:
1. The JSONConverter should really be a visitor pass. It has not been implemented as such, but @hanw is working on refactoring it so that it conforms to the rest of our passes.
2. Each IR node should know its source information (or know that it doesn't have it). We should define an interface that will be "mixed-in" (or even declare these methods in the IR::Node itself) so that when a node is visited it can output the source info -- just as you described before with the Node::addSourceInfoToJson(Util::JsonObject *jsonObj) function.
3. You can pick and chose which nodes output the info by adding it to the visitor functions.

The reason I'm proposing this is that you will visit each node in the JSONConverter visitor and moreover, outputing source info is not separable into another visitor (unless you can somehow re-traverse the whole JSON tree). The addSrcInfoData functions that you wrote would work well for such a separate visitor. By encapsulating the source info into the IR node, you can also carry it through transforms (e.g., if you inline a table and create clone nodes, both sites should refer to the original source).

Let me know what you think.

jafingerhut: @cc10512: When you say "it will adhere better to the design that we have", do you mean "it will adhere better to the design we hope to have in the near future"?  Because the current code doesn't work as you describe.

It seems I can put in something similar to what I have in this PR now, before that refactor, and let @hanw include these enhancements in his refactoring effort, or I can wait until after that refactoring effort is complete, and do as you suggest.  I suppose a third alternative is to do the refactoring that @hanw is planning to do, but I don't think I want to deny him that pleasure. :-)

cc10512: Umm, the design with visitors and mix-ins is there :) Unfortunately the implementation of the JSON converter is not conforming to the design. 

Or you could work together with @hanw and refactor it twice as fast! :)

jafingerhut:
I have made some changes similar to what @cc10512 suggested, at least in terms of adding a new method to the Node class that creates a JSON object in memory populated with key/value pairs for filename, line, column, and source_fragment.  This is added as a nested object inside of each relevant node, with a new key "source_info".  This enables adding new keys to the "source_info" object in the future without any chance of colliding with other keys people might want to add to nodes.

I have not attempted to do the code refactoring that Calin suggested is coming soon.  I have made every attempt so that when such refactoring is done, it is no harder to do after these changes, than without these changes.

Please let me know if you have any other comments.  I would prefer to commit these without first doing the refactoring Calin mentioned, but I'll leave that for the project owners to decide/prioritize.

hanw:
Thanks Andy, these changes looks good. I will rebase the refactored code on top of them.

lib/source_file.cpp
jafingerhut: @mbudiu-vmw Is this a change you would strongly prefer be part of this PR, or is it OK for it to wait until a later full pass through this file?

mbudiu-vmw: If you merge with master you will notice that `this->` has been removed.

jafingerhut: Doh!  OK, thanks for that.  I will now try to hit the big scary Github "Update branch" button and see if that causes the state of this PR to become unreadable.  If so, I will create a new separate PR that is clean but won't have all the nice review comments on it.

ChrisDodd: Generally the way to deal with this (upstream changes to master you want in your working branch that you've already made a pull request for) is to use

    git pull --rebase upstream master

(or possibly `origin master` if your local repo is clone of p4c.org/p4c with your fork added, rather than a clone of a fork, with p4c.org/p4c added as upstream)

Followed by

    git push -f

to update the pull request in place

jafingerhut:
I believe I have addressed all current requests for changes, including @mbudiu-vmw comment about running cpplint, and removing occurrences of 'this' from the code, both the ones I added and a few that were already there before my changes.

Github still thinks there are requested changes that are not addressed yet.  If anyone can see what those are, or has more changes they would like, please let me know.

jafingerhut:
I believe I have addressed all current requests for changes, including @mbudiu-vmw comment about running cpplint, and removing occurrences of 'this' from the code, both the ones I added and a few that were already there before my changes.

Github still thinks there are requested changes that are not addressed yet.  If anyone can see what those are, or has more changes they would like, please let me know.

jafingerhut:
@antoninbas Wanted to check with you quickly -- if the objects in a bmv2 JSON file have an extra key "source_info", which is itself a nested JSON object with keys like "filename", "line", "column", and "source_fragment", bmv2 should hopefully just silently ignore those right now, yes?

antoninbas:
@jafingerhut there should be no issue

jafingerhut:
@mbudiu-vmw I know you must be crazy busy, but Github's software is still in a state that this issue is marked as blocked for merging, because changes were requested, I think the ones requested by you.  I believe all changes requested have been implemented, and I think Github's state somehow got stale for this PR.  If you get some time, and could see if you approve of the current PR, that would be great.

antoninbas:
@jafingerhut In theory you do not need Mihai to change his review to be able to merge, since you already have an "approved review" from Calin. However, your branch is out of sync with master, but I can click the "Update branch" button for you.

antoninbas:
I have updated the branch, you should be able to merge once the Travis regression completes. If you wait too long, you may have to update the branch again.

ChrisDodd:
As a general rule, if you have a change like this with many separate changes where things are perhaps in an invalid state between them, please "squash and merge" to make a single merge commit.

jafingerhut:
@antoninbas Understood.  This being my first time committing anything to this project, the last thing I want to do is merge it in myself if people were expecting that someone else besides me should do it.  If you are saying that I am good to merge it in with the 2 approvals it has plus clean Travis build results, I am happy to give it a go.

jafingerhut:
@antoninbas I am a member of this project, not an owner.  I see no Merge button available to me, probably because of my member status.  I am perfectly content with that status, but it appears someone else must be the one to do the merge.

PR #514, opened 2017-04-24
Fix invalid dereferencing in bmv2/inlining.cpp

antoninbas:
Thanks to @teverman for reporting this issue. Only some compilers actually complain about this, which is very confusing to me...


antoninbas:
@teverman I submitted your patch.

ChrisDodd:
Ah -- I see the propblem now -- this code is not currently built at all; this source file does not appear in Makefile.am.  If we need it, it should probably be added to the makefile.

antoninbas:
That's reassuring... @ChrisDodd and @mbudiu-vmw should we remove this code then if it is not used?

mbudiu-vmw:
The file is probably checked-in by mistake; let's delete it.

teverman:
So it's breaking in our environment because it just globs all the cpp files from the bmv2 directory.

antoninbas:
PR #515 removes these files, as @mbudiu-vmw suggested

PR #515, opened 2017-04-24
Remove bmv2/inlining files

antoninbas:
These files must have been checked-in by error, they are not even
built.


antoninbas:
@teverman this pull request will solve your problem, these files were actually not needed so we are removing them, they will not be picked up anymore by your build system

PR #520, opened 2017-04-25
Strength reduction polishing.

jnfoster:
This commit reviews and adds Doxygen comments for strength reduction.

### DIFF
--- a/frontends/p4/strengthReduction.cpp
+++ b/frontends/p4/strengthReduction.cpp
@@ -18,6 +18,10 @@ limitations under the License.
 
 namespace P4 {
 
+///@section Helper methods
+
+
+/** Helper that returns `true` if `expr` is the constant `1`. */
 bool StrengthReduction::isOne(const IR::Expression* expr) const {
     auto cst = expr->to<IR::Constant>();
     if (cst == nullptr)
@@ -25,6 +29,7 @@ bool StrengthReduction::isOne(const IR::Expression* expr) const {
     return cst->value == 1;
 }
 
+/** Helper that returns `true` if `expr` is the constant `0`. */  
 bool StrengthReduction::isZero(const IR::Expression* expr) const {
     auto cst = expr->to<IR::Constant>();
     if (cst == nullptr)
@@ -32,6 +37,7 @@ bool StrengthReduction::isZero(const IR::Expression* expr) const {
     return cst->value == 0;
 }
 
+/** Helper that returns `true` if `expr` is the constant `true`. */
 bool StrengthReduction::isTrue(const IR::Expression* expr) const {
     auto cst = expr->to<IR::BoolLiteral>();
     if (cst == nullptr)
@@ -39,13 +45,16 @@ bool StrengthReduction::isTrue(const IR::Expression* expr) const {
     return cst->value;
 }
 
+/** Helper that returns `true` if `expr` is the constant `false`. */
 bool StrengthReduction::isFalse(const IR::Expression* expr) const {
     auto cst = expr->to<IR::BoolLiteral>();
     if (cst == nullptr)
         return false;
     return !cst->value;
 }
-
+/** Helper that returns the logarithm (base 2) of `expr` if it is
+  * positive and a power of `2`, or `-1` otherwise. 
+  */
 int StrengthReduction::isPowerOf2(const IR::Expression* expr) const {
     auto cst = expr->to<IR::Constant>();
     if (cst == nullptr)
@@ -61,8 +70,9 @@ int StrengthReduction::isPowerOf2(const IR::Expression* expr) const {
     return log;
 }
 
-///////////////// visitor methods ////////////////////
+/// @section Visitor Methods
 
+/** See class comment. */
 const IR::Node* StrengthReduction::postorder(IR::BAnd* expr) {
     if (isZero(expr->left))
         return expr->left;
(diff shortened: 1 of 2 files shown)


frontends/p4/strengthReduction.h
```diff
@@ -26,13 +26,37 @@ limitations under the License.
...
+  * 
+  * 2. A visitor that transforms arithmetic and boolean exprssions
+  * 
+  * @pre: None
+  * 
+  * @post: 
+  *   - all arithmetic and boolean expressions are simplified
+  *   - check for div / mod by `0`
+  */ 
+// TODO:
```
sethfowler: I'm not totally sure about this, but it's possible that separating the Doxygen comment from the class definition with these non-Doxygen comments will confuse Doxygen. Did you check?

One alternative is just to move the block of TODO comments one line down, inside `class StrengthReduction`.

frontends/p4/strengthReduction.h
```diff
@@ -26,13 +26,37 @@ limitations under the License.
...
 
+/** Implements a pass that replaces expensive arithmetic and boolean
+  * operations with cheaper ones -- i.e., strength reduction
+  * 
+  * Specifically, it provides:
+  * 
+  * 1. A collection of helper methods that determine whether a given
+  * expression is `0`, `1`, `true`, or `false`, or a power of `2`
+  * 
+  * 2. A visitor that transforms arithmetic and boolean exprssions
```
cole-barefoot: s/exprssions/expressions/

frontends/p4/strengthReduction.h
```diff
@@ -26,13 +26,37 @@ limitations under the License.
...
     bool isZero(const IR::Expression* expr) const;
-    bool isFalse(const IR::Expression* expr) const;
+    /// Helper that returns `true` if `expr` is the constant `true`.
     bool isTrue(const IR::Expression* expr) const;
-    // if expr is a constant value that is a power of 2 return the log2 of it,
-    // else return -1
+    /// Helper that returns `true` if `expr` is the constant `false`.
+    bool isFalse(const IR::Expression* expr) const;
+    /** Helper that returns the logarithm (base 2) of `expr` if it is
+     * positive and a power of `2`, or `-1` otherwise.
```
sethfowler: It's fine both ways, but just as an FYI, another option you have to write this in Doxygen is:

```
/// Helper that returns the logarithm (base 2) of `expr` if it is
/// positive and a power of `2`, or `-1` otherwise. 
```

In other words, consecutive lines that start with `///` will be joined.

In the past, I've often used the rule of thumb that for one or two lines, `///` is more readable, and for three or more `/** ... */` is more readable. It's a matter of taste, though.

frontends/p4/strengthReduction.h
```diff
@@ -26,13 +26,37 @@ limitations under the License.
...
+  * 
+  * @post: 
+  *   - all arithmetic and boolean expressions are simplified
+  *   - check for div / mod by `0`
+  */ 
+// TODO:
+//   - overlap with constant folding?
+//   - store constant values in IR?
 class StrengthReduction final : public Transform {
+    /// Helper that returns `true` if `expr` is the constant `1`.
```
sethfowler: Conventionally you'd use `@return` here, as in:

```
/// @return `true` if `expr` is the constant `1`.
```

This excludes the word `helper`, but my view is that there's a stronger way to indicate that a method is a helper, and that's to make it protected or private. (And these methods are.)

frontends/p4/strengthReduction.cpp
```diff
@@ -169,6 +170,7 @@ const IR::Node* StrengthReduction::postorder(IR::Mul* expr) {
 }
 
 const IR::Node* StrengthReduction::postorder(IR::Div* expr) {
+    // TODO: can this happen? Constant folding also checks for the same condition.
```
mbudiu-vmw: The fewer preconditions we have, the more robust the code is.
We can't be sure that constant folding has recently been run.

frontends/p4/strengthReduction.h
```diff
@@ -26,13 +26,37 @@ limitations under the License.
...
+  * 
+  * 1. A collection of helper methods that determine whether a given
+  * expression is `0`, `1`, `true`, or `false`, or a power of `2`
+  * 
+  * 2. A visitor that transforms arithmetic and boolean exprssions
+  * 
+  * @pre: None
+  * 
+  * @post: 
+  *   - all arithmetic and boolean expressions are simplified
```
mbudiu-vmw: I don't know if "all" expressions are simplified. Some of them are.

frontends/p4/strengthReduction.h
```diff
@@ -26,13 +26,37 @@ limitations under the License.
...
+  *   - check for div / mod by `0`
+  */ 
+// TODO:
+//   - overlap with constant folding?
```
mbudiu-vmw: I find these comments obscure; I don't know what is left to do.

frontends/p4/strengthReduction.h
```diff
@@ -26,13 +26,37 @@ limitations under the License.
...
+//   - store constant values in IR?
 class StrengthReduction final : public Transform {
+    /// Helper that returns `true` if `expr` is the constant `1`.
     bool isOne(const IR::Expression* expr) const;
+    /// Helper that returns `true` if `expr` is the constant `0.
     bool isZero(const IR::Expression* expr) const;
-    bool isFalse(const IR::Expression* expr) const;
+    /// Helper that returns `true` if `expr` is the constant `true`.
```
mbudiu-vmw: For these functions I don't think that the comment really adds much. The function name is quite descriptive.

jnfoster: Looks better with Doxygen though.

PR #521, opened 2017-04-25
Add support for new ActionRef message in p4runtime

antoninbas:
In reaction to the following PI update: https://github.com/p4lang/PI/pull/135

### DIFF
--- a/control-plane/p4RuntimeSerializer.cpp
+++ b/control-plane/p4RuntimeSerializer.cpp
@@ -231,6 +231,12 @@ struct MatchField {
                                         // to this field.
 };
 
+struct ActionRef {
+    const cstring name;  // The fully qualified external name of the action.
+    const IR::IAnnotated* annotations;  // If non-null, any annotations applied to this action
+                                        // reference in the table declaration.
+};
+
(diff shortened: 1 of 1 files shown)


control-plane/p4RuntimeSerializer.cpp
```diff
@@ -1020,7 +1026,12 @@ class P4RuntimeSerializer {
         }
 
         for (const auto& action : actions) {
-            auto id = symbols.getId(P4RuntimeSymbolType::ACTION, action);
+            auto id = symbols.getId(P4RuntimeSymbolType::ACTION, action.name);
+            auto action_ref = table->add_action_refs();
+            action_ref->set_id(id);
+            addAnnotations(action_ref, action.annotations);
+            // XXX(antonin): This is temporary for backward-compatibility; can be removed when all
+            // consummers gave moved from action_ids to action_refs.
```
sethfowler: s/consummers gave/consumers get

antoninbas:
annotations on actions show up in the `Action` message

sethfowler:
Ah yes, that's a good point. We now have both sets of annotations exposed, which is a good thing.

PR #523, opened 2017-04-25
documentation for IR in base.def

cc10512:
This PR started as documenting the IR. In the meantime several other fixes creeped in:
1. Fixing the ir-generation to include all the lines in a multiline comment (commit 3f47a59)
2. forcing the rebuild of documentation (commit 041f64b)
3. the comments themselves (commit 3a25ee8)


ir/base.def
```diff
@@ -32,18 +32,25 @@ abstract Type {
...
 }
 
-/// Implemented by all types that may be generic
+/// Implemented by all types that may be generic:
+/// Control, Extern, Method, Package, and Parser
 interface IMayBeGenericType {
+    /// Allows the retrieval of type parameters
     virtual const TypeParameters* getTypeParameters() const = 0;
-    const Type* toType() const { return to<Type>(); }
+    // cast to a Type object.
```
mbudiu-vmw: It should be removed completely, not just commented-out.

ir/base.def
```diff
@@ -32,18 +32,25 @@ abstract Type {
...
+    // There is a single invocation of this method (typeChecker.cpp:92) so removed
+    // const Type* toType() const { return to<Type>(); }
 }
 
-/// Implemented by objects with an 'apply' method
+/// Implemented by objects with an 'apply' method: Parser, Control, Table
+/// \TODO: Unclear why the corresponding Type objects of these objects also implement IApply
```
mbudiu-vmw: This comment does not belong here.

ir/base.def
```diff
@@ -141,6 +150,8 @@ abstract Operation : Expression {
     toString{ return getStringOp(); }
 }
 
+/// A Path represents a fully qualified name for a P4 object
+/// global (absolute) names are prefixed by a '.'
```
mbudiu-vmw: Today paths can only consist of a dot; we had a design where paths could be more complicated, but it is not currently used.

ir/type.def
```diff
@@ -458,6 +458,10 @@ class Type_MethodCall : Type {
 
 /// Actions look a lot like methods in many respects.
 /// However, invoking an action returns another action
+/// \TODO(mbudiu): where is the constraint of returning an action captured?
```
mbudiu-vmw: That is part of the type-checker.
The methods may be the same, but the pre/postorder methods will be different.
Also, look at methodInstance.h, in particular the class ActionCall.

tools/ir-generator/ir-generator.ypp
```diff
@@ -142,6 +142,16 @@ element
     : irclass           { global.push_back($1); }
     | COMMENTBLOCK      { global.push_back(new CommentBlock(@1, $1)); }
     | COMMENTBLOCK irclass {
+            /* find all the comments that have been pushed on the global stack */
+            while (global.size() > 0) {
+                auto c = dynamic_cast<CommentBlock *>(global.back());
+                if (c != nullptr) {
+                    $2->comments.push_back(c);
+                    global.pop_back();
```
ChrisDodd: This reverses the order of the comments.  If you want to pull in all the comments immediately before the class, get rid of the `COMMENTBLOCK irclass` rule and scan over the trailing comments to pull them into the class in the bare `irclass` rule.  Alternately, add a rule to consecutive comments into a single comment:

    comment_block : COMMENTBLOCK { $$ = new CommentBlock(@1, $1); }
        | comment_block COMMENTBLOCK {
            $$ = $1;
            $$->srcInfo += @2;
            $$->body += $2; }

ir/base.def
cc10512: @mbudiu-vmw I will remove the comment, but do you mind explaining?

mbudiu-vmw: I assume that you mean Type_Parser and Type_Control.
I think that in the end only the Type_* objects have the required information for implementing this interface, P4Parser and P4Control only delegate the query to them. I think that you should think of Type_Parser as an interface, and as P4Parser as a class implementing the interface. Then IApply is just a base interface extended by Type_Parser.

tools/ir-generator/ir-generator.ypp
cc10512: ah, yes, i should have pushed in front, but your solution is more elegant.

ir/base.def
cc10512: Thank you for that explanation. So if I understand correctly, for both objects and corresponding Type_objects, the call to ```getApplyMethodType()``` will actually resolve in the Type_object. Any particular reason (except for C++ inclusion) why the P4Parser does not inherit from Type_Parser?

mbudiu-vmw: We don't really use inheritance for IR classes; only abstract classes and interfaces are subclassed.
In other words, today all IR concrete classes are final.
Maybe we could use inheritance, I don't know if there are technical reasons.

cc10512:
@mbudiu-vmw, @ChrisDodd, thank you for your comments. I updated the PR.

PR #402, opened 2017-03-28
Handle list literals in IR::Cast case in constantFolding.cpp

jnfoster:
The IR::Cast case had a bug where it assumed that a constant was either a boolean or numeric literal. This pull request adds an extra case to handle list literals.

Fixes #401.

PR #525, opened 2017-04-26
Polish constant folding

jnfoster:
This commit reviews and adds Doxygen comments for constant folding.

### DIFF
--- a/frontends/common/constantFolding.h
+++ b/frontends/common/constantFolding.h
@@ -28,19 +28,16 @@ namespace P4 {
  * expressions.
  *
  * The pass can be invoked either with or without the reference and
- * type maps. When this information is not available, then several
- * cases revert to the identify function.
- *
- * The pass also eliminates arbitrary-precision integer literals,
- * replacing them with literals that have explicit widths.
+ * type maps. When type information is not available, constant folding
+ * is not done for some IR nodes.
  *
  * @pre: None
  *
  * @post: Ensures that
- *    - some constant expressions are eliminated (TODO: specify)
+ *    - some constant expressions are eliminated (TODO: specify?)
+ * 
  *    - no operations that involve only int-typed values; at least one
-        operand will have a known width, which is used to cast the
-        other operand.
+ *      operand will have a known width. 
  */
 class DoConstantFolding : public Transform {
  protected:
--- a/frontends/common/constantFolding.h
+++ b/frontends/common/constantFolding.h
@@ -23,36 +23,97 @@ limitations under the License.
 #include "frontends/p4/typeChecking/typeChecker.h"
 
 namespace P4 {
-// Can also be run before type checking, so it only uses types if
-// they are available.
+
+/** Implements a pass that statically evaluates many constant
+ * expressions.
+ *
+ * Currently the following expressions are treated as constants:
+ *  - `IR::Constant`
+ *  - `IR::BoolLiteral`
+ *  - `IR::ListExpression` where each element is also a constant
+ *  - `IR::TypeNameExpression`
+ *
+ * The pass can be invoked either with or without the reference and
+ * type maps. When this information is not available, then several
+ * cases revert to the identify function.
+ *
+ * The pass also eliminates arbitrary-precision integer literals,
+ * replacing them with literals that have explicit widths.
+ *
+ * @pre: None
+ *
+ * @post: Ensures that
+ *    - some constant expressions are eliminated (TODO: specify)
+ *    - arbitrary-precision integers replaced with fixed-width integers (TODO: confirm)
+ */
 class DoConstantFolding : public Transform {
  protected:
-    const ReferenceMap* refMap;  // if null no 'const' values can be resolved
-    TypeMap* typeMap;  // if null we have no types; updated for new constants
+    /// Used to resolve IR nodes to declarations.
+    /// If `null`, then `const` values cannot be resolved.
+    const ReferenceMap* refMap;
+
+    /// Used to resolve nodes to their types.
+    /// If `null`, then type information is not available.
+    TypeMap* typeMap;
+
+    /// Set to `true` iff type information is available.
     bool typesKnown;
-    bool warnings;  // if true emit warnings
-    // maps expressions and declarations to their constant values
+
+    /// If `true` then emit warnings
+    bool warnings;
+
+    /// Maps IR nodes to constants
     std::map<const IR::Node*, const IR::Expression*> constants;
 
  protected:
-    // returns nullptr if the expression is not a constant
+    /** Check whether @p expr is a constant. @returns @p expr if a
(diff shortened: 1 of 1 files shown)


frontends/common/constantFolding.h
```diff
@@ -23,36 +23,97 @@ limitations under the License.
...
 
 namespace P4 {
-// Can also be run before type checking, so it only uses types if
-// they are available.
+
+/** Implements a pass that statically evaluates many constant
+ * expressions.
+ *
+ * Currently the following expressions are treated as constants:
+ *  - `IR::Constant`
```
jnfoster: Mihai: Also, Declaration_Constant

frontends/common/constantFolding.h
```diff
@@ -23,36 +23,97 @@ limitations under the License.
...
+ *  - `IR::BoolLiteral`
+ *  - `IR::ListExpression` where each element is also a constant
+ *  - `IR::TypeNameExpression`
```
jnfoster: This also handles constants of enum types. Perhaps it should also handle constants of the error type...

frontends/common/constantFolding.h
```diff
@@ -23,36 +23,97 @@ limitations under the License.
...
+ *
+ * The pass can be invoked either with or without the reference and
+ * type maps. When this information is not available, then several
+ * cases revert to the identify function.
```
jnfoster: Mihai: for some IR classes constant folding is not done in the absence of type information.

frontends/common/constantFolding.h
```diff
@@ -23,36 +23,97 @@ limitations under the License.
...
+ *
+ * The pass also eliminates arbitrary-precision integer literals,
+ * replacing them with literals that have explicit widths.
```
jnfoster: Mihai: This is incorrect, that is done by the type checker. There isn't enough information to do this here.

frontends/common/constantFolding.h
```diff
@@ -23,36 +23,97 @@ limitations under the License.
...
 
  protected:
-    // returns nullptr if the expression is not a constant
+    /** Check whether @p expr is a constant. @returns @p expr if a
+     *  constant and `nullptr` if not.  The result is guaranteed to have one
+     *  of the following types:
+     *  - `IR::Constant`
+     *  - `IR::BoolLiteral`
+     *  - `IR::ListExpression`, where each element is also a constant
+     *  - `IR::TypeNameExpression`
```
jnfoster: Mihai: I don't think this is correct. This can be a Member(TypeName), where TypeName denotes an enum type.

frontends/common/constantFolding.h
```diff
@@ -23,36 +23,97 @@ limitations under the License.
...
+     * 
+     * @todo should we change the return type to a variant?
```
jnfoster: A variant would not solve anything, and would make the code much more complicated.

jnfoster:
@mbudiu-vmw I'm sorry, my previous pull request was in a bad state. I created a new one and manually copied your comments over to this one, prefixed by "Mihai".

frontends/common/constantFolding.h
jnfoster: Indeed, due to the use of the `constant` map, essentially arbitrary Expressions can be returned from `getConstant`. 

For example, the case for `LAnd` places `e->right` in the map, which could be anything (including something that is not a constant).

frontends/common/constantFolding.h
jnfoster: Better.

frontends/common/constantFolding.h
jnfoster: Fixed.

frontends/common/constantFolding.h
jnfoster: Indeed. My mistake. It can be many different kinds of expressions.

frontends/common/constantFolding.h
jnfoster: Again, my mistake. I thought only those types could get into `constants`. A variant would let the code that eliminates constants use the type system to check that all cases had been covered, avoiding situations like #402. I removed the comment.

frontends/common/constantFolding.h
```diff
@@ -23,36 +23,81 @@ limitations under the License.
...
+ * The pass can be invoked either with or without the reference and
+ * type maps. When type information is not available, constant folding
+ * is not done for some IR nodes.
+ *
+ * @pre: None
+ *
+ * @post: Ensures that
+ *    - some constant expressions are eliminated (TODO: specify?)
+ * 
+ *    - no operations that involve only int-typed values; at least one
```
mbudiu-vmw: This is not true. This invariant is established by the type-checker, not the constant folding pass.

frontends/common/constantFolding.h
```diff
@@ -23,36 +23,81 @@ limitations under the License.
...
+ * expressions.
+ *
+ * The pass can be invoked either with or without the reference and
+ * type maps. When type information is not available, constant folding
+ * is not done for some IR nodes.
+ *
+ * @pre: None
+ *
+ * @post: Ensures that
+ *    - some constant expressions are eliminated (TODO: specify?)
```
mbudiu-vmw: I would rather say that some expressions that provably evaluate to a constant are replaced with the constant value.

frontends/common/constantFolding.h
```diff
@@ -23,36 +23,81 @@ limitations under the License.
...
-    // returns nullptr if the expression is not a constant
+    /// @returns a constant equivalent to @p expr or `nullptr`
     const IR::Expression* getConstant(const IR::Expression* expr) const;
+
+    /// Add a mapping from @p node to @p result in `constants`. Note that
+    /// this method must be called by a visitor as it invokes
+    /// @ref Visitor#getOriginal.
     void setConstant(const IR::Node* expr, const IR::Expression* result);
 
+    /// Cast constant @p node to @p type at @p base.
```
mbudiu-vmw: Represented in the specified base.

frontends/common/constantFolding.h
jnfoster: It is a paraphrase from something you wrote...

> The spec also indicates that operations on int-typed objects (InfInt) are done using infinite precision, producing int results. The constant folding does these operations. After constant folding there are no operations which involve only int-typed values, at least one operand has a known width, which is used to cast the other operand. 

How do we make it accurate?

mbudiu-vmw: OK, maybe I misread the comment, it is roughly accurate, but not easy to parse.
Let's say that (all) operations that involve constant InfInt operands are evaluated to an InfInt value.

jnfoster: Sounds good. I'll update.

jnfoster:
Subsumed by #541. Closing.

PR #526, opened 2017-04-26
[do not merge]: replace typeMap->constants with IR-based constants

cc10512:
This PR attempts to replace the typeMap->isCompileTimeConstant/setCompileTimeConstant with that functionality handled by IR nodes.

Note that less than half of all tests pass with this change. Most fail with the dreaded:
```
terminating with uncaught exception of type Util::CompilerBug: COMPILER BUG: ../frontends/p4/typeChecking/typeChecker.cpp:118
<P4Program>(19779): typechecker mutated node <P4Program>(17743)
```
@mbudiu-vmw: I need help in figuring out which of the IR nodes in typeChecker is missing the right information. I traced all the changes and they seem to capture all the places where constants were set. For one of the test cases (testdata/p4_16_samples/inline-stack-bmv2.p4), I tracked this down to invoking the apply method on the aux instance in the sequence below. Commenting out the aux.apply line allows the program to go through.
```
control X {
  Aux();
  apply {
     aux.apply(hdr);
  }
}
```

### DIFF
--- a/frontends/common/constantFolding.cpp
+++ b/frontends/common/constantFolding.cpp
@@ -495,7 +495,6 @@ const IR::Node* DoConstantFolding::postorder(IR::Member* e) {
         result = list->components.at(index)->clone();
     }
     typeMap->setType(result, origtype);
-    typeMap->setCompileTimeConstant(result);
     setConstant(e, result);
     return result;
 }
@@ -670,7 +669,6 @@ const IR::Node *DoConstantFolding::postorder(IR::Cast *e) {
         auto result = expr->clone();
         auto origtype = typeMap->getType(getOriginal());
         typeMap->setType(result, origtype);
-        typeMap->setCompileTimeConstant(result);
         setConstant(e, result);
         return result;
     }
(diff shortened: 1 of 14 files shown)


cc10512:
Actually it should be used in constantFolding, no?
Also seems to be used in sideEffects, and I would expect to use it in the backend as well.

mbudiu-vmw:
No, constant folding maps nodes to actual values.
This pass only implements section 16.1 from the Spec, "compile-time known values". It only checks to see whether all arguments for directionless parameters can be evaluated statically. It is logically part of type-checking.

SideEffects does not need this, it just updates the information so it remains correct.
None of the public back-ends uses this information.

sethfowler:
Leaving aside the issue of *whether* we should do this, this came up again today and I thought it might be worth recording here that "isLeftValue()" is something else we could store directly on the nodes instead of in a table if we so chose.

ChrisDodd:
`isLeftValue` seems to be redundant with `P4WriteContext`, which tells you
whether an expression is used an an lvalue or not.  Storing it on/in a node
or in an external table indexed by node is impossible, as a node may be
referenced in mulitple places, some of which are lvalues and some of which
are not.  This is a fundamental problem with the typeMap approach and leads
to issues with dags (which is why some other passes make redundant/useless
clones -- to work around the limitations of the typeMap).

mbudiu-vmw:
@cc10512 : are you still interested to refactor this piece of code?
We can file an issue about it and close this PR.

mbudiu-vmw:
Since this is 3 years old I will close it.

Issue #366, opened 2017-03-18
Incorrect @name annotation for key


mbudiu-vmw:
See test p4_14_samples/simple_nat.p4 - the name generated for keys containing isValid() looks suspicious:

hdr.ipv4.isValid(): exact @name("hdr.ipv4.isValid()")

mbudiu-vmw:
@sethfowler : there are several issues related to naming things in the control-plane, including this one, #582, #595, #486. Maybe we can try to coordinate and find a solution to all of them?

sethfowler:
Yeah, I agree. I have a couple of other tasks I need to complete first, but this is something we need to address soon if possible.

We put some of this stuff off while we discussed language changes that unfortunately didn't end up materializing.

mbudiu-vmw:
I am assuming that this issue can now be closed.

PR #528, opened 2017-04-26
Add Doxygen comments for the TableKeyExtraction pass.

cole-barefoot:


frontends/p4/tableKeyNames.cpp
```diff
@@ -32,10 +38,16 @@ class KeyNameGenerator : public Inspector {
 
     void postorder(const IR::Expression* expression) override { error(expression); }
 
+    /** Compute a name annotation for @expression.  Eg. `@name("foo.bar")` for
```
mbudiu-vmw: These comments on the postorder methods are practically useless.
This one in particular is also wrong, since a PathExpression cannot be "foo.bar".
The code of these methods is so short that it is self-explanatory; I don't think that these comments help at all.
What should be documented is the "name" field of the KeyNameGenerator; for each expression is maintains a cstring representing the tentative name of the expression. The name of an expression is built recursively from its components. if an expression does not have a name in this map, then it is too complex.

frontends/p4/tableKeyNames.cpp
```diff
@@ -57,6 +72,9 @@ class KeyNameGenerator : public Inspector {
         name.emplace(expression, l + "[" + r + "]");
     }
 
+    /** Compute a name annotation for @expression.  Eg. `@name("bits")` for
+     * `bits & 0x3` or `0x3 & bits`.
```
mbudiu-vmw: Note that this method is somewhat of a hack designed to support masks for keys, which are a P4-14 feature only.
I can argue that this should not be done here, but in the P4-14 front-end.

frontends/p4/tableKeyNames.cpp
```diff
@@ -68,10 +86,17 @@ class KeyNameGenerator : public Inspector {
             error(expression); }
     }
 
+    /** Compute a name annotation for @expression.  Eg. `@name("16w4")` for
```
mbudiu-vmw: I expect that the result would be 4 instead of 16w4.
I don't think that the control-plane (the one that consumes these annotations) supports widths.

frontends/p4/tableKeyNames.cpp
```diff
@@ -68,10 +86,17 @@ class KeyNameGenerator : public Inspector {
...
+     * `16w4`.
+     */
     void postorder(const IR::Constant* expression) override {
         name.emplace(expression, expression->toString());
     }
 
+
+    /** Compute a name annotation for @expression.  Eg. `@name("foo[0:3]")` for
```
mbudiu-vmw: The slice syntax is really MSB:LSB.

frontends/p4/tableKeyNames.cpp
```diff
@@ -81,6 +106,12 @@ class KeyNameGenerator : public Inspector {
         name.emplace(expression, e0 + "[" + e1 + ":" + e2 + "]");
     }
 
+    /** Compute a name annotation if @expression is a method call for
+     * `isValid()`.
+     *
+     * @warning All other method calls in keys will cause the compiler to
```
mbudiu-vmw: The comment is inaccurate; the compiler will not reject other method calls; this method will invoke an error if it is invoked with another method call. However, if the key already has a name annotation, the compiler will not reject the program.

frontends/p4/tableKeyNames.h
```diff
@@ -23,13 +23,6 @@ limitations under the License.
 
 namespace P4 {
 
-/**
```
mbudiu-vmw: Technically this comment is correct, this is the pass that adds the annotations.
The PassManager below just calls this one.

frontends/p4/tableKeyNames.h
```diff
@@ -38,6 +31,24 @@ class DoTableKeyNames : public Transform {
     const IR::Node* postorder(IR::KeyElement* keyElement) override;
 };
 
+/** Adds a "@name" annotation to each table key that does not have a name.
+ * The string used for the name is derived from the expression itself - if
+ * the expression is "simple" enough.  If the expression is not simple the
+ * compiler will give an error.  Simple expressions are:
+ * - .isValid(),
+ * - ArrayIndex,
+ * - BAnd (where at least one operand is constant),
```
mbudiu-vmw: I think that the second operand must be constant.
And I think that this case does not belong here.

frontends/p4/tableKeyNames.h
```diff
@@ -38,6 +31,24 @@ class DoTableKeyNames : public Transform {
     const IR::Node* postorder(IR::KeyElement* keyElement) override;
 };
 
+/** Adds a "@name" annotation to each table key that does not have a name.
+ * The string used for the name is derived from the expression itself - if
+ * the expression is "simple" enough.  If the expression is not simple the
+ * compiler will give an error.  Simple expressions are:
+ * - .isValid(),
```
mbudiu-vmw: There is an open issue #366 about handling isValid(). I don't think that is correct either, but we don't know what else to use until we have a spec for the control-plane API.

frontends/p4/tableKeyNames.h
```diff
@@ -38,6 +31,24 @@ class DoTableKeyNames : public Transform {
...
+ * - ArrayIndex,
+ * - BAnd (where at least one operand is constant),
+ * - Constant,
+ * - Member,
+ * - PathExpression,
+ * - Slice.
+ *
+ * @pre None.
+ * @post All keys have `@name` annotations.
```
mbudiu-vmw: All key fields have annotations.

frontends/p4/tableKeyNames.cpp
jnfoster: To me, the fact that some of these comments are not accurate suggests that we need to either simplify the code, or add documentation so anyone can easily understand it.

I think having Doxygen comments for most methods is reasonable for uniformity, even if the meaning is clear from the name and type signature.

Finally, if a PhD-level engineer stays up past midnight to add documentation to code that somebody else wrote, it seems only polite to avoid characterizing their effort as "practically useless" ;-)

mbudiu-vmw: I don't imply that the work is useless, that particular comment is useless. That is different.

Also, there is such a thing as over-commenting; we have no tools to keep comments in sync with code, and a wrong comment is worse than no comment. So I think we should really write comments where they really help our understanding of the code/algorithms. Sometimes code is clearer than comments.

I don't think that uniform documentation for Doxygen is a worthwhile goal.

jnfoster: Sorry, I read "These comments" more broadly and misunderstood. 

I agree that too many comments can be overkill. I would be in favor of class comments that document the assumptions / guarantees of each pass and the overall strategy used to implement it. Members of the class and helper functions may also merit comments but we should use our judgment. I agree that commenting each pre/post-order method may be overkill if it's just implementing the functionality specified in the class comment, unless there is something exceptional going on in that method.

cole-barefoot: For this pass, I'd like to see examples of the names generated from the IR.  This is much faster than, say, digging through the parser to figure out how syntax corresponds to IR objects.

For example, `_` is apparently a valid PathExpression, but---looking at the parser---will not appear in an expression inside a KeyElement.  (In fact, it's not obvious to me why the `Path` class has an `isDontCare()` method.)  On the other hand, dot-prefixed identifiers are also valid PathExpressions and may indeed appear here, in which case the name generated includes the "." prefix.

This is not obvious from the implementation of the postorder method.

I don't have a strong preference as to whether this lives as part of comments on pre-/postorder methods, on the `name` field of the `KeyNameGenerator`, or as part of the class comment for the pass itself.

frontends/p4/tableKeyNames.cpp
jafingerhut: Why do you say it is a P4-14 feature only?  The current p4c-bm2-ss compiler takes 'field_name & 0xff00 : exact;' as a key in a table with no error, and uses the mask feature in the bmv2 JSON file to represent it.  You can even put other expressions there like "field_name | 0xff00 : exact @name("foo");", but in that case the P4-16 compiler creates a temporary variable to hold the result of the expression, and then sends that as the table search key.

The P4-16 draft spec allows both of these things in the syntax.

mbudiu-vmw: Key masks are a P4-14 feature only; it is not clear that they are semantically the same as using `&` on a key. In particular, we don't know how the control-plane specifies such keys.

mbudiu-vmw: The name annotation should be inserted by the P4-14 front-end when converting a key mask.
Otherwise we should not handle this specially at all.

cole-barefoot:
I consolidated examples in the class comment rather than with each postorder method.  This, coupled with better documentation of the IR, should be clear.

frontends/p4/tableKeyNames.h
cole-barefoot: Makes sense.  I'll open a low-priority issue.

Issue #512, opened 2017-04-23
p4test gives error for what looks like legal if code inside action


jafingerhut:
See the attached P4 program, in particular action 'foo'.  The error I get from both p4test and p4c-bm2-ss is:

    error: Could not find declaration for pred

There is no line number for the error, and nothing in the program named 'pred'.

There are 3 lines inside action 'foo', as explained in the comments, such that commenting out or deleting any one of them causes the error to go away.  That might help find a root cause for this issue.
[could-not-find-pred-declaration.p4.txt](https://github.com/p4lang/p4c/files/950082/could-not-find-pred-declaration.p4.txt)

jafingerhut:
Confirmed that after the commits for this issue, the test program I submitted no longer gives an error.  Thanks for the fix.

PR #531, opened 2017-04-26
Documentation improvements

mbudiu-vmw:
This is a continuation of the fix for #512.

### DIFF
--- a/midend/local_copyprop.h
+++ b/midend/local_copyprop.h
@@ -29,7 +29,7 @@ This pass is designed to be run after all declarations have received unique
 internal names.  This is important because the locals map uses only the
 declaration name, and not the full path.
 
-@preconditions
+@pre
 Requires expression types be stored inline in the expression
 (obtained by running Typechecking(updateProgram = true)).
 
--- a/midend/local_copyprop.h
+++ b/midend/local_copyprop.h
@@ -23,15 +23,21 @@ limitations under the License.
 
 namespace P4 {
 
-/* Local copy propagation and dead code elimination within a single action function.
-   This pass is designed to be run after all declarations have received unique
-   internal names.  This is important because the locals map uses only the
-   declaration name, and not the full path.
+/**
+Local copy propagation and dead code elimination within a single pass.
+This pass is designed to be run after all declarations have received unique
+internal names.  This is important because the locals map uses only the
+declaration name, and not the full path.
 
-   Requires expression types be stored inline in the expression
-   (obtained by running Typechecking(updateProgram = true)).
-   Requires that all declaration names be globally unique
-   (obtained by running UniqueNames)
+@preconditions
+Requires expression types be stored inline in the expression
+(obtained by running Typechecking(updateProgram = true)).
+
+Requires that all declaration names be globally unique
+(obtained by running UniqueNames).
+
+Requires that all variable declarations are at the top-level control scope
+(obtained using MoveDeclarations).
  */
 class DoLocalCopyPropagation : public ControlFlowVisitor, Transform, P4WriteContext {
     bool                        working = false;


frontends/p4/moveDeclarations.h
```diff
@@ -22,13 +22,17 @@ limitations under the License.
...
-// Also, if the control has nested actions, we move the declarations from
-// the actions to the enclosing control.
+/**
+Moves all local declarations to the top-level context.
+We move all declarations in a control or parser to the "top", including
+the ones in the Control body and Parser states.
+Also, if the control has nested actions, we move the declarations from
+the actions to the enclosing control.
+
+@preconditions
```
jnfoster: Should this be `@pre` ?

frontends/p4/moveDeclarations.h
```diff
@@ -62,10 +66,14 @@ class MoveDeclarations : public Transform {
...
-// Run after MoveDeclarations.
-// Some variable declarations in the "local" section of a parser and control
-// may still have initializers; these are moved into the start state,
-// and to the beginning of the apply body repectively.
+/**
+Some variable declarations in the "local" section of a parser and control
+may still have initializers; these are moved into the start state,
+and to the beginning of the apply body repectively.
+
+@preconditions
```
jnfoster: `@pre` ?

midend/local_copyprop.h
```diff
@@ -23,15 +23,21 @@ limitations under the License.
...
-   (obtained by running Typechecking(updateProgram = true)).
-   Requires that all declaration names be globally unique
-   (obtained by running UniqueNames)
+/**
+Local copy propagation and dead code elimination within a single pass.
+This pass is designed to be run after all declarations have received unique
+internal names.  This is important because the locals map uses only the
+declaration name, and not the full path.
+
+@preconditions
```
jnfoster: `@pre` ?

midend/local_copyprop.h
```diff
@@ -23,15 +23,21 @@ limitations under the License.
...
+
+@pre
+Requires expression types be stored inline in the expression
+(obtained by running Typechecking(updateProgram = true)).
+
+Requires that all declaration names be globally unique
+(obtained by running UniqueNames).
+
+Requires that all variable declarations are at the top-level control scope
+(obtained using MoveDeclarations).
```
ChrisDodd: This is actually false -- in particular it can only eliminate declarations from actions if they are in the actions.  So it should work correctly either with or without MoveDeclarations, but the things that it can eliminate differ, and generally it will do a better job of eliminating dead code if called before MoveDeclarations.

mbudiu-vmw: Actually it failed without an insertion of MoveDeclarations.
The outline was like follows:

```
{  bit e = ...; 
    { bit tmp = e; }
    { bit tmp1 = e; }
}
```

In this case it replaced `tmp1` with `tmp`, but they are not available in the same scope.

ChrisDodd: That seems like a bug in copyprop then -- the copy should not be available after the scope ends.

mbudiu-vmw: I don't know if copyProp looks at scoping.
It is much simpler to require everything to be in scope, which I put in the preconditions.

ChrisDodd: I may have misunderstood -- Yes, copyprop does need for declarations to be moved up out of BlockStatements into the containing P4Action or P4Control.  It does not need locals of a P4Action moved out of the action into the containing P4Control, and doing so will limit its ability to dead-code eliminate those declarations.

mbudiu-vmw:
I have rebased this; perhaps we can merge and close; this is just changes in comments.

PR #536, opened 2017-04-27
writes out json boolean values for valid match keys.

cc10512:
Trivial fix to print the json boolean value.


antoninbas:
@mbudiu-vmw is that a recent thing? cause I still see empty stf files in the testdata directory: https://github.com/p4lang/p4c/blob/master/testdata/p4_16_samples/action_selector_shared-bmv2.stf

mbudiu-vmw:
A month or so. If there is no file we use empty.stf. Maybe we haven't deleted all the empty ones.

PR #541, opened 2017-04-28
constant folding streamlining, bugfixes, and Doxygen comments

jnfoster:
Do not merge yet! 

Submitting this pull request for comments. More documentation and tidying are needed 

The current version of constant folding has two bugs:

1. It does not fold constants in all cases, or check that declaration constants are initialized with genuine constants. For example, on the input program,
```
const T t = { 32s10, 32s20 };
const S s = { { 32s15, 32s25}, t };
const int<32> x = t.t1;
const int<32> y = s.s1.t2;
const int<32> w = .t.t1;
const T t1 = (T)s.s1;
```
constant folding (with type information) yields:
```
const T t = { 32s10, 32s20 };
const S s = { { 32s15, 32s25 }, t };
const int<32> x = 32s10;
const int<32> y = 32s25;
const int<32> w = 32s10;
const T t1 = { 32s15, 32s25 };
```
Note that the RHS of the declaration of `s` is not a constant -- it contains the variable `t`.

The root cause of this behavior is due to the use of auxiliary data structure `constants`: the pass does not ensure that all expressions that are inserted into this map are genuine constants, yet it assumes that `getConstants`, which returns arbitrary entries from `constants`, are always a constant. Hence, the current check for `Declaration_Constants` (https://github.com/p4lang/p4c/blob/master/frontends/common/constantFolding.cpp#L76) is not correct.

The fix here is to streamline the use of `constants` so it is used in exactly two places: 
* It is populated with genuine constants in the `IR::Declaration_Constant` case
* It is read in the `IR::PathExpression` (aka variable) case

2. When it transforms the IR, the constant folding pass attempts to patch the `typeMap` with type information about the newly inserted nodes. However, the types it inserts are not always correct -- e.g., in particular, in the `IR::Member` case, when header stack sizes are folded, the type map is updated with `origtype, the type of the expression, but the arbitrary-precision integer may require a cast. Hence, subsequent type-checking passes will rely on the information in the typeMap and assume that no cast is required. However, if the typeMap is ever cleared, then the typeChecker will mutate the node, triggering an error.

This error is not triggered by any of our current tests -- in examples like `p4_16_samples/stack.p4`, the `IR::Member` is dead code and is eliminated before the `typeMap` is cleared. Hence, the error never manifests. However, a close variant can trigger the error:
```
#include <core.p4>
header h { }
control c(out bit<32> x) {
    apply {
        h[4] stack;
        bit<32> sz = stack.size;
        x = sz;
    }
}
control Simpler(out bit<32> x);
package top(Simpler ctr);
top(c()) main;
```
The solution proposed here is to change the `typeMap` to `const` (which streamlines the constant folding pass) and reconstruct it after type checking. In addition, we statically convert the header stacks length to the appropriate type.

### DIFF
--- a/frontends/common/constantFolding.cpp
+++ b/frontends/common/constantFolding.cpp
@@ -23,9 +23,6 @@ namespace P4 {
 
 const IR::Expression* DoConstantFolding::getConstant(const IR::Expression* expr) const {
     CHECK_NULL(expr);
-    auto cst = get(constants, expr);
-    if (cst != nullptr)
-        return cst;
     if (expr->is<IR::Constant>())
         return expr;
     if (expr->is<IR::BoolLiteral>())
@@ -35,38 +32,33 @@ const IR::Expression* DoConstantFolding::getConstant(const IR::Expression* expr)
         for (auto e : list->components)
             if (getConstant(e) == nullptr)
                 return nullptr;
-        return list;
+        return expr;
     }
     if (typesKnown) {
         auto ei = EnumInstance::resolve(expr, typeMap);
         if (ei != nullptr)
             return expr;
     }
-
     return nullptr;
 }
-
-// This has to be called from a visitor method - it calls getOriginal()
-void DoConstantFolding::setConstant(const IR::Node* node, const IR::Expression* result) {
-    LOG2("Folding " << node << " to " << result << " (" << result->id << ")");
-    auto orig = getOriginal();
-    constants.emplace(node, result);
-    constants.emplace(orig, result);
-}
-
+    
 const IR::Node* DoConstantFolding::postorder(IR::PathExpression* e) {
     if (refMap == nullptr)
         return e;
     auto decl = refMap->getDeclaration(e->path);
     if (decl == nullptr)
         return e;
-    auto v = get(constants, decl->getNode());
-    if (v == nullptr)
-        return e;
-    setConstant(e, v);
-    if (v->is<IR::ListExpression>())
-        return e;
-    return v;
+    if (auto dc = decl->to<IR::Declaration_Constant>()) {
+        auto cst = get(constants, dc);
+        if (cst == nullptr)
+            cst = getConstant(dc->initializer);
+        if (cst == nullptr)
+            return e;
+        if (!typesKnown && cst->is<IR::ListExpression>())
+            return e;
+        return cst;
+    }
+    return e;
 }
 
 const IR::Node* DoConstantFolding::postorder(IR::Declaration_Constant* d) {
@@ -76,27 +68,24 @@ const IR::Node* DoConstantFolding::postorder(IR::Declaration_Constant* d) {
             ::error("%1%: Cannot evaluate initializer for constant", d->initializer);
         return d;
     }
-
-    if (typesKnown) {
-        // If we typechecked we're safe
-        setConstant(d, init);
-    } else {
-        // In fact, this declaration may imply a cast, so the actual value of
-        // d is not init, but (d->type)init.  The typechecker inserts casts,
-        // but if we run this before typechecking we have to be more conservative.
+    if (!typesKnown) {
+        // This declaration may imply a cast, so the actual value of d
+        // is not init, but (d->type)init. The typechecker inserts
+        // casts, but if we run this before typechecking we have to be
+        // more conservative.
         if (init->is<IR::Constant>()) {
             auto cst = init->to<IR::Constant>();
             if (d->type->is<IR::Type_Bits>()) {
                 if (cst->type->is<IR::Type_InfInt>() ||
                     (cst->type->is<IR::Type_Bits>() &&
-                     !(*d->type->to<IR::Type_Bits>() == *cst->type->to<IR::Type_Bits>())))
+                     !(*d->type->to<IR::Type_Bits>() == *cst->type->to<IR::Type_Bits>()))) 
                     init = new IR::Constant(init->srcInfo, d->type, cst->value, cst->base);
-                setConstant(d, init);
             }
         }
+        if (init != d->initializer) 
+            d = new IR::Declaration_Constant(d->srcInfo, d->name, d->annotations, d->type, init);
     }
-    if (init != d->initializer)
-        d = new IR::Declaration_Constant(d->srcInfo, d->name, d->annotations, d->type, init);
+    constants.emplace(getOriginal<IR::Declaration_Constant>(), init);
     return d;
 }
 
@@ -124,9 +113,7 @@ const IR::Node* DoConstantFolding::postorder(IR::Cmpl* e) {
     }
 
     mpz_class value = ~cst->value;
-    auto result = new IR::Constant(cst->srcInfo, t, value, cst->base, true);
-    setConstant(e, result);
-    return result;
+    return new IR::Constant(cst->srcInfo, t, value, cst->base, true);
 }
 
 const IR::Node* DoConstantFolding::postorder(IR::Neg* e) {
@@ -151,9 +138,7 @@ const IR::Node* DoConstantFolding::postorder(IR::Neg* e) {
     }
 
     mpz_class value = -cst->value;
-    auto result = new IR::Constant(cst->srcInfo, t, value, cst->base, true);
-    setConstant(e, result);
-    return result;
+    return new IR::Constant(cst->srcInfo, t, value, cst->base, true);
 }
 
 const IR::Constant*
@@ -260,9 +245,7 @@ DoConstantFolding::compare(const IR::Operation_Binary* e) {
             return e;
         }
         bool bresult = (left->value == right->value) == eqTest;
-        auto result = new IR::BoolLiteral(e->srcInfo, bresult);
-        setConstant(e, result);
-        return result;
+        return new IR::BoolLiteral(e->srcInfo, bresult);
     } else if (typesKnown) {
         auto le = EnumInstance::resolve(eleft, typeMap);
         auto re = EnumInstance::resolve(eright, typeMap);
@@ -270,9 +253,7 @@ DoConstantFolding::compare(const IR::Operation_Binary* e) {
             BUG_CHECK(le->type == re->type,
                       "%1%: different enum types in comparison", e);
             bool bresult = (le->name == re->name) == eqTest;
-            auto result = new IR::BoolLiteral(e->srcInfo, bresult);
-            setConstant(e, result);
-            return result;
+            return new IR::BoolLiteral(e->srcInfo, bresult);
         }
     }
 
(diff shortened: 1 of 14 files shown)


mbudiu-vmw:
I realized that there are more cases to consider.

H[2] h;
H[h.size +1] h1;
H[h1.size + 1] h2;
Etc.

Also:

control c()(bit<32> sz) { H[sz] h; apply {}}

frontends/common/constantFolding.h
```diff
@@ -28,11 +28,11 @@ namespace P4 {
 class DoConstantFolding : public Transform {
  protected:
     const ReferenceMap* refMap;  // if null no 'const' values can be resolved
-    TypeMap* typeMap;  // if null we have no types; updated for new constants
+    const TypeMap* typeMap;  // if null we have no types; updated for new constants
     bool typesKnown;
     bool warnings;  // if true emit warnings
     // maps expressions and declarations to their constant values
-    std::map<const IR::Node*, const IR::Expression*> constants;
+    std::map<const IR::Declaration_Constant*, const IR::Expression*> constants;
```
mbudiu-vmw: You have to update the comment above.

mbudiu-vmw: Also, please update the reference outputs. This can be done by running each failed test with the -f flag - after you made sure that the produced output is the intended one. For example:
`./p4/testdata/p4_16_samples/stack.p4.test -f` will update the output produced by this test.

cc10512:
you should rebase on master, not merge master.

jnfoster:
The GitHub web interface triggers these merges...

cc10512:
That's why I never trusted GUIs ...

sethfowler:
Yeah. I wish the Github web interface would let you "Update branch" using a rebase instead of a merge. It would save a lot of time.

jnfoster:
* Fixed comments.
* Updated test outputs, which covers bug 1 above.
* Added a new test case for bug 2 above.

frontends/common/constantFolding.cpp
```diff
@@ -23,50 +23,49 @@ namespace P4 {
...
     CHECK_NULL(expr);
-    auto cst = get(constants, expr);
-    if (cst != nullptr)
-        return cst;
-    if (expr->is<IR::Constant>())
+    if (expr->is<IR::Constant>()) {
         return expr;
-    if (expr->is<IR::BoolLiteral>())
+    }
+    if (expr->is<IR::BoolLiteral>()) {
```
ChrisDodd: Seems like this could be just `expr->is<IR::Literal>()` to get Constant, BoolLiteral and StringLiteral

frontends/common/constantFolding.cpp
```diff
@@ -23,50 +23,49 @@ namespace P4 {
...
         return expr;
+    }
     if (expr->is<IR::ListExpression>()) {
         auto list = expr->to<IR::ListExpression>();
         for (auto e : list->components)
-            if (getConstant(e) == nullptr)
+            if (getConstant(e) == nullptr) {
```
mbudiu-vmw: Our style guidelines do not require braces for this case.
Since @ChrisDodd prefers not to squander vertical space, I have avoided using them; they take 1 extra line.

frontends/common/constantFolding.cpp
```diff
@@ -76,27 +75,26 @@ const IR::Node* DoConstantFolding::postorder(IR::Declaration_Constant* d) {
...
+        // is not init, but (d->type)init. The typechecker inserts
+        // casts, but if we run this before typechecking we have to be
+        // more conservative.
         if (init->is<IR::Constant>()) {
             auto cst = init->to<IR::Constant>();
             if (d->type->is<IR::Type_Bits>()) {
                 if (cst->type->is<IR::Type_InfInt>() ||
                     (cst->type->is<IR::Type_Bits>() &&
-                     !(*d->type->to<IR::Type_Bits>() == *cst->type->to<IR::Type_Bits>())))
+                     !(*d->type->to<IR::Type_Bits>() == *cst->type->to<IR::Type_Bits>())))
```
mbudiu-vmw: You should set-up your editor to delete spaces at the end of lines.

frontends/common/constantFolding.cpp
```diff
@@ -76,27 +75,26 @@ const IR::Node* DoConstantFolding::postorder(IR::Declaration_Constant* d) {
...
             if (d->type->is<IR::Type_Bits>()) {
                 if (cst->type->is<IR::Type_InfInt>() ||
                     (cst->type->is<IR::Type_Bits>() &&
-                     !(*d->type->to<IR::Type_Bits>() == *cst->type->to<IR::Type_Bits>())))
+                     !(*d->type->to<IR::Type_Bits>() == *cst->type->to<IR::Type_Bits>()))) 
                     init = new IR::Constant(init->srcInfo, d->type, cst->value, cst->base);
-                setConstant(d, init);
             }
         }
+        if (init != d->initializer)
```
mbudiu-vmw: So here no braces?

frontends/common/constantFolding.h
```diff
@@ -23,16 +23,15 @@ limitations under the License.
...
 class DoConstantFolding : public Transform {
  protected:
     const ReferenceMap* refMap;  // if null no 'const' values can be resolved
-    TypeMap* typeMap;  // if null we have no types; updated for new constants
+    const TypeMap* typeMap;      // if null we have no types; updated for new constants
     bool typesKnown;
     bool warnings;  // if true emit warnings
-    // maps expressions and declarations to their constant values
-    std::map<const IR::Node*, const IR::Expression*> constants;
+    // maps declaration constants to constant values
```
mbudiu-vmw: This comment could have three slashes

testdata/p4_16_samples_outputs/stack2.p4
```diff
@@ -0,0 +1,16 @@
+#include <core.p4>
```
mbudiu-vmw: How about some of the tests I suggested in comments, e.g.
```
H[h1.size+1] h2
```

jnfoster: In principle, this should be possible, but it would be a deeper change based on how the passes are currently organized:

* We don't fold IR::Members until after type checking.

* In particular, we need to know the type of the expression to create a constant with the appropriate width.

* But the type checker requires header stacks to have a compile-time known constant as their size.

mbudiu-vmw: This should be either a positive or a negative test. If it is negative it should have a good error message.

frontends/common/constantFolding.cpp
jnfoster: String literals are not currently folded, so this would change the logic of the function. Leaving it as-is for now.

testdata/p4_16_samples_outputs/stack2.p4
jnfoster: Added stack3.p4 as an error test.

jnfoster:
Okay, I think this is finally ready to go. Please let me know if you have more feedback.

frontends/common/constantFolding.cpp
```diff
@@ -35,38 +32,34 @@ const IR::Expression* DoConstantFolding::getConstant(const IR::Expression* expr)
...
         return e;
-    auto v = get(constants, decl->getNode());
-    if (v == nullptr)
-        return e;
-    setConstant(e, v);
-    if (v->is<IR::ListExpression>())
-        return e;
-    return v;
+    if (decl->is<IR::Declaration_Constant>()) {
+        auto dc = decl->to<IR::Declaration_Constant>();
```
ChrisDodd: Stylisticly I prefer combining these two lines as

    if (auto dc = decl->to<IR::Declaration_Constant>()) {

frontends/common/constantFolding.cpp
```diff
@@ -76,27 +69,26 @@ const IR::Node* DoConstantFolding::postorder(IR::Declaration_Constant* d) {
...
             }
         }
+        if (init != d->initializer)
+            d = new IR::Declaration_Constant(d->srcInfo, d->name, d->annotations, d->type, init);
     }
-    if (init != d->initializer)
-        d = new IR::Declaration_Constant(d->srcInfo, d->name, d->annotations, d->type, init);
+    auto orig = getOriginal()->to<IR::Declaration_Constant>();
+    BUG_CHECK(orig, "getOriginal() did not return a Declaration_Constant");
+    constants.emplace(orig, init);
```
ChrisDodd: These 3 lines could be replaced by

    constants.emplace(getOriginal<IR::Declaration_Constant>(), init);

-- the templated version of `getOriginal` does the cast and the BUG_CHECK.

testdata/p4_16_errors/stack3.p4
```diff
@@ -0,0 +1,42 @@
+/*
```
mbudiu-vmw: Where is the output generated by the compiler for this test?
There should be a file stack3.p4-stderr at the very least.

jnfoster:
My fault, I didn't understand the protocol for adding tests that fail to
compile. I'll grab the stderr and stash it in the repo now.

-N

mbudiu-vmw:
In general you should git add everything in testdata after running a test.

jnfoster:
Got it. Thanks!

antoninbas:
@jnfoster I doubt you meant to update the PI submodule ref pointer (you probably shouldn't)

jnfoster:
Gah. I think this is from pushing the update branch button here and then pushing additional changes from the command line. Let me fix this...

jnfoster:
Okay @sethfowler helped me detangle myself. 

Morale: don't ever push GitHub's "Update branch" button or follow their command-line instructions!

frontends/common/constantFolding.cpp
jnfoster: Done.

frontends/common/constantFolding.cpp
jnfoster: Cool. Done.

testdata/p4_16_errors/stack3.p4
jnfoster: Added.

testdata/p4_16_errors_outputs/stack3.p4-stderr
```diff
@@ -0,0 +1,36 @@
...
+../testdata/p4_16_errors/stack3.p4(35): error: header h[]: Size of header stack type should be a constant
+        h[x] s3;
+        ^^^^
+../testdata/p4_16_errors/stack3.p4(35): error: header h[]: Size of header stack type should be a constant
+        h[x] s3;
+        ^^^^
+../testdata/p4_16_errors/stack3.p4(35): error: header h[]: Size of header stack type should be a constant
+        h[x] s3;
+        ^^^^
+../testdata/p4_16_errors/stack3.p4(35): error: Could not find type of h[]
```
jnfoster: @mbudiu-vmw I'm confused about why the type of `h` can't be found. Is this indicating a deeper error?

mbudiu-vmw: No, if a sub-expression fails type-checking no type is installed in the typeMap, so all the expressions that involve it as a sub-expression will fail to type-check too. These additional error messages are somewhat redundant, but other compilers are equally confused.

mbudiu-vmw:
@jnfoster : this code does not build.

frontends/common/constantFolding.cpp
jnfoster: Actually, this doesn't compile. So I reverted.

jnfoster:
Sorry about that. Clearly I had not run `make check` when I committed @ChrisDodd's suggested changes in. I just pushed a new version that walks it back and verified that it compiles and passes all tests on my machine. Should be fixed now.

frontends/common/constantFolding.cpp
mbudiu-vmw: You forgot the `=` sign in your code. It would compile.

jnfoster: Mea culpa! I'll push one more time.

PR #542, opened 2017-04-28
P4_14 register related fixes

ChrisDodd:
- support for direct registers
- support for registers with layout instead of width


ChrisDodd:
Unfortunately only in our internal tests -- we have no tests that use registers that actually run on bmv2, though I hope to improve that in the future

PR #544, opened 2017-05-01
BMV2 multi-architecture support

hanw:
[do NOT merge yet]
  Bmv2 multi-architecture support
    
    This commit refactors bmv2 backend to support multiple pipeline
    architectures. JsonConverter is refactored into a pass manager
    to conform to the rest of the compiler design.
    
    Added files:
    backend.cpp       - the new backend pass manager
    convert*.cpp      - compiler passes for control, parser, deparser,
                        externs and headers
    inferArchitecture - a new frontend pass to parse the architecture p4
                        into a data object which is then used by json
                        generator
    model.cpp         - a v2model that oriented for p4-16 IR


frontends/common/model.h
```diff
@@ -62,4 +65,103 @@ class Model {
...
+
+/// Method_Model : Block_Model<Param_Model> : Type_Model
+struct Method_Model : public Block_Model<Param_Model> {
+    ::Model::Type_Model type;
+    explicit Method_Model(cstring name) :
+        Block_Model<Param_Model>(name), type("Method") {}
+};
+
+/// Extern_Model : Block_Model<Method_Model> : Type_Model
+struct Extern_Model : public Block_Model<Method_Model> {
```
cc10512: How do you distinguish between decls and methods in an extern? Can externs have only Methods?

frontends/common/model.h
```diff
@@ -62,4 +65,103 @@ class Model {
...
+};
+
+/// V2Model : Model::Model
+class V2Model : public ::Model::Model {
+ public:
+    std::vector<Parser_Model*>  parsers;
+    std::vector<Control_Model*> controls;
+    std::vector<Extern_Model*>  externs;
+    std::vector<Type_Model*>    match_kinds;
+    ::Model::Elem                        action_profile;
```
cc10512: Not externs?

frontends/Makefile.am
```diff
@@ -114,7 +116,8 @@ common_frontend_UNIFIED = \
 	frontends/common/resolveReferences/referenceMap.cpp \
 	frontends/common/resolveReferences/resolveReferences.cpp \
 	frontends/common/parseInput.cpp \
-	frontends/common/constantParsing.cpp
+	frontends/common/constantParsing.cpp \
```
cc10512: we should really keep these files in alphabetical order to reduce merge conflicts.

backends/bmv2/tov1model.cpp
```diff
@@ -0,0 +1,30 @@
...
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+#include "frontends/common/model.h"
+
+/// take v2model and turn it into v1model
```
cc10512: Do we really need to convert the model? I would have thought that we can serialize v2 to the json from the v1model.

backends/bmv2/backend.h
```diff
@@ -0,0 +1,92 @@
...
+#include "helpers.h"
+#include "expressionConverter.h"
+#include "frontends/p4/coreLibrary.h"
+#include "midend/convertEnums.h"
+#include "frontends/common/model.h"
+#include "frontends/p4/fromv1.0/v1model.h"
+
+namespace BMV2 {
+
+class DirectMeterMap final {
```
ChrisDodd: That does seem odd -- this class seems to be tracking the (color) output, but that should also be needed for meters and registers, which also have destination fields that are written by the extern.

hanw:
Close temporarily for now to reducing spam to people. Will reopen later.

backends/bmv2/backend.h
hanw: Yeah, this is odd. This is the last piece that needs to be taken out of jsonconverter.cpp (which is now backend.cpp). Will separate it to its own file.

PR #548, opened 2017-05-01
Document the RemoveUnusedDeclarations pass.

cole-barefoot:


frontends/p4/unusedDeclarations.h
```diff
@@ -70,14 +94,22 @@ class RemoveUnusedDeclarations : public Transform {
...
+ * If @warn is true, emit compiler warnings if an unused instance of an
+ * IR::P4Table or IR::Declaration_Instance is removed.
+ */
 class RemoveAllUnusedDeclarations : public PassManager {
  public:
     explicit RemoveAllUnusedDeclarations(ReferenceMap* refMap, bool warn = false) {
         CHECK_NULL(refMap);
         std::set<const IR::Node*> *warned = nullptr;
+
+        // TODO: The @warned set is discarded.  Is it necessary, or can it be replaced
```
mbudiu-vmw: The set is not discarded, it is used as an argument for the RemoveUnusedDeclarations below.
It is the same set for all iterations of the PassRepeated, thus it cannot be replaced with a bool.

cole-barefoot: Oh I see.  This avoids issuing multiple warnings for the same unreferenced extern definition, which is the only thing that can both trigger a warning and persist across multiple rounds.

I'll remove the TODO and add a comment to this effect.  Thanks for pointing it out.

frontends/p4/unusedDeclarations.h
```diff
@@ -22,10 +22,38 @@ limitations under the License.
...
+ * The following kinds of nodes are not removed even if they are unreferenced:
+ *  - IR::Declaration_MatchKind
+ *  - IR::Parameter
+ *  - IR::Type_Error
+ *  - IR::Type_Extern
+ *  - IR::Type_Method
+ *  - IR::Type_StructLike
+ *
+ * If @warned is non-null, unused IR::P4Table and IR::Declaration_Instance
+ * nodes are stored in @warned if they are unused and removed by this pass.
```
ChrisDodd: In addition, a warning message is emitted only if the declaration was not previously in `warned` -- that is in fact the main point of `warned` -- to suppress duplicate warnings

cole-barefoot:
Thanks @mbudiu-vmw and @ChrisDodd.  I updated the comments to (a) remove the TODO, which was wrong, and (b) clarify the behavior of the warning list.

PR #554, opened 2017-05-02
Use C++ reentrant parsers and lexers for P4-14 and P4-16

sethfowler:
This PR moves our P4-14 and P4-16 parsers and lexers to use Flex and Bison's C++ backends.

All of the commits have long-form commit messages; see those for the details.


frontends/parsers/parserDriver.cpp
```diff
@@ -0,0 +1,238 @@
...
+
+#include <boost/iostreams/device/file_descriptor.hpp>
+#include <boost/iostreams/stream.hpp>
+
+namespace {
+
+/// A RAII helper class that provides an istream wrapper for a stdio FILE*. This
+/// is the efficient implementation for users with boost::iostreams installed.
+struct AutoStdioInputStream {
+    explicit AutoStdioInputStream(FILE* in)
```
mbudiu-vmw: I bet @ChrisDodd would write this using fewer lines.

frontends/parsers/parserDriver.cpp
```diff
@@ -0,0 +1,238 @@
...
+    boost::iostreams::stream_buffer<boost::iostreams::file_descriptor_source> buffer;
+    std::istream stream;
+};
+
+}  // anonymous namespace
+
+#else
+
+/// A RAII helper class that provides an istream wrapper for a stdio FILE*. This
+/// is an inefficient fallback implementation.
```
mbudiu-vmw: The buffering of file I/O is the least of our performance worries at this point.

frontends/parsers/parserDriver.cpp
```diff
@@ -0,0 +1,238 @@
...
+}
+
+void AbstractParserDriver::onReadIdentifier(cstring id) {
+    lastIdentifier = id;
+}
+
+void AbstractParserDriver::onParseError(const Util::SourceInfo& location,
+                                        const std::string& message) {
+    static const std::string unexpectedIdentifierError =
+        "syntax error, unexpected IDENTIFIER";
```
mbudiu-vmw: bison's messages are terrible, so this is a nice improvement.

frontends/parsers/parserDriver.cpp
```diff
@@ -0,0 +1,238 @@
...
+
+/* static */ void P4ParserDriver::checkShift(const Util::SourceInfo& l,
+                                                 const Util::SourceInfo& r) {
+    if (!l.isValid() || !r.isValid())
+        BUG("Source position not available!");
+    const Util::SourcePosition& f = l.getStart();
+    const Util::SourcePosition& s = r.getStart();
+    if (f.getLineNumber() != s.getLineNumber() ||
+        f.getColumnNumber() != s.getColumnNumber() - 1)
+        ::error("Syntax error at shift operator: %1%", l);
```
mbudiu-vmw: I wonder whether this is a good error; it's not really a shift since there are spaces in the middle.
Anyway, I suspect this message is inherited from the old code.

frontends/common/constantParsing.h
```diff
@@ -17,11 +17,71 @@ limitations under the License.
...
+ * lexing. The parser is responsible for actually interpreting the raw text as a
+ * numeric value and transforming it into an IR::Constant using parseConstant().
+ *
+ * To illustrate how a numeric constant is represented using this struct,
+ * here is a breakdown of '16w0x12':
+ *
+ *          ___
+ *         /                                    ___
+ *         |                                   /
+ *         |   bitwidth (if @hasWidth)        |       16
```
sethfowler: It would've taken a lot fewer if I turned this diagram "sideways", but then I didn't have enough horizontal space. =\

frontends/common/constantParsing.h
```diff
@@ -17,11 +17,71 @@ limitations under the License.
...
+    unsigned base;  /// The base in which the constant is expressed.
+    bool hasWidth;  /// If true, a bitwidth and separator are present.
+};
+
+std::ostream& operator<<(std::ostream& out, const UnparsedConstant& constant);
+
+/**
+ * Parses an UnparsedConstant @constant into an IR::Constant object, with
+ * location information taken from @srcInfo. If parsing fails, by default null
+ * is returned, but callers can optionally specify a @defaultValue to return
```
sethfowler: It's actually *always* used. It might be better to make it non-optional.

frontends/common/parseInput.cpp
```diff
@@ -65,3 +74,53 @@ const IR::P4Program* parseP4File(CompilerOptions& options) {
     }
     return result;
 }
+
+const IR::P4Program* parseP4String(const std::string& input,
```
sethfowler: Yep. I tried merging them at one point, but wasn't happy with the result - they're quite similar, but the differences are in awkward places. I don't mind trying again, though; it might be that a second attempt would go better.

frontends/parsers/p4/p4lexer.ll
```diff
@@ -0,0 +1,199 @@
+%{
```
sethfowler: Even if it did, the formatting changes would unfortunately make it pretty hard to review. =\

frontends/parsers/parserDriver.cpp
sethfowler: This implementation copies the entire file multiple times, but yeah, it's still not really a limiting factor. I'd rather use only the boost::iostreams implementation, though. Waste not, want not.

frontends/parsers/parserDriver.cpp
sethfowler: It is, yeah. I could tweak it, though; maybe "syntax error, right shift operator contains whitespace" would be clearer and more consistent with the other Bison errors.

frontends/parsers/parserDriver.cpp
sethfowler: This is actually present in the old code too, although the implementation is now different.

sethfowler:
Thanks for the super fast review, Mihai! I'll make some changes based on your comments.

ChrisDodd:
I must admit, I don't really see the point of this change.  We don't actually need a reentrant parser.  You've added about 400 lines of comments (good) and 600 lines of obfuscation (bad) -- more code to read is never a good thing.  It doesn't help that the C++ interface of bison is particularly poorly designed.

frontends/parsers/p4/p4parser.ypp
```diff
@@ -0,0 +1,1012 @@
...
+%token      MUL "*"
+%token      DIV "/"
+%token      MOD "%"
+%token      BIT_OR "|"
+%token      BIT_AND "&"
+%token      BIT_XOR "^"
+%token      COMPLEMENT "~"
+%token      L_BRACKET "["
+%token      R_BRACKET "]"
+%token      L_BRACE "{"
```
ChrisDodd: Using symbolic constants for 1-character tokens in a grammar is THE WORST kind of gratuitous obfuscation and should NEVER be done.  It make the code harder to read, harder to understand, harder to debug, and more verbose -- for absolutely zero benefit.

Its bad enough that bison forces you to use symbol constants for 2-character tokens.

frontends/parsers/p4/p4lexer.ll
```diff
@@ -0,0 +1,199 @@
...
+<STRING>\"      { BEGIN(INITIAL);
+                  auto string = cstring(driver.stringLiteral);
+                  return Parser::make_STRING_LITERAL(string, driver.yylloc); }
+<STRING>.       { driver.stringLiteral += yytext; }
+<STRING>\n      { driver.stringLiteral += yytext; }
+
+"abstract"      { BEGIN(NORMAL); return Parser::make_ABSTRACT(driver.yylloc); }
+"action"        { BEGIN(NORMAL); return Parser::make_ACTION(driver.yylloc); }
+"actions"       { BEGIN(NORMAL); return Parser::make_ACTIONS(driver.yylloc); }
+"apply"         { BEGIN(NORMAL); return Parser::make_APPLY(driver.yylloc); }
```
ChrisDodd: One of the changes I've wanted to do was to unify the P4_14 and P4_16 lexers -- but this requires getting rid of the explicit keyword patterns and instead putting all the keywords in a table with their corresponding bison tokens.  The ID rule would then use that table to lookup keywords prior to the symbolkind lookup it already does, and we would just have two static tables (one for P4_14 and one for P4_16).

Unfortunately, it looks like this change would make that very difficult, as the tokens are no longer just enums.

frontends/parsers/p4/p4parser.ypp
sethfowler: It's actually possible to use the aliases that are defined here in the grammar; for example, we could refer to L_BRACE as "{". I waffled on whether to do that, and I ultimately went this direction because it keeps things consistent between the lexer and the parser. I'm happy to change it, though, since it sounds like you'd prefer to do it the other way.

frontends/parsers/p4/p4lexer.ll
sethfowler: We should still be able to do that kind of table lookup. Bison's C++ backend still generates an enumeration containing each token type (`tokentype`), and we can still retrieve it from the symbol via `symbol_type::type_get()`.

sethfowler:
Beyond using symbol names for tokens, is it the ParserDriver code you consider obfuscation, Chris? Those files are about 50% comments; CLOC reports only 245 lines of actual obfuscation. =)

frontends/parsers/parserDriver.cpp
ChrisDodd: The only reason we are using `FILE` at all is because iostreams don't support the equivalent of popen.  If boost has that, we could use it thoughout and not have to do any of this adaptor stuff.

sethfowler:
OK, I've updated the grammars to use literals instead of symbolic names for operator tokens. (We could do this for keywords too, if we wanted, but it'd actually make things more verbose.)

frontends/parsers/parserDriver.cpp
sethfowler: I'm not aware of an iostreams-based alternative to popen, unfortunately, though that would be ideal. My plan in the long term was to eliminate the need for `FILE` support by implementing the C preprocessor features we need internally.

frontends/common/parseInput.cpp
```diff
@@ -15,13 +15,50 @@ limitations under the License.
...
-#include "frontends/p4/p4-parse.h"
+#include "lib/error.h"
+#include "lib/source_file.h"
+
+namespace P4 {
+
+template <typename Input>
+static const IR::P4Program*
+parseV1Program(const char* name, Input& stream,
+               boost::optional<DebugHook> debugHook = boost::none) {
```
sethfowler: That's a good point. I'll make that change.

frontends/parsers/v1/v1parser.ypp
```diff
@@ -0,0 +1,1049 @@
+/*
```
sethfowler: It's buried deep in the reams of text I wrote in the commit messages, but P4_14 unfortunately won't work, because automake insists on actually *rewriting the include guards in the code Bison generates*, and it does it with a regular expression that collapses all non-alphabetic characters, which means that `P4_14` and `P4` (and `P4_16` for that matter) become indistinguishable. This was actually the problem that led me to move these files!

We can definitely rename them to something else. Anything that contains any letter other than a single `P` would work. =)

PR #555, opened 2017-05-02
adds an unimplemented exception and exception catching in drivers

cc10512:
Adds a new type of P4Exception: CompilerUnimplemented to replace
instances where we use BUG/BUG_CHECK as a placeholder for features
that we have not yet implemented.

Also adds exception catching in the compiler drivers, such that the
compiler exits cleanly.

### DIFF
--- a/backends/bmv2/bmv2.cpp
+++ b/backends/bmv2/bmv2.cpp
@@ -49,22 +49,38 @@ int main(int argc, char *const argv[]) {
     auto program = parseP4File(options);
     if (program == nullptr || ::errorCount() > 0)
         return 1;
-    P4::FrontEnd frontend;
-    frontend.addDebugHook(hook);
-    program = frontend.run(options, program);
+    try {
+        P4::FrontEnd frontend;
+        frontend.addDebugHook(hook);
+        program = frontend.run(options, program);
+    } catch (const Util::P4CExceptionBase &bug) {
+        std::cerr << bug.what() << std::endl;
+        return 1;
+    }
     if (program == nullptr || ::errorCount() > 0)
         return 1;
 
+    const IR::ToplevelBlock* toplevel = nullptr;
     BMV2::MidEnd midEnd(options);
     midEnd.addDebugHook(hook);
-    auto toplevel = midEnd.process(program);
-    if (options.dumpJsonFile)
-        JSONGenerator(*openFile(options.dumpJsonFile, true)) << program << std::endl;
+    try {
+        toplevel = midEnd.process(program);
+        if (options.dumpJsonFile)
+            JSONGenerator(*openFile(options.dumpJsonFile, true)) << program << std::endl;
+    } catch (const Util::P4CExceptionBase &bug) {
+        std::cerr << bug.what() << std::endl;
+        return 1;
+    }
     if (::errorCount() > 0 || toplevel == nullptr)
         return 1;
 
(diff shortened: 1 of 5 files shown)


lib/exceptions.h
```diff
@@ -48,13 +48,31 @@ class CompilerBug final : public P4CExceptionBase {
     template <typename... T>
     CompilerBug(const char* format, T... args)
             : P4CExceptionBase(format, args...)
-    { message = "COMPILER BUG:\n" + message; }
+    // \e[31m prints the text in red
+    { message = "\e[31mCOMPILER BUG\e[0m:\n" + message; }
```
jnfoster: No "\e[5m"? Kidding. Should we remove the all-caps?

cc10512: Not sure "\e5m" will convey the right message: can it blink slow enough to wait for someone to fix the bug? :)

Yeah, camel case will make it less threatening.

backends/bmv2/bmv2.cpp
```diff
@@ -49,22 +49,38 @@ int main(int argc, char *const argv[]) {
...
     if (program == nullptr || ::errorCount() > 0)
         return 1;
-    P4::FrontEnd frontend;
-    frontend.addDebugHook(hook);
-    program = frontend.run(options, program);
+    try {
+        P4::FrontEnd frontend;
+        frontend.addDebugHook(hook);
+        program = frontend.run(options, program);
+    } catch (const Util::P4CExceptionBase &bug) {
```
mbudiu-vmw: There is a benefit for not catching exceptions: makes debugging easier.
Otherwise you have to set breakpoints manually in all these places.

lib/exceptions.h
mbudiu-vmw: Is this portable, or will it look terrible in non-xterm terminals?
I usually have my shell in emacs.

lib/exceptions.h
```diff
@@ -48,13 +48,31 @@ class CompilerBug final : public P4CExceptionBase {
...
     template <typename... T>
     CompilerBug(const char* file, int line, const char* format, T... args)
             : P4CExceptionBase(format, args...)
-    { message = cstring("COMPILER BUG: ") + file + ":" + Util::toString(line) + "\n" + message; }
+    { message = cstring("In file: ") + file + ":" + Util::toString(line) + "\n" +
+                cstring("\e[31mCOMPILER BUG\e[0m: ") + message; }
 };
 
+// This class indicates an unimplemented feature in the compiler
+class CompilerUnimplemented final : public P4CExceptionBase {
```
mbudiu-vmw: You probably want a JavaDoc comment.
Also, this is not yet used anywhere, perhaps you want at least one use case.

backends/bmv2/bmv2.cpp
sethfowler: In theory for debugging purposes you could use `catch throw` in gdb. I haven't ever actually tried that, though, so I'm not sure how well it works in practice.

lib/exceptions.h
cc10512: We will start using it during code reviews. In particular, we are working on the bmv2 backend and there are a few places there. Coming ...

cc10512: It is definitely portable. Works even on mobile phones. Emacs shell as well: 
screen shot 2017-05-02 at 17 53 44

PR #556, opened 2017-05-03
Add/convert Doxygen comments for SideEffects and DoSimplifyExpressions visitors.

cole-barefoot:
@mbudiu-vmw: I added several TODO comments with questions that you may know the answer to.  If you think they are useful pointers to things we should address in the future, we can leave them in---if not, I'm happy to remove them.

### DIFF
--- a/frontends/p4/sideEffects.cpp
+++ b/frontends/p4/sideEffects.cpp
@@ -31,9 +31,10 @@ namespace {
 struct EvaluationOrder {
     ReferenceMap* refMap;
 
-    /// Output: The original expression with side-effecting sub-expressions
-    /// replaced by temporaries.
-    const IR::Expression* final;  // output
+    /// Output: An expression whose evaluation will produce the same result as
+    /// the original one, but where side-effects have been factored out such
+    /// that each statement produces at most one side-effect.
+    const IR::Expression* final;
 
     // Declaration instead of Declaration_Variable so it can be more easily inserted
     // in the program IR.
@@ -72,8 +73,8 @@ struct EvaluationOrder {
     }
 };
 
-/** @brief Replace expressions containing constructor or method invocations
- * with temporaries.
+/** @brief Replaces expressions containing constructor or method invocations
+ * with temporaries.  Also unrolls short-circuit evaluation.
  * 
  * The EvaluationOrder field accumulates lists of the temporaries introduced in
  * this way, as well as assignment statements that assign the original
@@ -89,9 +90,6 @@ struct EvaluationOrder {
  * left-value-ness and compile-time-contant-ness.
  */
 class DismantleExpression : public Transform {
-    // TODO: It may be worth refactoring DismantleExpression to purely use
-    // pre-/postorder methods, rather than invoking visit() in preorder
-    // methods, to reflect the design pattern used elsewhere.
     ReferenceMap* refMap;
     TypeMap* typeMap;
     EvaluationOrder *result;
@@ -203,7 +201,6 @@ class DismantleExpression : public Transform {
         visit(expression->expr);
         auto left = result->final;
         CHECK_NULL(left);
-        // TODO: why clone here?
         auto clone = expression->clone();
         clone->expr = left;
         typeMap->setType(clone, type);
@@ -223,7 +220,6 @@ class DismantleExpression : public Transform {
             CHECK_NULL(left);
             visit(expression->right);
             auto right = result->final;
-            // TODO: why clone here?
             auto clone = expression->clone();
             clone->left = left;
             clone->right = right;
@@ -273,7 +269,6 @@ class DismantleExpression : public Transform {
             auto block = new IR::BlockStatement(*ifFalse);
             auto ifStatement = new IR::IfStatement(expression->srcInfo, cond, ifTrue, block);
             result->statements->push_back(ifStatement);
-            // TODO: result->addAssignment() returns a clone; why clone again?
             result->final = path->clone();
         }
         typeMap->setType(result->final, type);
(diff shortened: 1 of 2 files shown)


frontends/p4/sideEffects.cpp
```diff
@@ -22,14 +22,19 @@ namespace P4 {
...
+/** Data structure used for making the order of evaluation explicit for
+ * sub-expressions in an expression.  An expression ```e``` will be represented
+ * as a sequence of temporary declarations, followed by a sequence of
+ * statements (mostly assignments to the temporaries, but which could also
+ * include conditionals for short-circuit evaluation).
+ */
 struct EvaluationOrder {
     ReferenceMap* refMap;
+
+    /// Output: The original expression with side-effecting sub-expressions
```
mbudiu-vmw: An expression whose evaluation will produce the same result as the original one, but where side-effects have been factored out such that each statement produces at most one side-effect.

frontends/p4/sideEffects.cpp
```diff
@@ -61,13 +72,36 @@ struct EvaluationOrder {
     }
 };
 
+/** @brief Replace expressions containing constructor or method invocations
```
mbudiu-vmw: This also deals with short-circuit evaluation.

frontends/p4/sideEffects.cpp
```diff
@@ -61,13 +72,36 @@ struct EvaluationOrder {
...
+ * assignments above the expression.
+ *
+ * @pre An up-to-date ReferenceMap and TypeMap.
+ *
+ * @post The RefenceMap and TypeMap are updated to reflect changes to the IR.
+ * This includes extra information stored in the TypeMap, specifically
+ * left-value-ness and compile-time-contant-ness.
+ */
 class DismantleExpression : public Transform {
+    // TODO: It may be worth refactoring DismantleExpression to purely use
```
mbudiu-vmw: We actually want to precisely control the visit order, to reflect the order that the evaluation should be performed. I don't want to rely on the visitor order, which is not well-defined. This is the only way to really control the precise order: manual calls to visit and prune. See the compiler-design document too.

frontends/p4/sideEffects.cpp
```diff
@@ -169,6 +203,7 @@ class DismantleExpression : public Transform {
         visit(expression->expr);
         auto left = result->final;
         CHECK_NULL(left);
+        // TODO: why clone here?
```
mbudiu-vmw: If you don't clone you have all sorts of weird problems (which are also a big pain to debug, BTW). For example, consider def-use analysis: the same expression has different uses and definitions depending of where it occurs in the program. So although something is syntactically identical, it should be represented by a different IR node. You will want different attributes computed for the two clones.

frontends/p4/sideEffects.cpp
```diff
@@ -237,6 +273,7 @@ class DismantleExpression : public Transform {
             auto block = new IR::BlockStatement(*ifFalse);
             auto ifStatement = new IR::IfStatement(expression->srcInfo, cond, ifTrue, block);
             result->statements->push_back(ifStatement);
+            // TODO: result->addAssignment() returns a clone; why clone again?
```
mbudiu-vmw: If the path is used someplace else, one of the two users has to clone it.

mbudiu-vmw: Also, better safe than sorry. Unless the contract for addAssignment is clear about this, I don't want to rely on it.

frontends/p4/sideEffects.cpp
```diff
@@ -279,8 +317,11 @@ class DismantleExpression : public Transform {
 
     // We don't want to compute the full read/write set here so we
     // overapproximate it as follows: all declarations that occur in
-    // an expression.  TODO: this could be made more precise, perhaps
-    // using LocationSets.
+    // an expression.
+
+    // TODO: this could be made more precise, perhaps using LocationSets.
+    // TODO: surely there are other passes that compute reads/writes?  Perhaps
```
mbudiu-vmw: Unfortunately the pass that computes reads/writes depends on this one being run first. The reason is that you cannot compute reads if you don't know what side-effects occur first...

frontends/p4/sideEffects.cpp
```diff
@@ -295,6 +336,8 @@ class DismantleExpression : public Transform {
         }
     };
 
+    // TODO: alias analysis would be generally useful... refactor this into
```
mbudiu-vmw: Alias analysis would use def-use information, which needs the information computed by this pass.

frontends/p4/sideEffects.h
```diff
@@ -26,15 +26,23 @@ limitations under the License.
 
 namespace P4 {
 
-// Determines whether an expression may have side-effects
-// (but not any sub-expression)
+/** @brief Determines whether an expression may have method or constructor
+ * invocations.
+ *
+ * The TypeMap and ReferenceMap arguments may be null, in which case every
+ * method call expression is counted.  With type information, invocations
```
mbudiu-vmw: "is assumed to have side-effects". Otherwise said, the analysis may be more conservative.

frontends/p4/sideEffects.h
```diff
@@ -75,39 +87,41 @@ class SideEffects : public Inspector {
...
+a[tmp1].x = tmp4;        // assign result of call of f to actual left value
+```
+ *
+ * Translating function calls proceeds as follows:
+ * - arguments are evaluated in order
+ * - inout and out arguments produce left-values---these are "saved"
+ * - function is called with temporaries for all arguments
+ * - out and inout temporaries are copied to the saved left-values in order
+ *
+ * An assignment statement ```e = e1;``` is treated as if it is a call to a
```
mbudiu-vmw: I was too pedantic, this just means that the LHS is evaluated first.

For such a simple language, P4 has a really nasty semantics due to side-effects and out parameter. This pass was one of the hardest to write. After it has been run everything is much simpler.

cole-barefoot:
Thanks @mbudiu-vmw!  I revised based on your suggestions and removed the TODOs.  I also added a comment explaining why we implement a conservative alias analysis as part of this pass.

PR #558, opened 2017-05-03
Support for 'bool' type in P4_14

ChrisDodd:
Trivial addition that was missing in the P4_14 frontend.


ChrisDodd:
The typechecker already has support for boolean types as all comparisons
are booleans.  Legacy P4_14 programs tend to use bit<1> as an alias for
boolean, so the P4_14 typechecker allows implicit conversions both ways.

Issue #562, opened 2017-05-03
Support nested structs in bmv2 compiler backend


antoninbas:
I'm attaching a sample program.
[nested_struct.p4.txt](https://github.com/p4lang/p4c/files/974694/nested_struct.p4.txt)

jafingerhut:
Not sure if this is a duplicate of https://github.com/p4lang/p4c/issues/455
Whether it is or not, please keep this issue in mind, too: https://github.com/p4lang/p4c/issues/486

jafingerhut:
If by "nested header" you mean enabling one header to be nested inside of another header, I would recommend _not_ allowing that, unless you have proposed behavior for that (e.g. what would it mean to do setValid on an inner header, when the outer header was invalid?)

There are no such corner cases that I am aware of with allowing structs inside of headers.

hesingh:
Yes.  Ah, sorry, I forgot about the setValid() behavior in nested headers! So, I agree with you to not allow nested headers.    So the trimmed down list is:

1. nested bit-vector structs.
2. Struct in #1 to be part of a p4 header?

hesingh:
Code review for, https://github.com/p4lang/p4c/pull/971/files, said to remove nested structs altogether in a new midend pass.   However, the PR has changed the frontend to support nested structs.   So, the frontend generates control-plane API with control, parser, action, method, table key with nested structs, but, subsequently, the new midend pass flattens the nested structs in API.   

Thus, a frontend pass is also needed to flatten nested structs before the control-plane API is generated.

hesingh:
@ Mihai  I could use some help with the frontend pass which flattens nested structs altogether and also changes any Casts.  Thus, the pass is used when the frontend has structs/types, Casts, and table keys available.    Where does the new pass get inserted in the list of frontend passes?  I have included a short list below to show where I think the new pass goes in.  Please confirm.  

      new PrettyPrint(options),
        // Simple checks on parsed program
        new ValidateParsedProgram(),
        // Synthesize some built-in constructs
        new CreateBuiltins(),
        new ResolveReferences(&refMap, true),  // check shadowing
        // First pass of constant folding, before types are known --
        // may be needed to compute types.
        new ConstantFolding(&refMap, nullptr),
        // Desugars direct parser and control applications
        // into instantiations followed by application
        new InstantiateDirectCalls(&refMap),
                                  NEW PASS                         <=== here?
        // Type checking and type inference.  Also inserts
        // explicit casts where implicit casts exist.
[...]

PR #564, opened 2017-05-03
Start a list of unsupported P4_16 fatures in bmv2 backend README

antoninbas:
See issue #563 
It would be nice if this list could be kept up-to-date as we implement more features in the bmv2 compiler backend or discover more issues. Feel free to update this list if you are aware of more issues.


backends/bmv2/README.md
```diff
@@ -17,3 +17,30 @@ To run and test this back-end you need some additional tools:
...
+  bit<8> f1;
+  bit<8> f2;
+};
+
+struct s1_t {
+  s1_t s01;
+  s1_t s02;
+};
+```
+
```
hanw: I am assuming this is about nested struct used in block constructor parameters? There is already a pass to eliminate nested struct in local variables.

sethfowler: Yep - the BMV2 backend fails if you try to define a metadata or header struct that contains other nested structs. (Or at least, it did last time I tried.)

antoninbas: @hanw This is related to https://github.com/p4lang/p4c/issues/562
You may be right, I think this only happens for structs used in block constructor parameters. I can update my PR to be more precise. Any idea why this case is handled differently?

mbudiu-vmw:
You should check this in so I can edit it.
Also, having the metadata contain nested structs does not work.

backends/bmv2/README.md
mbudiu-vmw: Headers cannot contain nested structs in P4, so only metadata is a problem.
There are no extern blocks in v1model which require structs as parameters, so this is not really an issue with v1model.p4.
There are many bugs filed with are related to unsupported features, including name collisions, and the error type.

sethfowler: Headers can't contain nested structs, but it seems to me that we could support nested structs in the "header type" that we instantiate the package with. For `v1model.p4`, that's the `H` type parameter to `package V1Switch<H,M>`. I'd expect to be able to do something like this:

```p4
header h1 {...}
struct s1 { h1 h; }
header h2 {...}
struct s2 {h2 h; }

// This type contains nested structs:
struct Headers { s1 nested1; s2 nested2; }

// This should work, but last I checked, it doesn't:
parser p(packet_in b, out Headers h, inout Meta m, inout standard_metadata_t sm) { ... }
```

hanw: One possible solution is to flatten the struct to only contain either bit<n> or header type, and then rewrite the parameter list of a block to use the new struct, and update all references to the struct field in the block, table, action, etc, to the new struct field. @mbudiu-vmw Do you other suggestions?

mbudiu-vmw: What kind of block are we talking about?
We cannot change APIs for externs of objects visible from the control-plane.

For controls and parsers we can do something like that, but it won't be trivial. You also have to cope with operations that may use the sub-structs that you are flattening, so you will have to reconstruct them if needed. If it was simple I would have implemented it by now.

hanw: I was thinking about `ControlBlock` and `ParserBlock`. I agree that we should not try to transform the parameter list of `ExternBlock`.

backends/bmv2/README.md
```diff
@@ -17,3 +17,36 @@ To run and test this back-end you need some additional tools:
...
+             inout standard_metadata_t standard_metadata) {
+  // ...
+}
+```
+
+- explicit transition to reject in parse state
+
+- compound action parameters (can only be `bit<>` or `int<>`)
+
+- functions or methods with a compund return type
```
hanw: @antoninbas can you please give an example of this?

antoninbas:
I am merging this, since @mbudiu-vmw approved and I think I have addressed all comments. Feel free to update this file if you discover more unsupported features.

PR #566, opened 2017-05-04
Generate header field names differently in p4runtime

antoninbas:
We generate the names differently based on whether the input is P4_14 or
P4_16. This allows us to:
 * be consistent with existing old tools in the P4_14 code
 * be more consistent for control plane name generation in P4_16. The
 existing behavior was suprising and did not match the language
 specification IMO.

### DIFF
--- a/control-plane/p4RuntimeSerializer.cpp
+++ b/control-plane/p4RuntimeSerializer.cpp
@@ -116,17 +116,21 @@ struct HeaderFieldPath {
                                  const TypeMap* typeMap) {
         auto type = typeMap->getType(expression->getNode(), true);
         if (expression->is<IR::PathExpression>()) {
-            // Unlike other top-level structs, top-level metadata structs are
-            // fully exposed to P4Runtime and show up in P4Runtime field names.
-            if (isMetadataType(type) && type->is<IR::Type_Declaration>()) {
-                auto name = controlPlaneName(type->to<IR::Type_Declaration>());
-                return HeaderFieldPath::root(name, type);
-            }
+            // for consistency with old P4_14 tools (e.g. the bmv2 JSON to p4runtime conversion
+            // tool), we generate names differently in P4_14 mode
+            if (refMap->isV1()) {
+                // Unlike other top-level structs, top-level metadata structs are fully exposed to
+                // P4Runtime and show up in P4Runtime field names.
+                if (isMetadataType(type) && type->is<IR::Type_Declaration>()) {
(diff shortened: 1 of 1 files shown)


antoninbas:
Here is an example program:
```
// This is a simple P4 program for testing user metadata mapping.
#include <core.p4>
#include <v1model.p4>

struct test_metadata_t {
  bit<32> test_meta1;
}

header ethernet_t {
  bit<48> dstAddr;
  bit<48> srcAddr;
  bit<16> etherType;
}

struct headers {
  ethernet_t ethernet;
}

parser TestParser(packet_in packet, out headers hdr, inout test_metadata_t meta,
  state start {
    packet.extract(hdr.ethernet);
    transition accept;
  }
}

control ingress(inout headers hdr, inout test_metadata_t meta_i,
                inout standard_metadata_t standard_metadata) { }

control egress(inout headers hdr, inout test_metadata_t meta_e,
  action nop() {
  }
  table egress_table {
    actions = {
      nop;
    }
    key = {
      meta_e.test_meta1: lpm;
      hdr.ethernet.dstAddr: exact;
      standard_metadata.egress_spec: exact;
    }
    default_action = nop();
  }
  apply {
    egress_table.apply();
  }
}

control DeparserImpl(packet_out packet, in headers hdr) {
  apply {}
}

control verifyChecksum(in headers hdr, inout test_metadata_t meta) {
  apply {}
}

control computeChecksum(inout headers hdr, inout test_metadata_t meta) {
  apply {}
}

V1Switch(TestParser(), verifyChecksum(), ingress(), egress(),
         computeChecksum(), DeparserImpl()) main;
```

With existing code, match field `meta_e.test_meta1` show up as `test_meta1` in p4runtime, while `standard_metadata.egress_spec` show up as `standard_metadata.egress_spec`. If we rename parameter `standard_metadata` in the `egress` control flow, the field still shows up as `standard_metadata.egress_spec` because of the v1model.p4 annotations (https://github.com/p4lang/p4c/blob/master/p4include/v1model.p4#L31).

The side-effect of this pull request is that `hdr.ethernet.dstAddr` will now show up as `hdr.ethernet.dstAddr` and no longer as `ethernet.dstAddr` but IMO that also makes more sense.

mbudiu-vmw:
We need to do something like this for the json too, this will solve several issues with duplicated names.

antoninbas:
@sethfowler Nice I was looking for something like this but couldn't find it, I'll make the change

antoninbas:
@sethfowler I made the change you suggested

PR #568, opened 2017-05-04
Add Doxygen-style documentation for the SimplifyControlFlow pass.

cole-barefoot:
There are also a couple minor TODOs that should be addressed.  I'm happy to update this PR with code fixes if the TODO comments make sense.


frontends/p4/simplify.cpp
```diff
@@ -33,7 +37,11 @@ const IR::Node* DoSimplifyControlFlow::postorder(IR::BlockStatement* statement)
         return statement;
     }
     bool withinBlock = statancestor != nullptr && statancestor->is<IR::BlockStatement>();
+
+    // TODO: withinAction will always be false: If parent is a P4Action, the if
```
mbudiu-vmw: This should probably be replaced with `findContext<IR::P4Action>() != null`.
I will file an issue about this.

PR #569, opened 2017-05-05
[do not merge] PSA definition

cc10512:
I'm providing this PR so that people can review the PSA definition as it stands today. I incorporated feedback from @hanw, @antoninbas, @ChrisDodd, @vgurevich, @mbudiu-vmw, and @jafingerhut. There are still a number of TODOs marked in this file. I would like to use this PR as the means to collect more feedback and provide a forum for the discussion that will allow us to finalize the PSA spec. Please share your thoughts.

### DIFF
--- a/p4include/psa.p4
+++ b/p4include/psa.p4
@@ -62,10 +62,6 @@ struct psa_ingress_input_metadata_t {
   ParserErrorLocation_t    parser_error_location;
 }
 
-struct psa_ingress_output_metadata_t {
-  PortId_t                 egress_port;
-}
-
 struct psa_egress_input_metadata_t {
   PortId_t                 egress_port;
   InstanceType_t           instance_type;  /// Clone or Normal
@@ -155,7 +151,7 @@ enum CloneMethod_t {
  *   From this, it follows that if there is a call that sets the
  *   multicast_group, the packet will be multicast to the group that
  *   last set the multicast group. Otherwise, the packet will be sent
- *   to the port set by either send_to_port.
+ *   to the port set by send_to_port.
  *
  * - multiple clone invocations will cause the packet to be cloned to
  *   the corresponding port. Any drop call in the pipeline will cause
@@ -346,7 +342,7 @@ enum HashAlgorithm {
   crc32_custom,
   crc16,
   crc16_custom,
-  random,
+  random,           /// are random hash algorithms useful?
   identity
 }
 
@@ -471,7 +467,7 @@ extern Meter<S> {
   @ControlPlaneAPI
   {
     reset(in MeterColor_t color);
-    setParams(in S committedRate, out S committedBurstSize
+    setParams(in S committedRate, in S committedBurstSize
               in S peakRate, in S peakBurstSize);
     getParams(out S committedRate, out S committedBurstSize
               out S peakRate, out S peakBurstSize);
@@ -488,7 +484,7 @@ extern DirectMeter {
   @ControlPlaneAPI
   {
     reset(in MeterColor_t color);
-    void setParams<S>(in S committedRate, out S committedBurstSize
+    void setParams<S>(in S committedRate, in S committedBurstSize
                       in S peakRate, in S peakBurstSize);
     void getParams<S>(out S committedRate, out S committedBurstSize
                       out S peakRate, out S peakBurstSize);
@@ -504,10 +500,10 @@ extern DirectMeter {
  * written in actions. Registers are similar to counters, but can be
  * used in a more general way to keep state.
  *
- * Although registers cannot be used directly in matching, they may be
- * used as the RHS of an assignment operation, allowing the current
- * value of the register to be copied into metadata and be available
- * for matching in subsquent tables.
+ * Although registers cannot be used directly in matching,
+ * register.read may be used as the RHS of an assignment operation,
+ * allowing the current value of the register to be copied into
+ * metadata and be available for matching in subsequent tables.
  */
 extern Register<T, S> {
   Register(S size);
@@ -533,6 +529,8 @@ extern Register<T, S> {
  */
 
 /// The set of distributions supported by the Random extern.
+/// \TODO: should this be removed in favor of letting the extern
+/// return whatever distribution is supported by the target?
 enum RandomDistribution {
   PRNG,
   Binomial,
@@ -654,11 +652,10 @@ control VerifyChecksum<H, M>(in H hdr, inout M user_meta);
 
 control Ingress<H, M>(inout H hdr, inout M user_meta,
                       PacketReplicationEngine pre,
-                      in  psa_ingress_input_metadata_t  istd,
-                      out psa_ingress_output_metadata_t ostd);
+                      in  psa_ingress_input_metadata_t  istd);
 
 control Egress<H, M>(inout H hdr, inout M user_meta,
-                     PacketReplicationEngine pre,
+                     BufferingQueueingEngine bqe,
                      in  psa_egress_input_metadata_t  istd);
 
 control ComputeChecksum<H, M>(inout H hdr, inout M user_meta);
--- a/p4include/psa.p4
+++ b/p4include/psa.p4
@@ -1,0 +1,675 @@
+/*
+Copyright 2013-present Barefoot Networks, Inc.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+#ifndef _PORTABLE_SWITCH_ARCHITECTURE_P4_
+#define _PORTABLE_SWITCH_ARCHITECTURE_P4_
+
+/**
+ *   P4-16 declaration of the Portable Switch Architecture
+ */
+
+/**
+ * These types need to be defined before including the architecture file
+ * and the macro protecting them should be defined.
+ */
+#ifndef PSA_CORE_TYPES
+#error "Please define the following types for PSA and the PSA_CORE_TYPES macro"
+typedef bit<unspecified> PortId_t;
+const   PortId_t         PORT_CPU = unspecified;
+
+typedef bit<unspecified> MulticastGroup_t;
+typedef bit<unspecified> PacketLength_t;
+typedef bit<unspecified> EgressInstance_t;
+typedef bit<unspecified> ParserStatus_t;
+typedef bit<unspecified> ParserErrorLocation_t;
+typedef bit<unspecified> entry_key;           /// for DirectCounters
+
+// typedef bit<unspecified> InstanceType_t;
+// const   InstanceType_t   INSTANCE_NORMAL = unspecified;
+#endif
+
+#include "core.p4"
+
+/* --------------------------------------------------------------------- *
+/**
+ * PSA supported metadata types
+ */
+typedef enum { NORMALINSTANCE_, CLONE_INSTANCE } InstanceType_t;
+
+struct psa_parser_input_metadata_t {
+  PortId_t                 ingress_port;
+  InstanceType_t           instance_type;
+}
+
+struct psa_ingress_input_metadata_t {
+  PortId_t                 ingress_port;
+  InstanceType_t           instance_type;  /// Clone or Normal
+  /// set by the runtime in the parser, these are not under programmer control
+  ParserStatus_t           parser_status;
+  ParserErrorLocation_t    parser_error_location;
+}
+
+struct psa_ingress_output_metadata_t {
+  PortId_t                 egress_port;
+}
+
+struct psa_egress_input_metadata_t {
+  PortId_t                 egress_port;
+  InstanceType_t           instance_type;  /// Clone or Normal
+  EgressInstance_t         instance;       /// instance coming from PRE
+}
+
+/* --------------------------------------------------------------------- */
+/**
+ * Additional supported match_kind types
+ */
+match_kind {
+    range,   /// Used to represent min..max intervals
+    selector /// Used for implementing dynamic_action_selection
+}
+
+/* --------------------------------------------------------------------- */
+/**
+ *  Cloning methods
+ */
+enum CloneMethod_t {
+  /// Clone method         Packet source             Insertion point
+  Ingress2Ingress,  /// original ingress,            Ingress parser
+  Ingres2Egress,    /// post parse original ingres,  Buffering queue
+  Egress2Ingress,   /// post deparse in egress,      Ingress parser
+  Egress2Egress     /// inout to deparser in egress, Buffering queue
+}
+
+/* --------------------------------------------------------------------- */
+/*
+ * PSA Externs
+ */
+
+/**
+ * The PacketReplicationEngine extern represents the non-programmable
+ * part of the PSA pipeline.
+ *
+ * Even though the PRE can not be programmed using P4, it can be
+ * configured both directly using control plane APIs and by setting
+ * intrinsic metadata. In this specification we opt to define the
+ * operations available in the PRE as method invocations. A target
+ * backend is responsible for mapping the PRE extern APIs to the
+ * appropriate mechanisms for performing these operations in the
+ * hardware.
+ *
+ * The PRE is instantiated by the architecture and a P4 program can
+ * use it directly. It is an error to instantiate the PRE multiple
+ * times.  The PRE is made available to the Ingress programmable block
+ * using the same mechanism as packet_in. A corresponding Buffering
+ * and Queueing Engine (BQE) extern is defined for the Egress
+ * pipeline (@see BufferingQueueingEngine).
+ *
+ * Note: some of these operations may not be implemented as primitive
+ * operations on a certain target. However, All of the operations can
+ * be implemented as a combination of other operations. Applications
+ * that rely on non-primitive operations may incur significant
+ * performance penalties, however, they should be functionally
+ * correct.
+ *
+ * Semantics of behavior for multiple calls to PRE APIs
+ *
+ * The semantics of calling the PRE APIs is equivalent to setting
+ * intrinsic metadata fields/bits and assumming that the PRE looks up
+ * the fields in the following order: drop, truncate, multicast,
+ * clone, output_port.
+ *
+ * Following this semantics, examples of the behaviors are:
+ *
+ * - any call to drop in the pipeline will cause the packet, and all
+ *   potential clone copies (see below) to drop.
+ *
+ * - any call to truncate, will cause the packet (and its clones
+ *   \todo: check) to be truncated.
+ *
+ * - multiple calls to send_to_port -- the last call in the ingress
+ *   pipeline sets the output port.
+ *
+ * - multiple calls to multicast -- the last in the ingress pipeline
+ *   sets the multicast group
+ *
+ * - interleaving send_to_port and multicast -- the semantics of multicast
+ *   is defined as below (https://github.com/p4lang/tutorials/issues/22):
+ *     if (multicast_group != 0)
+ *        multicast_to_group(multicast_group);
+ *     else
+ *        send_to_port(output_port);
+ *
+ *   From this, it follows that if there is a call that sets the
+ *   multicast_group, the packet will be multicast to the group that
+ *   last set the multicast group. Otherwise, the packet will be sent
+ *   to the port set by either send_to_port.
+ *
+ * - multiple clone invocations will cause the packet to be cloned to
+ *   the corresponding port. Any drop call in the pipeline will cause
+ *   the packet to drop, and no clone will be created (following the
+ *   analogy with intrinsic metadata bit fields, drop bits are
+ *   processed before clone bits are looked up).
+ *
+ * - resubmit
+ * - recirculate
+ * \TODO: finalize the semantics of calling multiple of the PRE APIs
+ *
+ */
+extern PacketReplicationEngine {
+
+  // PacketReplicationEngine(); /// No constructor. PRE is instantiated
+                                /// by the architecture.
+
+  /// Unicast operation: sends packet to port
+  ///
+  /// Targets may implement this operation by setting the appropriate
+  /// intrinsic metadata or through some other mechanism of
+  /// configuring the PRE.
+  ///
+  /// @param port The output port. If the port is PORT_CPU the packet
+  ///             will be sent to CPU
+  void send_to_port (in PortId_t port);
+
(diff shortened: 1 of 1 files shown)


p4include/psa.p4
```diff
@@ -0,0 +1,675 @@
...
+ * implementing controls that match these interfaces and instantiate
+ * them in a package definition.
+ *
+ * The current implementation uses the same user-defined metadata
+ * structure for all the controls.
+ * An alternative design is to split the user-defined metadata into an
+ * input parameter and an output parameter for each block. The
+ * compiler will have to check that the out parameter of a block
+ * matches the in parameter of the subsequent block.
+ *
```
jafingerhut: This is _not_ a suggestion to change anything, but just a comment on this choice for implementers.  I suspect what you have in the code here now is easier for a compiler to optimize out any fields of user_meta that are not needed by any later code, to minimize the number of bits passed down a pipeline implementation.

Easier, that is, than having a different user_meta type passed between each pair of consecutive 'blocks', requiring the user's P4 program to copy what they want to keep (in many cases likely unchanged, but requiring the assignments anyway).  Optimizing away those assignments sounds trickier (to me, not having thought about it very long) for a compiler than the 'optimize away unnecessary fields' approach aobve.

p4include/psa.p4
```diff
@@ -0,0 +1,675 @@
...
+ *   is defined as below (https://github.com/p4lang/tutorials/issues/22):
+ *     if (multicast_group != 0)
+ *        multicast_to_group(multicast_group);
+ *     else
+ *        send_to_port(output_port);
+ *
+ *   From this, it follows that if there is a call that sets the
+ *   multicast_group, the packet will be multicast to the group that
+ *   last set the multicast group. Otherwise, the packet will be sent
+ *   to the port set by either send_to_port.
```
jafingerhut: 'either' looks left over and should be removed.  Or there should be additional possibilities besides 'send_to_port' mentioned as alternatives.

p4include/psa.p4
```diff
@@ -0,0 +1,675 @@
...
+ * that rely on non-primitive operations may incur significant
+ * performance penalties, however, they should be functionally
+ * correct.
+ *
+ * Semantics of behavior for multiple calls to PRE APIs
+ *
+ * The semantics of calling the PRE APIs is equivalent to setting
+ * intrinsic metadata fields/bits and assumming that the PRE looks up
+ * the fields in the following order: drop, truncate, multicast,
+ * clone, output_port.
```
jafingerhut: Everything in this list except 'output_port' is one of the API calls.  Perhaps replace 'output_port' with 'send_to_port' for consistency?

Should recirculate be part of this list?

p4include/psa.p4
```diff
@@ -0,0 +1,675 @@
...
+
+control VerifyChecksum<H, M>(in H hdr, inout M user_meta);
+
+control Ingress<H, M>(inout H hdr, inout M user_meta,
+                      PacketReplicationEngine pre,
+                      in  psa_ingress_input_metadata_t  istd,
+                      out psa_ingress_output_metadata_t ostd);
+
+control Egress<H, M>(inout H hdr, inout M user_meta,
+                     PacketReplicationEngine pre,
```
jafingerhut: From comments earlier, perhaps this egress PacketReplicationEngine was intended to be a BufferingQueueingEngine instead?

p4include/psa.p4
```diff
@@ -0,0 +1,675 @@
...
+ * Hashes
+ */
+
+/// Supported hash algorithms
+enum HashAlgorithm {
+  crc32,
+  crc32_custom,
+  crc16,
+  crc16_custom,
+  random,
```
jafingerhut: Given the existence of the Random extern, of what use would a random HashAlgorithm be?  I can't think of one.  If it were there, it sounds like you would need to feed some packet header/metadata fields as input, which would simply be ignored, so why bother with the Hash extern?

p4include/psa.p4
```diff
@@ -0,0 +1,675 @@
...
+
+/// Counter types
+enum CounterType_t {
+    packets,
+    bytes,
+    packets_and_bytes
+}
+
+/// Counter: RFC-XXXX
+extern Counter<W, S> {
```
jafingerhut: I don't know all of the RFCs by a long shot, but I'd be a little surprised if there is an RFC covering counters the way there is for meters.  Counters are just too simple, unless maybe someone has gone to the trouble to make an RFC defining saturating vs. wrapping counters.

p4include/psa.p4
```diff
@@ -0,0 +1,675 @@
...
+struct psa_ingress_input_metadata_t {
+  PortId_t                 ingress_port;
+  InstanceType_t           instance_type;  /// Clone or Normal
+  /// set by the runtime in the parser, these are not under programmer control
+  ParserStatus_t           parser_status;
+  ParserErrorLocation_t    parser_error_location;
+}
+
+struct psa_ingress_output_metadata_t {
+  PortId_t                 egress_port;
```
jafingerhut: Is it intended to have both an egress_port field here, _and_ the send_to_port() method in the PRE object?  It seems simpler to have only one of the two, probably the PRE method, and eliminate this egress_port field.  I could imagine having both, and specifying what happens if you both assign to the egress_port field and call send_to_port(), but why bother creating the confusion?

p4include/psa.p4
```diff
@@ -0,0 +1,675 @@
...
+/// Meter
+extern Meter<S> {
+  Meter(S n_meters, MeterType_t type);
+  MeterColor_t execute(in S index, in MeterColor_t color);
+
+  /*
+  @ControlPlaneAPI
+  {
+    reset(in MeterColor_t color);
+    setParams(in S committedRate, out S committedBurstSize
```
jafingerhut: possible typo: 'out' should be 'in'?

p4include/psa.p4
```diff
@@ -0,0 +1,675 @@
...
+/// DirectMeter
+extern DirectMeter {
+  DirectMeter(MeterType_t type);
+  MeterColor_t execute(in MeterColor_t color);
+
+  /*
+  @ControlPlaneAPI
+  {
+    reset(in MeterColor_t color);
+    void setParams<S>(in S committedRate, out S committedBurstSize
```
jafingerhut: Same possible 'out' -> 'in' typo here

p4include/psa.p4
```diff
@@ -0,0 +1,675 @@
...
+ * Registers
+ *
+ * Registers are stateful memories whose values can be read and
+ * written in actions. Registers are similar to counters, but can be
+ * used in a more general way to keep state.
+ *
+ * Although registers cannot be used directly in matching, they may be
+ * used as the RHS of an assignment operation, allowing the current
+ * value of the register to be copied into metadata and be available
+ * for matching in subsquent tables.
```
jafingerhut: I don't see how one could use a register in the RHS of an assignment, without a read() method call.  If you mean the read() method call can be in the RHS of an assignment, then OK, but might be nice to be explicit about that in the comment.

typo: subsquent -> subsequent

p4include/psa.p4
```diff
@@ -0,0 +1,675 @@
...
+/**
+ * Random
+ *
+ * The random extern provides a reliable, target specific number generator
+ * in the min .. max range.
+ */
+
+/// The set of distributions supported by the Random extern.
+enum RandomDistribution {
+  PRNG,
```
jafingerhut: PRNG doesn't provide any information here.  I mean, every one of these is pseudo-random, yes?

Perhaps 'Uniform' instead?

p4include/psa.p4
cc10512: The advantage of defining metadata as subsets is that it sets clear bounds on the lifetime of the fields. Good liveness analysis in the compiler can make up for it. The disadvantage, as you say is the need to copy. However, if you have intrinsic metadata directly supported in the hardware, the copies will be unnecessary.  In such cases, the analysis to remove redundant copies is quite trivial. So, I think that in terms of performance the two approaches are equivalent, the analysis are both doable, and it really comes down to a matter of taste! :) And given that we currently already support the flattened view in the compiler, I took the easy path out ...

p4include/psa.p4
cc10512: It is a textual representation of the if above, because the if does not actually capture the semantics of multiple calls. Not sure why you think it doesn't fit ...

p4include/psa.p4
cc10512: they are actually the bits in the intrinsic metadata. I guess the statement needs clarification. We have intrinsic metadata defined in the v1model for most of these bits. What I intended to say is that one can think of those bits as "global state" and the API functions modifying the global state. The PRE looks at the state after all the functions have been called in the pipeline and makes the decision based on the state of the bits. I'll appreciate your help in rephrasing the paragraph accordingly.

p4include/psa.p4
cc10512: Thanks!

p4include/psa.p4
jafingerhut: The sentence makes perfect sense in English if you delete the word "either".  Having the word there cannot make it clearer in any way I can think of, only make it appear like you meant to write "either A or B", but you forgot to write the "or B" part.

p4include/psa.p4
cc10512: Not sure. A random permutation of the bits in the input? As opposed to a random number from the Random's distribution? I'll put a comment so that folks who could think of how to use this can chime in.

p4include/psa.p4
cc10512: Good point! Lost this in translation from one approach to the other.

p4include/psa.p4
cc10512: Indeed, it is the read operation on the register that can be on the RHS. Fixed.

p4include/psa.p4
cc10512: What do you think about specifying the distribution in general? It is likely that most hardware implementation of the Random extern will implement a fixed algorithm, and it is not clear how do we map all the PSA distribution to that one implementation. A software implementation could call different libraries and get the random distribution. So the question is: is it more portable to just say a random number generator with whatever algorithm is implemented in the target?

cc10512:
Andy, thanks a lot for the thoughtful feedback! I pushed another version addressing some of your comments.

p4include/psa.p4
jafingerhut: A uniform random generator is a useful thing on its own.

A uniform random generator plus some formulas using + - << and if's is enough to implement piecewise linear approximations of other generators.

A uniform random generator _plus_ a lookup table with an exact key that is a uniform random value enables one to implement whatever distribution you can imagine, if you are willing to spend the table's storage to do it (and it can even be modified by the control plane at run time).

Given all of those choices, and given that a high performance target would probably implement non-uniform distributions using one of the methods above, I would think that uniform random would be sufficient, and probably all that some targets would be willing to implement anyway.

p4include/psa.p4
jafingerhut: I think my comment was more basic than that.  The names drop, truncate, multicast, and clone are all names of methods in the PRE object.  The only one that is _not_ is output_port, which is the name of the argument for the method set_output_port.  Is there a good reason to make that name a different kind of name than the others in the list?

jafingerhut: Also, if you want to define the behavior in terms of state that is hidden beneath the API of an object, you may as well give a reference implementation of that API, yes?  Otherwise, it seems best to define the behavior in terms of the API calls made, not the internal state variables.

p4include/psa.p4
```diff
@@ -0,0 +1,672 @@
...
+ * written in actions. Registers are similar to counters, but can be
+ * used in a more general way to keep state.
+ *
+ * Although registers cannot be used directly in matching,
+ * register.read may be used as the RHS of an assignment operation,
+ * allowing the current value of the register to be copied into
+ * metadata and be available for matching in subsequent tables.
+ */
+extern Register<T, S> {
+  Register(S size);
```
vgurevich: I wonder if the constructor should be 
```
    Register<T>(S instance_count);
```
instead.

The reason is that the type of the value stored in the register should be provided to the constructor, not derived from usage.

Also, as a minor comment, I think that ```(S size)``` is not a very good name, because it is not clear whether this refers to the size of an individual register or the size of the register *array*. I propose we use ```(I instance_count)``` in the declaration and ```(I index)``` in the methods.

mbudiu-vmw: Constructors cannot have type parameters, they inherit the parameters of the class.

vgurevich:
As a general comment, I think we need to settle on a clear naming scheme. 

Should we use suffixes (e.g. _t) or not? When should we use them. What are they?
Should we use lower_case, CamelCase (or camelCase), or_moreComplex_scheme_t? (very often different objects use different naming schemes). 

All these need to be outlined in the beginning.

p4include/psa.p4
```diff
@@ -0,0 +1,672 @@
...
+
+  /// Unicast operation: sends packet to port
+  ///
+  /// Targets may implement this operation by setting the appropriate
+  /// intrinsic metadata or through some other mechanism of
+  /// configuring the PRE.
+  ///
+  /// @param port The output port. If the port is PORT_CPU the packet
+  ///             will be sent to CPU
+  void send_to_port (in PortId_t port);
```
vgurevich: Can one query the current value (set by an action in a preceding table)? If not, why not?

p4include/psa.p4
```diff
@@ -0,0 +1,672 @@
...
+  /// Multicast operation: sends packet to a multicast group or a port
+  ///
+  /// Targets may implement this operation by setting the appropriate
+  /// intrinsic metadata or through some other mechanism of
+  /// configuring the PRE.
+  ///
+  /// @param multicast_group Multicast group id. The control plane
+  ///        must program the multicast groups through a separate
+  ///        mechanism.
+  void multicast (in MulticastGroup_t multicast_group);
```
vgurevich: What is the interaction between send_to_port()/multicast() and drop()? 
Do they override each other? 
How can one see that using these actions creates an action dependency (or does not)?
If one action calls send_top_port() and another one calls multicast(), how can a table later in the flow determine where the packet is currently going?

p4include/psa.p4
```diff
@@ -0,0 +1,672 @@
...
+  /// The PSA specifies four types of cloning, with the packet sourced
+  /// from different points in the pipeline and sent back to ingress
+  /// or to the buffering queue in the egress (@see CloneMethod_t).
+  /// \TODO: needs both source and destination data
+  ///
+  /// @param clone_method  The type of cloning.
+  /// @param port          The port to send the cloned packet to.
+  ///
+  /// Note: based on discussions with Antonin, we can ignore session now
+  void clone     (in CloneMethod_t clone_method, in PortId_t port);
```
vgurevich: Are we saying that because  clone() and drop() are allowed in the Egress() control, that the ReplicationEngine extern can be called in it too? That makes little sense to me, because it means that some methods of the same extern can be called only in the ingress, some in the egress and others in both.

Also, it puts a lot of additional pressure on the backend, since it has to determine the legality of a call by analyzing the actual value of clone_method. I do not think this is a good idea either

p4include/psa.p4
```diff
@@ -0,0 +1,672 @@
...
+  ///
+  /// The PSA specifies four types of cloning, with the packet sourced
+  /// from different points in the pipeline and sent back to ingress
+  /// or to the buffering queue in the egress (@see CloneMethod_t).
+  /// \TODO: needs both source and destination data
+  ///
+  /// @param clone_method  The type of cloning.
+  /// @param port          The port to send the cloned packet to.
+  /// @param data additional header data attached to the packet
+  void clone<T>  (in CloneMethod_t clone_method, in PortId_t port, in T data);
```
vgurevich: For all these 3 methods (clone(), resubmit() and recirculate()) it is totally not clear what happens to the ```T data```.

p4include/psa.p4
```diff
@@ -0,0 +1,672 @@
...
+
+  /// Unicast operation: sends packet to port
+  ///
+  /// Targets may implement this operation by setting the appropriate
+  /// intrinsic metadata or through some other mechanism of
+  /// configuring the BQE.
+  ///
+  /// @param port The output port. If the port is PORT_CPU the packet
+  ///             will be sent to CPU
+  void send_to_port (in PortId_t port);
```
vgurevich: Sending to a port AFTER egress is typically not possible or extremely expensive on most modern architectures.

p4include/psa.p4
```diff
@@ -0,0 +1,672 @@
...
+  /// the semantics supported by the PSA is that the drop will be at
+  /// the end of the pipeline (ingress or egress).
+  void drop      ();
+
+  /// Truncate operation
+  ///
+  /// Truncate the outgoing packet to the specified length
+  ///
+  /// @param length packet length
+  void truncate(in bit<32> length);
```
vgurevich: What's the interaction of this primitive with the deparser and checksumming?

p4include/psa.p4
```diff
@@ -0,0 +1,672 @@
...
+/// @param Algo The algorithm to use for computation (@see HashAlgorithm).
+/// @param O    The type of the return value of the hash.
+///
+/// Example usage:
+/// parser P() {
+///   Hash<crc16, bit<56>> h();
+///   bit<56> hash_value = h.getHash(16, buffer, 100);
+///   ...
+/// }
+extern Hash<Algo, O> {
```
vgurevich: Algo is not used

p4include/psa.p4
```diff
@@ -0,0 +1,672 @@
...
+/// Counter types
+enum CounterType_t {
+    packets,
+    bytes,
+    packets_and_bytes
+}
+
+/// Counter: RFC-XXXX
+extern Counter<W, S> {
+  Counter(S n_counters, W size_in_bits, CounterType_t counter_type);
```
vgurevich: We have to be careful here. For example if this is packets_and_bytes, then the width for the packet and byte counter (if less than 64 bits) is typically different and thus might require two separate arguments.

p4include/psa.p4
```diff
@@ -0,0 +1,672 @@
...
+  void count(in S index, in W increment);
```
vgurevich: Why does one need to specify the increment? It should be 1 for packets and packet_length for bytes (the last one is often done using very tricky mechanisms). If someone needs anything else, that's what registers are for.

Also, why increment has the type W? Are you saying that you can't increment a 64-bit counter by more than 64 (or 128)?

p4include/psa.p4
```diff
@@ -0,0 +1,672 @@
...
+    void reset        (in S index);
+    void start        (in S index);
+    void stop         (in S index);
+  }
+  */
+}
+
+/// DirectCounter: RFC-XXXX
+extern DirectCounter<W> {
+  DirectCounter(entry_key key, CounterType_t counter_type);
```
vgurevich: What is entry_key????

p4include/psa.p4
```diff
@@ -0,0 +1,672 @@
...
+ * are 3-color meters.
+ */
+
+/// Meter types
+enum MeterType_t {
+    packets,
+    bytes
+}
+/// Meter colors
+enum MeterColor_t { RED, GREEN, YELLOW };
```
vgurevich: Enum is a bad idea here, because there is a need to store this value in both the metadata and the packet.

p4include/psa.p4
```diff
@@ -0,0 +1,672 @@
...
+  { }
+  */
+}
+
+/* --------------------------------------------------------------------- */
+/// Checksum computation
+///
+/// @param W    The width of the checksum
+extern Checksum<W> {
+  Checksum(HashAlgorithm hash);          /// constructor
```
vgurevich: This is probably an overkill (to use the same list of algorithms as for a hash unit).

p4include/psa.p4
```diff
@@ -0,0 +1,672 @@
...
+extern DirectCounter<W> {
+  DirectCounter(entry_key key, CounterType_t counter_type);
+  void count();
+
+  /*
+  @ControlPlaneAPI
+  {
+    W    read<W>      (in entry_key key);
+    W    sync_read<W> (in entry_key key);
+    void set          (in W seed);
```
vgurevich: What does ```seed``` mean in the counter context?

p4include/psa.p4
```diff
@@ -0,0 +1,672 @@
...
+/**
+ *  Action profiles are used as table implementation attributes
+ *
+ * Action profiles implement a mechanism to populate table entries
+ * with actions and action data. The only data plane operation
+ * required is to instantiate this extern. When the control plane adds
+ * entries (members) into the extern, they are essentially populating
+ * the corresponding table entries.
+ */
+extern ActionProfile {
```
vgurevich: Is the extern instantiated in the table or outside?

p4include/psa.p4
```diff
@@ -0,0 +1,672 @@
...
+
+
+
+/* --------------------------------------------------------------------- */
+/// \TODO: is generating a new packet and sending it to the stream or is
+/// it adding a header to the current packet and sending it to the
+/// stream (copying or redirecting).
+extern Digest<T> {
+  Digest(PortId_t receiver); /// define a digest stream to receiver
+  void emit(in T data);      /// emit data into the stream
```
vgurevich: How do we plan to ensure that the data is "latched", i.e. that changing the data after the method has been called does not affect the results.

mbudiu-vmw: This is a bmv2 specific question. The current compiler always copies the data to a temporary for digest.

p4include/psa.p4
```diff
@@ -0,0 +1,672 @@
...
+ * input parameter and an output parameter for each block. The
+ * compiler will have to check that the out parameter of a block
+ * matches the in parameter of the subsequent block.
+ *
+ */
+
+parser Parser<H, M>(packet_in buffer, out H parsed_hdr, inout M user_meta,
+                    in psa_parser_input_metadata_t istd);
+
+control VerifyChecksum<H, M>(in H hdr, inout M user_meta);
```
vgurevich: What's the mechanism to pass checksum errors? user_meta?

p4include/psa.p4
jafingerhut: @vgurevich I have never heard of a use case for storing this red/green/yellow value in the packet.  Have you?  Mostly I have seen a 3-way branch in the code after updating the meter, that does different things depending upon the value of red/green/yellow, e.g. drop on red, modify DSCP on yellow, no-op on green.  That said, I have no problem using types that are easily converted to/from bit vectors.

p4include/psa.p4
```diff
@@ -0,0 +1,672 @@
...
+struct psa_parser_input_metadata_t {
+  PortId_t                 ingress_port;
+  InstanceType_t           instance_type;
+}
+
+struct psa_ingress_input_metadata_t {
+  PortId_t                 ingress_port;
+  InstanceType_t           instance_type;  /// Clone or Normal
+  /// set by the runtime in the parser, these are not under programmer control
+  ParserStatus_t           parser_status;
```
vgurevich: What's the relationship between ```error``` and  ```ParserStatus_t```.  What about checksum errors?

cc10512:
superseded by #618

PR #575, opened 2017-05-05
Fix bmv2 backend crash for unused action selectors

antoninbas:
When an action selector is declared as a control local but never
referenced by a table, the compiler cannot figure out the selector
input, which means the action selector cannot be included in the bmv2
JSON.

### DIFF
--- a/testdata/p4_16_samples/action_selector_unused-bmv2.p4
+++ b/testdata/p4_16_samples/action_selector_unused-bmv2.p4
@@ -1,0 +1,41 @@
+#include <core.p4>
+#include <v1model.p4>
+
+struct H { };
+struct M {
+    bit<32> hash1;
+}
+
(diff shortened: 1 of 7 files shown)


testdata/p4_16_samples/action_selector_unused-bmv2.p4
```diff
@@ -0,0 +1,52 @@
...
+    state start { transition accept; }
+}
+
+action empty() { }
+
+control IngressI(inout H hdr, inout M meta, inout standard_metadata_t smeta) {
+
+    action drop() { smeta.drop = 1; }
+
+    action_selector (HashAlgorithm.identity, 32w1024, 32w10) as;
```
mbudiu-vmw: Where is the unused selector?

antoninbas:
@mbudiu-vmw my bad I pushed the wrong test file. I fixed it.

antoninbas:
@mbudiu-vmw could you update your review for this?

PR #577, opened 2017-05-07
Correct issue where getSourceLine() could return -1 mod 2^32 instead …

jafingerhut:
…of 0

### DIFF
--- a/lib/source_file.cpp
+++ b/lib/source_file.cpp
@@ -168,7 +168,12 @@ SourceFileLine InputSources::getSourceLine(unsigned line) const {
     // ----------
     // The first line indicates that line 2 is the first line in x.p4
     // line=2, it->first=1, it->second.sourceLine=1
-    // So we have to subtract one to get the real line number
+    // So we have to subtract one to get the real line number,
+    // except in the special case of all 0s, where 0 is preferable to
+    // (-1 mod 2^32)
+    if (line == 0 && it->first == 0 && it->second.sourceLine == 0) {
+        return SourceFileLine(it->second.fileName, 0);
+    }
     return SourceFileLine(it->second.fileName, line - it->first + it->second.sourceLine - 1);
 }
 


jafingerhut:
There are other cases I have noticed where the SourceInfo objects created by the compiler for synthesized code is a bit odd, but I will save enhancing those issues for later.

lib/source_file.cpp
```diff
@@ -168,7 +168,12 @@ SourceFileLine InputSources::getSourceLine(unsigned line) const {
     // ----------
     // The first line indicates that line 2 is the first line in x.p4
     // line=2, it->first=1, it->second.sourceLine=1
-    // So we have to subtract one to get the real line number
+    // So we have to subtract one to get the real line number,
+    // except in the special case of all 0s, where 0 is preferable to
+    // (-1 mod 2^32)
+    if (line == 0 && it->first == 0 && it->second.sourceLine == 0) {
+        return SourceFileLine(it->second.fileName, 0);
```
ChrisDodd: It seems to me that this should never happen, and if it does, it indicates a bug somewhere else.  So it might be better to have a BUG_CHECK here rather than silently covering up the problem.

jafingerhut: If you compile this program: https://github.com/jafingerhut/p4-guide/blob/master/demo5/demo5.p4_16.p4#L89

using the command `p4c-bm2-ss -o tmp.json demo5.p4_16.p4`, the action `top_level_action`, being defined at the top level, and having an `inout` struct of all headers, but only modifying one field of one header inside of that struct, is compiled to contain several `copy_header` calls synthesized by the compiler to implement the copy-in and copy-out semantics.  One of those has source info of 2^32-1 in the output, which becomes 0 if you make the diff suggested in this PR.

This might very well be a bug elsewhere in the compiler, but I haven't tried to trace it backwards from the point of this change.

ChrisDodd: I've added a fix (PR #588) that fixes the problem with the json generation code that was triggering the problem you see (and which isn't really fixed by this -- it will just output 0 as the line number to avoid crashing BMV2, rather than the correct line number)

jafingerhut: Closing this PR in favor of @ChrisDodd better one: https://github.com/p4lang/p4c/pull/588

PR #579, opened 2017-05-08
fix lookahead byte order issue

hanw:


hanw:
@mbudiu-vmw I updated the reference outputs on this one.

PR #580, opened 2017-05-08
Run tests with MAKEFLAGS=-j2

sethfowler:
Since we stopped persisting the `MAKEFLAGS` environment variable in the image, we need to pass it in manually when running the tests. Failing to do this was doubling the length of time our tests took to run.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,5 +7,5 @@ install:
   - docker build -t p4c --build-arg IMAGE_TYPE=test .
 
 script:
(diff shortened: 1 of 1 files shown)


sethfowler:
Travis doesn't provide such an environment variable, but we could add one ourselves so that it's set in one place.

sethfowler:
I went ahead and merged this just to speed up the tests ASAP. We can do cleanup in a followup PR.

PR #581, opened 2017-05-08
Set MAKEFLAGS in a single place in .travis.yml

sethfowler:
This PR sets MAKEFLAGS for the testsfrom a Travis environment variable.


sethfowler:
(Doesn't make things any less verbose, unfortunately, since you have to be explicit about which environment variables are exposed to the container.)

Issue #583, opened 2017-05-09
New p4c alpha compiler fails compile for old HLIR statefu.p4 file


familysingh:
In the p4-hlir/tests directory, there is a stateful.p4 file (https://github.com/p4lang/p4-hlir/blob/master/tests/stateful.p4).  My failure is at the bottom of this writeup while another user's failure is included below.

From: Antonin Bas [mailto:antonin@barefootnetworks.com] 
Sent: Monday, May 08, 2017 1:05 PM
To: hemant@mnkcg.com
Cc: Mihai Budiu <mbudiu@vmware.com>; p4-dev <p4-dev@lists.p4.org>
Subject: Re: [P4-dev] P4-16 HLIR with struct support?

Compilation fails for me as well, but with a different error message:

includes/headers.p4(18): error: header ipv4_t: Duplicates declaration header ipv4_t
header_type ipv4_t {
^
includes/headers.p4(18)
header_type ipv4_t {
^
error: 1 errors encountered, aborting compilation

 

------------------------------------------------------------------------
I used the following p4c compilation.  The compiler fails to compile this P4-14 file.  Did I do anything wrong using the p4c compiler to compile P4-14 code?  
~/p4-hlir/tests$ p4c -S -x p4-14 stateful.p4 
 
 
includes/parser.p4(46): error: : No header type ipv4_t
header ipv4_t ipv4;
^^^^^^^^^^^^^^^^^^^
includes/parser.p4(46)
header ipv4_t ipv4;
       ^^^^^^
error: 1 errors encountered, aborting compilation
 
~/p4-hlir/tests$

PR #587, opened 2017-05-09
Fix for issue #583

mbudiu-vmw:
Also, slightly improved comments and debugging output in the p4-14 typecheck file.

### DIFF
--- a/frontends/p4-14/typecheck.cpp
+++ b/frontends/p4-14/typecheck.cpp
@@ -16,15 +16,18 @@ limitations under the License.
 
 #include "typecheck.h"
 
-// P4 v1.0 and v1.1 type checking algorithm
-// Initial type setting based on immediate context:
-// - replace named reference to ActionParams in the bodies of ActionFunctions with the
-//   actual ActionParam
-// - link header/metadata instances to the Type
-// - replace named references to global header or metadata instances with ConcreteHeaderRef
-//   expressions that link directly to them.
-// - set type for Member and HeaderStackItemRefs
+/// P4-14 (v1.0 and v1.1) type checking algorithm
+/// Initial type setting based on immediate context:
+/// - replace named reference to ActionParams in the bodies of ActionFunctions with the
+///   actual ActionParam
+/// - link header/metadata instances to the Type
+/// - replace named references to global header or metadata instances with ConcreteHeaderRef
+///   expressions that link directly to them.
+/// - set type for Member and HeaderStackItemRefs
 class TypeCheck::AssignInitialTypes : public Transform {
+ public:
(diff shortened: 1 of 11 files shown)


frontends/p4-14/typecheck.cpp
```diff
@@ -16,15 +16,19 @@ limitations under the License.
...
-// - set type for Member and HeaderStackItemRefs
+/// P4-14 (v1.0 and v1.1) type checking algorithm
+/// Initial type setting based on immediate context:
+/// - replace named reference to ActionParams in the bodies of ActionFunctions with the
+///   actual ActionParam
+/// - link header/metadata instances to the Type
+/// - replace named references to global header or metadata instances with ConcreteHeaderRef
+///   expressions that link directly to them.
+/// - set type for Member and HeaderStackItemRefs
 class TypeCheck::AssignInitialTypes : public Transform {
```
cc10512: Is there any particular reason you're using nested classes rather than namespaces?

mbudiu-vmw: I have only changed the comments here. This code was written by @ChrisDodd.

frontends/p4/fromv1.0/programStructure.cpp
```diff
@@ -143,6 +145,13 @@ void ProgramStructure::createTypes() {
         auto type = it.first->type;
         createType(type, true, &converted);
     }
+    for (auto it : registers) {
+        if (it.first->layout) {
+            auto type = types.get(it.first->layout);
+            if (type->is<IR::Type_Header>())
+                createType(type, true, &converted);
```
ChrisDodd: Register layouts should always be treated as structs, not headers (should not have valid bits), They're only declared as header_types in P4_14, because that is the only way to define structs in P4_14 -- whether something is a header or metadata is determined by the instance declaration.

frontends/p4-14/typecheck.cpp
ChrisDodd: I generally define nested visitors for a PassManager subclass as nested classes in the PassManager subclass; seems cleaner to me that defining a separate namespace in addition to the class.

frontends/p4/fromv1.0/programStructure.cpp
mbudiu-vmw: Then let's not yet merge this yet.

mbudiu-vmw: It may be that the best place to do the conversion to a struct is somewhere earlier, in the P4-14 typechecker. I will try to figure it out.

mbudiu-vmw: Actually the layout is just a string, so that solution won't work.

PR #588, opened 2017-05-09
Don't use an invalid srcInfo to construct json source info.

ChrisDodd:
- Also fix off-by-one error tracking stdin line number

I think there are more off-by-one errors in the linenumber tracking code, combined with various +1/-1 adjustments in random places to correct for them.  Much of this comes from #line directives containing the line number of the next line, not the line number of the #line directive itself.

### DIFF
--- a/ir/node.cpp
+++ b/ir/node.cpp
@@ -83,7 +83,8 @@ Util::JsonObject* IR::Node::sourceInfoJsonObj() const {
         auto assign = to<IR::AssignmentStatement>();
         lhs = assign->left;
         rhs = assign->right;
-        lhs->srcInfo.toSourcePositionData(&lineNumber, &columnNumber);
+        if (lhs->srcInfo.isValid())
+            lhs->srcInfo.toSourcePositionData(&lineNumber, &columnNumber);
     }
     auto json = new Util::JsonObject();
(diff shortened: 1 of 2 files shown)


ChrisDodd:
The json code for getting the source fragment of assignments also seems broken to me.

jafingerhut:
I am aware that the code for getting the source fragment of assignments often produces weird results, at least for assignments synthesized by the compiler.

It does produce good results for most assignments that appear in the original source code.  I'm happy to open a separate issue to remember that those should be improved in some way.

ChrisDodd:
In general, you should be checking the SourceInfo objects with isValid() before trying to map them back to source code -- when isValid() is false, there's no source information (usually because the node was created by the compiler where it had no source info to put in it, but sometimes just because it failed to provide that argument to the constructor).

ChrisDodd:
In addition, you can "concatenate" SourceInfo objects with `+`.  If both are valid the resulting SourceInfo will be from the start of the first one to the end of the second.  If only one is valid, that one will be returned unchanged.

jafingerhut:
@ChrisDodd Thanks for the tip on "+" being defined for SourceInfo objects.  I will work on a patch to the assignment code I had to use that technique instead, which should improve the code.

Issue #584, opened 2017-05-09
v1model hash() yielding frontend compiler bug


Xingdao-Fan:
I apologize if this is intended behavior.
The following lines are yielding a compiler bug when compiled with `./p4c-bm2-ss --p4v 16` or `./p4test --p4v 16`.

# 1
```C++
control TopPipe(inout Parsed_packet hdr) {

   apply {
       bit<16> var;

       hash(var, HashAlgorithm.crc16, 0, hdr, 0xFFFF);
   }
}
```

> In file: ../frontends/p4/typeChecking/bindVariables.cpp:15
> Compiler Bug: ../frontends/p4/typeChecking/bindVariables.cpp:15: Null rtype

# 2
The bug goes away if I make explicit typecasts:
```C++
//this will compile just fine
hash(var, HashAlgorithm.crc16, (bit<16>) 0, hdr, (bit<16>) 0xFFFF); 
```

# 3
But does not accept typedef types.

```C++
typedef bit<16> Hash;

control TopPipe(inout Parsed_packet hdr) {

   apply {
       Hash var;

       hash(var, HashAlgorithm.crc16, (Hash) 0, hdr, (Hash) 0xFFFF);
   }
}
```
> ../../f/skeleton.p4(77): error: cast: Illegal cast from Hash to bit<16>
>  hash(var, HashAlgorithm.crc16, (Hash) 0, hdr, (Hash) 0xFFFF);

This message is printed twice, pointing at `(Hash) 0` and `(Hash) 0xFFFF`.

## Side note
The compiler is also demanding that the fourth parameter be of `struct` type, but I assume this is intentional (`field_list_calculation` is now `struct`).

# Complete P4 program
```C++
#include "./p4include/v1model.p4"

typedef bit<16> Hash;

//------------------------------------------
header head {
    bit<512> field;
}

//------------------------------------------
struct Meta {
    bit<5> metafield;
}

struct Parsed_packet {
    head h;
}

//------------------------------------------

parser TopParser(packet_in b, out Parsed_packet p, inout Meta m, inout standard_metadata_t metadata) {
  state start {
    transition accept;
  }
}

control VeryChecksum(in Parsed_packet hdr, inout Meta meta) {
	apply {  }
}

control IngressP(inout Parsed_packet hdr,
                      inout Meta meta,
                      inout standard_metadata_t standard_metadata) {

  apply {
     bit<16> var;

     hash(var, HashAlgorithm.crc16, 0, hdr, 0xFFFF);
[...]

PR #589, opened 2017-05-09
fix for issue #584

mbudiu-vmw:


frontends/p4/typeChecking/bindVariables.cpp
```diff
@@ -12,8 +12,7 @@ const IR::Type* BindTypeVariables::getP4Type(const IR::Type* type) const {
     if (type == nullptr)
         return type;
     auto rtype = type->getP4Type();
-    CHECK_NULL(rtype);
-    return rtype;
+    return rtype;  // This may be nullptr
```
cc10512: is the nullptr checked at the call site?

mbudiu-vmw: Yes, there is exactly one call of this method.
But as you can see from two lines above, this method could already return nullptr.

PR #594, opened 2017-05-09
Improve dependency checks in configure.ac

sethfowler:
This PR makes a few improvements to the dependency checks in `configure.ac`:

- We now explicitly depend on `libfl-dev`. For some reason Debian (and thus Ubuntu) decided to split out some of the files included with flex, including `FlexLexer.h` which we depend on, into a separate package. This issue wasn't noticed until Mihai ran into it because the package is installed as a dependency of something else in our Docker CI environment. No change is necessary on macOS; a full flex installation comes with the system by default.

- We don't depend `constraint_solver.h` (which is from Google OR-Tools), so we shouldn't be checking for it.

- We don't need to check for ranlib; `LT_INIT` takes care of it. The `AC_PROG_RANLIB` check was triggering warnings from autoconf.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -19,6 +19,7 @@ ENV P4C_DEPS automake \
              g++ \
              libboost-dev \
              libboost-iostreams1.58-dev \
+             libfl-dev \
              libgc-dev \
              libgmp-dev \
              libtool \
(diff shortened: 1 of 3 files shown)


ChrisDodd:
I know the call to AC_PROG_RANLIB is there because certain versions of autoconf require it; when I tried to remove it before some build somewhere failed as a result...

sethfowler:
OK, I'll remove the commit that takes out AC_PROG_RANLIB.

PR #596, opened 2017-05-10
Add/convert Doxygen-style comments for SpecializeAll.

cole-barefoot:


frontends/p4/specialize.h
```diff
@@ -24,31 +24,21 @@ limitations under the License.
...
+/// Describes how a parser or control is specialized.
 struct SpecializationInfo {
-    // Name to use for specialized object
+    /// Name to use for specialized object.
     cstring name;
-    // Actual parser or control that is being specialized
+    /// Actual parser or control that is being specialized.
     const IR::IContainer*              specialized;
-    // values to substitute for type arguments
+    /// values to substitute for type arguments.
```
mbudiu-vmw: Maybe capitalization should be uniform too.

frontends/p4/specialize.h
```diff
@@ -113,6 +117,42 @@ class FindSpecializations : public Inspector {
...
+c<bit<32>>(16) c_inst;
+```
+ *
+ * is converted to
+ *
+```
+control cspec(in bit<32> data) { ... }
+cspec() c_inst;
+```
+ *
```
mbudiu-vmw: Maybe you can add here that the body of cspec if obtaining from the body of c where T is substituted with bit<32> and size is substituted with 16.

frontends/p4/specialize.h
```diff
@@ -127,6 +167,21 @@ class Specialize : public Transform {
     const IR::Node* postorder(IR::Declaration_Instance*) override;
 };
 
+/** @brief Repeats ConstantFolding, TypeChecking, FindSpecializations, and
```
mbudiu-vmw: This description is too much tied to the implementation.
I would rather say that it specializes all parser and controls that are used in the program.

The post is actually that there are no more parsers and controls with generic types or constructor arguments in the program.

frontends/p4/specialize.h
```diff
@@ -113,6 +117,42 @@ class FindSpecializations : public Inspector {
...
+cspec() c_inst;
+```
+ *
+ * A different specialization is made for each constructor invocation
+ * and Declaration_Instance.
+ *
+ * @pre Actual arguments are suitable for substitution, meaning that it must
+ * run after ConstantFolding and SimplifyExpressions.
+ *
+ * @post No instantiations of Parsers or Controls with constructor parameters
```
mbudiu-vmw: Actually this may be incorrect; there won't remain such instantiations only after SpecializeAll.
We need to iterate because we specialize from the top-down: we need to know the arguments to specialize, so the arguments must be compile-time constants.

frontends/p4/specialize.h
cole-barefoot: Good catch.

frontends/p4/specialize.h
cole-barefoot: Yes, good idea.  I'll add that.

frontends/p4/specialize.h
cole-barefoot: Oh I see.  This handles the case where, say, a control is instantiated with a constructor argument that is itself a constructor parameter?

mbudiu-vmw: Or an expression that involves constructor parameters.
That's why you need constant folding; these are all compile-time constants, but you have to evaluate them.

frontends/p4/specialize.h
cole-barefoot: I used the names of the passes because Doxygen automatically inserts links to their documentation.  But I'm fine replacing that with a more generic description.

PR #601, opened 2017-05-10
Part 1 for PSA support

hanw:
This is a rather large commit to refactor bmv2 backend to enable easier support for PSA.

A few major changes in this commit:
1. Separate JSON generation from IR manipulation. The `jsonconverter.cpp` had two jobs: traversing the IR and generating the corresponding JSON template. We introduced a new `JsonObjects` class that handles the JSON template generation and refactored `jsonconverter.cpp` into individual `header.cpp`, `control.cpp`, `extern.cpp`, `action.cpp`, `parser.cpp`, `deparser.cpp` files to limit each file to one aspect of the IR.
2. Separate V1Model-specific code generation routine. `primitive`s are handled differently by v1model and PSA. We separated out the v1model-specific primitive generation into `v1model.cpp` to be backward-compatible with bm2-ss.
3. Preliminary extern generation for PSA.

As the code stands right now, it still generates valid JSON for bm2-ss. Our next step is to enable JSON generation for bm2-psa, and test with PSA behavior model.

In the interest of not keep a branch for too long, I hope we can fold this change in, to allow others to modify the refactored code, instead of adding changes to `jsonconverter.cpp`.


sethfowler:
Yeah, we should get this review and landed quickly, since it touches a huge number of files. It looks like you need to rebase already.

backends/bmv2/JsonObjects.cpp
```diff
@@ -0,0 +1,336 @@
...
+}
+
+/// insert a json array named 'parameters' in a parent json object
+Util::JsonArray*
+JsonObjects::create_parameters(Util::JsonObject* object) {
+    return insert_array_field(object, "parameters");
+}
+
+/// append a json object r to a parent json array
+/// insert a field 'op' with 'name' to r
```
cc10512: s/r/parent

backends/bmv2/action.h
```diff
@@ -0,0 +1,46 @@
...
+#define _BACKENDS_BMV2_ACTION_H_
+
+#include "ir/ir.h"
+#include "backend.h"
+
+namespace BMV2 {
+
+class ConvertActions : public Inspector {
+    Backend*               backend;
+    P4::ReferenceMap*      refMap;
```
cc10512: Is there a particular reason to have these as data members? You are initializing them from the backend data members.

backends/bmv2/backend.cpp
```diff
@@ -0,0 +1,107 @@
...
+#include "extern.h"
+#include "frontends/p4/methodInstance.h"
+#include "header.h"
+#include "metadata.h"
+#include "parser.h"
+#include "JsonObjects.h"
+
+namespace BMV2 {
+
+void Backend::createFieldAliases(const char *remapFile) {
```
cc10512: This is not needed anymore, is it? We're not using remap files.

backends/bmv2/metadata.cpp
```diff
@@ -0,0 +1,76 @@
...
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+/*
+Copyright 2013-present Barefoot Networks, Inc.
```
cc10512: two copyright headers is one too many :)

backends/bmv2/metadata.cpp
```diff
@@ -0,0 +1,76 @@
...
+    }
+    input.close();
+    return remap;
+}
+
+}  // end namespace Arch
+
+namespace BMV2 {
+
+}  // end namespace BMV2
```
cc10512: this is not needed anymore.

backends/bmv2/metadata.h
```diff
@@ -0,0 +1,67 @@
...
+    const IR::Node * postorder(const IR::Declaration_Variable* decl) override
+    { std::cerr << decl << std::endl; };
+};
+
+}  // namespace BMV2
+
+#endif
+
+#endif  // _BACKENDS_BMV2_METADATA_H_
+
```
cc10512: likewise. this should be removed.

backends/bmv2/metermap.cpp
```diff
@@ -0,0 +1,96 @@
...
+        if (!same)
+            ::error("On this target all meter operations must write to the same destination "
+                    "but %1% and %2% are different", destination, info->destinationField);
+    }
+}
+
+/**
+    Set the size of the table that a meter is attached to.
+
+    @param meter
```
cc10512: unless you comment the parameters themselves, it's not needed to just add them as a ```@param```.

backends/bmv2/metermap.h
```diff
@@ -0,0 +1,47 @@
...
+*/
+
+#ifndef _BACKENDS_BMV2_METERMAP_H_
+#define _BACKENDS_BMV2_METERMAP_H_
+
+#include "ir/ir.h"
+
+namespace BMV2 {
+
+class DirectMeterMap final {
```
cc10512: class comment?

p4include/p4d2model_bmss_meta.map
```diff
@@ -0,0 +1,29 @@
+# This file maps p4d2 metadata to the bmv2 simple switch standard and intrinsic metadata
```
cc10512: remove this file too.

testdata/p4_16_samples/ipv4_forward-bmv2.p4
```diff
@@ -0,0 +1,159 @@
+/* -*- P4_16 -*- */
+#include <core.p4>
+#include <p4d2model.p4>
```
cc10512: is this test still failing? should we convert it back to v1model?

testdata/p4_16_samples/ipv4_forward-bmv2.stf
```diff
@@ -0,0 +1,8 @@
+set_default ipv4_lpm drop
```
cc10512: before checking these in, we should make sure they are correct. I'm not convinced that the packet data is going to work for this test.

backends/bmv2/JsonObjects.h
```diff
@@ -0,0 +1,82 @@
...
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+#ifndef _BACKENDS_BMV2_JSONOBJECTS_H_
+#define _BACKENDS_BMV2_JSONOBJECTS_H_
+
+namespace bm {
+
```
antoninbas: this header should probably include its dependencies, like `<map>`, etc and forward declare `Util::JsonArray`, etc.

backends/bmv2/action.cpp
```diff
@@ -0,0 +1,191 @@
...
+            auto assign = s->to<IR::AssignmentStatement>();
+            l = assign->left;
+            r = assign->right;
+
+            cstring operation;
+            auto type = typeMap->getType(l, true);
+            if (type->is<IR::Type_StructLike>())
+                operation = "copy_header";
+            else
+                operation = "modify_field";
```
antoninbas: I would recommend using the new bmv2 core primitives here if possible, as they will be available for all architectures: `assign`, `assign_header`, etc (see https://github.com/p4lang/behavioral-model/blob/master/include/bm/bm_sim/core/primitives.h)

backends/bmv2/v1model.cpp
```diff
@@ -0,0 +1,484 @@
...
+            add_size(instance.action_profile.sizeParam.name);
+        } else {
+            add_size(instance.action_selector.sizeParam.name);
+            auto selector = new Util::JsonObject();
+            auto hash = eb->getParameterValue(
+                    instance.action_selector.algorithmParam.name);
+            BUG_CHECK(hash->is<IR::Declaration_ID>(), "%1%: expected a member", hash);
+            auto algo = convertHashAlgorithm(hash->to<IR::Declaration_ID>()->name);
+            selector->emplace("algo", algo);
+            // FIXME selector_check pass
```
antoninbas: why was this commented-out?

backends/bmv2/JsonObjects.h
antoninbas: this header file should probably include its dependencies (`<map>` etc) and fast-forward `Util::JsonArray`...

backends/bmv2/action.cpp
antoninbas: you should try to use the new bmv2 core primitives when possible given that they will be available on all archs: `assign`, `assign_header`... (see https://github.com/p4lang/behavioral-model/blob/master/include/bm/bm_sim/core/primitives.h). For example `assign` is strictly equivalent to the "old" `modify_field`.

hanw: can you point me to the mapping for `copy_header`, `add_header`, `remove_header` and `exit`?

backends/bmv2/v1model.cpp
hanw: good catch, I put it back.

backends/bmv2/action.cpp
antoninbas: there is no mapping (also some new core primitives have no old equivalent), you'll have to go by the names
for `exit`, you should keep using `exit`, the name is staying the same

backends/bmv2/JsonObjects.h
antoninbas: by the way this is important to people not using a unified build

backends/bmv2/midend.h
```diff
@@ -23,22 +23,22 @@ limitations under the License.
...
 #include "midend/inlining.h"
 #include "midend/convertEnums.h"
+#include "helpers.h"
 
 namespace BMV2 {
 
 class MidEnd : public PassManager {
     P4::InlineWorkList controlsToInline;
     P4::ActionsInlineList actionsToInline;
     cstring updateControlBlockName;
```
mbudiu-vmw: Is updateControlBlockName still used?

backends/bmv2/parser.cpp
```diff
@@ -0,0 +1,326 @@
...
+            }
+        }
+    }
+    ::error("%1%: not supported in parser on this target", stat);
+    return result;
+}
+
+// Operates on a select keyset
+void Parser::convertSimpleKey(const IR::Expression* keySet,
+                                     mpz_class& value, mpz_class& mask) const {
```
mbudiu-vmw: indentation

backends/bmv2/parser.cpp
```diff
@@ -0,0 +1,326 @@
...
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+#include "parser.h"
+
+namespace BMV2 {
+
+// TODO(hanw) refactor this function
+Util::IJson* Parser::convertParserStatement(const IR::StatOrDecl* stat) {
```
mbudiu-vmw: I would suggest using a different name for this class, perhaps ParserConverter.

backends/bmv2/parser.cpp
```diff
@@ -0,0 +1,326 @@
...
+Parser::createDefaultTransition() {
+    auto trans = new Util::JsonObject();
+    trans->emplace("value", "default");
+    trans->emplace("mask", Util::JsonValue::null);
+    trans->emplace("next_state", Util::JsonValue::null);
+    return trans;
+}
+
+bool Parser::preorder(const IR::P4Parser* parser) {
+    // TODO(hanw): hard-coded parser name
```
mbudiu-vmw: That is what BMv2 expects.

backends/bmv2/sharedActionSelectorCheck.h
```diff
@@ -0,0 +1,71 @@
...
+#include "frontends/p4/coreLibrary.h"
+#include "frontends/p4/frontend.h"
+#include "frontends/p4/typeMap.h"
+#include "frontends/p4/typeChecking/typeChecker.h"
+#include "frontends/common/resolveReferences/referenceMap.h"
+#include "helpers.h"
+
+namespace BMV2 {
+
+using Input = std::vector<const IR::Expression *>;
```
mbudiu-vmw: Input is too vague; perhaps SelectorInput?

backends/bmv2/v1model.cpp
```diff
@@ -0,0 +1,484 @@
...
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+/**
+ * This file implements the simple switch model
```
mbudiu-vmw: I think this file should be called simple_switch.
And the namespace P4V1 is inappropriate, since this can handle very well P4-16 programs.

frontends/p4/fromv1.0/v1model.h
```diff
@@ -20,7 +20,14 @@ limitations under the License.
 #include "lib/cstring.h"
 #include "frontends/common/model.h"
 #include "frontends/p4/coreLibrary.h"
+#include "frontends/p4/methodInstance.h"
 #include "ir/ir.h"
+#include "lib/json.h"
+#include "backends/bmv2/sharedActionSelectorCheck.h"
+
+namespace BMV2 {
+class Backend;
```
mbudiu-vmw: Please, do not mix the P4-14 model with the back-end. They are completely separate.

frontends/p4/fromv1.0/v1model.h
```diff
@@ -298,6 +306,20 @@ class V1Model : public ::Model::Model {
     DirectMeter_Model   directMeter;
     DirectCounter_Model directCounter;
 
+    static void convertExternObjects(Util::JsonArray *o, BMV2::Backend *bmv2,
```
mbudiu-vmw: These methods do not belong to this class, they are specific to the BMv2 simulator.
V1model is about a specific architecture, which could be compiled to run on many other targets.

backends/bmv2/v1model.cpp
```diff
@@ -0,0 +1,484 @@
...
+    else if (algorithm == V1Model::instance.algorithm.random.name)
+        result = "random";
+    else if (algorithm == V1Model::instance.algorithm.identity.name)
+        result = "identity";
+    else
+        ::error("%1%: unexpected algorithm", algorithm);
+    return result;
+}
+
+void V1Model::convertExternObjects(Util::JsonArray *result, BMV2::Backend *bmv2,
```
mbudiu-vmw: See comment in v1model.h; these methods do not belong to this class.

frontends/p4/fromv1.0/v1model.h
mbudiu-vmw: You cannot have references to a back-end in the front-end!

frontends/p4/inferArchitecture.cpp
```diff
@@ -0,0 +1,114 @@
...
+/// FIXME: going from param to p4type takes a lot of work, should be as simple as
+/// p4type = p4Type(param);
+/// if (p4type->is<IR::Type_Parser>) { visit(p4type); }
+/// else if (p4type->is<IR::Type_Control>) { visit(p4type); }
+bool InferArchitecture::preorder(const IR::Type_Package *node) {
+    for (auto param : node->constructorParams->parameters) {
+        BUG_CHECK(param->type->is<IR::Type_Specialized>(),
+                  "Unexpected Package param type");
+        auto baseType = param->type->to<IR::Type_Specialized>()->baseType;
+        auto typeObj = typeMap->getType(baseType)->getP4Type();
```
mbudiu-vmw: getP4Type is really there just for generating back P4 source code.
You are also mixing getting the type from the parameter and from the typeMap; use just the typeMap.
`typeMap->getType(param, true)` will give you the parameter type.

frontends/p4/inferArchitecture.h
```diff
@@ -0,0 +1,46 @@
...
+#define _P4_INFER_ARCHITECTURE_H_
+
+#include "ir/ir.h"
+#include "ir/visitor.h"
+#include "typeMap.h"
+#include "common/model.h"
+
+namespace P4 {
+
+class InferArchitecture : public Inspector {
```
mbudiu-vmw: I don't really understand what this visitor is doing.
I am assuming it is trying to extract information from the block model.
This is also working only just for V2Model, so the class should be perhaps called InferV2ModelArchitectures, or Extract..., since it does not really infer anything.

frontends/p4/modelInstances.cpp
```diff
@@ -16,10 +16,12 @@ limitations under the License.
 
 #include "coreLibrary.h"
 #include "fromv1.0/v1model.h"
+#include "common/model.h"
 
 namespace P4 {
 
 P4CoreLibrary P4CoreLibrary::instance;
+V2Model V2Model::instance;
```
mbudiu-vmw: You cannot have the front-end depend on any model.
V1Model is there just for P4-14 compatibility.

lib/assert.h
```diff
@@ -0,0 +1,30 @@
+/*
```
mbudiu-vmw: Why is this file called assert.h?
I think that this file is unneeded, and should be removed.
In any case, it has to be renamed.

midend/Makefile.am
```diff
@@ -23,6 +23,7 @@ midend_UNIFIED = \
 	midend/nestedStructs.cpp \
 	midend/inlining.cpp \
 	midend/interpreter.cpp \
+	midend/isolateMethodCalls.cpp \
```
mbudiu-vmw: This pass cannot be part of the mid-end, it has to be in the back-end if anywhere at all.

p4include/p4d2model.p4
```diff
@@ -0,0 +1,224 @@
...
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+/* P4-16 declaration of the P4-D2 (P4 Developer's Day) switch model */
+
+#include "core.p4"
```
mbudiu-vmw: This should be `#include <core.p4>`

p4include/v1model.p4
```diff
@@ -38,6 +38,18 @@ struct standard_metadata_t {
     bit<1>  drop;
     bit<16> recirculate_port;
     bit<32> packet_length;
+    // flattening fields that exist in bmv2-ss
+    // queueing metadata
+    @alias("queueing_metadata.enq_timestamp") bit<32> enq_timestamp;
```
mbudiu-vmw: The `@alias` annotation is not documented. What does it mean?

p4include/v1model.p4
```diff
@@ -133,14 +145,17 @@ parser Parser<H, M>(packet_in b,
                     inout standard_metadata_t standard_metadata);
 control VerifyChecksum<H, M>(in H hdr,
                              inout M meta);
+@pipeline
```
mbudiu-vmw: The @pipeline annotation is not documented anywhere. I think it should be removed.
You are essentially duplicating functionality provided by the package operation - the fact that something is a pipeline is given by the fact that it is used as an argument to the V1Switch package in the right position.

p4include/psa.p4
```diff
@@ -0,0 +1,582 @@
...
+
+/**
+ *   P4-16 declaration of the Portable Switch Architecture
+ */
+
+/**
+ * These types need to be defined before including the architecture file
+ * and the macro protecting them should be defined.
+ */
+#ifndef PSA_CORE_TYPES
```
mbudiu-vmw: Is this file the most recent version?
It looks like this file cannot compile.

p4include/psa.p4
```diff
@@ -0,0 +1,582 @@
...
+#error "Please define the core types for PSA and the PSA_CORE_TYPES macro"
+#endif
+
+#include "core.p4"
+
+/* --------------------------------------------------------------------- *
+/**
+ * PSA supported metadata types
+ */
+struct psa_parser_input_metadata_t {
```
mbudiu-vmw: indentation is wrong.

hanw:
Thanks very much for the review, Mihai. I will address your comments today.

p4include/v1model.p4
hanw: The @pipeline annotation is needed because BMv2 JSON mandates only ingress and egress can be in the `pipeline` section. The original BMv2 backend hard-coded the name of ingress and egress in v1model when generating the `pipeline` section in JSON. When I wrote the visitor, I did not figure out a way to traverse only `ingress` and `egress`, and not `verifychecksum` and `updatechecksum` without hard-coding the names. They are all `Controlblock` from the visitor's point of view.

p4include/psa.p4
hanw: We haven't fully tested psa.p4 yet. I will do it in the next PR.

p4include/v1model.p4
hanw: @alias is used to generate the `field_alias` section of the BMV2 JSON.  Field alias creates a mapping from the metadata name in P4 program to the behavioral model's internal metadata name. Here we use it to expose all metadata supported by behavior model to user through standard metadata, as requested by @vgurevich

p4include/v1model.p4
mbudiu-vmw: The annotation is on the type, not on the instance, so this way of doing it won't work.
For example, you could very well have multiple blocks with the same type required in the package, e.g., the same signature for checksum updates and for the main pipeline. 

The way you distinguish them is by the argument of package they bind to. That's why there is a pass which takes a policy about which controls to skip. What you have to do is something similar for the P2D2 architecture: a list of controls to skip.

p4include/psa.p4
mbudiu-vmw: Maybe you can then remove it from the PR?

midend/Makefile.am
hanw: This pass is disabled as of now, I will fix it in next PR.

p4include/v1model.p4
mbudiu-vmw: You have to add this description to the v1model.p4 file itself.

midend/isolateMethodCalls.h
```diff
@@ -0,0 +1,81 @@
...
+#include "frontends/p4/typeMap.h"
+#include "frontends/p4/coreLibrary.h"
+#include "frontends/common/resolveReferences/referenceMap.h"
+#include "frontends/p4/typeChecking/typeChecker.h"
+
+
+namespace P4 {
+
+/**
+ * Converts non-void extern functions/methods to void
```
hanw: I will remove this file from this PR, and add it back in next PR.

backends/bmv2/control.cpp
```diff
@@ -0,0 +1,769 @@
...
+            tables->append(j);
+        } else if (node->is<CFG::IfNode>()) {
+            auto j = convertIf(node->to<CFG::IfNode>(), cont->name);
+            if (::errorCount() > 0)
+                return false;
+            conditionals->append(j);
+        }
+    }
+
+#ifndef PSA
```
hanw: I removed code enclosed by ifdef PSA, and will add them back in next PR.

mbudiu-vmw: You can temporarily have #ifdefs in the code until you debug or fix something, but don't plan to have them permanently.

hanw: Yes, they are temporary. I will replace it with a flag from compiler options.

p4include/v1model.p4
hanw: Previously, the skipControls policy receives the list of control blocks to skip from the hard-coded names in v1model. https://github.com/p4lang/p4c/blob/b25ba2dc23d147a9efdcc3ed39c49e72fc84ec34/backends/bmv2/midend.cpp#L126  I was looking for a way to generalize how skipControls (and other policies) computes the list of control blocks. The solution I came up with was to use special annotations in the architecture file. That way, at least I don't have to change the backend code for a different architecture. By putting the annotation on the type, only the person who designs the architecture needs to care about this detail, and it's not visible to the P4 programmer. It also affects the checksum passes, as I need a way to identify which control block is updateChecksum. It would be argument 5 in v1model or argument n ( n != 5) in PSA, It may not have the name `updateChecksum` in PSA, or other architecture that people come up with. I am not sure how to identify updateChecksum without using annotation?

mbudiu-vmw: I understand what you are trying to do.
The problem comes from the fact that you are trying to support multiple architectures with a single back-end. That will be harder and harder as these architecture diverge in behaviors. For example, changes you make for PSA may break v1model programs. Also, the BMv2 back-end generates a program called p4c-bm2-ss, so you see that it targets a specific simulator on top of BMv2 (bmv2 can be used to implements lots of different simulators), and it it assumes a specific model for the input architecture, v1model. These things are tightly coupled in the code; the C++ code hardwires a lot of knowledge about v1model and BMv2 SS in particular. You are trying to de-entangle them, but that won't be easy.

However, v1model should be thought as a description of a fixed architecture, which cannot be changed. In general, you cannot assume you can edit it anymore after you have distributed it to customers.

Moreover, regarding the @pipeline annotations, all the information you need IS in the package. It is just annoying to extract. Maybe you can write some helper functions to make this easier. For v1model we know that the pipelines are arguments 3 and 4 of the main package, and the deparser is argument 6 (the code in fact uses the parameter names in v1model to retrieve them). This is a contract between the architecture and the compiler, and it can never change once you published the architecture, so you can rely on it.  Similarly, for PSA you will identify the pipelines from the arguments to the main package.  So what you have to do is write a generic class that can identify the blocks from the main package, provide two implementations for PSA and v1model and output the same data-structure that you can use from both. I would suggest not using annotations, but just set<Node>, the same way it was done in the BMv2 back-end.

PR #606, opened 2017-05-11
Front-end support for header_unions

mbudiu-vmw:
I renamed the IR node to header_union, in case we ever get a real union. Most changes are due to this renaming.
Def-use analysis now handles unions; not many changes were required.
Also, changed the grammar to allow empty list expressions - good for initializing empty structs.

### DIFF
--- a/frontends/p4/def_use.cpp
+++ b/frontends/p4/def_use.cpp
@@ -23,7 +23,7 @@ limitations under the License.
 
 namespace P4 {
 
-// name for header valid bit
+// internal name for header valid bit; used only locally
 const cstring StorageFactory::validFieldName = "$valid";
 const cstring StorageFactory::indexFieldName = "$lastIndex";
 const LocationSet* LocationSet::empty = new LocationSet();
@@ -46,8 +46,21 @@ StorageLocation* StorageFactory::create(const IR::Type* type, cstring name) cons
         type = typeMap->getTypeType(type, true);  // get the canonical version
         auto st = type->to<IR::Type_StructLike>();
         auto result = new StructLocation(type, name);
+
+        // For header unions we will model all of the valid fields
+        // for all components as a single shared field.  The
+        // reason is that updating one of may change all of the
+        // other ones.
+        StorageLocation* globalValid = nullptr;
+        if (type->is<IR::Type_HeaderUnion>())
+            globalValid = create(IR::Type_Boolean::get(), name + "." + validFieldName);
+
         for (auto f : st->fields) {
-            auto sl = create(f->type, name + "." + f->name);
+            cstring fieldName = name + "." + f->name;
+            auto sl = create(f->type, fieldName);
+            if (globalValid != nullptr)
+                dynamic_cast<StructLocation*>(sl)->replaceField(
+                    fieldName + "." + validFieldName, globalValid);
             result->addField(f->name, sl);
         }
         if (st->is<IR::Type_Header>()) {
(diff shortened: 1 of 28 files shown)


frontends/p4/def_use.cpp
```diff
@@ -46,8 +46,21 @@ StorageLocation* StorageFactory::create(const IR::Type* type, cstring name) cons
...
+        StorageLocation* globalValid = nullptr;
+        if (type->is<IR::Type_HeaderUnion>())
+            globalValid = create(IR::Type_Boolean::get(), name + "." + validFieldName);
+
         for (auto f : st->fields) {
-            auto sl = create(f->type, name + "." + f->name);
+            cstring fieldName = name + "." + f->name;
+            auto sl = create(f->type, fieldName);
+            if (globalValid != nullptr)
+                dynamic_cast<StructLocation*>(sl)->replaceField(
```
jnfoster: Is there a reason that `dynamic_cast` is used here instead of a `to` method?

frontends/p4/def_use.cpp
```diff
@@ -46,8 +46,21 @@ StorageLocation* StorageFactory::create(const IR::Type* type, cstring name) cons
         type = typeMap->getTypeType(type, true);  // get the canonical version
         auto st = type->to<IR::Type_StructLike>();
         auto result = new StructLocation(type, name);
+
+        // For header unions we will model all of the valid fields
+        // for all components as a single shared field.  The
+        // reason is that updating one of may change all of the
+        // other ones.
```
jnfoster: Expanding this overview to what is implemented below would be helpful. (Could be done later.)

frontends/p4/def_use.cpp
ChrisDodd: I would guess because the `to` methods currently all return `const` pointers.  We should probably overload them to return non-const pointers when called on non-const objects.

frontends/p4/def_use.cpp
mbudiu-vmw: This is the single call of this method.
This entire pass should be rewritten, so let's not clean it up.
There is an issue filed about that: #151.

frontends/p4/def_use.cpp
mbudiu-vmw: This is the single call of this method.
We have to rewrite this pass at some point: #151.

frontends/p4/def_use.h
```diff
@@ -49,38 +49,44 @@ class StorageLocation : public IHasDbPrint {
...
+    // We can use this for tuples because tuples have no field accessors,
+    // so we treat them as monolithic objects.
     BaseLocation(const IR::Type* type, cstring name) :
             StorageLocation(type, name)
-    { BUG_CHECK(type->is<IR::Type_Base>() || type->is<IR::Type_Enum>()
-                || type->is<IR::Type_Var>() || type->is<IR::Type_Tuple>()
-                || type-is<IR::Type_Error>() || type->is<IR::Type_Var>(),
+    { BUG_CHECK(type->is<IR::Type_Bits>() || type->is<IR::Type_Enum>() ||
+                type->is<IR::Type_Boolean>() || type->is<IR::Type_Var>() ||
+                type->is<IR::Type_Tuple>() || type-is<IR::Type_Error>(),
```
ChrisDodd: How is this not a syntax error  (missing `>` in `type-is<`...)

mbudiu-vmw: Isn't C++ beautiful? 'is' is a method on 'this', and this is just pointer arithmetic!
Sick, I will push a fix. Note that the bug has been there for a longer time.
And I was counting on the C++ type-checker...

jnfoster: Wow.

Issue #608, opened 2017-05-11
Compiler quietly accepts verify() in the controls


vgurevich:
Hello,

Through my ignorance, I've written the following:
```
error { BadIPv4Checksum }

control MyVerifyChecksum(
    in    my_headers_t   hdr,
    inout my_metadata_t  meta)
{
    Checksum16() ipv4_checksum;
    bit<16>      checksum;
    
    apply {
        if (hdr.ipv4.isValid()) {
            checksum = ipv4_checksum.get(
                {    
                    hdr.ipv4.version,
                    hdr.ipv4.ihl,
                    hdr.ipv4.diffserv,
                    hdr.ipv4.totalLen,
                    hdr.ipv4.identification,
                    hdr.ipv4.flags,
                    hdr.ipv4.fragOffset,
                    hdr.ipv4.ttl,
                    hdr.ipv4.protocol,
                    hdr.ipv4.srcAddr,
                    hdr.ipv4.dstAddr
                });
            verify(hdr.ipv4.hdrChecksum == checksum, error.BadIPv4Checksum);
        }
    }
}
```

The compiler happily accepted the code and generated a JSON. 

The right thing to do would be to reject the program, since verify() can appear in parser states only as per spec.

jnfoster:
Good catch. I pushed a possible fix.

jnfoster:
Fixed by #611

PR #611, opened 2017-05-11
Check that verify is only used in parsers.

jnfoster:
Fix for issue #608

### DIFF
--- a/frontends/p4/validateParsedProgram.cpp
+++ b/frontends/p4/validateParsedProgram.cpp
@@ -48,18 +48,6 @@ void ValidateParsedProgram::postorder(const IR::Method* m) {
     }
 }
 
-/// Check that verify is not used outside of parsers
-void ValidateParsedProgram::postorder(const IR::MethodCallExpression* e) {
-    if (e->method->is<IR::PathExpression>()) {
-        auto m = e->method->to<IR::PathExpression>();
-        if (m->path->asString() == IR::ParserState::verify) {
-            auto inParser = findContext<IR::P4Parser>();
-            if (inParser == nullptr)
-                ::error("%1%: may only be invoked in parsers", m);
-        }
-    }
-}
-
 /// Struct field names cannot be underscore
 void ValidateParsedProgram::postorder(const IR::StructField* f) {
     if (f->name.isDontCare())
--- a/frontends/p4/validateParsedProgram.cpp
+++ b/frontends/p4/validateParsedProgram.cpp
@@ -55,11 +55,11 @@ void ValidateParsedProgram::postorder(const IR::MethodCallExpression* e) {
         if (m->path->asString() == IR::ParserState::verify) {
             auto inParser = findContext<IR::P4Parser>();
             if (inParser == nullptr)
-                ::error("%1%: may only be invoked in parsers", m); 
+                ::error("%1%: may only be invoked in parsers", m);
         }
     }
 }
-    
+
 /// Struct field names cannot be underscore
 void ValidateParsedProgram::postorder(const IR::StructField* f) {
     if (f->name.isDontCare())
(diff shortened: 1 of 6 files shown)


jnfoster:
@mbudiu-vmw I'm not sure about this fix. Can you have a look?

In particular:
* Is validateParsedProgram the right place to do this check? I think so...

* Am I analyzing the IR correctly? That is, check for a `MethodCallExpression` whose `method` member is a `PathExpression` that is equal to `ParserState::verify` when converted to a `cstring`? Maybe there is a more elegant way to do this.

frontends/p4/validateParsedProgram.cpp
```diff
@@ -48,6 +48,18 @@ void ValidateParsedProgram::postorder(const IR::Method* m) {
     }
 }
 
+/// Check that verify is not used outside of parsers
+void ValidateParsedProgram::postorder(const IR::MethodCallExpression* e) {
```
mbudiu-vmw: This will work with the current language definition, but it would be more robust to use MethodInstance to resolve the method. This check could probably be better placed in `TypeInference::postorder(IR::MethodCallExpression* e)`. There is already a call to MethodInstance, one just needs to add a check for this particular built-in method.

jnfoster: Okay, I'll move it there. Thanks!

jnfoster: Whoops, actually I don't understand.

* The check is already in IR::MethodCallExpression. I extended the visitor to override that case. Are you suggesting moving it elsewhere? Or keeping it there?

* I do see that using `MethodInstance` would streamline the analysis but I don't see any uses in this pass. Which is the call you were referring to (maybe a link to a line number?)

mbudiu-vmw:
I suggested TypeInference, the typeChecker.cpp file.

frontends/p4/validateParsedProgram.cpp
jnfoster: Ahh, I see. Got it.

jnfoster:
Ahh, I see. Got it.

jnfoster:
Sorry, one more question: it's not obvious to me that `verify` is a built-in method according to `methodInstance.cpp`... The resolver seems to only construct a BuiltInMethod object for isValid/setValid/setInvalid and push_front/pop_front.

mbudiu-vmw:
In that case this is actually an ExternFunction.

jnfoster:
@mbudiu-vmw 

1. I can add the following code to TypeInference::postorder(IR::MethodCallExpression)

```
        // Check that verify is only used in parsers.
        if (auto mf = mi->to<ExternFunction>()) {
            if (auto pe = mf->expr->method->to<IR::PathExpression>())
                if (pe->path->asString() == IR::ParserState::verify)
                    if (!findContext<IR::P4Parser>())
                        ::error("%1%: may only be invoked in parsers", mf->expr);
        }
```
However, the `MethodFunction` object is not helping me much -- unless I've missed something, I still have to go unpack the original `MethodCallExpression` to access the name. So I'm not sure this is better. Since `validate` is a built-in function, why not do the check in `validateParsedProgram` instead? It seems simpler and just as robust...

2. Should we also add checks for `packet_in` and `packet_out` to make sure they are only invoked in the parser/deparser? Or are those implemented further downstream in the back-end?

mbudiu-vmw:
ef->method is a pointer to a Method object, which has a name.
Since it's an extern function you know that there is no object, and the name uniquely identifies the function.

One cannot create packet_in/out objects, so they only appear where the architecture allows them.
If the architecture is broken we can't do much in the compiler.

sethfowler:
I haven't read through the recent comments in this issue carefully, but I did want to point out that `frontends/p4/externInstance.h` may be relevant here.

jnfoster:
@cc10512 I'm going with Mihai's proposal. Perhaps this does belong earlier, but we are getting some mileage out of the `MethodInstance` class so it is definitely more robust. It can always be revisited later.

PR #617, opened 2017-05-12
replaces the bug with unimplemented when using the error enum in structs

cc10512:
This replaces the BUG (crashing the compiler) with a nicer message
specifying that this is unimplemented for BMV2. This is a temporary window
dressing until we properly retrieve errors from the bmv2 parser. Currently
BMv2 raises an exception on parser errors.

### DIFF
--- a/backends/bmv2/Makefile.am
+++ b/backends/bmv2/Makefile.am
@@ -80,3 +80,6 @@ bmv2tests.mk: $(GENTESTS) $(srcdir)/%reldir%/Makefile.am \
 	      $(srcdir)/testdata/p4_14_samples/switch_*/switch.p4 \
 	      $(srcdir)/testdata/p4_16_samples $(srcdir)/testdata/p4_14_samples
 	@$(GENTESTS) $(srcdir) bmv2 $(srcdir)/backends/bmv2/run-bmv2-test.py $^ >$@
+
+XFAIL_TESTS += \
+	bmv2/testdata/p4_16_samples/error_in_struct-bmv2.p4.test
--- a/backends/bmv2/header.cpp
+++ b/backends/bmv2/header.cpp
@@ -71,6 +71,9 @@ void ConvertHeaders::addTypesAndInstances(const IR::Type_StructLike* type, bool 
                 addHeaderField("scalars", newName, boolWidth, 0);
                 scalars_width += boolWidth;
                 backend->scalarMetadataFields.emplace(f, newName);
+            } else if (ft->is<IR::Type_Error>()) {
+                P4C_UNIMPLEMENTED("The %1% does not support declaring fields of type:\n %2%",
+                                  "BMV2 backend", ft);
             } else {
                 BUG("%1%: Unhandled type for %2%", ft, f);
             }
@@ -301,4 +304,3 @@ bool ConvertHeaders::preorder(const IR::PackageBlock *block) {
 }
 
 }  // namespace BMV2
-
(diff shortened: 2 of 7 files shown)


cc10512:
@mbudiu-vmw: provided your definition of XFAIL -- "The XFAIL denotes a well-known bug which we don't know how to fix, so for now we accept it." I think this falls straight into that category. A BMv2 fix for this issue may require significant work, so I think we should just accept that parser errors are not going to be implemented in the bmv2 backend for a while.

mbudiu-vmw:
This error actually should be fairly easy to fix. In fact, there is a pass which converts error to integers, I don't know why it hasn't done this for the struct field.

This has nothing to do with signalling errors from the parser, it is about the user manipulating explicitly values with type error. v1model simply does not expose errors, and that's not a bug, it is by design, so we can't change that.

cc10512:
which pass is that?

hanw:
errorcode.cpp in bmv2 backend does this.

cc10512:
The ErrorCodesVisitor pass simply takes all the enum fields and assigns them a numeric value (the index in the map). It does not replace error declarations. It is an inspector and not a transformer. I can check the size of that map and generate an appropriately sized integer ... but I'm really not sure what is the intent here: make sure that the program goes through the compiler (I shouldn't think so) or that the error member of the struct (used as metadata) is somehow populated by the parser. I assumed the latter.  @vgurevich? 

In the former case, yeah, I can patch it up so it goes through. For the latter, we need support from bmv2/hw to get the right values.

mbudiu-vmw:
Actually BMv2 has no notion of error that is exposed to the program, so we should be able to do the entire work without any support from BMv2. It is just that there are also many other things to do. I expect that a pass like the ConvertEnums would be sufficient, but it has to be done very late, after implementing verify in the back-end.

antoninbas:
Why don't we simply add a `error parser_error` field to the standard metadata struct in v1model.p4? This way bmv2 simple_switch can simply populate that field with the appropriate value (which would be a 3-line change in the simple_switch code).
BTW , a lot of work has gone into bmv2 to support parser errors and verify statements natively, so I'm not sure what @mbudiu-vmw means by "we should be able to do the entire work without any support from BMv2"

mbudiu-vmw:
It is not there in v1model. We can use it for newer targets.

antoninbas:
I don't see what is the issue with adding a standard metadata fields to v1model.p4, especially one I think is important. A bunch of fields were added just recently: https://github.com/p4lang/p4c/blob/master/p4include/v1model.p4#L43

cc10512:
I agree with @antoninbas here. It will not be possible to do this in the compiler only. For one, verify statements depend on runtime data. Adding a parse_error field to standard_metadata is actually the right solution.

mbudiu-vmw:
We cannot change the behavior of p4-14 programs. They did not have a notion of error, and we cannot introduce one. V1model is the p4-14 fixed architecture, in principle it cannot be changed, except fixing bugs. We can do these changes for psa.

antoninbas:
How does that change the behavior of any P4_14 programs? A field is added to the standard_metadata struct in v1model.p4. v1model.p4 is P4_16, so P4_14 programs cannot use it. When translating a P4_14 program to P4_16, this field will never be used. There is 0 impact on packet forwarding behavior as far as I can see.

cc10512:
@mbudiu-vmw note that this is orthogonal to transforming the error enum decl into an int field.

mbudiu-vmw:
It certainly can be done with BMv2 support, I just said that it could also be done without support.
Now if we have the support we'll use it. But I still think that the XFAIL test case does not belong to this PR.

cc10512:
agreed. i'll actually push a change to handle this by converting the enum and will work with @antoninbas to get the metadata handled as well.

mbudiu-vmw:
What's the plan about this PR?

mbudiu-vmw:
I just checked and this error no longer occurs, so I think we can close this pr.

PR #618, opened 2017-05-12
(Part 1) PSA support - addressing review comments

hanw:
Addressed most of the comments from Mihai.

Need more refactor to get rid of P4V1 namespace. Will do in a later PR.


backends/bmv2/control.cpp
```diff
@@ -184,10 +172,10 @@ cstring Control::getKeyMatchType(const IR::KeyElement *ke) {
 }
 
 bool
-Control::handleTableImplementation(const IR::Property* implementation,
-                                                    const IR::Key* key,
-                                                    Util::JsonObject* table,
-                                                    Util::JsonArray* action_profiles) {
+ControlConverter::handleTableImplementation(const IR::Property* implementation,
```
mbudiu-vmw: Indentation still wrong after you renamed the class.

backends/bmv2/extractArchInfo.cpp
```diff
@@ -47,10 +46,10 @@ bool InferArchitecture::preorder(const IR::Type_Control *node) {
     return false;
 }
 
-/// new Parser_Model object
-bool InferArchitecture::preorder(const IR::Type_Parser *node) {
+/// new P4::Parser_Model object
```
mbudiu-vmw: I don't understand this Doxygen comment, it is too short.

backends/bmv2/extractArchInfo.cpp
```diff
@@ -66,19 +65,19 @@ bool InferArchitecture::preorder(const IR::Type_Parser *node) {
 /// p4type = p4Type(param);
```
mbudiu-vmw: This is a strange comment, usually one does not write pseudocode in comments.

backends/bmv2/extractArchInfo.cpp
```diff
@@ -66,19 +65,19 @@ bool InferArchitecture::preorder(const IR::Type_Parser *node) {
...
+bool ExtractArchInfo::preorder(const IR::Type_Package *node) {
     for (auto p : node->getConstructorParameters()->parameters) {
         LOG1("package constructor param: " << p);
         auto ptype = typeMap->getType(p);
         if (ptype == nullptr)
             continue;
         if (ptype->is<IR::P4Parser>()) {
             LOG3("new parser model: " << p->toString());
-            v2model.parsers.push_back(new Parser_Model(p->toString()));
+            v2model.parsers.push_back(new P4::Parser_Model(p->toString()));
```
mbudiu-vmw: It is dangerous to rely on the behavior of toString; maybe you can use instead p->name.

backends/bmv2/extractArchInfo.h
```diff
@@ -14,22 +14,23 @@ See the License for the specific language governing permissions and
...
-#include "common/model.h"
+#include "frontends/p4/typeMap.h"
+#include "frontends/common/model.h"
+#include "frontends/p4/methodInstance.h"
 
-namespace P4 {
+namespace BMV2 {
 
-class InferArchitecture : public Inspector {
+class ExtractArchInfo : public Inspector {
```
mbudiu-vmw: Please document this class.

backends/bmv2/mapAnnotations.h
```diff
@@ -25,31 +25,18 @@ limitations under the License.
 namespace BMV2 {
 
 /**
- * Copy Annotations from architecture block to the corresponding
- * instance in user program.
+ * Build a map from architecture block to its declaration in uer program.
```
mbudiu-vmw: I am hoping that this pass will not be necessary; annotations are a brittle way to do what I think you are doing.
The architecture description language is rich enough to expose all the information you need, you are building another language layer on top of it with annotations, which is not as precise.

hanw:
I am going to work on the passes that deal with checksums today and add an STF test.

PR #620, opened 2017-05-12
Tiny clarification in core.p4 comment for error.NoMatch

jafingerhut:


jafingerhut:
This change was made in core.p4 in the P4_16 draft spec a while back, but didn't get transferred to the version in p4c.

'select' expression is precise, and you can look up what 'select' means in the spec.  "Match expression" that was there isn't very clear what it refers to.

cc10512:
please rebase and I'll merge it for you.

PR #621, opened 2017-05-12
Fix non unified build

antoninbas:
Some users compile this code with their own build system. Recent changes in the bmv2 backend have broken non-unified builds (`max-chunk-size` set to 1 in `bootstrap.sh`).


sethfowler:
I've been contemplating, someday when I have some free time, setting up IWYU on Travis. That would catch issues like these, but perhaps an even bigger benefit is that it would make sure that we remove `#include`'s that we no longer depend on, which would help speed up the build.

antoninbas:
@sethfowler that would be nice to have

PR #626, opened 2017-05-12
Issue430

mbudiu-vmw:
This fix makes the semantics of digest to be eager: the digest is evaluated at the place of the call, and not at the end of the pipeline. This can break P4-14 programs which depended on the old behavior, but we hope that no one really did. P4-16 does not allow such semantics.

The new test does not really exhibit this case, since the conversion is in the back-end, and the generated reference file is in the mid-end, but the test file can be used to manually inspect that the transformation is correct.


backends/bmv2/midend.cpp
```diff
@@ -204,6 +204,8 @@ MidEnd::MidEnd(CompilerOptions& options) {
         new P4::TableHit(&refMap, &typeMap),
         new P4::SynthesizeActions(&refMap, &typeMap, new SkipControls(skipv1controls)),
         new P4::MoveActionsToTables(&refMap, &typeMap),
+        new P4::MidEndLast(),
+        // Here is actually the start of the BMv2-specific back-end
```
hanw: Maybe we can consider moving them to backend.cpp?

mbudiu-vmw: What is "them"?
After this point there are still a few mid-end passes run, but logically what follows is bmv2-specific.

mbudiu-vmw: I don't have any problem moving the invocation of these passes to another file.

PR #627, opened 2017-05-12
remove spurious error meesage

ChrisDodd:


frontends/p4/fromv1.0/programStructure.cpp
```diff
@@ -285,12 +285,6 @@ class ProgramStructure::FixupExtern : public Modifier {
 };
 
 void ProgramStructure::createExterns() {
-    for (auto r : registers) {
```
mbudiu-vmw: Well, the back-end does not need to generate code for direct registers anywhere.
Where are they supported?

ChrisDodd: They're supported by other backends, just not by the bmv2 backend.

mbudiu-vmw: But also they do not seem to be converted to P4-16 data structures anywhere.
Where is that done?

mbudiu-vmw: I don't see where they generate P4-16 code.

mbudiu-vmw: Ok, they generate declarations, but the declarations will be removed as dead code, as they are not referenced anywhere. 

Also, if you remove this error message at least one test will fail.

mbudiu-vmw: Maybe you can consider moving this error message to the BMv2 back-end?

ChrisDodd: Moved the check to the bmv2 backend

Issue #630, opened 2017-05-14
emit() does not support structs


vgurevich:
According to p4_16 specification:

> For derived type, emit recursively proceeds on fields:
> . . .
> If the argument is a struct containing multiple fields, the emit is recursively applied to each component of the struct in the order of their declaration in the struct.

However, the attempt to pass a struct to emit results in the following message:
```
struct my_headers_t {
    ethernet_t     ethernet;
    vlan_tag_t [2] vlan_tag;
    mpls_t [5]     mpls;
    arp_t          arp;
    arp_ipv4_t     arp_ipv4;
    ipv4_t         ipv4;
    icmp_t         icmp;
}

. . . 

control MyDeparser(
    packet_out      packet,
    in my_headers_t hdr)
{
    apply {
        packet.emit(hdr);
   }
}
```

results in the following message:
```
l3_switch.p4(413): error: hdr: emit only supports header and stack arguments, not struct my_headers_t
        packet.emit(hdr);
                    ^^^
l3_switch.p4(110)
struct my_headers_t {
       ^^^^^^^^^^^^
```

mbudiu-vmw:
There is a PR queued with this feature.

mbudiu-vmw:
I lied, the PR is not yet submitted, it is only a commit in my repo, until I figure out some issues with BMv2.
I may split it into two PRs if a solution for this is needed soon.

vgurevich:
It is definitely an extremely useful feature that can relieve a lot of people from the necessity of spending time writing trivial deparsers. Once we allow header_unions in  structs it will be doubly-useful and will also make code a lot safer.

 I wanted to demonstrate it in the training slides, but... So far, I decided to keep the slide, but put a "Not Implemented yet" note :(

Issue #447, opened 2017-04-06
varbit fields are not properly implemented


mbudiu-vmw:
There are several issues with varbit:
- we don't properly convert P4-14 programs with varbit fields into P4-16 programs using v1model
- we do not generate proper code for handling varbit fields on BMv2

antoninbas:
I can look into how to support the 2 argument extract in bmv2. It's not such a big difference with what is supported today. Today in bmv2, the VL field size expression is a property of the VL field; I just need to support providing this expression as part of the extract call instead. The only difficulty is that I need to make sure the JSON stays backward-compatible.

mbudiu-vmw:
You can use another name for the method, like 'extractVarbits' if that helps.
The compiler should take care of that.

That will only be half of the equation, though, we also need to support converting P4-14 correctly, and also assignments between varbit values. I guess that today they can be done using copy_header.

Also, @akeep pointed out checksums; does the checksum work with a field list where a field is a varbit?

antoninbas:
Apart from the parsing, I don't see what else would need to change. Assignments and using VL fields in calculations / checksums should work fine.

antoninbas:
Support for this has been added in bmv2 with https://github.com/p4lang/behavioral-model/pull/330
2 notable things:
- there is a new parser operation, `extract_VL` which takes 2 parameters, the second being an expression used to compute the length
- I introduced the notion of core primitives in bmv2, which are available on all bmv2 targets. These are `assign`, `assign_VL` and `assign_header`. `assign` replaces `modify_field` (which is still available). `assign_VL` implements correct assignment for varbit fields, as per the P4_16 spec (it is important to call `assign_VL` and not assume that bmv2 will work properly when calling `assign` on a varbit field); `assign_VL` does not really perform runtime checks so the compiler is expected to verify that the varbit assignment is conform to the spec. Finally `assign_header` replaces `copy_header` (which is still available) and it works even if the header includes a varbit field.

mbudiu-vmw:
Thank you, this simplifies things. Can you have varbits in metadata?

antoninbas:
It should be possible to use a varbit metadata field as either operand of `assign_VL`; if I'm not mistaken that is the only thing that is allowed by the spec (if you consider that a metadata field cannot be used with extract / emit).

mbudiu-vmw:
[...]

PR #634, opened 2017-05-15
Support emits of structs; test parsing of varbit fields

mbudiu-vmw:
This does not solve issue #447 completely, only partially.
More fixes will be necessary.
It should solve issue #630 completely, though.


midend/expandEmit.cpp
```diff
@@ -0,0 +1,89 @@
...
+
+            auto arg0 = em->expr->arguments->at(0);
+            auto type = typeMap->getType(arg0, true);
+            std::vector<const IR::Expression*> expansion;
+            std::vector<const IR::Type*> expansionTypes;
+            if (expandArg(type, arg0, &expansion, &expansionTypes)) {
+                auto vec = new IR::Vector<IR::StatOrDecl>();
+                auto it = expansionTypes.begin();
+                for (auto e : expansion) {
+                    auto method = statement->methodCall->method->clone();
```
ChrisDodd: Why clone the method if you're not changing it?

mbudiu-vmw: These must be different IR nodes, otherwise all analyses which compute types, def-use, etc. will produce wrong results - the results are usually indexed in a hash-table by IR node.

mbudiu-vmw: A bit worried we may actually need a deeper clone.

ChrisDodd: This probably indicates that we should fix those analyses to work properly in the presence of dags -- just storing pointers to IR nodes is not adequate if it also depends on the context of those nodes.

mbudiu-vmw: It is really easiest to have nodes be distinct rather than look at the context. 

The question is whether these analyses should establish this invariant by themselves. Is setting `visitDagOnce = false` enough for an analysis to duplicate all nodes?

Issue #461, opened 2017-04-09
p4c-bm2-ss gives incorrect warning about unused counter


jafingerhut:
See the first line containing "direct_counter" in the attached P4_16 program.  2017-Apr-08 version of p4c-bm2-ss gives a warning about ipv4_da_lpm_stats being unused.  However, if I comment out that line, I get 2 errors, one for each time ipv4_da_lpm_stats.count() is called shortly afterwards.

Also, if I generate a bmv2 JSON file and run simple_switch on it and try to use the command 'counter_read ipv4_da_lpm_stats <id>' from simple_switch_CLI, it gives this error:

    Error: Invalid counter name (ipv4_da_lpm_stats)

A similar P4_14 program works correctly when compiled with p4c-bmv2.  I can create a separate issue for this behavior if desired.
[bad-warning-of-unused-counter.p4.txt](https://github.com/p4lang/p4c/files/908608/bad-warning-of-unused-counter.p4.txt)

mbudiu-vmw:
This issue was introduced in 2c131b8a65c42c0641bcd4175c51589b271e8bde, when the direct counters had a method count() added. Originally there was no such method in v1model.p4.
I believe that this method should not be present in v1model.p4.

mbudiu-vmw:
See all the comments at #421 where this problem was anticipated.

ChrisDodd:
There's some confusion as to how direct counters and meters work in the P4_16 standard architecture -- are they properties on tables, or are they externs with methods invoked in actions?  The bmv2 backend mostly expecting them to be properties, which is what causes problems with this code.  Interally it is buggy and is checking the wrong attribute name (a result of obfuscating the the names of evering behind static const cstring  variables rather than just using string literals).  With that fixed, the posted code still gives the warning as there are TWO counters, one created in the table and one created in the control.  It should be giving an error about the `count` method being called on a counter not attached to the table, but is not.  Fixing that (and the spurious warning requires changing the attribute in the table to be `counters = ipv4_da_lpm_stats`, attaching the counter created in the control to the table, rather than creating a second counter.

mbudiu-vmw:
The standard architecture is one place, but v1model is another one.
It would be very nice if there was some documentation about the way these objects should be used.
Also, if they are to be used differently between PSA and BMv2 we will need special handling for the two cases in the back-end. Also, if the P4-14 to P4-16 translation targets BMv2 or PSA, it may need to be different!

mbudiu-vmw:
[...]

PR #639, opened 2017-05-16
Fix counters attribute name

ChrisDodd:
Fix (partly) issue #461 

A good example why obfuscating string constants behind `static constexpr` "names" is a bad idea -- makes the code harder to read and maintain and makes it easier for bugs to creep in.

Arguably we should add a check for `count` calls on counters that are not named in the table attributes.  Whether that should be an error or should implicitly add said counter attribute is unclear.

### DIFF
--- a/backends/bmv2/control.cpp
+++ b/backends/bmv2/control.cpp
@@ -401,8 +401,11 @@ ControlConverter::convertTable(const CFG::TableNode* node,
         size = BMV2::TableAttributes::defaultTableSize;
 
     result->emplace("max_size", size);
-    auto ctrs = table->properties->getProperty(BMV2::TableImplementation::directCounterName);
+    auto ctrs = table->properties->getProperty(BMV2::TableAttributes::countersName);
     if (ctrs != nullptr) {
+        // The counters attribute should list the counters of the table, accessed in
+        // actions of the table.  We should be checking that this attribute and the
+        // actions are consistent?
         if (ctrs->value->is<IR::ExpressionValue>()) {
             auto expr = ctrs->value->to<IR::ExpressionValue>()->expression;
             if (expr->is<IR::ConstructorCallExpression>()) {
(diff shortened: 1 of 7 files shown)


mbudiu-vmw:
I will commit soon an example where the extra indirection is actually good: we had changed default_only to defaultonly in the spec, but not in the code.

A question you are raising is whether a table can have simultaneously both counters and direct counters; in that case you may need to use two different properties.

mbudiu-vmw:
Actually, maybe I am wrong, and counters do not need a table property at all.

backends/bmv2/control.cpp
```diff
@@ -400,7 +400,7 @@ ControlConverter::convertTable(const CFG::TableNode* node,
         size = BMV2::TableAttributes::defaultTableSize;
 
     result->emplace("max_size", size);
-    auto ctrs = table->properties->getProperty(BMV2::TableImplementation::directCounterName);
+    auto ctrs = table->properties->getProperty(BMV2::TableAttributes::countersName);
```
mbudiu-vmw: Maybe we can just delete these from TableImplementation?

ChrisDodd:
Arguably we should reorganize this code to not look at the table properties at all, but instead just look at the extern method calls in actions on the table.

mbudiu-vmw:
In this case you will also need to do some more checks, e.g., that counters are not shared between different tables or that method calls only occur in actions.

ChrisDodd:
We need those checks regardless, independent of how we recognize which table(s) a counter or meter is attached to.

backends/bmv2/control.cpp
mbudiu-vmw: This is the only change I am requesting: removal of the unused constants.

ChrisDodd: Are you suggesting replacing them with string constant literals?  Or something else?

mbudiu-vmw:
I don't think we need both TableImplementation::directCounterName and TableATtributes::countersName. One of them should be just deleted.

ChrisDodd:
They're different -- TableImplementation::directCounterName is "direct_counter" and TableAttributes::countersName is "counters" -- the "counters" table attribute refers to a list of externs that must be either "direct_counter" or "counter" instances (TableImplementation::directCounterName or TableImplementation::counterName).

Whether we need the "counters" attribute at all is another question -- or is mentioning the counters (with a method call in an action of the table) enough?  At the moment we check that the method calls and the "counters" attribute are consistent, and give an error if they are not (I think).

mbudiu-vmw:
OK, then I guess what we need is this kind of documentation - that you just typed - in the code.

Issue #638, opened 2017-05-15
default_action does not conform to spec


mbudiu-vmw:
The new version of the spec says that default_action is optional, and if missing it should be NoAction.

PR #640, opened 2017-05-16
Fix for issue #638

mbudiu-vmw:
The spec allows tables without default_actions; in this case we insert a default_action of NoAction very early in the compilation process.

The default_only property was misspelled in the code; fixing this causes lots of reference outputs to change.

### DIFF
--- a/frontends/p4/simplifyDefUse.cpp
+++ b/frontends/p4/simplifyDefUse.cpp
@@ -475,8 +475,10 @@ class FindUninitialized : public Inspector {
     bool preorder(const IR::Operation_Unary* expression) override {
         BUG_CHECK(!lhs, "%1%: Unary operation on LHS?", expression);
         visit(expression->expr);
-        auto storage = getReads(expression->expr, true);
-        reads(expression, storage);
+        // This expression in fact reads the result of the operation,
+        // which is a temporary storage location, which we do not model
+        // in the def-use analysis.
+        reads(expression, LocationSet::empty);
         registerUses(expression);
         return false;
     }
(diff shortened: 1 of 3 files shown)


ChrisDodd:
It seems to me that this is the wrong approach -- the point of the spec change was to simplify things, so we should be removing things that are no longer necessary, not addiing additional stuff:

- remove the code in fromV1 that creates/add NoAction default actions, as it is no longer needed
- remove NoAction from v1model
- remove the checks for not having default actions

That should be all that is needed.  We could think about having a midend pass that added NoAction as default actions, but bmv2 does not need it.

mbudiu-vmw:
We can probably remove these from the v1 path, since this path does the same thing. However, this path is necessary, because this is the semantics of tables, they must have some default action, it is in the spec. Moreover, many analyses rely on the fact that all actions are explicitly present in the actions list, it is much simpler to add NoAction there than to special-case the analyses. See also the discussion about this issue with the spec.

ChrisDodd:
The spec has been changed so that they no longer need a default action.

mbudiu-vmw:
They don't need it in the user program, but the semantics of the table application always requires the existence of default actions, and that is why the compiler synthesises them here. It's simpler for the programmer, but the same for us.

mbudiu-vmw:
I have removed the insertion of default action from the v1->p4-16 path.

This pass implements section 12.2 from the spec, the following paragraph:

The compiler may set the default_action to NoAction (and also insert it into the list of actions) for tables that do not define the default_action property. This is consistent with the semantics given in Section 12.2.1.3.

mbudiu-vmw:
Ooops, I have pushed this change to the wrong branch...

mbudiu-vmw:
This should have been in branch issue648, which was in a PR which has already been merged; I will submit another PR. I will also try to fix this one...

Issue #357, opened 2017-03-10
Range "set" not supported in parser transitions by bmv2 backend


antoninbas:
For example:
```
parser p(packet_in b, out Headers h,
         inout Meta m, inout standard_metadata_t sm) {
    state start {
        b.extract(h.hs.next);
        m.offset = m.offset + h.hs.last.f2;
        transition select(h.hs.last.f1) {
            0..4 : start;
            _: accept;
        }
    }
}
```
Attached program can be used to reproduce.
[range_set.p4.txt](https://github.com/p4lang/p4c/files/834809/range_set.p4.txt)

mbudiu-vmw:
Does BMv2 support this or do I have to synthesize a list of explicit labels?

antoninbas:
You would need to synthesize the list. I don't think it's worth changing the JSON format to support this. If there is a need to optimize the bmv2 parser implementation using set lookups in the future, then we can consider changing the JSON.

mbudiu-vmw:
Here is a discussion of how this should be done: https://github.com/jafingerhut/p4-guide/tree/master/match-range-using-tcam

jafingerhut:
This enhancement has been added with this PR: https://github.com/p4lang/p4c/pull/2125

PR #641, opened 2017-05-16
adds the list of open and currently unsupported issues.

cc10512:


README.md
```diff
@@ -193,3 +193,86 @@ containers used during the `docker build` process. On macOS in particular the
...
+  * Bypass_egress
+  * Sample_ primitives
+  * invalidate
+
+* No support for P4_14 parser exceptions.
+
+## Backends
+
+### Bmv2
+* All checksum verification happens in the parser.
```
mbudiu-vmw: I don't know what this means. Checksum verification has a separate stage in v1model.p4.
Checksum verification/update is never conditional - when it was implemented bmv2 didn't support conditional checksums.

README.md
```diff
@@ -193,3 +193,86 @@ containers used during the `docker build` process. On macOS in particular the
...
+* Range match types in tables
```
mbudiu-vmw: The compiler supports range matches, perhaps BMv2 doesn't.

README.md
```diff
@@ -193,3 +193,86 @@ containers used during the `docker build` process. On macOS in particular the
...
+* Tables with multiple apply calls
```
mbudiu-vmw: The semantics of such calls in P4-14 is not defined.

README.md
```diff
@@ -193,3 +193,86 @@ containers used during the `docker build` process. On macOS in particular the
...
+
+
+## Unsupported P4_16 language features
+
+These are some of the unsupported features we are aware of. We will
```
mbudiu-vmw: Frankly, all unsupported features that we are aware of already have issues filed. So this list is a duplicate.

README.md
```diff
@@ -193,3 +193,86 @@ containers used during the `docker build` process. On macOS in particular the
...
+  * Recirculate variants
+  * Bypass_egress
+  * Sample_ primitives
+  * invalidate
+
+* No support for P4_14 parser exceptions.
+
+## Backends
+
+### Bmv2
```
mbudiu-vmw: Antonin already created such a list in the BMv2 backend readme file.
Maybe we should have only one such list.

README.md
```diff
@@ -193,3 +193,86 @@ containers used during the `docker build` process. On macOS in particular the
...
+
+## Unsupported P4_16 language features
+
+These are some of the unsupported features we are aware of. We will
+update this list as more features get supported in the bmv2 compiler
+backend, as the spec evolves, and as we discover more issues.
+
+- header_union support is incomplete
+
+- nested structs in structs used as block constructor parameters
```
mbudiu-vmw: This is an issue only if using v1model.

README.md
```diff
@@ -193,3 +193,86 @@ containers used during the `docker build` process. On macOS in particular the
...
+};
+
+parser parse(packet_in pkt, out parsed_packet_t hdr,
+             inout s1_t my_metadata,
+             inout standard_metadata_t standard_metadata) {
+  // ...
+}
+```
+
+- explicit transition to reject in parse state
```
mbudiu-vmw: This is only a problem with bmv2.

README.md
```diff
@@ -193,3 +193,86 @@ containers used during the `docker build` process. On macOS in particular the
...
+
+- compound action parameters (can only be `bit<>` or `int<>`)
+
+- functions or methods with a compound return type
```
mbudiu-vmw: This is only a problem with bmv2. The compiler supports this with no problem. But there are no such functions in v1model, so this should not be an issue, unless one wants to provide additional extern blocks.

README.md
```diff
@@ -193,3 +193,86 @@ containers used during the `docker build` process. On macOS in particular the
...
+};
+
+extern s_t my_extern_function();
+
+controlc c() {
+    apply { s_t s1 = my_extern_function(); }
+}
+```
+
+- user-defined extern types / methods which are not defined in `v1model.p4`
```
mbudiu-vmw: To support these we first have to define how the calling conventions in P4 are mapped to BMv2 C++ code.

README.md
antoninbas: supported by bmv2

README.md
antoninbas: I do think it is nice to have an exhaustive list in one place

README.md
antoninbas: Agree with @mbudiu-vmw. Can you remove the list in the bmv2 backend readme if we are going to be using the top-level README? Or maybe replace it with a pointer to this new list?

cc10512: Antonin's list is specific to bmv2. We're trying here to capture what is the delta between what are the gaps in the compiler's bmv2 backend.

mbudiu-vmw: But the two lists overlap. It would be much easier for people to read only one list.

cc10512:
@mbudiu-vmw, I think it is good for someone who downloads the compiler to have a place to check for unsupported features. Even if the compiler frontend can parse all the P4 codes, it doesn't mean that all backends generate code that runs. Yes, there are opened issues for most of the items listed in this document, however, we often get questions on the mailing list asking about one issue or another. Having them in a file that is distributed hopefully will help people quickly understand what are some of the issues.

README.md
antoninbas: My list was specific to the bmv2 _backend_, not bmv2 itself. I am not sure I understand your remark @cc10512. A list of bmv2 limitations would hardly belong in this repo.

mbudiu-vmw:
The point I am trying to make is that we will have to be careful to update the README as issues are fixed or filed.

README.md
jafingerhut: @mbudiu-vmw What do you see lacking in terms of missing semantics in P4-14 for tables with multiple apply calls?  In the absence of direct counters and meters, at least, the desired behavior seems fairly clear to me -- repeat construction of a search key, the match operation, and repeat whatever action is found (which could be a different action than happened in a previous apply on the same table, because the search key contents could be different).

Even in the presence of side effects like counters and meters, repeating those sounds like the straightforward desired thing to do.

mbudiu-vmw: The question is whether it is the same table or another table with the same type (in P4-16 parlance, it is the same instance or another one?)

The spec was not clear originally, I have not looked lately.

README.md
cc10512: Sorry, my bad. I didn't realize that the list was for the bmv2 backend. I will link to that file.

README.md
cc10512: @mbudiu-vmw see this issue: https://github.com/p4lang/p4c/issues/357

README.md
cc10512: This may be a slip on precisely defining the semantics, however, it turns out that when folks call a table multiple times in what they potentially interpret as mutually exclusive paths, they do not go to check the spec, they complain about the compiler ... Since implementing this in bmv2 requires work that we're not currently planning to do, I think it is safer to call it out.

README.md
mbudiu-vmw: Yes the issue is not with tables, it is with select.

PR #643, opened 2017-05-18
Don't include json_loader.h in ir.h

sethfowler:
Here's a PR I've summoned from an old branch.

p4c takes a long time to compile, and one of the biggest factors is duplicate template instantiations in different translation units. We've worked around this to some degree with unified builds, but it'd be much better to resolve the underlying problem.

This patch takes one small step in that direction by eliminating the need to include `json_loader.h` in `ir.h`. This is accomplished by adding forward declarations for `JSONLoader` where needed (including in `ir-generated.cpp`) and moving the code that references `JSONLoader` from `ir-inline.h` into `json_loader.h` itself.

I had tried to do the same for `json_generator.h`, but while I have a patch that does it, it requires some icky explicit template instantiations. Getting rid of it cleanly would require a redesign that didn't implement `toJSON()` as a virtual method.

### DIFF
--- a/backends/p4test/p4test.cpp
+++ b/backends/p4test/p4test.cpp
@@ -15,6 +15,7 @@ limitations under the License.
 */
 
 #include "ir/ir.h"
+#include "ir/json_loader.h"
 #include "lib/log.h"
 #include "lib/error.h"
 #include "lib/exceptions.h"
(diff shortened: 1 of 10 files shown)


sethfowler:
No, it doesn't make much speed difference. It's just a baby step towards the larger goal of minimizing the amount of code that gets `#include`'d in each file.

PR #647, opened 2017-05-21
support for a binary distribution

cc10512:
allow loading of configuration files when no source is specified
Handle both the developer installation and the binary installation.


tools/driver/p4c
```diff
@@ -0,0 +1,37 @@
...
+    # this is the installed configuration
+    artifacts_dir = os.path.join(install_dir, "../share/p4c")
+else:
+    artifacts_dir = install_dir
+
+sys.path.insert(1, artifacts_dir)
+os.environ['P4C_BIN_DIR'] = install_dir
+os.environ['P4C_CFG_PATH'] = os.path.join(artifacts_dir, "p4c_src")
+os.environ['P4C_16_INCLUDE_PATH'] = os.path.join(artifacts_dir, "p4include")
+os.environ['P4C_14_INCLUDE_PATH'] = os.path.join(artifacts_dir, "p4_14include")
```
ChrisDodd: This seems like it expects `p4include` under bindir, which seems wrong -- it should be pkgdatadir

cc10512: artifacts_dir will point to  in the  ```<prefix>/share/p4c``` in the installed version and ```build``` in the developer version, so it is actually in pkgdatadir for the installed version. Unfortunately we do not follow the same structure in the developer build dir.

ChrisDodd: Its going to be `<bindir>/../share/p4`, which is not necessarily the same as `<prefix>/share/p4c`, if someone has tried to configure bindir and datadir to not be in the same relative positions (and we've already had a request to support that for setting up the docker container for p4 tutorials, as docker containers tend to use a different directory structure than normal linux setups?)

hanw: I tried to do `configure --prefix=/usr/local/test`, `make` and then `make install`, everything seems to be installed to the right place.

cc10512: I believe Chris is concerned when one configures with --prefix _and_ --datadir, and not when datadir is derived from prefix.

cc10512:
@ChrisDodd, @hanw: I modified the code to handle the case when the datadir is different from ${prefix}/share.

I am using ```@datarootdir@``` as the variable to be substituted as more recent versions of autoconf print a warning that datadir is no longer the preferred variable: 
```WARNING: 'tools/driver/p4c.in' seems to ignore the --datarootdir setting```

Issue #648, opened 2017-05-22
Slicing a cast crashes in simplifyDefUse


ChrisDodd:
This program:

```
#include <core.p4>
#include <v1model.p4>

header hdr {
    bit<32> a;
    bit<32> b;
    bit<8> c;
}

#include "arith-skeleton.p4"

control ingress(inout Headers h, inout Meta m, inout standard_metadata_t sm) {
    apply {
        h.h.a[7:0] = ((bit<32>)h.h.c)[7:0];
    }
}

V1Switch(p(), vrfy(), ingress(), egress(), update(), deparser()) main;
```

crashes with the error

    In file: ../frontends/p4/simplifyDefUse.cpp:79
    Compiler Bug: no locations known for <Cast>(24117)

PR #649, opened 2017-05-22
Fix for issue #648

mbudiu-vmw:
Every method is now just a preorder, makes the code easier to read.


mbudiu-vmw:
There is a tiny bug, I will submit a fix.

PR #653, opened 2017-05-23
Fix for issue #638

mbudiu-vmw:
By misusing git I lost this code, so I had to rewrite it.
This is a reimplementation of the fix which removes requirement for a default_action.
The compiler creates a default_action of NoAction is none is present; this makes all the dataflow analyses work correctly, since the dataflow analysis for a table is the join of the dataflow for all actions in the actions list. It also sets the stage for cross table code motion - one needs to have all possible actions in order to push code into all branches.


mbudiu-vmw:
The only logical solution seems to be to allow an empty actions list.
Is this the p4-14 spec?

mbudiu-vmw:
Let me then rework this PR by changing the grammar too.
We will need to amend the spec.

mbudiu-vmw:
Maybe I should not have merged... pulled some needless commits.
Anyway, this now allows empty actions lists in both P4-14 and P4-16, there are two test cases.

ChrisDodd:
Should be fine as long as you do a 'squash and merge' rather than 'rebase and merge' to merge the change -- that will get rid of the spurious merges.

Issue #619, opened 2017-05-12
run cpplint with make check?


mbudiu-vmw:
I see many Travis runs fail because of cpplint. Maybe we should consider running it together with make check.

jnfoster:
+1

sethfowler:
It's definitely easy to forget to run it.

mbudiu-vmw:
Anyone volunteer to do this?
Speaking of build, it looks like our process for merging into master is bottlenecked by the need to always merge with the latest and wait for Travis. There must be a better way to do this. We have had approved PRs waiting more than a week to be merged. Also, we don't know if someone else is working on merging them, so there is a race where we trigger needless merges with master.
In general, I have tried to avoid merging my own PRs, but I sometimes step up to do it if they were approved. Maybe we can add a label to the PR, something like "next to be merged"?

jnfoster:
I think we need a merge queue.

antoninbas:
Can't you just disable "Require branches to be up to date before merging" for this repo?
I agree that this has been a major pain, I keep having to click the "Update" button, and then I forget to merge and have to start again 2 hours later...

sethfowler:
It is definitely a pain.

The downside of allowing people to merge un-rebased changes is that they may have been broken by another commit that landed in the meantime. In my experience that happens quite frequently for non-trivial patches.

One approach that is used to solve this problem is to merge PRs to an integration branch (`dev` or something similar) and manually merge commits from that branch to master once we're confident that the tests pass. That is effectively a merge queue.

If we used that approach, someone would need to take responsibility for managing the integration branch. Doing things this way has the effect of reducing the effort expended by most team members at the cost of greatly increasing the amount of work for whoever manages that branch, since they'll have to do things like manually back broken patches out of the integration branch.

I don't think that approach is a good fit for this project, really.

A better idea would be to set up a bot that auto-rebases patches and auto-merges reviewed patches once they pass the tests. People have written a lot of bots that do this, including [bors](https://github.com/graydon/bors), which I've used before on other projects, and [homu](https://github.com/barosl/homu), which looks great, though I haven't used it before.

PR #656, opened 2017-05-24
Fix for issue #619

mbudiu-vmw:
I am not sure this is the best way to do this, but it seems to work.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -154,6 +154,8 @@ cpplint_FILES += $(noinst_HEADERS)
 clean-local:
 	-rm -f $(BUILT_SOURCES) $(CLEANFILES)
 
+check: cpplint
+
 cpplint:
 	@cd $(srcdir); tools/cpplint.py --root=. --extensions=h,hpp,c,cpp,ypp,l $(sort $(cpplint_FILES))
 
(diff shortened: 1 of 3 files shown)


Makefile.am
```diff
@@ -154,6 +154,8 @@ cpplint_FILES += $(noinst_HEADERS)
 clean-local:
 	-rm -f $(BUILT_SOURCES) $(CLEANFILES)
 
+check: cpplint
+
```
ChrisDodd: Is this correct, or should it be check-local?  I'm never sure with automake.  Possibly also redirect stdout of cpplint to avoid the spew of "processing x" messages that are mostly uninteresting?

mbudiu-vmw: stupid cpplint writes the message on stderror.
Want me to change the python file?

sethfowler: Re: `check-local`, I wondered that myself. It does seem more idiomatic, so perhaps that by itself is an argument in its favor.

mbudiu-vmw: Added a --quiet option to cpplint

ChrisDodd: The only difference I could find is that if you use `check-local` and `-j` it will run the cpplint check in parallel with other checks, while if you use `check` it will not (and will bail out if cpplint fails, before running any tests).

PR #657, opened 2017-05-24
Add utilities that make writing GTests easier

sethfowler:
This PR adds a number of facilities that make it easier to write GTests. I'm using these features in other PRs (which haven't landed yet, obviously) and they've been quite helpful. The changes included:

- Helpers for writing simple IR traversals using lambdas. `forAllMatching`, which behaves like a postorder Inspector which only matches a single node type, was already being used in the `p4RuntimeSerializer` code; I've just moved it to `visitor.h` so other code can use it. I also added Modifier and Transform variations of the concept.
- I reworked the Makefile a little bit to make it easier to link additional libraries into the GTest executable. This is necessary to cleanly add GTests in extensions.
- I generalized the types that you can pass to `BUG()` a little bit.
- I added a helper to get a combined total of warnings and errors. This makes it a bit easier to write assertions.
- I added `WARN_CHECK()` and `ERROR_CHECK()`, which behave like `BUG_CHECK()` but for warnings and errors respectively. This makes typical error and warning checks more terse.
- Finally, I added support for the same debugging flags we support in the compiler executables to the GTest executable. I've found this quite handy at times.

### DIFF
--- a/ir/visitor.h
+++ b/ir/visitor.h
@@ -309,4 +309,53 @@ class P4WriteContext : public virtual Visitor {
     // as it might be an 'inout' access or it might be unable to decide.
 };
 
+
+/**
+ * Invoke an inspector @function for every node of type @NodeType in the subtree
+ * rooted at @root. The behavior is the same as a postorder Inspector.
+ */
+template <typename NodeType, typename Func>
+void forAllMatching(const IR::Node* root, Func&& function) {
+    struct NodeVisitor : public Inspector {
+        explicit NodeVisitor(Func&& function) : function(function) { }
+        Func function;
+        void postorder(const NodeType* node) override { function(node); }
+    };
+    root->apply(NodeVisitor(std::forward<Func>(function)));
+}
+
+/**
+ * Invoke a modifier @function for every node of type @NodeType in the subtree
+ * rooted at @root. The behavior is the same as a postorder Modifier.
+ *
+ * @return the root of the new, modified version of the subtree.
+ */
+template <typename NodeType, typename RootType, typename Func>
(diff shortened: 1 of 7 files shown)


sethfowler:
Yeah, they're not used anywhere in this repo yet. I'm using them elsewhere, though.

sethfowler:
Heh, failed on cpplint. What else is new? =)

mbudiu-vmw:
I have a little PR for running cpplint with make check: #656

ir/visitor.h
```diff
@@ -309,4 +309,53 @@ class P4WriteContext : public virtual Visitor {
...
+ * rooted at @root. The behavior is the same as a postorder Modifier.
+ *
+ * @return the root of the new, modified version of the subtree.
+ */
+template <typename NodeType, typename RootType, typename Func>
+const RootType* modifyAllMatching(const RootType* root, Func&& function) {
+    struct NodeVisitor : public Modifier {
+        explicit NodeVisitor(Func&& function) : function(function) { }
+        Func function;
+        bool postorder(NodeType* node) override { function(node); return true; }
```
ChrisDodd: `Modifier::postorder` functions return void, not bool; I'm suprised this doesn't give a compile error?  Maybe its considered a covariant return type as bool > void?

mbudiu-vmw: My guess is that this class has not been used.

sethfowler: No, I've used it. I'm surprised as well, but I think Chris's theory (that the return type change is considered covariant) is probably correct.

At any rate, even if it compiles, it's confusing. I'll fix it.

PR #659, opened 2017-05-24
insert explicit casts in p4_14 typechecking

ChrisDodd:
- avoid redundant casts in P4_14->16 conversion
- fix some corner case P4_14 type inferencing

This new pass would seem like an ideal candidate for a "no automatic clone transform" visitor subclass, as it has to go out of its way to undo the clone when it is unwanted.


frontends/p4-14/typecheck.cpp
```diff
@@ -411,6 +416,34 @@ class TypeCheck::AssignActionArgTypes : public Modifier {
...
+        inAction = true;
+        visit(af->action);
+        inAction = false;
+        prune();
+        return af; }
+    IR::Expression *postorder(IR::Expression *op) override {
+        if (!inAction || isWrite()) return op;  // don't cast lvalues or refs outside actions
+        auto *type = inferTypeFromContext(getContext(), global);
+        if (type != IR::Type::Unknown::get() && !type->is<IR::Type_InfInt>() && type != op->type) {
+                LOG1("Need cast " << op->type << " -> " << type << " on " << op);
```
mbudiu-vmw: I thought you don't like LOG1 for this kind of stuff...

frontends/p4-14/typecheck.cpp
```diff
@@ -411,6 +416,34 @@ class TypeCheck::AssignActionArgTypes : public Modifier {
     explicit AssignActionArgTypes(TypeCheck &s) : self(s) { setName("AssignActionArgTypes"); }
 };
 
+class TypeCheck::InsertImplicitCasts : public Transform, P4WriteContext {
```
mbudiu-vmw: Maybe this should be called InsertExplicitCasts.

frontends/p4-14/typecheck.cpp
```diff
@@ -411,6 +416,34 @@ class TypeCheck::AssignActionArgTypes : public Modifier {
...
+        inAction = false;
+        prune();
+        return af; }
+    IR::Expression *postorder(IR::Expression *op) override {
+        if (!inAction || isWrite()) return op;  // don't cast lvalues or refs outside actions
+        auto *type = inferTypeFromContext(getContext(), global);
+        if (type != IR::Type::Unknown::get() && !type->is<IR::Type_InfInt>() && type != op->type) {
+                LOG1("Need cast " << op->type << " -> " << type << " on " << op);
+                const IR::Expression *e = getOriginal<IR::Expression>();
+                if (*op != *e) e = op;  // undo Transform clone if it wasn't needed
```
mbudiu-vmw: Why is this necessary?

frontends/p4/fromv1.0/programStructure.cpp
```diff
@@ -304,13 +304,9 @@ const IR::Expression* ProgramStructure::paramReference(const IR::Parameter* para
...
     Util::SourceInfo srcInfo, const IR::Expression* left,
     const IR::Expression* right, const IR::Type* type) {
-    const IR::Expression* cast;
-    if (type != nullptr)
-        cast = new IR::Cast(type, right);
-    else
-        cast = right;
-    auto result = new IR::AssignmentStatement(srcInfo, left, cast);
-    return result;
+    if (type != nullptr && type != right->type)
```
mbudiu-vmw: I wish I had thought of this earlier...

frontends/p4-14/typecheck.cpp
```diff
@@ -411,6 +416,34 @@ class TypeCheck::AssignActionArgTypes : public Modifier {
...
+                return new IR::Cast(type, e);
```
mbudiu-vmw: I could not figure out whether any of the test cases was modified by this line. There are too many changes. But I wonder how programs compiled without this previously. Do you have a new test case to exercise this line?

frontends/p4-14/typecheck.cpp
ChrisDodd: or MakeImplicitCastsExplicit?

mbudiu-vmw: Even better.

frontends/p4-14/typecheck.cpp
ChrisDodd: Yes, this should likely be LOG2 at least, though at least it only prints one line for each cast inserted, and doesn't produce any output if no casts are needed.

frontends/p4-14/typecheck.cpp
ChrisDodd: Strictly speaking it is not, as making an extra 1-node clone is pretty minor (the parent will always need cloning to reflect the new cast).  But the idea here is to insert a new (cast) node into the dag without modifying the node being cast if not otherwise needed.

That's why I said this would be a good candidate for a transform that does not autoclone (some discussions I've been having with Cole).

testdata/p4_14_samples_outputs/switch_20160512/switch-first.p4
```diff
@@ -2937,7 +2937,7 @@ control process_egress_acl(inout headers hdr, inout metadata meta, inout standar
     }
     @name(".egress_mirror") action egress_mirror(bit<32> session_id) {
         meta.i2e_metadata.mirror_session_id = (bit<16>)session_id;
-        clone3<tuple<bit<32>, bit<16>>>(CloneType.E2E, session_id, { meta.i2e_metadata.ingress_tstamp, meta.i2e_metadata.mirror_session_id });
+        clone3<tuple<bit<32>, bit<16>>>(CloneType.E2E, (bit<32>)(bit<16>)session_id, { meta.i2e_metadata.ingress_tstamp, meta.i2e_metadata.mirror_session_id });
```
sethfowler: This seems a bit odd...

frontends/p4-14/typecheck.cpp
ChrisDodd: Many test cases get casts inserted (and thus the dumped P4-16 code is different), but it turns out none of the test cases we had had implicit casts that were legal in P4_14 and not in P4_16, which is the one place where this is actually needed (implicit casts that are also legal implicit casts in P4_16 will get inserted by the P4_16 typechecker).

We do have some internal test cases that were hitting a 'type unification failed' error that this change fixes.

testdata/p4_14_samples_outputs/switch_20160512/switch-first.p4
ChrisDodd: That does seem odd -- I'll look into where it is coming from.

ChrisDodd: Turns out the problem is Dag vs tree in the visitor -- because this node has mulitple parents in the tree and those contexts have different types, it needs different casts in the different uses.  The fix is to set `visitDagOnce = false` in the `MakeImplicitCastsExplicit` transform, but this is a subtle issue that is easily missed.

Good catch!

frontends/p4-14/typecheck.cpp
ChrisDodd: After adding `visitDagOnce = false`, this now is strictly necessary to avoid duplicating ActionArg nodes, which we don't want.

PR #664, opened 2017-05-25
Fix undefined behavior in p4parser.ypp

sethfowler:
This PR fixes a (harmless) instance of undefined behavior in `p4parser.ypp`. See the comment in the patch for more details.

### DIFF
--- a/frontends/parsers/p4/p4parser.ypp
+++ b/frontends/parsers/p4/p4parser.ypp
@@ -38,7 +38,26 @@ limitations under the License.
 namespace P4 {
 class P4Lexer;
 class P4ParserDriver;
+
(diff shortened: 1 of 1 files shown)


sethfowler:
I like most languages better than C++. =)

In fairness, in this particular case the root cause is a bug in Bison's variant implementation. They should've just used boost's version instead of reinventing the wheel.

jnfoster:
I'm hearing two votes for rewriting the compiler in OCaml?

jnfoster:
We can do F# if @mbudiu-vmw prefers...

robertsoule:
+1 for OCaml

sethfowler:
Just imagine how much easier our lives would be if p4c was implemented in LabVIEW.

Issue #655, opened 2017-05-23
bmv2 back-end does not generate checksum computations


mbudiu-vmw:
Here is an example program:
```
#include <core.p4>
#include <v1model.p4>

header H {
    bit<16> d;
    bit<16> c;
}

struct Parsed_packet {
    H h;
}

struct Metadata {
}

control DeparserI(packet_out packet, in Parsed_packet hdr) {
    apply {
        packet.emit(hdr.h);
    }
}

parser parserI(packet_in pkt, out Parsed_packet hdr, inout Metadata meta, inout standard_metadata_t stdmeta) {
    state start {
        pkt.extract(hdr.h);
        transition accept;
    }
}

control cIngress(inout Parsed_packet hdr, inout Metadata meta, inout standard_metadata_t stdmeta) {
    apply {
        hdr.h.d = hdr.h.d + 1;
    }
}

control cEgress(inout Parsed_packet hdr, inout Metadata meta, inout standard_metadata_t stdmeta) {
    apply {}
}

control vc(in Parsed_packet hdr, inout Metadata meta) {
    Checksum16() ck;
    apply {
        bit<16> tmp = ck.get({ hdr.h.d, hdr.h.c });
        if (0 != tmp)
            mark_to_drop();
    }
}

control uc(inout Parsed_packet hdr, inout Metadata meta) {
    Checksum16() ck;
    apply {
        bit<16> tmp = ck.get({ hdr.h.d });
        hdr.h.c = tmp;
    }
}

V1Switch(parserI(), vc(), cIngress(), cEgress(), uc(), DeparserI()) main;
```

The checksums computation produced in json is empty.

PR #670, opened 2017-05-26
Fix for issue #655

hanw:


mbudiu-vmw:
The compiler crashes on 5 examples. So the patch is obviously not ready for review.
Also, please add a test case that actually runs a checksum computation. I think that the following would work as an STF file for the example filed with issue #655: 

```
packet 0 0000 0000
expect 0 0001 0001
packet 0 0001 0001
expect 0 0002 0002
```

If this example works you can also address issue #255.

hanw:
yes, I was dragged to a meeting before I can finish adding the test case, doing it now.

mbudiu-vmw:
This commit does not include the reference outputs from the new test case, so we cannot be sure that the test case has succeeded.

hanw:
Still debugging, just switched to the VM with bmv2 installed. I will let you know when it's complete.

hanw:
@mbudiu-vmw The checksum value computed by bmv2 seems odd to me, it is `0xfffe` instead of `0x0001` for an input of `0x0001`.

hanw:
but other than that, I have added the test case for p4test and bmv2.

mbudiu-vmw:
Probably my test is wrong: https://en.wikipedia.org/wiki/IPv4_header_checksum
It looks like you have to flip the bits.
Please update the stf file.

hanw:
I have flipped the bits in the committed stf file.

jafingerhut:
I would strongly recommend at least a few test cases that would detect the difference between doing a 1's complement sum vs. not doing a 1's complement sum, i.e. a case where adding two 16-bit numbers has a carry out into the 17th bit position.

jafingerhut:
Also at least 1 case where the received packet is dropped because of a bad checksum.

PR #672, opened 2017-05-26
Remove lib/default.h

sethfowler:
In C++03, forcing value-initialization of an unknown type (e.g., in a template) was tricky, but in C++11, there's a standard way to do it - uniform initialization. Here's a [blog post](https://akrzemi1.wordpress.com/2013/09/10/value-initialization-with-c/) on the subject, but the tldr is that you can just write `T t{};` and `T` will get value initialized, which means that its default constructor will be called if it's a class type, it'll get initialized to nullptr if it's a pointer type, 0 if it's a numeric type, etc.

Using this feature eliminates the need for `lib/default.h`, so we can remove it and the associated code. I noticed this opportunity for simplification in the process of turning the existing unit tests into GTests.


sethfowler:
Completely agreed.

PR #676, opened 2017-05-27
Fix #674 - avoid printing duplicate header_type entries in bmv2 JSON …

### LINKED ISSUE #674: bmv2 JSON config files can have duplicates in header_types section
switch.p4's bmv2 JSON config file, for example, has several header_types that are repeated 2 or 3 times.  I am not sure, but I think this occurs when more than one header has the same type, e.g. if there is an inner IPv4 header and an outer IPv4 header defined with the same type.

The waste of space in the bmv2 JSON file is not my primary concern here.  In the "headers" section, the "header_type" is given as a name, not an id, so this means that bmv2 code must have a way to look up the header_type by name, not id.  There are other memory corruption issues when there are name collisions in the JSON config file, e.g. https://github.com/p4lang/p4c/issues/486

I do not know whether behavioral-model code can cause memory corruption while reading such files, but why risk it?

jafingerhut:
…config file

### DIFF
--- a/backends/bmv2/JsonObjects.cpp
+++ b/backends/bmv2/JsonObjects.cpp
@@ -119,9 +119,9 @@ JsonObjects::add_meta_info() {
 unsigned
 JsonObjects::add_header_type(const cstring& name, Util::JsonArray*& fields, unsigned max_length) {
     std::string sname(name, name.size());
-    try {
-        return header_type_id.at(sname);
-    } catch (const std::out_of_range &x) {
+    auto header_type_id_it = header_type_id.find(sname);
+    if (header_type_id_it != header_type_id.end()) {
+        return header_type_id_it->second;
     }
     auto header_type = new Util::JsonObject();
     unsigned id = BMV2::nextId("header_types");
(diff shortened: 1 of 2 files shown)


backends/bmv2/JsonObjects.cpp
```diff
@@ -118,8 +118,14 @@ JsonObjects::add_meta_info() {
  */
 unsigned
 JsonObjects::add_header_type(const cstring& name, Util::JsonArray*& fields, unsigned max_length) {
+    std::string sname(name, name.size());
+    try {
+        return header_type_id.at(sname);
+    } catch (const std::out_of_range &x) {
```
cc10512: Using exceptions for control flow is not a good practice. I recommend using the following pattern:
```
auto header_type_id_it = header_type_id.find(sname);
if( header_type_id_it != header_type_id.end() )
  return header_type_id_it->second;
}
```

jafingerhut: @cc10512 Thanks for the sample code!  Obviously my ignorance of STL is vast.

PR #681, opened 2017-05-29
Fix "Argument list too long" make check errors on Linux

ChrisDodd:
As the number of tests increases, we're starting to hit a problem on Linux where "make check" fails with "Argument list too long" due to the way in which it recursively invokes make.  This change avoids the problem, and doesn't appear to introduce other problems or break parallel makes or recheck or other things.

It still has the problem that it needs to rebuild everything if you touch any Makefile.am file, which is annoying but not fatal.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -149,8 +149,6 @@ cpplint_FILES += $(noinst_HEADERS)
 clean-local:
 	-rm -f $(BUILT_SOURCES) $(CLEANFILES)
 
-check-local: cpplint-quiet
-
 cpplint:
 	@cd $(srcdir); tools/cpplint.py --root=. --extensions=h,hpp,c,cpp,ypp,l $(sort $(cpplint_FILES))
 
@@ -167,6 +165,19 @@ etags:
 	    backends extensions frontends ir lib tools midend
 	cd $(srcdir)/tools/ir-generator; ctags -e -R --langmap=Flex:+.l,YACC:+.ypp \
 	    . ../../lib
+
+# automake's default "make check" rule runs into "Argument list too long" errors on
+# Linux when TESTS gets to big, because of the peculiar way in which it recursively
+# invokes make with assignments on the command line.  These in turn end up in the
+# MAKEFLAGS environment variable which causes the problem when invoking bash, as
+# it exceeds Linux's 128K limit on a single environment variable.  So we override
+# with a simpler recipe that doesn't do that.  This gives us warnings from make
+# about multiple recipes for 'check-am', but the right one (this one) ends up being used
+check-am: all-am cpplint-quiet $(check_LTLIBRARIES) $(check_PROGRAMS)
+	@rm -f $(TEST_LOGS)
+	@rm -f $(TEST_LOGS:.log=.trs)
+	@rm -f $(TEST_SUITE_LOG)
+	@$(MAKE) $(TEST_SUITE_LOG)
 
 # XXX(seth): automake as of version 1.15 has a bug that makes "recheck" interact
 # very badly with parallel make.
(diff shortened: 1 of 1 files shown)


mbudiu-vmw:
I actually read this, but my knowledge of autotools/make is too weak to really understand why it works.
I think that @sethfowler would probably be a better reviewer.

ChrisDodd:
My main concern is that this not break anyones workflow -- I tried out all the various permutations of 'make check' I could think of and they seemed to work ok.

Makefile.am
```diff
@@ -168,6 +167,20 @@ etags:
...
+# invokes make with assignments on the command line.  These in turn end up in the
+# MAKEFLAGS environment variable which causes the problem when invoking bash, as
+# it exceeds Linux's 128K limit on a single environment variable.  So we override
+# with a simpler recipe that doesn't do that.  This gives us warnings from make
+# about multiple recipes for 'check', but the right one (this one) ends up being used
+check: all-am $(BUILT_SOURCES) $(check_LTLIBRARIES) $(check_PROGRAMS) cpplint-quiet
+	@rm -f $(TEST_LOGS)
+	@rm -f $(TEST_LOGS:.log=.trs)
+	@rm -f $(TEST_SUITE_LOG)
+	@$(MAKE) $(TEST_LOGS)
```
sethfowler: Doesn't the rule for TEST_SUITE_LOG already depend on TEST_LOGS?

Makefile.am
```diff
@@ -116,7 +116,8 @@ setup.h: $(srcdir)/setup.h.in Makefile
 	     -e 's|@prefix[@]|$(prefix)|g' \
 	     -e 's|@p4includedir[@]|$(p4includedir)|g' \
 	     -e 's|@p4_14includedir[@]|$(p4_14includedir)|g' \
-	$< >$@
+	$< >$@.tmp
+	@if cmp $@ $@.tmp; then rm $@.tmp; else mv $@.tmp $@; fi
```
sethfowler: What situation is this protecting you from? It's a bit tricky to infer but my best guess is that you're running into a problem with building `$(BUILT_SOURCES)` at the same time as `all-am` in the new `check` rule. If so, I think the right solution here is just to not do that; see below.

Makefile.am
```diff
@@ -168,6 +167,20 @@ etags:
...
 	    . ../../lib
 
+# automake's default "make check" rule runs into "Argument list too long" errors on
+# Linux when TESTS gets to big, because of the peculiar way in which it recursively
+# invokes make with assignments on the command line.  These in turn end up in the
+# MAKEFLAGS environment variable which causes the problem when invoking bash, as
+# it exceeds Linux's 128K limit on a single environment variable.  So we override
+# with a simpler recipe that doesn't do that.  This gives us warnings from make
+# about multiple recipes for 'check', but the right one (this one) ends up being used
+check: all-am $(BUILT_SOURCES) $(check_LTLIBRARIES) $(check_PROGRAMS) cpplint-quiet
```
sethfowler: The contract of `$(BUILT_SOURCES)` is that it should be built before everything else is; to do otherwise will break automake's dependency tracking. Given that, it'd be best to restructure this so that we don't build `$(BUILT_SOURCES)` and these other dependencies in parallel. The original automake rules have a rule that depends only on `$(BUILT_SOURCES)` which invokes a new instance of make that runs a second rule that does the real work; it may be best to just imitate that pattern.

Makefile.am
ChrisDodd: This is an attempt to not rebuild everything all the time -- all object files depend on setup.h, and setup.h depends on Makefile, so by not updating setup.h when there are no actual differences, we don't need to rebuild everything whenever Makefile changes.  Unfortunately there still seem to be dependencies on Makefile.am somewhere that rebuild everything when Makefile.am changes.

sethfowler: So it's unrelated to the other changes in this patch?

sethfowler: If you don't update setup.h then this command will run every time, by the way - i.e., setup.h will remain perpetually out of date. Given that all that happens for this target is a call to `sed`, that's not so bad, but it's worth keeping in mind.

ChrisDodd: The command is fast and silent, so running it on every make is not a big deal (there are other much slower things that happen on every make).  Its here largely because it makes trying variations on this patch (by editing Makefile) much easier/faster.

Makefile.am
ChrisDodd: So I cleaned this up by changing it to override check-am instead of check (the default check just depends on `$(BUILT_SOURCES)`, then does a `make check-am`.)  Hopefully that won't break horribly on other versions of automake.

Makefile.am
sethfowler: OK. I'd recommend either moving it into a separate commit (so `git blame` explains the reasoning for the change) or adding a comment (so viewing the source explains the reasoning). Future readers of this code will struggle otherwise because Makefile interactions are nonlocal and are hard to reason about just from a diff.

ChrisDodd: ok, removed from this change

sethfowler: That's another solution, but just to be clear: I wasn't saying you had to remove it, just that it should be documented somewhere. =)

ChrisDodd: I have it in another change where I'm trying to find the other causes of rebuilding everything when Makefile.am changes.

Issue #96, opened 2016-09-29
Testing bug: no negative tests for p4c-bm2-ss


mbudiu-vmw:
The bmv2 backend testing script does not really check the results of negative tests for bmv2.

mbudiu-vmw:
It looks like we will be using XFAIL for this purpose, so I am closing this issue.

PR #683, opened 2017-05-30
Fix bmv2 inspector which validates selector sharing

### LINKED ISSUE #679: p4c-bm2-ss fails to diagnose all cases of different selector inputs to the same action selector
The P4_16 file [keys.p4.txt](https://github.com/p4lang/p4c/files/1033899/keys.p4.txt)  contains two tables that share an action selector.  However, the selector inputs for these two tables are not the same:

    table tbl1 {
        key = {
            meta.a: selector;
            meta.b: selector;
        }
        implementation = as;
        ...
    }

    table tbl2 {
        key = {  // Note: order reversed from tbl1
            meta.b: selector;
            meta.a: selector;
        }
        implementation = as;
        ....
    }

But p4c-bm2-ss is perfectly happy with this invalid situation, producing only spurious warnings about a missing default_action property which is not supported on these tables anyway.

This same lack of diagnosis happens if you REMOVE some of the selector keys from the FIRST table.

Looking at p4c/backends/bmv2/sharedActionSelectorCheck.cpp lines 66 through 75 will show the problem.  The code there is completely incapable of determining whether or not the selector inputs consist of the same key expressions in the same order.  Surely the code ought to read something like:

        uint n = i1.size();
        if (n != i2.size()) return false;
        for (uint k = 0;  k < n;  k++) {
            if (!checkSameKeyExpr(i1[k], i2[k]))  return false;
        }
        return true;

antoninbas:
Selector input fields should appear in the same order in all tables.

fixes #679

### DIFF
--- a/backends/bmv2/sharedActionSelectorCheck.cpp
+++ b/backends/bmv2/sharedActionSelectorCheck.cpp
@@ -16,6 +16,8 @@ limitations under the License.
 
 #include "sharedActionSelectorCheck.h"
 
+#include <algorithm>
+
 namespace BMV2 {
 
 const SelectorInput*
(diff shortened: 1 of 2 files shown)


antoninbas:
@mbudiu-vmw that's exactly what I thought (p4test vs p4c-bm2-ss). I have a negative test ready though, so I can add it to the PR.

antoninbas:
@mbudiu-vmw I added the negative test

mbudiu-vmw:
Great, when we fix #96 hopefully these tests will be run as part of our testsuite.

PR #688, opened 2017-05-31
Fix circular dep in includes

antoninbas:


antoninbas:
@teverman FYI

antoninbas:
@teverman I updated my commit to add a fwd declaration for Backend and move the constructor to the cpp file. What I'm worried about is that I don't seem to be catching these errors when I run my "non-unified" build

sethfowler:
Not sure why you wouldn't...

antoninbas:
can I get an approve for this?

PR #691, opened 2017-06-01
p4c-bm2-ss should gracifully exit when compiling a program without main

hanw:


mbudiu-vmw:
Can you please add a negative test?
When someone fixes #96 all these tests should become part of our test suite...

PR #695, opened 2017-06-02
Reduce unnecessary rebuilds

sethfowler:
Chris got me interested in the problem of reducing the amount of unnecessary rebuilds we do. (And I hope this PR hasn't duplicated his work; I know he's been looking at it too!)

The class of problems I'm trying to solve here are "build amplification" problems: cases where a change to a single file leads us to rebuild most or all of the project for no good reason.

With that in mind, this PR makes the following changes:

- `setup.h` is reimplemented using CPP symbols, which eliminates it as a concern from the perspective of the build system.

- I reused `ylwrap` (which ships with autotools and is normally used with lex and yacc) to eliminate build amplification due to the IR generator. It was easy to change something in `lib/` and cause the whole project to rebuild because the IR generator depends on `libp4toolkit`. Using `ylwrap` avoids the problem because `ir-generated.h` and `gen-tree-macro.h` aren't updated unless their contents would actually change. A minor hack was required to work with `ylwrap`'s unnecessary limitations, but hey, it's autotools.

- While I was at it, IR generator invocations now respect automake's verbosity option for improved aesthetics.

- We no longer reinstall package data in all-local unless it changed. It just required a little trickery with stamp files. This puts an end to byte-compiling python modules every time we run `make`.

- We now use a `ylwrap`-like trick for the generated unified .cpp files. Previously, changing an automake file would cause all of the unified .cpp files to be rewritten, regardless of whether their contents has changed. This invalidated basically the whole project. With this patch, we update the unified compilation rules when the Makefile changes, but we don't update the unified .cpp files themselves unless they would actually change. This trick works so well that dependency tracking for unified sources is now better than the dependency tracking provided natively by automake - even removing files is handled transparently.

Taken together, these changes eliminated all of the build amplification scenarios I tested. Should save us all a lot of time.


frontends/common/setup.h
```diff
@@ -0,0 +1,23 @@
+/*
```
mbudiu-vmw: Does this file logically belong to the front-end?

Makefile.am
```diff
@@ -79,45 +83,37 @@ include $(srcdir)/regenerate-unified-compilation.am
...
+V_IRGEN = $(__v_IRGEN_$(V))
+__v_IRGEN_ = $(__v_IRGEN_$(AM_DEFAULT_VERBOSITY))
+__v_IRGEN_0 = @echo "  IRGEN   " $@;
+__v_IRGEN_1 = 
+
+# We wrap the IR generator in `ylwrap` to reduce unnecessary full-project
+# rebuilds. `ylwrap` won't overwrite the generated headers unless they've
+# changed.
+ir/ir-generated.cpp: irgenerator $(ir_DEF_FILES)
+	$(V_IRGEN)$(SHELL) $(YLWRAP) _ ir-generated.cpp $@ ir-generated.h $(dir $@)ir-generated.h gen-tree-macro.h $(dir $@)gen-tree-macro.h -- ./irgenerator -i ir-generated.cpp -t gen-tree-macro.h -o ir-generated.h $(abspath $(ir_DEF_FILES))
```
mbudiu-vmw: Maybe add a comment with a reference to ir-generator.ypp to explain the underscore?

ChrisDodd: Does this still run `fixup-line-directives`?  Without it, the build will still work fine, but any errors or warnings in the .def files will be much harder to track down due to incorrect/obscured line number info

tools/ir-generator/generator.cpp
```diff
@@ -28,6 +28,7 @@ void usage(const char* progname) {
     fprintf(stderr, "usage:\n");
     fprintf(stderr, "%s [options] file.def file2.def ... >ir.h\n", progname);
```
ChrisDodd: remove the `>ir.h` from the useage line as it no longer applies.

frontends/common/setup.h
ChrisDodd: Logically belongs with the preprocessor, which is in the frontend, but currently in options.h/options.cpp.  We could perhaps just put these two declarations in options.h and get rid of this file, if that is not too obscure.

Makefile.am
sethfowler: Whoops, that's an oversight. I'll upload a new version of this commit.

Actually, after thinking about it more this morning I think it'd be better to avoid `ylwrap` and just use a stamp file, like I'm doing in the unified compilation code in a later commit. That'll avoid the new to patch the IR generator, too.

frontends/common/setup.h
sethfowler: Yeah, maybe that would be simplest; it does seem to belong with `CompilerOptions::preprocess()`.

sethfowler:
OK, I made the changes discussed above - the definitions in `setup.cpp` are now located in `options.cpp`, and I've written a custom solution for the IR generator output files instead of using `ylwrap`, which eliminated the need for the `_` hack. (And should reduce unnecessary rebuilds even further, to boot!)

I've also added an additional commit which draws on the lessons learned in writing the other patches to clean up the unified compilation code a little bit.

PR #699, opened 2017-06-05
Interface for architecture-specific conversion of P4_14 externs to P4_16

ChrisDodd:
Extern types in P4 are by definition architecture specific -- different architectures will support different ones.  In addition, they will often need to be different between P4_14 and P4_16 as P4_16 has more and different capabilities.  So when converting code from P4_14 there's a need to convert externs as well, but that cannot be done in an architecture-neutral way in general.

This change adds a interface hook to the P4_14->P4_16 converter that allows a backend/architecture to supply a custom extension for converting externs from P4_14 to P4_16 rather than just copying them through unchanged.

### DIFF
--- a/frontends/common/parseInput.h
+++ b/frontends/common/parseInput.h
@@ -18,6 +18,7 @@ limitations under the License.
 #define _FRONTENDS_COMMON_PARSEINPUT_H_
 
 #include "options.h"
+#include "frontends/p4/fromv1.0/converters.h"
 
 namespace IR {
 class P4Program;
@@ -38,7 +39,8 @@ namespace P4 {
  * @return a P4-16 IR tree representing the contents of the given file, or null
  * on failure. If failure occurs, an error will also be reported.
  */
-const IR::P4Program* parseP4File(CompilerOptions& options);
+const IR::P4Program* parseP4File(CompilerOptions& options,
+                                 P4V1::ExternConverter *extCvt = nullptr);
 
 /**
  * Parse P4 source from the string @input, interpreting it as having language
@@ -55,7 +57,8 @@ const IR::P4Program* parseP4File(CompilerOptions& options);
  * null on failure. If failure occurs, an error will also be reported.
  */
 const IR::P4Program* parseP4String(const std::string& input,
-                                   CompilerOptions::FrontendVersion version);
+                                   CompilerOptions::FrontendVersion version,
+                                   P4V1::ExternConverter *extCvt = nullptr);
 
 /**
  * Clear global program state so that a new program can be parsed.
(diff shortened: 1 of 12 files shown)


frontends/p4/fromv1.0/converters.cpp
```diff
@@ -321,6 +321,68 @@ const IR::Type_StructLike *TypeConverter::postorder(IR::Type_StructLike *str) {
 ///////////////////////////////////////////////////////////////
 
 namespace {
+class FixupExtern : public Modifier {
```
mbudiu-vmw: I will assume that there aren't many changes to review here, since this code seems to be just moved between files.
If you want me to read this carefully, please let me know.

frontends/p4/fromv1.0/converters.h
```diff
@@ -78,12 +78,23 @@ class TypeConverter : public ExpressionConverter {
     explicit TypeConverter(ProgramStructure* structure) : ExpressionConverter(structure) {}
 };
 
+class ExternConverter {
+ public:
+    ProgramStructure *structure = nullptr;
+
+    virtual const IR::Type_Extern *convertExternType(const IR::Type_Extern *, cstring);
```
mbudiu-vmw: Can you please Doxygen the parameters?
Or at least give a readable name to the cstring parameter?

frontends/p4/fromv1.0/converters.h
```diff
@@ -78,12 +78,23 @@ class TypeConverter : public ExpressionConverter {
     explicit TypeConverter(ProgramStructure* structure) : ExpressionConverter(structure) {}
 };
 
+class ExternConverter {
```
mbudiu-vmw: I am guessing you are subclassing this somewhere. In this case, I think this should be an interface (abstract class), and there should be also a "default" implementation.

frontends/common/parseInput.h
```diff
@@ -18,6 +18,7 @@ limitations under the License.
 #define _FRONTENDS_COMMON_PARSEINPUT_H_
 
 #include "options.h"
+#include "frontends/p4/fromv1.0/converters.h"
```
sethfowler: I'd just forward declare `P4V1::ExternConverter` to minimize header bloat.

frontends/p4/fromv1.0/converters.cpp
ChrisDodd: Yes -- this class is unchanged, just moved

frontends/p4/fromv1.0/converters.h
ChrisDodd: This is an interface that can be implemented in a backend.  It is also a (default) implementation of that interface, with the existing code that was being used for extern conversion.  Arguably there should be no default implementation (eg, it should be a pure abstract class), and the default should give an error if there are any externs in the P4_14 program.

ChrisDodd:
The goal is not to extend or otherwise encourage P4_14; it is to be able to support all existing P4_14 programs allowing users to transition from older P4_14 only compilers to the new compiler.

frontends/p4/fromv1.0/converters.h
mbudiu-vmw: Yes, that is what I meant.
Or it could be a pure virtual class, and the users have to provide some implementation.
Most policies in the mid-end are described by pure virtual classes. This one is not in the mid-end, though.

PR #706, opened 2017-06-10
Support for temporaries with varbit type

mbudiu-vmw:
This change also removes the hardwired "scalars" name from the json, since it may conflict with other variable names.


backends/bmv2/action.cpp
```diff
@@ -45,7 +45,11 @@ ConvertActions::convertActionBody(const IR::Vector<IR::StatOrDecl>* body, Util::
             cstring operation;
             auto type = typeMap->getType(l, true);
             if (type->is<IR::Type_StructLike>())
-                operation = "copy_header";
+                operation = "assign_header";
+            else if (type->is<IR::Type_Varbits>())
+                operation = "assign_VL";
+            else if (type->is<IR::Type_HeaderUnion>())
```
ChrisDodd: `Type_HeaderUnion` is a subclass of `Type_StructLike`, so this test will never be true -- it would always go into the `if (type->is<IR::Type_StructLike>())` case instead.

PR #707, opened 2017-06-12
add --top4 and -T to top level options in p4c driver

hanw:


hanw:
These flags are frequently used during development, passing them through -Xp4c is painful. However, logically they belong to the "development" version of p4c, not "production" version. But we don't have that distinction.

cc10512:
Sure we do now. p4c.in defines os.environ['P4C_BUILD_TYPE'] as DEVELOPER or INSTALLED.

hanw:
Ah, okay, then I will update the patch to only add these flag to the developer version.

PR #714, opened 2017-06-14
Fix memory error when building bmv2 JSON

antoninbas:
Backend::convert was inserting json->meta into jsonTop, but that pointer
would later get overwritten in add_meta_info.


antoninbas:
@teverman can you see if this solves your issue

teverman:
Problem solved, thanks.

Issue #510, opened 2017-04-23
Including 'error' as a struct member crashes the compiler


cc10512:
As reported by @vgurevich, the following snippet:
```
struct my_metadata_t {
    error parser_error;
}
```
crashes the compiler
```
terminate called after throwing an instance of 'Util::CompilerBug'
  what():  COMPILER BUG: ../backends/bmv2/jsonconverter.cpp:2487
/home/vgurevich/hh/p4c/build/p4include/core.p4(22): error { NoError, PacketTooShort, NoMatch, StackOutOfBounds, OverwritingHeader, HeaderTooShort, ParserTimeout }: Unhandled type for error { NoError, PacketTooShort, NoMatch, StackOutOfBounds, OverwritingHeader, HeaderTooShort, ParserTimeout } parser_error
error {
^
/tmp/error_in_struct.p4(29)
    error parser_error;
          ^^^^^^^^^^^^

Aborted (core dumped)
```

Full program attached: [error_in_struct.p4.txt](https://github.com/p4lang/p4c/files/949895/error_in_struct.p4.txt)

mbudiu-vmw:
error is a funny one; as far as I can tell, BMv2 does not really expose the parser error at all - you cannot read the error generated by the parser, and it is not exposed as part of the architecture metadata.

But this does not preclude the user from using error as an enum.
So in the BMv2 back-end we can just treat it like a special enum.

PR #715, opened 2017-06-15
Fix for issue #510

mbudiu-vmw:


ChrisDodd:
It might make sense (for some targets where bits are a premium) to try to figure out the minimum number of bits to use for an error code (to allow all declared errors to have a distinct code), rather than just using 32, but that seems like significantly more work?

mbudiu-vmw:
The 32 bit size was hardwired by @hanw.
I only made it a constant.
This code is in the bmv2 back-end, and not in a reusable pass. We probably should have a mid-end pass for this purpose, which would work a lot like the enum conversion pass, by taking a policy class.

hanw:
wasn't me =), the 32 bit was there before the refactoring. but I agree with the policy class approach.

mbudiu-vmw:
In fact it was @antoninbas with this commit: https://github.com/p4lang/p4c/commit/6c3f06aef57bf794403591c1d22e48b08f0395e9
Sorry for the mis-attribution.

PR #716, opened 2017-06-15
Fix bug in LocalCopyPropagation pass in midend

ksubramanianBarefoot:
The LocalCopyPropagation pass in midend checks and removes assignment statements of the form: 
`modify_field(metadata.port, metadata.port)` 
However, these statements were not removed in practice because the check was comparing the IR nodes (which were different for the same variable), and thus, none of the no-op assignment statements were eliminated. This PR fixes that bug. Two conditions for equality are required based on the current methods: 1) For header/metadata fields and 2) variables. 

This PR requires changes to the testdata outputs because some of the sample programs have assignments of the form `a=a` which can be safely removed.


ksubramanianBarefoot:
@evanches @ChrisDodd

midend/local_copyprop.cpp
```diff
@@ -205,7 +205,16 @@ IR::AssignmentStatement *DoLocalCopyPropagation::preorder(IR::AssignmentStatemen
 }
 
 IR::AssignmentStatement *DoLocalCopyPropagation::postorder(IR::AssignmentStatement *as) {
-    if (as->left == as->right) {   // FIXME -- need deep equals here?
+    if (as->left->toString() == as->right->toString() &&
```
ChrisDodd: Converting an operation to a string generally just returns the operation.  So if you have an assignment like `a[0] = b[1]`, both `toString()` calls will return `"[]"` and this will incorrectly eliminate the assignment.

The right fix (as noted in the FIXME) is to use a deep equality comparison.  Doing that is probably best done by adding a `deep_equals` method in the IR generator that will automatically create a deep comparison operation that could be used here -- on my TODO list, but keeps getting pushed back by other things.

ChrisDodd: For now, we could create a small `equiv` function that takes two `const IR::Expression *` and does some deeper comparisons (eg two PathExpressions with Paths with the same same) that would fix this specific case and not break other things, defined as a static function right above this.  Be sure to include a FIXME comment about this being temporary until a proper general deep comparison is implemented.

PR #717, opened 2017-06-15
Add support for P4Runtime ControllerPacketMetadata

antoninbas:
This is a new P4Info / P4Runtime feature that enablese passing metadata
between the controller and the switch for Packet IO. The metadata format
is defined by P4 header objects which are tagged with a
@controller_metadata annotation.

### DIFF
--- a/control-plane/p4RuntimeSerializer.cpp
+++ b/control-plane/p4RuntimeSerializer.cpp
@@ -69,8 +69,13 @@ static bool isMetadataType(const IR::Type* type) {
 }
 
 /// @return true if @node has an @hidden annotation.
-static bool isHidden(const IR::Node *node) {
+static bool isHidden(const IR::Node* node) {
     return node->getAnnotation("hidden") != nullptr;
+}
+
+/// @return true if @type has an @controller_header annotation.
+static bool isControllerHeader(const IR::Type_Header* type) {
+    return type->getAnnotation("controller_header") != nullptr;
 }
 
 /// @return a version of @name which has been sanitized for exposure to the
@@ -437,6 +442,7 @@ externalId(const IR::IDeclaration* declaration) {
 enum class P4RuntimeSymbolType {
     ACTION,
     ACTION_PROFILE,
+    CONTROLLER_HEADER,
     COUNTER,
     METER,
     TABLE
@@ -583,6 +589,7 @@ public:
         symbols.computeIdsForSymbols(P4RuntimeSymbolType::ACTION_PROFILE);
         symbols.computeIdsForSymbols(P4RuntimeSymbolType::COUNTER);
         symbols.computeIdsForSymbols(P4RuntimeSymbolType::METER);
+        symbols.computeIdsForSymbols(P4RuntimeSymbolType::CONTROLLER_HEADER);
 
         return symbols;
     }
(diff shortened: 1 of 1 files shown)


antoninbas:
@teverman @wmohsin support added in compiler

antoninbas:
sample output:
```
controller_packet_metadata {
  preamble {
    id: 2868932736
    name: "packet_in"
    annotations: "@controller_header(\"packet_in\")"
  }
  metadata {
    id: 1
    name: "f1"
    annotations: "@proto_tag(9)"
    bitwidth: 16
  }
}
```

control-plane/p4RuntimeSerializer.cpp
```diff
@@ -439,7 +444,8 @@ enum class P4RuntimeSymbolType {
     ACTION_PROFILE,
     COUNTER,
     METER,
-    TABLE
+    TABLE,
+    CONTROLLER_HEADER
```
sethfowler: Nit: it'd be better to keep this list alphabetical.

control-plane/p4RuntimeSerializer.cpp
```diff
@@ -754,6 +761,9 @@ class P4RuntimeSymbolTable {
             case P4RuntimeSymbolType::COUNTER: return PI_COUNTER_ID;
             case P4RuntimeSymbolType::METER: return PI_METER_ID;
             case P4RuntimeSymbolType::TABLE: return PI_TABLE_ID;
+            // XXX(antonin): no support from PI library for now
+            // the choice for the resource type id has little importance here
+            case P4RuntimeSymbolType::CONTROLLER_HEADER: return 0xab;
```
sethfowler: I ran into this same situation in my own recent patch. I suppose we have a number of these resource types to add at this point.

control-plane/p4RuntimeSerializer.cpp
```diff
@@ -770,7 +780,8 @@ class P4RuntimeSymbolTable {
         { P4RuntimeSymbolType::ACTION_PROFILE, SymbolTable() },
         { P4RuntimeSymbolType::COUNTER, SymbolTable() },
         { P4RuntimeSymbolType::METER, SymbolTable() },
-        { P4RuntimeSymbolType::TABLE, SymbolTable() }
+        { P4RuntimeSymbolType::TABLE, SymbolTable() },
+        { P4RuntimeSymbolType::CONTROLLER_HEADER, SymbolTable() }
```
sethfowler: Nit: again, it'd be better to keep this list alphabetical.

control-plane/p4RuntimeSerializer.cpp
```diff
@@ -976,6 +987,50 @@ class P4RuntimeSerializer {
...
+            ::error("@controller_header should be a string for declaration %1%", type);
+            return;
+        }
+        auto controllerName = nameConstant->value;
+
+        auto header = p4Info->add_controller_packet_metadata();
+        header->mutable_preamble()->set_id(id);
+        // According to the p4info specification, we use the name specified in the annotation for
+        // the p4info preamble, not the P4 fully-qualified name.
+        header->mutable_preamble()->set_name(controllerName);
```
sethfowler: So we're computing the id based on the P4 name, but we're assigning the name based on the annotation. That's different than any other P4Info object. I don't really see why it'd be a problem, though.

teverman: What if someone wants to add an ID annotation to the metadata?

antoninbas: @sethfowler that's correct
@teverman you can assign an ID to the header with `@id` just like for a table, action etc. For field ids in the controller header, we do not support the `@id` annotation yet (as for match key entries & action parameters), we use the index instead; but we can change that soon.

antoninbas:
@sethfowler this is not an extern; we have decided to have built-in support in P4Runtime for controller headers, even though these headers have no special status at all in the P4 language
additionally, regarding externs, we are strongly leaning towards having all of the PSA externs supported directly in P4Runtime (as opposed to supporting them with `Any`) to emphasize their special status when it comes to writing portable programs.

sethfowler:
@antoninbas Including all of the PSA externs directly would be fine with me. If we chose to, we could also support them with Any, but ship `psa.proto` with P4Runtime as a way of privileging them. I was actually surprised at how easy it is to work with Any; it doesn't make things much harder.

Issue #497, opened 2017-04-19
p4c-bm2-ss writes incorrect JSON file if there are "Removing unused action parameter" warnings


jafingerhut:
See attached .p4 and .json files.  When compiled there are correct warnings about unused parameters for action set_bd_dmac_intf.  If you look at action set_bd_dmac_intf in the JSON file, it has 3 entries for the used arguments under key "runtime_data".  However, under the "primitives" key, every time there is a "modify_field" primitive action with source data from a parameter that comes after an unused (and thus removed) parameter, the index is the original index in the parameter list, before the unused ones were removed, which are now wrong.
[demo1-variant.p4.txt](https://github.com/p4lang/p4c/files/932756/demo1-variant.p4.txt)
[demo1-variant.p4.txt](https://github.com/p4lang/p4c/files/932757/demo1-variant.p4.txt)

jafingerhut:
And if anyone else hits this issue before it is fixed, it does seem that removing all unused action parameters until those warnings are gone does work around this problem, as one might expect.

sethfowler:
IMO, we should just stop removing unused action parameters.

We deliberately aren't doing it for P4Runtime.

Is there any reason not to make that change?

mbudiu-vmw:
I was about to work on a fix, but indeed, it would be better if we could keep the unused parameters.
This behavior is inherited from the P1-14 compiler.

jafingerhut:
I think continuing to warn about unused parameters would be useful, even if they are left in the bmv2 JSON file.

mbudiu-vmw:
I will attempt to provide a fix for this that uses an #if REMOVE_PARAMETERS. So then we can do it in either way. But, of course, I won't be able to test both cases until BMv2 handles unused parameters.

Note that removing parameters violates the contract that the control-plane API is readable from the program.

antoninbas:
There is nothing required from bmv2 AFAIK. If you leave an unused parameter in the JSON, nothing bad will happen, it will just not be used. And as pretty much everyone has mentioned, from a control-plane perspective, it is better to leave unused parameters and simply print a warning.

mbudiu-vmw:
OK, then I will just them in. Much simpler.
Note that there seem to be several Travis build failures related to this bug, I don't know why it didn't show up before.

antoninbas:
[...]

PR #719, opened 2017-06-15
Add support for PD generation via P4Runtime

sethfowler:
We'd like to start running PTF tests against p4c's compiler output. Existing PTF tests are written against the old control plane API tooling, PD, rather than the new P4Runtime tooling. For that reason, at least during this transition period it'd be nice to be able to generate a PD frontend from p4c's output. In practice that's best handled by generating it from the P4Runtime configuration that p4c can already produce.

Concretely, the goal I'm trying to achieve in this PR specifically is to generate a PD frontend for [switch.p4](https://github.com/p4lang/switch) that is identical to the output of [p4c-bm](https://github.com/p4lang/p4c-bm).

There were a couple of minor issues originating in the compiler itself that resulted in control plane API differences. We inferred a different type for `execute_meter`'s second argument. We also aren't removing unused action parameters for P4-14 programs on BMV2, but we need to do that for compatibility with `p4c-bm`. I fixed both of those issues.

We also need to extend the P4Runtime messages available in the PI repository with some additional Extern messages. For now, I just added Digest, since Register isn't used by switch.p4 in its default configuration; we'll eventually need it, though. To make this work, I reworked the way that `protoc` is invoked in the build system to make it easier to add new `.proto` files. I then added a new `v1model.proto` to hold architecture-specific Extern messages, and I added the Digest message and support for it in p4RuntimeSerializer.

That gets us to the point where we have all of the information we need in the P4Runtime, but we still need to generate the PD frontend. The final commit in this PR adds `p4runtime-to-pd`, a script which generates a PD frontend from a P4Runtime file. This script has a complicated heritage; the templates are from `p4c-bm` (to ensure that the output is identical), while the code is an extended version of `pi_gen_pd` that has been reworked to take P4Runtime as input directly instead of PI or BMV2 JSON. It might seem natural that this script should live in the `PI` repo, but unfortunately it depends on `.proto` files that live here. New architectures would need to either enhance `p4runtime-to-pd` or add their own implementation, anyway; it's beyond the scope of what the `PI` repo should contain.

The new script is only built if you configure with the `--enable-p4runtime-to-pd` flag; I don't want to encourage anyone to build new projects that use PD. Usage is as follows:

```bash
./p4c-bm2-ss --p4v 14 --p4runtime-file switch.p4runtime switch.p4
mkdir -p switch-pd
./p4runtime-to-pd  switch.p4runtime --pd switch-pd --target bm
```

Using this script, a PD frontend almost identical to that generated by `p4c-bm` can be generated using only the new `p4c`/`P4Runtime` toolchain. There are a couple of tiny differences:

- `pd_tables.cpp` uses table ids internally, and since we assign different ids than `p4c-bm` does, the output is different. This doesn't appear to matter, though, as the table ids are just used as unique identifiers internally and don't leak out of the generated code.
- `pd_learning.cpp` uses ids for learning quanta, and unfortunately those do leak out of the generated PD code. We can fix that with a small addition to the PD fixed API. We'll probably need to (or we'll need to find another solution), because P4Runtime ids are also not the same as the ones generated by the BMV2 backend, and I suspect that things need to match up for these APIs to work correctly.

There's one other todo: we should define PI resource types for Externs. The intention is to have a vendor-specific resource type for each vendor's ExternInstances, I believe, and in that case we'll want to define one for p4.org as well. This is something Antonin and I can get straightened out after this patch lands.


sethfowler:
It looks like 16 tests fail due to the type inference changes. At first glance all of the output changes seem OK. I'll go through them to be sure and then add a commit that updates the test outputs.

ir/v1.cpp
```diff
@@ -131,14 +133,10 @@ unsigned IR::Primitive::inferOperandTypes() const {
...
-    if ((name == "count" || name == "execute_meter") && operand == 1) {
-        if (auto obj = operands[0]->to<IR::GlobalRef>()) {
-            if (auto tbl = obj->obj->to<IR::Stateful>()) {
-                if (tbl->instance_count > 0) {
-                    int width = ceil_log2(tbl->instance_count);
-                    return IR::Type::Bits::get(width); } } } }
-    if (name.startsWith("execute_stateful") && operand == 1) {
-            return IR::Type::Bits::get(32); }
+    if ((name == "count" || name == "execute_meter") && operand == 1)
+        return IR::Type::Bits::get(32);
```
ChrisDodd: Is there a reason to remove the calculation of the correct size and use a fixed, incorrect size?

control-plane/p4RuntimeSerializer.cpp
```diff
@@ -578,11 +596,8 @@ class P4RuntimeSymbolTable {
         function(symbols);
 
         // Now that the symbol table is initialized, we can compute ids.
-        symbols.computeIdsForSymbols(P4RuntimeSymbolType::ACTION);
-        symbols.computeIdsForSymbols(P4RuntimeSymbolType::TABLE);
-        symbols.computeIdsForSymbols(P4RuntimeSymbolType::ACTION_PROFILE);
-        symbols.computeIdsForSymbols(P4RuntimeSymbolType::COUNTER);
-        symbols.computeIdsForSymbols(P4RuntimeSymbolType::METER);
+        for (auto& table : symbols.symbolTables)
+            symbols.computeIdsForSymbols(table.first);
```
cc10512: You'll have a conflict with @antoninbas's PR: https://github.com/p4lang/p4c/pull/717 on this.
That said, this is making the implementation more scalable when we add other symbol types.

ir/v1.cpp
sethfowler: Yep, I believe I mentioned it in the first comment. The calculation here, while clearly more accurate, is not backwards compatible with p4c-bm. We can't generate a compatible control plane API without this change.

sethfowler:
@mbudiu-vmw, I deliberately made this tool somewhat obscure because I don't want to encourage people to use it. If there's reason to draw more attention to it, we can certainly add it to the README.

sethfowler:
@cc10512, the pi_gen_pd tool generates PD that is quite different from what p4c-bm generates. It also requires some indirection through a different JSON format that makes things much more difficult. When we extend P4Runtime with additional Extern definitions, as I've done in this PR, we don't want to have to update the intermediate tools to plumb them through an unrelated JSON format. Generating PD directly from P4Runtime makes a lot more sense. Nothing has changed in this regard; this was always the path of least resistance.

As I mentioned in the first comment, such a tool is also inherently architecture specific to some degree. It's not clear that it even makes sense for it to live in the PI repo. (And it can't, really, unless we move v1model.proto in there too.) We will need a different version of this tool to support the features of the chips Barefoot makes, for example, because we will need both different templates and support in the code for different externs.

Regarding the templates included in this PR, they are identical to the ones in the p4c-bm repo. We could include that repo as a submodule. Given that p4c-bm development is not very active, that felt unnecessary to me, but I'm happy to go that way if you prefer.

cc10512:
Thanks for the explanation.

sethfowler:
Sure.

@mbudiu-vmw, on reflection I think you're absolutely right that we should have a `README.md` documenting `p4runtime-to-pd`. I've added a new commit that adds one in the `p4runtime-to-pd` directory. I still hesitate to mention this tool in the top-level `README.md` for the project, though, so for now I've left it out.

ir/v1.cpp
sethfowler: Interestingly, if you examine the test failures, we were always casting to bit<32> anyway when we actually pass the operands to the extern method. For example, before we had:

```p4
     @name(".m_action") action m_action_0(bit<14> idx) {
         my_indirect_counter.count((bit<32>)idx);
     }
```

After this patch, we have:

```p4
     @name(".m_action") action m_action_0(bit<32> idx) {
         my_indirect_counter.count(idx);
     }
```

The second version is consistent with what `p4c-bm` generates.

sethfowler:
I've updated the PR to include updated test outputs. That should hopefully turn Travis green.

backends/bmv2/midend.cpp
```diff
@@ -122,6 +123,11 @@ MidEnd::MidEnd(BMV2Options& options) {
         new P4::CompileTimeOperations(),
         new P4::TableHit(&refMap, &typeMap),
         new P4::RemoveLeftSlices(&refMap, &typeMap),
+
+        // p4c-bm removed unused action parameters. To produce a compatible
+        // control plane API, we remove them as well for P4-14 programs.
+        isv1 ? new P4::RemoveUnusedActionParameters(&refMap) : nullptr,
```
antoninbas: just checking, doesn't that mean this issue is going to come up again: https://github.com/p4lang/p4c/issues/497

sethfowler: I believe the answer is no, because in that case we were just skipping them when writing the JSON output, but in this case we're actually cleanly removing them from the IR at an earlier stage.

control-plane/p4RuntimeSerializer.cpp
```diff
@@ -1106,24 +1154,141 @@ class P4RuntimeSerializer {
...
+
+    auto type = refMap->getDeclaration(typeArg->to<IR::Type_Name>()->path, true);
+    if (!type->is<IR::Type_Struct>()) {
+        ::error("%1%: Expected a struct type", type);
+        return boost::none;
+    }
+
+    // The first argument is a numeric constant identifying the receiver.
+    auto receiver = call->arguments->at(0)->to<IR::Constant>();
+    if (!receiver->is<IR::Constant>()) {
```
mbudiu-vmw: This test should probably be for nullptr.

backends/bmv2/midend.cpp
mbudiu-vmw: We have been trying to remove all differences between P4-14 and P4-16 compilation. This flag means that if you convert a p4-14 program to p4-16 then the compilation will produce a different result than if you start with just p4-14.
So it looks to me like what you want is a new option flag instead; "--removeUnusedActionArgs", which is provided for compatibility with some old tools.

sethfowler:
@mbudiu-vmw I'd prefer not to introduce such a flag. The right approach is probably to recognize that the conversion from P4-14 to P4-16 is partially architecture-dependent, and handle this as part of the conversion for P4-14 programs written for BMV2. There are similar issues with P4-14 programs written for Barefoot chips, since the existing compilers have somewhat incompatible behavior with `p4c-bm`.

I think that'd best be handled as a followup to this pull request, since it's a somewhat complicated change in its own right. The patch in this PR reintroduces (for P4-14 only) a behavior that we only eliminated within the last month or two, so I don't think we'll suffer any consequences for reintroducing it.

control-plane/p4RuntimeSerializer.cpp
sethfowler: Good catch! I'll make that change.

sethfowler:
By the way, do we only run the frontend when converting from P4-14 to P4-16? If we run the midend passes before writing out the converted P4-16 source file, then this will work just fine in any case. I'm removing the unused arguments from the IR, so the serialized version will not have them, and compiling the converted file will produce exactly the same results.

sethfowler:
@mbudiu-vmw the bug in the conversion code you pointed out has been fixed in the version of the PR I just uploaded, btw.

mbudiu-vmw:
Pretty-printing (--pp) runs as the first front-end pass.
The missing arguments is an important question, though, it will also affect STF tests, and anything that touches the control-plane.

sethfowler:
Hmm. That confirms that we'll need to handle this architecture/target-specific stuff as part of the conversion process. I agree that it's a problem and we will need to implement a proper solution for it.

The effect on the control plane is intentional. I'm removing the unused arguments specifically to fix the control plane API we generate. Without this change, we can't run the same tests against `p4c-bm` and this compiler.

sethfowler:
Just as a status update: I've haven't merged this because I noticed that `p4runtime-to-pd` doesn't work correctly when installed with `make install`. I have a fix already locally (some paths just need to be changed in `p4runtime-to-pd.in`) but I'm too tired to finish up tonight. I'll merge once I get that fix uploaded.

I'll also file an issue where we can tackle the arch/target-specific P4-14 conversion problem.

PR #721, opened 2017-06-16
Added contiguous function to bitvec

evanches:
Just added a continuous function to bitvec.  I needed something in the Barefoot backend, but the function is useful for any bitvec.

Not optimized for speed, but is correct.  Could be optimized if that is a major concern for people.

### DIFF
--- a/lib/bitvec.cpp
+++ b/lib/bitvec.cpp
@@ -140,9 +140,7 @@ unsigned bitvec::ffz(unsigned start) const {
 
 bool bitvec::is_contiguous() const {
     // Empty bitvec is not contiguous
-    if (popcount() == 0)
+    if (empty())
         return false;
-    if (max().index() - min().index() + 1 == popcount())
-        return true;
-    return false;
+    return max().index() - min().index() + 1 == popcount();
 }
--- a/lib/bitvec.cpp
+++ b/lib/bitvec.cpp
@@ -137,3 +137,12 @@ unsigned bitvec::ffz(unsigned start) const {
 #endif
     return rv;
 }
+
+bool bitvec::is_contiguous() const {
+    // Empty bitvec is not contiguous
+    if (popcount() == 0)
+        return false;
+    if (max().index() - min().index() + 1 == popcount())
+        return true;
+    return false;
+}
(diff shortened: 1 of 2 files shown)


lib/bitvec.cpp
```diff
@@ -137,3 +137,12 @@ unsigned bitvec::ffz(unsigned start) const {
 #endif
     return rv;
 }
+
+bool bitvec::is_contiguous() const {
+    // Empty bitvec is not contiguous
+    if (popcount() == 0)
```
sethfowler: I know you said you weren't optimizing for speed, but it's an easy win to save `popcount()` to a temporary variable instead of calling it twice. I think it's worth doing.

lib/bitvec.cpp
```diff
@@ -137,3 +137,12 @@ unsigned bitvec::ffz(unsigned start) const {
...
+        return false;
+    if (max().index() - min().index() + 1 == popcount())
+        return true;
+    return false;
```
sethfowler: These three lines are the same as:

```c++
return max().index() - min().index() + 1 == popcount();
```

lib/bitvec.cpp
```diff
@@ -137,3 +137,12 @@ unsigned bitvec::ffz(unsigned start) const {
 #endif
     return rv;
 }
+
+bool bitvec::is_contiguous() const {
+    // Empty bitvec is not contiguous
```
sethfowler: I think a reasonable argument could be made for returning true if the bitvec is empty, as well. It'd be good to document this method's behavior in the header file.

lib/bitvec.cpp
ChrisDodd: Even better as then you don't need the check for `popcount() == 0` as when that is true, `max.index() == min.index() == -1;`.   Or is that too obscure?

lib/bitvec.cpp
ChrisDodd: using `if (empty()) return false;` might be clearer?  I guess if you're using popcount() anyways.

lib/bitvec.cpp
evanches: I felt it was very obscure, as the reason why I didn't have it.

lib/bitvec.cpp
evanches: The whole header file could use a range of documentation.  It has 0.

PR #726, opened 2017-06-19
support for @optional parameters to functions in type unification

ChrisDodd:


frontends/p4/typeChecking/typeUnification.cpp
```diff
@@ -119,15 +119,16 @@ bool TypeUnification::unifyFunctions(const IR::Node* errorPosition,
...
-        if (reportErrors)
-            ::error("%1%: Cannot unify functions with different number of arguments: %2% to %3%",
-                    errorPosition, src, dest);
-        return false;
-    }
 
     auto sit = src->parameters->parameters.begin();
     for (auto dit : *dest->parameters->getEnumerator()) {
+        if (sit == src->parameters->parameters.end()) {
+            if (dit->getAnnotation("optional"))
```
cc10512: this should increment the sit if the source parameter is optional. Otherwise, you handle optional parameters only at the end of the list.

ChrisDodd: At the moment, yes it only support trailing optional params (C++ style).  Supporting mid-list optionals will require generating multiple constraint sets (for with/without the arg), then pruning off the ones that are the wrong length, then solving all remaining possibilities (with an ambiguity check if more than one has a solution).  Possible, but much more work.

PR #728, opened 2017-06-19
Prepare the Dockerfile for running PD-based PTF tests

sethfowler:
We need to make a few changes to the Dockerfile to run PD-based PTF tests.

We need to use scapy-vxlan instead of Ubuntu's scapy package - here we're relying on scapy-vxlan being present in the upstream third-party image. (see p4lang/third-party#11)

We also need `p4runtime-to-pd` available in the Docker image, so I've enabled it. That requires us to install the `tenjin` Python package.

Finally, I noticed while making these changes that the `||` in the `RUN` statement makes `docker build` succeed even if `bootstrap.sh` returns an error. Multistage builds will eliminate the need for these huge shell expressions, but it may take as long as two weeks for us to be able to use them, it's worth fixing the problem now.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -25,6 +25,8 @@ ENV P4C_DEPS automake \
              libtool \
              pkg-config \
              python-ipaddr \
+             python-pip \
+             python-setuptools \
              tcpdump
 ENV P4C_RUNTIME_DEPS cpp \
                      libboost-iostreams1.58.0 \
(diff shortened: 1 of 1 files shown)


sethfowler:
The build failed because the `scapy-vxlan` patch isn't on the stable branch of the third-party repo. I'll fix that and retrigger.

sethfowler:
Done. It'll take some time for the cascade of Docker Hub builds to catch up, though.

PR #729, opened 2017-06-19
Maintain P4-14 names more accurately when generating BMV2 JSON

sethfowler:
We currently generate BMV2 JSON that uses P4-16-style hierarchical names. These don't match the output of `p4c-bm`, resulting in a different and incompatible control plane API that prevents us from running the existing suite of PTF tests for switch.p4.

Fixing this is straightforward: we need to be careful about always marking names as global (i.e., prefixing them with a dot) during 14-to-16 conversion. The exception is type names; they're "global" already in P4-16 (i.e., we don't have nested types or types which are scoped to controls) and there's some existing code which concatenates type names in a way that produces duplicate dots. It's probably not too hard to fix that code, but to limit the scope of this PR I haven't done so here.

There are existing functions in the BMV2 backend and in the P4Runtime serializer that transform these global names to remove the leading dot, since we don't want to expose that to the control plane. I've removed those functions and replaced them with a single implementation: IDeclaration::controlPlaneName().

With those changes, it's a simple (if tedious) matter to find all the places where we output names in the BMV2 JSON generation code and call controlPlaneName(). That fixes the control plane API that we generate.

I had to update a bunch of test outputs to replace `@name("foo")` with `@name(".foo")`.


frontends/p4/fromv1.0/programStructure.cpp
```diff
@@ -59,7 +59,14 @@ const IR::Annotations*
 ProgramStructure::addNameAnnotation(cstring name, const IR::Annotations* annos) {
     if (annos == nullptr)
         annos = IR::Annotations::empty;
-    return annos->addAnnotationIfNew(IR::Annotation::nameAnnotation, new IR::StringLiteral(name));
+    return annos->addAnnotationIfNew(IR::Annotation::nameAnnotation,
+                                     new IR::StringLiteral(name));
+}
+
+const IR::Annotations*
+ProgramStructure::addGlobalNameAnnotation(cstring name,
```
mbudiu-vmw: The spec mandates a `globalname` annotation, which we have not used so far. It should imply a leading dot.
I don't know if we also allow `name` annotations with a leading dot, and if there is any difference.

frontends/p4/fromv1.0/programStructure.h
```diff
@@ -169,7 +169,35 @@ class ProgramStructure {
  protected:
     const IR::Statement* convertPrimitive(const IR::Primitive* primitive);
     void checkHeaderType(const IR::Type_StructLike* hrd, bool toStruct);
-    const IR::Annotations* addNameAnnotation(cstring name, const IR::Annotations* annos = nullptr);
+
+    /**
+     * Extend the provided set of annotations with an '@name' annotation for the
+     * given name. This is used to preserve the original, P4-14 object names.
+     *
+     * In general, because P4-14 names do not have the hierarchical structure
```
mbudiu-vmw: The P4-14 names are not hierarchical, with one exception which is not specified in the P4-14 spec: calling a control multiple times. It is not clear what that means, and how names should be assigned for the tables inside the called control.

frontends/p4/fromv1.0/programStructure.cpp
sethfowler: Yeah, at some point we need to synchronize the way this stuff works with the spec. It's currently somewhat out of date.

frontends/p4/fromv1.0/programStructure.h
sethfowler: It sounds like we may need to tweak this if we decide to handle that case, yeah.

sethfowler:
Looks like this failed on Travis because there was a new test introduced in a commit I rebased on top of, which I didn't catch. It needs the `@name("foo")` -> `@name(".foo")` transformation.

sethfowler:
@ChrisDodd can we merge this?

ChrisDodd:
Sorry -- I thought I approved it previously, but apparently not

PR #730, opened 2017-06-19
Fix bug in LocalCopyPropagation pass in midend

ksubramanianBarefoot:
Note: This PR is addressing the same bug as #716 with the required changes to code and the test outputs (I added other commits in that PR by mistake, so created a new one with only my commits).  

The LocalCopyPropagation pass in midend checks and removes assignment statements of the form:
`modify_field(metadata.port, metadata.port)`
However, these statements were not removed in practice because the check was comparing the IR nodes (which were different for the same variable), and thus, none of the no-op assignment statements were eliminated. This PR fixes that bug by adding an function to check equivalence of two IR::Expression objects for 1) Variables 2) Array references 3) Packet header/metadata fields (struct). 

This PR changes the testdata outputs because some of the sample programs have assignments of the form a=a which can be safely removed.

### DIFF
--- a/midend/local_copyprop.cpp
+++ b/midend/local_copyprop.cpp
@@ -65,8 +65,14 @@ class DoLocalCopyPropagation::ElimDead : public Transform {
         if (s->ifTrue == nullptr) {
             /* can't leave ifTrue == nullptr, as that will fail validation -- fold away
              * the if statement as needed */
-            if (s->ifFalse == nullptr)
-                return nullptr;
+            if (s->ifFalse == nullptr) {
+                if (!hasSideEffects(s->condition)) {
+                    return nullptr;
+                } else {
+                    s->ifTrue = new IR::EmptyStatement();
+                    return s;
+                }
+            }
             s->ifTrue = s->ifFalse;
             s->ifFalse = nullptr;
             s->condition = new IR::LNot(s->condition); }
@@ -204,8 +210,34 @@ IR::AssignmentStatement *DoLocalCopyPropagation::preorder(IR::AssignmentStatemen
     return postorder(as);
 }
 
+/* Function to check if left and right side of
+ * assignment statement are equivalent.
+ * FIXME -- This function is temporary, proper deep comparison may be required */
+bool DoLocalCopyPropagation::equiv(const IR::Expression *left, const IR::Expression *right) {
+    // Compare names of variables (at this pass, all names are unique)
+    auto l1 = left->to<IR::PathExpression>();
+    auto r1 = right->to<IR::PathExpression>();
+    if (l1 && r1) {
+        return l1->path->name == r1->path->name;
+    }
+    // Compare references to arrays with indices
+    auto l2 = left->to<IR::Operation_Binary>();
+    auto r2 = right->to<IR::Operation_Binary>();
+    if (l2 && r2) {
+        return equiv(l2->left, r2->left) && equiv(l2->right, r2->right);
+    }
+    // Compare packet header/metadata fields
+    auto l3 = left->to<IR::Member>();
+    auto r3 = right->to<IR::Member>();
+    if (l3 && r3) {
+        return l3->member == r3->member && equiv(l3->expr, r3->expr);
+    } else {
+        return false;
+    }
+}
+
 IR::AssignmentStatement *DoLocalCopyPropagation::postorder(IR::AssignmentStatement *as) {
-    if (as->left == as->right) {   // FIXME -- need deep equals here?
+    if (equiv(as->left, as->right)) {
         LOG3("  removing noop assignment " << *as);
         return nullptr; }
     if (!working) return as;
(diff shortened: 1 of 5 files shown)


midend/local_copyprop.cpp
```diff
@@ -204,8 +210,34 @@ IR::AssignmentStatement *DoLocalCopyPropagation::preorder(IR::AssignmentStatemen
...
 
+/* Function to check if left and right side of
+ * assignment statement are equivalent.
+ * FIXME -- This function is temporary, proper deep comparison may be required */
+bool DoLocalCopyPropagation::equiv(const IR::Expression *left, const IR::Expression *right) {
+    // Compare names of variables (at this pass, all names are unique)
+    auto l1 = left->to<IR::PathExpression>();
+    auto r1 = right->to<IR::PathExpression>();
+    if (l1 && r1) {
+        return l1->path->name == r1->path->name;
```
mbudiu-vmw: `&& l1->path->isAbsolute == r1->path->isAbsolute`

midend/local_copyprop.cpp
```diff
@@ -204,8 +210,34 @@ IR::AssignmentStatement *DoLocalCopyPropagation::preorder(IR::AssignmentStatemen
...
+    }
+    // Compare references to arrays with indices
+    auto l2 = left->to<IR::Operation_Binary>();
+    auto r2 = right->to<IR::Operation_Binary>();
+    if (l2 && r2) {
+        return equiv(l2->left, r2->left) && equiv(l2->right, r2->right);
```
mbudiu-vmw: Shouldn't these also have the same operation?

midend/local_copyprop.cpp
```diff
@@ -204,8 +210,34 @@ IR::AssignmentStatement *DoLocalCopyPropagation::preorder(IR::AssignmentStatemen
...
+    }
+    // Compare packet header/metadata fields
+    auto l3 = left->to<IR::Member>();
+    auto r3 = right->to<IR::Member>();
+    if (l3 && r3) {
+        return l3->member == r3->member && equiv(l3->expr, r3->expr);
+    } else {
+        return false;
```
mbudiu-vmw: There are other kinds of expressions too, e.g., unary.

midend/local_copyprop.cpp
ChrisDodd: I think this is safe solely because the only `Operation_Binary` that could appear here (lhs of assignment) is `IR::ArrayIndex`.  For generality, should add `&& typeid(*l2) == typeid(*r2)`

midend/local_copyprop.cpp
```diff
@@ -228,6 +260,26 @@ IR::AssignmentStatement *DoLocalCopyPropagation::postorder(IR::AssignmentStateme
     return as;
 }
 
+IR::IfStatement *DoLocalCopyPropagation::postorder(IR::IfStatement *s) {
```
ChrisDodd: I would expect the immediately following call to `ElimDead` to do this?  Are there cases where it does not?

ksubramanianBarefoot: Yes, for `p4_16_samples/hit_ebpf.p4`, the compiler throws a bug of the if statement will a null `IfTrue` expression. The `IfStatement` is validated before the `ElimDead` is called on this object, which results in the compiler exiting with a bug.

midend/local_copyprop.cpp
ksubramanianBarefoot: Could any other `Operation_Unary` other than `IR::Member` appear on the LHS of an assignment?

mbudiu-vmw: The computation of the array index can be an arbitrary expression.

mbudiu-vmw: But you should check whether the side effect pass removes this case. Maybe it cannot happen anymore.

mbudiu-vmw: But you should check whether the side effect pass removes this case. Maybe it cannot happen anymore.

ChrisDodd: The way it's currently set up, expressions with side effectes (MethodCalls -- that's the only kind we have) will always compare as non-equiv, so that shouldn't be an issue here even if side effects didn't first remove them.  I'm not concerned with this `equiv` method getting all cases that are equivalent correct as long as it doesn't incorrectly decide non-equivalent things are equivalent (false negatives are ok, false positives are not).

ksubramanianBarefoot:
Added a more comprehensive equivalence check. `equiv` can check indices of array expressions: for e.g., `arr[x+1] = arr[x+1]`. While `equiv` still does not cover all cases, it will not report any false positives.

midend/local_copyprop.cpp
```diff
@@ -210,32 +210,61 @@ IR::AssignmentStatement *DoLocalCopyPropagation::preorder(IR::AssignmentStatemen
...
+        return ml->member == mr->member && equiv(ml->expr, mr->expr);
+    }
+    // Compare unary operations (can be used inside array indices)
+    auto ul = left->to<IR::Operation_Unary>();
+    auto ur = right->to<IR::Operation_Unary>();
+    if (ul && ur) {
+        return equiv(ul->expr, ur->expr) &&
+        typeid(*ul) == typeid(*ur);
+    }
+    // Compare literals (strings, booleans and integers)
```
ChrisDodd: You could eliminate the need for all these 'leaf' checks by a single check at the top:

    if (*left == *right) return true;

ksubramanianBarefoot: I tried this, but it could not detect `stack[2] = stack[2]`. The definition of == operator for `IR::Expression` is: 

`return typeid(*this) == typeid(a)
        && type == a.type;`

Would this detect equivalence?

mbudiu-vmw: This can handle only "leaf" cases of the recursion, in particular, all Literals would be covered.

ksubramanianBarefoot: `if (*left->to<IR::Node>() == *right)` does not work either, the definition of `==` for `IR::Node` is: 

`return typeid(*this) == typeid(a);`

ChrisDodd: Its virtual, so it does dynamic dispatch based on the dynamic type of *left. That should work without the explicit cast as well (all the operator== functions are virtual) but it doesn't in all cases?

ChrisDodd: Looking at the gtest tests, it is testing for simple `*e1 == *e2` comparison where e1 and e1 are both Expression *, so it should work without the cast.   So ignore my suggestion to use the cast -- it is not needed

PR #732, opened 2017-06-20
Asserts to prevent negative idx and sz parameters in bitvecs.

evanches:


lib/bitvec.cpp
```diff
@@ -36,6 +36,7 @@ std::ostream &operator<<(std::ostream &os, const bitvec &bv) {
 }
 
 bitvec &bitvec::operator>>=(size_t count) {
+    assert(static_cast<int>(count) >= 0);
```
antoninbas: I think `assert(count >= 0u)` looks better

antoninbas: sorry my bad, stupid comment.

ChrisDodd: This is wrong -- we can right shift by arbitrarily large amounts without a problem; if count is more than the largest bit (index) in the bitvec it will just result in a 0 (empty) bitvec and not any sort of problem.

cc10512: Yeah, it's an unsigned. I wonder though if we should set a more constrained limit on the size of bitvec. And we should assert that the return from new in the 'expand' routine is non-null.

lib/bitvec.cpp
```diff
@@ -60,6 +61,7 @@ bitvec &bitvec::operator>>=(size_t count) {
 }
 
 bitvec &bitvec::operator<<=(size_t count) {
+    assert(static_cast<int>(count) >= 0);
```
ChrisDodd: A too-large size should result in a bad_alloc here (from expand), so if we want to catch bad_alloc and deal with it, this might cause problems.

lib/bitvec.cpp
```diff
@@ -78,6 +80,8 @@ bitvec &bitvec::operator<<=(size_t count) {
 }
 
 bitvec bitvec::getslice(size_t idx, size_t sz) const {
+    assert(static_cast<int>(idx) >= 0);
+    assert(static_cast<int>(sz) >= 0);
```
ChrisDodd: calling getslice with `sz == (size_t)-1` is a perfectly valid -- it means get from idx to the end of the bitvec.  Equivalent to making a copy then right-shifting by index.

lib/bitvec.h
```diff
@@ -209,13 +213,16 @@ class bitvec {
             for (size_t i = sz; i < size; i++)
                 ptr[i] = 0; } }
     bool clrbit(size_t idx) {
+        assert(static_cast<int>(idx) >= 0);
```
ChrisDodd: Unnecessary as this one can't cause problems (clearing a bit past the end of the bitvec is a noop

lib/bitvec.h
```diff
@@ -235,8 +242,10 @@ class bitvec {
             if (i < size)
                 ptr[i] &= ~(((uintptr_t)1 << (idx%bits_per_unit)) - 1); } }
     bool getbit(size_t idx) const {
+        assert(static_cast<int>(idx) >= 0);
```
ChrisDodd: Likewise unnecessary

lib/bitvec.h
```diff
@@ -383,8 +393,12 @@ class bitvec {
         return true; }
     bitvec &operator>>=(size_t count);
     bitvec &operator<<=(size_t count);
-    bitvec operator>>(size_t count) const { bitvec rv(*this); rv >>= count; return rv; }
-    bitvec operator<<(size_t count) const { bitvec rv(*this); rv <<= count; return rv; }
+    bitvec operator>>(size_t count) const {
+        assert(static_cast<int>(count) >= 0);
+        bitvec rv(*this); rv >>= count; return rv; }
+    bitvec operator<<(size_t count) const {
+        assert(static_cast<int>(count) >= 0);
```
ChrisDodd: redundant with the asserts in operator >>=/<<=

lib/bitvec.cpp
antoninbas: what is this pull request trying to solve exactly? I'm not very familiar with the `bitvec` class but this check looks a little weird and arbitrary to me. If the limit for these parameters is the maximum positive integer that fits inside an integer, maybe the parameter type should be `int` and the assert should be `assert(count > 0)`. But that seems like an arbitrary limit to me, unless I'm missing something. If the bitvec can be larger than that (unlikely to be useful in the context of this compiler), then this check makes no sense to me.

ChrisDodd: The C++ spec says that operator new cannot ever return NULL (unlike malloc) -- it will throw a bad_alloc exception instead.  If we're seeing NULLs, that is a bug (probably in the garbage collector).

lib/bitvec.cpp
sethfowler: Yeah, I'd expect that `bad_alloc` would be thrown instead of `nullptr` being returned from `new`.

Of course, usually Linux will happily let you allocate a titanic amount of address space and just segfault or OOM-kill you when you actually touch the pages.

evanches:
The SEGFAULT comes from the expand function where the pointers returned from such massive function calls are nullptrs.  Instead of this, I can put an assert around this function instead.  The purpose of this PR was to try to help the debugging of negative parameters in bitvecs, but as you've stated Chris, this makes these functions unclear.

The only difference will be that the compiler will now assert instead of SEGFAULT for invalid bitvec operations, which is a step in the right direction, but doesn't really provide any extra debugging information.

evanches:
I think until we define an arbitrary bitvec size, it's not that important.  I'll just close it.

Issue #658, opened 2017-05-24
Generalize the P4Runtime serializer to support arbitrary backends


sethfowler:
We'd like to generate a P4Runtime API for arbitrary backends, including ones added by extensions. This is currently *close* to working, but there are some BMV2-specific things in the existing code. We should remove them.

mbudiu-vmw:
I don't think this can be done in general.
You can start by trying to adapt it to the ebpf back-end to see if it works.

sethfowler:
The main difficulty I see is that there needs to be a way to describe to the P4Runtime serializer how to deal with backend-specific externs. It's unclear to me to what extent PSA will help with this in the long term. In the short term we'll have to deal with that problem in a somewhat ad hoc manner.

mbudiu-vmw:
The bottom interface is the one I am wondering about, not the top one.
How do you actually perform the operations that the API describes?

sethfowler:
All the P4Runtime serializer does is generate a data structure that describes the structure of the program (which tables exist, which externs are instantiated, etc). That information is consumed by target-specific code that performs communication between the data plane and the control plane. When someone wants to provide P4Runtime support for a target, they need to write that code manually. All of that code lives outside of the compiler, though; in this issue I'm just concerned with generating that data structure, which is the part that the compiler needs to do.

mbudiu-vmw:
Is there a documentation about how to write such a target-specific code?

sethfowler:
That's a good question. @antoninbas, does such documentation exist yet?

antoninbas:
That's a very good question. As Seth writes, I believe the P4Runtime serializer output (which is actually called P4Info) should be completely backend-independent. It is still my belief that we should be able to run it after the frontend, before any midend or backend passes.
[...]

PR #733, opened 2017-06-22
Rewrite isValid() to explicit matches against a $valid$ field in the BMV2 backend

sethfowler:
We want to be able to use the P4Runtime serializer for any architecture and target. (This is issue #658.) There are currently two issues that need to be solved to do that:

1. We need to support extending the P4Runtime serializer with support for architecture-specific externs.
2. We need to remove all BMV2-specific code from the P4Runtime serializer.

This PR solves the second problem. At this point, the only BMV2-specific thing that's left is the implementation of `isValid()` in table keys. The P4Runtime serializer rewrites these to use either the `valid` match type, or to match against the special `$valid$` field that BMV2 adds to header types. This doesn't make sense to handle at that layer.

This PR adds a pass that adds the `$valid$` field to all header types and removes `isValid()` from the program. When `foo.isValid()` is used as a table key component, it's rewritten to `foo.$valid$`, which changes the type to `bit<1>`. In other contexts, it becomes `foo.$valid$ == 1`, which maintains the type of the overall expression. Because of the type change for key components, const table entries must also be rewritten to replace `true` and `false` with `1` and `0`.

This change lets us get rid of the `isValid()`-related code in the P4Runtime serializer. It also lets us remove similar code from the BMV2 JSON generator. Things become significantly simpler.

In addition to the tests in this repo, I tested this change with the `switch.p4` PTF tests. Everything works fine.


backends/bmv2/expression.cpp
```diff
@@ -201,9 +188,9 @@ void ExpressionConverter::postorder(const IR::Member* expression)  {
     if (parentType->is<IR::Type_StructLike>()) {
         auto st = parentType->to<IR::Type_StructLike>();
         auto field = st->getField(expression->member);
-        if (field != nullptr)
-            // field could be a method call, i.e., isValid.
-            fieldName = field->controlPlaneName();
+        BUG_CHECK(field != nullptr,
```
mbudiu-vmw: There are other methods that look like fields, e.g., push.
So maybe this should not be a BUG.

backends/bmv2/synthesizeValidField.cpp
```diff
@@ -0,0 +1,143 @@
...
+    bool preorder(IR::Type_Header* header) override {
+        const cstring validField = V1ModelProperties::validField;
+        if (header->getField(validField) != nullptr) return true;
+        auto field = new IR::StructField(validField, IR::Type::Bits::get(1));
+        header->fields.push_back(field);
+        return true;
+    }
+};
+
+/// Replaces calls to isValid() with references to the header's valid bit.
```
mbudiu-vmw: What about setValid?

backends/bmv2/synthesizeValidField.cpp
```diff
@@ -0,0 +1,143 @@
...
+};
+
+const IR::Node*
+RewriteIsValidCalls::postorder(IR::MethodCallExpression* expr) {
+    auto instance = P4::MethodInstance::resolve(expr, refMap, typeMap);
+    if (!instance->is<P4::BuiltInMethod>()) return expr;
+    auto builtin = instance->to<P4::BuiltInMethod>();
+    if (builtin->name != IR::Type_Header::isValid) return expr;
+
+    // If isValid() is being used as a table key element, it already behaves
```
mbudiu-vmw: Why do you want to make it `bit<1>` instead of bool?
Anyway, if you insist it being a bit, then there is a cast from bit to bool which should be used instead of a comparison.

backends/bmv2/synthesizeValidField.cpp
```diff
@@ -0,0 +1,143 @@
...
+ private:
+    const IR::Node* preorder(IR::P4Table* table) override {
+        using namespace boost;
+
+        // If this table has no key or no const entries, there's nothing to do.
+        auto key = table->getKey();
+        if (key == nullptr) return table;
+        if (table->getEntries() == nullptr) return table;
+
+        // Find the indices of key elements which are simple references to valid
```
mbudiu-vmw: All this would be unnecessary if you don't change the type to bit.

backends/bmv2/synthesizeValidField.cpp
mbudiu-vmw: Note that there are other perfectly legal expressions that evaluate to bool that can be used for table keys, so you still have to handle the other cases in the back-end.

backends/bmv2/expression.cpp
sethfowler: OK, I can change that.

backends/bmv2/synthesizeValidField.cpp
sethfowler: See the comments in the header. Currently we can't do that, because `setValid()` (when used in an action, at least) compiles down to the `add_header` primitive on BMV2. `add_header` does more than just set `$valid$` to 1 - for example, it's used in the implementation of header unions.

mbudiu-vmw: isValid is also defined for unions.

backends/bmv2/synthesizeValidField.cpp
sethfowler: I actually did that because the P4-16 spec says that you can't have a boolean header field. However, now that I look, the BMV2 JSON generator already supports that, so perhaps it's not such a big deal. Making it a boolean field would eliminate the need for rewriting const table entries, which would be nice.

mbudiu-vmw: The IR can represent programs that the spec disallows.
But indeed, the type-checker may be upset if it finds a boolean header field.

sethfowler: Yeah, I just tried it and it doesn't pass the type checker. We could always loosen that check, though. I don't feel strongly either way, but I suppose that rewriting to something that's as close to valid P4 code as possible has some advantages.

backends/bmv2/synthesizeValidField.cpp
sethfowler: Ack, that's a potential problem! Let me take a look at that; I may need to handle that case differently.

sethfowler: So I investigated this, and I don't think we're generating correct BMV2 JSON for header unions as it is. The `valid` match type won't work for header unions; we should be making use of the `valid_union` primitive. `valid_union` can't be used in table match keys (since they can't be general expressions) so what we need to do is use `valid_union` to set a scalar and use the scalar in the key. I think the right fix here is:

1. Don't remove `isValid()` calls on header unions in `SynthesizeValidField`. Since all other `isValid()` calls will be removed, these will be the only `isValid()` calls left.
2. Modify the `KeyIsComplex` policies to treat `isValid()` calls as complex, so they're moved out of match keys. (Or perhaps, for now, just add a new policy, and remove the old ones later once we're sure all backends have migrated over.)
3. In the backend, generate the `valid_union` primitive when we encounter `isValid()` calls on header unions.

Should all be pretty straightforward.

mbudiu-vmw: Yes, we have a very broad issue covering the support of header unions in the back-end, which is very incomplete. I am working on this right now, but I don't know how much I will have time to finish before going on vacation. I will add the KeyIsComplex to the patch I working on now.

sethfowler: Actually I already wrote the patch. =) I needed it as part of this PR. I'll add it in an additional commit, along with a test that exercises this functionality.

sethfowler:
My thinking was that it didn't matter because field names aren't first class in P4Runtime.

It hasn't happened yet because we've had our heads down with other things, but I still expect us to add a P4-16 language feature to provide explicit names for match key components. With that in place, the idea is that you'd either provide an explicit or accept that the compiler will generate some arbitrary name for you. That's the same way we handle any other complex expression in a match key.

If there are problems with that approach, though, we can definitely revisit it.

antoninbas:
I think the name should be the same across all possible targets. For `isValid` it would be best IMO if it wasn't too arbitrary as I expect it to be pretty commonly used in match keys. So it would be nice if the name was predictable and not too shady. `<hdr>.$valid$` is not too bad. I would expect P4Runtime consumers to manipulate this name directly. A fully arbitrary name makes the annotation a requirement and I think we should avoid that and have a nice default name.

That's a secondary issue until we can find a way to truly make P4Runtime serializer target-independent by moving it up in the pass sequence.

sethfowler:
Sounds reasonable to me. We can standardize on `$valid$`, then, unless we find a reason to use something else.

For the general problem of making P4Runtime's output totally target independent, we're going to need to tackle the issue of control plane names more holistically in the P4-16 language itself. The good news is that we've come up with a lot of ideas for doing that; the bad news is that nobody's had time to finalize and implement them yet. =)

sethfowler:
@mbudiu-vmw could you please take a look one more time? Things are mostly the same, but I believe I've addressed your concern with how header unions are handled, and d40d956 implements the change we discussed that moves `isValid()` calls out of match keys in SimplifyKey.

I've also added a new integration test for matching against `isValid() on header unions (which would have caught the bug) and a reasonably thorough unit test suite for SynthesizeValidField.

mbudiu-vmw:
Note that I also submitted a PR which changes the key policy: #735

backends/bmv2/control.cpp
```diff
@@ -342,32 +326,18 @@ ControlConverter::convertTable(const CFG::TableNode* node,
...
-                                        backend->getCoreLibrary().ternaryMatch.name;
-                            } else {
-                                expr = bim->appliedTo; }}}}
-            }
 
             auto keyelement = new Util::JsonObject();
             keyelement->emplace("match_type", match_type);
-            auto jk = conv->convert(expr);
+
+            // XXX(seth): Boolean variables are stored as data, so we normally
```
mbudiu-vmw: Why the XXX?

backends/bmv2/synthesizeValidField.cpp
```diff
@@ -0,0 +1,151 @@
...
+
+    // If isValid() is being used as a table key element, it already behaves
+    // like a bit<1> value; just replace it with a reference to the valid bit.
+    const cstring validField = V1ModelProperties::validField;
+    auto member = new IR::Member(expr->srcInfo, builtin->appliedTo, validField);
+    if (getParent<IR::KeyElement>() != nullptr) return member;
+
+    // In other contexts, rewrite isValid() into a comparison with a constant.
+    // This maintains a boolean type for the overall expression.
+    auto constant = new IR::Constant(IR::Type::Bits::get(1), 1);
```
mbudiu-vmw: Why not Cast to bool - it would generate more efficient code in most cases it would be a noop, unlike the comparison.

backends/bmv2/synthesizeValidField.h
```diff
@@ -0,0 +1,92 @@
...
+ * However, when `foo.isValid()` is a table key element (and isn't part of a
+ * larger expression), it's replaced with a simple match against `foo.$valid$`
+ * itself. This changes the type of the key element from boolean to bit<1>;
+ * to accomodate that change, the corresponding constant table entry keys are
+ * also rewritten to use bit<1> values instead of booleans.
+ *
+ * Calls to isValid() on header unions aren't rewritten; they can't be desugared
+ * to a simple read of a hidden field.
+ *
+ * XXX(seth): It would be cleaner to rewrite setValid() and setInvalid() too,
```
mbudiu-vmw: The XXX is probably not warranted.

backends/bmv2/control.cpp
sethfowler: Because the name "convertLeftValue()" doesn't really capture the reason we're calling the function. It needs a better name, or perhaps we need a different function with a clearer name, but I couldn't immediately think of one.

backends/bmv2/synthesizeValidField.h
sethfowler: I actually do think we're eventually going to be able to do this. It requires some BMV2 changes and they aren't in place yet, but the documentation (search for `$valid$` [here](https://github.com/p4lang/behavioral-model/blob/master/docs/JSON_format.md)) suggests they will probably arrive eventually.

backends/bmv2/synthesizeValidField.cpp
sethfowler: That's a good point. I'll make that change.

sethfowler: Hmm, I just tried this and while the compiler is happy, it breaks several STF tests. The cast leads us to generate some code that malfunctions at runtime.

I've spent too long on this patch already so I'm going to go ahead and stick with the comparison for now, but I'll file a followup bug to investigate the issue.

PR #735, opened 2017-06-23
Better complex key policy; more tests for header unions

mbudiu-vmw:


sethfowler:
As you mentioned in the comments on #733, the changes in `simplifyKey.cpp` may not be needed anymore since that PR has landed.

Really glad to see more header union tests; it became clear to me while working on #733 that we really need them. =)

mbudiu-vmw:
There are no conflicts, it looks like my changes to the KeyIsComplex policies are independent of yours.
We don't yet have support for arrays of unions.

PR #736, opened 2017-06-23
Speed up the build by reducing the amount of headers we're including

sethfowler:
Here's a patch I've had lingering for a while now that speeds up the build a little bit by moving some code from `.h` files to `.cpp` files. The biggest win is moving some code out of `ir-inline.h`; at least on clang, that code is pretty expensive to compile.

In my environment, a full rebuild of master (`time make -j8`) takes this long:

```
107.69 real 623.43 user 22.66 sys
```

With this patch:

```
95.80 real 544.69 user 22.00 sys
```

That's an 11% speedup. Not bad.

### DIFF
--- a/ir/visitor.cpp
+++ b/ir/visitor.cpp
@@ -355,6 +355,7 @@ void Transform::revisit(const IR::CLASS *o, const IR::Node *n) {                
     return revisit(static_cast<const IR::BASE *>(o), n); }                              \
 
 IRNODE_ALL_SUBCLASSES(DEFINE_VISIT_FUNCTIONS)
+#undef DEFINE_VISIT_FUNCTIONS
 
 class SetupJoinPoints : public Inspector {
     map<const IR::Node *, std::pair<ControlFlowVisitor *, int>> &join_points;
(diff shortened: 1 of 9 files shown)


mbudiu-vmw:
Does this affect the compiler running time (by reducing inlining?)

sethfowler:
Since most people working on the compiler are building at `-O0`, the answer is no. =)

If we want to distribute a fast executable in binary form, we should just build with LTO anyway.

mbudiu-vmw:
This looks fine to me, but I think that @ChrisDodd should review this.

ir/visitor.cpp
```diff
@@ -416,3 +417,49 @@ ControlFlowVisitor &ControlFlowVisitor::flow_clone() {
     assert(rv->check_clone(this));
     return *rv;
 }
+
+#define DEFINE_APPLY_FUNCTIONS(CLASS, TEMPLATE, TT)                                             \
+    TEMPLATE bool IR::CLASS TT::apply_visitor_preorder(Modifier &v)                      \
```
ChrisDodd: I don't like duplicating this big macro in both the header and the .cpp file -- need to define it once in the header and NOT `#undef` it (with a comment about leaving it valid so it can be expanded in visitor.cpp).  This does pollute the global namespace a bit, but that is unavoidable.

sethfowler: They're actually not identical, alas, because one macro includes the `inline` keyword and the other doesn't. I can set up something to eliminate the duplication, though.

PR #739, opened 2017-06-23
Add preliminary support for learning to p4runtime-to-pd

sethfowler:
This PR uses the newly-introduced `bm_get_id_from_name()` API to add support for learning to the PD frontend that `p4runtime-to-pd` generates.

Unfortunately, we still can't pass the learning PTF test for switch.p4. We get the correct learn quanta id, and everything appears to work fine, but we end up receiving a null `msg` in `notify_ack`. To avoid a crash, I've handled this case, but we'll need to debug it before the PTF test will pass.

With this change, though, the switch.p4 PTF tests at least all run without crashing the driver, so we can start using them.

### DIFF
--- a/p4runtime-to-pd/templates/src/pd.cpp
+++ b/p4runtime-to-pd/templates/src/pd.cpp
@@ -33,6 +33,7 @@ extern "C" {
 
 void ${pd_prefix}learning_notification_cb(const char *hdr, const char *data);
 p4_pd_status_t ${pd_prefix}learning_new_device(int dev_id);
+p4_pd_status_t ${pd_prefix}learning_setup_device(int dev_id);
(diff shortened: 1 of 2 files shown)


mbudiu-vmw:
Is there a plan to add the ptf tests to the regression tests for the compiler?
They take a long time and have additional dependences.
But it would be nice to have a simple way to run them.

sethfowler:
We could eventually add them to `make check`. They don't actually take that long to run, but building the PD frontend for a P4 program is pretty slow. Fortunately, we'd only really want to do it for one program (switch.p4) and it doesn't change very frequently.

The bigger problem is that, as you said, it adds more dependencies - a *lot* of them. In the short term, I'm probably going to avoid that by running the PTF tests inside a docker container. I'll set it up so that's easy to do, though, so you can run them locally if you want.

PR #740, opened 2017-06-26
Fix typechecking MakeImplicitCastExplicit to work properly with dags

ChrisDodd:
So using `visitDagOnce = false` in a Transform (or Modifier) seems to be a sure source of hard to track bugs.  Its far too easy for the Transform to inadvertently clone something that is referenced mulitple places, which seems to be ok at first, but leads to problems down the line.

This change dusts off an old change I've had sitting on a branch for awhile to add some new Visitor functions (`visitAgain()` and `visitOnce()`) that override the `visitDagOnce` flag on a per-node basis.  Any pre- or postorder function can call `visitAgain()` and it result in node being visited again if it is found as a child elsewhere in the IR Dag instead of reusing the results of the first call (as if `visitDagOnce = false`, but just for that node).  Similarly `visitOnce()` treats the node as if `visitDagOne = true` (in case you set it to `false` globally in the constructor.

This allows the fix for the P4_14 typechecker to insert casts on ActionArgs indepedently each time they appear in the tree while NOT duplicating extern object instances that are referenced multiple times.

This also adds the `int` type to the P4_16 parser (for integer constant of unspecified size in a constructor) so that P4_14 externs that use `int` can be converted to P4_16 without causing syntax errors in the result.

### DIFF
--- a/frontends/p4-14/typecheck.cpp
+++ b/frontends/p4-14/typecheck.cpp
@@ -249,10 +249,12 @@ class TypeCheck::InferExpressionsBottomUp : public Modifier {
             auto *lt = op->left->type->to<IR::Type::Bits>();
             auto *rt = op->right->type->to<IR::Type::Bits>();
             if (lt && rt) {
-                if (lt->size < rt->size)
+                if (lt->size < rt->size) {
+                    setType(op, rt);
                     op->left = new IR::Cast(rt, op->left);
-                else if (rt->size < lt->size)
-                    op->right = new IR::Cast(lt, op->right);
+                } else if (rt->size < lt->size) {
+                    setType(op, lt);
+                    op->right = new IR::Cast(lt, op->right); }
                 LOG3("Inserted cast in " << op);
             }
         }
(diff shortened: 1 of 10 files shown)


ChrisDodd:
Yes exactly.  This allows a visitor to control which nodes will be visited again and which nodes will not on a per-node basis.  So if node A has 2 parents and set visitAgain it will be visited twice.  If it's child B does not call visitAgain, then B will only be visited once (between the preorder and postorder of the first visit of A, and on the second visit of A it will *not* be visited again -- the result of the first visit of B will be substituted into B automatically.

frontends/parsers/p4/p4parser.ypp
```diff
@@ -602,6 +602,7 @@ baseType
     | BIT    { $$ = IR::Type::Bits::get(@1, 1); }
     | BIT "<" INTEGER ">"
       { $$ = IR::Type::Bits::get(@3, parseConstant(@3, $3, 0)->asInt(), false); }
+    | INT { $$ = new IR::Type_InfInt; }
```
mbudiu-vmw: Should we add this to the spec?

sethfowler:
(Sorry I didn't look at this quicker; fell a bit behind on my email this week.)

This seems like a good improvement. I agree with Mihai that this is subtle behavior, but it's probably the best that can be done without a more drastic change. I wonder if we could implement a model where the user expresses on a per-node basis whether they want the result of multiple visits to the same node to always result in the same object. We could verify that equivalent objects are always returned in that situation (treating it as a BUG if not) and then perform the deduplication. That might make things a little more intuitive, but without trying it out I'm not sure whether it'd be a win.

mbudiu-vmw:
I don't think you ever need this cast inside an expression, this is just to make the bmv2 typechecker happy.

mbudiu-vmw:
My last comment is on the wrong thread. Please ignore it

PR #745, opened 2017-06-28
Annotate bmv2 $valid$ fields as hidden

antoninbas:
Otherwise they can show up in the p4Runtime output.

### DIFF
--- a/backends/bmv2/synthesizeValidField.cpp
+++ b/backends/bmv2/synthesizeValidField.cpp
@@ -37,7 +37,9 @@ class AddValidField final : public Modifier {
     bool preorder(IR::Type_Header* header) override {
         const cstring validField = V1ModelProperties::validField;
         if (header->getField(validField) != nullptr) return true;
-        auto field = new IR::StructField(validField, IR::Type::Bits::get(1));
+        auto annotations = new IR::Annotations(
+            {new IR::Annotation(IR::Annotation::hiddenAnnotation, {})});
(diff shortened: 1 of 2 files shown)


backends/bmv2/synthesizeValidField.cpp
```diff
@@ -37,7 +37,10 @@ class AddValidField final : public Modifier {
     bool preorder(IR::Type_Header* header) override {
         const cstring validField = V1ModelProperties::validField;
         if (header->getField(validField) != nullptr) return true;
-        auto field = new IR::StructField(validField, IR::Type::Bits::get(1));
+        auto annotations = new IR::Annotations();
+        annotations->add(new IR::Annotation(IR::Annotation::hiddenAnnotation,
+                                            IR::Vector<IR::Expression>()));
```
ChrisDodd: The above 3 lines can be more compactly written as

    auto annotations = new IR::Annotations({new IR::Annotation(IR::Annotation::hiddenAnnotation, {})};

or even

    auto annotations = new IR::Annotations({new IR::Annotation("hidden", {})};

sethfowler:
I was worried that this would break PD generation, but since it only applies to `controller_header` it's OK. We cannot ignore `$valid$` fields in other situations (e.g. match keys), obviously.

I'm really looking forward to having some free time to make the PTF tests start running on Travis; it's fairly easy to break right now since there's no CI testing for it anywhere.

PR #749, opened 2017-06-29
Include SourceInfo in the JSON output of p4test.

orenbenkiki:
This patch embeds the SourceInfo data in the p4test JSON output.
This makes it accessible to any post-processing tool using this output.
The format of data is compatible with Andy's patch (which is only used in the BMV2 backend).

The commit also includes a minor tweak - cleaning up the trailing space following the "," in JSON nodes.

### DIFF
--- a/ir/node.cpp
+++ b/ir/node.cpp
@@ -25,7 +25,7 @@ void IR::Node::traceCreation() const { LOG5("Created node " << id); }
 int IR::Node::currentId = 0;
 
 void IR::Node::toJSON(JSONGenerator &json) const {
-    json << json.indent << "\"Node_ID\" : " << id << ", " << std::endl
+    json << json.indent << "\"Node_ID\" : " << id << "," << std::endl
          << json.indent << "\"Node_Type\" : " << node_type_name();
 }
 
@@ -67,10 +67,9 @@ cstring IR::dbp(const IR::INode* node) {
     return str.str();
 }
 
-Util::JsonObject* IR::Node::sourceInfoJsonObj() const {
-    unsigned lineNumber, columnNumber;
-
-    Util::SourceInfo si = srcInfo;
+cstring IR::Node::prepareSourceInfoForJSON(Util::SourceInfo& si,
+                                           unsigned *lineNumber,
+                                           unsigned *columnNumber) const {
     if (!si.isValid()) {
         return nullptr;
     }
@@ -78,7 +77,15 @@ Util::JsonObject* IR::Node::sourceInfoJsonObj() const {
         auto assign = to<IR::AssignmentStatement>();
         si = (assign->left->srcInfo + si) + assign->right->srcInfo;
     }
-    cstring fName = si.toSourcePositionData(&lineNumber, &columnNumber);
+    return si.toSourcePositionData(lineNumber, columnNumber);
+}
+
+// TODO: Find a way to eliminate the duplication below.
+
+Util::JsonObject* IR::Node::sourceInfoJsonObj() const {
+    Util::SourceInfo si = srcInfo;
+    unsigned lineNumber, columnNumber;
+    cstring fName = prepareSourceInfoForJSON(si, &lineNumber, &columnNumber);
     if (fName == nullptr) {
         // Do not add anything to the bmv2 JSON file for this, as this
         // is likely a statement synthesized by the compiler, and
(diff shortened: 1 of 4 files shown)


ir/node.cpp
```diff
@@ -94,4 +102,24 @@ Util::JsonObject* IR::Node::sourceInfoJsonObj() const {
     return json;
 }
 
+void IR::Node::sourceInfoToJSON(JSONGenerator &json) const {
+    Util::SourceInfo si = srcInfo;
+    unsigned lineNumber, columnNumber;
+    cstring fName = prepareSourceInfoForJSON(si, &lineNumber, &columnNumber);
```
orenbenkiki: I wondered about this adjustment. I refactored it out of the `sourceInfoJsonObj` code. If I switch to using `toSourcePositionData` in `sourceInfoToJSON`, what should I do in `sourceInfoJsonObj`? I thought it makes sense for both of them to be consistent.

ir/node.cpp
```diff
@@ -94,4 +102,24 @@ Util::JsonObject* IR::Node::sourceInfoJsonObj() const {
...
+        return;
+    }
+
+    json << "," << std::endl
+         << json.indent++ << "\"Source_Info\" : {" << std::endl;
+
+    json << json.indent << "\"filename\" : " << fName << "," << std::endl;
+    json << json.indent << "\"line\" : " << lineNumber << "," << std::endl;
+    json << json.indent << "\"column\" : " << columnNumber << "," << std::endl;
+    json << json.indent << "\"source_fragment\" : " << si.toBriefSourceFragment() << std::endl;
```
orenbenkiki: Refactoring it to an `operator<<` makes sense.

As for including the `end` coordinates, I didn't do it for the following reasons:

- The main reason for including `srcInfo` in the output is to help with error messages in post processing tools. The common practice in UNIX tools is to specify something in the format `filename:line` or, for extra credit, `filename:line:column`. Ranges are very rarely used and AFAIK there isn't even a standard format for specifying them in a way that IDEs etc. will understand.

- The  `sourceInfoJsonObj` also only printed the start position, and I wanted to be consistent with it. As a bonus, it includes a `source_fragment`, which would be useful in printing an improved error message without having to include an end position. 

I suggest that these are sufficient reasons for avoiding including an explicit end position in the output. That said, if you feel strongly that both should be included, should it also be done for the `sourceInfoJsonObj`, then?

ir/node.cpp
orenbenkiki: Refactoring it to an `operator<<` makes sense.

As for including the `end` coordinates, I didn't do it for the following reasons:

- The main reason for including `srcInfo` in the output is to help with error messages in post processing tools. The common practice in UNIX tools is to specify something in the format `filename:line` or, for extra credit, `filename:line:column`. Ranges are very rarely used and AFAIK there isn't even a standard format for specifying them in a way that IDEs etc. will understand.

- The  `sourceInfoJsonObj` also only printed the start position, and I wanted to be consistent with it. As a bonus, it includes a `source_fragment`, which would be useful in printing an improved error message without having to include an end position. 

I suggest that these are sufficient reasons for avoiding including an explicit end position in the output. That said, if you feel strongly that both should be included, should it also be done for the `sourceInfoJsonObj`, then?

Idea: We could emit the unadjusted position, and only used the adjusted position for computing the `source_fragment` - and use this combination in both `sourceInfoToJSON` and `sourceInfoJsonObj`. Would that be acceptable?

ir/node.cpp
ChrisDodd: Part of the problem is that we're not including the entire source code in the json, so just having start/end positions is not enough; we need the source fragments too.  But with the source fragments, you're right, we probably don't need the end points.  Of course, the source code could readily be packaged separately from the json -- with the start/end positions, a tool could easily extract the fragments from the source code itself.  No good answers.

ir/node.cpp
ChrisDodd: Alternately if what makes sense for the 'location' of an assignment is the whole thing rather than just the `=` in the assignment, we should change what the parser puts in the srcInfo for an assignment statement.  Lots of different ways to skin this cat.

jafingerhut:
I don't know if I will be any help here, but can at least shed some light on what I was thinking when adding source info to the bmv2 JSON configuration files output by p4c-bm2-ss.

My main goal was to enable simple_switch to produce extra log messages that indicated the file name and line number, and sometimes a source fragment, in at least the following 2 scenarios:

(a) when a condition was executed, I wanted to be able to quickly find that condition in the source code, and preferably see it in the log if it fit on a single line, and whether it evaluated to true or false.

(b) when a primitive inside of an action was executed, I wanted to be able to quickly find that primitive in the source code, and preferably see it in the log if it fit on a single line.

I was not trying to solve the world, but mainly get those use cases working.  For (b) with P4_16 source programs, the assignment statement source info was initially just the '=' symbol in the source code, which isn't informative as a source fragment in a log file, when you would prefer to see 'dest = expression;' instead.  So I added a special case for assignment statements that included the entire assignment statement, including left- and right-hand sides, not just the '='.

For those use cases, the code does what it needs to.  It might not meet the needs of other use cases.

If you have use cases for this extra source data, I would recommend trying to identify them, think about what is needed for them, and go from there.  Sorry, I know that is fairly general advice, but it is also the best advice I can think of.

PR #751, opened 2017-06-29
fix issue750

hanw:


sethfowler:
Mihai should probably review this one.

IMO, the way we're handling boolean conversion here is just not sound. In the contexts where we want to insert a `b2d` cast, we set a flag at the top level which causes us to do that for any boolean value in any subexpression, but that's not correct; we generally only want to do it at the top level.

At some point, we'll need to come up with a better method.

PR #755, opened 2017-07-02
adds cmake build system

cc10512:
This PR adds an alternate build system based on cmake. I kept it compatible with the current automake system, although, there are a number of places where we can do things a little better when we deprecate automake.


ChrisDodd:
We switched to automake in the first place because it was seen as the most portable for an open-source project.  If we want to abandon automake, why not just go back to GNU make which is much more powerful, flexible, and easier to use than cmake?

cc10512:
Because make lacks support for testing. With cmake we get CTest which is a very nice testing framework that allows us to do negative tests, check for specific error messages on a per-test basis, pass arguments to tests to individual tests, and so on. 

You may notice, cmake also it removes a need of the "helper" scripts that were put in place for getting automake to configure everything properly.

cmake is supported on all platforms that we are building on, and more. It actually works with Windows as well, and if you fancy Xcode, there is a generator for that too. It also supports packaging, with a few lines you can get an installable Debian package (or RPM, or DMG, etc.). While it is still fairly actively developed, it is stable and handles composing projects graciously.

ChrisDodd:
So I was able to get&build the branch with cmake, but running tests is a failure -- 561 out of 798 tests fail.  The failures all seem to be related to being unable to find v1model.p4 and/or core.p4, though it also gives messages "*** config.h indicates that BMv2 is not installed; will skip running BMv2 tests", even though BMv2 is installed.

cc10512:
For BMv2 I have a potential explanation: it looks for it in the path and in the install directory. It will not find it in a directory that is somewhere above the current directory.

For the header files, I have put in hooks to copy the headers in the right place. And I thought that the p4c_driver rule is always invoked. It also seem to have worked on Travis for all 798 tests. I wonder what in your env causes it to fail. Do you have the p4include directory in the same place with the executable?

ChrisDodd:
The BMv2 binaries (simple_switch and simple_switch_CLI) are installed in $PATH, it just fails to find them, apparently.  The build correctly puts the p4include files in the build directory (under p4include), but the compiler fails to find them there.

Tracking what is going on is hard because the cmake build apparently always builds without debug info (or at least there's no debug info that gdb can find).

hanw:
I did the following and all 798 tests passed in os x.
```
cd ~/p4c
mkdir build
cd build
cmake ..
make
ctest
```
Compiling or running ctest outside the `build` directory does not work.

ChrisDodd:
So the include failure problem is due to the fact that the p4include directory is under build, while all the executables are under build/backends -- so it looks for build/backends/bmv2/p4inlcude and fails to find it.  Adding a symlink there to ../../p4include fixes the problem.

It still can't seem to find the installed simple_switch and simple_switch_CLI, so won't run bmv2, though looking in the parent directory for a more recent built (but not installed) bmv2 would be preferable.

ChrisDodd:
Its also just the cmake/config process that fails to find bmv2 -- if I manually add `#define HAVE_SIMPLE_SWITCH 1` to config.h, it runs the bmv2 tests just fine.

README.md
```diff
@@ -194,6 +206,102 @@ default is 2GB, which is not enough to build p4c. Increase the memory limit to
...
+
+where `mybackend` is the name of the directory you added under extensions.
+The p4c CMakeLists.txt will use that name to figure the full path of the files to lint.
+
+## Target
+
+Define a target for your executable. The target should link against
+the core `P4C_LIBRARIES` and `P4C_LIB_DEPS`.  `P4C_LIB_DEPS` are
+package dependencies. If you need additional libraries for your
+project, add them to `P4C_LIB_DEPS`.
```
sethfowler: That's presumably just if you need additional libraries that are referenced by your IR classes, right? Other stuff should only be linked in for the backend that needs it.

README.md
```diff
@@ -194,6 +206,102 @@ default is 2GB, which is not enough to build p4c. Increase the memory limit to
...
+set (mybackend_CPPLINT_FILES ${MY_SOURCES_AND_HEADERS} PARENT_SCOPE)
+```
+
+where `mybackend` is the name of the directory you added under extensions.
+The p4c CMakeLists.txt will use that name to figure the full path of the files to lint.
+
+## Target
+
+Define a target for your executable. The target should link against
+the core `P4C_LIBRARIES` and `P4C_LIB_DEPS`.  `P4C_LIB_DEPS` are
```
sethfowler: Do we actually need to specify `P4C_LIB_DEPS` directly? I'd expect cmake to support transitive dependencies...

cmake/P4CUtils.cmake
```diff
@@ -0,0 +1,76 @@
...
+#   - driver is the script that is used to run the tests and compare the results
+#   - testsuite is a list of directory patterns, e.g.: testdata/p4_16_samples/*.p4
+#   - xfail is a set of tests that are expected to fail
+#
+# The macro generates the test files in a directory prefixed by tag.
+#
+macro(p4c_add_tests tag driver testsuites xfail)
+  set(__xfail_list "${xfail}")
+  foreach(ts "${testsuites}")
+    file (GLOB __testfiles RELATIVE ${P4C_SOURCE_DIR} ${ts})
```
sethfowler: Do you need to rerun cmake if you add a test? I'm not sure how smart its dependency management is in this case. (Not that it's a huge deal, since once we fully switch over to cmake we can implement this stuff in a more cmake-native way.)

CMakeLists.txt
```diff
@@ -0,0 +1,273 @@
...
+set (P4C_LIBRARIES controlplane midend frontend ir p4ctoolkit)
+
+# add extensions - before the frontends as they produce IR and extra frontend
+set(EXTENSION_FRONTEND_SOURCES)
+file (GLOB p4c_extensions RELATIVE ${P4C_SOURCE_DIR}/extensions ${P4C_SOURCE_DIR}/extensions/*)
+MESSAGE ("-- Available extensions ${p4c_extensions}")
+foreach (ext ${p4c_extensions})
+  if (EXISTS ${P4C_SOURCE_DIR}/extensions/${ext}/CMakeLists.txt)
+    add_subdirectory (extensions/${ext})
+    set (__cpplint_list "${${ext}_CPPLINT_FILES}")
```
sethfowler: There's a lot of gross machinery around cpplint in here, and unfortunately we will need similar stuff for some other cases as well. I wonder if we shouldn't just use the idiom we're using for `IR_DEF_FILES` for cpplint too. Something even cleaner might be to use source file properties, but I'm not sure if there are scoping issues there that would prevent us from using them.

README.md
cc10512: libraries do not currently depend on each other.

README.md
cc10512: correct.

cmake/P4CUtils.cmake
cc10512: You do need to re-run cmake. In my sandbox it does not rebuild anything else and just adds the test.

README.md
sethfowler: That's not true. For example, `libfrontend.la` depends on both `libcontrolplane.la` and `libp4ctoolkit.la`. `libcontrolplane.la` in turn depends on the Protobuf libraries. Handling that transparently is the reason we're using libtool in the current build system.

Dockerfile
```diff
@@ -39,8 +40,9 @@ WORKDIR /p4c/
 RUN apt-get update && \
     apt-get install -y --no-install-recommends $P4C_DEPS $P4C_RUNTIME_DEPS && \
     pip install tenjin && \
-    ./bootstrap.sh --enable-p4runtime-to-pd && \
+    mkdir build && \
     cd build && \
+    cmake .. && \
```
sethfowler: Since writing this comment, some of these issues have been addressed: the patch now includes GTest support, and while it does regress compilation time, it speeds up the tests, resulting in the overall time required for a Travis run being about the same.

In the interests of allowing the state of the two build systems to drift apart, it'd make life easier to switch over to cmake as soon as possible. I'd still recommend that we don't do it as part of this PR, though. Let's land this, then open a new PR to make the changeover official (i.e., change the build system we use on Travis). We can wait to land that PR until everyone has had a chance to try the new build system out locally and see if any issues crop up.

cc10512: I reverted that commit.

cc10512: Although obviously, Travis no longer tests this change!!

cc10512:
@ChrisDodd, please take another look. I logs the XFAIL tests in LastXfail.log. All the logs are in build/Testing/Temporary/Last*.log -- output, list of failed, list of xfailed. Let me know if this works for you.

Issue #530, opened 2017-04-26
Enhancement: program that accepts all legal P4_16 programs


jafingerhut:
p4test is pretty close to this already, but for various reasons it does not accept some syntactically and semantically legal P4_16 programs as of 2017-Apr-26, e.g. see https://github.com/p4lang/p4c/issues/513

Given how close p4test is to this, one possible implementation might be to add a new p4test command line option that changes its behavior somewhat, enabling it to accept all legal P4_16 programs, e.g. --validate

mbudiu-vmw:
The only thing needed is to stop compilation after the front-end is finished.

jafingerhut:
Sorry for my ignorance in asking, but is there already a p4test command line option that achieves this, and if so, what is it?

mbudiu-vmw:
No. But that would be the fix.

jafingerhut:
Tested with new --validate option on the same program on issue #513 that causes p4test to issue an error, and it gives no error.  Thanks!

PR #757, opened 2017-07-03
Add --parse-only option to p4test.

orenbenkiki:
This allows p4test to emit the JSON for the full unprocessed P4 source.
This JSON is useful for tools that would like to deal directly with the source code (e.g. IDEs).

For complete functionality, the JSON should include SourceInfo data (pull request #749).

### DIFF
--- a/backends/p4test/p4test.cpp
+++ b/backends/p4test/p4test.cpp
@@ -31,6 +31,18 @@ limitations under the License.
 #include "frontends/p4/toP4/toP4.h"
 #include "midend.h"
 
+class P4TestOptions : public CompilerOptions {
+ public:
+    bool parseOnly = false;
+    P4TestOptions() {
+        registerOption("--parse-only", nullptr,
+                       [this](const char*) {
+                           parseOnly = true;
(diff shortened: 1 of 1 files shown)


jafingerhut:
Looks like this might implement the enhancement suggested in this issue, too: https://github.com/p4lang/p4c/issues/530

orenbenkiki:
With regard to #530, the PR here makes `--parse-only` skip _all_ processing, including the front end; that is, it accepts invalid P4 as well as valid P4. This is an advantage when trying to feed meta-data about the source code into external tools.

I'm not sure I follow the logic in the #530 discussion; I thought the purpose of `p4test` (without the `-parse-only` flag) was to validate the P4 code. I'm not sure what the point is of running only the front-end without the mid-end.

Perhaps what is needed is a more nuanced flag which specifies whether to perform just the parsing, the parsing + the front end, or the full parsing + front end + mid end? I'm not sure what the exact form of such a flag would be, though. At any rate, if such a flag is added, then `--parse-only` would be equivalent to specifying it with a minimal value.

PR #758, opened 2017-07-03
Rework ExternConverter to be simpler and more flexible

ChrisDodd:
This reworks my previous ExternConverter change to be both simpler and more flexible.  Now instead of having a single ExternConverter passed into the frontend, you can register multiple ExternConverter subclass instances on a per extern type basis -- the converter will manage them and call the correct one.  This allows better mixing and matching of extern converters for related targets.  Most of the change is actually undoing the previous change to the frontend API.

There's also yet another P4_14 typechecking fix with getting the right implicit conversions inserted as explicit conversions.

### DIFF
--- a/frontends/p4/fromv1.0/converters.cpp
+++ b/frontends/p4/fromv1.0/converters.cpp
@@ -370,13 +370,13 @@ class FixupExtern : public Modifier {
 };
 }  // end anon namespace
 
-const IR::Type_Extern *
-ExternConverter::convertExternType(const IR::Type_Extern *ext, cstring name) {
+const IR::Type_Extern *ExternConverter::convertExternType(ProgramStructure *structure,
+            const IR::Type_Extern *ext, cstring name) {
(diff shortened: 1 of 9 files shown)


ChrisDodd:
BTW I plan on a mechanism similar to this for adding more per-targer pimitive conversions to the P4_14->16 conversion.

frontends/p4/fromv1.0/converters.cpp
```diff
@@ -399,7 +399,20 @@ ExternConverter::convertExternCall(const IR::Declaration_Instance *ext, const IR
     return new IR::MethodCallStatement(prim->srcInfo, mc);
 }
 
+std::map<cstring, ExternConverter *> *ExternConverter::cvtForType = nullptr;
 
+void ExternConverter::addConverter(cstring type, ExternConverter *cvt) {
```
cc10512: What is type? Extern name or extern type name? How qualified? It would be helpful to add a comment and an example.

PR #765, opened 2017-07-07
extensible PrimitveConverter framework

ChrisDodd:
Initial ideas for an extensible primitive P4_14->16 conversion framework.  The goal is to be able to easily support target architecture specific primitives, rather than having everything ties to v1model.

This ends up requiring some duplication of code in the conversion routines (for example, all of the simple binop conversions); it would be nice to figure out a way to reduce that.

This also has the disadvantage of being statically built into the executable, which will make it difficult to support mutiple targets in a single executable.  At the moment we're using separate executables for each target, so this is not an issue.

### DIFF
--- a/frontends/p4/fromv1.0/converters.cpp
+++ b/frontends/p4/fromv1.0/converters.cpp
@@ -416,6 +416,44 @@ ExternConverter *ExternConverter::get(cstring type) {
 
 ///////////////////////////////////////////////////////////////
 
+std::map<cstring, std::vector<PrimitiveConverter *>> *PrimitiveConverter::all_converters;
+
+PrimitiveConverter::PrimitiveConverter(cstring name, int prio) : prim_name(name), priority(prio) {
+    static std::map<cstring, std::vector<PrimitiveConverter *>> converters;
+    all_converters = &converters;
+    auto &vec = converters[name];
+    auto it = vec.begin();
+    while (it != vec.end() && (*it)->priority > prio) ++it;
+    if (it != vec.end() && (*it)->priority == prio)
+        BUG("duplicate primitive converter for %s at priority %d", name, prio);
+    vec.insert(it, this);
+}
+
+PrimitiveConverter::~PrimitiveConverter() {
+    auto &vec = all_converters->at(prim_name);
+    vec.erase(std::find(vec.begin(), vec.end(), this));
+}
+
+const IR::Statement *PrimitiveConverter::cvtPrimitive(ProgramStructure *structure,
+                                                      const IR::Primitive *primitive) {
+    if (all_converters->count(primitive->name))
+        for (auto cvt : all_converters->at(primitive->name))
+            if (auto *rv = cvt->convert(structure, primitive))
+                return rv;
+    return nullptr;
+}
+
+vector<const IR::Expression *>
+PrimitiveConverter::convertArgs(ProgramStructure *structure, const IR::Primitive *prim) {
+    ExpressionConverter conv(structure);
+    vector<const IR::Expression *> rv;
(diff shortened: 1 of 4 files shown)


frontends/p4/fromv1.0/converters.cpp
```diff
@@ -416,6 +416,35 @@ ExternConverter *ExternConverter::get(cstring type) {
 
 ///////////////////////////////////////////////////////////////
 
+std::map<cstring, std::vector<PrimitiveConverter *>> *PrimitiveConverter::all;
+
+PrimitiveConverter::PrimitiveConverter(cstring name, int prio) : prim_name(name), priority(prio) {
+    static std::map<cstring, std::vector<PrimitiveConverter *>> converters;
```
cc10512: Why do you need another local static variable to initialize 'all'? You could initialize all directly as a static data member to an empty map. And BTW, 'all' is not the most fortunate choice for this name.

frontends/p4/fromv1.0/programStructure.cpp
```diff
@@ -1326,11 +922,683 @@ const IR::Statement* ProgramStructure::convertPrimitive(const IR::Primitive* pri
...
     }
-
     BUG("Unhandled primitive %1%", primitive);
     return nullptr;
 }
 
+CONVERT_PRIMITIVE(modify_field) {
+    ExpressionConverter conv(structure);
+    if (primitive->operands.size() == 2) {
+        auto left = conv.convert(primitive->operands.at(0));
```
mbudiu-vmw: How about a convert method that converts all args and returns an array?

frontends/p4/fromv1.0/programStructure.cpp
```diff
@@ -1326,11 +922,683 @@ const IR::Statement* ProgramStructure::convertPrimitive(const IR::Primitive* pri
...
+    return nullptr;
+}
+
+CONVERT_PRIMITIVE(bit_xor) {
+    ExpressionConverter conv(structure);
+    OPS_CK(primitive, 3);
+    auto dest = conv.convert(primitive->operands.at(0));
+    auto left = conv.convert(primitive->operands.at(1));
+    auto right = conv.convert(primitive->operands.at(2));
+    if (!sameBitsType(left->type, right->type)) {
```
mbudiu-vmw: A utility method for inserting casts could be useful in the base class.

frontends/p4/fromv1.0/converters.cpp
ChrisDodd: Because of the !@$&! undefined order of global static initialization in C++.  If the map was in one compilation unit and an instance static was in another, it *might* initialize the instance before the map, leading to undefined behavior.  By making the map itself a static local in the constructor, it is guarenteed to be initialized the first time the constructor is run.  I'll rename the var to be more descriptive.

frontends/p4/fromv1.0/programStructure.cpp
```diff
@@ -1385,7 +1653,7 @@ ProgramStructure::convertAction(const IR::ActionFunction* action, cstring newNam
 
     for (auto p : action->action) {
         auto stat = convertPrimitive(p);
-        if (stat != nullptr)
+        if (stat != nullptr /* && !stat->is<IR::EmptyStatement>() */)
```
hanw: Do you still need this commented code?

ChrisDodd: This was an optimization I was considering, but its really independent of the change -- if the converter returns a EmptyStatement, its not really necessary to put it in the action, and a later call to SimplifyControlFlow will remove it.  But if it is meaningless, perhaps its better to never put it in in the first place.

PR #766, opened 2017-07-08
Strip out p4_14 extern support code in ResolveReferences

ChrisDodd:
- this code was to support P4_14-style externs with attributes and is no
  longer needed now that we have ExternConverters to convert those P4_14
  extern_types to P4_16 properly.


ChrisDodd:
The stripped out code was for resolving properties of extern instances, matching them up to the corresponding attributes of the extern type, and resolving references to the expression_local_variables of those attributes.  These are all features of P4_14 that are not present in P4_16.

The warning message could probably be better worded; I'll think about that.

PR #770, opened 2017-07-12
convert path passed by -I to absolute path

hanw:
This is required if we want to match on srcInfo to figure out whether a declaration comes from architecture or user program.

### DIFF
--- a/frontends/common/options.cpp
+++ b/frontends/common/options.cpp
@@ -32,6 +32,21 @@ const char* p4_14includePath = CONFIG_PKGDATADIR "/p4_14include";
 
 const char* CompilerOptions::defaultMessage = "Compile a P4 program";
 
+template <size_t N>
+static void convertToAbsPath(const char* const relPath, char (&output)[N]) {
+    output[0] = '\0';  // Default to the empty string, indicating failure.
+
+    char cwd[PATH_MAX];
+    getcwd(cwd, sizeof(cwd));
+    const size_t cwdLen = strlen(cwd);
+    if (cwdLen == 0) return;
+    const char* separator = cwd[cwdLen - 1] == '/' ? "" : "/";
+
+    // Construct an absolute path. We're assuming that @relPath is relative to
+    // the current working directory.
+    snprintf(output, N, "%s%s%s", cwd, separator, relPath);
+}
+
 CompilerOptions::CompilerOptions() : Util::Options(defaultMessage) {
     registerOption("--help", nullptr,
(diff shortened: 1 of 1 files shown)


mbudiu-vmw:
Using the path sounds very brittle and will break if people move files aroud. You should rely on type names.

hanw:
@mbudiu-vmw Declaration `srcInfo`s currently have a mix of the relative and absolute path to the source file, this change helps to make it consistent. I will look into a more appropriate way to distinguish an architecture declaration from user program declaration, but this change is useful nonetheless.

frontends/common/options.cpp
```diff
@@ -32,6 +32,21 @@ const char* p4_14includePath = CONFIG_PKGDATADIR "/p4_14include";
...
+    getcwd(cwd, sizeof(cwd));
+    const size_t cwdLen = strlen(cwd);
+    if (cwdLen == 0) return;
+    const char* separator = cwd[cwdLen - 1] == '/' ? "" : "/";
+
+    // Construct an absolute path. We're assuming that @relPath is relative to
+    // the current working directory.
+    snprintf(output, N, "%s%s%s", cwd, separator, relPath);
+}
+
```
cc10512: why not using realpath? https://linux.die.net/man/3/realpath

frontends/common/options.cpp
```diff
@@ -43,7 +58,9 @@ CompilerOptions::CompilerOptions() : Util::Options(defaultMessage) {
                        exit(0); return false; }, "Print compiler version");
     registerOption("-I", "path",
                    [this](const char* arg) {
-                       preprocessor_options += std::string(" -I") + arg; return true; },
+                       char buffer[PATH_MAX];
+                       convertToAbsPath(arg, buffer);
```
cc10512: this does not seem to change the files, but rather the include paths. I'm not sure this solves the issues you are seeing. If you are targeting the p4 14 to 16 converter (https://github.com/p4lang/p4c/blob/master/frontends/p4/fromv1.0/programStructure.cpp#L469) hardcodes the path already to a compile-time defined path.

frontends/common/options.cpp
hanw: because realpath will follow the symbolic name of p4c-bm2-ss in build/p4c to `build/backends/bmv2/p4c-bm2-ss`, and `p4includePath` will be generated as `build/backends/bmv2/p4c-bm2-ss/p4include`, instead of `/build/p4c/p4include`. I used p4includePath as prefix to match on the path to system p4 files, which wouldn't work.

PR #782, opened 2017-07-20
mangles the names used by the converter to avoid clashes with

### LINKED ISSUE #780: Bogus "duplicates declaration" error
Compiling the following P4-14 program,
```
header_type headers {
  fields {
    counterrevolution : 32;
    cartographers : 8;
    wadded : 8;
    terns : 32;
    miscellaneous : 8;
  }
}
header headers heartlands;
parser start {
  return ingress;
}
action add_heartlands() {
  add_header(heartlands);
}
control ingress { }
```
produces a mysterious error:
```
% p4test --p4-14 ~/Downloads/envelop.p4 
envelop.p4(1): error: struct headers: Duplicates declaration header headers
header_type headers {
^
```
If I change `headers` to `foo`, then it converts just fine:
```
#include <core.p4>
#include <v1model.p4>

header foo {
    bit<32> counterrevolution;
    bit<8>  cartographers;
    bit<8>  wadded;
    bit<32> terns;
    bit<8>  miscellaneous;
}

struct metadata {
}

struct headers {
    @name("heartlands") 
    foo heartlands;
}

    @name(".start") state start {
        transition accept;
    }
}

control ingress(inout headers hdr, inout metadata meta, inout standard_metadata_t standard_metadata) {
    apply {
    }
}

control egress(inout headers hdr, inout metadata meta, inout standard_metadata_t standard_metadata) {
    apply {
    }
}

control DeparserImpl(packet_out packet, in headers hdr) {
    apply {
    }
}

control verifyChecksum(in headers hdr, inout metadata meta) {
    apply {
    }
}

control computeChecksum(inout headers hdr, inout metadata meta) {
    apply {
    }
}

V1Switch(ParserImpl(), verifyChecksum(), ingress(), egress(), computeChecksum(), DeparserImpl()) main;
```
Hence, it seems that `p4c` is introducing names that may or may not be unique, and/or not mangling the names in the source program to ensure there are no clashes.

cc10512:
user defined structures.

Adds a test case covering the allegedly legal combinations.
standard_metadata_t and main are not replaceable.

Fixes #780.


jnfoster:
ResolveReferences provides a function to generate names that are unique for a given namespace. Why aren't we using that instead of any fixed strategy?

-N

mbudiu-vmw:
ResolveReferences can be used only after the program has been constructed, but in this case identifiers must be generated to translate the program to P4-16. But the ProgramStructure class, used for the conversion from P4-14 to P4-16, has a makeUniqueName method. 

I think that one solution would be to populate the list of existing identifiers with all the identifiers from v1model.p4; then the existing code which renames P4-14 identifiers would automatically assign new names.

cc10512:
The compiler seems to rely on the actual name of ingress/egress/standard_metadata for some of the "optimizations". basic_routing.p4 fails in bmv2 because the rewrite_mac table in egress is eliminated. I have to track this down, and I may as well implement a list of reserved names for v1model. As long as we give reasonable messages, it is likely to be an acceptable (and cheap) solution.

jnfoster:
In addition to `v1model.p4`, we also need to worry about names in `core.p4`...

cc10512:
I re-implemented this patch to include all the names in core.p4 and v1model.p4 as reserved symbols. Now the compiler prints an error as follows:
```
./p4test --pp ./tmpnaPBOz/issue780.p4 --dump ./tmpnaPBOz --top4 FrontEndLast,FrontEndDump,MidEndLast --testJson --p4-14 ../p4c/testdata/p4_14_errors/issue780.p4
Error compiling
error: 'egress' is reserved in P4_14 as a 'control' and can not be used as struct
error: 'egress' is reserved in P4_14 as a 'control' and can not be used as struct
error: 2 errors encountered, aborting compilation
```

I tried to rename these in different ways using the ProgramStructure::NamedObjectInfo, or simply changing the name in convertType (although, the later was not a good solution in any case, as it needed to be applied to multiple declarations). However, there is not enough information stored in this interface to be able to disambiguate different uses of the same name in different contexts. So I chickened out and decided to print an error very early in the DiscoverStructure process.

@mbudiu-vmw if you have other suggestions, please let me know. I left around a couple of instances where I added code to do the translation. I can remove those before the final merge.

mbudiu-vmw:
I was hoping that something simpler would work (but maybe it doesn't?).
The solution I was envisioning is just adding to structure.allNames the list of reserved words. (Perhaps even including the parameter names used in v1model.) Then the automatic renaming of objects with duplicate names should rename all these objects using `@name` annotations.

cc10512:
I tried that (that's why I have the addReservedWords method), but it didn't work.

mbudiu-vmw:
How does that fail?

cc10512:
One example: when the p4 program defines an `egress` control, `ProgramStructure::createControls()` is looking to convert `egress_0`, which does not exist. 
There were others, but I don't recall. @mbudiu-vmw, for your pleasure, I pushed a branch that has just that change here: https://github.com/cc10512/p4c/tree/cc/issue780-mb.

mbudiu-vmw:
I can see how that fails; sometimes "egress" has to be renamed, but sometimes it does not, depending on where it is used. Let me try your patch to see how it works on more test cases.

cc10512:
Right, allNames is not really a symbol table. That would be the right solution: build a true symbol table, so that names are associated with a kind (type is overloaded). Then we can add `egress` as header, control, table name, etc.

mbudiu-vmw:
We actually have a lot of symbol tables; there is one for each kind of symbol; these are all of the NamedObjectInfo fields. But if your solution works, it's good enough.
I have to also think about the parameter names; using underscores is not a complete solution, since these names are not reserved. So maybe something has to be done about them too.

mbudiu-vmw:
It looks like your patch will still reject correct P4-14 programs:
`error: 'main' is reserved in P4_14 as a 'package' and can not be used as struct`

Do you mind if I spend some time trying to modify your code?

cc10512:
@mbudiu-vmw please do modify it so that we reject the minimum number of programs.

PR #784, opened 2017-07-20
Fixed a small bug in the bitvec >>= function

evanches:
Essentially this was failing on the border of bits_per_unit.  The size should always be at least 1 in the calculation.


evanches:
If the shift was a multiple of bits_per_unit, this shift failed as well.

evanches:
I've written a very simple gtest that describes what the problem was when I found the bug.  Future fixes to bitvec should now also have a corresponding gtest.

cc10512:
@evanches please squash your commits when you merge.

Issue #785, opened 2017-07-21
p4test generates invalid "source_fragment" value in JSON file


brmarci:
I tried to generate JSON output using p4test from various P4 sources, but there are some `"source_fragment"` fields that are set to `"""` resulting in an invalid JSON file.

When running
`./p4test -I p4include/ ../testdata/p4_16_samples/vss-example.p4 --p4v 16 --toJSON temp.json`
the result file contains the following (from line 7067 to 7080):
```
                {
                  "Node_ID" : 90149,
                  "Node_Type" : "StringLiteral",
                  "type" : {
                    "Node_ID" : 3906
                  },
                  "value" : "NoAction",
                  "Source_Info" : {
                    "filename" : "/home/brmarci/work/p4/p4c/build/p4include/core.p4",
                    "line" : 67,
                    "column" : 15,
                    "source_fragment" : """
                  }
                }
```

Also when generating JSON from a P4_14 source file, there are similar problems referencing p4include/v1model.p4.

cc10512:
Fixed by #786

PR #786, opened 2017-07-21
Fix for issue #785; readme formatting

mbudiu-vmw:


Makefile.am
```diff
@@ -1,302 +0,0 @@
-# Copyright 2013-present Barefoot Networks, Inc.
```
cc10512: I don't understand why do you need to remove Makefile.am again. Have you rebased?

mbudiu-vmw: I have, this is built with cmake.
I don't know why the files are still there.

mbudiu-vmw: I guess maybe not this particular branch...

cc10512: git likely does not remove your local files once you've update the branch. You can run `git clean -dxf` after you rebased your master.

Removing them from .gitignore is a good catch!

ir/node.cpp
```diff
@@ -102,6 +102,17 @@ Util::JsonObject* IR::Node::sourceInfoJsonObj() const {
     return json;
 }
 
+static cstring quote(cstring s) {
```
cc10512: shouldn't this be a method of cstring?

mbudiu-vmw: This quoting is really specific to JSON.
For C you would need more cases.
So I decided to make it local for now.

mbudiu-vmw: I could move this method to cstring and call it jsonQuote.

cc10512: would be nice, because otherwise, nobody is going to find it and it will be bound to repeat. A comment on it would also help :)

PR #789, opened 2017-07-25
Check if expression uses a IR::Member object

ksubramanianBarefoot:
Check if an expression uses a `IR::Member` object. `exprUses` can be now used to check if an expression uses a packet header or metadata field.  

Edit: `make check` passes on my local machine, any reason why Travis would fail? The repo is synced with the latest commit.

### DIFF
--- a/midend/expr_uses.h
+++ b/midend/expr_uses.h
@@ -27,6 +27,9 @@ class exprUses : public Inspector {
     bool preorder(const IR::Path *p) override {
         if (p->name == look_for) result = true;
         return !result; }
+    bool preorder(const IR::Member *m) override {
+        if (m->toString() == look_for) result = true;
+        return !result; }
     bool preorder(const IR::Primitive *p) override {
         if (p->name == look_for) result = true;
         return !result; }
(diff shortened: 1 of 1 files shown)


midend/expr_uses.h
```diff
@@ -27,6 +27,9 @@ class exprUses : public Inspector {
     bool preorder(const IR::Path *p) override {
         if (p->name == look_for) result = true;
         return !result; }
+    bool preorder(const IR::Member *m) override {
+        if (m->toString() == look_for) result = true;
```
mbudiu-vmw: Do not use expression->toString() in the compiler; it is only meant for displaying information to the user. It is lossy.

What is the purpose of this commit? Who uses this method and why?

ksubramanianBarefoot:
Removed toString() usage and checking usage recursively. 
I am using this feature for different analyses/optimizations which deal with header and metadata fields, and I need `exprUses` to detect usage of these fields.

ChrisDodd:
Sorry for the delay; I was on vacation.  Mihai, I assume this can be approved now?

mbudiu-vmw:
It looks to me that this code is superseded by code in #904

ChrisDodd:
Yes, this is now obsolete, so I'll close it.

Issue #763, opened 2017-07-05
P4_14 to P4_16 conversion duplicate declarations


jnfoster:
The `p4c` compiler accepts this (P4_14) program, even though it has a duplicate top-level declaration of `perry`. It seems like this program should be rejected, or the compiler should at least issue a warning.

```
header_type morris {
  fields {
    nichols : 112;
  }
}

header_type irrelevancy {
  fields {
    tweedledee : 32;
    rumsfeld : 32;
    palavers : 8;
  }
}

header_type stammers {
  fields {
    childproofing : 32;
    unmitigated : 16;
    rathers : 32;
    souphanouvongs : 8;
  }
}

header morris egalitarians;

header irrelevancy perry;

header stammers perry;

parser start {
  return parse_egalitarians;
}

parser parse_egalitarians {
  extract(egalitarians);
  return parse_perry;
}

parser parse_perry {
  extract(perry);
  return select (latest.palavers) {
    0 : ingress;
    default : ingress;
  }
}

control ingress {
}
```

mbudiu-vmw:
I am assuming we can close this issue.

PR #798, opened 2017-07-27
Issue780

mbudiu-vmw:
This does not handle some cases which should be handled by #763.
This should extend the functionality of PR #782.
Most of the symbols that occur in v1model and core.p4 are just renamed if they occur in the user program.
There are three symbols that are reserved and should not be used in P4-14 programs for other purposes: egress, standard_metadata and standard_metadata_t.


cc10512:
@mbudiu-vmw, if I read this correctly, it is a full replacement of #782, not an extension. I shall close that PR.

frontends/p4/fromv1.0/programStructure.cpp
```diff
@@ -1902,7 +1912,7 @@ void ProgramStructure::createControls() {
     std::vector<cstring> knownControls;
 
     for (auto it : controls)
-        knownControls.push_back(it.second);
+        knownControls.push_back(it.first->name);
```
cc10512: `ct` in line 1924 can also end up null and crash the program. Can you please add a null check there as well?

PR #806, opened 2017-07-29
add policies to skip certain checks in frontend

hanw:
It is useful to be able to skip certain passes in the frontend pass manager, in the case that such checks are already performed in some pre-frontend passes. I ran into a case this is required, adding policies to the frontend passes seems to be the right approach.


hanw:
Closed as no longer needed.

PR #811, opened 2017-08-01
support make uninstall in cmake

hanw:


hanw:
Took the snippet from cmake wiki with minor changes. https://cmake.org/Wiki/CMake_FAQ#Can_I_do_.22make_uninstall.22_with_CMake.3F

PR #812, opened 2017-08-01
Revert "convert path passed by -I to absolute path internally (#770)"

hanw:
This reverts commit b4db551d5cb1106daf085baf547720f60905f7a4.


hanw:
The commit was introduced to allow me to use the file path to identify the system include file. It is no longer required, and it also appears to cause regression. So I am backing it out.

rhalstea:
@hanw 
thanks for fixing this.

Issue #781, opened 2017-07-19
p4c-bm2-ss generates invalid extract_VL code for variable-length fields


antoninbas:
This P4_14 header declaration:
```
header_type ipv4_t {
    fields {
        version     : 4;
        ihl         : 4;
        diffserv    : 8;
        totalLen    : 16;
        id          : 16;
        flags       : 3;
        fragOffset  : 13;
        ttl         : 8;
        protocol    : 8;
        hdrChecksum : 16;
        srcAddr     : 32;
        dstAddr     : 32;
        options_ipv4: *;
    }
    length          : (ihl << 2);
    max_length      : 60;
}
```
generates invalid JSON code when the corresponding header instance is extracted in a parser state.
According to the bmv2 JSON documentation:
```
extract_VL: introduced for P4_16, where the expression to dynamically compute the length of a variable-length field is an argument to the extract built-in rather than a property of the header. For this operation, we require 2 parameters, the first one of type regular and the second one of type expression (to compute the length in bits of the variable-length field in the header).
```
However, the bmv2 backend simply outputs `ihl << 2` as the second argument to `extract_VL`, instead of the correct value, which should be something like: `((ihl << 2) - 20) * 8`. Notice the discrepancy between the P4_14 spec definition of `length` (length of the entire header in bytes) and the expected value in the bmv2 JSON (length of the VL field in bits). This discrepancy is probably what created this issue.
I wonder if this bug affects P4_16 varbit extraction as well?

I am attaching a complete P4_14 program which exhibits this bug.
[simple_router_ipv4_VL.p4.txt](https://github.com/p4lang/p4c/files/1160699/simple_router_ipv4_VL.p4.txt)

mbudiu-vmw:
Yes, the p4-14 to p4-16 generates a @length annotation for the field, which is currently incorrect.
This seems easy to fix, but there is one little problem: ihl is on 4 bits, so just generating an annotation `@length(ihl << 2) * 8 - 160` will generate a computation on 4 bits (the width of ihl). The shift of ihl by 2 will overflow; so the compiler should really cast ihl to a wider value before making the shift. It is not obvious what this larger value should be. I guess we can use 32-bits... @ChrisDodd : do you have any suggestion?

mbudiu-vmw:
BTW: I have already implemented the fix for the expression, but I haven't done the casting.

mbudiu-vmw:
@antoninbas : I will close this, please re-open if you think it is not solved.

PR #813, opened 2017-08-01
Fix for issue #781

mbudiu-vmw:
This was quite difficult to fix, since the fixes uncovered a couple of other bugs in other parts of the compiler. I have filed some test cases for these as well.

### DIFF
--- a/frontends/p4/typeChecking/typeChecker.cpp
+++ b/frontends/p4/typeChecking/typeChecker.cpp
@@ -19,6 +19,7 @@ limitations under the License.
 #include "frontends/p4/substitution.h"
 #include "typeConstraints.h"
 #include "frontends/p4/coreLibrary.h"
+#include "frontends/p4/toP4/toP4.h"
 #include "syntacticEquivalence.h"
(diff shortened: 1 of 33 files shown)


frontends/p4/typeChecking/typeChecker.cpp
```diff
@@ -117,8 +118,15 @@ Visitor::profile_t TypeInference::init_apply(const IR::Node* node) {
 }
 
 void TypeInference::end_apply(const IR::Node* node) {
-    if (readOnly && !(*node == *initialNode))
-        BUG("%1%: typechecker mutated node %2%", dbp(node), dbp(initialNode));
+    if (readOnly && !(*node == *initialNode)) {
+        ToP4 top4(&std::cout, true, nullptr);
+        std::cout << "Initial program" << std::endl;
```
jnfoster: Seems like some debugging statements are left here. Intentional?

mbudiu-vmw: Yes, this block ends with a BUG call, so all these statements just print more information to help analyze the bug.

Issue #807, opened 2017-07-30
Compiler bug: "expected an instance declaration"


jnfoster:
While trying to write a devious cyclic example for @jafingerhut, I discovered a new (to me) compiler bug in the inlining pass of the midend:
```
#include <core.p4>

// Architecture    
control C1();
control C2();
control C3();
extern E {
  E();
  void set1(C1 c1);
  void set2(C2 c2);
  C1 get1();
  C2 get2();
}

package S(C3 c3);

// User Program            
control MyC1()(E e) {
  apply {
    C2 c2 = e.get2();
    c2.apply();
  }
}        
control MyC2()(E e) {
  apply {
    C1 c1 = e.get1();
    c1.apply();
  }
}        
control MyC3()(E e, C1 c1, C2 c2) {
  apply {
    e.set1(c1);
    e.set2(c2);
    c1.apply();
  }
}        

E() e;
    
S(MyC3(e,MyC1(e),MyC2(e))) main;
```
Error:
```
In file: p4c/midend/inlining.cpp:404
Compiler Bug: cycle.p4(18): C2 c2_0/c2 expected an instance declaration
    C2 c2 = e.get2();
    ^^^^^^^^^^^^^^^^^
```

mbudiu-vmw:
I think that we'll fix this one by forbidding methods to return controls (parsers, packages, or other externs).

jnfoster:
The following program produces the same error message:
```
#include <core.p4>

// Architecture    
control C();
package S(C c);
extern BoolRegister {
  BoolRegister();
  bool get();
  void flip();
}        

// User Program
control MyC1() {
  apply {
  }
}
control MyC2() {
  apply {
  }
}        
control MyC3()(BoolRegister r, C c1, C c2) {
  apply {
    r.flip();
    if(r.get()) c1.apply();
    else c2.apply();
  }
}
control MyC4() {
  BoolRegister() r;
  MyC1() c1;
  MyC2() c2;
  MyC3(r,c1,c2) c3;
  apply {
    c3.apply();
  }  
}        
S(MyC4()) main;
```

```
In file: p4c/midend/inlining.cpp:404
Compiler Bug: cyclic.p4(21):  C c1 expected an instance declaration
control MyC3()(BoolRegister r, C c1, C c2) {
                                 ^^
```

It only uses controls as constructor parameters, which is definitely legal. It also seems unlikely this is fixed by 7317e5fc5bb4d56761287cbc86be092aafadf322?

PR #815, opened 2017-08-01
Fix for issue #807

mbudiu-vmw:


cc10512:
Are we also updating the spec to reflect this change as discussed in https://github.com/p4lang/p4-spec/issues/361?

jafingerhut:
@cc10512 I have two issues open for p4lang/p4-spec that you might be thinking about.  Let me know if there are any other possible P4_16 spec changes you think ought to have issues created for them:

Propose documenting that methods cannot return references to stateful objects: https://github.com/p4lang/p4-spec/issues/367

Document clearly that controls and parsers cannot be passed as run-time parameters to controls/parameters: https://github.com/p4lang/p4-spec/issues/364

jnfoster:
The bug report was for the following error message:
```
In file: p4c/midend/inlining.cpp:404
Compiler Bug: cycle.p4(18): C2 c2_0/c2 expected an instance declaration
    C2 c2 = e.get2();
```
This pull request does not modify `inlining.cpp`. That's fine, of course. But for the purposes of deepening our collective understanding of the compiler, it would be great to add a brief note describing what invariant `inlining.cpp` was incorrectly assuming and why it is now true.

jnfoster:
It appears this fix addresses the symptom but not the root cause. See  https://github.com/p4lang/p4c/issues/807#issuecomment-319593487 for an example program that triggers the same bug but ever return a control...

mbudiu-vmw:
The root cause is that controls were not designed to be passed around as first-class values.
We just have to tighten the type checker to prevent that kind of usage.
This fixes this reported issue, there will be other fixes for other similar issues.

PR #825, opened 2017-08-04
Fix typo in P4Runtime

antoninbas:
This fixes commit 9b4e0db15fe9578be5f74052b993c010cbe8f454. For table
properties, the name annotation is on the property name, not the value.


cc10512:
@antoninbas On second thoughts I believe the intent was to name the value as well. IIRC, in the case where you have a key defined as an expression, you need a name for the expression.

antoninbas:
this method is not used for key & actions; it is only called for implementation, counters and meters AFAIK. I actually tried annotating the value in these cases and it is a syntax error (i.e. the parser complains).

PR #828, opened 2017-08-07
Mac OSX does not have rt (and ir-generator) doesn't need it anyway)

cc10512:


rhalstea:
When I take out this flag I get the following error message

```
[  9%] Linking CXX executable irgenerator
../../lib/libp4ctoolkit.a(log.cpp.o): In function `Log::Detail::operator<<(std::ostream&, Log::Detail::OutputLogPrefix const&)':
log.cpp:(.text+0x3b): undefined reference to `clock_gettime'
../../lib/libp4ctoolkit.a(log.cpp.o): In function `Log::addDebugSpec(char const*)':
log.cpp:(.text+0x5b7): undefined reference to `clock_gettime'
collect2: error: ld returned 1 exit status
make[2]: *** [tools/ir-generator/irgenerator] Error 1
make[1]: *** [tools/ir-generator/CMakeFiles/irgenerator.dir/all] Error 2
make: *** [all] Error 2
```

Can we insert the flag if https://github.com/p4lang/p4c/blob/master/CMakeLists.txt#L53 passes?

cc10512:
What configuration are you building? I'm curious because Travis does not show that error.
And yes, I will add the flag, just want to make sure I can see this too.

akeep:
I noticed there was a check for the library ``rt`` and the symbol``clock_gettime`` in the top-level ``CMakeLists.txt`` file.  Is it possible that we could include ``${P4C_LIB_DEPS}`` in place of ``rt`` in the ``target_link_libraries`` of the ``tools/ir-generator/CMakeLists.txt``?

I've not yet tested this out on Linux, but it does fix the problem on the Mac.  That said, I'm not sure why the ``ir-generator`` is using ``clock_gettime``.

akeep:
Oh, sorry, I guess it is really ``log.cpp`` that is using ``clock_gettime``.

rhalstea:
The base OS is CentOS 6.5, but a lot of the tools packages I link against are installed in a development folder (e.g. GCC, GMP, GC, Protobuf, etc.). As far as I can see there is nothing special about rt it's just not linked by default, but I'm not sure if that's a CentOS thing or not.

cc10512:
Thanks @akeep. Pushed.

ChrisDodd:
ir-generator uses log.cpp for optional debug logging, for debugging the
generator (by default it is not used, but can be enabled with command-line
flags)

The logging code optionally includes timestamps in the log messages for
debugging timing related issues (identifying slow parts of code).

The log timestamping code uses clock_gettime if it is available on your
system (defined in the system header files), but not otherwise.

So the fact that you're getting link time errors (not compile errors)
indicate that your system header files claim it is available, but your
system libs don't actually support it. Which seems like a system config
problem.

On Mon, Aug 7, 2017 at 9:50 AM, Robert J. Halstead <notifications@github.com

rhalstea:
@cc10512 thanks. That change works in my environment.

rhalstea:
@ChrisDodd CentOS 6.5 has a glibc version 2.12 which according to the man page for clock_gettime (http://man7.org/linux/man-pages/man2/clock_gettime.2.html) still needs -lrt since it's below version 2.17

I'm out of my depth here, so it could be a system config issue. Would ```CLOCK_MONOTONIC``` be set if gcc doesn't get the -lrt flag?

ChrisDodd:
It seems to me like `-lrt` should always be the default (along with `-lc`) on systems where it is required.

PR #832, opened 2017-08-08
add getApplyParameters() method to IApply interface

hanw:
The IR has `getConstructorParameters` and `getTypeParameters`, but not `getApplyParameters` which makes the IR API asymmetrical.

One thing I am not sure about in this commit is the `table.apply()` does not allow apply parameter in P4-16. Hence, `getApplyParameters()` on table will return a `nullptr`. We do the same for the `getP4Type()` function in `type.def`.

### DIFF
--- a/ir/ir.def
+++ b/ir/ir.def
@@ -96,6 +96,7 @@ class P4Parser : Type_Declaration, ISimpleNamespace, IApply, IContainer {
             return decl;
         return states.getDeclaration(name); }
     Type_Method getApplyMethodType() const override { return type->getApplyMethodType(); }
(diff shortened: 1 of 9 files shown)


ir/ir.def
```diff
@@ -289,6 +291,7 @@ class P4Table : Declaration, IAnnotated, IApply {
 
     Annotations getAnnotations() const override { return annotations; }
     Type_Method getApplyMethodType() const override;
+    ParameterList getApplyParameters() const override { return nullptr; }
```
mbudiu-vmw: I think this should be an empty parameter list.
You don't want to check for nullptr everywhere.
Moreover, this will make it easier to add parameters in the future if we decide to do that.

PR #833, opened 2017-08-08
Add graphs backend

antoninbas:
It can generate a graph representation of top-level controls as dot
files.


antoninbas:
Please look at the README for the graphs backend for information. I added a link to an example (switch.p4 egress).
There is no rush to merge this, but it is IMO convenient to have a graph representation for large P4 programs. There is still a lot of room for improvement (some ideas are listed in the README) and possibly some bugs, but I wanted to check if people thought this was a good start.

backends/graphs/CMakeLists.txt
```diff
@@ -0,0 +1,40 @@
...
+add_executable(p4c-graphs ${GRAPHS_SRCS} ${EXTENSION_P4_14_CONV_SOURCES})
+target_link_libraries (p4c-graphs ${P4C_LIBRARIES} ${P4C_LIB_DEPS})
+add_dependencies(p4c-graphs genIR)
+
+install (TARGETS p4c-graphs
+  RUNTIME DESTINATION ${P4C_RUNTIME_OUTPUT_DIRECTORY})
+
+# hack to get around the fact that the test scripts expect the backend
+# binary to be in the top level directory. This should go away when we
+# remove automake and fix the scripts.
```
mbudiu-vmw: We have removed automake already.
Is this comment still applicable?

backends/graphs/README.md
```diff
@@ -0,0 +1,34 @@
...
+```
+mkdir out
+p4c-graphs <prog.p4> --graphs-dir out
+cd out
+dot <name>.dot -Tpng > <name>.png
+```
+
+## Improvement ideas
+
+  - use different shapes for vertices of different types (e.g. rectangles for
```
mbudiu-vmw: I would highly recommend using rectangles instead of ellipses, the use the space much better. The output graphs will be much more compact.

backends/graphs/controls.cpp
```diff
@@ -0,0 +1,304 @@
...
+    enum class Branch { TRUE, FALSE };
+    explicit EdgeIf(Branch branch) : branch(branch) { }
+    std::string label() const override {
+        switch (branch) {
+          case Branch::TRUE:
+              return "TRUE";
+          case Branch::FALSE:
+              return "FALSE";
+        }
+        assert(0 && "unreachable");
```
mbudiu-vmw: We use BUG("unreachable") for this.

backends/graphs/controls.cpp
```diff
@@ -0,0 +1,304 @@
...
+#include "lib/path.h"
+
+#include "controls.h"
+
+namespace graphs {
+
+class EdgeUnconditional : public EdgeTypeIface {
+ public:
+    EdgeUnconditional() = default;
+    std::string label() const override { return ""; };
```
mbudiu-vmw: Not really important, but we have a cstring class which may be suitable.

backends/graphs/controls.h
```diff
@@ -0,0 +1,119 @@
...
+
+    // merge misc control statements (action calls, extern method calls,
+    // assignments) into a single vertex to reduce graph complexity
+    boost::optional<vertex_t> merge_other_statements_into_vertex();
+
+    vertex_t add_vertex(const std::string &name);
+    vertex_t add_vertex(const cstring &name);
+
+    bool preorder(const IR::PackageBlock *block) override;
+
```
mbudiu-vmw: I think that all these empty lines are unnecessary.

backends/graphs/controls.h
```diff
@@ -0,0 +1,119 @@
...
+    void writeGraphToFile(const Graph &g, const cstring &name);
+
+ private:
+    P4::ReferenceMap *refMap; P4::TypeMap *typeMap;
+    const cstring graphsDir;
+    Graph *g{nullptr};
+    vertex_t root{};
+    vertex_t exit_v{};
+    // boost::optional<vertex_t> local_root{};
```
mbudiu-vmw: Do you need these?

backends/graphs/graphs.cpp
```diff
@@ -0,0 +1,128 @@
...
+        evaluator,
+        new VisitFunctor([this, evaluator]() { toplevel = evaluator->getToplevelBlock(); }),
+    });
+}
+
+class Options : public CompilerOptions {
+ public:
+    cstring graphsDir{"."};
+    Options() {
+        registerOption("--graphs-dir", "dir",
```
mbudiu-vmw: You could also just reuse the existing -o option and treat it as a folder name, or as a filename prefix.
But there is no simple way to change its documentatition...

backends/graphs/graphs.cpp
```diff
@@ -0,0 +1,128 @@
...
+    if (::errorCount() > 0)
+        return 1;
+
+#ifdef HAVE_LIBBOOST_GRAPH
+    LOG1("Generating graphs under " << options.graphsDir);
+    LOG1("Generating control graphs");
+    graphs::ControlGraphs cgen(&midEnd.refMap, &midEnd.typeMap, options.graphsDir);
+    top->getMain()->apply(cgen);
+#else
+    ::error("This backend requires the boost graph headers, which could not be found");
```
mbudiu-vmw: This looks like a runtime error, but the problem happened when building the back-end.
Either the error message should be clearer, or the build should fail.

backends/graphs/controls.h
sethfowler: Not to be contrary, but my preference is for you to keep the empty lines. They greatly help readability in my opinion.

sethfowler:
Re: the number of dependencies, I think we should just accept that we depend on all of boost.

CMakeLists.txt
```diff
@@ -37,7 +37,14 @@ find_package (PythonInterp REQUIRED)
 find_package (FLEX REQUIRED)
 find_package (BISON REQUIRED)
 find_package (Protobuf 3.0.0 REQUIRED)
+# The boost graph headers are optional and only required by the graphs backend
```
sethfowler: It's not really worth making it optional, IMO, but since you've already done the work it's up to you.

backends/graphs/README.md
```diff
@@ -0,0 +1,34 @@
...
+
+In addition to other p4c dependencies, this backend requires the Boost graph
+headers. On a Debian system, they can be installed with `sudo apt-get install
+libboost-graph-dev`.
+
+## Usage
+
+```
+mkdir out
+p4c-graphs <prog.p4> --graphs-dir out
```
sethfowler: If you're only generating one file, why not just specify a file instead of a directory?

backends/graphs/README.md
sethfowler: Agreed. I think the rectangles are more "readable", as well.

backends/graphs/README.md
```diff
@@ -0,0 +1,34 @@
...
+  - change representation of controls: instead of adding a vertex labelled with
+    the control name when the control is invoked, group the control contents
+    into a "subgraph"
+
+
+## Example
+
+You can look at the control graph generated for the egress control of
+[switch.p4](https://github.com/p4lang/switch/tree/f219b4f4e25c2db581f3b91c8da94a7c3ac701a7/p4src)
+[here](http://bmv2.org/switch_egress.png).
```
sethfowler: Should this just live in this repo?

backends/graphs/graphs.cpp
sethfowler: Yeah, seems like this should be a `#error` and we just shouldn't build this backend at all if we don't have the boost graphs library. It'd be easier, of course, to just make it a hard dependency.

backends/graphs/controls.h
sethfowler: Heh, actually, after seeing the code *below* the portion that Mihai quoted (as opposed to the portion above, which is what shows up in the Github summary) I actually agree with Mihai on this. You're probably better off not having the blank line between every `preorder` method.

Sorry for the noise, Mihai. =)

backends/graphs/controls.h
```diff
@@ -0,0 +1,119 @@
...
+
+    bool preorder(const IR::P4Table *table) override;
+
+    void writeGraphToFile(const Graph &g, const cstring &name);
+
+ private:
+    P4::ReferenceMap *refMap; P4::TypeMap *typeMap;
+    const cstring graphsDir;
+    Graph *g{nullptr};
+    vertex_t root{};
```
sethfowler: I feel like writing `vertex_t root{};` doesn't add much over just `vertex_t root;`.

backends/graphs/CMakeLists.txt
antoninbas: updating comment and commands

backends/graphs/controls.h
antoninbas: removing empty lines between `preorder` method overloads

backends/graphs/controls.cpp
antoninbas: making the change

backends/graphs/controls.h
antoninbas: removing old data members

backends/graphs/graphs.cpp
antoninbas: I'd rather keep it as it is TBH; using -o would be confusing IMO

backends/graphs/README.md
antoninbas: I'm indeed planning on checking-in a smaller example in this repo and embedding the graph in the README

backends/graphs/README.md
sethfowler: Oh, I see. I suppose you'll need a directory then. =)

backends/graphs/controls.cpp
antoninbas: changing to use `cstring` consistently; guess it makes sense since all the strings I use are pretty much meant to be immutable

backends/graphs/README.md
antoninbas: improving the shapes

backends/graphs/graphs.cpp
antoninbas: updating to not compile graphs backend if boost graph headers not found

qin-nz:
I can't find `p4c-graphs` in build/backends/graphs.
Something wrong in CMakeLists.txt ?
image

cc10512:
You need to install the boost graph library, otherwise the graphs backend does not build. It is currently set as optional.

PR #837, opened 2017-08-10
Bug-fixes for ebpf code generation

mbudiu-vmw:
The changes look bigger than they actually are: most code changed because it is now guarded by if statements.


backends/ebpf/ebpfControl.cpp
```diff
@@ -271,6 +270,9 @@ void ControlBodyTranslator::processApply(const P4::ApplyMethod* method) {
     }
     builder->blockStart();
 
+    BUG_CHECK(method->expr->arguments->size() == 0, "%1%: table apply with arguments", method);
+#if 0
+    binding.populate(method->getActualParameters(), method->expr->arguments);
```
ChrisDodd: You should only comment stuff out with `#if 0` if you expect it will shortly be resurrected, in which case you should also add a comment to that effect (describe what/why on the change and what it is waiting for)

Otherwise, just remove the code.

PR #843, opened 2017-08-14
Sugar for making constructing calls easier.

ChrisDodd:
Avoid including the same support file multiple times when converting P4_14->16


frontends/p4/fromv1.0/programStructure.cpp
```diff
@@ -476,7 +472,9 @@ void ProgramStructure::createParser() {
     conversionContext.clear();
 }
 
-void ProgramStructure::include(cstring filename) {
+void ProgramStructure::include(cstring filename, cstring ppoptions) {
+    if (included_files.count(filename)) return;
```
mbudiu-vmw: Could this break the semantics of `#include` by preventing the inclusion of a file multiple times?

ChrisDodd: This isn't used for `#include` -- it's used to pull in (target specific) header files when translating some construct that requires said header file.  For this, it's nice to not have to worry about including the file mulitple times if mulitple such constructs are used.

PR #845, opened 2017-08-14
Fix the use of unofficial GC APIs

DGeorgeNetro:
The variable GC_start_call_back was being used directly, but is not part of the
official library interface, ie marked with GC_API. This means that some systems
may not have this call available (notably Windows/MSYS2). GC version 7.2.0
introduced an API GC_set_start_callback() for setting this callback. This
commit uses that call and places a minimum version of 7.2.0 for p4c builds.

Any manipulation of GC_print_stats has been bypassed on Windows systems,
which is also an unofficial and unavailable API on that system. There appears
to be no official way of supporting that.

Finally, the extern declaration for GC_get_heap_size was removed as it is
part of the official APIs.


DGeorgeNetro:
Pull request signed-off-by: David George <david.george@netronome.com>

DGeorgeNetro:
This PR wasn't complete, I will resubmit a complete version.

ChrisDodd:
Please do submit this -- this kind of portability and version verification is very needed.

PR #847, opened 2017-08-14
converts boolean expressions comparing bools to ints

cc10512:
Convert the expression `(boolExpr == constExpr)` to either
`(boolExpr)` or `(! boolExpr)` depending on the value of the
constant. Similarly for `(boolExpr != constExpr)` and permutations of
the expressions.

### DIFF
--- a/frontends/p4/fromv1.0/converters.cpp
+++ b/frontends/p4/fromv1.0/converters.cpp
@@ -195,11 +195,11 @@ const IR::Node* ExpressionConverter::postorder(IR::Equ *equ) {
 
     auto val = constExpr->to<IR::Constant>()->asInt();
     if (val == 1)
-        return boolExpr; // == 1 return the boolean
+        return boolExpr;  // == 1 return the boolean
     else if (val == 0)
-        return new IR::LNot(equ->srcInfo, boolExpr); // return the !boolean
+        return new IR::LNot(equ->srcInfo, boolExpr);  // return the !boolean
     else
-        return new IR::BoolLiteral(equ->srcInfo, false); // everything else is false
+        return new IR::BoolLiteral(equ->srcInfo, false);  // everything else is false
 }
 
 /// And the Neq
@@ -223,7 +223,7 @@ const IR::Node* ExpressionConverter::postorder(IR::Neq *neq) {
     else if (val == 1)
         return new IR::LNot(neq->srcInfo, boolExpr);
     else
-        return new IR::BoolLiteral(neq->srcInfo, true); // everything else is true
+        return new IR::BoolLiteral(neq->srcInfo, true);  // everything else is true
 }
 
 const IR::Node* StatementConverter::preorder(IR::Apply* apply) {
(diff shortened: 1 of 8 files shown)


frontends/p4/fromv1.0/converters.cpp
```diff
@@ -176,6 +176,53 @@ const IR::Node* ExpressionConverter::postorder(IR::GlobalRef *ref) {
             new IR::Path(ref->srcInfo, IR::ID(ref->srcInfo, ref->toString())));
 }
 
+/// P4_16 is stricter on comparing booleans with ints
+/// Therefore we convert such expressions into simply the boolean test
+const IR::Node* ExpressionConverter::postorder(IR::Equ *equ) {
+  const IR::Expression *boolExpr = nullptr;
+  const IR::Expression *constExpr = nullptr;
+  if (equ->left->type->is<IR::Type_Boolean>() &&
+      (equ->right->type->is<IR::Type_InfInt>() && equ->right->is<IR::Constant>())) {
```
ChrisDodd: Why check for InfInt here?  Something like `hdr.valid == 1w1` would have a constant that is not an InfInt.  Something like `hdr.valid == true` is legal already but could be simplified to just `hdr.valid`.

It seems like just removing the check for InfInt would deal with all these cases.  Something like `hdr.valid == "foo"` should fail typechecking before this.

frontends/p4/fromv1.0/converters.cpp
```diff
@@ -176,6 +176,53 @@ const IR::Node* ExpressionConverter::postorder(IR::GlobalRef *ref) {
...
+  } else if (equ->right->type->is<IR::Type_Boolean>() &&
+             (equ->left->type->is<IR::Type_InfInt>() && equ->left->is<IR::Constant>())) {
+    boolExpr = equ->right;
+    constExpr = equ->left;
+  }
+
+  if (boolExpr == nullptr)
+    return equ;
+
+  if (constExpr->to<IR::Constant>()->asInt() == 0)
```
ChrisDodd: What if someone writes something like `hdr.valid == 2`?  Should it be an error?  or always false?

cc10512: I followed the "C" convention: 0 is `false`, any other integer value is `true`.  Since we already have a boolean expression, testing it for equality with any other value than `0` or `1` is likely to be an error, so I should let it fail with the previous error by returning the original expression.

ChrisDodd: In C a comparison like `boolExpr == 2` will always be false (it converts bool->int and does an integer compare).  This code will convert it to just `boolExpr`, which is the equivalent of doing an int->bool conversion followed by a boolean compare.

frontends/p4/fromv1.0/converters.cpp
```diff
@@ -176,6 +176,53 @@ const IR::Node* ExpressionConverter::postorder(IR::GlobalRef *ref) {
             new IR::Path(ref->srcInfo, IR::ID(ref->srcInfo, ref->toString())));
 }
 
+/// P4_16 is stricter on comparing booleans with ints
+/// Therefore we convert such expressions into simply the boolean test
+const IR::Node* ExpressionConverter::postorder(IR::Equ *equ) {
```
mbudiu-vmw: Indentation should use 4 spaces.

cc10512:
@ChrisDodd: updated based on your comments. Also indented at 4 spaces.

PR #849, opened 2017-08-15
Fix use of unexported GC library variables

DGeorgeNetro:
The variable GC_start_call_back was being used directly, but is not part of the
official library interface, ie marked with GC_API. This means that some systems
may not have this call available (notably Windows/MSYS2). GC version 7.2.0
introduced an API GC_set_start_callback() for setting this callback. This
commit uses that call and places a minimum version of 7.2.0 for p4c builds.

Any manipulation of GC_print_stats has been bypassed on Windows systems,
which is also an unofficial and unavailable API on that system. There appears
to be no official way of supporting that.

Finally, the extern declaration for GC_get_heap_size was removed as it is
part of the official APIs.

Signed-off-by: David George <david.george@netronome.com>
Acked-by: Rando Wiesemann <rando.wiesemann@netronome.com>


lib/gc.cpp
```diff
@@ -51,24 +59,29 @@ void *operator new[](std::size_t size) {
 void operator delete(void *p) _GLIBCXX_USE_NOEXCEPT { return gc::operator delete(p); }
 void operator delete[](void *p) _GLIBCXX_USE_NOEXCEPT { return gc::operator delete(p); }
 
-extern "C" void (*GC_start_call_back)(void);
-extern "C" size_t GC_get_heap_size(void);
+#ifndef NO_GC_PRINT_STATS
```
cc10512: Would you mind avoiding double negation? That is, you can use cmakedefine to check for the GC_print_stats function, define HAS_GC_PRINT_STATS if it exists, and then use that to factor out the code. Then, you will not have any platform dependent code in your source.

lib/gc.cpp
```diff
@@ -51,24 +59,29 @@ void *operator new[](std::size_t size) {
...
 
 void silent(char *, GC_word) {}
 #endif  /* HAVE_LIBGC */
 
 void setup_gc_logging() {
 #if HAVE_LIBGC
-    GC_print_stats = LOGGING(2) ? 1 : 0;  // unfortunately goes directly to stderr!
-    GC_start_call_back = gc_callback;
+#ifndef NO_GC_PRINT_STATS
+    GC_print_stats = LOGGING(2) ? 1 : 0;
```
mbudiu-vmw: It looks like you have lost a comment.

lib/gc.cpp
DGeorgeNetro: Will do - using cmake to do the check is much better.

lib/gc.cpp
DGeorgeNetro: sorry - will fix

sethfowler:
By unfortunate coincidence, #852 touches this code as well. Just a heads up. It shouldn't be hard for either of us to rebase around the other's changes.

DGeorgeNetro:
It might be easier to submit a new PR with the above feedback on a new branch. If there are no objections, I'll do that.

sethfowler:
If my changes happen to land first, you can rebase your local branch and then force push it to update this PR. You don't need a new one.

CMakeLists.txt
```diff
@@ -67,9 +67,14 @@ check_include_file (execinfo.h HAVE_EXECINFO_H)
 check_include_file (ucontext.h HAVE_UCONTEXT_H)
 
 # check functions
+
+# set libraries to be used with check_function_exists
+set (CMAKE_REQUIRED_LIBRARIES ${CMAKE_REQUIRED_LIBRARIES} gc)
```
cc10512: Please use `${LIBGC_LIBRARIES}` to account for non-standard installations. The find_package command will return the library with the right path.
Also, please reset `CMAKE_REQUIRED_LIBRARIES` after the check. You do not necessarily want libgc to be linked to all targets.

Issue #838, opened 2017-08-10
p4test segfaults when provided with two -v flags


hanw:
happens with any p4-16 problem.

```
./p4test -I p4include --p4v=16 ~/p4-tests/mint.p4 -v -v
Invoking preprocessor
cc -E -x c -Wno-comment -C -undef -nostdinc  -Ip4include -I/Users/hanwang/bf-p4c-compilers/build/p4c/./p4include /Users/hanwang/p4-tests/mint.p4
Parsing P4-16 program /Users/hanwang/p4-tests/mint.p4
Segmentation fault: 11
```

mbudiu-vmw:
I cannot reproduce this problem; I have run successfully p4test with two -v flags using test programs from the testdata/p4_16_samples folder.

hanw:
With two -v flags, I had the issue with a program that is using another architecture definition (not v1model).

With three -v flags, following error happens.
518655.588:gc:1:****** GC called ****** (heap size 479232)
518655.588:gc:1:****** GC called ****** (heap size 479232)
518655.588:gc:1:****** GC called ****** (heap size 479232)
518655.588:gc:1:****** GC called ****** (heap size 479232)
518655.588:gc:1:****** GC called ****** (heap size 479232)
Segmentation fault: 11

mbudiu-vmw:
Can you please supply a program that reproduces this problem? apply_if is not part of the testsuite.
I have tried several of the tests and none has failed yet.

sethfowler:
Han, can you still reproduce now that #852 has landed?

mbudiu-vmw:
If this is fixed perhaps we can close it.

PR #852, opened 2017-08-16
Improve InstantiatedBlock, Vector, and JSONLoader APIs, and fix a GC bug

sethfowler:
This PR contains a number of small improvements and fixes that didn't seem worthy of their own individual PRs.

- `InstantiatedBlock::getParameterValue()` calls `BUG()` if the parameter doesn't exist. This PR adds `InstantiatedBlock::findParameterValue()` for cases where the parameter may not exist; it returns null on failure.
- It's a common idiom in p4c to take a Node and either append it to a Vector or, if it is itself a Vector, append its content to the Vector. Now this idiom is captured in the new method `Vector::pushBackOrAppend()`.
- Calling `LOG()` macros, which may allocate memory, is unsafe in `gc_callback()`. I've seen this trigger stack overflows, but not anymore, because the issue is fixed in this PR.
- Sometimes it's both more convenient and more efficient to return a value type from `fromJSON()`. Now you can.

### DIFF
--- a/lib/gc.cpp
+++ b/lib/gc.cpp
@@ -56,10 +56,17 @@ extern "C" size_t GC_get_heap_size(void);
 extern "C" int GC_print_stats;
 
 static void gc_callback() {
-    size_t count;
-    LOG1("****** GC called ****** (heap size " << GC_get_heap_size() << ")");
-    LOG2("cstring cache size " << cstring::cache_size(count) << " (count=" << count << ")");
-    GC_print_stats = LOGGING(2) ? 1 : 0;  // unfortunately goes directly to stderr!
+    if (Log::verbose())
+        std::clog << "****** GC called ****** (heap size " << GC_get_heap_size() << ")";
+
+    if (Log::verbosity() >= 2) {
+        size_t count;
+        std::clog << "cstring cache size " << cstring::cache_size(count)
+                  << " (count=" << count << ")";
+    }
+
+    // Maybe print GC statistics. Unfortunately they go directly to stderr!
+    GC_print_stats = Log::verbosity() >= 2 ? 1 : 0;
 }
 
 void silent(char *, GC_word) {}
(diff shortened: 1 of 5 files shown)


sethfowler:
It's worth noting that I actually doubt the remaining code in `gc_callback()` is truly safe, since we override the global `operator new` to allocate using libgc, but in practice this patch is enough to eliminate the stack overflows I was seeing. To be more thorough, we'd probably want to move all logging (and more generally anything that could allocate) out of `gc_callback()` altogether.

lib/gc.cpp
```diff
@@ -56,10 +56,17 @@ extern "C" size_t GC_get_heap_size(void);
 extern "C" int GC_print_stats;
 
 static void gc_callback() {
-    size_t count;
-    LOG1("****** GC called ****** (heap size " << GC_get_heap_size() << ")");
-    LOG2("cstring cache size " << cstring::cache_size(count) << " (count=" << count << ")");
-    GC_print_stats = LOGGING(2) ? 1 : 0;  // unfortunately goes directly to stderr!
+    if (Log::verbose())
```
mbudiu-vmw: Hopefully this will fix #838

sethfowler: Good call! That looks like the same symptom I was seeing.

ChrisDodd: This subtljy changes the logging behavior, basing it off of the global -v level rather than the -Tgc option, which defaults to one less than the -v level.  So you've changed these to require one less -v option to be triggered.  Will using the `LOGGING` macro run into the same memory allocation issues?

sethfowler: That defaulting to one less than the `-v` level has always confused me, honestly. We can change the levels in this code to maintain the previous behavior if you'd like.

`LOGGING()` will run into the same issues, yes. It can allocate.

ChrisDodd: I guess to do this safely, we need to lookup the logging level in `setup_gc_logging`, store it in a static var, and use that static var in `gc_callback`....

sethfowler: To truly do it safely I think we'll need to use C APIs and get rid of iostreams in this function, because I doubt that there's any guarantee that iostreams doesn't allocate.

PR #856, opened 2017-08-17
allow counter and meter enum names to be all capitalized

hanw:


ir/v1.def
```diff
@@ -297,9 +297,10 @@ abstract Stateful : Attached {
 abstract CounterOrMeter : Stateful {
     CounterType   type = CounterType::NONE;
     void settype(cstring t) {
-        if (t == "packets") type = CounterType::PACKETS;
-        else if (t == "bytes") type = CounterType::BYTES;
-        else if (t == "packets_and_bytes") type = CounterType::BOTH;
+        if (t == "packets" || t == "PACKETS") type = CounterType::PACKETS;
+        else if (t == "bytes" || t == "BYTES") type = CounterType::BYTES;
```
ChrisDodd: Should we allow mixed-case names like Packets and Bytes?  Maybe just a case-insensitive stricmp (strcasecmp?)

ir/v1.def
```diff
@@ -297,9 +297,10 @@ abstract Stateful : Attached {
...
+        else if (t == "packets_and_bytes" ||
+                 t == "PACKETS_AND_BYTES") type = CounterType::BOTH;
```
ChrisDodd: and then what about PacketsAndBytes -- that way lies madness.

hanw:
Thanks. Changed the code to use strcasecmp.

PR #857, opened 2017-08-18
Bmv2 backend calculations const fix

antoninbas:
This fixes a bug reported by @jnfoster 
This was caused by an omission in the bmv2 JSON format doc, which I fixed this morning: https://github.com/p4lang/behavioral-model/pull/426


antoninbas:
@hanw I saw some code duplication which I removed, please tell me if I missed something

Issue #840, opened 2017-08-12
Compiler bug: statements in compute checksum trigger "not handled yet" error with Bmv2 backend


jnfoster:
Compiling the following program 

```
#include <core.p4>
#include <v1model.p4>

header h_t {
  bit<32> src;
  bit<32> dst;
  bit<16> csum;
}

struct metadata {
}
struct headers {
  h_t h;
}

parser MyParser(packet_in packet, out headers hdr, inout metadata meta, inout standard_metadata_t standard_metadata) {
    state start {
        packet.extract(hdr.h);
        transition accept;
    }
}
control MyVerifyChecksum(in headers hdr, inout metadata meta) {
    apply {
    }
}
control MyIngress(inout headers hdr, inout metadata meta, inout standard_metadata_t standard_metadata) {
    apply {
    }
}
control MyEgress(inout headers hdr, inout metadata meta, inout standard_metadata_t standard_metadata) {
    apply {
    }
}
control MyDeparser(packet_out packet, in headers hdr) {
    apply {
    }
}
control MyComputeChecksum(inout headers hdr, inout metadata meta) {
    Checksum16() checksum;
    h_t h = hdr.h;
    apply {
        h.csum = 0;
	hdr.h.csum = checksum.get(h);
   }    
}
V1Switch(MyParser(), MyVerifyChecksum(), MyIngress(), MyEgress(), MyComputeChecksum(), MyDeparser()) main;
```
generates the following error when compiled using `p4c-bm2-ss`
```
In file: /home/parallels/p4c/backends/bmv2/simpleSwitch.cpp:583
Compiler Bug: foo.p4(42): h_1/h.csum = 0;: not handled yet
        h.csum = 0;
               ^
```

jnfoster:
Also, what is `h_1/h` ??

mbudiu-vmw:
The compiler bug messages do not have to be user-friendly.
Each identifier in the compiler has two names: the internal name (h1), which is unique, and the external name used by the program writer (h). The internal name is needed to disambiguate P4-14 names and for eliminating name clashes; the external name is needed for giving error reports. 

We have many duplicated bugs related to checksums in v1model. 
This particular program will probably never work in v1model, but we should give a better error message.

mbudiu-vmw:
The new checksum API makes this issue invalid, so I will close it.

Issue #775, opened 2017-07-13
P4_14 to P4_16 checksum conversion gives wrong comparison in verifyChecksum() control block


jafingerhut:
If you run the command below to auto-convert the P4_14 program in p4c/testdata/p4_14_samples/checksum.p4 to a supposedly equivalent P4_16 program:
```
p4test --p4v 14 --pp checksum-p4-16.p4 checksum.p4
```
The output P4_16 program has this in the verifyChecksum() control block:
```
control verifyChecksum(in headers hdr, inout metadata meta) {
    Checksum16() ipv4_checksum;
    apply {
        if (hdr.ipv4.ihl == 4w5 && hdr.ipv4.hdrChecksum == ipv4_checksum.get({ hdr.ipv4.version, hdr.ipv4.ihl, hdr.ipv4.diffserv, hdr.ipv4.totalLen, hdr.ipv4.identification, hdr.ipv4.flags, hdr.ipv4.fragOffset, hdr.ipv4.ttl, hdr.ipv4.protocol, hdr.ipv4.srcAddr, hdr.ipv4.dstAddr })) 
            mark_to_drop();
    }
}
```
Unless I am very confused (possible!) I think the comparison between hdr.ipv4.hdrChecksum and the return value of ipv4_checksum.get() should be not-equals (!=), not equals (==).

The code in the auto-generated computeChecksum control block appears to me to be doing the correct thing.

jnfoster:
+1 

(Just observed this bug in a different converted program.)

mbudiu-vmw:
We are having lots of problems with the checksum. I am thinking to change v1model.p4 to handle checksums differently, and more in line with bmv2 capabilities.

```
extern void checksum_update<T>(in T fields, out bit<16> checksum);
extern void checksum_verify(in T fields);
```

(Maybe the second method should also have a `bit<16> checksum` argument, haven't decided yet.)
Then:
* in the verify control you are only allowed to have calls to checksum_verify.
* in the update control you are only allowed to have calls to checksum_update.

The existing Checksum16 extern would be removed.

This will make code generation for bmv2 much more robust, and also will enable us to give much better error messages.

Note that neither control can drop packets.
@jafingerhut @cc10512 @ChrisDodd @jnfoster @antoninbas : I would appreciate comments.

ChrisDodd:
Seems reaonable to me -- allowing general code in checksum blocks seems problematic.  @sethfowler any opinions?

jafingerhut:
[...]

Issue #741, opened 2017-06-27
p4c-bm2-ss compiler bug: could not convert Checksum16 to Json


tomrodeheffer:
I wrote the following program to try out the Checksum16 extern: [checksum.p4.txt](https://github.com/p4lang/p4c/files/1104224/checksum.p4.txt)
But p4c-bm2-ss says:

```
In file: ../backends/bmv2/expression.cpp:82
Compiler Bug: checksum.p4(38): ck_1/ck;: could not convert to Json
    hdr.h.cksm = ck.get(hdr.h);
                 ^^
```

Is Checksum16 supported in p4c-bm2-ss?  If not, I would expect to see some remark about that in the documentation and a better error message from the compiler.

mbudiu-vmw:
The documentation is really lacking, indeed.
Checksums can be only instantiated in the verifyChecksum and updateChecksum controls.
The error message is bad, so we'll keep this as a bug.

jafingerhut:
I was running a test case yesterday with someone using v1model.p4 using the 2017-Jun-19 latest versions of p4c-bm2-ss and p4lang/behavioral-model's simple_switch, and it seems that nothing in the verifyCheckum and updateChecksum controls is executed in simple_switch.  Not 100% sure, but we had some evidence pointing that way.

mbudiu-vmw:
BMv2 ignores checksum verify statements, even in P4-14.
If the update does not work then you can't really modify IPv4 headers.
We have issue #655, which still seems to be open. @hanw assigned himself for this one.

antoninbas:
Actually bmv2 now logs an error message if a packet has an invalid checksum, but does not drop the packet. This was implemented a couple months ago.

mbudiu-vmw:
I hope that the new checksum APIs in v1model make this a non-issue,

Issue #675, opened 2017-05-27
bmv2 backend does not generate json for 'verify checksum' in v1model.


hanw:
Follow up with issue #655, the verify checksum block need to be fixed to drop the packets with a bad checksum.

jafingerhut:
Is this a suggested change in the P4 program so that it makes the necessary actions to drop the packet if it detects a bad checksum?

Or are you thinking of some kind of change where the architecture automatically drops a packet, regardless of whether the P4 code directs this to happen?

I am hoping you are thinking of the first approach.

hanw:
I am thinking of the first one. If the `verifyChecksum` control block in v1model computes and checks the checksum value, and decide to drop the packet if the checksum does not match to an expected value, then the packet should be dropped. An example is in issue655.p4.

hanw:
I am not sure which part of the JSON should be used for the verifyChecksum control block, maybe @antoninbas can chime in here.

vgurevich:
Dropping packets with bad IPv4 header checksum is only one option. Another option is, for example, to allow the Ingress() control to see them, so that they could be counted, copied to CPU, etc. 

Another concern that I have is that parser generates an `error`, but verifyChecksum does not have this ability, so from the Ingress() control perspective different errors that occurred before it, need to be processed using completely different mechanisms.

mbudiu-vmw:
This is a limitation of v1model. The language allows many more behaviors.

jafingerhut:
@vgurevich verifyChecksum can store the value of an error into a metadata variable of type error, one that is passed out of verifyChecksum and to a next control block, yes?  That would require a line or three of explicit P4 code, and the error value would be a second one, over and above the one filled in by the parser, so 2 error variable values might need to be checked in later places rather than one, but checking both could be limited to one place, that then combines the two error variables into one combined value if the programmer wished to do so.

I am probably not saying anything you haven't thought already here.  It just doesn't seem to me like "a completely different mechanism".  It is just explicit P4 code, not something hidden beneath the implementation of the parser.

vgurevich:
@mbudiu-bfn,

I agree that these are limitations of v1model, but I want to make sure we articulate them so that PSA does not have them.

Here are a couple of things the problem might be solved:
# Remove verifyChecksum() control, allow using Checksum16() extern in the parser instead. To me this sounds like the cleanest solution, since we can then use verify() statement то generate checksum errors as well, thus moving all parser-related checks into one place with one mechanism.
[...]

PR #858, opened 2017-08-19
New checksum APIs in v1model; better alias analysis in side-effects; …

mbudiu-vmw:
…removed some hacks from bmv2 backend which were related to checksums

This patch still does not generate JSON code for verifying checksum, but it should be very easy to add that capability now. There is a whole set of issues that should be addressed by this PR; I will list them separately.


mbudiu-vmw:
This PR should fix the following issues: #134, #445, #609, #741, #775, #840

mbudiu-vmw:
This PR simplifies the way checksums are computed and verified in v1model by using an extern function for each API; no extern objects exist anymore. This is much closer to the way BMv2 (and P4-14) handle checksums, so translation is much simpler in both directions. This new API also makes it much easier to give good error messages when users misuse the verify/update checksum controls.

The better alias analysis was needed so that the compiler does not mess up these nice extern function calls. On the positive side, a bunch of relatively hacky code to handle checksums in the BMv2 back-end has been completely removed.

frontends/p4/sideEffects.cpp
```diff
@@ -310,40 +310,206 @@ class DismantleExpression : public Transform {
...
-    // We don't want to compute the full read/write set here so we
-    // overapproximate it as follows: all declarations that occur in
-    // an expression.
-    // TODO: this could be made more precise, perhaps using LocationSets.
-    class ReadsWrites : public Inspector {
+    /// This class represents the path to a location.
+    /// Given a struct S { bit a; bit b; } and a variable S x;
+    /// a path can be x.a, or just x.  An array index is represented as a
+    /// number (encoded as a string) or as "*", denoting an unknown index.
+    struct LocationPath : public IHasDbPrint {
```
antoninbas: Would have been nice to make the changes to this pass in a separate PR

p4include/v1model.p4
```diff
@@ -140,6 +135,27 @@ extern void clone3<T>(in CloneType type, in bit<32> session, in T data);
...
+Verifies the 16-bit checksum of the supplied data.
+T must be a list expression where all the fields are bit-fields or varbits
+and where the total dynamic length of the fields is a multiple of 16 bits.
+If this method detects that a checksum of the data is not correct it
+sets an internal error flag.
+@param condition  If 'false' the verification always succeeds.
+@param data       Data whose checksum is verified.
+@param checksum   Expected checksum of the data.
+*/
+extern void verify_checksum<T>(in bool condition, in T data, in bit<16> checksum);
```
antoninbas: is there a reason to limit these new extern methods to the "csum16" algorithm?
why not support `verify_checksum<T, U>(in bool condition, in T data, in U checksum, in HashAlgorithm algo)`?

p4include/v1model.p4
```diff
@@ -140,6 +135,27 @@ extern void clone3<T>(in CloneType type, in bit<32> session, in T data);
 
 extern void truncate(in bit<32> length);
 
+/**
+Verifies the 16-bit checksum of the supplied data.
+T must be a list expression where all the fields are bit-fields or varbits
+and where the total dynamic length of the fields is a multiple of 16 bits.
+If this method detects that a checksum of the data is not correct it
+sets an internal error flag.
```
antoninbas: isn't the plan to expose this information as standard_metadata which can be consumed in ingress?

frontends/p4/sideEffects.cpp
mbudiu-vmw: yes, but this PR would depend on that other one for being merged first.

p4include/v1model.p4
mbudiu-vmw: Yes, but that will be a separate PR.

p4include/v1model.p4
mbudiu-vmw: No, but currently this is the only algorithm supported - at least in the back-end, so I didn't want to make the API more complicated than necessary.

antoninbas: I think that was my question. Is there a reason not to support the other algorithms in the backend? It looks like you would just need to map the enum value to the algo name and call `createCalculation` with that name as the first argument. It may be worth it to add support now rather than modify v1model.p4 in a non-backward compatible fashion in the future.

mbudiu-vmw: I will amend the PR with support for an algorithm parameter.
Only csum16 will be supported initially.

mbudiu-vmw:
@antoninbas : this additional commit adds the Algorithm argument to the two checksum methods.

antoninbas:
@mbudiu-vmw I believe that the destination field's type need to be a generic type (e.g. needs to be `bit<32>` for a crc32 algorithm).

mbudiu-vmw:
I hope I have addressed all of @antoninbas comments.
Still pending: doing checksum verification - issue #675 - and separating verification from update.
I will probably have a single separate PR for both of these.

PR #865, opened 2017-08-26
Fix for issue #675

mbudiu-vmw:
With this fix the BMv2 back-end also verifies checksums.
This PR also adds an intrinsic metadata field to be set by verify_checksum when verification fails. This will require BMv2 support; currently the field is unused.

Unfortunately, the BMv2 API for checksum verification requires a field for the checksum; it cannot be a constant. In consequence, to enforce this in P4-16 I had to change v1model.p4 to take an inout parameter for the checksum:
```
extern void verify_checksum<T, O>(in bool condition, in T data, inout O checksum, HashAlgorithm algo);
```

This requires the corresponding control headers to also be inout:

```
control VerifyChecksum<H, M>(inout H hdr, inout M meta);
```

This has caused a cascade of changes to all reference outputs and tests that use v1model.


mbudiu-vmw:
This PR requires another PR that has been reverted to be merged beforehand.

mbudiu-vmw:
I will close this PR and submit a new one which supersedes it.

PR #866, opened 2017-08-28
Update include files whe rebuilding

ChrisDodd:
I couldn't figure out a way of having this happen only when the include files change, but the copy is fast enough that it probably doesn't matter.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -217,6 +217,13 @@ add_custom_command (OUTPUT ${IR_GENERATED_SRCS}
 
 add_custom_target(genIR DEPENDS ${IR_GENERATED_SRCS})
 
+# Header files 
+add_custom_target(update_includes ALL
+  #FIXME -- should only run this when headers change -- how to accomplish that?
+  COMMAND ${CMAKE_COMMAND} -E copy_directory ${P4C_SOURCE_DIR}/p4include ${P4C_BINARY_DIR}/p4include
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -217,6 +217,13 @@ add_custom_command (OUTPUT ${IR_GENERATED_SRCS}
 
 add_custom_target(genIR DEPENDS ${IR_GENERATED_SRCS})
 
+# Header files 
+add_custom_target(update_includes ALL
+  #FIXME -- should only run this when headers change -- how to accomplish that?
+  COMMAND ${CMAKE_COMMAND} -E copy_directory ${P4C_SOURCE_DIR}/p4include ${P4C_BINARY_DIR}/p4include
+  COMMAND ${CMAKE_COMMAND} -E copy_directory ${P4C_SOURCE_DIR}/p4_14include ${P4C_BINARY_DIR}/p4_14include
+)
```
cc10512: add a DEPENDS directive so that you make it depend only on the include files, and use `cmake -E copy_if_different`

ChrisDodd: `copy_if_different` gives me an error trying to copy a directory (apparently only works for files?)

ChrisDodd: If I add a `DEPENDS` and remove the `ALL`, it then no longer runs the commands regardless of whether things are out of date or not.

PR #867, opened 2017-08-28
Added "deletePasses" method to PassManager.

zma-barefoot:
needed to delete existing passes and re-populate new passes.

### DIFF
--- a/ir/pass_manager.h
+++ b/ir/pass_manager.h
@@ -36,6 +36,7 @@ class PassManager : virtual public Visitor, virtual public Backtrack {
     void addPasses(const std::initializer_list<Visitor *> &init) {
         never_backtracks_cache = -1;
         for (auto p : init) if (p) passes.emplace_back(p); }
+    void deletePasses() { passes = vector<Visitor *>(); }
     void runDebugHooks(const char* visitorName, const IR::Node* node);
     profile_t init_apply(const IR::Node *root) override {
         running = true;


ir/pass_manager.h
```diff
@@ -36,6 +36,7 @@ class PassManager : virtual public Visitor, virtual public Backtrack {
     void addPasses(const std::initializer_list<Visitor *> &init) {
         never_backtracks_cache = -1;
         for (auto p : init) if (p) passes.emplace_back(p); }
+    void deletePasses() { passes = vector<Visitor *>(); }
```
cc10512: where is it used?

ChrisDodd: Not sure why this would be needed -- generally passes are only added when the PassManager is created and never change.  The only reason `addPasses` exists (rather than just using the initializer_list constructor) is historical accident.

zma-barefoot: In the backend refactor. I would like to fold PHV passes into a PassManager, and because of PHV analysis and allocation are split before and after table placement, need a way to clear and re-populate passes.

zma-barefoot: Please see reply to cc10512 above.

sethfowler: This sounds suspiciously like you're solving the wrong problem, though I'm prepared to be convinced otherwise. *Without giving details about Barefoot-internal code on this public repo*, could you please give an example of a thing you're trying to do that you can't do with the existing PassManager interface?

PR #868, opened 2017-08-29
Revert "New checksum APIs in v1model; better alias analysis in side-e…

cc10512:
…ffects; … (#858)"

This reverts commit f09a95dcda68ca18a1528be6a63a16f2b8074d46.

This commit breaks all existing programs using checksums. This is a
change that needs to be staged more carefully, announced on the
mailing list and deprecate the old way in a more graceful manner,
giving people the chance of transitioning their programs.


mbudiu-vmw:
Where is the original pull request?
We should not lose that code; it was a lot of work to implement it, and eventually it will probably have to be merged.

cc10512:
Agreed, we want to get this in, we just need to stage it better.

The code is in git (nothing really gets lost in git without extra effort!). If you no no longer have the branch, you can still get to the code by doing `git checkout f09a95d` (the sha is in the revert message.

ChrisDodd:
You can also easily reapply it on top of any branch by checking out that branch and doing `git cherry-pick f09a95d`

PR #872, opened 2017-08-30
P4_14 casts that change both size and signedness need two explicit casts

ChrisDodd:


ChrisDodd:
Yes, I just searched for all the places in the P4_14 typechecker where it was creating IR::Cast nodes -- there should be none before typecheck as P4_14 does not have explicit casts in the language.  I can't think of other places in the P4_14 code where they might be introduced as we don't do any other transforms on P4_14 programs before converting to P4_16.

Issue #870, opened 2017-08-29
Adding P4 Slice to EBPF backend causes extra text in Slice statement line 


familysingh:
I am using a modified EBPF backend to shoehorn P4 Slice code into this backend.  I have modified this file: “p4c/testdata/p4_16_samples/switch_ebpf.p4” as follows:

Added new struct below.

struct Token
{
    bit<64>    data;
};

Added the new struct as member of Headers_t

struct Headers_t
{
    Ethernet_h ethernet;
    IPv4_h     ipv4;
    Token      token;
}

Added one line to the control below for use of token.data.

control pipe(inout Headers_t headers, out bool pass)
{
    action Reject(IPv4Address add)
    {
        pass = false;
        headers.ipv4.srcAddr = add;
        headers.token.data[63:32] = add;  // <==== this is the line added
    }
…

I have modified p4c/backends/ebpf/codeGen.h and .cpp.  The .cpp has added this function and thus the modification to the .h file is obvious.

bool CodeGenInspector::preorder(const IR::Slice* expression) {
    int h = expression->getH();
    int l = expression->getL();
  
    std::cout << expression->toString() << h << " " << l  << "\n";    
    builder->append(expression->toString());
    return true;
}

When I use p4c-ebpf on switch_ebpf.p4, I see the printed output from code in the preorder Slice and the preorder AssigmentStatement functions showing correct text for this line of P4 code:

headers.token.data[63:32] = add;

However, the output file from p4c-ebpf adds extra text after the sliced data.  The text is shown in highlight below.

headers.token.data[63:32]headers.token.data6332 = value->u.Reject.add;     

Please test this issue and see where the problem is.

mbudiu-vmw:
I hope that we can close this one.

familysingh:
Sure, please go ahead and close this issue - thanks.

PR #874, opened 2017-08-30
Fix for issue #870: support slice and concat in ebpf back-end

mbudiu-vmw:
Some of this code is copied from the bmv2 back-end.
Arguably, some of this code should be moved into a mid-end pass.

### DIFF
--- a/backends/ebpf/lower.cpp
+++ b/backends/ebpf/lower.cpp
@@ -1,0 +1,118 @@
+/*
+Copyright 2017 VMware, Inc.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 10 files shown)


backends/ebpf/lower.cpp
```diff
@@ -0,0 +1,118 @@
...
+#include "lower.h"
+
+namespace EBPF {
+
+const IR::Expression* LowerExpressions::shift(const IR::Operation_Binary* expression) const {
+    auto rhs = expression->right;
+    auto rhstype = typeMap->getType(rhs, true);
+    if (rhstype->is<IR::Type_InfInt>()) {
+        auto cst = rhs->to<IR::Constant>();
+        mpz_class maxShift = Util::shift_left(1, LowerExpressions::maxShiftWidth);
```
ChrisDodd: This seems like an expensive and obfuscated way of doing `mpz_class(1) << maxShiftWidth`

mbudiu-vmw: I didn't realize that `<<` is defined for mpz_class.
The problem is really in the shift_left method, though.
I wonder whether it has the same semantics; the implementation is according to the spec.

Issue #871, opened 2017-08-30
type parameter on extern method


hanw:
Following code that uses a generic type parameter on extern method triggers a segmentation fault in the frontend.

```
#include <core.p4>

struct psa_parser_input_metadata_t {
}

struct psa_parser_output_metadata_t {
    error parser_error;
}

struct psa_ingress_input_metadata_t {
    error parser_error;
}

struct psa_ingress_output_metadata_t {
}

header user_metadata_t {
    bit<8> value;
}

extern lpf<I> {
    lpf(@optional I instance_count);
    T execute<T>(in T data, @optional in I index);
}

parser Parser ( in psa_parser_input_metadata_t mdi, out psa_parser_output_metadata_t mdo);

control Ingress ( in psa_ingress_input_metadata_t md, inout user_metadata_t umd );

control Deparser ( packet_out p, in user_metadata_t umd );

package Switch( Parser p, Ingress ig, Deparser dp );

parser myParser ( in psa_parser_input_metadata_t mdi,
                  out psa_parser_output_metadata_t mdo) {
    state start {
        mdo.parser_error = error.NoError;
        transition accept;
    }
}

control myIngress ( in psa_ingress_input_metadata_t md, inout user_metadata_t umd ) {
    lpf<bit<32>>() lpf_0;
    table color_table {
        key = {
            umd.value : exact;
        }
        actions = {
            NoAction;
        }
    }
    apply {
        umd.value = lpf_0.execute<bit<8>>(0);
        color_table.apply();
    }
}

control myDeparser( packet_out p, in user_metadata_t umd ) {
    apply {
        p.emit(umd);
    }
}

Switch( myParser(), myIngress(), myDeparser() ) main;
```

mbudiu-vmw:
This triggers an infinite recursion in the type-checking pass.

jafingerhut:
@mbudiu-vmw The latest version of p4c as of 2017-Sep-01 still gives a segmentation fault in my testing when trying to compile the sample program given in the original issue.  Is this what you expect?

mbudiu-vmw:
I hope that this is fixed, so I will close it.

PR #875, opened 2017-08-30
Fix for issue #871

mbudiu-vmw:
In a type variable substitution one has to look-up variables until reaching a fix-point; it is not enough to look up a variable once.

### DIFF
--- a/frontends/p4/substitutionVisitor.cpp
+++ b/frontends/p4/substitutionVisitor.cpp
@@ -35,7 +35,7 @@ const IR::Node* TypeVariableSubstitutionVisitor::preorder(IR::TypeParameters *tp
     for (auto it = tps->parameters.begin(); it != tps->parameters.end();) {
         const IR::Type* type = bindings->lookup(*it);
         if (type != nullptr && !replace) {
-            LOG1("Removing from generic parameters " << *it);
+            LOG3("Removing from generic parameters " << *it);
             it = tps->parameters.erase(it);
         } else {
             if (type != nullptr)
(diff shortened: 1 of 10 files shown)


jnfoster:
@mbudiu-vmw I haven't reviewed your implementation of type inference, unification, and substitution carefully. So it's possible the observation I'm about to make is off the mark. 

But the textbook approach does not require iterating substitution to a fixpoint -- if the substitution is implemented as a function from type variables to types, substitution is simply function application, lifted homomorphically to all types.  In my experience, when it seems that iterating to a fixpoint is needed, the issue is often that the implementation of the unification algorithm has actually produced an incorrect substitution.

To see this, consider this specification of unification from Benjamin Pierce's _Types and Programming Languages_:
tapl-unification
Now consider the third case -- i.e., where the left side of the constraint is a type variable X and the right side is a type T, where X is not free in T. The result is obtained by (i) substituting T for X in the remaining set of constraints C' and recursively unifying the result and then (ii) composing the resulting substitution (as functions!) with the singleton substitution that maps X to T. 

When I assign type inference to students as homework, a very common error is to implement composition incorrectly: they represent substitutions using a data structure like an association list or a finite map, and rather than implementing a proper composition operation, they simply add the mapping from X to T to the result of the recursive call to unify. Then when they notice that some occurrences of X remain after applying the resulting substitution, they think they need to iterate to a fixpoint. However, doing this can lead to other incorrect behaviors.

Again, I'm not sure if that's what's going on here, but I thought I'd pass these observations along just in case. 

If it would help, I'd be glad to make time to do a detailed code review of the type inference implementation with you sometime. It might be an efficient way to get the bugs out.

mbudiu-vmw:
Yes, this is due to the fact that in the past I have optimized the composition of two substitutions for a certain case; you will notice that there are TypeVariableSubtitution::compose and simpleCompose. The latter tries to elude some checks and substitutions. With code evolution some invariants assumed for simpleComposed have become more complicated. One solution is to get rid of simpleCompose and just use compose; then I expect that the transitive closure won't be necessary at lookup time. TypeChecking is one of the most frequently run passes, so this was done in the interest of speed.

frontends/p4/substitutionVisitor.cpp
```diff
@@ -47,22 +47,25 @@ const IR::Node* TypeVariableSubstitutionVisitor::preorder(IR::TypeParameters *tp
...
+        // This will end because there should be no circular chain of variables pointing
+        // to each other.
+        const IR::Type* type = bindings->lookup(current);
+        if (type == nullptr)
+            break;
+        replacement = type;
+        if (!type->is<IR::ITypeVar>())
+            break;
+        current = type->to<IR::ITypeVar>();
+    }
```
ChrisDodd: This loop could be written more concisely (and more clear?) as

    auto replacement = typeVariable->getNode();
    while(current) {
        if (auto type = bindings->lookup(current)) {
            replacement = type;
            current = type->to<IR::TypeVar>();
        } else {
            break; } }

jnfoster:
Do we know the overhead on a big program like `switch.p4` if we eliminate simpleCompose and just use compose? If it's negligible, then I'd be in favor of making the change you proposed and reverting to a standard non-fixpoint version of applying a type substitution. If it's not negligible, then this design seems okay.

mbudiu-vmw:
I tried to simplify the code, but I failed, so I will probably postpone investigating this until I have more time available.

Issue #774, opened 2017-07-13
Extract to the underscore identifier (Spec section 11.8.4) doesn't work


pofig37:
Compiling following test code fails:
```
#include <core.p4>

header Header {
    bit<32> data;
}

parser p0(packet_in p, out Header h) {
    state start {
        p.extract<Header>(_);
        transition next;
    }

    state next {
        p.extract(h);
        transition accept;
    }
}

parser proto(packet_in p, out Header h);
package top(proto p);
top(p0()) main;
```
fails with the  following error:
```
parser-extract-bug.p4(25):syntax error, unexpected _, expecting )
        p.extract<Header>(_
                          ^
error: 1 errors encountered, aborting compilation
```

mbudiu-vmw:
This one is actually quite tricky. The problem is in the grammar; currently it does not accept _ as an expression. An attempt to add _ as a legal expression makes the grammar ambiguous: the compiler cannot tell whether _ is a type or an expression. @ChrisDodd : do you see a workaround for this?

In the worst case we may need to go back to the spec and disallow the use of the underscore identifier for expressions...

jafingerhut:
At least for the case of an extract(_) method call, I believe the spec already mentions that another way to achieve the same effect is to use advance().  Note that regardless of which of these 2 methods is allowed or used, you could not later emit that header, so it is effectively removed from the incoming packet by doing these things in a parser.

In terms of what is possible semantically, there is no loss of functionality if _ is disallowed as an expression, is there?  It seems the same as declaring a temporary variable that can be put in place of the underscore, then never using that temporary variable's value anywhere.

mbudiu-vmw:
Yes, we can still write the same set of programs.

ChrisDodd:
Allowing `_` as an expression anywhere is hard to parse, but I'm not sure it makes sense.  Allowing it as an argument is simpler and doesn't introduce any conflicts.

    argument
        : expression  { $$ = $1; }
        | "_"         { $$ = new IR::DefaultExpression(@1); }
        ;

mbudiu-vmw:
Thinking about this it only makes sense to allow it for out parameters. So we may need to amend the spec anyway.

mbudiu-vmw:
This should work now; there is a pending fix to the spec too.

PR #876, opened 2017-08-31
Fix for issue #774

mbudiu-vmw:


mbudiu-vmw:
This fix will also require some fixes in the spec.

frontends/parsers/p4/p4parser.ypp
```diff
@@ -486,7 +486,7 @@ simpleKeysetExpression
     | expression "&&&" expression { $$ = new IR::Mask(@1 + @3, $1, $3); }
     | expression ".." expression  { $$ = new IR::Range(@1 + @3, $1, $3); }
     | DEFAULT                     { $$ = new IR::DefaultExpression(@1); }
-    | "_"                         { $$ = new IR::DefaultExpression(@1); }
+    | DONTCARE                    { $$ = new IR::DefaultExpression(@1); }
```
ChrisDodd: This is an unnecessary obfuscation -- its generally better to use the literals for textual tokens (punctuation and the like) rather than names, as this makes it look like DONTCARE is a keyword `dontcare` rather than a single underscore.

mbudiu-vmw: In this case we should get rid of the DONTCARE terminal altogether.

ChrisDodd: The problem is that a keyword is needed for the lexer to generate the token as flex (or more precisely the C++ interface between flex and bison) does not support referring to tokens by string literal.  So for all the punctuation-type tokens, we need to define an alias identifier that is used by flex to create the token object.

mbudiu-vmw:
@ChrisDodd : you may want to re-review this; I have added support for back-ends via a mid-end pass.

PR #882, opened 2017-09-04
Add test of v1model update_checksum implementation

jafingerhut:
It does not test the verify_checksum implementation.

### DIFF
--- a/backends/bmv2/CMakeLists.txt
+++ b/backends/bmv2/CMakeLists.txt
@@ -105,6 +105,10 @@ set (BMV2_TEST_SUITES
 # so we allow it to fail on BMv2.
 set (XFAIL_TESTS
   testdata/p4_14_samples/issue60.p4
+  # Program checksum1-bmv2.p4 uses new verify_checksum and
+  # update_checksum extern functions that were added but then reverted
+  # until they can be deployed without breaking existing programs.
+  testdata/p4_16_samples/checksum1-bmv2.p4
   )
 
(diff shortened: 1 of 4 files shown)


jafingerhut:
The new P4 program only compiles with the new v1model checksum APIs added via this commit: https://github.com/p4lang/p4c/commit/f09a95dcda68ca18a1528be6a63a16f2b8074d46

but later reverted on 2017-Aug-28.

mbudiu-vmw:
There should already be a few tests with stf files that exercise the checksums in bmv2.
But more tests are always better.

jafingerhut:
@mbudiu-vmw I found only one P4_16 test program in the p4c/testdata/p4_16_samples program where there is checksum calculation code, _and_ there is an .stf file to specify input packets, and expected output packets that check for the correct checksum being computed.  That is issue655-bmv2.p4 and issue655-bmv2.stf.  It only calculates checksums for a single 16-bit field.

The test program and .stf file proposed in this PR calculate the IPv4 header checksum over an entire 20-byte IPv4 header, plus a test packet that has 24-bit IPv4 header with options, parsed into a varbit field with a non-0 length.  Thus this test should be adding a little bit of functional coverage not already covered by other tests.

jafingerhut:
@cc10512 I have tried making the changes you suggested -- thanks for file name and line number pointers to guide me.  We'll see what the Travis build thinks of these changes in a bit.

PR #885, opened 2017-09-08
Try to convince Travis to actually cache the ccache db

sethfowler:
Travis seems determined to not cache anything. Let's try specifying the cache directory manually.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,7 +8,9 @@ branches:
 services:
   - docker
 
-cache: ccache
+cache:
+  directories:
+  - $HOME/.ccache
 
 env:
   - CTEST_PARALLEL_LEVEL=4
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -19,3 +21,4 @@ install:
 
 script:
   - docker run -w /p4c/build -e CTEST_PARALLEL_LEVEL p4c ctest --output-on-failure --schedule-random
+  - ls $HOME/.ccache
```
cc10512: Not sure I understand what is the lifetime and/or management of these directories between docker and Travis. Where does $HOME/.ccache from the travis image get mapped to the docker container?

Isn't it worth running `ccache --show-stats` as well to see what did docker cache?

sethfowler: This stuff certainly is a little bit confusing. `start_ccache` maps the `$HOME/.ccache` directory in the Travis VM into the `ccache` Docker container.

`ccache --show-stats` (or the moral equivalent) is actually already run inside the build container. It runs at the end of the build process; look for it in the log between the build and the tests.

PR #892, opened 2017-09-09
fixes hiding of overloaded constructors with variable number of argum…

cc10512:
…ents

CompilerBug and CompilerUnimplemented have a pair of constructors that
start with a const char *. The (const char *format, ...) constructor
hides the one with (const char *filename, int line, const char
*format, ...).

### DIFF
--- a/lib/exceptions.h
+++ b/lib/exceptions.h
@@ -55,7 +55,7 @@ class CompilerBug final : public P4CExceptionBase {
             : P4CExceptionBase(format, args...)
     { message = cstring(ANSI_RED) + "Compiler Bug" + ANSI_CLR + ":\n" + message; }
     template <typename... T>
-    CompilerBug(const char* file, int line, const char* format, T... args)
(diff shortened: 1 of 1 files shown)


ChrisDodd:
If this is necessary, I guess I don't understand C++ overload resolution rules -- I thought the `(const char *, int, const char *, ...)` overload would take preference over the `(const char *, ...)` as exact matches are better than matches with `...`.  The `(const char *, ...)` overload should only be used if the 2nd argument is not an int, or the 3rd is not a `char *`.  It was working properly for the `BUG` macro without this, at least I though so (I've seen many BUG failures and they all seem to have the right error message).

Or is the problem the reverse hiding?  Something like `CompilerBug("message with %d and %s", intvar, stringvar);` would call the one expecting file and line, and treat the stringvar as the format?

cc10512:
It was treating (filename, lineno, format, arg) as (format, ...) and throwing an exception in boost::format that it had too many arguments.

PR #898, opened 2017-09-12
Use the gold linker if available

sethfowler:
On a build where ccache gets a lot of hits, the linker rapidly becomes the bottleneck. Let's use the gold linker when it's available; it's much faster than the standard ld. It comes with binutils, so all we really need to do is flip it on, at least on recent Linux distros.

Unfortunately the compatibility situation between gold and clang (and symmetrically, between lld and gcc) doesn't seem to be perfect, so I hesitate to enable it when building with clang. We can add support for lld in a future PR.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -16,7 +16,7 @@ cmake_minimum_required (VERSION 3.0.2 FATAL_ERROR)
 
 find_program(CCACHE_PROGRAM ccache)
 if(CCACHE_PROGRAM)
-    MESSAGE("-- Enabling ccache")
+    MESSAGE(STATUS "Enabling ccache")
     set_property(GLOBAL PROPERTY RULE_LAUNCH_COMPILE "${CCACHE_PROGRAM}")
 endif()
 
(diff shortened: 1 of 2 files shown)


sethfowler:
@cc10512, I took a look at removing that `VERBOSE` argument to `BISON_TARGET`, and it works, but I hesitate to actually write the patch without understanding why it's there in the first place. It's not obvious to me why it was necessary.

ChrisDodd:
The VERBOSE argument to BISON_TARGET is to get it to produce a .output file in the build directory with the bison verbose output (dump of the parser states), which is crucial for debugging parser problems, but not actually needed to produce an executable if the parser is correct.

sethfowler:
Yeah, that part I get, but what's less clear to me is the relationship between that argument and the CMake version issues Calin and I were discussing on Friday. There's a secondary interaction happening as well, apparently, and I need to clarify with Calin precisely what that is.

cc10512:
Depending on the version of cmake, the VERBOSE target needs or not a file argument. 3.0.2 (the one we mandate) needs the file as an argument. The later ones, don't and generate a file with exactly the same name as the one we have listed, and then issue a warning that they are overwriting the file. It's annoying and silly. What I mentioned is, keep the argument for the older versions and use just VERBOSE for the newer ones. Or alternatively, rename the file we produce the output into to something else.

CMakeLists.txt
```diff
@@ -131,6 +131,21 @@ add_cxx_compiler_option ("-Wextra")
...
 
+# If we're on GCC, use the gold linker if available.
+if(CMAKE_COMPILER_IS_GNUCC)
+    execute_process(
+            COMMAND ${CMAKE_C_COMPILER} -fuse-ld=gold -Wl,--version
+            ERROR_QUIET OUTPUT_VARIABLE LD_VERSION)
+    if ("${LD_VERSION}" MATCHES "GNU gold")
+        set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -fuse-ld=gold")
+        set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fuse-ld=gold")
+        message("-- Using the GNU gold linker.")
```
cc10512: message (STATUS "Using the GNU gold linker.") will produce the "--" prefixed string, properly indented.

CMakeLists.txt
```diff
@@ -131,6 +131,21 @@ add_cxx_compiler_option ("-Wextra")
...
+    else ()
+        message("-- Using the default system linker.")
```
cc10512: same here.

sethfowler:
Ah, I see. Looks like 3.7.2 and up can use VERBOSE with no argument. I can add that in.

CMakeLists.txt
sethfowler: I'll make that change.

PR #900, opened 2017-09-12
allow CopyStructures pass to ignore methods

rhalstea:
In our back-end users have some leeway with writing their own functions/externs that get linked in later. Since they can choose to write a function that returns a struct I'd prefer not to reject the design outright. So I just put a conditional around the error call in this case.

### DIFF
--- a/midend/copyStructures.h
+++ b/midend/copyStructures.h
@@ -54,10 +54,14 @@ namespace P4 {
  */
 class DoCopyStructures : public Transform {
     TypeMap* typeMap;
-    bool ignoreMethods;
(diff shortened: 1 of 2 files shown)


midend/copyStructures.h
```diff
@@ -54,19 +54,22 @@ namespace P4 {
  */
 class DoCopyStructures : public Transform {
     TypeMap* typeMap;
+    bool ignoreMethods;
```
mbudiu-vmw: The name of this flag is not entirely suitable. Please add a comment describing what it does.
It skips methods whose output type is a struct; when this flag is false encountering such methods will produce an error message.

PR #902, opened 2017-09-13
Build with optimizations enabled on Travis

sethfowler:
This PR is just an experiment for now.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -40,7 +40,7 @@ RUN apt-get update && \
     pip install tenjin && \
     mkdir build && \
     cd build && \
-    cmake .. && \
+    cmake .. '-DCMAKE_CXX_FLAGS:STRING=-O3 -march=native' && \
     make && \
     make install && \
     /usr/local/bin/ccache -p -s && \


sethfowler:
I'm rerunning the test. With `-O3 -flto`, the time the tests take to run is halved, but the compile time increases so much that it's a wash. (Actually, it's a slight loss.) I'm hoping that some other combination of flags will provide comparable increases in test performance without increasing the compile time to that degree.

sethfowler:
The compile time slowdown is almost certainly mostly due to LTO, by the way, since the measurements I'm taking here are on a second run, after ccache has had a chance to kick in. ccache caches compilation output but not linker output, so we can be pretty confident that the additional link time accounts for the difference.

sethfowler:
OK, so here are the results. `-O3` seems to give us more or less the same performance as `-O3 -flto`, and since LTO slows down linking considerably, we can leave LTO out of the analysis. Comparing `-O0` and `-O3`:

```
-O0, no cache: 31 min
-O3, no cache: 27 min

-O0, cache: 14 min
-O3, cache: 9 min
```

Here "no cache" means a 100% ccache miss rate, and "cache" means a 100% ccache hit rate. In the latter case we are just measuring link time, the tests, and various overheads.

It's pretty clear that in either scenario, `-O3` wins, and the win is especially dramatic when ccache does well and we're mostly just running the tests.

Based on these results, I think we should switch to running our tests on Travis with `-O3`. I'll upload a cleaner patch that makes that change.

sethfowler:
@cc10512 This is no longer a WIP. Since this seems to be a win in all situations, let's go for it.

PR #904, opened 2017-09-14
LocalCopyprop improvements -- track struct fields and stack elements

ChrisDodd:


mbudiu-vmw:
I am traveling for the rest of this week, i will try to review this asap, but there may be a delay.

midend/expr_uses.h
```diff
@@ -21,19 +21,52 @@ limitations under the License.
 
 /* Should this be a method on IR::Expression? */
 
+/// Functor to check if an expression uses an lvalue.  The lvalue is specified as a
+/// a cstring, which can be the name of a variable with optional field names and constant
+/// array indexes for fields of headers or structs or unions or elements of stacks.
```
sethfowler: Am I wrong in thinking that we should be able to use the ReferenceMap to determine this stuff? (Perhaps with some additional munging for array indices.)

test/gtest/expr_uses_test.cpp
```diff
@@ -0,0 +1,43 @@
...
+limitations under the License.
+*/
+
+#include "gtest/gtest.h"
+#include "ir/ir.h"
+#include "ir/visitor.h"
+#include "lib/exceptions.h"
+#include "midend/expr_uses.h"
+
+TEST(expr_uses, expr_uses) {
```
sethfowler: Thanks for adding a GTest! 👍 👍 👍

midend/expr_uses.h
ChrisDodd: In theory, yes, but it would be more complicated, and it would depend on the refMap being up to date.

That latter thing is the big problem with the refMap (and typeMap) in general -- it means you can't use them reliably in Transform passes (only in Inspectors) because as soon as you transform anything, they're invalidated which may corrupt later transformations in the same pass.  Since we use this in LocalCopyProp and LocalCopyProp is a (complex) transform, it hits this problem in a big way.

This is also why LocalCopyProp uses strings to name lvalues rather than `IR::Expression *`s -- the latter can change when things change (and there may be multiple copies of them) so using them as map keys is a non-starter.

test/gtest/expr_uses_test.cpp
ChrisDodd: My first gtest attempt -- I'm still feeling my way,

testdata/p4_14_samples_outputs/switch_20160226/switch-midend.p4
```diff
@@ -2818,16 +2818,16 @@ control egress(inout headers hdr, inout metadata meta, inout standard_metadata_t
...
     @name(".egress_mirror_drop") action _egress_mirror_drop(bit<32> session_id) {
         meta.i2e_metadata.mirror_session_id = (bit<16>)session_id;
-        clone3<tuple_0>(CloneType.E2E, session_id, { meta.i2e_metadata.ingress_tstamp, meta.i2e_metadata.mirror_session_id });
+        clone3<tuple_0>(CloneType.E2E, session_id, { meta.i2e_metadata.ingress_tstamp, (bit<16>)session_id });
         mark_to_drop();
     }
     @name(".egress_redirect_to_cpu") action _egress_redirect_to_cpu(bit<16> reason_code) {
         meta.fabric_metadata.reason_code = reason_code;
-        clone3<tuple_1>(CloneType.E2E, 32w250, { meta.ingress_metadata.bd, meta.ingress_metadata.ifindex, meta.fabric_metadata.reason_code, meta.ingress_metadata.ingress_port });
+        clone3<tuple_1>(CloneType.E2E, 32w250, { meta.ingress_metadata.bd, meta.ingress_metadata.ifindex, reason_code, meta.ingress_metadata.ingress_port });
```
mbudiu-vmw: In all these cases you are converting a field member access to some arithmetic, such as a cast, which could be more expensive.

testdata/p4_16_samples_outputs/inline-stack-bmv2-midend.p4
```diff
@@ -30,12 +30,9 @@ control ComputeChecksumI(inout H hdr, inout M meta) {
 }
 
 control IngressI(inout H hdr, inout M meta, inout std_meta_t std_meta) {
-    h_t[1] hdr_1_stack;
     @hidden action act() {
-        hdr_1_stack[0] = hdr.stack[0];
-        hdr.stack[0] = hdr_1_stack[0];
-        hdr_1_stack[0] = hdr.stack[0];
-        hdr.stack[0] = hdr_1_stack[0];
+        hdr.stack[0] = hdr.stack[0];
```
mbudiu-vmw: These are no-ops. Do we have anything to detect this kind of code?

testdata/p4_16_samples_outputs/issue561-midend.p4
```diff
@@ -21,12 +21,12 @@ control c(out bit<32> x) {
         x = u.h1.f + u.h2.g;
         u.h1.setValid();
         u.h1.f = 32w0;
-        x = x + u.h1.f;
+        x = x + 32w0;
```
mbudiu-vmw: This is a sign we could do more strength reductions after this.

midend/expr_uses.h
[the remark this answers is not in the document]
ChrisDodd: Yes, but if they are not constants, they can't be copy-propagated.

testdata/p4_16_samples_outputs/inline-stack-bmv2-midend.p4
ChrisDodd: Yes, `DoLocalCopyPropagation::postorder(IR::AssignmentStatement *)` is supposed to eliminate these -- I'll have to look into why it is missing this one.

ChrisDodd:
On Mon, Sep 18, 2017 at 10:03 AM, Mihai Budiu <notifications@github.com>
wrote:

> *@mbudiu-vmw* commented on this pull request.
>
> This is a very different way to do things compared with all our current
> visitors. I wonder whether this is a sign that something is wrong with the
> Transform infrastructure.
> ------------------------------
>
This is the way that the Transform infrastructure is designed to work.
Many other visitors are needlessly complex because they aren't really using
it the right way (IMO).

mbudiu-vmw:
Do you think we should be using strings in more places?

testdata/p4_14_samples_outputs/issue894-midend.p4
```diff
@@ -150,13 +150,13 @@ control ingress(inout headers hdr, inout metadata meta, inout standard_metadata_
         hash<bit<16>, bit<16>, tuple_0, bit<32>>(meta.custom_metadata.hash_val1, HashAlgorithm.csum16, 16w0, { hdr.ipv4.srcAddr, hdr.ipv4.dstAddr, hdr.ipv4.protocol, hdr.tcp.srcPort, hdr.tcp.dstPort }, 32w16);
         heavy_hitter_counter1.read(tmp_6, (bit<32>)meta.custom_metadata.hash_val1);
         meta.custom_metadata.count_val1 = tmp_6;
-        meta.custom_metadata.count_val1 = meta.custom_metadata.count_val1 + 16w1;
-        heavy_hitter_counter1.write((bit<32>)meta.custom_metadata.hash_val1, meta.custom_metadata.count_val1);
+        meta.custom_metadata.count_val1 = tmp_6 + 16w1;
+        heavy_hitter_counter1.write((bit<32>)meta.custom_metadata.hash_val1, tmp_6 + 16w1);
```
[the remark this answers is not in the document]
ChrisDodd: Good question -- it depends whether adders or wires and dependencies are more expensive.  It could be converted back to a copy by CSE; any given target will need to balance which way is better.

testdata/p4_14_samples_outputs/switch_20160226/switch-midend.p4
ChrisDodd: Its also removing a data dependency within the action, which is usually a win.  But again, specific targets will need to decide on which is best and will generally need both CSE and copyprop in the backend.

testdata/p4_16_samples_outputs/issue561-midend.p4
ChrisDodd: Yes, which would then uncover more copy propagation opportunities.

ChrisDodd:
Yes, probably we should.  It makes the code easier to debug and understand when it is running (don't need to worry about when two complex data structures are equivalent).  It does depend on having unique names for things, but we do have a pass that ensures that early on.

PR #911, opened 2017-09-15
Make all backends optional

sethfowler:
People who are developing extensions won't necessarily want to build all of the backends - for example, they may want to avoid building anything unnecessary on their continuous integration infrastructure. (That's my motivation in writing this PR.)

What this PR does is:
- Add flags to enable or disable the backends. They all default to ON.
- Use the existing infrastructure to avoid including BMV2-specific GTests if the BMV2 backend is not enabled.

With those changes in place, things "just work". Disabling all the backends will give you a perfectly working test suite consisting of two tests, cpplint and gtest. =)

As a bonus, I've also including a commit in this PR to make `libboost-graph` available on Travis so we will build the `p4c-graphs` backend there.


CMakeLists.txt
```diff
@@ -30,6 +30,10 @@ set (CMAKE_USE_RELATIVE_PATHS 1)
 
 OPTION (ENABLE_DOCS "Build the documentation" OFF)
 OPTION (ENABLE_P4RUNTIME_TO_PD "Build the P4Runtime PD Generator" ON)
+OPTION (ENABLE_BMV2 "Build the BMV2 backend (required for the full test suite)" ON)
+OPTION (ENABLE_EBPF "Build the EBPF backend (required for the full test suite)" ON)
+OPTION (ENABLE_EBPF "Build the P4Test backend (required for the full test suite)" ON)
```
ChrisDodd: typo here? EBPF vs P4Test

ChrisDodd:
Should there also be something in `foreach (ext` loop around line 164 of CMakeLists.txt to create an `ENABLE_` flag for each subdirectory in extensions?  I guess it requires string manipulation to turn the directory name into a valid identifier.

CMakeLists.txt
sethfowler: Doh! You're right. I'll fix that.

sethfowler:
That does sound like a good improvement. Let me see if it's a simple thing to do.

sethfowler:
I honestly expected that to be harder than it was. OK, we now automatically generate an option to disable each extension we find.

CMakeLists.txt
```diff
@@ -193,6 +197,13 @@ file (GLOB p4c_extensions RELATIVE ${P4C_SOURCE_DIR}/extensions ${P4C_SOURCE_DIR
...
 foreach (ext ${p4c_extensions})
   if (EXISTS ${P4C_SOURCE_DIR}/extensions/${ext}/CMakeLists.txt)
+    # Generate an option that makes it possible to disable this extension.
+    string(MAKE_C_IDENTIFIER ${ext} EXT_AS_IDENTIFIER)
+    string(TOUPPER ${EXT_AS_IDENTIFIER} EXT_AS_OPTION_NAME)
+    string(CONCAT ENABLE_EXT_OPTION "ENABLE_" ${EXT_AS_OPTION_NAME})
+    string(CONCAT EXT_HELP_TEXT "Build the " ${ext} " backend")
+    OPTION (${ENABLE_EXT_OPTION} ${EXT_HELP_TEXT} ON)
+
     add_subdirectory (extensions/${ext})
```
ChrisDodd: should this be wrapped in `if (${ENABLE_EXT_OPTION})`?

CMakeLists.txt
```diff
@@ -30,6 +30,10 @@ set (CMAKE_USE_RELATIVE_PATHS 1)
 
 OPTION (ENABLE_DOCS "Build the documentation" OFF)
 OPTION (ENABLE_P4RUNTIME_TO_PD "Build the P4Runtime PD Generator" ON)
+OPTION (ENABLE_BMV2 "Build the BMV2 backend (required for the full test suite)" ON)
+OPTION (ENABLE_EBPF "Build the EBPF backend (required for the full test suite)" ON)
+OPTION (ENABLE_P4TEST "Build the P4Test backend (required for the full test suite)" ON)
+OPTION (ENABLE_P4C_GRAPHS "Build the p4c-graphs backend" ON)
```
cc10512: shouldn't you remove these if you've enabled them below now?

CMakeLists.txt
```diff
@@ -202,19 +213,25 @@ add_subdirectory (midend)
 add_subdirectory (control-plane)
 
 
+add_subdirectory (test)
+
+if (ENABLE_BMV2)
```
cc10512: and these are now also redundant.

CMakeLists.txt
sethfowler: Argh, yes, it should.

CMakeLists.txt
sethfowler: I don't understand... the foreach loop is for extensions. These are built-in backends. Am I missing something?

sethfowler:
OK, the version I just pushed adds the missing `if()` Chris pointed out, so disabling extensions actually works. I checked that this actually works locally.

I've also improved the formatting a little bit in `CMakeLists.txt`.

CMakeLists.txt
cc10512: You're right. I read is as all backends being captured in that loop. Sorry about that.

sethfowler: No worries. Thanks for reviewing so closely. =)

ChrisDodd: Though I guess you /could/ have a loop over the backends directory to create all of these.  I guess the help messages would then be less informational (unless there was a way of pulling the message from a file in the directory in question...)

PR #912, opened 2017-09-16
Add ability to serialize static table entries

antoninbas:
In the case where a P4_16 program provides compile-time table entries,
it is convenient for targets in general to have access to P4Runtime
messages which encode these static entries. This adds support for
serializing all static entries in a program as a P4Runtime.WriteRequest
message. This message can then be sent to a P4 target along with the
rest of the compiler output, and the message can be processed just like
runtime control-plane messages.

When using the bmv2 backend, the '--p4runtime-entries-file' command-line
option can be used to generate the message and write it to the provided
file. The format is determined by '--p4runtime-format'.

This commit was a good opportunity to update the submodule ref pointer
to p4lang/PI, which provides p4info.proto and p4runtime.proto.


antoninbas:
@teverman could you take a look at this?

teverman:
Code looks good.  I want to take a look at the output.  Please give me a little more time to patch this pull request and compile some of our.p4 programs.

antoninbas:
@cc10512 not a problem I can wait for #913 to be merged

backends/bmv2/bmv2.cpp
```diff
@@ -104,9 +104,13 @@ int main(int argc, char *const argv[]) {
     // Generate a PI control plane API for this program if requested.
     if (!options.p4RuntimeFile.isNullOrEmpty()) {
         std::ostream* out = openFile(options.p4RuntimeFile, false);
+        std::ostream *outEntries = nullptr;
+        if (!options.p4RuntimeEntriesFile.isNullOrEmpty())
+            outEntries = openFile(options.p4RuntimeEntriesFile, false);
```
teverman: Please add a warning if a runtime entries file is specified on the command line without the --p4runtime-file.

antoninbas: Will do

frontends/common/options.cpp
```diff
@@ -96,6 +96,10 @@ CompilerOptions::CompilerOptions() : Util::Options(defaultMessage) {
     registerOption("--p4runtime-file", "file",
                    [this](const char* arg) { p4RuntimeFile = arg; return true; },
                    "Write a P4Runtime control plane API description to the specified file.");
+    registerOption("--p4runtime-entries-file", "file",
```
mbudiu-vmw: I think that this should be part of the BMV2Options class instead.
Other front-ends to do not have this capability.

antoninbas: I disagree, this should be common to all backends
In terms of applicability, this is no different from `--p4runtime-file`

sethfowler: Agreed with Antonin. Keep in mind that while we're not there yet, we want to get to the point where the P4Runtime serializer isn't backend-specific at all; it should only require the frontend to be run.

mbudiu-vmw: You are assuming that all back-ends need the same control-plane, this will not be true.
A switch like OVS will have a separate control-plane.
If you think it is common to all back-ends you should explain how the ebpf back-end can benefit.

sethfowler:
@antoninbas I suppose the protobuf API makes it awkward to serialize multiple "unrelated" messages to one file, but perhaps we could definitely a wrapper message that includes both P4Info and an option stream of P4Runtime initialization commands? (Perhaps just a `repeated WriteRequest` for now, but we could imagine support other P4Runtime messages.) That way we could just write a single file that supports both.

frontends/common/options.cpp
sethfowler: That's, as I see it, a separate question. A P4Info message (which is what the P4Runtime serializer code, as it currently stands, really generates) just describes metadata about the program. That information is intended to be used by the control plane, but even if people using a certain backend or target will end up using a different control plane scheme, we can still *generate* a P4Info message. It's still well-defined.

sethfowler: In other words, nobody has to use it, and some people won't, but there's no real advantage to not providing it for certain backends and making things uniform adds convenience and makes the ecosystem simpler.

antoninbas:
@sethfowler The protobuf API indeed doesn't support multiple messages written to the same file very well.
I'm not sure I see the interest of defining another message at this point. I think the idea at that stage is that backends will include it in their device-specific config if needed (https://github.com/p4lang/PI/blob/master/proto/p4/p4runtime.proto#L404). If it turns out that it is a good idea to have this be promoted to a field in `ForwardingPipelineConfig`, we will probably do that. It could also be a new message that wraps both `P4Info` and a `WriteRequest`, but since I see no advantage in doing so right now, I'd rather avoid breaking existing code.

control-plane/p4RuntimeSerializer.h
```diff
@@ -35,12 +35,13 @@ enum class P4RuntimeFormat {
...
 
-void serializeP4Runtime(std::ostream* destination,
+void serializeP4Runtime(std::ostream* p4InfoDestination,
                         const IR::P4Program* program,
                         const IR::ToplevelBlock* evaluatedProgram,
                         ReferenceMap* refMap,
                         TypeMap* typeMap,
-                        P4RuntimeFormat format = P4RuntimeFormat::BINARY);
+                        P4RuntimeFormat format = P4RuntimeFormat::BINARY,
+                        std::ostream* p4RuntimeEntriesDestination = nullptr);
```
sethfowler: Once #913 gets merged, you will be able to implement this much more cleanly as a separate method on P4RuntimeAPI.

antoninbas: sounds good

sethfowler:
@antoninbas it's not too big a deal when there are only two files to be generated. If we think that in the long term we'll need even more of these things, that might be an argument to go ahead and add a wrapper message now so we don't end up needing to specify a ton of files on the command line. You have a much better idea of how the P4Runtime ecosystem will evolve than I do, so I'll defer to your expertise here; just a suggestion.

antoninbas:
@sethfowler I totally agree with you. There shouldn't be more than 2 files (one for P4info, one for static initialization of P4 objects). And if it does seem like the right thing to do as we start getting some mileage out of this new output, we can definitely bundle them into a single message that will replace P4Info in ForwardingPipelineConfig.

teverman:
I'm all for having p4c generate the individual proto messages and letting the backend have the ability to wrap the messages and package them into a file structure that's best suited to its target.  We are currently doing this now by having the runtime serializer write to stringstreams, then consolidating the streams according to our needs.

sethfowler:
@teverman you may find the API more convenient for your use-case after #913 lands. You'll then have direct access to the P4Info object (and probably, once Antonin rebases, the WriteRequest messages being added in this PR as well) and you'll be able to write your own custom serialization routines as needed.

antoninbas:
@sethfowler @cc10512 I rebased master and added a unit test for my changes, please take another look
@teverman I fixed the bug you reported to me

test/gtest/p4runtime.cpp
```diff
@@ -54,6 +56,9 @@ createP4RuntimeTestCase(const std::string& source,
     PassManager p4RuntimeFixups = {
         new BMV2::SynthesizeValidField(&refMap, &typeMap),
         new P4::EliminateTuples(&refMap, &typeMap),
+        // If DoLocalCopyPropagation::RewriteTableKeys is not run, we apparently
+        // cannot find the bitwidth of members of the match key
+        new P4::LocalCopyPropagation(&refMap, &typeMap),
```
cc10512: This sounds like we're breaking the abstraction that one pass does one job. The values in entries are constants from the beginning and they shouldn't need to be "propagated". @ChrisDodd can you please comment?

mbudiu-vmw:
@antoninbas : I still have problems building: 
```
/home/mbudiu/git/p4c/build/control-plane/p4/p4runtime.pb.h:32:34: fatal error: google/rpc/status.pb.h: No such file or directory
```

The problem is here:
```
#include "google/rpc/status.pb.h"
```

If I replace the quotes with `<` and `>` it works, though.

mbudiu-vmw:
That's a generated file, so I take it back.

mbudiu-vmw:
It still does not build for me; I take back the assertion that I can fix the problem.

antoninbas:
@mbudiu-vmw would you mind trying this out: https://github.com/p4lang/p4c/pull/926
In the end this is basically the same issue as https://github.com/p4lang/p4c/issues/924, which is that we run `make check` after `make install` in Travis

Issue #841, opened 2017-08-12
Type checking for headers initialized by tuples/lists broken.


jnfoster:
The P4_16 spec states:

> A header object can be initialized with a list expression, similar to a struct—the list fields are assigned to the header fields in the order they appear. In this case the header automatically becomes valid:
```
header H { bit<32> x; bit<32> y; }
H h;
h = { 10, 12 };  // This also makes the header h valid
```
However, the p4c compiler rejects the following program:
```
#include <core.p4>
#include <v1model.p4>

header h_t {
  bit<32> src;
  bit<32> dst;
  bit<16> csum;
}

struct metadata {
}
struct headers {
  h_t h;
}

parser MyParser(packet_in packet, out headers hdr, inout metadata meta, inout standard_metadata_t standard_metadata) {
    state start {
        packet.extract(hdr.h);
        transition accept;
    }
}
control MyVerifyChecksum(in headers hdr, inout metadata meta) {
    apply {
    }
}
control MyIngress(inout headers hdr, inout metadata meta, inout standard_metadata_t standard_metadata) {
    apply {
    }
}
control MyEgress(inout headers hdr, inout metadata meta, inout standard_metadata_t standard_metadata) {
    apply {
    }
}
control MyDeparser(packet_out packet, in headers hdr) {
    apply {
    }
}
control MyComputeChecksum(inout headers hdr, inout metadata meta) {
    Checksum16() checksum;
    h_t h = { hdr.h.src, hdr.h.dst, 16w0 };
    apply {
	hdr.h.csum = checksum.get(h);
   }    
}
V1Switch(MyParser(), MyVerifyChecksum(), MyIngress(), MyEgress(), MyComputeChecksum(), MyDeparser()) main;
```
with this error:
```
foo.p4(40): error: AssignmentStatement: Number of fields 3 in initializer different than number of fields in structure 4: Tuple(3) to header h_t
    h_t h = { hdr.h.src, hdr.h.dst, 0 };
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
foo.p4(40)
    h_t h = { hdr.h.src, hdr.h.dst, 0 };
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^
foo.p4(4)
header h_t {
       ^^^
```
(The compiler also rejects the example given in the spec with a similar error message.)

sethfowler:
[...]

PR #913, opened 2017-09-16
Fix P4Runtime digest and match field issues

sethfowler:
We've been running into problems with generating P4Runtime APIs for some of our test cases; the current code doesn't handle some `digest<T>()` and match field constructs that we need to support. This PR reworks things to resolve those issues and to finally let us write unit tests for P4Runtime.

For the `digest<T>()` issues, the basic problem is that we were too restrictive about what `T` could be. I've updated the code to be more flexible about supporting any `Type_StructLike`.

For match fields, the changes are much more involved. Rather than continue to work around issues in P4Runtime, this patch switches us to a model where P4Runtime *only* accepts explicit `@name` annotations for match fields. We have a pass, TableKeyNames, that adds those already, but it needed some work to be general enough to replace the existing P4Runtime code. That required many small changes:

- We generate global name annotations (i.e. `@name(".foo")`) for the top-level header and metadata structs introduced in P4-14->P4-16 conversion. This has the effect of hiding them from the control plane API.
- We no longer treat bitwise AND masks in keys specially during P4-14->P4-16 conversion. Instead, we allow TableKeyNames to handle them.
- TableKeyNames now permits bitwise AND in keys, and generates a name for them automatically. We needed this for P4-14, but as a side effect they can be used in P4-16 as well, which seems perfectly reasonable to me; we already support slices, and it doesn't seem like a big jump to support bitwise AND.
- TableKeyNames now generates annotations like `@name("foo.$valid$")` for `foo.isValid()`. This aligns with what P4Runtime needs.
- TableKeyNames now handles global names in Member expressions by dropping the left side of the expression. That hides the top-level structs introduced during P4-14->P4-16 conversion from the control plane API, which is what we want.

Unit tests for P4Runtime have been long overdue. To add support for them, I split P4Runtime generation into two phases: `generateP4Runtime()` returns a `P4RuntimeAPI` struct containing the P4Runtime data structures (just `P4Info`, really), and `P4RuntimeAPI` has a method allowing you to serialize its contents to an output stream. This allows us to write tests against the P4Runtime data structures without needing to serialize them.

The test suite I've added isn't complete by any means, but there's a basic test for ID generation as well as tests for `digest<T>()` and match fields, since those are the features impacted by the changes in this PR. Writing those tests identified some additional issues that need to be fixed in a future PR:

- `$valid$` fields show up in `digest<T>()` field lists, which we probably don't want.
- The P4Runtime generation code depends implicitly on `EliminateTuples` and `LocalizeAllActions`, which we also probably don't want. That's in addition to the known dependency on `SynthesizeValidField`.

There were other bugs that were fixed as part of this PR. Unit tests definitely help.

@antoninbas, currently if `headerUnion.isValid()` is used in a match key I rewrite it to `headerUnion.$valid$`, just like we do for headers and header stack elements. Should we change that? I know it's somewhat inconsistent from the point of view of BMV2 (since in BMV2's `SynthesizeValidField` we don't do that, because we want to generate a special primitive in that case), but not rewriting it feels inconsistent from the point of view of P4Runtime.


sethfowler:
I need to update more test outputs, apparently.

control-plane/p4RuntimeSerializer.cpp
```diff
@@ -76,6 +76,25 @@ static bool isControllerHeader(const IR::Type_Header* type) {
...
 
+/// @return the value of @item's explicit name annotation, if it has one. We use
+/// this rather than e.g. controlPlaneName() when we want to prevent any
+/// fallback.
+static boost::optional<cstring>
+explicitNameAnnotation(const IR::IAnnotated* item) {
+    auto* anno = item->getAnnotation(IR::Annotation::nameAnnotation);
+    if (!anno) return boost::none;
+    if (anno->expr.size() != 1) {
+        ::error("%1% should contain a string", anno);
```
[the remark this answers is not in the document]
sethfowler: Yeah, that could use a little cleanup. I'll make it nicer.

control-plane/p4RuntimeSerializer.cpp
```diff
@@ -792,9 +811,9 @@ class P4RuntimeSymbolTable {
     P4SymbolSuffixSet suffixSet;
 };
 
-/// A serializer which translates the information available in the P4 IR into a
+/// An analyzer which translates the information available in the P4 IR into a
 /// representation of the control plane API which is consumed by P4Runtime.
-class P4RuntimeSerializer {
+class P4RuntimeAnalyzer {
```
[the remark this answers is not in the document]
sethfowler: The fact that we need to deal in both the normal IR representation of the P4 program and with the evaluated version. The latter doesn't really work with the normal visitor idioms we use, unfortunately; that's been the problem since the beginning.

test/gtest/p4runtime.cpp
```diff
@@ -0,0 +1,611 @@
...
+                        inout standard_metadata_t sm) {
+            action noop() { }
+
+            table igTable {
+                actions = { noop; }
+                default_action = noop;
+            }
+
+            @name("igTableWithName")
+            table igTableWithoutName {
```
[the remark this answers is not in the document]
sethfowler: Precisely.

backends/bmv2/CMakeLists.txt
```diff
@@ -118,7 +118,12 @@ else()
 endif()
 
 set (GTEST_BMV2_SOURCES
-  ${CMAKE_SOURCE_DIR}/test/gtest/bmv2_isvalid.cpp
+  ${P4C_SOURCE_DIR}/test/gtest/bmv2_isvalid.cpp
+
+  # XXX(seth): This should not depend on BMV2. The source of the dependency is
+  # the SynthesizeValidField pass; we should either promote it to a midend pass
```
mbudiu-vmw: That pass is responsible for issue #841, so it cannot be moved to the mid-end, since it breaks some important invariants about the IR.

testdata/p4_14_samples_outputs/01-BigMatch-first.p4
```diff
@@ -24,14 +24,14 @@ header vag_t {
 }
 
 struct metadata {
-    @name("ing_metadata") 
+    @name(".ing_metadata")
```
mbudiu-vmw: There is a globalname annotation in the spec for this purpose. Are we planning to use it?

testdata/p4_14_samples_outputs/01-BigMatch-first.p4
```diff
@@ -51,7 +51,7 @@ control egress(inout headers hdr, inout metadata meta, inout standard_metadata_t
             @defaultonly NoAction();
         }
         key = {
-            hdr.ethernet.srcAddr: exact @name("hdr.ethernet.srcAddr") ;
+            hdr.ethernet.srcAddr: exact @name("ethernet.srcAddr") ;
```
mbudiu-vmw: is this what you expect?
what happens to 'hdr'? How does the control-plane make up that?

backends/bmv2/CMakeLists.txt
sethfowler: I know, yeah. It's on my to-do list to deal with that better. I have a plan, but I haven't implemented it yet. (I want to try it on our proprietary backend first, basically.)

testdata/p4_14_samples_outputs/01-BigMatch-first.p4
sethfowler: We could certainly switch over to it, but I'd prefer to do it in a separate PR. This just maintains the same mechanism we've been using up to now, but applied in more cases.

testdata/p4_14_samples_outputs/01-BigMatch-first.p4
sethfowler: Yeah, that change was one of the main things I was trying to accomplish in this PR. `hdr` shouldn't be visible to the control plane, because it's not part of the P4-14 program; it's just an artifact of the conversion from P4-14 to P4-16. That was already true, because the P4Runtime code ensured that, but now that I'm making the P4Runtime code lean more heavily on the frontend code, I need to make the appropriate changes to maintain that behavior. Otherwise, the control plane API we'll generate will be incompatible with the one we generate from the old compiler.

control-plane/p4RuntimeSerializer.cpp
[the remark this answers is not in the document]
sethfowler: Just cleaned this up in my newest push.

PR #918, opened 2017-09-18
refactor resolveReferences to use visitor infracture rather than main…

ChrisDodd:
…tining its ouw redundant stack

- should ultimately allow eliinating the refMap, as any pass can inherit
  from ResolutionContext and do the lookups itself

This is still WIP, but I thought I'd put it out there for comment.

The basic idea is to use the already existing support in the visitor infrastructure to do contextual name lookup in ResolveReferences rather than maintaining an extra stack object in the visitor.  This still needs more cleanup and reorganization.

### DIFF
--- a/frontends/common/resolveReferences/resolveReferences.cpp
+++ b/frontends/common/resolveReferences/resolveReferences.cpp
@@ -16,118 +16,152 @@ limitations under the License.
 
 #include "resolveReferences.h"
 #include <sstream>
+#include "frontends/common/options.h"
 
 namespace P4 {
 
-std::vector<const IR::IDeclaration*>*
-ResolutionContext::resolve(IR::ID name, P4::ResolutionType type, bool forwardOK) const {
-    static std::vector<const IR::IDeclaration*> empty;
+static const std::vector<const IR::IDeclaration*> empty;
 
-    std::vector<const IR::INamespace*> toTry(stack);
-    toTry.insert(toTry.end(), globals.begin(), globals.end());
+ResolutionContext::ResolutionContext() {
+    isv1 = P4CContext::get().options().isv1();
+}
 
-    for (auto it = toTry.rbegin(); it != toTry.rend(); ++it) {
-        const IR::INamespace* current = *it;
-        LOG3("Trying to resolve in " << current->toString());
-
-        if (current->is<IR::IGeneralNamespace>()) {
-            auto gen = current->to<IR::IGeneralNamespace>();
-            Util::Enumerator<const IR::IDeclaration*>* decls = gen->getDeclsByName(name);
-            switch (type) {
-                case P4::ResolutionType::Any:
-                    break;
-                case P4::ResolutionType::Type: {
-                    std::function<bool(const IR::IDeclaration*)> kindFilter =
-                            [](const IR::IDeclaration* d) {
-                        return d->is<IR::Type>();
-                    };
-                    decls = decls->where(kindFilter);
-                    break;
-                }
-                case P4::ResolutionType::TypeVariable: {
-                    std::function<bool(const IR::IDeclaration*)> kindFilter =
-                            [](const IR::IDeclaration* d) {
-                    return d->is<IR::Type_Var>(); };
-                    decls = decls->where(kindFilter);
-                    break;
-                }
-            default:
-                BUG("Unexpected enumeration value %1%", static_cast<int>(type));
-            }
-
-            if (!forwardOK && name.srcInfo.isValid()) {
-                std::function<bool(const IR::IDeclaration*)> locationFilter =
-                        [name](const IR::IDeclaration* d) {
-                    Util::SourceInfo nsi = name.srcInfo;
-                    Util::SourceInfo dsi = d->getNode()->srcInfo;
-                    bool before = dsi <= nsi;
-                    LOG3("\tPosition test:" << dsi << "<=" << nsi << "=" << before);
-                    return before;
-                };
-                decls = decls->where(locationFilter);
-            }
-
-            auto vector = decls->toVector();
-            if (!vector->empty()) {
-                LOG3("Resolved in " << dbp(current->getNode()));
-                return vector;
-            } else {
-                continue;
-            }
-        } else {
-            auto simple = current->to<IR::ISimpleNamespace>();
-            auto decl = simple->getDeclByName(name);
-            if (decl == nullptr)
-                continue;
-            switch (type) {
-                case P4::ResolutionType::Any:
-                    break;
-                case P4::ResolutionType::Type: {
-                    if (!decl->is<IR::Type>())
-                        continue;
-                    break;
-                }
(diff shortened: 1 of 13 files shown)


frontends/common/resolveReferences/resolveReferences.cpp
```diff
@@ -19,112 +19,106 @@ limitations under the License.
...
-                BUG("Unexpected enumeration value %1%", static_cast<int>(type));
-            }
-
-            if (!forwardOK) {
+    std::vector<const IR::IDeclaration*> *rv = &empty;
+
+    // FIXME -- we look up 'globals' solely to find match kinds (that the only thing
+    // in globals), which menas that match kinds are effectively reserved words.  Should
+    // probably move resolving match kinds out of here altogether
+    for (auto it = globals.begin(); rv->empty() && it != globals.end(); ++it)
```
mbudiu-vmw: In the other implementation the globals are tried only at the very end.
This seems to try them at the beginning.

frontends/common/resolveReferences/resolveReferences.cpp
```diff
@@ -19,112 +19,106 @@ limitations under the License.
...
+                    return &empty;
+                break; }
+            case P4::ResolutionType::TypeVariable: {
+                if (!decl->is<IR::Type_Var>())
+                    return &empty;
+                break; }
+        default:
+            BUG("Unexpected enumeration value %1%", static_cast<int>(type)); }
+
+        if (!forwardOK) {
```
mbudiu-vmw: Is this code duplicated?

frontends/common/resolveReferences/resolveReferences.cpp
```diff
@@ -240,186 +216,143 @@ void ResolveReferences::checkShadowing(const IR::INamespace* ns) const {
...
-    removeFromContext(p->getConstructorParameters());
-    removeFromContext(p->getApplyParameters());
-    removeFromContext(p->getTypeParameters());
-}
-
-bool ResolveReferences::preorder(const IR::Function* function) {
+bool ResolveReferences::preorder(const IR::Function *function) {
     refMap->usedName(function->name.name);
-    addToContext(function->type->parameters);
+    checkShadowing(function);
```
mbudiu-vmw: It looks like all these methods now have the same implementation, so perhaps you can factor this in a helper.

frontends/common/resolveReferences/resolveReferences.h
```diff
@@ -32,46 +32,24 @@ enum class ResolutionType {
 };
 
 /// Data structure representing a stack of nested namespaces.
-class ResolutionContext : public IHasDbPrint {
-    /// Stack of nested namespaces
-    std::vector<const IR::INamespace*> stack;
-    /// Root namespace for the program.
-    const IR::INamespace* rootNamespace;
+class ResolutionContext : public IHasDbPrint, virtual public Visitor {
```
mbudiu-vmw: I guess the above comment is no longer correct.

frontends/common/resolveReferences/resolveReferences.h
```diff
@@ -32,46 +32,24 @@ enum class ResolutionType {
...
+ protected:
     /// Stack of namespaces for global declarations (e.g., match_kind)
     /// @todo: what about errors?
     std::vector<const IR::INamespace*> globals;
 
- public:
-    explicit ResolutionContext(const IR::INamespace* rootNamespace) :
-            rootNamespace(rootNamespace)
-    { push(rootNamespace); }
+    std::vector<const IR::IDeclaration*>*
```
mbudiu-vmw: I think that the above todo is no longer necessary.

frontends/common/resolveReferences/resolveReferences.cpp
ChrisDodd: The old code appends the globals to the end of the `toTry` vector then iterates over `toTry` in *reverse*, so the globals are checked first, not last.  That seems wrong to me, but for now I was just attempting to duplicate the former behavior without change.

frontends/common/resolveReferences/resolveReferences.cpp
ChrisDodd: This is the old code (mostly) unchanged, but the indentation changed (moved it out of the loop into a separate `lookup` function, so that it could be called on just the global scope for absolute paths, rather than creating a new ResolveContext with just the global scope on its lookup stack).

I'm thinking of refactoring this by making much of it a method on ISimpleNamespace/IGeneralNamespace, using vitual dispatch rather than dynamic cast tests.

frontends/common/resolveReferences/resolveReferences.cpp
ChrisDodd: Yes, I'll look at that.  Unfortunately the visitor infrastructure does not support pre/postorder functions on IR `interface` types, which could make this much cleaner.  I'm thinking about how to extend the ir-generator to fix that.

frontends/common/resolveReferences/resolveReferences.h
ChrisDodd: I haven't yet tracked down what happens with errors -- it seems that we currently require them to be explicit (eg, `error.NoMatch` rather than just `NoMatch`).

ChrisDodd:
Updated to deal with overload resolution.

ChrisDodd:
I've been maintaining this change for over a year (rebasing it on master every couple weeks) and it has yet to have any problems, so I think it is safe to merge.  I haven't yet managed to get the removal of refmap from TypeInferencing working, but that is in a separate branch.

mbudiu-vmw:
If you want to merge this I think you should also check that it addresses #2267.
Does this affect the running time? The refMap acts as a cache, and it's not obvious to me that re-computing the references every time does not blow up. (On the other hand it could be cheaper if you don't recompute stuff you don't really need...).

ChrisDodd:
So far it does not seem to affect the running time measurably (not suprising as it does not ye get rid of the refmap anywhere), though overall getting rid of the reference map and its repeated recomputes should make things faster.

#2267 is a TypeInference issue -- it tries to look up a type for a table before it has added that type to the typeMap (because adding is done in the postorder) -- and is orthogonal to this.

mbudiu-vmw:
Whether #2267 is a type inference or a reference resolution problem depends on how we choose to resolve identifiers. Can the key of a table refer to the table itself? If yes, then it's a problem with type inference which can be fixed by checking for this particular pathological case. I tried to give a more general solution based on reference resolution forbidding an object to refer to the surrounding declaration (except in parser states).

ChrisDodd:
We resolve identifiers lexically -- any identifier is resolved to the lexically enclosing definition, in some cases restricted to just those definitions after the use, but not in others (IMO a confusing and somewhat broken restriction we should remove).  A table key cannot (sematically) refer to any table (itself or otherwise), but doing so is a semantic error, not a syntax error.  So #2267 is well-defined; it's a semantic type error and should give such an error diagnostic (invalid type in table key or some such).

PR #922, opened 2017-09-19
Better alias analysis in side-effects pass

mbudiu-vmw:
This PR contains a subset of the changes from the reverted PR (which changed v1model).
This PR only contains the improved alias analysis, which is useful in itself.
The main effect is to create fewer unneeded temporaries in the side-effect pass.

### DIFF
--- a/frontends/p4/sideEffects.cpp
+++ b/frontends/p4/sideEffects.cpp
@@ -310,21 +310,193 @@ class DismantleExpression : public Transform {
     const IR::Node* preorder(IR::LAnd* expression) override { return shortCircuit(expression); }
     const IR::Node* preorder(IR::LOr* expression) override { return shortCircuit(expression); }
 
-    // We don't want to compute the full read/write set here so we
-    // overapproximate it as follows: all declarations that occur in
-    // an expression.
-    // TODO: this could be made more precise, perhaps using LocationSets.
-    class ReadsWrites : public Inspector {
-     public:
-        std::set<const IR::IDeclaration*> decls;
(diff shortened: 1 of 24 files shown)


frontends/p4/sideEffects.cpp
```diff
@@ -430,6 +595,10 @@ class DismantleExpression : public Transform {
                         break;
                     if (!p1->hasOut() && !p2->hasOut())
                         continue;
+                    if (useTemporary.find(p1) != useTemporary.end())
```
ChrisDodd: idiomatically, I prefer `!useTemporary.count(p1)` to test if something is not in a set or map.  Mostly because it is easier to apply to something that has side effects (such as the result of a call that returns a set)

frontends/p4/sideEffects.cpp
```diff
@@ -310,40 +310,205 @@ class DismantleExpression : public Transform {
...
+        }
+    };
+
+    /// We represent a set of location set as a set of LocationPath
+    /// objects.
+    class SetOfLocations : public IHasDbPrint {
      public:
-        std::set<const IR::IDeclaration*> decls;
-        ReferenceMap* refMap;
+        std::set<const LocationPath*> paths;
```
ChrisDodd: Using sets of pointers can lead to non-deterministic behavior, as the order of iteration over the set depends on what the garbage collector does.  Since all the iterations here *should* be order independent, that's probably not an issue here, but that's what we have `ordered_set` and `ordered_map` for,

PR #923, opened 2017-09-20
Add unified compilation support for cmake

sethfowler:
Finally, the joys of unified compilation have returned to us.

Check out the comments in `UnifiedBuild.cmake` and in particular on the `build_unified()` function. It's pretty straightforward to use. `build_unified(FOO_SRCS)` generates a set of unified sources for the files in `FOO_SRCS` and then modifies `FOO_SRCS` to refer to those files instead of the original files.

Two options are exposed:
- `UNIFIED_COMPILATION_ENABLED` controls whether unified compilation is enabled at all.
- `UNIFIED_SOURCE_CHUNK_SIZE` controls the chunk size - i.e., the number of source files included in each unified source file. You can set it to `ALL` to unify as thoroughly as possible.

Both the standard `GENERATED` property and the custom `NONUNIFIED` property will exclude a source file from being unified, even if it gets passed to `build_unified()`.

When you call `build_unified()`, you can override the global `UNIFIED_SOURCE_CHUNK_SIZE`. I use this in a number of places to pass `ALL` for parts of the compiler that change very rarely, since in those cases there isn't much need to balance full build speed with incremental compilation speed.

One thing that I currently don't do is unify GTests added by backends or extensions. It's just too error-prone as things standard now; we should rework the GTest extension mechanism to use actual libraries. Consider that future work.

Some numbers:

```
NONUNIFIED BUILD: -j8, empty ccache
      161.26 real      1049.17 user        48.95 sys 
UNIFIED BUILD: -j8, empty ccache
       52.94 real       262.93 user        12.42 sys
```

1/4 the CPU time. 3x faster in terms of wall clock time. With this patch, a clean full build takes under a minute.


sethfowler:
Benchmarking on Travis can be a little tricky now that we're using ccache. It's important to compare two runs with no cache hits when evaluating build performance. I just ran a non-cache build for master so we could compare this PR with the status quo:

```
NONUNIFIED BUILD on Travis:
    - 20 min 5 sec
UNIFIED BUILD on Travis:
    - 7 min 42 sec
```

More than 2.5x as fast.

backends/bmv2/CMakeLists.txt
```diff
@@ -66,18 +67,18 @@ set (BMV2_BACKEND_HDRS
...
+build_unified(BMV2_SRCS)
 add_executable(p4c-bm2-ss ${BMV2_SRCS})
 target_link_libraries (p4c-bm2-ss bmv2backend ${P4C_LIBRARIES} ${P4C_LIB_DEPS})
 
 install(TARGETS p4c-bm2-ss RUNTIME DESTINATION ${P4C_RUNTIME_OUTPUT_DIRECTORY})
 
-set (BMV2_SOURCES ${BMV2_BACKEND_SRCS} ${BMV2_SRCS} ${BMV2_BACKEND_HDRS})
-add_cpplint_files (${CMAKE_CURRENT_SOURCE_DIR} "${BMV2_SOURCES}")
-
 # hack to get around the fact that the test scripts expect the backend
```
mbudiu-vmw: Does the comment below still apply?

sethfowler:
You do lose some potential ccache hits, sure, but the overall speed up is so drastic that it's a more than worthwhile trade off.

backends/bmv2/CMakeLists.txt
sethfowler: Nothing has really changed for code outside this file yet, so I believe that comment still applies (though I'll look over it again when I'm not on my phone). I plan to rework the GTest extension stuff in a followup PR, so expect some cleanup of this file soon.

sethfowler: Oh, I see. I thought this was in reference to a different comment in `test/CMakeLists.txt`.

I had the same question when I was looking at this file. I would expect this to be obsolete, since after all we've long since remove automake. @cc10512, can we remove this hack?

cc10512: We did remove automake, but the runscripts are still hardcoded for the build/p4c path. Ideally we'll change the runscripts to invoke p4c (the driver), and then all these can be removed. That and how we handle includes. Again, p4c does solve that issue as well.

sethfowler: OK. File it under "TODO", then.

cmake/UnifiedBuild.cmake
```diff
@@ -0,0 +1,174 @@
+# Options to control unified compilation. More information is available in the
+# documentation for `build_unified`.
+OPTION (UNIFIED_COMPILATION_ENABLED "Merge source files to speed up compilation" ON)
```
cc10512: can you please rename this to ENABLE_UNIFIED_COMPILATION? All our "enable" options start like this.

sethfowler: OK, will do.

mbudiu-vmw:
I think something is wrong with the unified build. I have changed one file in the front-end (frontends/p4/fromv1.0/converters.cpp) and the compilation is rebuilding lots of other object files which should be completely independent.

sethfowler:
@mbudiu-vmw this is what happens for me:

```
○ > touch ../frontends/p4/fromv1.0/converters.cpp
○ > make -j8
[  0%] Built target mkv1dirs
[  0%] Built target linkgraphs
[  0%] Built target mkp4dirs
[  0%] Built target update_includes
[  0%] Built target linkbmv2
[  1%] Built target linkp4test
[  3%] Built target p4ctoolkit
[  7%] Built target gtest
[  7%] Built target mkP4configdir
[  7%] Built target linkp4cebpf
[  8%] Built target p4c_driver
[ 17%] Built target irgenerator
[ 19%] Built target genIR
[ 25%] Built target ir
[ 32%] Built target midend
[ 44%] Built target controlplane
Scanning dependencies of target frontend
[ 50%] Built target bmv2backend
[ 51%] Building CXX object frontends/CMakeFiles/frontend.dir/unified_frontend_sources_4.cpp.o
[ 53%] Linking CXX static library libfrontend.a
[ 76%] Built target frontend
[ 80%] Linking CXX executable p4c-bm2-ss
[ 80%] Linking CXX executable p4c-ebpf
[ 82%] Linking CXX executable p4test
[ 83%] Linking CXX executable gtestp4c
[ 85%] Linking CXX executable p4c-graphs
[ 87%] Built target p4c-graphs
[ 89%] Built target p4test
[ 91%] Built target p4c-ebpf
[ 92%] Built target p4c-bm2-ss
[100%] Built target gtestp4c
○ >
```

Just one file gets rebuilt. Are you seeing something different?

mbudiu-vmw:
I can't reproduce this problem; if it happens again I will post some more detailed logs.
Sorry for the trouble.

sethfowler:
No worries. Thanks for the bug report. =) If there are any problems, I definitely want to get them fixed.

PR #926, opened 2017-09-20
Fix build for P4Runtime gtest

antoninbas:
p4runtime.pb.h includes google/rpc/status.pb.h which means that
build/control-plane has to be in the path


antoninbas:
I believe Cmake's add_directories will give priority to local files.

Issue #927, opened 2017-09-20
Build failure on OS X with xcode 9.0


hanw:
I, unfortunately, upgraded osx to use xcode 9.0 today. Now p4lang/p4c is failing to build.

Below is a snippet of the error messages.

```
make[2]: *** [frontends/CMakeFiles/frontend.dir/common/parseInput.cpp.o] Error 1
In file included from /Users/hanwang/p4c/backends/bmv2/control.cpp:17:
In file included from /Users/hanwang/p4c/backends/bmv2/backend.h:20:
In file included from /Users/hanwang/p4c/backends/bmv2/analyzer.h:20:
In file included from /Users/hanwang/p4c/ir/ir.h:25:
In file included from /Users/hanwang/p4c/lib/null.h:22:
In file included from /Users/hanwang/p4c/lib/error.h:26:
In file included from /Users/hanwang/p4c/lib/source_file.h:29:
In file included from /Users/hanwang/p4c/lib/map.h:20:
/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/include/c++/v1/map:820:5: error: static_assert failed "Allocator::value_type must be same type as value_type"
    static_assert((is_same<typename allocator_type::value_type, value_type>::value),
    ^             ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/Users/hanwang/p4c/lib/ordered_map.h:63:41: note: in instantiation of template class 'std::__1::map<const cstring *, std::__1::__list_iterator<std::__1::pair<const cstring, Util::IJson *>, void *>, ordered_map<cstring, Util::IJson *, std::__1::less<cstring>, std::__1::allocator<std::__1::pair<const cstring, Util::IJson *> > >::mapcmp, std::__1::allocator<std::__1::pair<const cstring, Util::IJson *> > >' requested here
    map_type                            data_map;
                                        ^
/Users/hanwang/p4c/lib/json.h:124:47: note: in instantiation of template class 'ordered_map<cstring, Util::IJson *, std::__1::less<cstring>, std::__1::allocator<std::__1::pair<const cstring, Util::IJson *> > >' requested here
class JsonObject final : public IJson, public ordered_map<cstring, IJson*> {
                                              ^
In file included from /Users/hanwang/p4c/backends/bmv2/control.cpp:17:
[...]

PR #928, opened 2017-09-20
fix ordered_map bug for clang 9.0 on osx

hanw:
fix for issue #927

### DIFF
--- a/lib/ordered_map.h
+++ b/lib/ordered_map.h
@@ -36,7 +36,7 @@ class ordered_map {
     typedef const value_type            &const_reference;
 
  private:
-    typedef std::list<value_type, ALLOC>                list_type;
+    typedef std::list<value_type>                       list_type;
     list_type                                           data;
 
  public:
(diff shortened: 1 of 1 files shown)


lib/ordered_map.h
```diff
@@ -36,7 +36,7 @@ class ordered_map {
     typedef const value_type            &const_reference;
 
  private:
-    typedef std::list<value_type, ALLOC>                list_type;
+    typedef std::list<value_type>                       list_type;
```
ChrisDodd: This is unnecessary -- the allocator was correct

lib/ordered_map.h
```diff
@@ -59,7 +59,7 @@ class ordered_map {
     struct mapcmp : std::binary_function<const K*, const K*, bool> {
         COMP    comp;
         bool operator()(const K *a, const K *b) const { return comp(*a, *b); } };
-    typedef std::map<const K *, iterator, mapcmp, ALLOC>  map_type;
+    typedef std::map<const K *, iterator, mapcmp>  map_type;
```
ChrisDodd: This avoids the problem by using the default allocator, which is wrong if anyone ever tries to use any other allocator...

lib/ordered_map.h
mrwtakev: Where is this file？And how can I find this file?

PR #932, opened 2017-09-22
Fix clang warnings in p4RuntimeSerializer.cpp

sethfowler:
Looks like some new unused parameter warnings have crept in recently.


mbudiu-vmw:
My last pr also fixes these

sethfowler:
@mbudiu-vmw this fix is just a bit more thorough; in this patch I fully removed the now-unneeded parameters.

PR #939, opened 2017-09-26
Introduced new checksum APIs in v1model.p4

mbudiu-vmw:


p4include/v1model.p4
```diff
@@ -57,11 +57,8 @@ struct standard_metadata_t {
     @alias("intrinsic_metadata.mcast_grp")     bit<16> mcast_grp;
     @alias("intrinsic_metadata.resubmit_flag") bit<1>  resubmit_flag;
     @alias("intrinsic_metadata.egress_rid")    bit<16> egress_rid;
-}
-
-extern Checksum16 {
```
ChrisDodd: Rather than deleting Checksum16 (which existing P4_16 programs depend on), it should be marked as `@deprecated`

mbudiu-vmw: But the compiler support for this block has been deleted.
It won't do much help to have programs that use it.

mbudiu-vmw: Unless the compiler errors for all deprecated constructs...

ChrisDodd: That's kind of the point -- you shouldn't remove support for it yet, just add support for the new way of dealing with things.

And yes, there should be a pass that checks for uses of `@deprecated` things and issues a warning about it -- the `@deprecated` could have a string literal arg with a message suggesting what it should be changed to.

mbudiu-vmw: The fundamental problem is that the old way of doing things was really broken; if I could have fixed that implementation I would have.

I should indeed add a pass for handing deprecated. It would be easiest to make it really an error rather than a warning. 

It will be quite difficult to support both variants, because not only the checksum extern changes, but also the signature of the verifyChecksum control.

ChrisDodd: It was only broken in the bmv2 backend and was working fine in other backends.

mbudiu-vmw: There are lots of filed issues about it for bmv2.

Also, I suspect there were no torture tests; technically you could have any combination of ifs, assignments and method calls in the update/verify checksum blocks. I somehow doubt that all these were handled.

mbudiu-vmw: On the other hand, the new methods should be easily converted to the old style code...
`update_checksum(c, list, field, HashAlgorithm.csum16)` is the same as
`if (c) field = ck.get(list);`

mbudiu-vmw: @ChrisDodd : assuming we merge PR #940 and we tag the Checksum16 block as deprecated, would this solution work for you? The bmv2 back-end would not support it, but your custom back-end could.

p4include/v1model.p4
```diff
@@ -151,8 +181,12 @@ extern void truncate(in bit<32> length);
...
                     out H parsedHdr,
                     inout M meta,
-                    inout standard_metadata_t standard_metadata);
-control VerifyChecksum<H, M>(in H hdr,
+inout standard_metadata_t standard_metadata);
+
+/* The only legal statements in the implementation of the
+VerifyChecksum control are: block statements, calls to the
+verify_checksum method, and return statements. */
+control VerifyChecksum<H, M>(inout H hdr,
```
ChrisDodd: Why did this first parameter become `inout`?  Is it legal for `VerifyChecksum` to change headers?

mbudiu-vmw: There was an explanation in the original PR. This is because in BMv2 you cannot use a constant for the checksum value. It *must* be a (p4-14) field. So you cannot, for example, check that the checksum is 0.

So, while we do not allow the block to change the checksum, this is how we enforce it being a left-value.

mbudiu-vmw: Alternatively, we could keep the parameter "in", but make another separate check in the back-end for the argument to the verify_checksum function.

antoninbas: Couldn't you create a field temporary in the backend if the parameter is a constant?

ChrisDodd: Not a big deal, but logically it seems that the result of verifying a checksum is a boolean -- either the checksum is valid or it is not.  I guess in this case you've added a single field to match the checksum against and in most uses it will always be a constant 0 or ~0?  This allows for weird non-standard checksum where the result is something else, but that bmv2 does not handle as well.

mbudiu-vmw: These constraints come all from modelling the way checksums are defined in p4-14 (and in consequence in bmv2), where the checksum has to indicate a fieldlist, a field, and then whether the checksum is verified or updated.

mbudiu-vmw:
These interfaces model how p4-14 and bmv2 simple_switch work very closely.
I am open to other proposals.

sethfowler:
Well, I don't necessarily have a better one; I'm more just trying to understand the motivation for the design decisions in this proposal.

p4include/v1model.p4
```diff
@@ -57,11 +57,8 @@ struct standard_metadata_t {
     @alias("intrinsic_metadata.mcast_grp")     bit<16> mcast_grp;
     @alias("intrinsic_metadata.resubmit_flag") bit<1>  resubmit_flag;
     @alias("intrinsic_metadata.egress_rid")    bit<16> egress_rid;
-}
-
-extern Checksum16 {
-    Checksum16();
-    bit<16> get<D>(in D data);
+    /// Indicates that a verify_checksum() method has failed.
+    bit<1>  checksum_error;
```
hanw: This does not work for ipv4-over-ipv4 tunnel protocol. Two bits are needed to represent checksum errors in both inner and outer IP headers. Further, why don't we use the `error` type to represent checksum error?

mbudiu-vmw:
V1model is p4-14. This is how p4-14 works.
You can invent a better way to do things that in psa. We cannot change p4-14 retroactively..

hanw:
P14 spec 1.0.4 says on page 16, `The verify option ... If the value is not equal, then a p4_-
pe_checksum exception is generated; see 6.6.1. Standard Parser Exceptions`. I interpret it as the v1model verify_checksum operation should raise the checksum_error as one of the `error` types, not setting the standard_metadata. Although, I understand the implementation might just be the same.

sethfowler:
I don't love the API (I'm particularly not a fan of the boolean arguments, though I understand that the motivation is to simplify the implementation in the compiler) but it's clear that it solves real problems and will enable us to move forward and focus on more important work.

Consider it approved on my end, at least.

sethfowler:
"More important work" should be understood to include designing a cleaner checksum API for PSA. =)

sethfowler:
I don't mean to imply that fixing the v1model checksum API *isn't* important work, by the way. It definitely is. Thank you for doing it, Mihai.

What I'm really trying to say is that I don't think we need to put a huge amount of effort into bike shedding the v1model architecture into perfection, because we know that in the long term, users will be migrating away from it. By being pragmatic about v1model, I'm hoping that we can direct our energy towards improving the architectures those users will be migrating *to*.

jnfoster:
Care to comment here? https://github.com/p4lang/p4-spec/pull/421

mbudiu-vmw:
Yes, I entirely agree. I think we should stop investing in v1model as soon as possible and focus on PSA. v1model is really there just for backward compatibility.

@ChrisDodd : does this version of v1model work for your legacy programs with a custom back-end?

Issue #935, opened 2017-09-22
Add support for p4runtime code generation to the EBPF backend


familysingh:
Currently, only the p4c-bm2-ss backend supports p4runtime code generation.  It would be very helpful to add the same support to at least the EBPF backend - thanks.

antoninbas:
I've added support in the p4test backend today (https://github.com/p4lang/p4c/pull/962)
However I'm not sure I see the point of adding support in the eBPF backend if no P4Runtime implementation exists for this target.

familysingh:
Adding to p4test sounds good - thanks!  You can close this issue.

PR #941, opened 2017-09-27
Make P4Runtime generation backend independent

sethfowler:
We're at the point now where we're ready to make P4Runtime generation backend independent. This PR eliminates all remaining dependencies on post-frontend passes. This is accomplished by moving the passes that *are* required inside `generateP4Runtime()`. This is a funny definition of "eliminate", I'll grant you. =) The point is that the users of the API don't need to know about these dependencies anymore, though. I've triaged them all; now that we have the full list, we can start gradually whittling it down until there's nothing left.

To make this work, we do have to move the `BMV2::SynthesizeValidField` pass into the midend code, because we need this available on all backends. Everyone, including me, has concerns about that pass, so I can understand some discomfort at making this move, but consider it a temporary matter until we can get something better in place. I added a warning to the documentation for that pass telling people that they shouldn't write new backends that use it.

I have really wanted for a while now to fix the cpplint and unified build woes that result from `ir/std.h` placing a `vector` class in the global namespace; that vector class triggers issues in the generated protobuf headers that we've been working around in various ways for a while. As part of this PR I went ahead and solved the root of the problem by renaming that `vector` class to `safe_vector` and moving it to `lib/safe_vector.h`. It turns out that if you make that change and add a few missing `std::`s and `#include`s here and there, we don't need `ir/std.h` at all anymore, so I went ahead and removed it as part of this patch. It pulled in a lot of STL headers, so that should improve compilation speed a little for translation units that otherwise didn't need them.

### DIFF
--- a/backends/bmv2/bmv2.cpp
+++ b/backends/bmv2/bmv2.cpp
@@ -104,12 +104,11 @@ int main(int argc, char *const argv[]) {
     // Generate a PI control plane API for this program if requested.
     if (!options.p4RuntimeFile.isNullOrEmpty()) {
         std::ostream* out = openFile(options.p4RuntimeFile, false);
-        std::ostream *outEntries = nullptr;
+        std::ostream* outEntries = nullptr;
         if (!options.p4RuntimeEntriesFile.isNullOrEmpty())
             outEntries = openFile(options.p4RuntimeEntriesFile, false);
         if (out != nullptr) {
-            auto p4Runtime =
-              generateP4Runtime(program, toplevel, &midEnd.refMap, &midEnd.typeMap);
+            auto p4Runtime = P4::generateP4Runtime(program);
             p4Runtime.serializeP4InfoTo(out, options.p4RuntimeFormat);
             if (outEntries != nullptr)
                 p4Runtime.serializeEntriesTo(outEntries, options.p4RuntimeFormat);
--- a/frontends/p4/fromv1.0/converters.cpp
+++ b/frontends/p4/fromv1.0/converters.cpp
@@ -508,10 +508,10 @@ const IR::Statement *PrimitiveConverter::cvtPrimitive(ProgramStructure *structur
     return nullptr;
 }
 
-vector<const IR::Expression *>
+safe_vector<const IR::Expression *>
 PrimitiveConverter::convertArgs(ProgramStructure *structure, const IR::Primitive *prim) {
     ExpressionConverter conv(structure);
-    vector<const IR::Expression *> rv;
+    safe_vector<const IR::Expression *> rv;
     for (auto arg : prim->operands)
         rv.push_back(conv.convert(arg));
     return rv;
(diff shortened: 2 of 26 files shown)


backends/bmv2/bmv2.cpp
```diff
@@ -104,12 +104,11 @@ int main(int argc, char *const argv[]) {
     // Generate a PI control plane API for this program if requested.
     if (!options.p4RuntimeFile.isNullOrEmpty()) {
         std::ostream* out = openFile(options.p4RuntimeFile, false);
-        std::ostream *outEntries = nullptr;
+        std::ostream* outEntries = nullptr;
```
mbudiu-vmw: You and @ChrisDodd are moving the stars around in different directions.
We should settle for one, or make this a don't care, otherwise we'll never converge.

frontends/p4/fromv1.0/converters.cpp
```diff
@@ -508,10 +508,10 @@ const IR::Statement *PrimitiveConverter::cvtPrimitive(ProgramStructure *structur
     return nullptr;
 }
 
-vector<const IR::Expression *>
+safe_vector<const IR::Expression *>
```
mbudiu-vmw: It won't be easy for people to remember to use safe_vector.

backends/bmv2/bmv2.cpp
```diff
@@ -104,12 +104,11 @@ int main(int argc, char *const argv[]) {
     // Generate a PI control plane API for this program if requested.
     if (!options.p4RuntimeFile.isNullOrEmpty()) {
```
antoninbas: Would you say it could be a good idea to move this code to a common location so that it doesn't have to be duplicated between backends (the code that looks at the command line options and invokes `generateP4Runtime` and the `serialize` methods?
This issue for example (https://github.com/p4lang/p4c/issues/935) adds for support for the p4runtime options in the eBPF backend and I think it would be a good idea to have support in the p4test backend as well.
All the backends would still need to call one method, but that's better than copying all these lines around. We also want backends to have to actively call a method anyway, as they may need to register hooks in the future to handle specific externs.

sethfowler: Yeah, we could add a wrapper that would take care of this; this pattern is probably going to be the same in every backend. I anticipate those extern hooks being ultimately passed to `generateP4Runtime()`, but the wrapper could funnel them along to their destination.

frontends/p4/fromv1.0/converters.cpp
sethfowler: That's true. I'd argue that remembering `vector` vs `std::vector` is even more difficult, though. =)

It's probably actually a better bet to just use the built-in debugging modes for the various C++ standard library implementations (i.e., define `_GLIBCXX_DEBUG` and `_LIBCPP_DEBUG` for debug builds). That way we get coverage everywhere, and nobody has to remember anything special. I'll run some benchmarks, and if it doesn't result in a massive slowdown I'll open a PR to just use that approach and completely remove `safe_vector`.

backends/bmv2/bmv2.cpp
sethfowler: It's for the greater good. We are all too aware of what will happen when the stars align.

Issue #937, opened 2017-09-22
gtest not finding protobuf include files


rhalstea:
I'm getting the following error when trying to build the project

```
[ 95%] Building CXX object test/CMakeFiles/gtestp4c.dir/gtest/p4runtime.cpp.o
In file included from /wrk/hdstaff/halstead/work_p4c/src/p4c/test/gtest/p4runtime.cpp:19:0:
/wrk/hdstaff/halstead/work_p4c/src/p4c/build/control-plane/p4/config/p4info.pb.h:9:42: fatal error: google/protobuf/stubs/common.h: No such file or directory
 #include <google/protobuf/stubs/common.h>
                                          ^
compilation terminated.
```

It doesn't look like CMake see's protobuf as a dependency because the file that needs it doesn't exist until I start compiling: ``` build/control-plane/p4/config/p4info.pb.h```

This may only be an issue because my protobuf isn't installed in the default directories. I'm not really sure how to fix this beyond adding ```include_directories(${PROTOBUF_INCLUDE_DIRS}) ``` to ```test/CMakeLists.txt```

sethfowler:
I'll take a look. Your proposed solution sounds like the way to go; I just want to double check that there are no surprises there.

rhalstea:
May be unrelated (may not) after doing this one test fails: gtestp4c

[error.txt](https://github.com/p4lang/p4c/files/1326002/error.txt)

I'm not sure what the expected output is here, but I can say I don't install anything to ```usr/local/share/p4c/``` so the first error is valid.

sethfowler:
Sounds like #924. In our test environment we run `make install` before running the tests, which allowed that bug in the build system to creep in. We need to fix that.

rhalstea:
oops, I didn't notice that issue. Yea, that looks like the same one.

rhalstea:
I don't think the original compiling issue is a duplicate, but the followup about gtestp4c is.

rhalstea:
thank you Mihai

PR #943, opened 2017-09-27
Fix for issue #937

rhalstea:


sethfowler:
Thanks for taking this on. This looks good to me, but I'd like to have @cc10512 review it before landing.

rhalstea:
@sethfowler if there are no objections could you pull this in?

Thanks.

PR #944, opened 2017-09-28
Clean up and fix log metalogging

ChrisDodd:


lib/log.cpp
```diff
@@ -177,6 +179,12 @@ void increaseVerbosity() {
     static std::mutex lock;
     std::lock_guard<std::mutex> acquire(lock);
 #endif
+#ifdef CLOCK_MONOTONIC
```
cc10512: Shall we use http://en.cppreference.com/w/cpp/chrono/high_resolution_clock? It's supposed to handle platform specific clocks a bit cleaner than `clock_gettime`, and I have seen in previous experiments, that it is indeed at ns resolution. Just a thought :)

sethfowler: If it's not too much trouble to make the switch, I'd say that is a good idea. The standard library developers have taken care of dealing with portability for us and the API is nicer.

If you do keep this approach with `clock_gettime`, it might be worth pulling this initialization code into a separate function since we now have two copies of it.

PR #945, opened 2017-09-29
fix missing includes and namespace qualifier for non-unified build

cc10512:


test/gtest/p4runtime.cpp
```diff
@@ -34,7 +34,7 @@ createP4RuntimeTestCase(const std::string& source,
                             = CompilerOptions::FrontendVersion::P4_16) {
     auto frontendTestCase = FrontendTestCase::create(source, langVersion);
     if (!frontendTestCase) return boost::none;
-    return generateP4Runtime(frontendTestCase->program);
```
sethfowler: Hmm, why did this work before? There must be a `using namespace P4` somewhere in the test suite, I guess...

cc10512:
```

sethfowler: Doh. I'll try to remember to get rid of those at some point.

Issue #946, opened 2017-09-29
Unsupported keyword parser_value_set in p4-14


hanw:
Current frontend does not support the `parser_value_set` keyword in P4-14 in section 4.3.

```
header_type ethernet_t {
    fields {
        dstAddr : 48;
        srcAddr : 48;
        etherType : 16;
    }
}

parser start {
    return parse_ethernet;
}

parser_value_set pvs;

parser parse_ethernet {
    extract(ethernet);
    return select(latest.etherType) {
        pvs : ingress;
        default : ingress;
    }
}

table dummy {
    reads {
        ethernet.dstAddr : exact;
    }
    actions {
        noop;
    }
    size : 512;
}


action noop() {
    no_op();
}

control ingress {
    apply(dummy);
}
```

Compiling the above example returns following error:
```
$ ./p4test --p4-14 pvs.p4
pvs.p4(14):syntax error, unexpected IDENTIFIER "parser_value_set"
parser_value_set
^^^^^^^^^^^^^^^^
/Users/hanwang/pvs.p4(19): error: : Non constant expression
        pvs :
        ^^^
error: 2 errors encountered, aborting compilation
```

mbudiu-vmw:
@hanw: I will close this one, but I will file a bug with the spec, since the value_set is not described in the p4-16 spec.

PR #948, opened 2017-10-02
Frontend support for parser_value_set

hanw:
This PR implements the frontend support for parser_value_set. 

The backend support for bmv2 is TBD. There is a placeholder for it.

I also corrected the section number in v1parser.ypp to correspond to the P14 v1.0.4 spec.

Related to #946

### DIFF
--- a/frontends/p4/fromv1.0/programStructure.cpp
+++ b/frontends/p4/fromv1.0/programStructure.cpp
@@ -412,7 +412,7 @@ const IR::ParserState* ProgramStructure::convertParser(const IR::V1Parser* parse
                     cases.push_back(sc);
                 } else if (auto first = v.first->to<IR::PathExpression>()) {
                     // XXX(hanw): handle parser_value_set
-                    P4C_UNIMPLEMENTED("parser_value_set is not yet implemented");
+                    ::warning("parser_value_set is not yet implemented");
                 } else {
                     ::error("Expected constant or parser value set in %1%", v.first);
                 }
--- a/frontends/p4/fromv1.0/programStructure.cpp
+++ b/frontends/p4/fromv1.0/programStructure.cpp
@@ -414,7 +414,7 @@ const IR::ParserState* ProgramStructure::convertParser(const IR::V1Parser* parse
                     // XXX(hanw): handle parser_value_set
                     P4C_UNIMPLEMENTED("parser_value_set is not yet implemented");
                 } else {
-                    BUG("Unknown type for select case", v.first);
+                    ::error("Expected constant or parser value set in %1%", v.first);
                 }
             }
         }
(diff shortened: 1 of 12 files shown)


frontends/p4/fromv1.0/programStructure.cpp
```diff
@@ -406,9 +406,14 @@ const IR::ParserState* ProgramStructure::convertParser(const IR::V1Parser* parse
...
-                auto expr = explodeLabel(v.first, v.second, sizes);
-                auto sc = new IR::SelectCase(c->srcInfo, expr, deststate);
-                cases.push_back(sc);
+                if (auto first = v.first->to<IR::Constant>()) {
+                    auto expr = explodeLabel(first, v.second, sizes);
+                    auto sc = new IR::SelectCase(c->srcInfo, expr, deststate);
+                    cases.push_back(sc);
+                }
+                if (auto first = v.first->to<IR::StringLiteral>()) {
+                    // XXX(hanw): handle parser_value_set
```
mbudiu-vmw: At least you could give an error here.

frontends/parsers/v1/v1parser.ypp
```diff
@@ -485,12 +494,14 @@ case_value:
       { $$ = CaseValue($1 ? $1 : new IR::Constant(-1), new IR::Constant(-1)); }
     | const_expression MASK const_expression
       { $$ = CaseValue($1 ? $1 : new IR::Constant(-1), $3 ? $3 : new IR::Constant(0)); }
+    | name
+      { $$ = CaseValue(new IR::StringLiteral(@1, $1), new IR::Constant(-1)); }
```
mbudiu-vmw: This should be an identifier, not a string literal.

ir/v1.def
```diff
@@ -252,8 +252,15 @@ class CalculatedField {
...
+class ParserValueSet {
+    optional ID                 name;
+    Annotations                 annotations;
+    dbprint { out << node_type_name() << " " << name; }
+    toString { return node_type_name() + " " + name; }
+}
+
 class CaseEntry {
-    safe_vector<std::pair<Constant, Constant>>  values = {};
+    safe_vector<std::pair<Literal, Constant>>    values = {};
```
ChrisDodd: This should be a `safe_vector<Expression>`, though that will require more changes in everything that visits a CaseEntry (the Expression must be a Constant, a Mask, or (now) a PathExpression).  The P4_16 SelectCase already has an Expression here, so it should actually be simpler to deal with.

frontends/parsers/v1/v1parser.ypp
ChrisDodd: Actually probably should be a PathExpression

ir/v1.def
ChrisDodd: Alternately, to keep the change smaller, could be `safe_vector<std::pair<Expression, Constant>>`, keeping the mask split off.

hanw:
@mbudiu-vmw I am going to map parser-value-set to the following extern in P16. ```// Parser value set
// The parser value set implements a run-time updatable values that is used to
// determine parser transition
extern value_set<D> {
    value_set(bit<8> size);
    bool is_member(in D data);
}```

hanw:
Thanks, I will do both bmv2 support and conversion to P16 in the next PR.

Issue #951, opened 2017-10-02
Pretty-printing loses declaration


mbudiu-vmw:
```
#include <core.p4>
const bit x= 1;
```

When pretty-printing with `--pp x.p4` the const is not output.

PR #954, opened 2017-10-03
Fix for issue #951

mbudiu-vmw:


testdata/p4_16_samples_outputs/issue951-frontend.p4
```diff
@@ -0,0 +1,2 @@
+#include <core.p4>
+
```
ChrisDodd: Why does this output not include the declaration?  Dead code elimination?

mbudiu-vmw: Yes - in fact unused declaration elimination.
Previously it was also missing in --pp, before any dead code was executed.

PR #956, opened 2017-10-03
Add comments for some ControlFlowVisitor methods.

cole-barefoot:
It always takes me a bit to remember how this works.  This time I took notes.

### DIFF
--- a/ir/visitor.h
+++ b/ir/visitor.h
@@ -201,7 +201,7 @@ class Visitor {
     virtual void init_join_flows(const IR::Node *) { assert(0); }
 
     /** If @n is a join point in the control flow graph (i.e. has multiple incoming
-     * edges), then:
+     * edges) and is not filtered out by `filter_join_point`, then:
      *
      *   - if this is the first time a visitor has visited @n, store a clone of the
      *   visitor with this node and return true, deferring visiting this node until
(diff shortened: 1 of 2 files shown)


ir/visitor.h
```diff
@@ -193,8 +197,31 @@ class Visitor {
     // flow_merge the visitor from all the parents before visiting the node and its
     // children.  This only works for Inspector (not Modifier/Transform) currently.
     bool joinFlows = false;
+
     virtual void init_join_flows(const IR::Node *) { assert(0); }
+
+    /** If @n is a join point in the control flow graph (i.e. has multiple incoming
+     * edges), then:
```
ChrisDodd: ... has multiple incoming edges AND is not filtered out by filter_join_point

ir/visitor.h
```diff
@@ -287,6 +314,11 @@ class ControlFlowVisitor : public virtual Visitor {
     virtual ControlFlowVisitor *clone() const = 0;
     void init_join_flows(const IR::Node *root) override;
     bool join_flows(const IR::Node *n) override;
+
+    /** @returns true if this node should be removed from the set of join
+     * points, where a 'join point' is a node with multiple incoming edges in
+     * the control flow graph.
```
ChrisDodd: This will be called for all nodes with multiple incoming edges, and should return 'false' if the node should be considered a join point, and 'true' if if should not be considered one.  Nodes with only one incoming edge are never join points.

PR #964, opened 2017-10-08
bitvec::operator< (needed to use bitvec as a map key)

ChrisDodd:


ChrisDodd:
I'm not sure what you mean.  This just compares bitvecs to create a total ordering of them, by essentially treating them as bigints.  So there's nothing in particular about the length.

Issue #950, opened 2017-10-02
Program generates wrong output


mbudiu-vmw:
This was reported on the mailing list by Kuo-Feng Hsu.

Here is the program:
```
/* -*- P4_16 -*- */
#include <core.p4>
#include <v1model.p4>

const bit<16> TYPE_IPV4 = 0x800;

typedef bit<9>  egressSpec_t;
typedef bit<48> macAddr_t;
typedef bit<32> ip4Addr_t;

header ethernet_t {
     macAddr_t dstAddr;
     macAddr_t srcAddr;
     bit<16>   etherType;
}

header ipv4_t {
     bit<4>    version;
     bit<4>    ihl;
     bit<8>    diffserv;
     bit<16>   totalLen;
     bit<16>   identification;
     bit<3>    flags;
     bit<13>   fragOffset;
     bit<8>    ttl;
     bit<8>    protocol;
     bit<16>   hdrChecksum;
     ip4Addr_t srcAddr;
     ip4Addr_t dstAddr;
}

header flag {
    bit<8>    flag;
}

struct metadata {
     bool test_bool;
}

struct headers {
    ethernet_t   ethernet;
    ipv4_t       ipv4;
    flag         f;
}

parser ParserImpl(packet_in packet,
                   out headers hdr,
                   inout metadata meta,
                   inout standard_metadata_t standard_metadata) {


     state start {
         packet.extract(hdr.ethernet);
         transition select(hdr.ethernet.etherType) {
             TYPE_IPV4: parse_ipv4;
             default: reject;
         }
     }

     state parse_ipv4 {
         packet.extract(hdr.ipv4);
         transition accept;
     }

}

control verifyChecksum(in headers hdr, inout metadata meta) {
     apply {  }
}

control ingress(inout headers hdr, inout metadata meta, inout standard_metadata_t standard_metadata) {
    action drop() {
        mark_to_drop();
    }

    action ipv4_forward(macAddr_t dstAddr, egressSpec_t port) {
        standard_metadata.egress_spec = port;
        hdr.ethernet.srcAddr = hdr.ethernet.dstAddr;
        hdr.ethernet.dstAddr = dstAddr;
        hdr.ipv4.ttl = hdr.ipv4.ttl-1;
        meta.test_bool = true;
     }

    table ipv4_lpm {
        key = {
            hdr.ipv4.dstAddr: lpm;
        }
        actions = {
            NoAction;
            ipv4_forward;  drop;
        }
        size = 1024;
[...]

PR #965, opened 2017-10-08
Fix for issue #950

mbudiu-vmw:


jafingerhut:
I don't know enough about compiler internals to know whether this is a complete fix, but at least with the example program from issue #950, the JSON produced with this proposed fix now has "not (d2b(meta.test_bool))" in the JSON expression of the "if" condition, versus the original "not meta.test_bool".  The extra d2b() call looks correct to me, and necessary in order for simple_switch to correctly evaluate the condition.

With non-IPv4 packets, simple_switch correctly evaluates the condition for 10 consecutive packets.

mbudiu-vmw:
@sethfowler: There seems to be something wrong with the Travis job, it gives some strange warnings and segfaults.
I cannot reproduce any of these build problems on my machine.

PR #966, opened 2017-10-09
allow annotation on extern methods

hanw:


frontends/parsers/p4/p4parser.ypp
```diff
@@ -556,12 +556,13 @@ functionPrototype
...
-    : functionPrototype ";" { $$ = $1; }
-    | ABSTRACT functionPrototype ";"    { $$ = $2; $$->setAbstract(); }  // experimental
-    | TYPE "(" parameterList ")" ";"  // constructor
-                                        { auto par = new IR::ParameterList(@3, *$3);
-                                          auto mt = new IR::Type_Method(@1, par);
-                                          $$ = new IR::Method(@1, IR::ID(@1, $1), mt); }
+    : optAnnotations functionPrototype ";" { $$ = $2; $2->annotations = $1; }
+    | optAnnotations ABSTRACT functionPrototype ";"    { $$ = $3; $$->setAbstract();
+                                                         $3->annotations = $1; }  // experimental
+    | optAnnotations TYPE "(" parameterList ")" ";"  // constructor
```
mbudiu-vmw: I don't know if annotations on the constructor are useful; you already have annotations on the type declaration.

mbudiu-vmw: In fact, you are not using the annotations here.

hanw: Annotation on constructor is used in the 4th argument in `new IR::Method()`

hanw: Without the `optAnnotations` on the constructor, there is a shift/reduce conflict.

hanw:
I tested with the following program, and `dump` the top level `IR::P4Program`. How do I add a test case with reference output to the test suite?

```
enum counter_type_t {
    BYTES,
    PACKETS
}
@annotate("test")
extern counter<I> {
    @annotate("test")
    counter(I inst);
    @annotate("test")
    void count(counter_type_t type, in I count);
}
```

mbudiu-vmw:
To add a test copy the file to the testsuite/p4_16_samples and rerun 'cmake ..; make check'.
This will generate all reference outputs, which you can then add to git.

hanw:
@mbudiu-vmw PR for spec https://github.com/p4lang/p4-spec/pull/439

testdata/p4_16_samples_outputs/annotations-first.p4
```diff
@@ -1,6 +1,7 @@
 @annotest const bit<1> b = 1w1;
 extern Annotated {
     Annotated();
+    void execute(bit<8> index);
```
ChrisDodd: looks like ToP4 is failing to output the annotations on the methods?  Or is something deleting them?

mbudiu-vmw: In this case you may have to do some more work in toP4.cpp

mbudiu-vmw:
This is not finished, you have to implement the support in toP4, as @ChrisDodd pointed.

hanw:
addressed Chris's comment in #967

PR #969, opened 2017-10-13
parser graph generation support for graphs backend

c3m3gyanesh:
Generates a graph representation of top-level parser block as a dot file.
Added an example of parser graph for reference.


backends/graphs/controls.cpp
```diff
@@ -67,6 +67,18 @@ class EdgeSwitch : public EdgeTypeIface {
     const IR::Expression *labelExpr;
 };
 
+class EdgeSelect : public EdgeTypeIface {
```
antoninbas: This is not used in this file, so I'm not sure why it is defined here

backends/graphs/graphs.h
```diff
@@ -14,8 +14,8 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-#ifndef _BACKENDS_GRAPHS_CONTROLS_H_
```
antoninbas: it's a little confusing to call this file `graphs.h`, given that none of the classes it prototypes are implemented in `graphs.cpp`.

backends/graphs/graphs.h
```diff
@@ -149,6 +149,78 @@ class ControlGraphs : public Inspector {
...
+        cstring name;
+        VertexType type;
+    };
+    class GraphAttributeSetter;
+    // The boost graph support for graphviz subgraphs is not very intuitive. In
+    // particular the write_graphviz code assumes the existence of a lot of
+    // properties. See
+    // https://stackoverflow.com/questions/29312444/how-to-write-graphviz-subgraphs-with-boostwrite-graphviz
+    // for more information.
+    using GraphvizAttributes = std::map<cstring, cstring>;
```
antoninbas: it seems you pasted a lot of code from `ControlGraphs` here, and that these 2 classes would benefit from inheriting from a common base class

backends/graphs/graphs.h
```diff
@@ -149,6 +149,78 @@ class ControlGraphs : public Inspector {
...
+    vertex_t add_vertex(const cstring &name, VertexType type);
+    vertex_t add_and_connect_vertex(const cstring &name, VertexType type);
+    void add_edge(const vertex_t &from, const vertex_t &to, const cstring &name);
+
+    bool preorder(const IR::PackageBlock *block) override;
+    bool preorder(const IR::ParserBlock *block) override;
+    bool preorder(const IR::P4Parser *pars) override;
+    bool preorder(const IR::ParserState *state) override;
+
+    void writeGraphToFile(const Graph &g, const cstring &name);
```
antoninbas: as underlined in my comment above, duplicating this method in `ControlGraphs` and `ParserGraphs` looks like a bad design

backends/graphs/parser.cpp
```diff
@@ -0,0 +1,294 @@
...
+
+bool ParserGraphs::preorder(const IR::ParserState *state) {
+    std::stringstream sstream;
+
+    if (state->selectExpression != nullptr) {
+        auto v = add_and_connect_vertex(state->name, VertexType::HEADER);
+        parents.clear();
+
+        if (state->selectExpression->is<IR::PathExpression>()) {
+            parents = {{v, new EdgeSwitch(new IR::DefaultExpression())}};
```
antoninbas: these edge classes are not visible from this file; this will break non-unified builds

backends/graphs/controls.cpp
c3m3gyanesh: This will be removed. Remnant code from my initial experimentation with code.

backends/graphs/graphs.h
c3m3gyanesh: It is the common header file for both control.cpp and parser.cpp.
If you have any other suggestion for names, i would appreciate that.

backends/graphs/parser.cpp
c3m3gyanesh: I don't understand what you said? I am pretty new to c++
Can you give little more details, plz. ?

antoninbas: Well there is no `EdgeSwitch` declaration visible from this file. It is in controls.cpp. I believe that this is only compiling because we use a unified-build by default, which merges all the source files in this directory before compiling them. Some people build this code in a non-unified way, it which case it would fail. You should try compiling without a unified build (run cmake with `UNIFIED_COMPILATION_ENABLED` set to false). It really should be a regression for this repo IMO but it's not.

backends/graphs/graphs.h
antoninbas: Maybe `graph_generators.h` if you want to have a single header file for both control and parser

ChrisDodd: If you refactor `ParserGraphs` and `ControlGraphs` into a single `Graphs` (abstract) base class with `ParserGraphs` and `ControlGraphs` as subclasses, then I think this organization does make a fair amount of sense.

mbudiu-vmw:
@sethfowler @cc10512 @ChrisDodd : there are several builds that fail due to some errors that should not happen, including this one. Is the Travis setup for testing wrong?

cc10512:
@mbudiu-vmw agreed. It is likely the -march=native that breaks it when running in a VM.

backends/graphs/graphs.h
c3m3gyanesh: I tried to refactor the classes and created a Graphs base class.

backends/graphs/graphs.h
c3m3gyanesh: Now the code duplication is minimized.

c3m3gyanesh:
@mbudiu-vmw I updated the graphs to use rectangle. Please take a look at the example graph which is part of the commit and let me know your thoughts.

backends/graphs/graphs.cpp
```diff
@@ -55,6 +55,116 @@ MidEnd::MidEnd(CompilerOptions& options) {
     });
 }
 
+class Graphs::GraphAttributeSetter {
```
antoninbas: I personally think that the `main` function for the backend and the `Graph` methods implementation belong in 2 different files; apart from that it looks ok.

c3m3gyanesh: created a new file "p4c-graphs.cpp" to accommodate `main` funcion.

c3m3gyanesh:
Is there something else remain to be done from my side for this pull request to be merged? I would appreciate any input here to understand the pull/merge process. 
Thank you.

PR #975, opened 2017-10-18
use the tag label to select tests when running make check

cc10512:
@ChrisDodd I believe the reason for using regular expressions instead of the labels was consistency with the way we list labels, however, using labels for the actual run and XFAIL for listing the failures works, since the xfails are defined per tag.


sethfowler:
@cc10512 will you revisit this, or should we close the PR?

mbudiu-vmw:
Is this PR still relevant or can we close it?

PR #976, opened 2017-10-18
update simple_router.p4 test case to set egress_spec, not egress_port

hanw:
original test program does not conform to the p4-14 spec.


hanw:
Thanks!

PR #977, opened 2017-10-18
Fail test if compiler crashes

mbudiu-vmw:


backends/bmv2/run-bmv2-test.py
```diff
@@ -208,10 +208,15 @@ def process_file(options, argv):
         args[0:0] = options.runDebugger.split()
         os.execvp(args[0], args)
     result = run_timeout(options, args, timeout, stderr)
+    # If the compiler crashed fail the test
+    if 'Compiler Bug' in open(stderr).readlines():
```
cc10512: @mbudiu-vmw: isn't the return code != SUCCESS in when throwing a 'Compiler Bug'?  Why do you think it needs special handling?
And if it is, then you could move this code after you print stderr in line 217.

mbudiu-vmw: Yes, because tests that are supposed to return with an error (illegal programs) should still fail when the compiler crashes. Today they will succeed.

mbudiu-vmw:
@cc10512 : is this better?

PR #994, opened 2017-10-27
remove v1.1 test case that is no longer supported

hanw:
Do we still need to keep this v1.1 test case around? It is not a valid P4-14 program.


sethfowler:
Perhaps in practice the standard should be that we support the program if `p4c-bm` supports it, since we'd ideally like to move users from that compiler to this one in the long term.

Issue #995, opened 2017-10-28
Incorrect masks generated for parser transitions in JSON


4tXJ7f:
Compiler version: master from October 26

p4c is generating wrong masks in the JSON file if only parts of a case tuple in a select statement contains masks. The attachment contains a small test case.

Select statement from the example:
```
...
        transition select(hdr.ethernet.srcAddr, hdr.ethernet.dstAddr) {
            (0x12f           , 0x456             ): a1;
            (0x12f &&& 0xffff, 0x456             ): a2;
            (0x12f           , 0x456 &&& 0xfff   ): a3;
            (0x12f &&& 0xffff, 0x456 &&& 0xfff   ): a4;
            default: accept;
        }
...
```

JSON output for the example (transitions to a1 and a4 are ok, a2 and a3 are not):
```
...
         "transitions" : [
            {
              "value" : "0x00000000012f000000000456",
              "mask" : null,
              "next_state" : "a1"
            },
            {
              "value" : "0x00000000012f000000000456",
              "mask" : "0x00000000000000000000ffff",
              "next_state" : "a2"
            },
            {
              "value" : "0x00000000012f000000000456",
              "mask" : "0x000000000000000000000fff",
              "next_state" : "a3"
            },
            {
              "value" : "0x00000000012f000000000456",
              "mask" : "0x00000000ffff000000000fff",
              "next_state" : "a4"
            },
            {
              "value" : "default",
              "mask" : null,
              "next_state" : null
            }
          ],
...
```

[mask_test.p4.txt](https://github.com/p4lang/p4c/files/1423847/mask_test.p4.txt)

jafingerhut:
Bug should be fixed now that PR #998 has been merged into master branch.

PR #998, opened 2017-10-28
Proposed fix for issue #995

jafingerhut:
Correct the masks generated in bmv2 backend for select expressions
with multiple fields.

### DIFF
--- a/backends/bmv2/parser.cpp
+++ b/backends/bmv2/parser.cpp
@@ -232,15 +232,18 @@ unsigned ParserConverter::combine(const IR::Expression* keySet,
             int width = type->width_bits();
             BUG_CHECK(width > 0, "%1%: unknown width", e);
 
-            mpz_class key_value, mask_value;
+            mpz_class key_value, mask_value, mask_value2;
             convertSimpleKey(keyElement, key_value, mask_value);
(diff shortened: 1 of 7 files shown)


backends/bmv2/parser.cpp
```diff
@@ -232,15 +232,18 @@ unsigned ParserConverter::combine(const IR::Expression* keySet,
             int width = type->width_bits();
             BUG_CHECK(width > 0, "%1%: unknown width", e);
 
-            mpz_class key_value, mask_value;
+            mpz_class key_value, mask_value, mask_value2;
```
cc10512: Please comment this code. It is not immediately obvious what mask_value2 means, unless one refers back to the issue.

jafingerhut: The extra local variable was not strictly necessary, so I removed it, as well as adding a comment explaining the new case.

jafingerhut:
@mbudiu-vmw I added a P4_16 program with STF test case that fails with the current p4c-bm2-ss compiler, but passes with the proposed fix.

jafingerhut:
@mbudiu-vmw I have added all files I saw created in the p4_16_samples_outputs directory for issue995-bmv2.p4.

PR #1003, opened 2017-11-01
list expression fixes

ChrisDodd:
This adds support for list expression constants in the front-end, for use with externs that need them in their constructors.  There's an issue with TypeInferencing around constructor arguments where it puts the wrong type in the typeMap when dealing with arguments that need to be inferenced (see the FIXME) that I'm not sure how to fix with the current typemap setup.

### DIFF
--- a/frontends/p4/evaluator/evaluator.h
+++ b/frontends/p4/evaluator/evaluator.h
@@ -67,6 +67,9 @@ class Evaluator final : public Inspector, public IHasBlock {
     { setValue(expression, expression); return false; }
     bool preorder(const IR::BoolLiteral* expression) override
     { setValue(expression, expression); return false; }
+    bool preorder(const IR::ListLiteral* expression) override
+    { setValue(expression, expression); return false; }
+    bool preorder(const IR::ListExpression* expression) override;
     bool preorder(const IR::Declaration_ID* expression) override
     { setValue(expression, expression); return false; }
 
--- a/frontends/p4/typeChecking/typeChecker.cpp
+++ b/frontends/p4/typeChecking/typeChecker.cpp
@@ -38,6 +38,11 @@ class ConstantTypeSubstitution : public Transform {
     { CHECK_NULL(subst); CHECK_NULL(typeMap); setName("ConstantTypeSubstitution"); }
     // This is needed to handle newly created expressions because
     // their children have changed.
+    // FIXME -- unfortunately this may also be wrong since the change in the
+    // FIXME -- child might change the type of this expression, which means
+    // FIXME -- blindly copying to old type puts the wrong type into the typeMap
+    // FIXME -- we really need to rerun inferencing for the sub-expression at
+    // FIXME -- this point.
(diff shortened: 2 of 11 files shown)


mbudiu-vmw:
Can you check-in an example that needs this feature too?
I will think about it.

frontends/p4/typeChecking/typeChecker.cpp
```diff
@@ -781,7 +786,11 @@ TypeInference::checkExternConstructor(const IR::Node* errorPosition,
         ConstantTypeSubstitution cts(tvs, typeMap);
         auto newArg = cts.convert(arg);
         result->push_back(newArg);
-        setType(newArg, paramType);
+        // FIXME -- ConstantTypeSubstitution will have already put a type in the typeMap
+        // FIXME -- for newArg, which might be wrong, so setting it correctly here will
+        // FIXME -- hit the BUG_CHECK in typeMap.cpp:70.  So we don't set it and hope it
+        // FIXME -- gets reset properly next time we run typeInference.
```
cc10512: And if that does not happen, does the compilation fail?

frontends/p4/evaluator/evaluator.h
```diff
@@ -67,6 +67,9 @@ class Evaluator final : public Inspector, public IHasBlock {
     { setValue(expression, expression); return false; }
     bool preorder(const IR::BoolLiteral* expression) override
     { setValue(expression, expression); return false; }
+    bool preorder(const IR::ListLiteral* expression) override
+    { setValue(expression, expression); return false; }
```
cc10512: I'm sure you have a good reason, but do you mind explaining why do we need to re-evaluate an ListLiteral? A comment will help, especially since we are just setting expression to expression.

ChrisDodd:
Added a trivial motivating example.  More complex uses are possible.

frontends/p4/evaluator/evaluator.h
ChrisDodd: Actually the point her is to NOT reevaluate the ListLiteral -- ListLiteral is a subclass of ListExpression, so if we did not define this preorder function (which does nothing, just adds the literal to the map as itself), the preorder(ListExpression) function would be called to create a bew ListLiteral.

mbudiu-vmw:
I have filed issue #1006; let me fix that one and then we can revisit this PR.
Indeed, the list expression is type-checked, and then unification tries to change its type, which causes a BUG.

mbudiu-vmw:
I will submit a separate PR (if all tests pass) which fixes this issue and also adds the ListLiteral class. I actually named it ListCompileTimeValue, because it can contain not just literals, but also other compile-time values, such as constructor invocations.

Indeed, @ChrisDodd , the solution is to "learn" the new type. This process, of learning new type objects that may be generated during type-checking, is done in many places by the type-checker by calling another TypeChecker recursively on the new objects but passing the same typeMap.

mbudiu-vmw:
I ran into some problems with my PR. It may need to be delayed a bit...

PR #1004, opened 2017-11-02
add an option -s to p4c driver to only run assembler

hanw:


tools/driver/p4c_src/main.py
```diff
@@ -133,6 +133,9 @@ def main():
     parser.add_argument("-S", dest="run_till_assembler",
                         help="Only run the preprocess and compilation steps",
                         action="store_true", default=False)
+    parser.add_argument("-s", dest="skip_compiler",
```
cc10512: How exactly is this supposed to work? How does one specify the input file(s)? Is it expecting the assembly files as inputs? If not, if the compiler generates multiple output files from a source, is the expectation that the names are kept consistent between runs? What if the previous invocation used --output to generate a different file?

I can see how this can be useful to allow you to select specific components to run, but seems a bit dangerous to expose in the current incarnation. Is there a better solution?

hanw: @amreshk and @sethfowler can chime in on why they think this is useful. My understanding is they sometimes only need to rerun the assembler + linker to debug issues with assembler. Compiling the p4 source code is unnecessary for their use case. In addition, having this option completes all the possible combinations to run preprocessor, compiler, assembler, and linker.

ChrisDodd: Analogy with 'normal' toolchains would suggest using the file name rather than an option for this (eg `p4c file.p4` run compiler+assembler+linker, while `p4c file.bfa` runs just the assembler and linker)

amreshk: @ChrisDodd 's suggestion is fine too as long as you have some way to doing this directly through the script. Ideally would be great to have support to,
1. Run just linker + test (allows modifying cfg regs manually)
2. Just the test without any compiler, linker (allows modifying context json manually)
3. gdb directly into the assembler (similar to `-g` for compiler)

sethfowler: For me the benefit of this feature would be mostly for debugging issues with the compiler. Right now if you want to check if some assembly change fixes a bug, the easiest approach is to hack the compiler to generate it. Being able to just edit the assembly directly without having to run all the tooling manually would allow for faster iteration.

ChrisDodd: My usual approach after hacking asm code is to (manually) rerun the assembler and linker and then manaually run the resulting binary...

sethfowler: That certainly works, but it's convenient to be able to use the driver. In fact, ideally we would support this in the test system as well; I often want to rerun a PTF test after changing the assembly, for example. That'd be straightforward to support (e.g. via an environment variable) if we have support in the driver.

I wouldn't be opposed to supporting this *only* via environment variable if that would help this PR get approved, really.

hanw: @cc10512 mentioned he is going to look into rework this PR along with some other stuff. I will let Calin decide what to do with this PR.

sethfowler: SGTM.

cc10512:
replaced by #1077

PR #1005, opened 2017-11-02
fixes includes and class definitions for non-unified build

cc10512:
The GraphAttributeSetter is used also in both controls.cpp and parser.cpp and thus it needs to be fully defined -- the forward declaration in graphs.h is not sufficient.

Also, the typeMap and referenceMap headers need to be explicitly included.

### DIFF
--- a/backends/graphs/graphs.cpp
+++ b/backends/graphs/graphs.cpp
@@ -25,74 +25,6 @@ limitations under the License.
 
 namespace graphs {
 
-class Graphs::GraphAttributeSetter {
- public:
-  void operator()(Graph &g) const {
-      auto vertices = boost::vertices(g);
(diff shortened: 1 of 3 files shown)


mbudiu-vmw:
@cc10512 @antoninbas : the Travis build failed with lots of errors like: simple_switch: error while loading shared libraries: libboost_filesystem.so.1.58.0: cannot open shared object file: No such file or directory
So there is a problem starting the simple_switch simulator.

antoninbas:
@mbudiu-vmw this is independent of this PR. I added a runtime dependency on this library in bmv2 yesterday, and I forgot to update the Dockerfile to reflect this. This should be fixed once this PR is merged: https://github.com/p4lang/behavioral-model/pull/459

PR #1007, opened 2017-11-02
Minor fix of p4c-graph.

Stumble:
1. Remove redundent __start__ and __exit__ nodes.
2. Change the logics of processing select expression, default keyset and rejects are handled
   correctly.


backends/graphs/parser.cpp
```diff
@@ -149,16 +149,21 @@ bool ParserGraphs::preorder(const IR::PackageBlock *block) {
             pGraph g_;
```
antoninbas: shouldn't you be clearing `acceptParents` and `rejectParents` at each iteration of the for loop?
It seems that if you have a P4 architecture with several parser blocks, you would not be resetting these data structures properly between each iteration.

Stumble: agree, fixed.

Issue #983, opened 2017-10-23
Wrong result for bit-wise negating then widening a calculation


jafingerhut:
There might be related problems that I haven't discovered yet, but I have found some issues where if you do bit-wise negation on a small bit vector, then assign it or use it in a calculation that casts it to a wider value, like this:

```
     bit<16> tmp = ~foo;      // e.g. suppose that foo is 16w0x0800, so tmp should become 16w0xf7ff
     bit<32> x32 = (bit<32>) tmp;   // x32 should become 32w0xf7ff, extending with 0 bits in MSBs
```
the generated JSON file for simple_switch is the same as that generated if you first cast the small bit vector to a wide one, then negate it.
```
    bit<32> y32 = ~((bit<32>) tmp);   // y32 should become 32w0xfffff7ff
```
with a recent version of p4c-bm2-ss and simple_switch, I see x32 become 32w0xfffff7ff, the same as y32 becomes.

I will attach a sample program here soon.

jafingerhut:
[issue983.p4.txt](https://github.com/p4lang/p4c/files/1407875/issue983.p4.txt)

I may also create an STF file to exercise this program, but the basic idea is that if you send in a packet with just an Ethernet header with etherType 0x0800, if everything is working as I understand the P4_16 spec, it should output a packet with an ethernet dstAddress that is all 0, indicating no errors.  Currently it outputs a non-0 dstAddress, indicating several unexpected results for arithmetic calculations.

jafingerhut:
I am not sure, but for one of the 'error codes' this program generates (the one for the last 'if' statement in the ingress control block), it might be due to a bug in simple_switch.

antoninbas:
I'm guessing that's because bmv2 uses libgmp, which assumes arbitrary size.
When doing a one's complement, the result is extended with infinite 1s. For example, the complement of 0000...0011 is ...1111...1100. This means that a masking operation has to be performed right after the one's complement operation to guarantee correctness.

jnfoster:
...but the compiler front-end passes are careful to insert the correct
coercions during type checking. So I'm not sure.

-N

jafingerhut:
[issue983.zip](https://github.com/p4lang/p4c/files/1427789/issue983.zip)

[...]

PR #1011, opened 2017-11-06
Fix for issue #983

mbudiu-vmw:
The test cases were provided by @jafingerhut


jafingerhut:
I don't know enough to review the code changes made thoroughly, but I have tried out the test program that is part of the commit, and another that helped me discover the issue in the first place, and they both appear to be working as expected with these changes.

Issue #955, opened 2017-10-03
compiler bug during RewriteIsValidCalls pass


antoninbas:
I get the following compiler bug:
```
SynthesizeValidField invoking RewriteIsValidCalls
In file: /home/antonin/Documents/p4lang/p4c/ir/visitor.cpp:457
Compiler Bug: ../testdata/p4_16_samples/uninit.p4(43): visitor returned non-MethodCallExpression type: stack[1].$valid$ == 1;
        b = stack[1].isValid();
```
when writing a program like this one:
```
control IngressI(inout H hdr, inout M meta, inout standard_metadata_t smeta) {
    apply { bool b = hdr.h.isValid(); }
};
```
I'm attaching the full P4 program which can be used to reproduce the issue.
[bug_RewriteIsValidCalls.p4.txt](https://github.com/p4lang/p4c/files/1353532/bug_RewriteIsValidCalls.p4.txt)

mbudiu-vmw:
@sethfowler : this bug is caused by the pass which synthesizes valid field bits.
I think we should remove this pass from the mid-end. 
This pass breaks some invariants assumed by many other passes about the meaning of headers; the valid field of headers is implicit. Other passes still assume that the field is implicit, and they will treat the explicit field as a normal field.

mbudiu-vmw:
I had a chat with @sethfowler and we think we have a way to fix the 3 bugs related to the valid bits synthesis. Stay tuned.

PR #1014, opened 2017-11-08
Fix for issue #841 and #955

mbudiu-vmw:
This PR fixes these two issues by removing the SynthesizeValidBits pass from the normal compiler flow.
That pass breaks invariants about the semantics of the IR, and does not seem to be necessary anymore.

I have also changed the policies used when to simplify table keys: they were using double negation, now they are easier to read and to combine.


mbudiu-vmw:
@sethfowler : I will wait for you to look at this before I merge it.

mbudiu-vmw:
@sethfowler : I will wait for you to look at this before I merge it.

mbudiu-vmw:
@sethfowler : I will wait for you to look at this before I merge it.

Issue #1020, opened 2017-11-08
SimplifyKey midend pass doesn't handle table references in assignments


AlexBernstein:
Statements like the one below are ignored when applying insertions
`hit = table.apply().hit;`

AlexBernstein:
[suggested fix ](https://github.com/p4lang/p4c/pull/1021)

mbudiu-vmw:
I hope everyone is happy about this solution.

PR #1021, opened 2017-11-08
Fix for issue #1020

AlexBernstein:
Statements like the one below are ignored in SimplifyKey midend pass when applying insertions
`hit = table.apply().hit;`


ChrisDodd:
I think this is correct, but it would be nice to have a testcase.

mbudiu-vmw:
Thank you. This is nice. The protocol is to run 'make check' and then submit all the files that have appeared in the 'testcase' folder. These are reference outputs that make sure that the compiler always does the same thing.

AlexBernstein:
Added a test case by modifying existing key-ebpb.p4 test to make key complex and force the move. in the midend output complex keys in the table definition get replaced by key_0 and key_1, and declarations are added before the table. However, assignments
```
key_0 = headers.ipv4.srcAddr + 1;
key_1 = headers.ipv4.dstAddr +1; 
```
are missing. They should appear before the `tmp_0 = t.apply().hit;`

AlexBernstein:
Added a test case by modifying existing key-ebpf.p4 test to make key complex and force the move. in the midend output complex keys in the table definition get replaced by key_0 and key_1, and declarations are added before the table. However, assignments
```
key_0 = headers.ipv4.srcAddr + 1;
key_1 = headers.ipv4.dstAddr + 1; 
```
are missing. They should appear before the `tmp_0 = t.apply().hit;`

mbudiu-vmw:
Yes, that is exactly what the reference outputs would be conveying.

AlexBernstein:
Sorry, first time submitting. I've run the 'make check', but having trouble finding the folder with the generated output. Is it somewhere in the 'build' directory?

mbudiu-vmw:
run 'cmake ..', then 'make check -j4', and then 'git status'
You will see some new p4 files that you should commit.

mbudiu-vmw:
Thank you. If the Travis build completes successfully we'll merge this.

Issue #737, opened 2017-06-23
Investigate why casting bit<1> fields to bool breaks BMV2 STF tests


sethfowler:
This issue was noticed [here](https://github.com/p4lang/p4c/pull/733#discussion_r123853884). I haven't investigated further. To reproduce it, it's sufficient to edit `synthesizeValidField.cpp` and in `RewriteIsValidCalls::postorder(IR::MethodCallExpression* expr)`, replace these lines:

```c++
    auto constant = new IR::Constant(IR::Type::Bits::get(1), 1);
    return new IR::Equ(expr->srcInfo, member, constant);
```

With this line:

```c++
    return new IR::Cast(expr->srcInfo, IR::Type::Boolean::get(), member);
```

A log file of the resulting test failures: [test-suite.log.txt](https://github.com/p4lang/p4c/files/1099171/test-suite.log.txt)

Xingdao-Fan:
Not sure if related, but the simple switch itself seems to be misbehaving when the bool type is involved.

See [here](https://github.com/p4lang/behavioral-model/issues/399).

sethfowler:
p4lang/behavioral-model#399 was closed; this problem is believed to be the compiler's fault. (We're not generating a necessary cast.)

sethfowler:
I'm sure this is the same problem, though. Thanks for providing a test case that demonstrates it!

hanw:
Although #751 only fixed the cast from bool -> bit<1>, not the other way around.

mbudiu-vmw:
@sethfowler : do we have a test case that fails?
This issue is very vague. Unless we can get a test case I am tempted to close it.

jafingerhut:
In general, I believe all `bit<W>` and `bool` _operations_ in the bmv2 JSON operate on different 'stacks' of values, e.g. "and" and "or" operators pop two values from a boolean data stack, and push a value back on the boolean data stack, but bitwise "&" and "|" pops two values from a bitvector data stack, and pushes a value back on the bitvector data stack.

bmv2 does _no_ error checking on the expressions it reads in the JSON file to ensure that the operands of an operator have the expected type (AFAIK).

p4c generating JSON with incorrect combinations of operator/operand types was the root cause for this issue: https://github.com/p4lang/p4c/issues/950

I wouldn't be surprised if it was the root cause for other issues involving expressions with boolean types.

It probably wouldn't be terribly difficult to write a checker for bmv2 JSON expressions to statically check for wrong nesting of operators.

[...]

PR #1022, opened 2017-11-08
Test case for issue #737, which seems to be fixed

mbudiu-vmw:
Test case adapted from https://github.com/p4lang/behavioral-model/issues/399


jafingerhut:
FYI, I haven't tested to make sure, but wouldn't be surprised if the fix for https://github.com/p4lang/p4c/issues/950 fixed this one, too.

PR #1024, opened 2017-11-08
Experimental: parser value set implemented as a first class type

hanw:
The PR implements an experimental feature to use parser value set as a first class type. 

The implemented parser value set type is:
```
value_set<T> pvs;
```

The name of a parser value set can be used as the select case in a parser state, just as we had in P4-14.

```
value_set<bit<16>> pvs0_0;
value_set<bit<16>> pvs1_0;
state parse_ethernet {
   packet.extract<ethernet_t>(hdr.ethernet);
       pvs0_0: accept;
       pvs1_0: parse_inner_ethernet;
       default: accept;
   }
}
```

It is up to the language design group to decide the merit of this approach. But at least we have a working prototype for the discussion.

### DIFF
--- a/backends/bmv2/parser.cpp
+++ b/backends/bmv2/parser.cpp
@@ -200,7 +200,8 @@ void ParserConverter::convertSimpleKey(const IR::Expression* keySet,
 
 unsigned ParserConverter::combine(const IR::Expression* keySet,
                                 const IR::ListExpression* select,
-                                mpz_class& value, mpz_class& mask) const {
+                                mpz_class& value, mpz_class& mask,
+                                bool& is_vset, cstring& vset_name) const {
     // From the BMv2 spec: For values and masks, make sure that you
     // use the correct format. They need to be the concatenation (in
     // the right order) of all byte padded fields (padded with 0
@@ -212,6 +213,7 @@ unsigned ParserConverter::combine(const IR::Expression* keySet,
     // Return width in bytes
     value = 0;
     mask = 0;
+    is_vset = false;
     unsigned totalWidth = 0;
     if (keySet->is<IR::DefaultExpression>()) {
         return totalWidth;
@@ -249,12 +251,16 @@ unsigned ParserConverter::combine(const IR::Expression* keySet,
             }
             mask = Util::shift_left(mask, w) + mask_value;
             LOG3("Shifting " << " into key " << key_value << " &&& " << mask_value <<
-                 " result is " << value << " &&& " << mask);
+                             " result is " << value << " &&& " << mask);
             index++;
         }
 
         if (noMask)
             mask = -1;
+        return totalWidth;
+    } else if (keySet->is<IR::PathExpression>()) {
+        is_vset = true;
+        vset_name = keySet->to<IR::PathExpression>()->path->name;
         return totalWidth;
     } else {
         BUG_CHECK(select->components.size() == 1, "%1%: mismatched select/label", select);
@@ -282,18 +288,26 @@ ParserConverter::convertSelectExpression(const IR::SelectExpression* expr) {
     for (auto sc : se->selectCases) {
         auto trans = new Util::JsonObject();
         mpz_class value, mask;
-        unsigned bytes = combine(sc->keyset, se->select, value, mask);
-        if (mask == 0) {
-            trans->emplace("value", "default");
-            trans->emplace("mask", Util::JsonValue::null);
+        bool is_vset;
+        cstring vset_name;
+        unsigned bytes = combine(sc->keyset, se->select, value, mask, is_vset, vset_name);
+        if (is_vset) {
+            trans->emplace("value", vset_name);
+            trans->emplace("mask", mask);
             trans->emplace("next_state", stateName(sc->state->path->name));
         } else {
-            trans->emplace("value", stringRepr(value, bytes));
-            if (mask == -1)
+            if (mask == 0) {
+                trans->emplace("value", "default");
                 trans->emplace("mask", Util::JsonValue::null);
-            else
-                trans->emplace("mask", stringRepr(mask, bytes));
-            trans->emplace("next_state", stateName(sc->state->path->name));
+                trans->emplace("next_state", stateName(sc->state->path->name));
+            } else {
+                trans->emplace("value", stringRepr(value, bytes));
+                if (mask == -1)
+                    trans->emplace("mask", Util::JsonValue::null);
+                else
+                    trans->emplace("mask", stringRepr(mask, bytes));
+                trans->emplace("next_state", stateName(sc->state->path->name));
+            }
         }
         result.push_back(trans);
     }
@@ -331,7 +345,13 @@ bool ParserConverter::preorder(const IR::P4Parser* parser) {
     auto parser_id = json->add_parser("parser");
 
     for (auto s : parser->parserLocals) {
-        if (s->is<IR::Declaration_Instance>()) {
+        if (auto inst = s->to<IR::Declaration_Instance>()) {
+            if (auto typeName = inst->type->to<IR::Type_Name>()) {
+                if (typeName->path->name == "value_set") {
+                    json->add_parse_vset(s->name);
+                    continue;
+                }
+            }
             ::error("%1%: not supported on parsers on this target", s);
             return false;
         }
--- a/backends/bmv2/parser.cpp
+++ b/backends/bmv2/parser.cpp
@@ -292,6 +292,7 @@ ParserConverter::convertSelectExpression(const IR::SelectExpression* expr) {
         cstring vset_name;
         unsigned bytes = combine(sc->keyset, se->select, value, mask, is_vset, vset_name);
         if (is_vset) {
+            trans->emplace("type", "parse_vset");
             trans->emplace("value", vset_name);
             trans->emplace("mask", mask);
             trans->emplace("next_state", stateName(sc->state->path->name));
@@ -301,6 +302,7 @@ ParserConverter::convertSelectExpression(const IR::SelectExpression* expr) {
                 trans->emplace("mask", Util::JsonValue::null);
                 trans->emplace("next_state", stateName(sc->state->path->name));
             } else {
+                trans->emplace("type", "hexstr");
                 trans->emplace("value", stringRepr(value, bytes));
                 if (mask == -1)
                     trans->emplace("mask", Util::JsonValue::null);
@@ -345,15 +347,18 @@ bool ParserConverter::preorder(const IR::P4Parser* parser) {
     auto parser_id = json->add_parser("parser");
 
     for (auto s : parser->parserLocals) {
-        if (auto inst = s->to<IR::Declaration_Instance>()) {
-            if (auto typeName = inst->type->to<IR::Type_Name>()) {
-                if (typeName->path->name == "value_set") {
-                    json->add_parse_vset(s->name);
-                    continue;
-                }
+        if (auto inst = s->to<IR::Declaration_Variable>()) {
+            if (!inst->type->is<IR::Type_ValueSet>())
+                continue;
+            auto value_set = inst->type->to<IR::Type_ValueSet>();
+            unsigned size = 0;
+            if (value_set->elementType->is<IR::Type_Name>()) {
+                auto type = typeMap->getTypeType(value_set->elementType, true);
+                size = type->width_bits();
+            } else {
+                size = value_set->elementType->width_bits();
             }
-            ::error("%1%: not supported on parsers on this target", s);
-            return false;
+            json->add_parse_vset(s->name, size);
         }
     }
 
--- a/frontends/p4/fromv1.0/programStructure.cpp
+++ b/frontends/p4/fromv1.0/programStructure.cpp
@@ -41,6 +41,7 @@ ProgramStructure::ProgramStructure() :
         actions(&allNames), counters(&allNames), registers(&allNames), meters(&allNames),
         action_profiles(nullptr), field_lists(nullptr), field_list_calculations(&allNames),
         action_selectors(nullptr), extern_types(&allNames), externs(&allNames),
+        value_sets(&allNames),
         calledActions("actions"), calledControls("controls"), calledCounters("counters"),
         calledMeters("meters"), calledRegisters("registers"), calledExterns("externs"),
         parsers("parsers"), parserPacketIn(nullptr), parserHeadersOut(nullptr),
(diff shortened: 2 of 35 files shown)


hanw:
@jafingerhut See the above extern proposal for parser value set, if accepted, we should update PSA to be the same.

backends/bmv2/parser.cpp
```diff
@@ -286,18 +292,26 @@ ParserConverter::convertSelectExpression(const IR::SelectExpression* expr) {
...
+        cstring vset_name;
+        unsigned bytes = combine(sc->keyset, se->select, value, mask, is_vset, vset_name);
+        if (is_vset) {
+            trans->emplace("value", vset_name);
+            trans->emplace("mask", mask);
             trans->emplace("next_state", stateName(sc->state->path->name));
         } else {
-            trans->emplace("value", stringRepr(value, bytes));
-            if (mask == -1)
+            if (mask == 0) {
```
ChrisDodd: minor nit:  I feel

   } else if (mask == 0) {

and not increasing the indent would be better.

mbudiu-vmw:
You cannot update PSA until this is part of the P4 spec.  We had lots of discussions about value sets in the P4 design working group. I would like to encourage you to write up a precise description of how this is supposed to work.

frontends/p4/fromv1.0/programStructure.cpp
```diff
@@ -405,14 +413,39 @@ const IR::ParserState* ProgramStructure::convertParser(const IR::V1Parser* parse
...
+                if (!first) continue;
+                auto value_set = value_sets.get(first->path->name);
+                if (!value_set) {
+                    ::error("Unable to find declaration for parser_value_set %s", first->path->name);
+                    return nullptr;
+                }
+
+                auto type = new IR::Type_Name("value_set");
+                unsigned vset_size = 4;  // default value_set size is 4.
+                if (auto size = value_set->getAnnotation("parser_value_set_size")) {
```
ChrisDodd: If this annotation is present, you probably want to remove it, as the size is now a ctor param...

testdata/p4_14_samples/parser_value_set2.p4
```diff
@@ -0,0 +1,77 @@
...
+        dstAddr: 32;
+    }
+}
+
+header ethernet_t ethernet;
+header ethernet_t inner_ethernet;
+header ipv4_t ipv4;
+
+@pragma parser_value_set_size 4
+parser_value_set pvs0;
```
ChrisDodd: Should probably have a testcase with a `pragma parser_value_set_size` other than the default size to make sure its getting set up properly...

frontends/p4/typeChecking/typeChecker.cpp
```diff
@@ -2746,6 +2746,11 @@ TypeInference::matchCase(const IR::SelectExpression* select, const IR::Type_Tupl
     if (caseType->is<IR::Type_Dontcare>())
         return selectCase;
 
+    // The caseType may be an extern, we use the type of the selectType
+    if (caseType->is<IR::Type_Extern>()) {
```
mbudiu-vmw: This is not really type checking.
Any extern would work here.

midend/simplifySelectCases.cpp
```diff
@@ -42,6 +42,15 @@ void DoSimplifySelectCases::checkSimpleConstant(const IR::Expression* expr) cons
     auto ei = EnumInstance::resolve(expr, typeMap);
     if (ei != nullptr)
         return;
+
+    // we allow value_set name to be used in place of select case;
+    if (expr->is<IR::PathExpression>()) {
```
mbudiu-vmw: You cannot hardwire the name of an extern block in the mid-end!
The name of this block is architecture-specific!

mbudiu-vmw:
I am sorry, but this is a hack.
I cannot accept this PR.
The whole point of P4-16 was to eliminate hacks from the language design.
This is moving back towards the same approach as P4-14.

cc10512:
@mbudiu-vmw: Accepting that P4_14 did some things right is the path to successful adoption of P4_16.

hanw:
@mbudiu-vmw I already tried to implement bmv2 support for other options discussed in the working group.

```
extern value_set<D> {
    value_set(bit<8> size);
    bool is_member();
}
```

Unfortunately, the previous proposal by the language group is not good in terms of usability. 

1. Users are forced to break the parser state into multiple parser states for no good reason. 

E.g. with the existing proposal, we need to write the following, it is not clear all these typing on dispatch state gain anything to the user. They are essentially typing out the internals of parser_value_set which should be encapsulated behind the extern abstraction.
```
parser TopParser(packet_in b, out Parsed_packet p) {
    ValueSet<bit16>>(1) trill_types;
    ValueSet<bit<16>>(4) tpid_types;

    state start {
        b.extract(p.ethernet);
        transition select(p.ethernet.etherType) {
            16w0x0800: parse_ipv4;
            16w0x0806: parse_arp;
            16w0x86DD: parse_ipv6;
            default: dispatch_tpid_value_set;
        }
    }
    state dispatch_tpid_value_set {
        bit<1> is_tpid = ethtype_kinds.is_member(p.ethernet.etherType);
        transition select(is_tpid) {
            1: parse_vlan_tag;
            default /* or 0 */: dispatch_trill_value_set;
        }

    state dispatch_trill_value_set {
        bit<1> is_trill = ethtype_kinds.is_member(p.ethernet.etherType);
        transition select(is_trill) {
            1: parse_trill;
            default /* or 0 */: accept;
        }
    }
}
```

With the proposal in this PR.
```
parser TopParser(packet_in b, out Parsed_packet p) {
    ValueSet(1) trill_types;
    ValueSet(4) tpid_types;

    state start {
        b.extract(p.ethernet);
        transition select(p.ethernet.etherType) {
            16w0x0800: parse_ipv4;
            16w0x0806: parse_arp;
            16w0x86DD: parse_ipv6;
            tpid_types: parser_tpid_types;
            trill_types: parse_trill_types;
        }
    }
```

2. support parser_value_set in bmv2 backend is difficult because bmv2 is modeled after P4-14. So p4c has to translate p4-14 program to p4-16 extern, then back to the p4-14 primitive.

3. the compiler backend on hardware is a lot harder to implement and optimize with the existing proposal.

midend/simplifySelectCases.cpp
hanw: The spec says "Some targets may require that all keyset expressions in a select expression be compile-time known values." In this case, the mid-end should not check the selection cases are all compile-time constant, because it is target-specific, thus should be done in the backend.

mbudiu-vmw:
@cc10512 : if I didn't believe this statement I would not have spent 3 years of my life working on this subject. What I meant is not a criticism of P4-14, but of the way P4-14 was evolved, by adding frequently completely new constructs to the language.

@hanw :
The usability of a language does not only consist in having programs be as short as possible (as anyone who tried to read a perl program can confirm). It consists in having the language be built of few basic concepts which are independent and work in any combination, without too many rules and exceptions. Moreover, experience teaches us that when we invent a new language concept we are probably wrong; we should try to build as much as possible in terms of well-understood concepts, such as types, control-flow, objects, lexical scoping, and so on.

The language group does not really have a proposal yet for value sets; we considered two proposals: one based on the is_member method, which can be done entirely in P4-16 as is now, and one where sets are first-class values, which resembles your proposal to some degree. I even had a PR for the latter, which I then withdrew, because I realized that it opens lots of difficult problems that we don't have an answer for (for example, what other operations are available between values that are sets?)

Let's not conflate the problem of user-facing usability and compiler-writing difficulty; these are different.
I claim that users can write code which isn't too bad:

```
transition select(set1.is_member(a, b), set2.is_member(x, y, z)) {
   case true, _: s1;
   case false, true: s2;
   default: s3;
}
```

I am very sympathetic to the problem of generating code in your back-end, but I am not sure that the way you wrote the program above, with several states, won't actually be easier. We can talk about details of your target. Indeed, you should not expect heroic efforts from compiler writers to generate code for a target. I don't think that generating code for bmv2 will be difficult starting from the is_member method.

However, you should also not leak your target architecture in the language; this will prohibit portability, inhibit the evolution of the language, and of your target architecture.

You have to agree that the implementation is a hack; you are essentially special-casing these constructs all thorough the front-end and mid-end. This is not a recipe for writing maintainable software.

I also understand that you may need to showcase this feature sooner rather than later. My proposal is to move this feature to a branch, and to discuss how to design this properly at the next language working group meeting, in a couple of weeks. There may be a solution which will allow you to reuse most of the existing code. If the solution requires extending P4, then we can hopefully adopt it using an experimental label, as we have done for other solutions, which would allow us to merge the branch into the compiler. I am open to discuss this in a conference call sonner too, so that we can have a proposal by the time of the working group.

midend/simplifySelectCases.cpp
mbudiu-vmw: You cannot change the midend files to depend on specifics of a particular architecture; you can change the midend.cpp file in the bmv2 back-end, which assembles the mid-end files that it needs.
If the mid-end has architecture-specific choices, these should be encapsulated in an abstract policy method which is defined in the back-end.

ChrisDodd: The right approach is probably to have a policy object that encapsulates the set of extern types that can appear in a select case.  The backend would create a policy containing value_set and provide it to this pass.  Bckends without value_set or anything similar would provide an empty policy.

jnfoster:
I agree with @mbudiu-vmw that this needs to be discussed with the LDWG (there has already been much discussion) and having gone to the trouble of generalizing the language with externs, would like to avoid one-off hacks. 

I also agree with @cc10512 that we should look for ways to make the ergonomics of the language attractive, which could mean adding or resurrecting domain-specific features.

So... more discussion is needed.

In the meantime, having a prototype implementation of a proposed extension is not a problem -- in fact, that's an important part of the process we want to follow when we evolve the language! However, it should be clearly identified as an experimental feature so people don't start unwittingly depending on it until the LDWG and PSAWG reach consensus.

frontends/p4/typeChecking/typeChecker.cpp
```diff
@@ -2748,6 +2762,10 @@ TypeInference::matchCase(const IR::SelectExpression* select, const IR::Type_Tupl
     // The caseType may be a simple type, and then we have to unwrap the selectType
     if (caseType->is<IR::Type_Dontcare>())
         return selectCase;
+    // The caseType may be an value set, we use the type of the selectType
```
mbudiu-vmw: This looks wrong. You should handle it exactly like a tuple is handled below.
Just extract the type argument of the value_set and use that to match the select case type.

jafingerhut:
My apologies if there is a better issue on which to discuss this - let me know where and I'll ask there instead.

In looking at the P4_16 language spec, it seems that in this `select` expression:
```
bit<16> etherType;

// ...

transition select (etherType) {
0x0800: parser_ipv4;
0x86dd: parse_ipv6;
my_parser_value_set: parse_my_custom_header;
// ...
default: accept;
}
```
The type of the expression "0x0800" is "a set of bit<16>".  The same is true for the expressions "0x86dd" and "default".  Those expressions all have constant compile-time values.

Is the thinking here that we could define a P4_16 extern with an instance named "my_parser_value_set" whose return type is also "a set of bit<16>", but the control plane could change that set at run time (via control plane operations)?

If so, awesome.  If it helps, PSA could easily state something like "a PSA implementation need not implement the use of such an extern's return value anywhere _except_ in a parser select expression."

The only other place that the P4_16 v1.0 spec says an expression of a set type can be used is in a constant table entry (according to section 8.12 "Operations on sets").  We already have tables with run-time variable entries, so I can think of no desire to want to introduce a new way to have run-time variable table entries using such an extern.

mbudiu-vmw:
This is what this PR does. But it's not an extern, it's a first-class object in the language.
BTW: @hanw : I think that the type of a value_set<T> object should be set<T>, then you don't need to do anything special about it in the implementation.

hanw:
@mbudiu-vmw If we make the type of `value_set<T>` to be `set<T>`, then when we dump IR to P4 code, aren't we going to print all IR nodes with type `set<T>` to set, instead of value_set?

hanw:
@jnfoster can I get a thumb-up for this?

mbudiu-vmw:
No, the type in the typeMap is a `set`, but the object is a `value_set`. 
This is no different from a typedef.

mbudiu-vmw:
I think @jnfoster wanted a compiler flag. But I think it's not necessary, since this does not impact other programs at all.

frontends/p4/fromv1.0/programStructure.h
```diff
@@ -19,6 +19,7 @@ limitations under the License.
 
 #include <set>
 #include <vector>
+#include <deque>
```
mbudiu-vmw: is this needed?

frontends/p4/typeChecking/typeChecker.cpp
```diff
@@ -1144,6 +1153,11 @@ const IR::Node* TypeInference::postorder(IR::Type_Set* type) {
     return type;
 }
 
+const IR::Node* TypeInference::postorder(IR::Type_ValueSet* type) {
+    (void)setTypeType(type);
```
mbudiu-vmw: Can you try to set this to `Type_Set(elementType)` and see if it works?
The you won't need to do anything in the select expression.
This will be a funny type declaration, because the type of the declaration is not a TypeType of itself.
Also, the case in canonicalize won't be necessary.

hanw: Like this? 
```
const IR::Node* TypeInference::postorder(IR::Type_ValueSet* type) {
    auto tt = new IR::Type_Set(type->elementType);
    (void)setTypeType(tt);
    return tt;
}
```

mbudiu-vmw: This may work.

mbudiu-vmw:
This looks great. I hope the tests pass.
And I hope that @vgurevich is happy.

mbudiu-vmw:
Let me see if I can fix this.

hanw:
Thanks!

mbudiu-vmw:
I fixed some problems.
1) Just calling setTypeType did not work, I had to specialize the function a bit. Also, the method for Value_set should return 'type' and not 'tt.
2) I had to adapt the mid-end pass which does tuple elimination for value_sets: you don't really want to eliminate tuples which are type arguments to value-sets, since these do not participate in any computation.

These are very reasonable changes.
Let me see if the tests succeed.

hanw:
Superceded by #1158

PR #1029, opened 2017-11-09
Fix for issue #530

mbudiu-vmw:
@jafingerhut : if this is good enough maybe we can close this issue.

### DIFF
--- a/backends/p4test/p4test.cpp
+++ b/backends/p4test/p4test.cpp
@@ -35,12 +35,19 @@ limitations under the License.
 class P4TestOptions : public CompilerOptions {
  public:
     bool parseOnly = false;
+    bool validateOnly = false;
     P4TestOptions() {
         registerOption("--parse-only", nullptr,
                        [this](const char*) {
                            parseOnly = true;
                            return true; },
                        "only parse the P4 input, without any further processing", true);
+        registerOption("--validate", nullptr,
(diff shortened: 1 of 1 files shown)


mbudiu-vmw:
This PR adds a "--validate" flag to p4test. If this flag is set only the front-end is executed.
This enables the compiler to just validate that a P4-16 source is correct.
The p4test mid-end has various passes which may fail for legal programs, because this is how they are tested.

jafingerhut:
Thanks for adding this option.  Maybe I am doing something wrong, but when I get the latest p4c source after this commit and rebuild everything, and try to run the command "p4test --validate foo.p4", it gives an error that the option does not exist.  I do not see --validate in the output of "p4test --help", either.  Then again, I don't see --parse-only in the output of "p4test --help", but I can run "p4test --parse-only foo.p4" and it gives no errors.

mbudiu-vmw:
I have filed a new issue for that: https://github.com/p4lang/p4c/issues/1034
It looks like the option works, but it is not printed by the usage() method.

jafingerhut:
And I must have done something wrong somewhere in my previous test, because re-building yet again, I can run with the --validate option as I can with --parse-only option.  Sorry for the noise, and thanks for the issue for adding those options to the help output.

Issue #961, opened 2017-10-04
STF files for several tests seem to use incorrect syntax


mbudiu-vmw:
All are p4_14 programs, named like counter*.p4 and gateway*.p4

jafingerhut:
I tried with recent versions of p4c and behavioral-model repos to run commands like this, to run individual STF tests:
```
% cd p4c/build
% ../backends/bmv2/run-bmv2-test.py .. -b -v ../testdata/p4_14_samples/gateway1.p4
```
For the 8 files gateway1.p4 through gateway8.p4 in the p4_14_samples directory, I found no STF syntax errors.  gateway1.p4 through gateway4.p4 ran STF tests that failed, because there were different number of actual vs. expected output packets.  Perhaps this was masked before some recent changes that caused STF tests to do more precise checking of actual vs. output packets?

gateway5.p4 through gateway8.p4 had no syntax errors in the STF files, and all tests passed, so no problems seen there.

Perhaps for gateway1.p4 through gateway4.p4, the best thing to do is to update the STF files to have the correct sequence of expected output packets, checking the new ones by hand to verify they are correct?

jafingerhut:
I also tried the command mentioned in the previous comment for p4_14_samples/counter1.p4 through counter5.p4.

All of them had STF syntax errors for lines like these:
```
wait
check_counter ...
```
Those look like commands intended to read counters and verify that they had particular values.  Those commands are not implemented in STF, but it appears that a syntax error on any line causes STF tests to just report them as a syntax error, and continue reading more lines looking for correct ones, so no real harm done other than maybe making people think these tests cover more than they actually do, and that those STF commands are implemented.

One option would be to enhance STF tests to implement those commands, but my impression is that the trend for such testing of other APIs like reading counters might be towards using PTF rather than STF, or whatever it is that P4 Runtime API testing is doing?

All of the tests for counter1.p4 through counter4.p4 failed in my attempts.  Only counter5.p4 passed.

The quickest thing I can think of doing would be to delete the STF lines that are not implemented, eliminating the syntax error warnings, and update the STF files for counter1.p4 through counter4.p4 to pass, probably by updating the expected output packets in the STF files.  Sound like a reasonable action to take?

ChrisDodd:
Some of the tests fail because bmv2 outputs (to port 0) packets that never
have their egress_spec set, instead of dropping them as the test expects...

jafingerhut:
[...]

PR #1030, opened 2017-11-09
adds STF parsing code based on ply

cc10512:
The code implements STF lexing and parsing based on ply.

We also include a base class that can be run to generate tests for
different other frameworks and an example that uses the class to just
evaluate and print STF statements.

Hopefully helps you with #961.


jafingerhut:
I am guessing that none of these files are currently in use by the p4c build system, nor the running of any automated tests.  The hope is that they will be useful for creating STF tests that can exercise more language features, e.g. reading counters and checking their values.

If so, then it seems 'harmless' to add these, and it is a bit difficult to know how easy it is to use them to write the proposed enhancements to STF tests without trying it out and seeing if it works.  I'm in favor of adding the files if they can't break anything (at least not until they are used, and then we can fix that use of them).

mbudiu-vmw:
These files should be probably part of the tools folder, since multiple back-ends will use them.
This is a biggish pile of code, I may need a bit of time to review it.

jafingerhut:
So my reading of the current STF test-runner code, and this newly proposed parser to add, is that STF is this special mini-language that requires new code to parse it and translate it into simple_switch_CLI commands, and packet sends and expects.

Why not just write tests in Python with a simple API for sending simple_switch_CLI commands as strings, and provide Python functions to send a packet, and expect packets?  Then you'd get the full power of Python for writing unit tests that interact with simple_switch, e.g. you could have arbitrary variables and Python data structures if you wanted, loops, etc. rather than not having those (unless you wanted to make your STF interpreter become a larger and larger subset of a general purpose programming language).

cc10512:
@jafingerhut the current run_bmv2 shell does issue cli. We also have PTF (https://github.com/p4lang/ptf), that talks thrift and/or p4runtime. So how many times are we going to rewrite the tests? Internally we use this parser to translate to these frameworks. 

The point you are making is valid and STF has limitations, in particular w.r.t. support for more advanced externs. It does have some benefits yoo, so as long as you stick to tables, it is simple, fast, and easily retargetable with this parser code.

jafingerhut:
I guess I could have made it clearer -- there is no need to rewrite any tests.  If you simply provide the option to write general tests in something like I described, there is no need to eliminate the existing STF option -- but there is also no need to generalize it further.

jafingerhut:
Are there existing PTF tests that can be run via some 'make' command that is part of the p4c repository?  From a quick reading of the PTF README, is it true that you have to generate some PD API from your program before you can use PTF?  Or can PTF use something like the simple_switch_CLI commands as strings?

jafingerhut:
Another possibility: See this P4 Runtime API test written in Python (another file in the same proposed commit shows the P4_16 program it is meant to test).  I know it isn't as short as an STF test, but it isn't a whole lot longer, either, and it should enable all counter, meter, action profile, action selector, etc. functionality that simple_switch has.

If we extend STF for those things, we are basically making up a command that is similar to, but different than, the corresponding command in simple_switch_CLI, and the knowledge of how to do that is useful nowhere except when writing STF files.

backends/bmv2/stf_lexer.py
```diff
@@ -0,0 +1,242 @@
...
+    @TOKEN(dec_constant)
+    def t_packetdata_DATA_DEC(self, t):
+        return t
+
+    @TOKEN(hex_constant_body)
+    def t_packetdata_DATA_HEX(self, t):
+        return t
+
+    def t_packetdata_DATA_TERN(self, t):
+        r'\*'
```
mbudiu-vmw: Can someone explain how this is evaluated? I looked at the Ply manual but I still don't understand.

mbudiu-vmw:
@jafingerhut makes a valid point, we have lots of "testing frameworks" for P4.
I regard STF as an internal tool part of the compiler testing tools; it should not be used for anything else.
STF is useful only insofar it is significantly simpler than PTF or raw Python. If it starts to accumulate all features of these other tools then it has exceeded its original purpose.
Let's say that STF should mostly model data-plane features, since it is mostly designed to test what the p4c compilers do, and these compiler only manipulate the data plane part of the program. The other frameworks support very complicated control-plane operations, which are much more target-specific. There is no reason we could not use STF with EBPF programs as well, for example, although there is some feature-creep into STF.

jafingerhut:
For example, soon we will want to extend p4c and bmv2 to add the PSA architecture.

Some of the tests for that, I believe, should include things that can verify that counters are working as expected, and action profiles, action selectors, registers, digests, etc.  simple_switch_CLI has commands to do nearly all of these things today.  P4 Runtime API does, or will soon (I realize it is in an earlier state than simple_switch code is, so may be a too-frequently-changing thing to rely upon at this particular moment in time).

Action profiles and action selectors cannot be tested without a control plane to do it.  Sure, we could make up STF syntax to create commands for doing that, but they already exist in simple_switch_CLI.  I don't know the history here, but if you look at it from the eyes of a relative newcomer (me), it seems like STF has different syntax than simple_switch_CLI, and less functionality, for no observable gain.

mbudiu-vmw:
The CLI is a CLI, it is not useful in itself for testing. You need a way to script invocations of the CLI.
There's also the problem of timing: you don't know how long to wait after sending a packet until the counters are updated. 

Testing an architecture is not the same thing as testing the compiler; STF is not a good tool for that.

jafingerhut:
The issue that we are talking about adds the ability to read counters and check their values against expected values, and it has a 'wait' statement in it that I don't know exactly what it does, but it probably waits some standard time like 1 or 2 seconds, with the expectation that by that time any packets sent before in the STF file should be done processing in bmv2, and the results of side effects like counter updates are ready to check.

If you don't want that, then you don't want these STF enhancements either.  But if you don't want that, then I don't see any other straightforward way to write automated tests for the bmv2 backend of the compiler.  At least, not the kinds of tests that verify that counters, tables, action profiles, action selectors, etc. work as desired.  Do you?

mbudiu-vmw:
The compiler does not do much to externs, so testing these is not part of the core compiler tests that we need.

backends/bmv2/stf_lexer.py
cc10512: if you see a '*' in the string for the packetdata, then the token is a DATA_TERN.

cc10512:
@mbudiu-vmw, it is not clear to me whether you would like the files moved to tools before this gets merged or can that be a future PR when we start using this code.

mbudiu-vmw:
Well, it would be nice to have it in the right place, but we can certainly move it at any time.

cc10512:
@mbudiu-vmw, I granted your wish!

PR #1031, opened 2017-11-10
Fix some of the failing STF tests described in #961

jafingerhut:
These failures all are due to bmv2 sending packets out of port 0 for a
P4_14 program that never assigns a value to the egress_spec intrinsic
metadata field.

Chris Dodd suggested changing the programs so that if they do not
explicitly assign a value to egress_spec, then they drop the packet.
This makes the existing STF tests for these programs pass, so these
tests appear to have been written with this behavior in mind.

### DIFF
--- a/backends/bmv2/CMakeLists.txt
+++ b/backends/bmv2/CMakeLists.txt
@@ -105,10 +105,6 @@ set (XFAIL_TESTS
   testdata/p4_14_samples/counter2.p4
   testdata/p4_14_samples/counter3.p4
   testdata/p4_14_samples/counter4.p4
-  testdata/p4_14_samples/gateway1.p4
-  testdata/p4_14_samples/gateway2.p4
-  testdata/p4_14_samples/gateway3.p4
-  testdata/p4_14_samples/gateway4.p4
   testdata/p4_14_samples/07-MultiProtocol.p4
 
   # This test defines two lpm keys for a table.
--- a/testdata/p4_14_samples/gateway1.p4
+++ b/testdata/p4_14_samples/gateway1.p4
@@ -29,6 +29,15 @@ parser start {
     return ingress;
 }
 
+action _drop() { drop(); }
+
+table set_default_behavior_drop {
+    actions {
+        _drop;
+    }
+    default_action: _drop;
+}
+
 action noop() { }
 action setb1(val, port) {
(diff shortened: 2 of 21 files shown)


mbudiu-vmw:
@jafingerhut : I think that @ChrisDodd should review this pr, since he wrote the original tests.
If he is too busy I can do it too. In the meantime I have assigned him.

ChrisDodd:
It seems like this is not legal P4_14 code, as the spec says `drop` can only be used in the egress pipeline:

> drop()
Summary
Drop the packet on egress.
Description
Indicate that the packet should not be transmitted. This primitive is in-
tended for the egress pipeline where it is the only way to indicate that the
packet should not be transmitted. On the ingress pipeline, this primitive
is equivalent to setting the
egress_spec
metadata to a drop value (spe-
cific to the target).
If  executed  on  the  ingress  pipeline,  the  packet  will  continue  through
the  end  of  the  pipeline.   A  subsequent  table  may  change  the  value  of
egress_spec
which will override the drop action.  The action cannot be
overridden in the egress pipeline.

jafingerhut:
@ChrisDodd I can change these drop() calls on ingress to setting egress_spec to 511 (or whatever magic value it is that simple_switch needs to specify dropping the packet) if you prefer, before these changes are merged in.

jafingerhut:
repeater.p4 and repeater.stf in testdata/p4_14_samples/ directory are doing ingress drop using drop() call, too, and there may be others.

Reading that part of the P4_14 spec that you quoted, it seems to imply that drop() is allowed to be used on ingress, and its effect is to assign a value to egress_spec that the implementation will cause to drop the packet, but that later code after that is allowed to overwrite that egress_spec value, which 'undoes' the drop effect.

If all that sounds reasonably kosher, I'll leave the use of drop() in an ingress table action in this PR.

ChrisDodd:
I don't think its necessary to change, as the P4_14 spec does have the language defining what `drop` does in the ingress pipe, different from what it does in the egress pipe.  So the backend will need to choose a "drop" `egress_spec` value can convert the `drop()` in ingress into the appropriate setting of the egress_spec somewhere in the midend or backend to conform with the spec.

mbudiu-vmw:
If the drop thing is a bug please file a separate issue for it.
In the meantime I will merge this PR.

Issue #1036, opened 2017-11-15
intrinsic_metadata in P4_14 is being aliased


theojepsen:
The JSON generated by the P4_14 compiler aliases `intrinsic_metadata` fields, which breaks simple_switch. For example, this is the generated JSON:
```
  "field_aliases" : [
    ...
    [
      "intrinsic_metadata.mcast_grp",
      ["standard_metadata", "mcast_grp"]
    ],
```

When a P4_14 program writes to `intrinsic_metadata.mcast_grp`, the value is "lost". In simple_switch, here:
https://github.com/p4lang/behavioral-model/blob/master/targets/simple_switch/simple_switch.cpp#L351

The value read is not what is expected (what was written by the P4 program).

mbudiu-vmw:
This is a duplicate of https://github.com/p4lang/p4c/issues/718
Unfortunately that bug has been open for a long time, and no one seems to be willing to tackle it.
@hanw : that bug is assigned to you. Why do we need this aliasing stuff?

hanw:
field_alias is not needed when compiling p4-14 program but is required when compiling p4-16 program targeting v1model. The reason is bmv2 internally assumes hard-coded names for intrinsic_metadata that are not part of P4-14 spec (e.g., multicast group). A P4-14 program defines a separate metadata structure for the intrinsic metadata that matches to what is expected in bmv2. But a P4-16 program that uses v1model defines the intrinsic metadata in `standard_metadata` and we need to use `field_alias` to map `standard_metadata.mcast_grp` to `intrinsic_metadata.mcast_grp`. I discussed with Theo, he is going to make a PR to fix the error. Sorry to drop the ball on this one.

Issue #718, opened 2017-06-15
Multicast does not work after converting from p4_14 code to p4_16 code


kecmu:
When converted from p4_14 to p4_16, multicast does not work with the simple_switch bmv2 target. Deleting the whole alias field section in the compiler generated JSON file fixes the problem.

hanw:
Hi @kecmu, can you post an example p4 program for this bug? Thanks.

kecmu:
Hi, please see attached.

[ example_code.tar.gz](https://github.com/p4lang/p4c/files/1079294/example_code.tar.gz)

mbudiu-vmw:
In a private conversation with @kecmu we looked at his program.
There is a mcast field both in intrinsic_metadata (which he defines) and in the standard_metadata.
The compiler generates a section in json which aliases these two fields; when deleting that 'alias' section broadcast works again.
It would be nice to have an stf test using broadcast.

MaoJianwei:
Is this problem fixed now?

mbudiu-vmw:
As far as I know this has not been fixed.

jrmcclurg:
It would be great to have a fix for this. Does anyone have an estimated time frame?

I'm building a tool that auto-generates P4_14 code and then runs `p4c-bm2-ss` to generate a JSON file, so I guess my only choice at the moment (since I need multicast) is to post-process the JSON file to remove the contents of the `field_aliases` section.

jafingerhut:
The field 'mcast_grp' is intended to enable a P4 program writer to control whether a packet is multicast-replicated after it completes ingress processing.

BMv2 simple_switch is currently coded to look for a field in the BMv2 JSON file with the name, or the alias, "intrinsic_metadata.mcast_grp" to decide whether to multicast-replicate a packet.  If it does not find a field with that name, it will never multicast replicate a packet after ingress.

The documentation here for BMv2 simple_switch: https://github.com/p4lang/behavioral-model/blob/master/docs/simple_switch.md

says that if you want to write a P4_14 program that uses mcast_grp, you should define the following header_type and header in your P4_14 program.  This is _separate from the standard_metadata_t header_type and standard_metadata header_ that are predefined for you, and need not (and perhaps should not) be defined in your P4_14 program.
```
header_type intrinsic_metadata_t {
    fields {
        ingress_global_timestamp : 48;
        egress_global_timestamp : 48;
        lf_field_list : 8;
        mcast_grp : 16;
        egress_rid : 16;
[...]

PR #1037, opened 2017-11-15
Only create aliases for metadata fields if compiling P4_16

theojepsen:
This is a fix for #718 and #1036.

### DIFF
--- a/backends/bmv2/header.cpp
+++ b/backends/bmv2/header.cpp
@@ -28,8 +28,8 @@ Util::JsonArray* ConvertHeaders::pushNewArray(Util::JsonArray* parent) {
 }
 
 ConvertHeaders::ConvertHeaders(Backend* backend, cstring scalarsName, BMV2Options& options)
-        : backend(backend), scalarsName(scalarsName), options(options), refMap(backend->getRefMap()),
-          typeMap(backend->getTypeMap()), json(backend->json) {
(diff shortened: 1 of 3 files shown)


backends/bmv2/header.cpp
```diff
@@ -223,19 +223,21 @@ void ConvertHeaders::addHeaderType(const IR::Type_StructLike *st) {
...
-            auto alias = new Util::JsonArray();
-            auto target_name = name_annotation->expr.front()->to<IR::StringLiteral>()->value;
-            LOG2("field alias " << target_name);
-            container->append(target_name);  // name on target
-            // break down the alias into meta . field
-            alias->append(name);      // metadata name
-            alias->append(f->name);   // field name
-            container->append(alias);
-            json->field_aliases->append(container);
+    if (options.langVersion == BMV2::BMV2Options::FrontendVersion::P4_16) {
```
mbudiu-vmw: There's a shortcut to this test (in fact the complement of it) called CompilerOptions::isv1().

mbudiu-vmw: I don't really understand how this fixes the problem when one converts a P4-14 program to a P4-16 program and then tries to compile the latter. Won't this cause the broadcast to not work?

hanw:
Closed as the issue is fixed in #1117

Issue #1038, opened 2017-11-15
p4c-graphs back end produces combinatorial explosion of parse states


jafingerhut:
[tmp-parser.p4.txt](https://github.com/p4lang/p4c/files/1475415/tmp-parser.p4.txt)
[parserI.dot.txt](https://github.com/p4lang/p4c/files/1475416/parserI.dot.txt)

If there are N ways to reach a parse state X, the p4c-graphs back end creates N copies of X in the dot file of the parser.

A small example of this is the sample dot file parserI.dot.txt attached to this issue, produced from the P4_16 source program tmp-parser.p4.txt with this command:
```
p4c-graphs -I path/to/my/p4c/p4include tmp-parser.p4.txt
```
Note that parse_tcp and parse_udp appear twice in the graph.  Ideally they should appear only once.

This combinatorial explosion is small in this example, but makes the dot file pretty much useless for larger parsers with many ways to reach the same parse state (e.g. hundreds).

jafingerhut:
I have tested with the fix now merged in, and things are much better.  Thanks!

PR #1044, opened 2017-11-17
fix for 1038

c3m3gyanesh:


c3m3gyanesh:
#1038

jafingerhut:
I tried this out, and while it is better in avoiding creating duplicate nodes for the same parser state, it creates duplicate edges between those parser states if there are multiple ways to reach the state.  Try running with these changes on the sample program tmp-parser.p4.txt on issue #1038 and you will see 2 arcs labeled 'default' out of parser states parse_tcp and parse_udp.

Also, I ran this modified version on a much larger parser, and it seems to have some kind of larger-than-linear time slowdown when there are multiple ways to reach a parser state.  That is a guess on my part, but there is no reason that the run time should be anything other than linear.  Once a parser state has had a node created, and arcs creates out of it or into it, there should be no reason to do any further processing on that node again, except to determine "I've been here, done that."

c3m3gyanesh:
@jafingerhut I have added another commit in a similar manner by restricting duplicate edges. Let me know if it fixes the issue at hand.
Regarding time taken to generate parser graphs, I am not sure why it is becoming slower. It will be helpful if you can point to any p4 file to test and verify. Also if you have any approach in mind to address the issue in parser graph generation, then we can discuss on that. It will be really helpful to improve the code.

jafingerhut:
[issue1038-parser2.p4.txt](https://github.com/p4lang/p4c/files/1488637/issue1038-parser2.p4.txt)

@c3m3gyanesh I have attached another example program to this comment.  I tried it out with your latest changes, and it also produces many duplicate edges.

In general, I would recommend doing something like a depth-first search of the parse graph (and this should work for control flow graphs of control blocks, too).  Remember for each node whether you have already visited it or not. If you haven't, mark it visited, and proceed as normal.  If you have visited it already, return immediately without doing anything.  This way, if there is more than one way to reach a node in a graph, you only do 'work' for it one time, not many times.

With such a 'visited' state for each node, there should be no reason to keep any other data structures to check whether you have generated a node or edge before -- you will never do it, without having to check anything other than 'visited'.

c3m3gyanesh:
@jafingerhut Thanks for the detailed input. I added "visited" state to the vertex and checking it before traversing the state. [issue1038-parser2.dot.txt](https://github.com/p4lang/p4c/files/1490935/issue1038-parser2.dot.txt)
If the current commit doesn't satisfy any other corner scenario, please let me know.
Also as we are not using the add_unique_edge() api, should i delete it, or we can keep it for future use?

jafingerhut:
Much faster now on a larger parser with many ways to reach the same state!

I have a couple of other programs that it is failing to produce correct results on, e.g. one with a loop in the parser because of parsing header stacks (it is not showing the parser loop in the graph), and it is crashing on a version of switch.p4 auto-translated from P4_14 to P4_16 for reasons I don't know yet -- I can file separate issues for those, if you prefer.  I will link to their source code here in case you want to work on fixing those for this issue, too:

Parser with several loops in the parse state graph:
https://github.com/p4pktgen/p4pktgen/blob/master/p4_programs/tcp-options-parser2.p4

switch.p4 auto-translated from P4_14 to P4_16 (also has parser loops):
https://github.com/p4pktgen/p4pktgen/blob/master/p4_programs/switch-p416.p4

This hand-edited version of switch-p416.p4 above that has had the parser loops commented out.  This one does not give an error when run with the latest code you have checked in for p4c-graphs, so the crash on the previous version might be due to those loops:
https://github.com/p4pktgen/p4pktgen/blob/master/p4_programs/switch-p416-nohdrstacks.p4

mbudiu-vmw:
You should try to run it on all the P4 programs in the testdata folders.

c3m3gyanesh:
@jafingerhut  Yes. Loops won't work with this current implementation. But i didn't realize we can have loops in parser too (restricted by p4 constructs). It will be better to follow up with a different bug or an enhancement for parser graph. Thanks for these detailed insights.

@mbudiu-vmw I will try to run with more number of p4 files. But it is difficult to validate the correctness of the graphs after generating them as to do that we need to go through the parsing logic for each p4 file and compare with the corresponding graphs. I would suggest if we can have some specific P4 files which target the corner cases (e.g. p4 files identified in this conversation) and put them inside p4c-graphs in git, then these can act as our base test cases for any further changes in the code.

mbudiu-vmw:
Maybe you should use the ParserCallGraph.cpp file from the front-end. It computes the call graph of the parser. Then you just have to visit each node and edge in the callgraph and dump it. The call graph has methods for depth-first traversal too.

jafingerhut:
@c3m3gyanesh Loops between states in a parser are the only kind of loop allowed in P4_16, and probably P4_14 as well.  They pretty much go hand in hand with header stacks -- it is difficult to extract into a header stack without a loop in a parser, and it is difficult to take advantage of a loop in a parser without header stacks.

mbudiu-vmw:
This is superseded by #1085, so I will close it.

PR #1046, opened 2017-11-17
Fix for issue #961

mbudiu-vmw:
This fixes the old stf file parser to ignore the key handle part of a command.
This makes the tests pass again, but the stf support in the old parser is still incomplete.

### DIFF
--- a/backends/bmv2/CMakeLists.txt
+++ b/backends/bmv2/CMakeLists.txt
@@ -100,13 +100,6 @@ set (BMV2_TEST_SUITES
   )
 
 set (XFAIL_TESTS
-  # The following tests are broken because of issue #961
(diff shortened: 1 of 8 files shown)


mbudiu-vmw:
I still have a failing test that I have to figure out.

mbudiu-vmw:
@ChrisDodd : I had to rewrite a test to make it pass. If the original was correct then there is a bug somewhere else.

ChrisDodd:
I think the rewrite is fine -- this test is old code that was trying to use `drop` metadata flag instead of the `drop` primitive (since the drop primitive used to only be available in the egress pipe).  IMO drop is still kind of a mess.

PR #1053, opened 2017-11-21
P4_14->16 conversion hooks for register/stateful conversion

ChrisDodd:
This adds some hooks so that convertExternInstance can add additional declarations to the scope of the control in which is being converted and/or get at other previously converted objects in the scope (registers is the particular need).


ChrisDodd:
scope is passed as an argument to convertExternInstance so that backend-specific subclasses of ExternConverter can use it in their override...

Issue #1057, opened 2017-11-24
p4c segfault when counter name and header name is the same.


rbooysen:
p4c crashes with this error:

terminate called after throwing an instance of 'boost::exception_detail::clone_impl >'
  what():  boost::too_few_args: format-string referred to more arguments than were passed
Aborted (core dumped)

[name_clash_crash.p4.txt](https://github.com/p4lang/p4c/files/1501630/name_clash_crash.p4.txt)

mbudiu-vmw:
I will assume that this issue can also be closed.

PR #1060, opened 2017-11-28
Fix for issue #1057

mbudiu-vmw:


mbudiu-vmw:
The format string seemed to be correct, but maybe the format library does not work with null pointers?

PR #1065, opened 2017-11-29
Eliminate global per-translation-unit state

sethfowler:
Recently I've been spending some time working on tasks involving translating between different architectures and writing unit tests against those architectures. Some tasks have been far harder than I expected, and I think now might be a good time to refactor things a little to make things easier. Here are the features I'd like to have:

1. We should be able to control whether individual compiler diagnostics are treated as errors, as warnings, or ignored. We should be able to control this via the command line, via pragmas in the program, and in GTest code. We've already implemented most of this internally; I'll pull it into this repo soon.

2. We should be able to parse P4 code and examine the resulting IR without affecting the "primary" program in any way.

3. We should be able to parse P4 code from multiple sources and combine it into a single program easily and flexibly. We should revise our GTest helper functions to use this approach.

This PR is architectural work to enable items 2 and 3. In a second PR, I'll add some new APIs that take advantage of this stuff, but that comes later.

The basic problem I repeatedly run into is that I'd like to combine snippets of code - different translation units, essentially - that either need to be compiled with different settings, or that generate side effects that pollute each other. As an example, I may need to enable a certain architecture and set certain compiler flags while compiling one snippet, while a second snippet may generate a ton of warnings that shouldn't pollute the state of the primary program or the other snippets. In some cases there are ad hoc solutions for these issues, but sometimes they're hard enough to work around that right now we need to use a different approach like generating IR manually.

The approach to fixing this that I've taken in this PR is to either eliminate that state or make it more explicit and controlled. Here are the concrete changes:

- There is no more global `InputSources` singleton. A `SourceInfo` object knows what `InputSources` it's associated with, and any number can exist in the program at one time.
- To hold per-translation-unit state - a "compilation context" - I've added `ICompilerContext`, and a `CompilerContextStack` for them to live in. Rather than accessing globals directly, we can access them through the compilation context stack. The stack can be pushed and popped as you might expect, and there's a RAII class, `AutoCompileContext`, to ensure we keep them properly nested.
- Rather than a global `ErrorReporter` singleton, an `ErrorReporter` now lives in the compilation context.
- `CompilerOptions` and friends also now live in the compilation context, though this doesn't make much of a difference yet because I haven't eliminated the code where we pass them around manually.

With these changes, we can compile a translation unit with different settings just by pushing a new compilation context onto the stack, and any warnings or errors will disappear when we pop that compilation context away. (The user will still see them for now, but item 1 above will give us the tools to fix even that.) Backends which have more per-translation-unit state than just an `ErrorReporter` and a `CompilerOptions` instance can create their own `ICompilerContext` subclasses to hold that data.

That's hopefully enough explanation (if not too much!) - the rest should be clear from the code.

One TODO about this design is that the current interaction between `ICompilerContext` subclasses and `CompilerOptions` subclasses doesn't feel right. If `FooOptions` inherits from `BarOptions` then we need `FooContext` to inherit from `BarContext`, and ideally `FooContext` wouldn't contain an unneeded and unused `BarOptions`! This is dealt with using a template helper in the current code, but I feel like the obvious right thing is to just move the state of the `CompilerOptions` subclasses directly onto the corresponding `ICompilerContext` subclasses. Then normal inheritance would "do the right thing" and no template magic would be needed. I'll deal with this in a later PR, though.


frontends/common/options.h
```diff
@@ -109,4 +109,36 @@ class CompilerOptions : public Util::Options {
     DebugHook getDebugHook() const;
 };
 
+
```
mbudiu-vmw: I expect an `#include` for lib/compile_cointext.h

test/gtest/p4runtime.cpp
```diff
@@ -379,7 +381,7 @@ TEST(P4Runtime, P4_16_MatchFields) {
 #if 0
 // Disabling this test because it does not work without installation
 // see issue https://github.com/p4lang/p4c/issues/924
-TEST(P4Runtime, P4_14_MatchFields) {
+TEST_F(P4Runtime, P4_14_MatchFields) {
```
mbudiu-vmw: How do you feel about fixing this test?

ir/pass_manager.cpp
```diff
@@ -41,7 +41,8 @@ const IR::Node *PassManager::apply_visitor(const IR::Node *program, const char *
                 program = program->apply(**it);
                 LOG3(log_indent << "heap after " << v->name() << ": in use " <<
                      n4(gc_mem_inuse(&maxmem)) << "B, max " << n4(maxmem) << "B");
-                int errors = ErrorReporter::instance.getErrorCount();
+                auto& context = BaseCompileContext::get();
+                int errors = context.errorReporter().getErrorCount();
```
ChrisDodd: perhaps there should be a global `::getErrorCount()` (analogous to the global `::getDiagnosticCount()` that already exists)?

test/gtest/p4runtime.cpp
sethfowler: I plan to fix that in a later patch in this series. We need to just make GTest use the same code for finding included headers that the normal compiler binary uses.

ir/pass_manager.cpp
sethfowler: There actually is, but I didn't update this call site to use it. =) I'll do that.

frontends/common/options.h
sethfowler: Thanks; I'll fix it.

PR #1069, opened 2017-11-30
support another blackbox primitive  in p4-14 type inference

hanw:
This is undoubtfully a hack, but we already have hacks in this file to support type inference on `execute_stateful_` primitives.


hanw:
Good =)

Issue #924, opened 2017-09-20
gtestp4c looks for inexistant directory


mbudiu-vmw:
One of the gtestp4c tests fails as follows on my machine:

```
[ RUN      ] P4Runtime.P4_14_MatchFields
cpp: error: /usr/local/share/p4c/p4include/v1model.p4: No such file or directory
cpp: warning: ‘-x c’ after last input file has no effect
cpp: fatal error: no input files
compilation terminated.
error: Preprocessor returned exit code 256; aborting compilation
error: 1 errors encountered, aborting compilation
Couldn't parse test case source
/home/mbudiu/git/p4c/test/gtest/p4runtime.cpp:458: Failure
Value of: test
  Actual: false
Expected: true
[  FAILED  ] P4Runtime.P4_14_MatchFields (17 ms)
```

I have never run 'make install', so the p4c directory under share does not exist indeed.

sethfowler:
Hmm, that scenario indeed has not been tested. I'll check it out. Thanks for letting me know, Mihai.

mbudiu-vmw:
This issue has been solved by brute-force by commenting out a gtest that was failing.
I don't know if this warrants us to mark the issue closed, though. I would rather fix the broken gtest.

PR #1070, opened 2017-11-30
Add support for diagnostics with configurable behavior

sethfowler:
We've found it useful to control whether a particular diagnostic is treated as a warning or an error on a per-program basis. This prevents us from losing test coverage when tests trigger spurious errors, and it allows us to write tests that fail if a certain warning is triggered. It's also a convenient feature for users of the compiler.

This PR brings this feature to `p4c`. The approach is to create a generic infrastructure for controlling whether particular diagnostic messages are treated as warnings, as errors, or simply ignored. Diagnostic behavior can be configured both via the command line and via the P4-16 `@diagnostic` annotation or the P4-14 `pragma diagnostic`.

Within the compiler code itself, you can add one of these controllable diagnostics by using the `DIAGNOSE_WARN` or `DIAGNOSE_ERROR` macros. Here's an example:

```c++
DIAGNOSE_ERROR("foo_failure", "Your program contained an invalid foo: %1%", fooExpr);
```

This reports a diagnostic called `foo_failure` which is an error by default. If the program includes `@diagnostic("foo_failure", "warn")` or `pragma diagnostic foo_failure warn`, or if the user specifies `--Wwarn=foo_failure` at the command line, then this diagnostic will be treated as a warning instead.

Hopefully the code is fairly clear, but here's a high level summary of the changes:
 - Support for `DIAGNOSE_WARN` and `DIAGNOSE_ERROR` has been added to `error.h`. I split `ErrorReporter` out into its own header to avoid some cyclic dependencies; this also enabled me to store `ErrorReporter` on the compilation context directly instead of via a pointer.
- The information about which actions (ignore, warn, or error) are configured for each diagnostic now lives on the compilation context. Default actions for warnings and errors also live there; this replaces `ErrorReporter::warningsAsErrors`.
- `CompilerOptions` has been extended to allow diagnostics to be configured via the command line.
- A new `ApplyOptionsPragmas` pass has been added; backends can run it after parsing to gather any diagnostic options specified by pragmas. You can use it for more than just diagnostics as well; the pass is extendable so that each backend can add support for their own pragmas.

While I was at it, I fixed the issue in #924 that prevented P4-14 GTests from working if `make install` wasn't run. The problem was that the GTests couldn't find `p4_14include/` because we search for it relative to the compiler executable and for GTests the executable is in `test/` instead of at the root of the build tree. I worked around this by checking for `p4include` and `p4_14include` one directory up, as well. There's more to be done in a future PR, but this was enough to reenable the disable P4Runtime test and to let me include P4-14 tests in the test suite for the new diagnostics code.


sethfowler:
Ah, looks like I had to update test outputs since making those `SimplifyDefUse` warnings into diagnostics affected the format of the output.

Hopefully I got the changes correct; for some reason the outputs ended up having the entire path to the file instead of just the filename when I ran the test suite locally. I had to fix them with perl.

mbudiu-vmw:
What is the semantics of such a pragma? It applies from that point onwards? Can you get first warnings and later errors for the same construct? (I assume the answer is no). 
If not, then what if you have conflicting requests?
What if you ask for warnings that don't exist?

The situation is even murkier for P4-16 which has no `@pragma` annotation, which indeed have to be attached to some syntactic object.

mbudiu-vmw:
Could you ask git to actually classify the error_helper file as a rename instead of a new file?
The diff is too large to read.

mbudiu-vmw:
We need a place to document the compiler implementation of features that do not fit in the spec.
For example, these pragmas/diagnostic options, it is not enough to have them in the program comments. We need compiler users to be able to read about them. It looks to me like we need to create a new manpage or a README.md for features of the compiler such as this one. Otherwise these features will not be discoverable.

ChrisDodd:
At the moment all of these pragma/annotation options are strictly global -- if they appear anywhere in the file they have global effect everywhere.  So it's not possible to turn off warnings for one block of code and then turn them back on for the rest of the code.

We definitely due need more comprehensive documentation on how to use the compiler (a man page at least), along with documentation of all the supported pragmas/annotations.

sethfowler:
Yep, we definitely need to create compiler documentation. That's probably too big a job for this PR, but if we can get that going in another PR, I'll be happy to add the documentation for this stuff.

@mbudiu-vmw, git only stores complete files (well, ignoring compression, at least); it decides what changes constitute a rename vs a new file on the fly. Newer versions do a better job. My local version of git does show this as a rename.

sethfowler:
@mbudiu-vmw just to add to what Chris said: the way things work right now is that it doesn't matter which object the pragmas/annotations are attached to. This lets us pretend that they're syntactically global.

They *are* semantically global. We apply them once, immediately after parsing the program. They are desugared into compiler options which are processed by the normal `CompilerOptions::process()` method, so the behavior you'd expect applies: if you have conflicting requests, the last one wins.

If you ask for warnings that don't exist, we will show you a warning; this is what the `isRecognizedDiagnostic()` method does. Chris has suggested elsewhere that we might generate the list using a script, and I think that's a great idea, but for now you need to add them manually. At any rate, that should help catch user mistakes, and if the user doesn't want to see those warnings, that warning itself is a diagnostic that can be disabled. =)

I'm sure we'll evolve this feature further. Consider this a starting point.

PR #1072, opened 2017-11-30
Make P4TEST_REPLACE strip paths correctly on macOS

sethfowler:
I cannot conceive how simple character classes like `\w` don't work correctly on the `sed` that ships with macOS, but somehow it's true. =\ This patch makes `P4TEST_REPLACE` strip paths correctly on both macOS and Linux.

### DIFF
--- a/backends/p4test/run-p4-sample.py
+++ b/backends/p4test/run-p4-sample.py
@@ -83,7 +83,14 @@ def run_timeout(options, args, timeout, stderr):
             # copy stderr to the specified file, stripping file path prefixes
             # from the start of lines
             outfile = open(stderr, "w")
-            local.filter = Popen(['sed', '-e', 's|^[\S*/]\S*/||'], stdin=PIPE, stdout=outfile)
+            # This regex is ridiculously verbose; it's written this way to avoid
+            # features that are not supported on both GNU and BSD (i.e., macOS)
+            # sed. BSD sed's character class support is not great; for some
+            # reason, even some character classes that the man page claims are
+            # available don't seem to actually work.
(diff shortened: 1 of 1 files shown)


sethfowler:
Doh. Looks like some messages in the logs don't include a colon.

backends/p4test/run-p4-sample.py
```diff
@@ -83,7 +83,14 @@ def target():
...
             # from the start of lines
             outfile = open(stderr, "w")
-            local.filter = Popen(['sed', '-e', 's|^[\S*/]\S*/||'], stdin=PIPE, stdout=outfile)
+            # This regex is ridiculously verbose; it's written this way to avoid
+            # features that are not supported on both GNU and BSD (i.e., macOS)
+            # sed. BSD sed's character class support is not great; for some
+            # reason, even some character classes that the man page claims are
+            # available don't seem to actually work.
+            local.filter = Popen(['sed', '-E',
+                r's|^[-[:alnum:][:space:]_/]*/([-[:alnum:][:space:]_]+\.p4\([[:digit:]]+\))|\1|'],
```
ChrisDodd: This seems to be wrong -- there's a colon (`:`) between the filename and the line number, so it should be:

    r's|^[-[:alnum:][:space:]_/]*/([-[:alnum:][:space:]_]+\.p4:[[:digit:]]+)|\1|'

?

jafingerhut:
It wouldn't be simpler to do this file modification in Python code, which is probably more portable across platforms in its behavior in this regard?

ChrisDodd:
There's probably a way to do this in python without invoking an external process and without reimplementing all of sed in python...

PR #1075, opened 2017-12-01
Support for shared action_profiles in P4_14->16 conversion

ChrisDodd:
- fix bmv2 backend to emit global action_profile/selector info when needed

The main change here is to make the action_profile/action_selector objects created in P4_14->16 conversion globals instead of just being anonymous constructor calls in the tables.  This allows them to be shared between tables (including tables in different control blocks).  Along with this, I fixed the bmv2 backend to include such global action_profile/selector objects in the pipeline(s) that use them; previously it would not include globals, leading to incorrect config that would cause bmv2 to crash.


ChrisDodd:
Yes, unfortunately.  The newline was needed to avoid getting errors when a global declaration was followed by a preprocessor directive.  Fortunately, you can look at the changes individually -- the last change is basically just a huge number of "add a newline at the end" changes and can be mostly ignored.

backends/bmv2/JsonObjects.h
```diff
@@ -24,9 +24,9 @@ limitations under the License.
...
 
 class JsonObjects {
- protected:
-    Util::JsonObject* find_object_by_name(Util::JsonArray* array,
-                                          const cstring& name);
+ public:
+    static Util::JsonObject* find_object_by_name(Util::JsonArray* array,
+                                                 const cstring& name);
 
  public:
```
sethfowler: This extra `public:` is redundant, so it could be removed.

backends/bmv2/simpleSwitch.cpp
```diff
@@ -531,6 +531,8 @@ SimpleSwitch::convertExternInstances(const IR::Declaration *c,
         backend->meter_arrays->append(jmtr);
     } else if (eb->type->name == v1model.action_profile.name ||
             eb->type->name == v1model.action_selector.name) {
+        if (BMV2::JsonObjects::find_object_by_name(action_profiles, name))
+            return;
```
sethfowler: Might be good to add a comment explaining why we exit early here.

frontends/p4/fromv1.0/programStructure.cpp
```diff
@@ -1904,6 +1900,23 @@ void ProgramStructure::createControls() {
         return;
     }
 
+    for (auto ap : action_profiles) {
+        auto action_profile = convertActionProfile(ap.first, ap.second);
+        declarations->push_back(action_profile);
```
sethfowler: Should you be checking for a null `action_profile` here? Looks like `convertActionProfile()` can return null.

Issue #1074, opened 2017-12-01
Compiler gives incorrect error message when it does not recognize an option


mbudiu-vmw:
```
 ./p4c-xdp --Werror -o xdp1.c xdp1.p4
[--Wwarn=unknown_diagnostic] warning: Unrecognized diagnostic: -o
error: Only one input file must be specified: xdp1.c,xdp1.p4
./p4c-xdp: Compile a P4 program
--help                                  Print this help message
```

The error message should be about --Werror, not about -o.

sethfowler:
Seems like Util::Options needs a tweak. I’d love if we could configure an option to only accept an argument if you use the “—foo=bar” form, and otherwise be treated as having no argument. That would not only fix this, but it would let “—Werror” by itself have the behavior that you’d expect from using GCC and Clang.

PR #1076, opened 2017-12-01
Allow options to have optional arguments

sethfowler:
As @mbudiu-vmw pointed out in #1074, when you forget the argument to an option, the error message that the compiler reports isn't very helpful. The ambiguity is unfortunately impossible to resolve in general, I think, but this specific case is fairly easy: in analogy with GCC, users expect that `--Werror` without an argument will be accepted and will make the compiler treat all warnings as errors, and we should make that work.

To that end, this PR adds a flag to `Util::Options::registerOption()` that lets you specify that an option's argument may be omitted. To avoid ambiguity, for such options we don't allow `--foo bar`; only `--foo=bar` is accepted. That means that you must use `--Werror=foo` and not `--Werror foo`. `--foo bar` still works for options with non-optional arguments, of course.

With that change in place, we can make `--Werror` and `--Wdisable` do what you might expect. `--Wwarn` also works and makes us treat all warnings as, well, warnings; this seems a bit silly but sometimes it's useful to do this kind of thing in Makefiles and the like to cancel flags introduced elsewhere.

Since it's now redundant, I also removed `--Wwarnings-as-errors`.

### DIFF
--- a/frontends/common/options.cpp
+++ b/frontends/common/options.cpp
@@ -120,31 +120,45 @@ CompilerOptions::CompilerOptions() : Util::Options(defaultMessage) {
     registerOption("-o", "outfile",
                    [this](const char* arg) { outputFile = arg; return true; },
                    "Write output to outfile");
-    registerOption("--Wwarnings-as-errors", nullptr,
-                    [](const char*) {
-                        auto action = DiagnosticAction::Error;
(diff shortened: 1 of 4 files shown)


sethfowler:
It was there for ~24 hours; doubtful that anyone but you has embraced it yet. =)

I can add a test.

sethfowler:
OK, I just added a test.

PR #1078, opened 2017-12-05
print the output directly to stdout rather than using parser.exit

cc10512:
Printing the message through parser.exit puts it to stderr, and thus
version and targets are not available in scripts without redirection.


shalinish:
+1

Issue #1079, opened 2017-12-05
update_checksum/verify_checksum for bmv2-ss-p4org target missing if_cond


ygottlieb-vencorelabs-com:
I think there may be a bug in the p4c compiler for the BMv2 backend that causes it not to generate the if_cond block.  For example, I have the p4 file and the generated json file below. Since the update/verify calls use false as the first parameter, I would expect the generated code to include an if_cond block telling the simple_switch not to actually compute the checksum.  I believe the problem is in backends/bmv2/simpleSwitch.cpp that is missing the following lines around line 661:

auto ifcond = conv->convert(mi->expr->arguments->at(0));
cksum->emplace("if_cond", ifcond);

adding those lines generates the if_cond block as expected.

[cksumtest.json.txt](https://github.com/p4lang/p4c/files/1531906/cksumtest.json.txt)
[cksumtest_p4.txt](https://github.com/p4lang/p4c/files/1531907/cksumtest_p4.txt)

ygottlieb-vencorelabs-com:
Thank you.

PR #1080, opened 2017-12-05
Fix for issue #1079

mbudiu-vmw:
Bug report and solution contributed by @ygottlieb-vencorelabs-com

### DIFF
--- a/backends/bmv2/simpleSwitch.cpp
+++ b/backends/bmv2/simpleSwitch.cpp
@@ -656,6 +656,8 @@ SimpleSwitch::convertChecksum(const IR::BlockStatement *block, Util::JsonArray* 
                     cksum->emplace("target", jleft->to<Util::JsonObject>()->get("value"));
                     cksum->emplace("type", "generic");
                     cksum->emplace("calculation", calcName);
+                    auto ifcond = conv->convert(mi->expr->arguments->at(0), true, false);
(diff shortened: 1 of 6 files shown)


mbudiu-vmw:
It turns out that parser_dc_full.stf will actually work for this case.
I know because the first attempt to fix this bug made that test case crash.

cc10512:
But obviously was not correctly covering this case, no? Otherwise how come it was passing before you fixed the bug?

mbudiu-vmw:
Before fixing this bug we were generating no condition for the checksum verify/update, so it was always unconditional. Note that BMv2 may still skip the verify/update if the field list contains fields of an invalid header (the actual semantics in bmv2 when some fields are invalid and some are not is not clear).

antoninbas:
@mbudiu-vmw if the target field is invalid, there will be no verify / update
if a field list contains fields of an invalid header, these fields will be skipped when computing the checksum and will therefore not contribute to the checksum
if all fields are invalid, the checksum will be 0

debatgithub:
When a register section is defined in a p4 file and compiled with p4_16,
the resulting json doesn't have any reference to it.
Has the registers/meters/counters support totally disappeared from p4_16?
Or is there a back door to re-enable it?

mbudiu-vmw:
Can you provide a test example?
We have several tests in testdata/*/*.p4 which use registers.

PR #1081, opened 2017-12-06
Change base Docker image to behavioral-model

antoninbas:
The behavioral-model image now has pi as a base image, not the other way around.


antoninbas:
Do not merge until corresponding pull requests have been merged for bmv2 & PI.

Issue #1082, opened 2017-12-06
Compiler Bug [p4-graphs]: IR loop detected


qin-nz:
Hi, I am using p4-graphs to compile the code and I got an error.

```
terminate called after throwing an instance of 'Util::CompilerBug'
  what():  In file: /home/dy/p4c/ir/visitor.cpp:267
Compiler Bug: IR loop detected
```

The following file is generator by front-end compiler.

[mapswitch.zip](https://github.com/p4lang/p4c/files/1534444/mapswitch.zip)

jafingerhut:
Your parser has a loop from the state parse_ethernet back to itself, extract'ing into an ethernet header, but it is not a header stack.  I am a little bit surprised that p4c-bm2-ss _doesn't_ give an error when compiling this.  I would suspect some P4 devices would eventually give a parser error if you gave it a packet with a repeated Ethernet header with ethertype 0x0001.

Question to p4c developers: Does this look like a program p4c-bm2-ss should quietly accept?  It does have an extract in the parser loop, but not into a header stack, so there is no way except the size of the input packet to bound the number of iterations.

The p4c-graphs back end does give the error you see, probably because it doesn't yet handle loops in the parser at all, not even correct ones that are made finite by limiting them to the maximum size of a header stack being extracted into.  This looks like a bug worth fixing in p4c-graphs.

mbudiu-vmw:
No program should crash the compiler, so this is a bug.
The program is legal if you can overwrite a valid header, something which is allowed by the spec.
The compiler could give a warning for such cases. I have implemented an analysis to detect such cases, but it is currently commented-out (in the loop unrolling pass). I hope to revive it someday.
At runtime such a loop will probably terminate with either a PacketTooShort or ParserTimeout error.

qin-nz:
Yes, my  parser has a loop.
I just need the last Ethernet header if there are more than one header.

I think give a warning for such cases is necessary.

qin-nz:
Thanks @mbudiu-vmw

PR #1085, opened 2017-12-08
Fix for issues #1082 and #1038

mbudiu-vmw:


mbudiu-vmw:
The edges are currently not labeled; if there are multiple edges between a pair of nodes it is because there are multiple transitions between that pair of nodes in the program.

backends/graphs/controls.cpp
```diff
@@ -107,6 +129,14 @@ bool ControlGraphs::preorder(const IR::PackageBlock *block) {
             controlStack.popBack();
             GraphAttributeSetter()(g_);
             writeGraphToFile(g_, name);
+        } else if (it.second->is<IR::ParserBlock>()) {
```
antoninbas: while I'm all for unifying code, it looks like this and the `writeGraphToFile` overload for `P4::ParserCallGraph` do not belong here, not in a class called `ControlGraphs` in any case
I believe it would be better to leave the code for parsers in its own file. It doesn't look that it is re-using a lot of the existing code from `ControlGraphs`which is mostly written around the boost graph library.

mbudiu-vmw:
Once you like the output we'll update the readme too.

backends/graphs/controls.cpp
mbudiu-vmw: The problem is that the preorder(IR::Package) method does not belong to the ControlGraphs file itself.

In fact, there is hardly any code used for the parser part.  If you want labels on the edges I will have to duplicate some of the code which I reused, because the ParserControlFlowGraph class does not have edge information.

I could not really find a benefit for using that boost library, it looks overly complicated. The dot file format has been stable for 20 years now, and it is really very simple.

antoninbas: With the removal of the old parser code, the current `Graphs` base class doesn't make sense any more. I believe it was just an attempt at unification based when the generation code was similar for parsers and controls. With your current proposal I'm not sure I see the need for a base class at all, or in any case a base class with boost members creeping in.
I don't see why you tried to piggyback the parser graph generation on top of the `ControlGraph` visitor by reusing the same visitor function. It looks like the parser graph generation can be done with its own, very simple, visitor in a cleaner way.

mbudiu-vmw:
@antoninbas : I have reworked this; it will now label the edges and also show in the node the expression that is used to dispatch to the next state.

mbudiu-vmw:
I have pushed a new figure; once the Travis build is done I will merge the PR.

PR #1087, opened 2017-12-08
debugging/logging improvements

ChrisDodd:
- LOG_FEATURE macro for logs messages triggered on a feature name
  instead of source file name
- Dump allows dumping Visitor::Context in a log message
- demangle class names when printing them

### DIFF
--- a/ir/visitor.cpp
+++ b/ir/visitor.cpp
@@ -468,3 +468,24 @@ IRNODE_ALL_NON_TEMPLATE_CLASSES(DEFINE_APPLY_FUNCTIONS, , , )
 
 std::ostream &operator<<(std::ostream &out, const IR::Vector<IR::Expression> *v) {
     return v ? out << *v : out << "<null>"; }
+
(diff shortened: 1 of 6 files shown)


ir/visitor.cpp
```diff
@@ -468,3 +468,24 @@ IRNODE_ALL_NON_TEMPLATE_CLASSES(DEFINE_APPLY_FUNCTIONS, , , )
...
 std::ostream &operator<<(std::ostream &out, const IR::Vector<IR::Expression> *v) {
     return v ? out << *v : out << "<null>"; }
+
+#if HAVE_CXXABI_H
+#include <cxxabi.h>
+
+cstring Visitor::demangle(const char *str) {
+    int status;
+    cstring rv;
+    if (char *n = abi::__cxa_demangle(str, 0, 0, &status)) {
```
mbudiu-vmw: I didn't know about this function. If I knew maybe we didn't need to use setName() for Passes; it is only used to avoid printing the mangled name.

Issue #1097, opened 2017-12-13
Bug [Simple_switch]: register_array is empty or duplicate.


qin-nz:
Hi,
I want to make a register available both in ingress and egress. So I declare it out of control.
The result json for bmv2 might be wrong.

`"register_arrays" : [],`

And this will lead to bmv2 quit unexpectedly. 
https://github.com/p4lang/behavioral-model/issues/492

Source code is following.
```
#include <core.p4>
#include <v1model.p4>

struct Headers {}

struct Meta {}

parser p(packet_in b, out Headers h, inout Meta m, inout standard_metadata_t sm) {
    state start {
        transition accept;
    }
}

register<bit<8>>(2) r;

control ingress(inout Headers h, inout Meta m, inout standard_metadata_t sm) {
    apply {
        bit<8> x;
        r.read(x,0);
    }
}

control vrfy(inout Headers h, inout Meta m) { apply {} }
control update(inout Headers h, inout Meta m) { apply {} }
control egress(inout Headers h, inout Meta m, inout standard_metadata_t sm) { apply {} }
control deparser(packet_out b, in Headers h) { apply {} }

V1Switch(p(), vrfy(), ingress(), egress(), update(), deparser()) main;
```

qin-nz:
Another example can make  `register_array` has duplicate register.

```
#include <core.p4>
#include <v1model.p4>

struct H {}
struct M {}
parser p(packet_in b, out H h, inout M m, inout standard_metadata_t sm) {
    state start {
        transition accept;
    }
}
control ingress(inout H h, inout M m, inout standard_metadata_t sm) {
    register<bit<8>>(2) r;
    apply {
        bit<8> x;
        r.read(x,0);
    }
}
control egress(inout H h, inout M m, inout standard_metadata_t sm) {
    register<bit<8>>(2) r;
    apply { 
        bit<8> y;
        r.read(y,0);} 
    }
control vrfy(inout H h, inout M m) { apply {} }
control update(inout H h, inout M m) { apply {} }
control deparser(packet_out b, in H h) { apply {} }

V1Switch(p(), vrfy(), ingress(), egress(), update(), deparser()) main;
```

jafingerhut:
Your second example, with duplicate names in the bmv2 JSON file because of using the same instance name in both ingress and egress, is not unique to register instances, but is a problem for all kinds of instances that have the same name and are instantiated in both ingress and egress in P4_16 programs.  That issue is covered by #1073 and/or #486.

[...]

Issue #696, opened 2017-06-05
Register cannot be declared at the top-level


qizhe:
I use v1model (v1model.p4) to build the program. When I tried to instantiate extern registers at the top-level of my program and read register values at one of actions, I got error:

P4 switch s1 did not start correctly.Check the switch log file.

S1 log:
terminate called after throwing an instance of 'std::out_of_range'
what(): _Map_base::at
Calling target program-options parser

After I changed to declare registers in the ingress control block, I can run my program.
I don't find in P4_16 spec that forbids us to declare extern objects in the top-level.

Sample code from issue242.p4 (only changes the place of declaring registers):

[issue242_revise.txt](https://github.com/p4lang/p4c/files/1050844/issue242_revise.txt)

jafingerhut:
A tiny bit of analysis on the cause of the crash.  If it is true that it should be allowed to declare a regsister at the top level of a P4-16 program, then it appears that p4c is generating and incorrect bmv2 JSON configuration file for the P4 progarm issue242_revise.txt attached above.

I have used versions of p4lang/p4c and p4lang/behavioral-model dated 2017-Jun-03 in my testing.

I used p4c to compile the program issue242_revise.txt.  It generates a bmv2 JSON configuration file that has an empty list for the value of its "register_arrays" key, yet it also generates actions with "register_write" and "register_read" primitives referring to register_array's named "debug" and "reg", as they are named in the P4 program.  This causes behavioral-model's simple_switch program to crash when trying to look up a non-existent register_array with those names.

hanw:
BMV2 backend does not support global register extern right now. A register extern instance must be local to a control block. The compiler should have returned an error in this case. However, we could add support to global extern instance since the spec does not explicitly disallow it.

mbudiu-vmw:
I think that we should write a midend pass which pushes global instantiations into controls/parsers if they are used only within a single control or parser. Then we only need to handle this case in the back end. The p4-14 front-end does something similar for tables.

cc10512:
Not sure I understand your comment. You need to handle the global instance if it is used in multiple controls regardless.

mbudiu-vmw:
Many architectures will not support registers shared between different control/parser blocks.
Whether BMv2/PSA do or not it is a matter of choice.

mbudiu-vmw:
This seems to be closeable.

PR #1099, opened 2017-12-15
Fix for issues #696 and #1097

mbudiu-vmw:


jafingerhut:
I am guessing that if someone provided tests using STF to check whether simple_switch is doing the expected thing for registers accessed from both ingress and egress, you wouldn't mind?

mbudiu-vmw:
We never have enough tests.
However, the semantics of a register accessed from both input and output is not clear; in particular, the interleaving of packets in the two pipelines is not specified. So I am not sure what tests one should write.

jafingerhut:
It is pretty easy to write an STF test that for each packet, only accesses a single element of the register array, but in both ingress and egress.  It is safe to assume that for the same packet, ingress must complete before egress begins.

For different packets, agreed that the order may not be consistent, but the test can be written to avoid that issue.  Let me try to make something simple here.

jafingerhut:
[issue1097.zip](https://github.com/p4lang/p4c/files/1564279/issue1097.zip)

I have attached a ZIP file with a P4 program and corresponding STF file that fail with the latest master version of p4c-bm2-ss, but succeed with the changes Mihai has in this pull request, because the changes in this PR do correctly include a definition for a globally defined register in the "register_arrays" section of the bmv2 JSON file, and simple_switch is processing the packets as I would expect it should.

Mihai, feel free to add these two files to your pull request if you wish.

Issue #1073, opened 2017-12-01
Bug fix for P4_16 bmv2 JSON and P4Runtime control plane names?


jafingerhut:
As p4c-bm2-ss is now, when compiling P4_16 programs, controlPlaneName returns names like foo when foo is a table or extern or control instantiated directly inside of any top-level control block (and probably also parser, but I haven't checked that).

This leads to, in bmv2 JSON files and p4Runtime API files, conflicting names if foo is instantiated in both ingress and egress control blocks, for example.

If controlPlaneName instead returned a name like ingress.foo or egress.foo, then I think this would solve most or all of the name conflicts that currently occur between same-named things inside of different top-level controls, in both bmv2 JSON and p4Runtime API files generated by p4c-bm2-ss.

controlPlaneName already returns hierarchical names if there are one or more controls instantiated in a tree structure within ingress.  This is good, and should not change.  For example, if inside of ingress you instantiate another control with an instance name of c1, and inside c1 there is a table t1, then currently controlPlaneName returns c1.t1.  If that changed to ingress.c1.t1 (where by 'ingress' I mean whatever the instance name that the ingress control instance has in the top level package instantiation of V1Model), then that is the change I am suggesting.

I don't know how to make such a change, or whether it would break something else, but it seems worth a try.

jafingerhut:
Even if someone could create a PR that had only the necessary changes for the names, but not tests and other things, I could run some experiments with it to investigate what else needs fixing, and perhaps update STF files, etc.

mbudiu-vmw:
The compiler changes may not be very difficult, but we don't want to update all stf files. If we do the change in the most straightforward way all stf files for all P4-16 programs will need changes.

However, the spec says that shorter names can be used in the control plane as long as they are not ambiguous. So we need to update the program that interfaces between the stf files and the CLI to automatically convert the short names into the long forms. That program loads the entire JSON, so it knows the long names.

jafingerhut:
I don't mind updating all stf files for P4_16 programs, if that is a clean long-term way to go.

mbudiu-vmw:
How is this issue different from #486?

jafingerhut:
With #486 the issue is names of fields need more global qualification.  This issue is names of tables, externs, control, parsers, etc.

mbudiu-vmw:
I will mark this as fixed.
But the PR has not been merged yet, so we'll keep it around.

jafingerhut:
[...]

PR #1106, opened 2017-12-19
Fix for issue #486

mbudiu-vmw:
This fixes a long-standing bug where the bmv2 JSON generated from a P4-16 program had name clashes. This happened because some local identifiers in P4-16 become global identifiers in JSON. The solution was to make sure that the names are unique. This is done using two tricks:
- we use hierarchical names for local variables, which are prefixed with the enclosing control/parser
- we make sure that the fields of the headers and usermetadata structures have different names, by renaming the fields in the userMetadata type structure

This PR also sorted the file names in several makefiles.
It also changed the name of a few functions from *update* to *compute*, because v1model.p4 uses the name ComputeChecksum for a control (and not UpdateChecksum).

The BMv2 testing script that runs STF programs has been updated to allow unambiguous names that are suffixes of full names when specifying actions or tables.

These changes may potentially break other control planes that are not being tested. In particular, I have filed https://github.com/p4lang/behavioral-model/issues/500 with the behavioral model, to allow the JSON to specify key names. In fact, I am already using this feature, because simple_switch does not complain about the new fields, but other control plane implementations may need this information too.


antoninbas:
When you write "may potentially break other control planes that are not being tested", do you mean other control planes that operates on the bmv2 JSON? This would only affect applications consuming the bmv2 JSON, correct?

jafingerhut:
I have tried out these changes on a few of the example programs that were producing bad bmv2 JSON files before, and could not find one that still produces bad bmv2 JSON with these changes.

These changes do not appear to make any changes to p4runtime output that is incorrect, i.e. it is still incorrect, for example, when duplicate table names appear in both ingress and egress.  I guess that is expected with these changes, and correcting that would be saved for some future fix?

mbudiu-vmw:
This should handle duplicate table names too.

mbudiu-vmw:
@antoninbas : It's not about the JSON, I am playing with @name annotations, which influence the control-plane API generated, which is platform-independent.

backends/bmv2/backend.cpp
```diff
@@ -78,6 +79,86 @@ class ProcessControls : public BMV2::RemoveComplexExpressionsPolicy {
     }
 };
 
+/**
+This pass adds @name annotations to all fields of the user metadata
+structure so that they do not clash with fields of the headers
+structure.  This is necessary because both of them become global
+objects in the output json.
+*/
```
jafingerhut: Is this change specific to the one user metadata structure and the one headers structure in v1model.p4?

PSA adds half a dozen more top level parameters for top level controls that may be structs, all of which might have identical member names.  I guess this method can be extended for any number of such 'top level structs' ?

antoninbas:
The important thing to me is that the names match in the bmv2 JSON and in the file generated by the compiler for p4runtime.
As long as they still match after this PR is merged, it is fine with me.

jafingerhut:
@antoninbas I believe they mismatch with these changes.  The bmv2 JSON names for a table named 'foo' inside of ingress becomes 'cIngress.foo' in the bmv2 JSON, and for a table named 'foo' in egress becomes 'cEgress.foo' in the bmv2 JSON (assuming the program uses the control names cIngress and cEgress for those controls), but the p4runtime file still has 'foo' for both, and the same instance id for both.

mbudiu-vmw:
@jafingerhut : we don't have a back-end for PSA yet. The same trick should work, though, when targeting BMv2.

@antoninbas : I think that @jafingerhut may be right, because the PI names are generated early, before the hierarchical names are constructed. However, the names that were generated were in fact incorrect; we are only now starting to adhere with the P4-16 spec. However, my changes are all in the BMv2 back-end, so they happen long after the PI is generated.

The hierarchical names pass is in fact a mid-end pass, so we could call it earlier in the pipeline.

mbudiu-vmw:
This highlights a problem: we need to have some tests in the compiler tree for the PI.

antoninbas:
I just checked and the names are indeed inconsistent now. This is an issue as this breaks the P4 Runtime implementation for bmv2. This seems like a good opportunity to also fix the names in P4 Runtime and solve https://github.com/p4lang/p4c/issues/1073
The burden will be on the applications (P4 Runtime users) to make sure they use the correct fully-qualified names, or rely on the alias (https://github.com/p4lang/PI/blob/master/proto/p4/config/p4info.proto#L55).

antoninbas:
@mbudiu-vmw there are some tests for P4 Runtime (https://github.com/p4lang/p4c/blob/master/test/gtest/p4runtime.cpp) but they don't check consistency with the bmv2 JSON which makes sense to me
they use the old incorrect names (without top-level control name)

mbudiu-vmw:
@antoninbas : can you help create a test case to validate the PI API?

mbudiu-vmw:
It may be as easy as moving the hierarchicalNames pass earlier in the compiler pipeline.

mbudiu-vmw:
So the p4runtime tests are wrong - they should check for hierarchical names.
I will look into updating them.

mbudiu-vmw:
Frankly I don't understand how the P4 runtime API is generated. It should be done after inlining, because the actual name of various tables won't be known before inlining. But the bmv2 back-end seems to do it after the front-end has run. @sethfowler : can we move this after the mid-end, or in the middle of the mid-end?  Then we can stick the pass that changes the hierarchical names before the runtime API generation.

mbudiu-vmw:
We should probably merge this PR after #1116. Then we'll push the hierarchicalNames pass into the front-end too. And we'll finally be compliant with the P4-16 spec on naming...

mbudiu-vmw:
@antoninbas @sethfowler : I have also updated the p4runtime API generation code to use the hierarchical names.

For the ptf tests we should allow the test to use a short table name if it is unambiguous (e.g., instead of ingress.tbl use just tbl if there is no other table named like that). If we do that then we probably won't need to change too many tests.

antoninbas:
@mbudiu-vmw I agree with your PTF comment. However, this may not be supported by all P4 Runtime server implementations at this time, so I need to reach out to P4 API WG participants and we may need some time to transition.

antoninbas:
@mbudiu-vmw are you still planning to run the hierarchical names pass to the frontend as you mentioned in a comment above?

mbudiu-vmw:
OK, let me then try to see whether it is easy to move it to the front-end.
Keep watching this space. I have to solve the conflicts anyway.

mbudiu-vmw:
I have moved the pass to the frontend.
All reference outputs have changed, in their `@name` annotations.

antoninbas:
@mbudiu-vmw thanks I'll try it out tomorrow and get back to you

mbudiu-vmw:
There's something weird, since even ingress names are prefixed with egress.
Let me check.

mbudiu-vmw:
But other than that it's ready.

backends/bmv2/backend.cpp
mbudiu-vmw: We only care about structures which generate objects in the BMV2 toplevel namespace which can clash.

mbudiu-vmw:
This is ready for review again.

antoninbas:
@mbudiu-vmw sorry for the delay, I'm working on testing this today and fixing what it breaks.
Currently I'm running into the following issue. The P4Info message generated by the compiler has multiple entries for the `NoAction` action, one for each top-level control. For example:
```
actions {
  preamble {
    id: 16821110
    name: "MyIngress.NoAction"
    alias: "NoAction"
  }
}
// ...
actions {
  preamble {
    id: 16821111
    name: "MyEgress.NoAction"
    alias: "NoAction"
  }
}
```
Would it be possible to declare `NoAction` as a top-level action, rather than introduce one per-control?

mbudiu-vmw:
It is not a difficult problem to change core.p4.
But the more important issue is that this is actually a legal program. The compiler will clone NoAction for each control, and make it into a separate action. Then it may actually modify these actions independently (it doesn't now). So in practice one may have two actions with the same name in different controls.

antoninbas:
From core.p4:
```
/// Built-in action that does nothing.
@name("NoAction")
action NoAction() {}
```
 So it seems that there should be a single instance of `NoAction` in P4Info, and its name should be "NoAction".

The issue is not about whether it's a valid program or not. My tests assume that the action is called "NoAction" because that's what core.p4 tells me.

I have updated my test infrastructure to access unique suffixes (of the fully-qualified names) as valid identifiers for objects. This means that I don't have to change names in all of my-test once this PR gets merged. It seems that the only issue I'm running into now is with `NoAction`. All my tests refer to it as "NoAction" (which I think is correct based on core.p4). However according to the P4Info that is currently generated with this PR, "NoAction" is no longer a valid unique suffix because both "MyIngress.NoAction" and "MyEgress.NoAction" exist.

mbudiu-vmw:
This does not prevent a programmer from writing:
```
control ingress(...) {
   action NoAction() { ... }
}
```
So you cannot assume that there is only one action called "NoAction". The second one would be called "ingress.NoAction".

In our case the compiler builds ingress.NoAction, but it could as well come from the programmer. So the API must be able to handle two actions with the same name (suffix) in different scopes.

jafingerhut:
Not sure if I am capturing Antonin's thoughts correctly here, but I think his idea might be that it would be nice if the compiler didn't generation multiple instances of NoAction with different hierarchical names on its own, if the P4 program does not contain any occurrences of it.

I suspect he is probably OK with the output having multiple instance of an action, even one called NoAction, if the P4 program went out of its way to explicitly include actions with that name in the source code?

antoninbas:
Andy captured it pretty well.
The programmer can indeed do whatever he wants, _and the P4Info should reflect the P4 program_. Right now there is a single declaration of `NoAction` in the program (in core.p4), it is even annotated with `@name(NoAction)`, yet the compiler produces an incorrect P4Info with `MyIngress.NoAction` and `MyEgress.NoAction`. This doesn't reflect the P4 source.

If I write a the following P4 program:
```
#include <core.p4>
control ingress(...) {
   action NoAction() { ... }
}
```
the I would indeed expect 2 action "instances" in the P4Info, with "NoAction" as the last segment of the fully-qualified name: "NoAction" (from core.p4) and "ingress.NoAction" (from my "own" source).

jafingerhut:
Thinking about this a bit more, Antonin, I suppose you would also expect _any_ user-written actions that were at the top level of a P4_16 program, but used from tables in multiple control blocks, to have the same top-level name everywhere it was used, even though the default hierarchical names of the tables that use them would be prefixed with the control they were in (or even a sequence of controls that the object instantiation tree implied by the constructor calls in the source code).

For tables, they cannot be at the top level in P4_16, so they will always have at least one control name in their default hierarchical name (by "default" here I simply mean "the one assigned by the compiler in the absence of a `@name` annotation).

Some extern objects can also be instantiated at the top level in P4_16 (although a PSA implementation is currently spec'd in the latest draft as "need not be supported" -- a particular PSA implementation might choose to allow that, though).  For an implementation that allows such things, I suppose it would be nice if the control plane name was at the root level, no matter where the instance was later used?

mbudiu-vmw:
I believe that all problems stem from the fact that NoAction has a name annotation; let me try to remove that and see if things improve. This may change lots of intermediate results.
I don't remember why we added the annotation. If it is there it should be an absolute name `@name(".NoAction")`.

antoninbas:
@mbudiu-vmw I think you are right. It seems that NoAction is the only top-level action with this issue. If I define action "my_action" at the top-level and use it in both ingress & egress, the P4Info is what I would expect for my_action (single entry with name "my_action").
It's funny because I remember the annotation was introduced to solve a similar problem months ago (there were many NoAction entries in the P4Info, as NoAction_0, NoAction_1, etc.)
I believe adding a leading dot to the name annotation may have the same effect as removing it, but removing it is probably better. I haven't tested it though.

mbudiu-vmw:
I have amended the PR by removing the annotation of NoAction from core.p4.
I have to think a bit of what we should say in the spec - the name annotation is given as a tool to rename an object for the control-plane, but obviously, it does not always work as intended. In particular, maybe a global object should have a global name - prefixed by dot.

antoninbas:
@mbudiu-vmw can we move ahead and merge this?

mbudiu-vmw:
I certainly won't object to that!

antoninbas:
@mbudiu-vmw thanks again for working on this

mbudiu-vmw:
@antoninbas Thank you for the very thorough review: this issue touches the control-plane too, and I could not have done that myself.

PR #1108, opened 2017-12-19
Make v1model 'random' extern type generic

ChrisDodd:
- bmv2 allows any type up to the bmv2 word size
- makes it easier to convert/deal with programs that use random numbers
  of sizes other than 32 bits.

### DIFF
--- a/p4include/v1model.p4
+++ b/p4include/v1model.p4
@@ -104,7 +104,7 @@ extern action_profile {
 }
 
 // Get a random number in the range lo..hi
-extern void random(out bit<32> result, in bit<32> lo, in bit<32> hi);
+extern void random<T>(out T result, in T lo, in T hi);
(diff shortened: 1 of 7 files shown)


p4include/v1model.p4
```diff
@@ -104,7 +104,7 @@ extern action_profile {
 }
 
 // Get a random number in the range lo..hi
-extern void random(out bit<32> result, in bit<32> lo, in bit<32> hi);
+extern void random<T>(out T result, in T lo, in T hi);
```
mbudiu-vmw: This is more convenient for software switches, but is it good for hardware switches?

ChrisDodd: Hardware switches may require that lo == 0 and hi == one less than a power of 2, but that's not really affected by this change.

Issue #1107, opened 2017-12-19
Compiler Bug: typechecker mutated program


antoninbas:
I'm hitting a pretty strange Compiler Bug with all backends, including p4test.
The following program (also attached as a file) triggers this compiler bug:
```
In file: /home/antonin/Documents/p4lang/p4c/frontends/p4/typeChecking/typeChecker.cpp:135
Compiler Bug: typechecker mutated program
```

The source:
```
#include <core.p4>
#include <v1model.p4>

struct H { };
struct M {
    bit<32> f1;
    bit<32> f2;
}

typedef bit<32> IPAddr;
const IPAddr MyIP = 0xffffffff;

// if the 2 lines above are replaced by this one, program compiles
// const bit<32> MyIP = 0xffffffff;

parser ParserI(packet_in pk, out H hdr, inout M meta, inout standard_metadata_t smeta) {
    state start { transition accept; }
}

action empty() { }

control myc(inout H hdr, inout M meta, inout standard_metadata_t smeta) {
    action set_eg(bit<9> eg) { smeta.egress_spec = eg; }

    table myt {
        key = {
            meta.f1 : exact;
            meta.f2 : exact;
        }
        actions = { set_eg; }
        const entries = {
            // if one entry instead of 2, program compiles
            (32w1, MyIP) : set_eg(9w1);
            (32w2, MyIP) : set_eg(9w2);
        }
    }

    apply {
        myt.apply();
    }
}

control IngressI(inout H hdr, inout M meta, inout standard_metadata_t smeta) {
    apply {
        myc.apply(hdr, meta, smeta);
    }
};

control EgressI(inout H hdr, inout M meta, inout standard_metadata_t smeta) {
    apply { }
};

control DeparserI(packet_out pk, in H hdr) {
    apply { }
}

control VerifyChecksumI(inout H hdr, inout M meta) {
    apply { }
}

control ComputeChecksumI(inout H hdr, inout M meta) {
    apply { }
}

V1Switch(ParserI(), VerifyChecksumI(), IngressI(), EgressI(),
         ComputeChecksumI(), DeparserI()) main;
```

[type_checker_mutated_program.p4.txt](https://github.com/p4lang/p4c/files/1570063/type_checker_mutated_program.p4.txt)

Some minor changes make the program compile fine. In particular, if I do either of the following:
- remove the typedef and use `bit<32>` directly
- provide a single const entry instead of 2 in my table

mbudiu-vmw:
[...]

PR #1111, opened 2017-12-20
Fix for issue #1107

mbudiu-vmw:
There were two subtle bugs in constant folding
- the constant declaration value was substituted, but the type in the constant declaration was ignored.
- cloning an Type_InfInt does not produce a new Type_InfInt, just another copy of the same type. Here we really wanted a new type, which can be bound to a different value when performing unification.

### DIFF
--- a/testdata/p4_16_samples/issue1107.p4
+++ b/testdata/p4_16_samples/issue1107.p4
@@ -10,9 +10,6 @@ struct M {
 typedef bit<32> IPAddr;
 const IPAddr MyIP = 0xffffffff;
 
-// if the 2 lines above are replaced by this one, program compiles
-// const bit<32> MyIP = 0xffffffff;
-
 parser ParserI(packet_in pk, out H hdr, inout M meta, inout standard_metadata_t smeta) {
     state start { transition accept; }
 }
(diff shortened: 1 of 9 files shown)


testdata/p4_16_samples/issue1107.p4
```diff
@@ -0,0 +1,66 @@
...
+struct H { };
+struct M {
+    bit<32> f1;
+    bit<32> f2;
+}
+
+typedef bit<32> IPAddr;
+const IPAddr MyIP = 0xffffffff;
+
+// if the 2 lines above are replaced by this one, program compiles
```
antoninbas: Nit: remove comment

testdata/p4_16_samples/issue1107.p4
```diff
@@ -0,0 +1,66 @@
...
+    action set_eg(bit<9> eg) { smeta.egress_spec = eg; }
+
+    table myt {
+        key = {
+            meta.f1 : exact;
+            meta.f2 : exact;
+        }
+        actions = { set_eg; }
+        const entries = {
+            // if one entry instead of 2, program compiles
```
antoninbas: Nit: remove comment

PR #1116, opened 2017-12-21
Move inlining to the front-end

mbudiu-vmw:
This PR just shuffles code around.
I have moved about 10 passes, including inlining, from the mid-end to the front-end. In consequence the 3 back-ends do not need them anymore, and the control-plane API generation can be much simpler. I have not pushed a corresponding PR for the control-plane repo yet, it is very simple.

It looks like inlining is useful for all back-ends. Moreover, the API generation must be done after inlining. This pushes the API generation to be done almost immediately after the front-end. 

Unfortunately many of the reference outputs have changed.

I have also sorted alphabetically some lists of included files and makefiles.

The file removeReturns has been split into two files: removeReturns, which is now in the front-end, and removeExits, which is still in the mid-end.


mbudiu-vmw:
If we accept this change the output of the p4c-graphs will also change - it will show the graph after inlining. It is not clear whether this is desirable or not.

jafingerhut:
I haven't looked at the code of these changes, and won't comment on that.

I have tried out p4c-graphs before and after these changes, and while they are different, if anything things look a bit more compact and readable with these changes vs. before these changes.

e.g. Before these changes, there are boxes drawn around sub-controls called from ingress or egress, and sub-tables have their names prefixed with control names.

After these changes, there are no boxes around sub-controls, and names of tables do not have control names prefixing them.

ChrisDodd:
It seems to me that doing API generation AFTER inlining is a bug, albeit one that some example programs are dependent on, which makes it tough to fix.  The API names (IMO) should have the fully qualified names, and not the post-inlining names, but it looks like I'm outvoted on this.

I can conceive of systems where not inlining would be desirable (ie, CPU based software which real function calls available), but we don't currently have any such backends, and it looks like they would require inlining anyway just to get the API right.

mbudiu-vmw:
The fully qualified names before and after inlining should produce the same result. The p4-16 spec rules for name generation satisfy these requirements.

jafingerhut:
@mbudiu-vmw I am not sure, but it sounds like you may be saying that it should be possible to generate correct hierarchical P4Runtime API names _without_ doing the compiler's inlining step first, by using the fully qualified names?

mbudiu-vmw:
Yes, the names that are generated for the final program should not depend on when and where the inlining pass is executed. The current implementation for generating the names in fact ran inlining by itself too.

PR #1117, opened 2017-12-22
remove alias fields from standard_metadata when translating p4-14

hanw:
This is a fix for #718 and #1036.

The fix is to remove the fields in v1model.p4 standard_metadata_t that are marked with `@alias`, which is only supposed to be used in P4-16 programs. If a P4-16 program is generated from a P4-14 program, these annotated fields should be removed, so they don't show up in the JSON file.

### DIFF
--- a/frontends/p4/fromv1.0/converters.cpp
+++ b/frontends/p4/fromv1.0/converters.cpp
@@ -1034,7 +1034,7 @@ class RemoveAnnotatedFields : public Transform {
                     fields->push_back(f);
                 }
             }
-            return new IR::Type_Struct(node->srcInfo, node->name, *fields);
+            return new IR::Type_Struct(node->srcInfo, node->name, node->annotations, *fields);
         }
         return node;
     }
--- a/frontends/p4/fromv1.0/converters.cpp
+++ b/frontends/p4/fromv1.0/converters.cpp
@@ -1020,6 +1020,26 @@ class DetectDuplicates: public Inspector {
     }
 };
 
+// The fields in standard_metadata in v1model.p4 should only be used if
+// the source program is written in P4-16. Therefore we remove those
+// fields from the translated P4-14 program.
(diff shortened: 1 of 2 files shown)


mbudiu-vmw:
I don't understand what problem this PR is solving.
It would be nice to address issue #718, which is tied to aliased fields.
Personally, I don't understand what aliased fields are, so I don't know how to address that issue.

hanw:
This part of v1model standard_metadata was added to support additional intrinsic metadata in bmv2 simple_switch in P4-16 programs.
```
    @alias("queueing_metadata.enq_timestamp") bit<32> enq_timestamp;
    @alias("queueing_metadata.enq_qdepth")    bit<19> enq_qdepth;
    @alias("queueing_metadata.deq_timedelta") bit<32> deq_timedelta;
    @alias("queueing_metadata.deq_qdepth")    bit<19> deq_qdepth;
    // intrinsic metadata
    @alias("intrinsic_metadata.ingress_global_timestamp") bit<48> ingress_global_timestamp;
    @alias("intrinsic_metadata.lf_field_list") bit<32> lf_field_list;
    @alias("intrinsic_metadata.resubmit_flag") bit<1>  resubmit_flag;
```
The documentation for these intrinsic metadata is in `https://github.com/p4lang/behavioral-model/blob/master/docs/simple_switch.md`. 

However, these metadata are not needed for P4-14 or a P4-16 program that is translated from P4-14. This PR basically removed these field from standard_metadata during the P4-14 to 16 translation. It should fix issue #718, but I haven't been able to get STF to work with multicast.

hanw:
Alias is a way to tell bmv2 backend that value assigned to say, `std_meta.mcast_grp` should be used as the value for `intrinsic_metadata.mcast_grp`.

mbudiu-vmw:
How is multicast supposed to work in P4-16?
Apparently if you strip out the alias fields multicast works in P4-16.

PR #1126, opened 2017-12-30
fixes for P4-14 to 16 register extern translation

hanw:
Global registers should remain global after 14-to-16 translation,
because multiple control blocks may share the same register. Creating
an instance in the local scope of the control block results in
duplicated register instance after inlining.

### DIFF
--- a/frontends/p4/fromv1.0/converters.h
+++ b/frontends/p4/fromv1.0/converters.h
@@ -89,7 +89,8 @@ class ExternConverter {
     virtual const IR::Type_Extern *convertExternType(ProgramStructure *,
                 const IR::Type_Extern *, cstring);
     virtual const IR::Declaration_Instance *convertExternInstance(ProgramStructure *,
-                const IR::Declaration_Instance *, cstring, IR::IndexedVector<IR::Declaration> *);
+                const IR::Declaration_Instance *, cstring, IR::IndexedVector<IR::Declaration> *,
+                IR::IndexedVector<IR::Node> *);
     virtual const IR::Statement *convertExternCall(ProgramStructure *,
                 const IR::Declaration_Instance *, const IR::Primitive *);
     virtual bool convertAsGlobal(ProgramStructure *, const IR::Declaration_Instance *) {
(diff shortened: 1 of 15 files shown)


frontends/p4/fromv1.0/converters.h
```diff
@@ -89,7 +89,8 @@ class ExternConverter {
     virtual const IR::Type_Extern *convertExternType(ProgramStructure *,
                 const IR::Type_Extern *, cstring);
     virtual const IR::Declaration_Instance *convertExternInstance(ProgramStructure *,
-                const IR::Declaration_Instance *, cstring, IR::IndexedVector<IR::Declaration> *);
+                const IR::Declaration_Instance *, cstring, IR::IndexedVector<IR::Declaration> *,
+                IR::IndexedVector<IR::Node> *);
```
ChrisDodd: should have comments or names here to make clear which argument is the local scape and which is the global scope.

frontends/p4/fromv1.0/converters.h
```diff
@@ -107,8 +108,9 @@ class ExternConverter {
         return get(e)->convertExternType(s, e, name); }
     static const IR::Declaration_Instance *cvtExternInstance(ProgramStructure *s,
                 const IR::Declaration_Instance *di, cstring name,
-                IR::IndexedVector<IR::Declaration> *scope) {
-        return get(di)->convertExternInstance(s, di, name, scope); }
+                IR::IndexedVector<IR::Declaration> *local_scope,
+                IR::IndexedVector<IR::Node> *global_scope) {
```
ChrisDodd: will `global_scope` always be the same as `s->declarations`?  If so we don't really need it as an extra argument.  If not, what is it?

frontends/p4/fromv1.0/programStructure.cpp
```diff
@@ -1911,8 +1914,9 @@ void ProgramStructure::createControls() {
         if (ExternConverter::cvtAsGlobal(this, ext.first)) {
             // FIXME -- 'declarations' is a vector Nodes instead of Declarations.  Should
             // FIXME -- fix the IR to use a NameMap for scopes anyways.
-            IR::IndexedVector<IR::Declaration> tmpscope;
-            auto e = ExternConverter::cvtExternInstance(this, ext.first, ext.second, &tmpscope);
+            IR::IndexedVector<IR::Node> tmpscope;
+            auto e = ExternConverter::cvtExternInstance(this, ext.first, ext.second,
+                                                        nullptr, &tmpscope);
```
ChrisDodd: Is passing a nullptr for the local scope argument safe?  That's why the code previously created tmpscope.  Similarly, for the global scope, should just pass `declarations` directly?

ChrisDodd:
Actually, is this change even required at all?  `ExternConverter:cvtExternInstance` already has access to the global scope as `structure->declarations`...

hanw:
Now I think about it, this change may not be required at all. =)

frontends/p4/fromv1.0/programStructure.cpp
```diff
@@ -1830,13 +1830,16 @@ ProgramStructure::convertControl(const IR::V1Control* control, cstring newName)
     for (auto c : registersToDo) {
         auto reg = registers.get(c);
         auto r = convert(reg, registers.get(reg));
-        stateful.push_back(r);
+        if (!declarations->getDeclaration(r->name)) {
+            declarations->push_back(r);
+        }
```
hanw: Probably the only needed change is here.

Issue #1127, opened 2018-01-06
Internal bmv2 backend 'Duplicate label' error for legal P4_16 program


jafingerhut:
[p4c-issue1127.p4.txt](https://github.com/p4lang/p4c/files/1608273/p4c-issue1127.p4.txt)

At least, I cannot see anything illegal about the P4_16 program I am attempting to compile that causes this error.  See the program in the attached file p4c-issue1127.p4.txt.  When I attempt to compile it with the following command line, using the latest version of p4c-bm2-ss as of 2018-Jan-05, it gives the following error:
```
% p4c-bm2-ss p4c-issue1127.p4.txt
terminate called after throwing an instance of 'std::logic_error'
  what():  Duplicate label in json object true_next
Aborted (core dumped)
```
There are 3 comments in the program showing small changes one can make to the program, either commenting out or uncommenting a few lines, that give a variant of the program for which no error message occurs from the compiler (I haven't verified whether the bmv2 JSON file is correct for those variations -- simply that the compiler gives no error message)

mbudiu-vmw:
Seems to be a bug in the control-flow graph computation.

jafingerhut:
With the latest p4c change, the compiler error is gone.  From a brief look, it appears that the generated BMv2 JSON file is correct, too.  Thanks for the fix.

Issue #1128, opened 2018-01-06
Bug in losing 'next' field of header stack during sub-parser invocation


jafingerhut:
I don't have a test case for this right now.  This is more of a case of seeing a bug in one place, where probably soon it will be fixed, but wondering whether a similar bug may also exist in another place.

Background: I am writing a P4_16 test program to test the new P4_16 header stack support made by this behavioral-model PR: https://github.com/p4lang/behavioral-model/pull/515

I started by writing a program and testing with those new changes, and then I wanted to verify that some of those tests, ones that should fail with the old/current behavior of 'push', actually do fail.

My P4_16 program has an ingress control block that passes all headers to a sub-control, modifying the headers there, then returning to the ingress control.  The sub-control does P4_16 push_front and pop_front() method calls on a header stack, passed as a field of a 'struct headers' type that is commonly seen in P4_16 programs.

The BMv2 JSON file compiled from this source code has 2 header stacks -- one for the 'global' one in the ingress control block, and another temporary one for the sub-control.  The JSON file has actions that when calling the sub-control, copies the elements of the size-5 header stack via 5 'assign_header' primitive actions from the global header stack to the temporary one, then copies them back out when returning.

That is correct for getting most of the state of the header stack from the ingress control to the sub-control being called, but it _does not_ cause the 'next' state of the header stack to be copied in and out on these calls to the sub-control.

Now maybe this is soon to be a historical footnote for control blocks, because with the new behavior of push and pop primitives, they will no longer depend upon the value of 'next' in a header stack.

However, if the compiler can generate similar BMv2 JSON code for a parser calling a sub-parser, then it is possible it can generate code that does not behave as the source code specifies, because of losing the 'next' state.

jafingerhut:
This is likely only an issue for P4_16 programs, since I am guessing that for P4_14 programs having all headers and header stacks global leaves no reason for the compiler to add copying of one header stack to another.

I don't think P4_14 allows copying one header stack variable to another, but P4_16 does, and if p4c implements such copying in a parser, that might be another situation where incorrect behavior could occur due to not copying the 'next' state from one header stack to another.  I haven't tested that.

jafingerhut:
[...]

PR #1129, opened 2018-01-06
Add STF tests of P4_16 header stack operation behavior

jafingerhut:
I have tested this with the version of p4lang/behavioral-model after this commit: https://github.com/p4lang/behavioral-model/commit/5d3e2f2bbeb4434941236712847dce0f62c9b398

and the tests only produce the expected results, which I believe are correct according to the P4_16 language spec, if you build simple_switch after running configure for behavioral-model with this option:
```
./configure --enable-WP4-16-stacks
```


antoninbas:
One more reason to wait for 1.11: the bmv2 docker image which is used for running the Travis CI tests doesn't use `--enable-WP4-16-stacks` (since it is not the default yet).

jafingerhut:
It seems reasonable to me to wait until behavioral-model release 1.11.0, when these tests should pass with the default build options for behavioral-model, before merging this new test in.

antoninbas:
Maybe this test should be merged and tagged as XFAIL? Then we will just need to remove it from the XFAIL list when bmv2 1.11 is released.

jafingerhut:
Tried adding this new test to XFAIL list, but Travis testing still runs it.  Maybe I didn't mark the test XFAIL correctly?  If I did it correctly, then something weird is going on to cause Travis to run the test anyway.

mbudiu-vmw:
The test should be run, but it should not cause a failure.

mbudiu-vmw:
There are two tests run for each p4 file: one with p4test and one with bmv2.
You have added the xfail to the wrong makefile.
You have to use the backends/bmv2 makefile.

jafingerhut:
@mbudiu-vmw "The test should be run, but it should not cause a failure."

If you meant "I've looked at the new test, and it should pass with the latest versions of p4c and behavioral-model code", then I disagree.  Some of packet-in/packet-out behavior relies upon changes to behavioral-model in how P4_16 push and pop primitives are implemented, that are not planned to be the default behavior until behavioral-model 1.11 is released in the future.

If you meant "We should not merge in changes that cause 'make check' to fail", then we are in agreement, hence marking the new test XFAIL.

jafingerhut:
@mbudiu-vmw I modified the test P4_16 program so that it no longer works around already-fixed issue #1127.

jafingerhut:
If the new behavioral-model 11.1 will avoid all of the failures, what benefit is there in splitting up the test cases?

It might be possible to write a different P4_16 test program that exhibits the issue #1128, with or without the changes coming in behavioral-model release 11.1, i.e. a program that has a parser calling a sub-parser, passing a header stack as an inout parameter.

jafingerhut:
I have created an STF test case in a PR that I believe demonstrates issue #1128, and its behavior will not be affected by releasing behavioral-model 1.11: https://github.com/p4lang/p4c/pull/1132

mbudiu-vmw:
If we merge this the build will (hopefully) fail when the bmv2 simulator is updated.
What should be do?

jafingerhut:
I am perfectly happy to wait to merge this after behavioral-model 1.11 is released, and change this PR at that time to remove the new test from XFAIL, since it should then pass.

mbudiu-vmw:
Yes, and your example from https://github.com/p4lang/p4c/pull/1132 will not generate legal JSON anymore with my recent PR: https://github.com/p4lang/p4c/pull/1133. We have to wait for support from bmv2.

mbudiu-vmw:
We should merge this after #1144.

antoninbas:
bmv2 1.11.0 has been released and `--enable-WP4-16-stacks` is now the default.
It will take some time for the new bmv2 docker image to build, but after that the Travis regression for this PR should be restarted - and hopefully it will pass.

antoninbas:
Looks like the new docker image has been built and uploaded to docker hub. I'm restarting the Travis job now.

antoninbas:
@jafingerhut @mbudiu-vmw the test is passing, can we merge?

mbudiu-vmw:
This is great!
Thank you.

jafingerhut:
@antoninbas Good to see that these tests are passing, as expected, and yes they are good to merge.

PR #1130, opened 2018-01-08
Fix for issue #1127

mbudiu-vmw:
I will lose all my self-respect one of these days.
How could I mess up so badly such a simple piece of code? I could not even read what I had written...
Hopefully it's much simpler now.


jafingerhut:
@mbudiu-vmw Don't lose your self respect over this!  I don't know when I started doing this, but when I find bugs I've written that take a bit of debugging, when I finally understand the root cause I often laugh out loud (then fix the bug).  Perhaps a sign of impending madness.  Not sure :-)

PR #1132, opened 2018-01-09
Add test case that fails due to issue #1128

jafingerhut:
I believe that the bug demonstrated here will not be fixed, even when
behavioral-model 1.11 changes the behavior of push/pop primitives for
P4_16 header stack behavior, because this program does not use those
primitives at all.


mbudiu-vmw:
I haven't merged this yet because I hope to push a fix first.

mbudiu-vmw:
Once #1144 is merged this PR should be closed.

jafingerhut:
#1144 is merged now, and includes the test cases added with this PR, so closing this one as redundant.

PR #1133, opened 2018-01-09
Remove incorrect expansion of array copy

mbudiu-vmw:
As @jafingerhut pointed out, the removed code was incorrect, since it was assuming that copying all elements of an array is the same as copying the array. This would actually be correct within a control block only, where the .next and .last fields are not accessible. But that code was not really necessary for removing nested structures, which is the point of copyStructures. I don't think this is enough to solve issue #1128.


mbudiu-vmw:
No one likes my little PR?

PR #1134, opened 2018-01-12
swap precedence of bitwise and relational operators in P4_14 parser

ChrisDodd:
Despite the fact that the P4_14 spec says that expressions have "syntax, semantics, precedence, and associtivity matching that of C", if you look closely at th egrammar in the spec, you find that the precedence of bitwise operators (`&`, `|`, and `^`) has been swapped with that of relations (`==`, `!=`, `<`, `<=`, `>`, `>=`).  The old p4-hlir (python) frontend follows the precedence as specified in the grammar, giving the bitwise operators higher precedence.

Turns out there are existing P4_14 programs that depend on this, and they fail when using our frontend.  In the interest of backwards compatibility, this change swaps the precedence to match the spec grammar and the old implementation.

Arguably, the precedence in C is a mistake -- a historical accident due to the language originally conflating bitwise and logical operations.  Changing this in P4_16 is a discussion for the language committee.

### DIFF
--- a/frontends/parsers/v1/v1parser.ypp
+++ b/frontends/parsers/v1/v1parser.ypp
@@ -177,11 +177,11 @@ typedef const IR::Type ConstType;
 %right ASSIGN
 %left OR
 %left AND
+%left EQ NE
+%left L_ANGLE R_ANGLE LE GE
 %left BIT_OR
 %left BIT_XOR
 %left BIT_AND
-%left EQ NE
-%left L_ANGLE R_ANGLE LE GE
 %left SHL SHR
 %left PLUS MINUS
 %left MUL DIV MOD


mbudiu-vmw:
I don't know how this happened, i thought i had copied that part of the grammar from the c grammar.

ChrisDodd:
The P4_16 grammar does in fact copy the C grammar and has the same precedence as C.  This is a difference between p4_14 and p4_16 specs, and means that code like:

    if (hdr.field & 0xf == 1) {

has different semantics in P4_14 and P4_16.  Its not a huge problem, as P4_16 will give you a type error for the above; you need to add parentheses to make it equivalent to the P4_14 code and fix the type error.

cc10512:
I would argue that we should not repeat C's mistakes and keep it uniform with P4_14, so we should bring it up in the LDWG for a P4_16 spec change.

PR #1138, opened 2018-01-18
make strength reduction pass in frontend optional

hanw:
strength reduction causes an issue with certain midend when the midend is trying to perform an architecture translation. In particular, to translate the following expression.

```
standard_metadata.packet_length - 14 + 1 
```

to another architecture with meta.pkt_length of type bit<16>.

```
meta.pkt_length - 14 + 1
```

If the strength reduction is done at the frontend, then the transformation in frontend goes as follows.
```
standard_metadata.packet_length - 14 + 1
```
is first type inferred to 
```
standard_metadata.packet_length - 32w14 + 32w1
```
then, it is strength reduced to
```
standard_metadata.packet_length + 32w4294967282 + 32w1
```

However, if the target architecture contains a pkt_length metadata that has a different bit width, say bit<16>, the translation needs to go as follows.
```
standard_metadata.packet_length + 32w4294967282 + 32w1
```
is first translated to 
```
meta.pkt_length + 32w4294967282 + 32w1
```
then, the translation has to fix the type of the constants by inserting casts for all constants in an arbitrary expression.
```
meta.pkt_length + (bit<16>)32w4294967282 + (bit<16)32w1
```

Alternatively, an easier way to do the translation is to disable the strength reduction in the frontend. 
```
meta.pkt_length  - 32w14 + 32w1
```
converts the constants to InfInt
```
meta.pkt_length  - 14 + 1
```
then rerun typeinference
```
meta.pkt_length  - 16w14 + 16w1
```
and then strengthReduction
```
meta.pkt_length  + 16w65522 + 16w1
```

Making the StrengthReduction pass optional will make certain midend easier to write.

### DIFF
--- a/frontends/p4/frontend.cpp
+++ b/frontends/p4/frontend.cpp
@@ -106,7 +106,8 @@ class FrontEndDump : public PassManager {
 
 // TODO: remove skipSideEffectOrdering flag
 const IR::P4Program *FrontEnd::run(const CompilerOptions &options, const IR::P4Program* program,
-                                   bool skipSideEffectOrdering) {
+                                   bool skipSideEffectOrdering,
+                                   bool skipStrengthReduction) {
     if (program == nullptr)
         return nullptr;
 
@@ -140,6 +141,7 @@ const IR::P4Program *FrontEnd::run(const CompilerOptions &options, const IR::P4P
         new ClearTypeMap(&typeMap),
         new TableKeyNames(&refMap, &typeMap),
         new ConstantFolding(&refMap, &typeMap),
+        new StrengthReduction(skipStrengthReduction),
         new UselessCasts(&refMap, &typeMap),
         new SimplifyControlFlow(&refMap, &typeMap),
         new FrontEndDump(),  // used for testing the program at this point
--- a/frontends/p4/frontend.cpp
+++ b/frontends/p4/frontend.cpp
@@ -140,7 +140,6 @@ const IR::P4Program *FrontEnd::run(const CompilerOptions &options, const IR::P4P
         new ClearTypeMap(&typeMap),
         new TableKeyNames(&refMap, &typeMap),
         new ConstantFolding(&refMap, &typeMap),
-        new StrengthReduction(),
         new UselessCasts(&refMap, &typeMap),
         new SimplifyControlFlow(&refMap, &typeMap),
         new FrontEndDump(),  // used for testing the program at this point
--- a/frontends/p4/frontend.h
+++ b/frontends/p4/frontend.h
@@ -29,7 +29,8 @@ class FrontEnd {
     explicit FrontEnd(DebugHook hook) { hooks.push_back(hook); }
     void addDebugHook(DebugHook hook) { hooks.push_back(hook); }
     const IR::P4Program* run(const CompilerOptions& options, const IR::P4Program* program,
-                             bool skipSideEffectOrdering = false);
+                             bool skipSideEffectOrdering = false,
+                             bool skipStrengthReduction = false);
 };
 
 }  // namespace P4
--- a/frontends/p4/strengthReduction.cpp
+++ b/frontends/p4/strengthReduction.cpp
@@ -20,34 +20,34 @@ namespace P4 {
 
 /// @section Helper methods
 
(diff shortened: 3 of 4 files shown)


mbudiu-vmw:
I don't understand what the problem is that you are addressing.
In general, all passes in the front-end should be completely architecture-independent. If that is not true, then we have a significant bug.
Moreover, if the front-end emits a program to the mid-end, then there is also a program that the user could have written that could produce the same result. In general, the mid-end should be prepared to accept any program that comes from the front-end, because they all are legal programs.
The need to disable passes in the front-end suggests that something is seriously wrong. So we need to get to the bottom of this issue.

The type of the packet length is part of core.p4, which is part of the P4 spec. The mid-end cannot suddenly declare that the type of an object is different from source-code type declaration. If you do that you should expect that everything can break, not just strength-reduction. None of the guarantees provided by a type system can be assumed valid anymore.

The fundamental issue that you are describing seems to be that core.p4 is wrong, and removing strength reduction is the wrong solution. The right solution is to take this to the P4 design group, and amend the design, by removing the length method from the packet_in extern, or making packet_in be architecture-dependent rather than being in core.p4.

jnfoster:
I may not have all the context, but I agree with Mihai.

all passes in the front-end should be completely architecture-independent.
> If that is not true, then we have a significant bug.


-N

hanw:
When a P4-14 program is translated to P4-16 at the beginning of the frontend, it assumes v1model. In v1model, the type of `packet_length` field is bit<32>
```
struct standard_metadata_t {
   ....
   bit<32> packet_length;
   ....
}
```

Now, if the P4-14 program uses the `packet_length` in an expression, for example, `standard_metadata.packet_length - 14 + 1 `, the end result after strength reduction is `standard_metadata.packet_length + 32w4294967282 + 32w1`. So even though strength reduction by itself is architecture independent, the result it produces actually depends on the architecture of the source program. 

Now suppose I would like to run the original P4-14 program to a different architecture that has a different type for packet length, say bit<16>, say, if the expression with standard_metadata.packet_length is translated to use intrinsic_metadata.pkt_length.

```
struct intrinsic_metadata_t {
   ....
   bit<16> pkt_length;
   ....
}
```

If I do such translation in the midend, the input to the translation pass is going to be:
`standard_metadata.packet_length + 32w4294967282 + 32w1`, which assumed v1model.

I would rather have `standard_metadata.packet_length - 14 + 1 ` as input and the translation would be a lot simpler: rename, type inference, and then strength reduction.

mbudiu-vmw:
The problem is not the compiler, the problem is v1model.p4 - it does not describe your architecture.
You have to use an appropriate model. Removing strength reduction will not fix your model.
The front-end does not make any assumptions about the model, only about core.p4.
In core p4 packet_in has a length field which is defined as bit<32>, but if you don't use that then it's not a problem. Only the P4-14 parser depends on v1model to some degree.

The simplest solution is probably to pass a -I flag to the compiler to direct it to load your custom modified v1model.p4 file instead of the standard one.

hanw:
The simplest solution is probably to update v1model.p4 to use bit<16> for packet_length, if there is no objection. BMV2 does not care the bit width of packet_length, and I am not aware of any other backend that mandate packet_length to be bit<32>. Neither do I understand why it is bit<32>, as the P4-14 spec does not specify the width.

mbudiu-vmw:
Why should it be 16 bits if P4-14 does not specify the width?
You can always stuff a small number into a large number of bits, but it does not work in reverse.
If you ever need to express a packet length above 16 bits you are dead. So 32 bits is much safer than 16 bits.

jnfoster:
"65K ought to be enough for anybody." -Han Wang, with apologies to Bill
Gates :-)

jnfoster:
Another possible solution could be to write a pass that implements
coercions to/from the type of standard_metadata.length in v1model (say S)
and the type in your architecture (say T).

In any well-typed source program, all expressions involving
standard_metadata.length will be of type S. So all arithmetic would be
carried out at that width and then converted when copying in to/out of that
field.

Now, these coercions would not preserve information if S and T are not
identical. But the compiler could issue a warning.

-N

hanw:
Thanks Mihai and Nate. This PR is no longer required.

PR #1145, opened 2018-01-24
Implement equality testing for structures

mbudiu-vmw:
This PR implements structure/header/array equality testing by lowering equality/inequality to a recursive comparison of all fields. The test does not completely succeed, probably because of issue https://github.com/p4lang/behavioral-model/issues/536 -- varbit comparison in the behavioral simulator.


mbudiu-vmw:
This is now ready for review. It implements the latest changes to the spec, which define the equality of various nested structures.

mbudiu-vmw:
It looks to me like we are running out of reviewers for PRs.

PR #1148, opened 2018-01-27
Use 1LL << 32, instead of 1U << 32

hanw:
1U << 32 is 0, we should use 1LL << 32.

Also, the BUG_CHECK condition is reversed, we were lucky to not hit on that check.


jafingerhut:
Is the p4c midend always converting enum values into 1-hot encoded bit vectors with N bits, not ids with lg(N) bits, where N is the number of possible values of the enum type?

hanw:
As far as I can see, the encoding in the converEnums is not one-hot.

Issue #122, opened 2016-10-28
Compilers do not handle side-effects in select expression labels


mbudiu-vmw:
The spec allows side-effects in select labels, but all our examples use constant labels.
Handling multiple side-effects in the right order will require some compiler support.

PR #1152, opened 2018-01-30
Check for constant arguments, as required by spec

mbudiu-vmw:


mbudiu-vmw:
What does one do when Travis gets stuck?

mbudiu-vmw:
Actually the spec says this:

> Some targets may require that all keyset expressions in a select expression be compile-time known values.

So the front-end should not check for the labels.

hanw:
Does that mean current spec allow any expression to be used in a selection expression?

mbudiu-vmw:
In the "keysets" of a select expression: yes.
But various back-ends may reject non-constant keysets.
Frankly, they are a pain, and the compiler does not support them correctly; technically they can be any expression, and the semantics requires these expressions to be evaluated only when necessary (when all previous cases have failed). Our sideEffects pass should handle that, but it currently does not.

mbudiu-vmw:
BTW: we have an issue open for this: #122

jafingerhut:
@mbudiu-vmw Don't be surprised if a P4_16 language work group meeting in the near future includes suggestions for, or questions about, using a keyset expression that is a run-time variable, as a more P4_14-like syntax for using parser value sets in PSA.

I have seen lots of the discussion on multiple Github issues on this controversial issue, and we don't have to turn this one into yet another place to discuss it, but wanted to mention it, since at least at this time, the P4_16 language spec doesn't actually prevent the keyset expressions from being run-time variables.

mbudiu-vmw:
I believe that my latest proposal for handling value_sets satisfies all the requirements you have. You should take a look; it is very simple.

PR #1155, opened 2018-02-06
Dot is not legal in an identifier in C

mbudiu-vmw:


mbudiu-vmw:
Could we get a quick review for this one?
It's breaking some builds for external projects...
And it is a bug.

PR #1159, opened 2018-02-09
Remove '$' from indexed names in stf

amreshk:


antoninbas:
Also I don't which problem you are trying to solve but I would say changing the STF format may not be the best or only solution. Isn't the issue with a specific STF interpreter / runner you are using? It seems that if the STF interpreter for your target doesn't support '$', then it should be its job to perform the substitution?

ChrisDodd:
This isn't really changing the stf -- its just changing the names of headers in the stf to match the changed names in the context.json, which were changed to make PD generation happy.  But you should certainly check to make sure that this doesn't break the bmv2 version of the test before checking in.

amreshk:
I am holding off on this for now and see if we can solve it in another way.

amreshk:
Closing this PR, dont need this change anymore.

PR #1161, opened 2018-02-14
P4runtime support is const table in p4info generation

antoninbas:


test/gtest/p4runtime.cpp
```diff
@@ -677,4 +678,54 @@ TEST_F(P4Runtime, StaticTableEntries) {
...
+                actions = { a; a_with_control_params; }
+                default_action = a;
+                const entries = {
+                    (0x01) : a_with_control_params(1);
+                }
+            }
+            table t_non_const {
+                key = { h.h.hfA : exact; }
+                actions = { a; a_with_control_params; }
+                default_action = a;
```
mbudiu-vmw: The P4-16 spec says:  Future versions of P4 may add the ability to mix mutable and immutable entries in the same table, by declaring additional entries properties without the const keyword. Are you prepared for that case?

antoninbas: That's a good point. We currently disallow that scenario - which is consistent with the language spec - and the "is_const_table" attribute in P4Info makes sense and is convenient, as it lets the application know whether are not it has write access to the table.

When - and if - we support mutable entries, this attribute will probably not make sense any more. However, I believe this will be a large change with other ramifications for us, so I believe it will probably translate to a new P4Runtime version in which we will deprecate that attribute (using the protobuf deprecation mechanism) and introduce a different one (different ones?) if needed.

jafingerhut: As long as there is something half-way plausible for a path to supporting that, _if_ it is ever needed, I think YAGNI applies here: https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it

mbudiu-vmw: I expect that there is a way which won't require deprecation. Const tables will always make sense. You will just have to add support for table initializers in some way: a table that starts with some entries populated by the P4 program - even though they can be modified later.

control-plane/p4RuntimeSerializer.cpp
```diff
@@ -1698,9 +1703,11 @@ static void analyzeTable(P4RuntimeAnalyzer& analyzer,
 
     bool supportsTimeout = getSupportsTimeout(table);
 
+    bool isConstTable = (table->getEntries() != nullptr);
```
hanw: Each table attribute is an `IR::Property` and the IR node has a member called `isConstant` member. You could return the value of `isConstant` here.

hanw:
```diff --git a/control-plane/p4RuntimeSerializer.cpp b/control-plane/p4RuntimeSerializer.cpp
index 713a7fa9..248da835 100644
--- a/control-plane/p4RuntimeSerializer.cpp
+++ b/control-plane/p4RuntimeSerializer.cpp
@@ -1126,7 +1126,8 @@ class P4RuntimeAnalyzer {
                   const boost::optional<DefaultAction>& defaultAction,
                   const std::vector<ActionRef>& actions,
                   const std::vector<MatchField>& matchFields,
-                  bool supportsTimeout) {
+                  bool supportsTimeout,
+                  bool isConstTable) {
         if (isHidden(tableDeclaration)) return;

         auto name = tableDeclaration->controlPlaneName();
@@ -1189,6 +1190,9 @@ class P4RuntimeAnalyzer {
         if (supportsTimeout) {
             table->set_with_entry_timeout(true);
         }
+
+        if (isConstTable) {
+        }
     }

     void addActionProfile(const ActionProfile& actionProfile,
@@ -1647,6 +1651,14 @@ static bool getSupportsTimeout(const IR::P4Table* table) {
     return expr->to<IR::BoolLiteral>()->value;
 }

+static bool getConstantTable(const IR::P4Table* table) {
+    auto const_table = table->properties->getProperty(P4V1::V1Model::instance
+                                                      .tableAttributes
+                                                      .entries.name);
+    if (const_table == nullptr) return false;
+    return const_table->isConstant;
+}
+
 static std::vector<ActionRef>
 getActionRefs(const IR::P4Table* table, ReferenceMap* refMap) {
     std::vector<ActionRef> actions;
@@ -1698,9 +1710,11 @@ static void analyzeTable(P4RuntimeAnalyzer& analyzer,

     bool supportsTimeout = getSupportsTimeout(table);

+    bool constantTable = getConstantTable(table);
+
     analyzer.addTable(table, tableSize, implementation, directCounter,
                       directMeter, defaultAction, actions, matchFields,
-                      supportsTimeout);
+                      supportsTimeout, constantTable);
 }

 /// Visit evaluated blocks under the provided top-level block. Guarantees that
diff --git a/frontends/p4/fromv1.0/v1model.h b/frontends/p4/fromv1.0/v1model.h
index ebbdb538..5150d11b 100644
--- a/frontends/p4/fromv1.0/v1model.h
+++ b/frontends/p4/fromv1.0/v1model.h
@@ -229,12 +229,14 @@ struct TableAttributes_Model {
     TableAttributes_Model() : tableImplementation("implementation"),
                               counters("counters"),
                               meters("meters"), size("size"),
-                              supportTimeout("support_timeout") {}
+                              supportTimeout("support_timeout"),
+                              entries("entries") {}
     ::Model::Elem       tableImplementation;
     ::Model::Elem       counters;
     ::Model::Elem       meters;
     ::Model::Elem       size;
     ::Model::Elem       supportTimeout;
+    ::Model::Elem       entries;
     const unsigned defaultTableSize = 1024;
 };```

antoninbas: I have added the `getConstantTable` method and I am checking that the property has a `const` modifier. However I am giving an error if there is no `const` modifier as this is disallowed by the current P4_16 spec (and enforced in the frontend) and is not supported by P4Runtime. It still makes sense to have the check in case the language restriction is relaxed and we are still not supporting it in P4Runtime.

I did not change v1model as in your diff. Unless I misunderstand what v1model.h is for, it doesn't make sense to change it as the notion of table initializer ("entries" property) is a P4_16 language built-in property. It has nothing to do with v1model and has no P4_14 equivalent.

PR #1163, opened 2018-02-16
Added Midend Pass to Optimize Complex Bitwise Operations

evanches:
In the p4-14 to p4-16 translation, the following instruction:
modify_field(field, param, mask)

becomes:
field = field & ~mask | param & mask

which is correct, but could be simplified for a backend as a vector of assignments of slices, as long as the masks don't have any collisions.

This also required a change to the equivalency check of Constants in local_copyprop.

### DIFF
--- a/midend/simplifyBitwise.cpp
+++ b/midend/simplifyBitwise.cpp
@@ -2,7 +2,8 @@
 
 namespace P4 {
 
-bool SimplifyBitwise::Scan::preorder(const IR::Operation *) {
+bool SimplifyBitwise::Scan::preorder(const IR::Operation *op) {
+    LOG1("Are we here? " << op);
     can_be_changed = false;
     return false;
 }
@@ -10,20 +11,26 @@ bool SimplifyBitwise::Scan::preorder(const IR::Operation *) {
 bool SimplifyBitwise::Scan::preorder(const IR::AssignmentStatement *as) {
     if (!as->right->is<IR::BOr>())
         return false;
+    auto action = findContext<IR::P4Action>();
+    if (action)
+        LOG1("Action name " << action->name.name);
     can_be_changed = true;
     total_mask = 0;
     return true;
 }
 
 bool SimplifyBitwise::Scan::preorder(const IR::BOr *bor) {
+    LOG1("Bor");
     if (!findContext<IR::AssignmentStatement>())
         can_be_changed = false;
     if (!(bor->left->is<IR::BAnd>() && bor->right->is<IR::BAnd>()))
         can_be_changed = false;
+    LOG1("Tests " << bor->left->is<IR::BAnd>() << " " << bor->right->is<IR::BAnd>());
     return can_be_changed;
 }
 
 bool SimplifyBitwise::Scan::preorder(const IR::BAnd *band) {
+    LOG1("Band");
     if (!findContext<IR::BOr>())
         can_be_changed = false;
     if (!band->left->is<IR::Constant>() && !band->right->is<IR::Constant>())
@@ -34,6 +41,7 @@ bool SimplifyBitwise::Scan::preorder(const IR::BAnd *band) {
 }
 
 bool SimplifyBitwise::Scan::preorder(const IR::Constant *constant) {
+    LOG1("Constant");
     if (!findContext<IR::BAnd>()) {
         can_be_changed = false;
         return false;
--- a/midend/simplifyBitwise.cpp
+++ b/midend/simplifyBitwise.cpp
@@ -2,11 +2,16 @@
 
 namespace P4 {
 
+bool SimplifyBitwise::Scan::preorder(const IR::Operation *) {
+    can_be_changed = false;
+    return false;
+}
+
 bool SimplifyBitwise::Scan::preorder(const IR::AssignmentStatement *as) {
     if (!as->right->is<IR::BOr>())
         return false;
     can_be_changed = true;
-    mpz_init(total_mask.get_mpz_t());
+    total_mask = 0;
     return true;
 }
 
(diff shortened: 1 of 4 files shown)


midend/simplifyBitwise.cpp
```diff
@@ -0,0 +1,98 @@
+#include "simplifyBitwise.h"
+
+namespace P4 {
+
+bool SimplifyBitwise::Scan::preorder(const IR::AssignmentStatement *as) {
+    if (!as->right->is<IR::BOr>())
+        return false;
+    can_be_changed = true;
+    mpz_init(total_mask.get_mpz_t());
```
ChrisDodd: I think this is not the correct way to initialize a `mpz_class` (it overwrites an already initialized mpz_t object, which would leak memory, but is innocuous due to garbage collection).  It should just be

    total_mask = 0;

midend/simplifyBitwise.cpp
```diff
@@ -0,0 +1,98 @@
...
+    if (!findContext<IR::BAnd>()) {
+        can_be_changed = false;
+        return false;
+    }
+    // Ensure that there are no collisions within the bitmask
+    mpz_class mask;
+    mpz_init(mask.get_mpz_t());
+    mpz_class zero;
+    mpz_init(zero.get_mpz_t());
+    mpz_and(mask.get_mpz_t(), total_mask.get_mpz_t(), constant->value.get_mpz_t());
```
ChrisDodd: The above should be

    mpz_class mask = total_mask & constant->value;

basically you should just be using `mpz_class` objects as if they were integers (and can intermix them with `int` and `long`) and they "just work"

midend/simplifyBitwise.cpp
```diff
@@ -0,0 +1,98 @@
...
+    if (mpz_cmp(mask.get_mpz_t(), zero.get_mpz_t()) == 0) {
+        mpz_ior(total_mask.get_mpz_t(), mask.get_mpz_t(), total_mask.get_mpz_t());
```
ChrisDodd: if (mask == 0) {
        total_mask = mask | total_mask;

...which is a noop?

midend/simplifyBitwise.cpp
```diff
@@ -0,0 +1,98 @@
...
+    } else {
+        can_be_changed = false;
+    }
+    return false;
+}
```
ChrisDodd: The above mask-testing code seems wrong to me -- it will always succeed, regardless of the masks?

You should be checking that the mask are disjoint AND cover all bits of the result.  If they don't cover all bits of the result, it could still work, but you would have to assign 0 to the slices that are not covered by either mask.

You could even make it work for masks that are not disjoint by assigning to the non-disjoint slices with a BAnd.

ChrisDodd:
This code looks like a golden opportunity for an expression tree pattern matcher.  Unfortunately, we don't have such a beast -- another idea for the "future cleanups"

midend/simplifyBitwise.cpp
```diff
@@ -0,0 +1,98 @@
...
+
+namespace P4 {
+
+bool SimplifyBitwise::Scan::preorder(const IR::AssignmentStatement *as) {
+    if (!as->right->is<IR::BOr>())
+        return false;
+    can_be_changed = true;
+    mpz_init(total_mask.get_mpz_t());
+    return true;
+}
```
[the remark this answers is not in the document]
evanches: I don't think that's the case, as the preorder in the AssignmentStatement will only proceed if the right is a BOr only.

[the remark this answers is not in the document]
evanches: But I can put it just incase.

midend/simplifyBitwise.cpp
evanches: You're right, mask should be replaced by constant->value.

midend/simplifyBitwise.cpp
evanches: You could extend this with a bunch of masks if you so chose to.  I was trying to handle this corner case that was arising.  You're right, let me rework the code to get this particular type of side case.

ChrisDodd: The important thing is to not accidentally catch other similar but not quite the same corner cases in a way that changes the semantics of the resulting code...

midend/simplifyBitwise.cpp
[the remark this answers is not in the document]
evanches: I agree with that.  Good catch.

midend/simplifyBitwise.cpp
```diff
@@ -0,0 +1,128 @@
+#include "simplifyBitwise.h"
+
+namespace P4 {
+
+bool SimplifyBitwise::Scan::preorder(const IR::Operation *op) {
+    LOG1("Are we here? " << op);
```
ChrisDodd: probably don't want LOG1 for such common messages -- LOG3 at least

midend/simplifyBitwise.cpp
```diff
@@ -0,0 +1,128 @@
...
+    return false;
+}
+
+void SimplifyBitwise::Scan::postorder(const IR::AssignmentStatement *as) {
+    if (can_be_changed)
+        self.changeable.insert(as);
+}
+
+const IR::Node *SimplifyBitwise::Update::preorder(IR::AssignmentStatement *as) {
+    auto orig_as = getOriginal()->to<IR::AssignmentStatement>();
```
ChrisDodd: As an aside, you can use `getOriginal<IR::AssignmentStatement>()` to do the getOriginal and dynamic cast and BUG_CHECK if its not of the expected type, all in one step (as this is such a common idiom).

PR #1165, opened 2018-02-16
Simplify SimplifyBitwise by using a general pattern matcher for expressions

ChrisDodd:
It seems so much simpler to do this with a basic pattern matcher that I couldn't resist refactoring it to use one...

### DIFF
--- a/ir/pattern.h
+++ b/ir/pattern.h
@@ -37,6 +37,16 @@ class Pattern {
         bool match(const IR::Node *n) override { return (m = n->to<T>()); }
         Match(const T *&m) : m(m) {}
     };
+    class Const : public Base {
+        mpz_class       value;
+     public:
+        bool match(const IR::Node *n) override {
+            if (auto k = n->to<IR::Constant>())
+                return k->value == value;
+            return false; }
+        Const(mpz_class v) : value(v) {}
+        Const(int v) : value(v) {}
+    };
     template<class T> class Unary : public Base {
         Base *expr;
      public:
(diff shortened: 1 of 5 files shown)


ir/pattern.h
```diff
@@ -0,0 +1,108 @@
...
+ * them with C++ operators to build larger patterns that can match tree fragments
+ */
+class Pattern {
+    class Base {
+     public:
+        virtual bool match(const IR::Node *) = 0;
+    } *pattern;
+    Pattern(Base *p) : pattern(p) {}
+    template<class T> class Match : public Base {
+        const T *&m;
```
mbudiu-vmw: This is very interesting; I was wondering how to do match + binding. This is a nice solution.
We talked about doing this for a long time, but we never got around to doing it...

ir/pattern.h
```diff
@@ -0,0 +1,108 @@
...
+     public:
+        bool match(const IR::Node *n) override {
+            if (auto b = n->to<T>()) {
+                if (left->match(b->left) && right->match(b->right))
+                    return true;
+                if (commutative && left->match(b->right) && right->match(b->left))
+                    return true; }
+            return false; }
+        Binary(Base *l, Base *r, bool commute = false) : left(l), right(r), commutative(commute) {}
+    };
```
mbudiu-vmw: How about the ability to match with a specific constant value?
This would enable much more compact strength-reduction code.

ChrisDodd: Should be easy enough to add, even though its not used in this example pass...

PR #1176, opened 2018-02-27
add sha to the p4c version

### LINKED ISSUE #1173: Compiler version number is not used properly
It would be very nice if we could arrange for the build process to automatically increment the minor version number of the compiler. Or maybe have the version print some commit id?

cc10512:
Updating the version should be done at "release" time, however, we
can automatically add the sha to the version number so people can
identify which release of the compiler are using.

Fixes #1173

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -38,9 +38,14 @@ OPTION (ENABLE_P4C_GRAPHS "Build the p4c-graphs backend" ON)
 OPTION (ENABLE_PROTOBUF_STATIC "Link against Protobuf statically" ON)
 
 if (NOT $ENV{P4C_VERSION} STREQUAL "")
+  # Allow the version to be set from outside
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -38,9 +38,14 @@ OPTION (ENABLE_P4C_GRAPHS "Build the p4c-graphs backend" ON)
 OPTION (ENABLE_PROTOBUF_STATIC "Link against Protobuf statically" ON)
 
 if (NOT $ENV{P4C_VERSION} STREQUAL "")
+  # Allow the version to be set from outside
   set (P4C_VERSION $ENV{P4C_VERSION})
 else()
-  set (P4C_VERSION "0.1")
+  execute_process (COMMAND git rev-parse --short HEAD
+    OUTPUT_VARIABLE P4C_GIT_SHA
```
mbudiu-vmw: This sure is a complicated language.

PR #378, opened 2017-03-22
p4c compiler driver

hanw:
Compiler driver for bmv2, ebpf and third-party backends.

Installing the driver:
```
cd p4c/tools/driver
python setup.py install
```

Usage
```
usage: p4c [-h] [-v] [-debug] [-###] [-Xpreprocessor <arg>] [-Xp4c <arg>]
           [-Xassembler <arg>] [-Xlinker <arg>] [-b BACKEND] [-E] [-e] [-S]
           [-c] [-x LANGUAGE] [-I SEARCH_PATH] [-o PATH]
           source_file

positional arguments:
  source_file           Files to compile

optional arguments:
  -h, --help            show this help message and exit
  -v, --version         show version and exit
  -debug                verbose
  -###                  print (but do not run) the commands
  -Xpreprocessor <arg>  Pass <arg> to the preprocessor
  -Xp4c <arg>           Pass <arg> to the compiler
  -Xassembler <arg>     Pass <arg> to the assembler
  -Xlinker <arg>        Pass <arg> to the linker
  -b BACKEND            specify target backend
  -E                    Only run the preprocessor
  -e                    Skip the preprocessor
  -S                    Only run the preprocess and compilation steps
  -c                    Only run preprocess, compile, and assemble steps
  -x LANGUAGE           Treat subsequent input files as having type language.
  -I SEARCH_PATH        Add directory to include search path
  -o PATH               Write output to the provided path
```

PR #1177, opened 2018-02-27
Support inferring which architecture file to include from --target command line option.

hanw:
Users don't have to include `core.p4` or `v1model.p4` in their program, as long as they specify which target the program should be compiled to, the compiler should be smart enough to include the correct file.

### DIFF
--- a/backends/ebpf/p4c-ebpf.cpp
+++ b/backends/ebpf/p4c-ebpf.cpp
@@ -39,14 +39,17 @@ void compile(EbpfOptions& options) {
         ::error("This compiler only handles P4-16");
         return;
     }
-    cstring device, arch, vendor;
-    std::tie(device, arch, vendor) = options.parseTarget();
 
-    options.preprocessor_options += " -D__TARGET_FRONTEND__";
-    if (arch == "ss") {
-        options.preprocessor_options += " -D__ARCH_V1MODEL__";
-    } else if (arch == "psa") {
-        options.preprocessor_options += " -D__ARCH_PSA__";
+    if (options.target != nullptr) {
+        cstring device, arch, vendor;
+        std::tie(device, arch, vendor) = options.parseTarget();
+
+        options.preprocessor_options += " -D__TARGET_FRONTEND__";
+        if (arch == "ss") {
+            options.preprocessor_options += " -D__ARCH_V1MODEL__";
+        } else if (arch == "psa") {
+            options.preprocessor_options += " -D__ARCH_PSA__";
+        }
     }
 
     auto program = P4::parseP4File(options);
--- a/backends/ebpf/p4c-ebpf.cpp
+++ b/backends/ebpf/p4c-ebpf.cpp
@@ -44,7 +44,7 @@ void compile(EbpfOptions& options) {
         cstring device, arch, vendor;
         std::tie(device, arch, vendor) = options.parseTarget();
 
-        options.preprocessor_options += " -D__TARGET_FRONTEND__";
+        options.preprocessor_options += " -D__TARGET_EBPF__";
         if (arch == "ss") {
             options.preprocessor_options += " -D__ARCH_V1MODEL__";
         } else if (arch == "psa") {
--- a/backends/ebpf/p4c-ebpf.cpp
+++ b/backends/ebpf/p4c-ebpf.cpp
@@ -39,6 +39,16 @@ void compile(EbpfOptions& options) {
         ::error("This compiler only handles P4-16");
         return;
     }
+    cstring device, arch, vendor;
+    std::tie(device, arch, vendor) = options.parseTarget();
+
+    options.preprocessor_options += " -D__TARGET_FRONTEND__";
+    if (arch == "ss") {
+        options.preprocessor_options += " -D__ARCH_V1MODEL__";
+    } else if (arch == "psa") {
+        options.preprocessor_options += " -D__ARCH_PSA__";
+    }
+
     auto program = P4::parseP4File(options);
     if (::errorCount() > 0)
         return;
--- a/frontends/common/options.cpp
+++ b/frontends/common/options.cpp
@@ -374,6 +374,20 @@ void CompilerOptions::dumpPass(const char* manager, unsigned seq, const char* pa
     }
 }
 
+std::tuple<cstring, cstring, cstring> CompilerOptions::parseTarget() {
+    std::vector<std::string> splits;
+    std::string target_str(target.c_str());
+    boost::split(splits, target_str, [](char c){return c == '-';});
+    if (splits.size() != 3)
+        BUG("Invalid target %s", target);
+
+    auto device = splits[0];
+    auto arch   = splits[1];
+    auto vendor = splits[2];
+
+    return {device, arch, vendor};
+}
+
 DebugHook CompilerOptions::getDebugHook() const {
     using namespace std::placeholders;
     auto dp = std::bind(&CompilerOptions::dumpPass, this, _1, _2, _3, _4);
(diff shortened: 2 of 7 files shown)


backends/ebpf/p4c-ebpf.cpp
```diff
@@ -39,6 +39,19 @@ void compile(EbpfOptions& options) {
         ::error("This compiler only handles P4-16");
         return;
     }
+
+    if (options.target != nullptr) {
+        cstring device, arch, vendor;
+        std::tie(device, arch, vendor) = options.parseTarget();
+
+        options.preprocessor_options += " -D__TARGET_EBPF__";
+        if (arch == "ss") {
```
mbudiu-vmw: This is wrong, you can never compile ebpf programs for v1model or PSA targets.

p4include/arch.p4
```diff
@@ -0,0 +1,40 @@
...
+#ifdef __TARGET_GRAPH__
+ #ifdef __ARCH_V1MODEL__
+  #include "v1model.p4"
+ #endif
+ #ifdef __ARCH_PSA__
+  #include "psa.p4"
+ #endif
+#endif
+
+#ifdef __TARGET_EBPF__
```
mbudiu-vmw: This does not work: you cannot list all possible back-end here, because you don't know what they are. Where is the Tofino back-end, or the p4c-xdp back-ends, which are loaded as extensions? This breaks the modularity of the compiler. So far there no one piece that has to know about all back-ends; when you add a back-end it just appears. 

If you really want to do this you should do it as part of the build process, and this file should be automatically generated, in the same way makefiles are.

p4include/arch.p4
```diff
@@ -0,0 +1,40 @@
+#ifndef __INCLUDE_ARCH_P4__
+#define __INCLUDE_ARCH_P4__
+
+#ifdef __TARGET_BMV2__
+ #ifdef __ARCH_V1MODEL__
+  #include "v1model.p4"
```
mbudiu-vmw: This should be `<v1model.p4>`. The angle brackets are important.

frontends/common/options.cpp
```diff
@@ -372,6 +374,20 @@ void CompilerOptions::dumpPass(const char* manager, unsigned seq, const char* pa
     }
 }
 
+std::tuple<cstring, cstring, cstring> CompilerOptions::parseTarget() {
+    std::vector<std::string> splits;
+    std::string target_str(target.c_str());
+    boost::split(splits, target_str, [](char c){return c == '-';});
+    if (splits.size() != 3)
+        BUG("Invalid target %s", target);
```
mbudiu-vmw: We have never specified as a requirement that target has to look this way. For example, the ebpf back-end has different requirements for the target string.

frontends/common/options.cpp
```diff
@@ -284,14 +284,16 @@ FILE* CompilerOptions::preprocess() {
...
         // the p4c driver sets environment variables for include
         // paths.  check the environment and add these to the command
         // line for the preprocessor
         char * driverP4IncludePath =
           isv1() ? getenv("P4C_14_INCLUDE_PATH") : getenv("P4C_16_INCLUDE_PATH");
         cmd += cstring(" -C -undef -nostdinc") + " " + preprocessor_options
             + (driverP4IncludePath ? " -I" + cstring(driverP4IncludePath) : "")
-            + " -I" + (isv1() ? p4_14includePath : p4includePath) + " " + file;
+            + " -I" + (isv1() ? p4_14includePath : p4includePath)
+            + " " + (isv1() ? "" : arch) + " " + file;
```
mbudiu-vmw: I think that this is not a good idea. A compiler should not magically include files based on esoteric and undocumented environment variables. It will be difficult for the user to understand what is going on. The compiler should include exactly the files that the user specifies in the program using `#include`. Then the compilation process is clear to everyone.

frontends/common/options.cpp
ChrisDodd: The 3-component target is a feature/requirement of the p4c driver program, so if you want a backend to be callable via the driver, it needs to use the same kind of target.  If that is not adequate for some target, we need to figure out what needs to be added to the p4c driver

backends/ebpf/p4c-ebpf.cpp
ChrisDodd: Is it impossible to support psa on the ebpf target/backend?  (psa and v1model are architectures, not targets -- ideally you could compile any architecture on any target, or at least get a clear message that that arhitecture/target combination is not supported).

p4include/arch.p4
ChrisDodd: Since the includes seem purely based on the `__ARCH_` symbols, is it even necessary to check `__TARGET_`anything?

frontends/common/options.cpp
ChrisDodd: If this is an issue about using the `p4c` driver program vs calling the backend executable directly, perhaps it should be the driver program adding the argument to its invocation of the preprocessor?  Actually, isn't it the driver that calls the preprocessor anyways and calls the backend executable with some flag to disable the preprocessor?  Or do we end up running the preprocessor twice when we use `p4c`?

frontends/common/options.cpp
mbudiu-vmw: I have never seen a specification of what the driver is supposed to do.
If the driver specification influences the design of the compiler we have to have it written down somewhere.

frontends/common/options.cpp
mbudiu-vmw: Yes, the driver would be the natural place for this.

backends/ebpf/p4c-ebpf.cpp
mbudiu-vmw: Take a look at p4include/ebpf_model.p4 and it will be obvious that it is very different from v1model.p4. For example, the ebpf back-end only supports packet filters. It does not have output ports or output packets, just a boolean. The set of externs supported is very different. Table properties are different, for example, there are no ternary key matches.

cc10512:
@mbudiu-vmw there's been an entire discussion on the driver design here: https://github.com/p4lang/p4c/pull/378. The triplets are documented in the front README file: https://github.com/p4lang/p4c/blob/1b5a81340eec04a7a2adf3a5d627fc65454cac5d/docs/README.md#compiler-driver

From a usability point of view, this is a move in the right direction. This is what we've been hearing from our customers. It also seems counterintuitive for users to invoke directly the backend for a particular target, especially when that backend is not the only executable that needs to be run. Look at gcc and clang, both are hiding backends, assembler, linker, etc. behind target options. So this is nothing out of ordinary, and it is totally necessary for usability on more complex targets (such as Barefoot's).

In the same vein, we should also have the compiler include standard includes based on the target. We have been equating the architecture with libc -- well, the compiler initializes the entire libc library (and all its standard objects) without any programmer intervention. This is why it seems quite unnecessary to ask the programmers to explicitly include core.p4 when there is nothing you can do in p4 without core.p4. Similarly, if you already specified a target-architecture triplet on the command line, you should not need to also specify the architecture file as an include. In most cases you will get a cryptic compiler error because some incompatibility between controls. It seems more natural to require specifying the target and the architecture in one place and drive the rest of the compiler to do the right thing.

That said, I think that the current PR is a temporary solution, but one that allows the insulation from user changes. We need a more involved solution in the compiler to handle everything properly, but that will take more time.

frontends/common/options.cpp
cc10512: The driver calls the preprocessor, but the preprocessor also is called from the compiler. This is why we need deeper changes in the way we handle these includes.

mbudiu-vmw:
Even if gcc handles libc, it does not insert `#include <stdlib.h>` for you in your source file.
I think it is very useful for the user to understand the target architecture because they include the file explicitly.
I don't have any problems with the driver doing this automatically, but I think that the compiler binary should not do that.

Issue #489, opened 2017-04-17
P4_14 parser exceptions are not recognized by bison parser


antoninbas:
parser exceptions have never been used much in P4_14 AFAIK, however people are complaining that p4c is rejecting valid (according to the current spec) P4_14 programs that use parser exceptions

mbudiu-vmw:
I have assigned @ChrisDodd; but he can probably only help with the grammar; we should discuss about how to represents exceptions in P4-16. It would be nice to have at least one test example with exceptions.
I never understood clearly their semantics personally.

ChrisDodd:
The P4_14 frontend has support for parser exceptions, so any legal program that uses them should be accepted.  They're just dropped before conversion to P4_16, however, so they won't be visible in the midend or backend, which might result in problems there.

mbudiu-vmw:
I have pushed a "fix" for this issue which only consists in emitting a warning when a parser exception is encountered. I suggest to close this issue - which seems to be about parsing programs with parser exceptions. If we want to implement properly a translation to P4-16 we should probably file a different issue. If someone can explain exactly how a parser exception is triggered then we can attempt to implement the translation.

PR #1178, opened 2018-02-28
Fix for issue #489

mbudiu-vmw:


frontends/p4/fromv1.0/converters.cpp
```diff
@@ -546,6 +546,8 @@ class DiscoverStructure : public Inspector {
     explicit DiscoverStructure(ProgramStructure* structure) : structure(structure)
     { CHECK_NULL(structure); setName("DiscoverStructure"); }
 
+    void postorder(const IR::ParserException* ex) override
```
mbudiu-vmw: Note: this line is currently never executed, since the bison parser never creates a ParserException object.

Issue #1093, opened 2017-12-12
Infinite loop


jnfoster:
The compiler notifies about a syntax error,
```
bug.p4(8):syntax error, unexpected END
  if(0 == 1) { }
                ^
```
and then appears to go into an infinite loop on this P4_14 input program.
```
parser start {
  return ingress;
}    
control ingress {
  d();
}
control d {
  if(0 == 1) { }
```

mbudiu-vmw:
@ChrisDodd @sethfowler  : I think this is a bison-related problem. The lexer reaches end of file but the parser gets into an infinite loop.

mbudiu-vmw:
I think this has something to do with some bison rules that are supposed to handle errors, which @ChrisDodd  wrote.

ChrisDodd:
The problem seems to be that the error recovery code tries to find the closing `}` to recover from the syntax error in the control block, but fails, getting stuck on the extra END symbol that got added at eof when the bison parser was switched to C++ mode.  I'll try to come up with a workaround.

PR #1181, opened 2018-03-01
Match explicit END tokens after error in bison parser

ChrisDodd:
bison C++ parser can't recover properly (automatically) from errors on eof, due to the explicit END token required.  Need to explicitly match END where it can occur in error recovery to avoid hang.

Fixes issue #1093


frontends/parsers/v1/v1parser.ypp
```diff
@@ -268,10 +268,12 @@ header_type_declaration: HEADER_TYPE name "{" header_dec_body "}"
                                 new IR::Annotations(*$4.annotations), *$4.fields),
             new IR::Type_Header(@1+@5, IR::ID(@2, $2),
                                 new IR::Annotations(*$4.annotations), *$4.fields))); }
+    | HEADER_TYPE name "{" header_dec_body error END
```
mbudiu-vmw: Sometimes you set `$$`, but sometimes you don't...

ChrisDodd: The non-terminals that have a type need to set `$$` to something for every action.  The non-terminals that don't have a type can't set `$$`...

PR #1182, opened 2018-03-02
Add parser support to pass saturating as a field annotation

amreshk:
This allows `isSaturating` to be passed as an `Annotation` on a field, which can be used in the backed to identify `saturating` fields.

### DIFF
--- a/frontends/parsers/v1/v1parser.ypp
+++ b/frontends/parsers/v1/v1parser.ypp
@@ -114,6 +114,9 @@ typedef const IR::Type ConstType;
 // types, and another with null checking for pointer types.
 %printer { yyoutput << $$; } <BBoxType> <CaseValue> <cstring> <HeaderType>
                              <int> <IR::Direction> <IR::ID> <UnparsedConstant>
+
+%printer { ; } <std::pair<ConstType*, int>>
+
 %printer {
     auto val = $$;
     if (val != nullptr) {
@@ -224,7 +227,8 @@ typedef const IR::Type ConstType;
 %type<IR::V1Parser*>                parser_statement_list
 %type<IR::Register*>                register_spec_list
 %type<IR::V1Table*>                 table_body
-%type<ConstType*>                   bit_width type
+%type<ConstType*>                   type
+%type<std::pair<ConstType*, int>>   bit_width
 
 %%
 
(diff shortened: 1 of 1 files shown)


cc10512:
Really? Wouldn't it be more natural to have a saturated data type and use the arithmetic operators with saturated semantics? Surely, externs are a bit too much of a hammer for this. And thus, an annotation seems a good way to prove that we can extend the data types we support in the language.

jafingerhut:
@cc10512 Do you mean a new type other than `bit<W>` and `int<W>` added to the language, with arithmetic operations defined on them with saturating behavior?  If so, I cannot think of any objections, other than the extra work of going through the language work group to add them to the language spec.

Extern method calls have the benefit of not needing to add new types to the language.

It isn't clear to me from the P4_16 language spec whether annotations on a variable/parameter are allowed to change the behavior of the built-in operators like plus, minus, etc. on them.

mbudiu-vmw:
Annotations changing the meaning of operations is a nightmare from the point of view of semantics. If fact, even the semantics of saturated arithmetic is completely unclear. I thought about this for a while and I am pretty sure that externs are the best route. The advantage is that you can define only the operations you need, and for these you can specify their semantics clearly in comments, and the compiler needs zero changes. We can talk about this at our next language meeting, or afterwards.

cc10512:
@jafingerhut, @mbudiu-vmw yes, I meant a new type such as saturated_bit<W> or saturated_int<W>. Indeed, it requires an addition to the language. I agree that annotations on operations themselves are cumbersome, and in fact, this PR attempts to annotate the fields, and thus the types, not the operations. We realize it is not sufficient, as it does not touch the type checker, so the compiler does not provide any safety for expressions that mix saturated and non-saturated terms.

I believe that it is much saner to define the semantics of the operations directly on a data type, and then define the proper conversion when used with other types. I shall try to get together a proposal for the Monday meeting -- @jnfoster eagerly added it to the agenda :)

I'm keen to hear Mihai's thoughts. In our view, using externs seems to gratuitously burden the programmer with the complexity for the advantage of saving work in the frontend of the compiler. One still needs to implement the extern in the backend, so there have to be changes to the compiler. Moreover, it is essentially a safe addition to the language, and one that can be safely ignored by backends that do not support saturating arithmetic. We already have multiple cases when we reject P4 programs because they can not be mapped to a target.

In the meantime, I would appreciate your help converging on this PR as an experimental compiler feature, especially as it is added only for P4_14. I will help us support legacy code in P4_14.

mbudiu-vmw:
I think we can close this PR, it is superseded by #1241

Issue #1043, opened 2017-11-17
How to use Resubmit in p4-16?


MaoJianwei:
P4 - 16

How to use Resubmit ?
Is there any example in the tutorial or somewhere else ?

Can I put user-defined metadata into the _resubmit()_?
How can I read the put data from the pipeline of resubmitted packet?

When I use like:

`resubmit<Mao_metadata>(meta);`
or:
`resubmit(meta);`

The p4c-bm-ss will reports errors:

![qq 20171117155330](https://user-images.githubusercontent.com/12800168/32936738-6baa12a4-cbb0-11e7-9ed3-8b29776f9b7d.jpg)

mbudiu-vmw:
This looks like a bug in the compiler, but this information is not sufficient to reproduce your error. Can you supply a small program that exhibits this symptom?

MaoJianwei:
Could you please write an available P4-16 program as a resubmit tutorial for us?

Corresponding with this:
https://github.com/p4lang/tutorials/blob/master/examples/resubmit/p4src/resubmit.p4

MaoJianwei:
eeee.....by the way,

This tutorial failed to resubmit too....

https://github.com/p4lang/tutorials/blob/master/examples/resubmit/p4src/resubmit.p4

mbudiu-vmw:
This is not the right forum for example programs, you should use one of the P4 mailing lists for this purpose. Did you use this compiler to compile that program?

It would be nice if you could help us locate the bug in the compiler which caused the error message you posted last week. Can you attach a P4 program that produces the compiler error that you described above?

MaoJianwei:
I aimed to the resubmit bug, but not the tutorial demo.

mbudiu-vmw:
I will tag this issue as "invalid" because we don't have enough information to reproduce it.
I will leave it open for a while in the hope that we will get more information.

JehandadKhan:
The tutorial mentioned above does not work, when debugged using the p4 debugging utility, the following is my diagnosis:

```
New event for packet 0.0
Entering action '_resubmit'
P4DBG:
New event for packet 0.0
Update for field mymeta.f1
New value is 01 (1)
P4DBG:
New event for packet 0.0
Update for field standard_metadata.resubmit_flag
New value is 00 (0)
P4DBG:
New event for packet 0.0
Exiting action '_resubmit'
P4DBG:
New event for packet 0.0
Exiting table 't_ingress_2'
P4DBG:
New event for packet 0.0
[...]

PR #1183, opened 2018-03-02
Fix resubmit in bmv2 backend

antoninbas:
Just like for learn lists, field list ids (which are used for resubmit)
need to be non-zero, since the id is actually used as a flag to indicate
whether or not resubmit was invoked.

Given that the "resubmit_flag" v1model metadata field is supposed to
hold a field list id, its bitwidth was increased from 1 to 32.


antoninbas:
Fixes https://github.com/p4lang/p4c/issues/1043

Issue #1184, opened 2018-03-06
Recirculate does not work


paulsubrata55:
I was trying out the recirculate mechanism and seems it does not work. Using simple switch grpc.

    control MyEgress(inout headers hdr,                                              
             inout metadata meta,                                            
             inout standard_metadata_t standard_metadata) {                                                                                  
    action ofpat_goto_table_(bit<7> table_id) {                                  
           meta.of_metadata.table_id = table_id;                                    
            recirculate(meta);                                                                                                              
     }                                                                            
    table ofpat_goto_table {                                                     
        key = {                                                                  
            meta.of_metadata.index : exact;                                      
        }                                                                        
                                                                             
        actions = {                                                              
            ofpat_goto_table_;                                                   
            NoAction;                                                            
        }                                                                        
        size = 1024;                                                             
        }                                                                            
        apply {                                                                      
            if ((meta.of_metadata.bmap & 4) == 4) {                                  
                ofpat_goto_table.apply();                                            
        }                                                                        
    }                                                                            
}

Seems like intrinsic_metadata.recirculate_flag metadata field in v1model.p4 is missing.

mbudiu-vmw:
We just merged a fix for this. Are you using a new build?

jafingerhut:
I see a recent commit with a fix for resubmit, but recirculate is controlled by a different metadata flag, according to this comment: https://github.com/p4lang/behavioral-model/issues/558#issuecomment-370564691

mbudiu-vmw:
@paulsubrata55: Maybe you can run a quick experiment and just add that declaration to v1model.p4.
It that's enough to make it work you can then just submit a PR...

paulsubrata55:
@mbudiu-vmw: I added recirculate flag in v1model.p4 and its started working as expected:
Changes done in v1model.p4
        @alias("intrinsic_metadata.recirculate_flag") bit<32> recirculate_flag; 

logs:
[05:11:09.711] [bmv2] [D] [thread 3940] [0.0] [cxt 0] Pipeline 'egress': end
[05:11:09.711] [bmv2] [D] [thread 3940] [0.0] [cxt 0] Deparser 'deparser': start
[05:11:09.711] [bmv2] [D] [thread 3940] [0.0] [cxt 0] Updating checksum 'cksum'
[05:11:09.711] [bmv2] [D] [thread 3940] [0.0] [cxt 0] Deparsing header 'ethernet'
[...]

PR #1186, opened 2018-03-07
Fix for issue #1184

mbudiu-vmw:


mbudiu-vmw:
Only if it is useful. I understand that you suggest that I should delete it?

antoninbas:
I know bmv2-simple_switch doesn't use it. I don't know if someone else uses it in their backend (@hanw, do you use `recirculate_port`?). Maybe in doubt we should keep it...

PR #1191, opened 2018-03-15
Support value_set in P4Info generation

antoninbas:
The P4 syntax may not be set in stone, but people are starting to use parser value sets and require support from the compiler to include the correct information in the P4Info message.

### DIFF
--- a/control-plane/p4RuntimeSerializer.cpp
+++ b/control-plane/p4RuntimeSerializer.cpp
@@ -536,7 +536,8 @@ enum class P4RuntimeSymbolType {
     EXTERN,
     EXTERN_INSTANCE,
     METER,
-    TABLE
+    TABLE,
+    VALUE_SET
 };
 
 /**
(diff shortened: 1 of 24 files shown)


control-plane/p4RuntimeSerializer.cpp
```diff
@@ -1219,11 +1223,53 @@ class P4RuntimeAnalyzer {
...
+            }
+            size = sizeConstant->value.get_ui();
+        }
+
+        auto vs = p4Info->add_value_sets();
+        auto id = symbols.getId(P4RuntimeSymbolType::VALUE_SET, name);
+        vs->mutable_preamble()->set_id(id);
+        vs->mutable_preamble()->set_name(name);
+        vs->mutable_preamble()->set_alias(symbols.getAlias(name));
+        vs->set_bitwidth(bitwidth);
```
mbudiu-vmw: Is this the only important thing, or do you care about the structure of the element type?

antoninbas: Thanks for the review. At the moment we only care about the total bitwidth but we are reviewing a proposal to include the structure information in P4Info. This is related to the type serialization discussion.

Issue #1179, opened 2018-02-28
core.p4 not found


mbudiu-vmw:
This commit: https://github.com/p4lang/p4c/commit/94af1b909d4af0caadee1948c406191a037b2c89
seems to cause problems for people who do not install the compiler.

Wasdns:
This problem occurs as I installing p4c today(180301). 

The command:
```
p4c-bm2-ss optimize.p4
```

The error log:
```
optimize.p4:2:19: error: no include path in which to search for core.p4
 #include <core.p4>
                   ^
optimize.p4:3:22: error: no include path in which to search for v1model.p4
 #include <v1model.p4>
                      ^
optimize.p4(50):syntax error, unexpected IDENTIFIER "packet_in"
parser MyParser(packet_in
                ^^^^^^^^^
error: Preprocessor returned exit code 256; aborting compilation
error: 2 errors encountered, aborting compilation
```

antoninbas:
I have a similar issue. I get errors when I install the compiler but not when I use the locally built version:
```
antonin@ubuntu:~/Documents/p4lang/p4c/build$ p4c-bm2-ss ../testdata/p4_16_samples/action_profile-bmv2.p4 --p4-16
../testdata/p4_16_samples/action_profile-bmv2.p4:17:19: error: no include path in which to search for core.p4
../testdata/p4_16_samples/action_profile-bmv2.p4:18:22: error: no include path in which to search for v1model.p4
../testdata/p4_16_samples/action_profile-bmv2.p4(25):syntax error, unexpected IDENTIFIER "packet_in"
parser ParserI(packet_in
               ^^^^^^^^^
error: Preprocessor returned exit code 256; aborting compilation
error: 2 errors encountered, aborting compilation
antonin@ubuntu:~/Documents/p4lang/p4c/build$ ./p4c-bm2-ss ../testdata/p4_16_samples/action_profile-bmv2.p4 --p4-16
../testdata/p4_16_samples/action_profile-bmv2.p4(35): warning: Target does not support default_action for IngressI.indirect (due to action profiles)
    table indirect {
          ^^^^^^^^
../testdata/p4_16_samples/action_profile-bmv2.p4(42): warning: Target does not support default_action for IngressI.indirect_ws (due to action profiles)
    table indirect_ws {
          ^^^^^^^^^^^
```

antoninbas:
Using the p4c driver instead of p4c-bm2-ss solves the issue.

[...]

PR #1198, opened 2018-03-28
use relative paths to package relocatable binaries

cc10512:
Take two on the right paths to find includes for both developer
and installed versions of the compiler.

We use relative paths for the "artifacts" while we use the full
path only for the command line define CONFIG_PKGDATADIR that is
used as a hardcoded path to find includes.


antoninbas:
Fixes #1179

PR #1199, opened 2018-03-29
separate --target flag to --target and --arch

hanw:
Separating the --target triplet to two flags --target and --arch in all backends and driver based on feedback from some users.

E.g.,
```
--target bmv2-ss-p4org
```
is replaced with
```
--target bmv2 --arch v1model
```


backends/bmv2/backend.cpp
```diff
@@ -181,14 +181,16 @@ class RenameUserMetadata : public Transform {
...
     auto evaluator = new P4::EvaluatorPass(refMap, typeMap);
     if (tlb->getMain() == nullptr)
         return;  // no main
 
-    if (options.arch != Target::UNKNOWN)
-        target = options.arch;
+    if (options.arch == "v1model")
+        target = Target::SIMPLE;
+    else if (options.arch == "psa")
+        target = Target::PORTABLE;
```
cc10512: Nit: I would use the full names: `Target::SIMPLE_SWITCH` and `Target::PSA`  (or `Target::PORTABLE_SWITCH_ARCHITECTURE`)

frontends/common/options.cpp
```diff
@@ -64,13 +64,13 @@ CompilerOptions::CompilerOptions() : Util::Options(defaultMessage) {
                    [this](const char*) {
                        langVersion = CompilerOptions::FrontendVersion::P4_14;
                        return true; },
-                    "[Deprecated] Specify language version to compile",
+                    "[Deprecated] Specify language version to compile.",
```
cc10512: Should these be removed all together? How long have they been "deprecated"?

hanw: the driver is using --std, can we use the same flag for all backends, instead of --p4v?

Issue #1202, opened 2018-03-29
Internal compiler error on function-style cast


ChrisDodd:
If you try to use a function-style cast (constructor call) for a builtin type, you get an internal compiler error (CompilerBug) with no line info about where the problem might be.  Eg,

    hdr.field1 = bit<32>(hdr.field2);

gives:

    terminate called after throwing an instance of 'Util::CompilerBug'
      what():  In file: /home/cdodd/bf-p4c/p4c/ir/expression.def:334
    Compiler Bug: bit<32>: unexpected type

This should be either treated as a cast, or at least give a clean error message.

mbudiu-vmw:
I guess that BUG_CHECK should be turned into an error.
Do you want to assign someone to this?

ChrisDodd:
The BUG_CHECK is the `verify` routine for IR::ConstructorCallExpression, so it should be checking and giving an error before trying to create the constructor call.

I don't have time to work on this just now, so I just dumped it here to avoid forgetting about it.  At some point I'll fix it, unless someone else wants to pick it up.

PR #1203, opened 2018-03-29
Fix for issue #1202; missing reference outputs

mbudiu-vmw:
Instead of checking the error message I have tightened the grammar.
We should also fix the spec if we approve this.


mbudiu-vmw:
@ChrisDodd: how about this simple fix for the issue you filed?

Issue #1208, opened 2018-04-03
Compiler bug: unexpected type


jnfoster:
I replaced the version of `psa.p4` in `p4c/p4include` with the official version, and then attempted to run the following program through `p4test`:
```p4
#include <core.p4>
#include <psa.p4>

struct EMPTY { };
    
parser MyIP(
    packet_in buffer,
    out EMPTY a,
    inout EMPTY b,
    in psa_ingress_parser_input_metadata_t c,
    in EMPTY d,
    in EMPTY e) {
    state start {
        transition accept;
    }
}

parser MyEP(
    packet_in buffer,
    out EMPTY a,
    inout EMPTY b,
    in psa_egress_parser_input_metadata_t c,
    in EMPTY d,
    in EMPTY e,
    in EMPTY f) {
    state start {
        transition accept;
    }
}

control MyIC(
    inout EMPTY a,
    inout EMPTY b,
    in psa_ingress_input_metadata_t c,
    inout psa_ingress_output_metadata_t d) {
    apply { }
}

control MyEC(
    inout EMPTY a,
    inout EMPTY b,
    in psa_egress_input_metadata_t c,
    inout psa_egress_output_metadata_t d) {
    apply { }
}

control MyID(
    packet_out buffer,
    out EMPTY a,
    out EMPTY b,
    out EMPTY c,
    inout EMPTY d,
    in EMPTY e,
    in psa_ingress_output_metadata_t f) {
    apply { }
}    

control MyED(
    packet_out buffer,
    out EMPTY a,
    out EMPTY b,
    inout EMPTY c,
    in EMPTY d,
    in psa_egress_output_metadata_t e,
    in psa_egress_deparser_input_metadata_t f) {    
    apply { }
}    
            
PSA_SWITCH(
    IngressPipeline(MyIP(), MyIC(), MyID()),
    EgressPipeline(MyEP(), MyEC(), MyED())) main;
```

I got a compiler bug:
```
In file: /Users/nate/shared/p4c/frontends/p4/specialize.cpp:57
Compiler Bug: /usr/local/share/p4c/p4include/psa.p4(666): package IngressPipeline<IH/94, IM/95, NM/96, CI2EM/97, RESUBM/98, RECIRCM/99>( IngressParser<IH, IM, RESUBM, RECIRCM> ip,  Ingress<IH, IM> ig,  IngressDeparser<IH, IM, CI2EM, RESUBM, NM> id);: unexpected type
[...]

PR #1211, opened 2018-04-04
Fix for issue #1208

mbudiu-vmw:


mbudiu-vmw:
@cc10512 : I am not sure I understand your suggestion. Do you want a different enum for each parameter?

cc10512:
@mbudiu-vmw my suggestion is based on the fact that all but one of the invocations of `checkParameters` are in the `canonicalize` method, you could declare in that method:
```
constexpr bool forbidModule = true;
constexpr bool forbidPackage = true;
```
and then use these as the arguments to checkParameters.

Issue #1210, opened 2018-04-04
Error compiling program that compares struct variable to struct constant


jafingerhut:
I tried compiling the attached program using the latest p4c-bm2-ss and p4test as of Apr 4, 2018, which is after this bug fix has been committed: https://github.com/p4lang/p4c/pull/1145

It appears there is no problem comparing two variables with the same struct type to each other, but comparing a struct variable to a constant of the same type gives an error.  Here is the error I see attempting to compile the attached program:
```
% p4test struct-variable-to-constant-compare-error.p4
struct-variable-to-constant-compare-error.p4(58): error: ==: not defined on struct PortId_t and Tuple(1)
        if (meta.foo == PSA_CPU_PORT) {
            ^^^^^^^^^^^^^^^^^^^^^^^^
```
[struct-variable-to-constant-compare-error.p4.txt](https://github.com/p4lang/p4c/files/1874794/struct-variable-to-constant-compare-error.p4.txt)

mbudiu-vmw:
The typechecker infers a tuple type for a list expression, and does not implicitly cast it to a struct.
Until we fix this bug a workaround is to define a constant with a struct type that you initialize with the list expression.

jafingerhut:
@mbudiu-vmw Sorry, I am not understanding your suggested workaround.

I thought that the attached program _is_ defining a constant with a struct type (PSA_PORT_CPU), and that it is initialized with a list expression, in this case `{9w192}`.

Do you have a code snippet example that uses your suggested workaround?

mbudiu-vmw:
I see, the problem is in the constant folding pass, which replaces the const object with its value. 
So you get something like this in your program:

```
if (meta.foo == { 9w192 }) 
```

And the type-checker is confused, since there is no implicit cast for equality, just for assignment. So there is no workaround until this is fixed.

mbudiu-vmw:
This was a bit tricky to fix, I had to write a fair amount of code.
The reason is that once we added support for structure equality this required us to support constant folding for struct equality, type-checking that compares structs to constant list expressions, and also comparisons between tuples. These were missing from the initial commit, but are there now. We could use a few more tests, because there are lots of possible combinations, including comparing headers with tuples.

jafingerhut:
With the changes in your fix: https://github.com/p4lang/p4c/commit/48a57a6ae4f96961b74bd13f6bdeac5add7bb815

[...]

PR #1212, opened 2018-04-05
Fix for issue #1210

mbudiu-vmw:


frontends/common/constantFolding.cpp
```diff
@@ -275,6 +279,32 @@ DoConstantFolding::compare(const IR::Operation_Binary* e) {
...
+                if (eqTest)
+                    tmp = new IR::Equ(li, ri);
+                else
+                    tmp = new IR::Neq(li, ri);
+                auto cmp = compare(tmp);
+                auto boolLit = cmp->to<IR::BoolLiteral>();
+                if (boolLit == nullptr)
+                    return e;
+                if (!boolLit->value)
+                    return boolLit;
```
ChrisDodd: This seems correct for `==` (false if any component differs) but seems wrong for `!=` (should be false only if all components are the same).  Should be `if (boolLit->value != eqTest)`?

testdata/p4_16_samples_outputs/list-compare-first.p4
```diff
@@ -11,6 +11,8 @@ const bool w = true;
 const S s = { 32w10, 32w20 };
 const bool v = true;
 const bool o = false;
+const bool vnot = true;
```
ChrisDodd: Something is definitely awry here, as both `s == x` and `s != x` got constfolded to true...

mbudiu-vmw: That is weird, because the first time I ran the test it actually was correct.
I will have to look into this.

PR #1213, opened 2018-04-06
change parser value_set to accept a size argument

hanw:
Follow up on issue https://github.com/p4lang/p4-spec/issues/565.

This PR implements the proposed syntax for value_set to take a size argument, instead of using `@size` annotation. 

```
value_set<tuple<bit<32>, bit<32>>> (4) pvs;
value_set<bit<32>> (8) pvs;
```

### DIFF
--- a/frontends/p4/typeChecking/typeChecker.cpp
+++ b/frontends/p4/typeChecking/typeChecker.cpp
@@ -1166,6 +1166,18 @@ const IR::Node* TypeInference::postorder(IR::Type_ValueSet* type) {
     return type;
 }
 
+const IR::Node* TypeInference::postorder(IR::P4ValueSet* decl) {
+    if (done())
+        return decl;
+    auto type = getTypeType(decl->type);
+    if (type == nullptr)
+        return decl;
+    auto orig = getOriginal<IR::P4ValueSet>();
+    setType(decl, type);
+    setType(orig, type);
+    return decl;
+}
+
 const IR::Node* TypeInference::postorder(IR::Type_Extern* type) {
     if (done()) return type;
     auto canon = setTypeType(type);
--- a/frontends/p4/typeChecking/typeChecker.cpp
+++ b/frontends/p4/typeChecking/typeChecker.cpp
@@ -1147,36 +1147,24 @@ const IR::Node* TypeInference::postorder(IR::Type_Set* type) {
     return type;
 }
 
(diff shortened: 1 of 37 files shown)


ir/ir.def
```diff
@@ -335,6 +335,13 @@ class P4Table : Declaration, IAnnotated, IApply {
     }
 }
 
+class P4ValueSet : Declaration, IAnnotated {
+    optional Annotations        annotations = Annotations::empty;
+    Type_ValueSet               type;
+    Expression                  expression;
```
mbudiu-vmw: Why don't you call this "size"?

ir/ir.def
```diff
@@ -335,6 +335,13 @@ class P4Table : Declaration, IAnnotated, IApply {
     }
 }
 
+class P4ValueSet : Declaration, IAnnotated {
+    optional Annotations        annotations = Annotations::empty;
+    Type_ValueSet               type;
```
mbudiu-vmw: Do we still need Type_ValueSet?
It looks to me like it can be deleted. 
Here you should just keep the element type.

frontends/p4/typeChecking/typeChecker.cpp
```diff
@@ -1147,6 +1147,7 @@ const IR::Node* TypeInference::postorder(IR::Type_Set* type) {
     return type;
 }
 
+#if 0
```
mbudiu-vmw: Do you want to also delete this?

hanw:
@antoninbas I adjusted the value_set gtest to have an `name` annotation on the value_set instance. I wasn't sure which pass in the frontend automatically adds the `name` annotation to the value_set instance for a P4-16 program. @mbudiu-vmw , do you remember?

antoninbas:
@hanw not sure, but don't you need to handle the new type in the `FindSymbols` pass (https://github.com/p4lang/p4c/blob/81dda93f64b4fc0fae47ff50b349ca14c79548a2/frontends/p4/uniqueNames.h#L68), otherwise the `RenameSymbols` pass will not give the correct names?
Can you figure it out before merging this, so that we don't have to explicitly add the annotations to the P4 source?

hanw:
Fixed the issue, it is the FindSymbols pass. @antoninbas can you check the gtest again? give me a thumb up to merge if it seems okay.

Issue #1216, opened 2018-04-09
Add compiler --emit-externs option


engjefersonsantiago:
Hello,

Following @mbudiu-vmw suggestion ([PR #1194](https://github.com/p4lang/p4c/pull/1194)), I am about to submit a PR to allow the json generation in the BMv2 backend.
This issue is to track the PR.

Jeferson

engjefersonsantiago:
Merged. Closing.

PR #1217, opened 2018-04-09
Support to --emit-externs

engjefersonsantiago:
Issue #1216


backends/bmv2/extern.cpp
```diff
@@ -75,9 +75,9 @@ bool Extern::preorder(const IR::Declaration_Instance* decl) {
 
 /// Custom visitor to enable traversal on other blocks
 bool Extern::preorder(const IR::PackageBlock *block) {
-    if (backend->target != Target::PORTABLE_SWITCH)
+    if (backend->target != Target::PORTABLE_SWITCH && !emitExterns)
         return false;
-
+    std::cout << "je suis ici\n";
```
mbudiu-vmw: This should be removed.
You should use LOG for log messages.

backends/bmv2/simpleSwitch.cpp
```diff
@@ -206,7 +207,25 @@ SimpleSwitch::convertExternObjects(Util::JsonArray *result,
             // Do not generate any code for this operation
         }
     } else {
-        error("Unknown extern type %1%", em->originalExternType->name);
+        if (emitExterns) {
+            ::warning("Found extern method %1%. Make sure that the BMv2 supports it",
```
mbudiu-vmw: I would think that this warning is unnecessary - if the user has passed the suitable flag to the compiler they already should know this.

frontends/common/options.cpp
```diff
@@ -121,6 +121,9 @@ CompilerOptions::CompilerOptions() : Util::Options(defaultMessage) {
                        }
                        return true; },
                    "Choose output format for the P4Runtime API description (default is binary).");
+    registerOption("--emit-externs", nullptr,
```
mbudiu-vmw: This option does not belong to common - it should be specific to the back-end.
Also, the documentation has to be more expansive, describing what will actually be generated in the output code.

backends/bmv2/simpleSwitch.h
```diff
@@ -46,12 +46,12 @@ class SimpleSwitch {
  public:
     void modelError(const char* format, const IR::Node* place) const;
     void convertExternObjects(Util::JsonArray *result, const P4::ExternMethod *em,
-                              const IR::MethodCallExpression *mc, const IR::StatOrDecl *s);
+                              const IR::MethodCallExpression *mc, const IR::StatOrDecl *s, const bool& emitExterns);
```
mbudiu-vmw: Did you run cpplint? These lines are too long.

backends/bmv2/simpleSwitch.cpp
```diff
@@ -586,6 +606,13 @@ SimpleSwitch::convertExternInstances(const IR::Declaration *c,
         }
 
         action_profiles->append(action_profile);
+    } else {
+        if (emitExterns) {
```
mbudiu-vmw: Shouldn't something happen for this case too?

backends/bmv2/simpleSwitch.h
engjefersonsantiago: I ran. I realized that the headers files were not include in the rule for the cpplint. My next commit will include it.

engjefersonsantiago:
Hi @mbudiu-vmw , I think I've addressed your comments.

backends/bmv2/extern.cpp
```diff
@@ -19,29 +19,28 @@ limitations under the License.
...
                             const IR::ExternBlock* block) {
     auto attributes = new Util::JsonArray();
-    auto paramIt = block->getConstructorParameters()->parameters.begin();
-    for (auto arg : *di->arguments) {
-        auto name = arg->toString();
-        if (arg->is<IR::Constant>()) {
-            auto cVal = arg->to<IR::Constant>();
-            if (arg->type->is<IR::Type_Bits>()) {
+    for (auto p : *block->getConstructorParameters()) {
+        auto name = p->toString();
```
mbudiu-vmw: You should use `p->name` and not `p->toString` here and below.

PR #1220, opened 2018-04-11
Handle "long long" primitive types in lib/json

antoninbas:
On some 64-bit sytems, int64 is a typedef to "long long" and not
"long". On those systems, trying to construct a JsonValue object from a
int64 or uint64 triggers a compiler error. In this commit, we add
support for constructing JsonValue objects from "long long" and
"unsigned long long" values. Note that the constructor for GMP's
mpz_class does not support "long long" values so we have to use a
workaround. This commit also makes handling of integral types in
lib/json more generic.

Note that this commit should also help for compiling on 32-bit systems,
where int64 is probably always a typedef to "long long".

### DIFF
--- a/lib/json.cpp
+++ b/lib/json.cpp
@@ -18,6 +18,7 @@ limitations under the License.
 #include <sstream>
 #include "json.h"
 #include "indent.h"
+#include "lib/gmputil.h"
 
 namespace Util {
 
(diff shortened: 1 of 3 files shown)


hanw:
Thanks for the quick fix.

PR #1222, opened 2018-04-12
Allow named arguments in calls; first step: types and grammar

mbudiu-vmw:
This is the first step in implementing support for named arguments in function calls.
We introduce a new IR class called Argument, which packages the argument information. The argument has an optional name, which is an ID. If the name is missing we use in fact an id whose `name` is a cstring which is nullptr - which is unusual, nowhere else in the compiler we allow null names.

This PR just converts the code to use Argument everywhere instead of expecting an Expression for the arguments. 

The code still matches arguments in order to parameters, ignoring the argument names supplied. A second PR will augment this one to make that change.  

I have also updated the bison grammar to allow names for arguments, to make sure that there are no ambiguities. But if you use them the result will be incorrect.

### DIFF
--- a/ir/base.def
+++ b/ir/base.def
@@ -260,6 +260,7 @@ class Argument {
     /// If an argument has no name the name.name is nullptr.
     optional ID name;
     Expression expression;
+    Argument { if (!srcInfo && expression) srcInfo = expression->srcInfo; }
     dbprint { out << (name.name.isNullOrEmpty() ? "" : name.name + " = ") << expression; }
     validate { CHECK_NULL(expression); }
     toString{
--- a/ir/base.def
+++ b/ir/base.def
@@ -147,8 +147,9 @@ abstract Operation : Expression {
     toString{ return getStringOp(); }
 }
 
-/// A Path is a vestige of a previous design.
-/// Today paths can only consist of a '.'
+/// Currently paths can be absolute (starting with a dot) or relative
+/// (just an identifier).  In a previous design paths could have
+/// multiple components.
 class Path {
     ID   name;
     optional bool absolute = false;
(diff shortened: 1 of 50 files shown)


ir/base.def
```diff
@@ -253,4 +254,20 @@ interface IInstance {
     virtual Type getType() const = 0;
 }
 
+/// An argument to a function call (or constructor call)
+/// Arguments may have optional names
+class Argument {
+    /// If an argument has no name the name.name is nullptr.
+    optional ID name;
+    Expression expression;
```
ChrisDodd: Add

    Argument { if (!srcInfo && expression) srcInfo = expression->srcInfo; }

to always initialize the srcInfo of the argument from the expression if its not otherwise overridden, and all the places you explicitly copy the srcInfo from the expression will no longer need to do that.  It will also deal with any spot you missed -- much less error prone.

mbudiu-vmw: I actually wanted to do this, but I wasn't sure about the syntax for constructors in the ir-generator...

ChrisDodd: The syntax above (class name with no argument list) is an ir-generator special for a chunk of code to be added to all to auto-generated constructors it generates.  You can also declare a constructor normally (which will declare that constructor, overriding any auto-generated ctor that would have had the same signature).

ir/base.def
```diff
@@ -253,4 +254,21 @@ interface IInstance {
...
```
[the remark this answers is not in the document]
mbudiu-vmw: Arguably the wrong decision was made in the original design, when we have not used an Argument class (which would have contained just one expression). We are only fixing this now.

However, if you look at the code everywhere, accessing IR fields directly by name is the standard procedure in this codebase. All IR fields are public, and are used as public. There are no accessor methods whatsoever.

Adding `operator()` would not reduce the number of changes.

Since I am not using any fancy IDE, refactoring expression to arg_value would have to be done again manually - probably a few hours of work. (Using Java that could have been done in 2 seconds...).

Moreover, the second part of this change that is upcoming will probably be even harder to make - match arguments to parameters; for the current change the type system pointed out most places that needed to be changed, but for the next one there will be no help. I expect that change to take at least a couple of days of work.

mbudiu-vmw:
@cc10512 : I am planning to merge this as it is, so I can get started on the second part.
If you really would like the changes you have suggested maybe it's best to file a new issue.

PR #1231, opened 2018-04-17
remove macro for PSA_SWITCH in psa.p4, update test case

hanw:
We already have a proposal with named argument to make the PacketReplicationEngine and BufferingQueueingEngine arguments optional. 

I would like to remove the PSA_SWITCH macro in psa.p4 to avoid creating more test cases that use a macro which is going to be retired later. 

Another reason to remove the macro is it does not play well with the gtest infrastructure, because gtest includes a raw string as test program which assumes the preprocessing is done, but there is no preprocessing support in the frontend, so all gtest cases that include psa.p4 is going to give a syntax error for the macro.

### DIFF
--- a/p4include/psa.p4
+++ b/p4include/psa.p4
@@ -639,15 +639,4 @@ package PSA_Switch<IH, IM, EH, EM, NM, CI2EM, CE2EM, RESUBM, RECIRCM> (
 
 // END:Programmable_blocks
 
-// Macro enabling the PSA program author to more conveniently create a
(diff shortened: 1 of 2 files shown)


jafingerhut:
Makes sense.  Making the corresponding changes in the p4-spec repo for psa.p4 and example programs would also resolve this issue: https://github.com/p4lang/p4-spec/issues/495

hanw:
@jafingerhut can you approve this PR? thanks.

Issue #1237, opened 2018-04-19
error: 4294967295: Value too large


jnfoster:
Compiling this program
```p4
header_type affairs {
  fields {
    flaccidly : 48;
    sleeting : 16;
  }
}

header_type backs {
  fields {
    sharps : 16;
    peptides : 16;
    whirs : 16;
    breasted : 16;
    browning : 16;
    jews : 32;
  }
}

header affairs kilometer;

header backs expressivenesss;

parser start {
  extract(kilometer);
  extract(expressivenesss);
  return ingress;
}

action mooneys() {
  subtract_from_field(expressivenesss.breasted, expressivenesss.peptides);
  subtract_from_field(kilometer.sleeting, 7);
}

table conceptualization {
  reads {
    kilometer : valid;
    kilometer.flaccidly mask 0 : lpm;
    expressivenesss : valid;
  }
  actions {
    mooneys;
  }
}

control ingress {
  apply(conceptualization);
}
```
using the p4c front-end produces the following mysterious error
```
p4test --std p4-14 allergically.p4 
error: 4294967295: Value too large
error: 4294967295: Value too large
error: 2 errors encountered, aborting compilation
```
If I delete "`mask 0`" from the `reads`, then there is no error.

jafingerhut:
Either someone has a program to pseudo-randomly generate P4 programs with an English dictionary for identifier names, or you are stretching P4 into domains it was never intended to go :-)

cc10512:
Are you saying that the value too large is not a bug but an allergic reaction? :)

mbudiu-vmw:
I think we can close this issue as well.

PR #1240, opened 2018-04-19
Fix for issue #1237

mbudiu-vmw:


hanw:
@mbudiu-vmw Do you want to rebase this one?

testdata/p4_14_samples_outputs/issue1237-first.p4
```diff
@@ -0,0 +1,81 @@
...
+        hdr.kilometer.sleeting = hdr.kilometer.sleeting + 16w65529;
+    }
+    @name(".conceptualization") table conceptualization {
+        actions = {
+            mooneys();
+            @defaultonly NoAction();
+        }
+        key = {
+            hdr.kilometer.isValid()      : exact @name("kilometer.$valid$") ;
+            48w0                         : lpm @name("0") ;
```
hanw: It's unclear to me if the control plane API for this argument would be generated correctly. Would the argument be named '0'?

mbudiu-vmw: That's a good question. Today this is a warning, maybe it should be an error?
@antoninbas : what should happen when a key field is a constant expression?

jafingerhut: I have a hard time imagining a situation where someone intentionally wants a compile-time constant value as a table key expression.  It seems that a correct implementation could leave that key in, and always search with that field equal to the constant value (0 in this example), but if the control plane ever installed a table entry that did not match 0 for that field, such a table entry would be impossible to match.

It seems like it justifies at least a warning from the compiler, if not an error.  If it is a warning, it isn't clear that much effort should be made in later tools to support this case, unless someone can think of a valid use case for it, so perhaps an error is preferable?

antoninbas: I think one acceptable solution that is probably also the easiest to implement is to print a warning in the p4RuntimeSerializer and omit the field altogether from the P4Info message. This way we do not put any burden on the P4Runtime implementers, for something that doesn't seem to have any actual use-cases.

mbudiu-vmw: Since we already have a warning I declare this PR ready for review.

PR #1241, opened 2018-04-20
adds support for saturating arithmetic in P4_16

### LINKED ISSUE #1229: implement saturating arithmetic operators
Implement the saturating arithmetic specification in http://github.com/p4lang/p4-spec/pull/609

cc10512:
We add support for |+| and |-| in the frontend of the compiler, as
well as bmv2 backend support. This includes defining two binary
operation IR nodes for saturating addition and subtraction, and
generating the json for bmv2 as defined in
https://github.com/p4lang/behavioral-model/pull/590. For that, we
update the json version to 2.18.

We also added a test that ensures that saturating arithmetic properly
bounds the min and max values for signed and unsigned integer
operations.

Still to do:
- translate P4_14 saturating operations to the new IR expression nodes.

Fixes #1229

### DIFF
--- a/backends/bmv2/expression.cpp
+++ b/backends/bmv2/expression.cpp
@@ -450,7 +450,11 @@ void ExpressionConverter::postorder(const IR::IntMod* expression)  {
 }
 
 void ExpressionConverter::postorder(const IR::Operation_Binary* expression)  {
-    binary(expression);
+    cstring op = expression->getStringOp();
+    if (op == "|+|" || op == "|-|")
+        saturated_binary(expression);
+    else
+        binary(expression);
 }
 
 void ExpressionConverter::binary(const IR::Operation_Binary* expression) {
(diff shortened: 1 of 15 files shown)


backends/bmv2/expression.cpp
```diff
@@ -450,7 +450,11 @@ void ExpressionConverter::postorder(const IR::IntMod* expression)  {
 }
 
 void ExpressionConverter::postorder(const IR::Operation_Binary* expression)  {
-    binary(expression);
+    cstring op = expression->getStringOp();
+    if (op == "|+|" || op == "|-|")
+        saturated_binary(expression);
```
ChrisDodd: Rather than doing string comparisons, you should just overload the `postorder` for AddSat and SubSat:

    void ExpressionConverter::postorder(const IR::AddSat *exp) { saturated_binary(exp); }
    void ExpressionConverter::postorder(const IR::SubSat *exp) { saturated_binary(exp); }

Issue #1157, opened 2018-02-07
Question: Is preserving typedef type aliases on table key fields, action parameters, etc. reasonable to implement?


jafingerhut:
In a few contexts when developing PSA, and now also in the P4 Runtime API working group, the observation has been made that it would be nice if an annotation could be placed on a typedef, e.g.:

```
typedef bit<13> PortId_t;
```
(Please mentally add one or more P4_16 annotations to the code above wherever they are syntactically legal -- I am being lazy and not looking that up right now.)

Then, wherever that type is used, e.g. for a field in a table's search key, or an action parameter, etc. the typedef name could be associated with it, and preserved in places like the P4Runtime Info file the compiler produces, and the P4 Runtime consumes.

My understanding is that right now, p4c treats typedefs as aliases for the underlying type, `bit<13>` in the example above, and it "forgets" the PortId_t type.

Is this "merely a convenience" for implementing the p4c compiler?  Something that in order to change, would require a change to the P4_16 language spec (and then the p4c implementation)?  Something else?

mbudiu-vmw:
I have a vague idea about what you mean, but to solve this issue we need a precise specification of what is desired. In general, if you have a declaration in your program, it is not guaranteed to map 1-1 to an object in the back-end. For example, the variable value may be constant-folded, or may be register allocated in a register which stores other values. So then there is no single object in the back-end that corresponds to the source declaration which could maintain source type information.

I think it is better to start from the problem that we need to solve by carrying type information around. Maybe we can state that more crisply?

jafingerhut:
Suppose I have a field in a table's key that is defined with type PortId_t (defined as `typedef bit<13> PortId_t`).  Right now P4Runtime Info will record that this is type `bit<13>` with no indication of anything about the named PortId_t.  It would be nice if it did, predictably, repeatably.

Same thing if such a value were used as an action parameter.

I can't think of any other desirable outcomes that aren't happening right now, but others might.

jafingerhut:
There may be other similar cases where such a value is an input value to some P4_16 extern, too, but the table keys and action parameters might be simplest to focus on first.

mbudiu-vmw:
Parameters are simpler, since they do not participate in optimizations.
Table keys are more problematic since they are expressions and the type of the expression is derived in general by the type-checker.
Is there a mechanism in P4Runtime to use symbolic type names?

Maybe what we need is a new kind of typedef which introduces a different type, which is not equivalent to the original one. E.g.:

```
[...]

PR #1244, opened 2018-04-20
Add newtype primitive

### LINKED ISSUE #1243: Add a newtype operation to create new types
Unlike typedef, the newtype is not equivalent to the original type.

mbudiu-vmw:
Fixes #1243 
Currently newtype only works with base types. A bit more work will be needed to make it work with tuples.
I don't think we should allow it to work with structs or other named types.

I also included a couple of test cases for older issues that are now fixed.


mbudiu-vmw:
There is some work needed to make this work with the control-plane API code generation. That is going to be the main consumer. @jafingerhut @antoninbas @vgurevich @ChrisDodd @chrispsommers

antoninbas:
I'm copying here a comment from the source code, which I think would be useful to have in the PR description:

> A newtype is like a typedef, but the type does not inherit any of the operations of the original type.  The new type has assignment, equality, and casts to/from the original type.

mbudiu-vmw:
These should be relatively easy to fix.
But do we want these values in headers? One way to get them there is to cast them...

jafingerhut:
This is not a "drop everything, implement this idea" question, but just a "what do you think of the difficulty of this idea?" question.

It seems like it should be a design choice in the language and compiler whether we support not only assignment, `==`, and `!=`, but _all_ of the unary and binary operations that we support on values of type `bit<7>`, also on values of type X where X is declared as `newtype bit<7> X;`

Is that true?  If so, would you guesstimate that is a lot more work in the compiler?  Moderate?  A little bit?

The reason I ask is that if this idea comes to be useful not only for type `PortId_t`, but also for other types like `newtype bit<32> IPv4Address_t;`, where:

+ the size in bits would be identical in the data plane and control plane, and
+ there would be _no_ numerical translation in values between control plane and data plane, and
+ the control plane has annotations on values of that type for purposes of debug, and conversion to/from strings for purposes of a nicer user interface

then it might be nice if P4 programmers could then do all of the usual `bit<32>` binary and unary operations on values of those types, where the binary operators would both have to be of the same newtype, and the resulting value would be of the same newtype.

I know that if that is not done, one could still do extra casting between `bit<32>` and `IPv4Address_t` values like this -- it just starts getting a bit verbose:

```
// Some arithmetic code on `newtype bit<32> IPv4Address_t` if we must cast to `bit<32>` first
newtype bit<32> IPv4Address_t;

IPv4Address_t addr1;
IPv4Address_t mask;

addr1 = (IPv4Address_t) ((bit<32>) addr1 & (bit<32>) mask);
```

```
// Similar code to above if we can use `bit<32>` operators directly on values of type `IPv4Address_t`

IPv4Address_t addr1;
IPv4Address_t mask;

addr1 = addr1 & mask;
```

With this extra operator support, we could still imagine having lint-like warnings if someone uses the operators on type `PortId_t` that do the funky control plane/data plane numerical translation, but not give such warnings for `IPv4Address_t`, perhaps via an annotation on type `PortId_t` that is not on type `IPv4Address_t`.

mbudiu-vmw:
It would not be particularly difficult to inherit the operations.
The modularity working group is working on type constraints, perhaps if they devise a language for type constraints we can reuse it here to say: the new type inherits only these operations from the basic type.
```
newtype bit<32> IPv4Address_t inherits & ~ |;
```

antoninbas:
I'm also in favor of having fine-grained control over which operators are supported - and if possible, not through annotations since it impacts actual program semantics. The IPv4 example is exactly what @vgurevich brought up at the last meeting. However, my immediate concern is having something that works for translation so I'm happy with the current PR. I think whether we add support for something like "inherits" right away or we punt it and take some time to think about the best syntax should be up to Mihai and the LDWG.

jafingerhut:
@antoninbas @mbudiu-vmw @vgurevich @samar-abdi @ChrisDodd I have attached a zip containing 2 files.  One is the latest p4c/p4include/v1model.p4 file, with a few replacements of type `bit<9>` with `PortId_t` in standard_metadata.  The second is a full (P4_16 + v1model-newtype architecture) program that is a small modification to Antonin's code in this comment: https://github.com/p4lang/p4c/issues/1157#issuecomment-377042731

It compiles with no errors using the latest version of Mihai''s newtype branch code as of this writing.

I found something that seemed odd, where I could not cast a value of type `newtype bit<9> PortId_t` to type `typedef bit<9> PortIdUInt_t` -- the compiler was giving errors whenever I tried to cast just about any type to PortIdUint_t.  Some of the comments in my code mention these things.  I also could not find a way to define `const PSA_CPU_PORT = (PortId_t) any constant expression here`, which seems like a useful thing to have working.

Except for those minor things, and the bigger question of whether to allow most/all arithmetic operations that work on `bit<W>` types to also work on `newtype bit<W> MyNewType_t` types, this is looking reasonable to me.
[psa-portid-using-newtype.zip](https://github.com/p4lang/p4c/files/1948735/psa-portid-using-newtype.zip)

mbudiu-vmw:
The errors that do not involve newtype, do they also happen for the current master?
If so, we should file issues for them.
Otherwise these are bugs introduced by this commit.

vgurevich:
Can we put the text of the proposal and the examples provided @jafingerhut into the pull request so that we can comment on those in context?

I think that we should allow all operations on the new types. Think of the guy who needs to write psa.p4: how he is supposed to know whether it makes sense to allow & or | on the port numbers or not?  He seriously would not know and thus, being a sensible person, he would leave the decision to the PSA user instead and allow all operations anyway.

jafingerhut:
@mbudiu-vmw Updated .zip file containing an edited sample P4_16 program is attached to this comment.  It turns out that I was misspelling the typedef name PortIdUInt_t as PortIdUint_t in many places.  My error, of course, but the compiler error messages for this case are not terribly helpful (and maybe it is difficult in the compiler to make such error messages as helpful as "expected a type specification but found PortIdUint_t instead").

A remaining issue, besides the question of allowing arithmetic ops on newtype types, is that I cannot find a way to make the compiler create a `const` with a newtype type.  For example, either of the attempted const lines below causes the compiler to give an error:
```
newtype bit<9> PortId_t;
const PortId_t PSA_CPU_PORT = (PortId_t) 9w192;
const PortId_t PSA_CPU_PORT = (PortId_t) ((bit<9>) 9w192);

error message:

psa-portid-using-newtype2.p4(16): error: cast: Cannot evaluate initializer for constant
const PortId_t PSA_CPU_PORT = (PortId_t) 9w192; // target-specific
                              ^^^^^^^^^^^^^^^^
```
[psa-portid-using-newtype2.zip](https://github.com/p4lang/p4c/files/1949555/psa-portid-using-newtype2.zip)

vgurevich:
A stupid question: is this allowed:
```
newtype bit<32> bar;
newtype bar foo;
```
?

mbudiu-vmw:
Yes

jafingerhut:
@vgurevich I have created a separate PR with the sample program that is one way of using newtype to define type PortId_t and use it in various parts of a P4 program.  https://github.com/p4lang/p4c/pull/1255   Sure, probably a little less convenient that it is a separate PR, but I am not sure we want to commit those files as-is, anyway, since they contain a modified version of v1model.p4

mbudiu-vmw:
The new PR adds the ability to initialize with constants - but casts are still necessary.
Maybe we'll improve this later.

mbudiu-vmw:
It looks to me like we are running out of reviewers. Lots of PRs stay in this queue for a long time.

ir/type.def
```diff
@@ -498,6 +498,17 @@ class Type_Typedef : Type_Declaration, IAnnotated {
 #nodbprint
 }
 
+/// A newtype is like a typedef, but the type does not inherit any
```
antoninbas: Additionally, `typedef` being just an alias it could probably be eliminated by the frontend, which is not the case for `type`.

PR #1246, opened 2018-04-23
Use correct ID prefixes in P4Runtime serializer

### LINKED ISSUE #1214: Use correct P4Info id prefixes in serializer
We need to use the prefixes defined in p4info.proto: https://github.com/p4lang/PI/blob/master/proto/p4/config/p4info.proto#L78, rather than hardcode the values in the serializer.
Currently, the serializer use the same id for counter / direct counter and for meter / direct meter. This needs to be fixed as well.

antoninbas:
The different values for the ID prefix (first byte of every ID) are now defined in p4info.proto. In this commit, we switch to using these values when generating the P4Info message. One notable difference is that the direct & indirect versions of counters / meters now use a different prefix value for their ids.

Fixes #1214


antoninbas:
@teverman I need your review on this. This CL updates the p4RuntimeSerializer code to use the ID prefixes that we defined with https://github.com/p4lang/PI/pull/336. As a result, after this is merged, generated ids will be different (as the id prefix value has changed for some objects and in particular direct resources).

PR #1255, opened 2018-04-26
Example code using proposed 'newtype' language feature for defining P…

jafingerhut:
…SA PortId_t

### DIFF
--- a/testdata/p4_16_samples/psa-portid-using-newtype2.p4
+++ b/testdata/p4_16_samples/psa-portid-using-newtype2.p4
@@ -1,0 +1,293 @@
+#include "core.p4"
+
+// From psa.p4
+
+#define PORT_BITWIDTH 9  // target-specific
+
+// PortIdUInt_t is only intended for casting values from type PortId_t
+// to an unsigned integer with the same width.
+typedef bit<PORT_BITWIDTH> PortIdUInt_t;
+
+// request translation by P4Runtime agent of all occurrences of PortId_t exposed
+// to control-plane
+@p4runtime_translate("port")
+newtype bit<PORT_BITWIDTH> PortId_t;
+
+//const PortId_t PSA_CPU_PORT = (PortId_t) 9w192; // target-specific
+// With (PortId_t) cast on unsigned int literal, I get this error from
+// newtype branch of p4c:
+//psa-portid-using-newtype1.p4(16): error: cast: Cannot evaluate initializer for constant
+
+//const PortId_t PSA_CPU_PORT = 9w192;  // target-specific
+// With no cast on unsigned int literal, I get this error:
+//psa-portid-using-newtype1.p4(14): error: PSA_CPU_PORT: Cannot unify bit<9> to PortId_t
+//newtype bit<9> PortId_t;
+//                        ^
+//psa-portid-using-newtype1.p4(14)
+//newtype bit<9> PortId_t;
+//               ^^^^^^^^
+
+// This also gives 'Cannot evaluate initializer for constant'.
+//const PortId_t PSA_CPU_PORT = ((PortId_t) ((bit<9>) 9w192));  // target-specific
+
+// This is intended as a workaround until 'const' can be used for
+// PSA_CPU_PORT, or I suppose it could be a long term solution, too.
+#define PSA_CPU_PORT ((PortId_t) 9w192)
+
+// The width of this is _not_ target-specific.  All targets will use
+// this bit width.  This type is only intended for copying data-plane
+// number space PortId_t values into headers to be sent to the CPU
+// port, or for parsing headers containing PortId_t values in certain
+// headers in packets from the CPU port.
+@p4runtime_translate("port")
+newtype bit<32> PortIdInHeader_t;
+
+// used to cast between PortId_t and PortIdInHeader_t the programmer needs to be
+// aware that he may be exposing non-translated values to the control-plane,
+// unless he use @p4runtime_translate appropriately
+
+// Note: all of these casts are necessary with the current newtype
+// code.  Eliminating any one of them results in a compiler error when
+// using the macro as done later in this program.  The nice thing is
+// that this macro could be added to psa.p4, and most people won't
+// need to see it.
+
+#define PSA_PORT_HEADER_TO_ID(p) ((PortId_t) (PortIdUInt_t) (bit<32>) (p))
+#define PSA_PORT_ID_TO_HEADER(p) ((PortIdInHeader_t) (bit<32>) (PortIdUInt_t) (p))
+
+// Included here simply so I can use type PortId_t in my
+// locally-modified version of v1model.p4, called v1model-newtype.p4.
+#include "v1model-newtype.p4"
+
+
+header ethernet_t {
+    bit<48> dstAddr;
+    bit<48> srcAddr;
+    bit<16> etherType;
+}
+
+header ipv4_t {
+    bit<4>  version;
+    bit<4>  ihl;
+    bit<8>  diffserv;
+    bit<16> totalLen;
+    bit<16> identification;
+    bit<3>  flags;
+    bit<13> fragOffset;
+    bit<8>  ttl;
+    bit<8>  protocol;
+    bit<16> hdrChecksum;
+    bit<32> srcAddr;
+    bit<32> dstAddr;
+}
+
+
+// From fabric.p4
+@controller_header("packet_in")
+header packet_in_header_t {
+    // Translation will occur in agent software for this field,
+    // because it is inside of a controller_header, and because
+    // PortIdInHeader_t is defined as a newtype with a
+    // @p4runtime_translate("port") annotation.  No such annotation is
+    // needed here.
+    PortIdInHeader_t ingress_port;
+}
+
+@controller_header("packet_out")
+header packet_out_header_t {
+    // No annotation needed here to get agent translation of
+    // egress_port, for same reason as in header packet_in_header_t
+    // above.
+    PortIdInHeader_t egress_port;
+}
+
+
+struct parsed_headers_t {
+    packet_in_header_t packet_in;
+    packet_out_header_t packet_out;
+    ethernet_t ethernet;
+    ipv4_t ipv4;
+}
+
+struct fabric_metadata_t {
+}
+
+action drop() {
+    mark_to_drop();
+}
+
+action nop() {
+}
+
+parser FabricParser(packet_in packet,
+                    out parsed_headers_t hdr,
+                    inout fabric_metadata_t meta,
+                    inout standard_metadata_t standard_metadata)
+{
+    state start {
+        transition accept;
+    }
+}
+
+control PacketIoIngress(inout parsed_headers_t hdr,
+                        inout fabric_metadata_t fabric_metadata,
+                        inout standard_metadata_t standard_metadata)
+{
+    apply {
+        if (hdr.packet_out.isValid()) {
+            // I know what I'm doing, I requested translation on
+            // packet_out.egress_port
+            standard_metadata.egress_spec =
+                PSA_PORT_HEADER_TO_ID(hdr.packet_out.egress_port);
+            hdr.packet_out.setInvalid();
+            exit;
+        }
+    }
+}
+
+control PacketIoEgress(inout parsed_headers_t hdr,
+                       inout fabric_metadata_t fabric_metadata,
+                       inout standard_metadata_t standard_metadata)
+{
+    apply {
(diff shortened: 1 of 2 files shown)


testdata/p4_16_samples/psa-portid-using-newtype2.p4
```diff
@@ -0,0 +1,293 @@
...
+#define PORT_BITWIDTH 9  // target-specific
+
+// PortIdUInt_t is only intended for casting values from type PortId_t
+// to an unsigned integer with the same width.
+typedef bit<PORT_BITWIDTH> PortIdUInt_t;
+
+// request translation by P4Runtime agent of all occurrences of PortId_t exposed
+// to control-plane
+@p4runtime_translate("port")
+newtype bit<PORT_BITWIDTH> PortId_t;
```
vgurevich: I do not understand why the annotation is required here. Again, maybe I mis-understand the feature, but at least my vision for it was that we are basically saying that variables, that are declared as PortId_t in the P4 program will behave as `bit<PORT_BITWIDTH>` in the data plane program, but will be exposed in the control plane as `control_plane_specific_representation_of(PortId_t)`. 

This "specific representation" consists of (for example):
 - The language-specific type that the control-plane will use for PortId_t (e.g. uint32_t regardless of whether the base type is bit<7> or <bit<9> or bit <17>)
- The control-plane-specific representation format of the underlying values (e.g. an integer (0, 1, 2, 3), or a formatted string such as "eth0", "port-3/5" or anything else)
- The control-plane-specific rules to convert from the control-plane-specific representation of the values into the underlying data plane type (bit<PORT_BITWIDTH>) and vice-versa.

If we agree on that, then there is no need to add the annotation, because translation falls in the category of the possible conversion rules. The data plane program still is unaware of that and can't do much about it (and it should not).

testdata/p4_16_samples/psa-portid-using-newtype2.p4
```diff
@@ -0,0 +1,293 @@
...
+
+//const PortId_t PSA_CPU_PORT = (PortId_t) 9w192; // target-specific
```
vgurevich: Note 1: Even if the control plane conversion procedure involves value translation, this constant must be the *translated* value. The compiler will have no idea of the control-plane translation rules anyway. 

Note 2: It is precisely of the Note 1 that we want to put a cast to PortId_t (or any newtype-derived-type) in front of constants: because we want to ensure that the programmer understands that he/she must translate the values themselves (or that they are appropriate to use for other operations). 

It might be easier to think of what is required if you think of a control plane as a CLI that uses strings to represent different values of PortId_t.

testdata/p4_16_samples/psa-portid-using-newtype2.p4
mbudiu-vmw: I think you are relying on a reserved type name: `PortId_t`. That's fine. I think @jafingerhut does not - so he wants to identify the interesting types somehow.

testdata/p4_16_samples/psa-portid-using-newtype2.p4
```diff
@@ -0,0 +1,293 @@
...
+//psa-portid-using-newtype1.p4(14): error: PSA_CPU_PORT: Cannot unify bit<9> to PortId_t
+//newtype bit<9> PortId_t;
+//                        ^
+//psa-portid-using-newtype1.p4(14)
+//newtype bit<9> PortId_t;
+//               ^^^^^^^^
+
+// This also gives 'Cannot evaluate initializer for constant'.
+//const PortId_t PSA_CPU_PORT = ((PortId_t) ((bit<9>) 9w192));  // target-specific
+
```
vgurevich: Out of all 3 options, I think all should work, but that the first one is the best.
The second is OK too -- it should generate a warning, but I believe it should still be legal
The third option is clearly a workaround that should still work, but doesn't seem a practical application, since the type of 9w192 must be bit<9> by definition.

testdata/p4_16_samples/psa-portid-using-newtype2.p4
```diff
@@ -0,0 +1,293 @@
...
+// PSA_CPU_PORT, or I suppose it could be a long term solution, too.
+#define PSA_CPU_PORT ((PortId_t) 9w192)
+
+// The width of this is _not_ target-specific.  All targets will use
+// this bit width.  This type is only intended for copying data-plane
+// number space PortId_t values into headers to be sent to the CPU
+// port, or for parsing headers containing PortId_t values in certain
+// headers in packets from the CPU port.
+@p4runtime_translate("port")
+newtype bit<32> PortIdInHeader_t;
```
vgurevich: Same comment as on the annotation to PortId_t. I do not believe it should be necessary at all.

testdata/p4_16_samples/psa-portid-using-newtype2.p4
```diff
@@ -0,0 +1,293 @@
...
+
+// Note: all of these casts are necessary with the current newtype
+// code.  Eliminating any one of them results in a compiler error when
+// using the macro as done later in this program.  The nice thing is
+// that this macro could be added to psa.p4, and most people won't
+// need to see it.
+
+#define PSA_PORT_HEADER_TO_ID(p) ((PortId_t) (PortIdUInt_t) (bit<32>) (p))
+#define PSA_PORT_ID_TO_HEADER(p) ((PortIdInHeader_t) (bit<32>) (PortIdUInt_t) (p))
+
```
vgurevich: I think that the simple 
```
#define PSA_PORT_HEADER_TO_ID(p) ((PortId_t)(p))
#define PSA_PORT_ID_TO_HEADER(p) ((PortIdHeader_t)(p))
```
should be enough. If so, perhaps these macros should not be needed.

testdata/p4_16_samples/psa-portid-using-newtype2.p4
```diff
@@ -0,0 +1,293 @@
...
+control PacketIoIngress(inout parsed_headers_t hdr,
+                        inout fabric_metadata_t fabric_metadata,
+                        inout standard_metadata_t standard_metadata)
+{
+    apply {
+        if (hdr.packet_out.isValid()) {
+            // I know what I'm doing, I requested translation on
+            // packet_out.egress_port
+            standard_metadata.egress_spec =
+                PSA_PORT_HEADER_TO_ID(hdr.packet_out.egress_port);
```
vgurevich: Given the comment above, I think that the simple cast should be enough and that this will be even more readable than the macro.
```
standard_metada.egress_spec = (PortId_t)hdr.packet_out.egress_port;
```

testdata/p4_16_samples/psa-portid-using-newtype2.p4
```diff
@@ -0,0 +1,293 @@
...
+    }
+}
+
+control Filtering (inout parsed_headers_t hdr,
+                   inout fabric_metadata_t fabric_metadata,
+                   inout standard_metadata_t standard_metadata)
+{
+    table t {
+        key = {
+            // translated by P4Runtime agent because of annotation on PortId_t
```
vgurevich: The comment is totally correct, but that's precisely the point: translated or not, P4 tables do not care.

testdata/p4_16_samples/psa-portid-using-newtype2.p4
```diff
@@ -0,0 +1,293 @@
...
+        actions = { drop; fwd; }
+    }
+
+    apply {
+        t.apply();
+
+        // One can perform arithmetic operations on PortId_t values if
+        // you use enough casts to convert it to an unsigned int, do
+        // the arithetic ops, then cast back to PortId_t.
+
```
vgurevich: In my view, this should not be required. We must be able to operate on the values of PortId_t as if they are bit<PORT_BITWIDTH>.

testdata/p4_16_samples/psa-portid-using-newtype2.p4
jafingerhut: PortId_t is just one of 3 types defined in PSA that has target-specific bit widths, each with a different numerical translation we expect the agent to be able to do: PortId_t, ClassOfService_t, and Timestamp_t.

There are 4 more types defined in PSA that we want target-specific bit widths, but no numerical translation in the agent: MulticastGroup_t, CloneSessionId_t, PacketLength_t, EgressInstance_t

You have mentioned in the meetings types like IPv4Address, MACAddress, IPv6Address, where you would like the control plane to know about these types in a P4 program, so it can provide custom debug and/or string<->numeric value conversions.

This set of types will only grow in the future, yes?

By what mechanism do you propose to inform the compiler about the properties of these types, if not annotations on the 'newtype' or 'typedef' definition of the type?

testdata/p4_16_samples/psa-portid-using-newtype2.p4
jafingerhut: The '// target-specific' comment on that line is intended to convey that every target can define its own custom data-plane-specific value for PSA_CPU_PORT.  For any one given target, this value is a constant.  Across different targets, different constants are expected to be used.  That line belongs in a target-specific psa.p4 include file.  Sorry if that wasn't clear in the example code.

testdata/p4_16_samples/psa-portid-using-newtype2.p4
vgurevich: Yes. This is correct.

mbudiu-vmw:
I have renamed 'newtype' to just 'type' in the latest PR.

mbudiu-vmw:
@jafingerhut @antoninbas : I will submit a new PR for the 'newtype' feature, it will also include these 2 files modified to use the new keyword `type`. After that we can probably close this PR.

testdata/p4_16_samples/psa-portid-using-newtype2.p4
jafingerhut: After a bug fix in p4c, the first const definition now works.

testdata/p4_16_samples/psa-portid-using-newtype2.p4
jafingerhut: With the current p4c, unfortunately your proposed alternate definitions are not enough.

I don't have any bright ideas to make it work with a single cast, but with the macro definitions, most people will not need to know that 3 casts are in use, anyway.

testdata/p4_16_samples/psa-portid-using-newtype2.p4
jafingerhut: Again, I don't have any good ideas on how to change the compiler so that one cast is enough, but the macro definition as proposed does do what is needed here.

testdata/p4_16_samples/psa-portid-using-newtype2.p4
jafingerhut: Given the current p4c compiler as of 2018-Aug-19, it is not possible to do arithmetic on two values of type PortId_t directly.  That is something that could be enhanced later, if the compiler is changed to support arithmetic on two values with the same `type` name, and produces a value with that same `type` name type.

testdata/p4_16_samples/psa-portid-using-newtype2.p4
jafingerhut: Given that a P4Runtime server will likely need to have at least a little bit of PSA-specific code in it to work with a PSA target, whether it has PSA-specific code to recognize a list of type names like `PortId_t` and `ClassOfService_t`, or annotation keys with specific names, seems to be six of one vs. half a dozen of another.  Both seem like they could be made to work about as easily as each other, and I can see reasons why one might prefer to identify these types by type names like `PortId_t` rather than an annotation key name.

mbudiu-vmw:
What do we have to do to make this pass tests?

jafingerhut:
@mbudiu-vmw Long term we want a way to run automated tests for p4c with the PSA back end.  The PSA back end may not be ready for that yet (I'm not sure).

I am guessing that updating p4include/psa.p4 to use `type` definitions like we plan to use in PSA, and having some test P4 program #include that, might be the same as running an automated test with p4c plus PSA back end.

In the mean time, if you want an automated test that exercises `type` in a way similar to the program in the PR, do you think it exercises anything about the `type` feature in the compiler that existing test programs for `type` do not?

mbudiu-vmw:
I'd rather have redundant tests than no tests for some features.

mbudiu-vmw:
@jafingerhut : do you plan to fix this or should we close it?

jafingerhut:
This recently merged PR https://github.com/p4lang/p4c/pull/1792 includes the only program in this PR as a test case, so I will close this one now.  It is no longer necessary.

PR #1256, opened 2018-04-26
Add correct support for digest in p4RuntimeSerializer

antoninbas:
See each individual commit message for details. Here are the different changes:
 * Add support for generating `P4DataTypeSpec` messages from `IR::Type` nodes (https://github.com/p4lang/PI/pull/328)
 * Generate correct P4Info information for v1model digest method calls (https://github.com/p4lang/PI/pull/337)
 * Remove `HeaderFieldPath` class in p4RuntimeSerializer, which was introduced back when we needed to generate names for IR path expressions
 * Remove p4runtime-to-pd code generator; AFAIK nobody is using this and I have no interest in maintaining it.


antoninbas:
@mbudiu-vmw @cc10512 the first commit is probably the most interesting to review, the fourth commit is just a bunch of code being removed.

control-plane/typeSpecConverter.cpp
```diff
@@ -0,0 +1,243 @@
...
+    auto decl = refMap->getDeclaration(type->path, true);
+    auto name = decl->controlPlaneName();
+    if (decl->is<IR::Type_Struct>()) {
+        typeSpec->mutable_struct_()->set_name(name);
+    } else if (decl->is<IR::Type_Header>()) {
+        typeSpec->mutable_header()->set_name(name);
+    } else if (decl->is<IR::Type_HeaderUnion>()) {
+        typeSpec->mutable_header_union()->set_name(name);
+    } else if (decl->is<IR::Type_Enum>()) {
+        typeSpec->mutable_enum_()->set_name(name);
```
cc10512: Nit: why sometimes you suffix the method name with `_` and sometimes you do not?

control-plane/typeSpecConverter.cpp
```diff
@@ -0,0 +1,243 @@
...
+            auto headerUnionTypeSpec = new p4::P4HeaderUnionTypeSpec();
+            for (auto f : type->fields) {
+                auto fType = f->type;
+                visit(fType);
+                auto fTypeSpec = map.at(fType);
+                BUG_CHECK(fTypeSpec->has_header(),
+                          "Only header fields expected in header union declaration %1%", type);
+                auto member = headerUnionTypeSpec->add_members();
+                member->set_name(f->controlPlaneName());
+                member->mutable_header()->CopyFrom(fTypeSpec->header());
```
cc10512: Are header_unions just as structs?

control-plane/typeSpecConverter.cpp
antoninbas: This is protobuf-generated code. When you use a field name in the protobuf definition that happens to be a reserved keyword in the language you are generating to, they append a `_`.

control-plane/typeSpecConverter.cpp
antoninbas: They are similar enough that it's tempting to unify the code, yet different enough that I couldn't find a good way to do it...

cc10512: Thanks! I can see the extra checks for members of a header union are only headers.

PR #1257, opened 2018-04-26
Added Another Policy to SimplifyKey

evanches:
Prevent potentially full tables to run SimplifyKey if a target wished to do so.


evanches:
I like that Visitor::Context thing.  That is a lot better than my solution.

PR #1260, opened 2018-04-27
Support v1model registers in P4RuntimeSerializer

### LINKED ISSUE #1248: Generate P4Info register information in p4RuntimeSerializer

antoninbas:
Update p4lang/PI submodule ref point

Support v1model registers in p4RuntimeSerializer
Fixes #1248


antoninbas:
@mbudiu-vmw I wasn't expecting to find `Type_SpecializedCanonical` in the IR based on this comment in ir/types.def:
> Canonical representation of a Type_Specialized; only used by the type-checker, never in the IR tree
Is that normal?

antoninbas:
Which looks right since the evaluator instantiates this field by doing a lookup in the type map :)
https://github.com/p4lang/p4c/blob/master/frontends/p4/evaluator/evaluator.cpp#L235

mbudiu-vmw:
I see, blocks are also not part of the program IR, so this is fine. They are the result produced by the evaluator, which is not stored as part of the program at any time.
I didn't realize that you are talking about a block.

control-plane/p4RuntimeSerializer.cpp
```diff
@@ -181,6 +181,50 @@ struct Digest {
     const p4::P4DataTypeSpec* typeSpec;  // The format of the packed data.
 };
 
+struct Register {
+    const cstring name;  // The fully qualified external name of this register.
+    const IR::IAnnotated* annotations;  // If non-null, any annotations applied
```
mbudiu-vmw: I think IAnnotated is the object which is annotated, not the annotations themselves. Perhaps this should be IR::Annotations instead.

antoninbas: I'm just following what the existing code is doing. We actually store the declaration as `IR::IAnnotated` and then we have a `addAnnotations` method later (when creating the protobuf message) that retrieves the annotations.
I modified my gtest to use an annotation to make sure things were working correctly though :)

PR #1271, opened 2018-05-04
Support struct as type parameter in value_set<T>

hanw:
as mentioned in https://github.com/p4lang/p4-spec/pull/602

### DIFF
--- a/frontends/p4/typeChecking/typeChecker.cpp
+++ b/frontends/p4/typeChecking/typeChecker.cpp
@@ -2854,7 +2854,7 @@ void TypeInference::explodeStructType(const IR::Type_StructLike* structType, IR:
         } else if (auto ft = field->type->to<IR::Type_StructLike>()) {
             explodeStructType(ft, tuple);
         } else {
-            ::error("Unhandled type %1% in explodeStructType", field->type);
+            BUG("Unexpected type %1% for struct field %2%", field->type, field);
         }
     }
 }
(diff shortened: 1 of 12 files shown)


jafingerhut:
@hanw Is the idea here that the transition select expression can be any 'flat' list of expressions, and with this change you plan to allow those expressions to correspond to fields of a struct, perhaps even a nested struct arbitrarily deep, in the order that the fields appear in the definitions of the structs?

For example, if you have these struct definitions:
```
struct level2_t {
bit<8> L2_a;
bit<16> L2_b;
}
struct level1_t {
bit<3> L1_a;
level2_t L1_b;
bit<10> L1_c;
}
struct level0_t {
level1_t L0_a;
bit<7> L0_b;
}
```
Then you could declare a parser value set with type `level0_t`, and a transition select statement with 5 comma-separated expressions would correspond to the fields `L0_a.L1_a, L0_a.L1_b.L2_a,  L0_a.L1_b.L2_b, L0_a.L1_c, L0_b`, in that order, because they appear in that order in the struct definitions?

hanw:
@jafingerhut yes, that's the idea. I realized in P4-16, there is no order in struct. but there is no better way to give names to the value_set fields, unless we go down the path with a table-like syntax.

frontends/p4/typeChecking/typeChecker.cpp
```diff
@@ -2847,6 +2847,18 @@ const IR::Node* TypeInference::postorder(IR::ConstructorCallExpression* expressi
     return expression;
 }
 
+void TypeInference::explodeStructType(const IR::Type_StructLike* structType, IR::Type_Tuple *tuple) {
```
mbudiu-vmw: Maybe you can make this a static file-level method (with no class), since it does not depend on "this".
I would call it "convertStructToTuple"

PR #1275, opened 2018-05-09
Improved documentation for two passes

mbudiu-vmw:


mbudiu-vmw:
Could we get a review for this PR? It is just comments, so it should be easy.

mbudiu-vmw:
I am afraid you don't have write access, so your approval is not sufficient to allow me to merge this.

PR #1278, opened 2018-05-10
DirectCounter min_width

evanches:
Some annotations on Counters were not saved on DirectCounters


cc10512:
@evanches you need to update the outputs.

Issue #1273, opened 2018-05-09
JSON to P4


aneeshakella17:
Does the code handle JSON to P4 conversion? If not, how feasible do you think it would be to implement?

mbudiu-vmw:
What is P4i?

aneeshakella17:
Hi, 
Yes, I would really be interested in seeing the converter. P4i is an intermediate representation. When you convert P4 to JSON, a P4i file should be generated as well.

aneeshakella17:
Just an update @debatgithub as to when and if you can share the converter

TakeshiTseng:
Hi,
You mean `p4i` generated by `p4c`? (e.g. `p4c -b bmv2-v1model my-p4-code.p4`)

This `p4i` is not the `IR` between Frontend and Backend compiler. It's like the output from preprocessor such as `cc` or `gcc`

TakeshiTseng:
By the way you can find how `p4c` generate `p4i` here:
https://github.com/p4lang/p4c/blob/master/tools/driver/p4c_src/p4c.bmv2.cfg#L40-L42

It actually execute this command:
`cc -E -x c -I /usr/local/share/p4c/p4include -o ./my-p4-code.p4i my-p4-code.p4`

hanw:
The `-###` flag on `p4c` will print the commands executed by p4c.

aneeshakella17:
Thank you, but I actually need something to convert JSON to some P4 format (p4-16, p4-14) or p4i.

hanw:
In the case of bmv2, it runs the c preprocessor and the p4c-bm-ss backend.

hanw:
If I understand correctly, p4i is just the preprocessed p4 source code, we could have named it my-p4-code.preprocessed.p4

aneeshakella17:
I'm performing a project where JSON manipulation was a key element. I was hoping to  convert the manipulated JSON back into a p4 program.

aneeshakella17:
I don't think --dumpJSON is a given option

mbudiu-vmw:
It is `--toJSON`.
We should file an issue, because the documentation says:
`--toJSON file                           Dump IR to JSON in the specified file.`
But it does not say when the IR is generated.

mbudiu-vmw:
By looking at the code the JSON seems to be generated just before the back-end is entered.

hesingh:
[...]

PR #1282, opened 2018-05-14
add --fromJSON option to p4test to load previously dumped json

ChrisDodd:
- fix json reading of strings with escaped quotes

for #1273

### DIFF
--- a/backends/p4test/p4test.cpp
+++ b/backends/p4test/p4test.cpp
@@ -37,6 +37,7 @@ class P4TestOptions : public CompilerOptions {
  public:
     bool parseOnly = false;
     bool validateOnly = false;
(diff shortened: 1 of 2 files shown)


backends/p4test/p4test.cpp
```diff
@@ -49,6 +50,12 @@ class P4TestOptions : public CompilerOptions {
                            return true;
                        },
                        "Validate the P4 input, running just the front-end");
+        registerOption("--fromJSON", nullptr,
```
mbudiu-vmw: Shouldn't we make this an option for all front-ends?

ChrisDodd:
Different backends probably want to run different passes on the loaded json and/or inject it into a different point in the pipeline.

PR #1283, opened 2018-05-15
Extend toString methods for Operation expressions

cole-barefoot:
IR nodes are equipped with a toString method intended to pretty-print the node.
The IR::Operation nodes, however, mostly inherit toString from IR::Operation,
which simply prints the operation string encoding, eg. "+" for addition,
without including the operands.

This commit changes the toString methods for Operation nodes to include
operands, eg. "left + right" rather than "+", and adds a gtest confirming that the
toString methods work as expected.

### DIFF
--- a/ir/expression.def
+++ b/ir/expression.def
@@ -12,6 +12,7 @@ abstract Operation_Unary : Operation {
         if (!srcInfo && expr) srcInfo = expr->srcInfo;
         if (type->is<Type::Unknown>() && expr) type = expr->type; }
     precedence = DBPrint::Prec_Prefix;
+    toString { return getStringOp() + expr->toString(); }
 }
 
 class Neg : Operation_Unary {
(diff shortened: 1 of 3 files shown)


cole-barefoot:
Edit: I'm not sure whether this is the right way to go.  I would have expected `toString` to print an IR subtree, not just the node, but the error message differences I'm seeing on Travis say differently.  My inclination would be to use `getStringOp()` to get the string representation of the operation itself vs. `toString()` to pretty-print the IR subtree.

ChrisDodd:
Mihai's summary is correct -- toString is intended to produce short descriptions appropriate for error messages to the P4 programmer.  dbprint is intended to pretty-print the IR node, appropriate for (internal) logging messages to aid in debugging the compiler.

cole-barefoot:
Hm, good to know.  The specific problem I ran into is that calling `toString()` on a Member projection of an ArrayIndex, eg. `foo.bar[3].f`, prints the ArrayIndex as `"[]"`, which results in `"[].f"`.  Maybe the thing to do is just to override the `toString()` method for ArrayIndex, but not all operations.

mbudiu-vmw:
What is the context in which you are calling toString()? Normally you should need to do this only when calling error. This behavior is also intentional, because the array index could be a complicated expression itself.

Note that the compiler will show the source-code fragment using carets `^^^^`, so showing only braces is usually not a problem, because the context of the error is displayed to the user.

cole-barefoot:
I see.  I'll write a visitor to get the pretty-printing I'd like.  Thanks for the clarification.

cole-barefoot:
The ToP4 visitor might do the trick.  Thanks!

PR #1284, opened 2018-05-17
Fix def_use/SimplifyDefUse to deal with extern methods properly

ChrisDodd:
- Track extern method calls, assuming that any call to any
  non-abstract method might trigger any abstract method implementation
- local_copyprop fixed for the same assumption.

### DIFF
--- a/frontends/common/resolveReferences/resolveReferences.cpp
+++ b/frontends/common/resolveReferences/resolveReferences.cpp
@@ -404,10 +404,15 @@ void ResolveReferences::postorder(const IR::Type_Method *t) {
 
 bool ResolveReferences::preorder(const IR::Type_Extern *t) {
     refMap->usedName(t->name.name);
-    addToContext(t->typeParameters); return true; }
+    // FIXME -- should the typeParamters be part of the extern's scope?
+    addToContext(t->typeParameters);
+    addToContext(t);
+    return true; }
 
 void ResolveReferences::postorder(const IR::Type_Extern *t) {
-    removeFromContext(t->typeParameters); }
+    removeFromContext(t);
+    removeFromContext(t->typeParameters);
+    }
 
 bool ResolveReferences::preorder(const IR::ParserState *s) {
     refMap->usedName(s->name.name);
--- a/frontends/p4/def_use.cpp
+++ b/frontends/p4/def_use.cpp
@@ -656,26 +656,29 @@ bool ComputeWriteSet::preorder(const IR::MethodCallExpression* expression) {
     }
 
     // Symbolically call some apply methods (actions and tables)
-    const IR::Node* callee = nullptr;
+    std::vector<const IR::IDeclaration *> callee;
     if (mi->is<ActionCall>()) {
         auto action = mi->to<ActionCall>()->action;
-        callee = action;
+        callee.push_back(action);
     } else if (mi->isApply()) {
         auto am = mi->to<ApplyMethod>();
         if (am->isTableApply()) {
             auto table = am->object->to<IR::P4Table>();
-            callee = table;
+            callee.push_back(table);
         }
-    }
-
-    if (callee != nullptr) {
-        LOG3("Analyzing " << dbp(callee));
+    } else if (auto em = mi->to<ExternMethod>()) {
+        // symbolically call all the methods that might be called via this extern method
+        callee = em->mayCall(); }
+    if (!callee.empty()) {
+        LOG3("Analyzing " << DBPrint::Brief << callee << DBPrint::Reset);
         ProgramPoint pt(callingContext, expression);
         ComputeWriteSet cw(this, pt, currentDefinitions);
-        (void)callee->apply(cw);
+        for (auto c : callee)
+            (void)c->getNode()->apply(cw);
         currentDefinitions = cw.currentDefinitions;
         exitDefinitions = exitDefinitions->joinDefinitions(cw.exitDefinitions);
-        LOG3("Definitions after call of " << expression << ": " << currentDefinitions);
+        LOG3("Definitions after call of " << DBPrint::Brief << expression << ": " <<
+             currentDefinitions << DBPrint::Reset);
     }
 
     auto result = LocationSet::empty;
@@ -752,11 +755,10 @@ bool ComputeWriteSet::preorder(const IR::P4Control* control) {
     enterScope(control->getApplyParameters(), &control->controlLocals, startPoint);
     exitDefinitions = new Definitions();
     returnedDefinitions = new Definitions();
-    for (auto l : control->controlLocals) {
-        if (l->is<IR::Declaration_Instance>())
-            visit(l);  // process virtual Functions if any
-    }
+    BUG_CHECK(controlLocals == nullptr, "Nested controls?");
+    controlLocals = &control->controlLocals;
     visit(control->body);
+    controlLocals = nullptr;;
     auto returned = currentDefinitions->joinDefinitions(returnedDefinitions);
     auto exited = returned->joinDefinitions(exitDefinitions);
     return setDefinitions(exited, control->body);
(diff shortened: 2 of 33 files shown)


frontends/p4/simplifyDefUse.cpp
```diff
@@ -367,20 +367,34 @@ class FindUninitialized : public Inspector {
...
+        if (auto ac = mcd.instance->to<ActionCall>()) {
+            callee = ac->action;
         } else if (mcd.instance->isApply()) {
             auto am = mcd.instance->to<ApplyMethod>();
             if (am->isTableApply()) {
                 auto table = am->object->to<IR::P4Table>();
                 callee = table;
             }
-        }
+        } else if (auto em = mcd.instance->to<ExternMethod>()) {
```
mbudiu-vmw: This code seems to be the same as in def_use.cpp. Maybe it can be refactored to be written only once?

midend/local_copyprop.cpp
```diff
@@ -419,16 +419,22 @@ IR::MethodCallExpression *DoLocalCopyPropagation::postorder(IR::MethodCallExpres
...
                     return mc;
                 } else {
-                    LOG3("extern method call " << mc->method << " does nothing");
-                    // FIXME -- is this safe? If an abstract method in outer scope that
-                    // we haven't seen a definition for, it might affects non-locals?
+                    // extern method call might trigger any concrete implementation
+                    // method in the object, so act as if all of them are applied
+                    LOG3("extern method call " << name);
+                    name = obj + '.';
+                    for (auto it = methods.upper_bound(name); it != methods.end(); ++it) {
```
mbudiu-vmw: I don't think this is conservative enough. The virtual methods could also be invoked asynchronously at any time, e.g., by a request from the control-plane. We have not agreed that abstract methods are only invoked when some other extern method is invoked. 

Also, consider action profiles: the action selector will be invoked when a table is invoked, without any other method being called explicitly. 

So I think you should always consider the abstract methods as "live" and scan them. You know nothing about the context they execute in, though; they could all execute, or one of them could execute multiple times. This code should not be necessary, you should always visit them.

testdata/p4_16_samples/virtual3.p4
```diff
@@ -0,0 +1,55 @@
...
+    bit<16> total();
+}
+
+// User code
+control c(inout bit<16> p) {
+    bit<16>     local;
+
+    Virtual() cntr = {  // implementation
+        bit<16> f(in bit<16> ix) {  // abstract method implementation
+            return (ix + local);
```
mbudiu-vmw: @ChrisDodd : I have an issue filed to forbid access to the context in an abstract method. We cannot give a semantics to this code. I don't think we should allow this.

ChrisDodd: Unfortunately we can't do that -- the whole point of abstract methods in per-control instances is to get access to the context -- if we disallow that, we might as well require all externs instances be global.

ChrisDodd:
I thin what we really need is a hook for backends to provide per-extern (and per method) information about when the method(s) might be invoked, so as to allow proper flow analysis.  So this is incomplete, but an improvement over the existing code which assumed that abstract methods could never be called in the middle of the control apply block.

frontends/p4/simplifyDefUse.cpp
```diff
@@ -103,13 +104,13 @@ class FindUninitialized : public Inspector {
...
         CHECK_NULL(hasUses);
-        setName("FindUninitialized"); }
+        visitDagOnce = false; }
 
     // we control the traversal order manually, so we always 'prune()'
     // (return false from preorder)
 
     bool preorder(const IR::ParserState* state) override {
-        LOG3("FU Visiting " << dbp(state));
+        LOG3("FU Visiting state " << state->name);
```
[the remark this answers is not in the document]
ChrisDodd: I find the id rarely useful, and dbp also just prints the id and type and none of the contents of the node (like the state name), making it mostly useless in my experience.  But of course, debugging methods differ, so YMMV.

mbudiu-vmw:
What is wrong with my proposal for a `@synchronous` annotation?

ChrisDodd:
I don't think @synchronous is enough.  We could create a little language in annotations -- annotate each extern method with the set of other methods it could invoke.  If an asynchronous method could be invoked anywhere, I'm not sure how to deal with it in these passes, however.

ChrisDodd:
So I've updated the code to use @synchronous annotations to indicate which methods might call abstract methods, so accesses to in-scope vars can be anayzed a bit better.  I would like to check this in as it fixes some problems we see with dead-code eliminating metadata fields that are only used by abstract method implementations.  @mbudiu-vmw any other comments?

mbudiu-vmw:
On vacation this week with no laptop

frontends/common/resolveReferences/resolveReferences.cpp
```diff
@@ -404,10 +404,15 @@ void ResolveReferences::postorder(const IR::Type_Method *t) {
 
 bool ResolveReferences::preorder(const IR::Type_Extern *t) {
     refMap->usedName(t->name.name);
-    addToContext(t->typeParameters); return true; }
+    // FIXME -- should the typeParamters be part of the extern's scope?
```
mbudiu-vmw: I don’t think it will work without them.

frontends/p4/def_use.cpp
```diff
@@ -932,3 +934,13 @@ bool ComputeWriteSet::preorder(const IR::MethodCallStatement* statement) {
 }
 
 }  // namespace P4
+
+// functions for calling from gdb
+void dump(const P4::StorageLocation *s) { std::cout << *s << std::endl; }
```
mbudiu-vmw: If these implement IHasDbPrint there should already exist a dump, but indeed gdb sometimes has trouble finding the dump.

frontends/p4/methodInstance.h
```diff
@@ -129,6 +129,11 @@ class ExternMethod final : public MethodInstance {
     const IR::Method*      method;
     const IR::Type_Extern* originalExternType;    // type of object method is applied to
     const IR::Type_Extern* actualExternType;      // with type variables substituted
+
+    /// Set of IR::Method and IR::Function objects that may be called by this method.
+    // If this method is abstract, will consist of (just) the concrete implemenation,
```
mbudiu-vmw: Typo

midend/local_copyprop.cpp
```diff
@@ -409,12 +441,19 @@ IR::MethodCallExpression *DoLocalCopyPropagation::postorder(IR::MethodCallExpres
         if (actions.count(fn->path->name)) {
             LOG3("action method call " << mc->method);
             apply_function(&actions[fn->path->name]);
+            return mc;
+        } else if (fn->path->name == "verify" || fn->path->name == "hash") {
```
mbudiu-vmw: Hash is not part of the standard, you should not hardwire its name here. You may want another annotation on methods to indicate that they have no side-effects. E.g. `@pure`

testdata/p4_16_samples/virtual3.p4
mbudiu-vmw: Do you need an annotation here?

testdata/p4_16_samples_outputs/array_field1-midend.p4
```diff
@@ -6,18 +6,29 @@ extern bit<1> f(inout bit<1> x, in bit<1> b);
 control c(out H[2] h);
 package top(c _c);
 control my(out H[2] s) {
+    bit<32> a_0;
```
mbudiu-vmw: These changes suggest that some analysis has become less precise.

ChrisDodd: The code here changed because we're now assuming that the unknown method call `f` might depend on the locals.  However, f is a global extern function, so we should probably assume it only depends on/modifies things passed to it explicitly.

frontends/common/resolveReferences/resolveReferences.cpp
ChrisDodd: Currently they're not part of the Type_Extern's scope, so need to be added as a separate scope (which is what this part of the change does)

frontends/p4/def_use.cpp
ChrisDodd: Yes, gdb seems to be unable to deal with virtual base classes properly (or at least to be able to call function overloads based on them), so I ended up adding all these overloads to make them callable.

testdata/p4_16_samples/virtual3.p4
ChrisDodd: The @synchronous annotation was added in to the abstract method declaration in the extern above.

ChrisDodd:
IMO we should be doing more in the architecture P4 file to describe how the architecture works -- perrhaps (simplified) bodies for extern methods and functions that summarize what they do and can be analyzed by common frontend/midend passes to drive their transformations.  I'm not a big fan of duplicating stuff between C++ classes and architecture files in a way that needs to be carefully managed but is subject to breakage when someone makes a change -- all dependencies of this nature should be strongly delineated by automatic tests that will diagnose the problem automatically when someone breaks something.

apinski-cavium:
I just noticed `@synchronous` is not even documented anywhere except here and in the source of p4c.
Seems like there could be some user documentation missing here ...

mbudiu-vmw:
@apinski-cavium we have long wanted to do that in the spec, but never got around to it: https://github.com/p4lang/p4-spec/issues/943

PR #1287, opened 2018-05-18
Do not eliminate annotated states.

Stumble:
Annotations on parser state may be useful for certain midend and backend. Do not eliminate the parser state that has an annotation other than `name`(because it is always added), but not invoked by any other parser state.

### DIFF
--- a/frontends/p4/simplifyParsers.cpp
+++ b/frontends/p4/simplifyParsers.cpp
@@ -55,6 +55,12 @@ class RemoveUnreachableStates : public Transform {
         if (state->name == IR::ParserState::start ||
             state->name == IR::ParserState::reject)
             return state;
+
+        // Do not eliminate states with annotations other than name.
+        for (const auto* anno : state->getAnnotations()->annotations) {
+            if (anno->name.name != "name") {
+                return state; } }
+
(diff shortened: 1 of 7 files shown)


hanw:
You may want to update the reference output as well.
P4TEST_REPLACE=1 make check-p4

Stumble:
no output changed.
```
100% tests passed, 0 tests failed out of 519

```

frontends/p4/simplifyParsers.cpp
```diff
@@ -55,6 +55,11 @@ class RemoveUnreachableStates : public Transform {
         if (state->name == IR::ParserState::start ||
             state->name == IR::ParserState::reject)
             return state;
+        // Do not eliminate states with annotations other than name.
+        if (state->getAnnotations()->annotations.size() > 1
```
mbudiu-vmw: The comment does not reflect what the code is doing. There is no check that there is any annotation that is not "name". Note that nothing prevents an object having multiple annotations named in the same way.

Stumble:
@mbudiu-vmw Thanks! just changed.

hanw:
If the change does not impact any existing test cases, we may need a new test case.

PR #1288, opened 2018-05-18
Support named arguments in frontend and midend

mbudiu-vmw:
This is the second PR in a series for supporting named arguments.
This should take care of most of the cases.

In an upcoming PR we will also fix the backends to properly handle named arguments.
The PR does not handle default values for parameters or overload resolution. These will be other future PRs.


mbudiu-vmw:
The PR has been augmented with an additional commit which adds support for named arguments in BMv2. This is done through a new midend pass which reorders the named arguments to always be in the order of the parameters. This only works if there are no optional arguments anywhere (and v1model does not have optional arguments). For other architectures the back-end should never reach into the argument list by position, but should instead use a ParameterSubstitution and find arguments by the name of the corresponding parameter. There were too many such places to change in the BMv2 back-end.

PR #1289, opened 2018-05-21
Emit extern functions with --emit-externs

omeranson:
This change builds on top of pull request #1217 to also support emitting
extern functions (in addition to objects and methods).

It assumes extern functions have the __extern_ prefix, to avoid name
collision.

### DIFF
--- a/backends/bmv2/CMakeLists.txt
+++ b/backends/bmv2/CMakeLists.txt
@@ -149,6 +149,8 @@ set (XFAIL_TESTS
   testdata/p4_14_samples/issue604.p4
   # This test uses an incorrect model
   testdata/p4_16_samples/issue1205-bmv2.p4
+  # This test requires --emit-externs and extern support from BMv2
+  testdata/p4_16_samples/extern-funcs-bmv2.p4
   # These psa tests are not ready to run on bmv2 yet
   testdata/p4_16_samples/psa-example-counters-bmv2.p4
   testdata/p4_16_samples/psa-example-digest-bmv2.p4
@@ -158,6 +160,8 @@ set (XFAIL_TESTS
 
 if (HAVE_SIMPLE_SWITCH)
   p4c_add_tests("bmv2" ${BMV2_DRIVER} "${BMV2_V1MODEL_TEST_SUITES}" "${XFAIL_TESTS}")
+  add_library(extern_func_module MODULE EXCLUDE_FROM_ALL "${P4C_SOURCE_DIR}/testdata/extern_modules/extern-funcs-bmv2.cpp" )
+  p4c_add_test_with_args("bmv2" ${BMV2_DRIVER} FALSE "bmv2_emit_externs" "testdata/p4_16_samples/extern-funcs-bmv2.p4" "-a --emit-externs --target-specific-switch-arg \"--load-modules $<TARGET_FILE:extern_func_module>\" --init \"make extern_func_module\"")
 else()
   MESSAGE(WARNING "BMv2 simple switch is not available, not adding v1model BMv2 tests")
 endif()
--- a/backends/bmv2/run-bmv2-test.py
+++ b/backends/bmv2/run-bmv2-test.py
@@ -51,10 +51,13 @@ class Options(object):
         self.verbose = False
         self.replace = False            # replace previous outputs
         self.compilerOptions = []
+        self.switchOptions = []
+        self.switchTargetSpecificOptions = []
         self.hasBMv2 = False            # Is the behavioral model installed?
         self.usePsa = False             # Use the psa switch behavioral model?
         self.runDebugger = False
         self.observationLog = None           # Log packets produced by the BMV2 model if path to log is supplied
+        self.initCommands = []
 
 def nextWord(text, sep = " "):
     # Split a text at the indicated separator.
@@ -112,9 +115,12 @@ def usage(options):
     print("          -p: use psa switch")
     print("          -f: replace reference outputs with newly generated ones")
     print("          -a option: pass this option to the compiler")
+    print("          --switch-arg option: pass this general option to the switch")
+    print("          --target-specific-switch-arg option: pass this target-specific option to the switch")
     print("          -gdb: run compiler under gdb")
     print("          --pp file: pass this option to the compiler")
     print("          -observation-log <file>: save packet output to <file>")
+    print("          --init <cmd>: Run <cmd> before the start of the test")
 
 
 def isError(p4filename):
@@ -183,9 +189,21 @@ def run_model(options, tmpdir, jsonfile):
     result = bmv2.checkOutputs()
     return result
 
+def run_init_commands(options):
+    if not options.initCommands:
+        return SUCCESS
+    for cmd in options.initCommands:
+        args = cmd.split()
+        result = run_timeout(options, args, timeout, None)
+        if result != SUCCESS:
+            return FAILURE
+    return SUCCESS
+
 def process_file(options, argv):
     assert isinstance(options, Options)
 
+    if (run_init_commands(options) != SUCCESS):
+        return FAILURE
     tmpdir = tempfile.mkdtemp(dir=".")
     basename = os.path.basename(options.p4filename)
     base, ext = os.path.splitext(basename)
@@ -276,6 +294,22 @@ def main(argv):
             else:
                 options.compilerOptions += argv[1].split();
                 argv = argv[1:]
+        elif argv[0] == "--switch-arg":
+            if len(argv) == 0:
+                reportError("Missing argument for --switch-arg option")
+                usage(options)
+                sys.exit(FAILURE)
+            else:
+                options.switchOptions += argv[1].split();
+                argv = argv[1:]
+        elif argv[0] == "--target-specific-switch-arg":
+            if len(argv) == 0:
+                reportError("Missing argument for --target-specific-switch-arg option")
+                usage(options)
+                sys.exit(FAILURE)
+            else:
+                options.switchTargetSpecificOptions += argv[1].split();
+                argv = argv[1:]
         elif argv[0][1] == 'D' or argv[0][1] == 'I' or argv[0][1] == 'T':
             options.compilerOptions.append(argv[0])
         elif argv[0] == "-gdb":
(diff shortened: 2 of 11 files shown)


omeranson:
I've added two tests: A negative test (compilation fails if an extern function is used without --emit-externs) and a positive test.

The positive test currently fails since BMv2 doesn't know how to execute the extern function. Once https://github.com/p4lang/behavioral-model/pull/605 is merged, then the test can be improved further. (And an end-to-end test can be written). This will also require updating the BMv2 backend testing infrastructure to allow passing parameters to simple_switch (i.e. the --load-modules option from the BMv2 PR).

backends/bmv2/simpleSwitch.cpp
```diff
@@ -36,6 +36,7 @@ namespace P4V1 {
 // need a class to generate simpleSwitch
 
 cstring jsonMetadataParameterName = "standard_metadata";
+cstring externFunctionPrefix = "__extern_";
```
antoninbas: I'm not sure why this prefix is needed at all. I believe we did remove the automatic prepending of this prefix when registering the function in your bmv2 PR. Did you actually experience a name conflict when using this?

omeranson: I added this prefix from the get-go. But I will remove it. On second thought, a collision like this would probably appear in the p4 level as well, not just after compilation.

omeranson:
@mbudiu-vmw My plan was to modify run-bmv2-test to accept additional command line arguments to pass directly to simple_switch (like the -a parameter is passed to the compiler). Then modify bmv2stf.py to accept these additional command line arguments. Lastly, I will have a dummy .cpp file that compiles to a .so file that can be loaded, and registers the expected externs.

What do you think?

omeranson:
I've added the tests and the infrastructure required for them (I can split it into two commits if needed). The test will fail until https://github.com/p4lang/behavioral-model/pull/605 is merged.

testdata/p4_16_samples/extern-funcs-bmv2.p4
```diff
@@ -0,0 +1,75 @@
+/*
```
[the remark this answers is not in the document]
omeranson: No. But I removed the line. It's also 2018 and not 2016 :)

backends/bmv2/CMakeLists.txt
```diff
@@ -119,10 +119,15 @@ set (XFAIL_TESTS
   testdata/p4_14_samples/issue604.p4
   # This test uses an incorrect model
   testdata/p4_16_samples/issue1205-bmv2.p4
+  # This test requires --emit-externs and extern support from BMv2
+  testdata/p4_16_samples/extern-funcs-bmv2.p4
   )
 
 if (HAVE_SIMPLE_SWITCH)
   p4c_add_tests("bmv2" ${BMV2_DRIVER} "${BMV2_TEST_SUITES}" "${XFAIL_TESTS}")
+
```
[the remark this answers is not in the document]
omeranson: I have not such plans at this time.

backends/bmv2/run-bmv2-test.py
```diff
@@ -110,9 +113,12 @@ def usage(options):
     print("          -v: verbose operation")
     print("          -f: replace reference outputs with newly generated ones")
     print("          -a option: pass this option to the compiler")
+    print("          --sa option: pass this option to the switch")
+    print("          --stsa option: pass this target-specific option to the switch")
     print("          -gdb: run compiler under gdb")
     print("          --pp file: pass this option to the compiler")
     print("          -observation-log <file>: save packet output to <file>")
+    print("          --make <target>: Make target before starting the test")
```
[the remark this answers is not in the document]
omeranson: I'll reword.

testdata/extern_modules/extern-funcs-bmv2.cpp
```diff
@@ -0,0 +1,6 @@
+#include <bm/bm_sim/data.h>
```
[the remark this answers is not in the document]
omeranson: Will do.

testdata/extern_modules/extern-funcs-bmv2.cpp
```diff
@@ -0,0 +1,6 @@
...
+#include <bm/bm_sim/extern.h>
+#include <bm/bm_sim/fields.h>
+
+void extern_func(bm::Field &, const bm::Data &) { /* Do nothing */ }
```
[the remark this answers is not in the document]
omeranson: Will do.

backends/bmv2/run-bmv2-test.py
```diff
@@ -172,6 +178,8 @@ def run_model(options, tmpdir, jsonfile):
         return SUCCESS
 
     bmv2 = RunBMV2(tmpdir, options, jsonfile)
+    bmv2.cmd_line_args = options.switchOptions
```
hanw: `options` is already passed in to RunBMV2 as an argument, there is no need for these two assignments outside the RunBMV2 function. Do the assignment inside RunBMV2?

backends/bmv2/run-bmv2-test.py
```diff
@@ -181,9 +189,21 @@ def run_model(options, tmpdir, jsonfile):
     result = bmv2.checkOutputs()
     return result
 
+def make_make_targets(options):
```
hanw: It's a lot of machinery to support one test case, why don't you put all these commands into a shell script and just invoke the shell script from cmake?

backends/bmv2/run-bmv2-test.py
```diff
@@ -51,9 +51,12 @@ def __init__(self):
         self.verbose = False
         self.replace = False            # replace previous outputs
         self.compilerOptions = []
+        self.switchOptions = []
```
hanw: Overall, I think it would be a lot cleaner if you could drive your test case with a single shell script, as opposed to adding three new arguments to run-bmv2-test.py to forward command options to bmv2 switch and to invoke make.

backends/bmv2/run-bmv2-test.py
```diff
@@ -265,6 +285,22 @@ def main(argv):
             else:
                 options.compilerOptions += argv[1].split();
                 argv = argv[1:]
+        elif argv[0] == "--sa":
```
hanw: If you use a separate script to drive your dynamic loading test case, then none of these command line arguments are needed.

backends/bmv2/run-bmv2-test.py
omeranson: Will do.

backends/bmv2/run-bmv2-test.py
omeranson: I thought it would be more reusable this way.

backends/bmv2/run-bmv2-test.py
omeranson: At a first glance, the shell script would only allow me to remove the --make option. I still need to pass arguments to simple_switch. Either I still add the arguments via `run-bmv2-test.py`, or I disable that part of the code and do it manually, or bypass `run-bmv2-test.py` altogether.

I think this way is more reusable, and there's less code duplication.

backends/bmv2/run-bmv2-test.py
omeranson: Please see my responses above.

hanw: Your test case does not use `--sa` option, how is it different from `--stsa`? What about changing it to `-s` for passing option to bmv2 switch? the `stsa` acrynym is difficult to understand and remember.

backends/bmv2/run-bmv2-test.py
hanw: Instead of extending run-bmv2-test.py to support invoking make, how about we extend the python script with an init script/step/action before executing the bmv2 switch. In your test case, the init script will call `make` and built a shared library. In other future test cases, the init script can call other linux utility tools. The run-bmv2-test.py script does not need to know the details of each init script and it certainly need not to know how to invoke `make`.

backends/bmv2/run-bmv2-test.py
omeranson: The syntax for invoking simple_switch is: ./simple_switch <general arguments> -- <target-specific arguments>. --sa is used to provide general arguments (Switch Arguments). --stsa is used for target-specific arguments (Switch Target Specific Argumenta). Even though --sa isn't used here, I included it for completeness.

omeranson: Sorry, I see github swallowed the syntax. I meant to say it's:
./simple_switch \<general arguments\> -- \<target-specific arguments\> .

testdata/p4_16_bmv_errors/no-externs-bmv2.p4
```diff
@@ -0,0 +1,75 @@
...
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+#include <v1model.p4>
+
+extern void extern_func(out bit<32> d, bit<32> s);
```
mbudiu-vmw: I think you should also document this function explaining what the external implementation does.
No one can guess without reading all the code, which is scattered among several files.
You should also document here where the function implementation is found.

omeranson: Sure. Will do (here and in the positive test)

omeranson:
Hi,

I've rebased on top of master.

Thanks.

backends/bmv2/run-bmv2-test.py
hanw: The `--sa` acronym is difficult to understand, can you spell it out please?

backends/bmv2/run-bmv2-test.py
```diff
@@ -276,6 +294,22 @@ def main(argv):
...
+            if len(argv) == 0:
+                reportError("Missing argument for --sa option")
+                usage(options)
+                sys.exit(FAILURE)
+            else:
+                options.switchOptions += argv[1].split();
+                argv = argv[1:]
+        elif argv[0] == "--stsa":
```
hanw: The `--stsa` acronym is difficult to understand, can you spell it out please?

hanw:
I am happy if you can make the command line options more readable.

omeranson:
@hanw Done

backends/bmv2/run-bmv2-test.py
omeranson: Done

PR #1294, opened 2018-05-23
Generate a deep-compare "equiv" function for IR nodes

ChrisDodd:
I talked about this awhile back -- this has the ir-generator generate a deep-compare method `equiv` for all IR classes.  It recursively compares all pointed at IR classes, so it is potentially expensive, but when it's needed, it's needed.

### DIFF
--- a/ir/node.h
+++ b/ir/node.h
@@ -109,7 +109,11 @@ class Node : public virtual INode {
     void toJSON(JSONGenerator &json) const override;
     void sourceInfoToJSON(JSONGenerator &json) const;
     Util::JsonObject* sourceInfoJsonObj() const;
+    /* operator== does a 'shallow' comparison, comparing two Node subclass objects for equality,
+     * and comparing pointers in the Node directly for equality */
     virtual bool operator==(const Node &a) const { return typeid(*this) == typeid(a); }
+    /* 'equiv' does a deep-equals comparison, comparing all non-pointer fields and recursing
+     * though all Node subclass pointers to compare them with 'equiv' as well. */
     virtual bool equiv(const Node &a) const { return typeid(*this) == typeid(a); }
 #define DEFINE_OPEQ_FUNC(CLASS, BASE) \
     virtual bool operator==(const CLASS &) const { return false; }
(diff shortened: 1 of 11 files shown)


ir/node.h
```diff
@@ -110,6 +110,7 @@ class Node : public virtual INode {
     void sourceInfoToJSON(JSONGenerator &json) const;
     Util::JsonObject* sourceInfoJsonObj() const;
     virtual bool operator==(const Node &a) const { return typeid(*this) == typeid(a); }
+    virtual bool equiv(const Node &a) const { return typeid(*this) == typeid(a); }
```
cole-barefoot: Can you add a comment, either here or wherever you think is appropriate, documenting `equiv`, including
 - the difference between it and `operator==`, and
 - the relationship between `equiv` and the visitors?

For example, it looks like `equiv` checks the equality even for fields that aren't visited.

tools/ir-generator/methods.cpp
```diff
@@ -56,6 +56,45 @@ const ordered_map<cstring, IrMethod::info_t> IrMethod::Generate = {
...
+        buf << cl->indent << cl->indent << "if (typeid(*this) != typeid(a_)) return false;\n";
+        if (auto parent = cl->getParent()) {
+            while (parent && parent->kind == NodeKind::Abstract)
+                parent = parent->getParent();
+            if (parent && parent->name != "Node")
+                buf << cl->indent << cl->indent << "if (!" << parent->name
+                    << "::equiv(a_)) return false;\n"; }
+        bool first = true;
+        for (auto f : *cl->getFields()) {
+            if (*f->type == NamedType::SourceInfo) continue;  // FIXME -- deal with SourcInfo
```
mbudiu-vmw: I don't know if there is anything to fix here; I suspect that you really want the equivalence to ignore the position.

ir/node.h
mbudiu-vmw: I think we should call this `syntacticallyEquivalent` to make it clear that this is not meant to say that the two trees evaluate to the same value or some other notion of equivalence.

PR #1295, opened 2018-05-23
Add sample command lines

jafingerhut:
I know there are many, many command line options for these programs, but it seems useful to have quick examples for the most basic command line options, and how to auto-translate P4_14 to P4_16 source code.

A sample of how the modified README looks in Github can be found here: https://github.com/jafingerhut/p4c/tree/add-sample-command-lines


cc10512:
I would really like to encourage people to use the p4c driver rather than directly invoking the backend.
You can get the same result of invoking `p4c-bm-ss` by invoking `p4c --target bmv2`.

jafingerhut:
@cc10512 I can update the example commands to use the `--target bmv2` option.

Do you want the sample commands to also specify `--arch <something>`, too?  If so, should the example have `--arch v1model`, or something else?

cc10512:
`--arch v1model` is the default, but then of course, we will have `--arch psa` soon :)
Making that explicit would be very nice. Thank you for doing this!

jafingerhut:
@cc10512 Please take a look at the latest version I've pushed up to now, and see if there is anything else you would like to see changed.

README.md
```diff
@@ -22,6 +22,57 @@ The code contains four sample backends:
...
   for now it only supports generating graphs of top-level control flows.
 
+Sample command lines:
+
+Compile P4_16 or P4_14 source code.  If your program successfully
+compiles, the command will create files with the same base name as the
+P4 program you supplied, and the following suffixes instead of the
+`.p4`:
+
++ a file with suffix `.p4i`, which is the output from running the
```
hanw: .p4i was the suffix I came up with when designing the p4c driver, but it is really just the output from c preprocessor, should we call it *.i or *.ii to follow the naming convention of C? No big deal, just feel like a waste of good suffix on something not very important.

jafingerhut: I personally don't mind the "p4i" suffix.  I didn't realize there was a conventional *.i or *.ii suffix for C after preprocessing.

jafingerhut:
@cc10512 Added `p4c --help` and `p4c --target-help` commands with a sentence about what they do.

PR #1298, opened 2018-05-24
initial support for psa architecture in bmv2 backend

hanw:
This PR restructured the bmv2 backend to add support for the psa architecture in addition to the existing v1model architecture. I am trying to maximize the code reuse between the two backends by factoring out a base Backend class which is extended by both PortableSwitch and SimpleSwitch backend classes. Most of the conversion code for parser, control, action, deparser are shared between the two backends. Extern conversion is target-specific, we have introduced the a mechanism to register target-specific extern converter following other similar mechanisms in p4c frontend and behavior model. 

cmake will generate two separate binaries, one for each target. The simple switch remains as p4c-bm-ss. The portable switch binary is called p4c-bm-psa. 

I have added a dozen of psa test programs.

Caveat: this PR does not fully support all psa features yet. I will submit a few more PRs as I finish them. But as of now, the change is already quite a lot, and I would appreciate other maintainers to take some time and give me some feedback. 

We must give credit to @rakeshsharma712 for getting the initial implementation done for this PR.


ChrisDodd:
Such a large change will take awhile to review -- I'll try to get time to look over it this weekend.

hanw:
Thanks! Take your time.

jnfoster:
Nicely done @hanw and @rakeshsharma712! 🎉

backends/bmv2/CMakeLists.txt
```diff
@@ -16,74 +16,84 @@
...
-  simpleSwitch.cpp
+set (BMV2_SIMPLE_SWITCH_SRCS
+    simple_switch/main.cpp
+    simple_switch/simpleSwitch.cpp
+    simple_switch/simpleSwitch.h
+    )
+add_cpplint_files (${CMAKE_CURRENT_SOURCE_DIR} "${BMV2_SIMPLE_SWITCH_SRCS}")
+
+set (BMV2_PORTABLE_SWITCH_SRCS
+    portable_switch/main.cpp
```
cc10512: Same suggestion that @antoninbas had for bmv2, shall we call it psa here as well?

backends/bmv2/CMakeLists.txt
```diff
@@ -122,7 +149,8 @@ set (XFAIL_TESTS
   )
 
 if (HAVE_SIMPLE_SWITCH)
-  p4c_add_tests("bmv2" ${BMV2_DRIVER} "${BMV2_TEST_SUITES}" "${XFAIL_TESTS}")
+  p4c_add_tests("bmv2" ${BMV2_DRIVER} "${BMV2_V1MODEL_TEST_SUITES}" "${XFAIL_TESTS}")
+  #p4c_add_tests("bmv2" ${BMV2_DRIVER} "${BMV2_PSA_TEST_SUITES}" "${XFAIL_TESTS}" ${testExtraFlags})
```
cc10512: I think we need to have a HAVE_PSA_SWITCH macro defined which finds the `portable_switch` binary (or  `psa_switch` if renamed)

backends/bmv2/common/action.cpp
```diff
@@ -15,11 +15,34 @@ limitations under the License.
 */
 
 #include "action.h"
+#include "extern.h"
 
 namespace BMV2 {
 
-void
-ConvertActions::convertActionBody(const IR::Vector<IR::StatOrDecl>* body, Util::JsonArray* result) {
+cstring ActionConverter::jsonAssignment(const IR::Type* type, bool inParser) {
```
cc10512: A comment on this function will help.

backends/bmv2/common/helpers.h
```diff
@@ -14,15 +14,19 @@ See the License for the specific language governing permissions and
...
 #include "ir/ir.h"
 #include "lib/cstring.h"
 #include "lib/json.h"
 #include "lib/ordered_map.h"
-#include "analyzer.h"
+#include "controlFlowGraph.h"
 #include "frontends/common/model.h"
+#include "programStructure.h"
+#include "expression.h"
+#include "JsonObjects.h"
```
cc10512: alphabetical order, please

backends/bmv2/common/helpers.h
```diff
@@ -62,18 +59,44 @@ class V1ModelProperties {
     static const cstring validField;
 };
 
-using ErrorValue = unsigned int;
-using ErrorCodesMap = ordered_map<const IR::IDeclaration *, ErrorValue>;
+// XXX(hanw): This convenience class stores pointers to the data structures
+// that are commonly used during the program translation. Due to the limitation
+// of current IR structure, these data structure are only refreshed by the
+// evaluator pass. In the long term, integrating these data structures as part
+// of the IR tree would simplify this kind of bookkeeping effort.
```
cc10512: +1

mbudiu-vmw: This may not be possible, since some of these data structures may have cycles, and the IR prohibits cycles.

mbudiu-vmw: I also think that the IR is complicated enough, we should not carry within the IR stuff that we only need occasionally.

backends/bmv2/common/midend.cpp
```diff
@@ -105,8 +105,12 @@ MidEnd::MidEnd(CompilerOptions& options) {
         new P4::LocalCopyPropagation(&refMap, &typeMap),
         new P4::ConstantFolding(&refMap, &typeMap),
         new P4::MoveDeclarations(),
-        new P4::ValidateTableProperties({ "implementation", "size", "counters",
-                                          "meters", "support_timeout" }),
+        (options.arch == "v1model") ?
```
mbudiu-vmw: can you refactor this into subclasses as well?
common should not need to know about various subclasses.
You may want to have different mid-ends for the two back-ends, I think that's reasonable.

backends/bmv2/portable_switch/portableSwitch.cpp
```diff
@@ -0,0 +1,823 @@
...
+    createHeaders(ctxt);
+    createExterns();
+    createParsers(ctxt);
+    createActions(ctxt);
+    createControls(ctxt);
+    createDeparsers(ctxt);
+    createGlobals();
+}
+
+void PsaProgramStructure::createStructLike(ConversionContext* ctxt, const IR::Type_StructLike* st) {
```
mbudiu-vmw: Why is this code PSA-specific?

mbudiu-vmw: Should you have a base class for ProgramStructure as well?

backends/bmv2/common/extern.h
```diff
@@ -0,0 +1,154 @@
...
+            const IR::Declaration* c, const IR::ExternBlock* eb) override;      \
+        Util::IJson* convertExternObject(ConversionContext* ctxt,               \
+            const P4::ExternMethod* em, const IR::MethodCallExpression* mc,     \
+            const IR::StatOrDecl* s) override; };
+
+#define EXTERN_CONVERTER_SINGLETON(extern_name, ...)                           \
+    ExternConverter_##extern_name##__VA_ARGS__                                 \
+       ExternConverter_##extern_name##__VA_ARGS__::singleton;
+
+#define CONVERT_EXTERN_INSTANCE(extern_name, ...)                                     \
```
mbudiu-vmw: I don't like macros much. In particular, these macros hardwire some parameter names. When you read the use of the macro you have no idea what these things are or where they come from (e.g., eb, em, mc).
If you use macros maybe you can at least expose all the parameters in the macro, and perhaps give longer names to the parameters.

backends/bmv2/portable_switch/portableSwitch.cpp
mbudiu-vmw: Really, a lot of this code looks independent of psa.

backends/bmv2/portable_switch/portableSwitch.cpp
```diff
@@ -0,0 +1,823 @@
...
+    auto primitive = mkPrimitive("Checksum");
+    return primitive;
+}
+
+CONVERT_EXTERN_OBJECT(InternetChecksum) {
+    auto primitive = mkPrimitive("InternetChecksum");
+    return primitive;
+}
+
+CONVERT_EXTERN_OBJECT(Counter) {
```
mbudiu-vmw: This is what I am talking about: it is not clear at all what is available in the body of this function.

backends/bmv2/simple_switch/simpleSwitch.cpp
```diff
@@ -0,0 +1,970 @@
...
+    if (id >= 0) {
+        auto cst = new IR::Constant(id);
+        ctxt->typeMap->setType(cst, IR::Type_Bits::get(32));
+        auto jcst = ctxt->conv->convert(cst);
+        parameters->append(jcst);
+    }
+    return primitive;
+}
+
+CONVERT_EXTERN_FUNCTION(clone3) {
```
mbudiu-vmw: I expect you may be able to reuse some of this code for psa as well. Maybe it will need to move in a base class.

testdata/p4_16_samples/psa-action-profile1.p4
```diff
@@ -0,0 +1,100 @@
+#include <core.p4>
```
mbudiu-vmw: Maybe we should name these files with -psa.p4 as a suffix.

mbudiu-vmw:
This pass orders arguments in the order of parameters, but if you have an optional parameter that has no corresponding supplied argument somewhere in the middle of the parameter list you cannot rewrite the call in this way.

mbudiu-vmw:
There are no declarations with optional parameters in v1model, so this will work for v1model.

hanw:
There is no optional arguments in psa either, unless user defines a control block with optional arguments.

testdata/p4_16_samples/psa-action-profile1.p4
hanw: I am using psa-*-bmv2.p4 for psa program that are ready for bmv2 test.
and psa-*.p4 for psa program that are only passing frontend/midend.

backends/bmv2/common/extern.h
hanw: I will think of a better way to write this macro, maybe in the next PR.

backends/bmv2/common/extern.cpp
```diff
@@ -0,0 +1,210 @@
...
+                                       const IR::MethodCallExpression* ,
+                                       const IR::StatOrDecl* ) {
+    ::error("Unknown extern function %1%", ef->method->name);
+    return nullptr;
+}
+
+void
+ExternConverter::modelError(const char* format, const IR::Node* node) const {
+    ::error(format, node);
+    ::error("Are you using an up-to-date v1model.p4?");
```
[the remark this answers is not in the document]
hanw: Yes, more work is needed in ExternConverter to make it a true base class. So far, a lot of the implementation is just reusing what's done in v1model.

backends/bmv2/common/control.cpp
```diff
@@ -501,6 +501,15 @@ ControlConverter::convertTable(const CFG::TableNode* node,
         result->emplace("direct_meters", Util::JsonValue::null);
     }
 
+    auto psa_counter = table->properties->getProperty("psa_direct_counter");
```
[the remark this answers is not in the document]
hanw: Similar here, most of the implementation is reusing what's available in v1model. I will try to implement a base class in next PR.

backends/bmv2/CMakeLists.txt
```diff
@@ -94,13 +104,30 @@ add_dependencies(p4c_driver linkbmv2)
...
+set (BMV2_V1MODEL_TEST_SUITES
   "${P4C_SOURCE_DIR}/testdata/p4_16_bmv_errors/*-bmv2.p4"
   "${P4C_SOURCE_DIR}/testdata/p4_14_samples/*.p4"
   "${P4C_SOURCE_DIR}/testdata/p4_14_samples/switch_*/switch.p4"
+  ${v1tests}
+  )
+
+set (PSA_INCLUDE_PATTERNS "include.*psa.p4")
+set (PSA_EXCLUDE_PATTERNS "include.*v1model.p4")
+set (P4TESTS_FOR_PSA "${P4C_SOURCE_DIR}/testdata/p4_16_samples/*-bmv2.p4")
```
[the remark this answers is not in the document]
hanw: Not all psa programs are ready to run on bmv2 yet, some of them needs more feature to be implemented in the backend.

mbudiu-vmw:
There aren't optional arguments at this point, but one point of adding them to the spec was that apparently PSA could use them if they were available.

hanw:
If it is okay with you, I would like to merge in this PR, so I can start the next one. There is still more work to be done to make the common completely target independent, but it will be easier to work on a smaller PR. All existing tests are passing for this PR.

mbudiu-vmw:
Let's merge it, but I really hope you will get rid of these macros in the next one.

hanw:
Thanks. I will find a way to make the macros better, behavioral-model has a nice mechanism to register new externs handlers, which might apply here as well.

PR #1305, opened 2018-05-29
Corrected syntax and inmport errors in p4include/psa.p4

SantiagoBautista:
+ As `psa.p4` uses the `packet_in` type without defining it, it should include `core.p4` .
+ There was a syntax error in the declaration of  `enum MeterColor_t` : as this enum declaration is not a typedef declaration, it cannot end with a semicolon.
+ Some non-executable files where in a mode allowing execution in linux; changed the mode by taking execution rights away


hanw:
Before we can accept this contribution, can you confirm you or your institution is a member of P4.org?

SantiagoBautista:
Created pull request #1311 with the desired changes.
I am currently working with the Cornell University, which is a member of P4.org

jafingerhut:
@SantiagoBautista I would recommend creating a PR in the p4-spec repository, which currently contains the latest version of the psa.p4 #include file.

While you are at it, I guess perhaps we should also add things like this around the body of psa.p4, as core.p4 and v1model.p4 include files currently have:
```
#ifndef __PSA_P4__
#define __PSA_P4__

<body goes here>

#endif   // __PSA_P4__
```

PR #1311, opened 2018-05-30
Fix syntax error and add missing include in psa.p4

SantiagoBautista:
This pull request contains 4 minor corrections to the psa.p4 file:
+ Updated file psa.p4 so it is up-to-date with p4-spec/p4-16/psa/psa.p4
+ There was a missing include in file psa.p4 : added it
+ There was an unexpected semicolon at the end of an enum declaration :  deleted it
+ Some non-executable files were in a mode allowing execution: changed the rights to correct it

This pull request does the changes asked for in PR1305 (I don't know how to modify a PR once created, so I created a new one).
I am currently working for the Cornell University, which is a member of p4.org


jafingerhut:
I haven't done a diff against the latest version of psa.p4 in the p4lang/p4-spec repository, but as long as the changes from that latest version in there are only the ones you mention, looks reasonable to me.  It would be great if you could open a PR for the spec to make its psa.p4 match this one.

SantiagoBautista:
I opened a PR in p4-spec (p4-spec/#632) as suggested :) .
I wrapped the psa.p4 body in `#ifndef ... #endif` as suggested in the comment on #1311 .

SantiagoBautista:
The pr#632 in p4-spec was merged, so now there is no difference between the psa.p4 file in p4-spec and the one in this commit.

SantiagoBautista:
I will close this pr in favor of #1324 , that is more up-to-date and simple/clear.

PR #1314, opened 2018-05-31
A framework for P4Info generation for arbitrary architectures

antoninbas:
More architectures are being developped and there is a desire to be able
to use P4Runtime with them. P4Runtime supports standard architectures
(v1model & PSA) natively and support architecture-specific extensions
(table properties & externs). In order to enable using P4Runtime with
vendor-specific architectures, we define a framework that can be used to
implement P4Info generation for these architectures more easily. v1model
& PSA also use this framework, but support for these standard
architectures is supported natively by p4c, while vendor-specific
architectures must be supported in the vendor's backend.

The goal of the framework is to reduce code redundancy. A
P4RuntimeArchHandlerIface implementation can be registered for each
architecture. The P4RuntimeSerializer singleton object takes care of all
the common logic, and calls the overridden P4RuntimeArchHandlerIface
when needed for architecture-specific processing.

There are still some minor v1model references in P4RuntimeSerializer
that will be solved in a future commit. These are minor and are
dependent on some fixes in p4info.proto.

P4Info generation for PSA is not supported yet (it will be in a future
commit) but most of the v1model code should be re-usable.

### DIFF
--- a/control-plane/p4RuntimeArchHandler.h
+++ b/control-plane/p4RuntimeArchHandler.h
@@ -1,0 +1,384 @@
+/*
+Copyright 2013-present Barefoot Networks, Inc.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+#ifndef CONTROL_PLANE_P4RUNTIMEARCHHANDLER_H_
+#define CONTROL_PLANE_P4RUNTIMEARCHHANDLER_H_
+
+#include <boost/optional.hpp>
+
+#include <set>
+
+#include "p4/config/v1/p4info.pb.h"
+
+#include "frontends/p4/externInstance.h"
+#include "frontends/p4/methodInstance.h"
+#include "ir/ir.h"
+#include "lib/ordered_set.h"
+
+namespace P4 {
+
+/** \addtogroup control_plane
+ *  @{
+ */
+namespace ControlPlaneAPI {
+
+using p4rt_id_t = uint32_t;
+
+/// Base class for the different types introduced by the P4 architecture. The
+/// class includes static factory methods for all the built-in P4 types common
+/// to all architectures. When defining the P4Info serialization logic for a new
+/// architecture, the first step is to inherit from this class and add factory
+/// methods for architecture-specific types. The class is used to separate all
+/// symbols by type in the symbol table which is responsible for control-plane
+/// id assignment.
+class P4RuntimeSymbolType {
+ public:
+    virtual ~P4RuntimeSymbolType() { }
+
+    /// An explicit conversion operator that is useful when building the P4Info
+    /// message. It returns the id prefix for that type, as defined in the
+    /// p4info.proto file.
+    explicit operator p4rt_id_t() const { return id; }
+
+    static P4RuntimeSymbolType ACTION() {
+        return P4RuntimeSymbolType(::p4::config::v1::P4Ids::ACTION);
+    }
+    static P4RuntimeSymbolType TABLE() {
+        return P4RuntimeSymbolType(::p4::config::v1::P4Ids::TABLE);
+    }
+    static P4RuntimeSymbolType VALUE_SET() {
+        return P4RuntimeSymbolType(::p4::config::v1::P4Ids::VALUE_SET);
+    }
+    static P4RuntimeSymbolType CONTROLLER_HEADER() {
+        return P4RuntimeSymbolType(::p4::config::v1::P4Ids::CONTROLLER_HEADER);
+    }
+
+    bool operator==(const P4RuntimeSymbolType& other) const {
+        return id == other.id;
+    }
+
+    bool operator!=(const P4RuntimeSymbolType& other) const {
+        return !(*this == other);
+    }
+
+    bool operator<(const P4RuntimeSymbolType& other) const {
+        return id < other.id;
+    }
+
+ protected:
+    static P4RuntimeSymbolType make(p4rt_id_t id) {
+        return P4RuntimeSymbolType(id);
+    }
+
+ private:
+    // even if the constructor is protected, the static functions in the derived
+    // classes cannot access it, which is why we use the make factory function
+    constexpr P4RuntimeSymbolType(p4rt_id_t id) noexcept
+        : id(id) { }
+
+    /// The 8-bit id prefix for that type, as per the p4info.proto file.
+    p4rt_id_t id;
+};
+
+/// A table which tracks the symbols which are visible to P4Runtime and their
+/// ids.
+class P4RuntimeSymbolTableIface {
+ public:
+    virtual ~P4RuntimeSymbolTableIface() { }
+    /// Add a @type symbol, extracting the name and id from @declaration.
+    virtual void add(P4RuntimeSymbolType type, const IR::IDeclaration* declaration) = 0;
+    /// Add a @type symbol with @name and possibly an explicit P4 '@id'.
+    virtual void add(P4RuntimeSymbolType type, cstring name,
+                     boost::optional<p4rt_id_t> id = boost::none) = 0;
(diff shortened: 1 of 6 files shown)


control-plane/p4RuntimeArchHandler.h
```diff
@@ -0,0 +1,384 @@
...
+/// The interface for defining the P4Info serialization logic for a specific P4
+/// architecture. The goal is to reduce code duplication between
+/// architectures. The @ref P4RuntimeSerializer will call these methods when
+/// generating the P4Info message to handle architecture-specific parts. @ref
+/// P4RuntimeSerializer generates the P4Info in two passes: first it collects
+/// all the control-plane visible symbols from the program into the symbol
+/// table, then it builds the P4Info message by adding each collected entity to
+/// the Protobuf message. The collect* methods are called in the first pass, the
+/// add* methods are called in the second pass.
+class P4RuntimeArchHandlerIface {
```
cc10512: I would have implemented this as a pass manager with two passes, just to take advantage of the visiting functions, rather than rolling out another visiting infra.

control-plane/p4RuntimeArchHandler.h
```diff
@@ -0,0 +1,384 @@
...
+
+/// @return true if the extern instance assigned to property @propertyName for
+/// the @table was constructed in-place or outside of teh @table declaration.
+bool isExternPropertyConstructedInPlace(const IR::P4Table* table,
+                                        const cstring& propertyName);
+
+/// Visit evaluated blocks under the provided top-level block. Guarantees that
+/// each block is visited only once, even if multiple paths to reach it exist.
+template <typename Func>
+void forAllEvaluatedBlocks(const IR::ToplevelBlock* aToplevelBlock, Func function) {
```
cc10512: Isn't this just what a FlowVisitor is supposed to achieve?

control-plane/p4RuntimeArchStandard.cpp
```diff
@@ -0,0 +1,715 @@
...
+namespace p4configv1 = ::p4::config::v1;
+
+namespace P4 {
+
+/** \addtogroup control_plane
+ *  @{
+ */
+namespace ControlPlaneAPI {
+
+namespace Standard {
```
cc10512: Interesting choice to call this namespace "Standard". I expect that indeed these will be part of the standard library of externs. However, you also chose to have v1model part of this file, and in fact part of the Standard namespace. Why this choice?

control-plane/p4RuntimeArchStandard.cpp
```diff
@@ -0,0 +1,715 @@
...
+                   externBlock->type->name == P4V1::V1Model::instance.action_selector.name) {
+            symbols->add(SymbolType::ACTION_PROFILE(), decl);
+        } else if (externBlock->type->name == P4V1::V1Model::instance.registers.name) {
+            symbols->add(SymbolType::REGISTER(), decl);
+        }
+    }
+
+    void collectExternFunction(P4RuntimeSymbolTableIface* symbols,
+                               const P4::ExternFunction* externFunction) override {
+        auto digest = getDigestCall(externFunction, refMap, typeMap, nullptr);
```
cc10512: How about clone, clone3, resubmit, recirculate? these are all v1model extern functions. No support in P4Runtime yet?

cc10512: Never mind, they are not controllable from the control plane.

control-plane/p4RuntimeSerializer.cpp
```diff
@@ -1049,60 +788,41 @@ class P4RuntimeAnalyzer {
...
-            auto id = symbols.getId(P4RuntimeSymbolType::ACTION, defaultAction->name);
+            auto id = symbols.getId(P4RuntimeSymbolType::ACTION(), defaultAction->name);
             table->set_const_default_action_id(id);
 
-            // XXX(seth): Generally the parameters of const default actions are
-            // bound at compile time, so this is a good default, but we need to
-            // add support for the cases where they aren't.
+            // TODO(antonin): Generally the parameters of const default actions
+            // are bound at compile time, so this is a good default, but we need
+            // to add support for the cases where they aren't.
```
cc10512: +1

control-plane/p4RuntimeSerializer.cpp
```diff
@@ -2174,5 +1393,44 @@ void serializeP4RuntimeIfRequired(const IR::P4Program* program,
...
+
+cstring
+P4RuntimeSerializer::resolveArch(const CompilerOptions& options) {
+    if (auto arch = getenv("P4C_DEFAULT_ARCH")) {
+        return cstring(arch);
+    } else if (options.arch != "") {
+        return options.arch;
+    } else {
+        return "v1model";
+    }
```
cc10512: Isn't this already resolved by the backend you use?

control-plane/p4RuntimeArchHandler.h
antoninbas: This was already heavily used in the original code so I'm going to trust that there was a reason for Seth to add this.
Do you mean `ControlFlowVisitor`? We use `forAllEvaluatedBlocks` for parsers as well, and just like I mentioned above, I'm not sure how convenient visitors are with evaluated blocks.

control-plane/p4RuntimeArchStandard.cpp
antoninbas: For me v1model + PSA constitute the standard architectures, and I was going to put all the code specific to these architectures under `Standard`. I don't have support for PSA but I expect I will try to re-use as much code as possible.

control-plane/p4RuntimeArchStandard.cpp
antoninbas: Clone is configured at runtime and there is support in P4Runtime, but no information is required in P4Info.

control-plane/p4RuntimeSerializer.cpp
antoninbas: A backend can support multiple architectures. For example ideally the bmv2 backend would be generic and support all architectures.

mbudiu-vmw: Unfortunately this code runs before the IR reaches a back-end.

control-plane/p4RuntimeArchStandard.cpp
```diff
@@ -0,0 +1,715 @@
...
+// whose declarator-id is not qualified shall be declared in the nearest
+// enclosing namespace of the template, or, if the namespace is inline (7.3.1),
+// any namespace from its enclosing namespace set. Such a declaration may also
+// be a definition. If the declaration is not a definition, the specialization
+// may be defined later (7.3.1.2).
+//
+// gcc reports an error when trying so specialize CounterlikeTraits<> for
+// Standard::CounterExtern & Standard::MeterExtern outside of the Helpers
+// namespace, even when qualifying CounterlikeTraits<> with Helpers::. It seems
+// to be related to thsi bug:
```
mbudiu-vmw: typo in comment.

control-plane/p4RuntimeArchStandard.cpp
```diff
@@ -0,0 +1,715 @@
...
+        return P4RuntimeSymbolType::make(p4configv1::P4Ids::METER);
+    }
+    static P4RuntimeSymbolType REGISTER() {
+        return P4RuntimeSymbolType::make(p4configv1::P4Ids::REGISTER);
+    }
+};
+
+/// The information about a digest call which is needed to serialize it.
+struct Digest {
+    const cstring name;       // The fully qualified external name of the digest
```
mbudiu-vmw: These could probably be /// comments as well.

control-plane/p4RuntimeArchStandard.cpp
```diff
@@ -0,0 +1,715 @@
...
+            if (isExternPropertyConstructedInPlace(tableDeclaration, propertyName))
+                addActionProfile(symbols, p4info, *implementation);
+        }
+
+        if (directCounter) {
+            auto id = symbols.getId(SymbolType::DIRECT_COUNTER(),
+                                    directCounter->name);
+            table->add_direct_resource_ids(id);
+            // no risk to add twice because direct counters cannot be shared
+            // auto propertyName = P4V1::V1Model::instance.tableAttributes.counters.name;
```
mbudiu-vmw: do you want to deleted the commented-out code?

control-plane/p4RuntimeSerializer.cpp
hanw: currently each bmv2 backend is a separate binary.

antoninbas:
@mbudiu-vmw sorry for merging so fast
I'll open a new PR that addresses your review comments early next week.

Issue #1310, opened 2018-05-30
Assignment from from tuple to header fails


ygottlieb-perspectalabs-com:
In a simple test (attached), I try to assign a header from a tuple, but I get the following error:

/home/ygottlieb/examples/assign.p4(34): error: AssignmentStatement: Number of fields 1 in initializer different than number of fields in structure 2: Tuple(1) to header test
    headers.test = { 1 };
                 ^
/home/ygottlieb/examples/assign.p4(34)
    headers.test = { 1 };
                   ^^^^^
/home/ygottlieb/examples/assign.p4(4)
header test {
       ^^^^
terminate called after throwing an instance of 'Util::CompilerBug'
  what():  In file: /home/ygottlieb/p4/p4c/control-plane/p4RuntimeSerializer.cpp:2044
Compiler Bug: Failed to transform the program into a P4Runtime-compatible form

Based on git bisect, the problem first appears in commit 9d836b33220571b1424c12e6864ff9cf275631e8, possibly due to control-plane/p4RuntimeSerializer.cpp calling P4::SynthesizeValidField.

/* -*- Mode:C; c-file-style:"gnu"; indent-tabs-mode:nil; -*- */
#include <v1model.p4>

header test {
  bit<1> test;
}

struct headers_t {
  test test;
};

struct metadata_t {
};

parser EmptyParser(packet_in b, out headers_t headers, inout metadata_t meta,
                   inout standard_metadata_t standard_metadata)
{
  state start {
    /* No explicit reject */
    /* transition reject; */
    transition accept;
  }
}

control EmptyVerifyChecksum(inout headers_t hdr, inout metadata_t meta)
{
  apply {}
}

control EmptyIngress(inout headers_t headers, inout metadata_t meta,
                     inout standard_metadata_t standard_metadata)
{
  apply {
    headers.test = { 1 };
  }
}

control EmptyEgress(inout headers_t hdr,
                    inout metadata_t meta,
                    inout standard_metadata_t standard_metadata)
{
  apply {}
}

control EmptyComputeChecksum(inout headers_t hdr,
                             inout metadata_t meta)
{
  apply {}
}

control EmptyDeparser(packet_out b, in headers_t hdr)
{
  apply {}
}

V1Switch(EmptyParser(),
         EmptyVerifyChecksum(),
         EmptyIngress(),
         EmptyEgress(),
         EmptyComputeChecksum(),
         EmptyDeparser()) main;

mbudiu-vmw:
[...]

PR #1317, opened 2018-06-01
Fix for issue #1310: remove SynthesizeValidField pass.

mbudiu-vmw:
Also added another cstring.rename() method (turns out it's not needed, but may prove useful).
The test suite does not really test this feature, because it does not pass the suitable flags, so that's why there is no attached test program.


test/gtest/cstring.cpp
```diff
@@ -105,4 +105,12 @@ TEST(cstring, substr) {
     ASSERT_TRUE(c.exceptLast(2) == "simple"); // NOLINT
 }
 
+TEST(cstring, replace) {
+    cstring c = "Original";
+    ASSERT_TRUE(c.replace("in", "out") == "Origoutal");  // NOLINT
```
antoninbas: Should probably be `EXPECT_EQ`

mbudiu-vmw: I am not sure what equality operator is called by EXPECT_EQ, that's why I was using this version.
That's why NOLINT is there too. But if you vouch for the right operator then I can change this.

antoninbas: I'm pretty sure it is equivalent to writing `==` (it will call the `==` operator for cstring). You can also add a negative test with `EXPECT_NE` to make sure it works as expected :)

PR #1319, opened 2018-06-01
Support typedef in P4Info TypeSpecConverter pass

antoninbas:
Typedef's are eliminated in the midend, not the frontend, but the P4Info
generation happens after the frontend, so we have to support the case
where a type name can be a typdef.


antoninbas:
@hanw done

PR #1320, opened 2018-06-02
Fix segfault when compiling with O3

antoninbas:
The issue seems to occur when compiling a P4_14 program which has a
direct meter with a pre_color property. It seems to be fixed by only
invoking the ExpressionConverter on pre_color if the field is not null.


antoninbas:
This happens for example with p4_samples/meter1.p4
I don't know why it only happens during our Barefoot-specific backend Travis regressions and not during the p4c Travis regressions, since they are both compiling with O3. Maybe a different compiler version.
I'm also unsure whether this is actually fixing the issue or just hiding it. Should it actually be okay to call `convert` with a nullptr here?

PR #1323, opened 2018-06-05
std::set, std::map, ordered_set now allowed in IR

evanches:
A JSON loader/JSON generator template class was required for std::set, std::map, and ordered_set in order to include it as part of the IR.

### DIFF
--- a/ir/json_generator.h
+++ b/ir/json_generator.h
@@ -25,6 +25,8 @@ limitations under the License.
 #include "lib/cstring.h"
 #include "lib/indent.h"
 #include "lib/match.h"
+#include "lib/ordered_map.h"
+#include "lib/ordered_set.h"
 #include "lib/safe_vector.h"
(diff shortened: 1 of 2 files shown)


ir/json_generator.h
```diff
@@ -98,6 +100,52 @@ class JSONGenerator {
...
+            out << ++indent;
+            generate(*it);
+            for (it++; it != v.end(); ++it) {
+                out << "," << std::endl << indent;
+                generate(*it);
+            }
+            out << std::endl << --indent;
+        }
+        out << "]";
+    }
```
ChrisDodd: There's probably a way of generating these 3 identical functions as a single template function with an appropriate `std::enabble_if`, but I don't know what it is.

ChrisDodd: :+1: :+1: :bangbang:

PR #1326, opened 2018-06-07
Add a userspace test target to the ebpf compiler backend.

fruffy:
This request adds a new target to the EBPF compiler backend. This target, which can be invoked using the --target test flag, runs only in userspace and does not depend on the BCC framework or kernel includes. It is intended to emulate the forwarding and filtering behaviour of kernel EBPF for simplified testing purposes. To mimic the shared tables of ebpf, a new registry has been added. It depends on the [uthash](https://troydhanson.github.io/uthash/) hashmap library, which is licensed under the [revised BSD license](https://en.wikipedia.org/wiki/BSD_licenses#3-clause_license_(%22BSD_License_2.0%22,_%22Revised_BSD_License%22,_%22New_BSD_License%22,_or_%22Modified_BSD_License%22)).

This request also makes substantial changes to the structure of the target.cpp file. Instead of generating static code a new header file has been generated for each target. These header files can be found in `ebpfinclude/` and contain most of the important structures necessary for compilation for each target. Target `bcc` has not yet been changed to the new style. 

The change to target.cpp will break the XDP compiler backend, which depends on the kernel code generator. However, the necessary changes to the extension are small.

The code generated by target `kernel` and `test` compiles but has not been integrated into the overall test library yet.


backends/ebpf/bpfinclude/ebpf_registry.c
```diff
@@ -0,0 +1,77 @@
...
+    struct map_register *tmp_reg;
+    HASH_FIND(hh, maps, name, VAR_SIZE, tmp_reg);
+    if (tmp_reg==NULL)
+        return NULL;
+    return tmp_reg->map;
+}
+
+int registry_add(struct bpf_map_def *map) {
+    struct map_register *tmp_reg = (struct map_register*)malloc(sizeof(struct map_register));
+    HASH_ADD_KEYPTR(hh, maps, map->name, VAR_SIZE, tmp_reg);
```
mbudiu-vmw: can this fail?
Do you always return 0?

backends/ebpf/bpfinclude/ebpf_registry.c
```diff
@@ -0,0 +1,77 @@
...
+void *bpf_map_lookup_elem(struct bpf_map_def *map, void *key) {
+    struct bpf_map *tmp_map;
+    HASH_FIND(hh, map->bpf_map, key, map->key_size, tmp_map);
+    if (tmp_map==NULL)
+        return NULL;
+    return tmp_map->value;
+}
+
+int bpf_map_update_elem(struct bpf_map_def *map, void *key, void *value,
+                  unsigned long long flags) {
```
mbudiu-vmw: what are the flags for?

backends/ebpf/bpfinclude/ebpf_user.h
```diff
@@ -0,0 +1,149 @@
...
+
+#define REGISTER_START() \
+struct bpf_map_def tables[] = {
+#define REGISTER_TABLE(NAME, TYPE, KEY_SIZE, VALUE_SIZE, MAX_ENTRIES) \
+    { #NAME, TYPE, KEY_SIZE, VALUE_SIZE, MAX_ENTRIES },
+#define REGISTER_END() \
+    { 0, 0, 0, 0, 0 } \
+};
+
+int registry_add(struct bpf_map_def *map);
```
mbudiu-vmw: put these in a separate registry.h file

backends/ebpf/ebpf_runtime.c
```diff
@@ -0,0 +1,55 @@
...
+    /* Skip over the program name. */
+    ++argv; --argc;
+    if ( argc != 1 ) {
+        fprintf(stderr, "Missing input trace file, exiting...\n");
+        return EXIT_FAILURE;
+    }
+
+    /* Initialize the registry of shared tables */
+    struct bpf_map_def* current = tables;
+    while (current->name != 0)
```
mbudiu-vmw: Please use statements and not the comma expression.

backends/ebpf/bpfinclude/ebpf_registry.c
fruffy: The flags define different behaviour regarding updates in the hash map.
#define BPF_ANY       0 /* create new element or update existing */
#define BPF_NOEXIST   1 /* create new element only if it didn't exist */
#define BPF_EXIST     2 /* only update existing element */

I initially just added the exact same prototype as the kernel version which is why the flags argument is still there. We can either add behaviour which emulates the kernel version (https://elixir.bootlin.com/linux/v4.17-rc7/source/kernel/bpf/hashtab.c#L796) or just scrap the flags entirely since we only support one type (BPF_ANY).

backends/ebpf/bpfinclude/ebpf_kernel.h
```diff
@@ -0,0 +1,61 @@
...
+#include <linux/skbuff.h>
+#include <linux/netdevice.h>
+#include <linux/version.h>
+#include <uapi/linux/bpf.h>
+
+/* helper macro to place programs, maps, license in
+ * different sections in elf_bpf file. Section names
+ * are interpreted by elf_bpf loader
+i*/
+#define SEC(NAME) __attribute__((section(NAME), used))
```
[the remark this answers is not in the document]
fruffy: As far as I know the only other header file defining these in the kernel is here:
https://github.com/torvalds/linux/blob/f7f4e7fc6c517708738d1d1984b170e9475a130f/tools/perf/include/bpf/bpf.h

afaik SEC(NAME) is a define used by BCC. They have a large header file called helpers.h which contains all kinds of definitions, some of which we are already using extensively. 
https://github.com/iovisor/bcc/blob/master/src/cc/export/helpers.h
I can take a look and see if it makes sense to integrate.

backends/ebpf/bpfinclude/ebpf_registry.c
fruffy: It can fail if the provided key in map is corrupt  (for example NULL or too long) or if malloc fails (unlikely). I can add checks for that.

backends/ebpf/bpfinclude/ebpf_registry.c
```diff
@@ -0,0 +1,77 @@
...
+    if (tmp_map==NULL) {
+        tmp_map = (struct bpf_map*)malloc(sizeof(struct bpf_map));
+        tmp_map->key = key;
+        HASH_ADD_KEYPTR(hh, map->bpf_map, key, map->key_size, tmp_map);
+    }
+    tmp_map->value = value;
+    return 0;
+}
+
+int bpf_map_delete_elem(struct bpf_map_def *map, void *key) {
```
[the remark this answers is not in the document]
fruffy: Both the data and control plane, these are the map operations. bpf_map_* are operations on tables, whereas registry_* are operations to retrieve or add tables to the registry. We can split the files, to logically separate them, but it would add more .c files to compile with the userspace version.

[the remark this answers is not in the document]
mbudiu-vmw: Yes, I think this is good software engineering practice: each module in a separate file. The table registry is in principle reusable, so the APIs should be in a separate file, and the implementation also in a separate file.

backends/ebpf/bpfinclude/ebpf_kernel.h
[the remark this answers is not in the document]
fruffy: https://github.com/torvalds/linux/blob/1c8c5a9d38f607c0b6fd12c91cbe1a4418762a21/tools/testing/selftests/bpf/bpf_helpers.h
This is likely the best include file for the kernel version, but I am not sure if we need all of that. We can either copy it in full and extend it with our customizations (e.g., REGISTER_START() etc) or remove most of the definitions which we do not need right now.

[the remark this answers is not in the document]
fruffy: I do not think so, most variations of this file I have seen are a slightly modified copy. There is no file in include/uapi/linux which contains these definitions. However, this file is generally included by all files in the samples/bpf folder. I think once we actually move on to building the kernel targets we will know how to best structure the header includes.

backends/ebpf/bpfinclude/ebpf_map.c
```diff
@@ -0,0 +1,92 @@
...
+    struct bpf_map *tmp_map;
+    HASH_FIND(hh, map, key, key_size, tmp_map);
+    if (tmp_map != NULL) {
+        HASH_DEL(map, tmp_map);
+        free(tmp_map);
+    }
+    return EXIT_SUCCESS;
+}
+
+int bpf_map_get_next_key(struct bpf_map *map, const void *key, void *next_key){
```
mbudiu-vmw: Maybe it's easier to just delete these methods for now - only keep the ones that are declared in the header file.

backends/ebpf/bpfinclude/ebpf_registry.c
```diff
@@ -0,0 +1,132 @@
...
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+#include <stdio.h>
+#include "ebpf_registry.h"
+
+/**
+ * @brief Defines the structure of the central register.
```
mbudiu-vmw: register or registry?

backends/ebpf/bpfinclude/ebpf_registry.c
```diff
@@ -0,0 +1,132 @@
...
+ * @details Defines a register type, which is a hashmap of
```
mbudiu-vmw: The registry maps names to tables and names to integer handles.

backends/ebpf/bpfinclude/ebpf_registry.c
```diff
@@ -0,0 +1,132 @@
...
+ */
+typedef struct {
+    char name[VAR_SIZE];         // name of the map
+    int map_fd;                  // id of the map
+    struct bpf_map_def *table;   // ptr to the map
+    UT_hash_handle hh;           // makes this structure hashable
+} table_register;
+
+/**
+ * @brief A mapping between names and file descriptors.
```
mbudiu-vmw: Let's call it "handle" instead of file descriptor. These do correspond to file descriptors in the ebpf implementation, but they are not file descriptors here.

backends/ebpf/bpfinclude/ebpf_registry.c
```diff
@@ -0,0 +1,132 @@
...
+ * @details Defines a register type, which is a hashmap of
+ * tables identified by their name. Also associates a unique
+ * integer value as descriptor.
+ */
+typedef struct {
+    char name[VAR_SIZE];         // name of the map
+    int map_fd;                  // id of the map
+    struct bpf_map_def *table;   // ptr to the map
+    UT_hash_handle hh;           // makes this structure hashable
+} table_register;
```
mbudiu-vmw: Let's call this "registry_entry".

backends/ebpf/bpfinclude/ebpf_registry.c
```diff
@@ -0,0 +1,132 @@
...
+
+
+static int map_indexer = 0;
+
+/* Instantiation of the central register */
+table_register *reg_maps = NULL;
+fd_mapping *map_fds = NULL;
+
+
+table_register *_find_table(const char *name) {
```
mbudiu-vmw: make this static and remove the underscore.

backends/ebpf/bpfinclude/ebpf_registry.c
```diff
@@ -0,0 +1,132 @@
...
+    unsigned int key_length = VAR_SIZE;
+    if (strlen(name) > VAR_SIZE)
+        perror("Warning: Key exceeds maximum length.\n");
```
mbudiu-vmw: You print an error, but you will return a potentially incorrect result?
How about you define a function called "fatal" which terminates the program on error?
Alternatively, you should return NULL to indicate that the lookup failed.

backends/ebpf/bpfinclude/ebpf_registry.c
```diff
@@ -0,0 +1,132 @@
...
+void registry_add(struct bpf_map_def *map) {
+    // Check if the register exists already
+    table_register *tmp_reg = _find_table(map->name);
+    if (tmp_reg != NULL)
+        return NULL;
+
+    // Check key size
+    unsigned int key_length = VAR_SIZE;
+    if (strlen(map->name) > VAR_SIZE)
+        perror("Warning: Key exceeds maximum length.\n");
```
mbudiu-vmw: Maybe have this function return an int and return failure here.

backends/ebpf/bpfinclude/ebpf_registry.c
```diff
@@ -0,0 +1,132 @@
...
+    else
+        key_length = strlen(map->name);
+
+    // Add the table
+    tmp_reg = malloc(sizeof(table_register));
```
mbudiu-vmw: You should also check that the allocation succeeded.

backends/ebpf/bpfinclude/ebpf_registry.c
```diff
@@ -0,0 +1,132 @@
...
+    fd_mapping *tmp_fd = malloc(sizeof(fd_mapping));
+    tmp_fd->map_fd = map_indexer;
+    HASH_ADD_INT(map_fds, map_fd, tmp_fd);
+    memcpy(tmp_reg->name, map->name, key_length);
+    map_indexer++;
+}
+
+void registry_delete(const char *name) {
+    table_register *tmp_reg = _find_table(name);
+    if (tmp_reg != NULL) {
```
mbudiu-vmw: You could as well return here success/failure.

backends/ebpf/bpfinclude/ebpf_registry.h
```diff
@@ -0,0 +1,133 @@
...
+ * intended to approximate the kernel ebpf object API as closely as possible.
+ * This library is currently not thread-safe.
+ */
+
+#ifndef BACKENDS_EBPF_BPFINCLUDE_EBPF_REGISTRY_H_
+#define BACKENDS_EBPF_BPFINCLUDE_EBPF_REGISTRY_H_
+
+#include "ebpf_map.h"
+
+#define VAR_SIZE 32  // maximum length of the table name
```
mbudiu-vmw: If that's what it means, please name it MAX_TABLE_NAME_LENGTH

backends/ebpf/bpfinclude/ebpf_map.h
```diff
@@ -0,0 +1,66 @@
...
+ * This file defines a library of simple hashmap operations which emulate the behavior
+ * of the kernel ebpf map API. This library is currently not thread-safe.
+ */
+
+#ifndef BACKENDS_EBPF_BPFINCLUDE_EBPF_MAP_H_
+#define BACKENDS_EBPF_BPFINCLUDE_EBPF_MAP_H_
+
+#include "../contrib/uthash.h"
+
+#define BPF_ANY       0  // create new element or update existing
```
mbudiu-vmw: Can we make this an enum?

backends/ebpf/bpfinclude/ebpf_registry.h
```diff
@@ -0,0 +1,133 @@
...
+ * strings as input.
+ */
+void registry_delete(const char *name);
+
+/**
+ * @brief Retrieve a table from the registry.
+ * @details Retrieves a table from the shared registry.
+ * This operation uses a char name as the key.
+ * map->name should not exceed VAR_SIZE.
+ * This function also assumes conventional, null-terminated
```
mbudiu-vmw: I don't think you need to repeat the strings constraints everywhere; just write it once in the table name declaration.

backends/ebpf/bpfinclude/ebpf_registry.h
```diff
@@ -0,0 +1,133 @@
...
+ * @brief A helper structure used to describe attributes.
+ * @details This structure describes various properties of the ebpf table
+ * such as key and value size and the maximum amount of entries possible.
+ * In userspace, this space is theoretically unlimited.
+ * This table definition points to an actual hashmap managed by uthash,
+ * the relation is many-to-one.
+ *
+ */
+struct bpf_map_def {
+    char *name;                 // table name longer than VAR_SIZE is not accessed
```
mbudiu-vmw: Put the constraints here, and remove them from all the functions accessing tables.

backends/ebpf/ebpf_runtime.c
```diff
@@ -0,0 +1,55 @@
...
+int main(int argc, char **argv) {
+    pcap_t *in_handle;
+    pcap_dumper_t *out_handle;
+    const unsigned char *packet;
+    char errbuf[PCAP_ERRBUF_SIZE];
+    struct pcap_pkthdr *pcap_hdr;
+
+    /* Skip over the program name. */
+    ++argv; --argc;
+    if ( argc != 1 ) {
```
mbudiu-vmw: No spaces inside ().

backends/ebpf/ebpf_runtime.c
```diff
@@ -0,0 +1,55 @@
...
+        fprintf(stderr, "Missing input trace file, exiting...\n");
+        return EXIT_FAILURE;
+    }
+
+    /* Initialize the registry of shared tables */
+    struct bpf_map_def* current = tables;
+    while (current->name != 0) {
+        registry_add(current);
```
mbudiu-vmw: You should also check the result once you have one.

backends/ebpf/ebpf_runtime.c
```diff
@@ -0,0 +1,55 @@
...
+    }
+
+    /* Parse each packet in the file and check the result*/
+    while ((packet = pcap_next(in_handle, pcap_hdr)) != NULL) {
+        struct sk_buff skb;
+        skb.data = (void *) packet;
+        skb.len = pcap_hdr->len;
+        int result = ebpf_filter(&skb);
+        printf("\nebpf result is: %d\n", result);
+        pcap_dump((unsigned char *) out_handle, pcap_hdr, packet);
```
mbudiu-vmw: You should only write the packet to the output if it was not filtered.

backends/ebpf/bpfinclude/ebpf_registry.c
fruffy: Returning NULL makes sense, should I still provide feedback with perror or just return NULL?

backends/ebpf/bpfinclude/ebpf_map.c
fruffy: Done, I realized they are basically eclipsed by the registry anyway.

backends/ebpf/bpfinclude/ebpf_registry.c
fruffy: Oh yeah right. i updated with some instrumentation.

backends/ebpf/ebpf_runtime.c
fruffy: right now we are always dropping, which is why I always dump for testing purposes. I added a check which only writes if the packet is not "filtered"

backends/ebpf/ebpf_runtime.c
fruffy: done

backends/ebpf/ebpf_runtime.c
fruffy: done

backends/ebpf/bpfinclude/ebpf_registry.h
fruffy: done

backends/ebpf/bpfinclude/ebpf_registry.h
fruffy: done

backends/ebpf/bpfinclude/ebpf_registry.h
fruffy: done

backends/ebpf/bpfinclude/ebpf_registry.c
fruffy: done

backends/ebpf/bpfinclude/ebpf_registry.c
fruffy: done

backends/ebpf/bpfinclude/ebpf_registry.c
fruffy: done

backends/ebpf/bpfinclude/ebpf_registry.c
fruffy: done

backends/ebpf/bpfinclude/ebpf_registry.c
fruffy: done

backends/ebpf/bpfinclude/ebpf_registry.c
fruffy: done

backends/ebpf/bpfinclude/ebpf_registry.c
fruffy: done

backends/ebpf/bpfinclude/ebpf_map.h
fruffy: done

backends/ebpf/bpfinclude/ebpf_map.c
```diff
@@ -0,0 +1,92 @@
...
+    // TODO: Implement
+    return EXIT_SUCCESS;
+}
+
+struct bpf_map *bpf_map_get_next_id(unsigned int start_id, unsigned int *next_id) {
+    // TODO: Implement
+    return EXIT_SUCCESS;
```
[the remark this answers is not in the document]
fruffy: done

backends/ebpf/bpfinclude/ebpf_map.c
```diff
@@ -0,0 +1,92 @@
...
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+#include "ebpf_map.h"
+
+int check_flags(void *elem, unsigned long long map_flags) {
```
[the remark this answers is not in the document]
fruffy: done

backends/ebpf/bpfinclude/ebpf_map.c
```diff
@@ -0,0 +1,73 @@
...
+                  unsigned long long flags) {
+    struct bpf_map *tmp_map;
+    HASH_FIND(hh, map, key, key_size, tmp_map);
+    int ret = check_flags(tmp_map, flags);
+    if (ret)
+        return ret;
+    if (!tmp_map) {
+        tmp_map = (struct bpf_map *) malloc(sizeof(struct bpf_map));
+        tmp_map->key = key;
+        HASH_ADD_KEYPTR(hh, map, key, key_size, tmp_map);
```
mbudiu-vmw: Can these fail?

backends/ebpf/bpfinclude/ebpf_map.c
```diff
@@ -0,0 +1,73 @@
...
+    }
+    tmp_map->value = value;
+    return EXIT_SUCCESS;
+}
+
+int bpf_map_delete_elem(struct bpf_map *map, void *key, unsigned int key_size) {
+    struct bpf_map *tmp_map;
+    HASH_FIND(hh, map, key, key_size, tmp_map);
+    if (tmp_map != NULL) {
+        HASH_DEL(map, tmp_map);
```
mbudiu-vmw: How about this one?

backends/ebpf/bpfinclude/ebpf_map.h
```diff
@@ -0,0 +1,67 @@
...
+
+#ifndef BACKENDS_EBPF_BPFINCLUDE_EBPF_MAP_H_
+#define BACKENDS_EBPF_BPFINCLUDE_EBPF_MAP_H_
+
+#include "../contrib/uthash.h"  // exports string.h, stddef.h, and stdlib.h
+
+// BPF_ANY create new element or update existing
+// BPF_NOEXIST: create new element only if it didn't exist
+// BPF_EXIST: only update existing element
+enum bpf_flags { BPF_ANY, BPF_NOEXIST, BPF_EXIST};
```
mbudiu-vmw: You can still put comments next to each value; some IDEs will pick them up and help when typing.

backends/ebpf/bpfinclude/ebpf_map.h
```diff
@@ -0,0 +1,67 @@
...
+    UT_hash_handle hh;  // makes this structure hashable
+};
+
+/**
+ * @brief Add/Update a value in the map
+ * @details Updates a value in the map based on the provided key.
+ * If the key does not exist, it depends the provided flags if the
+ * element is added or the operation is rejected.
+ *
+ * @return error if update operation fails
```
mbudiu-vmw: Maybe you want to say EXIT_FAILURE instead of "error".

backends/ebpf/bpfinclude/ebpf_registry.c
```diff
@@ -0,0 +1,147 @@
...
+static int map_indexer = 0;
+
+/* Instantiation of the central registry */
+registry_entry *reg_tables = NULL;
+handle_entry *table_handles = NULL;
+
+
+static registry_entry *find_table(const char *name) {
+    if (strlen(name) > MAX_TABLE_NAME_LENGTH){
+        fprintf(stderr, "%s", "Error: Key length exceeds maximum size!\n");
```
mbudiu-vmw: I don't think that the '%s' is necessary here, but you could print the name.
Also, you may check if name is NULL and return an error.

backends/ebpf/ebpf_runtime.c
```diff
@@ -0,0 +1,60 @@
+#include <string.h>
+#define PCAP_DONT_INCLUDE_PCAP_BPF_H
+#include <pcap/pcap.h>
+#include "test.h"
+
+int main(int argc, char **argv) {
```
mbudiu-vmw: You should probably document the arguments expected.

backends/ebpf/bpfinclude/ebpf_map.c
fruffy: In case of malformed pointers or arguments these may segfault, but I do not think they return errors. 
uthash has a debug mode: 
"If a program that uses this hash is compiled with -DHASH_DEBUG=1, a special internal consistency-checking mode is activated. In this mode, the integrity of the whole hash is checked following every add or delete operation. This is for debugging the uthash software only, not for use in production code."
We could run in in this mode by default to ensure correctness of the hash library.

PR #1332, opened 2018-06-12
bugfix: default_action translation to p4-16

hanw:
During the P4-14 to 16 translation, when a table has an attached direct_counter
or direct meter, the translation inserts a counter.count() or meter.execute()
method call to each of the action in the table.

As a result, if an action is used in the default_action property of the table,
the default_action has changed from the original unmodified action to the
updated action with counter.count() or meter.execute().

The current P4-14 to 16 translation has a bug which does not update the
default_action property in the table, and the wrong default_action is used when
a direct_counter or meter present.

### DIFF
--- a/frontends/p4/fromv1.0/programStructure.cpp
+++ b/frontends/p4/fromv1.0/programStructure.cpp
@@ -758,6 +758,7 @@ ProgramStructure::convertTable(const IR::V1Table* table, cstring newName,
     auto ctr = get(directCounters, table->name);
     const std::vector<IR::ID> &actionsToDo =
             action_profile ? action_profile->actions : table->actions;
+    cstring default_action = table->default_action;
     for (auto a : actionsToDo) {
         auto action = actions.get(a);
         cstring newname;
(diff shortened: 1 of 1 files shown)


ChrisDodd:
Its not entirely clear to me if this is/was a bug or a feature -- its not entirely clear what the semantics of direct counters/meters is in P4_14 with respect to miss actions.  Is there a counter for the miss/default action that is incremented, or only for each hit action?  If the former, then any action that can be used for either hit or miss actions must be split into two versions -- one with the counter/meter operation (for use as a hit action) and one without.

hanw:
@ChrisDodd  A possible solution is to name the miss action as `NoAction`? 

P4-14 spec says:
```
A default action is taken when no table entry matches. This action is specified at
run time. If no default action is specified and no entry matches, the table does not
affect the packet and processing continues according to the imperative control
flow.
```

Currently the miss action is named after the original action. As a result, we end up with two different actions with the same external name.

hanw:
@ChrisDodd I am merging this now to unblock others, but let's revisit the issue after you are back.

PR #1343, opened 2018-06-13
 Use dotPrefix non-terminal for expressions having a dot prefix

SantiagoBautista:
dotPrefix non-terminal was not used for expressions starting with dot prefixes, which meant they were parsed in the local context instead of being parsed at the toplevel context (and this can lead to local types interfering with toplevel nonTypeName s).

### DIFF
--- a/frontends/parsers/p4/p4parser.ypp
+++ b/frontends/parsers/p4/p4parser.ypp
@@ -1044,7 +1044,7 @@ expression
     | FALSE                              { $$ = new IR::BoolLiteral(@1, false); }
     | THIS                               { $$ = new IR::This(@1); }  // experimental
     | nonTypeName                        { $$ = new IR::PathExpression(*$1); }
-    | "." nonTypeName                    { $$ = new IR::PathExpression(new IR::Path(*$2, true)); }
+    | dotPrefix nonTypeName              { $$ = new IR::PathExpression(new IR::Path(*$2, true)); driver.structure->clearPath(); }
     | expression "[" expression "]"      { $$ = new IR::ArrayIndex(@1 + @4, $1, $3); }
     | expression "[" expression ":" expression "]" { $$ = new IR::Slice(@1 + @6, $1, $3, $5); }
     | "{" expressionList "}"             { $$ = new IR::ListExpression(@1 + @3, *$2); }


mbudiu-vmw:
Please add a test case which exhibits this bug.

SantiagoBautista:
For now I have this example, that should be recognized as syntactically valid
```
const bit X = 1;

bit f<X>() {
  return .X ;
}
```

but i am trying to think of an example that does not use functions.

mbudiu-vmw:
One example is good enough.

SantiagoBautista:
This example cannot yet be included in the testdata directory, because functions are not yet supported by the version of the compiler that is in the `master` branch.

But yes, this example is good enough for understanding the bug I am talking about:
in the return statement, `.X` corresponds to a toplevel constant and not to a type; which means `X` should be lexed as `IDENTIFIER` and not `TYPE_IDENTIFIER`, which requires a call to `driver.structure->startAbsolutePath()` to prevent local type `X` from interfering with `X` being recognized as an `IDENTIFIER`. This call to `driver.structure->startAbsolutePath()` is added by the fact of using `dotPrefix` instead of `'.'` , and it requires a call to `driver.structure->clearPath()` after the rule corresponding to `.X` has been reduced.

mbudiu-vmw:
We support abstract functions, so that would work in that setting. That's an experimental feature in the language. Grep for abstract in the samples folder.

SantiagoBautista:
This should be a minimal example of file syntactically correct without the bug but rejected with the bug:
```
const bit X = 1; 

extern Virtual {
    Virtual();
    abstract bit f<X>();
}

Virtual() cntr = {
  bit f<X>(){
    return .X
  }
}
```

SantiagoBautista:
Abstract method implementations cannot be generic, and I need type parameters to show what I want to show. (Otherwise there is no interference between a type and a constant).

So no, I cannot find an example that I can run with the current compiler; and hence I cannot find an example to add to the test cases.

As I cannot find an example that exhibits the bug maybe I shouldn't call it a bug.

But that doesn't change the fact that for expressions, when a dot prefix is encountered, the fact that the next token is `IDENTIFIER` or `TYPE_IDENTIFIER` should be decided whit the toplevel context and not with the local context.
And this pr does this, without changing the compiler behavior in any other way.

If my explanations does not suffice to convince you, I won't insist.

mbudiu-vmw:
It has nothing to do about convincing me, it is about having test cases to catch this example.
You can file an issue to add this test case once we have functions working.

SantiagoBautista:
Ok, I understand.

For now I cannot find a test case. I will add one when functions work :)

SantiagoBautista:
Meanwhile I think this particular pr is ready for review and merge.

PR #1348, opened 2018-06-18
Add Kernel Compilation and loading to the eBPF tests

fruffy:
This pull request adds the automatic creation, compilation, and loading of eBPF programs as part of the test framework.
The test uses clang with the option -emit-llvm to generate an intermediate representation for llc, which in turn compiles the input to ebpf byte code. The byte code can then be verified using the newly added, ebpf_loader, which checks if the kernel accepts the input.
- Currently, the test files are compiled over hardcoded python scripts. A more maintainable alternative is to convert the scripts to a Makefile with different targets. So far only the kernel target is implemented over makefiles.
- New dependencies are: libelf-dev, llvm (3.7.1+) , clang, iproute2, and the python package pyroute2
- If the kernel version is less than 4.15.x, the tests are ignored.
- Files for target test are now in folder "testinclude"
- Several improvements to code readability and comments have been added. The Python files now follow the PEP8 style more closely.
- The kernel tests require sudo in order to pass. (Adding a check to make sure this is the case is a todo)


Dockerfile
```diff
@@ -27,8 +27,12 @@ ENV P4C_DEPS bison \
              python-ipaddr \
              python-pip \
              python-setuptools \
-             tcpdump \
-             libpcap-dev
+             tcpdump
+ENV P4C_EBPF_DEPS libpcap-dev \
```
mbudiu-vmw: Please update the ebpf/README.md file to list all these dependences.
You may need a specific version of llvm.

backends/ebpf/CMakeLists.txt
```diff
@@ -94,6 +98,38 @@ set (EBPF_TEST_SUITES
...
+string (REGEX MATCH "[0-9]+[^-]*" KERNEL_VER ${CMAKE_SYSTEM})
+if (${KERNEL_VER} VERSION_LESS ${MIN_KERNEL} )
+  MESSAGE(WARNING "Kernel version too small, ignoring ebpf kernel tests...")
+  set (SUPPORTS_KERNEL False)
+endif()
+
+# Check if we have the right llvm version
+
+find_package(LLVM)
+if(LLVM_PACKAGE_VERSION)
```
mbudiu-vmw: space after if

backends/ebpf/CMakeLists.txt
```diff
@@ -94,6 +98,38 @@ set (EBPF_TEST_SUITES
...
   )
 
-p4c_add_tests("ebpf-kernel" ${EBPF_DRIVER_KERNEL} ${EBPF_TEST_SUITES} "${XFAIL_TESTS_KERNEL}")
+set (SUPPORTS_KERNEL True)
+
+# Check if the kernel version is new enough to support ebpf features
+set (MIN_KERNEL 4.16.0)
+string (REGEX MATCH "[0-9]+[^-]*" KERNEL_VER ${CMAKE_SYSTEM})
+if (${KERNEL_VER} VERSION_LESS ${MIN_KERNEL} )
+  MESSAGE(WARNING "Kernel version too small, ignoring ebpf kernel tests...")
```
mbudiu-vmw: Maybe you can output the minimum version required, so people know what to install

backends/ebpf/CMakeLists.txt
```diff
@@ -94,6 +98,38 @@ set (EBPF_TEST_SUITES
...
+  set (SUPPORTS_KERNEL False)
+endif()
+
+# Check if we have the right llvm version
+
+find_package(LLVM)
+if(LLVM_PACKAGE_VERSION)
+  message(STATUS "Found LLVM ${LLVM_PACKAGE_VERSION}")
+  if (${LLVM_PACKAGE_VERSION} VERSION_LESS 3.7.1 )
+    MESSAGE(WARNING "LLVM version too small, ignoring ebpf tests...")
```
mbudiu-vmw: Please output the required version in the message; perhaps also the found version.

backends/ebpf/CMakeLists.txt
```diff
@@ -94,6 +98,38 @@ set (EBPF_TEST_SUITES
...
+    set (SUPPORTS_KERNEL False)
+  endif()
+else()
+  message(STATUS "LLVM missing")
```
mbudiu-vmw: You should also say that this will disable the kernel tests

backends/ebpf/CMakeLists.txt
mbudiu-vmw: This only ignores the kernel tests.

backends/ebpf/ebpf_kernel.h
```diff
@@ -0,0 +1,157 @@
+/*
```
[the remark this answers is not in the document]
mbudiu-vmw: Will you do this move in this PR?

backends/ebpf/ebpf_runtime.c
```diff
@@ -1,12 +1,36 @@
...
 #define PCAP_DONT_INCLUDE_PCAP_BPF_H
 #include <pcap/pcap.h>
 #include "test.h"
 
+#define FILE_NAME_MAX 256
 
-void usage() {
-    printf("This program parses a pcap file, feeds the individual packets into a filter function, and returns the output.\n");
+void usage(char *name) {
+    printf("usage: %s [file].pcap\n"
```
mbudiu-vmw: File should probably not be in [], that usually means optional.

backends/ebpf/ebpftargets.py
```diff
@@ -23,24 +23,25 @@
...
+from pyroute2 import IPRoute
 
 SUCCESS = 0
 FAILURE = 1
 TIMEOUT = 10 * 60
 
 
-def nextWord(text, sep=None):
-    ''' Split a text at the indicated separator.
+def next_word(text, sep=None):
```
mbudiu-vmw: Just renaming functions that are otherwise identical will make it very hard to remove this duplicated code later if we decide to refactor it.

backends/ebpf/ebpftargets.py
```diff
@@ -50,116 +51,140 @@ def nextWord(text, sep=None):
...
-            reportError("Full received packed is ", received)
+            report_err(outputs["stderr"], "Received packet ", received)
+            report_err(outputs["stderr"], "Packet different at position", i,
+                       ": expected", expected[i], ", received", received[i])
+            report_err(outputs["stderr"], "Full received packed is ", received)
             return FAILURE
     return SUCCESS
 
 
+def require_root(outputs):
```
mbudiu-vmw: How are people supposed to run this test with root permissions?
They should not run all tests with root permissions.

backends/ebpf/ebpftargets.py
```diff
@@ -168,58 +193,67 @@ def interface_of_filename(self, f):
...
     def generate_model_inputs(self, stffile):
-        ''' To override '''
-        ''' Parses the stf file and creates a .pcap file with input packets.
+        """ To override """
+        """ Parses the stf file and creates a .pcap file with input packets.
             It also adds the expected output packets per interface to a global
             dictionary.
             TODO: Create pcap packet data structure and differentiate
-            between input interfaces.'''
+            between input interfaces."""
```
mbudiu-vmw: In general you don't want TODO in the comment describing the function. That should be a separate comment.

backends/ebpf/ebpftargets.py
```diff
@@ -251,98 +285,159 @@ def checkOutputs(self):
...
+        args.append("IFACE=" + ifname)
+        errmsg = "Failed to compile and load the eBPF byte code:"
+        return run_timeout(self.options, args, TIMEOUT, self.outputs, errmsg)
 
-    def checkOutputs(self):
+    def create_filter(self):
+
+        require_root(self.outputs)
+        # Create interface with unique id (allows parallel tests)
+        ifname = "test" + str(os.getpid())
```
mbudiu-vmw: Only one interface? What if you have multiple pcap files?

backends/ebpf/ebpftargets.py
```diff
@@ -251,98 +285,159 @@ def checkOutputs(self):
...
 
 
 class EBPFBCCTarget(EBPFTarget):
-    def __init__(self, tmpdir, options, template, stderr):
-        EBPFTarget.__init__(self, tmpdir, options, template, stderr)
+    def __init__(self, tmpdir, options, template, outputs):
+        EBPFTarget.__init__(self, tmpdir, options, template, outputs)
+
+    def compile_p4(self, argv):
+        """ To override """
```
mbudiu-vmw: This comment is no longer necessary

backends/ebpf/ebpftargets.py
```diff
@@ -168,58 +193,67 @@ def interface_of_filename(self, f):
                    lstrip(self.pcapPrefix).rsplit('_', 1)[0])
 
     def compile_p4(self, argv):
-        ''' To override '''
-        ''' Compile the p4 target '''
+        """ To override """
```
mbudiu-vmw: I think that this comment should not be the function documentation, it should be a separate comment with `#`

backends/ebpf/ebpftargets.py
```diff
@@ -251,98 +285,159 @@ def checkOutputs(self):
...
+
+    def check_outputs(self):
         # Not implemented yet, just pass the test
         return SUCCESS
 
 
 class EBPFBCCTarget(EBPFTarget):
-    def __init__(self, tmpdir, options, template, stderr):
-        EBPFTarget.__init__(self, tmpdir, options, template, stderr)
+    def __init__(self, tmpdir, options, template, outputs):
```
mbudiu-vmw: You must also check if bcc is installed in the makefile

backends/ebpf/ebpftargets.py
```diff
@@ -251,98 +285,159 @@ def checkOutputs(self):
...
     def run(self):
-        if self.options.verbose:
-            print("Running model")
-        # compiler
-        args = []
-        # main
-        args.append(self.tmpdir + "/filter")
+        report_output(self.outputs["stdout"],
+                      self.options.verbose, "Running model")
+        # main executable
```
mbudiu-vmw: why is the comment here twice?

backends/ebpf/kernel.mk
```diff
@@ -0,0 +1,91 @@
...
+INCLUDES+= -I./
+LIBS+=
+# Optimization flags to save space
+CFLAGS+= -O2 -g -D__KERNEL__ -D__ASM_SYSREG_H -Wno-unused-value -Wno-pointer-sign \
+		-Wno-compare-distinct-pointer-types \
+		-Wno-gnu-variable-sized-type-not-at-end \
+		-Wno-address-of-packed-member -Wno-tautological-compare \
+		-Wno-unknown-warning-option
+
+# Arguments for the P4 Compiler
```
mbudiu-vmw: Add a TODO here, we may need to use the installation folder for p4include, and not ./

backends/ebpf/kernel.mk
```diff
@@ -0,0 +1,91 @@
...
+TARGET=kernel
+# Extra arguments for the compiler
+ARGS=
+
+# If needed, bpf target files can be hardcoded here
+# This can be any file of type ".c", ".bc" or, ".o"
+BPFOBJ=
+# Get the source name of the object to match targets
+BPFNAME=$(basename $(BPFOBJ))
+# Interface to attach programs to
```
mbudiu-vmw: In general there will be multiple interfaces.

backends/ebpf/kernel.mk
```diff
@@ -0,0 +1,91 @@
...
+
+# Invoke the llvm on the generated .bc code and produce bpf byte code
+$(BPFNAME).o: %.o : %.bc
+	$(LLC) -march=bpf -mcpu=probe -filetype=obj $< -o $@
+
+clean: clean_loader
+	rm -f *.o *.bc $(BPFNAME).c $(BPFNAME).h
+
+# For actually attaching BPF programs
+attach:
```
mbudiu-vmw: I don't think this should be in the makefile; you should move this in the python script. And you will need to do this for every interface; in general there will be multiple interfaces. Use the makefiles only for compiling.

backends/ebpf/kernel.mk
```diff
@@ -0,0 +1,91 @@
...
+	ip link show $(IFACE); \
+	tc qdisc add dev $(IFACE) clsact; \
+	for TOOL in $(BPFOBJ); do \
+		tc filter add dev $(IFACE) ingress bpf da $(TOOL) sec prog verb; \
+	done
+
+# Detach the loaded programs
+detach:
```
mbudiu-vmw: This too.

backends/ebpf/ebpftargets.py
```diff
@@ -313,36 +393,33 @@ def create_filter(self):
...
         args = ["gcc"]
+        args.append("-g")
         # main
         args.append("-o")
         args.append(self.tmpdir + "/filter")
         # sources
         args.append(ebpfdir + "/ebpf_runtime.c")
-        args.append(ebpfdir + "/bpfinclude/ebpf_registry.c")
-        args.append(ebpfdir + "/bpfinclude/ebpf_map.c")
+        args.append(ebpfdir + "/testinclude/ebpf_registry.c")
```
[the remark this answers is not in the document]
fruffy: You mean testinclude or ebpf_runtime.c?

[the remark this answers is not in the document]
mbudiu-vmw: We should move all the hand-written C files in the runtime directory.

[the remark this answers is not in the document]
fruffy: It would also make sense then to move all the .cpp files into a compiler folder, or are there external dependencies on that folder structure which might break?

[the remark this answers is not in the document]
mbudiu-vmw: The other back-ends are structured similarly, so it's better to keep the cpp files here for homogeneity.

mbudiu-vmw:
Your PR has failed the checks.

fruffy:
Very annoying... I forgot to rebuild the compiler before running the local tests. Slipped through the cracks. Everything should work now.

PR #1351, opened 2018-06-20
fixes for issue 1350

hanw:


mbudiu-vmw:
We have never merged PR #1289. Maybe we should. @hanw : can you take a look at it?

isaac-ped:
I seem to be getting the same error as before. If I define an extern function in the p4 file, which is not explicitly defined as an extern in the backed, it still fails:
```
error: Unknown extern function extern_fn
extern void extern_fn(inout bit<16> hdr);
            ^^^^^^^^^
```
Is this what was supposed to be changed?

hanw:
Can you attach your program please?

hanw:
This PR does not handle extern function, extern function is handled in #1289, which has not been merged.

isaac-ped:
Ah - sorry. Did not realize what this was targeting. Running similar code with an extern object/method instead results in the same issue, if that is what this fix was for:

```
test.p4(48): error: Unknown extern method update from type ExternObj
    void update(inout bit<16> t);
         ^^^^^^
test.p4(46)
extern ExternObj {
       ^^^^^^^^^
```

With code containing:
```
extern ExternObj {
    ExternObj();
    void update(inout bit<16> t);
}

hanw:
@isaac-ped try again?

isaac-ped:
Working perfectly now!

Issue #1352, opened 2018-06-20
v1model digest extern fails bmv2 assertion


rosstex:
I am doing a simple digest test with bmv2-ss and the V1model, and calling the digest() extern causes the bmv2 switch to crash after the first packet it sees.

The issue is:
```
[21:15:00.182] [bmv2] [D] [thread 15611] [0.0] [cxt 0] Table 'tbl_act_1': miss
[21:15:00.182] [bmv2] [D] [thread 15611] [0.0] [cxt 0] Action entry is act_1 -
Assertion 'Default switch case should not be reachable' failed, file '../../include/bm/bm_sim/actions.h' line '318'.
```

My headers:
```
struct test_digest_t {
    macAddr_t in_mac_srcAddr;
}
struct metadata {
    bool                dfa_accept;
    dfa_accept_digest_t dfa_accept_digest;
    test_digest_t       test_digest;
}
```

In my Ingress control's apply func, I have:
```meta.test_digest.in_mac_srcAddr = hdr.ethernet.srcAddr;```

and in Egress, I have:
```digest(1, meta.test_digest);```

The generated .json is:
```
    {
      "name" : "act_1",
      "id" : 15,
      "runtime_data" : [],
      "primitives" : [
        {
          "op" : "assign",
          "parameters" : [
            {
              "type" : "header",
              "value" : "tmp_0"
            },
            {
              "type" : "header",
              "value" : "userMetadata.test_digest"
            }
          ],
          "source_info" : {
            "filename" : "ross_dfa_int.p4",
            "line" : 286,
            "column" : 18,
            "source_fragment" : "meta.test_digest"
          }
        },
        {
          "op" : "generate_digest",
          "parameters" : [
            {
              "type" : "hexstr",
              "value" : "0x00000001"
            },
            {
              "type" : "hexstr",
              "value" : "0x2"
            }
          ]
        }
      ]
    }
```

I'm not sure if this is a compiler issue or a bmv2 issue.

antoninbas:
The "op" should be "assign_header" not "assign" for header assignments. This is indeed a bug in the bmv2 compiler backend.

mbudiu-vmw:
The back-end code seems to generate "assign_header" for headers. Do you have a test programs we can use to reproduce this?

rosstex:
Sure, here's a stripped-down p4 implementation that has this bug.

[...]

PR #1353, opened 2018-06-20
Fix for issue #1352

mbudiu-vmw:


backends/bmv2/simple_switch/simpleSwitch.cpp
```diff
@@ -895,6 +895,10 @@ SimpleSwitchBackend::convert(const IR::ToplevelBlock* tlb) {
         evaluator,
         new VisitFunctor([this, evaluator]() { toplevel = evaluator->getToplevelBlock(); }),
     };
+
+    auto hook = options.getDebugHook();
```
antoninbas: was that just for debugging?

mbudiu-vmw:
We have a pass that eliminates structure copies, but this particular structure is created later in the backend.

backends/bmv2/simple_switch/simpleSwitch.cpp
mbudiu-vmw: Mostly; it prints the passes as they get executed, but it also enables you to dump the program after each pass.

PR #1358, opened 2018-06-22
shift has lower precedence than minus

hanw:


hanw:
Will update the reference outputs for two test cases.

PR #1361, opened 2018-06-22
Add support for PSA in p4RuntimeSerializer

antoninbas:
Because PSA and v1model share many similarities (not to mention that
they are both considered standard architectures in the context of
P4Runtime), we tried to unify the code for the two architectures as much
as possible using meta-programming. We define a "traits" class for each
extern type to handle the difference in naming (table property names,
extern type names, extern method parameter names). One major difference
between the two architectures is digest: v1model uses an extern method
while PSA defines an extern type.

We added a new gtest unit test for the PSA code, very similar to the
ones we already have for v1model. Unfortunately PSA requires a lot of
boiler-plate code so the test is pretty verbose. We may want to find a
better solution in the future. The unit test covers digest, since as
pointed out before this is the major difference with v1model.

### DIFF
--- a/control-plane/p4RuntimeArchStandard.cpp
+++ b/control-plane/p4RuntimeArchStandard.cpp
@@ -443,6 +443,13 @@ class P4RuntimeArchHandlerCommon : public P4RuntimeArchHandlerIface {
         }
     }
 
+    void collectExternFunction(P4RuntimeSymbolTableIface* symbols,
+                               const P4::ExternFunction* externFunction) override {
+        // no common task
+        (void)symbols;
+        (void)externFunction;
+    }
+
     void postCollect(const P4RuntimeSymbolTableIface& symbols) override {
         (void)symbols;
         // analyze action profiles and build a mapping from action profile name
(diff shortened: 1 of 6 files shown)


control-plane/p4RuntimeArchStandard.cpp
```diff
@@ -701,12 +743,210 @@ class P4RuntimeArchHandlerV1Model final : public P4RuntimeArchHandlerIface {
...
+        auto decl = externBlock->node->to<IR::Declaration_Instance>();
+        if (decl == nullptr) return;
+        auto p4RtTypeInfo = p4info->mutable_type_info();
+        if (externBlock->type->name == "Digest") {
+            auto digest = getDigest(decl, p4RtTypeInfo);
+            if (digest) addDigest(symbols, p4info, *digest);
+        }
+    }
+
+    void addExternFunction(const P4RuntimeSymbolTableIface& symbols,
```
mbudiu-vmw: Can these have this particular default implementation in the base class?

antoninbas: Usually I like to make sure the "interface" (here `P4RuntimeArchHandlerIface`) is a pure abstract class. However I don't mind putting the default implementations in the common base class (between v1model & PSA): `P4RuntimeArchHandlerCommon`.

antoninbas:
@mbudiu-vmw Thanks for the review!
I think it would be great if we had a way to have some PSA boiler-plate code in a companion library. However, I'm not sure how practical it would be. The type parameters (e.g. for user-defined metadata) would have to match across all controls (if I want to use some of the default empty ones and implement the other controls myself). I feel like that would be tricky to do.

PR #1363, opened 2018-06-22
Minor extension: prevent defined types from disabling table properties

SantiagoBautista:
In the spec, it is written that each architecture may have its own table properties, like, for example, an `implementation` table property.

But table property names, in the current grammar, cannot be type names.

So imagine that a programmer accidentally defines a type that has the same name as a table property for the architecture he is targeting (or includes another P4 program that does). He could no longer use the table property in tables, as it would be seen as a syntax error.

This pr allows table properties to be type names, preventing this problem.

### DIFF
--- a/frontends/parsers/p4/p4parser.ypp
+++ b/frontends/parsers/p4/p4parser.ypp
@@ -160,7 +160,7 @@ typedef const IR::Type ConstType;
 %token<cstring> IDENTIFIER TYPE_IDENTIFIER STRING_LITERAL
 %token<UnparsedConstant>  INTEGER
 
-%type<IR::ID*>              name nonTypeName
+%type<IR::ID*>              name nonTypeName nonTableKwName
 %type<IR::Direction>        direction
 %type<IR::Path*>            prefixedNonTypeName prefixedType
 %type<IR::IndexedVector<IR::Type_Var>*>  typeParameterList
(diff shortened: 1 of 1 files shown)


frontends/parsers/p4/p4parser.ypp
```diff
@@ -913,10 +921,10 @@ tableProperty
         { auto l = new IR::EntriesList(@3, *$6);
           auto id = IR::ID(@3+@7, "entries");
           $$ = new IR::Property(@3, id, $1, l, $2.isConst); }
-    | optAnnotations optCONST IDENTIFIER "=" initializer ";"
+    | optAnnotations optCONST nonTableKwName "=" initializer ";"
         { auto v = new IR::ExpressionValue(@5, $5);
-          auto id = IR::ID(@3, $3);
-          $$ = new IR::Property(@3, id, $1, v, $2.isConst); }
+          auto id = *$3;
+          $$ = new IR::Property(id, $1, v, $2.isConst); }
```
mbudiu-vmw: You should still have the srcInfo for the property to give good error messages.
Use `$3->srcInfo`.

cc10512:
@SantiagoBautista thank you for scratching an old itch! Would you be able to expand on this work and disambiguate table names from parameter names? And in general, enable the same name to be used for different entities depending on scope.

SantiagoBautista:
Thanks for your comment, @cc10512 !

For now I am just checking, in the grammar and in the parser, if the distinction between `IDENTIFIER` and `TYPE_IDENTIFIER` has no inconveniences (that is what originated this pr).

But sure, in my future work developing a P4 interpreter I can try and keep separate contexts for tables and parameters.

PR #1364, opened 2018-06-23
Add a cmake option to disable GC

dmakarov:
BDW garbage collector is extremely buggy and plain does not work.  It would be very useful to have a build option to disable the GC completely.  This PR adds such an option.  Please, take a look.


mbudiu-vmw:
I have never seen a problem related to the GC in the last 3 years since we started this codebase. Have you measured memory consumption for large programs?

ChrisDodd:
I have seen two problems with the garbage collection so far, both only on OSX, and both worked around in the code -- see the DANGER comments in lib/gc.cpp and lib/indent.cpp.  Disabling garbage collection on a large program (such as switch.p4) requires 10+ GB of memory, so may not be practical.

dmakarov:
I'm using p4c as a library in a multithreaded language server application. GC just doesn't work. I get an obscure error "collecting from unknown thread" from the GC, and have little interest debugging BDW GC.  The PR is not about disabling GC, but having an option to do so when building p4c for those who might need this.  For all purposes everything remains the same for those who do not want to disable the GC.

ChrisDodd:
That error message generally means you failed to enable the GC library's multlthreading support -- generally requires -DGC_PTHREADS and `#include <gc.h>` before calling any pthread thread creation routines.

jnfoster:
Have you read the page on using BDW in multithreaded contexts?
http://www.hboehm.info/gc/scale.html

I don't know how you are building libgc or your application, but seems
slightly unfair to criticize libgc as "extremely buggy" and say it "plain
just doesn't work" when the authors explicitly say it is not thread safe in
its default configuration.

-N

dmakarov:
I'm using Homebrew bdw-gc package (not building libgc from source).

Before I made the server a multithreaded application, it was single-threaded, and BDW corrupted the stack of my application.  From my point of view it's enough to call the software "extremely buggy".

So far, I disabled GC in my builds of p4c libraries.  If the proposed cmake option is an unacceptable change, feel free to close the PR. I'll just continue using my fork.

mbudiu-vmw:
Can you try to see whether the suggestions above can help with the multi-threading issues you experienced?

ChrisDodd:
I guess it's kind-of OK to have an option to disable the garbage collector -- its just that doing so will lead to a superficially ok, but generally broken config, so should be discouraged.

dmakarov:
I did. I wasted enough time trying to make it work. The language server uses only the p4c parser, so memory consumption is not huge.

dmakarov:
Out of curiosity, have p4c developers considered migrating to using smart pointers in the longer run?

mbudiu-vmw:
We chose GC to keep the code as simple as possible.
We don't have the resources to port the entire codebase to smart pointers.

hanw:
IIRC, Chris said using smart pointer resulting in 10x runtime overhead comparing to GC when he tried it initially.

ChrisDodd:
We started with smart pointers, but they caused compile times (with gcc) that were 10x slower, and were a lot more work, as well as making the code hard to read, so we stripped them out and moved to garbage collection.

dmakarov:
@mbudiu-vmw I added a note about the option to the README. Please, take a look.

PR #1368, opened 2018-06-27
Fix the userspace ebpf map, integrate control plane initializer.

fruffy:
This request fixes the control plane initializer and fixes several issues in the registry. It does not add new features.
- Three new .stf files have been added. action_call.stf which does not use a table, and key_ebpf.stf and valid_ebpf.stf which rely on table initialization.
- key_ebpf has been added to the xfail tests, as it requires explicit dynamic table updates via the control plane.
- Static control plane initialization is now fully functional 
- ebpf_registry has been slightly rewritten and now acts as a wrapper of ebpf_map.c. Maps are only modified via ebpf_registry now.
- Fixed a bug caused by pcap_next_ex(). This function seems to exhaust the stack somehow, causing the previously initialized hash map to be dropped. Moving the relevant code section into its own function fixed the issue for now.
- The user Makefile might benefit from automatic dependency generation, right now it is hardcoded.
- Fixed the delete operations of the ebpf user map. They now free all allocated values.
- Added an option handler to the runtime. It currently supports debug options and requires a file input.
- Switched from a simple dummy interface to a central virtual bridge. The bridge allows attachment of several interfaces. Any traffic will have to pass through the bridge, which is loaded with an ebpf program.


backends/ebpf/runtime/ebpf_map.c
```diff
@@ -43,36 +43,36 @@ static int check_flags(void *elem, unsigned long long map_flags) {
     return EXIT_SUCCESS;
 }
 
-void *bpf_map_lookup_elem(struct bpf_map *map, void *key, unsigned int key_size) {
+void *bpf_map_lookup_elem(struct bpf_map **map, void *key, unsigned int key_size) {
```
mbudiu-vmw: Why do you need a double pointer?
You never modify the value of map.
Perhaps you can also mark the pointers as const.

backends/ebpf/runtime/ebpf_registry.c
```diff
@@ -27,121 +27,123 @@ Implementation of ebpf registry. Intended to provide a common access interface b
...
-        fprintf(stderr, "Error: Table %s already exists!\n", map->name);
+        fprintf(stderr, "Error: Table %s already exists!\n", table->name);
         return EXIT_FAILURE;
     }
-
-    // Check key maximum length
-    if (strlen(map->name) > MAX_TABLE_NAME_LENGTH){
-        fprintf(stderr, "Error: Key name %s exceeds maximum size %d", map->name, MAX_TABLE_NAME_LENGTH);
+    /* Check key maximum length */
+    if (strlen(table->name) > MAX_TABLE_NAME_LENGTH){
```
mbudiu-vmw: missing a space

backends/ebpf/runtime/ebpf_runtime.c
```diff
@@ -33,13 +33,27 @@ void usage(char *name) {
            " into a filter function, and returns the output.\n" , name);
 }
 
+int pcap_while_loop(pcap_t *in_handle, pcap_dumper_t *out_handle) {
```
mbudiu-vmw: The name of this function is not very good. It should be called something like "process_packets".

backends/ebpf/runtime/ebpf_runtime.c
```diff
@@ -33,13 +33,27 @@ void usage(char *name) {
...
+    int ret;
+    while ((ret = pcap_next_ex(in_handle, &pcap_hdr, &packet)) == 1) {
+        /* Parse each packet in the file and check the result */
+        struct sk_buff skb;
+        skb.data = (void *) packet;
+        skb.len = pcap_hdr->len;
+        int result = ebpf_filter(&skb);
+        if (result)
+            pcap_dump((unsigned char *) out_handle, pcap_hdr, packet);
+         printf("\nResult of the eBPF parsing is: %d\n", result);
```
mbudiu-vmw: Please guard debugging messages with a flag. If (options.debug) or something similar.

backends/ebpf/runtime/ebpf_user.h
```diff
@@ -120,29 +123,28 @@ struct sk_buff {
...
-struct bpf_map_def tables[] = {
+struct bpf_table tables[] = {
 #define REGISTER_TABLE(NAME, TYPE, KEY_SIZE, VALUE_SIZE, MAX_ENTRIES) \
-    { #NAME, TYPE, KEY_SIZE, VALUE_SIZE, MAX_ENTRIES },
+    { MAP_PATH"/"#NAME, TYPE, KEY_SIZE, VALUE_SIZE, MAX_ENTRIES, NULL },
 #define REGISTER_END() \
     { 0, 0, 0, 0, 0 } \
 };
 
+/* TODO: We do not check for NULL output, which causes a segmentation fault.
```
mbudiu-vmw: I don't understand this comment. The lookup function seems to check for nulls.

Issue #1366, opened 2018-06-25
EBPF Tests: C-Code generated from two_ebpf.p4 does not compile


fruffy:
Variable naming is slightly off in the two_ebpf test case. The compiler produces a variable name pass_0 which does not exist.
[two_output.zip](https://github.com/p4lang/p4c/files/2134682/two_output.zip)

PR #1373, opened 2018-06-27
Fix for issue #1366

mbudiu-vmw:


mbudiu-vmw:
These PR are tiny, so it would not take much to review them...

Issue #1365, opened 2018-06-25
EBPF Tests: C-Code generated from hit_ebpf.p4 does not compile


fruffy:
The compiler is not able to handle .hit structures. The struct is simply left in the code.
Attached are all relevant files.

[hit_outputs.zip](https://github.com/p4lang/p4c/files/2134669/hit_outputs.zip)

PR #1374, opened 2018-06-28
Fix for issue #1365

mbudiu-vmw:


mbudiu-vmw:
@hanw @ChrisDodd @antoninbas @cc10512 if either of you could check these two tiny PR we could then merge them.
Thank you.

PR #1377, opened 2018-06-28
Fix non-unified build

antoninbas:
- fix use of static variables across translation units
- fix include statements

### DIFF
--- a/tools/ir-generator/irclass.cpp
+++ b/tools/ir-generator/irclass.cpp
@@ -19,23 +19,49 @@ limitations under the License.
 #include "lib/enumerator.h"
 
 const char* IrClass::indent = "    ";
-IrNamespace IrNamespace::global(nullptr, nullptr);
+IrNamespace& IrNamespace::global() {
+    static IrNamespace irn(nullptr, nullptr);
+    return irn;
+}
 static const LookupScope utilScope(nullptr, "Util");
 static const NamedType srcInfoType(Util::SourceInfo(), &utilScope, "SourceInfo");
-IrField *IrField::srcInfoField = new IrField(Util::SourceInfo(), &srcInfoType, "srcInfo", nullptr,
-                                             IrField::Inline | IrField::Optional);
-IrClass *IrClass::ideclaration = new IrClass(NodeKind::Interface, "IDeclaration");
-IrClass *IrClass::nodeClass = new IrClass(NodeKind::Abstract, "Node", {IrField::srcInfoField});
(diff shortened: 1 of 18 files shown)


antoninbas:
@teverman this will fix your issue
In addition to the static variable initialization issue, I fixed a bunch of missing includes. Note that there is currently a build issue with the ebpf backend, but I don't think you use it anyway.

tools/ir-generator/irclass.cpp
```diff
@@ -19,31 +19,57 @@ limitations under the License.
...
-IrField *IrField::srcInfoField = new IrField(Util::SourceInfo(), &srcInfoType, "srcInfo", nullptr,
-                                             IrField::Inline | IrField::Optional);
-IrClass *IrClass::ideclaration = new IrClass(NodeKind::Interface, "IDeclaration");
-IrClass *IrClass::nodeClass = new IrClass(NodeKind::Abstract, "Node", {IrField::srcInfoField});
-IrClass *IrClass::vectorClass = new IrClass(NodeKind::Template, "Vector");
-IrClass *IrClass::indexedVectorClass = new IrClass(NodeKind::Template, "IndexedVector");
-IrClass *IrClass::namemapClass = new IrClass(NodeKind::Template, "NameMap");
-IrClass *IrClass::nodemapClass = new IrClass(NodeKind::Template, "NodeMap");
+
+IrField* IrField::srcInfoField() {
```
ChrisDodd: This is overkill -- just need to make the static member an `IrField` instead of an `IrField *` and use the address of it everywhere, as other globals just need references to it in their ctor, and the object need not be actually initialized at that time.

antoninbas: I don't think this is overkill, unless you have a way to enforce that people writing new code never access the object directly and only use its address. IMO all static objects with dynamic initialization should be accessed through a pattern like this one to ensure that the object is initialized on first use. It can become quite difficult to keep track of this but adopting that pattern means you are probably not going to have an issue in the first place.

teverman: I'm not so sure about this.  In the crash I captured, the input to the NamedType constructor was nullptr.  If I look at the constructor implementation, it dereferences the input pointer when constructing a new LookupScope, so it still seems vulnerable to an indeterminate order of initialization.

ChrisDodd: You've convinced me -- it needs the full overkill treatment.

PR #1382, opened 2018-07-03
Fix .test script generation to include args

ChrisDodd:
The arguments for a test should go in the .test script so it can run directly, rather than just from ctest.


hanw:
Should we merge this?

cc10512:
Yes.

PR #1383, opened 2018-07-03
Implement multi-interface parsing in the runtime.

fruffy:
This pull request implements the parsing of multiple pcap files in the runtime. The runtime expects a file pattern separated by a delimiter (here it's *) and the number of expected files, and reconstructs the inputs. The input pcap files correspond to an input interface. The runtime opens an input and output pcap handle for each of the files, reads all the packets, and creates a packet (with metadata) list per file. The individual packet lists are then merged into a master list using quicksort and subsequently fed into the runtime interface. For now, only the userspace ebpf interface is specified. The kernel ebpf testing will follow in the next pull request.

There are several functions which are intended to change per target. These are feed_packets() and the initialization of the pcap handles. The decision where to place these functions is a ToDo.

### DIFF
--- a/backends/ebpf/ebpfstf.py
+++ b/backends/ebpf/ebpfstf.py
@@ -82,7 +82,7 @@ def create_table_file(actions, tmpdir, file_name):
     the control plane. """
     with open(tmpdir + "/" + file_name, "w+") as control_file:
         control_file.write("#include \"test.h\"\n\n")
-        control_file.write("static inline void generated_init() {\n\t")
+        control_file.write("static inline void run_generated_cmds() {\n\t")
         control_file.write("int ok;\n\t")
         generated_cmds = _generate_control_actions(actions)
         control_file.write(generated_cmds)
--- a/backends/ebpf/ebpftargets.py
+++ b/backends/ebpf/ebpftargets.py
@@ -338,20 +338,18 @@ class EBPFTestTarget(EBPFTarget):
         report_output(self.outputs["stdout"],
                       self.options.verbose, "Running model")
         direction = "in"
-        file = self.filename('*', direction)
+        pcap_pattern = self.filename('*', direction)
         num_files = len(glob(self.filename('*', direction)))
-
-        print (file)
+        report_output(self.outputs["stdout"],
+                      self.options.verbose,
+                      "Input file: %s" % pcap_pattern)
         # Main executable
         args = [self.template]
         # Input
-        args.extend(["-f", file])
+        args.extend(["-f", pcap_pattern])
         args.extend(["-n", str(num_files)])
         # Debug flag
         args.append("-d")
         errmsg = "Failed to execute the filter:"
-        result = run_timeout(self.options, args,
-                             TIMEOUT, self.outputs, errmsg)
-        if result != SUCCESS:
-            return FAILURE
-        return SUCCESS
+        return run_timeout(self.options, args,
+                           TIMEOUT, self.outputs, errmsg)
--- a/backends/ebpf/ebpftargets.py
+++ b/backends/ebpf/ebpftargets.py
@@ -33,6 +33,7 @@ from ebpfstf import create_table_file, parse_stf_file
 sys.path.insert(0, os.path.dirname(__file__) + '/../../tools')
 from testutils import *
 
+
 class EBPFFactory(object):
     """ Generator class.
      Returns a target subclass based on the provided target option."""
@@ -329,6 +330,7 @@ class EBPFTestTarget(EBPFTarget):
         args = self.get_make_args(self.ebpfdir, self.options.target)
         # List of bpf programs to attach to the interface
         args.append("BPFOBJ=" + self.template)
+        args.append("CFLAGS+=-DCONTROL_PLANE")
         errmsg = "Failed to build the filter:"
         return run_timeout(self.options, args, TIMEOUT, self.outputs, errmsg)
 
@@ -336,16 +338,20 @@ class EBPFTestTarget(EBPFTarget):
         report_output(self.outputs["stdout"],
                       self.options.verbose, "Running model")
         direction = "in"
-        for file in glob(self.filename('*', direction)):
-            # Main executable
-            args = [self.template]
-            # Input
-            args.extend(["-f", file])
-            # Debug flag
-            args.append("-d")
-            errmsg = "Failed to execute the filter:"
-            result = run_timeout(self.options, args,
-                                 TIMEOUT, self.outputs, errmsg)
-            if result != SUCCESS:
-                return FAILURE
+        file = self.filename('*', direction)
+        num_files = len(glob(self.filename('*', direction)))
+
+        print (file)
+        # Main executable
+        args = [self.template]
+        # Input
+        args.extend(["-f", file])
+        args.extend(["-n", str(num_files)])
+        # Debug flag
+        args.append("-d")
+        errmsg = "Failed to execute the filter:"
+        result = run_timeout(self.options, args,
+                             TIMEOUT, self.outputs, errmsg)
+        if result != SUCCESS:
+            return FAILURE
         return SUCCESS
--- a/backends/ebpf/runtime/ebpf_runtime.c
+++ b/backends/ebpf/runtime/ebpf_runtime.c
@@ -14,89 +14,270 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-/* The runtime emulates a loaded eBPF program and provides an interface to feed
-   packets to the emulated eBPF filter. */
+/* The runtime emulates a loaded eBPF program. It parses a set of given pcap files
+ and exposes an interface to feed packets to the emulated eBPF filter. */
 
-
-
-#include <string.h>
-#include <libgen.h>
+#include <unistd.h>     // getopt
+#include <ctype.h>      // isprint
 #define PCAP_DONT_INCLUDE_PCAP_BPF_H
 #include <pcap/pcap.h>
 #include "test.h"
+#ifdef CONTROL_PLANE
 #include "control.h"
-#include <unistd.h>
-#include <ctype.h>
+#endif
+
 #define FILE_NAME_MAX 256
+#define PCAPIN  "_in.pcap"
+#define PCAPOUT "_out.pcap"
+#define DELIM   "*"
 
 static int debug = 0;
 
 void usage(char *name) {
-    fprintf(stderr, "This program parses a pcap file, "
-            "feeds the individual packets into a filter function, "
+    fprintf(stderr, "This program expects a pcap file pattern, "
+            "extracts all the packets out of the matched files,"
+            "then feeds the individual packets into a filter function, "
             "and returns the output.\n");
-    fprintf(stderr, "Usage: %s [-d] -f file.pcap\n", name);
+    fprintf(stderr, "Usage: %s [-d] -f file.pcap [-n num_files]\n", name);
     fprintf(stderr, "Options:\n");
     fprintf(stderr, "\t-d: Turn on debug messages\n");
     fprintf(stderr, "\t-f: The input pcap file\n");
+    fprintf(stderr, "\t-n: Specifies the number of input files\n");
     exit(EXIT_FAILURE);
 }
 
-static int process_packets(pcap_t *in_handle, pcap_dumper_t *out_handle) {
-    struct pcap_pkthdr *pcap_hdr;
-    const unsigned char *packet;
-    int ret;
-    while ((ret = pcap_next_ex(in_handle, &pcap_hdr, &packet)) == 1) {
+/* Struct containing packet content and its metadata.
+ For now this is only the pcap header and the input interface.
+ */
+typedef struct {
+    char *data;
+    struct pcap_pkthdr pcap_hdr;
+    uint32_t ifindex;
+} pcap_pkt;
+
+/* Carries a list of packets and the total length as information.
+ Used as input stream for the packet parsing interface.
+ */
+typedef struct {
+    pcap_pkt *list;
+    uint64_t len;
+} pcap_pkt_list;
+
+/* Rank packets based on the timestamp of the pcap header */
+int compare(const void *s1, const void *s2) {
+  pcap_pkt *p1 = (pcap_pkt *)s1;
+  pcap_pkt *p2 = (pcap_pkt *)s2;
+  return p1->pcap_hdr.ts.tv_usec > p2->pcap_hdr.ts.tv_usec;
+}
+
+static pcap_pkt_list *sort_packets(pcap_pkt_list *pkt_lists, int num_lists) {
+    /* Get the total length of all individual lists combined first */
+    uint64_t total_len = 0;
+    for (int i = 0; i < num_lists; i++)
+        total_len+= pkt_lists[i].len;
+    /* Allocate the "master" list */
+    pcap_pkt_list *sorted_list = malloc(sizeof(pcap_pkt_list));
+    pcap_pkt *full_pkts = malloc(sizeof(pcap_pkt) * total_len);
+
+    /* Fill the master list by copying over the individual packets */
+    uint64_t offset = 0;
+    for (int i = 0; i < num_lists; i++) {
+        memcpy(&full_pkts[offset], pkt_lists[i].list, sizeof(pcap_pkt) *pkt_lists[i].len);
+        offset += pkt_lists[i].len;
+        /* We do not need the temporary allocated list anymore */
+        free(pkt_lists[i].list);
+    }
+    /* Sort the master list and return the meta information*/
+    qsort(full_pkts, total_len, sizeof(pcap_pkt), compare);
+    sorted_list->list = full_pkts;
+    sorted_list->len = total_len;
+    return sorted_list;
+}
+
+static void fill_packets(pcap_pkt_list *pkt_list, int num_files, pcap_t **in_handles) {
+    /* Extract all pcap packets per output file */
+    for (int i = 0; i < num_files; i++) {
+        struct pcap_pkthdr *pcap_hdr;
+        const unsigned char *packet;
+        char errbuf[PCAP_ERRBUF_SIZE];
+        int ret;
+        FILE *pcap_in_fp = pcap_file(in_handles[i]);
+        uint64_t pkt_index = 0;
+        /* Count packets in the pcap file for allocation*/
+        while ((ret = pcap_next_ex(in_handles[i], &pcap_hdr, &packet)) == 1)
+            pkt_index++;
+        if (ret == -1) {
+            pcap_perror(in_handles[i], "Error: Failed to parse ");
+            exit(EXIT_FAILURE);
+        }
+        /* Reset the file pointer and reopen the pcap handle */
+        fseek(pcap_in_fp, 0, 0);
+        in_handles[i] = pcap_fopen_offline(pcap_in_fp, errbuf);
+        if (in_handles[i] == NULL) {
+            fprintf(stderr, "Error: Failed to reopen pcap file!\n");
+            exit(EXIT_FAILURE);
+        }
+        /* Allocate a list of packets and reset the index */
+        pcap_pkt *pkts = malloc(sizeof(pcap_pkt) * pkt_index);
+        pkt_index = 0;
+        /* Fill the packet list while incrementing the counter */
+        while ((ret = pcap_next_ex(in_handles[i], &pcap_hdr, &packet)) == 1) {
+            /* Save the data we extracted from the pcap buffer */
+            pkts[pkt_index].data = malloc(pcap_hdr->len);
+            memcpy(pkts[pkt_index].data, packet, pcap_hdr->len);
+            /* Also save the header and the interface "index" */
+            pkts[pkt_index].pcap_hdr = *pcap_hdr;
+            pkts[pkt_index].ifindex = i;
+            pkt_index++;
+        }
+        if (ret == -1) {
+            pcap_perror(in_handles[i], "Error: Failed to parse ");
+            exit(EXIT_FAILURE);
+        }
+        /* Store the pointer to the list in the packet container */
+        pkt_list[i].list = pkts;
+        pkt_list[i].len = pkt_index;
+    }
+}
+
+static void free_pkt_list(pcap_pkt_list *pkt_list) {
+    for (uint64_t i = 0; i < pkt_list->len; i++) {
+        free(pkt_list->list[i].data);
+    }
+    free(pkt_list->list);
+    free(pkt_list);
+}
+
+static pcap_pkt_list *get_sorted_pkts(pcap_t **in_handles, int num_files) {
+    pcap_pkt_list pcap_pkt_lists[num_files];
+    /* Fill the packet lists with packets from the pcap files */
+    fill_packets(pcap_pkt_lists, num_files, in_handles);
+    /* Merge, sort, and return the list of the parsed packets */
+    return sort_packets(pcap_pkt_lists, num_files);
+}
+
+static void feed_packets(pcap_dumper_t **out_handles, pcap_pkt_list *pkt_list, int num_pcaps) {
+    for (uint64_t i = 0; i < pkt_list->len; i++) {
+        struct pcap_pkthdr *pcap_hdr = &pkt_list->list[i].pcap_hdr;
+        pcap_dumper_t *out_handle = out_handles[pkt_list->list[i].ifindex];
         /* Parse each packet in the file and check the result */
         struct sk_buff skb;
-        skb.data = (void *) packet;
-        skb.len = pcap_hdr->len;
+        skb.data = (void *) pkt_list->list[i].data;
+        skb.len = pkt_list->list[i].pcap_hdr.len;
         int result = ebpf_filter(&skb);
         if (result)
-            pcap_dump((unsigned char *) out_handle, pcap_hdr, packet);
+            pcap_dump((unsigned char *)out_handle, pcap_hdr, skb.data);
         if (debug)
             printf("\nResult of the eBPF parsing is: %d\n", result);
     }
-    return ret;
+}
+
+static void init_output_handles(char *first, char *second, pcap_t **in_handles, pcap_dumper_t **out_handles, int num_files) {
+    for (int i = 0; i < num_files; i++) {
+        char pcap_out_name[FILE_NAME_MAX];
+        memset(pcap_out_name, 0, FILE_NAME_MAX);
+        if (second !=NULL) {
+            snprintf(pcap_out_name, FILE_NAME_MAX, "%s%d"PCAPOUT, first, i);
+            /* We actually matched, create the filename */
+        } else {
+            /* No delimiter, try to create the file by replacing the input pattern */
+            char *replace = strstr(first, PCAPIN);
+            int offset =(replace - first);
+            memcpy(pcap_out_name, first, offset);
+            memcpy(&pcap_out_name[offset], PCAPOUT, strlen(PCAPOUT));
+        }
+        if (debug)
+            printf("Opening output file: %s\n", pcap_out_name);
+        /* Open pcap output file and store its pointer in the handles */
+        out_handles[i] = pcap_dump_open(in_handles[i], pcap_out_name);
+        if (out_handles[i] == NULL) {
+            pcap_perror(in_handles[i], "Error: Failed to create pcap output file ");
+            pcap_close(in_handles[i]);
+            exit(EXIT_FAILURE);
+        }
+    }
+}
+
+static void init_input_handles(char *first, char *second, pcap_t **in_handles, int num_files) {
+    char errbuf[PCAP_ERRBUF_SIZE];
+    for (int i = 0; i < num_files; i++) {
+        char pcap_in_name[FILE_NAME_MAX];
+        memset(pcap_in_name, 0, FILE_NAME_MAX);
+        if (second !=NULL)
+            /* We actually matched, create the filename */
+            snprintf(pcap_in_name, FILE_NAME_MAX, "%s%d"PCAPIN, first, i);
+        else
+            /* Just copy the filename into the buffer */
+            memcpy(pcap_in_name, first, FILE_NAME_MAX);
+        if (debug)
+            printf("Processing input file: %s\n", pcap_in_name);
+        if (access( pcap_in_name, F_OK) == -1) {
+            fprintf(stderr,"File %s does not exist!\n",  pcap_in_name);
+            exit(EXIT_FAILURE);
+        }
+        /* Open pcap input file and store its pointer in the handles */
+        in_handles[i] = pcap_open_offline(pcap_in_name, errbuf);
+        if (in_handles[i] == NULL) {
+            fprintf(stderr, "Error: Failed to read pcap file %s \n", errbuf);
+            exit(EXIT_FAILURE);
+        }
+    }
+}
+
+static void init_pcap_handles(char *pcap_base, pcap_t **in_handles, pcap_dumper_t **out_handles, int num_files) {
+    /* Store first half until DELIM */
+    char *first = strtok(pcap_base, DELIM);
+    /* Use the rest of the string to check if we actually matched */
+    char *second = strtok(NULL, DELIM);
+    /* Init input handles */
+    init_input_handles(first, second, in_handles, num_files);
+    /* Init output handles */
+    init_output_handles(first, second, in_handles, out_handles, num_files);
+}
+
+static void close_pcap_handles(pcap_t **in_handles, pcap_dumper_t **out_handles, int num_files) {
+    for (int i = 0; i < num_files; i++) {
+        pcap_close(in_handles[i]);
+        pcap_dump_close(out_handles[i]);
+    }
 }
 
 int main(int argc, char **argv) {
-    pcap_t *in_handle;
-    pcap_dumper_t *out_handle;
-    char errbuf[PCAP_ERRBUF_SIZE];
-    char *input_pcap = NULL;
+    char *base_pcap = NULL;
+    int num_pcaps = 0;
     int c;
-
     opterr = 0;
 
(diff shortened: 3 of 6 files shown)


backends/ebpf/ebpfstf.py
```diff
@@ -82,7 +82,7 @@ def create_table_file(actions, tmpdir, file_name):
     the control plane. """
     with open(tmpdir + "/" + file_name, "w+") as control_file:
         control_file.write("#include \"test.h\"\n\n")
-        control_file.write("static inline void generated_init() {\n\t")
+        control_file.write("static inline void run_generated_cmds() {\n\t")
```
mbudiu-vmw: I think we should name this function "initialize_tables".
There are no commands really at this level.

backends/ebpf/ebpftargets.py
```diff
@@ -329,23 +330,26 @@ def create_filter(self):
...
-            args.extend(["-f", file])
-            # Debug flag
-            args.append("-d")
-            errmsg = "Failed to execute the filter:"
-            result = run_timeout(self.options, args,
-                                 TIMEOUT, self.outputs, errmsg)
-            if result != SUCCESS:
-                return FAILURE
-        return SUCCESS
+        pcap_pattern = self.filename('*', direction)
```
mbudiu-vmw: this can be simplified - just find the last occurrence of the underscore

backends/ebpf/runtime/ebpf_runtime.c
```diff
@@ -14,89 +14,270 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-/* The runtime emulates a loaded eBPF program and provides an interface to feed
-   packets to the emulated eBPF filter. */
+/* The runtime emulates a loaded eBPF program. It parses a set of given pcap files
```
mbudiu-vmw: The runtime in fact executes a C program that processes packets.
The C program has a fixed interface: it takes an "skbuf" and returns a boolean for each packet.

backends/ebpf/runtime/ebpf_runtime.c
```diff
@@ -14,89 +14,270 @@ See the License for the specific language governing permissions and
...
+#ifdef CONTROL_PLANE
 #include "control.h"
-#include <unistd.h>
-#include <ctype.h>
+#endif
+
 #define FILE_NAME_MAX 256
+#define PCAPIN  "_in.pcap"
+#define PCAPOUT "_out.pcap"
+#define DELIM   "*"
```
mbudiu-vmw: You don't need these patterns

backends/ebpf/runtime/ebpf_runtime.c
```diff
@@ -14,89 +14,270 @@ See the License for the specific language governing permissions and
...
 
 static int debug = 0;
 
 void usage(char *name) {
-    fprintf(stderr, "This program parses a pcap file, "
-            "feeds the individual packets into a filter function, "
+    fprintf(stderr, "This program expects a pcap file pattern, "
+            "extracts all the packets out of the matched files,"
```
mbudiu-vmw: in the order given by the packet time

backends/ebpf/runtime/ebpf_runtime.c
```diff
@@ -14,89 +14,270 @@ See the License for the specific language governing permissions and
...
+ Used as input stream for the packet parsing interface.
+ */
+typedef struct {
+    pcap_pkt *list;
+    uint64_t len;
+} pcap_pkt_list;
+
+/* Rank packets based on the timestamp of the pcap header */
+int compare(const void *s1, const void *s2) {
+  pcap_pkt *p1 = (pcap_pkt *)s1;
```
mbudiu-vmw: use consistent indentation

backends/ebpf/runtime/ebpf_runtime.c
```diff
@@ -14,89 +14,270 @@ See the License for the specific language governing permissions and
...
+  pcap_pkt *p2 = (pcap_pkt *)s2;
+  return p1->pcap_hdr.ts.tv_usec > p2->pcap_hdr.ts.tv_usec;
```
mbudiu-vmw: Here you have to return the difference, not greater-than. 
The result has to be 0, 1 or -1. With your implementation it is never -1.

backends/ebpf/runtime/ebpf_runtime.c
```diff
@@ -14,89 +14,270 @@ See the License for the specific language governing permissions and
...
+/* Carries a list of packets and the total length as information.
+ Used as input stream for the packet parsing interface.
+ */
+typedef struct {
+    pcap_pkt *list;
+    uint64_t len;
+} pcap_pkt_list;
+
+/* Rank packets based on the timestamp of the pcap header */
+int compare(const void *s1, const void *s2) {
```
mbudiu-vmw: Call this function: comparePacketsByTime or something similar.

backends/ebpf/runtime/ebpf_runtime.c
```diff
@@ -14,89 +14,270 @@ See the License for the specific language governing permissions and
...
+  pcap_pkt *p1 = (pcap_pkt *)s1;
+  pcap_pkt *p2 = (pcap_pkt *)s2;
+  return p1->pcap_hdr.ts.tv_usec > p2->pcap_hdr.ts.tv_usec;
+}
+
+static pcap_pkt_list *sort_packets(pcap_pkt_list *pkt_lists, int num_lists) {
+    /* Get the total length of all individual lists combined first */
+    uint64_t total_len = 0;
```
mbudiu-vmw: I don't think you need a 64-bit int for the list length

backends/ebpf/runtime/ebpf_runtime.c
```diff
@@ -14,89 +14,270 @@ See the License for the specific language governing permissions and
...
+    for (int i = 0; i < num_lists; i++)
+        total_len+= pkt_lists[i].len;
```
mbudiu-vmw: use consistent spacing

backends/ebpf/runtime/ebpf_runtime.c
```diff
@@ -14,89 +14,270 @@ See the License for the specific language governing permissions and
...
+    pcap_pkt_list *sorted_list = malloc(sizeof(pcap_pkt_list));
+    pcap_pkt *full_pkts = malloc(sizeof(pcap_pkt) * total_len);
+
+    /* Fill the master list by copying over the individual packets */
+    uint64_t offset = 0;
+    for (int i = 0; i < num_lists; i++) {
+        memcpy(&full_pkts[offset], pkt_lists[i].list, sizeof(pcap_pkt) *pkt_lists[i].len);
+        offset += pkt_lists[i].len;
+        /* We do not need the temporary allocated list anymore */
+        free(pkt_lists[i].list);
```
mbudiu-vmw: In general you should delete something in the same place that allocated it, otherwise it is very confusing.
If a function deletes one of its arguments it should be clearly documented. But try to avoid doing this.

backends/ebpf/runtime/ebpf_runtime.c
```diff
@@ -14,89 +14,270 @@ See the License for the specific language governing permissions and
...
+}
+
+static void fill_packets(pcap_pkt_list *pkt_list, int num_files, pcap_t **in_handles) {
+    /* Extract all pcap packets per output file */
+    for (int i = 0; i < num_files; i++) {
+        struct pcap_pkthdr *pcap_hdr;
+        const unsigned char *packet;
+        char errbuf[PCAP_ERRBUF_SIZE];
+        int ret;
+        FILE *pcap_in_fp = pcap_file(in_handles[i]);
```
mbudiu-vmw: Here you are not checking for error

backends/ebpf/runtime/ebpf_runtime.c
```diff
@@ -14,89 +14,270 @@ See the License for the specific language governing permissions and
...
+        uint64_t pkt_index = 0;
+        /* Count packets in the pcap file for allocation*/
+        while ((ret = pcap_next_ex(in_handles[i], &pcap_hdr, &packet)) == 1)
```
mbudiu-vmw: This is somewhat confusing, you are accessing files both through file handles and FILE pointers. These are separate data structures and should not be mixed.

backends/ebpf/runtime/ebpf_runtime.c
```diff
@@ -14,89 +14,270 @@ See the License for the specific language governing permissions and
...
+    sorted_list->len = total_len;
+    return sorted_list;
+}
+
+static void fill_packets(pcap_pkt_list *pkt_list, int num_files, pcap_t **in_handles) {
+    /* Extract all pcap packets per output file */
+    for (int i = 0; i < num_files; i++) {
+        struct pcap_pkthdr *pcap_hdr;
+        const unsigned char *packet;
+        char errbuf[PCAP_ERRBUF_SIZE];
```
mbudiu-vmw: You can probably lift this allocation out of the loop.

backends/ebpf/runtime/ebpf_runtime.c
```diff
@@ -14,89 +14,270 @@ See the License for the specific language governing permissions and
...
+typedef struct {
+    char *data;
+    struct pcap_pkthdr pcap_hdr;
+    uint32_t ifindex;
+} pcap_pkt;
+
+/* Carries a list of packets and the total length as information.
+ Used as input stream for the packet parsing interface.
+ */
+typedef struct {
```
mbudiu-vmw: You should move all pcap-related data structures into a separate C file with an associated header.
Then you define these two structures and a set of functions:
- read file, return list
- merge lists, return list
- delete list
- split a list into sub-lists by interfaces
- write a list to a file
Files should probably be given just by name, not by handles. They should be closed when the corresponding functions return.

backends/ebpf/runtime/ebpf_runtime.c
```diff
@@ -14,89 +14,270 @@ See the License for the specific language governing permissions and
...
+        /* Count packets in the pcap file for allocation*/
+        while ((ret = pcap_next_ex(in_handles[i], &pcap_hdr, &packet)) == 1)
+            pkt_index++;
+        if (ret == -1) {
+            pcap_perror(in_handles[i], "Error: Failed to parse ");
+            exit(EXIT_FAILURE);
+        }
+        /* Reset the file pointer and reopen the pcap handle */
+        fseek(pcap_in_fp, 0, 0);
+        in_handles[i] = pcap_fopen_offline(pcap_in_fp, errbuf);
```
mbudiu-vmw: You haven't closed the file. Why do you need to reopen?
Why don't you just open offline from the start?

backends/ebpf/runtime/ebpf_runtime.c
```diff
@@ -14,89 +14,270 @@ See the License for the specific language governing permissions and
...
+static pcap_pkt_list *get_sorted_pkts(pcap_t **in_handles, int num_files) {
+    pcap_pkt_list pcap_pkt_lists[num_files];
+    /* Fill the packet lists with packets from the pcap files */
+    fill_packets(pcap_pkt_lists, num_files, in_handles);
+    /* Merge, sort, and return the list of the parsed packets */
+    return sort_packets(pcap_pkt_lists, num_files);
+}
+
+static void feed_packets(pcap_dumper_t **out_handles, pcap_pkt_list *pkt_list, int num_pcaps) {
+    for (uint64_t i = 0; i < pkt_list->len; i++) {
```
mbudiu-vmw: For symmetry I think you should consume a list and produce a list.

backends/ebpf/runtime/ebpf_runtime.c
```diff
@@ -14,89 +14,270 @@ See the License for the specific language governing permissions and
...
         if (debug)
             printf("\nResult of the eBPF parsing is: %d\n", result);
     }
-    return ret;
 }
 
-int main(int argc, char **argv) {
-    pcap_t *in_handle;
-    pcap_dumper_t *out_handle;
+static void init_output_handles(char *first, char *second, pcap_t **in_handles, pcap_dumper_t **out_handles, int num_files) {
```
mbudiu-vmw: This function will disappear: it will be replaced by a for loop that calls the function to write a list to a file.
Parts of this code will move into the write function.

backends/ebpf/runtime/pcap_util.h
```diff
@@ -0,0 +1,207 @@
...
+ * @param pkt_array List handle descriptor. Can be null
+ * @param pkt_list Pointer of the packet list to append.
+ *
+ * @return Returns the updated handle of the list array. This function causes
+ * an exit if allocation fails.
+ */
+pcap_list_array_t *append_list(pcap_list_array_t *pkt_array, pcap_list_t *pkt_list);
+
+/**
+ * @brief Allocates a packet list and returns the handle.
```
mbudiu-vmw: No need to have the `@return`, it says the same thing as the `@brief`.

backends/ebpf/runtime/pcap_util.h
```diff
@@ -0,0 +1,207 @@
...
+ *
+ * @return The handle to the packet list.
+ */
+pcap_list_t *allocate_pkt_list();
+
+
+/**
+ * @brief Allocates a packet list and returns the handle.
```
mbudiu-vmw: This one is copied from above and is wrong.
Also, don't use "handle".

backends/ebpf/runtime/pcap_util.h
```diff
@@ -0,0 +1,207 @@
...
+ * @details Deletes the given list. Also deletes the data referenced by the
+ * pcap packets in the list.
+ *
+ * @param pkt_list Handle to a list.
+ */
+void delete_list(pcap_list_t *pkt_list);
+
+/**
+ * @brief Completely erases a list array and its lists.
+ * @details Deletes the given array. Also deletes the list as well as the
```
mbudiu-vmw: Deletes all lists in the array.

backends/ebpf/runtime/ebpf_runtime.c
```diff
@@ -14,89 +14,164 @@ See the License for the specific language governing permissions and
...
 void usage(char *name) {
-    fprintf(stderr, "This program parses a pcap file, "
-            "feeds the individual packets into a filter function, "
+    fprintf(stderr, "This program expects a pcap file pattern, "
+            "extracts all the packets out of the matched files"
+            "in the order given by the packet time,"
+            "then feeds the individual packets into a filter function, "
             "and returns the output.\n");
-    fprintf(stderr, "Usage: %s [-d] -f file.pcap\n", name);
+    fprintf(stderr, "Usage: %s [-d] -f file.pcap [-n num_files]\n", name);
```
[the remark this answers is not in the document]
fruffy: Well, it defaults to zero if no explicit number is specified.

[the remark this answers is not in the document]
mbudiu-vmw: But zero does not make much sense, so it should not be optional.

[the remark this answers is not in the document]
fruffy: So the intention was to test the runtime also for programs who do not have any pcap or stf files associated with it. But specifying -n 0 works just as well so I can make it mandatory.

backends/ebpf/runtime/ebpf_runtime.c
```diff
@@ -14,89 +14,164 @@ See the License for the specific language governing permissions and
...
+static pcap_list_t *get_packets(const char *pcap_base, uint16_t num_files) {
+    pcap_list_array_t *tmp_list_array = allocate_pkt_list_array();
+    /* Retrieve a list for each file and append it to the temporary array */
+    for (uint16_t i = 0; i < num_files; i++) {
+        char pcap_in_name[FILE_NAME_MAX];
+        snprintf(pcap_in_name, FILE_NAME_MAX, "%s%d%s", pcap_base, i, PCAPIN);
+        if (debug)
+            printf("Processing input file: %s\n", pcap_in_name);
+        pcap_list_t *pkt_list = read_pkts_from_pcap(pcap_in_name, i);
+        tmp_list_array = append_list(tmp_list_array, pkt_list);
```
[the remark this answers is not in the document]
fruffy: Yes, because the pointer might get updated by realloc. I could also enforce passing a double pointer instead. The list array is deleted in merge_lists(). I will rename that function to make it more explicit.

backends/ebpf/runtime/ebpf_runtime.c
```diff
@@ -14,89 +14,164 @@ See the License for the specific language governing permissions and
...
+    /* Our pcap handles, do not forget to delete them after */
+    pcap_list_t *pkt_list_handle = allocate_pkt_list();
+    pcap_list_array_t *output_array_handle = allocate_pkt_list_array();
+
+    /* Create the basic pcap filename from the input */
+    const char *suffix = strrchr(pcap_name, DELIM);
+    int baselen = suffix - pcap_name;
+    char pcap_base[baselen];
+    memcpy(pcap_base, pcap_name, baselen);
+    pcap_base[baselen] = '\0';
```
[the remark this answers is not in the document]
fruffy: https://stackoverflow.com/questions/13050987/c-random-characters-after-using-strncpy-function
I think this is the correct way to do it, also why valgrind did not complain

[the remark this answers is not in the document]
mbudiu-vmw: The correct way is to allocate one extra character and use strncpy.

backends/ebpf/runtime/pcap_util.h
```diff
@@ -0,0 +1,207 @@
...
+/**
+ * @brief Appends a reference to a packet to a given list of packets.
+ * @details This function takes a pointer to a packet and appends it to the
+ * given list. If the list does not exist, it is allocated. For each new packet
+ * to append, the list is expanded by the size of the packet pointer.
+ *
+ * @param pkt_list List handle descriptor. Can be null
+ * @param pkt Pointer of the packet to append.
+ *
+ * @return Returns the updated handle of the list. This function causes an
```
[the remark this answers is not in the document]
fruffy: I had chosen exit instead of an error code since a remalloc error is fatal for the program anyway. It makes the code a bit simpler by avoiding the checks for errors for each append call. The pointer itself changes depending on the remalloc output iirc. I can make it a double pointer argument instead.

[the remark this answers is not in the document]
mbudiu-vmw: exit is fine

backends/ebpf/runtime/ebpf_runtime.c
```diff
@@ -94,51 +86,81 @@ static void write_pkts_to_pcaps(const char *pcap_base, pcap_list_array_t *output
     uint16_t arr_len = get_list_array_length(output_array);
     for (uint16_t i = 0; i < arr_len; i++) {
         char pcap_out_name[FILE_NAME_MAX];
-        snprintf(pcap_out_name, FILE_NAME_MAX, "%s%d%s", pcap_base, i, PCAPOUT);
+        if (snprintf(pcap_out_name, FILE_NAME_MAX, "%s%d%s", pcap_base, i, PCAPOUT) >=FILE_NAME_MAX) {
+            fprintf(stderr, "Filename %s%d%s too long. Truncated.\n"
```
mbudiu-vmw: There is no need to say "truncated" since you terminate the program anyway. Just say that the filename is too long.

mbudiu-vmw: But the best practice would be to actually allocate the filename buffer dynamically to be as big as needed. Using fixed-size buffers was done in the eighties when memories were too small.

backends/ebpf/runtime/ebpf_runtime.c
```diff
@@ -14,89 +14,192 @@ See the License for the specific language governing permissions and
...
+            "extracts all the packets out of the matched files"
+            "in the order given by the packet time,"
+            "then feeds the individual packets into a filter function, "
             "and returns the output.\n");
-    fprintf(stderr, "Usage: %s [-d] -f file.pcap\n", name);
+    fprintf(stderr, "Usage: %s [-d] -f file.pcap -n num_pcaps\n", name);
     fprintf(stderr, "Options:\n");
     fprintf(stderr, "\t-d: Turn on debug messages\n");
     fprintf(stderr, "\t-f: The input pcap file\n");
+    fprintf(stderr, "\t-n: Specifies the number of input interfaces\n");
```
mbudiu-vmw: You should use consistent naming: -n interface_count if your error message is about interfaces.

backends/ebpf/runtime/ebpf_runtime.c
```diff
@@ -14,89 +14,192 @@ See the License for the specific language governing permissions and
...
 }
-
-static int process_packets(pcap_t *in_handle, pcap_dumper_t *out_handle) {
-    struct pcap_pkthdr *pcap_hdr;
-    const unsigned char *packet;
-    int ret;
-    while ((ret = pcap_next_ex(in_handle, &pcap_hdr, &packet)) == 1) {
-        /* Parse each packet in the file and check the result */
+/**
+ * @brief Feed a list packets into an eBPF program.
```
mbudiu-vmw: This function has actually nothing to do with ebpf. 
What it does is to send all the packets in the list to a function called `ebpf_filter` and to return the packets for which the filter returned 1.
In fact, you could probably pass ebpf_filter using a function pointer.
Then you should call this funciton `run_filter`.

backends/ebpf/runtime/ebpf_runtime.c
```diff
@@ -14,89 +14,192 @@ See the License for the specific language governing permissions and
...
-        skb.data = (void *) packet;
-        skb.len = pcap_hdr->len;
+        pcap_pkt *input_pkt = get_packet(pkt_list, i);
+        skb.data = (void *) input_pkt->data;
+        skb.len = input_pkt->pcap_hdr.len;
         int result = ebpf_filter(&skb);
-        if (result)
-            pcap_dump((unsigned char *) out_handle, pcap_hdr, packet);
+        if (result != 0) {
+            /* We copy the entire content to emulate an outgoing packet */
```
mbudiu-vmw: This comment is not really needed.

backends/ebpf/runtime/ebpf_runtime.c
```diff
@@ -14,89 +14,192 @@ See the License for the specific language governing permissions and
...
+            pcap_pkt *out_pkt = copy_pkt(input_pkt);
+            output_pkts = append_packet(output_pkts, out_pkt);
+        }
+        if (debug)
+            printf("Result of the eBPF parsing is: %d\n", result);
```
mbudiu-vmw: There is no parsing really, this is the result of the packet filter.

backends/ebpf/runtime/ebpf_runtime.c
```diff
@@ -14,89 +14,192 @@ See the License for the specific language governing permissions and
...
+    }
+    return output_pkts;
+}
+
+static void write_pkts_to_pcaps(const char *pcap_base, pcap_list_array_t *output_array) {
+    uint16_t arr_len = get_list_array_length(output_array);
+    for (uint16_t i = 0; i < arr_len; i++) {
+        char pcap_out_name[FILE_NAME_MAX];
+        if (snprintf(pcap_out_name, FILE_NAME_MAX, "%s%d%s", pcap_base, i, PCAPOUT) >=FILE_NAME_MAX) {
```
mbudiu-vmw: You should avoid using fixed-size buffers. The right solution is to measure the string and allocate a buffer of the right size. You should still be defensive about snprintf in case you got it wrong.

backends/ebpf/runtime/ebpf_runtime.c
```diff
@@ -14,89 +14,192 @@ See the License for the specific language governing permissions and
...
+            fprintf(stderr, "Filename %s%d%s too long. Truncated.\n"
```
mbudiu-vmw: There is no reason to say "truncated" - you are terminating the program anyway.

backends/ebpf/runtime/ebpf_runtime.c
```diff
@@ -14,89 +14,192 @@ See the License for the specific language governing permissions and
...
+    }
+}
+
+static void *run_and_record_output(const char *pcap_base, pcap_list_t *pkt_list) {
+    /* Create an array of packet lists */
+    pcap_list_array_t *output_array = NULL;
+    /* Feed the packets into our "loaded" program */
+    pcap_list_t *output_pkts = feed_packets(pkt_list);
+    /* Split the output packet list by interface. This destroys the list. */
+    output_array = delete_and_split_list(output_pkts, output_array);
```
mbudiu-vmw: I would call this `split_and_delete`. The splitting is more important than the deletion.

backends/ebpf/runtime/ebpf_runtime.c
```diff
@@ -14,89 +14,192 @@ See the License for the specific language governing permissions and
...
+    /* Delete the array, including the data it is holding */
+    delete_array(output_array);
+}
+
+static pcap_list_t *get_packets(const char *pcap_base, uint16_t num_pcaps, pcap_list_t *merged_list) {
+    pcap_list_array_t *tmp_list_array = allocate_pkt_list_array();
+    /* Retrieve a list for each file and append it to the temporary array */
+    for (uint16_t i = 0; i < num_pcaps; i++) {
+        char pcap_in_name[FILE_NAME_MAX];
+        if (snprintf(pcap_in_name, FILE_NAME_MAX, "%s%d%s", pcap_base, i, PCAPIN) >=FILE_NAME_MAX) {
```
mbudiu-vmw: Same comment as above. Perhaps you want to make a function that allocates and returns the filename.

backends/ebpf/runtime/ebpf_runtime.c
```diff
@@ -14,89 +14,192 @@ See the License for the specific language governing permissions and
...
+    pcap_list_t *output_pkts = feed_packets(pkt_list);
+    /* Split the output packet list by interface. This destroys the list. */
+    output_array = delete_and_split_list(output_pkts, output_array);
+    /* Write each list to a separate pcap output file */
+    write_pkts_to_pcaps(pcap_base, output_array);
+    /* Delete the array, including the data it is holding */
+    delete_array(output_array);
+}
+
+static pcap_list_t *get_packets(const char *pcap_base, uint16_t num_pcaps, pcap_list_t *merged_list) {
```
mbudiu-vmw: If this function deletes its input argument merged_list the name should also reflect that fact.
In general people don't expect functions to deallocate their arguments.

backends/ebpf/runtime/ebpf_runtime.c
```diff
@@ -14,89 +14,192 @@ See the License for the specific language governing permissions and
...
+void launch_runtime(const char *pcap_name, uint16_t num_pcaps) {
+    /* Initialize the list of input packets */
+    pcap_list_t *input_list = NULL;
+    if (num_pcaps == 0)
+        return;
+
+    /* Create the basic pcap filename from the input */
+    const char *suffix = strrchr(pcap_name, DELIM);
+    if (suffix == NULL) {
+        fprintf(stderr, "Delimiter not found in %s! Exiting...\n",pcap_name);
```
mbudiu-vmw: You should give a better error message: "expected a filename of the form ...".
It is not clear what delimiter you are looking for from this message.

backends/ebpf/runtime/ebpf_runtime.c
```diff
@@ -14,89 +14,192 @@ See the License for the specific language governing permissions and
...
+        exit(EXIT_FAILURE);
     }
-    return ret;
+    int baselen = suffix - pcap_name;
+    char pcap_base[baselen + 1];
+    strncpy(pcap_base, pcap_name, baselen);
+    pcap_base[baselen] = '\0';
```
mbudiu-vmw: I think that strncpy already does this.

backends/ebpf/runtime/ebpf_runtime.c
```diff
@@ -106,37 +209,23 @@ int main(int argc, char **argv) {
...
-    pcap_dump_close(out_handle);
-    free(out_file_name);
+    init_tables();
+    /* Run all commands specified in the control file */
+    setup_control_plane();
+#endif
+
+    launch_runtime(pcap_name, num_pcaps);
+
+    /* Delete all the remaining tables in the user program */
```
mbudiu-vmw: For symmetry you should move this code into a function delete_tables.
Also, this code is only needed for the user-space testing. Is this file used when doing kernel-level testing too?

backends/ebpf/runtime/pcap_util.c
```diff
@@ -0,0 +1,237 @@
...
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+#include <stdlib.h>     // EXIT_SUCCESS, EXIT_FAILURE
+#include <string.h>     // memcpy()
+#include "pcap_util.h"
+
+#define DLT_EN10MB 1
```
mbudiu-vmw: This deserves a comment.

backends/ebpf/runtime/pcap_util.c
```diff
@@ -0,0 +1,237 @@
...
+    return pkt_list;
+}
+
+pcap_list_array_t *insert_list(pcap_list_array_t *pkt_array, pcap_list_t *pkt_list, uint16_t index) {
+    if (!pkt_array)
+        /* If the array is not allocated yet, create it */
+        pkt_array = allocate_pkt_list_array();
+    if (index >= pkt_array->len) {
+        pkt_array->lists = realloc(pkt_array->lists, (index+1) * sizeof(pcap_list_t *));
+        if (pkt_array->lists == NULL) {
```
mbudiu-vmw: You should set all lists between the current size and index to null. Currently there is garbage there.

backends/ebpf/runtime/pcap_util.c
```diff
@@ -0,0 +1,237 @@
...
+    return pkt_list;
+}
+
+pcap_list_array_t *allocate_pkt_list_array() {
+    pcap_list_array_t *pkt_list_arr = calloc(1, sizeof(pcap_list_array_t));
+    return pkt_list_arr;
+}
+
+void delete_list(pcap_list_t *pkt_list) {
+    for(uint32_t i = 0; i< pkt_list->len; i++) {
```
mbudiu-vmw: spaces missing

backends/ebpf/runtime/pcap_util.c
```diff
@@ -0,0 +1,237 @@
...
+        pkt_list->pkts[i]->data = NULL;
+        free(pkt_list->pkts[i]);
+    }
+    free(pkt_list->pkts);
+    free(pkt_list);
+}
+
+void delete_array(pcap_list_array_t *pkt_list_array) {
+    for(uint32_t i = 0; i< pkt_list_array->len; i++)
+        delete_list(pkt_list_array->lists[i]);
```
mbudiu-vmw: This will crash if some lists were never allocated.
You must check for null lists either here or in delete_list.

backends/ebpf/runtime/pcap_util.c
```diff
@@ -0,0 +1,237 @@
...
+
+    uint16_t max_index = 0;
+    for (uint32_t i = 0; i < input_list->len; i++){
+        if (input_list->pkts[i]->ifindex > max_index)
+            max_index = input_list->pkts[i]->ifindex;
+    }
+
+    /* Allocate as many lists as the maximum index */
+    for (int i = 0; i <= max_index; i++) {
+        pcap_list_t *pkt_list = calloc(1, sizeof(pcap_list_t));
```
mbudiu-vmw: You should use the function you already created to allocate a list.

backends/ebpf/runtime/pcap_util.c
```diff
@@ -0,0 +1,237 @@
...
+}
+
+/* Rank packets based on the timestamp of the pcap header */
+static int compare_pkt_time(const void *s1, const void *s2) {
+  pcap_pkt *p1 = *(pcap_pkt **)s1;
+  pcap_pkt *p2 = *(pcap_pkt **)s2;
+  return p1->pcap_hdr.ts.tv_usec - p2->pcap_hdr.ts.tv_usec;
+}
+
+void sort_pcap_list(pcap_list_t *pkt_list) {
```
mbudiu-vmw: This is a very nice file. It will be a very useful library going forward.

backends/ebpf/runtime/pcap_util.h
```diff
@@ -0,0 +1,207 @@
...
+
+#ifndef BACKENDS_EBPF_RUNTIME_EBPF_PCAP_UTIL_H_
+#define BACKENDS_EBPF_RUNTIME_EBPF_PCAP_UTIL_H_
+
+#define PCAP_DONT_INCLUDE_PCAP_BPF_H
+#include <pcap/pcap.h>
+#include <stdint.h>     // uint32_t, uint16_t
+
+/* Interfaces are named by integers */
+typedef uint16_t iface_index;
```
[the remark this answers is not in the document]
mbudiu-vmw: This comment is still unaddressed.

backends/ebpf/runtime/pcap_util.c
fruffy: I decided against that since I set the value with  pkt_list->pkts[pkt_list->len - 1] = pkt in the next line anway. So it seemed superfluous

backends/ebpf/runtime/ebpf_runtime.c
fruffy: Unfortunately not, I had to learn that the hard way. I decided to use snprintf instead, which is a bit more elegant.

backends/ebpf/runtime/pcap_util.h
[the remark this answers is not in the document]
fruffy: My intention was to limit the amount of possible files to 65535 for simplicity and sanity. I do a check for that in the runtime. Pcap files can be much larger, so I leave the indexing at uint64_t to avoid overflows.

backends/ebpf/runtime/pcap_util.c
mbudiu-vmw: What if you add a packet to list number 10 and lists 0 to 9 do not exist?

PR #1385, opened 2018-07-06
Extend P4-14 frontend to support multiple entry points to parser

hanw:
This PR extends the P4-14 to P4-16 converter to support multiple entry points to parser. This feature does not comply with the P4-14 specification, as the P4-14 spec only allow one entry point to the parser, however, it is a useful feature for certain targets. 

Previously, the frontend had a hack in unusedDeclaration.cpp and simplifyParser.cpp to prevent any state with `packet_entry` pragma from being eliminated. However, this hack does not work on a state that is used and only used by the the additional entry point states. As a result, the frontend gives a 'Cannot find declaration' error when a state used by the additional entry point is eliminated. This PR fixes this bug.

The added `start` state needs to be removed by the backend. However, backend does not have to support this feature, as the feature is not in P4-14 spec.

### DIFF
--- a/frontends/p4/fromv1.0/converters.cpp
+++ b/frontends/p4/fromv1.0/converters.cpp
@@ -1081,6 +1081,8 @@ class RemoveAnnotatedFields : public Transform {
     }
 };
 
+// If a parser state has a pragma @packet_entry, it is treated as a new entry
+// point to the parser.
 class CheckIfMultiEntryPoint: public Inspector {
     ProgramStructure* structure;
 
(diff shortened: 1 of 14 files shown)


mbudiu-vmw:
Our testing scripts already try to recompile the output programs.

hanw:
In that case, I can change to `__`.

mbudiu-vmw:
The safe way is to use a name generator.
I recall that structure has a name generator.

hanw:
I need to match on these names later in the backend, a random name from name generator makes it more difficult.

mbudiu-vmw:
The name generators I wrote add a suffix. If the prefix is unique you are in a good shape.

hanw:
Fixed, I moved `$start` to nameAnnotation on the parser state. We don't support annotation on IR::Type, do we?

mbudiu-vmw:
We should support them on an enum declaration if that's what you want.
In general, they can be on declarations, not on types.

hanw:
I would like to be able to write something like this. 
``` 
@name("$InstanceType")
enum bit<32> InstanceType_0 {
       START = 32w0,
       start_e2e_mirrored = 32w1,
       start_i2e_mirrored = 32w2
}
```
But IR::Type_SerEnum constructor does not accept IR::Annotations.

mbudiu-vmw:
This is an oversight. Both enum and serenum should have annotations.
You can either add them to the IR or file an issue about it.
It may be just enough to add IAnnotated to the interface and an optional Annotations annotations field. and the getAnnotations method. See for example Type_Newtype

mbudiu-vmw:
We may be able to push the annotations to Type_Declaration, but that may require more changes in the code

PR #1396, opened 2018-07-10
Fix irgenerator dependecy in cmake files

dmakarov:
This fix makes the build system work with the Xcode generator,
which otherwise fails to produce the correct dependencies for
the custom command that builds IR generated sources.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -194,7 +194,10 @@ set (IR_GENERATED_SRCS
   ${P4C_BINARY_DIR}/ir/ir-generated.h
   ${P4C_BINARY_DIR}/ir/ir-generated.cpp
   ${P4C_BINARY_DIR}/ir/gen-tree-macro.h)
-set (IR_GENERATOR ${P4C_BINARY_DIR}/tools/ir-generator/irgenerator)
+# IR_GENERATOR_DIRECTORY is used to set the RUNTIME_OUTPUT_DIRECTORY property
+# of the irgenerator target to the matching path
+set (IR_GENERATOR_DIRECTORY ${P4C_BINARY_DIR}/tools/ir-generator)
+set (IR_GENERATOR ${IR_GENERATOR_DIRECTORY}/irgenerator)
 
 # the order of adding subdirectories matters because of target dependencies
 add_subdirectory (tools/driver)
@@ -279,7 +282,7 @@ add_custom_command (OUTPUT ${IR_GENERATED_SRCS}
   COMMAND awk -v name=gen-tree-macro.h -f ${fixup_file} ir/gen-tree-macro.h.tmp > ir/gen-tree-macro.h.fixup
   COMMAND ${CMAKE_COMMAND} -E copy_if_different ir/gen-tree-macro.h.fixup ir/gen-tree-macro.h
   MAIN_DEPENDENCY ${IR_GENERATOR}
-  DEPENDS ${IR_DEF_FILES}
+  DEPENDS irgenerator ${IR_DEF_FILES}
   COMMENT "Generating IR class files")
 
 add_custom_target(genIR DEPENDS ${IR_GENERATED_SRCS})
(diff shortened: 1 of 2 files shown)


CMakeLists.txt
```diff
@@ -194,7 +194,10 @@ set (IR_GENERATED_SRCS
   ${P4C_BINARY_DIR}/ir/ir-generated.h
   ${P4C_BINARY_DIR}/ir/ir-generated.cpp
   ${P4C_BINARY_DIR}/ir/gen-tree-macro.h)
-set (IR_GENERATOR ${P4C_BINARY_DIR}/tools/ir-generator/irgenerator)
+# IR_GENERATOR_DIRECTORY is used to set the RUNTIME_OUTPUT_DIRECTORY property
+# of the irgenerator target to the matching path
+set (IR_GENERATOR_DIRECTORY ${P4C_BINARY_DIR}/tools/ir-generator)
```
ChrisDodd: This seems wrong -- the ir generator puts its generated files in  `${P4C_BINARY_DIR}/ir`.  The `${P4C_BINARY_DIR}/tools/ir-generator` is used for the lex&yacc output intermediate source files of the ir-generator.

dmakarov: This path is where the `irgenerator` _executable_ is placed when it's built by a "non-IDE" build tool (make, or ninja).  When the build tool is `xcodebuild`, another subdirectory (Debug or Release) is normally appended to the path, depending on the build configuration.  Since the path is explicitly used by the custom command that runs `irgenerator` ( https://github.com/p4lang/p4c/blob/39deb73c51a1eaec8cbebcbb20f7aa2b69cc442f/CMakeLists.txt#L277-L283 ), it doesn't work when the cmake generator is Xcode (I assume it shouldn't work with Visual Studio either).  The change fixes the issue for Xcode cmake generator, but doesn't change anything for other generators (Unix Makefile or Ninja).  Whether it's wrong that `irgenerator` executable is placed in `${P4C_BINARY_DIR}/tools/ir-generator` or not -- I don't know -- this is how your `CMakeLists.txt` file has been for a while.

ChrisDodd: So I guess its the comment that is confusing -- it implies that this is for the generated ir source files, not for the irgenerator binary.

tools/ir-generator/CMakeLists.txt
```diff
@@ -43,4 +43,9 @@ set_source_files_properties(${IRGENERATOR_GEN_SRCS} PROPERTIES GENERATED TRUE)
 set_source_files_properties(${IRGENERATOR_GEN_SRCS} PROPERTIES OBJECT_DEPENDS ${FLEX_IRgenLexer_OUTPUTS})
 build_unified(IRGENERATOR_SRCS ALL)
 add_executable (irgenerator ${IRGENERATOR_SRCS} ${IRGENERATOR_GEN_SRCS})
+# Unconditionally set the output directory for the irgenerator to the path
+# where the custom command that generates IR sources expects to find it.
+# The generator expression $<BOOL:1> prevents IDE generators like Xcode
+# from appending a per-configuration subdirectory to the specified path.
```
ChrisDodd: So this is not the output directory for the ir-generator, it is the binary location?

CMakeLists.txt
dmakarov: The `add_executable` cmake command uses the property `RUNTIME_OUTPUT_DIRECTORY` to determine the location of a built executable (a binary). Probably the name of the property is confusing. When I mention _output directory_ in this context I refer to the property `RUNTIME_OUTPUT_DIRECTORY` and the corresponding path where the executable (binary) is placed. This is _not_ where any output, produced by running the executable, may be placed.

tools/ir-generator/CMakeLists.txt
dmakarov: Yes, this is correct.

PR #1398, opened 2018-07-11
Ignore unknown match types (with warning) rather than crashing out.

ChrisDodd:
Targets/backends can define additional match types, so there should be some way of telling this code what to do with them.  For now though, its better to just ignore them so at least the compiler can get to the backend...


ChrisDodd:
No crash as such -- it was just exiting immediately after the error without running any of the rest of the compiler.

PR #1400, opened 2018-07-13
Generate P4info in p4test backend tests

antoninbas:
* Fix P4Info register serialization for PSA

The PSA register has 2 type parameters (while the v1model register has a
single one), so the BUG_CHECK didn't make sense.

* Generate P4Info as part of p4test backend tests

In order to improve test coverage of P4Info generation, we are now
generating it by default in p4test backend tests.
Tests can be explicitly added to P4RUNTIME_EXCLUDE if P4Info cannot be
generated even though they are valid v1model / PSA programs.
A new cmake macro was introduced to register tests which include P4Info
generation: p4c_add_tests_w_p4runtime. I didn't want to modify
p4c_add_tests to avoid breaking backends that may be using this macro to
register their own tests.

### DIFF
--- a/backends/p4test/CMakeLists.txt
+++ b/backends/p4test/CMakeLists.txt
@@ -50,11 +50,31 @@ set (P4TEST_SUITES
   "${P4C_SOURCE_DIR}/testdata/p4_16_samples/*.p4"
   "${P4C_SOURCE_DIR}/testdata/p4_16_errors/*.p4"
   )
+# Builds a list of tests for which P4Info generation is not supported. It makes
+# sense not to fail the test just because P4Info generation fails. It probably
+# would not make sense to run all the tests twice (one with and one without
+# P4Runtime).
+set (P4RUNTIME_EXCLUDE
+  # Two counter instances provided for psa_direct_counter, illegal according to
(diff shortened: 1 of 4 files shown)


backends/p4test/CMakeLists.txt
```diff
@@ -50,11 +50,31 @@ set (P4TEST_SUITES
...
+# would not make sense to run all the tests twice (one with and one without
+# P4Runtime).
+set (P4RUNTIME_EXCLUDE
+  # Two counter instances provided for psa_direct_counter, illegal according to
+  # PSA spec and fails in P4Info generation
+  testdata/p4_16_samples/psa-counter5.p4
+  # Need to convert PSA_MeterColor_t to a serializable enum
+  testdata/p4_16_samples/psa-meter1.p4
+  # Compiler Bug: Type_Name is not a canonical type, use getTypeType()?
+  testdata/p4_16_samples/pvs-nested-struct.p4
```
mbudiu-vmw: Can you please file an issue for these showing how to reproduce them?

antoninbas: yes will do. I was going to try and get a backtrace, but if you can take a look that would be great

PR #1416, opened 2018-07-21
Fix overflow bug in stringify. va_copy was missing.

fruffy:


test/gtest/stringify.cpp
```diff
@@ -0,0 +1,46 @@
...
+
+TEST(stringify, simple) {
+    cstring str = appendFormat("%s", "AAA");
+    EXPECT_EQ("AAA", str);
+}
+
+TEST(stringify, overflow) {
+    char test_str[129];
+    for (int i = 0; i < 129; i++)
+        test_str[0] = 'A';
```
mbudiu-vmw: you probably meant test_str[i].
Also, the last character should be 0.

mbudiu-vmw:
Also, please run make cpplint and fix the issue reported.

mbudiu-vmw:
The gtest failed.

fruffy:
This is pretty tricky, I cannot reproduce this bug locally, only with the travis build. The only way I can get the travis build to actually pass is when I declare the string manually as a literal, which does not seem right. What could cause such behaviour?

mbudiu-vmw:
Use a simpler test case which reproduces the bug we had. For example print a 10-character string ten times.

lib/stringify.cpp
```diff
@@ -88,8 +88,11 @@ cstring vprintf_format(const char* fmt_str, va_list ap) {
     if (size < 0)
         throw std::runtime_error("Error in vsnprintf");
     if (static_cast<size_t>(size) >= sizeof(buf)) {
+        va_list ap_copy;
+        va_copy(ap_copy, ap);
```
ChrisDodd: I think that you need to wrap the va_copy/va_end around the FIRST call to vsnprintf (line 87), rather than the second -- the first call leaves `ap` pointing after all the arguments so the second call gets garbage.  Copying after the first call still leaves it pointing after the arguments.

lib/stringify.cpp
```diff
@@ -81,15 +81,15 @@ cstring printf_format(const char* fmt_str, ...) {
 // printf into a string
 cstring vprintf_format(const char* fmt_str, va_list ap) {
     static char buf[128];
+    va_list ap_copy;
+    va_copy(ap_copy, ap);
```
ChrisDodd: You fail to call va_end if the string doesn't overflow.  I'm not sure if that will cause problems, but the spec does require that you call va_end exactly once for each va_start or va_copy, so it risks potential undefined behavior.

fruffy: Yeah I realized while pushing the commit, I am fixing both your comments in the next commit. Thanks for the help!

Issue #1412, opened 2018-07-19
Boolean expression not translated properly in bmv2 backend


antoninbas:
The following P4 code:
```
    action set_true() {
        if (meta.field == 0) {
            meta.cond = true;
        }
    }
```
is not translated to valid bmv2 JSON.
More precisely, the use of a condition in the action leads to the use of the ternary (`?:`) operator in the JSON, which is expected, however **the 2 operands have different types** according to bmv2 (one is a boolean, while the other is a 1-bit values):
```
"value" : {
                      "op" : "?",
                      "left" : {
                        "type" : "bool",
                        "value" : true
                      },
                      "right" : {
                        "type" : "field",
                        "value" : ["scalars", "metadata.cond"]
                      },
                      "cond" : { ... }
```
This issue was originally created in the bmv2 repo (https://github.com/p4lang/behavioral-model/issues/629). A sample P4 program (along with the generated JSON) can be found here: https://gist.github.com/xenerizes/49a5a61a25c001aea1d6414d8ddcd8e6

mbudiu-vmw:
in the conversion from IR to JSON some Boolean values are converted to 1-bit values.
it looks like the conversion of expressions is not systematic enough.
The right way to do this would be to do a separate pass to convert all expressions and insert casts. Today the casts are inserted only in the translation process. I am not sure I will have time to fix this too soon.

PR #1417, opened 2018-07-21
fix for issue #1412

mbudiu-vmw:


midend/predication.cpp
```diff
@@ -15,13 +15,22 @@ limitations under the License.
...
+    CHECK_NULL(result);
+    return result;
+}
+
 const IR::Node* Predication::postorder(IR::AssignmentStatement* statement) {
     if (!inside_action || ifNestingLevel == 0)
         return statement;
-    auto right = new IR::Mux(predicate(), statement->right, statement->left);
+
+    auto right = new IR::Mux(predicate(), statement->right, clone(statement->left));
```
cc10512: A comment why do you need to clone this would be helpful.

backends/bmv2/common/expression.h
```diff
@@ -119,6 +119,7 @@ class ExpressionConverter : public Inspector {
     void postorder(const IR::PathExpression* expression) override;
     void postorder(const IR::TypeNameExpression* expression) override;
     void postorder(const IR::Expression* expression) override;
+    void mapExpression(const IR::Expression* expression, Util::IJson* json);
```
cc10512: I suppose you added these for debugging ... is it worth keeping?

mbudiu-vmw: Yes, in fact is is very useful.

midend/predication.cpp
mbudiu-vmw: I am afraid I won't have time to add this comment in the next 3 weeks.
The reason is that in the end we generate different code for the two copies of the expression, so we want them to be distinct. This happens in many other passes, not only in this one.

mbudiu-vmw:
Can we get a quick approval? I think that I have addressed @cc10512 's comments.

midend/predication.cpp
```diff
@@ -15,13 +15,26 @@ limitations under the License.
...
 #include "predication.h"
+#include "frontends/p4/cloner.h"
 
 namespace P4 {
 
+const IR::Expression* Predication::clone(const IR::Expression* expression) {
+    // We often need to clone expressions.  This is necessary because
+    // in the end we will generate different code for the different clones of
+    // an expression.  This is most obvious if one clone is on the LHS and one
+    // on the RHS of an assigment.
```
ChrisDodd: You shouldn't actually need to clone expressions to do this -- but many of the later passes are broken when they run into dags rather than strict trees.  We should fix those passes, but that is a larger problem...

midend/predication.cpp
```diff
@@ -15,13 +15,26 @@ limitations under the License.
...
+    ClonePathExpressions cloner;
+    auto result = expression->apply(cloner)->to<IR::Expression>();
+    CHECK_NULL(result);
+    return result;
```
ChrisDodd: This could be more concise as `return expression->apply(ClonePathExpressions);` -- the cast is unnecessary and the CHECK_NULL is overkill.

midend/predication.cpp
mbudiu-vmw: Maybe what we should do is to run a pass that converts dags into graphs when it is necessary.
Unfortunately I no longer remember all places where this may be important...
In general this may be important when you maintain a hashmap from a Node* to some other data structure; sometimes the node context is important, so mapping just the Node* is not sufficient.

PR #1418, opened 2018-07-21
[under review] Allow removing GMP, using boost::multiprecision instead

ChrisDodd:
There was a GPL vs Apache licensing question that came up -- this change experiments with replacing GMP use (the only GPL component currently) with boost::mulitprecision.


antoninbas:
From a pure licensing point of view, I believe libgmp is distributed under LGPL, which based on my limited understanding means that we are not violating anything with our current distribution of p4c. That doesn't mean I oppose this change, I don't know if there is any drawback in moving to boost::mp with the cpp_int backend.

mbudiu-vmw:
If you really want to do this probably we should define a third class which wraps one of the other two, based on a compile-time `#define`. We could also give it a nicer API.

ChrisDodd:
So I've updated this change to by default use the GMP backend for boost::multiprecision, but one can configure with `-DENABLE_GMP=OFF` to disable that and use the boost-licensed cpp_int backend instead.  I've refactored the code to have a generic `big_int` type which supports all the operations we need and will be either GMP mpz_int or cpp_int depending on HAVE_GMP.

On change is in the printing of binary constants in toP4 -- we had been using `std::setbase` to set the base to be printed, but according to the C++11 standard, that function can only use the bases 8, 10, or 16 -- trying to use any other base is implementation defined and might not work.  So I changed the code to print base 2 constants in hex (with 0x) instead.

There's also a potential issue with boost and g++ versions -- it seems that boost::multiprecision has problems with boost v1.58 and g++ v6.3.0, which happen to be the default versions for Ubuntu 16.04; our Travis builds work because they use clang.

CMakeLists.txt
```diff
@@ -112,9 +113,15 @@ include_directories(${PROTOBUF_INCLUDE_DIRS})
 include_directories(${LIBGMP_INCLUDE_DIR})
```
cc10512: Should move this and the finding of the GMP package (line 110) under the ENABLE_GMP option.

backends/bmv2/common/control.cpp
```diff
@@ -71,7 +71,7 @@ void ControlConverter::convertTableEntries(const IR::P4Table *table,
                     key->emplace("key", stringRepr(km->left->to<IR::Constant>()->value, k8));
                     auto trailing_zeros = [](unsigned long n) { return n ? __builtin_ctzl(n) : 0; };
                     auto count_ones = [](unsigned long n) { return n ? __builtin_popcountl(n) : 0;};
-                    unsigned long mask = km->right->to<IR::Constant>()->value.get_ui();
+                    auto mask = static_cast<unsigned long>(km->right->to<IR::Constant>()->value);
```
cc10512: Why are static casts needed? Can't this operation be done directly on the big_int object?

control-plane/p4RuntimeSerializer.cpp
```diff
@@ -216,7 +216,7 @@ externalId(const IR::IDeclaration* declaration) {
             return boost::none;
         }
 
-        const uint32_t id = idConstant->value.get_si();
+        const uint32_t id = static_cast<uint32_t>(idConstant->value);
```
cc10512: No automatic casting?

ir/json_generator.h
```diff
@@ -19,7 +19,6 @@ limitations under the License.
 
 #include <assert.h>
 #include <boost/optional.hpp>
-#include <gmpxx.h>
```
cc10512: include lib/gmputil.h for `big_int` definition.

ChrisDodd: it's included indirectly via ir.h

backends/bmv2/common/control.cpp
ChrisDodd: This one could be done directly on big_int (and could have been done directly on mpz_class before), I was just minimalizing the changes.

control-plane/p4RuntimeSerializer.cpp
ChrisDodd: All of boost::multiprecision's conversion operators are marked 'explicit', so you can't implicitly convert to any integer type.  So need the static_cast for a simple conversion.

hesingh:
Regarding licensing and LGPL,  a commercial derivative of software is fine to ship and sell.  However, I believe, any derivative must publish source code as well.   We know p4lang/p4c is used by at least one commercial use but the use does not publish source code.  This is why this PR is needed so that licensing moves to Apache 2.0. and older implementation is deprecated.   No new c++ class is needed to support one of two implementations.

mbudiu-vmw:
The code actually looks nicer. We should file an issue about the binary representation, probably.

PR #1424, opened 2018-07-31
modified graphs::ParserGraphs for reuse

zma-barefoot:
Collect states/transitions for each parser, and save those in a map.


zma-barefoot:
Can I get a review on this one? Thanks!

mbudiu-vmw:
This looks fine, although i don't understand why it is useful.

zma-barefoot:
I would like to reuse the part that collects states/transitions in our backend (for parse graph analysis). Current implementation deletes those after dumping the dot file for each parser.

PR #1429, opened 2018-08-04
change the language used by the preprocessor

### LINKED ISSUE #1428: irrelevant warnings
The preprocessor uses the C language rules to process .p4 files, resulting in warnings such as:
```
constants.p4:53:6: warning: missing terminating ' character
     8'0;
      ^
constants.p4:53:6: warning: missing terminating ' character
     8'0;
      ^
```

cc10512:
Fixes #1428

### DIFF
--- a/frontends/common/options.cpp
+++ b/frontends/common/options.cpp
@@ -292,7 +292,7 @@ FILE* CompilerOptions::preprocess() {
         // line for the preprocessor
         char * driverP4IncludePath =
           isv1() ? getenv("P4C_14_INCLUDE_PATH") : getenv("P4C_16_INCLUDE_PATH");
-        cmd += cstring(" -C -undef -nostdinc") + " " + preprocessor_options
+        cmd += cstring(" -C -undef -nostdinc -x assembler-with-cpp") + " " + preprocessor_options
             + (driverP4IncludePath ? " -I" + cstring(driverP4IncludePath) : "")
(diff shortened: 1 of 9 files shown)


frontends/common/options.cpp
```diff
@@ -292,7 +292,7 @@ FILE* CompilerOptions::preprocess() {
         // line for the preprocessor
         char * driverP4IncludePath =
           isv1() ? getenv("P4C_14_INCLUDE_PATH") : getenv("P4C_16_INCLUDE_PATH");
-        cmd += cstring(" -C -undef -nostdinc") + " " + preprocessor_options
```
ChrisDodd: It seems interesting that here we also use `-C` (preserve comments), `-undef` (remove standard predefines) and `-nostdinc` (remove standard include directories), while in the .cfg files we do not.

tools/driver/p4c_src/p4c.bmv2.cfg
```diff
@@ -28,7 +28,7 @@ class BMV2Backend(BackendDriver):
         self.enable_commands(['preprocessor', 'compiler'])
 
         # options
-        self.add_command_option('preprocessor', "-E -x c")
+        self.add_command_option('preprocessor', "-E -x assembler-with-cpp")
```
ChrisDodd: It looks like `-undef` and `-nostdinc` appear in tools/driver/p4c_src/driver.py -- perhaps the `-x assembler-with-cpp` should be there instead of here in the .cfg files?

PR #1434, opened 2018-08-09
move args inside the .test script

hanw:


hanw:
make it easier to run a test by invoking the .test script directly without any argument.

ChrisDodd:
compare with https://github.com/p4lang/p4c/pull/1382/files

Unfortunately this breaks tests that rely on the same .test script with different arguments.

PR #1437, opened 2018-08-11
32-bit support

cc10512:
This PR ensures that p4c works in a 32-bit environment.

There are a number of assumptions in the code that unsigned long is 8 bytes. Therefore, unsigned long and uintptr_t data types are converted to uint64_t and appropriate constants.

There is still a potential issue with the asLong constant which also returns a long. However, the gmp library casts those values to long itself, and thus the 32-bit version of the library is consistent. Even though, the entire suite of tests passes, I'm not 100% sure that this does not break any other assumptions in the compiler.

### DIFF
--- a/lib/n4.h
+++ b/lib/n4.h
@@ -1,17 +1,18 @@
 #ifndef LIB_N4_H_
 #define LIB_N4_H_
 
+#include <cstdint>
 #include <iomanip>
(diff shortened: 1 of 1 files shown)


cc10512:
Rebased on master. @ChrisDodd please review.

PR #1438, opened 2018-08-13
Construct bitvec from integer types larger than uintptr_t

ChrisDodd:
This uses some template/enable_if magic to only do the overload if there exists a type larger than uintptr_t -- if uintptr_t and uintmax_t are the same, then a simple overload would fail, and just defining for uintmax_t would have trouble with shifts being out of range.

### DIFF
--- a/lib/bitvec.h
+++ b/lib/bitvec.h
@@ -144,6 +144,15 @@ class bitvec {
 
     bitvec() : size(1), data(0) {}
     explicit bitvec(uintptr_t v) : size(1), data(v) {}
+    template<typename T, typename = typename
+        std::enable_if<std::is_integral<T>::value && (sizeof(T) > sizeof(uintptr_t))>::type>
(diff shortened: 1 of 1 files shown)


mbudiu-vmw:
This looks fine, but I really don't have a deep understanding of the bitvec code so I can't tell for sure. Moreover, I believe that none of the tests we have in the public repo exercise this code. Maybe you want to write some unit tests?

ChrisDodd:
This code will be excercised if you build with -m32 (or on a 32-bit machine), but on a normal 64 bit machine (with 64 bit intmax_t), these additions are a noop.

mbudiu-vmw:
The question is "which compiler pass uses bitvecs"?

ChrisDodd:
Hmm -- looks like the frontend doesn't use bitvecs anywhere -- they're only used in private backends.  But they are generally useful.

PR #1439, opened 2018-08-13
fixes for issue-1436

hanw:


frontends/p4/fromv1.0/programStructure.cpp
```diff
@@ -421,13 +423,17 @@ const IR::ParserState* ProgramStructure::convertParser(const IR::V1Parser* parse
     const IR::Expression* select = nullptr;
     if (parser->select != nullptr) {
         auto list = new IR::ListExpression(parser->select->srcInfo, {});
-        std::vector<int> sizes;
+        std::vector<std::pair<bool, int>> signAndSizes;
```
ChrisDodd: How about just using a `std::vector<const IR::Type::Bits *>` -- when the type is a `Bits`, put it directly in the vector, otherwise just get the bits type.

frontends/p4/fromv1.0/programStructure.cpp
```diff
@@ -421,13 +423,17 @@ const IR::ParserState* ProgramStructure::convertParser(const IR::V1Parser* parse
...
         for (auto e : *parser->select) {
             auto c = conv.convert(e);
             list->components.push_back(c);
             int w = c->type->width_bits();
+            bool isSigned = false;
+            if (c->type->is<IR::Type_Bits>())
+                isSigned = c->type->to<IR::Type_Bits>()->isSigned;
```
ChrisDodd: So this becomes

    if (auto *t = c->type->to<IR::Type::Bits>())
        fieldTypes.push_back(t);
    else
        fieldTypes.push_back(IR::Type::Bits::get(c->type->width_bits());

and the explodeLabel/explodeType code become much simpler...

mbudiu-vmw:
you have a cpplint issue

PR #1444, opened 2018-08-15
Fix code generation bug for ebpf booleans

mbudiu-vmw:


mbudiu-vmw:
This is also a one-line change, it should be quick to review.

PR #1447, opened 2018-08-17
python-ply is a dependence if one wants to use tools/stf

mbudiu-vmw:


jafingerhut:
I had noticed 'make check' for p4c failing the ebpf tests without this change as well recently, but did not know the cause or fix.  With this extra package installed, they are all passing.

mbudiu-vmw:
Maybe then someone can approve this one-line documentation change.

Issue #1325, opened 2018-06-06
Error type in nested struct gives Compiler Bug error


Mastertrap21:
We have found out that declaring error types inside nested structs and compiling for the BMv2 target leads to compilation error. In the language specification (https://p4.org/p4-spec/docs/P4-16-v1.0.0-spec.html) it is written under chapter 7.2.7 (Type nesting rules) that both structs and errors are allowed in struct container types. It works fine with p4test so this might be a BMv2 back-end issue

Source file used:
[issue.p4.txt](https://github.com/p4lang/p4c/files/2077203/issue.p4.txt)

Compilation process:
`p4c-bm2-ss issue.p4 -o issue.json`

Error:
```
In file: /home/vagrant/p4c/backends/bmv2/header.cpp:204
Compiler Bug: /usr/local/share/p4c/p4include/core.p4(23): error { NoError, PacketTooShort, NoMatch, StackOutOfBounds, HeaderTooShort, ParserTimeout }: unexpected type for struct test_struct {
  error { NoError, PacketTooShort, NoMatch, StackOutOfBounds, HeaderTooShort, ParserTimeout } test_error; }.test_error
error {
^
issue.p4(6)
struct test_struct {
       ^^^^^^^^^^^
issue.p4(7)
  error test_error;
        ^^^^^^^^^^
```

mbudiu-vmw:
We can implement converting errors to integers quite easily, but we should implement a solution that allows us to also pass errors from the parser. In other words, we should give a solution that also solves #808. @antoninbas : we may need some support from BMv2 to do this properly. Do you have a suggestion?

antoninbas:
bmv2 expects all errors to have been converted to integers and expects the `error` array in the JSON to have been populated appropriately (https://github.com/p4lang/behavioral-model/blob/master/docs/JSON_format.md#errors)
I'm not sure how it relates to #808 but I am definitely recommending replacing `bit<1> checksum_error` in the standard metadata header by a more general `error parser_error` if possible. Then people would be able to run the following code in bmv2:
```
// in ingress
if (std_metata.parser_error != error.NoError) drop();
```
which in the bmv2 JSON would have been replaced by:
```
if (std_metata.parser_error != 0) drop();
```

mbudiu-vmw:
I will close this issue.

PR #1448, opened 2018-08-20
Fix for issue #1325, hopefully for #808 too

### LINKED ISSUE #808: Add parser_error field to v1model standard metadata
This is a proposal to extend v1model's standard metadata with a `bit<X> parser_error` field. The v1model architecture would assign integral values in the appropriate range to each parser error and set this metadata field to the appropriate value so it can be consumed by the ingress control.

I know that such a feature will be available in PSA, but it will take some time to bring PSA software to the level where it can fully replace v1model / bmv2 simple_switch. This enhancement will enable P4 programmers to use parser errors today, with minimal changes in the existing software.

mbudiu-vmw:
@antoninbas : I don't know if this is sufficient to fix #808, but it can't harm.
@hanw: the error code generation was done too early, so the errors[] array was always empty.
We should write some real tests for handling parser errors, but this fix should address #1325.

### DIFF
--- a/p4include/v1model.p4
+++ b/p4include/v1model.p4
@@ -63,7 +63,7 @@ struct standard_metadata_t {
     bit<1>  checksum_error;
     @alias("intrinsic_metadata.recirculate_flag") bit<32> recirculate_flag;
     /// Error produced by parsing
-    bit<32> parser_error;
+    error parser_error;
 }
 
 enum CounterType {
--- a/p4include/v1model.p4
+++ b/p4include/v1model.p4
@@ -59,8 +59,11 @@ struct standard_metadata_t {
     @alias("intrinsic_metadata.resubmit_flag") bit<32> resubmit_flag;
     @alias("intrinsic_metadata.egress_rid")    bit<16> egress_rid;
     /// Indicates that a verify_checksum() method has failed.
+    // TODO: this should be deprecated in favor of parser_error
     bit<1>  checksum_error;
     @alias("intrinsic_metadata.recirculate_flag") bit<32> recirculate_flag;
+    /// Error produced by parsing
+    bit<32> parser_error;
 }
 
 enum CounterType {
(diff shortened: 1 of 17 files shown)


p4include/v1model.p4
```diff
@@ -59,8 +59,11 @@ struct standard_metadata_t {
     @alias("intrinsic_metadata.resubmit_flag") bit<32> resubmit_flag;
     @alias("intrinsic_metadata.egress_rid")    bit<16> egress_rid;
     /// Indicates that a verify_checksum() method has failed.
+    // TODO: this should be deprecated in favor of parser_error
     bit<1>  checksum_error;
     @alias("intrinsic_metadata.recirculate_flag") bit<32> recirculate_flag;
+    /// Error produced by parsing
+    bit<32> parser_error;
```
antoninbas: Could this be of `type` error? Otherwise how can we use it properly in P4?

mbudiu-vmw:
@antoninbas: I wrote a new test which uses parser_error, but this one does not pass.
The core part of the test is this, in ingress:
```
 if (standard_metadata.parser_error == error.PacketTooShort) {
            hdr.eth.setValid();
            hdr.eth.type = 0;
            hdr.eth.src = 0;
            hdr.eth.dst = 0;
}
```

However, when I send a packet too short this specific condition is not true. Should I send you the complete p4 and json test?

antoninbas:
@mbudiu-vmw yes please send both, I can take a look today

antoninbas:
@mbudiu-vmw I just need to add a one-liner in the bmv2 v1model implementation to set this new metadata field after the parser (all the parser error handling logic is already there). I didn't have time to do it today but I will do it tomorrow.

antoninbas:
@mbudiu-vmw I opened this PR: https://github.com/p4lang/behavioral-model/pull/644
I noticed that you are using 1 for error.NoError. Any reason for not using 0?

mbudiu-vmw:
Actually the numbering is done by the backend code. I will have to see what is the order that values are added into the map.

mbudiu-vmw:
I will merge this now, and add the additional test when bmv2 supports it.

antoninbas:
It should support it. I merged my PR already.

PR #1450, opened 2018-08-22
bitvec gtest when no __int128_t defined

cc10512:


cc10512:
@ChrisDodd your guess is correct. At least Ubuntu 16.04 32-bit does not define it.

ChrisDodd:
Actually from the error messages, noone defines it as a macro -- its a builtin type, so #ifdef doesn't work to check for its existence.  StackOverflow suggests checking `#ifdef __SIZEOF_INT128__`

cc10512:
Thanks @ChrisDodd!

PR #1451, opened 2018-08-23
Save and restore ostream fill character in hex.cpp

ChrisDodd:


cc10512:
@ChrisDodd, we also need:
```
diff --git a/lib/hex.h b/lib/hex.h
index 6ba043f0..41c1cafe 100644
--- a/lib/hex.h
+++ b/lib/hex.h
@@ -33,7 +33,9 @@ class hex {

 inline std::ostream &operator<<(std::ostream &os, const hex &h) {
     auto save = os.flags();
+    auto fill = os.fill();
     os << std::hex << std::setw(h.width) << std::setfill(h.fill) << h.val;
+    os.fill(fill);
     os.flags(save);
     return os; }
```

PR #1453, opened 2018-08-26
add support for uint64_t constants in IR::Constant (for 32-bit)

cc10512:
Casting to long on 32-bit machines does not achieve the expected
result that the value is 64-bit. This PR adds support for uint64_t in
IR::Constant, by expanding the set of constructors with the
almost all used integer data types (exception is unsigned long because the
ir parser generator does not handle types that have two qualifiers).

TODO: deprecate the use of asLong throughout the code base. Right now,
it is used only in v1parser.ypp, and as far as I can tell, only for
sizes that should be well within 32-bits (ids, bit-widths, etc.). I do
recommend a review of the code, and replacement with asUnsigned().

Added a gtest to check different values for all the types -- which is really
what forced me to the solution adding all constructors!

### DIFF
--- a/ir/expression.def
+++ b/ir/expression.def
@@ -148,10 +148,14 @@ class Constant : Literal {
         Literal(new Type_InfInt()), value(v), base(base) {}
     Constant(unsigned v, unsigned base = 10) :
         Literal(new Type_InfInt()), value(v), base(base) {}
+#emit
+#if __WORDSIZE < 64
     Constant(long v, unsigned base = 10) :
         Literal(new Type_InfInt()), value(v), base(base) {}
-    // Constant(unsigned long v, unsigned base = 10) : needs fixing the ir parser
-    //     Literal(new Type_InfInt()), value(v), base(base) {}
+    Constant(unsigned long v, unsigned base = 10) :
+        Literal(new Type_InfInt()), value(v), base(base) {}
+#endif
+#end
     Constant(intmax_t v, unsigned base = 10) :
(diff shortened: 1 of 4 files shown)


ir/expression.def
```diff
@@ -141,8 +141,32 @@ class Constant : Literal {
...
+#else
+    Constant(long v, unsigned base = 10) :
+        Literal(new Type_InfInt()), value(v), base(base) {}
+    Constant(unsigned long v, unsigned base = 10) :
+        Literal(new Type_InfInt()), value(v), base(base) {}
     Constant(intmax_t v, unsigned base = 10) :
-            Literal(new Type_InfInt()), value((long)v), base(base) {}
+        Literal(new Type_InfInt()), base(base) {
+        mpz_import(value.get_mpz_t(), /* word count */1, /* msb first */0,
+                   /* word size */sizeof(v), /* native endian */0, /* nails */0, &v); }
```
ChrisDodd: I suspect what might be required here is to check the sign of `v` and negate it before calling mpz_import if it is negative (so the value passed to mpz_import is always positive), then negate the mpz_class afterwards if it was originally negative.  With that, you just need two constructors (intmax_t and uintmax_t) -- all other integer types should convert to one of those two, plus you don't need any special-case for WORDSIZE.

cc10512: Maybe true. We did not have an unsigned constructor before, we were simply using a cast to long, and that was causing problems. I suspect your solution could work, and I will give it a try later. I do want to get this in now to unblock the pipeline, and I will revisit.

PR #1455, opened 2018-08-27
Utilities minor fixes

DanilLutsenko:
Fixed operator std::string += cstring in cstring library. Implementation do nothing now, modifies copy of string and returns copy of modified copy.

Fixed memory leak in stringify utility. cstring makes copy of string internally in strings table,
memory will never be released without GC

LUTSENKO DANIL
SOFTWARE DEVELOPER
Jabil Logo
www.jabil.com

### DIFF
--- a/lib/stringify.cpp
+++ b/lib/stringify.cpp
@@ -80,7 +80,9 @@ cstring vprintf_format(const char* fmt_str, va_list ap) {
     if (static_cast<size_t>(size) >= sizeof(buf)) {
         char* formatted = new char[size + 1];
         vsnprintf(formatted, size + 1, fmt_str, ap_copy);
-        return cstring(formatted);
+        cstring result(formatted);
+        delete[] formatted;
+        return result;
     }
     va_end(ap_copy);
     return cstring(buf);
--- a/lib/cstring.h
+++ b/lib/cstring.h
@@ -199,7 +199,7 @@ inline cstring cstring::operator+=(const char *a) { *this = *this + a; return *t
 inline cstring cstring::operator+=(std::string a) { *this = *this + a; return *this; }
 inline cstring cstring::operator+=(char a) { *this = *this + a; return *this; }
 
-inline std::string operator+=(std::string a, cstring b) {
+inline std::string& operator+=(std::string& a, cstring b) {
     a.append(b.c_str());
     return a; }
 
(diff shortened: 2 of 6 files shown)


lib/stringify.cpp
```diff
@@ -80,7 +80,9 @@ cstring vprintf_format(const char* fmt_str, va_list ap) {
     if (static_cast<size_t>(size) >= sizeof(buf)) {
         char* formatted = new char[size + 1];
         vsnprintf(formatted, size + 1, fmt_str, ap_copy);
-        return cstring(formatted);
+        cstring result(formatted);
```
mbudiu-vmw: If you plan to fix all the leaks in the compiler you have a few thousands left to do.
We actually do not use "delete", this code is written using garbage-collected C++.

DanilLutsenko: Thank you for review! Since p4c has option to build without GC and in bitvec utility invokes delete [] explicitly too (https://github.com/p4lang/p4c/blob/master/lib/bitvec.cpp#L57), can we at least partially fix memory leaks in build without GC? Is delete keyword no-op for build with GC or this is a bug  (I don't see any magic for GC like here https://github.com/p4lang/p4c/blob/master/lib/indent.cpp#L45) in bitvec?

ChrisDodd: Calling delete explicitly should be fine with the garbage collector (its generally a noop unless the memory was marked non-collectable for some reason -- it makes it collectable).  It does explicitly call the destructor if that is relevant (the garbage collector never calls destructors -- only finalizers)

DHrpcs3: Oh, I found magic here https://github.com/p4lang/p4c/blob/5d83957346aef6d0f1abb3f65691744a4738cc3f/lib/bitvec.h#L424 sorry.
So I can rewrite allocation like here and this will work for both builds?

ChrisDodd: The bitvec code was carefully written to work with or without the garbage collector, as it is used in other projects that don't use the garbage collector.

mbudiu-vmw: There is no harm in calling delete if it is safe. But don't expect this to be the only leak.

DanilLutsenko: Thank you for explanation! Looks like there no more leaks which not require changing API or behavior to fix them, so other leaks out of the scope for this request with minor changes.

There any special limitations for this part? I mean, can we use supported by gcc 4.9 and clang 3.3 C++11 features like magic statics and move semantic?

mbudiu-vmw: We can use standard c++11 features.
We do want this code to be as portable as possible.

PR #1463, opened 2018-08-31
cstring optimization

DanilLutsenko:
Added hash utility with implemented FNV-1A hash
Do not make copy of string in cstring if possible
Minor cstring cleanup

LUTSENKO DANIL
SOFTWARE DEVELOPER
jabil
www.jabil.com

### DIFF
--- a/lib/CMakeLists.txt
+++ b/lib/CMakeLists.txt
@@ -19,6 +19,7 @@ set (LIBP4CTOOLKIT_SRCS
 	cstring.cpp
 	gc.cpp
 	gmputil.cpp
+	hash.cpp
 	hex.cpp
 	indent.cpp
 	json.cpp
@@ -46,6 +47,7 @@ set (LIBP4CTOOLKIT_HDRS
 	exceptions.h
 	gc.h
 	gmputil.h
+	hash.h
 	hex.h
 	indent.h
 	json.h
--- a/lib/cstring.cpp
+++ b/lib/cstring.cpp
@@ -15,10 +15,11 @@ limitations under the License.
 */
 
 #include "cstring.h"
-#include "hash.h"
 
 #include <string>
 #include <unordered_set>
+
+#include "hash.h"
 
 namespace {
 enum class table_entry_flags {
@@ -36,7 +37,7 @@ inline table_entry_flags operator |(table_entry_flags l, table_entry_flags r) {
     return static_cast<table_entry_flags>(static_cast<int>(l) | static_cast<int>(r));
 }
 
-//cache entry, ordered by string length
+// cache entry, ordered by string length
 class table_entry {
     std::size_t m_length = 0;
     table_entry_flags m_flags = table_entry_flags::none;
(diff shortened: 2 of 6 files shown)


ChrisDodd:
This seems like a lot of work to duplicate what std::string and std::hash already does -- did you profile it to see if it makes a difference?  If anything, I would expect it to make the code slower and less efficient...

DanilLutsenko:
@ChrisDodd Hi Chris, I wrote simple benchmarks https://github.com/DanilLutsenko/p4-bench (compiling without GC).
On my machine i got following results:
compiled with O0:

| Name | Old |  New |
|-------|-----|------|
|  test_unique_long   |   332   |   225|
| test_unique_short  |    178   |   173|
|  test_shared_long   |   234   |   157|
| test_shared_short    |  166  |    107|
| test_literal_long    |  171   |   156|
|test_literal_short   |   159   |   155|

with -O1:

| Name | Old |  New |
|-------|-----|------|
|  test_unique_long  |  332   |   225 |
| test_unique_short |   178  |    173 |
|  test_shared_long |  234   |   157 |
| test_shared_short  |   166  |    107 |
| test_literal_long      |171  | 156 |
|test_literal_short      |159 |   155 |

with -03:

| Name | Old |  New |
|-------|-----|------|
|  test_unique_long   |   217  |    116 |
| test_unique_short  |     75   |    53 |
|  test_shared_long    |  149   |   110 |
| test_shared_short |      67   |    50 |
| test_literal_long    |   69    |   43 |
|test_literal_short  |     42    |   39 |

Values are avg time of invoking constructors with different arguments in nanoseconds. For details see benchmark source code.

DanilLutsenko:
@mbudiu-vmw There are no comments in the commits about the fixed issues, so you are completely right, no opened issues weren't fixed by the changes.
Only one small memory leak was fixed (for build without GC of course), the cache table is "magic static" object now.
Leaving TODOs I assumed some discussion in inline comments, because I don't know purposes for the related decisions.


If I understand correctly, this is utilities part, which is used in each module, so I think, thing like cstring needs to be as optimized as possible. By the way, not sure we are really need cstring at all, super fast comparison for equality vs long creation, long iteration and no way to use in multi-threaded code. I think std::string usage will be faster for almost all cases, except shared owning of string (if it's really present in the compiler), since std::string with C++11 got move ctor.

Libraries has different code style in each file. Should cstring be Util::CString? Are there any plans for agreements about style? We can use utilities like clang-tidy and clang-format for automated code fixing into correct style.

mbudiu-vmw:
We are not style Nazis; if you pass review you are good enough.
We have way too many things to work on to spend too much time about reformatting existing code.
It's unlikely that the style of these tools will improve the codebase in a significant way. We don't have time to re-review the stylized code. We rely on our modified cpplint tool for catching some classes of style issues.
The reason for cstring is that the entire compiler IR is completely immutable, and we needed an immutable string. The C++ const enforces this thorughout the codebase.

PR #1464, opened 2018-08-31
Option for additional args to test scripts not in the script

ChrisDodd:


ChrisDodd:
This is because we have additional tests for our backend that use the same .test script with multiple different -D arguments added.  So the -D arguments can't be included in the .test script (one variation runs with no -D arguments).  Arguably we should come up with a variation that uses multiple .test scripts, but then we need a way of giving them different names.

mbudiu-vmw:
Could you generate the test name by concatenating the arguments?

Issue #1459, opened 2018-08-29
All ebpf-kernel/testdata/p4_16_samples/*.p4 tests failed


hesingh:
About few months back, before the EBPF backend was reorganized, p4c did not have a dependency on the llvm compiler nor did "make check|recheck" in p4c required sudo priviledge to run.   Both these issues were seen today when using a p4c from today.  More details below.

All ebpf-kernel/testdata/p4_16_samples/*.p4 tests failed.  Debugging the failure, I had to do two things which are not mentioned in the README.md file.  It would be good to update the README.md file.
 
1.	I had to install llvm using “sudo apt-get install llvm”. 
2.	Still “make recheck” failed.  Using “sudo make recheck” worked.  So, now one has to use sudo with “make check “or “make recheck”.

Fix:
----
1. Please add "llvm" to "sudo apt-get install" line in README.md.
2. Change this line in README.md : from "make -j4 check" to  "sudo make -j4 check"

For long-term, find out why ebpf-kernel tests need sudo privilege and change software to not require sudo.

mbudiu-vmw:
Loading ebpf programs into the kernel requires su privileges, so we cannot do it without it.
We should find a way so that you need to type your password only when running these particular tests, and not for all tests.

mbudiu-vmw:
The dependences for the ebpf backend are described in the README in that back-end folder.
We can duplicate this in the main readme, though.

hesingh:
I have pulled a branch to commit the changes to main README.md.  See the changes committed to the branch.  I can use a new Pull of p4c to send out the code review.

https://github.com/hesingh/p4c/commit/dc13f68b59cd40b8ffba4a10269e5033332c2ca8

mbudiu-vmw:
If you don't submit a PR I don't know how to comment.
I don't think we should run the tests with sudo. 
I would rather try to not run the ebpf kernel tests by default; if you want those you have to type a separate command-line.

Also, please note that the documentation is modular, so in theory the ebpf back-end-related documentation should be in the README in that folder, although no one can find it there.

hesingh:
Right now, the global "make check" is running the ebpf-kernel tests and fails on ebpf-kernel.  This is why sudo was required.

ChrisDodd:
[...]

PR #1465, opened 2018-08-31
Skip tests that require root rather than failing them

mbudiu-vmw:
This should fix issue #1459.
To run the tests that require root one should do `sudo make check-ebpf-kernel`.


README.md
```diff
@@ -168,7 +168,7 @@ Most dependencies can be installed using `apt-get install`:
 `sudo apt-get install g++ git automake libtool libgc-dev bison flex
 libfl-dev libgmp-dev libboost-dev libboost-iostreams-dev
 libboost-graph-dev pkg-config python python-scapy python-ipaddr python-ply
-tcpdump cmake`
+tcpdump cmake llvm`
```
cc10512: Would be nice to keep this in alphabetical order. I realize there was a previous offender, but it's not too hard to fix.

PR #1467, opened 2018-09-01
Fix for issue #1334

mbudiu-vmw:
@cc10512 : this implements overload resolution as defined in the newest version of the spec.
We still need to handle default parameter values to complete the new rules for procedure calling.


cc10512:
Thank you, Mihai! I will review over the weekend.

PR #1468, opened 2018-09-04
LocalCopyProp fix for abstract functions

ChrisDodd:
- abstract function calls might call any concrete function defined in
  the target object.
- propagate reads/writes from callee to caller properly.

### DIFF
--- a/midend/local_copyprop.cpp
+++ b/midend/local_copyprop.cpp
@@ -357,16 +357,26 @@ IR::MethodCallExpression *DoLocalCopyPropagation::postorder(IR::MethodCallExpres
                 LOG3("table apply method call " << mc->method);
                 apply_table(&tables[obj]);
                 return mc;
-            } else if (mem->expr->type->is<IR::Type_Extern>()) {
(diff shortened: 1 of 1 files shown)


mbudiu-vmw:
This is related to #1284. I think we need a real way to solve this at the language level, perhaps using some annotations. My proposal there was to use a `@synchronous(callers)` annotation on *callees* to indicate who may call them and when. I still think this is sound. If a method is not tagged as `@synchronous` then it is not allowed to capture any variables.

mbudiu-vmw:
It's not about blocking, it's about when they are being called.
`@synchronous(caller) void callee()` means that callee is only called by caller when caller executes, and the caller is expected to execute in the context of a packet.
It cannot be invoked by any other method, or, most importantly, by the control-plane.

mbudiu-vmw:
We should bring this to the design meetings.

ChrisDodd:
If we do add annotations it makes more sense to do it the other way around (eg, `@may_call(callee) abstract void caller();` since generally functions have no idea when they might be called.

ChrisDodd:
Superceded by #1284

Issue #1381, opened 2018-07-03
Remove the MethodCallDescription class


mbudiu-vmw:
Fold its functionality in the MethodInstance class.

PR #1472, opened 2018-09-05
 Fixes for issues #1333 and #1381 

mbudiu-vmw:
This completes the implementation of handling named arguments.

### DIFF
--- a/frontends/p4/defaultArguments.h
+++ b/frontends/p4/defaultArguments.h
@@ -41,6 +41,7 @@ class DoDefaultArguments : public Transform {
     { setName("DoDefaultArguments"); CHECK_NULL(refMap); CHECK_NULL(typeMap); }
     const IR::Node* postorder(IR::MethodCallExpression* expression) override;
     const IR::Node* postorder(IR::Declaration_Instance* inst) override;
+    const IR::Node* postorder(IR::ConstructorCallExpression* ccc) override;
 };
 
 class DefaultArguments : public PassManager {
--- a/frontends/p4/defaultArguments.h
+++ b/frontends/p4/defaultArguments.h
@@ -36,7 +36,7 @@ class DoDefaultArguments : public Transform {
     ReferenceMap* refMap;
     TypeMap*      typeMap;
 
-  public:
+ public:
     DoDefaultArguments(ReferenceMap* refMap, TypeMap* typeMap): refMap(refMap), typeMap(typeMap)
     { setName("DoDefaultArguments"); CHECK_NULL(refMap); CHECK_NULL(typeMap); }
     const IR::Node* postorder(IR::MethodCallExpression* expression) override;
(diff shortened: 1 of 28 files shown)


frontends/p4/defaultArguments.h
```diff
@@ -0,0 +1,58 @@
...
+    const IR::Node* postorder(IR::MethodCallExpression* expression) override;
+    const IR::Node* postorder(IR::Declaration_Instance* inst) override;
+    const IR::Node* postorder(IR::ConstructorCallExpression* ccc) override;
+};
+
+class DefaultArguments : public PassManager {
+ public:
+    DefaultArguments(ReferenceMap* refMap, TypeMap* typeMap) {
+        setName("DefaultArguments");
+        passes.push_back(new TypeChecking(refMap, typeMap));
```
cc10512: Type checking first and then adding the default arguments? I would have thought you would check after you've added them.

frontends/p4/typeChecking/typeChecker.cpp
```diff
@@ -1406,6 +1407,14 @@ const IR::Node* TypeInference::postorder(IR::Parameter* param) {
         return param;
     BUG_CHECK(!paramType->is<IR::Type_Type>(), "%1%: unexpected type", paramType);
 
+    if (param->defaultValue != nullptr) {
+        auto init = assignment(param, paramType, param->defaultValue);
+        if (init == nullptr)
+            return param;
+        if (param->defaultValue != init)
+            param->defaultValue = init;
```
cc10512: Are you overloading `defaultValue`? Or is assignment likely to return a simplified expression?

frontends/p4/typeChecking/typeUnification.cpp
```diff
@@ -123,9 +123,9 @@ bool TypeUnification::unifyCall(const IR::Node* errorPosition,
             ++paramIt;
     }
 
-    // Check remaining parameters: they must be all optional
+    // Check remaining parameters: they must be all optional or have default values
     for (auto p : left) {
-        bool opt = p.second->isOptional();
+        bool opt = p.second->isOptional() || p.second->defaultValue != nullptr;
```
cc10512: +1. And soon we can get rid of @optional!

frontends/p4/defaultArguments.h
mbudiu-vmw: I tried to to it the other way round, but it's much harder.
The problem is that given an expression a.b.c.isValid() to know what the function isValid is you need to know more about a.b.c and that information comes from type checking.

frontends/p4/typeChecking/typeChecker.cpp
mbudiu-vmw: Assignment does type inference, and may return a new expression with modified types. In particular, `int` types may be converted to `bit<>` types.

frontends/p4/typeChecking/typeUnification.cpp
mbudiu-vmw: At this point we don't have anything to replace optional.

PR #1482, opened 2018-09-13
Avoid crashing after errors

ChrisDodd:
- after an error, arbitrarily returning nullptr from a PassManager can
  result in a crash or error cascade -- we instead return the tree from
  before the error.
- stop_on_error should only abort after a NEW error, to allow some
  continued error checking after an error.

### DIFF
--- a/ir/pass_manager.cpp
+++ b/ir/pass_manager.cpp
@@ -28,6 +28,7 @@ const IR::Node *PassManager::apply_visitor(const IR::Node *program, const char *
     } nest_log_indent(log_indent);
 
     early_exit_flag = false;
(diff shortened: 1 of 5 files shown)


ir/pass_manager.cpp
```diff
@@ -28,6 +28,7 @@ const IR::Node *PassManager::apply_visitor(const IR::Node *program, const char *
     } nest_log_indent(log_indent);
 
     early_exit_flag = false;
+    int initial_error_count = ::errorCount();
```
mbudiu-vmw: We may need to do this in other places too...

PR #1486, opened 2018-09-16
Add P4_16 + v1model arch test program for testing multiple "packet pa…

jafingerhut:
…ths"

i.e. resubmit, recirculate, ingress-to-egress clone, egress-to-egress
clone, and multicast replication.  It has a P4_16 program plus an STF
test that checks that output packets are as they should be, checked
pretty carefully by hand.

### DIFF
--- a/backends/bmv2/bmv2stf.py
+++ b/backends/bmv2/bmv2stf.py
@@ -387,6 +387,14 @@ class RunBMV2(object):
             self.do_cli_command(self.parse_table_add(cmd))
         elif first == "setdefault":
             self.do_cli_command(self.parse_table_set_default(cmd))
+        elif first == "mirroring_add":
+            # Pass through mirroring_add commands unchanged, with same
+            # arguments as expected by simple_switch_CLI
+            self.do_cli_command(first + " " + cmd)
+        elif first == "mc_mgrp_create" or first == "mc_node_create" or first == "mc_node_associate":
+            # Pass through multicast group commands unchanged, with
+            # same arguments as expected by simple_switch_CLI
+            self.do_cli_command(first + " " + cmd)
         elif first == "packet":
             interface, data = nextWord(cmd)
             interface = int(interface)
--- a/testdata/p4_16_samples/v1model-special-ops-bmv2.p4
+++ b/testdata/p4_16_samples/v1model-special-ops-bmv2.p4
@@ -1,0 +1,470 @@
+/*
+Copyright 2018 Cisco Systems, Inc.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+#include <core.p4>
+#include <v1model.p4>
+
+#undef ENABLE_DEBUG_TABLES
+
+// These definitions are derived from the numerical values of the enum
+// named "PktInstanceType" in the p4lang/behavioral-model source file
+// targets/simple_switch/simple_switch.h
+
+// https://github.com/p4lang/behavioral-model/blob/master/targets/simple_switch/simple_switch.h#L126-L134
+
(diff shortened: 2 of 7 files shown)


mbudiu-vmw:
The STF files were first defined informally and the bmv2stf program was created. Then a more formal implementation was made in stf_parser.py, but the bmv2stf program was never converted to use the new implementation, and now it only handles a subset.  It is actually pretty easy to use, but we always had more pressing problems. If you want to extend the STF grammar we should probably rewrite bmv2stf as well, it's too difficult to keep two versions in sync.

testdata/p4_16_samples/v1model-special-ops-bmv2.p4
```diff
@@ -0,0 +1,470 @@
...
+            standard_metadata.deq_qdepth : exact;
+            standard_metadata.ingress_global_timestamp : exact;
+            standard_metadata.egress_global_timestamp : exact;
+            standard_metadata.lf_field_list : exact;
+            standard_metadata.mcast_grp : exact;
+            standard_metadata.resubmit_flag : exact;
+            standard_metadata.egress_rid : exact;
+            standard_metadata.checksum_error : exact;
+            standard_metadata.recirculate_flag : exact;
+            //standard_metadata.parser_error : exact;
```
mbudiu-vmw: Filed #1490 to keep track of this.

testdata/p4_16_samples/v1model-special-ops-bmv2.p4
```diff
@@ -0,0 +1,470 @@
...
+
+
+control fill_ipv4_address(out bit<32> ipv4_address,
+                          in bit<8> byte0,    // most significant byte
+                          in bit<8> byte1,
+                          in bit<8> byte2,
+                          in bit<8> byte3)    // least significant byte
+{
+    apply {
+        ipv4_address = (((bit<32>) byte0 << 24) |
```
mbudiu-vmw: must easier to use concatenation `++`

testdata/p4_16_samples/v1model-special-ops-bmv2.p4
```diff
@@ -0,0 +1,470 @@
...
+        // BMV2_V1MODEL_INSTANCE_TYPE_RESUBMIT.  The instance_type
+        // field is an exception to the "preserve metadata field
+        // value" rule.
+
+        // For the resubmit operation, standard_metadata.resubmit_flag
+        // is 0 for the resubmitted packet, which avoids that packet
+        // being resubmitted indefinitely, unless your program
+        // explicitly causes a separate call to resubmit() on each
+        // execution of ingress.
+        resubmit(standard_metadata);
```
mbudiu-vmw: You should an explicit type argument to resubmit if you want field names.
`resubmit<standard_metadata_t>(standard_metadata)`

backends/bmv2/bmv2stf.py
```diff
@@ -387,6 +387,14 @@ def do_command(self, cmd):
             self.do_cli_command(self.parse_table_add(cmd))
         elif first == "setdefault":
             self.do_cli_command(self.parse_table_set_default(cmd))
+        elif first == "mirroring_add":
```
mbudiu-vmw: Filed #1489 to unify this script with the stf parsing library

mbudiu-vmw:
I see no reason not to merge this.

testdata/p4_16_samples/v1model-special-ops-bmv2.p4
jafingerhut: I tried making that change to the source code, and at least with the version of p4c I compiled it with, it made no difference in the resulting compiled BMv2 JSON file.  It contained the list of field names with or without the `<standard_metadata_t>`.

jafingerhut: I believe at least one of the intermediate P4_16 programs generated by the compiler (maybe after front end, maybe after mid end) has this `<standard_metadata_t>` type specifier added to the P4_16 code automatically, so it shouldn't depend upon the back end.

testdata/p4_16_samples/v1model-special-ops-bmv2.p4
jafingerhut: I am making that change and updating this PR soon.  Doh!  Can't believe I forgot about concatenation while writing out that expression.

testdata/p4_16_samples/v1model-special-ops-bmv2.p4
mbudiu-vmw: that makes a difference when you use a constant (list) as an argument.

PR #1492, opened 2018-09-17
add equality operator for ordered_set

zma-barefoot:


mbudiu-vmw:
I think that @ChrisDodd should review this.
What does it mean for two ordered_sets to be equal?

zma-barefoot:
Shouldn't it be same as the equal operator for ordered_map where we compare the underlying std::list?

https://github.com/p4lang/p4c/blob/7a8fae074a44398f3843d74cda178639715af4d7/lib/ordered_map.h#L113

mbudiu-vmw:
Do you want to also write some gtests for this?
This seems to assume that the underlying list type has the correct equality operator.

ChrisDodd:
I hadn't realized we had an `ordered_map::operator ==` -- that is arguably wrong as it only compares the values in the ordered map and does not compare the keys at all (so two ordered_maps with the same values in the same order, but different keys will compare as equal -- that just seems wrong to me).

PR #1493, opened 2018-09-17
Add comments to v1model.p4 about simple_switch not using some metadat…

jafingerhut:
…a fields


mbudiu-vmw:
The tests will fail because the position in error messages will change, so you need to update the reference outputs.

jafingerhut:
@mbudiu-vmw If only you had a dollar for every time you had to remind someone of that, eh?  Updated line numbers and tests passing now.

Issue #677, opened 2017-05-27
p4c bmv2-v1model compiler bug when using typedef for standard_metadata_t


tomrodeheffer:
I got tired of writing standard_metadata_t on the parameter lists of the various v1model controls that need it.  So I abbreviated by saying

    typedef standard_metadata_t SM;

I expected this would be just fine, since section 7.3 in the 2017-05-22 P4_16 spec says "The two types are treated as synonyms, and all operations that can be executed using the original type can be also executed using the newly created type."

Here is my test file: [smeta.p4.txt](https://github.com/p4lang/p4c/files/1033818/smeta.p4.txt)
p4c says:

    In file: ../backends/bmv2/expression.cpp:239
    Compiler Bug: ../backends/bmv2/expression.cpp:239: Null name

hanw:
This is a regression caused by recent refactoring. Will fix.

mbudiu-vmw:
I think that this has been fixed.

PR #1494, opened 2018-09-18
Remove deprecated drop field from p4c tests for v1model arch

jafingerhut:
I know that the p4c test programs are often not complete functional
programs, but leaving a deprecated standard_metadata_t field like drop
in use in the test programs can be confusing to P4 developers looking
for examples.

### DIFF
--- a/testdata/p4_16_bmv_errors/issue679-bmv2.p4
+++ b/testdata/p4_16_bmv_errors/issue679-bmv2.p4
@@ -29,7 +29,7 @@ parser ParserI(packet_in pk, out H hdr, inout M meta, inout standard_metadata_t 
 
 control IngressI(inout H hdr, inout M meta, inout standard_metadata_t smeta) {
 
-    action drop() { smeta.drop = 1; }
+    action drop() { mark_to_drop(); }
 
     action_selector (HashAlgorithm.identity, 32w1024, 32w10) as;
 
--- a/testdata/p4_16_samples/action_profile-bmv2.p4
+++ b/testdata/p4_16_samples/action_profile-bmv2.p4
@@ -30,7 +30,7 @@ action empty() { }
 
 control IngressI(inout H hdr, inout M meta, inout standard_metadata_t smeta) {
 
-    action drop() { smeta.drop = 1; }
+    action drop() { mark_to_drop(); }
 
     table indirect {
         key = { }
--- a/testdata/p4_16_samples/action_selector_shared-bmv2.p4
+++ b/testdata/p4_16_samples/action_selector_shared-bmv2.p4
@@ -30,7 +30,7 @@ action empty() { }
 
 control IngressI(inout H hdr, inout M meta, inout standard_metadata_t smeta) {
 
-    action drop() { smeta.drop = 1; }
+    action drop() { mark_to_drop(); }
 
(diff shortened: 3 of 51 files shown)


jafingerhut:
Wherever a P4_16 + v1model architecture test program assigns a value of 1 to deprecated standard_metadata_t field `drop`, replace it with a call to `mark_to_drop()`, which is what BMv2 simple_switch actually uses to determine whether to drop a packet.

Wherever such a test program assigns a value of 0 to this deprecated field (less common, but a few occurrences), replace it with an assignment to field `egress_spec` instead.

Update a few expected test output files so that the tests continue to pass.

With these changes, I have checked that it is possible to comment out the fields `drop` and `recirculate_port` in file p4include/v1model.p4, and with only a few additional changes in the expected test output files, the tests still all pass.  This PR does _not_ include changes to remove those fields, but makes that change easier to make in the future, if we decide that is desirable.

jafingerhut:
@antoninbas There is one of those that I forgot to mention (only one).  It is the file issue677-bmv2.p4.  There is no STF file for this one, so packet forwarding behavior doesn't need to be preserved for that one to keep it working.  In looking at the issue #677 my guess was that having one explicit assignment statement to at least one field of the variable of type `standard_metadata_t` in place of `smeta.drop = 1;` was probably closer to the original reason the program was created.  My guess was that replacing that assignment with `mark_to_drop();` would have changed the compiler behavior enough that it might no longer be testing what it originally did.  Difficult to check without going back and rerunning the pre-fixed version of the compiler on the program, though.  I can try that if you think it is worth it.

testdata/p4_16_bmv_errors/issue679-bmv2.p4
```diff
@@ -29,7 +29,7 @@ parser ParserI(packet_in pk, out H hdr, inout M meta, inout standard_metadata_t
 
 control IngressI(inout H hdr, inout M meta, inout standard_metadata_t smeta) {
 
-    action drop() { smeta.drop = 1; }
+    action drop() { mark_to_drop(); }
```
mbudiu-vmw: One thing about all these programs called `issueXXX` is that they are testing a particular bug in the compiler. If we edit them we don't know for sure that they are triggering the same bug or not. It is unlikely that the issue is related to the drop field, so probably this is fine. But in general programs called `issue` should not be used as examples.

If you want to remove the drop field from the metadata then you will have to make these edits.

jafingerhut: I agree there is a danger in changing test programs that they no longer exercise the bug the compiler once had.  If you think it is important before changing these to check that the modified test program gives the same error, with the version of the compiler before the fix, as the original test program, I can do that.

jafingerhut: Understood.

One more quick thought, if it would be preferable to this PR: I could take the current v1model.p4, make a copy of it with a different name in the testsdata/p4_16_samples directory, e.g. v1model-old1.p4, and have the existing tests that refer to standard_metadata.drop `#include` that one instead of the standard one.  Then the test programs would keep doing what they do now, even if `p4include/v1model.p4` was later changed to remove the `drop` field.

mbudiu-vmw:
I will bite the bullet and merge this.

PR #1498, opened 2018-09-20
Struct initializer

mbudiu-vmw:
This PR was much harder to implement than I expected.
This introduces a new IR node, called StructInitializerExpression.
This will be very useful if we decide to add C99-like struct intializers to P4 - which we probably should.
This also makes it easier to write some analyses - because some ListExpressions were implicitly converted to structs, and now these conversions are explicit in the use of a struct initializer instead of a ListExpression.

(I also realized that the type-checker is too lenient: it allows some conversions between tuples and structs which should be illegal, but it's not easy to fix that with our current setup.)

### DIFF
--- a/ir/base.def
+++ b/ir/base.def
@@ -183,6 +183,11 @@ class Path {
     validate { BUG_CHECK(!name.name.isNullOrEmpty(), "Empty path"); }
 }
 
+/// Handy class used in several NamedMaps
+class NamedExpression : Declaration {
+    Expression expression;
+}
+
 /// Annotations are used to provide additional information to the compiler
 /// Most P4 entities can be optionally annotated
 class Annotation {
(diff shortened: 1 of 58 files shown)


ir/base.def
```diff
@@ -183,26 +183,30 @@ class Path {
...
 /// Annotations are used to provide additional information to the compiler
 /// Most P4 entities can be optionally annotated
 class Annotation {
     ID name;
     /// Annotations that are simple expressions
     inline Vector<Expression> expr;
     /// Annotations described as key-value pairs
-    // TODO: this should be a map from ID not from cstring
-    inline NameMap<Expression> kv;
+    inline IndexedVector<NamedExpression> kv;
```
ChrisDodd: why IndexedVector instead of NameMap?  You can use `NameMap<NamedExpression, ordered_map>` to specify an ordered map as the underlying map type (to preserve order semantics).

mbudiu-vmw: These two are pretty much equivalent in functionality; they have slightly different APIs. For example, the IndexedVector only works with IDeclaration objects, and cannot support multisets of keys. I can convert to a NameMap if you prefer.

mbudiu-vmw:
@ChrisDodd @cc10512 @hanw 
If we want to merge this you may want to test your back-ends first. This changes the IR from ListExpression to StructInitializerExpression in several places. The bmv2 back-end shows the typical change required.

mbudiu-vmw:
@cc10512 @ChrisDodd : would you guys like to test this PR with your back-end?
It changes the way struct initializers are represented in the IR, so it may break some of your test cases. If it doesn't I will merge it.

hanw:
will do.

PR #1501, opened 2018-09-20
Refactored generate() for pairs.

liujed:
Refactored `generate()` for pairs to introduce `toJSON()`, for cases where we just want to generate the body of a JSON object.

### DIFF
--- a/ir/json_generator.h
+++ b/ir/json_generator.h
@@ -80,11 +80,18 @@ class JSONGenerator {
 
     template<typename T, typename U>
     void generate(const std::pair<T, U> &v) {
-        out << "{" << std::endl << ++indent << "\"first\" : ";
+        ++indent;
+        out << "{" << std::endl;
+        toJSON(v);
+        out << std::endl << --indent << "}";
+    }
+
+    template<typename T, typename U>
+    void toJSON(const std::pair<T, U> &v) {
+        out << indent << "\"first\" : ";
         generate(v.first);
         out << "," << std::endl << indent << "\"second\" : ";
         generate(v.second);
-        out << std::endl << --indent << "}";
     }
 
     template<typename T>


ir/json_generator.h
```diff
@@ -80,11 +80,18 @@ class JSONGenerator {
...
     void generate(const std::pair<T, U> &v) {
-        out << "{" << std::endl << ++indent << "\"first\" : ";
+        ++indent;
+        out << "{" << std::endl;
+        toJSON(v);
+        out << std::endl << --indent << "}";
+    }
+
+    template<typename T, typename U>
+    void toJSON(const std::pair<T, U> &v) {
```
mbudiu-vmw: I think that the name of this function is misleading; someone may expect it to generate a full JSON object.
You should call it something else, like toJsonObjectBody.

liujed: I agree, but this name is actually consistent with `Node::toJSON()`: https://github.com/p4lang/p4c/blob/d37330b117dd4bb42e21067a75f67e5fa3a5749c/ir/node.cpp#L27

mbudiu-vmw:
Node is an abstract class, so you won't accidentally misuse it.

liujed:
@mbudiu-vmw Sorry, I'm not following. Because `Node` has `toJSON()`, all AST nodes will implement that method. One might then call `toJSON()` on an arbitrary AST node, expecting to get a full JSON object, but end up with just the body of one.

`JSONGenerator::toJSON(std::pair<T,U>&)` is intended to be analogous to `Node::toJSON(JSONGenerator&)` — if we were able to modify `std::pair` itself, then I would have instead added a `std::pair::toJSON(JSONGenerator&)`. I can rename `JSONGenerator::toJSON(std::pair<T,U>&)` to something else, but for consistency, I think `Node::toJSON(JSONGenerator&)` should also be renamed. If that's what you think should be done, please say so, and I can do that in a separate PR.

---

@ChrisDodd This refactoring is to support a bug fix in Barefoot's extension to p4c. Currently, there is an AST node whose `toJSON()` calls `generate` for `std::pair`, which leads to an extraneous set of braces.

liujed:
@mbudiu-vmw Because you requested changes, merging is blocked until you approve. Are you happy with things as they stand in this PR, or do you want me to rename all `toJSON` methods in the compiler?

mbudiu-vmw:
For consistency of naming the Node methods do not need to be renamed. 
Each method for a concrete (non-abstract) class will always generate full JSON.
The name of your method is misleading, because it does not generate JSON for a pair.

liujed:
Please see the implementation of `Node::toJSON` here: https://github.com/p4lang/p4c/blob/d37330b117dd4bb42e21067a75f67e5fa3a5749c/ir/node.cpp#L27 It just generates a JSON body, not a full object.

mbudiu-vmw:
Node is an abstract class, and this method is overridden in all it's derived classes. So this method is never executed except when called by other methods, as @ChrisDodd has explained.

liujed:
Here is the generated code for `EmptyStatement::toJSON()` and all overridden implementations that it calls. Note that it doesn't generate a full JSON object because `Node::toJSON()` doesn't. I'm fairly certain `EmptyStatement` is not abstract.
```
void IR::EmptyStatement::toJSON(JSONGenerator & json) const {
    Statement::toJSON(json);
}
void IR::Statement::toJSON(JSONGenerator & json) const {
    StatOrDecl::toJSON(json);
}
void IR::StatOrDecl::toJSON(JSONGenerator & json) const {
    Node::toJSON(json);
}
```

mbudiu-vmw:
You may be right in that none of the toJSON methods seems to generate a complete JSON value.

Issue #1499, opened 2018-09-20
using `latest` without preceding header extraction causes crash


zma-barefoot:
The incorrect program below, which uses the `latest` keyword without a preceding header extraction causes crash in the v1 converter.

```
header_type inboxes {
  fields {
    cornstarchs : 48;
    khans : 16;
  }
}

header_type whitesand_t {
  fields {
    contracts : 48;
  }
}

header inboxes extirpations;
metadata whitesand_t hesitation;

parser start {
  set_metadata(hesitation.contracts, latest.cornstarchs);
  extract(extirpations);
  return select (latest.khans) {
    1024 : ingress;
  }
}
control ingress { }
```

backtrace:

```
#0  0x0000000002a3ad0e in P4V1::ExpressionConverter::postorder (this=0x7fffffffcb80, field=0x46dbb00)
    at /home/zma/HEAD/bf-p4c-compilers/p4c/frontends/p4/fromv1.0/converters.cpp:77
#1  0x0000000002c90861 in IR::Member::apply_visitor_postorder (this=0x46dbb00, v=...)
    at /home/zma/HEAD/bf-p4c-compilers/p4c/ir/visitor.cpp:446
#2  0x0000000002c58ae0 in Transform::apply_visitor (this=0x7fffffffcb80, n=0x46c7a00, name=0x0)
    at /home/zma/HEAD/bf-p4c-compilers/p4c/ir/visitor.cpp:323
```

PR #1502, opened 2018-09-20
Fix for issue #1499

mbudiu-vmw:


mbudiu-vmw:
We have a bunch of (very) small fixes that need review. Anyone around?

Issue #1478, opened 2018-09-10
Check value of size property for keyless tables?


antoninbas:
Would it be a good idea to check the value of the `size` table property for keyless tables in a midend pass? If the `size` is greater than 0, I would suggest printing a warning and replacing the program-provided value with 0 in the IR.
```
table {
  key = { }
  actions = { a; }
  size = 2;
}
```
We actually ran into a situation recently where someone wrote a similar program, so it may be worth it to do this check in the compiler and I would imagine it would benefit most backends (thus my suggestion to make it a midend pass).

hesingh:
size is an optional property of a table.  If a table is keyless, a backend should be robust enough to ignore size.  I don't have a strong preference.  It's fine to add such a check to a midend pass, but we should make sure the midend pass is included by all backends.

mbudiu-vmw:
There are other interesting cases to consider, like specifying the size and specifying the const table intializers...

PR #1503, opened 2018-09-20
Fix for issue #1478

mbudiu-vmw:
Unfortunately the stderr is produced by p4test, and not by p4c-bm2-ss.
But when running p4c-bm2-ss on the sample it will produce two warnings:
```
issue1478-bmv2.p4(25): warning: 3: Size specified for table without keys
        size = 3;
               ^
issue1478-bmv2.p4(35): warning: 10: Size specified for table with constant entries
        size = 10;
               ^^
```

### DIFF
--- a/midend/checkSize.h
+++ b/midend/checkSize.h
@@ -22,21 +22,32 @@ limitations under the License.
 namespace P4 {
 
 /// Checks some possible misuses of the table size property
-class CheckTableSize : public Inspector {
+class CheckTableSize : public Modifier {
  public:
     CheckTableSize() { setName("CheckTableSize"); }
-    bool preorder(const IR::P4Table* table) override {
+    bool preorder(IR::P4Table* table) override {
         auto size = table->getSizeProperty();
         if (size == nullptr)
             return false;
+
+        bool deleteSize = false;
         auto key = table->getKey();
         if (key == nullptr) {
-            if (size->value != 1)
+            if (size->value != 1) {
                 ::warning("%1%: Size specified for table without keys", size);
+                deleteSize = true;
+            }
         }
         auto entries = table->properties->getProperty(IR::TableProperties::entriesPropertyName);
-        if (entries != nullptr && entries->isConstant)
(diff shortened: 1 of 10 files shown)


midend/checkSize.h
```diff
@@ -0,0 +1,46 @@
...
+
+#ifndef _MIDEND_CHECKSIZE_H_
+#define _MIDEND_CHECKSIZE_H_
+
+#include "ir/ir.h"
+
+namespace P4 {
+
+/// Checks some possible misuses of the table size property
+class CheckTableSize : public Inspector {
```
antoninbas: Would it make sense to not make this an inspector and actually replace the size by 0 in that case, so that the backend doesn't have to worry about it?

mbudiu-vmw: Why not just remove the size property rather than replace it by 0?

antoninbas: I don't really have a preference. It seems that either way each backend has to be careful. Currently the bmv2 backend unconditionally sets the table size to 1024 when the size is missing (or when the provided size is 0), which seems wrong: https://github.com/p4lang/p4c/blob/master/backends/bmv2/common/control.cpp#L386

mbudiu-vmw: I think that this behavior is inherited from the p4-14 compiler.
The back-end should probably do this only when it is really necessary - i.e., the table size can change dynamically.

antoninbas: bmv2 actually enforced fixed table sizes to emulate hardware targets
the default table size code should probably be updated to account for the key size, at least for exact match tables?

mbudiu-vmw: Then let's file a new issue. Probably @hanw is the right person to handle this, since I believe he has ownership of the bmv2-based backends. For constant tables the compiler should somehow measure the table size and tell bmv2 what that is.

mbudiu-vmw: In the meantime it looks to me like deleting the table size for these two cases is the right thing. The back-end is supposed to measure the table anyway. 0 seems a bit arbitrary.

PR #1506, opened 2018-09-21
Revert "Sanitize file path input in cmake, add travis check to skip t…

liujed:
…ests (#1476)"

This reverts commit 7fd89e95bf07916a864e0c9ec6521df1458f0ade.


liujed:
For context: the reverted commit broke Barefoot's internal regressions. There is some discussion in #1476 for a more backwards-compatible approach, which I think @fruffy will be trying out.

PR #1507, opened 2018-09-21
 Improve error message for a specific type inferencing failure that seems to come up frequently.

ChrisDodd:
The specific case that comes up is passing a list of values to validateChecksum -- if one value is an untyped constant, then the error message was just giving the line info about the validateChecksum declaration (in the header file) and not providing any clarity as to the actual problem.  Several people have run into this problem and been unable to figure out what was going wrong from the error message.

I'm not a big fan of elaborating special cases in error messages like this, so it would be better to come up with a more general scheme; perhaps saving the expressions that are used to infer types in type inferencing.

### DIFF
--- a/frontends/p4/typeChecking/bindVariables.cpp
+++ b/frontends/p4/typeChecking/bindVariables.cpp
@@ -2,6 +2,20 @@
 
 namespace P4 {
 
+class FindUntypedInt : public Inspector {
+    const IR::Node *pos;
+    bool preorder(const IR::Node *) override { return pos == nullptr; }
+    bool preorder(const IR::Type *) override { return false; }
+    void postorder(const IR::Constant *c) override { if (c->type->is<IR::Type_InfInt>()) pos = c; }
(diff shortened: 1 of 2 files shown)


mbudiu-vmw:
I think this falls under the umbrella of #19.
I think I know how to improve all these error messages, but I didn't have time to work on it.
What you should do is keep track of all the things you were trying to unify which led you to an error, and print them as part of the error message. That's what llvm does. This may take care of this problem too.

mbudiu-vmw:
The problem with your error message is that it may be wrong; if there are several ints you may be blaming the wrong one.

ChrisDodd:
If there are several untyped ints, then they are all at fault and will need to be corrected to make the code work; this shows only the first of them.

I added this specific fix as this is the 3rd time I've seen someone run into the error message that only identifies the verifyChecksum declaration in v1model.p4 and not the location in their source program.  It should at least not make the problem worse.

A general fix under #19 would be better and would make this unnecessary.

PR #1510, opened 2018-09-21
Fix inverted test

mbudiu-vmw:


mbudiu-vmw:
This is a one character fix, I would appreciate a review.

PR #1512, opened 2018-09-22
Sanitize CMake file paths. Second attempt.

fruffy:
This is the second attempt to fix the path inputs for the CMake testing script. Both tests and their corresponding xfail counterparts now follow identical patterns to make matching of file on different paths work. 
However, instead of casting the paths to absolute values after the `file GLOB` matching, it is now done before the function call. The original output of the `file GLOB` operation is preserved. In addition, the absolute value function does not use the REALPATH parameter anymore. It now does not follow symlinks in order to prevent odd behavior. I hope this prevents the regression tests from breaking.

### DIFF
--- a/cmake/P4CUtils.cmake
+++ b/cmake/P4CUtils.cmake
@@ -141,11 +141,26 @@ endmacro(p4c_add_test_list)
 
 # generate a list of test name suffixes based on specified testsuites
 function(p4c_find_test_names testsuites tests)
-  foreach(ts "${testsuites}")
+  set(__tests "")
+  p4c_sanitize_path("${testsuites}" abs_paths)
+  foreach(ts ${abs_paths})
     file (GLOB __testfiles RELATIVE ${P4C_SOURCE_DIR} ${ts})
     list (APPEND __tests ${__testfiles})
   endforeach()
   set(${tests} "${__tests}" PARENT_SCOPE)
+endfunction()
+
+# convert the paths from a list of input files to their absolute value,
+# does not follow symlinks.
+#   - files is a list of (relative) file paths
+#   - abs_files is the return set of absolute file paths
+function(p4c_sanitize_path files abs_files)
+  foreach(file ${files})
+    get_filename_component(__file "${file}"
+                       ABSOLUTE BASE_DIR "${P4C_SOURCE_DIR}")
+    list (APPEND __abs_files ${__file})
+  endforeach()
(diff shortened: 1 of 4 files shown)


fruffy:
Running Travis in[ sudo required](https://docs.travis-ci.com/user/reference/overview/#virtualization-environments) mode actually fixes the namespace issue. `sudo: required` uses a VM instead of Docker as far as I understand. I have not found a way to run the Travis instance as a privileged Docker container unfortunately.

cmake/P4CUtils.cmake
```diff
@@ -141,13 +141,28 @@ endmacro(p4c_add_test_list)
...
+  set(__tests "")
+  p4c_sanitize_path("${testsuites}" abs_paths)
+  foreach(ts ${abs_paths})
     file (GLOB __testfiles RELATIVE ${P4C_SOURCE_DIR} ${ts})
     list (APPEND __tests ${__testfiles})
   endforeach()
   set(${tests} "${__tests}" PARENT_SCOPE)
 endfunction()
 
+# convert the paths from a list of input files to their absolute value,
```
mbudiu-vmw: to absolute paths.
No comma at the end

cmake/P4CUtils.cmake
```diff
@@ -141,13 +141,28 @@ endmacro(p4c_add_test_list)
...
+# does not follow symlinks.
```
mbudiu-vmw: does this work when someone uses a symlink in the extensions folder?

tools/testutils.py
```diff
@@ -29,6 +29,7 @@
 FAILURE = 1
 SKIPPED = 2  # used occasionally to indicate that a test was not executed
 
+
```
mbudiu-vmw: no need for an extra empty line

cmake/P4CUtils.cmake
cc10512: yes.

fruffy:
Hold on, I have some small improvements ready soon.

mbudiu-vmw:
The p4c-xdp build has been broken for a few days, we should merge this so we can fix that one as well.
Some changes you made to that project assumed a specific position in the filesystem, and broke when people installed the code in a different place. In particular, relative paths ".." do not traverse symlinks backwards.

fruffy:
Yes, unfortunately I did not consider symlinks. I switched the top-level os.path.abspath import to a os.path.realpath import. This should fix the issue of build setups.

Issue #1520, opened 2018-09-27
God made the integers...


jnfoster:
... but compile-time known values are the work of man.

Compiling this program:
```p4
#include <core.p4>
#include <v1model.p4>

struct headers {
    bit<16> x;
}

struct metadata {
}

parser MyParser(packet_in packet,
                out headers hdr,
                inout metadata meta,
                inout standard_metadata_t standard_metadata) {
    
    state start {
	hdr.x = 0;
        transition accept;
    }
}

control MyVerifyChecksum(inout headers hdr, inout metadata meta) {   
    apply {  }
}

control C(inout headers hdr, inout metadata meta)(bool b) {
    register<bit<16>>(32w8) r;
    apply {
	r.read(hdr.x, 0);
    }
}

control H(inout headers hdr, inout metadata meta) {
    C(true) c1;
    C(false) c2;
    apply {
	c1.apply(hdr, meta);
	c2.apply(hdr, meta);
    }
}

control MyIngress(inout headers hdr,
                  inout metadata meta,
                  inout standard_metadata_t standard_metadata) {
    H() h;
    apply {
	h.apply(hdr, meta);
    }
}

control MyEgress(inout headers hdr,
                 inout metadata meta,
                 inout standard_metadata_t standard_metadata) {
    apply {  }
}

control MyComputeChecksum(inout headers  hdr, inout metadata meta) {
    apply {  }
}

control MyDeparser(packet_out packet, in headers hdr) {
    apply { }
}

V1Switch(
MyParser(),
MyVerifyChecksum(),
MyIngress(),
MyEgress(),
MyComputeChecksum(),
MyDeparser()
) main;
```
yields the error:
```
/home/p4/bug2.p4(27): error: 8: Cannot evaluate to a compile-time constant
    register<bit<16>>(32w8) r;
                      ^^^^
```
It's hard to see what could be more compile-time constant than a fixed-width integer literal :-)

The error does not arise if we pass the size of the register as a constructor parameter to `C` or if we only instantiate `C` once.

PR #1522, opened 2018-09-27
Fix for issue #1520

mbudiu-vmw:


frontends/p4/evaluator/evaluator.h
```diff
@@ -41,7 +41,7 @@ class Evaluator final : public Inspector, public IHasBlock {
  public:
     Evaluator(const ReferenceMap* refMap, const TypeMap* typeMap) :
             refMap(refMap), typeMap(typeMap), toplevelBlock(nullptr)
-    { CHECK_NULL(refMap); CHECK_NULL(typeMap); setName("Evaluator"); }
+    { CHECK_NULL(refMap); CHECK_NULL(typeMap); setName("Evaluator"); visitDagOnce = false; }
```
ChrisDodd: This change leads to a bunch of crashes in `setValue` -- by visiting the same node multiple times, it tries to call setValue multiple times, leading to the BUG_CHECK at ir/ir.cpp:194.  Why was it added?

mbudiu-vmw: The test case that this PR fixes is a good example.
Each block has a map of constant values as evaluated within that block.
If you have a diamond pattern, where two blocks instantiate a third one, some constants are evaluated only once if you don't do visitDagOnce = false, so they are missing from some maps.
This is the right way to do this. 

If you have a test case which crashes please share it.

PR #1523, opened 2018-09-28
support const default_action in p4-14

hanw:
This PR implements an experimental feature in P4-14 to support const default_action by introducing a new table property keyword 'const_default_action'.


hanw:
bmv2 currently does not support `action_const` on `indirect` table as it is following p4-14 spec. https://github.com/p4lang/behavioral-model/blame/master/docs/JSON_format.md#L549

This may be a question for Antonin: does bmv2 support experimental language feature that is not in the p4-14 spec yet? If so, then I can modify the bmv2 backend to generate the flag in the output json, but we also need to modify bmv2 itself to support the flag.

hanw:
Mergin this, and opened #1526 to track the issue in bmv2 backend.

Issue #1524, opened 2018-09-28
Another 'typechecker mutated program' error, even with PR #1519 patch applied


jafingerhut:
I tried out your patch from that PR because I had a different program that gave a 'typechecker mutated program', and was hoping that change would fix it, but it does not.  Program attached.
[typechecker-mutated-error.p4.txt](https://github.com/p4lang/p4c/files/2426902/typechecker-mutated-error.p4.txt)

Issue #1517, opened 2018-09-25
Error "Compiler bug: typechecker mutated program" for short P4_16 program


jafingerhut:
I am not sure if this is serious or not, but thought I would provide a cut-down example program that caused this error message to appear with latest p4c as of 2018-Sep-24, in case it turns out to be a compiler bug that could have more subtle consequences.

```
$ p4c --target bmv2 --arch v1model compiler-bug2.p4
In file: /home/jafinger/forks/p4c/frontends/p4/typeChecking/typeChecker.cpp:146
Compiler Bug: typechecker mutated program
```

[compiler-bug2.p4.txt](https://github.com/p4lang/p4c/files/2414014/compiler-bug2.p4.txt)

PR #1525, opened 2018-09-28
Fix for issue #1524

mbudiu-vmw:


jafingerhut:
I am a poor person to review the contents of the patch, but I can say that I have tried the latest p4c, with this patch applied, on top of the one for issue #1517, and with both of those patches combined, p4test completes with no 'type checker mutated program' errors on some slightly-updated PSA example programs that helped detect this issue.

frontends/common/constantFolding.cpp
```diff
@@ -37,7 +62,7 @@ const IR::Expression* DoConstantFolding::getConstant(const IR::Expression* expr)
         // Casts of a constant to a value with type Type_Newtype
         // are constants, but we cannot fold them.
         if (getConstant(cast->expr))
-            return expr;
+            return CloneConstants::clone(expr);
```
cc10512: If we haven't succeeded in folding the constant, why does it still need to be cloned?

mbudiu-vmw: We cannot replace these expressions with constant values, we have to leave them as (cast)constant. But we still have to clone the constant itself - due to the same reasons that we clone constants everywhere in this pass. (The cast will disappear after the `type` constructs are eliminated.) This was the case that was triggering the reported bug.

PR #1527, opened 2018-09-28
pad header in const_default_action.p4 to byte boundary

hanw:


jafingerhut:
@hanw Isn't it a mistake to have the line `header ingress_metadata_t ingress_metadata <name>` in this P4_14 test program?  All other P4_14 test programs that have a `header_type ingress_metadata_t` instead later define `metadata ingress_metadata_t <name>`.

Was there a reason you chose to use `header ingress_metadata_t <name>` in this test program?  If not, changing it to `metadata` should mean there is no need to have an explicit padding field, I would think.

PR #1532, opened 2018-09-30
Change bmv2 back end to give error if any header is not a multiple of…

jafingerhut:
… 8 bits

Previously it silently padded user-defined P4 headers to make them a
multiple of 8 bits in bmv2 simple_switch, which appears to violate
both the P4_14 and P4_16 specifications.


jafingerhut:
Sorry for such a long comment.  Maybe it will be useful to anyone reviewing this, in answering their questions about what these changes do, and why they are proposed.

Definition: Call a header whose fields have a total size that is a
multiple of 8 bits a "byte-aligned" header.  All other header are
"non-byte-aligned".


The P4_14 language spec, v1.0.4, Section 2.2 "Header and Metadata
Instances":

    "For header instances, the compiler must produce an error if the
    total length of all fields in a header type is not an integral
    number of bytes.  The compiler may pad the header to be byte
    aligned."

The P4_16 language spec, v1.0.0, Section 7.2.2 "Header Types"

    "There is no padding or alignment of the header fields.
    Architectures may impose additional constraints on header types --
    e.g., restricting headers to sizes that are an integer number of
    bytes."



I am confident that BMv2 simple_switch is a P4 target that imposes the
restriction that all headers must be byte aligned.

Evidence for this:

behavioral-model/src/bm_sim/header.cpp has several assert statements
like this:

  assert(nbytes_packet % 8 == 0);

These cause simple_switch to crash if you attempt to call extract() on
a non-byte-aligned header.  I have tested this by manually editing a
P4 program's compiled BMv2 JSON file to eliminate a padding field in a
header definition for a header that had an extract() call in its
parser, sending a packet in, and one of those assert statements causes
simple_switch to crash.

I have no interest in changing this simple_switch restriction.  I
cannot think of any good reasons to change it to support
non-byte-aligned headers.


If you compile a P4_14 or P4_16 program using p4c with simple_switch
as the target, the bmv2 back end code today pads all headers to be a
multiple of 8 bits long, if they are not already, by adding a padding
field from 1 to 7 bits in size.  It does this even for P4_14 metadata
and P4_16 structs, which have no such restriction -- the padding
fields are probably there for bmv2 internal implementation reasons
that I have not investigated.  I see no reason to change that bmv2
internal implementation detail, which should be invisible to its
users.

Adding such padding _silently_ for P4_14 or P4_16 headers seems to
violate the language specs, as quoted above.

The proposed change is: in the bmv2 back end, used when compiling both
P4_14 or P4_16 programs to the bmv2 target, issue an error message if
a user-defined header is encountered that is non-byte-aligned.

Before and after this proposed change, the p4c front end code allows
headers to be other than a multiple of 8 bits long.  This is useful,
in case anyone wants to use that front end with a back end that
supports this.

Before and after this change, p4test can be used to auto-translate
P4_14 source code with non-byte-aligned headers to P4_16 source code
with non-byte-aligned headers.  This seems reasonable to me, such
P4_16 programs might not be intended to run on a target with the
byte-aligned header restriction.

Existing p4c tests that currently contain non-byte-aligned headers,
and suggestions for what to do about them:

For the following programs, I propose simply increasing the length of
the last field in the non-byte-aligned header definition, so that it
becomes byte-aligned.  I have looked at the test program, and if
available, the issue created that led to the test program being added,
and as far as I can tell, the reason for the program has nothing to do
with testing non-byte-aligned headers, but other things.

    p4_14_samples/14-GatewayGreaterThan.p4
    p4_16_samples/inline-stack-bmv2.p4
    p4_16_samples/issue1062-1-bmv2.p4
    p4_16_samples/issue1062-bmv2.p4
    p4_16_samples/issue134-bmv2.p4
    p4_16_samples/table-entries-exact-bmv2.p4
    p4_16_samples/table-entries-exact-ternary-bmv2.p4
    p4_16_samples/table-entries-lpm-bmv2.p4
    p4_16_samples/table-entries-priority-bmv2.p4
    p4_16_samples/table-entries-range-bmv2.p4
    p4_16_samples/table-entries-ternary-bmv2.p4
    p4_16_samples/table-entries-valid-bmv2.p4

For this test:
    p4_14_samples/02-BadSizeField.p4

As for the previous one, this was added in April 2016 with the initial
set of files in the p4c repository, so it is not easy to determine if
there is a compiler bug that was fixed that it is intending to test.

The name of the file suggests that it is intended to deal with a badly
sized field, but it isn't clear from the program whether the badly
sized field is a metadata field inside of header_type vag_t, which is
byte-aligned, or the non-byte-aligned header ethernet_t, which
unusually has an ethertype of 15 bits instead of 16.

Because the ethertype field is not used anywhere in the program, and
one of the fields in the vag_t header is, I am guessing the intent was
to test the access to that field, which is not a multiple of 8 bits by
itself, but with the other field totals a multiple of 8 bits.

Proposal: Change the ethertype field from 15 to 16 bits, making header
ethernet_t byte-aligned, like it is in most programs.


For this test:

    p4_14_samples/14-SplitEthernetVlan.p4

It appears to have been added when the p4c repository was created in
April 2016, so there is no easy way to determine the reason it was
written, or if there was a compiler bug fix it was meant to test.

Normally if one writes a P4 program that deals with the Ethernet
802.1Q header, one would write it like this:

    header_type pcp_t {
        fields {
            pcp : 3;
            cfi : 1;
            vlan_id_t : 12;
            ethertype : 16;
        }
    }

Test program 14-SplitEthernetVlan.p4 instead defines 3 different
headers, each with one of the first 3 fields, and there is parser code
to extract those 3 headers in 3 different parse states, one after the
other.

That approach would work correctly on a target that supports
non-byte-aligned headers, but if the premise of these changes is
correct, it should give an error when bmv2 is the target, and the only
target the p4c P4_14 test programs have are: (a) bmv2, and (b)
auto-translating to P4_16 source code with the v1model architecture.

If you use today's p4c and behavioral-model code to compile this
program and run it on simple_switch, the bmv2 back end silently adds
padding to each of those 3 headers to make them byte-aligned, and when
parsing packets, it will _not_ be parsing an 802.1Q header, but
something else that has extra padding bits after each of the 3 fields.

Proposal: Delete this test.  It does not do anything that other tests
do not already cover that I can tell, other than using
non-byte-aligned headers.

Note: I have tried instead of deleting it, moving it to the
p4_14_errors directory.  It does give an error when compiling it to
target bmv2, but programs in that directory are currently also
auto-translated to P4_16 using p4test, and that does _not_ give an
error for this program, so 'make check' fails.

If we think we want to keep this test program, there needs to be a way
to cause 'make check' to expect and check error messages when
compiling it to target bmv2, but to expect no errors when translating
it to P4_16.

I would like to leave this file out of this commit for now, and if the
rest of these changes are merged in, I will try to find a way to add
this program back in so that there are no test failures.

jafingerhut:
@antoninbas You are welcome to review these changes if you wish, but even if you could only confirm my belief that bmv2 is only intended to handle headers that are a multiple of 8 bits wide, that would be helpful.  The basic change here is to cause p4c to give an error when targeting bmv2 if any user-defined header is not a multiple of 8 bits.

jafingerhut:
I will ping a couple of other people here, so they have at least a chance to object in case I am missing something undesirable in these changes: @cc10512 @vgurevich @ChrisDodd @jnfoster

antoninbas:
@jafingerhut I wanted to merge this but it seems that there are some merge conflicts that need to be resolved. Would you mind taking a look?

jafingerhut:
@antoninbas I have attempted to resolve the conflicts, which fortunately were pretty small and easy to decide what to do.  We'll see if the tests all pass soon.

jafingerhut:
@antoninbas OK, finally actually got a correct-looking merge.  Tests passing.

Issue #1538, opened 2018-10-01
p4test/p4c give errors for this P4_16 program with function calls


jafingerhut:
Program attached.  It seems like a fairly straightforward and expected use of functions in P4_16 to me.

[p4-functions.p4.txt](https://github.com/p4lang/p4c/files/2436037/p4-functions.p4.txt)

mbudiu-vmw:
The fix is small, but it will cause lots of reference outputs to change, because it needs to rename some temp variables...

jafingerhut:
My test program was added as a new test case in testdata, and I have compared the output files for it in testdata/p4_16_samples_outputs against the original program, and they appear to correctly implement the original source code.  Closing this issue.

PR #1539, opened 2018-10-02
Fix for issue #1538

mbudiu-vmw:
I apologize for the large set of changes to the outputs. I don't know how to avoid those with our current testing framework.


jafingerhut:
No addition of the test program that exhibits the bug?  Maybe it was in there but has neither the original name nor the name issue1538 so I missed it.

mbudiu-vmw:
I had called it issue1358 instead of 1538.
I have committed a renamed version.

mbudiu-vmw:
Can we get a review for this patch?

frontends/p4/uniqueNames.cpp
```diff
@@ -80,8 +80,8 @@ addNameAnnotation(cstring name, const IR::Annotations* annos) {
 }
 
 UniqueNames::UniqueNames(ReferenceMap* refMap) : renameMap(new RenameMap) {
-    setStopOnError(true);
     setName("UniqueNames");
+    visitDagOnce = false;
```
ChrisDodd: Won't this create multiple unique names for a single shared object?  Thus turning it into multiple objects?

Issue #1541, opened 2018-10-08
Action shadowing extern function triggers CompilerBug


jnfoster:
Compiling this program using `p4c`,
```p4
#include <core.p4>
#include <v1model.p4>
struct metadata { 
  bit<3> x;
}
struct headers { }
parser MyParser(packet_in packet,
                out headers hdr,
                inout metadata meta,
                inout standard_metadata_t standard_metadata) {
    
    state start {
        transition accept;
    }
}

control MyVerifyChecksum(inout headers hdr, inout metadata meta) {   
    apply {  }
}


control MyIngress(inout headers hdr,
                  inout metadata meta,
                  inout standard_metadata_t standard_metadata) {
    action hash() {
	hash(meta.x, HashAlgorithm.crc16, 3w0, { }, 3w7);
    }    
    apply { }
}

control MyEgress(inout headers hdr,
                 inout metadata meta,
                 inout standard_metadata_t standard_metadata) {
    apply {  }
}

control MyComputeChecksum(inout headers  hdr, inout metadata meta) {
    apply {  }
}

control MyDeparser(packet_out packet, in headers hdr) {
    apply { }
}

V1Switch(
MyParser(),
MyVerifyChecksum(),
MyIngress(),
MyEgress(),
MyComputeChecksum(),
MyDeparser()
) main;
```
generates the following bug report:
```
bug.p4(25): warning: hash shadows hash
    action hash() {
           ^^^^
/usr/local/share/p4c/p4include/v1model.p4(126)
extern void hash<O, T, D, M>(out O result, in HashAlgorithm algo, in T base, in D data, in M max);
            ^^^^
In file: /home/vagrant/p4c/frontends/p4/typeChecking/typeChecker.cpp:158
Compiler Bug: bug.p4(25): Could not find type of action hash() {
  hash(meta.x, <TypeNameExpression>(1147)HashAlgorithm.crc16, 0, {}, 7); }
    action hash() {
           ^^^^
```

PR #1543, opened 2018-10-10
Fix for issue #1541

mbudiu-vmw:


mbudiu-vmw:
Functions are resolved using argument names/argument counts, and not types.

mbudiu-vmw:
But actions do not participate in this disambiguation, that's why the compiler treats this as a recursive call.
You can call the function by writing `.hash(...)`. with a dot in front.

jnfoster:
Sounds good. Thanks as always for the quick fix!

PR #1552, opened 2018-10-14
add saturated arithmetic support to p4-14

cc10512:
Implements saturated arithmetic in P4-14. Since the fields are tagged
with the saturated annotation, we add isSaturated to the Type_Bits.

This also requires refactoring the map caching the bit
types. Previously, we were storing signed and unsigned in separate
maps. Adding more maps to accomodate the saturated types is
impractical, so we unified the maps and added a key consisting of
(width, sign, saturation).

Since P4-16 supports only operations as saturated, the conversion
routines for the supported primitives, add, subtract, add_to_field,
subtract_from_field, change the operation to saturated if any of its
arguments are saturated.

### DIFF
--- a/frontends/p4/fromv1.0/programStructure.cpp
+++ b/frontends/p4/fromv1.0/programStructure.cpp
@@ -945,6 +945,20 @@ static bool sameBitsType(
     return true;  // to prevent inserting a cast
 }
 
+static bool isSaturatedField(const IR::Expression *expr) {
+    auto member = expr->to<IR::Member>();
+    if (!member)
+        return false;
+    auto header_type = dynamic_cast<const IR::Type_StructLike *>(member->expr->type);
+    if (!header_type)
+        return false;
+    auto field = header_type->getField(member->member.name);
+    if (field && field->getAnnotation("saturating")) {
+        return true;
+    }
+    return false;
+}
+
 // Implement modify_field(left, right, mask)
 const IR::Statement* ProgramStructure::sliceAssign(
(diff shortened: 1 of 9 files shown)


cc10512:
There is really no need to do any additional type checking as the only way they are used are to convert the P4-14 primitives to the P4-16 defined operations, which have the proper semantics. As we discussed before, there are _no_ saturated types in P4-16. I was expecting to have an annotation, however, the type annotation was completely ignored in the v1 parser.

mbudiu-vmw:
Yes, I think that would help.
You are saying that the specific field is only used during p4-14 to p4-16 translation, and ignored afterwards.

cc10512:
I reworked the PR based on @ChrisDodd's annotation and left the Type_Bits as is, except for removing the maps.

frontends/p4/fromv1.0/programStructure.cpp
```diff
@@ -1090,9 +1104,17 @@ CONVERT_PRIMITIVE(bit_xnor) {
 CONVERT_PRIMITIVE(add) {
     OPS_CK(primitive, 3);
     auto args = convertArgs(structure, primitive);
+    bool isSaturated = false;
+    for (auto a : args)
+        isSaturated = isSaturated || isSaturatedField(a);  // any of the fields
```
ChrisDodd: would

    bool isSaturated = std::any_of(args.begin(), args.end(), isSaturatedField);

be simpler/clearer?

Issue #1544, opened 2018-10-10
Compiler BUG from bmv2 back end when compiling P4_16 functions


jafingerhut:
I get an error like this one when compiling any of the 3 attached programs, likely with the same root case and fix for all.
```
$ p4c --target bmv2 --arch v1model function-error-try1.p4.txt 
In file: /home/jafinger/p4c/backends/bmv2/common/expression.cpp:168
Compiler Bug: function-error-try1.p4.txt(81): sometimes_dec(tmp_0);: unhandled case
        hdr.ethernet.srcAddr[15:0] = sometimes_dec(hdr.ethernet.srcAddr[15:0]);
                                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```
Note: this error occurs with the latest p4c as of 2018-Oct-05, and also with the proposed patch for this PR: https://github.com/p4lang/p4c/pull/1539

In looking at the mid-end P4_16 files created, it appears that the function body is not being in-lined into the body of the control, leading to a function call in the IR sent to the bmv2 back end.  I am not sure, but maybe the front/mid-end of the compiler should be inlining the function body here?

[function-error-try1.p4.txt](https://github.com/p4lang/p4c/files/2465910/function-error-try1.p4.txt)
[function-error-try2.p4.txt](https://github.com/p4lang/p4c/files/2465911/function-error-try2.p4.txt)
[function-error-try3.p4.txt](https://github.com/p4lang/p4c/files/2465913/function-error-try3.p4.txt)

jafingerhut:
I have tested my sample programs with the recently committed fix.  The Compiler Bug message no longer occurs, and I have examined the P4_16 code from the last midend compiler pass, and it appears equivalent in behavior to the input program.  Closing this issue.

PR #1555, opened 2018-10-16
Fix for issue #1544

mbudiu-vmw:


mbudiu-vmw:
The original function inlining code comes with nested inlining tests.
The bug was essentially that when inlining in an action you forgot the things you had to inline in the parent control. So we need to keep a stack of inlining worklists.
I can add a test with a single function inlined multiple times if there isn't one.

Issue #1542, opened 2018-10-09
Action parameters that has a direction in const entries


dingrui37:
According to the description about default action in P4_16 spec (13.2.1.3 Default action)

_The default action must be one of the actions that appear in the actions list. In particular, the expressions passed as in, out, or inout parameters must be syntactically identical to the expressions used in one of the elements of the actions list._

When I write my P4 program,  I find if default action parameters not equal to the expressions in the action list, it will give a compile error, this result is as expected. 

```
    action multicast(in bit<32> multicast_group_id) {
        clone(CloneType.I2E, multicast_group_id);
    }

    table forward {
        key = {
            hdr.arp.isValid()      : exact;
            hdr.arp.oper           : ternary;
            hdr.arp_ipv4.isValid() : exact;
        }

        actions = {
            drop;
            multicast(1);
        }

        const default_action = multicast(2);
        const entries = {
            ( true, ARP_OPER_REQUEST, true) : multicast(2);
        }
    }

~$p4c-bm2-ss packet-replication.p4 --std p4-16 -o packet-replication.json
packet-replication.p4(166): error: 2: argument does not match declaration in actions list: 1
        const default_action = multicast(2);
                                         ^
packet-replication.p4(163)
            multicast(1);
                      ^
```
So I fix this bug and then go on the test, the compile result is successful.
```
    const default_action = multicast(2); 
->  const default_action = multicast(1);

```

But why don't report an error about `const entries`? If the action parameter in const entires live empty, it will give a compile error, but if the value is not equal to the expressions in the action list, it can still be compiled successfully. I think this a bug.

```
        actions = {
            drop;
            multicast(1);
        }
       ...
        const default_action = multicast(1);
        const entries = {
            ( true, ARP_OPER_REQUEST, true) : multicast();
        }
       ...

packet-replication.p4(168): error: multicast: Parameter multicast_group_id must be bound
            ( true, ARP_OPER_REQUEST, true) : multicast();
                                              ^^^^^^^^^^^
packet-replication.p4(150)
    action multicast(in bit<32> multicast_group_id) {
                                ^^^^^^^^^^^^^^^^^^
```

```
    action multicast(in bit<32> multicast_group_id) {
        clone(CloneType.I2E, multicast_group_id);
    }

    table forward {
        key = {
            hdr.arp.isValid()      : exact;
            hdr.arp.oper           : ternary;
            hdr.arp_ipv4.isValid() : exact;
        }

        actions = {
            drop;
[...]

PR #1556, opened 2018-10-16
Fix for issue #1542

mbudiu-vmw:


frontends/p4/typeChecking/typeChecker.cpp
```diff
@@ -3227,6 +3207,64 @@ const IR::Node* TypeInference::postorder(IR::KeyElement* elem) {
...
+
+    auto al = table->getActionList();
+    if (al == nullptr) {
+        typeError("%1% has no action list, so it cannot have %2%",
+                  table, actionCall);
+        return;
+    }
+
+    auto call = actionCall->to<IR::MethodCallExpression>();
+    CHECK_NULL(call);
```
ChrisDodd: If the expression is not a MethodCallExpression, should that be a error rather than a BUG_CHECK?

mbudiu-vmw: The pass createBuiltins will convert other expressions to method calls in this context.

mbudiu-vmw: But maybe there are other things that could happen, so an error would not be bad either.

PR #1558, opened 2018-10-17
Save/restore prune_flag in Transform::apply_visitor

ChrisDodd:
A couple of people have tried to call `prune()` in a Transform::preorder function before calling visit on some child, which doesn't work (as visiting the child will reset the prune flag.)  This avoids that problem by making the recursive visit call preserve the flag, so things will work as expected regardless of when prune() is called.


mbudiu-vmw:
How about a little review on one or more of my PRs instead?

PR #1561, opened 2018-10-18
fix for issue 1559

hanw:


mbudiu-vmw:
@ChrisDodd wrote this code, he's the right person to do it.
I don't know much about the semantics of P4-14, and especially the legacy programs that people want to run.

PR #1565, opened 2018-10-19
remove unnecessary annotations in v1model

hanw:
let's see what breaks..


mbudiu-vmw:
@hanw: can you please either close this or fix it?

mbudiu-vmw:
This is old enough to look like it is not maintained.

Issue #1566, opened 2018-10-19
p4c generates two extern instances with identical `@name` names


jafingerhut:
There is no compilation error when compiling this program for the BMv2 target, but simple_switch does correctly complain about the two instances with identical names when it attempts to read the BMv2 JSON file produced.

The attached .zip file contains the source program, the output of the compiler midend pass 36, which is the latest one that I believe was generated before the compiler began the BMv2 back end.  It shows two counter extern instances that have the same `@name` annotation names, which looks like a bug to me.
```
control cIngress(inout Parsed_packet hdr, inout metadata_t meta, inout standard_metadata_t stdmeta) {
    @name("cIngress.E.c1.stats") counter(32w65536, CounterType.packets) E_c1_stats;
    @name("cIngress.E.c1.stats") counter(32w65536, CounterType.packets) E_c1_stats_0;
    apply {
        hdr.ethernet.etherType = hdr.ethernet.etherType << 1;
        E_c1_stats.count((bit<32>)(hdr.ethernet.etherType + 16w1));
        hdr.ethernet.etherType = hdr.ethernet.etherType + 16w1;
        hdr.ethernet.etherType = hdr.ethernet.etherType << 3;
        E_c1_stats.count((bit<32>)(hdr.ethernet.etherType + 16w1));
        hdr.ethernet.etherType = hdr.ethernet.etherType + 16w1;
    }
}
```
I believe there should be only one counter instantiation, not two.  The `apply` block code in the midend pass 36 program does seem to correctly invoke a count method twice on the same instance, not using the second erroneous instance at all.

This code was written based on sample code discussed on this issue: https://github.com/p4lang/p4c/issues/1545

[prog5.zip](https://github.com/p4lang/p4c/files/2497536/prog5.zip)

jafingerhut:
By the way, creating a test case for this program by making its name have a '-bmv2.p4' suffix and creating a simple one-packet '-bmv2.stf' file for it should exhibit the bug, since no packet will come out.  I don't see a way to create an automated test without an STF file that will fail for this program, since the compiler itself gives no error messages.

jafingerhut:
Similar erroneous code was produced by two versions of p4c that I tested this with, both just before and just after the commit for issue #1538 that was merged on 2018-Oct-19.

jafingerhut:
Added test case here: https://github.com/p4lang/p4c/pull/1568

jafingerhut:
[...]

PR #1568, opened 2018-10-20
Add test case for issue #1566

jafingerhut:


jafingerhut:
The recent commits made with PR #1567 appear to have fixed issue #1566 as well.  This PR adds a test case that not only compiles the program, but also runs a packet through bmv2 using an STF file to verify that the BMv2 JSON file is doing what it should.

PR #1576, opened 2018-10-23
Insert zero initialized default arguments when not present in program

amreshk:


amreshk:
Checked, no change in outputs.

Issue #1580, opened 2018-10-27
Nice if compiler gave error on use of actions violating @defaultonly @tableonly annotations


jafingerhut:
The attached program has two uses of actions that the compiler seems like it ought to be able to catch and give an error for during compilation:

+ an action annotated @defaultonly that is used for an entry inside of `const entries`
+ an action annotated @tableonly that is assigned as the initial value of `default_action`

[action-use-violating-their-annotations.p4.txt](https://github.com/p4lang/p4c/files/2522065/action-use-violating-their-annotations.p4.txt)

jafingerhut:
Verified with latest p4c after the fix that it correctly flags errors where it should in my test program.  Thanks!

PR #1583, opened 2018-10-29
Fix for issue #1580

mbudiu-vmw:


frontends/p4/typeChecking/typeChecker.cpp
```diff
@@ -3300,15 +3310,25 @@ const IR::Node* TypeInference::postorder(IR::Property* prop) {
                 return prop;
             }
             auto at = type->to<IR::Type_Action>();
-            if (at->parameters->size() != 0)
+            if (at->parameters->size() != 0) {
                 typeError("Action for %1% has some unbound arguments", prop->value);
+                return prop;
```
cc10512: Would be nice to identify those arguments so that folks who do not debug the compiler can help themselves.

mbudiu-vmw: Actually I tried to create a test case that triggers this error, but I think other type checking cases will report an error first, so this never seems to happen. I think that this is defensive programming, so I will leave it in. I will commit a version with a better error message, but I could not create a test to exercise it.

Issue #1586, opened 2018-11-02
Incorrect "cannot infer bitwidth for integer-valued type parameter T" compiler error


jafingerhut:
This might be expected, and if so, feel free to close this issue.  Even if it is considered a bug, there is an easy workaround, so it could easily be considered a low priority bug.

The results I am seeing are for the latest version of p4c as of 2018-Oct-27.
```
bit<8> rand_val;

        // No error for this call to random:
        //random<bit<8>>(rand_val, 0, 255);
        

        // This line:
        //random(rand_val, 0, 255);

        // causes this error:
        
        // error-calling-random.p4(62): error: random: cannot infer bitwidth for integer-valued type parameter T
        //         random(rand_val, 0, 255);
        //         ^^^^^^^^^^^^^^^^^^^^^^^^
        // /home/jafinger/p4c/build/backends/p4test/p4include/v1model.p4(121)
        // extern void random<T>(out T result, in T lo, in T hi);
        //                    ^


        // This one:
        random((bit<8>) rand_val, 0, 255);

        // causes this error:

        // error-calling-random.p4(75): error: : Read-only value used for out/inout parameter result
        //         random((bit<8>) rand_val, 0, 255);
        //                ^^^^^^^^^^^^^^^^^
        // /home/jafinger/p4c/build/backends/p4test/p4include/v1model.p4(121)
        // extern void random<T>(out T result, in T lo, in T hi);
        //                             ^^^^^^
```

I have attached a full P4_16 + v1model arch sample program for this.
[error-calling-random.p4.txt](https://github.com/p4lang/p4c/files/2540861/error-calling-random.p4.txt)

mbudiu-vmw:
The read-only error is correct, the result of a cast is not a l-value, so you cannot write to it.

mbudiu-vmw:
The behavior is similar in other languages, e.g., C.

jafingerhut:
Makes sense.  Closing this issue since it is expected behavior.

mbudiu-vmw:
I will reopen the issue. the first error looks like a bug: the compiler should be able to infer the fact that T is `bit<8>`.

jafingerhut:
OK, changed title of issue to match that.

jafingerhut:
After the merged fix, there is no longer an incorrect "cannot infer bitwidth for integer-valued type parameter T" compiler error for the test program (with the line that caused the error uncommented).  Closing this issue.

PR #1587, opened 2018-11-02
Fix for issue #1586; added more test cases

mbudiu-vmw:
I think this is a bug in the design of C++.
If you get such an iterator: ` for (auto bound : binding)` over a map, the first and second fields should be const.


mbudiu-vmw:
This is a one-character fix in C++ (plus a few p4 files).
Shoudl be easy to review.

Issue #1569, opened 2018-10-20
p4c-graphs can create GraphViz dot files with illegal chars in node names


jafingerhut:
Found with latest p4c-graphs as of 2018-Oct-20.  The file tcp-options-parser.p4 in the command line below is this one: https://github.com/jafingerhut/p4-guide/blob/master/tcp-options-parser/tcp-options-parser.p4
```
p4c-graphs tcp-options-parser.p4
dot -Tpng ParserImpl.dot > ParserImpl.png
```
The file `ParserImpl.dot` is the one with slash characters (/) in some of the node names.

jafingerhut:
Tried out latest p4c-graphs after the fix, and looks good to me.  Closing this issue.

PR #1588, opened 2018-11-02
Fix for issue #1569

mbudiu-vmw:


mbudiu-vmw:
Tiny change, anyone can review it?

mbudiu-vmw:
Actually these identifiers are not visible to the outside; they are just internal graphviz node names.
So it is probably better to use the mangled unique names.
The node labels are the ones that are visible to the user.

PR #1589, opened 2018-11-02
a midend pass to convert mirror struct argument to a list expression

krishna-barefoot:


midend/expandMirror.cpp
```diff
@@ -0,0 +1,62 @@
...
+    }
+}
+
+void ExpandMirror::postorder(IR::MethodCallExpression* expr) {
+    auto mi = MethodInstance::resolve(expr, refMap, typeMap);
+    auto em = mi->to<P4::ExternMethod>();
+    if (!em) {
+      return;
+    }
+    if (em->originalExternType->name.name != "Mirror") {
```
mbudiu-vmw: I don't think you should hardwire such constants in a mid-end pass. The mid-end passes are to a large degree architecture-independent. If you want to do something like this you should consider using a policy class. See for example the ValidateTableProperties pass.

mihaibudiu (changes requested):

midend/expandMirror.cpp
krishna-barefoot: What is architecture dependent here? How is converting a one form of expression to other dependent on architecture?

mbudiu-vmw: The word "Mirror" does not appear in any of the P4 specs.
So it must be architecture-dependent.

krishna-barefoot: I haven't muddled too much in the front-end and so, I will followup with @cc10512 to make sure I understand your concern(s).  End of the day, it is a function called "foo" which I am changing the argument from one form to other which are orthogonal.

mbudiu-vmw: You should not hardwire constants in the compiler if they are not part of the language. The compiler should work for all legal input programs. "Mirror" is a constant which is not part of the input language.

krishna-barefoot: I get what you are saying. But, how does this make a legal program illegal?

ChrisDodd: It is ok for midend passes to be architecture dependent (many existing passes are); what they should not be is target depedent.  That said, it is desirable for them to be as general as possible, so as to be usable for multiple architectures.  As structured, this pass is closely tied to PSA (to support the PSA mirror extern), so it would be good to generalize the transform with a policy that decides when to expand things -- in the PSA case this would select argument 1 of any method call to the Mirror extern, but it would be nice if it could easily specify which arguments of which methods of which extern.

jafingerhut: Do you mean the PSA clone functionality?  There is no extern named Mirror in PSA.

krishna-barefoot: I am taking a function call (in this case "mirror") which is an identifier with an argument list and  match on that identifier and change the argument list in an orthogonal form.  And, the objection seems to be that identifier is not in the LRM as a string.  And, instead of hard-coding it, if I pass it as an argument, it is all OK. 

End of the day, the success of a language depends on the number of users and how it helps them - not how everything checks out.  My objective is to make a fix and move on.

jafingerhut: @krishna-barefoot I understand that this process can be frustrating, and you may be under tight time pressure, but please realize that p4c is intended to be used by other P4 targets that have no function called "Mirror" in them at all, that users might want to name something that of their own.  Does that make sense?

Issue #1560, opened 2018-10-18
P4_16 program that appears legal gives Compiler Bug in bmv2 back end


jafingerhut:
The error message is similar to that for issue https://github.com/p4lang/p4c/issues/1544, but the line number of code in the bmv2 back end is different, and the program exhibiting the bug does not use P4_16 functions at all, so perhaps a different root cause.  I do get the error with the latest version of p4c after the fix for #1544 was merged in.

Command and error message I see:
```
action-profile.p4(136): warning: Target does not support default_action for cIngress.t1 (due to action profiles)
    table t1 {
          ^^
In file: /home/jafinger/p4c/backends/bmv2/common/expression.cpp:636
Compiler Bug: action-profile.p4(176): hdr.ipv4.dstAddr[15:0];: Unhandled case
        meta.hash1 = hdr.ipv4.dstAddr[15:0];
                     ^^^^^^^^^^^^^^^^^^^^^^
```
[action-profile.p4.txt](https://github.com/p4lang/p4c/files/2493399/action-profile.p4.txt)

jafingerhut:
Adding a slightly larger example program, with some similar lines added in comments near the one that causes the problem.  If any of those commented-out lines replaces the one causing the problem, the compiler appears to create a correct BMv2 JSON output.  Perhaps the issue is something to do with the bit slice operator in particular places of the compiler's IR, that the bmv2 back end does not expect or handle yet.
[action-profile.p4.txt](https://github.com/p4lang/p4c/files/2493477/action-profile.p4.txt)

mbudiu-vmw:
@ChrisDodd @hanw: this is an issue with the LocalCopyPropagation pass. It propagates an expression into a table key which turns out to be too complicated for the back-end. This happens after the SimplifyKey pass in the midend, which can simplify such expressions. In this case the expression contains a bit slice.

jafingerhut:
Tried out the test program after the fix to the compiler was merged in.  No more compiler error, and a quick examination of the last midend pass P4_16 compiler output, and the BMv2 JSON, looks reasonable.  I didn't try testing it yet, but if any further problems detected will file a separate issue.  Closing this one.

PR #1590, opened 2018-11-02
Fix for issue #1560

mbudiu-vmw:
Interestingly, this also fixes a long-standing bug in the stf testing infrastructure.
When converting P4-14 programs with table keys (reads) that have masks we have to generate the correct key names.

### DIFF
--- a/backends/bmv2/bmv2stf.py
+++ b/backends/bmv2/bmv2stf.py
@@ -172,6 +172,13 @@ class TableKey(object):
         self.fields = OrderedDict()
     def append(self, name, type):
         self.fields[name] = type
+    def __str__(self):
+        result = ""
+        for f in self.fields.keys():
+            if result != "":
+                result += " "
+            result += f + ":" + self.fields[f]
+        return result
 
 class TableKeyInstance(object):
     def __init__(self, tableKey):
@@ -221,6 +228,7 @@ class TableKeyInstance(object):
             key = "$valid$"
             found = True
         if not found:
+            print(self.key.fields)
             raise Exception("Unexpected key field " + key)
         if self.key.fields[key] == "ternary":
             self.values[key] = self.makeMask(value)
@@ -315,7 +323,9 @@ class BMV2Table(object):
         self.key = TableKey()
         self.actions = {}
         for k in jsonTable["key"]:
-            name = k["target"]
+            name = k["name"]
+            if name is None:
+                name = k["target"]
             if isinstance(name, list):
                 name = ""
                 for t in k["target"]:
--- a/backends/bmv2/simple_switch/midend.cpp
+++ b/backends/bmv2/simple_switch/midend.cpp
@@ -89,6 +89,10 @@ SimpleSwitchMidEnd::SimpleSwitchMidEnd(CompilerOptions& options) : MidEnd(option
         new P4::ConstantFolding(&refMap, &typeMap),
         new P4::LocalCopyPropagation(&refMap, &typeMap),
         new P4::ConstantFolding(&refMap, &typeMap),
+        new P4::SimplifyKey(&refMap, &typeMap,
+                            new P4::OrPolicy(
+                                new P4::IsValid(&refMap, &typeMap),
+                                new P4::IsMask())),
         new P4::MoveDeclarations(),
         new P4::ValidateTableProperties({ "implementation",
                                           "size",
--- a/frontends/p4/fromv1.0/programStructure.cpp
+++ b/frontends/p4/fromv1.0/programStructure.cpp
@@ -799,10 +799,21 @@ ProgramStructure::convertTable(const IR::V1Table* table, cstring newName,
                 rt.name = p4lib.exactMatch.Id();
             auto ce = conv.convert(e);
 
+            const IR::Annotations* annos = IR::Annotations::empty;
+            // Generate a name annotation for the key if it contains a mask.
+            if (auto bin = e->to<IR::Mask>()) {
+                // This is a bit heuristic, but P4-14 does not allow arbitrary expressions for keys
+                cstring name;
+                if (bin->left->is<IR::Constant>())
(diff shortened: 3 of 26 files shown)


mbudiu-vmw:
@ChrisDodd @cc10512 @vgurevich : I have a question related to conversion from P4-14 to P4-16: if a table has a reads field with a mask, how do you name the key in the control-plane?

For example testdata/p4_14_samples/exact_match_mask1.p4:
```
table test1 {
    reads {
        data.f1 mask 0xff00ff : exact;
    }
...
```

The P4-14 to P4-16 converter generates a key with a `@name` annotation which looks wrong:
`hdr.data.f1 & 32w0xff00ff: exact @name("data.f1 & 16711935")`

What would you expect the name of this key to be? How should it be named, for example, in STF files?

antoninbas:
I remember trying to fix this a while ago. I would personally recommend using simply "data.f1".
do you know what name is generated for P4_16 slices?

mbudiu-vmw:
This PR does exactly this, but I am worried about ambiguities, like:
```
reads {
   a mask 0xF: exact;
   a mask 0xF0: exact;
}
```

mbudiu-vmw:
For slices they actually appear in the name: `@name("a[31:0]")`
This happens later, though. If the P4-14 converter adds a `@name`, it won't be changed.

antoninbas:
Do you think it would be possible to detect conflicts and resolve them with a simple `_X` suffix:
```
reads {
   a mask 0xF: exact;
   a mask 0xF0: exact;
}
```
would become `a_1` and `a_2` maybe?
This way we have a nice name in the common case and an easy way to handle collisions in the much rarer case where someone would apply different masks on the same field.

mbudiu-vmw:
We could write a pass just to detect such conflicts.
But we should have a rule that given a P4 program can tell exactly what the control-plane names for the key/reads fields will be, so people do not depend on the implementation.
Numbering them consecutively is definitely an option.

mbudiu-vmw:
OK, so I will fix the reference outputs in this PR.
Then we should file a separate issue about detecting (and resolving) ambiguous key field names.
We should probably add the rule in the P4 spec. Or at least in the P4Runtime spec.
But it would be nice to have a rule which does not even depend on which control-plane you are using.
@jafingerhut : what do you think?

mbudiu-vmw:
I have created issue https://github.com/p4lang/p4-spec/issues/691 and assigned to @antoninbas . I think he is the best person to propose a good solution to this. We will move the conversation about this there. I filed this as a spec issue.

jafingerhut:
Antonin's suggestion of adding uniquifying suffixes like "_1" and "_2" etc. if you need them seems reasonable to me.  It does seem pretty rare that they would be needed in production programs, but seems like a good idea to have a workable approach in case it does occur.

cc10512:
We should consider adding the mask as a suffix. For example, `a_0xF` and `a_0xF0`.

jafingerhut:
The syntax and semantics of the language don't prevent you from putting arbitrary arithmetic expressions there, though, do they?  Sure, you can make things nicer for a perhaps common case, but not sure how far it is worth going to create a new language to translate from expressions to legal identifier names.

cc10512:
Simple replacement of `[&|+-*/]` characters with `_` would suffice, no need to define a new language :)

mbudiu-vmw:
The question is most important for P4-14 programs.
For P4-16 programs we can also have a policy that gives an error if the user provides a key expression that is too complicated without providing a `@name` annotation.

hanw:
@mbudiu-vmw 
With the fix in this PR, frontend generates `@name("data.f1")` annotation for P4-14 program.
```
table test1 {
    reads {
        data.f1 mask 0xff00ff : exact;
    }
...
```

and it generates `@name("data.f116711935")`  for the equivalent P4-16 program
```
table test1 {
    keys = {
        data.f1 & 0xff00ff : exact;
    }
...
```
Which seems wrong.

Shall we fix the tableKeyNames pass to ignore the slices for P4-16 program? or not generating name annotation for P4-16 tables at all?

hanw:
If in P4-16, the frontend do not guess a name and use whatever is there, what is the use of the tableKeyNames pass? It does not generate any new information. In fact, the tableKeyNames pass makes midend/backend's job harder.  If midend/backend wants to generate the key name without the slices, it has to undo the name annotation added by tableKeyNames pass.

mbudiu-vmw:
We do this because other optimizations can change the key expression, e.g. By renaming variables.

jafingerhut:
This is really a separate issue, I suppose, but I don't mean it as a joke, but a very serious question about P4_16.  I am happy to file a separate issue on the p4lang/p4-spec repository if people think it has any hope of being adopted as a change to the language.

Expressions as keys of tables seem to cause a disproportionately large number of complexities in the design and implementation of P4, relative to any benefits that I can see.

Every single such thing _could_ be replaced by "new_temp_variable = <expression>;" followed by apply'ing the table with a key of "new_temp_variable", couldn't it?

Is there some significant benefit we are getting out of allowing expressions as table keys that is worth all of the hassle it causes?

jafingerhut:
`isValid` is much more restricted than "arbitrary expression".  P4_14 to P4_16 translation _could_ be done via introducing a temporary variable as described in my earlier comment.  The name of that temporary variable might even have well-defined rules for how it is constructed from the P4_14 expression contents, although it probably couldn't have dots inside its name.

I'm not going to push hard for such a change, in case it help put your mind at ease, but table key expressions seem to come up repeatedly as a thorn in the side of many implementation decisions.

mbudiu-vmw:
Table key expressions are trivial for the dataplane. The only issue is about the way they are exposed in the control-plane.  Whatever we allow, even if it is only as much as P4-14, we need some rules that define how keys are named in the control-plane. We also need some rules about what happens for the other kinds of expressions - reject, warning, require name, generate name?

Issue #1491, opened 2018-09-17
Automatically increment the version number on merges


mbudiu-vmw:
Today the compiler versions essentially never change, it would be nice to change the minor number automatically.

PR #1592, opened 2018-11-06
add support for versioning p4c backends

cc10512:
We add support for versioning different p4c backends independently of
each other. Currently we specify a global version in the root
CMakeLists.txt file, which will be reported by the p4c driver.
Individual backends get a configurable version string that is set at
configuration time based on macros derived from the global
version. However, each backend can customize the format of the
version, and see the example in the psa backend that hardcodes the
version to 0.0.1 and uses only the git SHA.

In honor of the language spec 1.1 release, I set the default version to 1.1.0-rc1.

Alternative solution to #1491.


cc10512:
Well, maybe just having all the reviewers look at it :) but thank you for your confidence!

PR #1594, opened 2018-11-07
add ior/xor operators to IR::Constant

zma-barefoot:


ir/expression.cpp
```diff
@@ -125,6 +125,20 @@ IR::Constant::operator&(const IR::Constant &c) const {
   return IR::Constant(v);
 }
 
+IR::Constant
```
mbudiu-vmw: all these operations ignore the types, but I guess they are patterned after the previous ones.

Issue #1595, opened 2018-11-08
p4c incorrectly optimizes away action assignment statements


jafingerhut:
With latest p4c as of 2018-Nov-07, I think it is pass 30 of the front end that incorrectly optimizes away the assignment statements in actions a1 through a4 of the attached P4_16 program, which seems like a bug to me.  Those are legitimate side effects that should be happening, and are visible in the output packet.  Perhaps the fact that bit slices are being used in the assignment left-hand side for later actions is causing the compiler to "believe" that the earlier assignments are killed?  Note that the left-hand side slices are assigning to _different_ bits of the field, so the earlier assignments also need to happen.

[switch-stmt-behavior1.p4.txt](https://github.com/p4lang/p4c/files/2559981/switch-stmt-behavior1.p4.txt)

mbudiu-vmw:
You are probably right, the compiler models the slices as writes to the whole field, so it thinks that the later ones kill the first ones.

jafingerhut:
If you want a program with corresponding STF test for this, let me know and I can cook one up.  I am guessing that fixing the bug should also be visible in the P4_16 program output by the front end, though.

jafingerhut:
A small corner of my mind is wondering whether you are considering doing def-use analysis on each of the W bits of a `bit<W>` independently :-)   Or perhaps only if a slice operator is used on a value at least once.

mbudiu-vmw:
The bug is not in the model of the slices, but in keeping track of what the write set is at each program point. The notion of program point is not consistent between the writes set computation and the read set computation. Haven't fixed it yet.

jafingerhut:
I have tried out the after-fix-merged version of p4c on my program exhibiting the problem, and the BMv2 JSON file and intermediate P4_16 source files output by the compiler look correct now.

PR #1596, opened 2018-11-10
Fix for issue #1595

mbudiu-vmw:
The fix is a one-liner (line 238 in simplifyDefUse), but I renamed a bunch of functions because TAGS were not useful to navigate the code at all.

### DIFF
--- a/frontends/p4/def_use.cpp
+++ b/frontends/p4/def_use.cpp
@@ -283,7 +283,7 @@ bool ProgramPoints::operator==(const ProgramPoints& other) const {
     return true;
 }
 
-Definitions* Definitions::join(const Definitions* other) const {
+Definitions* Definitions::joinDefinitions(const Definitions* other) const {
     auto result = new Definitions();
     for (auto d : other->definitions) {
         auto loc = d.first;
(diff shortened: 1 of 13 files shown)


mbudiu-vmw:
Anyone want to review this one-line fix (plus some code beautification)?

frontends/p4/def_use.cpp
```diff
@@ -817,8 +817,8 @@ bool ComputeWriteSet::preorder(const IR::AssignmentStatement* statement) {
 bool ComputeWriteSet::preorder(const IR::SwitchStatement* statement) {
     LOG3("CWS Visiting " << dbp(statement));
     visit(statement->expression);
-    auto locs = get(statement->expression);
-    auto defs = currentDefinitions->writes(getProgramPoint(), locs);
+    auto locs = getWrites(statement->expression);
+    auto defs = currentDefinitions->writes(getProgramPoint(statement->expression), locs);
```
ChrisDodd: What about this change?  It is not (just) a renaming -- there's a real change here.  Is it intentional?

ChrisDodd:
This would be much easier to review if you have separate commits for the rename change vs the bugfix change (when viewing the 'files changed' you can select single commits to review independently)

mbudiu-vmw:
Yes, the renaming was part of the debugging process. Next time I will commit the fix separately. You should trust that the c++ typechecker checks that the renaming is right.

PR #1598, opened 2018-11-13
add IR::Annotation to IR::ListExpression

hanw:
Passthrough annotation on FieldList to ListExpression to be handled in the backend.

### DIFF
--- a/frontends/p4/fromv1.0/converters.cpp
+++ b/frontends/p4/fromv1.0/converters.cpp
@@ -70,7 +70,7 @@ const IR::Node* ExpressionConverter::postorder(IR::Constant* expression) {
 
 const IR::Node* ExpressionConverter::postorder(IR::FieldList* fl) {
     // Field lists may contain other field lists
-    return new IR::ListExpression(fl->srcInfo, fl->fields);
+    return new IR::ListExpression(fl->srcInfo, fl->annotations, fl->fields);
 }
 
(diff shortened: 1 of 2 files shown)


mbudiu-vmw:
In general our expressions do not have annotations because the parser cannot insert them there and optimizations would not know what to do with them. For example, if a ListExpression is converted into a StructInitializerExpression, what should happen to the annotations?

hanw:
In this case, there is an annotation on the P4-14 FieldList object that we would like to pass-through to our backend. The P4-14 to 16 translation uses ListExpression to represent FieldList and this is probably the simplest way to get the annotation through. I am open to suggestions.

mbudiu-vmw:
It depends how the fieldlist is really used.
If we should have annotations on expressions they should be added to the base class Expression.
There is an understanding that these are solely for the compiler internal use - there is no way to write them out as P4, for example. They could be useful for other purposes, e.g., replacing hash-tables which are actually quite hard to use with IR nodes. 

What we don't have is a contract on what passes do to annotations on expressions. It would be very difficult to write one; the only guarantee we can have is that if a pass inserts annotations on expressions the next one will see them. But in the absence of a semantics of annotations it is not clear at all what should happen to them during compilation, when doing constant folding, common-subsexpression elimination, copy propagation, etc.

If you expect an annotation to make it from the front-end all the way to the back-end you are probably making many assumptions about what the implementation actually is.

Issue #117, opened 2016-10-26
[bmv2 backend] intrinsic meta data


smolkaj:
The field name `intrinsic_metadata` seems to be reserved or have some special meaning inside the meta data struct. Specifically, the program

``` C
#include <core.p4>
#include <v1model.p4>
typedef standard_metadata_t std_m;

header some_t { }

struct H { };
struct M {
    some_t intrinsic_metadata;
}

parser ParserI(packet_in pk, out H hdr, inout M meta, inout std_m smeta) {
    state start { transition accept; }
}

control VerifyChecksumI(in H hdr, inout M meta) {
    apply { }
}

control IngressI(inout H hdr, inout M meta, inout std_m smeta) {
    apply { }
}

control EgressI(inout H hdr, inout M meta, inout std_m smeta) {
    apply { }
}

control ComputeChecksumI(inout H hdr, inout M meta) {
    apply { }
}
control DeparserI(packet_out b, in H hdr) {
    apply { }
}

V1Switch(ParserI(), VerifyChecksumI(), IngressI(), EgressI(), ComputeChecksumI(),
         DeparserI()) main;
```

raises the exception

```
terminate called after throwing an instance of 'Util::CompilerBug'
  what():  COMPILER BUG: ../backends/bmv2/jsonconverter.cpp:1843
Expected a struct type
```

but if I rename the field `intrinsic_metadata`, the code compiles fine. Is this the intended behavior?

PR #1600, opened 2018-11-13
Fix for spec issue 692

mbudiu-vmw:
This is not part of the spec, but looks like a good idea:
https://github.com/p4lang/p4-spec/issues/692


frontends/p4/validateParsedProgram.cpp
```diff
@@ -180,6 +180,13 @@ void ValidateParsedProgram::postorder(const IR::SwitchStatement* statement) {
     auto inAction = findContext<IR::P4Action>();
     if (inAction != nullptr)
         ::error("%1%: switch statements not allowed in actions", statement);
+    bool defaultFound = false;
+    for (auto c : statement->cases) {
+        if (defaultFound)
```
cc10512: Can we please ensure that we print the error only once per select statement? And we also print the offending labels.

mbudiu-vmw: We are printing the offending label. Do you also want to see the default label?

mbudiu-vmw: The test case stderr shows the output.

cc10512: My bad on the label. I looked at the travis run and it shows:
```error: SwitchCase: label following default label
error: SwitchCase: label following default label
error: SwitchCase: label following default label```

I interpreted SwitchCase as the message.
Also, there are no line numbers printed.

Comment (mbudiu-vmw) [reply frontends/p4/validateParsedProgram.cpp]: Yes, the line number is the one in parentheses.
The error function does this automatically.
```
default-switch.p4(30): error: SwitchCase: label following default label
            b: { return; }
            ^^^^^^^^^^^^^^
default-switch.p4(31): error: SwitchCase: label following default label
            c:
            ^
```

Comment (cc10512) [reply frontends/p4/validateParsedProgram.cpp]: I see the following output on Travis:
```
978/1166 Test #117: bmv2/testdata/p4_14_samples/04-GatewayDefault.p4 ..............................***Failed    0.46 sec
Error compiling
error: SwitchCase: label following default label
error: SwitchCase: label following default label
*** Test failed
```

File: frontends/p4/validateParsedProgram.cpp [hunk 2]
```diff
@@ -175,11 +175,21 @@ void ValidateParsedProgram::postorder(const IR::EntriesList* l) {
...
         ::error("%1%: switch statements not allowed in actions", statement);
+    bool defaultFound = false;
+    for (auto c : statement->cases) {
+        if (defaultFound) {
+            ::error("%1%: label following default label", c);
+            break;
+        }
+        if (c->label->is<IR::DefaultExpression>())
+            defaultFound = true;
+    }
```
cc10512: How about the following:
```
    bool defaultFound = false;
    std::vector<IR::SwitchCase *> labelsAfterDefault;
    for (auto c : statement->cases) {
        if (defaultFound)
            labelsAfterDefault.push_back(c);
        if (c->label->is<IR::DefaultExpression>())
            defaultFound = true;
    }
   // since this may break existing programs, we should throw an error only if in strict compilation mode
   // otherwise, print a warning and let folks know that it may become an error
    if (options.isStrict())
         ::error("%1%: Switch statements require the default label to be the last one. The following labels are listed following the default: %2%", statement, labelsAfterDefault);
    else
         ::warning("%1%: Switch statements will require the default label to be the last one. This behavior will be enforced in a future version of the compiler. The following labels are listed following the default: %2%", statement, labelsAfterDefault);
```

It will of course, require adding the strict option, but I do believe that is better than breaking existing programs outside the p4lang/p4c test suite.

mbudiu-vmw: We can just make this a warning.
We have -Werror to be strict.

cc10512: That's a good idea. And we then we can turn it into an error later when people had a chance to migrate their code.

mbudiu-vmw:
@cc10512 : I have also fixed a bug in the p4-14 translation, so now all tests pass.

PR #1601, opened 2018-11-14
add hook to support custom expression converter visitor function

hanw:
Added a hook in P4-14 to P4-16 translation to allow customizing ExpressionConverter visitor function. Currently, it is only used to overload IR::FieldList visitor function.

### DIFF
--- a/frontends/p4/fromv1.0/converters.cpp
+++ b/frontends/p4/fromv1.0/converters.cpp
@@ -70,7 +70,7 @@ const IR::Node* ExpressionConverter::postorder(IR::Constant* expression) {
 
 const IR::Node* ExpressionConverter::postorder(IR::FieldList* fl) {
     // Field lists may contain other field lists
-    if (auto func = get(typeid(fl))) {
+    if (auto func = get(std::type_index(typeid(*fl)).name())) {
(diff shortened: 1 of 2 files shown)


mbudiu-vmw:
So this is used if you are extending the P4-14 IR classes with some additional ones?
I think that the right thing to do is actually to subclass the converter too and use that one in the frontend. Then you don't need a hook, just a factory for the converter.

frontends/p4/fromv1.0/converters.cpp
```diff
@@ -240,6 +242,20 @@ const IR::Node* ExpressionConverter::postorder(IR::Neq *neq) {
         return new IR::BoolLiteral(neq->srcInfo, true);  // everything else is true
 }
 
+std::map<cstring, ExpressionConverter::funcType>* ExpressionConverter::cvtForType = nullptr;
+
+void ExpressionConverter::addConverter(cstring type, ExpressionConverter::funcType cvt) {
+    static std::map<cstring, ExpressionConverter::funcType> tbl;
```
ChrisDodd: You could use `std::type_index` directly as the key type, rather than converting to a string.

PR #1602, opened 2018-11-16
Support lookahead in ebpf backend

mbudiu-vmw:
There is also a new midend pass which may prove useful in other places.


mbudiu-vmw:
Anyone have a little bandwidth to check out this PR?

PR #1605, opened 2018-11-21
Fix for issue #562

mbudiu-vmw:
I am not 100% that this fix is foolproof - but it may be better than nothing for now.
I will have to think whether it can be tripped up by some crafty programs that use the nested structs for other purposes. However, the only bad thing that can happen is that we trigger more BUGs because we rewrite the programs incorrectly. But such programs weren't supported on bmv2 anyway, so the situation won't be any worse than it is now.

### DIFF
--- a/midend/flattenInterfaceStructs.cpp
+++ b/midend/flattenInterfaceStructs.cpp
@@ -1,0 +1,157 @@
+/*
+Copyright 2018 VMware, Inc.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+#include "flattenInterfaceStructs.h"
+
+namespace P4 {
+
+static void flatten(const P4::TypeMap* typeMap,
+                    cstring prefix,
+                    const IR::Type* type,
+                    IR::IndexedVector<IR::StructField> *fields,
+                    Replacement* replacement) {
+    if (auto st = type->to<IR::Type_Struct>()) {
+        for (auto f : st->fields)
+            flatten(typeMap, prefix + "." + f->name, f->type, fields, replacement);
+        return;
+    }
+    cstring fieldName = prefix.replace(".", "_") +
+                        cstring::to_cstring(replacement->fieldNameRemap.size());
+    replacement->fieldNameRemap.emplace(prefix, fieldName);
+    fields->push_back(new IR::StructField(IR::ID(fieldName), type));
+}
+
+static Replacement* flatten(
+    const P4::TypeMap* typeMap, cstring name, const IR::Type_Struct* type) {
+    auto result = new Replacement();
+    auto vec = new IR::IndexedVector<IR::StructField>();
+    flatten(typeMap, "", type, vec, result);
+    result->type = new IR::Type_Struct(IR::ID(name), IR::Annotations::empty, *vec);
+    return result;
+}
+
+static const IR::Type_Struct* isNestedStruct(const P4::TypeMap* typeMap, const IR::Type* type) {
+    if (auto st = type->to<IR::Type_Struct>()) {
+        for (auto f : st->fields) {
+            auto ft = typeMap->getType(f, true);
+            if (ft->is<IR::Type_Struct>())
+                return st;
+        }
+    }
+    return nullptr;
+}
+
+void NestedStructMap::createReplacement(const IR::Type_Struct* type) {
+    auto repl = ::get(replacement, type);
+    if (repl != nullptr)
+        return;
+    cstring newName = refMap->newName(type->name);
+    repl = flatten(typeMap, newName, type);
+    LOG3("Replacement for " << type << " is " << repl);
+    replacement.emplace(type, repl);
+}
+
+bool FindTypesToReplace::preorder(const IR::Declaration_Instance* inst) {
+    auto type = map->typeMap->getTypeType(inst->type, true);
+    auto ts = type->to<IR::Type_SpecializedCanonical>();
+    if (ts == nullptr)
+        return false;
+    if (!ts->baseType->is<IR::Type_Package>())
+        return false;
+    for (auto t : *ts->arguments) {
+        if (auto st = isNestedStruct(map->typeMap, t))
+            map->createReplacement(st);
+    }
+    return false;
+}
+
+/////////////////////////////////
+
+const IR::Node* ReplaceStructs::preorder(IR::P4Program* program) {
+    if (replacementMap->empty()) {
+        // nothing to do
+        prune();
+    }
+    return program;
+}
+
+const IR::Node* ReplaceStructs::postorder(IR::Type_Struct* type) {
+    auto canon = replacementMap->typeMap->getTypeType(getOriginal(), true);
+    auto repl = replacementMap->getReplacement(canon);
+    if (repl != nullptr)
+        return repl->type;
+    return type;
+}
+
+const IR::Node* ReplaceStructs::postorder(IR::Type_Name* type) {
+    auto canon = replacementMap->typeMap->getTypeType(getOriginal(), true);
+    auto repl = replacementMap->getReplacement(canon);
+    if (repl != nullptr)
+        return repl->type->getP4Type();
+    return type;
+}
+
+const IR::Node* ReplaceStructs::postorder(IR::Member* expression) {
+    // Find out if this applies to one of the parameters that are being replaced.
+    const IR::Expression* e = expression;
+    cstring prefix = "";
+    while (auto mem = e->to<IR::Member>()) {
+        e = mem->expr;
+        prefix = cstring(".") + mem->member + prefix;
+    }
+    auto pe = e->to<IR::PathExpression>();
+    if (pe == nullptr)
+        return expression;
+    auto decl = replacementMap->refMap->getDeclaration(pe->path, true);
+    auto param = decl->to<IR::Parameter>();
+    if (param == nullptr)
+        return expression;
+    auto repl = ::get(toReplace, param);
+    if (repl == nullptr)
+        return expression;
+    auto newFieldName = ::get(repl->fieldNameRemap, prefix);
+    if (newFieldName.isNullOrEmpty())
+        return expression;
+    auto result = new IR::Member(pe, newFieldName);
+    LOG3("Replacing " << expression << " with " << result);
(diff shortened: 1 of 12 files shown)


midend/flattenInterfaceStructs.h
```diff
@@ -0,0 +1,144 @@
...
+#include "ir/ir.h"
+#include "frontends/p4/typeChecking/typeChecker.h"
+
+namespace P4 {
+
+/**
+   Describes how a type is replaced: the new type to
+   replace it and how each field is renamed.
+*/
+struct Replacement : public IHasDbPrint {
```
cc10512: You may want to find a better name for "Replacement". It is too generic and not nested in a specialized class.

midend/flattenInterfaceStructs.h
```diff
@@ -0,0 +1,144 @@
...
+    // The keys are strings of the form field.field.field.  If
```
cc10512: If I read this correctly it is supporting a single level of nesting ...
Even if you impose that limit at this time for the implementation, your data structures should be designed to support additional levels of nesting.

midend/flattenInterfaceStructs.h
```diff
@@ -0,0 +1,144 @@
...
+
+/**
+ * Maintains a map between an original struct type and its
+ * replacement.
+ */
+struct NestedStructMap {
+    P4::ReferenceMap* refMap;
+    P4::TypeMap* typeMap;
+
+    ordered_map<const IR::Type*, Replacement*> replacement;
```
cc10512: Are you keeping the replacement per field or per struct? From your comment on replacement, I assumed per field.

midend/flattenInterfaceStructs.h
```diff
@@ -0,0 +1,144 @@
...
+control proto<V>(inout V arg);
+package top<V>(proto<V> ctrl);
+
+top<T>(c()) main;
+
+This is transformed into:
+
+struct S { bit b; }
+struct T { S s; }
+struct Tflat {
```
cc10512: What is the reason for not replacing the nested structs in place? By generating a new type and replacing it in the program, you affect the control plane code generation (when done past midend which some backends may do) and make debugging a lot harder.

midend/flattenInterfaceStructs.h
mbudiu-vmw: No, this works to arbitrary levels of nesting. There are multiple dots in the keys.
The test case is a two-level nested struct.

midend/flattenInterfaceStructs.h
mbudiu-vmw: For each type that is to be replaced (a deeply nested struct) we create a Replacement. The replacement says how each field is renamed, and also it has a pointer to the new flattened struct. When I rename Replacement I will give that field a clearer name, now it's called type.

midend/flattenInterfaceStructs.h
mbudiu-vmw: You are really asking why I am not keeping the same type name. This is really a new type, it is not the same as the old type, so I liked the idea of having a new name. I don't think there isn't any reason we can't keep the old name. I will try the patch that way.

I am not aware of any place where the control-plane API depends on type names. I.e., I don't know if you alpha-rename types in a P4 program whether the control-plane API could change at all. But this is indeed not part of the P4 contract, so it is conceivable that some APIs could depend on type names.

The fact that the type structure changes is actually a much more disruptive change than the renaming of the type. I hope that is not going to break anything. That's one reason it took me so long to implement this, and I am still not sure that this is 100% correct (i.e., there are no source programs where this transformation could generate an output program that is not equivalent).

mbudiu-vmw:
One problem with the PR is that it does not show the output of the transformation, since we only save the files produced by p4test, not the files produced by p4c-bm2-ss after the midend.

midend/flattenInterfaceStructs.h
mbudiu-vmw: Actually, the transformation is not enough, some of the failing test cases show instances where it fails. So more work is needed.

cc10512:
I think the idea of renaming fields (aliasing really) is good also for handling overlays. So if you could work that out generically in your design, it will be a winner!

midend/flattenInterfaceStructs.h
jafingerhut: > I am not aware of any place where the control-plane API depends on type names. I.e., I don't know if you alpha-rename types in a P4 program whether the control-plane API could change at all. But this is indeed not part of the P4 contract, so it is conceivable that some APIs could depend on type names.

Part of the work done on the P4Runtime API to allow arbitrary types to be used as Digest extern message parameters, and in the future other places as well (e.g. hopefully action parameters, table keys, Register array element contents), uses type names in the control plane API to name those types.  I do not know the details of these changes and which pass they might appear in the compiler to know whether that could affect the output files used by the P4Runtime API, but wanted to address your comment, so you were aware of this potential issue.

hesingh: The PR is developing a new midend pass.  The P4Runtime API are generated by a backend once the backend code has completed running the frontend module.  See frontend.run(..,..) in p4c/backends/bmv2/simple_switch/main.cpp.  Thus, the new flatten midend pass of this PR (or any other midend pass) is run after the P4Runtime API are generated.

mbudiu-vmw: Yes, but there are compilers out there that do the API generation after mid-end.

mbudiu-vmw:
@cc10512 : I have updated the PR with another commit.
It was required to handle calls like `recirculate(meta.field)` which take struct-valued arguments.

midend/flattenInterfaceStructs.h
mbudiu-vmw: The ebpf back-end generates the control-plane API in the back-end. It does not use P4Runtime.
It does not need this pass, though.
@cc10512 may have other examples, he is the one who said this.
I am not working on any proprietary compilers for P4, all the code I write is on github, so there is nothing else that I know about this subject.

hesingh: > Yes, but there are compilers out there that do the API generation after mid-end.

Hmm, Antonin sent out an email to p4-dev in a discussion with me saying "it was a careful decision where to insert control-plane api generation."   See

http://lists.p4.org/pipermail/p4-dev_lists.p4.org/2018-September/003636.html

It would be good to understand why compilers are diverging.  I see even the p4test backend generates P4Runtime API after frontend.run() and before midend.process() is called.

midend/flattenInterfaceStructs.cpp
```diff
@@ -0,0 +1,187 @@
...
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+#include "flattenInterfaceStructs.h"
+
+namespace P4 {
+
+void StructTypeReplacement::flatten(const P4::TypeMap* typeMap,
```
cc10512: 👍

midend/flattenInterfaceStructs.cpp
```diff
@@ -0,0 +1,187 @@
...
+        return expression;
+    auto newFieldName = ::get(repl->fieldNameRemap, prefix);
+    const IR::Expression* result;
+    if (newFieldName.isNullOrEmpty()) {
+        // Prefix is a reference to a field of the original struct whose
+        // type is actually a struct itself.  We need to replace the field
+        // with a struct initializer expression.  (This won't work if the
+        // field is being used as a left-value.  Hopefully, all such uses
+        // of a struct-valued field as a left-value have been already
+        // replaced by the NestedStructs pass.)
```
cc10512: Could you BUG_CHECK this condition?

midend/flattenInterfaceStructs.h
```diff
@@ -0,0 +1,189 @@
...
+};
+
+class FlattenInterfaceStructs final : public PassManager {
+ public:
+    FlattenInterfaceStructs(ReferenceMap* refMap, TypeMap* typeMap) {
+        auto sm = new NestedStructMap(refMap, typeMap);
+        passes.push_back(new TypeChecking(refMap, typeMap));
+        passes.push_back(new FindTypesToReplace(sm));
+        passes.push_back(new ReplaceStructs(sm));
+        passes.push_back(new ClearTypeMap(typeMap));
```
cc10512: I assume you're clearing the typeMap because you've replaced the types. Which is the next pass that rebuilds it? I guess a different way of asking the question, aren't any invariants that are broken by where is `FlattenInterfaceStructs` inserted in the pass order?

midend/flattenInterfaceStructs.h
mbudiu-vmw: Compilers are not diverging, they support different architectures with different constraints. When you run eBPF locally in the Linux kernel often it does not make sense to have a remote control-plane that communicates using RPC. And you need a local control-plane anyway. 

The P4 spec does not mandate a specific compiler architecture; it only requires some entities to be exposed to the control-plane, i.e., tables and actions. As long as these are done properly it should not matter where or how the control-plane API is being generated. P4Runtime is *a* control-plane API, but not the only one that can work with P4 programs. That's exactly why P4Runtime is a separate spec. 

In our implementation we are indeed careful when we add passes to the front-end: these passes should not cause any changes in the control-plane API that will be generated. No such contract is promised by mid-end passes, but if you construct your mid-end carefully you can preserve this property after the mid-end as well.

midend/flattenInterfaceStructs.h
mbudiu-vmw: The contract enforced by the typeMap is that if an IR node is in a key in typeMap then the associated type is correct. This pass changes the types of some nodes, so it is easiest to clear the whole typeMap. In general, a pass runs type-checking if it needs to be sure that all nodes have types in the typeMap, but the type-checker skips over IR subtrees that are already in the typeMap. 

Due to the structure of our visitors, new IR nodes are created all the time, and it is quite frequent for nodes not to be in the typeMap, so typeChecking is run many times. It works as an integrity checker of the compiler. The alternative would be for each pass to carefully insert types for all new nodes created, but that is much more complicated than inserting nodes and leaving the type-checker compute their types.

midend/flattenInterfaceStructs.h
hesingh: > Compilers are not diverging, they support different architectures with different constraints. When you run eBPF locally in the Linux kernel often it does not make sense to have a remote control-plane that communicates using RPC. And you need a local control-plane anyway.
> 
> The P4 spec does not mandate a specific compiler architecture; it only requires some entities to be exposed to the control-plane, i.e., tables and actions. As long as these are done properly it should not matter where or how the control-plane API is being generated. P4Runtime is _a_ control-plane API, but not the only one that can work with P4 programs. That's exactly why P4Runtime is a separate spec.
> 
> In our implementation we are indeed careful when we add passes to the front-end: these passes should not cause any changes in the control-plane API that will be generated. No such contract is promised by mid-end passes, but if you construct your mid-end carefully you can preserve this property after the mid-end as well.

I was only going why what Antonin said that it's a hard rule where to insert P4Runtime generation.    However, I understand your perspective as well.

PR #1614, opened 2018-11-29
Added support for free-form annotations

liujed:
This adds support for free-form annotations, wherein annotation bodies may be any sequence of tokens containing balanced parentheses. At the same time, this introduces annotation checking, to ensure predefined annotations are used correctly. Backwards compatibility with P4₁₄-style pragmas is also added as an experimental feature.

On the initial parsing pass, annotation bodies are parsed as a sequence of tokens in a new field, `Annotation::body`. After the initial parse, `Annotation::expr` and `Annotation::kv` are empty. P4₁₄-style pragmas are translated by the parser into P4₁₆ annotations: `@pragma name body` is translated into `@name(body)`.

Early in the front end, a rewriting pass parses the bodies of spec-defined annotations, converting them into an expression list (which populates `Annotation::expr`) or a key-value list (`Annotation::kv`). This has the new side-effect of rejecting annotations with mismatched bodies. For instance, having an expression list where a key-value list is expected, or having a no-argument annotation with a non-empty body.

As of v1.0.0-draft of the spec, annotations are as follows:
  * `@optional` with an empty body (§18.2.1)
  * `@tableonly` with an empty body (§18.2.2)
  * `@defaultonly` with an empty body (§18.2.2)
  * `@name` with a string literal (§18.2.3)
  * `@hidden` with an empty body (§18.2.3)
  * `@atomic` with an empty body (§18.2.4)
  * `@pkginfo` with a key-value list (§18.3)

The front end also supports some annotations used by the tests, but not mentioned in the spec:
  * `@length` takes an expression, and
  * `@deprecated` takes a string literal; this is a candidate language feature: p4lang/p4-spec#655.

Unrecognized annotations are left untouched. For any target-specific annotations, back ends may do a similar rewriting pass as needed. For instance, the BMv2 back end additionally supports the following annotations:

  * `@metadata` has no body,
  * `@alias` takes a string literal,
  * `@priority` takes an integer constant.

The P4 Runtime serializer uses a `@controller_header` annotation that takes a string literal and an `@id` annotation that takes an integer constant. The GTest uses `@diagnostic` (takes an expression list) and `@my_anno` (takes a string literal).


liujed:
For context, this implements https://github.com/p4lang/p4-spec/issues/695.
@mbudiu-vmw: requesting a review from you, since I touched BMv2.
@antoninbas: requesting a review from you, since I touched the P4Runtime serializer.

control-plane/p4RuntimeSerializer.cpp
```diff
@@ -657,6 +658,15 @@ getMatchFields(const IR::P4Table* table, ReferenceMap* refMap, TypeMap* typeMap)
     return matchFields;
 }
 
+/// Parses P4Runtime-specific annotations.
+class ParseAnnotations : public P4::ParseAnnotations {
+ public:
+    ParseAnnotations() : P4::ParseAnnotations("P4Runtime") { }
+    virtual void postorder(IR::Annotation* annotation) override {
+        PARSE("controller_header", StringLiteral)
```
antoninbas: This is just checking that the `@controller_header` "payload" is a string and transforming it to an IR StringLiteral, correct? The code would work the same without it?

If this sanity checks the annotation payload, can we remove this code later on:
```
auto nameConstant = controllerAnnotation->expr[0]->to<IR::StringLiteral>();
        if (nameConstant == nullptr) {
            ::error("@controller_header should be a string for declaration %1%", type);
            return;
        }
```

Also, why not do the other annotations, such as `@id`?

frontends/parsers/p4/p4parser.ypp
```diff
@@ -316,8 +366,127 @@ annotations
...
-    | "@" name "(" expressionList ")" { $$ = new IR::Annotation(@1, *$2, *$4); }
-    | "@" name "(" kvList ")"         { $$ = new IR::Annotation(@1, *$2, *$4); }
+    | "@" name "(" annotationBody ")" { $$ = new IR::Annotation(@1, *$2, *$4); }
+    ;
+
+annotationBody
+    : /* empty */
+        { $$ = new IR::Vector<IR::AnnotationToken>; }
+    | "(" annotationBody ")"
+        { $$ = new IR::Vector<IR::AnnotationToken>;
```
ChrisDodd: This only accepts nested `(`..`)` at the very beginning of the annotationBody.  So something like `@anot( func(7) )` will cause a syntax error.  You need to change this rule to `annotationBody: annotationBody "(" annotationBody ")"`

frontends/parsers/parserDriver.cpp
```diff
@@ -156,6 +164,61 @@ P4ParserDriver::parse(FILE* in, const char* sourceFile,
     return parse(inputStream.get(), sourceFile, sourceLine);
 }
 
+template<typename T> const T*
+P4ParserDriver::parse(P4AnnotationLexer::Type type,
+                      Util::SourceInfo srcInfo,
+                      IR::Vector<IR::AnnotationToken>* body) {
+    if (Log::verbose())
+        std::cout << "Parsing P4-16 annotation " << srcInfo << std::endl;
```
ChrisDodd: Generally prefer using `LOG` macros to checking Log::verbose() directly.  This will generate lots of cruft for people not interested in it.

frontends/p4/parseAnnotations.cpp
```diff
@@ -0,0 +1,58 @@
...
+    return true;
+}
+
+void ParseAnnotations::postorder(IR::Annotation* annotation) {
+    // @tableonly, @defaultonly, @hidden, @atomic, and @optional have no
+    // arguments.
+    PARSE_NO_ARGS(IR::Annotation::tableOnlyAnnotation)
+    PARSE_NO_ARGS(IR::Annotation::defaultOnlyAnnotation)
+    PARSE_NO_ARGS(IR::Annotation::hiddenAnnotation)
+    PARSE_NO_ARGS(IR::Annotation::atomicAnnotation)
```
ChrisDodd: The linear search through annotation names bothers me a bit -- it would be better if this could be done with a `std::map` or even `std::unordered_map`.  A minor nit.

liujed: Agreed. This was the path of least resistance, since I'm not as comfortable in C++ as I am in Java. But now that you've brought this up, I'll fix this.

frontends/parsers/parserDriver.cpp
liujed: I've rewritten this to use `LOG3`, and rewritten the pre-existing checks to `Log::verbose()` to use `LOG1`.

frontends/parsers/p4/p4parser.ypp
liujed: Thanks, good catch. Fixed.

control-plane/p4RuntimeSerializer.cpp
liujed: Done. Added support for `@id` and `@hidden`. Removed the sanity check above and one for `@id`.

liujed:
@mbudiu-vmw: Sure, this change allows annotations that are not constrained by the P4 syntax. For instance, a programmer might want to add assertion annotations to their code. Previously, they had to write things like `@assert(!reach(table) || hdr.ethernet.ether_type == 0x0800)`. With this change, they can instead write `@assert(reach(table) ⇒ hdr.ethernet.ether_type == 0x0800)`, which I find more natural and easier to read.

This is a simple example, but I think opening up the syntax of annotation bodies makes annotations a lot more interesting: implementers are now free to have a DSL embedded in their annotations.

mbudiu-vmw:
Can you add a P4 test file that contains such a free-form annotation?
Do you want to own the spec issue and bring it to the discussion during one of our design meetings?

mbudiu-vmw:
The `@assert` example is probably not the best. assert should probably be a function, not an annotation.
But I can imagine that there are other useful cases.
We should probably stick to ASCII in programs, though.

frontends/parsers/p4/p4parser.ypp
```diff
@@ -316,8 +366,127 @@ annotations
...
+                  auto constInfo = new IR::UnparsedConstant(@2, uc.skip, uc.base, uc.hasWidth);
+                  $$->push_back(new IR::AnnotationToken(@2, $2.type, $2.text, constInfo));
+              }
+              break;
+          default:
+              $$->push_back(new IR::AnnotationToken(@2, $2.type, $2.text));
+          } }
+    ;
+
+annotationToken
```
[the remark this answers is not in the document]
liujed: Only when new lexer tokens are added, which shouldn't happen all that often.

test/gtest/helpers.cpp
```diff
@@ -126,6 +127,16 @@ P4CTestEnvironment::P4CTestEnvironment() {
 
 namespace Test {
 
+class ParseAnnotations : public P4::ParseAnnotations {
+ public:
+    ParseAnnotations() : P4::ParseAnnotations("FrontendTest") { }
+    void postorder(IR::Annotation* annotation) override {
+        PARSE_EXPRESSION_LIST("diagnostic")
+        PARSE("id", Constant)
```
antoninbas: out of curiosity does that serve any purpose since the test code itself doesn't access that annotation directly (and any invalid annotation - not a integer literal  - would be caught in the P4Runtime code)?

frontends/parsers/p4/p4parser.ypp
```diff
@@ -255,11 +290,26 @@ typedef const IR::Type ConstType;
    track of which identifiers represent types, and which represent namespaces.
 */
 
+// Simulate multiple start symbols to allow the parser to be reused for
```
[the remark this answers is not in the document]
liujed: I don't think these start symbols add all that much complexity, but I agree that we might not want to publish this detail in the spec. This doesn't seem to be a showstopper, though. I don't see a verbatim copy of `p4parser.ypp` in v1.1.0 of the spec. Instead, Appendix H has what looks like a processed copy. It doesn't seem difficult to omit the `start` and `fragment` productions as part of this processing. The spec version would continue to have `program` as its start symbol (which appears to be renamed `p4program`).

frontends/parsers/p4/p4AnnotationLexer.hpp
```diff
@@ -0,0 +1,35 @@
...
+    };
+
+ private:
+    Type type;
+    IR::Vector<IR::AnnotationToken> body;
+    bool needStart;
+    IR::Vector<IR::AnnotationToken>::const_iterator it;
+
+ public:
+    explicit P4AnnotationLexer(Type type,
```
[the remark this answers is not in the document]
liujed: Fixed.

ir/base.def
```diff
@@ -215,6 +238,8 @@ class Annotation {
     static const cstring hiddenAnnotation;  /// Object should not be exposed to the control-plane.
     static const cstring lengthAnnotation;  /// P4-14 annotation for varbit fields.
     static const cstring optionalAnnotation;  /// Optional parameter annotation
+    static const cstring pkginfoAnnotation;  /// Package documentation annotation.
+    static const cstring deprecatedAnnotation;  /// Deprecation annotation.
```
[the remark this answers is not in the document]
liujed: Done in 208490b.

test/gtest/helpers.cpp
liujed: Good point. I didn't fully understand what was going on in the test. Fixed in a371db8.

frontends/parsers/p4/p4AnnotationLexer.cpp
```diff
@@ -0,0 +1,50 @@
...
+        token->constInfo->base,
+        token->constInfo->hasWidth
+    };
+
+    return result;
+}
+
+P4AnnotationLexer::Token P4AnnotationLexer::yylex(P4::P4ParserDriver& driver) {
+    // Driver not used here. Suppress compiler warning.
+    (void) driver;
```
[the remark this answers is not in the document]
liujed: Thanks. Done in e4e3d70.

liujed:
@mbudiu-vmw: Simple test added in fa7fecb.

frontends/p4/parseAnnotations.cpp
liujed: Fixed in 036a0a1.

testdata/p4_16_samples/free-form-annotation.p4
```diff
@@ -0,0 +1,64 @@
+/*
```
mbudiu-vmw: you need to also commit the result files produced by running this test.
They should be in in p4_16_samples_outputs.

frontends/parsers/p4/p4parser.ypp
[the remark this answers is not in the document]
mbudiu-vmw: Maybe you can add a comment in the lexer file to let people know to look here as well.

backends/bmv2/simple_switch/simpleSwitch.h
```diff
@@ -26,6 +26,7 @@ limitations under the License.
 #include "frontends/p4/unusedDeclarations.h"
 #include "midend/convertEnums.h"
 #include "backends/bmv2/common/action.h"
+#include "backends/bmv2/common/annotations.h"
```
cc10512: Why not include only in .cpp?

testdata/p4_16_samples/free-form-annotation.p4
liujed: Done in afadac0.

frontends/parsers/p4/p4parser.ypp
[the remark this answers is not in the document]
liujed: Done in 4ce54f8.

backends/bmv2/simple_switch/simpleSwitch.h
liujed: Fixed in e0290d7.

cc10512:
@mbudiu-vmw can you please take another look at this PR and approve?

frontends/parsers/p4/p4parser.ypp
```diff
@@ -316,8 +374,121 @@ annotations
 
 annotation
     : "@" name                        { $$ = new IR::Annotation(@1, *$2, {}); }
-    | "@" name "(" expressionList ")" { $$ = new IR::Annotation(@1, *$2, *$4); }
-    | "@" name "(" kvList ")"         { $$ = new IR::Annotation(@1, *$2, *$4); }
+    | "@" name "(" annotationBody ")" { $$ = new IR::Annotation(@1, *$2, *$4); }
+    | PRAGMA name annotationBody END_PRAGMA
```
mbudiu-vmw: This is a non-standard language feature. Perhaps it should be marked as "experimental".
I thought pragmas were a p4-14 thing only.

mbudiu-vmw: So we can add a comment saying that this is not a standard P4-16 grammar construct.
We marked these as // experimental in the past.
We could consider standardizing `@pragma` in the spec too.

liujed: Done in 9667864. (Pragmas appear to be non-standard in P4₁₄ too!)

Issue #692, opened 2017-06-01
p4c-bm2-ss produces incorrect json when parser is selection on stack.next


antoninbas:
[stack-error.bmv2.p4.txt](https://github.com/p4lang/p4c/files/1053182/stack-error.bmv2.p4.txt) (edited)
I'm attaching a sample P4 program.
The compiler produces a JSON file which is rejected by simple_switch. Upon inspection of the JSON, the following is not correct:
```
"transition_key" : [
    {
        "type" : "field",
        "value" : ["hstack", "next.y"]
     }
]
```
I think it is okay for the bmv2 compiler to reject such a program for now, since there is no super-easy way of compiling this for bmv2. It is also not a very useful program because it selects on an uninitialized value.

mbudiu-vmw:
The program looks legal to me.
Why do you say last.y is uninitialized?
Would the following rewrite be compilable for bmv2? If so, then we have to do this in the back-end. We already to this kind of rewrite for other select expressions.

```
tmp = parseHdr.hstack.last;
transition select(tmp.y) {
   ...
}
```

Alternatively, the temporary could be parseHdr.hstack.last.y.

antoninbas:
My bad, I submitted the wrong P4 file. I have updated it. The parser is selecting on `parsedHdr.hstack.next.y`, which is uninitialized.

mbudiu-vmw:
The def-use analysis that we use is not smart enough to understand last/next, so it does not warn about uninitialized values. However, even if the value is uninitialized, this will probably be a warning and not an error. We should still generate a sensible output JSON. So this is a bug.

We may need to write a mid-end pass to convert next/last to explicit indexes; this was supposed to be part of the loop unrolling pass, which is currently broken and disabled. But this is not necessary for a target such as BMv2.

PR #1615, opened 2018-11-29
Fix for issue #692

mbudiu-vmw:


mbudiu-vmw:
No one wants to review this one-line fix?

backends/bmv2/common/expression.cpp
```diff
@@ -332,6 +332,8 @@ void ExpressionConverter::postorder(const IR::Member* expression)  {
     if (expression->expr->is<IR::Member>()) {
         auto mem = expression->expr->to<IR::Member>();
         auto memtype = typeMap->getType(mem->expr, true);
+        if (memtype->is<IR::Type_Stack>() && mem->member == IR::Type_Stack::next)
+            ::error("%1%: Reading next field, which is always uninitialized", mem);
```
jnfoster: Is there a test case for this behavior?

mbudiu-vmw: Yes, it is part of the PR.

mbudiu-vmw: It's the file called issue692-bmv2.p4.
Note that this is an error only in this back-end.
It's a warning in the front-end.

jnfoster: Got it.

PR #1617, opened 2018-11-30
Empty tuple

mbudiu-vmw:


mbudiu-vmw:
This is a one-line change in the grammar. I think we should take this change even before we approve it in the spec. Anyone care to review?

PR #1618, opened 2018-11-30
Support for 'int' type

mbudiu-vmw:


mbudiu-vmw:
@ChrisDodd : this solves a TODO you had left in a recent fix, about better error reporting when the width of an int constant cannot be inferred. This assumes that after type inference you cannot have Type_InfInt values (except in things like array maximum sizes), so we report any InfInt left in the program as an error. This can give better error messages.

PR #1619, opened 2018-11-30
Allow 1-bit int values

mbudiu-vmw:
I don't think that the v1errors folder was being used. It was in the wrong place anyway.


ChrisDodd:
We discussed adding bit<1> in the language design group (https://github.com/p4lang/p4-spec/issues/668), was a decision ever made?

mbudiu-vmw:
We will on Monday. Testing that it works.

mbudiu-vmw:
This is a small patch to allow `int<1>`.

mbudiu-vmw:
I think that the LDWG has accepted this change, so we should probably merge this PR.
This is a one-line change, can we get a review?

mbudiu-vmw:
I have pushed a change to the InvalidParserError test, it turns out that merging that PR and the mark_to_drop PR causes the build to fail. If this test passes it should fix the build when merged.

hesingh:
When I resolved my PSA11 PR conflict, indeed Travis failed most tests (3-4) due to mark_to_drop.

PR #1622, opened 2018-12-05
Fix for systems that default to python3.

liujed:


tools/cpplint.py
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python2
```
mbudiu-vmw: we have lots of other python files...

liujed: Sure, I'll go fix those too. I only noticed this one, since it was being used in a git hook in a different repo.

PR #1624, opened 2018-12-07
Fixes & improvements to annotation parsing

liujed:
* Improved the reporting of annotation-parsing errors:
  * We now report source-line information when the annotation parser runs off the end of the annotation.
  * Introduced a new `END_ANNOTATION` symbol for a better error message when the parser runs off the end of the annotation.
  * Added a negative test case for annotation parsing.

* Fixed & improved handling of unparsed annotations, and added the ability to warn about unknown annotations.
  * We now explicitly track whether an annotation needs parsing.

* Added a few more productions for parsing annotations with pairs or triples of expressions, string literals, and integer constants.
* Added ability to skip parsing of annotations. This is useful for avoiding warnings about annotations that are explicitly ignored.
* Renamed `PARSE_NO_BODY` to `PARSE_EMPTY`.
* Fixed a bug in the `annotationToken` production.

### DIFF
--- a/frontends/p4/parseAnnotations.cpp
+++ b/frontends/p4/parseAnnotations.cpp
@@ -46,7 +46,7 @@ bool ParseAnnotations::parseSkip(IR::Annotation*) {
 
 bool ParseAnnotations::parseEmpty(IR::Annotation* annotation) {
     if (!annotation->body.empty()) {
-        ::error("%1% should not have any argumentss", annotation);
+        ::error("%1% should not have any arguments", annotation);
         return false;
     }
 
--- a/frontends/p4/parseAnnotations.cpp
+++ b/frontends/p4/parseAnnotations.cpp
@@ -47,10 +47,6 @@ void ParseAnnotations::parseNoBody(IR::Annotation* annotation) {
 }
 
 void ParseAnnotations::parseExpressionList(IR::Annotation* annotation) {
-    if (!needsParsing(annotation)) {
-        return;
-    }
-
     const IR::Vector<IR::Expression>* parsed =
         P4::P4ParserDriver::parseExpressionList(annotation->srcInfo,
                                                 annotation->body);
@@ -60,10 +56,6 @@ void ParseAnnotations::parseExpressionList(IR::Annotation* annotation) {
 }
 
 void ParseAnnotations::parseKvList(IR::Annotation* annotation) {
-    if (!needsParsing(annotation)) {
-        return;
-    }
-
     const IR::IndexedVector<IR::NamedExpression>* parsed =
         P4::P4ParserDriver::parseKvList(annotation->srcInfo,
                                         annotation->body);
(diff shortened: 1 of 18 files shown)


ir/base.def
```diff
@@ -206,26 +206,41 @@ class AnnotationToken {
...
+
     /// Annotations that are simple expressions
     inline Vector<Expression> expr;
+
     /// Annotations described as key-value pairs
     inline IndexedVector<NamedExpression> kv;
+
     Annotation { if (!srcInfo) srcInfo = name.srcInfo; }
+
+    /// For annotations parsed from P4₁₆ source.
```
mbudiu-vmw: I worry a bit about non-ascii characters in source. The preferred spelling is P4-16 if you want to stick to ASCII.
Makes it also easier to grep.

ir/base.def
```diff
@@ -206,26 +206,41 @@ class AnnotationToken {
...
     Annotation(Util::SourceInfo si, ID n, const Vector<AnnotationToken> &a)
-    : Node(si), name(n), body(a) {}
+    : Node(si), name(n), body(a), needsParsing(true) {}
+
+    /// The remaining constructors are for compiler-generated annotations.
```
mbudiu-vmw: This should not be a triple-slash comment.

ir/base.def
liujed: I think we all should embrace the future and be less squeamish about using Unicode. But maybe my years of staring at lattices (viz. ⊔,⊓,⊑) has desensitized me! :smiley:

I buy the grep argument, though. Fixed.

ir/base.def
liujed: Fixed.

mbudiu-vmw:
I think there must be some dependency missing in the makefiles. Your new header files have broken the build of a dependent project: https://travis-ci.org/vmware/p4c-xdp/builds/465114164?utm_medium=notification&utm_source=email

The error is:
```
In file included from /home/p4c/frontends/parsers/parserDriver.h:25:0,
                 from /home/p4c/frontends/p4/parseAnnotations.h:21,
                 from /home/p4c/control-plane/p4RuntimeSerializer.cpp:45,
                 from /home/p4c/build/control-plane/unified_controlplane_sources_1.cpp:4:
/home/p4c/frontends/parsers/p4/abstractP4Lexer.hpp:4:10: fatal error: frontends/parsers/p4/p4parser.hpp: No such file or directory
 #include "frontends/parsers/p4/p4parser.hpp"
          ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

This must be because the Makefile does not realize that p4parser.hpp is produced by bison. Can you please look into this?

cc10512:
`set_source_files_properties(frontends/parsers/p4/p4parser.hpp PROPERTIES GENERATED TRUE)` to solve @mbudiu-vmw's breakage.

frontends/p4/parseAnnotations.cpp
```diff
@@ -87,7 +87,8 @@ void ParseAnnotations::postorder(IR::Annotation* annotation) {
 
     if (!handlers.count(annotation->name)) {
         // Unknown annotation. Leave as is, but warn if desired.
-        if (warnUnknown) {
+        if (warnUnknown && warned.count(annotation->name) != 0) {
```
cc10512: 👍

Issue #83, opened 2016-09-17
Trouble building on Mac OS X El Capitan


anirudhSK:
I am having some trouble building on Mac OS X.
1. It complains of yellow and yylval not being defined. For instance,

CXX      tools/ir-generator/ir-generator.o
In file included from ../tools/ir-generator/ir-generator.ypp:120:
<stdout>:857:1: error: use of undeclared identifier 'yylloc'
YY_RULE_SETUP
^
<stdout>:760:2: note: expanded from macro 'YY_RULE_SETUP'
        YY_USER_ACTION
        ^
../tools/ir-generator/ir-generator-lex.l:5:7: note: expanded from macro 'YY_USER_ACTION'
      yylloc = Util::SourceInfo(tmp, Util::InputSources::instance->getCurrentPosition()); }

<and more of the same>
1. I "fixed" this by adding two extern declarations at the top of tools/ir-generator/ir-generator-lex.l

extern YYSTYPE yylval;
extern YYLTYPE yylloc;

Not sure if this was correct, but it moved on.
1. Then, I ran into this. It seems like a value is being substituted for a template parameter, where it is expecting a type.

In file included from ../ir/ir.h:42:
../ir/nodemap.h:22:10: error: declaration of anonymous class must be a definition
template<class KEY, class VALUE,
         ^
../ir/nodemap.h:22:32: error: a non-type template parameter cannot have type 'class VALUE'
template<class KEY, class VALUE,
                               ^
../ir/nodemap.h:25:55: error: expected a type
         class ALLOC = std::allocator<std::pair<const KEY * const, const VALUE *>>>
                                                      ^
frontends/p4/p4-parse.cpp:161:13: note: expanded from macro 'KEY'
# define KEY 290

```
        ^
```

There are more errors, possibly resulting from the same cause, attached in the file here.
[err.txt](https://github.com/p4lang/p4c/files/478296/err.txt)
1. I seem to run into these where I use g++ or clang++ on Mac (specified using CXX="..." ./boostrap.sh).

Not sure if this is important, but I use ports, not brew for installing bison, flex, boost, and auto tools.

PR #1627, opened 2018-12-11
Fix Travis & missing build dependency

liujed:
* Added missing build dependency. The control plane now depends on the front end, so that the parser is built before the P4Runtime serializer. This fixes a heisenbug in the build process.

* Build now checks for `bison>=3.0.2`.

* Don't clobber the user's `.bash_profile` in `install_os_deps.sh`.

* Fix Travis OS X job. Modifying `.bash_profile` isn't good enough for Travis. Have to set
environment variables in `.travis.yml`.


liujed:
Travis job for OS X seems to be running into #83. @cc10512's theory is that this is caused by a recent OS X update.

Issue #1630, opened 2018-12-15
Verifying a checksum triggers a field update


edgar-costa:
This issue was originally opened in the `bmv2` repo:

p4lang/behavioral-model#691

mbudiu-vmw:
With the tests that @jafingerhut has committed I think we can close this issue.

PR #1632, opened 2018-12-17
Fix for issue #1630

mbudiu-vmw:


backends/bmv2/simple_switch/simpleSwitch.cpp
```diff
@@ -782,6 +782,10 @@ SimpleSwitchBackend::convertChecksum(const IR::BlockStatement *block, Util::Json
                     cksum->emplace("target", jleft->to<Util::JsonObject>()->get("value"));
                     cksum->emplace("type", "generic");
                     cksum->emplace("calculation", calcName);
+                    if (verify)
+                        cksum->emplace("verify", true);
+                    else
+                        cksum->emplace("update", true);
```
jafingerhut: Antonin said in a comment that BMv2 defaults to the value 'true' on these two keys, if they are absent.  I think it would be best in both cases to add both keys, both with explicit true/false values, otherwise I doubt it will fix the problem.

mbudiu-vmw:
This pass fixes a bug in the code generation for bmv2 for checksum verify/update; it is a two-line change in the bmv2 back-end.

ChrisDodd:
Would be nice if the one-line change summary at least mentioned bmv2 backend as well as the issue#

mbudiu-vmw:
Do you want the description in the commit message or is it enough if I add it to the PR description?

ChrisDodd:
It would be nice if it appeared in the history you see with `git show-branch` or various history browsers.  They generally show just the first line of the commit message (which you can set when you click "squash&merge" for the PR).

cc10512:
@mbudiu-vmw I found this blog post https://chris.beams.io/posts/git-commit/ incredibly useful.

mbudiu-vmw:
We should write something about this in the coding standards if we expect it to be used more widely. Perhaps post a link to the blog there, but we'd also need a shorter message.

The way you structure messages depends a lot on the tools you are using to browse them.

Issue #1083, opened 2017-12-07
ebpf support for p4 header stacks


zeeshanlakhani:
Discussed this with @mbudiu-vmw over email, but the ebpf target should support p4 header stacks.

PR #1633, opened 2018-12-17
Issue1083

mbudiu-vmw:


ChrisDodd:
Could I put in a request for more informative one-line change summaries as they appear in a history log?  That is the first line of the commit message should at least give an idea of what part of the compiler is affected and what is being done.  Including the issue# is great, but that should not be the only thing in the summary...

mbudiu-vmw:
This pass adds support for header stacks to the ebpf back-end. It only touches code generation in that back-end.

ChrisDodd:
So the one line summary in the change (which you can set when you squash&merge) should be something like

    ebpf support for header stacks (issue #1083)

PR #1637, opened 2018-12-18
fix for issue 1593: default type parameter in architecture definition

hanw:
Now TypeUnification won't complain about mismatching number of arguments if the control/parser block in architecture definition has a default value.

also fixed a related bug when inferring type for IR::Parameter with defaultValue in typeChecker.cpp


hanw:
Most of the parameters on control/parser blocks are `inout` type, when I tried to add defaultValue to any of them, p4c returns an error, such as:
```
./issue1593-bmv2.p4i(302): error: standard_metadata: out parameters cannot have default values
                      inout standard_metadata_t standard_metadata =
```

Is this expected? or should we allow default value on `inout` parameters.

mbudiu-vmw:
No, default values are constants, so they cannot be left values.

cc10512:
@mbudiu-vmw the lvalue is for the argument, not for the default value. Isn't the default value copied into the argument? And if the argument is written to, then it will be copied out? If it is not modified, the default value will be copied back out. Your concern is valid if we were to pass by reference.

mbudiu-vmw:
The default value is used for a parameter if no argument is supplied in a call. So if a parameter is `out` then value supplied as an argument must be a left-value. However, default values for parameters cannot be left values because they must be compile-time constants. C++ has a similar restriction: all default parameter values must be `static`; however, since P4 has no static left values, we only allow constants.

mbudiu-vmw:
I.e. consider
`void f(out bit<32> b = 0}`;
Where is f supposed to write the output value for b to if it is called without arguments?
In P4 you can say "don't care" for out values if that's what you want, but I don't think we allow don't care for a default parameter value. Maybe we should.

hanw:
I was asking about the default value for `inout` parameter, not `out`.

mbudiu-vmw:
Inout implies out. The argument must be a left-value.

hanw:
Let's think about this control block:

```
control C<H, M>(
    inout H hdr,
    inout M meta,
    inout intrinsic_metadata_t intr_md = {0, 0});
```

Suppose we instantiate the control block without the intr_md parameter

```
control MyC(inout hdr_t hdr, inout meta_t meta) {
   apply {}
}
```

The `inout` parameter is not used in the control block, the only source that can drive the `out` is the value that comes from the `in` direction. The default value is basically doing the following:

```
control (inout H hdr, inout M meta,
              inout intrinsic_metadata_t intr_md) {
   intrinsic_metadata_t tmp = {0, 0};
   intr_md = tmp;
}
```
except that the caller does not care which variable the `tmp` is written, as it is not used in the program.

If the `inout` parameter is used in the control block, then the default value is ignored.

In both case, the meaning of default value is pretty clear to me. I don't understand why we need to limit the default value of `inout` parameter to lvalue.

jnfoster:
Consider this code:
```
control MyD(inout H hdr,
    inout M meta,
    inout intrinsic_metadata_t intr_md )(C c) {
   c.apply(hdr,meta,intr_md);
} 
...
MyD(MyC()) myD;
```
What happens when we execution `myD.apply`? In particular, what value does its third argument have when the call completes?

mbudiu-vmw:
Parameters work in the same way for all sorts of objects, including extern methods and extern functions, controls and parsers, and user-written functions. Parameter directions are part of the function signature, and they describe how the function communicates with its environment. They allow us to typecheck all these objects independently from each other; if all objects use their signatures in a proper way then we can put them together and the result will typecheck, and we don't need to know how a function is implemented if we call it using arguments of the right type. That's why typechecking is the most important tool in programming languages, it is the best too we have for writing modular programs.

What you are saying is that: I want to implement a function that should have a signature with an inout parameter, but since my function does not happen to use its parameter, I want to change its signature to just "in". Well, you can write a new function, but the function will have a different signature.

This is no different from C++, if a function has a non-const parameter you cannot pass a const argument to it (const = non-left value). You cannot change the constness of a parameter when you override, or when you pass a function as an argument.

mbudiu-vmw:
Unfortunately when you write an architecture file you commit to some degree to a signature for its blocks - in particular you cannot change the directions later. You will have to supply implementations that match these directions always.

hanw:
Thanks, I understand now.

frontends/p4/typeChecking/typeUnification.cpp
```diff
@@ -382,8 +384,7 @@ bool TypeUnification::unify(const IR::Node* errorPosition,
 
         return true;
     } else if (dest->is<IR::Type_Declaration>() && src->is<IR::Type_Declaration>()) {
-        bool canUnify = typeid(dest) == typeid(src) &&
-                dest->to<IR::Type_Declaration>()->name == src->to<IR::Type_Declaration>()->name;
+        bool canUnify = typeid(dest) == typeid(src);
```
hanw: @mbudiu-vmw  Can you take a look at this line and see if removing the comparison on name makes sense to you?

The check fails for default-package-argument.p4, which it tries to compare M/8 with empty_t and the name of these two types does not match.

mbudiu-vmw:
This cannot be right. What is M?

hanw:
M is the type parameter in 
```
control C<H, M>(
    inout H hdr,
    inout M meta,
    in intrinsic_metadata_t intr_md);
```
in default-package-argument.p4

mbudiu-vmw:
M is a type variable, so it should unify, but not as a declaration. This point should not be reached.

hanw:
But it did. You can probably try the default-package-argument.p4 example on master.

mbudiu-vmw:
The fix is to take this out of `TypeInference::postorder(IR::Parameter*)`:
```
if (param->defaultValue != nullptr) {
        auto init = assignment(param, paramType, param->defaultValue);
        if (init == nullptr)
            return param;
        if (param->defaultValue != init)
            param->defaultValue = init;
    }
```

The downside is that a default value will never be typechecked if it is not used. But I think that in general we don't have enough information to typecheck the default value, as in the case you pointed out. Perhaps this can be fixed in a different way, but I think this is OK.

mbudiu-vmw:
I won't push this change, I will assume that you will, part of your PR.

hanw:
will do. thanks!

mbudiu-vmw:
Hopefully this won't break any test. I haven't rerun them all.
It should not break any positive test, it may break negative tests if there are any with this specific case where the default value is never used.

hanw:
The check was actually useful to catch the type error in this program.
```
struct intrinsic_metadata_t {
   bit<8> f0;
   bit<8> f1;
}

struct empty_t {}

control C<H, M>(
    inout H hdr,
    inout M meta,
    in intrinsic_metadata_t intr_md = {0, 0, 0});

package P<H, M>(C<H, M> c);

struct hdr_t { }
struct meta_t { }

control MyC(inout hdr_t hdr, inout meta_t meta) {
   apply {}
}

P(MyC()) main;
```
```
../testdata/p4_16_samples/default-control-argument.p4(11): error: intr_md: Number of fields 3 in initializer different than number of fields in structure 2: Tuple(3) to struct intrinsic_metadata_t
    in intrinsic_metadata_t intr_md = {0, 0, 0});
                            ^^^^^^^
../testdata/p4_16_samples/default-control-argument.p4(11)
    in intrinsic_metadata_t intr_md = {0, 0, 0});
                                      ^^^^^^^^^
../testdata/p4_16_samples/default-control-argument.p4(1)
struct intrinsic_metadata_t {
```

We probably need a different fix.

If I remove the code you mentioned, the compiler will accept the above program, even when the default value is used.

mbudiu-vmw:
The value is actually never used, so that's why it is never typechecked.
We agreed that default arguments for prototypes don't do really anything useful.

hanw:
The fix causes this test to fail with error.

```
#include <core.p4>

struct intrinsic_metadata_t {
   bit<8> f0;
   bit<8> f1;
}

struct empty_t {}

control C<H, M>(
    inout H hdr,
    inout M meta,
    in intrinsic_metadata_t intr_md);

package P<H, M>(C<H, M> c);

struct hdr_t { }
struct meta_t { bit<8> f0; bit<8> f1; bit<8> f2; }

control MyC2(in meta_t meta = {0, 0, 0}) {
  table a {
    key = {
      meta.f0 : exact;
    }
    actions = { NoAction; }
  }
  apply {
    a.apply();
  }
}

control MyC(inout hdr_t hdr, inout meta_t meta, in intrinsic_metadata_t intr_md) {
   MyC2() c2;
   apply {
     c2.apply();
   }
}

P(MyC()) main;
```

```
In file: p4c/frontends/p4/typeChecking/typeChecker.cpp:144
```

hanw:
Do I still need to remove the check in `TypeInference::postorder(IR::Parameter* param)`? I can wait a couple days till you fixed the pass orders.

mbudiu-vmw:
Please file a new issue with the example that failed originally.
If all other failing tests are part of the repo then that's enough. Otherwise please attach other tests that you tried to run and failed. I hope it won't be too difficult to fix, although I had surprises in the past.

hanw:
I filed https://github.com/p4lang/p4c/issues/1638 and https://github.com/p4lang/p4c/issues/1639

mbudiu-vmw:
I hope that #1640 will fix these issues.

mbudiu-vmw:
OK, let's remove issue1638-e.p4; it looks like the default value is not used anywhere.
I will let you do that, since my PR has been merged already.

BTW: in typeUnification.cpp I think you need to check for both source and destination parameters that they don't have default values.

hanw:
typeUnification.cpp checks the condition that there is no more source parameter && the destination parameter has a default value.

hanw:
are your referring to line 386 or line 166?

mbudiu-vmw:
In unifyFunctions there are two loops.

PR #1641, opened 2018-12-21
match on full file path when checking logging level

zma-barefoot:
I often find it useful to dump logs for all files under certain directory e.g. `-Tfoo/bar/*:3`. This PR proposes change that we match full file path, instead of individual file names. Individual file logging will continue to work as `-Tmyfile:5` with this change.

I also took the liberty of rewriting the `match` function using `std::regex`, though the current implementation seems to support certain patterns , such as `^`?, so I'm not sure if my change is backward compatible.

### DIFF
--- a/lib/log.cpp
+++ b/lib/log.cpp
@@ -32,6 +32,7 @@ limitations under the License.
 #include <string>
 #include <unordered_map>
 #include <vector>
+#include <regex>
 #ifdef MULTITHREAD
 #include <mutex>
 #endif  // MULTITHREAD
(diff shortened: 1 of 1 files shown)


lib/log.cpp
```diff
@@ -145,45 +146,15 @@ std::ostream& operator<<(std::ostream& out, const OutputLogPrefix& pfx) {
...
-        if (pattern == pend) return true;
-        while (*name && *name != *pattern) {
-            if (pbackup && *name == *pbackup) {
-                pattern = pbackup;
-                break; }
-            name++; }
-        pbackup = pattern;
+static bool match(const char *pattern, const char *file) {
+    if (strcmp(file + strlen(file) - 2, ".h") || strcmp(file + strlen(file) - 4, ".cpp")) {
+        std::regex expr(std::string(pattern, strchr(pattern, ':') - pattern));
```
ChrisDodd: This will crash if someone forgets the `:` on a `-T` option.  It also won't work for `-Tfile1,file2:1` -- better to use `strcspn(pattern, ",:")` as the old code did.

mbudiu-vmw:
@zma-barefoot : are you planning to address @ChrisDodd 's comments, or should we close the PR?

zma-barefoot:
Closed for now. It's not clear to me whether we need to be backward compatible with older version of GCC? `<regex>` is supported in GCC 4.9.0?

ChrisDodd:
The annoying thing about older versions of glibc++ is that they have a <regex> header and things using it will compile, but they won't actually work as the implementation in the library is stubbed out.

Issue #1504, opened 2018-09-20
bmv2 hardwired default table size


mbudiu-vmw:
The bmv2 back-end uses a size of 1024 when the size is not specified. This is wrong for tables with no entries or tables with a constant number of entries.

PR #1646, opened 2018-12-27
error reporting

cc10512:
This PR addresses the issue of repeated error and warning messages when
compiler passes are repeated, by introducing error/warning types, and
an error catalog. Error types classify the errors and impose a format
that allows the compiler to automatically filter repeated
messages. The filtering is based on the type of error and the source
code location of the object that reports the error. Thus, it allows
multiple error types per source code line, and ensures that only one
error is reported even if the message is raised multiple times. Error
codes and formats are defined in `lib/error_catalog.[h,cpp]`. Backends
can extend the codes and formats as needed (and they are encouraged to
do so).

To ease the transition to typed errors and warnings, free form
messages whose first argument (`%1%`) is an `IR::Node` (or more
precisely a class that implements `Util::IHasSourceInfo` interface),
will be converted automatically to typed errors that use the format
argument of the message rather than the error catalog predefined
formats.

We converted the errors in the bmv2 backend to typed errors, as well 
as a number of errors and warnings in the frontend. The conversion is
by no means finalized, however, the added support for transition does
handle a fair number of the duplicates. And I would like to get feedback
from others on what is their view before continuing to convert.

Also introduces the `FATAL_ERROR` macro to print a message and raise
an CompilationError exception.

And finally adds documentation on how to use to the coding standards.

**For reviewers**
I broke the PR in several commits to ease the review process:
- the first commit (https://github.com/p4lang/p4c/commit/fb2c8bcd01515f27ff0f11699fa63c20af0788b6) fixes a couple of warnings that point to potential issues with scoping
- the second commit (https://github.com/p4lang/p4c/commit/5f8b3c17b11e177f7e9fbbf5ed4c8578a943819d) adds the infrastructure support for typed errors. 
- and finally the https://github.com/p4lang/p4c/commit/7ac417bef01beca36e1e67304d2c7d6be64a2e29 converts bmv2 errors and https://github.com/p4lang/p4c/commit/9bcf80709b41cde3201821ecba9cc99f27d5e20e converts frontend errors and warnings.
It might help if you review these separately. 

**Still to do**
The sample outputs will need to be adjusted. Will do that after initial feedback.

### DIFF
--- a/backends/bmv2/common/action.cpp
+++ b/backends/bmv2/common/action.cpp
@@ -129,7 +129,7 @@ void ActionConverter::convertActionBody(const IR::Vector<IR::StatOrDecl>* body,
                 continue;
             }
         }
-        ::error("%1%: not yet supported on this target", s);
+        ::error(ErrorType::ERR_EXPRESSION, s, "not yet supported on this target");
     }
 }
 
@@ -144,7 +144,8 @@ ActionConverter::convertActionParams(const IR::ParameterList *parameters,
         param->emplace("name", p->name);
         auto type = ctxt->typeMap->getType(p, true);
         if (!type->is<IR::Type_Bits>())
-            ::error("%1%: Action parameters can only be bit<> or int<> on this target", p);
+            ::error(ErrorType::ERR_INVALID, p,
+                    "action parameters must be bit<> or int<> on this target");
         param->emplace("bitwidth", type->width_bits());
         params->append(param);
     }
--- a/backends/bmv2/common/action.cpp
+++ b/backends/bmv2/common/action.cpp
@@ -129,7 +129,7 @@ void ActionConverter::convertActionBody(const IR::Vector<IR::StatOrDecl>* body,
                 continue;
             }
         }
-        ::error(ErrorType::ERR_EXPRESSION, s, "not yet supported on this target");
+        ::error(ErrorType::ERR_UNSUPPORTED, "%1% not yet supported on this target", s);
     }
 }
 
@@ -138,14 +138,14 @@ ActionConverter::convertActionParams(const IR::ParameterList *parameters,
                                      Util::JsonArray* params) {
     for (auto p : *parameters->getEnumerator()) {
         if (!ctxt->refMap->isUsed(p))
-            ::warning("Unused action parameter %1%", p);
+            ::warning(ErrorType::WARN_UNUSED, "Unused action parameter %1%", p);
 
         auto param = new Util::JsonObject();
         param->emplace("name", p->name);
         auto type = ctxt->typeMap->getType(p, true);
         if (!type->is<IR::Type_Bits>())
-            ::error(ErrorType::ERR_INVALID, p,
-                    "action parameters must be bit<> or int<> on this target");
+            ::error(ErrorType::ERR_INVALID,
+                    "action parameters must be bit<> or int<> on this target", p);
         param->emplace("bitwidth", type->width_bits());
         params->append(param);
     }
--- a/backends/bmv2/common/control.cpp
+++ b/backends/bmv2/common/control.cpp
@@ -48,7 +48,7 @@ void ControlConverter::convertTableEntries(const IR::P4Table *table,
                     // booleans are converted to ints
                     key->emplace("key", stringRepr(k->to<IR::BoolLiteral>()->value ? 1 : 0, k8));
                 else
-                    ::error(ErrorType::ERR_UNSUPPORTED, k, "exact key expression");
+                    ::error(ErrorType::ERR_UNSUPPORTED, "exact key expression", k);
             } else if (matchType == corelib.ternaryMatch.name) {
                 if (k->is<IR::Mask>()) {
                     auto km = k->to<IR::Mask>();
@@ -61,7 +61,7 @@ void ControlConverter::convertTableEntries(const IR::P4Table *table,
                     key->emplace("key", stringRepr(0, k8));
                     key->emplace("mask", stringRepr(0, k8));
                 } else {
-                    ::error(ErrorType::ERR_UNSUPPORTED, k, "ternary key expression");
+                    ::error(ErrorType::ERR_UNSUPPORTED, "ternary key expression", k);
                 }
             } else if (matchType == corelib.lpmMatch.name) {
                 if (k->is<IR::Mask>()) {
@@ -72,7 +72,7 @@ void ControlConverter::convertTableEntries(const IR::P4Table *table,
                     unsigned long mask = km->right->to<IR::Constant>()->value.get_ui();
                     auto len = trailing_zeros(mask);
                     if (len + count_ones(mask) != keyWidth)  // any remaining 0s in the prefix?
-                        ::error(ErrorType::ERR_INVALID, k, "mask for LPM key");
+                        ::error(ErrorType::ERR_INVALID, "mask for LPM key", k);
                     else
                         key->emplace("prefix_length", keyWidth - len);
                 } else if (k->is<IR::Constant>()) {
@@ -82,7 +82,7 @@ void ControlConverter::convertTableEntries(const IR::P4Table *table,
                     key->emplace("key", stringRepr(0, k8));
                     key->emplace("prefix_length", 0);
                 } else {
-                    ::error(ErrorType::ERR_UNSUPPORTED, k, "LPM key expression");
+                    ::error(ErrorType::ERR_UNSUPPORTED, "LPM key expression", k);
                 }
             } else if (matchType == "range") {
                 if (k->is<IR::Range>()) {
@@ -93,10 +93,10 @@ void ControlConverter::convertTableEntries(const IR::P4Table *table,
                     key->emplace("start", stringRepr(0, k8));
                     key->emplace("end", stringRepr((1 << keyWidth)-1, k8));  // 2^N -1
                 } else {
-                    ::error(ErrorType::ERR_UNSUPPORTED, k, "range key expression");
+                    ::error(ErrorType::ERR_UNSUPPORTED, "range key expression", k);
                 }
             } else {
-                ::error(ErrorType::ERR_UNKNOWN, k, "key match type " + matchType);
+                ::error(ErrorType::ERR_UNKNOWN, "key match type '%2%' for key %1%", k, matchType);
             }
             matchKeys->append(key);
             keyIndex++;
@@ -105,7 +105,7 @@ void ControlConverter::convertTableEntries(const IR::P4Table *table,
         auto action = new Util::JsonObject();
         auto actionRef = e->getAction();
         if (!actionRef->is<IR::MethodCallExpression>())
-            ::error(ErrorType::ERR_INVALID, actionRef, "action in entries list");
+            ::error(ErrorType::ERR_INVALID, "Invalid action %1% in entries list.", actionRef);
         auto actionCall = actionRef->to<IR::MethodCallExpression>();
         auto method = actionCall->method->to<IR::PathExpression>()->path;
         auto decl = ctxt->refMap->getDeclaration(method, true);
@@ -121,11 +121,12 @@ void ControlConverter::convertTableEntries(const IR::P4Table *table,
         auto priorityAnnotation = e->getAnnotation("priority");
         if (priorityAnnotation != nullptr) {
             if (priorityAnnotation->expr.size() > 1)
-                ::error(ErrorType::ERR_INVALID, priorityAnnotation->expr, "priority value");
+                ::error(ErrorType::ERR_INVALID, "Invalid priority value %1%",
+                        priorityAnnotation->expr);
             auto priValue = priorityAnnotation->expr.front();
             if (!priValue->is<IR::Constant>())
-                ::error(ErrorType::ERR_INVALID, priorityAnnotation->expr,
-                        "priority value. Must be constant");
+                ::error(ErrorType::ERR_INVALID, "Invalid priority value %1%. Must be constant.",
+                        priorityAnnotation->expr);
             entry->emplace("priority", priValue->to<IR::Constant>()->value);
         } else {
             entry->emplace("priority", entryPriority);
@@ -155,7 +156,7 @@ cstring ControlConverter::getKeyMatchType(const IR::KeyElement *ke) {
         return mt->name.name;
     }
 
-    ::error(ErrorType::ERR_UNSUPPORTED, mt, "match type");
+    ::error(ErrorType::ERR_UNSUPPORTED, "match type not supported on this target", mt);
     return "invalid";
 }
 
@@ -171,7 +172,7 @@ ControlConverter::handleTableImplementation(const IR::Property* implementation,
     }
 
     if (!implementation->value->is<IR::ExpressionValue>()) {
-        ::error(ErrorType::ERR_EXPECTED, implementation, "expression for property");
+        ::error(ErrorType::ERR_EXPECTED, "expected expression for property", implementation);
         return false;
     }
     auto propv = implementation->value->to<IR::ExpressionValue>();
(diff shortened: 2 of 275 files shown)


mbudiu-vmw:
Can you point to an example which had redundant errors?
Some passes are run only once. Is it worth doing this transformation for such passes?

backends/bmv2/common/control.cpp
```diff
@@ -93,10 +93,10 @@ void ControlConverter::convertTableEntries(const IR::P4Table *table,
                     key->emplace("start", stringRepr(0, k8));
                     key->emplace("end", stringRepr((1 << keyWidth)-1, k8));  // 2^N -1
                 } else {
-                    ::error("%1% invalid range key expression", k);
+                    ::error(ErrorType::ERR_UNSUPPORTED, k, "range key expression");
                 }
             } else {
-                ::error("unkown key match type '%1%' for key %2%", matchType, k);
+                ::error(ErrorType::ERR_UNKNOWN, k, "key match type " + matchType);
```
mbudiu-vmw: Whenever you converted something to a concatenation you will get a different result; in particular you won't information about the source line position of the construct which is appended. Sometimes this may be desired.

backends/bmv2/common/action.cpp
```diff
@@ -129,7 +129,7 @@ void ActionConverter::convertActionBody(const IR::Vector<IR::StatOrDecl>* body,
                 continue;
             }
         }
-        ::error("%1%: not yet supported on this target", s);
+        ::error(ErrorType::ERR_EXPRESSION, s, "not yet supported on this target");
```
mbudiu-vmw: I find these errors harder to understand and write for the developer.
Wouldn't it be easier to add an enum to the error function describing the category?

docs/CodingStandardPhilosophy.md
```diff
@@ -77,3 +79,100 @@ repeating (visual) patterns should usually be avoided.  This is an aspect
...
+style comments. Comments should reflect the programmer's intention
+(rather than enumerating the code) and the reasoning behind certain
+decisions. Comments should also capture invariants that are not
+directly visible in the code. There are arguments that code should be
+self-documenting, and indeed we encourage the writing of clean
+self-explanatory code. Comments should contribute the additional
+meaning that helps extensibility and maintainability by a large group
+of developers.
+
+## Handling errors
```
mbudiu-vmw: This is too detailed for the coding standards. The longer this section, the less likely it is for anyone to read it.
Also, we should keep this very high level; these are really implementation details and not principles.

lib/CMakeLists.txt
```diff
@@ -17,6 +17,7 @@ set (LIBP4CTOOLKIT_SRCS
 	compile_context.cpp
 	crash.cpp
 	cstring.cpp
+        error_catalog.cpp
```
mbudiu-vmw: indentation?

cc10512:
Here is an example of the reduction in the number of warnings:
Using the bmv2 backend on p4c/testdata/p4_14_samples/switch_20160512/switch.p4
```
wc -l switch_trim.log switch.log
      38 switch_trim.log
      43 switch.log
```
The difference is:
```
diff switch_trim.log switch.log
23a24,25
> warning: 1024: Size specified for table without keys
> warning: 1024: Size specified for table without keys
38a41,43
> warning: Action selector '.ecmp_action_profile' is never referenced by a table and cannot be included in bmv2 JSON
> warning: Action selector '.fabric_lag_action_profile' is never referenced by a table and cannot be included in bmv2 JSON
> warning: Action selector '.lag_action_profile' is never referenced by a table and cannot be included in bmv2 JSON
```
The last three are filtered out because the errors are not typed, however, it is likely that you would like both warnings. Specifying a type allows you to issue different warnings for the same line.
The two logs are attached.
[switch_trim.log](https://github.com/p4lang/p4c/files/2713657/switch_trim.log)
[switch.log](https://github.com/p4lang/p4c/files/2713658/switch.log)

For an internal version of a switch profile, the reduction is quite more dramatic roughly 3x:
```
wc -l switch_MSDC_trim.log switch_MSDC.log
     145 switch_MSDC_trim.log
     494 switch_MSDC.log
```
Such a gap is the difference between people paying attention and fixing the warnings vs. continuously ignoring them.

backends/bmv2/common/control.cpp
cc10512: The line number is always printed for the first argument (k in this case).

backends/bmv2/common/action.cpp
cc10512: That's what ERR_EXPRESSION is (well, not really an enum because of C++ limitations on extending enum classes, and thus forcing me to use a class with static const members). In addition to the enum, this proposal structures the error so that we have a program object that we can use as source. If you prefer the original messages (typos included for some of them! :-) ) we can keep the messages and simply use the format from here rather than the error catalog.

docs/CodingStandardPhilosophy.md
cc10512: Where do we put the details?

mbudiu-vmw:
The first 2 errors are really a bug; they should display the table name as well, and probably they would not be repeated. This is tracked by https://github.com/p4lang/p4c/issues/1504.

So in this example I don't see any errors that are really repeated.
The fact that you lost the last 3 warning is also a bug.

Today most errors will prevent the next pass from running, so only warnings may be repeated.
Perhaps you could do the changes first just for warnings - this is likely the biggest payout.

docs/CodingStandardPhilosophy.md
mbudiu-vmw: We don't have a good place. The readme, where you removed some details from, isn't great either.
Some of this is also in the compiler-design powerpoint, but you can't find that using search.
People will have to read the code anyway to understand how to report errors, so perhaps this is implied.

lib/error_catalog.h
```diff
@@ -0,0 +1,53 @@
...
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+#ifndef P4C_LIB_ERROR_CATALOG_H_
+#define P4C_LIB_ERROR_CATALOG_H_
+
+/// enumerate supported errors
+/// Backends should extend this class with additional errors in the range 500-999 and
+/// warnings in the range 1500-2141.
```
hanw: wWhy is the range important? What happens if user decides to lower certain error to warning at compiler invocation time? Does the encoding move from range 500-999 to 1500-2141?

cc10512: No, the value of the enum won't move. The change in behavior is at the point of handling the warning/error, while the value just determines the format in the catalog.
The ranges are not essential, but I thought it would be good to partition the space for ease of further developing the notion of an error catalog.

lib/error_catalog.cpp
```diff
@@ -0,0 +1,37 @@
...
+limitations under the License.
+*/
+
+#include <map>
+#include <string>
+#include "error_catalog.h"
+#include "error_reporter.h"
+
+std::map<int, const std::string> ErrorReporter::errorCatalog = {
+    { ErrorType::ERR_UNKNOWN,     "%1%: Unknown %2%"},      // args: %1% - IR node, %2% - what
```
hanw: I was hoping each catalog can be a complete sentence about the error and then developer can provide more reasoning on the specific occurrence of the error. The error message we print should be like a git commit message, with a title and a body. Title is mostly fixed template by the compiler, body is provided by the developer.

For example:

```
Error: "file.p4", line xxx, column xxx: (error code)
  Unsupported primitive clone_egress_to_egress
  The supported primitives are clone_ingress_to_egress, clone_ingress_to_ingress.
```

mbudiu-vmw: Han's proposal would also make it easier to localize the errors if we decide to do it.

cc10512: I think this is reasonable. One potential issue is that we end up with too many errors codes, which will be confusing for developers to pick. So some amount of customization for the title will be needed as well. That said, I agree with giving it a try once we hear from a few more folks. 

I hope that it is clear what problem we're trying to address. Additional suggestions on how to make this work for both users and developers are more than welcome.

lib/error_catalog.cpp
```diff
@@ -0,0 +1,37 @@
...
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+#include <map>
+#include <string>
+#include "error_catalog.h"
+#include "error_reporter.h"
+
+std::map<int, const std::string> ErrorReporter::errorCatalog = {
```
antoninbas: Why is this declared in error_reporter.h and defined in error_catalog.cpp? Circular dependency?

I hate global variables like this as you don't know when they are constructed. I'd rather see a singleton pattern here (with a method that can be called by backends to register new error codes in the catalog?)

cc10512:
I've addressed the following reviewer comments:
- make the ErrorCatalog a singleton class and defined the format and title in error_catalog.cpp
- added methods to extract the error name and error format
- added support for naming the warnings so that classes of warnings can be disabled on the command line
- added types for most warnings (and a few errors)
- fixed the "bugs" that remove additional warnings on the same object. Note that there are several places where warnings do not get a source code object. Those were left un-typified to avoid eager filtering.
- updated the reference outputs.

I've tried Han's suggestion of having a "title" for the warnings and unfortunately there is no good way of covering entire classes with a single title -- it has to be customized, and thus, it will require the specialization that I implemented the first time. Since you found that hard to write messages for, I went for the simpler approach of simply printing the name of the "offending" object.

Please take a look and let me know what you think.

testdata/p4_16_samples_outputs/copy-frontend.p4
```diff
@@ -6,7 +6,7 @@ control c(inout bit<32> b) {
     S s1_0;
     S s2_0;
     @name("c.a") action a() {
-        s2_0 = { 32w0 };
+        s2_0 = {32w0};
```
mbudiu-vmw: there should be no whitespace difference here. is this because you are missing a rebase?

testdata/p4_16_samples_outputs/decl.p4-stderr
```diff
@@ -1,10 +1,10 @@
-decl.p4(33): warning: y shadows y
+decl.p4(33): [--Wwarn=shadow] warning: y: y shadows y
```
mbudiu-vmw: this error message has become worse.

testdata/p4_16_samples_outputs/default-switch.p4-stderr
```diff
@@ -1,4 +1,4 @@
-default-switch.p4(13): warning: SwitchCase: label following default label
+default-switch.p4(13): [--Wwarn=ordering] warning: SwitchCase: SwitchCase: label following default label
```
mbudiu-vmw: there are a whole bunch of errors from now on where there is a duplicated IR node name.

backends/bmv2/common/expression.cpp
```diff
@@ -433,7 +432,7 @@ void ExpressionConverter::postorder(const IR::Mux* expression)  {
     auto result = new Util::JsonObject();
     mapExpression(expression, result);
     if (simpleExpressionsOnly) {
-        ::error("%1%: expression to complex for this target", expression);
+        ::error(ErrorType::ERR_EXPRESSION, "too complex for this target", expression);
```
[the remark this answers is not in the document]
cc10512: Added UNSUPPORTED_ON_TARGET error type.

testdata/p4_16_errors_outputs/action-bind1.p4-stderr
```diff
@@ -1,4 +1,4 @@
-action-bind1.p4(22): error: a: Parameter b must be bound
+action-bind1.p4(22): [--Werror=legacy] error:  a: Parameter b must be bound
```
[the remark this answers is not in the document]
cc10512: Fixed.

testdata/p4_16_errors_outputs/issue306.p4-stderr
```diff
@@ -1,36 +1,6 @@
-issue306.p4(30): error: Structure struct my_packet does not have a field stack
+issue306.p4(30): [--Werror=legacy] error:  Structure struct my_packet does not have a field stack
```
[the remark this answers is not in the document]
cc10512: Legacy errors are errors that have not been typified. I added a TYPE_ERROR to address this issue.

testdata/p4_16_errors_outputs/stack3.p4-stderr
```diff
@@ -1,30 +1,6 @@
-stack3.p4(34): error: header h[]: Size of header stack type should be a constant
+stack3.p4(34): [--Werror=legacy] error:  header h[]: Size of header stack type should be a constant
```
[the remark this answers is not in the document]
cc10512: Added TYPE_ERROR.

testdata/p4_16_errors_outputs/twovarbit.p4-stderr
```diff
@@ -1,10 +1,4 @@
-twovarbit.p4(2): error: x and u: multiple varbit fields in a header
-    varbit<120> x;
-                ^
-twovarbit.p4(3)
-    varbit<120> u;
-                ^
-twovarbit.p4(2): error: x and u: multiple varbit fields in a header
+twovarbit.p4(2): [--Werror=legacy] error:  x and u: multiple varbit fields in a header
```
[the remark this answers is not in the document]
cc10512: It is printed in the latest.

testdata/p4_16_samples_outputs/copy-frontend.p4
cc10512: yes.

testdata/p4_16_samples_outputs/decl.p4-stderr
cc10512: Agreed. Removed the duplication.

testdata/p4_16_samples_outputs/default-switch.p4-stderr
cc10512: fixed.

cc10512:
@mbudiu-vmw I believe I addressed all your concerns. Can you please take another look?

lib/error_catalog.cpp
```diff
@@ -0,0 +1,98 @@
...
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+#include <map>
+#include <string>
+#include "error_catalog.h"
+
+// -------- Errors -------------
+const int ErrorType::LEGACY_ERROR      =   0;
```
mbudiu-vmw: Why can't these be just an enum?
Do we care about their concrete values outside of the compiler?

testdata/p4_16_errors_outputs/issue306.p4-stderr
```diff
@@ -1,36 +1,6 @@
-issue306.p4(30): error: Structure struct my_packet does not have a field stack
+issue306.p4(30): [--Werror=type-error] error: Structure struct my_packet does not have a field stack
```
mbudiu-vmw: It looks like for this error the two arguments should be swapped, because these are actually not duplicate messages.

cc10512: Attached is how the errors look if I switch it. It is true that it identifies all uses of the .stack member, but it does seem a bit redundant? Let me know if you prefer it this way.
[issue306.p4-stderr.txt](https://github.com/p4lang/p4c/files/2829218/issue306.p4-stderr.txt)

lib/error_catalog.cpp
cc10512: Per the class comment:
```
/// enumerate supported errors
/// Backends should extend this class with additional errors in the range 500-999 and
/// warnings in the range 1500-2141.
/// It is a class and not an enum class because in C++11 you can't extend an enum class
class ErrorType {
```

testdata/p4_16_errors_outputs/issue306.p4-stderr
mbudiu-vmw: I think it's better. We could avoid printing the struct position altogether, although it's probably useful the first time.

PR #1648, opened 2018-12-28
Pass field list and calc names through annotations

amreshk:
The `field_list` and `field_list_calculation` names specified in a `p4-14` program are lost during program conversion. We use annotations to get past this issue and save these names to be passed on.


frontends/p4/fromv1.0/programStructure.cpp
```diff
@@ -1556,7 +1556,14 @@ CONVERT_PRIMITIVE(modify_field_with_hash_based_offset) {
         new IR::Cast(max->srcInfo, IR::Type_Bits::get(2 * flc->output_width), max)));
     auto hash = new IR::PathExpression(structure->v1model.hash.Id());
     auto mc = new IR::MethodCallExpression(primitive->srcInfo, hash, args);
-    auto result = new IR::MethodCallStatement(primitive->srcInfo, mc);
+    auto annos = new IR::Annotations();
+    // Save the original field list calculation name in an annotation
+    annos->addAnnotation("field_list_calc_name", new IR::StringLiteral(flc->name));
+    // Save the original field list names in annotations
+    for (auto fl_name : flc->input->names) {
+        annos->addAnnotation("field_list_name", new IR::StringLiteral(fl_name));
```
cc10512: Doesn't the order matter? How is that preserved if the field list name annotation is not associated with the field?

frontends/p4/fromv1.0/programStructure.cpp
```diff
@@ -1752,8 +1760,16 @@ ProgramStructure::convertAction(const IR::ActionFunction* action, cstring newNam
 
     for (auto p : action->action) {
         auto stat = convertPrimitive(p);
-        if (stat != nullptr)
-            body->push_back(stat); }
+        if (stat != nullptr) {
+            body->push_back(stat); 
+            if (auto mcs = stat->to<IR::MethodCallStatement>()) {
```
cc10512: Why not construct `annots` using the copy constructor that takes annotations. I.e., `annots = new IR::Annotations(mcs->getSourceInfo(), mcs->getAnnotations()->annotations)` and directly add to the body?
Essentially, this whole code can be:
```
if (stats != nullptr) {
    body->push_back(stat);
    if (auto mcs = stat->to<IR::MethodCallStatement>())
        body->annotations = new IR::Annotations(mcs->getSourceInfo(), mcs->getAnnotations()->annotations);
}
```

mbudiu-vmw:
I don't think this will work. The problem with annotations on statements is that they have no semantics, and because of this reason the front-end does not know what to do with them. You can see this in the treatment of BlockStatements with annotations: the compiler has to leave them alone, since it does not know what they annotations may mean.

For assigments, your change is probably useless, because lots of passes in the front-end rewrite assignments in various ways, and thus the annotations will soon be lost. For MethodCallStatements you may argue that the compiler does not really understand them, so perhaps they may preserve the annotations. However, you will have to audit *all* the passes and make sure that they do preserve these annotations. For that purpose you should make the annotations non-optional fields of the IR.

However, I think that in general this is not an adequate solution. The compiler cannot operate on an IR if it does not understand its meaning, and thus the compiler cannot operate on statements with annotations. 

Why do you need to preserve field list names? The best way to do this would be to create types and instances of types that carry these names, not to annotation the statements that operate on these objects.

PR #1652, opened 2019-01-08
Fix def_use/SimplifyDefUse to deal with extern methods properly (#1284)

ChrisDodd:
- Track extern method calls, assuming that any call to any
  non-abstract method might trigger any abstract method implementation
- local_copyprop fixed for the same assumption.
- propagate reads/writes from callee to caller properly.
- @synchronized annotation for abstract methods

This is revised slightly from the previous change -- the additional change is to make RemoveUnused in simplifyDefUse ignore IR::Function nodes that have not been analyzed by FindUninitialized as there were no (direct or indirect) references to the function.  The problem being that visiting such a function would remove all the assignments in the function as none of them were in `hasUses`.  That would be ok if the function was in fact unused, but in the case of a non-@synchronous function, that would be a problem.  We don't know how such functions are called (from the control plane?) so we can't really analyze them, so we just want to leave them unchanged.

The problem is if such a function accesses a control local variable, the analysis won't see it, so might remove an (apparently unused) control local.  Its not clear if such functions should be allowed to access control locals; we don't really have any example use cases for non-@synchronous functions that would want to.

### DIFF
--- a/frontends/p4/simplifyDefUse.cpp
+++ b/frontends/p4/simplifyDefUse.cpp
@@ -58,17 +58,22 @@ class HasUses {
 };
 
 // Run for each parser and control separately
+// Somewhat of a misnamed pass -- the main point of this pass is to find all the uses
+// of each definition, and fill in the `hasUses` output with all the definitions that have
+// uses so RemoveUnused can remove unused things.  It incidentally finds uses that have
+// no definitions and issues uninitialized warnings about them.
 class FindUninitialized : public Inspector {
-    ProgramPoint    context;
+    ProgramPoint    context;    // context as of the last call or state transition
     ReferenceMap*   refMap;
     TypeMap*        typeMap;
     AllDefinitions* definitions;
     bool            lhs;  // checking the lhs of an assignment
-    ProgramPoint    currentPoint;
+    ProgramPoint    currentPoint;  // context of the current expression/statement
     // For some simple expresssions keep here the read location sets.
     // This does not include location sets read by subexpressions.
     std::map<const IR::Expression*, const LocationSet*> readLocations;
     HasUses*        hasUses;  // output
+    const IR::IndexedVector<IR::Declaration>    *controlLocals = nullptr;
 
     const LocationSet* getReads(const IR::Expression* expression, bool nonNull = false) const {
         auto result = ::get(readLocations, expression);
@@ -78,7 +83,7 @@ class FindUninitialized : public Inspector {
     }
     // 'expression' is reading the 'loc' location set
     void reads(const IR::Expression* expression, const LocationSet* loc) {
-        LOG3(dbp(expression) << " reads " << loc);
+        LOG3(expression << " reads " << loc);
         CHECK_NULL(expression);
         CHECK_NULL(loc);
         readLocations.emplace(expression, loc);
@@ -92,8 +97,8 @@ class FindUninitialized : public Inspector {
             context(context), refMap(parent->definitions->storageMap->refMap),
             typeMap(parent->definitions->storageMap->typeMap),
             definitions(parent->definitions), lhs(false), currentPoint(context),
-            hasUses(parent->hasUses)
-    { setName("FindUninitialized"); }
+            hasUses(parent->hasUses), controlLocals(parent->controlLocals)
+    { visitDagOnce = false; }
 
  public:
     FindUninitialized(AllDefinitions* definitions, HasUses* hasUses) :
@@ -103,13 +108,13 @@ class FindUninitialized : public Inspector {
             hasUses(hasUses) {
         CHECK_NULL(refMap); CHECK_NULL(typeMap); CHECK_NULL(definitions);
         CHECK_NULL(hasUses);
-        setName("FindUninitialized"); }
+        visitDagOnce = false; }
 
     // we control the traversal order manually, so we always 'prune()'
     // (return false from preorder)
 
     bool preorder(const IR::ParserState* state) override {
-        LOG3("FU Visiting " << dbp(state));
+        LOG3("FU Visiting state " << state->name);
         context = ProgramPoint(state);
         currentPoint = ProgramPoint(state);  // point before the first statement
         visit(state->components, "components");
@@ -159,7 +164,9 @@ class FindUninitialized : public Inspector {
     }
 
     bool preorder(const IR::P4Control* control) override {
-        LOG3("FU Visiting " << dbp(control));
+        LOG3("FU Visiting control " << control->name << "[" << control->id << "]");
+        BUG_CHECK(context.isBeforeStart(), "non-empty context in FindUnitialized::P4Control");
+        controlLocals = &control->controlLocals;
         currentPoint = ProgramPoint(control);
         for (auto d : control->controlLocals)
             if (d->is<IR::Declaration_Instance>())
@@ -168,11 +175,15 @@ class FindUninitialized : public Inspector {
         visit(control->body);
         checkOutParameters(
             control, control->getApplyMethodType()->parameters, getCurrentDefinitions());
+        controlLocals = nullptr;
         return false;
     }
 
     bool preorder(const IR::Function* func) override {
-        LOG3("FU Visiting " << dbp(func));
+        LOG3("FU Visiting function " << func->name << "[" << func->id << "]");
+        LOG5(func);
+        // FIXME -- this throws away the context of the current point, which seems wrong,
+        // FIXME -- but otherwise analysis fails
         currentPoint = ProgramPoint(func);
         visit(func->body);
         bool checkReturn = !func->type->returnType->is<IR::Type_Void>();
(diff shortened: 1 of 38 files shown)


mbudiu-vmw:
But Functions are global-scoped, so they cannot refer to any control-locals...

frontends/p4/simplifyDefUse.cpp
```diff
@@ -545,6 +551,13 @@ class RemoveUnused : public Transform {
         }
         return statement;
     }
+    const IR::Node* preorder(IR::Function* func) override {
```
mbudiu-vmw: do you have an example where this is important?

midend/local_copyprop.cpp
```diff
@@ -409,12 +435,21 @@ IR::MethodCallExpression *DoLocalCopyPropagation::postorder(IR::MethodCallExpres
         if (actions.count(fn->path->name)) {
             LOG3("action method call " << mc->method);
             apply_function(&actions[fn->path->name]);
+            return mc;
+        } else if (mi->is<P4::ExternFunction>()) {
+            LOG3("extern function call " << mc->method);
+            // assume it has no side-effects on anything not explicitly passed to it?
+            // maybe should have annotations if it does
```
mbudiu-vmw: even if it has side-effects it should be unable to change any named variables (locals). We do not have references, so you cannot change something you cannot name - due to copy-in copy-out.

mbudiu-vmw: it cannot call any methods, since it cannot refer to instances of the allocated objects that have these methods.

testdata/p4_16_samples/virtual3.p4
```diff
@@ -0,0 +1,55 @@
...
+*/
+#include <core.p4>
+
+// In arch file
+extern Virtual {
+    Virtual();
+    // abstract methods must be implemented
+    // by the users
+    void increment();
+    @synchronous(increment) abstract bit<16> f(in bit<16> ix);
```
mbudiu-vmw: So the meaning of this is that increment calls f?
Shouldn't the annotation be on increment instead?

frontends/p4/simplifyDefUse.cpp
ChrisDodd: I extended the virtual.p4 test case to include the problematic case -- an abstract method with an inout arg where the implementation assigns to the arg (could also happen with an out arg) -- not skipping the method here would remove the assignment (leaving the method empty).

testdata/p4_16_samples/virtual3.p4
ChrisDodd: Yes, the annotation means that increment might call f (but is not guarenteed to).  At the moment I have the annotations on the abstract methods, listing the other methods that might call them, but it could equivalently be done the opposite way (put @might_call annotations on methods listing the other methods it might call?)

midend/local_copyprop.cpp
ChrisDodd: We don't know for certain that the architecture does not provide a back-door method of referring to other objects -- for example, the packet_in and packet_out objects that are passed to parser and deparsers might be affected by other methods in some way?  Perhaps a clone object that can have a method invoked that triggers an additional copy of the packet?

If we want to allow such architectures, we really need to provide langauge that the architecture can be described in (in the arch.p4 file); I'm not sure annotations are good enough longer-term.  Something for the language design group.

testdata/p4_16_samples/virtual3.p4
mbudiu-vmw: If we want to support the case where the same method may be called both synchronously and asynchronously it may make sense to put the annotation on the caller.

I propose these rules:
- a method (virtual or non-virtual) that may be called asynchronously *must* be annotated with `@async`. This means that the method may be called at any time, from any thread. Such methods cannot access any P4 locals, only their own parameters. Such methods can also be called synchronously.
- a non-virtual method that may call `callee` is labeled with `@calls(callee)`

With these rules we have that asynchronous methods are always live, but can never reach into locals. We should always run def-use for them. Synchronous methods may use locals, but we should know all their callers, so if they are unreachable they are actually dead. Also, it is an error for an asynchronous method to call a synchronous one.

frontends/p4/simplifyDefUse.cpp
mbudiu-vmw: With the rules I proposed this would be correct, because no one is calling this method.
We need to know who is calling the method to analyze it.
If we don't know the method is either asynchronous or dead.

testdata/p4_16_samples/virtual3.p4
ChrisDodd: Problem is that we can't run def_use on these methods as its currently structured, as it tends to crash with a BUG_CHECK in Definitions::getPoints because it can't figure out where the function is called from.  We need some way of constructing a ProgramPoint corresponding to the (non-existant) call site before we can analyze the body.

As far a what the annotations should be and which functions they should be on, the two are equivalent -- you're now back to something more like the original suggestion before it was later suggested they should be on the abstract methods.

mbudiu-vmw: We should be able to create a ProgramPoints for asynchronous functions (which have no explicit caller). The same problem appears in several other places. There's a "beforeStart" ProgramPoint which may work.

ChrisDodd:
Ok, so I've gone back to (more like) the old code -- def_use now (continues to) analyze all functions with an assumption that they're called from outside the control, and additionally analyzes directly called and @synchronous functions per the direct call site.

Unfortunately it looks like Travis has begun failing the OSX CI build with `*** cannot import name RawPcapWriter` for all ebpf tests?

mbudiu-vmw:
How does one even debug a failing Travis build for a different OS?
RawPcapWriter should be part of scapy.

mbudiu-vmw:
@fruffy : if you have some time can you try to figure out what's wrong with the Python testing script?

mbudiu-vmw:
@ChrisDodd : I am not sure whether this will make a difference, but maybe you can try to modify the file backends/ebpf/targets/target.py as follows:

```
from scapy.utils import *
from scapy.layers.all import *
```
 The stars replace more specific names.

fruffy:
Looking at the output log of travis it seems that it fails to install scapy.

https://scapy.readthedocs.io/en/latest/installation.html#mac-os-x
Maybe these steps are missing?

mbudiu-vmw:
Actually the p4c-xdp build failed with the same error, so this does not seem to be OSX only:
https://travis-ci.org/vmware/p4c-xdp

fruffy:
I just tried to install scapy on my local machine
`pip install --upgrade --user scapy`
and it failed with the same error.
`pip install --upgrade --user scapy==2.4.0` 
worked fine. So this might be a quick and dirty workaround.

mbudiu-vmw:
This sounds like a good idea. @ChrisDodd : can you please try this fix?
We should indeed stick to specific versions of packages that we know are working.
I assume it's enough to replace in the Dockerfile `scapy` with `scapy==2.4.0`

ChrisDodd:
Looks like scapy==2.4.0 does not work on Travis's OSX -- see #1656.  I'm going to try just removing all the ebpf tests from the OSX build next.

mbudiu-vmw:
Actually by looking at the travis.yml file it looks to me like the ebpf tests should not be run on OSX.
```
script:
  - if [[ $TRAVIS_OS_NAME == 'linux' ]] ; then docker run -w /p4c/build -e CTEST_PARALLEL_LEVEL p4c ctest --output-on-failure --schedule-random ; fi
  - if [[ $TRAVIS_OS_NAME == 'osx' ]] ; then ctest --output-on-failure -j 2 --schedule-random -LE "ebpf$" ; fi
```

But I don't know much about how travis works.

mbudiu-vmw:
At least that was the intention, I guess.

ChrisDodd:
The tests were working (on OSX) a week or so ago -- something changed (in the Travis setup?) that causes them all to fail now.

mbudiu-vmw:
I actually think it is a change in the scapy python module that breaks things.
No one touched Travis or Docker lately.

mbudiu-vmw:
That's why I was hoping that moving to an old version would fix the problem.

mbudiu-vmw:
Making this change in the Dockerfile of the p4c-xdp project has fixed that build:
```
scapy==2.4.0
```

mbudiu-vmw:
The build there was failing with the same error.

fruffy:
Looking at #1656 it still seems to download scapy-2.4.2:
```
Collecting scapy
  Downloading https://files.pythonhosted.org/packages/d0/04
/b8512e5126a1816581767bf95cbf525e0681a22c055bcd45f47ecff60170/scapy-2.4.2.tar.gz (795kB)
```

The file `tools/install_os_deps.sh` installs scapy before the dockerfile runs I think.

mbudiu-vmw:
Good catch! Thanks @fruffy!

Issue #1049, opened 2017-11-19
Another case with missing d2b/b2d in bmv2 backend


jafingerhut:
[issue1049.p4.txt](https://github.com/p4lang/p4c/files/1485876/issue1049.p4.txt)

That attached program issue1049.p4.txt, when compiled with p4c-bm2-ss dated 2017-Nov-17 exhibits a problem similar to those seen in several other issues, including #737, where the bmv2 JSON file is missing d2b/b2d calls needed to convert types between bitvector and boolean in bmv2.

In particular, the condition node for evaluating the boolean condition meta.mystruct1.hash_drop should be wrapped in a d2b call to convert it from 1-bit bitvectors to bool, I believe.

jafingerhut:
[issue1049.zip](https://github.com/p4lang/p4c/files/1485977/issue1049.zip)

The attached zip file contains a somewhat modified P4 program exhibiting the problem.  It is modified to make it easier to write an STF test case for it.  I haven't run the STF test case with a corrected compiler, because I don't have a proposed patch for p4c that fixes this issue, but I believe it should be correct, as I have compared the input and output packets against what is produced by simple_switch with a hand-edited JSON file that is hopefully what p4c will produce when the bug is fixed, with d2b calls wrapped around uses of the mystruct1.hash_drop 1-bit bitvector (bool in the P4_16 source, but 1-bit bitvector according to bmv2).

Note that the failure with the incorrect JSON is very weird -- it isn't using an uninitialized value like some other test cases found for similar issues - it is causing the evaluation of a _different_ expression to give the wrong value, because of the multi-operand-stack method that bmv2 uses to evaluate expressions.  Makes perfect sense why if you know the bmv2 internals well enough.

jafingerhut:
Re-tested the provided test case and maybe one other, and things look better now.  Thanks!

PR #1654, opened 2019-01-10
Follow-on work for nested struct - add SerEnum and boolean to nested struct

hesingh:
Now that the nested struct/header support PR is checked into the repo, it's follow-on work is done in this PR.  Specifically, SerEnum and boolean are added to the bit-vector check.  Also, a change is made to allow a boolean inside a P4 header.


hanw:
Not a comment, just a thought. We now have struct that are serializable and non-serializable. I wonder if we should make the distinction clear by introducing separating IR classes for these two type of structs in the frontend. Currently we rely on a helper function to `t->is<IR::Type_Struct>() && onlyBitsOrBitStructs(t)` to tell the difference, which is perhaps less desirable in the long run.

hanw:
The syntax for structs are overloaded, but the semantics should be made clear with different IR classes.

mbudiu-vmw:
The notion of serialization is relative. Some things can be serialized as headers, some other things can be serialized in messages for the control-plane. Personally I don't think that we should change the IR every time we make updates to the language spec. At some point we will start freezing the IR classes, because other people may take dependencies on the program representation which we can no longer change.

hesingh:
Even before the latest revision of the P4-16 spec from November 2018, the spec has supported struct inside struct with no additional IR type or serialization-related text.   I agree with Mihai.

Also, note, I have started a PR for spec change due to nested struct p4c code changes.  Please see what I have updated in the spec in the PR below.

https://github.com/p4lang/p4-spec/pull/719

hesingh:
Right now, the more important coding issue is to fix the bmv2 backend for a more complex test case I just checked into this PR's workspace.   See

testdata/p4_16_samples/issue1653-complex-bmv2.p4

The failed log is below.  The tmp_2 variable should not have a op = d2b added by the bmv2 backend.  

```
hemant@ubuntu:~/p4c-1/build$ bmv2/testdata/p4_16_samples/issue1653-complex-bmv2.p4.test -b -v
Writing temporary files into  ./tmp64MP5b
Executing  ./p4c-bm2-ss -o bmv2/testdata/p4_16_samples/issue1653-complex-bmv2.json /home/hemant/p4c-1/testdata/p4_16_samples/issue1653-complex-bmv2.p4
Exit code  0
Check for  /home/hemant/p4c-1/testdata/p4_16_samples/issue1653-complex-bmv2.stf
Running model
Running /home/hemant/behavioral-model/targets/simple_switch/simple_switch --log-file switch.log --log-flush --use-files 0 --thrift-port 9459 --device-id 369 ../bmv2/testdata/p4_16_samples/issue1653-complex-bmv2.json
Invalid entry type 'expression' in field list
bad json:
{
   "type" : "expression",
   "value" : {
      "type" : "expression",
      "value" : {
         "left" : null,
         "op" : "d2b",
         "right" : {
            "type" : "field",
            "value" : [ "scalars", "tmp_2" ]
         }
      }
   }
}

Running /home/hemant/behavioral-model/targets/simple_switch/simple_switch_CLI --thrift-port 9459
STF Command: 
Could not connect to thrift client on port 9459
Exception  [Errno 3] No such process
hemant@ubuntu:~/p4c-1/build$ 
```

jafingerhut:
@hesingh Just a caution about boolean values in the BMv2 back end that has occurred before (and there may still be some remaining bugs, but none have been seen in a while): BMv2 simple_switch does not "catch" the compiler if it makes the mistake of trying to use a boolean value as a bit<1>, or vice versa.  If you forget a needed d2b or b2d conversion operation, simple_switch can silently misbehave.

Here are some earlier issues that were corrected related to this, in case you are curious:
https://github.com/p4lang/p4c/issues/737
https://github.com/p4lang/p4c/issues/1049

Note: I am _not_ saying that you are necessarily prone to introducing new such errors in the changes you are making, but I did want you to realize that simple_switch will not currently help you catch such a mistake in the generated BMv2 JSON expressions.

hesingh:
@jafingerhut:  Thanks for the references - I didn't know about them.  However, I did find the relevant code in the bmv2 backend that changed expression.cpp.  Today morning, I actually changed code in expression.cpp and fixed the test case issue1653-bmv2.p4.  I am now dealing with the next complex test case that causes the bmv2 backend to add a d2b op when the op should not be added.  This error is caused by the clone3 statement in the test.  I will continue to investigate the bmv2 backend code and see what to do.

hesingh:
@mbudiu-vmw : The tmp, tmp_1, tmp_2, ..., tmp_8 variables are generated due to the clone3 statement in the P4 program (issue1653-complex-bmv2.p4.test).  tmp_2 and tmp_8 deal with boolean in postorder(IR::PathExpression).  Both get incorrectly marked with an op = d2b because leftValue = 0, and simpleExpressionsOnly = 0.   We can't fix this bug unless we know we are in a clone3 operation and thus we should not go to the d2b else-if.   Or we need another condition which you may be more familiar with.  No matter what, we need a third condition to go into a non-d2b code section for boolean.

With a hack of hard-coding tmp_2 and tmp_8, the bug is fixed.  

```
} else if (type->is<IR::Type_Bits>() ||
          (type->is<IR::Type_Boolean>() &&
	  ((leftValue || simpleExpressionsOnly) ||
	  ((var->name == "tmp_2") || (var->name == "tmp_8"))))) {
```

mbudiu-vmw:
I have filed issue https://github.com/p4lang/p4c/issues/1660 to keep track of this.

hesingh:
> I have filed issue #1660 to keep track of this.

Thanks, Mihai!

frontends/p4/typeChecking/typeChecker.cpp
```diff
@@ -2720,7 +2720,8 @@ bool TypeInference::hasVarbitsOrUnions(const IR::Type* type) const {
 
 bool TypeInference::onlyBitsOrBitStructs(const IR::Type* type) const {
     // called for a canonical type
-    if (type->is<IR::Type_Bits>()) {
+    if (type->is<IR::Type_Bits>() || type->is<IR::Type_InfInt>() ||
```
pofig37: IR::Type_InfInt shouldn't be allowed here, we can only have fixed width integers (represented by IR::Type_Bits).

hesingh: Good catch.  Sorry, I missed that Type_InfInt does not represent ```int<W>```.

hesingh:
@mbudiu-vmw: First, thanks for fixing 1660 quickly!  
This PR is ready for review.  Just typeCheck.cpp is changed for source code.  Rest of the changes are test cases.  thanks.

testdata/p4_16_samples/header-bool-bmv2.p4
```diff
@@ -1,11 +1,12 @@
 /*
-Copyright 2017 VMware, Inc.
+Copyright 2019 MNK Consulting, LLC.
```
mbudiu-vmw: if you delete issue272.p4 you have to also delete the corresponding outputs

testdata/p4_16_samples/header-bool-bmv2.p4
```diff
@@ -14,8 +15,22 @@ See the License for the specific language governing permissions and
...
-header boolfield {
-    bool x;
+header hdr {
+    bit<31> f;
+    bool    x;
 }
+
+control compute(inout hdr h) {
+    apply {
+        hdr tmp;
```
mbudiu-vmw: your indentation is inconsistent.

testdata/p4_16_samples/issue383-bmv2.p4
```diff
@@ -50,27 +50,27 @@ struct parsed_packet_t {
...
 parser parse(packet_in pk, out parsed_packet_t h,
-             inout local_metadata_t local_metadata,
-             inout standard_metadata_t standard_metadata) {
-  state start {
-    pk.extract(h.bvh0);
-    pk.extract(h.bvh1);
-    pk.extract(local_metadata.col.bvh);
-    transition accept;
-  }
+inout local_metadata_t local_metadata,
```
mbudiu-vmw: this indentation is worse than it used to be for these two lines.

testdata/p4_16_samples/header-bool-bmv2.p4
hesingh: wow, sorry, I missed fixing the indentation for the new test.  I will fix the other one too.

hesingh: Is it?  The 'hdr tmp;' is indented by 4 spaces from the 'apply'.   Each data struct member is also indented by 4 spaces.

testdata/p4_16_samples/issue383-bmv2.p4
hesingh: This file's indentation also uses 4 spaces from the top statement 'state'.

hesingh:
Before developing the STF test, I tested header-bool-bmv2.p4 and found a bmv2 issue that needs fixing.  I hit the failure on line 297 in expression.cpp in the bmv2 backend.   To fix this issue, I had to emplace header fields in the bmv2 backend which only emplace struct fields so far.  Code is checked in.

Thereafter, I ran into yet another behavioral-model error:

```Invalid reference to object of type 'header' with name 'h' ```

Detailed log is included below.  
```
hemant@ubuntu:~/p4c-1/build$ bmv2/testdata/p4_16_samples/header-bool-bmv2.p4.test -v
Writing temporary files into  ./tmpHhCDWi
Executing  ./p4c-bm2-ss -o bmv2/testdata/p4_16_samples/header-bool-bmv2.json /home/hemant/p4c-1/testdata/p4_16_samples/header-bool-bmv2.p4
Exit code  0
Check for  /home/hemant/p4c-1/testdata/p4_16_samples/header-bool-bmv2.stf
Running model
Running /home/hemant/behavioral-model/targets/simple_switch/simple_switch --log-file switch.log --log-flush --use-files 0 --thrift-port 9186 --device-id 96 ../bmv2/testdata/p4_16_samples/header-bool-bmv2.json
Invalid reference to object of type 'header' with name 'h'
Running /home/hemant/behavioral-model/targets/simple_switch/simple_switch_CLI --thrift-port 9186
STF Command: 
Could not connect to thrift client on port 9186
Exception  [Errno 3] No such process
hemant@ubuntu:~/p4c-1/build$ 
```

mbudiu-vmw:
If you think you have found another bmv2 backend bug please file a new issue with a program that can reproduce it.

hesingh:
> If you think you have found another bmv2 backend bug please file a new issue with a program that can reproduce it.

Will do.   Right now, I am debugging the issue.

hesingh:
I filed a new Issue:

https://github.com/p4lang/p4c/issues/1672

PR #1655, opened 2019-01-11
no need to run ClearTypeMap

hanw:
running travis..

### DIFF
--- a/midend/flattenHeaders.cpp
+++ b/midend/flattenHeaders.cpp
@@ -129,7 +129,7 @@ const IR::Node* ReplaceHeaders::postorder(IR::Member* expression) {
     }
 
     result = new IR::Member(expression->type, e, newFieldName);
-    LOG3("FH Replacing " << expression << "(" << expression->type << ") with " << result << "(" << result->type << ")");
+    LOG3("FH Replacing " << expression << " with " << result);
     typeMap->setType(result, expression->type);
     typeMap->setLeftValue(result);
     return result;
--- a/midend/flattenHeaders.cpp
+++ b/midend/flattenHeaders.cpp
@@ -128,8 +128,10 @@ const IR::Node* ReplaceHeaders::postorder(IR::Member* expression) {
                   "FH cannot find replacement for %s in type %s", prefix, h);
     }
 
-    result = new IR::Member(e, newFieldName);
-    LOG3("FH Replacing " << expression << " with " << result);
+    result = new IR::Member(expression->type, e, newFieldName);
+    LOG3("FH Replacing " << expression << "(" << expression->type << ") with " << result << "(" << result->type << ")");
+    typeMap->setType(result, expression->type);
+    typeMap->setLeftValue(result);
     return result;
 }
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -44,7 +44,7 @@ ENV P4C_RUNTIME_DEPS cpp \
 ENV P4C_PIP_PACKAGES tenjin \
                      pyroute2 \
                      ply \
-                     scapy
+                     scapy==2.4.0
 COPY . /p4c/
 WORKDIR /p4c/
 RUN apt-get update && \
--- a/backends/ebpf/targets/target.py
+++ b/backends/ebpf/targets/target.py
@@ -26,8 +26,7 @@
 import os
 import sys
 from glob import glob
-from scapy.utils import rdpcap
-from scapy.layers.all import RawPcapWriter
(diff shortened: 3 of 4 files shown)


hesingh (commented):
So why does nestedStructs.h does use the same code?
```
class NestedStructs final : public PassManager {
 public:
    NestedStructs(ReferenceMap* refMap, TypeMap* typeMap) {
        auto values = new ComplexValues(refMap, typeMap);
        passes.push_back(new TypeChecking(refMap, typeMap));
        passes.push_back(new RemoveNestedStructs(values));
        passes.push_back(new ClearTypeMap(typeMap));
        setName("NestedStructs");
    }
}; ```

**Comment by hesingh** [comment]
Even the FlattenInterfaceStructs pass uses the same code.  Please have Mihai review your code changes.  Also, there may be other midend passes that use similar code that I pointed out.

```
class FlattenInterfaceStructs final : public PassManager {
 public:
    FlattenInterfaceStructs(ReferenceMap* refMap, TypeMap* typeMap) {
        auto sm = new NestedStructMap(refMap, typeMap);
        passes.push_back(new TypeChecking(refMap, typeMap));
        passes.push_back(new FindTypesToReplace(sm));
        passes.push_back(new ReplaceStructs(sm));
        passes.push_back(new ClearTypeMap(typeMap));
        setName("FlattenInterfaceStructs");
    }
};
```

**Comment by mbudiu-vmw** [comment]
I actually think it does need to use ClearTypeMap. Anytime the type of an IR node can change you need to rerun ClearTypeMap. If you replace a struct with another one the type of variables instantiated from that struct will change.

**Comment by hanw** [comment]
Not sure, I am running Travis to see what it breaks. Ideally, none of the midend pass should run ClearTypeMap() as a particular midend may introduce additional IR type. ClearTypeMap will throw away any modification to the typeMap including the target-specific type info.

**Comment by hanw** [comment]
Can we use typeMap->setType for this purpose?

**Comment by mbudiu-vmw** [comment]
You can certainly fix the types manually, but it is a difficult process.
In general we don't have any promises on the typeMap except the fact that it always represents types correctly. You should use other data structures if you want to carry information from one pass to the next.

**Comment by mbudiu-vmw** [comment]
The rule is that if an IR node is a key in the TypeMap then its type in the typeMap is correct.
We run typechecking frequently and it promises that after running all nodes in the IR that have types have a correct type in the typeMap. 
You should always be able to clear the typeMap - it should only have a performance impact, not a correctness impact. But you do have to clear it if some IR nodes do not have a correct type.

**Comment by mbudiu-vmw** [comment]
Unfortunately we do have some back-end specific passes which change node types. Including eliminating enums - the policy used to replace enums is target-dependent.

**Comment by hanw** [comment]
ClearTypeMap makes it difficult to reuse midend passes, because TypeInference does not understand any IR type that is not defined in the frontend. For instance, suppose I introduce a IR type for cast that only does sign conversion, e.g. IR::ReinterpretCast, and suppose I manually updated the type map with `typeMap->setType(newnode, oldnode->destType)`;  The program type checks fine until the next ClearTypeMap(), then the program won't typecheck anymore. 

Do you have any suggestion on how to extend TypeInference to understand user-defined IR types?

**Comment by mbudiu-vmw** [comment]
If this is a target-specific IR node then the only solution is to subclass the type inference and add a method for your new node, then use the new type inference pass everywhere.

**Review by mihaibudiu — COMMENTED** [review]
When I tried this change the program stopped working.

File: midend/flattenHeaders.cpp [hunk 1]
```diff
@@ -128,8 +128,10 @@ const IR::Node* ReplaceHeaders::postorder(IR::Member* expression) {
                   "FH cannot find replacement for %s in type %s", prefix, h);
     }
 
-    result = new IR::Member(e, newFieldName);
+    result = new IR::Member(expression->type, e, newFieldName);
```
mbudiu-vmw: This may not be sufficient. There may be other expressions that involve the struct whose type may change.
Also, in general expression->type is stale; it is not maintained by the compiler. If you want to set the type use the type in the typeMap.

mbudiu-vmw: I see that you ask the typechecker to set these types; maybe the type won't be stale at this point.

mbudiu-vmw: For example, the type of any method that handles this kind of structures will change too.

mihaibudiu (changes requested):

hanw:
closed and will take a different approach.

PR #1662, opened 2019-01-14
extend strength reduction pass to handle concat operator with slice

hanw:


frontends/p4/strengthReduction.cpp
```diff
@@ -265,4 +265,59 @@ const IR::Node* StrengthReduction::postorder(IR::Slice* expr) {
     return expr;
 }
 
+const IR::Node* StrengthReduction::simplifyConcat(IR::Slice* expr) {
+    if (!expr->e0->is<IR::Concat>())
+        return expr;
+
```
ChrisDodd: This seems massively more complex than it needs to be, and doesn't even handle 3+ element concats.  Why not a simple loop like:

        while (auto cat = expr->e0->to<IR::Concat>()) {
            unsigned rwidth = cat->right->type->width_bits();
            if (expr->getL() >= shift_amt) {
                shift_of = cat->left;
                shift_amt = -rwidth
            } else if (sl->getH() < rwidth) {
                expr->e0 = cat->right;
            } else {
                return new IR::Concat(sl->type, new IR::Slice(cat->left, slice->getH() - rwidth, 0), new IR::Slice(cat->right, rwdith-1, slice->getL());
            }
        }

Though the above would be even better with `MakeSlice` to fold constants in those newly created slices.

In particular, a Transform or Modifier pre- or postorder method should almost never call new to create a new object of the same type -- the visitor infrastructure has already created a clone that should be modified if need be.

This won't handle all weird combos like `((a ++ b) >> 3)[5:2]` which could in theory be simpified, but its not clear that's worth the work.  Could be a way of making it work by having the postorder function recursively call itself, instead of this while loop (but need to be careful of infinite loop)

hanw: what is `sl` in the above code snippet?

frontends/p4/strengthReduction.cpp
```diff
@@ -75,6 +86,10 @@ const IR::Node* DoStrengthReduction::postorder(IR::BAnd* expr) {
         return expr->left;
     if (isZero(expr->right))
         return expr->right;
+    if (isAllOnes(expr->left))
```
mbudiu-vmw: thanks, this may modify some reference outputs.

Issue #1660, opened 2019-01-12
BMv2 back-end sometimes generates incorrect code for booleans


mbudiu-vmw:
Here is an example
```
#include <v1model.p4>

struct HasBool {
    bool x;
}

struct parsed_packet_t {}
struct local_metadata_t {}

parser parse(packet_in pk, out parsed_packet_t h, inout local_metadata_t local_metadata,
             inout standard_metadata_t standard_metadata) {
    state start {
	transition accept;
    }
}

control ingress(inout parsed_packet_t h, inout local_metadata_t local_metadata,
                inout standard_metadata_t standard_metadata) {
    apply {
        HasBool b;
        b.x = true;
        clone3(CloneType.I2E, 0, b);
    }
}

control egress(inout parsed_packet_t hdr, inout local_metadata_t local_metadata,
               inout standard_metadata_t standard_metadata) {
    apply { }
}

control deparser(packet_out b, in parsed_packet_t h) {
    apply { }
}

control verify_checksum(inout parsed_packet_t hdr,
inout local_metadata_t local_metadata) {
    apply { }
}

control compute_checksum(inout parsed_packet_t hdr,
                         inout local_metadata_t local_metadata) {
    apply { }
}

V1Switch(parse(), verify_checksum(), ingress(), egress(),
compute_checksum(), deparser()) main;
```

PR #1664, opened 2019-01-15
Do not insert casts for fields in field lists; remove ugly macros. 

### LINKED ISSUE #1471: Replace macros with function declarations in bmv2/psa back-end code
Making this into an issue, since it has been left unresolved from a previous PR.

mbudiu-vmw:
Fixes #1471 and #1660.
The fix for #1660 is in file bmv2/common/extern.cpp.
All other files have fixes for #1471.

### DIFF
--- a/backends/bmv2/common/extern.h
+++ b/backends/bmv2/common/extern.h
@@ -72,94 +72,71 @@ class ExternConverter {
     static cstring convertHashAlgorithm(cstring algorithm);
 };
 
-#define EXTERN_CONVERTER_W_FUNCTION_AND_MODEL(extern_name, model_type, model_name, ...)  \
-    class ExternConverter_##extern_name##__VA_ARGS__ : public ExternConverter { \
+#define EXTERN_CONVERTER_W_FUNCTION_AND_MODEL(extern_name, model_type, model_name)  \
+    class ExternConverter_##extern_name : public ExternConverter {              \
         model_type&  model_name;                                                \
-        ExternConverter_##extern_name##__VA_ARGS__() :                          \
+        ExternConverter_##extern_name() :                                       \
             model_name(model_type::instance) {                                  \
             registerExternConverter(#extern_name, this); }                      \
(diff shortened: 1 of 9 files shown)


backends/bmv2/common/extern.h
```diff
@@ -72,95 +72,72 @@ class ExternConverter {
...
+            const IR::Declaration* c, const IR::ExternBlock* eb,                \
+            const bool& emitExterns) override;                                  \
         Util::IJson* convertExternObject(ConversionContext* ctxt,               \
             const P4::ExternMethod* em, const IR::MethodCallExpression* mc,     \
             const IR::StatOrDecl* s, const bool& emitExterns) override; };
 
-#define EXTERN_CONVERTER_SINGLETON(extern_name, ...)                           \
-    ExternConverter_##extern_name##__VA_ARGS__                                 \
-       ExternConverter_##extern_name##__VA_ARGS__::singleton;
-
```
ChrisDodd: I'm not sure that removing these 4 macros is a cleanup, or just makes things harder to maintain.

mbudiu-vmw: The program isn't any longer...

mbudiu-vmw: Actually this macro doesn't make the program shorter, the next 3 do.

ChrisDodd: I think it is desirable to make clear which externs are objects and which are functions.  To the extent that there is anything common, one would prefer to avoid replication, as that can lead to subtle problems when there are differences -- when looking at a repeating pattern of source code, one's eyes tend to concentrate on the replicated pattern and not notice the differences, when it is the differences that are important, not the pattern.

ChrisDodd: It may be that these 4 macros should be replaced by 2 (object and function), or 3 (if the object/instance difference is important), with the singleton declaration folded into the OBJECT or FUNCTION macro.

PR #1665, opened 2019-01-15
Avoid unneeded copies when inlining controls

ChrisDodd:
When inlining controls, we check for aliasing and avoid creating copies of arguments if there is no possibility of aliasing, but currently that is restricted to a single call site.  This change extends that to multiple call sites if all the call sites use the same arguments.  If the arguments are different, we would need copies in at least some cases (to get them into the same place), but if all call sites are the same, copies are not needed.

Currently this only avoids the copies if the call sites are completely the same -- if the call sites had the same arguments for some parameters and not others, copies would be needed only for the parameters that were different between calls sites, but that would be a bit more work.

There's also a bunch of cleanups that crept into this branch.

### DIFF
--- a/frontends/p4/commonInlining.h
+++ b/frontends/p4/commonInlining.h
@@ -157,7 +157,7 @@ class AbstractInliner : public Transform {
 };
 
 template <class InlineList, class InlineWorkList>
-class InlineDriver : public Transform {
+class InlineDriver : public Visitor {
     InlineList*     toInline;
     AbstractInliner<InlineList, InlineWorkList>* inliner;
 
@@ -166,17 +166,14 @@ class InlineDriver : public Transform {
         InlineList* toInline, AbstractInliner<InlineList, InlineWorkList> *inliner) :
             toInline(toInline), inliner(inliner)
     { CHECK_NULL(toInline); CHECK_NULL(inliner); setName("InlineDriver"); }
-    // Not really a visitor, but we want to embed it into a PassManager,
-    // so we make it look like a visitor.
-    const IR::Node* preorder(IR::P4Program* program) override {
+    const IR::Node* apply_visitor(const IR::Node *program, const char * = 0) override {
         LOG2("InlineDriver");
-        const IR::P4Program* prog = program;
         toInline->analyze();
         LOG3("InlineList size " << toInline->size());
         while (auto todo = toInline->next()) {
             LOG2("Processing " << todo);
             inliner->prepare(toInline, todo);
-            prog = prog->apply(*inliner);
+            program = program->apply(*inliner);
             if (::errorCount() > 0)
                 break;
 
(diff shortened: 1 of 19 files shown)


frontends/p4/commonInlining.h
```diff
@@ -157,7 +157,7 @@ class AbstractInliner : public Transform {
 };
 
 template <class InlineList, class InlineWorkList>
-class InlineDriver : public Transform {
+class InlineDriver : public Visitor {
```
mbudiu-vmw: This suggests that the driver does not change anything, which is not really true.

frontends/p4/inlining.cpp
```diff
@@ -500,9 +500,20 @@ const IR::Node* GeneralInliner::preorder(IR::P4Control* caller) {
             // Use temporaries for these parameters
             std::set<const IR::Parameter*> useTemporary;
 
-            auto call = workToDo->uniqueCaller(inst);
+            const IR::MethodCallStatement *call = nullptr;
+            for (auto m : workToDo->callToInstance) {
+                if (m.second != inst) continue;
+                if (call) {
+                    if (!call->equiv(*m.first)) {
```
mbudiu-vmw: is this precise enough to distinguish between two identifiers with the same name but different scopes?

frontends/p4/typeChecking/typeChecker.cpp
```diff
@@ -39,8 +39,7 @@ class ConstantTypeSubstitution : public Transform {
                              ReferenceMap* refMap,
                              TypeMap* typeMap) : subst(subst), refMap(refMap), typeMap(typeMap) {
         CHECK_NULL(subst); CHECK_NULL(refMap); CHECK_NULL(typeMap);
-        LOG3("ConstantTypeSubstitution " << subst);
-        setName("ConstantTypeSubstitution"); }
```
mbudiu-vmw: I use --top4 passSubstring often; I hope that removing the setName won't affect that.

testdata/p4_14_samples_outputs/issue-1426-midend.p4
```diff
@@ -109,78 +108,20 @@ control ingress(inout headers hdr, inout metadata meta, inout standard_metadata_
         size = 1024;
         default_action = NoAction_7();
     }
-    @hidden action act() {
```
mbudiu-vmw: these are nice savings

frontends/p4/inlining.cpp
ChrisDodd: Can't happen as the previous UniqueNames pass will have renamed them to be different.

frontends/p4/typeChecking/typeChecker.cpp
ChrisDodd: It fixes that -- when there are Visitor classes that inherit from each other, we want to use the name of the most-derived class, not the name of the base class.

frontends/p4/commonInlining.h
ChrisDodd: Visitor is the base class of all visitors, so encapsulates both visitors that change things and visitors that do not.  Deriving directly from Visitor means it does not use the normal Inspector/Modifier/Transform depth-first traversal.

ChrisDodd: All of our Visitors *except* Inspectors (potentially) mutate the program.  Other direct Visitor subclasses like PassManager and PassRepeated may modify the program.  I guess we could add a comment to all such passes that they might modify the program, but it seems obvious from them not being Inspectors.

PR #1666, opened 2019-01-15
Bitvec Barrel Shift

evanches:
Added a barrel shift for bitvec to rotationally shift around a pivot.  Not sure what to do with the corner case where pivot < max().index() of that bitvec.

### DIFF
--- a/lib/bitvec.cpp
+++ b/lib/bitvec.cpp
@@ -16,6 +16,7 @@ limitations under the License.
 
 #include "bitvec.h"
 #include "hex.h"
+#include "lib/exceptions.h"
 
 std::ostream &operator<<(std::ostream &os, const bitvec &bv) {
     if (bv.size == 1) {
@@ -144,4 +145,21 @@ bool bitvec::is_contiguous() const {
     if (empty())
         return false;
     return max().index() - min().index() + 1 == popcount();
+}
+
+/**
+ * The purpose of this function is to barrel_shift a bitvec around a pivot.  This function
+ * will crash if the pivot is smaller than the size of the bitvector
+ */
+bitvec bitvec::barrel_shift_left(size_t shift, size_t pivot) const {
+    if (max().index() > pivot) {
+        BUG("Barrel shifting around pivot %d is smaller than the max of the bitvec %d", pivot,
+            max().index());
+        return *this;
+    }
+    int single_shift = shift % pivot;
+    bitvec rv;
+    rv |= *this << single_shift;
+    rv |= *this >> (pivot - single_shift);
+    return rv & bitvec(0, pivot);
 }
--- a/lib/bitvec.h
+++ b/lib/bitvec.h
@@ -435,6 +435,7 @@ class bitvec {
     bitvec &operator<<=(size_t count);
     bitvec operator>>(size_t count) const { bitvec rv(*this); rv >>= count; return rv; }
     bitvec operator<<(size_t count) const { bitvec rv(*this); rv <<= count; return rv; }
+    bitvec barrel_shift_left(size_t shift, size_t pivot) const;
     int popcount() const {
         int rv = 0;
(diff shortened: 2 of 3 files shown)


lib/bitvec.cpp
```diff
@@ -145,3 +146,20 @@ bool bitvec::is_contiguous() const {
         return false;
     return max().index() - min().index() + 1 == popcount();
 }
+
+/**
+ * The purpose of this function is to barrel_shift a bitvec around a pivot.  This function
+ * will crash if the pivot is smaller than the size of the bitvector
+ */
+bitvec bitvec::barrel_shift_left(size_t shift, size_t pivot) const {
```
ChrisDodd: It seems this is a rotate, not a barrel shift.  Might be more cleanly/clearly expressed as such, with 3 args (rotate sz bits at offset idx by shift).

lib/bitvec.cpp
```diff
@@ -145,3 +146,20 @@ bool bitvec::is_contiguous() const {
...
+    if (max().index() > pivot) {
+        BUG("Barrel shifting around pivot %d is smaller than the max of the bitvec %d", pivot,
+            max().index());
```
ChrisDodd: This shouldn't be a problem -- the below code should work fine for this, rotating 0 bits from above max().index() just fine.

lib/bitvec.cpp
```diff
@@ -145,3 +146,20 @@ bool bitvec::is_contiguous() const {
...
+        return *this;
+    }
+    int single_shift = shift % pivot;
+    bitvec rv;
+    rv |= *this << single_shift;
+    rv |= *this >> (pivot - single_shift);
+    return rv & bitvec(0, pivot);
```
ChrisDodd: `rv.clrrange(pivot, ~0);` would be slightly more efficient for this last step (in place clear rather than creating a bitmask and anding)

lib/bitvec.cpp
ChrisDodd: suggest:

    // In place rotate (left) of sz bits starting at idx by shift
    bitvec &rotate(size_t idx, size_t sz, size_t shift) {
        bitvec bits = getslice(idx, sz);
        bits <<= idx + shift % sz;
        bits |= bits >> (sz - shift % sz);
        clrrange(idx, sz);
        bit.clrrange(0, idx);
        bit.clrrange(idx + sz, ~0);
        *this |= bits;
        return *this;
    }

could be done not in place as well.

evanches:
@ChrisDodd change to rotate function, more to reflect the std::rotate function of c++

lib/bitvec.cpp
```diff
@@ -145,3 +146,34 @@ bool bitvec::is_contiguous() const {
...
+    bitvec rotation_section = *this & rot_mask; 
+    int down_shift = rotation_idx - start_bit;
+    int up_shift = (end_bit - start_bit) - down_shift;
+    bitvec rv;
+    rv = (rotation_section >> down_shift) | (rotation_section << up_shift);
+    return rv & rot_mask; 
+}
+
+/**
+ * Designed to imitate the std::rotate/std::rotate_copy function for vectors.  Return a bitvec
```
ChrisDodd: Note that std::rotate uses big-endian indexing, so a left shift is shifting down (towards index 0), while we use little endian indexing for bitvecs, so a left shift is shifting up (away from index 0).  Very confusing.

evanches: Would you prefer a different implementation?  I can make the rotate function shift the other way.  I was just going off of a base implementation.

evanches: or I can rename the functions to rotate_right/rotate_copy_right if you'd prefer that as well, for much clearer naming.

ChrisDodd: I think its fine -- might just want to note the right/left difference in the comment.

PR #1667, opened 2019-01-16
Changed FlattenHeaders to explicitly handle Type_Header and Type_Struct

hanw:


hanw:
I have another data structure that is a subclass of Type_StructLike, which I am not willing to flatten by this pass. This pass is a little too general for me.

mbudiu-vmw:
then you should check explicitly for the classes that need to be flattened.

hesingh:
The pass is FlattenHeader and as the name signifies, it's flattening starts with a header which has a struct to be flattened.  Such a header flattening will fail.  You will see once you invoke 'make check' tests will fail.  Give us till tomorrow to think how to address your use case.

hanw:
@hesingh I am hoping we can make this pass more precise and to only flattens Type_Struct, Type_Header and Type_HeaderStack. Nothing else.

hesingh:
> @hesingh I am hoping we can make this pass more precise and to only flattens Type_Struct, Type_Header and Type_HeaderStack. Nothing else.

Midend passes in FlattenInterfaceStructs and NestedStructs flatten Type_Struct.  This is why the FlattenHeaders pass only deals with flattening Type_Header.   I will get back wrt Type_HeaderStack.

hesingh:
Regarding Header stack, the FlattenHeaders pass supports it.  See this test.

testdata/p4_16_samples/stack-bvec-bmv2.p4

Additionally, see this code in flattenHeaders.cpp which takes care of any Header stack.

```
bool FindHeaderTypesToReplace::preorder(const IR::Type_Header* type) {
    // check if header has structs in it and create flat replacement if it does
    for (auto f : type->fields) {
        auto ft = map->typeMap->getType(f, true);
        if (ft->is<IR::Type_Struct>()) {
            map->createReplacement(type);
            break;
        }
    }
    return false;
}
```

cc10512:
Let me see if I can clarify this request. There are two types of P4 programs objects that are defined in the language: headers and structs. In our view the "common" flattening optimization should act on those specific objects, rather than on the base type that may be derived from for other purposes. In this way, it is clear to a backend developer that if they derive from a `IR::Type_Header` or `IR::Type_Struct` they will get the flattening, otherwise they should have a choice of calling flattening on their derived type or not. Unfortunately, with the current implementation, FlattenHeaders will flatten all derived classes from `IR::Type_StructLike`. Does this make sense?

hesingh:
> Let me see if I can clarify this request. There are two types of P4 programs objects that are defined in the language: headers and structs. In our view the "common" flattening optimization should act on those specific objects, rather than on the base type that may be derived from for other purposes. In this way, it is clear to a backend developer that if they derive from a `IR::Type_Header` or `IR::Type_Struct` they will get the flattening, otherwise they should have a choice of calling flattening on their derived type or not. Unfortunately, with the current implementation, FlattenHeaders will flatten all derived classes from `IR::Type_StructLike`. Does this make sense?

We (I and Alex) have said numerous times in this Issue that the FlattenHeaders pass is ONLY handling flattening of Type_Header.  I am adding this caller code for flatten again to prove the statements.  Please see "<===" in the code below. 

```
HeaderTypeReplacement::HeaderTypeReplacement(
    const P4::TypeMap* typeMap, const IR::Type_Header* type) {  <===
    auto vec = new IR::IndexedVector<IR::StructField>();
    flatten(typeMap, "", type, vec);  <=== 
```

Likewise the other pass in FlattenInterfaceStructs is ONLY handling Type_Struct.  See its code below.
```
StructTypeReplacement::StructTypeReplacement(
    const P4::TypeMap* typeMap, const IR::Type_Struct* type) {  <===
    auto vec = new IR::IndexedVector<IR::StructField>();
    flatten(typeMap, "", type, vec); <===
```

hanw:
@hesingh The problem is orthogonal to what you described. The current issue is not that FlattenHeaders can only flatten Type_Header. 

The issue is that the FlattenHeaders flattens the StructField in a Type_Header that is Type_StructLike, which means that it not only flattens StructField that is Type_Struct, but also flattens any derived class of Type_StructLike, such as Type_Struct_my_random_struct_type, which may be undesirable.

hesingh:
> @hesingh The problem is orthogonal to what you described. The current issue is not that FlattenHeaders can only flatten Type_Header.
> 
> The issue is that the FlattenHeaders flattens any StructFields in a Type_Header that is Type_StructLike, which means that it not only flattens StructField that is Type_Struct, but also flattens any derived class of Type_StructLike, such as Type_Struct_my_random_struct_type, which may be undesirable.

No.  See the comments from pofig37 for this issue.  I have copied the comments here within quotes:

"First time flatten() is called, it's called on Type_Header().  So with this change it wouldn't recurse down into the fields. Checking for StructLike is OK, because frontend ensures that there's nothing but structs inside a header."  

Likewise, first time flatten in the FlattenInterfaceStructs pass is called, it's called on Type_Struct().  Even for this case, the frontend ensures that there is nothing but structs inside a struct.

See the frontend code in typeChecker.cpp and the onlyBitsOrBitStructs() method.

Nesting of P4 header is not supported.

hanw:
I have updated the code, hopefully the intent is clearer.

ChrisDodd:
Perhaps if the midend wants to avoid flattening nested structs in some other Type_StructLike subclass, it should not call FlattenHeaders directly, but should instead subclass is with a visitor the treats these other Type_StructLike subclasses specially and avoids flattening them?

hesingh:
> I have updated the code, hopefully the intent is clearer.

I disagree with the code changes.  What to flatten or not is not decided inside the flatten method.  The caller to flatten has nailed down the top-level object to flatten is Type_Header.   The frontend code has decided that a header can nest only a bit-vector struct.  So, flatten using Type_StructLike was fine.  It was brief code and now the changed code is duplicating code with explicit checks for struct and header.  Code duplication is not production code and causes maintenance issues -  fixing a bug in one place needs fixing in another.  

In regards to "I have another data structure that is a subclass of Type_StructLike, which I am not willing to flatten by this pass.", my question is how did your data structure even get to the flatten midend pass?  The frontend has not allowed your 'another data structure' to be nested inside a header.
 Now you checked in code to the master that hasn't solved anything.

PR #1668, opened 2019-01-16
change setTypeType to virtual to allow override.

hanw:
Otherwise, when method in derived class calls setTypeType, the setTypeType
method will create an instance of the base class TypeInference and apply it.

### DIFF
--- a/frontends/p4/typeChecking/typeChecker.h
+++ b/frontends/p4/typeChecking/typeChecker.h
@@ -159,7 +159,7 @@ class TypeInference : public Transform {
     static constexpr bool forbidPackages = true;
     bool checkParameters(const IR::ParameterList* paramList,
                          bool forbidModules = false, bool forbidPackage = false) const;
-    const IR::Type* setTypeType(const IR::Type* type, bool learn = true);
+    virtual const IR::Type* setTypeType(const IR::Type* type, bool learn = true);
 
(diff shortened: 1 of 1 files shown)


frontends/p4/typeChecking/typeChecker.h
```diff
@@ -159,7 +159,7 @@ class TypeInference : public Transform {
     static constexpr bool forbidPackages = true;
     bool checkParameters(const IR::ParameterList* paramList,
                          bool forbidModules = false, bool forbidPackage = false) const;
-    const IR::Type* setTypeType(const IR::Type* type, bool learn = true);
+    virtual const IR::Type* setTypeType(const IR::Type* type, bool learn = true);
```
mbudiu-vmw: it's strange that you need to override this.

hanw: It's because this line. https://github.com/p4lang/p4c/blob/master/frontends/p4/typeChecking/typeChecker.cpp#L1056
It creates an instance of the base class, not the derived class.

mbudiu-vmw:
This happens in several other places.
I don't know how easy it would be to create a clone method for the visitor, but that's what's really needed.

hanw:
@ChrisDodd ?

hanw:
I am going to merge this for now. Will talk to Chris to see if there is a better solution.

PR #1673, opened 2019-01-16
improved strength reduction pass

hanw:


testdata/p4_16_samples_outputs/strength4-first.p4
```diff
@@ -0,0 +1,55 @@
...
+}
+
+control ingress_impl(inout headers_t hdr, inout local_metadata_t local_metadata, inout standard_metadata_t standard_metadata) {
+    apply {
+    }
+}
+
+control egress_impl(inout headers_t hdr, inout local_metadata_t local_metadata, inout standard_metadata_t standard_metadata) {
+    apply {
+        local_metadata.m16 = (16w0 ++ local_metadata.f16)[14:0] ++ 1w0;
```
mbudiu-vmw: this seems sub-optimal, you should be able to drop the concat altogether

hanw: it is simplified in testdata/p4_16_samples_outputs/strength4-midend.p4

frontends/p4/strengthReduction.cpp
```diff
@@ -276,7 +276,13 @@ const IR::Node* DoStrengthReduction::postorder(IR::Slice* expr) {
         if (lo + shift_amt >= 0 && hi + shift_amt < shift_of->type->width_bits()) {
             expr->e0 = shift_of;
             expr->e1 = new IR::Constant(hi + shift_amt);
-            expr->e2 = new IR::Constant(lo + shift_amt); } }
+            expr->e2 = new IR::Constant(lo + shift_amt); }
```
mbudiu-vmw: is this only done when types are unsigned?
does this work for signed values?

mbudiu-vmw: maybe you can add a test for signed values too

hanw: It does not work for signed value. How do we represent the result of strenght-reducing right shift on a signed value in P4.

for instance. 
```
(local_meta.s16 >> 1)[15:0]
```
How do we know if which of the following expression is correct if we do not know the value of the signed bit?
```
1s0 ++ local_meta.s16[15:1]
```
or 
```
1s1 ++ local_meta.s16[15:1]
```

mbudiu-vmw: Yes, but you are not checking that the types are unsigned.
You may be applying this for signed values too.

hanw: I have added a test showing strength reduction on shift does nothing to signed values.

testdata/p4_16_samples_outputs/strength4-first.p4
ChrisDodd: I guess because the midend runs StrengthReduction again?  So if we had a recursive call here, it would get simplified in the frontend...

frontends/p4/strengthReduction.cpp
```diff
@@ -271,12 +271,22 @@ const IR::Node* DoStrengthReduction::postorder(IR::Slice* expr) {
...
         int lo = expr->getL();
         if (lo + shift_amt >= 0 && hi + shift_amt < shift_of->type->width_bits()) {
             expr->e0 = shift_of;
             expr->e1 = new IR::Constant(hi + shift_amt);
-            expr->e2 = new IR::Constant(lo + shift_amt); } }
+            expr->e2 = new IR::Constant(lo + shift_amt); }
+        if (lo + shift_amt < 0) {
+            expr->e0 = shift_of;
+            expr->e1 = new IR::Constant(hi + shift_amt);
+            expr->e2 = new IR::Constant(0);
```
ChrisDodd: maybe should call `expr = postorder(expr);` here to recursively deal with anything nested in the expresion?  Or perhaps not worth it for such an obscure case.

PR #1674, opened 2019-01-16
add a simple stf to test current status with psa

hanw:
currently fails due to lack of support for nested struct in p4c-bm2-psa backend.


mbudiu-vmw:
@hanw : we now have a few tests with psa that are running, AFAICT, so perhaps you want to close this PR?

PR #1675, opened 2019-01-16
Add p4info files as reference outputs

mbudiu-vmw:


frontends/p4/typeChecking/typeChecker.cpp
```diff
@@ -118,8 +118,8 @@ const IR::Type* TypeInference::cloneWithFreshTypeVariables(const IR::IMayBeGener
 }
 
 TypeInference::TypeInference(ReferenceMap* refMap, TypeMap* typeMap, bool readOnly) :
-        refMap(refMap), typeMap(typeMap), readOnly(readOnly),
-        initialNode(nullptr) {
+        refMap(refMap), typeMap(typeMap),
+        initialNode(nullptr), readOnly(readOnly) {
```
mbudiu-vmw: this is cleaning up a warning introduced in a recent commit by @hanw

hesingh:
@mbudiu-vmw Can you also please change this code in typeChecker.cpp for checkEmitType() to change the 'continue;' to a BUG like you told me in our code in the same file in a  review.  thanks.

```
if (type->is<IR::Type_Struct>()) {
        for (auto f : type->to<IR::Type_Struct>()->fields) {
            auto ftype = typeMap->getType(f);
            if (ftype == nullptr)
                continue;
            checkEmitType(emit, ftype);
        }
        return;
    }
```

mbudiu-vmw:
@hesingh : your request does not seem to have anything to do with the issue that we are attempting to solve here.

hesingh:
> @hesingh : your request does not seem to have anything to do with the issue that we are attempting to solve here.

I thought since you were making another change in the same .cpp file, one could make another change.  However, it's fine to keep this change out of this PR and use another PR to make the change.  thanks.

antoninbas:
I just realized that the p4Runtime serializer doesn't really sort the objects (e.g. tables) in any specific order. But it seems that Travis passed, so hopefully the order is relatively stable (because the IR is traversed deterministically?).

mbudiu-vmw:
The IR traversal is deterministic. We tried to make the entire compiler deterministic.

antoninbas:
I'm just hoping that most changes to the compiler (e.g. new transform visitor) will not cause all the P4Info files to be changed. If I add a transform pass, it creates new instances of the IR nodes, which may impact the ordering in the IR data structures (e.g. IndexedVector), which may impact P4Info generation. How are objects sorted in the IR data structures?

antoninbas:
NVM, IndexedVector was a bad example since it is ordered based on the name...

jnfoster:
Nice.

PR #1676, opened 2019-01-17
Validate headers type for v1model

### LINKED ISSUE #1672: bmv2 backend fails header generation sometimes
Please use the p4c in the URL below which is being used as a repo for PR to add boolean to P4 header.

https://github.com/hesingh/p4c-1

Use the test in testdata/p4_16_sample/header-bool-bmv2.p4.  Run the test as shown below and see the error where the arrow is "<===".  The error is from P4Objects.cpp in the behavioral model which fails to find the hdr or h in the output json. 

```
hemant@ubuntu:~/p4c-1/build$ bmv2/testdata/p4_16_samples/header-bool-bmv2.p4.test -b -v
Writing temporary files into  ./tmp9tR5ek
Executing  ./p4c-bm2-ss -o bmv2/testdata/p4_16_samples/header-bool-bmv2.json /home/hemant/p4c-1/testdata/p4_16_samples/header-bool-bmv2.p4
Exit code  0
Check for  /home/hemant/p4c-1/testdata/p4_16_samples/header-bool-bmv2.stf
Running model
Running /home/hemant/behavioral-model/targets/simple_switch/simple_switch --log-file switch.log --log-flush --use-files 0 --thrift-port 9124 --device-id 34 ../bmv2/testdata/p4_16_samples/header-bool-bmv2.json
Invalid reference to object of type 'header' with name 'h' <===
Running /home/hemant/behavioral-model/targets/simple_switch/simple_switch_CLI --thrift-port 9124
STF Command: 
Could not connect to thrift client on port 9124
Exception  [Errno 3] No such process
hemant@ubuntu:~/p4c-1/build$ 

```

mbudiu-vmw:
Fixes #1672

### DIFF
--- a/backends/bmv2/simple_switch/simpleSwitch.cpp
+++ b/backends/bmv2/simple_switch/simpleSwitch.cpp
@@ -908,6 +908,33 @@ SimpleSwitchBackend::convert(const IR::ToplevelBlock* tlb) {
     userMetaType = decl->to<IR::Type_Struct>();
     LOG2("User metadata type is " << userMetaType);
 
+    {
+        // Find the header types
+        auto headersParam = params->parameters.at(1);
+        auto headersType = headersParam->type;
+        if (!headersType->is<IR::Type_Name>()) {
+            ::error("%1%: expected type to be a struct", headersParam->type);
+            return;
+        }
+        decl = refMap->getDeclaration(headersType->to<IR::Type_Name>()->path);
+        auto st = decl->to<IR::Type_Struct>();
+        if (st == nullptr) {
+            ::error("%1%: expected type to be a struct", headersParam->type);
+            return;
+        }
+        LOG2("Headers type is " << st);
+        for (auto f: st->fields) {
+            auto t = typeMap->getType(f, true);
+            if (!t->is<IR::Type_Header>() &&
+                !t->is<IR::Type_Stack>() &&
+                !t->is<IR::Type_HeaderUnion>()) {
+                ::error("%1%: the type of should be a struct of headers, stacks, or unions",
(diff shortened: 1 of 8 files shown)


hesingh:
Any reason, we don't catch the issue in frontend code?

Also, I'd like to see normative text in the P4-16 spec where the v1model.p4 restrictions are specified.   I don't see the P4-16 spec saying only headers, stacks or header_unions are allowed with Parser.

antoninbas:
> Any reason, we don't catch the issue in frontend code?
> 
> Also, I'd like to see normative text in the P4-16 spec where the v1model.p4 restrictions are specified. I don't see the P4-16 spec saying only headers, stacks or header_unions are allowed with Parser.

If it's architecture-specific, there isn't really a way to catch it in the frontend. I also don't see why the v1model.p4 architecture should be documented in the P4_16 spec. The lack of documentation for v1model.p4 has indeed been a major pain point, but I don't think it belongs in the language spec. @jafingerhut has written some documentation for simple_switch, which is also applicable to v1model: https://github.com/p4lang/behavioral-model/blob/master/docs/simple_switch.md

hesingh:
ok, got it - thanks.

jafingerhut:
Here is a PR proposing documenting the restriction that the P4_16 type parameter `H` in v1model.p4 must only contain header, header stacks, and header_union members.  https://github.com/p4lang/behavioral-model/pull/701

Certainly more can be written documenting the details of the v1model architecture than that, but I'm going about it piece-by-piece as I learn about those restrictions myself.

hesingh:
@jafingerhut Thanks!

PR #1677, opened 2019-01-17
Fix equiv generation to be easier to override properly

ChrisDodd:
- add Declarator::equiv and Type_Declarator::equiv that ignores declid,
  so repeated specializations with the same types will compare
  equivalent.

### DIFF
--- a/ir/base.def
+++ b/ir/base.def
@@ -118,6 +118,7 @@ abstract Declaration : StatOrDecl, IDeclaration {
     ID          name;
     int declid = nextId++;
     ID getName() const override { return name; }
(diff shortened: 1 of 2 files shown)


ChrisDodd:
`equiv` is used for deep comparisons of IR subtrees.  This specific change is needed to make things like Method_Types compare as equiv if they are equivalent -- each mathod type will contain a vector of paramters (and type parameters) which we want to compare against each other, which ends up calling the equiv method on the parameter Declarations -- we want to consider them equivalent if they are the same in context.

Arguably when comparing the parameter types of a method type, we should even ignore the names, and just consider the types.

PR #1680, opened 2019-01-18
Move a message to assert

evanches:
Better for building for sub libraries.

### DIFF
--- a/lib/bitvec.cpp
+++ b/lib/bitvec.cpp
@@ -16,7 +16,6 @@ limitations under the License.
 
 #include "bitvec.h"
 #include "hex.h"
-#include "lib/exceptions.h"
 
 std::ostream &operator<<(std::ostream &os, const bitvec &bv) {
     if (bv.size == 1) {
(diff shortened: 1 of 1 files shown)


lib/bitvec.cpp
```diff
@@ -148,8 +147,7 @@ bool bitvec::is_contiguous() const {
 }
 
 bitvec bitvec::rotate_right_helper(size_t start_bit, size_t rotation_idx, size_t end_bit) const {
-    BUG_CHECK(start_bit <= rotation_idx && rotation_idx < end_bit, "Invalid rotation on bitvec, as "
-              "rotation_idx does not fall between start_bit and end_bit");
+    assert(start_bit <= rotation_idx && rotation_idx < end_bit);
```
cc10512: Asserts disappear when you compile with -DNDEBUG, which is the default for CMAKE_BUILD_TYPE=Release. So your test coverage should ensure that you never call this function with out of bounds indices.

Issue #1670, opened 2019-01-16
Use of FlattenHeaders in p4runtime_serializer.cpp


hanw:
With attached file.

[serializer2.p4.txt](https://github.com/p4lang/p4c/files/2765443/serializer2.p4.txt)

You can reproduce the issue by copying the file to p4c/testdata/p4_16_samples/
```
./bootstrap.sh
ctest -R serializer2.p4
```

or run the p4test command directly with --p4runtime-file flag.

```
./p4test --pp ./tmpwLs0oI/serializer2.p4 --dump ./tmpwLs0oI --top4 FrontEndLast,FrontEndDump,MidEndLast --testJson --arch v1model --p4runtime-format text --p4runtime-file ./tmpwLs0oI/serializer2.p4.p4info.txt /Users/hanwang/p4c/testdata/p4_16_samples/serializer2.p4
Error compiling
libc++abi.dylib: terminating with uncaught exception of type Util::CompilerBug: In file: /Users/hanwang/p4c/midend/flattenHeaders.cpp:130
Compiler Bug: /Users/hanwang/p4c/testdata/p4_16_samples/serializer2.p4(51): FH cannot find replacement for .meta in type header serialized_switch_metadata_t {
  bit<8> type;
  switch_metadata_t meta; }
header serialized_switch_metadata_t {
       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

hesingh:
Please first compile any P4 program using p4c-bm2-ss and if compiling passes, then you can test other programs.  I used your program and run into the error below.  There isn't anything we can do about this - it's a correct error from the frontend. 

```
hemant@ubuntu:~/p4c-1/build$ ./p4c-bm2-ss --std p4-16 ../testdata/p4_16_samples/serializer2.p4
../testdata/p4_16_samples/serializer2.p4(64): error: b.emit: argument must be a header, stack or union, or a struct or tuple of such types
        b.emit(h.s);
        ^^^^^^^^^^^
../testdata/p4_16_samples/serializer2.p4(64): error: b.emit: argument must be a header, stack or union, or a struct or tuple of such types
        b.emit(h.s);
        ^^^^^^^^^^^
hemant@ubuntu:~/p4c-1/build$ 
```

hesingh:
Also, see the Issue below which gives the same error.

https://github.com/p4lang/p4c/issues/1659

hesingh:
[...]

PR #1682, opened 2019-01-18
do not flatten headers for control-plane API; temporary solution for …

mbudiu-vmw:
…#1670

### DIFF
--- a/testdata/p4_16_samples/issue1670-bmv2.p4
+++ b/testdata/p4_16_samples/issue1670-bmv2.p4
@@ -1,0 +1,75 @@
+/*******************************************************************************
+ * BAREFOOT NETWORKS CONFIDENTIAL & PROPRIETARY
+ *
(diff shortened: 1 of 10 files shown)


testdata/p4_16_samples_outputs/issue1670-bmv2.p4.p4info.txt
```diff
@@ -0,0 +1,3 @@
+pkg_info {
```
antoninbas: does it make sense to have a p4info file if the test is in `P4RUNTIME_EXCLUDE`, how did the file even get generated?

testdata/p4_16_samples/issue1670-bmv2.p4
```diff
@@ -0,0 +1,75 @@
+/*******************************************************************************
```
antoninbas: This is a proprietary license header, not an Apache license header. So either this file should not be here or the header is wrong.

PR #1686, opened 2019-01-21
move clone() method from ControlFlowVisitor to Visitor

hanw:
A class that introducing new IR class needs to subclass TypeInference and overriding the clone() method.

### DIFF
--- a/ir/visitor.h
+++ b/ir/visitor.h
@@ -314,7 +314,7 @@ class Transform : public virtual Visitor {
 class ControlFlowVisitor : public virtual Visitor {
     std::map<const IR::Node *, std::pair<ControlFlowVisitor *, int>> *flow_join_points = 0;
  protected:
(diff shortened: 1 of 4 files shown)


ir/visitor.h
```diff
@@ -312,7 +314,7 @@ class Transform : public virtual Visitor {
 class ControlFlowVisitor : public virtual Visitor {
     std::map<const IR::Node *, std::pair<ControlFlowVisitor *, int>> *flow_join_points = 0;
  protected:
-    virtual ControlFlowVisitor *clone() const = 0;
+    ControlFlowVisitor* clone() const override;
```
ChrisDodd: Is it not possible to define this as a abstract method? (`= 0;`).  If not, its probably better to remove it completely rather than overriding Visitor::clone with an identical method.

hanw: fixed.

Issue #1687, opened 2019-01-21
p4c test program issue1653-bmv2.p4 is perhaps a bad test


jafingerhut:
https://github.com/p4lang/p4c/blob/master/testdata/p4_16_samples/issue1653-bmv2.p4#L49

The field list given to the clone operation contains a packet header.

There are other potential issues with clone/recirc/resub and lists of metadata fields that are given in other issues, but at least for P4_14, which is the origin of these operations and the reason they are present in P4_16's v1model architecture, the specification says:

"Only metadata (and not header) fields can be specified in the field_list." (documentation for clone_ingress_pkt_to_egress operation)
"Only meta- data (and not header) fields can be specified in the field_list." (documentation for clone_egress_pkt_to_ingress operation)

Similar phrases were added for recirculate and resubmit operations in v1.0.5 of the P4_14 spec.

p4c does not give an error when you attempt to use a field_list for these operations that violates those sentences, but perhaps it should?

In any case, it seems like a good idea not to create test cases that expect good behavior for an unsupported case like this.

hesingh:
Great catch.  Sorry, I missed the behavior about Clone3 during creating the test.  

Yes, I agree the compiler should give an error when a header is used with Clone3.

hesingh:
@mbudiu-vmw How come the bmv2 backend does not find an entry in the typeMap which I looked for during processing a clone3() call?  I changed the bmv2 backend code to try and catch a header inside a struct given to clone3 as 3rd arg.  

My test log is followed by the code changes.

```
hemant@ubuntu:~/p4c-cplane/build$ ./p4c-bm2-ss --std p4-16 ../testdata/p4_16_samples/issue1653-bmv2.p4
Checking Clone3 arg3: struct parsed_packet_t {
  bitvec_hdr bvh0;
  bitvec_hdr bvh1; }
error: ftype is null
error: ftype is null
hemant@ubuntu:~/p4c-cplane/build$ 
```

```
diff --git a/backends/bmv2/simple_switch/simpleSwitch.cpp b/backends/bmv2/simple_switch/simpleSwitch.cpp
index c19fa5ea..a7d0ada5 100644
--- a/backends/bmv2/simple_switch/simpleSwitch.cpp
+++ b/backends/bmv2/simple_switch/simpleSwitch.cpp
@@ -154,6 +154,22 @@ Util::IJson* ExternConverter_clone3::convertExternFunction(
         modelError("Expected 3 arguments for %1%", mc);
[...]

PR #1688, opened 2019-01-22
p4c does not catch clone3 with incorrect clone3 third argument

hesingh:
See https://github.com/p4lang/p4c/issues/1687

The fix adds the check to the bmv2 backend code.

Unit test log is included below.  The testdata file has been moved to a negative test. 

```
hemant@ubuntu:~/p4c/build$ ./p4c-bm2-ss --std p4-16 ../testdata/p4_16_bmv_errors/issue1653-bmv2.p4
../testdata/p4_16_bmv_errors/issue1653-bmv2.p4(30): error: v1 model does not support header bvh0 inside struct for clone3 arg h
    bitvec_hdr bvh0;
               ^^^^
../testdata/p4_16_bmv_errors/issue1653-bmv2.p4(49)
        clone3(CloneType.I2E, 0, h);
                                 ^
hemant@ubuntu:~/p4c/build$ 
```


backends/bmv2/simple_switch/simpleSwitch.cpp
```diff
@@ -154,6 +154,19 @@ Util::IJson* ExternConverter_clone3::convertExternFunction(
...
+        for (auto f : strct->fields) {
+            auto ftype = f->type;
+            if (ftype->is<IR::Type_Header>()) {
+                ::error("v1 model does not support header %1% inside struct for clone3 arg %2%",
+                        f, exp);
+                return nullptr;
+            }
+        }
+    }
+
```
jafingerhut: The same restriction applies to resubmit and recirculate field lists, too.  Would it be as straightforward to add such a check for them, too?

jafingerhut: PSA does clone, recirculate, and resubmit preservation of selected metadata fields in a different way than the current implementation does, and it isn't currently working yet anyway, so nothing working there yet to update.

jafingerhut:
Hemant, I think perhaps this change is a little premature.  Sorry for any part I played in pushing something like this forward, if so.

The changes to P4_16 + v1model architecture for handling the preservation of programmer-selected metadata fields for recirculated/resubmitted/cloned packets sounds like it could be big enough of a change that it is on the schedule for discussion at the next P4 language design workgroup meeting, the first Monday of Feb 2019.

After some more details have been worked out in that forum, and implemented, it seems somewhat likely that all of the P4_16 + v1model arch programs in the testdata/p4_16_samples directory will need to be modified to conform to the new way of doing it.

It also seems likely to me that the restrictions on what fields can be named in such a field_list could become P4_14-specific, and not applicable to P4_16 programs at all.

hesingh:
Andy, no worries at all.  The plan sounds good.

PR #1693, opened 2019-01-23
Remove standard_metadata from preserved metadata in test program

jafingerhut:
It seems unlikely that standard or intrinsic metadata should be
allowed to be explicitly preserved in this way, since the architecture
defines how these should be filled in at the beginning of an ingress
and/or egress pipeline.

### DIFF
--- a/testdata/p4_16_samples/v1model-special-ops-bmv2.p4
+++ b/testdata/p4_16_samples/v1model-special-ops-bmv2.p4
@@ -218,31 +218,22 @@ control ingress(inout headers_t hdr,
         // the packet that will be processed by the ingress control
         // block in the future.
 
-        // Note: There is what might be considered a bug in p4c that
-        // if you give an individual metadata field name in the list
-        // below, and the compiler can simplify it to a constant
-        // value, e.g. because you assign that field a constant value
-        // shortly before the resubmit() call, then the BMv2 JSON file
-        // will have that constant value in it instead of the field
-        // name.  I believe in that case that BMv2 simple_switch will
+        // Note: There is a bug in p4c that in some cases can cause
+        // one or more of the fields _not_ to be preserved.  See the
+        // "Caveat emptor" section of the README.md file here:
+        // https://github.com/jafingerhut/p4-guide/tree/master/v1model-special-ops#caveat-emptor
         // _not_ have that metadata field value preserved.
 
-        // If you give an entire struct like standard_metadata, it
-        // includes all fields inside of that struct.  Even though
-        // standard_metadata.instance_type is one of the fields inside
-        // of this struct, that field's value will _not_ be preserved
-        // across the resubmit operation -- the resubmitted packet
-        // will have standard_metadata.instance_type ==
-        // BMV2_V1MODEL_INSTANCE_TYPE_RESUBMIT.  The instance_type
-        // field is an exception to the "preserve metadata field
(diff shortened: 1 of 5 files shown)


antoninbas:
I disagree with the analysis that standard_metadata cannot be preserved like this with recirculation. I see the preserved metadata as values which are serialized at the beginning the packet as it goes from the egress deparser back to the ingress parser. These values are deserialized and written back to the original fields before the user-provided parser code is executed. This makes sense for P4_14 where all fields are global identifiers, and shared between ingress and ingress.

jafingerhut:
FYI, if a test like this, with STF packet tests, had existed for a similar P4_14 program, with or without preserving standard_metadata, then this bug would have been caught before it was committed: https://github.com/p4lang/p4c/issues/1694

That bug happened, in part, because there is no such P4_14 program with STF test that could have helped catch it.

jafingerhut:
> I disagree with the analysis that standard_metadata cannot be preserved like this with recirculation. I see the preserved metadata as values which are serialized at the beginning the packet as it goes from the egress deparser back to the ingress parser. These values are deserialized and written back to the original fields before the user-provided parser code is executed, therefore overwriting the architecture values. This makes sense for P4_14 where all fields are global identifiers, and shared between ingress and ingress.


Antonin, BMv2 simple_switch lets one _specify_ that they want to preserve a field like standard_metadata.instance_type, but as implemented, this field value is always (or at least almost always) overwritten by "the architecture" (code in simple_switch.cpp) for recirculated/resubmitted/cloned packets.  This seems like reasonable "hard-coded" behavior to me that should not be modifiable via one's P4 program.

If you agree that the instance_type field should be impossible to preserve, no matter what the user's P4 program says, then perhaps you may also think there are others that should be similarly "hard-coded" in how they are filled in by the architecture, and the P4 program should not be able to change that.

It would be good to make a list of which ones you think would be reasonable to specify preserving their values, and which should not.

From v1model's current standard_metadata_t struct, here are some that seem reasonable to disallow a P4 program from preserving, since the architecture would always fill it in:
```
    bit<32> instance_type;
    bit<32> packet_length;
    bit<1>  checksum_error;
    error parser_error;
    bit<9>  egress_port;
    @alias("queueing_metadata.enq_timestamp") bit<32> enq_timestamp;
    @alias("queueing_metadata.enq_qdepth")    bit<19> enq_qdepth;
    @alias("queueing_metadata.deq_timedelta") bit<32> deq_timedelta;
    @alias("queueing_metadata.deq_qdepth")    bit<19> deq_qdepth;
    @alias("intrinsic_metadata.ingress_global_timestamp") bit<48> ingress_global_timestamp;
    @alias("intrinsic_metadata.egress_global_timestamp") bit<48> egress_global_timestamp;

    bit<32> clone_spec;
    @alias("intrinsic_metadata.lf_field_list") bit<32> lf_field_list;
    @alias("intrinsic_metadata.resubmit_flag") bit<32> resubmit_flag;
    @alias("intrinsic_metadata.recirculate_flag") bit<32> recirculate_flag;
```

and the rest below are ones that I could see someone wanting to explicitly preserve, and the architecture allow it to be preserved:
```
    bit<9>  ingress_port;
    bit<9>  egress_spec;
```

mbudiu-vmw:
I will accept this PR, it is very useful. We can have other ones for handling the standard metadata.

PR #1695, opened 2019-01-24
get rid of an unnecessary warning

zma-barefoot:
This warning is redundant because all v1model programs that have parser branches will not have an unique header order in the deparser?

### DIFF
--- a/frontends/p4/fromv1.0/programStructure.cpp
+++ b/frontends/p4/fromv1.0/programStructure.cpp
@@ -675,9 +675,7 @@ void ProgramStructure::createDeparser() {
     }
 
     std::vector<const IR::Expression*> sortedHeaders;
-    bool loop = headerOrder.sccSort(startHeader, sortedHeaders);
-    if (loop)
-        ::warning("The order of headers in deparser is not uniquely determined by parser!");
+    headerOrder.sccSort(startHeader, sortedHeaders);
 
     auto params = new IR::ParameterList;
     auto poutpath = new IR::Path(p4lib.packetOut.Id());


mihaibudiu (changes requested):
That is not true, as a proof the warning does not fire for most programs.

jafingerhut:
It sounds like Zhao reads the warning, and thinks of a parser like this:

Parse an ethernet header first always.  If etherType is 0x0800, transition to a state that parses ipv4 header next and then accepts.  If etherType is 0x86dd, transition to a state that parses ipv6 header next and then accepts.

With such a parser, both of these deparsers would be correct:
```
pkt.emit(hdr.ethernet);
pkt.emit(hdr.ipv4);
pkt.emit(hdr.ipv6);
```
or this one:
```
pkt.emit(hdr.ethernet);
pkt.emit(hdr.ipv6);
pkt.emit(hdr.ipv4);
```
One could read the error message text "The order of headers in deparser is not uniquely determined by parser!" and think of a scenario like the above, which is very very common in P4 programs.

I do not know off hand a good way to describe the scenario when that warning message is printed, but perhaps this issue is a good sign that perhaps the warning message could be changed to be more descriptive, perhaps even containing a link to a document describing why it occurs, and what you can do about it?

zma-barefoot:
@jafingerhut Yes, this is what I interpret from this warning msg. Any parse graph that has branches will have more than one valid topological sorting order, and which is true for any parse graph that supports multiple protocols. Am I missing anything?

jafingerhut:
@zma-barefoot I suspect that the condition in which the warning is printed is a sign of something potentially significant in one's parser definition, but different than the scenario I described -- it must be different than the scenario I described, because the warning is _not_ printed for P4 programs like that.

A more precise, different warning message could probably be written that you would have no objection with.

jafingerhut:
_If_ this message is accurate (I don't yet know that it is), it might be a good replacement for the one printed now:

"For the parser <parser_name>, there is no order of emitting headers in the deparser that is consisent with all paths through the parser.  That is, there is some execution path through the parser where header A can be extracted, followed later by header B, and another execution path where header B is extracted, followed by header A."

mbudiu-vmw:
The explanations of @ChrisDodd and @jafingerhut are correct.

jafingerhut:
@zma-barefoot Would you be willing to update the PR so that the compiler still prints a warning message, but it modifies the warning message to something that is clearer?

zma-barefoot:
@jafingerhut Sure, I can change the message.

But a couple of more thoughts, just to clarify:

1. If the compiler cannot uniquely determine header order in de-parser and chooses one for the user, that means user headers may get reordered when parser execute certain path. Should we just reject such programs?

2. It seems the issue can be circumvented by compiler generated header validity bits?

Consider the example above, and suppose we designate two bits `parse_ab.$hit` and `parse_ba.$hit`, and have the following order in the de-parser, then the de-parse order should be correct at runtime?

```
emit(a) if parse_ab.$hit
emit(b) if parse_ab.$hit
emit(b) if parse_ba.$hit
emit(a) if parse_ba.$hit

```

ChrisDodd:
In fact you only need one (extra) bit (in addition to the pov bits) and you can do:

    if (ab_path) emit(a);
    emit(b);
    if (!ab_path) emit(a);

This depends on support for conditions in the deparser (in addition to the implicit condition of the valid bit).

mbudiu-vmw:
The bmv2 deparser does not support conditionals.
Moreover, there may be exponentially many combinations to consider in general. In general one has to record the history of the parse states to do this dynamically.

zma-barefoot:
I'm in favor of rejecting such programs then. Otherwise, the warning should read `warning: The order of headers in deparser is not uniquely determined by parser! Headers may get re-ordered at deparser based on parser runtime execution!`, by which the compiler is really doing user a disservice.


I'm also seeing the warning get tripped on header stack, which I think is a bug? @mbudiu-vmw  

```
header_type stack_t {
    fields {
        f : 16;
    }
}

header stack_t stack[4];

parser start {
    extract(stack[next]);
    return select(latest.f) {
        0xffff : start;
        default: ingress;
    }
}

control ingress { }
```

jafingerhut:
That P4_14 program does cause that warning with a recent build of p4c, and does look like a bug of some kind.  Could perhaps file a separate issue for that problem?

mbudiu-vmw:
We cannot reject such programs; we have several p4-14 test examples that generate this warning, and we have to continue to compile them.

mbudiu-vmw:
I have filed #1696

zma-barefoot:
Thanks, @mbudiu-vmw. I'll let you do the message change as part of #1696 then. One more suggestion to improve the message, in addition to the ones above, is to print the headers that do not have total order in de-parser, something like:

```
The order of header "a", "b" in deparser is not uniquely determined by parser! The compiler chooses the order below:
    emit(a)
    emit(b)
```

Thanks all for the discussion. Closing this PR.

mkeda:
@mbudiu-vmw , I do not agree that p4-14 test examples are sacrosanct, such that they cannot be modified if we want the compiler to do more stringent error checking.  

It's hard to know about user intent, but I have certainly never written a P4-14 parser where this was desired behavior.

PR #1700, opened 2019-01-26
allow annotation on parser select case.

hanw:


mbudiu-vmw:
I find this a bit dangerous. The compiler plays a lot of select expressions moving things around; it won't know what to do with these annotations.

mbudiu-vmw:
These annotations may not be preserved by code transformations which manipulate parser states. No pass was written with the assumptions that select cases may have annotations; all that code should be audited. We should say what needs to happen to the annotations when select expressions are being transformed.

hanw:
Close this PR as I have decided to rewrite the P4 program.

PR #1708, opened 2019-02-04
Handle complex headers (with nested structs) in P4Info generation

antoninbas:
The P4Info generation code now handles complex headers by flattening
them. This is independent from the FlattenHeaders pass, which is a
midend pass. This code is much simpler and has no dependency on non
frontend pass. It simply performs a "local" flattening of a given
Type_Header, without updating any program references. It also preserves
the fully-qualified name using a @name annotation (unlike the
FlattenHeaders pass), along with annotations of the nested
fields. Because of the local nature of this flattening, it seemed silly
to write an actual pass (I tried and it was clumsy, with a call to
visitAgain).

The flattening is performed wherever the p4RuntimeSerializer has to deal
with header types, so only 2 places: for controller headers and in the
typeSpecConverter.

### DIFF
--- a/backends/p4test/CMakeLists.txt
+++ b/backends/p4test/CMakeLists.txt
@@ -66,9 +66,6 @@ set (P4RUNTIME_EXCLUDE
   # error not supported in static entries, not strictly required for P4Info
   # generation though
   testdata/p4_16_samples/issue1062-bmv2.p4
-  # issue #1670: no support for headers containing structs
-  testdata/p4_16_samples/issue1670-bmv2.p4
-  testdata/p4_16_samples/p4rt_digest_complex.p4
 )
 # Do not generate P4Info for tests which are expected to fail P4 compilation.
 p4c_find_test_names ("${P4C_SOURCE_DIR}/testdata/p4_16_errors/*.p4" P4RUNTIME_EXCLUDE_ERRORS)
--- a/control-plane/CMakeLists.txt
+++ b/control-plane/CMakeLists.txt
@@ -71,6 +71,7 @@ add_custom_command(OUTPUT ${P4RUNTIME_GEN_SRCS} ${P4RUNTIME_GEN_HDRS}
 #PROTOBUF_GENERATE_PYTHON (P4RUNTIME_GEN_PYTHON P4RUNTIME_INFO_GEN_HDRS ${P4RUNTIME_INFO_PROTO})
 
 set (CONTROLPLANE_SRCS
+  flattenHeader.cpp
   p4RuntimeArchHandler.cpp
   p4RuntimeArchStandard.cpp
   p4RuntimeSerializer.cpp
@@ -82,6 +83,7 @@ set (CONTROLPLANE_SOURCES
   )
 
 set (CONTROLPLANE_HDRS
+  flattenHeader.h
   p4RuntimeArchHandler.h
   p4RuntimeArchStandard.h
   p4RuntimeSerializer.h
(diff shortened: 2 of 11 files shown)


hesingh:
It would be good to test the changes using the P4 program included below.  It's Han's serializer2.p4.

```
#include <core.p4>
#include <v1model.p4>

#define ETHERTYPE_IPV4 0x0800
#define IP_PROTOCOLS_TCP 6
#define IP_PROTOCOLS_UDP 17

@flexible
struct switch_metadata_t {
    bit<9> port;
    bit<10> session_id;
    bit<1> check;
    bit<4> pad;
}

@flexible
struct bridge_metadata_t {
    bit<9> port;
    bit<1> check;
    bit<6> pad;
}

typedef bit<16> switch_nexthop_t;

header serialized_bridge_metadata_t {
    bit<8> type;
    bridge_metadata_t meta;
}

header serialized_switch_metadata_t {
    bit<8> type;
    switch_metadata_t meta;
}

struct parsed_packet_t {
    serialized_switch_metadata_t mirrored_md;
}

struct local_metadata_t {
    serialized_bridge_metadata_t bridged_md;
}

parser parse(packet_in pk, out parsed_packet_t h, inout local_metadata_t local_metadata,
    state start {
	transition accept;
    }
}

control ingress(inout parsed_packet_t h, inout local_metadata_t local_metadata,
    apply {
	h.mirrored_md.setValid();
	h.mirrored_md.type = 8w0;
	h.mirrored_md.meta = { 9w0, 10w0, 1w0, 4w0 };
        clone3(CloneType.I2E, 0, h.mirrored_md);
    }
}

control egress(inout parsed_packet_t hdr, inout local_metadata_t local_metadata,
    apply { }
}

control deparser(packet_out b, in parsed_packet_t h) {
    apply { }
}

control verify_checksum(inout parsed_packet_t hdr,
inout local_metadata_t local_metadata) {
    apply { }
}

control compute_checksum(inout parsed_packet_t hdr,
                         inout local_metadata_t local_metadata) {
    apply { }
}

V1Switch(parse(), verify_checksum(), ingress(), egress(),
compute_checksum(), deparser()) main;
```

hesingh:
Why not also change typeSpecConverter.cpp for ```preorder(const IR::Type_Struct* type)```.  A digest could contain a nested struct.

antoninbas:
@hesingh I tried this program and it works fine. Note that this is a poor example, because this programs doesn't expose any programmable objects to the control-plane and therefore an empty P4Info is generated.

hesingh:
> @hesingh I tried this program and it works fine. Note that this is a poor example, because this programs doesn't expose any programmable objects to the control-plane and therefore an empty P4Info is generated.

ok, so good.

antoninbas:
> Why not also change typeSpecConverter.cpp for `preorder(const IR::Type_Struct* type)`. A digest could contain a nested struct.

AFAIK nested structs are already supported and no changes are required. If this is not the case, please provide an example P4 program for which P4Info generation fails.

hesingh:
> > Why not also change typeSpecConverter.cpp for `preorder(const IR::Type_Struct* type)`. A digest could contain a nested struct.
> 
> AFAIK nested structs are already supported and no changes are required. If this is not the case, please provide an example P4 program for which P4Info generation fails.

If they are, then never mind.  There is p4rt_digest_complex.p4 in p4c/testdata/p4_16_samples/.   You can change its 'h_t' from header to struct and test.  'h_t' is used by the digest_t struct.

PR #1709, opened 2019-02-05
Add XFAIL P4_14 test program that fails because of issue #1669

jafingerhut:
It is nearly identical to the test program p414-special-ops.p4
proposed to be added in the PR #1704, except that program explicitly
lists individual fields to be preserved during resubmit
etc. operations, whereas this one gives the name of an entire metadata
'header' to be preserved.

For p414-special-ops-2-bmv2.p4, p4c as of this writing introduces
temporary metadata variables and preserves those instead of the ones
named in the source file, and since the STF packet tests cases rely
for their correct operation on the metadata fields being preserved,
they fail.


jafingerhut:
Done.

PR #1712, opened 2019-02-06
Additional slice-related strength reductions

ChrisDodd:
- slice on slice
- slice on cast

### DIFF
--- a/frontends/p4/strengthReduction.cpp
+++ b/frontends/p4/strengthReduction.cpp
@@ -318,6 +318,12 @@ const IR::Node* DoStrengthReduction::postorder(IR::Slice* expr) {
             return new IR::Concat(expr->type, expr,
                     new IR::Constant(IR::Type_Bits::get(-(lo + shift_amt)), 0)); } }
 
+    while (auto cast = expr->e0->to<IR::Cast>()) {
+        if (size_t(expr->getH()) < cast->expr->type->width_bits()) {
+            expr->e0 = cast->expr;
+        } else {
+            break; } }
+
     while (auto cat = expr->e0->to<IR::Concat>()) {
         unsigned rwidth = cat->right->type->width_bits();
         if (expr->getL() >= rwidth) {
@@ -335,11 +341,11 @@ const IR::Node* DoStrengthReduction::postorder(IR::Slice* expr) {
 
     // out-of-bound error has been caught in type checking
     if (auto sl = expr->e0->to<IR::Slice>()) {
-        auto e = sl->e0;
-        auto hi = expr->getH() + sl->getL();
-        auto lo = expr->getL() + sl->getL();
-        return new IR::Slice(e, hi, lo);
-    }
+        int delta = sl->getL();
+        expr->e0 = sl->e0;
+        if (delta != 0) {
+            expr->e1 = new IR::Constant(expr->getH() + delta);
+            expr->e2 = new IR::Constant(expr->getL() + delta); } }
 
     auto slice_width = expr->getH() - expr->getL() + 1;
     if (slice_width == (unsigned)expr->e0->type->width_bits())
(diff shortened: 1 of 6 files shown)


frontends/p4/strengthReduction.cpp
```diff
@@ -288,6 +288,12 @@ const IR::Node* DoStrengthReduction::postorder(IR::Slice* expr) {
             return new IR::Concat(expr->type, expr,
                     new IR::Constant(IR::Type_Bits::get(-(lo + shift_amt)), 0)); } }
 
+    while (auto cast = expr->e0->to<IR::Cast>()) {
+        if (size_t(expr->getH()) < cast->expr->type->width_bits()) {
```
mbudiu-vmw: since you are doing casts you may perhaps want to also handle unsigned casts by converting them into either slices or concatenations

mbudiu-vmw: Why does this work for a signed cast that widens the number?

testdata/p4_16_samples_outputs/strength5-midend.p4
```diff
@@ -0,0 +1,75 @@
...
+    bit<8> x;
+    bit<8> y;
+}
+
+control ingress(inout Headers h, inout Meta m, inout standard_metadata_t sm) {
+    @name("ingress.case0") action case0() {
+        h.h.c = h.h.a[11:4];
+    }
+    @name("ingress.case1") action case1() {
+        h.h.c = (bit<8>)h.h.a[11:7];
```
mbudiu-vmw: for example, this could be done by just prepending some zeros.

testdata/p4_16_samples/strength5.p4
```diff
@@ -0,0 +1,68 @@
...
+struct Meta {
+    bit<8> x;
+    bit<8> y;
+}
+
+control ingress(inout Headers h, inout Meta m, inout standard_metadata_t sm) {
+    action case0() { h.h.c = ((bit<12>)h.h.a)[11:4]; }
+    action case1() { h.h.c = (bit<8>)(h.h.a[14:3][8:4]); }
+    action case2() { h.h.c = (bit<8>)((16w0 ++ h.h.a)[31:8][15:8]); }
+    action case3() { h.h.c = (bit<8>)((int<32>)(int<16>)h.h.a)[14:2][10:3]; }
```
mbudiu-vmw: You should test a slice that captures the sign-extended bits, e.g. ((int<32>)(int<16>)h.h.a)[22:15]

frontends/p4/strengthReduction.cpp
ChrisDodd: This dealing with the case where the slice make the cast irrelevant -- eg, if its a widening cast, but the slice only gets bits from the original (pre-cast) value, the extra widened bits are just thrown away anayways.  So the cast is irrelvant and can be removed.  The test here is 'are all the bits used by the slice present before the cast?'  If so drop the cast.

testdata/p4_16_samples/strength5.p4
ChrisDodd: I'll add that -- that's the case that this code specifically does not modify.

mbudiu-vmw:
@ChrisDodd : do you want to rebase this PR so we can merge it?

frontends/p4/strengthReduction.cpp
ChrisDodd: Concatenation may well be more expensive than casts, so converting casts to concatenations should only be done for targets that want that, and not in the target-independent front end code.

mbudiu-vmw: That's fine with me.

mbudiu-vmw:
@ChrisDodd : do you want to fix this PR so we can also merge it?

mbudiu-vmw:
@ChrisDodd : do we already have these optimizations in the compiler?
Maybe we can close this PR.

ChrisDodd:
There's one optimization left here that hasn't been picked up by other changes -- slice on cast that makes the cast unnecessary.  eg:

    bit<16> x;
    ((bit<32>)x)[5:0]     -->  x[5:0]

PR #1718, opened 2019-02-07
Support multiple p4runtime files

### LINKED ISSUE #1705: Allow specifying multiple p4info formats/files on cmd-line (enhancement)
It would be convenient to allow generation of multiple p4info files at the same time. For example, I'd like to generate both .proto and .txt files at the same time so I"m ready to use .proto in a system but also read .text in an editor.

Proposal - allow the following command-line syntax:

`p4c --p4runtime-format format1[,format2,...] --p4runtime-file file1[,file2,...] [opts] srcfile`

e.g.` p4c --p4runtime-format binary,text--p4runtime-file myprog.p4.p4info.proto,myprog.p4.p4info.txt myprog.p4`

Another approach would be allowing multiple appearances of the respective options.

mbudiu-vmw:
Fixes #1705.

I haven't deleted the old options. Together they are a bit redundant, but I don't want to break any scripts.

### DIFF
--- a/control-plane/p4RuntimeSerializer.cpp
+++ b/control-plane/p4RuntimeSerializer.cpp
@@ -1506,7 +1506,47 @@ P4RuntimeSerializer::serializeP4RuntimeIfRequired(const IR::P4Program* program,
                                                   const CompilerOptions& options) {
     // The options parser in the frontend already prints a warning if
     // '--p4runtime-entries-file' is used without '--p4runtime-file'.
-    if (options.p4RuntimeFile.isNullOrEmpty()) return;
+    std::vector<cstring> files;
+    std::vector<P4::P4RuntimeFormat> formats;
+
+    if (!options.p4RuntimeFile.isNullOrEmpty()) {
+        files.push_back(options.p4RuntimeFile);
+        formats.push_back(options.p4RuntimeFormat);
+    }
+    if (!options.p4RuntimeFiles.isNullOrEmpty()) {
+        const char* current = options.p4RuntimeFiles.c_str();
+        while (current != nullptr) {
+            const char* comma = strchr(current, ',');
+            cstring name;
+            if (comma != nullptr) {
+                name = cstring(current, comma - current);
+                current = comma + 1;
+            } else {
+                name = cstring(current);
+                current = nullptr;
+            }
+            files.push_back(name);
+
+            const char* dot = name.find('.');
+            if (dot == nullptr) {
+                formats.push_back(options.p4RuntimeFormat);
+            } else {
+                cstring suffix = cstring(dot);
+                if (suffix == ".json") {
+                    formats.push_back(P4::P4RuntimeFormat::JSON);
+                } else if (suffix == ".p4info") {
+                    formats.push_back(P4::P4RuntimeFormat::BINARY);
+                } else if (suffix == ".txt") {
+                    formats.push_back(P4::P4RuntimeFormat::TEXT);
+                } else {
+                    ::error("%1%: Could not detect p4runtime info file format from file suffix %2%",
+                            name, suffix);
+                    return;
+                }
+            }
+        }
+    }
+    if (files.empty()) return;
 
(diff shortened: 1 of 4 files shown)


frontends/common/options.cpp
```diff
@@ -97,6 +97,11 @@ CompilerOptions::CompilerOptions() : Util::Options(defaultMessage) {
     registerOption("--toJSON", "file",
                    [this](const char* arg) { dumpJsonFile = arg; return true; },
                    "Dump the compiler IR after the midend as JSON in the specified file.");
+    registerOption("--p4runtime-files", "filelist",
```
antoninbas: you should probably update the p4c compiler driver to expose this option as well?

frontends/common/options.cpp
```diff
@@ -97,6 +97,11 @@ CompilerOptions::CompilerOptions() : Util::Options(defaultMessage) {
...
+                   [this](const char* arg) { p4RuntimeFiles = arg; return true; },
+                   "Write the P4Runtime control plane API description to the specified\n"
+                   "files (comma-separated list).  The format is inferred from the file\n"
+                   "suffix: .info, .json, .p4info");
```
antoninbas: I don't understand these suffixes apart from the .json one
The one for binary should probably be .bin (if you want to accept .p4info for binary in addition to .bin, that could be a good idea), the one for text should probably be .txt.

control-plane/p4RuntimeSerializer.cpp
```diff
@@ -1506,7 +1506,47 @@ P4RuntimeSerializer::serializeP4RuntimeIfRequired(const IR::P4Program* program,
...
+            const char* dot = name.find('.');
+            if (dot == nullptr) {
+                formats.push_back(options.p4RuntimeFormat);
+            } else {
+                cstring suffix = cstring(dot);
+                if (suffix == ".json") {
+                    formats.push_back(P4::P4RuntimeFormat::JSON);
+                } else if (suffix == ".p4info") {
+                    formats.push_back(P4::P4RuntimeFormat::BINARY);
+                } else if (suffix == ".txt") {
```
antoninbas: it seems you got the suffix right here (but not in the help message below)

control-plane/p4RuntimeSerializer.cpp
```diff
@@ -1515,12 +1555,16 @@ P4RuntimeSerializer::serializeP4RuntimeIfRequired(const IR::P4Program* program,
...
+        std::ostream* out = openFile(file, false);
+        if (!out) {
+            ::error("Couldn't open P4Runtime API file: %1%", file);
+            continue;
+        }
+        p4Runtime.serializeP4InfoTo(out, format);
     }
-    p4Runtime.serializeP4InfoTo(out, options.p4RuntimeFormat);
 
     if (!options.p4RuntimeEntriesFile.isNullOrEmpty()) {
```
antoninbas: we use `options.p4RuntimeFormat` here:
```
p4Runtime.serializeEntriesTo(out, options.p4RuntimeFormat);
```
It seems that maybe the same should be done for the entries file: add `--p4runtime-entries-files` and detect the format based one the extension?

control-plane/p4RuntimeSerializer.cpp
```diff
@@ -1506,7 +1506,47 @@ P4RuntimeSerializer::serializeP4RuntimeIfRequired(const IR::P4Program* program,
...
+                name = cstring(current);
+                current = nullptr;
+            }
+            files.push_back(name);
+
+            const char* dot = name.find('.');
+            if (dot == nullptr) {
+                formats.push_back(options.p4RuntimeFormat);
+            } else {
+                cstring suffix = cstring(dot);
```
ChrisDodd: The above 5 lines could be more concise as

    if (cstring suffix = name.find('.')) {

If there's no suffix, shouldn't that be an error (like an unrecognized suffix?)

chrispsommers:
Thanks Mihai!

Issue #1717, opened 2019-02-07
Develop sizeof solution until module system is available in p4c


hesingh:
It is better to file this new issue so that implementation discussions can take place here.  This way, the spec changes don't get cluttered with implementation details.

See spec-changes that are discussing the problem.

https://github.com/p4lang/p4-spec/issues/660
https://github.com/p4lang/p4-spec/pull/712

Right now, the implementation is focused on supporting ```type.size()```.  After a LDWG discussion, we may change the implementation.

hesingh:
p4c has been changed to support ```type.size``` in a P4 program.  See a commit in this repo.

https://github.com/hesingh/p4c

What is TODO is to convert the expression to an IR::Constant.  The typeChecker.cpp code does not deal with transforming IR.  Thus, a home is needed in frontend code where the conversion code can be added. 

@mbudiu-vmw Would you know a frontend .cpp file where this TODO code could go in?

mbudiu-vmw:
Either constantFolding.cpp or typeChecker.cpp

hesingh:
> Either constantFolding.cpp or typeChecker.cpp

thanks!

hesingh:
Prototyped code for type.size() has pinged.  I will look into concatenated type.size() next and make sure that code works.  

Log is included below.
```
hemant@ubuntu:~/int/p4c/build$ ./p4test --std p4-16 ../testdata/p4_16_samples/v1model-special-ops-bmv2.p4
sizeBits: 112
sizeBytes: 14
hemant@ubuntu:~/int/p4c/build$
```
P4 code used:
```
state parse_ethernet {
        packet.extract(hdr.ethernet);
        len = hdr.ethernet.sizeBits();     <===
        len = hdr.ethernet.sizeBytes();  <===
        transition select(hdr.ethernet.etherType) {
            ETHERTYPE_IPV4: parse_ipv4;
            default: accept;
        }
}
```

hesingh:
I fixed a bug in the code to support multiple sizeof computation with the arithmetic Add operator.   This was the P4 program used to test. 

```
len = hdr.ethernet.sizeBits() + hdr.ethernet.sizeBits();
len = hdr.ethernet.sizeBytes() + hdr.ethernet.sizeBytes();
```

hesingh:
I added nested struct inside header test and testing passed.  Few tests were added and 'make check' passed for p4c.  Thus, I have issued a code review PR.

https://github.com/p4lang/p4c/pull/1720

PR #1720, opened 2019-02-08
Fix for Issue # 1717 to add sizeof to p4c using type.size()

hesingh:
See https://github.com/p4lang/p4c/issues/1717


testdata/p4_16_errors_outputs/struct_e.p4-stderr
```diff
@@ -4,9 +4,7 @@ struct_e.p4(24): [--Werror=type-error] error: Field data is not a member of stru
...
-struct_e.p4(25): [--Werror=type-error] error: Cannot extract field data from b which has type bit<1>
-        b.data = 5; // no such field
-          ^^^^
-struct_e.p4(25)
-        b.data = 5; // no such field
-        ^
+terminate called after throwing an instance of 'Util::CompilerBug'
+  what():  In file: /home/hemant/int/p4c/lib/crash.cpp:229
+[31mCompiler Bug[0m: Exiting with SIGSEGV
+
```
jafingerhut: Not sure whether it is a good idea to check in code with known crashes like this?  Not really for me to judge -- I was just skimming the changes and noticed this.

hesingh: I will look into this one in the parent branch and well is this PR repo and get back - thanks.

hesingh: I had added one line of new code (an ::error message) to typeChecker.cpp to debug my code.  I forgot to remove this line of code.  Once I removed it, the test passed.  I have checked in code changes.  thanks.

hesingh:
Code is ready to be reviewed.  

Once review is complete, I will cleanup typeChecker.cpp by removing ```#include <iostream>" and any commented out cout.  Until then, I am keep the code to debug any issues raised. 

thanks.

hesingh:
Sorry, when I sent out the prior note for code review, I had not tested header union and header stack.  Adding this support led to more p4c changes.  Please review the new round of diffs - thanks.

hesingh:
Since Mihai is away for two weeks starting today, could anyone else who has worked with typeChecker.cpp please review the code changes.

frontends/p4/typeChecking/typeChecker.cpp
```diff
@@ -2435,6 +2438,41 @@ const IR::Node* TypeInference::postorder(IR::Member* expression) {
         return expression;
     }
 
+    int sz = 0;
+    int size = 0;
```
cc10512: Both `sz` and size should go inside the if and I don't see why you need both.
You can do: 
```
auto size = ...
if (ht) size *= ht->width_bits()
if (member == R::Type_Header::sizeBytes) size = ROUNDUP(size, 8);
```

frontends/p4/typeChecking/typeChecker.cpp
```diff
@@ -2448,19 +2486,49 @@ const IR::Node* TypeInference::postorder(IR::Member* expression) {
                 return expression;
             }
         }
-        if (type->is<IR::Type_Header>()) {
-            if (member == IR::Type_Header::setValid ||
-                member == IR::Type_Header::setInvalid) {
+        if (type->is<IR::Type_Header>() || type->is<IR::Type_HeaderUnion>()) {
+            // Built-in methods
+            if ((member == IR::Type_Header::setValid ||
```
cc10512: Please add an error if setValid/setInvalid is called on a header union.
```
if (type->is<IR::Type_HeaderUnion>() && (member == IR::Type_Header::setValid || member == IR::Type_Header::setInvalid)) {
    ::error(ErrorType::ERR_INVALID, "setValid and setInvalid are not defined for header unions", expression);
    return expression;
}
```

frontends/p4/typeChecking/typeChecker.cpp
```diff
@@ -2448,19 +2486,49 @@ const IR::Node* TypeInference::postorder(IR::Member* expression) {
...
+                                                new IR::ParameterList);
                 auto ctype = canonicalize(type);
                 if (ctype == nullptr)
                     return expression;
                 setType(getOriginal(), ctype);
                 setType(expression, ctype);
                 return expression;
+            } else if ((member == IR::Type_Header::sizeBits) ||
+                       (member == IR::Type_Header::sizeBytes)) {
+                const IR::Expression* e = expression;
```
cc10512: `expression` is already an `IR::Member`, why do you need to recast it?

frontends/p4/typeChecking/typeChecker.cpp
```diff
@@ -2435,6 +2438,41 @@ const IR::Node* TypeInference::postorder(IR::Member* expression) {
         return expression;
     }
 
+    int sz = 0;
+    int size = 0;
+    if (type->is<IR::Type_Stack>()) {
+        if ((member == IR::Type_Header::sizeBits) ||
+            (member == IR::Type_Header::sizeBytes)) {
+            const IR::Expression* e = expression;
+            auto mem = e->to<IR::Member>();
```
cc10512: `expression` is already an `IR::Member`, why do you need to recast it?

frontends/p4/typeChecking/typeChecker.cpp
```diff
@@ -2835,7 +2903,11 @@ void TypeInference::checkCorelibMethods(const ExternMethod* em) const {
 
 const IR::Node* TypeInference::postorder(IR::MethodCallExpression* expression) {
     if (done()) return expression;
+    auto method = expression->to<IR::MethodCallExpression>()->method;
```
cc10512: `expression` is already an `IR::MethodCallExpression`. Please move after log message.

frontends/p4/typeChecking/typeChecker.cpp
```diff
@@ -25,6 +26,8 @@ limitations under the License.
 #include "frontends/common/resolveReferences/resolveReferences.h"
 #include "frontends/p4/methodInstance.h"
 
+using namespace std;
```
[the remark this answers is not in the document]
hesingh: Actually, in this PR, an earlier comment by me said "I will remove this temp code being used to debug issues."

frontends/p4/typeChecking/typeChecker.cpp
```diff
@@ -175,7 +178,7 @@ const IR::Type* TypeInference::getTypeType(const IR::Node* element) const {
     // See comment in getType() above.
     if (result == nullptr) {
         if (::errorCount() == 0)
-            BUG("Could not find type of %1%", element);
+            BUG("Could not find Typetype of %1%", element);
```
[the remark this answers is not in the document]
hesingh: Ok, will do.

frontends/p4/typeChecking/typeChecker.cpp
hesingh: Ok.  The variable in 'sz' is used by two if-statements.  First one is 'if Stack' and the other one is below it for 'if StructLike'.  Thus, I will use only 'sz' and declare it outside both if-statements.

cc10512: you do not use it outside the if body, so it can be local to each branch of the if.

frontends/p4/typeChecking/typeChecker.cpp
```diff
@@ -2435,6 +2438,41 @@ const IR::Node* TypeInference::postorder(IR::Member* expression) {
...
+            const IR::Expression* e = expression;
+            auto mem = e->to<IR::Member>();
+            auto t = typeMap->getType(mem->expr, true);
+            const IR::Type_StructLike* ht;
+            auto stack = type->to<IR::Type_Stack>();
+            size = type->to<IR::Type_Stack>()->getSize();
+            auto stype = stack->elementType;
+            if (stype->is<IR::Type_Header>()) {
+                ht = stype->to<IR::Type_Header>();
+                if (ht == nullptr) {
```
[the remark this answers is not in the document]
hesingh: Good catch.  Will do.

frontends/p4/typeChecking/typeChecker.cpp
hesingh: will do.

frontends/p4/typeChecking/typeChecker.cpp
hesingh: ok.

frontends/p4/typeChecking/typeChecker.cpp
hesingh: ok

frontends/p4/typeChecking/typeChecker.cpp
hesingh: Code changes from code review caused this legacy test to be updated.

testdata/p4_16_errors_outputs/issue561-1.p4-stderr

Travis has passed.  Current round of changes are ready for review.

hesingh:
Calin or anyone else who has merge access to p4c, please merge this PR - thanks.

hesingh:
2nd request: When anyone gets a chance, please merge this code or let me know if there is any issue - thanks.

hanw:
In current P4-16 specification, we can already compute the size of a header stack with 
`headerStack.size() * hdr.sizeInBits()`. Why do we need to define `headerStack.sizeInBits()`, isn't it redundant?

jafingerhut:
If that is the biggest remaining comment on the proposal, then we are pretty much done, I would say :-)

Multiplication only works portably in P4_16 if both operands are compile-time constants, though.  Are you thus content with both operands `headerStack.size()` and `hdr.sizeInBits()` being compile-time constants?

hesingh:
> In current P4-16 specification, we can already compute the size of a header stack with
> `headerStack.size() * hdr.sizeInBits()`. Why do we need to define `headerStack.sizeInBits()`, isn't it redundant?

Minor correction.  It's not 'headerStack.size()'.  It's 'headerStack.size' - no parenthesis.   

It's convenient to the user to just type one operator to get the headerStack size.  This way, all types (header, header union, header stack) are consistent for use with the new sizeBits() and sizeBytes() methods.   Both expression types make sense to support.

hesingh:
Indeed both ```headerStack.size``` and ```hdr.sizeBits()``` are compile-time constants.  For good measure, I added this new line of code to the header stack test program for this PR and testing passed.

```len = tmp.size * h.sizeBits();```

See p4c/testdata/p4_16_samples/sizeof-stack-bmv2.p4

**Comment by jafingerhut** [comment]
Sorry, I forgot that a lot of the discussion wasn't about whether they were compile time constants, but whether they were converted to constant int values early in the compiler's passes, or late.  You can saafely ignore my most recent comment, I think.

**Comment by hesingh** [comment]
> Sorry, I forgot that a lot of the discussion wasn't about whether they were compile time constants, but whether they were converted to constant int values early in the compiler's passes, or late. You can safely ignore my most recent comment, I think.

No worries.  The conversion is done in the p4c frontend typeChecker which is not a frontend pass.  The typeChecker is an early gate to processing the P4 program's types to be checked.

**Comment by hesingh** [comment]
Any other reviewer? Or are we good to merge this code today?

**Comment by hesingh** [comment]
If there is no objection, could anyone with merge permissions, please merge this PR - thanks.

File: frontends/p4/typeChecking/typeChecker.cpp [hunk 10]
```diff
@@ -2448,19 +2472,51 @@ const IR::Node* TypeInference::postorder(IR::Member* expression) {
                 return expression;
             }
         }
-        if (type->is<IR::Type_Header>()) {
-            if (member == IR::Type_Header::setValid ||
-                member == IR::Type_Header::setInvalid) {
+        if (type->is<IR::Type_Header>() || type->is<IR::Type_HeaderUnion>()) {
+            if (type->is<IR::Type_HeaderUnion>() &&
```
mbudiu-vmw: This code is not necessary. You get a fine error message without it.

frontends/p4/typeChecking/typeChecker.cpp
```diff
@@ -2448,19 +2472,51 @@ const IR::Node* TypeInference::postorder(IR::Member* expression) {
...
+                       (member == IR::Type_Header::sizeBytes)) {
+                int sz = 0;
+                auto t = typeMap->getType(expression->expr, true);
+                const IR::Type_StructLike* ht;
+                if (t->is<IR::Type_Header>()) {
+                    ht = t->to<IR::Type_Header>();
+                } else if (t->is<IR::Type_HeaderUnion>()) {
+                    ht = t->to<IR::Type_HeaderUnion>();
+                }
+                if (ht == nullptr) {
```
mbudiu-vmw: This should be a BUG not an error. It does not make any sense as an error for the user.

frontends/p4/typeChecking/typeChecker.cpp
```diff
@@ -2448,19 +2472,51 @@ const IR::Node* TypeInference::postorder(IR::Member* expression) {
...
+                    ::error("%1%: null header/headerUnion?", expression);
+                    return expression;
+                }
+                sz = ht->width_bits();
+                if (member == IR::Type_Header::sizeBytes) {
+                    sz = ((sz + 7) >> 3);
+                }
+                auto result = new IR::Constant(sz);
+                result->type = IR::Type_Bits::get(32);
+                auto ctype = canonicalize(result->type);
```
mbudiu-vmw: I don't think this is necessary.

frontends/p4/typeChecking/typeChecker.cpp
```diff
@@ -2448,19 +2472,51 @@ const IR::Node* TypeInference::postorder(IR::Member* expression) {
...
+                if (t->is<IR::Type_Header>()) {
+                    ht = t->to<IR::Type_Header>();
+                } else if (t->is<IR::Type_HeaderUnion>()) {
+                    ht = t->to<IR::Type_HeaderUnion>();
+                }
+                if (ht == nullptr) {
+                    ::error("%1%: null header/headerUnion?", expression);
+                    return expression;
+                }
+                sz = ht->width_bits();
```
mbudiu-vmw: this won't work for Type_HeaderUnion. I think that the function width_bits() for Type_HeaderUnion is in fact buggy, since it calls width_bits on types which are all Type_Name.

frontends/p4/typeChecking/typeChecker.cpp
```diff
@@ -2836,6 +2892,10 @@ void TypeInference::checkCorelibMethods(const ExternMethod* em) const {
 const IR::Node* TypeInference::postorder(IR::MethodCallExpression* expression) {
     if (done()) return expression;
     LOG2("Solving method call " << dbp(expression));
+    auto method = expression->method;
```
mbudiu-vmw: This is clearly wrong. The type of a method call is not the method type, but the result of the method type.

testdata/p4_16_errors_outputs/issue561-1.p4-stderr
```diff
@@ -7,9 +7,6 @@ issue561-1.p4(29)
 issue561-1.p4(19)
 header_union U {
              ^
-issue561-1.p4(30): [--Werror=type-error] error: Field setValid is not a member of structure header_union U
+issue561-1.p4(30): [--Werror=invalid] error: u.setValid: Invalid setValid and setInvalid are not defined for header unions
```
mbudiu-vmw: why change this error message?

testdata/p4_16_samples/sizeof-bmv2.p4
```diff
@@ -0,0 +1,115 @@
...
+
+parser parserI(packet_in pkt,
+               out Parsed_packet hdr,
+               inout mystruct1 meta,
+               inout standard_metadata_t stdmeta) {
+    bit<32> len = 0;
+    state start {
+        bit<8> my_local = 1;
+        pkt.extract(hdr.ethernet);
+        len = hdr.ethernet.sizeBits() + hdr.ethernet.sizeBits();
```
mbudiu-vmw: this is dead code. You should use the result for something.
Also, we claim that sizeBytes is a constant, so you should try to assign it to a `const bit<32>` value to see if it really works.

frontends/p4/typeChecking/typeChecker.cpp
hesingh: Another reviewer asked to add the new error.  It makes sense because now HeaderUnion is also allowed (line 2473) and one should check for setValid() and setInvalid() not supported by HeaderUnion

frontends/p4/typeChecking/typeChecker.cpp
hesingh: I don't see a Type_Name.  Please see this code.

https://github.com/p4lang/p4c/blob/master/ir/type.def#L258

frontends/p4/typeChecking/typeChecker.cpp
hesingh: In the Member postorder method where sz is computed, the result type is set to Type_Bits.  Subsequently the MethodCallExpression is run.  I think it is semantics because once the type was set, it is checked for the same type here and then method is returned.  What would be a code change you would suggest?

testdata/p4_16_errors_outputs/issue561-1.p4-stderr
hesingh: It is due to the new error message added in the Member postoder method.

testdata/p4_16_samples/sizeof-bmv2.p4
hesingh: The `len` is computed correctly as verified by a debugged message added to typeChecker.cpp.  I can certainly add code to use the result.

frontends/p4/typeChecking/typeChecker.cpp
hesingh: It is needed.  Without the code we get a Compiler bug.
```
hemant@ubuntu:~/int/p4c/build$ ./p4test --std p4-16 ../testdata/p4_16_samples/sizeof-bmv2.p4
SZ: 112
SZ: 112
In file: /home/hemant/int/p4c/frontends/p4/typeChecking/typeChecker.cpp:169
Compiler Bug: Could not find type of 112

hemant@ubuntu:~/int/p4c/build$ 
```

ChrisDodd: `canonicalize` on a Type_Bits type is a noop (always returns the argument), so the call to `canonicalize is not needed.  In addition, the type could/should be set in the ctor:

    auto result = new IR::Constant(IR::Type::Bits::get(32), sz);
    setType(result, result->type);

ChrisDodd: Though arguably, the result of one of these `size` operations should be an untyped constant (a Type_InfInt), not a bit<32>

hesingh: Thanks, Chris.   Hmm, when would a header size be greater than `2^32`?  If we go with TypeIInfInt, I will have to see what to use replace `get(32)` with.

ChrisDodd: It might be much smaller than 2^32 -- it would be reasonable to store the result of a .size in eg a bit<8> field.  It should be possible to use just `result = IR::Constant(sz);` giving a constant of unspecified size and have the typechecker infer the size from the context.

mbudiu-vmw:
I actually think it would be better to do this in the constant folding pass. Type inference should just return type_infint for these function results.

hesingh:
> You should make the tests much simpler; we don't need to use v1model to test this feature.
> Also, you should test on a header that has a field already called sizeInBits. I suspect that won't really work.

Most of the controls are empty and the parser is being used with two headers.  I think, we can stick to using v1model.p4.  I did add an assignment of computed sizeinBits() to a `bit<32>` and testing passed.

hesingh:
> I actually think it would be better to do this in the constant folding pass. Type inference should just return type_infint for these function results.

To be clear, are you saying, move all the code changes from typeChecker.cpp to the constant folding pass?  The code may not port as is, but that is the gist of it?

frontends/p4/typeChecking/typeChecker.cpp
hesingh: Soon as I use InfInt this code does not even compile.  My impression is that an InfInt is written in a P4 program as `int`.

```
struct Parsed_packet {
    Ethernet_h    ethernet;
    ipv4_t        ipv4;
    int           length;
}
```
```
hemant@ubuntu:~/int/p4c/build$ ./p4test --std p4-16 ../testdata/p4_16_samples/sizeof-bmv2.p4
../testdata/p4_16_samples/sizeof-bmv2.p4(51):syntax error, unexpected IDENTIFIER, expecting <
    int length
        ^^^^^^
hemant@ubuntu:~/int/p4c/build$ 
```

jafingerhut: It is expected that the compiler should give an error for `P4_16` code like that.  Type `int` is not allowed in headers.

It is allowed to assign a value of type `int` to an L-value of type `bit<W>` or `int<W>` in P4_16 programs, with a warning if the compile-time constant value of the `int` is too wide for the L-value being assigned to.  p4c already implements all of this, I believe.

I don't think anyone is suggesting that we should change P4_16 so that headers can contain values of type `int`.  I believe the suggested change is that `sizeBits()` should return a value of type `int`.

hesingh: Ah, I got it about int not allowed in headers.  I will what to do towards assigning the computed length to another const type.  thanks.

hesingh:
The typeChecker is able to infer the size when the typeChecker.cpp is changed as shown below.

```
 auto result = new IR::Constant(sz);
 setType(result, result->type);

...
if (method->type->is<IR::Type_InfInt>()) {
        return method;
}
```
The test P4 line of code that works is:

`hdr.ipv4.totalLen = hdr.ethernet.sizeBits();`

However, use of InfInt fails to infer binary arithmetic when the P4 program is changed as shown below. 

`hdr.ipv4.totalLen = hdr.ethernet.sizeBits() + hdr.ethernet.sizeBits();`

We get the following error
```
./p4test --std p4-16 ../testdata/p4_16_samples/sizeof-bmv2.p4 

In file: /home/hemant/int/p4c/frontends/p4/typeChecking/typeChecker.cpp:148
```

I debugged the node when the above error happens.  The node is `Add`.  

Moving the feature to constantFolding pass does not avoid the above problem.   

Other changes to P4 test program tested fine. I checked in a new sizeof-bmv2.p4

1. Regarding the sizeof-bmv2.p4 test, I removed using v1model.p4 and trimmed down the test.   After p4c code changes are finalized, I will trim down other test programs. 

2. I also added `bit<8>  sizeInBits;` to the ipv4_t header and used this line of code to test:

`  hdr.ipv4.sizeInBits = hdr.ethernet.sizeBits();`

3. Now the computation results is also assigned to another variable instead of dead code.

mbudiu-vmw:
Probably moving it to the constant folding pass would be more difficult, so let's keep it in the typechecker pass. Why don't you let me try to take a stab at it?

hesingh:
Sure, please go ahead.  thanks.

mbudiu-vmw:
I will close this since it is superseded by #1751

PR #1722, opened 2019-02-08
implementation for serializable struct in the frontend

hanw:
In case I convinced anyone that having a narrower sense of nested struct is a good idea, here is the implementation to support serializable struct in the frontend.


mihaibudiu (commented):
I don't understand why we need a serializable struct.
We can have a struct be serializable if all its fields are serializable. And we define some base types to be serializable.
We should take this to the LDWG.

mihaibudiu (commented):
What is the point of this new type?
Why can't we say that a struct is serializable if all the fields are serializable?
(We should first discuss what serializable means, anyway)

hanw:
Superseded by a PR to spec via https://github.com/p4lang/p4-spec/pull/736

PR #1723, opened 2019-02-08
Propagate --p4runtime-files from p4c driver to compiler

antoninbas:


antoninbas:
> Are we going to deprecate the other forms?

I think so, I'll add a warning

antoninbas:
@chrispsommers FYI

chrispsommers:
@antoninbas thanks!

PR #1725, opened 2019-02-10
 add expressions to several type checking warnings

cc10512:
Also extend the --Wdisable, --Wwarn, --Werror to the error catalog, as it was used only for pragmas.


cc10512:
Indeed the strength reduction is not handling this correctly and this fix is not addressing that issue, only the fact that we're not printing the messages correctly. I'll keep the issue open for the actual fix.

PR #1730, opened 2019-02-15
Update PSA include file psa.p4 to PSA version 1.1, plus psa_switch sp…

jafingerhut:
…ecifics

I have proposed some particular bit widths for the psa_switch back
end, basing their proposed sizes on BMv2 simple_switch v1model's
current sizes in bits for similar fields, in order to keep
simple_switch and psa_switch from different in those kinds of details.

### DIFF
--- a/p4include/psa.p4
+++ b/p4include/psa.p4
@@ -29,68 +29,276 @@ limitations under the License.
  * These types need to be defined before including the architecture file
  * and the macro protecting them should be defined.
  */
-#define PSA_CORE_TYPES
-#ifdef PSA_CORE_TYPES
+#define PSA_ON_BMV2_CORE_TYPES
+#ifdef PSA_ON_BMV2_CORE_TYPES
+/* The bit widths shown below are specifc to the BMv2 psa_switch
+ * target.  These types do _not_ dictate what sizes these types should
+ * have for any other PSA implementation of PSA.
+ *
+ * In many cases, the bit widths of these types have been chosen to be
+ * the same as corresponding types in the v1model architecture, but
+ * this is simply because it should help keep the differences between
+ * BMV2 simple_switch and BMv2 psa_switch a tiny bit smaller. */
+
+/* These are defined using `typedef`, not `type`, so they are truly
+ * just different names for the type bit<W> for the particular width W
+ * shown.  Unlike the `type` definitions below, values declared with
+ * the `typedef` type names can be freely mingled in expressions, just
+ * as any value declared with type bit<W> can.  Values declared with
+ * one of the `type` names below _cannot_ be so freely mingled, unless
+ * you first cast them to the corresponding `typedef` type.  While
+ * that may be inconvenient when you need to do arithmetic on such
+ * values, it is the price to pay for having all occurrences of values
+ * of the `type` types marked as such in the automatically generated
+ * control plane API.
+ *
+ * Note that the width of typedef <name>Uint_t will always be the same
+ * as the width of type <name>_t. */
+
+/* The comments after each line give the name of a similar field in
+ * the BMv2 simple_switch implementation, on which the width of this
+ * PSA type is based. */
+typedef bit<9>  PortIdUint_t;          // ingress_port, egress_port
+typedef bit<16> MulticastGroupUint_t;  // mcast_grp
+typedef bit<16> CloneSessionIdUint_t;  // clone_spec, but see below
+typedef bit<3>  ClassOfServiceUint_t;  // priority in p4c PR #1704
+typedef bit<32> PacketLengthUint_t;    // packet_length
+typedef bit<16> EgressInstanceUint_t;  // egress_rid
+typedef bit<48> TimestampUint_t;       // ingress_global_timestamp, egress_global_timestamp
+
+/* Note: clone_spec in BMv2 simple_switch v1model is 32 bits wide, but
+ * it is used such that 16 of its bits contain a clone/mirror session
+ * id, and 16 bits contain the numeric id of a field_list.  Only the
+ * 16 bits of clone/mirror session id are comparable to the type
+ * CloneSessionIdUint_t here.  See occurrences of clone_spec in this
+ * file for details:
+ * https://github.com/p4lang/behavioral-model/blob/master/targets/simple_switch/simple_switch.cpp
+ */
+
+@p4runtime_translation("p4.org/psa/v1/PortId_t", 32)
+type PortIdUint_t         PortId_t;
+type MulticastGroupUint_t MulticastGroup_t;
+type CloneSessionIdUint_t CloneSessionId_t;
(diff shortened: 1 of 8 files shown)


p4include/psa.p4
```diff
@@ -29,69 +29,277 @@ limitations under the License.
...
+    return (PortId_t) (PortIdUint_t) (PortIdInHeaderUint_t) x;
+}
+MulticastGroup_t psa_MulticastGroup_header_to_int (in MulticastGroupInHeader_t x) {
+    return (MulticastGroup_t) (MulticastGroupUint_t) (MulticastGroupInHeaderUint_t) x;
+}
+CloneSessionId_t psa_CloneSessionId_header_to_int (in CloneSessionIdInHeader_t x) {
+    return (CloneSessionId_t) (CloneSessionIdUint_t) (CloneSessionIdInHeaderUint_t) x;
+}
+ClassOfService_t psa_ClassOfService_header_to_int (in ClassOfServiceInHeader_t x) {
+    return (ClassOfService_t) (ClassOfServiceUint_t) (ClassOfServiceInHeaderUint_t) x;
```
hanw: If I am not mistaken, the input to this function is of type `ClassOfServiceInHeader_t` which is 8 bits, and the output from this function is of type `ClassOfService_t` which is also 8 bits. Can you show a use case of these functions?

jafingerhut: What you say is true if a PSA implementation chooses to define ClassOfService_t to be 8 bits wide, but they can choose any smaller width for their PSA implementation, too, e.g. 3 bits.

p4include/psa.p4
```diff
@@ -29,69 +29,277 @@ limitations under the License.
...
 
-const   CloneSessionId_t PSA_CLONE_SESSION_TO_CPU = 0;
-#endif  // PSA_CORE_TYPES
-#ifndef PSA_CORE_TYPES
-#error "Please define the following types for PSA and the PSA_CORE_TYPES macro"
+const CloneSessionId_t PSA_CLONE_SESSION_TO_CPU = (CloneSessionId_t) 0;
+
+#endif  // PSA_EXAMPLE_CORE_TYPES
+
+#ifdef 0
```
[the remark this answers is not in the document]
jafingerhut: The part of the file between the line "// BEGIN:Type_defns" and "// END:Type_defns" is auto-included in the published PSA specification.

If that code is not part of the output of the C preprocessor, in what way has it annoyed you to deal with the `bit<unspecified>`?  In your text editor somehow?

p4include/psa.p4
jafingerhut: There are example uses of one of these conversion functions in this example program: https://github.com/p4lang/p4-spec/blob/master/p4-16/psa/examples/psa-example-parser-error-handling.p4#L380

jafingerhut:
Looks like I probably need to update some of the example PSA programs in the p4c test suite, to match what they have been updated to in the p4-spec repo.  Will add another commit doing that soon.

p4include/psa.p4
```diff
@@ -29,69 +29,233 @@ limitations under the License.
...
+ * of the `type` types marked as such in the automatically generated
+ * control plane API.
+ *
+ * Note that the width of typedef <name>Uint_t will always be the same
+ * as the width of type <name>_t. */
+
+/* The comments after each line give the name of a similar field in
+ * the BMv2 simple_switch implementation, on which the width of this
+ * PSA type is based. */
+typedef bit<9>  PortIdUint_t;          // ingress_port, egress_port
```
antoninbas: I wonder that for the sake of simplicity we should try to have these types match the P4Runtime ones for bmv2 PSA.
For example, when creating a bmv2 instance, one can choose arbitrary port numbers and there is no reason for the user not to choose the SDN port numbers (see P4Runtime spec). So we could just use `bit<32>` for ports and say that for bmv2 port translation is never required. This way the `PSA_PORT_CPU` can just match the P4Runtime one: 0xfffffffd (https://github.com/p4lang/p4runtime/blob/master/proto/p4/v1/p4runtime.proto#L696).

@jafingerhut don't you think it's something worth considering? Otherwise, if we go with a `PSA_PORT_CPU` which is different from the P4Runtime one for example, then the P4Runtime server implementation for bmv2 PSA is required to perform translation.

I think there is a lot of value in achieving consistency in this. We are less likely to confuse people :)

The only drawback potentially is that it may be harder to test port translation code on bmv2, since no translation is actually needed but I doubt it is an issue. One can always use an identity mapping...

BTW I took ports as an example, but this applies to the other types which appear in P4Runtime. I think the goal of consistency with P4Runtime should be more important than consistency with v1model.

jafingerhut: Sounds like a reasonable idea to me.  I will soon update the PR based on this suggestion...

jafingerhut: So the BMv2 psa_switch would need to handle 32-bit port numbers in the data path code?  That could make for some interesting psa_switch command lines, e.g.:
```
psa_switch --log-console -i 939280@veth2 -i 1004902@veth4
```

antoninbas: bmv2 already supports arbitrary 32-bit port numbers, otherwise I would not have suggested it
I expect that "front-panel" ports will always be small integers. The CPU port & recirculate port are large integers but will be hardcoded in the code.
If someone uses some unusual port numbers, it is likely that they are simulating large topologies of bmv2 switches and orchestrating it through scripts.
It's much more likely to be `psa_switch --log-console -i 1@veth2 -i 2@veth4`

jafingerhut: PortId_t is 32 bits in P4Runtime, and ClassOfService_t is 8 bits I think.

Are there P4Runtime sizes specified for the other 5 types, that at least so far do not have a requirement for runtime numeric translation?

+ MulticastGroup
+ CloneSessionId
+ PacketLength
+ EgressInstance
+ Timestamp

antoninbas: I would recommend to just use the width of the `InHeader` variant of the type for these. P4Runtime uses 32-bit number for all of them but for non-translated types, the value must fit in the data plane type and the `InHeader` variant is supposed to be an upper bound for all targets.

jafingerhut:
I changed 4 PSA test programs to XFAIL on p4test.  My best guess is that they are now failing because PortId_t changed from `typedef bit<W> PortId_t` to `type bit<W> PortId_t`.  At least, that is my best guess for the reason why these tests are now failing.

jafingerhut:
@hanw OK to commit this PR as is, with the 4 tests newly marked XFAIL so that the remaining automated tests pass?  We can debug those 4 newly marked XFAIL tests after this commit.

hanw:
yes. please go ahead.

hesingh:
I found the root cause and have fixed the issue with p4c code changes.  I will issue a PR soon and have filed a new Issue.

https://github.com/p4lang/p4c/issues/1734

PR #1731, opened 2019-02-16
Update P4Runtime version to v1.0.0

antoninbas:
The only change that was required was to Value Set serialization. Value
Set support was improved between P4Runtime v1.0.0-rc4 and v1.0.0 and
is now consistent with the P4_16 spec. In particular, we support
mutliple match fields with potentially different match kinds.


antoninbas:
I also added support for the `@match` annotation to the `ParseAnnotations` run in the frontend. I don't know why it was omitted earlier since the annotation is part of the spec.

control-plane/p4RuntimeSerializer.cpp
```diff
@@ -604,6 +606,38 @@ getActionRefs(const IR::P4Table* table, ReferenceMap* refMap) {
...
+        return MatchField::MatchTypes::LPM;
+    } else if (matchTypeName == P4CoreLibrary::instance.ternaryMatch.name) {
+        return MatchField::MatchTypes::TERNARY;
+    } else if (matchTypeName == P4V1::V1Model::instance.rangeMatchType.name) {
+        return MatchField::MatchTypes::RANGE;
+    } else if (matchTypeName == P4V1::V1Model::instance.selectorMatchType.name) {
+        // Nothing to do here, we cannot even perform some sanity-checking.
+        return boost::none;
+    } else {
+        return MatchField::MatchTypes::UNSPECIFIED;
```
cc10512: Do you really need UNSPECIFIED? The spec says: "If the @match annotation is not specified on a struct field, by default it is assumed to be @match(exact)." Shouldn't then just return EXACT?

control-plane/p4RuntimeSerializer.cpp
```diff
@@ -929,12 +943,79 @@ class P4RuntimeAnalyzer {
...
-        vs->set_bitwidth(bitwidth);
         vs->set_size(size);
+
+        /// Look for a @match annotation on the struct field and set the match
+        /// type of the match field appropriately.
+        auto setMatchType = [this](const IR::StructField* sf, p4configv1::MatchField* match) {
+            auto matchAnnotation = sf->getAnnotation(IR::Annotation::matchAnnotation);
+            // default is EXACT
+            if (!matchAnnotation) {
+                match->set_match_type(MatchField::MatchTypes::EXACT);  // default match type
```
cc10512: Oh, I see, you set the default here. Nevermind!

PR #1735, opened 2019-02-17
support slice expression in bmv2 backend

hanw:


jafingerhut:
I have not checked, but it seems likely that the reason the BMv2 JSON file produced from psa-example-hello-world.p4, with these proposed changes in p4c, is because the assignments to the PSA-specific standard metadata is considered 'dead code' that is optimized out, because nothing later 'reads' the results of the assignments.  In a real PSA device, of course, the architecture would.

hanw:
@mbudiu-vmw you are right. I see that https://github.com/p4lang/p4c/blob/master/backends/bmv2/psa_switch/psaSwitch.cpp#L431 has a few passes commented out (probably by me a while ago). I will try to re-enable them.

PR #1737, opened 2019-02-18
add MoveActionsToTables pass to psa midend

hanw:


hanw:
with this change, helloworld.p4 now generates correct table calls in ingress.

PR #1748, opened 2019-02-26
Add varbits last field check

hemant-avatar3:
To make sure, length of varbits is never before the varbits field in a header, this PR adds a check to p4c typeChecker.  The check ensures varbits is the last field.


hanw:
I vaguely remember we had a requirement for varbit to be at the end of a header. However, I am not able to find the corresponding text in P4-16 spec. Do you know which section specifies that?

hesingh:
I don't see any text in the latest P4-16 spec that specifies the fact.  Network specification for IPv6 specifies TLV format and some IPv4 options too.  Thus the L(length) cannot be after V(Value).

jafingerhut:
I did a quick check and found nothing in the P4_14 or P4_16 language specs restricting varbit to be the last field of a header.

The P4_14 spec in Section 2.1 "Header Type Declarations" has this reasonable restriction mentioned: "Fields referenced in the length attribute must be located before the variable length field."

That roughly corresponds to the P4_16 restriction that the length value given in the 2-arg extract() method call must have an initialized value.

I also scanned through the BMv2 simple_switch JSON spec here: https://github.com/p4lang/behavioral-model/blob/master/docs/JSON_format.md and the only mention I found was that a header may have at most one varbit field in it, not more than one.  It does not say anything about whether it must be last or not, and I would not be surprised if it supported varbit fields that were not last in the header, as long as the length can be calculated from fields that are all before the position of the varbit field.

mbudiu-vmw:
This is only a p4-14 problem. In p4-16 you cannot write a sensible program that has the varbit before the length.

The varbits does not have to be last. I also wanted the compiler to support headers with multiple varbit fields, but it doesn't yet. So i think the p4 spec should not put artificial constraints.

hesingh:
> I did a quick check and found nothing in the P4_14 or P4_16 language specs restricting varbit to be the last field of a header.
> 
> The P4_14 spec in Section 2.1 "Header Type Declarations" has this reasonable restriction mentioned: "Fields referenced in the length attribute must be located before the variable length field."
> 
> That roughly corresponds to the P4_16 restriction that the length value given in the 2-arg extract() method call must have an initialized value.
> 
> I also scanned through the BMv2 simple_switch JSON spec here: https://github.com/p4lang/behavioral-model/blob/master/docs/JSON_format.md and the only mention I found was that a header may have at most one varbit field in it, not more than one. It does not say anything about whether it must be last or not, and I would not be surprised if it supported varbit fields that were not last in the header, as long as the length can be calculated from fields that are all before the position of the varbit field.

Even the P4-16 spec mentions one varbit field.  See section 12.8.2  about extract that says "The expression headerLvalue must be a l-value representing a header that contains exactly one varbit field. ".
Additionally, the P4-16 spec says "P4 architectures may impose additional constraints on varbit types"

mbudiu-vmw:
Yes, extract can support only one currently, but we can add support for multiple varbits, where the compiler translates an extract on multiple into many extracts of single varbits. So we do allow headers with multiple varbit fields, we just cannot extract into them yet.

jafingerhut:
I agree the language spec says that architectures may impose additional restrictions on varbit fields, but if such restrictions exist for a particular architecture, it seems best putting such checks for those restrictions in architecture-specific parts of p4c, only enabled for those architectures.

It seems that the restriction "varbit field must be last in a header" is _not_ a restriction for any architecture or target that p4c supports today.

hesingh:
> This is only a p4-14 problem. In p4-16 you cannot write a sensible program that has the varbit before the length.

With p4-16, if one cannot write a sensible program with varbit before the length, why not add a check because any naive or newbie may do the reverse?  Also, folks will ask (zma has already), why p4-14 is different from p4-16?  What is the answer?

> 
> The varbits does not have to be last. I also wanted the compiler to support headers with multiple varbit fields, but it doesn't yet. So i think the p4 spec should not put artificial constraints.

mbudiu-vmw:
In p4-14 declarations do not have to come before their use.

jafingerhut:
In P4_14, the length _formula_ is a declarative part of the header definition, if it has a variable-length field.  It is pretty straightforward and "local" for a compiler to check whether any names in that formula are names of fields after the variable-length field, and give an error message in that case.

It is not so "local" of a check to see if names in that length formula occur in headers parsed later, or earlier.  I do not know whether P4_14 allows the length formula to contain field names from other headers, or standard or user-defined metadata, but if it does, then it becomes much tricker in the general case to statically catch the use of an uninitialized value, just as it is potentially tricky to do so in a P4_16 program.

hesingh:
@mbudiu-vmw Even with multiple varbit fields, the requirement does not change.  Lengths of fields cannot be after the varbit fields.  

Thanks for the replying about p4-14.  

Final input from folks - would anyone still like the check to be added for p4-16?

jafingerhut:
I personally think that adding a restriction to p4-16 that a varbit field must be _last_ in a header is adding a restriction that doesn't seem to be imposed by the language itself, and therefore doesn't belong in the front-end of the compiler.  Architectures or targets can implement their own restrictions in appropriate architecture- or target-specific parts of the compiler if they wish.

hesingh:
Ok, then let me close this PR.  Right now the bmv2 backend doesn't seem to care about this check, so I won't bother adding the check to bmv2.

PR #1749, opened 2019-02-26
Reject varbit fields that do not come last in header in P4-14

### LINKED ISSUE #1745: Should varbit field appear after its associated length field in header?
When compiling the P4_14 program below,

```
header_type hd_t {
    fields {
        options        : *;
        length_        : 8;
    }
    length: length_;
    max_length: 60;
}

header hd_t     hd;

parser start {
    extract(hd);
    return ingress;
}

control ingress { }
```

I get this error. I'm assuming it means `length_` needs to appear before `options` in the header?

```
issue_2.p4(6): [--Werror=not-found] error: length_: Not found declaration
    length: length_;
            ^^^^^^^
```

The equivalent program in P4_16, compiles fine however.

```
#include <v1model.p4>

header hd_t {
    varbit<320> options;
    bit<8>  length;
}

struct packet_t {
    hd_t  hd;
}

struct user_metadata_t {
    bit<8>  data;
}

parser TopParser(packet_in b, out packet_t p,
                 inout user_metadata_t m, inout standard_metadata_t meta) {
    state start {
        b.extract(p.hd, (bit<32>)(p.hd.length));
        transition accept;
    }
}
```

Q: Should varbit field appear after its associated length field in header?

One would assume `length` should always be a priori to `options`, since it comes in first "on the wire"?

mbudiu-vmw:
Fixes #1745

### DIFF
--- a/frontends/p4/fromv1.0/converters.cpp
+++ b/frontends/p4/fromv1.0/converters.cpp
@@ -392,6 +392,34 @@ const IR::Type_Varbits *TypeConverter::postorder(IR::Type_Varbits *vbtype) {
     return vbtype;
 }
 
+namespace {
+
+/// Checks that all references in a header length expression
+/// are to fields prior to the varbit field.
+class ValidateLenExpr : public Inspector {
+    std::set<cstring> prior;
+    const IR::StructField* varbitField;
+
+ public:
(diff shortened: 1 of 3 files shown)


mbudiu-vmw:
I will rework this to be more in line with the spec.

PR #1750, opened 2019-02-27
Ability to serialize existing P4Info message

antoninbas:
THis is useful for p4c extensions wich want 2 separate APIs to generate
P4Info and serialize it later.


antoninbas:
Thanks for the quick review Mihai!

PR #1736, opened 2019-02-17
Fix 4 psa tests due to new psa.p4

familysingh:
Fix  #1734 Issue.

Note, typeMap.h[.cpp] would be one place to fix the unify failure bug by adding a method.  However, then, a new method in typeMap.h[.cpp] would have to be defined as static for use in typeUnification.cpp.  However, a static mehtod cannot use variables or other methods in typeMap.  The need is to use getTypeType() which uses getType().  This is why code has been changed to use typeMap in typeUnification.    

Regarding the change of the constraints API from single to 2 args is so that less code changes in the lower layers.

```constraints(typeMap->getSubstitutions(), typeMap)```

PR #1751, opened 2019-02-27
Implement sizeInBits

### LINKED ISSUE #1717: Develop sizeof solution until module system is available in p4c

mbudiu-vmw:
Fixes #1717.

### DIFF
--- a/frontends/p4/typeChecking/typeChecker.cpp
+++ b/frontends/p4/typeChecking/typeChecker.cpp
@@ -743,7 +743,12 @@ TypeInference::assignment(const IR::Node* errorPosition, const IR::Type* destTyp
         return sourceExpression;
     if (!tvs->isIdentity()) {
         ConstantTypeSubstitution cts(tvs, refMap, typeMap, this);
-        return cts.convert(sourceExpression);  // sets type
+        sourceExpression = cts.convert(sourceExpression);  // sets type
+    }
+    if (initType->is<IR::Type_InfInt>()) {
+        sourceExpression = new IR::Cast(destType->getP4Type(), sourceExpression);
+        setType(sourceExpression, destType);
+        setCompileTimeConstant(sourceExpression);
     }
     return sourceExpression;
 }
@@ -1870,6 +1875,13 @@ const IR::Node* TypeInference::binaryArith(const IR::Operation_Binary* expressio
         typeError("%1%: cannot be applied to %2% of type %3%",
                   expression->getStringOp(), expression->right, rtype->toString());
         return expression;
+    } else if (ltype->is<IR::Type_InfInt>() && rtype->is<IR::Type_InfInt>()) {
+        auto t = new IR::Type_InfInt();
+        setType(getOriginal(), t);
+        setType(expression, t);
+        setCompileTimeConstant(expression);
+        setCompileTimeConstant(getOriginal<IR::Expression>());
+        return expression;
     }
 
     const IR::Type* resultType = ltype;
@@ -2882,8 +2894,8 @@ const IR::Node* TypeInference::postorder(IR::MethodCallExpression* expression) {
             if (mem->member == IR::Type_StructLike::sizeInBits &&
                 type->is<IR::Type_StructLike>()) {
                 LOG3("Folding sizeInBits");
-                unsigned w = typeMap->width_bits(type, expression);
-                if (w == 0)
+                int w = typeMap->width_bits(type, expression);
+                if (w < 0)
                     return expression;
                 auto result = new IR::Constant(w);
                 auto tt = new IR::Type_Type(result->type);
--- a/frontends/p4/typeChecking/typeChecker.cpp
+++ b/frontends/p4/typeChecking/typeChecker.cpp
@@ -1469,7 +1469,15 @@ const IR::Node* TypeInference::postorder(IR::Operation_Relation* expression) {
 
     bool equTest = expression->is<IR::Equ>() || expression->is<IR::Neq>();
 
-    if (ltype->is<IR::Type_InfInt>() && rtype->is<IR::Type_Bits>()) {
+    if (ltype->is<IR::Type_InfInt>() && rtype->is<IR::Type_InfInt>()) {
+        // This can happen because we are replacing some constant functions with
+        // constants during type checking
+        setType(getOriginal(), IR::Type_Boolean::get());
+        setType(expression, IR::Type_Boolean::get());
+        setCompileTimeConstant(expression);
+        setCompileTimeConstant(getOriginal<IR::Expression>());
+        return expression;
+    } else if (ltype->is<IR::Type_InfInt>() && rtype->is<IR::Type_Bits>()) {
         auto e = expression->clone();
         auto cst = expression->left->to<IR::Constant>();
         CHECK_NULL(cst);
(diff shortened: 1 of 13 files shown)


hesingh:
It would be good to add a nested struct to the test program and use the struct inside a header.  Test the nested struct for width_bits and also the header.  Looking at the code, the test will pass.  Still it's good to add the test.  

thanks for the quick fix.

hesingh:
It would be also good to add sizeInBytes() to the code to save many developers from doing the math in their p4 code to get the sizeInBytes.

hesingh:
Please also add a varbits field to a header and test.  Varbits width_bits() should be zero and compile-time.

hanw:
Is there any useful program we could write with varbit size being zero? It would be nice if we could have a real use case for this design decision. Otherwise, I rather left it out as unspecified, and come back to it when we have a use case.

hesingh:
This is header in a P4 program

```
header bitvec_hdr {
    row_t row;
    varbit<32> v;
}
```
During compile-time, the width_bits() of such a header will count sizeof `v` as zero.  

One has to test p4c code changes with any field allowed in a header.

hanw:
A better example would be use sizeof varbit in a program that parse IPv4 options. I understand what the current proposal is, but I am not convinced if the proposal makes sense in practice.

jafingerhut:
The reason I proposed that the sizeInBits/Bytes() of a header that contains varbit fields, to count the varbit fields as 0, is exactly because of a use case like defining an IPv4 header with varbit options at the end.

The most common use case I know of for wanting to do sizeInBits/Bytes is tunnel encap/decap length modification calculations.

Getting back sizeInBytes(ipv4_with_possible_options) = 20 bytes is useful in such a program (the length of an IPv4 header without the options).  If you need to include the options length in your calculation, then that can be done by explicitly storing the length of the varbits in a user-defined metadata field.

I would argue that getting back sizeInBytes(ipv4_with_possible_options) = 60 bytes is not very useful in such a program (the maximum length of an IPv4 header with longest possible options).

hesingh:
sizeInBits has the same issue for accounting for possible ipv4 options.  I would support both sizeInBits and sizeInBytes.  Also, my preference is sizeBits and sizeBytes with no "In" - two words are better than three to type.

hesingh:
> Is there any useful program we could write with varbit size being zero? It would be nice if we could have a real use case for this design decision. Otherwise, I rather left it out as unspecified, and come back to it when we have a use case.

Whatever long-winded or useful P4 program you write, p4c still returns zero bits for Varbits at compile-time.  So what's the use of a useful program?  This PR is dealing with compile-time Varbits only.

hesingh:
I pulled a new p4lang/p4c workspace, patched code changes from this PR, and tested sizeof-bmv2.p4.  The test fails.  The P4 program is included below.

```
hemant@ubuntu:~/int/ttt/p4c/build$ ./p4test --std p4-16 ../testdata/p4_16_samples/sizeof-bmv2.p4
In file: /home/hemant/int/ttt/p4c/frontends/p4/typeChecking/typeChecker.cpp:144

hemant@ubuntu:~/int/ttt/p4c/build$ 
```

```
#include <core.p4>

typedef bit<48>  EthernetAddress;

header Ethernet_h {
    EthernetAddress dstAddr;
    EthernetAddress srcAddr;
    bit<16>         etherType;
}

header ipv4_t {
    bit<4>  version;
    bit<4>  ihl;
    bit<6>  dscp;
    bit<2>  ecn;
    bit<16> totalLen;
    bit<16> identification;
    bit<1>  flag_rsvd;
    bit<1>  flag_noFrag;
    bit<1>  flag_more;
    bit<13> fragOffset;
    bit<8>  ttl;
    bit<8>  protocol;
    bit<16> hdrChecksum;
    bit<32> srcAddr;
    bit<32> dstAddr;
    bit<8>  sizeInBits;    
}

struct Parsed_packet {
    Ethernet_h    ethernet;
    ipv4_t        ipv4;
}

struct mystruct1 {}

parser parserI(packet_in pkt,
               out Parsed_packet hdr) {
    state start {
        bit<8> my_local = 1;
        pkt.extract(hdr.ethernet);
        hdr.ipv4.totalLen = (hdr.ethernet.sizeInBits() + hdr.ethernet.sizeInBits());
            16w0x0800: parse_ipv4;
            default: accept;
        }
    }
    state parse_ipv4 {
        bit<16> my_local = 2;
        pkt.extract(hdr.ipv4);
        transition select(hdr.ipv4.version, hdr.ipv4.protocol) {
            (4w0x4, 8w0x06): accept;
            (4w0x4, 8w0x17): accept;
            default: accept;
        }
    }
}
```

mbudiu-vmw:
In this PR is it illegal to call sizeInBits on a header with varbits.
That's probably a bit harsh. If we follow @jafingerhut's suggestion we should also probably add a maxSizeInBits.

hesingh:
> In this PR is it illegal to call sizeInBits on a header with varbits.
> That's probably a bit harsh. If we follow @jafingerhut's suggestion we should also probably add a maxSizeInBits.

Seem ok to do so.  

What about my note about also supporting sizeInBytes?  This way few lines of code in p4c saves P4 programmers to convert bits to bytes in all programs.

mbudiu-vmw:
Since this is not a standard feature it is more imporant to understand how it would work than to implement if fully. If the LDWG decides we need sizeInBytes we can add it as well.

jafingerhut:
I have no objection to adding maxSizeInBits/Bytes as a separate operation that counts varbit fields as their maximum size, rather than 0.  Off hand, it does not seem to me as useful in as many cases as the one that counts varbit as 0, but if it saves more time in discussing the options of what to do, vs. the time to implement it, then I am 100% for it :-)

hesingh:
Agree with Andy that the other work can be a subsequent PR if the LDWG decides to do so.  thanks.

hesingh:
I see p4c code changes to compute width_bits() for `SerEnum`, Varbits, and header stack, but no unit tests exist in the P4 program for these types.  Please add the tests.

mbudiu-vmw:
I have added the methods minSizeInBits and minSizeInBytes to stacks as well.

antoninbas:
@mbudiu-vmw is this PR delayed because of LDWG discussions or are you simply waiting for someone to review?
I'm asking because it would be great if `TypeMap::minWidthBits` could be used for #1736

mbudiu-vmw:
The PR is ready for review. I don't think we need to wait for LDWG. First, it was converging to this solution, and second, we can still make changes if we don't advertise this feature in the spec.

hesingh:
With the checked in code, if a base type in `Type_Bits` is passed to minSizeInBits(), the function causes a compiler bug.  minSizeInBits() invokes this line of code.

https://github.com/p4lang/p4c/blob/master/frontends/p4/typeMap.cpp#L307

getTypeType() calls getType() which reaches this line of code to cause the compiler bug.

https://github.com/p4lang/p4c/blob/master/frontends/p4/typeMap.cpp#L84

I have fixed the issue in my PSA 1.1 PR .  The changed code is ready for review in my PR.  See

https://github.com/p4lang/p4c/pull/1736/commits/38dd644eb813ce8b80a58af975b9f4e7726f607d#diff-85933acf29d903849fbfdece52c797ccR307

PR #1753, opened 2019-02-28
Do not generate negative slice indexes; fixes #1724

### LINKED ISSUE #1724: Corner case handling in strength reduction
This trivial program fragment

    action test(inout bit<16> a, inout bit<2> b) {
        b = (a << 3)[1:0];
    }

gives the error "[--Werror=type-error] error: -2: negative bit index" (with no line number or context information!) when run with p4test.  There are probably other such untested corner cases in strength reducing various combinations of things; I've also seen cases where a single pass of StrengthReduction doesn;t get everything, but repeating it does.

The lack of an error message with any location info is also a quality problem

mbudiu-vmw:


mbudiu-vmw:
@ChrisDodd you should like this PR, it's just 2 lines.

mbudiu-vmw:
This PR is literally 3 lines, including test. Can anyone review it?

Issue #1280, opened 2018-05-10
Compiler cannot read the IR dumped using --toJSON


mbudiu-vmw:
There should be a flag to indicate to the compiler to read the program from a JSON file dumped previously instead of reading a P4 file. An important question is which passes should be run on this file; currently the compiler dumps JSON after the mid-end.

lbanduka:
Hi,
I think that this is fixed in following commit
https://github.com/p4lang/p4c/commit/46609cd1c393ecf1741edd3f5209e16a70bafb8c

mbudiu-vmw:
That fix only allows you to read the json before the mid-end.
But I will close this issue, because it looks like that's enough for now.

PR #1754, opened 2019-03-01
Added support for --fromJSON flag (#1280)

vidakjelena:
* Added --fromJSON flag, as registerOption for all backends.
* Construction : "--fromJSON" "file", where file represents
IR representation json instead of p4 file.

* Modified all backends to skip parsing stage and FrontEnd pass.

* ConvertEnums is skipped in MidEnd pass(when --fromJson flag is used)
since it was already processed during initial IR generation
(checksum already been computed).

* ParseAnnotation is skipped in backend(while using --fromJSON flag),
because InputSources are empty

* Extended class JsonObject with fields and methods
for making JsonObject object, when program is reading from jsonFile
dumped after midEnd previously

* Extended class SourceInfo,
to make object SourceInfo from parameters read from the jsonFile

* Modified function IR::Node::sourceInfoJsonObj() to generate
source_info information when --fromJson flag is used.

Signed-off-by: Jelena Vidakovic <jelena.vidakovic@rt-rk.com>

### DIFF
--- a/backends/bmv2/common/options.h
+++ b/backends/bmv2/common/options.h
@@ -28,6 +28,8 @@ class BMV2Options : public CompilerOptions {
     bool emitExterns = false;
     // file to output to
     cstring outputFile = nullptr;
+    // read from json
+    bool loadIRFromJson = false;
 
     BMV2Options() {
         registerOption("--emit-externs", nullptr,
@@ -37,6 +39,10 @@ class BMV2Options : public CompilerOptions {
         registerOption("-o", "outfile",
                 [this](const char* arg) { outputFile = arg; return true; },
                 "Write output to outfile");
+        registerOption("--fromJSON", "file",
+                [this](const char* arg) { loadIRFromJson = true; file = arg; return true; },
+                "Use IR representation from JsonFile dumped previously,"\
+                "the compilation starts with reduced midEnd.");
     }
 };
 
--- a/backends/bmv2/psa_switch/main.cpp
+++ b/backends/bmv2/psa_switch/main.cpp
@@ -32,6 +32,8 @@ limitations under the License.
 #include "backends/bmv2/psa_switch/midend.h"
 #include "backends/bmv2/psa_switch/psaSwitch.h"
 #include "backends/bmv2/psa_switch/version.h"
+#include "ir/json_loader.h"
+#include "fstream"
 
 int main(int argc, char *const argv[]) {
     setup_gc_logging();
@@ -41,8 +43,10 @@ int main(int argc, char *const argv[]) {
     options.langVersion = CompilerOptions::FrontendVersion::P4_16;
     options.compilerVersion = BMV2_PSA_VERSION_STRING;
 
-    if (options.process(argc, argv) != nullptr)
-        options.setInputFile();
+    if (options.process(argc, argv) != nullptr) {
+            if (options.loadIRFromJson == false)
+                    options.setInputFile();
+    }
     if (::errorCount() > 0)
         return 1;
 
@@ -50,28 +54,49 @@ int main(int argc, char *const argv[]) {
 
     // BMV2 is required for compatibility with the previous compiler.
     options.preprocessor_options += " -D__TARGET_BMV2__";
-    auto program = P4::parseP4File(options);
-    if (program == nullptr || ::errorCount() > 0)
-        return 1;
-    try {
-        P4::P4COptionPragmaParser optionsPragmaParser;
-        program->apply(P4::ApplyOptionsPragmas(optionsPragmaParser));
 
-        P4::FrontEnd frontend;
-        frontend.addDebugHook(hook);
-        program = frontend.run(options, program);
-    } catch (const Util::P4CExceptionBase &bug) {
-        std::cerr << bug.what() << std::endl;
-        return 1;
+    const IR::P4Program *program = nullptr;
+    const IR::ToplevelBlock* toplevel = nullptr;
+
+
+    if (options.loadIRFromJson == false) {
+        program = P4::parseP4File(options);
+
+        if (program == nullptr || ::errorCount() > 0)
+            return 1;
+        try {
+            P4::P4COptionPragmaParser optionsPragmaParser;
+            program->apply(P4::ApplyOptionsPragmas(optionsPragmaParser));
+
+            P4::FrontEnd frontend;
+            frontend.addDebugHook(hook);
+            program = frontend.run(options, program);
+        } catch (const Util::P4CExceptionBase &bug) {
+            std::cerr << bug.what() << std::endl;
+            return 1;
+        }
+        if (program == nullptr || ::errorCount() > 0)
+            return 1;
+    } else {
+        std::filebuf fb;
+        if (fb.open(options.file, std::ios::in) == nullptr) {
+            ::error("%s: No such file or directory.", options.file);
+            return 1;
+        }
+        std::istream inJson(&fb);
+        JSONLoader jsonFileLoader(inJson);
+        if (jsonFileLoader.json == nullptr) {
+            ::error("Not valid input file");
+            return 1;
+        }
+        program = new IR::P4Program(jsonFileLoader);
+        fb.close();
     }
-    if (program == nullptr || ::errorCount() > 0)
-        return 1;
 
     P4::serializeP4RuntimeIfRequired(program, options);
     if (::errorCount() > 0)
         return 1;
 
-    const IR::ToplevelBlock* toplevel = nullptr;
     BMV2::PsaSwitchMidEnd midEnd(options);
     midEnd.addDebugHook(hook);
     try {
@@ -80,7 +105,7 @@ int main(int argc, char *const argv[]) {
             toplevel->getMain() == nullptr)
             return 1;
         if (options.dumpJsonFile)
-            JSONGenerator(*openFile(options.dumpJsonFile, true)) << program << std::endl;
+            JSONGenerator(*openFile(options.dumpJsonFile, true), true) << program << std::endl;
     } catch (const Util::P4CExceptionBase &bug) {
         std::cerr << bug.what() << std::endl;
         return 1;
--- a/backends/bmv2/psa_switch/midend.cpp
+++ b/backends/bmv2/psa_switch/midend.cpp
@@ -86,54 +86,63 @@ class PsaEnumOn32Bits : public P4::ChooseEnumRepresentation {
 PsaSwitchMidEnd::PsaSwitchMidEnd(CompilerOptions& options) : MidEnd(options) {
     auto evaluator = new P4::EvaluatorPass(&refMap, &typeMap);
     auto convertEnums = new P4::ConvertEnums(&refMap, &typeMap, new PsaEnumOn32Bits("psa.p4"));
-    addPasses({
-        new P4::EliminateNewtype(&refMap, &typeMap),
-        new P4::EliminateSerEnums(&refMap, &typeMap),
-        new P4::RemoveActionParameters(&refMap, &typeMap),
-        convertEnums,
-        new VisitFunctor([this, convertEnums]() { enumMap = convertEnums->getEnumMapping(); }),
-        new P4::OrderArguments(&refMap, &typeMap),
-        new P4::TypeChecking(&refMap, &typeMap),
-        new P4::SimplifyKey(&refMap, &typeMap,
-                            new P4::OrPolicy(
-                                new P4::IsValid(&refMap, &typeMap),
-                                new P4::IsMask())),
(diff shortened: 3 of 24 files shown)


backends/bmv2/common/options.h
```diff
@@ -37,6 +39,9 @@ class BMV2Options : public CompilerOptions {
         registerOption("-o", "outfile",
                 [this](const char* arg) { outputFile = arg; return true; },
                 "Write output to outfile");
+        registerOption("--fromJSON", "file",
+                [this](const char* arg) { loadIRFromJson = true; file = arg; return true; },
+                "Use IR representation from JsonFile dumped previously.");
```
mbudiu-vmw: The message should probably indicate that compilation starts after mid-end in this case.

backends/bmv2/psa_switch/main.cpp
```diff
@@ -41,37 +43,60 @@ int main(int argc, char *const argv[]) {
...
+
+            P4::FrontEnd frontend;
+            frontend.addDebugHook(hook);
+            program = frontend.run(options, program);
+        } catch (const Util::P4CExceptionBase &bug) {
+            std::cerr << bug.what() << std::endl;
+            return 1;
+        }
+        if (program == nullptr || ::errorCount() > 0)
+            return 1;
```
mbudiu-vmw: It looks like --toJSON dumps after the mid-end, but you are loading BEFORE the mid-end.
Shouldn't you also skip the mid-end when loading json?

backends/bmv2/psa_switch/midend.cpp
```diff
@@ -90,8 +90,18 @@ PsaSwitchMidEnd::PsaSwitchMidEnd(CompilerOptions& options) : MidEnd(options) {
         new P4::EliminateNewtype(&refMap, &typeMap),
         new P4::EliminateSerEnums(&refMap, &typeMap),
         new P4::RemoveActionParameters(&refMap, &typeMap),
-        convertEnums,
-        new VisitFunctor([this, convertEnums]() { enumMap = convertEnums->getEnumMapping(); }),
+    });
```
mbudiu-vmw: If you load the json after the mid-end you don't need to modify the mid-end.
The fact that you do indicates that something is wrong.

backends/bmv2/simple_switch/simpleSwitch.cpp
```diff
@@ -936,8 +936,17 @@ SimpleSwitchBackend::convert(const IR::ToplevelBlock* tlb) {
...
     auto program = tlb->getProgram();
     // These passes are logically bmv2-specific
-    PassManager simplify = {
-        new ParseAnnotations(),
+    PassManager simplify = {};
+
+    if (BMV2::BMV2Context::get().options().loadIRFromJson == false) {
+        // ParseAnnotations is added only in case of not using --fromJson flag
+        simplify.addPasses({
+            new ParseAnnotations(),
```
mbudiu-vmw: We should make it that ParseAnnotations is a noop if the annotations are already parsed.

backends/bmv2/psa_switch/main.cpp
ChrisDodd: There should probably be a --toJSON option to dump before the midend (after just the frontend).  Possibly other points make sense as well.  So perhaps a 3rd option that specifies which point in the compile that the json is dumped (for --toJSON) and which point we continue with (for --fromJSON).  But for now, the --toJSON/--fromJSON options should be duals, acting at the same point in the compile.  Whether that point is before the midend or after is IMO not that important, though others may desire a particular spot.  So for minimal change, that should probably be after the midend (for now), as that is what --toJSON currently does

vidakjelena: Thank you for all comments. I will investigate if we can skip the midEnd, but the issue was generating toplevel without midEnd.process.

vidakjelena:
@mbudiu-vmw, @ChrisDodd 
Sorry for taking me so long to respond, I was trying to find solution to avoid midEnd pass and I came up with two approaches. I realized that evaluator pass, which is called inside of midEnd pass, must be called anyway. So, the frist solution is to call midEnd pass, but with some condition inside of it, so this time only evaluator will be called instead of calling all of them. That approach will not change the rest of the code and there will be less changes in instantiation the backend-pass.
And the other solution is to create new pass, for example with name reducedMidEnd, which will have its own refMap, typeMap and enumMap, and inside of it the evaluator pass will be called.
In the meantime I also realized that IR representation that is dumped after midEnd pass, does not contain informations about user defined enums. Therefore I supposed that enumMap will be empty and output .json file will not contain information about that enums. Right now I will try to investigate if there is something that I have missed, otherwise I will try to resolve it.
Thank you again for your responses.

mbudiu-vmw:
Calling the evaluator should be easy. You could either call the midend, or just an evaluator pass.

The enum map is interesting. The only solution to that is that we could make an enumMap part of the P4Program IR representation.

In general, if there are any data structures that carry information to the back-end, like the enumMap, these would be lost. We should consider folding all of these into the P4Program.

mbudiu-vmw:
@ChrisDodd @antoninbas @hanw : it looks to me like we have to make the enumMap part of the P4Program IR. In this map the compiler maintains the representation of the enum values that were converted to ints. In this way the information is available in the IR, and it is serialized as json automatically. What do you think?

mbudiu-vmw:
I have filed #1784 to keep track of this.

ChrisDodd:
I think it is expected that information about things that have been removed from the program by midend passes (such as enums) will be lost if you dump and reload the IR as json.

mbudiu-vmw:
The trouble is that the enumMap has to be dumped as JSON in the back-end.
This is information that we carry along the IR from one pass to a later one.

jafingerhut:
Are the JSON files discussed here some format specific to p4c IR?  And thus the BMv2 JSON format is not relevant in this discussion?

mbudiu-vmw:
We are talking about the JSON serialization of the IR. The BMv2 JSON is not relevant.

vidakjelena:
Thank you for discussion about this issue. 
I will update my pull request with the latest changes which you suggested (reduced midEnd pass and changed registerOption message), if you agree. In the meantime I will try to investigate and solve the problem about enumMaps.
I have one more doubt about ParseAnnotations. If I understood well, it should be called from backend pass, but the body should simulate noop. I tried this in the following way, it just returns from ParseAnnotations::postorder method if some condition is true(loadIRfromJson), the compilation is done correctly but about 50% of tests failed. I am not sure if it's possible to achieve or this is an incorrect approach.

mbudiu-vmw:
@liujed : how are annotations saved as json?

liujed:
> @liujed : how are annotations saved as json?

Below is an excerpt of running with `--toJson` on `testdata/p4_16_samples/key-issue-1020_ebpf.p4`, corresponding to the `@name(" headers.ipv4.srcAddr")` annotation [here](https://github.com/p4lang/p4c/blob/0b534f1f21d67197ca91a3aba5dbbf1c1f0d2f42/testdata/p4_16_samples/key-issue-1020_ebpf.p4#L37).
* The `name`, `body`, `expr`, and `kv` fields are as before my changes in #1614.
* The `needsParsing` field is as described in [`base.def`](https://github.com/p4lang/p4c/blob/0b534f1f21d67197ca91a3aba5dbbf1c1f0d2f42/ir/base.def#L215-L220).
* The `body` field is a vector of [`AnnotationToken`](https://github.com/p4lang/p4c/blob/0b534f1f21d67197ca91a3aba5dbbf1c1f0d2f42/ir/base.def#L197-L205)s, each of which represents a lexer token in the annotation body. The `AnnotationToken` class itself is essentially a copy of `P4::P4Parser::symbol_type`.
```
{
  "Node_ID" : 9010,
  "Node_Type" : "Annotation",
  "name" : "name",
  "body" : {
    "Node_ID" : 9011,
    "Node_Type" : "Vector<AnnotationToken>",
    "vec" : [
      {
        "Node_ID" : 503,
        "Node_Type" : "AnnotationToken",
        "token_type" : 357,
        "text" : " headers.ipv4.srcAddr"
      }
    ]
  },
  "needsParsing" : false,
  "expr" : {
    "Node_ID" : 9012,
    "Node_Type" : "Vector<Expression>",
    "vec" : [
      {
        "Node_ID" : 9015,
        "Node_Type" : "StringLiteral",
        "type" : {
          "Node_ID" : 4056,
          "Node_Type" : "Type_String"
        },
        "value" : " headers.ipv4.srcAddr"
      }
    ]
  },
  "kv" : {
    "Node_ID" : 9013,
    "Node_Type" : "IndexedVector<NamedExpression>",
    "vec" : [],
    "declarations" : {}
  }
}
```

mbudiu-vmw:
So the body is preserved even after parsing, and it is saved as JSON as well.
It looks to me like you don't need to call ParseAnnotations again if you have parsed them once.

vidakjelena:
Hello everyone, I have updated my changes as you suggested, please take a look. Thank you :)

backends/p4test/p4test.cpp
```diff
@@ -100,71 +104,70 @@ int main(int argc, char *const argv[]) {
...
     if (program != nullptr && ::errorCount() == 0) {
-        P4::P4COptionPragmaParser optionsPragmaParser;
-        program->apply(P4::ApplyOptionsPragmas(optionsPragmaParser));
+        P4::serializeP4RuntimeIfRequired(program, options);
 
-        if (!options.parseOnly) {
+        if (!options.parseOnly && !options.validateOnly) {
+            P4Test::MidEnd midEnd(options);
+            midEnd.addDebugHook(hook);
+#if 0
```
mbudiu-vmw: maybe this can be deleted. But perhaps that's not the job of this PR.

backends/p4test/p4test.cpp
```diff
@@ -100,71 +104,70 @@ int main(int argc, char *const argv[]) {
...
             }
+            log_dump(program, "After midend");
+            log_dump(top, "Top level block");
         }
-        log_dump(program, "Initial program");
-        if (program != nullptr && ::errorCount() == 0) {
-            P4::serializeP4RuntimeIfRequired(program, options);
+        if (options.dumpJsonFile)
+            JSONGenerator(*openFile(options.dumpJsonFile, true), true) << program << std::endl;
+        if (options.debugJson) {
```
mbudiu-vmw: I wonder whether we still need the debugJson option.

ChrisDodd: We use it by default on all p4test tests, so we have regression tests on the toJson/fromJson code.

vidakjelena:
I solved problem with user defined enums (issue #1784), but the changes depend on this pull request, so I am in doubt. Should I wait for this PR to be merge, and then create new PR with the solutions, or just modify and update this one.

mbudiu-vmw:
How did you solve the problem of the enums?
Anyone want to review this PR or can I just merge it?
Note that I don't think we are testing this PR currently. @vidakjelena : do you have any suggestions on how we could exercise the new code paths without saving and restoring for all tests? Perhaps for some P4 programs we could save the json, reload and check that the P4 output is the same?

vidakjelena:
I have two proposals for implementation, one is to change the convertEnums postorder method to return type instead of nullptr and after that to read enums directly from program and print them in output file. The lack of this solution is that enumMap is saved in program, but when --fromJSON flag is used it is still empty (because convertEnum pass should not be called again). The other one, that I think is better is to make new pass fillEnumMap that will be called in MidEnd pass, but in the new reduced branch, and will fill midEnd.enumMap. Now printing user defined enums will be the same as before.
As far as tests, I was thinking about it in the similar way, but haven't found satisfactory solution. I'll consider it again.

mbudiu-vmw:
If I understand right, you won't rely on the P4Program serialization as json to do everything, you will supplement the output json manually with some extra information. I still think we should try to carry this information in the program. I have started working on this, but got interrupted by other bugs.

vidakjelena:
The other solution, which I prefer, saved all enums in the program and in case of --fromJSON there is fillEnumMap pass. It fills enumMap with user defined enums just like convertEnums does, so now enumMap is not empty anymore. I actually won't supplement the output json manually because the enums are printed from the enumMap as before. 
In this way, the changes are minimal and the code structure remains similar to the former.
I will make commit with these changes and post link of it here, in order to make it easier for you to have a look at the solution and to give suggestions for change.
It is probably possible to change program to carry this information(enumMap exactly), but there may be a need for additional changes in the code structure and passes.

vidakjelena:
Here is link to the solution https://github.com/vidakjelena/p4c/commit/f27be01c46c15a2a43be48a8b9d8c93dfe60384f . Please take a look and give me your feedback and suggestions for change.

mbudiu-vmw:
Since no one else wants to comment I will merge this. We can then build the next stage on top of it.

PR #1757, opened 2019-03-03
p4c-psa emit Counter and DirectCounter externs in json dump 

derekso1:
Implemented counters and direct counters into JSON output for p4c. All tests passed.


mbudiu-vmw:
Are there any tests that exercise this code?

derekso1:
> Are there any tests that exercise this code?

There are not tests in p4c that test the output JSON file with counters and direct counters. No changes were made to the frontend or IR. These changes just allow p4c-psa to include information about the Counter and Direct Counter externs into the "extern_instances" node of the output JSON file.

mbudiu-vmw:
I see what the changes are. The question is whether there is any P4 file in the testdata folder which makes these functions execute. Has the json been validated in some way?

derekso1:
> I see what the changes are. The question is whether there is any P4 file in the testdata folder which makes these functions execute. Has the json been validated in some way?

Oh I see, my apologies, yes, in p4_16_samples, there are P4 files that use Counters and Direct Counters I have been using those files to verify the JSON output to conform with standards outlined in psa_switch.

cc10512:
@mbudiu-vmw it's a chicken and egg issue for the PSA development. They need to generate some JSON to be able to add the support in BMv2 to test it, and while they validated by hand some of this, they need the compiler to generate it. Once the BMv2 support is there, we'll certainly have test cases to exercise this. For now they are using as tests the programs in the PSA repo: https://github.com/p4lang/p4-spec/tree/master/p4-16/psa/examples

backends/bmv2/psa_switch/psaSwitch.cpp
```diff
@@ -684,6 +684,35 @@ void ExternConverter_Counter::convertExternInstance(
...
+    extern_obj->emplace("source_info", inst->sourceInfoJsonObj());
+    ctxt->json->externs->append(extern_obj);
+    Util::JsonArray *arr = ctxt->json->insert_array_field(extern_obj, "attributes");
+
+    // first argument to create a counter is just a number, convert and dump to json
+    // we get a name from param, type and value from the arguments
+    auto attr_obj = new Util::JsonObject();
+    auto arg1 = inst->arguments->at(0)->expression->to<IR::Constant>();
+    auto param1 = eb->getConstructorParameters()->getParameter(0);
+    auto bitwidth = arg1->type->width_bits();
```
mbudiu-vmw: This will crash if the argument is not a constant. You should check and give an error message in this case rather than crashing.

backends/bmv2/psa_switch/psaSwitch.cpp
```diff
@@ -684,6 +684,35 @@ void ExternConverter_Counter::convertExternInstance(
...
+    attr_obj->emplace("name", param1->toString());
+    attr_obj->emplace("type", "hexstr");
+    attr_obj->emplace("value", repr);
+    arr->append(attr_obj);
+
+    // second argument is the counter type, this is psa metadata, the converter
+    // in conversion context will handle that for us
+    auto arg2 = inst->arguments->at(1)->expression->to<IR::Member>();
+    auto param2 = eb->getConstructorParameters()->getParameter(1);
+    auto mem = arg2->member.name;
```
mbudiu-vmw: Same here. You may want to validate the member name as well.

backends/bmv2/psa_switch/psaSwitch.cpp
```diff
@@ -702,6 +731,23 @@ void ExternConverter_DirectCounter::convertExternInstance(
...
+        extern_obj->emplace("id", nextId("extern_instances"));
+        extern_obj->emplace("source_info", inst->sourceInfoJsonObj());
+        ctxt->json->externs->append(extern_obj);
+        Util::JsonArray *arr = ctxt->json->insert_array_field(extern_obj, "attributes");
+
+        // Direct Counter only has a single argument, which is psa metadata
+        // converter in conversion context will handle this for us
+        auto param = eb->getConstructorParameters()->getParameter(0);
+        auto arg = inst->arguments->at(0)->expression->to<IR::Member>();
+        auto mem = arg->member.name;
```
mbudiu-vmw: same here

backends/bmv2/psa_switch/psaSwitch.h
```diff
@@ -47,7 +47,35 @@ class PsaSwitchExpressionConverter : public ExpressionConverter {
...
+          auto bitwidth = param->type->width_bits();
+
+          // encode the counter type from enum -> int
+          if (fieldName == "BYTES") {
+            cstring repr = BMV2::stringRepr(0, ROUNDUP(bitwidth, 32));
+            jsn->emplace("value", repr);
+          } else if (fieldName == "PACKETS") {
+            cstring repr = BMV2::stringRepr(1, ROUNDUP(bitwidth, 32));
+            jsn->emplace("value", repr);
+          } else {
```
mbudiu-vmw: you may want to validate the third case as well and give an error otherwise.

mbudiu-vmw:
You may want to add a negative test, which causes these crashes without the fixes.

backends/bmv2/psa_switch/psaSwitch.cpp
derekso1: Yes, looks like I did some redundant work as well, I switched to using a previous variable that has already been verified "sz".

backends/bmv2/psa_switch/psaSwitch.cpp
derekso1: There is some redundant logic in here as well, changed to match existing formats.

backends/bmv2/psa_switch/psaSwitch.cpp
derekso1: changed

backends/bmv2/psa_switch/psaSwitch.h
derekso1: Added an error reporting function as well, thanks

backends/bmv2/psa_switch/psaSwitch.cpp
```diff
@@ -684,6 +684,41 @@ void ExternConverter_Counter::convertExternInstance(
...
+    extern_obj->emplace("id", nextId("extern_instances"));
+    extern_obj->emplace("source_info", inst->sourceInfoJsonObj());
+    ctxt->json->externs->append(extern_obj);
+    Util::JsonArray *arr = ctxt->json->insert_array_field(extern_obj, "attributes");
+
+    // first argument to create a counter is just a number, convert and dump to json
+    // we get a name from param, type and value from the arguments
+    auto attr_obj = new Util::JsonObject();
+    auto arg1 = sz->to<IR::Constant>();
+    auto param1 = eb->getConstructorParameters()->getParameter(0);
```
mbudiu-vmw: even here you may want to check whether there exists a parameter 0. if someone compiles the wrong model file (or edits the model file) there won't be a parameter 0 and this will crash.

backends/bmv2/psa_switch/psaSwitch.cpp
```diff
@@ -684,6 +684,41 @@ void ExternConverter_Counter::convertExternInstance(
...
+    auto bitwidth = arg1->type->width_bits();
```
mbudiu-vmw: You should use the new function TypeMap::minWidthInBits instead of doing this. This won't work if the type is a typedef, for example.

backends/bmv2/psa_switch/psaSwitch.cpp
```diff
@@ -684,6 +684,41 @@ void ExternConverter_Counter::convertExternInstance(
...
+    cstring repr = BMV2::stringRepr(arg1->value, ROUNDUP(bitwidth, 8));
+    attr_obj->emplace("name", param1->toString());
+    attr_obj->emplace("type", "hexstr");
+    attr_obj->emplace("value", repr);
+    arr->append(attr_obj);
+
+    // second argument is the counter type, this is psa metadata, the converter
+    // in conversion context will handle that for us
+    auto tp = eb->findParameterValue("type");
+    CHECK_NULL(tp);
```
mbudiu-vmw: This will still crash the compiler, but with a nicer error message. I would instead call modelError here as well.

backends/bmv2/psa_switch/psaSwitch.cpp
```diff
@@ -684,6 +684,41 @@ void ExternConverter_Counter::convertExternInstance(
...
+    if (!tp->is<IR::Declaration_ID>()) {
+      modelError("%1%: expected a declaration_id", tp->getNode());
+      return;
+    }
+    auto arg2 = tp->to<IR::Declaration_ID>();
+    auto param2 = eb->getConstructorParameters()->getParameter(1);
```
mbudiu-vmw: Same problem here. there may not exist 2 parameters.

backends/bmv2/psa_switch/psaSwitch.cpp
```diff
@@ -702,6 +737,29 @@ void ExternConverter_DirectCounter::convertExternInstance(
...
+        extern_obj->emplace("id", nextId("extern_instances"));
+        extern_obj->emplace("source_info", inst->sourceInfoJsonObj());
+        ctxt->json->externs->append(extern_obj);
+        Util::JsonArray *arr = ctxt->json->insert_array_field(extern_obj, "attributes");
+
+        // Direct Counter only has a single argument, which is psa metadata
+        // converter in conversion context will handle this for us
+        auto tp = eb->findParameterValue("type");
+        CHECK_NULL(tp);
+        if (!tp->is<IR::Declaration_ID>()) {
```
mbudiu-vmw: Perhaps you can make this if !tp || !tp->is...

backends/bmv2/psa_switch/psaSwitch.cpp
derekso1: Changed

backends/bmv2/psa_switch/psaSwitch.cpp
derekso1: Yes, changed to minWidthInBits

backends/bmv2/psa_switch/psaSwitch.cpp
derekso1: Changed

backends/bmv2/psa_switch/psaSwitch.cpp
derekso1: Changed

backends/bmv2/psa_switch/psaSwitch.cpp
derekso1: Changed in both places.

backends/bmv2/psa_switch/psaSwitch.cpp
mbudiu-vmw: there's no need to reply to every comment, git will close some of them automatically

derekso1:
> Do you want a review from @hanw?

Open to as much feedback as possible, this is my first time working with P4 and p4c, want to be producing good quality work for future PR's thanks!

hanw:
I have looked at the changes, looks good to me. Please work towards adding a stf test based on the changes you made here in the next PR.

mbudiu-vmw:
@hanw, can we even run psa? do we have any stf test for psa program?

hanw:
@mbudiu-vmw Derek and friends in his class are working on psa backend and bmv2-psa implementation. Their first task was to run a helloworld psa program in bmv2-psa, the second task is to implement the support for PSA counters.

jafingerhut:
@hanw If the code changes otherwise look good, it should be reasonable to merge in PSA-specific changes even with no STF test quite yet, true?  I suspect there may be other changes to p4c _and_ psa_switch required before an STF test will pass in p4c's auto-run tests.

hanw:
yes, that's why I approved Derek's PR.

mbudiu-vmw:
Can I merge this?

hanw:
yes.

PR #1760, opened 2019-03-04
Support for header unions in bmv2 simple_switch back-end; multiple test examples

### LINKED ISSUE #561: header_union support
The compiler support for header unions is incomplete.
This applies to both the front-end and to the existing back-ends.

mbudiu-vmw:
We can claim that this fixes #561, but I will file a separate issue with support for header_unions in the ebpf back-end.

### DIFF
--- a/backends/bmv2/common/JsonObjects.cpp
+++ b/backends/bmv2/common/JsonObjects.cpp
@@ -256,6 +256,23 @@ JsonObjects::add_header_stack(const cstring& type, const cstring& name,
     header_stacks->append(stack);
 }
 
+void
+JsonObjects::add_header_union_stack(const cstring& type, const cstring& name,
+                                    const unsigned size, std::vector<unsigned>& ids) {
+    auto stack = new Util::JsonObject();
+    unsigned id = BMV2::nextId("union_stack");
+    stack->emplace("name", name);
+    stack->emplace("id", id);
+    stack->emplace("union_type", type);
+    stack->emplace("size", size);
+    auto members = new Util::JsonArray();
+    stack->emplace("header_union_ids", members);
+    for (auto id : ids) {
+        members->append(id);
+    }
+    header_union_stacks->append(stack);
+}
+
 /// Add an error to json.
 void
 JsonObjects::add_error(const cstring& name, const unsigned type) {
--- a/backends/bmv2/common/JsonObjects.cpp
+++ b/backends/bmv2/common/JsonObjects.cpp
@@ -241,7 +241,7 @@ JsonObjects::add_metadata(const cstring& type, const cstring& name) {
 
 void
 JsonObjects::add_header_stack(const cstring& type, const cstring& name,
-                              const unsigned size, std::vector<unsigned>& ids) {
+                              const unsigned size, const std::vector<unsigned>& ids) {
     auto stack = new Util::JsonObject();
     unsigned id = BMV2::nextId("stack");
     stack->emplace("name", name);
@@ -258,7 +258,7 @@ JsonObjects::add_header_stack(const cstring& type, const cstring& name,
 
 void
 JsonObjects::add_header_union_stack(const cstring& type, const cstring& name,
-                                    const unsigned size, std::vector<unsigned>& ids) {
+                                    const unsigned size, const std::vector<unsigned>& ids) {
     auto stack = new Util::JsonObject();
     unsigned id = BMV2::nextId("union_stack");
     stack->emplace("name", name);
--- a/backends/bmv2/common/deparser.cpp
+++ b/backends/bmv2/common/deparser.cpp
@@ -42,6 +42,7 @@ void DeparserConverter::convertDeparserBody(const IR::Vector<IR::StatOrDecl>* bo
                         auto arg = mc->arguments->at(0);
                         auto type = ctxt->typeMap->getType(arg, true);
                         if (type->is<IR::Type_Stack>()) {
+                            BUG("unreachable");
                             // This branch is in fact never taken, because
                             // arrays are expanded into elements.
                             int size = type->to<IR::Type_Stack>()->getSize();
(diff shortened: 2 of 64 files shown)


testdata/p4_16_samples/issue561-3-bmv2.stf
```diff
@@ -0,0 +1,7 @@
+packet 0 00 00 00 00
+packet 0 01 00 00 00
+packet 0 02 00 00 00
+
+expect 0 00 FF FF
+expect 0 01 FF
```
jafingerhut: I do not see how the above two expected packet output lines can be correct.

I would have expected, I think:
```
expect 0 00 FF FF 00 00
expect 0 01 FF 00
```

mbudiu-vmw: I think that STF specifies a prefix of the packet. So my file implies yours. But yours is more precise.

mbudiu-vmw:
@antoninbas : this Travis test was started after your PR was merged in the behavioral-model, but it failed with the error you have just fixed. Let's see if it still happens, maybe there is some weak consistency in the access to the sources.

jafingerhut:
I really, really hope that STF is intended to match byte for byte, with the exact number of bytes in the STF file compared against the packets sent out by BMv2.  I will do a quick test and report back.

jafingerhut:
Ugh.  You are correct.  This seems unfortunate to me that STF checking isn't "exactly this expected length packet, or failure", but I realize there can be use cases for both.

mbudiu-vmw:
The way @ChrisDodd designed STF, it does not match exactly. This is so you can actually use real packets with payloads, which you can ignore in expect. We could consider adding a flag to the python testing scripts so that they expect the STF expected packets to be exact. This would improve testing. If you think it's a good idea please file an issue. Alternatively, we could add a command 'expectExact', but this would entail editing quite a number of files.

jafingerhut:
Understood.  I have created a PR proposing a new STF command here: https://github.com/p4lang/p4c/issues/1762

Not high priority, but seems nice to have the little bit of extra checking for those who want to write such STF tests.

mbudiu-vmw:
Hm. Travis for Linux fails with an error which @antoninbas just fixed in bmv2.
Why does travis not use a recent build of bmv2?
```
bad json:
{
   "type" : "header_union_stack",
   "value" : "tcp_options_vec"
}
```

cc10512:
It takes some time to rebuild the docker image for bmv2 and propagate that change to p4c.

backends/bmv2/common/JsonObjects.cpp
```diff
@@ -256,6 +256,23 @@ JsonObjects::add_header_stack(const cstring& type, const cstring& name,
     header_stacks->append(stack);
 }
 
+void
+JsonObjects::add_header_union_stack(const cstring& type, const cstring& name,
+                                    const unsigned size, std::vector<unsigned>& ids) {
```
antoninbas: nit: looks like `ids` should be a `const` reference

backends/bmv2/common/deparser.cpp
```diff
@@ -54,13 +55,15 @@ void DeparserConverter::convertDeparserBody(const IR::Vector<IR::StatOrDecl>* bo
                                 ref += "[" + Util::toString(i) + "]";
                                 result->append(ref);
                             }
-                        } else if (type->is<IR::Type_Header>()) {
+                        } else if (type->is<IR::Type_Header>() ||
+                                   type->is<IR::Type_HeaderUnion>()) {
```
antoninbas: I'm not sure this is correct but maybe I'm missing something that i8s happening in the expression converter. The bmv2 JSON documentation says "For stacks and unions, all the header instances need to be listed in the appropriate order.". So shouldn't you iterate over all the union members here (the header instances that you are generating) and add each one to the `result` array?

backends/bmv2/common/header.cpp
```diff
@@ -135,18 +135,41 @@ void HeaderConverter::addHeaderStacks(const IR::Type_Struct* headersStruct) {
...
+            std::vector<unsigned> ids;
+            for (unsigned i = 0; i < stack_size; i++) {
+                cstring hdrName = f->controlPlaneName() + "[" + Util::toString(i) + "]";
+                int id;
+                if (isUnion) {
+                    // We have to add separately a header instance for all
+                    // headers in the union.  Each instance will be named with
+                    // a prefix including the union name, e.g., "u.h"
+                    Util::JsonArray* fields = new Util::JsonArray();
+                    for (auto uf : type->to<IR::Type_HeaderUnion>()->fields) {
```
antoninbas: maybe we could have a helper function for that for loop, since it seems you have the exact same loop below (for `HeaderConverter::init_apply`)

backends/bmv2/common/parser.cpp
```diff
@@ -73,53 +73,83 @@ Util::IJson* ParserConverter::convertParserStatement(const IR::StatOrDecl* stat)
...
+                }
+                auto value = j->to<Util::JsonObject>()->get("value");
+                param->emplace("type", type);
+                param->emplace("value", value);
+
+                if (argCount == 2) {
+                    auto arg2 = mce->arguments->at(1);
+                    auto jexpr = ctxt->conv->convert(arg2->expression, true, false);
+                    auto rwrap = new Util::JsonObject();
+                    // The spec says that this must always be wrapped in an expression
```
antoninbas: I thought `convert` could do that for you already (`wrap` parameter)?
So maybe if you call `auto jexpr = ctxt->conv->convert(arg2->expression, true, true);` you won't need to wrap yourself?

backends/bmv2/common/deparser.cpp
mbudiu-vmw: In fact stacks and header_unions were expanded into components by the expandEmit midend pass. So this code is useless.

PR #1763, opened 2019-03-04
Add option to STF "expect" command for stricter checking of output pa…

jafingerhut:
…ckets

Before and after this change, the default behavior for the STF
"expect" command is to check that the actual output packet from BMv2
is at least as long as the one specified in the "expect" statement,
and all bytes in the "expect" packet data are present at the beginning
of the actual output packet.  If the actual output packet has
additional data after that, the test still passes.

With this change, if you add a "$" character at the end of the packet
data in an "expect" statement, the test will fail in all cases where
"expect" fails today, plus it will also fail if the actual output
packet is longer than the one specified in the "expect" statement.
This enables slightly more strict checking of actual output packets.

### DIFF
--- a/backends/bmv2/bmv2stf.py
+++ b/backends/bmv2/bmv2stf.py
@@ -678,8 +678,15 @@ class RunBMV2(object):
     def comparePacket(self, expected, received):
         received = ''.join(ByteToHex(str(received)).split()).upper()
         expected = ''.join(expected.split()).upper()
+        strict_length_check = False
+        if expected[-1] == '$':
+            strict_length_check = True
+            expected = expected[:-1]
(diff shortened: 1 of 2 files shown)


jafingerhut:
I have added "$" at the end of all expect statements in an STF file that I added earlier, to verify that the new code is working as expected.

I can add such things to many more existing STF tests if people do not mind, but I already know from a quick test that there are existing STF files that only specify a prefix of the actual output packet, so "$" should not be added to the end of all existing expect statements.

jafingerhut:
@mbudiu-vmw Probably ready to merge, unless you have further comments.  Feel free to do so if you think it is ready.

PR #1766, opened 2019-03-06
Add program that tests p4c's error-detection capability for bad type's

jafingerhut:
This is the test program I wrote while determining which types were
supported for typedef and type while creating this proposed table to
add to the P4_16 language spec:

https://github.com/p4lang/p4-spec/pull/731

### DIFF
--- a/testdata/p4_16_errors/typedef-and-type-definitions.p4
+++ b/testdata/p4_16_errors/typedef-and-type-definitions.p4
@@ -106,7 +106,7 @@ typedef tuple<bit<8>, varbit<17> > Tuple2D_t;
 // Every line marked '// error' below causes the compiler to issue an
 // error message of this form:
 
-// error: <type_name>: 'type' can only be applied to base types or tuple types
+// error: <type_name>: 'type' can only be applied to base types
 
 
 type    int        intT_t;
@@ -123,8 +123,8 @@ type    h1_t[3]    h1StackT_t;  // error
 type    hu1_t      hu1T_t;  // error
 type    s1_t       s1T_t;  // error
 type    struct PointB_t { int<32> x; int<32> y; } PointBD_t;  // error
-type    tuple<bit<8>, bit<17> > Tuple1T_t;
-type    tuple<bit<8>, varbit<17> > Tuple2T_t;  // should this be an error, since it contains a varbit element?
+type    tuple<bit<8>, bit<17> > Tuple1T_t;  // error
+type    tuple<bit<8>, varbit<17> > Tuple2T_t;  // error
 
 
 // typedef on top of another typedef or type are allowed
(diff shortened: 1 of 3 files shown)


jafingerhut:
Mihai, this is the program I mentioned at the last LDWG meeting that you suggested adding as a test case to p4c.

I added one more quick line before submitting it, and it looks a little bit odd to me.  The compiler gives an error for `type varbit<17> foo;`, but not for `type tuple <bit<8> varbit<17> > foo;`.  Would you expect it should give an error if the type of a tuple element is not supported "directly" in a `type` definition?

testdata/p4_16_errors/typedef-and-type-definitions.p4
```diff
@@ -0,0 +1,158 @@
...
+// syntax error, unexpected VOID, expecting ENUM or HEADER or HEADER_UNION or STRUCT
+//type    void       voidT_t;
+// syntax error, unexpected MATCH_KIND, expecting ENUM or HEADER or HEADER_UNION or STRUCT
+//type    match_kind matchT_t;
+
+
+// Every line marked '// error' below causes the compiler to issue an
+// error message of this form:
+
+// error: <type_name>: 'type' can only be applied to base types or tuple types
```
mbudiu-vmw: The spec does not seem to say much about restrictions on type. that looks like a bug. Do we really want to allow tuple?

mbudiu-vmw: From the error message it looks like we do want to allow tuples, but I don't remember what the use cases for tuples are.

jafingerhut: I do not recall reading any description of why tuples are (or should be) supported for `type` definitions, but did notice them mentioned in the error message explicitly like that.

I am not aware of any reasons for the PSA architecture's use of `type` definitions, for runtime translation, that it needs support for `type` on top of a tuple type.  So far it only needs `type` on top of a `bit<W>` type.

testdata/p4_16_errors/typedef-and-type-definitions.p4
```diff
@@ -0,0 +1,158 @@
...
+// order in which they have been "stacked", e.g. EthDT_t is a `type`
+// (the T is last) defined on type of a `typedef` (the D just before
+// the T).
+
+
+//////////////////////////////////////////////////////////////////////
+// typdef
+//////////////////////////////////////////////////////////////////////
+
+// int, void, match_kind cause syntax error that causes compiler to
```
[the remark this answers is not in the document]
jafingerhut: I had not tried `typedef int foo;` again after latest p4c changes.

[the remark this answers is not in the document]
jafingerhut: I have updated the PR to include lines with type and typedef definitions based on `int` type, which do behave more like the other types given recent p4c changes for supporting `int`.

mbudiu-vmw:
This program will have to be modified if we accept #1772

jafingerhut:
Understood, and happy to update it after that other change is merged.

Issue #1767, opened 2019-03-06
Compiler validation for types allowed in Newtype statements is too weak.


mbudiu-vmw:
See #1766

mbudiu-vmw:
I think we can close this now.

PR #1772, opened 2019-03-07
Do not allow "type tuple<...> X"

mbudiu-vmw:
The spec does not require tuples, and we currently have no uses for tuples.


mbudiu-vmw:
Fixes #1767

PR #1773, opened 2019-03-07
actionSynthesis propagate source info

ChrisDodd:
- propagate source info from statements to newly created actions and
  tables; allows later passes to generate errors/warnings referring to
  the code that the tables were created from rather than just a
  meaningless name like tbl_act_14

### DIFF
--- a/midend/actionSynthesis.cpp
+++ b/midend/actionSynthesis.cpp
@@ -73,7 +73,7 @@ const IR::Node* DoMoveActionsToTables::postorder(IR::MethodCallStatement* statem
 
     auto annos = new IR::Annotations();
     annos->add(new IR::Annotation(IR::Annotation::hiddenAnnotation, {}));
-    auto tbl = new IR::P4Table(tblName, annos, props);
+    auto tbl = new IR::P4Table(statement->srcInfo, tblName, annos, props);
     tables.push_back(tbl);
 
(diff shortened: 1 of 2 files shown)


midend/actionSynthesis.cpp
```diff
@@ -191,7 +193,7 @@ const IR::Statement* DoSynthesizeActions::createAction(const IR::Statement* toAd
     actions.push_back(action);
     auto actpath = new IR::PathExpression(name);
     auto repl = new IR::MethodCallExpression(actpath);
-    auto result = new IR::MethodCallStatement(repl->srcInfo, repl);
+    auto result = new IR::MethodCallStatement(toAdd->srcInfo, repl);
```
mbudiu-vmw: Maybe the expression above should also use toAdd->srcInfo.

mbudiu-vmw:
The tests fail because you are probably using a source position which is after the place where it is used, so resolveReferences refuses to use it. In the generated programs declarations should still come after uses.

ChrisDodd:
Actually, they were failing because they were using a name with no source position (it was compiler generated).  Such names should always be considered after anything with a source location, not before.

midend/actionSynthesis.cpp
ChrisDodd: makes sense

PR #1776, opened 2019-03-08
conditionally disable warnings on constant folding

zma-barefoot:


mihaibudiu (commented):
It would be nice to have an example program on which this warning is too verbose.

zma-barefoot:
Yes, see `p4_16_samples/constant_folding.p4 `.

We warn the overflow twice. Once in constant folding, and again in IR::Constant constructor.

```
590/1257 Test #579: p4/testdata/p4_16_samples/constant_folding.p4 .................................***Failed    0.50 sec
--- /p4c/testdata/p4_16_samples_outputs/constant_folding.p4-stderr	2019-03-08 22:38:05.000000000 +0000
+++ ./tmpMCknte/constant_folding.p4-stderr	2019-03-08 22:47:22.162108135 +0000
@@ -1,9 +1,6 @@
 constant_folding.p4(95): [--Wwarn=overflow] warning: <<: Shifting 8-bit value with 9
         z = 8w1 << 9;
             ^^^^^^^^
-constant_folding.p4(95): [--Wwarn=mismatch] warning: 512: value does not fit in 8 bits
-        z = 8w1 << 9;
```

mbudiu-vmw:
This means that your PR will fail, since you haven't submitted the updated file.
But that looks like a bug, since 8w1 << 9 should not be 512, but zero.
Can you submit another test to validate this fact?

zma-barefoot:
I verified in the debugger the evaluation of 8w1 << 9 is correct. IR::Constant constructs 0, and hence the warning.

mihaibudiu (changes requested):
I don't think you should mute this warning. This would give no warning for z = 8w2 << 8.
Perhaps you want to add this line as a test case.

zma-barefoot:
You're right. Actually, the first warning issued by `DoConstantFolding` is redundant. The warning given by `IR::Constant` constructor should be sufficient.

mbudiu-vmw:
The other warning is also necessary for right shifts.

cc10512:
@zma-barefoot if you set both warnings to have the same warning type, only one will be printed. Right now one is tagged as `ErrorType::WARN_OVERFLOW` and one is tagged as `ErrorType::WARN_MISMATCH`.

zma-barefoot:
What I actually want is to turn off warning in IR::Constant if warning of constant folding is disabled, so let me try again. See latest.

cc10512:
@mbudiu-vmw we are using expression evaluation internally to unroll parser loops and this causes the same warning to be repeatedly issued on the same line. The current mechanism of line number and IR::Node does not work since we iterate through the possible values of the expression and thus the expression changes. 

Can you propose a solution that's acceptable to you to suppress warnings when we evaluate such loops internally?

mbudiu-vmw:
Ideally you should generate a constant that does not need to give a warning. E.g, the value is in the expected range.

zma-barefoot:
Resuming this thread, all this PR does is to propagate the `warnings` flag from `DoConstantFolding` to `IR::Constant`'s ctor. So, can I get a review again? @mbudiu-vmw

cc10512 (approved):
LGTM: Since we are initializing this pass with a warnings flag, it does make sense to use it throughout the pass and squash warnings in the methods it calls.

mbudiu-vmw:
The reference outputs should change.

zma-barefoot:
> The reference outputs should change.

Done in https://github.com/p4lang/p4c/pull/1776/commits/2b8ce312cbef0aa609f92fedfb24535eb6e8f954

testdata/p4_14_samples_outputs/issue1013.p4-stderr
```diff
@@ -0,0 +1,3 @@
+issue1013.p4(13): [--Wwarn=mismatch] warning: -0x1: negative value with unsigned type
+    modify_field(h.x, h.x, 0x0);
```
mbudiu-vmw: Something is wrong here. There are a few other strange error messages.

zma-barefoot: Strange indeed. Investigating this further, this is caused by the (strange?) 14-to-16 translation. In particular, the `modify_field` is translated as below for some reason. That said, I think this PR didn't cause this strange error message but simply revealed it.

```
771: ./tmp6u3htctk/issue1013.p4(24): [--Wwarn=mismatch] warning: -0x1: negative value with unsigned type
771:         hdr.h.x = hdr.h.x & ~32w0x0 | hdr.h.x & 32w0x0;
771:                              ^^^^^^
```

testdata/p4_16_samples_outputs/issue1334.p4-stderr
```diff
@@ -0,0 +1,3 @@
+issue1334.p4(28): [--Wwarn=mismatch] warning: -2: negative value with unsigned type
+    d(b = 2) dinst;
+          ^
```
zma-barefoot: The strangeness here is caused by parameter substitution. Somehow, the source info for the expression is given for `d(b = 2) dinst;` instead of `x = -b;`. Again, I think this PR simply exposed this issue ...

```
control d(out bit<32> x)(bit<32> b) {
    apply {
        x = -b;
    }
}

...

    d(b = 2) dinst;

```

testdata/p4_16_samples_outputs/named-arg1.p4-stderr
```diff
@@ -0,0 +1,3 @@
+named-arg1.p4(36): [--Wwarn=mismatch] warning: -3: negative value with unsigned type
+        a(bi = 3, mb = xv);
```
zma-barefoot: See previous comment.

testdata/p4_14_samples_outputs/issue1013.p4-stderr
ChrisDodd: The translation looks ok -- this is a modify field with mask, so it becomes `dest = dest & ~mask | src & mask`; setting just the masked bits of the destination, leaving other bits unchanged.  The weirdness come from why it thinks `~32w0x0` is a negative constant -- it is not.  It should be `32w0xffff` after constant folding.  Not sure where the -1 comes from.

testdata/p4_16_samples_outputs/synth-action.p4-stderr
```diff
@@ -0,0 +1,3 @@
+synth-action.p4(22): [--Wwarn=mismatch] warning: 4294967302: value does not fit in 32 bits
+            x = x - 6;
```
zma-barefoot: This one is truly strange and makes me scratch my head as to where the negative value is coming from?

```
control c(inout bit<32> x) {
    apply {
        x = 10;
        if (x == 10) {
            x = x + 2;
            x = x - 6;
        } else {
            x = x << 2;
        }
    }
}
```

ChrisDodd: If you put a breakpoint in the `warning` function, you can figure out which `IR::Node` that value is coming from.  You can then put a breakpoint in IR::Node::traceCondition with a condition `id == `... to see where that specific IR::Node gets created...

zma-barefoot: Okay, this is because we evaluate

```
x = x - 6;
```

as

```
x= x + (-6);
```

And `IR::Constant` thinks -6 is is a large constant ...

testdata/p4_14_samples_outputs/issue1013.p4-stderr
zma-barefoot: This seems to be simply caused by the comparison `< 0` with `value` of type `bit_int` which is `boost::multiprecision::mpz_int` or `cpp_int` which treats all values as signed?

https://github.com/p4lang/p4c/blob/90a8f3f7cbd242c8759f66cb7cf8e3966a173e18/ir/expression.cpp#L94

zma-barefoot:
@mbudiu-vmw As I mentioned in the comments, the strangeness and impreciseness of the error messages are largely caused by the treatment of the signed and negative values in the IR. If it's okay, I would like to have this PR merged and address those issues in the future.

mbudiu-vmw:
At the very least we should have an issue filed.

zma-barefoot:
Created issue https://github.com/p4lang/p4c/issues/2122

mbudiu-vmw:
Can you explain again what the point of this PR is? It seems to cause the compiler to generate worse diagnostics overall, and I don't see any case which is improved. The point of the PR seems to be improving the diagnostics, but, at least on the test cases we have, it does not seem to do that. I would really like to see a fix for #2122 bundled within this PR.

fruffy:
@mbudiu-vmw Many old pull requests such as this one are also not being worked on any more. Consider closing them, too?

mbudiu-vmw:
I would really prefer the authors of the PRs themselves to close them.

fruffy:
I believe a lot of these are by inactive accounts or people that do not work at the respective company any more. So it is unclear if they will ever be updated, is there a cut-off time?

mbudiu-vmw:
Technically your github account is not tied to a company, but to a person.
The should be able to close them.

PR #1786, opened 2019-03-11
move annotations on header to fields

hanw:
This change will keep the annot on struct with the fields that the struct is flattened to.
```
@my_annot
struct st {
    bit<8> f0; 
}

header hd{
   st s;
}
```

```
header hd {
   @my_annot _s_f0;
}
```

### DIFF
--- a/midend/flattenHeaders.cpp
+++ b/midend/flattenHeaders.cpp
@@ -19,17 +19,6 @@ limitations under the License.
 
 namespace P4 {
 
-template <typename Func>
-const IR::Annotations* filterAnnotation(const IR::Annotations* annot, Func function) {
-    auto annotations = new IR::Annotations();
-    for (auto an : annot->annotations) {
-        if (function(an))
-            continue;
-        annotations->add(an);
-    }
-    return annotations;
-}
-
 void HeaderTypeReplacement::flatten(const P4::TypeMap* typeMap,
                                     cstring prefix,
                                     const IR::Type* type,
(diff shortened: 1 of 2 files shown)


mbudiu-vmw:
You can't do this for all annotations; if you move for example `@name` annotations this will break the back-end which uses the original annotations or names to output the names.

hanw:
I added code to avoid copying name annotation.

midend/flattenHeaders.cpp
```diff
@@ -19,20 +19,31 @@ limitations under the License.
 
 namespace P4 {
 
+template <typename Func>
```
mbudiu-vmw: The Annotations class already has a function called 'where' which does exactly this.

midend/flattenHeaders.h
```diff
@@ -177,6 +178,7 @@ class FlattenHeaders final : public PassManager {
         passes.push_back(new FindHeaderTypesToReplace(sm));
         passes.push_back(new ReplaceHeaders(sm));
         passes.push_back(new ClearTypeMap(typeMap));
+        passes.push_back(new TypeChecking(refMap, typeMap, true));
```
mbudiu-vmw: In general you should not end with type checking; if the next pass needs it, it should call it itself.

midend/flattenHeaders.cpp
hanw: fixed.

midend/flattenHeaders.h
hanw: fixed.

midend/flattenHeaders.cpp
cc10512: please remove this function then.

hanw:
> This code will probably work, but it may need to be revisited if we add other annotations.
> It may also interfere with other back-end specific annotations.

I added a policy to this pass to allow backend to customize the annotation that it is willing to keep during flattening. By default, all annotation will be lost. Backend must implement the policy if it needs certain annotation to be kept from struct to fields.

PR #1787, opened 2019-03-12
Handle Newtype for table keys; fixes #1739

### LINKED ISSUE #1739: p4test gives error for table key with `type` type
The attached program p416-type-use1.p4 gives an error even when one attempts to compile it using `p4test`, e.g.:
```
$ p4test p416-type-use1.p4
p416-type-use1.p4(99): [--Werror=type-error] error: Key hdr.ipv4.dstAddr field type must be a scalar type; it cannot be IPv4AddressUint_t
            hdr.ipv4.dstAddr: lpm;
            ^^^^^^^^^^^^^^^^
p416-type-use1.p4(110): [--Werror=type-error] error: Key hdr.ipv4.srcAddr field type must be a scalar type; it cannot be IPv4Address_t
            hdr.ipv4.srcAddr : ternary;
            ^^^^^^^^^^^^^^^^
```
This is the result with the latest version of p4c source as of 2019-Feb-18:
```
$ git log -n 1 | cat
commit 6b761be48da3b1f1c19b6c9f6d911b378ead5bb6
Author: Han Wang <hanw@users.noreply.github.com>
Date:   Mon Feb 18 07:21:06 2019 -0800
```

I can understand if a P4_16 architecture rejects table search keys that do not have one of a restricted set of types, but it seems odd that p4test does.

p4c-bm2-ss also fails with the same errors.  Here it seems plausible that the desired behavior is to allow this, if the underlying type of the `type` of the key fields is `bit<W>` or `int<W>`, which it is in this case.

Tried this example while looking into support for PSA in p4c-bm2-psa.  If architecture-independent parts of p4c reject such programs, is there a straightforward way for a particular architecture to override that?

[p416-type-use1.p4.txt](https://github.com/p4lang/p4c/files/2877537/p416-type-use1.p4.txt)

mbudiu-vmw:


hesingh:
I have asked several times to review my PSA 1.1. PR (https://github.com/p4lang/p4c/pull/1736) which has changed the same code as this PR.  It would have been good to review my PR rather then sending out this PR.  The change to p4RuntimeSeriliazer.cpp has already been made in my PR.  The changes to typeChecker.cpp are along the same lines.

hesingh:
Seeing this PR's Travis failures, here is a thought.

It makes sense to first check in my PSA 1.1 PR <https://github.com/p4lang/p4c/pull/1736>.  

The reason is because my PR has new tests and also merged with minSizeInBits code.  The merge has uncovered two issues with minSizeInBits code, viz., a crash and gtestp4c failure.  I have one comment not addressed from today regarding gtestp4c failure - I have contacted Antonin and cced Mihai.  I don't see any other issue pending from review today - please keep me honest.  

I have addressed all of Antonin's comments as well.

control-plane/p4RuntimeSerializer.cpp
```diff
@@ -661,10 +661,11 @@ getMatchFields(const IR::P4Table* table, ReferenceMap* refMap, TypeMap* typeMap)
           typeMap->getType(keyElement->expression->getNode(), true);
         BUG_CHECK(matchFieldType != nullptr,
                   "Couldn't determine type for key element %1%", keyElement);
-
+        while (matchFieldType->is<IR::Type_Newtype>())
```
antoninbas: This seems to be just a quick work around to get the P4Info to generate. According to the P4Runtime spec, the type information needs to be preserved for match fields.

hesingh: I think it's a workaround to get around the gtestp4c failure.  I have proposed that the code below is added to minWidthBits() and then the gtestp4c failure is fixed.  This is the code change is included in my PR.  I think this PR should only touch typeChecker.cpp and leave the rest to my PR. 

```
   // Check first if incoming type is bits, otherwise getTypeType will
    // cause failure in getType and p4runtime gtestp4c tests fail.
    if (type->is<IR::Type_Bits>()) {
	  return type->width_bits();
    }
```

hesingh:
@mbudiu-vmw You provided a review comment in my PR to use a function and not write code similar to how you changed p4RuntimeSerializer.cpp in this PR. 

https://github.com/p4lang/p4c/pull/1736/#discussion_r260529211

I also added a note yesterday, saying to please don't change p4RuntimeSerializer.cpp.  Having used the changes to p4RuntimeSerializer.cpp from this PR, a table key in a `SerEnum` fails to compute bitwidth.   Please see this test in my PR (https://github.com/p4lang/p4c/pull/1736/)

`testdata/p4_16_samples/table-key-serenum.p4`

The code has to use a function as follows:
```
      int width = getTypeWidth(matchFieldType, typeMap);
        if (width < 0)
            return matchFields;
```

The only catch is, if one uses the above code, gtestp4c test fails p4runtime.cpp's P4_16_MatchFields sub-test in getType().  I don't fully understand the failure and this is what needs chasing down.

mbudiu-vmw:
If you found a bug please file an issue with a P4 program that reproduces it.

hesingh:
My PR has an Issue filed.  The scope of my PR is narrow:

Support `SerEnum` and `type` (nested) for table key and Action param for p4info generation.

The code changes update two files in p4c/control-plane/.  I have closed most issues from Antonin's review yesterday.  I would like to close the old PR and issue a new one to expedite code review.   The only issue blocking the closing is to find root cause to gtestp4c p4runtime sub-tests failure.  Fixing the failure would require more changes which I'd make in the current PR, then close it.

PR #1792, opened 2019-03-15
p4c changes to support PSA1.1

hesingh:
This is a new PR after closing the old one in https://github.com/p4lang/p4c/pull/1736

The new PR consolidates all changes into one for ease of code review.


hesingh:
Code in this PR is ready for review - thanks.

hesingh:
2nd try.  Please review this PR.

control-plane/p4RuntimeSerializer.cpp
```diff
@@ -639,6 +640,45 @@ getMatchType(cstring matchTypeName) {
     }
 }
 
+static int
+getTypeWidth(const IR::Type* type, TypeMap* typeMap) {
+    // p4runtime tests use bit Slices which are Type_Bits
+    // which will cause failure in minWidthBits.
+    // Thus compute width here for Slices and any other bit-field.
+    if (auto tb = type->to<IR::Type_Bits>()) {
```
antoninbas: @mbudiu-vmw This sounds like an issue with `minWidthBits` or the type map, and not something for which we should have a workaround here. Mihai, do you know what could be happening. If I understand the issue correctly, when a match field is a Slice expression, and we call this function (`getTypeWidth`) on the match field's type (which is `Type_Bits`), the compiler crashes (I'm sure Hemant can provide a backtrace). Things work fine when the match field is just a plain field path.

hesingh: If a Slice's type is sent to `getTypeType`, it calls `getType` which ends up into this `BUG` condition.

https://github.com/p4lang/p4c/blob/master/frontends/p4/typeMap.cpp#L84

The reason is because the `element` is not found in the typeMap.  `getTypeType` is called by `minWidthBits` and also my new code in p4runtime serializer.  The reason the type is not found in the typeMap is because any caller to minWidthBits or getTypeName is only getting the type for the Slice.  The actual Slice expression is `h.h.headerField[13:4]` which exists in typeMap.

The Slice IR dump from getMatchFields is included below when gtestp4c is run.

```
[1223064] KeyElement
  annotations: [1145486] Annotations
    [1145488] Annotation name=name needsParsing=0
      [1145492] StringLiteral value=h.h.headerField[13:4]
        type: [11866] Type_String
  expression: [1223072] Slice
    type: [959116] Type_Bits size=10 isSigned=0
    e0: [1223073] Member member=headerField
      type: [6419] Type_Bits size=16 isSigned=0
      expr: [1223074] Member member=h
        type: [1119732] Type_Header name=Header declid=3516
          annotations: [2] Annotations
          [1119729] StructField name=headerField declid=9657
            annotations: [2] Annotations
            type: [6419] Type_Bits size=16 isSigned=0...
        expr: [1223075] PathExpression
          type: [1217347] Type_Struct name=Headers declid=3703
            annotations: [2] Annotations
            [1217342] StructField name=h declid=10176
              annotations: [2] Annotations
              type: [1119732] Type_Header name=Header declid=3516...
            [1217343] StructField name=hStack declid=10177
              annotations: [2] Annotations
              type: [1208349] Type_Stack
                elementType: [1119732] Type_Header name=Header declid=3516...
                size: [1119771] Constant value=4 base=10
                  type: [1119770] Type_InfInt declid=1162
            [1217344] StructField name=hUnion declid=10178
              annotations: [2] Annotations
              type: [1217284] Type_HeaderUnion name=HeaderUnion declid=3702
                annotations: [2] Annotations
                [1217279] StructField name=a declid=10170
                  annotations: [2] Annotations
                  type: [1119732] Type_Header name=Header declid=3516...
                [1217280] StructField name=b declid=10171
                  annotations: [2] Annotations
                  type: [1119740] Type_Header name=AnotherHeader declid=3517
                    annotations: [2] Annotations
                    [1119737] StructField name=anotherHeaderField declid=9658
                      annotations: [2] Annotations
                      type: [190009] Type_Bits size=8 isSigned=0
                [1217281] StructField name=c declid=10172
                  annotations: [2] Annotations
                  type: [1217245] Type_Header name=YetAnotherHeader declid=3701
                    annotations: [2] Annotations
                    [1217242] StructField name=yetAnotherHeaderField declid=10168
                      annotations: [2] Annotations
                      type: [1119725] Type_Newtype name=CustomT_t declid=3515
                        annotations: [1119723] Annotations
                        type: [190009] Type_Bits size=8 isSigned=0...
          path: [1120128] Path name=h absolute=0
    e1: [1120132] Constant value=13 base=10
      type: [1120131] Type_InfInt declid=1174
    e2: [1120134] Constant value=4 base=10
      type: [1120133] Type_InfInt declid=1175
  matchType: [1145502] PathExpression
    type: [11847] Type_MatchKind
    path: [1120137] Path name=ternary absolute=0
```

mbudiu-vmw: The function getTypeType is being called with a type that has not been inserted in the typeMap. Typechecking inserts all types that appear in the IR in the typemap, so some other pass has created this type node without inserting it into the typeMap. The way this is debugged is by setting a conditional breakpoint on the creation of the id of the IR::Node, which is unique.

control-plane/p4RuntimeSerializer.cpp
```diff
@@ -197,6 +197,7 @@ struct MatchField {
     const uint32_t bitwidth;  // How wide this field is.
     const IR::IAnnotated* annotations;  // If non-null, any annotations applied
                                         // to this field.
+    const cstring type_name;  // Optional field, used if field is user-defined type.
```
mbudiu-vmw: The comment has to explain what the field is, not only when it is used.

control-plane/p4RuntimeSerializer.cpp
```diff
@@ -639,6 +640,45 @@ getMatchType(cstring matchTypeName) {
...
+    auto ann = type->getAnnotation("p4runtime_translation");
+    if (ann != nullptr) {
+        auto sdnB = ann->expr[1]->to<IR::Constant>();
+        return static_cast<int>(sdnB->value.get_si());
+    }
+    return typeMap->minWidthBits(type, type->getNode());
+}
+
+/*
+ * The function returns a cstring for use as type_name for any nested type.
```
mbudiu-vmw: this comment is not very helpful. The fact it's a cstring is read from the function type. There is no notion of "nested type" in the P4 spec, so it is not clear what this means. This looks to me like it is returning the name of a type node if the node is Type_Newtype.

control-plane/p4RuntimeSerializer.cpp
mbudiu-vmw: If the type is not expected to be in the typeMap maybe we should not use typeMap->getMinWidth anyway. The function to compute the width should be very simple for the types supported by the control-plane.

control-plane/p4RuntimeSerializer.cpp
```diff
@@ -767,15 +813,23 @@ class P4RuntimeAnalyzer {
...
             }
-            if (paramType->is<IR::Type_Boolean>()) {
-                param->set_bitwidth(1);
-            } else {
-                param->set_bitwidth(paramType->width_bits());
+            int w = getTypeWidth(paramType, typeMap);
+            if (w < 0)
+                return;
+            param->set_bitwidth(w);
+            if (paramType->is<IR::Type_Newtype>()) {
```
mbudiu-vmw: This "if" is not necessary.

control-plane/typeSpecConverter.cpp
```diff
@@ -111,6 +115,48 @@ bool TypeSpecConverter::preorder(const IR::Type_Name* type) {
     return false;
 }
 
+// This function is invoked if psa.p4 has a Newtype. The function
```
mbudiu-vmw: This code has nothing to do with PSA, so there should be no comments related to PSA in it. This file applies to all architectures.

There is no notion of user-defined NewType. I don't know what this comment means.

control-plane/typeSpecConverter.cpp
```diff
@@ -111,6 +115,48 @@ bool TypeSpecConverter::preorder(const IR::Type_Name* type) {
...
+        const IR::StringLiteral* uri;
+        const IR::Constant* sdnB;
+        auto ann = type->getAnnotation("p4runtime_translation");
+        if (ann != nullptr) {
+            orig_type = false;
+            uri = ann->expr[0]->to<IR::StringLiteral>();
+            BUG_CHECK(uri, "P4runtime annotation does not have uri: %1%",
+                      type->getP4Type());
+            sdnB = ann->expr[1]->to<IR::Constant>();
+            BUG_CHECK(sdnB, "P4runtime annotation does not have sdn: %1%",
```
mbudiu-vmw: This should be an error, not a BUG; the user can supply an incorrect expression, and this should not crash the compiler.

control-plane/typeSpecConverter.cpp
```diff
@@ -111,6 +115,48 @@ bool TypeSpecConverter::preorder(const IR::Type_Name* type) {
...
+    if (p4RtTypeInfo) {
+        bool orig_type = true;
+        const IR::StringLiteral* uri;
+        const IR::Constant* sdnB;
+        auto ann = type->getAnnotation("p4runtime_translation");
+        if (ann != nullptr) {
+            orig_type = false;
+            uri = ann->expr[0]->to<IR::StringLiteral>();
+            BUG_CHECK(uri, "P4runtime annotation does not have uri: %1%",
+                      type->getP4Type());
```
mbudiu-vmw: Why call getP4Type here and not pass just type?

control-plane/p4RuntimeSerializer.cpp
```diff
@@ -639,6 +640,45 @@ getMatchType(cstring matchTypeName) {
...
+/*
+ * The function returns a cstring for use as type_name for any nested type.
+*/
+static cstring
+getTypeName(const IR::Type* type, TypeMap* typeMap) {
+    cstring type_name = nullptr;
+    if (type == nullptr)
+        return type_name;
+
+    // p4runtime uses bit slices which are Type_Bits and cause
```
mbudiu-vmw: This comment indicates that there is a bug somewhere else in the code. 
How does p4runtime use bit slices?

control-plane/p4RuntimeSerializer.cpp
```diff
@@ -639,6 +640,45 @@ getMatchType(cstring matchTypeName) {
...
+    // a BUG in typeMap.cpp if getTypeType() is used on such a
+    // Type_Bits.
+    if (type->is<IR::Type_Bits>())
+        return type_name;
+
+    auto t = typeMap->getTypeType(type, true);
+    if (t->is<IR::Type_Newtype>()) {
```
mbudiu-vmw: This whole function can be simplified to 'if (auto n = t->to<IR::Type_Newtype>()) return n->name; else return null;`

control-plane/p4RuntimeSerializer.cpp
```diff
@@ -1317,6 +1383,22 @@ class P4RuntimeEntriesConverter {
             return stringRepr(k->to<IR::Constant>(), keyWidth);
         } else if (k->is<IR::BoolLiteral>()) {
             return stringRepr(k->to<IR::BoolLiteral>(), keyWidth);
+        } else if (k->is<IR::Member>()) {
```
mbudiu-vmw: We already have a class to do this work, it is called SerEnumInstance.

control-plane/typeSpecConverter.cpp
```diff
@@ -111,6 +115,48 @@ bool TypeSpecConverter::preorder(const IR::Type_Name* type) {
...
+        }
+
+        auto name = std::string(type->controlPlaneName());
+        auto types = p4RtTypeInfo->mutable_new_types();
+        if (types->find(name) == types->end()) {
+            auto newTypeSpec = new p4configv1::P4NewTypeSpec();
+            auto newType = type->type;
+            auto n = newType->to<IR::Type_Name>();
+            visit(n);
```
mbudiu-vmw: I don't understand this, but it looks wrong. We very seldom need to recursively call a visitor.
Why is n not nullptr?
I expect this will fail if the same Type_Name appears multiple times in the program.

frontends/p4/typeMap.cpp
```diff
@@ -304,6 +304,8 @@ const IR::Type* TypeMap::getCanonical(const IR::Type* type) {
 }
 
 int TypeMap::minWidthBits(const IR::Type* type, const IR::Node* errorPosition) {
+    if (type == nullptr)
```
mbudiu-vmw: type should never be nullptr. You can replace this with CHECK_NULL(type);

frontends/p4/typeMap.cpp
```diff
@@ -333,7 +335,10 @@ int TypeMap::minWidthBits(const IR::Type* type, const IR::Node* errorPosition) {
         return 0;
     }
 
-    ::error("%1%: width not well-defined", errorPosition);
+    if (errorPosition)
```
mbudiu-vmw: errorPosition should never be nullptr, so you can replace this with CHECK_NULL(errorPosition);

control-plane/p4RuntimeSerializer.cpp
hesingh: > If the type is not expected to be in the typeMap maybe we should not use typeMap->getMinWidth anyway. The function to compute the width should be very simple for the types supported by the control-plane.

This is precisely what I have done in this PR at this line of code and avoided calling getMinWidth.

https://github.com/p4lang/p4c/blob/05e7aba7c6f597e2ccc9303db8af7ad5dc5b58cd/control-plane/p4RuntimeSerializer.cpp#L649

frontends/p4/typeMap.cpp
hesingh: My first use of midWidthBits was in my getTypeWidth() function where I thought the errorPosition was not avaiable.  So I sent a nullptr to the minWidthbits for errorPosition and crashed the minWidthbits code.   No matter what, the code should not crash.  Later, I realized I could use type->getNode().  I will add a CHECK_NULL.

control-plane/p4RuntimeSerializer.cpp
```diff
@@ -639,6 +640,45 @@ getMatchType(cstring matchTypeName) {
...
+    return typeMap->minWidthBits(type, type->getNode());
+}
+
+/*
+ * The function returns a cstring for use as type_name for any nested type.
+*/
+static cstring
+getTypeName(const IR::Type* type, TypeMap* typeMap) {
+    cstring type_name = nullptr;
+    if (type == nullptr)
```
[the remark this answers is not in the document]
hesingh: Defensive programming.  Odds are not likely.  I changed to CHECK_NULL

control-plane/typeSpecConverter.cpp
hesingh: There is `type` in a model file such as p4c/p4include/psa.p4.  This method is able to process only `type` in psa.p4.  If a user writes a P4 program and uses `type`, this is what I call as user-defined NewType.  A user-defined NewType is not handled by this method.

control-plane/p4RuntimeSerializer.cpp
hesingh: As a MatchField.

control-plane/typeSpecConverter.cpp
hesingh: > I don't understand this, but it looks wrong. We very seldom need to recursively call a visitor.
> Why is n not nullptr?

First, `TypeSpecConverter::preorder(const IR::Type_Name* type)` is called.  This function has processed the `type->path` in refMap to get the `decl`.  The decl type is NewType who type is a Type_Name.  The same Type_Name is `n` in this code and is not a nullptr.

There is no recursion. A Newtype is found and its Type_Name is sent to `visit` to emplace in the map.  Then typeSpec is obtained from map.at(n).  The same two-step process is used by many preorder methods in this file. 

> I expect this will fail if the same Type_Name appears multiple times in the program.

This code is processing psa.p4 which has PortId_t uses multiple times in the program with no problems found.

control-plane/p4RuntimeSerializer.cpp
hesingh: I changed the code to use SerEnumInstance.  It needs TypeMap, so I had to change more p4runtime serializer code.  

Incidentally, I checked in code changes last night, but this PR does not show any code segment commented upon as "Outdated".

control-plane/p4RuntimeSerializer.cpp
hesingh: > The function getTypeType is being called with a type that has not been inserted in the typeMap. Typechecking inserts all types that appear in the IR in the typemap, so some other pass has created this type node without inserting it into the typeMap. The way this is debugged is by setting a conditional breakpoint on the creation of the id of the IR::Node, which is unique.

There is nothing to debug here and no other pass is involved.  In p4runtime Serializer,  the function in getMatchFields calls getTypeName and getTypeWidth for a table's MatchFields.  When the MatchField is a bit Slice, the 'matchFieldType` variable in the function is set to Type_Bits.  This is why, as discussed in another comment, getTypeWidth does not call minWidthBits and returns tb->width_bits().  In the same way, getTypeName has been changed so that if incoming type is Type_Bits, bail out because (a) the subsequent getTypeType will fail with a BUG, and (b) if type is Type_Bits, a type_name is nullptr for such a type - this behavior is by design.  In summary, the code change is fine.

hesingh:
I have no more code changes to make and have responded to all review comments.  Travis has passed.

control-plane/p4RuntimeSerializer.cpp
antoninbas: I agree with Mihai (at least with his first comment). There *is* something to debug, unless you have a good explanation for why it is not in the typeMap. So far all I see from you is a description of the problem, not the explanation.
I'm not saying it is a problem with your code, but I believe that if something is not in the typeMap, there is an issue somewhere. We do run the type checker in p4RuntimeSerializer (https://github.com/p4lang/p4c/blob/master/control-plane/p4RuntimeSerializer.cpp#L1532) after a few minor passes, so it seems that the typeMap should be up-to-date.

hesingh: If the Slice is not in typeMap, why didn't this line of code get invoked to cause an error?

https://github.com/p4lang/p4c/blob/master/control-plane/p4RuntimeSerializer.cpp#L662

hesingh: > The function getTypeType is being called with a type that has not been inserted in the typeMap. Typechecking inserts all types that appear in the IR in the typemap, so some other pass has created this type node without inserting it into the typeMap. The way this is debugged is by setting a conditional breakpoint on the creation of the id of the IR::Node, which is unique.

@mbudiu-vmw Could you please give an example for how to set the breakpoint in gdb.  I don't know the id for IR::Node and what is the creation function.

hesingh: @mbudiu-vmw All bit-field MatchFields in p4c/test/getest/p4runtime.cpp that succeed with getTypeType() have this IR as being Type_Type.  I obtained this information by debugging getType() in typeMap.cpp.

```
[1214926] Type_Type
  type: [6419] Type_Bits size=16 isSigned=0
```

The failing bit Slice MatchField has this IR

```
[959116] Type_Bits size=10 isSigned=0
```

This is why getTypeType() fails for Slice.  I think, in typeChecker or elsewhere where a table key is being added to typeMap, the Slice has to be marked as Type_Type for top-level type.   Please see this line of p4runtime.cpp above which several bit-fields exist which all have Type_Type.  

https://github.com/p4lang/p4c/blob/master/test/gtest/p4runtime.cpp#L428

If the Slice IR cannot be set to Type_Type, then we do have root cause and have to avoid the issue in p4runtimeSerializer code.

mbudiu-vmw:
Can you file an issue with a test case for which getMinWidth fails with a crash?

hesingh:
> Can you file an issue with a test case for which getMinWidth fails with a crash?

Hmm, when you checked in getMinWidth PR, `make check` passed, right?  Then gtestp4c, which is the last test run by `make check`, also passed.  So, if I file a new issue, how will you reproduce the issue?  If I am correct thus far, then it's better for me to checkin this PR with workaround and then file the issue.

The gtestp4c test invokes p4runtime.cpp which uses bit Slices for table key MatchField.  I am not sure about the program being rewritten. Soon as the bit Slices MatchFields are removed from p4runtime.cpp, gtestp4c test passes.   Otherwise when getTypeWidth() is called, it calls minWidthBits who calls getTypeType() which called getType() and ends up into the BUG inside getType() for Sliced key.

control-plane/p4RuntimeSerializer.cpp
hesingh: I don't see any other operations on bit slices in p4runtime.cpp.  The bit slice is merely used as a table key MatchField.

hesingh:
I can file a new issue because if the following line of code is changed, you can reproduce the bit slices issue of gtestp4c failing.

https://github.com/p4lang/p4c/blob/master/control-plane/p4RuntimeSerializer.cpp#L666

Change the above line to use `typeMap->minWidthBits(matchFieldType, matchFieldType->getNode());`

hesingh:
The issue has been filed.

https://github.com/p4lang/p4c/issues/1806

control-plane/p4RuntimeSerializer.cpp
hesingh: > According to your own comments there is something to debug: a program for which a type is not present in the typeMap.

I did debug as far as I could - I don't know the unique ID for IR::Node to debug its constructor.  I also debugged postorder(IR::Slice*) in typeChecker and didn't see anything obviously awry.  For good cases, getType() returns Type_Type as result.  I have added a log in another note below.

control-plane/typeSpecConverter.cpp
hesingh: First, the Type_Name is terminal. It has no Type_Bits to traverse to.  As I said above, the visit(n) emplaces the Type_Name in the map (defined in typeSpecConverter.h).

Second, I deliberately set n to nullptr before visit(n) and visit does not have a problem.

ChrisDodd: Visit should handle nullptrs (and anything else) just fine.  You need to be a bit careful calling visit in a Modifier or Transform like this, but for an Inspector, the only thing you need to worry about is triggering a recursive traversal (which will trigger an "IR loop detected" BUG if it happens.)

control-plane/p4RuntimeSerializer.cpp
antoninbas: One thought: if you try generating P4Info for a P4_16 program with a slice expression in a match key, do you observe the same issue? If not, maybe it is an issue with how the GTest test case is crafted and we can look into it.

hesingh: Good thought, Antonin.  I prepared a small P4 program with a table key as Slice expression and see the same failure as Gtest.  I will add the test program to the issue I filed last night.  thanks.

```
hemant@ubuntu:/tmp/p4c/build$ ./p4test --std p4-16 ../testdata/p4_16_samples/foo.p4 --p4runtime-files tmp.txt
  what():  In file: /tmp/p4c/frontends/p4/typeMap.cpp:84
Compiler Bug: Could not find type for <Type_Bits>(1161) bit<10>

hemant@ubuntu:/tmp/p4c/build$ 
```

hesingh:
It's been two days since 1806 was filed.  It may take longer to fix 1806.  It would be good to checkin this PR with the workaround.

hesingh:
Having merged with Mihai's commit for fixing missing Slice in typeMap, the code is ready for review - thanks, Antonin & Mihai.

control-plane/p4RuntimeSerializer.cpp
```diff
@@ -630,6 +632,33 @@ getMatchType(cstring matchTypeName) {
...
+    auto ann = type->getAnnotation("p4runtime_translation");
+    if (ann != nullptr) {
+        auto sdnB = ann->expr[1]->to<IR::Constant>();
+        return static_cast<int>(sdnB->value.get_si());
+    }
+    return typeMap->minWidthBits(type, type->getNode());
+}
+
+/*
+ * The function returns a cstring for use as type_name for any daisy-chained
```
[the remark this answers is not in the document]
hesingh: Please see updated diffs - thanks.

hesingh:
@antoninbas: Mihai has approved.  When you get a chance, please look at latest diffs - thanks.  If all is well, if one of you could please commit this code, I'd appreciate it.

control-plane/p4RuntimeSerializer.cpp
```diff
@@ -1305,13 +1355,13 @@ class P4RuntimeEntriesConverter {
             auto matchType = getKeyMatchType(tableKey, refMap);
 
             if (matchType == P4CoreLibrary::instance.exactMatch.name) {
-                addExact(protoEntry, fieldId++, k, keyWidth);
+			  addExact(protoEntry, fieldId++, k, keyWidth, typeMap);
```
antoninbas: what happened to the indentation here? it seems you added tabs

control-plane/typeSpecConverter.cpp
```diff
@@ -111,6 +115,50 @@ bool TypeSpecConverter::preorder(const IR::Type_Name* type) {
     return false;
 }
 
+// This function is invoked if an architecure's model .p4 file has a Newtype.
```
antoninbas: That doesn't make any sense to me, your comment should at least explain why this is the case
And if this is a bug (looks like it), add a TODO

control-plane/typeSpecConverter.cpp
```diff
@@ -111,6 +115,50 @@ bool TypeSpecConverter::preorder(const IR::Type_Name* type) {
...
+    if (p4RtTypeInfo) {
+        bool orig_type = true;
+        const IR::StringLiteral* uri;
+        const IR::Constant* sdnB;
+        auto ann = type->getAnnotation("p4runtime_translation");
+        if (ann != nullptr) {
+            orig_type = false;
+            uri = ann->expr[0]->to<IR::StringLiteral>();
+            if (!uri)
+                ::error("P4runtime annotation does not have uri: %1%",
```
antoninbas: This code is not written very well IMO. In case of an error here, `uri` will be set to `nullptr` and the code will segfault later on when you do `dataType->set_uri(std::string(uri->value));`. Same for `sdnB`.

test/gtest/p4runtime.cpp
```diff
@@ -519,6 +522,7 @@ TEST_F(P4Runtime, P4_16_MatchFields) {
         { 36, "lShift", 16, MatchField::TERNARY },
         { 37, "plusSix", 16, MatchField::EXACT },
         { 38, "plusSix", 16, MatchField::TERNARY },
+        { 39, "h.hUnion.c.yetAnotherHeaderField", 8, MatchField::EXACT },
```
antoninbas: That is hardly a sufficient test. You should check that the `type_name` field is set properly and that there is a corresponding entry in the `new_types` map in P4Info.

control-plane/typeSpecConverter.cpp
hesingh: Sorry, I thought `::error` would exit the running program.  I will change the code.

control-plane/typeSpecConverter.cpp
hesingh: I didn't get the time to look into this one.  I added a TODO.

control-plane/p4RuntimeSerializer.cpp
```diff
@@ -630,6 +632,30 @@ getMatchType(cstring matchTypeName) {
     }
 }
 
+static int
+getTypeWidth(const IR::Type* type, TypeMap* typeMap) {
+    auto ann = type->getAnnotation("p4runtime_translation");
+    if (ann != nullptr) {
+        auto sdnB = ann->expr[1]->to<IR::Constant>();
```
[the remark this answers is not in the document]
hesingh: Please see new changes that handle both issues.

test/gtest/p4runtime.cpp
hesingh: This will take sometime to write this code since I do not have documentation on this code.  I can add a TODO and complete the code in a future PR.  I have addressed rest of the comments - please take a look - thanks.

hesingh: Additionally, there are numerous .p4 tests added in this PR where `type` has been used and new_types exist for the `type` in p4info.

control-plane/p4RuntimeSerializer.cpp
```diff
@@ -818,10 +855,14 @@ class P4RuntimeAnalyzer {
             addAnnotations(metadata, headerField->to<IR::IAnnotated>());
 
             auto fieldType = typeMap->getType(headerField, true);
-            BUG_CHECK(fieldType->is<IR::Type_Bits>(),
-                      "Header field %1% has a type which is not bit<> or int<>",
+            BUG_CHECK((fieldType->is<IR::Type_Bits>() ||
```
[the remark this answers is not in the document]
jafingerhut: I think this issue might be relevant to this comment, but maybe not: https://github.com/p4lang/p4runtime/issues/192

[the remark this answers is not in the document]
hesingh: If P4Runtime 1.0 has a restriction for serializable enum, we may allow this code change.  Let a user not use a P4 program with a serializable enum in a header or an Action parameter. When P4Runtime 1.x is developed for p4lang/PI, p4c p4runtime serializer is already ready to support P4Runtime 1.x.

hesingh:
Note to reviewers: I'm done making changes.  Travis has passed.

control-plane/p4RuntimeSerializer.cpp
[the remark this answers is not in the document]
hesingh: I changed code to use p4c function to check if sndB fits integer.

hesingh:
Please review updated diffs - thanks.

hesingh:
@antoninbas I have developed and tested the code to check if the `value` fits into an integer.  I don't plan to add any more code changes to this PR.   Please review the latest diffs - thanks.

hesingh:
@antoninbas Please confirm if this PR review is not blocked due to the fact, that I am not changing p4runtime.cpp any more.  This PR has several P4 test programs that test `type_name`  appears in `new_types`.  

Then, the only issue left is to review new code to test if the `value` fits in an integer.  I have tested the code using a value of 0xffffffff.

control-plane/p4RuntimeSerializer.cpp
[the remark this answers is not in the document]
hesingh: A test for larger number than allowed for MatchField is an error test.  However, any P4 program in testdata/p4_16_errors/ is excluded for running with "--p4runtime-files" option.  Thus, I added a test to p4runtime.cpp and checked in code.

test/gtest/p4runtime.cpp
hesingh: > That is hardly a sufficient test. You should check that the `type_name` field is set properly and that there is a corresponding entry in the `new_types` map in P4Info.

Please see the latest commit to this PR.  How did you intend to set `type_name` in the  P4_16_MatchFields test P4 code?  All one can do is define the `type` and use the type as the table MatchField.  Thus, in the `expected` MatchFields it seems contrived to set null strings and one string for type_name.

mbudiu-vmw:
Is this ready to be merged or are we waiting for another review?
I notice there's a conflict.

hesingh:
I have resolved the conflict but let's wait for Travis to verify the conflict resolution.   I am waiting for @antoninbas to review the most-recent changes for 12 days.  I think, it may be good to merge this change and I can deal with any other comments in a subsequent PR.

hesingh:
@mbudiu-vmw I resolved the conflict with master and Travis testing has passed.

hesingh:
@mbudiu-vmw If @antoninbas has comments by end of today, please merge this code - thanks.  Thanks to you both for code reviews in many iterations.

mbudiu-vmw:
Can we merge this?

hesingh:
I would appreciate it if this PR is merged - thanks.

hesingh:
@antoninbas I have even added type_name to p4runtime.cpp via this commit.

https://github.com/p4lang/p4c/pull/1792/commits/a1c777d01bec811fc2e86c1c23c5f76a93326ba2

As I also said before, correct type_name is also generated by several P4 test programs added by this PR.  

Please see if you can approve this PR or provide any more comments - thanks.

hesingh:
Thanks Mihai!  Also thanks go to Andy and Antonin as well.

PR #1794, opened 2019-03-16
Fix to issue 1788

hesingh:
The parameterIndex has to be incremented.


hesingh:
@antoninbas Please review this one-line PR change.

hesingh:
> LGTM
> I don't think we are really testing entry generation as part of CI, so maybe we should try adding "--p4runtime-entries-files" to all frontend tests.

One file the "--p4runtime-entries-files" option generates is "p4info.txt".  Mihai did change all p4c tests to generate p4info via the commit below.  If there is more to test by adding the option to all frontend tests, I am all for it.  We can file a new Issue if folks agree - thanks.

https://github.com/p4lang/p4c/commit/62b8d528545fd0f1f42a4989172e72731e1481ac

antoninbas:
You are confusion --p4runtime-files and --p4runtime-entries-files

PR #1795, opened 2019-03-16
Add psa specific metadata to bmv2 psa_switch backend

alexwong:
Changelog:
Add psa specific metadata to the psa_switch backend in accordance with https://github.com/jafingerhut/p4-guide/blob/master/psa-examples/README-psa-implementation.md.

Test:
All tests passed in travis on fork

### DIFF
--- a/backends/bmv2/psa_switch/psaSwitch.cpp
+++ b/backends/bmv2/psa_switch/psaSwitch.cpp
@@ -372,19 +372,14 @@ void InspectPsaProgram::addTypesAndInstances(const IR::Type_StructLike* type, bo
     }
 }
 
-bool InspectPsaProgram::isStandardMetadata(const IR::Type_StructLike* st) {
-    cstring stName = st->toString();
-    if (!strcmp(stName, "struct psa_ingress_parser_input_metadata_t") ||
-        !strcmp(stName, "struct psa_egress_parser_input_metadata_t") ||
-        !strcmp(stName, "struct psa_ingress_input_metadata_t") ||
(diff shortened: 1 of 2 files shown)


jafingerhut:
I do not have any comments on the C++ code changes themselves, as I am not terribly familiar with the code there.

I have run p4c-bm2-psa with these changes on psa-example-drop-all.p4, and inspected the BMv2 JSON file produced, and it does not have what looks like the correct contents for the 7 PSA header types and headers.  Nice.

It does not yet compile psa-unicast-or-drop.p4 without error, but I expect that this change represents progress, and it is OK if later changes get that one compiling correctly.

alexwong:
> I have run p4c-bm2-psa with these changes on psa-example-drop-all.p4, and inspected the BMv2 JSON file produced, and it does not have what looks like the correct contents for the 7 PSA header types and headers. Nice.

Just want to make sure that this was a typo and it does print the 7 PSA header types and headers. I just compiled above and it looks to do so on my local build.

> It does not yet compile psa-unicast-or-drop.p4 without error, but I expect that this change represents progress, and it is OK if later changes get that one compiling correctly.

Ah, I just tried to compile this as well and it looks like there is a compile error. Because my next goals involve getting some of the things in this program compiling correctly (mainly assign and other statements referencing the new header), I agree that later commits will probably address this.

jafingerhut:
That "not" in my first paragraph was definitely a typo.  Sorry for the confusion.

jafingerhut:
Given Calin's approval, and since this only affects PSA back end of compiler, I am merging this in.

PR #1797, opened 2019-03-17
[testing -- DO NOT MERGE] Extend generic control plane name generation to externs

amreshk:


jafingerhut:
Out of curiosity, do these changes enable P4Info generation for any P4 programs that was not possible without these changes?  If so, a test P4 program with expected P4Info output would be nice (not necessary, as far as I am concerned, but nice).

antoninbas:
@jafingerhut I don't think I can come up with an example written for v1model or PSA.
This is for an architecture that instantiates multiple pipeline objects under a single switch instance:
```
package Switch<IH0, IM0, EH0, EM0, IH1, IM1, EH1, EM1(
    Pipeline<IH0, IM0, EH0, EM0> pipe0,
    Pipeline<IH1, IM1, EH1, EM1> pipe1);
```
where `Pipeline` is itself a `package` with ingress parser, ingress control, ingress deparser, ...
For this architecture, `pipe0` and `pipe1` can be exactly the same (use exactly the same parser definition, control definitions, etc.), they can be completely different, or something in between.
When they share a top-level control definition, we'd like the tables inside this control to have different names for `pipe0` and `pipe1`, in order to program them differently at runtime (since they are different physical entities). One could argue that the names should be the same, and that the runtime APIs could take an additional pipe identifier, to enable the application to program pipes selectively. However, we have the following issues:
 * P4Runtime currently doesn't have the notion of pipe, or anything like it (so we would have to "abuse" the device id for this)
 * nothing prevents the P4 programmer from using different compile-time constructor arguments for both instances of the top-level control (for `pipe0` and `pipe1`), in which case it doesn't make much sense to use the same name anymore, since the 2 control instances become fundamentally different

With Amresh's changes, we have the possibility of prepending the pipe name to each fully-qualified name (`pipe0.MyIngress.t0`, `pipe1.MyIngress.t1`) which solves the issue. I can think of one tricky case where that wouldn't work (if the same control definition can be used for both ingress & egress within a pipe, but usually this is not the case because the architecture requires different "signatures" for ingress and egress controls).

One could argue that our approach is wrong and that these pipes should be considered 2 different devices, but we are still talking about a single switching chip, so I fear it would create more confusion.

control-plane/p4RuntimeSerializer.cpp
```diff
@@ -925,13 +925,13 @@ class P4RuntimeAnalyzer {
...
 
-    void addValueSet(const IR::P4ValueSet* inst) {
+    void addValueSet(const IR::P4ValueSet* inst, const cstring name) {
         // guaranteed by caller
         CHECK_NULL(inst);
 
         auto vs = p4Info->add_value_sets();
 
-        auto name = inst->controlPlaneName();
+        // auto name = inst->controlPlaneName();
```
antoninbas: ?

control-plane/p4RuntimeSerializer.cpp
```diff
@@ -1156,29 +1156,34 @@ static void collectTableSymbols(P4RuntimeSymbolTable& symbols,
...
     auto parser = parserBlock->container;
     CHECK_NULL(parser);
 
+    auto name = archHandler->getControlPlaneName(parserBlock);
+
     for (auto s : parser->parserLocals) {
         if (auto inst = s->to<IR::P4ValueSet>()) {
-            symbols.add(P4RuntimeSymbolType::VALUE_SET(), inst);
+            // symbols.add(P4RuntimeSymbolType::VALUE_SET(), inst);
+            symbols.add(P4RuntimeSymbolType::VALUE_SET(), name);
```
antoninbas: that doesn't look right: you are using the name of the parser here, not the name of the PVS

control-plane/p4RuntimeSerializer.cpp
```diff
@@ -1156,29 +1156,34 @@ static void collectTableSymbols(P4RuntimeSymbolTable& symbols,
...
     CHECK_NULL(parserBlock);
 
+    auto name = archHandler->getControlPlaneName(parserBlock);
     auto parser = parserBlock->container;
     CHECK_NULL(parser);
 
     for (auto s : parser->parserLocals) {
         if (auto inst = s->to<IR::P4ValueSet>()) {
-            analyzer.addValueSet(inst);
+            analyzer.addValueSet(inst, name);
```
antoninbas: same, I think you need to fix how you handle PVS here

amreshk:
@antoninbas updated pvs name generation to prefix control plane name for parser with pvs name, please take a look

antoninbas (approved):
LGTM, looks like there is no change in any of the checked-in P4Info files

antoninbas (approved):
Don't know why I need to approve again, but here you go

control-plane/p4RuntimeArchHandler.cpp
```diff
@@ -39,16 +39,20 @@ namespace ControlPlaneAPI {
...
                               bool *isConstructedInPlace) {
+    const IR::P4Table* table = tableBlock->container;
+    CHECK_NULL(table);
+    auto tableName = archHandler->getControlPlaneName(tableBlock);
     auto property = table->properties->getProperty(propertyName);
     if (property == nullptr) return boost::none;
     if (!property->value->is<IR::ExpressionValue>()) {
         ::error("Expected %1% property value for table %2% to be an expression: %3%",
-                propertyName, table->controlPlaneName(), property);
+                propertyName, tableName, property);
```
mbudiu-vmw: since you are touching all these "error" calls you may consider migrating them to our new error function prototype which has an extra argument indicating the error kind.

mbudiu-vmw:
Can you add a test case which exercises the new code paths?

mihaibudiu (changes requested):

Issue #1548, opened 2018-10-11
assert and assume


jafingerhut:
Proposal for adding 'assert' and 'assume' statements to P4.

These could probably be introduced into the P4_16 language as extern
functions.  I would guess that their implementation in p4c and BMv2
simple_switch would be fairly straightforward, and an experimental
implementation there could precede adding them to the language spec
(if adding them to the language spec is even desirable).

Assert statements are part of the standard library of many programming
languages.  During testing of a P4 program, it can be useful to have
statements like this:

    assert(<boolean expression>)

The default behavior would be that such statements are no-ops, and for
optimization purposes, that they are removed early during compilation
so as not to hinder andy optimizations.

When assert statements are enabled in a software switch like BMv2
simple_switch, and the expression evaluates to false, debugging
information and a distinctive error message should be printed that can
be searched for by test case post-processing code to cause test cases
to fail, if desired.

It should be easy for the developer to get debugging information, such
as:

+ file name
+ line number
+ text of condition that failed
+ The values of the header fields or variables involved in the
  expression, at least, and perhaps _all_ header and metadata values
  for the packet, too.  Ideally the developer would probably also like
  the entire path of execution up to that point, similar to what
  simple_switch outputs when the `--log-console` command line option
  is given.
+ Optionally, at run time drop into an interactive debugger session
  that lets the developer examine the values of more things.

In an FPGA or ASIC, if the compiler supported enabling assert
statements, they could be implemented as if replaced by this code:
```
    if (! (<boolean expression>)) {
        mark_to_drop();   // for v1model, or the equivalent for PSA.

	// Probably also a good idea to change any state that would
	// cause the packet to be resubmitted, recirculated, cloned,
	// generate_digest "off" for this packet.

        // Record as much information in an implementation-specific
	// manner that allows the software to derive as much of the
	// above debugging information as is practical.
        exit;
    }
```
A formal verification tool could use assert statements, too.  For
example, a tool like p4pktgen [1] could attempt to find example input
packets and table entry contents that would cause the assert condition
to evaluate to false.

[1] https://github.com/p4pktgen/p4pktgen

The statement:

    assume(<boolean expression>)

looks similar to assert, but has a fairly different meaning in the
context of a formal verification tool.  Whereas an assert statement
gives a condition that you _hope_ is true at that point of execution
of the program, and you want to know if it is ever false,
assume(condition) means that you as a developer somehow know or
strongly believe that this condition _will always be true_.

[...]

PR #1799, opened 2019-03-18
Added support for assert and assume statement (#1548)

enisah:
* Added removeAssertAssume pass in frontend which will remove
 all assert and assume statements from program when BUILD_TYPE=RELEASE

* Not supported in deparsers and verifyChecksum block with assumption
that it is not needed

* Added support for code gen for built-in methods in ebpf backend

* Changed generating error code value in bmv2 backend to be
generated like hex value into json file

Signed-off-by: Enisa Hadzic <enisa.hadzic@rt-rk.com>


jafingerhut:
@liujed @jnfoster Pinging you folks on this, as I know you are interested in this topic, and may be working on an implementation of these ideas that are hopefully in alignment with this?  Or does this PR represent independent implementation work?

backends/bmv2/common/expression.cpp
```diff
@@ -251,7 +251,9 @@ void ExpressionConverter::postorder(const IR::Member* expression)  {
         result->emplace("type", "hexstr");
         auto decl = type->to<IR::Type_Error>()->getDeclByName(expression->member.name);
         auto errorValue = structure->errorCodesMap.at(decl);
-        result->emplace("value", Util::toString(errorValue));
+        // this generates error constant like hex value
+        auto reprValue = stringRepr(errorValue);
```
mbudiu-vmw: if this is a bug fix you should probably file a separate PR for it, including a test case.

backends/bmv2/common/parser.cpp
```diff
@@ -187,6 +187,22 @@ Util::IJson* ParserConverter::convertParserStatement(const IR::StatOrDecl* stat)
                     params->append(jexpr);
                 }
                 return result;
+            } else if (extfn->method->name.name == "assert") {
```
mbudiu-vmw: the code is almost the same, can you unify it?

backends/bmv2/psa_switch/psaSwitch.cpp
```diff
@@ -997,4 +999,42 @@ void ExternConverter_Digest::convertExternInstance(
     UNUSED const IR::ExternBlock* eb, UNUSED const bool& emitExterns)
 { /* TODO */ }
 
+Util::IJson* ExternConverter_assert::convertExternFunction(
```
mbudiu-vmw: lots of code duplicated here as well. Can you refactor?

backends/bmv2/simple_switch/simpleSwitch.cpp
```diff
@@ -107,6 +107,46 @@ ExternConverter_direct_counter ExternConverter_direct_counter::singleton;
 ExternConverter_direct_meter ExternConverter_direct_meter::singleton;
 ExternConverter_action_profile ExternConverter_action_profile::singleton;
 ExternConverter_action_selector ExternConverter_action_selector::singleton;
+ExternConverter_assert ExternConverter_assert::singleton;
+ExternConverter_assume ExternConverter_assume::singleton;
+
+Util::IJson* ExternConverter_assert::convertExternFunction(
```
mbudiu-vmw: and again. Could this code go to common?

backends/ebpf/ebpfControl.cpp
```diff
@@ -47,21 +47,65 @@ bool ControlBodyTranslator::preorder(const IR::PathExpression* expression) {
...
 
 void ControlBodyTranslator::processFunction(const P4::ExternFunction* function) {
-    ::error("%1%: Not supported", function->method);
+    if (function->method->name.name == "assert") {
+        BUG_CHECK(function->expr->arguments->size() == 1,
+                "%1%: Expected one argument in ", function);
+        builder->emitIndent();
+        builder->append("if (");
+        visit(function->expr->arguments->at(0)->expression);
+        builder->append(" == false) ");
```
mbudiu-vmw: This should be rather !expr

backends/ebpf/ebpfControl.cpp
```diff
@@ -47,21 +47,65 @@ bool ControlBodyTranslator::preorder(const IR::PathExpression* expression) {
...
+        builder->blockStart();
+        builder->emitIndent();
+        builder->appendFormat("% s = % s;",
```
mbudiu-vmw: You chose a particular semantics for assert which is not the standard C one.
You should probably document this in a summary of the PR.
This should also go to the language design working group.

backends/ebpf/ebpfControl.cpp
```diff
@@ -47,21 +47,65 @@ bool ControlBodyTranslator::preorder(const IR::PathExpression* expression) {
...
+        builder->appendFormat(
+            "fprintf(stderr, \"Assertion error: ' % s ' failed, file ' % s ', line % s \\n\");",
+            sourceFragment.c_str(), sourceFile.c_str(), linePosition.c_str());
+        builder->newline();
+        builder->emitIndent();
+        builder->appendFormat("goto %s;", IR::ParserState::reject.c_str());
+        builder->newline();
+        builder->blockEnd(true);
+        return;
+    } else if (function->method->name.name == "assume") {
```
mbudiu-vmw: this code also has a lot of duplication.
what is the difference between assert and assume really?

backends/ebpf/ebpfParser.cpp
```diff
@@ -335,8 +326,79 @@ bool StateTranslationVisitor::preorder(const IR::MethodCallExpression* expressio
             BUG("Unhandled packet method %1%", expression->method);
             return false;
         }
+    } else if (mi->is<P4::ExternFunction>()) {
```
mbudiu-vmw: duplicated code

backends/ebpf/ebpfParser.cpp
```diff
@@ -335,8 +326,79 @@ bool StateTranslationVisitor::preorder(const IR::MethodCallExpression* expressio
...
+                sourceFragment.c_str(), sourceFile.c_str(), linePosition.c_str());
+            builder->newline();
+            builder->emitIndent();
+            builder->appendFormat("goto %s;", IR::ParserState::reject.c_str());
+            builder->newline();
+            builder->blockEnd(true);
+            return false;
+        }
+    } else if (mi->is<P4::BuiltInMethod>()) {
+    // Added this because compiler didn't recognize assert(hdr.ethernet.isValid()), for expample
```
mbudiu-vmw: if this is a bug maybe you can file a separate PR and a test case for it.
When does the compiler not recognize isValid, only within assert? 
isValid should be handled by the expressionConverter.

backends/ebpf/ebpfTable.cpp
```diff
@@ -54,6 +54,67 @@ class ActionTranslationVisitor : public CodeGenInspector {
         return false;
     }
 
+    bool preorder(const IR::MethodCallExpression* expression) {
+        auto mi = P4::MethodInstance::resolve(expression, refMap, typeMap);
+        if (mi->is<P4::ExternFunction>()) {
```
mbudiu-vmw: this code should be refactored and should appear only once.

backends/ebpf/target.cpp
```diff
@@ -21,7 +21,8 @@ namespace EBPF {
 
 
 void KernelSamplesTarget::emitIncludes(Util::SourceCodeBuilder* builder) const {
-    builder->append("#include \"ebpf_kernel.h\"\n");
+    builder->append("#include \"ebpf_kernel.h\"\n"
```
mbudiu-vmw: Is there stdio.h in the kernel?

frontends/p4/coreLibrary.h
```diff
@@ -30,6 +30,8 @@ enum class StandardExceptions {
     OverwritingHeader,
     HeaderTooShort,
     ParserTimeout,
+    AssertError,
```
mbudiu-vmw: This is a big assumption: that we are changing code.p4. 
I would rather expect that assert and assume are library functions - at least for a while.
Putting them in core.p4 may break programs that happen to use identifiers with these names.
This is a subject for the LDWG.

frontends/p4/frontend.cpp
```diff
@@ -148,6 +149,9 @@ const IR::P4Program *FrontEnd::run(const CompilerOptions &options, const IR::P4P
         new TypeInference(&refMap, &typeMap, false),  // insert casts
         new ValidateMatchAnnotations(&typeMap),
         new DefaultArguments(&refMap, &typeMap),  // add default argument values to parameters
+#ifdef NDEBUG
```
mbudiu-vmw: First of all, this NDEBUG is for the compiler source code. You should remove assert/assume only if the p4 compiler is invoked with NDEBUG, not gcc.
Second, this is too early to remove the calls. If the calls have side-effects they should still happen. So the removal should happen after the SideEffects pass.
Third, I am not sure we actually want this part of the front-end; perhaps the mid-end.

backends/ebpf/ebpfControl.cpp
```diff
@@ -47,21 +47,65 @@ bool ControlBodyTranslator::preorder(const IR::PathExpression* expression) {
...
+        builder->appendFormat("% s = % s;",
+             control->program->errorVar.c_str(), P4::P4CoreLibrary::instance.assertError);
+        builder->newline();
+        auto srcInfo = function->expr->srcInfo;
+        auto sourceFragment = srcInfo.toBriefSourceFragment();
+        auto sourceFile = srcInfo.getSourceFile();
+        auto linePosition = srcInfo.toPositionString();
+        builder->emitIndent();
+        builder->appendFormat(
+            "fprintf(stderr, \"Assertion error: ' % s ' failed, file ' % s ', line % s \\n\");",
```
mbudiu-vmw: Have you tested this with kernel code? Can you just call fprintf from an ebpf program running in the kernel?

backends/ebpf/ebpfParser.cpp
enisah: Compiler didn't recognize isValid in assert, but also it didn't recognize setValid and setInvalid method calls for that ebpf backend when they are called from a parser state.
Also, I notice that compiler doesn't recognize verify call in parser state for ebpf backend, but I didn't add code generation for that case. In that case following error is reported:
error: Unexpected method call in parser verify
 verify(false, error.NoError);
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
If I understood it well, it should be enabled to call  methods, that are mentioned above, in parser state.

backends/bmv2/common/expression.cpp
enisah: Thank you for all comments.
I think that it is a bug fix, because when I extended core.p4 errors, I had noticed  that two tests had failed: checksum3-bmv2.p4 and checksum2-bmv2.p4. These tests reported out_of_range exception when make check is done, because tests added new errors which will have values 9 and 10, but it didn't generate error values as hex values (0x9 and 0xa) in json file. For example, in case of error value 10 it was generated like:
    "type" : "hexstr",
     "value" : "10"
which will be read as hex value 16, not 10 as it should be.
Now it is generate like:
  "type" : "hexstr",
   "value" : "0xa"
I will include a test for this case and make it as a seperate PR.

backends/ebpf/ebpfControl.cpp
enisah: I was trying to test it, but I couldn't manage to do it because I got some errors when I tried to compile .c file that is p4c output. So I did it according to the rest of code gen for ebpf backend.
Maybe I didn't understand well how to test it, so I would be thankful if you can give me more information  how to do it.

backends/ebpf/ebpfControl.cpp
enisah: I thought that assert reports error when expr == false and assume reports it when expr == true. 
From comments in [behavioral_model pull request](https://github.com/p4lang/behavioral-model/pull/737/commits/9846038dfc2d9b98dc10e4e679a8b3c3efe333ea) I see now that my assumption was wrong.
As I understood from the comments, both of them should report error when expr == false, but now I have some doubts about their differences in usage. If they should have similar behavior, what is the real difference between them?

jafingerhut: The intended difference is in how formal proof tools use them, not in how behavioral-model uses them.  I have tried to explain the difference in the original issue here: https://github.com/p4lang/p4c/issues/1548  If that does not make it clear, please consider asking a question in that issue's comments (or here).

backends/ebpf/ebpfControl.cpp
mbudiu-vmw: If you get errors compiling a C file it's a bug in the ebpf back-end - please file an issue.
The ebpf testing tries to test the files in multiple ways, depending on your kernel version.
But the ebpf code is supposed to run in the kernel, so it may not use fprintf.
The more important question is how to handle an assertion failure in ebpf.

backends/ebpf/ebpfParser.cpp
mbudiu-vmw: It's possible that verify is not supported yet in the back-end. 
But isValid should work; if you have a counter-example please file another issue.

backends/ebpf/ebpfControl.cpp
mbudiu-vmw: The question remains: how should the bmv2 and ebpf back-ends implement these functions?
Are these compile-time only - should we always remove them?

enisah: I have tried to understand more from the original issue, but I still do not comprehend the differences that you mentioned. @jafingerhut  could you give me an example that will help me to better understand it, please?

mbudiu-vmw:
Usually assume gives preconditions (which are assumed to be true before you enter a piece of code), and assert enforces postconditions (which are true when you exit a piece of code). Usually a static analyzer (a prover) tries to prove the postconditions assuming that the preconditions hold. The prover cannot prove the precoditions - it will take them for granted. That's the difference between assume and assert. This behavior is true for static analyses. 

The (dynamic) runtime behavior of both statements is the same, though: if the condition is false, then something is very wrong. For assume it means that the assumption was incorrect. For asserts it would mean that the proof made by the prover was wrong.

frontends/p4/frontend.cpp
enisah: If I understood you well, p4 compiler should remove assert/assume if it is invoked with  -DNDEBUG (and default) option and leave it if it is invoked with -UNDEBUG. I moved this pass to be the first in the mid-end, before any of the optimizations is done. Feel free to correct me if I misunderstood it.

antoninbas: What Mihai means is that when you write `#ifdef NDEBUG`, you are using a preprocessor flag that is meant for the C++ compiler with which you compile p4c. The `assert` / `assume` statements should be removed if the *p4c* compiler is invoked with `-DNDEBUG`. When I compile a C / C++ program, I choose whether asserts are removed by providing `-DNEBUG` or not. Whether the C / C++ compiler was itself compiled with `-DNEBUG` is completely irrelevant. The p4c invocation `p4c -DNEBUG <path to P4 program>` should remove the `assert` / `assume` statements. I would argue that using a preprocessor flag for this is probably not ideal. What about `p4c -ndebug <path to P4 program>` instead?

enisah:
Hello everyone, I have done changes according to your suggestions. Unfortunately I didn't manage to test implementation for ebpf backend, so I made a new PR with support for bmv2 backend only. Please take a look, here is link to new PR https://github.com/p4lang/p4c/pull/1859, meanwhile I will try to resolve problem with ebpf.

PR #1802, opened 2019-03-19
Allow 'void' as a type argument.

ChrisDodd:


frontends/parsers/p4/p4parser.ypp
```diff
@@ -918,6 +919,7 @@ typeArgumentList
 
 realTypeArg
     : typeRef                     { $$ = $1; }
+    | VOID                        { $$ = IR::Type_Void::get(); }
```
mbudiu-vmw: you should send a PR to the spec.

testdata/p4_16_samples/extern3.p4
```diff
@@ -0,0 +1,28 @@
...
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+extern ext<H, V> {
+    ext(H v);
+    V method1<T>(in H h, in T t);
```
mbudiu-vmw: I suspect this only works in the return position.

PR #1804, opened 2019-03-19
Run p4test tests with --p4runtime-entries-files

### LINKED ISSUE #1800: Add --p4runtime-entries-files to frontend tests
We should consider adding this flag to https://github.com/p4lang/p4c/blob/master/backends/p4test/run-p4-sample.py, in order to test serialization of static table entries.

antoninbas:
Most of the checked in files are empty because very few programs have
static table entries.

Fixes #1800

### DIFF
--- a/testdata/p4_16_samples_outputs/bvec-hdr-bmv2.p4.entries.txt
+++ b/testdata/p4_16_samples_outputs/bvec-hdr-bmv2.p4.entries.txt
@@ -1,0 +1,46 @@
+updates {
+  type: INSERT
+  entity {
+    table_entry {
+      table_id: 33562569
+      match {
+        field_id: 1
+        exact {
+          value: "\001"
+        }
+      }
+      action {
+        action {
+          action_id: 16837978
+          params {
+            param_id: 1
+            value: "\000\001"
+          }
+        }
+      }
+    }
+  }
+}
+updates {
+  type: INSERT
+  entity {
+    table_entry {
+      table_id: 33562569
+      match {
+        field_id: 1
+        exact {
+          value: "\002"
+        }
+      }
+      action {
+        action {
+          action_id: 16837978
+          params {
+            param_id: 1
+            value: "\000\002"
+          }
+        }
+      }
+    }
+  }
+}
--- a/testdata/p4_16_samples_outputs/table-entries-exact-ternary-bmv2.p4.entries.txt
+++ b/testdata/p4_16_samples_outputs/table-entries-exact-ternary-bmv2.p4.entries.txt
@@ -1,0 +1,124 @@
+updates {
+  type: INSERT
+  entity {
+    table_entry {
+      table_id: 33616996
+      match {
+        field_id: 1
+        exact {
+          value: "\001"
+        }
+      }
+      match {
+        field_id: 2
+        ternary {
+          value: "\021\021"
+          mask: "\000\017"
+        }
+      }
+      action {
+        action {
+          action_id: 16837978
+          params {
+            param_id: 1
+            value: "\000\001"
+          }
+        }
+      }
+      priority: 1
+    }
+  }
+}
+updates {
+  type: INSERT
+  entity {
+    table_entry {
+      table_id: 33616996
+      match {
+        field_id: 1
+        exact {
+          value: "\002"
+        }
+      }
+      match {
+        field_id: 2
+        ternary {
+          value: "\021\201"
+          mask: "\377\377"
+        }
+      }
+      action {
+        action {
+          action_id: 16837978
+          params {
+            param_id: 1
+            value: "\000\002"
+          }
+        }
+      }
+      priority: 2
+    }
+  }
+}
+updates {
+  type: INSERT
+  entity {
+    table_entry {
+      table_id: 33616996
+      match {
+        field_id: 1
+        exact {
+          value: "\003"
+        }
+      }
+      match {
+        field_id: 2
+        ternary {
+          value: "\021\021"
+          mask: "\360\000"
+        }
+      }
+      action {
+        action {
+          action_id: 16837978
+          params {
+            param_id: 1
+            value: "\000\003"
+          }
+        }
+      }
+      priority: 3
+    }
+  }
+}
+updates {
+  type: INSERT
+  entity {
+    table_entry {
+      table_id: 33616996
+      match {
+        field_id: 1
+        exact {
+          value: "\004"
+        }
+      }
+      match {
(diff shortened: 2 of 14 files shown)


testdata/p4_16_samples_outputs/bvec-hdr-bmv2.p4.entries.txt
```diff
@@ -0,0 +1,46 @@
...
+        exact {
+          value: "\001"
+        }
+      }
+      action {
+        action {
+          action_id: 16837978
+          params {
+            param_id: 1
+            value: "\000\001"
```
jafingerhut: This is a question for my curiosity, not necessarily a problem with the commit.  Should these values use the minimum-length encoding specified in the P4Runtime spec for unsigned integers?  They do not seem to be, but the value looks correct.

hesingh:
Should we also change the bmv2 backed python file with this PR's change to the p4test backend?

testdata/p4_16_samples_outputs/table-entries-exact-ternary-bmv2.p4.entries.txt
```diff
@@ -0,0 +1,124 @@
...
+        field_id: 1
+        exact {
+          value: "\001"
+        }
+      }
+      match {
+        field_id: 2
+        ternary {
+          value: "\021\021"
+          mask: "\000\017"
```
jafingerhut: Does BMv2 simple_switch and/or P4Runtime allow values in value/masks where (value & mask) != value ?  Should the compiler be warning/erroring about this?

testdata/p4_16_samples_outputs/table-entries-exact-ternary-bmv2.p4.entries.txt
```diff
@@ -0,0 +1,124 @@
...
+        field_id: 1
+        exact {
+          value: "\003"
+        }
+      }
+      match {
+        field_id: 2
+        ternary {
+          value: "\021\021"
+          mask: "\360\000"
```
jafingerhut: Similar (value & mask) != value case as mentioned earlier.

testdata/p4_16_samples_outputs/table-entries-exact-ternary-bmv2.p4.entries.txt
```diff
@@ -0,0 +1,124 @@
...
+        field_id: 1
+        exact {
+          value: "\004"
+        }
+      }
+      match {
+        field_id: 2
+        ternary {
+          value: "\000\000"
+          mask: "\000\000"
```
jafingerhut: Should P4Runtime canonical representation of omitting fields that are completely "don't care" apply in this file?

testdata/p4_16_samples_outputs/table-entries-lpm-bmv2.p4.entries.txt
```diff
@@ -0,0 +1,71 @@
+updates {
+  type: INSERT
+  entity {
+    table_entry {
+      table_id: 33555353
+      match {
+        field_id: 1
+        lpm {
+          value: "\021"
+          prefix_len: 4
```
jafingerhut: Similar (value & mask) != value case, but for a prefix, here.

testdata/p4_16_samples_outputs/table-entries-lpm-bmv2.p4.entries.txt
```diff
@@ -0,0 +1,71 @@
...
+}
+updates {
+  type: INSERT
+  entity {
+    table_entry {
+      table_id: 33555353
+      match {
+        field_id: 1
+        lpm {
+          value: "\000"
```
jafingerhut: Similar question to a previous one about a field that is completely "don't care", and whether it should be omitted from this file, or present like this.

testdata/p4_16_samples_outputs/table-entries-priority-bmv2.p4.entries.txt
```diff
@@ -0,0 +1,75 @@
...
+      action {
+        action {
+          action_id: 16837978
+          params {
+            param_id: 1
+            value: "\000\001"
+          }
+        }
+      }
+      priority: 1
```
jafingerhut: @priority annotation in source program appears to be ignored here.  I thought the P4_16 language spec didn't even support explicit priority here -- I thought it went with "entries always in source code order" for ternary tables.  Should this test be removed?  If not, it definitely seems wrong in priority values, at least.

testdata/p4_16_samples_outputs/table-entries-range-bmv2.p4.entries.txt
```diff
@@ -0,0 +1,75 @@
...
+updates {
+  type: INSERT
+  entity {
+    table_entry {
+      table_id: 33595117
+      match {
+        field_id: 1
+        range {
+          low: "\000"
+          high: "\377"
```
jafingerhut: Same comment here about "don't care" fields, whether they should be omitted or present in the file.

testdata/p4_16_samples_outputs/table-entries-ternary-bmv2.p4.entries.txt
```diff
@@ -0,0 +1,100 @@
+updates {
```
jafingerhut: Repeat comments on this file of things that were present in earlier ones, too: (value & mask) != mask in one entry.  Should field be present or absent in file if it is completely don't-care.

jafingerhut:
The following source files in the testdata/p4_16_samples directory have 'const entries', and are able to produce a non-empty entries file in my quick testing.  Should they also have non-empty expected output files for table entries?

action-two-params.p4
calc-ebpf.p4
init_ebpf.p4

antoninbas:
@jafingerhut I was actually missing 4 files (tests added recently that didn't get picked up when I ran ctest locally), one of them action-two-params.p4. Note that the *-ebpf programs are not written for v1model or PSA, and the compiler should not generate a P4Info message / and entries message for these programs (unless you trick the compiler by setting the architecture to something else or you omit the architecture flag?).

I think your other comments are correct. Please open a Github issue to document the list of problem you found with the generated messages, I will fix them later. The whole objective of this PR was to detect problems with the entry generation code, and looks like it was a success!

jafingerhut:
Issue created here: https://github.com/p4lang/p4c/issues/1805

Thanks for these changes, by the way, but I do sometimes get concerned that we will enshrine checking of incorrect expected output files. :-)

PR #1807, opened 2019-03-20
allow custom typechecker in SimplifyDefUse pass

hanw:


hanw:
Thanks, I will try to implement a convenience function in a follow-up PR.

Issue #1805, opened 2019-03-19
Several small bugs in generation of table entries for P4_16 'const entries'


jafingerhut:
Details of problems found are given in comments on this PR: https://github.com/p4lang/p4c/pull/1804

The main categories of problems found were:

+ When the P4_16 source file had a ternary match field with (value & mask) != mask, an entries file was generated with that value/mask pair, anyway, in violation of P4Runtime rules against such value/mask pairs.  Seems like the best thing to do would be to at least warn about such bad value/mask pairs in the source file, or even give an error during compilation.

+ P4Runtime canonical representation is to omit the description of table search key fields, if the field is ternary, lpm, or range, and is completely "don't care", i.e. can match any legal value for the field.  The current entries files generated mention such fields, although it appears with correct "match anything" rules (value=mask=0 for ternary, value=prefix_length=0 for lpm, min=0 max=2^W-1 for range).

+ P4Runtime canonical representation of unsigned bit strings is to use the fewest number of bytes in the byte sequence possible, omitting any leading bytes that are completely 0.  The existing files produce values with unnecessary leading 0 bytes in some cases.

+ The test program testdata/p4_16_samples/table-entries-priority-bmv2.p4 has a ternary table with `@priority` annotations on several of the entries.  Such annotations were considered by the P4 language design work group, but ultimately rejected (or at least postponed) for the latest P4_16 language spec (v1.1 as of this writing), in favor of always having the relative priorities of multiple table entries determined by the order they appear in the source file.  I am not sure whether `@priority` is desired as an experimental extension for the `p4c` compiler or not.  If so, they should be used and carried through into the entries file.  If not, maybe this test program should be removed?  It seems odd to give an error for an ignored annotation, but at least a warning from the compiler that they are ignored would be good, if we want to keep this test program around, else people might infer that this is supported.

jafingerhut:
One more related issue from Antonin: "I was actually missing 4 files (tests added recently that didn't get picked up when I ran ctest locally), one of them action-two-params.p4"

I agree that it seems reasonable to leave out the source file names ending with "-epbf.p4" from generating and checking entries files.

antoninbas:
@jafingerhut I don't think that last part is an actual issue. My branch was probably out-of-date and the tests got picked up after rebasing from master and running cmake again.

jafingerhut:
Sounds good.  I assumed it was left out in the commit because I say only the 1 commit in the PR.  Glad to know the other files are in there already.

[...]

PR #1808, opened 2019-03-20
Make --p4runtime-entries outputs more conformant to P4Runtime spec

antoninbas:
See #1805

 * omit don't care matches
 * toggle off value bits which are masked off for ternary and LPM
 * emit a warning for ignored @priority annotation, and generate correct
   priority values
 * I couldn't move to the canonical binary string representation for
   bit<W> values because it is not currently supported by p4lang/PI

### DIFF
--- a/control-plane/p4RuntimeSerializer.cpp
+++ b/control-plane/p4RuntimeSerializer.cpp
@@ -76,15 +76,6 @@ using Helpers::setPreamble;
 
 static const p4rt_id_t INVALID_ID = p4configv1::P4Ids::UNSPECIFIED;
 
-// TODO(antonin): Here are the known issues:
-// - We don't currently distinguish between the case where the const default
-//   action has mutable params and the case where it doesn't.
-// - Locals are intentionally not exposed, but this prevents table match keys
-//   which involve complex expressions from working, because those expressions
-//   are desugared into a match against a local. This could be fixed just by
-//   exposing locals, but first we need to decide how it makes sense to expose
-//   this information to the control plane.
-
 /// @return true if @node has an @hidden annotation.
 static bool isHidden(const IR::Node* node) {
     return node->getAnnotation("hidden") != nullptr;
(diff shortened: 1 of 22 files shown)


control-plane/p4RuntimeSerializer.cpp
```diff
@@ -1414,6 +1467,15 @@ class P4RuntimeEntriesConverter {
         auto bitsRequired = static_cast<size_t>(mpz_sizeinbase(value.get_mpz_t(), 2));
         BUG_CHECK(static_cast<size_t>(width) >= bitsRequired,
                   "Cannot represent %1% on %2% bits", value, width);
+        // TODO(antonin): P4Runtime defines the canonical representation for
+        // bit<W> value as the smallest binary string required to represent the
+        // value (no 0 padding). Unfortunately the reference P4Runtime
+        // implementation (https://github.com/p4lang/PI) does not currently
+        // support the canonical representation, so instead we use a padded
+        // binary string, which according to the P4Runtime specification is also
+        // valid (but not the canonical representation, which means to RW
```
jafingerhut: typo?  "to" -> "no"

testdata/p4_16_samples_outputs/table-entries-exact-ternary-bmv2.p4-stderr
```diff
@@ -0,0 +1,6 @@
+table-entries-exact-ternary-bmv2.p4(65): [--Wwarn=mismatch] warning: P4Runtime requires that Ternary matches have masked-off bits set to 0, updating value 4369 to conform to the P4Runtime specification
```
mbudiu-vmw: it would be nice if the error message used the same radix as the source code.

cc10512: Do all passes save the base when operating on constants?

jafingerhut: I won't argue against that if it is straightforward to do, but note that the source code excerpt with the original base is marked with '^^^^' later in the warning message, so that is good.

antoninbas: @mbudiu-vmw I made a small change to `IR::Constant::toString` to use the base, so now the values match. The base is not included in the string (e.g. abcd and not 0xabcd), because that's a bigger change.

antoninbas: I changed my mind and decided it was also a bit confusing without the base, so I'm amending my commit :)

PR #1810, opened 2019-03-21
extracting metadata is a user error not a compiler issue

cc10512:


testdata/p4_16_errors/nasty_meta.p4
```diff
@@ -0,0 +1,38 @@
+/* -*- P4_14 -*- */
```
mbudiu-vmw: this is in the wrong folder, it should be in p4_14_errors

testdata/p4_16_errors_outputs/nasty_meta.p4-stderr
```diff
@@ -0,0 +1,4 @@
+nasty_meta.p4(6):syntax error, unexpected {, expecting (
```
mbudiu-vmw: and this one should be deleted

testdata/p4_14_errors_outputs/nasty_meta.p4-stderr
```diff
@@ -0,0 +1,4 @@
+nasty_meta.p4(6):syntax error, unexpected {, expecting (
```
mbudiu-vmw: Something is wrong, this is not the error message that you should expect.

cc10512: thanks for catching this. updated to the proper error message.

testdata/p4_14_errors_outputs/nasty_meta.p4-stderr
```diff
@@ -0,0 +1,4 @@
+nasty_meta.p4(6): [--Werror=invalid] error: struct ethernet_t: Invalid expected a header
```
mbudiu-vmw: Sorry to be such a nuisance, but I think that the error should point to the extract and not to the header type declaration.

cc10512: No nuisance, I actually appreciate it! Better error messages are very important.

cc10512:
@mbudiu-vmv I hope the message is clear now.

PR #1813, opened 2019-03-23
Convert some error messages to the new format

mbudiu-vmw:
I have also create a new class of tests in the CMakeFile, called "err". This includes all tests from testdata/p4_16_errors. Now you can say 'make check-err'.

I am not 100% happy about the way error messages are printed now. E.g. the following is a bit clumsy:
```
missing_actions.p4(19): [--Werror=expected] error: t: Expected Table t does not have an `actions' property
```

@cc10512 : maybe you have a suggestion on how to improve this.

### DIFF
--- a/frontends/p4/validateParsedProgram.cpp
+++ b/frontends/p4/validateParsedProgram.cpp
@@ -32,19 +32,19 @@ void ValidateParsedProgram::postorder(const IR::Constant* c) {
 /// Check that extern constructor names match the enclosing extern
 void ValidateParsedProgram::postorder(const IR::Method* m) {
     if (m->name.isDontCare())
-        ::error(ErrorType::ERR_INVALID, "method/function name", m->name);
+        ::error(ErrorType::ERR_INVALID, "method/function name.", m->name);
     if (auto ext = findContext<IR::Type_Extern>()) {
         if (m->name == ext->name && m->type->returnType != nullptr)
-            ::error(ErrorType::ERR_INVALID, "type. Constructor cannot have a return type", m);
+            ::error(ErrorType::ERR_INVALID, "type. Constructor cannot have a return type.", m);
         if (m->type->returnType == nullptr) {
             if (m->name != ext->name) {
-                ::error(ErrorType::ERR_INVALID, "type. Method has no return type", m);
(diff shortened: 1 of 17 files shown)


cc10512:
@mbudiu-vmw based on the feedback https://github.com/p4lang/p4c/pull/1646#discussion_r244247893 the errors messages are prefixed by a fixed string that includes `%1%`. I can see from your usage pattern that you want to write the entire error message. Agreed that makes it more readable in the code, but awkward on the the screen. I see two possibilities:
- keep the prefix and format the message appropriately. Here is an example:
```
                ::error(ErrorType::ERR_INVALID, "argument. Expected a header not %2%",
                        primitive, destType->toString());
```
Results in:
```
nasty_meta.p4(24): [--Werror=invalid] error: extract: Invalid argument. Expected a header not struct ethernet_t
    extract(ethernet);
```
`%1%: Invalid` is the prefix, so the rest of the message is formatted accordingly.

- change the catalog to have empty prefixes and rely on the developer to write the full message. This is what we do for warnings, where the diversity was too large to constrain them to a fixed set of prefixes.

mbudiu-vmw:
I will rework this to skip %1%. Didn't realize it.

mbudiu-vmw:
@cc10512 : is this better?

frontends/p4/validateParsedProgram.cpp
```diff
@@ -110,15 +114,17 @@ void ValidateParsedProgram::distinctParameters(
...
                     it->second, p);
         else
             found.emplace(p->getName(), p);
     }
     for (auto p : constr->parameters) {
         auto it = found.find(p->getName());
         if (it != found.end())
-            ::error("Duplicated parameter name: %1% and %2%",
+            ::error(ErrorType::ERR_INVALID,
+                    "name. Duplicates %2%",
```
cc10512: Since this is the same name, it would be good to find a way in which we can distinguish between the current name and its duplicate. In the current error we mark only one of the 'p's:
```
dup-param1.p4(17): [--Werror=invalid] error: p: Invalid name.  Duplicates p
control MyIngress<p>(inout bit<32> p) {
                  ^
```
Same for the other duplicates.

PR #1816, opened 2019-03-23
Fix P4 code output

ChrisDodd:
- only output full struct decl with declaration -- uses just use name
- squash symlinks when looking for #include possibilities

### DIFF
--- a/frontends/common/options.cpp
+++ b/frontends/common/options.cpp
@@ -233,6 +233,11 @@ static void convertToAbsPath(const char* const relPath, char (&output)[N]) {
 bool setIncludePathIfExists(const char*& includePathOut,
                             const char* possiblePath) {
     struct stat st;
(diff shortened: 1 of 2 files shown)


ChrisDodd:
This is really preliminary to fixing type inferencing to actually put the inferenced types into the IR instead of requiring the typeMap everywhere.  So the 'test' is the fact that none of the expected outputs from p4test change at the moment -- they should continue to not change if you add a pass that replaces all Type_Names with the referred to type (without this change, the dumped P4 code becomes invalid, with type declarations repeated at every use of the type).

mbudiu-vmw:
Another solution is to call getP4Type on these types. This converts a Type_Struct into the corresponding Type_Name, for example. 

I think you will still need the typeMap; only expressions have type fields, but there are other entities that have types, which will only be in the typeMap.

PR #1817, opened 2019-03-24
add a bound to the number of errors that are printed before exiting

cc10512:
The compiler error reporting is based on the principle that printing
as many errors as possible will allow users to fix multiple of them in
a single compilation pass. However large number of errors is usually
too hard to handle, and people will incrementally fix errors as they
go. What is that magic number? Clang uses 20, which seems quite
reasonable, so we use that as a base.

We also provide the `ErrorReporter::setMaxErrorCount` function, so
that different backends can chose their own threshold matching their
user's expectations.

And finally, a bit more cleanup to remove redundancies in the error
reporting code. The main reason for splitting error_helper and
bug_helper functions into a separate header is to avoid circular
include dependencies:
error.h -> error_reporter.h -> exceptions.h -> error.h

### DIFF
--- a/backends/p4test/CMakeLists.txt
+++ b/backends/p4test/CMakeLists.txt
@@ -83,7 +83,7 @@ set (P4_XFAIL_TESTS
   testdata/p4_16_samples/p4rt_digest_complex.p4
   testdata/p4_16_samples/psa-example-counters-bmv2.p4
   )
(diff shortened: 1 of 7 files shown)


jafingerhut:
As you can see from the automated test results, there is one recently added test P4_16 program that has expected error output with more than 20 error messages, hence the test failure.

It does seem like it would give better test coverage for such tests if there were a command line option used in the automated tests that did not limit the number of error messages displayed.

cc10512:
Fair enough. I added the argument and used it for the test that exceeds.

PR #1818, opened 2019-03-24
Proposed additional documentation of v1model.p4

jafingerhut:


jafingerhut:
There are a couple of TBDs in comments I wrote with questions I do not know the answer to yet.  I could track them down given some more time, but if Antonin beats me to the answers, I am happy to update the proposed docs with those answers.

p4include/v1model.p4
```diff
@@ -88,27 +88,116 @@ enum MeterType {
...
+     * of the colors green, yellow, or red to be written to the result
+     * out parameter.
+     *
+     * @param index The index of the meter state in the array to be
+     *              updated, normally a value in the range [0,
+     *              size-1].  If index >= size, no meter state will be
+     *              updated.
+     * @param result Type T must be bit<W> with W >= 2.  When index is
+     *              in range, the value of result will be assigned 0
+     *              for color GREEN, 1 for color YELLOW, and 2 for
```
jafingerhut: @antoninbas Hoping you can do a quick double check on the values I documented here for GREEN, YELLOW, and RED.

antoninbas: The bmv2 meter implementation doesn't restrict how many buckets / colors you can have, but yes, for a 2rTCM meter, the values are correct.

p4include/v1model.p4
```diff
@@ -88,27 +88,123 @@ enum MeterType {
...
     direct_counter(CounterType type);
+    /***
+     * count() on a direct_counter object may only be called within an
+     * action executed as a result of matching a table entry, of a
+     * table that has a direct_counter associated with it.  Calling
+     * count() causes the counter state associated with the matched
+     * table entry to be read, modified, and written back, atomically
+     * relative to the processing of other packets.
+     *
+     * TBD: Is the direct_counter state updated even if an action is
```
antoninbas: In bmv2 simple_switch, either a table has a direct counter or doesn't, and if it does the counter will be updated for every table lookup (that doesn't miss), irrespective of the action. I assume that the simple switch backend in p4c just looks at the presence / absence of the table property and doesn't verify that count is called in every action (although that would be a good check to have, and we could print a warning if count is not called in all actions).

Implementing PSA direct counters will require some work because the spec doesn't mandate that the counter state has to be updated for every action, but only for actions that call count. I'm hoping the Cornell team can get to it.

p4include/v1model.p4
```diff
@@ -88,27 +88,123 @@ enum MeterType {
...
+     * read() may only be called within an action executed as a result
+     * of matching a table entry, of a table that has a direct_meter
+     * associated with it.  Calling read() causes the meter state
+     * associated with the matched table entry to be read, modified,
+     * and written back, atomically relative to the processing of
+     * other packets, and an integer encoding of one of the colors
+     * green, yellow, or red to be written to the result out
+     * parameter.
+     *
+     * TBD: Is the meter state updated even if an action is executed
```
antoninbas: same answer as for direct counter above
in a way for direct meters, this is implied by the method name. The meter is always executed, but you only call read() if you want to output the value to a field, whereas for (indirect) meters, the method is called "execute_meter".

p4include/v1model.p4
```diff
@@ -88,27 +88,123 @@ enum MeterType {
...
+     *              be ignored by the caller.
+     */
     void read(out T result, in bit<32> index);
+    /***
+     * write() writes the state of the register array at the specified
+     * index, with the value provided by the value parameter.  If you
+     * wish to perform a read() followed later by a write() to the
+     * same register array element, and you wish the read-modify-write
+     * sequence to be atomic relative to other processed packets, you
+     * must execute them in a P4_16 block annotated with an @atomic
```
antoninbas: In practice, bmv2 considers that all actions and parser states are `@atomic`. However, for bmv2 it is not possible at the moment to have control blocks be `@atomic`

p4include/v1model.p4
```diff
@@ -117,8 +213,19 @@ extern action_profile {
     action_profile(bit<32> size);
 }
 
-// Get a random number in the range lo..hi
+/***
+ * Generate a random number in the range lo..hi, inclusive, and write
+ * it to the result parameter.
+ *
+ * TBD: Do p4c and behavioral-model support run-time variable values
```
antoninbas: You can read `hi` and `lo` from registers in theory, so they can be set at runtime. They can also come from packet headers or metadata.
The implementation is here: https://en.cppreference.com/w/cpp/numeric/random/uniform_int_distribution/uniform_int_distribution. At the moment `hi` and `lo` have to fit in 64-bit numbers (but that could be generalized fairly easily). Behavior is undefined if `lo > hi` (see https://en.cppreference.com/w/cpp/numeric/random/uniform_int_distribution/uniform_int_distribution), but a check could be added.

p4include/v1model.p4
```diff
@@ -134,7 +241,35 @@ enum HashAlgorithm {
...
+ * ingress and egress processing are complete.
+ */
 extern void mark_to_drop();
+
+/***
+ * Calculate a hash function of the value specified by the data
+ * parameter.  Call this value H.  The value written to the result out
+ * parameter is then: (base + (H % max)).
+ *
+ * TBD: Do p4c and behavioral-model support run-time variable values
```
antoninbas: For the bmv2 implementation specifically, I expect that simple_switch will segfault for a `max` of 0 because it will attempt to do a modulo 0 operation. Just like for `random` we can add a check and a log message.

mbudiu-vmw:
Whenever you update v1model.p4 the error messages will change because the line numbers change. So some tests will fail.

p4include/v1model.p4
jafingerhut: Thanks.  I am adding another commit for the direct_counter constructor explaining how to associate a direct_counter instance to a table using the "counters = <object_name>;" table property, and changing the documentation to the read() method to indicate that it is really an unnecessary method in v1model, since the counter state for each matched table entry will be updated regardless of whether the count() method is called or not.

p4include/v1model.p4
jafingerhut: Thanks.  I have another commit that will make similar changes to direct_meter comments as I mentioned above for direct_counter.

jafingerhut:
@mbudiu-vmw Yes, I will update those failing tests after we review and update the changes to the documentation comments, since that process will likely change those line numbers again.

p4include/v1model.p4
jafingerhut: Can BMv2 have multiple threads performing ingress processing in parallel, for different subsets of input ports?

Or multiple threads performing egress processing in parallel, for different subsets of output ports?

If so, does it cause action execution to be atomic by using a lock every time an action is executed?

If not, how is action execution atomic?

p4include/v1model.p4
jafingerhut: Are you OK with leaving that generality of bmv2's implementation out of these comments?  I could mention it briefly if you want, but am afraid it is fairly far off the beaten track of how most people would use meters, and may only confuse some people.

p4include/v1model.p4
jafingerhut: I have another commit for the documentation that explicitly makes the value written to the out parameter unspecified if lo > hi.  I could imagine adding a run-time warning in BMv2, but at least with that behavior documented, people will have some kind of forewarning.

p4include/v1model.p4
jafingerhut: I will add another commit in a moment that proposes the behavior that if max=0, then the value written to the out parameter will always be equal to base, the same behavior as if max=1.  A warning in the log seems reasonable, but if explicitly documented maybe not high priority, since people may even come to rely upon such documented behavior.

If you prefer the documentation to say that the value written to the out parameter is unspecified if max=0, that also seems reasonable to me.

The current bmv2 implementation does segfault, as you suspected, according to my testing.

p4include/v1model.p4
antoninbas: Each register has an associated mutex. When executing an action, we acquire the mutexes for all the registers referenced in the action, in a predetermined order.

jafingerhut: Interesting.  I've just pushed another commit that attempts to describe this atomicity guarantee provided by the BMv2 implementation.

antoninbas: I wonder if that kind of details belong in v1model.p4 or in behavioral-model/docs. Based on the P4 spec, programmers should still be using `@atomic`, and the bmv2 compiler should complain if the scope of `@atomic` is too large. In the future, there is nothing preventing bmv2 from supporting arbitrary `@atomic` blocks.

mbudiu-vmw: Yes, v1model can be supported by other implementations besides bmv2. We should not be too specific in this file. I actually don't know if there are other implementations, though.

jafingerhut: I am fine moving BMv2 specific parts of this documentation elsewhere if you prefer.  I do think the BMv2-specific support/limitations should be documented at least once somewhere.  I will make another commit that attempts to remove BMv2 specifics from these changes.

jafingerhut: I have added a commit to this PR that removes the parts of the documentation that I believe to be BMv2-specific.  Those parts have been put into a separate PR here: https://github.com/p4lang/behavioral-model/pull/741

jafingerhut: To be strictly accurate, any or all of what is included in this PR _might_ differ in another v1model implementation.  I do not have any knowledge of other v1model implementations to check, though.

jafingerhut:
@antoninbas @mbudiu-vmw I _think_ that except for correcting the few failing tests to adjust line numbers in v1model.p4, I am done making changes here.  You may want to let the dust settle for a day or so to see if I try to throw in anything else, but I do not plan to.

jafingerhut:
@mbudiu-vmw Antonin has approved, and I have added the changes to the few error messages that need updating for the tests to pass.  I do not have any current plans to make any more changes to v1model.p4 (but hey, I may find something else that bugs me and want to document it better -- some other PR if so).

p4include/v1model.p4
```diff
@@ -134,7 +265,36 @@ enum HashAlgorithm {
     xor16
 }
 
+/***
+ * mark_to_drop() is a primitive action that modifies
+ * standard_metadata.egress_spec to an implementation-specific special
```
mbudiu-vmw: I would add that these changes happen after the pipeline has been completed. These do not happen when the operation is executed.

jafingerhut: I believe that is not true.  I believe that in the current p4c + simple_switch implementation, calls to mark_to_drop make modifications to standard_metadata.egress_spec and standard_metadata.mcast_grp _immediately_, which is why you can change those fields later and have the effect of the drop undone.

jafingerhut: It is recirculate/resubmit/clone operations that in P4_14 have a delayed effect, when they are requested to preserve metadata fields.  Even in P4_14, the drop action has an immediate effect that can later be changed before the end of ingress/egress processing, such that the drop can be undone.

mbudiu-vmw: If that's true, this is another instance of magic action performed by an operation that does not refer to some fields, which should be impossible in p4-16. The compiler will assume that these fields cannot be changed by the mark_to_drop call. The only solution is to convert this call to something like `mark_to_drop(out standard_metadata meta)`;

jafingerhut: Wow, I can't believe I discovered all of this about recirculate/resubmit/clone, and didn't notice this :-)  You are right, mark_to_drop() is modifying something that it doesn't have explicit access to.

mbudiu-vmw: I think you got the emoticon wrong, it should be :-(
Having this documentation early would have helped avoid this problem.
We should file another issue. We have to fix the p4-14 converter and invalidate a bunch of user programs.

jafingerhut: And I think it should be 'inout' direction if we make the entire standard_metadata struct a parameter to mark_to_drop(), because it should not change any of the many other fields inside that struct, except egress_spec and mcast_grp.

Adding such a parameter to mark_to_drop() is going to have an small effect on a lot of P4_16 test programs, I suspect, and other extant v1model code out there.  I am not saying that it should not be done -- just pointing out the potential impact.

jafingerhut: I can file a separate issue.

Would you prefer that I remove the part of these changes that document mark_to_drop?

Or keep them as they are, perhaps plus a note that mark_to_drop is violating the P4_16 language spec?

mbudiu-vmw: We should eventually mark this as `@deprecated` and add the inout version.

jafingerhut: I have created the PR you described here: https://github.com/p4lang/p4c/issues/1828

Let me know if you want any further changes to this PR before approving it.

jafingerhut: @mbudiu-vmw No rush on this PR, since it is only documentation changes.  I just wanted to ensure you knew that unless you ask for other changes, or remind me of some you have asked for that I have forgotten, I am done changing this.

mbudiu-vmw:
We should wait until we merge #1835, update the documentation for mark_to_drop, and then merge this one as well.

mbudiu-vmw:
Once I merge the mark_to_drop changes we can update this and merge.

mbudiu-vmw:
@jafingerhut : I think you can now revise this file and then it will be ready to merge. You may also need to resolve the conflicts.

mbudiu-vmw:
@jafingerhut @antoninbas I hit a nasty problem: with these definitions in v1model the queueing_metadata header cannot have a size which is a multiple of 8 bits... (in particular, enq_qdepth are problematic).
I don't know how to work around this problem...

jafingerhut:
I know that I made changes so that the bmv2 back end code rejects _headers_ that are not a multiple of 8 bits long.

But non-header "structs" (i.e. those things with the "metadata" field equal to true in the BMv2 JSON file) should still be allowed to have arbitrary lengths, I thought.

Either that or the error you are hitting is completely outside of my experience.  Happy to look at some PR changes on my local machine and reproduce what you are seeing, if you have pushed commits that are up to date with what you have locally.

hesingh:
It would be good to document the meter API because some folks are confused what is "size" in the API. 

https://github.com/p4lang/p4c/blob/master/p4include/v1model.p4#L100

mbudiu-vmw:
The problematic program is p4_14_samples/queueing.p4. This program explicitly declares a header queueing_hdr of type intrinsic_metadata_t. It actually has different definitions for queueing_metadata_t, but I had to bring it in line with the standard_metadata_t definitions in v1model.

I guess the only solution is to use a different type for the header...

PR #1820, opened 2019-03-25
Refactor GeneralInliner common code

ChrisDodd:
- template combining P4Control and P4Parser inlining
- extend avoidance of unnecessary copying to parser inlining

This would be much simpler if P4Control and P4Parser had a common base class, but using a template allows the two cases to share common code.


backends/bmv2/CMakeLists.txt
```diff
@@ -177,7 +177,7 @@ set (XFAIL_TESTS
   testdata/p4_16_samples/psa-example-counters-bmv2.p4
   testdata/p4_16_samples/psa-example-digest-bmv2.p4
   testdata/p4_16_samples/psa-example-register2-bmv2.p4
-  testdata/p4_16_samples/psa-fwd-bmv2.p4
```
hanw: removed, but the intermediate file is still updated?

ChrisDodd: removed from XFAIL -- this test now passes...

PR #1835, opened 2019-03-28
Deprecate mark_to_drop(); replace with mark_to_drop(standard_metadata) function

### LINKED ISSUE #1828: P4_16 + v1model mark_to_drop violates P4_16 language semantics
First pointed out by Mihai Budiu here: https://github.com/p4lang/p4c/pull/1818#discussion_r269311117

When BMv2 simple_switch executes a P4_16 + v1model architecture program that executes a `mark_to_drop()` action, it implements it by immediately modifying `standard_metadata.egress_spec` and `standard_metadata.mcast_grp`.  See here for the implementation: https://github.com/p4lang/behavioral-model/blob/master/targets/simple_switch/primitives.cpp#L143-L150

The problem from the P4_16 language specification point of view is: `mark_to_drop` should not be able to do so, because it does not take the standard_metadata struct as a parameter.  Yes, externs have "arbitrary power", but not really arbitrary, because they are not permitted to access variables in your program that are not run-time parameters.  See the first two paragraphs of the P4_16 language specification, version 1.1, Section 6.7.1 titled "Justification".

This is similar to issues raised about the current P4_16 + v1model implementations of recirculate, resubmit, and clone when preserving user-defined metadata fields.  Now that I write this, I realize that those operations _also_ currently modify fields within the standard_metadata struct (e.g. clone_spec, recirculate_flag, resubmit_flag, lf_field_list) without having access to them, according to P4_16 language specification restrictions.

One could argue in those cases that perhaps those fields in standard_metadata should not even be visible to the user's P4 program, and are instead "hidden state inside the architecture", but that argument does not apply for the mark_to_drop case, since egress_spec and mcast_grp fields are fully expected to be visible to, both read and write, by the user's P4 program.

Possible change that could be made: change the definition of mark_to_drop to take the standard_metadata struct as an inout parameter, like so:
```
extern void mark_to_drop(inout standard_metadata_t stdmeta);
```
This would require updating a fair number of P4_16 test programs in p4c, as well as P4 tutorials based on P4_16 + v1model architecture, etc.

mbudiu-vmw:
Fixes #1828. 
This also replaces some IndexedVectors with Vectors - because the top-level program can contains multiple declarations with the same name.

### DIFF
--- a/frontends/p4/fromv1.0/converters.cpp
+++ b/frontends/p4/fromv1.0/converters.cpp
@@ -1120,26 +1120,6 @@ class DetectDuplicates: public Inspector {
     }
 };
 
-// The fields in standard_metadata in v1model.p4 should only be used if
-// the source program is written in P4-16. Therefore we remove those
-// fields from the translated P4-14 program.
-class RemoveAnnotatedFields : public Transform {
- public:
-    RemoveAnnotatedFields() { setName("RemoveAnnotatedFields"); }
-    const IR::Node* postorder(IR::Type_Struct* node) override {
-        if (node->name == "standard_metadata_t") {
-            auto fields = new IR::IndexedVector<IR::StructField>();
-            for (auto f : node->fields) {
-                if (!f->getAnnotation("alias")) {
-                    fields->push_back(f);
-                }
-            }
-            return new IR::Type_Struct(node->srcInfo, node->name, node->annotations, *fields);
-        }
-        return node;
-    }
-};
-
 // If a parser state has a pragma @packet_entry, it is treated as a new entry
 // point to the parser.
 class CheckIfMultiEntryPoint: public Inspector {
@@ -1166,7 +1146,7 @@ class CheckIfMultiEntryPoint: public Inspector {
 // does not use the @packet_entry pragma.
 class InsertCompilerGeneratedStartState: public Transform {
     ProgramStructure* structure;
-    IR::IndexedVector<IR::Node>        allTypeDecls;
+    IR::Vector<IR::Node>               allTypeDecls;
     IR::IndexedVector<IR::Declaration> varDecls;
     IR::Vector<IR::SelectCase>         selCases;
     cstring newStartState;
@@ -1284,7 +1264,6 @@ Converter::Converter() {
     passes.emplace_back(new ComputeTableCallGraph(structure));
     passes.emplace_back(new Rewriter(structure));
     passes.emplace_back(new FixExtracts(structure));
-    passes.emplace_back(new RemoveAnnotatedFields);
     passes.emplace_back(new FixMultiEntryPoint(structure));
 }
 
--- a/backends/bmv2/CMakeLists.txt
+++ b/backends/bmv2/CMakeLists.txt
@@ -172,7 +172,6 @@ set (XFAIL_TESTS
   testdata/p4_16_samples/psa-example-counters-bmv2.p4
   testdata/p4_16_samples/psa-example-digest-bmv2.p4
   testdata/p4_16_samples/psa-example-register2-bmv2.p4
-  testdata/p4_16_samples/psa-fwd-bmv2.p4
   # This reads stack.next
   testdata/p4_16_samples/issue692-bmv2.p4
   # These test use computations in the verify/update checksum controls - unsupported
--- a/backends/bmv2/CMakeLists.txt
+++ b/backends/bmv2/CMakeLists.txt
@@ -144,11 +144,6 @@ set (XFAIL_TESTS
   testdata/p4_14_samples/issue60.p4
   # compiler claims (incorrectly?) that c2 has mulitple successors, so is not supported
   testdata/p4_14_samples/issue-1426.p4
-  # As of 2019-Feb-04 latest p4c, this program fails due to the root
-  # cause of both issues #1694, but is not affected by #1669.  I have
-  # tested it with the proposed fix for issue #1694 that is on PR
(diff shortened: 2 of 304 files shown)


mbudiu-vmw:
Yes, I think Andy's suggestion is better, but it will take a bit more work to implement.
What I am thinking to do is to actually make it an external function, and then "inline" it in the back-end. The back-end would still hardwire a value for the port, but that's architecture-dependent, and not exposed in the v1model file. The current implementation in bmv2 calls a drop primitive, but that won't be exactly the same semantically (e.g., the user could pass a copy of the standard_metadata to mark_to_drop).

jafingerhut:
Currently simple_switch supports port numbers in the range [0,510], with 511 reserved as a "drop port".

There is a request for, and a PR Antonin has developed, to extend the port number range to arbitrary 32-bit values, but I believe that the default behavior without extra command line options will be to stay at the current [0, 510] range.

Related to that change, he is planning to make the "drop port" number a command line option to simple_switch, so that those who want to extend the port range, e.g. to [0, 4094], could configure 4095 as the "drop port".

One way to support that in p4c bmv2 v1model back end would be to _not_ ever put a hard-coded "drop port" number like 511 in the BMv2 JSON file.  Today, no such number is put into the BMv2 JSON file -- a call to a primitive drop action is put there, and simple_switch is the only code that enshrines the specialness of the number 511.  It may be preferable to continue that approach, never placing a hard-coded "drop port" number anywhere in the BMv2 JSON file.

antoninbas:
That sounds good to me @mbudiu-vmw  but I'd like to have the ability to change the drop port in simple_switch dynamically based on a command-line option. I wanted to use the following implementation:
```
class mark_to_drop2 : public ActionPrimitive<const Header &> {
  void operator ()(const Header &std_hdr) {
    get_field(std_hdr, "egress_spec").set(user_config.DROP_PORT);
    get_field(std_hdr, "mcast_grp").set(0);
  }
};
```
This will respect the P4 semantics because `std_hdr` can he any header with type `standard_metadata_t`. So the P4 programmer is free to shoot himself in the leg by passing a copy of the architecture-provided standard metadata header.

mbudiu-vmw:
So what should the JSON generated be?  Can I assume that we have a mark_to_drop2 primitive action?
Note: in v1model I will still call it mark_to_drop, because functions can be overloaded.

antoninbas:
There is no existing `mark_to_drop` in simple_switch so I can use that name as well
Yes, you can assume `mark_to_drop` is available and call `mark_to_drop(<hdr>)`, I'll get started on it

antoninbas:
BTW, I made a typo above. The bmv2 parameter type will be `Header &` not `const Header &` since we modify a field in the header.

mbudiu-vmw:
This PR will fail, because it assumes that BMv2 has a primitive action mark_to_drop(stdmeta). Once that is available it should hopefully succeed. But I still submitted it for review.

I have also filed #1836, which will improve the code quality for some of the cases when mark_to_drop is called from global actions.

antoninbas:
@mbudiu-vmw  I have added `mark_to_drop`: https://github.com/p4lang/behavioral-model/pull/748

mbudiu-vmw:
The reference outputs for P4-14 programs that I committed are still wrong; let me fix these as well.

mbudiu-vmw:
See https://github.com/p4lang/behavioral-model/pull/748

mbudiu-vmw:
@antoninbas : this converts from mark_to_drop() to mark_to_drop(standard_metadata).

antoninbas:
@hanw maybe you should look at this before Mihai merges it and confirm this is not too disruptive

jafingerhut:
@mbudiu-vmw Just to make sure I understand:

One could imagine trying to make changes to p4c such that existing v1model programs that call `mark_to_drop();` give a warning that this call is deprecated, but they still work, and they produce a BMv2 JSON file, as well as they do today with the latest p4c without these changes.

That is _not_ the intent of these changes.

The intent with these changes seems to be: "give a deprecation warning if the v1model program calls mark_to_drop(), and also give a compilation error, and do not produce a BMv2 JSON file".

So, not merely deprecate `mark_to_drop()`, but also completely disallow?

I guess that is OK, if it is what folks want.  It seems a bit sudden in one commit of the compiler, but of course the older versions of the compiler are still available for people that want to compiler existing v1model programs without changing them, and the changes required are pretty straightforward.

mbudiu-vmw:
We can keep the old one around, but I think it's wrong, because it may not work (depending on what else the user does to standard_metadata and how the compiler optimizes the program). We should better tell people that right away rather than letting them ignore the warning and get possible incorrect results.

frontends/p4/fromv1.0/converters.cpp
```diff
@@ -1166,7 +1146,7 @@ class CheckIfMultiEntryPoint: public Inspector {
 // does not use the @packet_entry pragma.
 class InsertCompilerGeneratedStartState: public Transform {
     ProgramStructure* structure;
-    IR::IndexedVector<IR::Node>        allTypeDecls;
+    IR::Vector<IR::Node>               allTypeDecls;
```
ChrisDodd: Why change this from a more efficent IndexedVector to a less efficient Vector?  Particularly as this ends up being the global scope, having to do a linear search through the vector could get expensive for large programs?

mbudiu-vmw:
Because we now have two declarations called mark_to_drop. With overloading allowed the top level declarations can have duplicates.

Issue #1829, opened 2019-03-26
v1model direct_meter-using program that would be nice if it gave better message, not CompilerBug


jafingerhut:
Program is attached.  It does have an error, but it gives a Compiler Bug output with the latest p4c as of 2019-Mar-23:
```
$ p4c --target bmv2 --arch v1model v1model-direct-meter-compiler-bug1.p4 
In file: /home/jafinger/p4c/backends/bmv2/common/metermap.cpp:41
Compiler Bug: /home/jafinger/p4c/backends/bmv2/common/metermap.cpp:41: Null info
```
An error message something like the following would be more useful:
```
direct_meter 'da_meter' must have at least one action for table 'mac_da' with a 'da_meter.read(...)' method call
```
[v1model-direct-meter-compiler-bug1.p4.txt](https://github.com/p4lang/p4c/files/3010685/v1model-direct-meter-compiler-bug1.p4.txt)

I believe that this issue lies completely within the bmv2 v1model back end code.

jafingerhut:
3 other test programs.  One of these is the same case as the one attached to the original comment, the other three are different variations.  All give Compiler Bug with following command for latest version of p4c as of 2019-Mar-23:
```
p4c --target bmv2 --arch v1model progname.p4
```
[issue1829-1-bmv2.p4.txt](https://github.com/p4lang/p4c/files/3015478/issue1829-1-bmv2.p4.txt)
[issue1829-2-bmv2.p4.txt](https://github.com/p4lang/p4c/files/3015480/issue1829-2-bmv2.p4.txt)
[issue1829-3-bmv2.p4.txt](https://github.com/p4lang/p4c/files/3015482/issue1829-3-bmv2.p4.txt)
[issue1829-4-bmv2.p4.txt](https://github.com/p4lang/p4c/files/3015483/issue1829-4-bmv2.p4.txt)

jafingerhut:
And I forgot to mention in my previous comment that it seems to me that all four of those programs should _not_ successfully compile to a BMv2 JSON file.  They all have an error of some kind that I believe should prevent that.

PR #1837, opened 2019-03-29
Fix issue #1829 - better error messages for some wrong uses of direct…

jafingerhut:
…_meter


jafingerhut:
These are some alternate proposed changes for issue #1829.  I have tested that the error messages seems reasonable when run with `p4c --target bmv2 --arch v1model myprog.p4` for all 4 of the test programs added, but those error messages do not show up in the expected outputs here, I believe because all of the `-stderr` files are stderr outputs from `p4test`.

jafingerhut:
Is there a way to have a -bmv2.p4 source file, so that the BMv2 v1model back end is run, and have an expected stderr output file for them?

hesingh:
If I understand correctly what you want, from `p4c/build/` run this example command:

`bmv2/testdata/p4_16_samples/issue1829-1-bmv2.p4.test`

Then, if any error exists, the error should be emitted to the console and the bmv2 backend is run.

Or try `.\p4c-bm2-ss --std p4-16 ../testdata/p4_16_samples/issue1829-1-bmv2.p4`

A place to add a new negative test for bmv2 is `testdata/p4_16_bmv_errors/`

jafingerhut:
Thanks for the help.  Moving 3 of the tests that should give error messages in the v1model back end code into the p4_16_bmv_errors directory seems to do the trick.

mbudiu-vmw:
If you are happy about the errors produced by this we should merge it.

jafingerhut:
@antoninbas @cc10512 Do one of you want to look over the new error messages introduced by these changes to see if you have any suggested changes?  The patch is short, and the quickest way to see what the new error messages look like.

backends/bmv2/common/metermap.cpp
```diff
@@ -87,7 +93,14 @@ void DirectMeterMap::setDestination(const IR::IDeclaration* meter,
  */
 void DirectMeterMap::setSize(const IR::IDeclaration* meter, unsigned size) {
     auto info = getInfo(meter);
-    CHECK_NULL(info);
+    if (info == nullptr) {
+        /* This case may is reached if a table has a direct_meter
```
antoninbas: typo: `may be`

jafingerhut:
Three of the new test programs only produce errors when you run the bmv2 v1model back end, and are in the `testdata/p4_16_bmv_errors` directory.  The p4c test infrastructure expects those to fail, but as far as I can tell does not check the error output against any file contents.

You can check out my branch locally, build p4c, and try running `p4c --target bmv2 --arch v1model prog.p4` on those 3 programs if you want to see the error messages "live", or you can simply read the C++ code changes to see what those error messages look like.

backends/bmv2/common/metermap.cpp
jafingerhut: fixed

jafingerhut:
@mbudiu-vmw Antonin approved after reviewing error messages, so going ahead and merging this.

Issue #1784, opened 2019-03-11
P4Program should carry the enumMap


mbudiu-vmw:
There may be other data structures that should be part of the P4Program instead of being side maps.

mbudiu-vmw:
Actually this may be unnecessary; @vidakjelena has a solution which involves simply not deleting the enum type declarations from the program once the enums are substituted.

PR #1842, opened 2019-04-04
Load support for ir enum map when --fromJSON is used (#1784)

vidakjelena:
*Modifed convertEnums to save user defined enums in program
instead of removing them.

*Added pass fillEnumMap, which is called in reduced midEnd branch,
that fills enumMap

Signed-off-by: Jelena Vidakovic <jelena.vidakovic@rt-rk.com>

### DIFF
--- a/backends/bmv2/psa_switch/midend.cpp
+++ b/backends/bmv2/psa_switch/midend.cpp
@@ -53,6 +53,7 @@ limitations under the License.
 #include "midend/expandEmit.h"
 #include "midend/tableHit.h"
 #include "midend/midEndLast.h"
+#include "midend/fillEnumMap.h"
 
 namespace BMV2 {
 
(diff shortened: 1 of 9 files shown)


mbudiu-vmw:
This seems to have failed because the switch.p4 test took more than 10 minutes and Travis terminated the job.

mbudiu-vmw:
If you can rebase and rerun I hope this will succeed and then I plan to merge this.

vidakjelena:
Hello,
I updated PR. According to Travis results on previous version of my PR, I found out that on OSX it behaves different, I test it, and on my Mac machine that I used for testing and it doesn't fail. Just in case, I supported for Linux, for now.  In meantime I will figure out what the problem with this result is for OSX and update if I manage.

mbudiu-vmw:
The behavioral model simulator changes frequently.

Issue #1803, opened 2019-03-19
Introducing conflicting names with the @name annotation, should it be a frontend error?


antoninbas:
I wrote the following P4 program:
```
#include <core.p4>
#include <v1model.p4>

struct H { };
struct M {
    bit<32> hash1;
}

parser ParserI(packet_in pk, out H hdr, inout M meta, inout standard_metadata_t smeta) {
    state start { transition accept; }
}

action empty() { }

control MyC(inout standard_metadata_t smeta)() {
    action drop() { mark_to_drop(); }

    @name(".t0")
    table t0 {
        key = { smeta.ingress_port : exact; }
        actions = { drop; NoAction; }
        const default_action = NoAction();
    }
    apply {
        t0.apply();
    }
}

control IngressI(inout H hdr, inout M meta, inout standard_metadata_t smeta) {

    MyC() c1;
    MyC() c2;

    apply {
        c1.apply(smeta);
        c2.apply(smeta);
    }

};

control EgressI(inout H hdr, inout M meta, inout standard_metadata_t smeta) {
    apply { }
};

control DeparserI(packet_out pk, in H hdr) {
    apply { }
}

control VerifyChecksumI(inout H hdr, inout M meta) {
    apply { }
}

control ComputeChecksumI(inout H hdr, inout M meta) {
    apply { }
}

V1Switch(ParserI(), VerifyChecksumI(), IngressI(), EgressI(),
         ComputeChecksumI(), DeparserI()) main;
```

I compiled it with the the following command: `./p4test same_name.p4 --p4runtime-files=same_name.p4info.pb.txt`.

I didn't get any error and a P4Info message is generated. The generated P4Info is invalid however, because it includes 2 entries for table "t0":
```
tables {
  preamble {
    id: 33580881
    name: "t0"
    alias: "t0"
  }
  match_fields {
    id: 1
    name: "smeta.ingress_port"
    bitwidth: 9
    match_type: EXACT
  }
  action_refs {
    id: 16815162
  }
  action_refs {
    id: 16800567
  }
  const_default_action_id: 16800567
  size: 1024
}
tables {
  preamble {
    id: 33580881
    name: "t0"
    alias: "t0"
  }
  match_fields {
    id: 1
    name: "smeta.ingress_port"
    bitwidth: 9
    match_type: EXACT
  }
  action_refs {
    id: 16790297
  }
  action_refs {
    id: 16800567
  }
  const_default_action_id: 16800567
  size: 1024
}
```
[...]

PR #1844, opened 2019-04-05
Detect duplicate names in P4Info

antoninbas:
See issue #1803

We iterate over the P4Info after generating it to look for duplicate
names, and generate an error if we found one. Unfortunately we cannot
rely on the SymbolTable for duplicate detection, because exsting code
relies heavily on the ability to call add() multiple times with the same
name. For example top-level actions are inlined by the compiler in all
the controls using them and we end up calling add() for each copy. In
this case, it is not considered an error because these "duplicate" calls
actually refer to the same object as far as the control-plane is
concerned and we do not end up with duplicates in the P4Info message.

A negative test case is added to p4_16_errors. This required slightly
changing the testing framework so that programs under p4_16_errors are
compiled with --p4runtime-files.


testdata/p4_16_errors_outputs/issue1803_same_table_name.p4-stderr
```diff
@@ -0,0 +1,2 @@
+[--Werror=legacy] error: Name 't0' is used for multiple objects in the P4Info message
```
mbudiu-vmw: It would be nice to improve this error by displaying the P4 constructs that have the name in the source code. Do you have references to the IR nodes at this point?

control-plane/p4RuntimeSerializer.cpp
```diff
@@ -1611,9 +1687,12 @@ P4RuntimeSerializer::generateP4Runtime(const IR::P4Program* program, cstring arc
     auto* p4RuntimeProgram = program->apply(p4RuntimeFixups);
     auto* evaluatedProgram = evaluator->getToplevelBlock();
 
-    BUG_CHECK(p4RuntimeProgram && evaluatedProgram,
-              "Failed to transform the program into a "
-              "P4Runtime-compatible form");
+    if (!p4RuntimeProgram || !evaluatedProgram) {
+        ::error(ErrorType::LEGACY_ERROR,
```
mbudiu-vmw: I don't think you should use LEGACY_ERROR. I think there's a DUPLICATE error.

control-plane/p4RuntimeSerializer.cpp
```diff
@@ -1611,9 +1687,12 @@ P4RuntimeSerializer::generateP4Runtime(const IR::P4Program* program, cstring arc
...
+                "Failed to transform the program into a "
```
mbudiu-vmw: This message is a bit confusing. How about "error while generating P4Runtime API?"

testdata/p4_16_errors_outputs/issue1803_same_table_name.p4-stderr
antoninbas: We don't have the IR nodes at this point

control-plane/p4RuntimeSerializer.cpp
antoninbas: There wasn't, but I just added one. I don't know if it will be useful outside of the P4Runtime code.

antoninbas:
> 
> 
> It this comment accurate?

I hope so. Why makes you think it's not?

PR #1846, opened 2019-04-08
fix stf test

hanw:
`expect` command should run before `packet` command to avoid race condition in test where a test packet is generated before the `expect` command is executed.


jafingerhut:
@hanw Doesn't this mean that nearly all STF tests have this same bug?

I know that I have written STF tests based upon what I saw working in existing STF tests, which every one I can remember looking at always had 'packet' before the corresponding 'expect' commands.

The STF test infrastructure used by p4c right now works, if I recall correctly, by collecting all expect statements up and saving them in memory, then going back and generating all input packets from 'packet' statements.  It does not work by executing things like the lines of the STF file are a sequential execution of anything, as far as I can tell.

ChrisDodd:
Most of the 129 stf tests in the repo have expect commands before packet commands for precisely this issue; these two were cases where we were seeing intermiitent failures.

jafingerhut:
File arith1-bmv2.stf has packet-before-expect pairs.  It does not give you intermittent failures, too?

jafingerhut:
I have only examined 29 of the STF files in p4c/testdata/p4_16_samples so far, but 25 of those have packet lines before the corresponding expect lines.

Perhaps there is some other reasons those 2 tests were giving intermittent failures?  I don't see how it can be solely due to the fact that they had packet lines before the corresponding expect lines.

jafingerhut:
Anyway, for the current test infra for p4c, it probably doesn't hurt anything if all of the STF test files are updated to put expect lines before the corresponding packet lines, but if for some reason only those 2 were causing you problems, it is certainly less work to only update the failing ones.  Just seems awfully weird that the majority of them do not cause such failures for the same reason, since as far as I can tell, the majority of the tests are written to put a 'packet' line before the corresponding 'expect' lines for output packets generated by the input packet.

hesingh:
Makes sense to change all or someone else in future will ask why some STF files has code in a certain way while others do not.  It's a mechanical change - let's change all, please.

PR #1848, opened 2019-04-10
add bmv2 psa_switch stf tests for drop all, unicast, multicast 

peteli3:
Adds the following (x.p4, y.stf) test pairs for bmv2 - intended for psa switch target:
- drop all packets by doing nothing in the p4 (since `psa_ingress_output_metadata.drop` is marked true by default)
- unicast or drop (authored by Andy)
- multicast


backends/bmv2/CMakeLists.txt
```diff
@@ -183,6 +183,11 @@ set (XFAIL_TESTS
   # These test use computations in the verify/update checksum controls - unsupported
   testdata/p4_16_samples/issue1765-bmv2.p4
   testdata/p4_16_samples/issue1765-1-bmv2.p4
+  # These tests require bmv2's psa_switch to support std metadata passing at minimum
```
mbudiu-vmw: what do we need to do to have these tests pass?

peteli3: Broadly, we'd need to make bmv2's psa_switch stop referencing v1model std metadata and start using psa metadata names instead.

More specifically for each test,

To get unicast to work:
When ingress_thread moves a pkt onto the egress buffers, it should enqueue the pkt using `psa_ingress_output_metadata.egress_port` instead of `standard_metadata.egress_spec`

To get multicast working:
Check if `psa_ingress_output_metadata.multicast_group` is nonzero, and use PRE to replicate & enqueue the new packets to go out their respective ports.

To get drop-all working:
This one is still a bit unclear. I'll be looking into it. Getting the following exception when running the STF test
`lt-psa_switch: ../../include/bm/bm_sim/data.h:192: unsigned int bm::Data::get_uint() const: Assertion `arith' failed.`

mbudiu-vmw: These all seem to be bmv2 issues. If that's true, perhaps you can file an issue there.

peteli3: Oh yes, forgot to mention. I'm part of the Cornell student team that's working on getting a minimal PSA switch in bmv2 working.

These changes are underway and will be submitted for review in the next few days. I'll submit a subsequent PR to remove these from xfail after those changes are checked in.

jafingerhut:
I know this has already been merged and I am late to the reviewing game here, but I have looked over the STF files, and they all appear to be good tests to me.

PR #1849, opened 2019-04-12
Restrict duplicate name check for P4Info to objects of the same "type"

antoninbas:
It seems that some people use P4Runtime with P4_14 programs. In P4_14 it
is valid (and quite common) when defining a keyless table with a const
default action to use the same same for both the table and the default
action. When the compiler translates the program to P4_16, it ensures
that the names are different but preserves the original names with a
@name annotation. The @name annotation is used when generating the
P4Info message. In order to avoid breaking backward-compatibility for
these P4_14 programs, it seems that we need to restrict the duplicate
name check to objects of the same type. This is a weaker guarantee than
what is required by both the P4_16 and the P4Runtime spec but in my
experience consumers of P4Info are capable of handling non-unique names,
as longs as names are unique for a given object type.

### DIFF
--- a/control-plane/p4RuntimeSerializer.cpp
+++ b/control-plane/p4RuntimeSerializer.cpp
@@ -725,25 +725,21 @@ class P4RuntimeAnalyzer {
         return p4Info;
     }
 
-    /// Check for objects with duplicate names in the generated P4Info message
-    /// and @return the number of duplicates.
-    size_t checkForDuplicates() const {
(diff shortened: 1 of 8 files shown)


antoninbas:
@mbudiu-vmw let me know if you think it would be better to still fail for non P4-14 programs (e.g. using the isv1 information from the refMap). Even though in that case, it would mean that the behavior would be different when invoking the compiler directly on a P4_14 program or on the same program previously translated translated to P4_16 by the compiler...

PR #1850, opened 2019-04-12
remove dependency between controlplane and frontend

hanw:
controlplane should probably only depend on genIR, not on frontend.


mbudiu-vmw:
It looks like this broke the build. 
```
In file included from /home/p4c/frontends/parsers/parserDriver.h:25:0,
                 from /home/p4c/frontends/p4/parseAnnotations.h:22,
                 from /home/p4c/control-plane/p4RuntimeSerializer.cpp:46,
                 from /home/p4c/build/control-plane/unified_controlplane_sources_1.cpp:5:
/home/p4c/frontends/parsers/p4/abstractP4Lexer.hpp:4:10: fatal error: frontends/parsers/p4/p4parser.hpp: No such file or directory
 #include "frontends/parsers/p4/p4parser.hpp"
          ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

hanw:
How does it pass regression if it fails the build?

mbudiu-vmw:
You were lucky in the Travis test; if the front-end is built before the control-plane the file gets generated.
The build for p4c-xdp was not lucky.
It's a race.

mbudiu-vmw:
The point is that if you have this dependence on the header file you have to build the front-end first.

hanw:
Hopefully this PR https://github.com/p4lang/p4c/pull/1851 fixes the issue.

mbudiu-vmw:
It's strange, I still see the same build failure. So it looks like the fix wasn't enough for some reason.
https://travis-ci.org/vmware/p4c-xdp/builds/519758019?utm_medium=notification&utm_source=email

hanw:
I tried to reproduce the issue locally, but was unable to. Can you reproduce the problem on a local machine?

mbudiu-vmw:
I cannot reproduce reliably a race. You can try with -j1.
But the log in the travis build shows the build order, so you can see that the parser generator has not yet been run.

Issue #1856, opened 2019-04-15
extern method typeParameter not populated in methodInstance.cpp


hanw:
@mbudiu-vmw 

I come across a case that need this TODO item to be implemented. Do you remember why this is not implemented before?
https://github.com/p4lang/p4c/blob/master/frontends/p4/methodInstance.cpp#L96

hanw:
For example
` mirror.emit<mirror_metadata_h>(meta.session_id, {meta.srcAddr, meta.lkp_flag2}); `

The `mirror_metadata_h` type information is lost.

hanw:
Tracing down the typeSubstitutionVisitor.cpp pass
```
T/920 -> <Type_Name>(3565) mirror_metadata_h
Removing from generic parameters T/920
Visiting <Type_Var>(399172) T
Replacing T/920 with mirror_metadata_h
Specialized void _<T/920>(in bit<10> session_id, in T/920 hdr)
        into void _<>(in bit<10> session_id, in mirror_metadata_h hdr)
```

The question is why is the type parameter T on the emit method removed?

The extern definition is
```
extern Mirror {
  Mirror();
  void emit<T>(in MirrorId_t session_id, in T hdr);
}

**Comment by mbudiu-vmw** [comment]
After specializing the type parameter the method no longer has a type parameter. Type parameters are only variables. The type variable substitution can be obtained in multiple ways. For example, the type of the second parameter.

**Comment by mbudiu-vmw** [comment]
What do you want to achieve? This part of type checking works quite well, I don't think you should need to change it.

**Comment by jnfoster** [comment]
You've both taken CS 6110! (https://www.cs.cornell.edu/courses/cs6110/2016sp/lectures/lecture30.pdf)

**Comment by hanw** [comment]
I was trying to implement a check in the midend to allow this P4 program,
```
mirror.emit<mirror_metadata_h>(meta.session_id, {meta.srcAddr, meta.lkp_flag2});
```
and disallow this P4 program.
```
mirror.emit(meta.session_id, {meta.srcAddr, meta.lkp_flag2});
```

I tried to use the
`MethodInstance->actualMethodType->typeParameters` to find out if the type has been provided by the programmer and got confused when I found the vector was empty for the first case. But it makes sense now.

Thanks, Nate! Clearly I have put the class material into practice. =P.

PR #1858, opened 2019-04-15
do not remove extern method type parameter during type inference

hanw:
This is supposed to fix the issue in https://github.com/p4lang/p4c/issues/1856

Unfortunately, the fix does not work with extern defined as

```
extern Hash<W> {
   Hash(HashAlgorithm_t algo, CRCPolynomial<_> poly);
}
```

```
  what():  In file: /home/hanwang/bf-p4c-release/p4c/frontends/p4/typeChecking/typeSubstitutionVisitor.cpp:43
Compiler Bug: /home/hanwang/bf-p4c-release/build/p4c/p4include/tofino.p4(470): cannot replace a type parameter T/39 with _:
```

### DIFF
--- a/frontends/p4/typeChecking/typeChecker.cpp
+++ b/frontends/p4/typeChecking/typeChecker.cpp
@@ -293,7 +293,7 @@ const IR::Type* TypeInference::specialize(const IR::IMayBeGenericType* type,
 
     LOG2("Translation map\n" << bindings);
 
-    TypeVariableSubstitutionVisitor tsv(bindings);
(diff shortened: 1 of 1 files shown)


hanw:
It seems the problem the currently `Type_Dontcare` is a subclass of `Type_Base` and Type_Base represents the base type defind in the spec.
https://github.com/p4lang/p4c/blob/master/ir/type.def#L176

But in the p4-16 spec, base type does not include `dontcare`. Should `dontcare` be a `Type_Var`?

PR #1859, opened 2019-04-16
Added support for assert and assume statement in bmv2 backend (#1548)

enisah:
* Added removeAssertAssume pass in midend which will remove
 all assert and assume statements from program when --ndebug = true

* Not supported in ebpf backend yet

Signed-off-by: Enisa Hadzic <enisa.hadzic@rt-rk.com>

### DIFF
--- a/backends/bmv2/common/extern.cpp
+++ b/backends/bmv2/common/extern.cpp
@@ -253,4 +253,39 @@ ExternConverter::convertHashAlgorithm(cstring algorithm) {
     return result;
 }
 
+ExternConverter_assert ExternConverter_assert::singleton;
+ExternConverter_assume ExternConverter_assume::singleton;
+
+Util::IJson*
+ExternConverter::convertAssertAssume(ConversionContext* ctxt,
+    const IR::MethodCallExpression* methodCall, const P4::ExternFunction* ef) {
+     if (methodCall->arguments->size() != 1) {
+        ::error("Expected 1 arguments for %1%", methodCall);
+        return nullptr;
+    }
+    auto primitive = mkPrimitive(ef->method->name.name);
+    auto parameters = mkParameters(primitive);
+    auto cond = methodCall->arguments->at(0);
+    // wrap expression in an additional JSON expression block
+    // cast the result of expression to b2d
+    auto jsonExpr = ctxt->conv->convert(cond->expression, true, true, true);
+    parameters->append(jsonExpr);
+    primitive->emplace_non_null("source_info", methodCall->sourceInfoJsonObj());
+    return primitive;
+}
+
+Util::IJson* ExternConverter_assert::convertExternFunction(
+    UNUSED ConversionContext* ctxt, UNUSED const P4::ExternFunction* ef,
+    UNUSED const IR::MethodCallExpression* mc, UNUSED const IR::StatOrDecl* s,
+    UNUSED const bool emitExterns) {
+    return convertAssertAssume(ctxt, mc, ef);
+}
+
+Util::IJson* ExternConverter_assume::convertExternFunction(
+    UNUSED ConversionContext* ctxt, UNUSED const P4::ExternFunction* ef,
+    UNUSED const IR::MethodCallExpression* mc, UNUSED const IR::StatOrDecl* s,
+    UNUSED const bool emitExterns) {
+    return ExternConverter::convertAssertAssume(ctxt, mc, ef);
+}
+
 }  // namespace BMV2
--- a/frontends/p4/reservedWords.cpp
+++ b/frontends/p4/reservedWords.cpp
@@ -25,7 +25,7 @@ std::set<cstring> reservedWords = {
     "else", "enum", "error", "exit", "extern", "false", "header", "header_union",
     "if", "in", "inout", "int", "key", "match_kind", "out", "parser", "package",
     "return", "select", "set", "state", "struct", "switch", "table", "this",  // experimental
-    "transition", "true", "tuple", "typedef", "varbit", "verify", "void", "_",
+    "transition", "true", "tuple", "typedef", "varbit", "verify", "void", "_", "assert", "assume",
     "NoAction"  // core.p4
 };
 
--- a/frontends/p4/reservedWords.cpp
+++ b/frontends/p4/reservedWords.cpp
@@ -25,7 +25,7 @@ std::set<cstring> reservedWords = {
     "else", "enum", "error", "exit", "extern", "false", "header", "header_union",
     "if", "in", "inout", "int", "key", "match_kind", "out", "parser", "package",
     "return", "select", "set", "state", "struct", "switch", "table", "this",  // experimental
-    "transition", "true", "tuple", "typedef", "varbit", "verify", "void", "_", "assert", "assume",
+    "transition", "true", "tuple", "typedef", "varbit", "verify", "void", "_",
     "NoAction"  // core.p4
 };
 
--- a/backends/bmv2/common/extern.h
+++ b/backends/bmv2/common/extern.h
@@ -70,6 +70,9 @@ class ExternConverter {
     cstring createCalculation(ConversionContext* ctxt, cstring algo, const IR::Expression* fields,
                               Util::JsonArray* calculations, bool usePayload, const IR::Node* node);
     static cstring convertHashAlgorithm(cstring algorithm);
+    Util::IJson*
+    convertAssertAssume(ConversionContext* ctxt, const IR::MethodCallExpression* methodCall,
+                        const P4::ExternFunction* ef);
 };
 
 #define EXTERN_CONVERTER_W_FUNCTION_AND_MODEL(extern_name, model_type, model_name)  \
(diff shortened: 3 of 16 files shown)


frontends/p4/reservedWords.cpp
```diff
@@ -26,7 +26,7 @@ std::set<cstring> reservedWords = {
     "if", "in", "inout", "int", "key", "match_kind", "out", "parser", "package",
     "return", "select", "set", "state", "struct", "switch", "table", "this",  // experimental
     "transition", "true", "tuple", "typedef", "varbit", "verify", "void", "_",
-    "NoAction"  // core.p4
+    "NoAction", "assert", "assume"  // core.p4
```
mbudiu-vmw: I don't see any changes to core.p4.

backends/bmv2/common/extern.cpp
```diff
@@ -252,4 +252,39 @@ ExternConverter::convertHashAlgorithm(cstring algorithm) {
     return result;
 }
 
+ExternConverter_assert ExternConverter_assert::singleton;
+ExternConverter_assume ExternConverter_assume::singleton;
+
+Util::IJson*
+ExternConverter::convertAssertAssume(ConversionContext* ctxt,
```
mbudiu-vmw: What will bmv2 do with this call? Is there a corresponding primitive implemented?

enisah: Yes, I have made a new PR in behavioral-model for adding assert/assume primitives in the simulation.

mbudiu-vmw: Are these changes already merged?

mbudiu-vmw:
@jafingerhut : you should take a look at this. We should probably also discuss this at the language design working group; the question is where to put the declarations for assert and assume.

backends/bmv2/common/extern.cpp
enisah: I think they are not, I just made this as a separate PR from previous one, because this one does not include support for assert/assume in ebpf backend yet.

jafingerhut:
I was able to invoke `p4c-bm2-ss` with the new `--ndebug` option and it did correctly leave assert and assume statements out of the MidEndLast output file, as well as not having them in the BMv2 JSON file.  That is all good.

I could not find a way to invoke `p4c` and have it recognize the `--ndebug` option.  For example, this attempted command gives an error:
```
$ p4c --ndebug --target bmv2 --arch v1model assert-assume-try1.p4
usage: p4c [-h] [-V] [-v] [-###] [-Xpreprocessor <arg>] [-Xp4c <arg>]
           [-Xassembler <arg>] [-Xlinker <arg>] [-b TARGET] [-a ARCH] [-c]
           [-D PREPROCESSOR_DEFINES] [-E] [-e] [-g] [-I SEARCH_PATH] [-o PATH]
           [--p4runtime-file P4RUNTIME_FILE]
           [--p4runtime-files P4RUNTIME_FILES]
           [--p4runtime-format {binary,json,text}] [--target-help] [-S]
           [--std {p4-14,p4-16}] [-T LOG_LEVELS] [--top4 PASSES]
           [--dump DUMP_DIR] [--toJson JSON] [--fromJson JSON_SOURCE]
           [--pp PRETTY_PRINT]
           [source_file]
p4c: error: unrecognized arguments: --ndebug
```
It would be nice if this option could be passed through from p4c to the various back ends.  I do not know what Calin @cc10512 and others have in mind for rules of passing through such options to back ends that do not support it, and whether the p4c front end needs to keep track of which back ends support which options.

jafingerhut:
I did try this out with the v1model back end, and ran my test program on simple_switch with this PR's changes applied: https://github.com/p4lang/behavioral-model/pull/762 and it all seems to work as desired.  simple_switch aborted immediately with a descriptive message in the log when either an assert or an assume expression evaluated to false, but continued on when they evaluate to true, as desired.

It was also nice to see that these statements can be used within a parser, not only within a control.  Cool.

jafingerhut:
@mbudiu-vmw Are there any legal P4_16 expressions that evaluate to a boolean that can have side effects within them that you know of?  e.g. embedded assignment statements a la C/C++?

Hmm, I guess extern function calls could have side effects, and be used within an expression that evaluates to a boolean, or any other type.  I guess we just need to warn people, as we do for C/C++, that assert/assume expressions should ideally have no side effects.

jafingerhut:
Regarding where assert/assume are defined, are there any other considerations other than the following?

Approach 1: Add assert/assume definitions to core.p4
Advantages:
+ All P4_16 architectures have them available, without having to define them in the architecture's #include file.  They each have some custom work to do for their back ends and the target, of course.
Disadvantages:
+ User programs using the names assert/assume must be changed to avoid the name conflict.

Approach 2: Add assert/assume to every architecture-specific include file that wants to support them, e.g. v1model.p4, psa.p4, ebpf.p4, etc.
Advantages:
+ Not clear to me what the advantages are.  If assert/assume are added to every architecture a P4 developer actually wants to use (even if not all of them), then as far as they are concerned, it is no better or worse than if they were defined in core.p4
Disadvantages:
+ Not a big deal, but their definitions must be copied and pasted into each of a few architecture include files.

Did I miss anything above?

If not, it seems slightly preferable to me to add them to core.p4, but that is only a slight preference.

backends/bmv2/common/extern.cpp
jafingerhut: For reference, the corresponding changes to bmv2 are in this PR: https://github.com/p4lang/behavioral-model/pull/762

Antonin has suggested some modifications to that PR before he would be willing to accept it.

enisah:
Hello everyone, I have updated https://github.com/p4lang/behavioral-model/pull/762 PR. The changes about core.p4 that you  mentioned were in the previous version of PR and @mbudiu-vmw suggestion was to change it, at least for a while. According to that suggestion I have made changes and post it like this. Let me know which approach you prefer and  about all other suggestions that you may have.

jafingerhut:
I do not see any updates to the https://github.com/p4lang/behavioral-model/pull/762 PR.  It still has only the 1 commit that it started with.

enisah:
Yes, it still has the only 1 commit, because I did amend and force push on the previous one.

jafingerhut:
Just a note that the corresponding PR for behavioral-model to add support for assert and assume primitive operations in the BMv2 JSON file has been merged in now: https://github.com/p4lang/behavioral-model/pull/762

Unless there are other issues with these changes, the only open question I am aware of is whether the definitions of the new extern functions assert and assume belong in the architecture include files v1model.p4 and psa.p4 (and any other architecture-specific include files created for P4_16 in the future), or whether they should go into the core.p4 include file.

That question seems best to bring up at a P4 language design work group meeting.

frontends/p4/reservedWords.cpp
```diff
@@ -25,7 +25,7 @@ std::set<cstring> reservedWords = {
     "else", "enum", "error", "exit", "extern", "false", "header", "header_union",
     "if", "in", "inout", "int", "key", "match_kind", "out", "parser", "package",
     "return", "select", "set", "state", "struct", "switch", "table", "this",  // experimental
-    "transition", "true", "tuple", "typedef", "varbit", "verify", "void", "_",
+    "transition", "true", "tuple", "typedef", "varbit", "verify", "void", "_", "assert", "assume",
```
mbudiu-vmw: If these aren't in core.p4, then they are not reserved.

enisah:
Hello everyone :) 
 Did you decide to extern functions assert and assume belong in the architecture include files v1model.p4 and psa.p4 or should I move them into core.p4? Please let me know what is your decision.

jafingerhut:
I did bring up that topic during the meeting, and unfortunately I do not recall a clear cut answer to the question, except perhaps to merge this in as is, and decide later whether to put assert and assume into a P4_16 module (also discussed at the meeting, but not finished designing them yet) or core.p4.

@mbudiu-vmw @jnfoster Do you have any notes from the meeting on a decision that would help advance this PR to merging?

mbudiu-vmw:
I recall we settled for v1model...

jnfoster:
I'm very sorry but my minutes for Monday's meeting are quite a bit more sparse than normal. I got interrupted partway through (perhaps you heard the background noise before I muted myself) and I also lost some state in my editor. I tried to reconstruct the topics we discussed but it basically only records the list of items discussed, not the details of the discussion or the action items.

jafingerhut:
If v1model.p4 and psa.p4, then this PR is already doing that, so it is fine in that regard.

I am not aware of anything else holding this back, but could be forgetting something.

frontends/p4/reservedWords.cpp
mbudiu-vmw: I think these should be removed from this list.

jnfoster:
Alternatively, we could keep them reserved?

-N

mbudiu-vmw:
We should check that reserving them may not break programs that may have been using these names.

cc10512:
+1 for reserving them. Even if they are used as pragmas, they would be in a different context, so we should be able to disambiguate in the grammar.

jnfoster:
We can also reserve in the spec for 1.2 but not in the compiler until
later...

-N

enisah:
Hello everyone,
Just want to check what is the final decision? Should I remove them from reserved or leave it like this?

jafingerhut:
Sorry, I am not necessarily helping answer your question here.  What is the effect in p4c of putting something into reservedWords vs. not putting it there?

mbudiu-vmw:
Travelling right now, so it's not easy to check.
We can reserve these at anytime, and I think that the PR will work just fine without these being reserved.

enisah:
I removed it from reserved keywords, so it can be added at anytime. Well, as I understood when it is in reservedWords it is added used names in reference map when it is constructed, but it also works without being reserved so I made a commit that removes it.

PR #1862, opened 2019-04-18
Make changes to 3 PSA tests that are causing the build to fail

jafingerhut:
Two of them are passing now, and should be removed from the XFAIL list.

One of them was written in a way that it was passing, but it should
not have been.  Updated the STF test so it fails.  A separate PR on
behavioral-model code is required to make that test pass.

### DIFF
--- a/testdata/p4_16_samples/psa-unicast-or-drop-bmv2.stf
+++ b/testdata/p4_16_samples/psa-unicast-or-drop-bmv2.stf
@@ -7,6 +7,9 @@ expect 2 000000000002 000000000000 ffff
 packet 4 000000000003 000000000000 ffff
 expect 3 000000000003 000000000000 ffff
 
-# This packet should be dropped
-packet 2 000000000000 000000000000 ffff
+# This packet should be dropped.  We send it into port 0, because if
+# there are no packets sent into nor expected on port 0, then the test
+# infrastructure does not check any of the packets that come out port
+# 0, or whether the right number come out port 0.
+packet 0 000000000000 000000000000 ffff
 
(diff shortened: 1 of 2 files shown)


jafingerhut:
@derekso1 FYI

jafingerhut:
I believe that until this PR is merged, all other PRs will fail the automated tests.

Until a recent change made to behavioral-model for implementing more of the PSA architecture, several STF tests for PSA programs in p4c were marked as XFAIL, and with that recent behavioral-model change several of them are working now, so should no longer be XFAIL.

testdata/p4_16_samples/psa-unicast-or-drop-bmv2.stf
```diff
@@ -7,6 +7,9 @@ expect 2 000000000002 000000000000 ffff
 packet 4 000000000003 000000000000 ffff
 expect 3 000000000003 000000000000 ffff
 
-# This packet should be dropped
-packet 2 000000000000 000000000000 ffff
+# This packet should be dropped.  We send it into port 0, because if
+# there are no packets sent into nor expected on port 0, then the test
+# infrastructure does not check any of the packets that come out port
+# 0, or whether the right number come out port 0.
```
peteli3: Ahh we were unaware of this. So it would've simply passed since the check wouldn't have happened anyway. Is this specific to port 0; do you know if there are other ports that have this behavior?

peteli3:
added a question to Andy's comment in CMakeLists.txt
but lgtm

testdata/p4_16_samples/psa-unicast-or-drop-bmv2.stf
jafingerhut: This is not specific to port 0.  If there is a port number mentioned in any packet or expect statement of an STF file, then the test infrastructure will already do complete checking of every packet that comes out of that port, or does not come out of that port but is expected to.

If a port number is mentioned in no packet or expect statement of an STF file at all, the test infrastructure will do absolutely no checking of any packets that come out of those ports, whether 0 packets come out, or 1000.

PR #1861, opened 2019-04-18
Add STF test exercising new ParserInvalidArgument parser error

jafingerhut:

PR #1865, opened 2019-04-19
Forbid 'last' and 'next' outside parser; fixes #1853

### LINKED ISSUE #1853: Using last outside the parser.
I was using last for a header_stack on p4_14 outside the parser, what by the specification should not be possible, but the code has compiled well, with no erros from the compiller, but i was gettin a error at the setup of the switch at mininet (probably because i was with a invalid code)... So, i think there is a bug in the compiler, because the code should't be copilled in the first place.

mbudiu-vmw:


mbudiu-vmw:
This build has failed because of a different PR that we have merged #1861, and it has only failed on MacOS. I wish I understood how these computers work...

PR #1868, opened 2019-04-20
 Make an extra copy when source and destination of copy are aliased 

### LINKED ISSUE #1863: Structure flattening pass is probably buggy
I suspect it does not handle correctly this case:
```
struct S { bit a; bit b; }
S s = { 0, 1 };
s = { s.b, s.a };
```

mbudiu-vmw:
Fixes #1863 
I have moved out a pass from sideEffects to its own file in order to reuse it.


jafingerhut:
I am no good for reviewing the compiler code changes, but I can confirm that this does fix the bugs in the generated code for the kinds of test case added, and a similar one I ran independently on my own.

frontends/p4/alias.h
```diff
@@ -0,0 +1,248 @@
...
+#include "frontends/common/resolveReferences/referenceMap.h"
+#include "frontends/p4/typeChecking/typeChecker.h"
+#include "frontends/p4/methodInstance.h"
+
+namespace P4 {
+
+/// This class represents the path to a location.
+/// Given a struct S { bit a; bit b; } and a variable S x;
+/// a path can be x.a, or just x.  An array index is represented as a
+/// number (encoded as a string) or as "*", denoting an unknown index.
```
cc10512: What about a slice?

mbudiu-vmw:
This is a visitor; it is a pretty complicated recursive definition, so I think it belongs in a visitor rather than as a method in the IR. Also, this is a conservative analysis, we have a more precise one which can be used when you have types. This alias analysis only works for expressions that occur within the same statement. I will add a comment to make it clear.

frontends/p4/alias.h
mbudiu-vmw: I don't see slices handled, I should probably handle these as well. Currently this will approximate a slice with the whole value that is sliced.

mbudiu-vmw: Actually slice is handled by Operation_Ternary.

frontends/p4/alias.h
```diff
@@ -0,0 +1,255 @@
...
+        for (auto p : path)
+            out << "." << p;
+    }
+};
+
+/// We represent a set of location set as a set of LocationPath
+/// objects.
+class SetOfLocations : public IHasDbPrint {
+ public:
+    std::set<const LocationPath*> paths;
```
ChrisDodd: nit: is there a need for the indirection here, or could this be `std::set<LocationPath>` instead?  Is a single LocationPath object ever required in two sets such that it being the same object instead of a copy important?

frontends/p4/alias.h
```diff
@@ -0,0 +1,255 @@
...
+
+    void postorder(const IR::Operation_Ternary* expression) override {
+        auto e0 = ::get(rw, expression->e0);
+        auto e1 = ::get(rw, expression->e1);
+        auto e2 = ::get(rw, expression->e2);
+        rw.emplace(expression, e0->join(e1)->join(e2));
+    }
+
+    void postorder(const IR::Slice* expression) override {
+        auto e = ::get(rw, expression->expr);
```
ChrisDodd: I'm confused how this even compiles?  IR::Slice has no field called `expr`...

mbudiu-vmw: well, it didn't...
sorry about that.

frontends/p4/alias.h
mbudiu-vmw: No, we don't even test the LocationPaths for equality really; there may be probably some duplicates. This should not affect correctness. So we can probably use a vector instead of a set.

PR #1871, opened 2019-04-20
Fix Issue 1867

familysingh:


familysingh:
Due to churn of in CMakeLists.txt, Travis failed.  The failure is not seen once the churn has been fixed.

PR #1883, opened 2019-04-23
Fix #1881 - enhance BMv2 backend, support const entries range with 1 …

### LINKED ISSUE #1881: Enhancement: enable BMv2 const entries with range field having exact match value
Currently the BMv2 back end supports a field specified with a single exact value for ternary, lpm, and exact match fields, but not for range.  That is, this works:
```
    table t1 {
 	key = { hdr.h1.f1 : ternary; }
        actions = { a; }
        const entries = {
            0xfe  : a(3);
           _      : a(9);
        }
    }
```
but it gives an error if you change `ternary` to `range`.

This is a small easy enhancement, for which I will soon create a PR.

jafingerhut:
…value


jafingerhut:
Slightly modifies one test program to exercise this new capability.  Added '$' to the end of several STF tests that check const entries of various flavors, to make their packet content checking more strict.

PR #1884, opened 2019-04-24
Delete useless PassManagers; fixes #1876

### LINKED ISSUE #1876: Core dump on trivial program
Compiling this program
```
#include<core.p4>
#include<v1model.p4>
struct H {}
struct M {}
parser P(packet_in pkt, out H h, inout M m, inout standard_metadata_t meta) { state start { transition accept; } }
control G(inout H h, inout M m, inout standard_metadata_t meta) { apply { } }
control C(inout H h, inout M m) { apply{ } }
control D(packet_out pkt, in H h) { apply { } }
V1Switch(P(), C(), G(), G(), C(), D()) main;
```
dumps core using the Bmv2 back-end. This behavior can be seen with the current version of P4C in GitHub (f8876762cdc76c37c88afa82ea56fb6a04c76861).

mbudiu-vmw:
@hanw @jnfoster The bug was caused by assuming that each control block is unique, and using the control blocks as keys in a hashmap. In this program the same control block is used for multiple purposes. Fortunately the hashmap wasn't really necessary. 

The PSA back-end may have the same bug, though.

### DIFF
--- a/backends/bmv2/simple_switch/simpleSwitch.cpp
+++ b/backends/bmv2/simple_switch/simpleSwitch.cpp
@@ -44,7 +44,7 @@ bool ParseV1Architecture::preorder(const IR::PackageBlock* main) {
         modelError("%1%: main package  match the expected model", main);
         return false;
     }
-    structure->block_type.emplace(prsr->to<IR::ParserBlock>()->container, V1_PARSER);
+    structure->parser = prsr->to<IR::ParserBlock>()->container;
 
     auto ingress = main->findParameterValue(v1model.sw.ingress.name);
     if (ingress == nullptr || !ingress->is<IR::ControlBlock>()) {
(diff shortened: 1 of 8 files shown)


jafingerhut:
I have poked Travis to force it to re-test this PR, after merging in the changes that allow all PSA tests to pass again.

antoninbas:
I think the PSA backend also has that block_type map (and it may also be unused...)
However, I don't think the PSA architecture allows for the same control block to be used for multiple purposes, unlike v1model. All the parsers / controls / deparsers have different signatures IIRC.

jafingerhut:
Yes, no pair of the 6 top-level 'blocks' in PSA have the same signature.  Ingress and Egress controls differ in the types of some of their parameters, unlike v1model.

PR #1891, opened 2019-04-24
Move updating of search_tail; fixes #1879

### LINKED ISSUE #1879: Incorrect handling of output parameter when using subparser
To simplify some code that I'm writing I started to use a subparser. The subparser currently consists of only one state and looks roughly as follows:
```
parser SubParser(packet_in packet, out subheader_t sub, inout metadata meta, in bit<8> selectedPos) {
    state start {
        packet.extract(sub);

        bool currentFieldSelected = selectedPos == meta.currPos;
        meta.currentVal.value = meta.currentVal.value + (bit<1>)currentFieldSelected * sub.value;
        meta.currPos = meta.currPos + 1;
        transition accept;
    }
}
```
When compiling this code seems to be simplified in an incorrect way however. When I run this code with `simple_switch` and `currentFieldSelected` evaluates to true, I see the following line in the debug output:
```
Parser set: setting field 'meta_0_currentVal.value from expression, new value is 1
```
This is correct, but then a bit later (presumably at the end of the subparser) there is the following line:
```
Parser set: setting field 'scalars.metadata._currentVal_value7' from expression, new value is 0
```
Instead of copying the previously computed value, the expression seems to be evaluated again, resulting in an incorrect value (as meta.currPos was updated in the meantime).

If i move the condition outside the parser and pass it as an argument, it does function correctly and I get the following line in the output instead of the one above:
```
Parser set: setting field 'scalars.metadata._currentVal_value7' from field 'meta_0_currentVal.value' (1)
```
So now the computed value is actually assigned and the expression not evaluated again.

jderuiter:
Move updating of `search_tail` in `exprUses` to `preorder` and `revisit` of PathExpression instead of Path.

If Path is ever used outside of a PathExpression this might break, as then `search_tail` would not be updated, but I'm not sure this ever happens.

### DIFF
--- a/midend/expr_uses.h
+++ b/midend/expr_uses.h
@@ -28,16 +28,24 @@ class exprUses : public Inspector {
     cstring look_for;
     const char *search_tail = nullptr;  // pointer into look_for for partial match
     bool result = false;
-    bool preorder(const IR::Path *p) override {
-        if (look_for.startsWith(p->name)) {
-            search_tail = look_for.c_str() + p->name.name.size();
+
+    void set_search_tail(const IR::PathExpression *e) {
+        if (look_for.startsWith(e->path->name)) {
+            search_tail = look_for.c_str() + e->path->name.name.size();
             if (*search_tail == 0 || *search_tail == '.' || *search_tail == '[')
-                result = true; }
-        return !result; }
+                result = true; } }
+
+    bool preorder(const IR::Path *) override { return !result; }
     bool preorder(const IR::Primitive *p) override {
         if (p->name == look_for) result = true;
         return !result; }
     bool preorder(const IR::Expression *) override { return !result; }
+    bool preorder(const IR::PathExpression *e) override {
+        set_search_tail(e);
+        return !result; }
+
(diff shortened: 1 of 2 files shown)


midend/expr_uses.h
```diff
@@ -28,16 +28,24 @@ class exprUses : public Inspector {
...
+    bool preorder(const IR::Path *) override { return !result; }
     bool preorder(const IR::Primitive *p) override {
         if (p->name == look_for) result = true;
         return !result; }
     bool preorder(const IR::Expression *) override { return !result; }
+    bool preorder(const IR::PathExpression *e) override {
+        set_search_tail(e);
+        return !result; }
+
+    void revisit(const IR::PathExpression *e) override {
```
ChrisDodd: would be simpler to just add `visitAgain()` to the preorder(PathExpression) method.  Add that and a call to `set_search_tail` to the preorder(Path) method and it would work for paths that are not under PathExpressions (if there are any).

jderuiter:
Thanks for your input! It seems your suggestion is now already implemented in #1895

PR #1901, opened 2019-04-26
ordered_set::operator<

ChrisDodd:
- allows sets of ordered_sets or ordered_sets as map keys in a sensible way.
  Compares the underlying sets independent of the order, so that it is
  possible to have ordered_sets where !(a < b) && !(b < a) && !(a == b).
  Such sets contain the same elements but in a different order.

### DIFF
--- a/lib/ordered_set.h
+++ b/lib/ordered_set.h
@@ -24,7 +24,7 @@ limitations under the License.
 #include <set>
 #include <utility>
 
-// Remembers items intertion order
+// Remembers items in insertion order
 template <class T, class COMP = std::less<T>, class ALLOC = std::allocator<T>>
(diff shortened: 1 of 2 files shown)


lib/ordered_set.h
```diff
@@ -77,6 +77,18 @@ class ordered_set {
     ordered_set &operator=(ordered_set &&a) = default; /* move is ok? */
     bool operator==(const ordered_set &a) const { return data == a.data; }
     bool operator!=(const ordered_set &a) const { return data != a.data; }
+    bool operator<(const ordered_set &a) const {
```
mbudiu-vmw: you could add a comment saying that this is a lexicographic comparison

Issue #1900, opened 2019-04-26
gtestp4c infinite loop ?


jafingerhut:
On an Ubuntu 18.04 Linux system with latest updates as of 2019-Apr-26, but no other packages installed on it, and `p4c` source code with this version:
```
commit ae6fb228a0d307466f482affd2b47fbd079f3c9e
Author: Mihai Budiu <mbudiu@vmware.com>
Date:   Thu Apr 25 16:46:31 2019 -0700
```
I tried installing behavioral-model, p4c, and the packages that they depend upon, and then running p4c's `cd build ; make check`.  It reached the point in the output shown below of running `./test/gtestp4c`, and then continued using CPU for over 30 minutes without making further progress.
```
[----------] 1 test from FromJSONTest
[ RUN      ] FromJSONTest.load_ir_from_json
```
Later I can add a link to the full build logs on this machine where I tried it, and then try reproducing it with another Ubuntu 18.04 VM from scratch to see if it is reproducible.

jafingerhut:
I have also seen this with same version of p4c on an Ubuntu 16.04 Linux system (also fresh install, also with latest updates as of as of 2019-Apr-26).

Same commit of p4c involved.

I did not see gtestp4c running a long time like this the last time I ran the same install script to install p4c (this one https://github.com/jafingerhut/p4-guide/blob/master/bin/install-p4dev-p4runtime.sh ) when I ran it, then ran `cd build ; make check` on 2019-Apr-21, with a slightly earlier version of behavioral-model and p4c being installed then.

mbudiu-vmw:
This test forks separate processes and waits for them. We should probably disable it until we figure out what's wrong.

PR #1899, opened 2019-04-26
Updates to p4c automated tests to handle recent changes in psa_switch

jafingerhut:

PR #1902, opened 2019-04-26
Temporarily disable test which hangs

mbudiu-vmw:


mbudiu-vmw:
This addresses temporarily #1900

mbudiu-vmw:
I just submitted a version which uses system.

mbudiu-vmw:
I believe that this test failed because we need to merge #1899.
All this churn in the behavioral model makes it hard to synchronize. Is there a better solution - perhaps based on versioning the compiler and simulator to keep them in sync - or at least fail quickly?

jafingerhut:
I think the main reason we have recently had several test failures in p4c because of changes in behavioral-model is due to the ongoing PSA implementation work, _plus_ the fact that several XFAIL tests were checked into p4c that were waiting on updates to behavioral-model to work.

If we instead work in this order:

+ update behavioral-model for a new feature
+ only then check in the PSA test programs that rely on the new behavioral-model feature, not as XFAIL, but as a normally working test

then this breakage should not happen as often as it has for the last week or two.

There are not many XFAIL tests for PSA left, but if we instead of making them XFAIL could leave them in the testdata directory, but somehow mark them as DONT_RUN_YET (I think there is at least one *_EXCLUDE symbol in the CMakeList.txt files somewhere for this?), then p4c test results would change less often for this reason.

jafingerhut:
I have restarted the Travis job for this PR, which will likely pass this time (or at least should not fail because of the behavioral-model commit earlier today).

PR #1904, opened 2019-04-27
Add resubmit stf test

kevinye2:


peteli3:
This test requires resubmit support in bmv2: https://github.com/p4lang/behavioral-model/pull/773

kevinye2:
@jafingerhut 
We added the old test that looped for resubmit; before it was pointed out that it should fail due to the specifications.

jafingerhut:
@kevinye2 Which P4 program do you believe should have been failing according to the specifications?  You may be perfectly correct in that -- just wanted to double check to see if we are on the same page about what the PSA specification requires.

jafingerhut:
OK, reading the test program in this PR again and thinking about it more clearly, yes, the test program psa-resubmit-loop-bmv2.p4 should resubmit a packet forever, if it resubmits it when it first arrives from an input port.  What does the test infrastructure do with that test?  Run it for some maximum amount of time and then kill it?

peteli3:
@jafingerhut 
Yep, it keeps going for a period of time before stopping. Probably not a very useful/portable metric but we observe that runs until ~ 40k iterations before stopping.

jafingerhut:
I think that once resubmit is working in psa_switch, a more useful test than something that intentionally goes into an infinite loop, would be to modify the working test case so that it intentionally tries to modify some header field of a packet during ingress, and verify via the STF expected packets that these modifications _do not_ make it into the packet that eventually goes out of an output port.  That would test the same desirable behavior for resubmit that the infinite loop test does, without having an infinite loop.

peteli3:
Agreed. I will write up the test and push it tonight.

jafingerhut:
@peteli3 Looks like Github believes this has conflicts that need to be resolved before it can be merged.

Also, now that the behavioral-model code should implement resubmitted packets correctly, the program testdata/p4_16_samples/psa-resubmit-loop-bmv2.p4 should no longer be in the XFAIL list, should it?

jafingerhut:
Never mind about my comment that testdata/p4_16_samples/psa-resubmit-loop-bmv2.p4 should no longer be on the XFAIL list.  It should be, since the packet should resubmit in a loop infinitely.

My earlier comment here still applies: https://github.com/p4lang/p4c/pull/1904#issuecomment-489910049

peteli3:
> Never mind about my comment that testdata/p4_16_samples/psa-resubmit-loop-bmv2.p4 should no longer be on the XFAIL list. It should be, since the packet should resubmit in a loop infinitely.
> 
> My earlier comment here still applies: [#1904 (comment)](https://github.com/p4lang/p4c/pull/1904#issuecomment-489910049)

This latest commit adds your suggestion to the existing resubmit test!

peteli3:
Please ignore all my subsequent commits. 

It appears I've made a mess by attempting to rebase this branch using master. I'm just going to make a new PR with these 2 tests. @jafingerhut

jafingerhut:
Since the new test was added to this PR: https://github.com/p4lang/p4c/pull/1959  I am closing this PR.

PR #1905, opened 2019-04-27
add support for printing help for pragmas

cc10512:
Printing help is a backend specific option. It requires a bit more
flexible parsing of annotations (i.e., registering handlers outside
the constructor) and a bit of driver tweaking.

### DIFF
--- a/tools/driver/p4c_src/main.py
+++ b/tools/driver/p4c_src/main.py
@@ -136,11 +136,11 @@ def main():
                         "description (default is binary). "
                         "[Deprecated; use '--p4runtime-files' instead]",
                         action="store", default="binary")
-    parser.add_argument("--pragmas-help", "--help-pragmas", "--help-pragma", "--help-annotations",
-                        "--annotations-help",
+    parser.add_argument("--help-pragmas", "--pragma-help", "--pragmas-help",
+                        "--help-annotations", "--annotation-help", "--annotations-help",
                         dest="help_pragmas", action="store_true", default=False,
                         help = "Print the documentation about supported annotations/pragmas and exit.")
-    parser.add_argument("--target-help", "--help-targets", "--help-traget",
+    parser.add_argument("--help-targets", "--target-help", "--targets-help",
                         dest="show_target_help",
                         help="Display target specific command line options.",
                         action="store_true", default=False)
--- a/tools/driver/p4c_src/main.py
+++ b/tools/driver/p4c_src/main.py
@@ -116,7 +116,7 @@ def main():
     parser.add_argument("-I", dest="search_path",
                         help="Add directory to include search path",
                         action="append", default=[])
-    parser.add_argument("-o", dest="output_directory",
+    parser.add_argument("-o", "--output", dest="output_directory",
                         help="Write output to the provided path",
                         action="store", metavar="PATH", default=".")
     parser.add_argument("--p4runtime-file",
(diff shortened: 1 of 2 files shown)


tools/driver/p4c_src/main.py
```diff
@@ -184,20 +192,30 @@ def main():
         print display_supported_targets(cfg)
         sys.exit(0)
 
+    # When using --help-* options, we don't necessarily need to pass an input file
+    # However, by default the driver checks the input and fails if it does not exist.
+    # Also, loading the backend configuration files requires a source file to be set,
+    # so in that case, we set the input to dummy.p4 and expect that the backend itself
+    # will do its printing and ignore the input file.
+    # If not, the compiler will error out anyway.
+    checkInput = True if not opts.help_pragmas else False
```
liujed: Why not just `checkInput = not opts.help_pragmas`?

tools/driver/p4c_src/main.py
```diff
@@ -184,20 +192,30 @@ def main():
...
     if opts.source_file:
         input_specified = True
     if (os.environ['P4C_BUILD_TYPE'] == "DEVELOPER"):
         if opts.json_source:
             input_specified = True
-    if not input_specified:
+    if checkInput and not input_specified:
         parser.error('No input specified.')
+    elif not checkInput:
+        opts.source_file = "dummy.p4"
```
liujed: This risks having confusing/misleading error messages. I think a better approach would be to thread `checkInput` through to the relevant places. Granted, that would be a more disruptive change.

tools/driver/p4c_src/main.py
```diff
@@ -136,14 +136,18 @@ def main():
                         "description (default is binary). "
                         "[Deprecated; use '--p4runtime-files' instead]",
                         action="store", default="binary")
-    parser.add_argument("--target-help", dest="show_target_help",
+    parser.add_argument("--pragmas-help", "--help-pragmas", "--help-pragma",
+                        dest="help_pragmas", action="store_true", default=False,
+                        help = "Print the documentation about supported pragmas and exit.")
```
liujed: P4₁₆ officially calls these annotations, not pragmas. Neither annotations nor pragmas appear in the P4₁₄ spec.

tools/driver/p4c_src/main.py
cc10512: good point.

tools/driver/p4c_src/main.py
cc10512: The changes are quite significant to handle at this level vs. local in each backend. I took the lazy approach and as backends implement this change, the confusing error message disappears.

tools/driver/p4c_src/main.py
cc10512: Even though it is not in the spec, it is in the grammar -- this is how we write annotations in P4-14:
```
v1lexer.ll:68:"@pragma"[ \t]*[A-Za-z_][A-Za-z0-9_]* {
v1lexer.ll:71:"@pragma"[ \t]* { BEGIN((driver.saveState = PRAGMA_LINE));
v1lexer.ll:72:                  return Parser::make_PRAGMA("pragma", driver.yylloc); }
v1parser.ypp:229:                                    field_or_masked_ref opt_condition pragma_operand
v1parser.ypp:232:                                    field_match_list expressions pragma_operands
v1parser.ypp:271:    | input PRAGMA pragma_operands NEWLINE
v1parser.ypp:957:pragma_operands: /* epsilon */
v1parser.ypp:959:    | pragma_operands pragma_operand
v1parser.ypp:961:    | pragma_operands "," pragma_operand
```

liujed: I guess I am advocating for consistency in what these things are called. Granted, `--annotation-help` seems relatively unwieldy. Also, if we are keeping with "pragma" over "annotation", what about adding `--pragma-help` to the list of variants?

liujed: Having seen your most recent changes, these seem the most natural to me: `--annotation-help`, `--help-annotations`, `--pragma-help`, `--help-pragmas`. Of course, it doesn't hurt to have all eight(!) variants.

cc10512: I like this. Also, it helps that args_parse will actually handle `--help-pragma` if it resolves uniquely, and it does resolve to `--help-pragmas`. So we only need 6 variants ...

tools/driver/p4c_src/main.py
```diff
@@ -166,6 +171,10 @@ def main():
     user_defined_version = os.environ.get('P4C_DEFAULT_VERSION')
     if user_defined_version != None:
         opts.language = user_defined_version
+    # accept multiple ways of specifying which language, and ensure that it is a consistent
```
mbudiu-vmw: I think this code would be even more useful in the C++ compiler.

cc10512: I can add it to the backend as well, however, for anything a bit more complicated, running the driver is much more friendly than directly running the backend. And the driver normalizes the params, so the backend arguments are insulated from changes.

Issue #1906, opened 2019-04-28
STF issues: using table key as Slice fails, syntax, etc.


hesingh:
1. STF is a simple framework and thus I am not surprised, this line of code in my STF file fails.

```
STF Command: add lb_tbl meta.meta.lb_hash[19:12]:0x0 ingress.set_egress_spec(spec:11)
OrderedDict([(u'meta.meta.lb_hash[19:12]', u'exact')])
Exception  Unexpected key field meta.meta.lb_hash[19
```
`add lb_tbl meta.meta.lb_hash[19:12]:0x0 ingress.set_egress_spec(spec:11)`

What would it take to extend STF to support bit Slice? 

2. Also, why does behavioral-model "RuntimeCmd" use ```table_add <table name> <action name> <match fields> => <action parameters> [priority]``` but STF uses a different cmd and syntax shown above?  It would have been good for STF and RuntimeCmd to use the same cmd and syntax.

3. I think STF supports "packet 0" and "expect 2" which means packet ingressed at port 0 and egressed at port 2.  It would be good to add a P4 program to p4c repo to show how this is done in P4 code.  I have a rough idea which is to get_egress_port in the egress control.  Most STF tests I have seen use port 0 for both `packet` and `expect`

hesingh:
I see dead code in the bmv2 backend related to a table key, if the key uses a bit Slice.  The dead code starts at the line below. 

https://github.com/p4lang/p4c/blob/master/backends/bmv2/common/control.cpp#L345

I am using a table key which is a bit Slice, but the code above did not get hit during testing.

For this issue, I have to make sure, the bmv2 backend is good to support a table key as bit Slice.   I don't think testdata/p4_16_sample/ includes any bmv2 test with table key as bit Slice. 

Thereafter, I can look at the behavioral-model for any code changes.

ChrisDodd:
What would such a slice mean?  When you specify a table entry, you must specify the ENTIRE key required by the table -- specifying only part of the key makes no sense?

Most STF tests use expect 2 or 3 for output as until (relatively) recently, bmv2 would not output packets on port 0 properly (something to do with internal drop handling?).  So only a few recent STF tests use expect 0

jafingerhut:
P4_16 lets you specify a bit slice expression (and many other kinds of arithmetic expressions) as a key of a table, it appears that the existing STF test infrastructure does not let you refer to this key as the name of the field of which a slice is being taken, nor does it let you specify the slice expression, either.

[...]

PR #1910, opened 2019-04-28
Add test program with bit slices and bitwise AND for table key fields

jafingerhut:


jafingerhut:
@ChrisDodd @antoninbas  Here is a sample P4_16 program without @name annotations on two of its search key fields, one a bit slice expression, the other a bitwise AND expression of a field and a constant.

p4c today gives no error or warning messages when compiling such a program, and auto-generates @name annotations somewhere in the front end of the compiler that textually look like the expressions, i.e. "field[5:1]" for the slice expression, and "field & 0x11" for the bitwise AND expression.

Currently p4c gives an error message if you have an expression like "field - 1" as a search key field, saying that you must create a @name annotation for it.

Is that what should happen for bit slices and bitwise AND expressions, too?  Or only one of those cases?

jafingerhut:
According to a comment from Mihai on the issue that caused me to ask the questions above: https://github.com/p4lang/p4c/issues/1906#issuecomment-487650340

there are no warnings or errors for P4_16 table key expressions that are supported in P4_14 and auto-translated to P4_16.  P4_14 doesn't have bit slices.  It apparently does have bitwise AND.

mbudiu-vmw:
"mask" expressions in p4-14 are translated to bitwise ands.
The important question to answer is whether control-plane API generation tools (including the STF tests, which are a crude replacement for a control-plane) can handle all these shapes of name annotations.

mbudiu-vmw:
I will wait for merging this until we get an answer to the above question.

ChrisDodd:
There's no STF (or other) test included with this, so its hard to know...

jafingerhut:
I know that the currently-used STF test infrastructure gives an error if you attempt to add a table entry with a table key field named like a slice, e.g. "field[5:2]".  I do not know whether the new not-yet-in-use-by-p4c STF test infrastructure handles such field names.

I haven't tried the currently-used STF test infrastructure with a field name like "field & 0x11", but the space characters in the name make me willing to bet $100 that it will not, given the simplistic kinds of separate-by-whitespace parsing techniques are used on STF files in the current test infra.

jafingerhut:
Note: My reason for creating this PR was not so much to get STF tests working for them.  We can make STF tests work fine with the appropriate @name annotations on such table key fields, and that is good enough for me, personally.  The new test program and STF test in this PR demonstrates this capability: https://github.com/p4lang/p4c/pull/1909

My reason for creating this PR were to ask the questions above for P4Runtime and the P4Info file that p4c generates for its use.  Should _it_ have auto-generated names like "field[5:2]" and "field & 0x11"?  Does that hurt P4Runtime in any way?  Or does it handle arbitrary sequences of characters in field names just fine?  I do not yet know.

testdata/p4_16_samples/match-on-exprs2-bmv2.p4
```diff
@@ -0,0 +1,103 @@
...
+    }
+    action foo(bit<9> out_port) {
+        hdr.ethernet.dstAddr[22:18] = hdr.ethernet.srcAddr[5:1];
+        stdmeta.egress_spec = out_port;
+    }
+    table t1 {
+        key = {
+            hdr.ethernet.srcAddr[22:18] : exact;
+            hdr.ethernet.dstAddr & 0x010101010101 : exact;
+            hdr.ethernet.etherType - 10 : exact @name("etherType_less_10");
```
antoninbas: The P4Info message is exactly what I would expect. I guess ideally the `@name` annotation on `hdr.ethernet.etherType - 10` would not be required, and name `"hdr.ethernet.etherType - 10"` would be auto-generated by default.
Given that in P4Runtime, match field names are just strings which are mapped to ids (with ids being used in P4Runtime messages), the presence of unusual characters  in the names, such as whitespaces, should not be an issue for any implementation. This is unlike the bmv2 JSON and the tools which consume it, which require the match field to be a plain field reference. But the bmv2 CLI doesn't really use these names when receiving commands from the user, instead it treats the match key as a vector of fields (and so does the bmv2 Thrift API). It seems that on the other hand STF uses names to identify match fields.

ChrisDodd:
In this case though, the autogenerated key names should be just `hdr.ethernet.srcAddr` and `hdr.ethernet.destAddr` (that's what they would be for the equivalent P4_14 code), so using just them in the STF test should work.

testdata/p4_16_samples/match-on-exprs2-bmv2.p4
antoninbas: By the way @jafingerhut, I think once this is settle, this information would be a great addition to your https://github.com/p4lang/p4runtime/blob/master/docs/v1/guidance-for-generating-p4info.md document

mbudiu-vmw:
Can we merge this?

jafingerhut:
Antonin mentioned in a comment "Once this is settled ..." here https://github.com/p4lang/p4c/pull/1910#discussion_r279465870

Before / when merging this, if there are still-open questions raised by this PR, I would hope we would create at least one other issue that leads to resolving the still-open questions.

mbudiu-vmw:
@jafingerhut can you please check that this does not need rebasing?
In general after merging changes to v1model everything needs rebasing.

jafingerhut:
@mbudiu-vmw I went ahead and rebased it and forced it to run the tests again, which passed.  There were no merge conflicts.  I believe that most PRs are not affected by v1model.p4 changes -- the only conflicts I know of are if the other PR also modifies v1model.p4, which this one does not.

Issue #1908, opened 2019-04-28
Counter/meter/register shared between controls does not work


ChrisDodd:
The attached test case used to work, but recent changes cause the compiler to give errors about the counter conflicting with itself

ChrisDodd:
[counter6.p4](https://github.com/p4lang/p4c/files/3125415/counter6.p4.txt)
[counter6.stf](https://github.com/p4lang/p4c/files/3125416/counter6.stf.txt)

jafingerhut:
The counter6-first.p4 P4_16 program created from P4_14 to P4_16 conversion has a separate control `processA` and `processB` for each of the corresponding controls in the P4_14 program, but each of those has its own independent "copy" of the counter `cntDum` instantiated within it.

From that starting point, it is already not equivalent in behavior to the P4_14 program, which has only a single counter `cntDum`.  Am I getting that right?

If so, it seems like the correction should be in the P4_14 to P4_16 translation, perhaps by having only a single instantiation of `cntDum` at the top level that both controls `processA` and `processB` call its `count` method on.

ChrisDodd:
Yes, that would be correct.  That is the way it used to work, but it changed some time fairly recently.

jafingerhut:
Do you have a version of p4c that it used to work?  Out of curiosity, I have looked for one, about every month in 2019, and about every 3 months before that in 2018 and late 2017, and every one of the versions I tried auto-translated P4_14 to P4_16 in the same way for program counter6.p4

mbudiu-vmw:
This happens in the conversion from p4-14 to p4-16.

ChrisDodd:
I've narrowed down the problem to

    commit 7fbe8d3ad56b056f544c8f5907a2227459dab05c
    Author: Antonin Bas <antonin@barefootnetworks.com>
    Date:   Tue Apr 9 09:58:10 2019 -0700

        Detect duplicate names in P4Info (#1844)
    
        * Detect duplicate names in P4Info

prior to this change, the attached program works fine with the barefoot backend (the only place we'd been testing this sort of thing previously, apparently), while afterwards it give the error

    [--Werror=duplicate] error: Name 'cntDum' is used for multiple objects in the P4Info message
    [--Werror=duplicate] error: Found 1 duplicate name(s) in the P4Info
    ./p4c FAILED

Looking into it a bit more, it seems that we have some code working around the duplication in the barefoot backend, but after this change, it fails in the frontend and never gets to the backend.

[...]

PR #1912, opened 2019-04-29
P4_14->16 convert counter/meter/register as globals

ChrisDodd:
#1908


mbudiu-vmw:
This cannot pass since some reference outputs are bound to change.
You can rerun only these tests with make check-p14_to_16

PR #1913, opened 2019-04-30
psa stf tests should extract and emit evenly, expectations should be strict

peteli3:
@jafingerhut not entirely sure this is necessary but felt like it'd be nice to fix these little things
- make all psa stf test output expectations stricter by adding a hard ending character (`$`) at the end of all `expect` lines
- the egress parsers in all our stf test switch programs don't extract anything, but the egress deparsers emit, which is uneven, and resulted in the doubly-emitted packets we observed earlier. while this change doesn't directly fix any tests, it demonstrates more "proper" behavior of a switch than what we had before


jafingerhut:
Making STF tests more strict is a good thing, in my opinion.  Without that strictness, we might not have noticed the "double headers emitted" bug for much longer.

Changing the _existing_ tests, that helped us notice these bugs, seems like a bad idea to me.  Better to leave them as they are in case someone messes up the implementation in a way similar to the way it was messed up before :-)

Adding _new_ tests that are more typical seems like a better idea to me.  You can just put a "2" at the end of the program name if it is very similar to an existing test program.

peteli3:
> Making STF tests more strict is a good thing, in my opinion. Without that strictness, we might not have noticed the "double headers emitted" bug for much longer.
> 
> Changing the _existing_ tests, that helped us notice these bugs, seems like a bad idea to me. Better to leave them as they are in case someone messes up the implementation in a way similar to the way it was messed up before :-)
> 
> Adding _new_ tests that are more typical seems like a better idea to me. You can just put a "2" at the end of the program name if it is very similar to an existing test program.

@jafingerhut 
Makes sense to me! Should I:
1) close this PR, and open a new one in which I add in my edited p4 code + stf tests as new copies named [associated-old-name]-2-bmv2.[p4 | stf]? or
2) build off this PR, leave in my p4 code edits, AND make copies of the edited p4 to go with the stricter stf tests?

(sorry for my potentially confusing wording...)

jafingerhut:
Your choice of whatever you think is easier between closing this PR and opening a new one, or modifying this PR with additional commits.

Either way, it would be best if existing test P4 programs remain unchanged, I think.  OK to add strictness of checking to existing STF tests.

jafingerhut:
FYI, you can add another commit to an existing PR that "undoes" changes in previous commits.  Github makes it easy to see the contents of individual commits, or the "combined effect" of all commits.

peteli3:
I'll probably open a new PR since it seems like merging caused some problems in this one.

Issue #1918, opened 2019-05-03
PSA back end generates incorrect action ids in BMv2 JSON


jafingerhut:
Derek So first found an example program and the cause of why it was not executing correctly in psa_switch around 2019-Apr-14.  I am just creating an issue to remember what we learned about the problem so far, with one of the simplest example P4_16 + PSA arch programs that exhibits the problem.

The BMv2 JSON file generated via this command:
```
$ p4c-bm-psa myprog.p4 -o myprog.json
```
does contain both of the actions from the source program, but the two "dummy tables" generated that cause those actions to be executed both use action id 0 to identify which action they should run, which is incorrect.  One should be action id 0 and the other 1.

Here are some additional notes from Derek after working on debugging the issue: "I did a lot of tracing through the compiler and pinpointed the problem to a single helper function which retrieves the ID of a P4 Action in a map. The map actually contains the correct mapping of ID to P4 Action, but the helper function is showing unexpected behavior.

The issue disappears when I assign a `default_action` to `table t` in which case the JSON behaves as expected"

And further notes from Derek later, when asked what causes this problem: "A helper function in p4c `get(action_map, id)` which returns the wrong action given the id.
I verified that the map is populated correctly and that the id is also correct, should be something in the internals of the `get` helper which is mis-behaving"

jafingerhut:
Attaching test program with an STF file that should pass if the program executed correctly in psa_switch, but fails now because of the bug.
[psa-issue1918-bmv2.p4.txt](https://github.com/p4lang/p4c/files/3142504/psa-issue1918-bmv2.p4.txt)
[psa-issue1918-bmv2.stf.txt](https://github.com/p4lang/p4c/files/3142505/psa-issue1918-bmv2.stf.txt)

jafingerhut:
With the proposed enhancements to the p4c PSA back end code in logging and internal error checking, proposed in this PR: https://github.com/p4lang/p4c/pull/1922

compiling the attached program using the `-T` option to give a little logging output gives the output shown below:
```
$ p4c-bm2-psa psa-issue1918-bmv2.p4  -o psa-issue1918-bmv2.json -T control:3,action:3
add action with id 0 name ingressImpl.act1 action act1() {
  ostd.drop = 0; }
add action with id 1 name ingressImpl.act2 action act2() {
  ostd.egress_port = (bit<32>)hdr.ethernet.dstAddr[2:0]; }
Processing <P4Table>(214682) tbl_act1
look up id action act1() {
[...]

PR #1922, opened 2019-05-05
Enhancements to logging and internal error checking for PSA back end

jafingerhut:


jafingerhut:
@peteli3 @derekso1 FYI on these minor proposed enhancements.  They helped me learn a little bit more about why this issue is happening: https://github.com/p4lang/p4c/issues/1918

One of the extra BUG_CHECK calls causes the back end to quit with an error, preventing an incorrect BMv2 JSON file from being written, which will help debug p4c more quickly if/when this happens.

jafingerhut:
As of writing this comment, about 6 PSA automated BMv2 tests are failing because of the new BUG_CHECK calls.  Apparently these tests have been producing correct BMv2 JSON files "by accident", that is, because they could not find the action and used action id 0 by default, which just happened to be the correction action id in that case.

We can either mark those 6 programs as XFAIL tests and then merge in these changes, or find and fix whatever the root cause of issue #1918 is and merge the fix and these enhancements all at once, which should keep the existing PSA test programs that pass, continuing to pass.

jafingerhut:
Even though this is now passing tests, it is probably a good idea to wait until the Cornell team gets their other PSA changes for resubmitted packets working and merged in before merging this PR -- it would likely disrupt some of their latest work, and there is no urgency to merge in these changes yet.

jafingerhut:
Updated this PR so there are no more merge conflicts and all tests pass again.  I think this is useful to merge in.  It is specific to the PSA back end of p4c, and doesn't affect any others, and I am not aware of any other pending PSA improvements in other PRs.

jafingerhut:
In case anyone looks at these comments again, a correction to my previous comment: This PR _does_ affect all bmv2 back ends of p4c, including the one for the v1model architecture.  However, in all of the v1model automated tests included with p4c, it does not ever find a problem.  Thus the problem caught by the new BUG_CHECK() calls in this PR seem to be unique to the PSA back end code.

Issue #1926, opened 2019-05-07
extern_instances are not generated


Pfeifenjoy:
Hey,

maybe I am missing something here but somehow when I compile
file with externs the extern instances are missing.

They are describe here.
https://github.com/p4lang/behavioral-model/blob/master/docs/JSON_format.md#extern_instances

My command is:
```
	p4c-bm2-ss \
		--std p4-16 \
		--p4runtime-files ${BUILD_DIR}/p4c/basic.txt \
		--emit-externs \
		--target bmv2 \
		--arch v1model \
		-o ${BUILD_DIR}/p4c/basic.json \
		p4/switch.p4
```

In my p4 file I instantiate the extern as follows.

```
extern myextern {
	...
}

myextern() myExtern;  // instantiation
```

When I manually add the extern to the extern_instances everything works.
However I would like them to be generated.

antoninbas:
Duplicate of https://github.com/p4lang/p4c/issues/1882

mbudiu-vmw:
Please test the fix for #1882 and let us know if it solves the problem.

PR #1927, opened 2019-05-07
Emit instances for unknown externs

### LINKED ISSUE #1882: Extern object instances not emitted by p4c-bm2-ss
When compiling a p4 program with p4c-bm2-ss which uses an extern object and method, the call to the extern method is added to the output json, but the extern instantiation is not added to `extern_instances`.

Attempting to run the behavioral model with the created dataplane results in an error:
```
Invalid reference to object of type 'extern' with name 'extr_0'
```

Manually adding the extern instance to `extern_instances` in the json works around the issue.

With the extern defined in https://github.com/p4lang/behavioral-model/blob/master/tests/test_extern.cpp implemented in a modified version of the simple_switch, the following code excerpts reproduce the issue (full code attached):

In p4:
```p4
extern ExternCounter {
    ExternCounter();
    void increment();
}

control SampleProcess(inout headers_t hdr, inout metadata_t meta, inout standard_metadata_t smd) {
    ExternCounter() extr;

    apply {
        extr.increment();
    }
}
```

Produces:
```json
  "actions" : [
    {
      "name" : "act",
      "id" : 0,
      "runtime_data" : [],
      "primitives" : [
        {
          "op" : "_ExternCounter_increment",
          "parameters" : [
            {
              "type" : "extern",
              "value" : "extr_0"
            }
          ],
          "source_info" : {
            "filename" : "Sources/ExternObj.p4",
            "line" : 24,
            "column" : 8,
            "source_fragment" : "extr.increment()"
          }
        }
      ]
    }
  ],
...
  "extern_instances" : [],
```

Adding the object:
```json
    {
      "type" : "ExternCounter",
      "name" : "extr_0",
      "id" : 0
    }
```
to `extern_instances` resolves the error.

P4 code and output json: 
[ExternObj.zip](https://github.com/p4lang/p4c/files/3109547/ExternObj.zip)

mbudiu-vmw:
I don't know if this fixes #1882, since I don't know exactly how an extern instance should be represented in the json. But perhaps this is good enough.

I have also added an empty-bmv2.p4 test, which was used to create issue1882-bmv2.p4.

### DIFF
--- a/backends/bmv2/common/extern.cpp
+++ b/backends/bmv2/common/extern.cpp
@@ -88,13 +88,21 @@ ExternConverter::convertExternObject(ConversionContext* ctxt,
     }
 }
 
+/// This method is invoked for all externs that do not have a registered
+/// conversion, i.e., unknown by the architecture.
 void
-ExternConverter::convertExternInstance(ConversionContext* ,
-                                       const IR::Declaration* ,
+ExternConverter::convertExternInstance(ConversionContext* ctxt,
+                                       const IR::Declaration* decl,
                                        const IR::ExternBlock* eb,
                                        const bool& emitExterns) {
     if (!emitExterns)
         ::error(ErrorType::ERR_UNKNOWN, "extern instance", eb->type->name);
+    auto attr = new Util::JsonArray();
+    // TODO: add attributes.  It is not clear how
+    // JsonObjects::add_extern_attribute can be used to insert attributes,
+    // since it does not seem to have enough fields to represent
+    // all the data in an attribute.
+    ctxt->json->add_extern(decl->controlPlaneName(), eb->type->getName(), attr);
 }
 
 Util::IJson*
--- a/backends/bmv2/common/extern.cpp
+++ b/backends/bmv2/common/extern.cpp
@@ -101,7 +101,7 @@ ExternConverter::convertExternInstance(ConversionContext* ctxt,
     // TODO: add attributes.  It is not clear how
     // JsonObjects::add_extern_attribute can be used to insert attributes,
     // since it does not seem to have enough fields to represent
-    // all the data in an attribute.
+    // all the data in an IR::Attribute.
     ctxt->json->add_extern(decl->controlPlaneName(), eb->type->getName(), attr);
 }
 
--- a/backends/bmv2/common/extern.cpp
+++ b/backends/bmv2/common/extern.cpp
@@ -95,14 +95,29 @@ ExternConverter::convertExternInstance(ConversionContext* ctxt,
                                        const IR::Declaration* decl,
                                        const IR::ExternBlock* eb,
                                        const bool& emitExterns) {
-    if (!emitExterns)
+    if (!emitExterns) {
         ::error(ErrorType::ERR_UNKNOWN, "extern instance", eb->type->name);
-    auto attr = new Util::JsonArray();
-    // TODO: add attributes.  It is not clear how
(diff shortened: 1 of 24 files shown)


mbudiu-vmw:
This has also failed because some stupid Python package cannot be installed because someone hasn't tested the Python code they rolled out in a public package.

cc10512:
We need a solution for getting our builds go through. Since we now have support for Ubuntu 18.04, do we want to change our containers to it? @antoninbas how would that work with the third party and bmv2 repos?

cc10512:
Alternatively we can pin the version of the tenjin package to something we know works.

mbudiu-vmw:
I don't know if tenjin is the problem. The error occurs here:
```
  File "/usr/share/python-wheels/urllib3-1.13.1-py2.py3-none-any.whl/urllib3/util/retry.py", line 228, in increment
    total -= 1
TypeError: unsupported operand type(s) for -=: 'Retry' and 'int'
```
So this looks like a bug in the urllib3 package, which seems to be used by pip indirectly.

Pfeifenjoy:
@mbudiu-vmw This fixes my problem #1926.
Also as I can tell the format of extern_instances is defined here: https://github.com/p4lang/behavioral-model/blob/master/docs/JSON_format.md#extern_instances

antoninbas:
Ultimately it does seem to be an issue with pip, but it only occurs because the tenjin package fails to download. Why do we even install tenjin in the p4c Dockerfile anyway?

backends/bmv2/common/extern.cpp
```diff
@@ -88,13 +88,21 @@ ExternConverter::convertExternObject(ConversionContext* ctxt,
...
-                                       const IR::Declaration* ,
+ExternConverter::convertExternInstance(ConversionContext* ctxt,
+                                       const IR::Declaration* decl,
                                        const IR::ExternBlock* eb,
                                        const bool& emitExterns) {
     if (!emitExterns)
         ::error(ErrorType::ERR_UNKNOWN, "extern instance", eb->type->name);
+    auto attr = new Util::JsonArray();
+    // TODO: add attributes.  It is not clear how
+    // JsonObjects::add_extern_attribute can be used to insert attributes,
```
antoninbas: The "attributes" here should be the (compile-time) arguments to the extern constructor (as hexadecimal strings). I think this is very different from `IR::Attribute`, which seems to be used for P4_14 "externs" only.

mbudiu-vmw: OK, I will add the constructor arguments.
I assume that only scalar values are supported.

antoninbas: Ideally enum members would be supported as well. We have 2 choices: the P4 name of the enum members or the numerical value allocated by the backend (since the attribute type can be "hexstr" or "string"). My preference is to use the numerical value, and I think that's what @derekso1 did for PSA counters.

derekso1: Yes, I opted for numerical values since as Antonin said, them being "hexstr" makes for easier integration with bmv2. At least in psa_switch, many enums defined in psa.p4 are used to instantiate externs so making them numerical made it easier to create said externs which support only "hexstr" and "string" for attributes.

mbudiu-vmw:
I will merge this. Enums may "just work", but have not been tested. In general, such enums would appear in an architecture file (since they are particular to the extern that uses them), so they may need to be handled specially.

PR #1929, opened 2019-05-09
Remove tenjin Python package from Docker dependencies

antoninbas:
tenjin installation seems to be creating issues in CI. It is unclear whether
this is because of tenjin or pip, but nothing in this repo uses tenjin anyway.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -23,9 +23,7 @@ ENV P4C_DEPS bison \
              libgc-dev \
              libgmp-dev \
              pkg-config \
-             python-ipaddr \
              python-pip \
-             python-setuptools \
              tcpdump
 ENV P4C_EBPF_DEPS libpcap-dev \
              libelf-dev \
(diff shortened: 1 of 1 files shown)


cc10512:
Looks like if not tenjin, then pyroute2 ... 18.04 then? Custom pip?

jafingerhut:
Looks like tenjin isn't the culprit.  The list of Ubuntu packages installed in the Dockerfile has some differences from what is recommended in the p4c/README.md.  e.g. python-setuptools is installed by Dockerfile, but is mentioned nowhere in the p4c/README.md.  I don't know which package install is actually failing from the error messages, though.

antoninbas:
Let me try to upgrade pip.

jafingerhut:
Is it significant that early log output shows several mentions of "trusty" (Ubuntu 14.04) but later package installations all mention "xenial" (Ubuntu 16.04)?

antoninbas:
> Is it significant that early log output shows several mentions of "trusty" (Ubuntu 14.04) but later package installations all mention "xenial" (Ubuntu 16.04)?

No, it just means that the Host OS (Travis VM) is Ubuntu 14.04, but the docker image uses Ubuntu 16.04.

antoninbas:
Replaced by https://github.com/p4lang/p4c/pull/1931

PR #1933, opened 2019-05-09
Framework for non-local branches in ControlFlowVisitor

ChrisDodd:
- add support for tracking named 'global' flow paths between different
  parts of the IR, useful for tracking data flow over things like
  exit/return or calls or state transitions.

### DIFF
--- a/ir/visitor.h
+++ b/ir/visitor.h
@@ -134,6 +134,12 @@ class Visitor {
      * control flow graph.  Should update @this and leave the other unchanged.
      */
     virtual void flow_merge(Visitor &) { }
+    /** Support methods for non-local ControlFlow computations */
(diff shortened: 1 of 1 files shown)


ir/visitor.h
```diff
@@ -134,6 +134,12 @@ class Visitor {
      * control flow graph.  Should update @this and leave the other unchanged.
      */
     virtual void flow_merge(Visitor &) { }
+    /** Support methods for non-local ControlFlow computations */
```
mbudiu-vmw: do you mean inter-procedural?

ChrisDodd: Can be, but can also be used for goto-like things within a function (eg, exit/return/break).  They're 'non local' in that they mean control flow from an IR node to another IR node that is not a child or parent of that node.

PR #1934, opened 2019-05-09
Report errors for duplicate objects with the same name; fixes #1932

### LINKED ISSUE #1932: Which kinds of P4_16 objects should be allowed to have the same name?
For example, latest `p4test` as of 2019-May-08 gives no error for this program, which gives the name `foo` to both an action defined at the top level, and a control declaration.
```
#include <core.p4>
#include <v1model.p4>
// nameconflict-actionDeclaration-controlDeclaration.p4
action foo (in bit<8> x, out bit<8> y) { y = (x >> 2); }
control foo (in bit<8> x, out bit<8> y) { apply { y = x + 7; } }

struct headers_t { }
struct metadata_t { }
parser parserImpl(packet_in packet, out headers_t hdr, inout metadata_t meta, inout standard_metadata_t stdmeta) { state start { transition accept; } }
control verifyChecksum(inout headers_t hdr, inout metadata_t meta) { apply { } }
control ingressImpl(inout headers_t hdr, inout metadata_t meta, inout standard_metadata_t stdmeta) { apply { } }
control egressImpl(inout headers_t hdr, inout metadata_t meta, inout standard_metadata_t stdmeta) { apply { } }
control updateChecksum(inout headers_t hdr, inout metadata_t meta) { apply { } }
control deparserImpl(packet_out packet, in headers_t hdr) { apply { } }
V1Switch(parserImpl(), verifyChecksum(), ingressImpl(), egressImpl(), updateChecksum(), deparserImpl()) main;
```

I have a simple Python program that produces 20*20=400 little test programs like this, with the same name for each of 20 different "kinds" of named things that can appear at the top level in a P4_16 program.

Most of those 400 programs cause `p4test` to give an error message like `foo duplicates foo` or `error: Re-declaration of foo with different type:` as I would expect, but 36 of those 400 pairs give no error at all.  A handful give an error message other than the one I would expect.

This question is motivated by the design of a P4_16 module system.  If a module can have multiple named "things" at the top level with the same name, and this is legal, then does that mean we should have a way to import one of those things but not the other?  If so, we need something more than just a name to specify which one to import.

mbudiu-vmw:


jafingerhut:
Looks like all of the failures are due to a dozen or so P4_16+v1model programs defining a control called `verify_checksum` that matches the name of the extern function in v1model.p4.  Renaming the control to something like `verifyChecksum` as many of the v1model test programs do should make them pass.

jafingerhut:
It isn't clear to me whether there is any conflict in intent between these changes and this other recent change by @antoninbas https://github.com/p4lang/p4c/commit/a7aa7d0d3ab8c8502bf15b9823a3b0012e7ad313

At least in the specific case mentioned there, a P4_16 table cannot be a top level object, so it is not involved in the changes in this PR.

mbudiu-vmw:
@antoninbas looks at `@name` annotations (some of which are generated automatically, while this looks at identifier names. These are related, but not the same.

mbudiu-vmw:
Fixes #1936

mbudiu-vmw:
This is a one-line fix; it shouldn't be that hard to review it.

jafingerhut:
The difficult part may be deciding whether it is a desirable change. :-)

mbudiu-vmw:
@liujed : I noticed you had enough time to take a look at my prior PR, so perhaps you can look at this one too, it's very short. I have one more PR like that...

frontends/p4/validateParsedProgram.cpp
```diff
@@ -225,7 +225,8 @@ void ValidateParsedProgram::postorder(const IR::P4Program* program) {
         cstring name = decl->getName();
         auto existing = declarations.getDeclaration(name);
         if (existing != nullptr) {
-            if (!existing->is<IR::IFunctional>() || !decl->is<IR::IFunctional>()) {
+            if (!existing->is<IR::IFunctional>() || !decl->is<IR::IFunctional>()
+                || typeid(*existing) != typeid(*decl) || decl->is<IR::P4Action>()) {
```
liujed: I'm surprised that there is a `decl->is<IR::P4Action>()`, but not a matching `existing->is<IR::P4Action>()`. Is that missing here?

mbudiu-vmw: I think that the typeid will take care of that.

mbudiu-vmw: I guess not, it's an or.

mbudiu-vmw: yes, but it's a negated or. so it should work

liujed:
> The difficult part may be deciding whether it is a desirable change. :-)

Agreed. As a rule of thumb, I generally prefer these kinds of things to live in different namespaces. When, say, a parser, a control, and an action all have the same name, I tend to think of it not in terms of overloading, but more as entities existing in separate universes, but happen to have the same name.

frontends/p4/validateParsedProgram.cpp
liujed: Ah, I see. Because of the `typeid` check, we'd expect `decl` and `existing` to have the same type, and so, adding an `|| existing->is<IR::P4Action>()` would be redundant.

mbudiu-vmw:
Well, the actions and parsers live in different universes, but this is easy for the declarations.
The problem is resolving the calls. In a call you have less information about what you are trying to do.
We mostly use the name of an object to disambiguate a call, not much of the context.

jafingerhut:
> > The difficult part may be deciding whether it is a desirable change. :-)
> 
> Agreed. As a rule of thumb, I generally prefer these kinds of things to live in different namespaces. When, say, a parser, a control, and an action all have the same name, I tend to think of it not in terms of overloading, but more as entities existing in separate universes, but happen to have the same name.

One of the reasons motivating my creating issue https://github.com/p4lang/p4c/issues/1932, as it says there, is thinking about the P4_16 module system.

If you want to be able to write something like "from module_foo import name1 as alias1", but "name1" is the name of a parser, a control, and an action, do you always want to import all three, and give them the same alias "alias1"?

If you want the ability to pick and choose which are imported, and with which local aliases, then it at least raises the question of whether all top level names are lumped together, or not.  I am not saying that just because that issue and this PR exists, that we must make the change -- the issue exists to raise the question.

PR #1938, opened 2019-05-15
Don't replace mark_to_drop action if standard_metadata is the only use in the action

pkotikal:


cc10512:
This is a backend specific, so it should be subclassed in the backend.

PR #1940, opened 2019-05-16
Consistent naming for extern instances; fixes #1882

### LINKED ISSUE #1882: Extern object instances not emitted by p4c-bm2-ss

mbudiu-vmw:


cc10512:
Reference outputs?

mbudiu-vmw:
The previous fix already contains reference outputs.
However, the json file (the only one that changes) is not a reference output that we save - there is too much churn in the json.

mbudiu-vmw:
The previous fix is #1927

PR #1941, opened 2019-05-17
Add a getter for invocations and move a BMV2 specific test under ENABLE_BMV2 flag

pkotikal:


midend/removeParameters.h
```diff
@@ -122,6 +122,9 @@ class DoRemoveActionParameters : public Transform {
     const IR::Node* postorder(IR::P4Action* table) override;
     const IR::Node* postorder(IR::ActionListElement* element) override;
     const IR::Node* postorder(IR::MethodCallExpression* expression) override;
+    const ActionInvocation* getInvocations() {
```
mbudiu-vmw: Is this better than accessing the field itself?
Also, do you want to make the method const?

cc10512: The field is private, so you can't access it directly in a subclass. And can't be const since you can add to it.

PR #1945, opened 2019-05-21
Improve error messages; fixes #1944

### LINKED ISSUE #1944: Bit Shifting on A Very Large Bit Vector
Hello,

I was wanting to do bit shifting on a 2147483648 bit-vector, but I noticed that the P4 compiler would return a compilation error stating that "Width cannot be negative or zero". But my bit-vector width was neither negative nor zero. I referred to the P416 language specification which states that "P4 architectures may impose additional constraints on bit types: for example, they may limit the max size". I then checked the BMv2 specification, which states that "For most of these fields, their is not strict requirement to bit-width". Do you know why I am getting this message? 

Attached is a screen shot of the error message and my code

Thanks
Aatish

Screen Shot 2019-05-17 at 5 39 14 PM

[Bit_vector_error.pdf](https://github.com/p4lang/behavioral-model/files/3194364/Bit_vector_error.pdf)

mbudiu-vmw:


testdata/p4_16_errors_outputs/issue1944.p4-stderr
```diff
@@ -0,0 +1,4 @@
+issue1944.p4(1): [--Werror=overlimit] error: : value too large for type width
```
antoninbas: Should this error message somehow convey that this is a compiler limitation and not a P4 language restriction?

Issue #1376, opened 2018-06-28
llvm dependencies in ebpf backend can cause cmake problems.


ChrisDodd:
Pulling the latest p4c, when I run cmake I get some errors:

    CMake Warning at p4c/backends/ebpf/CMakeLists.txt:107 (MESSAGE):
      Kernel version 4.10.0 too small, expected 4.16.0.  Ignoring ebpf kernel
      tests...

    CMake Error at /usr/share/llvm-3.8/cmake/LLVMConfig.cmake:178 (include):
      include could not find load file:

        /usr/share/llvm/cmake/LLVMExports.cmake
    Call Stack (most recent call first):
      p4c/backends/ebpf/CMakeLists.txt:114 (find_package)

    CMake Error at /usr/share/llvm-3.8/cmake/LLVMConfig.cmake:181 (include):
      include could not find load file:

        /usr/share/llvm/cmake/LLVM-Config.cmake
    Call Stack (most recent call first):
      p4c/backends/ebpf/CMakeLists.txt:114 (find_package)

If I comment out the `find_package(LLVM)` line in that last cmake file, everything configures, but some stuff is disabled.

The README.md file just says that llvm 3.7.1 **or later** is required.  Is the problem llvm 3.8?

antoninbas:
I had the same issue, I had to run `cmake` with `-DEENABLE_EBPF` to get rid of it.

mbudiu-vmw:
This seems to be a bug in the llvm package for some ubuntu versions. The makefile has a link to a stackoverflow post describing a fix. It would be nice if the find-package would not crash and would just return false...

mbudiu-vmw:
Here is the link describing the problem and the solution:
https://stackoverflow.com/questions/38171543/error-when-using-cmake-with-llvm
It is in a pending PR.
I am not sure what to do about it in the makefile.

fruffy:
BCC seems to have a similar problem. This issue discusses several more workarounds:
https://github.com/iovisor/bcc/issues/492

PR #1946, opened 2019-05-23
Make eBPF CMAKE robust against llvm check failures

fruffy:
This pull request is intended to introduce a workaround for #1376.
Instead of using CMAKE's `find_package` we manually check the LLVM version and compare. This prevents crashes with specific versions of LLVM and Ubuntu. 
The downside of this method is that it does not catch versions of LLVM that were installed (via ` apt install llvm-3.8` for example) and disables the eBPF backend despite an LLVM version that exists. 
However, because the eBPF framework uses `llc` anyway instead of, say, `llc-3.8` it is probably for the better that this discrepancy is caught.

Let me know what you think.

### DIFF
--- a/backends/ebpf/CMakeLists.txt
+++ b/backends/ebpf/CMakeLists.txt
@@ -105,19 +105,24 @@ endif()
 
 # Check if we have the right llvm version
 set (MIN_LLVM 3.7.1)
-
-# If you get an error here see
-# https://stackoverflow.com/questions/38171543/error-when-using-cmake-with-llvm
-find_package(LLVM QUIET)
-if (LLVM_PACKAGE_VERSION)
+# Grab the LLVM version, do not use find_package because it is unstable
+# https://github.com/p4lang/p4c/issues/1376
+set(LLVM_CMD "llvm-config --version")
+message(STATUS "Check LLVM version with '${LLVM_CMD}'")
+exec_program(${LLVM_CMD}
+	 RETURN_VALUE LLVM_RET
+	 OUTPUT_VARIABLE LLVM_PACKAGE_VERSION)
+if (NOT LLVM_RET)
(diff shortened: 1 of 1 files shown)


backends/ebpf/CMakeLists.txt
```diff
@@ -105,19 +105,24 @@ endif()
...
   if (${LLVM_PACKAGE_VERSION} VERSION_LESS ${MIN_LLVM})
-    MESSAGE(WARNING "LLVM version ${LLVM_PACKAGE_VERSION} too small, expected ${MIN_LLVM}.
+    message(WARNING "LLVM version ${LLVM_PACKAGE_VERSION} too small, expected ${MIN_LLVM}.
     Ignoring ebpf tests...")
     set (SUPPORTS_KERNEL False)
   endif()
 else()
-  message(STATUS "LLVM missing, disabling kernel tests...")
+  message(WARNING
+	 "Problem with LLVM, disabling kernel tests...\n"
```
mbudiu-vmw: This message could be more user-friendly.
Perhaps "Could not detect a version of LLVM which is needed to compile the eBPF kernel tests".

Did you test this on both systems that have working and broken llvms?

fruffy: Yes, I tested it on an Ubuntu Xenial image with the broken llvm package and on my local version. Both worked fine. The Travis test also passed. This method does not seem to work on macOS, but we do not support that anyway as far as I remember.

mbudiu-vmw: On Mac we don't support eBPF, but the build should still work, of course.
Travis will test on Mac, so I think that if that passes (and it did previously) then we can merge this.

fruffy: Okay, it should pass, it will just complain that it cannot find llvm, which also might be just a setup issue in the macOS Travis image.

PR #1948, opened 2019-05-30
Bug fix in typechecker: fixes #1937

### LINKED ISSUE #1937: Should parsers, controls, and actions support default parameter values?
Will add test programs in a later comment.

mbudiu-vmw:


frontends/p4/typeChecking/typeChecker.cpp
```diff
@@ -956,6 +956,8 @@ const IR::Node* TypeInference::preorder(IR::Declaration_Instance* decl) {
             decl, decl->arguments, simpleType->to<IR::IContainer>());
         auto type = typeAndArgs.first;
         auto args = typeAndArgs.second;
+        auto *learn = clone();
+        (void)type->apply(*learn);
```
ChrisDodd: Can `type == nullptr;` here?  You check for it just below, which is too late.  If the *learn transform does something, do you really want to throw it away?  That seems wrong.  Maybe just want `visit(type, "type")`?

mbudiu-vmw: I will push this lower, after the null check.
This pattern is used all over the place in the typechecker; the recursive call updates the typeMap as a side-effect, so the result is not used. This visit may revisit some graph nodes that were already visited, that's the reason for the clone call. I believe that visit will be upset if you try to do that.

testdata/p4_16_samples_outputs/issue1937-1-bmv2-frontend.p4
```diff
@@ -0,0 +1,60 @@
...
+}
+
+struct metadata_t {
+}
+
+control ingressImpl(inout headers_t hdr, inout metadata_t meta, inout standard_metadata_t stdmeta) {
+    @name(".foo") action foo(out bit<8> x, in bit<8> y=5) {
+        x = y >> 2;
+    }
+    @name(".foo") action foo_0(out bit<8> x_1, in bit<8> y_1=5) {
```
jafingerhut: Having two actions with `@name(".foo")` annotations looks like wrong to me.  If the user did this in their original source program, the compiler does not catch it as an error?

testdata/p4_16_samples_outputs/issue1937-2-bmv2-first.p4
```diff
@@ -0,0 +1,59 @@
...
+parser foo(out bit<8> x, in bit<8> y=5) {
+    state start {
+        x = y >> 2;
+        transition accept;
+    }
+}
+
+parser parserImpl(packet_in packet, out headers_t hdr, inout metadata_t meta, inout standard_metadata_t stdmeta) {
+    @name("foo") foo() foo_inst;
+    @name("foo") foo() foo_inst_0;
```
jafingerhut: There are probably other occurrences of duplicate @name annotations in the intermediate compiler pass results, too, but I will only point out this one more explicitly.

testdata/p4_16_samples_outputs/issue1937-1-bmv2-frontend.p4
```diff
@@ -0,0 +1,60 @@
...
+        x_1 = y_1 >> 2;
+    }
+    bit<8> tmp;
+    bit<8> tmp_0;
+    apply {
+        tmp_0 = hdr.h1.f1;
+        foo(tmp, tmp_0);
+        hdr.h1.f1 = tmp;
+        foo_0(x = hdr.h1.f2, y = 8w5);
```
jafingerhut: I do not see how this intermediate code can compile, given that action foo_0 above has parameters named x_1 and y_1, not x and y.

mbudiu-vmw: This was a bug in the code that emitted P4, internally the names were correct. That's why the programs would compile correctly. I will add a commit to fix this in this PR.

testdata/p4_16_samples_outputs/issue1937-1-bmv2-frontend.p4
jafingerhut: Does it make sense that the compiler is taking one action, and creating two separate actions from it?  Why is that useful?

testdata/p4_16_samples_outputs/issue1937-3-bmv2-first.p4
```diff
@@ -0,0 +1,58 @@
...
+
+control foo(out bit<8> x, in bit<8> y=5) {
+    apply {
+        x = y >> 2;
+    }
+}
+
+control ingressImpl(inout headers_t hdr, inout metadata_t meta, inout standard_metadata_t stdmeta) {
+    @name("foo") foo() foo_inst;
+    @name("foo") foo() foo_inst_0;
```
jafingerhut: Similar question here and for the parser case: Is it good/important/useful that the compiler is creating two instantiations of control foo?  If foo contained stateful elements like tables, counters, meters, etc., this would be instantiating two of each of those things, yes?

mbudiu-vmw: Yes, the meaning of invoking a parser is that you create a new instance which you then invoke. So this is exactly what should happen. If you want the other way you should instantiate the parser once and invoke it twice.

testdata/p4_16_samples_outputs/issue1937-1-bmv2-frontend.p4
mbudiu-vmw: The code makes sure that each action is used in a single table. For some targets there is no action reuse anyway; if you have the same code you have to implement it twice. 
This also opens the opportunity to specialize the action body for each call site. Perhaps this should be a controllable option - whether to duplicate or not. But it's not wrong, in the sense that it preserves the program semantics.

mbudiu-vmw: This is not an error, you can have multiple actions with the same name; the control-plane disambiguates them through the tables that invoke them.

jafingerhut: These are both invoked by the same table, so no such through-the-table disambiguation is possible here, is it?

testdata/p4_16_samples_outputs/issue1937-3-bmv2-first.p4
jafingerhut: Yes, sorry for my unnecessary comment.  I realized this was a case of a parser or control being auto-instantiated via an apply call, so two instantiations is correct.

The @name annotations on the two instantiations being identical still looks wrong, though.

testdata/p4_16_samples_outputs/issue1937-2-bmv2-first.p4
jafingerhut: Hmm.  A closer reading on my part of the P4_16 language spec, version 1.1, Section 14.1 "Direct type invocation" seems to me to mean that this example program is illegal, according to this text:

"In the same scope, direct type invocation will result in a different local instance per invocation—however, instances of the same type will share the same global name, via the @name annotation. If the type contains controllable entities, then invoking it directly more than once in the same scope is illegal, because it will produce multiple controllable entities with the same fully-qualified control name."

If that interpretation is correct, this example program should cause the compiler to give an error because of the multiple "automatic instantiations" of parser "foo".  The same would hold for the example program that does multiple "automatic instantiations" of a control.

mbudiu-vmw: Parser are not controllable entities. They do not expose any APIs to the control-plane.

jafingerhut: Parsers can contain value_set instantiations, which do have a control plane API and internal state.

mbudiu-vmw: If you will instantiate such a parser twice you should get an error. But if you instantiate a parser without any stateful objects I don't see why you should get an error. The problem is that the value_set has two instances with the same name, not the parser that contains it.

jafingerhut: So the P4_16 language allows arbitrary extern objects to be instantiated within parsers, too.

Are you saying that only a back-end compiler that knows whether the extern object has a control plane API or not, is the only place that such a program should be determined to be legal or illegal?

jafingerhut: I understand wanting to make more programs legal, but it seems like a reasonable simplification to the compiler, to tighten things up here and say it is always illegal to use multiple direct type invocations on the same parser or control name, regardless of whether it contains a control plane API or not.

mbudiu-vmw: The code that generates the control-plane API should complain if it doesn't like something.
The fact that there is an extern does not mean that the extern is controllable by the control-plane using its name. After inlining these two parser instances will disappear anyway, so by the time you reach the back-end there aren't any more objects named the same way.

frontends/p4/toP4/toP4.cpp
```diff
@@ -254,7 +254,7 @@ bool ToP4::preorder(const IR::Type_Specialized* t) {
 
 bool ToP4::preorder(const IR::Argument* arg) {
     if (!arg->name.name.isNullOrEmpty()) {
-        builder.append(arg->name.toString());
+        builder.append((cstring)arg->name);
```
ChrisDodd: explicit casts are always an indication something is not right.  This should work with an implicit conversion and so the cast could just be removed?

frontends/p4/typeChecking/typeChecker.cpp
```diff
@@ -960,6 +960,8 @@ const IR::Node* TypeInference::preorder(IR::Declaration_Instance* decl) {
             prune();
             return decl;
         }
+        auto *learn = clone();
+        (void)type->apply(*learn);
```
ChrisDodd: This applying a clone and ignoring the return value just to update the typeMap still bugs me -- its a symptom of the messiness of having the typeMap separate from the type field(s) in the IR that I really would like to clean up eventually.

mbudiu-vmw: Removing the referenceMap should be easier.
The typechecking pass has grown quite complicated. Some of the complications are probably unnecessary - like maintaining canonical types. But I don't see how you can maintain such a complicated graph of types without an auxiliary data structure. 
The most annoying thing about type checking is that it needs to run over and over, any time the graph changes. But I think that the alternative of having each pass compute types on its own is worse. At least this way the type checking is centralized in a single pass.

PR #1950, opened 2019-05-31
Some (aspirational) tests for parser inlining

liujed:
The expected output for `inline-parser2.p4` is aspirational, and the test is currently xfailed as a result. When inlining, copying seems to be always introduced if the parser is invoked more than once. This copying should be avoided.

### DIFF
--- a/backends/p4test/CMakeLists.txt
+++ b/backends/p4test/CMakeLists.txt
@@ -84,6 +84,8 @@ set (P4_XFAIL_TESTS
   # appear as the value of the psa_direct_counter table attribute for at most
   # one table."
   testdata/p4_16_samples/psa-counter6.p4
+
+  testdata/p4_16_samples/inline-parser2.p4
   )
 # we invoke p4c with --p4runtime-files even for programs in p4_16_errors. This
 # enables us to use p4_16_errors even for programs which pass compilation but
--- a/testdata/p4_16_samples/inline-parser1.p4
+++ b/testdata/p4_16_samples/inline-parser1.p4
@@ -1,0 +1,72 @@
+#include <v1model.p4>
+
+header data_t {
+  bit<8> f;
+}
+
+struct headers_t {
+  data_t data1;
+  data_t data2;
+}
+
+struct user_metadata_t {
+}
+
+parser SubParser(packet_in pkt, inout data_t data) {
+  state start {
+    transition accept;
+  }
+}
+
+parser MyParser(
+    packet_in pkt,
+    out headers_t hdr,
+    inout user_metadata_t md,
+    inout standard_metadata_t smd) {
+
+  SubParser() subparser;
+
+  state start {
+    subparser.apply(pkt, hdr.data1);
+    transition accept;
+  }
+}
+
+control MyIngress(
+    inout headers_t hdr,
+    inout user_metadata_t md,
+    inout standard_metadata_t smd) {
+
+  apply {}
+}
+
+control MyEgress(
+    inout headers_t hdr,
+    inout user_metadata_t md,
+    inout standard_metadata_t smd) {
+
+  apply {} 
+}
+
+control MyDeparser(
+    packet_out pkt,
+    in headers_t hdr) {
+
+  apply {}
+}
+
+control MyVerifyChecksum(inout headers_t hdr, inout user_metadata_t md) {
+  apply {}
+}
+
+control MyComputeChecksum(inout headers_t hdr, inout user_metadata_t md) {
+  apply {}
+}
+
+V1Switch(
+    MyParser(),
+    MyVerifyChecksum(),
+    MyIngress(),
+    MyEgress(),
+    MyComputeChecksum(),
+    MyDeparser()) main;
--- a/testdata/p4_16_samples/inline-parser2.p4
+++ b/testdata/p4_16_samples/inline-parser2.p4
@@ -1,0 +1,73 @@
+#include <v1model.p4>
+
+header data_t {
+  bit<8> f;
+}
+
+struct headers_t {
+  data_t data1;
+  data_t data2;
+}
+
+struct user_metadata_t {
+}
+
+parser SubParser(packet_in pkt, inout data_t data) {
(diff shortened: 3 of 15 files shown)


mbudiu-vmw:
Ideally the copying should make no difference in performance, because later passes like copy propagation should remove redundant copies.

The semantics of P4 calls is copy-in/copy-out; the inliner is conservative if it cannot prove that direct substitution provides the same semantics and it uses copy-in/copy-out.

liujed:
@mbudiu-vmw I was just in the middle of writing up an issue for this, but since I have your attention, I guess the discussion can happen here.

This method seems to be the place to look: https://github.com/p4lang/p4c/blob/master/frontends/p4/inlining.cpp#L457. I'm not entirely understanding this code, though. I'd expect each invocation to be inlined independently, but this code appears to be doing inlining for all invocations? Can you shed some light on how inlining works at a high level?

mbudiu-vmw:
The inliner is pretty complicated, so I am never sure myself without consulting the code. One thing is that you have instances and invocations; each instance can be invoked multiple times. When you inline some parameters (constructor parameters) come from the instance, and some come from the invocation (from the call site). So you have to keep track of both. Moreover, constructor parameters and direction-less parameters can be inlined by direct name substitution (since they are supposed to be compile-time constants), while the other parameters should be done using copy-in/copy-out. 

For parameters with a direction I avoid copy-in/copy-out and perform just name substitution when I am certain that this causes no problems. Arguments may not alias to each other, for example. But if the inlined object is invoked multiple times then I don't just do substitution.

Another thing is that inlining is recursive: it keep running until it runs out of things to inline. I believe that inlining goes bottom up, from leaf objects (that have no other callees that can be inlined). 

The inline driver is just a while loop that keeps calling a specific inliner until it runs out of work. There are several inliners specialized for their roles: functions, actions, controls, parsers - but some are shared. The general inliner you are pointing at is for controls and parsers.

I don't know what else to write here; perhaps it would be best to review the code in person if you want to know more. I would have to also refresh my mind, it's been a while.

mbudiu-vmw:
For a parser you first instantiate and then you can "apply" it. Instantiating it allocates all objects that it contains - other instances, or externs. So you have one copy of each extern for an instance. But you may then call an instance repeatedly - these have to share the extern instances. Functions and actions are simpler, since they may not contain stateful objects inside.

liujed:
Okay, thanks. I didn't realize that we'd lumped constructor parameters and invocation parameters together like that. Wouldn't it be simpler to have them separated out, and inline the object before inlining the invocations?

mbudiu-vmw:
Nothing is simpler than leaving alone code that works...
If it doesn't work then we can perhaps take a look.
If you want to avoid copies then you just have to refine the policy which decides when to use temporaries, you can leave most of the other code in place.

liujed:
Here is an example that doesn't look quite right to me. Before inlining:
```
parser
SubParser(packet_in pkt, in data_t read, inout data_t data)(bit<8> init) {
  bit<8> x = init;
  state start {
    transition select(x) {
    42: accept;
    default: foo;
    }
  }

  state foo {
    x = read.f;
    transition accept;
  }
}

parser MyParser(
    packet_in pkt,
    out headers_t hdr,
    inout user_metadata_t md,
    inout standard_metadata_t smd) {

  SubParser(54) subparser;

  state start {
    pkt.extract(hdr.data0);
    subparser.apply(pkt, hdr.data0, hdr.data1);
    subparser.apply(pkt, hdr.data0, hdr.data2);
    transition accept;
  }
}
```
After inlining:
```
parser MyParser(packet_in pkt, out headers_t hdr, inout user_metadata_t md, inout standard_metadata_t smd) {
    data_t read_0;
    data_t data_0;
    bit<8> subparser_x;
    state start {
        pkt.extract<data_t>(hdr.data0);
        read_0 = hdr.data0;
        data_0 = hdr.data1;
        transition SubParser_0_start;
    }
    state SubParser_0_start {
        subparser_x = 8w54;
        transition select(subparser_x) {
            8w42: start_0;
            default: SubParser_0_foo;
        }
    }
    state SubParser_0_foo {
        transition start_0;
    }
    state start_0 {
        hdr.data1 = data_0;
        read_0 = hdr.data0;
        data_0 = hdr.data2;
        transition SubParser_0_start_0;
    }
    state SubParser_0_start_0 {
        subparser_x = 8w54;
        transition select(subparser_x) {
            8w42: start_1;
            default: SubParser_0_foo_0;
        }
    }
    state SubParser_0_foo_0 {
        transition start_1;
    }
    state start_1 {
        hdr.data2 = data_0;
        transition accept;
    }
}
```
There are a couple of issues I see here:
* `subparser_x` is re-initialized to 54 on the second invocation, and
* The assignment to `subparser_x` is lost in `SubParser_0_foo` and `SubParser_0_foo_0`.
Am I misunderstanding the intended semantics?

mbudiu-vmw:
x is a local variable, which lives on the "stack", and is allocated every time you invoke the parser. It does not live between invocations. Only things that are instantiated live across invocations of the parser. In fact, probably the last assignment to x was removed before inlining, because it is dead. This semantics for local variables makes it much easier to understand and analyze the code.

BTW: do you know that you can dump the program at any point in the pipeline?
Try `./p4test --top4 42_Inline x.p4`

jnfoster:
I'm surprised by this comment.

Given the control declarations,

control C(...) {
   bit<8> x = 0xff;
   ...
}
control D(...)(bit<8> n) {
   bit<8> x = n;
   /* Otherwise identical to C */
}

I would expect

C() myC;
myC.apply();
myC.apply();

and

D(0xff) myD;
myD.apply();
myD.apply();

to exhibit the same data-plane behavior, even if the apply block modifies
the variable x.

Is that not the case?

-N

mbudiu-vmw:
They will have the same behavior.

liujed:
I agree that a later copy-propagation pass should remove redundancies introduced by inlining. After all, the programmer may write code similar to what the inliner is currently producing.

Issue #1953, opened 2019-06-03
bug in match_t output stream operator?


zma-barefoot:
When `match_t` assumes its default value,

https://github.com/p4lang/p4c/blob/b806cecb70c19620acbb86c963d4f84b48c0a51f/lib/match.h#L33

the `<<` operator will simply print `0`.

https://github.com/p4lang/p4c/blob/b806cecb70c19620acbb86c963d4f84b48c0a51f/lib/match.cpp#L31-L48

Is this a bug? Shouldn't we print `*` (don't care) for the default value?

ChrisDodd:
The default value is a special invalid/match nothing code (and returns false for operator bool -- every other pattern returns true).  So it doesn't really have a proper printable form.

zma-barefoot:
I see. What's the recommended way of creating a don't care match?

zma-barefoot:
If the default value means invalid/null, then the `<<` operator should probably have a `CHECK_NULL`, instead of outputting it as `0`.

ChrisDodd:
A don't care would be something like `match_t(width, 0, 0)`,  where `width` is the width of the don't care.  So `match_t(4, 0, 0)` would print as `0x*`, while `match_t(12, 0, 0)` will print as `0x***`.  I don't recall why we wanted to the special "don't care" pattern -- perhaps just to catch improperly initialized match_t values.

PR #1954, opened 2019-06-03
fix stream output for default value of match_t

zma-barefoot:
As per out discussion in https://github.com/p4lang/p4c/issues/1953, it seems more correct to print the default value of `match_t` (which means "no match") as `*`.

The other idea is to print it as `-`.

### DIFF
--- a/lib/match.cpp
+++ b/lib/match.cpp
@@ -14,7 +14,6 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-#include <assert.h>
 #include "match.h"
 
 static int chkmask(const match_t &m, int maskbits) {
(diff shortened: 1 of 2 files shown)


lib/match.cpp
```diff
@@ -29,6 +28,10 @@ static int chkmask(const match_t &m, int maskbits) {
 }
 
 std::ostream &operator<<(std::ostream &out, match_t m) {
+    if (!m) {
+        out << "*";
+        return out;
+    }
```
ChrisDodd: Could fold this into the final `else` just below (that case only runs when `m.word0|m.word1 == 0`, the same as `!m`)

PR #1959, opened 2019-06-06
add resubmit test for psa

peteli3:
@jafingerhut 
this tests checks that:
1) packet is actually resubmitted
2) the packet that was resubmitted == the original, but with `packet_path` metadata set to resubmit

redo of #1904

### DIFF
--- a/testdata/p4_16_samples/psa-resubmit-bmv2.p4
+++ b/testdata/p4_16_samples/psa-resubmit-bmv2.p4
@@ -55,7 +55,7 @@ control cIngress(inout headers_t hdr,
 
     action resubmit() {
         // this write should not make it to the output packet
-        hdr.ethernet.dstAddr = 48w256;
+        hdr.ethernet.srcAddr = 48w256;
         ostd.resubmit = true;
     }
 
--- a/testdata/p4_16_samples/psa-resubmit-bmv2.p4
+++ b/testdata/p4_16_samples/psa-resubmit-bmv2.p4
@@ -1,0 +1,144 @@
+/*
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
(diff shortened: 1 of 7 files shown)


cc10512:
@peteli3 he means run `P4TEST_REPLACE=1 make check-p4` and add all the changes that happen in the testdata/p4_16_* directories.

mbudiu-vmw:
Or after you run the test just make sure to add all files that show up in p4_16_samples_outputs.

jafingerhut:
The test program and STF test look good to me.  Thanks for the update.  Please do follow one of the suggestions in earlier comments about adding the expected output files that should be in the testdata/p4_16_samples_outputs/ directory for this test program.

testdata/p4_16_samples/psa-resubmit-bmv2.p4
```diff
@@ -0,0 +1,144 @@
...
+        ostd.resubmit = true;
+    }
+
+    action pkt_write() {
+        ostd.drop = false;
+        hdr.ethernet.dstAddr = 48w4;
+    }
+
+    apply {
+        pkt_write();
```
jafingerhut: Thinking about this test case a bit more, it would be a better test if resubmit() and pkt_write() modified _different_ fields of the packet header.  Why?  Because as the test is written now, if psa_switch were incorrectly preserving modifications to the packet from before resubmitting, to after resubmitting, this test will not catch that problem.

For example, if resubmit() had the assignment `hdr.ethernet.srcAddr = 48w256;` _instead of_ the assignment to `hdr.ethernet.dstAddr`, then the test would catch a buggy psa_switch implementation that preserved that modification to srcAddr across the resubmit operation.

peteli3: True. My bad, should have caught this earlier.

PR #1960, opened 2019-06-06
Fix borked ebpf-kernel tests.

fruffy:
Setting pinning to 1 broke some ebpf tests because the maps could not be found. I think this is an issue with `tc`. Using libbpf directly instead of loading with `ip` or `tc` could fix this problem. I don't know how to deal with the namespace and global map problem yet. I think the testing pipeline has to change for that.

The bug was not discovered because SKIPPED set to 2 masked the failures (apparently the error output of the call was 2). Maybe we should clean up the error reporting pipeline. I am not sure which one is better. Escalating the error code of the original program or overriding error codes with the `FAILURE` variable. 

Another issue are IGMP packets which generate a lot of noise and falsify the packet count. We filter them for now.


mbudiu-vmw:
@williamtu : can you please review this PR?

williamtu:
Thanks. Looks good to me.

Setting to 1 fixes the issue in p4c-xdp 
https://github.com/vmware/p4c-xdp/issues/105
but broke the test here. I will have to think about other ways.

Issue #1919, opened 2019-05-03
PSA back end generates BMv2 JSON that refers to non-existent header


jafingerhut:
Compiling the attached P4_16+PSA architecture program using the latest version of p4c as of 2019-May-03:
```
$ p4c-bm2-psa psa-issue1919-bmv2.p4 -o psa-issue1919-bmv2.json
```
yields an incorrect BMV2 JSON file, which the latest version of `psa_switch` as of the same date correctly gives an error message when reading and validating it:
```
$ psa_switch --log-console -i 0@veth2 -i 1@veth4 ~/forks/p4c/testdata/p4_16_samples/psa-issue1919-bmv2.json 
Calling target program-options parser
Invalid reference to object of type 'header' with name 'scalars'
```
If this issue is corrected, the attached STF file test should work.

[psa-issue1919-bmv2.p4.txt](https://github.com/p4lang/p4c/files/3142622/psa-issue1919-bmv2.p4.txt)
[psa-issue1919-bmv2.stf.txt](https://github.com/p4lang/p4c/files/3142623/psa-issue1919-bmv2.stf.txt)

yunheL:
@jafingerhut Hi Andy, I just ran the psa-issue1919-bmv2.p4 and .stf locally. It looks like it has a very similar problem that this new [PR](https://github.com/p4lang/p4c/pull/2420). 

Specifically, `metadata_t.port` is trying to refer a header called `scalars` but in the headers section, the header is somehow called `null` but not `scalars`. Below are screenshots of generated json for `psa-issue1919-bmv2.p4`

Here is the use of `metadata_t.port`, it has a header named `scalars`:
![scalar](https://user-images.githubusercontent.com/11049009/83954464-63c83200-a817-11ea-8059-8772ee1da458.png)

But in the headers section of the generated json file, there is **NO** header object named `scalar`. There is a header object whose name is null and has `scalar_t` type.
![null](https://user-images.githubusercontent.com/11049009/83954466-6460c880-a817-11ea-8b67-fe0b32ac12b2.png)

I think this is the bug and causing bmv2 to give the error `Invalid reference to object of type 'header' with name 'scalars'`.

I can look into how to solve this problem.

jafingerhut:
@yunheL Have you tried running psa-issue1919-bmv2.p4 with your changes for PR https://github.com/p4lang/p4c/pull/2420 ?

If not, please do.  If it fixes the problems you were addressing there, and this one, that is excellent.

[...]

PR #1964, opened 2019-06-07
fix for issue 1923: assignments in ingress apply block not represented

hanw:


jafingerhut:
I am surprised Mihai didn't ask for the front/midend/etc. expected compiler outputs before I mentioned them :-)  I will do some local testing with these changes to see if they work as advertised.  I scanned through the code changes, but as is common I do not know the code well enough to spot anything wrong that way.

jafingerhut:
The test program I wrote, and that you include in the PR, fails for me, with output from psa_switch like this:
```
[15:42:24] $ ./bmv2/testdata/p4_16_samples/psa-simple-reflector-bmv2.p4.test
Check for  /home/jafinger/forks/p4c/testdata/p4_16_samples/psa-simple-reflector-bmv2.stf
Invalid reference to object of type 'header' with name 'scalars'
*** Test failed
```
which is the same error message for issue https://github.com/p4lang/p4c/issues/1919 and I suspect semi-independent of the issue where top-level assignments in the ingress/egress control are not appearing in the BMv2 JSON file.

jafingerhut:
If the p4c code changes are good, though, perhaps the best approach is to remove the failing test program from the PR.

Maybe I can find a different test program that breaks without these changes, and works with them, but I don't think that should hold up the PR.

jafingerhut:
I found a slightly different test program and STF file that fails without these changes, and passes with them.  Attaching them to this comment.
[psa-top-level-assignments-bmv2.p4.txt](https://github.com/p4lang/p4c/files/3267955/psa-top-level-assignments-bmv2.p4.txt)
[psa-top-level-assignments-bmv2.stf.txt](https://github.com/p4lang/p4c/files/3267956/psa-top-level-assignments-bmv2.stf.txt)

hanw:
There seems to be another bug related to 'scalars' field in generated json. I am using the test program in issue 1919 to debug the issue.

hanw:
It's not clear to me what types of fields should be counted as 'scalar'.

mbudiu-vmw:
BMv2 supports only header and struct instances. Everything else (e.g., local variables) is packed in a "scalars" struct.

jafingerhut:
Certainly very happy if you would like to fix both issues in the same PR, Han, but fixing one but not the other also seems like a viable option to me.

PR #1966, opened 2019-06-12
some refactoring to enable translating from p4-14 to other architectures in p4-16 at the beginning of the frontend.

hanw:


frontends/p4/fromv1.0/converters.cpp
```diff
@@ -1369,8 +1376,10 @@ Converter::Converter() {
     passes.emplace_back(new ComputeTableCallGraph(structure));
     passes.emplace_back(new Rewriter(structure));
     passes.emplace_back(new FixExtracts(structure));
-    passes.emplace_back(new FixMultiEntryPoint(structure));
-    passes.emplace_back(new MoveIntrinsicMetadata(structure));
+    if (CompilerOptions().arch == "v1model") {
```
mbudiu-vmw: This is a bit strange, to allocate here a new set of Options.

hanw: Agreed. These two passes assumes v1model, yet are currently implemented in frontend.

PR #1969, opened 2019-06-13
Add IPv6 switch ml test

hesingh:
See https://datatracker.ietf.org/doc/draft-hsingh-ipv6-coin-ml/

This test adds sending an IPv6 link-local packet to a multicast destination.  The multicast packet is replicated to few ports.   It would be good to add this test to p4c for folks in the switch ML community to leverage.

### DIFF
--- a/testdata/p4_16_samples/ipv6-switch-ml-bmv2.p4
+++ b/testdata/p4_16_samples/ipv6-switch-ml-bmv2.p4
@@ -1,0 +1,133 @@
+/*
+* Copyright 2019, MNK Consulting
+* http://mnkcg.com
+*
+* Licensed under the Apache License, Version 2.0 (the "License");
+* you may not use this file except in compliance with the License.
+* You may obtain a copy of the License at
+*
+*    http://www.apache.org/licenses/LICENSE-2.0
+*
+* Unless required by applicable law or agreed to in writing, software
+* distributed under the License is distributed on an "AS IS" BASIS,
+* WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+* See the License for the specific language governing permissions and
+* limitations under the License.
+*
+*
+*    How to test this P4 code:
+* $ ./p4c-bm2-ss --std p4-16 ../testdata/p4_16_samples/ipv6-switch-ml-bmv2.p4 -o tmp.json
+*
+*/
+
+#include <v1model.p4>
+#include "ml-headers.p4"
+
+const bit<32> BMV2_V1MODEL_INSTANCE_TYPE_REPLICATION   = 5;
+#define IS_REPLICATED(std_meta) (std_meta.instance_type == BMV2_V1MODEL_INSTANCE_TYPE_REPLICATION)
+
+parser MyParser(packet_in packet, out headers hdr, inout metadata_t meta,
+                inout standard_metadata_t standard_metadata) {
+    state start {
+        packet.extract(hdr.ethernet);
+        transition select(hdr.ethernet.etherType) {
+            TYPE_IPV6: ipv6;
+            default: accept;
+        }
+    }
+    state ipv6 {
+        packet.extract(hdr.ipv6);
+        transition select(hdr.ipv6.nextHdr) {
+            PROTO_UDP: parse_udp;
+            PROTO_ICMP6: icmp6;
+            default: accept;
+        }
+    }
+    state icmp6 {
+        packet.extract(hdr.icmp6);
+        transition accept;
+    }
+    state parse_udp {
+        packet.extract(hdr.udp);
+        transition accept;	
+    }
+}
+
+// Our switch table comprises of IPv6 addresses vs. egress_port.
+// This is the table we setup here.
+control ingress(inout headers hdr, inout metadata_t meta,
+                inout standard_metadata_t standard_metadata) {
+    action set_mcast_grp(bit<16> mcast_grp, bit<9> port) {
+        standard_metadata.mcast_grp = mcast_grp;
+	meta.egress_port = port;
+    }
+    table ipv6_tbl {
+	key = {
+            (hdr.ipv6.dstAddr[127:120] == 8w0xff) : exact @name("mcast_key");
+	}
+	actions = {set_mcast_grp;}
+    }
+
+    apply {
+        if (hdr.ipv6.isValid()) {
+	    ipv6_tbl.apply();
+        }
+    }
+}
+
+control egress(inout headers hdr, inout metadata_t meta,
+               inout standard_metadata_t standard_metadata) {
+    action set_out_bd (bit<24> bd) {
+        meta.fwd.out_bd = bd;
+    }
+    table get_multicast_copy_out_bd {
+        key = {
+            standard_metadata.mcast_grp  : exact;
+            standard_metadata.egress_rid : exact;
+        }
+        actions = { set_out_bd; }
+    }
+
+    action drop() {
+        mark_to_drop(standard_metadata);
+    }
+    action rewrite_mac(bit<48> smac) {
+        hdr.ethernet.srcAddr = smac;
+    }
+    table send_frame {
+        key = {
+            meta.fwd.out_bd: exact;
+        }
+        actions = {rewrite_mac; drop;}
+        default_action = drop;
+    }
+
+    apply {
+        if (IS_REPLICATED(standard_metadata)) {
+            get_multicast_copy_out_bd.apply();
+	    send_frame.apply();
+	}
+    }
+}
+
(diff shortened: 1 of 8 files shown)


jafingerhut:
The automated tests pass, so there is nothing functionally wrong with this proposed new test program plus STF test.

Reading it over, one thing I do not recall seeing before in another test program before (or any P4_16 program for that matter) is this table definition:
```
    table ipv6_tbl {
	key = {
            (hdr.ipv6.dstAddr[127:120] == 8w0xff) : exact @name("mcast_key");
	}
	actions = {set_mcast_grp;}
    }
```
I was mildly surprised that STF could add a table entry to that table, given that the key has type boolean.

Multicast-wise, this seems to exercise similar functionality to one or two existing STF tests, but if we think the above ought to work, adding a test that it does seems worth it.

hesingh:
Thanks much for the expedited review.  When you get a chance, please merge this PR with p4lang/p4c - thanks.

jafingerhut:
I would recommend not closing the PR if you want it merged.  Otherwise it falls off the list of open PRs and few people will notice it.

hesingh:
Sorry about closing the PR sooner than the PR merge,

jafingerhut:
@antoninbas Any objections to including this test for p4c?  The only weird part that I can see is the table that I mentioned in a previous comment that has a boolean expression as a key field.  It seems to work right now, including adding a table entry using the existing STF infrastructure, using a 1-bit integer instead of a boolean value.  I wanted to double-check with you that this is something that isn't just working by accident, because if it is by accident, perhaps we do not want such a test checked in.

hesingh:
I have personally never used boolean's true/false in a STF file.  I did try it today, but `true` does not work.  `0x1` does.

hesingh:
Just add three P4 programs of this PR into `p4c/testdata/p4_16_samples/`.
Then, from `p4c/build/` invoke `cmake ..`
Next, test, using this command from `p4c/build/`

`bmv2/testdata/p4_16_samples/ipv6-switch-ml-bmv2.p4.test`

The test will pass.  

Next, edit  `ipv6-switch-ml-bmv2.stf` to change the "0x1" in key to "0x0".  Re run the test.  This time, the test fails.  

One can also inspect the bmv2 backend.  This backend converts boolean to integer of width 1 and values in 1 or 0.  

Thus, I don't see any evidence of the program and compiler working fortuitously.

jafingerhut:
If you are willing to change the table key expression from having a boolean result to a `bit<1>` type result, I would definitely merge it in, no problem.

I just don't want to check in a test that might be relying on an accident of the current implementation, rather than an on-purpose property of the current implementation.  I know that the test as written works with the current implementation -- that isn't the issue.  The issue is whether we want to hold this up as an example of how things ought to be done, and we want to support doing it this way long term.

hesingh:
If I do that, the code gets ugly.  I have to use a Cast.  See changed code below.

```
control ingress(inout headers hdr, inout metadata_t meta,
    bit<1> res = (bit<1>)(hdr.ipv6.dstAddr[127:120] == 8w0xff);    <===
    action set_mcast_grp(bit<16> mcast_grp, bit<9> port) {
        standard_metadata.mcast_grp = mcast_grp;
	meta.egress_port = port;
    }
    table ipv6_tbl {
	key = {
            res : exact @name("mcast_key");  <===
	}
	actions = {set_mcast_grp;}
    }

    apply {
        if (hdr.ipv6.isValid()) {
	    ipv6_tbl.apply();
        }
    }
}
```

Regarding the way of doing things, here is the code from typeChecker.cpp.

https://github.com/p4lang/p4c/blob/master/frontends/p4/typeChecking/typeChecker.cpp#L3390

A boolean is legal for a table key expression.  Thus, the test uses a valid table key expression.  Then, it's only if STF accepts a boolean as 0x1 or not.  STF does.  

Maybe, when Mihai gets back from the road, he can weigh in from a compiler perspective.

jafingerhut:
Well, there is more than one way to do it, of course.  No need to use a cast if you consider it ugly for some reason (I personally don't).

One way:
```
control ingress(inout headers hdr, inout metadata_t meta,
    bit<1> ipv6_da_is_mcast;
    action set_mcast_grp(bit<16> mcast_grp, bit<9> port) {
        standard_metadata.mcast_grp = mcast_grp;
	meta.egress_port = port;
    }
    table ipv6_tbl {
	key = {
            ipv6_da_is_mcast : exact;
	}
	actions = {set_mcast_grp;}
    }

    apply {
        if (hdr.ipv6.isValid()) {
            ipv6_da_is_mcast = (bit<1>) (hdr.ipv6.dstAddr[127:120] == 0xff);
	    ipv6_tbl.apply();
        }
    }
}
```
another:
```
            ipv6_da_is_mcast = (hdr.ipv6.dstAddr[127:120] == 0xff) ? 1 : 0;
```
another:
```
            // after initializing ipv6_da_is_mcast to 0 where it is declared
            if (hdr.ipv6.dstAddr[127:120] == 0xff) {
                ipv6_da_is_mcast = 1;
            }
```
These are all just minor variations, of course.  Happy to wait until someone else chimes in on the desirability of P4_16 code examples with boolean table key fields.  I just wanted to suggest an alternative where I do not think anyone would object.

hesingh:
I did a p4runtime API generation test as well in the past, but didn't add unit test results.  Here they are:

```
hemant@ubuntu:~/p4c/build$ ./p4c-bm2-ss --std p4-16 ../testdata/p4_16_samples/ipv6-switch-ml-bmv2.p4 --p4runtime-files tmp.json
hemant@ubuntu:~/p4c/build$ 
```

`tmp.json` is generated which has the following json for the boolean `mcast_key`.

```
{
 "pkgInfo": {
  "arch": "v1model"
 },
 "tables": [
  {
   "preamble": {
    "id": 33606302,
    "name": "ingress.ipv6_tbl",
    "alias": "ipv6_tbl"
   },
   "matchFields": [
    {
     "id": 1,
     "name": "mcast_key",   <===
     "bitwidth": 1,
     "matchType": "EXACT"
    }
   ],
```

hesingh:
I have shown a unit test for p4runtime.  Please check-in this code if there is no word from Antonin by end of today - thanks.

jafingerhut:
If you want me to merge it in with no word from Antonin, then this earlier comment of mine still applies: https://github.com/p4lang/p4c/pull/1969#issuecomment-502249528

Meaning, if you change the program to use type `bit<1>` in the table search key field, I have no problem making the decision on my own to merge it in, without further feedback.  It is completely non-controversial in that case.

hesingh:
Got it, Andy.  I can wait for Mihai to check-in.  The used of Boolean table key is fine for p4c and also for p4runtime.

jafingerhut:
The current p4c generation of P4Info files might produce useful results, but the P4Runtime v1.0 specification does not allow boolean types as fields of table keys.

You can show me what the implementation does all day long, and I believe you 100%.  That isn't the issue.  The issue is: Is this a case where the implementation is running too far ahead of the spec, so that _we should not yet recommend to others that they write their code this way_?

hesingh:
Wow, got it @jafingerhut.  In that case, I am in no rush.  

I think Antonin  is working on p4runtime1.0+ software.  He should weigh in.  If he agrees to support such a table key in 1.0+, Antonin can also collate with client (ONOS controller) folks to add support.

thanks.

antoninbas:
@jafingerhut I think we should update the spec for P4Runtime v1.0.0 so that boolean values are treated as `bit<1>` for match fields and action parameters. There is existing code that relies on this, since it is very common to match on header validity (e.g. https://github.com/opennetworkinglab/onos/blob/master/pipelines/fabric/src/main/resources/include/control/filtering.p4#L54).

jafingerhut:
@antoninbas Sounds reasonable to me.  A couple of questions, and I can create a proposed PR for the P4Runtime spec:

+ I suppose the idea is that if a table search key field has type boolean, then as far as the P4Info message/file and the control software is concerned, it cannot be distinguished from a P4 program that had a table search key field with that name, and type `bit<1>` instead?

+ Did you mean to update the spec for P4Runtime v1.0.0?  Or the next published version, which perhaps will be numbered v1.1.0?  It seems a bit weird to change a published version of a document.

antoninbas:
@jafingerhut I meant P4Runtime v1.0.1. I think we will soon be in a position to release it (as soon as the master arbitration changes go through).

jafingerhut:
Created a PR for the P4Runtime API spec here: https://github.com/p4lang/p4runtime/pull/207

Likely that or something substantially similar will be merged into that spec, but probably only after it has been discussed at one of the P4 API working group meetings first.

mbudiu-vmw:
It looks like merging this test has broken the build.
I will submit a fix.

hesingh:
ok, thanks.

PR #1970, opened 2019-06-17
Fix for issue 1923

jafingerhut:
This is exactly the same code changes to PSA back end that @hanw wrote in this PR: https://github.com/p4lang/p4c/pull/1964

The only difference is that the test program added in this PR is a different one.  It passes with these changes, and fails without them.


jafingerhut:
I am perfectly happy if Han updates PR https://github.com/p4lang/p4c/pull/1964 to use this test program instead of the failing one it has now.  If so, that PR should be merged and this one closed.  I only created this one in hopes of getting the existing fix merged in.

PR #1975, opened 2019-06-20
Support function-type arguments in IR generator

liujed:
This adds support for template arguments of the form `T(T1, ..., Tn)`.


mbudiu-vmw:
Functions are not first-class values in P4. There may be passes in the compiler which assume this implicitly.

Also, introducing this type without any programs or implementing the type inference algorithms handling this type is not very useful.

liujed:
@mbudiu-vmw To be clear, this adds function-type support to the IR generator, not the IR itself. This allows us to write IR methods that take in a `std::function<void(int)>`, for instance.

mbudiu-vmw:
OK, that's different.
It looks fine from my pov, but probably @ChrisDodd should review this, since he wrote most of the IR generator.

PR #1982, opened 2019-07-01
experimental feature: allowing constructor parameter to be non-constant

hanw:
This feature is experimental, disabled by default and does not comply with the current P4-16 spec.

It violates the principle that there shall be no reference to field in P4-16. However, I hope we can allow this experimental feature to exist in the front-end to allow me to experiment with some non-spec-compliance way of writing P4-16. =)

One example that I have in mind:
```
/// extern requires two parameters.
/// partition_index: run-time value of a field.
/// number_partitions: compile-time constant.
extern ATCAM<T> {
   ATCAM(T partition_index, bit<32> number_partitions);
}

control ingress (header_t hdr, metadata_t md) {
  table {
    key = ...,
    actions = ...,
    implementation = ATCAM(md.meta.partition_index, 1024);
  }
}
```

I have thought about other way to implement this in P4-16.
```
/// use an extern function
extern void atcam(in T partition_index, bit<32> number_partitions);

control ingress (header_t hdr, metadata_t md) {
  table {
    key = ...,
    actions = ...,
    implementation = atcam(md.meta.partition_index, 1024); (this is a bit strange, because return value is void).
  }
}
```

We currently use annotation to solve the issue.
```
control ingress (header_t hdr, metadata_t md) {
  @atcam_partition_index(md.meta.partition_index)
  @atcam_number_partitions(1024)
  table {
    key = ...,
    actions = ...
  }
}
```

### DIFF
--- a/frontends/p4/evaluator/evaluator.cpp
+++ b/frontends/p4/evaluator/evaluator.cpp
@@ -129,7 +129,8 @@ Evaluator::evaluateArguments(
         CHECK_NULL(folded);
         visit(folded);  // recursive evaluation
         if (!hasValue(folded)) {
-            ::error("%1%: Cannot evaluate to a compile-time constant", arg->expression);
+            if (!typeMap->isCompileTimeConstantCheckDisabled())
+                error("%1%: Cannot evaluate to a compile-time constant", arg->expression);
             popBlock(context);
             return nullptr;
         } else {
--- a/frontends/p4/frontend.cpp
+++ b/frontends/p4/frontend.cpp
@@ -114,7 +114,8 @@ class FrontEndDump : public PassManager {
 
 // TODO: remove skipSideEffectOrdering flag
 const IR::P4Program *FrontEnd::run(const CompilerOptions &options, const IR::P4Program* program,
-                                   bool skipSideEffectOrdering) {
+                                   bool skipSideEffectOrdering,
+                                   bool disableCompileTimeConstantCheck) {
     if (program == nullptr)
         return nullptr;
 
@@ -122,6 +123,10 @@ const IR::P4Program *FrontEnd::run(const CompilerOptions &options, const IR::P4P
     ReferenceMap  refMap;
     TypeMap       typeMap;
     refMap.setIsV1(isv1);
+    // XXX(hanw) experimental feature
+    // set true if constructor parameter does not have to be compile-time
+    // constant.
+    typeMap.disableCompileTimeConstantCheck(disableCompileTimeConstantCheck);
 
     auto evaluator = new P4::EvaluatorPass(&refMap, &typeMap);
 
--- a/frontends/p4/frontend.h
+++ b/frontends/p4/frontend.h
@@ -40,7 +40,8 @@ class FrontEnd {
     }
     void addDebugHook(DebugHook hook) { hooks.push_back(hook); }
     const IR::P4Program* run(const CompilerOptions& options, const IR::P4Program* program,
-                             bool skipSideEffectOrdering = false);
+                             bool skipSideEffectOrdering = false,
+                             bool disableCompileTimeConstantCheck = false);
 };
 
 }  // namespace P4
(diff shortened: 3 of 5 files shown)


hesingh:
Personally, I don't like using Annotations to solve any issue in P4.  p4c should have an intelligent backend to solve issues but you are only adding code to p4c frontend - this is another issue because a feature should develop any code for bmv2 backend as well.   By just writing frontend code, all issues may not be flushed out.  

However, I am neutral and not blocking this work.

hanw:
I don't like using annotation to solve issue in p4c either. The proposal here is to put in an experimental feature in the frontend to allow me to refer to a field in extern constructor parameter (assuming name is in scope). Currently, such code fails to type check in the frontend. If proven useful, maybe one day this feature will be included in the spec, or maybe it will never be.

I provided the annotation approach as a negative example of how things are done today, hopefully gives reviewers an idea of why the feature is needed.

mbudiu-vmw:
The constructor does not return void, it returns an instance of the extern object.

mbudiu-vmw:
The spec says that constructors are evaluated at compile-time. They direct the compiler to do resource-allocation. If you need to do something that depends on runtime values then these should be passed as arguments to functions.

If I can guess what you are trying to do, then the most natural way is to define a new property of the table "index" and use the expression for that purpose.

```
table {
    key = ...,
    actions = ...,
    index = md.meta.partition_index;
  }
```

mbudiu-vmw:
That may not work, since the compiler may not know that the table call will evaluate the "index" expression.
The only safe way to do it is part of the key:
```
table {
   key = { ..., 
                 md.meta.partition_index : exact @index;
            }
   ...
}
```

hanw:
> The spec says that constructors are evaluated at compile-time. They direct the compiler to do resource-allocation. If you need to do something that depends on runtime values then these should be passed as arguments to functions.

This is true, but what I don't understand is why the resource-allocation has to be enforced by the frontend (in this case, the evaluator). When the name of a field is passed as the constructor parameter of an extern instance, it should be the job of the backend to ensure that the field is used appropriately for resource allocation, no? Maybe I am just confused about the division of labor between front and backend.

mbudiu-vmw:
The front-end does not do any resource allocation, but it does type-checking, and also def-use analysis, and perhaps other dataflow analyses. The front-end assumes that constructor calls are evaluated at compile-time, so it does not include them in any of these analyses.

hanw:
okay, suppose in this case, the information that the extern actually needs from the parameter is the 'name' of the field and perhaps the width of the field, are they compile-time constant? The name is a constant string, and width is a constant int. Everything can be evaluated at compile-time, isn't it?

hanw:
We don't support string in P4, maybe that is the root of the issue with supporting the reference to a field?

jafingerhut:
I have nothing against loosening of restrictions in the compiler to enable experiments in various back ends to extend functionality.

Is the root of this issue the following?

* How can one add more behavior to a `my_table.apply()` call, such that the additional behavior uses the run-time value of one or more fields as inputs, even if those fields are not one of the table's search key fields?

That is one way that I think accurately describes the collection of issues raised for the P4_16 language sometimes called virtual or abstract functions.  There is a list of several related Github issues in the row of the table for topic "Abstract methods in extern objects" here: https://github.com/p4lang/p4-spec/wiki/Language-Design-Working-Group

If this proposal has a different goal than that, feel free to correct me where I've missed the issue's intent.

hanw:
The intent of this PR is to loosen the restrictions in the compiler frontend to enable experiment with extern interface design. 

At the moment, there are at least three separate externs that can be designed more intuitively if the restriction is removed.

jafingerhut:
Out of curiosity, are all three of the externs you have in mind directly associated with tables, similar to how the one in your example is?  In that case, it seems that you want the run-time value of the variable to be read at the time of a `my_table.apply()` call, true?

If some of your examples involve externs that can be used independent of tables, when would you want the run-time value of those variables to be accessed by the extern?  Only when a method of the extern object is called?  If not only then, at what other times?

vgurevich:
Why not to do it this way:
```
match_kind {
     ...
     atcam_partition_index
}

control ingress (header_t hdr, metadata_t md) {
  table {
    key = {
         md.meta.partition_index : atcam_partition_index;  //a new match_kind
         //other key fields
     }
    actions = ...,
    implementation = atcam(1024); 
}
```

mbudiu-vmw:
I would also suggest closing this PR

PR #1988, opened 2019-07-09
Fix construction of bitref or bitranges object with a temp bitvec to not dangle

ChrisDodd:
- obscure corner case -- when a ctor is called with an rvalue ref to a
  temp, the lifetime of that temp is only until the end of the ctor --
  we need to have it persist until the dtor is called.

### DIFF
--- a/lib/bitrange.h
+++ b/lib/bitrange.h
@@ -43,6 +43,7 @@ class bitranges {
 
  public:
     explicit bitranges(const bitvec &b) : bits(b) {}
+    explicit bitranges(bitvec &&b) : tmp(b), bits(tmp) {}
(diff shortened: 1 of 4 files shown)


cc10512:
This would be an ideal candidate for adding some gtests.

mbudiu-vmw:
cpplint does not like this change

ChrisDodd:
cpplint was failing to recognize the `&&` as an rvalue reference modifer on the method rather than a logical and operator, so I modified the script for this case (its a pretty hacky collection of regexes to classify things).  Also added gtest

mbudiu-vmw:
Is there a newer version of cpplint we can rebase our prior changes on?
Or perhaps we can contribute this fix back as well?

PR #1991, opened 2019-07-10
Fix broken build

mbudiu-vmw:
Change the reference output of a test.


mbudiu-vmw:
The mac build has stalled in some download step.
If the Ubuntu test passes it would be useful to merge this to fix the build...

mbudiu-vmw:
The Ubuntu test has passed, and the MacOS test has failed due to unknown reasons.
It would be still nice to merge this to fix the broken build and allow other PRs to be rebased.

PR #1993, opened 2019-07-10
Fix gc.cpp tracing

ChrisDodd:
This was broken awhile back when logging was reorganized -- the caching routines can't be called from the gc_callback, so the LOG macros can't be used there -- and was never fixed properly.

### DIFF
--- a/lib/gc.cpp
+++ b/lib/gc.cpp
@@ -23,6 +23,7 @@ limitations under the License.
 #include "log.h"
 #include "gc.h"
 #include "cstring.h"
+#include "n4.h"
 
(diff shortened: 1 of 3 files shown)


lib/gc.cpp
```diff
@@ -23,6 +23,7 @@ limitations under the License.
 #include "log.h"
 #include "gc.h"
 #include "cstring.h"
+#include "n4.h"
```
mbudiu-vmw: where is this file?

ChrisDodd: in lib/n4.h

mbudiu-vmw: Sorry, I missed that.

PR #1994, opened 2019-07-10
type checker fix for inferring type parameter of extern

hanw:
I tried to use ExternInstance::resolve() to collect information on an extern that looks like this:

```
DirectCounter<W> {
    DirectCounter();
}

DirectCounter<bit<32>>() dc;
```

But the ExternInstance struct does not contain any info about `bit<32>`. After some debugging, I tracked the issue to typeChecker, where it removes the generic parameter, but did not replace it with the specialized type. Hence the fix. I am not sure why typeChecker is implemented in this way initially.

### DIFF
--- a/frontends/p4/typeChecking/typeChecker.cpp
+++ b/frontends/p4/typeChecking/typeChecker.cpp
@@ -295,7 +295,7 @@ const IR::Type* TypeInference::specialize(const IR::IMayBeGenericType* type,
 
     LOG2("Translation map\n" << bindings);
 
-    TypeVariableSubstitutionVisitor tsv(bindings);
+    TypeVariableSubstitutionVisitor tsv(bindings, true);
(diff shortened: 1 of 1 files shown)


hanw:
The problem seems to be related to a inconsistency between the type of a `Declaration_Instance` in the typeMap and the type of the `Declaration_Instance` in the IR.

https://github.com/p4lang/p4c/blob/master/frontends/p4/externInstance.cpp#L59

mbudiu-vmw:
With the change you made most programs do not compile.
ExternInstance::resolve is really independent on the type checker, it just tries to use the type information to gather everything related to an extern instance in a single place.

The bit<32> will be visible in the type field of the Declaration_instance; that type is a Type_Specialized node.

hanw:
I found a way to get the info i need, will submit a different PR.

PR #1995, opened 2019-07-10
emit braces for ifs; fixes #1992

### LINKED ISSUE #1992: Dangling else
Hi, I have a problem with a dangling else when converting from p4-14 to p4-16. The converter essentially modifies the semantics of the initial p4-14 program in a subsequent pass.
Sample p4-14 code (dangling.p4):
```
header_type data_t {
    fields {
        f1 : 32;
        f2 : 32;
    } }
action _nop() {}
metadata data_t m;
table t1 { actions { _nop; } }
table t2 { actions {  _nop; } }
parser start {
    return ingress; }
control ingress {
    if (m.f1 == 32w0) {
        if (m.f2 == 32w0) {
            apply(t1);
        }
    } else { 
        apply(t2);  
    }
}
```
`p4test --top4 FrontEndLast --std p4-14 dangling.p4 && mv dangling-*FrontEndLast* dangling16.p4` produces the following ingress code:
```
if (meta.m.f1 == 32w0) 
   if (meta.m.f2 == 32w0) 
      t1_0.apply();
else
   t2_0.apply()
```
which contains a dangling else. When running again `p4test --top4 FrontEndLast dangling16.p4` the ingress code looks like this:
```
if (meta.m.f1 == 32w0) 
   if (meta.m.f2 == 32w0) 
       t1.apply();
   else 
       t2.apply()
```
which is different from the original p4 14 program.

Quick workaround: modified ToP4 pass to add braces to if and else whenever the inner statement is not a BlockStatement and now it appears to work

mbudiu-vmw:
Unfortunately this changes lots of reference outputs.


frontends/p4/toP4/toP4.cpp
```diff
@@ -1051,25 +1051,35 @@ bool ToP4::preorder(const IR::IfStatement* s) {
     visit(s->condition);
     builder.append(") ");
     if (!s->ifTrue->is<IR::BlockStatement>()) {
+        builder.append("{");
```
ChrisDodd: Only actually necessary if s->ifTrue is an IR::IfStatement with ifFalse == nullptr?  I guess the redundant braces are not too ugly though (and are probably clearer even when ifFalse != nullptr)

frontends/p4/toP4/toP4.cpp
```diff
@@ -1051,25 +1051,35 @@ bool ToP4::preorder(const IR::IfStatement* s) {
...
         builder.decreaseIndent();
+        builder.emitIndent();
+        builder.append("}");
+    }
     if (s->ifFalse != nullptr) {
         builder.newline();
         builder.emitIndent();
         builder.append("else ");
         if (!s->ifFalse->is<IR::BlockStatement>()) {
+            builder.append("{");
```
ChrisDodd: Might be even cleaner to make the if (on lines 1070 and 1077) `if (!s->ifFalse->is<IR::BlockStatement>() && !s->ifFalse->is<IR::IfStatement>())`.  That way chained if ... else if ... else if ... else if ... else ... will line up nicely rather than with increasing indent...

frontends/p4/toP4/toP4.cpp
```diff
@@ -1051,25 +1051,35 @@ bool ToP4::preorder(const IR::IfStatement* s) {
...
-    if (!s->ifTrue->is<IR::BlockStatement>())
+    if (!s->ifTrue->is<IR::BlockStatement>()) {
+        builder.newline();
         builder.decreaseIndent();
+        builder.emitIndent();
+        builder.append("}");
+    }
     if (s->ifFalse != nullptr) {
         builder.newline();
         builder.emitIndent();
```
ChrisDodd: maybe remove these two, so `else` ends up on the same line as the previous `}`...

mbudiu-vmw:
I have made the changes suggested by @ChrisDodd

PR #2000, opened 2019-07-11
Warn on unsupported use of recirculate/resubmit/clone

mbudiu-vmw:
As discussed in the last LDWG, the bmv2 back-end currently does not implement correctly recirculate, resubmit and clone3. So we give a warning to the users. This would be ideally a temporary solution until we provide a working implementation of these primitives. A solution, proposed in #1698, has not been adopted yet, and no one has proposed an implementation for an alternative solution.

### DIFF
--- a/backends/bmv2/common/JsonObjects.cpp
+++ b/backends/bmv2/common/JsonObjects.cpp
@@ -51,6 +51,18 @@ JsonObjects::JsonObjects() {
     field_aliases = insert_array_field(toplevel, "field_aliases");
 }
 
+Util::JsonArray*
+JsonObjects::get_field_list_contents(unsigned id) const {
+    for (auto e : *field_lists) {
+        auto obj = e->to<Util::JsonObject>();
+        auto val = obj->get("id")->to<Util::JsonValue>();
+        if (val != nullptr && val->isNumber() && val->getInt() == (int)id) {
+            return obj->get("elements")->to<Util::JsonArray>();
+        }
+    }
+    return nullptr;
+}
+
 Util::JsonObject*
 JsonObjects::find_object_by_name(Util::JsonArray* array, const cstring& name) {
     for (auto e : *array) {
--- a/backends/bmv2/common/JsonObjects.cpp
+++ b/backends/bmv2/common/JsonObjects.cpp
@@ -56,7 +56,7 @@ JsonObjects::get_field_list_contents(unsigned id) const {
     for (auto e : *field_lists) {
         auto obj = e->to<Util::JsonObject>();
         auto val = obj->get("id")->to<Util::JsonValue>();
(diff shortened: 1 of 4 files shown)


jafingerhut:
Is this intended to warn when compiling P4_14 programs, too, or only when the program provided by the user is written in P4_16?

mbudiu-vmw:
They all use this backend

mbudiu-vmw:
Travis has completed but github doesn't seem to notice that...

jafingerhut:
I forced it to restart the build.  We'll see if Github notices when it finishes this time...

edgar-costa:
Why would the warning of using `clone3` (same applies for resubmit and recirculate) with a non empty parameter be :`clone3: clone with non-empty argument not supported`. Why no supported? I mean, its currently the only way we have to pass our metadata through the new packet cycle or replica, why would that be considered unsupported if it still works (for some cases only) and its currently the only way of copying the meta. Isn't that warning misleading for people that do not know about this issue?

jafingerhut:
I believe the intended meaning of "not supported" is that if it does not work, yeah, we know about that.  If it happens to work in your case, you got lucky.

jafingerhut:
That said, a link to this issue in the error message might help clarify things for people who want more details.

edgar-costa:
I would add a link in the error message. However, I still think that `not supported` can be understood as if you are doing something wrong that you should not do, yet passing metadata to clone3, recirculate and resubmit is supported in v1model, and it was always supported until this bug found.  So, indeed some clarification would be helpful for the people that do not follow the issues and might be quite confused when they see that warning in their newer versions of p4c.

jafingerhut:
So there is no paid support model for p4c that I am aware of, but there is support in the sense that if someone uses p4c and finds a bug in something that they and the p4c authors believe ought to work, it is somewhat likely that one or more of the p4c developers will devote some time to fix the problem, or suggest workarounds.

After significant time looking at the root cause of why recirculate, resubmit, and clone3 operations with metadata does not work in some cases in v1model, and multiple hours of discussion and thought by p4c developers of how it could be made to work 100% of the time, we came to the conclusion that it is so much effort to fix it, that we will work on other issues instead.  Anyone who wants to become educated on the p4c internals and wants to devote time to it is welcome to propose fixes, of course, but the people who traditionally enhance and make fixes to p4c are not planning to do so, even though they know about it.  In that sense, it is not supported.

jafingerhut:
If anyone wants to dig deeper on what the p4c developers considered to actually fix this issue, rather than say it is not supported, probably these two comments, and perhaps earlier ones on the same issue, are a good place to read as many gory details as you want:

* https://github.com/p4lang/p4c/issues/1669#issuecomment-510717499
* https://github.com/p4lang/p4c/issues/1669#issuecomment-511067622

PR #2002, opened 2019-07-12
Add documentation of assert and assume statements

jafingerhut:


jafingerhut:
@jnfoster @liujed Please take a look and see if this documentation for assert/assume looks reasonable to you.

jafingerhut:
@mbudiu-vmw I know this will fail the automated runs until after I update some expected output for v1model.p4 line numbers.  Will wait to update those until after getting review comments for the added text.

jafingerhut:
@enisah Thank you once again for implementing this feature in p4c and behavioral-model.  Sending you this notice of adding documentation for this feature, in case you have any comments or corrections.

jafingerhut:
Hmm, I guess this _didn't_ cause any need to update expected output files, because the changes are so late in v1model.p4 that no error messages reference line numbers after this.  So it is good to go.

enisah:
> @enisah Thank you once again for implementing this feature in p4c and behavioral-model. Sending you this notice of adding documentation for this feature, in case you have any comments or corrections.

Thank you for the notice! I don't have any comment, it looks good to me.

PR #2005, opened 2019-07-17
Support for zeros_as_ones

AashleshaPatil-BFN:
Added an annotation to indicate use of csum16_udp in p4-14  to activate zeros_as_ones feature.


hesingh:
Do we have to support csum16_tcp and csum16_ip?

AashleshaPatil-BFN:
> Do we have to support csum16_tcp and csum16_ip?

No. csum16_udp is used to indicate that zeros_as_ones feature is in use. (This is a feature is used for protocols like udp where optional checksums are used) No need to create or support csum16_tcp and csum16_ip.

hesingh:
ok, thanks.

PR #2006, opened 2019-07-19
Support serializable enums in const table entries

antoninbas:
Most of the changes are in the p4RuntimeSerializer. One change was
required in type unification to support:
`((bit<16>)MyEnum2B.MBR2) &&& 0xff00`

`MyEnum2B.MBR2 &&& 0xff00` gives a compiler error, but AFAIK that 
behavior conforms to the P4_16 spec.


antoninbas:
@mbudiu-vmw I'm not sure I should file an issue, because I don't think it violates the spec, which says:

> In this way, an enum with an underlying type can be thought of as being a type derived from the underlying type carrying equality, assignment, and casts to/from the underlying type.

It doesn't say anything about being able to apply operators such a `&&&` directly to an enum member.

PR #2009, opened 2019-07-26
keep annotation when converting action selector

hanw:


hanw:
wasn't able to get the append to work.

PR #2014, opened 2019-08-01
fix shadow warning issue1172

hanw:


mbudiu-vmw:
I am not sure what you mean by "proper scoping". As far as I know the name resolution is always based on proper contexts. This is about not giving warnings when a name shadows an existing name in the context in a case which is considered benign.

cc10512:
What I meant is that based on context, we can use the same name for two entities without considering them shadowed. While in this context it is not possible to determine whether W defines an int or an extern, since it is only an interface, there are many other cases in which we should be able to disambiguate based on context whether a name represents an object type or another. I guess, we should put a formal proposal in front of the language working group.

PR #2015, opened 2019-08-01
refactoring to allow overriding the default p4-14 converter 

hanw:
This PR is to enable p4 implementer to override the default implementation for P4-14 conversion to v1model, to enable P4-14 conversion to other target architecture directly. It may help simplify the implementation by avoiding the two steps translation : p4-14 -> v1model -> vendor architecture to one-step translation p4-14 -> vendor architecture. For example, it will allow us to implement p4-14 -> psa conversion directly.

It changes the `parseV1Program` function from

```
template <typename Input>
static const IR::P4Program*
parseV1Program(Input& stream, const char* sourceFile, unsigned sourceLine,
               boost::optional<DebugHook> debugHook = boost::none) {
```

to

```
template <typename Input, typename C = P4V1::Converter>
static const IR::P4Program*
parseV1Program(Input& stream, const char* sourceFile, unsigned sourceLine,
               boost::optional<DebugHook> debugHook = boost::none) {
```

It also moves the following passes from converters.cpp to converters.h to be reused by other user-defined P4-14 converters.

```
class DiscoverStructure : public Inspector
class ComputeCallGraph : public Inspector
class ComputeTableCallGraph : public Inspector
class Rewriter : public Transform
class FixExtracts final : public Transform
class AdjustLengths : public Transform
class DetectDuplicates: public Inspector
class FixMultiEntryPoint : public PassManager
class MoveIntrinsicMetadata : public Transform
```

In pariticular, MoveIntrinsicMetadata is a v1model-specific pass which is not required for other targets. Other targets can also add more post-translation fix passes after the initial p4-14 translation, before Frontend starts.


mbudiu-vmw:
It's not very easy to review this, since a lot of code has been moved between files.
Can you point out the important substantive changes?

hanw:
@mbudiu-vmw  I edited the PR note.

frontends/p4/fromv1.0/converters.h
```diff
@@ -158,11 +158,771 @@ class PrimitiveConverter {
     const IR::Statement *PrimitiveConverter_##NAME##_##__VA_ARGS__::convert(                    \
         ProgramStructure *structure, const IR::Primitive *primitive)
 
-// Is fed a P4-14 program and outputs an equivalent P4-16 program
-class Converter : public PassManager {
+///////////////////////////////////////////////////////////////
+
+class DiscoverStructure : public Inspector {
```
ChrisDodd: I'm assuming this large change is just moving code from .cpp to .h without changing it at all.  Would be nice if github reviews gave a way of seeing the differences (if any) here.

PR #2017, opened 2019-08-06
Implementation of string type

mbudiu-vmw:


frontends/p4/typeChecking/typeChecker.cpp
```diff
@@ -263,8 +263,9 @@ bool TypeInference::checkParameters(
             typeError("%1%: parameter cannot be a package", p);
             return false;
         }
-        if (p->direction != IR::Direction::None && type->is<IR::Type_Extern>()) {
-            typeError("%1%: a parameter with an extern type cannot have a direction", p);
+        if (p->direction != IR::Direction::None &&
+            (type->is<IR::Type_Extern>() || type->is<IR::Type_String>())) {
+            typeError("%1%: a parameter with type %2% cannot have a direction", p, type);
```
ChrisDodd: Can a type parameter used for an argument with a direction be inferred as Type_String?  What haapens for code like

        extern bit<16> hash<T>(in T data);
        hash("testing...");

should probably be a reasonable error message.

mbudiu-vmw:
@ChrisDodd : I added one more negative test for the case you suggested. It works.

PR #2018, opened 2019-08-06
Inherit all operations for serEnum from underlying type

mbudiu-vmw:
This allows a value of type serEnum participate in all operations that a value of the underlying type supports.


mbudiu-vmw:
Addresses https://github.com/p4lang/p4-spec/issues/774

mbudiu-vmw:
@vgurevich : please let us know if this implementation solves the problems you have raised.

antoninbas:
I'd like to merge https://github.com/p4lang/p4c/pull/2089 first. Then once this PR is rebased, if none of the tests fail, I will have high confidence that this change will not break P4Info generation.

mbudiu-vmw:
Actually let's wait a bit for this, since it's too lenient according to the spec. It allows implicit casts in both directions, but the spec only allows them in only one direction. So it has to be fixed a bit.

antoninbas:
@mbudiu-vmw sounds good, ping me when you need a review

mbudiu-vmw:
The LDWG suggested this fix in the spec: https://github.com/p4lang/p4-spec/pull/841

mbudiu-vmw:
Fixed by https://github.com/p4lang/p4c/pull/2018, which should adhere to the spec.

mbudiu-vmw:
although this has been rebased on master it's missing a constant in v1model in all reference files: `+const bit<32> __v1model_version = 32w20180101;` I don't know why the repository reference files do not contain this constant.

mbudiu-vmw:
In fact, this constant should not appear at all in the reference output. But on my machine it does.
It looks like a bug in the handling of the source line numbers; for some reason, the constant declaration for __v1_model_version, which is in v1model.p4 shows up as being in the including file. This seems to be a bug in the processing of `#line` directives. This only happens if the constant declaration is the last one in v1model.p4.

mbudiu-vmw:
In fact, this constant should not appear at all in the reference output. But on my machine it does.
It looks like a bug in the handling of the source line numbers; for some reason, the constant declaration for __v1_model_version, which is in v1model.p4 shows up as being in the including file. This seems to be a bug in the processing of `#line` directives. This only happens if the constant declaration is the last one in v1model.p4.

p4include/v1model.p4
```diff
@@ -55,6 +55,8 @@ limitations under the License.
 #define V1MODEL_VERSION 20180101
 #endif
 
+const bit<32> __v1model_version = V1MODEL_VERSION;
+
```
ChrisDodd, line 59: On my machine, putting this in the beginning of the file causes it to misbehave, and output this in the dump files (incorrectly), which is why I put it at the end originally...

mbudiu-vmw: I will put it in the middle. But it would be nice to understand why this happens.
The input from cpp seems correct.

mbudiu-vmw:
Also fixed #2329

mbudiu-vmw:
If we can review this we can hopefully put the corresponding language change in P4 1.2.1.

ir/expression.def
```diff
@@ -272,6 +272,7 @@ class Slice : Operation_Ternary {
 }
 
 class Member : Operation_Unary {
+    precedence = DBPrint::Prec_Postfix;
```
fruffy, line 275: Operation_Unary also includes Neg, Cmpl, LNot, which are prefixes, aren't they?

mbudiu-vmw: Yes, but this field is just for Member.

fruffy: Oh for some reason I read this as the Operation Unary definition... sorry about that.

mbudiu-vmw:
This PR fixes several problems:
- allows implicit casts for serializable enums
- adds ErrorType to all ::error calls made by the constant folding engine
- fixes the priority of the Member IR class for printing
- makes a &&& b have the type of a; this is important when a is a serializable enum

It would be great to get a review so we can merge this for P4 1.2.1.
@ChrisDodd ?

mbudiu-vmw:
I think that this PR is also ready for review. The list of changes is in the above comment.

PR #2022, opened 2019-08-10
Do not constant fold on the lhs of =

### LINKED ISSUE #2021: Confusing error message
Compiling this program
```
#include <core.p4>

control C();
package S(C c);
const bit<8> x = 8w1;
control MyC() {
  apply {
    .x = 8w3;
  }
}
S(MyC()) main;
```
generates the following error message:
```
[--Werror=type-error] error: Expression 1 cannot be the target of an assignment
const bit<8> x = 8w1;
                 ^^^
```
I can imagine why the compiler is treating 1 as the left-hand side of an assignment, but it's pretty confusing...

mbudiu-vmw:
Fixes #2021

### DIFF
--- a/frontends/common/constantFolding.cpp
+++ b/frontends/common/constantFolding.cpp
@@ -78,7 +78,7 @@ const IR::Expression* DoConstantFolding::getConstant(const IR::Expression* expr)
 }
 
 const IR::Node* DoConstantFolding::postorder(IR::PathExpression* e) {
-    if (refMap == nullptr /*|| leftValue*/)
+    if (refMap == nullptr || leftValue)
         return e;
     auto decl = refMap->getDeclaration(e->path);
     if (decl == nullptr)
--- a/frontends/common/constantFolding.cpp
+++ b/frontends/common/constantFolding.cpp
@@ -78,7 +78,7 @@ const IR::Expression* DoConstantFolding::getConstant(const IR::Expression* expr)
 }
 
 const IR::Node* DoConstantFolding::postorder(IR::PathExpression* e) {
-    if (refMap == nullptr)
+    if (refMap == nullptr /*|| leftValue*/)
         return e;
     auto decl = refMap->getDeclaration(e->path);
     if (decl == nullptr)
@@ -162,6 +162,25 @@ const IR::Node* DoConstantFolding::postorder(IR::Declaration_Constant* d) {
     LOG3("Constant " << d << " set to " << init);
     constants.emplace(getOriginal<IR::Declaration_Constant>(), init);
     return d;
+}
+
+const IR::Node* DoConstantFolding::preorder(IR::AssignmentStatement* statement) {
+    leftValue = true;
+    visit(statement->left);
+    leftValue = false;
+    visit(statement->right);
+    prune();
+    return statement;
+}
+
+const IR::Node* DoConstantFolding::preorder(IR::ArrayIndex* e) {
+    visit(e->left);
+    bool save = leftValue;
+    leftValue = false;
+    visit(e->right);
+    leftValue = save;
+    prune();
+    return e;
 }
 
 const IR::Node* DoConstantFolding::postorder(IR::Cmpl* e) {
--- a/frontends/common/constantFolding.cpp
+++ b/frontends/common/constantFolding.cpp
@@ -78,7 +78,7 @@ const IR::Expression* DoConstantFolding::getConstant(const IR::Expression* expr)
 }
 
 const IR::Node* DoConstantFolding::postorder(IR::PathExpression* e) {
-    if (refMap == nullptr || leftValue)
+    if (refMap == nullptr || assignmentTarget)
         return e;
     auto decl = refMap->getDeclaration(e->path);
     if (decl == nullptr)
@@ -165,9 +165,9 @@ const IR::Node* DoConstantFolding::postorder(IR::Declaration_Constant* d) {
 }
 
 const IR::Node* DoConstantFolding::preorder(IR::AssignmentStatement* statement) {
-    leftValue = true;
+    assignmentTarget = true;
     visit(statement->left);
-    leftValue = false;
+    assignmentTarget = false;
     visit(statement->right);
     prune();
     return statement;
@@ -175,10 +175,10 @@ const IR::Node* DoConstantFolding::preorder(IR::AssignmentStatement* statement) 
 
 const IR::Node* DoConstantFolding::preorder(IR::ArrayIndex* e) {
     visit(e->left);
-    bool save = leftValue;
-    leftValue = false;
+    bool save = assignmentTarget;
+    assignmentTarget = false;
     visit(e->right);
-    leftValue = save;
+    assignmentTarget = save;
     prune();
     return e;
 }
--- a/backends/bmv2/CMakeLists.txt
+++ b/backends/bmv2/CMakeLists.txt
@@ -88,13 +88,13 @@ build_unified(BMV2_BACKEND_COMMON_SRCS)
 add_library(bmv2backend ${BMV2_BACKEND_COMMON_SRCS})
 add_dependencies(bmv2backend genIR frontend)
 
-build_unified(BMV2_SIMPLE_SWiTCH_SRCS)
+build_unified(BMV2_SIMPLE_SWITCH_SRCS)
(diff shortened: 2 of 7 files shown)


mbudiu-vmw:
Interestingly, there were some typos in the CMakefiles which disabled the unified builds...

frontends/common/constantFolding.cpp
```diff
@@ -78,7 +78,7 @@ const IR::Expression* DoConstantFolding::getConstant(const IR::Expression* expr)
 }
 
 const IR::Node* DoConstantFolding::postorder(IR::PathExpression* e) {
-    if (refMap == nullptr)
+    if (refMap == nullptr || assignmentTarget)
```
ChrisDodd: If you use P4WriteContext here, there's no need for all the extra cruft with assignmentTarget and this becomes a one-line change

    if (refMap == nullPtr || isWrite())

This will incidentally also not fold into an out or inout param of a method/function

frontends/common/constantFolding.cpp
```diff
@@ -164,6 +164,25 @@ const IR::Node* DoConstantFolding::postorder(IR::Declaration_Constant* d) {
     return d;
 }
 
+const IR::Node* DoConstantFolding::preorder(IR::AssignmentStatement* statement) {
+    assignmentTarget = true;
```
ChrisDodd: making these methods unnecessary

mbudiu-vmw:
@ChrisDodd : there may be a bug in the P4WriteContext; it prevents several constants from being folded. It does not seem to work properly for "in" parameters of functions.

mbudiu-vmw:
The problem is that when we do constant folding for the first time we do not have type information, and P4WriteContext needs type information to be precise. In the absence of that it treats method calls conservatively.

This is a benign problem, since these arguments will be folded later, but I am thinking to revert to my previous implementation.

ChrisDodd:
It seems like not folding into arguments that might be out or inout (but we don't know yet) is the right thing to do -- folding incorrectly makes sorting out what went wrong later hard.  Though `isWrite` is supposed to be 'conservative' returning `true` for unknown sounds suspect (there are probably other cases that handle "don't know yet" the same as "true" when what should be happening is a three-way return value perhaps?)

ChrisDodd:
Looking at the `isWrite` code, it does at least document what it means by "conservative" -- true just means might be written and false means definitely not written.  So that at least seems reasonable.

ChrisDodd:
Should perhaps add an error test case that trues to call an out/inout arg with a const value to ensure it gives a reasonable error?

mbudiu-vmw:
How about I rename isWrite to mayWrite?
Should I do the same for isRead?

mbudiu-vmw:
There may be another problem: it is not clear that the 'type' field of the method is actually ever set, so isWrite may be always too conservative. At least, when compiling fabric.p4, the constant folding never folds an argument to an action (set_forwarding_type) which is directionless, so compilation fails.

ChrisDodd:
That sounds like a bug in TypeChecking.

mbudiu-vmw:
Type checking never promised to set the field; it does only if you call it with the flag "updateExpressions", which is "false" most of the time.

ChrisDodd:
Just because its a longstanding bug in TypeChecking doesn;t make it any less of a bug.

mbudiu-vmw:
If there is a bug it is not in the type checker, it is in the callers of the type checker.

mbudiu-vmw:
I have reverted to the less elegant but working implementation, without isWrite.
It gives an appropriate error message for function arguments; there's a test for that.

ChrisDodd:
I would suggest abandoning this change altogether until the type problems can be fixed -- the practice of piling hack upon hack without fixing the underlying stuff is what got us into this problem in the first place...

mbudiu-vmw:
I am not aware of any hack that I may have introduced in the codebase at any time.
I also do not see how the "hacks" have caused this bug to happen. It's a bug like any other bug, partly due to insufficient testing.

I think that this PR is perfectly legitimate, and it solves the issue. 
isWrite is too conservative if it does not have type information, and in the first constant-folding stage we do not have type information yet.

ChrisDodd:
When we don't (yet) have the type information to allow the constant propagation and folding, we don't (yet) want to do the constant propagation and folding -- doing it anyways will result in problems later in the case where it turns out it was the wrong thing to do.  While the PR does solve the one (very narrow) issue, it fails with a variety of other issues.

The "hacks" that I am referring to are all the cases where type information is squirrelled away in side data structures (like the type map) rather than updating the IR to correctly reflect the results of typechecking.  While it works, its pretty fragile.

mbudiu-vmw:
I don't think this solves a narrow problem, I think that the fix is fine.
Anyway, I won't have time to work on refactoring type checking to work without a typeMap for a while. You can try to see whether the scheme will work fine by changing the "updateExpressions" in the constructor of TypeChecking to have a default value of "true". If this does not break anything there is some hope, otherwise we can find out what the issues are.

mbudiu-vmw:
@ChrisDodd: I have started working on removing the typeMap.
The "easy" way of always using ApplyTypesToExpressions actually does not work, because running this pass after typeChecking changes the program, so the typeMap is no longer valid.
So I will try to actually get rid completely of ApplyTypesToExpressions and have the typeChecker always set the "type" field of an expression.
It this causes no problems we can get rid of ApplyTypesToExpressions.
To get rid of the typeMap more work will be needed, because not only expressions have types. We will have to create some new non-visited field on several other types of nodes, similar to the 'type' field of an expression. BTW: I would like to change this field name to "canonicalType" to make it clearer that this field is special.

mbudiu-vmw:
@ChrisDodd : I have a dilemma: I would like the typechecking pass not to mutate the IR graph - except the first time, when it inserts casts. However, if the typechecker sets the 'type' field of an expression, then it always changes the IR graph, invalidating for example the reference map. Some other passes may also assume that by calling type checking the graph does not change...
How do I work around this?

ChrisDodd:
Tricky problem to do without having two versions.  I've thought about making it a template, so you could have TypeCheck<Inspector> or TypeCheck<Modifier>.  Then, you would just have two versions of each `preorder`/`postorder` function -- one that takes a `const *` (for the Inspector) and the other that takes a non-const `*` (for the Modifier).  Problem is that you'd need to duplicate all of them (though in cases where the IR did not need to be modified, you could have the Modifier version just call the Inspector version.

The other possibility is similar to the current -- have a flag in the visitor -- when set, do the inference and modify the IR; when clear do the inference and check the result is `.equiv` to what's in the IR and flag an error (BUG?) if not.  Less code duplication, but need the check of the flag all over the place.  Could also have a BUG_CHECK in the `end_apply` method that checks the resulting tree is the same as the original from `init_apply` if the flag is clear.

mbudiu-vmw:
My question is different: To remove the typeMap I have to update the IR always to write the new types. So the typechecher will always mutate the program, invalidating whatever other maps were computed previously.

mbudiu-vmw:
@ChrisDodd : I will merge this PR. It does the job, and the alternative solution currently does not.
I have tried to remove the typeMap but I don't know how given our visitor infrastructure; if we do that we can revisit this PR.

ChrisDodd:
The alternative solution does the job AND also prevents folding into `out` and `inout` params, which is IMO better.

mbudiu-vmw:
With the alternative change a couple of tests do not pass because at the time this pass is run the type information is not available.

ChrisDodd:
My recollection is the only reason they don't pass is the expected output changes.

mbudiu-vmw:
They do not pass because constant folding does actually not constant fold some values that should be replaced with constants.

ChrisDodd:
But that's not a problem -- they'll get folded on a subsequent constant folding pass and getting them folded on the very first pass is not necessary for correctness (indeed, constant folding at all here is not  a correctness issue, merely an optimization).

mbudiu-vmw:
At least with the current structure of the compiler they were never folded. 
The reason is that in general expression->type does not hold an accurate type for the expression; some method call expressions had this field always NULL at the time of constant folding, so the compiler acted conservatively when evaluating the isWrite() method for a method call expression. 
Unfortunately if you update expression->type in the typeChecker you also mutate the graph; that's the reason I could not make the typeChecker always update types.

PR #2027, opened 2019-08-27
Fix const default action check

amreshk:
Support for declaring a constant default action using a `const` keyword was added a while back by @hanw - #1523 . Fixing check to use this parsed info within `V1Table` into table property in generated `P4Table`.


mihaibudiu (changes requested):
Why remove the previous check? 
How about a test case?

Issue #762, opened 2017-07-05
Future extension: Convenient initialization of header/structs


orenbenkiki:
Implement https://github.com/p4lang/p4-spec/issues/341 - at least as an option - to promote its acceptance into the spec. That is, allow:

```
header_or_struct = {};
```

To initialize all the leaf fields to 0 (or `false` for a Boolean, or zero length for `varbit`).

marko-stanoj:
Is there any particular reason why structs and headers, that are declared and initialized at the same time, are not processed by StructInitializer pass in frontend?

For example:
Bool_int_hdr hdr = { false, 0 };
this kind of initialization would not be processed by StructInitializer pass,

Bool_int_hdr hdr;
hdr = { false, 0 };
but this kind of initialization would.

Wouldn't it be better to visit those assignments after MoveInitializers pass, so they get the same treatment?

Second question is, is it even possible to set dynamic length of varbit in compile time and initialize it to zero length? I'm on the right track to solve this issue and answers on this question would help me a lot.

mbudiu-vmw:
Please do not ask two different questions in a single issue.
We need to run StructInitializer as early as possible. MoveInitializers has a few other dependencies.
You are welcome to try to reorganize them.
If StructInitializer does not process declaration intializers this should be a bug which you can probably highlight using a specific test case.

There is no operation to set the length of a varbit field.

mbudiu-vmw:
BTW: we should first settle on the syntax for doing initialization. I would prefer something like `{ ... }`.

Jovana-Syrmia:
I'm currently working on adjusting  PR #2368 (referencing this issue) to work with the current main.
Is default initialization addition still needed?

mbudiu-vmw:
Yes, it is still needed. Thank you.

Issue #2036, opened 2019-09-25
Type error due to inference/substitution?


jnfoster:
Compiling this program,
```p4
#include <core.p4>

//Architecture
parser C();
package S(C p);

//User Program
struct s {
  bit<8> x;
}

parser D(in s z) { 
  state start {
    transition accept;
  } 
}

parser E() {
  tuple<bit<8>> a = { 0 };
  s b = { 1 };
  D() d;
  state start {
    d.apply(a);
    d.apply(b);	
    transition accept;
  }
}

S(E()) main;
```
using `p4test` yields an error:
```
foo.p4(19): [--Werror=type-error] error: AssignmentStatement: Cannot assign a Tuple(1) to a struct s
  tuple<bit<8>> a = { 0 };
  ^^^^^^^^^^^^^
foo.p4(8)
struct s {
       ^
```
However, if we comment out either one of the `d.apply(...)` calls, it compiles fine.

My best conjecture is that type inference on the first occurrence is substituting one type for the other, which then causes a mismatch later on? 

This is also reminiscent of #1830.

mbudiu-vmw:
We should not allow a tuple to be assigned to a struct. This happens because a the type of a list expression is in fact a tuple, and we allow list expressions to be assigned to structs. But we don't allow tuples to be assigned to structs...

Probably the solution is to introduce a new type internally for list expressions which can be unified with tuples and structs.

jnfoster:
Actually, this variant of the program, which `p4test` accepts, should actually be rejected, right? Because we are using a variable whose type is a tuple in a place where a struct is expected.

```p4
#include <core.p4>

//Architecture
parser C();
package S(C p);

//User Program
struct s {
  bit<8> x;
}

parser D(in s z) { 
  state start {
    transition accept;
  } 
}

parser E() {
  tuple<bit<8>> a = { 0 };
  s b = { 1 };
  D() d;
  state start {
    d.apply(a);
    transition accept;
  }
}

S(E()) main;
```

mbudiu-vmw:
Yes, this program is incorrect.

PR #2038, opened 2019-09-30
Structs, headers and tuples initialisation support (#762) (#2017)

marko-stanoj:
Added support for struct/header/tuple initialisation with empty tuple.

Types that are supported as leaf fields are int, bool and bit.
Int and bit type fields will be initialized to zero, bool to false.
If any other type is used as a lief field compiler will report error.

along with assignment statemants, which is the cause of changes in expected oputputs
of some tests.

### DIFF
--- a/frontends/p4/frontend.cpp
+++ b/frontends/p4/frontend.cpp
@@ -148,7 +148,7 @@ const IR::P4Program *FrontEnd::run(const CompilerOptions &options, const IR::P4P
         new ValidateMatchAnnotations(&typeMap),
         new DefaultArguments(&refMap, &typeMap),  // add default argument values to parameters
         new BindTypeVariables(&refMap, &typeMap),
-        new StructInitializers(&refMap, &typeMap),
+        new StructInitializers(&refMap, &typeMap, false),
         new TableKeyNames(&refMap, &typeMap),
         // Another round of constant folding, using type information.
         new ConstantFolding(&refMap, &typeMap),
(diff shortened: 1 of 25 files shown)


mbudiu-vmw:
The language design working group has not yet converged on a syntax for initializing default fields in structures. We can use this PR as a basis for discussion, but the syntax may be different.

mbudiu-vmw:
We also have issue #2036 which may require changes to this PR (they could be part of a separate PR too).

marko-stanoj:
Hello @mbudiu-vmw ,
Thank you for your time and comments.

I apologize for not being precise enough in commit message, I've added support for struct/header/tuple initialisation with empty expression list, not tuples in general.

OK, neither different syntax, or introducing a new type for list expressions should require major changes in this pr.

marko-stanoj:
After adding a new type for list expressions, this solution requires changes, are there any news from LDWG about this theme(initialization syntax)?

mbudiu-vmw:
We haven't had a meeting since this PR; the next one is on 12/15. You are welcome to participate.
Let me know if you don't have a zoom link; it was sent on the p4-design mailing list.
BTW: I am planning to write a spec for an alternative proposal for default initializers, but some of it will apply to whatever solution we choose.

jnfoster:
12/16!

marko-stanoj:
Thank you for the invitation, it would be my pleasure to join. I'm not on p4-design mailing list, so I just sent a request.
Should part of this solution related to frontend reorganization, written to enable declaration initializers processing apart of assignment initializers in StructInitializes pass, be part of different PR?

mbudiu-vmw:
Here is a first PR against the spec to enable this: https://github.com/p4lang/p4-spec/pull/797
More to follow.

mbudiu-vmw:
This PR seems to be superseded by #2224; can we close it?

marko-stanoj:
Yes, sure, this was before dots were introduced.

PR #2039, opened 2019-09-30
fix type inference for modify_conditionally primitive

hanw:


mbudiu-vmw:
If this does not break any existing programs it's fine.

PR #2043, opened 2019-10-05
add a compiler option to disable specific annotations/pragmas

cc10512:
Add the --disable-annotations option to disable annotations by name.
It accepts a comma separated list of annotations that will not be
added in the IR tree. A warning will be printed that the annotation
was explicitly disabled. We do print the warning for every instance of
the annotation, which may be an overkill, however, it is intentional,
so that users can make sure that a particular annotation was not taken
into account.

The option only disable annotations present in the source
file. Compiler passes may continue to add annotations and process them
as before. Note that users may be able to disable annotations that may
cripple the code generation -- backend or control-plane. That's
another reason why it is important to print those warnings!

### DIFF
--- a/frontends/parsers/p4/p4parser.ypp
+++ b/frontends/parsers/p4/p4parser.ypp
@@ -39,6 +39,7 @@ limitations under the License.
 #include <iostream>  // NOLINT(build/include_order)
 
 #include "frontends/common/constantParsing.h"
+#include "frontends/common/options.h"
 #include "ir/ir.h"
 #include "lib/cstring.h"
 #include "lib/source_file.h"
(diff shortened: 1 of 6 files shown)


frontends/parsers/p4/p4parser.ypp
```diff
@@ -440,8 +441,14 @@ optAnnotations
     ;
 
 annotations
-    : annotation  { $$ = new IR::Vector<IR::Annotation>(); $$->push_back($1); }
-    | annotations annotation { $$ = $1; $$->push_back($2); }
+    : annotation  {
+       $$ = new IR::Vector<IR::Annotation>();
+       if (! P4CContext::get().options().isAnnotationDisabled($1))
```
mbudiu-vmw: wouldn't it be somewhat cleaner to write a pass to remove the annotations rather than not insert them?

cc10512: Why would that be cleaner? Especially if you want to ensure that only the user defined annotations are removed and the compiler can insert such annotations for its used, by not adding the annotations in the parser you avoid issues with pass ordering.

mbudiu-vmw: It's cleaner because the parser just parses and does not look at options.
Also, a visitor for this would be very simple indeed: just return null for the annotation that matches the name.

PR #2049, opened 2019-10-09
Change that seems to enable EBPF back end to build on Ubuntu 16.04

jafingerhut:


jafingerhut:
This is just an experimental change I tried that seems to enable the EBPF back end to build on a freshly installed Ubuntu 16.04 Linux system.  Don't trust that I know what I'm doing here :)

mbudiu-vmw:
Does this enable the kernel-based tests?
It's quite easy to write a condition to disable them, what we want is to be able to run them on systems that support it.

backends/ebpf/CMakeLists.txt
```diff
@@ -56,8 +56,8 @@ execute_process(COMMAND uname -r
   OUTPUT_STRIP_TRAILING_WHITESPACE
   RESULT_VARIABLE rc)
 message(STATUS "Detected kernel version: ${P4C_EBPF_KERNEL_VER}")
-if (EXISTS /usr/src/linux-headers-${P4C_EBPF_KERNEL_VER}/include/uapi/linux)
-  include_directories(/usr/src/linux-headers-${P4C_EBPF_KERNEL_VER}/include/uapi/linux)
+if (EXISTS /usr/src/linux-headers-${P4C_EBPF_KERNEL_VER}/include)
+  include_directories(/usr/src/linux-headers-${P4C_EBPF_KERNEL_VER}/include)
```
cc10512: I'm afraid this is not going to include the right header that EBPF needs.
Can out try:
```if (EXISTS /usr/src/linux-headers-${P4C_EBPF_KERNEL_VER}/include)
  include_directories(/usr/src/linux-headers-${P4C_EBPF_KERNEL_VER}/include /usr/src/linux-headers-${P4C_EBPF_KERNEL_VER}/include/uapi/linux)
```

mbudiu-vmw: What is the error? The compilation should never fail, running the tests may fail.

jafingerhut:
With the alternate change in the attached file patch1.txt suggested by Calin, on a freshly installed Ubuntu 16.04 Linux system, plus all dependencies recommended in the p4c README's installed, I get the error message in the attached file patch1-build-error-msg.txt

[patch1.txt](https://github.com/p4lang/p4c/files/3708666/patch1.txt)
[patch1-build-error-msg.txt](https://github.com/p4lang/p4c/files/3708667/patch1-build-error-msg.txt)

backends/ebpf/CMakeLists.txt
jafingerhut: See this comment on this same PR: https://github.com/p4lang/p4c/pull/2049#issuecomment-540106445
for attached files with what I think is a slightly different change suggested by Calin, and the error message I see when I attempt to build p4c with that different change.

mbudiu-vmw:
Hm, we should not set include_directories for gcc, only for the tests.

jafingerhut:
If you have a suggestion for how to change the p4c repo to set include directories for tests, not for gcc, I can try it out.  I don't know how to do that.

mbudiu-vmw:
I think we just have to take this line out:
`include_directories(/usr/src/linux-headers-${P4C_EBPF_KERNEL_VER}/include/uapi/linux)`
We have to do something equivalent in the python scripts which compile the resulting C files, though. I will work on it.

jafingerhut:
I have pushed another commit on this PR so that I believe it follows Mihai's suggestion of his previous comment -- removing the include_directories line.

jafingerhut:
I have confirmed that with the change proposed in the latest version of this PR (the one that only deletes one line, an `include_directories` directive), the latest version of p4c source code builds without errors on both a freshly installed & updated Ubuntu 16.04 system, and Ubuntu 18.04.

mbudiu-vmw:
Let's merge this and we'll have to do something more to make the ebpf tests properly compile on some architectures.

cc10512:
Does it also pass the tests?

mbudiu-vmw:
At least on Travis...

jafingerhut:
@cc10512 @mbudiu-vmw It also passed all p4c tests _except_ the ebpf-specific ones on my freshly installed Ubuntu 16.04 and 18.04 machines.  We shouldn't be surprised about failing the ebpf-specific ones given the lack of changes that Mihai is planning to make, and that I have done nothing special to install anything that the ebpf back end might need, except llvm.

Issue #2048, opened 2019-10-09
Ubuntu Linxu build problem for ebpf back end - missing kernel include file?


jafingerhut:
Trying to build the latest p4c source code on a freshly installed Ubuntu 16.04 Linux system, with all updates from Canonical as of 2019-Oct-09, I get an error message while it attempts to compile p4c with ebpf back end, due to a missing include file.  Full error message below.

Perhaps it requires not only the "user-facing" kernel include files, but actual full kernel source code to be installed?  I haven't yet determined what changes might enable it to compile successfully.

```
[ 80%] Building CXX object backends/ebpf/CMakeFiles/p4c-ebpf.dir/unified_p4c_ebpf_srcs_1.cpp.o
In file included from /usr/include/stdio.h:33:0,
                 from /home/jafinger/p4c/backends/ebpf/p4c-ebpf.cpp:17,
                 from /home/jafinger/p4c/build/backends/ebpf/unified_p4c_ebpf_srcs_1.cpp:2:
/usr/src/linux-headers-4.15.0-65-generic/include/uapi/linux/stddef.h:2:34: fatal error: linux/compiler_types.h: No such file or directory
compilation terminated.
backends/ebpf/CMakeFiles/p4c-ebpf.dir/build.make:62: recipe for target 'backends/ebpf/CMakeFiles/p4c-ebpf.dir/unified_p4c_ebpf_srcs_1.cpp.o' failed
make[2]: *** [backends/ebpf/CMakeFiles/p4c-ebpf.dir/unified_p4c_ebpf_srcs_1.cpp.o] Error 1
CMakeFiles/Makefile2:1597: recipe for target 'backends/ebpf/CMakeFiles/p4c-ebpf.dir/all' failed
make[1]: *** [backends/ebpf/CMakeFiles/p4c-ebpf.dir/all] Error 2
make[1]: *** Waiting for unfinished jobs....
```

cc10512:
Is that file present in any of the directories under `/usr/src/linux-headers-4.15.0-65-generic/include/`?

jafingerhut:
Yes:
```
$ find /usr/src/linux-headers-4.15.0-65 | grep compiler_types
/usr/src/linux-headers-4.15.0-65/include/linux/compiler_types.h
```

jafingerhut:
Actually, correction to my previous comment.  I was searching in a different directory than the one you asked about in my previous comment.

The correct answer is no for the directory with `-generic` in its name:
```
[...]

PR #2050, opened 2019-10-10
Disable kernel ebpf tests

mbudiu-vmw:
I don't know how to fix the building of the ebpf kernel-based tests so that they work on any kernel/distribution combination, so I am disabling them for now. @fruffy has actually created these tests. If anyone has time to work on them we can re-enable them.
We are still testing the p4c-ebpf compiler, but just with portable user-level tests.


mbudiu-vmw:
Fixes #2048

PR #2053, opened 2019-10-13
Remove use of __WORDSIZE

antoninbas:
__WORDSIZE is not well-documented and we should probably avoid using
it. I'm compiling p4c on a 64-bit system that apparently does not define
__WORDSIZE. This proposal sticks to standard C++ constructs. We define
an IR::Constant constructor manually for each constructor defined in
gmpxx, plus function template constructors for all other integral types
(signed and unsigned) which use mpz_import. On systems for which
"unsigned long long" and "long long" are different from "unsigned long"
and "long", the template constructors will be used.

### DIFF
--- a/ir/expression.def
+++ b/ir/expression.def
@@ -141,32 +141,45 @@ class Constant : Literal {
 #noconstructor
     /// if noWarning is true, no warning is emitted
     void handleOverflow(bool noWarning);
-    // We need to enumerate all the integer types because we need proper 64-bit handling on
-    // 32-bit systems (which ain't long!) and mpz_import is too big a hammer because and it loses
-    // the signess of the value.
+#emit
+    // One-one correspondence between these constructors and the ones in libgmp (gmpxx.h header).
+    Constant(char v, unsigned base = 10) :
+        Literal(new Type_InfInt()), value(v), base(base) {}
+    Constant(unsigned char v, unsigned base = 10) :
+        Literal(new Type_InfInt()), value(v), base(base) {}
+    Constant(short v, unsigned base = 10) :
+        Literal(new Type_InfInt()), value(v), base(base) {}
+    Constant(unsigned short v, unsigned base = 10) :
+        Literal(new Type_InfInt()), value(v), base(base) {}
     Constant(int v, unsigned base = 10) :
         Literal(new Type_InfInt()), value(v), base(base) {}
-    Constant(unsigned v, unsigned base = 10) :
+    Constant(unsigned int v, unsigned base = 10) :
         Literal(new Type_InfInt()), value(v), base(base) {}
-#emit
-#if __WORDSIZE == 64
-    Constant(intmax_t v, unsigned base = 10) :
-        Literal(new Type_InfInt()), value(v), base(base) {}
-#else
     Constant(long v, unsigned base = 10) :
         Literal(new Type_InfInt()), value(v), base(base) {}
     Constant(unsigned long v, unsigned base = 10) :
         Literal(new Type_InfInt()), value(v), base(base) {}
-    Constant(intmax_t v, unsigned base = 10) :
+    // For other integral types, we default to using mpz_import since there are no constructors
+    // available in gmpxx. Note that the constructors above will always take precedence (plain old
+    // functions over function template) if the type matches.
+    template<typename T,
+             typename std::enable_if<std::is_integral<T>::value &&
+                      std::is_signed<T>::value, int>::type = 0>
+        Constant(T v, unsigned base = 10) :
+        Literal(new Type_InfInt()), base(base) {
+        mpz_import(value.get_mpz_t(), /* word count */1, /* msb first */0,
+                   /* word size */sizeof(v), /* native endian */0, /* nails */0, &v);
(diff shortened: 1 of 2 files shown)


antoninbas:
I modified my solution slightly. I realized it was not ideal and would not work if for some reason `intmax_t` was neither `long` nor `long long` which is probably very unlikely. There is now a template constructor for all integral types, but the plain old constructors (which should be faster) will have precedence when the type matches.

ChrisDodd:
Seems like it should be simpler to just remove all of these constructors and just construct from an mpz_class (which should be implicitly convertable from any integer type.)  If that doesn't work, perhaps we should be changing to using boost::mulitprecision (see #1418 ) for which that does work?  That was "on pause" pending further discussion (for legal and licensing reasons) but we never moved forward on it.

cc10512:
@antoninbas I'm getting an error on a 32-bit system with g++ 5.4.0 when using the following constructor:
```
IR::Constant(IR::Type::Bits::get(4), bitvec.getrange(0, max));
```
where `uintmax_t getrange(size_t idx, size_t sz) const;` is declared in `lib/bitvec.h:273`
The error is:
```error: conversion from ‘uintmax_t {aka long long unsigned int}’ to ‘mpz_class {aka __gmp_expr<__mpz_struct [1], __mpz_struct [1]>}’ is ambiguous
In file ...
usr/include/gmpxx.h:1502:3: note: candidate: __gmp_expr<__mpz_struct [1], __mpz_struct [1]>::__gmp_expr(double)
   __GMPXX_DEFINE_ARITHMETIC_CONSTRUCTORS
   ^
/usr/include/gmpxx.h:1502:3: note: candidate: __gmp_expr<__mpz_struct [1], __mpz_struct [1]>::__gmp_expr(float)
   __GMPXX_DEFINE_ARITHMETIC_CONSTRUCTORS
   ^
/usr/include/gmpxx.h:1502:3: note: candidate: __gmp_expr<__mpz_struct [1], __mpz_struct [1]>::__gmp_expr(long unsigned int)
   __GMPXX_DEFINE_ARITHMETIC_CONSTRUCTORS
   ^
/usr/include/gmpxx.h:1502:3: note: candidate: __gmp_expr<__mpz_struct [1], __mpz_struct [1]>::__gmp_expr(long int)
   __GMPXX_DEFINE_ARITHMETIC_CONSTRUCTORS
   ^
/usr/include/gmpxx.h:1502:3: note: candidate: __gmp_expr<__mpz_struct [1], __mpz_struct [1]>::__gmp_expr(short unsigned int)
   __GMPXX_DEFINE_ARITHMETIC_CONSTRUCTORS
   ^
/usr/include/gmpxx.h:1502:3: note: candidate: __gmp_expr<__mpz_struct [1], __mpz_struct [1]>::__gmp_expr(short int)
   __GMPXX_DEFINE_ARITHMETIC_CONSTRUCTORS
   ^
/usr/include/gmpxx.h:1502:3: note: candidate: __gmp_expr<__mpz_struct [1], __mpz_struct [1]>::__gmp_expr(unsigned int)
   __GMPXX_DEFINE_ARITHMETIC_CONSTRUCTORS
   ^
/usr/include/gmpxx.h:1502:3: note: candidate: __gmp_expr<__mpz_struct [1], __mpz_struct [1]>::__gmp_expr(int)
   __GMPXX_DEFINE_ARITHMETIC_CONSTRUCTORS
   ^
/usr/include/gmpxx.h:1502:3: note: candidate: __gmp_expr<__mpz_struct [1], __mpz_struct [1]>::__gmp_expr(unsigned char)
   __GMPXX_DEFINE_ARITHMETIC_CONSTRUCTORS
   ^
/usr/include/gmpxx.h:1502:3: note: candidate: __gmp_expr<__mpz_struct [1], __mpz_struct [1]>::__gmp_expr(signed char)
   __GMPXX_DEFINE_ARITHMETIC_CONSTRUCTORS
...
ir/expression.def:187:5: note:   initializing argument 2 of ‘IR::Constant::Constant(const IR::Type*, mpz_class, unsigned int, bool)’
     Constant(const Type *t, mpz_class v, unsigned base = 10, bool noWarning = false) :
```

ChrisDodd:
Where is the actual error here?  It will be just before all those 100s of repeated 'note: candidate' messages.  It looks like it is failing to instantiate the enable_if template constructor properly (needed to initialize an mpz_class from any type larger than `long` since GMP does not provide a direct constrruction for those)

cc10512:
I remember pasting that in as well ...:
```error: conversion from ‘uintmax_t {aka long long unsigned int}’ to ‘mpz_class {aka __gmp_expr<__mpz_struct [1], __mpz_struct [1]>}’ is ambiguous```

antoninbas:
I didn't touch the constructor whose first parameter is a `Type *`, so it's not obvious to me how the error is related. Any idea?

mbudiu-vmw:
Should we update or close this PR?

jafingerhut:
@antoninbas @ChrisDodd Is this PR still relevant?  I see that #1418 was mentioned in the comment conversation above, and has been merged, but I do not know if that covers everything this PR was intended to do.

antoninbas:
Given how old this is, I'll just close it.

PR #2061, opened 2019-10-16
Make argument in only; fixes #2057

### LINKED ISSUE #2057: Can't add more than one verify_checksum in Checksum Verification block
I want to add checksum verification for two different headers.  Here it is the code in the Checksum Verification block of my bmv2 simple switch:

```p4
/*************************************************************************
************   C H E C K S U M    V E R I F I C A T I O N   *************
*************************************************************************/

control MyVerifyChecksum(inout headers hdr,
                        inout metadata meta) {
    apply {

/*
* Verify checksum for EFCP packets
*/
            verify_checksum(hdr.efcp.isValid(),
            { hdr.efcp.vers,
	          hdr.efcp.dstAddr,
              hdr.efcp.srcAddr,
              hdr.efcp.qosID,
              hdr.efcp.dstCEPID,
              hdr.efcp.srcCEPID,
              hdr.efcp.pduType,
              hdr.efcp.flags,
              hdr.efcp.length,
              hdr.efcp.seqnum },
            hdr.efcp.hdrChecksum,
            HashAlgorithm.csum16);

/*
* Verify checksum for IPv4 packets
*/

            verify_checksum(hdr.ipv4.isValid(),
            { hdr.ipv4.version,
	          hdr.ipv4.ihl,
              hdr.ipv4.diffserv,
              hdr.ipv4.totalLen,
              hdr.ipv4.identification,
              hdr.ipv4.flags,
              hdr.ipv4.fragOffset,
              hdr.ipv4.ttl,
              hdr.ipv4.protocol,
              hdr.ipv4.hdrChecksum,
              hdr.ipv4.srcAddr,
              hdr.ipv4.dstAddr
              },
            hdr.ipv4.hdrChecksum,
            HashAlgorithm.csum16);

    }
}
```

This is the compiler output that I have:

```
[--Werror=legacy] error: AssignmentStatement: Only calls to verify_checksum or verify_checksum_with_payload allowed
```
I think it should be no problem for calling `verify_checksum` more than one time, but actually the compiler does not allow me to do it. I am missing something?

Thank you in advance

mbudiu-vmw:


mbudiu-vmw:
This change actually has no impact as far as I can tell. As a proof, all tests have passed.
Changing the control prototype is very disruptive, though.

jnfoster:
Ugh. But how does `verify_checksum` implement the behavior described in the comment?

> If this method detects that a checksum of the data is not correct, then the value of the standard_metadata checksum_error field will be equal to 1 when the packet begins ingress processing.

I don't see an `inout` parameter of type `standard_metadata_t` here... :-/

mbudiu-vmw:
This is not very pretty, but the standard_metadata is defined by the architecture, and it *can* be changed between control blocks. This works because the current control block (verify_checksum) cannot actually refer to the metadata, so it is fine for this block to actually change the value of the metadata seen by the next block. This is similar to the parser_error field.

jafingerhut:
Adding a little to what Mihai said, it is perfectly allowed for an extern function to modify state within the architecture that is not directly visible to the P4 program.  As long as we say that the checksum_error field within standard metadata cannot be read by the user's program until the ingress control begins, after the VerifyChecksum control is finished, the architecture can read that architecture-private state modified by any verify_checksum calls performed earlier on the packet, and use that to assign the correct value to the standard_metadata.checksum_error field that is sent into the ingress control.

I am pretty sure this is what I was thinking when I wrote the following comment for the behavior of the verify_checksum extern function:


Verifies the checksum of the supplied data.  If this method detects
 * that a checksum of the data is not correct, then the value of the
 * standard_metadata checksum_error field will be equal to 1 when the
 * packet begins ingress processing.

Note that it does not say "immediately changes checksum_error to 1", but "will be equal to when the packet begins ingress processing".

p4include/v1model.p4
```diff
@@ -415,7 +415,7 @@ extern Checksum16 {
  *                   may be supported).  Must be a compile-time
  *                   constant.
  */
-extern void verify_checksum<T, O>(in bool condition, in T data, inout O checksum, HashAlgorithm algo);
+extern void verify_checksum<T, O>(in bool condition, in T data, in O checksum, HashAlgorithm algo);
```
jafingerhut: If we make this change, we should also make the corresponding change to the signature of function verify_checksum_with_payload.

jnfoster:
I agree that an `extern` can modify state that is not visible to the P4 program. However, in V1 model, the standard metadata is exposed to other controls. Here is the snippet from the spec that constrains what effects as `extern` can have (and that caused us to change the `mark_to_drop` extern):

> In general, extern functions are arbitrarily powerful: they can store information in global storage, spawn separate threads, “collude” with each other to share information — but they cannot access any variable in a P4 program. With copy-in/copy-out semantics the compiler can still reason about P4 programs that invoke extern functions. 

I suppose there is a lawyerly reading by which the standard metadata maintained by the architecture that is silently manipulated by `VerifyChecksum` is not the same as the `standard_meta` variable exposed to `Parser`, `Ingress`, etc. 

As @mbudiu-vmw said, this is not very pretty...

mbudiu-vmw:
In this case we are indeed not modifying any program variables, we are modifying the metadata in transit between two other blocks that produce and consume it.

jafingerhut:
There is nothing in the P4_16 language spec that says what the inputs to top level controls are, is there?  That is what is defined by the architecture (I thought).

If we say for v1model "the value of standard_metadata.checksum_error sent into the ingress control for a packet is equal to 1 if any of the verify_checksum* extern function calls during the previous VerifyChecksum control was executed found an error, otherwise 0", then that is the behavior we have with v1model now.  To implement this, the verify_checksum* extern functions only need to modify state hidden from the user's P4 program inside the hidden architecture state.

If we forget for a moment about all the issues around preserving user-defined metadata with resubmitted packets, and just focus on "which packets get resubmitted, vs. which do not?"  The answer is "the ones for which a resubmit() extern function call was performed while the ingress control was being executed".

Is there any state change that one can read in a v1model P4 program that tells you, during ingress execution, whether that same packet earlier in its ingress execution called resubmit()?  No, there is not.  It is hidden from the P4 program, in the architecture's private "per packet" state about each packet.  The architecture can view it when ingress processing is complete, and steer the packet to its next processing 'place' after that.

jafingerhut:
A response to this statement: "I suppose there is a lawyerly reading by which the standard metadata maintained by the architecture that is silently manipulated by VerifyChecksum is not the same as the standard_meta variable exposed to Parser, Ingress, etc."

The state modified by verify_checksum* calls during VerifyChecksum control execution _is not_ the standard_metadata struct at all.  The architecture can wait until VerifyChecksum control execution is complete, see what standard_metadata values came out of there, modify them however it is defined to do, then send that modified version into the ingress control.

v1model and PSA architectures _explicitly, by design_ control what values get sent into parsers and controls as inputs, and they _need not be_ exact copies of what was output with that packet from the previous control/parser.

mbudiu-vmw:
Yes, and this is made possible by the fact that standard_metadata is standardized.
It would more difficult to allow the architecture to manipulate headers, for example - because we have no promises on what the type of the headers is.
This is also why it's not easy to write the semantics of P4 programs - a lot happens in parts which are not programmable. And something different may happen for each architecture. And because these architectures are not programmable we don't have a choice of what to do about their behavior - it is given to us.

jnfoster:
@jafingerhut Yes, yes, I understand and I agree with your response. I was wrong to suggest that this requires a particular reading of "cannot access any variable in a P4 program." Mea culpa.

However, I would argue that having hidden side-effects like this is not the best design, especially when those effects are on state that is maintained by the architecture exposed to the programmer in other controls. It's subtle for programmers and it's fragile. For example, if a compiler pass wanted to inline the `VerifyChecksum` block into either the `Parser` or `Ingress` controls, the result would not be legal P4_16 code, because then the `verify_checksum` extern would be invoked in a scope where `standard_metadata` is a variable. That is, we'd be back in the situation we had with `mark_to_drop` where we were forced to make a chance.

Wouldn't it be a better design, and one more consistent with the rest of `v1model.p4` be to have `verify_checksum` (and also the `VerifyChecksum` controll) take `standard_metadata` as an argument, or at least a `bool` flag indicating if a checksum error has been found? Or are there reasons to prefer our current design?

jafingerhut:
If we were creating v1model from scratch, I would have no objections to have verify_checksum with a different API that explicitly returned whether it found an error, rather than behave as it does now.  It is certainly a more explicit API with less hidden machinery involved.

As a side point (not the main point, which is above): With the current verify_checksum behavior, I believe it could still be in-lined in the parser without breaking any P4_16 language restrictions.  The 'magic' happens between VerifyChecksum and ingress controls right now.  I don't see any hidden machinery involved between the parser and VerifyChecksum control, assuming that somehow the VerifyChecksum computations are all restricted to be performed after parsing is complete.

jnfoster:
Ahh, true. Again this is all so icky my ivory-tower brain can't imagine doing that!

To be precise, to satisfy the restrictions on `externs` quoted above, we'd have to say that `verify_checksum` doesn't actually modify the state stored in `standard_metadata` from the parser, but rather some other hidden version of that state. And the two versions get reconciled between the parser and ingress. Sigh.

PR #2068, opened 2019-10-26
remove test_args from p4c_add_test_with_args, always pass args throug…

hanw:
…h cmake_args

Use could add multiple tests for the same p4 program, the corresponding *.test script is also generated multiple times, with the latter one overwrites the earlier ones. As a result, if the earlier test uses a different test_args than the latter ones, the test_args will be lost due to overwrite. 

the solution is always pass argument to the *.test script, instead of saving the argument in the *.test script. For example, `myp4program.test -arg0 -arg1`.


ChrisDodd:
It seems like this change just broke all the .test scripts?  (or at least a lot of them) -- it used to be the case that you could run a test by running the .test script (without having to use the whole heavyweight ctest infrastructure), particularly useful for running a test under gdb.  But that no longer works as the .test scripts are now missing necessary arguments?

hanw:
relying on .test script to contain all necessary parameters inside it does not work for us, as we tends to generate multiple tests for the same program with different parameters (e.g., different #defines), which means in the end, the last .test script generated will persist on disk, and earlier ones will be overwritten.

mbudiu-vmw:
You should still provide an easy way for people to rerun manually a test.

hanw:
`ctest -R testname -V` which will print the .test script with appropriate parameters

ChrisDodd:
That is the entire point of the .test scripts -- if you can't use them to run the tests manually, why even bother to create them at all?

ChrisDodd:
An alternative if you really need to have only one .test script for multiple tests is to put only the common arguments in the .test script and the arguments that differ between tests supplied by ctest.  Klunky and harder to use, but would work.

mbudiu-vmw:
@chrisdodd, feel free to revert this merge

PR #2070, opened 2019-10-29
Throw error in case tuple type to struct like type conversion #2036

marko-stanoj:
Corner case in which tuple type expressions, that are not list expression,
were unifiable with struct like types fixed.

### DIFF
--- a/frontends/p4/typeChecking/typeChecker.cpp
+++ b/frontends/p4/typeChecking/typeChecker.cpp
@@ -3002,15 +3002,26 @@ const IR::Node* TypeInference::postorder(IR::MethodCallExpression* expression) {
         auto rettype = new IR::Type_Var(IR::ID(refMap->newName("R"), nullptr));
         auto args = new IR::Vector<IR::ArgumentInfo>();
         bool constArgs = true;
+        size_t i = 0;
         for (auto aarg : *expression->arguments) {
             auto arg = aarg->expression;
             auto argType = getType(arg);
             if (argType == nullptr)
                 return expression;
+            // Unification allows tuples to be assigned to headers and structs.
+            // We should only allow this if the source expression is a ListExpression.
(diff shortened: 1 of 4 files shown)


marko-stanoj:
There was already support for handling illegal assignments of tuple types (non list expression) to struct like types (structs and headers), as you can see in tuple-to-header.p4 test. Also I don't think that in, out and inout arguments affect type checking, but I can surely add more tests if you think it's necessary.
These changes handle the corner case if method is called once, or more times with same parameters, so there's no new assignment statements, of actual to formal method arguments, added after inlining pass where error would be reported. 

The reverse, assigning structs and headers to tuple, should be handled by deleting these lines of code:

`if (src->is<IR::Type_Struct>() || src->is<IR::Type_Header>()) {
            // swap and try again: handled below
            return unify(errorPosition, src, dest, reportErrors);
 }`
if destintation type is tuple type
in  `bool unify(const IR::Node* errorPosition, const IR::Type* dest,
               const IR::Type* src, bool reportErrors)` method of TypeUnification class.
This will result by treating struct like types as any other data type that is not tuple which is in conformance with specification.
But if you do so 99 of existing tests will fail with message "Cannot unify tuple type with non tuple-type struct". So I suppose that assigning tuples to struct like types should be referenced to another issue.

mbudiu-vmw:
I actually think that the right fix is to introduce a new type for the list initializer and to allow only that to unify with structs and tuples, while forbidding both directions for structs and tuples. That should take care of all cases. Do you feel like reworking the pr that way?

marko-stanoj:
Splitting tuples and list initializer into two different data types would require a lot of changes of existing code, but wouldn't bring new functionalities. With existing constraints and these changes, direction tuples to structs is already forbidden and problem with opposite direction, I mentioned in previous comment, would stay either new type is added or not.

mbudiu-vmw:
Should we close this?

marko-stanoj:
Yes, there's no need for this solution any more.

PR #2085, opened 2019-11-14
Add a Type_List for list expressions; fixes #2036

### LINKED ISSUE #2036: Type error due to inference/substitution?

mbudiu-vmw:
This generalizes the solution in #2070.
The solution is to assign to tuple Type_Tuple and to lists Type_List. Both extend Type_BaseList.
Type_List can be however unified with both Type_Tuple and Type_Struct, while Type_Tuple cannot be unified with Type_Struct.
There are quite a few changes because many passes which were looking at Type_Tuple now have to handle Type_List or Type_BaseList instead.

### DIFF
--- a/control-plane/p4RuntimeSerializer.cpp
+++ b/control-plane/p4RuntimeSerializer.cpp
@@ -1147,7 +1147,7 @@ class P4RuntimeAnalyzer {
                     [](cstring name) { return name == IR::Annotation::matchAnnotation; });
                 addDocumentation(match, f);
             }
-        } else if (et->is<IR::Type_Tuple>()) {
+        } else if (et->is<IR::Type_BaseList>()) {
             ::error(ErrorType::ERR_UNSUPPORTED,
                     "type parameter for Value Set; "
                     "this version of P4Runtime requires the type parameter of a Value Set "
--- a/control-plane/typeSpecConverter.cpp
+++ b/control-plane/typeSpecConverter.cpp
@@ -121,7 +121,7 @@ bool TypeSpecConverter::preorder(const IR::Type_Name* type) {
 bool TypeSpecConverter::preorder(const IR::Type_Newtype* type) {
     if (p4RtTypeInfo) {
         bool orig_type = true;
-        const IR::StringLiteral* uri;
+        const IR::StringLiteral* uri = nullptr;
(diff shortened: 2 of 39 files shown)


marko-stanoj:
Any news about this PR? It seems like a great solution to make type checking easier and code cleaner, especially with this few changes. Also, I think I have a solution that should go on top of this PR and should fix problem with struct to list unification we discussed.

ChrisDodd:
So I probably didn't review this enough first time through, as I've come up with a couple of questions trying to merge this with some of our backend code.

- what exactly is the difference between Type_List and Type_Tuple and which is used in different contexts?  It seems that sometimes a ListExpression has Type_List and sometimes Type_Tuple -- what does the difference signify?

- when dumping P4 code, both Type_Tuple and Type_List are dumped as `tuple<...>`, so dumping and reloading will convert Type_List into Type_Tuple.  Does this indicate a problem?

mbudiu-vmw:
Semantically they are the same, but Type_List is only used as the type of a list expression. A list expression can be assigned to either a struct or a tuple, so essentially we do not commit to which one it is until we see how the list is used. The difference between Type_Tuple and Type_List is that Tuple does not unify with Type_Struct, but List does.

Since the Type_List is only assigned as the type of a list expression, it can never appear in a type declaration, it is only created by the type inference.

PR #2089, opened 2019-11-18
Fix SerEnum member values in generated P4Info

antoninbas:
The value should be a P4Runtime "bytestring" and not a human-readable
hexadecimal string.

Thanks to @mbudiu-vmw for his help making the SerEnum Google Test work.

### DIFF
--- a/test/gtest/p4runtime.cpp
+++ b/test/gtest/p4runtime.cpp
@@ -34,6 +34,7 @@ limitations under the License.
 #include "control-plane/typeSpecConverter.h"
 #include "frontends/common/parseInput.h"
 #include "frontends/common/resolveReferences/referenceMap.h"
+#include "frontends/common/resolveReferences/resolveReferences.h"
 #include "frontends/p4/parseAnnotations.h"
 #include "frontends/p4/typeChecking/typeChecker.h"
 #include "frontends/p4/typeMap.h"
@@ -1406,7 +1407,8 @@ class P4RuntimeDataTypeSpec : public P4Runtime {
         auto pgm = P4::parseP4String(programStr, CompilerOptions::FrontendVersion::P4_16);
         if (pgm == nullptr) return nullptr;
         PassManager  passes({
-            new P4::TypeChecking(&refMap, &typeMap)
+            new P4::ResolveReferences(&refMap),
+            new P4::TypeInference(&refMap, &typeMap, false)
         });
         pgm = pgm->apply(passes);
         return pgm;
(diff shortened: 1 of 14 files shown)


test/gtest/p4runtime.cpp
```diff
@@ -1660,6 +1660,36 @@ TEST_F(P4RuntimeDataTypeSpec, Enum) {
     EXPECT_EQ("MBR_2", it->second.members(1).name());
 }
 
+// Currently throws an exception:
```
antoninbas: @mbudiu-vmw I could use your help with this test.

I thought I would add a test for serializable enums, just like I wrote a while ago for enums, but the test throws a Compiler Bug in the type checker for some reason. If I write a similar program and run it through `p4test`, there is no issue. The different is that these tests only run the `TypeChecking` pass (see line 1409) and not the whole frontend. Is there another pass we should run? What's so special about serializable enums here?

testdata/p4_16_samples_outputs/v1model-digest-containing-ser-enum.p4.p4info.txt
```diff
@@ -263,15 +263,15 @@ type_info {
...
       }
       members {
         name: "bar"
-        value: "17"
+        value: "\021"
       }
       members {
         name: "gah"
-        value: "42"
+        value: "*"
```
jafingerhut: Where is this "*" coming from?  Judging by the changes several lines above, shouldn't it be `"\052"`, i.e. escaped octal for the decimal value 42?

jafingerhut: Oh, I guess judging from the previous file's diffs, `"*"` has the ASCII value of decimal 42?  Kind of confusing the first time you see this.

antoninbas: yes these values prioritize size over human-readability :)

I got confused when I reviewed my diff as well...

mbudiu-vmw:
how can cpplint complain for unified builds but not for non-unified builds?

mbudiu-vmw:
@antoninbas : here is a fix that makes it work.
Normally the typechecker pass behaves like a Visitor, but you need to make it behave like a Transform, because it needs to change some nodes in the IR. That's the 'false' third argument.

mbudiu-vmw:
https://github.com/p4lang/p4c/commit/ac366556e288fc2d29d9485100050b1a74e49d06

antoninbas:
Thanks a lot @mbudiu-vmw, I appreciate you taking the time to look into this.

I'm guessing this line is mutating the IR so SerEnum members: https://github.com/p4lang/p4c/blob/master/frontends/p4/typeChecking/typeChecker.cpp#L1243

mbudiu-vmw:
Yes, inferring a type for int constants is one of the most frequent reasons for the typechecker to mutate the program.

antoninbas:
@mbudiu-vmw regarding the cpplint error, it only shows up for the unified builds because cpplint is not run for non-unified builds.

PR #2091, opened 2019-11-20
Insert cast for signed slices; fixes #2090

### LINKED ISSUE #2090: Type error in Bmv2 MidEnd?
Compiling the attached [switch.p4.txt](https://github.com/p4lang/p4c/files/3869430/switch.p4.txt) using the following command
```
p4c-bm2-ss --p4v 16 -Tpass_manager:1 switch.p4 
```
produces a type error:
```
...
BMV2::SimpleSwitchMidEnd invoking ExpandEmit
  ExpandEmit invoking P4::TypeChecking
    P4::TypeChecking invoking ResolveReferences
    P4::TypeChecking invoking P4::TypeInference
[--Werror=legacy] error: AssignmentStatement: Cannot unify bit<32> to int<32>
```

mbudiu-vmw:


mbudiu-vmw:
The test submitted with #2090 does not compile because it uses conditionals in actions. That's a feature that is currently not yet implemented.

midend/expandLookahead.cpp
```diff
@@ -56,7 +56,11 @@ const IR::Expression* DoExpandLookahead::expand(
     } else if (type->is<IR::Type_Bits>() || type->is<IR::Type_Boolean>()) {
         unsigned size = type->width_bits();
         BUG_CHECK(size > 0, "%1%: unexpected size %2%", type, size);
-        auto expression = new IR::Slice(base->clone(), *offset - 1, *offset - size);
+        const IR::Expression* expression =
+                new IR::Slice(base->clone(), *offset - 1, *offset - size);
+        auto tb = type->to<IR::Type_Bits>();
+        if (tb && tb->isSigned)
+            expression = new IR::Cast(tb, expression);
```
ChrisDodd: Why doesn't it need the cast if the type is Boolean?  I guess we have no tests with extracting `bool` fields?  Do we even allow `bool` (as opposed to `bit<1>` in headers?

mbudiu-vmw: good point, added a new commit to handle this case.

hesingh: If you see section 7.2.7 of P4-16 working spec, you will see the header supports a `bool`.

PR #2092, opened 2019-11-20
Fix for python3

mbudiu-vmw:


mbudiu-vmw:
Can someone review this quickly?
I think it is responsible for some xdp tests failing.

PR #2096, opened 2019-11-21
Fix FindPythonModule to avoid warnings

### LINKED ISSUE #2095: cmake failing on Ubuntu 18.04 because python links to python 2.7 by default
This is the error I get:
```
CMake Error at /usr/share/cmake-3.10/Modules/FindPackageHandleStandardArgs.cmake:137 (message):
  Could NOT find PythonInterp: Found unsuitable version "2.7.15", but
  required is at least "3" (found /usr/bin/python)
  /usr/share/cmake-3.10/Modules/FindPackageHandleStandardArgs.cmake:376 (_FPHSA_FAILURE_MESSAGE)
  /usr/share/cmake-3.10/Modules/FindPythonInterp.cmake:152 (FIND_PACKAGE_HANDLE_STANDARD_ARGS)
  CMakeLists.txt:78 (find_package)


-- Configuring incomplete, errors occurred!
See also "/home/abas/p4lang/p4c/build/CMakeFiles/CMakeOutput.log".
See also "/home/abas/p4lang/p4c/build/CMakeFiles/CMakeError.log".
```

I am using a not too old VM running Ubuntu 18.04. I believe that it is fairly common for even recent distributions to have `python` default to `python2.7` even when `python3` is installed. Could `CMakeLists.txt` check for `python3` and not just `python`? It seems that it is fairly easy to do with recent cmake versions (https://cmake.org/cmake/help/latest/module/FindPython3.html#module:FindPython3) but I couldn't find an easy way to do it with pre 3.12 versions with a quick Google search.

antoninbas:
Not sure what the purpose of all the find_python_module directives is,
since we only check for presence of standard library stuff. We also
use find_python_module to "find" the "re" module, and the
find_python_module implementation itself imports "re" :)

Fixes #2095

### DIFF
--- a/cmake/FindPythonModule.cmake
+++ b/cmake/FindPythonModule.cmake
@@ -7,7 +7,7 @@ function(find_python_module module)
     # A module's location is usually a directory, but for binary modules
     # it's a .so file.
     execute_process(COMMAND "${PYTHON_EXECUTABLE}" "-c"
-      "import re, ${module}; print re.compile('/__init__.py.*').sub('',${module}.__file__)"
+      "import re, ${module}; print(re.compile('/__init__.py.*').sub('',${module}.__file__))"
       RESULT_VARIABLE _${module}_status
       OUTPUT_VARIABLE _${module}_location
       ERROR_QUIET OUTPUT_STRIP_TRAILING_WHITESPACE)


cc10512:
We should add the non-standard modules that should be installed and remove the standard ones.

PR #2098, opened 2019-11-23
Without these changes, cmake fails to find the Python3 modules

jafingerhut:


jafingerhut:
It seems that the build succeeds without these changes, so I am not sure what good it does for cmake to look for these Python modules, if failing to find them still allows the build to succeed.  Still, finding them seems better than failing to find them.

antoninbas:
This seems to be a duplicate of my PR: https://github.com/p4lang/p4c/pull/2096. I forgot to merge it earlier, but it's done now.

jafingerhut:
Yep, sorry for the duplicate.  Will close this now.

PR #2103, opened 2019-12-04
Return a BlockStatement instead of a Vector; fixes #2102

### LINKED ISSUE #2102: Compiler Bug: visitor returned non-Statement type
While generating random P4 code we encountered this error in the compiler. If header assignments are placed within an if-statement the compiler emits a bug after the simplify_comparisons pass. This does not happen if the assignment is not within an if-statement. 
Here is a minimal example to trigger this crash.

```
#include <core.p4>

header H {
    bit<1> a;
}

struct headers {
    H h;
}

control c(inout headers hdr) {
    apply {
        if (hdr.h.a < 1) {
            hdr = hdr;
        }
    }
}

control e<T>(inout T t);
package top<T>(e<T> e);

top(c()) main;
```

[log.txt](https://github.com/p4lang/p4c/files/3923005/log.txt)

mbudiu-vmw:


mbudiu-vmw:
This one should be very easy to review if anyone has a minute.

PR #2107, opened 2019-12-07
Track unreachable statements; fixes #2104

### LINKED ISSUE #2104: Compiler Bug: no definitions
There is another issue we keep encountering while generating programs. When passing variables to actions or methods by reference, the compiler sometimes emit "no definitions" for a variable. I do not exactly understand the key factor behind the bug but we have two different version where this occurred:

Using local actions:
```
#include <core.p4>
#include <v1model.p4>

control c() {
    bit<16> FHaSkb = 0;
    bit<128> YIafjU = 0;
    action rxNET() {
        YIafjU = (bit<128>) FHaSkb;

    }
    action vKNsz() {
        return; // the return comes before the action call
        rxNET();
    }
    apply {
        vKNsz();
    }

}

control e<T>();
package top<T>(e<T> e);

top(c()) main;
```
Using global methods:
```
#include <core.p4>
#include <v1model.p4>

// adding the inout qualifier leads to a compiler crash
bit<8> test(inout bit<8> x) {
    return x;
}

header H {
    bit<8> a;
    bit<8> b;
}

struct headers {
    H h;
}

control c(inout headers hdr) {
    apply {
        test(hdr.h.a);
    }
}

control e<T>(inout T t);
package top<T>(e<T> e);

top(c()) main;

```

The first program is nonsensical but still causes a crash. @mbudiu-vmw  What do you think?

mbudiu-vmw:


mbudiu-vmw:
The def-use analysis used to analyze unreachable statements too, but that's wrong, since they have no effect. Also, return used to kill all definitions, including the ones from outer scopes. That was wrong too.

hesingh:
With code changes which definitions are not being killed?  I know the ones from outer scope are not.   Does it make sense to document the behavior in p4-16 specification?

mbudiu-vmw:
This is all standard behavior, i am just fixing a bug.

mbudiu-vmw:
I think that we already do have an unreachable code elimination pass, but it probably runs after this one. Even if you do eliminate unreachable code, other optimizations may generate more unreachable code, e.g. constant folding if expressions. So I think that in general we need to be conservative in the def-use analysis and handle unreachable code as well.

PR #2108, opened 2019-12-09
[WIP -- testsing]: refactoring sideEffects.cpp to fix problems

ChrisDodd:
- don't pull hit/miss out of an if, as it will just be converted into a (redundant) table and temp by
  actionSynthesis, so is a complete waste
- WIP -- needs refactoring of DismantleExpression/DoSimplifyExpressions
  into a single pass to get the context properly.
- don't introduce unnecessary block statements.
- don't try to create Type_Dontcare temps


frontends/p4/sideEffects.cpp
```diff
@@ -436,19 +446,18 @@ class DismantleExpression : public Transform {
...
             visit(arg);  // May mutate arg!  Recursively simplifies arg.
             auto newarg = result->final;
             CHECK_NULL(newarg);
 
             if (useTemp) {
                 // declare temporary variable
-                auto tmp = refMap->newName("tmp");
+                auto paramtype = typeMap->getType(p, true);
+                if (paramtype->is<IR::Type_Dontcare>())
+                    paramtype = typeMap->getType(arg, true);
```
mbudiu-vmw: you may want to apply getP4Type to the paramType. This will, for example, convert a Type_Struct to a Type_Name that refers to the struct, which is what you should use for a declaration.

frontends/p4/sideEffects.cpp
```diff
@@ -172,6 +173,15 @@ class DismantleExpression : public Transform {
             // the MethodCallExpression.
             if (TableApplySolver::isHit(expression, refMap, typeMap) ||
                 TableApplySolver::isMiss(expression, refMap, typeMap)) {
+                if (!getContext()) {
```
mbudiu-vmw: is this the problem or is the bug where the second table reference is introduced?
In general, the number of table invocations on each program path should not be changed by the compiler, so perhaps there is a bug somewhere?
Which pass introduces an extra table access?

frontends/p4/sideEffects.cpp
ChrisDodd: createTemporary already calls getP4Type on the type passed in -- just trying to remove redundant repeated calls here.

frontends/p4/sideEffects.cpp
```diff
@@ -600,6 +608,8 @@ const IR::Node* DoSimplifyExpressions::postorder(IR::AssignmentStatement* statem
     auto right = parts->final;
     CHECK_NULL(right);
     parts->statements->push_back(new IR::AssignmentStatement(statement->srcInfo, left, right));
+    if (parts->statements->size() == 1)
```
mbudiu-vmw: these blocks will be eliminated later. Is there a good reason to avoid creating them altogether?

ChrisDodd: Wasteful churn slowing down the compiler, making it harder to debug what is going on when trying to track where stuff is changing across all the passes of the compiler.

frontends/p4/sideEffects.cpp
ChrisDodd: ActionSynthesis creates new tables and actions for assignments in a control block.  So we end up with

    if (table.apply().hit) {
        t1.apply();  // run action { tmp = true; }
    } else {
        t2.apply(); // run action { tmp = false; }
    }
    if (tmp) {
        ...

and the apply().hit is back inside an if, and we have two extra (useless) tables.

mbudiu-vmw:
If looks like in at least one case you are giving the bmv2 back-end a pattern it does not know how to handle.

ChrisDodd:
Yes, looks like the bmv2 backend has trouble with `if (!table.apply().hit)`.

backends/p4test/run-p4-sample.py
```diff
@@ -336,8 +346,10 @@ def main(argv):
                 argv = argv[1:]
         elif argv[0][1] == 'D' or argv[0][1] == 'I' or argv[0][1] == 'T':
             options.compilerOptions.append(argv[0])
-        elif argv[0] == "-gdb":
+        elif argv[0][0:4] == "-gdb":
```
mbudiu-vmw: what happens if this is not long enough?

ChrisDodd: If the argument is shorter, the slice will get the entire arg (so the test will be false, but that is correct)

PR #2109, opened 2019-12-09
Handle MUX in FindUses; fixes #2105

### LINKED ISSUE #2105: Compiler Bug:  no locations known for <Mux>
This is a weird one.
I do not understand what is going on here but this statement leads to another crash. I reduced it to a minimum but all the components (the slice and the bor) are required to generate this crash. 

```
#include <core.p4>
#include <v1model.p4>

control c() {
    apply {
        bit<8> x = 0;
        bit<8> y = 0;
        // the crash happens when reassigning y while referencing x
        // removing either the slice or the bor operation will fix the crash
        y = (x < 4 ? 8w2 : 8w1)[7:0] | 8w8;
    }
}

control e<T>();
package top<T>(e<T> e);

top(c()) main;
```
[tmp.txt](https://github.com/p4lang/p4c/files/3928909/tmp.txt)

mbudiu-vmw:


frontends/p4/simplifyDefUse.cpp
```diff
@@ -515,6 +515,19 @@ class FindUninitialized : public Inspector {
...
+        visit(expression->e0);
+        visit(expression->e1);
+        visit(expression->e2);
+        // This expression in fact reads the result of the operation,
+        // which is a temporary storage location, which we do not model
+        // in the def-use analysis.
+        reads(expression, LocationSet::empty);
+        registerUses(expression);
+        return false;
+    }
```
ChrisDodd: Why have this be a preorder function that manually visits all the children prunes?  Make it a postorder function and remove the explicit visits and it becomes much simpler (more comment than actual code)

mbudiu-vmw:
This was done in this way mostly for symmetry with the existing code, but I have simplified it a little bit according to your comments.

PR #2112, opened 2019-12-12
Allow extern methods to be marked @noSideEffects

ChrisDodd:
- allows transformations/optimizations to have more knowledge of
  architecture-specific externs so as to better optimize things.

### DIFF
--- a/midend/has_side_effects.h
+++ b/midend/has_side_effects.h
@@ -18,18 +18,39 @@ limitations under the License.
 #define MIDEND_HAS_SIDE_EFFECTS_H_
 
 #include "ir/ir.h"
+#include "frontends/p4/typeChecking/typeChecker.h"
+#include "frontends/common/resolveReferences/referenceMap.h"
 
-/* Should this be a method on IR::Expression? */
+/* Should this be a method on IR::Expression?  Maybe after the refMap/typeMap go away */
 
 class hasSideEffects : public Inspector {
+    P4::ReferenceMap    *refMap = nullptr;
+    P4::TypeMap         *typeMap = nullptr;
+
     bool result = false;
-    bool preorder(const IR::AssignmentStatement *) override { return !(result = true); }
-    /* FIXME -- currently assuming all calls and primitves have side effects */
-    bool preorder(const IR::MethodCallExpression *) override { return !(result = true); }
-    bool preorder(const IR::Primitive *) override { return !(result = true); }
+    bool preorder(const IR::AssignmentStatement *) override { result = true; return false; }
+    bool preorder(const IR::MethodCallExpression *mc) override {
+        if (result) return false;
+        /* assume has side effects if we can't look it up */
+        if (refMap && typeMap) {
+            auto *mi = P4::MethodInstance::resolve(mc, refMap, typeMap, true);
+            if (auto *em = mi->to<P4::ExternMethod>()) {
+                if (em->method->getAnnotation(IR::Annotation::noSideEffectsAnnotation))
+                    return true; } }
+        result = true;
+        return false; }
+    bool preorder(const IR::Primitive *) override { result = true; return false; }
     bool preorder(const IR::Expression *) override { return !result; }
+
  public:
     explicit hasSideEffects(const IR::Expression *e) { e->apply(*this); }
+    hasSideEffects(P4::ReferenceMap *rm, P4::TypeMap *tm) : refMap(rm), typeMap(tm) {}
+    hasSideEffects(P4::ReferenceMap *rm, P4::TypeMap *tm, const IR::Expression *e)
+    : refMap(rm), typeMap(tm) { e->apply(*this); }
+    bool operator()(const IR::Expression *e) {
+        result = false;
+        e->apply(*this);
+        return result; }
     explicit operator bool () { return result; }
 };
 
--- a/frontends/p4/parseAnnotations.cpp
+++ b/frontends/p4/parseAnnotations.cpp
@@ -27,6 +27,7 @@ ParseAnnotations::HandlerMap ParseAnnotations::standardHandlers() {
             PARSE_EMPTY(IR::Annotation::hiddenAnnotation),
             PARSE_EMPTY(IR::Annotation::atomicAnnotation),
             PARSE_EMPTY(IR::Annotation::optionalAnnotation),
+            PARSE_EMPTY(IR::Annotation::noSideEffectsAnnotation),
 
             // @name and @deprecated have a string literal argument.
             PARSE(IR::Annotation::nameAnnotation, StringLiteral),
--- a/ir/base.def
+++ b/ir/base.def
@@ -252,7 +252,8 @@ class Annotation {
     static const cstring optionalAnnotation;  /// Optional parameter annotation
     static const cstring pkginfoAnnotation;  /// Package documentation annotation.
     static const cstring deprecatedAnnotation;  /// Deprecation annotation.
-    static const cstring synchronousAnnotation;  /// Synchronous annotation.
+    static const cstring synchronousAnnotation;  /// noSideEffects annotation.
+    static const cstring noSideEffectsAnnotation;  /// Synchronous annotation.
     static const cstring matchAnnotation;  /// Match annotation (for value sets).
     toString{ return cstring("@") + name; }
     validate{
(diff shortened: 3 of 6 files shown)


midend/has_side_effects.h
```diff
@@ -18,18 +18,38 @@ limitations under the License.
...
-    bool preorder(const IR::MethodCallExpression *) override { return !(result = true); }
+    bool preorder(const IR::MethodCallExpression *mc) override {
+        if (result) return false;
+        /* assume has side effects if we can't look it up */
+        if (refMap && typeMap) {
+            auto *mi = P4::MethodInstance::resolve(mc, refMap, typeMap, true);
+            if (auto *em = mi->to<P4::ExternMethod>()) {
+                if (em->method->getAnnotation(IR::Annotation::noSideEffectsAnnotation))
+                    return true; } }
+        return !(result = true); }
```
liujed: Same here.

midend/has_side_effects.h
```diff
@@ -18,18 +18,38 @@ limitations under the License.
...
 
-/* Should this be a method on IR::Expression? */
+/* Should this be a method on IR::Expression?  Maybe after the refMap/typeMap go away */
 
 class hasSideEffects : public Inspector {
+    P4::ReferenceMap    *refMap = nullptr;
+    P4::TypeMap         *typeMap = nullptr;
+
     bool result = false;
     bool preorder(const IR::AssignmentStatement *) override { return !(result = true); }
```
liujed: My preference is for `result = true; return false;`. As currently written, this is easily misinterpreted as being equivalent to `return !result;`.

midend/has_side_effects.h
```diff
@@ -18,18 +18,38 @@ limitations under the License.
...
+#include "frontends/common/resolveReferences/referenceMap.h"
 
-/* Should this be a method on IR::Expression? */
+/* Should this be a method on IR::Expression?  Maybe after the refMap/typeMap go away */
 
 class hasSideEffects : public Inspector {
+    P4::ReferenceMap    *refMap = nullptr;
+    P4::TypeMap         *typeMap = nullptr;
+
     bool result = false;
```
liujed: Is `result` supposed to be a memoized result? If so, I'd change its type to be something like `boost::optional<bool>`, to avoid conflating a memoized `false` with a result that still needs to be computed.

midend/has_side_effects.h
```diff
@@ -18,18 +18,38 @@ limitations under the License.
...
     bool preorder(const IR::Primitive *) override { return !(result = true); }
     bool preorder(const IR::Expression *) override { return !result; }
+
  public:
     explicit hasSideEffects(const IR::Expression *e) { e->apply(*this); }
+    hasSideEffects(P4::ReferenceMap *rm, P4::TypeMap *tm) : refMap(rm), typeMap(tm) {}
+    hasSideEffects(P4::ReferenceMap *rm, P4::TypeMap *tm, const IR::Expression *e)
+    : refMap(rm), typeMap(tm) { e->apply(*this); }
+    bool operator()(const IR::Expression *e) {
+        result = false;
```
liujed: Why reset the value of  `result ` here? When would we expect the computed result to change?

mbudiu-vmw:
Would you like to propose this annotation to the design working group?
I will add it to the agenda for Monday.

midend/has_side_effects.h
ChrisDodd: `false` means we have not yet seen a side effect anywhere we've looked.  Once we see a side effect, we set it to true and prune all future traversals.

jafingerhut:
Out of curiosity, what precisely constitutes a side effect here?

For example, v1model.p4's mark_to_drop modified a field in a `standard_metadata` parameter passed with direction`inout`.  Would you expect it to be correct to annotate that extern function with this annotation?  `update_checksum` is similar in this regard.

What about v1model's `register` extern `read` method?  No side effects, I think, but the return value is a function of the extern's internal state, not a parameter.  Do the kinds of compiler optimizations you have in mind for this annotation expect that an extern method can do this, and thus should not be reordered with any other side-effecting calls to the same extern?

I am guessing `random` would be incorrect to annotate with this method, given the typical implementation of a pseudo-random number generator that mutates internal bits of state on each call.

`verify_checksum` as defined in v1model today _does_ have a side effect, i.e. remembering whether a bad checksum was detected or not, which is not stored in an `out` or `inout` parameter, but in internal state somewhere, so seems wrong to use this annotation for that method.

midend/has_side_effects.h
ChrisDodd: This is implicitly starting a new search for a side effect.  The idea being that this is a functor you call like: `hasSideEffects(expr)` and it returns true or false if there are any side effects in the expression.  Its conservative in that some "don't know" cases are treated as true (might have a side effect).

ChrisDodd:
Not fully clear, but I would expect those cases could not be marked @noSideEffects.  The optimizations I'm considering enabling here are dead code elimination and reordering.  So what I'm considering is that the method should not modify any state that might be seen by some other method, other than what is explicitly modified by out/inout arguments and the return value.  In addition it probably should not depend on any state other than what is in part of its arguments, though perhaps that should be a different annotation?

This comes up for functions like `extern T min<T>(in T a, in T b);` which we have in our architecture to compute the minimum of two values.  If there are multiple `min` calls, reordering them should not be a problem, and it should be possible to dead-code eliminate them if the result is dead.

mbudiu-vmw:
Here is an uncontroversial (I hope) definition for `@noSideEffects`: a function has not side effects if the compiler can safely memoize a call: given the same arguments the result is guaranteed to be the same.

For methods this is a bit more complicated, because we can consider the `this` pointer to be an argument. 

In general, this annotation is useful for the front-end; the back-end will know the semantics of each function, but the front-end can benefit as well without much knowledge about the architecture.
Note that this feature is also related to the abstract methods we are discussing.

jafingerhut:
@mbudiu-vmw I am not saying your proposed definition is controversial -- but I do not think I understand what it means without some examples.

Would your definition of `@noSideEffects` be correctly applicable to method `inc_arg` below?
```
inc_arg (inout bit<8> x) {
    x = x + 1;
}
```
All of the changes `inc_arg` makes can be completely determined from its input parameter values, but it isn't clear to me whether one can memoize it.

Would your definition be applicable to v1model's `read` method on the `register` extern?  I don't see how those could be memoized, but I may be lacking imagination there.

mbudiu-vmw:
Maybe we need two annotations: `@pure`, which means that the function is solely a result of it's `in` and `inout` arguments, and `@noSideEffects`, which means that the function does not mutate any other state except it's `out` and `inout` arguments. So `read` has no side effects, but it is not pure.

mbudiu-vmw:
`inc_arg` would be pure if we consider the result to be the returned result and all its `out` and `inout` arguments.

jafingerhut:
I mainly want to point out that without additional explanation, while I believe I know what it means to say that a function with only `in` parameters and return value is memoizable, the meaning of "it is memoizable" isn't clear to me at all for methods with direction `out` and `inout` parameters.  So if you want to define something by saying it is legal to do when a method is memoizable, please explain what that means for methods with `out` and `inout` parameters, otherwise I am just left with questions :-)

ChrisDodd:
For memoization, the out and inout parameters are the result of the memoization.  So for a @noSideEffects function, you memoize based on the values of the in and inout params and save the results for the out and inout params.  If you have a later call with the same in and inout param values, you can substitute the saved out and inout results directly without actually recomputing the function.

Note the for compile-time 'memoization' (combining redundant calls -- CSE or PRE), you can only do it for calls with provably identical input values (constants or vars that are not modifed between call sites), which precludes any function with inout params.

ChrisDodd:
I guess its not entirely clear what "depends solely on its input (in and input) arguments" means in the case of an argument that is an extern object.  Depends only on the identity of the extern, or depends on the extern's state?  If the latter, `read` could be considered @noSideEffects.

I you want to have both @noSideEffects and @pure, need to be clear what all four possible combos mean (have both, or either and not both, or neither).

jafingerhut:
During 2019-Dec-16 P4 language design work group meeting, I volunteered to create a collection of examples, and which annotations we think are applicable or not applicable for each one of those examples, and some words explaining why.  I will not attempt any formal definition of when the proposed annotations are applicable, as I don't know how.

I will explicitly try to address the proposed `@noSideEffects` and `@pure` annotations discussed earlier here, and see if there are any examples that would distinguish those as different, or whether they always seem to be the same.

I also plan to add a new proposed annotation that would restrict a method to only modify internal state of the one extern object upon which the method is invoked.  This limits the power of an extern method below what the P4_16 language spec currently allows, which is that an extern method call can modify global state of the architecture, and/or communicate and cause side effects in other extern instances.  For example, the current v1model register write extern method is guaranteed never to modify any state outside of the one register instance upon which the method is called.

jafingerhut:
Created an issue to track this here: https://github.com/p4lang/p4-spec/issues/800

PR #2116, opened 2019-12-19
Use libgc for all malloc calls, so they are scanned properly.

ChrisDodd:
- glibstdc++ uses malloc to allocate exceptions being thrown, which
  could result in them not being scanned by the garbage collector,
  leading to memory corruption.

### DIFF
--- a/lib/gc.cpp
+++ b/lib/gc.cpp
@@ -19,6 +19,7 @@ limitations under the License.
 #include <gc/gc_cpp.h>
 #include <gc/gc_mark.h>
 #endif  /* HAVE_LIBGC */
+#include <unistd.h>
 #include <new>
 #include "log.h"
 #include "gc.h"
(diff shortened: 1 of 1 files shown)


lib/gc.cpp
```diff
@@ -33,26 +34,57 @@ limitations under the License.
...
      * memory, we need to force initialization */
     if (!done_init) {
+        started_init = true;
         GC_INIT();
         done_init = true; }
     return ::operator new(size, UseGC, 0, 0);
 }
 void *operator new[](std::size_t size) {
     if (!done_init) {
+        started_init = true;
```
cc10512: why not use `std::call_once`?

lib/gc.cpp
```diff
@@ -33,26 +34,57 @@ limitations under the License.
...
         GC_INIT();
         done_init = true; }
     return ::operator new(size, UseGC, 0, 0);
 }
 void operator delete(void *p) _GLIBCXX_USE_NOEXCEPT { return gc::operator delete(p); }
 void operator delete[](void *p) _GLIBCXX_USE_NOEXCEPT { return gc::operator delete(p); }
 
+void *realloc(void *ptr, size_t size) {
```
cc10512: Why aren't you using the glibc hooks? https://www.gnu.org/software/libc/manual/html_node/Hooks-for-Malloc.html

lib/gc.cpp
ChrisDodd: `call_once` doesn't work in the recursive call case -- we need to detect if we've started to do GC_INIT but it has not yet finished.

lib/gc.cpp
ChrisDodd: Do they work on OSX libc?

Issue #2007, opened 2019-07-20
bmv2 "array index must be a constant" error


hesingh:
I am using a header stack in v1model p4-16 code.  The index into the header stack array is obtained from a header in the packet and index copied to metadata.  Obviously, the header is parsed in the parser.  p4c-bm2-ss fails because the index is not a compile-time constant.  So, how is such code supposed to work?  The p4test backend does not care and compiles the code fine.

```
./testdata/p4_16_samples/ipv6-switch-ml-bmv2.p4(111): [--Werror=invalid] error: meta._ml_idx4: Invalid all array indexes must be constant
             hdr.vector[0].e = pool[meta.ml_idx].val;
                                    ^^^^
```

jafingerhut:
Section 8.15 "Operations on header stacks" of the P4_16 language spec: https://p4.org/p4-spec/docs/P4-16-v1.1.0-spec.html#sec-expr-hs

says this: "Some architectures may impose the constraint that the index expression evaluates to a compile-time known value"

v1model is such an architecture that imposes this constraint.

If you want to access one of a handful of different header stack elements in actions depending upon a run-time varying value, one workaround for the v1model restriction is to include the run-time variable index as a table search key, which selects between several different actions that are identical, except for the constant array index values they use.

hesingh:
Right - I do understand why bmv2 imposes the restriction because there is no JSON one can generate for a non-compile-time index.  Sorry, I forgot about using tables - make sense, thanks!

hesingh:
It's not a handful of elements.  I have 128 and each can be used a index for the header stack array.

hesingh:
Both the issues I filed today are related to switch Machine Learning (ML) where array elements are added and array indices are used for aggregating ML data into different aggregators.  I think the switch ML on Tofino may be using p4-14.  I am writing p4-16 and v1model for the IETF hackathon this weekend.

jafingerhut:
The approach of multiple actions works for an arbitrary finite number of elements.  I only mentioned a handful since I was guessing that might be the case, but apparently it is not for what you want.  Again, auto-generation/printing of repetitive code might be your friend here.

I agree that it would be nice if all P4 targets supported run-time variable indexes of arrays.  The fact is that the statement is in the language spec precisely because someone wanted to caution people about expecting that, at least at the time it was written, and I doubt that the highest speed implementations have changed significantly since that writing.

hesingh:
[...]

PR #2118, opened 2019-12-22
Change p4c bmv2 backend to support runtime array index

hesingh:
See https://github.com/p4lang/p4c/issues/2007

### DIFF
--- a/testdata/p4_16_samples/ipv6-switch-ml-bmv2.p4
+++ b/testdata/p4_16_samples/ipv6-switch-ml-bmv2.p4
@@ -96,7 +96,9 @@ control ingress(inout headers hdr, inout metadata_t meta,
         if (hdr.ipv6.isValid()) {
 	    ipv6_tbl.apply();
         }
-        hdr.vector[0].e = pool[hdr.ml.idx].val;
+//        hdr.vector[0].e = pool[hdr.ml.idx].val;
+        hdr.vector[0].e = pool[1].val + 2;
+//          hdr.ipv6.version = hdr.ipv6.version + 2;
     }
 }
 
--- a/testdata/p4_16_samples/ipv6-switch-ml-bmv2.p4
+++ b/testdata/p4_16_samples/ipv6-switch-ml-bmv2.p4
@@ -49,14 +49,29 @@ parser MyParser(packet_in packet, out headers hdr, inout metadata_t meta,
     }
     state parse_udp {
         packet.extract(hdr.udp);
-        transition accept;	
+        transition parse_ml;
     }
+    state parse_ml {
+        packet.extract(hdr.ml);
+	transition parse_ml_vector;
+    }
+    state parse_ml_vector {
+        packet.extract(hdr.vector.next);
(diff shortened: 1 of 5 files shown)


hesingh:
> You need to add at least one test case, probably with an stf file to make sure that the code executes correctly.

Sure.  Right now, I am using https://github.com/IETF-Hackathon/p4-ipv6-switch-ml/blob/master/ipv6-switch-ml-bmv2.p4 as one test case.  This test also has a STF file.  I will also test simple_switch to see if the behavioral-model accepts the generated JSON for DEREFERENCE_STACK op.  I sent out the PR so that folks such as Antonin could take the diffs and see if the generated JSON looks ok.

hesingh:
Here is JSON output due to code changes of this PR:

```
{
          "op" : "assign",
          "parameters" : [
            {
              "type" : "field",
              "value" : ["vector[0]", "e"]
            },
            {
              "type" : "field",
              "value" : [
                {
                  "type" : "header",
                  "value" : {
                    "value" : {
                      "op" : "DEREFERENCE_STACK",
                      "left" : {
                        "type" : "stack",
                        "value" : "pool_0"
                      },
                      "right" : {
                        "type" : "field",
                        "value" : ["hdr", "ml", "idx"]
                      }
                    },
                    "type" : "expression"
                  }
                }
              ]
            }
          ],
          "source_info" : {
            "filename" : "/home/hemant/p4-ipv6-switch-ml/ipv6-switch-ml-bmv2.p4",
            "line" : 109,
            "column" : 13,
            "source_fragment" : "hdr.vector[0].e = pool[hdr.ml.idx].val"
          }
}
```

hesingh:
The P4 test code used for this PR does not work yet with latest behavioral-model and its simple_switch.  So far, I have been able to see that at least `behavioral-model/src/bm_sim/P4Objects.cpp` needs a change if the generated JSON is acceptable.  Let's first review the JSON output and then one can look into any changes to the behavioral-model code.

hesingh:
The full JSON file output from code changes of this PR has been added to a new issue filed against the behavioral-model.  Please see 

https://github.com/p4lang/behavioral-model/issues/835

mbudiu-vmw:
these tests are failing.

testdata/p4_16_samples/ipv6-switch-ml-bmv2.p4
```diff
@@ -49,14 +49,29 @@ parser MyParser(packet_in packet, out headers hdr, inout metadata_t meta,
     }
     state parse_udp {
         packet.extract(hdr.udp);
-        transition accept;	
+        transition parse_ml;
```
mbudiu-vmw: in general we do not modify tests, but we add new ones.

hesingh:
> these tests are failing.

This PR code is not complete yet because I got diverted to other priorities.  When the code is complete and tests pass, I will add a note to this PR.

hesingh:
Another PR is handling the code changes of this PR. 
https://github.com/p4lang/p4c/pull/2139

PR #2119, opened 2019-12-24
Allow primitive calls in bbox expressions

ChrisDodd:


ChrisDodd:
I"ve thought about adding test cases, but the only thing that should be affected by this is P4_14 blackbox definitions (a v1.1 extension from the later withdrawn v1.1 spec) that pretty much only come up if you have an old v1.1-based code your trying to port from the old python p4_14 hlir to the new compiler.  There are no P4_14 blackboxes in the v1model/simple_switch stuff we have.

I though there migh be something with the factory1/2 tests that broke under the first version of this change to at least cover the typeChecking/bindVariables fix, but I can't figure out how those were passing before without this fix...

PR #2121, opened 2019-12-30
Add pure/localState/packetStata annotations

ChrisDodd:
- add annotations in appropriate places in v1mode.p4 and psa.p4
- fix resolveReferences to look at parameters


p4include/v1model.p4
```diff
@@ -340,6 +348,7 @@ enum HashAlgorithm {
 }
 
 @deprecated("Please use mark_to_drop(standard_metadata) instead.")
+@packetState
```
jafingerhut: To be nit-picky here, the reason this is deprecated in v1model is because it isn't a legal P4_16 extern function at all.  It violates the one restriction that externs have -- it accesses P4_16 program variables that are not parameters.

p4include/psa.p4
```diff
@@ -517,12 +526,15 @@ extern Checksum<W> {
   /// time the object is instantiated, that is, whenever the parser or control
   /// containing the Checksum object are applied.
   /// All state maintained by the Checksum object is independent per packet.
+  @packetState
```
mbudiu-vmw: It looks to me like we may want this annotation to be on the extern itself rather than on each method.
Does it ever make sense to have only some methods of an extern be `@packetState`?

jafingerhut:
In the interests of moving things forward here, if `@pure` and `@noSideEffects` are the only two annotations that we really have good use cases for taking advantage of in compiler optimization passes, would it be better to only add those two, and leave out all of the others?

I apologize if I introduced confusion by describing the other kinds of annotations -- just my over-exuberance in looking for other broader kinds of categories that extern functions/methods fit into, regardless of whether they enabled compiler optimizations.

jafingerhut:
The parts of these changes that add the `@pure` annotation, and changes to v1model.p4 and psa.p4 architectures to add `@pure` and `@noSideEffects` annotations, have been merged with this commit: https://github.com/p4lang/p4c/commit/301411d7ee4cc22d381a923037f840f575635aa2

As of this writing, there is currently no PR that makes the rest of these changes.  I do not know if such a thing is of interest, either, until and unless there is a known compiler optimization use case for additional similar annotations.

jafingerhut:
@ChrisDodd @mihaibudiu Does it seem reasonable to close this, since `@pure` and `@noSideEffects` have been in the P4 spec and p4c implementation for a couple of years now, and no one seems interested in pursuing other annotations like `packetState` since then?

mihaibudiu:
I have no problem closing this, but it is Chris' PR.

Issue #2123, opened 2020-01-01
p4c bmv2 backend does not support infix operator


hesingh:
These were the code changes made to a P4 program to reproduce the issue.

```
diff --git a/testdata/p4_16_samples/basic_routing-bmv2.p4 b/testdata/p4_16_samples/basic_routing-bmv2.p4
index b3efcb7c..408dae90 100644
--- a/testdata/p4_16_samples/basic_routing-bmv2.p4
+++ b/testdata/p4_16_samples/basic_routing-bmv2.p4
@@ -40,8 +40,8 @@ struct headers {
 parser ParserImpl(packet_in packet, out headers hdr, inout metadata meta, inout standard_metadata_t standard_metadata) {
     state parse_ethernet {
         packet.extract(hdr = hdr.ethernet);
-        transition select(hdr.ethernet.etherType) {
-            16w0x800: parse_ipv4;
+        transition select(true, hdr.ethernet.etherType) {
+            (true, 16w0x800 .. 16w0x806): parse_ipv4;
             default: accept;
         }
     }
```

Then the changed program was tested.  See error below.

```
hemant@ubuntu:/tmp/p4c/build$ ./p4c-bm2-ss --std p4-16 ../testdata/p4_16_samples/basic_routing-bmv2.p4
../testdata/p4_16_samples/basic_routing-bmv2.p4(44): [--Werror=invalid] error: ..: Invalid must evaluate to a compile-time constant
            (true, 16w0x800 .. 16w0x806): parse_ipv4;
                   ^^^^^^^^^^^^^^^^^^^^
hemant@ubuntu:/tmp/p4c/build$ 
```

hesingh:
Closing because there is already an open issue: https://github.com/p4lang/p4c/issues/357

PR #2124, opened 2020-01-01
Fix p4c bmv2 parser to support Range operator

hesingh:
See https://github.com/p4lang/p4c/issues/2123


hesingh:
CLosing due to other diffs included in the branch besides range fix.  I will pull another PR later.

PR #2125, opened 2020-01-01
Fix supporting Range operator for bmv2 backend

hesingh:
See https://github.com/p4lang/p4c/issues/2123


jafingerhut:
If this is added to p4c, it seems also reasonable to have a compile time check that the left operand of the range is less than or equal to the right operand, and issue an error if this is not true, with a test program for that case.

jafingerhut:
Also perhaps checks that the left and right operands fit within the size of the field.  It isn't clear to me whether it is better to have an error if the values are outside of the size of the field, or a warning that they are being truncated -- probably an error?

hesingh:
> Also perhaps checks that the left and right operands fit within the size of the field. It isn't clear to me whether it is better to have an error if the values are outside of the size of the field, or a warning that they are being truncated -- probably an error?

p4c already includes such a check.  I changed the end of the range to 65536 for 16-bit integer and p4c catches the issue.

```
hemant@ubuntu:~/p4c/build$ ./p4c-bm2-ss --std p4-16 issue-2123.p4 -o tmp.json
issue-2123.p4(44): [--Wwarn=mismatch] warning: 0x10000: value does not fit in 16 bits
            (true, 16w0x800 .. 16w0x10000
```                               ^^^^^^^^^^

**Comment by hesingh** [comment]
@mbudiu-vmw @ChrisDodd and anyone else who has worked with p4c including its bmv2 backed, please review this PR - thanks.

File: backends/bmv2/common/parser.cpp [hunk 1]
```diff
@@ -287,16 +287,23 @@ void ParserConverter::convertSimpleKey(const IR::Expression* keySet,
...
+        auto left = range->left->to<IR::Constant>()->value;
         if (!range->right->is<IR::Constant>()) {
-            ::error(ErrorType::ERR_INVALID, "must evaluate to a compile-time "
-            "constant", range->right);
+            ::error(ErrorType::ERR_INVALID, "must evaluate to a compile-time"
+                    "constant", range->right);
+            return;
+        }
+        auto right = range->right->to<IR::Constant>()->value;
+        if (right <= left) {
```
jafingerhut: I think this should be `right < left`.  While it might be odd for someone to make the left and right equal to each other, it doesn't seem like it ought to be an error.

hesingh:
I have also generated relevant JSON with these code changes:

```
 {
              "type" : "hexstr",
              "value" : "0x010800",
              "mask" : "0x010806",
              "next_state" : "parse_ipv4"
            },
```

jafingerhut:
If the BMv2 JSON file uses keys "value" and "mask" to store the min and max value of a range, it seems that simple_switch will behave the same as if you wrote "value &&& mask" in the P4 source code, which is not the same as a range at all.

There needs to be a different way to encode a range in the BMv2 JSON file, doesn't there?

jafingerhut:
@antoninbas As far as I can tell from the JSON_format.md documentation and the src/bm_sim/P4Objects.cpp source code, there is currently no way in a BMv2 JSON file to represent a parser transition with a range [min, max] of values for a field.  Does that sound correct?

If it is correct, then changes to the p4c bmv2 back end are not enough to get this feature to work.  It also requires changes in behavioral-model code.

hesingh:
> If the BMv2 JSON file uses keys "value" and "mask" to store the min and max value of a range, it seems that simple_switch will behave the same as if you wrote "value &&& mask" in the P4 source code, which is not the same as a range at all.
> 
> There needs to be a different way to encode a range in the BMv2 JSON file, doesn't there?

Yes.  

I didn't know how to encode range end yet.  

I propose we use a new `parser_range` transition type since `range` is already used for a table key match_type.  Then, this is what the JSON should look like.  `value` represents start of range. 

```
            {
              "type" : "parse_range",
              "value" : "0x010800",
              "end" : "0x010806",
              "next_state" : "parse_ipv4"
            },
```

Since we have new JSON, the behavioral-model will also change.

egecemkirci:
I tried to compile my P4 code after the new commits but I still get the same error.

egecemkirci:
I tried to compile my P4 code after the new commits but I still get the same error.

jafingerhut:
If there are multiple keys in the select expression, each of which uses a range, or some that use a range, some a ternary expression, I don't think the BMv2 JSON file is currently general enough to handle that.  It seems like it would require a separate sub-data structure in the BMv2 JSON file for each field in the select expression.

For example, consider this in the source code:
```
        transition select(hdr.ethernet.etherType, hdr.ethernet.srcAddr[7:0]) {
            (0x0800 &&& 0x0fff, 0x08 &&& 0x08): parse_ipv4;
            default: accept;
        }
```
Right now simple_switch expects in the BMv2 JSON file that all of those value/masks are concatenated into one long bit string.  If you did that for some that had ranges, some ternary, then you need some way to specify to simple_switch that some of them are range, some are ternary.

jafingerhut:
Or, alternately, one could imagine implementing this in a different way, that does not require any changes to the BMv2 JSON file format at all.

That could be done by the technique of taking every range, and creating one or more value/mask expressions out of it, that together match all of the same values that the range does, and nothing outside of the range.  As a short example, with an 8-bit field and the range 16 .. 25, the two value/masks 0x10 &&& 0xf8 (matching values in the range 16 through 23, inclusive) and 0x18 &&& 0xfe (matching values in the range 24 through 25, inclusive), match the same range.  In that case you would create two transitions in the BMv2 JSON file, not one.  In general, there could be dozens or hundreds of value/mask transitions in the BMv2 JSON file for a single line in the P4_16 source code, depending on how many fields there were, and what the ranges were.

This is exactly the kind of "range expansion" done in some traditional switches that support range matching for ACLs, and have only a TCAM for doing the matching.

hesingh:
> If there are multiple keys in the select expression, each of which uses a range, or some that use a range, some a ternary expression, I don't think the BMv2 JSON file is currently general enough to handle that. It seems like it would require a separate sub-data structure in the BMv2 JSON file for each field in the select expression.
> 
> For example, consider this in the source code:
> 
> ```
>         transition select(hdr.ethernet.etherType, hdr.ethernet.srcAddr[7:0]) {
>             (0x0800 &&& 0x0fff, 0x08 &&& 0x08): parse_ipv4;
>             default: accept;
>         }
> ```
> 
> Right now simple_switch expects in the BMv2 JSON file that all of those value/masks are concatenated into one long bit string. If you did that for some that had ranges, some ternary, then you need some way to specify to simple_switch that some of them are range, some are ternary.

With the code changes of this PR, there is a new `parse_range` to disambiguate range with infix (&&&) operator.  With such a design, a separate data structure and data structure default exists in JSON file for each line with a key. What is the concern with such a design besides the fact that behavoiral-model has to change?

jafingerhut:
How would you represent this P4 code using your new parse_range keyword?  I don't see how.
```
    transition select(hdr.ethernet.etherType, hdr.ethernet.srcAddr[7:0]) {
            (0x0800 .. 0x0806, 0x08 .. 0x11): parse_ipv4;
            default: accept;
        }
```

hesingh:
> How would you represent this P4 code using your new parse_range keyword? I don't see how.
> 
> ```
>     transition select(hdr.ethernet.etherType, hdr.ethernet.srcAddr[7:0]) {
>             (0x0800 .. 0x0806, 0x08 .. 0x11): parse_ipv4;
>             default: accept;
>         }
> ```

Break up into two lines:

```
(0x0800 .. 0x0806): parse_ipv4;
(0x08 .. 0x11): parse_ipv4;
```

jafingerhut:
The original code implies an AND condition, i.e. it means "if hdr.ethernet.etherType is in the range [0x0800, 0x0806], AND hdr.ethernet.srcAddr[7:0] is in the range [0x08, 0x11], then transition to state parse_ipv4".

Your code split over two lines is not syntactically correct, because it does not have match criteria for both fields.

Even if you added them, somehow, multiple lines in a transition select are "if line1 condition, then transition to this state, ELSE IF line2 condition then transition to this other state".  Not equivalent.

hesingh:
> The original code implies an AND condition, i.e. it means "if hdr.ethernet.etherType is in the range [0x0800, 0x0806], AND hdr.ethernet.srcAddr[7:0] is in the range [0x08, 0x11], then transition to state parse_ipv4". Just like you had a table with two fields in its key.
> 
> Your code split over two lines is not syntactically correct, because it does not have match criteria for both fields.
> 
> Even if you added them, somehow, multiple lines in a transition select are "if line1 condition, then transition to this state, ELSE IF line2 condition then transition to this other state". Not equivalent.

Sorry, I missed two conditions in the select statement.  Just take the cartesian product of the two sets of ranges.  For example, few product elements are shown below.
```
(0x0800, 0x08) : parse_ipv4;
(0x0800, 0x09) : parse_ipv4
etc.
```

Each pair above has the values concatenated by existing code.

jafingerhut:
You could do the Cartesian product of all exact match values, but that could blow up exponentially in size, and probably very soon hit limits on memory sizes or data structure sizes in simple_switch.

You can also break it into the fewest number of ternary ranges, as I mention above, which has the guarantee that a range on any W-bit wide field requires at most 2W-2 ternary value/masks to represent the range exactly, as opposed to the much much larger 2^W exact values.

hesingh:
In the worst case, the cartesion product blows up exponentially.  In our example, one range is small with `0x08 ... 0x11` .

Changing parser range to infix operation with value/mask you suggested makes sense.  I will go with this.  Thanks.

hesingh:
Now, I know why this feature was sitting around as unimplemented until now.  

The first complexity is writing a function to change range to multiple infix (`&&&`) operations.
The second involves removing the range Node from IR and adding few more `Mask` nodes to IR.  Currently the code in p4c/backends/bmv2/common/parser.h uses an Inspector class.  We need a Transform class to change IR.  Or perhaps write a backend or midend pass to morph range to multiple Mask operations.

jafingerhut:
Others can provide more informed answers than I can on this subject, but if the technique to be implemented expands ranges into one or more equivalent ternary entries, then it seems reasonable to make such a transformation specific to the bmv2 back end.  I would guess that it is preferable to leave the decision to do such expansion to the target-specific part of the p4c back end, and not force this transformation on back ends that do not want/need it.

hesingh:
Right, my goal is to develop the feature only for the bmv2 backend.  

I will start on the range conversion to ternary entries next.  Thereafter, I will see what to do with the compiler IR to emit multiple ternary entries for a range and also be able concatenate the entries.

antoninbas:
Sorry I'm late to the party here. @jafingerhut is correct, this requires changes to the bmv2 JSON and the bmv2 C++ code if you want to support range matches "natively" in bmv2. This is also not as simple as adding a "match type" field because at the moment all the different fields in the select key are concatenated (so each select case is mapped to a single value, mask pair). See https://github.com/p4lang/behavioral-model/blob/master/docs/JSON_format.md#parsers for more details and an example.

The right thing to do do IMO would be to support this natively in bmv2, in order to keep bmv2 log messages consistent with the P4 program. However, translating the range matches into a minimum set of ternary matches is representative of what an HW target using TCAMs for the parser would do, so it's also a good exercise to implement this as a **midend** pass (that could be used by other p4c backends). Is there a well-known algorithm or some literature that describes a solution to perform this transformation?

hesingh:
I haven't looked hard, but I don't see any algorithm in literature to perform the transformation.  This is why I am developing an algorithm and will publish it when it's ready.

Things to consider in the algorithm are:

(a) The value of Mask (value, mask) is a function of the min of the Range. E.g. min is 0x0800, and the value could be 0x080F.
(b) Is the size of the Range a power of 2 (e.g. `0x0800 .. 0x808`)  If yes, then the above value in (a) is used with a mask of 0xFFF0.
(c) If the size is not power of 2, but even or odd, then multiple Masks are needed - this is the (value, mask ) creation I am working out to complete the algorithm.

Sure, writing a midend pass for the transformation is fine so that other backends can use the feature.

thanks.

mbudiu-vmw:
We've had this issue open for a long time: https://github.com/p4lang/p4c/issues/357 but I never found the time to work on it.

jafingerhut:
"Is there a well-known algorithm or some literature that describes a solution to perform this transformation?"  There is an algorithm that is well known at least among people who study and/or implement ACL-like packet classification in TCAMs.  I have seen the "at most 2W-2 TCAM entries" to do the expansion of arbitrary ranges on W-bit wide fields mentioned in at least a couple of research papers from around the year 2000, and am pretty sure the expansion is described in some published research paper, but haven't found which one (references welcome if you know).

I can hack up a simple Python implementation and publish that as a sample, probably in a day or three.  It is pretty simple.

jafingerhut:
OK, here is a README about, and a Python implementation of code, that produces a small set of prefix value/masks to implement a range match: https://github.com/jafingerhut/p4-guide/tree/master/match-range-using-tcam

Comments/question/corrections welcome.  I am happy to add a C/C++ implementation if someone else wants to contribute one.

hesingh:
Thanks @jafingerhut.  I am looking at the code.

mbudiu-vmw:
You should an stf file to check that indeed the ranges work.

hesingh:
Sure, I will add STF tests.  

Yesterday, I abandoned porting Andy's Python code to C++ because its cumbersome work and also because a another algorithm with 20 lines of C code produces the same results as Andy's.  The C code also supports integer of any size used with Range.  I have tested the code with 64-bit integer.  

This is the C code checked in few minutes back to this PR.   

The C code's algorithm tracks trailing zeroes and creates Mask(s) which is a (value, mask) pair.

hesingh:
I added a stand-alone app to run the C code to convert ranges to Masks (value, mask) being used in this PR.

https://github.com/hesingh/range-to-tcam-val-mask

jafingerhut:
@hesingh I have created a slightly modified version of your C++ code that compiled with GCC, and includes the ability to do the exhaustive test for all ranges of a given width, doing its own internal sanity checks to verify that the results are correct.

If I keep your copyright notice, any objection if I add that version to my p4-guide repository in the same directory as I link in my earlier comment, next to the Python version?

jafingerhut:
Oh, and I ran the self-checking and try-all-possible-ranges-for-a-given-width version for all ranges with bit widths from 1 up to 12, inclusive, and it found no mistakes, so your version of the code seems to be working to me, for a single range.

For a full solution in p4c, it would also need to do cross products of these range expansions when there are multiple fields doing range matching in the same select expression.

hesingh:
Andy, sure thing, you can take the C code - thanks!  I do have a Apache 2.0 license in the C code. We are a team together!

Yes, I do know about the cross product because I mentioned it very early on in this PR about cartesian product.  Right now, I am working on the midend pass.

hesingh:
Also, since I saw recently that p4c changed using mpz to big_int, I have used big_int in the C code for range to tcam masks.  Thus infinite integer of p4c is supported by the C code.

jafingerhut:
No promises, but I can probably cook up an STF test for this -- I know it will not compile or work correctly with current p4c, yet.  I can make one that exercises both a single field with a range in a select expression, and also two or even three fields with ranges in a single select expression, to exercise the cross-producting code.

hesingh:
Thanks @jafingerhut 

From my side, I have a P4 program that tests for single and multiple ranges in a keyset,

At this time tonight, I have been able to call computation of tcam entries from range in the midend pass.  Next, I will add code to compute cross or cartesian product.  Thereafter, I am ready to change the IR tree to replace `IR::Range` with one or more `IR::Mask`.

When the above code blocks are complete, I will check in code to this PR's repo.

hesingh:
In the printf below, a vector is represented between parenthesis.  A vector of vectors is between open and close curly braces.

This is the P4 code that generates the p4c midend pass log below.  Three ranges are used. 

The cartesian product uses c++ return "auto" and lambda parameter declaration which is supported only with c++14 or gnu++14.  So, I changed the p4c/CMakeList from "11" to "14".  

```
transition select(hdr.ethernet.etherType, hdr.ethernet.srcAddr[7:0],  hdr.ethernet.dstAddr[7:0]) {
            (0x0800 .. 0x0806, 0x08 .. 0x11, 0x08 .. 0x10): parse_ipv4;
            default: accept;
        }
```

```
hemant@ubuntu:~/p4c-range/build$ ./p4c-bm2-ss --std p4-16 ../testdata/p4_16_samples/issue-2123.p4
Found multiple Ranges, of width: 16 min 0x800 max 0x806
(0x800/0xfffc 0x804/0xfffe 0x806/0xffff ) 
Found multiple Ranges, of width: 0x8 min 0x8 max 0x11
(0x8/0xf8 0x10/0xfe ) 
Found multiple Ranges, of width: 0x8 min 0x8 max 0x10
(0x8/0xf8 0x10/0xff ) 
Vector of vectors is: 
{ (0x800/0xfffc 0x804/0xfffe 0x806/0xffff ) (0x8/0xf8 0x10/0xfe ) (0x8/0xf8 0x10/0xff ) }
Cartesian product is: 
{ (0x800/0xfffc 0x8/0xf8 0x8/0xf8 ) (0x800/0xfffc 0x8/0xf8 0x10/0xff ) (0x800/0xfffc 0x10/0xfe 0x8/0xf8 ) (0x800/0xfffc 0x10/0xfe 0x10/0xff ) (0x804/0xfffe 0x8/0xf8 0x8/0xf8 ) (0x804/0xfffe 0x8/0xf8 0x10/0xff ) (0x804/0xfffe 0x10/0xfe 0x8/0xf8 ) (0x804/0xfffe 0x10/0xfe 0x10/0xff ) (0x806/0xffff 0x8/0xf8 0x8/0xf8 ) (0x806/0xffff 0x8/0xf8 0x10/0xff ) (0x806/0xffff 0x10/0xfe 0x8/0xf8 ) (0x806/0xffff 0x10/0xfe 0x10/0xff ) }
```

mbudiu-vmw:
We cannot require another compiler version, some people have constraints in their production environments on what tools they can deploy.

hesingh:
This is why I gave a heads up.  Thanks for the feedback.  I will change the cartesian product code and make it work with c++11.

hesingh:
The midend passed is complete and checked in for code review.
The bmv2 backend changes have been backed out.

Here is a test log.

```
hemant@ubuntu:~/p4c/build$ ./p4c-bm2-ss --std p4-16 --top4 ReplaceSelectRange --dump ./test ../testdata/p4_16_samples/issue-2123.p4 -o tmp.json
hemant@ubuntu:~/p4c/build$ 
hemant@ubuntu:~/p4c/build$ less test/issue-2123-BMV2\:\:SimpleSwitchMidEnd_25_ReplaceSelectRange.p4
    state parse_ipv4 {
        packet.extract<ipv4_t>(hdr = hdr.ipv4);
        transition select(hdr.ipv4.ihl, hdr.ipv4.protocol) {
            (4w0x5, 8w0x1): parse_icmp;
            (4w0x5, 8w0x6): parse_tcp;
            (4w0x5, 8w0x11): parse_udp;
            (default, default): accept;
        }
    }
    state parse_icmp {
        transition accept;
    }
    state parse_tcp {
        transition accept;
    }
    state parse_udp {
        transition accept;
    }
    state start {
        packet.extract<ethernet_t>(hdr = hdr.ethernet);
        transition select(hdr.ethernet.etherType, hdr.ethernet.srcAddr[7:0], hdr.ethernet.dstAddr[7:0]) {
            (16w0x800 &&& 16w0xfffc, 8w0x8 &&& 8w0xf8, 8w0x8 &&& 8w0xf8): parse_ipv4;
            (16w0x800 &&& 16w0xfffc, 8w0x8 &&& 8w0xf8, 8w0x10 &&& 8w0xff): parse_ipv4;
            (16w0x800 &&& 16w0xfffc, 8w0x10 &&& 8w0xfe, 8w0x8 &&& 8w0xf8): parse_ipv4;
            (16w0x800 &&& 16w0xfffc, 8w0x10 &&& 8w0xfe, 8w0x10 &&& 8w0xff): parse_ipv4;
            (16w0x804 &&& 16w0xfffe, 8w0x8 &&& 8w0xf8, 8w0x8 &&& 8w0xf8): parse_ipv4;
            (16w0x804 &&& 16w0xfffe, 8w0x8 &&& 8w0xf8, 8w0x10 &&& 8w0xff): parse_ipv4;
            (16w0x804 &&& 16w0xfffe, 8w0x10 &&& 8w0xfe, 8w0x8 &&& 8w0xf8): parse_ipv4;
            (16w0x804 &&& 16w0xfffe, 8w0x10 &&& 8w0xfe, 8w0x10 &&& 8w0xff): parse_ipv4;
            (16w0x806 &&& 16w0xffff, 8w0x8 &&& 8w0xf8, 8w0x8 &&& 8w0xf8): parse_ipv4;
            (16w0x806 &&& 16w0xffff, 8w0x8 &&& 8w0xf8, 8w0x10 &&& 8w0xff): parse_ipv4;
            (16w0x806 &&& 16w0xffff, 8w0x10 &&& 8w0xfe, 8w0x8 &&& 8w0xf8): parse_ipv4;
            (16w0x806 &&& 16w0xffff, 8w0x10 &&& 8w0xfe, 8w0x10 &&& 8w0xff): parse_ipv4;
            (16w0x800, 8w0x8 &&& 8w0xf8, 8w0x6 &&& 8w0x11): parse_ipv4;
            (16w0x800, 8w0x10 &&& 8w0xff, 8w0x6 &&& 8w0x11): parse_ipv4;
            default: accept;
        }
```

hesingh:
@jafingerhut You can use the code changes in the PR to generate STF tests - thanks.  

The repo is:

https://github.com/hesingh/p4c

hesingh:
@jafingerhut In moving the code from backend to midend, I forgot to add the sanity check for range.  I just added the check to the midend pass and tested the check failure with a new test P4 program.  Code is checked in.

midend/replaceSelectRange.cpp
```diff
@@ -0,0 +1,177 @@
...
+
+    return (zeros < width) ? zeros : width;
+}
+
+bool DoReplaceSelectRange::checkRange(const IR::Range* range) {
+    if (range == nullptr)
+        return false;
+
+    if (!range->left->is<IR::Constant>()) {
+        std::cout << "must evaluate to a compile-time "
```
mbudiu-vmw: use ::error to report errors.

midend/replaceSelectRange.cpp
```diff
@@ -0,0 +1,177 @@
...
+        << " is less than start " << left << std::endl;
+        return false;
+    }
+    return true;
+}
+
+std::vector<const IR::Mask *>
+DoReplaceSelectRange::rangeToMasks(const IR::Range *r) {
+    bool st = checkRange(r);
+    BUG_CHECK(st, "Range check failed");
```
mbudiu-vmw: This cannot be a bug, people can write programs with such ranges and the compiler should not crash.

midend/replaceSelectRange.cpp
```diff
@@ -0,0 +1,177 @@
...
+
+    return masks;
+}
+
+std::vector<IR::Vector<IR::Expression>>
+DoReplaceSelectRange::cartesianAppend(std::vector<IR::Vector<IR::Expression>> vecs,
+                                      std::vector<const IR::Mask *> masks) {
+    std::vector<IR::Vector<IR::Expression>> newVecs;
+
+    if (vecs.empty()) {
```
mbudiu-vmw: is this case necessary?

midend/replaceSelectRange.h
```diff
@@ -0,0 +1,66 @@
...
+               std::vector<IR::Vector<IR::Expression>> vecs,
+               const IR::Expression *e);
+};
+
+class ReplaceSelectRange final : public PassManager {
+ public:
+        ReplaceSelectRange(ReferenceMap* refMap, TypeMap* typeMap) {
+            passes.push_back(new TypeChecking(refMap, typeMap));
+            passes.push_back(new DoReplaceSelectRange(typeMap));
+            passes.push_back(new ClearTypeMap(typeMap));
```
mbudiu-vmw: I don't think this is necessary.

midend/replaceSelectRange.cpp
hesingh: This is necessary because when the first Cartesian product is taken, the left vector is empty. Without explicitly handling this case, the resulting product would be another empty vector.

midend/replaceSelectRange.h
hesingh: You are correct because the IR::Range nodes are removed and do not change type.

hesingh:
I have no clue why Travis fails on this test because in my workspace, I test the failed case and no file gets changed.  In the Travis failure, the tests results is adding zero masks to the midend file which I cannot reproduce for an issue.  The Travis log is at the bottom of this note.

```
hemant@ubuntu:~/p4c/build$ p4/testdata/p4_16_samples/issue-2123.p4.test -f
./p4test --pp ./tmp9u4xkwfw/issue-2123.p4 --dump ./tmp9u4xkwfw --top4 FrontEndDump,FrontEndLast,MidEndLast --testJson --maxErrorCount 100 --arch v1model --p4runtime-files ./tmp9u4xkwfw/issue-2123.p4.p4info.txt --p4runtime-entries-files ./tmp9u4xkwfw/issue-2123.p4.entries.txt /home/hemant/p4c/testdata/p4_16_samples/issue-2123.p4
./p4test -I. --pp ./tmp9u4xkwfw/issue-2123.p4-x --std p4-16 ./tmp9u4xkwfw/issue-2123.p4 --maxErrorCount 100
./p4test -I. --pp ./tmp9u4xkwfw/issue-2123-midend.p4-x --std p4-16 ./tmp9u4xkwfw/issue-2123-midend.p4 --maxErrorCount 100
hemant@ubuntu:~/p4c/build$ git status
On branch master
Your branch is up to date with 'origin/master'.

nothing to commit, working tree clean
hemant@ubuntu:~/p4c/build$
```

This is the Travis failure.

```
--- /p4c/testdata/p4_16_samples_outputs/issue-2123-midend.p4	2020-01-11 01:13:50.000000000 +0000
+++ ./tmp5l1ieaur/issue-2123-midend.p4	2020-01-11 01:26:42.904645889 +0000
@@ -62,20 +62,20 @@
     state start {
         packet.extract<ethernet_t>(hdr = hdr.ethernet);
         transition select(hdr.ethernet.etherType, hdr.ethernet.srcAddr[7:0], hdr.ethernet.dstAddr[7:0]) {
-            (16w0x800 &&& 16w0xfffc, 8w0x8 &&& 8w0xf8, 8w0x8 &&& 8w0xf8): parse_ipv4;
-            (16w0x800 &&& 16w0xfffc, 8w0x8 &&& 8w0xf8, 8w0x10 &&& 8w0xff): parse_ipv4;
-            (16w0x800 &&& 16w0xfffc, 8w0x10 &&& 8w0xfe, 8w0x8 &&& 8w0xf8): parse_ipv4;
-            (16w0x800 &&& 16w0xfffc, 8w0x10 &&& 8w0xfe, 8w0x10 &&& 8w0xff): parse_ipv4;
-            (16w0x804 &&& 16w0xfffe, 8w0x8 &&& 8w0xf8, 8w0x8 &&& 8w0xf8): parse_ipv4;
-            (16w0x804 &&& 16w0xfffe, 8w0x8 &&& 8w0xf8, 8w0x10 &&& 8w0xff): parse_ipv4;
-            (16w0x804 &&& 16w0xfffe, 8w0x10 &&& 8w0xfe, 8w0x8 &&& 8w0xf8): parse_ipv4;
-            (16w0x804 &&& 16w0xfffe, 8w0x10 &&& 8w0xfe, 8w0x10 &&& 8w0xff): parse_ipv4;
-            (16w0x806 &&& 16w0xffff, 8w0x8 &&& 8w0xf8, 8w0x8 &&& 8w0xf8): parse_ipv4;
-            (16w0x806 &&& 16w0xffff, 8w0x8 &&& 8w0xf8, 8w0x10 &&& 8w0xff): parse_ipv4;
-            (16w0x806 &&& 16w0xffff, 8w0x10 &&& 8w0xfe, 8w0x8 &&& 8w0xf8): parse_ipv4;
-            (16w0x806 &&& 16w0xffff, 8w0x10 &&& 8w0xfe, 8w0x10 &&& 8w0xff): parse_ipv4;
-            (16w0x800, 8w0x8 &&& 8w0xf8, 8w0x6 &&& 8w0x11): parse_ipv4;
-            (16w0x800, 8w0x10 &&& 8w0xff, 8w0x6 &&& 8w0x11): parse_ipv4;
+            (16w0x800 &&& 16w0x0, 8w0x8 &&& 8w0x0, 8w0x8 &&& 8w0x0): parse_ipv4;
+            (16w0x800 &&& 16w0x0, 8w0x8 &&& 8w0x0, 8w0x10 &&& 8w0x0): parse_ipv4;
+            (16w0x800 &&& 16w0x0, 8w0x10 &&& 8w0x0, 8w0x8 &&& 8w0x0): parse_ipv4;
+            (16w0x800 &&& 16w0x0, 8w0x10 &&& 8w0x0, 8w0x10 &&& 8w0x0): parse_ipv4;
+            (16w0x804 &&& 16w0x0, 8w0x8 &&& 8w0x0, 8w0x8 &&& 8w0x0): parse_ipv4;
+            (16w0x804 &&& 16w0x0, 8w0x8 &&& 8w0x0, 8w0x10 &&& 8w0x0): parse_ipv4;
+            (16w0x804 &&& 16w0x0, 8w0x10 &&& 8w0x0, 8w0x8 &&& 8w0x0): parse_ipv4;
+            (16w0x804 &&& 16w0x0, 8w0x10 &&& 8w0x0, 8w0x10 &&& 8w0x0): parse_ipv4;
+            (16w0x806 &&& 16w0x0, 8w0x8 &&& 8w0x0, 8w0x8 &&& 8w0x0): parse_ipv4;
+            (16w0x806 &&& 16w0x0, 8w0x8 &&& 8w0x0, 8w0x10 &&& 8w0x0): parse_ipv4;
+            (16w0x806 &&& 16w0x0, 8w0x10 &&& 8w0x0, 8w0x8 &&& 8w0x0): parse_ipv4;
+            (16w0x806 &&& 16w0x0, 8w0x10 &&& 8w0x0, 8w0x10 &&& 8w0x0): parse_ipv4;
+            (16w0x800, 8w0x8 &&& 8w0x0, 8w0x6 &&& 8w0x11): parse_ipv4;
+            (16w0x800, 8w0x10 &&& 8w0x0, 8w0x6 &&& 8w0x11): parse_ipv4;
             default: accept;
         }
     }
```

hesingh:
The root cause to Travis failure has been found.  The `mask` is code was declared as `auto`.  Such code worked on my 64-bit laptop and Ubuntu 18.04.  However, on the github machine the `mask` was always zero.  Changing `mask`  from `auto` to `big_int` fixed Travis failure.

hesingh:
If one takes code changes from this repo (https://github.com/hesingh/p4c) because parser Range is supported, please see the following P4 program snippet which causes an interesting error.

```
	    (0x0806 .. 0x0800): parse_ipv4;
            default: accept;
        }
```

```
hemant@ubuntu:~/p4c/build$ ./p4c-bm2-ss --std p4-16 ../testdata/p4_16_errors/issue-2123_e.p4 
../testdata/p4_16_errors/issue-2123_e.p4(61):syntax error, unexpected ), expecting ","
     (0x0806 .. 0x0800)
                      ^
[--Werror=overlimit] error: 1 errors encountered, aborting compilation
hemant@ubuntu:~/p4c/build$ 
```

If I change the P4 code by adding a `true, as follows,  compilation passes.

```
 transition select(true, hdr.ethernet.etherType) {
	    (true, 0x0806 .. 0x0800): parse_ipv4;
            default: accept;
        }
```

I wonder what is the proper syntax in P4 to use the parser Range in code shown above.

jafingerhut:
When you have a single field in a select expression, there is no error if you avoid using parentheses around the key expressions.  I believe they are required if you have multiple fields in the select expression, and required _not_ to be there if you have one field in the select expression, and that perhaps this might be the same for const entries in tables?

jafingerhut:
The attached zip file contains 2 P4_16 programs similar to yours, Hemant, so I left your copyright notice in them.  One has an STF file that has test packets that exercise a select expression with ranges on a single field, the other has an STF file that has test packets exercising a select expression with ranges on two fields, so the cross-producting code is exercised.

I have also included the expected frontend and midend output files, which I have checked the expanded select expressions, which all look correct to me.

Thanks for implementing this!
[issue-2123-stf-tests.zip](https://github.com/p4lang/p4c/files/4050308/issue-2123-stf-tests.zip)

jafingerhut:
@hesingh Please add those files to your PR, if you do not mind.

hesingh:
> When you have a single field in a select expression, there is no error if you avoid using parentheses around the key expressions. I believe they are required if you have multiple fields in the select expression, and required _not_ to be there if you have one field in the select expression, and that perhaps this might be the same for const entries in tables?

Got it, @jafingerhut - thanks!  Regarding const entries, if you see this line of P4 code, parenthesis are used with a single field when parenthesis are not needed.

https://github.com/p4lang/p4c/blob/master/testdata/p4_16_samples/action-two-params.p4#L72

hesingh:
Thanks much for the STF tests, @jafingerhut

hesingh:
> You should an stf file to check that indeed the ranges work.

STF files have been checked in and Travis has passed.   Please review, thanks.

testdata/p4_16_errors_outputs/issue-2123_e.p4-stderr
```diff
@@ -0,0 +1,2 @@
+error: 2054-2048: Range end is less than start.
```
jafingerhut: Is there a way to make the ranges appear the same as they did in the source code?  No big deal if not, but thought I would ask.  I know that the source code could have it appear in any base that P4_16 supports, so this would involve some compiler internal to access the source code form of the value.

hesingh: In the p4c code, it's a cinch to find the base of the constant defined in P4 code.  However, I am using p4c's `::error` API for which I don't know how to provide `std::hex` and `std::showbase` to print the constant in hex.

jafingerhut: I just noticed that there is no location within the source code associated with this error message.  A file name and line number would be good, yes?

hesingh: Totally agree.  The error reporting API I am using does not print the information.  Let me take a look.

hesingh: I am using the same `::error` API in midend code as the bmv2 backend simple_switch code.  The backend code, on hitting the error, does print the line number of P4 code.  However, my midend p4c code does not.  I will look into this one.  

The log below is from a p4c build without this PR's changes.  See that line numbers are printed,

```
hemant@ubuntu:~/p4c/build$ ./p4c-bm2-ss --std p4-16 /tmp/p4c/testdata/p4_16_errors/issue-2123_e.p4 
/tmp/p4c/testdata/p4_16_errors/issue-2123_e.p4(61): [--Werror=invalid] error: ..: Invalid must evaluate to a compile-time constant
            0x0806 .. 0x0800 : parse_ipv4;
            ^^^^^^^^^^^^^^^^
/tmp/p4c/testdata/p4_16_errors/issue-2123_e.p4(62): [--Werror=invalid] error: ..: Invalid must evaluate to a compile-time constant
            2054 .. 2048 : parse_ipv4;
            ^^^^^^^^^^^^
hemant@ubuntu:~/p4c/build$ 

```

Printing in same base is fixed by the new checkin.

hesingh: A `IR::Node` needs to be included in the error message and then the source file and line number are printed.  I checked in code.

midend/replaceSelectRange.cpp
```diff
@@ -0,0 +1,175 @@
...
+        return false;
+    }
+    return true;
+}
+
+std::vector<const IR::Mask *>
+DoReplaceSelectRange::rangeToMasks(const IR::Range *r) {
+    bool st = checkRange(r);
+    if (!st)
+        ::error("Range check failed");
```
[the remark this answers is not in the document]
mbudiu-vmw: Also, what are you returning in case of error? You are still going through the loop below, which is probably not what you want.

midend/replaceSelectRange.cpp
```diff
@@ -0,0 +1,175 @@
...
+
+    while (n > 0 && (n & 1) == 0) {
+        zeros += 1;
+        n >>= 1;
+    }
+
+    return (zeros < width) ? zeros : width;
+}
+
+bool DoReplaceSelectRange::checkRange(const IR::Range* range) {
```
mbudiu-vmw: This function is only called once, I think it is better to inline it. Then your code will become much more compact.
E.g.: 
```
auto l = range->left->to<IR::Constant>();
if (!l) {  ::error("..."); 
```

midend/replaceSelectRange.cpp
```diff
@@ -0,0 +1,175 @@
...
+* See the License for the specific language governing permissions and
+* limitations under the License.
+*
+*/
+
+#include "replaceSelectRange.h"
+
+namespace P4 {
+
+static int trailingZeros(big_int n, int width) {
```
mbudiu-vmw: There is a function in gmputil.h called ffs which does this work already.

midend/replaceSelectRange.cpp
```diff
@@ -0,0 +1,175 @@
...
+    std::vector<const IR::Mask *> masks;
+    big_int range_size_remaining = max - min + 1;
+
+    while (range_size_remaining > 0) {
+        big_int range_size = ((big_int) 1) << trailingZeros(min, width);
+
+        while (range_size > range_size_remaining)
+            range_size >>= 1;
+
+        auto constType = typeMap->getType(r->left, true);
```
mbudiu-vmw: I don't think you need the type map, `r->left->type` should have all the information you need.

midend/replaceSelectRange.cpp
```diff
@@ -0,0 +1,175 @@
...
+        auto valConst = new IR::Constant(constType, min, 16, true);
```
mbudiu-vmw: Why use base 16 and not the base of the original constant?

midend/replaceSelectRange.cpp
```diff
@@ -0,0 +1,175 @@
...
+    }
+
+    return newVecs;
+}
+
+const IR::Node*  DoReplaceSelectRange::postorder(IR::SelectCase* sc) {
+    auto newCases = new IR::Vector<IR::SelectCase>();
+    auto keySet = sc->keyset;
+
+    if (keySet->is<IR::Range>()) {
```
mbudiu-vmw: This can be written shorter as `if (auto r = keySet->to<IR::Range>())`

midend/replaceSelectRange.cpp
hesingh: checkRange() is called twice because rangeToMasks() is called twice.

mbudiu-vmw: I can see exactly 1 line in your code with a call to checkRange().

hesingh: Yes, but rangeToMasks() is called twice and thus checkRange() ends up getting calls twice.

jafingerhut: I think Mihai's comment is that regardless of the number of times that checkRange() is called at execution time, it is only called from one place in the source code, so in his opinion, the code would read nicer if there were no separate function called checkRange(), and that its body was simply incorporated into the code where checkRange() is called from.  It isn't a correctness concern, but a code readability style suggestion.

hesingh: Thanks for clarifying, @jafingerhut.  I will change the code.

hesingh: My viewpoint was that the code could modular by having a different checmRange() method and keeping the computing guts in a different rangeToMasks().  I have no strong opinion and can do whatever we go with.

mbudiu-vmw:
If no one has additional comments we can probably merge this.

jafingerhut:
@antoninbas The current proposed changes to p4c require no changes to bmv2, because there is a new midend p4c pass that expands select expression ranges in a parser state into one or more ternary expressions, which have the same effect as the original range.  Thus the existing support in bmv2 for ternary match expressions is enough to do the correct execution of the parser.

Any thoughts before this is merged in?

midend/replaceSelectRange.cpp
```diff
@@ -0,0 +1,152 @@
...
+        masks.push_back(m);
+        range_size_remaining -= range_size;
+        min += range_size;
+    }
+
+    return masks;
+}
+
+std::vector<IR::Vector<IR::Expression>>
+DoReplaceSelectRange::cartesianAppend(std::vector<IR::Vector<IR::Expression>> vecs,
```
antoninbas: why are the vector parameters passed by value and not const reference?

midend/replaceSelectRange.cpp
```diff
@@ -0,0 +1,152 @@
...
+            copy.push_back(mask);
+            newVecs.push_back(copy);
+        }
+    }
+
+    return newVecs;
+}
+
+std::vector<IR::Vector<IR::Expression>>
+DoReplaceSelectRange::cartesianAppend(std::vector<IR::Vector<IR::Expression>> vecs,
```
antoninbas: same as above

antoninbas:
@jafingerhut I did a quick review.

Longer term I think it would be beneficial to add native support for this to bmv2, to facilitate debugging and avoid performance hits for range matches which translate to a large number of ternary rules.

jafingerhut:
@antoninbas Understood on the desire to add native support to bmv2 for this.  I think the most sane way to do so would be to no longer concatenate the values of multiple fields in a select expression together, but keep them as separate fields, similar to how P4 tables keep their fields separate in the BMv2 JSON file.

Regarding a warning, are you thinking that it would be some kind of global thing, e.g. if across the entire P4 program, a total of N parser state transitions is transformed into N+100 or more state transitions, then issue the warning?  It could be localized to individual state transitions, or separately per select expression, but perhaps the "total expansion is at least 100" is what you were thinking?

hesingh:
The cartersian product Antonin spoke of is per select Case for P4 code.  The compiler implementation has a `newCases` vector who size can be checked to exceed hundred to emit an error.

antoninbas:
@jafingerhut yes, there may be some unification possible between table & parser code. But it requires some significant changes to the JSON, thus the "longer term" :)

I was thinking of a per-transition check.

hesingh:
Let's try and understand what you want.  

This is relevant code:

```
transition select(hdr.ethernet.etherType, hdr.ethernet.srcAddr[7:0],  hdr.ethernet.dstAddr[7:0]) {
            // Test three ranges for cartesian product.
            (0x0800 .. 0x0806, 0x08 .. 0x11, 0x08 .. 0x10): parse_ipv4;
            // Test a constant, range, and ternary op in one keyset.
            (0x0800, 0x08 .. 0x10, 0x06 &&& 0x11): parse_ipv4;
            default: accept;
        }
```
I just checked in code to limit each select Case to not exceed a threshold.  However, I think you are asking that all the select Cases inside the `transition select` should not exceed a hundred.  If this is true, the current code does support tracking `transition select` and will make the code changes complicated.

hesingh:
@mbudiu-vmw Would not know how to pass an arg from p4c CLI to a midend pass?

mbudiu-vmw:
This should probably be a warning, if the size grows too much. then there's no need for a command line argument

hesingh:
Thanks, @mbudiu-vmw.  I will change the error to a warning.

Regarding the threshold, the current code can only track a threshold per selectCase.  I can't think of a good number because each selectCase could have large number of ranges or each selectCase has only two ranges, but the number of selectCases is large in one parser state.  There are multiple parser states as well.  It's getting complicated to decide the threshold.   

Folks should notice, the midend pass is only processing `postorder(IR::SelectCase* sc)` and thus tracking a threshold per `transition select`/parser state will complicate the code.  

Maybe, we can allow a threshold of 30 per selectCase?

hesingh:
Right now, on reaching the threshold for a selectCase, the code returns failure:

https://github.com/p4lang/p4c/pull/2125/files#diff-e9f74513df031210ac7355c6025666a1R148

I could change the code to not return failure and let the user continue at his/her own peril for testing with simple_switch.  After, all, the user could run simple_switch on a cloud machine with 32G of RAM, and 500G disk space and the machine is a 64-bit machine.

antoninbas:
I meant per individual transition, so for each "select case".

jafingerhut:
@hesingh @antoninbas Saying a "per-transition check" is unfortunately easy to interpret multiple ways, since the syntax of P4_16 has a single transition keyword in the entire statement "transition select (...) { ... }"

I believe that when Antonin was saying a "per-transition check", he meant per _line in the select expression_, that is, per state transition in the parser state diagram.  That is what Hemant has implemented as an error, and which Mihai suggests should be a warning instead (and I agree it should be a warning, not an error).

If so, a threshold of 30 seems as reasonable as 100.  Both are quite large, and you will likely only hit them if you exercise the Cartesian product across multiple fields, or have a range on a field wider than 16 bits.

Ergonomically, I prefer more detailed messages when possible, so perhaps something like: "Warning: select key set expression with a range expands into 38 ternary key set expressions, which may lead to run-time performance or parser configuration space issues in some targets"

and if it is easy to show the file name, line number, and ^^^^ under the expression like is possible for error messages, that would be good, too.

Note: keysetExpression is the name for this part of the syntax in the P4_16 specification, hence the use of "key set expression" in the proposed warning message.

hesingh:
Thanks @antoninbas.

Then, the current code in this midend pass is tracking the threshold as you suggested.  I can log a warning and still let the user continue?

hesingh:
@jafingerhut I changed the 100 to 4 in the .h file, rebuilt the compiler image, and here is what the error looks like with testing.  The line number that caused the warning is also printed.

```
hemant@ubuntu:/tmp/p4c/build$ ./p4c-bm2-ss --std p4-16 ../testdata/p4_16_samples/issue-2123.p4
../testdata/p4_16_samples/issue-2123.p4(62): [--Wwarn=overlimit] warning: select key set expression with a range expands into 4 ternary key set expressions, which may lead to run-time performance or parser configuration space issues in some targets.
            (0x0800 .. 0x0806, 0x08 .. 0x11, 0x08 .. 0x10): parse_ipv4;
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
hemant@ubuntu:/tmp/p4c/build$ 

```

midend/replaceSelectRange.cpp
```diff
@@ -0,0 +1,159 @@
...
+        for (auto v : newVectors) {
+            auto le = new IR::ListExpression(oldList->srcInfo, v);
+            newCases->push_back(new IR::SelectCase(sc->srcInfo, le, sc->state));
+        }
+        if (newCases->size() > MAX_CASES) {
+            ::warning(ErrorType::ERR_OVERLIMIT,
+                      "select key set expression with a range expands into %2% "
+                      "ternary key set expressions, which may lead to run-time "
+                      "performance or parser configuration space issues in some"
+                      " targets.", sc, MAX_CASES);
```
jafingerhut: I would recommend replacing MAX_CASES in this line with newCases->size(), so that the warning message shows the actual number of ternary key set expressions expanded into.  As it is now, it will always print 100, even if the code causes an expansion into more than 100.

jafingerhut:
I think we have beaten this enhancement into submission :-). Merging it in.

PR #2129, opened 2020-01-08
Correctly parse action references in const table entries

jnfoster:
This pull request fixes a discrepancy between the language specification and `p4c` with respect to the syntax for action references in `const` table entries.

The grammar in the language specification allows action references to be written with or without parentheses:
```
entry
    : keysetExpression ':' actionRef optAnnotations ';'
    ;

actionRef
    : optAnnotations prefixedNonTypeName
    | optAnnotations prefixedNonTypeName '(' argumentList ')'
    ;
```
However, the grammar in the `p4c` parser requires all action references to be given with parentheses, even for actions that do not take arguments:
```
entry
    : keysetExpression ":" actionBinding optAnnotations ";"
        { ... }
    ;

actionBinding
    : lvalue "(" argumentList ")"
        { ... }
    | lvalue "<" typeArgumentList ">" "(" argumentList ")"
        { ... }
    ;
```
This pull request adds a new clause to the `actionBinding` rule in the grammar for the `p4c` parser to handle actions without parentheses. It also includes a unit test (`p4_16_samples/table-entries-no-arg-actions.p4`) to check that such action references are parsed correctly.

### DIFF
--- a/frontends/p4/createBuiltins.cpp
+++ b/frontends/p4/createBuiltins.cpp
@@ -47,6 +47,13 @@ void CreateBuiltins::postorder(IR::ExpressionValue* expression) {
             new IR::Vector<IR::Type>(), new IR::Vector<IR::Argument>());
 }
 
+void CreateBuiltins::postorder(IR::Entry* entry) {
+  // convert a const table entry with action "a;" into "a();"
+  if(entry->action->is<IR::PathExpression>())
+    entry->action = new IR::MethodCallExpression(
+      entry->action->srcInfo, entry->action,
+      new IR::Vector<IR::Type>(), new IR::Vector<IR::Argument>());                                      }
+
 void CreateBuiltins::postorder(IR::ParserState* state) {
     if (state->selectExpression == nullptr) {
         warning(ErrorType::WARN_PARSER_TRANSITION, "%1%: implicit transition to `reject'", state);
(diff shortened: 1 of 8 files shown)


jnfoster:
I just noticed that a second discrepancy: `p4c` does not allow annotations before each action in the list...

ChrisDodd:
Is there a semantic difference between an action with no argument list and an action with an empty argument list?

jnfoster:
In other contexts there is sometimes difference in how the control plane is allowed to configure a table. But tables with `const` entries cannot be modified by the control plane. So I don't think that distinction is relevant here.

jnfoster:
Is the proposal for the specification modification to remove per-action-binding annotations in const table entries? 

FWIW I do think that's the right way to resolve the second discrepancy. 

For the `actions` property in tables, we do allow per-action-ref annotations. And that feature is used for things like `@tableonly` and `@defaultonly`. But note that each action can only appear once. In contrast, the same action can and typically will occur many times in the `const entries`. So it's not clear what different annotations on the same action would even mean.

mbudiu-vmw:
Make the grammar changes in the p4-16 spec.
We don't really have any use at this point for annotations for action entries.

jnfoster:
Sounds good will do.

jnfoster:
Two more thing I just noticeds. Why does `actionBinding` have `lvalue` (which then requires type arguments). Would we lose anything by tightening this down to be just `prefixedNonTypeName`?

mbudiu-vmw:
I think that @cc10512 implemented this feature, so perhaps he can tell us whether there's a reason for allowing lvalue. I expect we don't need it.

mbudiu-vmw:
I think we can merge this and the corresponding spec changes.

jnfoster:
Done! Happy to be an active `p4c` developer again :-)

PR #2137, opened 2020-01-13
Correct checksum pragma collection

AashleshaPatil-BFN:


ir/ir.def
```diff
@@ -452,14 +452,16 @@ class BlockStatement : Statement, ISimpleNamespace, IAnnotated {
     Annotations getAnnotations() const override { return annotations; }
 }
 
-class MethodCallStatement : Statement {
+class MethodCallStatement : Statement, IAnnotated {
```
mbudiu-vmw: The P4-16 spec does not allow annotations on MethodCallStatements.
You will not be able to parse a program that has them.

Issue #2135, opened 2020-01-11
Protobuf documentation is incorrect for p4c


hesingh:
At this URL, the protobuf version asked to be installed is 3.2.0.

https://github.com/p4lang/p4c#ubuntu-dependencies

However, p4c has already moved on to v3.6.1.  

The documentation at the URL should change.

Issue #2117, opened 2019-12-21
Unused code in p4c bmv2 backend


hesingh:
In `p4c/backends/bmv2/common/JsonObjects.cpp`

`JsonObjects::create_primitive(Util::JsonArray* parent, cstring name)`

is not used by any p4c code.  

Should we delete it?

mbudiu-vmw:
Feel free to submit a PR that deletes it.

hesingh:
> Feel free to submit a PR that deletes it.

Will do. thanks.

PR #2138, opened 2020-01-15
Misc minor fixes

hesingh:
Fixes:

https://github.com/p4lang/p4c/issues/2135

https://github.com/p4lang/p4c/issues/2117

and a minor spelling mistake.

### DIFF
--- a/README.md
+++ b/README.md
@@ -221,7 +221,7 @@ versions in the 3 series may not be supported by other p4lang projects, such as
 well, but all our CI testing is done with version 3.2.0. After cloning protobuf
 and before you build, check-out version 3.2.0:
 
-`git checkout v3.2.0`
+`git checkout v3.6.1`
 
 Please note that while all protobuf versions newer than 3.0 should work for
 `p4c` itself, you may run into trouble with some extensions and other p4lang
--- a/README.md
+++ b/README.md
@@ -215,17 +215,17 @@ which is not available until Ubuntu 16.10. For earlier releases of Ubuntu,
 you'll need to install from source. You can find instructions
 [here](https://github.com/google/protobuf/blob/master/src/README.md). **We
 recommend that you use version
-[3.2.0](https://github.com/google/protobuf/releases/tag/v3.2.0)**. Earlier
+[3.6.1](https://github.com/google/protobuf/releases/tag/v3.6.1)**. Earlier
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -221,7 +221,7 @@ versions in the 3 series may not be supported by other p4lang projects, such as
 well, but all our CI testing is done with version 3.2.0. After cloning protobuf
 and before you build, check-out version 3.2.0:
 
-`git checkout v3.2.0`
+`git checkout v3.6.1`
```
jafingerhut: There are multiple occurrences of the version number 3.2.0 in the README.  It would be very confusing only to update this one.

hesingh: Since p4c is now using protobuf 3.6.1, would folks mind if I changed other instances of 3.2.0  to 3.6.1?  thanks.hh

hesingh: I just verified p4lang/PI and even this repo has upgraded to using protobug 3.6.1.  See

https://github.com/p4lang/PI/commit/1539ecd8a50c159b011d9c5a9c0eba99f122a845

jafingerhut: As far as I know, everything still works with protobuf 3.2.0, too.  @antoninbas Any opinion on what the p4c README should say about protobuf versions?  In particular, OK to replace all mentions of 3.2.0 with 3.6.1?  It doesn't seem critical to make such a change if both work, but if other open source P4 tools are "officially supporting" version 3.6.1, it seems to make sense to update all relevant README/install scripts eventually.

hesingh:
I don't know about Bazel being used to build any p4runtime subset of p4c, but the URL in the following file is incorrect.  https://github.com/p4lang/protobuf/archive does not exist.

control-plane/p4runtime/bazel/rules.bzl:            "url": "https://github.com/p4lang/protobuf/archive/v3.2.0-patched.zip",

README.md
hesingh: One may note that p4lang/PI documentation has already moved to using protobuf 3.6.1.

antoninbas:
We should use 3.6.1 consistently for the version. This is the version we used in the docker images in which we run CI.

hesingh:
If there aren't any more comments, anyone will merge permission, please merge this PR - thanks.

hesingh:
thanks!

PR #2139, opened 2020-01-15
Complete p4c bmv2 backend header stack runtime index support

hesingh:
See a prior PR whose repo was deleted to work on another issue.  

https://github.com/p4lang/p4c/pull/2118

This new PR completes the same code and has addressed all comments from the prior PR.

A new p4 program has been added to p4_16_samples.  A STF test has also been added.

This PR also had to wait until runtime index was supported as left-value by the behavioral-model - this support was added around 12/31/2019.


jafingerhut:
A git detail, not related to the changes here, but note that you can create branches in your local git repository where you do development, and make commits to those branches that do not affect master or any other branch, and create PRs from those branches.  This can enable the same repository to help you create, and make updates in response to comments, to multiple PRs from the same repository.

jafingerhut:
Your comment says there is a test P4 program with an STF test, but this is not present in your PR.

hesingh:
Thanks, @jafingerhut Sorry, I missed "git add ." in my repo before the first checkin.  I checked in the pending files.

Yes, I do know about `git checkout -b bug-fix` command to reuse a view but other changes in the branch were interfering with another bug's changes.

hesingh:
Something wrong with Travis right now.  If one clicks on "Details" next to travis-ci/pr, one will see Travis does not report any time elapsed when it should.

hesingh:
I had Travis not working for 10 minutes.  Now, it looks like it's back to working.

antoninbas:
@hesingh all the p4lang projects share a small set of Travis workers, so sometimes there are ongoing CI runs for other repositories and your job is queued until a worker becomes available.

jafingerhut:
I would recommend more than one test packet in the STF file, testing at least writes to the minimum and maximum index in the header stack being assigned to with a run-time variable index.  A single test packet is too easy to make pass even when the implementation has bugs.

Also note that the STF file can be made easier to read by putting spaces between headers.  The spaces within packet data are ignored by the code reading STF files.

hesingh:
Will add more STF tests with min and max.  

Right now, the single test tests for left- and right-value of assignment statement with runtime indices,

jafingerhut:
Also, I believe that unless you rename your files like `runtime-index-bmv2.p4` and `runtime-index-bmv2.stf`, the STF tests will not actually run.

testdata/p4_16_samples/runtime-index-bmv2.p4
```diff
@@ -0,0 +1,95 @@
...
+
+parser MyParser(packet_in packet, out headers hdr, inout metadata_t meta,
+                inout standard_metadata_t standard_metadata) {
+    state start {
+        packet.extract(hdr.ethernet);
+        packet.extract(hdr.ml);
+        packet.extract(hdr.vector[0]);
+        packet.extract(hdr.vector[1]);
+        packet.extract(hdr.vector[2]);
+        packet.extract(hdr.vector[0]);
```
jafingerhut: Did you intend to extract hdr.vector[0] twice?  I would recommend leaving this out, if you didn't intend it -- it makes the test a bit more confusing, and doesn't seem necessary at all to test the new feature.

If you do want to keep it for some reason, the packets in the STF test are shorter than the number of bytes extracted by this parser code.

testdata/p4_16_samples/runtime-index-bmv2.stf
```diff
@@ -0,0 +1,10 @@
+packet abababab00000002000000010000000200000003
+expect abababab000000020000000100000002abababab
```
jafingerhut: I would recommend putting a `$` at the end of each `expect` line, to cause the STF test to verify that the output packet is no longer than the one shown in the `expect` line, and causing the test to fail if the actual output packet is longer than the expected one.

hesingh: Will do.  I have been away from STF for a long time and didn't add port.  I will fix the Travis error.

jafingerhut: You are also missing input port numbers on the packet lines, and output port numbers on the expect lines.

hesingh: I already said in my reply, I didn't add port - I added it.

I am using latest behavioral-model code which has Antonin's l-value changes to runtime index.  But bmv2 test with simple_switch fails (this is the Travis error).   I have narrowed down the error to be caused the the line of P4 code using a runtime index. 

```
hemant@ubuntu:/tmp/p4c/build$ bmv2/testdata/p4_16_samples/runtime-index-bmv2.p4.test -f
Check for  /tmp/p4c/testdata/p4_16_samples/runtime-index-bmv2.stf
terminate called after throwing an instance of 'std::runtime_error'
  what():  in Json::Value::operator[](char const*)const: requires objectValue
*** Test failed
hemant@ubuntu:/tmp/p4c/build$ 
```

hesingh:
[runtime-JSON-clarify.pdf](https://github.com/p4lang/p4c/files/4069191/runtime-JSON-clarify.pdf)

@antoninbas I have attached a pdf to clarify JSON.  The bmv2 test of this PR fails and I suspect it's due to issues in the PR's JSON output.  I have clarification questions in the pdf.

hesingh:
I wrote a new method in `expression.cpp` to handle adding left and right JSON to runtime index array index P4 code. 

```
void ExpressionConverter::riai(cstring left, cstring right) {
    //    mapExpression(expression, result); // TODO

    result->emplace("type", "expression");
    auto e = new Util::JsonObject();
    result->emplace("value", e);
    cstring op = "access_field";

    e->emplace("op", op);
    e->emplace("left", left);
    e->emplace("right", right);
    std::cout << "result: " << result->toString() << std::endl;
}
```

The JSON inside the new method for "result" now follows Antonin's format. See below.  I have to now see what to create for IR::Expression to provide to mapExpression() to complete the code. 

```
result: {
  "type" : "expression",
  "value" : {
    "op" : "access_field",
    "left" : "{
  "type" : "expression",
  "value" : {
    "op" : "dereference_header_stack",
    "left" : {
      "type" : "header_stack",
      "value" : "pool"
    },
    "right" : {
      "type" : "field",
      "value" : ["ml", "idx"]
    }
  }
}",
    "right" : "val"
  }
}
```

antoninbas:
The "right" operand for the "access_field" operator must be the field offset in the header type (so an integer not a string). This is documented in the bmv2 JSON format documentation.

hesingh:
I see - thanks.  

So, where is the "val" value emitted in JSON?  The val is in this line of P4 code:

` hdr.pool[hdr.ml.idx].val = hdr.vector[0].e;`

jafingerhut:
I think the integer index is a pointer/reference to the "val" field, not by name, but by its position in the list of fields that is in the definition of the header inside the BMv2 JSON file.  The string "val" will not be in the BMv2 JSON file for that assignment, only some integer like 2 (if the field "val" was the 3rd field in the header type being assigned to).  In your particular code, it should be 0 since the header type only has one field in it.

jafingerhut:
Or perhaps you are asking where header types are defined in the BMv2 JSON file?  In a dictionary that is the value of the key "header_types"

hesingh:
Thanks @jafingerhut

hesingh:
After clarification of JSON format, I pinged runtime index for the bmv2 backend and checked in code.   There are two TODOs to work on next.  

One is to compute offset of a header field and the other is to look at what's wrong with JSON output for a commented out line of P4 code.

hesingh:
This PR also changed p4c/lib/json.cpp for one error message.  During working with bmv2 backend JSON, many times, I hit the error in this file and the error was useless to help diagnose the problem.  So, I augmented the error message by printing the value as well.

hesingh:
Anyone seen this error working with simple_switch and a bmv2 test? A google search draws a blank for the error.  The first two lines in the log below are from behavioral-mode code.  I tracked the code and see that `get_config` just returns `current_config`.  I don't see anything wrong in the JSON output of my test.  Maybe, the JSON file is too big? 

The P4 program used is from this PR will no lines commented out.

```
ls -l runtime-index-bmv2.json
-rw-r--r-- 1 hemant hemant 14235 Jan 16 19:37 runtime-index-bmv2.json
```

```
Obtaining JSON from switch...
Error when requesting JSON config from switch
*** simple_switch died with return code -11
*** CLI process failed with exit code 1
Execution completed
Removing ./tmp8q93hade
*** Test failed
```

Full log is included below.

```
gemant@ubuntu:/tmp/p4c/build$ bmv2/testdata/p4_16_samples/runtime-index-bmv2.p4.test -v
Writing temporary files into  ./tmp8q93hade
Executing  ./p4c-bm2-ss -o bmv2/testdata/p4_16_samples/runtime-index-bmv2.json /tmp/p4c/testdata/p4_16_samples/runtime-index-bmv2.p4
Exit code  0
Check for  /tmp/p4c/testdata/p4_16_samples/runtime-index-bmv2.stf
Running model
Running simple_switch --log-file switch.log --log-flush --use-files 0 --thrift-port 9341 --device-id 251 -i 0@pcap0 ../bmv2/testdata/p4_16_samples/runtime-index-bmv2.json
Adding interface pcap0 as port 0 (files)
Running simple_switch_CLI --thrift-port 9341
STF Command: packet 0 ca0107fc001c1111111111110800 02 010203 040404
STF Command: expect 0 CA0107FC001C1111111111110800 02 040203 040404 $
STF Command: packet 0 ca0107fc001c1111111111110800 00 010203 040404
STF Command: expect 0 CA0107FC001C1111111111110800 00 040203 040404 $
Obtaining JSON from switch...
Error when requesting JSON config from switch
*** simple_switch died with return code -11
*** CLI process failed with exit code 1
Execution completed
Removing ./tmp8q93hade
*** Test failed
hemant@ubuntu:/tmp/p4c/build$ 
```

jafingerhut:
I have not seen that error, but have you tried running a simple_switch by itself using your BMv2 JSON file, without running the Python scripts around it that execute the STF test?

Given that you are in the process of creating new BMv2 JSON files that may or may not have errors in them, according to simple_switch, that seems like a good thing to try, perhaps running simple_switch inside gdb to see if additional debug info is available there about the problem.

Realize that while simple_switch does some basic sanity checking of parts of the BMv2 JSON file contents when it reads them, for most of the contents it does very little checking, assuming it is correct.

jafingerhut:
If you have not already done so, I would recommend trying to hand-edit the BMv2 JSON file to create what you think is expected by simple_switch, and seeing if you can get it to run functionally correctly, posting copies of the BMv2 JSON files you are trying if they do not work, and you cannot determine why they fail.  Once you have a working BMv2 JSON file, then you can have more confidence knowing that the changes to the p4c back end you make are what is needed.

Or perhaps you already have a working BMv2 JSON file that you edited by hand?

hesingh:
I debugged more and realized the JSON file is not too large.

Soon as I add this line to the P4 program, the error occurs.  Thus, it's confirmed the JSON for this line is the curlprit.

`hdr.pool[hdr.ml.idx].val = hdr.pool[hdr.ml.idx].val + 1;`

thanks.

hesingh:
Yes, I already used hand-edit the JSON file in the morning to debug another issue.  thanks.

hesingh:
The current P4 file checked in passes testing without any hand-editing.  
runtime-index-bmv2.p4.  

Only one line in this program fails simple_switch testing and this is why the line is commented out and marked as TODO.  This is what I am working on right now.

hesingh:
I don't see a way forward with one test in this PR unless the behavioral-model  changes.  This is the line of P4 code  used in the test.

`hdr.pool[hdr.ml.idx].val = hdr.pool[hdr.ml.idx].val + 1;`

Soon as the runtime index is used with an arithmetic expression such as the right-side of the assignment above, JSON output fails simple_switch test. 

First, this is the JSON ouput for runtime index which was proposed and works with simple_switch.

```
{
              "type" : "expression",
              "value" : {
                "type" : "expression",
                "value" : {
                  "op" : "access_field",
                  "left" : {
                    "type" : "expression",
                    "value" : {
                      "op" : "dereference_header_stack",
                      "left" : {
                        "type" : "header_stack",
                        "value" : "pool"
                      },
                      "right" : {
                        "type" : "field",
                        "value" : ["ml", "idx"]
                      }
                    }
                  },
                  "right" : 0
                }
              }
            }
```

This is the JSON output for right of assignment statement.  This JSON does fails simple_switch.  The error is "Error when requesting JSON config from switch".  I have reason to believe, this errors is generated when JSON is not accepted by the simple_switch.  I have included arrows "<===" to highlight one issue.  There are two "type, value" pairs from runtime index JSON when runtime index is merged with arithmetic operations.   I manually edited the json to include only one "type,value" pair and still the error persists when testing with simple_switch.  I edited bmv2stf.py to read my json file.  
If the JSON generated is fine, then doesn't the behavioral-model needs to change?


```
{
              "type" : "expression",
              "value" : {
                "type" : "expression",
                "value" : {
                  "op" : "&",
                  "left" : {
                    "type" : "expression",
                    "value" : {
                      "op" : "+",
                      "left" : {
                        "type" : "expression",  <=== 
                        "value" : {
                          "type" : "expression", <===
                          "value" : {
                            "op" : "access_field",
                            "left" : {
                              "type" : "expression",
                              "value" : {
                                "op" : "dereference_header_stack",
                                "left" : {
                                  "type" : "header_stack",
                                  "value" : "pool"
                                },
                                "right" : {
                                  "type" : "field",
                                  "value" : ["ml", "idx"]
                                }
                              }
                            },
                            "right" : 0
                          }
                        }
                      },
                      "right" : {
                        "type" : "hexstr",
                        "value" : "0x01"
                      }
                    }
                  },
                  "right" : {
                    "type" : "hexstr",
                    "value" : "0xff"
                  }
                }
              }
            }
```

jafingerhut:
It seems to me if you have the ingredients below, and can publish them in an issue on the p4lang/behavioral-model repository, that would help others debug the issue.

+ a complete BMv2 JSON file that you believe is correct, but gives an error when run with the command `simple_switch my-file.json` from the command line (no bmv2stf.py needs to be involved -- that only introduces other possibilities for problems)

+ point out the range of line numbers of that file that you think are causing the problem

+ The strongest evidence that those line numbers are causing a problem would be to publish a second variation of that BMV2 JSON file, maybe hand-edited from the first one, where `simple_switch my-file-edited.json` gives no error, and works as desired, because it has no access to a run-time variable index of a header stack on the right-hand side of the assignment, but simply a constant value (for example).

hesingh:
Found the root cause.

In my prior note from today, I said there are two "type, value" pairs before the "op : access_field".  I removed the first one and then simple_switch testing passes.   In my prior note I said even after removing one "type, value", test fails, but it was due to a missed edit in bmv2stf.py.  I edited the .py code in another place and then testing passed.   These are the changes.

```
hemant@ubuntu:/tmp/p4c$ git diff backends/bmv2/bmv2stf.py
diff --git a/backends/bmv2/bmv2stf.py b/backends/bmv2/bmv2stf.py
index e6ac0737..3d35e95d 100755
--- a/backends/bmv2/bmv2stf.py
+++ b/backends/bmv2/bmv2stf.py
@@ -366,7 +366,8 @@ class RunBMV2(object):
         self.target_specific_cmd_line_args = getattr(options, 'switchTargetSpecificOptions', ())
 
     def readJson(self):
-        with open(self.jsonfile) as jf:
+#        with open(self.jsonfile) as jf:
+        with open("/tmp/p4c/build/tmp.json") as jf:
             self.json = json.load(jf)
         for a in self.json["actions"]:
             self.actions.append(BMV2Action(a))
@@ -578,7 +579,7 @@ class RunBMV2(object):
             runswitch = [FindExe("behavioral-model", switch),
                          "--log-file", self.switchLogFile, "--log-flush",
                          "--use-files", str(wait), "--thrift-port", thriftPort,
-                         "--device-id", str(rand)] + self.interfaceArgs() + ["../" + self.jsonfile]
+                         "--device-id", str(rand)] + self.interfaceArgs() + ["/tmp/p4c/build/tmp.json"]
             if self.cmd_line_args:
                 runswitch += self.cmd_line_args
             if self.target_specific_cmd_line_args:

```

As shown in the new issue filed against the behavioral-model, using simple_switch CLI does not reproduce the issue.  This is why I used the other test shown below to test.

```
hemant@ubuntu:/tmp/p4c/build$ bmv2/testdata/p4_16_samples/runtime-index-bmv2.p4.test -v
Writing temporary files into  ./tmppv2vnkp2
Executing  ./p4c-bm2-ss -o bmv2/testdata/p4_16_samples/runtime-index-bmv2.json /tmp/p4c/testdata/p4_16_samples/runtime-index-bmv2.p4
Exit code  0
Check for  /tmp/p4c/testdata/p4_16_samples/runtime-index-bmv2.stf
Running model
Running simple_switch --log-file switch.log --log-flush --use-files 0 --thrift-port 9473 --device-id 383 -i 0@pcap0 /tmp/p4c/build/tmp.json
Adding interface pcap0 as port 0 (files)
Running simple_switch_CLI --thrift-port 9473
STF Command: packet 0 ca0107fc001c1111111111110800 02 010203 040404
STF Command: expect 0 CA0107FC001C1111111111110800 02 050203 040405 $
STF Command: packet 0 ca0107fc001c1111111111110800 00 010203 040404
STF Command: expect 0 CA0107FC001C1111111111110800 00 050203 050404 $
Obtaining JSON from switch...
Done
RuntimeCmd: 
simple_switch exit code -15
Execution completed
Comparing outputs
Removing ./tmppv2vnkp2
```

At least for me, it's tricky to change the p4c bmv2 backend to fix the JSON problem.  If one "type, value" pair is removed for the arithmetic and runtime index case, then the "runtime index only" case break down for JSON output because this case needs the extra "type, value" pair.

jafingerhut:
@hesingh Let me know when you think this is ready to go.  I want to go through the test program and STF file one more time before commit, and perhaps add another test or two.

hesingh:
@jafingerhut I plan to add a test or two and this is why I did not send a note for folks to review.  I will do so when the code is ready.

hesingh:
@jafingerhut I plan to add a test or two and this is why I did not send a note for folks to review.  I will do so when the code is ready.

hesingh:
@jafingerhut I plan to add a test or two and this is why I did not send a note for folks to review.  I will do so when the code is ready.

hesingh:
I tried to add more STF tests by changing the data structure from

```
header aggregator_t {
    bit<8> val;
}
```
to 

```
header aggregator_t {
    bit<8> base;
    bit<8> val;
}
```

but when using the P4 code such as `hdr.pool[hdr.ml.idx].val =`, the runtime index causes vector size checking to lose its brain and cause this error:

```
Running simple_switch --log-file switch.log --log-flush --use-files 0 --thrift-port 9144 --device-id 54 -i 0@pcap0 ../bmv2/testdata/p4_16_samples/runtime-index-bmv2.json
Adding interface pcap0 as port 0 (files)
Running simple_switch_CLI --thrift-port 9144
STF Command: packet 0 ca0107fc001c1111111111110800 02 010203 000100010001
STF Command: expect 0 CA0107FC001C1111111111110800 02 050203 000100010001 $
terminate called after throwing an instance of 'std::out_of_range'
  what():  vector::_M_range_check: __n (which is 8) >= this->size() (which is 3)
```

It will take me longer to figure out this issue.   I think, we could punt such tests to another PR.

Please review this PR - thanks.

hesingh:
One Travis job has passed with latest checkin.

@antoninbas and/or @mbudiu-vmw please review the p4c changes.

@jafingerhut please review the test cases.

thanks.

jafingerhut:
I have only reviewed the test program and STF file in this PR, and do not plan to take the time to learn the modified parts of the p4c code well enough to determine whether they are correct.

I would recommend changing the test program and STF file to the one in the attached .zip file.

First, the test input packets given are shorter than the headers extracted by the parser, so the last two headers attempted to be extracted will always be invalid.

Second, several of the assignments made in the P4_16 program later have their results overwritten by another assignment, so in a test that only looks at the output packet, it is impossible to tell whether those assignments operated correctly or not.  The proposed small changes to the test program should make the result of every assignment visible in the output packets, so that if the implementation does not do those assignments correctly, or not at all, the test will fail.

[runtime-index-bmv2.zip](https://github.com/p4lang/p4c/files/4089288/runtime-index-bmv2.zip)

hesingh:
Thanks @jafingerhut!  Sorry, I had the packet length correct when the aggregator_t header had only one 8-bit field.  Thanks for the fix and the other code.

hesingh:
@antoninbas Code is ready for review.

Also, @mbudiu-vmw If you'd like to review changes to type.def and any other file.

mbudiu-vmw:
I have already written my review.

hesingh:
@mbudiu-vmw I have addressed all your review comments with latest code.

mbudiu-vmw:
I don't see any new commits after my review

hesingh:
Sorry,  your comments were made with an old PR: https://github.com/p4lang/p4c/pull/2118

Due to a repo deletion on my machine, I had to pull a new PR which is this one.  This one has  new commits after your review on the old PR.   I have addressed these issues raised by you.

1. I stopped modifying an existing test and wrote a new one.
2. I had STF file with packet tests.
3. All new and existing tests have passed.

backends/bmv2/common/action.cpp
```diff
@@ -70,9 +70,14 @@ void ActionConverter::convertActionBody(const IR::Vector<IR::StatOrDecl>* body,
             auto primitive = mkPrimitive(operation, result);
             auto parameters = mkParameters(primitive);
             primitive->emplace_non_null("source_info", assign->sourceInfoJsonObj());
-            auto left = ctxt->conv->convertLeftValue(l);
-            parameters->append(left);
             bool convertBool = type->is<IR::Type_Boolean>();
+            Util::IJson* left;
+            if (ctxt->conv->IsArrayIndexRuntime(l)) {
+                left = ctxt->conv->convert(l, true, true, convertBool);
```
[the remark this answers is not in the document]
hesingh: It is needed to deal with the case when a runtime index is a left value and is the only expression in JSON (without any other arithmetic operations).  The JSON format needs an extra wrapping `"type": "expression"`.

Regarding convertLeftValue, note, the right value is also using `convert` because simple_switch supports runtime index in right value with simple_switch.  See this code few lines below the left code.

`auto right = ctxt->conv->convert(r, true, true, convertBool);`

backends/bmv2/common/expression.cpp
```diff
@@ -184,32 +184,57 @@ void ExpressionConverter::postorder(const IR::Constant* expression)  {
...
-        auto path = expression->left->to<IR::PathExpression>();
         elementAccess = path->path->name;
     }
 
     if (!expression->right->is<IR::Constant>()) {
-        ::error(ErrorType::ERR_INVALID, "all array indexes must be constant", expression->right);
+        result->emplace("type", "expression");
+        const IR::Expression* ex = expression->right;
+        std::vector<cstring> cv;
+        while (auto rmem = ex->to<IR::Member>()) {
```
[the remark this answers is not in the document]
hesingh: Good catch.  Let me see what I can do.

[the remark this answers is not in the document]
hesingh: It's not a visit subexpression and recursive call issue. 
Arithmetic expression such as `a+b` for constat ArrayIndex is not processed in the backend.  

I decided to debug a constant array index `a+b`and see that the expression is simplified to the sum of `a+b` by the time I dump FrontEndLast.  

`hdr.vector[0+1].e = hdr.pool[1].val + 1;`

changes to

`hdr.vector[1].e = hdr.pool[1].val + 1;`

Since a runtime index cannot be constant folded, the ArrayIndex reaches the bmv2 backend as `hdr.pool[hdr.ml.idx + 8s1].val`  and is processed first by 'void ExpressionConverter::postorder(const IR::IntMod* expression)' before processing calls `void ExpressionConverter::postorder(const IR::ArrayIndex* expression)`.  

With `a+b` for runtime index an IntMod expression is sent to postorder(ArrayIndex).  With this new code added

```
if (!expression->right->is<IR::Constant>()) {
        result->emplace("type", "expression");
        const IR::Expression* ex = expression->right;
        std::vector<cstring> cv;
        if (auto im = ex->to<IR::IntMod>()) { <=== start new code
	  if (auto bio = im->expr->to<IR::Add>()) {
	    ex = bio->left;
	    std::cout << "AIM:" << bio->left << std::endl;
	  }
        }  <=== end new code.
```

bio->left is `hdr.ml.idx;` and rest of the code runs to completion. 

A question for Antonin:

If the P4 code is ` hdr.pool[hdr.ml.idx+1]`, what should the JSON look like because we are in uncharted territory.  Would it be the one shown below?

"value" : ["ml", "idx+1"]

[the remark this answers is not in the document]
jafingerhut: I am almost certain that is not what should be in the BMv2 JSON.

There is already defined JSON expressions for expressions like (hdr.ml.idx + 1) appearing on the right hand side of an assignment.

Most likely whatever that JSON is, is what should be in place of the field member access hdr.ml.idx now.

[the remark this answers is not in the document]
jafingerhut: FYI, attached is a proposed second test P4 program and STF file that exercises arithmetic expressions inside of the run time array indexes.  I have not tested whether the STF is correct or not, since there is no implementation to try it on, but there are extensive comments in the STF file explaining why I think it might be correct.

jafingerhut:
Trying to attach a zip file to this comment, since it wasn't working for the other comment I recently made.
[runtime-index-2-bmv2.zip](https://github.com/p4lang/p4c/files/4095103/runtime-index-2-bmv2.zip)

backends/bmv2/common/expression.cpp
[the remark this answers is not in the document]
hesingh: @jafingerhut I know.  However, the postorder(ArrayIndex) method does not deal with arithmetic expressions.  Some other code does.   I am not sure right now, if I add arithmetic to the runtime index processing in postorder(ArrayIndex).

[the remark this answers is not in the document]
hesingh: Arithmetic expressions with runtime index is not supported by this PR code changes yet.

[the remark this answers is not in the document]
jafingerhut: I understand it is not supported yet.  If/when it is supported, it will be useful to have a test case to exercise that it works.  Hence my second attached zip file.

hesingh:
With constant array index, frontend constant folding simplifies the array index.  Support complex arithmetic in bmv2 backend for runtime index will get very complicated.

jafingerhut:
So I do not know for certain, but Mihai mentioned calling something
recursively to get the job done.

For example, if you have a complex nested arithmetic expression like
(something1) + (something2), where something1 and something2 are themselves
complex arithmetic expressions, I don't think that people write into the
compiler special cases for every possible subexpression something1 and
something2.  They emit a + operation with a left operand that is the output
of a recursive call on a function that emits the JSON for "something1".
When that returns, they make another recursive call to the same function,
but this time for "something2".  When that returns, the the JSON for
"(something1) + (something2)" has now been successfully generated.

Andy

hesingh:
As I said earlier, I am on call path from IntMod to ArrayIndex postorder methods.  I added debugged code to expression.cpp and got to the bottom of what's going on.

First this is how I test:

```
hemant@ubuntu:/tmp/p4c/build$ ./p4c-bm2-ss --std p4-16 ../testdata/p4_16_samples/runtime-index-bmv2.p4 -o tmp.json -Texpression.cpp:2
hemant@ubuntu:/tmp/p4c/build$ 
```

These are the debugged changes to expression.cpp.

```
diff --git a/backends/bmv2/common/expression.cpp b/backends/bmv2/common/expression.cpp
index b04b53fd..ff5f7d31 100644
--- a/backends/bmv2/common/expression.cpp
+++ b/backends/bmv2/common/expression.cpp
@@ -205,11 +205,19 @@ void ExpressionConverter::postorder(const IR::ArrayIndex* expression)  {
         result->emplace("type", "expression");
         const IR::Expression* ex = expression->right;
         std::vector<cstring> cv;
+        if (auto im = ex->to<IR::IntMod>()) {
+	    if (auto bio = im->expr->to<IR::Add>()) {
+	        ex = bio->left;
+	        std::cout << "AIM:" << bio->left << std::endl;
+	    }
+        }
+
         while (auto rmem = ex->to<IR::Member>()) {
             ex = rmem->expr;
             cv.push_back(rmem->member.toString());
         }
-
+        if (cv.empty())
+	    ::error("ArrayIndex runtime failed");
         std::reverse(cv.begin(), cv.end());
 
         // create a DEREFERENCE_HEADER_STACK json node
@@ -501,6 +509,10 @@ void ExpressionConverter::postorder(const IR::Mux* expression)  {
 }
 
 void ExpressionConverter::postorder(const IR::IntMod* expression)  {
+    if (auto im = expression->to<IR::IntMod>()) {
+        std::cout << "IntMod:" << im->expr << std::endl;
+    }
+
     mapExpression(expression, result);
     result->emplace("type", "expression");
```

This is the log from the test of debugged code.  Notice IntMod method processes the arithmetic fully before sending a consolidated IntMod to the ArrayIndex method.  This is the line of P4 code that led to the log below.  In the log you will see that first the "Shr" (shift right) is processed.  Next the subtraction is processed.  Then the consolidated subtraction IntMod is sent to ArrayIndex.  So the ArrayIndex method can now do what Mihai is saying and process the consolidated IntMod. 

```
hdr.vector[hdr.ml.idx1 - (hdr.ml.idx2 >> 1)].e =
            hdr.ethernet.etherType[15:8] + 7;
```

```
left: hdr.vector; right: 0
left: hdr.vector; right: 1
left: hdr.vector; right: 2
left: hdr.vector; right: 3
left: hdr.vector; right: 4
left: hdr.vector; right: 5
left: hdr.vector; right: 6
left: hdr.vector; right: 7
IntMod:hdr.ml.idx2 >> 1;
IntMod:hdr.ml.idx1 - <IntMod>(150732);
left: hdr.vector; right: <IntMod>(150733)
error: ArrayIndex runtime failed
left: hdr.vector; right: 0
left: hdr.vector; right: 1
left: hdr.vector; right: 2
left: hdr.vector; right: 3
left: hdr.vector; right: 4
left: hdr.vector; right: 5
left: hdr.vector; right: 6
left: hdr.vector; right: 7
```

hesingh:
@mbudiu-vmw and @antoninbas  A visit and recursive function was not needed.  By the time postorder(ArrayIndex) is hit, the `map` already has the arithmetic JSON generated.  So, I find the generated JSON and use the JSON in rest of the runtime index code to generate the op = dereference_header_stack.

hesingh:
Once testing passes with runtime index and arithmetic, I will remove the std::cout from the code.   I realized I just checked in expression.cpp with a std::cout.

jafingerhut:
@hesingh I have attached an updated version of the P4 program and STF test.  The previous one with the same file names inside the zip file definitely did seem to attempt to access a header stack index that was out of range, and should have caused simple_switch to crash because of that.

Hopefully this one should not.
[runtime-index-2-bmv2-corrected.zip](https://github.com/p4lang/p4c/files/4100112/runtime-index-2-bmv2-corrected.zip)

hesingh:
@jafingerhut I had to comment out the second packet in the latest STF file.  Then I changed the first expected packet to 
`expect 0 CA0107FC001C111111111111F00D08050123456789ABF7EF`

and this line of P4 code passed testing for arithmetic with runtime index.  

```
hdr.vector[hdr.ml.idx1 - (hdr.ml.idx2 >> 1)].e =
            hdr.ethernet.etherType[15:8] + 7;
```

If I use the 2nd packet and only the above line of P4 code, simple_switch crashes.  If I enable the 2nd line of P4 code which is

`hdr.ethernet.etherType[7:0] = hdr.vector[(hdr.ml.idx2 ^ 0x07) & 0x7].e;`

simple_switch crashes.

If you download my latest p4c code (https://github.com/hesingh/p4c), this code should work for any arithmetic expression with runtime index as far as the STF packet does not cause a std::out_of_range exception.

thanks.

hesingh:
One of two packets in updated STF file that Andy sent out, passed testing of arithmetic with runtime index.  The two line of P4 code (runtime-index-2-bmv2.p4) has the arithmetic.

hesingh:
@mbudiu-vmw or anyone else, could you please use your latest p4c repo and run this test and see if you all fail like I do?

```
hemant@ubuntu:~/p4c/build$ ./p4c-bm2-ss --std p4-16 ../testdata/p4_16_samples/issue1989-bmv2.p4 -o tmp.json
../testdata/p4_16_samples/issue1989-bmv2.p4(58): [--Werror=invalid] error: meta.color: Invalid all array indexes must be constant
    hdr.ethernet_stack[1] = hdr.ethernet_stack[meta.color];
                                               ^^^^^^^^^^
hemant@ubuntu:~/p4c/build$ 
```

I am chasing this test because Travis for this PR fails this test and does not give any obvious error.  To  debug, I decided to test latest p4c without my changes.   If this test fails with p4c, why doesn't a negative -stderr exist for this test?

With this PR's code changes, this test does not fail.  Thanks.

hesingh:
This is Travis failure iwith this PR with no obvious error that I can discern.

```
695/1414 Test #216: bmv2/testdata/p4_16_samples/issue1989-bmv2.p4 .................................***Failed    3.47 sec
Obtaining JSON from switch...
Done
RuntimeCmd: 
Check for  /p4c/testdata/p4_16_samples/issue1989-bmv2.stf
```

hesingh:
I have fixed the crash with 2nd packet in STF file with arithmetic with runtime index.

Please review the latest code changes.

hesingh:
Why is issue1989-bmv2.p4 named with a tag `-bmv2.p4` when this P4 program does not have a corresponding STF file?

jafingerhut:
There are many P4_16 source files with a file name ending in -bmv2.p4 for which there are no corresponding .stf files.

Someone else can probably answer better why.  I think perhaps the intent might be that they use an empty .stf file by default.  One possible testing advantage to running a test with an empty .stf file over not running an STF test at all, is that it confirms that it produces a BMv2 JSON file that simple_switch can read without error.  Not sure if that is the intent, but it would be a reasonable one.  Actual packets in and comparisons to packets out are stricter tests, of course.

jafingerhut:
Probably also that without the suffix -bmv2.p4 on the file name, it does not execute the p4c bmv2 back end code.

jafingerhut:
It is not clear to me why the bmv2 test for file issue1989-bmv2.p4 is failing in Travis.  It passes on my local machine, with and without the changes in this PR.

jafingerhut:
@hesingh Please update the runtime-index-2-bmv2.stf file to the one attached to this comment.  It has the same input and output packets, but the comments are corrected in the attached version, which should make it easier to update the test when the P4 compiler correctly handles the commented-out line of code in runtime-index-2-bmv2.p4

[runtime-index-2-bmv2.stf.txt](https://github.com/p4lang/p4c/files/4101439/runtime-index-2-bmv2.stf.txt)

mbudiu-vmw:
All programs that end in -bmv2.p4 are run also with the p4c-bm2-ss compiler, and the json produced is validated by bmv2. If there is no stf file they are run with an empty stf file.

hesingh:
Thanks @mbudiu-vmw and @jafingerhut .  I pulled a new repo of p4lang/p4c yesterday on my local machine.  The test fails for me.

```
hemant@ubuntu:~/p4c/build$ bmv2/testdata/p4_16_samples/issue1989-bmv2.p4.test -v
Writing temporary files into  ./tmppc11hz82
Executing  ./p4c-bm2-ss -o bmv2/testdata/p4_16_samples/issue1989-bmv2.json /home/hemant/p4c/testdata/p4_16_samples/issue1989-bmv2.p4
Exit code  1
Error compiling
/home/hemant/p4c/testdata/p4_16_samples/issue1989-bmv2.p4(58): [--Werror=invalid] error: meta.color: Invalid all array indexes must be constant
    hdr.ethernet_stack[1] = hdr.ethernet_stack[meta.color];
                                               ^^^^^^^^^^

Removing ./tmppc11hz82
*** Test failed
hemant@ubuntu:~/p4c/build$ 
```

mbudiu-vmw:
The test is blacklisted in backends/bmv2/CMakeLists.txt; just remove the corresponding line and comment.

mbudiu-vmw:
We don't have a directory for tests that should pass with p4test but succeed with p4c-bm2-ss.

backends/bmv2/common/expression.cpp
```diff
@@ -184,32 +184,51 @@ void ExpressionConverter::postorder(const IR::Constant* expression)  {
...
-        auto path = expression->left->to<IR::PathExpression>();
         elementAccess = path->path->name;
     }
 
     if (!expression->right->is<IR::Constant>()) {
-        ::error(ErrorType::ERR_INVALID, "all array indexes must be constant", expression->right);
+        const IR::Expression* ex = expression->right;
+        auto res = map.find(ex);
+        Util::JsonObject* fres;
+        if (res != map.end()) {
```
mbudiu-vmw: Please use the pattern used everywhere else: `get(expression)`, it does it's own error reporting.

backends/bmv2/common/expression.cpp
```diff
@@ -184,32 +184,51 @@ void ExpressionConverter::postorder(const IR::Constant* expression)  {
...
+            if ((fres = res->second->to<Util::JsonObject>())) {
+                LOG2("found result: " << fres->toString());
+            } else {
+                ::error("Failure in ArrayIndex runtime processing");
+            }
+        }
+        result->emplace("type", "expression");
+
+        // create a DEREFERENCE_HEADER_STACK json node
```
mbudiu-vmw: this comment is not particularly useful; the next line says the same thing.

backends/bmv2/common/expression.cpp
```diff
@@ -403,6 +429,15 @@ void ExpressionConverter::postorder(const IR::Member* expression)  {
                 } else if (lv->is<Util::JsonValue>()) {
                     e->append(lv);
                     e->append(fieldName);
+                } else if (auto jo = l->to<Util::JsonObject>()) {
+                    if (st) {
```
mbudiu-vmw: what happens if st is false? If it cannot be false then please add a `BUG_CHECK(st)`.

lib/json.cpp
```diff
@@ -192,8 +192,11 @@ JsonObject* JsonObject::emplace(cstring label, IJson* value) {
     if (label.isNullOrEmpty())
         throw std::logic_error("Empty label");
     auto j = get(label);
-    if (j != nullptr)
-        throw std::logic_error(cstring("Duplicate label in json object ") + label.c_str());
+    if (j != nullptr) {
+      cstring s = value->toString();
+      throw std::logic_error(cstring("Duplicate label in json object ")
```
mbudiu-vmw: this error seems now a bit incorrect; perhaps it should say: "attempt to add to json object a *value* for a label *label* which already exists.

backends/bmv2/common/expression.cpp
hesingh: If the P4 program does not use runtime index, then `st` is false and this else if will not even be hit:

https://github.com/hesingh/p4c/blob/master/backends/bmv2/common/expression.cpp#L432

hesingh:
> The test is blacklisted in backends/bmv2/CMakeLists.txt; just remove the corresponding line and comment.

Removing the line fixed the Travis error - thanks!

hesingh:
At this juncture, I am not sure what changes are needed to p4c bmv2 backend's LowerExpressions pass to fix https://github.com/p4lang/p4c/issues/2150.

I think, we can checkin this PR after code review is complete, and have a separate PR to fix https://github.com/p4lang/p4c/issues/2150.

frontends/p4/typeChecking/typeChecker.h
```diff
@@ -83,8 +83,21 @@ class TypeInference : public Transform {
...
+                    return true;
+                }
+            }
+        }
+        return false;
+    }
+
     bool isLeftValue(const IR::Expression* expression) const
-    { return typeMap->isLeftValue(expression) || expression->is<IR::DefaultExpression>(); }
+    { return typeMap->isLeftValue(expression) || expression->is<IR::DefaultExpression>()
```
mbudiu-vmw: This is incorrect; the definition of left values is very clear in the spec.

hesingh: I searched in the p4-16 spec and do not see LeftValue in regards to AssignmentStatement.

Could you please do me a favor and use this URL below to show me where in the spec is the definition of values.

https://github.com/p4lang/p4-spec/blob/master/p4-16/spec/P4-16-spec.mdk

jafingerhut: This might be the section Mihai is thinking of: https://p4.org/p4-spec/docs/P4-16-v1.2.0.html#sec-lvalues

That describes the syntax and the meaning of expressions that can appear as an L-value.

How that ought to be implemented in the compiler to enable an expression like `hdr.vector[hdr.vector[hdr.ethernet.etherType[2:0]].e].e` to be allowed on the left hand side of an assignment, I have no idea.

mbudiu-vmw:
There is nothing wrong with the typechecker, which maintains a map with all expressions that are left values, and applies these rules to compute whether another expression is also a left value. If there is a bug it must be in the lower pass.

frontends/p4/typeChecking/typeChecker.h
hesingh: Thanks @jafingerhut 

Maybe the code is correct.

This is text from the same L-Value section of the spec.

```
The following is a legal l-value: `headers.stack[4].field`
```
A constant array index is allowed in the above expression.  Ours is a runtime index expression.  Why should ours not be allowed?  The only reason I can think of is that expression count is greater than zero.

Additionally, I was also looking at the grammar in p4c/frontend/p4/parsers/p4/p4parser.ypp and clearly see this grammar:

```
lvalue
    : prefixedNonTypeName                { $$ = new IR::PathExpression($1); }
    | THIS                               { $$ = new IR::This(@1); }  // experimental
    | lvalue "." name                    { $$ = new IR::Member(@1 + @3, $1, *$3); }
    | lvalue "[" expression "]"          { $$ = new IR::ArrayIndex(@1 + @4, $1, $3); }
    | lvalue "[" expression ":" expression "]" { $$ = new IR::Slice(@1 + @6, $1, $3, $5); }
    ;
```

A lvalue can include an ArrayIndex.

hesingh:
Right now, all of bmv2 backend's runtime index code in this PR works because the left and right values of assignment are received by the backend as is from the from the P4 program.  If the lower pass changes the IR, it's more work for this PR.  

I'd like to know what would the lower pass do with this expression?

` hdr.vector[hdr.vector[hdr.ethernet.dstAddr[39:32] & 0x7].e & 0x7].e`

hesingh:
I read the one-line description of the lower pass.  Here it is.

https://github.com/p4lang/p4c/blob/master/backends/bmv2/common/lower.h#L27

The pass does not need to rewrite anything because the runtime array index expression can be natively handled in the bmv2 backend by code changes of this PR.  So, I question if the lower pass has any bug or the pass description could use more details to explain this case.

mbudiu-vmw:
The JSON does not support slice expressions, so that needs to be lowered to something else.

hesingh:
> The JSON does not support slice expressions, so that needs to be lowered to something else.

Indeed today, I debugged the lower pass and the runtime index expression hits the Slice portion of lower pass.  This is why the expression could be messed up.  thanks.

hesingh:
Sorry, I didn't know that Slice is not supported by the backend and duh, this is why it's in lower.cpp.

ir/type.def
```diff
@@ -233,6 +233,14 @@ abstract Type_StructLike : Type_Declaration, ISimpleNamespace, IAnnotated {
         return fields.getDeclarations(); }
     StructField getField(cstring name) const {
         return fields.getDeclaration<StructField>(name); }
+    int getOffset(cstring name) const {
```
mbudiu-vmw: Please change this function name to something more explicit, such as getFieldIndex.

ir/type.def
```diff
@@ -233,6 +233,15 @@ abstract Type_StructLike : Type_Declaration, ISimpleNamespace, IAnnotated {
...
     StructField getField(cstring name) const {
         return fields.getDeclaration<StructField>(name); }
+    int getFieldIndex(cstring name) const {
+        int index_pos = 0;
+        for (auto f : fields) {
+            if (f->name == name)
+                return index_pos;
+	    index_pos++;
+        }
+        return index_pos;
```
mbudiu-vmw: How about returning -1 for an error?
Otherwise this could be confused with a regular index.

hesingh: Agree.  However, I see that lot of test P4 programs uses data struct or header with zero fields.  So many Travis tests failed with the new error.  I am looking code for final fix.

hesingh:
Anyone else has any comments?

hesingh:
@jafingerhut or @mbudiu-vmw As Andy has said in the past "we have beaten this PR into submission".  Please check in the code - thanks.

testdata/p4_16_samples/runtime-index-2-bmv2.p4
```diff
@@ -0,0 +1,105 @@
...
+                inout standard_metadata_t standard_metadata) {
+    apply {
+        // Test runtime index with arithmetic expression as index, as
+        // L-value.
+        hdr.vector[hdr.ml.idx1 - (hdr.ml.idx2 >> 8w1)].e =
+            hdr.ethernet.etherType[15:8] + 7;
+        // similar, but with runtime index as R-value.
+        hdr.ethernet.etherType[7:0] = hdr.vector[(hdr.ml.idx2 ^ 8w0x07) & 8w0x7].e;
+
+        // TODO: The line of code below is commented out
```
mbudiu-vmw: Can you please submit a separate PR with this test case once this is merged?

hesingh: Sure, will do.

PR #2141, opened 2020-01-17
Add new match kind `optional` to v1model

stefanheule:
`optional` is a new match kind that matches either exactly, or is a wildcard.  Note that unlike `ternary`, there is no mask, it is either a full match on all bits, or a complete wildcard.  This was proposed to be added to the language, and in the LDWG meeting from Jan 6, 2020, we decided the newly proposed match kind  should be added to v1model and P4RT first, and possibly later consider promoting it to the `core.p4` standard library.  This PR is to add it to v1model, and a companion PR will add support to P4RT soon.

More details on the original discussion in the LDWG meeting notes and here: https://github.com/p4lang/p4-spec/issues/794.

### DIFF
--- a/p4include/v1model.p4
+++ b/p4include/v1model.p4
@@ -52,6 +52,8 @@ limitations under the License.
 
 match_kind {
     range,
+    // Either an exact match, or a wildcard (matching any value).
+    optional,
     // Used for implementing dynamic_action_selection
     selector
 }


jafingerhut:
I am not sure, but was it perhaps the intent of the LDWG request to implement this feature in v1model to make all changes necessary for a P4_16 program with the v1model architecture compile and run with the desired behavior using the simple_switch behavioral-model?

If so, I suspect a fair number of additional changes would be needed, including at least one test program with STF test, and generation of the correct match_kind in the P4Info file written by p4c.

Were you planning to look into those kinds of changes, too?  Or perhaps someone else has volunteered for the other parts?

jafingerhut:
If you or someone you know of are thinking of looking into everything that needs to be updated to support optional in p4c and/or simple_switch, please let me know.  I have spent a few hours today hacking around in p4c to see what changes I think would be needed, and can publish at least a partial patch/PR separately from this one if that sounds useful.

jafingerhut:
Related PR for behavioral-model: https://github.com/p4lang/behavioral-model/issues/844

jafingerhut:
Here are at least some of the kinds of changes I would expect to be made to p4c in order to add support for match_kind optional to p4c: https://github.com/p4lang/p4c/pull/2146

As mentioned in the a comment there, they require changes to p4lang/behavioral-model (maybe), and for p4c's control-plane/p4runtime directory / git-submodule to be updated to a version that includes the additional of the optional match_kind.

stefanheule:
Thank you Andy, it is excellent to see that you already did most of the work required to make this happen. I really appreciate it!

jafingerhut:
You are welcome for the part I have done (it scratched a curiosity itch I had about the changes required), but I suspect it might be small compared to the PI changes required, and am happy to let someone else figure that part out :-)

mbudiu-vmw:
I think we can close this

PR #2142, opened 2020-01-18
simplify slices

hanw:
Simplify `n[7:4][3:0][3:0] = 4w0;` to `n[7:4] = 4w0` and `m[7:0] = 8w0` to `m = 8w0` if m is 8bit wide.


jafingerhut:
Are there any existing test P4_16 programs that contain what I would guess to be expressions that should be errors, e.g. `n[7:4][5:3]` ?  If not, seems worth adding one with a few such bad expressions.

hanw:
I added a negative test and a few more positive tests.

PR #2145, opened 2020-01-18
Add test program that tests catching of incorrect lpm const entries

jafingerhut:
From looking at the C++ code, I thought there might be some cases that
were not caught, but it seems that all of them get at least a warning,
if not an error, in some cases in the target-independent code, others
in the P4Runtime back end.

### DIFF
--- a/testdata/p4_16_errors/table-entries-lpm-2.p4
+++ b/testdata/p4_16_errors/table-entries-lpm-2.p4
@@ -1,0 +1,92 @@
+/*
+Copyright 2020 Cisco Systems, Inc.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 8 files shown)


mbudiu-vmw:
Are you happy with these error messages?
Do you want the keys to be 8 bits? This triggers all these warnings about values too wide.

jafingerhut:
This test program is intended specifically to exercise p4c's ability to detect bad values and/or masks in const entries, because I thought there might be a bug in catching certain kinds of bad ones, but from my testing it appears there is no such bug.  So that is good.

Is it true that there is an option on p4c's command line to turn all, or a subset of particular kinds of, warnings into errors?  It seems that there might be some who would wish a warning like this:
```
table-entries-lpm-2.p4(68): [--Wwarn=mismatch] warning: 0x100: value does not fit in 8 bits
             0x100 &&& 0xF0 : a_with_control_params(11);
             ^^^^^
```
to cause a program to fail to compile.

jafingerhut:
In response to my question of whether I am happy with all of the warnings/errors, they all look correct to me, and I was not able to write anything that looked like a wrong value/mask for LPM fields that the compiler did not give at least a warning about, so that is good.

PR #2146, opened 2020-01-18
Add match_kind optional to v1model architecture and bmv2 back end

jafingerhut:
This is at least a partial addition of match_kind optional to the v1model architecture and the bmv2 back end of p4c.

As written here, getting the STF test to pass would require changes to the behavioral-model repository so that it accepts match_type "optional" in addition to the others currently supported.  See https://github.com/p4lang/behavioral-model/issues/844 asking whether that is even the desired approach for changes to behavioral-model at all, and if yes, see https://github.com/p4lang/behavioral-model/pull/845 for only the beginnings of the changes required for that.

They also require that the git submodule control-plane/p4runtime be updated to a version of the p4lang/p4runtime repository that contains the addition of the optional match_kind, so that P4Info files can be generated with the new match_kind of "optional".


jafingerhut:
Warning to reviewers, or those considering merging: This code does not compile unless the control-plane/p4runtime git submodule version is updated to a version that adds match_kind "optional" to the list of the match_kinds allowed in a P4Runtime P4Info file, which p4c can optionally generate while compiling a P4 program.

This PR, or something similar: https://github.com/p4lang/p4runtime/pull/259

must be merged into p4lang/p4runtime, and this p4c PR updated to point at that committed version, before this PR can possibly be ready to merge.

control-plane/p4RuntimeSerializer.cpp
```diff
@@ -1636,6 +1641,49 @@ class P4RuntimeEntriesConverter {
...
+            valueStr = stringReprConstant(*value, keyWidth);
+        } else {
+            valueStr = convertSimpleKeyExpression(k, keyWidth, typeMap);
+        }
+        if (valueStr == boost::none) return;
+        auto protoMatch = protoEntry->add_match();
+        protoMatch->set_field_id(fieldId);
+        // TODO(jafingerhut): Should there be a mutable_optional()
+        // method, or is using mutable_exact() OK?
+        auto protoOptional = protoMatch->mutable_exact();
```
antoninbas: There should already be a `mutable_optional` with Stefan's CL: https://github.com/p4lang/p4runtime/pull/259/files#diff-ddc87b5c18cbea0bd61c6774edbd07c9R236

backends/bmv2/common/control.cpp
```diff
@@ -100,6 +107,42 @@ void ControlConverter::convertTableEntries(const IR::P4Table *table,
...
                 }
+            } else if (matchType == "optional") {
+                // Table key fields with match_kind optional with
+                // "const entries" in the P4 source code will be
+                // represented using the same "key" and "mask" keys in
+                // the BMv2 JSON file as table key fields with
+                // match_kind ternary.  We will perform here the more
+                // strict checking of the allowed masks that optional
+                // imposes.
+                if (k->is<IR::Mask>()) {
```
antoninbas: do you think we should even allow mask expressions in the P4? We already have `_` for the don't care case. We would no longer need this code then.

jafingerhut: That makes a lot of sense.  Good review comment!  I will change the code not to use masks for const entries on optional fields at all, but only exactly value or `_`/`default`

control-plane/p4RuntimeSerializer.cpp
jafingerhut: I will try that out.  Thanks.

backends/bmv2/common/control.cpp
jafingerhut: @antoninbas If you can stand to take another look at this PR, I think I have addressed your two comments.

testdata/p4_16_errors_outputs/table-entries-optional-2-bmv2.p4-stderr
```diff
@@ -0,0 +1,12 @@
+table-entries-optional-2-bmv2.p4(71): [--Wwarn=mismatch] warning: 0x100: value does not fit in 8 bits
+            (0x100, default): a_with_control_params(2);
+             ^^^^^
+table-entries-optional-2-bmv2.p4(73): [--Wwarn=mismatch] warning: 0x10000: value does not fit in 16 bits
+            (default, 0x10000): a_with_control_params(3);
+                      ^^^^^^^
+table-entries-optional-2-bmv2.p4(69): [--Werror=legacy] error: &&& invalid key expression
```
antoninbas: are the output files up-to-date? I would have expected this error message: https://github.com/p4lang/p4c/pull/2146/files#diff-7d9ca24e25fbe2b12ec7df5a3e6f2bc3R125

jafingerhut: That is weird.  The Travis tests all pass, which suggests the output files are up to date, if they are being checked by the test scripts properly.  But the line numbers look off, so not sure why the tests are passing.  I will try regenerating the output files on my machine and see if anything changes.

jafingerhut: Double-checked, and I was wrong about the line numbers -- the line numbers in these warning and error messages do match what they should be from the P4_16 source file.

Also regenerated the output files from the code in this PR, and it matches what you see here.  I haven't dug hard to see why this error message is given instead of the one you linked to.

I guess the question is: is it worth digging more?  I was just happy to get an error message at all for that line :-)

antoninbas: I just checked and this is because the error is given by the P4Runtime serializer and not the bmv2 backend (since the test is run with `--p4runtime-files`). We could make the error more helpful, but that can be a future improvement. After all, the same error would be output if you were to use a `Mask` expression for an exact match.

antoninbas: So, no changes required :)

mbudiu-vmw:
@stefanheule: do you have any comments?

mbudiu-vmw:
Do we need to change bmv2? Can't we just compile this into `ternary`?

mbudiu-vmw:
Apparently you do compile them into `ternary`. So I guess the original comment is no longer accurate.

control-plane/p4RuntimeSerializer.cpp
```diff
@@ -1636,6 +1641,19 @@ class P4RuntimeEntriesConverter {
...
+    void addOptional(p4v1::TableEntry* protoEntry, int fieldId,
+                     const IR::Expression* k, int keyWidth,
+                     TypeMap* typeMap) const {
+        if (k->is<IR::DefaultExpression>())  // don't care, skip in P4Runtime message
+            return;
+        auto protoMatch = protoEntry->add_match();
+        protoMatch->set_field_id(fieldId);
+        auto protoOptional = protoMatch->mutable_optional();
+        auto value = convertSimpleKeyExpression(k, keyWidth, typeMap);
+        if (value == boost::none) return;
```
stefanheule: I'm not super familiar with this code, but I think this check needs to happen before creating the match (`protoEntry->add_match()`). If value is none, then the match is a wildcard, and should not appear in the P4RT message.

jafingerhut:
mbudiu-vmw wrote: "Apparently you do compile them into ternary. So I guess the original comment is no longer accurate."

Yes, that is confusing for readers of this PR.  My plan for how to represent these new match kind `optional` fields changed to "just use `ternary` for all of them in the BMv2 JSON file" after I started working on this PR, based on feedback from Antonin in other issues on the behavioral-model repo.

control-plane/p4RuntimeSerializer.cpp
jafingerhut: I confess that this part of the code I patterned after the `addExact` method elsewhere in the same file, with the primary difference being that `addOptional` checks for the use of a DefaultExpression, but addExact does not (and should not).

@antoninbas Is it a bug if `addExact` is doing this kind of a check and `return`?

antoninbas: The code as it is now is correct.

The check at the beginning of the function skips wildcard matches:
```
if (k->is<IR::DefaultExpression>())
    return;
```

This check:
```
if (value == boost::none) return;
```
just skips the field if there is an error in `convertSimpleKeyExpression`.

jafingerhut: And most likely, if the code reaches this point after many other sanity checks have been done, there might not be any cases left where `convertSimpleKeyExpression` has an error?  At least, I would guess it would be difficult to find a scenario where it does.

antoninbas: actually your very own example hits an error in `convertSimpleKeyExpression`: https://github.com/p4lang/p4c/pull/2146/files#diff-a414f2e1b347d4c3f183729cb058e3c1R7

this leads to an empty `optional` match in the P4Runtime entries file: https://github.com/p4lang/p4c/pull/2146/files#diff-f4f53336a8d660ad338f787608f89c6cR8

IMO this is the right thing to do here: we generate an invalid file in case of error rather than treating it the same as a wildcard.

But I think Stefan was just confused as to which check was responsible for handling wildcards.

stefanheule: Thanks for all the comments, I was indeed just confused (which is why I said I don't know the code very well) :)

testdata/p4_16_errors_outputs/table-entries-optional-2-bmv2-frontend.p4
```diff
@@ -0,0 +1,80 @@
...
+        }
+        actions = {
+            a();
+            a_with_control_params();
+        }
+        default_action = a();
+        const entries = {
+                        (8w0xaa &&& 8w0xff, 16w0x1111 &&& 16w0xffff) : a_with_control_params(9w1);
+
+                        (8w0x0, default) : a_with_control_params(9w2);
```
stefanheule: What does default mean in this context? I've looked through the P4 spec, but can't find a reference to it. (Likely there is nothing wrong here, I just wanted to understand).

jafingerhut: See section 8.12.2 "The universal set" in the latest P4_16 language spec.  You can use `default` or `_` in this context to mean "match anything", the same as 0 &&& 0 would mean for a ternary field.

stefanheule: Ah, thank you, not sure how I missed that.

mbudiu-vmw:
Anyone objects to merging this?

jafingerhut:
@mbudiu-vmw @antoninbas Maybe I misunderstand, but I believe these changes depend upon at least two things happening first:

+ This PR, or a similar one, are merged in for the p4lang/p4runtime repo: https://github.com/p4lang/p4runtime/pull/259
+ This PR is updated so that the git submodule version of p4runtime used by p4c points at that new commit version of p4lang/p4runtime.

Right now I think this is only passing Travis tests because I have committed a git submodule update of the p4lang/p4runtime version to some commit on a PR branch of that repo, not of its master branch.  But I don't necessarily understand how git submodules work, either. :-)

antoninbas:
We should be able to merge the p4runtime PR in the upcoming week I hope. We first need to release P4Runtime v1.1.0, and for that I need to close this issue (https://github.com/p4lang/p4runtime/issues/258): I'm planning on opening a PR for it ahead of the Wednesday P4 API WG meeting.

jafingerhut:
Closing this PR, since I created this new updated one to replace it: https://github.com/p4lang/p4c/pull/2184

PR #2149, opened 2020-01-22
visitDagOnce = false for control flow simplifications; fixes #2148

### LINKED ISSUE #2148: p4c-bm2-ss crashes on undefined header conditional in method
This is a minor, very specific crash, which we have encountered. This program is nonsensical and returns a value based on a conditional that compares against an undefined variable.

When compiling the program with p4c-bm2-ss,  the compiler crashes at the ValidateTableProperties pass with a segmentation fault. Changing any of the code leads to success. Also just using ./build/p4c instead works. 

[undefined_header_cond_ValidateTableProperties.txt](https://github.com/p4lang/p4c/files/4098258/undefined_header_cond_ValidateTableProperties.txt)

[undefined_header_cond.txt](https://github.com/p4lang/p4c/files/4098240/undefined_header_cond.txt)

mbudiu-vmw:
This was happening because an EmptyStatement is replaced with nullptr in a block, but the same EmptyStatement was a child of an IfStatement. The solution is `visitDagOnce = false` for this pass.

### DIFF
--- a/frontends/p4/simplify.cpp
+++ b/frontends/p4/simplify.cpp
@@ -20,7 +20,7 @@ limitations under the License.
 namespace P4 {
 
 const IR::Node* DoSimplifyControlFlow::postorder(IR::BlockStatement* statement) {
-    LOG1("Visiting " << dbp(getOriginal()));
+    LOG3("Visiting " << dbp(getOriginal()));
     if (statement->annotations->size() > 0)
         return statement;
     auto parent = getContext()->node;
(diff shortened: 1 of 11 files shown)


mbudiu-vmw:
Thank you @rst0git, but it looks like more reviews are needed to merge this.

mbudiu-vmw:
This is a one-line fix, it should be easy to review. There's another PR of a similar complexity awaiting.

PR #2154, opened 2020-01-24
Update leftValues when lowering; fixed #2150

### LINKED ISSUE #2150: Why a different error message for some 'complex' assignment left hand sides?
While developing test programs for the PR that aims to add support for run-time variable indexes to header stacks https://github.com/p4lang/p4c/pull/2139 we came across a kind of odd difference in some error messages that seem to indicate that some run-time variable index assignments are treated quite a bit differently in the compiler internals, and we were wondering if this is a sign of a limitation of the current compiler internals, and/or imply that adding this feature fully generally will require different kinds of changes than have been included in that PR so far.

For example, in the attached test program, an assignment like this:
```
hdr.vector[hdr.ml.idx1].e =
            hdr.ethernet.dstAddr[47:40];
```
causes the latest p4c code as of 2020-Jan-22 using the command `p4c filename.p4` to give an error message "hdr.ml.idx1: Invalid all array indexes must be constant".  The only place in the p4c code that contains the string "all array indexes must be const" is the file p4c/backends/bmv2/common/expression.cpp, so that error message is issued from the bmv2 back end code.

The above is expected.  No problem there.

However, if you make the left hand side more complex, like this:
```
        hdr.vector[hdr.vector[hdr.ethernet.dstAddr[2:0]].e].e =
            hdr.ethernet.dstAddr[47:40];
```
and run `p4c filename.p4`, you get a different error message:
```
runtime-index-4-bmv2.p4(121): [--Werror=type-error] error: Expression [].e cannot be the target of an assignment
        hdr.vector[hdr.vector[hdr.ethernet.dstAddr[2:0]].e].e =
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```
The only error message containing the string "cannot be the target of an assign" occurs in file: p4c/frontends/p4/typeChecking/typeChecker.cpp, so this message is caught in a completely different way, and in a part of the code where there are existing test cases for doing things like attempting to assign a value to an `in` parameter, or to the name of an extern object, or a few other kinds of things that are completely illegal in the language, whereas I thought the expression above is legal.

The same expression gives no error or warning message at all when you run `p4test filename.p4` instead of `p4c filename.p4`.

Any idea what the difference is here?  Should it be giving a different error message, or should it be the same "all array indexes must be constant" even for the more complex assignment?

[runtime-index-4-bmv2.p4.txt](https://github.com/p4lang/p4c/files/4103676/runtime-index-4-bmv2.p4.txt)

mbudiu-vmw:
There is no test associated with this file, because the PR #2139 contains a test.

### DIFF
--- a/backends/bmv2/common/lower.cpp
+++ b/backends/bmv2/common/lower.cpp
@@ -83,8 +83,13 @@ const IR::Node* LowerExpressions::postorder(IR::Cast* expression) {
 
 const IR::Node* LowerExpressions::postorder(IR::Expression* expression) {
     // Just update the typeMap incrementally.
-    auto type = typeMap->getType(getOriginal(), true);
+    auto orig = getOriginal<IR::Expression>();
+    auto type = typeMap->getType(orig, true);
     typeMap->setType(expression, type);
+    if (typeMap->isCompileTimeConstant(orig))
+        typeMap->setCompileTimeConstant(expression);
+    if (typeMap->isLeftValue(orig))
+        typeMap->setLeftValue(expression);
     return expression;
 }
 


hesingh:
I have tested this fix with my PR  (https://github.com/p4lang/p4c/pull/2139) and the fix works.

With the fix it's good to see a runtime index expression being added to the leftValues std::set in p4c.

hesingh:
Code changes look good to me.  If the orig expression is a leftValue, then the lowered expression is added to the leftValues std::set.

hesingh:
I poked around p4c code to see if any other pass needs to set left value like what this PR is doing.  I am not sure.  However, I noticed code duplication.  The code changes of this PR also exist here.

https://github.com/p4lang/p4c/blob/master/frontends/p4/sideEffects.cpp#L51

Why not stick the code in a new function and call the function from two places?   Since the current users of the new function are a midend and backend pass, I'd stick the new function is some code in p4c/ir.

hesingh:
Here are diffs for a new common function.  

https://github.com/hesingh/p4-code/blob/master/diffs-func-setLeftVal

For some reason I could not get `auto orig = getOriginal<IR::Expression>();` to compile as a line of code in typeMap.cpp and this is why this line is still in lower.cpp and sideEffects.cpp.

hesingh:
I saw that `getOriginal` is defined in `visitor.h` and derived the typeMap class also from Inspector and then compiling passes.  

https://github.com/hesingh/p4-code/blob/master/diffs-func-setLeftVal#L65

Maybe, one does not allow multiple inheritance for a class like this in p4c, but if we do, the final code  has been checked at the above URL.

mbudiu-vmw:
This one is a 5-line fix, should also be easy to review.

PR #2159, opened 2020-01-28
Exit makes statements unreachable; fixes #2151

### LINKED ISSUE #2151: Follow-up to issue 2104, exit also kills all variables
A follow up to a previous bug. In issue #2104  return statements killed all subsequent variables. This also happens with exit statements, we missed to cover these.
[2151.txt](https://github.com/p4lang/p4c/files/4109385/2151.txt)

mbudiu-vmw:


frontends/p4/simplifyDefUse.cpp
```diff
@@ -177,16 +177,16 @@ class FindUninitialized : public Inspector {
     }
 
     bool preorder(const IR::Function* func) override {
-        LOG3("FU Visiting function " << func->name << "[" << func->id << "]");
+        if (findContext<IR::P4Program>() != nullptr) {
+            LOG3("FU skip function " << func->name << "[" << func->id << "]");
```
ChrisDodd: How can this ever be true?  This pass is only ever called via DoSimplifyDefUse::process, which is only ever called directly on IR::Function, IR::P4Parser, and IR::P4Control nodes and never on full programs...

frontends/p4/simplifyDefUse.cpp
```diff
@@ -378,7 +385,13 @@ class FindUninitialized : public Inspector {
     }
 
     bool preorder(const IR::P4Action* action) override {
-        LOG3("FU Visiting " << action);
+        if (findContext<IR::P4Program>() != nullptr) {
+            // We want to visit an action only if invoked by a new
```
ChrisDodd: same here -- there's no way there could ever be a P4Program in the context.

mbudiu-vmw:
Changed the checks with assertions.
But there was an issue with virtual methods too, which I hopefully fixed.

PR #2162, opened 2020-01-28
Enable Slice test with runtime index

hesingh:
After checkin of https://github.com/p4lang/p4c/commit/b281cddc79de75f048a2f3c663ebce5b732d2c63, this PR enables Slice and runtime index test line of code.

This PR has also changed p4RuntimeSerializer.cpp to use ErrorType so that when the error occurs, the line number of P4 code causing the error is also printed. 

I will file a separate issue against p4RuntimeSerializer.cpp because many of its errors do not print the line of P4 code where the error occurs.  The code should change.

### DIFF
--- a/control-plane/p4RuntimeSerializer.cpp
+++ b/control-plane/p4RuntimeSerializer.cpp
@@ -1623,9 +1623,8 @@ class P4RuntimeEntriesConverter {
             // earlier.
             // For e.g. 16 bit key has a max value of 65535, Range of (1..65536)
             // will be converted to (1..0) and will fail below check.
-	    if (*start > *end)
-                ::error(ErrorType::ERR_INVALID, "%2% Invalid range for table entry",
-                        k, kr->srcInfo);
+            if (*start > *end)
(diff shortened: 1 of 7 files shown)


control-plane/p4RuntimeSerializer.cpp
```diff
@@ -1623,8 +1623,9 @@ class P4RuntimeEntriesConverter {
             // earlier.
             // For e.g. 16 bit key has a max value of 65535, Range of (1..65536)
             // will be converted to (1..0) and will fail below check.
-            if (*start > *end)
-                ::error("%s Invalid range for table entry", kr->srcInfo);
+	    if (*start > *end)
```
mbudiu-vmw: He said that he didn't like the error message produced.

hesingh: I restored the original code and checked in.

hesingh:
The workers are slow at github today.   Three Travis tests have passed proving the code changes are good.

Status is here: https://travis-ci.org/p4lang/p4c/builds/643160656?utm_source=github_status&utm_medium=notification

PR #2164, opened 2020-01-29
tools-driver: Move shebang on first line

rst0git:
A script will be executed using the interpreter specified on the first line.


cc10512:
main.py in this case is a misnomer and that file is not invoked on the command line. The p4c driver file is https://github.com/p4lang/p4c/blob/master/tools/driver/p4c.in, and it will itself invoke main.py after setting the env. So you can actually remove that line altogether.

rst0git:
@cc10512 thank you for the code review, I have updated the PR to remove the shebang line.

Issue #2153, opened 2020-01-23
Switchstatement: Assignment in switch-case removed


fruffy:
Hello again

I have a question about the semantics of switch statements. During code generation and validation we have encountered a pass which removes the assignment to a variable, which is later used for a condition. The example program is attached.
The pass in question is `SimpleSwitchMidEnd_29_LocalCopyPropagation`. It changes the program from

    control ingress(inout Headers h, inout Meta m, inout standard_metadata_t sm) {
        bit<8> tmp_condition_0;
        @name(".NoAction") action NoAction_0() {
        }
        @name("ingress.do_something") action do_something() {
            h.h.a = 8w7;
        }
        @name("ingress.simple_table") table simple_table {
            key = {
                h.h.b: exact @name("h.h.b") ;
            }
            actions = {
                NoAction_0();
                do_something();
            }
            default_action = do_something();
        }
        apply {
            tmp_condition_0 = 8w0;
            switch (simple_table.apply().action_run) {
                NoAction_0: {
                    tmp_condition_0 = 8w1;
                }
            }

            if (tmp_condition_0 > 8w0) {
                sm.egress_spec = 9w1;
            }
        }
    }

to

    control ingress(inout Headers h, inout Meta m, inout standard_metadata_t sm) {
        @name(".NoAction") action NoAction_0() {
        }
        @name("ingress.do_something") action do_something() {
            h.h.a = 8w7;
        }
        @name("ingress.simple_table") table simple_table {
            key = {
                h.h.b: exact @name("h.h.b") ;
            }
            actions = {
                NoAction_0();
                do_something();
            }
            default_action = do_something();
        }
        apply {
            switch (simple_table.apply().action_run) {
                NoAction_0: {
                }
            }

            if (8w1 > 8w0) {
                sm.egress_spec = 9w1;
            }
        }
    }

The variable `tmp_condition_0` is removed because it is assumed that `NoAction` is always executed. Is this correct? Our assumption was this expression either does nothing or, if `NoAction` has previously been executed, changes the value of `tmp_condition_0`. The spec says

> At runtime, if no case matches, execution of the program simply continues. However, no label can appear twice in a switch statement. 

which would imply that it might happen that `tmp_condition_0 = 8w1;` is not executed.

[switch_statement.txt](https://github.com/p4lang/p4c/files/4105950/switch_statement.txt)

[2153.p4.txt](https://github.com/p4lang/p4c/files/4109437/2153.p4.txt)
[...]

PR #2168, opened 2020-01-30
new SwitchAddDefault pass to add default: {} to switch

ChrisDodd:
- only adds if it does not already have a default and not all cases can
  be shown to be handled.

Fix for #2153


mbudiu-vmw:
Why not even in the front-end?

ChrisDodd:
I guess it would be simpler to do it in the frontend -- fewer changes...

ChrisDodd:
Moved to the frontend, so now it is only in one place instead of in all the backends.

PR #2171, opened 2020-01-31
Fix uninitialized 'unreachable' field in FindUninitialized

ChrisDodd:


jafingerhut:
I have written many, many bugs in my life, so please take this comment in the spirit it is offered, from one developer to another.  Sometimes I laugh out loud at my desk when I discover the root cause of a bug in my code.  Probably a bit wacky that way.

Does anyone else find it amusing that there is a bug in the compiler code for detecting uninitialized variables in P4 programs, that is itself an uninitialized variable?

hesingh:
Where is the unit test P4 program for this issue?  An existing P4 program in testdata/p4_16_samples/?  If so, please point us to it.  thanks.

jafingerhut:
There might not be a test program, because for uninitialized variables, the behavior is non-deterministic depending upon the rest of the code/control flow leading up to the use of the uninitialized variable.

Issue #2163, opened 2020-01-29
Compiling a v1model program using p4c and value_set gives compiler bug error message


jafingerhut:
I have not gone to earlier versions of p4c to see how far back this behavior goes, but least with the latest version of p4c source code as of 2020-Jan-28, compiling any of these programs in the testdata/p4_16_samples directory using the command `p4c file.p4` or `p4c --target bmv2 --arch v1model file.p4`:

+ pvs-struct.p4
+ pvs-nested-struct.p4
+ v1model-p4runtime-most-types1.p4

gives a compiler bug error:
```
$ p4c pvs-struct.p4
In file: /home/andy/forks/p4c/ir/type.def:365
Compiler Bug: Type_Name is not a canonical type, use getTypeType()?
```

None of them give any errors when compiled using the `p4test file.p4` command.

I think the last two test programs above were created to test out the edge cases of supported types, and may not be intended to compile without any errors, but a compiler bug doesn't seem appropriate.

The attached program pvs-2-bmv2.p4 is a small modification of the existing test program pvs-struct.p4, except that it compiles without error using `p4c` and `p4c --target bmv2 --arch v1model`.  Something like it should perhaps be added as another test program, and might be useful in comparing the compiler's behavior against the other test programs when investigating why only the ones above give a compiler bug.

[pvs-2-bmv2.p4.txt](https://github.com/p4lang/p4c/files/4126118/pvs-2-bmv2.p4.txt)

jafingerhut:
It seems to me like the program pvs-struct.p4 _should_ be supported and compile to working BMv2 JSON.  It is pretty much the same as pvs-2-bmv2.p4, except that the type on which the value_set is created is a struct with one bit<32> field, instead of pvs-2-bmv2.p4 which has a bit<32> field directly.

If pvs-struct.p4 can be made to work, I can try a test program that would address issue https://github.com/p4lang/p4c/issues/2143

hesingh:
@mbudiu-vmw If you didn't get a chance to look into this one, I have found a fix.   I will issue a PR shortly.

PR #2172, opened 2020-01-31
Fix 2163

hesingh:
Fixes 

https://github.com/p4lang/p4c/issues/2163


hesingh:
Other diffs are included which is not correct.  I will issue another PR.

Issue #2143, opened 2020-01-18
No test cases for @match annotation in p4c test suite


jafingerhut:
At least, none I could find by grep'ing for "@match" in the testdata/p4_16_samples directory.

Search for "@match" in the P4_16 language specification for a description of it.

I suspect that at least the P4Info file generated from a program with a "@match(ternary)" annotation on some field of a struct used in a P4_16 value_set should be different than the same program without such an annotation, so a test case that someone hand-verified the contents of the P4Info file was correct, and checked in, would be better than no test case at all.

jafingerhut:
Closing this issue, as there is now one such test program, and the automated tests should check that the P4Info file is generated with the correct match kinds.

The test program is named: testdata/p4_16_samples/pvs-struct-3-bmv2.p4
added in this commit: https://github.com/p4lang/p4c/commit/9a2abd7ead8f7f71f1a5387aa299fe5f81fc6c32

PR #2173, opened 2020-01-31
Fix 2163

hesingh:
Fixes 

https://github.com/p4lang/p4c/issues/2163

### DIFF
--- a/backends/bmv2/common/parser.cpp
+++ b/backends/bmv2/common/parser.cpp
@@ -443,8 +443,7 @@ bool ParserConverter::preorder(const IR::P4Parser* parser) {
 
     for (auto s : parser->parserLocals) {
         if (auto inst = s->to<IR::P4ValueSet>()) {
-            auto type = ctxt->typeMap->getTypeType(inst->elementType, true);
-            auto bitwidth = type->width_bits();
+            auto bitwidth = inst->elementType->width_bits();
             auto name = inst->controlPlaneName();
             ctxt->json->add_parse_vset(name, bitwidth);
         }
--- a/backends/bmv2/common/parser.cpp
+++ b/backends/bmv2/common/parser.cpp
@@ -443,7 +443,8 @@ bool ParserConverter::preorder(const IR::P4Parser* parser) {
 
     for (auto s : parser->parserLocals) {
         if (auto inst = s->to<IR::P4ValueSet>()) {
-            auto bitwidth = inst->elementType->width_bits();
+            auto type = ctxt->typeMap->getTypeType(inst->elementType, true);
+            auto bitwidth = type->width_bits();
             auto name = inst->controlPlaneName();
             ctxt->json->add_parse_vset(name, bitwidth);
         }
--- a/frontends/p4/typeChecking/typeChecker.cpp
+++ b/frontends/p4/typeChecking/typeChecker.cpp
@@ -1262,7 +1262,6 @@ const IR::Node* TypeInference::postorder(IR::P4ValueSet* decl) {
             (void)canon->apply(*tc);
             if (::errorCount() > e)
                 return nullptr;
-            decl->elementType = getTypeType(canon);
         }
         auto tt = new IR::Type_Set(canon);
         setType(getOriginal(), tt);
--- a/frontends/p4/typeChecking/typeChecker.cpp
+++ b/frontends/p4/typeChecking/typeChecker.cpp
@@ -1262,6 +1262,7 @@ const IR::Node* TypeInference::postorder(IR::P4ValueSet* decl) {
             (void)canon->apply(*tc);
             if (::errorCount() > e)
                 return nullptr;
+            decl->elementType = getTypeType(canon);
         }
         auto tt = new IR::Type_Set(canon);
         setType(getOriginal(), tt);


hesingh:
There is no test included because the issue does not happen with `p4test` which is used in testing.  The issue happens when using `p4c`.  `p4c` failure occurs with P4 programs which already exist in the p4c repo.  The test programs are listed in the issue.

jafingerhut:
I have not tried to understand the code changes involved, and whether or not they are a good solution to this problem.

I have tried applying this patch on top of latest master p4c source code, and run the resulting p4c on the test programs mentioned in the issue, and there is no more compiler bug.  I checked more carefully the output of the last midend pass, the BMv2 JSON file, and the P4Info file, for one of the existing test programs, pvs-struct.p4, and they all look correct.

I have not tried running simple_switch and configuring a parser value set and running packets through to try it, but may do so within a week or so.

I do not think the existing STF test infra enables creating such an automated STF test, because I don't think it supports creating entries in a parser value_set.  I am also not sure it could be done without changes to behavioral-model, as there do not appear to be simple_switch_CLI commands for adding entries to value sets.

jafingerhut:
I am happy to suggest additional P4 test programs, that could be added either in this PR, or a later one, that would check the output of the midend, and the contents of the generated P4Info file against an expected one.

hesingh:
Thanks for testing @jafingerhut.  Your plan also sounds good. 

The bug fix is in the bmv2 backend code, so one does not have to test any frontend or midend p4c code.   

Also, having checked the `git log` for p4lang/behavioral-model` , I see parse value sets is supported via this checkin: https://github.com/p4lang/behavioral-model/commit/0bc948ed5b6b43c8c5416033916f24206e5974a6#diff-418314966c57fea403ba29bf6b0793c7  The bmv2 backend does seem to have some support for parse value sets in JSON.  Look for ``` "type" : "parse_vset" ``` in JSON.  I don't know about STF.

I see that the error also happen with `p4c-bm2-ss`.  See log below.

```
hemant@ubuntu:~/p4c/build$ ./p4c-bm2-ss --std p4-16 ~/int/p4c/testdata/p4_16_samples/pvs-struct.p4
In file: /home/hemant/p4c/ir/type.def:365
Compiler Bug: Type_Name is not a canonical type, use getTypeType()?

hemant@ubuntu:~/p4c/build$ 
```

I think we could checkin code from this PR after code review.  Later we can add as many tests as we'd like.  I'd be happy to support p4c for those tests with any issue.

hesingh:
@jafingerhut Regarding

"I can try a test program that would address issue #2143"

Please see what you can do to test 2143 - thanks.

jafingerhut:
I have tested issue #2143, and it also works with the changes in this PR.  Perhaps it might have worked using p4test rather than p4c even before -- I haven't tried that.

hesingh:
p4test is not a comprehensive test compared to p4c or p4c-bm2-ss.   p4test processes the P4 program only up till midend.  In contrast p4c and p4c-bm2-ss run the bmv2 backend on the P4 program.

hesingh:
I had to make a change to p4lang/behavioral-model as shown below before a parse value set works with attached P4 and STF file. 

```
diff --git a/include/bm/bm_sim/bytecontainer.h b/include/bm/bm_sim/bytecontainer.h
index 9dcdea7..8ad7ee4 100644
--- a/include/bm/bm_sim/bytecontainer.h
+++ b/include/bm/bm_sim/bytecontainer.h
@@ -28,6 +28,7 @@
 #include <vector>
 #include <iterator>
 #include <string>
+#include <iostream>
 
 #include "short_alloc.h"
 
@@ -191,7 +192,7 @@ class ByteContainer {
 
   //! @copydoc operator[]
   const_reference operator[](size_type n) const {
-    assert(n < size());
+    std::cout << "op: " << n << " " << size() << std::endl;
     return bytes[n];
   }
```

Test log:  In the log see that the 0-4 values of parser set is `n`

```
hemant@ubuntu:~/p4c/build$ bmv2/testdata/p4_16_samples/pvs-struct-bmv2.p4.test -v
Writing temporary files into  ./tmp1zi98ni6
Executing  ./p4c-bm2-ss -o bmv2/testdata/p4_16_samples/pvs-struct-bmv2.json /home/hemant/p4c/testdata/p4_16_samples/pvs-struct-bmv2.p4
Exit code  0
Check for  /home/hemant/p4c/testdata/p4_16_samples/pvs-struct-bmv2.stf
Running model
Running /home/hemant/behavioral-model/targets/simple_switch/simple_switch --log-file switch.log --log-flush --use-files 0 --thrift-port 9771 --device-id 681 -i 0@pcap0 ../bmv2/testdata/p4_16_samples/pvs-struct-bmv2.json
Adding interface pcap0 as port 0 (files)
Running /home/hemant/behavioral-model/targets/simple_switch/simple_switch_CLI --thrift-port 9771
STF Command: packet 0 01020304 05060708
STF Command: expect 0 01020304 05060708 $
op: 0 1
op: 1 1
op: 2 1
op: 3 1
Obtaining JSON from switch...
Done
RuntimeCmd: 
simple_switch exit code -15
Execution completed
Comparing outputs
Removing ./tmp1zi98ni6
```
[pvs-struct-bmv2-stf.txt](https://github.com/p4lang/p4c/files/4142089/pvs-struct-bmv2-stf.txt)
[pvs-struct-bmv2-p4.txt](https://github.com/p4lang/p4c/files/4142090/pvs-struct-bmv2-p4.txt)

hesingh:
The reason this PR fixes the issue is because `value_set<value_set_t>(4) pvs;` parse value set is using a struct `value_set_t` vs. `bit<32>`.  So the backend receives the inst->elementType as `Type_Name` and this is why getTypeType() API is needed to fix the code.

backends/bmv2/common/parser.cpp
```diff
@@ -443,7 +443,8 @@ bool ParserConverter::preorder(const IR::P4Parser* parser) {
 
     for (auto s : parser->parserLocals) {
         if (auto inst = s->to<IR::P4ValueSet>()) {
-            auto bitwidth = inst->elementType->width_bits();
+            auto type = ctxt->typeMap->getTypeType(inst->elementType, true);
```
cc10512: I think it is better to ensure that elementType expression has the proper type assigned.

hesingh: I did look at the typeChecker code which canonicalizes elementType as shown below.

https://github.com/p4lang/p4c/blob/master/frontends/p4/typeChecking/typeChecker.cpp#L1256

The output from typeChecker has elementType as Type_Name which looks correct to me.   The backend needs  to use getTypeType() if type is Type_Name.

hesingh: I debugged `const IR::Node* TypeInference::postorder(IR::P4ValueSet* decl)` and see an issue.  The `decl` coming in and being returned does not change `decl->elementType`.  The coming in type for `decl->elementType`  in `Type_Name`.  I changed the frontend code, backed out the bmv2 backend code and the issue is fixed.  I will update the code changes shortly.

hesingh:
@mbudiu-vmw A question related to typeChecker and P4ValueSet when PVS uses a struct instead of `bit<W>`

In this method, at this line of code,

https://github.com/p4lang/p4c/blob/master/frontends/p4/typeChecking/typeChecker.cpp#L1255

decl->elementType is of type Type_Name.

At the end of the method, when decl is returned, the type for decl->elementType is still Type_Name?

When PVS uses a struct, the canonicalization code is run.

hesingh:
> This is incorrect. The typeChecker should never modify an IR node like this. It is only supposed to modify the typeMap. It can only insert new nodes, e.g., casts.

Ok, then the method is incomplete because a new node for P4ValueSet should be created with elementType set to Type_Struct/Type_Bits for `if (canon != decl->elementType)`.

Or we don't change any frontend code and just add two lines of fix to bmv2 backend as shown below.

https://github.com/p4lang/p4c/pull/2173/commits/656ad4a8a6544bec57f9bf8b257fc63428d722b0

backends/bmv2/common/parser.cpp
hesingh: It will be too complicated to change frontend typeChecker for P4ValueSet to set the elementType to a proper type.  The reason is because the postorder(IR::P4ValueSet) method is entered multiple times and additional checks are needed to not create new node with proper element type. 

This is why you may note, Mihai copied the backend changes of this PR and issued another PR to fix the issue in the backend as I proposed here.

PR #2174, opened 2020-02-01
New and updated tests of parser value sets that exercise bmv2 back end

jafingerhut:
At least they partially do.  They run the p4c bmv2 back end and
generate BMv2 JSON files, and ensure that simple_switch can read them,
but they do not send packets through.  Getting that to work appears
like it may require changes to behavioral-model code.


jafingerhut:
At least some of these new tests will fail until this issue is addressed: https://github.com/p4lang/p4c/issues/2163

I have tested that the bmv2 back end code changes in this PR enable these tests to pass, but I do not know the code well enough to otherwise approve the code changes there: https://github.com/p4lang/p4c/pull/2173

hesingh:
I would appreciate it if @mbudiu-vmw could take a look at https://github.com/p4lang/p4c/pull/2173 to expedite its review  - thanks.

antoninbas:
The generated JSON is not correct, at least for the pvs-struct-2-bmv2.p4 program.
Looking at this:
```
          "transitions" : [
            {
              "type" : "parse_vset",
              "value" : "MyParser.pvs",
              "mask" : 0,
              "next_state" : null
            },
            {
              "type" : "hexstr",
              "value" : "0x081001",
              "mask" : null,
              "next_state" : "foo"
            }
          ],
```
The `"mask"` attribute for the first transition is not correct. It should be `null` (or `0x7FFFF` should work as well since entries are 19 bits, but really the mask is not needed here).

hesingh:
This PR only adds tests.  If an issue is found in generated JSON, then don't we have to file a new issues against p4c to fix the JSON?

hesingh:
If I understand correctly that parse_vset should always have a null mask, then this one line of p4c change will fix the issue raised by Antonin.

The "combine" function in parser.cpp does not compute a mask for vset and this is why the mask with 0 value was going to generate the JSON for parse_vset.

```
diff --git a/backends/bmv2/common/parser.cpp b/backends/bmv2/common/parser.cpp
index 4536ac43..7f3e8675 100644
--- a/backends/bmv2/common/parser.cpp
+++ b/backends/bmv2/common/parser.cpp
@@ -390,7 +390,7 @@ ParserConverter::convertSelectExpression(const IR::SelectExpression* expr) {
         if (is_vset) {
             trans->emplace("type", "parse_vset");
             trans->emplace("value", vset_name);
-            trans->emplace("mask", mask);
+            trans->emplace("mask", Util::JsonValue::null);
             trans->emplace("next_state", stateName(sc->state->path->name));
         } else {
             if (mask == 0) {
```

mbudiu-vmw:
I think that only now we have all the required changes merged, we should try this again.

jafingerhut:
Poked the build box to force it to run tests for this PR again.  We will see what happens.

PR #2179, opened 2020-02-04
Use the type from the typeMap; fixes #2163

### LINKED ISSUE #2163: Compiling a v1model program using p4c and value_set gives compiler bug error message

mbudiu-vmw:
The tests are in #2174

### DIFF
--- a/backends/bmv2/common/parser.cpp
+++ b/backends/bmv2/common/parser.cpp
@@ -443,7 +443,8 @@ bool ParserConverter::preorder(const IR::P4Parser* parser) {
 
     for (auto s : parser->parserLocals) {
         if (auto inst = s->to<IR::P4ValueSet>()) {
-            auto bitwidth = inst->elementType->width_bits();
+            auto etype = ctxt->typeMap->getTypeType(inst->elementType, true);
+            auto bitwidth = etype->width_bits();
             auto name = inst->controlPlaneName();
             ctxt->json->add_parse_vset(name, bitwidth);
         }


hesingh:
This was my first fix with https://github.com/p4lang/p4c/pull/2173.

Why did you duplicate the PR when I asked my PR to be reviewed?

hesingh:
I care about bugs getting fixed ASAP.  If anyone approves this PR, go ahead and check it in since you have checkin permission and I don't.

mbudiu-vmw:
I have reviewed your current fix, I cannot review fixes you made in the past.
The current fix is wrong.
I was on vacation this week-end, I cannot review code continuously.

hesingh:
No problem/  These were my first changes in my PR:

https://github.com/p4lang/p4c/pull/2173/commits/656ad4a8a6544bec57f9bf8b257fc63428d722b0

hesingh:
@mbudiu-vmw When you get a chance please reply to this question I asked in my PR today:

It is correct behavior that TypeInference::postorder(IR::P4ValueSet) does not change type for decl->elementType when elementType is struct?  

I didn't get an answer to this question and this is why I was looking at frontend changes.

hesingh:
Could someone please review this PR so that the code can be checked in today.  The changes will unblock Andy's pvs P4 tests and also my follow on work related to pvs.  thanks.

Changes look good to me, but my vote doesn't count.

mbudiu-vmw:
TypeInference never changes any IR node in the graph. It can only insert new nodes in the graph, and only if it is called with a `readOnly=false` flag, which happens only very early in the front-end.

PR #2182, opened 2020-02-04
Quarantine four tests broken by behavioral model

hesingh:
See https://github.com/p4lang/behavioral-model/issues/862.

This way p4c Travis tests will pass and gives time to behavioral model to fix its code.


hesingh:
As expected, Travis has passed.

PR #2183, opened 2020-02-04
Fix mask generation for PVS

mbudiu-vmw:


backends/bmv2/common/parser.cpp
```diff
@@ -390,7 +393,7 @@ ParserConverter::convertSelectExpression(const IR::SelectExpression* expr) {
         if (is_vset) {
             trans->emplace("type", "parse_vset");
             trans->emplace("value", vset_name);
-            trans->emplace("mask", mask);
+            trans->emplace("mask", stringRepr(mask, bytes));
```
antoninbas: For a PVS, you should probably set the mask to `null` unconditionally, given that the cube operator is not supported for PVS lookups in P4_16. To support P4_14 masks, you would probably need some code to distinguish between missing mask and mask set to 0.

If this just extends the mask to the right amount of bytes (e.g. `0` -> `"\x00\x00"`), then the JSON will still be incorrect: the transition will always fail. Maybe having a STF test that actually sends a packet for one of these P4 programs would be a good idea.

hesingh: The code changes are in the bmv2 backend.  The p4-14 to p4-16 translation already happened in the p4c frontend.  How does the backend know  if the mask is to be computed for p4-14 or p4-16?  If we don't know, it's time to extend the IR to carry a translated flag for pvs mask or something like that.

Also, sorry, I don't know what is the cube operator?

jafingerhut: Cube I have heard to refer, I think, to P4's `&&&` operator.  The word 'cube' apparently never appears in the P4 language spec, so not sure the origin of the name, but probably refers to the fact that a value/mask matches all keys in a hypercube of dimension N, where N is the number of wildcard bit positions.

hesingh: I figured as much.  Thanks @jafingerhut.  We all should use consistent terminology.   

A simple fix regarding p4-14 vs. p4-16 masks, would be to augment the `P4ValueSet` in p4c/ir/ir.def.

https://github.com/p4lang/p4c/blob/master/ir/ir.def#L344

Add a mask member and set it during p4-14 to p4-16 translation.

Also, while we are at it, the P4ValueSet is poorly commented code in ir.def.  The `size` could be number of elements in PVS or is it a product of number of elements and size of each element?  I know it is the former, but some folks may not know.   Some comments could also be provided regarding `elementType`.

jafingerhut: I do not have any great suggestions, but note that this p4-14 vs p4-16 mask thing appears to be a case where there might not be a good way to represent it, such that the compiler printing out intermediate P4_16 programs will be either syntactically illegal P4_16, or not exactly equivalent to the P4_14 program.  Perhaps the best thing to do is make the intermediate P4_16 programs dumped syntactically legal, but not exactly the same in behavior as the P4_14 program?

jafingerhut: Seems worth creating a P4_16 language spec issue for this mask difference, in case others who defined P4_16 value_set believe that this is a significant oversight when they added it.

jafingerhut: Before I create an issue on the language spec, follow up question here on this behavioral-model issue: https://github.com/p4lang/behavioral-model/issues/854#issuecomment-582065240.   It would be good to have a working P4_14 program that uses this feature before creating a spec issue, I think.

hesingh:
Two days back on conversing with Antonin, it was decided to use a null mask.

See https://github.com/p4lang/behavioral-model/issues/856

where a public gitlab repo checked in null mask. 

I don't know what do we do with p4-14 masks in this PR.

Any plans to add a STF test?

PR #2185, opened 2020-02-04
Add pvs size to bmv2 JSON

hesingh:
Please see this issue against behavioral model where is it asked that p4c support pvs size in bvm2 JSON.

https://github.com/p4lang/behavioral-model/issues/856

### DIFF
--- a/backends/bmv2/common/JsonObjects.cpp
+++ b/backends/bmv2/common/JsonObjects.cpp
@@ -386,7 +386,7 @@ JsonObjects::add_parser_transition_key(const unsigned state_id, Util::IJson* new
 
 void
 JsonObjects::add_parse_vset(const cstring& name, const unsigned bitwidth,
-                            const big_int size) {
+                            const big_int& size) {
     auto parse_vset = new Util::JsonObject();
     unsigned id = BMV2::nextId("parse_vsets");
     parse_vset->emplace("name", name);
--- a/backends/bmv2/common/JsonObjects.cpp
+++ b/backends/bmv2/common/JsonObjects.cpp
@@ -385,14 +385,14 @@ JsonObjects::add_parser_transition_key(const unsigned state_id, Util::IJson* new
 }
 
 void
-JsonObjects::add_parse_vset(const cstring& name, const unsigned& size,
-                            const big_int& n) {
+JsonObjects::add_parse_vset(const cstring& name, const unsigned bitwidth,
+                            const big_int size) {
     auto parse_vset = new Util::JsonObject();
     unsigned id = BMV2::nextId("parse_vsets");
     parse_vset->emplace("name", name);
     parse_vset->emplace("id", id);
-    parse_vset->emplace("compressed_bitwidth", size);
-    parse_vset->emplace("size", n);
+    parse_vset->emplace("compressed_bitwidth", bitwidth);
+    parse_vset->emplace("size", size);
     parse_vsets->append(parse_vset);
 }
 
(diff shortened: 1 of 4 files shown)


hesingh:
@antoninbas and anyone else, please review - thanks.

backends/bmv2/common/JsonObjects.cpp
```diff
@@ -385,12 +385,14 @@ JsonObjects::add_parser_transition_key(const unsigned state_id, Util::IJson* new
 }
 
 void
-JsonObjects::add_parse_vset(const cstring& name, const unsigned size) {
+JsonObjects::add_parse_vset(const cstring& name, const unsigned& size,
+                            const big_int& n) {
```
antoninbas: I suggest renaming the parameters from `size` to `bitwidth` and from `n` to `size`.

backends/bmv2/common/JsonObjects.h
```diff
@@ -49,7 +49,7 @@ class JsonObjects {
     void add_parser_transition(const unsigned id, Util::IJson* transition);
     void add_parser_op(const unsigned id, Util::IJson* op);
     void add_parser_transition_key(const unsigned id, Util::IJson* key);
-    void add_parse_vset(const cstring& name, const unsigned size);
+    void add_parse_vset(const cstring& name, const unsigned& size, const big_int& n);
```
antoninbas: The change from `const unsigned` to `const unsigned&` is pointless, and I believe doesn't match the rest of the code base

hesingh:
@antoninbas Travis has passed.  Please let me know if you have any more comments.

backends/bmv2/common/JsonObjects.h
```diff
@@ -49,7 +49,8 @@ class JsonObjects {
     void add_parser_transition(const unsigned id, Util::IJson* transition);
     void add_parser_op(const unsigned id, Util::IJson* op);
     void add_parser_transition_key(const unsigned id, Util::IJson* key);
-    void add_parse_vset(const cstring& name, const unsigned size);
+    void add_parse_vset(const cstring& name, const unsigned bitwidth,
+                        const big_int size);
```
antoninbas: I said it was pointless to use `const unsigned&`, not that you shouldn't use `const big_int&`

hesingh: What is the reason that `const unsigned` is used without reference in the code base?  Using a reference with unsigned doesn't cause any damage. 
  
Then `const big_int` is used with reference because `big_int` is a typedef?

hesingh:
All but one Travis test passed.  The failed test has its failure log snippet included below.  The failure has got nothing to do with this PR code changes.  I don't know how to force the failed test to run again.

```
Reading package lists...
W: Failed to fetch http://ppa.launchpad.net/ubuntu-toolchain-r/test/ubuntu/dists/xenial/InRelease  Could not connect to ppa.launchpad.net:80 (91.189.95.83), connection timed out
```

jnfoster:
@hesingh may I propose a bit of electronic etiquette? How about as a group we agree to not comment when Travis changes status.

There are over 200 people who watch this repository, and around 70 of us watch it, so GitHub populates our feed and/or sends us an email on every commit, comment, etc. 

To keep the signal-to-noise ration high, we need to be slightly judicious about writing comments that trigger a wave of notifications. Of course, folks could stop watching the repository or set up an email filter, but then they would miss many notifications that they do want to see.

To be clear: any an all discussion of technical matters is welcome. And it's fine to ask for help here, especially from newcomers who are getting used to the workflow we use. But nudging people here to review or approve a PR or informing the group that a Travis run has passed (or not passed) is probably best avoided in most cases.

backends/bmv2/common/JsonObjects.h
antoninbas: There is a cost difference in copying an unsigned int vs. copying a big_int.

Actually passing the unsigned int by `const` reference can increase complexity and negatively impact performance: https://stackoverflow.com/a/3009568

PR #2189, opened 2020-02-05
Fixes for issues found by the klockwork static analyzer

ChrisDodd:


frontends/p4/typeChecking/typeUnification.cpp
```diff
@@ -46,8 +46,7 @@ bool TypeUnification::unifyCall(const IR::Node* errorPosition,
         if (dest->typeParameters->size() != src->typeArguments->size()) {
             TypeInference::typeError(
                 "%1% has %2% type parameters, but is invoked with %3% type arguments",
-                errorPosition, (dest->typeParameters ? dest->typeParameters->size() : 0),
-                src->typeArguments->size());
+                errorPosition, dest->typeParameters, src->typeArguments->size());
```
mbudiu-vmw: This seems to be different in semantics; you are showing the arguments rather than their count.

ChrisDodd: yes -- trivial typo in the change

PR #2191, opened 2020-02-07
Constified some function parameters.

liujed:
The `contains` and `intersects` predicates for `ordered_set` can take `const` parameters.


hesingh:
After this PR checkin, we should inspect all p4c/lib/ code and see if such a PR change is need elsewhere.   Why stop at lib.  We should look at other p4c/frontend and p4c/backend code for such changes - thanks.

PR #2192, opened 2020-02-07
out parameters checked in the right context; fixes #2175

### LINKED ISSUE #2175: Assignment removed despite conditional return
Consider the following function:

```c
void do_something(inout bit<8> val) {
    if (val == 8w0) {
        val = 8w1;
        return;
    }
    val = 8w2;
    return;
}
```
After `FrontEnd_32_SimplifyDefUse` this is turned into

```c
void do_something(inout bit<8> val) {
    if (val == 8w0) {
        ;
        return;
    }
    val = 8w2;
    return;
}

```
It seems that SimplifyDefUse does not consider the return statement and erroneously removes the assignment.

[return_in_function_if.p4.txt](https://github.com/p4lang/p4c/files/4148407/return_in_function_if.p4.txt)
[return_in_function_if.stf.txt](https://github.com/p4lang/p4c/files/4148408/return_in_function_if.stf.txt)

mbudiu-vmw:


mbudiu-vmw:
@ChrisDodd : I have 6 small bug-fix PRs in the pipeline, including this one.
Some of them are still running checks.

mbudiu-vmw:
This does symbolic evaluation and creates new visitors for callees; I would have to think to see whether the existing context would do the job.

PR #2193, opened 2020-02-07
Add only compatible values; fixes #2190

### LINKED ISSUE #2190: Compiler Bug: At this point in the compilation typechecking should not infer new types  anymore, but it did.
Hello again,
after commit 1159ced13676cba54520474d4e416b314302c1db we encounter the compiler bug in the title for specific programs. I attached an example.
The exact command is `p4c/build/p4c -v bad_type_checking.p4`.
[bad_type_checking.p4.txt](https://github.com/p4lang/p4c/files/4161569/bad_type_checking.p4.txt)

mbudiu-vmw:


jafingerhut:
It seems that when transforming an expression like this `((a >> b) >> c)` into this `(a >> (b+c))`, we need to be pretty cautious with when this is done, or we could change the value of the result.

For example, if the types of b and c are both `bit<W>`, and the values of b and c are such that (b+c) wraps around from two large values to a small value, e.g. 1, then ((a >> b) >> c) should be 0, but `(a >> (b+c))` would be `(a >> 1)`, often non-0.

I do not have a proposed list of restrictions that make the transformation safe, off-hand.  Both b and c being compile time constants that are non-negative seems pretty safe.

jafingerhut:
I don't have a comprehensive proof, but I believe that it can be proved that ((a >> b) >> c) where b is type `bit<W1>`, and c is type `bit<W2>`, is equivalent to (a >> (((bit<W3>) b) + ((bit<W3>) c))), where W3 = max(W1,W2)+1, because by widening both b and c to the widest of the two, plus one bit more, the addition is guaranteed not to wrap around.

mbudiu-vmw:
I have 5 very small PRs queued which fix bugs; if someone has a bit of time to take a look we can perhaps close these.

mbudiu-vmw:
If we want to fix the issues and bugs in the compiler we probably need to find some time to review the fixes, especially when they are small. Should I stop making fixes?

Issue #2188, opened 2020-02-05
Should P4_14 and/or P4_16 implement `parser_value_set mask 0x0ff0` construct?


jafingerhut:
@antoninbas @vgurevich 

While adding more full support for parser value sets in P4_16 to p4c and behavioral-model code, Antonin Bas pointed out that there is code in behavioral-model intended to support code that looks like this in P4_14, which p4c accepts today without any errors:
```
parser_value_set pvs0;

parser start {
    return parse_ethernet;
}

parser parse_ethernet {
    extract(ethernet);
    return select(latest.etherType) {
        pvs0 mask 0x0ff0 : ingress;       //  <--- this is the most significant line of the example
        default : ingress;
    }
}
```
Today p4c when invoked to generate a BMv2 JSON file for the v1model architecture, or when used to auto-translate P4_14 source code to P4_16 source code, produces exactly the same output regardless of whether the line marked in the example above is as shown, or when you remove the `mask 0x0ff0` part.

The P4_14 language specification does not mention this possibility, but p4c seems to implement it anyway, or at least silently accept it and at least somewhere in the implementation, either ignore it, or pay attention to it for a while but forget about it before issuing BMv2 JSON or P4_16 program output.

Questions:

+ Should the output be different for these two programs?
+ Should p4c treat the program with `mask 0x0ff0` present as a syntax error?
+ What is the intended difference in behavior?
+ If the behavior should be different, how should it be represented differently when auto-translating to P4_16 source code?

Note that v1.0.5 of the P4_14 language spec does not mention this syntax at all.  In section 4.4 "BNF" v1.0.5 of the P4_14 language spec gives this grammar rule:
```
value_or_masked ::=
    field_value | field_value mask field_value | value_set_name
```
Note that there is no `value_set_name mask field_value` on the right hand side mentioned.

Commands I used to compare outputs.  pvs1.p4 and pvs2.p4 are in the attached zip file.
```
$ p4test --std p4-14 pvs1.p4 --pp pvs1b.p4
$ p4test --std p4-14 pvs2.p4 --pp pvs2b.p4
$ diff pv1b.p4 pvs2b.p4

$ p4c --std p4-14 --target bmv2 --arch v1model pvs1.p4
$ p4c --std p4-14 --target bmv2 --arch v1model pvs2.p4
$ diff pvs1.json pvs2.json
```
The only difference in output was a couple of occurrences of the source file name in the debug info part of the BMv2 JSON files.  The 'functional part' was identical.

[...]

PR #2194, opened 2020-02-07
Add detection of p4-14 pvs

hesingh:
See 

https://github.com/p4lang/p4c/pull/2183#discussion_r374828806

https://github.com/p4lang/behavioral-model/issues/854

and

https://github.com/p4lang/p4c/issues/2188

With this detection of p4-14, the json output is now different between p4-16 and p4-14.  Now, I am waiting for the discussion to resolve for how is p4-14 mask stored in IR so that the bmv2 backend can emit appropriate JSON.

### DIFF
--- a/backends/bmv2/common/parser.cpp
+++ b/backends/bmv2/common/parser.cpp
@@ -358,6 +358,11 @@ unsigned ParserConverter::combine(const IR::Expression* keySet,
         auto vset = decl->to<IR::P4ValueSet>();
         CHECK_NULL(vset);
         auto type = ctxt->typeMap->getTypeType(vset->elementType, true);
+        int width = type->width_bits();
+        // TODO: Temporary code below because it is not known
+        // how does p4-14 send mask in IR.
+        if (vset->wasP414 && !mask)
+            mask = width;
         return ROUNDUP(type->width_bits(), 8);
     } else {
         BUG_CHECK(select->components.size() == 1, "%1%: mismatched select/label", select);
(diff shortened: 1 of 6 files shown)


hesingh:
Regarding https://github.com/p4lang/p4c/issues/2188,

If one sees the code changes in this PR, viz., this change:

https://github.com/p4lang/p4c/pull/2194/files#diff-2bb7830fa23705a8b613d9b7a96db48fR486

The P4ValueSet constructor has no mask field.   At this juncture where the constructor is called the p4-14 code has been translated to p4-16 for pvs.  Clearly, no mask from p4-14 is carried over to p4-16. So how can bmv2 JSON cater to the p4-14 mask?  The bmv2 backend cannot.

jafingerhut:
One possibility, and we will not know until questions around issue https://github.com/p4lang/p4c/issues/2188 have been answered, discussed, argued over, picked apart in detail, etc., is that Vladimir or someone will come up with a way to translate each of the P4_14 examples into an equivalent P4_16 program that does not use the `pvs mask 0xff00` syntax, or anything like it, but instead perhaps introduces formulas in the select expression instead.

I do not know the intended behavior that Vladimir or others have in their heads for all of those code snippets I asked about.  Until I know the intended meaning, I don't know if that is actually possible or not.

In short: I would suggest not worrying about the details of how to modify p4c, until we actually know how this language feature works.  Without that knowledge, changes to p4c are premature.

hesingh:
Right.  I am not modifying any more p4c.  This changes in the PR just show how to differentiate between p4-14 and p4-16 pvs and now the output JSON is not the same for each type of language program test.   I have no plans to merge this PR yet.

It's only looking at code that I asked my question.  It's a pity that p4_14_samples does not include a test using mask.

No worries.  After this note, I will wait for Vladimir or anyone else.

hesingh:
I am going by what Antonin gave for guidance in https://github.com/p4lang/p4c/pull/2183
He said "To support P4_14 masks, you would probably need some code to distinguish between missing mask and mask set to 0."

PR #2198, opened 2020-02-10
One instantiation per value set; fixed #2196

### LINKED ISSUE #2196: p4c does not support the use of parser value sets in multiple parser states for P4_14
The P4_14 spec states that:

> The width of the values is inferred from the place where the value set is referenced. If the
set is used in multiple places and they would infer different widths, then the compiler
must raise an error.

which implies that following program (https://github.com/p4lang/behavioral-model/pull/859#issuecomment-581560962) is valid:
```
// Standard L2 Ethernet header
header_type ethernet_t {
    fields {
        dst_addr        : 48; // width in bits
        src_addr        : 48;
        ethertype       : 16;
    }
}

header ethernet_t ethernet;

header_type my_header_t {
    fields {
        f1 : 3;
        f2 : 13;
    }
}

header my_header_t my_header;

parser_value_set pv1;
parser_value_set pv2;

header_type my_meta_t {
    fields {
        pv_parsed : 8;
    }
}

metadata my_meta_t my_meta;

parser start {
    extract(ethernet);
    return select(ethernet.ethertype) {
        0x0800 : ingress;
        pv1 : parse_pv1;
        pv2 : parse_pv2;
        default : ingress;
    }
}

parser parse_pv1 {
    set_metadata(my_meta.pv_parsed, 1);
    extract(my_header);
    return select(my_header.f1, my_header.f2) {
        pv2 : parse_pv2;
        default : ingress;
    }
}

parser parse_pv2 {
    set_metadata(my_meta.pv_parsed, 2);
    return ingress;
}

action route_eth(egress_spec, src_addr) {
    modify_field(standard_metadata.egress_spec, egress_spec);
    modify_field(ethernet.src_addr, src_addr);
}
action noop() { }

table routing {
    reads {
	ethernet.dst_addr : lpm;
    }
    actions {
	route_eth;
	noop;
    }
}

control ingress {
    apply(routing);
}
```

I verified that the program compiles (and produces a correct bmv2 JSON) with the old p4-hlir / p4c-bm toolchain.

With p4c, I get the following error:
```
[--Werror=duplicate] error: .pv2: Duplicates declaration .pv2
[--Werror=overlimit] error: 1 errors encountered, aborting compilation
```

Note that even if I make the program "easier" by replacing transition key `select(my_header.f1, my_header.f2)` with `select(ethernet.ethertype)` so that the 2 transition keys (in the 2 parser states) are the same, I get the same error (p4c tries to duplicate the PVS definition, with the same name).

I am not sure that there is even a good way to translate this into P4_16 - at least when the transition keys are different. In the above example, we have a 16bit key and a 3+13bit key.

mbudiu-vmw:


hesingh:
Please get rid of compile warnings by adding a pragme about each v1 pvs declaration as follows to issue2196.p4

```
@pragma parser_value_set_size 4
parser_value_set pv1;
@pragma parser_value_set_size 4
parser_value_set pv2;
```

@antoninbas How do we generate p4runtime info for the P4 code so that we can check what pvs is generated.   Or this P4 program does not generate any p4info?  I tried --p4runtime-entries-file but get a file with output as zero bytes.

jafingerhut:
@hesingh If you look in the p4c/testdata/p4_16_samples_outputs/ directory you will see many files with p4info in their names, and all of them have `-bmv2.p4` at the end of the test program name.  The p4c test infrastructure always tries to generate the p4info files for P4_16 programs named with that suffix, and checks against those expected output file names if they exist in the p4_16_samples_outputs/ directory.

If you want to generate it manually on your own local system, you can use the command line option `--p4runtime-files some-file-name.txt`

hesingh:
@jafingerhut I am aware of all those things.
I did change the P4 program name to use "-bmv2.p4" but still no luck.

`$ ./p4c-bm2-ss --std p4-14 issue2196-bmv2.p4 --p4runtime-files tmp.txt`

tmp.txt does not include a parse_vset.  Also, this is a p4-14 program and if stick the program in p4c/testdata/p4_14_sample/, and run `p4/p14_to_16/testdata/p4_14_samples/issue2196.p4.test -f`, no p4info.txt file is generated.

hesingh:
I also used

`$ p14_to_16/testdata/p4_14_samples/issue2196-bmv2.p4.test -f`

and still no p4info.txt file is generated.

jafingerhut:
P4Runtime API is not supported for P4_14, only P4_16.  I do not know whether p4c disables P4Info generation when the input program is P4_14, but it might.

hesingh:
I know p4runtime is not supported for p4-14.  But the p4-14 program has been translated to p4-16 as seen in this file.

https://github.com/p4lang/p4c/pull/2198/files#diff-15828abd5b6a3af399b584ae183363b4

The file exists in this PR's list of files. 

We should edit any .py file to then run p4runtime generation on the translated file.

jafingerhut:
I would not be surprised if p4c does not generate P4Info files when you use the `--std p4-14` command line option, indicating that the input file is P4_14, even though as an intermediate step it internally can create P4_16 code from it.

jafingerhut:
If you really, really want an automated check for the P4Info output, then create a P4_16 source file however you like, whether it is auto-generated from a P4_14 program or not.  There is no need to complicate the test infra for a few special cases, is there?

hesingh:
We can discuss the issue outside of this PR.

PR #2199, opened 2020-02-11
Reject value sets with masks

mbudiu-vmw:
This addresses #2188 somewhat: it will reject such programs instead of converting them silently into something incorrect. But #2188 also asks whether we want to support this idiom.

### DIFF
--- a/frontends/p4/fromv1.0/programStructure.cpp
+++ b/frontends/p4/fromv1.0/programStructure.cpp
@@ -494,7 +494,7 @@ const IR::ParserState* ProgramStructure::convertParser(const IR::V1Parser* parse
                 } else {
                     auto c = v.second->to<IR::Constant>();
                     CHECK_NULL(c);  // this is enforced elsewhere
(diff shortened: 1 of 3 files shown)


jafingerhut:
Mihai, not sure if your previous comment was intentional, and intended for this PR, but if it was, I think someone would need some more detail in the question in order to have a chance of answering.

mbudiu-vmw:
Maybe some comments were deleted.

hesingh:
> Why?

The bmv2 backend does not support "pvs mask <hex>" yet.  If supported, the pvs has to be treated as a ternary operation with (value, mask) pair.  The mask is handy for the pair, but where does the value come from?  The pvs value is known during runtime.  Thus, this PR should say that the feature in UNIMPLEMENTED if the code changes do not already say so.   If the mask is "-1", the there is no mask with pvs.  See comments related to mask in at this link as well.

https://github.com/p4lang/p4c/pull/2194/files#diff-77d3a89cfadf7b688394438e594b58d7R268

PR #2204, opened 2020-02-20
[Testing] Always ApplyTypesToExpressions after Typechecking

ChrisDodd:


mbudiu-vmw:
I tried this, and it doesn't work because inserting types changes the graph and invalidates any hash table that may have been built. That's why typeChecking can't do  this in the ir either.

Issue #2210, opened 2020-02-25
Compiler warnings when building p4c with gcc 9.2.1 and bison 3.4.1


krzycz:
There is a lot of warnings when building **p4c** with gcc 9.2.1 / bison 3.4.1 (Fedora 31).
```
p4c/lib/source_file.h:182:40: warning: implicitly-declared ‘Util::SourcePosition& Util::SourcePosition::operator=(const Util::SourcePosition&)’ is deprecated [-Wdeprecated-copy]
  182 |             start = start.min(rhs.start);
```
```
p4c/frontends/common/options.cpp:279:25: warning: ‘snprintf’ output may be truncated before the last format character [-Wformat-truncation=]
  279 |     snprintf(output, N, "%s%s%s", cwd, separator, relPath);
      |                         ^~~~~~~~
```
```
parsers/v1/v1parser.ypp:24.1-36: warning: deprecated directive, use ‘%define api.parser.class {V1Parser}’ [-Wdeprecated]
   24 | %define parser_class_name {V1Parser}
      | ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
parsers/v1/v1parser.ypp:34.1-14: warning: deprecated directive, use ‘%define parse.error verbose’ [-Wdeprecated]
   34 | %error-verbose
      | ^~~~~~~~~~~~~~
```
```
p4c/backends/bmv2/common/expression.cpp:449:26: warning: ‘e’ may be used uninitialized in this function [-Wmaybe-uninitialized]
  449 |                 e->append(fieldName);
      |                 ~~~~~~~~~^~~~~~~~~~~
```
```
p4c/ir/type.def: In member function ‘virtual bool IR::Type_InfInt::equiv(const IR::Node&) const’:
p4/p4c/ir/type.def:175:15: warning: unused variable ‘a’ [-Wunused-variable]
  175 |  #endif
      |               ^
```

mbudiu-vmw:
I tried to fix the bison warnings, but they work with older versions of Bison, so I think we'll live with them for now.

PR #2211, opened 2020-02-25
Fix gcc warnings

mbudiu-vmw:
Fix gcc warnings from #2210

### DIFF
--- a/lib/source_file.h
+++ b/lib/source_file.h
@@ -56,6 +56,7 @@ class SourcePosition final {
               columnNumber(0) {}
 
     SourcePosition(unsigned lineNumber, unsigned columnNumber);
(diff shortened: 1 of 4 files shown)


lib/source_file.h
```diff
@@ -56,6 +56,7 @@ class SourcePosition final {
               columnNumber(0) {}
 
     SourcePosition(unsigned lineNumber, unsigned columnNumber);
+    SourcePosition& operator=(const SourcePosition&) = default;
```
krzycz: I noticed the same issue in a few other places: See: **ProgramPoint** and **Node**.

mbudiu-vmw: Do you get warnings about them from gcc?

mbudiu-vmw:
Feel free to submit a fix for bison.

PR #2214, opened 2020-02-26
Warn in bmv2 backend about unsupported match types in value_sets

antoninbas:
bmv2 only supports exact matches, and a warning is better than silently
ignoring the @match annotation.

### DIFF
--- a/backends/bmv2/common/parser.cpp
+++ b/backends/bmv2/common/parser.cpp
@@ -18,6 +18,7 @@ limitations under the License.
 #include "JsonObjects.h"
 #include "backend.h"
 #include "extern.h"
+#include "frontends/p4/coreLibrary.h"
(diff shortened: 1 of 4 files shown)


jafingerhut:
I have tried out these changes and the new warning messages look good to me.

I am confused about the automated tests, though.  The test program testdata/p4_16_samples/pvs-struct-3-bmv2.p4 without these changes has no warnings, but now it does (as desired).  So why don't the tests fail because the expected stderr output file is empty for that one?

mbudiu-vmw:
The tests only check stderr for p4test.

jafingerhut:
Thanks, Mihai.  Makes sense.  I feel like I need to write this down somewhere before I forget, in a place I remember to look for it the next time I am confused about how the test infra works :)

PR #2215, opened 2020-02-26
Use relative paths for symlinks

ChrisDodd:
Using absolute paths for symlinks interacts badly with trying to use docker containers that want to mount things with different paths.  Relative symlinks work better.

### DIFF
--- a/backends/bmv2/CMakeLists.txt
+++ b/backends/bmv2/CMakeLists.txt
@@ -103,10 +103,10 @@ install(TARGETS p4c-bm2-psa RUNTIME DESTINATION ${P4C_RUNTIME_OUTPUT_DIRECTORY})
 # binary to be in the top level directory. This should go away when we
 # remove automake and fix the scripts.
 add_custom_target(linkbmv2
-  COMMAND ${CMAKE_COMMAND} -E create_symlink ${CMAKE_CURRENT_BINARY_DIR}/p4c-bm2-ss ${P4C_BINARY_DIR}/p4c-bm2-ss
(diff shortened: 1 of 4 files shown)


mbudiu-vmw:
@ChrisDodd : I am now getting the following error when trying to run a compiler from the build directory:

```
 ./p4test -v issue2206.p4 
Invoking preprocessor 
cpp -C -undef -nostdinc -x assembler-with-cpp  -I/usr/local/share/p4c/p4include issue2206.p4
issue2206.p4:1:19: error: no include path in which to search for core.p4
 #include <core.p4>
                   ^
```

mbudiu-vmw:
This happens if I didn't `make install`

ChrisDodd:
Hmm -- looks like something is causing it to generate an incorrect `-I` option, which should not depend on the symlinks (and whether they are relative or absolute)

ChrisDodd:
Looks like it depends on the envvar `P4C_16_INCLUDE_PATH` to get the correct include path -- you need to run the `p4c` driver program to get this set rather than running the compiler directly.  I thought I had fixed the code to not need need that at one point, but it seems to have been broken...

This is completely independent of whether the symlinks are absolute or relative.

mbudiu-vmw:
That is weird; I have never run the driver. I don't understand why this broke suddenly.
This is the only change that seemed related.

ChrisDodd:
Running this as `P4C_16_INCLUDE_PATH=p4include ./p4test -v issue2206.p4` should also avoid the problem.

asl:
This seems to break mac builds:

1. There is no realpath builtin in /bin/sh (which executes the shell part there)
2. The realpath builtin in /bin/fish does not support --relative-to
3. realpath from coreutils does support --relative-to, but I do not have them installed and if installed it will be “grealpath” in order not to mess with the system utilities

mbudiu-vmw:
If this is a bug, please open a new issue. Commenting on a closed PR is not as useful.

PR #2218, opened 2020-02-28
Remove broken link-handling code (stat follows links anyways)

ChrisDodd:


ChrisDodd:
This should fix the different behavior between absolute and relative links

PR #2219, opened 2020-02-29
Update stf test

pkotikal:
Change the expected counter values in the test +4 bytes per packet

### DIFF
--- a/testdata/p4_14_samples/counter3.stf
+++ b/testdata/p4_14_samples/counter3.stf
@@ -7,13 +7,13 @@ expect 3
 expect 5
 packet 0 a1a2a3a4a5a6 0000
 wait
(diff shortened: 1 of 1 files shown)


jafingerhut:
Is your comment about running afoul of ethernet 64-byte minimum packet size specific to only this test, or all STF tests in general?  Because there are probably a large fraction of STF tests that use shorter packets, since apparently veth interfaces let you pass them, as long as they are at least 14 bytes long.

ChrisDodd:
It potentially applies to all tests, but is really only relevant for tests that will be affected by added padding in a driver or kernel, like this one that counts the number of packet bytes.  Tests that only look at what they look at and don't care that padding might be added to the end of the packet should not be affected.

PR #2222, opened 2020-03-03
Some small changes to run-bmv2-test.py

fruffy:
Added options.compilerSrcDir to a few paths so that `run-bmv2-test.py` can be run outside of the build directory.
Plus a few autoformatting adjustments.

### DIFF
--- a/backends/bmv2/run-bmv2-test.py
+++ b/backends/bmv2/run-bmv2-test.py
@@ -40,6 +40,7 @@ import pdb
 SUCCESS = 0
 FAILURE = 1
 
+
 class Options(object):
     def __init__(self):
         self.binary = ""                # this program's name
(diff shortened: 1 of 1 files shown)


backends/bmv2/run-bmv2-test.py
```diff
@@ -220,13 +235,13 @@ def process_file(options, argv):
         raise Exception("No such file " + options.p4filename)
 
     if options.usePsa:
-        binary = "./p4c-bm2-psa"
+        binary = options.compilerSrcDir + "/build/p4c-bm2-psa"
     else:
-        binary = "./p4c-bm2-ss"
+        binary = options.compilerSrcDir + "/build/p4c-bm2-ss"
```
ChrisDodd: There should also be a way to specify the compiler build subdirectory, which might not be `build` -- I often have `build-release` and `build-debug` directories.  This will prevent running tests at all in such directories.

fruffy: I think we even can replace compilerSrcDir with compilerBuildDir entirely with a couple minor changes. Is that okay? How is `options.testName`, from a superficial look it seems to specify the output test folder?

ChrisDodd: That would seem to be fine.  `options.testName` is only set if the name on the command line includes the path to the compiler source directory (so depends on having that), implying that you might be able to use the script to run tests that aren't in the compiler tree?  Probably for people trying to make new tests or get them from the tutorials or some such.

Not sure how many different ways people regularly use the test scripts -- everyone ends up having their own preferred workflow.

fruffy: Yes, it is a little bit inconvenient to run new tests because they require the python file to be executed from the build folder. I would also like to write external tools that call `run-bmv2-test.py`.
Hmm the downside of switching to `compilerBuildDir`is that the command that generates the `.test` files in `P4CUtils.cmake` has to be updated. I guess that is a larger change impacting not only the bmv2 backend. So we are probably better off leaving this as is. I can add the option.

backends/bmv2/run-bmv2-test.py
```diff
@@ -40,12 +40,14 @@
...
 FAILURE = 1
 
+
 class Options(object):
     def __init__(self):
         self.binary = ""                # this program's name
         self.cleanupTmp = True          # if false do not remote tmp folder created
         self.p4Filename = ""            # file that is being compiled
         self.compilerSrcDir = ""        # path to compiler source tree
+        self.compilerBuildDir = "build"  # the build subfolder
```
ChrisDodd: should default this based on the current working directory -- if you're running in a subdirectory of the compilerSrcDir, then that should be the build dir.  Maybe do that in main (after setting compilerSrcDir) rather than here?

fruffy: Okay pushed a version that uses the working directory as default.

Issue #2213, opened 2020-02-25
Compiler bug when attempting to compile nested struct initializer


jafingerhut:
The attached program causes a compiler bug when attempting to compile it with p4c 1.2.0 SHA: 62fd40a0.  Commenting out the following line enables the same version of p4c to compile the rest of the code:
```
        meta.s2 = {s1={f1=2, f2=3}, f3=5, f4=8};
```
Similar lines that use struct initializers that are not nested appear to work correctly, so perhaps the issue is because of the nested struct initializer expression.

Is this an expected use of struct initializer expressions according to the language spec?  It seems grammatically correct, at least, but may be bad for other reasons.

[nested-struct-initializers.p4.txt](https://github.com/p4lang/p4c/files/4252345/nested-struct-initializers.p4.txt)

mbudiu-vmw:
The problem is that typeInference does not convert the list initializer for the inner struct to a struct-valued expression.

Issue #797, opened 2017-07-27
P4-14 compiler does not complain about duplicate definitions


mbudiu-vmw:
```
header_type X {
    fields {
        x : 32;
    }
}

header_type X {
    fields {
        y : 32;
    }
}

header X x;

parser start {
    return ingress;
}

control ingress {}
```

The P4-14 front-end should reject this program, or at least give a warning.

mbudiu-vmw:
This is a duplicate of #763

PR #2224, opened 2020-03-04
Structs, headers and tuples initialization support (#762) (#2017)

marko-stanoj:
Struct, header or tuple can be initialized automatically with a default value of
the suitable type using the syntax `...` or using a mix of explicit
values and default values by using the notation `...` in a list
expression initializer(and struct initializer expression for structs).

Mix of explicit values and default values can also be used for assignment.

For `struct`, `tuple` and `header` types the default values are `struct`,
`tuple` and `header` respectively, where each field has the default value
of the suitable field type -- if all such default values are defined.

Default value for `header` type could be easily changed to invalid, as proposed
in p4-spec pull request #797, when a new literal for initializing `header` to
invalid is introduced.

Header stacks and `header_union` are not supported as a leaf fields of structs
and tuples if using this kind of initialization, because in current version of
spec there are no constructs to initialize these types.
If they are used as a leaf fields compiler will report an error.

along with assignment statemants, which is the cause of changes in expected outputs
of some tests.

### DIFF
--- a/frontends/p4/structInitializers.cpp
+++ b/frontends/p4/structInitializers.cpp
@@ -13,11 +13,11 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 See the License for the specific language governing permissions and
 limitations under the License.
 */
-
 #include "structInitializers.h"
 
 namespace P4 {
-
+const IR::Expression*
+convert(const IR::Expression* expression, const IR::Type* type);
 /// Given an expression and a destination type, convert ListExpressions
 /// that occur within expression to StructInitializerExpression if the
 /// destination type matches.
@@ -28,26 +28,129 @@ convert(const IR::Expression* expression, const IR::Type* type) {
         auto si = new IR::IndexedVector<IR::NamedExpression>();
         if (auto le = expression->to<IR::ListExpression>()) {
             size_t index = 0;
+            auto l = le->size();
             for (auto f : st->fields) {
+                if (index >= l) {
+                    break;
+                }
                 auto expr = le->components.at(index);
                 auto conv = convert(expr, f->type);
                 auto ne = new IR::NamedExpression(conv->srcInfo, f->name, conv);
                 si->push_back(ne);
                 index++;
             }
+            if (le->defaultInitializer) {
+                modified = true;
+                IR::Expression* expr;
+                for (; index < st->fields.size(); index++) {
+                    auto f = st->fields.at(index);
+                    if (f->type->is<IR::Type_Bits>()) {
+                        auto conv = new IR::Constant(0);
+                        expr = conv;
+                        auto ne = new IR::NamedExpression(expression->srcInfo, f->name, expr);
+                        si->push_back(ne);
+                    } else if (f->type->is<IR::Type_Boolean>()) {
+                        auto conv = new IR::BoolLiteral(false);
+                        expr = conv;
+                        auto ne = new IR::NamedExpression(expression->srcInfo, f->name, expr);
+                        si->push_back(ne);
+                    } else if (f->type->is<IR::Type_Error>()) {
+                        auto path = new IR::Path(IR::ID(expression->srcInfo, "error"));
+                        auto typeName =  new IR::Type_Name(expression->srcInfo, path);
+                        auto id = new IR::ID(expression->srcInfo, "NoError");
+                        auto tne = new IR::TypeNameExpression(expression->srcInfo, typeName);
+                        expr = new IR::Member(expression->srcInfo, tne, *id);
+                        auto ne = new IR::NamedExpression(expression->srcInfo, f->name, expr);
+                        si->push_back(ne);
+                    } else if (auto t = f->type->to<IR::Type_Enum>()) {
+                        auto tne =  new IR::TypeNameExpression(expression->srcInfo,
+                                                               new IR::Type_Name(t->name));
+                        expr = new IR::Member(expression->srcInfo, tne, t->members.at(0)->name);
+                        auto ne = new IR::NamedExpression(expression->srcInfo, f->name, expr);
+                        si->push_back(ne);
+
+                    } else if (auto t = f->type->to<IR::Type_SerEnum>()) {
+                        int size = t->type->size;
+                        auto conv = new IR::Constant(new IR::Type_Bits(size, false), 0);
+                        expr = conv;
+                        auto ne = new IR::NamedExpression(expression->srcInfo, f->name, expr);
+                        si->push_back(ne);
+                    } else {
+                        auto vec = new IR::Vector<IR::Expression>();
+                        auto conv = convert(new IR::ListExpression(expression->srcInfo, *vec, true),
+                                            f->type);
+                        auto ne = new IR::NamedExpression(expression->srcInfo, f->name, conv);
+                        si->push_back(ne);
+                    }
+                }
+            }
             auto type = st->getP4Type()->to<IR::Type_Name>();
             auto result = new IR::StructInitializerExpression(
                 expression->srcInfo, type, type, *si);
             return result;
         } else if (auto sli = expression->to<IR::StructInitializerExpression>()) {
-            for (auto f : st->fields) {
-                auto ne = sli->components.getDeclaration<IR::NamedExpression>(f->name.name);
-                BUG_CHECK(ne != nullptr, "%1%: no initializer for %2%", expression, f);
-                auto convNe = convert(ne->expression, f->type);
-                if (convNe != ne->expression)
-                    modified = true;
-                ne = new IR::NamedExpression(ne->srcInfo, f->name, convNe);
-                si->push_back(ne);
+            if (!sli->defaultInitializer) {
+                for (auto f : st->fields) {
+                    auto ne = sli->components.getDeclaration<IR::NamedExpression>(f->name.name);
+                    BUG_CHECK(ne != nullptr, "%1%: no initializer for %2%", expression, f);
+                    auto convNe = convert(ne->expression, f->type);
+                    if (convNe != ne->expression)
+                        modified = true;
+                    ne = new IR::NamedExpression(ne->srcInfo, f->name, convNe);
+                    si->push_back(ne);
+                }
+            } else {
+                modified = true;
+                IR::Expression* expr;
+                for (auto f : st->fields) {
+                    auto ne = sli->components.getDeclaration<IR::NamedExpression>(f->name.name);
+                    if (ne == nullptr) {
+                        if (f->type->is<IR::Type_Bits>()) {
+                            auto conv = new IR::Constant(0);
+                            expr = conv;
+                            ne = new IR::NamedExpression(expression->srcInfo, f->name, expr);
+                            si->push_back(ne);
+                        } else if (f->type->is<IR::Type_Boolean>()) {
+                            auto conv = new IR::BoolLiteral(false);
+                            expr = conv;
+                            ne = new IR::NamedExpression(expression->srcInfo, f->name, expr);
+                            si->push_back(ne);
+                        } else if (f->type->is<IR::Type_Error>()) {
+                            auto path = new IR::Path(IR::ID(expression->srcInfo, "error"));
+                            auto typeName =  new IR::Type_Name(expression->srcInfo, path);
+                            auto id = new IR::ID(expression->srcInfo, "NoError");
+                            auto tne = new IR::TypeNameExpression(expression->srcInfo, typeName);
+                            expr = new IR::Member(expression->srcInfo, tne, *id);
+                            auto ne = new IR::NamedExpression(expression->srcInfo, f->name, expr);
+                            si->push_back(ne);
+                        } else if (auto t = f->type->to<IR::Type_Enum>()) {
+                            auto tne =  new IR::TypeNameExpression(expression->srcInfo,
+                                                                   new IR::Type_Name(t->name));
+                            expr = new IR::Member(expression->srcInfo, tne, t->members.at(0)->name);
+                            auto ne = new IR::NamedExpression(expression->srcInfo, f->name, expr);
+                            si->push_back(ne);
+
+                        } else if (auto t = f->type->to<IR::Type_SerEnum>()) {
+                            int size = t->type->size;
+                            auto conv = new IR::Constant(new IR::Type_Bits(size, false), 0);
+                            expr = conv;
+                            auto ne = new IR::NamedExpression(expression->srcInfo, f->name, expr);
+                            si->push_back(ne);
+                        } else {
+                            auto vec = new IR::Vector<IR::Expression>();
+                            auto conv_le = new IR::ListExpression(expression->srcInfo, *vec, true);
+                            auto conv = convert(conv_le, f->type);
+                            ne = new IR::NamedExpression(expression->srcInfo, f->name, conv);
+                            si->push_back(ne);
+                        }
+                    } else {
+                        auto convNe = convert(ne->expression, f->type);
+                        if (convNe != ne->expression)
+                            modified = true;
+                        ne = new IR::NamedExpression(ne->srcInfo, f->name, convNe);
+                        si->push_back(ne);
+                    }
+                }
             }
             if (modified || sli->type->is<IR::Type_Unknown>()) {
                 auto type = st->getP4Type()->to<IR::Type_Name>();
@@ -62,12 +165,48 @@ convert(const IR::Expression* expression, const IR::Type* type) {
             return expression;
 
         auto vec = new IR::Vector<IR::Expression>();
-        for (size_t i = 0; i < le->size(); i++) {
-            auto expr = le->components.at(i);
-            auto type = tup->components.at(i);
+        size_t index = 0;
+        auto l = le->size();
+        for ( ; index < l; index++) {
+            auto expr = le->components.at(index);
+            auto type = tup->components.at(index);
             auto conv = convert(expr, type);
             vec->push_back(conv);
             modified |= (conv != expr);
+        }
+        if (le->defaultInitializer) {
+            modified = true;
+            for ( ; index < tup->components.size(); index++) {
+                auto f = tup->components.at(index);
+                if (f->is<IR::Type_Bits>()) {
+                    auto conv = new IR::Constant(0);
+                    vec->push_back(conv);
+                } else if (f->is<IR::Type_Boolean>()) {
+                    auto conv = new IR::BoolLiteral(false);
+                    vec->push_back(conv);
+                } else if (f->is<IR::Type_Error>()) {
+                    auto path = new IR::Path(IR::ID(expression->srcInfo, "error"));
+                    auto typeName =  new IR::Type_Name(expression->srcInfo, path);
+                    auto tne = new IR::TypeNameExpression(expression->srcInfo, typeName);
+                    auto conv = new IR::Member(expression->srcInfo, tne,
+                                               *(new IR::ID(expression->srcInfo, "NoError")));
+                    vec->push_back(conv);
+                } else if (auto t = f->to<IR::Type_Enum>()) {
+                    auto tne = new IR::TypeNameExpression(expression->srcInfo,
+                                                          new IR::Type_Name(t->name));
+                    auto conv = new IR::Member(expression->srcInfo, tne, t->members.at(0)->name);
+                    vec->push_back(conv);
+                } else if (auto t = f->to<IR::Type_SerEnum>()) {
+                    int size = t->type->size;
+                    auto conv = new IR::Constant(new IR::Type_Bits(size, false), 0);
+                    vec->push_back(conv);
+                } else {
+                    auto vect = new IR::Vector<IR::Expression>();
+                    auto conv_le = new IR::ListExpression(expression->srcInfo, *vect, true);
+                    auto conv = convert(conv_le, f);
+                    vec->push_back(conv);
+                }
+            }
         }
         if (modified) {
             auto result = new IR::ListExpression(expression->srcInfo, *vec);
@@ -75,14 +214,6 @@ convert(const IR::Expression* expression, const IR::Type* type) {
         }
     }
     return expression;
-}
-
-const IR::Node* CreateStructInitializers::postorder(IR::AssignmentStatement* statement) {
-    auto type = typeMap->getType(statement->left);
-    auto init = convert(statement->right, type);
-    if (init != statement->right)
-        statement->right = init;
-    return statement;
 }
 
 const IR::Node* CreateStructInitializers::postorder(IR::MethodCallExpression* expression) {
@@ -122,6 +253,14 @@ const IR::Node* CreateStructInitializers::postorder(IR::Operation_Relation* expr
     if (rtype->is<IR::Type_StructLike>() && ltype->is<IR::Type_List>())
         expression->left = convert(expression->left, rtype);
     return expression;
+}
+
+const IR::Node* CreateStructAssignInitializers::postorder(IR::AssignmentStatement* statement) {
+    auto type = typeMap->getType(statement->left);
+    auto init = convert(statement->right, type);
+    if (init != statement->right)
+        statement->right = init;
+    return statement;
 }
 
 }  // namespace P4
--- a/frontends/p4/structInitializers.cpp
+++ b/frontends/p4/structInitializers.cpp
@@ -18,6 +18,37 @@ limitations under the License.
 namespace P4 {
 const IR::Expression*
 convert(const IR::Expression* expression, const IR::Type* type);
+
+const IR::Expression* defaultValue(const IR::Type* type, const Util::SourceInfo srcInfo) {
+    IR::Expression* expr;
+    if (type->is<IR::Type_Bits>()) {
+        expr = new IR::Constant(type, 0);
+    } else if (type->is<IR::Type_Boolean>()) {
+        expr = new IR::BoolLiteral(false);
+    } else if (type->is<IR::Type_Error>()) {
+        auto path = new IR::Path(IR::ID(srcInfo, "error"));
+        auto typeName =  new IR::Type_Name(srcInfo, path);
+        auto id = new IR::ID(srcInfo, "NoError");
+        auto tne = new IR::TypeNameExpression(srcInfo, typeName);
+        expr = new IR::Member(srcInfo, tne, *id);
+    } else if (auto t = type->to<IR::Type_Enum>()) {
+        BUG_CHECK(t->members.size(), "unknown default value for enum with no members");
+        auto tne = new IR::TypeNameExpression(srcInfo,
+                                               new IR::Type_Name(t->name));
+        expr = new IR::Member(srcInfo, tne, t->members.at(0)->name);
+    } else if (auto t = type->to<IR::Type_SerEnum>()) {
+        expr = new IR::Constant(t->type, 0);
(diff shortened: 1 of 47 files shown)


frontends/p4/structInitializers.cpp
```diff
@@ -28,26 +28,129 @@ convert(const IR::Expression* expression, const IR::Type* type) {
...
                 auto ne = new IR::NamedExpression(conv->srcInfo, f->name, conv);
                 si->push_back(ne);
                 index++;
             }
+            if (le->defaultInitializer) {
+                modified = true;
+                IR::Expression* expr;
+                for (; index < st->fields.size(); index++) {
+                    auto f = st->fields.at(index);
+                    if (f->type->is<IR::Type_Bits>()) {
```
mbudiu-vmw: I would make a separate function for this, `IR::Expression* defaultValue(const IR::Type*)`.

frontends/p4/structInitializers.cpp
```diff
@@ -28,26 +28,129 @@ convert(const IR::Expression* expression, const IR::Type* type) {
...
+                        auto typeName =  new IR::Type_Name(expression->srcInfo, path);
+                        auto id = new IR::ID(expression->srcInfo, "NoError");
+                        auto tne = new IR::TypeNameExpression(expression->srcInfo, typeName);
+                        expr = new IR::Member(expression->srcInfo, tne, *id);
+                        auto ne = new IR::NamedExpression(expression->srcInfo, f->name, expr);
+                        si->push_back(ne);
+                    } else if (auto t = f->type->to<IR::Type_Enum>()) {
+                        auto tne =  new IR::TypeNameExpression(expression->srcInfo,
+                                                               new IR::Type_Name(t->name));
+                        expr = new IR::Member(expression->srcInfo, tne, t->members.at(0)->name);
```
mbudiu-vmw: Are we checking somewhere that enums have at least 1 member?
Can you add a test for this case in testdata/p4_16_errors?

frontends/p4/structInitializers.cpp
```diff
@@ -28,26 +28,129 @@ convert(const IR::Expression* expression, const IR::Type* type) {
...
+                        auto ne = new IR::NamedExpression(expression->srcInfo, f->name, expr);
+                        si->push_back(ne);
+
```
mbudiu-vmw: why empty line here?

frontends/p4/structInitializers.cpp
```diff
@@ -28,26 +28,129 @@ convert(const IR::Expression* expression, const IR::Type* type) {
...
+                    } else if (auto t = f->type->to<IR::Type_SerEnum>()) {
+                        int size = t->type->size;
+                        auto conv = new IR::Constant(new IR::Type_Bits(size, false), 0);
```
mbudiu-vmw: you should use `t->type` instead. E.g., it can be signed.

frontends/p4/structInitializers.cpp
```diff
@@ -28,26 +28,129 @@ convert(const IR::Expression* expression, const IR::Type* type) {
...
+                        expr = conv;
+                        auto ne = new IR::NamedExpression(expression->srcInfo, f->name, expr);
+                        si->push_back(ne);
+                    } else {
```
mbudiu-vmw: I am uncomfortable about this else; this will handle types that you do not allow as well.
You should rather check that the type is legal and give an error in the else branch.

frontends/p4/structInitializers.cpp
```diff
@@ -28,26 +28,129 @@ convert(const IR::Expression* expression, const IR::Type* type) {
...
+                    ne = new IR::NamedExpression(ne->srcInfo, f->name, convNe);
+                    si->push_back(ne);
+                }
+            } else {
+                modified = true;
+                IR::Expression* expr;
+                for (auto f : st->fields) {
+                    auto ne = sli->components.getDeclaration<IR::NamedExpression>(f->name.name);
+                    if (ne == nullptr) {
+                        if (f->type->is<IR::Type_Bits>()) {
```
mbudiu-vmw: having a function will also remove all this code duplication

frontends/p4/structInitializers.cpp
```diff
@@ -28,26 +28,129 @@ convert(const IR::Expression* expression, const IR::Type* type) {
...
                 si->push_back(ne);
                 index++;
             }
+            if (le->defaultInitializer) {
+                modified = true;
+                IR::Expression* expr;
+                for (; index < st->fields.size(); index++) {
+                    auto f = st->fields.at(index);
+                    if (f->type->is<IR::Type_Bits>()) {
+                        auto conv = new IR::Constant(0);
```
mbudiu-vmw: Why not make the constant of the type `f->type`?

frontends/p4/typeChecking/typeUnification.cpp
```diff
@@ -303,64 +302,142 @@ bool TypeUnification::unify(const IR::Node* errorPosition,
             return false;
         }
         auto ts = src->to<IR::Type_BaseList>();
-        if (td->components.size() != ts->components.size()) {
+        bool convenient;
```
mbudiu-vmw: you have to initialize convenient. Doesn't the C compiler give you a warning about this?

mbudiu-vmw:
Fixes #2213 as well

frontends/parsers/p4/p4lexer.ll
```diff
@@ -194,6 +194,7 @@ using Parser = P4::P4Parser;
                           return Parser::make_INTEGER(constant, driver.yylloc); }
 
 "&&&"           { BEGIN(driver.saveState); return makeToken(MASK); }
+"..."           { BEGIN(driver.saveState); return makeToken(CONVINIENT_INITIALIZER); }
```
mbudiu-vmw: "convenient"
But perhaps we should call this DEFAULT_INITIALIZER, or DOTS?

frontends/p4/typeChecking/typeUnification.cpp
```diff
@@ -303,64 +302,142 @@ bool TypeUnification::unify(const IR::Node* errorPosition,
...
+                if (!success)
+                    return false;
+            }
+        } else {
+            for (size_t i=0; i < td->components.size(); i++) {
+                auto di = td->components.at(i);
+                if (di->is<IR::Type_Varbits>() || di->is<IR::Type_Stack>()
+                    || di->is<IR::Type_HeaderUnion>()) {
+                    TypeInference::typeError("%1%: Structs, headers and tuples containing %2% data "
+                                             "type as a lief field cannot be zero initialized with "
```
mbudiu-vmw: `Containing %2% fields cannot be initialized using "..."`

frontends/p4/typeChecking/typeUnification.cpp
```diff
@@ -303,64 +302,142 @@ bool TypeUnification::unify(const IR::Node* errorPosition,
...
+                    bool success = unify(errorPosition, destt, tplField, reportErrors);
+                    if (!success)
+                        return false;
+                    index++;
+                }
+            } else {
+                for (const IR::StructField* f : strct->fields) {
+                    if (f->type->is<IR::Type_Varbits>() || f->type->is<IR::Type_Stack>()
+                        || f->type->is<IR::Type_HeaderUnion>()) {
+                        TypeInference::typeError("%1%: Structs, headers and tuples "
```
mbudiu-vmw: same error as above

frontends/p4/typeChecking/typeUnification.cpp
```diff
@@ -303,64 +302,142 @@ bool TypeUnification::unify(const IR::Node* errorPosition,
...
+                    if (strctField == nullptr) {
+                        TypeInference::typeError("%1%: No field named %2% in structure %3%",
+                                                 errorPosition, f->name, strct);
+                        return false;
+                    }
+                }
+                for (const IR::StructField* f : strct->fields) {
+                    if (f->type->is<IR::Type_Varbits>() || f->type->is<IR::Type_Stack>()
+                        || f->type->is<IR::Type_HeaderUnion>()) {
+                        TypeInference::typeError("%1%: Structs, headers and tuples "
```
mbudiu-vmw: use the same error message.

frontends/p4/typeChecking/typeUnification.cpp
```diff
@@ -303,64 +302,142 @@ bool TypeUnification::unify(const IR::Node* errorPosition,
...
+                                                 "containing %2% data type as a lief "
+                                                 "field cannot be zero initialized with "
```
mbudiu-vmw: haven't looked very carefully, but there is a fair amount of code duplication, please see if you can refactor the common code in a function.

frontends/parsers/p4/p4parser.ypp
```diff
@@ -1260,6 +1261,7 @@ optInitializer
 
 initializer
     : expression                          { $$ = $1; }
+    | "..."                               { $$ = new IR::ListExpression(@1, *(new IR::Vector<IR::Expression>()), true); }
```
mbudiu-vmw: I am a little uncomfortable that this is a list expression. What if this is assigned to a scalar?
(Do we allow scalars to be initialized this way? `bit<32> b = ...;`

ir/expression.def
```diff
@@ -384,14 +384,28 @@ class ConstructorCallExpression : Expression {
 
 /// Represents a list of expressions separated by commas
 class ListExpression : Expression {
+    bool defaultInitializer = false;
+    ListExpression(Util::SourceInfo srcInfo, IR::Vector<IR::Expression>& components, bool convenientInit) : Expression(srcInfo), components(components){
+        defaultInitializer = convenientInit;
+        validate();
```
mbudiu-vmw: can you please move the duplicated code to a helper function?

ir/expression.def
```diff
@@ -400,6 +414,19 @@ class ListExpression : Expression {
 /// IR representation of an initializer for a struct.
 /// Currently it does not have a direct representation as P4 source.
 class StructInitializerExpression : Expression {
+    bool defaultInitializer = false;
+    StructInitializerExpression(Util::SourceInfo srcInfo, const IR::Type* type, const IR::Type_Name* typeName, IR::IndexedVector<IR::NamedExpression>& components, bool convenientInit)
+                                : StructInitializerExpression(srcInfo, type, typeName, components) {
+        defaultInitializer = convenientInit;
+
```
mbudiu-vmw: no empty lines

ir/type.def
```diff
@@ -268,6 +268,16 @@ class Type_Struct : Type_StructLike {
 /// and some information about their types.
 class Type_UnknownStruct : Type_StructLike {
 #nodbprint
+    bool fromDefaultInitializer = false;
```
mbudiu-vmw: i think you should use a consistent name everywhere. Just use "defaultInitializer" for all fields and constructor arguments.

ir/expression.def
```diff
@@ -384,14 +384,28 @@ class ConstructorCallExpression : Expression {
 
 /// Represents a list of expressions separated by commas
 class ListExpression : Expression {
+    bool defaultInitializer = false;
+    ListExpression(Util::SourceInfo srcInfo, IR::Vector<IR::Expression>& components, bool convenientInit) : Expression(srcInfo), components(components){
```
mbudiu-vmw: Can you also set defaultInitializer after a comma instead of doing it in the body of the constructor?
This is our consistent style. 
This applies to all other changes to the def files.

mbudiu-vmw:
I don't know if this allow structure-valued expressions with default initializers. But perhaps we don't want that. A test case is something like `(a == S { x = 2, ...))`, where `a` is a structure of type `S`.
Also, can you add the test case for #2213 if it is indeed fixed?
This is great!

frontends/parsers/p4/p4parser.ypp
```diff
@@ -207,6 +207,7 @@ inline std::ostream& operator<<(std::ostream& out, const P4::Token& t) {
 %token<Token>      PP "++"
 %token<Token>      DONTCARE "_"
 %token<Token>      MASK "&&&"
+%token<Token>      CONVINIENT_INITIALIZER "..."
```
jafingerhut: Typo?  There is an English word "convenient", but no English word "convinient".  Not sure if the intent is to be an English word or not, though.

jafingerhut:
Are there existing test programs that attempt to use a struct initializer to initialize a header/struct, but should be a compile-time error because they do not have `...`, and do not assign values to all fields?

If not, should this PR add some, with expected error outputs?

testdata/p4_16_samples/convenient_init.p4
```diff
@@ -0,0 +1,47 @@
...
+
+
+extern void f<D>(in D data);
+control c(inout bit<1> r) {
+    S s_0;
+    H h_0;
+    bit<1> tmp;
+    apply {
+        S s_1;
+        s_1 = { b = false, h={3, true, ...}, ...};
```
jafingerhut: I would like to please request that this test program be changed slightly to make it easier for a person to check the program against the frontend and midend versions, for example by making the following changes in variable names:

+ s_0 -> sa
+ s_1 -> sb
+ h_0 -> ha
+ h -> hb

With the suffixes like `_0` and `_1`, the automatic renaming that some compiler passes do can make it very difficult to see how the renamed variables correspond with the original ones.

marko-stanoj:
Thank you everyone for your time and reviews, I will try to follow suggestions and make these changes as soon as possible.

frontends/p4/structInitializers.cpp
marko-stanoj: If there are types that we don't allow as a field, compiler will report an error in typeUnification.cpp before StructInitializers pass, so the only types that could be handled in the else branch are header, struct or tuple. But you are right, if some new data types are introduced into P4, this could be a problem.

frontends/parsers/p4/p4parser.ypp
marko-stanoj: This solution doesn't allow default initializer for scalar types. Using `...` makes initializing structs, headers and tuples easier and notation shorter, but I don't see any benefits of using it for scalar types. If you think it has a use case I can try to implement it for scalar types too.

frontends/p4/structInitializers.cpp
marko-stanoj: Is it possible for enum not to have at least 1 member? According to parser it should be a syntax error, am I right?

marko-stanoj:
Unfortunately #2213 is not fixed, but there's no compiler bug when default initializer is used instead of a regular struct initializer expression.

marko-stanoj:
> I don't know if this allow structure-valued expressions with default initializers. But perhaps we don't want that. A test case is something like `(a == S { x = 2, ...))`, where `a` is a structure of type `S`.

Currently this example produces a syntax error because it is a production of `expression: typeName "{" kvList "}"` and `kvList` cannot contain dots, they can only be part of expression or initializer. Do you want me to change something about that?

mbudiu-vmw:
Let's leave structure-valued expressions for later.

frontends/p4/structInitializers.h
```diff
@@ -33,17 +34,36 @@ class CreateStructInitializers : public Transform {
...
         setName("StructInitializers");
         passes.push_back(new TypeChecking(refMap, typeMap));
-        passes.push_back(new CreateStructInitializers(refMap, typeMap));
+        if (doProcessAssignments == false) {
+            passes.push_back(new CreateStructInitializers(refMap, typeMap));
+        } else {
+            passes.push_back(new CreateStructAssignInitializers(refMap, typeMap));
+            passes.push_back(new ResolveReferences(refMap));
+            // may insert new constants
+            passes.push_back(new TypeInference(refMap, typeMap, false));
```
mbudiu-vmw: But aren't all inserted constants having the correct type already?
I suspect this may not be necessary. Can you try without it as well? If I am right, the ResolveReferences is not necessary either.

marko-stanoj: When I try to compile programs in which mix of explicit
values and default values are used (e.g ` h={3, true, ...}`), without those two additional passes I get compiler bug:
p4c/frontends/p4/typeChecking/typeChecker.cpp:144

I suppose that is because constants used for explicit initialization of fields don't have the correct type already. So ResolveReferences and TypeInference are needed, am I right?

mbudiu-vmw: Then you should add a comment that these passes are useful not for the newly introduced initializers, but for the existing ones. I was assuming these had already been processed, but I guess the first instance of StructIntializers with a 'false' flag no longer does that.

marko-stanoj: They are useful only if a mix of explicit and default values are used, because type unification of struct fields and explicit values in list expression listed before `...` is postponed after StructInitializers pass (with true flag). That is why those constants do not have correct type already. 
Should I try to make some changes in typeUnification.cpp and make these passes unnecessary again, or should I leave it like this (with added comment, as you said)?

mbudiu-vmw: The important thing is for passes in the front-end to use the correct assumptions; none of the passes before the second invocation of StructInitializers should assume that constants have been cast implicitly, for example. In general we try to minimize the number of assumptions that pass writers have to understand, so if there could be only one version of StructInitializers it would be better - but if the trade-off is increased complexity then it may not be worth it.

mbudiu-vmw:
If no one has any more comments I think we should merge this.

jafingerhut:
Can anyone summarize, perhaps with a list of example assignments, how this behaves in terms of assigning headers that are valid, versus forcing the header assigned to to become invalid?

marko-stanoj:
Currently this doesn't force headers to become invalid. Default value for header is header of same type where each field has the default value of the suitable field type -- if all such default values are defined.

But as I said in the commit message:
"Default value for header type could be easily changed to invalid, as proposed
in p4-spec pull request #797, when a new literal for initializing header to invalid is introduced."

jafingerhut:
@mar-stanoj Is it easy in the parsing code, and does it make sense, to make it work as follows?

`my_hdr = ...;` makes `my_hdr` become invalid (which according to the current spec, means all header fields are undefined)

`my_hdr = { ... };` makes `my_hdr` become valid, with default values for all fields.

The `{ ... }` syntax for a valid header, with default values for all field, seems like a reasonable generalization of `{f1=5, ...}`, which is default for all fields except `f1`.

marko-stanoj:
@jafingerhut It does make sense, and in my opinion is the best solution, but when new construct which enables initializing headers to invalid is introduced (as it was mentioned on last LDWG meeting).
Right now, with current specification, it is impossible to initialize header, or assign value to header and leave the header invalid. So it does not make sense for me to allow that by introducing `...` , which should just represent shorten notation of initialization.

jafingerhut:
@marko-stanoj I have added this comment on the pull request to add default values to the language specification, which if accepted by the language design work group, may be what you are hoping for?  https://github.com/p4lang/p4-spec/pull/797#discussion_r396023841

mbudiu-vmw:
While there is no expression denoting an uninitialized header, it does not mean that the compiler cannot expand such intializers. They would have to be expanded to a series of calls to setInvalid, which is considerably more complicated..

marko-stanoj:
As far as I know headers are invalid by default, when declared. Wouldn't transforming

`Hdr hdr = ... ; `

into:
```
Hdr hdr;
hdr.setInvalid(); 
```
where Hdr is header type be redundant?

But if header is field of a struct:
```
struct S {
    error       errorNoOne;
    Hdr           hdr;
    bool        b;
    int<8>      i;
}
/* some P4 code */
S s = { errorNoOne = error.NoError, b= false, i = 0, ... };
/* some P4 code */
```
what should be used to replace the initializer for hdr, if we want it to stay invalid?

jafingerhut:
In both cases, making the header become invalid is correct.  Redundant things are an optimizing compiler's job to eliminate.  These assignments are not _only_ for initialization, though, correct?  We could make one of these assignments to a header that is currently valid, and then it is not a no-op.

mbudiu-vmw:
The only way to implement with our current syntax it is to blow up the initializer into a series of assignments to the various fields, skipping the invalid header.

marko-stanoj:
> The only way to implement with our current syntax it is to blow up the initializer into a series of assignments to the various fields, skipping the invalid header.

That would work if we use `...` for initialization, but what if we are using `...` inside curly brackets (can be used for assignments too), header doesn't have to be invalid in that case?

marko-stanoj:
I suppose this solution is good compromise solution until construct that allows initializing headers to invalid is introduced. Then it will be possible to make it work like @jafingerhut proposed with very few changes, keeping solution simple and readable.

marko-stanoj:
Is there any news on this topic? Should I resolve conflicts that have appeared meanwhile or you came up with different idea to implement default initializers?

mbudiu-vmw:
The main question is how does this align with the current P4 spec proposal.
If they are the same, then we can probably merge this.
If not, we should change one of the two so that they are in alignment. If we need to change the spec we should discuss it at our next design meeting. Can you please summarize the state of the world at this point?

marko-stanoj:
The state of the world at this point (things that are not in conformance with spec proposal):
- dots are used only for initialization of struct, headers and tuples, not scalar types
- default value of header is header of same type where each field has the default value of the 
suitable field type (if there is a plan for introducing literal which would initialize headers to invalid, default value could be easily changed to invalid too)
- default values of varbit<N>, headers stack and header union are undefined, those types are not supported as a struct or tuple field if dots are used

mbudiu-vmw:
We discussed this in the language design working group today. The agreement is that ideally we should support initialization to headers and unions too. This cannot be done solely as you have done it. First, it cannot be done as initializers, is has to be done only as statements. We already have a pass moveInitializers which moves initializers into statements; this one should come after that. Then, for headers and unions it would have to generate setInvalid calls as consecutive statements. Are you interested to adapt your PR?

marko-stanoj:
As you can see it is done as statements, all the processing is in function ```const 
IR::Expression*
convert(const IR::Expression* expression, const IR::Type* type
)``` which is called from  `const IR::Node* CreateStructAssignInitializers::postorder(IR::AssignmentStatement* statement)`. That's the reason I moved the processing after moveInitializers pass. For headers and header unions, all that needs to be done is to generate adequate number of `setInvalid()` calls and emplace ` h = ... ` and `hu = ...` with it, is that right? (h stands for header type and hu for header union) 

Problem remains if header union is a field of a struct. Logic behind my solution is to fill struct initializer expression with default values of corresponding fields. If header type is struct field, I can initialize it with list expression* of random values of adequate type, generate `setInvalid()` for the header after struct initialization, and get desired effect. But if header union is a struct field there's no literal to initialize it all. Do you have idea how to solve that problem?

*that list expression is a component of struct initializer expression

mbudiu-vmw:
Since we don't have a literal to stand for an invalid header, I am guessing that the only thing to do is to actually generate a sequence of assignment statements for all structs that contain headers or unions, assigning to each field separately, but for the header just calling setInvalid.

jafingerhut:
I would encourage some kind of implementation that leaves out header_union support for now, but solves all of the other issues.

Reason: The P4_16 language spec does not currently define any way to assign a value to a variable `hu` with a header_union type, without also explicitly referring to one of the headers that is a member of `hu`.  For example, you can write `hu.h1 = { 5, 7 };` if `h1` is a member of header_union `hu` with 2 `bit<16>` fields, but there is no syntax like `hu = { 5, 7 };`.  Until and unless some syntax is defined in the spec that can go on the right-hand side of an assignment to `hu`, there is also no syntax for what can go on the right hand side of an assignment to a struct that contains a header_union as one of its fields.

I do not think the implementation of all of the other good things in this PR should wait for the P4 language design work group to design that.  If you have proposals for it, great, and perhaps p4c devs would not mind an experimental solution for it, but it seems best to me to have an implementation of everything but header_union, earlier.

mbudiu-vmw:
The problem is that his solution doesn't work for headers either, because you cannot say `h = something` for making a header invalid. So it's exactly the same for headers and unions; for unions you have to do `f.setInvalid()` for all fields, and for headers you have to do `h.setInvalid()`.

jafingerhut:
I agree that the problem of making a header or a header_union member of a struct invalid is the same for both.

I also believe that for making a header_union member of a struct _valid_ using a single assignment with list syntax on the right hand side, the problem I mentioned earlier exists -- there is no agreed-upon way to do this in the specification.

That is why I believe we should move forward with a change to p4c to make all of this work as proposed for everything except header_union, and wait on header_union support until after some syntax for that has been worked out.  I'm not saying it is difficult to create a syntax for that -- just that it hasn't been done yet.

jafingerhut:
Also, as an aside that need not affect this PR contents at all if it doesn't support header_union yet (as I hope it need not to be merged in), one does not have to do setInvalid on all members of a header_union in order to make it invalid.  You only have to do it on one.  This is a semi-subtle point, which I only know about because I suggested changing it so that one need not have to make all header_union members invalid in order to make the header_union variable invalid -- that just seems like a waste of unnecessary operations to me (and others who approved the change 3 years ago, it seems).

marko-stanoj:
> The problem is that his solution doesn't work for headers either, because you cannot say `h = something` for making a header invalid. So it's exactly the same for headers and unions; for unions you have to do `f.setInvalid()` for all fields, and for headers you have to do `h.setInvalid()`.

That is true, but if we have something like this:
```
header Hdr {
    bit<8>      f;
}

struct S {
    Hdr           hdr;
    bool          f1;
    int<8>        f2;
}
/* some P4 code */
S s;
/* some P4 code */
s = { ... };
/* some P4 code */
```
I can transform that assignment statement into:
```
S = {hdr = {8w0}, f1 = false, f2 = 8s0};
S.hdr.setInvalid( );
```
which would result in desired effect, wouldn't it?

mbudiu-vmw:
Seems a bit wasteful to initialize invalid fields; the compiler may even give warnings.
Probably the right way is to use a struct expression if there are no header fields, and otherwise to explode the initialization into many separate statements. Note that we already have some passes that do similar things, like the copyStructures pass.

marko-stanoj:
OK, I will try to implement it as you proposed. Should we retain functionality to initialize headers with mix of explicit and default values?

jafingerhut:
@marko-stanoj You asked: "Should we retain functionality to initialize headers with mix of explicit and default values?"

If you mean that if there is a header type defined like this:
```
header ethernet_t {
bit<48> dstAddr;
bit<48> srcAddr;
bit<16> etherType;
}
```
and then later an assignment like this:
```
ethernet_t h1;

h1 = { 5, ... };
```
Then I am pretty sure that the intent of the change in the language specification is that after that assignment, `h1` should be a valid header with field dstAddr equal to 5, and the other two fields equal to the default value of 0.

marko-stanoj:
That is exactly what I meant, thank you.

marko-stanoj:
As far as I know there is no possibility of assigning single field of a tuple. If I'm right, there is no way to explode the initialization of tuple into many separate statements, which is behavior we want for headers and structs. If you share my opinion, do you agree to forbid assigning or initializing tuples with `...` for now?

P.S. Sorry for not implementing the solution yet, I will try to come up with it by the end of the week.

mbudiu-vmw:
Yes, we cannot assign individual tuple elements. For a tuple you would have to create temporaries and then use a list expression to assign the whole tuple - at least in the case when the tuple has headers as elements.

But I think we should add tuple field access as a feature of the language. Maybe I will write up something about this.

mbudiu-vmw:
We now have a syntax to access tuple elements, but tuples still have to be initialized entirely - fields are not left values.
(Also, we haven't merged that yet.)
We should try to make progress on merging this PR, since it's already in the spec.
One thing we can do is to add in the spec some syntax to initialize stacks and invalid headers, and then this task becomes significantly easier.

mbudiu-vmw:
Superseded by #3968

PR #2227, opened 2020-03-04
Support for structured annotations

mbudiu-vmw:
This implements the spec PR https://github.com/p4lang/p4-spec/pull/796/files

@jafingerhut, @stefanheule, @chrispsommers, @antoninbas : please take a look, especially at the tests.

Since I didn't want to change the reference outputs, all annotations that were already being used have been marked as unstructured, including `@name` annotations. If we want to change this we can do it in a separate pull request.


jafingerhut:
To check my understanding, here is what I think is true:

+ If in the source code, @annotationName is followed by a left parenthesis `(`, then it is always an unstructured annotation, regardless of the characters of the name string following the `@` symbol.
+ If the next character is `[`, it is always a structured annotation.
+ If the annotation name has neither a `(` nor `[` character after it, then I do not know if the spec explicitly says which it should be, but for backwards compatibility sake it should probably be an unstructured annotation with an empty body?

Thus changing `@name` annotations, which are always followed by `(`, from unstructured to structured, would violate the currently proposed spec for structured annotations.  We could change the spec further, of course, so that whether an annotation was structure or unstructured depended not only on the character being `(` or `[`, but also on the contents of the annotation name, but I have not heard anyone yet proposing that.

testdata/p4_16_errors/structured-annotation-e.p4
```diff
@@ -0,0 +1,4 @@
+@IllegalMixing[key=4, 5]
```
jafingerhut: There is no expected output file for this test program?

testdata/p4_16_errors/structured-annotation-e2-ConstantFolding_0_DoConstantFolding.p4
```diff
@@ -0,0 +1,5 @@
+@badType[a={ 2, 3 }] control c() {
```
jafingerhut: Also no expected output for this test program, nor the one named structured-annotation-e2-FrontEnd_5_ConstantFolding.p4

testdata/p4_16_samples_outputs/structured-annotation.p4-stderr
```diff
@@ -0,0 +1 @@
+warning: Program does not contain a `main' module
```
jafingerhut: There is no test program that contains a generated P4Info file.  Is it planned that one will be added later after appropriate changes are made in the PI project on which p4c depends?  Or should that be possible to create now?

jafingerhut: Answering my own comment: I am pretty sure that this PR, or something that similarly updates the .proto files defining the allowed contents of the P4Info files, would need to be merged in first, before it would be possible to write an automated test for p4c that checks that P4Info files are written by p4c with the contents of structured annotations: https://github.com/p4lang/p4runtime/pull/265

So any such tests for p4c will have to wait for that merging to be done first.

antoninbas: @jafingerhut the plan is to separate IR changes (which Mihai is taking care of) from the P4Info generation changes (which I will take care of).

frontends/p4/typeChecking/typeChecker.cpp
```diff
@@ -760,6 +760,29 @@ TypeInference::assignment(const IR::Node* errorPosition, const IR::Type* destTyp
...
 
+const IR::Node* TypeInference::postorder(IR::Annotation* annotation) {
+    auto checkAnnotation = [this] (const IR::Expression* e) {
+        if (!isCompileTimeConstant(e))
+            typeError("%1%: structured annotation must be compile-time constant values", e);
+        auto t = getType(e);
+        if (!t->is<IR::Type_InfInt>() &&
+            !t->is<IR::Type_String>() &&
+            !t->is<IR::Type_Boolean>())
+            typeError("%1%: illegal type for structured annotation; must be int, string or bool", e);
```
chrispsommers: Do we want to be more specific about the `int` type in the error message, i.e. specify infinite-precision?

mbudiu-vmw:
`int` is infinite precision

frontends/parsers/p4/p4parser.ypp
```diff
@@ -459,10 +459,14 @@ annotation
           $$ = new IR::Annotation(@1, *$2, body); }
     | "@" name "(" annotationBody ")"
         { $$ = new IR::Annotation(@1, *$2, *$4); }
```
chrispsommers: Do you need to add `, false` to args?

mbudiu-vmw: If your body is a sequence of tokens then by definition it's not structured. This constructor will set `structured` to false.

testdata/p4_16_errors/structured-annotation-e.p4
mbudiu-vmw: Error programs may not have outputs, depending on where the compiler exits.

testdata/p4_16_errors/structured-annotation-e2-ConstantFolding_0_DoConstantFolding.p4
mbudiu-vmw: This one shouldn't be here.

frontends/parsers/p4/p4parser.ypp
chrispsommers: OK,  thanks.

testdata/p4_16_errors/structured-annotation-e.p4
jafingerhut: This test program does have an error output, when I try it out with your changes on my local dev machine:
```
$ ~/forks/p4c/build/p4test structured-annotation-e.p4 
structured-annotation-e.p4(1):syntax error, unexpected INTEGER
@IllegalMixing[key=4, 5
                      ^
```

testdata/p4_16_samples_outputs/structured-annotation.p4-stderr
jafingerhut: Sounds good.  Resolving this conversation on Github.

ir/base.def
```diff
@@ -284,20 +295,20 @@ class Annotations {
         annotations.push_back(annot);
         return this; }
     Annotations add(Annotation annot) const { return clone()->add(annot); }
-    Annotations addAnnotation(cstring name, Expression expr) {
-        return add(new Annotation(name, Vector<Expression>(expr))); }
-    Annotations addAnnotation(cstring name, Expression expr) const {
-        return add(new Annotation(name, Vector<Expression>(expr))); }
+    Annotations addAnnotation(cstring name, Expression expr, bool structured) {
```
antoninbas: You didn't want to use a default value of `false` for these methods to avoid having to add `false` in all the existing code?

frontends/parsers/p4/p4parser.ypp
```diff
@@ -459,10 +459,14 @@ annotation
...
+    | "@" name "[" expressionList "]"
+        { $$ = new IR::Annotation(@1, *$2, *$4, true); }
+    | "@" name "[" kvList "]"
+        { $$ = new IR::Annotation(@1, *$2, *$4, true); }
```
antoninbas: I though we would go for a enum with 3 possible values: `UNSTRUCTURED`, `EXPRESSION_LIST`, `KV_LIST`, but I see that you opted for a boolean. I think it makes sense because for every structured annotation you set `needsParsing` to `true`, which means that exactly one of these fields will not be empty:
```

    inline IndexedVector<NamedExpression> kv;
```
unless the annotation body is empty (`@foo[]`), in which case both of these vectors will be empty.

Please let me know if I got this wrong. The P4Info-generation code needs to be able to "easily" distinguish between both cases.

ir/base.def
mbudiu-vmw: There are lots of overloads, and the preprocessor we wrote generates even more, so I wasn't sure which one will get used.

frontends/parsers/p4/p4parser.ypp
mbudiu-vmw: If it is structured then only one of the two can be non-empty.
If both are empty it's the kv that is assumed to be used.
needsParsing should be false for structured annotations. It can only even be true for unstructured annotations. I don't think you need to look at it.
I can add a getter that would return one of these three if that makes your job easier.

testdata/p4_16_samples/structured-annotation.p4
```diff
@@ -0,0 +1,13 @@
+#define TEXT_CONST "hello"
+#define NUM_CONST 6
+
+@number[2]
+@empty[]
+@kv[k1=1,k2=2]
+@number(3)
```
liujed-intel: This program has `number` as both a structured and an unstructured annotation on the control `c`. This should be an error according to the intent of the proposed spec change p4lang/p4-spec#796.

ir/base.def
```diff
@@ -266,6 +285,14 @@ class Annotation {
 
     /// Useful helper: extracts name value from a name annotation
     static cstring getName(const Annotation* annotation);
+
+    AnnotationKind annotationKind() const {
```
mbudiu-vmw: @antoninbas : here is a helper function which may be useful.

testdata/p4_16_samples/structured-annotation.p4
chrispsommers: @liujed The intent of the spec change was to disallow  duplicate use of names  on a given *structured* annotation. Unstructured annotations have no such restriction. So `@number[2]` and `@number(3)` should be considered as non-colliding. By that reasoning, it would be OK to have another unstructured annotation in this program `number(4)` but not OK to add another structured annotation `number[5]`

mbudiu-vmw: This is how the implementation was, but your last comment made it look like it's wrong, so I just changed it in my last PR.

liujed-intel: @chrispsommers, as Mihai says, your comment here suggests otherwise: https://github.com/p4lang/p4-spec/pull/796#issuecomment-598002044

mbudiu-vmw:
@liujed : I think that this is now conformant with the spirit of the spec - if not the letter.
I have reverted to allowing duplicate names if only one of them is structured.

ir/base.def
antoninbas: based on the latest proposals for both specifications, it seems that we really have 4 cases: unstructured, structured empty, structured kv list and structured list. However, I think the implementation can be adjusted later and we can merge this as it is.

PR #2233, opened 2020-03-11
Struct initializers should infer types for nested structs; fixes #2213

### LINKED ISSUE #2213: Compiler bug when attempting to compile nested struct initializer

mbudiu-vmw:
This is literally a (half) line fix.

### DIFF
--- a/testdata/p4_16_samples/issue2213.p4
+++ b/testdata/p4_16_samples/issue2213.p4
@@ -1,0 +1,24 @@
+struct mystruct1_t {
+    bit<16> f1;
+}
+
+struct mystruct2_t {
+    mystruct1_t s1;
+}
+
+struct metadata_t {
+    mystruct1_t s1;
+    mystruct2_t s2;
+}
+
+control ingressImpl(inout metadata_t meta)
+{
+    apply {
+        meta.s2 = {s1={f1=2}};
+    }
+}
+
+control c(inout metadata_t meta);
+package top(c _c);
+
+top(ingressImpl()) main;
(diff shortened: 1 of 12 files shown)


testdata/p4_16_samples/issue2213.p4
```diff
@@ -0,0 +1,24 @@
...
+    mystruct1_t s1;
+    mystruct2_t s2;
+}
+
+control ingressImpl(inout metadata_t meta)
+{
+    apply {
+        meta.s2 = {s1={f1=2}};
+    }
+}
```
marko-stanoj: If you change ingressImpl into:
```
control ingressImpl(inout metadata_t meta)
{
    mystruct2_t helper = { s1 = { f1 = 2 }};
    apply {
        meta.s2 = helper;
    }
}
```
you would still get a compiler bug, so this solution fixes only assignments not declarations.

PR #2238, opened 2020-03-12
Remove a stray print

fruffy:
This print snuck in and should not be there.


mbudiu-vmw:
You will need to rebase on master for this to pass.

PR #2240, opened 2020-03-13
Fix type_info generation

### LINKED ISSUE #2239: Empty type_info section of P4Info file, when it should be non-empty
The two attached example programs both cause the indicated version of p4c (which includes the fix in recent PR https://github.com/p4lang/p4c/pull/2235 ) to generate a P4Info file with an empty `type_info` section, when that section should contain the definition for at least one type `IPv4Addr_t` used as the type of a table key field, or an action parameter (different case for each test program).
```
$ p4c --version
p4c 1.2.0 (SHA: a1f21cb9)
```
[custom-type-table-key.p4.txt](https://github.com/p4lang/p4c/files/4327139/custom-type-table-key.p4.txt)
[custom-type-action-parameter.p4.txt](https://github.com/p4lang/p4c/files/4327140/custom-type-action-parameter.p4.txt)

antoninbas:
P4Runtime v1 resticts the type of some fields (match fields, action
parameters, controller metadata fields) to unsigned bitstrings. When a
user-defined type is used for one of these, P4Info will use the
underlying type, but will also populate a type_name field in the
Protobuf message so that P4Info consummers can be aware of the
user-defined type. The type_name can then be looked-up in the type_info
"database" of types. This commit fixes an issue which caused type_info
not to be populated correctly. It also makes sure that type_name is set
properly for counter / meter / register indices.

Fixes #2239

### DIFF
--- a/control-plane/p4RuntimeArchHandler.h
+++ b/control-plane/p4RuntimeArchHandler.h
@@ -32,6 +32,7 @@ limitations under the License.
 #include "frontends/p4/typeMap.h"
 #include "ir/ir.h"
 #include "lib/ordered_set.h"
+#include "typeSpecConverter.h"
 
 namespace P4 {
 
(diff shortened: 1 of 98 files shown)


antoninbas:
@jafingerhut correct, the `type_name` (match fields, action params, ...) does not apply to serializable enums.

I thought it would be a quick fix when I saw the issue, and then I realized I had to support `type_name` for counter / meter / register indices as well, which was definitely more work...

PR #2243, opened 2020-03-15
Ensure that ID prefixes are correct in P4Info when '@id' is used

### LINKED ISSUE #2241: Incorrect P4Info id generation
In https://github.com/p4lang/p4runtime/issues/268 I noticed that the compiler is not generating P4Info correctly when the programmer uses the `@id` annotation.  The compiler should fill in the top 8 bits with the object type if those bits aren't already present, and if they are present, check that they are correct. Instead, the compiler seems to just use the value from the annotation without adding those bits. https://github.com/p4lang/p4runtime/issues/268 has a testcase.

antoninbas:
As per the P4Runtime specification, when the '@id' annotation is used:
  * if a prefix is missing (i.e. it is 0), assign the correct prefix
    based on the resource type.
  * if a prefix is present (i.e. not 0), ensure that the provided value
    is correct for the resource type - this is not part of P4Runtime
    v1.1, but is being added to the unreleased v1.2 specification.

Fixes #2241

### DIFF
--- a/control-plane/p4RuntimeSerializer.cpp
+++ b/control-plane/p4RuntimeSerializer.cpp
@@ -457,8 +457,8 @@ class P4RuntimeSymbolTable : public P4RuntimeSymbolTableIface {
     void computeIdsForSymbols(P4RuntimeSymbolType type) {
         // The id for most resources follows a standard format:
         //
-        //   [resource type] [zero byte] [name hash value]
-        //    \____8_b____/   \__8_b__/   \_____16_b____/
+        //   [resource type] [name hash value]
+        //    \____8_b____/   \_____24_b____/
         auto& symbolTable = symbolTables.at(type);
         auto resourceType = static_cast<p4rt_id_t>(type);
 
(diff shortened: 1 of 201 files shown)


jafingerhut:
This looks good to me from what I know of the desired changes (which is excellent), and the implementation code (which is not good).

Is this PR an appropriate place to add tests for, or change the current behavior, where bits [23:16] of an id are zeroed out?  Described in more detail in this comment: https://github.com/p4lang/p4runtime/issues/268#issuecomment-596791315

antoninbas:
@jafingerhut these bits were only zeroed out for generated ids, not programmer-provided ids. This was for legacy reasons which no longer apply: originally in PI, the id for a match field was `<match field type> (8b) ++ <table id without prefix> (16b) ++ <offset in table match key> (8b)`, which meant that an id (without prefix) for a table (or an action, the same applies to action params) could only be 16 bits. Because this no longer applies today (match fields and action params can be arbitrary), and in order to be consistent with programmer-provided ids, it makes sense to use 24 bits (coming from a hash value computed on the name) instead of 16. I added a commit to this PR to implement this.

PR #2255, opened 2020-03-21
Better syntax for structure-valued expressions

mbudiu-vmw:
This simplifies the P4 grammar, removing the rule that allowed one to write:
```
struct S = { bit a; };
S s = S { a = 0 };
```

The way to do this is now using the cast syntax:

```
struct S = { bit a; };
S s = (S) { a = 0 };
```

This is already allowed by the grammar. Since we do not allow casts to structs, the typechecker converts casts to structs to StructureInitializerExpressions, and from that point on the IR is exactly as it was before.


jnfoster:
From the Travis error logs, looks like just a `cpplint` issue? Otherwise looks good!

PR #2268, opened 2020-03-26
fix ninja builds

boblytton:
Hi,
My ninja build stopped working with:
    ninja: error: build.ninja:1517: bad $-escape (literal $ must be written as 19185)
The fix was to add the second '$' in the CMakeList.txt file.

### DIFF
--- a/backends/ubpf/CMakeLists.txt
+++ b/backends/ubpf/CMakeLists.txt
@@ -57,7 +57,7 @@ set(P4C_UBPF_HEADERS
 
 set (P4C_UBPF_DIST_HEADERS p4include/ubpf_model.p4)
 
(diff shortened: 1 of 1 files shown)


backends/ubpf/CMakeLists.txt
```diff
@@ -57,7 +57,7 @@ set(P4C_UBPF_HEADERS
 
 set (P4C_UBPF_DIST_HEADERS p4include/ubpf_model.p4)
 
-add_cpplint_files(${CMAKE_CURRENT_SOURCE_DIR} "$(P4C_UBPF_SOURCES)")
+add_cpplint_files(${CMAKE_CURRENT_SOURCE_DIR} "$$(P4C_UBPF_SOURCES)")
```
cc10512: The right solution is to replace the '()' with '{}' and keep the one '$'.

PR #2272, opened 2020-03-27
Disable checkOutParams on functions due to breakge

ChrisDodd:
- add virtual4.p4 test that shows breakage

### DIFF
--- a/frontends/p4/simplifyDefUse.cpp
+++ b/frontends/p4/simplifyDefUse.cpp
@@ -200,8 +200,11 @@ class FindUninitialized : public Inspector {
         // We now check the out parameters using the definitions
         // produced *after* the function has completed.
         LOG3("Context after function " << currentPoint);
+#if 0
+        // FIXME -- doesn't work in the presence of abstract functions
         auto current = getCurrentDefinitions();
         checkOutParameters(func, func->type->parameters, current);
+#endif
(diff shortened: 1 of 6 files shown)


ChrisDodd:
I ran into this problem trying to update our internal code with the latest p4c code; we use abstract functions on externs quite a bit,  I was able to extract this down to a test case I could include in the open source code, but was not able to figure out exactly what was going wrong -- it seems to be having trouble with the arguments to the abstract function implementations

mbudiu-vmw:
You have an `@optional` parameters, including `out` parameters, for a function that you implement. I don't really know what this means; the spec only allows optional parameters for extern functions or methods, but then, abstract methods are not yet in the spec.

That's not the problem in this program, but it's something we should probably discuss in the design group.

mbudiu-vmw:
The problem is the following: the def-use pass makes a context-sensitive analysis for function calls, and it uses a not-very-clearly-specified method of representing program points where various definitions are alive. In this example some method are analyzed repeatedly, in different context (because of the `@synchronous` annotation), and the results of one analysis overwrite the results of the other one, because the context is not clearly represented. I don't have a fix yet, but hopefully it won't take too long.

mbudiu-vmw:
I will assume we can close this one.

PR #2274, opened 2020-03-27
Implement @id annotation support for non top-level objects

### LINKED ISSUE #2242: Implement use of @id annotation on action parameters, table match keys and packetio metadata
The P4RT working group has slightly expanded the use of the `@id` annotation beyond just the P4 objects (like tables, actions, etc), to the other entities in P4Info that get IDs, such that the programmer can also choose IDs for those.  The updated spec is here: https://github.com/p4lang/p4runtime/pull/272.

This is a feature request for the compiler to implement the change in the specification. Unlike for P4 Objects, there is no tag in the upper 8 bits for these, so it should be slightly simpler to implement (and in particular will not need the fix asked for in #2241).

antoninbas:
@id annotation support for match fields, action parameters and
controller metadata fields, as per the current P4Runtime specification
(dev version of P4Runtime 1.2.0).

Fixes #2242

### DIFF
--- a/control-plane/p4RuntimeSerializer.cpp
+++ b/control-plane/p4RuntimeSerializer.cpp
@@ -1206,8 +1206,12 @@ class P4RuntimeAnalyzer {
             match->set_bitwidth(et->width_bits());
             match->set_match_type(MatchField::MatchTypes::EXACT);
         } else if (et->is<IR::Type_Struct>()) {
-            int fieldId = 1;
-            for (auto f : et->to<IR::Type_Struct>()->fields) {
+            auto fields = et->to<IR::Type_Struct>()->fields;
+            // Allocate ids for all match fields, taking into account
+            // user-provided @id annotations if any.
+            FieldIdAllocator<decltype(fields)::value_type> idAllocator(
+                fields.begin(), fields.end());
+            for (auto f : fields) {
                 auto fType = f->type;
                 if (!fType->is<IR::Type_Bits>()) {
                     ::error(ErrorType::ERR_UNSUPPORTED,
@@ -1218,7 +1222,8 @@ class P4RuntimeAnalyzer {
                     continue;
                 }
                 auto* match = vs->add_match();
-                match->set_id(fieldId++);
+                auto fieldId = idAllocator.getId(f);
+                match->set_id(fieldId);
                 match->set_name(f->controlPlaneName());
                 match->set_bitwidth(fType->width_bits());
                 setMatchType(f, match);
--- a/control-plane/p4RuntimeSerializer.cpp
+++ b/control-plane/p4RuntimeSerializer.cpp
@@ -24,6 +24,7 @@ limitations under the License.
 
 #include <algorithm>
 #include <iostream>
+#include <iterator>
 #include <set>
 #include <typeinfo>
 #include <unordered_map>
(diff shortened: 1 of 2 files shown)


antoninbas:
@mbudiu-vmw I should be testing all the cases in my GTest test case (I just improved the test a bit, as I realized I had missed an edge case). Any other test you had in mind?

mbudiu-vmw:
I see them. I assumed that when you have an error the allocations don't happen anymore.
I didn't read the spec; is the header field f1 supposed to have one too?

antoninbas:
A lot of these google tests rely on the fact that even in the case of error, we still process other objects correctly. However, we don't really care about which id ends up being allocated when there is an error. In the case below:
```
            table t2 {
                key = {
                    m.f1: exact @id(99);
                    m.f2: exact @id(99);
                    m.f3: exact @id(0);
                }
                actions = { NoAction; }
            }
```
we expect the compiler to flag 2 errors and `m.f1` will probably receive id 99, but really the exact P4Info ids (or whether they are actually "correct") for `m.f1`, `m.f2` and `m.f3` don't matter.

antoninbas:
> I see them. I assumed that when you have an error the allocations don't happen anymore.
> I didn't read the spec; is the header field f1 supposed to have one too?

header fields don't have ids, unless they are part of a "controller header".

mbudiu-vmw:
f1 is in a controlled header.

antoninbas:
> f1 is in a controlled header.

ids for both controller header fields should be verified there:
```
    {
        // Check the ids for the packet-in header fields.
        auto* packetInHeader = findControllerHeader(*test, "packet_in");
        ASSERT_TRUE(packetInHeader != nullptr);
        const auto &m1 = packetInHeader->metadata(0);
        const auto &m2 = packetInHeader->metadata(1);
        EXPECT_NE(1u, m1.id());
        EXPECT_EQ(1u, m2.id());
    }
```
I don't use there actual names, but the P4Runtime terminology

jafingerhut:
Does the existing p4c test infrastructure enable writing a P4_16 source program where it is syntactically correct, and its only errors are in the values of `@id` annotations, presumably only caught in the P4Info generation back end code of p4c, and the error messages output are compared exactly against an expected output file?  I can try making such a thing on my local system in the p4_16_errors directory and see what happens, and report back.  That would enable negative test cases for this change in P4_16 source code format, which I know is in a way redundant with the Gtest tests, but perhaps still useful to point at as examples.

antoninbas:
@jafingerhut such tests can be placed under testdata/p4_16_errors; I believe the P4Info will be generated for these tests as well. However, in that case you are checking-in an invalid P4Info message, although still a "deterministic" one.

jafingerhut:
This question isn't necessarily 100% relevant for this PR, but it seems a semi-appropriate way to ask: What is the mechanism in the P4Info generation code to prevent two auto-generated ids from being identical to each other, assuming there is one?

cc10512:
The implementation of the `getId()` method in this interface: https://github.com/p4lang/p4c/blob/d7783e1036de5c1d2e6b65cdc2d1c3bc47db68ed/control-plane/p4RuntimeArchHandler.h#L114

antoninbas:
The `P4RuntimeSymbolTableIface` is for top-level P4 objects only. It is not used as part of this PR, as local ids work differently.

jafingerhut:
Does this include match fields for value_set's as well as tables?

antoninbas:
> Does this include match fields for value_set's as well as tables?

Good point, we forgot them in https://github.com/p4lang/p4runtime/pull/272.

I opened a new PR against the P4Runtime spec to add them (https://github.com/p4lang/p4runtime/pull/276) and updated this p4c PR accordingly.

PR #2231, opened 2020-03-10
Predication pass optimization for nested if statements (p4lang#2083)

matijasyrmia:
Refactored Predication pass, because of the issue p4lang#2083 in which LocalCopyProp pass
creates quadratically more complex code for chained if-else statements.

Predication pass is optimized by looking for same assignment statements in both then
and else branches and then converting them to the ternary expression if possible.
It will not be possible to convert more than one assignment if they have dependencies
between each other.

For the basic example with one if-else statement like this:
```
if(e)
	x = 5;
else
	x = 10;
```
Code will be transformed to this:

`x = e ? 5 : 10;`

New implementation of Predication pass relies on keeping track of the transformed
live assignments which will be inserted into the right place in IR, after visiting
both if and else statements of the corresponding block.

ExpressionReplacer transforms expression on the right side of the assignment into
ternary operator(IR::Mux) or nested ternary operators.

ReplaceChecker inspects if the assignment statement could be emplaced into the
nested ternary operator.

PR #2276, opened 2020-03-30
Predication fixes and cleanup

matijasyrmia:
This PR is follow-up of the #2231 
expr_name refactored to lvalue_name because it's used only for lvalue in every assignemnt.
visitThen and visitElse  merged to visitBranch with extra bool parameter.
Comments for functions fixed, unnecessary name generator removed.

### DIFF
--- a/midend/predication.cpp
+++ b/midend/predication.cpp
@@ -136,13 +136,10 @@ const IR::Node* Predication::clone(const IR::AssignmentStatement* statement) {
 const IR::Node* Predication::preorder(IR::AssignmentStatement* statement) {
     if (!inside_action || ifNestingLevel == 0)
         return statement;
-    auto context = getContext();
+    const Context * ctxt = nullptr;
     std::vector<const IR::Expression*> conditions;
-    while (context != nullptr) {
-        if (context->node->is<IR::IfStatement>()) {
-            conditions.push_back(context->node->to<IR::IfStatement>()->condition);
-        }
-        context = context->parent;
+    while (auto ifs = findContext<IR::IfStatement>(ctxt)) {
+        conditions.push_back(ifs->condition);
     }
     auto clonedStatement = clone(statement)->to<IR::AssignmentStatement>();
     ExpressionReplacer replacer(clonedStatement, travesalPath, conditions);
--- a/midend/predication.cpp
+++ b/midend/predication.cpp
@@ -193,7 +193,6 @@ const IR::Node* Predication::preorder(IR::IfStatement* statement) {
 
     ++ifNestingLevel;
     auto rv = new IR::BlockStatement;
-
     if (!statement->condition->is<IR::PathExpression>()) {
         cstring conditionName = generator->newName("cond");
         auto condDecl = new IR::Declaration_Variable(conditionName, IR::Type::Boolean::get());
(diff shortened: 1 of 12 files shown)


midend/predication.cpp
```diff
@@ -15,21 +15,26 @@ limitations under the License.
...
-                return base + "." + std::to_string(k->asInt()); }
+            if (auto base = lvalue_name(a->left))
+                return base + "." + std::to_string(k->asInt());
+    } else if (auto s = exp->to<IR::Slice>()) {
+        if (auto base = lvalue_name(s->e0))
+            if (auto h = s->e1->to<IR::Constant>())
+                if (auto l = s->e2->to<IR::Constant>())
+                    return base + "." +
+                        std::to_string(h->asInt()) + ":" + std::to_string(l->asInt());
+    }
```
mbudiu-vmw: If "" should never happen then you should make this a BUG_CHECK.

matijasyrmia: lvalue_name should return "" cstring for every expression that is not lvalue.

mbudiu-vmw:
Can you please fix the conflicts? This should also kick Travis again.

matijasyrmia:
@mbudiu-vmw can you check new commits? Also, I have pulled new commits for p4runtime, but there is still conflict.

mbudiu-vmw:
To resolve the remaining conflict I suspect you have to follow the instructions in the README and run `git clone --recursive https://github.com/p4lang/p4c.git` and then push again.

mbudiu-vmw:
Do this in master, and then rebase on master.

midend/predication.cpp
```diff
@@ -148,46 +136,57 @@ const IR::Node* Predication::clone(const IR::AssignmentStatement* statement) {
 const IR::Node* Predication::preorder(IR::AssignmentStatement* statement) {
     if (!inside_action || ifNestingLevel == 0)
         return statement;
-    ExpressionReplacer replacer(clone(statement->right), travesalPath, getContext());
+    auto context = getContext();
+    std::vector<const IR::Expression*> conditions;
+    while (context != nullptr) {
+        if (context->node->is<IR::IfStatement>()) {
```
mbudiu-vmw, line 142: you could also write this as `if (auto ifs = context->node->to<IR::IfStatement>) conditions.push_back(ifs->condition)`

midend/predication.cpp
```diff
@@ -148,46 +136,57 @@ const IR::Node* Predication::clone(const IR::AssignmentStatement* statement) {
...
+            conditions.push_back(context->node->to<IR::IfStatement>()->condition);
+        }
+        context = context->parent;
+    }
```
ChrisDodd, line 146: You could write this more concisely as

    Context *ctxt = nullptr;
    std::vector<const IR::Expression*> conditions;
    while (auto ifs = findContext<IR::IfStatement>(ctxt))
        conditions.push_back(ifs->condition);

matijasyrmia: It must be `const Context *`.

matijasyrmia:
> 
> 
> Thank you for doing this.
> The logic is a bit involved, so I cannot promise that I cannot guarantee that there are no missed corner cases, but this is certainly better than it used to be.
@mbudiu-vmw 
I'm here for any other corner cases. If there is anything more that you want from me just let me know.

PR #2277, opened 2020-03-30
@noWarnUnused annotation to supress unused warnings

ChrisDodd:
- mostly for declarations in arch headers for optional components


p4include/core.p4
```diff
@@ -66,6 +66,7 @@ extern packet_out {
 extern void verify(in bool check, in error toSignal);
 
 /// Built-in action that does nothing.
+@noWarnUnused
```
mbudiu-vmw: This has caused lots of reference outputs to change.

mbudiu-vmw:
@ChrisDodd : we discussed this in the LDWG and we'll change it to `@noWarn("unused")`.
I will file an issue with the compiler.

mbudiu-vmw:
We should also change the spec.

PR #2278, opened 2020-03-31
Do not resolve identifiers in the enclosing context

mbudiu-vmw:


frontends/common/resolveReferences/resolveReferences.cpp
```diff
@@ -20,15 +20,50 @@ limitations under the License.
...
+            }
+
+            auto ns = d->to<IR::INamespace>();
+            if (!ns || !ns->isRecursive()) {
+                // Check to see whether d is a parent of the current
+                // node in the IR DAG; in this case we cannot resolve
+                // to it.  An example would be: table t { key = { t: exact }}
+                // However, we allow "recursive" namespaces resolve to identifiers
+                // within their context; this is necessary for example to allow
+                // transitions from a state to itself.
```
ChrisDodd, line 55: This seems like `t` should resolve to the table (not to whatever t might be in the enclosing scope), which would then be an error (can't use a table as a key expression)...

ChrisDodd:
This seems like the wrong solution to this problem -- either symbol resolution should take the type of the symbol into account (so that when resolving eg, a name in a key expression, we would only consider objects that can be used in such an expression (instance variables that are headers or metadata, basically), or we should resolve to the nearest definition of the name (as we do currently), and flag an error if it is the wrong type.

testdata/p4_16_samples_outputs/issue2267-first.p4
```diff
@@ -0,0 +1,16 @@
...
+        }
+        actions = {
+            @defaultonly NoAction();
+        }
+        default_action = NoAction();
+    }
+    apply {
+    }
+}
+
```
ChrisDodd, line 16: Consider changing this test case to

    control MyC(bit<8> t, bit<8> u) {
        table t {
            key = { u: exact @name("u") ; }
        }
        table u {
            key = { t: exact @name("t") ; }
        }
        apply {
            t.apply();
            u.apply();
        }
    }

mbudiu-vmw:
We will give a simpler solution to this problem.

PR #2282, opened 2020-04-01
P4 runtime corner case handling

ChrisDodd:
I've run into some problems trying to update fonrtend refpoint with some of our internal tests

- avoid crashing for Type_Dontcare in API generation
  we have a number of tests that use a `_` (dontcare) type for the index type -- the backend will infer the index type based on the counter/meter/register size, but the API generation has trouble.  Should probably fix API generation to do the same inference.

- 16 bit id masking to make things work
  test cases fail without this due to disagreeing on what the ids of various things should be.

Antonin, if you could give some advice on better ways of dealing with these problems and/or pointer on where to look?  We won't merge this change but may need it as a temporary hack to make things work.

### DIFF
--- a/control-plane/p4RuntimeArchHandler.h
+++ b/control-plane/p4RuntimeArchHandler.h
@@ -403,8 +403,9 @@ struct Counterlike {
             auto typeArg = type->arguments->at(*indexTypeParamIdx);
             // We ignore the return type on purpose, but the call is required to update p4RtTypeInfo
             // if the index has a user-defined type.
-            TypeSpecConverter::convert(refMap, typeMap, typeArg, p4RtTypeInfo);
-            index_type_name = getTypeName(typeArg, typeMap);
+            if (!dynamic_cast<const IR::Type_Dontcare *>(typeArg)) {
+                TypeSpecConverter::convert(refMap, typeMap, typeArg, p4RtTypeInfo);
+                index_type_name = getTypeName(typeArg, typeMap); }
         }
 
         return Counterlike<Kind>{declaration->controlPlaneName(),
--- a/control-plane/p4RuntimeSerializer.cpp
+++ b/control-plane/p4RuntimeSerializer.cpp
@@ -489,7 +489,7 @@ class P4RuntimeSymbolTable : public P4RuntimeSymbolTableIface {
             // resolve hash collisions, the id that we select depends on the order in
             // which the names are hashed. This is why we sort the names above.
             boost::optional<p4rt_id_t> id = probeForId(nameId, [=](uint32_t nameId) {
-                return (resourceType << 24) | (nameId & 0xffffff);
+                return (resourceType << 24) | (nameId & 0xffff);
             });
 
             if (!id) {


control-plane/p4RuntimeArchHandler.h
```diff
@@ -403,8 +403,9 @@ struct Counterlike {
             auto typeArg = type->arguments->at(*indexTypeParamIdx);
             // We ignore the return type on purpose, but the call is required to update p4RtTypeInfo
             // if the index has a user-defined type.
-            TypeSpecConverter::convert(refMap, typeMap, typeArg, p4RtTypeInfo);
-            index_type_name = getTypeName(typeArg, typeMap);
+            if (!dynamic_cast<const IR::Type_Dontcare *>(typeArg)) {
```
mbudiu-vmw, line 406: We have a function: `typeArg->is<IR::Type_Dontcare>()`

ChrisDodd: Unfortunately, this seems to trigger a gcc bug -- if I use that here, it chokes.  Even though it works fine just a few lines above.

control-plane/p4RuntimeArchHandler.h
```diff
@@ -403,8 +403,9 @@ struct Counterlike {
...
+                TypeSpecConverter::convert(refMap, typeMap, typeArg, p4RtTypeInfo);
```
antoninbas, line 407: a better solution may be to handle `Type_Dontcare` in the `TypeSpecConverter` visitor?

also is there any PSA / v1model program that can illustrate this case? I'm having trouble figuring out what `Type_Dontcare` means here. If this code is exercised by your internal compiler, did you add the correct `indexTypeParamIdx` methods (see https://github.com/p4lang/p4c/commit/d7783e1036de5c1d2e6b65cdc2d1c3bc47db68ed#diff-36306e8998b70ba36e5831d493ce90f4)?

control-plane/p4RuntimeSerializer.cpp
```diff
@@ -489,7 +489,7 @@ class P4RuntimeSymbolTable : public P4RuntimeSymbolTableIface {
             // resolve hash collisions, the id that we select depends on the order in
             // which the names are hashed. This is why we sort the names above.
             boost::optional<p4rt_id_t> id = probeForId(nameId, [=](uint32_t nameId) {
-                return (resourceType << 24) | (nameId & 0xffffff);
+                return (resourceType << 24) | (nameId & 0xffff);
```
antoninbas, line 492: AFAIK, there is no reason to revert back to 16-bit of hash value. Are you sure it is not just about updating reference outputs for your internal tests? If there is something making use of the 2nd most significant byte in your internal compiler, you may want to try and figure out if this is strictly required or you can update that code.

control-plane/p4RuntimeArchHandler.h
ChrisDodd: I modified the psa-basic-counter-bmv2.p4 to show this.  Basically, requiring the user to come up with a bit size for the indexes is painful -- they generally just want a size that is 'big enough' and efficient on the target (which may vary from target to target).  Dontcare (`_`) seeme like a reasonable way of saying this, and had been working previously.

control-plane/p4RuntimeSerializer.cpp
ChrisDodd: Using the full 24 bits causes tests to fail with 
```
Traceback (most recent call last):
  File "/mnt/p4c/extensions/p4_tests/p4testutils/p4runtime_base_tests.py", line 852, in handle
    return f(*args, **kwargs)
  File "/mnt/p4c/extensions/p4_tests/p4_16/tor.ptf/test.py", line 71, in runTest
    self.runCommonTest("write_lpm_entries.pb.txt")
  File "/mnt/p4c/extensions/p4_tests/p4_16/tor.ptf/test.py", line 22, in runCommonTest
    response = self.write_request(p4runtime_request)
  File "/mnt/p4c/extensions/p4_tests/p4testutils/p4runtime_base_tests.py", line 514, in write_reques
t
    rep = self._write(req)
  File "/mnt/p4c/extensions/p4_tests/p4testutils/p4runtime_base_tests.py", line 511, in _write
    raise P4RuntimeWriteException(e)
P4RuntimeWriteException: <unprintable P4RuntimeWriteException object>
```
restricting it to 16 bits works fine

antoninbas: That's an easy one. IIRC this test hardcodes ids in the "write_lpm_entries.pb.txt" file. So you just need to edit that file manually to include the new ids (with the 24-bit hash value) and it will pass.

control-plane/p4RuntimeArchHandler.h
antoninbas: Interesting, I wasn't aware of this feature. Then it seems that you should be able to add a PSA test case as part of your PR so we don't break this in the future.

control-plane/p4RuntimeSerializer.cpp
ChrisDodd: That is extremely helpful -- thank you.  Hopefully that will make this issue go away.

ChrisDodd:
I'll put together an issue/proposal for the Dontcare index type -- I think it would be useful to support.

PR #2285, opened 2020-04-04
use /usr/bin/env bash instead of /bin/bash

flokli:
Like most of the shebangs in the rest of the repo, /usr/bin/env to pick
bash from `$PATH`, instead of invoking it from /bin/bash.

This fixes execution on systems that don't have bash in /bin/bash, or
have explicitly configured another one to be used.

### DIFF
--- a/backends/ebpf/targets/ebpfenv.py
+++ b/backends/ebpf/targets/ebpfenv.py
@@ -67,7 +67,7 @@ class Bridge(object):
 
     def ns_proc_open(self):
         """ Open a bash process in the namespace and return the handle """
-        cmd = self.get_ns_prefix() + " /bin/bash "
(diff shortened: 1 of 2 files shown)


flokli:
Travis doesn't seem to report back, but when showing the details, one can see the builds all succeeded.

jafingerhut:
I have on occasion hit a button on the Travis CI page for a p4c PR that causes the tests to be run again, but that button seems not to be available in this case, where all individual tests report success.  Weird.

flokli:
I amended the commit with a new date and pushed again - let's see if Travis reports back this time or is just broken…

flokli:
https://travis-ci.org/github/p4lang/p4c/builds/671723793 again is green when visiting the Travis Website, but not reporting back the status. Can someone with the appropriate permissions debug the Travis Integration?

PR #2286, opened 2020-04-04
tools/cpplint: bump to a later release supporting python2

flokli:
While the README says only python3 is required, running the tests (which
run cpplint) still requires python2 to be around.

Fix this, by bumping our version of `cpplint.py` to their latest release
(1.4.5, d0465e6b7b52335d21c45b5769e9eb0d648a7ea5), which works on both
python2 and python3.

### DIFF
--- a/tools/cpplint.py
+++ b/tools/cpplint.py
@@ -1,4 +1,4 @@
-#!/usr/bin/env python2
+#!/usr/bin/env python
 #
 # Copyright (c) 2009 Google Inc. All rights reserved.
 #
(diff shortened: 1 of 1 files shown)


flokli:
Urgh, I wasn't aware you maintain your own custom fork of cpplint.

The upstream version doesn't know about `filematch` (from `CPPLINT.cfg`), and seems to have some P4-specific tweaks (like 8acc550d1cefe0222abbfa219484e736897c70f5, b67652697405922ff3bb41490c44d95753ddc9c0, 8cf091313a9cf4523185a127986f1d36a090ff02, 6c596703ecd1171db06aa7b37e854a169a4498f1, 8b6f0e65061ad22b4dd7f991406b0496de5cae94, 496c150d754368384ea00955184a2818be0bc9dc, e3a13f976a2dd46f6eb00bd2b2bde2fb5feef662.

I didn't check if there's more changes before the initial code drop (bc7b75bd1892ab230c2ec2d125af8529cf092f4a)

Do you have any plans on moving to some official cpplint, or move this script to python3?

cc10512:
I've done a quick attempt to move cpplint to python3 when I moved the other scripts, and I ran into the same issues you're seeing. I did not have cycles to spend on it, so I left it as is. If you can spare some time to move it to the official cpplint (or at least factor out the custom changes so that we can upgrade to the official version, it would be much appreciated!

flokli:
@cc10512 do you know whom to ask from which repo and rev the `cpplint.py` in the repo initially came from?

This would make it easier to do a diff on the initial codedrop commit to figure out all p4c-specific quirks that existed in there, and iterate from there.

Another idea: What about moving to another linter, like `clang-tidy`?

cc10512:
@flokli The original developers for p4c who are still active are @mbudiu-vmw, @ChrisDodd, and @antoninbas. They may know the different instantiations of how cpplint was forked and modified.

As for another linter ... I seem to recall someone trying `clang-format` at some point, but don't remember who and why the transition did not happen. I agree that having a set of customized rules with a standard tool is less maintenance.

flokli:
I'll close this PR for now as it obviously doesn't work with just a new upstream cpplint. I opened #2293 to track this.

PR #2295, opened 2020-04-06
Return the proper vectors; fixes #2289

### LINKED ISSUE #2289: Compiler Bug: visitor returned invalid type Vector<Declaration> for IndexedVector<StatOrDecl>
The programs are becoming a bit more esoteric now.
In this case using a nested struct inside a function that is called by another function leads to a compiler bug. I tried to reduce the case as much as possible.

[nested_structs_and_functions.p4.txt](https://github.com/p4lang/p4c/files/4437648/nested_structs_and_functions.p4.txt)

mbudiu-vmw:


mbudiu-vmw:
You can file an issue about a solution based on modifying the visitor. I will merge this for now.

PR #2298, opened 2020-04-07
When errors happen do not bug-check; fixes #2290

### LINKED ISSUE #2290: Compiler Bug: Unexpected type <Type_Name> for nested headers
Declaring a header that has another header as member leads to this compiler bug. While this is not allowed in the frontend it still seems to cause a compiler bug.

[nested_header.p4.txt](https://github.com/p4lang/p4c/files/4437667/nested_header.p4.txt)

mbudiu-vmw:


mbudiu-vmw:
@ChrisDodd : this one and a couple of other bug-fixes PRs are in fact very small, in case you have some time to review.

PR #2299, opened 2020-04-08
Infer counter/meter/register index sizes from P4_14 instance count

ChrisDodd:
The first of these changes is fairly straight-forward and non-controversial -- it infers the size of action arguments use as indexes for counters/meters/registers in P4_14 code based on the size of the object in question.  This is the common way P4_14 compilers (like the old p4-hlir based stuff) worked.

The second change is wider-ranging in effect.  It changes the v1model to have an index size type parameter for counter/meter/register externs (much as is done in PSA), and uses the inferred size as above when translating from P4_14.  This is a "first step" to changing the P4_14 translation code to generating PSA model code instead of v1model (and getting rid of v1model altogether).

### DIFF
--- a/ir/v1.cpp
+++ b/ir/v1.cpp
@@ -130,6 +130,19 @@ unsigned IR::Primitive::inferOperandTypes() const {
     return 0;
 }
 
+// infer the index width of a meter/counter/register based on the instance count
+// default to 32 bits if we can't figure it out
+int IR::Stateful::index_width() const {
+    return instance_count > 1 ? ceil_log2(instance_count) : 32;
+}
+
+static int inferIndexWidth(const IR::Expression *obj) {
+    if (auto *glob = obj->to<IR::GlobalRef>())
+        if (auto *sful = glob->obj->to<IR::Stateful>())
+            return sful->index_width();
+    return 32;
+}
+
 const IR::Type *IR::Primitive::inferOperandType(int operand) const {
     const IR::Type *rv = IR::Type::Unknown::get();
     unsigned infer = 0;
@@ -147,7 +160,7 @@ const IR::Type *IR::Primitive::inferOperandType(int operand) const {
     if (name == "truncate")
         return IR::Type::Bits::get(32);
     if ((name == "count" || name == "execute_meter") && operand == 1)
-        return IR::Type::Bits::get(32);
+        return IR::Type::Bits::get(inferIndexWidth(operands.at(0)));
     if (name.startsWith("execute_stateful") && operand == 1)
         return IR::Type::Bits::get(32);
     if ((name == "clone_ingress_pkt_to_egress" || name == "clone_i2e" ||
@@ -155,9 +168,13 @@ const IR::Type *IR::Primitive::inferOperandType(int operand) const {
         operand == 0) {
         return IR::Type::Bits::get(32); }
     if ((name == "execute") && operand == 2)
-        return IR::Type::Bits::get(32);
+        return IR::Type::Bits::get(inferIndexWidth(operands.at(0)));
     if (name == "modify_field_conditionally" && operand == 1)
         return IR::Type::Bits::get(1);
+    if (name == "register_read" && operand == 2)
+        return IR::Type::Bits::get(inferIndexWidth(operands.at(1)));
+    if (name == "register_write" && operand == 1)
+        return IR::Type::Bits::get(inferIndexWidth(operands.at(0)));
     if (name == "shift_left" && operand == 1) {
         if (operands.at(0)->type->width_bits() > operands.at(1)->type->width_bits())
             return operands.at(0)->type;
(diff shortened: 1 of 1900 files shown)


ir/v1.cpp
```diff
@@ -130,6 +130,19 @@ unsigned IR::Primitive::inferOperandTypes() const {
     return 0;
 }
 
+// infer the index width of a meter/counter/register based on the instance count
+// default to 32 bits if we can't figure it out
+int IR::Stateful::index_width() const {
+    return instance_count > 1 ? ceil_log2(instance_count) : 32;
```
jafingerhut, line 136: A minor edge case, but if someone creates an object with size=1, should the index_width also be 1?

jafingerhut:
Are there people who actively use the v1model architecture that might want a heads-up about a backwards-incompatible change in its definition?  I have no opposition to these changes.  It just seems polite to let them know that something like this is coming their way when they upgrade to p4c code after this is merged in.

The only examples that spring to mind are the people interested in P4 at ONF and Google, e.g. @konne88 @stefanheule @bocon13 or anyone they would like to ping.

hesingh:
I do use the v1model on generic compute to prototype new P4-16, not P4-14 code.  Anyway, thanks for the heads up @jafingerhut

ir/v1.cpp
ChrisDodd: Possibly -- there's no good answer.  There really should not be any index in that case, but we can't use `bit<0>` and don't have a good way of having it specialized

ChrisDodd:
The impact on anyone who might be using v1model is why this is "controversial".  On the one hand, we don't want to cause problems for people and make them unhappy with P4, but on the other, we'd really prefer to have people using PSA.  We could just back-burner this until someone is ready to take on the task of moving the P4_14 translation completely over to PSA (at which point we simply "obsolete" v1model).

Another possibility that would allow incrementally moving v1model->PSA would be to version v1model.  That would add a bunch of "temporary" complexity while the transition was happening, but would keep backwards compatibility for people who have chosen to use v1model.

stefanheule:
Thanks Andy! We only use P4-16, so looks like we won't be affected.

hanw:
We need to work on versioning architecture API before introducing this change. The change here would break all existing p4-16 v1model programs.

p4include/v1model.p4
```diff
@@ -270,7 +276,7 @@ extern register<T> {
      *
      * allocates storage for 512 values, each with type bit<32>.
      */
-    register(bit<32> size);
+    register(bit<32> size);  // FIXME -- arg should be `int` but that breaks typechecking
```
mbudiu-vmw, line 279: please file an issue about this; probably a very small reproduction will be enough

p4include/v1model.p4
```diff
@@ -117,7 +119,7 @@ enum MeterType {
     bytes
 }
 
-extern counter {
+extern counter<I> {
```
mbudiu-vmw, line 122: I don't think we can break compatibility with existing programs; there are a lot of tutorials out there that may be using this file. I can see two ways forward:
- version the file; add a `const string v1model_version = "1.0.0";` and have the back-end do things depending on this version. We keep multiple v1model files around, and the user symlinks or uses -I to indicate the desired one
- deprecate the existing counter, the way we did it with the checksum apis, and add a new counter class which is generic. Old things will keep working, but we may duplicate some significant amount of code in the backend. The front-end should not really care.
We should version v1model anyway; the question is whether we add a new file or keep modifying things in the old one.
Maybe we can start with a PR with just the versioning. This would have been a great subject to discuss during the LDWG.

antoninbas:
If we are going to make changes to v1model, I suggest changing digest to use an extern type instead of an extern method. @jafingerhut

antoninbas:
@stefanheule these changes affect v1model.p4 and therefore all (or rather most) P4_16 programs written for the v1model architecture

ChrisDodd:
I've added a versioning scheme to maintain backwards compatibility with old P4_16 programs that use v1model.  There's now a V1MODEL_VERSION macro that must be set to specify what (newer) version you want before including v1model.p4 in a p4_16 program.  The VERSION being an 8-digit date stamp, as "v1model" is not a standard architecture, but specific to this implementation's P4_14 support.  The P4_14->16 conversion sets the version in the dumped code it produces.

testdata/p4_14_samples_outputs/01-BigMatch-first.p4
```diff
@@ -1,4 +1,5 @@
 #include <core.p4>
+#define V1MODEL_VERSION 20200408
```
mbudiu-vmw, line 2: We could have the test script set this flag.

ChrisDodd: It's the toP4 dumping code that sets/outputs this -- whenever you dump a P4 program that included v1model.p4, it also dumps a #define specifying which version of v1model it was that was included.

PR #2324, opened 2020-04-18
Fix P4_16 name resolution

ChrisDodd:
- when multiple namespaces exist in a single IR::Node, and symbols in
  those namespaces may shadow each other, lookup depends on the location
  of the name.

This is a subtle corner case that is a consequence of the decision to prohibit uses before definitions in P4_16 -- it also means that symbols in inner scopes only partly shadow symbols in enclosing scopes (only in the part of the scope after they are defined.)  Which means that in objects embodying multiple scopes, we must keep those scopes separate for symbol lookup.

### DIFF
--- a/ir/ir.def
+++ b/ir/ir.def
@@ -81,25 +81,20 @@ class ParserState : ISimpleNamespace, Declaration, IAnnotated {
 }
 
 // A parser that contains all states (unlike the P4 v1.0 parser, which is really just a state)
-class P4Parser : Type_Declaration, ISimpleNamespace, IApply, IContainer {
+class P4Parser : Type_Declaration, INestedNamespace, ISimpleNamespace, IApply, IContainer {
     Type_Parser                                 type;
     optional ParameterList                      constructorParams = new ParameterList;
     optional inline IndexedVector<Declaration>  parserLocals;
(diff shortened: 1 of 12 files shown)


ir/ir.def
```diff
@@ -81,25 +81,20 @@ class ParserState : ISimpleNamespace, Declaration, IAnnotated {
 }
 
 // A parser that contains all states (unlike the P4 v1.0 parser, which is really just a state)
-class P4Parser : Type_Declaration, ISimpleNamespace, IApply, IContainer {
+class P4Parser : Type_Declaration, INestedNamespace, ISimpleNamespace, IApply, IContainer {
```
mbudiu-vmw, line 84: Are these the only two nested namespaces?

ChrisDodd: These are the only two objects that have both arguments and locals that might shadow the arguments, so need this special handling to get the P4_16 name lookup correct.  Function and Action objects don't have any locals directly (any locals will be nested in the body BlockStatement)

ChrisDodd: Actually, I take that back -- extern type might also, so I should extend this to them (though it can't happen in a user program, as users can't define extern types -- only architectures)

Issue #13, opened 2016-04-19
The P4-16 grammar has the wrong priority for method calls with templates


mbudiu-bfn:
The testdata/v1_2_samples/cast-call.p4 program is a repro of this bug.

ChrisDodd:
I actually have a fix for this old issue...

mbudiu-vmw:
Fixed by #2326

PR #2326, opened 2020-04-20
Change precedence of '<' based on symbol immediately before it

ChrisDodd:
- if the symbol before it is something that expects template args, we
  use a higher precedence (same as postfix operators) rather than
  normal less-than precedence.

Fixes issue #13

### DIFF
--- a/backends/p4test/CMakeLists.txt
+++ b/backends/p4test/CMakeLists.txt
@@ -77,8 +77,6 @@ set (P4RUNTIME_EXCLUDE
 )
 
 set (P4_XFAIL_TESTS
-  # issue #13
-  testdata/p4_16_samples/cast-call.p4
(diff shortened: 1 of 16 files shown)


mbudiu-vmw:
@ChrisDodd : do you think we should change the spec to reflect this fix?

ChrisDodd:
I don't think there's any spec change -- this fixes the implementation to match the spec.

Unfortunately, it looks like this is inadequate for more complex cases, mostly because the `Util::ProgramStructure` object is incomplete -- it doesn't contain any info on the types of objects (nor on parameters, though that is fixable), which is needed for knowing the type of things like `a.b`.  The fix is probably to toss the ProgramStructure and use the actual IR::INamespace objects we're constructing in the parser to do symbol lookup.

PR #2337, opened 2020-04-24
Stronger checks for int values

### LINKED ISSUE #2335: Compiler Bug message with program defining variable with type int
With the version of p4c source code indicated below, the attached program causes a Compiler Bug to occur.  I believe the program should be illegal, since it attempts to define a variable with type `int`, which I think is considered illegal in the language spec (values of type `int` should all be compile-time known values).

```
$ p4c --version
p4c 1.2.0 (SHA: b1374a14)

$ p4c type-int-variable.p4 
In file: /home/andy/p4c/frontends/p4/typeChecking/typeChecker.cpp:2009
Compiler Bug: /home/andy/p4c/frontends/p4/typeChecking/typeChecker.cpp:2009: Null cst
```

Discovered by @hackedy and reported in this comment on another issue, but seemed worth creating its own issue to track separately from that one: https://github.com/p4lang/p4c/issues/2334#issuecomment-618437733

[type-int-variable.p4.txt](https://github.com/p4lang/p4c/files/4524081/type-int-variable.p4.txt)

mbudiu-vmw:
Fixes #2335

### DIFF
--- a/frontends/common/resolveReferences/referenceMap.h
+++ b/frontends/common/resolveReferences/referenceMap.h
@@ -44,7 +44,7 @@ class MinimalNameGenerator : public NameGenerator, public Inspector {
     }
 
     /// Generate a name from @p base that fresh for the program.
-    cstring newName(cstring base);
+    cstring newName(cstring base) override;
 };
 
 // FIXME -- temp cpmmon base class to allow use of ReferenceMap or ResolutionContext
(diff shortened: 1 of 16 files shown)


mbudiu-vmw:
@ChrisDodd : this Travis check failed due some more stringent checks from clang on MacOS.
You may want to fix some of these warnings.

mbudiu-vmw:
@ChrisDodd : maybe you can also review this one; it's mostly warning fixes with clang, and one bug fix.

PR #2338, opened 2020-04-24
Pay more attention to types during constant folding

### LINKED ISSUE #2332: Untyped constant folding not sound
The compiler front-end does several transformations on the input program prior to type checking. One of these is constant folding, which appears to be transforming the following program, which is not well-typed,
```
const bit<32> x = 1w1;
```
into this program, which is:
```
const bit<32> x = 32w1;
```

(H/T @hackedy for finding this bug.)

mbudiu-vmw:
Fixes #2332


jnfoster:
Seems like e39a079 is not a part of this fix?

mbudiu-vmw:
I was hoping to merge that one first...
But I have to duplicate the work.

Issue #2340, opened 2020-04-27
bmv2 back end gives error for const entries lpm field with mask 0


jafingerhut:
The indicated version of p4c gives a compile-time error in the bmv2 back end code for the attached program, which as far as I can tell should be perfectly legal.  The bug appears to be specific to the bmv2 back end code, when generating the Bmv2 JSON for a table with const entries and an lpm match_kind field with mask 0.
```
$ p4c --version
p4c 1.2.0 (SHA: b1374a14)

$ p4c --target bmv2 --arch v1model v1model-const-entries-bmv2.p4
v1model-const-entries-bmv2.p4(71): [--Werror=invalid] error: &&&: invalid mask for LPM key
            0x000000000000 &&& 0x000000000000: act_hit(3);
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```
[v1model-const-entries-bmv2.p4.txt](https://github.com/p4lang/p4c/files/4537007/v1model-const-entries-bmv2.p4.txt)

jafingerhut:
Closing this issue, since it should be fixed with latest master as of this commit: https://github.com/p4lang/p4c/pull/2341

PR #2341, opened 2020-04-27
Fix bug in BMv2 back end handling const entries with lpm field with m…

jafingerhut:
…ask 0


jafingerhut:
Fix for issue #2340

mbudiu-vmw:
you have to fix the cpplint warnings

PR #2346, opened 2020-04-28
Constant fold slices always

### LINKED ISSUE #2342: Compile-time-known and slices
I have the following expression
```p4

    apply {
        const bool tmp = 1 != 8w2[7:0];
    }
}
```

which gives me this error:

```bash
compile_time_known.p4.txt(31): [--Werror=type-error] error: !=: Cannot evaluate initializer to a compile-time constant
        const bool tmp = 1 != 8w2[7:0];

```

Is this correct? According to the spec, slices are acceptable when their operand is a compile-time-known value.
[compile_time_known.p4.txt](https://github.com/p4lang/p4c/files/4538467/compile_time_known.p4.txt)

mbudiu-vmw:
Fixes #2342

### DIFF
--- a/frontends/common/constantFolding.cpp
+++ b/frontends/common/constantFolding.cpp
@@ -521,8 +521,6 @@ const IR::Node* DoConstantFolding::postorder(IR::Slice* e) {
         return e;
     }
 
-    if (!typesKnown)
-        return e;
     auto e0 = getConstant(e->e0);
     if (e0 == nullptr)
         return e;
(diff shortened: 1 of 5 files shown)


frontends/common/constantFolding.cpp
```diff
@@ -559,9 +557,7 @@ const IR::Node* DoConstantFolding::postorder(IR::Slice* e) {
     big_int mask = 1;
     mask = (mask << (m - l + 1)) - 1;
     value = value & mask;
-    auto resultType = typeMap->getType(getOriginal(), true);
-    if (!resultType->is<IR::Type_Bits>())
-        BUG("Type of slice is not Type_Bits, but %1%", resultType);
+    auto resultType = IR::Type_Bits::get(m - l + 1);
```
ChrisDodd, line 560: Should this be signed or unsigned?  This uses unsigned unconditionaly, but if the constant is a signed should it be signed?  I forget what we decided to say in the spec, but the old code would preserve the signedness of the source

mbudiu-vmw: This is from the 1.2.0 spec changelog:
```
Define the type of bit slices from signed and unsigned values to be unsigned. 
```

Issue #2357, opened 2020-05-04
Int as parameter and assignments


fruffy:
This comes back to the question in #2354. Are these sorts of parameters allowed?

```p4
bit<8> do_function1(in int val1) {
    return 8w1;
}

bit<8> do_function2(inout bit<8> val0, in int val1) {
    val0 = val1;
    return 8w1;
}

control ingress(inout Headers h, inout Meta m, inout standard_metadata_t sm) {

    apply {
        bit<8> tmp = 8w1;
        do_function1(1);
        do_function2(tmp, 1);
    }
}
```

In this case, I get an error which says:

```
...
InlineActions_3_RemoveAllUnusedDeclarations
FrontEnd_44_InlineActions
P4::TypeChecking_0_ResolveReferences
P4::TypeChecking_1_TypeInference
InlineFunctions_0_TypeChecking
InlineFunctions_1_DiscoverFunctionsInlining
InlineFunctions_2_InlineDriver
PassRepeated_0_ResolveReferences
PassRepeated_1_RemoveUnusedDeclarations
PassRepeated_2_ResolveReferences
PassRepeated_3_RemoveUnusedDeclarations
RemoveAllUnusedDeclarations_0_PassRepeated
InlineFunctions_3_RemoveAllUnusedDeclarations
P4::TypeChecking_2_ResolveReferences
[--Werror=type-error] error: val1_1: Cannot declare variables with type int
```
Which also does not seem correct, maybe caused by the latest changes. Should declaration be a `Declaration_Constant`?

[int_function_arguments.p4.txt](https://github.com/p4lang/p4c/files/4576204/int_function_arguments.p4.txt)

mbudiu-vmw:
This should be fixed by #2360, but it's not easy to check since this is not a full program.

fruffy:
Forget the attachment.

PR #2360, opened 2020-05-04
int parameters must be directionless

### LINKED ISSUE #2354: Compiler Bug: Null cst for InfInt Parameters
New week, new bugs!

I was fuzzing integer types for a bit. The attached program crashes the type checker.
I also do not know if this is a legal construct, is it allowed to define a read-only parameter with type `InfInt`?

```p4
    action do_action1(in int val) {
        bool test = 8w1 == val;
    }
    action do_action2(in int val) {
        h.eth_hdr.eth_type = h.eth_hdr.eth_type + val;
    }

    apply {
        do_action1(1);
        do_action2(1);
    }
}
```

[type_checking_crash.p4.txt](https://github.com/p4lang/p4c/files/4574069/type_checking_crash.p4.txt)

mbudiu-vmw:
Fixes #2354

### DIFF
--- a/frontends/p4/typeChecking/bindVariables.cpp
+++ b/frontends/p4/typeChecking/bindVariables.cpp
@@ -12,7 +12,7 @@ class ErrorOnInfInt : public Inspector {
     void postorder(const IR::Expression *expression) override {
         auto t = typeMap->getType(expression, true);
         if (t->is<IR::Type_InfInt>())
-            ::error("%1%: could not infer a width", expression);
+            ::error(ErrorType::ERR_TYPE_ERROR, "%1%: could not infer a width", expression);
     }
 };
 
--- a/frontends/p4/typeChecking/typeChecker.cpp
+++ b/frontends/p4/typeChecking/typeChecker.cpp
@@ -165,7 +165,7 @@ const IR::Type* TypeInference::getType(const IR::Node* element) const {
     // for some node; we are now just trying to typecheck a parent node.
     // So an error should have already been signalled.
     if ((result == nullptr) && (::errorCount() == 0))
-        BUG("Could not find type of %1%", element);
+        BUG("Could not find type of %1% %2%", dbp(element), element);
     return result;
 }
 
@@ -1474,14 +1474,16 @@ const IR::Node* TypeInference::postorder(IR::Parameter* param) {
     if (!readOnly && paramType->is<IR::Type_InfInt>()) {
         // We only give these errors if we are no in 'readOnly' mode:
         // this prevents giving a confusing error message to the user.
-        if (param->direction != IR::Direction::None)
+        if (param->direction != IR::Direction::None) {
             typeError("%1%: parameters with type %2% must be directionless",
                       param, paramType);
+            return param;
+        }
         if (findContext<IR::P4Action>()) {
             typeError("%1%: actions cannot have parameters with type %2%",
                       param, paramType);
+            return param;
         }
-        return param;
     }
 
     // The parameter type cannot have free type variables
--- a/frontends/p4/typeChecking/typeChecker.cpp
+++ b/frontends/p4/typeChecking/typeChecker.cpp
@@ -235,7 +235,7 @@ const IR::ParameterList* TypeInference::canonicalizeParameters(const IR::Paramet
     bool changes = false;
     auto vec = new IR::IndexedVector<IR::Parameter>();
     for (auto p : *params->getEnumerator()) {
-        auto paramType = getType(p);
+        auto paramType = getTypeType(p->type);
         if (paramType == nullptr)
             return nullptr;
         BUG_CHECK(!paramType->is<IR::Type_Type>(), "%1%: Unexpected parameter type", paramType);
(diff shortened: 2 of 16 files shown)


mbudiu-vmw:
Fixes #2357

jafingerhut:
Is it true that all values of type `int` in a P4_16 program must be compile-time known values?

Is a directionless parameter of an action a compile-time known value?

If so, I do not understand how it can be, since the control plane software could install table entries using such an action with values for such parameters that are not known at compile time.

mbudiu-vmw:
I guess this is necessary, but not sufficient. Actions cannot have any int parameters at all. I should add a check for that.

jafingerhut:
Not only actions, but any run-time parameters in general, whether they be for actions, functions (extern or written-in-P4 kinds), methods, controls, or parsers.

Constructor parameters for extern objects, parsers, and controls seem reasonable to have type `int`, but I cannot think of any other kinds of parameters where type `int` makes sense to me.

mbudiu-vmw:
No, functions can have int parameters. For example, the compiler typechecks the declaration of pop_front as if it is an extern method of a stack that has the signature `pop_front(int index);`. This requires the argument to be a constant.

jafingerhut:
I can see that it makes sense for a control, parser, or function to have a run-time parameter that is directionless or `in` with type `int`, as long as all calls in the program source code have compile-time known integer values given in the calls, for all such parameters.

Can actions have directionless parameters with type `int`?  I do not see how that would make sense.

jnfoster:
I think it could be legal if the action were used as a default_action?

action a(int x) { ...}
> table t = { ... default_action = a(42); }
>

Unclear if this is useful of course :-)

-N

mbudiu-vmw:
Actually in https://github.com/p4lang/p4c/pull/2360 I do reject all `int` parameters for actions.
So that should probably make it to the spec too.

mbudiu-vmw:
Are these changes in line with the spec?
If so, perhaps someone can approve this PR?

jafingerhut:
@mbudiu-vmw I am currently unclear on what the spec actually says should be allowed, versus not allowed, for values with type `int`.  I mean, I can look at some example programs and see 'yeah, that is clearly a run-time value with type `int`, so not allowed', and I can look at another program like the issue2354.p4 added as part of this P4, excerpted below:
```
    action do_action1(in int val) {
        bool test = 8w1 == val;
    }
    action do_action2(in int val) {
        h.eth_hdr.eth_type = h.eth_hdr.eth_type + val;
    }
    apply {
        do_action1(1);
        do_action2(1);
    }
}
```
and it seems to me that it is open to interpretation of the English text in the current spec whether it allows that, or not.

mbudiu-vmw:
this would be illegal using the latest PRs against the spec.
any `int` parameter must be directionless.

jafingerhut:
Perhaps if there was a brief description of these p4c changes, along with proposed changes to the language spec that we think are consistent with those, then I could try comparing those two things to each other, and the p4c changes to the code changes (or at least the example programs?)

jafingerhut:
@mbudiu-vmw "this would be illegal using the latest PRs against the spec."  Can you provide a link to the PR you mean?

mbudiu-vmw:
I actually thought I had filed a PR against the spec, but I don't see one.
This PR against the compiler requires `int` arguments to be directionless and forbids them altogether for actions.

jafingerhut:
Mihai, if you want me to take a shot at writing a spec PR for the changes below, let me know.  If you already have one in the works, I can wait and review it.

Directionless parameters are used for two completely different purposes in P4_16 (and maybe others, but these are the two I could find in a quick scan for the term 'direction' in the spec):

+ for everything other than an action (e.g. parser, control, function, ...), it represents the restriction that the argument value supplied in all calls must be a compile-time known value.
+ for an action, it represents the fact that the parameter value is (at least usually) supplied by the control plane software when adding entries to a table that use that action.  I say "at least usually" because a P4_16 program can also supply an expression as the argument for such directionless action parameters in explicit calls to the action, or in the table property `actions` (at least I think the last part might be true).

Given those two very different meanings for directionless parameters, it does seem reasonable to me to restrict type `int` parameters to be directionless for everything except actions, and to completely disallow type `int` parameters for actions.

If we wanted to be slightly more permissive, I could imagine allowing type `int` parameters for action parameters that were directionless, but only if all calls to the action supplied a compile-time known value for those parameters.  However, the benefit of allowing such calls seems to me to be lower than the complexity of specifying that corner case, and implementing it in the compiler.

Aside: It does strike me as a bit odd that directionless parameters are used for these two completely different purposes, but changing that seems impossible without breaking backwards compatibility and confusing everyone familiar with the current language.

jnfoster:
I agree. Unfortunately we didn't fix this issue (better syntax for
control-plane-supplied action data) before P4_16 got finalized.

mbudiu-vmw:
I wasn't planning to write anything, it is rather implicit in the spec. But we can always be more explicit.

jafingerhut:
I created this proposed PR: https://github.com/p4lang/p4-spec/pull/861

mbudiu-vmw:
I believe that whatever we do in the spec will not invalidate these assumptions, so I will merge this.
If this turns out to be false we'll file another issue.

PR #2365, opened 2020-05-05
Support new @p4runtime_translation syntax.

### LINKED ISSUE #2283: Support P4RT string translation
Add support for strings in the `@p4runtime_translation` annotation as described in:
https://github.com/p4lang/p4runtime/pull/274
https://github.com/p4lang/p4runtime/issues/270

smolkaj:
Fixes #2283.
Related: p4lang/p4runtime#274 and p4lang/p4runtime#270.
@antoninbas @stefanheule @jafingerhut @konne88


mbudiu-vmw:
No tests?

stefanheule:
I'm not very familiar with the compiler, but from my side this looks good. Thanks so much for working on this.

frontends/parsers/p4/p4parser.ypp
```diff
@@ -157,6 +157,9 @@ inline std::ostream& operator<<(std::ostream& out, const P4::Token& t) {
                    START_INTEGER_TRIPLE
                    START_STRING_LITERAL_TRIPLE
 
+                   // P4Runtime annotations
+                   START_P4RT_TRANSLATION_ANNOTATION  // @p4runtime_translation
```
antoninbas, line 161: @mbudiu-vmw I think the feedback we are looking for here is whether you are ok with "hardcoding" support for this annotation in the P4 parser? Or would you prefer a different approach? We could treat is as an unstructured annotation and "parse" it dynamically using C++ regular expressions. Would like to have and opinion from @liujed @liujed-intel as well if he can think of a better approach.

mbudiu-vmw: Sure, this is fine. If this is a "standard" annotation this is a reasonable way to do it.

liujed-intel: Agreed, this is fine if we expect all P4 compilers to support this annotation.

smolkaj:
I am observing some strange behavior when trying to write gtests for this, see #2379. I would be really helpful for some help understanding this; #2379 is a tiny example demonstrating the issue. @antoninbas any chance you know what's going on here?

antoninbas:
@smolkaj I replied to your issue

jafingerhut:
Wouldn't it achieve even better functional test coverage to have a test P4 program, check the generated P4Info file by hand, and then commit that as an expected output file that will fail if the generated P4Info file is ever different?

antoninbas:
> Wouldn't it achieve even better functional test coverage to have a test P4 program, check the generated P4Info file by hand, and then commit that as an expected output file that will fail if the generated P4Info file is ever different?

My preference is usually to have both types of tests; I agree that having a test P4 program and check-in the outputs is a requirement here.

smolkaj:
I still have two e2e tests failing for reasons I do not understand. In particular, psa.p4 contains a`@p4runtime_translation` annotation which is rejected by my new code:
```p4
@p4runtime_translation("p4.org/psa/v1/PortId_t", 32)
type PortIdUint_t         PortId_t;
```
This causes `testdata/p4_16_samples/psa-example-register2-bmv2.p4` and `testdata/p4_16_samples/p4rt_digest_complex.p4 ` to fail.

 @antoninbas, any chance you could take a look and see if you know what's going on there? It almost seems like the failing tests are parsed using a different frontend? In particular, note that the test `testdata/p4_16_samples/issue2283_1-bmv2.p4` uses the same syntax and works fine.
```p4
@p4runtime_translation("p4.org/psa/v1/PortId1_t", 32)
type PortIdUint_t PortId1_t;
```

mbudiu-vmw:
Have you tried running them under a debugger?

control-plane/typeSpecConverter.cpp
```diff
@@ -41,33 +41,49 @@ namespace P4 {
...
+    const IR::Expression* second_arg = ann->expr[1];
+
+    auto uri = first_arg->to<IR::StringLiteral>();
+    BUG_CHECK(uri != nullptr,
+              "%1%: the first argument to @p4runtime_translation must be a "
+              "string literal",
+              type);
+    payload->original_type_uri = std::string(uri->value);
+
+    BUG_CHECK(second_arg->to<IR::DefaultExpression>() != nullptr,
```
antoninbas, line 64: @smolkaj @mbudiu-vmw this doesn't seem right. Isn't a `DefaultExpression` basically a "don't care" expression? I think you meant to use `Expression` here but I am not very familiar with the IR in general.

frontends/parsers/p4/p4parser.ypp
```diff
@@ -395,6 +399,23 @@ fragment
           result->push_back(new IR::StringLiteral(@4, $4));
           result->push_back(new IR::StringLiteral(@6, $6));
           $$ = result; }
+
+      // P4Runtime annotations
+    | START_P4RT_TRANSLATION_ANNOTATION STRING_LITERAL "," p4rtControllerType
+        { auto* result = new IR::Vector<IR::Expression>();
+          result->push_back(new IR::StringLiteral(@2, $2));
+          result->push_back(new IR::DefaultExpression($4));
```
antoninbas, line 407: @smolkaj as I mentioned in the other comment, I doubt you meant to use a `DefaultExpression` here (it corresponds to don't care or `_` in P4). I'm pretty sure this is connected to your test failures as well, because I just saw this error message in the Travis logs:
```
psa.p4(78): [--Werror=invalid] error: PortId_t: the second argument to @p4runtime_translation must be one of the basic types 'string' or 'bit<W>', but got _
```

control-plane/typeSpecConverter.cpp
smolkaj: You're right, a DefaultExpression is a "don't care". I really want an IR::Type here, however the parser must return a vector of `IR::Expression`s for things to type check. In other words, using a `DefaultExpression` (that carries the type I'm interested in) is a hack;  I cannot use `Expression` directly since that is an abstract class.

frontends/parsers/p4/p4parser.ypp
smolkaj: See my comments above, this is a hack.

Could it be that there is some type checking pass that overwrites my p4rtControllerType with a type `_`, because that's the type assigned to `DefaultExpression`? Maybe @mbudiu-vmw would know?

mbudiu-vmw: Why not just encode the type into a string (a string literal) and decode it back when you need to look at it?
The real solution is to allow annotations also contain types, but for that we need to change the spec.

liujed-intel: I don't think we need a spec change here. To me, the issue is that the compiler's internal representation of parsed annotations simply isn't rich enough to capture what is needed for this annotation. One can imagine addressing this by specializing `Annotation` for the various kinds of annotations a compiler might need to handle.

ChrisDodd: We should probably restructure `Annotation` that way anyways -- currently it contains an unparsed token vector, a parsed expression vector and a parsed kv-pair vector, even though the one of the latter two will always be empty.  It should instead be an abstract base class with subclasses `UnparsedAnnotation`, `ExpressionListAnnotation`, and `KVListAnnotation` -- the annotation parser would replace the unparsed annotation with the appropriate subclass.  Then adding a `TypeAnnotation` subclass would be easy.

mbudiu-vmw: But this annotation has both an expression and a type.

liujed-intel: I think it would be a `P4rtTranslationAnnotation` class containing a string literal and a type.

smolkaj: Having a `P4rtTranslationAnnotation` class seems nice.

In the meantime, I used a different encoding:
* `IR::StringLiteral()` encodes the type "string"
* `IR::Constant(W)` encodes the type "bit\<W\>"

It's still a hack, of course, but this seems to work.

mbudiu-vmw: at the very least you should document the hack, and probably add a FIXME or TODO.

antoninbas: +1 for documenting this with a comment

at least the new hack is not as hacky as the old hack IMO :)

smolkaj: Sorry, I just forgot to push my documentation.
I also added a TODO

smolkaj:
All tests pass now, so it looks like this PR is finally ready.

smolkaj:
Thanks everyone for helping me through the technical challenges, I really appreciate it!

PR #2367, opened 2020-05-07
Check nullptr before invoking methods.

scweng:


test/gtest/arch_test.cpp
```diff
@@ -89,13 +90,15 @@ TEST_F(P4CArchitecture, duplicatedDeclarationBug) {
     )");
     auto pgm = P4::parseP4String(program, CompilerOptions::FrontendVersion::P4_16);
 
-    ReferenceMap refMap;
-    TypeMap      typeMap;
-    PassManager  passes({
-        new TypeChecking(&refMap, &typeMap)
-    });
+    if (pgm != nullptr) {
```
mbudiu-vmw, line 93: why would this suddenly not parse?
if it used to parse should't you instead use an ASSERT_TRUE(pgm != nullptr)?

scweng: I think it never parses. It's just that all the visitors are safe against null pointer so when compiler doesn't specifically check, calling `apply` on the null program simply flow through all the visitors without any effect.

Calling methods on a null pointer is, however, an undefined behavior by th C++ standard and technically the program could do arbitrary things.

PR #2368, opened 2020-05-07
Structs, headers and tuples initialization/assignment support (#762) …

marko-stanoj:
…(#2017)

Struct, header or tuple can be initialized automatically with a default value of
the suitable type using the syntax `...` or using a mix of explicit values and
default values by using the notation `...` in a list expression
(and struct expression for structs and headers).

Syntax `{ ... }` and `{/* expressionList | kvList */, ... }` can also be used in assignments.

Default values are as proposed in p4-spec pull request #797.

Header stacks and varbits are not supported as a leaf fields of structs
nor elements of tuples if using this kind of initialization/assignment, because in current
version of spec there are no constructs to initialize these types.
If they are used as a leaf fields compiler will report an error.

If header union is used as an element of tuple, compiler will produce warning
that temporary header union variable used for the element initialization may not be
completely initialized, despite setting it invalid.

along with assignment statemants, which is the cause of changes in expected
outputs of some tests.


mbudiu-vmw:
Before I review this I want to mention that during the last design working group we have briefly discussed the possibility of using `{}` to indicate an invalid header. If we did adopt this syntax (or some other equivalent syntax) it could probably simplify this PR...

frontends/p4/structInitializers.cpp
```diff
@@ -134,4 +322,44 @@ const IR::Node* CreateStructInitializers::postorder(IR::Operation_Relation* expr
     return expression;
 }
 
+const IR::Node* CreateStructAssignInitializers::postorder(IR::AssignmentStatement* statement) {
```
mbudiu-vmw, line 325: will your code handle structure expressions that are not used just as initializers?
e.g. `if (x == { a = 2, ... })`?

mbudiu-vmw: if not, we should make sure that the spec does not allow this.

frontends/p4/structInitializers.cpp
```diff
@@ -13,10 +13,166 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
...
+    auto strct = ltype->to<IR::Type_StructLike>();
+    if (auto si = statement->right->to<IR::StructExpression>()) {
+        for (auto f : strct->fields) {
+            auto right = si->components.getDeclaration<IR::NamedExpression>(f->name);
+            auto left = new IR::Member(statement->left, f->name);
+            if (f->type->is<IR::Type_Header>()) {
+                auto le = right->expression->to<IR::ListExpression>();
+                auto se = right->expression->to<IR::StructExpression>();
+                if ((le && le->defaultInitializer) ||
+                    (se && se->defaultInitializer)) {
```
mbudiu-vmw, line 139: much of this code seems to be common with the previous function (not 100% clear at a cursory reading); could you refactor some of it?

frontends/p4/structInitializers.cpp
```diff
@@ -13,10 +13,166 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
...
+        expr = new IR::Constant(t->type, 0);
+    } else if (type->is<IR::Type_StructLike>() || type->is<IR::Type_BaseList>()) {
+        auto vec = new IR::Vector<IR::Expression>();
+        auto conv = convert(new IR::ListExpression(srcInfo, *vec, true), type);
+        return conv;
+    } else {
+        ::error(ErrorType::ERR_TYPE_ERROR,
+                "%1%: Unknown default value for type %2%", srcInfo, type);
+    }
+    return expr;
```
mbudiu-vmw, line 49: this may return an initialized variable expr; please initialize expr to nullptr.

frontends/p4/structInitializers.cpp
```diff
@@ -25,31 +181,71 @@ const IR::Expression*
...
+            if (le->defaultInitializer) {
+                if (index == 0 && isHdr) {
+                    hdrInvalid = true;
+                }
+                const IR::Expression* expr;
+                modified = true;
+                if (!hdrInvalid) {
+                    for (; index < st->fields.size(); index++) {
+                        auto f = st->fields.at(index);
+                        expr = defaultValue(f->type, expression->srcInfo);
```
mbudiu-vmw, line 209: you probably should check if this returns nullptr

frontends/p4/structInitializers.cpp
```diff
@@ -25,31 +181,71 @@ const IR::Expression*
...
+                    auto ne = sli->components.getDeclaration<IR::NamedExpression>(f->name.name);
+                    BUG_CHECK(ne != nullptr, "%1%: no initializer for %2%", expression, f);
+                    auto convNe = convert(ne->expression, f->type);
+                    if (convNe != ne->expression)
+                        modified = true;
+                    ne = new IR::NamedExpression(ne->srcInfo, f->name, convNe);
+                    si->push_back(ne);
+                }
+            } else {
+                const IR::Expression* expr;
```
mbudiu-vmw, line 231: It does not look like you need this declaration here, please push it inwards.

frontends/p4/structInitializers.cpp
```diff
@@ -25,31 +181,71 @@ const IR::Expression*
...
+                modified = true;
+                for (auto f : st->fields) {
+                    auto ne = sli->components.getDeclaration<IR::NamedExpression>(f->name.name);
+                    if (ne == nullptr) {
+                        expr = defaultValue(f->type, expression->srcInfo);
```
mbudiu-vmw, line 236: nullptr check?

frontends/p4/typeChecking/typeUnification.cpp
```diff
@@ -346,22 +389,54 @@ bool TypeUnification::unify(const IR::Node* errorPosition,
...
+                    auto strctField = strct->getField(f->name);
+                    if (strctField == nullptr) {
+                        TypeInference::typeError("%1%: No field named %2% in structure %3%",
+                                                 errorPosition, f->name, strct);
+                        return false;
+                    }
+                }
+                for (const IR::StructField* f : strct->fields) {
+                    if (f->type->is<IR::Type_Varbits>() || f->type->is<IR::Type_Stack>()
+                        /*|| f->type->is<IR::Type_HeaderUnion>()*/) {
```
mbudiu-vmw, line 425: do you want to delete this comment?

frontends/parsers/p4/p4parser.ypp
```diff
@@ -1357,9 +1359,13 @@ expression
     | dotPrefix nonTypeName              { $$ = new IR::PathExpression(new IR::Path(*$2, true)); driver.structure->clearPath(); }
     | expression "[" expression "]"      { $$ = new IR::ArrayIndex(@1 + @4, $1, $3); }
     | expression "[" expression ":" expression "]" { $$ = new IR::Slice(@1 + @6, $1, $3, $5); }
+    | "{" expressionList "," "..." "}"   { $$ = new IR::ListExpression(@1 + @5, *$2, true); }
     | "{" expressionList "}"             { $$ = new IR::ListExpression(@1 + @3, *$2); }
+    | "{" "..." "}"                      { $$ = new IR::ListExpression(@1 + @3, *(new IR::Vector<IR::Expression>()), true); }
     | "{" kvList "}"                     { $$ = new IR::StructExpression(
                                                   @1 + @3, IR::Type::Unknown::get(), (IR::Type_Name*)nullptr, *$2); }  // experimental
+    | "{" kvList "," "..." "}"                     { $$ = new IR::StructExpression(
```
mbudiu-vmw, line 1367: I guess this is no longer experimental, that comment (and the previous one) can be removed.

ir/expression.def
```diff
@@ -390,21 +390,47 @@ class ConstructorCallExpression : Expression {
...
+        }
+    }
+
     validate { components.check_null(); }
     size_t size() const { return components.size(); }
     void push_back(Expression e) { components.push_back(e); }
 }
 
-/// An expression that evaluates to a struct.
+/// IR representation of an initializer for a struct.
```
mbudiu-vmw, line 420: This is not entirely correct, it can also be a structure-valued expression.

midend/copyStructures.cpp
```diff
@@ -13,7 +13,7 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 See the License for the specific language governing permissions and
 limitations under the License.
 */
-
+#include <iostream>
```
mbudiu-vmw, line 16: is this needed?

frontends/p4/structInitializers.cpp
marko-stanoj: Currently that use case is not covered at all, I made an omission. Do we want to support those use cases, or compiler should report an error?

frontends/p4/structInitializers.cpp
marko-stanoj: Function defaultValue() will return adequate default value for type passed as an argument, or report error and return nullptr. Why do we need nullptr check here, when error is reported and we won't use those named expressions anyway?

mbudiu-vmw: You don't want the compiler to crash after reporting an error. And defensive programming can prove useful when code moves around.

marko-stanoj:
Hi @mbudiu-vmw , I applied requested changes. Please leave a comment if you want anything more from me in order to merge this PR.

mbudiu-vmw:
250 commits will pollute the history, can you please rebase on main in the recommended way instead?
https://github.com/p4lang/p4c/tree/main/docs#git-usage

Jovana-Syrmia:
@mbudiu-vmw Yes, I made a mistake, sorry, I'm currently working on that.

Jovana-Syrmia:
This PR is now in accordance with the main and a test for using this initialization in operations is added. Leave a comment if more change is needed.

mbudiu-vmw:
Did you address the previous requests?

ir/expression.def
```diff
@@ -404,21 +404,52 @@ class ConstructorCallExpression : Expression {
 
 /// Represents a list of expressions separated by commas
 class ListExpression : Expression {
```
mbudiu-vmw, line 406: You can add a new interface IHasDefaultInitializer with a method bool hasDefaultInitializer() and have all these classes implement that interface.

ir/type.def
```diff
@@ -300,6 +300,16 @@ class Type_Struct : Type_StructLike {
 /// and some information about their types.
 class Type_UnknownStruct : Type_StructLike {
 #nodbprint
+    bool defaultInitializer = false;
```
mbudiu-vmw, line 303: Let's call this "incomplete" rather than "defaultInitializer".
It is the type of an expression containing a default initializer (for now).

frontends/p4/typeChecking/typeUnification.cpp
```diff
@@ -296,32 +295,73 @@ bool TypeUnification::unify(const EqualityConstraint* constraint) {
...
+                auto si = ts->components.at(i);
+                auto di = td->components.at(i);
+                constraints->add(constraint->create(di, si));
+            }
+        } else {
+            for (size_t i=0; i < td->components.size(); i++) {
+                auto di = td->components.at(i);
+                if (di->is<IR::Type_Varbits>() || di->is<IR::Type_Stack>()) {
+                    return constraint->reportError(constraints->getCurrentSubstitution(),
+                                                "Structs, headers and tuples "
```
mbudiu-vmw, line 317: frankly this is not really the job of the type unification. that error should be given somewhere else.
type unification should only say whether the two types can be made equal. moreover, if we decide to have default initializers for these types this code should probably not change.

frontends/p4/typeChecking/typeUnification.cpp
```diff
@@ -296,32 +295,73 @@ bool TypeUnification::unify(const EqualityConstraint* constraint) {
...
+                                                "containing %1% fields cannot be initialized "
+                                                "using ...: %2% = {...}.",
+                                                 td->components.at(i), td);
+                } else if (di->is<IR::Type_Struct>() || di->is<IR::Type_Header>()
+                           || di->is<IR::Type_Tuple>()) {
+                    IR::Type_List* src_tmp =
+                                       new IR::Type_List(*(new IR::Vector<IR::Type>()), true);
+                    constraints->add(constraint->create(di, src_tmp));
```
mbudiu-vmw, line 325: can this ever fail? if not, there is no need to do it.

frontends/p4/typeChecking/typeUnification.cpp
```diff
@@ -296,32 +295,73 @@ bool TypeUnification::unify(const EqualityConstraint* constraint) {
...
+                }
```
mbudiu-vmw, line 326: probably you need an `else return false;` or some error report

frontends/p4/typeChecking/typeUnification.cpp
```diff
@@ -334,23 +374,52 @@ bool TypeUnification::unify(const EqualityConstraint* constraint) {
...
-                    { stField->type, f->type, f, strct });
-                constraints->add(c);
+                    st->fields.size(),
+                    strct->fields.size(), st, strct);
+            }
+            if (!defaultInitializer) {
+                for (const IR::StructField* f : strct->fields) {
+                    auto stField = st->getField(f->name);
+                    if (stField == nullptr) {
+                        return constraint->reportError(constraints->getCurrentSubstitution(),
```
mbudiu-vmw, line 393: this code is not about initialization, it is about matching types. the errors should be about not being able to unify two types.

mbudiu-vmw (changes requested):

frontends/p4/structInitializers.cpp
```diff
@@ -17,43 +17,247 @@ limitations under the License.
...
+    } else if (type->is<IR::Type_Boolean>()) {
+        expr = new IR::BoolLiteral(false);
+    } else if (type->is<IR::Type_Error>()) {
+        auto path = new IR::Path(IR::ID(srcInfo, "error"));
+        auto typeName =  new IR::Type_Name(srcInfo, path);
+        auto id = new IR::ID(srcInfo, "NoError");
+        auto tne = new IR::TypeNameExpression(srcInfo, typeName);
+        expr = new IR::Member(srcInfo, tne, *id);
+    } else if (auto t = type->to<IR::Type_Enum>()) {
+        BUG_CHECK(t->members.size(), "unknown default value for enum with no members");
```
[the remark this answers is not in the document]
jboskovic: I'm not sure how can the user declare an enum with no members.
Wouldn't parser complain because it has an enum declaration defined as a list of names that cannot be empty?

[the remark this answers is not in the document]
Jovana-Syrmia: I'm not sure how the user can declare an enum with no members.
Wouldn't parser complain because it has an enum declaration with a list of names that cannot be empty?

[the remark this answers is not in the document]
mbudiu-vmw: If the spec says that it can't be empty then having a BUG is ok.

frontends/p4/structInitializers.cpp
```diff
@@ -65,36 +269,47 @@ convert(const IR::Expression* expression, const IR::Type* type) {
...
-}
-
 const IR::Node* CreateStructInitializers::postorder(IR::ReturnStatement* statement) {
     if (statement->expression == nullptr)
         return statement;
     auto func = findOrigCtxt<IR::Function>();
     if (func == nullptr)
         return statement;
 
+    if (isDefaultInit(statement->expression)) {
```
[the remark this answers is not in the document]
Jovana-Syrmia: Should I do the same thing for operation relations in order for them to be supported with default initialization?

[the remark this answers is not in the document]
mbudiu-vmw: That may be a tricky case. Consider something like this: `{ a=1, ... } == { b = 2, ... }`. 
This part should probably be left for later.

Jovana-Syrmia:
Hi @mbudiu-vmw, I applied the requested changes. 
For operation relations, I left it to be an unsupported operation.
For now, default initializations are implemented as separate assignments.
Because of the use in arguments and return statements, it's a repeated pass. If a literal for an invalid header were to be introduced, implementation would be more consistent with the current, not default initialization.
Please leave a comment if more changes are needed.

frontends/p4/structInitializers.cpp
```diff
@@ -17,43 +17,247 @@ limitations under the License.
...
+        auto left = new IR::Member(exp, hdr->name);
+        auto method = new IR::Member(srcInfo, left,
+                                     IR::Type_Header::setInvalid);
+        auto mc = new IR::MethodCallExpression(srcInfo, method, new IR::Vector<IR::Argument>());
+        return new IR::MethodCallStatement(mc->srcInfo, mc);
+    }
+}
+
+// rexpr is struct expression or list expression used to assign a value to struct or tuple
+// respectively; ltype is destination type
```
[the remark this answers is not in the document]
mbudiu-vmw: We usually use JavaDoc-like comments, using /// and `@param` to describe parameters:
```
/// @param rexpr  struct expression or list expression...
```

frontends/p4/structInitializers.h
```diff
@@ -19,34 +19,66 @@ limitations under the License.
...
     const IR::Node* postorder(IR::Operation_Relation* expression) override;
     const IR::Node* postorder(IR::Declaration_Variable* statement) override;
     const IR::Node* postorder(IR::ReturnStatement* statement) override;
-};
+    const IR::Node* postorder(IR::AssignmentStatement* statement) override;
+    const IR::Node* postorder(IR::P4Control* control) override;
 
+    static const IR::Expression* defaultValue(const IR::Type* type, const Util::SourceInfo srcInfo);
+};
+/// first repeat splits default initializations into assignments with the default values
```
mbudiu-vmw, line 48: I would think that these should be two different passes: one which creates a temporary for every structexpression that shows up in an inconvenient place, and one which expands the initializers - which you already wrote. 
This should probably not entail much code writing, just reorganizing the existing code.
What do you think?

ir/expression.def
```diff
@@ -404,21 +404,54 @@ class ConstructorCallExpression : Expression {
...
+        initUnknown();
+    }
+
     inline Vector<Expression> components;
     ListExpression {
         validate();
+        initUnknown();
+    }
+
+    void initUnknown(){
```
mbudiu-vmw, line 425: In general the `type` field of an expression is not maintained by the compiler. It is only populated by TypeInference when invoked with a boolean flag.
Is this field used for the ListExpression?

frontends/p4/structInitializers.h
mbudiu-vmw: In fact, this approach may make it easy to handle comparisons too - you will no longer have comparisons with default initializers. Creating the temporaries will require the type to be known - which is only possible if the type of the structexpression can be inferred. So I think the whole structure of the pass will be simplified.

mbudiu-vmw (changes requested):

ir/type.def
```diff
@@ -300,14 +300,14 @@ class Type_Struct : Type_StructLike {
 /// and some information about their types.
 class Type_UnknownStruct : Type_StructLike {
 #nodbprint
-    bool defaultInitializer = false;
-    Type_UnknownStruct(Util::SourceInfo srcInfo, IR::ID name, IR::IndexedVector<IR::StructField>& fields, bool defaultInitializer)
-                       : Type_StructLike(srcInfo, name, fields), defaultInitializer(defaultInitializer) {
+    bool incomplete = false;
```
mbudiu-vmw, line 303: please document the field incomplete.

testdata/p4_16_errors/default-values-operations-err.p4
```diff
@@ -12,9 +12,6 @@ struct S {
     bool   b;
 }
 
-S f<D>(in D data) {
```
mbudiu-vmw: why delete this?

frontends/p4/structInitializers.h
```diff
@@ -33,52 +33,47 @@ class CreateStructInitializers : public Transform {
...
-        ///     returnTmp.i = 0;
-        ///     returnTmp.b = false;
-        ///     /*ommited*/
-        ///     return returnTmp;
-        /// }
-        auto passRepeat = new PassRepeated{new TypeChecking(refMap, typeMap),
-                                    new CreateStructInitializers(refMap, typeMap)};
-        passRepeat->setRepeats(2);
-        passes.push_back(passRepeat);
+        passes.push_back(new TypeChecking(refMap, typeMap));
```
mbudiu-vmw, line 72: the comment in your previous commit was actually good, can you preserve it and adapt it to the current structure?

mbudiu-vmw (approved):
I am approving, but I still have a couple of small requests.
I also have a question about a change in a test case.

testdata/p4_16_errors/default-values-operations-err.p4
Jovana-Syrmia: Call to this function with default values is enabled now. I left this file in p4_16_errors by accident. I removed it and added a new test with this function in p4_16_samples.

Jovana-Syrmia:
Hi @mbudiu-vmw, I applied the requested changes but I don't know why my tests failed on validation.

fruffy:
These validation failures should normally be caught, I will take care of this.

Jovana-Syrmia:
Hi @mbudiu-vmw. All checks are passing now, is there anything more you want me to do in order to merge this solution?

testdata/p4_16_samples_outputs/default-values-operations-1.p4-stderr
```diff
@@ -0,0 +1,18 @@
+[--Wwarn=uninitialized_use] warning: returnTmp may not be completely initialized
+default-values-operations-1.p4(27): [--Wwarn=uninitialized_use] warning: hdr_or_hu may not be completely initialized
```
mbudiu-vmw, line 2: I find these error messages puzzling. Why are the objects not initialized?

Jovana-Syrmia: returnTmp is a structer initialized with {b=false, ....} (which means field H is initialized with {...}). Specification says that default value for header, initialized like that, is to set header to be Invalid. So we have a structure that has a header field set to invalid, that's why warning is given that struct may not be completely initialized. Same goes for hdr_or_hu, that's also a header initialized with {...}.

mbudiu-vmw: An invalid header is a properly initialized value.
However, I assume that this happens in a later statement.

mbudiu-vmw: This warning comes from the def-use analysis pass.

mbudiu-vmw: This warning signals either a bug in this pass or in the def-use pass. We should figure out which one is true and fix it.
Technically the code generated by this pass should have no uninitialized structs or headers - in the -first.p4 file everything seems to be properly initialized. Can you create a small reproduction just for this case?

mbudiu-vmw: If the bug is in the def-use analysis then we can merge this code.

mbudiu-vmw: Here is the generated code:
```
H hdr_or_hu;
H hdr_or_hu_0;
rightTmp_0 = { false, hdr_or_hu, (S){i = 1w0,h = hdr_or_hu_0,b = false} };
```
hdr_or_hu is not set to invalid. I suspect that this is the problem.

Jovana-Syrmia: I understand, i overlooked that invalid header is a properly initialized value.
This happens because, whenever a header is initialized with {...}, StructInitializer will just set declaration variable initializer to empty, and leave it to ResetHeader pass to set it to invalid.  Then SimplifyDefUse sees a declaration without an initialization and generates a warning.

Jovana-Syrmia: First implementation had a set invalid header inside StructInitializer, but then after the pass ResetHeaders, there were redudant setInvalid() calls, so i thought this resolved the problem. I will try to go back to that solution and correct it a little to have just one call.

mbudiu-vmw: So do you have a plan to fix this or should we do it somewhere else?

Jovana-Syrmia: Problem is that ResetHeaders sets to invalid every header that has an empty declaration. Tuples, that have headers, are initialized with headers extracted in declaration variable. I think thats the main problem, if this is done differently, StructInitializer could set headers to invalid without extracting it to declaration and ResetHeaders wouldnt put redudant call. I will think about this solution and try to implement it.

mbudiu-vmw: the right solution may be to improve the ResetHeaders pass or to run it sooner.

Jovana-Syrmia: Easiest solution would be to put ResetHeaders before SimplifyDefUse, if it can be done.

mbudiu-vmw: Feel free to try that if it's not too much work. Hopefully it won't affect too many reference outputs either.

Jovana-Syrmia: I found that this problem was already present before my changes. 
This function segment also throws a warning because returnTmp.h is set to invalid and SimplifyDefUse doesn't see this as initialization.
```
S returnTmp;
returnTmp.i = 1;
returnTmp.h.setInvalid();
returnTmp.b = false;
return returnTmp;
```
```
[--Wwarn=uninitialized_use] warning: returnTmp may not be completely initialized
    return returnTmp;
           ^^^^^^^^^
```
Also if we have 
```
H h; 
h.setInvalid()
```
the same warning is thrown.

Jovana-Syrmia:
Hi @jfingerh.
The problem is that ... and {...} calls are different just in case of a header initialization, in other cases like a struct or tuple initialization or in the future for enums, these two calls act the same. Is it okay to have this inconsistency?

.. and {...} are implemented as ListExpressions. I can add a new variable to distinguish these two expressions and implement it differently when used in headers. But, if an invalid header literal is to be introduced is there a need for ... to do the same thing?

I suggest that ... and {...} do the same thing (set all header fields to default values) and when an invalid literal is introduced use it to set headers to invalid.

Yes, you are right. This PR does not yet handle enums, neither the serializable nor the non-serializable kinds. It just handles structs, headers, and tuples default initialization, and if their fields are of type enum, serializable or non-serializable kind, it sets them to default values.

jfingerh:
@Jovana-Syrmia wrote in an earlier comment:

.. and {...} are implemented as ListExpressions. I can add a new variable to distinguish these two expressions and implement it differently when used in headers. But, if an invalid header literal is to be introduced is there a need for ... to do the same thing?

I suggest that ... and {...} do the same thing (set all header fields to default values) and when an invalid literal is introduced use it to set headers to invalid.

My response:

The current P4_16 language specification currently says that `...` should initialize a header to invalid, and `{...}` should initialize a header to valid, with default values for all fields.

As far as how this might best be implemented in p4c, I defer to @mbudiu-vmw for ideas there, as I am fairly ignorant of the p4c compiler IR.  If he agrees with your idea that `...` assigned to a header is the same as assigning `{...}` to a header (i.e. both make the header valid), then we should change the P4_16 language spec to match what is implemented in p4c at some point in the future.

It has not yet been decided if there will be a literal syntax for an invalid header, and if so, what that syntax will be.  See this issue open for the language spec: https://github.com/p4lang/p4-spec/issues/1021

mbudiu-vmw:
I think that the spec is correct, we should follow that.

mbudiu-vmw:
We don't need to have the enums as part of this PR, should be easy to add later, but @Jovana-Syrmia will have to fix the problem with `...` for headers. I apologize for not having noticed this.

jfingerh:
@marko-stanoj @mbudiu-vmw @jnfoster I would be interested in discussing this PR, or any PR for p4c that would lead us towards `...` syntax being implemented according to the P4 language spec.

Marko, do you know which of these two PRs you consider more well developed at this time?  Do you consider either of them obsolete?  Were you being paid to work on this PR, Marko, and are no longer being paid to work on it?  Or have other reasons to no longer be interested in working on it?
+ https://github.com/p4lang/p4c/pull/2224
+ https://github.com/p4lang/p4c/pull/2368

Would any or all of you be interested in discussing this at the next P4.org open source developer days meeting, scheduled for 11am-noon Pacific time on Tuesday, October 25, 2022?  Details for joining the Zoom meeting are available on this calendar of events: https://p4.org/working-groups/

If that time is inconvenient, let me know and we can try to arrange a different time.

mbudiu-vmw:
This would be much easier if we had a literal for invalid headers.

jfingerh:
@mbudiu-vmw I understand, which is why I created this issue: https://github.com/p4lang/p4-spec/issues/1021

It might have been discussed once, but it led to a discussion that perhaps we should simply use `...` for that.  I am perfectly OK to create a different syntax that is a literal for an invalid header -- we just need to pick one and move forward.

mbudiu-vmw:
The way forward is for me to make up a syntax for invalid headers and perhaps stacks too and implement that.
Then we can see it works and we can just argue about what the syntax should be.

jfingerh:
@mbudiu-vmw I did a quick check, and the string "invalid_header" does not appear anywhere in any of the p4c P4_16 test programs, except inside of a string `@noWarn("invalid_header")`, so `invalid_header` as a literal syntax for a value that is an invalid header would not cause any conflicts with existing names used in existing compiler test programs.  I am fine with other proposed literal syntax for experimentation purposes, too -- just throwing in my 2 cents.

mbudiu-vmw:
Superseded by #3968

PR #2378, opened 2020-05-15
DO NOT MERGE: Mysterious p4runtime backend behavior

smolkaj:
DO NOT SUBMIT.

This is merely to demonstrate the mysterious behavior detailed in issue #2379.


smolkaj:
Safe to close, since it is for demonstration purposes only.

PR #2380, opened 2020-05-15
Faster crash dumps

ChrisDodd:
- run addr2line through a pipe so it does not need to reload.

### DIFF
--- a/lib/crash.cpp
+++ b/lib/crash.cpp
@@ -72,50 +72,65 @@ static void sigint_shutdown(int sig, siginfo_t *, void *) {
  * else that might be problematic if there's memory corruption or exhaustion
  */
 const char *addr2line(void *addr, const char *text) {
-    int pfd[2], len;
-    pid_t child;
+MTONLY(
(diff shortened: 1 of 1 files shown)


lib/crash.cpp
```diff
@@ -72,50 +72,65 @@ static void sigint_shutdown(int sig, siginfo_t *, void *) {
...
+        p += t-text;
+        if (!memchr(text, '/', t-text)) *p++ = ')';
+        child = -1;
+#if HAVE_PIPE2
+        if (pipe2(pfd1, O_CLOEXEC) < 0) return 0;
+        if (pipe2(pfd2, O_CLOEXEC) < 0) return 0;
+#else
+        if (pipe(pfd1) < 0) return 0;
+        if (pipe(pfd2) < 0) return 0;
+        fcntl(pfd1[0], F_SETFD, FD_CLOEXEC | fcntl(pfd[0], F_GETFL));
```
mbudiu-vmw, line 102: this didn't compile on the mac. There's no pfd

PR #2385, opened 2020-05-19
Set headers valid when expanding initializations

### LINKED ISSUE #2383: Clarification question on uninitialized local headers
After the recent discussions and spec merges I have a (hopefully) last question on header initialization. Are these constructs equivalent?

```p4
ethernet_t tmp;
tmp = { 48w1, 48w1, 16w1 };

```

```p4
ethernet_t tmp;
tmp.dst_addr = 48w1;
tmp.src_addr = 48w1;
tmp.eth_type = 16w1;
```

My initial guess was **no**, since the in the first example the header ist set to valid, whereas in the second I assume the header stays invalid. Otherwise any assignment to an invalid header would set it to valid again.

Based on my assumptions I have observed potentially incorrect behavior in the "CopyStructures" pass. The example program this time is actually something a sane person might write:

```p4
ethernet_t do_function() {
    return  {1, 1, 1};
}
    apply {
        h.eth_hdr = do_function();
    }
}
```

With  `BMV2::SimpleSwitchMidEnd_19_CopyStructures`
```p4
    bool hasReturned;
    ethernet_t retval;
    apply {
        {
            hasReturned = false;
            hasReturned = true;
            retval = { 48w1, 48w1, 16w1 };
            h.eth_hdr = retval;
        }
    }
}
```

is converted to 

```p4
    bool hasReturned;
    ethernet_t retval;
    apply {
        {
            hasReturned = false;
            hasReturned = true;
            {
                retval.dst_addr = 48w1;
                retval.src_addr = 48w1;
                retval.eth_type = 16w1;
            }
            h.eth_hdr = retval;
        }
    }
}
```
Which lacks the `setValid` call. Maybe there is already a plan to fix this with the `setHeaders` pass?. 

[header_function_cast.p4.txt](https://github.com/p4lang/p4c/files/4643522/header_function_cast.p4.txt)
[header_function_cast.stf.txt](https://github.com/p4lang/p4c/files/4643524/header_function_cast.stf.txt)

mbudiu-vmw:
Fixes #2383


testdata/p4_14_samples_outputs/09-IPv4OptionsUnparsed-midend.p4
```diff
@@ -77,6 +77,7 @@ parser ParserImpl(packet_in packet, out headers hdr, inout metadata meta, inout
     @name(".parse_ipv4") state parse_ipv4 {
         tmp = packet.lookahead<bit<160>>();
         tmp_hdr_0.setValid();
+        tmp_hdr_0.setValid();
```
ChrisDodd, line 80: So it looks like we get redundant setValid calls in some cases?  Harmless, but do we know why this happens?

mbudiu-vmw: I guess this happens if the user was already setting the header to valid before assigning to it.
We should write an analysis which can remove the useless setValid; it could probably be folded in the def-use analysis.

mbudiu-vmw: See #2396

PR #2386, opened 2020-05-20
Support if statements in parsers

mbudiu-vmw:
@vgurevich : this is actually a two-line change to the compiler.
We should add this to the spec as well; right now it's labeled as experimental.

### DIFF
--- a/frontends/p4/frontend.cpp
+++ b/frontends/p4/frontend.cpp
@@ -152,6 +152,7 @@ const IR::P4Program *FrontEnd::run(const CompilerOptions &options, const IR::P4P
         new ValidateMatchAnnotations(&typeMap),
         new DefaultArguments(&refMap, &typeMap),  // add default argument values to parameters
         new BindTypeVariables(&refMap, &typeMap),
+        new RemoveParserControlFlow(&refMap, &typeMap),
         new StructInitializers(&refMap, &typeMap),
         new SpecializeGenericFunctions(&refMap, &typeMap),
         new TableKeyNames(&refMap, &typeMap),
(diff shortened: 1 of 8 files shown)


hesingh:
It would be good to add a packet test to this PR changes, thanks.

mbudiu-vmw:
I don't see any reason not to merge this.

hksoni:
I modified the example in 5.3 of the spec to show possible use of transition statement in body of If-stmt.
1. will transition statements be allowed in if-else body?
2. Currently states have only one explicit conditional, select expression, that can be only at the end of the state. Having multiple conditional would essential create a DAG within the state. If the transfer of control is outside the state using transition statement, compilers need to do some work to some more work.     
 
look at `parse_ipv4` state.
```
parser TopParser(packet_in b, out Parsed_packet p) {
    Checksum16() ck;  // instantiate checksum unit

    state start {
        b.extract(p.ethernet);
        transition select(p.ethernet.etherType) {
            0x0800: parse_ipv4;
            // no default rule: all other packets rejected
        }
    }

    state parse_ipv4 {
        b.extract(p.ip);
        if (p.ip.version == 4w4) {
            transition  my_error_state;
        }
        verify(p.ip.ihl == 4w5, error.IPv4OptionsNotSupported);
        ck.clear();
        ck.update(p.ip);
        // Verify that packet checksum is zero
        verify(ck.get() == 16w0, error.IPv4ChecksumError);
        transition accept;
    }
    state my_error_state {
        // update user-defined metadata
    }
}
```

mbudiu-vmw:
transition is still allowed only as the last statement of a state, and it cannot be conditional

PR #2388, opened 2020-05-20
Make IR DAG into tree to remove ambiguities

### LINKED ISSUE #2375: Crash when running end-to-end tests with simple switch
When doing end-to-end tests I have observed strange crashes with simple switch. 
I run the following command
`python p4_tv/p4c/backends/bmv2/run-bmv2-test.py p4_tv/p4c -v -b -bd p4c/build/  p4_tv/bugs/simple_switch/two_bools.p4 `

and I get this output:

```shell
Writing temporary files into  ./tmp1xr6wjtr                                        
Executing  p4c/build//p4c-bm2-ss -o ./tmp1xr6wjtr/two_bools.json p4_tv/bugs/simple_switch/two_bools.p4
Exit code  0
Check for  p4_tv/bugs/simple_switch/two_bools.stf
Running model
Running behavioral-model/targets/simple_switch/simple_switch --log-file switch.log --log-flush --use-files 0 --thrift-port 9972 --device-id 882 -i 0@pcap0 .././tmp1xr6wjtr/two_bools.json
Invalid 'type' in expression: ''
bad json:
{
   "left" : null,
   "op" : "d2b",
   "right" : {
      "type" : "field",
      "value" : [ "scalars", "tmp_0" ]
   }
}

*** Test failed
```
It looks like this might be a problem in the P4C BMV2 back-end and not the behavioral model?
I have two examples, they look related to me:

[two_bools.p4.txt](https://github.com/p4lang/p4c/files/4627728/two_bools.p4.txt)
[two_bools.stf.txt](https://github.com/p4lang/p4c/files/4627729/two_bools.stf.txt)
[two_bools_2.stf.txt](https://github.com/p4lang/p4c/files/4627731/two_bools_2.stf.txt)
[two_bools_2.p4.txt](https://github.com/p4lang/p4c/files/4627730/two_bools_2.p4.txt)

mbudiu-vmw:
The problem here is that the IR DAG has the same expression object in the LHS and RHS, but the code generated needs to be different.

This also fixes another clang warning.

Fixes #2375


backends/bmv2/common/lower.h
```diff
@@ -20,13 +20,14 @@ limitations under the License.
...
 #include "frontends/p4/typeMap.h"
 #include "frontends/common/resolveReferences/resolveReferences.h"
+#include "frontends/p4/cloner.h"
 
 namespace BMV2 {
 
 /**
   This pass rewrites expressions which are not supported natively on BMv2.
 */
-class LowerExpressions : public Transform {
```
ChrisDodd, line 29: Seems like this is actually a bug in some later pass that can't handle DAGs properly.  We should at least document which one(s), nothing where they fail when expression nodes appear in mulitple places in the DAG and why.

mbudiu-vmw: The translation to JSON uses a hashtable to map an IR node to its translation.
Potentially all passes that do something similar are vulnerable.

ChrisDodd: If the translation to JSON of a node depends on the parent context in which the node appears, then it shouldn't do that -- its a bug.  Other passes that do similar things are also buggy -- we should at least document all the potentially unsafe assumptions that passes make.

mbudiu-vmw: Reviewing the code for all passes would take a significant amount of time, so I am not sure when this could happen. In principle any time you use a set or map with a node as a key you have to worry about this, and there are a lot of cases. If we had more help with the front-end development I would consider doing this, but at this point I don't have enough time.

Unfortunately I didn't really internalize clearly the consequences of sharing nodes; even very early code (2015) that was written assumes sometimes implicitly that the IR DAG is a tree. Ideally we should have caught this in code review at the time.

mbudiu-vmw: It also doesn't help that writing tests for these conditions is very difficult.

ChrisDodd: All true and reasonable.  But if you know about specific issues in specific passes from the work done in this change, it would be good to add a comment to them documenting the dependency.

jafingerhut: This might defeat the purpose of the hashtable mapping IR nodes to other data, but would it be correct if instead it mapped a _path of IR nodes_, from a "root" to the IR node in question, to that data, instead of only the last IR node on that path?

mbudiu-vmw: We do have a data structure for that, it's called a "context", but it exists only while you are traversing in a visitor. You would need to copy it separately and add equality tests... quite a bit of work.

hesingh:
Please fix this phrase in the documentation:

"// WARNING: this assumes that each the same IR node has the same"

backends/bmv2/common/lower.h
hesingh: @mbudiu-vmw Please consider taking a small (size of code) frontend pass and make changes to it akin to this PR.  Then, the community could help with frontend code and also think of how to test.  thanks.

mbudiu-vmw: I don't understand this request.

Regarding @jafingerhut's question: it is not clear that you want equality testing for the path; you may still treat two nodes the same even if they have different paths; for example many passes treat left-values different from the other expressions, but similar to each other, no matter where they show up in the tree. So I don't know if such a data structure would actually solve all the problems.

hesingh: > I don't understand this request.
> 

Since you said "any time you use a set or map with a node as a key you have to worry about this, and there are a lot of cases.", I meant, why not look at std::map and std::set use in frontend pass and see how to fix the issue.

mbudiu-vmw:
Fixes #2390

Issue #2391, opened 2020-05-22
Enum Underlying Representation Using Typedef


CommitThis:
Apologies if I have missed something again but I am having some trouble with `p4c`. 
I am using version 1.2.0, targeting bmv2/v1model and using the simple switch.

I was expecting to be able to do something like:
```
typedef bit<16> EtherType_t;

enum EtherType_t EtherType_e {
    IPv4 = 0x0800
}
```

However this returns an error:
```
./simple.p4(15):syntax error, unexpected IDENTIFIER, expecting {
enum EtherType_t EtherType_e
                 ^^^^^^^^^^^
```
Compiling the following is ok though:
```
enum bit<16> EtherType_e {
    IPv4 = 0x0800
}
```

The p4-spec issue https://github.com/p4lang/p4-spec/issues/774 seemed to be somewhat relevant however @vgurevich used a similar syntax in their motivating example. I replicated this (providing a const int for the bit width) which didn't work either.

This isn't a big issue for me, I was just wondering If I had missed something.

Thanks

jnfoster:
The grammar in the [language specification](https://p4.org/p4-spec/docs/P4-16-v1.2.0.html#sec-enum-types) requires serializable `enum`s to be declared with this syntax:
```
enumDeclaration
    : optAnnotations ENUM name '{' identifierList '}'
    | optAnnotations ENUM BIT '<' INTEGER '>' name '{' specifiedIdentifierList '}'
    ;
```
I just checked and the grammar seems fine if we replace this with `typeRef`, so we should probably generalize this...

vgurevich:
Interesting... I didn't realize that the current grammar doesn't allow (a) bit<N>, where N is an expression and, more importantly  (b) int<N>. 

Replacing with typeRef seems like a good idea as long as we check that the underlying type is a scalar one with a specific bit width.

hesingh:
`typeRef` does not include `TYPEDEF` which is included in p4 lang grammar to support typedef.  `typeRef` includes `baseType` which has checks for `bit<W>` type.   I am working with latest p4c.

jnfoster:
@hesingh you don't want to literally write `typedef` in a serializable `enum`, so the token `TYPEDEF` is neither here nor there.

In the example, we want to refer to a name that was previously `typedef`'d. Hence `typeRef` is the right thing to put here: it includes both `baseType` and `typeName`.

hesingh:
[...]

PR #2394, opened 2020-05-23
Support typedef, int<W>, and expression for serEnum

hesingh:
See 

https://github.com/p4lang/p4c/issues/2391

Reviewers: Please see comments in the .h and .ypp files.


hesingh:
Regarding serious problem 1, I am testing the new `type` operator (section 7.5 of working p4-16 spec) using serenum-type.p4.   The other test file called serenum-typedef.p4 is testing `typedef`. 

At least I don't see a formal typeMap, as used in frontend code such as typeChecker.cpp, available in parser or parser driver code.   However, the lexer does build types from symbol map and if that is what you call as typeMap, I understand.  

I agree, a general solution of nested typedefs or `type` makes sense to be out of the parser.  However, how do you deal with calling constructor of Type_Serenum in the parser if the serEnum type is not Type_Bits and is instead a Type_Name, due to use of typedef or `type`?

https://github.com/p4lang/p4c/blob/master/frontends/parsers/p4/p4parser.ypp#L1080

We cannot change constructor of `Type_SerEnum` because the Type_Bits is specified in p4runtime protobuf.   

This is why I suggested that we add a boolean flag to a regular enum class, and in parser code, when a serEnum needs creation, just create a regular enum with boolean flag set.  Using regular enum gets around trying to find Type_Bits from Type_Name.  Then, in a frontend pass, the regular enum with boolean flag set, is swapped with a serEnum. The frontend code (p4c/frontends/p4) knows how to get Type_Bits from any typedef or `type`.  Any other ideas are welcome.

hesingh:
Before the next round of changes are published, I am adding design notes for the changes here.
All facts in the notes are to the best of my understanding.

1. Once a serEnum's type is tydefed or "typed" in a P4 program, the p4c parser yacc code sees the type as a `Type_Name`.  There is no easy means in yacc to convert `Type_Name` to `Type_Bits` needed by serEnum constructor.  So, typde.def serEnum class is changed as follows:

```
$ git diff ir/type.def
diff --git a/ir/type.def b/ir/type.def
index 3f5458cc..6be5a628 100644
--- a/ir/type.def
+++ b/ir/type.def
@@ -461,7 +461,8 @@ class SerEnumMember : Declaration, CompileTimeValue {
 /** A serializable enumeration with a backing type */
 class Type_SerEnum : Type_Declaration, ISimpleNamespace, IAnnotated {
     optional Annotations annotations = Annotations::empty;
-    Type_Bits type;
+    NullOK Type_Bits type;
+    optional NullOK Type_Name nameType;
     inline IndexedVector<SerEnumMember> members;
     Util::Enumerator<IDeclaration>* getDeclarations() const override {
```

Once a `typeRef` is detected for `enumDeclaration` in yacc code, the new constructor with `nameType` is used.  Any p4runtime concerns with changing type.def?

2.  Moving on to frontend code, a `preorder(IR::Type_SerEnum*)` is added to `typeChecker.h` Transform class.  In the implementation of the new preorder method, the declaration is found using refMap and eventually the `Type_Bits` for use with serEnum is found and the IR tree nodes are swapped.  Old node has nullptr for `type` and valid `nameType`.  New node has valid type and nullptr `nameType`.

3.  Next set of changes relate to how `make check` tests each P4 program.  The Python script for each test pretty prints (using the PrettyPrint pass) the P4 program first.  PrettyPrint calls `ToP4` to output the P4 program.  However, `ToP4` finds the serEnum->type to be null and fails to write the type for serEnum which is incorrect.  So a new Transform class in added to `toP4.cpp` to find `Type_Bits` from `Type_Name` and then write the serEnum.  To find the mapping of one type to another the refMap is needed.  So, ToP4 has new constructors to with refMap and typeMap passed in.  The typeMap may have future use in case typeMap->getTypeType() is needed.  

4. Now that PrettyPrint needs refMap and TypeMap, the PrettyPrint pass is moved down in frontend passes list when both refMap and typeMap are available.  Such a move causes many test cases outputs to change.  A sample output change is shown below.  Is the changed output acceptable because the PrettyPrint passes find change data to print.

```
$ git diff testdata/p4_16_samples_outputs/table-key-serenum.p4
diff --git a/testdata/p4_16_samples_outputs/table-key-serenum.p4 b/testdata/p4_16_samples_outputs/table-key-serenum.p4
index bed4b19f..24fabcdc 100644
--- a/testdata/p4_16_samples_outputs/table-key-serenum.p4
+++ b/testdata/p4_16_samples_outputs/table-key-serenum.p4
@@ -43,7 +43,8 @@ control c(inout Headers h, inout standard_metadata_t sm) {
             h.eth.type: exact;
         }
         actions = {
-            do_act;
+            do_act();
         }
         const entries = {
                         EthTypes.IPv4 : do_act(0x800);
@@ -52,6 +53,7 @@ control c(inout Headers h, inout standard_metadata_t sm) {
 
         }
 
+        default_action = NoAction();
     }
     apply {
         tns.apply();
```

hesingh:
Having mulled over today about PrettyPrint pass I realized, PrettyPrint can just print the typedef or `type` in the P4 program instead of the original `bit<16>` type.  This way I can avoid changing toP4.h[.cpp] and frontend.cpp files and change 400 existing test cases for p4c.

For example, if the P4 code is `typedef bit<16> EthT; enum EthT EthTypes {};`, let PrettyPrint generate `enum EthT EthTypes {};`.  

So, I will backout the toP4.h[.cpp] and frontend.cpp in next checkin.

mbudiu-vmw:
I would suggest discussing this in the design group before implementing it.
Why is this feature needed?

hesingh:
> I would suggest discussing this in the design group before implementing it.
> Why is this feature needed?

Ok.  
This issue is being worked on by this PR: https://github.com/p4lang/p4c/issues/2391

hesingh:
No `p4c/control-plane` changes are needed because the frontend TypeChecker code changes any typedef to `bit<W>` or `int<W>`.  Likewise, the frontend ConstantFolding pass changes expression to W in `bit<W>` or `int<W>`.   Both `bit<W>` and `int<W>` are supported by `Type_Bits` in p4c.

frontends/parsers/p4/p4parser.ypp
```diff
@@ -1074,10 +1074,15 @@ enumDeclaration
...
         ENUM name { driver.structure->declareType(*$3); }
         "{" identifierList "}" { $$ = new IR::Type_Enum(@3, *$3, *$6); }
-    | optAnnotations ENUM BIT l_angle INTEGER r_angle name { driver.structure->declareType(*$7); }
-        "{" specifiedIdentifierList "}" {
-            auto type = IR::Type::Bits::get(@3+@6, parseConstantChecked(@5, $5));
-            $$ = new IR::Type_SerEnum(@7, *$7, type, *$10); }
+    | optAnnotations ENUM typeRef name { driver.structure->declareType(*$4); }
+	 "{" specifiedIdentifierList "}" {
+              auto type = $typeRef;
+              if (auto tn = type->to<IR::Type_Name>()) {
```
mbudiu-vmw, line 1080: You should not check the type here; this should be left for the typechecker.
Also, this is wrong, because if neither condition is true $$ will be left uninitialized.

ir/type.def
```diff
@@ -461,7 +461,8 @@ class SerEnumMember : Declaration, CompileTimeValue {
 /** A serializable enumeration with a backing type */
 class Type_SerEnum : Type_Declaration, ISimpleNamespace, IAnnotated {
     optional Annotations annotations = Annotations::empty;
-    Type_Bits type;
+    NullOK Type_Bits type;
```
mbudiu-vmw, line 464: This should instead just have a Type field.

frontends/p4/toP4/toP4.cpp
```diff
@@ -351,7 +351,10 @@ bool ToP4::preorder(const IR::Type_SerEnum* t) {
     dump(1);
     visit(t->annotations);
     builder.append("enum ");
-    visit(t->type);
+    if (t->type)
```
mbudiu-vmw, line 354: This should not be necessary with the changes proposed below.

frontends/p4/typeChecking/typeChecker.cpp
```diff
@@ -1755,6 +1755,31 @@ const IR::Node* TypeInference::preorder(IR::EntriesList* el) {
     return el;
 }
 
+const IR::Node* TypeInference::preorder(IR::Type_SerEnum* se) {
+    const IR::Type_Bits* type;
+    if (se->type == nullptr) {
```
mbudiu-vmw, line 1760: Here you should just lookup `se->type` in the typeMap and check it is one of the allowed types.

frontends/p4/toP4/toP4.cpp
hesingh: It is necessary.  This is a note I added in a prior comment of this PR.

"Next set of changes relate to how `make check` tests each P4 program. The Python script for each test pretty prints (using the `PrettyPrint` pass) the P4 program first. `PrettyPrint` calls ToP4 to output the P4 program. However, ToP4 finds the serEnum->type to be null and fails to write the type for serEnum which is incorrect."

frontends/p4/typeChecking/typeChecker.cpp
hesingh: From the `p4parser.ypp` code, either `se->type` is `nullptr` or  `se->nameType` is `nullptr`.  So this check is checking for `se->type` being `nullptr` so that the `nameType` can be converted to Type_Bits.  I will add a comment above the check to explain.

hesingh:
> You will have to address this in the P4Runtime code, because as @antoninbas has explained, if the type is not `bit<>` today this will generate an incorrect output. Instead you should give an error message that this feature is currently unsupported.

With last checkin, please see a new P4 program to test  `int<W>`  serEnum with p4runtime.  See this line of code.

https://github.com/p4lang/p4c/pull/2394/files#diff-4466ad8bdbfcbedaa51b0521ac45d70fR120

I don't see any issue with p4info generation for `int<W>` .

frontends/p4/typeChecking/typeChecker.cpp
```diff
@@ -1755,6 +1755,37 @@ const IR::Node* TypeInference::preorder(IR::EntriesList* el) {
     return el;
 }
 
+// If e->type is Type_Bits, this method doesn't do anything.
+const IR::Node* TypeInference::preorder(IR::Type_SerEnum* se) {
```
mbudiu-vmw, line 1759: This code should be moved into the postorder for SerEnum.
Then it becomes much simpler; you do `typeMap->getTypeType(se->type)` and check that it is `bit<>` or `int<>` and give an error otherwise. Also, your error messages are not clear, the error should be "type %1% cannot be used for the underlying type of an enum"

hesingh: It's tricky to change the postorder method because without preorder, the SerEnum input into the postorder method has type as Type_Name.  We would have to morph the SerEnum to be of Type_Bits.  If you see the preorder method, a new SerEnum is allocated to change the type.  

I have sent detailed logs in an email.

hesingh: This was the note in an email to @mbudiu-vmw.

Its tricky to remove the preorder and add code to postorder SerEnum.  These are the diffs for postorder as per your comment.

```p4
diff --git a/frontends/p4/typeChecking/typeChecker.cpp b/frontends/p4/typeChecking/typeChecker.cpp
index 3256483a..c0070473 100644
--- a/frontends/p4/typeChecking/typeChecker.cpp
+++ b/frontends/p4/typeChecking/typeChecker.cpp
@@ -1223,7 +1223,14 @@ const IR::Node* TypeInference::postorder(IR::Type_Enum* type) {
     return type;
}

const IR::Node* TypeInference::postorder(IR::Type_SerEnum* type) {
+    auto t = typeMap->getTypeType(type->type, true);
+    CHECK_NULL(t);
+    if (!t->is<IR::Type_Bits>()) {
+        typeError("%1%: cannot use type %2%", type, t);
+        return type;
+    }
     auto canon = setTypeType(type);
     for (auto e : *type->getDeclarations())
         setType(e->getNode(), canon);
```
But when I test with `issue2391.p4` which has a SerEnum using tydedef, The “type” coming in to the postorder method has type as Type_Name.  This is why I didn’t change rest of the postorder code using “type” because the “type” is incorrect.  

```p4
[2126] Type_SerEnum name=EthTypes declid=11
  annotations: [3] Annotations
  type: [171] Type_Name
    path: [170] Path name=EthT absolute=0
  [2132] SerEnumMember name=IPv4 declid=25
    value: [2139] Constant value=2048 base=16
      type: [151] Type_Bits size=16 isSigned=0
  [2142] SerEnumMember name=ARP declid=26
    value: [2149] Constant value=2054 base=16
      type: [151] Type_Bits size=16 isSigned=0...
  [2152] SerEnumMember name=RARP declid=27
    value: [2159] Constant value=32821 base=16
      type: [151] Type_Bits size=16 isSigned=0...
  [2162] SerEnumMember name=EtherTalk declid=28
    value: [2169] Constant value=32923 base=16
      type: [151] Type_Bits size=16 isSigned=0...
  [2172] SerEnumMember name=VLAN declid=29
    value: [2179] Constant value=33024 base=16
      type: [151] Type_Bits size=16 isSigned=0...
  [2182] SerEnumMember name=IPX declid=30
    value: [2189] Constant value=33079 base=16
      type: [151] Type_Bits size=16 isSigned=0...
  [2192] SerEnumMember name=IPv6 declid=31
    value: [2199] Constant value=34525 base=16
      type: [151] Type_Bits size=16 isSigned=0...
```

This is why I was morphing the SerEnum with type of Type_Name in preorder with a “new” allocation of SerEnum.  If I add a new allocation of SerEnum in postorder, I get an error saying 

```bash
“In file: /home/hemant/int/p4c/frontends/p4/typeChecking/typeChecker.cpp:144
[31mCompiler Bug[0m: At this point in the compilation typechecking should not infer new types anymore, but it did.”
```

In summary, I think we could keep the preorder Serenum method.

mbudiu-vmw:
I think this is included in #2424, so we can close this.

PR #2395, opened 2020-05-26
Adding register support to PSA

yunheL:
Adding register support for PSA:
1. register_write().
2. register_read(). Syntax exposed to programmer is `dst = register_read(index)`. Internally, the compiler backend generate .json for `register_read(dst, index)`. This way, the generated .json shares the same .json format as v1model and is compatible with existing bmv2 backend.
3. Add register control plane commands support to STF: register_read, register_write, register_reset.  

Testing filename: testdata/p4_16_samples/psa-register-read-write-bmv2.p4 (.stf)

### DIFF
--- a/backends/bmv2/common/action.cpp
+++ b/backends/bmv2/common/action.cpp
@@ -47,7 +47,7 @@ void ActionConverter::convertActionBody(const IR::Vector<IR::StatOrDecl>* body,
         // TODO(jafingerhut) - add line/col at all individual cases below,
         // or perhaps it can be done as a common case above or below
         // for all of them?
-        
+
         IR::MethodCallExpression *mce2;
         auto isR = false;
         if (s->is<IR::AssignmentStatement>()) {
@@ -57,18 +57,18 @@ void ActionConverter::convertActionBody(const IR::Vector<IR::StatOrDecl>* body,
             r = assign->right;
             auto mce = r->to<IR::MethodCallExpression>();
             if (mce != nullptr) {
-                auto mi = P4::MethodInstance::resolve(mce, ctxt->refMap, ctxt->typeMap); 
+                auto mi = P4::MethodInstance::resolve(mce, ctxt->refMap, ctxt->typeMap);
                 auto em = mi->to<P4::ExternMethod>();
-                if (em != nullptr){
-                    if (em->originalExternType->name.name == "Register" || 
+                if (em != nullptr) {
+                    if (em->originalExternType->name.name == "Register" ||
                     em->method->name.name == "read") {
                         isR = true;
                         // l = l->to<IR::PathExpression>();
                         // BUG_CHECK(l != nullptr, "register_read dest cast failed");
                         auto dest = new IR::Argument(l);
                         auto args = new IR::Vector<IR::Argument>();
-                        args->push_back(dest); //dest
-                        args->push_back(mce->arguments->at(0)); //index
+                        args->push_back(dest);  // dest
+                        args->push_back(mce->arguments->at(0));  // index
                         mce2 = new IR::MethodCallExpression(mce->method, mce->typeArguments);
                         mce2->arguments = args;
                         s = new IR::MethodCallStatement(mce);
--- a/backends/bmv2/common/action.cpp
+++ b/backends/bmv2/common/action.cpp
@@ -47,6 +47,36 @@ void ActionConverter::convertActionBody(const IR::Vector<IR::StatOrDecl>* body,
         // TODO(jafingerhut) - add line/col at all individual cases below,
         // or perhaps it can be done as a common case above or below
         // for all of them?
+        
+        IR::MethodCallExpression *mce2;
+        auto isR = false;
+        if (s->is<IR::AssignmentStatement>()) {
+            auto assign = s->to<IR::AssignmentStatement>();
+            const IR::Expression *l, *r;
+            l = assign->left;
+            r = assign->right;
+            auto mce = r->to<IR::MethodCallExpression>();
+            if (mce != nullptr) {
+                auto mi = P4::MethodInstance::resolve(mce, ctxt->refMap, ctxt->typeMap); 
+                auto em = mi->to<P4::ExternMethod>();
+                if (em != nullptr){
+                    if (em->originalExternType->name.name == "Register" || 
+                    em->method->name.name == "read") {
+                        isR = true;
+                        // l = l->to<IR::PathExpression>();
+                        // BUG_CHECK(l != nullptr, "register_read dest cast failed");
+                        auto dest = new IR::Argument(l);
+                        auto args = new IR::Vector<IR::Argument>();
+                        args->push_back(dest); //dest
+                        args->push_back(mce->arguments->at(0)); //index
+                        mce2 = new IR::MethodCallExpression(mce->method, mce->typeArguments);
+                        mce2->arguments = args;
+                        s = new IR::MethodCallStatement(mce);
+                    }
+                }
+            }
+        }
+
         if (!s->is<IR::Statement>()) {
             continue;
         } else if (auto block = s->to<IR::BlockStatement>()) {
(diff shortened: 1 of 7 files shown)


midend/local_copyprop.h
```diff
@@ -131,6 +131,20 @@ class LocalCopyPropagation : public PassManager {
         if (!typeChecking)
             typeChecking = new TypeChecking(refMap, typeMap, true);
         passes.push_back(typeChecking);
+        policy =
```
hanw, line 134: Please move this to a separate psa backend pass.

ChrisDodd: This will break any backend that tries to use LocalCopyProp with a different policy.  This should be moved to backends/bmv2/psa_switch/midend.cpp, and passed as an argument to the LocalCopyProp constructor...

testdata/p4_16_samples/psa-register-read-write-bmv2.stf
```diff
@@ -0,0 +1,25 @@
...
+expect 3 000000000003 000000000000 ffff $
+
+packet 4 000000000003 000000000000 ffff
+expect 3 000000000003 000000000000 ffff $
+
+packet 0 000000000003 000000000000 ffff
+expect 3 000000000003 000000000000 ffff
+
+register_read cIngress.regfile 1
+# expect 3
```
ChrisDodd, line 15: should be `check_register cIngress.regfile(1) = 3`?

backends/bmv2/common/action.cpp
```diff
@@ -47,6 +47,36 @@ void ActionConverter::convertActionBody(const IR::Vector<IR::StatOrDecl>* body,
...
+        if (s->is<IR::AssignmentStatement>()) {
+            auto assign = s->to<IR::AssignmentStatement>();
+            const IR::Expression *l, *r;
+            l = assign->left;
+            r = assign->right;
+            auto mce = r->to<IR::MethodCallExpression>();
+            if (mce != nullptr) {
+                auto mi = P4::MethodInstance::resolve(mce, ctxt->refMap, ctxt->typeMap);
+                auto em = mi->to<P4::ExternMethod>();
+                if (em != nullptr) {
```
mbudiu-vmw, line 62: Does some pass make sure that reads are only used in assignments on the rhs?
You can have an expression `a = 2 + r.read(5)`

backends/bmv2/common/action.cpp
```diff
@@ -47,6 +47,36 @@ void ActionConverter::convertActionBody(const IR::Vector<IR::StatOrDecl>* body,
...
+                    if (em->originalExternType->name.name == "Register" ||
```
mbudiu-vmw, line 63: In general we avoid using strings in the code, and have these strings be defined as constants in a model header file. I don't recall whether there's a psaModel.h file.

hesingh: The bmv2 backend is using `frontends/p4/fromv1.0/v1model.h`.  This header has definition for a register. 

There is no psaMode.h file in p4c just yet.  I have used `find ./ -name "*model*.h"` and `find ./ -name "*Model*.h"` to check. For now, we could add `#include "frontends/p4/fromv1.0/v1model.h"` to actions.cpp.

testdata/p4_16_samples/psa-register-read-write-bmv2.stf
yunheL: Thanks! I use `register_read` because in the help menu of psa_switch_CLI (behavioral-model/targets/psa_switch), there is a `register_read` RuntimeCmd. I modified STF a little so that it can take in this RuntimeCmd for register.

backends/bmv2/common/action.cpp
yunheL: Thanks! I am adding a pass to psa backend to convert all `r.read()` into  `dst = r.read(index)` format. So that expression of arbitrary complexity can be handled.

backends/bmv2/common/action.cpp
yunheL: Thanks! I found that there is a `psa.p4` file in p4c/p4include which is the model header file for psa. I will have the strings define in this file.

yunheL: Thanks! I will try to see whether there is a model header file for psa. If not, I can create one.

hesingh: Note p4c/p4include/psa.p4 is a PSA switch architecture file.  In this file, externs supported by PSA are defined.  The model file is a .h file, not a .p4 file.  As I have said above, the model file being used by simple_switch is v1model.h.  v1model.h is included by several .cpp programs in p4c/backends/bmv2.   See if existing v1model.h serves your code changes.  If not define a new one, say, psamodel.h.

testdata/p4_16_samples/psa-register-read-write-bmv2.stf
yunheL: Hi Chris, sorry I might have misunderstood you earlier. I think I now understand: you mean we need to automate the check of the result of the register_read command (like assert in c). I don't think STF has implemented check_register yet. But I think it would be nice to have I think look into it.

backends/bmv2/common/action.cpp
mbudiu-vmw: Creating psa.h from scratch may be a lot of work, this is not really necessary to merge this PR.

yunheL:
> I don't see a corresponding PR in bmv2 for adding a `PSA_Register` extern implementation: https://github.com/p4lang/behavioral-model/tree/master/targets/psa_switch/externs ?

Hi Antonin, I changed p4c backend so that p4c backend can generate .json for PSA registers in the same format as v1model registers. Thus, currently, even no change on bmv2 is made, bmv2 can still support PSA registers. Thanks!

antoninbas:
> > I don't see a corresponding PR in bmv2 for adding a `PSA_Register` extern implementation: https://github.com/p4lang/behavioral-model/tree/master/targets/psa_switch/externs ?
> 
> Hi Antonin, I changed p4c backend so that p4c backend can generate .json for PSA registers in the same format as v1model registers. Thus, currently, even no change on bmv2 is made, bmv2 can still support PSA registers. Thanks!

I believe we should really follow the work that was previously done for the PSA counter extern (https://github.com/p4lang/behavioral-model/pull/767/files), and treat those PSA registers as generic `extern_instances` in the JSON and not as `register_arrays`. This is how we can eventually simply the process of adding new externs, and this is how we may be able to get to a point where the bmv2 p4c backend is practically architecture-independent. Eventually we may even be able to simplify the JSON format by removing `counter_arrays`, `meter_arrays`, `register_arrays`, etc.

hanw:
I spoke to Yunhe offline and explained the need to use `extern_instances`, instead of `register_arrays`. I think it is fine to let this PR merge. It allows us to enable PSA register support and move it out of critical path, while yunhe makes another PR to clean up the JSON generation and implement the externs in BMV2.

backends/bmv2/common/action.cpp
yunheL: Actually, I just realized one of the frontend passes `SideEffectOrdering` ensures all r.read will be transformed into the format `dst = r.read(index)`
e.g. `a = r.read(1) + r.read(2) - 5` will be translated as  
`tmp_0 = r.read(1);`  
`tmp_1 = r.read(2);`  
`a = tmp_0 + tmp_1 - 5;`

midend/local_copyprop.h
yunheL: Done moving to backends/bmv2/psa_switch/midend.cpp

yunheL:
(1) moved architecture specific code to p4c backend  
(2)frontend pass `SideEffectOrdering` transforms all r.read into the simple format `dst = r.read(idx)` thus r.read in arbitrary complex expressions can be handled.  
PSA register support should be good. More comments or merge is appreciated. Thanks!

PR #2400, opened 2020-05-28
update PSA action selector/profile tests to include action data

zma-barefoot:


zma-barefoot:
@mbudiu-vmw There are no (packet) tests for these -- need to write some. This PR is to improve the program -- action selection without action data not really make sense.

PR #2403, opened 2020-05-28
Add libbpf to maintain Ubuntu 16.04 compatibility

### LINKED ISSUE #1961: ebpf_common.h should not include linux/bpf.h
We do not need to depend on any kernel headers when doing user-level testing.

fruffy:
This pull request adds [libbpf](https://github.com/libbpf/libbpf) as dependency to make the eBPF back end more stable. It makes the generated code less dependent on the kernel headers of a particular distribution. This may help with the problem that different distributions of Ubuntu cannot compile the code generated by p4c-ebpf. Also fixes #1961. 

Let's see if this works out.

Changes:
- Added [libbpf](https://github.com/libbpf/libbpf) as dependency.
- Reenabled eBPF kernel tests.
- Added iproute2 to the Travis build.
- Runs Travis in privileged mode to allow the creation of network namespaces.
- Add stf test for `lpm_ebpf.p4`, also marked this test as xfail.
- Added the extern object tests to the user-space eBPF tests.
- Cleaned up the `ebpf_kernel.h` file, using only the imports from `libbpf`.
- Added `BPF_USER_MAP_UPDATE_ELEM` to the user-space test framework.
- Cleaned up the eBPF makefiles a little.


backends/ebpf/README.md
```diff
@@ -206,8 +208,7 @@ Additionally, the eBPF compiler test suite has the following python dependencies
 
 You can install these using:
 ```
-$ sudo apt-get install clang llvm libpcap-dev libelf-dev iproute2 net-tools
-$ sudo pip3 install pyroute2 ply==3.8 scapy==2.4.0
+$ pip3 install --user pyroute2 ply==3.8 scapy==2.4.0
```
mbudiu-vmw, line 211: why remove the apt-get line?

backends/ebpf/runtime/ebpf_kernel.h
```diff
@@ -23,30 +23,48 @@ limitations under the License.
...
 #define BACKENDS_EBPF_RUNTIME_EBPF_KERNEL_H_
 
-#include <linux/pkt_cls.h>  // TC_ACT_OK, TC_ACT_SHOT
-#include <string.h>         // memset()
-#include <unistd.h>         // syscall()
-#include <sys/syscall.h>    // __NR_bpf
 #include "ebpf_common.h"
 
 
+/* If we operate in user space we only need to include bpf.h and
```
mbudiu-vmw, line 29: Will this compile with any kernel version?

backends/ebpf/CMakeLists.txt
```diff
@@ -141,6 +141,15 @@ else()
   set (SUPPORTS_KERNEL False)
 endif()
 
+if (EXISTS "${CMAKE_CURRENT_SOURCE_DIR}/runtime/usr/include/bpf")
+  message(STATUS "Found libbpf")
+else()
```
mbudiu-vmw, line 146: did you test this on MacOS?

backends/ebpf/runtime/ebpf_kernel.h
```diff
@@ -23,30 +23,48 @@ limitations under the License.
...
+#define bpf_printk(fmt, ...)                                            \
                 ({                                                      \
                         char ____fmt[] = fmt;                           \
                         bpf_trace_printk(____fmt, sizeof(____fmt),      \
                                      ##__VA_ARGS__);                    \
                 })
 
 /* helper functions called from eBPF programs written in C */
-static void *(*bpf_map_lookup_elem)(void *map, void *key) =
+static void *(*bpf_map_lookup_elem)(void *map, const void *key) =
```
mbudiu-vmw, line 54: I was hoping we would not need this junk anymore.
Isn't there a header with this stuff in libbpf?

backends/ebpf/runtime/runtime.mk
```diff
@@ -25,12 +27,12 @@ BUILDDIR:= $(BPFDIR)build
 override INCLUDES+= -I./$(SRCDIR) -include ebpf_runtime_$(TARGET).h
 # Optimization flags to save space
 override CFLAGS+=-O2 -g # -Wall -Werror
-LIBS+=-lpcap
+LIBS+=-lpcap -lelf $(ROOT_DIR)usr/lib64/libbpf.a -lelf -lz
```
mbudiu-vmw, line 30: is elf needed twice?

backends/ebpf/CMakeLists.txt
mbudiu-vmw: Let's see what Travis will say.

backends/ebpf/runtime/ebpf_kernel.h
fruffy: Well this is what libbpf says: "No ties to any specific kernel, transparent handling of older kernels. Libbpf is designed to be kernel-agnostic and work across multitude of kernel versions. It has built-in mechanisms to gracefully handle older kernels, that are missing some of the features, by working around or gracefully degrading functionality. Thus libbpf is not tied to a specific kernel version and can/should be packaged and versioned independently." 

I try to use their custom headers everywhere I can.

backends/ebpf/runtime/ebpf_kernel.h
fruffy: I will see if I can strip more later. Last time I checked those were still necessary.

backends/ebpf/build_libbpf
```diff
@@ -0,0 +1,64 @@
+#!/usr/bin/env python
+
+#!/usr/bin/env python2
```
jafingerhut, line 3: Is it intentional that both /usr/bin/env lines are in this script?

jafingerhut: And should this also perhaps explicitly call out Python 3, to avoid introducing new Python 2 scripts in 2020?

mbudiu-vmw:
There are some problems with your Python scripts.

fruffy:
Trying to support kernel eBPF tests from source is just pure masochism. At this point we also need to build the iproute2 package from source... These are the same problems we had with the xdp back ends so at least we have the fixes already.

fruffy:
Trying to fix the build unearthed a bunch of other issues. Right now, I have no idea what to do about the failing tests. One of them requires a change to how tables are registered, the other is the LPM_TRIE map getting rejected. They are failing for reasons we have missed before.

mbudiu-vmw:
I thought we didn't support lpm in the ebpf backends.

fruffy:
There is a test that fails for the kernel
https://github.com/p4lang/p4c/blob/master/testdata/p4_16_samples/lpm_ebpf.p4
We can mark it xfail for now.

fruffy:
The other tests that fail are the extern object ones, they do not play nicely for various reasons.

fruffy:
I think this is ready for review now. I also created some issues based on problems I have encountered.

mbudiu-vmw:
@fruffy : is this finished?

fruffy:
From my point of view it is. The remaining issues should probably be handled separately. Merging this will break the xdp back end, but I have a pull request ready for it.

PR #2415, opened 2020-06-04
Add serializable enum support for p4runtime generation of parser value

amreshk:
sets

Currently we support only `Type_Bit` and `structs` for parser value sets. This PR extends support for `serializabe enums`.

### DIFF
--- a/control-plane/p4RuntimeSerializer.cpp
+++ b/control-plane/p4RuntimeSerializer.cpp
@@ -1234,16 +1234,36 @@ class P4RuntimeAnalyzer {
                     [](cstring name) { return name == IR::Annotation::matchAnnotation; });
                 addDocumentation(match, f);
             }
+        } else if (et->is<IR::Type_SerEnum>()) {
+            auto serEnum = et->to<IR::Type_SerEnum>();
+            auto fType = serEnum->type;
+            if (!fType->is<IR::Type_Bits>()) {
+                ::error(ErrorType::ERR_UNSUPPORTED,
+                        "Unsupported type argument for Value Set; "
+                        "this version of P4Runtime requires that when the type parameter "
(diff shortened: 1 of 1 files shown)


mbudiu-vmw:
Looks fine from the compiler pov, but someone who knows about the P4Runtime standard should also take a look.

jafingerhut:
@antoninbas If it is not part of those specs today, perhaps it is an extension of the existing specs that some might wish to have added, given the interest in implementing this change?

antoninbas:
> @antoninbas If it is not part of those specs today, perhaps it is an extension of the existing specs that some might wish to have added, given the interest in implementing this change?

yes my point was actually that we should open issues against both specs if we merge this change

amreshk:
@jafingerhut @antoninbas thanks for the inputs. Looks like we are in general agreement on supporting serializable enums for value sets.

I will merge this in for now and we can open issues to update the specs.

hesingh:
> I think this change is fine but technically I would say it violates both the P4_16 spec and the P4Runtime spec.
> 
> If I quote the P4_16 spec (https://p4.org/p4-spec/docs/P4-16-v1.2.0.html#sec-select):
> 
> > Some targets may support parser value set, see Section 12.11. Given a type T for the type parameter of the value set, the type of the value set is set. The type of the value set must match to the type of all other keysetExpression in the same select expression. If there is a mismatch, the compiler must raise an error. The type of the values in the set must be one of bit<>, tuple, and struct.
> 
> The spec doesn't say anything about serializable enums

Why does the section not allow `int<>` because the p4c checks for `Type_Bits` which allows `int<>`?

amreshk:
@hesingh , we can continue the discussion here - https://github.com/p4lang/p4-spec/issues/868

PR #2417, opened 2020-06-05
Unset bitwidth for string controller types in P4Info

antoninbas:
When a user-defined type is translated to a string controller type, the
bitwidth fields for match fields, action params and controller metadata
fields with this type must be unset in the P4Info message.

See https://github.com/p4lang/p4runtime/pull/294


antoninbas:
@smolkaj @stefanheule

Issue #2418, opened 2020-06-05
Preserve source locations of annotations in p4info


smolkaj:
P4Runtime recently added support for preserving the source location of annotations in p4info.proto (https://github.com/p4lang/p4runtime/pull/286). The rationale for this is explained in https://github.com/p4lang/p4c/issues/2333 and https://github.com/p4lang/p4runtime/issues/285.

We need to update p4c to populate the new source location fields in p4info.

smolkaj:
I will prepare a PR for this.

PR #2419, opened 2020-06-05
Safe source location of annotations in p4info protobuf.

smolkaj:
Addresses #2418.
@antoninbas

### DIFF
--- a/control-plane/p4RuntimeArchHandler.cpp
+++ b/control-plane/p4RuntimeArchHandler.cpp
@@ -115,6 +115,15 @@ int64_t getTableSize(const IR::P4Table* table) {
     return tableSize == 0 ? defaultTableSize : tableSize;
 }
 
+p4configv1::SourceLocation
+serializeSourceLocation(const Util::SourceInfo& source_info) {
+    p4configv1::SourceLocation src_loc;
+    src_loc.set_file(source_info.filename.c_str());
+    src_loc.set_line(std::max(0, source_info.line));
+    src_loc.set_column(std::max(0, source_info.column));
+    return src_loc;
+}
+
 std::string serializeOneAnnotation(const IR::Annotation* annotation) {
     // we do not need custom serialization logic here: the P4Info should include
     // the annotation as it was in P4.
@@ -159,6 +168,8 @@ void serializeOneStructuredAnnotation(
     const IR::Annotation* annotation,
     p4configv1::StructuredAnnotation* structuredAnnotation) {
     structuredAnnotation->set_name(annotation->name.name);
+    *structuredAnnotation->mutable_source_location() =
+        serializeSourceLocation(annotation->getSourceInfo());
     switch (annotation->annotationKind()) {
         case IR::Annotation::Kind::StructuredEmpty:
             // nothing to do, body oneof should be empty.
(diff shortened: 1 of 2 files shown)


smolkaj:
Is there a way to automatically update the expected output of the golden tests? @mbudiu-vmw 

It also looks like there is still some work in the frontend needed, currently the source location is not populated properly.

antoninbas:
`P4TEST_REPLACE=1 make check-p4` will re-generate outputs

smolkaj:
Awesome, thanks @antoninbas.
I will look into fixing the frontend issue next week. Have a good weekend, everyone!

smolkaj:
P4C now correctly populates p4info with source locations.

Unfortunately, we have a new issue with golden testing now:
```
--- /p4c/testdata/p4_16_samples_outputs/inline-bmv2.p4.p4info.txt	2020-06-12 23:42:48.000000000 +0000
+++ ./tmpgd00hnu1/inline-bmv2.p4.p4info.txt	2020-06-12 23:50:13.879618783 +0000
@@ -25,7 +25,7 @@
     alias: "NoAction"
     annotations: "@noWarn(\"unused\")"
     annotation_locations {
-      file: "/usr/local/google/home/smolkaj/src/p4c/build/p4include/core.p4"
+      file: "/p4c/build/p4include/core.p4"
       line: 90
     }
   }
```
Since p4info files now includes source file paths, and these file paths naturally differ from machine to machine, we will see a diff between p4info files on different machines in general.

Would it be acceptable to assume that p4c must be located in the folder `/p4c` for the purposes of testing? Any other ideas how we might address this? @mbudiu-vmw @antoninbas

antoninbas:
@smolkaj I suggest adding some kind of switch to turn source location generation on / off. For the golden tests, we can turn it off, and validate the P4Info generation code using gtests only.

Not sure what the best switch mechanism may be here. If we are just targeting tests, an environment variable works fine.

But I can imagine that some users may want to turn this off in production in some cases, which means a flag may be a good idea. We can have a larger discussion about this, but maybe the flag does not need to be super specific. We could have a "debug" mode that would turn source location generation on and reuse that flag for other features in the future (e.g. to include additional information in `PkgInfo`). I believe there is some value in having the ability to generate a P4Info message which is "immutable" and does not differ based on the location of the original source P4 program.

smolkaj:
@antoninbas I think that makes a lot of sense. My preference would be for the locations to be included by default so that error reporting in downstream tools will work "out of the box".
Maybe this would be discussing in a bit more detail at the next p4runtime meeting.

antoninbas:
Enabling source locations by default makes sense to me. Kind of like you have to explicitly define `NDEBUG` in C to turn off some debugging features.

We will discuss this at the next meeting, but if you want to make progress on this PR now, you can try adding a `--p4runtime-mode` flag with 2 possible values, `debug` and `release`, and default it to `debug` which will include source information.

mbudiu-vmw:
We already have the diff problem for the p4 sources that we generate and for the error messages. Our testing scripts try to hide path differences; for example, in run-p4-sample.py run_timeout() normalizes things that look like pathnames that come out of the compiler stdout. A bit ugly. Normally we would write a "smart" diff.

smolkaj:
Okay, agreed, that seems nice.

smolkaj:
@mbudiu-vmw thanks for the pointer. Looks like in this case we can do better with @antoninbas's suggestion but this is good to know for future reference.

jafingerhut:
@smolkaj Is there any interest in pursuing this enhancement?

smolkaj:
Unfortunately, I am unlikely to ever finish this up at this time.
Someone from our team might pursue it in the future if the issue ever becomes pressing.

PR #2420, opened 2020-06-06
Supporting scalar type in PSA

yunheL:
Adding scalar type support in PSA:

When source code contains a more complex expression, for example:
`T a = register.read(x) + register.read(y);` (assume T has a width of 48 bits.)

p4c will transform the source code into:
```c
bit<48> tmp_0;
bit<48> tmp_1; 
tmp_0 = register.read(x);
tmp_1 = register.read(y);
a = tmp_0 + tmp_1;
```

In v1model, `bit<x> tmp;` is supported by:  
(1) create a `scalars` object in the headers section of the generated json file.    
(2) create a `scalars_t` object in the header_types section of the generated json file.   
(3) For each `bit<x> varName;` add a field to header_type `scalars_t`.  

This is not implemented for PSA yet and more complex expressions like the one mentioned above fails to run on bmv2. This PR implements the three points listed above for PSA to support `bit<x> varName`, thus more complex expressions like one mentioned above can run on bmv2.

### DIFF
--- a/backends/bmv2/psa_switch/psaSwitch.cpp
+++ b/backends/bmv2/psa_switch/psaSwitch.cpp
@@ -22,6 +22,7 @@ namespace BMV2 {
 void PsaProgramStructure::create(ConversionContext* ctxt) {
     createTypes(ctxt);
     createHeaders(ctxt);
+    createScalars(ctxt);
     createExterns();
     createParsers(ctxt);
     createActions(ctxt);
@@ -114,6 +115,24 @@ void PsaProgramStructure::createTypes(ConversionContext* ctxt) {
     /* TODO */
     // add errors to json
     // add enums to json
+}
+
+void PsaProgramStructure::createScalars(ConversionContext* ctxt) {
+    auto name = scalars.begin()->first;
+    ctxt->json->add_header("scalars_t", name);
+    ctxt->json->add_header_type("scalars_t");
+
+    for (auto kv : scalars) {
+        LOG5("Adding a scalar field " << kv.second << " to generated json");
+        auto field = new Util::JsonArray();
+        auto ftype = typeMap->getType(kv.second, true);
+        if (auto type = ftype->to<IR::Type_Bits>()) {
+            field->append(kv.second->name);
+            field->append(type->size);
+            field->append(type->isSigned);
+        }
+        ctxt->json->add_header_field("scalars_t", field);
+    }
 }
 
 void PsaProgramStructure::createHeaders(ConversionContext* ctxt) {
@@ -388,6 +407,19 @@ void InspectPsaProgram::addTypesAndInstances(const IR::Type_StructLike* type, bo
             }
         }
     }
+}
+
+
+bool InspectPsaProgram::preorder(const IR::Declaration_Variable* dv) {
+        auto ft = typeMap->getType(dv->getNode(), true);
+        cstring scalarsName = refMap->newName("scalars");
+
+        if (ft->is<IR::Type_Bits>()) {
+            LOG5("Adding " << dv << " into scalars map");
+            pinfo->scalars.emplace(scalarsName, dv);
+        }
+
+        return false;
 }
 
 // This visitor only visits the parameter in the statement from architecture.
--- a/backends/bmv2/psa_switch/psaSwitch.cpp
+++ b/backends/bmv2/psa_switch/psaSwitch.cpp
@@ -130,6 +130,8 @@ void PsaProgramStructure::createScalars(ConversionContext* ctxt) {
             field->append(kv.second->name);
             field->append(type->size);
             field->append(type->isSigned);
+        } else {
+            BUG_CHECK(kv.second, "%1 is not of Type_Bits");
         }
         ctxt->json->add_header_field("scalars_t", field);
     }
--- a/backends/bmv2/CMakeLists.txt
+++ b/backends/bmv2/CMakeLists.txt
@@ -173,7 +173,6 @@ set (XFAIL_TESTS
   testdata/p4_16_samples/issue1205-bmv2.p4
   # These psa tests are not ready to run on bmv2 yet
   testdata/p4_16_samples/psa-example-digest-bmv2.p4
-  testdata/p4_16_samples/psa-example-register2-bmv2.p4
   # This reads stack.next
   testdata/p4_16_samples/issue692-bmv2.p4
   # These test use computations in the verify/update checksum controls - unsupported
--- a/backends/bmv2/psa_switch/psaSwitch.h
+++ b/backends/bmv2/psa_switch/psaSwitch.h
@@ -121,9 +121,8 @@ class PsaProgramStructure : public ProgramStructure {
 
  public:
     // We place scalar user metadata fields (i.e., bit<>, bool)
-    // in the scalarsName metadata object, so we may need to rename
-    // these fields.  This map holds the new names.
-    std::vector<const IR::StructField*> scalars;
+    // in the scalars map.
+    ordered_map<cstring, const IR::Declaration_Variable*> scalars;
     unsigned                            scalars_width = 0;
     unsigned                            error_width = 32;
     unsigned                            bool_width = 1;
@@ -160,6 +159,7 @@ class PsaProgramStructure : public ProgramStructure {
     void createStructLike(ConversionContext* ctxt, const IR::Type_StructLike* st);
     void createTypes(ConversionContext* ctxt);
     void createHeaders(ConversionContext* ctxt);
+    void createScalars(ConversionContext* ctxt);
     void createParsers(ConversionContext* ctxt);
     void createExterns();
     void createActions(ConversionContext* ctxt);
@@ -220,6 +220,7 @@ class InspectPsaProgram : public Inspector {
     void addTypesAndInstances(const IR::Type_StructLike* type, bool meta);
     void addHeaderType(const IR::Type_StructLike *st);
     void addHeaderInstance(const IR::Type_StructLike *st, cstring name);
+    bool preorder(const IR::Declaration_Variable* dv) override;
     bool preorder(const IR::Parameter* parameter) override;
 };
 
@@ -243,7 +244,7 @@ class ConvertPsaToJson : public Inspector {
         CHECK_NULL(structure); }
 
     void postorder(UNUSED const IR::P4Program* program) override {
-        cstring scalarsName = refMap->newName("scalars");
+        cstring scalarsName = "scalars";
         // This visitor is used in multiple passes to convert expression to json
         auto conv = new PsaSwitchExpressionConverter(refMap, typeMap, structure, scalarsName);
         auto ctxt = new ConversionContext(refMap, typeMap, toplevel, structure, conv, json);
--- a/testdata/p4_16_samples/psa-example-register2-bmv2.p4
+++ b/testdata/p4_16_samples/psa-example-register2-bmv2.p4
@@ -121,6 +121,9 @@ control ingress(inout headers hdr,
 
     apply {
         ostd.egress_port = (PortId_t) 0;
+        hdr.ipv4.setValid();
+        // dstAddr (48 bits), srcAddr (48 bits), etherType (16 bits)
+        hdr.ipv4.totalLen = (48 + 48 + 16) / 8;
         if (hdr.ipv4.isValid()) {
             @atomic {
                 PacketByteCountState_t tmp;
--- a/testdata/p4_16_samples/psa-example-register2-bmv2.stf
+++ b/testdata/p4_16_samples/psa-example-register2-bmv2.stf
@@ -1,1 +1,13 @@
-# empty stf for testing
+# cmd | port | dest | src | ethertype
+packet 4 000000000001 000000000002 ffff
+
+register_read ingress.port_pkt_ip_bytes_in 4
+# expect 281474976710670
+
+# Here is the reason:
+# PACKET_BYTE_COUNT_WIDTH = 80
+# PACKET_COUNT_WIDTH = 32
+# BYTE_COUNT_WIDTH = 48
+# PACKET_BYTE [79:0] = PACKET[79:48] | BYTE[47:0]
+# The test case generates 1 packet which has 14 bytes
+# Thus we expect (1 << 48) + 14 = 281474976710670
--- a/testdata/p4_16_samples/psa-example-register2-bmv2.stf
+++ b/testdata/p4_16_samples/psa-example-register2-bmv2.stf
@@ -1,0 +1,1 @@
+# empty stf for testing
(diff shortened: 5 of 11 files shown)


backends/bmv2/psa_switch/psaSwitch.cpp
```diff
@@ -116,6 +117,24 @@ void PsaProgramStructure::createTypes(ConversionContext* ctxt) {
...
+void PsaProgramStructure::createScalars(ConversionContext* ctxt) {
+    auto name = scalars.begin()->first;
+    ctxt->json->add_header("scalars_t", name);
+    ctxt->json->add_header_type("scalars_t");
+
+    for (auto kv : scalars) {
+        LOG5("Adding a scalar field " << kv.second << " to generated json");
+        auto field = new Util::JsonArray();
+        auto ftype = typeMap->getType(kv.second, true);
+        if (auto type = ftype->to<IR::Type_Bits>()) {
```
mbudiu-vmw, line 129: What if it's not?

yunheL: I use this function to handle all statements that is in the format `bit<x> varName;`, thus I believe all targeted statements will go in the if block.

If a statement does not go into the if block, the statement is not of format `bit<x> varName;` and should not be caught by this function.

mbudiu-vmw:
Shouldn't that be an error?

yunheL:
> Shouldn't that be an error?

Ah, that's right. I will add an else block in which I will do error handling. In this case, it should report a compiler bug but not a user bug right? Thanks!

yunheL:
> > Shouldn't that be an error?
> 
> Ah, that's right. I will add an else block in which I will do error handling. In this case, it should report a compiler bug but not a user bug right? Thanks!

Updated!

jafingerhut:
I am not claiming to understand the compiler changes being made here.  I simply wanted to point out that there are pretty basic bugs in the current PSA-specific compiler implementation that seem to me to have nothing to do with any of the externs at all.

For example, the test programs attached to the issue below still fails, in a similar way as it did in May 2019, and uses no PSA externs at all.  I do not know the precise problem, or what changes to p4c would be required to make them work, but since these test program does not use externs, I am pretty sure that the p4c issues involved with that test program have nothing to do with externs, but something more basic in the p4c PSA implementation.

+ https://github.com/p4lang/p4c/issues/1919

yunheL:
> I am not claiming to understand the compiler changes being made here. I simply wanted to point out that there are pretty basic bugs in the current PSA-specific compiler implementation that seem to me to have nothing to do with any of the externs at all.
> 
> For example, the test program attached to the issue below still fails, in a similar way as it did in May 2019, and uses no PSA externs at all. I do not know the precise problem, or what changes to p4c would be required to make them work, but since this test program does not use externs, I am pretty sure that the p4c issues involved have nothing to do with externs, but something more basic in the p4c PSA implementation.
> 
> * #1919

Hi Andy, thank you for the information and linking to issue #1919 . Looking at the  error message in issue 1919 (`Invalid reference to object of type 'header' with name 'scalars'`), I believe it is the problem I am trying to solve with this PR.  

I believe the problem is:
With v1model program, the json generated by v1model compiler (p4c-bm2-ss) has three categories of headers: 
(1) scalars: handles non-struct like declarations, for example declarations of type `bit<x>` and `bool`.
(2) standard_metadata: ingress_port, egress_port etc.
(3) user defined headers.  

Here is what happens with the these categories in PSA: 
(3) user defined headers is supported the same way as v1model. 
(2) standard_metadata has been decomposed into a several headers including `psa_ingress_parser_input_metadata` and `psa_ingress_output_metadata` etc.
(1) **scalars, however, is not included as a header object** anymore in PSA compiler (p4c-bm2-psa) generated json file. If a statement like `bit<48> tmp_0` exists in the source code after all compiler midend passes, p4c-bm2-psa will generate this in the .json when `tmp_0` is being used:
```json
           "parameters" : [
             {
               "type" : "field",
               "value" : ["scalars", "tmp_0"]
             }
```
Thus, with `tmp_0`, bmv2 will try to find a header called scalars. but since there is no scalars header in the json file's header list, bmv2 will give the error: `Invalid reference to object of type 'header' with name 'scalars'`.  

With this PR, I added scalars into the headers list and scalars_t into the header_types list of the generated json to solve this problem. However, I have seen 1 test is failing with travis build (`psa-example-register2-bmv2.p4`) and is currently investigating.   

Thank you very much for bringing up issue 1919. I think it is very relevant and will provide insights in my investigation.

yunheL:
Currently, there is 1 test failing with Travis build which gives an error message:

```java
1176/1573 Test #415: bmv2/testdata/p4_16_samples/psa-example-register2-bmv2.p4 .....................***Failed    5.17 sec
Obtaining JSON from switch...
Done
RuntimeCmd: 
Check for  /p4c/testdata/p4_16_samples/psa-example-register2-bmv2.stf
``` 

Looks like travis build is looking for an stf for this program which did not exists. Thus I added an empty stf test (psa-example-register2-bmv2.stf). But travis is still failing and giving the same error message. 

I am not sure what the problem is, any suggestions is much appreciated!

hesingh:
There are several tests in `p4c/testadata/p4_16_samples` for bmv2 without use of any stf file.  It's fine.  Further, note, the test your are testing is quarantined [here](https://github.com/p4lang/p4c/blob/master/backends/bmv2/CMakeLists.txt#L176)

I think Travis is broken trying to run this test.

jafingerhut:
The test psa-example-register2-bmv2.p4 is in an 'XFAIL' set of tests, in the file backends/bmv2/CMakeLists.txt

NOTE: When a test is marked XFAIL, it does NOT mean "do not run this test".

Instead, it means, run this test normally, and check for succeed or fail result.  If the result is succeed, invert that and fail the entire larger test.  If the result is fail, invert that and keep doing the larger test.  It is used for checking in tests that should some day in the future succeed, but right now the implementation causes them to fail.

Most likely what is happening is that your changes improve the compiler enough that psa-example-register2-bmv2.p4 now passes.  Because it is marked XFAIL, that causes the entire test suite to fail.  The remedy is straightforward: remove it from the XFAIL list, because now we expect it to pass.

yunheL:
> There are several tests in `p4c/testadata/p4_16_samples` for bmv2 without use of any stf file. It's fine. Further, note, the test your are testing is quarantined [here](https://github.com/p4lang/p4c/blob/master/backends/bmv2/CMakeLists.txt#L176)
> 
> I think Travis is broken trying to run this test.

Thank you Hemant, that is exactly what I a looking for!

yunheL:
> The test psa-example-register2-bmv2.p4 is in an 'XFAIL' set of tests, in the file backends/bmv2/CMakeLists.txt
> 
> NOTE: When a test is marked XFAIL, it does NOT mean "do not run this test".
> 
> Instead, it means, run this test normally, and check for succeed or fail result. If the result is succeed, invert that and fail the entire larger test. If the result is fail, invert that and keep doing the larger test. It is used for checking in tests that should some day in the future succeed, but right now the implementation causes them to fail.
> 
> Most likely what is happening is that your changes improve the compiler enough that psa-example-register2-bmv2.p4 now passes. Because it is marked XFAIL, that causes the entire test suite to fail. The remedy is straightforward: remove it from the XFAIL list, because now we expect it to pass.

Thank you Andy, that makes a lot of sense!

hesingh:
>

@yunheL Please take a look at the json format in 

https://github.com/p4lang/behavioral-model/blob/master/tests/testdata/pvs_struct_2.json

and search in the file for "scalars".  This is provide you with more clarity for what to generate for JSON .

@antoninbas If scalars are supports by the v1model, how come the simple_switch [documentation ](https://github.com/p4lang/behavioral-model/blob/master/docs/JSON_format.md)does not mention any scalars format?

jafingerhut:
@hesingh asked: "If scalars are supported by the v1model, how come the simple_switch documentation does not mention any scalars format?"

Antonin can answer more authoritatively, but most likely there is nothing special in the BMv2 JSON format for the header name "scalars".  That is just a name that wasn't used by any header names by the P4 developer's P4 program, so someone decided to write the p4c simple_switch back end to create such a 'header' in the BMv2 JSON file to hold all of the things that doesn't have a header defined for it, because BMv2 JSON doesn't let you have fields that aren't in headers/structs.

You could change that name to "blibbidy_bloop" everywhere and the BMv2 JSON would be handled by simple_switch just as well as if it is "scalars".

hesingh:
I have been scratching my head since past few weeks when Vladimir mentioned "scalar" in a serEnum typdef issue.  I checked the p4-16 working doc and didn't see any mention of any scalar.  Finally, looking at this PR I  looked at p4c code and see the term scalar being used in p4c frontend and midend code but not as any IR - this inspection made me clear what Valdimir was talking about.   Then I saw the use of "scalars" in the bmv2 backend JSON output.   If we decide to use such a term in JSON, I am humbly saying, we should document it because when fixing issues in PSA the "scalars" name is null and one jumps through hoops trying to figure out what the name should be.

jafingerhut:
Additional hints that the name "scalars" is simply a design choice for a name by whoever wrote the p4c simple_switch back end:

+ https://github.com/p4lang/p4c/blob/master/backends/bmv2/simple_switch/simpleSwitch.cpp#L1103
+ I haven't tried it, but I bet if you change the string `"scalars"` to `"blibbidy_bloop"`, probably everything still works, in that C++ source file, probably everything still works.  Similarly if you hand-edit a BMv2 JSON file occurrence of `"scalars"` everywhere it occurs.

jafingerhut:
The term 'scalars' is commonly used for many programming languages, not just P4, to refer to a 'single value' as opposed to a 'compound value'.  That use of the term likely originates in mathematics, where the contrast is a math vector, a compound of many scalars, versus an individual real number, a scalar.  https://en.wikipedia.org/wiki/Variable_(computer_science)

yunheL:
> Additional hints that the name "scalars" is simply a design choice for a name by whoever wrote the p4c simple_switch back end:
> 
> * https://github.com/p4lang/p4c/blob/master/backends/bmv2/simple_switch/simpleSwitch.cpp#L1103
> * I haven't tried it, but I bet if you change the string `"scalars"` to `"blibbidy_bloop"`, probably everything still works, in that C++ source file, probably everything still works.  Similarly if you hand-edit a BMv2 JSON file occurrence of `"scalars"` everywhere it occurs.

I tried, I changed the string name from `"scalars"` to `"breakfast"` and it still works :)

hesingh:
> The term 'scalars' is commonly used for many programming languages, not just P4, to refer to a 'single value' as opposed to a 'compound value'. That use of the term likely originates in mathematics, where the contrast is a math vector, a compound of many scalars, versus an individual real number, a scalar. https://en.wikipedia.org/wiki/Variable_(computer_science)

Understood.  It was just that first I look for any term mentioned for P4 in the p4-16 doc or in p4c IR.  Looking at p4c source code i figured out what scalar was being referred to.  thanks.

hesingh:
@jafingerhut If it were one entity, I understand, use any name.  But there are two.  One is header_type and another is header, the instance.  For v1model, header_type is `scalars_0` and the instance is `scalars`. 

```
"header_types" : [
    {
      "name" : "scalars_0",
      "id" : 0,
      "fields" : [
        ["tmp", 16, false],
        ["tmp_0", 3, false],
        ["key_0", 32, false],
        ["_padding_0", 5, false]
      ]
    },

"headers" : [
    {
      "name" : "scalars",
      "id" : 0,
      "header_type" : "scalars_0",
      "metadata" : true,
      "pi_omit" : true
    },
```

PSA is using `scalars_t` for header_type and `scalars` for instance.   So, now one has to change the simple_switch for PSA scalars JSON.

hanw:
@hesingh the psa backend targets psa_switch, there is no need to update simple_switch. @yunheL will take care of the change if name conflicts becomes an issue.

hesingh:
> @hesingh the psa backend targets psa_switch, there is no need to update simple_switch. @yunheL will take care of the change if name conflicts becomes an issue.

Duh, sorry, I forgot about psa_switch.  Both switches share some code and this is why one has to test psa_switch and see if any changes are needed for scalars, thanks.

jafingerhut:
@hesingh As far as I know, there is nothing in simple_switch, and nothing in psa_switch, that is looking specially for any of these names with 'scalars' in them, neither in the "header_types" section, nor the "headers" section.  As I mentioned earlier, I am pretty sure you could globally search those names in a BMv2 JSON file for either simple_switch, or psa_switch, replace them with some other name not in use in that file, and it would work the same as the original.

Are you thinking that the psa_switch and simple_switch back ends ought to be using the same name in header_types and/or headers sections, for some reason that it is incorrect if they differ?  I don't think it is incorrect, just .... different by choice of people writing the back ends.

hesingh:
@jafingerhut It's fine for simple_switch and psa_switch to use different header_type and header.

castilhorosa:
> > Additional hints that the name "scalars" is simply a design choice for a name by whoever wrote the p4c simple_switch back end:
> > 
> > * https://github.com/p4lang/p4c/blob/master/backends/bmv2/simple_switch/simpleSwitch.cpp#L1103
> > * I haven't tried it, but I bet if you change the string `"scalars"` to `"blibbidy_bloop"`, probably everything still works, in that C++ source file, probably everything still works.  Similarly if you hand-edit a BMv2 JSON file occurrence of `"scalars"` everywhere it occurs.
> 
> I tried, I changed the string name from `"scalars"` to `"breakfast"` and it still works :)



> > Additional hints that the name "scalars" is simply a design choice for a name by whoever wrote the p4c simple_switch back end:
> > 
> > * https://github.com/p4lang/p4c/blob/master/backends/bmv2/simple_switch/simpleSwitch.cpp#L1103
> > * I haven't tried it, but I bet if you change the string `"scalars"` to `"blibbidy_bloop"`, probably everything still works, in that C++ source file, probably everything still works.  Similarly if you hand-edit a BMv2 JSON file occurrence of `"scalars"` everywhere it occurs.
> 
> I tried, I changed the string name from `"scalars"` to `"breakfast"` and it still works :)

Is there any update on this issue? I'm getting the error (`Invalid reference to object of type 'header' with name 'scalars'`) in my P4/PSA program. I've tried to change the string name from `"scalars"` to something else in the JSON file like you and @jafingerhut said but it didn't work. When I change the string name from  `"scalars"` to `"blibbidy_bloop"` in JSON file, for instance, I get the same error: (`Invalid reference to object of type 'header' with name 'blibbidy_bloop'`). Other than changing the string name in JSON, do I need to change it in https://github.com/p4lang/p4c/blob/master/backends/bmv2/simple_switch/simpleSwitch.cpp#L1143 and compile? It seems that #1919 issue has not been solved yet. If it is the case, is there any temporary solution to get my code executed?

mbudiu-vmw:
@castilhorosa I am not sure what you are trying to achieve. This pull request has been merged, so it was supposed to solve this problem.

Issue #2421, opened 2020-06-06
Broken STF test


jnfoster:
It looks to me like these two lines in the `runtime-index-bmv2` STF test are wrong.

https://github.com/p4lang/p4c/blob/master/testdata/p4_16_samples/runtime-index-bmv2.stf#L46
https://github.com/p4lang/p4c/blob/master/testdata/p4_16_samples/runtime-index-bmv2.stf#L78

hesingh:
@jnfoster Yes, this was temporary code.  The P4 program has to change to use registers and then STF test cases will be updated.  I nor others at the IETF hackathon did not get the time to work on these issues yet.

jafingerhut:
Do you mean that those lines should be comments, but have no '#' as a prefix?  If so, I agree, and the fix is simply to prefix those lines with '#' characters.

mbudiu-vmw:
Finally, a bug which is easy to fix.

PR #2423, opened 2020-06-08
Fix stf file with missing comments

mbudiu-vmw:


mbudiu-vmw:
Fixes #2421

mbudiu-vmw:
@jnfoster : how about a little review for #2404? it just moves one line lower in the code...

PR #2424, opened 2020-06-08
Generalize SerEnum

mbudiu-vmw:
This PR is built on top of #2394.

### DIFF
--- a/backends/ebpf/targets/xdp_target.py
+++ b/backends/ebpf/targets/xdp_target.py
@@ -1,1 +1,0 @@
-/home/mbudiu/git/p4c/extensions/p4c-xdp/xdp_target.py
--- a/backends/ebpf/targets/xdp_target.py
+++ b/backends/ebpf/targets/xdp_target.py
@@ -1,0 +1,1 @@
+/home/mbudiu/git/p4c/extensions/p4c-xdp/xdp_target.py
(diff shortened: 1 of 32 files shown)


hesingh:
Would all the changes made to typeChecker.cpp be necessary if we merely added a preorder(IR::SerEnum) method which would use getTypeType()?

mbudiu-vmw:
You can go ahead and try it. In preorder the type has not yet been computed. The types are computed bottom-up, so they are only available in postorder.

backends/ebpf/targets/xdp_target.py
```diff
@@ -0,0 +1 @@
+/home/mbudiu/git/p4c/extensions/p4c-xdp/xdp_target.py
```
jnfoster, line 1: This seems unrelated to the PR and also references a file in @mbudiu-vmw's machine...

mbudiu-vmw:
@jnfoster : I think your change request has been satisfied.

jnfoster:
Whoops. Sorry for letting this sit for 2 days. Thanks for the nudge.

jafingerhut:
I do not know yet whether this will impact the P4Runtime API, but have created this issue to track the request for the P4 API working group to think about it: https://github.com/p4lang/p4runtime/issues/300

smolkaj:
This PR got merged with a symlink to @mbudiu-vmw's machine, see @jnfoster's comment above.

PR #2425, opened 2020-06-09
convert psa meter color to serializable enum with width 8

hanw:
open to suggestion. bit<8> is what we use currently.

### DIFF
--- a/p4include/psa.p4
+++ b/p4include/psa.p4
@@ -646,7 +646,7 @@ enum PSA_MeterType_t {
 // END:MeterType_defn
 
 // BEGIN:MeterColor_defn
-enum PSA_MeterColor_t { RED, GREEN, YELLOW }
+enum bit<8> PSA_MeterColor_t { GREEN = 8w0, YELLOW = 8w1, RED = 8w3 }
 // END:MeterColor_defn
 
 // BEGIN:Meter_extern
--- a/testdata/p4_16_errors_outputs/psa-meter2.p4-stderr
+++ b/testdata/p4_16_errors_outputs/psa-meter2.p4-stderr
@@ -1,3 +1,3 @@
-psa-meter2.p4(53): [--Werror=type-error] error: cast: Cannot unify PSA_MeterColor_t to bit<16>
+psa-meter2.p4(53): [--Werror=type-error] error: cast: Cannot unify bit<16> to bit<8>
         b.data = (bit<16>)meter0.execute(index, PSA_MeterColor_t.GREEN);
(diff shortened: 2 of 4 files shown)


mbudiu-vmw:
Some tests failef

hanw:
@jafingerhut FYI, I'm not sure what other options we have here. We need meter color to be serializable enum to enable P4Runtime API generation for action that uses meter color as action parameter. However, not sure who to talk to achieve a consensus on the width and encoding. So I just put what we have out here.

jafingerhut:
@hanw Thanks for checking.  I will send out a time for regular every-2-week series of P4 Arch working group meetings to address desired updates for a new version of the PSA spec, which I suspect should be numbered version 2.x because some of those changes will not be backwards compatible, and we should hash out the details there.

For this particular proposed change, bit<8> seems reasonable.  bit<2> is wide enough, but understood if that is probably not an ideal size for other implementation-specific reasons.

The particular numeric values I do not care so much about, but it might be nice to have a sentence of motivation why they aren't simply a consecutive range of numeric values 0, ...., (N-1).  Again, I suspect implementation-specific motivations here are the underlying reasons.

hanw (commented):
@jafingerhut @mbudiu-vmw Can I get an approval for this?

We did not get a chance to talk about this issue in today's PSA meeting, we can talk about it next week. In the interest of making progress, I would propose to merge it this week to unblock Yunhe and other folks on this issue. I will make another PR against PSA spec for this change.

jafingerhut:
@hanw and I have discussed this a bit off line, and I believe we might have reached a consensus that any approach that tries to standardize the numeric encodings of a meter's 3-color green/yellow/red markings seems to go against portability across targets.  I would not even be surprised if the numeric values used in the data plane of different products within a _single company's_ portfolio of products might not all be identical to each other in this numeric encoding, especially given acquisitions of one networking company by another.

My personal belief is that the right answer is to encourage the P4Runtime API developers to add support in the control plane APIs for non-serializable enums as action parameters and table key fields, which is the plan for version 2.x, I believe, and the only reason it is not in version 1.0 was an explicit design decision to get that version published sooner, versus any extra time it might have taken to make 1.0 more general in this regard.  I have added this motivating use case as an additional known reason to generalize this in version 2.x here: https://github.com/p4lang/p4runtime/issues/191

The only programs that I am aware of that this causes problems for are ones where the developer wishes to make a value of type MeterColor_t an action parameter.  The most likely case where someone would want to do so would be one where they want to implement color-aware policing, and convert the IPv4 ToS/DSCP, or IPv6 Traffic Class, or MPLS EXP field, into one of the 3 colors.  This can be worked around for the moment by making 3 separate actions, one per color.

PR #2426, opened 2020-06-09
Packet truncation support for uBPF backend

tatry:
I've implemented support for `truncate()` extern in the uBPF backend. This allows cut off some last bytes from a packet, but dataplane must support that.

### DIFF
--- a/backends/ubpf/p4include/ubpf_model.p4
+++ b/backends/ubpf/p4include/ubpf_model.p4
@@ -99,8 +99,15 @@ extern Register<T, S> {
  */
 extern bit<48> ubpf_time_get_ns();
 
-/*
- * Truncate packet
+/***
+ * Truncate packet to the maximum size
+ *
+ * @param len   Maximum length of the packet (from beginning of packet) in
+ *              bytes, further bytes will be removed. If deparsed packet is
+ *              shorter than len, this extern has no effect.
+ *              Minimum value for len is 14, as a packet in Open vSwitch must
+ *              have at least Ethernet header. Smaller values than 14 will be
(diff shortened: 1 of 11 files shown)


backends/ubpf/p4include/ubpf_model.p4
```diff
@@ -99,6 +99,10 @@ extern Register<T, S> {
  */
 extern bit<48> ubpf_time_get_ns();
 
+/*
+ * Truncate packet
+ */
+extern void truncate(in bit<32> len);
```
jafingerhut, line 105: It would be good to have a sentence or three describing at least the units of the parameter `len` (bits?  bytes?), whether it starts from the first byte of an Ethernet header, if the packet is Ethernet, or some other location, and whether the intent is to support any integer number value of `len`, or only some subset of those values are supported, e.g. only multiples of 16 bytes.

backends/ubpf/p4include/ubpf_model.p4
```diff
@@ -99,8 +99,15 @@ extern Register<T, S> {
...
 
-/*
- * Truncate packet
+/***
+ * Truncate packet to the maximum size
+ *
+ * @param len   Maximum length of the packet (from beginning of packet) in
+ *              bytes, further bytes will be removed. If deparsed packet is
+ *              shorter than len, this extern has no effect.
+ *              Minimum value for len is 14, as a packet in Open vSwitch must
```
osinstom, line 108: This may be removed as uBPF could be used for different software switches.

osinstom:
LGTM, I supervise this work and we've tested this feature while implementing INT with uBPF backend.

PR #2427, opened 2020-06-09
Implementation of lookahead<bit<w>>() for parser in uBPF backend

tatry:
I've implemented `lookahead<>()` in the uBPF parser with support only for `bit<w>` type where 0 < w < 65. So far other types are not supported. This allows to read a few bytes from a packet without cursor advancing.

Implementation is based on header's field extraction.


jafingerhut:
I would recommend examining whether you can make your implementation of lookahead such that if it is passed a number of bits that is not supported in your implementation, that the parser immediately goes to the reject state and records the error called `error.ParserInvalidArgument` in the P4_16 language spec.  I did not check -- perhaps your proposed changes already do this.

tatry:
@jafingerhut 
Number of bits passed to `lookahead()` is checked during compilation time. Too many bits to read cause call to `BUG()`.
During runtime parser checks whether there are sufficient bits to read from a packet - if not, goes to reject state. Error codes are not supported yet in the uBPF backend.

osinstom:
LGTM too, I supervise this work and we've tested this feature while implementing INT with uBPF backend. Regarding support for error codes in uBPF backend, this is still on the roadmap, but it doesn't have a high priority.

PR #2428, opened 2020-06-09
Update PSA programs

amreshk:


jafingerhut:
Any motivation for these changes?  It isn't clear from the PR itself.

hanw:
This is an instance of the problem that a program that can be expressed in PSA cannot be supported due to resource constraints on our end. We would like to change the program slightly so that it is both a valid psa program and can be compiled to our target.

jafingerhut:
Makes sense, and seems fine to me to make such changes.  Looks like some expected output files need to be updated, too, before the tests will pass.

PR #2429, opened 2020-06-09
Update PSA programs - 2

amreshk:
Updates to programs due to resource constraints


antoninbas:
I am not sure what the motivation is for merging these changes in this repository. It's a pretty fair statement that it makes the programs harder to read, and it seems that splitting the equality test for resource constraints reason should be done by a backend pass.

jafingerhut:
I definitely agree that my personal order of preference would be, from highest to lowest:
(1) use 48-bit comparisons in P4 source code, and make the compiler support it, e.g. perhaps by splitting it up into multiple comparisons if needed
(2) use `(a[47:32] == 0) && (a[31:0] == 0)`
(3) use nested `if`s as are in the current PR as I write this.

antoninbas:
Better to fail these tests in your backend or write your own version IMO, if you do not want to implement this support. It is unreasonable to expect that you can constrain all the PSA programs in this repo to those that can compile on a resource-constrained backend.

hesingh:
May I ask which p4c backend has resource constraints for the P4 program being changed?

Issue #2271, opened 2020-03-26
Bazel build files?


stefanheule:
It would be useful to have Bazel build files for p4c, like other p4lang projects already do (e.g. p4runtime). If I go ahead and write them, could we add them to this repository?

mbudiu-vmw:
If you commit to maintain them, certainly.
The build system is quite complicated, though, you should expect this will be a non-trivial effort.
If there is a tool to convert cmake to bazel you may save some time.

cc10512:
@stefanheule how do you plan to test the Bazel builds?

stefanheule:
Just invoking them would be enough, I imagine? Or is there anything else that you would want to test?

jafingerhut:
Using the existing cmake-based build, a good check that things went well (or at least most of it went well), is to do "cd build ; make check".   If a Bazel-based build also enabled something similar, that would be a fairly thorough way to check the build results.

I do not know that people would want to immediately enable the Travis/CI to automatically do Bazel-based builds and use them as additional prerequisites for merging PRs, but I suppose it is a possibility if the extra compute resources are not an issue (they could be done in parallel to avoid extra latency).

cc10512:
Yes, but they need to be invoked any time someone changes any of the cmake files. Since we do not have such smarts, the next best thing is to have a bazel build run as part of Travis. That support is not there yet. I suppose it should not be too hard to add, but it does imply adding another row to the build matrix.

mbudiu-vmw:
Travis seems less and less reliable; doubling the build time in CI is not really a great option. But I agree that one would have to run the bazel builds periodically to make sure they don't break.

hesingh:
I don't consider p4runtime as any yardstick to borrow from for p4c.  p4runtime is essentially worked by one person in Antonin while p4c, in comparison, has an army. 

Second, if one sees this comparison between bazel and cmake, I am underwhelmed by bazel.  

https://tinyurl.com/ybq89q88

stefanheule:
Nobody is suggesting to replace cmake, only to offer bazel in addition to cmake. If you prefer cmake, you can continue to use cmake.

hesingh:
@stefanheule I see, thanks.

So, by default, does Travis run Bazel or cmake to test a PR code changes to p4c?
Which one of Bazel or make (after cmake) runs faster?

What is the biggest advantage to also add Bazel to p4c build?

stefanheule:
[...]

PR #2430, opened 2020-06-11
Add Bazel build files for p4c

EileenFeng:
Adding bazel files for p4c as discussed in [p4lang/p4c#2271](https://github.com/p4lang/p4c/issues/2271).


README.md
```diff
@@ -322,6 +322,21 @@ default is 2GB, which is not enough to build p4c. Increase the memory limit to
 at least 4GB via Docker preferences or you are likely to see "internal compiler
 errors" from gcc which are caused by low memory.
 
+# Bazel
```
mbudiu-vmw, line 325: This is cool.
One nice thing about the makefiles was that you could just add another project and if you respect the format of the makefiles it would build. For example, on my machine I checkout the p4c-xdp repo, and then I create a symlink in `extensions` to the root directory, and it is automatically picked up by cmake, including all tests. I don't really know bazel, but with this structure it doesn't look to me like this would behave in the same way.

smolkaj: With Bazel this would work the opposite way, in the sense that your extension (p4c-xdp) would include p4c as a dependency. There is an example in the bazel/example folder of how to depend on p4c in a Bazel project.

smolkaj:
It seems that GitHub Actions are currently disabled for this repository. @mbudiu-vmw, could you take a look at enabling them? https://help.github.com/en/github/administering-a-repository/disabling-or-limiting-github-actions-for-a-repository

For context, this PR adds a Github Action that tests the Bazel build for every PR/commit to master.

mbudiu-vmw:
I don't think I have permissions to enable actions.

stefanheule:
@jnfoster Nate, do you have permissions to enable them? Or can point us to someone who does?

jnfoster:
The p4lang was on a legacy paid plan that did not support actions. I changed that about a week ago on @antoninbas's request. However, the change won't take effect until 13 June and I don't see a way to override that. So in 2 days we should be able to do this.

smolkaj:
Interestingly they already work for p4lang/p4-constraints and p4lang/p4runtime.

jnfoster:
Weird. I must not be finding the right settings then.

smolkaj:
Weird indeed, because they should be enabled by default in the first place (though this may be explained by the legacy thing).

jnfoster:
Here's what I see for p4c... 

Screen Shot 2020-06-11 at 4 37 55 PM

smolkaj:
Looks right. Could you try disabling and then reenabling? GitHub is owned by Microsoft now so it probably needs a restart ;)

jnfoster:
When I changed the settings, my whole browser froze up and turned blue...

Just kidding. 

I disabled, waited for the green check mark to appear in the right place, and then re-enabled. Try again?

smolkaj:
Never mind, I think I know what's going on. The action is not being run because the PR is based on a branch in @EileenFeng's repository. If we want to run the action before merging, we need to move the PR to a branch on p4lang/p4c. I don't currently have the necessary privileges to do that though.

smolkaj:
Alternatively, reviewers can convince themselves that this works by looking at Eileen's repository: https://github.com/EileenFeng/p4c/actions

BUILD.bazel
```diff
@@ -0,0 +1,379 @@
...
+    name = "p4c_includes",
+    includes = [
+        "ir/",
+        "lib/",
+        "tools/ir-generator",
+    ],
+    visibility = [":__subpackages__"],
+)
+
+# This rule helps split some circular dependencies between subdirectories.
```
mbudiu-vmw, line 98: If we have circular dependencies this may be a bug, perhaps we should investigate.
Could you file an issue describing the dependency you found?

BUILD.bazel
```diff
@@ -0,0 +1,379 @@
...
+        ":p4c_ir_generator_lib",
+        ":p4c_toolkit",
+    ],
+)
+
+genrule(
+    name = "ir_generated_files",
+    srcs = [
+        "frontends/p4-14/ir-v1.def",
+        "backends/bmv2/bmv2.def",
```
mbudiu-vmw, line 168: Could this be factored so that the backend and the frontend are not mingled together?
Ideally the backend should be treated as an add-on that depends on the frontend.
I..e, if we add a new backend we should not need to edit the main makefile, esp. if the backend is an extension.
This is not a must, but would be nice to have.

README.md
mbudiu-vmw: An extension works by adding a symlink from the extensions folder to a source directory. Then the *.def files and the Cmakefile are picked up by the build automatically. There are some conventions on the structure of the makefile. As an example you can see http://github.com/vmware/p4c-xdp. It would be nice to make sure that one can be added to the Bazel build without editing any of the files in this repository.

bazel/bison.bzl
```diff
@@ -0,0 +1,65 @@
+"""Build rule for generating C or C++ sources with Bison."""
```
mbudiu-vmw, line 1: I don't really know how Bazel works, but it would be nice if you didn't need to copy this file to every project using bison. Is there some kind of a (trusted) repository of such rules which would allow them to be imported dynamically?

bazel/p4c_deps.bzl
```diff
@@ -0,0 +1,52 @@
...
+        #     name = "com_github_p4lang_p4runtime",
+        #     path = "@com_github_p4lang_p4c//:control-plane/p4runtime/proto",
+        # )
+        #
+        # We use git_repository as a workaround; the version used here should
+        # ideally be kept in sync with the submodule control-plane/p4runtime.
+        git_repository(
+            name = "com_github_p4lang_p4runtime",
+            remote = "https://github.com/p4lang/p4runtime",
+            commit = "776797d4bc7c2e5f1a7249f73f1c879a91688e75",
```
mbudiu-vmw, line 29: So we have to edit this manually every time we make a commit to p4runtime?

smolkaj: Yes and no. We can refer to any p4runtime commit here, it doesn't necessarily have to be the latest one. So we do not have to update this every time we make a commit to p4runtime.
This works much like git submodules, which p4c is already using for the cmake-based build. As explained in the comment, we would ideally have this rule point at the submodule, so that cmake and Bazel would be guaranteed to use the same version of p4runtime, but due to a Bazel limitation we cannot currently do so. The good news is that if something brakes, CI will tell us.

antoninbas:
Stratum comes with its own p4c backend for fixed-function pipelines. It may be a better exercise to see if they can switch over to these new targets, as opposed to doing a POC with p4c-xdp which already has a different build system (cmake based) that works fine. Then Stratum does not need to maintain its own Bazel BUILD file for p4c.

Even if p4c-xdp can be updated to support both a Bazel build and a cmake build, the changes may not be neglectable. It's probably better to validate p4c Bazel support by working with a backend that already made the choice to use Bazel.

[The Stratum p4c backend](https://github.com/stratum/stratum/tree/master/stratum/p4c_backends)
[The Stratum p4c BUILD file](https://github.com/stratum/stratum/blob/master/bazel/external/p4c.BUILD)

@pudelkoM FYI

smolkaj:
That makes sense. @pudelkoM, would you wanna give this a try? I briefly tried myself but couldn't get Stratum to build out of the box

BUILD.bazel
smolkaj: Are you suggestion that we invoke the ir generator twice? Is that possible? If so I'm happy to refactor things that way if you'd prefer.

BUILD.bazel
smolkaj: This comment originates from the Stratum authors, so I don't know the details here.

bazel/bison.bzl
smolkaj: Agreed, that would be nice. At the moment, there is no trusted rule for Bison.

pudelkoM:
Interesting effort. In general this is good and makes p4c more accessible. 

Here's what I found by just switching out the dependency source:

- `p4c_frontend_midend` is not visible to Stratum: https://github.com/p4lang/p4c/pull/2430/files#diff-0ea22445d900db774c8deb00f4a4df55R248

- We're missing a target called `p4c_backend_p4test_lib` (was `backends/p4test:p4c_common_midend_pass` before):
https://github.com/stratum/stratum/blob/9f5bd2b285badbef11e81eca6c31d4a3c4342843/bazel/external/p4c.BUILD#L430-L448

smolkaj:
@pudelkoM thanks for the feedback. Those concerns should now be addressed.

pudelkoM:
Initial compilation proceeds now, but I have found a blocker. Stratum has custom IR definitions for the FPM backend: https://github.com/stratum/stratum/commit/8772c199480c85b6edc609638a34e6866c1d1c5c#diff-bfd2d8e393f767320fbcc0aa008a69cb

Maybe there is a way to inject them into the dependency.

mbudiu-vmw:
Indeed, this is the point of the exercise. The Makefiles pick up def files automatically without changing the base makefiles.

smolkaj:
@pudelkoM @mbudiu-vmw Are custom IR definitions the only place where p4c needs to know about extensions? This seems to be the case for Stratum at least. If that's the case I think we could figure something out to make things work.

mbudiu-vmw:
These are the hardest probably, because they generate C++ which is fed to the compiler to create the visitors, so they need to be known statically. We also do stuff like register tests to execute for each backend, but perhaps these are easier to deal with.

smolkaj:
Alright, I added support for custom IR definitions.
@pudelkoM, could you give this a shot? The example/bazel folder contains an example of how to integrate your IR extensions.

bazel/example/WORKSPACE.bazel
```diff
@@ -3,6 +3,12 @@ workspace(name = "example_p4_project")
 local_repository(
   name = "com_github_p4lang_p4c",
   path = "../..",
+  # This part is optional and only needed for custom backends with IR
+  # extensions.
+  repo_mapping = {
+    # Tells p4c where to look for `:ir_extensions` target: in this project.
+    "@com_github_p4lang_p4c_extension" : "@example_p4_project",
```
antoninbas, line 10: pretty neat way to resolve that circular dependency :)

antoninbas: @smolkaj BTW is `repo_mapping` supported for `http_archive` and `git_repository`?
It does not say so in the documentation, but this seems to imply it is supported: https://github.com/bazelbuild/bazel/issues/6845. I believe the documentation is just inaccurate?

pudelkoM: I'll try this tomorrow.

Probs for finding this! I'm not sure yet if this is a really great idea or horribly undermining Bazel concepts :D

smolkaj: It's worth noting that `repo_mapping` is not needed to do this. Something like `new_local_repository(name = "com_github_p4lang_p4c_extension", ...)` should also do the trick, though a bit less elegantly. I'll document this tomorrow since it looks like `repo_mapping` might indeed not be available for `http_archive` and `git_repository`.

pudelkoM: The mapping scheme seems to work!
Currently solving linker errors related to the move from gmp to boost::multiprecision

pudelkoM:
The current include path for p4c headers is not optimal:

```c++
#include "external/com_github_p4lang_p4c/frontends/p4/coreLibrary.h"
#include "external/com_github_p4lang_p4c/frontends/p4/methodInstance.h"
```

Could probably fixed with a bit of `include_prefix` / `strip_prefix`.

bazel/example/WORKSPACE.bazel
smolkaj: Awesome! :)

smolkaj: I also just confirmed that `repo_mapping` works for `http_archive` and `git_repository` as well, so we can leave it as is.

smolkaj:
@pudelkoM do you happen to know if we can make it so that a p4c prefix is used externally but not internally?

smolkaj:
@pudelkoM Actually, I may have misunderstood your question. Are you using
```
#include "external/com_github_p4lang_p4c/frontends/p4/coreLibrary.h"
#include "external/com_github_p4lang_p4c/frontends/p4/methodInstance.h"
```
and that works? What would you like this to look? The following should also work:
```
#include "frontends/p4/coreLibrary.h"
#include "frontends/p4/methodInstance.h"
```

pudelkoM:
Sorry, I should have been more clear.
`#include "external/com_github_p4lang_p4c/frontends/p4/coreLibrary.h"` is what we currently have.

Ideally it would look something like this: 
`#include "p4c/frontends/p4/coreLibrary.h"`

The `external/...` prefix is an undesirable Bazel artifact and just `frontend/...` does not express the origin of these headers. `p4c` is just a suggestion because of the project name. 


For additional context, here is how other project's includes look like:
```c++
#include "absl/synchronization/mutex.h"
#include "grpcpp/grpcpp.h"
#include "p4/v1/p4runtime.grpc.pb.h"
#include "tofino/bf_pal/bf_pal_port_intf.h"
```

fruffy:
So with this, if I have a backend that uses Bazel, can I just drop it into the extension folder and then compile as usual?

smolkaj:
@pudelkoM I agree `#include "p4c/frontends/p4/coreLibrary.h"` is what we want but I don't know how to make that happen, technically speaking.

@fruffy Not quite. Things work the other way around with Bazel in the sense that your extension would have to depend on (i.e., include) p4c. Very concretely, take a look at the bazel/example folder (included in this PR) for how to set things up.

smolkaj:
Let me know if there is anything else to do here before we can merge.

mbudiu-vmw:
what is the verdict about extensions?

smolkaj:
They now work, see the `bazel/example` folder.
@pudelkoM were you able to make this work for Stratum (modulo the slightly annoying include paths)?

mbudiu-vmw:
this Travis build was queued 5 days ago, and prevents us from merging

pudelkoM:
Sorry for the radio silence. I got wrapped up in other tasks.

I think this PR is an improvement of the current state and therefore should be merged. If we (Stratum project) find other issues when switching later, we'll report them.

smolkaj:
I just added an empty commit hoping this will help Travis out of the deadlock.

PR #2440, opened 2020-06-16
Fix issue 2307

hesingh:


hesingh:
@robertmacdavid or anyone else, please review this PR, thanks.

hesingh:
@jnfoster thanks for finding root cause for the issue of this PR.  Maybe, you could please review this PR.  It has few lines of code changes - thanks.

backends/bmv2/common/control.cpp
```diff
@@ -352,7 +352,9 @@ ControlConverter::convertTable(const CFG::TableNode* node,
             // 1) if there is at least one RANGE field, then the table is RANGE
             // 2) if there is at least one TERNARY or OPTIONAL field, then the table is TERNARY
             // 3) if there is a LPM field, then the table is LPM
-            // 4) otherwise the table is EXACT
+            // 4) if there is a LPM field, then the table is ternary or EXACT
```
antoninbas, line 355: cases 3) and 4) seems at odds with each other:
> // 3) if there is a LPM field, then the table is LPM
// 4) if there is a LPM field, then the table is ternary or EXACT

Besides, I think the rules are correct and do not need to be changed.

backends/bmv2/common/control.cpp
```diff
@@ -352,7 +352,9 @@ ControlConverter::convertTable(const CFG::TableNode* node,
...
+            // TODO: Do we also check if table is range or optional in 4?
```
antoninbas, line 356: I don't understand this TODO. The rules above are pretty straightforward and just need to be applied.

backends/bmv2/common/control.cpp
```diff
@@ -352,7 +352,9 @@ ControlConverter::convertTable(const CFG::TableNode* node,
...
+            LOG3("Tbl match key: " << match_type << " " << table_match_type);
```
antoninbas, line 357: This is not a very useful log for developers, you need to indicate which field is which (individual match key element match type vs deducted bmv2 table match type)

hesingh:
> I do not think this change is correct. The match type of this table:
> 
> ```
> table t_ternary {
> 
>   	key = {
>             h.h.t : ternary;
>             h.h.e : lpm;
>         }
> 
> 	actions = {
>             a;
>             a_with_control_params;
>         }
> 
> 	default_action = a;
>     }
> ```
> 
> will be set to "lpm" with your change, when it should be "ternary".

Sorry, I didn't know for this P4 program what the table type should be.  Thanks for letting me know.

hesingh:
> I do not think this change is correct. The match type of this table:
> 
> ```
> table t_ternary {
> 
>   	key = {
>             h.h.t : ternary;
>             h.h.e : lpm;
>         }
> 
> 	actions = {
>             a;
>             a_with_control_params;
>         }
> 
> 	default_action = a;
>     }
> ```
> 
> will be set to "lpm" with your change, when it should be "ternary".

According to Rule 3, the table should be "lpm", right?

antoninbas:
No, according ti rule 2 which has higher precedence, the table should be ternary.

hesingh:
After sending my new question, I too saw the precedence rule.  Sorry, my bad.

The fix should be that the rule 0 should be independently checked is check first because the rule's precedence is highest.

backends/bmv2/common/control.cpp
```diff
@@ -412,6 +415,7 @@ ControlConverter::convertTable(const CFG::TableNode* node,
             tkey->append(keyelement);
         }
     }
+    LOG1("table_match_type: " << table_match_type);
```
antoninbas, line 418: isn't `LOG1` for the most "important" log messages?

hesingh: Yes, I thought, to debug any future issue in the method, we could use a LOG1 and leave other logging clutter out.

However, I am open to change.

Would you like me to change LOG1 to LOG3?

antoninbas: It's a pretty minor piece of information that is also available in the generated JSON.
It has to be at least a LOG3, since the name of the table itself is logged with that verbosity:
```
LOG3("Processing " << dbp(table));
```

Issue #1914, opened 2019-04-30
Can p4c handle default parameter values for packages that are controls with type variables?


jafingerhut:
Attached is a program that tries to make one of the package parameters have a default value for one parameter.  I have tried a few variations, but nothing I have thought of enables this program to compile using `p4test`.

Should this be possible?  If so, what is the syntax?

FYI, this is related to a possible change/enhancement to the PSA architecture definition Calin and I are mulling over.  It would be nice if this were legal to do in some way.

[default-package-params.p4.txt](https://github.com/p4lang/p4c/files/3129630/default-package-params.p4.txt)

hanw:
Similar issue reported here? https://github.com/p4lang/p4c/issues/1639

jafingerhut:
Closing this one, as it does look like it is probably a duplicate of that earlier issue.

jafingerhut:
Had a chat with Mihai, who I believe mentioned that he thought these issues were different.  Re-opening this one, in case they are different.  Feel free to close as a duplicate if I've misjudged this.

jafingerhut:
Attaching 3 separate minor variants of the P4_16 program attached to this issue originally.  The latest version of p4c as of 2020-Jun-14 gives compile time errors for all 3 variants.  I am happy to try other variants if people can suggest what should work, if there is anything that currently does.

```
$ ./compile.sh 
+ p4c --version
p4c 1.2.0 (SHA: ebd446fac BUILD: DEBUG)
+ p4c --target bmv2 --arch v1model default-package-params-try1.p4
default-package-params-try1.p4(31):syntax error, unexpected IDENTIFIER "H"
control EmptyVerifyChecksum(inout H
                                  ^
error: 1 errors encountered, aborting compilation
+ p4c --target bmv2 --arch v1model default-package-params-try2.p4
In file: /home/andy/p4c/frontends/p4/specialize.h:50
Compiler Bug: /home/andy/p4c/frontends/p4/specialize.h:50: Null insertion

+ p4c --target bmv2 --arch v1model default-package-params-try3.p4
default-package-params-try3.p4(49): [--Werror=not-found] error: H: declaration not found
                       VerifyChecksum<H, M> vr = EmptyVerifyChecksum<H, M>(),
                                                                     ^
default-package-params-try3.p4(49): [--Werror=not-found] error: M: declaration not found
                       VerifyChecksum<H, M> vr = EmptyVerifyChecksum<H, M>(),
```
[...]

PR #2441, opened 2020-06-16
Substitute type parameters as well in default arguments

### LINKED ISSUE #2019: Handling of generic arguments with default values is incorrect
The following program does not compile; in particular, the type argument for the default value should be scoped to the `vr` parameter's type argument H.

```
control V<H>(inout H hdr);
control EmptyV<H>(inout H hdr) {
    apply { }
}

struct E {}

package S<H>(V<H> vr = EmptyV<H>());
S<E>() main;
```

mbudiu-vmw:
Fixes #2019


mbudiu-vmw:
Fixes #1914

jafingerhut:
These changes add a new test file named issue2341-bmv2.p4.  Is that name what you intended?  I do not see how it is related to this PR, if that is the intent: https://github.com/p4lang/p4c/pull/2341

testdata/p4_16_samples_outputs/issue1914-1-first.p4
```diff
@@ -0,0 +1,67 @@
...
+control EmptyVerifyChecksum<H, M>(inout H hdr, inout M meta) {
+    apply {
+    }
+}
+
+control Ingress<H, M>(inout H hdr, inout M meta, inout standard_metadata_t standard_metadata);
+control Egress<H, M>(inout H hdr, inout M meta, inout standard_metadata_t standard_metadata);
+control ComputeChecksum<H, M>(inout H hdr, inout M meta);
+control Deparser<H>(packet_out b, in H hdr);
+package MySwitch<H, M>(Parser<H, M> p, VerifyChecksum<H, M> vr=EmptyVerifyChecksum<_, _>(), Ingress<H, M> ig, Egress<H, M> eg, ComputeChecksum<H, M> ck, Deparser<H> dep);
```
jafingerhut, line 18: This intermediate output for this test program has `EmptyVerifyChecksum<_, _>()`.  Is this expected?  Is there some brief explanation of why it is `<_, _>` instead of `<H, M>` for this test program, whereas most of the others have `<H, M>`, I think?

jafingerhut:
I have looked at all of the test programs and expected output files added in this PR, and they look reasonable to me, with one question that I have added in a separate comment.

I am not knowledgeable enough of p4c compiler internals to review that part of this PR.

testdata/p4_16_samples_outputs/issue1914-1-first.p4
mbudiu-vmw: Yes, this was the symptom of a bug.

mbudiu-vmw:
Unfortunately the fix introduces a small regression: if a type variable is not used anywhere the compiler will complain that it cannot infer a value for it. Previously it would infer "don't care". The solution is to explicitly specify "don't care" for the type variable when instantiating. This is not ideal, but it's correct. The reason for this change is that we have to wait until the default arguments are substituted to find out whether a variable may be unused, and the default arguments pass has to be run after we have inferred type substitutions for type variables... This could probably be fixed by doing default argument substitution in typeChecking, but that pass is complicated enough as it is.

hesingh:
> Unfortunately the fix introduces a small regression: if a type variable is not used anywhere the compiler will complain that it cannot infer a value for it. Previously it would infer "don't care". The solution is to explicitly specify "don't care" for the type variable when instantiating. This is not ideal, but it's correct. The reason for this change is that we have to wait until the default arguments are substituted to find out whether a variable may be unused, and the default arguments pass has to be run after we have inferred type substitutions for type variables... This could probably be fixed by doing default argument substitution in typeChecking, but that pass is complicated enough as it is.

The solution looks reasonable to me.

Issue #2435, opened 2020-06-12
p4test: x: declaration not found


fruffy:
Running `p4test` there seem to be a problem with the way variables are propagated in the mid end. This only happens in combination with exit statements. There are two examples. One happens after inlining a global action, the other happens after a constant key is simplified. In both cases `MidEnd_25_LocalCopyPropagation` removes variables that it should not. 

[key_after_exit.p4.txt](https://github.com/p4lang/p4c/files/4769830/key_after_exit.p4.txt)
```
MidEnd_20_FlattenInterfaceStructs
MidEnd_21_ReplaceSelectRange
MidEnd_22_Predication
MidEnd_23_MoveDeclarations
MidEnd_24_ConstantFolding
MidEnd_25_LocalCopyPropagation
[--Werror=not-found] error: key_0: declaration not found
Done.*/
```

[global_action_after_exit.p4.txt](https://github.com/p4lang/p4c/files/4769831/global_action_after_exit.p4.txt)
```
MidEnd_20_FlattenInterfaceStructs
MidEnd_21_ReplaceSelectRange
MidEnd_22_Predication
MidEnd_23_MoveDeclarations
MidEnd_24_ConstantFolding
MidEnd_25_LocalCopyPropagation
bugs/crash/global_action_after_exit.p4(348): [--Werror=not-found] error: val: declaration not found
action do_action(inout ethernet_t val) {
```

hesingh:
@mbudiu-vmw and @ChrisDodd  In our prior discussion, we seem to be converging on removing exit statements inside action if the action includes any `hasOut` parameter.  

This issue uses an exit inside the control apply block and global action.  I thought we could leave this exit for the bmv2 backend to deal with.  If we do, the midend screws up the IR.  

```p4
action do_action(inout ethernet_t val) {
    val.eth_type = 0xDEAD;
}
control ingress(inout Headers h, inout Meta m, inout standard_metadata_t sm) {

    apply {
        exit;
        do_action(h.eth_hdr);
    }
}
```

hesingh:
@mbudiu-vmw Having poked around more, I see p4c already supports an exit statement inside if-else.  For example, I found `p4c/testdata/p4_16_samples/exit2.p4`.  This test case do not include any action with `hasOut` parameter or even a parameter.  But the control the action is inside has an out parameter.  So now, one should also check if the control has a `hasOut` parameter and remove exit statement(s) from control actions.

[...]

Issue #2382, opened 2020-05-18
Another pass that incorrectly converts exit statements 


fruffy:
I have another instance where there are issues with exit, this time in `InlineActions`.

```p4
control ingress(inout Headers h, inout Meta m, inout standard_metadata_t sm) {
    @name("do_action") action do_action_0(inout bit<8> val_0) {
        val_0 = 8w1;
        exit;
    }
    @name("do_action_ingress") action do_action_ingress_0() {
        do_action_0(h.h.a);
    }
    apply {
        do_action_ingress_0();
    }
}
```

is converted to

```p4
control ingress(inout Headers h, inout Meta m, inout standard_metadata_t sm) {
    @name("do_action_ingress") action do_action_ingress_0() {
        {
            bit<8> val = h.h.a;
            val = 8w1;
            exit;
            h.h.a = val;
        }
    }
    apply {
        do_action_ingress_0();
    }
}

```

Should I collect all the examples of exit problems in one thread or keep filing individual issues?
[inline_actions_exit.p4.txt](https://github.com/p4lang/p4c/files/4643455/inline_actions_exit.p4.txt)
[inline_actions_exit.stf.txt](https://github.com/p4lang/p4c/files/4643457/inline_actions_exit.stf.txt)

mbudiu-vmw:
It is more useful to have these in one place; if the fix is common to all of them we should have all these examples as regression tests. I would at least cross-link from the original issue to this one.

mbudiu-vmw:
Tagged as duplicate, but kept for the test case.

hesingh:
What is expected output for this issue with a fixed p4c?

Should the "exit;" just disappear from the output?

mbudiu-vmw:
For this particular program removing exit would preserve the semantics.

hesingh:
These are the code changes to p4c to fix this issue.  I don't see a need to check any use because the statements just have to be removed. 

```
diff --git a/frontends/p4/simplifyDefUse.cpp b/frontends/p4/simplifyDefUse.cpp
index 14615fab..7e89c01f 100644
--- a/frontends/p4/simplifyDefUse.cpp
+++ b/frontends/p4/simplifyDefUse.cpp
@@ -640,6 +640,12 @@ class RemoveUnused : public Transform {
         }
         return statement;
     }
+    const IR::Node* postorder(IR::ExitStatement* statement) override {
+        return new IR::EmptyStatement();
+    }
[...]

Issue #2374, opened 2020-05-14
Def-Use and exit statements


fruffy:
Another case, where exit statements combined with out parameters are not correctly handled in respect to the recent spec changes:

After `FrontEnd_32_SimplifyDefUse`, the following program:

```p4
control ingress(inout Headers h, inout Meta m, inout standard_metadata_t sm) {
    @name("exit_action") action exit_action_0() {
        exit;
    }
    @name("assignment_action") action assignment_action_0(out bit<16> val) {
        val = 16w0xbeef;
    }
    @name("simple_action") action simple_action_0(inout bit<16> val) {
        val = 16w0xdead;
        exit_action_0();
        assignment_action_0(val);
    }
    apply {
        simple_action_0(h.eth_hdr.eth_type);
    }
}
```
is transformed into:

```p4
control ingress(inout Headers h, inout Meta m, inout standard_metadata_t sm) {
    @name("exit_action") action exit_action_0() {
        exit;
    }
    @name("assignment_action") action assignment_action_0(out bit<16> val) {
        val = 16w0xbeef;
    }
    @name("simple_action") action simple_action_0(inout bit<16> val) {
        ;
        exit_action_0();
        assignment_action_0(val);
    }
    apply {
        simple_action_0(h.eth_hdr.eth_type);
    }
}
```
This disregards the exit call in `exit_action_0`. 

[nested_exit_out_param.p4.txt](https://github.com/p4lang/p4c/files/4627702/nested_exit_out_param.p4.txt)
[nested_exit_out_param.stf.txt](https://github.com/p4lang/p4c/files/4627703/nested_exit_out_param.stf.txt)

mbudiu-vmw:
I have labeled this as a duplicate, probably the same fix should address this one - once we have it.

mbudiu-vmw:
the exit statement should be eliminated before reaching this point.

jafingerhut:
@fruffy Do you know if this bug was fixed by later commits?

fruffy:
This bug is still not fixed.

Issue #2358, opened 2020-05-04
Follow-up issue on exit statements


fruffy:
As @jafingerhut pointed out in #2225 , there might be some fallout from the exit changes. One of them is an edge-case in how to handle `SideEffectOrdering`.

`SideEffectOrdering` transforms the following program:

```p4
control ingress(inout Headers h, inout Meta m, inout standard_metadata_t sm) {
    action NoAction() {
    }
    action call_exit(inout bit<48> val, in bit<48> val1) {
            val = 1;
            exit;
    }
    apply {
        call_exit(h.eth_hdr.src_addr, h.eth_hdr.src_addr);
    }
}
```
into

```p4
control ingress(inout Headers h, inout Meta m, inout standard_metadata_t sm) {
    @name("call_exit") action call_exit_0(inout bit<48> val, in bit<48> val1) {
        val = 48w1;
        exit;
    }
    bit<48> tmp;
    bit<48> tmp_0;
    apply {
        {
            tmp = h.eth_hdr.src_addr;
            tmp_0 = h.eth_hdr.src_addr;
            call_exit_0(tmp, tmp_0);
            h.eth_hdr.src_addr = tmp;
        }
    }
}
```
which still looks okay, but ultimately this change causes this transformation in the MidEnd pass `RemoveActionParameters`:

```p4
control ingress(inout Headers h, inout Meta m, inout standard_metadata_t sm) {
    bit<48> tmp;
    bit<48> tmp_0;
    bit<48> val;
    bit<48> val1;
    @name("ingress.call_exit") action call_exit() {
        val = tmp;
        val1 = tmp_0;
        val = 48w1;
        tmp = val;
        exit;
        tmp = val;
    }
    apply {
        tmp = h.eth_hdr.src_addr;
        tmp_0 = h.eth_hdr.src_addr;
        call_exit();
        h.eth_hdr.src_addr = tmp;
    }
}

```
Here, `h.eth_hdr.src_addr` will never be assigned the value `1`, from what I can tell. 

[side_effect_ordering_exit.p4.txt](https://github.com/p4lang/p4c/files/4574281/side_effect_ordering_exit.p4.txt)
[side_effect_ordering_exit.stf.txt](https://github.com/p4lang/p4c/files/4574282/side_effect_ordering_exit.stf.txt)

ChrisDodd:
[...]

PR #2443, opened 2020-06-17
Fix exit bugs

hesingh:
This PR fixes the following issues

https://github.com/p4lang/p4c/issues/2358
https://github.com/p4lang/p4c/issues/2374
https://github.com/p4lang/p4c/issues/2382

Also, fixes both test P4 programs included in the following issue.
https://github.com/p4lang/p4c/issues/2435

Existing tests in p4c/testdata/p4_16_samples/exit?.p4 are left untouched in case the bmv2 backend uses exit testing of these tests.  The tests are special because they include exit statements in if-else code blocks and actions have no parameters to check for inout/out.  If asked of, I can look into removing exits from test tests as well. 

Other test outputs have changed in two ways:

1. exit statements have been removed in output if the action parameter have inout/out property.
2. Some outputs repeat the same warning. This is because the `RemoveAllUnusedDeclarations` pass is called an extra time by the new pass of this PR.


hesingh:
@jafingerhut Please review the outputs of these P4 exit programs which fruffy filed - thanks.

```bash
global_action_after_exit.p4
inline_actions_exit.p4
key_after_exit.p4
nested_exit_out_param-bmv2.stf
nested_exit_out_param-bmv2.p4
side_effect_ordering_exit.p4
```

mbudiu-vmw:
I don't see any test with a conditional exit statement: `if (cond) exit;`
The way the code is structured I don't think it handles this case.

hesingh:
> I don't see any test with a conditional exit statement: `if (cond) exit;`
> The way the code is structured I don't think it handles this case.

If you see my notes in the PR, I said if-else statements with exit exist in p4c/testdata/p4_16_sample/exit?.p4.  One example is exit2.p4.  exit2.p4 does not have any action with a parameter so there is no way to check for inout or out.  I'd be happy to change p4c code of this PR to perform an additional check which will catch that action with exit resides in control with an out parameter and thus exit statements should be removed.  Sound reasonable?

mbudiu-vmw:
I don't think that your approach will work. You are not really handling the case of actions with copy-out arguments where the action does not exit on all paths. Also, you are not handling the case of a table where some actions may call exit.

hesingh:
> I don't think that your approach will work. You are not really handling the case of actions with copy-out arguments where the action does not exit on all paths. Also, you are not handling the case of a table where some actions may call exit.

Let me look into handling the case of actions with copy-out args where the action does not exit on all paths. 

Regarding handling the case of a table where some actions may call exit, I am handling the first exit encountered because then the control has to exit.

mbudiu-vmw:
I don't understand your argument about tables. Consider two actions a and b where a will exit but b will not. If both can appear in table t how are you handling t.apply()?

fruffy:
Did a really quick check on this. It passes the files I have submitted and also the exit_combination files in https://github.com/p4gauntlet/p4_tv/tree/master/false_friends. 
However, here is an example that leads to a wrong result:
[exit_combination_18.p4.txt](https://github.com/p4lang/p4c/files/4794188/exit_combination_18.p4.txt)
The exit is erroneously removed because there is a subsequent return.

hesingh:
> I don't understand your argument about tables. Consider two actions a and b where a will exit but b will not. If both can appear in table t how are you handling t.apply()?

Did you see p4c/testdata/p4_16_samples/nested_exit_out_param-bmv2.p4?  One action exits while another does not.  However, since these are nested actions, there is only one t.apply() for which test passes.

I will add the test case you speak of and handle its t.apply() with new p4c changes.

hesingh:
> Did a really quick check on this. It passes the files I have submitted and also the exit_combination files in https://github.com/p4gauntlet/p4_tv/tree/master/false_friends.
> However, here is an example that leads to a wrong result:
> [exit_combination_18.p4.txt](https://github.com/p4lang/p4c/files/4794188/exit_combination_18.p4.txt)
> The exit is erroneously removed because there is a subsequent return.

Thanks for checking.  Let me look into your new exit_combination_18.p4 test.  As one can see, I have built a framework in a new frontend pass to catch any exit in action or control.  So, the new tests asked of by Mihai and fruffy should be possible to support with more changes which won't be major.

hesingh:
This is the program in `p4c/testdata/p4_16_samples/exit2.p4`.

```p4
control ctrl(out bit<32> c) {
    bit<32> x;

    action e() {
        exit;
        x = 1;
    }

    apply {
        bit<32> a;
        bit<32> b;

        a = 0;
        b = 1;
        c = 2;
        if (a == 0) {
            b = 2;
            e();
            c = 3;
        } else {
            b = 3;
            e();
            c = 4;
        }
        c = 5;
    }
}

control noop(out bit<32> c);
package p(noop _n);
p(ctrl()) main;
```

Does the following transformation as `exit2-frontend.p4` look ok to @jafingerhut and @mbudiu-vmw 

```p4
control ctrl(out bit<32> c) {
    bit<32> a_0;
    @name("ctrl.e") action e() {
    }
    @name("ctrl.e") action e_2() {
    }
    apply {
        a_0 = 32w0;
        if (a_0 == 32w0) {
            e();
        } else {
            e_2();
        }
        c = 32w5;
    }
}

control noop(out bit<32> c);
package p(noop _n);
p(ctrl()) main;
```

mbudiu-vmw:
Change this example to only call e on one branch of the if.

hesingh:
@mbudiu-vmw Now that I got time to think, I totally agree with you that this pass will only handle simple cases of exit statement removal such as in fruffy's issues.  In the next commit fruffy's `exit_combination_18.p4` is fixed.

More complex cases which you gave examples of (one table with two actions, or if has exit but else does not) are dealt with complicated logic which you have with `hasReturned/hasReturns` in the `RemoveReturns` pass. 

Right now this PR's new pass runs earlier than the `RemoveReturns` or `SideEffectsOrdering` passes.  I think, we could do simple processing in the `rmExits` pass of this PR and handle the complicated cases in subsequent passes.  Right now the `rmExits` pass only processes `MethodCallStatement` (to catch nested actions with exit) and the `ExitStatement`.

mbudiu-vmw:
The general solution should handle all these special cases. I don't think there is any point to implement a solution that only works for some programs if we can give a general solution.

hesingh:
Ok, got it.  Let me mull over a general solution and get back.

ChrisDodd:
random question -- how do you mark a pull request as a 'draft' (as this one is), so it can't be merged?

mbudiu-vmw:
I think that when you submit you get a choice of whether it's a draft or not.

hesingh:
> random question -- how do you mark a pull request as a 'draft' (as this one is), so it can't be merged?

Moved to draft because after Mihai's review I have to work on generalizing the pass.

hesingh:
First the if-else case has to be clarified.  
This is the P4 program with exit in if-block but not in the else-block.

```p4
control ctrl(out bit<32> c) {
    bit<32> x;
    bit<32> a;
    bit<32> b;

    action e() {
        if (a == 0) {
            b = 2;
            exit;
            c = 3;
        } else {
            b = 3;
            c = 4;
        }
        x = 1;
    }

    apply {
        a = 0;
        b = 1;
        c = 2;
        e();
        c = 5;
    }
}
```

For a first analysis of the program, it should be this:
Move else blocks to a new action.
If any statements remain in the original action using if-else, the statements move to a new action. Then the actions are called in a specific sequence.  For any statements in action before the if block, move the statements to yet another new action.  In a subsequent analysis get rid of the if-block in action e. 

The generalization is new work for p4c because I don't think a test like this exits in the p4c repo. I admit haven't looked hard. 

```p4
control ctrl(out bit<32> c) {
    bit<32> x;
    bit<32> a;
    bit<32> b;

    action e() {
        if (a == 0) {
            b = 2;
            exit;
            c = 3;
        }
    }

    action ee() {
        b = 3;
        c = 4;
    }

    action ex() {
        x = 1;
    }

    apply {
        a = 0;
        b = 1;
        c = 2;
        e();
        ee();
        ex();
        c = 5;
    }
}
```

ChrisDodd:
It is already the case that some targets/architectures may not support conditionals (if/else) in action at all.  For such a target it is necessary to convert all conditionals in actions into separate actions (and perhaps separate tables); this will likely be done as a midend pass.  So if such is transform is done before exit removal, then exit removal need not worry about conditionals.

mbudiu-vmw:
The exit removal should be done when the expansion of copy-out is done, otherwise the semantics of the program after copy-out expansion is actually incorrect. This happens in the front-end, in the side-effect ordering. So I am afraid that side-effect ordering should also do exit removal right away.

hesingh:
Thanks @ChrisDodd and @mbudiu-vmw.

@mbudiu-vmw The test case you wanted me to support with only one of if- or else- blocks, should we only support the test in a control apply block or also inside an action?  As Chris said, some targets do not support if-else inside an action, so we could skip the action?

mbudiu-vmw:
As I said, the removeExits pass is almost correct, we just have to fix it to treat correctly copy-out.
It handles all these cases already.

fruffy:
It looks like this may not be needed any more. Should we close this? #2374 and #2382 are still not fixed, but they might be resolved with a different pull request.

PR #2448, opened 2020-06-22
Clarified an STF test case with a too-short packet

jnfoster:
The pseudo-code for `packet_in.extract` in the P4 Language Specification implies that the packet buffer should be unmodified if the packet is too short.

The second test case in `parser_error-bmv2.stf` is such a case:
```
packet 0 012345678910
expect 0 000000000000 000000000000 0000
```
The parser attempts to extract the 14-byte Ethernet header, which should fail. Then in the `ingress` control, it adds an all-zero `Ethernet` header. 

The `expect` statement for this test is correct, but potentially confusing -- STF only checks that the prefix of the emitted packet matches the bitstring. In particular, it's not obvious that the original packet is preserved in the output.

This commit clarifies the test case, by explicitly showing that the original packet should not have been extracted.


ChrisDodd:
It seems like this test also suffers from the "not well-defined what happens when standard_metadata.egress_spec is not set" problem -- bmv2-ss initializes it to port 0, but the language spec doesn't say.

jnfoster:
Good point. Added an assignment to clarify this...

PR #2451, opened 2020-06-25
Tuple elim

mbudiu-vmw:
This is paired with https://github.com/p4lang/p4-spec/pull/877
Note that tuple fields are not (yet) left values.

### DIFF
--- a/frontends/common/constantFolding.cpp
+++ b/frontends/common/constantFolding.cpp
@@ -585,31 +585,42 @@ const IR::Node* DoConstantFolding::postorder(IR::Member* e) {
         auto expr = getConstant(e->expr);
         if (expr == nullptr)
             return e;
-        auto structType = type->to<IR::Type_StructLike>();
-        if (structType == nullptr)
-            BUG("Expected a struct type, got %1%", type);
-        if (auto list = expr->to<IR::ListExpression>()) {
-            bool found = false;
-            int index = 0;
-            for (auto f : structType->fields) {
-                if (f->name.name == e->member.name) {
-                    found = true;
-                    break;
-                }
-                index++;
-            }
 
-            if (!found)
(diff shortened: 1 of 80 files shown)


frontends/common/constantFolding.cpp
```diff
@@ -580,31 +580,42 @@ const IR::Node* DoConstantFolding::postorder(IR::Member* e) {
...
-            BUG_CHECK(ne != nullptr, "Could not find field %1% in initializer %2%", e->member, si);
-            return CloneConstants::clone(ne->expression);
+            if (auto list = expr->to<IR::ListExpression>()) {
+                result = CloneConstants::clone(list->components.at(static_cast<size_t>(index)));
+            }
         } else {
-            BUG("Unexpected initializer: %1%", expr);
+            auto structType = type->to<IR::Type_StructLike>();
+            if (structType == nullptr)
+                BUG("Expected a struct type, got %1%", type);
```
ChrisDodd, line 594: probably should be

    } else if (auto structType = type->to<IR::Type_StructLike>()) {
            :
    } else {
        BUG("Expected a struct or tuple type, got %1%", type);

ChrisDodd (approved):

mbudiu-vmw:
@fruffy : this fails in the gauntlet check, but the diagnostic is not clear. How do we interpret this?

mbudiu-vmw:
I actually suspect that what happened is that this is introducing a new feature in the language (tuple fields), which it not yet modeled in Gauntlet, so Gauntlet fails.

fruffy:
Ugh. I knew this would happen eventually. Yes this construct is not supported yet. I think I should let the check only fail if it reports an actual validation error. Otherwise this will keep happening. Implemented a hotfix that should pass this test at least.

fruffy:
Actually the fix for this was really simple. It was just a field naming issue.

PR #2452, opened 2020-06-25
Disable merging of specific state

AashleshaPatil-BFN:


zma-barefoot:
I think @dontMerge sound better than @noMerge (grammatically). Why is gress required?

ir/type.cpp
```diff
@@ -49,6 +49,7 @@ const cstring IR::Annotation::pureAnnotation = "pure";
 const cstring IR::Annotation::noSideEffectsAnnotation = "noSideEffects";
 const cstring IR::Annotation::noWarnAnnotation = "noWarn";
 const cstring IR::Annotation::matchAnnotation = "match";
+const cstring IR::Annotation::noMergeStateAnnotation = "nomerge";
```
zma-barefoot, line 52: dontMerge

AashleshaPatil-BFN:
> I think @dontMerge sounds better than @noMerge (grammatically). Why is gress required?

gress is required for p4-14 programs

jafingerhut:
Is there some recommended place to document this annotation and its intended behavior?  If the P4_16 language spec, then the LDWG can take it up.  If that does not seem like the appropriate place, then where?

mbudiu-vmw:
Also, it would be nice to have a way for targets to add such annotations without needing to change the common base.def file.
This PR will change the way programs are compiled, because all other annotations on a state will be ignored when merging. Previously states with annotations were never merged.

frontends/p4/simplifyParsers.cpp
```diff
@@ -113,6 +113,9 @@ class CollapseChains : public Transform {
         // has no other incoming edges.
         for (auto oe : *transitions) {
             auto node = oe.first;
+            if (!node->annotations->annotations.empty()) {
+                continue;
+            }
```
liujed-intel, line 118: Isn't this redundant with https://github.com/p4lang/p4c/pull/2452/files#diff-52d560b4ecf435649572dd0911559f23R130-R132?

AashleshaPatil-BFN: Consider a chain of states : s1 -> s2(@dontMerge) -> s3. 
s2 should not merge with s1 or s3. . Line 130 - 132 will avoid merging of s1 and s2. Line 116-118 will avoid merging of s2 and s3.

liujed-intel: Ah ha, I didn't notice the `node` vs. `next` difference. Thanks.

PR #2453, opened 2020-06-26
refactor bmv2 backend to reuse more code for psa support

hanw:
I took the extern traits definition from control-plane. this results in some code duplication, but it might be better to keep  bmv2 backend implementation separate from the control-plane code. There isn't that much change. The main change is to make ControlConverter a template class, so we customize the type name for each architecture, as was done in p4runtime generation. With this change, we could support psa action_profile and action_selector @yunheL  FYI.


mbudiu-vmw:
Could you indicate in a comment which code has been just moved without changes so we can avoid trying to chase differences between different files?

backends/bmv2/common/control.h
```diff
@@ -39,18 +48,777 @@ class ControlConverter : public Inspector {
...
+                    keyelement->emplace("mask", Util::JsonValue::null);
+                tkey->append(keyelement);
+            }
+        }
+        LOG3("table_match_type: " << table_match_type);
+        result->emplace("match_type", table_match_type);
+        ctxt->conv->simpleExpressionsOnly = false;
+
+        auto propertyName = Standard::ActionProfileTraits<arch>::propertyName();
+        auto impl = table->properties->getProperty(propertyName);
```
hanw, line 148: Used to be `auto impl = table->properties->getProperty("implementation");`

backends/bmv2/common/control.h
```diff
@@ -39,18 +48,777 @@ class ControlConverter : public Inspector {
...
+                    ::error(ErrorType::ERR_EXPECTED, "%1% must be a constant", size_expr);
+                    size = 0;
+                } else {
+                    size = size_expr->to<IR::Constant>()->asInt();
+                }
+                action_profile->emplace("max_size", size);
+            };
+
+            auto actionSelectorName = Standard::ActionSelectorTraits<arch>::typeName();
+            if (implementationType->name == actionSelectorName) {
```
hanw, line 648: Used to be `implementationType->name == BMV2::TableImplementation::actionSelectorName`

backends/bmv2/common/control.h
```diff
@@ -39,18 +48,777 @@ class ControlConverter : public Inspector {
...
+            apname = decl->controlPlaneName();
+            auto dcltype = ctxt->typeMap->getType(pathe, true);
+            if (!dcltype->is<IR::Type_Extern>()) {
+                ::error(ErrorType::ERR_UNEXPECTED, "%1%: unexpected type for implementation", dcltype);
+                return false;
+            }
+            auto type_extern_name = dcltype->to<IR::Type_Extern>()->name;
+            auto actionProfileName = Standard::ActionProfileTraits<arch>::propertyName();
+            auto actionSelectorName = Standard::ActionSelectorTraits<arch>::propertyName();
+            if (type_extern_name == actionProfileName) {
```
hanw, line 700: Used to be `implementationType->name == BMV2::TableImplementation::actionProfileName`

backends/bmv2/common/control.h
```diff
@@ -39,18 +48,777 @@ class ControlConverter : public Inspector {
...
+                table->emplace("type", "indirect");
+            } else if (type_extern_name == actionSelectorName) {
```
hanw, line 702: Used to be `type_extern_name== BMV2::TableImplementation::actionSelectorName`

backends/bmv2/common/helpers.h
```diff
@@ -66,11 +56,204 @@ class V1ModelProperties {
     static const cstring validField;
 };
 
+namespace Standard {
```
hanw, line 59: Code below was copied from control-plane/p4RuntimeArchStandard.cpp.

backends/bmv2/common/helpers.h
```diff
@@ -85,7 +268,14 @@ struct ConversionContext {
 
     // for action profile conversion
     Util::JsonArray*                 action_profiles;
-    SharedActionSelectorCheck*       selector_check;
+
```
hanw, line 271: Changed to a map to avoid making ConversionContext a template class. As the SharedActionSelectorCheck needs to be templated.

backends/bmv2/common/sharedActionSelectorCheck.cpp
```diff
@@ -1,85 +0,0 @@
-/*
```
hanw, line 1: Moved to .h, becuase this class need to templatized.

backends/bmv2/common/sharedActionSelectorCheck.h
```diff
@@ -33,10 +35,11 @@ using SelectorInput = std::vector<const IR::Expression *>;
 // This pass makes sure that when several match tables share a selector, they use the same input for
 // the selection algorithm. This is because bmv2 considers that the selection key is part of the
 // action_selector while v1model.p4 considers that it belongs to the table match key definition.
+template <Standard::Arch arch>
 class SharedActionSelectorCheck : public Inspector {
+    BMV2::ConversionContext* ctxt;
```
hanw, line 40: Mostly moved from cpp, changed constructor to take ConversionContext, instead of refmap and typemap. Refmap and typemap are taken from the pointer inside ConversionContext. Same implementation as before.

backends/bmv2/common/sharedActionSelectorCheck.h
```diff
@@ -59,11 +62,64 @@ class SharedActionSelectorCheck : public Inspector {
...
+            ::error(ErrorType::ERR_EXPECTED, "%1%: expected a reference to an instance", pathe);
+            return false;
+        }
+        auto dcltype = typeMap->getType(pathe, true);
+        if (!dcltype->is<IR::Type_Extern>()) {
+            ::error(ErrorType::ERR_UNEXPECTED, "%1%: unexpected type for implementation", dcltype);
+            return false;
+        }
+        auto type_extern_name = dcltype->to<IR::Type_Extern>()->name;
+        auto actionSelectorName = Standard::ActionSelectorTraits<arch>::typeName();
```
hanw, line 92: Only change in this class.

backends/bmv2/common/control.h
```diff
@@ -39,18 +48,777 @@ class ControlConverter : public Inspector {
  protected:
     Util::IJson* convertTable(const CFG::TableNode* node,
                               Util::JsonArray* action_profiles,
-                              BMV2::SharedActionSelectorCheck& selector_check);
-    void convertTableEntries(const IR::P4Table *table, Util::JsonObject *jsonTable);
-    cstring getKeyMatchType(const IR::KeyElement *ke);
+                              BMV2::SharedActionSelectorCheck<arch>* selector_check) {
```
hanw, line 51: Mostly moved from cpp, except the lines where I commented.

backends/bmv2/common/helpers.h
mbudiu-vmw: Do we need it to be copied, or can we share it somehow?

hanw: The `mapUnitName` method in CounterlikeTraits depends on protobuf, which does not make much sense to carried over to bmv2 backend. See line 190 in p4RuntimeArchStandard.cpp. @antoninbas FYI.

Issue #2201, opened 2020-02-18
Unsupported types for log_msg calls


jafingerhut:
I had a fairly decent sized v1model P4 program that was causing simple_switch to give an error message very soon after starting up, about some unexpected contents inside of the BMv2 JSON file, and it took a while to whittle it down and discover it was because I was trying to do log_msg on a value of type `bool`.  Casting it to a `bit<1>` type caused things to work.  It would be nice to at least document what types are supported, and preferably give a compile time error message for unsupported types.

From my experiments using the attached P4 program, it appears that these types cause either cryptic errors when starting simple_switch, or a run-time error and simple_switch crash when the log_msg statement is executed:

+ bool - cryptic error at simple_switch start time
+ header - crash when attempting to execute log_msg call, but at least fairly easy to localize in the code
+ struct - same as header, even if all fields of the struct are of type `bit<W>`

I did not try header_union, header stack, or anything more 'exotic' than header or struct, but would expect they have the same behavior as header/struct.

One enhancement possibility - Change p4c bmv2 back end to give an error if an unsupported type is used as argument to log_msg calls.

Whatever list of supported/unsupported types we decide upon, I am happy to document, either in the v1model.p4 #include file, or here, whichever seems most appropriate: https://github.com/p4lang/behavioral-model/blob/master/docs/simple_switch.md

jafingerhut:
Actually attaching example program I mentioned in comment above.
[log-msg-supported-types.p4.txt](https://github.com/p4lang/p4c/files/4221413/log-msg-supported-types.p4.txt)

jafingerhut:
Confirmed that this PR fixed this issue.  Closing this issue.

PR #2462, opened 2020-07-01
Validate arguments to log_msg

mbudiu-vmw:


mbudiu-vmw:
Fixes #2201

backends/bmv2/simple_switch/simpleSwitch.cpp
```diff
@@ -810,7 +810,25 @@ Util::IJson* ExternConverter_log_msg::convertExternFunction(
...
     params->append(str);
     if (mc->arguments->size() == 2) {
-        auto val = ctxt->conv->convert(mc->arguments->at(1)->expression);
+        auto arg1 = mc->arguments->at(1)->expression;
+        // this must be a list expression, with all components
+        // evaluating to integral types.
+        auto argType = ctxt->typeMap->getType(arg1);
+        if (auto ts = argType->to<IR::Type_List>()) {
+            for (auto tf : ts->components) {
+                if (!tf->is<IR::Type_Bits>()) {
```
antoninbas, line 819: does it still work with SerEnum arguments?

backends/bmv2/simple_switch/simpleSwitch.cpp
```diff
@@ -810,7 +810,25 @@ Util::IJson* ExternConverter_log_msg::convertExternFunction(
...
+                    return primitive;
+                }
+            }
+        } else {
+            ::error(ErrorType::ERR_UNSUPPORTED_ON_TARGET,
+                    "%1%: Second argument must be a list expression: %2%", mc, arg1);
+            return primitive;
+        }
+
+        auto val = ctxt->conv->convert(arg1);
```
antoninbas, line 831: can you call `convert` (https://github.com/p4lang/p4c/blob/master/backends/bmv2/common/expression.h#L105) with `convertBool` set to `true`? Then it will work fine with boolean values as well.

backends/bmv2/simple_switch/simpleSwitch.cpp
mbudiu-vmw: Line 134 from issue2201-bmv2.p4 tests for this case

backends/bmv2/simple_switch/simpleSwitch.cpp
mbudiu-vmw: Did that, but since this is a list I had to iterate over the elements first.

backends/bmv2/simple_switch/simpleSwitch.cpp
antoninbas: Since the testdata/p4_16_samples/issue2201-bmv2.p4 is marked as XFAIL, I don't think anything is really tested as far as these backend changes are concerned? But if `tf->is<IR::Type_Bits>()` is true for SerEnum types, that works for me.

mbudiu-vmw: I added a new test 2201-1-bmv2.p4 which tests this case.

PR #2463, opened 2020-07-01
Support regex matching for toP4 dumps.

fruffy:
Currently it is not possible to perform more complex queries for toP4 compiler pass dumps. We can only check for substrings. This pull request enables matches using regular expressions. Functionality should remain the same, but now it is possible to use commands like `p4c/build/p4test --dump dmp prog.p4 --top4 "FrontEnd.*PassRepeated"` which only dumps PassRepeated in the FrontEnd. Or `p4c/build/p4test --dump dmp prog.p4 --top4 ".*"` which dumps all compiler passes. 

Afaik C++ regex matching is supported after `gcc4.9`, which is the minimum version `p4c` requires.

### DIFF
--- a/frontends/common/options.cpp
+++ b/frontends/common/options.cpp
@@ -396,14 +396,14 @@ void CompilerOptions::dumpPass(const char* manager, unsigned seq, const char* pa
     for (auto s : top4) {
         bool match = false;
         try {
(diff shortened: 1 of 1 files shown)


mbudiu-vmw:
@ChrisDodd : any objection to merging this?

PR #2466, opened 2020-07-01
Implementation of advance() for parser in uBPF backend

tatry:
I've implemented `advance()` in the uBPF parser. This allows to skip some bits without need to define headers.

In this implementation parser advances the packet cursor. When the packet is too short, it goes to reject state.

### DIFF
--- a/backends/ubpf/ubpfParser.cpp
+++ b/backends/ubpf/ubpfParser.cpp
@@ -341,22 +341,23 @@ void
 UBPFStateTranslationVisitor::compileAdvance(const IR::Expression* expr) {
     auto type = state->parser->typeMap->getType(expr);
     auto etype = UBPFTypeFactory::instance->create(type);
-    const char * tmpVarName = "__tmp_9507";  // need to be some random name
+    cstring tmpVarName = refMap->newName("tmp");
 
     builder->emitIndent();
     builder->blockStart();
     // declare temp var
     builder->emitIndent();
     etype->emit(builder);
-    builder->appendFormat(" %s = ", tmpVarName);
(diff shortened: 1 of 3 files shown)


backends/ubpf/ubpfParser.cpp
```diff
@@ -335,6 +337,31 @@ UBPFStateTranslationVisitor::compileLookahead(const IR::Expression* destination)
     builder->endOfStatement(true);
 }
 
+void
+UBPFStateTranslationVisitor::compileAdvance(const IR::Expression* expr) {
+    auto type = state->parser->typeMap->getType(expr);
+    auto etype = UBPFTypeFactory::instance->create(type);
+    const char * tmpVarName = "__tmp_9507";  // need to be some random name
```
mbudiu-vmw, line 344: Use the refMap, it can generate fresh unique names.

backends/ubpf/ubpfParser.cpp
```diff
@@ -367,6 +394,16 @@ bool UBPFStateTranslationVisitor::preorder(const IR::MethodCallExpression* expre
                 compileExtract(expression->arguments->at(0)->expression);
                 return false;
             }
+
+            if (extMethod->method->name.name == p4lib.packetIn.advance.name) {
+                if (expression->arguments->size() != 1) {
+                    ::error("Exactly one argument expected %1%", expression);
```
mbudiu-vmw, line 400: Use the version of error with an additional argument of type ErrType. Or perhaps call modelError if there is such a function.

tatry:
I think it would be better if @osinstom will review the code.

osinstom:
LGTM, I think we can merge.

PR #2467, opened 2020-07-02
Illegal P4 to report error rather than  compiler-error

boblytton:
Hi,
This is a tweak to the way an error is reported from:
```
In file: /home/rlytton/repos/p4c/ir/ir.def:479
Compiler Bug: ../testdata/p4_16_errors/switch_expression.p4(3): 85: Unexpected expression for switch case
const bit<8> NumConstant = 0x55;
                           ^^^^
```
to:
```
switch_expression.p4(3): [--Werror=expr] error: 0x55: Unexpected expression for switch case
const bit<8> NumConstant = 0x55;
                           ^^^^
```

### DIFF
--- a/ir/ir.def
+++ b/ir/ir.def
@@ -476,7 +476,7 @@ class SwitchCase {
     validate{
         if (!(label->is<IR::DefaultExpression>() || label->is<IR::PathExpression>()))
             ::error(ErrorType::ERR_EXPRESSION,
-                    "%1%: Unexpected expression for switch case", label);
+                    "%1%: Unexpected expression for switch case", *this);
         BUG_CHECK(statement == nullptr || statement->is<IR::BlockStatement>(),
                   "%1%: Expected a block statement",
                   statement);
--- a/ir/ir.def
+++ b/ir/ir.def
@@ -474,7 +474,7 @@ class SwitchCase {
     NullOK Statement statement;  // If missing then it's a fall-through
 #nodbprint
     validate{
-        if (label->is<IR::DefaultExpression>() || label->is<IR::PathExpression>())
+        if (!(label->is<IR::DefaultExpression>() || label->is<IR::PathExpression>()))
             ::error(ErrorType::ERR_EXPRESSION,
                     "%1%: Unexpected expression for switch case", label);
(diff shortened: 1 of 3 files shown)


boblytton:
I have add a `testdata/p4_16_errors/switch_expression.p4`, but as it stands it adds little additional value.
Q: should I remove it, or can I regex the expected output?

I have left the second conditional as a BUG_CHECK()
Q: should this also be an ::error()? and if so, how do I exercise it?

hesingh:
If you add a new test P4 program to p4_16_errors, run "cmake .." from `p4c/build`.   After cmake a new .test exists in the path to run the test.  From `p4c/build`, run `err/testdata/p4_16_errors/switch_expression.p4.test -f`.  If the test has no issues, output files will be generated.  

If you are working with p4c, please go through the `p4c/doc/compiler-design.pptx`.  See slide 130.

boblytton:
Hi @hesingh , the output file` ../testdata/p4_16_errors_outputs/switch_expression.p4-stderr` has been created and has the expected contents.
1. Is there a test that checks its output with this file?
2. Does this need to be committed too?

hesingh:
> Hi @hesingh , the output file` ../testdata/p4_16_errors_outputs/switch_expression.p4-stderr` has been created and has the expected contents.
> 
> 1. Is there a test that checks its output with this file?

Yes.  See slide 129.  `make check` runs all tests and matches expected output. 

> 2. Does this need to be committed too?

Yes.  See p4/testdata/p4_16_errors_outputs/ where lot of `-stderr` outputs exist.

> 
> (I also need to look at other tests that are now failing)

Debug any failing test by running the individual test.  I have shown in my earlier post how to run one test.  To debug run the test with "-v".  See slide 130.

boblytton:
Thank you - excellent.

boblytton:
I could use *this for the whole switch case:
```
switch_expression.p4(13): [--Werror=expr] error: SwitchCase: Unexpected expression for switch case
            NumConstant: { }
            ^^^^^^^^^^^^^^^^
```
Is there an alternative?
How about using `getSourceInfo() `to see if `label` is on the same line and if not, use `this`?

mbudiu-vmw:
the produced output will change, so the test will fail

PR #2469, opened 2020-07-13
let @name annotation states merge

AashleshaPatil-BFN:


liujed-intel:
Suppose we have
```
@name("one")   state s1 { ... ; transition s2; }
               state s2 { ... ; transition s3; }
@name("three") state s3 { ... ; transition s4; }
@name("four")  state s4 { ... ; transition accept; }
```
If `s1` and `s2` are merged, I'd naïvely expect the result
```
@name("one") state merged { ... ; transition s3; }
```
Similarly, if `s2` and `s3` are merged, we might expect the result
```
@name("three") state merged { ... ; transition s4; }
```
A few questions.
1. Are these expectations reasonable? What are the consequences of putting a `name` annotation on a parser state?
2. Do these expectations match the implementation here? Please provide test cases to demonstrate this.
3. What should we expect when `s3` and `s4` are merged?

AashleshaPatil-BFN:
> Suppose we have
> 
> ```
> @name("one")   state s1 { ... ; transition s2; }
>                state s2 { ... ; transition s3; }
> @name("three") state s3 { ... ; transition s4; }
> @name("four")  state s4 { ... ; transition accept; }
> ```
> 
> If `s1` and `s2` are merged, I'd naïvely expect the result
> 
> ```
> @name("one") state merged { ... ; transition s3; }
> ```
> 
> Similarly, if `s2` and `s3` are merged, we might expect the result
> 
> ```
> @name("three") state merged { ... ; transition s4; }
> ```
> 
> A few questions.
> 
> 1. Are these expectations reasonable? What are the consequences of putting a `name` annotation on a parser state?
> 2. Do these expectations match the implementation here? Please provide test cases to demonstrate this.
> 3. What should we expect when `s3` and `s4` are merged?

In this example `s1` and `s2` will merge. The merging algorithm is such that `s1` will swallow `s2` and the name of the state will be `s1` . I think if `@name` annotation is given, the contents in that annoted state should be associated with that state only. Thats why `s3 ` and `s4` will not merge and neither will `s2` and `s3`.  I will add a test.

liujed-intel:
Why allow the merging of `s1` and `s2` but not the merging of `s2` and `s3`?

AashleshaPatil-BFN:
> Why allow the merging of `s1` and `s2` but not the merging of `s2` and `s3`? I'm not seeing the difference in those two cases.

That will need a change in the algorithm of merging. It seems in the current algorithm, a state is absorbed by its preceding state. eg `s1`  absorbs `s2`. But in `s2` and `s3` , `s3` will have to absorb `s2`.  Though that should happen, currently my intention is to alleviate the restriction I added in my last PR.

Issue #2330, opened 2020-04-21
Another issue with Predication


fruffy:
I apologize for being pesky regarding the Predication issue but I believe I have found another case which could lead to unexpected behaviour. This still happens even after the recent fixes.

In the Predication pass

```p4
control ingress(inout Headers h, inout Meta m, inout standard_metadata_t sm) {
    bool pointless_bool_0;
    bit<48> tmp;
    bool hasReturned;
    bit<16> val_0;
    @name("ingress.do_action") action do_action() {
        hasReturned = false;
        pointless_bool_0 = true;
        if (h.eth_hdr.dst_addr != 48w0) {
            ;
        } else {
            hasReturned = true;
        }
        if (!hasReturned) {
            {
                val_0 = h.eth_hdr.eth_type;
                h.eth_hdr.eth_type = val_0;
            }
            if (pointless_bool_0) {
                tmp = 48w1;
            } else {
                tmp = 48w2;
            }
            h.eth_hdr.src_addr = tmp;
        }
    }
    apply {
        do_action();
    }
}
```

is turned into

```p4
control ingress(inout Headers h, inout Meta m, inout standard_metadata_t sm) {
    bool pointless_bool_0;
    bit<48> tmp;
    bool hasReturned;
    bit<16> val_0;
    @name("ingress.do_action") action do_action() {
        hasReturned = false;
        pointless_bool_0 = true;
        {
            bool cond;
            cond = h.eth_hdr.dst_addr != 48w0;
            hasReturned = (cond ? hasReturned : true);
        }
        {
            bool cond_0;
            cond_0 = !hasReturned;
            {
                {
                    // this expression may be undefined
                    h.eth_hdr.eth_type = (cond_0 ? val_0 : h.eth_hdr.eth_type);
                    tmp = (cond_0 ? (pointless_bool_0 ? 48w1 : 48w2) : tmp);
                }
            }
            {
                val_0 = (cond_0 ? h.eth_hdr.eth_type : val_0);
            }
            h.eth_hdr.src_addr = (cond_0 ? tmp : h.eth_hdr.src_addr);
        }
    }
    apply {
        do_action();
    }
}
```

Which may leave `h.eth_hdr.eth_type` undefined in the case that the program does not return. This should never be the case judging from the source program.

Triggering this issue is fairly tricky, it requires certain arrangement of function calls and variables (attached).

[...]

PR #2471, opened 2020-07-16
Fix: Predication issue #2330

anasyrmia:
Changes made in preorder of assignment and if statements, in order to fix the loss of assignment when dependencies exist.
New Predication class members:
Unsigned depNestingLevel used to track how deep in nesting do dependencies go.
AssignmentStatement pointer dependencyAssignement, added to indicate that dependencies exist, in which case no assignment is lost.
cstring dependantName used to track the assignemnt statements that are dependants.
vector of cstrings statNames used for storing assignement statements in preorder of assignment statements.
map isAssignemntPushed controls if the liveAssignemnts were already pushed on rv block.

Changes made:

In preorder of Assignment statements:
    Filling the liveAssignemnts map depending on the current dependencies nesting level and checking if the current statement is a dependant statement.
    elseStatementName cstring used for creating a new name for else branches in orderedNames.

In preorder of If statements:
    Checking if there were dependencies before pushing anything else on rv block (if dependnecyAssig$
    If the bool value of corresponding assignment is false, in the isAssignmentPushed map, the assignment should be pushed on rv block. If the bool value is true, it means that assignment has already been pushed when dependencies were handled.

In visitBranch of ExpressionReplacer:
    Fixing the assignment of condition variable from the right conditions vector element.

### DIFF
--- a/midend/predication.cpp
+++ b/midend/predication.cpp
@@ -82,7 +82,7 @@ void Predication::ExpressionReplacer::emplaceExpression(IR::Mux * mux) {
 }
 
 void Predication::ExpressionReplacer::visitBranch(IR::Mux * mux, bool then) {
-    auto condition = conditions[conditions.size() - currentNestingLevel];
+    auto condition = conditions[conditions.size() - currentNestingLevel - 1];
     auto leftName = lvalue_name(statement->left);
     auto thenExprName = lvalue_name(mux->e1);
     auto elseExprName = lvalue_name(mux->e2);
@@ -134,8 +134,9 @@ const IR::Node* Predication::clone(const IR::AssignmentStatement* statement) {
 }
 
 const IR::Node* Predication::preorder(IR::AssignmentStatement* statement) {
-    if (!inside_action || ifNestingLevel == 0)
+    if (!inside_action || ifNestingLevel == 0) {
         return statement;
+    }
     const Context * ctxt = nullptr;
     std::vector<const IR::Expression*> conditions;
     while (auto ifs = findContext<IR::IfStatement>(ctxt)) {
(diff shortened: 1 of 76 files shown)


mbudiu-vmw:
@fruffy : please take a look.

midend/predication.cpp
```diff
@@ -151,24 +152,52 @@ const IR::Node* Predication::preorder(IR::AssignmentStatement* statement) {
...
             liveAssignments.erase(dependency);
+            depNestingLevel = ifNestingLevel;
+            dependantName = lvalue_name(statement->left);
         }
     }
     auto statementName = lvalue_name(statement->left);
+    statNames.push_back(statementName);
+    bool isStatementDependant = false;
+    auto depName = dependantName;
+    for_each(statNames.begin(), statNames.end(),
```
mbudiu-vmw, line 167: I think that a simple for loop would be clearer.

midend/predication.cpp
```diff
@@ -151,24 +152,52 @@ const IR::Node* Predication::preorder(IR::AssignmentStatement* statement) {
...
+                new IR::AssignmentStatement(statement->left, rightStatement);
+        } else {
+            cstring elseStatementsName = generator->newName("elseStatement");
+            orderedNames.push_back(elseStatementsName);
+            auto rightStatement = clone(statement->right)->apply(replacer);
+            liveAssignments[elseStatementsName] =
+                new IR::AssignmentStatement(statement->left, rightStatement);
+        }
+    } else {
+        if (foundedAssignment != liveAssignments.end()) {
```
mbudiu-vmw, line 188: foundedAssignment is not used in the previous block, why not move it in here?

midend/predication.cpp
```diff
@@ -134,8 +134,9 @@ const IR::Node* Predication::clone(const IR::AssignmentStatement* statement) {
 }
 
 const IR::Node* Predication::preorder(IR::AssignmentStatement* statement) {
-    if (!inside_action || ifNestingLevel == 0)
+    if (!inside_action || ifNestingLevel == 0) {
```
mbudiu-vmw, line 137: The field inside_action can be replaced by `findContext<IR::P4Action>() != null`. This will make the code easier to read and more robust too.

I would suggest commenting all data structures of the visitor as well, since your code structure has become quite complicated.

midend/predication.h
```diff
@@ -73,6 +73,10 @@ class Predication final : public Transform {
     std::vector<IR::BlockStatement*> blocks;
     bool inside_action;
     unsigned ifNestingLevel;
+    unsigned depNestingLevel;
```
mbudiu-vmw, line 76: Please add a comment for each of these fields.

fruffy:
Looks good. No errors in the test suite. I also generated a bunch of random programs and have not found any issues so far.

Issue #2472, opened 2020-07-20
Calculating TCP and UDP checksum in p4 switch


sarashakeri:
Hi,

I am using a P4 switch for implementing nat between two network namespaces and as I am changing the IP addresses I want to calculate l3 and l4 checksum. There is no problem with calculating l3, however, I did not find any example of calculating l4 checksum. Currently, I am using the v1model, but don't have any limitation for the p4 switch model. Is there any way to implement tcp or udp checksum in P4 switch?

hesingh:
First, the UDP checksum is defined in this RFC: https://tools.ietf.org/html/rfc768
UDP uses the same checksum at TCP.

Pseudo code for the checksum algorithm can be leveraged from the PSA specification.  

https://p4.org/p4-spec/docs/PSA.html#appendix-internetchecksum-implementation

sarashakeri:
Thanks for your reply. 
What I was looking for, is the already tested example of calculating TCP checksum especially in the V1model ( like what there is for IP header), but it seems there is not a complete and available code for that, right? So, I have to write the code?

hesingh:
I poked around (not hard enough) the p4lang/behavioral-model repo for simple_switch and do not see a checksum computation.  Also, since you are looking for a checksum computation, this issue does not belong to p4c, a compiler.  You should file an issue with p4lang/behavioral-model,

What the PSA specification has for checksum computation is correct and the code can be used.

mbudiu-vmw:
In general you don't compute the checksum, you incrementally update it, which should be much easier.

hesingh:
The update_checksum is a simple API you can apply for updating layer-4 checksum.  

https://github.com/p4lang/p4c/blob/master/p4include/v1model.p4#L465

There are several example uses in `p4c/testdata/p4_16_sample/*.p4` for update_checksum.  Even though, the example use IPv4 header, you can replace such data with layer-4 data.

sarashakeri:
I tried to add something like there is for IPv4 checksum update for TCP header as below, however it is not working. Is there any problem with the code:


`
//Calculating tcp length

    state parse_ipv4 {
        packet.extract(hd.ipv4);
        verify(hd.ipv4.version == 4w4, error.IPv4IncorrectVersion);
        verify(hd.ipv4.ihl == 4w5, error.IPv4OptionsNotSupported);
        meta.tcpLength = hd.ipv4.total_len - 20;
        transition accept;
    }

//Updating checksum

[...]

PR #2474, opened 2020-07-21
Add TCP and UDP Checksum Test

hesingh:
Seeing the recent [issue ](https://github.com/p4lang/p4c/issues/2472) filed against p4c, I decided to add layer-4 (TCP and UDP) checksum example or test to p4c.  The `expect` packet in STF was loaded in Wireshark and Wireshark verified the checksum as correct.


hesingh:
To test, copy the `expect` packet from `checksum-tcp-bmv2.stf` and load the packet in Scapy as follows:

```bash
>>> p = Raw("525400123502080027018BBC080045000030000100003F0665B30A00020F0A01000516A2005000000000000000005002200027890000DEADBEEFDEADBEEF")
```
Then dump the packet as follows in Scapy and save the output to a file.  The file is imported into Wireshark as hexdump to verify checksum.  In the Wireshark Import dialog box, for Encapsulation, select TCP.

```bash
>>> hexdump(p)
0000   35 32 35 34 30 30 31 32  33 35 30 32 30 38 30 30   5254001235020800
0010   32 37 30 31 38 42 42 43  30 38 30 30 34 35 30 30   27018BBC08004500
0020   30 30 33 30 30 30 30 31  30 30 30 30 33 46 30 36   0030000100003F06
0030   36 35 42 33 30 41 30 30  30 32 30 46 30 41 30 31   65B30A00020F0A01
0040   30 30 30 35 31 36 41 32  30 30 35 30 30 30 30 30   000516A200500000
0050   30 30 30 30 30 30 30 30  30 30 30 30 35 30 30 32   0000000000005002
0060   32 30 30 30 32 37 38 39  30 30 30 30 44 45 41 44   200027890000DEAD
0070   42 45 45 46 44 45 41 44  42 45 45 46               BEEFDEADBEEF
>>> 
```

hesingh:
@jafingerhut and anyone else, please review this PR - thanks.

hesingh:
I tried to add TCP options to the P4 program of this PR using @jafingerhut's P4 TCP options [code](https://github.com/jafingerhut/p4-guide/blob/master/tcp-options-parser/tcp-options-parser.p4).  Even p4c has similar parsing of TCP options in `issue2090.p4`.

The parser and emit in deparser work fine.  However, when I give the header union to verify or update checksum API, the simple_switch fails with the error below?

Does simple_switch support header_union in checksum verify/update API?

```
~/p4c/build$ ./p4c-bm2-ss --std p4-16 ../testdata/p4_16_samples/checksum-l4-bmv2.p4
~/p4c/build$ bmv2/testdata/p4_16_samples/checksum-l4-bmv2.p4.test -v
Writing temporary files into  ./tmpcld9gzmk
Executing  ./p4c-bm2-ss -o bmv2/testdata/p4_16_samples/checksum-l4-bmv2.json /home/hemant/p4c/testdata/p4_16_samples/checksum-l4-bmv2.p4
Exit code  0
Check for  /home/hemant/p4c/testdata/p4_16_samples/checksum-l4-bmv2.stf
Running model
Running /home/hemant/behavioral-model/targets/simple_switch/simple_switch --log-console --log-flush --use-files 0 --thrift-port 9795 --device-id 705 -i 0@pcap0 ../bmv2/testdata/p4_16_samples/checksum-l4-bmv2.json
Ignoring --log-flush option because --log-file not specified
Unsupported input type 'header_union_stack' for calculation 'calc_3'
bad json:
{
   "type" : "header_union_stack",
   "value" : "tcp_options_vec"
}

Removing ./tmpcld9gzmk
*** Test failed
~/p4c/build$ 
```

jafingerhut:
@hesingh That is most likely a limitation of p4lang/behavioral-model repository code, that it doesn't support a stack of header_union as an argument to the checksum functions.  It might not support header stacks as arguments to those functions, either.

hesingh:
Seems right, @jafingerhut.  

I am mulling over plan B which is to copy parsed options to varbit which is allowed in checksum.  I will also change your code to use minSizeInBits().  However, I am afraid using bit Slice in varbit to copy options to will run into compile-time constant issue with Slice index.  I did change p4c to support runtime-index in the bmv2 backend, but I think the Slice may need some p4c frontend changes.

hesingh:
Using runtime-index with Slice index, the Constant Folding pass barfs on this line of code.

https://github.com/p4lang/p4c/blob/master/frontends/common/constantFolding.cpp#L522

There is a typo in the error message too which should replace indexes to indices. 

@mbudiu-vmw What do you think about Slice with runtime-index?  Nightmare to change for p4c?  We could skip Costant Folding but I am not sure how many other frontend or midend passes will barf because if not, let the bmv2 backend deal with the index.  The bmv2 backend already supports runtime-index and so does simple_switch, but not with Slice index.

jafingerhut:
@hesingh In P4_16, types `bit<8>` and `bit<9>` are different, for example, and I believe that any expression of the form `foo[msb:lsb]` where the value of `msb-lsb` is not a constant cannot therefore have a single type.  Even if `msb-lsb` can be inferred to be a constant, e.g. `foo[x+7:x]` where `x` is a run-time variable, is not supported by p4c, even though one could imagine inferring that it had type `bit<8>` if `foo` was type `bit<W>`.

The best workarounds for doing something like that, that I know of, are an if/then/else-if daisy chain, or a table with as many actions as needed for all cases, where every branch or action has compile-time constant slice indexes.  Tedious, but easy to auto-generate using a Python program, similar to examples here: https://github.com/jafingerhut/p4-guide/blob/master/code-generation/README.md

hesingh:
@antoninbas  I am debugging TCP checksum discrepancy for another packet today.  I added printing `len` in 
`behavioral-model/src/bm_sim/calculations.cpp` as shown below.

```c++
struct cksum16 {
  uint16_t operator()(const char *buf, size_t len) const {
    std::cout << "Printing data: " << len << std::endl;
```

The hexump of packet is : 

```
52 54 00 12 35 02 08 00 27 01 8B BC 08 00 45 00 00 38 00 01 00 00 40 06 64 AF 0A 00 02 0F 0A 01 00 01 16 A1 00 50 00 00 00 00 00 00 00 00 50 02 20 00 47 A6 00 00 64 65 61 64 62 65 65 66 64 65 61 64 62 65 65 66"
```

For IPv4, the len printed is 18 when I expected 20.  Likewise the len printed for TCP should be 36 but it's 48?

I could use help with how is len computed by behavioral-model for IPv4 header and TCP checksum.  thanks.

hesingh:
@antoninbas Sorry, for verify TCP checksum the length in 48 is correct because I forgot the 12 byte pseudo header to add to the TCP header + payload. 

I will double-check tomorrow about IPv4 length as 18 instead of 20. 

I fixed the TCP and UDP checksum l4len mistake.  I had also asked @jafingerhut to test my code with STF packet in case I had made any mistake in STF packet prep.  Andy too found that the l4len computing was wrong.

testdata/p4_16_samples/checksum-l4-bmv2.p4
```diff
@@ -0,0 +1,313 @@
...
+
+control vc(inout headers hdr,
+           inout metadata meta)
+{
+    apply {
+        // There is code similar to this in Github repo p4lang/p4c in
+        // file testdata/p4_16_samples/flowlet_switching-bmv2.p4
+        // However in that file it is only for a fixed length IPv4
+        // header with no options.  When I try to do this, it gives an
+        // error for having a varbit<> element in the tuple.
```
jafingerhut, line 193: Is this last sentence of the comment correct?  I do not understand it, given that hdr.ipv4.options appears in the list given as input to verify_checksum below.

hesingh: I will remove all comments in next checkin.

testdata/p4_16_samples/checksum-l4-bmv2.p4
```diff
@@ -0,0 +1,468 @@
...
+    apply {
+        packet.emit(hdr.ethernet);
+        packet.emit(hdr.ipv4);
+        packet.emit(hdr.tcp);
+        packet.emit(hdr.tcp_opt_ss);
+        packet.emit(hdr.tcp_opt_s);
+        packet.emit(hdr.tcp_opt_sack);
+        packet.emit(hdr.tcp_opt_nop);
+        packet.emit(hdr.tcp_opt_end);
+        packet.emit(hdr.tcp_opt_padding);
```
jafingerhut, line 458: This code for TCP option handling is wrong in general.  The options need not appear in this order, thus if you want to emit the same packet you received, this will not do so unless the options happen to be received in this order.  Also, some options can appear more than once in the same packet, and this code does not handle that case, either.

I would recommend that rather than trying to handle the general case, which I think is maybe a bit of a stretch for P4 without full support for stacks of header_union, it might be best to simply use varbit for TCP options in this example/test program, with no parsing of which options appear within it.  Or, ignore TCP options completely.

hesingh: I wrote the TCP options code to prove verify and update checksums works for TCP options.  Parsing TCP options for all use cases and emitting options in a correct sequence was not a goal because I cannot use a header_union. 

The order is preserved and parsing same option multiple times if I setup data structs similar to parsing MPLS labels using a header stack.  However, I don't want to bother with this one either.

Let me ignore TCP options completely in this PR.

testdata/p4_16_samples/checksum-l4-bmv2.p4
```diff
@@ -0,0 +1,360 @@
...
+    bit<16> srcPort;
+    bit<16> dstPort;
+    bit<32> seqNo;
+    bit<32> ackNo;
+    // dataOffset in TCP hdr uses 4 bits but set to 8 bits below.
+    // If 4 bits are used for it, p4c-bm2-ss complains the header
+    // is not a multiple of 8 bits. In the lookahead we use bit Slice
+    // to get 4 bits.
+    bit<8>  dataOffset;
+}
```
jafingerhut, line 96: You could also use something like this, which would avoid need for slice after lookahead:
```
    bit<4> dataOffset;
    bit<4> dontCare;
```
since you really do not care what the contents of any of the fields are except `dataOffset` for this header that is only used for lookahead purposes.

testdata/p4_16_samples/checksum-l4-bmv2.p4
```diff
@@ -0,0 +1,360 @@
...
+        }
+    }
+    state parse_tcp {
+        // The 4-bit dataOffset field of the TCP base header is the number
+        // of 32-bit words in the entire TCP header.  It is an error
+        // for it to be less than 5.  There are only TCP options
+        // present if the value is at least 6.  The length of the TCP
+        // options alone, without the 20-byte base header, is thus ((4
+        // * ihl) - 20) bytes, or 8 times that many bits.
+        pkt.extract(hdr.tcp,
```
jafingerhut, line 170: `ihl` in comment above should be `dataOffset`

frontends/common/constantFolding.cpp
```diff
@@ -519,7 +519,7 @@ const IR::Node* DoConstantFolding::postorder(IR::Slice* e) {
     const IR::Expression* msb = getConstant(e->e1);
     const IR::Expression* lsb = getConstant(e->e2);
     if (msb == nullptr || lsb == nullptr) {
-        ::error(ErrorType::ERR_EXPECTED, "%1%: bit indexes must be compile-time constants", e);
+        ::error(ErrorType::ERR_EXPECTED, "%1%: bit indices must be compile-time constants", e);
```
jfingerh, line 522: @mbudiu-vmw I have checked over the rest of this PR quite carefully, and the rest of it only adds a test program with STF file that looks good to me.  I wanted to call your attention to this change in case you had any comment for/against it.  I believe Hemant considers this an English spelling fix, but if 'indexes' appears in a bunch of other places already, it might as well be a 'standard alternate spelling' as far as p4c is concerned.

PR #2482, opened 2020-07-31
Use type inference for constructor type arguments

### LINKED ISSUE #2480: Compiler does not infer type arguments for some constructors.
See https://github.com/p4lang/p4-spec/issues/884 for a reproduction.

mbudiu-vmw:
Fixes #2480 
We actually had all the code needed for functions, but we were not using it for constructors.
This fix mostly copies and moves some code around in the type-checker.


testdata/p4_16_samples/issue2480-1.p4
```diff
@@ -0,0 +1,49 @@
...
+
+control Ingress(inout headers_t hdr, inout meta_t meta)
+{
+    Hash<bit<32>>(algo=HashAlgorithm_t.CRC32)       hash1;
+    Hash<bit<32>>(poly=CRCPolynomial(coeff=16w0x107))  hash2;
+    LCGMatrix(8w4, 10w7,
+        "1000100
+         0100011
+         0010110
+         0001001") hamming;
```
ChrisDodd, line 41: Since this is (almost) a duplicate of specIssue884.p4 you recently added, it probably makes more sense just to modify that rather than running what is essentially the same test twice...

frontends/p4/typeChecking/typeChecker.cpp
```diff
@@ -867,30 +861,24 @@ TypeInference::checkExternConstructor(const IR::Node* errorPosition,
...
 
-        result->push_back(new IR::Argument(arg->srcInfo, arg->name, newArg));
-        setType(newArg, paramType);
-        if (newArg != arg->expression)
-            changes = true;
-    }
-    if (changes)
-        return result;
-    else
-        return arguments;
```
ChrisDodd, line 893: Looks like the code is no longer checking for "no changes" and returning the original call?

mbudiu-vmw: Yes, all changes will be made in the cts.convert call. There's nothing else left for us to track.

PR #2483, opened 2020-08-04
 Analyze copy-in before calling functions 

### LINKED ISSUE #2147: SimplifyDefUse incorrectly removes assignment in actions with slices as arguments
Because of copy-out semantics, the compiler removes assignments to global variables if they are also passed as an argument.
For example:

        action do_action(inout bit<8> val) {
            h.h.a = 8w7;
        }

        apply {
            do_action(h.h.a);
        }
    }

turns into

        @name("ingress.do_action") action do_action() {
        }
        apply {
            do_action();
        }
    }

However, when slices are passed as argument and the whole variable is modified inside the action, the statement is still removed. 

        action do_action(inout bit<7> val) {
            h.h.a[0:0] = 0;
        }
        apply {
            do_action(h.h.a[7:1]);
        }
    }
turns into 

        bit<7> val;
        @name("ingress.do_action") action do_action() {
            val = h.h.a[7:1];
            h.h.a[7:1] = val;
        }
        apply {
            do_action();
        }
    }
and then ultimately all statements are removed. This incorrectly removes the  h.h.a[0:0] = 0; assignment. 

I attached a sample program and the corresponding out I would have expected. Feel free to correct me if I am making the wrong assumptions here.

[slice_arg.stf.txt](https://github.com/p4lang/p4c/files/4098079/slice_arg.stf.txt)
[slice_arg.txt](https://github.com/p4lang/p4c/files/4098080/slice_arg.txt)

mbudiu-vmw:
Fixes #2147 
Previously we were analyzing the effects of arguments after symbolically calling the functions for def-use analysis. Copy-in effects have to be performed before calling the function.


mbudiu-vmw:
Note that there are 3 tests, which were produced by simplifying the problem more and more. I think that all are valuable, since the simpler tests make it much easier to debug.

Issue #2485, opened 2020-08-04
Range starting from zero: Can not shift by a negative value


fruffy:
Submitting some old concerns and questions I have collected.
I have the following parser:

```p4
parser p(packet_in packet, out Headers hdr) {
    state start {
        packet.extract(hdr.h);
        transition select(hdr.h.a) {
            0x0000 .. 0x0004: parse;
            default: accept;
        }
    }
    state parse {
        hdr.h.a = 1;
        transition accept;
    }

}
```
When I try to compile this with `p4test` I get the following error:
```
MidEnd_11_SimplifyParsers
MidEnd_12_StrengthReduction
MidEnd_13_EliminateTuples
MidEnd_14_SimplifyComparisons
MidEnd_15_CopyStructures
MidEnd_16_NestedStructs
MidEnd_17_SimplifySelectList
MidEnd_18_RemoveSelectBooleans
MidEnd_19_FlattenHeaders
MidEnd_20_FlattenInterfaceStructs
Can not shift by a negative value.
```
Is this correct? The error does not make sense to me. 

[negative_shift.p4.txt](https://github.com/p4lang/p4c/files/5021296/negative_shift.p4.txt)

mbudiu-vmw:
@hesingh : this bug is in your pass, ReplaceSelectRange.

hesingh:
I will take a look and get back - thanks.

PR #2489, opened 2020-08-04
Fix Range error found with Issue 2485

hesingh:
See https://github.com/p4lang/p4c/issues/2485 and this error from it:

`Can not shift by a negative value.`

This issue is caused by using ffs() from gmputil.h which returns `-1` when v is 0.

### DIFF
--- a/lib/gmputil.h
+++ b/lib/gmputil.h
@@ -84,7 +84,7 @@ static inline unsigned bitcount(big_int v) {
 }
 
 static inline int ffs(big_int v) {
-    if (v <= 0) return (v < 0) ? -1 : 0;
+    if (v <= 0) return -1;
     return boost::multiprecision::lsb(v);
 }
 
--- a/lib/gmputil.h
+++ b/lib/gmputil.h
@@ -84,7 +84,7 @@ static inline unsigned bitcount(big_int v) {
 }
 
 static inline int ffs(big_int v) {
-    if (v <= 0) return -1;
+    if (v <= 0) return (v < 0) ? -1 : 0;
     return boost::multiprecision::lsb(v);
 }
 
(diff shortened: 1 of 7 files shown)


lib/gmputil.h
```diff
@@ -84,7 +84,7 @@ static inline unsigned bitcount(big_int v) {
 }
 
 static inline int ffs(big_int v) {
-    if (v <= 0) return -1;
+    if (v <= 0) return (v < 0) ? -1 : 0;
```
mbudiu-vmw, line 87: we should probably not change the definition of ffs; if all bits are 0 returning 1 seems incorrect. Instead probably you should move this check to the function that calls ffs.

mbudiu-vmw:
Fixes #2485

testdata/p4_16_samples_outputs/issue2485-midend.p4
```diff
@@ -0,0 +1,34 @@
...
+}
+
+parser p(packet_in packet, out Headers hdr) {
+    state start {
+        packet.extract<H>(hdr.h);
+        transition select(hdr.h.a) {
+            16w0x0 &&& 16w0xffff: parse;
+            16w0x1 &&& 16w0xffff: parse;
+            16w0x2 &&& 16w0xfffe: parse;
+            16w0x4 &&& 16w0xffff: parse;
```
jafingerhut, line 21: This range can be represented using only 2 ternary value/masks, i.e. (16w0x0 &&& 16w0xfffc) and (16w0x4 &&& 16w0xffff).  It seems odd to represent it using more ternary value/masks than needed.

hesingh: This is an issue when the start of Range is zero.  I will have to mull over this one.  Right now, the exhaustive test you wrote in range-to-tcam-entries.c is fine with four entries.

PR #2497, opened 2020-08-11
Strength-reduce !!

### LINKED ISSUE #2495: p4c-bm2-ss: Compiler Bug: Could not convert to Json
The following program produces a compiler bug in the simple-switch back end. 
The output is 
```
In file: p4c/backends/bmv2/common/expression.cpp:97
[31mCompiler Bug[0m: empty_table_miss.p4.txt(37): dummy_table_0/dummy_table;: could not convert to Json
        if (dummy_table.apply().miss) {
```
This only seems to happen when using an empty table in combination with a `miss` expression and an if statement with an else clause. 

[empty_table_miss.p4.txt](https://github.com/p4lang/p4c/files/5061493/2495.p4.txt)

mbudiu-vmw:
Fixes #2495.
The reason this works is that our code generator only handles very limited expressions in `if` statements that contain a table apply `if (table.apply()...)`.  LocalCopyProp would create an expression `if (!!table.apply().hit)`. The expressions in `if` statements are simplified very early in the compilation, and there isn't a pass in the back-end that would handle more complex expressions - under the assumption that no other pass could make the expression more complex.

We can probably simplify this even further to an unconditional table application.

### DIFF
--- a/backends/bmv2/psa_switch/midend.cpp
+++ b/backends/bmv2/psa_switch/midend.cpp
@@ -147,6 +147,7 @@ PsaSwitchMidEnd::PsaSwitchMidEnd(CompilerOptions& options, std::ostream* outStre
             new P4::ConstantFolding(&refMap, &typeMap),
             new P4::LocalCopyPropagation(&refMap, &typeMap, nullptr, policy),
             new P4::ConstantFolding(&refMap, &typeMap),
+            new P4::StrengthReduction(&refMap, &typeMap),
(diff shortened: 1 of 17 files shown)


jnfoster:
Minor: should "Git status" be in the commit message?

mbudiu-vmw:
It's funny, I typed "strength reduce !!" in the commit message, and the shell has substituted the !!.

PR #2499, opened 2020-08-17
[psa] i2e cloning bmv2 stf tests

peteli3:
This PR is comprised of: psa-i2e-cloning-basic-bmv2.p4/stf and generated sample output files.

i2e cloning implementation:
https://github.com/p4lang/behavioral-model/pull/935


mbudiu-vmw:
So I guess we'll have to wait for the PR in bmv2 to be merged first. You should then trigger again the build to validate your changes.

testdata/p4_16_samples/psa-i2e-cloning-basic-bmv2.p4
```diff
@@ -0,0 +1,147 @@
...
+    action clone() {
+        ostd.clone = true;
+        ostd.clone_session_id = (CloneSessionId_t) (CloneSessionIdUint_t) 3;
+    }
+
+    apply {
+        clone();
+
+        if (hdr.ethernet.dstAddr == 0) {
+            // This action should overwrite the ostd.drop field that
```
jafingerhut, line 66: This comment is probably copied and pasted from another PSA test program, and doesn't appear to be relevant for this test program, since there is no send_to_port() call above this comment.

testdata/p4_16_samples/psa-i2e-cloning-basic-bmv2.stf
```diff
@@ -0,0 +1,19 @@
+
+mc_mgrp_create 3
+mc_node_create 0 6
+mc_node_create 0 7
+mc_node_create 0 8
+mc_node_associate 3 0
+mc_node_associate 3 1
+mc_node_associate 3 2
+
+packet 1 000000000002 000000000000 ffff
```
jafingerhut, line 10: I believe Antonin made similar comments on the corresponding changes you propose in the p4lang/behavioral-model PR to add cloning support for PSA, but the clone session ids should not directly use multicast group ids.  That is, multicast group id 3 and clone session id 3 refer to different configurable 'objects' in the switch, and clone session id 3 should not use the multicast group 3 you are using the 'mc_' commands above to configure.

It would be a good idea to add support for some or perhaps all of the 'mirroring_' commands in simple_switch_CLI, probably copied and pasted into psa_switch_CLI, that are used for configuring mirroring sessions, where mirroring is another name for cloning in the P4 world.  https://github.com/p4lang/behavioral-model/blob/master/docs/runtime_CLI.md#mirroring_add-mirroring_add_mc-mirroring_delete-mirroring_get

I can probably find time to help add support for those commands to the STF files, since I added the corresponding support for the 'mc_' commands in STF files a couple of years ago, and the changes for this should be very similar.  The addition of the 'mc_' command support in STF files was added with this commit: https://github.com/p4lang/p4c/commit/7a8fae

Looking at that commit, it adds support for the 'mirroring_add' command, too, but not 'mirroring_add_mc', which would be useful to add for STF testing purposes, and looks like it is probably a one-line change to the bmv2stf.py program.

peteli3: @jafingerhut
With the latest commit, I added support in pswitch_CLI and bmv2stf for those mirroring commands and fixed up this STF test to not mistake multicast group ids for clone session ids. Please review when you get a chance.

testdata/p4_16_samples/psa-i2e-cloning-basic-bmv2.p4
peteli3: Ah yep my bad, fixing this in next commit.

testdata/p4_16_samples/psa-i2e-cloning-basic-bmv2.p4
```diff
@@ -0,0 +1,146 @@
...
+        clone();
+
+        if (hdr.ethernet.dstAddr == (EthernetAddress) 9) {
+            // PROPERTY - dropped packets are still cloned
+            ingress_drop(ostd);
+        } else {
+            // GENERAL CASE - all other packets are sent to port number from packet
+            send_to_port(ostd, (PortId_t) (PortIdUint_t) hdr.ethernet.dstAddr);
+        }
+    }
```
jfingerh, line 72: @peteli3 One more suggested enhancement to this test -- during ingress control, do a modification to the packet, e.g. something like `hdr.ethernet.etherType = 0xcafe;`, so that the STF test can verify that this change _is_ made to the normal unicast packets that are output, but _is not_ made to the ingress-to-egress cloned packets.  Having a comment like that before the assignment explaining why you are doing it is good for later people reading and understanding the tests.

peteli3: Good idea, added this in my next commit.

testdata/p4_16_samples/psa-i2e-cloning-basic-bmv2.p4
```diff
@@ -0,0 +1,148 @@
...
+control cEgress(inout headers_t hdr,
+                inout metadata_t user_meta,
+                in    psa_egress_input_metadata_t  istd,
+                inout psa_egress_output_metadata_t ostd)
+{
+    apply {
+        // mark the clone packets so stf can tell the difference
+        if (istd.packet_path == PSA_PacketPath_t.CLONE_I2E) {
+            hdr.ethernet.srcAddr = 48w5;
+        }
```
jfingerh, line 100: @peteli3 Sorry for one more request on changes to this test program.  Right now this test program modifies hdr.ethernet.srcAddr on ingress, which should not affect clone packets, but it also assigns a value to the same field in egress, only for cloned packets, so from the output packet it isn't really possible to tell that the ingress assignment DIDN'T take effect, like it should not.

Could you change either the ingress or the egress assignment to a different field, e.g. hdr.ethernet.etherType?  That way whichever field you try to assign in ingress, that does not come out modified for cloned packets, we can know that it is not being modified at all by the P4 program, as it should not.

peteli3: That's a good point - thanks for the suggestion. I'll make this change.

peteli3: the next commit adds this suggested change

jafingerhut:
There is only one failing test, and it is one that I checked in, and I suspect I know the reason why.  I think it is because that test has some nondeterminism in the order of ingress/egress packet processing, and different input packets being processed, which leads to nondeterminism in the output packets.  I believe I know how to make it deterministic, and yet remain an STF test.

jafingerhut:
@peteli3 If you take the attached file, which is the same as the current file testdata/p4_16_samples/psa-end-of-ingress-test-bmv2.stf, except with some of the 'packet' and 'expect' lines commented out, I would hope that it will pass.  Please copy this modified file over the one in your repository, and push that as another commit on this PR, and I hope everything should pass then.

I could not get the STF test as currently written to fail on my local machine, even after about 15 consecutive runs, so I suspect there is something about the machine being used for p4c automated tests that more often gives a different BMv2 packet processing order than mine.  In any case, making it 100% predictable results, or as close as we know how, is a good idea, and this new file at least gets closer to predictable results, and hopefully deterministic.
[psa-end-of-ingress-test-bmv2.stf.txt](https://github.com/p4lang/p4c/files/5158553/psa-end-of-ingress-test-bmv2.stf.txt)

peteli3:
@jafingerhut @mbudiu-vmw
just wanted to bump this to get merged in asap since checks are passing now.

thanks for the review and suggestions!

Issue #1985, opened 2019-07-05
Compiler Bug in local copy propogation


jnfoster:
The `makeSideEffectStatement` (https://github.com/p4lang/p4c/blob/master/midend/local_copyprop.cpp#L37) is either incomplete or is called elsewhere in a way that violates its preconditions.

Here is a minimal P4 file that triggers the bug:
```p4
#include <core.p4>
#include <v1model.p4>
header h_t {
     bit<8> x;
}
struct headers {
    h_t h;
}
struct metadata { }
parser p(packet_in pkt,out headers hdr,inout metadata meta,inout standard_metadata_t std_meta) {
    state start {
        transition accept;
    }
}
control c1(inout headers hdr,inout metadata meta) {
    apply { }
}
control c2(inout headers hdr,inout metadata meta,inout standard_metadata_t std_meta) {
    apply { }
}
control c3(inout headers hdr,inout metadata meta,inout standard_metadata_t std_meta) {
    action a(in bool b) {
        hdr.h.x = 8w0;
    }
    table t {
        key = { }
        actions = {
            a(hdr.h.isValid() || (bool) 1w1);
        }
    }
    apply {
        t.apply();
    }
}
control c4(inout headers hdr,inout metadata meta) {
    apply { }
}
control c5(packet_out pkt,in headers hdr) {
    apply { }
}
V1Switch(p(),c1(),c2(),c3(),c4(),c5()) main;
```

When compiled using the command `p4c --std p4-16 --arch v1model makeSideEffectStatement-shrunk.p4` the compiler produces the following error:

```
In file: /mnt/p4c/midend/local_copyprop.cpp:42
Compiler Bug: Unhandled expression in makeSideEffectStatement
```

mbudiu-vmw:
@ChrisDodd : I think that this happens because your hasSideEffects pass assumes that a call to isValid has side effects.

fruffy:
This looks fixed now.

PR #2500, opened 2020-08-17
Fix compiler Bug in local copy propagation #1985

milicaperic:
Change made in makeSideEffectStatement where the case which processed binary operator was missing.
After that a new bug appeared because isValid has not yet been processed in convertActionBody.

Change made:

In makeSideEffectStatement:
-Processing of unary operation was replaced with processing of unary and binary
-The given example illustrates the situation where the change needs to be applied

~~~ P4
control c3(inout headers hdr,inout metadata meta,inout standard_metadata_t std_meta) {
    action a(in bool b) {
        hdr.h.x = 8w0;
    }
    table t {
        key = { }
        actions = {
            a(hdr.h.isValid() || (bool) 1w1);
        }
    }
    apply {
        t.apply();
    }
}
~~~

or more complicated:

~~~ P4
control c3(inout headers hdr,inout metadata meta,inout standard_metadata_t std_meta) {
    action a(in bool b) {
        hdr.h.x = 8w0;
    }
    table t {
        key = { }
        actions = {
            a(!(!hdr.h.isValid() || !hdr.g.isValid()) ||  !(!hdr.f.isValid() || !hdr.e.isValid()) ||  !hdr.i.isValid());
        }
    }
    apply {
        t.apply();
    }
}
~~~

In convertActionBody:
-Added case that handles isValid
-Set prim with "no_op"
-The way we use isValid allows us to ignore the validity as opposed to the if-statement where this property is crucial
as in example:

    action a(in bool b) {
    if(b)
        hdr.h.x = 8w0;
    }
    table t {
        key = { }
        actions = {
            a(!(!hdr.h.isValid() || !hdr.g.isValid()) ||  !(!hdr.f.isValid() || !hdr.e.isValid()) ||  !hdr.i.isValid());
        }
    }
    apply {
        t.apply();
    }

    where no needed for makeSideEffectStatement because local variable is live

### DIFF
--- a/midend/has_side_effects.h
+++ b/midend/has_side_effects.h
@@ -30,15 +30,26 @@ class hasSideEffects : public Inspector {
     bool result = false;
     bool preorder(const IR::AssignmentStatement *) override { result = true; return false; }
     bool preorder(const IR::MethodCallExpression *mc) override {
-        if (result) return false;
+        if (result) {
+            return false;
+        }
         /* assume has side effects if we can't look it up */
         if (refMap && typeMap) {
             auto *mi = P4::MethodInstance::resolve(mc, refMap, typeMap, true);
+            if (auto *bm = mi->to<P4::BuiltInMethod>()) {
+                if (bm->name == "isValid") {
+                    return true;
+                }
+            }
             if (auto *em = mi->to<P4::ExternMethod>()) {
                 if (em->method->getAnnotation(IR::Annotation::noSideEffectsAnnotation))
-                    return true; } }
+                    return true;
+                }
+            }
         result = true;
-        return false; }
+        return false;
+    }
+
     bool preorder(const IR::Primitive *) override { result = true; return false; }
     bool preorder(const IR::Expression *) override { return !result; }
 
--- a/testdata/p4_16_samples/issue1985.p4
+++ b/testdata/p4_16_samples/issue1985.p4
@@ -1,0 +1,41 @@
+#include <core.p4>
+#include <v1model.p4>
+header h_t {
+     bit<8> x;
+}
+struct headers {
+    h_t h;
+}
+struct metadata { }
+parser p(packet_in pkt,out headers hdr,inout metadata meta,inout standard_metadata_t std_meta) {
+    state start {
+        transition accept;
+    }
+}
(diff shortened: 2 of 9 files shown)


jafingerhut:
Should the test program that fails with the latest p4c code, without this fix, from issue #1985 be added to this PR, with expected test outputs?  It seems a bad idea to leave out an existing known test program, when there is one.

milicaperic:
> Should the test program that fails with the latest p4c code, without this fix, from issue #1985 be added to this PR, with expected test outputs? It seems a bad idea to leave out an existing known test program, when there is one.

You are right I added two tests for verifying solution

mbudiu-vmw:
@ChrisDodd : do you want to take a look at this?

ChrisDodd:
will do (was on vacation last week, so missed this).

midend/local_copyprop.cpp
```diff
@@ -35,10 +35,26 @@ static const IR::Expression *lvalue_out(const IR::Expression *exp) {
...
+    auto unary = exp->to<IR::Operation_Unary>();
+
+    while (binary || unary) {
+        if (unary) {
+            exp = unary->expr;
+        }
+        if (binary) {
+            if (hasSideEffects(binary->left)) {
+                exp = binary->left;
+            } else if (hasSideEffects(binary->right)) {
```
ChrisDodd, line 48: This is incorrect if both `->left` and `->right` have side effects (the side effects in `->right` will be lost).  Add a `BUG_CHECK(!hasSideEffects(binary->right)`... at line 47 at least.

ChrisDodd:
So my suggested "right" fix would be in `hasSideEffects::preorder(MethodCallExpression`, if the `mi` returned by `MethodInstance::resolve` is a `BuiltInMethod` with the name `"isValid"`, return true and don't set result.

milicaperic:
Thank you for suggestion @ChrisDodd , I've made requested changes.

milicaperic:
Is there anything more that I should change in order to merge this PR?

mbudiu-vmw:
We are waiting for another review from @ChrisDodd

ChrisDodd (approved):

PR #2504, opened 2020-08-18
Small updates to CMakeLists.txt for PSA test programs

jafingerhut:
Remove mention of a test program psa-ingress-input-meta-bmv2.p4, which
as far as I can tell was never added to the p4c repository.

Update some comments in backends/p4test/CMakeLists.txt to reflect the
reason why some tests are excluded.


jafingerhut:
Is it clear to anyone else what the cause of the build failures are here?  They seem unrelated to my changes, and maybe have something to do with Python packages failing to be installed?

hesingh:
Please update your repo with latest p4c code which includes a python-xdist change and see if your error is fixed.

PR #2505, opened 2020-08-19
Make PSA packet recirculation test more strict

jafingerhut:
It still does not exercise recirculation of packets with preserved
user-defined metadata, since that is not yet implemented in p4c for
PSA, but it does copy the value received for ingress_port by the
ingress control into the output packet, so the STF test can check
whether it has the expected value.


jafingerhut:
@hanw FYI

I expect that this modified P4 program and STF test will fail, until after this PR or something similarly effective is merged into the p4lang/behavioral-model repo.  https://github.com/p4lang/behavioral-model/pull/936

This modified test was used to catch the misbehavior in psa_switch.

Issue #2470, opened 2020-07-15
Inlining Controls and out parameters


fruffy:
This is something I have observed in an actual large P4 program. 
So I have the following snippet:

```p4
void inline_dummy(in bit<16> eth_type, out bit<8> dummy_val) {}
control sub_ctrl(in bit<16> eth_type, out bit<8> dummy_val){
    apply {
    }
}

control ingress(inout Headers h, inout Meta m, inout standard_metadata_t sm) {
    sub_ctrl() sub_c;

    apply {
        sub_c.apply(h.eth_hdr.eth_type, h.h.a);
        inline_dummy(h.eth_hdr.eth_type, h.h.b);

    }
}
```
After the inlining passes (Inline and InlineFunctions), this turns into:

```p4
control ingress(inout Headers h, inout Meta m, inout standard_metadata_t sm) {
    apply {
        {
        }
        {
            bit<8> dummy_val_0;
            h.h.b = dummy_val_0;
        }
    }
}
```
In this case, the semantics of the out parameter for the function call are preserved, but not for the control `apply()` statement. Which one is correct? Or are both okay? One could argue that not changing the input value at all is a subset of undefined behaviour. Still, this seems inconsistent.  

[control_nested_inline.p4.txt](https://github.com/p4lang/p4c/files/4925372/control_nested_inline.p4.txt)

mbudiu-vmw:
In either case you get an undefined result.
So this looks correct to me.

jafingerhut:
It definitely seems like this is a case where the compiler is transforming the `sub_c.apply()` call into a program that is "less undefined" or "fewer possible output states" than the input program, and if one believed that the compiler transformations should somehow preserve undefined behavior, the `sub_c.apply()` call should also assign an undefined value to `h.h.a`.

I do not currently have any strong opinions on whether compiler transformations should maximally preserve undefined behavior, or should be allowed to reduce the possible behaviors.

mbudiu-vmw:
Perhaps it's a bug, I will investigate further.

fruffy:
For what it's worth, I cannot think of a case where this could be harmful. What I found odd was the difference in behaviour between function/action inlining and control inlining.

mbudiu-vmw:
[...]

PR #2509, opened 2020-08-20
Make stricter PSA tests that verify packet_path and instance fields i…

jafingerhut:
…n egress

### DIFF
--- a/testdata/p4_16_samples/psa-multicast-basic-2-bmv2.p4
+++ b/testdata/p4_16_samples/psa-multicast-basic-2-bmv2.p4
@@ -48,6 +48,10 @@ control packet_path_to_int (in PSA_PacketPath_t packet_path,
                             out bit<32> ret)
 {
     apply {
+        // Unconditionally assign a value of 8 to ret, because if all
+        // assignments to ret are conditional, p4c gives warnings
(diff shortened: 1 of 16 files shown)


jafingerhut:
These tests fail without the changes proposed in this bmv2 PR, or something similarly effective: https://github.com/p4lang/behavioral-model/pull/937

The tests were written to be more comprehensive in their checking that PSA intrinsic metadata fields are assigned the correct values.

jafingerhut:
@fruffy Would you have any time to look at an equivalence violation found for one of the test programs modified by this PR?  I am not yet knowledgable enough of the program's output to determine what kind of problem it found.

fruffy:
I think it is related to this issue #2470. I have not implemented a fix because the resolution was unclear. Will update something tomorrow.

jafingerhut:
@fruffy Thanks for taking a look.  I changed the test program so that the control always assigned a value to the 'out' parameter 'ret', which eliminates a warning from p4c that it might be uninitialized, in hopes that it will also avoid this error from the equivalence checker.  If you want to keep a copy of the version of the test program that caused the error report, let me know.  It is the one after the 3rd of the commits on this PR.

PR #2517, opened 2020-08-24
More fixes from static analysis tool

ChrisDodd:


backends/bmv2/psa_switch/psaSwitch.cpp
```diff
@@ -1060,6 +1060,9 @@ void ExternConverter_ActionProfile::convertExternInstance(
     action_profile->emplace_non_null("source_info", eb->sourceInfoJsonObj());
 
     auto sz = eb->findParameterValue("size");
+    BUG_CHECK(sz, "%1%Invalid declaration of extern ActionProfile ctor: no size param",
```
mbudiu-vmw, line 1063: Should this be a bug check or an error?
This will crash the compiler when supplied with a mismatched model file.

ChrisDodd: IMO it should be an internal compiler error if the arch file is messed up or mismatched with the compiler, though that could perhaps happen with an installation problem.

PR #2519, opened 2020-08-25
Add tests to verify end-of-ingress behavior of a PSA implementation

jafingerhut:


jafingerhut:
@peteli3 @hanw @yunheL Pinging you for your information.  These test cases did not find any bugs in the latest p4lang/behavioral-model PSA implementation, so that is good news.  They do provide more test coverage than any existing PSA tests I know of that the end-of-ingress relative priority of the drop and resubmit flags vs. multicast_group being 0 or non-0 is implemented as stated in the PSA specification.

testdata/p4_16_samples/psa-end-of-ingress-test-bmv2.p4
```diff
@@ -0,0 +1,217 @@
...
+                inout metadata_t user_meta,
+                in    psa_egress_input_metadata_t  istd,
+                inout psa_egress_output_metadata_t ostd)
+{
+    Register<bit<16>, bit<8>>(256) egress_pkt_seen;
+
+    apply {
+        bit<8> idx = hdr.ethernet.etherType[7:0];
+        bit<16> cur_count = egress_pkt_seen.read(idx);
+        if (hdr.ethernet.etherType[15:8] == 0xc0) {
```
mbudiu-vmw, line 136: perhaps there should be some named constants for these values.
an interesting idea would be to have such named constants possibly exposed to the control-plane too.
Perhaps if a constant has a `@name` annotation then it's part of the P4Runtime API?

mbudiu-vmw: Another useful idea would be to build a mechanism to allow the use of such constants in STF files.

jafingerhut: I have no objections to such capabilities.

(a) The mechanisms you describe sound like reasonable and useful things.  For production systems, it is common to have common to have thing that are either the following, or functionally equivalent to it with appropriate build scripts for generating multiple files from common sources: #include files used by both P4 program and control plane software that define names with identical numeric values.

(b) this is a tiny test program, and I hope to keep it fairly tiny so that it is easier to understand for anyone who wants to understand it.  The mechanisms described in (a) above seem perhaps a bit overkill for a tiny test program.

Personally I think that enabling something like PTF tests for p4c, where one can mingle sending and expecting packets with explicit pauses, and control plane API operations that are anything the Thrift and/or P4Runtime API support, e.g. reading and verifying counter values, creating action profile and action selector groups, could enable not only what you describe, but enable additional kinds of tests that STF does not support today.  In particular, this test program uses a Register extern where I would have been happy to use a counter if I could have read it in the STF file and compare its value to an expected one in the test script.  I also came across an issue I worked around where STF tests blast packets in as fast as possible, and as far as I can tell there is no capability to say "wait until you've checked all expect packets out up to this time so far, and only when you've done that, send in the next packet".  I know that enabling automated PTF tests for p4c is a sizable chunk of work, and I'm not volunteering for that at this time :-)

mbudiu-vmw: If you think such features would be useful please file issues asking for them.
S stands for "simple".

jafingerhut: I do think they would be useful, but at some point providing something like PTF tests provides all of those features, and much more, e.g. PTF would enable loops for creating table entries and other control plane API requests, packets, etc).  Gradually adding more and more special things to STF tests seems like moving them farther and farther from the 'S', and still not enabling what PTF tests can do.

jafingerhut:
Going ahead and merging this one in, since tests are passing and there is one approval.  Mihai, some of the enhancements suggested in your comments are probably best preserved as issues on the P4Runtime API spec repo: https://github.com/p4lang/p4runtime

PR #2524, opened 2020-09-02
Rename ERROR and WARNING macros

### LINKED ISSUE #2523: p4c macros clash with user code
User code including `lib/log.h`, either directly or through something like `ir/ir.h`, will also pull in a bunch of p4c internal macros like `ERROR` or `LOGN` into the global namespace: https://github.com/p4lang/p4c/blob/ebde76c05d14edbea62add9675b973eb0e6fc56c/lib/log.h#L128

These are very common names and bound to create naming conflicts in the user application. E.g. glog also defines `ERROR`.

Stratum currently work around this by `#undefine`ing the offending macros directly after the p4c includes. This is error prone and repetitive.

Common practice for libraries suggests prefixing library / internal macros with `P4C_` to prevent these clashes: [libuv](https://github.com/libuv/libuv/blob/43c74a6ee92a8e646201eef1fd056799cdccbfde/include/uv/errno.h#L23), [abseil](https://github.com/abseil/abseil-cpp/search?q=define&unscoped_q=define), [Boost](https://www.boost.org/doc/libs/master/libs/config/doc/html/boost_config/boost_macro_reference.html)

Unless there is a lot of external code that depends on these, this change could be a simple search and replace job.

pudelkoM:
This PR prefixes the `ERROR` and `WARNING` macro with `P4C_` to prevent collisions with other libraries (`windows.h`, glog).

After inspecting the code, I went for a minimal change here. There are no other uses of these inside p4c.

Alternatives:
If the code were better structured (i.e. would not include internal headers in public headers), this rename would not be necessary. This would require a larger restructure, though.

There is also this piece of code, which might cause surpises: https://github.com/p4lang/p4c/blob/5e05765e0c0d8d69e0ffe4ed4392e0e976203937/lib/log.h#L150-L166
I'm not sure if publicly overloading `ostream` for STL types is good practice.

Closes #2523

### DIFF
--- a/lib/log.h
+++ b/lib/log.h
@@ -1,5 +1,5 @@
 /*
-Copyright 2013-present Barefoot Networks, Inc. 
+Copyright 2013-present Barefoot Networks, Inc.
 
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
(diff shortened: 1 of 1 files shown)


mbudiu-vmw (approved):

mbudiu-vmw:
You have one test failures, which is rather unexpected.

pudelkoM:
> You have one test failures, which is rather unexpected.

Seems to be a flaky test.

jafingerhut:
Does anyone remember which test failed?  I ask mainly because I fixed one of my flaky tests recently, and was wondering if it might have been that one, or some other test I know about in any detail.

mbudiu-vmw:
It was a psa test.

jfingerh:
If that happens in the future, feel free to point me at it, preferably before the test results are overwritten by a following commit.  And/or copy and paste the failing test results into another issue labeled something like "PSA test with inconsistent failures" and I can take a look at it.

ChrisDodd:
We should probably just remove these macros completely as they are now unused (having removed all uses of them from the compiler)

PR #2526, opened 2020-09-04
Add a test case verifying that p4c gives compile time error for bad s…

jafingerhut:
…witch label


mbudiu-vmw (approved):

jafingerhut:
There is no bug in p4c I have found in this area.  It seems to correctly give a compile-time error if a switch statement mentions an action that is not an action for the table mentioned in the switch expression.  This PR simply adds an automated test case to confirm that this is caught as a compile-time error.

PR #2529, opened 2020-09-08
Rejet switch statements with two default labels

### LINKED ISSUE #2525: No compile-time error if a switch statement has multiple default labels
Regardless of one's opinion whether p4c should give a warning or an error if a default label is not last in a switch statement, it definitely seems like it ought to give an error if there is more than one default label in the same switch statement.  As of the version of p4c shown here, it does not give an error for the attached test program:
```
commit e3d00878698f7acdf57fba13260e086eef7aac1f (HEAD -> master, origin/master, origin/HEAD)
Author: Peter J. Li <pl488@cornell.edu>
Date:   Wed Sep 2 15:34:04 2020 -0700
```

Note: p4c already does give an error if a switch statement has a duplicate label that is not 'default'.  Test case testdata/p4_16_errors/duplicate-label.p4 covers this already.

mbudiu-vmw:
Fixes #2525


jafingerhut (approved):
LGTM

PR #2530, opened 2020-09-08
Add test case attempting to name an action 'default'

jafingerhut:
It gives compile-time syntax error, which seems better than allowing
such a confusing name, given the special meaning of 'default' in
switch statements.


jafingerhut:
I copied and pasted most of the program from another similar test program that was copyright VMware, so just left the copyright as is since my changes were minimal.

mbudiu-vmw (approved):

PR #2531, opened 2020-09-08
Improve error handing and error messages

mbudiu-vmw:


fruffy:
Just saw this. I can remove the warning so the function can be removed. But what is a suitable replacement used by p4c?

mbudiu-vmw:
Use `warn` with an additional ErrorType argument instead.

fruffy:
Should be fixed now:
https://github.com/p4gauntlet/toz3/blob/master/codegen.cpp#L1081

mbudiu-vmw:
Can you check that you got all of them by removing the function `warning` without this argument from the lib/error.h?

fruffy:
Yeah everything looks good.

backends/bmv2/psa_switch/psaSwitch.h
```diff
@@ -187,6 +187,11 @@ class ParsePsaArchitecture : public Inspector {
     explicit ParsePsaArchitecture(PsaProgramStructure* structure) :
         structure(structure) { CHECK_NULL(structure); }
 
+    void modelError(const char* format, const IR::INode* node) {
+        ::error(format, node->getNode());
```
ChrisDodd, line 191: should perhaps be `ErrorType::ERR_INVALID`?  Or perhaps a new ERR_ARCH_INTERNAL or some such

control-plane/bytestrings.cpp
```diff
@@ -28,7 +28,7 @@ namespace ControlPlaneAPI {
 boost::optional<std::string> stringReprConstant(big_int value, int width) {
     // TODO(antonin): support negative values
     if (value < 0) {
-        ::error("%1%: Negative values not supported yet", value);
+        ::error(ErrorType::ERR_INVALID, "%1%: Negative values not supported yet", value);
```
ChrisDodd, line 31: probably should be ERR_UNSUPPORTED

ChrisDodd (approved):

mbudiu-vmw:
@osinstom : there are a bunch of calls to `::error()` in your code that don't have an error code; when you have some time please convert these as well.

PR #2533, opened 2020-09-09
Avoid linking errors.

smolkaj:
This PR does not make semantic changes; it refactors code that caused nasty linking errors on some platforms.
Removing the relative indexing fixed the issue.


mbudiu-vmw (approved):

smolkaj:
Thanks for approving @mbudiu-vmw, I think this is ready to be merged then?

PR #2535, opened 2020-09-09
[PSA] add support for extern psa-internetChecksum

yunheL:
Testing filename: testdata/p4_16_samples/psa-internetChecksum-bmv2.p4 (.stf)

A corresponding PR has been submitted to bmv2 as well: [please see here](https://github.com/p4lang/behavioral-model/pull/945).

TODO:
1. Currently `internetChecksum.add()` can only take 1 field instead of a list of fields. For example:

```
InternetChecksum ck;
ck.add(fieldA); // this is supported
ck.add({fieldA, fieldB}) // this is not supported yet
```

I have not found an elegant way to support this. I am not entirely sure whether this should be handled on the p4c end or the bmv2 end. Advices and ideas are greatly appreciated!

2. Currently, `internetChecksum` can only be used inside a `control block` but not a `parser block`. I am looking at how to enable `internetChecksum` to be used in a parser block. Any pointers would be greatly appreciated! Thanks :-)


backends/bmv2/psa_switch/psaSwitch.cpp
```diff
@@ -578,7 +578,65 @@ Util::IJson* ExternConverter_InternetChecksum::convertExternObject(
...
+    } else if (em->method->name == "set_state" && mc->arguments->size() != 1) {
+        modelError("Expected 1 arguments for %1%", mc);
+        return nullptr;
+    } else if (em->method->name == "subtract" && mc->arguments->size() != 1) {
+        modelError("Expected 1 arguments for %1%", mc);
+        return nullptr;
+    } else if (em->method->name == "add" && mc->arguments->size() != 1) {
+        modelError("Expected 1 arguments for %1%", mc);
+        return nullptr;
+    }
```
rst0git, line 608: The `return` statement in the `if` body will prevent fall through to subsequent cases. Thus, it is not necessary to use `else if`.

PR #2537, opened 2020-09-09
Fix compiler bug confusing physical equality with semantic equality.

smolkaj:
The removed code checked if two CompileTimeValues were equal by checking pointer equality.
This can cause errors such as the following:

    In file: third_party/p4lang_p4c/ir/ir.cpp:210
    Compiler Bug: <redacted>: hdr.ipv4_base.isValid(); already set in <ControlBlock>(2443131) ingress(); instance type=control ingress<>(inout parsed_packet_t hdr, inout local_metadata_t local_metadata, inout standard_metadata_t standard_metadata); to hdr.ipv4_base.isValid();, not hdr.ipv4_base.isValid();
        if (hdr.ipv4_base.isValid()) {

Unfortunately I cannot provide a small example provoking the error at this time.

### DIFF
--- a/ir/ir.cpp
+++ b/ir/ir.cpp
@@ -205,7 +205,7 @@ void Block::setValue(const Node* node, const CompileTimeValue* value) {
     CHECK_NULL(node);
     auto it = constantValue.find(node);
     if (it != constantValue.end())
-        BUG_CHECK(value == constantValue[node],
+        BUG_CHECK(value->equiv(*constantValue[node]),
                       "%1% already set in %2% to %3%, not %4%",
                   node, this, value, constantValue[node]);
     else
--- a/ir/base.def
+++ b/ir/base.def
@@ -17,7 +17,11 @@
 #end
 
 /// a value that can be evaluated at compile-time
-interface CompileTimeValue {}
+interface CompileTimeValue {
+    bool equiv(const CompileTimeValue& other) const {
+        return this->getNode()->equiv(*other.getNode());
+    }
+}
 
 /// Base class for P4 types
 abstract Type {


ir/ir.cpp
```diff
@@ -203,13 +203,7 @@ const Type_Method* Type_Table::getApplyMethodType() const
...
 void Block::setValue(const Node* node, const CompileTimeValue* value) {
     CHECK_NULL(node);
-    auto it = constantValue.find(node);
-    if (it != constantValue.end())
-        BUG_CHECK(value == constantValue[node],
-                      "%1% already set in %2% to %3%, not %4%",
-                  node, this, value, constantValue[node]);
-    else
-        constantValue[node] = value;
+    constantValue[node] = value;
```
mbudiu-vmw, line 206: how about using equals() instead?

smolkaj: Do you mean equiv? I tried, but IR::CompileTimeValue is a .def "interface" without any member functions, so you cannot check for equivalence as far as I know. I do not see an easy fix for this, seems like it is a deeper issue with how the IR is structured.

FYI, this is the error I get when trying equals:
```
third_party/p4lang_p4c/ir/ir.cpp:208:26: error: no member named 'equals' in 'IR::CompileTimeValue'
        BUG_CHECK(value->equals(*constantValue[node]),
                  ~~~~~  ^
```

mbudiu-vmw: This is the code generated for CompileTimeValue:
```
/// a value that can be evaluated at compile-time
class CompileTimeValue : public virtual INode {
 public:
#line 21 "/home/mbudiu/git/p4c/ir/base.def"
    bool equiv(IR::Node const & a_) const override;
#line 24 "/home/mbudiu/git/p4c/build/ir/ir-generated.h"
};
```
So it does have an equiv.

mbudiu-vmw: All interfaces really inherit from INode, so they have the INode operations.

mbudiu-vmw: This part of the IR I think is actually not that useful: the Block hierarchy and CompileTimeConstants looked like a good idea, but in fact are probably not. But they are part of the P4 spec, so we cannot remove them... and several passes depend on them. They were designed to make backend writing easier, but it looks like backends don't use them much.

smolkaj: Thanks for clarifying. Not sure what's going on there, but on my machine I get
```
class CompileTimeValue : public virtual INode {
};
```

mbudiu-vmw: That is weird. 
But you could also just call `getNode()->equiv(*other->getNode())`.

mbudiu-vmw: Or define this:
```
/// a value that can be evaluated at compile-time
interface CompileTimeValue {
    bool eq(const CompileTimeValue* a) const { return getNode()->equiv(*a->getNode()); }
}
```

Unfortunately it looks like you cannot just override equiv.

mbudiu-vmw: equiv does not guarantee semantic equality, but it should be better than pointer equality. It may in fact work just fine for CompileTimeValues, since their value should not depend on the context.

smolkaj: The `getNode()` trick works, thanks!
Do you have a preference between doing this inline vs defining `eq` in the IR?

mbudiu-vmw: Not really. Hopefully it will give the correct result, though.

smolkaj: Found a nice solution.

Regarding correctness, it at least fixes the issue that I ran into previously (sorry that I cannot provide an example), so hopefully this is strictly better than pointer equality?

mbudiu-vmw (approved):

smolkaj:
Ready to be merged.

PR #2539, opened 2020-09-10
Make p4_library Bazel rule more portable

smolkaj:
...making it work on platforms where cc may not be on the PATH.

### DIFF
--- a/bazel/p4_library.bzl
+++ b/bazel/p4_library.bzl
@@ -1,5 +1,7 @@
 """P4 compilation rule."""
 
+load("@bazel_tools//tools/cpp:toolchain_utils.bzl", "find_cpp_toolchain")
+
 def _p4_library_impl(ctx):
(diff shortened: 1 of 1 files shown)


antoninbas (approved):
This looks fine to me, but I feel like p4c should have a more generic way of accepting a preprocessor executable, that would also apply when not using Bazel.

I have run into issues in the past, both when using the p4c backend binaries directly: https://github.com/p4lang/p4c/blob/f79b21033e0e942710dc00739806a6f835e11438/frontends/common/options.cpp#L324-L328 and when using the p4c driver: https://github.com/p4lang/p4c/blob/f79b21033e0e942710dc00739806a6f835e11438/tools/driver/p4c_src/p4c.bmv2.cfg#L24

For the first case, it seems that this information could come from the build system. Not sure what to do about the second one. It's also a bit suboptimal that we have 2 code paths for preprocessing.

Issue #2541, opened 2020-09-11
IR loop detected with minor mod to ebpf_model.p4


hesingh:
Sorry, this issue needs a minor modification to the ebf_model.p4 to be reproduced.  I have dumped FrontEndLast and MidEndLast passes but I don't see any obvious loop in the dumps.

This is the modification to ebpf_model.p4

```p4
diff --git a/backends/ebpf/p4include/ebpf_model.p4 b/backends/ebpf/p4include/ebpf_model.p4
index f9b4e1ef73708e8a918255298e187d7ea41e77cc..4904627509f4a24487ad705110ea3f5e99d81c0a 100644
--- a/backends/ebpf/p4include/ebpf_model.p4
+++ b/backends/ebpf/p4include/ebpf_model.p4
@@ -63,10 +63,10 @@ extern hash_table {
 
 /* architectural model for EBPF packet filter target architecture */
 
-parser parse<H>(packet_in packet, out H headers);
+parser parse<H, M>(packet_in packet, out H headers, out M metadata);
 control filter<H>(inout H headers, out bool accept);
 
-package ebpfFilter<H>(parse<H> prs,
-                      filter<H> filt);
+package ebpfFilter<H, M>(parse<H, M> prs,
+                         filter<H> filt);

```

Please see this repo (https://gitlab.com/hemant_mnkcg/p4c-temp) for all code changes to p4c and switch_ebpf.p4 to cause the issue. 

```bash
$ ./p4c-ebpf -o tmp.c ../testdata/p4_16_samples/switch_ebpf.p4
In file: /tmp/p4c-temp/ir/visitor.cpp:280
Compiler Bug: IR loop detected

``` 

The relevant line of code that causes the IR loop is with this code snippet from ebpf_switch.p4

```p4
state start
    {
        p.extract(headers.ethernet);
        meta.l4port = 0;
        transition select(headers.ethernet.etherType)
        {
            16w0x800 : ip;
            default : reject;
        }
    }
```

Soon as ` meta.l4port = 0;` is removed from the above program, the IR loop detected problem is not seen.

hesingh:
For good measure, I did add metadata as a 3rd parameter to the filter control as well, thinking metadata is used later to the parser.  Still the error persists. This is why I didn't include the other changes to this issue.  However, if folks would like to see the filter control changes, they are included in the second commit to the repo.

[...]

PR #2547, opened 2020-09-14
Do not call super::visit

mbudiu-vmw:


mbudiu-vmw:
Fixes #2541

hesingh:
Should we make the change in upbf backend as well to keep the code consistent between the two backends?

mbudiu-vmw:
First you should test this and see if it works for you. Feel free to submit a PR against the ubpf back-end and ask for a review.

hesingh:
Ok, let me test right away, thanks.

mbudiu-vmw:
I think this can be tested without making any changes to the model, just by inserting some assignments statements in the parser.

hesingh:
I tested this fix and it works.

hesingh:
> I think this can be tested without making any changes to the model, just by inserting some assignments statements in the parser.

In a P4 program, I did declare a local variable at the top of the parser and set the variable equal to zero in a parser state.  The p4c ebpf backend ebpfParser.cpp preorder(const IR::AssignmentStatement*) is not hit.  This means a pass already removed the local variable in the parser.  However, since ebpf model has "headers" as a parameter to the parser, one could set a header field to a value in a parser state and test this PR's code change.  You could add such a test case.

ChrisDodd (approved):

PR #2550, opened 2020-09-16
uBPF target: add support for LPM match

osinstom:
Similarily to the eBPF backend, this PR introduces support for Longest-Prefix Match for the uBPF target.

Obviously, a uBPF target must implement `UBPF_MAP_TYPE_LPM_TRIE` type of the BPF map. In the upcoming commits, I'm going to prepare a specification of functionalities/APIs that a switch should support to work properly with `p4c-ubpf`.

### DIFF
--- a/backends/ubpf/ubpfRegister.cpp
+++ b/backends/ubpf/ubpfRegister.cpp
@@ -58,6 +58,31 @@ namespace UBPF {
         }
     }
 
+    void UBPFRegister::emitInstance(EBPF::CodeBuilder *builder) {
+        cstring keyTypeStr;
+        if (keyType != nullptr && keyType->is<IR::Type_Bits>()) {
+            auto tb = keyType->to<IR::Type_Bits>();
+            auto scalar = new UBPFScalarType(tb);
+            keyTypeStr = scalar->getAsString();
+        } else {
+            keyTypeStr = cstring("struct ") + keyTypeName.c_str();;
+        }
+
+        cstring valueTypeStr;
+        if (valueType != nullptr && valueType->is<IR::Type_Bits>()) {
+            auto tb = valueType->to<IR::Type_Bits>();
+            auto scalar = new UBPFScalarType(tb);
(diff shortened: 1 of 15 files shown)


backends/ubpf/ubpfRegister.cpp
```diff
@@ -58,6 +58,29 @@ namespace UBPF {
...
     }
 
+    void UBPFRegister::emitInstance(EBPF::CodeBuilder *builder) {
+        cstring keyTypeStr;
+        if (keyType != nullptr && keyType->is<IR::Type_Bits>()) {
+            auto tb = keyType->to<IR::Type_Bits>();
+            auto scalar = new UBPFScalarType(tb);
+            keyTypeStr = scalar->getAsString();
+        } else {
+            keyTypeStr = cstring("struct ") + keyTypeName.c_str();;
```
mbudiu-vmw, line 68: Is this always true? Maybe you want to code defensively and add a BUG_CHECK for the type?

backends/ubpf/ubpfTable.cpp
```diff
@@ -200,6 +153,53 @@ UBPFTable::UBPFTable(const UBPFProgram *program,
     actionList = table->container->getActionList();
 
     setTableSize(table);
+    setTableKind();
+}
+
+void UBPFTable::emitInstance(EBPF::CodeBuilder *builder) {
+    cstring keyTypeStr;
```
mbudiu-vmw, line 160: This code looks very much like UBPFRegister. It would be nice to have a helper function implement both.

mbudiu-vmw (approved):

backends/ubpf/ubpfRegister.cpp
osinstom: I've refactored the `emitInstance()` method and added some additional BUG_CHECKs.

backends/ubpf/ubpfTable.cpp
osinstom: Done.

Issue #43, opened 2016-05-06
Mutation through IndexedVector iterator will produce wrong results.


mbudiu-bfn:
The iterator of the IndexedVector can be used for reading; however, overwriting an element through the iterator will result in an inconsistent IndexedVector (the replace method can be used to achieve the desired effect).

PR #2552, opened 2020-09-17
Fix bugs related to StructureExpressions

### LINKED ISSUE #2543: Some problems with function calls in struct initialization
I think one of the recent merges has unearthed some problems with structure initializations and function calls within them.
The following two programs throw errors that do not seem correct. 

[return_hdr_struct.p4.txt](https://github.com/p4lang/p4c/files/5217665/return_hdr_struct.p4.txt)
```
MidEnd_10_HandleNoMatch
MidEnd_11_SimplifyParsers
MidEnd_12_StrengthReduction
bugs/crash/return_hdr_struct.p4(17): [--Werror=type-error] error: AssignmentStatement: Cannot unify tuple_0 with ethernet_t
    return { { 1, 1, produce_val() }};
    ^^^^^^
bugs/crash/return_hdr_struct.p4(2)
header ethernet_t {
       ^^^^^^^^^^
```

[struct_function_call.p4.txt](https://github.com/p4lang/p4c/files/5217668/struct_function_call.p4.txt)
```
FrontEnd_47_InlineFunctions
SetHeaders_0_TypeChecking
In file: p4c/frontends/p4/typeMap.cpp:95
[31mCompiler Bug[0m: Could not find type for <StructExpression>(7655)
```

mbudiu-vmw:
IndexedVector had a validation function, but it was never called. If it was, it would have caught one of these two bugs.
Fixes #2543

### DIFF
--- a/frontends/p4/sideEffects.cpp
+++ b/frontends/p4/sideEffects.cpp
@@ -160,7 +160,7 @@ const IR::Node* DoSimplifyExpressions::preorder(IR::StructExpression* expression
         auto path = addAssignment(expression->srcInfo, tmp, v->expression);
         typeMap->setType(path, t);
         v = new IR::NamedExpression(v->name, path);
-	vec.push_back(v);
+        vec.push_back(v);
     }
     expression->components = vec;
     prune();
--- a/frontends/p4/sideEffects.cpp
+++ b/frontends/p4/sideEffects.cpp
@@ -151,7 +151,6 @@ const IR::Node* DoSimplifyExpressions::preorder(IR::StructExpression* expression
         return expression;
     // allocate temporaries for all members in order.
     // this will handle cases like a = (S) { b, f(b) }, where f can mutate b.
-    IR::IndexedVector<IR::NamedExpression> vec;
     LOG3("Dismantling " << dbp(expression));
     for (auto &v : expression->components) {
         auto t = typeMap->getType(v->expression, true);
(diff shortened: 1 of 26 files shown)


frontends/p4/sideEffects.cpp
```diff
@@ -160,7 +160,9 @@ const IR::Node* DoSimplifyExpressions::preorder(IR::StructExpression* expression
         auto path = addAssignment(expression->srcInfo, tmp, v->expression);
         typeMap->setType(path, t);
         v = new IR::NamedExpression(v->name, path);
+        vec.push_back(v);
     }
+    expression->components = vec;
```
ChrisDodd, line 165: This seems pointless?  `v` is already a reference to the pointer in expression->components.  Simpler would be to just get rid of the unused `vec`

frontends/p4/structInitializers.cpp
```diff
@@ -89,6 +89,25 @@ const IR::Node* CreateStructInitializers::postorder(IR::AssignmentStatement* sta
...
+        return statement;
+
+    auto ftype = typeMap->getType(func);
+    BUG_CHECK(ftype->is<IR::Type_Method>(), "%1%: expected a method type for function", ftype);
+    auto mt = ftype->to<IR::Type_Method>();
+    auto returnType = mt->returnType;
+    CHECK_NULL(returnType);
+
+    auto init = convert(statement->expression, returnType);
+    if (init != statement->expression)
```
ChrisDodd, line 106: pointless `if` test?  If the pointers are equal, the assignment would have no effect, so why bother?

ChrisDodd (approved):

frontends/p4/sideEffects.cpp
mbudiu-vmw: Actually this does not work because of issue https://github.com/p4lang/p4c/issues/43

PR #2561, opened 2020-09-22
ir generator fixes

ChrisDodd:
- use qualified name for classes when the unqualified would be
  ambiguous/incorrect (multiple classes with the same name in different
  namespaces)
- Fix gc memory reporting in pass manager (pre C++20 undefined order of eval)

### DIFF
--- a/lib/n4.h
+++ b/lib/n4.h
@@ -40,8 +40,13 @@ class n4 {
         } else if (v < UINT64_C(9950000000)) {
             v = (v + UINT64_C(50000000))/UINT64_C(100000000);
             os << v/10 << '.' << v%10 << 'G';
-        } else {
+        } else if (v < UINT64_C(999500000000)) {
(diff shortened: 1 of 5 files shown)


lib/n4.h
```diff
@@ -40,8 +40,13 @@ class n4 {
         } else if (v < UINT64_C(9950000000)) {
             v = (v + UINT64_C(50000000))/UINT64_C(100000000);
             os << v/10 << '.' << v%10 << 'G';
-        } else {
+        } else if (v < UINT64_C(999500000000)) {
             os << std::setw(3) << (v + UINT64_C(500000000))/UINT64_C(1000000000) << 'G';
+        } else if (v < UINT64_C(9950000000000)) {
```
mbudiu-vmw, line 45: I have trouble counting these zeros, I hope you have tested this.

mbudiu-vmw (approved):

Issue #2345, opened 2020-04-27
Incorrect transformation in Predication pass


jafingerhut:
With this version of p4c and the attached program, the Predication pass reorders some assignment statements in such a way that the final state is different than the original program will produce.
```
$ p4c --version
p4c 1.2.0 (SHA: d86cafa6)
```
It is possible that the root cause is the same as one of the other p4c issues, but I wanted to file this one separately, in case it has its own distinct root cause.  For example, this issue may be related: https://github.com/p4lang/p4c/issues/2330

Note: I did not write the test program.  It came from here: https://github.com/khooi8913/p4-op-in-condition/blob/e64e22b0d5b9b2f69cebe4700db869decc09cfb8/src/p4-xor-test.p4#L119

It was reported by the program's author on the P4 language Slack channel.  I will see if they will provide permission to add it as a test case to p4c.  If that does not work out, it should not be difficult to create another test program that produces similar behavior, I would guess.

[p4-xor-test.p4.txt](https://github.com/p4lang/p4c/files/4542111/p4-xor-test.p4.txt)

khooi8913:
Hi @jafingerhut, thanks for filing the issue! 

Feel free to use my program as a test case for p4c. 
I have also signed the ONF CLA as well. :)

mbudiu-vmw:
@matijasyrmia I will assume you will take a look at this one too.
Let us know if you are unable. Unfortunately it looks like I cannot formally assign it to you.

matijasyrmia:
@mbudiu-vmw I will see what is happening here, but I think it is related to #2330.

fruffy:
I can still reproduce this issue. In addition there seems to be another case that is distinct from this one. It is really tricky. 

The following program

```p4
action dummy(inout Headers val) {}
control ingress(inout Headers h) {
    action simple_action() {
        if (h.eth_hdr.eth_type == 1) {
            return;
        }
        h.eth_hdr.src_addr = 48w1;
        // this serves as a barrier
        dummy(h);
    }

    apply {
        h.eth_hdr.src_addr = 48w2;
        h.eth_hdr.dst_addr = 48w2;
        h.eth_hdr.eth_type = 16w2;

        simple_action();
    }
}
```
is transformed into

```p4
control ingress(inout Headers h) {
    bool hasReturned;
    ethernet_t val_eth_hdr;
    @name("ingress.simple_action") action simple_action() {
        hasReturned = false;
        {
            bool cond;
[...]

PR #2564, opened 2020-09-23
Fix: Predication issue #2345

anasyrmia:
This PR fixes issue #2345.

The issue was the incorrect order of assignment statements with a function like:
	void dummy(inout Headers val) { }
	...
	action simple_action() {
        if (h.eth_hdr.eth_type == 1) {
            return;
        }
        h.eth_hdr.src_addr = 48w1;
        // this serves as a barrier
        dummy(h);
    }

Predication output now looks like the folowing code:
	@name("ingress.simple_action") action simple_action() {
		hasReturned = false;
		{
			bool cond;
			cond = h.eth_hdr.eth_type == 16w1;
			hasReturned = (cond ? true : hasReturned);
		}
		{
			bool cond_0;
			cond_0 = !hasReturned;
			h.eth_hdr.src_addr = (cond_0 ? 48w1 : h.eth_hdr.src_addr);
			val_eth_hdr = (cond_0 ? h.eth_hdr : val_eth_hdr);
			h.eth_hdr = (cond_0 ? val_eth_hdr : h.eth_hdr);
		}
	}

and before these changes were implemented these three lines were falsely aranged:

	{
		val_eth_hdr = (cond_0 ? h.eth_hdr : val_eth_hdr);
	}
	h.eth_hdr.src_addr = (cond_0 ? 48w1 : h.eth_hdr.src_addr);
	h.eth_hdr = (cond_0 ? val_eth_hdr : h.eth_hdr);

since h.eth_hdr.src_addr should be assigned to 1 bofore the function call.

The cause of the problem was the need to push a block with dependencies before liveAssignments. Which is now not an issue because all assignments are pushed on rv block from the same vector (liveAssigns), in preorder of if statements. In preorder of assignment statements, statements are pushed on liveAssigns in order they have on the input of predication pass.

The xor test was added since it first showed this problem. This test now also has a correct output. Also, some tests were added, which show adequate behavior when mixing nested if else-es and dependencies.

### DIFF
--- a/midend/predication.cpp
+++ b/midend/predication.cpp
@@ -150,52 +150,64 @@ const IR::Node* Predication::preorder(IR::AssignmentStatement* statement) {
     // print out dependencies
     for (auto dependency : dependencies) {
         if (liveAssignments.find(dependency) != liveAssignments.end()) {
-            // print out dependecy
-            blocks.back()->push_back(liveAssignments[dependency]);
-            dependencyAssignment = liveAssignments[dependency];
+            // Save statement's name if it is dependent
+            dependentNames.push_back(lvalue_name(statement->left));
             // remove from names to not duplicate
-            orderedNames.erase(dependency);
             liveAssignments.erase(dependency);
             depNestingLevel = ifNestingLevel;
-            dependantName = lvalue_name(statement->left);
         }
     }
     auto statementName = lvalue_name(statement->left);
-    statNames.push_back(statementName);
-    bool isStatementDependant = false;
-    for (auto it  = statNames.begin(); it != statNames.end(); it++) {
-        if (dependantName == *it) {
-            isStatementDependant = true;
-        }
+    // Set value to true in isStatementDependent map
+    // if the name of a current statement is the same as
+    // the name of any in dependentNames.
+    auto it = std::find(dependentNames.begin(), dependentNames.end(), statementName);
+    if (it != dependentNames.end()) {
+        isStatementDependent[statementName] = true;
+    } else {
+        isStatementDependent[statementName] = false;
     }
-    if (depNestingLevel < ifNestingLevel && isStatementDependant) {
+    // Push liveAssignments in liveAssigns in adequate order
+    // If current statement is dependent, it should be pushed on liveAssigns.
+    if (depNestingLevel < ifNestingLevel && isStatementDependent[statementName]) {
         statement->right = new IR::Mux(conditions.back(), statement->right, statement->left);
         if (travesalPath[ifNestingLevel - 1]) {
-            orderedNames.push_back(statementName);
+            // push back dependent statements in if branch
             auto rightStatement = clone(statement->right)->apply(replacer);
             liveAssignments[statementName] =
                 new IR::AssignmentStatement(statement->left, rightStatement);
+            liveAssigns.push_back(liveAssignments[statementName]);
         } else {
+            // push back dependent statements in else brach
             cstring elseStatementsName = generator->newName("elseStatement");
-            orderedNames.push_back(elseStatementsName);
             auto rightStatement = clone(statement->right)->apply(replacer);
             liveAssignments[elseStatementsName] =
                 new IR::AssignmentStatement(statement->left, rightStatement);
+            liveAssigns.push_back(liveAssignments[elseStatementsName]);
         }
     } else {
         auto foundedAssignment = liveAssignments.find(statementName);
         if (foundedAssignment != liveAssignments.end()) {
             statement->right = foundedAssignment->second->right;
-            // move the lvalue assignment to the back
-            orderedNames.erase(statementName);
+            // Remove statement for 'then' if there is an else branch
+            if (!travesalPath[ifNestingLevel - 1]) {
+                liveAssigns.erase(std::remove(liveAssigns.begin(), liveAssigns.end(),
+                    foundedAssignment->second), liveAssigns.end());
+            }
         } else if (!statement->right->is<IR::Mux>()) {
             auto clonedLeft = clone(statement->left);
             statement->right = new IR::Mux(conditions.back(), clonedLeft, clonedLeft);
         }
-        orderedNames.push_back(statementName);
+        // Remove statement for 'then' if there is an a statement
+        // with the same statement name in the else branch.
+        if (liveAssigns.size() > 0 && !isStatementDependent[statementName] &&
+            lvalue_name(liveAssigns.back()->left) == lvalue_name(statement->left)) {
+            liveAssigns.pop_back();
+        }
         auto rightStatement = clone(statement->right)->apply(replacer);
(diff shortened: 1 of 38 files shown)


midend/predication.cpp
```diff
@@ -150,52 +150,62 @@ const IR::Node* Predication::preorder(IR::AssignmentStatement* statement) {
...
-    statNames.push_back(statementName);
-    bool isStatementDependant = false;
-    for (auto it  = statNames.begin(); it != statNames.end(); it++) {
-        if (dependantName == *it) {
-            isStatementDependant = true;
+    // Set value to true in isStatementDependant map
+    // if the name of a current statement is the same as
+    // the name of any in dependantNames.
+    isStatementDependent[statementName] = false;
+    for (auto it : dependantNames) {
```
mbudiu-vmw, line 165: Could you use the `find` method instead?

midend/predication.cpp
```diff
@@ -250,18 +248,12 @@ const IR::Node* Predication::preorder(IR::IfStatement* statement) {
...
-    rv->push_back(blocks.back());
-    blocks.pop_back();
-    blocks.push_back(new IR::BlockStatement);
-    for (auto exprName : orderedNames) {
-        if (!isAssignmentPushed[liveAssignments[exprName]]) {
-            rv->push_back(liveAssignments[exprName]);
-        }
-        isAssignmentPushed[liveAssignments[exprName]] = false;
+    // Push assignments which are correctly aranged in liveAssigns vector on rv block.
+    for (auto it : liveAssigns) {
```
mbudiu-vmw, line 252: use vector::insert?

midend/predication.h
```diff
@@ -75,29 +75,27 @@ class Predication final : public Transform {
...
-    // the assignment has already been pushed when handeling dependencies.
-    // False of corresponding assignemnt means that
-    // the assignment should be pushed on rv block at that point.
-    std::map<const IR::AssignmentStatement *, bool> isAssignmentPushed;
+    // Vector of assignment statements which collects assignments from
+    // liveAssignments and dependencies in adequate order. In preorder
+    // of if statements assignments from liveAssigns are pushed on rv block.
+    std::vector<const IR::AssignmentStatement*> liveAssigns;
+    // Map that shows if the current statement is dependant.
+    // Bool value is true for dependant statements,
```
mbudiu-vmw, line 96: your spelling of dependa/ent is inconsistent

testdata/p4_16_samples_outputs/issue2345-1-midend.p4
```diff
@@ -0,0 +1,83 @@
...
+        pkt.extract<ethernet_t>(hdr.eth_hdr);
+        transition accept;
+    }
+}
+
+control ingress(inout Headers h, inout Meta m, inout standard_metadata_t sm) {
+    ethernet_t val1_eth_hdr;
+    ethernet_t val_eth_hdr;
+    @name("ingress.simple_action") action simple_action() {
+        h.eth_hdr.src_addr = (!(h.eth_hdr.eth_type == 16w1 ? true : false) ? 48w1 : h.eth_hdr.src_addr);
```
mbudiu-vmw, line 29: I have filed #2565 to optimize such a pattern.

mbudiu-vmw (approved):
I am approving this PR, although I have to confess I don't understand the logic of the code anymore, but we need to fix this bug.
This is a case where I am relying fundamentally on Gauntlet and the Z3 verifier to ensure correctness.
@fruffy : I assume you will torture this logic further; your tool found the bug fixed here previously.

fruffy:
Gauntlet does not seem to have any issues with the tests. I will fuzz test until tomorrow and report back.

mbudiu-vmw:
This test failed due to a network connectivity issue.

fruffy:
Fuzz testing revealed an issue with header stacks:

The predication pass collapses 

```p4
    @name("ingress.simple_action") action simple_action() {
        if (h.eth_hdr.src_addr != h.eth_hdr.dst_addr) {
            h.h[0].a = 8w2;
        } else {
            h.h[1].a = 8w1;
        }
    }
    apply {
        simple_action();
    }
}
```
into

```p4
    @name("ingress.simple_action") action simple_action() {
        {
            bool cond;
            cond = h.eth_hdr.src_addr != h.eth_hdr.dst_addr;
            h.h[1].a = (cond ? h.h[1].a : 8w1);
        }
    }
    apply {
        simple_action();
    }
}
```
[2564.p4.txt](https://github.com/p4lang/p4c/files/5274414/2564.p4.txt)
[2564.stf.txt](https://github.com/p4lang/p4c/files/5274415/2564.stf.txt)

anasyrmia:
Thank you both for your replies.
I've implemented the requested changes regarding the code structure.
The problem with header stacks has been solved the following way:
The condition in if block on the line 203 is what originally caused the problem. This if block is used to remove the statement in liveAssigns vector, in order to not duplicate some assignments. If the code looks something like this (this is just a pseudocode):

bool a = true;
if (a) {
    hdr.x = hdr.x + 1;
}
else {
    hdr.x = hdr.x + 2;
}

Without the if block on line 203 the output of predication would look like:

hdr.x = a ? hdr.x + 1 : hdr.x;
hdr.x = a ? hdr.x + 1 : hdr.x + 2;

Which is wrong because it adds 1 to hdr.x twice in case 'a' is true.
This code segment checks if statement->left is the same in if block and in else block of p4 code. If it is the same, it removes the first and only keeps the second statement (hdr.x = a ? hdr.x + 1 : hdr.x + 2; in this example).

The condition wasn't correct, so some assignments were falsely removed (in 2564.p4 test the line: h.h[0].a = (cond ? 8w2 : h.h[0].a);). In this example the condition didn't see the difference between h.h[0].a and h.h[1].a because toString() function converted the left sides of both statements to [].a. Now that the condition consists of the comparison of lvalue_names of these headers, this if block won't effect these statements, since their lvlaue_names differ.

testdata/p4_16_samples_outputs/issue2345-1-midend.p4
anasyrmia: I will try to handle this optimization.

midend/predication.cpp
anasyrmia: vector::insert can't be used in this case because rv is a BlockStatement pointer, not a vector. Function push_back is defined for BlockStatements. The only way insert could be used here is if it would be defined for BlockStatements.

mbudiu-vmw:
In general, you should not depend on toString(), it is designed for human consumption, not for compiler internals.
Please add a test with arrays.

anasyrmia:
Thank you for pointing that out. Since I replaced the condition with the comparison of lvalue_names there are no more toString function calls in the solution.
Also, I added the test for header stacks.
Let me know if there is anything else regarding this solution that needs fixing.

Issue #2565, opened 2020-09-23
Strength-reduction opportunity


mbudiu-vmw:
We can strength-reduce the following expression: `a ? true : false` (or viceversa) as just `a`.

mbudiu-vmw:
Fixed by #2566

PR #2566, opened 2020-09-23
More strength-reduction optimizations

mbudiu-vmw:


mbudiu-vmw:
Fixes #2565

frontends/p4/strengthReduction.cpp
```diff
@@ -107,6 +109,10 @@ const IR::Node* DoStrengthReduction::postorder(IR::BOr* expr) {
     auto r = expr->right->to<IR::Cmpl>();
     if (l && r)
         return new IR::Cmpl(new IR::BAnd(expr->srcInfo, l->expr, r->expr));
+    if (hasSideEffects(expr))
+        return expr;
+    if (expr->left->equiv(*expr->right))
+        return expr->left;
```
ChrisDodd, line 115: You check for side effects before folding here, but not in the BAnd case above?  They should be consistent.

ChrisDodd (approved):

frontends/p4/strengthReduction.cpp
mbudiu-vmw: There is a similar check in BAnd as well.

Issue #2554, opened 2020-09-18
UBPF fails examples P4 program


hesingh:
I picked the first example in p4c/backends/ubpf/examples/gtp.p4.  I am using latest p4c downloaded today.  Even though gtp.p4 includes a `#define UBPF_MODEL_VERSION 20200304` before including 'ubpf_model.p4`, the P4 program fails to compile.

```bash
$ ./p4c-ubpf -o tmp.c ../backends/ubpf/examples/gtp.p4
error: Expected parser to have exactly 4 parameters
hemant@hemant-S5500BC:/tmp/p4c/build$ 
```

Since the older UBPF model version is being used by gtp.p4, I expected p4c-upbf to accept 3 parameters in parser.

osinstom:
@hesingh I'll fix it soon.

PR #2567, opened 2020-09-24
Fix: UBPF fails examples P4 program #2554

osinstom:
This PR fixes issue #2554. 

The `__ubpf_model_version` was not handled in the `p4c` code. I had to introduce new frontend implementation for uBPF to avoid problems with dependencies.


backends/ubpf/frontend.cpp
```diff
@@ -0,0 +1,166 @@
...
+    if (program == nullptr && options.listFrontendPasses == 0)
+        return nullptr;
+
+    bool isv1 = options.isv1();
+    P4::ReferenceMap refMap;
+    P4::TypeMap typeMap;
+    refMap.setIsV1(isv1);
+
+    auto evaluator = new P4::EvaluatorPass(&refMap, &typeMap);
+    std::initializer_list<Visitor *> frontendPasses = {
```
mbudiu-vmw, line 69: If this list is the same as in the standard front-end, then I would rather suggest to insert your pass in front, then call the standard front-end unmodified. This will make maintenance much simpler.

backends/ubpf/p4c-ubpf.cpp
```diff
@@ -47,7 +48,7 @@ void compile(EbpfOptions& options) {
     if (::errorCount() > 0)
         return;
 
-    P4::FrontEnd frontend;
+    UBPF::FrontEnd frontend;
```
mbudiu-vmw, line 51: yes, here just call `program = new UBPF::getUBPFModel()->run(program);`

mbudiu-vmw (changes requested):

backends/ubpf/frontend.cpp
osinstom: Done

backends/ubpf/p4c-ubpf.cpp
osinstom: Done

mbudiu-vmw (approved):

PR #2570, opened 2020-09-25
Fix EBPF parenthesis issue

hesingh:
I tested the generated C file with clang compiler and no warnings are seen.  Of course, visual inspection of the if-statement in the generated C file is correct with no extra parenthesis.


testdata/p4_16_samples/parenthesis-test_ubpf.p4
```diff
@@ -0,0 +1,107 @@
+/*
```
mbudiu-vmw, line 1: this test in itself is not useful, because we do not check the resulting C file.
Ideally you should have an stf file to check that the evaluated expression has the correct semantics.

mbudiu-vmw (changes requested):
How about Member?

hesingh:
I changed the test file because its if-statement with "do nothing" gets removed in a pass before the backend.

mbudiu-vmw (approved):

PR #2571, opened 2020-09-28
Add a simple PSA test program and STF test to verify parser_error ope…

jafingerhut:
…ration

That is, verify that parser errors caught during ingress parsing are
passed to ingress control in standard metadata field, and similarly
for parser errors caught during egress parsing.


antoninbas (approved):
LGTM

I think that since this is an STF test, the test case could have been made even simpler by not trying to use standard IPv4 / IPv6 headers and using lookaheads in the parser states to trigger `PacketTooShort` errors. However, maybe there is value in using standard headers if you are trying to run this test on  PSA switch that's not bmv2.

fruffy:
@jafingerhut You think this is ready to be merged?

jafingerhut:
Looks like I forgot about this one.  Thanks for the reminder.  The "Squash and merge" button is disabled for me on this one.  Maybe I need to do a comment-only change commit to force it to run CI tasks?  I will try that.

PR #2572, opened 2020-09-30
uBPF: support for default action

osinstom:
Currently, a table miss is not handled in the uBPF backend. This PR introduces the implementation of default action, which is used if a table miss occurs.


mbudiu-vmw (approved):
This design seems relatively similar to ebpf.
I don't see a reason you cannot support initializers.
Enforcing a table to be immutable is something you probably cannot do, though.

PR #2575, opened 2020-10-02
Support for general-purpose switch statements

mbudiu-vmw:
This is a still experimental feature under discussion: https://github.com/p4lang/p4-spec/pull/887
This implementation follows the proposal but uses 'case' labels for general-purpose switch statements, since otherwise the grammar is ambiguous.


fruffy:
I have a question on the semantics of the `eliminateSwitch` pass. The example program is changed from

```p4
control c(inout bit<32> b) {
    apply {
        switch (b) {
            case 32w16: 
            case 32w32: {
                b = 32w1;
            }
            case 32w64: {
                b = 32w2;
            }
            case 32w92: 
            default: {
                b = 32w3;
            }
        }
    }
}
```

to 

```p4
control c(inout bit<32> b) {
    bit<32> switch_0_key;
    @hidden action switch_0_case() {
    }
    @hidden action switch_0_case_0() {
    }
    @hidden action switch_0_case_1() {
    }
    @hidden table switch_0_table {
        key = {
            switch_0_key: exact;
        }
        actions = {
            switch_0_case();
            switch_0_case_0();
            switch_0_case_1();
        }
        const default_action = switch_0_case_1();
        const entries = {
                        32w16 : switch_0_case();
                        32w32 : switch_0_case();
                        32w64 : switch_0_case_0();
                        32w92 : switch_0_case_1();
        }
    }
    apply {
        {
            switch_0_key = b;
            switch (switch_0_table.apply().action_run) {
                switch_0_case: {
                    b = 32w1;
                }
                switch_0_case_0: {
                    b = 32w2;
                }
                switch_0_case_1: {
                    b = 32w3;
                }
            }
        }
    }
}
```
Does a `@hidden` annotation imply that this table cannot be configured from the control plane? So the only possible entries are the constant entries in the list?

mbudiu-vmw:
The entries are const, so they cannot be changed. Hidden says  that the control plane  does not need to know that this table exists

fruffy:
So I was confused about this particular section in the spec:

>Table entries are immutable (const)—i.e., they can only be read and cannot be changed or removed by the control plane. It follows that tables that define entries in the P4 source are immutable. This design choice has important ramifications for the P4 runtime since it does not have to keep track of different types of entries in one table (mutable and immutable). Future versions of P4 may add the ability to mix mutable and immutable entries in the same table, by declaring additional entries properties without the const keyword. 

What the current spec implies is that, once you mark `entries`  `const`, the entire table is immutable and the control plane can not add any additional entries, even in cases where the table is not hidden? So in the example above it would be impossible to add another match-action like `key: 32w16 action: switch_0_case_1`, right?

mbudiu-vmw:
Yes, this is correct.

hesingh:
Please see the p4runtime spec and search for `is_const_flag` for a table for additional clarification.

jfingerh:
@fruffy

For a table with on `@hidden` annotation, it can read the entries and read the default action, regardless of `const` declarations.
If it has `const entries`, the control plane can read but not modify the entries.
If it has `const default_action`, the control plane can read but not modify the default action.

For a table with the `@hidden` annotation, the control plane API cannot see the table at all, so regardless of whether there are `const` declarations on entries or default_action, it does not really matter much, since it can't name the table in order to access it in any way.  The transformation proposed here uses `const default_action` mainly for explicitness that it cannot be changed, even though that could be considered redundant with `@hidden` on the table.

fruffy:
Thanks for all the clarifying comments! I can work with this.

mbudiu-vmw:
Will rework this to remove the 'case' labels.

mbudiu-vmw:
I have removed the 'case' labels by creating a new grammar non-terminal for nonBraceExpression, which represents expressions that cannot start with a brace.

ChrisDodd (approved):

PR #2577, opened 2020-10-07
 Support lookahead with scalars in ebpf 

### LINKED ISSUE #2532: EBPF backend does not support non-header for lookahead
In `p4c/testdata/p4_16_samples/cal-ebpf.p4`, I see lookahead is supported by the ebpf backend.  However, I need a bit-field returned from lookahead to be also supported because, for example, I have to parse inner type of a GTP tunnel to be ipv4 or ipv6.  Thus, I am forced to use this lookahead which fails compiling with p4c-ebpf.

```
state parse_gtp_payload {
        bit<4> version = pkt.lookahead<bit<4>>();
        transition select(version) {
            4w4: parse_inner_ipv4;
            4w6: parse_inner_ipv6;
        }
    } 
```

To reproduce the issue, I changed switch_epbf.p4 for a contrived change and the problem is reproduced.  See log below.

```
$ ./p4c-ebpf -o tmp.c ../testdata/p4_16_samples/switch_ebpf.p4
../testdata/p4_16_samples/switch_ebpf.p4(43): [--Werror=legacy] error: Cannot extract to a non-struct type version
 bit<4> version = p.lookahead<bit<4>>();
        ^^^^^^^
```

P4 program is attached.

[switch1_ebpf_p4.txt.txt](https://github.com/p4lang/p4c/files/5192164/switch1_ebpf_p4.txt.txt)

mbudiu-vmw:
Fixes #2532


mbudiu-vmw:
I think I have lost a commit here, the actual intended fix is not here.

mbudiu-vmw:
I was pretty sure I had committed this code already, but I could not find it anywhere. This PR should now be ready for review.
I would appreciate a review, since this also fixes the broken build for p4c-xdp, which was broken by a prior commit which I merged without enough testing.

mbudiu-vmw:
@hanw: perhaps you can review this quickly; it is a small change.
The ebpf back-end is now broken until we merge this.

backends/ebpf/targets/xdp_target.py
```diff
@@ -0,0 +1 @@
+/home/mbudiu/git/p4c/extensions/p4c-xdp/xdp_target.py
```
rst0git, line 1: The path in this symlink should be relative.

mbudiu-vmw: This file should not be here at all, in fact.

testdata/p4_16_samples_outputs/issue2543-2-midend.p4
```diff
@@ -12,6 +12,7 @@ struct Headers {
 
 control ingress(inout Headers h) {
     @name("ingress.tmp") ethernet_t tmp;
+    ethernet_t tmp;
```
rst0git, line 15: Was this line added intentionally?

mbudiu-vmw: These are automatically generated files

liujed-intel (approved):
Approving to unblock, and to honour a [contract](https://github.com/p4lang/p4c/pull/2600#issuecomment-721477394). I'm not entirely familiar with the EBPF back end, but these changes look reasonable.

rst0git (approved):

PR #2578, opened 2020-10-07
Drop support for Python 2

fruffy:
Since bmv2 now supports Python 3 we can drop Python 2 from the Travis build. This hopefully reduces the total amount of requests, too, so that we do not hit the rate limiter as frequently. The only thing that uses Python 2 after this is the [cpplint.py](https://github.com/p4lang/p4c/blob/master/tools/cpplint.py) file. That change is more involved with the customizations and all.


tools/travis-build
```diff
@@ -57,12 +53,16 @@ apt-get install -y --no-install-recommends \
...
+ln -sf /usr/bin/python3 /usr/bin/python
+ln -sf /usr/bin/pip3 /usr/bin/pip
+
+pip install wheel
+pip install $P4C_PIP_PACKAGES
 
 # Build libbpf for eBPF tests.
 cd /p4c
-python3 backends/ebpf/build_libbpf
+python backends/ebpf/build_libbpf
```
rst0git, line 65: [PEP 394](https://www.python.org/dev/peps/pep-0394/) recommends using `python3` instead of `python`.

mbudiu-vmw (approved):

PR #2579, opened 2020-10-07
Use temporaries for slices used as out arguments

### LINKED ISSUE #2498: Follow-up on slice arguments
I started fuzzing slices again after the recent fix to #2147 (hurray). Unfortunately, there seem to be more nasty edge cases. 

In this action we assign 2 (0000 0010) to `h.h.a` while also assigning 1 to the least significant bit. The final value of `h.h.a` should be 3 (0000 0011). However, in `35_SimplifyDefUse`, the compiler removes the assignment to `h.h.a`. Copy-out should only copy back the least significant bit, but it looks like the pass assumes that the entire header is copied back. At least that is how I understand it. 

```p4
    @name("slice_action") action slice_action_0(inout bit<1> sliced_val) {
        h.h.a = 8w2;
        sliced_val = 1w1;
    }
    apply {
        slice_action_0(h.h.a[0:0]);
    }
}
```

to 

```p4
    @name("slice_action") action slice_action_0(inout bit<1> sliced_val) {
        ;
        sliced_val = 1w1;
    }
    apply {
        slice_action_0(h.h.a[0:0]);
    }
}
```
[slice_assignment.p4.txt](https://github.com/p4lang/p4c/files/5063607/slice_assignment.p4.txt)
[slice_assignment.stf.txt](https://github.com/p4lang/p4c/files/5063606/slice_assignment.stf.txt)

mbudiu-vmw:
Fixes #2498


ChrisDodd (approved):
This probably makes the backend's task harder -- we already introduce too many unnecessary copies in the frontend.

mbudiu-vmw:
this is probably a relatively uncommon case.

PR #2580, opened 2020-10-09
Do not constant fold structs before typing

mbudiu-vmw:
The code contains an example of a bug that is being fixed.


mbudiu-vmw:
@ChrisDodd @liujed @hanw I have about ten open pull requests which could use a review, in case any of you has some time.
Most of them are quite small. Many bug fixes.

liujed-intel (approved):

PR #2584, opened 2020-10-14
Update CI script to improve caching.

smolkaj:
This should speed up CI quite a bit, starting with the first run after this PR.

### DIFF
--- a/.github/workflows/ci-bazel.yml
+++ b/.github/workflows/ci-bazel.yml
@@ -19,12 +19,13 @@ jobs:
     - name: Mount bazel cache
       uses: actions/cache@v2
       with:
-        path: "~/.cache/bazel"  # See https://docs.bazel.build/versions/master/output_directories.html
-        key: bazel1
+        # See https://docs.bazel.build/versions/master/output_directories.html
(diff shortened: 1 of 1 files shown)


mbudiu-vmw:
I will approve this, but I really have no idea whether it works.

smolkaj:
Should be good. We are using the same pattern here: https://github.com/google/p4-pdpi/blob/master/.github/workflows/ci-build.yml

antoninbas (approved):
LGTM

smolkaj:
The Bazel builds are down to ~ 2.5 minutes, so this seems to have worked.

mbudiu-vmw:
It looks like we have some not-so-deterministic code that generates some hidden table names, which caused these tests to fail.
You should probably check-in the modified reference files, or we can merge another PR which fixes this and you can rebase.

smolkaj:
@mbudiu-vmw The travis failure is unrelated to my PR, right? Seems to have failed for several of the last commits to master as well. Happy to rebase once it is fixed.

smolkaj:
Ready to be merged.

PR #2586, opened 2020-10-14
Small changes for better P4 symbol naming

TomasZavodnik:


ir/expression.def
```diff
@@ -298,6 +298,7 @@ class ArrayIndex : Operation_Binary {
     ArrayIndex {
         if (auto st = left ? left->type->to<IR::Type_Stack>() : nullptr)
             type = st->elementType; }
+    toString{ return left->toString() + "[" + right->toString() + "]"; }
```
mbudiu-vmw, line 301: toString was defined in this way especially to highlight where the problem is in an expression involving an array indexing operation. If you print the whole index expression the user may not know what part of the expression is the cause. Before accepting this change we should check whether it impacts the usability of the error messages. Perhaps a better way to output this would be to use "[...]" to make it clear that the index expression is elided.

mbudiu-vmw (changes requested):
Your PR should include all the reference files produced by the tests that change.

mbudiu-vmw:
I never heard of p4c-2019. What is it?

jnfoster:
@mbudiu-vmw looks like a Jira issue number :-)

TomasZavodnik:
> @mbudiu-vmw looks like a Jira issue number :-)

Correct, this probably doesn't make sense here. I'll change the PR's name.

ir/expression.def
TomasZavodnik: Thank you for your comment, @mbudiu-vmw. 

I'm proposing this change with respect to the implementation of `IR::Member::toString()` [here](https://github.com/p4lang/p4c/blob/f40c5d92fd62ebe04c35003ad733005b0ef1535a/ir/expression.def#L281). `IR::Member` and `IR::ArrayIndex` are often nested in each other and in that case their different `toString()` implementations lead to weird results, for example: `Member([].$valid) / ArrayIndex([]) / Member(hdr.stack)`. The proposed change leads to, IMHO, a much better (and very useful) result: `Member(hdr.stack[0].$valid) / ArrayIndex(hdr.stack[0]) / Member(hdr.stack)`.

I've checked the test logs for relevant error messages and found the following differences:
```
-binary_e.p4(32): [--Werror=type-error] error: Array indexing [] applied to non-array type int<2>
+binary_e.p4(32): [--Werror=type-error] error: Array indexing a[2] applied to non-array type int<2>

-issue1696-1.p4(10): [--Wwarn=ordering] warning: []: the order of headers in deparser is not uniquely determined by parser!
+issue1696-1.p4(10): [--Wwarn=ordering] warning: hdr.stack[0]: the order of headers in deparser is not uniquely determined by parser!

-stack_e.p4(27): [--Werror=type-error] error: []: Negative array index -2
+stack_e.p4(27): [--Werror=type-error] error: stack[-2]: Negative array index -2
```
They seem OK to me. What do you think?

mbudiu-vmw (approved):
What will the name annotations of the variables be used for?
Please note that for each identifier the compiler does maintain the original name internally; the original name is used in all error messages. (The IR::ID class has an originalName).

TomasZavodnik:
> What will the name annotations of the variables be used for?
> Please note that for each identifier the compiler does maintain the original name internally; the original name is used in all error messages. (The IR::ID class has an originalName).

I'm trying to get fully qualified names for variables so that I can reference them globally by name. Since the name annotations are processed by the `HierarchicalNames` pass that transforms the names of the annotated nodes to fully qualified ones, adding the annotations seems to be the easiest way to achieve that. The resulting names can be accessed later via `externalName()`. I would like to do the same for `Parameter`s but that seems to be a bit trickier.

mbudiu-vmw:
The compiler has no contract with the outside world to maintain variable names; in fact, there are several passes that assign new names to the variables, e.g., to make them unique in the enclosing scope, or within the whole program. Having unique names can make other passes significantly simpler. We do maintain the original names to give meaningful error messages.

TomasZavodnik:
Yes, that makes sense. Nevertheless, I'm not sure if I got the entire message of yours. It's true that the compiler (e.g. the `UniqueNames` pass) renames variables to make their names unique and it's also true that their original names are preserved e.g. to print error messages, I understand that. Is there anything else that you wanted to express? I'm asking because, unfortunately, none of this helps me to achieve what I need. Unique names assigned to variables by the compiler may not be pretty or even intelligible to the user and, on the other hand, original names themselves are not usually unique. I'm looking for a way to get names that are unique and correspond to the original source code (like ["FabricIngress.spgw_normalizer.hasReturned"](https://github.com/p4lang/p4c/blob/08bfbede779270e63596120287417e8d0eec2228/testdata/p4_16_samples_outputs/fabric_20190420/fabric-frontend.p4#L674)) so that the user knows exactly what they refer to. Does this make sense to you?

mbudiu-vmw:
* When we give an error message we usually indicate the source line and display a little source fragment. This is an unambiguous designation of the identifier involved.
* There are unnamed scopes in P4, like block statements. For such scopes there is no good way to indicate the position
* The 'hasReturned' variable in your example is actually a variable synthesized internally by the compiler; ideally the user should never see a message from the compiler related to this variable.
* Finally, you still didn't say what the intended outcome of this change is. Is it for providing better error messages? If that's indeed the case, you should show some test outputs on which the error messages are improved.

TomasZavodnik:
Thank you for the reaction @mbudiu-vmw, I understand and am sorry for not explaining the desired outcome enough. My example was just an example of a name format I'm seeking, I agree that this one particularly shouldn't be exposed to the user.

My goal is to take all important P4 symbols (named blocks, tables, actions, parse states, parameters, variables etc.) present in the original source code and create a collection of locations (i.e. source info) of definitions and uses for those symbols. The names that we are talking about shall identify and differentiate symbols in that collection, and therefore should be unique. In a way, this collection shall be exported by the compiler and exposed to the user, so the names must be meaningful to them and are expected to correspond to the original source code. 

I'm proposing the name annotations for Declaration_Variable mainly because such annotations are already being used for Declaration_Instance, which I find a useful feature. The proposed changes shall not have any impact on error messages (or other parts of the open-source part of p4c) other than we have discussed. The feature I'm working on shall be an external extension to the compiler.

Issue #1149, opened 2018-01-27
Possible bug in constant folding pass


antoninbas:
I have the following non-sensical program:
```
#include <core.p4>
#include <v1model.p4>

struct H { };
struct M { }

parser ParserI(packet_in pk, out H hdr, inout M meta, inout standard_metadata_t smeta) {
    state start { transition accept; }
}

action empty() { }

control IngressI(inout H hdr, inout M meta, inout standard_metadata_t smeta) {
    apply {
        smeta.egress_spec = smeta.ingress_port + 32 - 16;
    }

};

control EgressI(inout H hdr, inout M meta, inout standard_metadata_t smeta) {
    apply { }
};

control DeparserI(packet_out pk, in H hdr) {
    apply { }
}

control VerifyChecksumI(inout H hdr, inout M meta) {
    apply { }
}

control ComputeChecksumI(inout H hdr, inout M meta) {
    apply { }
}

V1Switch(ParserI(), VerifyChecksumI(), IngressI(), EgressI(),
         ComputeChecksumI(), DeparserI()) main;
```
I noticed that the bmv2 JSON still had constants 32 and -16, even though I was expecting the compiler to do constant folding.
So I dumped the P4 program at the end of the bmv2 midend with `-top4 MidEndLast`:
```
control IngressI(inout H hdr, inout M meta, inout standard_metadata_t smeta) {
    apply {
        smeta.egress_spec = smeta.ingress_port + 9w32 + 9w496;
    }
}
```

I thought this might have been a small issue so I used gdb to walk through the constant folding pass. This is the IR AssignmentStatement:
```
(gdb) p dump(this)
[3029] AssignmentStatement
  left: [1138] Member member=egress_spec
    type: [123] Type_Unknown
    expr: [1137] PathExpression
      type: [123] Type_Unknown
      path: [1136] Path name=smeta absolute=0
  right: [1147] Sub
    type: [123] Type_Unknown
    left: [1144] Add
      type: [123] Type_Unknown
      left: [1141] Member member=ingress_port
        type: [123] Type_Unknown
        expr: [1139] PathExpression
          type: [123] Type_Unknown
          path: [1140] Path name=smeta absolute=0
      right: [1143] Constant value=32 base=10
        type: [1142] Type_InfInt declid=37
    right: [1146] Constant value=16 base=10
      type: [1145] Type_InfInt declid=38
$13 = void
```
[...]

PR #2587, opened 2020-10-14
Reassociation pass; two more strength-reduction optimizations

### LINKED ISSUE #2585: ConstantFolding not folding as expected
For something like this:
```p4
h.data = h.data2 - 8 - 8 - 2 - 16;
```
I would expect `ConstantFolding` to fold this to:
```p4
h.data = h.data2 - 34;
```
however, it does not fold the constants at all.

If parens are added:
```p4
h.data = h.data2 - (8 + 8 + 2 + 16);
```
then the constants are folded as expected.

test: 
[constant-fold.p4.txt](https://github.com/p4lang/p4c/files/5375005/constant-fold.p4.txt)

mbudiu-vmw:
Fixes #2585, #1149

### DIFF
--- a/frontends/p4/reassociation.h
+++ b/frontends/p4/reassociation.h
@@ -1,0 +1,48 @@
+/*
+Copyright 2020 VMware, Inc.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+#ifndef _P4_REASSOCIATION_H_
+#define _P4_REASSOCIATION_H_
+
+#include "ir/ir.h"
+
+namespace P4 {
+
+/** Implements a pass that reorders associative operations when beneficial.
+  * For example, (a + c0) + c1 is rewritten as a + (c0 + c1) when cs are constants.
+  */
+class Reassociation final : public Transform {
+ public:
+    Reassociation() { visitDagOnce = true; setName("Reassociation"); }
+    using Transform::postorder;
+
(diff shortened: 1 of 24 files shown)


jafingerhut:
All of the updated expected outputs look reasonable to me, but as usual, I'm not a good one for reviewing p4c pass changes.

A question: I assume that saturating add and subtract operations are not treated as associative by p4c?

I know that `(a - constant1)` expressions are often (maybe always?) transformed by p4c into a functionally equivalent expression `(a + constant2)` where `constant2` is, I believe, an unsigned representation of the 2's complement of `constant1`.  I suppose that such a transformation is never made for expressions like `(a |-| constant1)` ?

mbudiu-vmw:
I believe that saturated operations are not associative. a + 1 + (-1) is not the same as a if a is max.
Yes, the saturated operations have different opcodes, and I don't think we have many optimizations (if any) for them.
We could probably do some strength-reduction, though.

mbudiu-vmw:
Actually I think this is really one of the simplest passes one could write, so perhaps you would like to take a look at the implementation too.

ChrisDodd (approved):

frontends/p4/reassociation.h
```diff
@@ -0,0 +1,48 @@
...
+    { return reassociate(expr); }
+    const IR::Node* postorder(IR::Mul* expr) override
+    { return reassociate(expr); }
+    const IR::Node* postorder(IR::BOr* expr) override
+    { return reassociate(expr); }
+    const IR::Node* postorder(IR::BAnd* expr) override
+    { return reassociate(expr); }
+    const IR::Node* postorder(IR::BXor* expr) override
+    { return reassociate(expr); }
+};
```
jphurl, line 44: How does the testcase work? since we don't visit `IR::Sub`

jfingerh: If I understand your question correctly, and what the compiler currently does, it transforms all expressions of the form `(variable - constant1)` into another expression guaranteed to evaluate to the same result that looks like `(variable + constant2)`, where I believe `constant2` is the 2's complement negation of `constant1`.  After that transformation, there are no more subtractions by constant any more.

jphurl: ah ok thank you!

Issue #2544, opened 2020-09-14
Compiler Bug: Could not find type of <Declaration_Variable> for declaration of same name


fruffy:
Simply declaring a new local header with the same name leads to a compiler bug currently.

```p4
control ingress(inout Headers h) {
    apply {
        Headers h = h;
    }
}
```

Output:

```
Deprecated_1_CheckDeprecated
FrontEnd_9_Deprecated
FrontEnd_10_CheckNamedArgs
In file: p4c/frontends/p4/typeChecking/typeChecker.cpp:168
[31mCompiler Bug[0m: bugs/crash/same_name_crash.p4(24): Could not find type of <Declaration_Variable>(238) h/34 Headers h = h;
        Headers h = h;
        ^^^^^^^^^^^^^^*/
```
[same_name_crash.p4.txt](https://github.com/p4lang/p4c/files/5217691/same_name_crash.p4.txt)

mbudiu-vmw:
@ChrisDodd : this looks like a bug in the resolveReferences pass; the h on the right-hand side is resolved as the local variable instead of being resolved as the parameter. I believe that this is due to your recent refactoring of this pass. The variable declaration does introduce a new symbol, but it should be only visible after the declaration.

dbrajovic-syrmia:
@ChrisDodd : If possible, I would like to assign this issue to myself for further investigation - I'm interested in contributing to the project and the issue seems fit for a starting point.

mbudiu-vmw:
If you submit a PR and it's good then we'll accept it. I have added you to the list of assignees.

mbudiu-vmw:
Fixed by #2589

PR #2589, opened 2020-10-15
Declaration_Variable shadowing fix (#2544)

dbrajovic-syrmia:
* Parameter shadowing omitted.
* Outer scope variable shadowing is now viable.

Previously, shadowing a parameter, such as:

		apply {
       			Headers h = h;
    		}
	}

would result in a compiler bug. checkShadowing() now handles this case correctly.
Further investigation showed that, and for the same underlying reason, the compiler was able to "resolve" cases like this:

	bit<8> foo = foo;

again, resulting in a bug (later in the TypeInference pass) due to the rhs 'foo' being equated with the lhs.
As of now, this scenario is valid only if we have the following:

	bit<8> foo = 4;
	{
		bit<8> foo = foo;
	}

that is, only when the rhs 'foo' can be resolved to a different declaration (of a different scope).


frontends/common/resolveReferences/resolveReferences.cpp
```diff
@@ -107,6 +107,18 @@ ResolutionContext::lookup(const IR::INamespace *current, IR::ID name,
                 Util::SourceInfo dsi = decl->getNode()->srcInfo;
                 bool before = dsi <= nsi;
                 LOG3("\tPosition test:" << dsi << "<=" << nsi << "=" << before);
+
+                // Type foo = foo (don't resolve rhs to lhs)
+                if (auto *decl_var = decl->to<IR::Declaration_Variable>()) {
```
mbudiu-vmw, line 112: This  fix is very specific to variable declarations, but I suspect that the same problem exists for other declarations and instantiations.

dbrajovic-syrmia: I've expanded the logic in _checkShadowing()_ to cover other parameter shadowing scenarios. As for the code in the _lookup()_ method, I must elaborate:

Since the bug seems to be occurring (at least for now) only when there is an identical name on both sides of the '=' sign, then our possible node candidates become:

1. Declaration_Variable (has _initializer_ as member)
2. Declaration_Constant (also has _initializer_ as member)
3. and you mentioned instantiation (which is used for Declaration_Instance).

The first 2 are now accounted for. As for 3. I have to ask - could you elaborate on how instantiation is involved ? (test examples are much appreciated) 
From what I've seen in the specification, instantiations don't have a right-hand side. Unless you're referring to the experimental feature (https://github.com/p4lang/p4c/blob/c0907946eeb46a96bc8fee822ba083fb3c442bd2/frontends/parsers/p4/p4parser.ypp#L673)

mbudiu-vmw: Yes, we have to support the experimental feature too.
I haven't checked, but I expect that this could also apply to other declarations that introduce scopes, e.g., `struct S { S x; }`

dbrajovic-syrmia: > I haven't checked, but I expect that this could also apply to other declarations that introduce scopes, e.g., `struct S { S x; }`

You were right, it does. I'm working on a fix at the moment.
> Yes, we have to support the experimental feature too.

I haven't had much luck in finding some examples that demonstrate use of this feature. Any short examples you could share ?

mbudiu-vmw:
I suspect that the right way to fix this is to use `getConext<IR::Declaration>()`. This will tell you whether the current reference is a child of the declaration node, i.e., it is nested within the declaration.

dbrajovic-syrmia:
Thanks to the `findContext<>()` method I was able to arrive at the safe conclusion that the origin of the bug is in the `lookup()` method: 

When a given `name` (parameter) is attempted at being resolved, "faulty" declaration candidates are being given a pass (in certain scenarios, such as the one from the parameter shadowing example and yours). This is mostly due to the somewhat "lax" critetium of the Position Test. All it does is determine whether a declaration was "written earlier" than the given name (in the source code).

This is why my initial approach was to narrow down this selection of candidates, but using the same idea based on SourcePosition. `findContext<>()` allows this problem to be expressed more verbally: _anytime the found declaration matches the node retrieved from `findContext<T>(), where T is either a `Type_Declaration` or `Declaration`, that declaration becomes a potential candidate of this bug._ Depending on the namespace in which we are resolving and the type of the `name` parameter, we can throw an error or simply reject the candidate (and continue looking up, which allows us to keep the aforementioned shadowing).  

As a consequence, only 4 tests failed during the `make check`. Namely their outputs differed after applying the fix. Two of these test were a fortunate capture of this very bug only later in the TypeInference pass: error such as `Cannot refer to control inside itself` etc. This fix recognizes those situations earlier.  The other two tests had an incorrect shadowing warning and an incorrect 'recursive action call` warning. I claim this because when inspecting the actual tests, none of the code seemed to actually reflect these errors.

testdata/p4_16_errors_outputs/functors2_e.p4-stderr
```diff
@@ -1,3 +1,6 @@
-functors2_e.p4(20): [--Werror=type-error] error: p: Cannot refer to control inside itself
+functors2_e.p4(20): [--Werror=invalid] error: Recursive declaration of 'p' in 'control p'
```
mbudiu-vmw, line 1: This message is not accurate, since this is not a declaration. The previous message was more precise.

testdata/p4_16_errors_outputs/issue1541.p4-stderr
```diff
@@ -1,9 +0,0 @@
-issue1541.p4(25): [--Wwarn=shadow] warning: 'hash' shadows 'hash'
```
mbudiu-vmw, line 1: why isn't this a warning? Isn't this declaration shadowed?

testdata/p4_16_errors_outputs/issue2544_recursive.p4-stderr
```diff
@@ -0,0 +1,6 @@
+issue2544_recursive.p4(2): [--Werror=invalid] error: Recursive declaration of 'S' in 'struct S'
```
mbudiu-vmw, line 1: This message isn't appropriate here either. This is not a declaration of S, but of x, so it is not really a recursive declaration.

mbudiu-vmw (changes requested):
While this indeed fixes the issue we had, the error messages themselves are not clear enough.
Also, you have lost some shadowing errors, this looks like a bug.

testdata/p4_16_errors_outputs/functors2_e.p4-stderr
dbrajovic-syrmia: I very much agree, which is why I've been meaning to ask: is there a more suitable error message (one I can use in resolveReferences.cpp) or is adding a new error acceptable?

testdata/p4_16_errors_outputs/issue1541.p4-stderr
dbrajovic-syrmia: It still is but I've moved the file from errors to samples - it doesn't output the `Recursive action call` error it did previously. I don't think this was an error because the inner `hash` is a method included in the header, whereas `action hash() { ... }` simply encapsulates it (semantically there should be no recursion here).

testdata/p4_16_errors_outputs/functors2_e.p4-stderr
mbudiu-vmw: The main goal is to make the life of the user easy; if adding a new message is the best way, it should be added.

testdata/p4_16_errors_outputs/issue1541.p4-stderr
mbudiu-vmw: I don't see a -stderr file for issue 1541 in samples_outputs.

mbudiu-vmw (commented):
Can you also add the test from #2594?

dbrajovic-syrmia:
> Can you also add the test from #2594?

Of course, it has been added (as well as issue1541-stderr). 

Considering how these errors stem from the same source, but their messages differ in context, instead of adding several new `ErrorTypes`, I've decided to use the same message for all of them (for the sake of brevity of the output and code readability). Hopefully, the message:

`[--Werror=unsupported] error: Self-referencing types not supported`

 will provide some indication to the user just how the compiler reasons about these "nested" references.

mbudiu-vmw (approved):

Issue #2004, opened 2019-07-15
parser duplicated matches not optimized out


zma-barefoot:
As title says, duplicated matches 0x0800 are not optimized out.

```
header_type h_t {
    fields {
        f : 16;
    }
}

header h_t a;
header h_t b;

parser start {
    extract(a);
    return select(latest.f) {
        0x0800         : parse_b;
        0x0800         : parse_b;
        default        : ingress;
    }
}

parser parse_b {
    extract(b);
    return ingress;
}

control ingress { }
```

`FrontEndLast:`

```
    @name(".start") state start {
        packet.extract<h_t>(hdr.a);
        transition select(hdr.a.f) {
            16w0x800: parse_b;
            16w0x800: parse_b;
            default: accept;
        }
    }
```

It would also be nice to give a warning of duplicated matches.

mbudiu-vmw:
I general it's not easy to tell whether the sets of two labels intersect or not. Moreover, when keysets include value_sets, this cannot even be determined statically. We could have some heuristics, but a complete solution is not very simple.

zma-barefoot:
I agree with value sets we cannot determine duplicated matches statically.

With constant matches, can you give an example that we cannot trivial determine sets of two labels intersect or not?

mbudiu-vmw:
The problem in general is not checking whether two labels overlap, but if label M is included in the union of the prior M-1 labels. That's not trivial for a mix of range and don't care queries, e.g., `5...8` and `6 &&& 4`

Note that the user may want overlapping labels for convenience; the only clear error case is when a label is completely redundant.

zma-barefoot:
Okay, I see. Sounds like the general solution is to expand all labels and compute set inclusion based on ternary match, which I agree may not be trivial.

jafingerhut:
It is an NP-complete problem, equivalent to satisfiability, and solvable using off-the-shelf SAT solver software, to determine whether rule M is matchable, vs. it is unmatchable because it is completely "covered" by the union of the previous M-1 rules, if they can be ternary.

zma-barefoot:
> It is an NP-complete problem, equivalent to satisfiability

I'm genuinely intrigued to a see the reduction to boolean satisfiability.

jafingerhut:
[...]

PR #2591, opened 2020-10-19
Fix: Issue #2004 parser duplicated matches not optimized out

DarinkaNesovic:
This fix contains solution for overlaping constant and range labels inside SelectExpression.
For example:
	3 : accept;
	0x3: reject; (This label will not exist in optimized code)
	5 .. 10: accept;
	1 .. 15: reject; (This range will be parsed in two ranges:
				1 .. 2: reject; (This part has 2 ovelapings- with label "3 : accept;" and "5 .. 10: accept;")
				11 .. 15: reject; (This part ovelaps only with label "5 .. 10: accept;"))
	6 .. 9: reject;	(This range will be removed because it's already asigned insidelabel "5 .. 10: accept;")

So, parsed code will look like this:
	3: accept;
	5 .. 10: accept;
	1 .. 2: reject;
	11 .. 15: reject;

This code covers all posible overlapings between constants, ranges, and constants with ranges.

New SelectExpression will be created with new SelectCasses and it will be forwarded to further processing.
If labels are not valid, this pass will not exclude or optimize them, they will be forwarded to further processing.

This code doesn't cover overlaping optimization of masks, Select Expression with structs and constants that are not Constant type.


mbudiu-vmw:
I think that this pass should be decomposed into two separate passes. The detection of completely redundant labels is arguably useful for any target, so it probably belongs to the front-end. The detection of overlapping labels may or may not be beneficial, depending on the cost criteria of the target; perhaps some matches are much faster/cheaper on a target (e.g., powers of two limits), so having overlapping ranges could actually be desirable. So the splitting of ranges into non-overlapping ranges should be a separate mid-end pass, which is optional (only architectures that need it should use it).

jafingerhut:
Completely agreed with Mihai's recommendations.  If you can prove an entry is impossible to take, that seems useful for anyone compiling P4 code for any target, and is worth at least a warning.

The other changes look potentially useful for some targets (maybe those running on general purpose CPUs?), but actually _harmful_ for those with TCAMs in their hardware parsers.  So yes, optional is the way to go there.

DarinkaNesovic:
@mbudiu-vmw @jafingerhut I agree with your propositions on parsing code into two parts. Code has been modified, so please have a look.

mbudiu-vmw:
@fruffy : this test failed because your tool claims that the two programs are not equivalent. Is this a false positive, or is this a real bug?

fruffy:
Both actually. The parser is not well tested so it is likely that there are still bugs in my tool. I fixed one. However, this pull request also seems to have uncovered a bug in the `ReplaceSelectRange` pass? 

I reduced the issue to the following program:

```p4
    state start {
        packet.extract<H>(hdr = hdr.h);
        transition select(hdr.h.a) {
            4w4 .. 4w7: reject;
            4w1 .. 4w4: accept;
            default: reject;
        }
    }
}
```

After ReplaceSelectRange this normally looks like:
```p4
    state start {
        packet.extract<H>(hdr = hdr.h);
        transition select(hdr.h.a) {
            4w4 &&& 4w12: reject;
            4w1 &&& 4w15: accept;
            4w2 &&& 4w14: accept;
            4w4 &&& 4w15: accept;
            default: reject;
        }
    }
}
```
However with this pull request, this is now optimized to: 

```p4
    state start {
        packet.extract<H>(hdr = hdr.h);
        transition select(hdr.h.a) {
            4w4 &&& 4w12: reject;
            4w1 &&& 4w0: accept;
            4w2 &&& 4w0: accept;
            default: reject;
        }
    }
}
```
Which is incorrect, I believe. `4w1 &&& 4w0: accept` also covers the values `4w13`, `4w14`, and `4w15` although these should lead to a reject.

I think the reason for this is that `optimizeSelectLabels` loses some information. The range [here](https://github.com/p4lang/p4c/pull/2591/files#diff-384c213a31bf8d6970d73d823883636ee481ef8895e85babe428edf10c4f4297R25) does not have a bit width?

DarinkaNesovic:
@fruffy You're right, there is a bug inside OptimizeSelectLabels. When new Range or Constant are created, there is no information about width of bits. Because of that method rangeToMasks create wrong mask, and result is this:
4w1 &&& 4w0: accept;
4w2 &&& 4w0: accept;
I'm working now on this problem.

DarinkaNesovic:
Code is updated, so please have a look.

midend/optimizeSelectLabels.cpp
```diff
@@ -0,0 +1,428 @@
...
+
+IR::Range* DoOptimizeSelectLabels::getRange(SelectLabel* range) {
+    const IR::Expression* left = new IR::Constant(range->rangeConstantTypeInfo,
+                                                        range->rangeValue.first, range->base);
+    const IR::Expression* right = new IR::Constant(range->rangeConstantTypeInfo,
+                                                        range->rangeValue.second, range->base);
+    return new IR::Range(range->typeInfo, left, right);
+}
+
+SelectLabel* DoOptimizeSelectLabels::insertValues(SelectLabel* root, big_int value) {
```
mbudiu-vmw, line 30: to me it looks like this should be a method of SelectLabel.

midend/optimizeSelectLabels.cpp
```diff
@@ -0,0 +1,428 @@
...
+            else
+                leafNode->right = new SelectLabel(rightSubRange, root->base,
+                                                  root->typeInfo, root->rangeConstantTypeInfo);
+        }
+    }
+    return leafNode;
+}
+
+SelectLabel* DoOptimizeSelectLabels::insertValues(SelectLabel* root,
+                                                            std::pair<big_int, big_int> value) {
```
mbudiu-vmw, line 124: inconsistent indentation.
Should this also be a method of SelectLabel?

midend/optimizeSelectLabels.h
```diff
@@ -0,0 +1,95 @@
...
+
+namespace P4 {
+
+/**
+   This class contains methods that optimize select labels.
+   If some labels overlaping or one contains the other, user will get warning
+   and code will be optimized.
+*/
+
+// This struct define node in binarry tree
```
mbudiu-vmw, line 33: typo

midend/optimizeSelectLabels.h
```diff
@@ -0,0 +1,95 @@
...
+                          const IR::SelectCase* currentExpression);
+    // Handler for all optimizations of Select Casses
+    const IR::Node* postorder(IR::SelectExpression* expression) override;
+};
+
+class OptimizeSelectLabels : public PassManager {
+ public:
+    explicit OptimizeSelectLabels(ReferenceMap* refMap, TypeMap* typeMap) {
+        passes.push_back(new DoOptimizeSelectLabels(refMap, typeMap));
+        passes.push_back(new ResolveReferences(refMap));
```
mbudiu-vmw, line 87: you don't need to resolve references or compute the type map after you change program; you should do this *before* you change the program, if you need these to be accurate. If you don't need the typeMap or the referenceMap, then you should remove OptimizeSelectLabels altogether and keep just DoOptimizeSelectLabels.

frontends/p4/removeDuplicatedSelectLabels.cpp
```diff
@@ -0,0 +1,101 @@
...
+namespace P4 {
+
+const IR::Node* DoRemoveDuplicatedSelectLabels::postorder(IR::SelectExpression* selectExpression) {
+    auto vecCases = selectExpression->selectCases;
+    std::vector<const IR::Constant*> vecOfConstLabels;
+    std::vector<const IR::Range*> vecOfRangeLabels;
+    // Vector that will contain optimized select casses
+    IR::Vector<IR::SelectCase> newVecCasses;
+
+    for (auto currentExpression : vecCases) {
```
mbudiu-vmw, line 12: This is not really an expression, it's a select case, so the variable name is not the most appropriate.

midend/optimizeSelectLabels.cpp
```diff
@@ -0,0 +1,428 @@
...
+       getElementsOfParsedRange(root->right, elementsOfRange);
+}
+
+const IR::Node* DoOptimizeSelectLabels::postorder(IR::SelectExpression* selectExpression) {
+    auto vecCases = selectExpression->selectCases;
+    std::vector<SelectLabel*> vecOfLabels;
+    // Vector that will contain optimized select casses
+    IR::Vector<IR::SelectCase> newVecCasses;
+
+    for (auto currentExpression : vecCases) {
```
mbudiu-vmw, line 319: You are never looking at the state component of the SelectCase. I would think that could be useful for the optimization too: for example, if you have two overlapping ranges you can optimize them differently if they have different states or the same state.

midend/optimizeSelectLabels.cpp
```diff
@@ -0,0 +1,428 @@
...
+SelectLabel* DoOptimizeSelectLabels::insertValues(SelectLabel* root,
+                                                            std::pair<big_int, big_int> value) {
+    SelectLabel* mainRoot = root;
+    SelectLabel* leafNode = nullptr;
+
+    // Search leaf node where value will be inserted
+    while (mainRoot != nullptr) {
+        leafNode = mainRoot;
+        if (mainRoot->type == SelectLabel::RangeLabel) {
+            // Overlaping two ranges on left side (eg. 1..6 and 4..9 -> 5..9)
```
mbudiu-vmw, line 132: I believe that this comment is wrong.

midend/optimizeSelectLabels.cpp
```diff
@@ -0,0 +1,428 @@
...
+                if (mainRoot->left != nullptr) {
+                    if (mainRoot->left->type == SelectLabel::RangeLabel) {
+                        if (((value.first >= mainRoot->left->rangeValue.first) &&
+                                    (value.first <= mainRoot->left->rangeValue.second)) ||
+                                        ((value.second >= mainRoot->left->rangeValue.first) &&
+                                            (value.second <= mainRoot->left->rangeValue.second)))
+                            mainRoot = mainRoot->left;
+                        else
+                            mainRoot = mainRoot->right;
+                    } else {
```
mbudiu-vmw, line 150: If you handle the else case first you can reduce one level of indentation and make the code more readable:
`if (... != ...) return nullptr;`

midend/optimizeSelectLabels.h
```diff
@@ -0,0 +1,95 @@
...
+    union {
+      big_int constValue;
+      std::pair<big_int, big_int> rangeValue;
+    };
+
+    int base;
+    const IR::Type *typeInfo;
+    const IR::Type *rangeConstantTypeInfo;
+
+    SelectLabel(big_int value, int base, const IR::Type *typeInfo) :
```
mbudiu-vmw, line 45: It looks like this is a decision tree. You should probably describe in a comment the invariants maintained by this tree - each node represents a range, the ranges are non-overlapping, etc. Calling this DecisionTree is probably a better idea, a SelectLabel does not seem an appropriate name.

midend/optimizeSelectLabels.cpp
```diff
@@ -0,0 +1,428 @@
...
+    for (auto currentExpression : vecCases) {
+        auto expression = currentExpression->keyset;
+        // Current label is constant
+        if (expression->is<IR::Constant>()) {
+            auto constant = expression->to<IR::Constant>();
+            auto value = constant->value;
+            bool alreadyAssigned = false;
+
+            SelectLabel* label = new SelectLabel(value, constant->base, constant->type);
+            for (SelectLabel* previousLabel : vecOfLabels) {
```
mbudiu-vmw, line 328: Since you have already run the front-end passes, can this still happen?

midend/optimizeSelectLabels.cpp
```diff
@@ -0,0 +1,428 @@
...
+                }
+            }
+            if (!alreadyAssigned) {
+                vecOfLabels.push_back(label);
+                newVecCasses.push_back(currentExpression);
+            }
+        // Current label is range
+        } else if (expression->is<IR::Range>()) {
+            std::vector<SelectLabel*> elementsOfRange;
+            auto range = expression->to<IR::Range>();
```
mbudiu-vmw, line 359: move up: `if (auto range = ...)`

midend/optimizeSelectLabels.cpp
```diff
@@ -0,0 +1,428 @@
...
+                               new IR::Constant(previousLabel->constValue, previousLabel->base);
+                            ::warning(ErrorType::WARN_OVERLAPING,
+                                "Element of range label %1% .. %2% is already assigned "
+                                "by constant label %3%. Optimization will parse this range.",
+                                range->left, range->right, previousConstant);
+                            insertValues(treeRoot, previousLabel->constValue);
+                            insideRange = true;
+                        }
+                    } else {
+                        if ((rangeValues.first >= previousLabel->rangeValue.first &&
```
mbudiu-vmw, line 389: How about creating a utility function that can take two ranges and return an enum with values like "disjoint/inside/overlapleft/overlapright"? The you can reuse it everywhere and make the code much less prone to errors due to incorrect copy-and-paste.

mbudiu-vmw (changes requested):

PR #2592, opened 2020-10-20
Support for generic structs

mbudiu-vmw:
This PR adds support for an experimental features that allows generic structs, headers, and unions (but not yet generic stacks).
A generic struct declaration looks like this:

```
struct S<T> {
    T field;
    bit<32> otherField;
}
```

A generic struct must be instantiated with some type parameters:

```
typedef S<bit<32>> SB;
const SB b = { field = 0, otherField = 1 };
```

This is just syntactic sugar; a new pass in the front-end will actually replace the generic types with concrete types specialized with their arguments, e.g., in this case

```
struct S_0 {
    bit<32> field;
    bit<32> otherField;
}
typedef S_0 SB;
```

### DIFF
--- a/backends/bmv2/common/lower.cpp
+++ b/backends/bmv2/common/lower.cpp
@@ -213,7 +213,7 @@ RemoveComplexExpressions::simplifyExpression(const IR::Expression* expression, b
         auto simpl = simplifyExpressions(&si->components);
         if (simpl != &si->components)
             return new IR::StructExpression(
-                si->srcInfo, si->typeName, si->typeName, *simpl);
+                si->srcInfo, si->structType, si->structType, *simpl);
         return expression;
     } else {
         ComplexExpression ce;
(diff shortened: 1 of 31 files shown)


mbudiu-vmw:
Recursive type instantiations, where a type is specialized with itself are not allowed (this was already true).
Fields can also be generic:
```
struct S<T> ...
struct G<T> {
   S<T> field;
}
```

mbudiu-vmw:
@fruffy : this build has failed because I have renamed a field in the IR and the z3 code generator does not compile anymore.
We should find a way to coordinate such changes. Perhaps you can also not fail the build if compilation fails - but it would still be nice to get an alert when such a thing happens.

fruffy:
Yeah it is unfortunate, I do not have a good solution for build issues currently. For now I monitor pull requests manually to see if things go wrong. Will think about it in the next couple of days. One thing that we can do at least is to exclude Gauntlet from being built in the other tests.

I will try to get this fixed by Friday.

mbudiu-vmw:
@fruffy : this has still failed because the code generator for z3 could not be compiled.

fruffy:
I think #2593 needs to be merged first. Otherwise I have no means to handle issues like these.

fruffy:
Fwiw I tested a run on my own branch and this is the output:
https://github.com/fruffy/p4c/runs/1330991461

mbudiu-vmw:
@fruffy : this still fails the build because the validation step fails to compile.

ChrisDodd (approved):

PR #2593, opened 2020-10-22
Separate Gauntlet from the normal P4C tests

fruffy:
When the toz3 extension does not build only the validation test should fail. Now the question is what to do when the build for Gauntlet fails. Just ignore or try to issue some sort of warning? 
There is also the possibility of using [allow_failures](https://docs.travis-ci.com/user/build-matrix/#rows-that-are-allowed-to-fail) in Travis but I am not exactly sure if that has the intended effect.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -53,9 +53,9 @@ script:
 # We only test the front end and some mid end passes for now.
 jobs:
   include:
-    - install:
-        - docker build --network ccache_network -t p4c --build-arg IMAGE_TYPE=test --build-arg ENABLE_UNIFIED_COMPILATION=$UNIFIED --build-arg ENABLE_GMP=$ENABLE_GMP --build-arg VALIDATION=ON .
     - name: "Validation"
       os: linux
+      install:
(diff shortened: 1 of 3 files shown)


mbudiu-vmw:
I can't really guess what this does. Can you explain?

fruffy:
The `toz3` extension which is needed for Gauntlet Validation tests will only build for the Gauntlet test in Travis. For all the other normal tests the extension will be skipped. So a failure in the Gauntlet build will not affect them. 

With `allow_failures` it looks like Travis will pass the test if there is a failure. However, this may mask actual validation errors...

mbudiu-vmw:
Someone should see validation errors.
Ideally only validation errors should be fatal.
Building the tool or error in compiling the program should be ignored.

fruffy:
So I added [`travis_terminate`](https://github.com/travis-ci/travis-build/blob/f603c0089116364ccb820cad790b0837f358fb53/lib/travis/build/templates/header.sh#L163) to the build command:
`docker build --network ccache_network -t p4c --build-arg IMAGE_TYPE=test --build-arg ENABLE_UNIFIED_COMPILATION=$UNIFIED --build-arg ENABLE_GMP=$ENABLE_GMP --build-arg VALIDATION=ON . || travis_terminate 0`
With that the script should terminate if the build fails, otherwise continue. I hope this works, this is hard to test.

mbudiu-vmw:
We can see whether our PRs with new features will work now without crashing.

mbudiu-vmw (approved):

PR #2596, opened 2020-10-27
Support header and struct for logging (fixes #2201)

### LINKED ISSUE #2201: Unsupported types for log_msg calls

lcikaric:
Changing the condition enabled to json file generation. This is a prerequisite for the execution of behavioral-model, where other changed were made.

    Example 1:
    header Hdr {
            int<4> a;
            bit<4> b;
    }
    Hdr h = {3,5};
    log_msg("h = {}", {h});
    Output: h = {'a': 3, 'b': 5, '$valid$': 1}

    Example 2:
    struct Str {
            int<4> a;
            bit<4> b;
            bool c;
    }
    Str s = {3,5,true};
    log_msg("s = {}", {s});
    Output: s = {'a': 3, 'b': 5, 'c': 1}


testdata/p4_16_samples/issue2201-bmv2-test1.p4
```diff
@@ -0,0 +1,120 @@
...
+control ingressImpl(inout headers_t hdr,
+                    inout metadata_t meta,
+                    inout standard_metadata_t stdmeta)
+{
+
+    apply {
+
+        Header hdr1 = {3,-1,true};
+        log_msg("hdr1 = {}",{hdr1});
+        // Output: hdr1 = {'val1': 3, 'val2': -1, 'val3': 1, '$valid$': 1}
```
jafingerhut, line 85: I know there is currently no convenient way to automate the checking of the log message output in p4c's automated tests, but even so, I think it would be good to add a call to log_msg() on a header that is currently invalid, to at least test that this doesn't cause behavioral-model to crash.

mbudiu-vmw:
This is very nice. Hopefully more architectures will support this.

mbudiu-vmw (approved):
Still, is there any reason not to merge this? No tests we have are impacted.

jafingerhut:
I guess the test program does not have a name ending in '-bmv2.p4' and thus _will not_ run bmv2 with the compiled output of this program as input.  It seems best for long term slightly better test coverage if it _did_ run bmv2.  If so, then merging these changes should wait for the corresponding p4lang/behavioral-model merge, plus enough hours for that to be used for p4lang/p4c automated tests.

lcikaric:
On my local machine (Linux, Ubuntu 18.04) all tests are passing, same as for macOS on Travis. Can you possibly help me interpret why those two tests which I added fail, on Linux on Travis, with message:
`terminate called after throwing an instance of 'std::runtime_error'
  what():  in Json::Value::operator[](char const*)const: requires objectValue
Could not connect to any of [('127.0.0.1', 9398), ('127.0.0.1', 9398)]
Could not connect to thrift client on port 9398
Check for  /p4c/testdata/p4_16_samples/issue2201-test1-bmv2.stf
*** simple_switch died with return code -6
*** CLI process failed with exit code 1
*** Test failed`
because I can't reproduce error locally?

mbudiu-vmw:
What happens when you run this on your machine: `./bmv2/testdata/p4_16_samples/issue2201-test1-bmv2.p4.test`?

PR #2600, opened 2020-11-04
Enable -Werror in CI

liujed-intel:
This ensures that GCC warnings are addressed in PRs before they are merged. Also addressed a couple of uninitialized variables (one seemingly real, one a false positive).

### DIFF
--- a/.github/.travis.yml.disabled
+++ b/.github/.travis.yml.disabled
@@ -41,7 +41,7 @@ install:
   # build before continuing with the rest, which produces a unified build.
   # This is done here on MacOS; for Linux, this is done in Dockerfile.
   - if [[ $TRAVIS_OS_NAME == 'linux' ]] ; then docker build --network ccache_network -t p4c --build-arg IMAGE_TYPE=test --build-arg ENABLE_UNIFIED_COMPILATION=$UNIFIED --build-arg ENABLE_GMP=$ENABLE_GMP . ; fi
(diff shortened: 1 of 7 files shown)


mbudiu-vmw:
I will trade you a review of this one for one of #2577

mbudiu-vmw (approved):

jnfoster:
Wow, we should open a market for PR reviews.

mbudiu-vmw:
If that works out the way I expect I can retire soon.

liujed-intel:
Thanks, @mbudiu-vmw. I've submitted a review for your PR.

We seem to be out of credits on Travis, so none of our tests are running.

mbudiu-vmw:
Is there anything that we can do about the Travis builds? I thought that the credits are per day, but none of the builds seems to run now. We also haven't had too many builds lately for this project. Should we ask ONF to pay for this?

jnfoster:
I'll start a discussion with ONF staff.

antoninbas:
It seems that the credits don't ever renew, and it would have taken about 10 p4c builds to consume the 10k credits :P

The $129/month plan with the 2 concurrent workers would definitely unstuck us. I think a lot (all?) of what we do with Travis CI could also be done with Github actions, but they may eventually enforce limits there as well, and it would be a bit of an effort to migrate everything.

fruffy:
I personally do not see a bright future for Travis, so it might make sense to slowly switch. I opened an experimental pull requests that tries out workflows here (#2601).

mbudiu-vmw:
@liujed : you may want to adapt this to use github actions.

mbudiu-vmw:
It looks like your warnings are too strict, now bison warnings fail the build.

liujed-intel:
Yes, I saw. I'll look into ignoring files generated from bison and flex, but this is pretty low priority for me right now, so it might be a bit before I come back to this.

ChrisDodd:
In general, we put `#pragma`s into the .l/.y files to get copied into the generated code an silence any warnings from that code we can't fix.

Issue #2573, opened 2020-10-01
#2215 breaks mac builds


asl:
81a1ae6 uses `realpath` to make relative links. However, this solution is not portable and effectively breaks mac build, since:

1. There is no `realpath` builtin in `/bin/sh` (which executes the shell part there)
2. The `realpath` builtin in `/bin/fish` does not support `--relative-to`
3. `realpath` from `coreutils` does support `--relative-to` (and the intent of the original PR seems was to use it), but `coreutils` are not installed by default, if installed it will be `grealpath` in order not to mess with the system utilities

Tagging @ChrisDodd @mbudiu-vmw

PR #2602, opened 2020-11-06
Do not assume the presence of realpath

rst0git:
`realpath` is not present by default on macOS, and `realpath` provided from coreutils does support `--relative-to`.

Instead use the equivalent functionality provided by the cmake's `file` command.

Fixes: #2573


rst0git:
cc @mbudiu-vmw  @ChrisDodd

mbudiu-vmw (approved):

mbudiu-vmw:
We are working to get more Travis credits; until then I am afraid that the tests will not run.

jnfoster:
I think `git commit --allow-empty` will do this?

PR #2604, opened 2020-11-09
Re-enable MacOS build

mbudiu-vmw:


fruffy:
So it seems like the MacOS build is getting stuck at `[ 17%] Generating IR class files`.
I do not know how to fix this but I have seen it happen before. What could be the cause?

jnfoster:
I have seen this problem before.

IIRC, clang (or maybe it was gcc?) has a bad interaction with the BDW garbage collector. If you switch compilers or build with debugging information enabled, it will go through.

I have a long work queue but will try to find time to rebuild p4c on my Mac in the background today.

mbudiu-vmw:
The C compiler identification is AppleClang 12.0.0.12000032.

mbudiu-vmw:
This kind of problem is one of the reasons we are running these MacOS tests. I don't see an easy workaround - we can use gcc for MacOS, but we probably also need to document this problem for everyone who may use MacOS.

asl:
My experience shows that BDW is very unstable and unusable on Mac. With GC disabled everything works fine.

jnfoster:
I'm doing a fresh build now and will report what I find...

jnfoster:
I just confirmed that compiling for `DEBUG` rather than `RELEASE` will work on OS X. 

Specifically, I'm on 
```
Apple clang version 12.0.0 (clang-1200.0.32.21)
Target: x86_64-apple-darwin19.6.0
```

A few issues:
* The old Travis script runs this command
```
ctest --output-on-failure -j 2 --schedule-random -LE "bpf$"
```
rather than
```
make -j4 check
```
The former passes all tests whereas the latter fails about 50 eBPF and uBPF tests. So it seems the CMakefiles need to be updated. For example, I see this in the output generated by `cmake`:
```
-- Start configuring eBPF back end
-- Detected kernel version: 19.6.0
```

* The CMakefiles do not check that `scapy` and `ply` are installed.

mbudiu-vmw:
It sounds fragile to rely on debug working. So I will just disable gc for the macos tests for now.

jnfoster:
I haven't been able to get a `gcc`-based build to go through, but mostly because of Homebrew nanny state nonsense around `flex`.

antoninbas (approved):
LGTM - it seems the one significant change is disabling the garbage collector for the macOS build

mbudiu-vmw:
Yes, that is correct, people have pointed out that the GC does not play well with newer versions of clang on OSX.
@ChrisDodd did all sorts of tweaks to the GC in the past, I don't know if he has any suggestions or a new Mac to test this.

mbudiu-vmw:
I will merge this now so we get tests running on osx, and we can look into the GC later.

antoninbas:
For what it's worth I remember running into a similar issue (for me it was a crash of the IR generator at `[ 17%] Generating IR class files`) on some Linux systems as well. I finally tracked it down to the use of `sbrk` which is implemented differently based on the C standard library you use (in my case `musl`).

So I wrote a patch to replace `sbrk` with `mmap`:

```text
diff --git a/lib/gc.cpp b/lib/gc.cpp
index 899cd8a47..1816f7725 100644
--- a/lib/gc.cpp
+++ b/lib/gc.cpp
@@ -19,7 +19,7 @@ limitations under the License.
 #include <gc/gc_cpp.h>
 #include <gc/gc_mark.h>
 #endif  /* HAVE_LIBGC */
-#include <unistd.h>
+#include <sys/mman.h>
 #include <new>
 #include "log.h"
 #include "gc.h"
@@ -69,16 +69,33 @@ void operator delete(void *p) _GLIBCXX_USE_NOEXCEPT {
 void *operator new[](std::size_t size) { return ::operator new(size); }
 void operator delete[](void *p) _GLIBCXX_USE_NOEXCEPT { ::operator delete(p); }
 
+void *raw_alloc(size_t size) {
+    void *p = mmap(0, size + 16, PROT_READ|PROT_WRITE, MAP_PRIVATE|MAP_ANONYMOUS, -1, 0);
+    memcpy(p, static_cast<void *>(&size), sizeof(size));
+    return static_cast<void *>(p) + 16;
+}
+
+size_t raw_size(void *ptr) {
+    size_t size;
+    memcpy(static_cast<void *>(&size), static_cast<void *>(ptr) - 16, sizeof(size));
+    return size;
+}
+
+void raw_free(void *ptr) {
+    size_t size = raw_size(ptr);
+    munmap(static_cast<void *>(ptr) - 16, size + 16);
+}
+
 void *realloc(void *ptr, size_t size) {
     if (!done_init) {
         if (started_init) {
             // called from within GC_INIT, so we can't call it again.  Fall back to using
-            // sbrk and let it leak.
-            size = (size + 0xf) & ~0xf;
-            void *rv = sbrk(size);
+            // mmap.
+            void *rv = raw_alloc(size);
             if (ptr) {
-                size_t max = reinterpret_cast<char *>(rv) - reinterpret_cast<char *>(ptr);
-                memcpy(rv, ptr, max < size ? max : size); }
+                size_t max = raw_size(ptr);
+                memcpy(rv, ptr, max < size ? max : size);
+                raw_free(ptr); }
             return rv;
         } else {
             started_init = true;
@@ -87,9 +104,10 @@ void *realloc(void *ptr, size_t size) {
     if (ptr) {
         if (GC_is_heap_ptr(ptr))
             return GC_realloc(ptr, size);
-        size_t max = reinterpret_cast<char *>(sbrk(0)) - reinterpret_cast<char *>(ptr);
+        size_t max = raw_size(ptr);
         void *rv = GC_malloc(size);
         memcpy(rv, ptr, max < size ? max : size);
+        raw_free(ptr);
         return rv;
     } else {
         return GC_malloc(size);
```

Maybe it will solve the macOS issue as well... Regardless it's probably a good idea to remove the use of `sbrk`. Maybe I'll get around to submitting that patch next week.

Issue #2599, opened 2020-11-02
Compiler Bug when calling an extern multiple times with a default argument


jphurl:
here is a minimal reproducer: 
[bug-repro.p4.txt](https://github.com/p4lang/p4c/files/5477611/bug-repro.p4.txt)

Seems like the bug is a result of the default arg. Bug is only thrown if there is more than 1 call to the extern **and** the default arg is used for both calls. Using an overloaded extern instead of the default arg does not have this issue.

jphurl:
the bug:
```
In file: p4c/frontends/p4/typeChecking/typeChecker.cpp:144
Compiler Bug: At this point in the compilation typechecking should not infer new types anymore, but it did.
```

PR #2608, opened 2020-11-11
Default arguments values should be fresh clones

mbudiu-vmw:


mbudiu-vmw:
Without this change when a default argument value is substituted in multiple places the same sub-dag is used. This prevents type inference from inferring a new type for each use.

mbudiu-vmw:
Fixes #2599

frontends/p4/defaultArguments.cpp
```diff
@@ -27,14 +27,18 @@ class TypeNameSubstitutionVisitor : public TypeVariableSubstitutionVisitor {
...
     const IR::Node* preorder(IR::Type_Name* tn) override {
         auto t = typeMap->getTypeType(getOriginal<IR::Type>(), true);
         if (auto tv = t->to<IR::ITypeVar>())
             return replacement(tv, tn)->to<IR::Type>()->getP4Type();
         return tn;
     }
+    // When cloning the value of an argument we want to make a fresh
+    // copy for each new InfInt type, and not share the same one.
+    const IR::Node* postorder(IR::Type_InfInt*) override
+    { return new IR::Type_InfInt(); }
```
jphurl, line 41: Curious: Why do we visit the default arg in the second statement as `Type_InfInt` but in the first statement as `Type_Bits`? This is just what I saw from IR dump

mbudiu-vmw: In general the IR representation is a DAG, and not a graph; some nodes can share children. The node ID is unique, so in the IR dump look for nodes that have the same ID. In type inference InfInt is treated like a type variable, and unified with some other type to infer a width. That's why you want to create different type variables for each instantiation of the default value: so each can be unified independently. 

The DAG representation is subtle and has lead to several bugs so far. It is also related to the visitDagOnce flag of the visitor, which indicates whether shared nodes are visited only once or once for each parent. One has to keep in mind that the IR is a DAG, and whether shared nodes represent the "same" value really depends on the transformations that you are applying.

jphurl (approved):
LGTM, though I do not have write permissions. Thank you for fixing this and the detailed explanation!

ChrisDodd (approved):

Issue #2583, opened 2020-10-13
simplifyDefUse does not remove redundant assignments to slices 


jphurl:
The following assignments are reduced by `simplifyDefUse` pass:
```p4
h.data = 8;
h.data = 8;
```
to just:
```p4
h.data = 8;
```
but the same is not true if the lhs is a slice:
```p4
h.data[7:0] = 8;
h.data[7:0] = 8;
```
this does not get simplified by `simplifyDefUse`

here is a test: 
[simplify-slice.p4.txt](https://github.com/p4lang/p4c/files/5373502/simplify-slice.p4.txt)

mbudiu-vmw:
Yes, our def-use analysis is not precise enough to track bits, only whole values. Would be nice to do this someday.

dbrajovic-syrmia:
> Yes, our def-use analysis is not precise enough to track bits, only whole values. Would be nice to do this someday.

If you approve, I would like this issue assigned to me.

mbudiu-vmw:
We value any contributions. However, I have to warn you that I am not particularly proud of the def-use pass implementation. I find it difficult to debug myself. There are in fact two passes: def-use computes reaching definitions, and simplifyDefUse uses this information to remove dead assignments. The interplay between them isn't very nice. Perhaps our IR is too high-level? Modelling the semantics of some operations, like method calls with out arguments, is quite tricky. The pass needs to be interprocedural, because some "methods" (e.g., actions) can have side-effects and can access variables in the environment that are not passed as arguments.

mbudiu-vmw:
Note that this pass runs after the SideEFfectOrdering pass, which simplifies some even trickier code cases into simpler ones (e.g., method calls where an argument is another method call that can have side-effects).

dbrajovic-syrmia:
>There are in fact two passes: def-use computes reaching definitions, and simplifyDefUse uses this information to remove dead assignments.

Speaking of SimplifyDefUse and its decision-making to remove dead statements; it seems that for non-slice statements (such as the above `h.data = 8`), the `getReads()` method returns an empty `LocationSet` (in `registerUses()`) which is why the the pass is able to remove the duplicate statement further down the stack. The duplicate doesn't make it to the `HasUses::used` map and for that reason it gets thrown out.

[...]

PR #2610, opened 2020-11-18
Remove redundant slice statements during SimplifyDefUse (#2583)

dbrajovic-syrmia:
* Overwritten slice statements skipped
* Improved logging in FindUninitialized

When FU is registering uses in a lhs slice of an assignment statement, it skips those
which the given slice can overwrite.

To make these changes more comprehensible, I'll start by explaining my view of the problem.

How removing redundant assignment statements works (no-slice lhs assignments):
The issue starts by pointing out that the following
```
h.data = 8;
h.data = 8;
```
gets reduced to
```
h.data = 8;
```
But these snippets are misleading - in reality, the above holds true only when
```
h.data = 8;
h.data = 8;
// a third, certain kind of statement (not necessarily right after the one above) 
```
Why is this third statement required? Because the second `h.data = 8;` would disappear just as well as the first.
This reveals the underlying notion of redundancy the compiler has: 

_Consider all no-slice lhs assignments (with no lvalue on the rhs) as redundant until proven otherwise._

The proof comes from the "third" statement finding use for the second statement (more precisely, it means reading from the same `LocationSet` as `h.data = 8;`). I use the term "floater" to describe a statement that has yet to find its use. Floaters are what is actually getting thrown out during `RemoveUnused` - statements that none other found useful. Similarly, I'll use "anchoring" to describe the moment when floaters find their use. 
 
Every assignment starts out as a floater, but only assignments whose expression(s) can read a `LocationSet`, can anchor a floater (related to the same location). This explains the _assignments with no lvalue on the rhs_ part; lvalues on the rhs read a non-empty location.

Slices are treated differently, in a sense they always read some non-empty `LocationSet` whether they're on the lhs or rhs (why this doesn't apply to no-slice expressions on the lhs is unclear to me).  Consequently, slices always anchor a floater which is why some redundant assignments get to stay.

My solution comes in the form of an additional constraint on the floaters during their time of anchoring done by another slice assignment:

- Is the floater a slice assignment?
- Will the "anchorer" (slice assignment) overwrite the floater?

If both answers are true, the floating slice assignment is not anchored for this particular use (this does **not** mean it won't find its use somewhere else).

### DIFF
--- a/frontends/p4/simplifyDefUse.cpp
+++ b/frontends/p4/simplifyDefUse.cpp
@@ -27,34 +27,74 @@ namespace {
 class HasUses {
     // Set of program points whose left-hand sides are used elsewhere
     // in the program together with their use count
-    std::unordered_map<const IR::Node*, unsigned> used;
+    std::set<const IR::Node*> used;
+
+    class SliceTracker {
+        const IR::Slice* trackedSlice = nullptr;
+        bool active = false;
+        bool overwritesPrevious(const IR::Slice *previous) {
+            if (trackedSlice->getH() >= previous->getH() &&
+                trackedSlice->getL() <= previous->getL())
+                // current overwrites the previous
+                return true;
+
+            return false;
+        }
+
+     public:
+        SliceTracker() = default;
+        explicit SliceTracker(const IR::Slice* slice) :
+            trackedSlice(slice), active(true) { }
+        bool isActive() const { return active; }
+
+        // main logic of this class
+        bool overwrites(const ProgramPoint previous) {
+            if (!isActive()) return false;
+
+            auto last = previous.last();
+            if (auto *assign_stmt = last->to<IR::AssignmentStatement>()) {
+                if (auto *slice_stmt = assign_stmt->left->to<IR::Slice>()) {
+                    // two slice stmts writing to same location
+                    // skip use of previous if it gets overwritten
+                    if (overwritesPrevious(slice_stmt)) {
+                        LOG4("Skipping " << dbp(last) << " " << last);
+                        return true;
+                    }
+                }
+            }
+
+            return false;
+        }
+    };
+
+    SliceTracker tracker;
 
  public:
     HasUses() = default;
     void add(const ProgramPoints* points) {
         for (auto e : *points) {
+            // skips overwritten slice statements
+            if (tracker.overwrites(e)) continue;
+
             auto last = e.last();
             if (last != nullptr) {
                 LOG3("Found use for " << dbp(last) << " " <<
                      (last->is<IR::Statement>() ? last : nullptr));
-                auto it = used.find(last);
-                if (it == used.end())
-                    used.emplace(last, 1);
-                else
-                    used.emplace(last, it->second + 1);
+                used.emplace(last);
             }
         }
     }
-    void remove(const ProgramPoint point) {
-        auto last = point.last();
-        auto it = used.find(last);
-        if (it->second == 1)
-            used.erase(it);
-        else
-            used.emplace(last, it->second - 1);
-    }
     bool hasUses(const IR::Node* node) const
     { return used.find(node) != used.end(); }
+
+    void watchForOverwrites(const IR::Slice* slice) {
+        BUG_CHECK(!tracker.isActive(), "Call to SliceTracker, but it's already active");
+        tracker = SliceTracker(slice);
+    }
+
+    void doneWatching() {
+        tracker = SliceTracker();
+    }
 };
 
 // Run for each parser and control separately
@@ -83,7 +123,7 @@ class FindUninitialized : public Inspector {
             BUG_CHECK(result != nullptr, "no locations known for %1%", dbp(expression));
         return result;
     }
-    // 'expression' is reading the 'loc' location set
+    /// 'expression' is reading the 'loc' location set
     void reads(const IR::Expression* expression, const LocationSet* loc) {
         BUG_CHECK(!unreachable, "reached an unreachable expression in FindUninitialized");
         LOG3(expression << " reads " << loc);
@@ -93,6 +133,7 @@ class FindUninitialized : public Inspector {
     }
     bool setCurrent(const IR::Statement* statement) {
         currentPoint = ProgramPoint(context, statement);
+        LOG3(IndentCtl::unindent);
         return false;
     }
     profile_t init_apply(const IR::Node *root) override {
@@ -132,17 +173,19 @@ class FindUninitialized : public Inspector {
 
     Definitions* getCurrentDefinitions() const {
         auto defs = definitions->getDefinitions(currentPoint, true);
-        LOG3("FU Current point is (after) " << currentPoint << " definitions are " << defs);
+        LOG3("FU Current point is (after) " << currentPoint <<
+                " definitions are " << IndentCtl::endl << defs);
         return defs;
     }
 
     void checkOutParameters(const IR::IDeclaration* block,
                             const IR::ParameterList* parameters,
                             Definitions* defs) {
-        LOG2("Checking output parameters; definitions are " << defs);
+        LOG2("Checking output parameters; definitions are " << IndentCtl::endl << defs);
         for (auto p : parameters->parameters) {
             if (p->direction == IR::Direction::Out || p->direction == IR::Direction::InOut) {
                 auto storage = definitions->storageMap->getStorage(p);
+                LOG3("Checking parameter: " << p);
                 if (storage == nullptr)
                     continue;
 
@@ -170,6 +213,7 @@ class FindUninitialized : public Inspector {
         visit(control->body);
         checkOutParameters(
             control, control->getApplyMethodType()->parameters, getCurrentDefinitions());
+        LOG3("FU Returning from " << control->name << "[" << control->id << "]");
         return false;
     }
 
@@ -234,16 +278,19 @@ class FindUninitialized : public Inspector {
 
         auto outputDefs = acceptdefs->joinDefinitions(rejectdefs);
         checkOutParameters(parser, parser->getApplyMethodType()->parameters, outputDefs);
+        LOG3("FU Returning from " << parser->name << "[" << parser->id << "]");
         return false;
     }
 
     bool preorder(const IR::AssignmentStatement* statement) override {
-        LOG3("FU Visiting " << dbp(statement) << " " << statement);
+        LOG3("FU Visiting " << dbp(statement) << " " << statement << IndentCtl::indent);
         if (!unreachable) {
             lhs = true;
             visit(statement->left);
+            LOG3("FU Returned from " << statement->left);
             lhs = false;
             visit(statement->right);
+            LOG3("FU Returned from " << statement->right);
         } else {
             LOG3("Unreachable");
         }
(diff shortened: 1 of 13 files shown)


mbudiu-vmw:
The def-use analysis is relatively standard. For each statement we compute a conservative approximation of all the locations it will write to (ComputeWriteSet). Then a separate pass computes for each statement the set of locations it reads. Using these and following the program control-flow we can compute the flow of dependences: a statement is useful if it wrote a location set that some other statement reads. 

There are several things that makes this analysis harder:
- we need to do inter-procedural analysis, since a table apply call can invoke actions that can affect variables in scope.
- we need a representation of the location sets. Currently we use abstract memory locations for each base type. A struct is represented as a collection of fields. This is tricky because on procedure invocations new variables may appear.
- writes happen not only on assignments, but also on procedure calls in out arguments.

Some of the corner cases are eliminated by the execution of the SideEffects pass, which removes many constructs that would be very tricky to analyze - e.g., the order of evaluation of arguments that may have side-effects.

Our current representation of location sets does not track bits within a base type; that's why we didn't support removal of redundant slices. 

I will take a look at this PR.

testdata/p4_16_samples_outputs/issue2583_simplify_slice-frontend.p4
```diff
@@ -0,0 +1,20 @@
+#include <core.p4>
+
+header Header {
+    bit<32> data;
+}
+
+parser p(packet_in pckt, out Header h) {
+    state start {
+        h.data = 32w0;
+        h.data = 32w7;
```
jphurl, line 10: shouldn't this be simplified as well?

frontends/p4/simplifyDefUse.cpp
```diff
@@ -553,20 +627,33 @@ class FindUninitialized : public Inspector {
...
         auto fields = storage->getField(expression->member);
+        // LOG3("storage: " << storage << "\nfields: " << fields);
         reads(expression, fields);
+        if (fields == nullptr || fields->isEmpty()) {return false;}
         registerUses(expression);
         return false;
     }
 
-    bool preorder(const IR::Slice* expression) override {
+    bool  preorder(const IR::Slice* expression) override {
```
jphurl, line 637: extra space

jphurl (commented):
Comment on coding style: I think it is best to conform to surrounding style when adding to a file. The main style difference I see is the use of the ending `}` being different than the rest of the file. I think it is best to make the usage consistent

testdata/p4_16_samples_outputs/issue2583_simplify_slice2-frontend.p4
```diff
@@ -0,0 +1,34 @@
...
+            default: next2;
+        }
+    }
+    state next {
+        h.data = 32w8;
+        transition final;
+    }
+    state next2 {
+        h.data[7:2] = 6w9;
+        h.data[7:2] = 6w3;
```
jphurl, line 22: shouldn't this be simplified as well?

frontends/p4/simplifyDefUse.cpp
```diff
@@ -132,17 +180,19 @@ class FindUninitialized : public Inspector {
 
     Definitions* getCurrentDefinitions() const {
         auto defs = definitions->getDefinitions(currentPoint, true);
-        LOG3("FU Current point is (after) " << currentPoint << " definitions are " << defs);
+        LOG3("FU Current point is (after) " << currentPoint <<
```
hanw, line 183: What does the 'FU' acronym stands for? =)

jnfoster: Presumably "Find Uninitialized." But it could easily be confused for something else.

testdata/p4_16_samples_outputs/issue2583_simplify_slice-frontend.p4
dbrajovic-syrmia: Yes, you're right, it should be. But this makes for a great segue into another point I wanted to make. Consider this simplification (snippet):
```
parser p(packet_in pckt, out Header h, out Header h2) {
    state start {
        h.data = 8;  // A
        h2.data = h.data; // B
        h.data = 8; // C
        h2.data = 5; // D
        transition accept;
    }
}

// turns into

parser p(packet_in pckt, out Header h, out Header h2) {
    state start {
        h.data = 32w8;
        ;
        h.data = 32w8;
        h2.data = 32w5;
        transition accept;
    }
}
```
Looks familiar. Pardon the nonsensical piece of code (hard to imagine anyone writing this), but it does expose the inner workings of the pass. Despite the lack of slices, even after the simplifying, there still seems to be a redundant statement or two. To understand why, let's remind ourselves what `RemoveUnused` actually does.

Conceptually, it's intended to skim through all the assignments in the program and remove those not found in the `HasUses::used` map. Using the terminology from before, this means "remove all the floaters who aren't anchored".
But what if the floater being removed was anchoring some earlier statement (previously a floater) ? That's exactly what happened in both examples.

C and D are saved thanks to `FindUninitialized::checkOutParameters`, so they get to stay (I won't explain how, it's not relevant to the point). B gets thrown out due to having no anchor later on (C doesn't do this, no `LocationSet` to read from), but **statement A stays because statement B was its anchor**. The `h.data` on the rhs of B makes sure of this anchoring. 

So, `RemoveUnused` does what it's told and in effect, kills off the only anchor (use) statement A was having. Thus A becomes an _orphaned_ floater - a useful statement that became redundant once we removed the (known) redundant statements.

I am not entirely sure, but there may be a workaround to this. The simplest solution is to run this pass twice, but this might not scale well (how many times is enough? maybe some recursive strategy might help). 

The core problem with this pass (I think) is that we can't detect all the redundant statements in one go. We need `RemoveUnused` to expose the orphaned floaters and then remove them or find a way to detect them before we reach the removing stage.

mbudiu-vmw: SimplifyDefUse currently removes unused assignments, but it does not remove transitively unused assignments.
This can be fixed simply by converting the pass from a PassManager to a PassRepeated, and then it will be executed until a fixed-point is reached.

dbrajovic-syrmia:
I suppose there are a few things that should be addressed given the latest commit. @mbudiu-vmw 's suggestion of converting the pass to `PassRepeated` does exactly what's expected given what was discussed. But in turn, as indicated by the dozen of tests whose outputs I've changed, it brings about some other possible "issues" worth mentioning.

The most obvious one is the duplicated outputs of this pass. Since the pass could run several times, warnings about (for example ) uninitialized values come up more than once. While not the most daunting difference, it certainly is a nuisance.

The other issue is the significant difference in the `-frontend`, `-midend` outputs. In these files a lot of constructs get removed because they are considered `transitively unused`. Due to my lack of understanding of what these tests were designed for (as well as the P4 compiler in general), I'm unable to decide whether these changes break the tests' functionality or not.

mbudiu-vmw:
We have some mechanism that should prevent the same error from being emitted twice. Does it fail in this case?
Also, if some reference outputs change, the workflow is to check them manually for being correct, and then to commit the new versions together with the PR. The rule is that the emitted program must have the same semantics as the original one.

jafingerhut:
Given the number of changed outputs, hopefully we can get some help (or have already gotten some help from, part of Travis / Github Actions test results?) from the gauntlet formal verification tool.

Given the number of assignments that were deleted, perhaps some of those test programs might achieve better test coverage if more of their outputs were used, and thus could not be correctly deleted?

Just as one example that I spot-checked, this test output https://github.com/p4lang/p4c/pull/2610/files#diff-d92c92626088f62e18df7a4f6624b324702b38fead1e57f0bcab4994fc8519c8 seems much shorter than it was before, and while it may be correct, it also seems to no longer contain some enum casts, which from the name of the test program I would guess were a motivation for creating the test program in the first place.

mbudiu-vmw:
We should make these two changes in separate PRs.
We should have just one PR that removes recursive unused statements so we can validate that independently.

mbudiu-vmw (approved):
This seems sound.

frontends/p4/simplifyDefUse.cpp
```diff
@@ -554,19 +634,31 @@ class FindUninitialized : public Inspector {
 
         auto fields = storage->getField(expression->member);
         reads(expression, fields);
+        if (fields == nullptr || fields->isEmpty()) {return false;}
```
mbudiu-vmw, line 637: why this change?

mbudiu-vmw (commented):
I will suggest the following changes nevertheless:
- let's do the PassRepeated change in a separate PR
- As you noticed, HasUses has a useless count; please change the map to a set and delete the unused `remove` method 
- let's improve a bit the logging appearance, please replace `std::endl` in def_use.* and simplifyDefUse.* with `IndentCtl::endl`. Then change the `LOG` calls for all statements in simplifyDefUse as follows: ` LOG3("FU Visiting " << dbp(statement) << " " << statement << IndentCtl::indent);` Change `setCurrent` by adding a `LOG3(IndentCtl::unindent);`
- In the next PR, when we add PassRepeated, we have to add some support so that no warning is printed twice, probably by maintaining a set of the warnings already given.

mbudiu-vmw:
Also, can you see whether repeating this pass makes the compilation of the large programs (switch.p4) longer? A concern was that this analysis looks expensive.

frontends/p4/simplifyDefUse.cpp
dbrajovic-syrmia: Apologies in advance, it is a leftover from when I first started out examining the pass; to clear up the logs. I figured if a LocationSet was empty, no uses could be registered (at least in my test examples).

dbrajovic-syrmia:
> Just as one example that I spot-checked, this test output https://github.com/p4lang/p4c/pull/2610/files#diff-d92c92626088f62e18df7a4f6624b324702b38fead1e57f0bcab4994fc8519c8 seems much shorter than it was before, and while it may be correct, it also seems to no longer contain some enum casts, which from the name of the test program I would guess were a motivation for creating the test program in the first place.

Yes, that specific test was the reason I wanted to address the `PassRepeated` change. Another interesting note about that test is when you compare the `-first` and `-frontend` outputs (even before this PR), the very statements the test was founded upon get removed during the simplifying (only one run).  Here's what I mean:
https://github.com/p4lang/p4c/blob/f0b37d9a26151386f3b0011f97a1eaafdd28c1ad/testdata/p4_16_samples_outputs/enumCast-first.p4#L44
https://github.com/p4lang/p4c/blob/f0b37d9a26151386f3b0011f97a1eaafdd28c1ad/testdata/p4_16_samples_outputs/enumCast-frontend.p4#L42
So, when SimplifyDefUse is converted to a PassRepeated, it picks up where it left off, which is why I was uncertain about all the other tests as well. 

> I will suggest the following changes nevertheless:
> 
> 
>     * As you noticed, HasUses has a useless count; please change the map to a set and delete the unused `remove` method
> 
>     * let's improve a bit the logging appearance, please replace `std::endl` in def_use.* and simplifyDefUse.* with `IndentCtl::endl`. Then change the `LOG` calls for all statements in simplifyDefUse as follows: ` LOG3("FU Visiting " << dbp(statement) << " " << statement << IndentCtl::indent);` Change `setCurrent` by adding a `LOG3(IndentCtl::unindent);`
> 

Fortunately, I've kept around the commit representing the original change (without the PassRepeated). I've reset this branch to that commit as well as added the requested improvements above. The logs are much more readable now.

> Also, can you see whether repeating this pass makes the compilation of the large programs (switch.p4) longer? A concern was that this analysis looks expensive.

PassManager: `bmv2/testdata/p4_14_samples/switch_20160512/switch.p4 .........................   Passed   83.93 sec`
PassRepeated: `bmv2/testdata/p4_14_samples/switch_20160512/switch.p4 .........................   Passed   89.44 sec`

Both tests were run with `make check -j16` on my virtual machine (Ubuntu 18).

Regarding the second PR, for clarification, did you mean including both the slice logic _and_ converting SimplifyDefUse to PassRepeated or just the latter?

PR #2619, opened 2020-11-27
Handle Type_Dontcare args when creating Counterlike instances

boblytton:
The added gtest P4Runtime.ArgTypeDontCare demonstrates the problem viz:
```
meter<_>(1024, MeterType.bytes) mtr;  // Use `result` argument to infer type.
```
without the fix fails with:
```
error: Unexpected type _
            meter<_>(1024, MeterType.bytes) mtr;
```
Please advise as to the correctness of the fix, and the location of the gtest.
Thank you.

### DIFF
--- a/control-plane/p4RuntimeArchHandler.h
+++ b/control-plane/p4RuntimeArchHandler.h
@@ -414,11 +414,13 @@ struct Counterlike {
             BUG_CHECK(type->arguments->size() > *indexTypeParamIdx,
                       "%1%: expected at least %2% type arguments",
                       instance, *indexTypeParamIdx + 1);
-            auto typeArg = type->arguments->at(*indexTypeParamIdx);
-            // We ignore the return type on purpose, but the call is required to update p4RtTypeInfo
-            // if the index has a user-defined type.
-            TypeSpecConverter::convert(refMap, typeMap, typeArg, p4RtTypeInfo);
-            index_type_name = getTypeName(typeArg, typeMap);
+            const IR::Type* typeArg = type->arguments->at(*indexTypeParamIdx);
+            if (!typeArg->is<IR::Type_Dontcare>()) {
+                // We ignore the return type on purpose, but the call is required to update p4RtTypeInfo
+                // if the index has a user-defined type.
+                TypeSpecConverter::convert(refMap, typeMap, typeArg, p4RtTypeInfo);
+                index_type_name = getTypeName(typeArg, typeMap);
+            }
         }
 
         return Counterlike<Kind>{declaration->controlPlaneName(),
--- a/control-plane/p4RuntimeArchHandler.h
+++ b/control-plane/p4RuntimeArchHandler.h
@@ -416,8 +416,8 @@ struct Counterlike {
                       instance, *indexTypeParamIdx + 1);
             const IR::Type* typeArg = type->arguments->at(*indexTypeParamIdx);
             if (!typeArg->is<IR::Type_Dontcare>()) {
-                // We ignore the return type on purpose, but the call is required to update p4RtTypeInfo
-                // if the index has a user-defined type.
+                // We ignore the return type on purpose, but the call is required to
+                // update p4RtTypeInfo if the index has a user-defined type.
                 TypeSpecConverter::convert(refMap, typeMap, typeArg, p4RtTypeInfo);
                 index_type_name = getTypeName(typeArg, typeMap);
             }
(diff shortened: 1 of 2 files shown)


control-plane/p4RuntimeArchHandler.h
```diff
@@ -414,11 +414,13 @@ struct Counterlike {
             BUG_CHECK(type->arguments->size() > *indexTypeParamIdx,
                       "%1%: expected at least %2% type arguments",
                       instance, *indexTypeParamIdx + 1);
-            auto typeArg = type->arguments->at(*indexTypeParamIdx);
-            // We ignore the return type on purpose, but the call is required to update p4RtTypeInfo
-            // if the index has a user-defined type.
-            TypeSpecConverter::convert(refMap, typeMap, typeArg, p4RtTypeInfo);
-            index_type_name = getTypeName(typeArg, typeMap);
+            const IR::Type* typeArg = type->arguments->at(*indexTypeParamIdx);
+            if (!typeArg->is<IR::Type_Dontcare>()) {
```
mbudiu-vmw, line 418: Is this the right fix? index_type_name will be nullptr. 
Ideally you would use the inferred value for the type.
Perhaps the compiler should replace Type_Dontcare with the inferred value.

boblytton: Thank you for the comments.
Looking at `RemoveDontcareArgs` it is handling `IR::MethodCallExpression` but not `IR::Declaration_Instance`.
I will amend the PR by handling the need there.

mbudiu-vmw:
This is not so simple, there may be not enough info here.

control-plane/p4RuntimeArchHandler.h
boblytton: > Ideally you would use the inferred value for the type
Does this mean using the `TypeInference` class?
It is unclear to me if the `Type_Dontcare` should have been already inferred, or if the `TypeSpecConverter::convert()`  viz `TypeSpecConverter::preorder(const IR::Type* type)` should be using `TypeInference`.

boblytton:
> This is not so simple, there may be not enough info here.

Given the attached failing unit-test, who would be a good person to discuss a fix with?

mbudiu-vmw:
Let me take a look at this.

mbudiu-vmw:
If I try to run this test as is with the sources from the repository it fails because the version of v1model used is too old - and it uses the meter without any type arguments. So you must have some other changes in your environment too.

mbudiu-vmw:
Actually, I take it back, as a gtest it does compile fine.

mbudiu-vmw:
I am afraid that this uncovers a different bug in the compiler. In fact, the P4 program you supplied should be rejected: the compiler does not look at the invocation of execute_meter to infer the type in the declaration. The declaration itself is used to infer types, and in this case it does not provide enough information to infer the type of the index. As a proof that this is wrong, the compiler even accepts the following program:
```
    bit<8> index = 0;
    bit<8> result = 0;
    bit<16> index1 = 10;
    meter<_>(1024, MeterType.bytes) mtr;  // Use `result` argument to infer type.
    apply {
        mtr.execute_meter(index, result);
        mtr.execute_meter(index1, result);
    }
```

So what you will have to do is to explicitly specify the index type when you instantiate the meter. The compiler can infer the index type only if it is used somewhere in the instantiation itself.

mbudiu-vmw:
I have filed https://github.com/p4lang/p4c/issues/2623 to track this.
It would be nice to have the compiler infer the value of the don't care type, but for now we'll have to require the user to specify it. If you think this is an important feature you can file an issue for a feature request, but I suspect it won't be a high priority issue for a while.

vgurevich:
@mbudiu-vmw ,

I think that an important use case is when a constant is used as an index, especially in a case when there is really only one (or very few instances). For example:

```
Meter<_>(2, MeterType_t.BYTES) byte_meter;
...
apply {
    if (hdr.ipv4.isValid()) {
        color = byte_meter.execute(0);
    } else {
        color = byte_meter.execute(1);
    }
}
```

This is the case, where one might really want to leave the choice to the compiler.

mbudiu-vmw:
"Don't care" for types is meant to indicate that the type is not really used, not that it should be inferred.

boblytton:
> "Don't care" for types is meant to indicate that the type is not really used, not that it should be inferred.

I wonder how many programmers read `_` as    "Don't care because I don't use it"
and how many read it as     "leaving it for argument dependant lookup - keep things generic and 'DRY' "
I guess the later are mostly modern C++ programmers, so a small audience! :-)

jnfoster:
I would expect a don't-care type to work the same way as a completely fresh type variable...

@mbudiu-vmw is that not the case?

PR #2620, opened 2020-11-29
Add INode::checkedTo for checked casts.

liujed:
Added a convenience method `INode::checkedTo` for checked casts. The effect of `n->checkedTo<N>()` is similar to `&n->as<N>()`, except the former produces an error message that is easier to debug. The former also has clearer intent: the latter could be easily mistaken as being equivalent to `n->to<N>()`.


mbudiu-vmw (approved):

liujed:
So, I've signed the CLA. Do I need to push a trivial change, or is there some other way to re-trigger the CLA check?

liujed:
My CLA signature doesn't seem to be recognized.

jnfoster:
The CLA check is advisory at this point AFAIK

liujed:
Ah, indeed, it is. Regardless, I think I've sorted out the issue. Didn't notice I had to "Sign Addendum" after adding my GitHub account to my agreement.

PR #2621, opened 2020-11-30
Convert SimplifyDefUse to a PassRepeated (#2610)

dbrajovic-syrmia:
This (second) PR is a continuation of the discussion in https://github.com/p4lang/p4c/pull/2610.
Originally suggested as part of the solution, it was decided to become a separate concern.
As a result, the SimplifyDefUse pass runs until convergence, instead of only once.

### DIFF
--- a/testdata/p4_16_samples_outputs/issue2221-bmv2-frontend.p4
+++ b/testdata/p4_16_samples_outputs/issue2221-bmv2-frontend.p4
@@ -23,17 +23,10 @@ parser p(packet_in pkt, out Headers hdr, inout Meta m, inout standard_metadata_t
 }
 
 control ingress(inout Headers h, inout Meta m, inout standard_metadata_t sm) {
-    @name("ingress.tmp") bit<16> tmp;
     @name("ingress.tmp_0") bit<16> tmp_0;
-    @name("ingress.tmp_1") bit<16> tmp_1;
-    @name("ingress.tmp_2") bit<16> tmp_2;
     @name("ingress.tmp_3") bit<16> tmp_3;
-    @name("ingress.tmp_4") bit<16> tmp_4;
-    @name("ingress.tmp_5") bit<16> tmp_5;
-    @name("ingress.tmp_6") bit<16> tmp_6;
-    @name("ingress.tmp_7") bit<16> tmp_7;
+    @name("ingress.tmp_6") bit<16> tmp_4;
     apply {
-        tmp = 16w0;
         {
             @name("ingress.eth_type_0") bit<16> eth_type_0 = h.eth_hdr.eth_type;
             @name("ingress.hasReturned") bool hasReturned = false;
@@ -44,8 +37,6 @@ control ingress(inout Headers h, inout Meta m, inout standard_metadata_t sm) {
             h.eth_hdr.eth_type = eth_type_0;
             tmp_0 = retval;
         }
-        tmp_1 = tmp & tmp_0;
-        tmp_2 = 16w0;
         {
             @name("ingress.eth_type_1") bit<16> eth_type_1 = h.eth_hdr.eth_type;
             @name("ingress.hasReturned") bool hasReturned_1 = false;
@@ -56,8 +47,6 @@ control ingress(inout Headers h, inout Meta m, inout standard_metadata_t sm) {
             h.eth_hdr.eth_type = eth_type_1;
             tmp_3 = retval_1;
         }
-        tmp_4 = tmp_2 * tmp_3;
-        tmp_5 = 16w0;
         {
             @name("ingress.eth_type_2") bit<16> eth_type_2 = h.eth_hdr.eth_type;
             @name("ingress.hasReturned") bool hasReturned_2 = false;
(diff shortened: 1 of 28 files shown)


testdata/p4_16_samples_outputs/issue2221-bmv2-frontend.p4
```diff
@@ -66,9 +55,8 @@ control ingress(inout Headers h, inout Meta m, inout standard_metadata_t sm) {
             hasReturned_2 = true;
             retval_2 = 16w2;
             h.eth_hdr.eth_type = eth_type_2;
-            tmp_6 = retval_2;
+            tmp_4 = retval_2;
```
mbudiu-vmw, line 58: Why isn't this one removed too?

testdata/p4_16_samples_outputs/issue561-frontend.p4
```diff
@@ -16,9 +16,8 @@ package top(ct _ct);
 control c(out bit<32> x) {
     @name("c.u") U u_0;
     @name("c.u2") U[2] u2_0;
-    @name("c.b") bool b_0;
     apply {
-        b_0 = u_0.isValid();
+        u_0.isValid();
```
mbudiu-vmw, line 20: Please file an issue so that we add code to detect such method calls that have no effect and remove them too. It's funny that we did that in some places for assignments, but not just for method calls.

testdata/p4_16_samples_outputs/stack-frontend.p4
```diff
@@ -5,26 +5,23 @@ header h {
...
     @name("p.stack") h[4] stack_0;
-    @name("p.b") h b_0;
     state start {
         stack_0[0].setInvalid();
         stack_0[1].setInvalid();
         stack_0[2].setInvalid();
         stack_0[3].setInvalid();
         stack_0[3].setValid();
-        b_0 = stack_0.last;
-        stack_0[2] = b_0;
```
mbudiu-vmw, line 16: Why is this statement eliminated? We have to make sure this is not a bug.

testdata/p4_16_samples_outputs/uninit-frontend.p4
```diff
@@ -33,12 +31,9 @@ parser p1(packet_in p, out Header h) {
         }
     }
     state next1 {
-        d_0 = false;
         transition next3;
     }
     state next2 {
-        c_0 = true;
-        d_0 = c_0;
```
mbudiu-vmw, line 41: please check also that these assignments are truly dead.

mbudiu-vmw (approved):
I will approve this, but I will wait to merge until we check that all reference outputs are indeed correct.

testdata/p4_16_samples_outputs/stack-frontend.p4
fruffy: I checked this out and it looks like this happens because the header stack is unused overall. Attached is the same example but with output that matters. In this case the assignment is not removed. Concerning to the comment above, the `set(In)Valid` calls could also be removed here? Btw Gauntlet does not pick up changes in dead code. This can be a blind spot for some of these tests.

Modified `stack.p4`: 
```p4
#include <core.p4>

header H {
    bit<8> a;
}


struct Headers {
    H[4] stack;
}

parser p(packet_in pkt, out Headers h) {
    state start {
        H[4] stack;
        stack[3].setValid();
        H b = stack[3];
        b = stack.last;
        stack[2] = b;
        b = stack.next;
        bit<32> e = stack.lastIndex;
        h.stack = stack;
        transition accept;
    }
}

    apply {
        H[4] stack;
        stack[3].setValid();
        H b = stack[3];
        stack[2] = b;
        stack.push_front(2);
        stack.pop_front(2);
        bit<32> sz = stack.size;
        h.stack = stack;
    }
}

parser Parser(packet_in b, out Headers hdr);
control Ingress(inout Headers hdr);
package top(Parser p, Ingress ig);
top(p(), ingress()) main;
```

mbudiu-vmw: Calls like setInvalid have side-effects, and we do not have any analysis which would remove calls with side-effects. We do not understand currently the fact that the side-effects are confined just to an unused variable.

testdata/p4_16_samples_outputs/issue2221-bmv2-frontend.p4
dbrajovic-syrmia: I believe this is a rename of the temporary variable. Just above that line:
`@name("ingress.tmp_6") bit<16> tmp_4;`

testdata/p4_16_samples_outputs/issue561-frontend.p4
dbrajovic-syrmia: I've opened up a new issue: https://github.com/p4lang/p4c/issues/2628

testdata/p4_16_samples_outputs/uninit-frontend.p4
dbrajovic-syrmia: As far as I can tell (mostly by looking at the output), they do seem truly dead.

PR #2622, opened 2020-12-01
remove duplicated library function in favor of big_int version

hanw:
The implementation in bitops.h only supports 32bit unsigned int.


ChrisDodd (approved):
I guess this means that when these functions are called on an unsigned, that value will first be converted to a big_int, then will do the calculation with that?  So less efficient, and means we don't ever use the __builtin functions?  I suppose the difference is not measurable?  I guess if we ever do notice the performance, adding back optimized versions would be easy enough.

PR #2627, opened 2020-12-08
Remove generic types after specialization

mbudiu-vmw:
This fixes a bug found as a result of reviewing this design in the language design working group.
After specializing generic structs the original type definitions are removed from the program so not further passes have to worry about them. Without these passes the tuple elimination pass generates incorrect code.

### DIFF
--- a/frontends/p4/specializeGenericTypes.h
+++ b/frontends/p4/specializeGenericTypes.h
@@ -149,10 +149,10 @@ struct S0 { bit<32> data; }
 S0 s;
 ```
  */
-class SpecializeGenericTypes : public PassRepeated {
+class DoSpecializeGenericTypes : public PassRepeated {
     TypeSpecializationMap specMap;
  public:
-    SpecializeGenericTypes(ReferenceMap* refMap, TypeMap* typeMap) {
+    DoSpecializeGenericTypes(ReferenceMap* refMap, TypeMap* typeMap) {
         passes.emplace_back(new PassRepeated({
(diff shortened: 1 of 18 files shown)


mbudiu-vmw:
This is a tiny PR; it essentially removes all types that are generic - at that point they must have been replaced with non-generic specialized versions. For example, struct S<T> { T x; } should not be needed anymore. For example S<bit<32>> should be replaced with another struct S_0 { bit<32> x; }.

mbudiu-vmw:
This pass only deletes the declaration of S<T>.

frontends/p4/specializeGenericTypes.h
```diff
@@ -166,6 +166,27 @@ class SpecializeGenericTypes : public PassRepeated {
...
         specMap.refMap = refMap;
         specMap.typeMap = typeMap;
+        setName("DoSpecializeGenericTypes");
+    }
+};
+
+class RemoveGenericTypes : public Transform {
+ public:
+    const IR::Node* postorder(IR::Type_StructLike* type) override {
+        if (!type->typeParameters->empty()) return nullptr;
```
fruffy, line 176: There is a nullpointer dereference according to the tests. I am guessing because the type is now null because of this?

mbudiu-vmw: In fact this is a bug in the type checker which does not handle some expressions. I will add a commit to this PR to fix that.

fruffy (approved):

PR #2631, opened 2020-12-10
Add Two New Keywords to P4-16 for Modularity

hesingh:
This is work in progress and when all is well, I will let folks know.

What is pending:

1. Fix 4-5 failing tests with `make check`.
2. Fix one TODO due to new code in yacc.
3. Add tests to prove modularity works.  I have large P4 programs which have been tested but I have to pare them down for this PR because the programs are proprietary to my company.


frontends/common/options.cpp
```diff
@@ -76,12 +76,14 @@ CompilerOptions::CompilerOptions() : Util::Options(defaultMessage) {
                        return true; },
                     "[Deprecated; use --std instead] Specify language version to compile.",
                     OptionFlags::Hide);
-    registerOption("--std", "{p4-14|p4-16}",
+    registerOption("--std", "{p4-14|p4-16|p4-18}",
```
rst0git, line 79: I'm aware that this is still work in progress, but as far as I know `14` and `16` was meant to represent the year in which the P4 language specification was developed: https://youtu.be/GslseT4hY1w?t=651

Is there going to be P4_18 language specification?

hesingh: I am not aware of any p4-18 specification on the horizon.  The p4-18 flag was a convenience to tell p4c that a P4 program is using new `override` and `super` keywords and also tell p4c to generate merged output p4-16 program.   I updated README.md to show how `ptest` is used.

Later, I can add an explicit flag or do something else.

Regarding code review, the few tests failing with this code also fail for me with a p4c without these changes.  So, this PR is not an issue with `make check`.  We can get a head start on code review.  The guts of the changes are in `parserDriver.h` and `p4parser.ypp`. thanks

mbudiu-vmw: > Is there going to be P4_18 language specification?

Certainly not. Not even a P4_20. So far we have only incrementally extended P4_16 remaining backwards-compatible, so there was no need for changing the name of the language. I personally hope that we can stick with P4_16 for a long time.

jnfoster: Mihai and I haven't colluded, I swear, but I'm also not a big fan of the years numbering. I think we'll stick with semantic versioning for a while...

README.md
```diff
@@ -490,3 +490,27 @@ access them from the IR
 * Tables with multiple apply calls
 
 See also [unsupported P4_16 language features](backends/bmv2/README.md#unsupported-p4_16-language-features).
+
+## P4-16 Modularity
```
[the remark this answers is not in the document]
mbudiu-vmw: On the contrary, I would suggest have a more elaborate commit message, or comment on this PR, which describes the language extensions and how they work.

frontends/parsers/p4/p4parser.ypp
```diff
@@ -664,12 +666,46 @@ packageTypeDeclaration
 
 instantiation
       : annotations typeRef "(" argumentList ")" name ";"
-                     { $$ = new IR::Declaration_Instance(@6, *$6, new IR::Annotations(*$1),
-                                                         $2, $4);
+                     {
```
mbudiu-vmw, line 669: your implementation is done completely in the bison parser. I would actually suggest constructing an appropriate IR representation of the entire program, including the overridden objects, and then having a compiler pass do this elaboration.

mbudiu-vmw (changes requested):

frontends/parsers/p4/p4parser.ypp
hesingh: The bison implementation is efficient (fastest compiling) by preparing final IR as two programs are being parsed. 
However, for compiling on host machines, code maintenance is key, so why complicate the yacc code?  The bison changes are simple for all, but one case, by checking for a parent before adding an object to IR.  It's only the code that changed to handle `parserState` which is more complicated, but it's only one code object. 

I can document the two new keywords for full scope of their use.  Thereafter, we can decide to use bison or a frontend pass.  If it's a frontend pass, I won't have cycles to work on it anytime soon.

hesingh: > your implementation is done completely in the bison parser. I would actually suggest constructing an appropriate IR representation of the entire program, including the overridden objects, and then having a compiler pass do this elaboration.

Preparing a representation of the entire program in bison for processing later in a frontend pass is easier said than done.  Bison adds IR objects to the symbol table.  The base code is always processed first (modularity entails including base-code.p4) in a P4 program and then writing a new P4 program.  During base code processing, say, `struct foo {}` is added to the symbol table. Then `struct foo {} override` is processed. You cannot add another `foo` to symbol table.  So now bison gets complicated by not adding override and super objects to symbol table.  By the time bison prepares a representation for both programs, most of the bison code in this PR ends up getting used because one has to store the representation of overriden and supered information somewhere.

One could rename `foo` to `foo_ov` during processing override and then add `foo_ov` to symbol table.  But, when the frontend pass processes the representation, the pass needs to remove `foo_ov` from the symbol table.

PR #2632, opened 2020-12-10
Set idle timeout behavior based on table property in PSA 

amreshk:
For p4info generation on PSA, the idle timeout is always set to `NOTIFY_CONTROL` which shows up for tables which dont have any 'psa_idle_timeout' property on the table. PR checks for the property and sets the value accordingly.

### DIFF
--- a/control-plane/p4RuntimeArchStandard.cpp
+++ b/control-plane/p4RuntimeArchStandard.cpp
@@ -991,8 +991,14 @@ class P4RuntimeArchHandlerPSA final : public P4RuntimeArchHandlerCommon<Arch::PS
                             const IR::TableBlock* tableBlock) override {
         P4RuntimeArchHandlerCommon<Arch::PSA>::addTableProperties(
             symbols, p4info, table, tableBlock);
-        // TODO(antonin): not supported yet in PSA
-        table->set_idle_timeout_behavior(p4configv1::Table::NOTIFY_CONTROL);
+
+        auto tableDeclaration = tableBlock->container;
+        bool supportsTimeout = getSupportsTimeout(tableDeclaration);
+        if (supportsTimeout) {
+            table->set_idle_timeout_behavior(p4configv1::Table::NOTIFY_CONTROL);
+        } else {
+            table->set_idle_timeout_behavior(p4configv1::Table::NO_TIMEOUT);
+        }
     }
 
     void addExternInstance(const P4RuntimeSymbolTableIface& symbols,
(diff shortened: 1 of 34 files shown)


control-plane/p4RuntimeArchStandard.cpp
```diff
@@ -1024,6 +1030,31 @@ class P4RuntimeArchHandlerPSA final : public P4RuntimeArchHandlerCommon<Arch::PS
...
     }
+
+    /// @return true if @table's 'support_timeout' property exists and is true. This
+    /// indicates that @table supports entry ageing.
+    static bool getSupportsTimeout(const IR::P4Table* table) {
+        auto timeout = table->properties->getProperty("psa_idle_timeout");
+
+        if (timeout == nullptr) return false;
+
+        if (auto exprValue = timeout->value->to<IR::ExpressionValue>()) {
```
mbudiu-vmw, line 1041: This will only work if the expression is a compile-time known value (i.e., constant).
Is this documented? Perhaps the error message should say that.

mbudiu-vmw (approved):

control-plane/p4RuntimeArchStandard.cpp
amreshk: @mbudiu-vmw, we do have usage mentioned in the psa spec,

https://p4.org/p4-spec/docs/PSA-v1.1.0.html

> 7.2.1. Table entry timeout notification
> PSA uses the psa_idle_timeout to enable a table implementation send notifications from the PSA device when a configurable time has passed since an entry was last matched. The property may take one of two values – NO_TIMEOUT, and NOTIFY_CONTROL. 

In `psa.p4` this is represented as an enum,
https://github.com/p4lang/p4c/blob/84051411d31a39d5670b4721f438fa2f3d232d60/p4include/psa.p4#L263-L267

and will show up as,

```
[268478] ExpressionValue
  expression: [268479] Member member=NOTIFY_CONTROL
    type: [919] Type_Enum name=PSA_IdleTimeout_t declid=36
      annotations: [2] Annotations
      [916] Declaration_ID name=NO_TIMEOUT declid=70
      [917] Declaration_ID name=NOTIFY_CONTROL declid=71
    expr: [268480] TypeNameExpression
      type: [253042] Type_Type
        type: [919] Type_Enum name=PSA_IdleTimeout_t declid=36...
      typeName: [2481] Type_Name
        path: [2480] Path name=PSA_IdleTimeout_t absolute=0
```

I can explicitly add the allowed values in the error message.

mbudiu-vmw: How about this case: 
```
PSA_IdleTimeout_t v = PSA_IdleTimeout_t.NO_TIMEOUT;
// v could also come from an external function call, or some other dynamic expression.
table t {
   ... 
    psa_idle_timeout = v;
}
```

amreshk: @mbudiu-vmw , your example results in a `IR::PathExpression`

```
[191647] ExpressionValue
  expression: [191648] PathExpression
    type: [919] Type_Enum name=PSA_IdleTimeout_t declid=36
      annotations: [2] Annotations
      [916] Declaration_ID name=NO_TIMEOUT declid=70
      [917] Declaration_ID name=NOTIFY_CONTROL declid=71
    path: [147079] Path name=v_0/v absolute=0

```
Since this is not resolved, I am not sure how to handle it here.

mbudiu-vmw: For this case you should just give an error requiring the use of a constant value.

jafingerhut:
@amreshk Out of curiosity, have you looked at the bmv2 JSON files produced by the compiler before and after your changes (or does it produce a BMv2 JSON file at all for your new PSA test program)?

Do you know if it fills in the idle timeout for tables so that bmv2 recognizes and uses it?

amreshk:
@jafingerhut , my changes will reflect in any p4info generation for PSA. I see quite a few PSA programs had the idle time out behavior set by default which is no longer the case.

For BMV2, do we support psa-bmv2 target? I tried running this on my example but it doesnt seem to be the case.

amreshk:
@jafingerhut , i was able to generate JSON and p4info for psa-bmv2 using `./p4c-bm2-psa`. There is no change in json and this is reflective of the input p4 with / without my change. However, p4info output will have idle time behavior property only when specified on the p4 table after my changes.

control-plane/p4RuntimeArchStandard.cpp
amreshk: Added an error message for this case.

PR #2634, opened 2020-12-14
Add -o option to p4test backend

hesingh:
p4c backends in bmv2, ebpf, and ubpf support the `-o` option.  It's only the p4test backend that does not.  P4 modularity is one example of an app which would use the `-o` option to dump the merged P4 program.   This PR moves the option to `frontend/common/options.h[.cpp]`.

### DIFF
--- a/lib/options.cpp
+++ b/lib/options.cpp
@@ -40,14 +40,15 @@ void Util::Options::registerOption(const char* option, const char* argName,
 
 void Util::Options::unRegisterOption(const char* option) {
     auto it = options.find(option);
-    auto itv = find (optionOrder.begin(), optionOrder.end(), option);
-    if (it != options.end()) {
-        auto o = it->second;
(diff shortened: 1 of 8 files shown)


lib/options.cpp
```diff
@@ -38,6 +38,18 @@ void Util::Options::registerOption(const char* option, const char* argName,
     optionOrder.push_back(option);
 }
 
+void Util::Options::unRegisterOption(const char* option) {
+    auto it = options.find(option);
+    auto itv = find (optionOrder.begin(), optionOrder.end(), option);
+    if (it != options.end()) {
```
mbudiu-vmw, line 44: probably this should be a BUG_CHECK.

mbudiu-vmw (changes requested):
cpplint is not happy about your program.

mbudiu-vmw (approved):

PR #2637, opened 2020-12-18
Tweak type checking to accomodate generic functions that handle generic structs

### LINKED ISSUE #2635: p4c should generate warning for unused type in generic struct
I have the following generic struct.  N is not used by the struct.  p4c does not catch the unused N.

```p4
struct S<T, N> {
    T field;
    bit<32> otherField;
}
```

### LINKED ISSUE #2636: generic functions do not work with generic structs
```
struct S<T> {
    T field;
}

extern S<T> f<T>();

control c(out bit<32> o) {
    apply {
        S<bit<32>> s;
        s = f<bit<32>>();
        o = s.field;
    }
}
```

mbudiu-vmw:
The bulk of this PR tweaks the typechecker to use a flag `strictStruct` which defines how type equality for structs is defined. With `strictStruct=true` two structs are equal if they are the same struct. With `strictStruct=false` they are equal only if their fields are identical (and have equal types). The compiler uses strict equality initially, and then the less strict one. The less strict one is necessary only when the program contains generic functions that are invoked with generic structures as arguments.

Consider this example:
```
struct S<T> { ... }
extern S<T> f<T>();
S<bit<32>> s = f<bit<32>>();
```

Currently generic structures are eliminated in the compiler front-end by being replaced with concrete structures. In this example, `S<bit<32>>` may be replaced by a structure called `S_0`. This works as long as all uses of `S<bit<32>>` in the program are replaced with `S_0`. However, the result type of `f<bit<32>>` will always be computed by type inference to be `S<bit<32>>`, and never `S_0`. The more relaxed typing rules will allow the assignment between values with type `S_0` and `S<bit<32>>`.

Fixes #2635 
Fixes #2636


mbudiu-vmw:
The PR also contains a commit that tightens the checks for unused declarations to handle type variables and parameters; as a result there are quite a few more warnings produced.

mbudiu-vmw:
S_0 y;
S<bit<32>> x = y;

ChrisDodd (commented):
This seems like a lot of changes to pass around a single flag.  Perhaps we could make the flag part of the TypeMap object and make TypeMap::equivalent a member function rather than a static function?  Then too, the getType/getTypeType calls could be folded into equivalent, making the code that uses it cleaner.

mbudiu-vmw:
This is a good idea. However, the flag changes during compilation, so it will have to be a mutable field of the typeMap.

mbudiu-vmw:
@ChrisDodd I have applied the change you have suggested. Could you please review again?

jnfoster:
@mbudiu-vmw can you rebase?

mbudiu-vmw:
@jnfoster this should be ready for review. There are lots of changed files, but most of them are just reference outputs with new warnings.

mbudiu-vmw:
@rst0git I could not get anyone to review this PR, maybe you can take a look at it. Let me know if additional explanations are required. It is fixing a bug in the current compiler related to the typechecking of generic structs. The fix is not beautiful, but it seems to work.

jnfoster:
@mbudiu-vmw My apologies. I lost track of this. I can still take a look.

testdata/p4_16_samples_outputs/psa-action-selector3.p4-error
```diff
@@ -1 +1,11 @@
...
 [--Wwarn=mismatch] warning: Mismatched header/metadata struct for key elements in table tbl. Copying all match fields to metadata
+=======
+psa.p4(546): [--Wwarn=unused] warning: 'W' is unused
+extern Counter<W, S> {
+               ^
+psa.p4(575): [--Wwarn=unused] warning: 'W' is unused
+extern DirectCounter<W> {
+                     ^
+[--Wwarn=unsupported] warning: Mismatched header/metadata struct for key elements in table tbl. Copying all match fields to metadata
+>>>>>>> Changed reference outputs
```
rst0git: This looks like auto-generated diff from `git`

testdata/p4_16_samples_outputs/psa-dpdk-lpm-match-err4.p4-error
```diff
@@ -1 +1,11 @@
...
 [--Wwarn=mismatch] warning: Mismatched header/metadata struct for key elements in table tbl. Copying all match fields to metadata
+=======
+psa.p4(534): [--Wwarn=unused] warning: 'W' is unused
+extern Counter<W, S> {
+               ^
+psa.p4(563): [--Wwarn=unused] warning: 'W' is unused
+extern DirectCounter<W> {
+                     ^
+[--Wwarn=unsupported] warning: Mismatched header/metadata struct for key elements in table tbl. Copying all match fields to metadata
+>>>>>>> Changed reference outputs
```
rst0git: auto-generated diff from `git`?

testdata/p4_16_samples_outputs/psa-dpdk-table-key-consolidation-if-1.p4-error
```diff
@@ -1,2 +1,13 @@
...
+=======
+psa.p4(534): [--Wwarn=unused] warning: 'W' is unused
+extern Counter<W, S> {
+               ^
+psa.p4(563): [--Wwarn=unused] warning: 'W' is unused
+extern DirectCounter<W> {
+                     ^
+[--Wwarn=unsupported] warning: Mismatched header/metadata struct for key elements in table tbl. Copying all match fields to metadata
+[--Wwarn=unsupported] warning: Mismatched header/metadata struct for key elements in table foo. Copying all match fields to metadata
+>>>>>>> Changed reference outputs
```
rst0git: ditto

testdata/p4_16_samples_outputs/psa-dpdk-table-key-consolidation-if.p4-error
```diff
@@ -1 +1,11 @@
...
 [--Wwarn=mismatch] warning: Mismatched header/metadata struct for key elements in table tbl. Copying all match fields to metadata
+=======
+psa.p4(534): [--Wwarn=unused] warning: 'W' is unused
+extern Counter<W, S> {
+               ^
+psa.p4(563): [--Wwarn=unused] warning: 'W' is unused
+extern DirectCounter<W> {
+                     ^
+[--Wwarn=unsupported] warning: Mismatched header/metadata struct for key elements in table tbl. Copying all match fields to metadata
+>>>>>>> Changed reference outputs
```
rst0git: ditto

testdata/p4_16_samples_outputs/psa-dpdk-table-key-consolidation-mixed-keys-4.p4-error
```diff
@@ -1,6 +1,15 @@
...
             ^^^^^^
+<<<<<<< HEAD
+=======
+psa.p4(534): [--Wwarn=unused] warning: 'W' is unused
+extern Counter<W, S> {
+               ^
+psa.p4(563): [--Wwarn=unused] warning: 'W' is unused
+extern DirectCounter<W> {
+                     ^
+>>>>>>> Changed reference outputs
```
rst0git: ditto

testdata/p4_16_samples_outputs/psa-dpdk-table-key-consolidation-mixed-keys-4.p4-stderr
```diff
@@ -1,6 +1,15 @@
...
             ^^^^^^
+<<<<<<< HEAD
+=======
+psa.p4(546): [--Wwarn=unused] warning: 'W' is unused
+extern Counter<W, S> {
+               ^
+psa.p4(575): [--Wwarn=unused] warning: 'W' is unused
+extern DirectCounter<W> {
+                     ^
+>>>>>>> Changed reference outputs
```
rst0git: ditto

testdata/p4_16_samples_outputs/psa-dpdk-table-key-consolidation-mixed-keys-4.p4.bfrt.json
```diff
@@ -2,7 +2,11 @@
   "schema_version" : "1.0.0",
   "tables" : [
     {
+<<<<<<< HEAD
       "name" : "ep.egress.tbl",
+=======
+      "name" : "egress.tbl",
+>>>>>>> Changed reference outputs
```
rst0git: ditto

frontends/p4/typeChecking/typeChecker.h
```diff
@@ -85,8 +85,8 @@ class TypeInference : public Transform {
     const IR::Node* initialNode;
 
  public:
-    // If readOnly=true it will assert that it behaves like
-    // an Inspector.
+    // @param readOnly If ftrue it will assert that it behaves like
```
rst0git, line 88: *If true

rst0git (approved):
LGTM

frontends/p4/typeChecking/typeChecker.h
jnfoster: Also, I wish there was a way to enforce this invariant better...

frontends/p4/typeChecking/typeChecker.cpp
```diff
@@ -811,8 +812,10 @@ TypeInference::assignment(const IR::Node* errorPosition, const IR::Type* destTyp
             }
             if (!changes)
                 vec = si->components;
-            if (initType->is<IR::Type_UnknownStruct>() || changes)
+            if (initType->is<IR::Type_UnknownStruct>() || changes) {
                 sourceExpression = new IR::StructExpression(type, type, vec);
+                setType(sourceExpression, destType);
```
jnfoster, line 817: @mbudiu-vmw can you explain what this block is doing?

jnfoster:
LGTM, modulo a few questions...

Also, a philosophical question: is there a way to avoid squirreling away the `typeMap` in various pieces of program structure? Possibly I've misunderstood, but it seems like it would be cleaner if we could have it as a global structure during type checking rather than embedding it in terms.

jnfoster (approved):

frontends/p4/typeChecking/typeChecker.h
mbudiu-vmw: This has been here for at least 7 years.
It's not too bad, but indeed, I sometimes wished there was another way to do this.

mbudiu-vmw:
> LGTM, modulo a few questions...
> 
> Also, a philosophical question: is there a way to avoid squirreling away the `typeMap` in various pieces of program structure? Possibly I've misunderstood, but it seems like it would be cleaner if we could have it as a global structure during type checking rather than embedding it in terms.

I am not sure what you mean by "embedding into terms". The TypeInference pass receives the typeMap as an argument, and populates it with new types discovered.

frontends/p4/typeChecking/typeChecker.cpp
mbudiu-vmw: This is one of these cases where a type has been inferred for a StructExpression used in the RHS of an assignment.
A struct expression written without a type, e.g. `{a = 1}` has type `Type_UnknownStruct`. If the type is inferred it is written like `(S) {a = 1}` and the type becomes `Type_Struct`. In this case the type has been inferred from the LHS of the assignment (that's why it's `destType`).

Notice that the situation handled in this block is more complicated, since you can have something like:
`S a = (S){ a = { b = 1 }}` when `S` is a nested struct. So, even if the outer struct has the correct type `S`, the inner one does still have a `Type_UnknownStruct`, so the code prior to this has to recurse over the struct expression to make sure that all its fields are correct as well. Their types are inferred from the fields of `S`.

mbudiu-vmw: A similar inference pattern is used when structs are compared.

mbudiu-vmw: This code also handles mixed assignments that use list expressions, casting them to StructExpressions: e.g.:
`S s = { a = { 1 } }`

frontends/p4/typeChecking/typeChecker.h
mbudiu-vmw: I will fix this typo in a separate PR, so we don't have to wait for another build before merging.

Issue #2288, opened 2020-04-06
SideEffectOrdering: Regression?


fruffy:
Hello.

I have the following program, which I reduced to look just like the example in the spec:
```p4
bit<8> f(inout bit<8> x, in bit<8> z) {
    return 8w4;
}
bit<8> g(inout bit<8> z) {
    z = 8w3;
    return 8w1;
}
control ingress(inout Headers h, inout Meta m, inout standard_metadata_t sm) {
    apply {
        f(h.h.a, g(h.h.a));
    }
}
```

After `SideEffectOrdering`, the program looks like this
```p4
control ingress(inout Headers h, inout Meta m, inout standard_metadata_t sm) {
    bit<8> tmp;
    bit<8> tmp_0;
    apply {
        {
            tmp = g(h.h.a);
            tmp_0 = tmp;
            f(h.h.a, tmp_0);
        }
    }
}

```
which implies that `h.h.a` is `3` instead of being unchanged.

According to section `6.7`, I would expect the transformation to look something like
```p4
control ingress(inout Headers h, inout Meta m, inout standard_metadata_t sm) {
    bit<8> tmp;
    bit<8> tmp_0;
    apply {
        tmp = h.h.a;     // evaluate h.h.a; save result
        tmp_0 = g(h.h.a);  // evaluate g(h.h.a); save result; modifies h.h.a
        f(tmp, tmp_0);    // evaluate f; modifies tmp
        h.h.a = tmp;         // copy inout result back into h.h.a
    }
}
```
so the value of `h.h.a` does not change since f does not modify the input.

[2288.p4.txt](https://github.com/p4lang/p4c/files/5465546/2288.p4.txt)
[2288.stf.txt](https://github.com/p4lang/p4c/files/5465547/2288.stf.txt)

fruffy:
The sample program is missing the `pkt.extract(hdr.h);` call, for some reason I cannot update the original file.

fruffy:
Just adding another instance of this issue with side-effects in table calls. 

The program
```p4
control ingress(inout Headers h, inout Meta m, inout standard_metadata_t sm) {
    action simple_action_0() {
        h.eth_hdr.eth_type = 16w3;
    }
    action dummy_action_0(inout bit<16> val1, in bit<16> val2) {
    }
    table simple_table_0 {
        key = {
            h.eth_hdr.src_addr: exact @name("key1") ;
        }
        actions = {
            simple_action_0();
            @defaultonly NoAction();
        }
[...]

PR #2638, opened 2020-12-18
SideEffectOrdering #2288

milicaperic:
Change made in sideEffects.cpp and sideEffects.h

There are side efects when two arguments are aliases and one of them is in and other is out or inout.
In addition to a simple example such as this f(val, val) (T f(in T x, inout T y) )
the following cases have been treated
- f(g(val), s[val].x)
- f(s[val].x, g(val))
- f(val, g(val))
- f(g(val), val)
- and other similar cases with multiple arguments.


frontends/p4/sideEffects.cpp
```diff
@@ -323,6 +323,59 @@ bool DoSimplifyExpressions::containsHeaderType(const IR::Type* type) {
...
+        }
+        if (param1->type == param2->type && mayAlias(expr1, expr2)) {
+            returnParam = param2;
+            break;
+        }
+    }
+    return returnParam;
+}
+
+cstring DoSimplifyExpressions::declareTemporary(const IR::Parameter* p,
```
mbudiu-vmw, line 351: The name of this function seems misleading, since it does not just declare the temporary, but it also initializes it.

frontends/p4/sideEffects.cpp
```diff
@@ -391,39 +432,282 @@ const IR::Node* DoSimplifyExpressions::preorder(IR::MethodCallExpression* mce) {
...
-                if (!p1->hasOut() && !p2->hasOut())
+
+            LOG3("p1=" << dbp(p1) << " p2=" << dbp(p2));
+            if (p2 == p1)
+                break;
+
+            auto arg2 = mi->substitution.lookup(p2);
+
+            if (auto m = arg1->expression->to<IR::Member>()) {
+                auto u = arg1->expression->to<IR::Operation_Unary>();
```
mbudiu-vmw, line 453: Why do you need this? Can't you just use m instead of u everywhere?

frontends/p4/sideEffects.cpp
```diff
@@ -391,39 +432,282 @@ const IR::Node* DoSimplifyExpressions::preorder(IR::MethodCallExpression* mce) {
...
+                if  (u->expr->is<IR::ArrayIndex>()) {
+                    auto b = u->expr->to<IR::Operation_Binary>();
```
mbudiu-vmw, line 455: here too you can write if (auto b = u->expr->to<IR::ArrayIndex>())

frontends/p4/sideEffects.cpp
```diff
@@ -391,39 +432,282 @@ const IR::Node* DoSimplifyExpressions::preorder(IR::MethodCallExpression* mce) {
...
+                    // we handle this case T f(in T x, inout T y){...}; T g(inout T z){...}
+                    // f(g(val), s[val].x)
```
mbudiu-vmw, line 457: this comment does not seem to match the code. The code checks if arg1 is that shape a[].m

frontends/p4/sideEffects.cpp
```diff
@@ -391,39 +432,282 @@ const IR::Node* DoSimplifyExpressions::preorder(IR::MethodCallExpression* mce) {
...
+                    // tmp_2 = g(val);
+                    // f(tmp_2, tmp_1);
+                    // s[tmp_0].x = tmp_1;
+                    if (!b->right->is<IR::MethodCallExpression>() &&
+                        arg2->expression->is<IR::MethodCallExpression>()) {
+                        auto mce2 = arg2->expression->to<IR::MethodCallExpression>();
+                        auto p3 = findAlias(p1, b->right, mce2);
+                        if  (p3) {
+                            auto mi2 = MethodInstance::resolve(mce2, refMap, typeMap);
+                            auto arg3 = mi2->substitution.lookup(p3);
```
mbudiu-vmw, line 469: I don't understand why arg3 is not null. You seem to assume that p3 is the same as p2 here. What if the call you are analyzing has more than 2 arguments and the aliasing one is the third one?

frontends/p4/sideEffects.h
```diff
@@ -128,6 +128,14 @@ class DoSimplifyExpressions : public Transform, P4WriteContext {
     const IR::Expression* addAssignment(Util::SourceInfo srcInfo, cstring varName,
                                         const IR::Expression* expression);
     bool mayAlias(const IR::Expression* left, const IR::Expression* right) const;
+    const IR::Parameter* findAlias(const IR::Parameter* p,
```
mbudiu-vmw, line 131: please document this function.  In general expr1 seeems to be a subexpression of an argument assigned to p.

testdata/p4_16_samples_outputs/issue2288-first (copy).p4
```diff
@@ -0,0 +1,68 @@
+#include <core.p4>
```
mbudiu-vmw, line 1: I suspect this file should not exist.

frontends/p4/sideEffects.cpp
```diff
@@ -391,39 +432,282 @@ const IR::Node* DoSimplifyExpressions::preorder(IR::MethodCallExpression* mce) {
...
-                    break;
-                if (!p1->hasOut() && !p2->hasOut())
+
+            LOG3("p1=" << dbp(p1) << " p2=" << dbp(p2));
+            if (p2 == p1)
+                break;
+
+            auto arg2 = mi->substitution.lookup(p2);
+
+            if (auto m = arg1->expression->to<IR::Member>()) {
```
mbudiu-vmw, line 452: This expression seems overly specific. What if you have `val + a[val].index`?

mbudiu-vmw (changes requested):
It looks like I started reviewing the commit in your own repository instead of this PR.
Please address the comments I made so far - using fewer variables and having the comments match the code will make the review easier. I will re-review once you submit a corrected PR.
Also, perhaps you can make an argument in a comment why you are covering all possible cases.

fruffy:
Does this pull request also cover https://github.com/p4lang/p4c/issues/2288#issuecomment-719430077? It looks like it leads to a compiler bug.

milicaperic:
> Does this pull request also cover [#2288 (comment)](https://github.com/p4lang/p4c/issues/2288#issuecomment-719430077)? It looks like it leads to a compiler bug.

You are right, there is a bug when I tasting your example 2288.p4.
I can fix it easily with a few line

if(arg1->expression->is<IR::Operation_Ternary>() || arg2->expression->is<IR::Operation_Ternary>()) {
                continue;
  }

but that lines just ignore ternary operator don't analyze and remove problem of side effect.
I have problem with fixing ternary operations at all.
If we have function like
func(f(val)?g(val):h(val), val) or func(f(val)?g(val):h(val), s[val].x)
we have to find aliases between 
-all arguments in f and val
-all arguments in g and val
-all arguments in h and val
-and all arguments in f and g
-and all arguments in f and h
-and all arguments in g and h
And what if we have function like this
func(f1(val)?g1(val):h1(val), f2(val)?g2(val):h2(val))  ?

And if we resolve all these cases, still exists side effect in 
assignment statement and binary operators like:
- s[a].z = g(a)
- s[g(a)].z = a
- f(val + g(val))
- f(a + g(val), val)
- f(val + g(a), val).

Are all these cases supported? @mbudiu-vmw 
Should they be resolved, ignored or a warning written?

frontends/p4/sideEffects.cpp
milicaperic: It may not be so obvious (it's not intuitive) but nested for-loop visit arguments in that order.
Each argument is processed with all previous arguments.

For example:
Let f(a, b, c, d, e, f)
arg1: b; arg2: a;
arg1: c; arg2: a;
arg1: c; arg2: b;
arg1: d; arg2: a;
arg1: d; arg2: b;
arg1: d; arg2: c;
arg1: e; arg2: a;
arg1: e; arg2: b;
arg1: e; arg2: c;
arg1: e; arg2: d;

That is why my comment matched the code.
(for f(g(val), s[val].x)  
arg1: s[val].x; arg2: g(val); )

frontends/p4/sideEffects.cpp
milicaperic: I didn't understand the question.
Maybe the name (arg3) is wrong.
arg3 is an argument of g what an alias with val. 
It is not third argument of f.
Maybe you didn't mean that.

here are 2 examples:

extern bit f(in bit x, inout bit y);
extern bit g(inout bit x, inout bit y);

f(g(b, a), s[a].z);

after SideEffectOrdering:
tmp = a_0;
tmp_0 = s_0[tmp].z;
tmp_1 = g(b_0, a_0);
f(tmp_1, tmp_0);
s_0[tmp].z = tmp_0;

extern bit f(inout bit x, in bit y, in bit z);
extern bit g(inout bit x);

f(s[a].z, b, g(a));
after SideEffectOrdering:
tmp = a_0;
tmp_0 = s_0[tmp].z;
tmp_1 = g(a_0);
f(tmp_0, b_0, tmp_1);
s_0[tmp].z = tmp_0;

mbudiu-vmw:
@milicaperic based on your proposal I went and improved the sideEffects.cpp code. I noticed that the work you do in findAlias should actually be done by the existing mayAlias function, so I tweaked that one. It had at least one bug in the handling of array indexes too.

milicaperic:
What should be my further steps regarding this PR since you opened a [https://github.com/p4lang/p4c/pull/2640]?

mbudiu-vmw:
First, I very much appreciate the review on my failed attempts to far to fix the issue.
I am looking for a solution that does not need to enumerate too many special cases (I hope it's possible).
For now leave this open and in the worst case we will close it once we solve the problem.
If you are willing to rework it and find a simple-looking solution I will review it again.

fruffy:
I think this issue has been fixed with #2640. Should we close this request?

PR #2639, opened 2020-12-19
minor fix: check for nullptr returned from ProgramPoint.last()

hanw:


hanw:
Bug found when running with larger customer programs.

mbudiu-vmw (approved):

PR #2643, opened 2021-01-06
Update the testcase to use xor for overlapping write

lizhengxing:
Overlapping set will be supported. Update the test case to use xor for overlapping write


mbudiu-vmw:
In general we don't update tests, we write new ones.
Also, if you change a test you have to submit as part of the PR the changes in the reference outputs produced by the compiler.

lizhengxing:
> In general we don't update tests, we write new ones.
> Also, if you change a test you have to submit as part of the PR the changes in the reference outputs produced by the compiler.

Got it, Let me withdraw this PR and try again by your suggestion, thanks!

PR #2644, opened 2021-01-06
Make declarations global in parser before converting ifs to states

### LINKED ISSUE #2409: Implementation of 'if' statements in parsers is incorrect
The current implementation should first make all state-local variables global in the parser.

mbudiu-vmw:
Fixes #2409


jnfoster (approved):
Looks good. Thanks! 

One issue I didn't see a test for: I assume the state name generation works when there are multiple conditionals?

PR #2652, opened 2021-01-15
Remove non local dead code post LocalCopyPropagation

boblytton:
Non local dead code was not being removed by LocalCopyPropagation.
See gtest LocalCopyPropTest.DeadCodeRemoval in test\gtest\local_copyprop.cpp for an example.

### DIFF
--- a/midend/local_copyprop.h
+++ b/midend/local_copyprop.h
@@ -18,6 +18,7 @@ limitations under the License.
 
 #include "ir/ir.h"
 #include "frontends/p4/typeChecking/typeChecker.h"
+#include "frontends/p4/simplifyDefUse.h"
 #include "frontends/common/resolveReferences/referenceMap.h"
 #include "has_side_effects.h"
 
@@ -135,6 +136,7 @@ class LocalCopyPropagation : public PassManager {
             typeChecking = new TypeChecking(refMap, typeMap, true);
         passes.push_back(typeChecking);
         passes.push_back(new DoLocalCopyPropagation(refMap, typeMap, policy, elimUnusedTables));
+        passes.push_back(new SimplifyDefUse(refMap, typeMap, typeChecking));
     }
 };
 
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -49,6 +49,7 @@ set (GTEST_UNITTEST_SOURCES
   gtest/format_test.cpp
   gtest/helpers.cpp
   gtest/json_test.cpp
+  gtest/local_copyprop.cpp
   gtest/midend_test.cpp
   gtest/opeq_test.cpp
   gtest/ordered_map.cpp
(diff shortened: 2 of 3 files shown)


mbudiu-vmw:
In general we do not expect each optimization pass to produce an "optimal" program, but only to perform one particular optimization. We rely on the ordering of passes in frontend.cpp and midend.cpp to apply a reasonable sequence of optimizations. 
Your solution could be acceptable if every time you run LocalCopyPropagation you do expect to need SimplifyDefUse - but that pass is relatively expensive, so I don't know if we need to run it that often.

Also, if your change will affect any programs, then your PR should contain the reference outputs from testdata/*_outputs that are being affected.

boblytton:
Hi @mbudiu-vmw

>We rely on the ordering of passes in frontend.cpp and midend.cpp to apply a reasonable sequence of optimizations.

I had previously made the change in our local midend pass manager, but was in two minds as to if it was better coupled tightly to LocalCopyPropagation.

> Your solution could be acceptable if every time you run LocalCopyPropagation you do expect to need SimplifyDefUse.

Good question, in the midend I would expect LocalCopyPropagation to be run once and SimplifyDefUse to be run once after it.
The use-case (test\gtest\local_copyprop.cpp) suggests this is a wise thing to do...

I am happy to drop this PR and would be pleased to be directed.

boblytton:
How about a boolean?
```
LocalCopyPropagation(ReferenceMap* refMap, TypeMap* typeMap,
        TypeChecking* typeChecking = nullptr,
        std::function<bool(const Context *, const IR::Expression *)> policy =
            [](const Context *, const IR::Expression *) -> bool { return true; },
        bool elimUnusedTables = false,
        bool reRunSimplifyDefUse= false);

```
(I am unclear as to whether others should be running SimplifyDefUse or not)

mbudiu-vmw:
There are three levels of organization of passes:
- individual Inspec/Transform/Modify, which perform some indivisible traversal
- actual compiler passes, that do one atomic transformation. Often these contain several Inspects followed by a Transform. In general, the inspectors gather information and the Transform applies changes based on this information.
- finally, the compiler stages, like front-end or mid-end. These are sequence of passes organized in some order where earlier passes usually establish invariants necessary for the latter ones, but where the ordering is not necessarily total (i.e., some passes could be reordered without affecting correctness).

I personally vote against customizing LocalCopyPropagation; think of it as a brick that you can use to erect your building. The building is the frontend or the midend, but I would not customize this brick in such a way.

There are no hard and fast rules, though, so your proposal could be considered reasonable if LocalCopyPropagation always leaves lots of reside. We run it several times.

Let's see if @ChrisDodd can weigh on this, he wrote the LocalCopyPropagation pass.

boblytton:
> I personally vote against customizing LocalCopyPropagation; think of it as a brick that you can use to erect your building. The building is the frontend or the midend, but I would not customize this brick in such a way.

This seems very reasonable and correct.
Thank you.

PR #2656, opened 2021-01-22
Improve Bazel CI build times through smarter caching

smolkaj:
GitHub's caching mechanism never overwrites cache entries for existing keys. Since we were using a static key, this caused our cache to never be refreshed. 

With this PR, a new cache entry is generated at the end of every CI run; at the beginning of every CI run, the freshest cache entry is restored.

### DIFF
--- a/.github/workflows/ci-bazel.yml
+++ b/.github/workflows/ci-bazel.yml
@@ -21,7 +21,11 @@ jobs:
       with:
         # See https://docs.bazel.build/versions/master/output_directories.html
         path: "~/.cache/bazel"
-        key: ${{ runner.os }}-build-direct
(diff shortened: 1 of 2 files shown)


mbudiu-vmw:
Does this address the recent build failures?

smolkaj:
No, but let me take care of this as well while I'm at it.

mbudiu-vmw (approved):

smolkaj:
Just to confirm, the latest Bazel CI run took only 2 minutes. It won't be quite as fast when there are actual changes to the C++ code, but it should be a lot faster than before (~ 20 minutes).

smolkaj:
Ready to be merged

PR #2658, opened 2021-01-26
Allow casts between int and bool

### LINKED ISSUE #2444: Casting integers to bool
The following code is accepted.
```
const int a =  1;
const bool b = (bool)a;
```

If I change the 1 to a 2w1, then the program is rejected with a message that "only bit<1> values can be cast to bool":
```
const int a =  2w1;
const bool b = (bool)a;
```

I would expect the first program to fail since it's casting an int and not a bit<1> to bool, and I'd think the second program would result in an error on the first line because it's assigning a bit<2> to an int variable.

mbudiu-vmw:
Fixes #2444

### DIFF
--- a/frontends/p4/typeChecking/typeChecker.cpp
+++ b/frontends/p4/typeChecking/typeChecker.cpp
@@ -705,6 +705,8 @@ bool TypeInference::canCastBetween(const IR::Type* dest, const IR::Type* src) co
             return f->size == 1 && !f->isSigned;
         } else if (auto de = dest->to<IR::Type_SerEnum>()) {
             return TypeMap::equivalent(src, getTypeType(de->type));
+        } else if (auto ii = dest->to<IR::Type_InfInt>()) {
+            return true;
         }
     } else if (src->is<IR::Type_Boolean>()) {
         if (dest->is<IR::Type_Bits>()) {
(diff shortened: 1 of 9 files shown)


jnfoster (approved):
Looks good.

frontends/p4/typeChecking/typeChecker.cpp
```diff
@@ -705,14 +705,16 @@ bool TypeInference::canCastBetween(const IR::Type* dest, const IR::Type* src) co
...
+            return true;
         }
     } else if (src->is<IR::Type_Boolean>()) {
         if (dest->is<IR::Type_Bits>()) {
             auto b = dest->to<IR::Type_Bits>();
             return b->size == 1 && !b->isSigned;
         }
     } else if (src->is<IR::Type_InfInt>()) {
-        return dest->is<IR::Type_Bits>();
+        return dest->is<IR::Type_Bits>() || dest->is<IR::Type_Boolean>();
```
jnfoster, line 717: Where do you check that the value of `src` is `0` or `1` as mentioned in the P4 Spec PR?

jnfoster (changes requested):

frontends/p4/typeChecking/typeChecker.cpp
mbudiu-vmw: The check is in the constant folding code.

jnfoster: Sorry, I'm dense and not seeing it. Can you point me to a line number?

jnfoster: Nevermind. I found it. 

https://github.com/p4lang/p4c/blob/40c458898de221c9902e1febf85ac5d438f57c47/frontends/common/constantFolding.cpp#L804

jnfoster (approved):

PR #2661, opened 2021-01-29
Add the Gauntlet tests to p4c

fruffy:
This pull request contributes the roughly 100 tests we have auto-generated while developing Gauntlet. The files are distinct from the existing P4 tests in that they caused issues in our interpreter, whereas p4c compiled them correctly. They might be useful for other developers writing tools for the P4 language. I can also contribute the .stf tests for most of these, if requested. 

Because of the sheer volume reviewing may be an issue. However, I believe these files are compiled correctly, otherwise I would have filed an issue at an earlier point.


mbudiu-vmw:
I think that STF files would be very valuable. If it is not too much manual work to add them, it would be great.

jnfoster:
+1 Totally agree.

fruffy:
Converting the tests to v1model programs and generating STF tests unearthed some issues. They probably have to be looked at on a case-by-case basis. Should I mark them XFAIL in the meantime?

mbudiu-vmw:
Since there is a small number that fail the tests, I think that removing them altogether as tests is probably a better idea than making them xfail. xfail tests have a tendency to never get revisited.

fruffy:
I removed the offending tests. I will create a pull request to add them back once https://github.com/p4lang/behavioral-model/issues/976 is resolved.

mbudiu-vmw (approved):
As you can imagine, I haven't actually read all these test files. I rely on the fact that they have been formally validated, and that the tests all pass.

mbudiu-vmw:
Looks like I broke the build when merging this. I will try to fix it.
Some reference outputs are different.

Issue #2628, opened 2020-12-08
SimplifyDefUse doesn't remove redundant method calls


dbrajovic-syrmia:
Unassigned (duplicate) method calls are not removed during this pass. For example, 
```
        header hdr {
             bit<32> a;
        }

        hdr h;
        // ...
        h.isValid();
        h.isValid();
        h.isValid();
```
All three `isValid()` statements are preserved after the pass. Assigning them to a variable only strips them of the lhs, turning the assignment into a method call statement.
[simplify_method_calls.txt](https://github.com/p4lang/p4c/files/5658058/simplify_method_calls.txt)

jafingerhut:
I verified that the latest p4c as of 2022-Jan-04 has output from the SimplifyDefUse pass that has only one setInvalid call, not a sequence of several identical ones.  Closing this issue.

PR #2662, opened 2021-01-29
SimplifyDefUse removing redundant method calls now, BugFix #2628:

milicaperic:
MCE has not produced side effects in two cases:
* isValid()
* function with only in parameters

### DIFF
--- a/frontends/p4/sideEffects.h
+++ b/frontends/p4/sideEffects.h
@@ -90,9 +90,11 @@ class SideEffects : public Inspector {
         auto mce = exp->to<IR::MethodCallExpression>();
         if (mce == nullptr)
             return false;
-        // mce does not produce a side effect in the two cases:
+        // mce does not produce a side effect in few cases:
         //  * isValid()
         //  * function with all in parameters
+        //  * extern function with noSideEffectsAnnotation
+        //  * extern method with noSideEffectsAnnotation
         auto mi = MethodInstance::resolve(mce, refMap, typeMap);
         if (mi->is<FunctionCall>()) {
             for (auto p : *mi->substitution.getParametersInArgumentOrder()) {
@@ -101,6 +103,18 @@ class SideEffects : public Inspector {
                 }
             }
             return false;
+        }
+        if (auto em = mi->to<P4::ExternMethod>()) {
+            if (em->method->getAnnotation(IR::Annotation::noSideEffectsAnnotation)) {
+                return false;
+            }
+            return true;
+        }
+        if (auto ef = mi->to<P4::ExternFunction>()) {
+            if (ef->method->getAnnotation(IR::Annotation::noSideEffectsAnnotation)) {
+                return false;
+            }
+            return true;
         }
         if (auto bim = mi->to<BuiltInMethod>()) {
             if (bim->name.name == IR::Type_Header::isValid) {
--- a/frontends/p4/sideEffects.h
+++ b/frontends/p4/sideEffects.h
@@ -83,6 +83,32 @@ class SideEffects : public Inspector {
         expression->apply(se);
         return se.nodeWithSideEffect != nullptr;
     }
+    /// @return true if the expression may have side-effects.
+    static bool hasSideEffect(const IR::Expression* exp,
+                      ReferenceMap* refMap,
+                      TypeMap* typeMap) {
+        auto mce = exp->to<IR::MethodCallExpression>();
+        if (mce == nullptr)
+            return false;
+        // mce does not produce a side effect in the two cases:
+        //  * isValid()
+        //  * function with all in parameters
+        auto mi = MethodInstance::resolve(mce, refMap, typeMap);
+        if (mi->is<FunctionCall>()) {
+            for (auto p : *mi->substitution.getParametersInArgumentOrder()) {
+                if (p->hasOut()) {
+                    return true;
+                }
+            }
+            return false;
+        }
+        if (auto bim = mi->to<BuiltInMethod>()) {
+            if (bim->name.name == IR::Type_Header::isValid) {
+                return false;
+            }
+        }
+        return true;
+    }
 };
 
 /** @brief Convert expressions so that each expression contains at most one
--- a/frontends/p4/simplifyDefUse.cpp
+++ b/frontends/p4/simplifyDefUse.cpp
@@ -706,14 +706,17 @@ class FindUninitialized : public Inspector {
 class RemoveUnused : public Transform {
     // TODO: remove transitively unused
     const HasUses* hasUses;
+    ReferenceMap*   refMap;
+    TypeMap*        typeMap;
 
  public:
-    explicit RemoveUnused(const HasUses* hasUses) : hasUses(hasUses)
-    { CHECK_NULL(hasUses); setName("RemoveUnused"); }
+    explicit RemoveUnused(const HasUses* hasUses, ReferenceMap* refMap, TypeMap* typeMap)
+                        : hasUses(hasUses), refMap(refMap), typeMap(typeMap)
+    { CHECK_NULL(hasUses);  CHECK_NULL(refMap);  CHECK_NULL(typeMap); setName("RemoveUnused"); }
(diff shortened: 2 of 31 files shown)


frontends/p4/sideEffects.h
```diff
@@ -83,6 +83,46 @@ class SideEffects : public Inspector {
...
+                    return false;
+            } else if (basetype->is<IR::Type_Header>()) {
+                if (mem->member == IR::Type_Header::isValid)
+                    return false;
+            }
+        } else if (mce->method->is<IR::PathExpression>()) {
+            auto pe = mce->method->to<IR::PathExpression>();
+            auto decl = refMap->getDeclaration(pe->path, true);
+            if (decl->is<IR::Function>()) {
+                auto mi = MethodInstance::resolve(mce, refMap, typeMap);
```
mbudiu-vmw, line 115: This is not precise enough, you are missing several cases:
- If you invoke an action or a table.apply() it may have side-effects.
- some external functions may have side-effects that are not visible to the data-plane. For these you may want to check annotations too.
- evaluating the function's arguments may have side-effects

frontends/p4/simplifyDefUse.cpp
```diff
@@ -726,10 +729,21 @@ class RemoveUnused : public Transform {
...
                 return new IR::MethodCallStatement(statement->srcInfo, mce);
             }
+            // removing
             return new IR::EmptyStatement();
         }
         return statement;
     }
+    const IR::Node* postorder(IR::MethodCallStatement* mcs) override {
+        if (!hasUses->hasUses(getOriginal())) {
+            if (SideEffects::hasSideEffect(mcs->methodCall, refMap, typeMap)) {
```
mbudiu-vmw, line 739: Why can't you just reuse the existing check method?

mbudiu-vmw (changes requested):

frontends/p4/sideEffects.h
milicaperic: Than you for quick review. You are right that actions and table.apply() may produce side-effects, but hasSideEffect function returns true in all cases except  isValid() case and case with function for all in arguments.
There is a test simplify_method_calls.p4 that I have added, there is some simple action that produce side effect.

frontends/p4/simplifyDefUse.cpp
milicaperic: I have tried with check method, but `T f( in T a)` function didn't removed.

milicaperic: `bit<32> f(out bit<32> a) {
    a = 32w0;
    return a;
}
bit<32> g(in bit<32> a) {
    return a;
}
control c(out bit<32> x, out bit<32> y) {
    action simple_action() {
        y = g(x);
        f(x);
        g(x);
    }
    apply {
        hdr h = { 0 };
        bool b = h.isValid();
        b = h.isValid();
        b = h.isValid();
        b = h.isValid();
        h.isValid();
        h.isValid();
        h.isValid();
        h.setValid();
        if (b) {
            x = h.a;
            x = h.a;
            f(h.a);
            f(h.a);
            g(h.a);
            g(h.a);
        } else {
            x = f(h.a);
            x = g(h.a);
            f(h.a);
        }
        simple_action();
    }
}` after SimplifyDefUse pass 
`bit<32> f(out bit<32> a) {
    a = 32w0;
    return a;
}
bit<32> g(in bit<32> a) {
    return a;
}
control c(out bit<32> x, out bit<32> y) {
    @name("h") hdr h_0;
    @name("b") bool b_0;
    @name("simple_action") action simple_action_0() {
        y = g(x);
        f(x);
        ;
    }
    apply {
        h_0 = (hdr){a = 32w0};
        ;
        ;
        ;
        b_0 = h_0.isValid();
        ;
        ;
        ;
        h_0.setValid();
        if (b_0) {
            ;
            x = h_0.a;
            f(h_0.a);
            f(h_0.a);
            ;
            ;
        } else {
            x = f(h_0.a);
            x = g(h_0.a);
            f(h_0.a);
        }
        simple_action_0();
    }
}
`

mbudiu-vmw: Not all functions with only 'in' arguments will have no side-effects.

milicaperic: Can you give me an example?

mbudiu-vmw: log_msg from v1model

milicaperic: You are right! 
Are we only sure for isValid() that it doesn't  produce side effects?

mbudiu-vmw: There are "standard" methods, like isValid, for which the compiler can make assumptions.
Anything that is "extern" we should play it safe. 
Some annotations were added to the spec for this purpose; they are used in the compiler in some places, but not consistently everywhere (and certainly not here). See this section: https://p4.org/p4-spec/docs/P4-16-v1.2.1.html#sec-extern-annotations. So if you combine your test for 'in' arguments with this annotation the result should be safe.

milicaperic: What do you suggest to change my code in that way, or use existing method check?
I will definitely take a look at that section.

mbudiu-vmw: The SideEffects visitor does a deep traversal of the expression. That is necessary if you use it before the SideEffects pass. After SideEffects an expression has at most one operation with side-effects.
We should document when the visitor is safe to use. 
Ideally we should build some set of functions that can be reused for checking for side-effects, today many passes reimplement their own. 
Whatever you build, it must be "safe" for the intended use cases.

mbudiu-vmw:
This bazel build failed because of an error reported by the caching service.

frontends/p4/simplifyDefUse.cpp
milicaperic: Thank you for the explanation. I'm going to make some changes now.

frontends/p4/sideEffects.h
milicaperic: `if( decl->is<IR::Function>())` this condition will not be met in log_msg (and other extern methods or functions).
log_msg (and other extern methods or functions) will be met in following condition
`if(decl->is<IR::Method>())`
(but I didn't include that condition because all conditions except the first one return true).
You can look at the tests with log_msg (logging.p4 and issue2201-bmv2.p4).
Method wasn't removed from program and test wasn't changed.

milicaperic: since this case is actually covered, is there anything else I should change? @mbudiu-vmw

mbudiu-vmw, line 115: Why not use Resolve first and the look at the result to categorize the function? 
It does most of the work for you. It even has a case for built-in methods.

frontends/p4/sideEffects.h
```diff
@@ -83,6 +83,32 @@ class SideEffects : public Inspector {
...
+                      ReferenceMap* refMap,
+                      TypeMap* typeMap) {
+        auto mce = exp->to<IR::MethodCallExpression>();
+        if (mce == nullptr)
+            return false;
+        // mce does not produce a side effect in the two cases:
+        //  * isValid()
+        //  * function with all in parameters
+        auto mi = MethodInstance::resolve(mce, refMap, typeMap);
+        if (mi->is<FunctionCall>()) {
```
mbudiu-vmw, line 97: This is fine, but you can also add to this list extern methods and extern functions that have the right annotations. We can merge as it is, or we can do this in a separate pull request.

mbudiu-vmw (approved):

mbudiu-vmw:
I don't understand the two failures in this build. The error report is something like this:
```
Error response from daemon: unauthorized: authentication required
Unable to find image 'webhare/ccache-memcached-server:latest' locally
docker: Error response from daemon: unauthorized: authentication required.
See 'docker run --help'.
```

I hope this is just some transient network failure. In that case the PR is ready for review.

mbudiu-vmw (approved):

PR #2663, opened 2021-01-29
Add more flexibility to compiler configuration.

liujed-intel:
This turns the `P4CConfiguration` into an abstract class to allow overriding of compiler configuration parameters. The active configuration can be obtained from `P4CContext`.


mbudiu-vmw:
This looks good, but it's still a draft, so I won't review it now.

mbudiu-vmw (approved):

PR #2667, opened 2021-02-03
Do not crash when handling unknown externs

### LINKED ISSUE #2664: BMV2 Back end Compiler Bug:  could not convert to Json
Sourced from #2661.
Compiling the following program with `p4c-bm2-ss` causes a compiler bug in the bmv2 back end of p4c:

```bash
fruffy@fruffy /s/P/p4_tv (master) [1]> p4c/build/p4c-bm2-ss p4c/testdata/p4_16_samples/gauntlet_option
al-bmv2.p4                                                                                                            
In file: modules/p4c/backends/bmv2/common/expression.cpp:97                                   
Compiler Bug: p4c/testdata/p4_16_samples/gauntlet_optional-bmv2.p4(443): ipv4_checksum_0/ipv4_checksum;: could not convert to Json                                                                                                  
        h.h.result = ipv4_checksum.update({ h.eth_hdr.dst_addr, h.eth_hdr.src_addr, h.eth_hdr.eth_type});  
```

[gauntlet_optional-bmv2.p4.txt](https://github.com/p4lang/p4c/files/5904654/gauntlet_optional-bmv2.p4.txt)

mbudiu-vmw:
Fixes #2664

### DIFF
--- a/backends/bmv2/common/expression.cpp
+++ b/backends/bmv2/common/expression.cpp
@@ -479,7 +479,7 @@ void ExpressionConverter::postorder(const IR::Member* expression)  {
 
 Util::IJson* ExpressionConverter::fixLocal(Util::IJson* json) {
     if (!json)
-        return new Util::JsonValue(); // null
+        return new Util::JsonValue();  // null
     if (auto jo = json->to<Util::JsonObject>()) {
         auto to = jo->get("type");
         if (to != nullptr && to->to<Util::JsonValue>() != nullptr &&
--- a/backends/bmv2/common/expression.cpp
+++ b/backends/bmv2/common/expression.cpp
@@ -94,7 +94,9 @@ Util::IJson* ExpressionConverter::get(const IR::Expression* expression) const {
             LOG3(" " << it.first << " " << it.second);
         }
     }
-    BUG_CHECK(result, "%1%: could not convert to Json", expression);
+    if (result == nullptr)
+        ::error(ErrorType::ERR_UNSUPPORTED,
+                "%1%: could not convert expression to Json", expression);
     return result;
 }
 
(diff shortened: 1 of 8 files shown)


jafingerhut:
The error messages look reasonable to me when an unsupported extern.

Looks like a non-empty -stderr file needs to be checked in for the tests to pass.

mbudiu-vmw:
Actually I believe that no -stderr file needs to be present: we only check the -stderr file generated by p4test, and this new test passes with p4test, but fails with p4c-bm2-ss. We do not check the -stderr files for the latter compiler.

backends/bmv2/common/expression.cpp
```diff
@@ -298,6 +304,8 @@ void ExpressionConverter::postorder(const IR::Member* expression)  {
             result->emplace("value", fieldName);
         } else if (parentType->is<IR::Type_HeaderUnion>()) {
             auto l = get(expression->expr);
+            if (!l)
```
fruffy, line 307: Would it make sense to convert this into a macro? So that this check does not have to be implemented manually for all these calls.

mbudiu-vmw: Well, you can create a macro that also declares the variable, like this:
`#define GET2(var, exp) auto var = get(exp); if (!var) return;`
But this does not make it obvious that GET2 declares a new identifier when you invoke it:
`GET2(l, expression);`. It will also generate 3 statements, not just one. You cannot use the do-while trick in macros to generate a single statement, since you need `var` to be declared in the same scope.

The alternative solution is to use a GCC extension, which uses statement expressions:
```
#define GET1(expression) ({ \
    auto r = get(expression); \
    if (!r) return; \
    r; \
})
```
However, this may give wrong results if the `r` variable is shadowed.
So I don't like any of these solutions using a macro. Do you have an alternative proposal?

fruffy: No unfortunately not. There does not seem a nice way to implement this. But since the changes are localized to a single file the impact seems low.

fruffy (approved):

PR #2669, opened 2021-02-03
DPDK: add support for h.setValid and h.setInvalid

hanw:
pretty straightforward translation.
`h.setValid() -> validate h`
`h.setInvalid() -> invalidate h`


hesingh:
What about adding support for isValid()?

mbudiu-vmw (approved):

hesingh:
Later, please also think about adding support for isHit() and isActionRun() - thanks.

hanw:
> 
> 
> What about adding support for isValid()?

isValid() is already supported.

hanw:
> 
> 
> Later, please also think about adding support for isHit() and isActionRun() - thanks.

Will do.

Issue #2670, opened 2021-02-11
expressionPrecedence broken for Operation_Binary in UBPF p4c backend


hesingh:
I have attached a P4 program in this issue.

The following line in the P4 code causes an interesting C code generated by the UBPF backend. The issue should be reproducible even with the EBPF backend.

```p4
headers.ipv6.ip_version_traffic_class_and_flow_label[31:12] = 52;
```
The generated C code in included below.

```c
 headers.ipv6.ip_version_traffic_class_and_flow_label = headers.ipv6.ip_version_traffic_class_and_flow_label & ~4294963200 | (uint32_t)52 << 12 & 4294963200;
```
Issue are:

First, gcc fails compiling unless parenthesis are added as follows:

 headers.ipv6.ip_version_traffic_class_and_flow_label = (headers.ipv6.ip_version_traffic_class_and_flow_label & ~4294963200) | ((uint32_t)52 << 12 & 4294963200);

`p4c/backends/ebpf/codeGen.cpp` Operation_Binary preorder needs change and maybe other methods.

Second, there are junk numbers after the `&` operator.

This is how I test:

```bash
p4c/build/$ ./p4c-ubpf --std p4-16 ../testdata/p4_16_samples/op_bin.p4 -o tst.c
```

[op_bin-p4.txt](https://github.com/p4lang/p4c/files/5968340/op_bin-p4.txt)

hesingh:
On this line of p4c code:

https://github.com/p4lang/p4c/blob/master/backends/ebpf/codeGen.cpp#L53

`prec` is 1 and `b->getPrecedence()` is 8 and no parenthesis are added but needed. 

Of course, these code changes to p4c fix the parenthesis issue:

```c
diff --git a/backends/ebpf/codeGen.cpp b/backends/ebpf/codeGen.cpp
index 3e84cb1ba..d89319f7d 100644
--- a/backends/ebpf/codeGen.cpp
+++ b/backends/ebpf/codeGen.cpp
@@ -51,7 +51,7 @@ bool CodeGenInspector::preorder(const IR::Operation_Binary* b) {
     widthCheck(b);
     int prec = expressionPrecedence;
     bool useParens = prec > b->getPrecedence();
-    if (useParens)
+    //    if (useParens)
         builder->append("(");
     visit(b->left);
     builder->spc();
@@ -59,7 +59,7 @@ bool CodeGenInspector::preorder(const IR::Operation_Binary* b) {
     builder->spc();
[...]

PR #2671, opened 2021-02-12
Fix for p4c Issue 2670

hesingh:
Add parenthesis to Operation_Binary to fix gcc compile error reported in https://github.com/p4lang/p4c/issues/2670

`expressionPrecedence` doesn't work for Operation_Binary yet that would also fix the gcc compile error.  

Another fix is to print, e.g., `0xffff0000` instead of its decimal format.

### DIFF
--- a/backends/ebpf/codeGen.cpp
+++ b/backends/ebpf/codeGen.cpp
@@ -27,7 +27,9 @@ void CodeGenInspector::substitute(const IR::Parameter* p, const IR::Parameter* w
 { substitution.emplace(p, with); }
 
 bool CodeGenInspector::preorder(const IR::Constant* expression) {
-    builder->append(Util::toString(expression->value, 0, false));
+    // Until EBPF/UBPF applications set the IR::Constant base to hex vs.
+    // decimal all emits will use hex.
+    builder->append(Util::toString(expression->value, 0, false, 16));
     return true;
 }
 
@@ -49,19 +51,13 @@ bool CodeGenInspector::preorder(const IR::Declaration_Variable* decl) {
 
 bool CodeGenInspector::preorder(const IR::Operation_Binary* b) {
     widthCheck(b);
-    int prec = expressionPrecedence;
-    bool useParens = prec > b->getPrecedence();
-    if (useParens)
-        builder->append("(");
+    builder->append("(");
     visit(b->left);
     builder->spc();
     builder->append(b->getStringOp());
     builder->spc();
-    expressionPrecedence = b->getPrecedence() + 1;
     visit(b->right);
-    if (useParens)
-        builder->append(")");
-    expressionPrecedence = prec;
+    builder->append(")");
     return false;
 }
 
--- a/backends/ebpf/codeGen.cpp
+++ b/backends/ebpf/codeGen.cpp
@@ -27,9 +27,8 @@ void CodeGenInspector::substitute(const IR::Parameter* p, const IR::Parameter* w
 { substitution.emplace(p, with); }
 
 bool CodeGenInspector::preorder(const IR::Constant* expression) {
-    // Until EBPF/UBPF applications set the IR::Constant base to hex vs.
-    // decimal all emits will use hex.
-    builder->append(Util::toString(expression->value, 0, false, 16));
+    builder->append(Util::toString(expression->value, 0, false,
+                                   expression->base));
     return true;
 }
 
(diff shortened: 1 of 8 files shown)


backends/ebpf/codeGen.cpp
```diff
@@ -27,7 +27,9 @@ void CodeGenInspector::substitute(const IR::Parameter* p, const IR::Parameter* w
 { substitution.emplace(p, with); }
 
 bool CodeGenInspector::preorder(const IR::Constant* expression) {
-    builder->append(Util::toString(expression->value, 0, false));
+    // Until EBPF/UBPF applications set the IR::Constant base to hex vs.
+    // decimal all emits will use hex.
+    builder->append(Util::toString(expression->value, 0, false, 16));
```
mbudiu-vmw, line 32: Why is this necessary?

backends/ebpf/codeGen.cpp
```diff
@@ -49,19 +51,13 @@ bool CodeGenInspector::preorder(const IR::Declaration_Variable* decl) {
 
 bool CodeGenInspector::preorder(const IR::Operation_Binary* b) {
     widthCheck(b);
-    int prec = expressionPrecedence;
-    bool useParens = prec > b->getPrecedence();
-    if (useParens)
-        builder->append("(");
+    builder->append("(");
```
mbudiu-vmw, line 54: Perhaps you can review the code and remove all uses of expressionPrecedence and remove the field too.

backends/ebpf/codeGen.cpp
hesingh: Otherwise `4294963200` is emitted instead of `0xfffff000` for 

`headers.ipv6.ip_version_traffic_class_and_flow_label & ~4294963200`

mbudiu-vmw: Why is that incorrect?

hesingh: Won't be incorrect.  A mask used with `&` is more readable as hex.  Any other constant would also be printed in hex, but that is fine too.  Note, even the bmv2 JSON format uses hex to emit any mask.

mbudiu-vmw: This code is used for all constants, not just masks.
The right solution is to make sure that the constants created when expanding slices in the Lower pass actually use base 16.

backends/ebpf/codeGen.cpp
hesingh: If I remove all uses of expressionPrecedence, then all uses of its comparisons with object->getPrecedence() will be removed as well.  Can't have that or more gcc compiler failures crop up.  It's tricky.

mbudiu-vmw:
The reference output you have checked-in for op_bin is different from the one computed by the compiler, and this is why the tests fail.

backends/ebpf/codeGen.cpp
hesingh: With the code changes of this PR to cogeGen.cpp, another test breaks.  It's as follows:

The P4 code below

`if (meta->_acl_outer21 == 1)`

is emitted as 

`if ((meta->_acl_outer21 == 1))`

With double parenthesis, gcc will cause an error.

In a subsequent checkin, I will add the line above and another line that can break gcc compiling with wrong C code generation.  The checkin will also add a fix that fixes all three issues - the `&` line of code has appropriate parenthesis, and so will the two lines of if-statements.

backends/ebpf/codeGen.cpp
```diff
@@ -50,7 +51,10 @@ bool CodeGenInspector::preorder(const IR::Declaration_Variable* decl) {
 bool CodeGenInspector::preorder(const IR::Operation_Binary* b) {
     widthCheck(b);
     int prec = expressionPrecedence;
-    bool useParens = prec > b->getPrecedence();
+    bool useParens = false;
+    if ((prec > b->getPrecedence()) ||
```
mbudiu-vmw, line 55: I think that what you want here is to check whether the parent is an if statement:
`useParents = getParent<IR::IfStatement>() == nullptr;`
This will use parentheses for everything but not for the expression in an if condition.

mbudiu-vmw:
@hesingh : these tests failed because of the issue fixed by https://github.com/p4lang/p4c/pull/2676.
As soon as we merge that PR perhaps you can just rebase on master and we can merge this PR as well.

hesingh:
> @hesingh : these tests failed because of the issue fixed by #2676.
> As soon as we merge that PR perhaps you can just rebase on master and we can merge this PR as well.

ok.

mbudiu-vmw:
I have merged the other PR, please rebase this on master.

mbudiu-vmw (approved):

PR #2673, opened 2021-02-16
add return statement to ordered_set::emplace_back

### LINKED ISSUE #2672: ordered_set.emplace_back is missing a return statement
`ordered_set::emplace_back` is missing a return statement. It should return `std::pair<iterator, bool>` according to its declaration.

grg:
The emplace_back function is missing a return statement. Add the missing return statement.

Fixes #2672


jnfoster (approved):
Looks good.

PR #2676, opened 2021-02-24
Reference outputs changed

mbudiu-vmw:
@fruffy : our improved dead code detection has now removed some functions that have no side-effects.
So these four gauntlet tests fail. For now I have updated the reference outputs, but you may want to fix the tests so that the code is not actually dead. If the code is dead it also means that your stf files are not checking anything really.


fruffy (approved):
Ah yeah. I think these were tests that exercised some parsing issue. Will create a pull request that actually makes these tests do something.

Issue #2681, opened 2021-03-07
Compiler Bug compiling test program tuple4.p4


jafingerhut:
I was creating a new Ubuntu 20.04 Linux VM this weekend with most recent version of open source P4 dev tools compiled from source code and experienced a compiler bug when attempting to compile one of the newer test program tuple4.p4:
```
$ p4test --version
p4test
Version 1.2.0 (SHA: 01b63cb3d BUILD: RELEASE)

$ cd testdata/p4_16_samples
$ p4test tuple4.p4
In file: /home/vagrant/p4c/lib/crash.cpp:256
Compiler Bug: Exiting with SIGSEGV
```
I did not use any options to `cmake` or `make` when building -- just whatever the defaults there are.

jnfoster:
Weird. Could you run with `gdb` to point the finger at a specific place in the code?

jafingerhut:
I am not sure how much this gdb stack trace narrows down a code location, since it doesn't seem to have line numbers, only function names, but they might help.

I tried with the same version of p4c source code on both Ubuntu 18.04 and 20.04 systems (which probably have some differences in version of g++ installed, which I can check if it seems important).

On both systems, a similar error occurred when I built p4c with the default options to cmake and make.

On both systems, running p4test built using `cmake .. -DCMAKE_BUILD_TYPE=DEBUG` gives no error at all.
```
$ build/p4test --version
build/p4test
Version 1.2.0 (SHA: 01b63cb3d BUILD: RELEASE)

$ gdb build/p4test 
GNU gdb (Ubuntu 8.1.1-0ubuntu1) 8.1.1
Copyright (C) 2018 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.  Type "show copying"
and "show warranty" for details.
This GDB was configured as "x86_64-linux-gnu".
Type "show configuration" for configuration details.
For bug reporting instructions, please see:
<http://www.gnu.org/software/gdb/bugs/>.
Find the GDB manual and other documentation resources online at:
<http://www.gnu.org/software/gdb/documentation/>.
For help, type "help".
Type "apropos word" to search for commands related to "word"...
Reading symbols from build/p4test...done.
[...]

PR #2682, opened 2021-03-07
Update run-p4-sample.py

fruffy:
This pull request adds `-fsanitize=null` to the CI build process (only the CI build process) and also fixes an issue, where compiler bugs were incorrectly classified in the Python test scripts. The sanitize options by GCC and Clang seem fairly useful to find obscure bugs like these. However, they also increase compilation time substantially and, considering that the tests already run very slow, it makes sense to stick with only `-fsanitize=null`for now.  

This should be merged once #2681 and #2684 are resolved.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -17,7 +17,7 @@ ARG ENABLE_UNIFIED_COMPILATION=ON
 # Whether to enable translation validation
 ARG VALIDATION=OFF
 
-# Delegate the build to tools/travis-build.
+# Delegate the build to tools/ci-build.
 COPY . /p4c/
 WORKDIR /p4c/
-RUN chmod u+x tools/travis-build && tools/travis-build
+RUN chmod u+x tools/ci-build.sh && tools/ci-build.sh
--- a/README.md
+++ b/README.md
@@ -110,7 +110,7 @@ dot -Tpdf ParserImpl.dot > ParserImpl.pdf
 2.  Install [dependencies](#dependencies). You can find specific instructions
     for Ubuntu 16.04 [here](#ubuntu-dependencies) and for macOS 10.12
     [here](#macos-dependencies).  You can also look at the
-    [travis installation script](tools/travis-build).
+    [CI installation script](tools/ci-build.sh).
 
 3.  Build. Building should also take place in a subdirectory named `build`.
     ```
--- a/backends/bmv2/run-bmv2-test.py
+++ b/backends/bmv2/run-bmv2-test.py
@@ -254,9 +254,9 @@ def process_file(options, argv):
 
     if result != SUCCESS:
         print("Error compiling")
-        print("".join(open(stderr).readlines()))
+        print(open(stderr).read())
         # If the compiler crashed fail the test
-        if 'Compiler Bug' in open(stderr).readlines():
+        if 'Compiler Bug' in open(stderr).read():
             return FAILURE
 
     expected_error = isError(options.p4filename)
--- a/backends/p4test/run-p4-sample.py
+++ b/backends/p4test/run-p4-sample.py
@@ -264,9 +264,9 @@ def process_file(options, argv):
 
     if result != SUCCESS:
         print("Error compiling")
-        print("".join(open(stderr).readlines()))
+        print(open(stderr).read())
         # If the compiler crashed fail the test
-        if 'Compiler Bug' in open(stderr).readlines():
+        if 'Compiler Bug' in open(stderr).read():
             return FAILURE
 
     expected_error = isError(options.p4filename)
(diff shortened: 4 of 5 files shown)


hesingh:
The script in bmv2 backend has the same bug.  Please fix that script as well - thanks.

hesingh:
CI failed for the P4 program fruffy used to catch the failure.  Yay!

jafingerhut:
That is good, but the part that confuses me is why the test does not fail for the tuple4.p4 test program.

hesingh:
If you see the latest code changes of this PR, tools/travis-build is changed to build RELEASE.  So, if the new P4 file use to test is removed, CI should fail on tuple4.p4.

fruffy:
That was just a sanity check to test whether the current scripts catch compiler bugs at all. Still seems to work and probably also has in the past. So there is either something specific in the CI setup that causes the bug to disappear, or the libs that come with Ubuntu 16.04 simply do not trigger it. I do not have a VM at hand so I can not test that. The default compilation mode was already in RELEASE, but I made it explicit. 

We had several bugs in the past, which could only be produced in release mode. Not sure why, maybe a difference in garbage collection?

hesingh:
I don't understand Ubuntu 16.04.  CI tests OS show either Ubuntu 20.04 or Mac OS in this PR.  Let's keep this PR open until CI catches tuple4.p4.

hesingh:
CI invokes these two steps which I don't use in my p4 build.  I ran the steps with no luck - tuple4.p4 still crashes.

```bash
/usr/bin/git config --local gc.auto 0
tools/start_ccache
```

I see CI use a unified build.  I don't even know what a unified build is.

fruffy:
The script uses a custom Ubuntu 16.04 container that is defined here: https://github.com/p4lang/third-party/blob/master/Dockerfile#L27
The Github runner runs on Ubuntu 20.04, but that is unrelated to the Docker container running Ubuntu 16.04.

hesingh:
@jafingerhut If you patch these diffs in your p4c repo and build RELEASE, you won't be able to crash p4test with tuple4.p4. Something to this effect is happening in the CI build.

```c
diff --git a/CMakeLists.txt b/CMakeLists.txt
index ba27880e7..950200c3e 100644
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -192,6 +192,7 @@ add_cxx_compiler_option ("-Wall")
 add_cxx_compiler_option ("-Wextra")
 add_cxx_compiler_option ("-Wno-overloaded-virtual")
 add_cxx_compiler_option ("-Wno-deprecated")
+add_cxx_compiler_option ("-fno-delete-null-pointer-checks")
```

hesingh:
With the last checkin, tuple4.p4 fails CI, but so do other 99 tests.

fruffy:
Okay so a couple things

1. There seems to be some caching issue with Bazel that are unrelated to this pull request. 
2. Using `-fsanitize=null` for CI should catch issues like these even in Ubuntu 16.04. The runtime issue correctly points at the line where more modern versions of GCC cause a crash. 
3. It also caught another issue in the `json_loader.h`. It looks like `get_node` returns a `nullptr` sometimes, which hints at a JSON parsing problem. But I do not know how this loader is used. So, not sure how this can be fixed. I added a TODO line, but that was mostly just to see if `tuple4` is the only crash that remains after that.

mbudiu-vmw:
Please file a separate issue about json_loader.h

fruffy:
Okay so this pull request is basically done. 
It adds `-fsanitize=null` to the CI build process (only the CI build process) and also fixes an issue, where compiler bugs were incorrectly classified in the Python test scripts. The sanitize option by GCC and Clang seem fairly useful to find obscure bugs like this. However, they also increase compilation time substantially and, considering that the tests already run very slow, it makes sense to stick with only `-fsanitize=null`for now.  

This should be merged once #2681 and #2684 are resolved.

mbudiu-vmw (approved):

mbudiu-vmw:
@fruffy can you please solve the conflicts?

mbudiu-vmw:
the failures have funny error messages

fruffy:
Yeah I have no idea what this is.

fruffy:
Seems to be some issue with `p4c-dpdk` but the error message is not present even though we have `--output-on-failure`. I am guessing some segmentation fault?

fruffy:
Yeah it is a ` Segmentation fault (core dumped)`. @hanw can you check for 
./p4c-dpdk -I p4include testdata/p4_16_samples/psa-register-complex-bmv2.p4
and 
./p4c-dpdk -I p4include testdata/p4_16_samples/psa-register-read-write-bmv2.p4

PR #2683, opened 2021-03-08
Add correct Bazel cc toolchain dependency

katre:
In upcoming versions of Bazel, cc toolchains will be resolved via toolchain resolution, which requires the `toolchains` argument to the `rule` function.

The `incompatible_use_toolchain_transition` flag is set due to a migration, we will remove this when it is no longer needed (on the order of a half year, I suspect).


katre:
I signed it.

mbudiu-vmw:
@smolkaj can you please review this?

smolkaj:
LGTM.

mbudiu-vmw (approved):

PR #2685, opened 2021-03-11
Add regression script for dpdk backend

hanw:
Reusing the p4test script with minor modification

### DIFF
--- a/backends/dpdk/run-dpdk-test.py
+++ b/backends/dpdk/run-dpdk-test.py
@@ -1,0 +1,357 @@
+#!/usr/bin/env python3
+# Copyright 2013-present Barefoot Networks, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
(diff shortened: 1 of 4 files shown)


backends/dpdk/run-dpdk-test.py
```diff
@@ -0,0 +1,357 @@
+#!/usr/bin/env python3
```
mbudiu-vmw, line 1: I suspect much of this file is borrowed from the other run-* scripts. Ideally we should factor out the common parts into a module and put it in tools.

mbudiu-vmw (approved):

mbudiu-vmw:
It looks like you have a syntax error in your CMakefile

backends/dpdk/run-dpdk-test.py
fruffy: I started working on a template in the eBPF back end with run-ebpf-test.py, but never got around to generalizing it to the other back ends. Maybe one day I will get around to porting it.

hanw: @fruffy thanks, let me know when the template is ready.

PR #2690, opened 2021-03-15
Missing null pointer check

### LINKED ISSUE #2681: Compiler Bug compiling test program tuple4.p4

mbudiu-vmw:
Fixes #2681

### DIFF
--- a/frontends/common/constantFolding.cpp
+++ b/frontends/common/constantFolding.cpp
@@ -220,10 +220,12 @@ const IR::Node* DoConstantFolding::preorder(IR::ArrayIndex* e) {
                 return e;
             }
             auto value = getConstant(e->left);
+            if (!value)
+                return e;
             if (auto list = value->to<IR::ListExpression>()) {
                 if (static_cast<size_t>(index) >= list->size()) {
                     ::error(ErrorType::ERR_INVALID,
-                            "Tuple index %1% too large", e->right);
+                            "Tuple index %1% out of bounds", e->right);
                     return e;
                 }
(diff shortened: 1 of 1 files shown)


frontends/common/constantFolding.cpp
```diff
@@ -220,10 +220,12 @@ const IR::Node* DoConstantFolding::preorder(IR::ArrayIndex* e) {
                 return e;
             }
             auto value = getConstant(e->left);
+            if (!value)
```
fruffy, line 223: What causes this expression to be null in the first place?

fruffy (approved):

frontends/common/constantFolding.cpp
mbudiu-vmw: This returns null if the expression is not a constant.
For an array it will return a non-null constant only if the array is a list expression.
When we add array initializers we'll be able to have constant arrays too (I am not sure that having constant headers is useful, though, but it should work).

PR #2691, opened 2021-03-16
Do not dump json when program is null because of an error

### LINKED ISSUE #2684: Nullpointer dereference in json_loader.h
Enabling sanitizer checks on P4C (`-fsanitize=null`) caused an issue in the json loader. At [line 261](https://github.com/p4lang/p4c/blob/master/ir/json_loader.h#L261) a null pointer is sometimes dereferenced. This indicates that there was a parsing exception for the affected programs.

https://github.com/p4lang/p4c/runs/2059440163 contains a list of test that fail

mbudiu-vmw:
Fixes #2684

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -192,6 +192,7 @@ add_cxx_compiler_option ("-Wall")
 add_cxx_compiler_option ("-Wextra")
 add_cxx_compiler_option ("-Wno-overloaded-virtual")
 add_cxx_compiler_option ("-Wno-deprecated")
+# add_cxx_compiler_option ("-fsanitize=null")
 
 # If we're on GCC, use the gold linker if available.
 if(CMAKE_COMPILER_IS_GNUCC)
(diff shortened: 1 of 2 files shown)


CMakeLists.txt
```diff
@@ -192,6 +192,7 @@ add_cxx_compiler_option ("-Wall")
 add_cxx_compiler_option ("-Wextra")
 add_cxx_compiler_option ("-Wno-overloaded-virtual")
 add_cxx_compiler_option ("-Wno-deprecated")
+# add_cxx_compiler_option ("-fsanitize=null")
```
jnfoster, line 195: Why this line?

jnfoster (approved):

CMakeLists.txt
mbudiu-vmw: This is what makes this program crash.
@fruffy made a similar change in his testing code that will run on CI, but this is a reminder how it can be done when running normal tests too. We may consider adding this always as well (I haven't measured the cost.)

mbudiu-vmw:
This test fails with this error:
```

CMakeLists.txt
jnfoster: Ahh, makes sense. Thanks.

PR #2693, opened 2021-03-17
dpdk: support table.apply().hit and table.apply().miss expression in …

hanw:
…if statement


testdata/p4_16_samples_outputs/psa-table-hit.p4.spec
```diff
@@ -0,0 +1,76 @@
...
+	default_action ifHit args none 
+	size 0
+}
+
+
+apply {
+	rx m.psa_ingress_input_metadata_ingress_port
+	extract h
+	table tbl_0
+	jmph LABEL_0END
```
mbudiu-vmw, line 71: is this the correct output or is it inverted?

mbudiu-vmw (approved):

testdata/p4_16_samples_outputs/psa-table-hit.p4.spec
hanw: This is correct to my understanding, see comment from Crisitian below.

>“Jump if table lookup hit” (jmph) example:
table vxlan_table
//some instructions here, but none of them is “table”, as jmph and jmpnh will refer to the table from the last “table” instruction
jmph GO_HERE_ON_VXLAN_TABLE_HIT
//some instructions
GO_HERE_ON_VXLAN_TABLE_HIT:
//some instructions to execute on vxlan_table hit

hesingh:
In the P4 code, I don't see a test for table.apply().miss?

testdata/p4_16_samples_outputs/psa-table-hit.p4.spec
mbudiu-vmw: This still seems to me to be inverted. On hit you skip the tbl_ifHit application.

hanw: Thanks, I fixed the issue.

hanw:
> 
> 
> In the P4 code, I don't see a test for table.apply().miss?

Thanks, added test for apply().miss.

Issue #2679, opened 2021-03-01
Split CompilerOptions into frontend and parser options


robotrobo:
Hello,
a lot of our tools are implemented using p4c's infrastructure. This includes the CompilerOptions class. However, we only use the IR parser, not the front end. And the default options parser has too many options we do not need. Would it make sense to split the CompilerOptions into options only required for the parser and front end options, which are derived from the parser options? We can make a pull request for that.
@fruffy

mbudiu-vmw:
Sure, why not? If you make a good argument this could be done.

PR #2694, opened 2021-03-18
Seperate parser options from frontend options

robotrobo:
Implements issue #2679. The CompilerOptions class now inherits from a new class called ParserOptions that is defined in the parser_options file, which sits between Util::Options and CompilerOptions. This change should be invisible to other users, and all changes are local to the frontends folder. Note we also changed the signature of the parseP4File function in parseInput which will now take in a ParserOptions object instead of a CompilerOptions object as an argument. 
This would allow for an added level of granularity for extensions like ours in regards to the number of options that we have at hand, which would tidy up our usage message. 
@fruffy

### DIFF
--- a/frontends/CMakeLists.txt
+++ b/frontends/CMakeLists.txt
@@ -151,6 +151,7 @@ set (COMMON_FRONTEND_SRCS
   common/constantFolding.cpp
   common/constantParsing.cpp
   common/options.cpp
+  common/parser_options.cpp
   common/parseInput.cpp
   common/resolveReferences/referenceMap.cpp
   common/resolveReferences/resolveReferences.cpp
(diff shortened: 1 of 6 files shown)


mbudiu-vmw:
I am assuming that you have kept the same options but split them across multiple files. (It's not easy to review such diffs in detail)
You still have a cpplint failure. If there any other changes, please list them.

robotrobo:
Yes that is correct, they are now split into ParserOptions and CompilerOptions, CompilerOptions now derives from ParserOptions and ParserOptions derives from Util::Options ,
 There are no more changes. (Except the one mentioned in P4::parseP4File)

mbudiu-vmw (approved):

PR #2695, opened 2021-03-18
Fix Gauntlet install.

fruffy:
With the changes to the base image, this package was missing for the Gauntlet installation. This just installs 'software-properties-common'. Once we switch to Ubuntu 20.04 all of this will not be necessary anymore.


jnfoster (approved):

jnfoster:
LGTM

PR #2703, opened 2021-03-24
Upgrade cpplint to Python3

fruffy:
Migrate the custom `cpplint.py` p4c has to a more standard and recent version from [here](https://github.com/cpplint/cpplint/blob/develop/cpplint.py) that does not require Python 2. There were a couple small changes necessary. 
I was looking at the changes mentioned here: https://github.com/p4lang/p4c/pull/2286
The only big difference to the standard version are two customizations:
`filematch`:
https://github.com/p4lang/p4c/blob/master/CPPLINT.cfg#L12
I do not quite understand the purpose of it, so I have readded it to `cpplint.py`.
There is also [this custom check ](https://github.com/p4lang/p4c/commit/496c150d754368384ea00955184a2818be0bc9dc#diff-cc07e0218fa142ad884be87236d14fbeefcd49b8d624c96b2b93046256142409R5850)added by @ChrisDodd , which I do not have included yet. Should it be added for this new version? 

I also had to add the following filters to stop cpplint from complaining: 
filter=-build/header_guard
filter=-build/namespaces
filter=-readability/casting
Should any of these be applied?


hanw:
Tried this on our internal repo, it reported a few warnings like `Found C++ system header after other header.` I think we can fix them internally, otherwise looks good to me.

CPPLINT.cfg
```diff
@@ -1,12 +1,15 @@
...
+filter=-build/include_subdir
+filter=-build/c++11
 filter=-build/include_what_you_use
 filter=-legal/copyright
 filter=-runtime/int
 filter=-runtime/references
-filter=-readability/completely_reasonable_casts
 filter=-readability/todo
 filter=-readability/function
+filter=-build/header_guard
```
mbudiu-vmw, line 10: won't these filters ignore things we care about?

fruffy: I disabled them because they cause a deluge of complaints. The customized pylint uses a different style from the recommended one. I readded some of the customizations, but they cause issues. So the next push will definitely break.

Actually, I am confused about the existence of [some .hpp files](https://github.com/p4lang/p4c/tree/master/frontends/parsers/p4) in the compiler. They are the only ones in the code base. Should they be converted to h files?

mbudiu-vmw:
We should merge this. @fruffy can you please see what the failures are?

mbudiu-vmw (approved):

fruffy:
This custom pass that was added https://github.com/p4lang/p4c/pull/2703/files#diff-cc07e0218fa142ad884be87236d14fbeefcd49b8d624c96b2b93046256142409R6247 creates false positives in some cases and I do not know how to robustly fix it. For example, it complains that [this](https://github.com/p4lang/p4c/blob/master/ir/dump.cpp#L51) call is missing an override. I would like to disable it. 

Also I will add a diff file to keep track of the modifications to `cpplint.py`

PR #2706, opened 2021-03-25
Make v1model immutable

mbudiu-vmw:
In particular, v1model::instance::version was a global static variable. That's bad.
The weird thing is that some of the classes I have introduced to represent the two versions of v1model are currently not used.
@antoninbas some of these are in the control-plane API generation code.
In particular, V1MODEL2020 is never used, but all tests pass.

### DIFF
--- a/frontends/p4/fromv1.0/programStructure.cpp
+++ b/frontends/p4/fromv1.0/programStructure.cpp
@@ -609,7 +609,7 @@ void ProgramStructure::include(cstring filename, cstring ppoptions) {
 void ProgramStructure::loadModel() {
     // This includes in turn core.p4
     std::stringstream versionArg;
-    versionArg << "-DV1MODEL_VERSION=" << V1Model::instance.version;
+    versionArg << "-DV1MODEL_VERSION=20200408";
     include(V1Model::instance.file.name, versionArg.str());
 
     metadataInstances.insert(v1model.standardMetadataType.name);
--- a/frontends/p4/uniqueNames.h
+++ b/frontends/p4/uniqueNames.h
@@ -80,8 +80,14 @@ class FindSymbols : public Inspector {
     }
     void postorder(const IR::Declaration_Variable* decl) override
     { doDecl(decl); }
-    void postorder(const IR::Declaration_Constant* decl) override
-    { doDecl(decl); }
+    void postorder(const IR::Declaration_Constant* decl) override {
+        // Skip toplevel constants with names like __
+        // We assume that these do not clash and no new symbols with
+        // these names will be added.
+        if (decl->getName().name.startsWith("__") && getParent<IR::P4Program>())
+            return;
+        doDecl(decl);
+    }
     void postorder(const IR::Declaration_Instance* decl) override
     { if (!isTopLevel()) doDecl(decl); }
     void postorder(const IR::P4Table* decl) override
--- a/midend/interpreter.cpp
+++ b/midend/interpreter.cpp
@@ -79,7 +79,7 @@ bool SymbolicValueFactory::isFixedWidth(const IR::Type* type) const {
 }
 
 unsigned SymbolicValueFactory::getWidth(const IR::Type* type) const {
-    type = typeMap->getType(type, true);
+    type = typeMap->getTypeType(type, true);
     if (type->is<IR::Type_Bits>())
         return type->to<IR::Type_Bits>()->size;
     if (type->is<IR::Type_Boolean>())
@@ -739,7 +739,7 @@ void ExpressionEvaluator::postorder(const IR::Operation_Relation* expression) {
 
 void ExpressionEvaluator::postorder(const IR::Member* expression) {
     auto type = typeMap->getType(expression, true);
-    if (type->is<IR::Type_MethodBase>()) {
+    if (type->is<IR::Type_MethodBase>() || type->is<IR::Type_Error>()) {
         // not really void, but we can't do anything with this anyway
         set(expression, SymbolicVoid::get());
         return;
@@ -783,7 +783,7 @@ bool ExpressionEvaluator::preorder(const IR::ArrayIndex* expression) {
     visit(expression->left);
     evaluatingLeftValue = lv;
     visit(expression->right);
-    return false;  // prune
+    return true;  // don't prune
 }
 
 void ExpressionEvaluator::postorder(const IR::ArrayIndex* expression) {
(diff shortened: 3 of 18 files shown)


midend/interpreter.cpp
```diff
@@ -79,7 +79,7 @@ bool SymbolicValueFactory::isFixedWidth(const IR::Type* type) const {
 }
 
 unsigned SymbolicValueFactory::getWidth(const IR::Type* type) const {
-    type = typeMap->getType(type, true);
+    type = typeMap->getTypeType(type, true);
```
fruffy, line 82: This seems unrelated to the rest of the changes. What is this for?

midend/interpreter.cpp
```diff
@@ -739,7 +739,7 @@ void ExpressionEvaluator::postorder(const IR::Operation_Relation* expression) {
 
 void ExpressionEvaluator::postorder(const IR::Member* expression) {
     auto type = typeMap->getType(expression, true);
-    if (type->is<IR::Type_MethodBase>()) {
+    if (type->is<IR::Type_MethodBase>() || type->is<IR::Type_Error>()) {
```
fruffy, line 742: Same here.

test/gtest/helpers.cpp
```diff
@@ -128,7 +128,7 @@ P4CTestEnvironment::P4CTestEnvironment() {
     // XXX(seth): We should find a more robust way to locate these headers.
     _coreP4 = readHeader("p4include/core.p4");
     _v1Model = readHeader("p4include/v1model.p4", true,
-                          "V1MODEL_VERSION", P4V1::V1Model::instance.version);
+                          "V1MODEL_VERSION", 20200408);
```
fruffy, line 131: Shouldn't this be using getV1ModelVersion?

fruffy (approved):
I do not know much about how the v1model versioning works or what it is used for. So my feedback on this is only high level.

frontends/p4/uniqueNames.h
```diff
@@ -80,8 +80,14 @@ class FindSymbols : public Inspector {
     }
     void postorder(const IR::Declaration_Variable* decl) override
     { doDecl(decl); }
-    void postorder(const IR::Declaration_Constant* decl) override
-    { doDecl(decl); }
+    void postorder(const IR::Declaration_Constant* decl) override {
+        // Skip toplevel constants with names like __
+        // We assume that these do not clash and no new symbols with
+        // these names will be added.
+        if (decl->getName().name.startsWith("__") && getParent<IR::P4Program>())
```
fruffy, line 87: Could this cause an issue with inlining and constants that have been declared with __?

hesingh:
The original intent to support more than one version of v1model.p4 makes sense to me.  How we support is another question.

midend/interpreter.cpp
mbudiu-vmw: Yes, this is correct. This file is in fact not currently used, but these changes were triggered by @VolodymyrPeschanenkoLitSoft
and his changes to the loop unrolling pass. This file is part of loop unrolling, and these changes are indeed fixing some unrelated bugs in this file.

test/gtest/helpers.cpp
mbudiu-vmw: No, getV1ModelVersion extracts the version from an existing loaded program, but here we haven't loaded any program yet. What we are doing is preparing to load the program using this specific version of v1model.

frontends/p4/uniqueNames.h
mbudiu-vmw: I don't think we ever move objects into the global namespace currently, so I don't think this is an issue with the current implementation. You are right that in the future this may cause problems. The right solution is to reserve some names in the spec for internal use of the compiler, and then we can guarantee that this is safe. I could make this safer by checking explicitly just for __v1model_version, although that won't work for other architectures. Maybe we should discuss this in the LDWG.

fruffy:
Sounds good. From my point, I have no further comments.

test/gtest/helpers.cpp
fruffy: I do not quite understand the versioning. Where does 20200408 come from? Last change to the v1model.p4 file?

mbudiu-vmw:
> 
> 
> I do not know much about how the v1model versioning works or what it is used for. So my feedback on this is only high level.

If you look at the file v1mode.p4 it starts this way: 

```
#ifndef V1MODEL_VERSION
#define V1MODEL_VERSION 20180101
#endif

const bit<32> __v1model_version = V1MODEL_VERSION;

#if V1MODEL_VERSION >= 20200408
typedef bit<9>  PortId_t;       // should not be a constant size?
#endif
```

So we have in fact two versions of v1model.p4 in the same file, depending on the macro V1MODEL_VERSION when you include the file. Currently there are only 2 versions, but the file can evolve further in the future.

When we compile P4-14 programs we convert them to P4-16 programs that need to use the 20200408 version, so that's why you see the macro being defined this way before invoking the preprocessor. This behavior is identical prior to this refactoring, the V1Model class had a default value for the version field of 20200408.

The problem we are fixing in this PR is that the field V1Model::instance.version was a mutable global, which could be used to carry information outside the IR or pass arguments, which is not a good idea if you want the code to be readable and especially thread-safe. We still have a V1Model::instance global, but now it's a constant. To figure out which version of v1model was used when compiling a program you have to look at the __v1model_instance global constant, which a visitor pass provides.

mbudiu-vmw:
I will merge this PR to unblock #2687

frontends/p4/fromv1.0/programStructure.cpp
```diff
@@ -609,7 +609,7 @@ void ProgramStructure::include(cstring filename, cstring ppoptions) {
 void ProgramStructure::loadModel() {
     // This includes in turn core.p4
     std::stringstream versionArg;
-    versionArg << "-DV1MODEL_VERSION=" << V1Model::instance.version;
+    versionArg << "-DV1MODEL_VERSION=20200408";
```
ChrisDodd, line 612: It would be better to leave V1Model::instance.version as a static constexpr in the class, rather than hardcoding this value here.

ChrisDodd:
> In particular, v1model::instance::version was a global static variable. That's bad.

I disagree with this -- having a global static constant is fine.  It might be questionable if it was mutable, but it is not.

mbudiu-vmw:
I was mutated during compilation.

mbudiu-vmw:
We can have two constexpr values for the two values of v1model that we support. But we should not have a global mutable static.

PR #2707, opened 2021-03-25
Fix Dpdk regression failure: Duplicates declaration

usha1830:
This commit fixes the "Duplicates declaration error with p4c-dpdk.
The changes are made as per the comment specified in backends/dpdk/dpdkArch.h.
All header and metadata references should be prefixed with "h." and "m." respectively.


mbudiu-vmw (approved):

hanw:
looks like the fix is causing other issues, perhaps you can continue with debugging the 'declaration not found' and fix both in the same PR?

usha1830:
I am looking into the failures

PR #2708, opened 2021-03-25
Replace "master" with "main" 

jnfoster:
This is in preparation for renaming p4c's primary branch. 

In particular, update references to "master" in the CI scripts. Also update references to main branch in behavioral-model.

This PR should be merged at the same time as we (manually) rename the branch on GitHub's website.

### DIFF
--- a/.github/.travis.yml.disabled
+++ b/.github/.travis.yml.disabled
@@ -2,7 +2,7 @@
 
 branches:
   only:
-  - master
+  - main
   - stable
 
 dist: xenial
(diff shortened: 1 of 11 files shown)


mbudiu-vmw (approved):

jnfoster:
CI seems stalled here. Any way for me to see why? Presumably it's because I renamed the branches in the config files... 

Not sure how to unblock. I could use admin privileges to merge?

mbudiu-vmw:
If you have run all the tests locally then you could try merging by hand.
I don't understand why the tests are stalled. We had lots of PRs lately, could github be upset that we use too much CPU for free?

fruffy:
It is probably related to the branch protection rules in the settings. Those expect these tests to run. However, they do not since they now only run for main. This pull request still is against master though so no CI is triggered.

PR #2711, opened 2021-03-26
Some build fixes

dreibh:
- Improved CMakeLists.txt:
  * Added support for creating a tarball with "make dist".
  * Need to compile with -Wno-deprecated-declarations.
- Fixed define in includes to correspond to the actual include file name.

### DIFF
--- a/lib/algorithm.h
+++ b/lib/algorithm.h
@@ -14,8 +14,8 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-#ifndef P4C_LIB_ALGORITHM_H_
-#define P4C_LIB_ALGORITHM_H_
+#ifndef _LIB_ALGORITHM_H_
+#define _LIB_ALGORITHM_H_
 
 #include <algorithm>
 #include <set>
@@ -76,4 +76,4 @@ inline Fn for_each(std::pair<Iter, Iter> range, Fn fn) {
 template<class Iter> Iter begin(std::pair<Iter, Iter> pr) { return pr.first; }
 template<class Iter> Iter end(std::pair<Iter, Iter> pr) { return pr.second; }
 
-#endif /* P4C_LIB_ALGORITHM_H_ */
+#endif /* _LIB_ALGORITHM_H_ */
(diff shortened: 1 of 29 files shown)


lib/algorithm.h
```diff
@@ -14,8 +14,8 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-#ifndef P4C_LIB_ALGORITHM_H_
-#define P4C_LIB_ALGORITHM_H_
+#ifndef _LIB_ALGORITHM_H_
```
mbudiu-vmw, line 17: All these changes in lib are fine, but they should make the reuse of these files by other projects harder. For example, it is much more likely that other projects have a macro `_LIB_ERROR_H_` than a macro `_P4C_LIB_ERROR_H_`.

mbudiu-vmw (approved):

lib/algorithm.h
dreibh: The change was necessary to pass the "cpplint" test, at least under Ubuntu 21.04.

Without this fix:
```"cpplint" start time: Mar 29 11:16 UTC
Output:
----------------------------------------------------------
/build/p4c-1.2.0+g202103252040~feb3f3/lib/algorithm.h:17:  #ifndef header guard has wrong style, please use: _BUILD_P4C_1_2_0_G202103252040_FEB3F3_LIB_ALGORITHM_H_  [build/header_guard] [5]
/build/p4c-1.2.0+g202103252040~feb3f3/lib/algorithm.h:79:  #endif line should be "#endif  /* _BUILD_P4C_1_2_0_G202103252040_FEB3F3_LIB_ALGORITHM_H_ */"  [build/header_guard] [5]
Total errors found: 2
<end of output>
Test time =  29.47 sec
----------------------------------------------------------
Test Failed.
"cpplint" end time: Mar 29 11:17 UTC
"cpplint" time elapsed: 00:00:29
----------------------------------------------------------

End testing: Mar 29 11:17 UTC

cpplint =  29.47 sec*proc
```

ChrisDodd: This is a bug in cpplint and should be fixed there.  Having an include guard that starts with `_` followed by an upper-case letter is illegal perr the C and C++ standard (such macros are reserved for use by the implementation)

Issue #2710, opened 2021-03-25
Name collisions during inlining controls


amreshk:
Not sure if this has already been discussed, but there seems to be an issue with the `InlineDriver` pass.

Here is an example,

The control 'Wrapped' code,

```
control Wrapped(inout bit<8> valueSet) {
    @name("set") action set_0(bit<8> value) {
        valueSet = value;
    }
    @name("doSet") table doSet_0 {
        actions = {
            set_0();
        }
        default_action = set_0(8w0);
    }
    apply {
        doSet_0.apply();
    }
}
```
gets inlined into the control 'Wrapper'

```
control Wrapper(inout bit<8> value) { 
    @name("wrapped") Wrapped() wrapped_0; 
    apply { wrapped_0.apply(value); } 
}
```
as,

```
control Wrapper(inout bit<8> value) {
    @name("wrapped") Wrapped() wrapped_0;
    @name("wrapped.set") action wrapped_set(bit<8> value) {
        value = value;
    }
    @name("wrapped.doSet") table wrapped_doSet {
        actions = {
            wrapped_set();
        }
        default_action = wrapped_set(8w0);
    }
    apply {
        {
            wrapped_doSet.apply();
        }
    }
}
```
'value' from control param collides with local 'value' inside action param and creates an invalid
```
    @name("wrapped.set") action wrapped_set(bit<8> value) {
        value = value;
    }
```
This gets optimized out eventually which is incorrect.

Does the pass need a fix to create new variable / param names for such collisions?

mbudiu-vmw:
If this report is correct this is a bug. We actually have a pass that gives unique names to various objects, but perhaps we don't use it appropriately for parameters.

amreshk:
@mbudiu-vmw can you point me to the pass in question?

mbudiu-vmw:
UniqueNames

fruffy:
Can you post a full program example? I would like to check this out.

amreshk:
@fruffy , copying p4 below 

```
#include <core.p4>
#include <psa.p4>

struct EMPTY { };

header hdr_t {
    bit<8> field;
}

control Wrapped(inout bit<8> valueSet) {
    @name("set") action set_0(bit<8> value) {
        valueSet = value;
    }
    @name("doSet") table doSet_0 {
        actions = {
            set_0();
        }
        default_action = set_0(8w0);
    }
   apply {
        doSet_0.apply();
   }
}

control Wrapper(inout bit<8> value) { 
[...]

PR #2713, opened 2021-03-26
Rename all params in FindParameters pass

amreshk:


mbudiu-vmw:
Let's see if these tests pass.

amreshk:
Seeing fails on name mismatch due to the pass now renaming all the action params,

```
-    @name(".setb1") action setb1_2(bit<8> val, bit<9> port) {
-        hdr.data.b1 = val;
-        standard_metadata.egress_spec = port;
+    @name(".setb1") action setb1_2(bit<8> val_1, bit<9> port_1) {
+        hdr.data.b1 = val_1;
+        standard_metadata.egress_spec = port_1;
```

fruffy:
Should the original name be preserved with an annotation? Are there any control plane programs relying on the name? Also this is missing the example program.

jfingerh:
action parameter names are represented in P4Info files that P4Runtime API controller software has full access to all of its contents, including the parameter names, as a (partial) description of the P4 program.

For example, this action in this p4c test program: https://github.com/p4lang/p4c/blob/master/testdata/p4_16_samples_outputs/action-two-params.p4#L46
```
    action actTbl(bit<24> id, bit<32> ip) {
    }
```
is represented by this part of the P4Info file generated by p4c.  Note the "id" and "ip" names, which are intended to match the original action parameter names.  (I do not recall whether P4_16 supports @name annotations on action parameters, but if so, then those annotated names should override the one in the P4 source code.): https://github.com/p4lang/p4c/blob/master/testdata/p4_16_samples_outputs/action-two-params.p4.p4info.txt#L33
```
actions {
  preamble {
    id: 23137650
    name: "MyIngress.actTbl"
    alias: "actTbl"
  }
  params {
    id: 1
    name: "id"
    bitwidth: 24
  }
  params {
    id: 2
    name: "ip"
    bitwidth: 32
  }
}
```

jfingerh:
@fruffy My previous comment does not definitively answer your question: "Are there any control plane programs relying on the name?", but it is extremely easy for a P4Runtime API controller to examine the P4Info file and examine names there, and to rely on particular names existing there, and/or to use names to help distinguish the semantics of which action parameters mean what.

jfingerh:
I should also add that the P4Info file is today generated not at the end of the last midend compiler pass.  If I recall correctly, it is generated from the IR of the input P4 program at the end of the last front end pass, before the first mid end pass.  Generating a correct expected P4Info file thus puts several constraints on what a front end pass is allowed to do.

fruffy:
Thanks, I was wondering about that. The runtime [relies](https://github.com/p4lang/p4c/blob/master/control-plane/p4RuntimeSerializer.cpp#L962) on [externalName](https://github.com/p4lang/p4c/blob/1b5722bbe2fc94db19c28f74069aa4982788c83b/ir/declaration.h#L36) to get a parameter name, but that uses annotations. I am surprised it works after the FrontEnd transformation. Annotations can also be used with parameters, but it is a little ugly.

mbudiu-vmw:
The `@name` annotations are for P4 source code and programmers.
The compiler also maintains internally the original name of each identifier to use in error reporting and in code generation.
If you look at the ID class you will notice that it has two strings inside: the current and original name.
The rules for disambiguation of identifiers are really dependent on the context. For example, in P4 identifiers in the same scope have to be disambiguated, but in the P4Runtime API the names of local variables are irrelevant, but action parameter names matter.

mbudiu-vmw:
The reference outputs are expected to change.
However, the p4info files should probably not change. If they do this is a problem.

amreshk:
Added the testcase from issue #2710 

The p4info has the original param name `value`,

```
  actions {
    preamble {
      id: 20299687
      name: "MyIC.wrapper.wrapped.set"
      alias: "set"
    }
    params {
      id: 1
      name: "value"
      bitwidth: 8
   }
 }

```

mbudiu-vmw:
It looks like this does not work. The code generated for BMV2 will have the new action argument names in some instances, and the control plane will report an error when using the original names.

mbudiu-vmw:
The right solution is for the compiler to use a `@name` annotation on action arguments and for the back-ends and the control plane to honor that. But I don't know if they do.

mbudiu-vmw:
I think I have a working solution, I will post it in a separate PR. It is based on this PR, with a few small changes for bmv2.

mbudiu-vmw:
Included in #2715

mbudiu-vmw:
I assume it's ok to close this.

amreshk:
yes, thank you.

PR #2715, opened 2021-03-29
Give unique names to action directionless parameters

### LINKED ISSUE #2710: Name collisions during inlining controls

mbudiu-vmw:
Fixes #2710
The code changes are small, but most reference outputs are changed.
I have also modified the test script to make sure that when new reference outputs are changed the p4info files do not change.

### DIFF
--- a/backends/p4test/run-p4-sample.py
+++ b/backends/p4test/run-p4-sample.py
@@ -119,7 +119,8 @@ def run_timeout(options, args, timeout, stderr):
 timeout = 10 * 60
 
 def compare_files(options, produced, expected, ignore_case):
-    if options.replace:
+    # p4info files should not change
+    if options.replace and "p4info" not in produced:
         if options.verbose:
             print("Saving new version of ", expected)
         shutil.copy2(produced, expected)
--- a/backends/bmv2/common/action.cpp
+++ b/backends/bmv2/common/action.cpp
@@ -183,7 +183,7 @@ ActionConverter::convertActionParams(const IR::ParameterList *parameters,
             ::warning(ErrorType::WARN_UNUSED, "Unused action parameter %1%", p);
 
         auto param = new Util::JsonObject();
-        param->emplace("name", p->name);
+        param->emplace("name", p->externalName());
         auto type = ctxt->typeMap->getType(p, true);
         // TODO: added IR::Type_Enum here to support PSA_MeterColor_t
         // should re-consider how to support action parameters that is neither bit<> nor int<>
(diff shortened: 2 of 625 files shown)


mbudiu-vmw:
This was built on #2713

backends/p4test/run-p4-sample.py
```diff
@@ -119,7 +119,8 @@ def target():
 timeout = 10 * 60
 
 def compare_files(options, produced, expected, ignore_case):
-    if options.replace:
+    # p4info files should not change
+    if options.replace and "p4info" not in produced:
```
fruffy, line 123: Is the check also implemented for the other back end `run-p4-sample.py` files?

mbudiu-vmw: I think that only this script generates and compares the p4info files.
This is the only script that generates and compares the reference output files.

fruffy: Oh I see. So when you have files such as `xyz-bmv2.p4` tests are run twice. Once with p4test and the other time with the bmv2 script.

fruffy (approved):

PR #2719, opened 2021-03-31
Missing override keyword

mbudiu-vmw:
Looks like the lastest merge broke the build for MacOs. I hope that this commit fixes it.

PR #2720, opened 2021-03-31
dpdk: support drop operation

hanw:


mbudiu-vmw:
@hanw, your build fails in the same way on MacOS. Do you happen to have a Mac machine to check why the dependencies are not being computed?

mbudiu-vmw:
See #2719

hanw:
Unfortunately, I don't have a mac at the moment.

backends/dpdk/dpdkHelpers.cpp
```diff
@@ -21,12 +21,34 @@ namespace DPDK {
...
+
 bool ConvertStatementToDpdk::preorder(const IR::AssignmentStatement *a) {
     auto left = a->left;
     auto right = a->right;
     IR::DpdkAsmStatement *i = nullptr;
-    // handle Binary Operation
-    if (auto r = right->to<IR::Operation_Binary>()) {
+
+    if (isAssignToDrop(left)) {
+        if (auto rhs = right->to<IR::BoolLiteral>()) {
```
mbudiu-vmw, line 40: This looks wrong - what if you change your mind and assign later in the pipeline "false"?

jfingerh: Han, I do not know the behavior of this 'drop' primitive action in DPDK, but if its behavior is "drop the packet immediately, and do not perform any more table lookups", then it definitely should not be part of the DPDK output until the last statement of the ingress control (or egress control) has been executed.  That is similar to how drop is done in all other P4 switch architecture I have seen -- drop takes affect at the end of a control, not in the middle of a control.  Controls are always completely executed until the end in P4.  (where "until the end" can be executing an "exit" statement, of course, but in the absence of that, sequential execution until the last statement).

hanw: This is not a problem for a hardware switch target, when a pipeline is always running at line rate.

jnfoster: This raises a minor semantic question: is an extern allowed to terminate execution of a P4 block? 

On the one hand, the spec says "extern functions are arbitrarily powerful: they can store information in global storage, spawn separate threads..." which suggests the answer might be yes.

On the other hand, the spec also says that "at runtime each P4 functional block is executed to completion" which suggests the answer might be no.

It's possible this question is resolved elsewhere in the spec. It wasn't clear to me as I poked around this morning.

I would observe that we also have an `exit` primitive which terminates execution of the current P4 block. So even if we determined the spec does not allow an extern to terminate execution, the desired semantics for DPDK could be obtained with `drop(); exit`.

jafingerhut: @hanw You are free to implement it that way, of course -- I won't step you.  If I were you, I would call that out as a difference in the behavior of your PSA target, versus what the PSA specification says.

jafingerhut: @jnfoster I had the same question, actually :-).  Related to this discussion on multicast https://github.com/p4lang/p4-spec/issues/893 it seems that several P4 experts are at least assuming it is not possible to call an extern method and have it return multiple times, only exactly once.  Causing an exit from a control as a result of calling an extern function /method seems a bit to me like "returning 0 times".

jafingerhut: @hanw I would also say "expensive" is a relative term here.  We are talking about a branch instruction, right?  Every tie I hear someone say "it is expensive", it would be better to quantify it, when it seems like the difference in performance might very well be fairly small.  If you have a P4 program that does 4 table lookups per packet on tables with reasonably large sizes, e.g. 10,000 IPv4 longest prefix match entries, can you even measure the difference in the PSA-compliant drop implementation vs. the modified semantics?

jafingerhut: @jnfoster @hanw Again, I do not know the behavior of the DPDK `drop` primitive action, but if it is as if `exit` were called immediately afterward in P4_16, note that I believe the proposed change Han is suggesting here is that in a PSA program where one does an _assignment statement_ (not an extern function call) like `stdmeta.drop = true;`, it will behave as if `exit` were _always_ executed immediately afterwards.  Han, if I'm wrong on that, feel free to explain what you are actually proposing changing here.  I am just giving my best guess, based on too little knowledge.

jnfoster: Ahh, if it's done in an assignment, then this is definitely a non-standard semantics.

hesingh: There is precedence with using an assignment for drop. See

https://github.com/barefootnetworks/Open-Tofino/blob/master/share/p4c/p4include/tofino.p4

Look for "drop" in  the file. PSA uses `ingress_drop(ostd);`

hesingh: Also, if P4 code is written with a permit() and drop() actions, for a table, the code does not incur any delay in per-packet processing.   

@hanw Could you please point me to a specification for DPDK that covers the `spec` file which the p4c dpdk backend generates? I'd like to understand dpdk better - thanks.

cristian-dumitrescu: @mbudiu-vmw  @jafingerhut @jnfoster Why would you ever change your mind once you assigned drop = true and set drop = false later? I know it is in the PSA spec, I know this fits well with switch pipelines, but for a CPU pipeline:

(1) it is just a waste of cycles for packets to be dropped: set the drop flag, waste cycles to apply the rest of the pipeline processing for a packet that will get dropped, then test the drop flag, then finally do the only sensible thing i.e. drop the packet;

(2) it also potentially has side effects, such as counting this packet that should not be counted, or changing the state of a meter, etc;

(3) more importantly, it is wasting cycles for packets that are not to be dropped: need to test the drop flag for all packets, and branches are very expensive on the CPU.

Also, the P4-16 spec is confusing for the exit statement: it basically says that the execution of the current control block and all its ancestors is terminated, but will the packet be dropped or will it continue with the deparser / the traffic manager / the egress control block?

mbudiu-vmw: That's how the programming language works. We do not dictate how people will use it.
The right way is to define a predefined action "dropAction()" (in some P4 library that you offer) which assigns the drop metadata to true and calls exit.
This action can be compiled the way you do. People can use it when this is what they want.
"exit" means "stop processing and leave all metadata the way it currently looks.
The metadata will dictate what happens to the packet.

jafingerhut: @cristian-dumitrescu If someone writing a P4_16 program for the PSA architecture wants to drop a packet and do no more processing, i.e. no more table apply's, no more counter or meter updates, etc., then they can assign a value of true to the drop standard metadata in PSA, and then execute an `exit` statement, which causes the enclosing control to immediately stop executing.  What happens to the packet after the control finishes executing (whether that is via executing an `exit` statement, or reaching the end of the `apply` block of the control) is not determined by the P4_16 language specification.  It is determined by the architecture being used.  The PSA specification has detailed pseudocode describing what happens to the packet after the ingress and egress pipelines are complete, which depends upon the final values of the standard metadata fields.

"Branches are very expensive on the CPU" - agreed that mispredicted branches cost on the order of 20 cycles on many deeply-pipelined general purpose CPU architectures.  Correctly predicted branches are cheap.  If most packets are not dropped, then the "not dropped" branch costs one instruction to test and take.

I am not arguing against optimization here -- I am arguing against claiming that one is implementing the PSA specification, but then changing it in several undocumented ways that will surprise anyone who has read the PSA specification and expects a device to follow it.  If you want to create your own P4 architecture and implement that for DPDK, you are welcome to do so.

jafingerhut: @cristian-dumitrescu @hanw Note that if you want to faithfully implement the PSA architecture, then if the compiler back end sees the _combination_ of statements `stdmeta.drop = true; exit;` somewhere, then except for checking whether the `clone` standard metadata flag is true or false and creating a clone if it is true, you are completely done with everything, both on ingress and egress.

If you see `stdmeta.drop = true;` with _no_ `exit` statement after it, then by the P4_16 language specification the expected behavior is to continue executing the code of the enclosing control, because unless the P4 developer forgot the `exit` statement, they _intend_ for the rest of the code to be executed, and it would be incorrect to skip it.

Is it _common_ for a P4 developer to write a program that assigns `true` to the PSA `drop` standard metadata flag, then later change it to `false`?  I do not know.

hanw: I'm making changes to the PR to implement the PSA semantics. Will update the PR later today. Thanks for the good discussion.

hanw: I have updated the implementation, please review again. thanks.

jafingerhut: @cristian-dumitrescu My comment earlier about defining a different P4 architecture was seroius, by the way.  If you believe that a different P4 architecture that had, e.g. a drop() extern function that could not be undone, and if followed by an `exit` statement could be very efficiently executed on a DPDK target, and if there are half a dozen other small optimization opportunities you can see to defining something differently than PSA does, e.g. maybe it doesn't make sense in a CPU-based DPDK target to have egress processing at all, but only ingress processing, or at least the option during ingress to indicate that you want to skip egress processing completely, then those little things could very well add up to a different P4 architecture that is better suited to DPDK.  It will not be PSA, or v1model, or TNA, or PNA -- it will be its own thing.  If you think there is benefit to creating such a thing, go for it!

hesingh: +1 to @jafingerhut

Create a new dpdk-model.p4 in p4c/p4include/.

It's fairly straightforward to change p4c to support a new model.p4.

backends/dpdk/dpdkHelpers.h
```diff
@@ -149,6 +149,7 @@ class ConvertStatementToDpdk : public Inspector {
     }
     void branchingInstructionGeneration(cstring true_label, cstring false_label,
                                         const IR::Expression *expr);
+    bool isAssignToDrop(const IR::Expression*);
```
jafingerhut, line 152: Is `isAssignToDrop` used anywhere any more, or is this obsolete?

testdata/p4_16_samples_outputs/psa-drop.p4.spec
```diff
@@ -0,0 +1,62 @@
...
+	default_action NoAction args none 
+	size 0
+}
+
+
+apply {
+	rx m.psa_ingress_input_metadata_ingress_port
+	extract h
+	table tbl_0
+	jmpneq LABEL_DROP m.psa_ingress_output_metadata_drop 0x0
```
jafingerhut, line 57: Is the assumption here that something outside of the .spec file will cause m.psa_ingress_output_metadata_drop to be initialized to 0 for each new packet processed?  How is that accomplished?

testdata/p4_16_samples_outputs/psa-unicast-or-drop-corrected-bmv2.p4.spec
```diff
@@ -44,10 +44,12 @@ apply {
 	cast  h.ethernet.dstAddr bit_32 m.psa_ingress_output_metadata_egress_port
 	jmpneq LABEL_0END h.ethernet.dstAddr 0x0
 	mov m.psa_ingress_output_metadata_drop 1
-	LABEL_0END :	emit h.ethernet
+	LABEL_0END :	jmpneq LABEL_DROP m.psa_ingress_output_metadata_drop 0x0
+	emit h.ethernet
 	extract h.ethernet
 	emit h.ethernet
 	tx m.psa_ingress_output_metadata_egress_port
```
jafingerhut, line 51: Is the behavior of the `tx` primitive such that execution stops there?  In other words, what prevents `tx` from executing, and then falling through to execute the `drop` statement below?

hanw: yes, the processing will terminate after `tx`.

testdata/p4_16_samples_outputs/psa-drop.p4.spec
hanw: Good catch, I added initialization to drop right after the `rx` instruction.

backends/dpdk/dpdkHelpers.h
hanw: it's obsolete, removed.

testdata/p4_16_samples_outputs/psa-isvalid.p4.spec
```diff
@@ -46,10 +46,12 @@ table tbl {
 
 apply {
 	rx m.psa_ingress_input_metadata_ingress_port
+	mov m.psa_ingress_output_metadata_drop 0x0
 	extract h
 	jmpv LABEL_0END h
-	table tbl_0
+	table tbl
 	LABEL_0END :	tx m.psa_ingress_output_metadata_egress_port
+	LABEL_DROP : drop
```
jafingerhut, line 54: Should this .spec file have a conditional jump based on the value of m.psa_ingress_output_metadata_drop that might jump to LABEL_DROP?  Or has it been optimized out for a valid reason?

hanw: the original program does not have any action that drop of the packet. I would say line 54 is dead code.

jafingerhut: So are you saying that your _back end_ code somehow checked that there were no assignments to the drop field, and thus avoided generating a jump instruction based on its value?

hanw: yes, for some reason, the optimization remove the jmp instruction. I can look into that as a separate PR.

jafingerhut (approved):
The .spec output files all look good to me, with the one exception where it seems fishy that there is no conditional jump instruction testing the value of the drop field.  I am happy if a separate issue is created to look into that, to see if it is a bug, or normal behavior.

hanw:
thanks!

testdata/p4_16_samples_outputs/psa-isvalid.p4.spec
jafingerhut: I created an issue to track this here: https://github.com/p4lang/p4c/issues/2729

PR #2721, opened 2021-04-01
More options cleanup

fruffy:
Clean up options some more. Push some more options down from the parser options and differentiate the additional usage section from the rest of the arguments.

### DIFF
--- a/frontends/common/options.cpp
+++ b/frontends/common/options.cpp
@@ -20,7 +20,7 @@ limitations under the License.
 CompilerOptions::CompilerOptions() : ParserOptions() {
     registerOption(
         "--excludeFrontendPasses", "pass1[,pass2]",
-        [this](const char *arg) {
+        [this](const char* arg) {
             excludeFrontendPasses = true;
             auto copy = strdup(arg);
             while (auto pass = strsep(&copy, ","))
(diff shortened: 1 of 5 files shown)


fruffy:
After #2687 has been merged this is now ready.

frontends/common/options.cpp
```diff
@@ -17,59 +17,156 @@ limitations under the License.
...
-                       } else if (!strcmp(arg, "text")) {
-                           p4RuntimeFormat = P4::P4RuntimeFormat::TEXT;
-                       } else {
-                           ::error(ErrorType::ERR_INVALID, "Illegal P4Runtime format %1%", arg);
-                           return false;
-                       }
-                       return true; },
-                   "Choose output format for the P4Runtime API description (default is binary).\n"
-                   "[Deprecated; use '--p4runtime-files' instead].");
+    registerOption(
```
mbudiu-vmw, line 21: The way git shows diffs it's not easy to make sure that these two sets of options are the same.
I suspect you just moved them around. Can you make sure you didn't lose any?

fruffy: I just double checked and it looks okay but I can not make a 100% guarantee.

mbudiu-vmw (approved):

PR #2722, opened 2021-04-01
dpdk: fix table and action name to be original name, optimize direct action invocation call in apply block

hanw:
This PR implemented a couple fixes/optimizations to dpdk assembly.

- Output table and action name used in original P4 program, instead of compiler internal name.
- If an action is invoked directly in apply block, inline the action.


backends/dpdk/dpdkProgram.h
```diff
@@ -134,5 +135,51 @@ class ConvertToDpdkControl : public Inspector {
...
+// frontend generates multiple copies of the same action in the localizeActions pass,
+// they are not useful for dpdk datapath. Removed the duplicated actions in
+// this pass.
+class EliminateUnusedAction : public PassManager {
+    ordered_set<cstring> actions;
+
+ public:
+    EliminateUnusedAction() {
+        addPasses( {
+            new CollectActionUses(actions),
```
mbudiu-vmw, line 178: are you sure that the pass that eliminates unused declarations does not do this too?

midend/actionSynthesis.cpp
```diff
@@ -60,7 +60,7 @@ const IR::Node* DoMoveActionsToTables::postorder(IR::MethodCallStatement* statem
         ++arg;
     }
     BUG_CHECK(arg == mc->arguments->end(), "%1%: mismatched arguments", mc);
-    auto amce = new IR::MethodCallExpression(mc->srcInfo, mc->method, mc->typeArguments, otherArgs);
+    auto amce = new IR::MethodCallExpression(mc->srcInfo, actionPath, mc->typeArguments, otherArgs);
```
mbudiu-vmw, line 63: why change this?

mbudiu-vmw (approved):

backends/dpdk/dpdkProgram.h
hanw: I'm not sure if the unusedDeclarations pass visit the dpdk backend IRs, probably not.

mbudiu-vmw: I don't see why not. It will run reference resolution and then remove all declarations that are not referenced anywhere.

midend/actionSynthesis.cpp
hanw: The name of an action in action list is different from the same action in default action, e.g.
```
action = { action_1; }
default_action = action_1/action;
```

The name of the action has original name, but the action in action list has not.

I reverted the change in this PR, since I do not need the actionSynthesis pass in dpdk backend anyway.

PR #2725, opened 2021-04-06
Remove inlined constant from code

mbudiu-vmw:
@ChrisDodd: this addresses the comment you made in an already merged PR.

### DIFF
--- a/frontends/p4/fromv1.0/v1model.h
+++ b/frontends/p4/fromv1.0/v1model.h
@@ -309,7 +309,7 @@ class V1Model : public ::Model::Model {
     static V1Model instance;
     // The following match constants appearing in v1model.p4
     static const char* versionInitial;  // 20180101
-    static const char* version2020;     // 20200408
+    static const char* versionCurrent;  // 20200408
 };
 
 /// Stores the version of the global constant __v1model_version used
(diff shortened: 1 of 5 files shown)


frontends/p4/fromv1.0/v1model.h
```diff
@@ -307,6 +307,9 @@ class V1Model : public ::Model::Model {
     const DirectCounter_Model directCounter;
 
     static V1Model instance;
+    // The following match constants appearing in v1model.p4
+    static const char* versionInitial;  // 20180101
+    static const char* version2020;     // 20200408
```
ChrisDodd, line 312: This should probably be `versionCurrent` (or even just `version` -- not sure anyone needs other than the current version?)  This is being used primarily to get the version number to dump as a #define when we dump the IR as P4_16 code.

ChrisDodd (approved):

frontends/p4/fromv1.0/v1model.h
mbudiu-vmw: It is the current one, but if we add further variants suddenly old programs will stop working if current is always the latest.
I have come to appreciate specifying dependencies precisely since our builds fail due to random changes in the tools that we use.

ChrisDodd: The whole point of having the version number appear in the dumped code (the version used to translate/dump the code) is so that it will continue to work for later versions of the compiler.  Those later versions will use a later version number for themselves, but when picking up the older code will get the older version in it.

Issue #2727, opened 2021-04-06
Annotations are lost when a ControlBlock or ParserBlock is inlined


grg:
@mbudiu-vmw: You wrote most of the code below:
https://github.com/p4lang/p4c/blob/ca4915d41f7cd5329b49c4235b906b859c0620cd/frontends/p4/inlining.cpp#L437-L455

Annotations on the block's instanceType are dropped during inlining since they're not copied into the parent's instanceType. Is this intentional?

mbudiu-vmw:
Actually, this is an issue. Please add an example and mark it as bug.

grg:
Here's a sample program that demonstrates the issue for both a ControlBlock and a ParserBlock: [issue-2727.p4.txt](https://github.com/p4lang/p4c/files/6266675/issue-2727.p4.txt)
(Please remove the .txt extension. It wouldn't allow me to directly attach a .p4 file :( )

One question that needs to be addressed is _when_ an annotation should be preserved or not. For example, which annotations should we keep when inlining the following parser block?
```
@my_anno_1                                                                                                                                                                            
@name("inner_parser")                                                                                                                                                                 
parser InnerParser(...) { ... }
```
It's probably a mistake to copy the `@name(...)` annotation to the parent ParserBlock, but we might reasonably expect `@my_anno_1` to be preserved.

Run p4c as follows to see the problem in the transformed p4 code. All annotations on the inlined blocks are discarded.
```
p4c --top4 FrontEndLast issue-2727.p4
```

PR #2728, opened 2021-04-06
propagate annotations when inlining blocks (#2727)

grg:
Copy annotations when inlining ControlBlock and ParserBlock objects.

Please let me know what you think of this approach to solve #2727 

There's also the open question of whether we should propagate all or a subset of annotations. Any opinions on how to proceed here?

### DIFF
--- a/frontends/p4/inlining.h
+++ b/frontends/p4/inlining.h
@@ -252,6 +252,7 @@ will be enough.  Multiple iterations are necessary only when instances are
 passed as arguments using constructor arguments.
 */
 class Inline : public PassRepeated {
+    static std::set<cstring> noPropagateAnnotations;
  public:
     Inline(ReferenceMap* refMap, TypeMap* typeMap, EvaluatorPass* evaluator)
     : PassManager({
@@ -259,6 +260,16 @@ class Inline : public PassRepeated {
         // After inlining the output of the evaluator changes, so
         // we have to run it again
         evaluator }) {}
+
+    /// Do not propagate annotation \p name during inlining
+    static void setAnnotationNoPropagate(cstring name) {
+        noPropagateAnnotations.emplace(name);
+    }
+
+    /// Is annotation \p name excluded from inline propagation?
+    static bool isAnnotationNoPropagate(cstring name) {
+        return noPropagateAnnotations.count(name);
+    }
 };
 
 }  // namespace P4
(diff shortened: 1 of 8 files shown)


mbudiu-vmw:
You should add your example as a test in testdata/p4_16_samples.
I would also use a policy function to decide which annotations to skip copying. 
The main problem with block annotations is that the compiler does not understand them so they inhibit some optimizations. But ignoring them is not a solution.

mbudiu-vmw:
Also, please add the outputs generated when running your test.

grg:
I've added a mechanism to exclude propagation of specific annotations. Right now I have a static set containing the excluded annotation names in the `Inline` class, plus static accessors. Please let me know if you think this belongs elsewhere.

I've also added a test to verify that annotation propagation, and exclusion of the `@name` annotation from propagation.

mbudiu-vmw:
You have one failure because apparently the output that you checked in for the example does not match (auto-generated action names do not match).
Is this the output you expect? Notice that the parser annotation is not preserved, since the parser is not inlined as a block.
If you are happy with just control annotations this implementation may be what you want.
We should probably discuss this topic in the language design working group.

frontends/p4/inlining.h
```diff
@@ -225,7 +225,9 @@ class GeneralInliner : public AbstractInliner<InlineList, InlineSummary> {
     }
     // controlled visiting order
     const IR::Node* preorder(IR::MethodCallStatement* statement) override;
-    template<class T> void inline_subst(T *caller, IR::IndexedVector<IR::Declaration> T::*locals);
+    template<class T, class U>
+    void inline_subst(T *caller, IR::IndexedVector<IR::Declaration> T::*locals,
```
mbudiu-vmw, line 229: Please document the U argument.

mbudiu-vmw (changes requested):

frontends/p4/inlining.h
grg: I've replaced `T` and `U` with `P4Block` and `P4BlockType`, matching the definition in the cpp file. I've also copied the comment from the cpp file into the header.

grg:
> You have one failure because apparently the output that you checked in for the example does not match (auto-generated action names do not match).

I must have missed rerunning the test after my last change to the input file. Corrected.


> Is this the output you expect? Notice that the parser annotation is not preserved, since the parser is not inlined as a block.
> If you are happy with just control annotations this implementation may be what you want.
> We should probably discuss this topic in the language design working group.

The parser annotations (`@my_anno_1` and `@my_anno_2`) propagate through to the midend P4 file. I think this is all I need to get them to the Barefoot compiler 🤞
I'll double-check this in the next day or two.

mbudiu-vmw (approved):

mbudiu-vmw:
Should we merge this?

PR #2731, opened 2021-04-07
Enable static builds for p4c

fruffy:
Sharing compiler binaries across Linux distribution has been difficult because the libraries are included dynamically. Library versions (Boost, glibc, protobuf, etc) vary slightly, which often causes failures when running binaries that were compiled on a different system. This pull request adds an option to address this. It also includes link time optimization (`-flto`) , which reduces the size of the generated binaries considerably. 

~There is a problem with the garbage collection library though. Some of the symbols are already defined in `gc.cpp`. So the customizations have to be disabled. I do not understand how this compiles with dynamic builds in the first place, though. I do not want to disable GC completely, but I am not sure how to resolve this issue cleanly. Right now, I just disable specific customizations in the case that `GC_CPP_H` is already defined. But this does not seem correct. A similar problem applies to `backtrace.cpp`.~ This can be fixed by adding the `-Wl,-z,muldefs` flags to the linker, which allows multiple definitions. I do not know if that is the intended behavior but with this change, no modifications to `gc.cpp` and `backtrace.cpp` are necessary.


fruffy:
So the way I understand this is that the customizations in `gc.cpp` rebind the function symbols dynamically. This is why compiles when we link `p4ctoolkit`. However, when building static binaries, this does not work because of the One Definition Rule. `new`, `delete`, etc.  were already defined in `libgccpp.a`.

So these customizations need to be disabled for static builds. How important are they? @ChrisDodd do you maybe know?

hesingh:
I think what `p4c/lib/gc.cpp` has useful functionality for me.  But, for a static p4c build, I see no way out unless one skips linking `gc.cpp` from p4c.  One alternative is to rename functions in gc.cpp and then use these functions in rest of p4c source code - it's a mechanical change.  We should also link libgccp.a statically in a p4c build.

hesingh:
@fruffy Why not go ahead and make changes to create a static p4c build without `p4c/lib/gc.cpp` code?  At least it would be a start and one can test a static p4c build.  Thanks.

fruffy:
There is already a preprocessor directive that disables the GC customizations in the case a static build is created. I also added a CI test (release-p4c) that tests the build.

hesingh:
@fruffy I used your github repo to this PR to try and build a static build.  I used

```cmake -DENABLE_PROTOBUF_STATIC=ON BUILD_STATIC_RELEASE=ON ..`

But testing one app, I see it's not statically built.

```
ldd p4c-bm2-ss
	linux-vdso.so.1 (0x00007ffc1d981000)
	librt.so.1 => /lib/x86_64-linux-gnu/librt.so.1 (0x00007f9898d15000)
	libstdc++.so.6 => /usr/lib/x86_64-linux-gnu/libstdc++.so.6 (0x00007f989898c000)
	libgcc_s.so.1 => /lib/x86_64-linux-gnu/libgcc_s.so.1 (0x00007f9898774000)
	libpthread.so.0 => /lib/x86_64-linux-gnu/libpthread.so.0 (0x00007f9898555000)
	libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007f9898164000)
	libm.so.6 => /lib/x86_64-linux-gnu/libm.so.6 (0x00007f9897dc6000)
	/lib64/ld-linux-x86-64.so.2 (0x00007f989b85d000)
```

What am I doing wrong?  thanks.

**Comment by fruffy** [comment]
Made the settings very aggressive, could you try again?

**Comment by hesingh** [comment]
> Made the settings very aggressive, could you try again?

I updated my repo and resolved a conflict in CMakELists.txt, rebuilt, but still see dynamic linking.

```
$ ldd p4c-bm2-ss 
	linux-vdso.so.1 (0x00007fff86a6e000)
	librt.so.1 => /lib/x86_64-linux-gnu/librt.so.1 (0x00007f1818b33000)
	libstdc++.so.6 => /usr/lib/x86_64-linux-gnu/libstdc++.so.6 (0x00007f18187aa000)
	libgcc_s.so.1 => /lib/x86_64-linux-gnu/libgcc_s.so.1 (0x00007f1818592000)
	libpthread.so.0 => /lib/x86_64-linux-gnu/libpthread.so.0 (0x00007f1818373000)
	libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007f1817f82000)
	libm.so.6 => /lib/x86_64-linux-gnu/libm.so.6 (0x00007f1817be4000)
	/lib64/ld-linux-x86-64.so.2 (0x00007f181b729000)
```

**Comment by fruffy** [comment]
Hmmm I can not reproduce this. 
I get 
```
 ldd ./p4c-bm2-ss
        not a dynamic executable   
```
Have you rebuilt from scratch and tested the binaries in the build folder? Maybe there is mismatch.

**Comment by hesingh** [comment]
Let me try from scratch - thanks.

**Comment by hesingh** [comment]
> Hmmm I can not reproduce this.
> I get
> 
> ```
>  ldd ./p4c-bm2-ss
>         not a dynamic executable   
> ```
> 
> Have you rebuilt from scratch and tested the binaries in the build folder? Maybe there is mismatch.

What do you use for cmake?  I use `cmake -DENABLE_PROTOBUF_STATIC=ON BUILD_STATIC_RELEASE=ON ..`

**Comment by hesingh** [comment]
Clearly, since I use `cmake -DENABLE_PROTOBUF_STATIC=ON BUILD_STATIC_RELEASE=ON ..` this is not correct.  Could you please send me the steps you use for cmake and make. thanks.

**Comment by fruffy** [comment]
I used:
```
rm -rf build
mkdir build
cd build
cmake .. -DBUILD_STATIC_RELEASE=ON
make -j4
ldd ./p4c-bm2-ss
```

hesingh:
> I used:
> 
> ```
> rm -rf build
> mkdir build
> cd build
> cmake .. -DBUILD_STATIC_RELEASE=ON
> make -j4
> ldd ./p4c-bm2-ss
> ```

Your steps work for me! thanks.

mbudiu-vmw (approved):

PR #2733, opened 2021-04-09
Fix dpdk regression failure: declaration not found

usha1830:
This commit fixes the "declaration not found" error with p4c-dpdk. 
- Added a new preorder function which works on the Pathexpressions for replacing the header and metadata variable references.
- With this change, some of the tests in XFAIL list of dpdk backend are compiling fine. Updated the XFAIL list for the same.


hanw (approved):
Looks good to me. Please address the missing header type issue in a follow-up PR.

PR #2739, opened 2021-04-19
API for KV-tables

mihaibudiu:
Marking as draft PR, since this needs support for unions.
To be accompanied by support in the compiler for some back-ends and test cases.

### DIFF
--- a/p4include/kvtable.p4
+++ b/p4include/kvtable.p4
@@ -1,0 +1,101 @@
+#ifndef _KVTABLE_P4_
+#define _KVTABLE_P4_
+
+/*
+ * This is a target-independent P4 library that provides an abstraction
+ * of two table-like objects.
+ * Such tables are key-value mappings (like a HashMap).  They differ
+ * in the way they handle missing keys (which do not "hit" in the table).
+ * A compiler can transform all operations on such tables into
+ * operations on standard P4 tables.
+ */
+
+/// A type that has one single possible value: the empty tuple.
+typedef tuple<> Unit;
+
+/// An Either type has one of two possible values:
+/// - the left value
+/// - the right value
+union Either<LV, RV> {
+    LV left;
+    RV right;
+}
+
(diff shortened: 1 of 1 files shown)


jfingerh:
@mbudiu-vmw I was reminded of this PR in an Intel-internal discussion, and we may want to use a variant of one of these extern objects you propose, one that does not need the `union` keyword.

One variant of this could be restricted to only have exact match key fields, but in looking at your proposed extern object definition that takes a list of match_kind values, do you have an example instantiation of such an extern object that would use match_kind values other than `exact` that gives no errors with the latest `p4test` command?  I tried briefly to find one, but kept getting syntax error messages that appeared to come from the parser.

mbudiu-vmw:
We should write some tests. Your question is really independent on the patch, it would even apply to cases like:
```
extern void f<T>(in T data);
... 
f(ternary);
f(match_kind.ternary);
```

This should in principle work, but if it doesn't we may need to tweak the grammar.

jfingerh:
The program named v1model_valuelut_matchkinds2.p4 fails to compile with the latest p4test, and contains several commented-out lines that also fail to compile.  There might be some variation I have not tried that works, but if so, I am not sure what it would be.

The attached code includes a slightly different definition of a ValueExactLookupTable that avoids the issue, but is intended to be restricted to all `exact` match_kind values for all key fields.
[lookup-table-externs.zip](https://github.com/p4lang/p4c/files/8120836/lookup-table-externs.zip)

mbudiu-vmw:
perhaps you should file that as a new issue... It will get lost here. At least if we expect the program to compile.
Do we need to discuss this in the LDWG?

jfingerh:
Created this issue: https://github.com/p4lang/p4c/issues/3091

@mbudiu-vmw It would be wonderful if you could take a look at the example program and compile-time error messages attached to that issue, and see whether you think these look like they raise questions for the LDWG, or if they are perhaps easily-enhanced things in the existing compiler implementation.

PR #2741, opened 2021-04-20
Adding loops unrolling feature

VolodymyrPeschanenkoLitSoft:
These changes contain loops unroll feature from the master branch which is needed for future usage on dependent of this branch project.


jafingerhut:
The commits in this PR seem to be a mingling of some by you, and some by other p4c developers that I think are already merged in.  Were you perhaps developing on the main branch while getting updates from the main branch of the official repository?  It is far more common to see PRs that contain only commits by the submitter, without being mingled with commits to the main branch.  It makes the diff HUGE in this particular case, and difficult and/or impractical to review in its current form.

VolodymyrPeschanenkoLitSoft:
This branch is already in master. So, Is it possible to merge the bf-stable branch with the master? Speaking true, we need to update only two classes here: [parserUnroll](https://github.com/p4lang/p4c/blob/master/midend/parserUnroll.h) and [ExpressionEvaluator](https://github.com/p4lang/p4c/blob/master/midend/interpreter.h). What is the best solution for it:
1) to merge with the master
2) to merge with our branch which is already merged with the master
3) to update these two classes from the master.

Could you, please, advise?

jafingerhut:
Maybe try creating a fresh clone of the main branch of p4lang/p4c, and doing a 'diff' of that against your latest source code with your changes, and see how big that diff is?  If it contains only your changes, and is smaller than 1,789 changed files as this PR currently is, hopefully only a few files that have changed, then apply that diff to the unmodified p4lang/p4c clone and create a PR from that?

VolodymyrPeschanenkoLitSoft:
If I tried to merge bf-stable with the main brach the number of changed files is 1899. So, maybe the better solution is to make needed changes in the forked bf-stable and after that to make PR.

fruffy:
Someone else should probably update `bf-stable`, right? The simplest option is likely to `git reset --hard [commit_id]` and set bf-stable to be identical to the unrolling commit. Or just fast-forward to that point.

If not, the other option is to branch off `bf-stable,` then cherry pick the commit that adds the loop unrolling. This may or may not work, depending on what that commit relies on. There is no CI for `bf-stable` afaik. Another issue is that there will be diverging branches after this.

VolodymyrPeschanenkoLitSoft:
Probably, if somebody can merge bf-stable with the master then it will be good. “reset —hard” looks very dangerous for me, because I don’t known what changes were done there, and who are the users of this branch. I saw that the last changes on bf-stable were done by Jed and I”m not sure if these changes in master (for example). 

I think that if we make changes in bf-stable which connected to the loops unrolling only then it should be most safe.

So, I count upon your decision which will help me to add this feature into bf-stable.

jafingerhut:
@VolodymyrPeschanenko If you are actively coordinating efforts with people committing to the bf-stable branch, e.g. @hanw @ChrisDodd @grg then they can advise you better than I can on how to proceed.

PR #2742, opened 2021-04-20
dpdk: fix support for bitwise and/or/xor

hanw:


backends/bmv2/psa_switch/psaSwitch.cpp
```diff
@@ -474,6 +474,10 @@ void InspectPsaProgram::addTypesAndInstances(const IR::Type_StructLike* type, bo
                 LOG5("Field is a Type_Error");
                 pinfo->scalars_width += 32;
                 pinfo->scalarMetadataFields.emplace(f, newName);
+            } else if (ft->is<IR::Type_Enum>()) {
```
mbudiu-vmw, line 477: We have an enum elimination pass, you can use that one and they will never show up at the end.

mbudiu-vmw (approved):

PR #2744, opened 2021-04-21
Fix dpdk regression failure: Duplicates declaration

usha1830:
This commit fixes the "Duplicates declaration" error with p4c-dpdk.

Added new preorder functions for Type_Control and Type_Parser. These functions updates the names of header and metadata parameters to "h" and "m" respectively.

With this change, some of the tests in XFAIL list of dpdk backend are compiling fine. Updated the XFAIL list and sample outputs for the same.

### DIFF
--- a/backends/dpdk/dpdkArch.cpp
+++ b/backends/dpdk/dpdkArch.cpp
@@ -338,8 +338,8 @@ const IR::Node *ReplaceMetadataHeaderName::preorder(IR::Member *m) {
 
 /* This function replaces the header and metadata parameter names in control
  * blocks with "h" and "m" respectively.
- * It assumes that  ConvertToDpdkArch pass has converted the control blocks to
- * this form for DPDK Architecture
+ * It assumes that ConvertToDpdkArch pass has converted the control blocks to
+ * the following form for DPDK Architecture:
  *         control ingressDeparser(packet_out bufffer, header hdr, metadata md);
  *         control egressDeparser(packet_out bufffer, header hdr, metadata md);
  *         control ingress(header hdr, metadata md);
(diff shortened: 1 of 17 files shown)


jafingerhut:
@hanw Why do many of these PSA architecture test programs _not_ have a struct like `headers_t` that contains members, one for each header you want to be an output of the parser?  They seem to be using exactly one such header type instead.

I am not saying that there is anything written down in the PSA spec today that prohibits this (or maybe there is and I am forgetting it), but v1model does not support that, and I'm not sure whether TNA does, either.

Would it make the PSA back end any simpler if it only needed to handle a struct type only, one that could contain only headers and header stacks and header union members?

mbudiu-vmw:
cpplint does not like your program.

backends/dpdk/dpdkArch.cpp
```diff
@@ -336,15 +336,64 @@ const IR::Node *ReplaceMetadataHeaderName::preorder(IR::Member *m) {
...
+ * It assumes that  ConvertToDpdkArch pass has converted the control blocks to
+ * this form for DPDK Architecture
+ *         control ingressDeparser(packet_out bufffer, header hdr, metadata md);
+ *         control egressDeparser(packet_out bufffer, header hdr, metadata md);
+ *         control ingress(header hdr, metadata md);
+ *         control egress(header hdr, metadata md);
+*/
+const IR::Node *ReplaceMetadataHeaderName::preorder(IR::Type_Control *c) {
+    auto applyParams = new IR::ParameterList();
+    auto param0 = c->applyParams->parameters.at(0);
```
mbudiu-vmw, line 350: several of these have the opportunity to crash if applied to the wrong architecture.
in general it is more robust to check that parameters has at least one element, and that the type below is indeed a type_name and give an error if not.

backends/dpdk/dpdkArch.cpp
```diff
@@ -336,15 +336,64 @@ const IR::Node *ReplaceMetadataHeaderName::preorder(IR::Member *m) {
...
+    cstring param0_type = param0->type->to<IR::Type_Name>()->path->name;
```
mbudiu-vmw, line 351: this won't work if someone typedef-ed the type. It is more reliable to get the type from the typemap and then check that it is an extern type. But this will work for most cases.

backends/dpdk/dpdkArch.cpp
```diff
@@ -336,15 +336,64 @@ const IR::Node *ReplaceMetadataHeaderName::preorder(IR::Member *m) {
...
+
+    /* For IngressDeparser and EgressDeparser, first parameter is packet_out type.
+       This is followed by header and metadata parameters */
+    if (param0_type == "packet_out"){
```
mbudiu-vmw, line 355: our style rules require a space before the { and the else below on the same line

backends/dpdk/dpdkArch.cpp
```diff
@@ -336,15 +336,64 @@ const IR::Node *ReplaceMetadataHeaderName::preorder(IR::Member *m) {
...
+        auto header = c->applyParams->parameters.at(1);
```
mbudiu-vmw, line 356: same here, the robust way is to check that these exist first.

backends/dpdk/dpdkArch.cpp
```diff
@@ -336,15 +336,64 @@ const IR::Node *ReplaceMetadataHeaderName::preorder(IR::Member *m) {
...
+        auto local_metadata = c->applyParams->parameters.at(2);
+        header = new IR::Parameter(IR::ID("h"), header->direction, header->type);
+        local_metadata = new IR::Parameter(IR::ID("m"), local_metadata->direction, local_metadata->type);
+        applyParams->push_back(c->getApplyParameters()->getParameter(0));
+        applyParams->push_back(header);
+        applyParams->push_back(local_metadata);
+    }
+    else {
+        /* For Ingress and Egress blocks, there are only two parameters: header and metadata */
+        auto header = c->applyParams->parameters.at(0);
```
mbudiu-vmw, line 366: there is some opportunity to reuse the code between these two branches

mbudiu-vmw (approved):

hanw:
> @hanw Why do many of these PSA architecture test programs _not_ have a struct like `headers_t` that contains members, one for each header you want to be an output of the parser? They seem to be using exactly one such header type instead.
> 
> I am not saying that there is anything written down in the PSA spec today that prohibits this (or maybe there is and I am forgetting it), but v1model does not support that, and I'm not sure whether TNA does, either.
> 
> Would it make the PSA back end any simpler if it only needed to handle a struct type only, one that could contain only headers and header stacks and header union members?

no particular reason other than I was too lazy to type? =)

usha1830:
@mbudiu-vmw I modified the implementation a little. Please review if the changes look sane now. Thanks!

fruffy:
There is a merge conflict that needs to be resolved first. Also can #2707 be closed?

usha1830:
> There is a merge conflict that needs to be resolved first. Also can #2707 be closed?

Merge conflict resolved. Closed #2707.

backends/dpdk/dpdkArch.cpp
```diff
@@ -334,14 +334,69 @@ const IR::Node *ReplaceMetadataHeaderName::preorder(IR::Member *m) {
...
+ * It assumes that ConvertToDpdkArch pass has converted Ingress/Egress
+ * parser blocks to the following form for DPDK Architecture:
+ *         parser ingressParser(packet_in bufffer, header hdr, metadata md);
+ *         parser egressParser(packet_in bufffer, header hdr, metadata md);
+*/
+const IR::Node *ReplaceMetadataHeaderName::preorder(IR::Type_Parser *p) {
+    auto applyParams = new IR::ParameterList();
+    auto paramSize = p->applyParams->size();
+
+    if (paramSize == 3) {
```
mbudiu-vmw, line 386: Wouldn't you want to give an error if this is not true?
Here is how the bmv2 backend does error handling: https://github.com/p4lang/p4c/blob/5fa90dae7ebba31e5aa71bc7580edacb02ea502a/backends/bmv2/simple_switch/simpleSwitch.cpp#L43

jafingerhut:
> > @hanw Why do many of these PSA architecture test programs _not_ have a struct like `headers_t` that contains members, one for each header you want to be an output of the parser? They seem to be using exactly one such header type instead.
> > I am not saying that there is anything written down in the PSA spec today that prohibits this (or maybe there is and I am forgetting it), but v1model does not support that, and I'm not sure whether TNA does, either.
> > Would it make the PSA back end any simpler if it only needed to handle a struct type only, one that could contain only headers and header stacks and header union members?
> 
> no particular reason other than I was too lazy to type? =)

Thanks for the info.  I created a PR here to update these PSA test programs to define a struct of all parsed headers, as most P4 programs tend to do for other architectures: https://github.com/p4lang/p4c/pull/2746

backends/dpdk/dpdkArch.cpp
usha1830: Updated. I think at this point of compilation, it should be a bug if we do not have the right number of parameters. It is based on the assumption that a previous pass has converted the program to this expected format.

mbudiu-vmw: The rule of thumb I have followed is that users should never see a compiler bug message, no matter what input they supply to the compiler. To stress test this you can try invoking the dpdk compiler without using psa as a model. Just try compiling a regular v1model.p4 program with this backend.

usha1830: I got your point. I will update the BUG to a modelError as you suggested. But, that also raises a doubt about the usage of BUG for error handling. In what scenarios would we want to emit a message using BUG? I see a lot of places where it is used in dpdk backend.

I also tried compiling the v1model.p4 program with p4c-dpdk without --arch psa option, we currently emit a model error for PSA followed by a Compiler Bug.

[--Werror=Target model error] error: Package package V1Switch has no parameter named 'ingress'Are you using an up-to-date 'psa.p4'?
         DeparserImpl()) main;
In file: ../ir/ir.cpp:230
Compiler Bug: No parameter named ingress

Shouldn't we exit at the model error itself?

usha1830:
One of my commits [8dfdc6e](https://github.com/usha1830/p4c/commit/8dfdc6e87c6d00b20788d7e84c875ce3bb3a2462) hasn't triggered build and test for this PR. Any idea what could be wrong?

backends/dpdk/dpdkArch.cpp
mbudiu-vmw: You can choose where you want to exit.
Usually we finish the current pass and after each pass check for errors and exit.
This allows us to report multiple errors in one pass.
Another solution is to make the compiler bug conditional on no errors:
`BUG_CHECK(condition && errorCount() == 0, ...);`

hanw (approved):
LGTM.

PR #2745, opened 2021-04-22
dpdk: add support for header stack

hanw:


backends/dpdk/spec.cpp
```diff
@@ -157,12 +169,22 @@ std::ostream &IR::DpdkHeaderType::toSpec(std::ostream &out) const {
...
                 out << "header " << (*it)->name << " instanceof "
                     << t->path->name;
-            else {
-                std::cout << (*it)->type->node_type_name() << std::endl;
-                BUG("Unsupported type");
+            } else if (auto t = (*it)->type->to<IR::Type_Stack>()) {
+                if (!t->elementType->is<IR::Type_Name>())
+                    BUG("%1% Unsupported type", t->elementType);
+                cstring type_name = t->elementType->to<IR::Type_Name>()->path->name;
+                if (!t->size->is<IR::Constant>()) {
```
mbudiu-vmw, line 179: you'd think that dpdk would support dynamic indexes.

mbudiu-vmw (approved):

backends/dpdk/spec.cpp
hanw: True, will fix when there is a use case for it.

PR #2746, opened 2021-04-22
Update PSA test programs to always put parsed headers inside a struct

jafingerhut:
Many of these test programs were written so that a single Ethernet
header type was being passed as the parameter to top level parsers and
controls in the PSA architecture, rather than the more typical struct
of multiple headers.  This was not necessarily an error, but at least
the v1model architecture explicitly documents the user-defined headers
type H must be a struct containing header/stack/union types, but not
itself a header.  It would be best to avoid any special-case code in
the PSA back end to handle this unlikely use case.  I call it unlikely
because almost no one ever writes a program that only parses a single
header.

### DIFF
--- a/testdata/p4_16_errors/psa-meter2.p4
+++ b/testdata/p4_16_errors/psa-meter2.p4
@@ -15,16 +15,20 @@ header ethernet_t {
     bit<16>         etherType;
 }
 
+struct headers_t {
+    ethernet_t       ethernet;
+}
+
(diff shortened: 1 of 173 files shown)


mbudiu-vmw (approved):

hanw:
You need to enable DPDK backend with -DENABLE_DPDK=ON when bootstrapping.

hanw (approved):

PR #2751, opened 2021-04-28
Add two proposed target-specific customizations of psa.p4 include file

jafingerhut:


jafingerhut:
@hanw This is one possible idea for how to handle any desired differences between data-plane specific bit widths for different targets, e.g. for BMv2 psa_switch process, versus DPDK back end implementing PSA architecture.

One straightforward approach that was at least discussed in PSA meetings a long while back, but never actually carried out, is to have each target device create its own customized version of the psa.p4 include file, as demonstrated by the two files in this pull request.  We can certainly discuss other approaches, too -- this PR is primarily to get the ball rolling on discussion of what we think we should do.

hanw (commented):
It probably makes sense to factor out the common definitions into a separate file to avoid code duplication. 

How about in p4include/bmv2/psa.p4, we use macros to define the constants.
```
#define PORT_WIDTH 32
#include "common/psa.p4"
```

and in p4include/common/psa.p4

```
typedef bit<BITWIDTH> PortIdUint_t;
```

jafingerhut:
We can do that, but it wouldn't be surprising if one back end wanted to put some annotations on extern definitions / etc. that are different than what other PSA target device implementers might want, yes?  I don't think we want to limit those kinds of things.

I understand that _might_ be unusual, but if it were reasonably common, such a target device PSA implementer wouldn't be helped much by factoring out the bit widths into a separate file.

jfingerh:
Closing this PR, as Han's more complete PR has now been merged in: https://github.com/p4lang/p4c/pull/2818

PR #2754, opened 2021-04-30
Update StrengthReduction pass relative to signed shift left operation

CGeoffrion:
The current code does not slice a shift left operation if the source
field is signed, e.g.:
   unsigned_val[7:0] = (signed_val << 3)[15:8];
to something like this:
   unsigned_val[7:0] = signed_val[12:5];

This is not in-line with the P4 Spec "8.6. Operations on fixed-width
signed integers" chapter which state "Shifting left produces the exact
same bit pattern as a shift left of an unsigned value. Shift left can
thus overflow, when it leads to a change of the sign bit."

Slicing the operation also help reduce field constraints down the road.

### DIFF
--- a/frontends/p4/strengthReduction.cpp
+++ b/frontends/p4/strengthReduction.cpp
@@ -379,7 +379,7 @@ const IR::Node* DoStrengthReduction::postorder(IR::Slice* expr) {
     if (shift_of) {
         if (!shift_of->type->is<IR::Type_Bits>())
             return expr;
-        if (shift_of->type->to<IR::Type_Bits>()->isSigned)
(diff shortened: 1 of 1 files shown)


jafingerhut:
It would be best to add a test program and its expected output that fails without your changes, but passes with them.

CGeoffrion:
> It would be best to add a test program and its expected output that fails without your changes, but passes with them.

Well it depend on your backend. Having the signed and unsigned shift left using the same conversion simplify things and in our case it also reduce the data constraints. On the other hand, the original code or this fixed version produce good result from a functional perspective.

mbudiu-vmw:
If I understand right, this was not a bug, but a missed optimization opportunity.
So there is no case that fails when you don't do this.

mbudiu-vmw (approved):

PR #2760, opened 2021-05-05
dpdk: remove unreachable code after jmp instruction

hanw:


backends/dpdk/dpdkAsmOpt.cpp
```diff
@@ -172,4 +168,29 @@ const IR::Node *RemoveLabelAfterLabel::postorder(IR::DpdkListStatement *l) {
     return l;
 }
 
+const IR::Node* RemoveDeadCodeAfterUncondJmp::postorder(IR::DpdkListStatement *l) {
+    const IR::DpdkJmpLabelStatement *cache = nullptr;
+    auto new_l = new IR::IndexedVector<IR::DpdkAsmStatement>();
+    bool changed = false;
+    for (auto stmt : l->statements) {
```
mbudiu-vmw, line 175: This only works if there are no backwards jumps. Is that true?

PR #2762, opened 2021-05-05
dpdk: convert always-false verify to drop and remove always-true verify sta…

hanw:
…tement


mbudiu-vmw:
I don't think this is conformant with the PSA specification. On a "false" verify the ingress block should be entered with an error code as specified by the second argument. Drop is not the mandated action.

jfingerh:
I was looking through the PSA specification, and noticed that it is not as explicit as it could be on this point.  There is an example program in the specification where a parser uses `verify` to conditionally enter a reject state with a parser error, and later the control examines the parser error passed as an input to the control, from the parser, to decide what to do with the packet.

It does seem worth being more explicit about this in the PSA specification itself.  I have created an issue on the PSA spec for this here: https://github.com/p4lang/p4-spec/issues/936

mbudiu-vmw:
Can we close this PR?

PR #2763, opened 2021-05-05
dpdk: fix non-determinism in parser state assembly generation

hanw:


mbudiu-vmw (approved):

mbudiu-vmw:
cpplint does not like this

PR #2766, opened 2021-05-06
Eliminate redundant validation code.

jnfoster:
The `ifTrue` branch doesn't have a `NullOK` annotation, so the IR
generator already adds the same check:
```
void IR::IfStatement::validate() const {
        CHECK_NULL(condition);
        CHECK_NULL(ifTrue);
{ CHECK_NULL(ifTrue); } }
```

### DIFF
--- a/ir/ir.def
+++ b/ir/ir.def
@@ -444,7 +444,6 @@ class IfStatement : Statement {
         clone.visit(ifFalse, "ifFalse");
         v.flow_merge(clone);
     }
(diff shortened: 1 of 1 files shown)


mbudiu-vmw (approved):
This seems correct, but I wonder whether the implicit validation is equivalent to this one.

jnfoster:
Doesn't the generated code I typed in demonstrate that it is?

mbudiu-vmw:
True.

PR #2771, opened 2021-05-19
Add width_bits to varbits

fruffy:
`Type_Varbits` is missing an implementation for `width_bits()`. If you call it you will always receive zero, even if size is set. This requests adds the function call to `Type_Varbits`.

### DIFF
--- a/ir/type.def
+++ b/ir/type.def
@@ -110,6 +110,7 @@ class Type_Varbits : Type_Base {
     NullOK optional Expression  expression = nullptr;  // only used temporarily
     static Type_Varbits get(Util::SourceInfo si, int size = 0);
     static Type_Varbits get();
+    int width_bits() const override { return size; }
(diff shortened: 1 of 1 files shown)


mbudiu-vmw (commented):
I think this is on purpose. The dynamic width is not the same as the static width, and it is not well-defined what this function should return.

mbudiu-vmw:
According to section https://p4lang.github.io/p4-spec/docs/P4-16-v1.2.2.html#sec-ops-on-hdrs when measuring the header size the varbits fields are not counted. So probably this is the right behavior.
> The expression h.minSizeInBits() is defined for any value h that has a header type. The expression is equal to the sum of the sizes of all of header h's fields in bits, counting all varbit fields as length 0. An expression h.minSizeInBits() is a compile-time constant with type int.

fruffy:
Ah I missed that, thanks. Okay this request is not needed then.

PR #2772, opened 2021-05-21
Cleanups to help with resolving methodcalls without a typeMap

ChrisDodd:
Once type inferencing is finished, the typemap is no longer needed to resolve method calls, so this allows not providing one.  This change allows calling MethodInstance::resolve with a nullptr for the typeMap, so it can be used when a type map is not available.

Also clean up some redundant clones that were making things harder to debug (and just wasting space/time), and better allow ControlFlowVisitor to be extended to allow more general stuff.


mbudiu-vmw (approved):

ChrisDodd:
Looks like the MacOS builds are all broken by some protobuf issue?

PR #2774, opened 2021-05-25
A utility function is added to get the field offset of aggregate type…

maheswari-s:
For structs, the field offset is computed and returned.
For union, offset for all the fields is 0. So, the same function is defined under Type_HeaderUnion to return 0 for all the valid fields.
If the invalid field is given as argument, then -1 is returned.

### DIFF
--- a/ir/type.def
+++ b/ir/type.def
@@ -263,7 +263,12 @@ abstract Type_StructLike : Type_Declaration, INestedNamespace, ISimpleNamespace,
         }
         return -1;
     }
-    // This function returns offset in bits.
+    /// This function returns start offset of the given field name in bits.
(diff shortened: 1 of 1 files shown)


ir/type.def
```diff
@@ -263,6 +263,17 @@ abstract Type_StructLike : Type_Declaration, INestedNamespace, ISimpleNamespace,
         }
         return -1;
     }
+    // This function returns offset in bits.
```
mbudiu-vmw, line 266: Please use a /// comment to describe what this function does. You also need to document the -1 case.
You should also mention what this does when fields do not have a defined width (e.g., enums or varbits).

mbudiu-vmw (approved):

PR #2778, opened 2021-05-28
Conditional operator support in dpdk backend

usha1830:
This commit disables the PSA switch midend predication pass as  DPDK architecture does not support conditional operators. This change allows to retain the frontend transformation that converts the conditional operator into if-else statement.

Added a test and reference output for conditional operator.

### DIFF
--- a/backends/dpdk/midend.cpp
+++ b/backends/dpdk/midend.cpp
@@ -46,6 +46,7 @@ limitations under the License.
 #include "midend/noMatch.h"
 #include "midend/orderArguments.h"
 #include "midend/predication.h"
+#include "midend/parserUnroll.h"
 #include "midend/removeAssertAssume.h"
 #include "midend/removeLeftSlices.h"
 #include "midend/removeExits.h"
@@ -143,6 +144,7 @@ PsaSwitchMidEnd::PsaSwitchMidEnd(CompilerOptions &options,
             new P4::RemoveSelectBooleans(&refMap, &typeMap),
             new P4::FlattenHeaders(&refMap, &typeMap),
             new P4::FlattenInterfaceStructs(&refMap, &typeMap),
+            new P4::ParsersUnroll(true, &refMap, &typeMap),
             new P4::ReplaceSelectRange(&refMap, &typeMap),
             new P4::Predication(&refMap),
             new P4::MoveDeclarations(),  // more may have been introduced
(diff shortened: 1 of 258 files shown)


backends/dpdk/midend.cpp
```diff
@@ -146,7 +146,8 @@ PsaSwitchMidEnd::PsaSwitchMidEnd(CompilerOptions &options,
             new P4::FlattenInterfaceStructs(&refMap, &typeMap),
             new P4::ParsersUnroll(true, &refMap, &typeMap),
             new P4::ReplaceSelectRange(&refMap, &typeMap),
-            new P4::Predication(&refMap),
+            // DPDK architecture does not currently support predicated instructions
```
mbudiu-vmw, line 149: I don't really understand what the problem is. This generates conditional expressions ?:, and you seem to handle these just fine.

usha1830: DPDK does not have predicated instructions. 
So the statement val = cond ? a: b; 
needs to be converted into  if (cond) val = a; else val = b;

Looking at the intermediate dumps of frontend and midend passes, I see that frontend actually breaks the conditional operator into corresponding if-else and then later midend "predication" pass converts it back into conditional operator. I disabled the "Predication" pass as our ultimate goal is to convert the conditional operator into if-else which the frontend is already doing.

Please correct me if my understanding is wrong here and if you have any other suggestions for handling this.

Currently, the test program included in this PR is not compiling with p4c-dpdk as we do not handle it in the backend.

mbudiu-vmw: If you are sure that there is no other way ?: could show up this could be just fine.
Please rebase on main so that the macos tests have a chance to pass.

usha1830: I looked around the code further to confirm the code flow for conditional operator. The frontend always converts ?:; to if-else in the DoSimplifyExpressions pass and the backends which supports it reconverts it into ?:;

In case of dpdk backend, PSA midend passes already converts it back into ?:;, hence disabling this conversion seems fine to me.

I have rebased on main. All checks are passing now.

mbudiu-vmw: You don't need to keep the commented-out pass here, unless you plan to reinstate it eventually in the future, assuming that dpdk will support ?:.

mbudiu-vmw (approved):

Issue #2647, opened 2021-01-08
Some more predication issues


fruffy:
I have collected some more issues with the predication pass that I have stumbled upon recently. Unfortunately, it is hard for me to distinguish whether they are unique or not. However, it looks like they have a similar source issue. @anasyrmia would you mind taking a look at this? I am really testing the limits of this pass.

The following program:

```p4
bool simple_check() {
    return true;
}

action assign(inout bit<16> eth_t, inout bit<48> target_addr, bool check_bool) {
    if (check_bool) {
        if (simple_check() && 0xDEAD != eth_t) {
            target_addr = 48w1;
        }
    }
}
```
is transformed to 

```
control ingress(inout Headers h, inout Meta m, inout standard_metadata_t sm) {
    @name("ingress.tmp") bool tmp;
    @name("ingress.tmp_0") bool tmp_0;
    @name("ingress.hasReturned") bool hasReturned;
    @name("ingress.retval") bool retval;
    bit<16> eth_t;
    bit<48> target_addr;
    bool check_bool;
    @name(".assign") action assign() {
        eth_t = h.eth_hdr.eth_type;
        target_addr = h.eth_hdr.dst_addr;
        check_bool = true;
        {
            {
                {
                    bool cond;
                    cond = !tmp;
                    hasReturned = (check_bool ? true : hasReturned);
                    retval = (check_bool ? true : retval);
                    tmp = (check_bool ? retval : tmp);
                    tmp_0 = (check_bool ? (cond ? false : 16w0xdead != eth_t) : tmp_0);
                }
                {
                    target_addr = (check_bool ? (tmp_0 ? 48w1 : target_addr) : target_addr);
                }
            }
        }
        h.eth_hdr.eth_type = eth_t;
        h.eth_hdr.dst_addr = target_addr;
    }
    apply {
        assign();
    }
}
```
As far as I can tell, the problem is this assignment `cond = !tmp;`, where tmp is still undefined. So later expressions that depend on this value will be undefined. For example, `(check_bool ? (cond ? false : 16w0xdead != eth_t) : tmp_0)`
[predication_issue_1.p4.txt](https://github.com/p4lang/p4c/files/5787572/predication_issue_1.p4.txt)
[predication_issue_1.stf.txt](https://github.com/p4lang/p4c/files/5787573/predication_issue_1.stf.txt)

fruffy:
Another similar issue is this one:

```p4
struct Headers {
    ethernet_t eth_hdr;
[...]

PR #2779, opened 2021-05-28
Fix: Issue #2647 predication condition assignment of value

filipsyrmia:
This fixes the problem where a condition variable has it's value assigned before the variables that it depends on are defined.
For example, predication pass without this fix can produce something like this, even when the source code is written properly:
`
bool cond;
`
`
cond = !tmp;
`
`
tmp = ...;
`

With this fix, the output from the predication pass does not lead to undefined behaviour:
`
bool cond;
`
`
tmp = ...;
`
`
cond = !tmp;
`

  
 
This solution essentially changes only a small part of the code responsible for declaring and defining these conditions correctly(in IfStatement preorder in predication.cpp). The change delays the pushing of the definition for that condition by pushing it into "liveAssigns" vector instead, and the logic in the rest of the code(which is left unchanged) makes sure that the contents of that vector find their way onto the output BlockStatement in the correct order.

I also added comments in places that were lacking them, some log messages, shortened some parts of code where redundant variables were used(or same block of code was repeating several times when it could be substituted with only one block) and corrected some grammatical errors in variable names.


midend/predication.cpp
```diff
@@ -38,6 +39,8 @@ static cstring lvalue_name(const IR::Expression *exp) {
     return cstring();
 }
 
+// Remove all empty statements from the code.
+// For each assignmentStatement in the source code an emptyStatement is generated.
```
mbudiu-vmw, line 43: This comments seems wrong.
Also, please use /// comments when describing these functions.
Moreover, this code could be much simpler using a visitor: just return nullptr for an EmptyStatement. In fact, we have some other passes which do this already.

midend/predication.cpp
```diff
@@ -56,23 +59,28 @@ bool Predication::EmptyStatementRemover::preorder(IR::BlockStatement * block) {
     return false;
 }
 
+// Used to emplace appropriate values in the Mux expression, allows nesting of Mux expressions
```
mbudiu-vmw, line 62: Frankly this comment is rather vague, I am not sure what "appropriate" means.

midend/predication.h
```diff
@@ -22,7 +22,7 @@ limitations under the License.
 
 namespace P4 {
 
-/**
+/*
```
mbudiu-vmw, line 25: This docstring comment is intentional, using this form (or triple dashes) causes the tools to generate documentation for this function.

mbudiu-vmw (changes requested):
Please add the examples from #2647 as tests to see whether this fix addresses the problems.

filipsyrmia:
Hi @mbudiu-vmw I applied the requested changes and added all of the test files from the original issue. If anything else seems off please leave a comment.

mbudiu-vmw:
@fruffy I assume gauntlet is happy about these changes. Can you please confirm that all issues that you have identified with predication are addressed by this PR?

mbudiu-vmw (approved):

midend/predication.cpp
```diff
@@ -38,41 +39,37 @@ static cstring lvalue_name(const IR::Expression *exp) {
...
-            iter++;
-        }
-    }
-    return false;
+const IR::Node* Predication::EmptyStatementRemover::postorder(IR::EmptyStatement* statement) {
+    return nullptr;
 }
 
+const IR::Node* Predication::EmptyStatementRemover::postorder(IR::BlockStatement* statement) {
+    if (statement->components.empty()) return nullptr;
```
mbudiu-vmw, line 47: Note that sometimes empty blocks and empty statements cannot be removed safely, for example, if they are the only statement of an 'if-then' branch. Hopefully you know that this case cannot occur.

mbudiu-vmw:
Actually, I have one more request: #2647 and #2345 come with STF files that test end-to-end. Please add them to your tests for the corresponding fixes.

mbudiu-vmw (changes requested):
Please add the STF files from the fixed issues (see comment below).
The STF files contain input/output packets that test the functionality end-to-end.
It is enough to have the STF files added with names corresponding to the test files. E.g., test file x.p4 should have a corresponding x.stf.

fruffy:
I fuzzed this and found no new issues so far. Also all the programs I filed pass now.

midend/predication.cpp
filipsyrmia: Yes I'm aware of this, I implemented this class based on a pass that does a similar simplification and figured out that most checks it had are not needed here since the only empty statements and blocks that are being removed here are the ones that the Predication code itself created, so no problematic cases can occur.

filipsyrmia:
I added the STF files, hopefully all is good now.

fruffy:
I spoke too soon. I think there is another issue with this pass. Turned up after more fuzzing.
The following program:

```p4
    bool bool_val = h.eth_hdr.eth_type == 0xDE;
    action perform_action() {
        if (bool_val) {
            h.h[get_bound(1w0, 1w1)].a = 1;
        }
    }
    apply {
        perform_action();
    }
}
```
generates this sequence:

```p4
    @name("ingress.bool_val") bool bool_val_0;
    @name("ingress.tmp_0") bit<1> tmp;
    @name("ingress.tmp_1") bit<1> tmp_0;
    @name("ingress.val_0") bit<1> val_0;
    @name("ingress.bound_0") bit<1> bound_0;
    @name("ingress.hasReturned") bool hasReturned;
    @name("ingress.retval") bit<1> retval;
    @name("ingress.tmp") bit<1> tmp_1;
    @name("ingress.perform_action") action perform_action() {
        {
            {
                {
                    {
                        bool cond;
                        val_0 = (bool_val_0 ? 1w0 : val_0);
                        bound_0 = (bool_val_0 ? 1w1 : bound_0);
                        hasReturned = (bool_val_0 ? false : hasReturned);
                        cond = val_0 < bound_0;
                        tmp_1 = (bool_val_0 ? (cond ? val_0 : tmp_1) : tmp_1);
                        tmp_1 = (bool_val_0 ? (cond ? val_0 : bound_0) : tmp_1);
                    }
                }
            }
            hasReturned = (bool_val_0 ? true : hasReturned);
            retval = (bool_val_0 ? tmp_1 : retval);
            tmp = (bool_val_0 ? retval : tmp);
            tmp_0 = (bool_val_0 ? tmp : tmp_0);
            h.h[tmp_0].a = (bool_val_0 ? 8w1 : h.h[tmp_0].a);
        }
    }
    apply {
        bool_val_0 = h.eth_hdr.eth_type == 16w0xde;
        perform_action();
    }
}
```
Here the index `tmp_0` can be undefined if `h.eth_hdr.eth_type == 16w0xde` is false. No stf test yet. 


[2647d.p4.txt](https://github.com/p4lang/p4c/files/6584281/2647d.p4.txt)

filipsyrmia:
Doesn't seem like this new issue is caused by the code added by this fix. Nonetheless, I will start looking into a possible solution for this issue as well.

filipsyrmia:
I've figured out a possible solution but I would like to know your opinions on this idea before i start implementing it. @fruffy @mbudiu-vmw 

My main idea is to produce a solution which replaces this line:
`h.h[tmp_0].a = (bool_val_0 ? 8w1 : h.h[tmp_0].a)`
with something like this:
`h.h[(bool_val_0 ? tmp_0 : 1w0)].a = (bool_val_0 ? 8w1 : h.h[(bool_val_0 ? tmp_0 : 1w0)].a)` 
, since by doing this we avoid having an undefined value as an index for accessing an element of an array, by using the condition variable(s).

mbudiu-vmw:
I would like this better if the common subexpression would be assigned to a new temporary.
`tmp_1 = bool_val_0 ? tmp_0 : 1w0`.
But if it is correct, we'll take it.

mbudiu-vmw:
Should we merge this and file a new issue with the case that is not handled?

filipsyrmia:
I planned to just update this pull request with the new fix, which is now complete. I think it's better because I needed to generate new outputs for some of the test files already added in this pull request, since they include accessing of header stack elements and that logic was changed with the generating of new indexes.
If this is fine with you I will update this pull request.

filipsyrmia:
I have updated the PR, the new issue should now be fixed. I also added support for correctly transforming assignment statements that have a header stack on the 'left' side using a ternary op even when they are found in multiple nested If-Else blocks, since this feature was unsupported before.

midend/predication.cpp
```diff
@@ -25,9 +26,13 @@ static cstring lvalue_name(const IR::Expression *exp) {
         if (auto base = lvalue_name(m->expr))
             return base + "." + m->member;
     } else if (auto a = exp->to<IR::ArrayIndex>()) {
-        if (auto k = a->right->to<IR::Constant>())
-            if (auto base = lvalue_name(a->left))
-                return base + "." + std::to_string(k->asInt());
+                if (auto k = a->right->to<IR::Constant>()) {
```
mbudiu-vmw, line 29: your indentation is inconsistent.

midend/predication.cpp
```diff
@@ -199,11 +224,17 @@ const IR::Node* Predication::preorder(IR::AssignmentStatement* statement) {
...
-            new IR::AssignmentStatement(statement->left, rightStatement);
-        liveAssigns.push_back(liveAssignments[statementName]);
     }
+
+    // Push statement which now contains a Mux expression in the statement->right.
+    // ExpressionReplacer is applied here which takes care of correct transforming.
+    liveAssignments[statementName] =
+        new IR::AssignmentStatement(statement->left, clone(statement->right)->apply(replacer));
+    liveAssigns.push_back(liveAssignments[statementName]);
+    LOG1("Finished visiting statement");
```
mbudiu-vmw, line 234: In general we reserve level 1 for more important messages, perhaps you can make this into a higher level.

midend/predication.cpp
```diff
@@ -218,6 +249,38 @@ const IR::Node* Predication::preorder(IR::Member * member) {
...
+    while (auto ifs = findContext<IR::IfStatement>(ctxt)) {
+        conditions.push_back(ifs->condition);
+    }
+    // The index modification isn't necessary if the index is a Constant
+    if (!(arrInd->right->is<IR::Constant>()) && modifyIndex) {
+        // Creates a new variable that has the value of the original index when the
+        // condition is fulfilled, and in any other case it has a default value of '1w0'
+        cstring indexName = generator->newName("index");
+        auto indexDecl = new IR::Declaration_Variable(indexName,
+                                new IR::Type_Bits(arrInd->right->type->width_bits(), false));
```
mbudiu-vmw, line 264: you may be able to use arrInd->right->type->getP4Type() as well

mbudiu-vmw (changes requested):

filipsyrmia:
I applied the requested changes, and updated the test files so all checks should now pass. Let me know if you need anything else from me.

mbudiu-vmw:
It looks like the Gauntlet static checker has found a case where your transformation does not preserve the original meaning of the program.

fruffy:
Hmmm
I have been staring at this transformation 

```p4
@name("ingress.bool_val") bool bool_val_0;
@name("ingress.tmp_0") bit<1> tmp;
@name("ingress.tmp_1") bit<1> tmp_0;
@name("ingress.val_0") bit<1> val_0;
@name("ingress.bound_0") bit<1> bound_0;
@name("ingress.hasReturned") bool hasReturned;
@name("ingress.retval") bit<1> retval;
@name("ingress.tmp") bit<1> tmp_1;
@name("ingress.perform_action") action perform_action() {
    if (bool_val_0) {
        {
            val_0 = 1w0;
            bound_0 = 1w1;
            hasReturned = false;
            if (val_0 < bound_0) {
                tmp_1 = val_0;
            } else {
                tmp_1 = bound_0;
            }
            hasReturned = true;
            retval = tmp_1;
            tmp = retval;
        }
        tmp_0 = tmp;
        h.h[tmp_0].a = 8w1;
    }
}
apply {
    bool_val_0 = h.eth_hdr.eth_type == 16w0xde;
    perform_action();
}
```

to 

```p4
@name("ingress.bool_val") bool bool_val_0;
@name("ingress.tmp_0") bit<1> tmp;
@name("ingress.tmp_1") bit<1> tmp_0;
@name("ingress.val_0") bit<1> val_0;
@name("ingress.bound_0") bit<1> bound_0;
@name("ingress.hasReturned") bool hasReturned;
@name("ingress.retval") bit<1> retval;
@name("ingress.tmp") bit<1> tmp_1;
@name("ingress.perform_action") action perform_action() {
    {
        {
            {
                {
                    bool cond;
                    val_0 = (bool_val_0 ? 1w0 : val_0);
                    bound_0 = (bool_val_0 ? 1w1 : bound_0);
                    hasReturned = (bool_val_0 ? false : hasReturned);
                    cond = val_0 < bound_0;
                    tmp_1 = (bool_val_0 ? (cond ? val_0 : tmp_1) : tmp_1);
                    tmp_1 = (bool_val_0 ? (cond ? val_0 : bound_0) : tmp_1);
                }
            }
        }
        bit<1> index_0;
        hasReturned = (bool_val_0 ? true : hasReturned);
        retval = (bool_val_0 ? tmp_1 : retval);
        tmp = (bool_val_0 ? retval : tmp);
        tmp_0 = (bool_val_0 ? tmp : tmp_0);
        index_0 = (bool_val_0 ? tmp_0 : 1w0);
        h.h[index_0].a = (bool_val_0 ? 8w1 : h.h[index_0].a);
    }
}
apply {
    bool_val_0 = h.eth_hdr.eth_type == 16w0xde;
    perform_action();
}
```
For a while now, but I can not find a flaw. So I believe this is correct. There must be a bug in Gauntlet, although I do not understand how yet.

mbudiu-vmw:
Doesn't Gauntlet give you a counter-example indicating how these two programs could evaluate to different results?
In any case, it would be good to figure out which of the two is right.

fruffy:
Yes, the counterexample did not make sense to me. I have also checked with a generated STF test. It passes. So there is something subtle with undefined behavior going on. I will look at this with fresh eyes later, but for now I added the test to the xfails of Gauntlet so that this pull request passes.

filipsyrmia:
I'm guessing the problem is that in the original program there is no explicit access to `h.h[1w0].a`, but the index transformation is structured in a way where this access is possible if the right conditions are met. If so then the problem is the idea itself, which adds these 'default' index values if the original ones are undefined.

mbudiu-vmw:
So your program chooses some values for uninitialized variables.
In principle we could always do that, but people objected that initializing everything may be too expensive in some cases.

mbudiu-vmw:
Can you please add a list of all issues that are fixed by this PR so they can be closed automatically when we merge?

mbudiu-vmw (approved):

fruffy:
> I'm guessing the problem is that in the original program there is no explicit access to `h.h[1w0].a`, but the index transformation is structured in a way where this access is possible if the right conditions are met. If so then the problem is the idea itself, which adds these 'default' index values if the original ones are undefined.

Yes, but the value that is assigned is correct. If the function call is inlined, this should be the return value. So semantically, there is no difference from what I can tell. That is why I believe there must be a bug in the analysis tool. I need some more time to review the bug, but this pull request can be merged from my point of view.

filipsyrmia:
Do you need anything else from me before this PR is merged?

mbudiu-vmw:
References to the issues closed would be nice

filipsyrmia:
The original issue and the other issue covered by this fix are already linked, those are #2647 and #2345. The issue that @fruffy mentioned on this PR does not have it's own issue filed and therefore it isn't linked. I'm not sure I understand what you mean by 'references'.
Also, this PR appears closed, is that a mistake since I thought it was good to merge?

mbudiu-vmw:
Fixes #2647 
Fixes #2345

mbudiu-vmw:
There are 4 failed tests in this PR:
```
	903 - p4/testdata/p4_16_samples/predication_issue_2.p4 (Failed)
	1151 - p4/testdata/p4_16_samples/predication_issue_3.p4 (Failed)
	1393 - p4/testdata/p4_16_samples/predication_issue_4.p4 (Failed)
	1396 - p4/testdata/p4_16_samples/predication_issue_1.p4 (Failed)
```
It looks like you have to update the reference outputs. This may be due to other PRs that have been merged in the meantime.

filipsyrmia:
Updated the outputs, should be fine now.

Issue #2775, opened 2021-05-26
Fix use of removed constant google::protobuf::util::status_internal::Status::OK


kitlei-robert:
On line https://github.com/p4lang/p4c/blob/f824b59cb5bd3610ac04bee1e067817a2346500e/control-plane/p4RuntimeSerializer.cpp#L152 the constant `Status::OK` is used. However, it has been removed some 2 months ago, and should be replaced by `OkStatus()` as in https://github.com/protocolbuffers/protobuf/blob/fa31183250230b4af6ea485e23fb36e705b3f884/src/google/protobuf/stubs/status.cc#L124.

PR #2780, opened 2021-05-28
Attempt to fix Mac build breakage

mbudiu-vmw:
Instead of using brew to install the latest version of protobuf we compile explicitly the recommended version 3.6.1

### DIFF
--- a/tools/install_mac_deps.sh
+++ b/tools/install_mac_deps.sh
@@ -8,8 +8,18 @@ fi
 
 $BREW update
 $BREW install autoconf automake bdw-gc bison boost ccache cmake git \
-      libtool openssl pkg-config protobuf python
+      libtool openssl pkg-config python
 $BREW install gmp --c++11
+
+# Install specific version of Protobuf, since newer versions break compatibility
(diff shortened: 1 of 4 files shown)


mbudiu-vmw:
Fixes #2775 in a different way.

tools/install_mac_deps.sh
```diff
@@ -8,9 +8,19 @@ fi
 
 $BREW update
 $BREW install autoconf automake bdw-gc bison boost ccache cmake git \
-      libtool openssl pkg-config protobuf python
+      libtool openssl pkg-config python
 $BREW install gmp --c++11
 
+# Install specific version of Protobuf, since newer versions break compatibility
```
fruffy, line 14: Is it possible to just use brew like this?
https://stackoverflow.com/questions/3987683/homebrew-install-specific-version-of-formula

mbudiu-vmw: It may work but as far as I can tell there is no guarantee that the required version will be available.

fruffy (approved):

PR #2781, opened 2021-05-28
Expand structs into lists for bmv2 backend computations

### LINKED ISSUE #2657: Bmv2 Json file generated by p4c causes simple_switch throwing runtime error
Hi,

After I loaded a bmv2 json file which is generated by p4c without issues, simple_switch complained:
```
terminate called after throwing an instance of 'std::runtime_error'
  what():  Type is not convertible to string
Aborted
```
I pinpointed the bug to this statement:
```
hash<bit<16>, bit<9>, tuple_0, bit<18>>(tmp, HashAlgorithm.crc32, 9w0, { pkt.ethernet.etherType }, 18w512);
```
This statement looks semantically correct, and p4c complained nothing. After I changed this statement to:
```
hash(tmp, HashAlgorithm.crc32, 9w0, { pkt.ethernet.etherType }, 18w512);
```
simple_switch was able to load the new json file. Therefore, I think this might be some issue with p4c.

I have provided a minimal p4 file that can reproduce this: [bug.txt](https://github.com/p4lang/p4c/files/5870089/bug.txt)

mbudiu-vmw:
Fixes #2657 
As the issue discusses, there are actually 2 bugs in that issue, but one of them is addressed already by #2637 which is waiting for review, so I won't replicate that code here. This fix is however enough to make this test pass.

### DIFF
--- a/backends/bmv2/common/extern.cpp
+++ b/backends/bmv2/common/extern.cpp
@@ -237,9 +237,9 @@ ExternConverter::createCalculation(ConversionContext* ctxt,
         } else {
             for (auto f : type->to<IR::Type_StructLike>()->fields) {
                 auto e = new IR::Member(fields, f->name);
-            	auto ftype = ctxt->typeMap->getType(f);
-            	ctxt->typeMap->setType(e, ftype);
-            	list->push_back(e);
+                auto ftype = ctxt->typeMap->getType(f);
+                ctxt->typeMap->setType(e, ftype);
+                list->push_back(e);
             }
         }
         fields = list;
--- a/backends/bmv2/common/extern.cpp
+++ b/backends/bmv2/common/extern.cpp
@@ -229,11 +229,18 @@ ExternConverter::createCalculation(ConversionContext* ctxt,
             modelError("%1%: expected a struct", fields);
             return calcName;
         }
-        for (auto f : type->to<IR::Type_StructLike>()->fields) {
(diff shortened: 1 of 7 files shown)


mbudiu-vmw:
The failures are due to the MacOs build; hopefully rebasing this on #2780 will fix these too.
The fix consists in handling specially StructExpressions by converting them to ListExpressions in the bmv2 backend.

backends/bmv2/common/extern.cpp
```diff
@@ -229,11 +229,18 @@ ExternConverter::createCalculation(ConversionContext* ctxt,
...
-            auto ftype = ctxt->typeMap->getType(f);
-            ctxt->typeMap->setType(e, ftype);
-            list->push_back(e);
+        if (auto se = fields->to<IR::StructExpression>()) {
+            for (auto f : se->components) {
+                auto e = f->expression;
+                list->push_back(e);
+            }
+        } else {
+            for (auto f : type->to<IR::Type_StructLike>()->fields) {
```
fruffy, line 238: I guess the code before already makes sure this is a Type_StructLike?

fruffy (approved):

backends/bmv2/common/extern.cpp
mbudiu-vmw: Yes, there is a check on line 228 that does exactly this.

PR #2783, opened 2021-05-31
Added underscore to "#no.." regex to fix syntax error for "#novisit_children"

maheswari-s:


mbudiu-vmw:
Where is novisit_children used and documented?

mbudiu-vmw (approved):

maheswari-s:
> Where is novisit_children used and documented?

It is documented in ir/ir.def file as below:
---
  #noXXX          -> do not emit the specified implementation for the XXX method
                              e.g., #noconstructor, #nodbprint, **#novisit_children**
----

In our target specific implementation, we are creating backend IRs.
In that, we need this #novisit_children directive support.
While using this directive, we have got syntax error because of '_' in it.

PR #2785, opened 2021-06-02
Call setInvalid on all uninitialized headers.

### LINKED ISSUE #2344: Inlining functions and duplicate table calls
This is a really niche issue I stumbled across on accident but I thought I might still file it. I have not been able to reproduce this with other declarations, only with undefined headers.

When calling a table twice, `RemoveActionParameters` may not always maintain the isolation of variable declarations. 

In this case 

```p4
    @name("ingress.simple_action") action simple_action() {
        {
            bool hasReturned = false;
            bit<32> retval;
            H tmp1_0;
            if (tmp1_0.a != 32w10) {
                tmp1_0.a = tmp1_0.a + 32w10;
            }
            hasReturned = true;
            retval = tmp1_0.a;
            h.h.a = retval;
        }
    }
    @name("ingress.simple_table") table simple_table_0 {
        key = {
            h.h.b: exact @name("key") ;
        }
        actions = {
            simple_action();
            @defaultonly NoAction_0();
        }
        default_action = NoAction_0();
    }
    apply {
        simple_table_0.apply();
        simple_table_0.apply();
    }
}
```
turns into 

```p4
    bool hasReturned;
    bit<32> retval;
    H tmp1_0;
    @noWarn("unused") @name(".NoAction") action NoAction_0() {
    }
    @name("ingress.simple_action") action simple_action() {
        {
            hasReturned = false;
            if (tmp1_0.a != 32w10) {
                tmp1_0.a = tmp1_0.a + 32w10;
            }
            hasReturned = true;
            retval = tmp1_0.a;
            h.h.a = retval;
        }
    }
    @name("ingress.simple_table") table simple_table_0 {
        key = {
            h.h.b: exact @name("key") ;
        }
        actions = {
            simple_action();
            @defaultonly NoAction_0();
        }
        default_action = NoAction_0();
    }
    apply {
        simple_table_0.apply();
        simple_table_0.apply();
    }
}
```
where `tmp1_0` may be set by the previous table call and influence the execution of the second call, despite being defined inside the action body.
No stf test because the simple switch does not supported multiple table calls just yet. 

[two_tables.p4.txt](https://github.com/p4lang/p4c/files/4538605/two_tables.p4.txt)

mbudiu-vmw:
Fixes #2344 
Also changes how def-use analysis treats setInvalid. It does not really define the fields of a header, although it may clobber them.

### DIFF
--- a/testdata/p4_16_samples_outputs/gauntlet_hdr_assign_2-bmv2-midend.p4
+++ b/testdata/p4_16_samples_outputs/gauntlet_hdr_assign_2-bmv2-midend.p4
@@ -36,17 +36,18 @@ control ingress(inout Headers h, inout Meta m, inout standard_metadata_t sm) {
     @hidden action gauntlet_hdr_assign_2bmv2l43() {
         h.h.setValid();
     }
-    @hidden action gauntlet_hdr_assign_2bmv2l38() {
+    @hidden action gauntlet_hdr_assign_2bmv2l37() {
+        h.h.a = 8w1;
         h.h.setInvalid();
     }
     @hidden action gauntlet_hdr_assign_2bmv2l45() {
         h.h.b = 8w2;
     }
-    @hidden table tbl_gauntlet_hdr_assign_2bmv2l38 {
+    @hidden table tbl_gauntlet_hdr_assign_2bmv2l37 {
         actions = {
-            gauntlet_hdr_assign_2bmv2l38();
+            gauntlet_hdr_assign_2bmv2l37();
         }
-        const default_action = gauntlet_hdr_assign_2bmv2l38();
+        const default_action = gauntlet_hdr_assign_2bmv2l37();
     }
     @hidden table tbl_gauntlet_hdr_assign_2bmv2l41 {
         actions = {
@@ -67,7 +68,7 @@ control ingress(inout Headers h, inout Meta m, inout standard_metadata_t sm) {
         const default_action = gauntlet_hdr_assign_2bmv2l45();
     }
     apply {
-        tbl_gauntlet_hdr_assign_2bmv2l38.apply();
+        tbl_gauntlet_hdr_assign_2bmv2l37.apply();
         if (sm.egress_port == 9w2) {
             tbl_gauntlet_hdr_assign_2bmv2l41.apply();
         } else {
--- a/testdata/p4_16_samples_outputs/issue561-2-bmv2-frontend.p4
+++ b/testdata/p4_16_samples_outputs/issue561-2-bmv2-frontend.p4
@@ -66,8 +66,10 @@ control ingress(inout headers hdr, inout metadata meta, inout standard_metadata_
     apply {
         debug_hdr_0.apply();
         if (hdr.u.short.isValid()) {
+            hdr.u.short.data = 16w0xffff;
             hdr.u.short.setInvalid();
         } else if (hdr.u.byte.isValid()) {
+            hdr.u.byte.data = 8w0xff;
             hdr.u.byte.setInvalid();
         }
     }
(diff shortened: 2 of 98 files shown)


testdata/p4_16_samples_outputs/issue561-2-bmv2-frontend.p4
```diff
@@ -66,8 +66,10 @@ control ingress(inout headers hdr, inout metadata meta, inout standard_metadata_
     apply {
         debug_hdr_0.apply();
         if (hdr.u.short.isValid()) {
+            hdr.u.short.data = 16w0xffff;
```
fruffy, line 69: What's happening here?

testdata/p4_16_samples_outputs/gauntlet_hdr_assign_2-bmv2-midend.p4
```diff
@@ -36,17 +36,18 @@ control ingress(inout Headers h, inout Meta m, inout standard_metadata_t sm) {
     @hidden action gauntlet_hdr_assign_2bmv2l43() {
         h.h.setValid();
     }
-    @hidden action gauntlet_hdr_assign_2bmv2l38() {
+    @hidden action gauntlet_hdr_assign_2bmv2l37() {
+        h.h.a = 8w1;
```
fruffy, line 40: So the pass readds assignments that have no effect?

fruffy (approved):

testdata/p4_16_samples_outputs/gauntlet_hdr_assign_2-bmv2-midend.p4
mbudiu-vmw: The spec does not say that this assignment has no effect. Assigning to invalid header fields is really undefined, so having no effect is only one possible semantics, but this is a legal interpretation too. I have tweaked the meaning of setInvalid not to overwrite all fields of the header in the def-use analysis, that's why this assignment is no longer deleted.

testdata/p4_16_samples_outputs/issue561-2-bmv2-frontend.p4
mbudiu-vmw: This is an instance of the same change as below: the setInvalid call that follows this statement is no longer considered as overwriting all fields, so this assignment is no longer considered dead.

testdata/p4_16_samples_outputs/gauntlet_hdr_assign_2-bmv2-midend.p4
mbudiu-vmw: https://p4lang.github.io/p4-spec/docs/P4-16-v1.2.1.html#sec-uninitialized-values-and-writing-invalid-headers

jfingerh: I am confused.  What sentence of the spec do you believe says that assigning to a field in a currently invalid header is "really undefined"?  I thought it was defined as having no effect upon any defined fields anywhere.

jfingerh: From the section of the spec that you link above, in particular: "If any of these kinds of writes are performed:

+ a write to a field in a currently invalid header, either a regular header or an element of a header stack with an index that is in range, and that header is not part of a header_union
[ two bullet points irrelevant to my point omitted ]

then that write must not change any state that is currently defined in the system, neither in header fields nor anywhere else. In particular, if an invalid header is involved in the write, it must remain invalid."

mbudiu-vmw: I think that the interpretation is that if you invalidate a header, write a field, and the make it valid again you *may* read back the value you wrote (or you may not). This is the semantics currently assumed by this analysis pass - it does not treat writes to invalid headers as "dead".

jfingerh: Agreed. you might get back the value you wrote, but that is covered by the fact that when reading from a header that was recently made valid, which has not been assigned a value since the header became valid, it might return any bit pattern for the type of that field.  Such a bit pattern might happen to be one that was written to that field at some point in time.

PR #2786, opened 2021-06-02
Predictable parser : Adding extract_greedy, extract_atomic, lookahead_greedy, lookahead_atomic

AashleshaPatil-BFN:
This is the implementation of the few things mentioned in https://github.com/p4lang/p4-spec/issues/880

### DIFF
--- a/frontends/p4/coreLibrary.h
+++ b/frontends/p4/coreLibrary.h
@@ -68,10 +68,16 @@ class PacketIn : public Model::Extern_Model {
  public:
     PacketIn() :
             Extern_Model("packet_in"),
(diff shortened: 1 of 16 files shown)


hanw:
Can you explain the context that motivated this change? Changes to core.p4 most likely involves changing the P4-16 spec.

hanw:
> Can you explain the context that motivated this change? Changes to core.p4 most likely involves changing the P4-16 spec.

Nvm, saw the link above.

mbudiu-vmw:
This cannot be sufficient, there are several passes that manipulate extract, lookahead, emit. You would need to modify all these passes to handle these functions as well.

PR #2787, opened 2021-06-04
Add support for mask(&&&) and range (.. ) operations in dpdk backend and fixed few typos

usha1830:
This commit adds support for (Mask &&&) and Range (..) operations in p4c-dpdk.

The range operation is converted to Mask by midend and hence only &&& needs to be supported.

Mask operation is implemented as per the P4 Language spec **a &&& b = { c of type bit<W> where a & b = c & b }**
transition select (c) {
a &&& b : state1 is converted to 
if(( a& b) == c &b), goto state1

-Dpdk assembly does not have a syntax for adding local variable. Hence, a temporary variable to hold the result of (c&b)  is added to the metadata. One new pass is added to insert this temporary variable to parserLocals. Later another pass collects all local variables to metadata struct. This temporary variable is later used by the assembly generation pass.

### DIFF
--- a/backends/dpdk/dpdkProgram.cpp
+++ b/backends/dpdk/dpdkProgram.cpp
@@ -36,6 +36,7 @@ const IR::DpdkAsmStatement *ConvertToDpdkProgram::createListStatement(
 }
 
 const IR::DpdkAsmProgram *ConvertToDpdkProgram::create(IR::P4Program *prog) {
+    const IR::PathExpression *parserMask = nullptr;
     IR::IndexedVector<IR::DpdkHeaderType> headerType;
     for (auto kv : structure.header_types) {
         LOG3("add header type " << kv.second);
@@ -52,6 +53,14 @@ const IR::DpdkAsmProgram *ConvertToDpdkProgram::create(IR::P4Program *prog) {
                     {new IR::Annotation(IR::ID("__metadata__"), {})});
                 for (auto anno : s->annotations->annotations)
                     annotations->add(anno);
+
+                /* Fetch temporary variable created earlier for holding b & c for &&& operation */
+                for (auto field : s->fields) {
+                   if ((field->name.name).find("tmpMask_b_AND_c")) {
+                       parserMask = new IR::PathExpression(IR::ID("m."+field->name.name));
+                   }
+                }
+
                 auto st = new IR::DpdkStructType(s->srcInfo, s->name,
                                                  annotations, s->fields);
                 structType.push_back(st);
@@ -88,9 +97,9 @@ const IR::DpdkAsmProgram *ConvertToDpdkProgram::create(IR::P4Program *prog) {
 
     IR::IndexedVector<IR::DpdkAsmStatement> statements;
     auto ingress_parser_converter =
-        new ConvertToDpdkParser(refmap, typemap, collector, csum_map);
+        new ConvertToDpdkParser(refmap, typemap, collector, csum_map, parserMask);
     auto egress_parser_converter =
-        new ConvertToDpdkParser(refmap, typemap, collector, csum_map);
+        new ConvertToDpdkParser(refmap, typemap, collector, csum_map, parserMask);
     for (auto kv : structure.parsers) {
         if (kv.first == "ingress")
             kv.second->apply(*ingress_parser_converter);
@@ -210,8 +219,29 @@ bool ConvertToDpdkParser::preorder(const IR::P4Parser *p) {
                 switch_var = e->select->components[0];
                 for (auto v : e->selectCases) {
                     if (!v->keyset->is<IR::DefaultExpression>()) {
-                        add_instr(new IR::DpdkJmpEqualStatement(
-                            append_parser_name(p, v->state->path->name), switch_var, v->keyset));
+                        if (v->keyset->is<IR::Mask>()) {
+                            auto maskexpr = v->keyset->to<IR::Mask>();
+                            auto left = maskexpr->left;
+                            auto right = maskexpr->right;
+                            /* Dpdk architecture requires that both operands of &&& be constants */
+                            if (!left->is<IR::Constant>() || !left->is<IR::Constant>())
+                                ::error(ErrorType::ERR_UNSUPPORTED,
+                                        "Non constant values are not supported in Mask operation");
+                            if (tmpMask) {
+                                unsigned value = right->to<IR::Constant>()->asUnsigned() &
+                                                 left->to<IR::Constant>()->asUnsigned();
+                                add_instr(new IR::DpdkMovStatement(tmpMask, switch_var));
+                                add_instr(new IR::DpdkAndStatement(tmpMask, tmpMask, right));
+                                add_instr(new IR::DpdkJmpEqualStatement(
+                                          append_parser_name(p, v->state->path->name),
+                                                            tmpMask, new IR::Constant(value)));
+                            } else {
+                               BUG("Temporary variable to hold input mask is not found.");
+                            }
+                        } else {
+                            add_instr(new IR::DpdkJmpEqualStatement(
+                              append_parser_name(p, v->state->path->name), switch_var, v->keyset));
+                        }
                     } else {
                         auto i = new IR::DpdkJmpLabelStatement(
                                 append_parser_name(p, v->state->path->name));
--- a/backends/dpdk/dpdkProgram.cpp
+++ b/backends/dpdk/dpdkProgram.cpp
@@ -38,7 +38,7 @@ const IR::DpdkAsmStatement *ConvertToDpdkProgram::createListStatement(
 const IR::DpdkAsmProgram *ConvertToDpdkProgram::create(IR::P4Program *prog) {
     IR::IndexedVector<IR::DpdkHeaderType> headerType;
     for (auto kv : structure.header_types) {
-        // std::cout << kv.second << std::endl;
+        LOG3("add header type " << kv.second);
         auto h = kv.second;
         auto ht = new IR::DpdkHeaderType(h->srcInfo, h->name, h->annotations,
                                          h->fields);
--- a/backends/dpdk/dpdkProgram.cpp
+++ b/backends/dpdk/dpdkProgram.cpp
@@ -146,14 +146,14 @@ bool ConvertToDpdkParser::preorder(const IR::P4Parser *p) {
     for (auto l : p->parserLocals) {
         collector->push_variable(new IR::DpdkDeclaration(l));
     }
-    std::unordered_map<cstring, int> degree_map;
-    std::unordered_map<cstring, IR::ParserState> state_map;
-    std::vector<IR::ParserState> stack;
+    ordered_map<cstring, int> degree_map;
+    ordered_map<cstring, const IR::ParserState*> state_map;
+    std::vector<const IR::ParserState*> stack;
     for (auto state : p->states) {
         if (state->name == "start")
-            stack.push_back(*state);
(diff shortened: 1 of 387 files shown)


backends/dpdk/dpdkProgram.cpp
```diff
@@ -52,6 +53,14 @@ const IR::DpdkAsmProgram *ConvertToDpdkProgram::create(IR::P4Program *prog) {
                     {new IR::Annotation(IR::ID("__metadata__"), {})});
                 for (auto anno : s->annotations->annotations)
                     annotations->add(anno);
+
+                /* Fetch temporary variable created earlier for holding b & c for &&& operation */
+                for (auto field : s->fields) {
+                   if ((field->name.name).find("tmpMask_b_AND_c")) {
+                       parserMask = new IR::PathExpression(IR::ID("m."+field->name.name));
```
mbudiu-vmw, line 60: This isn't really safe, if the user happens to have a variable with this name you will use the wrong one.
Moreover, this is also wrong if you have multiple such variables with different types. Consider the case where you have two states with select statements with different types. Then you will have two such variables with different types.
Perhaps the simplest solution is to have a pass that converts a select into a series of cascading if statements.
The current IR cannot represent an if that transitions, so you would need to define a new IR node for that, but that's not particularly difficult.

backends/dpdk/dpdkProgram.cpp
```diff
@@ -210,8 +219,29 @@ bool ConvertToDpdkParser::preorder(const IR::P4Parser *p) {
...
                 for (auto v : e->selectCases) {
                     if (!v->keyset->is<IR::DefaultExpression>()) {
-                        add_instr(new IR::DpdkJmpEqualStatement(
-                            append_parser_name(p, v->state->path->name), switch_var, v->keyset));
+                        if (v->keyset->is<IR::Mask>()) {
+                            auto maskexpr = v->keyset->to<IR::Mask>();
+                            auto left = maskexpr->left;
+                            auto right = maskexpr->right;
+                            /* Dpdk architecture requires that both operands of &&& be constants */
+                            if (!left->is<IR::Constant>() || !left->is<IR::Constant>())
```
mbudiu-vmw, line 227: you should probably check this in the earlier pass.

mbudiu-vmw (changes requested):

backends/dpdk/dpdkProgram.cpp
usha1830: Yeah, I missed the point that searching for the variable using substring would be erroneous in case of multiple such variables and user defined variables.

And regarding writing a different pass for this: the select statement in dpdk is always converted into conditional jumps. I just made use of the existing framework to generate these conditional jumps to support Mask operator. Writing a new pass would be cleaner but would also require overall design change. 
@hanw any comments/concerns for this?

The issue with accessing variable name correctly could be fixed in current framework by defining a new map with "<blockname_statename> => path expression" as key=>value pair for these temp variables. And then, instead of using substring matching, access the variables using this map.

>> The current IR cannot represent an if that transitions, so you would need to define a new IR node for that, but that's not particularly difficult.

I don't quite understand this. Even in general case (without mask operation), the select transition statement is converted to Ifs, so not sure what difference it would make in IR for handling this case.

mbudiu-vmw: 1) I don't think your last proposal would work either. Consider the following:
`select (a, b) {  8w4 &&& 8w2, 16w4 &&& 16w 2: a }`
There are two masks in the same select.
2) You already have a translation for if, so if there is no select there is one fewer case to consider
3) Why do you even need to insert variable declarations in the P4 code? These variables are not really used anywhere. Can't you actually just insert dpdk variable declarations in your IR while you scan the select expression?

hanw: Some existing compiler infra may help you on this:
-  https://github.com/p4lang/p4c/blob/main/ir/pattern.h help you to write code to match on a particular expression pattern, instead of hard-code string match on names.
-  findContext() to filter IR usage in a paritcular context, e.g. a mask expression under IR::TableKey.

usha1830: >> There are two masks in the same select.

Yes, I think the "key" used in my last proposal would have to be modified further (using the select expression probably) to uniquely identify each mask.

>> Why do you even need to insert variable declarations in the P4 code? These variables are not really used anywhere. 

The general assembly format for dpdk is:
\<structure declarations for header and metadata\>
\<instance of header and metadata\>
\<extern declarations\>
\<actions\>
\<tables\>
\<apply block\>

There is single apply block which consists of all the statements from parser and control blocks. All local variables from parser and control blocks are encapsulated within a single metadata struct and accessed using "m.<var_name>" within the apply block. So we need to insert the temporary variable declaration into the metadata struct.

>> Can't you actually just insert dpdk variable declarations in your IR while you scan the select expression?

Thanks for pointing that out. I had a deeper look in the pass where we scan the transition select expression. I do believe that I can modify the metadata struct in the same pass to insert the new temp variables.

So the question is: Should we update that pass for Mask handling or we absolutely require to pull this select expression handling out of current framework and write a new pass as you suggested earlier?

mbudiu-vmw: It looks to me like you should be able to handle select masks in only one pass, without the preparation pass that inserts variables. Just insert the metadata dpdk information right when you need it, in the same pass that converts masks.

usha1830: I have modified the implementation to add the temp variables to metadata struct in the same pass where the select expressions are being scanned.
Some of the code is reordered in the  ConvertToDpdkProgram::create() function. This is to enable the new temp variables to be added to metadata before converting the metadata struct to DPDK IR.

backends/dpdk/dpdkProgram.cpp
```diff
@@ -36,61 +36,22 @@ const IR::DpdkAsmStatement *ConvertToDpdkProgram::createListStatement(
...
-                structType.push_back(st);
-            } else if (s->name.name == info->header_type) {
-                auto *annotations = new IR::Annotations(
-                    {new IR::Annotation(IR::ID("__packet_data__"), {})});
-                for (auto anno : s->annotations->annotations)
-                    annotations->add(anno);
-                auto st = new IR::DpdkStructType(s->srcInfo, s->name,
-                                                 s->annotations, s->fields);
-                structType.push_back(st);
+                metadataStruct = s->clone();
```
mbudiu-vmw, line 44: No break?

mbudiu-vmw: This seems a bit brittle; in other backends we use the type of the suitable parameters (e.g., third parameter of ingress) to find the metadata type. But perhaps this will work, I don't remember how local_metadata_type is computed.

backends/dpdk/dpdkProgram.cpp
```diff
@@ -134,6 +95,66 @@ const IR::DpdkAsmProgram *ConvertToDpdkProgram::create(IR::P4Program *prog) {
...
+        headerType.push_back(ht);
+    }
+    IR::IndexedVector<IR::DpdkStructType> structType;
+
+    /* Insert the metadata structure updated with tmp variables created during parser conversion */
+    auto new_objs = new IR::Vector<IR::Node>;
+    for (auto obj : prog->objects) {
+        if (auto s = obj->to<IR::Type_Struct>()) {
+            if (s->name.name == info->local_metadata_type) {
+                auto *annotations = new IR::Annotations(
```
mbudiu-vmw, line 113: why don't you factor this code into a separate function that returns a IndexedVector?

backends/dpdk/dpdkProgram.cpp
```diff
@@ -149,6 +170,10 @@ cstring ConvertToDpdkParser::append_parser_name(const IR::P4Parser *p, cstring l
     return p->name + "_" + label;
 }
 
+void ConvertToDpdkParser::add_metadata_field(IR::Declaration_Variable *d) {
+    metadataStruct->fields.push_back(new IR::StructField(IR::ID(d->name.name), d->type));
```
mbudiu-vmw, line 174: what happens if a local variable has a name identical to one of the existing fields?

mbudiu-vmw (changes requested):

backends/dpdk/dpdkProgram.cpp
usha1830: Yes. Break is needed here. I will update the code.

I used clone here as the metadata struct returned here as program object is const and I need to insert new fields to it. Any other way to do that?

local_metadata_type is collected from the Ingress Parser block parameters in an earlier pass "CollectMetadataHeaderInfo". The information collected in this pass is extensively used in many other passes.

 ```
auto local_metadata = parser->getApplyParameters()->getParameter(2);
 local_metadata_type =
                local_metadata->type->to<IR::Type_Name>()->path->name;
```

File: backends/dpdk/dpdkProgram.cpp (hunk 4 above)
Comment (usha1830) [reply backends/dpdk/dpdkProgram.cpp:113]: I will do that.

File: backends/dpdk/dpdkProgram.cpp (hunk 5 above)
Comment (usha1830) [reply backends/dpdk/dpdkProgram.cpp:174]: I assumed getting a unique name using refMap would take care of any name conflicts. Please correct me if I am wrong and please also suggest how we should handle it in this particular case(creating a new temporary variable and inserting it into an existing Structure).

Also, as per my understanding of the dpdk backend code, it is not possible to have an identical name in the metadata struct field from elsewhere. The metadata struct fields from user program are prefixed with "local_metadata_" and the locals within different parser/control blocks are prefixed with the block name. Ex: a variable named "tmpMask" in "Ingress" block is inserted in metadata struct as "Ingress_tmpMask_".

I am thinking if these new temporary variables for Mask operations should also be prefixed with something similar?

Comment (mbudiu-vmw) [reply backends/dpdk/dpdkProgram.cpp:174]: Looking at the generated code can you reverse engineer an input program that would produce the wrong results because it generates the same name twice? If you can, you should add that program as a test case and fix the code to make sure it uses everywhere unique identifiers. In this particular line you are not generating a fresh name, you are using the variable name directly.

Comment (usha1830) [reply backends/dpdk/dpdkProgram.cpp:174]: A fresh name is created right before calling this function and this function is called only from one place. 


                         auto tmpDecl = new IR::Declaration_Variable(IR::ID(
                                                 refmap->newName("tmpMask")), switch_var->type);

                            .......
                         add_metadata_field(tmpDecl);

I tried a few more modifications to my test program to use the variable with same name in different parser and control blocks. There seems to be no conflict.

But I agree, to use add_metadata_field to be called from other places, we will have to make sure that the caller creates a fresh name. To fix this, I will modify the function to create a new field name, insert it to metadata and return the same declaration to the caller. The caller can then use the returned declaration to create a pathexpression.

File: backends/dpdk/dpdkProgram.cpp [hunk 6]
```diff
@@ -35,13 +35,58 @@ const IR::DpdkAsmStatement *ConvertToDpdkProgram::createListStatement(
     return new IR::DpdkListStatement(name, *stmts);
 }
 
+/* Insert the metadata structure updated with tmp variables created during parser conversion
+   Add annotations to metadata and header structures and add all the structures to DPDK
+   structtype.
+*/
+IR::IndexedVector<IR::DpdkStructType> ConvertToDpdkProgram::UpdateHeaderMetadata
```
mbudiu-vmw, line 42: Does our linter like this? I expect it wants the open parens on the previous row.

mbudiu-vmw (approved):

mbudiu-vmw:
Is this ready to be merged?

PR #2789, opened 2021-06-04
keep unused method declaration

hanw:
Unused extern function definition is removed by the end of frontend. As a result, if a midend pass generates code that invokes the extern function, we run into an 'declaration not found' error.

The culprit is the the RemoveAllUnusedDeclarations pass which removes unused method declaration including the unused declaration in an architecture definition. 

For instance, 
```
// in arch.p4
extern void funnel_shift_right<T>(inout T dst, in T src1, in T src2, int shift_amount);

// p4 action generated by a midend pass
action foo() {
   funnel_shift_right(dst, src1, src2, 8);  // triggers a 'declaration not found' error
}
```

This declaration is removed at the end of the frontend.

The extern function declaration is parsed as: (note it's IR::Method, not IR::Type_Method).
```
  [1288] Method name=funnel_shift_right declid=205 isAbstract=0
    type: [1287] Type_Method name=funnel_shift_right
      typeParameters: [1270] TypeParameters
        [1268] Type_Var name=T declid=52
      returnType: [1] Type_Void
      parameters: [1285] ParameterList
        [1274] Parameter name=dst declid=201 direction=inout
          annotations: [5] Annotations
          type: [1273] Type_Name
            path: [1272] Path name=T absolute=0
        [1278] Parameter name=src1 declid=202 direction=in
          annotations: [5] Annotations
          type: [1277] Type_Name
            path: [1276] Path name=T absolute=0
        [1281] Parameter name=src2 declid=203 direction=in
          annotations: [5] Annotations
          type: [1280] Type_Name
            path: [1279] Path name=T absolute=0
        [1283] Parameter name=shift_amount declid=204 direction=
          annotations: [5] Annotations
          type: [1282] Type_InfInt declid=108
    annotations: [5] Annotations
```

### DIFF
--- a/frontends/p4/unusedDeclarations.cpp
+++ b/frontends/p4/unusedDeclarations.cpp
@@ -154,12 +154,15 @@ const IR::Node* RemoveUnusedDeclarations::preorder(IR::ParserState* state) {
 // Try to guess whether a file is a "system" file
 bool RemoveUnusedDeclarations::isSystemFile(cstring file) {
     if (file.startsWith(p4includePath)) return true;
+    // if the backend is invoked directly with '-I p4include'
+    if (file.startsWith("p4include")) return true;
     return false;
 }
 
 cstring RemoveUnusedDeclarations::ifSystemFile(const IR::Node* node) {
     if (!node->srcInfo.isValid()) return nullptr;
     auto sourceFile = node->srcInfo.getSourceFile();
+    LOG1("source file " << sourceFile << " " << p4includePath);
     if (isSystemFile(sourceFile))
         return sourceFile;
     return nullptr;
--- a/frontends/p4/unusedDeclarations.cpp
+++ b/frontends/p4/unusedDeclarations.cpp
@@ -16,6 +16,7 @@ limitations under the License.
 
 #include "unusedDeclarations.h"
 #include "sideEffects.h"
+#include "frontends/common/parser_options.h"
 
 namespace P4 {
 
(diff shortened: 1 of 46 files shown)


mbudiu-vmw:
This is actually related to https://github.com/p4lang/p4-spec/issues/929
The code will not remove any declarations that start with a double underscore (see `process(const IR::IDeclaration*)`).
So perhaps if you can rename your method to start with a double underscore you don't need to do anything.

hanw:
That means I should to add `__` to all extern function in tofino.p4? This further implies all system-defined functions should start with `__` in P4? Otherwise, if midend synthesizes a function call, the generated code will fail to resolve references.

```
// -----------------------------------------------------------------------------
// EXTERN FUNCTIONS
// -----------------------------------------------------------------------------
extern T max<T>(in T t1, in T t2);
extern T min<T>(in T t1, in T t2);
extern void funnel_shift_right<T>(inout T dst, in T src1, in T src2, int shift_amount);
extern void invalidate<T>(in T field);
```

hanw:
Should extern function be parsed as IR::Type_Method, instead of IR::Method?

mbudiu-vmw:
> Should extern function be parsed as IR::Type_Method, instead of IR::Method?

No, Type_Method is the IR node used to represent the type of any method. Technically the type should not include the name (but it does for better error messages).

mbudiu-vmw:
Indeed, if the compiler synthesizes calls to some functions that were not in the original program you will have to avoid deleting their declarations earlier. This is true for the `valid` function, and you can see that the code handles that specially too. But that function is part of core.p4.

That's why the spec issue is about reserved identifiers, although "reserved" is not exactly the same as "uses of it may be synthesized by the compiler".

We can solve this problem in several ways. One is the way you propose here, by not deleting method declarations (extern or not). The other would be to do like the toP4 pass does: detect whether the declaration comes from a "system" header file (see `toP4::isSystemFile`), and in that case not delete it. The third would be to use double underscores for all methods that can be invoked by the compiler internally. Choose one.

hanw:
I changed PR to use the ToP4 approach by checking whether a method came from 'system' file. The helper function is slightly different from the one used in ToP4.

mbudiu-vmw:
I expect this change will modify some reference outputs.

mbudiu-vmw (approved):

PR #2792, opened 2021-06-07
Checksum support for PSA model added

dukak-syrmia:
*Added support for padding in scalars*
-Length of scalars header need to be multiple of 8. This fixes the problem when the length of scalars header is not multiple of 8. It extends it with padding which size is 8-(scalars_header_length % 8).

*Added support for booleans in scalars header*
-Boolean variable need to be defined as a field in scalars header. This fixes registration of a boolean variable in scalars header.

*Added support for checksum in PSA model*
-Parser and deparser recognize PSA Checksum instances in their locals and convert them into JSON.
-Parser and deparser recognize PSA checksum methods and convert them into JSON.


dukak-syrmia:
build-linux tests *test-p4c / build-linux (ON, ON) * and *test-p4c / build-linux (ON, OFF)* failed because the changes in behavioral-model haven't been merged yet.
 p4lang/behavioral-model#1005

PR #2796, opened 2021-06-11
Typecheck properly casts of header expressions

### LINKED ISSUE #2795: Direct casts of headers are rejected by compiler
```
#include <core.p4>

header H {
    bit<32> a;
    bit<32> b;
}

control c(packet_out p) {
    apply {
        p.emit((H){0, 1});
    }
}

control ctr(packet_out p);
package top(ctr _c);

top(c()) main;
```

mbudiu-vmw:
Fixes #2795


mbudiu-vmw:
It is however not clear that this is enough to support direct emits `packet.emit((H){0, 1});`, where `H` is a header type. The compiler currently expands the argument of emit into a series of explicit header initialization steps.

mbudiu-vmw:
@vgurevich

mbudiu-vmw:
This change effectively allows casts for list expressions to structs by recursively casting all fields to the proper type.

fruffy (approved):

fruffy:
Does this cover [this comment](https://github.com/p4lang/p4c/issues/2795#issuecomment-860150167), too?

mbudiu-vmw:
I am making more changes to support `p.emit<H>({0, 1});`
This is more involved. The compiler does not cast a list expression as a function argument to a header type.
Unfortunately this change will be a bit more involved, so I will ask for a second code review.
It will essentially iterate over the arguments of a function call and cast each of them if necessary.

mbudiu-vmw:
I had to mark a few tests as xfail to make this work:
- 2 tests due to issue https://github.com/p4lang/p4c/issues/1669, see #2811
- 1 test due to #2808 (it does not seem to the be only one in that category)

frontends/p4/checkCoreMethods.cpp
```diff
@@ -0,0 +1,129 @@
+/*
```
mbudiu-vmw, line 1: This code used to be in the typeChecker, but was moved to this separate file.
Otherwise it is unchanged.

frontends/p4/typeChecking/typeChecker.cpp
```diff
@@ -3155,98 +3165,6 @@ bool TypeInference::onlyBitsOrBitStructs(const IR::Type* type) const {
     return false;
 }
 
-void TypeInference::checkEmitType(const IR::Expression* emit, const IR::Type* type) const {
```
mbudiu-vmw, line 3158: this code has been moved into the checkCoreMethods pass.

frontends/p4/typeChecking/typeUnification.cpp
```diff
@@ -44,28 +44,11 @@ bool TypeUnification::unifyCall(const EqualityConstraint* constraint) {
     }
     constraints->addUnifiableTypeVariable(src->returnType);  // always a type variable
 
+    // Adding type parameters last will ensure that they are matched first
+    // since the algorithms pops constraints from the back of the list
     for (auto tv : dest->typeParameters->parameters)
         constraints->addUnifiableTypeVariable(tv);
 
-    if (src->typeArguments->size() != 0) {
```
mbudiu-vmw, line 50: this code has just been moved a few lines lower in this function

mbudiu-vmw:
To make this work I had to modify the typechecker to convert ListExpression arguments to StructExpression arguments where the function expects a struct. This has cascaded through the compiler finding several other places where the passes were not handling StructExpression but only ListExpression. So most of the code in this PR really adds handling of StructExpression in various passes. A few other bugs were hit when this conversion was performed.

 @fruffy : I hope you can review this again; the code is more involved, but the changes are fairly small except moving some code around, which I have tagged in the PR as comments.

backends/bmv2/simple_switch/simpleSwitch.cpp
```diff
@@ -803,6 +803,36 @@ void ExternConverter_action_selector::convertExternInstance(
     ctxt->action_profiles->append(action_profile);
 }
 
+namespace {
```
fruffy, line 806: LGMT. Only a question here. Why is this extra namespace necessary?

fruffy (approved):

backends/bmv2/simple_switch/simpleSwitch.cpp
mbudiu-vmw: This prevents the linker from creating this as a global symbol. We've had bugs in the past where local passes (no header files) clashed between files. These cause extremely difficult to diagnose bugs (runtime crashes). The C++ linker does not complain, and is non-deterministic.

frontends/p4/typeChecking/typeChecker.cpp
```diff
@@ -2527,12 +2527,22 @@ const IR::Node* TypeInference::postorder(IR::Cast* expression) {
                 return expression;
             }
         } else if (auto le = expression->expr->to<IR::ListExpression>()) {
-            if (st->fields.size() == 0 && le->size() == 0) {
-                // Empty structs
+            if (st->fields.size() == le->size()) {
```
MichalKekely, line 2530: @mbudiu-vmw This change essentially converts most of the ListExpressions to StructExpressions. This means that midend/backend passes that were expecting a ListExpression might need to also be updated to handle StructExpression. Is there a possibility for the work with ListExpression and StructExpression to somehow be unified? Currently those two classes are not derived from each other, which means they (to some extend) need to be handled separately. The idea is that both of those classes represent pretty similar (if not the same) thing - StructExpression just adds some more structure to the data (NamedExpressions and structType). So what are the possibilities here?

mbudiu-vmw: Yes, all the midend passes that are part of the open-source repository have been converted to handle StructExpressions.
I don't really see how to unify the two classes. They are relatively simple.
In general, the StructExpression has strictly more information than the ListExpression.
Note that there are some passes, like the tuple elimination pass, which require this conversion to be done again. Tuple elimination converts tuples to structs, and thus tuple initializers, which are ListExpressions, need to change to struct initializers, which are StructExpression.

PR #2799, opened 2021-06-13
Add some more required dependencies for the validation tests.

fruffy:
I recently switched to a C++-based interpreter for validation. The functionality should stay the same, but it is easier to maintain and much faster. However, there are some additional dependencies required when building with Ubuntu 16.04. I hope I can sort them out over time. For now they have to  be added to get the tests to run again.

### DIFF
--- a/tools/ci-build.sh
+++ b/tools/ci-build.sh
@@ -82,29 +82,37 @@ rm -rf /tmp/pip
 # ! ------  BEGIN VALIDATION -----------------------------------------------
 
 function build_gauntlet() {
-  # for add-apt-repository
+  # For add-apt-repository.
   apt-get install -y software-properties-common
-  # Also pull Gauntlet for translation validation
+  # Also pull Gauntlet for translation validation.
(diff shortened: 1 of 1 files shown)


jfingerh:
Are p4lang/p4c CI tests using Ubuntu 16.04 still?  Given that Ubuntu 16,04 is past "free support" and into "extended software maintenance" (https://ubuntu.com/about/release-cycle), it seems best to update that some time soon-ish.  Switching to Ubuntu 18.04 would give another 2 years of free support there on the base OS, and perhaps be less hassle than updating all the way to Ubuntu 20.04.

mbudiu-vmw (approved):

fruffy:
Yes, I also vote for deprecating Ubuntu 16.04. In my experience p4c and bmv2 compile and work well with Ubuntu 18.04 and Ubuntu 20.04. However, switching is fairly involved. We will need to start at [third party](https://github.com/p4lang/third-party/blob/master/Dockerfile#L27) here.

mbudiu-vmw:
@fruffy : if the "build status" badge is no longer being used, can you please remove it?

PR #2803, opened 2021-06-15
Optimize subparser inlining in specific case

mvido:
The goal of this change is to save up the resources needed for parser implementation.

Before this change, for each subparser invocation all subparser states are inlined which may mean that some states are duplicated unnecessarily.

In a specific case when there is a same sequence of parser statements and a same transition statement after multiple invocations of the same subparser instance, it is possible to inline the states only once and use them for all such invocations.

In this PR, GeneralInliner is changed to inline the subparser states only for the first such invocation and for all the other invocations there are used the same states which were already inlined for that first subparser invocation.

Example testing P4 program is added too.

If newly generated states created during inlining have the original name according to their original subparser states, it may cause collisions in Backend algorithms after this change (f.e. in loops unrolling).
Therefore the original name from the states generated during inlining is removed.


frontends/p4/inlining.h
```diff
@@ -100,10 +100,52 @@ struct PerInstanceSubstitutions {
...
+
         /// For each instance (key) the container that is intantiated.
         std::map<const IR::Declaration_Instance*, const IR::IContainer*> declToCallee;
         /// For each instance (key) we must apply a bunch of substitutions
         std::map<const IR::Declaration_Instance*, PerInstanceSubstitutions*> substitutions;
+        /// For each distinct invocation of the subparser identified by InlinedInvocationInfo
+        /// we store the name of the next caller parser state which is a new state replacing
+        /// the start state of the callee parser (subparser).
+        /// Transition to this state is used in case there is another subparser invocation
+        /// which has the equivalent InlinedInvocationInfo.
```
mbudiu-vmw, line 146: What is "equivalent"? Can you define it precisely?

testdata/p4_16_samples/parser_subparser_v1model.p4
```diff
@@ -0,0 +1,190 @@
...
+            5: p_5;
+            default: accept;
+        }
+    }
+
+    state p_0 {
+        hdr2.h1.f = 100;
+        hdr2.h1.setValid();
+
+        p.apply(packet, hdr, hdr2);
```
mbudiu-vmw, line 88: in your example you always apply the sub-parser to the same arguments, which are toplevel variables in the caller.
Can you make a few simpler tests?
And some which use locals with the same name or different arguments?

frontends/p4/inlining.h
```diff
@@ -100,10 +100,52 @@ struct PerInstanceSubstitutions {
...
+                    e->dbprint(oss);
+                }
+                return std::hash<const IR::Declaration_Instance *>{}(std::get<0>(k)) ^
+                        std::hash<std::string>{}(oss.str());
+            }
+        };
+
+        /// Binary equality predicate for InlinedInvocationInfo used as a key for unordered_map
+        struct key_equal {
+            bool operator() (const InlinedInvocationInfo &v0,
```
mbudiu-vmw, line 130: I can see several problems with this equality test:
- IR pointers are often cloned during traversal; it is not clear to me that you will be comparing the same version of an IR node
- I am not sure that equiv is the right test for expressions; two expressions can be syntactically equivalent but they can evaluate to different results *if they are invoked in different contexts*. For example, an expression that is just a variable 'a' will evaluate to a different result depending on where it is invoked.

mbudiu-vmw (changes requested):

frontends/p4/inlining.cpp
```diff
@@ -733,7 +733,7 @@ class RenameStates : public Transform {
     const IR::Node* preorder(IR::Path* path) override {
         // This is certainly a state name, by the way we organized the visitors
         cstring newName = ::get(stateRenameMap, path->name);
-        path->name = IR::ID(path->name.srcInfo, newName, path->name.originalName);
+        path->name = IR::ID(path->name.srcInfo, newName, nullptr);
```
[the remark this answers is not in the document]
mvido: I see, I already took this change back in later version and the issue it causes for backend will be handled in the respective backend.

frontends/p4/inlining.cpp
```diff
@@ -776,14 +776,14 @@ const IR::Node* GeneralInliner::preorder(IR::ParserState* state) {
     auto srcInfo = state->srcInfo;
     auto annotations = state->annotations;
     IR::ID name = state->name;
-    for (auto e : state->components) {
-        if (!e->is<IR::MethodCallStatement>()) {
-            current.push_back(e);
+    for (auto e = state->components.begin(); e != state->components.end(); ++e) {
```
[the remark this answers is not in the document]
mvido: The point is to be able to iterate over the components after the current one and store them in a vector on the line 832:
`remainingComponents->insert(remainingComponents->end(), e + 1, state->components.end());`

If you have a better idea how to achieve that I'm all ears.

testdata/p4_16_samples/parser_subparser_v1model.p4
mvido: Ok, I'll work on more test cases.

frontends/p4/inlining.h
mvido: * If other maps in the same class (`declToCallee`, `substitutions`, `callToInstance`) can use pointer (`const IR::Declaration_Instance*`) I thought that I could rely on that it is not cloned in this pass and it should be all right to use it here as well. Or am I wrong?

* I think I see where you are heading, but I'm still not sure if that is an issue. If you have a variable 'a' it is not evaluated in compile time, but in run time, isn't it? So during compilation it is still just a variable 'a' and it may have different values during run time.
So f.e. these two parsers should be equivalent, shouldn't they?
```
parser Subparser(packet_in packet) {...}

parser ParserImpl(packet_in packet) {
  Subparser() p;
  bit<8> a;
  bit<8> b;

...

  state s1 {
    a = 1;
    p.apply(packet);
    b = a;
    transition accept;
  }

  state s2 {
    a = 2;
    p.apply(packet);
    b = a;
    transition accept;
  }
}
```
```
parser Subparser(packet_in packet) {...}

parser ParserImpl(packet_in packet) {
  Subparser() p;
  bit<8> a;
  bit<8> b;

...

  state s1 {
    a = 1;
    p.apply(packet);
    transition s3;
  }

  state s2 {
    a = 2;
    p.apply(packet);
    transition s3;
  }

  state s3 {
    b = a;
    transition accept;
  }
}
```
Could you please try to provide some counterexample which would be problematic?

frontends/p4/inlining.h
mvido: "Equivalent" here has following meaning:
Two subparser invocations are equivalent when following conditions are met:
* subparser instance declarations are same for both invocations
* the sequences of parser statements following the subparser invocation are same for both invocations
* transition statement is same for both invocations

I will add the information what equality means here to the comment, once we agree on the other details, as it is possible that it will change here as well.

frontends/p4/inlining.cpp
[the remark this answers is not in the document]
mbudiu-vmw: Ok, now I see.

mbudiu-vmw:
Isn't there a more general, and simpler, solution available? Which does not depend on the calling context?
Use a new global variable per subparser of type "enum", where you encode the call location. Then you inline the body only once no matter what the code looks like. On return from the subparser use a select on this enum to dispatch to the right state.

```
subparser p();
state s1: 
    p(...);
    ...

state s2:
    p(...)
    ....
```

is compiled to:

```
enum Caller { s1, s2 };
Caller c;

state s1:
    c = Caller.s1;
    // copy-in arguments
    transition p_start;
    
state s1bottom:
    ...
    
state s2:
    c = Caller.s2;
    // copy-in arguments
    transition p_start;

p_start:
    //
    ...

p_accept:
    // copy-out arguments
    transition select (c) {
       Caller.s1: s1bottom;
       Caller.s2: s2bottom;
    }
}
```

mbudiu-vmw:
This is essentially encoding the call stack into these static variables, taking advantage of the fact that there are no recursive parser invocations - so the stack is bounded in size.

mbudiu-vmw:
There still are some subtleties about the copy-in/copy-out, but these should better be handled by the current inliner as well. We have some outstanding bugs related to `exit`, which bypasses copy-out sometimes, but I don't think these apply to `transition reject`.

mvido:
Yes, I was thinking about this too. As far as I know it should be viable. As this solution has also support from you, I'm going to implement it this way.

mvido:
@mbudiu-vmw After further analysis, we found out that suggested general solution using local variable in a parser is not suitable as it uses a construct which may impose more requirements on the target than original program (possibility of using local variable in expression list of select expression).

It means that if originally a program can be compiled for a given target, the same program could not be compiled for the given target after this change as it would be internally transformed to a program which violates the constraints of the given target.

Because of that I continued with the original approach to optimise only a specific case. I tried to address your comments and I narrowed down the set of cases when this optimisation can be used. Please see the first comment in the PR which I edited to reflect the latest changes.

mbudiu-vmw:
Can you explain what the constraint is?
Is it the number of states?

mvido:
If a target does not support using local variable in an expression list of the select expression (meaning that only a field parsed from a packet can be used in the expression list).

mvido:
Which would mean that internally created state from your example above:
```
state p_accept {
    transition select (c) {
       Caller.s1: s1bottom;
       Caller.s2: s2bottom;
    }
}
```
could cause compilation error for a target which does not support that because of using local variable c in select expression even though the original program does not use any local variable.

mbudiu-vmw:
Perhaps you should contribute a midend pass to check for such a for constraint. But this is really independent on this contribution.

mvido:
Well, it already has to be handled in the backend for the respective targets which has such constraint, backend probably already deals with more such parser constraints and checks in more complex way. It would require some refactoring and so on to move it to midend and it would still not be clear if it was usable by various targets. So I'm not sure about the added value at the moment. It is always possible to introduce such pass to midend later if needed as it is independent of this change. So please, let's focus on this change in inlining here.

mbudiu-vmw:
I have been on vacation; I will review this soon.

mvido:
I see, no worries, there is still some ongoing work on our backend related to this change, which needs to be resolved anyway.
So even if you approved this PR, it would still be better if merging of this change waits until that work on the backend is done.
Thank you very much for reviewing this!

mbudiu-vmw:
Then just mark this as a draft until ready

mvido:
Optimization of subparser inlining transforms the parser in a way,
which may cause troubles for some backends in some special cases by
adding more constraints and requirements for the target which may
cause compilation error because some constraints of the target are
not satisfied, while compilation without the optimization may pass.

Because of that new command line option is introduced which allows
to disable the optimization.

See also help for a new option in the code and the comments in the code.

mbudiu-vmw:
If you disable parser inlining then many passes in midends and backends will not work correctly because they do not expect calls to parsers. What is a backend supposed to do if it finds a parser that calls another parser?

mvido:
It is not disabling the inlining. It is disabling only the optimization of inlining introduced in this PR, so if the optimization is disabled, the inlining should work as before (with the exception of lines 859 and 887 in inlining.cpp, where I added originalName to the new IDs, which I think could be there, but was missing before).

mvido:
@mbudiu-vmw It is ready for review, so if you had time for that I would appreciate it.

frontends/common/parser_options.cpp
```diff
@@ -223,6 +223,21 @@ ParserOptions::ParserOptions() : Util::Options(defaultMessage) {
             return true;
         },
         "[Compiler debugging] Folder where P4 programs are dumped\n");
+    registerOption(
```
mbudiu-vmw, line 226: You said that this optimization causes some problems. If the problems are related to the existing open-source passes can you please file an issue describing the problems so we can address them?

mvido:
@ChrisDodd, @jafingerhut, @jnfoster Your opinion on this PR, and especially addition of a new command line option (commit https://github.com/p4lang/p4c/pull/2803/commits/04a425ff95d2c48a960125f6d26905c6c1464fbe) would be also appreciated. Thanks.

frontends/p4/inlining.h
```diff
@@ -100,12 +100,67 @@ struct PerInstanceSubstitutions {
 struct InlineSummary : public IHasDbPrint {
     /// Various substitutions that must be applied for each instance
     struct PerCaller {  // information for each caller
+        /// Pair identifying all the invocations of the subparser which can use the same
```
mbudiu-vmw, line 103: Please indicate that these structures are only used if that particular optimization is enabled and are unused otherwise.

frontends/p4/inlining.h
mbudiu-vmw: Can you also add a comment show the effect of this additional optimization on a small example?

mbudiu-vmw (commented):
It's not very easy to read the transformed code and make sure it's equivalent to the original one.
I have two proposals that would increase our confidence:
- add some simple STF files for the many tests you wrote. Since your tests are designed for BMV2 they should be immediately executable.
- one way to check that the optimization preserves semantics is to run these tests with and without this optimization on these STF files. This way of running does not have to be necessarily committed in our test suite once you checked that the results are the same; if you commit the STF files that should be enough.
I realize that this is a fair amount of work, but it should be less work than you've done already.

frontends/common/parser_options.cpp
mvido: I have seen problems only in proprietary backend passes so far.

mvido: But I haven't been testing open-source backends. I was relying on the existing CI tests in regards to open source backends.

frontends/p4/inlining.h
mvido: I added a note in the comment.

frontends/p4/inlining.h
mvido: I checked that this is eliminated by UniqueNames and MoveDeclarations passes, which are run before Inline pass, so it should be all right. I added also a note in the comment that it relies on the fact that these passes are run before that.
I added also an additional test (subparser_inlining13.p4) which contains local variables declared in states.

I added also an example of the optimization in the comment.

mvido:
Ok, I'm working on adding STF tests.

mvido:
@mbudiu-vmw I changed the command line option from disabling the optimization to enabling the optimization instead.
So now the default behaviour is as before, and the optimization can be enabled by using command line option `--parser-inline-opt`.
It makes probably more sense, as there might exist some special cases (P4 programs) when the optimization produces worse results, and thus it makes sense to use the optimization only intentionally when we know that the resources are limited and the optimization can save up some resources.

Besides that I renamed the test cases and I added option to enable the optimization only to newly add test cases (13 tests).
And I added new labels for bmv2 test driver to run newly added test cases with optimization enabled (bmv2-parser-inline-opt-enabled) and disabled (bmv2-parser-inline-opt-disabled).

mvido:
@mbudiu-vmw I added also STF tests for all 13 test cases.

I encountered an issue with behavioral-model (https://github.com/p4lang/behavioral-model) when working on STF tests for bmv2.
When p4 program contains header field assignment as this:
```
        h2.f = h1.f;
        h2.setValid();
```
(which was used in the test cases), then `simple_switch` ignores the value assigned to `h2.f`.

In bmv2 json file it is represented by following two operations (set and primitive/add_header):
```
{
              "parameters" : [
                {
                  "type" : "field",
                  "value" : ["h2", "f"]
                },
                {
                  "type" : "field",
                  "value" : ["h1", "f"]
                }
              ],
              "op" : "set"
            },
            {
              "parameters" : [
                {
                  "parameters" : [
                    {
                      "type" : "header",
                      "value" : "h2"
                    }
                  ],
                  "op" : "add_header"
                }
              ],
              "op" : "primitive"
            }
```

Considering that `h1` is valid before the assignment and the header contains only one field `f`, it can be rewritten to:
```
        h2 = h1;
```

In bmv2 json:
```
               {
                  "parameters" : [
                    {
                      "type" : "header",
                      "value" : "h4"
                    },
                    {
                      "type" : "header",
                      "value" : "phdr_0_h1"
                    }
                  ],
                  "op" : "assign_header"
                }
```

In this case `simple_switch` works as expected.

I tracked down the cause of the issue to https://github.com/p4lang/behavioral-model/blob/main/targets/simple_switch/primitives.cpp#L231:
```
class add_header : public ActionPrimitive<Header &> {
  void operator ()(Header &hdr) {
    // TODO(antonin): reset header to 0?
    if (!hdr.is_valid()) {
      //hdr.reset();
      hdr.mark_valid();
      // updated the length packet register (register 0)
      auto &packet = get_packet();
      packet.set_register(RegisterAccess::PACKET_LENGTH_REG_IDX,
          packet.get_register(RegisterAccess::PACKET_LENGTH_REG_IDX) +
          hdr.get_nbytes_packet());
    }
  }
};
```

Where the header is reset before marking it valid in `add_header`, so the previously assigned value in `set` operation is cleared.

It seems to be intentional because of this commit (https://github.com/p4lang/behavioral-model/commit/c4b01883a35c54ec0eaae057347152d8fa0c56d0), but probably it could be made more generic to support such constructs in P4_16 as I mentioned before, as I don't see anything mentioning that `setValid()` method should reset the header in P4_16 spec (feel free to correct me if I'm wrong).

Therefore I changed the test cases to contain the header assignment instead of assigning value to field and setting the header valid by `setValid()` method.

mbudiu-vmw:
If a header is invalid any write to its fields is not guaranteed to have an effect.
However, if the header is valid and you call setValid, there should be no effect on any of the fields that are set.
The implementation in bmv2 looks right to me.
Are you saying otherwise?

mvido:
You're right. I missed that before. It's specified in section 8.22 of the spec:

> The result of reading a value in any of the situations below is that some unspecified value will be used for that field.
> ...
> * reading a field from a header that is currently valid, but the field has not been initialized since the header was last made valid.

Anyway, I already fixed the test cases, so this is not an issue anymore.

mbudiu-vmw:
This looks good.
I can't tell for sure from reading the code whether the tests use the stf files twice, once with the optimization and once without.
If that is correct, then this looks ready to merge.

mbudiu-vmw (approved):

mvido:
@mbudiu-vmw Thank you! Sorry for the late response, I have been on a vacation. When you check the details from CI, f.e. for test parser-inline-test1.p4 you can see the following output in the log (f.e. this log: https://github.com/p4lang/p4c/runs/3413639504):
```
        Start 624: bmv2-parser-inline-opt-enabled/testdata/p4_16_samples/parser-inline/parser-inline-test1.p4
1833/2098 Test #624: bmv2-parser-inline-opt-enabled/testdata/p4_16_samples/parser-inline/parser-inline-test1.p4 .....   Passed    2.89 sec
```
```
        Start 611: bmv2-parser-inline-opt-disabled/testdata/p4_16_samples/parser-inline/parser-inline-test1.p4
1911/2098 Test #611: bmv2-parser-inline-opt-disabled/testdata/p4_16_samples/parser-inline/parser-inline-test1.p4 ....   Passed    2.90 sec
```

Which corresponds to the following code from `backends/bmv2/CMakeLists.txt`:
```
  p4c_add_tests("bmv2-parser-inline-opt-disabled" ${BMV2_DRIVER} "${BMV2_PARSER_INLINE_TESTS}" "")
  p4c_add_tests("bmv2-parser-inline-opt-enabled" ${BMV2_DRIVER} "${BMV2_PARSER_INLINE_TESTS}" "" "-a=--parser-inline-opt")
```

So yes, the tests should be executed really twice.

PR #2804, opened 2021-06-16
Refactor error logging

jneruda:
Motivation behind this PR is to be able to use custom sinks for logging errors and to have maximum flexibility when formatting these messages.

Originally, all error/warning messages were send to `error_helper` functions which used template metaprogramming to serialize arguments into `boost::format` object and to add diagnostic information to the message. It used four string values to hold intermediate results and returned one formatted string at the end.

This PR uses dedicated struct for holding those intermediate values, most notably storing vector of `SourceInfo`s instead of serializing it into strings. This struct (`ErrorMessage`) contains `toString`` method which yields the same formatted strings as before, but `ErrorReporter::emit_message` function has been modified to accept this `ErrorMessage` object rather than preformatted string and this method has been marked `virtual` so any user can overload it and their own log sink.

Also, `CompileContext::errorReporter` accessor has been declared virtual so users can instantiate their own `ErrorReporter`s with custom sinks withing their custom `CompileContext  and due to nature of implementation of `CompileContext` stack, their `ErrorReporter` will be automatically used once the appropriate `CompileContext` is pushed to the shared stack.

> This change is **NOT** backwards compatible, but migration is very easy, as demonstrated in `typeConstraints.h`. A version of `error_helper` method with backwards compatible list of arguments has been preserved, also one simplified version has been added and since these methods now return `ErrorMessage` instead of `std::string`, users have to call `toString` on the returned object.

### DIFF
--- a/lib/CMakeLists.txt
+++ b/lib/CMakeLists.txt
@@ -19,6 +19,7 @@ set (LIBP4CTOOLKIT_SRCS
 	crash.cpp
 	cstring.cpp
         error_catalog.cpp
+		error_message.cpp
         exename.cpp
 	gc.cpp
 	gmputil.cpp
@@ -47,6 +48,7 @@ set (LIBP4CTOOLKIT_HDRS
 	enumerator.h
 	error.h
         error_catalog.h
+		error_message.h
         error_helper.h
 	error_reporter.h
 	exceptions.h
--- a/lib/error_helper.h
+++ b/lib/error_helper.h
@@ -151,7 +151,6 @@ auto error_helper(boost::format& f, ErrorMessage out, const T *t, Args... args) 
 template<typename T, class... Args>
 auto error_helper(boost::format& f, ErrorMessage out, const T &t, Args... args) ->
     typename std::enable_if<std::is_base_of<Util::IHasSourceInfo, T>::value, ErrorMessage>::type {
-
     auto info = t.getSourceInfo();
     if (info.isValid()) out.locations.push_back(info);
     return error_helper(f % t.toString(), out, std::forward<Args>(args)...);
--- a/lib/error_helper.h
+++ b/lib/error_helper.h
@@ -28,169 +28,247 @@ limitations under the License.
 #include "lib/source_file.h"
 #include "lib/stringify.h"
 
+struct ErrorMessage {
+    enum class Type : std::size_t { None, Error, Warning };
+
+    Type type = Type::None;
+    std::string prefix = "";  // Typically error/warning type from catalog
+    std::string message = "";  // Actual formatted message
+    std::vector<Util::SourceInfo> locations; // Relevant source locations for this error
+    Util::SourceInfo position;  // Position in P4 source
+    std::string tail = "";  // Extra P4 source positions
+    std::string suffix = "";  // Used by errorWithSuffix
+
+    ErrorMessage() {}
+    // Invoked from backwards compatible error_helper
+    ErrorMessage(const std::string &prefix, const Util::SourceInfo &info, std::string &suffix)
+        : prefix(prefix), position(info), suffix(suffix) {}
+    // Invoked from error_reporter
+    ErrorMessage(Type type, const std::string &prefix, const std::string &suffix)
+        : type(type), prefix(prefix), suffix(suffix) {}
+
+    std::string getPrefix() const {
+        std::string p = prefix;
+        if (type == Type::Error) {
+            if (p.empty()) p = "error: ";
+            else p = "[--Werror=" + p + "] error: ";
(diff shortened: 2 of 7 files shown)


lib/error_helper.h
```diff
@@ -322,8 +310,9 @@ auto bug_helper(boost::format& f, std::string message, std::string position,
         position = posString;
         posString = "";
     } else {
-        if (!posString.isNullOrEmpty())
+        if (!posString.isNullOrEmpty()) {
             posString += "\n";
+        }
```
jneruda, line 315:
```suggestion
```

lib/error_helper.h
```diff
@@ -322,8 +310,9 @@ auto bug_helper(boost::format& f, std::string message, std::string position,
         position = posString;
         posString = "";
     } else {
-        if (!posString.isNullOrEmpty())
+        if (!posString.isNullOrEmpty()) {
```
jneruda, line 313:
```suggestion
        if (!posString.isNullOrEmpty())
```

lib/error_helper.h
```diff
@@ -301,8 +288,9 @@ std::string bug_helper(boost::format& f, std::string message, std::string positi
...
             posString += "\n";
+        }
```
jneruda, line 293:
```suggestion
```

lib/error_helper.h
```diff
@@ -301,8 +288,9 @@ std::string bug_helper(boost::format& f, std::string message, std::string positi
         position = posString;
         posString = "";
     } else {
-        if (!posString.isNullOrEmpty())
+        if (!posString.isNullOrEmpty()) {
```
jneruda, line 291:
```suggestion
        if (!posString.isNullOrEmpty())
```

lib/error_helper.h
```diff
@@ -341,8 +330,9 @@ auto bug_helper(boost::format& f, std::string message, std::string position,
...
```
jneruda, line 333:
```suggestion
        if (!posString.isNullOrEmpty())
```

lib/error_helper.h
```diff
@@ -341,8 +330,9 @@ auto bug_helper(boost::format& f, std::string message, std::string position,
...
             posString += "\n";
+        }
```
jneruda, line 335:
```suggestion
```

lib/CMakeLists.txt
```diff
@@ -47,6 +48,7 @@ set (LIBP4CTOOLKIT_HDRS
 	enumerator.h
 	error.h
         error_catalog.h
+		error_message.h
```
jneruda, line 51:
```suggestion
	error_message.h
```

lib/CMakeLists.txt
```diff
@@ -19,6 +19,7 @@ set (LIBP4CTOOLKIT_SRCS
 	crash.cpp
 	cstring.cpp
         error_catalog.cpp
+		error_message.cpp
         exename.cpp
```
jneruda, line 23:
```suggestion
        error_catalog.cpp
	error_message.cpp
        exename.cpp
```

lib/error_message.h
```diff
@@ -0,0 +1,73 @@
...
+ *  might have SourceInfo attribute. ::error_helper parse those parameters, format
+ *  parameters to output message and extracts SourceInfo wherever possible.
+ *
+ *  Populated structure can be serialized to canonical error message with toString() method.
+ *
+ *  This structure is mainly used inside ErrorReporter, but some uses invoke ::error_helper
+ *  directly and those uses need to call toString() on returned object.
+ */
+struct ErrorMessage {
+    enum class Type : std::size_t { None, Error, Warning };
```
mbudiu-vmw, line 37: Let's name this something else, the IR::Type class is used too many times in the code.
How about ErrorType?

mbudiu-vmw (commented):
This does look nicer.
However, you will have to make sure that all tests pass. In particular, if the reference outputs produced have changed you have to check in the modified versions. Prior to that you will have to manually check that the output is actually reasonable; I expect many tests from the ./err group will be affected, so please make sure that the error messages produced are readable.

lib/error_message.h
jneruda: Changed it to MessageType

mbudiu-vmw:
Is this ready for review?

mbudiu-vmw (approved):

mbudiu-vmw:
I am not sure @cc10512 is still reviewing code for p4c.

mbudiu-vmw:
If this is ready we can merge it.

jneruda:
> If this is ready we can merge it.

Sorry, I was out of office, thank you for review, I merged the branch.

PR #2813, opened 2021-06-22
DPDK backend: Add support for meter and counter extern

usha1830:
This PR adds support for meter and counter PSA extern in DPDK backend and a minor change to fix jump statements in action body.

**DPDK target's requirements**
- Current PSA Meter extern API does not allow the user the specify a packet length, but DPDK target expects the packet length to come from the user. To handle this, Meter execute method API is customized to take an additional optional parameter for packet length.

` PSA_MeterColor_t execute(in S index, in PSA_MeterColor_t color, @optional in bit<32> pkt_len);
` 

- Similarly, for PSA counter extern, packet length is required for PACKET counter. The Counter count method API is customized to take additional optional parameter for packet length.

`void count(in S index, @optional in bit<32> increment);`

- DPDK target implements counter using register. Register declaration "regarray" instruction is used for counter declaration and atomic addition of registers "regadd" is used for counter incrementing. The packet counters are incremented by length specified by the user as argument to count method and byte counters are incremented by 1.


backends/dpdk/dpdkHelpers.cpp
```diff
@@ -73,6 +73,27 @@ bool ConvertStatementToDpdk::preorder(const IR::AssignmentStatement *a) {
...
                         left, e->object->getName(), intermediate);
                 }
+            } else if (e->originalExternType->getName().name == "Meter") {
+                if (e->method->getName().name == "execute") {
+                    auto argSize = e->expr->arguments->size();
+
+                    // DPDK target needs index and packet length as mandatory parameters
+                    if (argSize < 2)
+                        ::error(ErrorType::ERR_UNEXPECTED, "Expected atleast 2 arguments for %1%",
+                                e->object->getName());
```
mbudiu-vmw, line 83: do you want to return in this case? otherwise your executestatement below will have some nullptr arguments.

backends/dpdk/spec.cpp
```diff
@@ -163,10 +163,41 @@ std::ostream &IR::DpdkExternDeclaration::toSpec(std::ostream &out) const {
         }
     }
     else if ( DPDK::toStr(this->getType()) == "Counter") {
-    //TODO yet to be implemented
+        auto args = this->arguments;
+        unsigned value = 0;
+        if (args->size() < 2) {
+          ::error ("Counter extern declaration %1% must contain 2 parameters\n", this->Name());
```
mbudiu-vmw, line 169: your indentation is not consistent with our standard

p4include/dpdk/psa.p4
```diff
@@ -0,0 +1,863 @@
+/* Copyright 2013-present Barefoot Networks, Inc.
```
mbudiu-vmw, line 1: Is this psa.p4 different from the other one?
If not, does it have to be here?

mbudiu-vmw (changes requested):

p4include/dpdk/psa.p4
usha1830: Yes. This is slightly different from the standard psa.p4 in p4include directory.

This psa.p4 extends the API for Meter execute and Counter count methods. A new optional parameter is added to both of these methods to take packet length/counter increment value  as a parameter.

Meter extern's execute method:
`PSA_MeterColor_t execute(in S index, in PSA_MeterColor_t color, @optional in bit<32> pkt_len);`

Counter extern's count method
`void count(in S index, @optional in bit<32> increment);`

This API is required to meet DPDK target's requirements.

usha1830: Please also refer to https://github.com/p4lang/p4c/pull/2751 and https://github.com/p4lang/p4c/pull/2818 PRs which proposes similar extensions to PSA

usha1830:
@mbudiu-vmw Thanks for the review. I will address rest of the comments and resubmit the changes.

backends/dpdk/dpdkArch.h
```diff
@@ -446,32 +446,95 @@ class ConvertInternetChecksum : public PassManager {
...
+
+    bool preorder(const IR::Declaration_Instance *d) override {
+        if (d->type->is<IR::Type_Specialized>()) {
+            auto type = d->type->to<IR::Type_Specialized>();
+            auto externTypeName = type->baseType->path->name.name;
+            if (externTypeName == "Meter"){
+                if (d->arguments->size() != 2) {
+                    ::error("%1%: expected number of meters and type of meter as arguments", d);
+                } else {
+                    if (d->arguments->at(1) == 0)
```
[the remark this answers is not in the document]
usha1830: I am facing an issue w.r.t adding a test for this case.

I added a test in testdata/p4_16_samples/ for emitting this warning and added the warning in reference stderr output "psa-example-dpdk-meter1.p4-stderr". This test compiles as expected and produces a warning when compiled with p4c-dpdk.  However, **make check** fails while compiling this test using _p4test_. It fails because of output mismatch as there is no warning when compiling with p4test and reference stderr output has warning. What should I do in this case?

p4include/dpdk/psa.p4
mbudiu-vmw: I find this very confusing. The whole point of PSA is to be portable. If we have two different PSA.p4 I don't see how portability is helped. Only confusion can ensue. 
We should find a resolution to the issues you cite before merging this file.
I would expect that extensions can be done by including psa.p4 and then additional libraries. But the externs in these libraries would have to have different names.
In the meantime we could perhaps use a different name for this file.

backends/dpdk/dpdkArch.h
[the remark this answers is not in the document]
mbudiu-vmw: So far the p4 reference outputs have always been produced by p4test.
Is the dpdk backend testing framework also saving reference outputs?
Ideally it should only save the .spec output. Then there should be no conflict: the .p4 and -stderr outputs come from p4test, and the .spec output from p4c-dpdk. 
If both compilers produce reference outputs we have a second problem, since the test has an inherent data race. When we run tests on multiple cores they could overwrite the reference outputs files. So this should not be done. The likelihood of this happening is low, since we have many tests, but it's not zero.
Finally, note that we have already this problem where a test should pass with p4test and fail with a specific backend which has more constraints. Then the solution is to list the test as xfail into the makefile of the backend where it is supposed to fail, with comments explaining why.

usha1830:
@mbudiu-vmw I have completed the requested changes.  Could you please review it again?
 - Combined the extern declaration collection pass for Register, Counters and meters. 
 - Added one more test for using multiple externs in same example. 
 - Updated the dpdk test script to avoid saving error and P4 runtime outputs.

Regarding the customization of psa.p4, can we conclude on going with the proposed approach of having customized psa.p4 in target specific include directory (similar to https://github.com/p4lang/p4c/pull/2818)?

backends/dpdk/dpdkHelpers.cpp
```diff
@@ -412,21 +435,46 @@ bool ConvertStatementToDpdk::preorder(const IR::MethodCallStatement *s) {
...
             if (a->method->getName().name == "execute") {
-                auto args = a->expr->arguments;
-                auto index = args->at(0)->expression;
-                auto color = args->at(1)->expression;
-                auto meter = a->object->getName();
-                add_instr(
-                    new IR::DpdkMeterExecuteStatement(meter, index, color));
+                // DPDK target requires the result of meter execute method is assigned to a
+                // variable of PSA_MeterColor_t type.
+                ::error(ErrorType::ERR_UNSUPPORTED, "LHS of meter execute statement is missing " \
```
mbudiu-vmw, line 440: no return here?

p4include/dpdk/psa.p4
mbudiu-vmw: What is the relation of this file and the one from https://github.com/p4lang/p4c/pull/2818?

mbudiu-vmw (approved):
I am accepting this, but I still have a couple of questions. Please take a look.

usha1830:
>>What is the relation of this file and the one from #2818?

Both this PR and #2818 PR add the same new file "dpdk/psa.p4".

This new file is derived from the top-level psa.p4. This PR extends the top-level psa.p4 by modifying the signatures for **Meter _execute_ and Counter _count_** methods.

The PR #2818 modifies the DPDK target data plane bit widths.

Both of these customized psa.p4 files should be merged.

backends/dpdk/dpdkHelpers.cpp
usha1830: I don't think so, there is already a return just after the if-else chain.

Issue #2791, opened 2021-06-05
Support tuple with parser select for EBPF backend


hesingh:
This is the TODO in code.

https://github.com/p4lang/p4c/blob/08ef61a007b47b48945f86d60fc47b81b4d0fef1/backends/ebpf/ebpfParser.cpp#L128

mbudiu-vmw:
Have you tried it? It actually works fine up for select labels up to 64 bits width.

hesingh:
Wow, sorry no.  Gergely's code uses `_` don't care in the tuple and maybe that is what we need to support.

mbudiu-vmw:
Tuples in select computations are converted into concatenations, which are furthermore converted into | and shifts. As long as the total width of the select expression does not exceed 64 bits this should work. There is an issue if the width is larger, though.

hesingh:
> Tuples in select computations are converted into concatenations, which are furthermore converted into | and shifts. As long as the total width of the select expression does not exceed 64 bits this should work. There is an issue if the width is larger, though.

I checked Gergely's code and see all tuple using don't care or not use less than 64 bits of total width - Yay!

PR #2815, opened 2021-06-22
Fix off-by-one bug in ebpf backend mask computation

mbudiu-vmw:


mbudiu-vmw:
Fixes #2791

hesingh:
Does it make sense to add a tuple P4 test with don't care element in tuple?

mbudiu-vmw:
Sure, please create one.

hesingh:
> Sure, please create one.

Please add this parser state to your existing test.  I haven't compiled this code. thanks.
```p4
state ip {
        p.extract(headers.ipv4);
	transition select(headers.ipv4.protocol, diffserv) {
	(_, _) : accept;
        default : reject;
    }
```

mbudiu-vmw:
This is a correct observation, the pass that eliminates multi-argument selects fails with partial "default" labels.

mbudiu-vmw:
We should file a new issue for this case.

hesingh:
> We should file a new issue for this case.

https://github.com/p4lang/p4c/issues/2816

hesingh:
Just curious where is the tuple eliminated?

mbudiu-vmw:
The pass is called SingleArgumentSelect, but it is not general enough unfortunately.

mbudiu-vmw:
Maybe we should modify it to produce a value and a mask always

hesingh:
> Maybe we should modify it to produce a value and a mask always

Makes sense.

hesingh:
I tested the Gergely (without his  don't care tuple) code and it works - thanks.

mbudiu-vmw:
This is a 1-line change really, should be very easy to review.

hanw (approved):

PR #2818, opened 2021-06-23
Support target-specific customizations of psa.p4 include file

hanw:
This PR supersedes https://github.com/p4lang/p4c/pull/2751

This PR implements the mechanism to handle any desired differences between data-plane specific bit widths for different targets, e.g. for BMv2 psa_switch process, versus DPDK back end implementing PSA architecture.

Each backend searches for psa include file in the "p4include/dpdk" directories first. If the file is found, it will use in preprocessing, otherwise, the backend will continue to search in the "p4include" directories and use the default psa.p4 file.

The customized psa.p4 files are slightly different from the default psa.p4 provided in psa spec.


mbudiu-vmw:
I think the file should not be called psa.p4. Having more than 1 psa.p4 sounds to me like a recipe for disaster.
You can call it dpdk.p4.

jfingerh:
There can be multiple implementations of the PSA architecture, certainly for different targets like BMv2 and DPDK.

Why would you call the psa.p4 include file for DPDK `dpdk.p4`?

What if later the DPDK back end implements the v1model architecture, or the PNA architecture?   What would you call those include files?

mbudiu-vmw:
There can be multiple implementations of PSA, but not multiple definitions of PSA. This PR seems to add an alternative definition of PSA.
With respect to the name we can be more specific, e.g., psa-dpdk.p4. Or almost-psa-dpdk.p4.

jfingerh:
The psa.p4 include file is _intended by design_ to be customized for each back end, e.g. to supply target-specific bit widths for types like PortId_t.  That should be the primary difference between the psa.p4 include file for different targets, but there might be cases where different back ends would like to add target-specific annotations to extern definitions.

mbudiu-vmw:
Is there a definition of what are allowed customizations? Where is the boundary? Can you add new methods? Can you change the semantics of existing methods? Can you remove metadata fields? Can you change the signature of controls?
Even assuming that these changes are within scope, duplicating the whole file hardly seems like a good engineering solution. We should really break psa.p4 into pieces that are included into the main psa.p4; this way derived versions can reuse these pieces and substitute some.

jfingerh:
Do any of your suggestions avoid the need of separate target-specific customized psa.p4 include files?  I don't see any, but I could have missed them.

If they still need it, then please approve either this PR, or something functionally equivalent that lets there be multiple top level psa.p4 include files for different targets.

mbudiu-vmw:
The main question we have to agree on is "how these files are named".
If they are all named psa.p4, then the include path is the way to differentiate them.
That is not part of the p4 program, it is part of the build system around p4. So when you share a p4 file that includes psa.p4 you will have to give extra information about what that is supposed to mean.
The experience with C header files with the same name is not really great.
I don't like the solution for v1model.p4 based on `#define` symbols, but that seems a bit nicer.
Can we get @ChrisDodd to weigh on this issue?

jfingerh:
The intent of PSA is that including psa.p4 means that you get the definitions that the PSA specification has, at least.

When you compile such a P4 program, you will use either a target-specific compiler, e.g. p4c-dpdk, or the p4c driver with a `--target` option.  The `--target` option seems to be perhaps the "extra information" you mention in your previous comment, but I'm not sure if it is what you meant by that.  Ideally you can give someone a file that includes psa.p4, and not know what the target device is that someone else might compile it for.  So you shouldn't have to include any information about the target in the P4 source file at all, yes?

If you want to take the same P4 source program and compile it for two different targets, without changing it, and if psa.p4 needs different contents for different targets, then that is what leads to this desire for different include directory paths for different targets.

Most of the other proposals I can think of would require changing the P4 source code.  I'd rather avoid those.

mbudiu-vmw (approved):

mbudiu-vmw:
I have approved this, but we should really factor out the common contents in some other file and include it. psa-common.p4.
Perhaps we can also have some common files with definitions that are not supposed to be changed by anyone.

jfingerh:
We can do that, Mihai, but even if we factor out 99.9% of it as common, how do you propose handling the different parts, without separate include paths?

mbudiu-vmw:
I have approved it, so it may be that accepted this solution.

jnfoster:
Without being obstinate, I tend to agree with Mihai. Allowing multiple but different files named `psa.p4` to proliferate seems high risk. 

Don't want to hold thing up but his suggestions to factor out the common and therefore immutable parts -- so the parts that can be customized by a target are more explicit -- and to come up with a naming scheme that is clearer, both seem like good ideas.

jfingerh:
Nate Foster wrote: "Without being obstinate, I tend to agree with Mihai. Allowing multiple but different files named psa.p4 to proliferate seems high risk.

Don't want to hold thing up but his suggestions to factor out the common and therefore immutable parts -- so the parts that can be customized by a target are more explicit -- and to come up with a naming scheme that is clearer, both seem like good ideas."

I understand the desire to keep common things common.  I'm not against that.  But all the ways of making common things common still leaves the top level psa.p4 include file for different targets _different_, even if they include a lot of common parts within them.

jfingerh:
I should amend my previous comment a bit, in particular the phrase "all of the ways to make common things common".

If folks want to create a convention in P4 of having some #define symbol, e.g. named something like "TARGET", and assign distinct target names to all P4 targets, then we could do #ifdef's inside of a psa.p4 include file that had target-specific parts to it.  If you want to do that, please put forth such a proposal and we can consider it.

mbudiu-vmw:
It is really a question of interpretation. If we think of psa.p4 as a *family* of architectures then it makes more sense to have multiple files (the naming scheme won't matter that much, but using the same name may be ok). The psa spec should specify this explicitly: psa switches can differ in some respects but some things are invariant. What are the invariant things? What can be different in a target but still allow it to adhere to the psa spec? Can you add new externs? Can you add new annotations? Can you elide some externs?

jfingerh:
@mbudiu-vmw @jnfoster I'm glad this issue has been approved and we can move forward on this.

This is a side comment, regarding earlier comments in this Github comment thread from Mihai on worries that PSA implementations might have differences between them, and wanting to be very precise in exactly how they are allowed to differ.

I understand that is a desirable goal, but I wanted to remind you of this sentence near the beginning of the P4_16 language specification, upon which PSA rests: "It is understood that some implementations may be unable to implement the behavior defined here in all cases, or may provide options to eliminate some safety guarantees in exchange for better performance or handling larger programs. They should document where they deviate from this specification."  If the language spec isn't hyper-vigilant on exactly what constitutes a conformant implementation, it seems a bit unfair to say that PSA is somehow supposed to be _more_ precise and demanding in its specification than the language upon which it rests.

mbudiu-vmw:
We have already merged a version of this file #2813 
Should we close this PR?

hanw:
I will rebase this PR. Changes are still required to find the dpdk/psa.p4 file in include path.

p4include/dpdk/psa.p4
```diff
@@ -57,10 +47,16 @@ limitations under the License.
...
-typedef bit<32> MulticastGroupUint_t;
-typedef bit<16> CloneSessionIdUint_t;
-typedef bit<8>  ClassOfServiceUint_t;
+
+/* WARNING!  Andy Fingerhut made up the following numbers for the DPDK
+ * target data plane bit widths.  No one should believe they are
+ * accurate until they * have been approved by the p4c-dpdk
+ * implementers. */
+
+typedef bit<32>  PortIdUint_t;
```
jfingerh, line 56: @hanw Do you want DPDK data plane type for PortId_t to be 32 bits wide?  OK if so, just making sure that is what you intended.

The const values for the CPU and recirculate ports in the original psa.p4 include file are simply the numeric values that are used by BMv2, and are also intended to be customized for each PSA implementation, and should be as many bits wide as this type is.

hanw: I sent an email to Cristian to confirm what would be a good value for PortId_t and whether non-power-of-2 width are okay for DPDK.

PR #2820, opened 2021-06-24
Move RemoveActionParameters pass to frontend

### LINKED ISSUE #2266: "Compiler Bug Null stat" also triggered in action properties
PR https://github.com/p4lang/p4c/pull/2259 does not catch function calls in action arguments. Are there any other table properties where this could occur?

[2258b.p4.txt](https://github.com/p4lang/p4c/files/4384221/2258b.p4.txt)

mbudiu-vmw:
Fixes #2266 
This pass is used in all midends and in the control plane API generation.
Without it we cannot really solve #2266: if we do not remove action parameters there is no place where to place the body of inlined functions that are arguments of actions.
The code changes are very small, but unfortunately this change has renamed many reference outputs because the names of temporary variables generated have changed.


frontends/p4/validateParsedProgram.cpp
```diff
@@ -201,13 +201,8 @@ void ValidateParsedProgram::postorder(const IR::EntriesList* l) {
                 "%1%: invalid initializer. Table initializers must be constant.", l);
 }
 
-/// Switch statements are not allowed in actions.
```
fruffy, line 204: Is this related to https://github.com/p4lang/p4-spec/pull/945?

fruffy (approved):

frontends/p4/validateParsedProgram.cpp
mbudiu-vmw: There is a little change related to that.
But, as I said in the spec PR, because you can't call tables in actions you can't use the old switch statements anyway.
So I think this check does not change anything really.

PR #2821, opened 2021-06-24
Cancel active workflows on new push

mbudiu-vmw:
I don't really understand github actions, but I copied some code I found on the net.
When we have multiple push requests to the same PR or branch we run needless workflows.
I hope that this cancels all runs of the same branch.

### DIFF
--- a/.github/workflows/ci-test.yml
+++ b/.github/workflows/ci-test.yml
@@ -25,6 +25,11 @@ jobs:
       with:
         submodules: recursive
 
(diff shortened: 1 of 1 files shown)


fruffy:
https://stackoverflow.com/a/67939898/3215972

The second option might be more elegant since it relies on Github-native code. The other is a custom actions. But I have no idea if either will work as expected. Let's try it out?

fruffy (approved):

PR #2827, opened 2021-06-26
Allow masks in select expressions

### LINKED ISSUE #2816: Add support for don't care element in tuple for parser select
```p4
state ip {
        p.extract(headers.ipv4);
	transition select(headers.ipv4.protocol, headers.ipv4.diffserv) {
	(_, _) : accept;
        default : reject;
    }
```

mbudiu-vmw:
Fixes #2816 
@osinstom : you may want to consider applying similar fixes to the ubpf back-end.
In fact, this scheme is better than it used to be, but it is still not general enough: it won't handle select expressions where the expression has more than 64 bits combined.

### DIFF
--- a/midend/singleArgumentSelect.cpp
+++ b/midend/singleArgumentSelect.cpp
@@ -2,29 +2,69 @@
 
 namespace P4 {
 
-static const IR::Expression* convertList(const IR::Expression* expression) {
+DoSingleArgumentSelect::Pair::Pair(const IR::Expression* e, const IR::Type* type) {
+    auto srcInfo = e->srcInfo;
+    if (auto m = e->to<IR::Mask>()) {
+        expr = m->left;
+        mask = m->right;
+        hasMask = true;
+    } else if (e->is<IR::DefaultExpression>()) {
+        expr = new IR::Constant(srcInfo, type, 0, 16);
+        mask = new IR::Constant(srcInfo, type, 0, 16);
+        hasMask = true;
+    } else {
+        expr = e;
+        mask = new IR::Constant(srcInfo, type, Util::maskFromSlice(type->width_bits()-1, 0), 16);
+        hasMask = false;
+    }
+}
+
+static const IR::Expression* convertList(
+    const IR::Expression* expression, const IR::Type* selectListType) {
+    if (expression->is<IR::DefaultExpression>()) {
+        int width = selectListType->width_bits();
+        auto type = new IR::Type_Bits(width, false);
+        return new IR::Mask(expression->srcInfo,
+                            new IR::Constant(type, 0, 16),
+                            new IR::Constant(type, 0, 16));
+    }
     auto list = expression->to<IR::ListExpression>();
     if (list == nullptr)
         return expression;
     BUG_CHECK(list->components.size() > 0, "%1%: No components", list);
-    auto expr = list->components.at(0);
-    for (size_t i = 1; i < list->components.size(); i++)
-        expr = new IR::Concat(expr, list->components.at(i));
-    return expr;
+    auto tt = selectListType->checkedTo<IR::Type_List>();
+    BUG_CHECK(list->size() == tt->components.size(),
+              "%1% and %2% do not have the same size",
+              list, tt);
+    auto p = new DoSingleArgumentSelect::Pair(list->components.at(0), tt->components.at(0));
+    auto hasMask = p->hasMask;
+    auto expr = p->expr;
+    auto mask = p->mask;
+    for (size_t i = 1; i < list->components.size(); i++) {
+        p = new DoSingleArgumentSelect::Pair(list->components.at(i), tt->components.at(i));
+        expr = new IR::Concat(expr, p->expr);
+        mask = new IR::Concat(mask, p->mask);
+        hasMask = hasMask || p->hasMask;
+    }
+    if (!hasMask)
+        return expr;
+    else
+        return new IR::Mask(expression->srcInfo, expr, mask);
 }
 
-bool SingleArgumentSelect::preorder(IR::SelectExpression* expression) {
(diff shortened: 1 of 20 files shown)


hesingh:
Not a surprise that 37 ebpf tests have failed.  The reference output for these tests have to be updated.

hesingh:
The same error is seen for  few tests I checked.
```
*** Error 2: Failed to compile P4:
/p4c/testdata/p4_16_samples/key-issue-1020_ebpf.p4(16): [--Werror=unsupported] error: DefaultExpression: not yet implemented
            default : reject;
            ^^^^^^^
make: *** [/p4c/build/tmpv41__tqz/test.c] Error 1
```

hesingh:
```
285/1982 Test #658: ubpf/testdata/p4_16_samples/lookahead_ubpf.p4 .................................***Failed    0.32 sec
*** Error 2: Failed to compile P4:
/p4c/testdata/p4_16_samples/lookahead_ubpf.p4(37): [--Werror=unsupported] error: &&&: not yet implemented
            default: accept;
            ^^^^^^^
make: *** [/p4c/build/tmpiggk360_/test.c] Error 1
```

mbudiu-vmw:
@osinstom : I had to modify the ubpf back-end in a similar way with the ebpf back-end. 
There is a lot of code duplication between the two, it would have been nice to reuse it, but given the actual code structure at this point it would be a lot of work to unify them. So I had to essentially make the same changes in your back-end.
Can you please review these? 
The changes strictly increase the range of programs accepted: they can now handle selects with masks in parsers.
Note that default is now converted to a mask of all 0s in the SingleArgumentSelect pass, that's why I had to modify the ubpf backend as well.

hesingh:
The code changes need more work.

Only if an if-statement in generated C code includes an equality check with a mask, then double parenthesis are needed, e.g.,

`if ((select_7 & 0x0) == (0x0 & 0x0))`

However, if the generated code is this, the double parenthesis generated code is wrong and causes a C compiler error. 

`if ((select_5 == 65359))`

```
error: equality comparison with extraneous parentheses [-Werror,-Wparentheses-equality]
        if ((select_5 == 65359))goto parse_rlc;
             ~~~~~~~~~^~~~~~~~
```

hesingh:
The double parenthesis are only needed in generated code if the tuple includes any don't care, such as in the P4 code example below.

```
state parse_gtpu {
	pkt.extract(hdr.gtpu_hdr);
	transition select(hdr.gtpu_hdr.e, hdr.gtpu_hdr.s, hdr.gtpu_hdr.pn) {
            (_, _, _) : parse_gtpopt;
	    (_, 1,_) : parse_gtpopt;
	    (_, _, 1) : parse_gtpopt;
	    default   : parse_ue_hdr;
	}
    }
```

because this is the generated code

`if ((select_7 & 0x0) == (0x0 & 0x0))goto parse_gtpopt;`

hesingh:
This code fixes the issue I just reported today.

```
bool StateTranslationVisitor::preorder(const IR::SelectCase* selectCase) {
    builder->emitIndent();
    if (auto mask = selectCase->keyset->to<IR::Mask>()) {
        builder->appendFormat("if ((%s", selectValue);
        builder->append(" & ");
        visit(mask->right);
        builder->append(") == (");
        visit(mask->left);
        builder->append(" & ");
        visit(mask->right);
	builder->append("))");
    } else {
        builder->appendFormat("if (%s", selectValue);
        builder->append(" == ");
        visit(selectCase->keyset);
	 builder->append(")");
    }
    builder->append(" goto ");
```

hesingh:
This PR lacks a test program to test don't care in a tuple used by parser select.  I have attached a P4 program.

[ebpf-tuple-p4.txt](https://github.com/p4lang/p4c/files/6795373/ebpf-tuple-p4.txt)

However, when I use the program with this PR code changes, I hit a `Compiler Bug`.  Why should  20 bits not be allowed in the tuple that `widthCheck` in `p4c/backends/ebpf/codeGen.cpp` fails.

```$ ./p4c-ebpf ../testdata/p4_16_samples/ebpf-tuple.p4 -o foo.c
In file: /home/hemant/p4c-clone/backends/ebpf/codeGen.cpp:395
Compiler Bug: ../testdata/p4_16_samples/ebpf-tuple.p4(43): (bit<20>)((bit<12>)headers.ipv4.ihl << 8 | (bit<12>)headers.ipv4.protocol & 255) << 8 | (bit<20>)headers.ipv4.diffserv & 255;: Computations on 20 bits not yet supported
 transition select(headers.ipv4.ihl, headers.ipv4.protocol, headers.ipv4.diffserv) {
                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

hesingh:
@mbudiu-vmw Please expedite this PR when you get a chance.  Also, do we allow non-multiple of 8 bits for sum of all elements in he tuple? thanks.

midend/singleArgumentSelect.cpp
```diff
@@ -2,29 +2,67 @@
...
+        expr = m->left;
+        mask = m->right;
+        hasMask = true;
+    } else if (e->is<IR::DefaultExpression>()) {
+        expr = new IR::Constant(srcInfo, type, 0, 16);
+        mask = new IR::Constant(srcInfo, type, 0, 16);
+        hasMask = true;
+    } else {
+        expr = e;
+        mask = new IR::Constant(srcInfo, type, Util::maskFromSlice(type->width_bits()-1, 0), 16);
```
fruffy, line 17: What happens with zero-width bits here?

fruffy (approved):

midend/singleArgumentSelect.cpp
mbudiu-vmw: Good point. I will add a pass to eliminate operations on zero width values.

PR #2829, opened 2021-06-27
Enable automatic release on version change

rst0git:
This pull request enables automatic GitHub release (and git tag creation) every time the version of P4C has been changed.


hesingh:
Is this work discussed in any meeting and asked to do? I can always use a git commit id to pull a branch for p4c.

rst0git:
>Is this work discussed in any meeting and asked to do?

I'm sorry, I should have mentioned this. We discussed it on the [PSA meeting](https://docs.google.com/document/d/1BEERtI5dbZbd3EltqCHSrmc10Snpkf5PPYfRLRgq_Vw/edit#heading=h.rf4vp05f5s69).

Version.txt
```diff
@@ -0,0 +1 @@
+1.2.0
```
mbudiu-vmw, line 1: Perhaps this should be 1.2.2?

.github/workflows/ci-static-build.yml
```diff
@@ -0,0 +1,24 @@
+name: "static-build-p4c"
```
mbudiu-vmw, line 1: I find these files uncreadable. Can you add a little comment about what they provide?

mbudiu-vmw (commented):
@fruffy can you please review these files? You have written our github actions automation.
Thank you.

.github/workflows/ci-static-build.yml
fruffy-bfn: I added this one. I can make a separate pull request with a description. This one just tests for Ubuntu whether we can create a statically linked build of P4C.

fruffy: Reposting with the right acc. Yeah I added this workflow with the last pull request. The file shows up as new because it was renamed. Will add some more description in a new request. Essentially, it just compiles p4c with the BUILD_STATIC_RELEASE=ON option. To test whether a statically linked build works with Ubuntu 16.04.

.github/workflows/ci-release.yml
```diff
@@ -1,24 +1,26 @@
...
+      - name: Checkout
+        uses: actions/checkout@v2
+
+      - name: Get version
+        run: |
+          VERSION="v$(cat Version.txt)"
+          echo "VERSION=$VERSION" >> $GITHUB_ENV
+
+      - name: Release
+        uses: softprops/action-gh-release@v1
```
fruffy, line 22: Is there something more official? A lot of these are by random Github repositories.

Version.txt
fruffy: The version string is set in CMakeLists.txt at line 61. It would be great if those could be unified so that it is not necessary to maintain multiple version strings.

.github/workflows/ci-release.yml
```diff
@@ -1,24 +1,26 @@
...
+        with:
+          tag_name: ${{ env.VERSION }}
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
fruffy, line 26: This is a secret provided by Github right? This is not something a user has to add?

fruffy (changes requested):

.github/workflows/ci-release.yml
rst0git: I was hoping to use [`actions/create-release`](https://github.com/) but it is no longer maintained and the README recommends four alternatives. This one looks like the most prominent.

.github/workflows/ci-release.yml
rst0git: Yes, GitHub automatically creates [`GITHUB_TOKEN`](https://docs.github.com/en/enterprise-server@3.0/actions/reference/authentication-in-a-workflow#about-the-github_token-secret).

Version.txt
rst0git: CMakeLists.txt line 61 has been changed as follows:
```diff
-  set (P4C_SEM_VERSION_STRING "1.2.0+g202103291035~a69e52")
+  file(READ "Version.txt" P4C_SEM_VERSION_STRING)
```
This change should allow us to maintain a single version string stored in `Version.txt`. By keeping the version string as a separate file we could detect changes that would trigger the workflow run:
```yaml
on:
  push:
    branches:
      - main
    paths:
      - 'Version.txt'
```

rst0git: >Perhaps this should be 1.2.2?

Thanks, I have updated the PR to use `1.2.2`.

fruffy (approved):

PR #2833, opened 2021-07-01
Modernize run-bmv2-test a little

fruffy:
Update the file to use proper arguments etc instead of custom parsing. Makes extending the script much easier.
Also add a new option `--testfile`, which allows for specifying your own path for test files.

### DIFF
--- a/backends/bmv2/run-bmv2-test.py
+++ b/backends/bmv2/run-bmv2-test.py
@@ -61,9 +61,13 @@ PARSER.add_argument("--init", dest="init_cmds", default=[],
                     help="Run <cmd> before the start of the test")
 PARSER.add_argument("--observation-log", dest="obs_log",
                     help="save packet output to <file>")
+PARSER.add_argument("-tf", "--testfile", dest="testfile",
+                    help="Provide the path for the stf file for this test. "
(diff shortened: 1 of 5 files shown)


backends/bmv2/run-bmv2-test.py
```diff
@@ -18,41 +18,65 @@
...
+from scapy.layers.all import *
+from scapy.utils import *
 from bmv2stf import RunBMV2
-import pdb
 SUCCESS = 0
 FAILURE = 1
 
 
-class Options(object):
+PARSER = argparse.ArgumentParser()
```
mbudiu-vmw, line 34: I personally think that global variables are evil. Why not move this to the function using it?

backends/bmv2/run-bmv2-test.py
```diff
@@ -399,5 +327,45 @@ def main(argv):
     sys.exit(result)
 
 
+def check_path(path):
+    """Checks if a path is an actual directory and converts the input
```
mbudiu-vmw, line 331: I don't see where the check for being a directory is being made.

mbudiu-vmw (changes requested):

mbudiu-vmw (approved):

PR #2836, opened 2021-07-02
DPDK backend: Add support for tuple expressions in transition select statement

usha1830:
This PR adds support for the tuple expressions in transition select statement.
The changes include:
- New function for handling the tuple expressions.
- New helpers functions to refactor some of existing code.
- Added new tests and reference outputs

   The tuple expressions are evaluated as exact match of each element in the tuple and is
   translated to dpdk assembly by short-circuiting the conditions where beginning of each keyset
   starts with a unique case label.

   ```
      transition select(a,b) {      
             (a1, b1) : state1;
             (a2, b2) : state2;        
             (a3, b3) : state3;        
             }                         
         is evaluated as               
         if (a == a1 && b == b1)       
             goto state1               
         else (a == a2 && b == b2)     
             goto state2               
         .....
         
      Equivalent dpdk assembly:                            
       
      jmpneq case1 a, a1
      jmpneq case1 b, b1
      jmp state1
      case1:   jmpneq case2 a, a2
      jmpneq case2 b, b2
      jmp state2
      case2:    jmpneq case a, a3
      jmpneq case2 b, b3
      jmp state3
      ....
```

### CODE REVIEW

**Comment by hesingh** [comment]
Why don't you first use the SingleArgumentSelect midend pass in dpdk midend.cpp and then change the dpdk backend like https://github.com/p4lang/p4c/pull/2827 has done?

**Comment by mbudiu-vmw** [comment]
If dpdk is limited to 64 bits per value this implementation is more general

File: backends/dpdk/dpdkProgram.cpp [hunk 1]
```diff
@@ -187,6 +187,75 @@ IR::Declaration_Variable *ConvertToDpdkParser::addNewTmpVarToMetadata(cstring na
...
+   expressions. When the keyset is Mask expression "a &&& b", it inserts temporary variables in
+   Metadata structure and populates the left and right operands of comparison with
+   "input & b" and "a & b" */
+void ConvertToDpdkParser::getCondVars(const IR::Expression *sv, const IR::Expression *ce,
+                                      IR::Expression **leftExpr, IR::Expression **rightExpr) {
+    if (auto maskexpr = ce->to<IR::Mask>()) {
+        auto left = maskexpr->left;
+        auto right = maskexpr->right;
+        /* Dpdk architecture requires that both operands of &&& be constants */
+        if (!left->is<IR::Constant>() || !left->is<IR::Constant>()) {
```
mbudiu-vmw, line 202: You are using the same expression twice.

backends/dpdk/dpdkProgram.cpp
```diff
@@ -187,6 +187,75 @@ IR::Declaration_Variable *ConvertToDpdkParser::addNewTmpVarToMetadata(cstring na
...
+void ConvertToDpdkParser::handleTupleExpression(const IR::ListExpression *cl,
+                                                const IR::ListExpression *input, int inputSize,
+                                                cstring trueLabel, cstring falseLabel) {
+    IR::Expression *left;
+    IR::Expression *right;
+    /* Compare each element in the input tuple with the keyset tuple */
+    for (auto i = 0; i < inputSize; i++) {
+        auto switch_var = input->components.at(i);
+        auto caseExpr = cl->components.at(i);
+        if (!caseExpr->is<IR::DefaultExpression>()) {
```
mbudiu-vmw, line 251: A simpler solution may be to treat each DefaultExpression like 0 &&& 0.  See the SingleElementSelect pass too.

mbudiu-vmw (changes requested):

backends/dpdk/dpdkProgram.cpp
usha1830: > If dpdk is limited to 64 bits per value this implementation is more general

Yes, header and metadata field width is limited to 64-bit and immediate values width is limited to  32-bit.

backends/dpdk/dpdkProgram.cpp
usha1830: Thanks for catching this. When I tried hitting this error with a testcase, I see this condition is already checked by midend in DoSimplifySelectCases pass. So will remove this error check altogether.

backends/dpdk/dpdkProgram.cpp
usha1830: > A simpler solution may be to treat each DefaultExpression like 0 &&& 0. See the SingleElementSelect pass too.

I am not really sure how converting default expr to 0 &&& 0 would be any better than current code. This code would generate a jmpneq for 0 &&& 0 (which would be false always and the jmpneq would I think be optimized away later) but the current approach would not generate any such jmpneq. Am I missing something here?

mbudiu-vmw: In the implementation mentioned the total width of the select expression is limited to whatever the code generator supports.

usha1830:
> Why don't you first use the SingleArgumentSelect midend pass in dpdk midend.cpp and then change the dpdk backend like #2827 has done?

I don't see any benefit in using the SingleArgumentSelect pass as there is no corresponding assembly instruction for concat operator and we will have to ultimately convert the case value into series of jmpneqs.

backends/dpdk/dpdkProgram.cpp
mbudiu-vmw: It's not better but there is one fewer case to handle.

usha1830: I think we can give an error till we do this transformation.

backends/dpdk/dpdkProgram.cpp
```diff
@@ -187,6 +187,79 @@ IR::Declaration_Variable *ConvertToDpdkParser::addNewTmpVarToMetadata(cstring na
...
+/* This is a helper function for handling the transition select statement. It populates the left
+   and right operands of comparison, to be used  in conditional jump instructions. When the keyset
+   of select case is simple expressions, it populates the left and rigt operands with the input
+   expressions. When the keyset is Mask expression "a &&& b", it inserts temporary variables in
+   Metadata structure and populates the left and right operands of comparison with
+   "input & b" and "a & b" */
+void ConvertToDpdkParser::getCondVars(const IR::Expression *sv, const IR::Expression *ce,
+                                      IR::Expression **leftExpr, IR::Expression **rightExpr) {
+    if (sv->is<IR::Constant>() && sv->type->width_bits() > 32) {
+        ::error(ErrorType::ERR_UNEXPECTED,
```
mbudiu-vmw, line 199: This is a back end constraint so perhaps you should use one of the unsupported errors.

backends/dpdk/dpdkProgram.cpp
```diff
@@ -240,41 +313,59 @@ bool ConvertToDpdkParser::preorder(const IR::P4Parser *p) {
...
+                    }
+                } else {
+                    auto tupleInputExpr = e->select;
+                    auto inputSize = tupleInputExpr->components.size();
+                    auto count = 0;
+                    cstring trueLabel, falseLabel;
+                    /* For each select case, emit the block start label and then a series of
+                       jmp instructions. */
+                    for (auto sc : caseList) {
+                        cstring labelName = state->name + "_" + std::to_string(count++);
```
mbudiu-vmw, line 355: This looks like it could generate collisions if some states already have such names. I would always recommend using a name generator, but you have to use it everywhere so it keeps track of all used symbols.

mbudiu-vmw (approved):
I will approve, but please consider using a name generator. Maybe you want to file an issue about it.

backends/dpdk/dpdkProgram.cpp
usha1830: Updated

backends/dpdk/dpdkProgram.cpp
usha1830: I have updated the code to use refmap->newName() for generating the case label.

usha1830:
@mbudiu-vmw I have updated the code as per your comments. If there are no further comments, could you please merge this PR?

PR #2837, opened 2021-07-03
Remove unnecessary p4include/psa.p4

jafingerhut:
It is no longer needed now that p4include/bmv2/psa.p4 and
p4include/dpdk/psa.p4 have been created, and each of those should be
used for their respective targets.


mbudiu-vmw (approved):

jafingerhut:
The failing test runs show that p4test uses the p4include/psa.p4 include file, which I should have anticipated, but did not realize until seeing these failed test runs.

@hanw What concerns me more are the failing tests involving p4c-dpdk commands.  Do those make sense to you?  I would have thought that p4c-dpdk with your recent changes should always find p4include/dpdk/psa.p4, and not need p4include/psa.p4 to exist.

rst0git:
@jafingerhut I think we should also update `test/gtest/helpers.cpp`:
https://github.com/p4lang/p4c/blob/b3679a5e0f04b443659299540da064f342866c45/test/gtest/helpers.cpp#L132
and `backends/bmv2/CMakeLists.txt`
https://github.com/p4lang/p4c/blob/b3679a5e0f04b443659299540da064f342866c45/backends/bmv2/CMakeLists.txt#L141

rst0git (approved):

PR #2842, opened 2021-07-07
Missed optimization: Struct-assignment in action prevents ConstantFolding (#2249)

### LINKED ISSUE #2249: Missed optimization: Struct-assignment in action prevents ConstantFolding
I am filing this per request.

Consider the following program snippet:

```p4
control ingress(inout Headers hdr, inout md meta, inout standard_metadata_t stdmeta) {
    simple_struct test = { 128w0 };
    action pointless_action() {
        test = test;
    }

    apply {
        pointless_action();
        if ((test.a != 1) && (test.a <= 1)) {
            hdr.eth_hdr.eth_type = 16w1;
        }
    }
}
```

ConstantFolding should normally reduce this to something like:

```p4
control ingress(inout Headers hdr, inout md meta, inout standard_metadata_t stdmeta) {
    simple_struct test = { 128w0 };
    action pointless_action() {
        test = test;
    }

    apply {
        pointless_action();
        hdr.eth_hdr.eth_type = 16w1;
    }
}
```

However, likely because of the assignment in the action call, ConstantFolding is skipped and the program after the midend looks like this:

```p4
control ingress(inout Headers hdr, inout md meta, inout standard_metadata_t stdmeta) {
    simple_struct test_0;
    @name("ingress.pointless_action") action pointless_action() {
    }
    apply {
        test_0.a = 128w0;
        pointless_action();
        if (test_0.a != 128w1 && test_0.a <= 128w1) {
            hdr.eth_hdr.eth_type = 16w1;
        }
    }
}
```

 If the method call `pointless_action();` is removed the compiler optimizes the expressions as expected.
Note that I do not consider this a bug but a missing optimization.

[missed_constant_folding.p4.txt](https://github.com/p4lang/p4c/files/4361238/missed_constant_folding.p4.txt)

kostasyrmia:
Fix #2249 
This optimizes _LocalCopyPropagation Pass_ in case there is an assignment in the body of the action, where the left side and the right side of that assignment are equivalent, by not updating the writes set of that action.
As a consequence of this change, the test _issue-2147-1.p4_ failed, so I updated it as well.

### DIFF
--- a/midend/local_copyprop.cpp
+++ b/midend/local_copyprop.cpp
@@ -245,8 +245,8 @@ const IR::Expression *DoLocalCopyPropagation::copyprop_name(cstring name) {
     LOG6("  copyprop_name(" << name << ")" << (isWrite() ? " (write)" : ""));
     if (isWrite()) {
         dropValuesUsing(name);
-        if (inferForFunc)
-            inferForFunc->writes.insert(name);
+        if (inferForFunc) {
+            inferForFunc->is_first_write_insert = inferForFunc->writes.insert(name).second; }
         if (isRead() || findContext<IR::MethodCallExpression>()) {
             /* If this is being used as an 'out' param of a method call, its not really
              * read, but we can't dead-code eliminate it without eliminating the entire
@@ -322,6 +322,9 @@ IR::AssignmentStatement *DoLocalCopyPropagation::preorder(IR::AssignmentStatemen
 IR::AssignmentStatement *DoLocalCopyPropagation::postorder(IR::AssignmentStatement *as) {
     if (as->left->equiv(*as->right)) {
         LOG3("  removing noop assignment " << *as);
+        if (inferForFunc && inferForFunc->is_first_write_insert) {
+            inferForFunc->writes.erase(expr_name(as->left));
+            inferForFunc->is_first_write_insert = false; }
         return nullptr; }
     // FIXME -- if as->right is an uninitialized value, we could legally eliminate this
     // assignment, which would simplify (and minimize) the code.  This could be a separate
--- a/testdata/p4_16_samples_outputs/issue2147-1-midend.p4
+++ b/testdata/p4_16_samples_outputs/issue2147-1-midend.p4
@@ -5,9 +5,6 @@ control ingress(inout bit<8> h) {
     @hidden action issue21471l5() {
         tmp_0 = h;
     }
-    @hidden action issue21471l7() {
-        h = tmp_0;
-    }
     @hidden table tbl_issue21471l5 {
         actions = {
             issue21471l5();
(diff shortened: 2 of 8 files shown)


kostasyrmia:
@mbudiu-vmw, could you do a review, please?

mbudiu-vmw:
@ChrisDodd has written this pass, so I have requested a review from him.
He has been busy lately, but this is a small change, and hopefully he will review it.

kostasyrmia:
@ChrisDodd 
Just a reminder to take a look at this in case you have forgotten. I am sorry to bother you if you are still busy.

kostasyrmia:
@mbudiu-vmw @ChrisDodd 
Any updates?

midend/local_copyprop.cpp
```diff
@@ -322,6 +322,9 @@ IR::AssignmentStatement *DoLocalCopyPropagation::preorder(IR::AssignmentStatemen
 IR::AssignmentStatement *DoLocalCopyPropagation::postorder(IR::AssignmentStatement *as) {
     if (as->left->equiv(*as->right)) {
         LOG3("  removing noop assignment " << *as);
+        if (inferForFunc && inferForFunc->is_first_write_insert) {
```
mbudiu-vmw, line 325: So why is this only done for the first write?
Can you also please add a comment to the is_first_write_insert field definition below documenting its meaning?

kostasyrmia: Here the self-assignment is removed, but the expression on its left side still remains in the writes set, even though this assignment does nothing. So, we need to know whether that expression was in the set before this assignment (from some other assignments) or we are the first who put it there. If we are the first, then we need to remove it too.

I know this might not seem so efficient (removing a just inserted expression), but I suppose that in a 'reasonable' program the self-assignment should occur rarely (practically never). And constantly writing to the bool should be negligible overhead for this improvement.

testdata/p4_16_samples_outputs/issue2147-1-midend.p4
```diff
@@ -20,16 +17,9 @@ control ingress(inout bit<8> h) {
...
-    @hidden table tbl_issue21471l7 {
-        actions = {
-            issue21471l7();
-        }
-        const default_action = issue21471l7();
-    }
     apply {
         tbl_issue21471l5.apply();
         tbl_a.apply();
-        tbl_issue21471l7.apply();
```
kostasyrmia, line 32: Here is a similar situation. Even though the body of the action a was empty, the compiler added two assignments:

    b = tmp;
    tmp = b;

The writes set of that action had both tmp and b, even though the second assignment was removed (after propagation), so the propagation after the action call wasn't possible.

mbudiu-vmw (commented):
@ChrisDodd any chance you could review this 3-line patch?

mbudiu-vmw:
I am not entirely convinced that there are many opportunities for this particular optimization.
Let's merge it, though.

mbudiu-vmw (approved):

PR #2845, opened 2021-07-09
DPDK backend: Added error checks for limitations imposed by DPDK target on table keys. Convert table such that all match keys come from same struct

usha1830:
This PR adds error checks for the following limitations on table keys as imposed by DPDK target:
    - Only one LPM match field allowed per table.
     - Maximum allowed key size of header/metadata field is 64 bits.
     - If there is a key field with lpm match kind, the other match fields, if any, must all be exact match.
     - All the match fields should be part of the same header/metadata struct. Added a pass to transform the program to adhere  to this target limitation.


jfingerh:
All of the restrictions you mention make sense to me, except this one sounds confusing:

"All the match fields should be part of the same header/metadata struct."

Does DPDK support parsing, say, an IPv4 header, followed by a TCP header, and having a single table with some match fields from the IPv4 header, and other match fields from the TCP header?

If it does support that, then perhaps you can change the description of the restriction to be more clear?

usha1830:
>>Does DPDK support parsing, say, an IPv4 header, followed by a TCP header, and having a single table with some match fields from the IPv4 header, and other match fields from the TCP header?

DPDK supports parsing of headers one after the other but does not allow a single table to have match fields from different headers. All the fields in a table should be part of the same header.

jfingerh:
@hanw Can you confirm the restriction mentioned in the previous comment?  This seems very restrictive, and a restriction that should be removed fairly soon.

usha1830:
> @hanw Can you confirm the restriction mentioned in the previous comment? This seems very restrictive, and a restriction that should be removed fairly soon.

I have checked it by loading a spec file into DPDK pipeline application. It does not accept a table violating this condition. In any case, we can check with Cristian on this restriction.

mbudiu-vmw (approved):

usha1830:
> > @hanw Can you confirm the restriction mentioned in the previous comment? This seems very restrictive, and a restriction that should be removed fairly soon.
> 
> I have checked it by loading a spec file into DPDK pipeline application. It does not accept a table violating this condition. In any case, we can check with Cristian on this restriction.

For performance reasons, the DPDK target requires all the match fields of a given table to be in the same structure (same header instance or in meta-data), in this case the compiler can create mirror copies for the match fields in meta-data.
I have added a new pass to transform the P4 program to match this criteria.

backends/dpdk/dpdkArch.cpp
```diff
@@ -930,4 +930,129 @@ const IR::Node *PrependPDotToActionArgs::preorder(IR::PathExpression *path) {
     return path;
 }
 
+/* This function updates the control block with newly created declarations and
+   statements for table match keys */
+const IR::Node *copyMatchKeysToSingleStruct::postorder(IR::P4Control *c) {
```
mbudiu-vmw, line 935: The convention is to use uppercase names for classes.

backends/dpdk/dpdkArch.cpp
```diff
@@ -930,4 +930,129 @@ const IR::Node *PrependPDotToActionArgs::preorder(IR::PathExpression *path) {
...
+       ...
+       table tbl {
+           key = {
+               tbl_ethernet_srcAddr: lpm;
+               tbl_ipv4_totalLen   : exact;
+           }
+	   ...
+       }
+       apply {
+           tbl_ethernet_srcAddr = h.ethernet.srcAddr;
```
mbudiu-vmw, line 985: this kind of change is not trivial, since the table apply call may be part of an expression.
Most such expressions may be normalized by prior midend passes, but the default midends we use tend to leave alone table.apply calls within switch and if statements. The ones in switch statements in particular cannot be moved out.

mbudiu-vmw: Please note that this problem is tackled by the KeysSideEffect pass. You may want to reuse the code there if you want to insert copy statements in the right places. The code there detects where the insertions must be performed.

mbudiu-vmw: Please add a test case where a table apply is within an if or switch.

mbudiu-vmw (changes requested):

backends/dpdk/DpdkXfail.cmake
```diff
@@ -65,3 +65,18 @@ p4c_add_xfail_reason("dpdk"
...
+
+p4c_add_xfail_reason("dpdk"
+  "Non 'exact' match kind not permitted"
+  testdata/p4_16_samples/psa-dpdk-lpm-match-err2.p4
+  )
+
+p4c_add_xfail_reason("dpdk"
+  "Key field wider than 64-bit is not permitted"
+  testdata/p4_16_samples/psa-dpdk-lpm-match-err3.p4
+  )
```
jfingerh, line 82: If the message "key field wider than 64-bit is not permitted" is an accurate description of current DPDK back end restrictions, does that mean that if someone wants to do a match on an IPv6 address, they must break it up into 2 separate 64-bit key fields, e.g. hdr.ipv6.dstAddress[127:64] and hdr.ipv6.dstAddress[63:0] ?

backends/dpdk/dpdkArch.cpp
jfingerh: If there is not already such a test case, I would also recommend adding a test case where an earlier table has an action that can _modify_ one of the header fields of a packet, that is then later matched as a table key field, and verify that any transformation you do modifies not only the original header field in hdr.ipv4.dstAddress (for example), but also in any copies that your transformation creates.

hanw:
I suspect it is not sufficient to simply create a copy of the header field in metadata struct to satisfy the constraint in dpdk about have all keys in the same buffer. Cristian suggested that 

```
table classification_table {
     match {
          h.ipv4.dst_addr exact
          h.udp.dst_port exact
     }
}

… should get translated to something similar to …

struct metadata_t {
      bit<32> classification_table_ipv4_dst_addr
      bit<16> classification_table_udp_dst_port
}

metadata instanceof metadata_t

table classification table {
    match {
         m.classification_table_ipv4_dst_addr exact
         m.classification_table_udp_dst_port exact
    }
}

apply {
      extract h.ethernet
      extract h.ipv4
      extract h.udp
      mov  m.classification_table_ipv4_dst_addr h.ipv4.dst_addr
      mov m.classification_table_udp_dst_port h.udp.dst_port
}
```

Do we need to ensure the fields are contiguous in the metadata struct in order to be allocated to the same 'buffer' in dpdk? It's better to have Cristian explain in terms of actual C implementation in DPDK about where this constraint is from. Otherwise, this constraint is very mysterious to me..

cristian-dumitrescu:
1. In order for the mirror fields in packet meta-data to have a unique name, we need to prefix the field name with the table name and also include the full match field name. Example: h.ipv4.dst_addr match field for table classification needs to be mirrored by the m.table_classification_h_ipv4_dst_addr meta-data field.

2. Internally in the back-end, the set of match fields for a specific table get translated to the following:
a) a buffer: the array of bytes representing the header or metadata structure
b) an offset into the above buffer: offset of the "left-most" field within the buffer
c) a key size: the number of bytes read from the buffer, covering all the match fields from the "left-most" to the "right-most", including any unneeded fields in between.
d) a key mask (of key size bytes): indicates which of the key size bytes are relevant (part of the table match key) and which not (other fields in the same struct that are not included in the table match field list).
In order to keep the key size as short as possible, we need to put all meta-data fields that mirror the same table contiguously one after the other. For example, a 5-tuple of 16 bytes results in comparing keys of 16 bytes for every packet if those 5 fields are contiguous in meta-data or in comparing keys of 64 bytes per packet if for whatever reason the compiler inserts some other fields of 48 bytes in total in between the 5 fields of the tuple. So please make sure the mirror fields are contiguous.

jfingerh:
@cristian-dumitrescu So does this mean that it might be fairly common that if a packet header field is used as a key in 5 tables ,there will often be 5 copies of that field that you are calling "mirror fields"?  If so, that makes sense.  It isn't exactly avoiding copies that I can see, but it should only be making a run-time copy of a field once for each time it is used as a table key field (or perhaps less, if multiple tables happen to have exactly the same set of key fields, but that would be the exception, not the rule, in typical P4 programs).

Since a field like hdr.ipv4.dstAddr might be used as one of the table key fields for 3 different tables, and because there can be an arbitrary number of actions before or between any of those 3 tables that might assign a value to hdr.ipv4.dstAddr, I would recommend that when using this technique, the copy of the hdr.ipv4.dstAddr to each of the 3 mirror fields for the 3 tables should be made immediately before the table is applied.  That ensures that the mirror field's value is equal to the current value of hdr.ipv4.dstAddr, not a perhaps old value that it had earlier in the P4 program's execution.

FYI, given that in P4 "mirror" is often used as a synonym for packet cloning, it might be nice to use a different name than "mirror fields" for this implementation technique, but that is really up to you.

cristian-dumitrescu:
@jfingerh
1. It is about avoiding copies, but only if possible. If not possible, we do the copy.

2. I expect a scenario where the same header field is used in 5 different tables to be very infrequent. But yes, if the same header field is used by multiple tables, we should have one copy of it for each table, at least at this point. Later on, maybe the compiler can use more sophisticated logic in order to take the decision on whether to use different copies for each table (smaller lookup key) or use the same copy for multiple tables (potentially larger lookup keys due to "holes" in the middle),

3. Yes, you are absolutely right (thank you!): we need to read the header field into its meta-data copy just before the table lookup takes place, in order to avoid missing any updates of the header field. @usha1830 @hanw Please do the copy at this time. Maybe later on the compiler could decide to pull in the moment of the copy based on some dependency analysis, but absolutely not essential for now.

backends/dpdk/dpdkArch.cpp
usha1830: @mbudiu-vmw Thanks for pointing to the KeySideEffect pass. I will update this PR with some code reused from this pass.

usha1830: Will add a few tests as well for covering the above mentioned scenarios.

backends/dpdk/DpdkXfail.cmake
usha1830: Yes. Currently, we are only checking this and giving an error. But we do plan to remove this restriction soon by splitting such fields into fields with width within 64-bit.

usha1830:
Thanks all for this discussion and clarifying on the requirements here.

hanw:
@cristian-dumitrescu I still don't understand why the copying of key to metadata has to be done in the compiler. Is it to simplify the implementation of the dpdk backend? or is it to provide a chance to optimize (by saving some copy sometimes)?

Why cannot the dpdk backend internally create a struct and perform the copy before doing a table match? That would take less 'assembly' instructions and better performance?

cristian-dumitrescu:
@hanw
1. There are many cases when all the match fields are part of the same header or are all meta-data fields. We try to avoid the copy whenever possible.
2. It would be a significant change in the table layer of the back-end to do what you suggest, and it would also not be the right choice for performance. Why would the back-end copy all the match fields for every table for every packet if not necessary? It is relatively trivial for the compiler to detect when the copy is needed and insert the right instructions as opposed to the back-end always doing the copy at run-time (bad) or spending cycles as run-time to decide whether the copy is required or not (even worse).

backends/dpdk/dpdkProgram.cpp
```diff
@@ -483,6 +523,7 @@ boost::optional<int> ConvertToDpdkControl::getNumberFromProperty(const IR::P4Tab
 
 
 bool ConvertToDpdkControl::preorder(const IR::P4Table *t) {
+    checkTableValid(t);
```
mbudiu-vmw, line 526: Are you OK not returning here on error?
You could have checkTableValid return a bool.

backends/dpdk/dpdkProgram.h
```diff
@@ -43,6 +43,8 @@ limitations under the License.
 #include "lib/json.h"
 namespace DPDK {
 
+#define DPDK_MAX_HEADER_METADATA_FIELD_SIZE 64
```
mbudiu-vmw, line 46: can you please document the meaning of this value?

mbudiu-vmw (approved):

backends/dpdk/dpdkProgram.cpp
usha1830: Updated checkTableValid to return bool

usha1830:
@mbudiu-vmw Could you please merge this if the changes look fine now.

PR #2846, opened 2021-07-10
dpdk: Add reference to pipeline user guide

rst0git:
This PR adds a reference to the [Pipeline Application User Guide](https://doc.dpdk.org/guides/sample_app_ug/pipeline.html) in the DPDK backend README file.


mbudiu-vmw (approved):
Why would this fail the tests?

rst0git:
> Why would this fail the tests?

Perhaps there has been a glitch and if we restart the MacOS tests they would likely pass.

PR #2847, opened 2021-07-11
dpdk: add support for action selector

hanw:


backends/dpdk/dpdkArch.cpp
```diff
@@ -930,4 +934,201 @@ const IR::Node *PrependPDotToActionArgs::preorder(IR::PathExpression *path) {
...
+const IR::Node* SplitActionSelectorTable::postorder(IR::P4Table* tbl) {
+    bool isConstructedInPlace = false;
+    auto instance = Helpers::getExternInstanceFromProperty(tbl, "psa_implementation",
+                                                           refMap, typeMap, &isConstructedInPlace);
+    if (!instance)
+        return tbl;
+    if (instance->type->name != "ActionSelector")
+        return tbl;
+
+    if (!instance->arguments->at(1)->expression->is<IR::Constant>())
```
mbudiu-vmw, line 987: maybe first you should validate that this many arguments exist

backends/dpdk/dpdkArch.cpp
```diff
@@ -930,4 +934,201 @@ const IR::Node *PrependPDotToActionArgs::preorder(IR::PathExpression *path) {
...
+        ::error(ErrorType::ERR_UNEXPECTED,
+                "The 'size' argument of ActionSelector %1% must be a constant", *instance->name);
```
mbudiu-vmw, line 989: no return on error on any of these cases?

backends/dpdk/dpdkArch.h
```diff
@@ -596,6 +624,7 @@ class RewriteToDpdkArch : public PassManager {
         auto *evaluator = new P4::EvaluatorPass(refMap, typeMap);
         auto *parsePsa = new ParsePsa();
         info = new CollectMetadataHeaderInfo(&parsePsa->toBlockInfo);
+        passes.push_back(new ConvertActionSelector(refMap, typeMap));
```
mbudiu-vmw, line 627: will this work no matter where the table is called?
e.g., what if it is called in an if or switch statement?

backends/dpdk/dpdkProgram.h
```diff
@@ -141,7 +144,11 @@ class ConvertToDpdkControl : public Inspector {
 
     void add_inst(const IR::DpdkAsmStatement *s) { instructions.push_back(s); }
     void add_table(const IR::DpdkTable *t) { tables.push_back(t); }
+    void add_table(const IR::DpdkSelector *s) { selectors.push_back(s); }
     void add_action(const IR::DpdkAction *a) { actions.push_back(a); }
+
+	boost::optional<cstring> getIdFromProperty(const IR::P4Table*, cstring);
```
mbudiu-vmw, line 150: indentation?

mbudiu-vmw (changes requested):

backends/dpdk/dpdkArch.h
hanw: Good point, in this case, since action selector table is converted into three p4 table (routing, group, member), and applied in sequence. We could transform the apply call in if or switch statement to use the apply call on the third table.

backends/dpdk/dpdk.def
```diff
@@ -47,13 +47,27 @@ class DpdkTable {
...
+}
+
 // Toplevel Assembly representation
 class DpdkAsmProgram {
     inline IndexedVector<DpdkHeaderType> headerType;
     inline IndexedVector<DpdkStructType> structType;
     inline IndexedVector<DpdkExternDeclaration> externDeclarations;
     inline IndexedVector<DpdkAction> actions;
     inline IndexedVector<DpdkTable> tables;
+	inline IndexedVector<DpdkSelector> selectors;
```
mbudiu-vmw, line 70: indentation?

backends/dpdk/dpdkArch.cpp
```diff
@@ -930,4 +935,370 @@ const IR::Node *PrependPDotToActionArgs::preorder(IR::PathExpression *path) {
...
+    member_keys.push_back(tableKeyEl);
+    IR::IndexedVector<IR::Property> member_properties;
+    member_properties.push_back(new IR::Property("key", new IR::Key(member_keys), false));
+
+    IR::IndexedVector<IR::ActionListElement> memberActionList;
+    for (auto action : tbl->getActionList()->actionList)
+        memberActionList.push_back(action);
+    member_properties.push_back(new IR::Property("actions",
+                                                 new IR::ActionList(memberActionList), false));
+    if (tbl->getSizeProperty()) {
```
mbudiu-vmw, line 1098: you have done this already above, you could save the result.

mbudiu-vmw (approved):

PR #2849, opened 2021-07-13
Bazel rules for the graphs backend.

maemre:
This PR adds
- a target (`:p4c_graphs`) to build the compiler binary with the graphs
  backend, and
- a build rule `p4_graphs` to invoke this backend and merge the graph
  files

so that the graphs backend can be used in other projects using Bazel.

We need a new build rule (rather than specifying a different binary for
`_p4c` in the `p4_library` rule) because the graphs backend does
not use the normal output file arguments and uses a separate argument to
specify the output *directory* to write the graphs in. The names of the
output files depend on the content of the input p4 file, and blaze
outputs cannot be directories (they can be file groups but we do not
know the exact file names without effectively running the p4 compiler).
So, we cannot directly use the output files in a blaze target. Instead,
this build rule concatenates the output graphs into a single GraphViz
file. The output file can later be processed by `gvpack` and `dot` to
produce a single large graph (the original graphs are embedded as
subgraphs). It can also be parsed by other tools to extract specific
graphs.

### DIFF
--- a/bazel/p4_library.bzl
+++ b/bazel/p4_library.bzl
@@ -120,3 +120,103 @@ p4_library = rule(
     incompatible_use_toolchain_transition = True,
     toolchains = ["@bazel_tools//tools/cpp:toolchain_type"],
 )
+
+def _p4_graphs_impl(ctx):
+    p4c = ctx.executable._p4c
+    p4file = ctx.file.src
+    p4deps = ctx.files._p4include + ctx.files.deps
+    output_file = ctx.outputs.out_file
+    graph_dir = output_file.path + "-graphs-dir"
+    args = [
+        p4file.path,
+        "--std",
+        ctx.attr.std,
+        "--arch",
+        ctx.attr.arch,
+        "--graphs-dir",
+        graph_dir,
+    ]
+    if ctx.attr.extra_args:
+        args.append(ctx.attr.extra_args)
+
+    include_dirs = {d.dirname: 0 for d in p4deps}  # Use dict to express set.
+    include_dirs["."] = 0  # Enable include paths relative to workspace root.
+    args += [("-I" + dir) for dir in include_dirs.keys()]
+
+    cpp_toolchain = find_cpp_toolchain(ctx)
+    ctx.actions.run_shell(
+        command = """
+            # p4c invokes cc for preprocessing; we provide it below.
+            function cc () {{ "{cc}" "$@"; }}
+            export -f cc
+            # Create the output directory
+            mkdir "{graph_dir}"
+            # Run the compiler
+            "{p4c}" {p4c_args}
+            # Merge all output graphs
+            cat "{graph_dir}"/* > {output_file}
+        """.format(
+            cc = cpp_toolchain.compiler_executable,
+            p4c = p4c.path,
+            p4c_args = " ".join(args),
+            graph_dir = graph_dir,
+            output_file = output_file.path,
+        ),
+        inputs = p4deps + [p4file],
+        tools = depset(
+            direct = [p4c],
+            transitive = [cpp_toolchain.all_files],
+        ),
+        outputs = [output_file],
+        progress_message = "Generating the graphs for P4 program %s" % p4file.short_path,
+        use_default_shell_env = True,
+    )
+
+p4_graphs = rule(
+    doc = "Generates the graphs for a P4 program using p4c-graphs, and merges them in a single GraphViz file. This file can be later processed by gvpack and dot.",
+    implementation = _p4_graphs_impl,
+    attrs = {
+        "src": attr.label(
+            doc = "P4 source file to pass to p4c.",
+            mandatory = True,
+            allow_single_file = [".p4"],
+        ),
+        "deps": attr.label_list(
+            doc = "Additional P4 dependencies (optional). Use for #include-ed files.",
+            mandatory = False,
+            allow_files = [".p4", ".h"],
+            default = [],
+        ),
+        "out_file": attr.output(
+            doc = "The name of the output DOT file",
+            mandatory = True,
+        ),
+        "arch": attr.string(
+            doc = "The --arch argument passed to p4c (default: v1model).",
+            mandatory = False,
+            default = "v1model",
+        ),
+        "std": attr.string(
+            doc = "The --std argument passed to p4c (default: p4-16).",
+            mandatory = False,
+            default = "p4-16",
+        ),
+        "extra_args": attr.string(
+            doc = "String of additional command line arguments to pass to p4c.",
+            mandatory = False,
+            default = "",
+        ),
+        "_p4c": attr.label(
+            default = Label("@com_github_p4lang_p4c//:p4c_bmv2"),
+            executable = True,
+            cfg = "target",
+        ),
+        "_p4include": attr.label(
+            default = Label("@com_github_p4lang_p4c//:p4include"),
+            allow_files = [".p4", ".h"],
+        ),
+        "_cc_toolchain": attr.label(default = Label("@bazel_tools//tools/cpp:current_cc_toolchain")),
+    },
+    incompatible_use_toolchain_transition = True,
+    toolchains = ["@bazel_tools//tools/cpp:toolchain_type"],
+)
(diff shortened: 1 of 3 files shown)


maemre:
@smolkaj can you review this PR?

bazel/p4_library.bzl
```diff
@@ -120,3 +120,103 @@ p4_library = rule(
     incompatible_use_toolchain_transition = True,
     toolchains = ["@bazel_tools//tools/cpp:toolchain_type"],
 )
+
+def _p4_graphs_impl(ctx):
```
smolkaj, line 124: It looks like this is duplicating a lot of code from _p4_library_impl. Could we reuse the code somehow?

bazel/p4_library.bzl
```diff
@@ -120,3 +120,103 @@ p4_library = rule(
...
+            mandatory = True,
+            allow_single_file = [".p4"],
+        ),
+        "deps": attr.label_list(
+            doc = "Additional P4 dependencies (optional). Use for #include-ed files.",
+            mandatory = False,
+            allow_files = [".p4", ".h"],
+            default = [],
+        ),
+        "out_file": attr.output(
```
smolkaj, line 190: Nit: just "out" would be a bit more idiomatic.

bazel/p4_library.bzl
```diff
@@ -120,3 +120,103 @@ p4_library = rule(
...
+            doc = "The name of the output DOT file",
```
smolkaj, line 191: Nit: a check that the file actually ends in ".dot" would make this rule easier to use.
You can use `str.endswith(str) -> bool` and `fail(str)`.

smolkaj (changes requested):
Looks good to me at a high level.

Could you also extend bazel/example with the new rule? This will make it both easier to use for others. It will also cause the rule to be executed as part of continuous integration, so it will serve as a basic safety net ensuring the rule is at least not completely broken.

bazel/p4_library.bzl
```diff
@@ -120,3 +120,103 @@ p4_library = rule(
...
+        command = """
+            # p4c invokes cc for preprocessing; we provide it below.
+            function cc () {{ "{cc}" "$@"; }}
+            export -f cc
+            # Create the output directory
+            mkdir "{graph_dir}"
+            # Run the compiler
+            "{p4c}" {p4c_args}
+            # Merge all output graphs
+            cat "{graph_dir}"/* > {output_file}
```
smolkaj, line 157: Nit: "{graph_dir}"/* -> "{graph_dir}/*"
(unless this brakes the * somehow)

bazel/p4_library.bzl
maemre: Done.

bazel/p4_library.bzl
maemre: Done.

bazel/p4_library.bzl
maemre: That breaks the shell globbing by *, I added a comment to explain it.

bazel/p4_library.bzl
maemre: I added two helpers to reuse argument extraction and shell/cc environment setup

maemre:
@smolkaj I made the changes pointed in the code review. Can you review it again?

The two big changes are:

- I split off two common functionality to separate functions: argument extraction, and running commands in the required shell environment.
- I added a rule to `bazel/example/p4/BUILD` to show how to use the `p4_graphs` rule and to test it.

bazel/p4_library.bzl
```diff
@@ -2,26 +2,62 @@
...
-        args.append(ctx.attr.extra_args)
 
     include_dirs = {d.dirname: 0 for d in p4deps}  # Use dict to express set.
     include_dirs["."] = 0  # Enable include paths relative to workspace root.
     args += [("-I" + dir) for dir in include_dirs.keys()]
+    inputs = p4deps + [p4file]
+
+    return (args, inputs)
+
+def _run_p4c_with_cc(ctx, p4c, command, **run_shell_kwargs):
```
smolkaj, line 24: How about simplifying this to "_run_shell_cmd_with_p4c(ctx, command)"? In particular:
- The function can assert that `ctx.executable._p4c` is defined.
- No need to mention cc, that's arguably an implementation detail.
- We can add `run_shell_kwargs` if we need it in the future, currently it is unused.

bazel/p4_library.bzl
```diff
@@ -2,26 +2,62 @@
 
 load("@bazel_tools//tools/cpp:toolchain_utils.bzl", "find_cpp_toolchain")
 
-def _p4_library_impl(ctx):
-    p4c = ctx.executable._p4c
+def _extract_common_p4c_args(ctx):
```
smolkaj, line 5: Could we make this function only do one thing, i.e. extract the common p4c args, as the name suggests?
I think that would make things easier to understand, and extracting the inputs is just a one-liner that we can either duplicate or make a separate function.

bazel/p4_library.bzl
smolkaj: Thanks for adding the coment!

smolkaj (changes requested):
Thanks for the changes. Comments inline.

bazel/p4_library.bzl
maemre: The build rule implementations pass their inputs, outputs, and progress messages via `run_shell_kwargs`. I can add them as explicit parameters instead, if that makes sense.

I simplifed it with other changes. I used `if` and `fail` because Starlark does not have assertions (outside unit tests, which do not fail immediately) as far as I am aware of.

bazel/p4_library.bzl
maemre: I separated extracting the inputs to a helper function `_extract_p4c_inputs` that is called by the build rule implementations

bazel/p4_library.bzl
```diff
@@ -2,27 +2,73 @@
...
 
     include_dirs = {d.dirname: 0 for d in p4deps}  # Use dict to express set.
     include_dirs["."] = 0  # Enable include paths relative to workspace root.
     args += [("-I" + dir) for dir in include_dirs.keys()]
 
+    return args
+
+def _extract_p4c_inputs(ctx):
+    """Extract input p4 files to give to p4c from the build rule context."""
+    p4file = ctx.file.src
```
smolkaj, line 25: Optionally: consider making this a one-liner.

bazel/p4_library.bzl
```diff
@@ -2,27 +2,73 @@
...
+    p4deps = ctx.files._p4include + ctx.files.deps
+    return p4deps + [p4file]
+
+def _run_shell_cmd_with_p4c(ctx, command, **run_shell_kwargs):
+    """Run given sequence of shell commands using `run_shell` action after
+setting up the C compiler toolchain.
```
smolkaj, line 31: Nit: please indent.

bazel/p4_library.bzl
```diff
@@ -2,27 +2,73 @@
...
+    return args
+
+def _extract_p4c_inputs(ctx):
+    """Extract input p4 files to give to p4c from the build rule context."""
+    p4file = ctx.file.src
+    p4deps = ctx.files._p4include + ctx.files.deps
+    return p4deps + [p4file]
+
+def _run_shell_cmd_with_p4c(ctx, command, **run_shell_kwargs):
+    """Run given sequence of shell commands using `run_shell` action after
```
smolkaj, line 30: "Runs the given shell command" (since it is only a single string, rather than a list of strings)

smolkaj (approved):
Some final minor comments. Otherwise LGTM to me.

@mbudiu-vmw, could you take a look when you get a chance?

bazel/p4_library.bzl
maemre: I was wary of doing it because of readability but it looks readable as a one-liner, so I made it a one-liner.

bazel/p4_library.bzl
maemre: done

bazel/p4_library.bzl
maemre: done

maemre:
@mbudiu-vmw I just wanted to ping you, in case this PR has fallen through the cracks.

mbudiu-vmw:
I really know nothing about Bazel, so I cannot review this. But if @smolkaj approved them they are good for me.

smolkaj:
Confirming that the changes look good from the technical (Bazel) perspective.

mbudiu-vmw (approved):

Issue #2851, opened 2021-07-15
psa.p4 include issue on main branch


DusanKrdzic:
Some tests fail because of `psa.p4` include on main branch.

EXAMPLE:
![Screenshot from 2021-07-15 11-44-05](https://user-images.githubusercontent.com/77269536/125767338-b20403b9-d03e-4bcc-8709-296346de8227.png)

hanw:
This is expected, the reference output needs to be updated.

P4TEST_REPLACE=1 ctest -R p4/testdata/p4_16_samples/psa-action-selector2.p4

PR #2852, opened 2021-07-15
update psa.p4 include in output tests

DusanKrdzic:
fix : https://github.com/p4lang/p4c/issues/2851
Updated `psa.p4` include in outputs.

### DIFF
--- a/backends/dpdk/DpdkXfail.cmake
+++ b/backends/dpdk/DpdkXfail.cmake
@@ -10,7 +10,7 @@ p4c_add_xfail_reason("dpdk"
   )
 
 p4c_add_xfail_reason("dpdk"
-  "Ambiguous method execute"
+  "LHS of meter execute statement is missing"
   testdata/p4_16_samples/psa-meter1.p4
 )
 
--- a/testdata/p4_16_samples/issue1208-1.p4
+++ b/testdata/p4_16_samples/issue1208-1.p4
@@ -1,5 +1,5 @@
 #include <core.p4>
-#include <psa.p4>
+#include <bmv2/psa.p4>
 struct EMPTY { };
 
 parser MyIP(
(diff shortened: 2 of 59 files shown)


jafingerhut (approved):
Looks good to me.

DusanKrdzic:
Test `dpdk/testdata/p4_16_samples/psa-meter1.p4` also fails on main branch. This failure has nothing to do with changes in this pr.

mbudiu-vmw:
Can you figure out what changed to cause this failure?

mbudiu-vmw:
If there is a clear explanation you can just add this file to the list of xfail tests in the respective backend.

hanw:
This is caused by the removal of p4include/psa.p4 and addition of p4include/bmv2/psa.p4 and p4include/dpdk/psa.p4.

The consequence is p4test does not have a corresponding backend psa.p4 anymore. As a result, bmv2/psa.p4 was used as a default psa.p4 for p4test.

DusanKrdzic:
I'm not sure but I think it is already in xfail list.
https://github.com/p4lang/p4c/blob/e3375706258042d86beb3f287ab01880e3d9f4a4/backends/dpdk/DpdkXfail.cmake#L12-L15

mbudiu-vmw:
Indeed, but the xfail indicates that an error message containing "Ambiguous method execute" is expected, but the actual error is different.

DusanKrdzic:
> Indeed, but the xfail indicates that an error message containing "Ambiguous method execute" is expected, but the actual error is different. 

Do I need to add this test file with new error message in dpdk xfails or?

Is there anything else you want me to do in order to fix and merge this pr?

mbudiu-vmw:
I believe that changing the error message in the xfail test should be sufficient.
You should be able to just run that specific test on your machine to verify that it works.

DusanKrdzic:
I've changed the error massage. It works on my machine now.

PR #2857, opened 2021-07-21
For code blocks with '@atomic' use entire block for action synthesis

amreshk:
The `SynthesizeActions` pass tries to synthesize actions from code blocks. However, if the code block contains a pragma `@atomic` user intention is to atomically execute the entire block and this pass should honor the pragma. 

This PR adds a check for atomic pragma and uses the entire block for action synthesis.

### DIFF
--- a/midend/actionSynthesis.cpp
+++ b/midend/actionSynthesis.cpp
@@ -125,6 +125,10 @@ const IR::Node* DoSynthesizeActions::postorder(IR::P4Control* control) {
 }
 
 const IR::Node* DoSynthesizeActions::preorder(IR::BlockStatement* statement) {
+    if (statement->annotations->getSingle("atomic")) {
+        return createAction(statement);
+    }
+
     // Find a chain of statements to convert
     auto actbody = new IR::BlockStatement;  // build here new action
     auto left = new IR::BlockStatement(statement->annotations);  // leftover statements


midend/actionSynthesis.cpp
```diff
@@ -125,6 +125,10 @@ const IR::Node* DoSynthesizeActions::postorder(IR::P4Control* control) {
 }
 
 const IR::Node* DoSynthesizeActions::preorder(IR::BlockStatement* statement) {
+    if (statement->annotations->getSingle("atomic")) {
```
mbudiu-vmw, line 128: Nothing really promises that action bodies are atomic.
Perhaps the created action should also have an atomic annotation?

mbudiu-vmw: Atomic must be honored by the backend as well.

amreshk: Good point, I see that `createAction` function does end up with creating an action with a `hidden` and the action body with the `atomic` annot.

I can update the action to also have the `atomic` annotation. Would it make sense to extend this to other annotations on the block as well?

mbudiu-vmw: It would make sense to preserve the other annotations, but it is very difficult to make a blanket statement about annotations, since their semantics is not standardized (except the ones defined by the spec). Preserving them would not hurt.

amreshk: Thanks, I added a fix to copy the annotations to the newly created action.

mbudiu-vmw: The question remains why move an atomic block to an action?
Can't an action itself have inside one or more atomic blocks?
Perhaps this should be guided by a policy property.

jfingerh: I suspect that the reason is that there is an "oral tradition" of P4 implementations that treat action bodies as if they have @atomic around their bodies, as BMv2 does.  It is certainly nothing from the language spec that encourages this.

On the other hand, I would guess that statements like "it is very difficult to make a blanket statement about annotations" probably also discourage people from wanting to rely on the language-standard @atomic annotation for things that they care about whether they are preserved by the p4c front end.  If there are no rules or invariants that annotations follow that are well known among compiler developers, they become mysterious scary things that lead to compiler bugs.

mbudiu-vmw: The compiler knows about atomic, so it should definitely handle that.
The compiler front-end does not know anything about other target-specific annotations that may exist.

jafingerhut:
@amreshk Is there any interest in pursuing this PR, or similar changes?

Issue #2860, opened 2021-07-28
Parser symbolic interpreter unable to evaluate lookahead statement


hanw:
I am trying to compile the following psa program on dpdk. It uses varbit and lookahead.

[psa-example-dpdk-varbit.p4.txt](https://github.com/p4lang/p4c/files/6889686/psa-example-dpdk-varbit.p4.txt)

The error reported by the compiler is
```
./p4c-dpdk ../testdata/p4_16_samples/psa-example-dpdk-varbit.p4
In file: /home/hanwang/p4c-dpdk2/midend/interpreter.h:165
Compiler Bug: no evaluation for tmp_1[15:8];
```
You can reproduce with this branch: hanw/dpdk-varbit
Any help about how to resolve this issue is much appreciated.

mbudiu-vmw:
The latest changes to the unrolling pass were made by @VolodymyrPeschanenkoLitSoft 
I will probably be offline for a few days, so if you are in a hurry you can possibly ask him to take a look.
I will only be able to look at this next week otherwise.

hanw:
Thanks. @VolodymyrPeschanenkoLitSoft Please take a look if yo could.

VolodymyrPeschanenkoLitSoft:
The main problem here is ExpressionEvaluator doesn't support Operation_Ternary. So, you can use #2861 code to solve this problem. After these changes I got the following error on your branch:
In file: p4c/backends/dpdk/dpdkHelpers.cpp:106
Compiler Bug: p4c/build/p4include/core.p4(34): packet_in Not implemented
extern packet_in {
       ^^^^^^^^^

mbudiu-vmw:
Is this solved?

mbudiu-vmw:
@hanw can you please check whether this issue can be closed?

hanw:
Yes the problem is solved. Thanks.

PR #2834, opened 2021-07-02
Added support for Type_HeaderUnion, IfStatement, Varbits

vladyslav-dubina:
Added support for IR :: Type_HeaderUnion to interpreter.
Added support for IR :: IfStatement in evaluateState function to ParserUnroll.
Both problems are caused in the example "issue1897-bmv2.p4"

PR #2861, opened 2021-07-28
Support of Operation_Ternary for ExpressionEvaluator

VolodymyrPeschanenkoLitSoft:


hanw (approved):
Thanks for the quick fix.

midend/interpreter.cpp
```diff
@@ -585,6 +585,56 @@ SymbolicVoid* SymbolicVoid::instance = new SymbolicVoid();
...
+        return;
+    }
+    auto e2 = get(expression->e2);
+    if (e2->is<SymbolicError>()) {
+        set(expression, e2);
+        return;
+    }
+    auto clone = expression->clone();
+    BUG_CHECK(e0->is<SymbolicInteger>(), "%1%: expected an SymbolicInteger", e0);
+    BUG_CHECK(e1->is<SymbolicInteger>(), "%1%: expected an SymbolicInteger", e1);
```
mbudiu-vmw, line 606: Why should e1 be an integer? Any other type would work here.

midend/interpreter.cpp
```diff
@@ -585,6 +585,56 @@ SymbolicVoid* SymbolicVoid::instance = new SymbolicVoid();
...
+        set(expression, e1);
+        return;
+    }
+    auto e2 = get(expression->e2);
+    if (e2->is<SymbolicError>()) {
+        set(expression, e2);
+        return;
+    }
+    auto clone = expression->clone();
+    BUG_CHECK(e0->is<SymbolicInteger>(), "%1%: expected an SymbolicInteger", e0);
```
mbudiu-vmw, line 605: Shouldn't this be a SymbolicBool?

mbudiu-vmw (changes requested):
This does not look right. Also, a test case would be nice.

midend/interpreter.cpp
hanw: Looks like this follows the implementation of IR::Operation_Binary below. Why wasn't SymbolicInteger an issue before?

mbudiu-vmw: There aren't many binary operations on non-integers. Comparisons are the only ones I can think of.
There are some passes in the midend which eliminate structure comparisons into field-wise comparisons.
So perhaps that's why. I don't think we eliminate ternary operations as well, but maybe we should.
Anyway, this deserves a test.

VolodymyrPeschanenkoLitSoft: > Looks like this follows the implementation of IR::Operation_Binary below. Why wasn't SymbolicInteger an issue before?

I think that this class was used for integer calculations only, but [this](https://github.com/p4lang/p4c/blob/main/frontends/common/constantFolding.h) class is used for calculation and it supports integer and booleans.  So, I have tried to add boolean support

VolodymyrPeschanenkoLitSoft: > There aren't many binary operations on non-integers. Comparisons are the only ones I can think of.
> There are some passes in the midend which eliminate structure comparisons into field-wise comparisons.
> So perhaps that's why. I don't think we eliminate ternary operations as well, but maybe we should.
> Anyway, this deserves a test.

I think that my last changes help to increase the functionality of this class. Now, It makes an evaluation of the boolean expressions.

midend/interpreter.cpp
VolodymyrPeschanenkoLitSoft: I think that for Mux it should be SymbolicBool, but for Slice it could be SymbolicInteger. So, I made some changes to support SymbolicInteger and SymbolicBool. Now, I think that implementation is much better.

mbudiu-vmw:
Would it be possible to add the test case from #2860?

mbudiu-vmw:
Fixes #2860

midend/interpreter.cpp
```diff
@@ -585,6 +585,77 @@ SymbolicVoid* SymbolicVoid::instance = new SymbolicVoid();
...
+        clone->e0 = getConstant(e0i);
+        clone->e1 = getConstant(e1i);
+        clone->e2 = getConstant(e2i);
+        DoConstantFolding cf(refMap, typeMap);
+        auto result = clone->apply(cf);
+        checkResult(expression, result);
+        return;
+    }
+    auto type = typeMap->getType(expression, true);
+    set(expression, new SymbolicInteger(ScalarValue::ValueState::NotConstant,
```
mbudiu-vmw, line 655: why is this always an integer? Can't it be a Bool, for example?

VolodymyrPeschanenkoLitSoft:
> Would it be possible to add the test case from #2860?

I have added this test, but it failed, because of unimplemented support of BlockStatement inside ParserUnroll. This feature is implemented in #2834. So, we have two possibilities here: remove this test (to ask [Vladyslav Dubina](https://github.com/vladyslav-dubina) to add this test to his PR) or we can wait for that PR, and after merging we will check this sample again.

mbudiu-vmw:
Looks like #2834 is ready for review, so let's merge that first, then you can rebase. Thank you.

VolodymyrPeschanenkoLitSoft:
> Looks like #2834 is ready for review, so let's merge that first, then you can rebase. Thank you.

Ok, I will wait for #2834

fruffy:
This is failing a test of the dpdk back end.

hanw:
Can we move forward with this PR? It's blocking me from making progress on other features.

VolodymyrPeschanenkoLitSoft:
> Can we move forward with this PR? It's blocking me from making progress on other features.

We can't move forward until #2834 will be merged, because of support for the BlockStatement in ParserUnroll. All changed after the review was made here, but your test still hasn't passed. So, when it will be merged I check your test in this PR. @mbudiu-vmw, could you, please, comment on this question?

mbudiu-vmw:
how about you just steal the code for the block statement from #2834 and add it to this PR?
We can't get that PR to pass review. It is still wrong.

hanw:
@VolodymyrPeschanenkoLitSoft Please do as Mihai suggested, my change does not depend on #2834

VolodymyrPeschanenkoLitSoft:
> @VolodymyrPeschanenkoLitSoft Please do as Mihai suggested, my change does not depend on #2834

I will do it tomorrow morning Ukrainian time.

VolodymyrPeschanenkoLitSoft:
> @VolodymyrPeschanenkoLitSoft Please do as Mihai suggested, my change does not depend on #2834

Your test was failed, because of packet_in:
Compiler Bug: /p4c/build/p4include/core.p4(34): packet_in Not implemented
extern packet_in {

@mbudiu-vmw, @hanw, What are the next steps?

hanw:
You can add the test to https://github.com/p4lang/p4c/blob/main/backends/dpdk/DpdkXfail.cmake

That is the error message that I'm trying to resolve.

PR #2863, opened 2021-07-30
Fix "dontcare" type method argument

vhavel:
Since ee77e40f0864c1157eb858bb14d6ea94f9837721 frontend (correctly)
inserts type casts of method call arguments if necessary.

However, if a method accepts "dontcare" input parameter and the method
is called with a InfInt value, a cast between these types is also
inserted. But the cast is later rejected during typechecking as
invalid. This commit prevents any cast to Dontcare type to be
inserted.


mbudiu-vmw:
Let me know when this is ready for review

vhavel:
@mbudiu-vmw it is ready for review.

mbudiu-vmw (approved):

PR #2864, opened 2021-08-03
DPDK backend: Fixing criteria for table match key consolidation into metadata

usha1830:
The criteria for making copy of header match keys in metadata was being incorrectly checked by the below condition:
`if (key->expression->toString() != firstKey->toString())`

Updated  to check various conditions in which the copy should be done. 
Added a few tests for covering various such scenarios.


usha1830:
@mbudiu-vmw @hanw Could you please review this PR?

backends/dpdk/dpdkArch.cpp
```diff
@@ -1006,30 +1006,44 @@ const IR::Node* CopyMatchKeysToSingleStruct::preorder(IR::Key* keys) {
...
         return keys;
     }
 
-    firstKey = keys->keyElements.at(0)->expression;
+    cstring firstKeyStr = "";
+    bool firstKeyHdr = false;
+
+    if (auto firstKeyField = keys->keyElements.at(0)->expression->to<IR::Member>()) {
+        firstKeyStr = firstKeyField->expr->toString();
+        if (firstKeyStr.startsWith("h"))
```
mbudiu-vmw, line 1020: can you explain what 'h' means?

usha1830: A previous pass converts all header fields to the form "h.<header_name>.<field_name> and similarly metadata fields are prefixed with "m.". So, here we are checking if the match key is part of a header by checking the "h" prefix.

mbudiu-vmw: That must be a comment in the code.

usha1830: Added a comment for the same.

mbudiu-vmw: I liked your comment on github better than the one in the code.

mbudiu-vmw (approved):
I will approve the implementation.
You should use rebase to merge the code from the repository, as described in the README.
You have picked up many unrelated commits which will be very confusing.
Maybe you can actually squash all these commits resubmit the PR with just one commit?

usha1830:
> Maybe you can actually squash all these commits resubmit the PR with just one commit?
I have created a new PR with just one commit https://github.com/p4lang/p4c/pull/2867

usha1830:
Created a New PR with these changes. #2867
Closing this PR.

PR #2865, opened 2021-08-03
annotation to disable strength reduction in a block

ChrisDodd:
This is a very trivial addition to allow disabling strength reduction in pieces of the code that is not really relevant to to language at all (its an annotation that does not affect program semantics at all), but I'm of mixed mind because it is probably not useful for programmers in any meaningful way.

The impetus for this actually comes from some hardware people who are trying to test corner cases in our hardware and are having difficulties because the compiler keeps modifying/eliminating their code, so the program ends up bypassing what they are trying to test.  This particular case happpens very early on in the frontend where we can't really avoid it in our custom midend/backend code.  I'm open to discussions of better ways similar things might be done -- perhaps some kind of target-specific policies that could be applied to frontend passes (as is used in the midend code in a number of places?)

### DIFF
--- a/frontends/p4/parseAnnotations.cpp
+++ b/frontends/p4/parseAnnotations.cpp
@@ -28,6 +28,7 @@ ParseAnnotations::HandlerMap ParseAnnotations::standardHandlers() {
             PARSE_EMPTY(IR::Annotation::optionalAnnotation),
             PARSE_EMPTY(IR::Annotation::pureAnnotation),
             PARSE_EMPTY(IR::Annotation::noSideEffectsAnnotation),
+            PARSE_EMPTY("disable_optimization"),
 
             // string literal argument.
             PARSE(IR::Annotation::nameAnnotation, StringLiteral),
--- a/frontends/common/constantFolding.h
+++ b/frontends/common/constantFolding.h
@@ -145,6 +145,10 @@ class DoConstantFolding : public Transform {
     const IR::Node* postorder(IR::IfStatement* statement) override;
     const IR::Node* preorder(IR::AssignmentStatement* statement) override;
     const IR::Node* preorder(IR::ArrayIndex* e) override;
+    const IR::BlockStatement *preorder(IR::BlockStatement *bs) override {
+        if (bs->annotations->getSingle("disable_optimization"))
+            prune();
+        return bs; }
 };
 
 /** Optionally runs @ref TypeChecking if @p typeMap is not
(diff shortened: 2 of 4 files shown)


mbudiu-vmw:
Can you supply justification? Perhaps we can give a more principled solution.
Which strength-reduction optimization is not useful?

mbudiu-vmw:
I see no reason not to approve this, but I would like to understand better what the problem solved is.

mbudiu-vmw (approved):

ChrisDodd:
After thinking about this a bit more, I decided to rename the annotation to `@disable_optimization` and extend it to also disable ConstantFolding and Reassociation in the block.  There may be other optimization that could be controlled this way.  This makes the annotation more "blunt" but also clearer in intent.  The fine grained control over precisely which statements are covered is still good.

The problem being solved here is people wanting to write very directed tests involving sub-optimal code to hit particular corner cases more easily.

mbudiu-vmw (approved):
It would be nice to document this somewhere.
I am sure what the best place is, though.

ChrisDodd:
Isn't that the point of making it open source?  Source code is self-documenting ;-)

frontends/p4/parseAnnotations.cpp
```diff
@@ -28,6 +28,7 @@ ParseAnnotations::HandlerMap ParseAnnotations::standardHandlers() {
             PARSE_EMPTY(IR::Annotation::optionalAnnotation),
             PARSE_EMPTY(IR::Annotation::pureAnnotation),
             PARSE_EMPTY(IR::Annotation::noSideEffectsAnnotation),
+            PARSE_EMPTY("disable_optimization"),
```
fruffy, line 31: Should this be a constant member like all the other annotations? This is the only string here. The annotations are also documented in base.def

fruffy (commented):
Is there any way to generalize this to the entire transform visitor and block statements? So that a particular block wrapped with the annotation is not optimized. 
In GCC, pragmas were used to locally disable optimizations afaik. Are annotations preferred here?

apinski-cavium:
> In GCC, pragmas were used to locally disable optimizations afaik. Are annotations preferred here?

Well GCC supports both really. It has the attribute optimize (which the pragma is converted into internally). Though it is only at the function granularity rather than block granularity like this annotation.

Now warnings on the other hand are on line granularity and uses pragmas.

Also yes it would be nice if this was documented even in an internals documentation and not part of the source (many compilers have this kind of thing really including GCC).

PR #2866, opened 2021-08-06
initial commit for pna.p4

hanw:
I have added pna.p4 to p4include to get folks started on writing pna programs. So far, people only need p4test to work to write syntactically correct pna programs. 

Backend support is work-in-progress.

### DIFF
--- a/p4include/pna.p4
+++ b/p4include/pna.p4
@@ -1,0 +1,700 @@
+/* Copyright 2020-present Intel Corporation
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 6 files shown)


p4include/pna.p4
```diff
@@ -0,0 +1,700 @@
...
+// + drop_packet
+// + send_to_port
+
+
+// drop_packet() - Cause the packet to be dropped when it finishes
+// completing the main control.
+//
+// Invoking drop_packet() is supported only within the main control.
+
+extern void drop_packet();
```
mbudiu-vmw, line 575: so there is no way to undo this?

mbudiu-vmw (approved):

p4include/pna.p4
hanw:
```// The following extern functions are "forwarding" functions -- they
// all set the destination of the packet.  Calling one of them
// overwrites and replaces the effect of any earlier call to any of
// the functions in this set.  Only the last one executed will
// actually take effect for the packet.

// + drop_packet
// + send_to_port
```

If send_to_port is called after drop_packet, then the packet would be forwarded.

PR #2870, opened 2021-08-11
DPDK backend: Fixing couple of issues in counter implementation and extend jump and label optimizations in action body

usha1830:
This PR includes the following changes
1) Fixing the increment value for packet and byte counters. The count method should increment the counter by packet length for byte counter and by 1 for packet counter.
2) Added parameter checks for the counters.
3) Modified tests related to counters as per the above change
4) In some cases, two labels are generated back to back. Extended the Asm optimization passes for removing redundant labels and jumps in action body also.
5) Added a test for label optimization in action body.
6) Updated p4test XFAIL list as p4test does not support conditional execution in action body
7) Updated reference outputs


mbudiu-vmw (approved):

usha1830:
@mbudiu-vmw If there are no review comments, could you please merge this PR?

PR #2871, opened 2021-08-11
DPDK backend: Add support for using isValid as table key

usha1830:
This PR includes changes for
- Adding support for using isValid as table key. 
- Handling logical operations as RHS of assignment

The table key having _isValid_ method call is first assigned to a temporary and then the temporary variable is used a table key.
This temporary variable is moved to metadata as DPDK requires all temporary variables to be moved to metadata.

```
table tbl {
    key = {
         hdr.ethernet.isValid() : exact @name ("hdrValid");
         hdr.ethernet.srcAddr : exact ;
      }
}

```
gets translated to 

```
struct metadata {
    ....
      bit<8> key;
}
..
table tbl {
    key = {
          key : exact @name ("hdrValid");
          hdr.ethernet.srcAddr : exact ;
      }
}

apply {
       key = hdr.ethernet.isValid();
       tbl.apply();
}

```

### DIFF
--- a/backends/dpdk/dpdkHelpers.cpp
+++ b/backends/dpdk/dpdkHelpers.cpp
@@ -55,12 +55,12 @@ void ConvertStatementToDpdk::process_relation_operation(const IR::Expression* ds
    Hence we convert the expression var = a LOP b into if-else statements.
    var = a || b            ||     var = a && b
 => if (a){                 ||=>   if (!a)
-       var = 1;            ||         var = 0;
+       var = true;         ||         var = false;
    } else {                ||     } else {
        if (b)              ||         if (!b)
-           var = 1;        ||             var = 0;
+           var = true;     ||             var = false;
        else                ||         else
-           var = 0;        ||             var = 1;
+           var = false;    ||             var = true;
    }                       ||     }
 
    This function assumes complex logical operations are converted to simple expressions by
--- a/backends/dpdk/dpdkHelpers.cpp
+++ b/backends/dpdk/dpdkHelpers.cpp
@@ -51,9 +51,53 @@ void ConvertStatementToDpdk::process_relation_operation(const IR::Expression* ds
     add_instr(new IR::DpdkLabelStatement(end_label));
 }
 
+/* DPDK target does not support storing result of logical operations such as || and &&.
+   Hence we convert the expression var = a LOP b into if-else statements.
+   var = a || b            ||     var = a && b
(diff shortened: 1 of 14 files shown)


backends/dpdk/dpdkHelpers.cpp
```diff
@@ -51,9 +51,53 @@ void ConvertStatementToDpdk::process_relation_operation(const IR::Expression* ds
...
+   ConvertLogicalExpression pass.
+*/
+void ConvertStatementToDpdk::process_logical_operation(const IR::Expression* dst,
+                                                        const IR::Operation_Binary* op) {
+    if (!op->is<IR::LOr>() && !op->is<IR::LAnd>())
+        return;
+    auto true_label = Util::printf_format("label_%dtrue", next_label_id);
+    auto false_label = Util::printf_format("label_%dfalse", next_label_id);
+    auto end_label = Util::printf_format("label_%dend", next_label_id++);
+    if (op->is<IR::LOr>()) {
```
mbudiu-vmw, line 76: I find this rather inefficient. On most architectures I know branches are actually significantly more expensive than Boolean operations. I expect you may want to reconsider this design decision.

mbudiu-vmw (approved):
I didn't check the translations in detail.

backends/dpdk/dpdkHelpers.cpp
```diff
@@ -51,9 +51,53 @@ void ConvertStatementToDpdk::process_relation_operation(const IR::Expression* ds
     add_instr(new IR::DpdkLabelStatement(end_label));
 }
 
+/* DPDK target does not support storing result of logical operations such as || and &&.
+   Hence we convert the expression var = a LOP b into if-else statements.
+   var = a || b            ||     var = a && b
+=> if (a){                 ||=>   if (!a)
+       var = 1;            ||         var = 0;
```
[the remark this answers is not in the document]
usha1830: Updated to true and false.

backends/dpdk/dpdkHelpers.cpp
usha1830: Yes. That's right. But in the absence of instructions to store the result of these operations or use them directly in jump instructions, I see no other way to handle these. We can revisit this if and when DPDK target introduces instructions for these logical operations.

usha1830:
@mbudiu-vmw If there are no further comments, could you please merge this PR?

PR #2794, opened 2021-06-10
[PSA] Checksum support for PSA model added

DusanKrdzic:
*Added support for padding in scalars*
-Length of scalars header need to be multiple of 8. This fixes the problem when the length of scalars header is not multiple of 8. It extends it with padding which size is 8-(scalars_header_length % 8).

*Added support for booleans in scalars header*
-Boolean variable need to be defined as a field in scalars header. This fixes registration of a boolean variable in scalars header.

*Added support for checksum in PSA model*
-Parser and deparser recognize PSA Checksum instances in their locals and convert them into JSON.
-Parser and deparser recognize PSA checksum methods and convert them into JSON.

link to behavioral-model pull request: https://github.com/p4lang/behavioral-model/pull/1007

PR #2872, opened 2021-08-12
[PSA] Internet Checksum support for PSA model added

DusanKrdzic:
Added support for padding in scalars
-Length of scalars header need to be multiple of 8. This fixes the problem when the length of scalars header is not multiple of 8. It extends it with padding which size is 8-(scalars_header_length % 8).

Added support for booleans in scalars header
-Boolean variable need to be defined as a field in scalars header. This fixes registration of a boolean variable in scalars header.

Added support for internet checksum in PSA model
-Parser and deparser recognize PSA Checksum instances in their locals and convert them into JSON.
-Parser and deparser recognize PSA checksum methods and convert them into JSON.

InternetChecksum implementation requires some parts of code from current basic Checksum implementation, so
some parts are the same as in the pr for basic Checksum: #2794

behavioral-model PSA_InternetChecksum pr: https://github.com/p4lang/behavioral-model/pull/1018

behavioral-model PSA_InternetChecksum : https://github.com/p4lang/behavioral-model/commit/6459a750a512916435014af12cea5ba9f1121d7d


DusanKrdzic:
@mbudiu-vmw  The InternetChecksum PR is ready for review. Behavioral-model InternetChecksum implementation has been merged successfully into p4lang/behavioral-model: https://github.com/p4lang/behavioral-model/commit/6459a750a512916435014af12cea5ba9f1121d7d

DusanKrdzic:
I am not sure why it fails here, locally it passes.

DusanKrdzic:
I am not sure why it fails here, locally passes.

mbudiu-vmw:
These are the tests that fail in CI:
bmv2/testdata/p4_16_samples/internet_checksum1-bmv2.p4 (Failed)
p4/testdata/p4_16_samples/internet_checksum1-bmv2.p4 (Failed)
You can try them separately on your machine.
Sometimes if there are changes in bmv2 I believe that the CI may be using a stale version.

backends/bmv2/common/deparser.cpp
```diff
@@ -24,14 +24,53 @@ void DeparserConverter::convertDeparserBody(const IR::Vector<IR::StatOrDecl>* bo
...
+        IR::MethodCallExpression *mce2 = nullptr;
         if (auto block = s->to<IR::BlockStatement>()) {
             convertDeparserBody(&block->components, order, primitives);
             continue;
         } else if (s->is<IR::ReturnStatement>() || s->is<IR::ExitStatement>()) {
             break;
         } else if (s->is<IR::EmptyStatement>()) {
             continue;
-        } else if (s->is<IR::MethodCallStatement>()) {
+        } else if (s->is<IR::AssignmentStatement>()) {
```
mbudiu-vmw, line 36: Should these changes be in common/deparser.cpp? They are specific to PSA, and not common with v1model for example. Can you find a way to move these changes to the PSA backend without duplicating too much code?

backends/bmv2/common/deparser.cpp
```diff
@@ -24,14 +24,53 @@ void DeparserConverter::convertDeparserBody(const IR::Vector<IR::StatOrDecl>* bo
...
+                json->emplace("op", "set");
+                auto l = ctxt->conv->convertLeftValue(assign->left);
+                bool convertBool = type->is<IR::Type_Boolean>();
+                auto r = ctxt->conv->convert(assign->right, true, true, convertBool);
+                params->append(l);
+                params->append(r);
+                primitives->append(json);
+                ctxt->conv->simpleExpressionsOnly = true;
+                continue;
+            }
```
mbudiu-vmw, line 71: shouldn't there be an else clause here that gives an error for other kinds of expressions on the right?

DusanKrdzic:
> These are the tests that fail in CI:
> 
> bmv2/testdata/p4_16_samples/internet_checksum1-bmv2.p4 (Failed)
> 
> p4/testdata/p4_16_samples/internet_checksum1-bmv2.p4 (Failed)
> 
> You can try them separately on your machine.
> 
> Sometimes if there are changes in bmv2 I believe that the CI may be using a stale version.

Both of them passes on my machine.

mbudiu-vmw:
This is one error:
```
Invalid reference to extern type 'InternetChecksum'
bad json:
{
   "id" : 0,
   "name" : "IngressParserImpl.ck",
   "source_info" : {
      "column" : 23,
      "filename" : "/p4c/testdata/p4_16_samples/internet_checksum1-bmv2.p4",
      "line" : 62,
      "source_fragment" : "ck"
   },
   "type" : "InternetChecksum"
}
```

This looks like a bmv2 problem.

DusanKrdzic:
> This is one error:
> 
> ```
> 
> Calling target program-options parser
> 
> Invalid reference to extern type 'InternetChecksum'
> 
> bad json:
> 
> {
> 
>    "id" : 0,
> 
>    "name" : "IngressParserImpl.ck",
> 
>    "source_info" : {
> 
>       "column" : 23,
> 
>       "filename" : "/p4c/testdata/p4_16_samples/internet_checksum1-bmv2.p4",
> 
>       "line" : 62,
> 
>       "source_fragment" : "ck"
> 
>    },
> 
>    "type" : "InternetChecksum"
> 
> }
> 
> ```
> 
> 
> 
> This looks like a bmv2 problem.

Yes, and locally it works. It seems like CI doesn't recognize new bm code version.

backends/bmv2/common/parser.cpp
```diff
@@ -48,6 +48,61 @@ cstring ParserConverter::jsonAssignment(const IR::Type* type, bool inParser) {
 Util::IJson* ParserConverter::convertParserStatement(const IR::StatOrDecl* stat) {
     auto result = new Util::JsonObject();
     auto params = mkArrayField(result, "parameters");
+    auto isR = false;
+    IR::MethodCallExpression *mce2 = nullptr;
+    if (stat->is<IR::AssignmentStatement>()) {
```
mbudiu-vmw, line 53: same comment about changing common files.

mbudiu-vmw:
The solution to that is to wait until it picks a new version. I think it caches the old one.

backends/bmv2/common/parser.cpp
```diff
@@ -48,6 +48,61 @@ cstring ParserConverter::jsonAssignment(const IR::Type* type, bool inParser) {
...
+                    isR=true;
+                    auto dest = new IR::Argument(l);
+                    auto args = new IR::Vector<IR::Argument>();
+                    args->push_back(dest);  // dest
+                    mce2 = new IR::MethodCallExpression(mce->method, mce->typeArguments);
+                    mce2->arguments = args;
+                    stat = new IR::MethodCallStatement(mce);
+                }
+            }
+        } else if (assign->right->is<IR::Equ>()
```
mbudiu-vmw, line 74: this is awfully specific. 
what happens to some equivalent rewrites of this expression, e.g., assigning the result to a temporary and then comparing?

DusanKrdzic:
> The solution to that is to wait until it picks a new version. I think it caches the old one.

How do I know when to push new commit to check if it picks a new version?

backends/bmv2/psa_switch/psaSwitch.cpp
```diff
@@ -648,7 +667,36 @@ Util::IJson* ExternConverter_InternetChecksum::convertExternObject(
...
+    parameters->append(cksum);
+    if (em->method->name == "add" || em->method->name == "subtract") {
+        auto fieldList=new Util::JsonObject();
+        fieldList->emplace("type","field_list");
+        auto fieldsJson = ctxt->conv->convert(mc->arguments->at(0)->expression, true, false);
+        fieldList->emplace("value",fieldsJson);
+        parameters->append(fieldList);
+    } else if (em->method->name != "clear") {
+        if (mc->arguments->size() == 2) {  // get_verify
+        auto dst = ctxt->conv->convertLeftValue(mc->arguments->at(0)->expression);
```
mbudiu-vmw, line 691: indentation is off

backends/bmv2/psa_switch/psaSwitch.cpp
```diff
@@ -648,7 +667,36 @@ Util::IJson* ExternConverter_InternetChecksum::convertExternObject(
...
+    cksum->emplace("value", em->object->controlPlaneName());
+    parameters->append(cksum);
+    if (em->method->name == "add" || em->method->name == "subtract") {
+        auto fieldList=new Util::JsonObject();
+        fieldList->emplace("type","field_list");
+        auto fieldsJson = ctxt->conv->convert(mc->arguments->at(0)->expression, true, false);
+        fieldList->emplace("value",fieldsJson);
+        parameters->append(fieldList);
+    } else if (em->method->name != "clear") {
+        if (mc->arguments->size() == 2) {  // get_verify
```
mbudiu-vmw, line 690: If people use the wrong model psa.p4 when compiling these assumptions may not be true. 
You should really check them and give a model error if they are not satisfied.

mbudiu-vmw (changes requested):

mbudiu-vmw:
> 
> 
> > The solution to that is to wait until it picks a new version. I think it caches the old one.
> 
> How do I know when to push new commit to check if it picked a new version?

I don't know, perhaps 48 hours will be enough. In the meantime you can work on the other comments and when they are addressed perhaps it will just work.

backends/bmv2/common/parser.cpp
DusanKrdzic: This is used when **get** method is in **verify** functin. For example **verify(ipv4.hdrChecksum == ck.get(), flag)**. Compiler will always transform this in:

**bool tmp;
tmp = ipv4.hdrChecksum == ck.get(); 
verify(tmp, flag)**

Equ expression is for the second line. To process **tmp** and **ipv4.hdrChecksum** for bm **PSA_InternetChecksum::get_verify()** method.

mbudiu-vmw: For example, what if the compiler generates:
```
a = ck.get();
verify(a == ipv4.checksum, "")
```
This is a perfectly legal transformation.

antoninbas:
Docker has finally killed automated builds for free plans, so the bmv2 docker image is not getting updated any more. I'll try to push a band aid patch for this today.

mbudiu-vmw:
We could change our CI build process to also build bmv2 from sources, but that certainly is expensive.

antoninbas:
@DusanKrdzic the docker image for bmv2 should be updated now, if you want to start the tests again

backends/bmv2/common/deparser.cpp
DusanKrdzic: I put this in common for several reasons:

1. If someone wants to use `InternetChecksum` in simple switch and v1model, program will fail before it gets here, because ss arch doesn't recognize it.
2. To avoid redundant code.
3. Other structures that are only for one architecture are also used in common section.. for example:
https://github.com/p4lang/p4c/blob/2c2c928e972f92dd5fb1de9509e7bd4d91c354ca/backends/bmv2/common/action.cpp#L63-L80
https://github.com/p4lang/p4c/blob/2c2c928e972f92dd5fb1de9509e7bd4d91c354ca/backends/bmv2/common/parser.cpp#L206-L214

backends/bmv2/common/deparser.cpp
DusanKrdzic: I will add it.

backends/bmv2/common/parser.cpp
```diff
@@ -48,6 +48,61 @@ cstring ParserConverter::jsonAssignment(const IR::Type* type, bool inParser) {
...
+                    auto equOp = new IR::Argument(r);
+                    auto args = new IR::Vector<IR::Argument>();
+                    args->push_back(dest);  // dest
+                    args->push_back(equOp);
+                    mce2 = new IR::MethodCallExpression(mce->method, mce->typeArguments);
+                    mce2->arguments = args;
+                    stat = new IR::MethodCallStatement(mce);
+                }
+            }
+        }
```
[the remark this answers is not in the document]
DusanKrdzic: If it isn't the extern method I'm looking for, it won't get in that `if clause`, it will just skip the condition.

DusanKrdzic:
Now it passes bm part of tests. It fails in p4c test because of include `<psa.p4> -> <bmv2/psa.p4>`.:
``` 
BadIPv4HeaderChecksum
 }
 #include <core.p4>
-#include <psa.p4>
+#include <bmv2/psa.p4>
 ```
 It passes locally, but I will change it on code and push again. That should work.

File: backends/bmv2/common/parser.cpp (hunk 4 above)
Comment (DusanKrdzic) [reply backends/bmv2/common/parser.cpp:74]: There are two possible cases.

1. If the temporary variable `a` is assigned in the same block as `verify` function. 
 ```
a=ck.get();
 verify(hdr.ipv4.hdrChecksum == ck.get(), error.BadIPv4HeaderChecksum);
```

Then the compiler will transform it before it gets in this part of code:
Screenshot from 2021-08-13 09-30-01
This case has been resolved with this `Equ` `if clause` in `p4c` and with `get_verify(Field &dst, Field &equOp) const;` method in bm project.
First it asks if there is an `Equ` on the right side of `=`. If it is than it asks if there is an extern method. If there is it will precess it only if there is `InternetChecksum::get`.

2. If the temporary variable `a` is assigned in the other block in parser.
```
state x {
...
a = ck.get();
...
}
state y {
...
verify(hdr.ipv4.hdrChecksum == a, error.BadIPv4HeaderChecksum);
...
}
```
Then the compiler will transform it:
Screenshot from 2021-08-13 09-38-05
Here there is no `get` method in `Equ` and "this is no more extern method `get` problem". 
This will be processed in classic `Assignment statement`:
https://github.com/p4lang/p4c/pull/2872/files#diff-d0c07a9f6f6d88c535125cb0a83255a81340b8f55f87f645b0bfcbf58109ea0eR106-R126
In JSON this case looks like this (`tmp_10 = hdr.ipv4.hdrChecksum == a_0;`):
Screenshot from 2021-08-13 09-50-22
So in bm it is not `get` method problem.

backends/bmv2/common/parser.cpp
```diff
@@ -522,6 +595,27 @@ bool ParserConverter::preorder(const IR::P4Parser* parser) {
...
+    for (auto p : parser->parserLocals) {
+        if (p->is<IR::Declaration_Constant>() ||
+        p->is<IR::Declaration_Variable>() ||
+        p->is<IR::P4Action>() ||
+        p->is<IR::P4Table>())
+        continue;
+        if (p->is<IR::Declaration_Instance>()) {
+            auto bl = ctxt->structure->resourceMap.at(p);
+            CHECK_NULL(bl);
+            if (bl->is<IR::ControlBlock>() || bl->is<IR::ParserBlock>())
```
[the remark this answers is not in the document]
DusanKrdzic: I will delete it.

backends/bmv2/psa_switch/psaSwitch.cpp
DusanKrdzic: I will add it.

backends/bmv2/psa_switch/psaSwitch.cpp
DusanKrdzic: I will add the number of arguments check.

backends/bmv2/common/deparser.cpp
mbudiu-vmw: Perhaps, but this is actually bad for long term maintenance. People who will modify this code in the future will be confused about why this is here, as I am now. The fact that it was done before does not mean it's right, it means that the review didn't catch it.
If you really insist having this code here please add a commend explaining that it is in fact intended just for the PSA back-end.

DusanKrdzic: In fact I was also confused when I saw structures that are not common in common folder. I will add this part of code in `psaSwitch.cpp.`

DusanKrdzic: I added comments. 
In current `p4c PSA switch` implementation this is the best approach to preserve performance and avoid the same processing multiple times (processing parsers - first time for common structures and second only for `InternetChecksum`).

DusanKrdzic:
@mbudiu-vmw Is there anything else you want me to do in order to fix and merge this pr?

backends/bmv2/common/parser.cpp
mbudiu-vmw: This code is not incorrect, it is just overly specific. I hope it won't break when other changes are made to passes that precede it.

mbudiu-vmw (approved):

PR #2881, opened 2021-08-24
Giving warnings on accessing fields of invalid headers

### LINKED ISSUE #2784: Compiler does not give warnings on accessing fields of invalid headers

kostasyrmia:
Fix #2784 
This improves the FindUninitialized pass to report warnings on accessing fields of invalid headers. Currently, the warnings are reported only for headers and header fields of structs, but not for header stacks and header unions.

According to the specification, in the following cases headers become valid:

 - calling the setValid method of a header
 - calling the extract extern method of the packet_in type
 - initialization with a list expression
 - initialization with some valid header

In the following cases headers become invalid:

 - calling the setInvalid method of a header
 - initialization with some invalid header

During visiting of the statements in the program, the current values of header valid bits are stored in a map, which is updated each time some of these cases are encountered. When encountering if and switch statements, each branch is entered with the same information that the pass had until that point, and at the end of these statements, the common part of all branches is taken.

Currently, the conditions of if statements are not processed. The pass is visiting each branch with the information that it had until that point. This can lead to giving some false warnings in the situation where there are calls to the isValid method in the condition. For example, the following snippet of code would give warnings for every access to hdr fields in the if block, even though the header hdr is guaranteed to be valid inside the if block.

    hdr.setInvalid();
    if (hdr.isValid()) {   
      // read some hdr fields here   
    }
The expression in the condition can become more complex by adding &&, || and !. So to prevent giving false warnings in this situation, each time isValid call is encountered, that header is temporarily put in a set of 'not-to-report' headers until the end of the block or until that header is again explicitly set to valid or invalid inside that block.

### DIFF
--- a/frontends/p4/simplifyDefUse.cpp
+++ b/frontends/p4/simplifyDefUse.cpp
@@ -18,6 +18,7 @@ limitations under the License.
 #include "frontends/p4/def_use.h"
 #include "frontends/p4/methodInstance.h"
 #include "frontends/p4/tableApply.h"
+#include "frontends/p4/ternaryBool.h"
 #include "frontends/p4/sideEffects.h"
 #include "frontends/p4/parserCallGraph.h"
 
@@ -102,14 +103,13 @@ class HeaderDefinitions {
     ReferenceMap* refMap;
     StorageMap* storageMap;
 
-    /// The current values of the header valid bits are stored here. If a header is in the
-    /// map and the value of its valid bit is true, then the header is currently valid. If
-    /// the header is in the map and the value of its valid bit is false, then the header is
-    /// currently invalid. If the header is not in the map, then the header is potentially
-    /// invalid (for example, this can happen when the header is valid at the end of the then
-    /// branch and invalid at the end of the else branch of an if statement or if the header
-    /// is valid entering a parser state on some input branches and invalid on some other)
-    ordered_map<const StorageLocation*, bool> defs;
+    /// The current values of the header valid bits are stored here. If the value in the map is Yes,
+    /// then the header is currently valid. If the value in the map is No, then the header is
+    /// currently invalid. If the value in the map is Maybe, then the header is potentially invalid
+    /// (for example, this can happen when the header is valid at the end of the then branch and
+    /// invalid at the end of the else branch of an if statement, or if the header is valid entering
+    /// a parser state on some input branches and invalid on some other)
+    ordered_map<const StorageLocation*, TernaryBool> defs;
 
     /// Currently isValid() expressions in if conditions are not processed, so all headers
     /// for which isValid() is called are temporarly stored here until the end of the block
@@ -144,7 +144,7 @@ class HeaderDefinitions {
                 "location %1% is not a header", storage->name);
     }
 
-    void add(const StorageLocation* storage, bool valid) {
+    void add(const StorageLocation* storage, TernaryBool valid) {
         if (!storage)
             return;
 
@@ -153,7 +153,7 @@ class HeaderDefinitions {
         notReport.erase(storage);
     }
 
-    void add(const IR::Expression *expr, bool valid) {
+    void add(const IR::Expression *expr, TernaryBool valid) {
         if (!expr)
             return;
 
@@ -161,7 +161,7 @@ class HeaderDefinitions {
         add(storage, valid);
     }
 
-    void update(const StorageLocation *storage, bool valid) {
+    void update(const StorageLocation *storage, TernaryBool valid) {
         if (!storage)
             return;
 
@@ -170,7 +170,7 @@ class HeaderDefinitions {
         notReport.erase(storage);
     }
 
-    void update(const IR::Expression* expr, bool valid) {
+    void update(const IR::Expression* expr, TernaryBool valid) {
         if (!expr)
             return;
 
@@ -178,19 +178,19 @@ class HeaderDefinitions {
         update(storage, valid);
     }
 
-    const bool* find(const StorageLocation* storage) const {
+    TernaryBool find(const StorageLocation* storage) const {
         if (!storage)
-            return new bool(true);
+            return TernaryBool::Yes;
 
         if (notReport.find(storage) != notReport.end())
-            return new bool(true);
+            return TernaryBool::Yes;
 
-        return ::getref(defs, storage);
+        return ::get(defs, storage, TernaryBool::Maybe);
     }
 
-    const bool* find(const IR::Expression* expr) const {
+    TernaryBool find(const IR::Expression* expr) const {
         if (!expr)
-            return new bool(true);
+            return TernaryBool::Yes;
 
         auto storage = getStorageLocation(expr);
         return find(storage);
@@ -221,9 +221,8 @@ class HeaderDefinitions {
     HeaderDefinitions* intersect(const HeaderDefinitions* other) const {
         HeaderDefinitions* result = new HeaderDefinitions(refMap, storageMap);
         for (auto def : defs) {
-            auto val = other->find(def.first);
-            if (val && *val == def.second)
-                result->add(def.first, def.second);
+            auto valid = other->find(def.first);
+            result->add(def.first, valid == def.second ? valid : TernaryBool::Maybe);
         }
         return result;
     }
@@ -332,7 +331,7 @@ class FindUninitialized : public Inspector {
         return defs;
     }
 
-    void initHeaderParam(const StorageLocation* storage, bool value) {
+    void initHeaderParam(const StorageLocation* storage, TernaryBool value) {
         if (auto struct_storage = storage->to<StructLocation>()) {
             if (struct_storage->isHeader()) {
                 headerDefs->update(struct_storage, value);
@@ -346,8 +345,11 @@ class FindUninitialized : public Inspector {
     // Called at the beginning of controls, parsers and functions
     void initHeaderParams(const IR::ParameterList* parameters) {
         for (auto p : parameters->parameters)
-            if (auto storage = definitions->storageMap->getStorage(p))
-                initHeaderParam(storage, p->direction != IR::Direction::Out);
+            if (auto storage = definitions->storageMap->getStorage(p)) {
+                initHeaderParam(storage, p->direction != IR::Direction::Out
+                                         ? TernaryBool::Yes
+                                         : TernaryBool::No);
+            }
     }
 
     void checkOutParameters(const IR::IDeclaration* block,
@@ -488,9 +490,9 @@ class FindUninitialized : public Inspector {
                     inputHeaderDefs[n] = headerDefs->clone();
                     toRun.emplace(n);
                 } else {
-                    auto joined = inputHeaderDefs[n]->intersect(headerDefs);
-                    if (*joined != *inputHeaderDefs[n]) {
-                        inputHeaderDefs[n] = joined;
+                    auto newInputDefs = inputHeaderDefs[n]->intersect(headerDefs);
+                    if (*newInputDefs != *inputHeaderDefs[n]) {
+                        inputHeaderDefs[n] = newInputDefs;
                         toRun.emplace(n);
                     }
                 }
@@ -586,11 +588,8 @@ class FindUninitialized : public Inspector {
             return;
 
         if (dst_struct_storage->isHeader() && src_struct_storage->isHeader()) {
-            auto val = headerDefs->find(src);
-            if (val)
-                headerDefs->update(dst, *val);
-            else
-                headerDefs->remove(dst);
+            auto valid = headerDefs->find(src);
+            headerDefs->update(dst, valid);
 
             return;
         }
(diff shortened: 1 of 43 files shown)


kostasyrmia:
@mbudiu-vmw
I have fixed the issues u have requested and I have double-checked the -error files and they seem ok to me. Could you please check them too?

kostasyrmia:
I forgot to mention the issue in the gauntlet_hdr_assign_1-bmv2.p4 test. If I am not wrong, the first FindUninitialized pass ends without warnings, then the RemoveUnused pass removes the assignment x = h.h in the ingress control, and finally, when these passes are repeated, we get some warnings, which I think should not be reported.

kostasyrmia:
Also in addition to warnings for invalid headers, I have added warnings for potentially invalid headers because I thought that it would be useful too.  I am interested in your opinion on that. Do you think they should be reported or rather not?

mbudiu-vmw:
> 
> 
> Also in addition to warnings for invalid headers, I have added warnings for potentially invalid headers because I thought that it would be useful too. I am interested in your opinion on that. Do you think they should be reported or rather not?

If such warnings generate too many false positives they should be avoided. People really don't like spurious warnings.

testdata/p4_16_samples_outputs/gauntlet_hdr_assign_1-bmv2.p4-stderr
```diff
@@ -0,0 +1,6 @@
+gauntlet_hdr_assign_1-bmv2.p4(38): [--Wwarn=invalid_header] warning: accessing a field of an invalid header x
```
mbudiu-vmw, line 1: I think this message is wrong. The previous line was `x = h.h`, so x.a may actually be valid if h.h.a was valid.

mbudiu-vmw:
Without reading your code buy by looking at the produced error message your analysis is still producing wrong warnings.

testdata/p4_16_samples_outputs/gauntlet_hdr_assign_1-bmv2.p4-stderr
kostasyrmia: Yes, I am aware of this issue. I commented today (above) about this situation, I think that the RemoveUnused pass removed that line.

> I forgot to mention the issue in the gauntlet_hdr_assign_1-bmv2.p4 test. If I am not wrong, the first FindUninitialized pass ends without warnings, then the RemoveUnused pass removes the assignment x = h.h in the ingress control, and finally, when these passes are repeated, we get some warnings, which I think should not be reported.

Here is the final output after all passes in simplifyDefUse finish:

    apply /* 
    <BlockStatement>(74255) */
    {
        /* 
      <MethodCallStatement>(62620)
        <MethodCallExpression>(62621)
          <Member>(62622)setInvalid
            <PathExpression>(62623)
              x_0/x
          <Vector<Type>>(61096), size=0
          <Vector<Argument>>(61094), size=0 */
        x_0.setInvalid();
        /* 
        <EmptyStatement>(74270) */    <-------------- this line
        ;
        /* 
      <AssignmentStatement>(62635)
        <Member>(62636)a
          <PathExpression>(62637)
            x_0/x
        <Constant>(1313) 2
          <Type_Bits>(289) */
        x_0.a = 16w2;
        /* 
      <AssignmentStatement>(62642)
        <Member>(46198)eth_type
          <Member>(46199)eth_hdr
            <PathExpression>(46200)
              h
        <Member>(62647)a
          <PathExpression>(62648)
            x_0/x */
        h.eth_hdr.eth_type = x_0.a;
    }

If you want, I can try to find out why this is happening.

kostasyrmia:
> > Also in addition to warnings for invalid headers, I have added warnings for potentially invalid headers because I thought that it would be useful too. I am interested in your opinion on that. Do you think they should be reported or rather not?
> 
> If such warnings generate too many false positives they should be avoided. People really don't like spurious warnings.

Well, I think it works well for these small programs, but I am not sure how it would behave in some more serious programs. In any case, we can remove it if it starts bothering us.

testdata/p4_16_samples_outputs/gauntlet_hdr_assign_1-bmv2.p4-stderr
mbudiu-vmw: If that's what's happening it's clearly a bug, since x is not unused. Please file an issue about it.

mbudiu-vmw:
Let's merge first https://github.com/p4lang/p4c/pull/2891, then you can rebase and I will review the code.
This looks like a very useful analysis based on the number of warnings produced.

kostasyrmia:
I have updated the branch, I think everything is ready for review.

testdata/p4_16_samples_outputs/psa-table-hit-miss.p4-error
```diff
@@ -0,0 +1,3 @@
+psa-table-hit-miss.p4(62): [--Wwarn=invalid_header] warning: accessing a field of a potentially invalid header hdr.ethernet
```
mbudiu-vmw, line 1: what is this file

testdata/p4_16_samples_outputs/psa-table-hit-miss.p4-stderr
```diff
@@ -0,0 +1,3 @@
+psa-table-hit-miss.p4(62): [--Wwarn=invalid_header] warning: accessing a field of a potentially invalid header hdr.ethernet
+            hdr.ethernet.srcAddr : exact;
```
mbudiu-vmw, line 2: I am not sure that this message is desirable. Why does it appear here?
Because you are assuming the fields of hdr may be invalid?

frontends/p4/simplifyDefUse.cpp
```diff
@@ -97,6 +98,150 @@ class HasUses {
...
+    }
+    void update(const IR::Expression* expr, bool valid) {
+        if (!expr)
+            return;
+
+        auto storage = getStorageLocation(expr);
+        update(storage, valid);
+    }
+
+    const bool* find(const StorageLocation* storage) const {
```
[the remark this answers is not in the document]
mbudiu-vmw: There is also an enum in removeReturns.h that has 3 values (Yes, No, Maybe).
We could move that to a separate file and name it "TernaryBool" and reuse it.

frontends/p4/simplifyDefUse.cpp
```diff
@@ -179,6 +332,24 @@ class FindUninitialized : public Inspector {
...
+                    initHeaderParam(f, value);
+            }
+        }
+    }
+
+    // Called at the beginning of controls, parsers and functions
+    void initHeaderParams(const IR::ParameterList* parameters) {
+        for (auto p : parameters->parameters)
+            if (auto storage = definitions->storageMap->getStorage(p))
+                initHeaderParam(storage, p->direction != IR::Direction::Out);
```
mbudiu-vmw, line 350: I guess you can't really have header constants (with IR::Direction::None), but perhaps you should exclude these too.

frontends/p4/simplifyDefUse.cpp
```diff
@@ -271,8 +451,63 @@ class FindUninitialized : public Inspector {
...
+            FindUninitialized fu(this, currentPoint);
+            (void)state->apply(fu);
+
+            auto next = transitions.getCallees(state);
+            for (auto n : *next) {
+                if (inputHeaderDefs.find(n) == inputHeaderDefs.end()) {
+                    inputHeaderDefs[n] = headerDefs->clone();
+                    toRun.emplace(n);
+                } else {
+                    auto joined = inputHeaderDefs[n]->intersect(headerDefs);
```
mbudiu-vmw, line 491: usually "join" means "union", so perhaps you want to rename this variable to something like 'common'.

frontends/p4/simplifyDefUse.cpp
```diff
@@ -339,6 +574,83 @@ class FindUninitialized : public Inspector {
...
+
+            auto it1 = dst_fields.begin();
+            auto it2 = src_fields.begin();
+            while (it1 != dst_fields.end() && it2 != src_fields.end()) {
+                processHeadersInAssignment(*it1, *it2);
+                ++it1;
+                ++it2;
+            }
+        }
+    }
```
mbudiu-vmw, line 610: BUG_CHECK to make sure that no other cases exist?

frontends/p4/simplifyDefUse.cpp
```diff
@@ -339,6 +574,83 @@ class FindUninitialized : public Inspector {
...
+        if (auto dst_struct_storage = dst->to<StructLocation>()) {
+            if (dst_struct_storage->isHeader()) {
+                if (src->is<IR::StructExpression>()) {
+                    // always sets the valid bit to true
+                    headerDefs->update(dst, true);
+                } else if (src->is<IR::MethodCallExpression>()) {
+                    // return value of a function
+                    headerDefs->update(dst, true);
+                } else if (src->is<IR::ArrayIndex>()) {
+                    // TODO: header stacks
```
mbudiu-vmw, line 627: Can't you just keep track of a valid bit for each slot of the array?
If the index is not constant you can assume the "or" of all valid bits of all elements for reads.
You can do this in a subsequent PR.

frontends/p4/simplifyDefUse.cpp
```diff
@@ -558,20 +897,67 @@ class FindUninitialized : public Inspector {
...
         return false;
     }
 
+    void reportWarningIfInvalidHeader(const IR::Expression* expression, const IR::Type *expr_type) {
+        if (!reportInvalidHeaders)
+            return;
+
+        if (expr_type->is<IR::Type_Header>()) {
+            if (auto member = expression->to<IR::Member>()) {
+                // TODO: header unions
```
mbudiu-vmw, line 933: you may be able to model a header_union as a struct of headers.
also, writing to a union field implicitly invalidates all other fields.
this can be a separate PR.

frontends/p4/simplifyDefUse.cpp
```diff
@@ -589,18 +975,47 @@ class FindUninitialized : public Inspector {
...
+            auto am = mi->to<ApplyMethod>();
+            isControlOrParserApply = !am->isTableApply();
+        }
         for (auto p : *mi->substitution.getParametersInArgumentOrder()) {
             auto expr = mi->substitution.lookup(p);
             if (p->direction != IR::Direction::Out) {
                 visit(expr);
             }
+
+            // We assume control and parser apply calls set all output headers to valid
```
mbudiu-vmw, line 1003: that should be true of extern functions too.

frontends/p4/simplifyDefUse.cpp
```diff
@@ -614,6 +1029,18 @@ class FindUninitialized : public Inspector {
                 callee.push_back(table);
             }
         } else if (auto em = mi->to<ExternMethod>()) {
+            auto args = em->expr->arguments;
+            if (args->size() > 0) {
+                auto header = args->at(0);
+                if (em->originalExternType->getName().name == "packet_in" &&
+                    em->method->getName().name == "extract") {
```
mbudiu-vmw, line 1036: if you assume that extern methods will potentially set the valid bit of every header you probably don't need this case anymore.

frontends/p4/simplifyDefUse.cpp
```diff
@@ -614,6 +1029,18 @@ class FindUninitialized : public Inspector {
...
+                        headerDefs->update(header->expression, true);
+                } else if (em->originalExternType->getName().name == "packet_out" &&
+                         em->method->getName().name == "emit") {
+                    reportWarningIfInvalidHeader(header->expression,
```
mbudiu-vmw, line 1040: why is this method special? all methods that have an 'in' argument of type containing a header should be handled in the same way.

mbudiu-vmw (approved):
I will approve, but perhaps you can still take a look at the comments.

testdata/p4_16_samples_outputs/gauntlet_ctrl_plane_hdr.p4-stderr
```diff
@@ -0,0 +1,3 @@
+gauntlet_ctrl_plane_hdr.p4(25): [--Wwarn=invalid_header] warning: accessing a field of a potentially invalid header ctrl_hdr
```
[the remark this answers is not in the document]
kostasyrmia: Making the warning function check the context itself would be the best solution in my opinion, but I am not sure how exactly it could be done. I agree about the annotation, I will check it for some of the subsequent commits.

testdata/p4_16_samples_outputs/psa-table-hit-miss.p4-error
kostasyrmia: Well, ironically, when I pushed yesterday, this test failed. I checked the logs and there was a message that this output was expected in the file. This was very strange considering that I didn't change any of the files and the tests were passing before. I will remove this file and try again.

testdata/p4_16_samples_outputs/psa-table-hit-miss.p4-stderr
kostasyrmia: If I am not wrong, after the first if statement in that control, the header will be invalid after the then branch is executed, or will be valid if the condition in the if is false. So for the second apply call, the header might be invalid.
But I am considering removing the line that gives warnings about them for now (until I test the potentially invalid header analysis thoroughly).

frontends/p4/simplifyDefUse.cpp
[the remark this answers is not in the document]
kostasyrmia: That looks like exactly what I needed.

frontends/p4/simplifyDefUse.cpp
kostasyrmia: That's a great idea!

frontends/p4/simplifyDefUse.cpp
kostasyrmia: I didn't want to put everything in a single PR because I thought it would make the review harder. I agree with your idea, it looks good to me.

frontends/p4/simplifyDefUse.cpp
kostasyrmia: Yes, I agree. This method isn't much more special than other extern methods.

kostasyrmia:
I will add these suggestions, please don't merge until then.

testdata/p4_16_samples_outputs/gauntlet_ctrl_plane_hdr.p4-stderr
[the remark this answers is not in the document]
mbudiu-vmw: I have filed #2895 and I will try to do it myself.

kostasyrmia:
I have added a new commit be38a4b91a712a58af7890444da74db18d6d56e4 with the changes.

kostasyrmia:
> I have added a new commit be38a4b with the changes.

@mbudiu-vmw 
I didn't tag you in the previous comment, this is in case you haven't seen it.

mbudiu-vmw (approved):

PR #2888, opened 2021-08-31
Remove colors from error messages when redirected to a file

msedlace:
This change makes sure that no escape characters for terminal colorizations are present in error logs when standard error output is redirected to a file.

### DIFF
--- a/lib/exceptions.h
+++ b/lib/exceptions.h
@@ -39,7 +39,7 @@ inline bool is_cerr_redirected() {
 
 /// Function used to delete color in case of cerr output being redirected
 inline const char *cerr_colorize(const char *color) {
-    if(is_cerr_redirected()){
+    if (is_cerr_redirected()) {
         return "";
     }
     return color;
@@ -47,7 +47,7 @@ inline const char *cerr_colorize(const char *color) {
 
 /// Used to clear colors on terminal
 inline const char *cerr_clear_colors() {
-    if(is_cerr_redirected()){
+    if (is_cerr_redirected()) {
         return "";
     }
     return ANSI_CLR;
--- a/lib/exceptions.h
+++ b/lib/exceptions.h
@@ -31,8 +31,26 @@ constexpr char ANSI_BLUE[] = "\e[34m";
 constexpr char ANSI_CLR[]  = "\e[0m";
 
 /// Checks if stderr is redirected to a file
+/// Check is done only once and then saved to a static variable
 inline bool is_cerr_redirected() {
-    return ttyname(fileno(stderr)) == nullptr;  // NOLINT(runtime/threadsafe_fn)
+    static bool is_redir = ttyname(fileno(stderr)) == nullptr;  // NOLINT(runtime/threadsafe_fn)
+    return is_redir;
+}
+
+/// Function used to delete color in case of cerr output being redirected
+inline const char *cerr_colorize(const char *color) {
+    if(is_cerr_redirected()){
+        return "";
+    }
+    return color;
+}
+
+/// Used to clear colors on terminal
+inline const char *cerr_clear_colors() {
+    if(is_cerr_redirected()){
+        return "";
+    }
+    return ANSI_CLR;
 }
 
 /// Base class for all exceptions.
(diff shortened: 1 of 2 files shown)


lib/exceptions.h
```diff
@@ -54,27 +56,47 @@ class CompilerBug final : public P4CExceptionBase {
  public:
     template <typename... T>
     CompilerBug(const char* format, T... args)
-            : P4CExceptionBase(format, args...)
-    { message = cstring(ANSI_RED) + "Compiler Bug" + ANSI_CLR + ":\n" + message; }
+            : P4CExceptionBase(format, args...) {
+        // Check if output is redirected and if so, then don't color text so that
+        // escape characters are not present
+        const bool isCerrRedirected = ttyname(fileno(stderr)) == nullptr;  // NOLINT(runtime/threadsafe_fn)
```
mbudiu-vmw, line 62: Please move this code into a helper function, e.g., `colorize`.
You can also probably cache the result of the check.

mbudiu-vmw (changes requested):

lib/exceptions.h
msedlace: Sure thing, I moved it to its own function.

lib/exceptions.h
```diff
@@ -29,6 +30,11 @@ constexpr char ANSI_RED[]  = "\e[31m";
 constexpr char ANSI_BLUE[] = "\e[34m";
 constexpr char ANSI_CLR[]  = "\e[0m";
 
+/// Checks if stderr is redirected to a file
+inline bool is_cerr_redirected() {
```
mbudiu-vmw, line 34: I would suggest having a static bool to save the result after the first call. It won't change later, so you can avoid calling ttyname. It is unlikely that error is called many times, but it is still healthier to save the value.

lib/exceptions.h
```diff
@@ -54,27 +60,40 @@ class CompilerBug final : public P4CExceptionBase {
  public:
     template <typename... T>
     CompilerBug(const char* format, T... args)
-            : P4CExceptionBase(format, args...)
-    { message = cstring(ANSI_RED) + "Compiler Bug" + ANSI_CLR + ":\n" + message; }
+            : P4CExceptionBase(format, args...) {
+        // Check if output is redirected and if so, then don't color text so that
+        // escape characters are not present
+        message = (is_cerr_redirected() ? "" : cstring(ANSI_RED)) + "Compiler Bug"
```
mbudiu-vmw, line 66: I suggested moving the entire expression `(is_cerr_redirected() ? "" : cstring(ANSI_RED))` into a helper function. Perhaps colorize(ANSI_RED) and clear_colors().

mbudiu-vmw (changes requested):

lib/exceptions.h
```diff
@@ -31,8 +31,26 @@ constexpr char ANSI_BLUE[] = "\e[34m";
 constexpr char ANSI_CLR[]  = "\e[0m";
 
 /// Checks if stderr is redirected to a file
+/// Check is done only once and then saved to a static variable
 inline bool is_cerr_redirected() {
-    return ttyname(fileno(stderr)) == nullptr;  // NOLINT(runtime/threadsafe_fn)
+    static bool is_redir = ttyname(fileno(stderr)) == nullptr;  // NOLINT(runtime/threadsafe_fn)
```
mbudiu-vmw, line 36: you are not saving anything, the call to ttyname is still made every time.
you need to also remember whether is_redir has been initialized, using a second static variable.

lib/exceptions.h
```diff
@@ -31,8 +31,26 @@ constexpr char ANSI_BLUE[] = "\e[34m";
...
+    return is_redir;
+}
+
+/// Function used to delete color in case of cerr output being redirected
+inline const char *cerr_colorize(const char *color) {
+    if(is_cerr_redirected()){
```
mbudiu-vmw, line 42: we leave spaces in ifs.

mbudiu-vmw (changes requested):

mbudiu-vmw (approved):

Issue #2893, opened 2021-09-02
P4C-DPDK: Egress does not generate tables and their actions


Paxyk:
Hello,
As stated in the title, egress does not generate tables and their actions. Same happens when you try to use user defined control either in ingress or egress. Sample application with this bug is attached. 
[issue_egress.zip](https://github.com/p4lang/p4c/files/7097969/issue_egress.zip)

rst0git:
@Paxyk Thank you for opening this issue. I was able to replicate the problem and opened a PR with a fix: https://github.com/p4lang/p4c/pull/2894

PR #2894, opened 2021-09-02
dpdk: export egress tables and actions

rst0git:
This commit appends tables and actions from the egress pipeline to be included in the output spec file.

Fixes: #2893


mbudiu-vmw (changes requested):
Please add the test case from #2893 and all the corresponding outputs generated by the compiler.

rst0git:
@mbudiu-vmw I've added a test case with the corresponding output generated by the compiler. The test is expected to fail before the fix is applied and pass after.

hanw:
Looks good to me, thanks for the fix.

hanw (approved):

mbudiu-vmw (approved):

PR #2902, opened 2021-09-17
Rebased version of #1698

mbudiu-vmw:
This is another attempt to fix the problems with recirculate/resubmit/clone3.
The solution is to 
- use an annotation (called `@field_list` currently) on the user metadata fields that need to be recirculated
- the annotation has a numeric argument
- pass the annotation numeric argument as the last argument to the recirculate/etc call
This will cause the runtime to initialize the specified fields before entering the pipeline.
This correctly implements the semantics of P4-14 recirculate/etc.
Here is an excerpt from the `resubmit.p4` p4-14 benchmark, as automatically translated by the compiler:
```
enum bit<8> FieldLists {
    resubmit_FL = 0
}

struct mymeta_t {
    @field_list(FieldLists.resubmit_FL) 
    bit<8> f1;
}

...
    @name("._resubmit") action _resubmit() {
        meta.mymeta.f1 = 8w1;
        resubmit((bit<8>)FieldLists.resubmit_FL);
    }
```

This changes the signature of the respective v1model functions. We may chose instead to add new functions and deprecate the old ones.


smolkaj:
Awesome, thanks. I agree that we would want to instead add new functions and deprecate the old ones.
Maybe something like `clone3_with_field_list`?

mbudiu-vmw:
@smolkaj can you please test to see whether this variant works for you?

smolkaj:
Thanks, @mbudiu-vmw, I will give it a try!

kheradmandG:
Question, how can we specify not passing any fields to recirculate/resubmit/clone3. I.e. the equivalent of `clone3(…, …, {}) `? 
I assume we should use a number that is not already used in any `@field_list`. If true, isn’t that a bit problematic? Say one uses number 0 to specify that `clone3_field_list(…, …, 0)` and later end up using that number in a @field_list, forgetting that 0 was used as an equivalent of specifying nothing (`{}`). This can lead to subtle bugs.

mbudiu-vmw:
Yes, just use a number that has no fields attached. You could use -1, that's less likely to be a common number.
I expect people don't frequently refactor programs in this way.

kheradmandG:
Thanks for the response @mbudiu-vmw. @smolkaj also mentioned that there the `clone` primitive that can be used for this purpose.

p4include/v1model.p4
```diff
@@ -582,17 +582,12 @@ extern void resubmit<T>(in T data);
...
-
-/***
- * clone is in most ways identical to the clone3 operation, with the
- * only difference being that it never preserves any user-defined
- * metadata fields with the cloned packet.  It is equivalent to
- * calling clone3 with the same type and session parameter values,
- * with empty data.
- */
+extern void recirculate_field_list(bit<8> index);
 extern void clone(in CloneType type, in bit<32> session);
```
kheradmandG, line 586: There is another declaration of `extern void clone(in CloneType type, in bit<32> session)` in line 537. This causes compilation errors: 

```
 [--Werror=duplicate] error: clone: multiple matching declarations
...
v1model.p4(537): [--Werror=duplicate] error: Candidate: clone
extern void clone(in CloneType type, in bit<32> session);
            ^^^^^
v1model.p4(586): [--Werror=duplicate] error: Candidate: clone
extern void clone(in CloneType type, in bit<32> session);
            ^^^^^
```

mbudiu-vmw: how come our tests pass?

kheradmandG: Not sure what flags/options we are using for the compilation (maybe @smolkaj can shed some light?). Anyway that declaration is indeed duplicated in v1model.p4 in this pull request (lines 537 and 586).

mbudiu-vmw: I guess if you never call the method you don't get the error.
I will remove one of the two declarations, in the meantime can you please do the same if you want to try this?

kheradmandG: That makes sense, also maybe somebody should add a test for `clone` to p4c's test suite to avoid future regressions. Yeah I commented out that line and the error goes away.

kheradmandG:
One more question, I have something like:
```
enum bit<8> PreservedFieldList {
  CLONE_I2E = 8w1
};

struct metadata_t {
  ...
  @field_list(PreservedFieldList.CLONE_I2E)
  ipv4_addr_t ip;
  ...
}
```

I get the error:
```
[--Werror=not-found] error: PreservedFieldList: declaration not found
  @field_list(PreservedFieldList.CLONE_I2E)
              ^^^^^^^^^^^^^^^^^^
```

However, if I change `metadata_t` to:
```
struct metadata_t {
  ...
  @field_list(8w1)
  ipv4_addr_t ip;
  ...
}
``` 

I was wondering whether I'm missing something here.

mbudiu-vmw:
This is a bit strange, because this exactly the kind of code that is generated by the P4-14 front-end.
It must be a bug related to name resolution.
Can you please post a complete example and I will make sure it compiles and add it as a test?

kheradmandG:
I randomly selected `testdata/p4_14_samples_outputs/packet_redirect-frontend.p4` and ran into the same problem. Do you run that test successfully?

mbudiu-vmw:
If you look at this web page you will see that all tests have passed. I will give it a shot.

castilhorosa:
Sorry if this is not the right place for this kind of question but I'd like to get a sneak peek of this version of p4c, with the extern `resubmit_field_list,` and I wondering if there is any script that I can use to compile this branch. I've been trying unsuccefully to adapt the script `install-p4dev-v2.sh` by using the command 
``git clone -b recirculate --single-branch https://github.com/mbudiu-vmw/p4c-clone.git
``
I would really appreciate if someone could provide me such script/commands...

mbudiu-vmw:
Can you already build the main branch of the compiler?
Is yes, then the only thing you need to do is to switch to this branch (git checkout -b recirculate) and then run the same steps.
For me the steps are `cmake ..; make`

mbudiu-vmw:
Don't use my clone of the repository, use the official one from https://github.com/p4lang/p4c
My clone is just for development and the main branch may be behind the origin.

castilhorosa:
> Can you already build the main branch of the compiler? Is yes, then the only thing you need to do is to switch to this branch (git checkout -b recirculate) and then run the same steps. For me the steps are `cmake ..; make`

Great. It worked. Thanks @mbudiu-vmw

kheradmandG:
@mbudiu-vmw @smolkaj 
I did some internal testing and seems like (beside the 2 issues mentioned above), the pull request works as expected. 
A suggestion: would be good idea to name the the action with something that conveys that the field list is being preserved. E.g. `clone_preserve_field_list`.

smolkaj:
+1, how about `clone_preserving_field_list` to make it roll off the tongue a bit easier?

smolkaj:
(As you might be able to tell, Google has trained me to advocate for long, descriptive function names... 😆)

mbudiu-vmw:
I will submit soon an update which fixes the issues you discussed. There is a new test using an enum, and that passes too, I could not reproduce Ali's error. Make sure that the enum is declared prior to its first use.

mbudiu-vmw:
Take it back, found and solved the problem with the enums too.
@smolkaj @kheradmandG please take another look.

kheradmandG:
@mbudiu-vmw Thanks for the changes. One question regarding the naming: what does `3` in `clone3` mean? And does it makes sense to keep it in `clone3_preserving_field_list` ?

mbudiu-vmw:
It comes from the fact that there was a clone with 2 arguments, and one with 3.

kheradmandG:
@mbudiu-vmw Thanks for the info. nit: now that this action is completely distinguishable from the clone with 2 arguments, would it make sense to rename it to `clone_preserving_field_list` (i.e. drop `3`)?

mbudiu-vmw:
I will make the change to clone3, but we won't be able to merge this until someone approves it.
I asked @jafingerhut for a review. Hopefully he can find the time.

control-plane/p4RuntimeSerializer.cpp
```diff
@@ -729,6 +729,10 @@ getTypeWidth(const IR::Type* type, TypeMap* typeMap) {
         // W if the type is bit<W>, and 0 if the type is string
         return annotation.controller_type.width;
     }
+    /* Treat error type as string */
+    if (type->is<IR::Type_Error>())
+        return 0;
+
```
jfingerh, line 735: I'm not reviewing the C++ implementation code parts of this PR very carefully -- simply scanning through them looking for anything odd.  Is this change something that was discovered as a small side issue while developing the fixes for preserving user-defined metadata fields for resubmit/recirculate/clone?  Perhaps something that was needed for the failing test cases exercising those features to pass, so good to combine them into this PR while you were at it?  I don't have any objection, main curious why it is thrown in with this PR.

frontends/p4/fromv1.0/programStructure.cpp
```diff
@@ -2588,9 +2675,12 @@ void ProgramStructure::populateOutputNames() {
         "mark_to_drop",
         "hash",
         "resubmit",
+        "resubmit_preserving_field_list",
```
jfingerh, line 2678: I will double-check this with my own small tests, but is resubmit_preserving_field_list a new primitive internal to p4c only, or do these changes also require new primitives in the BMv2 JSON file that BMv2 reads?

mbudiu-vmw: This is compiled to the standard recirculare bmv2 primitive

p4include/v1model.p4
```diff
@@ -537,43 +528,65 @@ extern void verify_checksum_with_payload<T, O>(in bool condition, in T data, in
...
  * documentation (Note 1) for more details.
+ *
+ * For example, the user metadata fields can be annotated as follows:
+ * struct UM {
+ *    @field_list(1)
+ *    bit<32> x;
+ *    @field_list(1, 2)
+ *    bit<32> y;
+ *    bit<32> z;
+ * }
```
jfingerh, line 568: Thanks for adding a short example to this documentation!

testdata/p4_14_samples_outputs/acl1-midend.p4
```diff
@@ -134,16 +134,27 @@ header data_t {
 }
 
 struct metadata {
+    @pa_solitary("ingress", "acl_metadata.if_label")
```
jfingerh, line 137: How were tests passing before without these annotations?  The input program had a pragma or P4_14 annotation-like thing since before this PR, so did something change in this PR to make these annotations appear in the output?  Or were the tests somehow passing before even though the files did not match?

mbudiu-vmw: This is probably because I rebased on top of other changes. This pr does not know anything about this annotation. I don't know where it's coming from.

testdata/p4_14_samples_outputs/copy_to_cpu-first.p4
```diff
@@ -73,7 +77,7 @@ control egress(inout headers hdr, inout metadata meta, inout standard_metadata_t
 
 control ingress(inout headers hdr, inout metadata meta, inout standard_metadata_t standard_metadata) {
     @name(".do_copy_to_cpu") action do_copy_to_cpu() {
-        clone3<tuple<standard_metadata_t>>(CloneType.I2E, 32w250, { standard_metadata });
+        clone_preserving_field_list(CloneType.I2E, 32w250, (bit<8>)FieldLists.copy_to_cpu_fields);
```
jfingerh, line 80: Does this implementation work to preserve standard metadata fields?  If yes, how does that work, if one cannot add annotations to the standard metadata struct definition?

jfingerh: Regardless of the answer, it seems worth adding a sentence or three on this in the v1model.p4 comments that document how this works.

mbudiu-vmw: You cannot. The idea is that you should not be able to spoof standard Metadata. It's debatable, but I don't know how to do it anyway.

testdata/p4_16_samples/issue1653-bmv2.p4
```diff
@@ -43,10 +43,9 @@ control ingress(inout parsed_packet_t h, inout local_metadata_t local_metadata,
     bitvec_hdr bh;
 
     apply {
-
         // Copy another header's data to local variable.
         bh.x = h.bvh0.x;
-        clone3(CloneType.I2E, 0, h);
+        clone_preserving_field_list(CloneType.I2E, 0, 0);
```
jfingerh, line 48: It seems like if we wanted to remain faithful to the original meaning of this test program, there should be `@field_list(0)` annotations on all fields of struct `parsed_packet_t`?

frontends/p4/fromv1.0/programStructure.cpp
jfingerh: Sounds good.  Thanks.

p4include/v1model.p4
jfingerh: I created a separate issue to remind me to either myself, or find some other volunteer to help, update other documentation inside and outside of the p4lang/p4c repository on how these operations work.  I think it is best of those changes are made in their own separate PRs after this one is merged: https://github.com/p4lang/p4c/issues/2977

testdata/p4_14_samples_outputs/acl1-midend.p4
jfingerh: OK, I will not worry about it then.  There is a `pa_solitary` pragma in the original P4_14 program, so something in the P4_14 to P4_16 translation must be picking it up.

testdata/p4_14_samples_outputs/copy_to_cpu-first.p4
jfingerh: Seems like a reasonable restriction, and I will put it on a list of details to document elsewhere, including a suggested workaround that of course one is welcome to copy a standard metadata field value into a user-defined metadata field, and preserve the user-defined metadata field.

jfingerh (approved):
Looks good to me.  As usual, I have focused on the v1model.p4 include file changes, and the changes to the test programs and their expected outputs.  Except for one minor comment where it appears that a test program may have changed its behavior by no longer preserving some fields that the original did, I did not notice any problems.

testdata/p4_16_samples/issue1653-bmv2.p4
mbudiu-vmw: The clone only preserves metadata fields,. Adding annotations to parsed_packet_t should have no effect.

control-plane/p4RuntimeSerializer.cpp
mbudiu-vmw: I think that this is coming from the main branch from a merge.

mbudiu-vmw:
@jafingerhut if I have your approval please give it officially so we can merge this.
Or suggest alternative reviewers...

jfingerh:
jafingerh is also my user account, and I approved using that one already.  I can add approval using my account jafingerhut in a few minutes, in case that makes a difference for tracking purposes.

jafingerhut (approved):
Looks good to me. As usual, I have focused on the v1model.p4 include file changes, and the changes to the test programs and their expected outputs. Except for one minor comment where it appears that a test program may have changed its behavior by no longer preserving some fields that the original did, I did not notice any problems.

PR #2907, opened 2021-09-23
dpdk: initialize member_id and group_id for action profile and selector

hanw:


backends/dpdk/dpdkArch.cpp
```diff
@@ -1265,8 +1268,23 @@ const IR::Node* SplitP4TableCommon::postorder(IR::MethodCallStatement *statement
...
+            auto member_id = member_ids.at(tableName);
+            decls->push_back(new IR::AssignmentStatement(
+                        new IR::PathExpression(member_id),
+                        new IR::Constant(IR::Type_Bits::get(32), 0))); }
+
+        if (group_ids.count(tableName) != 0) {
+            auto group_id = group_ids.at(tableName);
+            decls->push_back(new IR::AssignmentStatement(
+                        new IR::PathExpression(group_id),
+                        new IR::Constant(IR::Type_Bits::get(32), 0))); }
```
mbudiu-vmw, line 1284: always 0?

mbudiu-vmw (approved):

backends/dpdk/dpdkArch.cpp
hanw: always initialized to zero, a subsequent table will set it based on control plane information.

PR #2909, opened 2021-09-27
DPDK Backend: Fix verify statement support and few other minor issues

usha1830:
This PR fixes the following issues:

1) Handled boolean literals in verify statement
2) Moving error id from verify method to standard metadata for parser error
3) Emitting missing Label before drop instruction at the end.
4) Change warning type for mismatched header fields
5) Change cast to mov instruction as it is not a valid instruction.

### DIFF
--- a/backends/dpdk/dpdkHelpers.cpp
+++ b/backends/dpdk/dpdkHelpers.cpp
@@ -21,15 +21,6 @@ namespace DPDK {
 
 int ConvertStatementToDpdk::next_label_id = 0;
 
-IR::Declaration_Variable *ConvertStatementToDpdk::addNewTmpVarToMetadata(cstring name,
-                                                                         const IR::Type* type) {
-    auto newTmpVar =  new IR::Declaration_Variable(IR::ID(
-                                                   refmap->newName(name)), type);
-    metadataStruct->fields.push_back(new IR::StructField(IR::ID
-                                     (newTmpVar->name.name), newTmpVar->type));
-    return newTmpVar;
-}
-
 // convert relation comparison statements into the corresponding branching
 // instructions in dpdk.
 void ConvertStatementToDpdk::process_relation_operation(const IR::Expression* dst,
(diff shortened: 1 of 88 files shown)


backends/dpdk/dpdkHelpers.cpp
```diff
@@ -619,11 +628,20 @@ bool ConvertStatementToDpdk::preorder(const IR::MethodCallStatement *s) {
                 ::error("%1%: must be one of the existing errors", s);
             auto error_id = structure->error_map.at(error->expression->to<IR::Member>()->member);
             auto end_label = Util::printf_format("label_%dend", next_label_id++);
-            add_instr(new IR::DpdkJmpEqualStatement(
-                        end_label,
-                        condition->expression, new IR::BoolLiteral(false)));
-            add_instr(new IR::DpdkMovStatement(
-                        new IR::PathExpression("metadata"), new IR::Constant(error_id)));
+            if (condition->expression->is<IR::BoolLiteral>()) {
```
mbudiu-vmw, line 631: are these two branches of the if swapped?

mbudiu-vmw (changes requested):

backends/dpdk/dpdkHelpers.cpp
usha1830: No. This code is for the case where both the  operands of conditional jump are constant. It moves the first operand to a new temporary variable.
Actually, I think I can remove the conditional branch altogether for this case.

backends/dpdk/dpdkHelpers.h
```diff
@@ -127,12 +127,13 @@ class ConvertStatementToDpdk : public Inspector {
     P4::ReferenceMap *refmap;
     DpdkProgramStructure *structure;
     const IR::P4Parser *parser = nullptr;
+    IR::Type_Struct *metadataStruct;
```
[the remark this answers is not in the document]
usha1830: This structure is not needed anymore and is removed.
Agree with you on keeping the interface clean. Will keep in mind in future.

backends/dpdk/spec.cpp
```diff
@@ -545,9 +545,7 @@ std::ostream &IR::DpdkGetChecksumStatement::toSpec(std::ostream &out) const {
 }
 
 std::ostream &IR::DpdkCastStatement::toSpec(std::ostream &out) const {
-    out << "cast "
-        << " " << DPDK::toStr(dst) << " " << DPDK::toStr(type) << " "
-        << DPDK::toStr(src);
+    out << "mov " << " " << DPDK::toStr(dst) << " " << DPDK::toStr(src);
```
hanw, line 548: nit: there is an extra space after 'mov'

hanw (approved):

backends/dpdk/spec.cpp
usha1830: @hanw Thanks for the review.
Removed extra space and updated reference outputs.

backends/dpdk/dpdkHelpers.cpp
usha1830: @mbudiu-vmw Thanks for the review.. 
Could you please have a look at the updated code and see if it looks fine now?

mbudiu-vmw (approved):

PR #2911, opened 2021-09-30
Silence warnings automatically if @nowarn appears in the context

### LINKED ISSUE #2895: Automatically inhibit warnings if the `@nowarn` annotation is used

mbudiu-vmw:
The base change is relatively easy: a warn function has been added to the visitor, and it checks the context of a node to see if it has a noWarn suitable annotation. 
The more complicated changes come because some visitors create other visitors and invoke them; for this purpose a new field called_by has been added to visitors. The check for warnings thus can span multiple visitors, but the called_by field must be maintained manually.
There may be still cases where this does not work, which will need to be handled manually probably.
Fixes #2895

### DIFF
--- a/ir/visitor.h
+++ b/ir/visitor.h
@@ -57,7 +57,11 @@ class Visitor {
     virtual ~Visitor() = default;
 
     mutable cstring internalName;
+    // Some visitors are created and applied by other visitors.
+    // This field keeps track of the caller.
+    const Visitor* called_by = nullptr;
 
+    const Visitor& setCalledBy(const Visitor* visitor) { called_by = visitor; return *this; }
     // init_apply is called (once) when apply is called on an IR tree
     // it expects to allocate a profile record which will be destroyed
     // when the traversal completes.  Visitor subclasses may extend this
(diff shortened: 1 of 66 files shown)


testdata/p4_16_samples_outputs/uninit-nowarnings.p4-stderr
```diff
@@ -0,0 +1,6 @@
+uninit-nowarnings.p4(73): [--Wwarn=uninitialized_out_param] warning: out parameter 'v' may be uninitialized when 'c' terminates
```
mbudiu-vmw, line 1: note that not all warnings were inhibited, so these should appear here.

hanw (approved):

ir/visitor.h
```diff
@@ -57,7 +57,11 @@ class Visitor {
     virtual ~Visitor() = default;
 
     mutable cstring internalName;
+    // Some visitors are created and applied by other visitors.
+    // This field keeps track of the caller.
+    const Visitor* called_by = nullptr;
 
+    const Visitor& setCalledBy(const Visitor* visitor) { called_by = visitor; return *this; }
```
ChrisDodd, line 64: This addition seems completely unnecessary -- you could just call `apply` with the context as a second argument .  You then just need a simple search up the context.

mbudiu-vmw: Didn't realize you can thread context across visitors. You can file an issue to make this change if you think it's worth the effort

PR #2914, opened 2021-10-01
Elimination of non-concrete indexes for header stacks from controls

VolodymyrPeschanenkoIntel:
These changes contain one additional class for midend which eliminates non-concrete indexes of the header stacks from the controls.


fruffy:
There are no tests for this. I believe it might be worthwhile to add this to the p4test back end.

VolodymyrPeschanenkoIntel:
> There are no tests for this. I believe it might be worthwhile to add this to the p4test back end.

Tests were placed in testdata/p4_16_samples : [control-hs-index-test1.p4](https://github.com/VolodymyrPeschanenkoIntel/p4c/blob/LitSoft/hs-index-control-unroll/testdata/p4_16_samples/control-hs-index-test1.p4), [control-hs-index-test2.p4](https://github.com/VolodymyrPeschanenkoIntel/p4c/blob/LitSoft/hs-index-control-unroll/testdata/p4_16_samples/control-hs-index-test2.p4)

The corresponded outputs are here: testdata/p4_16_samples_outputs/parser-unroll

Midend path for p4test was updated [here](https://github.com/VolodymyrPeschanenkoIntel/p4c/blob/f0e8919d53741d86624eb3455b33eb427e7c52b3/backends/p4test/midend.cpp#L145).

fruffy:
So it would be great if this could be run on the entire P4-16 test suite, so that we get more coverage. Currently, we only run two tests with this pass activated.

midend/hsIndexSimplify.h
```diff
@@ -0,0 +1,46 @@
...
+#define _MIDEND_HSINDEXSIMPLIFY_H_
+
+#include "frontends/common/resolveReferences/referenceMap.h"
+#include "frontends/p4/typeChecking/typeChecker.h"
+#include "frontends/p4/typeMap.h"
+#include "ir/ir.h"
+
+namespace P4 {
+
+/// The main class for finding of non-concrete haader stack/arrya index.
```
fruffy, line 11: I would write a proper description here. For an example program. what is the before and after? What is this pass meant to do?

backends/p4test/midend.cpp
```diff
@@ -141,6 +142,7 @@ MidEnd::MidEnd(CompilerOptions& options, std::ostream* outStream) {
         new P4::SynthesizeActions(&refMap, &typeMap, new SkipControls(v1controls)),
         new P4::MoveActionsToTables(&refMap, &typeMap),
         options.loopsUnrolling ? new P4::ParsersUnroll(true, &refMap, &typeMap) : nullptr,
+        options.hsIndexSimplify ? new P4::HSIndexSimplifier(&refMap, &typeMap) : nullptr,
```
fruffy, line 145: I personally think to properly test it, this pass should be used on the entire test suite. We do not have the same restrictions as with loops unrolling.

midend/hsIndexSimplify.cpp
```diff
@@ -0,0 +1,127 @@
...
+#include <sstream>
+
+namespace P4 {
+
+const IR::Node* HSIndexFinder::postorder(IR::ArrayIndex* curArrayIndex) {
+    if (curArrayIndex->right->is<IR::Constant>()) {
+        return exprIndex2Member(curArrayIndex->type, curArrayIndex->left,
+                                curArrayIndex->right->checkedTo<IR::Constant>());
+    }
+    // Finding the first occurrence of non-concrete array index.
```
fruffy, line 13: Find*

midend/hsIndexSimplify.cpp
```diff
@@ -0,0 +1,127 @@
...
+IR::Node* HSIndexSimplifier::preorder(IR::MethodCallStatement* methodCallStatement) {
+    auto* newMethodCall = methodCallStatement->clone();
+    HSIndexFinder aiRewriter(nullptr, 0);
+    newMethodCall->methodCall =
+        newMethodCall->methodCall->apply(aiRewriter)->to<IR::MethodCallExpression>();
+    return newMethodCall;
+}
+
+IR::Node* HSIndexSimplifier::preorder(IR::SelectExpression* selectExpression) {
+    // All non-concrete indexes should be eliminated by ParserUnroll.
```
fruffy, line 118: Is this a hard requirement? If that is not the case, a BUG should be thrown here.

midend/hsIndexSimplify.cpp
```diff
@@ -0,0 +1,127 @@
...
+                                                      const IR::Expression* expression,
+                                                      const IR::Constant* constant) {
+    return new IR::ArrayIndex(type, expression, constant);
+}
+
+IR::Node* HSIndexSimplifier::eliminateArrayIndexes(IR::Statement* statement) {
+    // Checking non-concrete array indexes.
+    HSIndexFinder aiFinder;
+    const IR::Statement* elseBody = nullptr;
+    bool isIf = false;
```
fruffy, line 43: Rewrite this to const auto* isIf = statement->is<IR::IfStatement>() and use that to check.

frontends/common/options.cpp
```diff
@@ -154,6 +154,13 @@ CompilerOptions::CompilerOptions() : ParserOptions() {
             return true;
         },
         "Unrolling all parser's loops");
+    registerOption(
```
fruffy, line 157: I am not sure if adding an extra option is a good either. Either this should run as part of the default suite, or right after parserUnroll, if it depends on it.

midend/hsIndexSimplify.cpp
fruffy: Or simply do nothing and do not convert the index here.

fruffy (changes requested):

frontends/common/options.cpp
```diff
@@ -154,6 +154,13 @@ CompilerOptions::CompilerOptions() : ParserOptions() {
...
+        "--hsIndexSimplify", nullptr,
+        [this](const char*) {
+            hsIndexSimplify = true;
+            return true;
+        },
+        "Replace all non-cocrete indexes for header stacks in controls");
```
mbudiu-vmw, line 163: This help message has a typo. But more importantly, I don't think it is clear.
Can you be more explicit?

midend/hsIndexSimplify.h
```diff
@@ -0,0 +1,46 @@
...
+class HSIndexFinder : public Transform {
```
mbudiu-vmw, line 12: A "finder" implies an Inspector.
This class must be doing more since it is a Transform. This suggests that it is not named properly.

midend/hsIndexSimplify.h
```diff
@@ -0,0 +1,46 @@
...
+    friend class HSIndexSimplifier;
```
mbudiu-vmw, line 13: Please document these fields.

midend/hsIndexSimplify.cpp
```diff
@@ -0,0 +1,127 @@
...
+
+size_t HSIndexFinder::getArraySize() {
+    const auto* typeStack = arrayIndex->left->type->checkedTo<IR::Type_Stack>();
+    return typeStack->getSize();
+}
+
+const IR::ArrayIndex* HSIndexFinder::exprIndex2Member(const IR::Type* type,
+                                                      const IR::Expression* expression,
+                                                      const IR::Constant* constant) {
+    return new IR::ArrayIndex(type, expression, constant);
```
mbudiu-vmw, line 36: you are losing the source code information srcInfo. This can lead to poor error messages.
I frankly do not understand the name of this function, there is no member involved.
Do you really need this function? IT is a one line, it would be clearer to inline it.

midend/hsIndexSimplify.h
```diff
@@ -0,0 +1,46 @@
...
+    HSIndexFinder(const IR::ArrayIndex* arrayIndex, int index)
+        : arrayIndex(arrayIndex), isFinder(false), index(index) {}
+    const IR::Node* postorder(IR::ArrayIndex* curArrayIndex) override;
+    size_t getArraySize();
+    static const IR::ArrayIndex* exprIndex2Member(const IR::Type* type,
+                                                  const IR::Expression* expression,
+                                                  const IR::Constant* constant);
+};
+
+/// The main class for concretization of array index.
```
mbudiu-vmw, line 29: This comment is not very helpful.

mbudiu-vmw (changes requested):
I will re-review this once it is a bit better documented.

VolodymyrPeschanenkoIntel:
I have added one test with nested header stack and methodCallStatement with header stack. I have updated p4c test with header stacks. So, I think that the code is ready for the next review.

midend/hsIndexSimplify.cpp
```diff
@@ -0,0 +1,116 @@
...
+IR::Node* HSIndexSimplifier::preorder(IR::IfStatement* ifStatement) {
+    return eliminateArrayIndexes(ifStatement);
+}
+
+IR::Node* HSIndexSimplifier::preorder(IR::MethodCallStatement* methodCallStatement) {
+    return eliminateArrayIndexes(ifStatement);
+}
+
+IR::Node* HSIndexSimplifier::preorder(IR::SelectExpression* selectExpression) {
+    // Ignore SelectExpression.
```
fruffy, line 107: If the select expression has a header stack, I would throw a warning that it was not resolved and that the parserUnroll pass should be used for that.

testdata/p4_16_samples_outputs/control-hs-index-test2.p4
```diff
@@ -0,0 +1,66 @@
+#include <core.p4>
```
fruffy, line 1: These tests are a little flawed. First, the index should not be part of the ethernet stack. Use h.h.a instead. Also you should extract a in the parser. That is missing. Here is the setup I would use for all the tests you added. 

```p4
header ethernet_t {
    bit<48> dst_addr;
    bit<48> src_addr;
    bit<16> eth_type;
}

header h_stack {
    bit<8> a;
}

header h_index {
    bit<32> i;
}

struct headers {
    ethernet_t eth_hdr;
    h_stack[3] h;
    h_index    i;
}
```

with extract like this:

```p4
pkt.extract(hdr.eth_hdr);
pkt.extract(hdr.h[0]);
pkt.extract(hdr.h[1]);
pkt.extract(hdr.h[2]);
pkt.extract(hdr.i);
```

fruffy (changes requested):

fruffy:
You also need to sign the CLA.

mbudiu-vmw:
I think that the right way to do this is to actually factor out any array index expression with non-constant index into a temporary assignment and then handle only this case in your code. This works for expressions that "read" an array index on the RHS. For assignments on the LHS this is less of a problem. And there is also `isValid()`, which should be handled probably in the same way.

VolodymyrPeschanenkoIntel:
> I think that the right way to do this is to actually factor out any array index expression with non-constant index into a temporary assignment and then handle only this case in your code. This works for expressions that "read" an array index on the RHS. For assignments on the LHS this is less of a problem. And there is also `isValid()`, which should be handled probably in the same way.

@mbudiu-vmw, @fruffy. I got some problem wit understanding of the right way to create a local variable.  I create new variable [here](https://github.com/VolodymyrPeschanenkoIntel/p4c/blob/8d176ffaba0e09ddad673801750aa0f77bec637d/midend/hsIndexSimplify.cpp#L21) and push it to a BlockStatement [here](https://github.com/VolodymyrPeschanenkoIntel/p4c/blob/8d176ffaba0e09ddad673801750aa0f77bec637d/midend/hsIndexSimplify.cpp#L134). When I tried to call HSIndexSimplifier [here](https://github.com/VolodymyrPeschanenkoIntel/p4c/blob/8d176ffaba0e09ddad673801750aa0f77bec637d/backends/p4test/midend.cpp#L142). The following error apear [here](https://github.com/VolodymyrPeschanenkoIntel/p4c/blob/8d176ffaba0e09ddad673801750aa0f77bec637d/frontends/common/resolveReferences/resolveReferences.cpp#L200). My question is how to add local variable for a BlockStatement in a correct way?

fruffy:
This is the transformation:

```p4
control ingress(inout headers hdr, inout test_metadata_t meta, inout standard_metadata_t standard_metadata) {
    apply {
        hdr.ethernet_stack[1] = hdr.ethernet_stack[meta.color];
    }
}

```

```p4
control ingress(inout headers hdr, inout test_metadata_t meta, inout standard_metadata_t standard_metadata) {
    apply {
        if (meta.color == 32w0) {
            hdr.ethernet_stack[1] = hdr.ethernet_stack[0];
        } else if (meta.color == 32w1) {
            hdr.ethernet_stack[1] = hdr.ethernet_stack[1];
        }
    }
}
```
So the issue is that the simplifier discards the case where the enum access is out of bounds. In that case, hdr.ethernet_stack[1] may read from random memory and the behavior is undefined. I am not sure whether this transformation is considered a refinement or a bug. Ultimately, it removes the undefined behavior from the program.

fruffy:
> @mbudiu-vmw, What kind of approval we need to wait?

Github has added an restriction that first-time contributors need their workflows approved. This is to avoid crypto mining and other abuse. I have to manually start the tests, which is a little cumbersome. Maybe we should disable this restriction.

mbudiu-vmw:
You mean, in the else case leave the index be a dynamic value?
I am afraid that will defeat the purpose of this optimization.

jnfoster:
I'm not sure I agree. If the compiler front-end can rewrite P4 code that uses run-time data with indexes into equivalent code that only uses compile-time known data with indexes, why should the backend care?

It seems to me that the front-end should be allowed to do any semantics-preserving transformation it wants to...

midend/hsIndexSimplify.cpp
```diff
@@ -0,0 +1,221 @@
...
+
+IR::Node* HSIndexSimplifier::eliminateArrayIndexes(IR::Statement* statement) {
+    if (ignoreParser)
+        return statement;
+    // Check non-concrete array indexes.
+    HSIndexFindOrTransform aiFinder(locals, refMap, typeMap);
+    const IR::Node* updatedStatement = nullptr;
+    const IR::Statement* elseBody = nullptr;
+    auto* curIf = statement->to<IR::IfStatement>();
+    if (curIf != nullptr) {
```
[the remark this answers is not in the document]
mbudiu-vmw: That's why we have visitors -- to dispatch on statement types.
Your code probably works, but it is not in the spirit of the visitor pattern.
Why not break this code into parts and inline them in place where you called this function?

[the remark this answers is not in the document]
mbudiu-vmw: I.e., the `if` should really be handled in the `HSIndexSimplifier::preorder(IR::IfStatement* ifStatement)` function.

midend/hsIndexSimplify.h
```diff
@@ -0,0 +1,72 @@
...
+#include "frontends/common/resolveReferences/referenceMap.h"
+#include "frontends/p4/typeChecking/typeChecker.h"
+#include "frontends/p4/typeMap.h"
+#include "ir/ir.h"
+
+namespace P4 {
+
+/// This class finds a header stack with non-concrete index 
+/// which should be eliminated in unfolding if @a isFinder is true.
+/// Otherwise it substitutes index of a header stack in all occuarence of @a arrayIndex.
```
mbudiu-vmw, line 13: typo in comment.

midend/hsIndexSimplify.h
```diff
@@ -0,0 +1,72 @@
...
+class HSIndexFindOrTransform : public Transform {
```
mbudiu-vmw, line 14: I see no benefit for this class to do double duty. Why don't you break it into two classes, an Inspector and a Transform.
They can communicate via a struct that keeps the fields that need to be used in common. This pattern is used a lot in the compiler.

mbudiu-vmw (changes requested):
I have not finished reviewing the code, please address these comments before I make another pass.
Also, please add a test where expressions using dynamic array indexes are used as function arguments. In particular, I am interested in having them used as in/out/inout arguments.

fruffy:
This time I believe the validator is wrong. This is looking good now.

midend/hsIndexSimplify.h
fruffy: Are these fields documented now?

midend/hsIndexSimplify.cpp
```diff
@@ -0,0 +1,201 @@
...
+            newIf = newStatement->to<IR::IfStatement>()->clone();
+            newIf->condition = new IR::LAnd(newCondition, newIf->condition);
+        } else {
+            newIf = new IR::IfStatement(newCondition, newStatement, nullptr);
+        }
+        if (result == nullptr) {
+            result = newIf;
+            curResult = newIf;
+        } else {
+            curResult->ifFalse = newIf;
```
fruffy, line 92: Is there a point to this assignment? curResult is always set to newIf. I think you just need the nullptr check for `.result`

midend/hsIndexSimplify.cpp
```diff
@@ -0,0 +1,201 @@
...
+    IR::IfStatement* newIf = nullptr;
+    const auto* typeStack = aiFinder.arrayIndex->left->type->checkedTo<IR::Type_Stack>();
+    size_t sz = typeStack->getSize();
+    for (size_t i = 0; i < sz; i++) {
+        HSIndexTransform aiRewriter(aiFinder, i);
+        auto* newStatement = statement->apply(aiRewriter)->to<IR::Statement>();
+        auto* newIndex = new IR::Constant(aiFinder.arrayIndex->right->type, i);
+        auto* newCondition = new IR::Equ(aiFinder.newVariable, newIndex);
+        if (curIf != nullptr) {
+            newIf = newStatement->to<IR::IfStatement>()->clone();
```
fruffy, line 83: Make this a `checkedTo()` since this will fail with a nullptr access if newStatement is not an IfStatement.

midend/hsIndexSimplify.cpp
```diff
@@ -0,0 +1,201 @@
...
+        if (result == nullptr) {
+            result = newIf;
+            curResult = newIf;
+        } else {
+            curResult->ifFalse = newIf;
+            curResult = newIf;
+        }
+    }
+    if (expr != nullptr && locals != nullptr) {
+        // Add case for out of bound.
```
fruffy, line 97: Out of bond what, read or write?

fruffy (commented):
Just a couple small comments. Functionally, it does look better now. Regarding the behavior around index access out of range we should document that this pass should only be used if you do not want to reject access out of bounds. Alternatively, you can add a condition to the pass where it throws an error if that is the case.

midend/hsIndexSimplify.cpp
VolodymyrPeschanenkoIntel: I think that it is correct, because we always need to keep innermost if-then-else and result should keep outward if-then-else.

fruffy: `curResult = newIf;`
You have this assignment in both branches, so does `curResult->ifFalse = newIf;` even do anything?

fruffy: curResult->ifFalse = newIf;

is pointless, isn't it?

fruffy (approved):
Not much more to add from me for now. I will fuzz this once I get the time.

fruffy:
The tests are failing because of mismatched ref files.

VolodymyrPeschanenkoIntel:
@fruffy, It looks like we forgot about elimination of the indexes inside keys of the tables. So, I added new test and made some changes.

fruffy:
Good catch, but I think the changes may be a little too aggressive. It looks like any key with a stack is caught by this.

VolodymyrPeschanenkoIntel:
> Good catch, but I think the changes may be a little too aggressive. It looks like any key with a stack is caught by this.

It works for any keys where are ArrayIndex has non constant index. For that case this class should unroll this non-concrete index.

VolodymyrPeschanenkoIntel:
> Good catch, but I think the changes may be a little too aggressive. It looks like any key with a stack is caught by this.

Fixed

fruffy:
Okay, so I fixed the semantics in Gauntlet (it was an overflow bug) and fuzzed this pull request. I did not find any issues, but may have founds some problems related to the `GlobalCopyPropagation` pass. I will check those and file an issue some time this week if there is indeed a bug.

fruffy (approved):

fruffy (commented):
@mbudiu-vmw From my point of view, all the changes are addressed. Any issues that are missing from your side?

midend/hsIndexSimplify.h
```diff
@@ -0,0 +1,96 @@
...
+#define _MIDEND_HSINDEXSIMPLIFY_H_
+
+#include "frontends/common/resolveReferences/referenceMap.h"
+#include "frontends/p4/typeChecking/typeChecker.h"
+#include "frontends/p4/typeMap.h"
+#include "ir/ir.h"
+
+namespace P4 {
+
+typedef std::map<cstring, const IR::PathExpression*> generatedVariablesMap;
```
mbudiu-vmw, line 11: we use capitalized names for all types.

midend/hsIndexSimplify.h
```diff
@@ -0,0 +1,96 @@
...
+
+ public:
+    HSIndexTransform(HSIndexFinder& finder, int index) : index(index), hsIndexFinder(finder) {}
+    const IR::Node* postorder(IR::ArrayIndex* curArrayIndex) override;
+};
+
+/// This class eliminates all non-concrete indexes of the header stacks in the controls.
+/// It generates new variables for all expressions in the header stacks indexes and
+/// checks their values for substitution of concrete values.
+/// Restriction : in/out parameters should be replaced by correspondent assignments.
```
mbudiu-vmw, line 57: you have another precondition, which is that array variable names are really unique within a scope (there is no aliasing).
I think you should list that. This is ensured by prior passes. 
So you won't have `{ H[2] a; { H[3] a; ... } }`

mbudiu-vmw (approved):
I am approving this, but perhaps you want to address the two minor comments before merging.

mbudiu-vmw:
I will merge this, if you plan to do the suggested fixes perhaps you can bundle them in a subsequent PR.

PR #2915, opened 2021-10-04
Bool to bit<1> cast 

vladyslav-dubina:
Solve problems with errors: `expected an integer` and  `expected a boolean`. 
Errors caused by cast bool to bit<1>.
For example test: https://github.com/p4lang/p4c/blob/main/testdata/p4_16_samples/issue361-bmv2.p4


fruffy:
So the problem is that the parserUnroll pass does not run across the entire p4_16_samples folder, right? Then maybe we should enable it for that, too. @mbudiu-vmw What do you think?

midend/interpreter.cpp
```diff
@@ -735,6 +747,18 @@ void ExpressionEvaluator::postorder(const IR::Operation_Unary* expression) {
         auto li = l->to<SymbolicBool>();
         clone->expr = new IR::BoolLiteral(li->value);
         DoConstantFolding cf(refMap, typeMap);
+        if (auto cast = expression->to<IR::Cast>()) {
+            if (cast->destType->is<IR::Type_Bits>()) {
+                const IR::Constant* constant;
+                if (clone->expr) {
```
fruffy, line 753: What are you checking here? Shouldn't this be value?

midend/interpreter.cpp
```diff
@@ -727,6 +727,18 @@ void ExpressionEvaluator::postorder(const IR::Operation_Unary* expression) {
         auto li = l->to<SymbolicInteger>();
         clone->expr = li->constant;
         DoConstantFolding cf(refMap, typeMap);
+        if (auto cast = expression->to<IR::Cast>()) {
+            if (cast->destType->is<IR::Type_Bits>()) {
```
fruffy, line 731: This might be lacking defensive programming. There is only once specific bool cast allowed as far as I recall and that is casting to bits of width 1.

midend/interpreter.cpp
```diff
@@ -735,6 +747,18 @@ void ExpressionEvaluator::postorder(const IR::Operation_Unary* expression) {
         auto li = l->to<SymbolicBool>();
         clone->expr = new IR::BoolLiteral(li->value);
         DoConstantFolding cf(refMap, typeMap);
+        if (auto cast = expression->to<IR::Cast>()) {
+            if (cast->destType->is<IR::Type_Bits>()) {
+                const IR::Constant* constant;
```
fruffy, line 752: You can write this as a mux.

midend/interpreter.cpp
```diff
@@ -735,6 +747,18 @@ void ExpressionEvaluator::postorder(const IR::Operation_Unary* expression) {
...
+                if (clone->expr) {
+                    constant = new IR::Constant(cast->destType, 1);
+                } else {
+                    constant = new IR::Constant(cast->destType, 0);
+                }
+                set(expression, new SymbolicInteger(constant->to<IR::Constant>()));
```
fruffy, line 758: Why is this cast necessary?

fruffy (changes requested):

midend/interpreter.cpp
```diff
@@ -727,6 +727,18 @@ void ExpressionEvaluator::postorder(const IR::Operation_Unary* expression) {
         auto li = l->to<SymbolicInteger>();
         clone->expr = li->constant;
         DoConstantFolding cf(refMap, typeMap);
+        if (auto cast = expression->to<IR::Cast>()) {
+            if (cast->destType->is<IR::Type_Bits>()) {
+                const IR::Constant* constant;
+                if (clone->expr) {
```
mbudiu-vmw, line 733: I don't think this does what you expect it does.
It compares the expr pointer to nullptr. It does not compare the value of the expression to true or false.

mbudiu-vmw (changes requested):

mbudiu-vmw:
> 
> 
> So part of the problem is that the parserUnroll pass does not run across the entire p4_16_samples folder, right? Then maybe we should enable it for that, too. Analogous to #2914. @mbudiu-vmw What do you think?

We cannot run the unroll over all programs because it only succeeds when the program has an acyclic parser. It is perfectly legal to have P4 programs with parser loops, but these cannot be unrolled. Some back-ends, like ebpf, require acyclic parsers. For these this pass is useful.

fruffy:
Should we apply it to the eBPF back end then? Or is that already the case? The main problem is that this pass is undertested. We are currently using it but it causes various problems. I also remember you mentioning that it may generate millions of additional states. 
What if we throw an error when we detect a cycle that does not advance the cursor? I am assuming the simple switch back end also does not support such cycles.

mbudiu-vmw:
We already run parserUnroll in the ebpf backend, but none of the tests we have are affected since we didn't write any program with loops in a parser for ebpf. The unrolling pass has managed to avoid the state explosion of the previous implementation; one of the tests is running it on switch.p4.

midend/interpreter.cpp
vladyslav-dubina: Oh, i will solve it

midend/interpreter.cpp
vladyslav-dubina: We also need to check the size of this Bit? To make it fit the size 1

midend/interpreter.cpp
vladyslav-dubina: Set type `bit<1>` for new constant. Wouldn't it be better to use `new IR :: Type_Bits` instead of `cast->destType`?

midend/interpreter.cpp
```diff
@@ -738,6 +755,23 @@ void ExpressionEvaluator::postorder(const IR::Operation_Unary* expression) {
         auto li = l->to<SymbolicBool>();
         clone->expr = new IR::BoolLiteral(li->value);
         DoConstantFolding cf(refMap, typeMap);
+        if (auto cast = expression->to<IR::Cast>()) {
+            auto bitsType = cast->destType->to<IR::Type_Bits>();
+            if (bitsType->size == 1) {
+                const IR::Constant* constant;
+                auto boolLoteral = clone->expr->to<IR::BoolLiteral>();
+                if (boolLoteral) {
+                    if (boolLoteral->value == true)
```
fruffy, line 764: This could should look exactly like the other.

vladyslav-dubina:
@fruffy 
Сan we do test in gtest for this case? It seems to me that it will not work, because this is called because of the midend in our tool?

fruffy:
You can also just craft a test for sections where LoopsUnrolling already runs. For example, the eBPF tests. 
gtest should be fine since you are isolating the case.

frontends/common/constantFolding.cpp
```diff
@@ -50,6 +50,12 @@ class CloneConstants : public Transform {
 
 const IR::Expression* DoConstantFolding::getConstant(const IR::Expression* expr) const {
     CHECK_NULL(expr);
+    if (auto pathexpr = expr->to<IR::PathExpression>()) {
```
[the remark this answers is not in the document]
vladyslav-dubina: The reason for this is that the processing for the cast has already been implemented, but it did not work because the expression was PathExpression and its search returned Declaration_Variable instead of Declaration_Constatn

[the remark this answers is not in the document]
fruffy: I do not know the constantFolding pass very well but this change does not appear correct to me. For example, you do not even check if you are returning a constant. And it breaks a lot of existing tests. Shouldn't you handle this in the unrolling pass? Or do you think there is a bug in constantFolding?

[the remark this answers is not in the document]
vladyslav-dubina: It seems to me that there was not enough implementation for the cases with PathExpression, but they are caused by our change of bool to cast. Probably back to the original version

vladyslav-dubina:
This could be caused by a difference in types. For example, when instead of bit <1> (bool) return not bit, but bool, then parsers.second-> states.size () = 4

fruffy:
It looks like you are doing copy propagation here, I do not think constant folding should do this. How does it relate to the issue with casting? Can you show an example where the cast becomes a problem?  issue361.p4 is an empty parser, so it seems unrelated.

vladyslav-dubina:
I don't need to do CloneConstants :: clone (dv-> initializer)?

fruffy:
But then the expression just remains as is, right? Meaning `interpreter.cpp` might still have to handle a case like that. Or does this cause a segmentation fault?

vladyslav-dubina:
This can cause a segmentation fault. Perhaps there is a solution that will bypass the segmentation fault, but I have not found it.

vladyslav-dubina:
There were cases when in constantFolding.cpp, when trying to find a type in typeMap or when trying to get a declaration from a refmap, a segmentation fault was obtained

fruffy:
So I checked this and I saw no issues in constantFolding. There are two ways to solve this. Either 
1) You fix the order of mid end passes that cause parserUnroll to have issues with the input
2) You make parserUnroll robust and able to handle casts in a select expression.

midend/interpreter.cpp
```diff
@@ -730,6 +730,18 @@ void ExpressionEvaluator::postorder(const IR::Operation_Unary* expression) {
         clone->expr = li->constant;
         DoConstantFolding cf(refMap, typeMap);
         cf.setCalledBy(this);
+        if (auto cast = expression->to<IR::Cast>()) {
+            auto bitsType = cast->destType->to<IR::Type_Bits>();
+            const IR::Constant* constant;
+            auto resConst = clone->expr->to<IR::Constant>();
+            if (resConst) {
+                constant = new IR::Constant(bitsType, resConst->value);
```
mbudiu-vmw, line 738: why are you doing this?
The call to DoConstantFolding above should be doing exactly the same thing below, when it calls apply(cf).

midend/interpreter.cpp
```diff
@@ -731,6 +731,20 @@ void ExpressionEvaluator::postorder(const IR::Operation_Unary* expression) {
         DoConstantFolding cf(refMap, typeMap);
         cf.setCalledBy(this);
         auto result = expression->apply(cf);
+        if (!result->is<IR::Constant>()) {
```
mbudiu-vmw, line 734: This is not how this should be done.
If you want to handle casts, use the postorder function for casts.
Also, you still have not explained why this case can ever happen. 
After you call constantFolding one of two things should be true
- result is a constant, and then this code is never executed
- the cast argument is not a constant, and then you hit the BUG case below
So I think that all this code does actually nothing.

vladyslav-dubina:
The reason for such errors is the generation of cast in which expression is of type PathExpression. Earlier I reported that constantFolding is unable to process a cast with an expression of type PathExpression, and it simply returns the same cast, which causes errors like "expected an integer" and "expected a boolean"

fruffy:
>The reason for such errors is the generation of cast in which expression is of type PathExpression. Earlier I reported that constantFolding is unable to process a cast with an expression of type PathExpression, and it simply returns the same cast, which causes errors like "expected an integer" and "expected a boolean"
Can you show an example to reproduce this? Is this the test you have added?

midend/interpreter.cpp
```diff
@@ -913,6 +932,220 @@ void ExpressionEvaluator::postorder(const IR::BoolLiteral* expression) {
...
+        }
+    }
+    BUG("%1%: unexpected type", l);
+}
+
+void ExpressionEvaluator::postorder(const IR::Neq* expression) {
+    auto l = get(expression->left);
+    auto clone = expression->clone();
+    DoConstantFolding cf(refMap, typeMap);
+    cf.setCalledBy(this);
```
mbudiu-vmw, line 988: This makes no sense. You do not call the constant folding pass. Instead of do constant folding manually.
There is no reason to have such a code duplication. 
This is true for all the cases below.

vladyslav-dubina:
On the example with parser-conditional.p4 launched with the argument `"--loopsUnroll"`, you can see that at least https://github.com/vladyslav-dubina/p4c/blob/6587ce96ab7b5a5bfc251efd9d2726f9cba8b608/midend/interpreter.cpp#L1120  makes sense. It's not the same with gauntlet_bounded_loop.p4.

mbudiu-vmw:
I already requested some changes on this PR. You *should* use the constant folding pass, and not do the constant folding manually.

fruffy:
@vladyslav-dubina What is the state of this pull request?

vladyslav-dubina:
@fruffy  I kind of finished what Mihai was talking about

vladyslav-dubina:
@mbudiu-vmw  this PR is ready for review. I seem to have done everything you asked.

mbudiu-vmw (changes requested):

vladyslav-dubina:
At least the test parser-conditional.p4 belongs here  https://github.com/vladyslav-dubina/p4c/blob/cda421d9bd3c180ee2ef94f508e52136c482c4bd/midend/interpreter.cpp#L989
I couldn't get similar behavior with the other three functions, so I'll remove them

vladyslav-dubina:
I added three additional tests, they show the cases in which we enter the other three functions

midend/interpreter.cpp
```diff
@@ -913,6 +932,114 @@ void ExpressionEvaluator::postorder(const IR::BoolLiteral* expression) {
     set(expression, new SymbolicBool(expression));
 }
 
+void ExpressionEvaluator::postorder(const IR::Lss* expression) {
+    auto l = get(expression->left);
+    auto clone = expression->clone();
+    if (l->is<SymbolicInteger>()) {
```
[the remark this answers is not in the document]
mbudiu-vmw: I apologize, this statement was incorrect, since this check is about the type of the left operand, and not of the result.
However, this case should be already handled by Operation_Relation. There is a bug in that function, though, where `expression->apply(cf)` should be replaced by `clone->apply(cf)` (in two places). That is the correct fix. Your code here leaves aside many important cases which are handled by that function and it also duplicates code needlessly.

vladyslav-dubina:
Thanks for the advice @mbudiu-vmw . I don't know how it happened that I didn't notice `expression->apply(cf)` in the first place. I replaced `expression->apply(cf)` with `clone->apply(cf)` as you said, it solved the problem and I removed unnecessary functions

midend/interpreter.cpp
```diff
@@ -878,6 +878,25 @@ void ExpressionEvaluator::postorder(const IR::Operation_Unary* expression) {
         DoConstantFolding cf(refMap, typeMap);
         cf.setCalledBy(this);
         auto result = expression->apply(cf);
+        if (!result->is<IR::BoolLiteral>()) {
```
mbudiu-vmw, line 881: are you positive that this case can happen?
I would think that this would be impossible - the result of constant folding a boolean expression has to a boolean. You are folding here expression, not l.
Which one of the tests exercises this case?

mbudiu-vmw: You have the same bug here: the previous statement should be `clone->apply(cf)`.
Then all this code is no longer needed.

vladyslav-dubina:
@mbudiu-vmw 
If we do `clone->apply(cf)` then we will get the error `"Could not find type for <Cast>(159758)"`

vladyslav-dubina:
The error comes from here https://github.com/vladyslav-dubina/p4c/blob/9da2e5f096fd6bc5216f0788022ed2f4c3515120/frontends/common/constantFolding.cpp#L795 , it might be worth doing `etype = typeMap->getType(getOriginal(), false);` and make an additional check that in case of getting `nullptr `we will go to `etype = e->destType;`

vladyslav-dubina:
Even if we do what I wrote about in the previous comments, we will still get constant , not BoolLiteral, and the need for this https://github.com/vladyslav-dubina/p4c/blob/9da2e5f096fd6bc5216f0788022ed2f4c3515120/midend/interpreter.cpp#L881

mbudiu-vmw:
What is the test that triggers failures?
If there is a bug, it is somewhere else.

mbudiu-vmw:
Before you answer my questions please rebase on main to make sure you picked some fixes that have recently been merged.

vladyslav-dubina:
`constant_folding` returns `Constant `, which is true, but that is returned in the function for `BoolLiteral`. This is due to the `midend `pass in `bf-p4c-tools`, as the pass replaces all bool in selects with cast bool to bit. This is the reason why we are here https://github.com/vladyslav-dubina/p4c/blob/053c4c6d2973799d1ff465a742027e73e9599472/midend/interpreter.cpp#L880 

The test for this example is `issue361-bmv2.p4`

vladyslav-dubina:
I redid the check and added a cast_test test for parserUnroll Gtest. If you remove this https://github.com/vladyslav-dubina/p4c/blob/337309b79c573b0b48d6ddb0491559bf56ad7c78/midend/interpreter.cpp#L881 code, the test will fail.

vladyslav-dubina:
@mbudiu-vmw  Changes have been made and the PR is ready for review

midend/interpreter.cpp
```diff
@@ -1199,15 +1216,17 @@ void ExpressionEvaluator::postorder(const IR::MethodCallExpression* expression)
...
-            } else if (em->method->name.name == P4CoreLibrary::instance.packetIn.lookahead.name) {
-                // If lookahead returns a header, it is always valid.
-                auto type = typeMap->getTypeType(mi->actualMethodType->returnType, true);
-                auto res = factory->create(type, false);
-                if (auto sh = res->to<SymbolicHeader>()) {
-                    sh->setValid(true);
-                }
-                set(expression, res);
-                return;
+            } else if (em->method->name.name ==
```
mbudiu-vmw, line 1219: why are you changing the indentation to this code?
We use 4 spaces anyway.

midend/interpreter.cpp
```diff
@@ -878,6 +878,23 @@ void ExpressionEvaluator::postorder(const IR::Operation_Unary* expression) {
         DoConstantFolding cf(refMap, typeMap);
         cf.setCalledBy(this);
         auto result = expression->apply(cf);
+        if (auto cast = result->to<IR::Cast>()) {
```
mbudiu-vmw, line 881: as I said, you need to call `clone->apply(cf)` above.

mbudiu-vmw:
This PR is superseded by https://github.com/p4lang/p4c/pull/3046

vladyslav-dubina:
@mbudiu-vmw  I ran the test using main with changes from #3046 , but that doesn't solve the problem. All also need changes from this branch

mbudiu-vmw:
Please point me to a test which is failing without these changes.
Also, please use the recommended indentation settings, and refrain from arbitrarily reindenting existing code, this does not respect our style guidelines, and makes reviewing harder since it introduces nonexistent changes.

midend/interpreter.cpp
```diff
@@ -877,9 +877,22 @@ void ExpressionEvaluator::postorder(const IR::Operation_Unary* expression) {
     } else if (l->is<SymbolicBool>()) {
```
VolodymyrPeschanenkoLitSoft, line 877: These changes look much better.  This example contains cast to bit<1>. So, the `get` [here](https://github.com/p4lang/p4c/blob/1750e2f36aa73b64c0707006f15a24cf75d25c79/midend/interpreter.cpp#L848) should return SymbolicInteger type, but not a SymbolicBool. In this case, the additional checking for a Constant will be not necessary below.

VolodymyrPeschanenkoLitSoft:
I think that these tests should cover all possible cases with bool->bit<1>, bit<1>->bool casts. I agree with these changes.  @mbudiu-vmw what do you think?

mbudiu-vmw:
Please address the review comments that were already made several times.
The indentation is incorrect in several places. I have said this at least 6 times for your PRs. 
Reviewing takes a lot of time. At this point I am better off fixing the bugs myself than reading the same code over and over again.
You also have 88 commits in this PR. This is not how commits are supposed to work: see our recommendations here: https://github.com/p4lang/p4c/blob/main/docs/README.md, in particular, please read the section on how to write a commit message: https://github.com/p4lang/p4c/blob/main/docs/CodingStandardPhilosophy.md#Git-commits-and-pull-requests
I would suggest dropping this PR and submitting all your changes in a single commit.

vladyslav-dubina:
Okay, I'll make a new PR with one commit that will have all the changes

mbudiu-vmw:
> I think that these tests should cover all possible cases with bool->bit<1>, bit<1>->bool casts. I agree with these changes. @mbudiu-vmw what do you think?

I am not sure that the code supplied actually is necessary. Isn't the code in main handling all these cases? I would really like to see an issue filed that shows what the actual problem is, then a PR that fixes the issue. The issue should be reproducible.

PR #2916, opened 2021-10-04
Initial support for generating DPDK context JSON

usha1830:
This PR adds support to generate context json files for DPDK backend. The context json contains information about the tables and actions within a P4 program as well as any compiler generated tables. This information can be used by the control plane software to manipulate these entities.

The generated JSON file should comply with a JSON schema (also added as part of this PR)

Some refactoring is done to reuse the printing routines across the .spec file generation and context json file generation.


stolsma:
Maybe a stupid question 😉 but why aren't you using the P4C  P4Info file creation options for this? A seperate control description file creation possibility looks a little bit like double work?🤔

usha1830:
> why aren't you using the P4C P4Info file creation options for this? A seperate control description file creation possibility looks a little bit like double work

As per my understanding the  P4C P4Info file creation option ("--p4runtime-files", "filelist") is used to generate only the P4Info file and the _filelist_ is only to allow generating P4Info in different file formats such as .json, .txt, .bin
Please correct me if I am wrong.

The context json is different from the P4Info as it contains information about additional compiler generated tables as well (For ex, tables generated for implementing action profile and action selectors)

stolsma:
> As per my understanding the P4C P4Info file creation option ("--p4runtime-files", "filelist") is used to generate only the P4Info file and the _filelist_ is only to allow generating P4Info in different file formats such as .json, .txt, .bin Please correct me if I am wrong.

No, you're right I think...

> The context json is different from the P4Info as it contains information about additional compiler generated tables as well (For ex, tables generated for implementing action profile and action selectors)

So it describes the compiler generated DPDK tables and actions and not the P4 program ones. Yeah, thats different then P4Info... Thanks for the clarification !

backends/dpdk/dpdkArch.cpp
```diff
@@ -329,6 +329,28 @@ bool CollectMetadataHeaderInfo::preorder(const IR::Type_Struct *s) {
...
+    }
+    /* Push all non-selector keys to the key_map for this table.
+       Selector keys become part of the selector table */
+    for (auto key : keys->keyElements) {
+        cstring keyTypeStr = key->expression->toString();
+        if (key->matchType->toString() != "selector")
+            tableKeys.push_back(keyTypeStr);
+    }
+
+    auto control =  findOrigCtxt<IR::P4Control>();
```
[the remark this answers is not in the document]
mbudiu-vmw: Do you want a CHECK_NULL for control and table?

backends/dpdk/dpdkContext.cpp
```diff
@@ -0,0 +1,582 @@
...
+    }
+
+    // Keep the compiler generated (hidden) tables at the end
+    for (auto d : selector_tables)
+        tables.push_back(d);
+    for (auto d : action_data_tables)
+        tables.push_back(d);
+}
+
+bool CollectTablesForContextJson::preorder(const IR::P4Program * /*p*/) {
```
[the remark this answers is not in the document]
mbudiu-vmw: BTW: your indentation is off here and below.

backends/dpdk/dpdkContext.cpp
```diff
@@ -0,0 +1,582 @@
...
+        actionAttrMap.emplace(act->getName(), attr);
+    }
+}
+
+/* Print a single table object of match_table type in Context JSON*/
+void WriteContextJson::printTableCtxtJson(const IR::P4Table *tbl, std::ostream &out) {
+    bool hasActionProfileSelector = false;
+    auto mainTableAttr = ::get(tableAttrmap, tbl->name.originalName);
+    cstring tableName = mainTableAttr.controlName + "." + tbl->name.originalName;
+    add_space(out, 8); out << "{\n";
```
mbudiu-vmw, line 228: this is a big ugly, it would be nicer to have indentation maintained automatically and automatically increased on blocks or arrays. But I won't be upset if you don't change this.
We already have several places where we write json. Maybe we should have a common JsonWriter class. Alternatively, you could generate a JSON object and just serialize it as a string.

backends/dpdk/dpdkContext.cpp
```diff
@@ -0,0 +1,582 @@
...
+        out << "\"name\": \"" <<  tableName <<"\",\n";
+        add_space(out, 20); out << "\"handle\": " << tableAttr.tableHandle   <<"\n";
+        add_space(out, 16); out << "}\n";
+        add_space(out, 12);
+    }
+    out << "],\n";
+
+    // Reference to compiler generated group table in case of action selector
+    add_space(out, 12); out << "\"selection_table_refs\": [";
+    if (structure->group_tables.count(tbl->name)) {
```
mbudiu-vmw, line 270: this code is almost the same as above. could it be factored into a separate function?

backends/dpdk/dpdkContext.cpp
```diff
@@ -0,0 +1,582 @@
...
+   /*     if (match_keys->keyElements.size() != mainTableAttr.tableKeys.size()) {
+            std::cout << match_keys->keyElements.size() << " " <<  mainTableAttr.tableKeys.size() << std::endl;
+            BUG("Inconsistent match keys between input program and transformed code");
+        }*/
+//        for (auto mkf : match_keys->keyElements) {
+//            if (mkf->expression->is<IR::Member>()) {
+          for (auto matchKeyFromPrg : mainTableAttr.tableKeys) {
+                auto mkf = match_keys->keyElements.at(index);
+//                auto matchKeyFromPrg = mainTableAttr.tableKeys.at(index);
+               if (mkf->expression->is<IR::Member>()) {
```
mbudiu-vmw, line 303: incorrect indentation

backends/dpdk/dpdkContext.cpp
```diff
@@ -0,0 +1,582 @@
...
+            if (attr.params) {
+                int index = 0;
+                for (auto param : *(attr.params)) {
+                    if (position) out << ",";
+                    out << "\n";
+                    add_space(out, 24); out << "{\n";
+                    add_space(out, 28); out << "\"name\": \""
+                                            << param->name.originalName << "\",\n";
+                    add_space(out, 28); out << "\"start_bit\": 0,\n";
+                    add_space(out, 28); out << "\"bit_width\": "
```
mbudiu-vmw, line 391: are you checking someplace whether the parameters all have scalar types?

mbudiu-vmw: there is a request to support structs for actions arguments.

backends/dpdk/dpdkContext.cpp
```diff
@@ -0,0 +1,582 @@
...
+            }
+            out << "]\n";
+            add_space(out, 16); out << "}";
+            first = false;
+        }
+    }
+    if (!first) out << "\n";
+    add_space(out, 12); out << "],\n";
+
+    // These fields is similar to the actions array with slightly different information
```
mbudiu-vmw, line 410: are

backends/dpdk/dpdkContext.cpp
```diff
@@ -0,0 +1,582 @@
...
+        out << "\n";
+        add_space(out, 28); out << "{\n";
+        add_space(out, 20); out << "\"action_name\": \"" << actName << "\",\n";
+        add_space(out, 32); out << "\"action_handle\": " << attr.actionHandle << ",\n";
+        add_space(out, 32); out << "\"immediate_fields\": [";
+
+        int position = 0;
+        if (attr.params) {
+            int index = 0;
+            for (auto param : *(attr.params)) {
```
mbudiu-vmw, line 435: inconsistent indentation

backends/dpdk/dpdkContext.h
```diff
@@ -0,0 +1,128 @@
...
+Context JSON is a JSON file used by the control plane software for manipulating tables and
+actions. It contains all relevant information regarding the tables and actions.
+The context JSON is based on the JSON Schema defined in DPDK_context_schema.json.
+*/
+
+namespace DPDK {
+
+/* This structure holds table attributes required for context JSON which are not
+    part of P4Table */
+struct TableAttributes {
```
mbudiu-vmw, line 35: Could you just have a method in this struct that converts it to JSON?
Then you could do the json generation by calling a few methods here and there and then serializing the resulting JSON object. It would be much nicer to read.

backends/dpdk/dpdkProgramStructure.h
```diff
@@ -39,6 +39,9 @@ struct DpdkProgramStructure {
     std::map<const IR::Declaration_Instance *, cstring>         csum_map;
     std::map<cstring, int>                                      error_map;
     std::vector<const IR::Declaration_Instance *>               externDecls;
+    std::map<cstring, std::vector<cstring>>                     key_map;
+    std::map<cstring, const IR::P4Table *> group_tables;
```
mbudiu-vmw, line 43: why not indent these the same as the previous ones?

backends/dpdk/printUtils.cpp
```diff
@@ -0,0 +1,125 @@
...
+cstring toStr(const IR::MethodCallExpression *const m) {
+    if (auto path = m->method->to<IR::PathExpression>()) {
+        return path->path->name.toString();
+    } else {
+        ::error("%1% is not a PathExpression", m->toString());
+    }
+    return "";
+}
+
+cstring toStr(const IR::Expression *const exp) {
```
mbudiu-vmw, line 74: Couldn't this be done nicer using a visitor?

backends/dpdk/CMakeLists.txt
```diff
@@ -104,6 +108,6 @@ set(DPDK_COMPILER_DRIVER "${CMAKE_CURRENT_SOURCE_DIR}/run-dpdk-test.py")
 set (P4_16_SUITES
   "${P4C_SOURCE_DIR}/testdata/p4_16_samples/psa-*.p4"
   "${P4C_SOURCE_DIR}/testdata/p4_16_samples/pna-*.p4")
-p4c_add_tests("dpdk" ${DPDK_COMPILER_DRIVER} "${P4_16_SUITES}" "" "--bfrt")
+p4c_add_tests("dpdk" ${DPDK_COMPILER_DRIVER} "${P4_16_SUITES}" "" "--bfrt '--context'")
```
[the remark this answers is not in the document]
usha1830: These files are going to be used by a control plane software and we need to track changes in these files to ensure they do not break the processing of these files.

backends/dpdk/backend.cpp
```diff
@@ -54,11 +55,12 @@ void DpdkBackend::convert(const IR::ToplevelBlock *tlb) {
         new P4::ConstantFolding(refMap, typeMap, false),
         new P4::TypeChecking(refMap, typeMap),
         new P4::RemoveAllUnusedDeclarations(refMap),
-        new ConvertActionSelectorAndProfile(refMap, typeMap),
+        new ConvertActionSelectorAndProfile(refMap, typeMap, &structure),
         new CollectAddOnMissTable(refMap, typeMap, &structure),
         new P4::MoveDeclarations(),  // Move all local declarations to the beginning
         new CollectProgramStructure(refMap, typeMap, &structure),
         new CollectMetadataHeaderInfo(&structure),
+        new CollectTableInfo(&structure),
```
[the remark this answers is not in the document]
usha1830: Are you referring to ConvertActionSelectorAndProfile and CollectTableInfo passes? 
I can't think of any reason why the structure would have invalid data.
These are two independent passes updating different members of DpdkProgramStructure class.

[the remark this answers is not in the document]
mbudiu-vmw: Because the program can change between these passes

backends/dpdk/printUtils.cpp
usha1830: Sorry, I don't think I follow this comment completely. Could you please elaborate a little?

backends/dpdk/dpdkProgramStructure.h
usha1830: Ack. Will fix it in the next commit.

backends/dpdk/dpdkContext.cpp
usha1830: Ack. Will fix it in the next commit.

backends/dpdk/dpdkContext.cpp
```diff
@@ -0,0 +1,582 @@
...
+        return ACTION_HANDLE_PREFIX | newActionHandle++;
+}
+
+// Helper function for pretty printing into JSON file
+void WriteContextJson::add_space(std::ostream &out, int size) {
+    out << std::setfill(' ') << std::setw(size) << " ";
+}
+
+/* Main function for generating context JSON */
+bool WriteContextJson::preorder(const IR::P4Program * /*p*/) {
```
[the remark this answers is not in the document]
usha1830: Ack. Will fix it in the next commit.

backends/dpdk/dpdkContext.cpp
```diff
@@ -0,0 +1,582 @@
...
+            cstring progName =  options.file;
+            progName = progName.findlast('/');
+            auto fileext = progName.find(".");
+            progName = progName.replace(fileext, "");
+            progName = progName.trim("/\t\n\r");
+
+            /* Print program level information*/
+            out << "{\n";
+            add_space(out, 4); out << "\"program_name\": \"" << progName <<"\",\n";
+            if (!options.testcontextJson) {
```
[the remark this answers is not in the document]
usha1830: Yes. This option is only internally used in the regression test.

backends/dpdk/dpdkContext.cpp
usha1830: Ack. I will try to reorganize the code with some additional functions to maintain the indentation and write to json.

backends/dpdk/dpdkContext.cpp
```diff
@@ -0,0 +1,582 @@
...
+        out << "\"action_profile\": \"" << memberTable->name.originalName <<"\",\n";
+    }
+
+    // Match Key Fields for this table
+    add_space(out, 12); out << "\"match_key_fields\": [";
+    auto match_keys = tbl->getKey();
+
+    if (match_keys) {
+        int index = 0;
+   /*     if (match_keys->keyElements.size() != mainTableAttr.tableKeys.size()) {
```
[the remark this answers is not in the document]
usha1830: Yes. Will delete this.

backends/dpdk/printUtils.cpp
mbudiu-vmw: A visitor class that has a preorder method for each kind of expression could achieve the same result in a nicer way.

control-plane/bfruntime.h
```diff
@@ -58,6 +58,7 @@ static inline constexpr P4Id getIdPrefix(P4Id id) {
 }
 
 static inline Util::JsonObject* findJsonTable(Util::JsonArray* tablesJson, cstring tblName) {
+    tblName = "pipe." + tblName;
```
hanw, line 61: "pipe." is not a prefix, it comes from the name of the pipeline.

control-plane/bfruntime.cpp
```diff
@@ -292,7 +292,8 @@ BFRuntimeGenerator::addKeyField(Util::JsonArray* dataJson, P4Id id, cstring name
 BFRuntimeGenerator::initTableJson(const std::string& name,
         P4Id id, cstring tableType, int64_t size, Util::JsonArray* annotations) {
     auto* tableJson = new Util::JsonObject();
-    tableJson->emplace("name", name);
+    const std::string tableName = "pipe." + name;
```
hanw, line 295: Here as well, this is incorrect. 'pipe' is the name of the pipeline, not a fixed prefix.

backends/dpdk/CMakeLists.txt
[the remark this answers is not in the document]
hanw: @usha1830 please check with the team to see if it is possible to setup a regression internally?

backends/dpdk/options.cpp
```diff
@@ -4,11 +4,19 @@
 
 namespace DPDK {
 
+// Collect the command line arguments into DpdkCompCmd
+cstring DpdkOptions::DpdkCompCmd = "";
```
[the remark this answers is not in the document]
usha1830: https://github.com/p4lang/p4c/blob/9a251715b027af92068ec65d84ac1268b167b1be/backends/dpdk/dpdkContext.cpp#L108
https://github.com/p4lang/p4c/blob/9a251715b027af92068ec65d84ac1268b167b1be/backends/dpdk/dpdkContext.cpp#L122

backends/dpdk/options.h
```diff
@@ -45,6 +54,13 @@ class DpdkOptions : public CompilerOptions {
         registerOption("-o", "outfile",
                 [this](const char* arg) { outputFile = arg; return true; },
                 "Write output to outfile");
+        registerOption("--context", "file",
+                [this](const char *arg) { ctxtFile = arg; return true; },
+                "Generate and write context JSON to the specified file");
+        registerOption("--testcontextJson", nullptr,
```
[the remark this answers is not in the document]
usha1830: If we remove the context json output comparison from regressions, this option will no longer be needed.

BTW, I need to avoid printing some fields based on the flag for this options. Would it be possible to achieve that without introducing an option?
https://github.com/p4lang/p4c/blob/4b5f7b43b17d6a6db7a994e2f41a4f6177a6231e/backends/dpdk/dpdkContext.cpp#L120

backends/dpdk/CMakeLists.txt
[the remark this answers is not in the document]
usha1830: Sure. Let me check with the team.

backends/dpdk/printUtils.cpp
usha1830: Sorry about my naïve question.
I think I still don't understand this correctly. These are helper routines for printing. If we make it a visitor how would we return a string there and how the call would be replaced with a visitor function?
For Example here:
https://github.com/p4lang/p4c/blob/4b5f7b43b17d6a6db7a994e2f41a4f6177a6231e/backends/dpdk/spec.cpp#L55

mbudiu-vmw: The visitor has a stringbuilder that it appends strings to. An Inspector is enough. Look at the ebpf backend for an example.

usha1830: Can we take this refactoring as a seperate PR as it would impact .spec file generation as well? 

This is existing code for .spec file generation. I just moved it from _backend/dpdk/spec.cpp_ to this new file _backend/dpdk/printUtils.cpp_ for reusing some of the print routines in context JSON file generation.

Perhaps, I can file an issue and work on this refactoring after this context JSON related PR?

mbudiu-vmw: This is called "Technical debt". If you indeed plan to make these changes that's fine, but we've had contributions in the past which were not followed by the promised cleanups. 
We can merge this part of the code as it is, it's not terrible.
But if I am right that a visitor would solve this nicely, it should be actually quite a quick fix. You would essentially define a visitor class, and convert each of these cases of the 'if' into a preorder method. That's it.

backends/dpdk/dpdkContext.h
usha1830: I have updated the code use lib/json.h to create JSON object and serializing it

backends/dpdk/CMakeLists.txt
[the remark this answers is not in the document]
usha1830: Removed the context json outputs.

backends/dpdk/dpdkContext.cpp
usha1830: Done

backends/dpdk/dpdkContext.cpp
```diff
@@ -0,0 +1,582 @@
...
+    add_space(out, 12); out << "\"direction\": \"" << mainTableAttr.direction << "\",\n";
+    add_space(out, 12); out << "\"handle\": " << mainTableAttr.tableHandle << ",\n";
+    add_space(out, 12); out << "\"name\": \"" << tableName << "\",\n";
+    add_space(out, 12); out << "\"size\": ";
+
+    if (auto size = tbl->properties->getProperty("size")) {
+        out << size->value << ",\n";
+    } else {
+        // default table size for DPDK
+        out << "65536" << ",\n";
```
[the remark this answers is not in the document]
usha1830: Done

backends/dpdk/dpdkContext.cpp
usha1830: Added checks

backends/dpdk/constants.h
```diff
@@ -0,0 +1,27 @@
...
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+#ifndef BACKENDS_DPDK_CONSTANTS_H_
+#define BACKENDS_DPDK_CONSTANTS_H_
+
+/* Unique handle for action and table */
+#define TABLE_HANDLE_PREFIX 0x00010000
```
mbudiu-vmw, line 21: I think that the preferred C++ way is to use a const variable, but this will work too.

backends/dpdk/dpdkContext.cpp
```diff
@@ -0,0 +1,429 @@
...
+        if (actName != "NoAction")
+            actName = ctrlName + "." + actName;
+        oneAction->emplace("action_name", actName);
+        oneAction->emplace("action_handle", attr.actionHandle);
+        auto* immFldArray = new Util::JsonArray();
+        if (attr.params) {
+            int index = 0;
+            int position = 0;
+            for (auto param : *(attr.params)) {
+                // TODO Handle other types of parameters
```
mbudiu-vmw, line 233: do you want to BUG when other parameter types appear?

backends/dpdk/dpdkContext.cpp
```diff
@@ -0,0 +1,429 @@
...
+               act->emplace("is_compiler_added_action", attr.is_compiler_added_action);
+               act->emplace("allowed_as_hit_action", attr.allowed_as_hit_action);
+               act->emplace("allowed_as_default_action", attr.allowed_as_default_action);;
+           }
+           auto* paramJson = new Util::JsonArray();
+           if (attr.params) {
+                int index = 0;
+                int position = 0;
+                for (auto param : *(attr.params)) {
+                // TODO Handle other types of parameters
```
mbudiu-vmw, line 290: same here

backends/dpdk/printUtils.cpp
```diff
@@ -0,0 +1,130 @@
...
+    if (auto path = m->method->to<IR::PathExpression>()) {
+        return path->path->name.toString();
+    } else {
+        ::error("%1% is not a PathExpression", m->toString());
+    }
+    return "";
+}
+
+bool ConvertExprToString::preorder(const IR::Constant *e) {
+    str = toStr(e);
```
mbudiu-vmw, line 59: why not just inline here the specific overloaded function that you are calling (the one for Constant).
This is the only call of that function, the function can then be deleted.
This is the main point of having the visitor: you replace a toStr(IR::Constant* e) function with a
preorder(IR::Constant* e) function.

backends/dpdk/printUtils.cpp
```diff
@@ -0,0 +1,130 @@
...
+    return false;
+}
+
+cstring toStr(const IR::Expression *const exp) {
+    auto exprToString = new ConvertExprToString;
+    exp->apply(*exprToString);
+    return exprToString->str;
+}
+
+cstring toStr(const IR::Type *const type) {
```
mbudiu-vmw, line 105: this function can also become several visitor preorder functions

mbudiu-vmw (changes requested):

backends/dpdk/constants.h
usha1830: Fixed

backends/dpdk/printUtils.cpp
usha1830: Fixed

backends/dpdk/printUtils.cpp
usha1830: Fixed

backends/dpdk/printUtils.cpp
usha1830: Fixed.

control-plane/bfruntime.h
usha1830: Will work on this in a follow up PR to update the bfRT JSON output with fully qualified table names. Created an issue for the same.
https://github.com/p4lang/p4c/issues/2927

backends/dpdk/printUtils.cpp
```diff
@@ -0,0 +1,121 @@
...
+    type->apply(*typeToString);
+    return typeToString->str;
+}
+
+bool ConvertPropertyValToString::preorder(const IR::ExpressionValue *property) {
+    str = toStr(property->expression);
+    return false;
+}
+
+cstring toStr(const IR::PropertyValue *const property) {
```
mbudiu-vmw, line 116: this implementation works, but I would have made the stringstream a field of the visitor and each method would have appended to it. the result would be calling str on the stringstream. that is the point of using a stringbuilder - you reuse it everywhere, you don't allocate a fresh one for each string you append.

backends/dpdk/printUtils.h
```diff
@@ -0,0 +1,70 @@
...
+    bool preorder(const IR::BoolLiteral *b);
+    bool preorder(const IR::Member *m);
+    bool preorder(const IR::PathExpression *p);
+    bool preorder(const IR::TypeNameExpression *p);
+    bool preorder(const IR::MethodCallExpression *m);
+    bool preorder(const IR::Cast *e);
+    bool preorder(const IR::ArrayIndex *e);
+};
+
+class ConvertTypeToString : public Inspector {
```
mbudiu-vmw, line 53: Do you even need 3 separate visitors?
What happens if you merge them into a single one?

backends/dpdk/printUtils.cpp
```diff
@@ -0,0 +1,121 @@
...
+
+namespace DPDK {
+
+bool ConvertToString::preorder(const IR::Expression *e) {
+    BUG("%1% not implemented", e);
+    return false;
+}
+
+bool ConvertToString::preorder(const IR::Type *t) {
+    std::cerr << t->node_type_name() << std::endl;
```
mbudiu-vmw, line 26: You know that BUG can display a message?
You can add a format string `BUG("not implemented %1%", p->node_type_name());`
Don't use std::cerr.

mbudiu-vmw (approved):

backends/dpdk/printUtils.cpp
usha1830: Updated as per suggestion.

usha1830:
@mbudiu-vmw  Can this PR be merged? Thanks!

PR #2917, opened 2021-10-04
Support minSizeInBits and minSizeInBytes for scalars and type names

mbudiu-vmw:
A proposal solution for https://github.com/p4lang/p4-spec/issues/937


fruffy:
The semantics of this are similar to the internal IR function `width_bits`()? E.g., Varbits return 0.

mbudiu-vmw:
This is a generalization of the minSizeInBits suffix method which already exists in the spec, except that we now allow it to be applied to more expressions. In fact, the implementation already supported all these cases, we were just disallowing them. This will be discussed in the LDWG.

fruffy (approved):
Just one request.
Can the older test https://github.com/p4lang/p4c/blob/main/testdata/p4_16_samples/issue1717.p4
be changed to have a main function and use the output that was calculated?

mbudiu-vmw:
We won't merge this until the LDWG has approved this syntax.
If all tests I will modify the old test.

PR #2919, opened 2021-10-05
Added type to PathExpression for simplified key variables.

vladyslav-dubina:
Solve problems with Z3 solver, where simplified key variables don`t have a type.

### DIFF
--- a/midend/simplifyKey.cpp
+++ b/midend/simplifyKey.cpp
@@ -58,12 +58,12 @@ const IR::Node* DoSimplifyKey::postorder(IR::KeyElement* element) {
     auto type = typeMap->getType(element->expression, true);
     auto decl = new IR::Declaration_Variable(tmp, type, nullptr);
     insertions->declarations.push_back(decl);
-    auto left = new IR::PathExpression(tmp);
+    auto left = new IR::PathExpression(type, new IR::Path(tmp));
     auto right = element->expression;
     auto assign = new IR::AssignmentStatement(element->expression->srcInfo, left, right);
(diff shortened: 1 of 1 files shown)


midend/simplifyKey.cpp
```diff
@@ -58,12 +58,12 @@ const IR::Node* DoSimplifyKey::postorder(IR::KeyElement* element) {
     auto type = typeMap->getType(element->expression, true);
     auto decl = new IR::Declaration_Variable(tmp, type, nullptr);
     insertions->declarations.push_back(decl);
-    auto left = new IR::PathExpression(tmp);
+    auto left = new IR::PathExpression(type, new IR::Path(tmp));
```
mbudiu-vmw, line 61: I am not sure how this helps.
In general, the type field of an expression is not maintained by the compiler. 
There is one single pass which is used to populate it, called ApplyTypesToExpressions, which is uses sometimes to avoid looking up types in the typeMap. But any subsequent changes to the IR could make the type in the Expression incorrect.
This change is correct, but won't guarantee that the type field is correct after any subsequent pass that modifies the code.

mbudiu-vmw:
I don't know how you are using the Z3 solver, but perhaps the formulas you generate for it should actually use the types from the typeMap.

mbudiu-vmw (approved):
The tests are failing because there is a wrong reference output committed in the repository.
I have committed a fix for that, hopefully I will be able to merge it and you can rebase on main to get that fixed.

vladyslav-dubina:
Ok, thanks. It`s really important for us

PR #2921, opened 2021-10-07
Enable loopsunrolling by default.

fruffy:
This is a quick edit to check what happens if you enable parserunrolling for the entire p4test back end. Let's see what breaks.
@VolodymyrPeschanenkoIntel  @vladyslav-dubina


fruffy:
Yes, this was mostly because of how it impacts the reference file output.

fruffy:
Supersedd by #3142.

PR #2923, opened 2021-10-11
Added nullptr checks to changes introduced by commit 51611f0 (#2881)

MichalKekely:


MichalKekely:
> Can these be triggered by real programs? If so, can you supply some test cases?

Yes, these can be triggered by real programs. I will try to provide some examples.

> If not, perhaps these should be CHECK_NULL. Can you argue why ignoring the null pointer is the right action?

`getStorageLocation` can return `nullptr` if the storage location does not exist which is fine. When we are trying to get a storage location for `IR::Member` all we are doing is simply getting the storage location of its expression. Therefore if the storage location of this expression does not exist it does not exist for the member as well so we should return `nullptr`.

`initHeaderParam` is then in some cases called with whatever `getStorageLocation` returns, which obviously might be `nullptr` if the storage location does not exist. This function then updates only some specific cases of the storage location (more specifically structures); if the storage location does not exist nothing should be updated (in the same way as if it existed but was of a different subtype).

mbudiu-vmw (approved):
I will approve this, but it would be nice to have a test case if you can come up with one.
If not, we can merge as it is.  According to @fruffy issue561 is an example (but I am not sure whether it covers all these changes or only some).

PR #2926, opened 2021-10-12
Throw an error when reference files are missing.

### LINKED ISSUE #2892: Fail CI if a test does not contain the expected reference outputs.
I.e., the outputs are not submitted as part of the PR.
Today we only fail if the outputs are different.

fruffy:
Fixes #2892.

With this pull request, tests will fail if they are missing their reference files.


fruffy:
I added this message to help users in the case of an error:
`"Missing reference for file %s. Please rerun the test with the -f option turned on or rerun all tests using \"P4TEST_REPLACE=True make check\"."`

PR #2930, opened 2021-10-15
dpdk: support of PNA extern send_to_port

mvido:
send_to_port() extern for PNA architecture has been added recently, but instructions were not generated properly.
This commit fixes the generated instructions.
Metadata field used as operand for 'tx' instruction is added to the main metadata structure.
The name of the field set when send_to_port extern is used is unified with the name of the field used as operand for 'tx' instruction.

### DIFF
--- a/backends/dpdk/dpdkArch.cpp
+++ b/backends/dpdk/dpdkArch.cpp
@@ -24,6 +24,7 @@ limitations under the License.
  */
 
 #include "dpdkArch.h"
+#include "dpdkHelpers.h"
 #include "frontends/p4/coreLibrary.h"
 #include "frontends/common/resolveReferences/referenceMap.h"
 #include "frontends/p4/externInstance.h"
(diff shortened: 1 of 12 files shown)


backends/dpdk/dpdkArch.cpp
```diff
@@ -342,6 +343,15 @@ const IR::Node *InjectJumboStruct::preorder(IR::Type_Struct *s) {
     return s;
 }
 
+const IR::Node *InjectOutputPortMetadataField::preorder(IR::Type_Struct *s) {
```
mbudiu-vmw, line 346: why isn't this field part of pna.p4?

mvido: There is no metadata field for output port in PNA spec. I don't know precisely why, the question should be rather directed to the authors of the spec. But my guess is that instead of metadata field which is in other architectures, here we have just send_to_port() extern function instead. DPDK backend needs the metadata field to store the port number internally so that it could be used for tx instruction. That's why there has to be this internally created metadata field.

We could consider different name for this metadata field though, so that it was more clear that it is not standard field defined by architecture, but field needed internally by target backend. I just kept the same name which was already used there before.

PR #2931, opened 2021-10-18
DPDK Backend: Fix context Json generation when table size is explicitly specified

usha1830:
Minor fix to use the getSizeProperty() method instead of getProperty for retrieving the table size.


usha1830:
@hanw Thanks for the review. Can this PR be merged?

Issue #2279, opened 2020-03-31
Missing StrengthReduction for complex expressions in actions


fruffy:
It looks like `StrengthReduction` is not triggered here:

```p4
control ingress(inout Headers hdr, inout Metadata meta,
                 inout standard_metadata_t stdmeta) {
    simple_struct test = { 128w0 };
    bit<16> tmp_val = 16w3;
    action do_action() {
        hdr.eth_hdr.eth_type = 16w3 + (tmp_val > 2 ? 16w3 : 16w1);
    }
    apply {
        do_action();
    }
}
```
after `MidEndLast` the program looks like this:

```p4 
control ingress(inout Headers hdr, inout Metadata meta, inout standard_metadata_t stdmeta) {
    bit<16> tmp_val_0;
    @name("ingress.do_action") action do_action() {
        hdr.eth_hdr.eth_type = 16w3 + (tmp_val_0 > 16w2 ? 16w3 : 16w1);
    }
    apply {
        tmp_val_0 = 16w3;
        do_action();
    }
}
```
Instead of, for example:
```p4 
control ingress(inout Headers hdr, inout Metadata meta, inout standard_metadata_t stdmeta) {
    bit<16> tmp_val_0;
    @name("ingress.do_action") action do_action() {
        hdr.eth_hdr.eth_type = 16w6;
    }
    apply {
        do_action();
    }
}
```

Example program:

[missed_strength_reduction.p4.txt](https://github.com/p4lang/p4c/files/4409547/missed_strength_reduction.p4.txt)

mbudiu-vmw:
There is no interprocedural constant-propagation in the front-end. I think this is not a real issue; after the action inlining pass (done in the mid-end) this optimization should happen. I will close the issue, but if this hasn't happened by the end of the midend please reopen.

fruffy:
I updated the post, this is also not handled in the mid-end.

fruffy:
@mbudiu-vmw I cannot reopen this issue.

fruffy:
This is fixed with #2932.

PR #2932, opened 2021-10-18
Fix: Issue #2279 - Implementing the GlobalCopyPropagation pass

filipsyrmia:
This PR introduces a new pass with the idea of implementing copy propagation logic, but on a 'global' level (regarding control blocks), with the goal of optimizing situations similar to the one described in the #2279 issue. The optimization described isn't a responsibility of the StrengthReduction pass' logic (as mentioned in the issue), it seems like it should be covered in the copy propagation logic of this compiler.

The logic was implemented as a standalone pass rather than expanding upon the existing LocalCopyPropagation pass, this was done to avoid needlessly complicating the already complex logic of the LocalCopyPropagation pass. Having said that, the new pass was designed with the idea of being run in tandem with the LocalCopyPropagation pass and doesn't introduce some of the optimizations that are already included in the existing pass. (This may warrant putting these 2 passes in a new pass manager and running them as such, for now I have just placed them as one after another in the 'p4test/midend.cpp', where the GlobalCopyPropagation is run first because I believe this order is more logical when you look at the optimizations it introduces)

The optimization mentioned in the issue #2279 is accomplished by visiting the body of the action after the 'ActionCall' node, that corresponds to the action mentioned, is visited. This way we retroactively edit the body of the action and propagate any values that are known at that time.

The only thing that seems problematic to me, and was during the process of developing this pass, is that doesn't seem to be a way of retroactively editing the body of the action without using the 'const_cast'  feature. This is needed because after resolving the 'ActionCall' in the preorder function of IR node 'MethodCallExpression' the action pointer that is available to use is a 'const IR::P4Action*' and doing any editing to the body of the action pointed to by this pointer is impossible, therefore the solution I found is forcefully getting rid of the problematic 'const' part. 
Any suggestions regarding this problem are welcome.

### DIFF
--- a/midend/global_copyprop.cpp
+++ b/midend/global_copyprop.cpp
@@ -1,0 +1,313 @@
+#include "global_copyprop.h"
+
+namespace P4 {
+
+/// Convert an expression into a string that uniqely identifies the lvalue referenced.
+/// Return null cstring if not a reference to a lvalue.
+static cstring lvalue_name(const IR::Expression *exp) {
+    if (auto p = exp->to<IR::PathExpression>())
+        return p->path->name;
+    if (auto m = exp->to<IR::Member>()) {
+        if (auto base = lvalue_name(m->expr))
+            return base + "." + m->member;
+    } else if (auto a = exp->to<IR::ArrayIndex>()) {
+        if (auto k = a->right->to<IR::Constant>()) {
+            if (auto base = lvalue_name(a->left))
+                return base + "[" + std::to_string(k->asInt()) + "]";
+        }
+    } else if (auto sl = exp->to<IR::Slice>()) {
+        if (auto e0 = lvalue_name(sl->e0))
+            return e0;
+    }
+    return cstring();
+}
+
+/// Test to see if names denote overlapping locations.
+bool names_overlap(cstring name1, cstring name2) {
+    if (name1 == name2) return true;
+    if (name1.startsWith(name2) && strchr(".[", name1.get(name2.size())))
+        return true;
+    if (name2.startsWith(name1) && strchr(".[", name2.get(name1.size())))
+        return true;
+    return false;
+}
+
+// Removes outdated values for variables.
+void removeVarsContaining(std::map<cstring, const IR::Expression*> *vars, cstring name) {
+    LOG6("removeVarsContaining(" << name << ")");
+    for (auto &var : *vars) {
+        LOG7("  checking entry: " << var.first << " = " << var.second);
+        if (names_overlap(var.first, name)) {
+            LOG4("  dropping the value for '" << var.first << "' as '"
+                     << name << "' is being assigned to");
+            var.second = nullptr;
+        }
+    }
+}
+
+// Removes values if they have changed
+void compareValuesInMaps(std::map<cstring, const IR::Expression*> *oldValues,
+                            std::map<cstring, const IR::Expression*> *newValues) {
+    for (auto it : *newValues) {
+        auto oldValue = (*oldValues)[it.first];
+        if (((it.second == nullptr) ^ (oldValue == nullptr)) ||
+                it.second && oldValue && !(it.second->equiv(*oldValue)))
+            removeVarsContaining(oldValues, it.first);
+    }
+}
+
+// Removes values if they are used as Out/InOut parameter
+void checkParametersForMap(const IR::ParameterList *params,
+                            std::map<cstring, const IR::Expression*> *vars) {
+    for (auto param : params->parameters)
+        if (param->hasOut())
+            removeVarsContaining(vars, param->name.name);
+}
+
+bool FindVariableValues::preorder(const IR::P4Control *ctrl) {
+    LOG2("FindVariableValues working on control: " << ctrl->name);
+    working = true;
+
+    return true;
+}
+
+void FindVariableValues::postorder(const IR::P4Control *ctrl) {
+    LOG2("FindVariableValues finished working on control: " << ctrl->name);
+    // Clear map as the pass works on every control block separately
+    vars.clear();
+    working = false;
+}
+
+bool FindVariableValues::preorder(const IR::P4Action *act) {
+    return false;
+}
+
+bool FindVariableValues::preorder(const IR::P4Table *tab) {
+    return false;
+}
+
+// When visiting IfStatement nodes the literal values assigned to variables in those
+// 'ifTrue' and 'ifFalse' blocks can't be retained in the 'vars' container since the
+// path taken can't be determined in compile time. Therefore a copy is made to preserve
+// the state of the map before those blocks, and all variables that are possibly changed
+// in these blocks are removed from the original map.
+bool FindVariableValues::preorder(const IR::IfStatement *stat) {
+    std::map<cstring, const IR::Expression*> copyOfVars(vars);
+    LOG3("Working on 'IfStatement->ifTrue' block: " << stat->ifTrue);
+    visit(stat->ifTrue);
+    // Check if some variables had their values changed when visiting 'ifTrue' block.
+    compareValuesInMaps(&copyOfVars, &vars);
+    // Restore old state of map
+    vars = copyOfVars;
+    LOG3("Finished 'IfStatement->ifTrue' block: " << stat->ifTrue);
+    LOG3("Working on 'IfStatement->ifFalse' block: " << stat->ifFalse);
+    visit(stat->ifFalse);
+    // Check if some variables had their values changed when visiting 'ifFalse' block.
+    compareValuesInMaps(&copyOfVars, &vars);
+    // Restore old state of map
+    vars = copyOfVars;
+    LOG3("Finished 'IfStatement->ifFalse' block: " << stat->ifFalse);
+
+    return false;
+}
+
+// Update the value for the 'stat->left' variable.
+bool FindVariableValues::preorder(const IR::AssignmentStatement *stat) {
+    if (!working || lvalue_name(stat->left).isNullOrEmpty())
+        return false;
+
+    LOG5("Working on statement: " << stat);
+    // Remove old values
+    if (vars[lvalue_name(stat->left)] == nullptr ||
+            !(stat->right->equiv(*(vars[lvalue_name(stat->left)]))))
+        removeVarsContaining(&vars, lvalue_name(stat->left));
+    // Set value
+    if (auto lit = stat->right->to<IR::Literal>()) {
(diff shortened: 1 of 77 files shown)


mbudiu-vmw:
I assume this pass works only for actions that are directly called and have exactly 1 call. It won't handle actions that appear in tables. Is that correct?

mbudiu-vmw:
Then you should document this fact: this pass has to be run after LocalizeAllActions, which ensures that each action is invoked only once. If there is an easy way to check this and throw if not, it would be even better.

filipsyrmia:
I looked around in the other passes of the compiler and haven't found a way of doing a check and throw, so I'll add a '@pre' comment to document this dependency.
I'll also add TODO comments for expanding the pass to work on actions inside of tables as well.

Is there anything else that needs to be added/changed?

mbudiu-vmw:
I still have to do an in-depth review, but understanding these preconditions makes the review much easier.
I would not add a TODO necessarily, just specify clearly what the limitations of the pass are.
If you want a TODO you better file an issue.
Frankly I don't think this will apply much to actions in tables.

testdata/p4_16_samples_outputs/psa-resubmit-bmv2-midend.p4
```diff
@@ -36,7 +36,6 @@ parser IngressParserImpl(packet_in pkt, out headers_t hdr, inout metadata_t user
 
 control cIngress(inout headers_t hdr, inout metadata_t user_meta, in psa_ingress_input_metadata_t istd, inout psa_ingress_output_metadata_t ostd) {
     @noWarnUnused @name(".send_to_port") action send_to_port_0() {
-        ostd.drop = false;
```
mbudiu-vmw, line 39: why is this removed?
this looks wrong to me.

mbudiu-vmw (changes requested):

midend/global_copyprop.cpp
```diff
@@ -0,0 +1,324 @@
...
+    if (auto mem = mc->method->to<IR::Member>())
+        return;
+
+    LOG5("Working on 'MethodCallexpression': " << mc);
+    auto *mi = MethodInstance::resolve(mc, refMap, typeMap, true);
+    // Add an entry in the 'actions' map for the action being called.
+    // Remove entries in the 'vars' map for variables that are used as 'Out' or 'InOut' parameters.
+    if (auto aCall = mi->to<ActionCall>()) {
+        LOG6("  Is 'ActionCall'. Adding entry for action: " << aCall->action);
+        (*actions)[aCall->action] = new std::map<cstring, const IR::Expression*>(vars);
```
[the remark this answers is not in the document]
mbudiu-vmw: This is called "defensive programming". We have no way to mechanically check that preconditions are satisfied, so if a pass can validate preconditions it's more robust. We frequently reorder passes or add additional invocations of passes.

testdata/p4_16_samples_outputs/gauntlet_mux_validity-bmv2-midend.p4
```diff
@@ -29,30 +29,20 @@ parser p(packet_in pkt, out Headers hdr, inout Meta m, inout standard_metadata_t
...
-    }
-    @hidden action gauntlet_mux_validitybmv2l37() {
-        b_0 = false;
-    }
-    @hidden table tbl_gauntlet_mux_validitybmv2l37 {
-        actions = {
-            gauntlet_mux_validitybmv2l37();
-        }
-        const default_action = gauntlet_mux_validitybmv2l37();
+        dummy = dummy;
```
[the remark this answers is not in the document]
mbudiu-vmw: repeated would probably be too much, but at least once seems to be very beneficial

midend/local_copyprop.cpp
```diff
@@ -661,4 +661,24 @@ IR::ParserState *DoLocalCopyPropagation::postorder(IR::ParserState *state) {
     return state;
 }
 
+// Reset the state of internal data structures after traversing IR,
+// needed for this pass to function correctly when used in a PassRepeated
+IR::P4Program *DoLocalCopyPropagation::postorder(IR::P4Program *prog) {
```
mbudiu-vmw, line 666: this can be done in the init_apply function.

testdata/p4_16_samples/issue2279.p4
```diff
@@ -0,0 +1,24 @@
+header ethernet_t {
```
mbudiu-vmw, line 1: this is much more readable. thank you.

testdata/p4_16_samples_outputs/psa-register-complex-bmv2-midend.p4
```diff
@@ -37,7 +37,7 @@ control cIngress(inout headers_t hdr, inout metadata_t user_meta, in psa_ingress
         regfile_0.write(32w1, 48w3);
         regfile_0.write(32w2, 48w4);
         tmp = regfile_0.read(32w1);
-        hdr.ethernet.dstAddr = tmp + regfile_0.read(32w2) + 48w281474976710651;
+        hdr.ethernet.dstAddr = regfile_0.read(32w1) + regfile_0.read(32w2) + 48w281474976710651;
```
mbudiu-vmw, line 40: this is a weird one.
many passes assume that there aren't two operations with side-effects in the same statement.
so this one I would think should not happen

mbudiu-vmw (changes requested):
We are very close.
Some of the improvements are very impressive.
The main question is about the side effects.

testdata/p4_16_samples_outputs/psa-register-complex-bmv2-midend.p4
filipsyrmia: The weird thing is that the mechanism used in the LocalCopyPropagation pass (the hasSideEffects class) to detect when a value of a variable has side-effects does not detect that the value of '_tmp_' variable here has any side-effects, this is the reason why it's value does get propagated into the following statement. Since the mechanism used here is the same for all of the passes (I assume the hasSideEffects class is responsible for this throughout the compiler) I don't think there would be any problems with other passes, unless I am missunderstanding something.

The reason this value is not propagated when the pass is run only once is because of the ordering of the statements in the program. Since the ordering be changed arbitrarily, the number of repeated runs of LocalCopyProp pass needed for this kind of behaviour to show can also vary. This is the reason why this situation did not appear before, or at least not consistently enough to be noticed.

If there is still a need to avoid these scenarios, an Inspector can be added in LocalCopyProp which is run on the right side of every AssignmentStatement and sets a flag if it detects that in that right side an ExternMethod exists, and then the LocalCopyPropagation pass can use this flag to determine if it should propagate further variables whose values are also a ExternMethod.

mbudiu-vmw: SideEffects is very simple. We have to figure out why the expression reading the register file is not detected as having side effects. Do you have time to look at this? 

Mihai

filipsyrmia: It seems to have a noSideEffects annotation attached to it.

mbudiu-vmw: So the thing to do is to understand whether the meaning of this annotation is the same as what the compiler worries about. Let me think about it a little bit.

mbudiu-vmw: This is from the spec:
```
@noSideEffects - Weaker than @pure and describes a function that does not change any hidden state, but may depend on hidden state. One example is a hash function that computes a deterministic hash of its arguments, plus some internal state that can be modified via control plane API calls such as a seed or initialization vector. Another example is a read of one element of a register array extern object. Such a function may be dead code eliminated, and may be reordered or combined with other @noSideEffects or @pure calls (subject to the limits imposed by copy-out behavior of out and inout parameters), but not with other function calls that may have side effects that affect the function. 
```

filipsyrmia: If I understand correctly, we need to know if other function calls within a statement affect the function call we are about to propagate. This doesn't seem to be checked anywhere in the LocalCopyProp pass, so how would we achieve this?

mbudiu-vmw: That is a vague statement unfortunately.
The problem is not with this pass, the work done by this pass is correct.
The problem is with other passes, that may expect that there aren't two method calls in the same statement.
It's not easy to review all other passes to make sure that this is what they expect.

mbudiu-vmw: We can leave it like this and see whether anything breaks.
I will approve this for now.

filipsyrmia: I was wondering, can BuiltInMethods be problematic as well when having multiple in one statement?

mbudiu-vmw: BTW: this invariant is established by SideEffectOrdering, early in the compiler.
The invariant is that each statement has at most one side effect. 
But the definition of "side-effect" is subtle, as can be seen from the spec, where `@pure` and `@noSideEffect` are two forms of possible non-side effects. So it is not exactly clear what each pass may expect.

mbudiu-vmw: What is different about BuiltInMethods is that the compiler fully understands what they do. But for extern methods the compiler only has a promise provided by their API and these annotations.

mbudiu-vmw (approved):

testdata/p4_16_samples_outputs/psa-register-complex-bmv2-midend.p4
filipsyrmia: Understood, would you like for me to push the changes for the init_apply function and then merge this or should we skip that as well?

mbudiu-vmw: That is to some degree a cosmetic change. They are equivalent if you always start the visitor from the P4Program node. But if you invoke the visitor on a subtree, they are not equivalent. If you plan to do it I will wait.

filipsyrmia: I will update the PR with that change and then we can move forward with the merge.

PR #2936, opened 2021-10-29
improve include path to bpf_endian.h

trixirt:
On fedora rawhide with clang 13, there are multiple false positve
test failures similar to

p4c/backends/ebpf/runtime/ebpf_kernel.h:28:10: fatal error:
  'bpf_endian.h' file not found

The libbfp-devel package installs bpf_endian.h to /usr/include/bpf/
The ./build_libbpf script installs it to
${CMAKE_CURRENT_SOURCE_DIR}/runtime/usr/include/bpf/

So the include should have a bpf/ prefix.
And since it is a system header the ""'s changed to <>'s

### DIFF
--- a/backends/ebpf/runtime/ebpf_kernel.h
+++ b/backends/ebpf/runtime/ebpf_kernel.h
@@ -25,7 +25,7 @@ limitations under the License.
 
 #include "ebpf_common.h"
 
-#include "bpf_endian.h" // definitions for bpf_ntohs etc...
+#include <bpf/bpf_endian.h> // definitions for bpf_ntohs etc...
 
 #undef htonl
 #undef htons
@@ -47,7 +47,7 @@ limitations under the License.
  */
 #ifdef CONTROL_PLANE // BEGIN EBPF USER SPACE DEFINITIONS
 
-#include "bpf.h" // bpf_obj_get/pin, bpf_map_update_elem
+#include <bpf/bpf.h> // bpf_obj_get/pin, bpf_map_update_elem
 
 #define BPF_USER_MAP_UPDATE_ELEM(index, key, value, flags)\
     bpf_map_update_elem(index, key, value, flags)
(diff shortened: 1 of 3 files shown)


mbudiu-vmw (approved):

rst0git (approved):
LGTM

mbudiu-vmw:
This seems to break the tests on the platforms that we currently test on.

rst0git:
> This seems to break the tests on the platforms that we currently test on.

Thank you for pointing this out. It looks like we need to update the following include path:
```diff
--- a/backends/ebpf/runtime/kernel.mk
+++ b/backends/ebpf/runtime/kernel.mk
@@ -3,7 +3,7 @@ ROOT_DIR := $(dir $(abspath $(lastword $(MAKEFILE_LIST))))
 # Argument for the CLANG compiler
 LLC ?= llc
 CLANG ?= clang
-override INCLUDES+= -I$(ROOT_DIR) -I$(ROOT_DIR)usr/include/bpf/ -I$(ROOT_DIR)contrib/libbpf/include/uapi/
+override INCLUDES+= -I$(ROOT_DIR) -I$(ROOT_DIR)usr/include/ -I$(ROOT_DIR)contrib/libbpf/include/uapi/
 override LIBS+=
 # Optimization flags to save space
 override CFLAGS+= -O2 -g -D__KERNEL__ -D__ASM_SYSREG_H \
```

trixirt:
is -I$(ROOT_DIR)usr/include/ really needed ?
$(ROOT_DIR)usr/include should already be an include path or you would not find much of anything

rst0git:
> is -I$(ROOT_DIR)usr/include/ really needed ? $(ROOT_DIR)usr/include should already be an include path or you would not find much of anything

https://github.com/p4lang/p4c/pull/2403 introduced a git submodule for libbpf that is used for testing.
`ROOT_DIR` is the location of the Makefile (i.e. `./backends/ebpf/runtime`)

trixirt:
Updated with suggested changes

PR #2937, opened 2021-10-31
dpdk: fix memory leak in main()

trixirt:
clang static analysis reports this issue
backends/dpdk/main.cpp:108:13: warning: Potential leak of
  memory pointed to by 'p4rt'

When out fails to open and returns, it does not free p4rt.

Because out only depends on options, move the out assignment
to earlier in the block.


mbudiu-vmw:
Have you seen many free calls in the code? There aren't any. We use the c++ garbage collection library.

PR #2938, opened 2021-11-02
"Null element" errors which are related with HeaderUnion type

vladyslav-dubina:
Solving "Null element" errors  which are related with HeaderUnion type.
Tests for example : issue561-7-bmv2.p4 , issue561-4-bmv2.p4,  issue561-5-bmv2.p4,  issue561-6-bmv2.p4.
Use `--loopsUnroll` argument for programs starting

### DIFF
--- a/midend/interpreter.cpp
+++ b/midend/interpreter.cpp
@@ -342,16 +342,16 @@ SymbolicValue* SymbolicHeaderUnion::clone() const {
 
 void SymbolicHeaderUnion::assign(const SymbolicValue* other) {
     if (other->is<SymbolicError>()) return;
-    BUG_CHECK(other->is<SymbolicHeaderUnion>(), "%1%: expected a header union", other);
     auto hv = other->to<SymbolicHeaderUnion>();
+    BUG_CHECK(hv, "%1%: expected a header union", other);
     for (auto f : hv->fieldValue)
         fieldValue[f.first]->assign(f.second);
     valid->assign(hv->valid);
 }
 
 bool SymbolicHeaderUnion::merge(const SymbolicValue* other) {
-    BUG_CHECK(other->is<SymbolicHeaderUnion>(), "%1%: expected a header union", other);
     auto hv = other->to<SymbolicHeaderUnion>();
+    BUG_CHECK(hv, "%1%: expected a header union", other);
     bool changes = false;
     for (auto f : hv->fieldValue)
         changes = changes || fieldValue[f.first]->merge(f.second);
--- a/midend/interpreter.cpp
+++ b/midend/interpreter.cpp
@@ -777,7 +777,7 @@ void ExpressionEvaluator::postorder(const IR::Operation_Ternary* expression) {
         clone->e1 = getConstant(e1i);
         clone->e2 = getConstant(e2i);
         DoConstantFolding cf(refMap, typeMap);
-        //cf.setCalledBy(this);
+        cf.setCalledBy(this);
         auto result = clone->apply(cf);
         checkResult(expression, result);
         return;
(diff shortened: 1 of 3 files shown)


fruffy (commented):
Can not really assess whether the implementation makes sense but can you add a gtest for this? Since we can not test this implementation using the normal p4 test suite.

midend/interpreter.cpp
```diff
@@ -309,6 +309,81 @@ void SymbolicStruct::dbprint(std::ostream& out) const {
...
+    BUG_CHECK(other->is<SymbolicHeaderUnion>(), "%1%: expected a header union", other);
+    auto hv = other->to<SymbolicHeaderUnion>();
+    for (auto f : hv->fieldValue)
+        fieldValue[f.first]->assign(f.second);
+    valid->assign(hv->valid);
+}
+
+bool SymbolicHeaderUnion::merge(const SymbolicValue* other) {
+    BUG_CHECK(other->is<SymbolicHeaderUnion>(), "%1%: expected a header union", other);
+    auto hv = other->to<SymbolicHeaderUnion>();
```
fruffy, line 354: Used checkedTo for all of these checks.

test/gtest/parser_unroll.cpp
```diff
@@ -311,4 +311,12 @@ TEST_F(P4CParserUnroll, t1_Cond) {
     ASSERT_EQ(parsers.first->states.size(), parsers.second->states.size());
 }
 
+TEST_F(P4CParserUnroll, header_union) {
+    auto parsers =  loadExample("issue561-7-bmv2.p4");
+    ASSERT_TRUE(parsers.first);
+    ASSERT_TRUE(parsers.second);
+    // adding outOfBound state
```
fruffy, line 318: What does this mean? Why is the number of parser states one less?

fruffy (changes requested):
Not much to say about this. @mbudiu-vmw do you have any comments on this one?

test/gtest/parser_unroll.cpp
vladyslav-dubina: This is because the parser adds an outOfBound state, which is correct. If we look at parser-unroll-test1.p4 we see similar behavior, as far as I understand it is caused by hitting similar parts of the code "return new SymbolicException (node, P4 :: StandardExceptions :: StackOutOfBounds);"

midend/interpreter.cpp
vladyslav-dubina: `P4::SymbolicValue` has no `checkedTo ` function

fruffy: If it is not too much effort, I would just add it then. Clearly, there is a use case. Alternatively, do the cast with `to` first, then check if `to` is a nullptr. So you do not have to cast twice.

test/gtest/parser_unroll.cpp
fruffy: But the tests says `ASSERT_EQ(parsers.first->states.size(), parsers.second->states.size() - 1)`;.
What states are removed here and why is this test correct? Is it because you do expect that executing the test does not crash and you get the right amount of parser states?  
In general, the explanations you provide here are helpful and should also be part of the test itself as part of comments.

midend/interpreter.cpp
```diff
@@ -309,6 +309,81 @@ void SymbolicStruct::dbprint(std::ostream& out) const {
     out << " }";
 }
 
+SymbolicHeaderUnion::SymbolicHeaderUnion(const IR::Type_HeaderUnion* type,
```
fruffy, line 312: How much of this code is duplicated from SymbolicHeader?

test/gtest/parser_unroll.cpp
vladyslav-dubina: There I removed the outOfBound state, similar to other tests. Well, I'll add a more extensive comment to the code to make it clearer what exactly I removed.

midend/interpreter.cpp
vladyslav-dubina: Ok, i`ll do this

midend/interpreter.cpp
vladyslav-dubina: This code is based on SymbolicHeader as they are quite similar

fruffy:
I assumed you had no comments on this one since it had been open for a while.

PR #2942, opened 2021-11-03
ParserState skipping

vladyslav-dubina:
Skips ParserState in parserUnroll pass.
This is necessary due to the fact that the appearance of 
```
state start @name(".$start") {
    select{(bit<32>)standard_metadata.instance_type} {
      0: start_0
      1: start_e2e_mirrored
      2: start_i2e_mirrored
      default: noMatch } }
state start_0 @name(".start") {
    accept; }
```
is similar to the bug of the translator from p4_14 to p4_16 standard.
With these changes, parser_pragma2.p4 should pass without errors.

### DIFF
--- a/midend/parserUnroll.cpp
+++ b/midend/parserUnroll.cpp
@@ -262,6 +262,9 @@ class ParserSymbolicInterpreter {
             result->set(p, value);
         }
         for (auto d : parser->parserLocals) {
+            if (d->to<IR::ParserState>()) {
+                continue;
+            }
             auto type = typeMap->getType(d);
             SymbolicValue* value = nullptr;
             if (d->is<IR::Declaration_Constant>()) {


mbudiu-vmw:
What bug are you referring to?

mbudiu-vmw:
If you look at the declaration of P4Parser you will notice that a ParserState can not be part of parserLocals.
https://github.com/p4lang/p4c/blob/ecf04030942ce44eb83f52e982f93b48547be08b/ir/ir.def#L87
So your change does nothing.

vladyslav-dubina:
If we look at this PR https://github.com/p4lang/p4c/pull/2921
We will see an error in parser_pragma2.p4. Analysis of this error showed that parserState gets here https://github.com/vladyslav-dubina/p4c/blob/NotFoundWithParserState/midend/parserUnroll.cpp#L265 , and then in getType, which causes the error.  We believe that this is a translation bug from p4_14 to p4_16. Therefore, we skipped ParserState because it does not make sense for the specified part of parserUnroll

vladyslav-dubina:
The main problem is that `ParserState ` gets to `parserLocals`. That I am a mistake. In addition, as far as I can judge, the `ParserState ` that got into `parserLocals ` is also a duplicate. The presence of `ParserState ` in `parserLocals ` results in an error in `getType`. Can we somehow check if there is a fix for this problem in one of your PRs (if I understood everything correctly)?

fruffy:
>We believe that this is a translation bug from p4_14 to p4_16.

Is there any way to reproduce this?

vladyslav-dubina:
You can run parser_pragma2.p4 with the --loopsUnroll argument. Will result in "Not found" ( https://github.com/vladyslav-dubina/p4c/blob/736df9ebdc219f10801ba5675818728604467746/frontends/p4/typeMap.cpp#L123 ) error.  It is caused by the fact that ParserState ( https://github.com/vladyslav-dubina/p4c/blob/cfebb42ceeee5bdfed97433f07afdc101c396b82/midend/parserUnroll.cpp#L265 ) gets to parserUnroll. This should not happen, especially since it is a duplicate.

```
state start @name(".$start") {
    select{(bit<32>)standard_metadata.instance_type} {
      0: start_0
      1: start_e2e_mirrored
      2: start_i2e_mirrored
      default: noMatch } }
```

mbudiu-vmw:
I will close this PR since the correct solution is in #2953. It was a bug in the p4 14->16 converter

PR #2943, opened 2021-11-03
DPDK Backend: Add support for switch statement in p4c-dpdk

usha1830:
This PR adds support for switch statement in p4c-dpdk

- This assumes that the EliminateSwitch pass is run and only expression in switch is of type tbl.apply().action.run
- Couple of tests started passing because of these changes. Adding reference outputs for these tests and updated the XFAIL list.
- Each case within switch is converted to **jmpa** instruction (jump on action run)

### DIFF
--- a/backends/dpdk/dpdk.def
+++ b/backends/dpdk/dpdk.def
@@ -151,6 +151,29 @@ class DpdkJmpMissStatement : DpdkJmpStatement {
         DpdkJmpStatement("jmpnh", label) { }
 }
 
+abstract DpdkJmpActionStatement : DpdkJmpStatement {
+    cstring action;
+    std::ostream& toSpec(std::ostream& out) const override;
+#nodbprint
+#noconstructor
+    DpdkJmpActionStatement(cstring instruction, cstring label, cstring action) :
+        DpdkJmpStatement(instruction, label), action(action) { }
+}
+
+class DpdkJmpOnActionRunStatement : DpdkJmpActionStatement {
+#nodbprint
+#noconstructor
+    DpdkJmpOnActionRunStatement(cstring label, cstring act) :
+        DpdkJmpActionStatement("jmpa", label, act) { }
+}
+
+class DpdkJmpNotOnActionRunStatement : DpdkJmpActionStatement {
+#nodbprint
+#noconstructor
+    DpdkJmpNotOnActionRunStatement(cstring label, cstring act) :
+        DpdkJmpActionStatement("jmpna", label, act) { }
+}
+
 abstract DpdkJmpHeaderStatement : DpdkJmpStatement {
     Expression header;
     std::ostream& toSpec(std::ostream& out) const override;
--- a/backends/dpdk/dpdk.def
+++ b/backends/dpdk/dpdk.def
@@ -160,17 +160,17 @@ abstract DpdkJmpActionStatement : DpdkJmpStatement {
         DpdkJmpStatement(instruction, label), action(action) { }
 }
 
-class DpdkJmpOnActionRunStatement : DpdkJmpActionStatement {
+class DpdkJmpIfActionRunStatement : DpdkJmpActionStatement {
 #nodbprint
 #noconstructor
-    DpdkJmpOnActionRunStatement(cstring label, cstring act) :
+    DpdkJmpIfActionRunStatement(cstring label, cstring act) :
         DpdkJmpActionStatement("jmpa", label, act) { }
 }
 
-class DpdkJmpNotOnActionRunStatement : DpdkJmpActionStatement {
+class DpdkJmpIfActionNotRunStatement : DpdkJmpActionStatement {
 #nodbprint
 #noconstructor
-    DpdkJmpNotOnActionRunStatement(cstring label, cstring act) :
+    DpdkJmpIfActionNotRunStatement(cstring label, cstring act) :
         DpdkJmpActionStatement("jmpna", label, act) { }
 }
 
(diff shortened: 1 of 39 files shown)


backends/dpdk/dpdk.def
```diff
@@ -151,6 +151,29 @@ class DpdkJmpMissStatement : DpdkJmpStatement {
...
+}
+
+class DpdkJmpOnActionRunStatement : DpdkJmpActionStatement {
+#nodbprint
+#noconstructor
+    DpdkJmpOnActionRunStatement(cstring label, cstring act) :
+        DpdkJmpActionStatement("jmpa", label, act) { }
+}
+
+class DpdkJmpNotOnActionRunStatement : DpdkJmpActionStatement {
```
mbudiu-vmw, line 170: that' an awkward name

backends/dpdk/dpdkArch.cpp
```diff
@@ -1526,6 +1526,8 @@ const IR::Node* SplitP4TableCommon::postorder(IR::IfStatement* statement) {
 
 const IR::Node* SplitP4TableCommon::postorder(IR::SwitchStatement* statement) {
     auto expr = statement->expression;
+    if (!expr->is<IR::Member>())
```
mbudiu-vmw, line 1529: could swap these or check if member is null below.
There is also a set of utilities in https://github.com/p4lang/p4c/blob/main/frontends/p4/tableApply.h that make it easier to detect this pattern.

backends/dpdk/dpdkHelpers.cpp
```diff
@@ -16,6 +16,7 @@ limitations under the License.
 #include <iostream>
 #include "dpdkHelpers.h"
 #include "ir/ir.h"
+#include "frontends/p4/tableApply.h"
```
mbudiu-vmw, line 19: I guess you know about it.

backends/dpdk/dpdkHelpers.cpp
```diff
@@ -681,9 +682,52 @@ bool ConvertStatementToDpdk::preorder(const IR::MethodCallStatement *s) {
...
+        } else {
+            auto pe = caseLabel->label->to<IR::PathExpression>();
+            CHECK_NULL(pe);
+            label = Util::printf_format("label_action%d", next_label_id++);
+            add_instr(new IR::DpdkJmpOnActionRunStatement(label, pe->path->name.name));
+        }
+        labels.push_back(label);
+    }
+
+    if (!isdefaultPresent)
```
mbudiu-vmw, line 714: don't we have a pass which ensures that this is always true (or that the cases are exhaustive?)

backends/dpdk/dpdkHelpers.cpp
```diff
@@ -681,9 +682,52 @@ bool ConvertStatementToDpdk::preorder(const IR::MethodCallStatement *s) {
...
+    auto size = s->cases.size();
+    std::vector <cstring> labels;
+    cstring label;
+    bool isdefaultPresent = false;
+    cstring default_label = "";
+    auto end_label = Util::printf_format("label_endswitch%d", next_label_id++);
+    // Emit jmp on action run statements and collect labels for each case statement
+    for (unsigned i = 0; i < size; i++) {
+        auto caseLabel = s->cases.at(i);
+        if (caseLabel->label->is<IR::DefaultExpression>()) {
```
mbudiu-vmw, line 700: we definitely have a pass that makes sure the default label is always last if it exists.

backends/dpdk/dpdkHelpers.cpp
```diff
@@ -681,9 +682,52 @@ bool ConvertStatementToDpdk::preorder(const IR::MethodCallStatement *s) {
...
+            label = Util::printf_format("label_default%d", next_label_id++);
+            add_instr(new IR::DpdkJmpLabelStatement(label));
+            default_label = label;
+            isdefaultPresent = true;
+        } else {
+            auto pe = caseLabel->label->to<IR::PathExpression>();
+            CHECK_NULL(pe);
+            label = Util::printf_format("label_action%d", next_label_id++);
```
mbudiu-vmw, line 708: are you sure these label names are unique?
could you use our existing tools to generate unique names (but then you have to do it for everything).

mbudiu-vmw (approved):

backends/dpdk/dpdkHelpers.cpp
usha1830: There is a frontend pass for adding default case. However, later the midend pass EliminateSwitch converts switch statements with other expressions to switch case with action_run expression. These newly created switch statements do not have a default case, the default case is also converted to an action and that action is marked as the const default action for the switch table.
https://github.com/p4lang/p4c/blob/main/midend/eliminateSwitch.h#L73

backends/dpdk/dpdkHelpers.cpp
usha1830: Yes. I updated the code to move the last case handling outside the loop.

backends/dpdk/dpdkHelpers.cpp
usha1830: Updated the code to use refmap->newName utility function to generate unique names.

usha1830:
@mbudiu-vmw Thanks for the review! I have updated the code as per your comments. Please take a look.

backends/dpdk/dpdkHelpers.cpp
mbudiu-vmw: One solution is then to rerun this pass after the switch statements are eliminated...

mbudiu-vmw (approved):

PR #2948, opened 2021-11-11
DPDK Backend: Fix minor issues in handling Unary operations

usha1830:
The changes include the following:

1) Fix handling of Unary operations in ExpressionUnroll and LogicalExpressionUnroll passes.
2) Remove explicit preorder function for Cast. Cast expression should be checked for further  unrolling and hence the preorder function for Cast operation would exactly be the same as the one for Unary operations.
3) Update XFAIL list and reference outputs for the tests that started passing with these changes.


mbudiu-vmw (approved):

usha1830:
@mbudiu-vmw Thanks for the review. Can this be merged?

PR #2953, opened 2021-11-19
Pragma fix

mbudiu-vmw:
A rarely used code path inside the p4-14 to p4-16 conversion was inserting parser states in the local variables instead of the states array. This is the real fix to the problem reported in #2942.

### DIFF
--- a/midend/global_copyprop.cpp
+++ b/midend/global_copyprop.cpp
@@ -51,7 +51,7 @@ void compareValuesInMaps(std::map<cstring, const IR::Expression*> *oldValues,
     for (auto it : *newValues) {
         auto oldValue = (*oldValues)[it.first];
         if (((it.second == nullptr) ^ (oldValue == nullptr)) ||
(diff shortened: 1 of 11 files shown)


midend/global_copyprop.cpp
```diff
@@ -51,7 +51,7 @@ void compareValuesInMaps(std::map<cstring, const IR::Expression*> *oldValues,
     for (auto it : *newValues) {
         auto oldValue = (*oldValues)[it.first];
         if (((it.second == nullptr) ^ (oldValue == nullptr)) ||
-                it.second && oldValue && !(it.second->equiv(*oldValue)))
+            (it.second && oldValue && !(it.second->equiv(*oldValue))))
```
fruffy, line 54: Does this fix something else or are these just cosmetic changes?

fruffy (approved):

midend/global_copyprop.cpp
mbudiu-vmw: this is a clang warning

PR #2954, opened 2021-11-19
Def-use should not treat array.next as a structure field

mbudiu-vmw:


mbudiu-vmw:
Without this fix the test program would crash the compiler.

mbudiu-vmw:
@fruffFri am surprised your validator does not like this

fruffy:
Yes, it is fixed now This was code that was poorly transposed in the C++ rewrite. I was not aware that you could also write to `.next` and `.last`.

fruffy (approved):

PR #2955, opened 2021-11-19
DPDK Backend: Fix a corner case in handling isValid() as table key

usha1830:
DPDK assembly does not have a syntax for having isValid() as a table key. This expression should be moved to metadata and the metadata should be used as table key.

This PR fixes a case where all the table keys are isValid() method calls and the isValid() expression is not moved to metadata.
- Added a new test and reference outputs for this case.


mbudiu-vmw (approved):

usha1830:
@mbudiu-vmw Thanks for reviewing. Could you please merge this PR?

PR #2962, opened 2021-11-24
Continue unrolling for parser after evaluation error

VolodymyrPeschanenkoIntel:
These changes allow to continue unrolling after evaluation error for one of the statements. These changes fix a bug which removes some states from a parser (issue1520-bmv2, issue1937-2-bmv2, invalid-hdr-warnings1).


midend/interpreter.h
```diff
@@ -41,6 +41,8 @@ class SymbolicValue {
     virtual void dbprint(std::ostream& out) const = 0;
     template<typename T> T* to() {
         auto result = dynamic_cast<T*>(this);
+                if (result == nullptr)
+            std::cout << 1;
```
rst0git: The indentation on these lines looks incorrect.

mbudiu-vmw:
Let us know when this is ready for review.

VolodymyrPeschanenkoIntel:
@mbudiu-vmw, It is ready for review

midend/parserUnroll.cpp
```diff
@@ -280,7 +282,7 @@ class ParserSymbolicInterpreter {
             if (value == nullptr)
                 value = factory->create(type, true);
             if (value->is<SymbolicError>()) {
-                ::error(ErrorType::ERR_EXPRESSION,
+                ::warning(ErrorType::ERR_EXPRESSION,
```
mbudiu-vmw, line 285: is there a way to distinguish between possible errors and certain errors? We want to give warnings for the former and errors for the latter. ("may" and "must" in compiler terminology.)

test/gtest/parser_unroll.cpp
```diff
@@ -318,7 +318,7 @@ TEST_F(P4CParserUnroll, header_union) {
     // The extra state in parsers.second is the outOfBound state that's correct.
     // But due to the fact that parsers.first does not have such an outOfBound state,
     // we have to subtract it from parsers.second.
-    ASSERT_EQ(parsers.first->states.size(), parsers.second->states.size() - 1);
+    ASSERT_EQ(parsers.first->states.size(), parsers.second->states.size());
```
mbudiu-vmw, line 321: is the comment still accurate?

midend/parserUnroll.cpp
mbudiu-vmw: There are some comments on SymbolicError and SymbolicStaticError which suggest that this is indeed true.
But the comments on SymbolicError seem wrong.

mbudiu-vmw (commented):
This looks fine, but I will wait for an answer to my questions to approve.

midend/parserUnroll.cpp
VolodymyrPeschanenkoIntel: I think that ParserUnroll should generate some message which helps user to understand the reason for unsuccessful unrolling. In this case a program should continue execution, because it is just a problem of interpretation inside parserUnroll.

mbudiu-vmw (approved):

PR #2966, opened 2021-11-29
DPDK Backend: Convert error codes to its constant values

usha1830:
The error constants defined using `error { }` should be replaced with the constant values for DPDK backend.

- Existing class BranchingInstructionGeneration (with minor additions ) is reused for generating jumps on error code comparison.
- As  error type only supports equality (==) and inequality (!=) comparisons. The operands of these operations are checked.
- Added new testcases and corresponding reference outputs.


mbudiu-vmw (commented):
I think we should have a general midend pass with a policy, like the one which eliminates enum. Why can't you just convert enums to short Integers?

usha1830:
> I think we should have a general midend pass with a policy, like the one which eliminates enum. Why can't you just convert enums to short Integers?

I first tried to implement this similar to eliminateSerEnums midend pass. 
```
const IR::Node* DoEliminateErrorCode::preorder(IR::Type_Error* err) {
    return new IR::Type_Bits(32, false);
}

```
However, when I tried to change the type of variables of `error` type to` bit<W>` type, the typechecker failed with the error
_"At this point in the compilation typechecking should not infer new types anymore, but it did."_

I also tried writing a backend pass to achieve this, but the typechecker failed with the same error as above. 
Am I doing something wrong? Is there a way to change the type of existing struct fields and other local variables?

Thanks!

hanw:
From my experience, typically, the error _"At this point in the compilation typechecking should not infer new types anymore, but it did"_ occurs when you add a variable of Type_InfInt to the program and was later inferred to Type_Bits.

mbudiu-vmw:
That too, if you replace with constants make sure they have a width

usha1830:
> From my experience, typically, the error _"At this point in the compilation typechecking should not infer new types anymore, but it did"_ occurs when you add a variable of Type_InfInt to the program and was later inferred to Type_Bits.

This maybe the issue. 
I replaced the error constant with  _IR::Constant(error_id)_ which is Type_InfInt

usha1830:
> That too, if you replace with constants make sure they have a width



> That too, if you replace with constants make sure they have a width

How do we specify width for constants?

mbudiu-vmw:
Use the right constructor of IR::constant

usha1830:
@mbudiu-vmw , @hanw Thanks for the pointers. The typechecker error is gone. Running into some other issues. Will update the PR after fixing the issues.

usha1830:
I have updated the PR with a midend pass. The code is similar to convertEnums midend pass, but this new pass should be called towards the end of the Midend passes, so that error types created by other passes (such as HandleNoMatch, ParserUnroll etc) are also converted.

backends/dpdk/midend.cpp
```diff
@@ -89,10 +90,26 @@ class EnumOn32Bits : public P4::ChooseEnumRepresentation {
     EnumOn32Bits() {}
 };
 
+/**
+This class implements a policy suitable for the ConvertErrors pass.
+The policy is: convert all errors to bit<16>
+*/
+class ErrorBits : public P4::ChooseErrorRepresentation {
```
mbudiu-vmw, line 97: I would call this ErrorOn16Bits or something similar.
Alternatively, the class should be called ErrorWidth and have a constructor argument for the width.

midend/convertErrors.cpp
```diff
@@ -0,0 +1,56 @@
...
+    BUG_CHECK(count <= (1ULL << width),
+              "%1%: not enough bits to represent %2%", width, type);
+    errorRepr = new ErrorRepresentation(type->srcInfo, width);
+    for (auto d : type->members)
+        errorRepr->add(d->name.name);
+    return new IR::Type_Bits(width, false);
+}
+
+const IR::Node* DoConvertErrors::postorder(IR::Type_Name* type) {
+    if (type->toString() == IR::Type_Error::error) {
```
mbudiu-vmw, line 36: In general relying on strings is not safe, you should probably use equiv or the typeMap type equivalence checker.

midend/convertErrors.cpp
```diff
@@ -0,0 +1,56 @@
...
+            return type;
+        else
+            return new IR::Type_Bits(16, false);
+    }
+    return type;
+}
+
+/// process error expression, e.g., X.a
+const IR::Node* DoConvertErrors::postorder(IR::Member* expression) {
+    if (expression->expr->toString() == "error") {
```
mbudiu-vmw, line 48: Do not use strings. Rather check the type of the expression to see if it's Error.

midend/convertErrors.cpp
```diff
@@ -0,0 +1,56 @@
+/* Copyright 2021 Intel Corporation
```
mbudiu-vmw, line 1: There is an alternative way to do this which you may find simpler: convert the Error to a serializable Enum and the rely on the serializable enum conversion pass. This should be simpler than what you have here.

mbudiu-vmw (changes requested):

mbudiu-vmw:
Thinking of this I believe that the policy should also include a function to assign a numeric value to each identifier. Then backends that have hardwired values for the error codes can still use this pass.

mbudiu-vmw (approved):

PR #2967, opened 2021-11-29
Add minWidthBits for Type_Error

usha1830:


mbudiu-vmw:
I am afraid this does not confirm to the spec

usha1830:
> I am afraid this does not conform to the spec

If we do not handle _Type_Error_ here, it leads to an error _"width not well-defined"_ while generating P4Info in certain cases. 
Any suggestion on how we can fix this? Should we avoid calling this function **minWidthBits()** when the type is _Type_Error_?

mbudiu-vmw:
That is my point, I belive that the spec days that it is an error to call this function on values of type error.

mbudiu-vmw:
Why does the compiler call width on an error type?

usha1830:
> Why does the compiler call width on an error type?

In P4info, each table match key has a field for bitwidth. When an error type variable is used as match key field of a table, this function is called.


```
table parser_error {
    key = {
        istd.parser_error : exact;
    }
   // rest of the table properties
}
```

P4info:		
 ```
match_fields {              
      id: 1                     
      name: "istd.parser_error" 
	  bitwidth : 0
      match_type: EXACT         
    }   
```

From the P4Runtime spec (section 8.5.5), 
"_For enum types with no underlying type — as well as error — there is no integer value associated with each symbolic member entry (whether assigned automatically by the compiler or directly in the P4 source). We therefore use a human-readable string in P4Data to represent enum and error values."_

As an error type does not have any underlying bitwidth (which is what the error is), I think we can treat the error type as a string.

mbudiu-vmw (approved):

PR #2970, opened 2021-11-30
Dpdk: skip ExpandLookAhead pass in dpdk midend.

hanw:
Dpdk datapath implements lookahead on a header instance directly, e.g.:

`headers.gtop = packet.lookahead<gtp1_top_t>();`

is supported as 

`lookahead h.gtop`

MidEnd does not need to expand lookahead on a header type into lookahead into raw bits and assign bit slices to header fields.


hanw:
This PR uncovers a bug in the ParsersUnroll pass where it incorrectly eliminated statement from a parser state, hence the big difference in the reference output.

mbudiu-vmw (approved):

PR #2971, opened 2021-12-01
Compiler Bug : overwriting definitions in defuse (#2900)

### LINKED ISSUE #2900: Compiler Bug: overwriting definitions in defuse analysis

Jovana-Syrmia:
Fixes #2900 

Bug in file def_use in method setDefinitionsAt class AllDefinitions
 In that method definition in map atPoint is overridden

This happens beacuse there is more than one AsssignStatement with the same id
Pass KeySideEffect is creating those AssignStatements (in method KeySideEffect::doStatement)
This method creates a BlockStatement and inserts AssignStatements from TableInsertions variable and then inserts that block before tbl.apply()

Bug is removed by cloning an AssignStatement before inserting it in a BlockStatement

Backend pass CopyMatchKeysToSingleStruct inherits KeySideEffect
Backend uses this pass to transform tables such that all the Match keys are part of the same header/metadata struct
With cloning two test fail (psa-dpdk-table-key-consolidation-if*) in spec file
This happens because for every assignment DpdkMovStatement is created and inserted in instractions (pass ConvertStatementToDpdk)
There is no need for this new DpdkMovStatements because src is already in dest (it is done the first time we call tbl.apply)

This is fixed by setting SideEffect::doStatement to be virtual and  overriding it in class CopyMatchKeysToSingleStruct
In this override there is no cloning

### DIFF
--- a/backends/dpdk/dpdkArch.cpp
+++ b/backends/dpdk/dpdkArch.cpp
@@ -1038,6 +1038,24 @@ const IR::Node* CopyMatchKeysToSingleStruct::postorder(IR::KeyElement* element) 
     }
     return element;
 }
+const IR::Node* CopyMatchKeysToSingleStruct::doStatement(const IR::Statement* statement,
+                                           const IR::Expression *expression) {
+    LOG3("Visiting " << getOriginal());
+    P4::HasTableApply hta(refMap, typeMap);
+    hta.setCalledBy(this);
+    (void)expression->apply(hta);
+    if (hta.table == nullptr)
+        return statement;
+    auto insertions = get(toInsert, hta.table);
+    if (insertions == nullptr)
+        return statement;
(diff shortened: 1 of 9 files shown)


mbudiu-vmw (approved):

mbudiu-vmw:
The problem is not that two nodes have the same id, the problem is that the to IR nodes are in fact the same exact node. We keep pointers to nodes in hash-tables, so a node cannot be mapped to two different values in the hash-table. See also the discussion from #2969

PR #2972, opened 2021-12-01
Add temporary IR nodes to typeMap

### LINKED ISSUE #2958: Compiler Bug: Could not find type for <Member>
Compiling with 
```
fruffy@fruffy /s/P/p4_tv (master) [1]> p4test --version                                                        
p4test
Version 1.2.2 (SHA: d5ea272e0 BUILD: RelWithDebInfo)
```
for the following program: 

[triple_nested_header.p4.txt](https://github.com/p4lang/p4c/files/7575203/triple_nested_header.p4.txt)

Leads to a crash bug:

```bash
FrontEnd_36_MoveDeclarations
SimplifyDefUse_0_TypeChecking
In file: p4c/frontends/p4/typeMap.cpp:126
Compiler Bug: Could not find type for <Member>(18751) .g
```

mbudiu-vmw:
Fixes #2958

### DIFF
--- a/frontends/p4/simplifyDefUse.cpp
+++ b/frontends/p4/simplifyDefUse.cpp
@@ -714,9 +714,10 @@ class FindUninitialized : public Inspector {
             if (auto list = src->to<IR::StructExpression>()) {
                 auto it = list->components.begin();
                 for (auto field : dst_struct->fields) {
-                    IR::Member member(dst, field->name);
-                    processHeadersInAssignment(&member, (*it)->expression,
-                                               typeMap->getType(field, true),
+                    auto ftype = typeMap->getType(field, true);
+                    auto member = new IR::Member(dst, field->name);
+                    typeMap->setType(member, ftype);
+                    processHeadersInAssignment(member, (*it)->expression, ftype,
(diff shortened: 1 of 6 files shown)


mbudiu-vmw:
@kostasyrmia please review this; it looks like the bug is in your code.

kostasyrmia:
Yes, you are right, this is definitely in my code.

I tried to avoid inserting these temporary fields into the typeMap, but it seems it has to be done...

The only thing that I would do additionally is removing `dst_type` and `src_type` from the signature of `processHeadersInAssignment`, because they no longer have any usage now that everything is in the typeMap.

I think that this:
```
void processHeadersInAssignment(const IR::Expression* dst, const IR::Expression* src,
                                    const IR::Type* dst_type, const IR::Type* src_type) {
    // code inside the function
}
```
can be changed to something like this:
```
void processHeadersInAssignment(const IR::Expression* dst, const IR::Expression* src) {
    auto dst_type = typeMap->getType(dst, true);
    auto src_type = typeMap->getType(src, true);

    // code inside the function
}
```

But also it's okay for me as it is.

Thank you for the fix!

fruffy (approved):

Issue #2956, opened 2021-11-20
GlobalCopyPropagation is too aggressive in switch statements


fruffy:
GlobalCopyPropagation transforms the following program segment

```p4
control ingress(inout Headers h) {
    @noWarn("unused") @name(".NoAction") action NoAction_1() {
    }
    @name("ingress.simple_assign") action simple_assign() {
        h.eth_hdr.eth_type = 16w1;
    }
    @name("ingress.simple_assign") action simple_assign_1() {
        h.eth_hdr.eth_type = 16w1;
    }
    @name("ingress.simple_assign") action simple_assign_2() {
        h.eth_hdr.eth_type = 16w1;
    }
    @name("ingress.dummy_table") table dummy_table_0 {
        key = {
            h.eth_hdr.src_addr: exact @name("key") ;
        }
        actions = {
            NoAction_1();
            simple_assign();
        }
        default_action = NoAction_1();
    }
    apply {
        switch (dummy_table_0.apply().action_run) {
            simple_assign: {
                simple_assign_1();
            }
            NoAction_1: {
            }
        }
        simple_assign_2();
    }
}
```
into 
```p4
control ingress(inout Headers h) {
    @noWarn("unused") @name(".NoAction") action NoAction_1() {
    }
    @name("ingress.simple_assign") action simple_assign() {
        h.eth_hdr.eth_type = 16w1;
    }
    @name("ingress.simple_assign") action simple_assign_1() {
        h.eth_hdr.eth_type = 16w1;
    }
    @name("ingress.simple_assign") action simple_assign_2() {
        ;
    }
    @name("ingress.dummy_table") table dummy_table_0 {
        key = {
            h.eth_hdr.src_addr: exact @name("key") ;
        }
        actions = {
            NoAction_1();
            simple_assign();
        }
        default_action = NoAction_1();
    }
    apply {
        switch (dummy_table_0.apply().action_run) {
            simple_assign: {
                simple_assign_1();
            }
            NoAction_1: {
            }
        }
        simple_assign_2();
    }
}
```
But this may be too aggressive. The table may not necessarily execute `NoAction_1`. 

[global_copy_prop_switch.p4.txt](https://github.com/p4lang/p4c/files/7576518/global_copy_prop_switch.p4.txt)

jfingerh:
It seems true that the table must either execute `NoAction_1` or `simple_assign`, and in either case, the field is assigned a value of 1.  But maybe I am missing something.

[...]

PR #2974, opened 2021-12-02
GlobalCopyPropagation is too agressive in switch statements (#2956)

Jovana-Syrmia:
Fixes issue #2956 
* Added preorder of switch statement for FindVariableValue pass
* There is no need for these in DoGlobalCopyPropagation because this pass operates on action bodies only and  switch statements aren't allowed within an action

### DIFF
--- a/midend/global_copyprop.cpp
+++ b/midend/global_copyprop.cpp
@@ -111,6 +111,23 @@ bool FindVariableValues::preorder(const IR::IfStatement *stat) {
     return false;
 }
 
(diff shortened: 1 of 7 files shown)


mbudiu-vmw (approved):

mbudiu-vmw:
Please add the failing example as a test case.

mbudiu-vmw (changes requested):
Test case is missing.

mbudiu-vmw (approved):

PR #2985, opened 2021-12-13
v1model metadata preservation updates

jafingerhut:


jafingerhut:
Some of this is simply reforming paragraphs to fit within 72/80 column width, which I know is annoying to review, but I do believe it makes it easier to read, and for comments, people reading them is their primary purpose.

Also several replacements of the old names to new names, e.g. recirculate->recirculate_preserving_field_list

I also updated a fair number of comments, but only comments, in one of the v1model test programs that exercises recirculate, resubmit, and clone heavily.

jafingerhut:
Pinging @smolkaj @kheradmandG who were involved in the p4c changes that led to these documentation changes, to see if they have any comments or suggestions here.

mbudiu-vmw (approved):
I only checked the changes. I didn't read the unmodified text to see whether more changes are needed.

PR #2991, opened 2021-12-15
DPDK Backend: Update DPDK Readme.md

usha1830:


backends/dpdk/README.md
```diff
@@ -20,17 +22,48 @@ generates the 'spec' file to configure the DPDK pipeline.
...
+- Direct Meter
+- Direct Counter
+- Basic Checksum
+- Packet Cloning/Recirculation/Resubmission
+
+### DPDK target limitations
+- Currently, programs written for DPDK target should limit the functionality in Ingress blocks.  When egress block support gets added to the DPDK target, compiler should generate separate spec file for ingress and egress.
+- DPDK architecture assumes the following signatures for programmable block of PSA. P4C-DPDK converts the input program to this form.
+
+```bash
```
rst0git, line 51: GitHub supports syntax highlighting for P4, you can use it with ```P4

backends/dpdk/README.md
```diff
@@ -20,17 +22,48 @@ generates the 'spec' file to configure the DPDK pipeline.
...
+
+parser EgressParser (packet_in buffer, out H parsed_hdr, inout M user_meta);
+
+control Egress (inout H hdr, inout M user_meta);
+
+control EgressDeparser (packet_out buffer, inout H hdr, in M meta);
+```
 
+- Size of all structure fields to be multiple of 8 bits and should not be greater than 64-bit. (This limitation is planned to be removed by grouping fields into fields which are multiple of 8-bits)
+- Combination of header and metadata as table key is not allowed. (This limitation is removed by P4C by copying the differing match key fields to metadata)
```
jafingerhut, line 66: "is removed" meaning "is planned to be removed", similar to the first restrction?

Both of these restrictions seem pretty limiting for what typical P4 targets support, so great to hear there is a plan to remove these restrictions.

hanw: Combining header and metadata as table key is already implemented by Usha.

hanw (approved):

backends/dpdk/README.md
jafingerhut: Already implemented before this change to the README and merged in?  Or planned to be merged by a future PR?

If it is already implemented before this PR, then why list it as a limitation in the README?

hanw: DPDK pipeline still has this limitation, but compiler was able to hide the limitation from p4 programmer by copying header field to metadata prior to table lookup.

jafingerhut: Understood.  Makes sense now.

jafingerhut (approved):
LGTM

rst0git (approved):

PR #2993, opened 2021-12-16
dpdk: change varbit tests for psa and pna arch

mvido:
PNA and PSA tests for varbit did not allow to distinguish the cases
when the header with varbit field was parsed and when it was not
and was just considered as payload.

Examples of packets and expected outputs are also in a new stf file.
Using this stf file, the test for PSA arch can be also run on bmv2
target.

When the tests are compiled for dpdk target, following warning is
displayed when a lookahead with header is used:
warning:  pkt.extract is ignored. Error: Reading field from invalid header
        pkt.extract(hdr.ipv4_option_timestamp, (bit<32>)(hdr.ipv4_option.len) * 8 - 16);

The origin of this warning is in ParsersUnroll pass during symbolic
evaluation of lookahead method.

When lookahead with header is used, the validation of programs also fails,
therefore in commit 4c125a6d9edc01ae4fb828de5e12dc17f5ccd74a, lookahead
with header is replaced by combination of lookahead and slicing.


testdata/p4_16_samples_outputs/pna-example-dpdk-varbit.p4.spec
```diff
@@ -119,19 +119,19 @@ apply {
 	lookahead m.MainParserT_parser_tmp_0
 	jmpeq MAINPARSERIMPL_PARSE_IPV4_OPTION_TIMESTAMP m.MainParserT_parser_tmp_0 0x44
 	jmp MAINPARSERIMPL_ACCEPT
-	MAINPARSERIMPL_PARSE_IPV4_OPTION_TIMESTAMP :	lookahead m.MainParserT_parser_tmp
-	lookahead m.MainParserT_parser_tmp_len_0
-	mov m.MainParserT_parser_tmp_2 m.MainParserT_parser_tmp_len_0
+	MAINPARSERIMPL_PARSE_IPV4_OPTION_TIMESTAMP :	lookahead m.MainParserT_parser_tmp16_0
+	mov m.MainParserT_parser_tmp_1 m.MainParserT_parser_tmp16_0
+	mov m.MainParserT_parser_tmp_2 m.MainParserT_parser_tmp_1
 	mov m.MainParserT_parser_tmp_3 m.MainParserT_parser_tmp_2
```
hanw, line 125: We should do something about these redundant moves.

hanw (approved):

PR #2994, opened 2021-12-17
FlattenInterfaceStruct: add type to a new member

VolodymyrPeschanenkoIntel:
This small change sets corresponded type for a new member. Otherwise, [this](https://github.com/p4lang/p4c/blob/17edf85a8421df2d151c79440c94a354d17ae195/testdata/p4_16_samples/issue983-bmv2.p4#L70) key will have metadata type, but not bit<16>.

### DIFF
--- a/midend/flattenInterfaceStructs.cpp
+++ b/midend/flattenInterfaceStructs.cpp
@@ -116,7 +116,7 @@ const IR::Node* ReplaceStructs::postorder(IR::Member* expression) {
         // replaced by the NestedStructs pass.)
         result = repl->explode(pe, prefix);
     } else {
-        result = new IR::Member(pe, newFieldName);
+        result = new IR::Member(expression->type, pe, newFieldName);
     }
     LOG3("Replacing " << expression << " with " << result);
     return result;
--- a/test/gtest/transforms.cpp
+++ b/test/gtest/transforms.cpp
@@ -63,7 +63,7 @@ class MidEndForFlatten : public PassManager {
             evaluator,
             new P4::MidEndLast()
         });
- 
+
         toplevel = evaluator->getToplevelBlock();
     }
     IR::ToplevelBlock* process(const IR::P4Program *&program) {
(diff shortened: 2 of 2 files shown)


mbudiu-vmw:
Types for expressions are optional and are not used in the type checking. Why is this necessary?

VolodymyrPeschanenkoIntel:
> Types for expressions are optional and are not used in the type checking. Why is this necessary?

Because after rewriting this key will have metadata type, but not bit<16>.  It not possible to get right type if you have key expression only.  I think that it could be considered as a bug because before rewriting this key has bit<16> type, but after it will have metadata type. I think that rewriting should save the type of the key expression if it just makes a new name. What do you think?

mbudiu-vmw:
If there is a bug it is not here. Please supply an example which fails. Where is the 16 bit coming from?

VolodymyrPeschanenkoIntel:
> 



> If there is a bug it is not here. Please supply an example which fails. Where is the 16 bit coming from?

bit<16> comes from [here](https://github.com/p4lang/p4c/blob/17edf85a8421df2d151c79440c94a354d17ae195/testdata/p4_16_samples/issue983-bmv2.p4#L35). In general It looks fine if you are not trying to use type of the key. Ok, I will try to add a new test for this situation.

VolodymyrPeschanenkoIntel:
@mbudiu-vmw , I added a gtest for this case. Without these changes this test failed.

mbudiu-vmw:
Of course the type is not known after you run flattenInterfaceStructs, you have to run another type checking pass if you need types every time you modify the IR representation.

mbudiu-vmw:
Moreover, please note that the 'type' field of an expression is not populated by default, only if you call type checking with an additional boolean flag.
In general, reproductions as gtest tests are not very easy to use, I really would prefer a stand-alone p4 program and a compiler invocation that fails. In your case, you make up a custom midend, and even midend passes depend on each other, you cannot assume that any combination of midend passes will produce a working compiler.

VolodymyrPeschanenkoIntel:
I didn't know about this magic flag in TypeChecking, thanks. It works with it.

PR #2997, opened 2021-12-17
dpdk: move definition of main metadata structure to the end

mvido:
Main metadata structure in the spec file may contain fields
which has structure type.
In such case the definition of those structures should be placed
before the definition of main metadata structure.
But currently the definitions of other structures are placed
after the definition of main metadata structure.
This commit moves the definition of main metadata structure
after the definitions of all other structures.

### DIFF
--- a/backends/dpdk/dpdkProgram.cpp
+++ b/backends/dpdk/dpdkProgram.cpp
@@ -188,11 +188,13 @@ const IR::DpdkAsmProgram *ConvertToDpdkProgram::create(IR::P4Program *prog) {
         headerType.push_back(ht);
     }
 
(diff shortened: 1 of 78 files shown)


mvido:
This PR is based on the changes from https://github.com/p4lang/p4c/pull/2996

mvido:
I'm not sure if the DPDK pipeline requires that a definition of the structure precedes its first use. If yes, then this change is needed, because otherwise the programs like a new example `testdata/p4_16_samples/pna-structure-lookahead.p4` generate invalid spec file. If not, then this change is optional.

mbudiu-vmw (approved):

mvido:
Rebased on top of current main with merged #2996.

PR #3000, opened 2021-12-21
DPDK Backend: Add support for SelectByDirection PNA extern

usha1830:
This PR adds support for SelectByDirection PNA extern.

SelectByDirection is implemented as if-else statement based on the direction.
 
`bit <T> var = SelectByDirection<T>(in PNA_Direction_t direction, in T n2h_value, in T h2n_value);`

is replaced by 

```
if (direction == PNA_Direction_t.NET_TO_HOST) {
     var = n2h_value;
 } else {
     var = return h2n_value;
 }
```

Some tests are added to cover statements where SelectByDirection can possibly be used.
1) as table key
2) as an operand of comparison
3) as action parameter

In all the cases, a temporary variable is inserted by midend passes and the call to SelectByDirection is assigned to this temporary. Backend needs to handle this Method call only as RHS of assignment.

### DIFF
--- a/backends/dpdk/dpdkHelpers.cpp
+++ b/backends/dpdk/dpdkHelpers.cpp
@@ -128,73 +128,6 @@ bool ConvertStatementToDpdk::preorder(const IR::AssignmentStatement *a) {
             BUG("%1% not implemented.", right);
         }
     } else if (auto m = right->to<IR::MethodCallExpression>()) {
-        /* PNA SelectByDirection extern is implemented as
-            SelectByDirection<T>(in PNA_Direction_t direction, in T n2h_value, in T h2n_value) {
-                if (direction == PNA_Direction_t.NET_TO_HOST) {
-                    return n2h_value;
-                } else {
-                    return h2n_value;
-                }
-            }
-            Example:
-                table ipv4_da_lpm {
-                    key = {
-                        SelectByDirection(istd.direction, hdr.ipv4.srcAddr, hdr.ipv4.dstAddr):
-                                                                      lpm @name ("ipv4_addr");
-                    }
-                    ....
-            }
-
-            In this example, KeySideEffect pass inserts a temporary for holding the result of
-            complex key expression and replaces the key expression with the temporary. An
-            assignment is inserted in the apply block before the table apply to assign the complex
-            expression into this temporary variable. After KeySideEffect pass, SelectByDirection
-            extern invocation is present as RHS of assignment and hence is evaluated here in this
-            visitor.
-
-            After keySideEffect pass:
-            bit <> key_0;
-            ...
-            table ipv4_da_lpm {
-                key = {
-                    key_0: lpm @name ("ipv4_addr");
-                }
-                ....
-            }
-
-            apply {
-                key_0 = SelectByDirection<bit<32>>(istd.direction, hdr.ipv4.srcAddr, hdr.ipv4.dstAddr);
-                ...
-            }
-
-            This is replaced by an assignment of the resultant value into a temporary.
-            if (istd.direction == PNA_Direction_t.NET_TO_HOST) {
-                key_0 = hdr.ipv4.srcAddr;
-            } else {
-                key_0 = hdr.ipv4.dstAddr;
-            }
-        */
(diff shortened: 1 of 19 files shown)


hanw (approved):

backends/dpdk/dpdkHelpers.cpp
```diff
@@ -128,6 +128,73 @@ bool ConvertStatementToDpdk::preorder(const IR::AssignmentStatement *a) {
...
+            }
+
+            This is replaced by an assignment of the resultant value into a temporary.
+            if (istd.direction == PNA_Direction_t.NET_TO_HOST) {
+                key_0 = hdr.ipv4.srcAddr;
+            } else {
+                key_0 = hdr.ipv4.dstAddr;
+            }
+        */
+        if (m->method->toString() == "SelectByDirection") {
```
hanw, line 177: Move this block further down
```
if (auto e =mi->to<P4::ExternMethod>()) {
   ...
} else if (auto e = mi->to<P4::ExternFunction>()) {
   // move to here.
}
```

hanw (changes requested):
See comments.

hanw (approved):

backends/dpdk/dpdkHelpers.cpp
```diff
@@ -184,6 +184,75 @@ bool ConvertStatementToDpdk::preorder(const IR::AssignmentStatement *a) {
...
+                    }
+                    ....
+                }
+
+                apply {
+                    key_0 = SelectByDirection<bit<32>>(istd.direction, hdr.ipv4.srcAddr, hdr.ipv4.dstAddr);
+                    ...
+                }
+
+                This is replaced by an assignment of the resultant value into a temporary.
```
mbudiu-vmw, line 227: The description is not 100% accurate - you don't actually generate these instructions in P4 IR, you generate directly assembly equivalent to this code.

backends/dpdk/dpdkHelpers.cpp
```diff
@@ -184,6 +184,75 @@ bool ConvertStatementToDpdk::preorder(const IR::AssignmentStatement *a) {
...
+                if (istd.direction == PNA_Direction_t.NET_TO_HOST) {
+                    key_0 = hdr.ipv4.srcAddr;
+                } else {
+                    key_0 = hdr.ipv4.dstAddr;
+                }
+            */
+            if (e->method->name == "SelectByDirection") {
+                auto args = e->expr->arguments;
+                if (args->size() != 3)
+                    ::error("Unexpected number of arguments\n");
```
mbudiu-vmw, line 237: Please use the error function that specifies an error type. Also, you should show something about the source code, otherwise no one will understand what the problem is when it occurs. This is probably a modelError.

testdata/p4_16_samples/pna-example-SelectByDirection.p4
```diff
@@ -0,0 +1,179 @@
...
+}
+
+control PreControlImpl(
+    in    headers_t  hdr,
+    inout main_metadata_t meta,
+    in    pna_pre_input_metadata_t  istd,
+    inout pna_pre_output_metadata_t ostd)
+{
+    apply {
+        // Note: This program does not demonstrate all of the code
```
mbudiu-vmw, line 85: smaller test examples are always easier to read, maintain and debug.

mbudiu-vmw (approved):

backends/dpdk/dpdkHelpers.cpp
usha1830: These error checks may not be required here. Frontend/typechecker is already checking if the number of arguments for extern invocation match with the declaration or not.

backends/dpdk/dpdkHelpers.cpp
usha1830: Description updated

testdata/p4_16_samples/pna-example-SelectByDirection.p4
usha1830: Simplified the tests. Removed unnecessary comments and declarations

usha1830:
@mbudiu-vmw, @hanw Thanks for the review! If there are no further comments, can this be merged?

PR #3003, opened 2021-12-24
Remove BUG_CHECK which is too conservative

### LINKED ISSUE #3001: Compiler bug attempting to call isValid() on headers returned from functions
I am filing this issue after trying out a test program related to this P4 spec issue: https://github.com/p4lang/p4-spec/issues/1000

Note: Whether this is legal P4 or not depends upon the answer to the questions raised against the language spec.

The primary reason for filing this issue is not to answer that question, but because the latest version of p4test gives a compiler bug when attempting to compile it, which ideally it should not even if the program is illegal according to the spec.

```
$ p4c --version
p4c 1.2.2 (SHA: 37cd30ee9 BUILD: DEBUG)

$ p4test p4-spec-1000-bmv2.p4
In file: /home/andy/p4c/frontends/p4/sideEffects.cpp:385
Compiler Bug: f();: method on left hand side?
```
[p4-spec-1000-bmv2.p4.txt](https://github.com/p4lang/p4c/files/7758318/p4-spec-1000-bmv2.p4.txt)

mbudiu-vmw:
Fixes #3001

### DIFF
--- a/frontends/p4/sideEffects.cpp
+++ b/frontends/p4/sideEffects.cpp
@@ -386,7 +386,8 @@ const IR::Expression* GetWrittenExpressions::everything = new IR::Constant(0);
 }  // namespace
 
 const IR::Node* DoSimplifyExpressions::preorder(IR::MethodCallExpression* mce) {
-    BUG_CHECK(!isWrite(), "%1%: method on left hand side?", mce);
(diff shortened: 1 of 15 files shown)


jafingerhut (approved):
I checked the test cases and the expected output files from the compiler for them, and they look correct to me.  I did not attempt to review the code changes to the compiler internals.

Issue #2957, opened 2021-11-20
SimplifyDefUse does not respect method calls in stack indices in parsers


fruffy:
This is a little specific. 

Given this program 

```p4 
extern bit<8> extern_call(inout H val);
parser p(packet_in pkt, out Headers hdr) {
    H tmp = {extern_call(hdr.h[max(3w1, 3w2)])};
    state start {
        transition parse_hdrs;
    }
    state parse_hdrs {
        pkt.extract(hdr.eth_hdr);
        pkt.extract(hdr.h.next);
        pkt.extract(hdr.h.next);
        transition accept;
    }
}
```

SimplifyDefUse transforms the following segment

```p4
extern bit<8> extern_call(inout H val);
parser p(packet_in pkt, out Headers hdr) {
    @name("tmp") H tmp_0;
    bit<8> tmp_1;
    bit<3> tmp_2;
    bit<3> tmp_3;
    H tmp_4;
    bit<8> tmp_5;
    state start {
        tmp_2 = max(3w1, 3w2);
        tmp_3 = tmp_2;
        tmp_4 = hdr.h[tmp_3];
        tmp_5 = extern_call(tmp_4);
        hdr.h[tmp_3] = tmp_4;
        tmp_1 = tmp_5;
        tmp_0 = (H){a = tmp_1};
        transition start_0;
    }
    state start_0 {
        pkt.extract<ethernet_t>(hdr.eth_hdr);
        pkt.extract<H>(hdr.h.next);
        pkt.extract<H>(hdr.h.next);
        transition accept;
    }
}
```
into 

```p4
extern bit<8> extern_call(inout H val);
parser p(packet_in pkt, out Headers hdr) {
    @name("tmp") H tmp_0;
    bit<8> tmp_1;
    bit<3> tmp_2;
    bit<3> tmp_3;
    H tmp_4;
    bit<8> tmp_5;
    state start {
        tmp_2 = max(3w1, 3w2);
        tmp_3 = tmp_2;
        tmp_4 = hdr.h[tmp_3];
        extern_call(tmp_4);
        ;
        ;
        ;
        transition start_0;
    }
    state start_0 {
        pkt.extract<ethernet_t>(hdr.eth_hdr);
        pkt.extract<H>(hdr.h.next);
        pkt.extract<H>(hdr.h.next);
        transition accept;
    }
}
```
Removing the assignments after the extern call discards any potential writes are made to `hdr.h[tmp_3]`

[...]

PR #3004, opened 2021-12-24
SimplifyDefUse does not respect method calls in stack indices in parsers. Bug fixed (#2957)

Jovana-Syrmia:
Bug fixed (#2957)

Bug occurs because  ComputeWriteSet pass, when visiting packet.extract(hdr.h.next), sets definition to whole stack that the packet.extract is the last to wrote there.

FindUninitialized pass after a BlockStatements saves ProgramPoints that last wrote to InOut/Out parameters in that block. (in issue2957.p4 after parser block, pass sees that the packet last wrote to stack and saves just that point, there is no information that it was partially written)

- Added fix when visiting MethodCallExpression. 
 Whenever parameter is a member with member name next, save ProgramPoints that wrote last to that location to hasUses  (because if it is a member with next there is now information which element of stack was written, so points that wrote before need to be saved)

- Added test for issue2957

- Changed output for invalid-hdr-warnings4 because there can be a case where assignments shouldn't be removed (like in issue2957) 
In invalid-hdr-warning4-frontend because of pkt.extract(hdr.h1.next) points that last changed hdr.h1 are saved. In this case lines 24 and 25 (hdr.h1[0].data = 32w1; hdr.h1[1].data = 32w1;)
This makes sense because with packet.extract we don't know which element was written, so we need to save both of these assignments.
And if we are keeping a point packet.extract(hdr.h1.next) there is a justification for keeping assignment statements  in line 24 and 25.

### DIFF
--- a/frontends/p4/simplifyDefUse.cpp
+++ b/frontends/p4/simplifyDefUse.cpp
@@ -1286,6 +1286,16 @@ class FindUninitialized : public Inspector {
                 if (isControlOrParserApply || mi->is<ExternMethod>() || mi->is<ExternFunction>()) {
                     if (typeMap->getType(expr->expression, true)->is<IR::Type_Header>()) {
                         headerDefs->update(expr->expression, TernaryBool::Yes);
+                        if (expr->expression->is<IR::Member>()) {
+                            auto member = expr->expression->to<IR::Member>();
+                            if (member->member == IR::Type_Stack::next ||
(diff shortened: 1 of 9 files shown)


mbudiu-vmw (changes requested):
I don't think this is really fixing the problem. The real fix is to mark writes of next or last as also reading the whole stack. With your fix these fields are always live.

PR #3005, opened 2021-12-24
Binary operation with bit and int types

vladyslav-dubina:
Helps to solve the error
```
[--Werror=invalid] error: ==: operands have different types: bit<8> and int<8>
        hdr.pool[hdr.ml.idx].val = hdr.vector[0].e;
                 ^^^^^^^^^^
```

`int `and `bit ` are the same and cast to `Type_Bits`. But `Type_Bits-> operator ==` will return `false ` in case of comparing `bit ` with `int`, due to the fact that they have different `baseName`

Example test: runtime-index-bmv2.p4
Add test for ParserUnroll Gtest?

### DIFF
--- a/frontends/common/constantFolding.cpp
+++ b/frontends/common/constantFolding.cpp
@@ -494,16 +494,10 @@ DoConstantFolding::binary(const IR::Operation_Binary* e,
     if (!lunk && !runk) {
         // both typed
         if (!ltb->operator==(*rtb)) {
-          if (ltb->baseName() == "bit" && rtb->baseName() == "int") {
-            if (ltb->size == rtb->size) {
-              resultType = ltb;
-            }
-          } else {
-            ::error(ErrorType::ERR_INVALID,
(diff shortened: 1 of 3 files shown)


mbudiu-vmw (changes requested):
The spec does not allow operations between different types. This will not be considered for review without a test

vladyslav-dubina:
My first decision was wrong, I found the reason for the different types, fixed them in a new commit

vladyslav-dubina:
Without these modifications, the ` int_type_in_slice.p4` test will generate an 
```
--Werror=invalid] error: ==: operands have different types: bit<4> and int<4>
        if (8s1[3:0] == 4s1) {
```
 error. Which is caused by the fact that in https://github.com/p4lang/p4c/blob/056eda9b3c67d3e2dbff7ff3deaadab831ca3097/frontends/common/constantFolding.cpp#L621 . `IsSignet `is not taken into account and it turns out that slice converts to `bit`, although `int `was expected

PR #3009, opened 2021-12-29
Possible problem with flatten structs

VolodymyrPeschanenkoIntel:
This draft contains simple test which is failed if user tries to get P4Table from the Type_Table. Field exp_etherType has wrong type in second case. It looks that new P4::TypeChecking(&refMap, &typeMap, true) doesn't help here. So, how to update P4Table inside Type_Table?

### DIFF
--- a/midend/flattenInterfaceStructs.cpp
+++ b/midend/flattenInterfaceStructs.cpp
@@ -116,7 +116,8 @@ const IR::Node* ReplaceStructs::postorder(IR::Member* expression) {
         // replaced by the NestedStructs pass.)
         result = repl->explode(pe, prefix);
     } else {
-        result = new IR::Member(expression->type, pe, newFieldName);
+        //result = new IR::Member(expression->type, pe, newFieldName);
+        result = new IR::Member(pe, newFieldName);
     }
     LOG3("Replacing " << expression << " with " << result);
     return result;
--- a/midend/flattenInterfaceStructs.cpp
+++ b/midend/flattenInterfaceStructs.cpp
@@ -116,7 +116,7 @@ const IR::Node* ReplaceStructs::postorder(IR::Member* expression) {
         // replaced by the NestedStructs pass.)
         result = repl->explode(pe, prefix);
     } else {
-        result = new IR::Member(pe, newFieldName);
+        result = new IR::Member(expression->type, pe, newFieldName);
     }
     LOG3("Replacing " << expression << " with " << result);
     return result;
--- a/test/gtest/transforms.cpp
+++ b/test/gtest/transforms.cpp
@@ -63,7 +63,7 @@ class MidEndForFlatten : public PassManager {
             evaluator,
             new P4::MidEndLast()
         });
- 
+
         toplevel = evaluator->getToplevelBlock();
     }
     IR::ToplevelBlock* process(const IR::P4Program *&program) {
@@ -77,7 +77,8 @@ const IR::Expression* getKeyExpression(const IR::P4Program* program, const char*
     CHECK_NULL(controlName);
     CHECK_NULL(tableName);
     CHECK_NULL(fieldName);
-    const auto* control = program->getDeclsByName(controlName)->toVector()->at(0)->to<IR::P4Control>();
+    const auto* control =
+        program->getDeclsByName(controlName)->toVector()->at(0)->to<IR::P4Control>();
     const auto* table = control->getDeclByName(tableName)->to<IR::P4Table>();
     for (auto key : table->getKey()->keyElements) {
         std::ostringstream output;
--- a/test/gtest/transforms.cpp
+++ b/test/gtest/transforms.cpp
@@ -60,6 +60,7 @@ class MidEndForFlatten : public PassManager {
 
         addPasses({
             new P4::FlattenInterfaceStructs(&refMap, &typeMap),
+            new P4::TypeChecking(&refMap, &typeMap, true),
             evaluator,
             new P4::MidEndLast()
         });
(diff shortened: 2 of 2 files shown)


fruffy:
So it looks like that TypeChecking infers for these headers and this table:

```p4
struct fwd_meta_t {
    bit<16> tmp;
    bit<32> x1;
    bit<16> x2;
    bit<32> x3;
    bit<32> x4;
    bit<16> exp_etherType;
    bit<32> exp_x1;
    bit<16> exp_x2;
    bit<32> exp_x3;
    bit<32> exp_x4;
}

struct metadata {
    bit<16> _fwd_meta_tmp0;
    bit<32> _fwd_meta_x11;
    bit<16> _fwd_meta_x22;
    bit<32> _fwd_meta_x33;
    bit<32> _fwd_meta_x44;
    bit<16> _fwd_meta_exp_etherType5;
    bit<32> _fwd_meta_exp_x16;
    bit<16> _fwd_meta_exp_x27;
    bit<32> _fwd_meta_exp_x38;
    bit<32> _fwd_meta_exp_x49;
}

@name("ingress.debug_table_cksum1") table debug_table_cksum1_0 {
    key = {
        hdr.ethernet.srcAddr              : exact @name("hdr.ethernet.srcAddr") ;
        hdr.ethernet.dstAddr              : exact @name("hdr.ethernet.dstAddr") ;
        hdr.ethernet.etherType            : exact @name("hdr.ethernet.etherType") ;
        user_meta._fwd_meta_exp_etherType5: exact @name("user_meta.fwd_meta.exp_etherType") ;
        user_meta._fwd_meta_tmp0          : exact @name("user_meta.fwd_meta.tmp") ;
        user_meta._fwd_meta_exp_x16       : exact @name("user_meta.fwd_meta.exp_x1") ;
        user_meta._fwd_meta_x11           : exact @name("user_meta.fwd_meta.x1") ;
        user_meta._fwd_meta_exp_x27       : exact @name("user_meta.fwd_meta.exp_x2") ;
        user_meta._fwd_meta_x22           : exact @name("user_meta.fwd_meta.x2") ;
        user_meta._fwd_meta_exp_x38       : exact @name("user_meta.fwd_meta.exp_x3") ;
        user_meta._fwd_meta_x33           : exact @name("user_meta.fwd_meta.x3") ;
        user_meta._fwd_meta_exp_x49       : exact @name("user_meta.fwd_meta.exp_x4") ;
        user_meta._fwd_meta_x44           : exact @name("user_meta.fwd_meta.x4") ;
    }
    actions = {
        NoAction_1();
    }
    default_action = NoAction_1();
}
```
That ` user_meta._fwd_meta_exp_etherType5` has type `struct metadata`, which is `Type_Struct`. It should be bit<16> instead. I do not quite understand how this is possible. It's either a problem in how the key is retrieved or in type checking.

VolodymyrPeschanenkoIntel:
@mbudiu-vmw, Do you have any ideas how to deal with that? Is it possible to pass the test with some modification of MidEnd?

mbudiu-vmw:
I didn't realize that you are waiting for me on this. In general I find that standalone tests are much easier to debug than gtest programs. But I assume that you have a specific combination of midend passes which requires a gtest to reproduce. Unfortunately I won't be able to look at this problem until next Tuesday, but I will put it first on my TODO list.

VolodymyrPeschanenkoIntel:
@mbudiu-vmw. This example demonstrates some problem connected with IR::Type_Table in IR::MethodCallExpression. Is it a problem of midend or not? If it is in midend then how we need to use midend in that case? If not then how to solve it?
In general, we just remove a source code connected with IR::Type_Table at all, but this problem makes impossible to use types from IR::Type_Table::table after midend if it makes there some changes.

mbudiu-vmw:
What do you want to achieve?
I suspect that there is a simpler way to do it.
In general IR nodes that are reachable through types are not necessarily part of the IR tree.
The `type` field of an expression points to a tree that may actually not be part of the program IR - it is a representation of the type that is separate from the IR tree (e.g., created when the type is canonicalized). So the tree does not necessarily reflect the latest version of the program. 
If you want to find the correct version of a table's key the safest way to do it is to actually locate the IR node that contains the table key declaration and use that IR node to query the typeMap.
Your code is traversing several trees expression->type->field->type->etc. The `field` of type that you are accessing is not part of the IR tree.
This is caused by a limitation of the ApplyTypesToExpressions pass: as soon as it modifies the `type` field of an expression the expression itself is changed and no longer part of the IR tree! The types of simple expressions that appear in the IR tree are always correct after that pass, but not necessarily the types of all nodes that are reachable from fields of the `type` fields of such expressions. Does this make sense?

mbudiu-vmw:
In your code this is problematic:
`auto key : type->table->getKey()->keyElements`
The table that you reach from the `type` field is not part of the current program IR tree, and may not have a correct `type` field ifself - and neither do any of its children nodes.

mbudiu-vmw:
See slide 81 in this document: https://github.com/p4lang/p4c/blob/main/docs/compiler-design.pdf
You can see how types are represented using IR trees that are not part of the program IR.

VolodymyrPeschanenkoIntel:
@mbudiu-vmw, Thanks for your answers. I think now it is more clear for me.  However, Does p4 have some criteria to use sometype->type instead of TypeMap::find?

mbudiu-vmw:
I think that the following hold:
- the type you get from the `type` field is always correct as a "canonical" type of the expression 
- however, this does not hold for any node reachable from the `type` field
- children of the `type` field node may not be part of the program IR, and thus may not modified by any passes

VolodymyrPeschanenkoLitSoft:
Thanks, How to detect which code is not a part of program IR in the general case?

mbudiu-vmw:
The only reliable way to detect whether something is part of the IR tree is to traverse the IR tree from the root and see whether the node is reached. Note that the IR tree itself is immutable, and nodes are removed from it all the time: every time you remove a node all the nodes all the way to the root are replaced. See slide 35 in the presentation. So nodes that are currently part of the IR tree may soon no longer be. (This is why we use getOriginal() frequently - the original node may no longer be in the tree by the time you look at the node...).

The situation is even more complicated if your backend backtraks and manipulates several trees simultaneously: then each IR node could be part of 0, 1, or multiple trees at the same time.

But let me move up a level and ask again: what do you want to do?

mbudiu-vmw:
BTW: see also slide 70. This is one of the most difficult things about using this IR representation.

PR #3011, opened 2021-12-30
 SimplifyDefUse does not respect method calls in stack indices in parsers #2957 

Jovana-Syrmia:
Bug #2957  
Discussion in PR #3004 

- For member expression added write for next/last as a read/write to the whole stack

- Added test for example issue2957.p4

- Changed test output for invalid-hdr-warnings4.p4
Besides writing pkt.extract(hdr.k1.next) is also reading a whole stack, so assignments hdr.h1[0].data = 32w1; hdr.h1[1].data = 32w1; (preceding) are saved.
They won't be deleted because they are preserved by pkt.extract(hdr.k1.next) and this ProgramPoint is always saved in hasUses, because it has a side effect.

### DIFF
--- a/frontends/p4/simplifyDefUse.cpp
+++ b/frontends/p4/simplifyDefUse.cpp
@@ -1335,6 +1335,7 @@ class FindUninitialized : public Inspector {
         if (basetype->is<IR::Type_Stack>()) {
             if (expression->member.name == IR::Type_Stack::next ||
                 expression->member.name == IR::Type_Stack::last) {
+                // Accessing these fields implies reading the whole stack
                 auto save  = lhs;
                 lhs = false;
                 visit(expression->expr);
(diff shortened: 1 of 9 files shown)


frontends/p4/simplifyDefUse.cpp
```diff
@@ -1335,6 +1335,11 @@ class FindUninitialized : public Inspector {
         if (basetype->is<IR::Type_Stack>()) {
             if (expression->member.name == IR::Type_Stack::next ||
                 expression->member.name == IR::Type_Stack::last) {
+                auto save  = lhs;
```
mbudiu-vmw, line 1338: please add a comment explaining that accessing these fields implies reading the whole stack since these behave like a read-modify-write operation on the LHS.

mbudiu-vmw (approved):

mbudiu-vmw (approved):

mbudiu-vmw:
Fixes #2957

PR #3012, opened 2022-01-03
Inline Alloc{1,2}D storage

vhavel:
This changes `Alloc{1,2}Dbase` to be a non-owning view into one- or two-dimensional array, and `AllocXD` to hold the actual data inline instead of owning it through `AllocXDbase` class (which stores elements in manually managed heap-allocated memory).

The change improves mainly performance of the class, it also simplifies the class a bit as a bonus.

### DIFF
--- a/lib/alloc.h
+++ b/lib/alloc.h
@@ -25,11 +25,9 @@ template<class T> class Alloc1Dbase {
     int         size;
     T           *data;
  protected:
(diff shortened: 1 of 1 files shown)


mbudiu-vmw (approved):
I am approving, but there are no uses of this class in the open-source part of the code, so perhaps you want an approval from the vendors who use it.

ChrisDodd (approved):

PR #3014, opened 2022-01-03
Disallow instantiations within a block

jafingerhut:
This is a proposed change to address issue #975 in the P4
specification: https://github.com/p4lang/p4-spec/issues/975

### DIFF
--- a/frontends/parsers/p4/p4parser.ypp
+++ b/frontends/parsers/p4/p4parser.ypp
@@ -1227,7 +1227,6 @@ statementOrDeclaration
     : variableDeclaration      { $$ = $1; }
     | constantDeclaration      { $$ = $1; }
     | statement                { $$ = $1; }
-    | instantiation            { $$ = $1; }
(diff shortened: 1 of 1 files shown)


jafingerhut:
I see only two failures when I run the p4c tests on my local machine, for these two test programs:

+ p4c/testdata/p4_16_errors/local_instance.p4
+ p4c/testdata/p4_16_errors/control-inline.p4

Note: It appears perfectly reasonable to delete at least one of these two test programs, since they are identical except for comments.

It appears that the only reason these test programs were added was because they caused an internal compiler error some time in 2016.  If the decision is made in the P4 specification to disallow such instantiations, then it seems that we should delete one of these test programs, and change the expected error message for the remaining one.

jafingerhut:
Deleting this PR, in favor of one that keeps the grammar transition but adds a comment why it is there: https://github.com/p4lang/p4c/pull/3016

PR #3015, opened 2022-01-03
dpdk: use header instead of metadata for lookahead in varbit tests

mvido:
DPDK pipeline library supports lookahead instruction only with header
not with metadata.
Local variable in parser is internally transformed to a main metadata
field which is currently not allowed as an operand for a lookahead
instruction in DPDK pipeline library.
Therefore we currently need to use a header which is part of main
headers structure.

When we use a header with lookahead in source program,
following warning is generated in ParsersUnroll pass in the midend:
```
warning:  packet.extract is ignored. Error: Reading field from invalid header
        packet.extract(hdr.ipv4_option_timestamp, (bit<32>)hdr.option.len * 8 - 16);
```

It is not clear to me whether header returned by lookahead method
should be valid or not and thus whether that warning is false or not.

If lookahead returns valid header then symbolic evaluation used
in ParsersUnroll pass needs to be fixed so that such warning is not
emitted.
If lookahead returns invalid header then I don't see how it could
be used so that we could use the return value.

Also we should handle somehow the case when lookahead is used with
local variable, because that is definitely valid scenario, but
at the moment the generated spec file can not be processed by DPDK
backend.

These tests are intended to test varbit feature, not lookahead, so
this is fixing only the tests so that they can be used to test
varbit with DPDK pipeline library.

Lookahead with local variables and symbolic evaluation of headers
in lookahead can be fixed separately.

### DIFF
--- a/testdata/p4_16_samples/pna-example-dpdk-varbit.p4
+++ b/testdata/p4_16_samples/pna-example-dpdk-varbit.p4
@@ -26,13 +26,11 @@ header ethernet_t {
 }
 
 header ipv4_base_t {
-    bit<4>  version;
-    bit<4>  ihl;
+    bit<8>  version_ihl;
     bit<8>  diffserv;
     bit<16> totalLen;
     bit<16> identification;
-    bit<3>  flags;
-    bit<13> fragOffset;
+    bit<16> flags_fragOffset;
     bit<8>  ttl;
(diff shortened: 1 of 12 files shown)


mvido:
After https://github.com/p4lang/p4c/pull/3080 was merged, this one is not needed anymore, so I'm closing it.

PR #3016, opened 2022-01-04
Add a comment on a not strictly necessary grammar rule

jafingerhut:
Also remove redundant test local_instance.p4, which was identical
except for comments to control-instance.p4


mbudiu-vmw (approved):

jnfoster:
I'm not sure the error message is more helpful. 

If we accept the proposed language change, and disallow instantiations in blocks, wouldn't it be better to do it in the grammar instead of the type system?

Also, are we certain this change is backward compatible? I'm not...

jafingerhut:
I have no strong opinion on the matter.  I created this PR following a suggestion from Mihai, but happy to change it to something else if someone has recommendations.

mbudiu-vmw:
I don't think we'll reject any program that used to be accepted.
We'll just reject it in a different way.

jnfoster:
Can you point me to the line in the spec where we say that instantiations within a block are not allowed? (It's probably there, my memory is just faulty.)

mbudiu-vmw:
There isn't any, but the implication of the evaluation section is that all instantiations are "static" - performed at compile-time.
So having them in a block statement is not really well-defined, since the block statements have a dynamic scope.
Or course, they could be treated like C statics, which are declared within blocks, and this restricts their visibility but not their lifetime, which is still static.

jafingerhut:
@jnfoster This PR was created in response to this _proposed_ change to the P4 spec that would explicitly prohibit this: https://github.com/p4lang/p4-spec/pull/1006

Adding the comment to the p4c grammar change should probably be made if and only if that PR or something similar is approved.

jnfoster:
It would be nice to signal the error directly in the Bison grammar...

jafingerhut:
I have no problems if someone can find a better change to make to p4c for this.  If it involves issuing error messages from the Bison grammar, I think I'm currently ill-prepared to write it.  All of the changes in this PR are under Apache 2.0 license as far as I am concerned, so feel free to reuse as much or as little of this PR as it helps someone else to write a different one, and we can close this if it becomes useless.

jafingerhut:
@mihaibudiu @jnfoster https://github.com/p4lang/p4-spec/pull/1006 was merged into P4 language spec in early 2022.

Is it thus reasonable to consider merging this PR?

mihaibudiu:
I am all for it

jnfoster (approved):

jnfoster:
Agree. I've gone ahead and merged this as it's a non-substantive change.

PR #3017, opened 2022-01-04
Experimental support for typeof keyword

mihaibudiu:
Here is a simple example:
```
const bit<32> b = 0;
typedef typeof(b) T;
const T c = 1;
```

Here `b` can be replaced by an arbitrary expression. The expression is never evaluated, so it only needs to typecheck.
`typeof` is an "expression" that returns a type. In this implementation it can only be used within a `typedef`.

### DIFF
--- a/frontends/p4/toP4/toP4.cpp
+++ b/frontends/p4/toP4/toP4.cpp
@@ -285,6 +285,14 @@ bool ToP4::preorder(const IR::Type_Typedef* t) {
     return false;
 }
 
+bool ToP4::preorder(const IR::Typeof* t) {
+    dump(2);
+    builder.append("typeof(");
+    visit(t->expression);
+    builder.append(")");
+    return false;
+}
+
 bool ToP4::preorder(const IR::Type_Newtype* t) {
     dump(2);
     visit(t->annotations);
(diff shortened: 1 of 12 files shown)


mbudiu-vmw:
Since this must be used in a typedef it is restricted to being used only at the toplevel.

mbudiu-vmw:
Discussed at the P4 language design working group in relation to https://github.com/p4lang/p4-spec/issues/978
This is a more general solution to that problem.

mbudiu-vmw:
Actually this is a problem since you cannot have toplevel variables or constants with stack types. So this design is not good enough.

vgurevich:
@mbudiu-vmw -- by the way, why did we decide to restrict typedefs (and other type definitions) to the top-level only? For example, I noticed that is it not super-convenient when I need to declare a struct used to instantiate a register -- usually it is needed only inside the corresponding control, but I have to put it at top-level as of today.

mbudiu-vmw:
We have restricted typedefs because I was afraid that local types may cause subtle semantics problems. Today all type declarations are global, so you are sure that all types are visible in inner contexts. This does not prohibit moving code that depends on these types around. It may be that local typedefs would work just fine, and it would be one solution around this problem. Is this a question that could be answered quickly by modelling it in Petra? Perhaps there are some subtleties that I don't envision.

mbudiu-vmw:
An alternative solution would be to change the grammar to allow `typeof(expression)` anywhere where a type is allowed. This will probably work, but will introduce a new keyword `typeof` which may break existing programs.

mbudiu-vmw:
This would be much more useful if merged after #3027

PR #3020, opened 2022-01-05
Restored optional priority for CONVERT_PRIMITIVE

MichalKekely:


frontends/p4/fromv1.0/converters.h
```diff
@@ -148,14 +148,14 @@ class PrimitiveConverter {
  * next highest will be run, etc.  The macro invocation is followed by the body of the
  * converter function.
  */
-#define CONVERT_PRIMITIVE(NAME) \
-    class PrimitiveConverter_##NAME : public PrimitiveConverter {               \
+#define CONVERT_PRIMITIVE(NAME, ...) \
```
mbudiu-vmw, line 151: Can you please document why this is needed so it won't be removed again in the future?

mbudiu-vmw (approved):

frontends/p4/fromv1.0/converters.h
MichalKekely: There is some explanation in the comment already:
https://github.com/p4lang/p4c/blob/550c96228383c85c20d70665efb09169474f62da/frontends/p4/fromv1.0/converters.h#L145-L149

Basically having the optional priority parameter there gives us the possibility to define multiple converters for the same primitive. For example one can be a general converter, that works across multiple platforms/cases (but is not very efficient) and then there can be specific converters for specific platforms/cases that can better leverage certain targer platform features. The more specific converters will have higher priority and will be invoked first - they try to do the conversion and if the fail/recognize that they are not applicable they can return `nullptr` and the next converter will be invoked.

For example, lets assume a made up case where we have a target platform that directly supports `bit_nor`. A general converter would convert `bit_nor(a,b)` to `not(a or b)` and then there would most likely be a midend/backend pass that just finds all of the subtrees of IR that represent the `not(x or y)` and convert them to something that represents the resource of the target platform that does the `bit_nor` directly. Instead we can simply write a specific converter (with higher priority) that checks if the target platform directly supports the `bit_nor` and if it does then convert it to the appropriate resource directly, if it doesn't it just returns `nullptr` and the general conversion happens anyway.

PR #3023, opened 2022-01-06
Error to constant in verify

vladyslav-dubina:
Create constant for `StackOutOfBounds ` error for `verify ` function in `parserUnroll`.

This change is necessary because the verify implementation expects constant instead of error. Using `convertErrors `did not produce any result causing errors
```
[--Werror=type-error] error: verify
  ---- Actual error:
  ./p4include/core.p4(23): Cannot unify type 'error' with type 'bit<32>'
  error {
  ^
  ---- Originating from:
  ./p4include/core.p4(23): Type of argument 'error.StackOutOfBounds' (error) does not match type of parameter 'toSignal' (bit<32>)
  error {
  ^
  ./p4include/core.p4(66)
  extern void verify(in bool check, in error toSignal);
                                             ^^^^^^^^
  ---- Originating from:
  ./p4include/core.p4(66): Function type 'verify' does not match invocation type '<Method call>'
  extern void verify(in bool check, in error toSignal);
              ^^^^^^
error: Testgen: Encountered errors during preprocessing. Exiting
```

We decided to initially set a constant value instead of error. To do this, we needed to get the error enum, but using `refMap-> getDeclaration ()` returned `nullptr`, and `typeMap-> getType` also returned `nullptr`

Because of this, we had to get error enum from P4Program


mbudiu-vmw:
This could only happen if you unroll after errors were converted into another type.
What is the pass ordering?

fruffy:
There is a custom pass similar to `ConvertEnums `that also converts errors. Should `parserUnroll` run before that or instead handle the case where `toSignal` has been converted to a bit type?

vladyslav-dubina:
@fruffy Are you suggesting to run `parserUnroll `after `ConvertEnums `?

PR #3024, opened 2022-01-07
When doing copyprop use original source position

### LINKED ISSUE #2678: LocalCopyPropagation loses source position information
When copy propagating an expression we should use the source position information of the destination expression. Otherwise errors are hard to read.

mbudiu-vmw:
Fixes #2678 
This has been cherry-picked from https://github.com/mbudiu-vmw/p4c-clone/commit/d22c2a2
This should only influence error messages give to users.


rst0git (approved):
LGTM

PR #3025, opened 2022-01-07
Support p4runtime info for serenum action arguments

### LINKED ISSUE #2904: Binding action args to serializable enums
It appears that compiler does not support binding of action args to simple enum types. This helps with readability.

Ask is to support accepting serializable enum types as action arguments.

```
enum bit<5> TableType {
  TT_ACL = 0,
  TT_FWD = 1
};

action A(TableType t) {
  <some_metadata> = t;
}

table X {
   key = {
      meta.m: ternary;
  }
  actions = {
    A;
  }
  const entries = {
    (0x0): A(TableType.TT_ACL);
    (0x1): A(TableType.TT_FWD);
    (_): A(TableType.TT_FWD);
  }
}
```

mbudiu-vmw:
Fixes #2904

### DIFF
--- a/control-plane/p4RuntimeSerializer.cpp
+++ b/control-plane/p4RuntimeSerializer.cpp
@@ -1539,11 +1539,16 @@ class P4RuntimeEntriesConverter {
             protoParam->set_param_id(parameterId++);
             auto parameter = actionDecl->parameters->parameters.at(parameterIndex++);
             int width = getTypeWidth(parameter->type, typeMap);
+            auto ei = EnumInstance::resolve(arg->expression, typeMap);
             if (arg->expression->is<IR::Constant>()) {
                 auto value = stringRepr(arg->expression->to<IR::Constant>(), width);
                 protoParam->set_value(*value);
             } else if (arg->expression->is<IR::BoolLiteral>()) {
                 auto value = stringRepr(arg->expression->to<IR::BoolLiteral>(), width);
+                protoParam->set_value(*value);
+            } else if (ei != nullptr && ei->is<SerEnumInstance>()) {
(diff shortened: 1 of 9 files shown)


control-plane/p4RuntimeSerializer.cpp
```diff
@@ -1539,12 +1539,17 @@ class P4RuntimeEntriesConverter {
...
             if (arg->expression->is<IR::Constant>()) {
                 auto value = stringRepr(arg->expression->to<IR::Constant>(), width);
                 protoParam->set_value(*value);
             } else if (arg->expression->is<IR::BoolLiteral>()) {
                 auto value = stringRepr(arg->expression->to<IR::BoolLiteral>(), width);
                 protoParam->set_value(*value);
+            } else if (ei != nullptr && ei->is<SerEnumInstance>()) {
+                auto sei = ei->to<SerEnumInstance>();
+                auto value = stringRepr(sei->value->to<IR::Constant>(), width);
+                protoParam->set_value(*value);
```
antoninbas, line 1552: I notice that in other places, we only call `EnumInstance::resolve` if the expression is actually an `IR::Member`, but I don't know if one approach is better than the other.

antoninbas (approved):
LGTM

control-plane/p4RuntimeSerializer.cpp
mbudiu-vmw: I think this is safe - if it's not member, it won't resolve...
Thanks.

jafingerhut (commented):
Already merged, but I also checked the P4Info and entries files generated by the compiler, and they look correct for the input program.  Thanks!

PR #3028, opened 2022-01-13
dpdk: use varbit field size for extract in bytes instead of bits

mvido:
DPDK target expects the size of varbit field for extract instruction
in bytes, but compiler was using the same value as in the P4 source
code which is a size in bits.
This commit adds the instructions to make a computation of the size
in bytes instead of bits.

### DIFF
--- a/backends/dpdk/README.md
+++ b/backends/dpdk/README.md
@@ -64,6 +64,12 @@ control EgressDeparser (packet_out buffer, inout H hdr, in M meta);
 
 - Size of all structure fields to be multiple of 8 bits and should not be greater than 64-bit. (This limitation is planned to be removed by grouping fields into fields which are multiple of 8-bits)
 - Combination of header and metadata as table key is not allowed. (This limitation is removed by P4C by copying the differing match key fields to metadata)
+- Extract instruction in DPDK target with 2 arguments expects the second argument to be the number
+of bytes to be extracted into the varbit field of the header.
+In P4 the second argument of the extract method is the number of bits.
+Compiler generates instructions which compute the size in bytes from the value in bits.
+If the value in bits is not aligned to 8 bits, the value is rounded down to the lower
+mutiple of 8 bits.
 
 ## Contacts
 
--- a/backends/dpdk/README.md
+++ b/backends/dpdk/README.md
@@ -68,8 +68,8 @@ control EgressDeparser (packet_out buffer, inout H hdr, in M meta);
 of bytes to be extracted into the varbit field of the header.
 In P4 the second argument of the extract method is the number of bits.
 Compiler generates instructions which compute the size in bytes from the value in bits.
-If the value in bits is not aligned to 8 bits, the value is rounded down to the lower
-mutiple of 8 bits.
+If the value in bits is not a multiple of 8, the value is rounded down to the lower
+multiple of 8 bits.
 
 ## Contacts
 
--- a/backends/dpdk/dpdkHelpers.h
+++ b/backends/dpdk/dpdkHelpers.h
@@ -138,7 +138,7 @@ class TypeWidthValidator : public Inspector {
         LOG3("Validating Type_Varbits: " << type);
         if (type->size % 8 != 0) {
             ::error(ErrorType::ERR_UNSUPPORTED,
-                    "%1% Maximum varbit width (%2%) not aligned to 8 bits",
+                    "%1% varbit width (%2%) not aligned to 8 bits",
                     type->srcInfo, type->size);
         }
     }
--- a/backends/dpdk/dpdkHelpers.h
+++ b/backends/dpdk/dpdkHelpers.h
@@ -133,18 +133,34 @@ class BranchingInstructionGeneration {
     bool generate(const IR::Expression *, cstring, cstring, bool);
 };
 
+class TypeWidthValidator : public Inspector {
+    void postorder(const IR::Type_Varbits *type) override {
+        LOG3("Validating Type_Varbits: " << type);
+        if (type->size % 8 != 0) {
+            ::error(ErrorType::ERR_UNSUPPORTED,
+                    "%1% Maximum varbit width (%2%) not aligned to 8 bits",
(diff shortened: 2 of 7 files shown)


jafingerhut:
I am curious what you plan to do if the extract length in bits is not a multiple of 8.  Cause a parsing error during packet parsing?

jafingerhut:
And whatever the answer to that question is, I would recommend adding that answer to a document describing properties/restrictions of the P4 DPDK back end, which other targets may or may not share.  Preferably some prominent document about the DPDK back end linked from its README.

As an example of the kinds of details you can learn about the BMv2 simple_switch back end from public documentation, in one place, see here: https://github.com/p4lang/behavioral-model/blob/main/docs/simple_switch.md.  I point people asking questions about the v1model architecture and v1model at that document frequently, and it would be good to have a similar document to read about such details for the DPDK back end, too.

mvido:
@jafingerhut 

> I am curious what you plan to do if the extract length in bits is not a multiple of 8. Cause a parsing error during packet parsing?

In such case the length will be rounded down to the nearest multiple of 8 (in bits). I have written it in the code comment here: https://github.com/p4lang/p4c/blob/4f87dd6f86faed7dc0363f385d8fe533f94c5eb6/backends/dpdk/dpdkHelpers.cpp#L635


> And whatever the answer to that question is, I would recommend adding that answer to a document describing properties/restrictions of the P4 DPDK back end, which other targets may or may not share. Preferably some prominent document about the DPDK back end linked from its README.
> 
> As an example of the kinds of details you can learn about the BMv2 simple_switch back end from public documentation, in one place, see here: https://github.com/p4lang/behavioral-model/blob/main/docs/simple_switch.md. I point people asking questions about the v1model architecture and v1model at that document frequently, and it would be good to have a similar document to read about such details for the DPDK back end, too.

As far as I know, the only documents which contain the information about DPDK target at the moment are:
https://doc.dpdk.org/guides/sample_app_ug/pipeline.html (this one is also referenced from the readme for dpdk backend in compiler)
https://doc.dpdk.org/guides/prog_guide/packet_framework.html#the-software-switch-swx-pipeline (this one is not referenced from the readme)

Maybe @cristian-dumitrescu could know more about the existing documentation.

I also see that there is a section about DPDK target limitations in https://github.com/p4lang/p4c/blob/main/backends/dpdk/README.md so I could add this information there too.

jafingerhut:
@mvido The document https://github.com/p4lang/p4c/blob/main/backends/dpdk/README.md that you linked looks like a perfect place to add a note about the details of DPDK-specific behavior like you describe.  I know that bmv2 back end does not round down extract's to a lower multiple of 8 bits, so this would be good to note in that DPDK README.

mvido:
> @mvido The document https://github.com/p4lang/p4c/blob/main/backends/dpdk/README.md that you linked looks like a perfect place to add a note about the details of DPDK-specific behavior like you describe. I know that bmv2 back end does not round down extract's to a lower multiple of 8 bits, so this would be good to note in that DPDK README.

@jafingerhut Sure, I updated it. Thank you for the feedback.

backends/dpdk/dpdkHelpers.h
```diff
@@ -133,18 +133,34 @@ class BranchingInstructionGeneration {
     bool generate(const IR::Expression *, cstring, cstring, bool);
 };
 
+class TypeWidthValidator : public Inspector {
+    void postorder(const IR::Type_Varbits *type) override {
+        LOG3("Validating Type_Varbits: " << type);
+        if (type->size % 8 != 0) {
+            ::error(ErrorType::ERR_UNSUPPORTED,
+                    "%1% Maximum varbit width (%2%) not aligned to 8 bits",
```
mbudiu-vmw, line 141: Why 'maximum' in the error message?

mbudiu-vmw (approved):

backends/dpdk/dpdkHelpers.h
mvido: I wanted to make it clear that this is checking the statically-declared maximum width of the varbit field which is stated in the definition of that field, f.e. if you have a header field defined as `varbit<304> data;`, this check only checks that the maximum width  `304` is aligned to 8 bits, not a dynamic width of the value which is parsed into that field.

The spec in section [7.1](https://p4.org/p4-spec/docs/P4-16-v1.2.2.html#sec-base-types) states:
> Bit-strings of dynamically-computed width with a fixed maximum width varbit<>

So I thought that it will be better to use also the word 'maximum'.

But I'm open to any other suggestion.

mbudiu-vmw: Just put yourself in the shoes of a user who gets this error. I am not sure they will know what the maximum is taken over.
I would just remove that word.

mvido: Ok, makes sense. Done.

backends/dpdk/README.md
```diff
@@ -64,6 +64,12 @@ control EgressDeparser (packet_out buffer, inout H hdr, in M meta);
 
 - Size of all structure fields to be multiple of 8 bits and should not be greater than 64-bit. (This limitation is planned to be removed by grouping fields into fields which are multiple of 8-bits)
 - Combination of header and metadata as table key is not allowed. (This limitation is removed by P4C by copying the differing match key fields to metadata)
+- Extract instruction in DPDK target with 2 arguments expects the second argument to be the number
+of bytes to be extracted into the varbit field of the header.
+In P4 the second argument of the extract method is the number of bits.
+Compiler generates instructions which compute the size in bytes from the value in bits.
+If the value in bits is not aligned to 8 bits, the value is rounded down to the lower
+mutiple of 8 bits.
```
jafingerhut, line 72: typo: mutiple -> multiple
I would suggest a slightly different wording: "If the value in bits is not a multiple of 8"

Thanks for adding the note to the docs!

mvido: Good catch, thanks! Fixed typo and reworded.

PR #3032, opened 2022-01-18
Update the p4c Docker image periodically

antoninbas:
Other p4lang Docker images are now updated periodically. We do the same
thing for p4c to pick up updates to the base images even when the p4c
repo itself is not updated.
We update the image every 4 hours. To avoid using using too much compute
in Github actions, we update the workflow definition to 1) use native
Docker actions and 2) leverage the Github actions cache. When the base
image (p4lang/behavioral-model) has not been updated and there are no
changes to the p4c code, the workflow will complete in less than a
minute.

### DIFF
--- a/.github/workflows/ci-container-image.yml
+++ b/.github/workflows/ci-container-image.yml
@@ -1,14 +1,17 @@
 name: Build and push latest image to container registry
 
-# This job should ideally also execute when the base image (p4lang/pi) is updated
 on:
   push:
     branches:
       - main
+  schedule:
+    # Ideally, we would update the image every time the base image (p4lang/behavioral-model)
+    # is updated, this is good enough.
+    - cron: '15 0-23/4 * * *' # "At minute 15 past every 4th hour from midnight through 23."
 
 jobs:
   build:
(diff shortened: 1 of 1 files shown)


.github/workflows/ci-container-image.yml
```diff
@@ -1,14 +1,17 @@
...
 
-# This job should ideally also execute when the base image (p4lang/pi) is updated
 on:
   push:
     branches:
       - main
+  schedule:
+    # Ideally, we would update the image every time the base image (p4lang/behavioral-model)
+    # is updated, this is good enough.
+    - cron: '15 0-23/4 * * *' # "At minute 15 past every 4th hour from midnight through 23."
```
mbudiu-vmw, line 10: even every midnight would be good from my pov

mbudiu-vmw (approved):
I am approving, although I can't claim I understand all the nuances of these commands.

.github/workflows/ci-container-image.yml
antoninbas: this is staggered compared to the other images so that the jobs don't run at the same time and so that all images can be updated in the same "cycle"

PR #3033, opened 2022-01-19
Disallow programs that call tables in the 'actions' table property

### LINKED ISSUE #2835: Compiler bug for table action parameter expression containing table_name.apply().hit
This test program is very similar to the one for https://github.com/p4lang/p4c/issues/2321 except that it has a table action parameter with table_name.apply().hit in an expression for an action parameter, instead of a function call.

This is also related to https://github.com/p4lang/p4-spec/issues/840

I am pretty sure there is another p4-spec or p4c issue where Nate Foster and I had some discussion in the comments of P4-16 programs like the test program for this issue, where I was arguing that this seems so complex to handle in general, that it seems to me to be perfectly reasonable to either disallow it in the language spec, or at least call it out as a case where support for it is not required.  I will link to that issue here if I can find it.

The primary reason I am filing an issue is because of the 'Compiler Bug' in the output.  Personally I would consider it a fix if there were simply an error message that said something like "table apply() calls are not supported in parameters to actions within a table's `actions` property", but if someone thinks this should be fully supported somehow then I can write even trickier test programs to make the compiler output correct, e.g. programs where such tables with these kinds of action parameters can nest 3 or 4 levels deep.

```
$ p4c --version
p4c 1.2.0+g202103291035~a69e52 (SHA: 9de755b7e BUILD: DEBUG)

$ p4c --target bmv2 --arch v1model p4c-issue-2835-bmv2.p4 
In file: /home/andy/p4c/frontends/p4/def_use.h:487
Compiler Bug: p4c-issue-2835-bmv2.p4(86): No location set known for t_two_0/t_two.apply().hit;
            a_two(t_two.apply().hit ? bump_val(hdr.ethernet.etherType) : hdr.ethernet.etherType, bump_val(hdr.ethernet.etherType));
                  ^^^^^^^^^^^^^^^^^
```

[p4c-issue-2835-bmv2.p4.txt](https://github.com/p4lang/p4c/files/6755803/p4c-issue-2835-bmv2.p4.txt)

mbudiu-vmw:
Fixes #2835

### DIFF
--- a/frontends/p4/sideEffects.h
+++ b/frontends/p4/sideEffects.h
@@ -344,6 +344,31 @@ class TablesInKeys : public Inspector {
     }
 };
 
+/// Discovers action arguments that may involve table applications.
+/// These idioms are currently not supported:
+/// table s { ... }
+/// table t {
+///    actions { a(s.apply().hit ? ... ); }
+class TablesInActions : public Inspector {
+    ReferenceMap* refMap;
+    TypeMap*      typeMap;
+
+ public:
+    TablesInActions(ReferenceMap* refMap, TypeMap* typeMap)
+            : refMap(refMap), typeMap(typeMap) {}
+    void postorder(const IR::MethodCallExpression* expression) override {
+        if (findContext<IR::ActionList>()) {
+            HasTableApply hta(refMap, typeMap);
+            hta.setCalledBy(this);
+            (void)expression->apply(hta);
+            if (hta.table != nullptr) {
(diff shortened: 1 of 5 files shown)


jafingerhut (approved):
As usual, I have not tried to review the p4c code changes in any kind of detail.  The test program and the error messages that the compiler outputs with these changes look quite clear to me.

rst0git (approved):
LGTM

Issue #2905, opened 2021-09-21
Allow tables with an empty entries list.


mbudiu-vmw:

jnfoster:
Can't wait for the breathless unveiling of P4's new features:`bit<0>` values and tables without entries :-)

vgurevich:
@jnfoster -- given that the compiler does accept an empty P4 program, I think we are on the right path :)
```
$ touch /tmp/null.p4
$ ./p4test /tmp/null.p4
[--Wwarn=missing] warning: Program does not contain a `main' module
```

Seriously, though, the empty constructs are very useful and add a degree of completeness and orthogonality to the language and we should be on the lookout for such misses, although they are often not of the highest priority.

jnfoster:
Totally :-)

PR #3035, opened 2022-01-19
Allow empty entries lists for tables

mbudiu-vmw:

### DIFF
--- a/frontends/parsers/p4/p4parser.ypp
+++ b/frontends/parsers/p4/p4parser.ypp
@@ -1301,7 +1301,7 @@ entry
     ;
 
 entriesList
-    : entry                          { $$ = new IR::Vector<IR::Entry>(); $$->push_back($1); }
+    : %empty                         { $$ = new IR::Vector<IR::Entry>(); }
     | entriesList entry              { $$ = $1; $$->push_back($2); }
     ;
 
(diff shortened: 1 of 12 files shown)


mbudiu-vmw:
Fixes #2905

jfingerh:
It would be nice to have a test case that generates a P4Info file and a P4 entries file, and verifies the entries file contains no entries for such a table.  It probably works, but may as well exercise it with an automated test.  I can create such a test case in a separate PR after this one is merged, if you prefer.

jfingerh (approved):
LGTM

mbudiu-vmw:
If you have a handy program with an empty table we can attach it to this PR.

jfingerh:
I can create one tonight if that is soon enough to include it in this PR.

jafingerhut:
Attached is a zip file that if you unzip it in the root directory of your p4c repo copy, while checked out on your branch, will add one test program to p4_16_samples and its expected outputs in p4_16_samples_outputs.  The test is the same as the one called table-entries-exact-bmv2.p4 except that its `const entries` list is empty.  The output file `table-entries-exact-0entries-bmv2.p4.entries.txt` is empty, as it should be.

mbudiu-vmw:
@jfingerh added the test you supplied. Please approve if we can merge this.

jafingerhut (approved):
LGTM

Issue #2755, opened 2021-04-30
Incorrect P4Info generation when different actions have same name annotation


jafingerhut:
Attached zip file issue2755.zip has two similar but slightly different P4_16 source files, and the P4Info files I obtained by running with a recent (2021-Apr-30) version of p4c source code.  Both P4Info files appear to be wrong in a similar way.

In my opinion, the correct behavior here would be to give a compile time error because two actions in the source code have the same `@name` annotation.

[issue2755.zip](https://github.com/p4lang/p4c/files/6408320/issue2755.zip)

jafingerhut:
I have collected most of the P4 programs given on issues that I have added the label control_plane_api to (currently 4 of these open as of 2022-Jan-19), and added them to the `name-annotations` directory of my p4-guide repo here: https://github.com/jafingerhut/p4-guide/tree/master/name-annotations

I have also written a description of what each of them do with the latest version of p4c as of 2022-Jan-19 (with the particular commit SHA I tested with), and my opinion of which of them have correct behavior from p4c, and which seem to be bugs.

This is a link to what I believe is one possible way to fix the root cause of these issues. I am not saying it is the only way, but I think it goes to the heart of the issue, i.e. it isn't any kind of hack or workaround: https://github.com/jafingerhut/p4-guide/tree/master/name-annotations#one-potential-way-to-fix-the-root-cause-of-some-of-these-issues

jafingerhut:
Copying a note from Mihai Budiu to this open issue, so it remains easily found while this issue is still open: "There is one more wrinkle to consider: for P4-14 programs the compiler will actually create a copy of each action for each context where it is used (e.g., each table), and it will attach a @name annotation to it. All the copies will have the same @name."

jfingerh:
I discussed this issue with Vladimir Gurevich on 2022-Mar-16 for his opinion on this troublesome P4_16 program in particular: https://github.com/jafingerhut/p4-guide/blob/master/name-annotations/actions-5-same-name-annot.p4#L7-L15 with the same @name annotation on two functionally different actions, and he agreed with me that except for those @name annotations, this looks like a perfectly reasonable program, but with the @name annotations there appears to be nothing that a P4 API generation pass could do but give an error for it (which the open source p4c's P4Info generation does _not_ as of 2022-Mar-16, but instead writes an incorrect P4Info file).

[...]

Issue #2749, opened 2021-04-26
Name clash in Localize All Actions


jnfoster:
The following program compiles incorrectly:
```p4
#include <core.p4>
#include <v1model.p4>

struct headers { }

struct metadata { }

parser MyParser(packet_in packet, out headers hdr, inout metadata meta, inout standard_metadata_t standard_metadata) {
    state start {
        transition accept;
    }
}

control MyVerifyChecksum(inout headers hdr, inout metadata meta) {
    apply { }
}

control MyIngress(inout headers hdr, inout metadata meta, inout standard_metadata_t standard_metadata) {
    action a(bit<9> port) {
        standard_metadata.egress_spec = port;
    }
    table t {
        actions = {
            a;
            NoAction;
        }
        key = { }
        default_action = NoAction();
    }
    apply {
      a(0);
      t.apply();
    }
}

control MyEgress(inout headers hdr, inout metadata meta, inout standard_metadata_t standard_metadata) {
    apply { }
}

control MyComputeChecksum(inout headers hdr, inout metadata meta) {
    apply { }
}

control MyDeparser(packet_out packet, in headers hdr) {
    apply { }
}

V1Switch(MyParser(),
	 MyVerifyChecksum(),
	 MyIngress(),
	 MyEgress(),
	 MyComputeChecksum(),
	 MyDeparser()) main;
```
Note that action `a` is invoked via table `t` and directly in `MyIngress`.

However, the `LocalizeAllActions` and `RemoveActionParameters` passes generate this program:
```p4
#include <core.p4>
#include <v1model.p4>

struct headers {
}

struct metadata {
}

parser MyParser(packet_in packet, out headers hdr, inout metadata meta, inout standard_metadata_t standard_metadata) {
    state start {
        transition accept;
    }
}

control MyVerifyChecksum(inout headers hdr, inout metadata meta) {
    apply {
    }
}

control MyIngress(inout headers hdr, inout metadata meta, inout standard_metadata_t standard_metadata) {
    @name(".NoAction") action NoAction_0() {
    }
    @name("MyIngress.a") action a(bit<9> port) {
        standard_metadata.egress_spec = port;
    }
    bit<9> port_0;
    @name("MyIngress.a") action a_2() {
        port_0 = 9w0;
        standard_metadata.egress_spec = port_0;
    }
    @name("MyIngress.t") table t_0 {
        actions = {
            a();
            NoAction_0();
        }
        key = {
        }
        default_action = NoAction_0();
    }
[...]

PR #3037, opened 2022-01-19
Check name annotations for duplicates in original program

### LINKED ISSUE #1949: Compiler silently allows duplicate name annotations
I will add another comment with example program soon.

mbudiu-vmw:
We give a warning if there are two objects in the program that have the same `@name` annotation.
It's not clear that this fix is sufficient - some names are generated internally by the compiler.
Moreover, it is not clear that the control-plane does not tolerate multiple objects with the same name - e.g., if they have different types.
Fixes #1949

### DIFF
--- a/backends/bmv2/CMakeLists.txt
+++ b/backends/bmv2/CMakeLists.txt
@@ -184,6 +184,8 @@ set (XFAIL_TESTS
   testdata/p4_16_samples/issue2664-bmv2.p4
   # This one uses arguments of incorrect types to log_msg
   testdata/p4_16_samples/issue2201-bmv2.p4
+  # Two objects with the same @name annotation
+  testdata/p4_16_samples/issue1949-bmv2.p4
 )
 
 set (BMV2_PARSER_INLINE_TESTS "${P4C_SOURCE_DIR}/testdata/p4_16_samples/parser-inline/*.p4")
--- a/frontends/CMakeLists.txt
+++ b/frontends/CMakeLists.txt
@@ -18,6 +18,7 @@ set (P4_FRONTEND_SRCS
   p4/checkConstants.cpp
   p4/checkCoreMethods.cpp
   p4/checkNamedArgs.cpp
+  p4/checkNameAnnotations.cpp
   p4/createBuiltins.cpp
   p4/def_use.cpp
   p4/defaultArguments.cpp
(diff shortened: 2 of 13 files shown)


mbudiu-vmw:
This causes a lot of failures in tests. There are many programs where there are different entities with the same `@name` annotation that cause no problems. So the question is "when is it wrong to have the same name on two different objects"? For this we need some rules for when objects names may be ambiguous to the control-plane.

jafingerhut:
I suspect that this comment and the test programs it links to, and the discussion above it, give some clues as to what I consider would be a correct fix.  I don't know if those ideas actually are correct, but I trust that I was on to something with the test case mentioned there :-).  https://github.com/p4lang/p4c/issues/2749#issuecomment-830432803

jafingerhut:
If we want to get to the bottom of #1949 and #2749, I suspect it would take a dedicated task force of Mihai, Antonin, myself, and several people from Intel and/or other companies (especially those who are experts in Intel's TDI control plane API) with P4 compilers to spend N hours on the problem until we had all of the following:

(a) examples of programs that demonstrate the corner cases involving duplicate action names and `@name` and/or `@id` annotations on actions here, and an agreement on which of them are legal, and which are not, and why.
(b) What P4Info file each should generate, if we agree it is a legal P4 program.
(c) if the compiler is allowed to duplicate an action used by multiple tables for optimization purposes, whether it makes sense to allow the compiler to do that before control plane API generation or not (my guess: no, it makes no sense to have a compiler do that)

I can collect some interesting corner case example P4 program I have already written and attached to existing p4c issues for (a), especially #2755, but I can't by myself generate agreement on which of them are legal, and which are not.  I can only provide my own opinion on that question.

jafingerhut:
I have collected most of the P4 programs given on issues that I have added the label control_plane_api to, and added them to the `name-annotations` directory of my p4-guide repo here: https://github.com/jafingerhut/p4-guide/tree/master/name-annotations

I have also written a description of what each of them do with the latest version of p4c as of 2022-Jan-19 (with the particular commit SHA I tested with), and my opinion of which of them have correct behavior from p4c, and which seem to be bugs.

This is a link to what I believe is one possible way to fix the root cause of these issues.  I am not saying it is the only way, but I think it goes to the heart of the issue, i.e. it isn't any kind of hack or workaround: https://github.com/jafingerhut/p4-guide/tree/master/name-annotations#one-potential-way-to-fix-the-root-cause-of-some-of-these-issues

mbudiu-vmw:
There is one more wrinkle to consider: for P4-14 programs the compiler will actually create a copy of each action for each context where it is used (e.g., each table), and it will attach a `@name` annotation to it. All the copies will have the same `@name`.

jafingerhut:
Thanks for mentioning the wrinkle.  I've added a copy of it to open issue #2755 for the same reason as I moved my comment: keeping it easy to find while that issue remains open.

PR #3042, opened 2022-01-20
Handle packet_in.lookahead in symbolic interpreter

### LINKED ISSUE #3041: Bug when lookahead is used with header type
When lookahead is used with a header type as in following example:
```
header option_t {
    bit<8> type;
    bit<8> len;
}

struct headers_t {
    ethernet_t ethernet;
    ipv4_base_t             ipv4_base;
    ipv4_option_timestamp_t ipv4_option_timestamp;
    option_t                option;
}

...

    state parse_ipv4_options {
        hdr.option = pkt.lookahead<option_t>();
        transition select(hdr.option.type) {
            8w0x44 &&& 8w0xff: parse_ipv4_option_timestamp;
            default : accept;
        }
    }
    state parse_ipv4_option_timestamp {
        pkt.extract(hdr.ipv4_option_timestamp, (bit<32>)hdr.option.len * 8 - 16);
        transition accept;
    }
```
the whole sample program [pna-example-dpdk-varbit.p4.txt](https://github.com/p4lang/p4c/files/7905845/pna-example-dpdk-varbit.p4.txt) is a modification of https://github.com/p4lang/p4c/blob/main/testdata/p4_16_samples/pna-example-dpdk-varbit.p4 (this modification is also available in draft PR: https://github.com/p4lang/p4c/pull/3015) and we compile it with p4c-dpdk as follows:
`./build/p4c-dpdk ./testdata/p4_16_samples/pna-example-dpdk-varbit.p4 -o out.spec
`
we get following warning produced inside `ParsersUnroll` pass:
```
./testdata/p4_16_samples/pna-example-dpdk-varbit.p4(85): warning:  pkt.extract is ignored. Error: Reading field from invalid header
        pkt.extract(hdr.ipv4_option_timestamp, (bit<32>)hdr.option.len * 8 - 16);
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

P4_16 spec does not explicitly state (or I didn't find it) whether lookahead returns a valid or invalid header, but it does not make sense to me if the returned header should be invalid.
If we assume that the header returned by lookahead should be valid, then there is probably a bug in `ParsersUnroll` pass which causes that warning.

There is also probably another problem in `ExpandLookahead` pass too because validator used in validate-p4c CI pipeline reports some error in that pass used in p4test, the details can be seen in the output of CI run in draft PR https://github.com/p4lang/p4c/pull/3015 : https://github.com/p4lang/p4c/runs/4695836576?check_suite_focus=true

mbudiu-vmw:
Fixes #3041


mbudiu-vmw:
@VolodymyrPeschanenko @VolodymyrPeschanenkoIntel please review.
I think we need at least one more review, though.

mvido:
I have observed thig bug with dpdk backend and I'm not sure if it is manifesting with any other backend too.
The test you added is not run with dpdk backend, following tests are run with dpdk backend:
```
set (P4_16_SUITES
  "${P4C_SOURCE_DIR}/testdata/p4_16_samples/pna-*.p4")
```
so I suggest to rename the test, or add it to the list of tests run with dpdk backend.

mbudiu-vmw:
I have renamed the file to "pna-issue3041.p4".

jafingerhut (approved):
The test programs and expected outputs look reasonable to me, under the assumption that lookahead on a header type should make the header become valid.  As usual, I'm not the right person to comment on the compiler code changes.

mvido:
@mbudiu-vmw FYI validate job seems to be broken (@fruffy is probably working on a fix in https://github.com/p4lang/p4c/pull/3043) and because of that the Validate step was not run and the pipeline passed. But this new program pna-issue3041.p4 would otherwise fail when run with Gauntlet. When the validate job is fixed, it will be probably failing on this pna-issue3041.p4 program until the issue with setValid in Gauntlet is fixed.

PR #3043, opened 2022-01-21
Migrate validation tests to CMake

fruffy:
This pull request migrates the P4C validation tests that use Gauntlet away from pytest to CMake. This change also removes a couple unnecessary dependencies and simplifies the build process. Gauntlet is now a simple P4C back end and does not require any additional dependencies other than `boost-filesystem`.

### DIFF
--- a/.github/workflows/ci-validation.yml
+++ b/.github/workflows/ci-validation.yml
@@ -30,7 +30,7 @@ jobs:
     - name: Validate
       if: env.BUILD_SUCCESS == 'true'
       run: |
-          docker run -w /gauntlet p4c python3 -m pytest test.py -vrf -k "test_p4c" -n $CTEST_PARALLEL_LEVEL --suppress-crashes
+          docker run --privileged -w /p4c/build -e $CTEST_PARALLEL_LEVEL p4c ctest  -R toz3-validate-p4c --output-on-failure --schedule-random
 
     - name: Build Failed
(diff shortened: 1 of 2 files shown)


fruffy:
This is tricky to fix. deadsnakes has dropped support for the installation of Python 3.6 on Ubuntu 16.04. https://github.com/deadsnakes/issues/issues/195  
Python3.6 is required to work for the Gauntlet tests because of some of the pytest features that are being used. Getting pytest to function with Ubuntu 16.04 and Python 3.5 seems difficult. It might take me a little longer to fix this. I might switch Gauntlet to cmake.

jfingerh:
Would it perhaps become easier to fix if we updated to Ubuntu 18.04 or Ubuntu 20.04 for the CI tests?  I don't know how much work remains to achieve that, but I think Antonin Bas has been making some changes recently that might have gotten us closer?

fruffy:
Yes, there are few (or even no) changes required with Ubuntu 18.04 or 20.04.

antoninbas:
I would suggest not putting too much effort into this, since I am currently working in updating the base image to Ubuntu 20.04

fruffy:
Awesome, thanks for doing this. This will get rid of so many build hacks.
Anyway, I am using this pull request as an opportunity to migrate to CMake, too. Removes a few unnecessary dependencies.

mbudiu-vmw (approved):

PR #3047, opened 2022-01-22
Unification for generic types must compare type arguments too

### LINKED ISSUE #3045: Typechecking bugs
This code should not type check (i.e., with `p4test`) but does:
```
#include <core.p4> 

extern register<T> {
  register(bit<32> size);
  void write(in bit<32> index, in T value);
}

void f<T>(register<T> r, in T t) {
  r.write(0, t);
}

control c()() {
    register<bool>(1) r;

    apply {
        f(r, exact);
    }
}
```
This code also should not type check but does -- note the type parameter on `c` now.
```
#include <core.p4> 

extern register<T> {
  register(bit<32> size);
  void write(in bit<32> index, in T value);
}

void f<T>(register<T> r, in T t) {
  r.write(0, t);
}

control c<T>()() {
    register<T>(1) r;

    apply {
        f(r, exact);
    }
}
```

mbudiu-vmw:
Fixes #3045

### DIFF
--- a/backends/ebpf/targets/xdp_target.py
+++ b/backends/ebpf/targets/xdp_target.py
@@ -1,1 +1,0 @@
-/home/mbudiu/git/p4c/extensions/p4c-xdp/xdp_target.py
--- a/backends/ebpf/targets/xdp_target.py
+++ b/backends/ebpf/targets/xdp_target.py
@@ -1,0 +1,1 @@
+/home/mbudiu/git/p4c/extensions/p4c-xdp/xdp_target.py
--- a/ir/type.def
+++ b/ir/type.def
@@ -493,16 +493,7 @@ class Type_SpecializedCanonical : Type {
         BUG_CHECK(substituted->is<IMayBeGenericType>(), "substituted %1% is not generic", substituted);
     }
     const Type* getP4Type() const override;
-    toString{
-        cstring result = baseType->toString() + "<";
-        cstring sep = "";
-        for (auto t : *arguments) {
-            result += sep + t->toString();
-            sep = ", ";
-        }
-        result += ">";
-        return result;
-    }
+    toString{ return baseType->toString() + "<...>"; }
 }
 
 /// A declaration that consists of just an identifier, e.g., an enum member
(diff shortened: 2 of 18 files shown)


jnfoster (approved):
This fix looks good to me. Thanks!

backends/ebpf/targets/xdp_target.py
```diff
@@ -0,0 +1 @@
+/home/mbudiu/git/p4c/extensions/p4c-xdp/xdp_target.py
```
rst0git, line 1: This file appears to have been added unintentionally

mbudiu-vmw: yes, I keep adding it.

ir/type.def
```diff
@@ -493,16 +493,7 @@ class Type_SpecializedCanonical : Type {
...
-        cstring result = baseType->toString() + "<";
-        cstring sep = "";
-        for (auto t : *arguments) {
-            result += sep + t->toString();
-            sep = ", ";
-        }
-        result += ">";
-        return result;
-    }
+    toString{ return baseType->toString() + "<...>"; }
```
rst0git, line 497: This change appears revert https://github.com/p4lang/p4c/commit/8eebedea818a2c788e0675d052a170be904a8e7f#diff-ea1da9796c6c2b602731a169985ebd06ed51a4b37dc65b5f5c8e67cba889c974R496

However, if we compare the output, the additional information about the arguments looks useful.

```diff
-issue3045-1.p4(16): Type of argument 'r' (register<...>) does not match type of parameter 'r' (register<...>)
+issue3045-1.p4(14): Type of argument 'r' (register<bool>) does not match type of parameter 'r' (register<T>)
```

mbudiu-vmw: It looks useful, but it is sometimes wrong. When the compiler substitutes types this will print the substituted types instead of the original ones the programmer wrote. That's why I decided it's better this way. We had a similar issue with variable names, but that is easier to solve, since variables remember their original names, while other IR constructs don't.

mbudiu-vmw: "wrong" meaning that it displays to the user types they never wrote in their program

rst0git (approved):

PR #3049, opened 2022-01-26
dpdk: improvement of extract and lookahead support

mvido:
This PR improves and extends the support of extract and lookahead methods in backend for DPDK target.
Until now only a limited set of extract and lookahead usages were supported and unsupported usages in source P4 code produced invalid spec file and an author of the P4 code was not even warned during compilation.

Following changes are included:

* Injecting local variables of header types into header struct instead of metadata struct

Instructions extract and lookahead in DPDK target assembler support only operand which is a header, metadata fields are not supported as operands.
Until now, all local variables were injected into main metadata structure.
As a result, if there was an extract or lookahead call which parsed the data from a packet into a header, there was generated an instruction with metadata operand producing invalid spec file.
As of now every local variable which is of header type is instantiated as a header in resulting spec file instead of adding it to main metadata structure.
This fixes the cases when a lookahead or extract parse a header into a local variable, f.e. as follows:
```
state parse_header {
    my_header_t local_hdr;
    pkt.extract(local_hdr);
}
```

* Lookahead of base or structure type is transformed into a lookahead of header type

Lookahead implementation in DPDK target supports only header instance as an operand.
Variables or metadata structure fields of other than header types need to be either rejected or transformed so that there is used lookahead with header instance operand.
This commit implements a transformation which converts all usages of lookahead with other than header type argument to lookahead with header type argument.
After P4::ExpandLookahead pass lookahead method may occur only as an assignment to a header or base bit type as follows:
```
T var_name;

state state_name {
  var_name = pkt.lookahead<T>();
}
```
Lookahead with header types arguments are unchanged by newly added pass ConvertLookahead.
Lookahead with other than header type argument is transformed to following:
```
header lookahead_tmp_hdr {
  T f;
}

T var_name;
lookahead_tmp_hdr lookahead_tmp;

state state_name {
  lookahead_tmp = pkt.lookahead<lookahead_tmp_hdr>();
  var_name = lookahead_tmp.f;
}
```
As part of this change, P4::ExpandLookahead pass is updated to have an optional parameter which allows to disable the expansion of headers (structures are still expanded).
Note (limitation):
Current maximum supported width of header fields in DPDK target is 64 bits.
As ExpandLookahead pass expands structures used as type arguments of lookahead method into bit type of the same width as the width of the whole structure and there is created a temporary variable of the given width which is than part of metadata structure, maximum supported width of the structure used as a type argument for lookahead is also 64 bits.

* Test cases are updated


testdata/p4_16_samples/pna-lookahead-structure.p4
```diff
@@ -0,0 +1,103 @@
+/*
+Copyright 2021 Intel Corporation
```
mbudiu-vmw, line 2: 2022?

midend/expandLookahead.h
```diff
@@ -36,6 +36,12 @@ class DoExpandLookahead : public Transform {
     P4::ReferenceMap* refMap;
     P4::TypeMap* typeMap;
     IR::IndexedVector<IR::Declaration> newDecls;
+    /**
```
mbudiu-vmw, line 39: I think you will have to document this flag in the class documentation as well, otherwise people may not notice it.
This is in fact a "policy" flag.

backends/dpdk/dpdkArch.h
```diff
@@ -67,6 +67,49 @@ class ConvertToDpdkArch : public Transform {
     }
 };
 
+// DPDK target supports lookahead instruction only with header operand.
```
mbudiu-vmw, line 70: Maybe you can use /// or /**, I think this generates documentation differently.

backends/dpdk/dpdkArch.cpp
```diff
@@ -239,6 +253,166 @@ const IR::Node *ConvertToDpdkArch::preorder(IR::Member *m) {
...
+
+    IR::ID newHeaderFieldName;
+    const IR::Type *newHeaderFieldType = nullptr;
+
+    /**
+     * Thanks to the fact that lookaheads are already expanded by P4::ExpandLookahead pass
+     * and structures are already flattened, following cases should cover all possibilities.
+     */
+    if (auto left = statement->left->to<IR::PathExpression>()) {
+        /**
```
mbudiu-vmw, line 280: The actual expression on the LHS does not really matter. Since you know that the header you synthesize has a single field you can always use the same field name. So this code can be simplified. You do need fresh names for the type and the variable.

backends/dpdk/dpdkArch.cpp
```diff
@@ -772,26 +948,112 @@ const IR::Node *ConvertBinaryOperationTo2Params::postorder(IR::P4Parser *a) {
...
+     * It may happen that there are header definitions in the source P4 program
+     * which are placed after the definition of main header structure, which is
+     * perfectly fine if they are not used in the main header structure.
+     * But if such header is used to declare local variable there is created
+     * a new field of the same header type in the main header structure.
+     * If the main header structure is placed before the definition of that
+     * header, it is a problem for Type Checking.
+     * There might also be header definitions internally created by compiler
+     * which would cause the same problem.
+     * Therefore here we move all structure and header definitons directly
```
mbudiu-vmw, line 979: I don't know if this is enough for the resulting code to always work. The headers may be used in other structs, not just in the main headers structure. In general, you need to insert the declarations before their first use and after the last declaration that they themselves use. 

But why can there be such headers? The headers you synthesize are not part of the main headers structure.

backends/dpdk/dpdkArch.cpp
```diff
@@ -772,26 +948,112 @@ const IR::Node *ConvertBinaryOperationTo2Params::postorder(IR::P4Parser *a) {
...
-                    s->fields.push_back(new IR::StructField(
-                        IR::ID(kv.first + "_" + dv->name.name), dv->type));
+                    auto type = typeMap->getType(dv, true);
+                    if (type->is<IR::Type_Header>()) {
+                        LOG3("Variable: " << dv << std::endl <<
+                             " type: " << type << std::endl <<
+                             " already added to: " << structure->header_type);
+                    } else {
+                        auto sf = new IR::StructField(
+                            IR::ID(kv.first + "_" + dv->name.name), dv->type);
```
mbudiu-vmw, line 1028: it's not obvious to me that this field name could not clash with another field name of the same struct.

mbudiu-vmw (changes requested):

backends/dpdk/dpdkArch.cpp
mvido: This is only because of new headers created by ConvertLookahead pass which creates new header definitions in:
```
const IR::Node *ConvertLookahead::postorder(IR::P4Program *program) {
    auto origProgram = getOriginal()->to<IR::P4Program>();
    if (newHeaderMap.count(origProgram)) {
        IR::Vector<IR::Node>::iterator insertIter = std::find_if(program->objects.begin(),
                program->objects.end(), [](const IR::Node *n){return n->is<IR::P4Parser>();});
        auto newHeaders = newHeaderMap.at(origProgram);
        program->objects.insert(insertIter, newHeaders.begin(), newHeaders.end());

        LOG2("Following new headers inserted into program " << dbp(program) << ":");
        for (auto h : newHeaders) {
            LOG2(" " << h);
        }
    }
    return program;
}
```
and places them before first parser definition and it also creates new local variables of those header types. So after ConvertLookahead pass the locations of those header definitions are before they are first used. But in this pass, for each local variable which is a header (which applies also for new local variables created in ConverLookahead pass) there is created a corresponding field in the main headers structure, and therefore the definitions of those headers must be before the main headers structure. The headers which are synthesized in ConverLookahead pass are used only in the declarations of local variables (which are already after the definitions since the beginning when they are created) and in the main headers structure (there are created new fields of that header type during this pass), so that's why I move all the header and structure definitions before the main headers structure. This also moves other header and structure definitions which are not synthesized in ConvertLookahead pass, but the order of definitions is kept the same and they are only moved forward, so the only structure definition they get ahead is the main header structure and I assume that no header or structure definition can use the main header structure to define its fields.

Other approach could be inserting new header definitions synthesized in ConvertLookahead pass directly before main header structure at the moment when they are created and not before the first parser. That way this code would not be needed. That would probably be better approach. Currently the problem is that the name of the main headers structure is parsed in CollectMetadataHeaderInfo pass which is after the place where I placed ConvertLookahead pass, but probably ConvertLookahead pass could be placed after CollectMetadataHeaderInfo or the name of the main header structure could be parsed sooner.
I'm going to try this approach.

mbudiu-vmw: This makes sense. However, if the code is never typechecked again after this pass you don't need the declarations in order really, you could simply generate code for them in the right order.

mvido: But TypeChecking is called at least twice after this CollectLocalVariables pass.

mvido: So do you think I should rather try this approach:

> Other approach could be inserting new header definitions synthesized in ConvertLookahead pass directly before main header structure at the moment when they are created and not before the first parser. That way this code would not be needed. That would probably be better approach. Currently the problem is that the name of the main headers structure is parsed in CollectMetadataHeaderInfo pass which is after the place where I placed ConvertLookahead pass, but probably ConvertLookahead pass could be placed after CollectMetadataHeaderInfo or the name of the main header structure could be parsed sooner.

or leave it as it is right now?

mbudiu-vmw: Then probably it is best to insert everything there in `postorder(Type_Struct*)`. If the struct is the user metadata (which you have already identified, just check the type name) just return a Vector containing all new types inserted in front of the type_struct itself.

mbudiu-vmw: BTW: this pattern is used several places in the compiler already, see for example ReplaceStructs.

mbudiu-vmw: Or better `DoReplaceTuples::insertReplacements`

mvido: Ok, will do. Thank you very much!

backends/dpdk/dpdkArch.cpp
mvido: In fact it was already done this way before. I think it relies on the fact that it is appending control/parser block name which has to be unique and local variable name which has to be unique in context of its control/parser block. But I think it could be theoretically possible to have the same name in the original P4 source code which could clash with this newly created name. I'm rather adding refMap->newName() here.

mvido: So it's not that simple. Same name, as here, needs to be used also in `const IR::Node *CollectLocalVariables::postorder(IR::PathExpression *p)`.
That would mean that the name would have to be generated in:
```
const IR::Node *CollectLocalVariables::preorder(IR::P4Program *p) {
    for (auto kv : structure->parsers) {
        locals_map.emplace(kv.first+"_parser", kv.second->parserLocals);
    }
    for (auto kv : structure->pipelines) {
        locals_map.emplace(kv.first, kv.second->controlLocals);
    }
    for (auto kv : structure->deparsers) {
        locals_map.emplace(kv.first+"_deparser", kv.second->controlLocals);
    }
```
which would require iterating over all controlLocals and creating the new names.

So I suggest to leave it as it is right now.

What do you think?

mbudiu-vmw: Many of our analyses have two passes: an Inspector which generates information (e.g., new names), and a Transform which makes the changes according to the generated information. This probably won't require much work - just moving some code around and storing results in a HashMap. I would say that safety is the most important goal, so you really should make sure names are unique.

backends/dpdk/backend.cpp
```diff
@@ -55,6 +55,9 @@ void DpdkBackend::convert(const IR::ToplevelBlock *tlb) {
         new BMV2::LowerExpressions(typeMap),
         new P4::RemoveComplexExpressions(refMap, typeMap,
                 new DPDK::ProcessControls(&structure.pipeline_controls)),
+        new P4::TypeChecking(refMap, typeMap),
```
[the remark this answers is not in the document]
mvido: done in 1c7ed7ce89e81b22ff1d4ff6d8dee98ea95d1508

midend/expandLookahead.h
mvido: done in 3a2ab3e2190678c87b9a8cba35fc181f18c91eea

backends/dpdk/dpdkArch.h
mvido: done in 0bf49e4d209d1f46741814913bc7f32627c099e2

backends/dpdk/dpdkArch.cpp
mvido: done in 91c72491f4eb19f08fe83e1c4140e03aee2cfc1f

backends/dpdk/dpdkArch.cpp
mvido: done in 81951e53812cf830549b10fec008a371266659d1

backends/dpdk/dpdkArch.cpp
mvido: I suggest to merge this change in separate PR, as there is a diff in quite a lot of spec files, so it would pollute this PR with unrelated changes. I prepared another PR with these changes based on this PR: https://github.com/p4lang/p4c/pull/3080, which could be merged when this PR is merged. What do you think?

mvido:
Changes from this PR were included in PR https://github.com/p4lang/p4c/pull/3080, which has been already merged to main, so I'm closing this PR.

PR #3052, opened 2022-01-26
Optional parameters may not have initializers

### LINKED ISSUE #3051: Issues with optional parameters
Not sure if the following program should be legal, but it makes `p4c` crash with a `SIGSEGV`.

```
control C();
package P(C a);

control MyC(@optional inout bool b) {
  apply {
     b = !b;
  } 
}

control MyD() {
  MyC() c;
  apply {
    c.apply();
  }
}

P(MyD()) main;
```

This is a testcase to try to make progress on https://github.com/p4lang/p4-spec/pull/987.

mbudiu-vmw:
Fixes #3051

### DIFF
--- a/frontends/p4/inlining.cpp
+++ b/frontends/p4/inlining.cpp
@@ -568,7 +568,10 @@ void GeneralInliner::inline_subst(P4Block *caller,
                     substs->paramSubst.add(param, initializer);
                     continue;
                 }
-                if (call != nullptr && (useTemporary.find(param) == useTemporary.end())) {
+                if (call != nullptr && (useTemporary.find(param) == useTemporary.end()) &&
+                    mi->substitution.contains(param))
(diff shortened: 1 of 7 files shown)


jnfoster (approved):
Confirm that this fixes the bug. 

Source changes LGTM.

jafingerhut:
I am completely confused about what the behavior of this program ought to be, and echo Nate's "I don't know if this should be a legal program or not" (paraphrased) comment on the original issue.

Why is the midend output considered equivalent to the input program?

mbudiu-vmw:
You could ask the same question about the frontend output.
The two programs have the same observable behavior from the control and data plane.

jnfoster:
Is it true that optional parameters that are not filled in by the invocation are just treated as if they were local uninitialized variables? That's a simple answer...

PR #3054, opened 2022-01-26
Solve some problems with cast

vladyslav-dubina:
Solving problem with `Cast` when the expression is a `SymbolicValue` .
Test for example : bool_to_bit_cast.p4,
@mbudiu-vmw Changes in the past PR proved useless. We found out that `PassRepeated` pass which is set in your midend does similar calculations. We believe that it is necessary to add a note that for the correct operation of `ParserUnroll`, should use `PassRepeated` before `ParserUnroll`.

### DIFF
--- a/midend/interpreter.cpp
+++ b/midend/interpreter.cpp
@@ -859,6 +859,16 @@ void ExpressionEvaluator::postorder(const IR::Operation_Unary* expression) {
         return;
     }
     if (sv->isUnknown()) {
+        if (auto cast = expression->to<IR::Cast>()) {
+            if (cast->destType->is<IR::Type_Boolean>() &&
+                l->is<SymbolicInteger>()) {
+                l = new SymbolicBool(sv->state);
+
+            } else if (cast->destType->is<IR::Type_Bits>() &&
+                       l->is<SymbolicBool>()) {
+                l = new SymbolicInteger(sv->state, cast->destType->to<IR::Type_Bits>());
+            }
+        }
         set(expression, l);
         return;
     }
@@ -1240,5 +1250,4 @@ SymbolicValue* ExpressionEvaluator::evaluate(const IR::Expression* expression, b
     auto result = get(expression);
     return result;
 }
-
 }  // namespace P4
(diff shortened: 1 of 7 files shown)


midend/interpreter.cpp
```diff
@@ -859,6 +859,16 @@ void ExpressionEvaluator::postorder(const IR::Operation_Unary* expression) {
         return;
     }
     if (sv->isUnknown()) {
+        if (auto cast = expression->to<IR::Cast>()) {
```
mbudiu-vmw, line 862: This does not handle the case of enums.
Can you please add a test with an enum and see if this needs to change as well?

mbudiu-vmw (changes requested):

midend/interpreter.cpp
vladyslav-dubina: @mbudiu-vmw  We've updated the test to include Enums and casts related to Enums. There is no need to make changes to the code, since in midend there is an `EliminateSerEnums ` pass that replaces Enums with the specified type

midend/interpreter.cpp
```diff
@@ -859,6 +859,16 @@ void ExpressionEvaluator::postorder(const IR::Operation_Unary* expression) {
...
+            if (cast->destType->is<IR::Type_Boolean>() &&
+                l->is<SymbolicInteger>()) {
+                l = new SymbolicBool(sv->state);
+
+            } else if (cast->destType->is<IR::Type_Bits>() &&
+                       l->is<SymbolicBool>()) {
+                l = new SymbolicInteger(sv->state, cast->destType->to<IR::Type_Bits>());
+            }
```
mbudiu-vmw, line 870: ok, then how about an additional check here: `else BUG("%1% unexpected type %2% in cast", cast, cast->destType)`.
This would ensure that the pass is "safe" even if someone changes the order of the midend passes. It will still crash the compiler, but it will give a somewhat clearer explanation.

mbudiu-vmw (approved):
I am approving, but perhaps you can make the suggested change before we merge.

midend/interpreter.cpp
```diff
@@ -859,6 +859,17 @@ void ExpressionEvaluator::postorder(const IR::Operation_Unary* expression) {
         return;
     }
     if (sv->isUnknown()) {
+        if (auto cast = expression->to<IR::Cast>()) {
+            if (cast->destType->is<IR::Type_Boolean>() &&
+                (l->is<SymbolicInteger>() || l->is<SymbolicBool>())) {
```
vladyslav-dubina, line 864: Additional conditions are needed in order not to fall into the `else` case when processing replaced enums in cast

PR #3055, opened 2022-01-27
DPDK Backend: Add support for general switch statements

usha1830:


backends/dpdk/dpdkHelpers.cpp
```diff
@@ -786,24 +786,49 @@ bool ConvertStatementToDpdk::preorder(const IR::MethodCallStatement *s) {
     return false;
 }
 
-// This function assumes EliminateSwitch midend pass is applied and only handles
-// action_run in switch expression.
+// If we get a switch statement with constant switch expression, only emit the body of
+// matching case label
+bool ConvertStatementToDpdk::handleConstSwitch(const IR::SwitchStatement *s) {
```
mbudiu-vmw, line 791: I think we already have this optimization in the compiler, don't we?
Do you see this triggered?

backends/dpdk/dpdkHelpers.cpp
```diff
@@ -786,24 +786,49 @@ bool ConvertStatementToDpdk::preorder(const IR::MethodCallStatement *s) {
...
+    auto constSwitchExpr = s->expression->to<IR::Constant>();
+    bool constSwitchFlag = false;
```
mbudiu-vmw, line 793: Just for safety I would add here a `CHECK_NULL(constSwitchExpr)`.
In general, you should check invariants that must be ensured by callers, because code may be refactored. Alternatively, you could have here the check for Constant and return right away if it is not constant. I think that's a better design.

mbudiu-vmw (approved):

backends/dpdk/dpdkHelpers.cpp
usha1830: Yes. SimplifySwitch works on optimizing the constant expression in switch.
However, this case is hitting after Midend does copy propagation. Calling the SimplifySwitch pass again after local copy propagation and constant folding works. Shall I update the PR with this change?

mbudiu-vmw: Should be shorter too

usha1830: This isn't working as expected for certain cases. I will update the PR after a little more testing.

usha1830: @mbudiu-vmw I updated the code to use the SimplifySwitch pass. 
Updated the SimplifySwitch pass to handle the case where the matching case label does not have associated block statement body and should fallthrough.

```
switch(16) {
    case 16:
    case 32: {   // block statement to be executed for both case 16 and case 32}
}
```

Please take a look.

mbudiu-vmw: Is this a bug in the original code?

frontends/p4/simplifySwitch.cpp
```diff
@@ -41,9 +41,15 @@ bool DoSimplifySwitch::matches(const IR::Expression* left, const IR::Expression*
 const IR::Node* DoSimplifySwitch::postorder(IR::SwitchStatement* stat) {
     if (!typeMap->isCompileTimeConstant(stat->expression))
         return stat;
+    bool foundMatch = false;
     for (auto ss : stat->cases) {
-        if (matches(ss->label, stat->expression))
+        if (matches(ss->label, stat->expression)){
```
mbudiu-vmw, line 46: so it looks like this was a bug.
Would be great if there was a test case for this bug.

backends/dpdk/dpdkHelpers.cpp
```diff
@@ -786,24 +786,29 @@ bool ConvertStatementToDpdk::preorder(const IR::MethodCallStatement *s) {
...
+    // each case statement.
     for (unsigned i = 0; i < size - 1; i++) {
         auto caseLabel = s->cases.at(i);
-        auto pe = caseLabel->label->to<IR::PathExpression>();
-        CHECK_NULL(pe);
-        label = refmap->newName("label_action");
-        add_instr(new IR::DpdkJmpIfActionRunStatement(label, pe->path->name.name));
+        label = refmap->newName("label_switch");
+        if (tc != nullptr) {
+            auto pe = caseLabel->label->to<IR::PathExpression>();
```
mbudiu-vmw, line 806: alternatively, we have a toChecked function...

backends/dpdk/dpdkHelpers.cpp
```diff
@@ -786,24 +786,29 @@ bool ConvertStatementToDpdk::preorder(const IR::MethodCallStatement *s) {
...
+
     auto size = s->cases.size();
     std::vector <cstring> labels;
     cstring label;
     cstring default_label = "";
     auto end_label = refmap->newName("label_endswitch");
-    // Emit jmp on action run statements and collect labels for each case statement
+
+    // Emit jmp on action run/jmp on matching label statements and collect labels for
+    // each case statement.
```
mbudiu-vmw, line 801: can this crash if size == 0?

mbudiu-vmw (approved):

frontends/p4/simplifySwitch.cpp
mbudiu-vmw: missing a space here

usha1830:
@mbudiu-vmw If there are no further comments, could you please merge this?

mbudiu-vmw (approved):

PR #3061, opened 2022-01-28
Solve problems with verify which is created in ParserUnroll

vladyslav-dubina:
Helps solve the problem of converting the second argument of `verify ` to a constant using the `ConvertEnums ` pass after `ParserUnroll`. This only applies to `verify` which is created in `ParserUnroll`.

### DIFF
--- a/midend/parserUnroll.h
+++ b/midend/parserUnroll.h
@@ -205,6 +205,7 @@ class RewriteAllParsers : public Transform {
             arguments->push_back(
                 new IR::Argument(new IR::BoolLiteral(IR::Type::Boolean::get(), false)));
             arguments->push_back(new IR::Argument(new IR::Member(
+                new IR::Type_Error(IR::ID("error")),
                 new IR::TypeNameExpression(new IR::Type_Name(IR::ID("error"))),
                     IR::ID("StackOutOfBounds"))));
             IR::IndexedVector<IR::StatOrDecl> components;


VolodymyrPeschanenkoLitSoft (approved):
I think that this is a good solution, because in that case `error.StackOutOfBounds` after  `parserUnroll` will be completely compatible with `ConvertEnums`. Otherwise, the `ConvertEnums` can't do anything with generated `error.StackOutOfBounds`.

VolodymyrPeschanenkoLitSoft:
@vladyslav-dubina Can't you specify a type for error.StackOutOfBounts when it comes from a p4 program in p4test before parserUnroll? If it is Type_Error then these changes looks fine. If not then you need to translate it here. Otherwise, all users should eliminate it after the application of parserUnroll.

VolodymyrPeschanenkoLitSoft:
I think that this is not a failure. it's more about the usability of `parserUnroll`. Let's imagine the situation where a user should represent `errors.StackOutOfbounds` as a constant after an application of the parserUnroll. So, this user should add the following passes after `parserUnroll` : 1) `TypeChecking` to setup type for `errors.StackOutOfbounds`;  2) `ConvertEnums` to replace Type_Error into bit.

From my point of view, it's too expensive to use two passes for one instance of `errors.StackOutOfBound`.

I think that the interface of the parserUnroll should provide a choice, how a user wants to represent `errors.StackOutofBound`: as `Type_Error` or `bit`. In this case, no additional passes are needed. 

What do you think about it?

vladyslav-dubina:
I agree with the words of Volodymyr, after conducting an experiment, I went through a lot of midend passes, but I could not achieve that after converting error to constant, a working program was obtained.

mbudiu-vmw:
One way is indeed to parameterize this pass with information about how errors are represented.
The better solution would be to reorder the midend passes so that error elimination is done after unrolling.

vladyslav-dubina:
We tried setting `TypeChecking` pass  as well but it didn't set the types, in which case you would have to check each member for Error

VolodymyrPeschanenkoLitSoft:
I think that` IR::Type_Error` is not a part of `IR::Program` and in this case `TypeCheckinng` can't help here. It means that if `errors.StackOutOfBound` exists without a type then the pass should analyze each name of a member for replacement of errors. So, I think that the current changes should be committed. @mbudiu-vmw What do you think about it?

So, now we have two options according to @mbudiu-vmw suggestion:
> One way is indeed to parameterize this pass with information about how errors are represented. The better solution would be to reorder the midend passes so that error elimination is done after unrolling.

@fruffy, Could you, please, advise your opinion on that?

mbudiu-vmw:
You do have some dependencies: error elimination should probably be done before enum elimination, and currently parserUnroll may assume that enums have been eliminated. So you may need to make sure the interpreter handles correctly enum values. But otherwise I think moving the enum passes after unrolling is the way to go.

fruffy:
I think reordering and handle enums/errors is the best choice. I would add a note to the parserUnroll pass that it must be executed before any enum/error elimination since it inserts `error.StackOutOfbounds`. Are there any other problems with this approach?

VolodymyrPeschanenkoLitSoft:
@fruffy, @mbudiu-vmw. I think that reordering is a good solution, but I think that we need to add a type for `errors.StackOutOfBound`. Otherwise, the elimination pass should check all members' names without any restriction like `IR::Type_Error`.

mbudiu-vmw:
That is why almost every pass runs typechecking before it does its work: to compute types for the new IR nodes.
It should just work, and add the right type for error.StackOutOfBounds.

VolodymyrPeschanenkoLitSoft:
> That is why almost every pass runs typechecking before it does its work: to compute types for the new IR nodes. It should just work, and add the right type for error.StackOutOfBounds.

@mbudiu-vmw, Thanks. The typechecking helps here.

mbudiu-vmw:
If the typechecking call is required here it means there is a bug somewhere else in the compiler which we should figure out.

PR #3062, opened 2022-01-28
DPDK Backend: Fix a bug in handling GEQ and LEQ operations

usha1830:
This PR fixes an issue with the order in which the labels and jumps were generated for GEQ and LEQ operations. 

Since dpdk does have a conditional jump for GEQ and LEQ, we negate the condition. However, the true and false blocks were not swapped which was leading to unreachable code.


mbudiu-vmw (approved):

usha1830:
@mbudiu-vmw Thanks for the review! Could you please merge this PR?

PR #3063, opened 2022-01-31
Solving problems with Header Union verify function 

vladyslav-dubina:
For testing we can use tests: issue1897-bmv2.p4 , invalid-hdr-warnings6.p4 , issue561-bmv2.p4 and invalid-hdr-warnings5.p4 with --loopsUnroll argument.

This PR refers to https://github.com/p4lang/p4c/pull/2980


midend/interpreter.cpp
```diff
@@ -309,22 +309,38 @@ void SymbolicStruct::dbprint(std::ostream& out) const {
...
+    }
+  }
+  valid = new SymbolicBool(v);
 }
 
 SymbolicValue* SymbolicHeaderUnion::get(const IR::Node* node, cstring field) const {
-    if (valid->isKnown() && !valid->value)
-        return new SymbolicStaticError(node, "Reading field from invalid header union");
-    return SymbolicStruct::get(node, field);
+  if (valid->isKnown() && !valid->value)
```
mbudiu-vmw, line 340: I don't really understand you. How hard is it to use 4 spaces for indentation?
So let's say that you do use 2 spaces only, but why reindent code that you don't even modify?
Can you please check the PRs before you submit them to see if indentation is wrong?

vladyslav-dubina: I checked with cpplint and clang-format and they didn't show any indentation errors, but if there are I'll correct them manually

vladyslav-dubina: I made indentation fixes

midend/interpreter.cpp
```diff
@@ -309,21 +309,36 @@ void SymbolicStruct::dbprint(std::ostream& out) const {
     out << " }";
 }
 
-SymbolicHeaderUnion::SymbolicHeaderUnion(const IR::Type_HeaderUnion* type,
-                               bool uninitialized,
-                               const SymbolicValueFactory* factory) :
+SymbolicHeaderUnion::SymbolicHeaderUnion(const IR::Type_HeaderUnion *type,
```
fruffy, line 312: I believe p4c uses left-hand pointer style.

midend/interpreter.cpp
```diff
@@ -1095,11 +1110,25 @@ void ExpressionEvaluator::postorder(const IR::MethodCallExpression* expression)
         auto bim = mi->to<BuiltInMethod>();
         auto base = get(bim->appliedTo);
         cstring name = bim->name.name;
+        const IR::Expression *node = nullptr;
```
fruffy, line 1113: Here too.

midend/interpreter.cpp
```diff
@@ -1124,11 +1153,17 @@ void ExpressionEvaluator::postorder(const IR::MethodCallExpression* expression)
...
-            auto hv = base->to<SymbolicHeader>();
-            auto v = hv->valid;
-            set(expression, v);
-            return;
+            if (auto hv = structVar->to<SymbolicHeader>()) {
+                auto v = hv->valid;
+                set(expression, v);
+                return;
+            } else if (auto huv = structVar->to<SymbolicHeaderUnion>()) {
+                auto v = huv->valid;
```
fruffy, line 1161: Considering that validity of a header union is dependent on its members, you may just want to check if one member is valid. If multiple members are valid that should be an assertion error.

midend/interpreter.cpp
```diff
@@ -1095,11 +1110,25 @@ void ExpressionEvaluator::postorder(const IR::MethodCallExpression* expression)
         auto bim = mi->to<BuiltInMethod>();
         auto base = get(bim->appliedTo);
         cstring name = bim->name.name;
+        const IR::Expression* node = nullptr;
```
fruffy, line 1113: Please add a comment that this code is necessary to find the parent and to check whether the parent is a header union.

midend/interpreter.h
```diff
@@ -420,7 +420,7 @@ class SymbolicHeaderUnion : public SymbolicStruct {
     SymbolicBool* valid = nullptr;
     SymbolicHeaderUnion(const IR::Type_HeaderUnion* type, bool uninitialized,
                    const SymbolicValueFactory* factory);
-    virtual void setValid(bool v);
+    virtual void setValid(bool v, cstring field = nullptr);
```
fruffy, line 423: What does it mean for the field to be nullptr? I would not allow this.

midend/interpreter.cpp
```diff
@@ -310,20 +310,35 @@ void SymbolicStruct::dbprint(std::ostream& out) const {
...
-        setAllUnknown();
+void SymbolicHeaderUnion::setValid(bool v, cstring field) {
+    if (field) {
+        fieldValue[field]->to<SymbolicHeader>()->setValid(v);
+        for (auto f : type->to<IR::Type_StructLike>()->fields) {
+            if (f->name.name != field)
+                fieldValue[f->name.name]->setAllUnknown();
+        }
+    }
     valid = new SymbolicBool(v);
```
fruffy, line 336: Should a Header Union have a valid field? Considering it is dependent on its members.

fruffy (changes requested):

midend/interpreter.h
vladyslav-dubina: This is necessary to check and separate the setValid call for the entire HeaderUnion or for its field

midend/interpreter.cpp
vladyslav-dubina: I'm not sure about this, I think it should be

midend/interpreter.cpp
vladyslav-dubina: I do not quite understand how this relates to this part of the code, but here I just set the received value

fruffy: This is a method call on isValid right? A union is actually valid according to its members (https://p4.org/p4-spec/docs/P4-16-v1.2.2.html#sec-expr-hu). It technically does not have a validity bit. This relates to the other part of the comments, which ask whether we even need a valid field for a header union.

midend/interpreter.h
fruffy: You can not really set a header union valid or invalid, since that depends on its members. So similarly, we should not need this.

testdata/p4_16_samples_outputs/pna-extract-local-header.p4-error
```diff
@@ -10,3 +10,12 @@ parser MainParserT<PM, MH, MM>(
 pna.p4(680): [--Wwarn=unused] warning: 'PM' is unused
 control MainControlT<PM, MH, MM>(
                      ^^
+pna-extract-local-header.p4(58): [--Wwarn=ignore-prop] warning: Result of AssignmentStatement is not defined: Error: Reading field from invalid header
+        hdr.h2.type1 = local_hdr.type1;
+                     ^
+pna-extract-local-header.p4(59): [--Wwarn=ignore-prop] warning: Result of AssignmentStatement is not defined: Error: Reading field from invalid header
```
fruffy, line 16: This error message does not make sense. Is it an error or a warning?

midend/interpreter.cpp
```diff
@@ -1105,11 +1129,27 @@ void ExpressionEvaluator::postorder(const IR::MethodCallExpression* expression)
         auto bim = mi->to<BuiltInMethod>();
         auto base = get(bim->appliedTo);
         cstring name = bim->name.name;
+        // This code is necessary to find the parent and to check whether the parent
+        // is a header union.
+        const IR::Expression* node = nullptr;
+        cstring memberName = nullptr;
+        if (auto member = expression->method->to<IR::Member>()
```
fruffy, line 1136: Use checkedTo or do not stack to calls, this can easily lead to a nullptr access.

midend/interpreter.cpp
```diff
@@ -1105,11 +1129,27 @@ void ExpressionEvaluator::postorder(const IR::MethodCallExpression* expression)
...
+                                ->expr->to<IR::Member>()) {
+            node = member->expr;
+            memberName = member->member.name;
+        } else if (auto expr = expression->method->to<IR::Member>()->expr) {
```
fruffy, line 1140: Same here.

midend/interpreter.cpp
```diff
@@ -374,6 +396,8 @@ bool SymbolicHeaderUnion::equals(const SymbolicValue* other) const {
 void SymbolicHeaderUnion::dbprint(std::ostream& out) const {
     out << "{ ";
     out << "valid=>";
+    auto hu = new SymbolicHeaderUnion(type->to<IR::Type_HeaderUnion>());
```
fruffy, line 399: This should also use checkedTo

midend/interpreter.cpp
```diff
@@ -355,15 +373,19 @@ bool SymbolicHeaderUnion::merge(const SymbolicValue* other) {
...
+    changes = changes || valid->merge(hv->isValid());
     return changes;
 }
 
 bool SymbolicHeaderUnion::equals(const SymbolicValue* other) const {
     if (!other->is<SymbolicHeaderUnion>())
         return false;
     auto sh = other->to<SymbolicHeaderUnion>();
-    if (!valid->equals(sh->valid))
+    auto hu = new SymbolicHeaderUnion(type->to<IR::Type_HeaderUnion>());
```
fruffy, line 386: checkedTo

midend/interpreter.cpp
```diff
@@ -310,33 +310,49 @@ void SymbolicStruct::dbprint(std::ostream& out) const {
...
+    if (hu->isValid()->isKnown()  && !hu->isValid()->value)
+        return new SymbolicStaticError(node,"Reading field from invalid header union");
     return SymbolicStruct::get(node, field);
 }
 
 void SymbolicHeaderUnion::setAllUnknown() {
     SymbolicStruct::setAllUnknown();
-    valid->setAllUnknown();
+    for (auto f : type->to<IR::Type_StructLike>()->fields) {
```
fruffy, line 347: checkedto

mbudiu-vmw:
Give me a chance to test this. The way it's written no tests will run as part of the CI.
Perhaps it's time to have a special class of tests with loop unrolling.

testdata/p4_16_samples_outputs/pna-extract-local-header.p4-error
vladyslav-dubina: These errors appeared when running the test on git, they are not on the local, I had to add

vladyslav-dubina:
#2921 does not have these changes, as I understand it, judging by the interpreter.cpp in #2921 . For this reason, part of the tests in #2921 failed with an error
`Compiler Bug: Error: Reading field from invalid header union: expected a header`

midend/interpreter.cpp
```diff
@@ -310,33 +310,49 @@ void SymbolicStruct::dbprint(std::ostream& out) const {
...
+                return new SymbolicBool(true);
+         }
+    }
+    return new SymbolicBool(false);
 }
 
 SymbolicValue* SymbolicHeaderUnion::get(const IR::Node* node, cstring field) const {
-    if (valid->isKnown() && !valid->value)
-        return new SymbolicStaticError(node, "Reading field from invalid header union");
+    auto hu = new SymbolicHeaderUnion(type->to<IR::Type_HeaderUnion>());
```
mbudiu-vmw, line 339: why do you allocate here a new SymbolicHeaderUnion? 
that one will never be valid by construction, so the next statement does nothing.

midend/interpreter.cpp
```diff
@@ -346,7 +362,9 @@ void SymbolicHeaderUnion::assign(const SymbolicValue* other) {
     BUG_CHECK(hv, "%1%: expected a header union", other);
     for (auto f : hv->fieldValue)
         fieldValue[f.first]->assign(f.second);
-    valid->assign(hv->valid);
+    auto hu = new SymbolicHeaderUnion(type->to<IR::Type_HeaderUnion>());
```
mbudiu-vmw, line 365: same question here, this looks wrong.

midend/interpreter.cpp
```diff
@@ -355,15 +373,19 @@ bool SymbolicHeaderUnion::merge(const SymbolicValue* other) {
...
+    changes = changes || valid->merge(hv->isValid());
     return changes;
 }
 
 bool SymbolicHeaderUnion::equals(const SymbolicValue* other) const {
     if (!other->is<SymbolicHeaderUnion>())
         return false;
     auto sh = other->to<SymbolicHeaderUnion>();
-    if (!valid->equals(sh->valid))
+    auto hu = new SymbolicHeaderUnion(type->checkedTo<IR::Type_HeaderUnion>());
```
mbudiu-vmw, line 386: same here.

midend/interpreter.cpp
```diff
@@ -310,33 +310,49 @@ void SymbolicStruct::dbprint(std::ostream& out) const {
...
 
-void SymbolicHeaderUnion::setValid(bool v) {
-    if (!v)
-        setAllUnknown();
-    valid = new SymbolicBool(v);
+SymbolicBool* SymbolicHeaderUnion::isValid() const {
+    for (auto f : type->to<IR::Type_StructLike>()->fields) {
+         auto fieldsClone = fieldValue;
+         if (fieldsClone[f->name.name]) {
+             if (fieldsClone[f->name.name]->to<SymbolicHeader>()->valid->value)
```
mbudiu-vmw, line 331: you should check whether the value is unknown before you look at it.

mbudiu-vmw (changes requested):

midend/interpreter.cpp
```diff
@@ -310,33 +310,50 @@ void SymbolicStruct::dbprint(std::ostream& out) const {
...
-                               bool uninitialized,
-                               const SymbolicValueFactory* factory) :
-        SymbolicStruct(type, uninitialized, factory),
-        valid(new SymbolicBool(false)) {}
+                                         bool uninitialized,
+                                         const SymbolicValueFactory* factory) :
+        SymbolicStruct(type, uninitialized, factory) {}
+
+void SymbolicHeaderUnion::setValid(bool v, cstring field) {
+    if (field) {
```
mbudiu-vmw, line 318: why would field here be the empty string?

midend/interpreter.cpp
```diff
@@ -310,33 +310,50 @@ void SymbolicStruct::dbprint(std::ostream& out) const {
...
+        fieldValue[field]->to<SymbolicHeader>()->setValid(v);
+        for (auto f : type->to<IR::Type_StructLike>()->fields) {
+            if (f->name.name != field)
+                fieldValue[f->name.name]->setAllUnknown();
```
mbudiu-vmw, line 322: I think you need to call `setValid(false)` here.

midend/interpreter.cpp
```diff
@@ -310,33 +310,50 @@ void SymbolicStruct::dbprint(std::ostream& out) const {
...
+    }
+}
 
-void SymbolicHeaderUnion::setValid(bool v) {
-    if (!v)
-        setAllUnknown();
-    valid = new SymbolicBool(v);
+SymbolicBool* SymbolicHeaderUnion::isValid() const {
+    for (auto f : type->to<IR::Type_StructLike>()->fields) {
+         auto fieldsClone = fieldValue;
```
mbudiu-vmw, line 329: why make a copy here?

midend/interpreter.cpp
```diff
@@ -310,33 +310,50 @@ void SymbolicStruct::dbprint(std::ostream& out) const {
...
+         if (fieldsClone[f->name.name]) {
```
mbudiu-vmw, line 330: why would this be nullptr?

midend/interpreter.cpp
```diff
@@ -310,33 +310,50 @@ void SymbolicStruct::dbprint(std::ostream& out) const {
...
+             if (auto fildValid = fieldsClone[f->name.name]->to<SymbolicHeader>()->valid) {
```
mbudiu-vmw, line 331: why would this be nullptr?

midend/interpreter.cpp
```diff
@@ -310,33 +310,50 @@ void SymbolicStruct::dbprint(std::ostream& out) const {
...
+             }
+         }
+    }
+    return new SymbolicBool(false);
 }
 
 SymbolicValue* SymbolicHeaderUnion::get(const IR::Node* node, cstring field) const {
-    if (valid->isKnown() && !valid->value)
-        return new SymbolicStaticError(node, "Reading field from invalid header union");
+    if (this->isValid()->isKnown()  && !this->isValid()->value)
```
mbudiu-vmw, line 341: is this line really necessary?

midend/interpreter.cpp
```diff
@@ -310,33 +310,50 @@ void SymbolicStruct::dbprint(std::ostream& out) const {
...
+        return new SymbolicStaticError(node,"Reading field from invalid header union");
     return SymbolicStruct::get(node, field);
 }
 
 void SymbolicHeaderUnion::setAllUnknown() {
     SymbolicStruct::setAllUnknown();
-    valid->setAllUnknown();
+    for (auto f : type->checkedTo<IR::Type_StructLike>()->fields) {
```
mbudiu-vmw, line 348: how is this for loop different from the one above?

midend/interpreter.cpp
```diff
@@ -346,7 +363,8 @@ void SymbolicHeaderUnion::assign(const SymbolicValue* other) {
     BUG_CHECK(hv, "%1%: expected a header union", other);
     for (auto f : hv->fieldValue)
         fieldValue[f.first]->assign(f.second);
-    valid->assign(hv->valid);
+    auto valid = this->isValid();
+    valid->assign(hv->isValid());
```
mbudiu-vmw, line 367: this does nothing

midend/interpreter.cpp
```diff
@@ -1105,11 +1126,27 @@ void ExpressionEvaluator::postorder(const IR::MethodCallExpression* expression)
...
+        CHECK_NULL(node);
+        auto structVar = get(node);
         if (name == IR::Type_Header::setInvalid ||
             name == IR::Type_Header::setValid) {
-            BUG_CHECK(base->is<SymbolicHeader>(), "%1%: expected a header", base);
-            auto hv = base->to<SymbolicHeader>();
-            hv->setValid(name == IR::Type_Header::setValid);
+            if (auto hv = structVar->to<SymbolicHeader>()) {
+                hv->setValid(name == IR::Type_Header::setValid);
+            } else if (auto headerUnion = structVar->to<SymbolicHeaderUnion>()) {
```
mbudiu-vmw, line 1146: I don't think there is a setInvalid method call on a header_union.

mbudiu-vmw: this is not the right place to evaluate the method call, it should be done in the `postorder(MethodCallExpression* )` method.

mbudiu-vmw (changes requested):

midend/interpreter.cpp
OleksandrLetychevskyiIntel: I'm not even sure, can we read header from unvalid headerUnion?

midend/interpreter.cpp
OleksandrLetychevskyiIntel: I don't quite understand what you mean

midend/interpreter.cpp
OleksandrLetychevskyiIntel: It is the same in  `postorder(MethodCallExpression* )` . 
And this call is needed so structVar will not be passed to SymbolicHeader. And for HeadeUnion, we have implemented the setValid function in which the name of the field is passed to which setValid will be called and additionally valid will be set to other fields

midend/interpreter.cpp
jafingerhut: The language spec definitely says that for a normal header, if you attempt to use one of its fields while it is invalid, then the "read" of the field is allowed to return any arbitrary value (basically, giving free reign to the implementation for whatever it wants to do).  The language spec should say somewhere that this is also the case for header unions, but I have not scanned the language spec for that question specifically on header unions today.  If you look for it and do not find any such mention, let me know.  It should be in this section: https://p4.org/p4-spec/docs/P4-16-v1.2.2.html#sec-uninitialized-values-and-writing-invalid-headers

jnfoster: Reading from that section...

> Where a header is mentioned, it may be a member of a header_union, an element in a header stack, or a normal header.

midend/interpreter.cpp
mbudiu-vmw: setAllUnknown() should do the same thing.

midend/interpreter.cpp
OleksandrLetychevskyiIntel: I relied on the Header implementation here https://github.com/vladyslav-dubina/p4c/blob/b1034c8742c5d301913584b0be59291cc5575251/midend/interpreter.cpp#L342
And tried to make similar behavior. I don't quite understand what's wrong.

OleksandrLetychevskyiIntel: I added the extract_for_header_union.p4 test, it shows the ability to read a field from an invalid header union , and it prints a warning about reading from an invalid header union

midend/interpreter.cpp
OleksandrLetychevskyiIntel: I fixed it

midend/interpreter.cpp
OleksandrLetychevskyiIntel: I fixed it

midend/interpreter.cpp
OleksandrLetychevskyiIntel: I fixed it

midend/interpreter.cpp
OleksandrLetychevskyiIntel: I fixed it

midend/interpreter.cpp
OleksandrLetychevskyiIntel: I fixed it

midend/interpreter.cpp
OleksandrLetychevskyiIntel: I fixed it

midend/interpreter.cpp
OleksandrLetychevskyiIntel: I fixed it

OleksandrLetychevskyiIntel:
Ready for review

midend/interpreter.cpp
mbudiu-vmw: Let me repeat: there is no setValid or setInvalid call on a header_union. So this case can never happen.

mbudiu-vmw (changes requested):

midend/interpreter.cpp
```diff
@@ -310,33 +310,50 @@ void SymbolicStruct::dbprint(std::ostream& out) const {
...
+        if (f->name.name != field)
+            fieldValue[f->name.name]->checkedTo<SymbolicHeader>()->setValid(false);
+    }
+}
 
-void SymbolicHeaderUnion::setValid(bool v) {
-    if (!v)
-        setAllUnknown();
-    valid = new SymbolicBool(v);
+SymbolicBool* SymbolicHeaderUnion::isValid() const {
```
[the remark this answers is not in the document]
OleksandrLetychevskyiIntel: I fixed it

midend/interpreter.cpp
```diff
@@ -310,33 +310,50 @@ void SymbolicStruct::dbprint(std::ostream& out) const {
...
+    for (auto f : type->to<IR::Type_StructLike>()->fields) {
+         if (fieldValue.count(f->name.name)) {
+             if (const auto fildValid = fieldValue.at(f->name.name)
+                                       ->checkedTo<SymbolicHeader>()
+                                       ->valid) {
+                 if (fildValid->isKnown())
+                     return new SymbolicBool(fildValid->value);
```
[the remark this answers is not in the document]
OleksandrLetychevskyiIntel: I fixed it

midend/interpreter.cpp
```diff
@@ -355,15 +371,17 @@ bool SymbolicHeaderUnion::merge(const SymbolicValue* other) {
     bool changes = false;
     for (auto f : hv->fieldValue)
         changes = changes || fieldValue[f.first]->merge(f.second);
-    changes = changes || valid->merge(hv->valid);
+    auto valid = this->isValid();
```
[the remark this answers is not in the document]
OleksandrLetychevskyiIntel: I fixed it

midend/interpreter.cpp
```diff
@@ -355,15 +371,17 @@ bool SymbolicHeaderUnion::merge(const SymbolicValue* other) {
...
+    changes = changes || valid->merge(hv->isValid());
     return changes;
 }
 
 bool SymbolicHeaderUnion::equals(const SymbolicValue* other) const {
     if (!other->is<SymbolicHeaderUnion>())
         return false;
     auto sh = other->to<SymbolicHeaderUnion>();
-    if (!valid->equals(sh->valid))
+    auto valid = this->isValid();
```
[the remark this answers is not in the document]
OleksandrLetychevskyiIntel: I fixed it

midend/interpreter.cpp
```diff
@@ -374,6 +392,7 @@ bool SymbolicHeaderUnion::equals(const SymbolicValue* other) const {
 void SymbolicHeaderUnion::dbprint(std::ostream& out) const {
     out << "{ ";
     out << "valid=>";
+    auto valid = this->isValid();
```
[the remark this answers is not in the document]
OleksandrLetychevskyiIntel: I fixed it

OleksandrLetychevskyiIntel:
I've made changes and the PR is ready for review.

midend/interpreter.cpp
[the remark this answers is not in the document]
mbudiu-vmw: The rule to me seems to be:
- if any field has the value "unknown" or "uninitialized", the result is "unknown"
- otherwise if exactly one field is valid the result SymblicBool(true), which is the really field->value.
 (it should not be possible for more than 1 to be true, that should be a BUG).
- otherwise if no field is valid the result is SymbolicBool(false).

The code below does not seem to do that, in particular I don't see the check for uninitialized or unknown.

mbudiu-vmw (changes requested):

midend/interpreter.cpp
```diff
@@ -310,33 +310,45 @@ void SymbolicStruct::dbprint(std::ostream& out) const {
...
-void SymbolicHeaderUnion::setValid(bool v) {
-    if (!v)
-        setAllUnknown();
-    valid = new SymbolicBool(v);
+                                         bool uninitialized,
+                                         const SymbolicValueFactory* factory) :
+        SymbolicStruct(type, uninitialized, factory) {}
+
+SymbolicBool* SymbolicHeaderUnion::isValid() const {
+    int validFields = 0;
```
OleksandrLetychevskyiIntel, line 318: This is necessary to calculate if `HeaderUnion `has more than one valid field

midend/interpreter.cpp
```diff
@@ -310,33 +310,45 @@ void SymbolicStruct::dbprint(std::ostream& out) const {
...
+    for (auto f : type->to<IR::Type_StructLike>()->fields) {
+         if (fieldValue.count(f->name.name)) {
+             auto fieldValid = fieldValue.at(f->name.name)->checkedTo<SymbolicHeader>()->valid;
+             if (!fieldValid->isKnown() || fieldValid->isUninitialized()) {
+                return new SymbolicBool(false);
```
OleksandrLetychevskyiIntel, line 323: if any field has the value "unknown" or "uninitialized", the result is "unknown"

midend/interpreter.cpp
```diff
@@ -310,33 +310,45 @@ void SymbolicStruct::dbprint(std::ostream& out) const {
...
+                validFields +=1;
+             }
+         } else {
+             BUG("The number of fields in %1% is different from HeaderUnion fieldValue", type);
+         }
+    }
+    if (validFields == 1) {
+        return new SymbolicBool(true);
+    } else if (validFields > 1) {
+        BUG("In HeaderUnion cannot be more than one valid field");
```
OleksandrLetychevskyiIntel, line 334: Throw an error if more than one field is valid

midend/interpreter.cpp
```diff
@@ -310,33 +310,45 @@ void SymbolicStruct::dbprint(std::ostream& out) const {
...
+    }
+    return new SymbolicBool(false);
```
OleksandrLetychevskyiIntel, line 336: if no field is valid the result is SymbolicBool(false)

midend/interpreter.cpp
jfingerh: If I understand your statement correctly (perhaps I do not), according to the P4_16 language specification, it is guaranteed that at most one member of a header union is valid at a time.  If member m1 of a header union is valid, and later you make a different member m2 valid, then automatically at the same time that member m2 becomes valid, m1 must become invalid.

mbudiu-vmw: Indeed, below he calls BUG if this happens.

midend/interpreter.cpp
OleksandrLetychevskyiIntel: Oh I got the error

midend/interpreter.cpp
OleksandrLetychevskyiIntel: But this function doesn't validate anyone, it just evaluates valid for HeaderUnion.

midend/interpreter.cpp
OleksandrLetychevskyiIntel: I do not quite understand how we can return "unknown", do something similar to this?
`return new SymbolicBool(ScalarValue::ValueState::NotConstant))`

midend/interpreter.cpp
OleksandrLetychevskyiIntel: I probably understood that when I call setValid for Header from HeaderUnion, I need to check that HeaderUnion is not valid now, if valid make it setAllUnknown and then set valid for Header

OleksandrLetychevskyiIntel:
@mbudiu-vmw  This PR is ready for review

mbudiu-vmw (changes requested):

midend/interpreter.cpp
```diff
@@ -310,33 +310,45 @@ void SymbolicStruct::dbprint(std::ostream& out) const {
...
+                                         const SymbolicValueFactory* factory) :
+        SymbolicStruct(type, uninitialized, factory) {}
+
+SymbolicBool* SymbolicHeaderUnion::isValid() const {
+    int validFields = 0;
+    for (auto f : type->to<IR::Type_StructLike>()->fields) {
+         if (fieldValue.count(f->name.name)) {
+             auto fieldValid = fieldValue.at(f->name.name)->checkedTo<SymbolicHeader>()->valid;
+             if (!fieldValid->isKnown() || fieldValid->isUninitialized()) {
+                return new SymbolicBool(ScalarValue::ValueState::NotConstant);
```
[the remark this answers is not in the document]
OleksandrLetychevskyiIntel: I fixed it

mbudiu-vmw (approved):
Can you please squash your commits into just one so we can merge?

PR #3064, opened 2022-02-01
Add types for tables keys after switchElimination

VolodymyrPeschanenkoIntel:
This small changes fix integration bug between `p4c-bm2-ss` and `behavioral-model`.
The `EliminateSwitch` [here](https://github.com/p4lang/p4c/blob/bbd35694c379bca47acbd8da466d9c386afe6673/backends/bmv2/simple_switch/midend.cpp#L121)  doesn't add type for new tables keys.  It leads to the situation  where each new table key will have `Type_Unknown` and zero width [here](https://github.com/p4lang/p4c/blob/bbd35694c379bca47acbd8da466d9c386afe6673/backends/bmv2/common/control.h#L447). 
Next, if a user tries to run `behavioral-model` with a resulting json then the sizes of expected and real key are not equal [here](https://github.com/p4lang/behavioral-model/blob/3baf709e6ef1dea6097c2f230cb6b8bfc8dbe0d3/src/bm_sim/match_units.cpp#L576). 

You need to run the following commands to obtain this situation:
1) `p4c/build/p4c-bm2-ss  -o ./build/11/invalid-hdr-warnings3.json ./p4c/testdata/p4_16_samples/invalid-hdr-warnings3.p4`
2) 
```
behavioral-model/targets/simple_switch/.libs/simple_switch  
 --log-file switch.log --log-flush --use-files 0 --thrift-port 10012 --device-id 922 -i 0@pcap0 "p4c/build/11/invalid-hdr-warnings3.json"
```


backends/bmv2/simple_switch/midend.cpp
```diff
@@ -119,6 +119,7 @@ SimpleSwitchMidEnd::SimpleSwitchMidEnd(CompilerOptions& options, std::ostream* o
             new P4::CompileTimeOperations(),
             new P4::TableHit(&refMap, &typeMap),
             new P4::EliminateSwitch(&refMap, &typeMap),
+            new P4::TypeChecking(&refMap, &typeMap, true),
```
mbudiu-vmw, line 122: You are fixing a symptom and not the actual problem.
If a pass needs to make sure that types are known and fresh it should run typechecking itself.
The next pass, RemoveLeftSlices does this (but without the 'true' flag).
Perhaps this has to be done just before the backend. The types you set here are not guaranteed to be accurate by the time you reach the backend anyway.
Moreover, you should add a check in the code generating part of the backend to give an error if a type is Type_Unknown instead of generating incorrect code.

mbudiu-vmw (changes requested):

backends/bmv2/simple_switch/midend.cpp
VolodymyrPeschanenkoIntel: I moved down TypeChecking as much as possible.

backends/bmv2/simple_switch/midend.cpp
```diff
@@ -125,6 +125,7 @@ SimpleSwitchMidEnd::SimpleSwitchMidEnd(CompilerOptions& options, std::ostream* o
             isv1 ? new P4::RemoveUnusedActionParameters(&refMap) : nullptr,
             new P4::TypeChecking(&refMap, &typeMap),
             options.loopsUnrolling ? new P4::ParsersUnroll(true, &refMap, &typeMap) : nullptr,
+            new P4::TypeChecking(&refMap, &typeMap, true),
```
mbudiu-vmw, line 128: This is not late enough, there are many passes run in the backend that could potentially modify this.
Look for example at simpleSwitch.cpp.

backends/bmv2/common/control.h
```diff
@@ -444,6 +444,8 @@ class ControlConverter : public Inspector {
             for (auto k : keyset->components) {
                 auto key = new Util::JsonObject();
                 auto tableKey = table->getKey()->keyElements.at(keyIndex);
+                BUG_CHECK(!tableKey->expression->type->is<IR::Type_Unknown>(),
```
mbudiu-vmw, line 447: Why do you use the `type` field instead of looking up the type in the typeMap? That one is most certainly accurate.

mbudiu-vmw (changes requested):

backends/bmv2/common/control.h
```diff
@@ -17,274 +17,283 @@ limitations under the License.
...
 
-#include "ir/ir.h"
-#include "lib/json.h"
 #include "controlFlowGraph.h"
-#include "frontends/p4/coreLibrary.h"
-#include "frontends/p4/typeMap.h"
-#include "frontends/p4/typeChecking/typeChecker.h"
-#include "frontends/common/resolveReferences/referenceMap.h"
-#include "midend/convertEnums.h"
 #include "expression.h"
```
fruffy, line 21: Are all these changes caused by autoformatting?

backends/bmv2/simple_switch/simpleSwitch.cpp
```diff
@@ -1173,6 +1173,7 @@ SimpleSwitchBackend::convert(const IR::ToplevelBlock* tlb) {
         // This is important later for conversion to JSON.
         new P4::ClonePathExpressions(),
         new P4::ClearTypeMap(typeMap),
+        new P4::TypeChecking(refMap, typeMap, true),
```
fruffy, line 1176: This clears the typeMap and immediately recomputes it? Does this make sense?

VolodymyrPeschanenkoIntel: You are right, it should be removed.

VolodymyrPeschanenkoIntel: You are right, it should be removed.

mbudiu-vmw: I don't know about this case, but whenever we rewrite the IR causing some  nodes to have a different type this is necessary.

VolodymyrPeschanenkoIntel: I think that ClearTypeMap call is needed. P4c have a bug without it for` bmv2/testdata/p4_16_samples/pvs-struct-3-bmv2.p4`: 
`Compiler Bug: /home/vpeschax/p4c/testdata/p4_16_samples/pvs-struct-3-bmv2.p4(17): Cannot find declaration for ternary @match(ternary) bit<16> field1; `

fruffy:
Can you change the invalid header warnings test such that the behavioral model is actually executed on it? Then we can see the fixes. 
@mbudiu-vmw Any other comments?

mbudiu-vmw (approved):

PR #3067, opened 2022-02-02
Remove unnecessary check in conversion of log_msg to a JSON

VolodymyrPeschanenkoIntel:
These small changes fix integration bug between `behavioral-model` and `p4c-bm2-ss`. 
The p4c-bm2-ss generated the error if second argument is not `Type_Bits` and `Type_Boolean`.
The `ExternConverter_log_msg::convertExternFunction` makes conversion to a json.  I think that in second argument could be any type of expression.

To repeat this bug, please, run the following command:
`./p4c-bm2-ss -o ./11/issue2201-bmv2.json ../testdata/p4_16_samples/issue2201-bmv2.p4`

The `behavioral-model` accepts created json without any errors.


jfingerh:
Look for the earlier commit that added these checks.  They are there specifically because at least at the time they were added, BMv2 did NOT support these types in its logging action.  Are you saying that has changed since that time?  In what behavioral-model commit was the logging support generalized to other types?

jfingerh:
Just because bmv2 can READ the BMv2 JSON file, does not mean that it can execute the log action without errors.  Test actually _executing_ the log message action by sending through appropriate packets to find out whether BMv2 gives an error then.

VolodymyrPeschanenkoIntel:
@jfingerh, Is it possible to create a issue for behavioral-model to support `log_msg` with other types? It looks like the v1model supports [log_msg](https://github.com/p4lang/p4c/blob/5090df8679d65bedb73241346f59509845378dc3/p4include/v1model.p4#L711) with any type?

jfingerh:
It is trivial to create an issue for anything you want.  Whether it gets implemented or not is the bigger question, and by whom :-)

mbudiu-vmw:
It may be easier to implement this entirely in the compiler.

VolodymyrPeschanenkoLitSoft:
> It may be easier to implement this entirely in the compiler.

@mbudiu-vmw, Do you have any suggestions for that? 
Now this example crashed [here](https://github.com/p4lang/behavioral-model/blob/3baf709e6ef1dea6097c2f230cb6b8bfc8dbe0d3/include/bm/bm_sim/actions.h#L342), because tag == ActionParam::HEADER

mbudiu-vmw:
We already have a pass doing this "flattenInterfaceStructs", but it doesn't touch values that are passed as arguments to functions. So a pass like this should be written for the arguments to log_msg, expanding one log_msg into many.

VolodymyrPeschanenkoLitSoft:
> We already have a pass doing this "flattenInterfaceStructs", but it doesn't touch values that are passed as arguments to functions. So a pass like this should be written for the arguments to log_msg, expanding one log_msg into many.

Great!!!! I will continue with that on this PR, thanks.

VolodymyrPeschanenkoIntel:
@jfingerh , @mbudiu-vmw PR is ready for review, but we need to clarify one question: what we need to do with errors inside log_msg? We have some pass which could replace them into bit constants. Do we need to replace them in the SimpleSwitch?

jfingerh:
I see these example v1model P4 program lines you are referring to in your links:
```
        log_msg("hdr.ethernet={}", {hdr.ethernet});
```
and this one:
```
        log_msg("stdmeta={}", {stdmeta});
```
I still do not know what the errors are that you are asking about.

If you want to enhance the v1model implementation so that those log_msg() calls are supported, by printing out all of the fields of the stdmeta struct, for example, or the valid true/false of a header and the value of all of its fields, that sounds like a great idea.

If you don't want to make changes so that those are supported, then I recommend that whichever cases you do not support should give a compile time error when running p4c.

If you have a question not answered by the suggestions above, please ask a more specific question.

VolodymyrPeschanenkoIntel:
I try to enhance the v1model implementation. The result of this transformation could be accepted by behavioral-model. Speaking true, I forgot to extract valid filed for header, thanks.  

Without this transformation it gives the error:
// Assertion 'Default switch case should not be reachable' failed, file '../../include/bm/bm_sim/actions.h' line '334'.

jfingerh:
Maybe I am assuming something here that needs to be said explicitly:

If you make enhancements in the p4c v1model back end, either:

(a) with those enhancements, the simple_switch/ simple_switch_grpc processes should be not only able to read the BMv2 JSON file without error, it should also be able to execute those log_msg calls without error.

(b) or, there should be corresponding enhancements to simple_switch/simple_switch_grpc code made first, such that (a) is true.

I am perfectly happy if you do not make any changes to simple_switch/simple_switch_grpc, as long as (a) is true.  If you violate (a), then you are introducing new cases where the compiler gives no errors, but simple_switch/simple_switch_grpc will give a run-time error when trying to execute the log_msg calls.

jfingerh:
My apologies, I might understand your original question now.

Are you asking: "What if someone tries to do a log_msg call on an expression with type `error`?  Or a struct that contains a field with type `error`?"

If that is your question, then my answer would be one of:

(a) Give a compile-time error that log_msg calls are not supported on values with type `error`

or

(b) figure out a useful string representation to print for values of type `error`, and make the p4c v1model back end print the appropriate string representation somehow.

Approach (a) sounds easier to me for now.  I do not have specific suggestions for how to implement (b).

VolodymyrPeschanenkoIntel:
Ok, I will replace errors by string. For example, error.PacketTooShort => "PacketTooShort", thanks

VolodymyrPeschanenkoIntel:
> If you do that, note that a P4 developer can define new constant values with type `error` in their P4 program, and it would be nice to support printing those names, too.

Does P4 have such examples in p4_16_samples folder?

jfingerh:
Probably several, but this is the first one that I found: https://github.com/p4lang/p4c/blob/main/testdata/p4_16_samples/verify-bmv2.p4

mbudiu-vmw:
Also note that midend passes may remove enums and errors, so you would have to do this before these passes are run.

jfingerh:
Basically, if it gets to be too much work to print out names for these things, we understand :-)  The next best thing would be to print out the numeric values for them, and we can add some documentation to the log_msg call explaining to developers where to determine the mapping between the names and numeric values.

mbudiu-vmw (changes requested):
Which test exercises this change?
I would expect that the test used to fail and now it doesn't.
Does this require changes to bmv2?

VolodymyrPeschanenkoIntel:
I think that the best way to deal with errors is to replace them by constants like it was done [here](https://github.com/p4lang/p4c/blob/289431c0de84543adf26372fe6c157bdfcf76733/backends/dpdk/midend.cpp#L223).

All enums eliminated [here](https://github.com/p4lang/p4c/blob/289431c0de84543adf26372fe6c157bdfcf76733/backends/bmv2/simple_switch/midend.cpp#L78). So, we can ignore them.

To repeat this problem, please, run the following command:
`./p4c-bm2-ss -o ./11/issue2201-bmv2.json ../testdata/p4_16_samples/issue2201-bmv2.p4`

For that case we make log_msg flatten and replace Type_Error. After that p4c-bm2-ss successfully finished and result can be used to run behavioral-model without any errors.

mbudiu-vmw:
Looks like bmv2 crashes for some of these tests.

VolodymyrPeschanenkoIntel:
It looks like behavioral-models support Type_Error for verify and log_msg.

midend/flattenLogMsg.cpp
```diff
@@ -0,0 +1,198 @@
+#include "midend/flattenLogMsg.h"
+#include "flattenLogMsg.h"
+
+namespace P4 {
+
+const IR::Type_StructLike* FindTypeInLogMsgForReplace::hasStructInParameter(
+        const IR::MethodCallStatement* methodCallStatement) {
+    if (auto* path = methodCallStatement->methodCall->method->to<IR::PathExpression>()) {
+        if (path->path->name.name == "log_msg") {
+            auto* param1 = methodCallStatement->methodCall->arguments->at(0)->expression;
```
mbudiu-vmw, line 10: you should first check that there exist 2 arguments. Users are free to declare a log_msg function with a different signature. In that case you should give an error; your approach will crash the compiler if the signature does not have at least 2 arguments.

midend/flattenLogMsg.h
```diff
@@ -0,0 +1,92 @@
...
+#include "frontends/p4/typeChecking/typeChecker.h"
+
+namespace P4 {
+
+using TypeLogMsgParams = std::pair<IR::IndexedVector<IR::NamedExpression>, std::string>;
+
+/*
+   Find types in log_msg function for replace.
+*/
+class FindTypeInLogMsgForReplace : public Inspector {
```
mbudiu-vmw, line 14: I would call this "FindTypesInLogMsgInvocationToReplace"

midend/flattenLogMsg.h
```diff
@@ -0,0 +1,92 @@
...
+   P4::TypeMap* typeMap;
+   ordered_map<cstring, const IR::Type_StructLike*> replacement;
+   ordered_map<unsigned, const IR::MethodCallStatement* > logMsgReplacament;
```
mbudiu-vmw, line 17: can you document this structure? What is the key?

mbudiu-vmw, line 17: typo: replacement, not replacament

midend/flattenLogMsg.h
```diff
@@ -0,0 +1,92 @@
...
+    void createReplacement(const IR::Type_StructLike* type);
+    const IR::MethodCallStatement* prepareLogMsgStatement(
+         const IR::MethodCallStatement* methodCallStatement);
+    const IR::Type_StructLike* getReplacement(const cstring name) const {
+         return ::get(replacement, name); }
+    const IR::MethodCallStatement* getReplacementMethodCall(unsigned id) const {
+       return ::get(logMsgReplacament, id);
+    }
+    bool empty() const { return replacement.empty(); }
+   const IR::Type_StructLike* hasStructInParameter(
```
mbudiu-vmw, line 35: inconsistent indentation

midend/flattenLogMsg.cpp
```diff
@@ -0,0 +1,198 @@
...
+
+const IR::Node* ReplaceLogMsg::preorder(IR::P4Program* program) {
+    if (findTypeInLogMsgForReplace->empty()) {
+        // nothing to do
+        prune();
+    }
+    return program;
+}
+
+const IR::Node* ReplaceLogMsg::postorder(IR::Type_Struct* typeStruct) {
```
mbudiu-vmw, line 178: why do you replace the type? If this type is used in other places besides the logmsg call it should remain in the program. So you should append the new type, not replace the old one.

midend/flattenLogMsg.cpp
```diff
@@ -0,0 +1,198 @@
...
+    if (repl != nullptr) {
+        LOG3("Replace " << typeStruct << " with " << repl);
+        BUG_CHECK(repl->is<IR::Type_Struct>(), "%1% not a struct", typeStruct);
+        return repl;
+    }
+    return typeStruct;
+}
+
+const IR::Node* ReplaceLogMsg::postorder(IR::MethodCallStatement* methodCallStatement) {
+    if (auto* newMethod =
```
mbudiu-vmw, line 191: I would index with getOriginal(), not with clone_id.

midend/flattenLogMsg.h
mbudiu-vmw, line 35: it seems to be enough to return a boolean here, you don't use the type anywhere.

midend/flattenLogMsg.cpp
```diff
@@ -0,0 +1,198 @@
...
+            auto* param2 = methodCallStatement->methodCall->arguments->at(1)->expression;
+            if (!param1->is<IR::StringLiteral>() || !param2->is<IR::StructExpression>()) {
+                return nullptr;
+            }
+            auto* type =
+                typeMap->getTypeType(methodCallStatement->methodCall->typeArguments->at(0), true);
+            if (auto* typeStruct = type->to<IR::Type_StructLike>()) {
+                for (auto field : typeStruct->fields) {
+                    if (field->type->is<IR::Type_StructLike>()) {
+                        return typeStruct;
```
mbudiu-vmw, line 20: change this to 'true'

midend/flattenLogMsg.cpp
```diff
@@ -0,0 +1,198 @@
...
+
+const IR::MethodCallStatement* FindTypeInLogMsgForReplace::prepareLogMsgStatement(
+        const IR::MethodCallStatement* methodCallStatement) {
+    if (auto* newMethod = getReplacementMethodCall(methodCallStatement->id)) {
+        return newMethod;
+    }
+    // Create new statement
+    auto* param1 = methodCallStatement->methodCall->arguments->at(0)->expression;
+    std::string strParam1 = param1->to<IR::StringLiteral>()->value.c_str();
+    auto* argType =
```
mbudiu-vmw, line 37: even here you have to check that the method has 1 type argument.

midend/flattenLogMsg.cpp
```diff
@@ -0,0 +1,198 @@
...
+        typeMap->getTypeType(methodCallStatement->methodCall->typeArguments->at(0), true);
+    auto* structType = argType->to<IR::Type_StructLike>();
```
mbudiu-vmw, line 39: checkedTo?

midend/flattenLogMsg.cpp
```diff
@@ -0,0 +1,198 @@
...
+    index = 0;
+    auto exprVector =
+        unfoldStruct(methodCallStatement->methodCall->arguments->at(1)->expression, strParam1);
+    structType = generateNewStructType(structType, exprVector.first);
+    auto* newMethodCall = methodCallStatement->methodCall->clone();
```
mbudiu-vmw, line 44: alternatively you can let the type inference compute this argument after clearTypeMap, by invoking it with a suitable 'true' flag to infer missing types.

midend/flattenLogMsg.cpp
```diff
@@ -0,0 +1,198 @@
...
+    auto* newString = newArgument->expression->to<IR::StringLiteral>()->clone();
+    newString->value = exprVector.second;
+    newArgument->expression = newString;
+    newArguments->at(0) = newArgument;
+    newArgument = newMethodCall->arguments->at(1)->clone();
+    auto* oldType = newArgument->expression->to<IR::StructExpression>()->structType;
+    auto newStructExpression =new IR::StructExpression(oldType->srcInfo, structType,
+        structType->getP4Type()->template to<IR::Type_Name>(), exprVector.first);
+    newStructExpression->components = exprVector.first;
+    typeMap->setType(structType, structType);
```
mbudiu-vmw, line 60: why do you need to set these types? you will clear the type map immediately after this.

midend/flattenLogMsg.cpp
```diff
@@ -0,0 +1,198 @@
...
+    return IR::ID(ostr.str());
+}
+
+const IR::Type_StructLike* FindTypeInLogMsgForReplace::generateNewStructType(
+    const IR::Type_StructLike* structType, IR::IndexedVector<IR::NamedExpression> &v) {
+    IR::IndexedVector<IR::StructField> fields;
+    for (auto namedExpr : v) {
+        auto* structField = new IR::StructField(namedExpr->srcInfo, namedExpr->name,
+                                                namedExpr->expression->type);
+        typeMap->setType(structField, structField->type);
```
mbudiu-vmw, line 84: I am not sure you need to put these in the typeMap

midend/flattenLogMsg.cpp
```diff
@@ -0,0 +1,198 @@
...
+        unfoldStruct(methodCallStatement->methodCall->arguments->at(1)->expression, strParam1);
+    structType = generateNewStructType(structType, exprVector.first);
+    auto* newMethodCall = methodCallStatement->methodCall->clone();
+    auto* newTypeArguments = newMethodCall->typeArguments->clone();
+    newTypeArguments->clear();
+    newTypeArguments->push_back(structType);
+    newMethodCall->typeArguments = newTypeArguments;
+    auto* newArguments = newMethodCall->arguments->clone();
+    auto* newArgument = newMethodCall->arguments->at(0)->clone();
+    auto* newString = newArgument->expression->to<IR::StringLiteral>()->clone();
```
mbudiu-vmw, line 51: why clone this and not just create a new stringliteral?

midend/flattenLogMsg.cpp
```diff
@@ -0,0 +1,198 @@
...
+            return result;
+        }
+        result.first.push_back(new IR::NamedExpression(expr->srcInfo, newName(), expr));
+        result.second = curName + std::string(" : {}");
+        return result;
+    }
+    auto* structExpr = expr->to<IR::StructExpression>();
+    std::string strResult;
+    for (auto namedExpr : structExpr->components) {
+        size_t n = strParam.find("{}");
```
mbudiu-vmw, line 126: what happens if you don't find this substring?
should you give a warning?

mbudiu-vmw (changes requested):

midend/flattenLogMsg.cpp
VolodymyrPeschanenkoIntel: Could you, please, suggest a place where is I can to insert new type?

midend/flattenLogMsg.cpp
VolodymyrPeschanenkoIntel: This situation is impossible here, because this code executable if we have found corresponded MethodCallStatement for replacement in the map. The record in the map appears if log_smg has two arguments only.

midend/flattenLogMsg.cpp
VolodymyrPeschanenkoIntel: This type is used for creation of IR::StructExpression. The IR::StructExpression haven't  any constructors without IR::Type. Do I need to use IR::Type_Unknown instend of structType?

midend/flattenLogMsg.cpp
VolodymyrPeschanenkoIntel: I think that this type couldn't be used in other places, because it should be unique for each log_msg function and each log_msg function will be replace for new one.

midend/flattenLogMsg.cpp
mbudiu-vmw: Then you need to create the type, there isn't enough information for the typechecked to infer the type from the argument.

mbudiu-vmw (approved):
I will approve this, but perhaps you can make sure that removing the type will not result in an incorrect program before we merge.

midend/flattenLogMsg.cpp
VolodymyrPeschanenkoIntel: It should add other arguments in StructExpression to this {} like in your example below.

VolodymyrPeschanenkoIntel:
> I will approve this, but perhaps you can make sure that removing the type will not result in an incorrect program before we merge.

The p4c has tuple type in second argument for all examples in testdata. This code doesn't change the type S, it changes type for `{s,s}`.  In the program you have the type `Tuple(S,S)` for `{s,s}`, after it will be `Tuple(bit<8>, bit<8>)` and` log_msg("{}", {s,s})` will be translated into `log_msg("(f={}, f={})", {s.f,s.f})` and corresponded type will become Tuple(bit<8>, bit<8>) for `{s.f,s.f}`.

I removed generation of BUG see above, because of two variables s for one {} in log_msg.

PR #3069, opened 2022-02-03
Support resubmit and recirculate with no arguments in p4-14

### LINKED ISSUE #3068: FindRecirculated does not check number of operands before accessing them
The following code should probably first check if the given operand is actually there (before checking what it is):
https://github.com/p4lang/p4c/blob/5090df8679d65bedb73241346f59509845378dc3/frontends/p4/fromv1.0/converters.cpp#L640-L672
I am not sure if this should lead to an error or not (is clone, recirculate, resubmit allowed to be used without the field list provided?). For p4_14 at least it seems that the field list parameter should be optional.
In any case, having `resubmit()` in code leads to compiler crashing.

mbudiu-vmw:
Fixes #3068

### DIFF
--- a/testdata/p4_14_samples_outputs/copy_to_cpu.p4
+++ b/testdata/p4_14_samples_outputs/copy_to_cpu.p4
@@ -3,7 +3,8 @@
 #include <v1model.p4>
 
 enum bit<8> FieldLists {
-    copy_to_cpu_fields = 0
+    none = 0,
+    copy_to_cpu_fields = 1
 }
 
 struct intrinsic_metadata_t {
@@ -26,9 +27,9 @@ struct metadata {
 }
 
 struct headers {
-    @name(".cpu_header")
+    @name(".cpu_header") 
     cpu_header_t cpu_header;
-    @name(".ethernet")
+    @name(".ethernet") 
     ethernet_t   ethernet;
 }
 
@@ -106,3 +107,4 @@ control computeChecksum(inout headers hdr, inout metadata meta) {
 }
 
 V1Switch(ParserImpl(), verifyChecksum(), ingress(), egress(), computeChecksum(), DeparserImpl()) main;
+
--- a/testdata/p4_14_samples_outputs/switch_20160512/switch-frontend.p4
+++ b/testdata/p4_14_samples_outputs/switch_20160512/switch-frontend.p4
@@ -3,12 +3,13 @@
 #include <v1model.p4>
 
 enum bit<8> FieldLists {
-    i2e_mirror_info = 8w0,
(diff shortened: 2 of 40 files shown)


jfingerh (approved):
I have reviewed the changes in the expected outputs files, and the new test program and its expected output, and they all look good to me.  As is my usual, I have not attempted to review the changes in the C++ code, although it appears that the basic idea is that when converting from P4_14 to P4_16, a field list with a name like "none" is always created, and always represents an empty field list, which seems like a reasonable strategy.

testdata/p4_14_samples_outputs/switch_20160512/switch-frontend.p4
```diff
@@ -657,166 +658,166 @@ header vlan_tag_t {
...
+    @name(".vxlan_gpe_int_header") 
     vxlan_gpe_int_header_t                  vxlan_gpe_int_header;
-    @name(".int_val")
+    @name(".int_val") 
     int_value_t[24]                         int_val;
-    @name(".mpls")
+    @name(".mpls") 
     mpls_t[3]                               mpls;
-    @name(".vlan_tag_")
+    @name(".vlan_tag_")
```
rst0git, line 820: @mbudiu-vmw Do you know what is generating these trailing white-spaces?

mbudiu-vmw: Yes, the code in toP4 for annotations always adds a trailing space.
I don't know why it wasn't there before... the code has been there for 5 years.
But our diff testing tool is insensitive to whitespace, so these changes would not trigger test failures.

testdata/p4_14_samples_outputs/copy_to_cpu.p4
```diff
@@ -3,7 +3,8 @@
 #include <v1model.p4>
 
 enum bit<8> FieldLists {
-    copy_to_cpu_fields = 0
+    none = 0,
```
rst0git, line 6: Is adding this `none` field intentional?

rst0git: Thank you Andy!

rst0git (approved):
LGTM

testdata/p4_14_samples_outputs/copy_to_cpu.p4
mbudiu-vmw: Yes, that is correct, just to make sure that id 0 is not used by other field lists we allocate it to none and no fields will be tagged with none. So calling recirculate() will not recirculate any fields.

PR #3070, opened 2022-02-04
Handle implicit serenum casts in more circumstances

### LINKED ISSUE #3056: Implicit cast of serializable enum
In testing out the situations where a serializable enum can be cast implicitly to its underlying type (relates to p4lang/p4-spec#958), we find several cases where the implicit cast should be allowed, but not so in the p4c compiler. We agreed to add to the spec that "_For enums with an underlying type, it can be implicitly cast to its underlying type whenever appropriate, including but not limited to in shifts, concatenation, bit slicing indexes, header stack indexes as well as other unary and binary operations._" (p4lang/p4-spec#982) Given this decision, bugs need to be fixed in the p4c compiler.
```
enum bit<8> E {
  e1 = 1,
  e2 = 2
}

control compute(inout hdr h) {
    apply {
        E e = E.e1;
        bit<8> x1 = 8w0 << e;
        bit<8> x2 = -e; 
        bit<4> x3 = e[3:0];     
        bool x4 = 8w0 == e;
        bit<8> x5 = 8w0 | e;
        // bit<8> x6 = e << 3;     // p4c error: "<< left operand of shift must be a numeric type, not E"
        // bit<16> x7 = e ++ 8w0;  // p4c error: "Concatenation not defined on E and bit<8>"
        // bit<4> x8 = 8w0[4:E.e1];   // p4c error: "bit indexes must be compile-time constants"
    }
}
```
For `x6`, it seems to be a problem with the compiler since shifting an enum field is a valid example in the spec (in section 8.3).
For `x7`, it does not really make sense to prohibit it given that the compiler allows implicit casts in shift (example `x1`).
For `x8`, it still seems to be a problem with the compiler since enum field is compile-time known as stated in 17.1.

Overall, I think the implicit casts of serializable enum should probably be allowed everywhere, especially considering this sentence in the spec "...it is always safe to cast from an enum to its underlying fixed-width integer type..." (from section 8.3).

One more thing to note is that if I change from `e` to `E.e1`, more expressions failed in the p4c compiler
```
        // bit<16> x1 = 16w0 << E.e1;
        // bit<8> x2 = -E.e1; 
        // bit<4> x3 = E.e1[3:0];     
```
all with the same `error: E.e1: expected an integer value`, leaving only x4 and x5 still valid.
Again, I think it is more likely a problem with the compiler.

mbudiu-vmw:
Fixes #3056

### DIFF
--- a/frontends/common/constantFolding.cpp
+++ b/frontends/common/constantFolding.cpp
@@ -582,8 +582,16 @@ static bool overflowWidth(const IR::Node* node, int width) {
 const IR::Node* DoConstantFolding::postorder(IR::Slice* e) {
     const IR::Expression* msb = getConstant(e->e1);
     const IR::Expression* lsb = getConstant(e->e2);
-    if (msb == nullptr || lsb == nullptr) {
-        ::error(ErrorType::ERR_EXPECTED, "%1%: bit indices must be compile-time constants", e);
+    if (msb == nullptr) {
+        if (typesKnown)
+            ::error(ErrorType::ERR_EXPECTED,
+                    "%1%: slice indexes must be compile-time constants", e->e1);
+        return e;
+    }
+    if (lsb == nullptr) {
+        if (typesKnown)
+            ::error(ErrorType::ERR_EXPECTED,
+                    "%1%: slice indexes must be compile-time constants", e->e2);
         return e;
     }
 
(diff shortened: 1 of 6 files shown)


rst0git (approved):
LGTM

PR #3072, opened 2022-02-05
Fixes #3071

### LINKED ISSUE #3071: Unable to do or,and,sub when 2nd argument passed is const
I think there is bug in ordered_set at this line https://github.com/p4lang/p4c/blob/main/lib/ordered_set.h#L224

all three function should take const ref for 2nd parm.

kamleshbhalui:
Accept const qualified ref for 2nd argument in ordered_set
* In ordered_set operator overloads like |=, &=, -= should accept
2nd argument as const ref, it would then accept lvalue as well as rvalue as a argument.

### DIFF
--- a/lib/ordered_set.h
+++ b/lib/ordered_set.h
@@ -222,15 +222,15 @@ class ordered_set {
 };
 
 template<class T, class C1, class A1, class U> inline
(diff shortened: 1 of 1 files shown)


mbudiu-vmw (approved):

kamleshbhalui:
Fixes #3071

kamleshbhalui:
@mbudiu-vmw can please merge it?

PR #3073, opened 2022-02-06
Use ccache with Github actions

fruffy:
Playing around with ccache a little. Maybe it can speed up some of the builds. It has worked well for my own repositories.


Dockerfile
```diff
@@ -18,12 +18,15 @@ ARG VALIDATION=OFF
...
+# Toggle usage of the GNU Multiple Precision Arithmetic Library.
+ARG ENABLE_GMP=ON
 # No questions asked during package installation.
 ARG DEBIAN_FRONTEND=noninteractive
 # Whether to install dependencies required to run PTF-ebpf tests
 ARG INSTALL_PTF_EBPF_DEPENDENCIES=OFF
 
 # Delegate the build to tools/ci-build.
 COPY . /p4c/
+RUN --mount=type=bind,source=.ccache,target=/root/.ccache,rw  chmod u+x /p4c/tools/ci-build.sh && /p4c/tools/ci-build.sh
```
rst0git, line 30: @fruffy it might be better to add the `--mount` option to the `docker run` command instead because we also use this Dockerfile to build the [p4lang/p4c](https://hub.docker.com/r/p4lang/p4c) container image in Dockerhub.

fruffy: So the problem is that the `docker run` command is just running ctest. However, we need to map ccache correctly when building p4c. I am not sure how to untangle this yet.

rst0git: We could use something like the following:
```diff
-RUN chmod u+x tools/ci-build.sh && tools/ci-build.sh
+RUN (test "$IMAGE_TYPE" = "build" && \
+        chmod u+x tools/ci-build.sh && tools/ci-build.sh) || \
+    (test "$IMAGE_TYPE" = "test" && \
+      echo 'Test image ready')
```

and

```yaml
    - name: Build (Linux)
      run: |
        docker build -t p4c --build-arg IMAGE_TYPE=test .
        docker run --mount type=bind,source=/usr/lib/ccache,target=/root/.ccache -e ENABLE_UNIFIED_COMPILATION=ON -e BUILD_STATIC_RELEASE=ON -e ENABLE_GMP=ON p4c bash -c 'chmod u+x tools/ci-build.sh && tools/ci-build.sh'

```

fruffy:
This pull request is finally ready now. Properly configuring .ccache actually with actions reduces build times to about 1-2 minutes instead of 15-20. This reduces the test time of the slowest pull request by about 15-20 minutes, too. I had to add a separate dockerfile that exports the ccache from the image using docker buildkit. This seems to work. 

The next bottleneck is the amount of cores available for the tests. We have only 2.

Dockerfile
fruffy: Does the current solution work for you?

tools/install_fedora_deps.sh
```diff
@@ -69,7 +70,7 @@ pushd "${tmp_dir}"
 git clone --depth=1 https://github.com/p4lang/behavioral-model
```
fruffy, line 70: Do we really need to build the behavioral model for the fedora tests? Can't we just run the normal p4 tests like we do with MacOS?

Dockerfile
rst0git: Yes

export_ccache.dockerfile
```diff
@@ -0,0 +1,2 @@
+FROM scratch AS export-stage
+COPY --from=p4c /p4c/.ccache ./.ccache
```
rst0git, line 2: @fruffy  Could we replace this dockerfile with a shell script?
```bash
$ cat tools/export_ccache.sh 
#!/bin/bash
docker cp $(docker create --rm p4c):/p4c/.ccache .
```

Dockerfile
rst0git: Yes, I tested it locally and it seems to work.

Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM p4lang/behavioral-model:latest
+FROM p4lang/behavioral-model:latest as p4c
```
rst0git, line 1: It looks like `as p4c` is no longer necessary?

backends/ebpf/targets/xdp_target.py
```diff
@@ -0,0 +1 @@
+/storage/Projekte/p4_tv/modules/p4c/extensions/xdp/xdp_target.py
```
rst0git, line 1: added unintentionally?

tools/ci-build.sh
```diff
@@ -46,15 +46,18 @@ export P4C_PIP_PACKAGES="ipaddr \
...
-
 apt-get update
 apt-get install -y --no-install-recommends \
   ${P4C_DEPS} \
   ${P4C_EBPF_DEPS} \
   ${P4C_RUNTIME_DEPS} \
   git
 
+# TODO: Remove this rm -rf line once the ccache memcache config is removed.
+rm -rf /usr/local/etc/ccache.conf
```
rst0git, line 57: Is the `rm -rf` line intended to be removed before merging this pull request?

fruffy: No, we can only remove it once we have removed the corresponding ccache code in https://github.com/p4lang/third-party.

rst0git (approved):
LGTM

PR #3075, opened 2022-02-08
Solution when p4include may be within the path

vladyslav-dubina:
Solves `verify: declaration not found` error when p4include  within the path
This pull request addresses the problem when the path is not at the beginning.
For example launch `parser-unroll-test2.p4` test  with `-I ./p4include` argument

### DIFF
--- a/frontends/p4/unusedDeclarations.cpp
+++ b/frontends/p4/unusedDeclarations.cpp
@@ -172,6 +172,9 @@ bool RemoveUnusedDeclarations::isSystemFile(cstring file) {
     if (file.startsWith(p4includePath)) return true;
     // if the backend is invoked directly with '-I p4include'
     if (file.startsWith("p4include")) return true;
+    // if "-I ./p4include" is specified in the arguments
+    if (file.find("p4include")) return true;
+
     return false;
 }
 
--- a/frontends/p4/unusedDeclarations.cpp
+++ b/frontends/p4/unusedDeclarations.cpp
@@ -169,6 +169,9 @@ const IR::Node* RemoveUnusedDeclarations::preorder(IR::ParserState* state) {
 
 // Try to guess whether a file is a "system" file
 bool RemoveUnusedDeclarations::isSystemFile(cstring file) {
+    if (file.startsWith(p4includePath)) return true;
+    // If the backend is invoked directly with '-I p4include'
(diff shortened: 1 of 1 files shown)


frontends/p4/unusedDeclarations.cpp
```diff
@@ -172,6 +172,9 @@ bool RemoveUnusedDeclarations::isSystemFile(cstring file) {
     if (file.startsWith(p4includePath)) return true;
     // if the backend is invoked directly with '-I p4include'
     if (file.startsWith("p4include")) return true;
+    // In cases such as  "-I ./p4include", p4include may be within the path
+    if (file.find("p4include")) return true;
```
mbudiu-vmw, line 176: perhaps it's better to check for `file.endsWith("p4include")`

mbudiu-vmw (approved):

mbudiu-vmw (changes requested):
please see comment

frontends/p4/unusedDeclarations.cpp
vladyslav-dubina: We can add another check with endsWith. Since in our case endsWith will not be executed because p4include is not at the end, because the `file `variable contains something like `"path"/p4include/core.p4`

vladyslav-dubina: In order to get only the path from `file`, we will have to do cumbersome string processing to exclude the file name. 
Isn't it easier to leave only one check, for the presence of p4Include and remove `if (file.startsWith("p4include")) return true;` ?

mbudiu-vmw: That's what I said.

vladyslav-dubina: Okay then I'll make a change now

frontends/p4/unusedDeclarations.cpp
```diff
@@ -169,10 +169,6 @@ const IR::Node* RemoveUnusedDeclarations::preorder(IR::ParserState* state) {
 
 // Try to guess whether a file is a "system" file
 bool RemoveUnusedDeclarations::isSystemFile(cstring file) {
-    if (file.startsWith(p4includePath)) return true;
```
mbudiu-vmw, line 172: this one should not be removed

mbudiu-vmw (approved):

PR #3074, opened 2022-02-08
Updates for Ubuntu 20.04

antoninbas:
Some changes are required for the Docker build now that the base
image is Ubuntu 20.04.

PR #3076, opened 2022-02-08
Build improvements

antoninbas:
* use same Scapy version as PTF (2.4.5)
* set Boost_USE_STATIC_RUNTIME when looking for Boost libraries with cmake

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -90,24 +90,9 @@ if (BUILD_STATIC_RELEASE)
   set(BUILD_SHARED_LIBS OFF)
   set(CMAKE_FIND_LIBRARY_SUFFIXES .a)
   # Link Boost statically
+  # See https://cmake.org/cmake/help/latest/module/FindBoost.html for details
   set(Boost_USE_STATIC_LIBS ON)
-  set(Boost_USE_STATIC_RUNTIME ON)
-  # After updating to the Ubuntu 20.04 base OS and to Boost 1.71 for CI builds,
-  # the cmake invocation for static builds started failing, with no suitable
-  # iostream library being found:
-  # ***************************
-  # The following variants have been tried and rejected:
-  # * libboost_iostreams.so.1.71.0 (shared, Boost_USE_STATIC_LIBS=ON)
-  # * libboost_iostreams.a (shared runtime, Boost_USE_STATIC_RUNTIME=ON)
-  # ***************************
-  # It was suggested online to try building with -DBoost_NO_BOOST_CMAKE=ON,
-  # which ignores some cmake files shipping with Boost (from Boost 1.70 on).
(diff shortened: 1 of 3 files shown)


mbudiu-vmw:
These do not seem conditional on the Ubuntu version. Will they work on older versions too?

jafingerhut:
@mbudiu-vmw I am NOT making a unilateral decision on this, but there are only so many OS/toolchain version combinations that P4.org has resources to support.

Support for Ubuntu 16.04 seems like a good candidate for discontinuing support, given that since April 2021 you must pay for a commercial support license from Ubuntu to continue to get security updates for it.

Whether we officially continue to support Ubuntu 18.04 until April 2023, or not, is a separate decision that could reasonably be argued about, I think.  I think Antonin is volunteering N hours of his time to get us from Ubuntu 16.04 to 20.04.  Assuming that begins working smoothly in a week or so, it seems worth at that time deciding if someone wants to extend these scripts to support other OS/toolchain combinations as well.

I can put this on an item for discussion with the P4 TST, and talk to others within Intel to see if they might devote some paid employee time for these kinds of tasks, but in the end, this is either up to volunteer effort, or some company paying someone to create and maintain such things.

Oh, and by the way, thanks Antonin for those N hours, and I hope your manager likes you doing so! :-)

antoninbas:
> These do not seem conditional on the Ubuntu version. Will they work on older versions too?

Yes, they are not directly related. These just came up during review for https://github.com/p4lang/p4c/pull/3074.
In particular, the cmake changes for the Boost libraries work fine on my VM, which has Ubuntu 18.04 and version 1.65.1 of Boost.
I will edit the commit message when I merge.

mbudiu-vmw (approved):

jafingerhut:
Ugh.  And if all you were asking about, Mihai, was whether the title of this PR should be changed or not, then my sincere apologies for completely missing the point and addressing a question that you did not even ask.

mbudiu-vmw:
The issue is that this code is used by companies (like Barefoot) who may give the SDK to their customers, who may have constraints on the software they can install and when. I remember how they explained to me (a long time ago) why we could not use a recent version of gcc. But perhaps this is not such a difficult problem, they only need the software on a build machine.

jafingerhut:
@mbudiu-vmw Understood on the concerns of when to drop support for older versions.  I will check with Intel folks to see which set of Ubuntu versions they would like to see tested, and if it is more than only Ubuntu 20.04, ask which of them volunteers to help add the CI tests that cover them :-)

rst0git:
I would like to volunteer to extend the CI tests to run with Fedora in addition to Ubuntu. In contrast to Ubuntu, Fedora is a rolling-release distribution and we can install most dependencies with a package manager (`dnf`).

PR #3078, opened 2022-02-09
Run CI tests with Fedora

rst0git:
As a follow up on the discussion in https://github.com/p4lang/p4c/pull/3076, this pull request extends the CI to run P4C tests with Fedora in addition to Ubuntu.

### DIFF
--- a/.github/workflows/ci-test.yml
+++ b/.github/workflows/ci-test.yml
@@ -30,7 +30,7 @@ jobs:
       with:
         access_token: ${{ github.token }}
 
-    - name: Build (Linux)
+    - name: Build (Ubuntu Linux)
       run: |
         tools/start_ccache
         # To flush out issues with unified vs. non-unified builds,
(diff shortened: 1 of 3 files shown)


.github/workflows/ci-test.yml
```diff
@@ -40,11 +40,48 @@ jobs:
...
+  build-fedora-linux:
+    strategy:
+      fail-fast: false
+      matrix:
+        unified: [ON, OFF]
+        enable_gmp: [ON, OFF]
+        exclude:
+          - unified: OFF
+            enable_gmp: OFF
+    runs-on: ubuntu-latest
```
jafingerhut, line 58: I really do not know the format of these files, so this is just a visual "is this what you meant to do?" that could be perfectly correct.  Just surprised me to see build-fedora-linux running on ubuntu-latest

rst0git: Hi Andy, thank you for asking. I have to admit that it looks confusing and it would be good to add a comment here.

GitHub supports a very limited set of [virtual environments](https://github.com/actions/virtual-environments). The following means that the GitHub action would run inside a Fedora container that runs in a Ubuntu VM:

```yml
runs-on: ubuntu-latest
    container:
      image: registry.fedoraproject.org/fedora:latest
```

tools/install_fedora_deps.sh
```diff
@@ -0,0 +1,83 @@
+#!/bin/bash
```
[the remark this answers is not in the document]
rst0git: Thanks Mihai, I've added a link in the README file.

PR #3085, opened 2022-02-15
Release 1.2.2.1

rst0git:
The current versioning scheme of the P4 compiler corresponds to the P4 language specification. In particular, the current compiler version corresponds to v1.2.2 of P4_16 Language Specification.

This patch updates the P4 compiler version to the following format

	<major>.<minor>.<patchlevel>.<releasenumber>

Where `<releasenumber>` increments with each release of the compiler. The increment in version number allows to distribute updates for the P4 compiler using package management systems such as DPKG (Debian/Ubuntu) and RPM (Fedora/CentOS).

This change was discussed at the P4 open source developer days on November 23, 2021.


jafingerhut:
+1 from me to have something where we can have distinct version numbers for more frequently released versions of p4c (monthly, perhaps?)

mbudiu-vmw:
Can the incrementation of the number be automated so we make sure we always increment when merging something in main?

mbudiu-vmw (approved):

rst0git:
Thank you Andy and Mihai! I have also discussed creating a Fedora package for the P4 compiler with @trixirt and @adrianreber.
The git tag / release version in GitHub is used to download the sources to build a package. I suspect that most users would not need to update on every merged commit, but perhaps monthly update might be a good idea.

Issue #3092, opened 2022-02-23
make check-xdp failed


apple-zkk:
i have been install p4c and xdp,now i want to try p4c-xdp test,but when i input make check-xdp. it failed,i don't know why? Is the version not match?

rst0git:
What is error message do you see?

apple-zkk:
This is the error message when i input check-xdp, I only passed two tests when I typed this this afternoon, but most of them pass now

coco@ubuntu20:~/p4c/build$ make check-xdp
[100%] Running tests for tag xdp
Test project /home/coco/p4c/build
      Start  2: xdp/extensions/p4c-xdp/tests/xdp.p4
      Start  3: xdp/extensions/p4c-xdp/tests/xdp1.p4
 1/18 Test  #2: xdp/extensions/p4c-xdp/tests/xdp.p4 .....   Passed    0.22 sec
      Start  4: xdp/extensions/p4c-xdp/tests/xdp10.p4
 2/18 Test  #3: xdp/extensions/p4c-xdp/tests/xdp1.p4 ....   Passed    0.27 sec
      Start  5: xdp/extensions/p4c-xdp/tests/xdp11.p4
 3/18 Test  #4: xdp/extensions/p4c-xdp/tests/xdp10.p4 ...   Passed    0.30 sec
      Start  7: xdp/extensions/p4c-xdp/tests/xdp13.p4
 4/18 Test  #5: xdp/extensions/p4c-xdp/tests/xdp11.p4 ...   Passed    0.27 sec
      Start  8: xdp/extensions/p4c-xdp/tests/xdp14.p4
 5/18 Test  #7: xdp/extensions/p4c-xdp/tests/xdp13.p4 ...   Passed    0.29 sec
      Start  9: xdp/extensions/p4c-xdp/tests/xdp15.p4
 6/18 Test  #8: xdp/extensions/p4c-xdp/tests/xdp14.p4 ...   Passed    0.30 sec
      Start 10: xdp/extensions/p4c-xdp/tests/xdp16.p4
 7/18 Test  #9: xdp/extensions/p4c-xdp/tests/xdp15.p4 ...   Passed    0.24 sec
      Start 11: xdp/extensions/p4c-xdp/tests/xdp17.p4
 8/18 Test #10: xdp/extensions/p4c-xdp/tests/xdp16.p4 ...   Passed    0.31 sec
      Start 12: xdp/extensions/p4c-xdp/tests/xdp2.p4
[...]

PR #3093, opened 2022-02-23
Emit semicolon after method call

mbudiu-vmw:
Fixes https://github.com/p4lang/p4c/issues/3092


fruffy:
It's a little strange that this was not caught. I wonder what caused this failure. Does the XDP CI run regularly (daily or weekly)?

mbudiu-vmw:
I think that the xdp repo was using Travis, so it's stopped running tests altogether.
It was never updated to github actions.

fruffy (approved):
I can try to migrate to actions in the next couple of days. Should not be too much work.

PR #3094, opened 2022-02-23
Preserve type information in constant folding

fruffy:
Constant folding was missing some type information for bool literals, which are added with this pull request. Please let me know if I am missing some more. 

I am aware that there is no guarantee for these passes to preserve type information. However, ConstantFolding and StrengthReduction are also useful in isolation. This is because they can be applied to any well-typed IR::Expression construct to simplify it.

### DIFF
--- a/frontends/common/constantFolding.cpp
+++ b/frontends/common/constantFolding.cpp
@@ -160,7 +160,8 @@ const IR::Node* DoConstantFolding::postorder(IR::Declaration_Constant* d) {
             } else {
                 // Destination type is InfInt; we must "erase" the width of the source
                 if (!cst->type->is<IR::Type_InfInt>()) {
-                    init = new IR::Constant(cst->srcInfo, cst->value, cst->base);
+                    init = new IR::Constant(cst->srcInfo, new IR::Type_InfInt(), cst->value,
+                                            cst->base);
                 }
             }
         }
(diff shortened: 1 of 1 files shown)


mbudiu-vmw:
Currently almost no passes assume that the type field of an expression is correct.
This is not wrong, but it is not useful either.

fruffy:
We are developing an interpreter in which we are building up expressions that use the P4C IR. To simplify these expression we reuse some of the compiler passes, primarily StrengthReduction and ConstantFolding. They are quite useful since they handle a lot of cases and can be applied without needing a type or reference map. However, sometimes the results are missing type information, which the interpreter requires. By adding this type information we avoid running into this issue.

mbudiu-vmw:
This will only store type information in some literals, not even sure that in all of them.
You already know that all Bool literals have type bool, there's no choice there.
Other literals won't have type information, e.g., strings, enum, errors, etc.
Recall that we also have another symbolic interpreter for loop unrolling which does some of this work.

mbudiu-vmw (approved):

fruffy:
Yes, the problem was that well-typed expressions such as `IR::Equ` were reduced to a literal that had no type. The interpreter relies on `->type` to be correct since at that particular stage it just works with `IR::Expression` objects. Another way to solve this were to explicitly check the returned expression from `ConstantFolding` and manually assign types. But that fix would be cumbersome since we would need to do this for all sorts of expressions. It was easier to just edit `ConstantFolding`. 

For `ConstantFolding` I was only able to identify these literals without a type, but there might be more.

PR #3096, opened 2022-02-24
dpdk: Eliminate unused metadata fields

kamleshbhalui:
The PSA and PNA architecture require a lot of architecture-specific meta-data fields, but most of them are not actually used by a typical P4 program.

If an architecture-specific meta-data field is defined in the P4 because it is required by the architecture (e.g. PSA or PNA), but this field is not used at all by the P4 program, then this field should not be carried into the output .spec file generated by the P4C-DPDK compiler.

### DIFF
--- a/backends/dpdk/dpdkAsmOpt.cpp
+++ b/backends/dpdk/dpdkAsmOpt.cpp
@@ -163,4 +163,35 @@ const IR::IndexedVector<IR::DpdkAsmStatement> *RemoveLabelAfterLabel::removeLabe
     }
     return new_l;
 }
+
+const IR::Node* RemoveUnusedMetadataFields::preorder(IR::DpdkAsmProgram *p) {
+    IR::IndexedVector<IR::DpdkStructType> usedStruct;
+    bool isMetadataStruct = false;
+    for (auto st : p->structType) {
+        if (!isMetadataStruct) {
+            for (auto anno : st->annotations->annotations) {
+                if (anno->name == "__metadata__") {
+                    isMetadataStruct = true;
+                    IR::IndexedVector<IR::StructField> usedMetadataFields;
+                    for (auto field : st->fields) {
+                        if (used_fields.count(field->name.name)) {
+                            usedMetadataFields.push_back(field);
+                        }
+                    }
+                    auto newSt = new IR::DpdkStructType(st->srcInfo, st->name,
+                                                   st->annotations, usedMetadataFields);
+                    usedStruct.push_back(newSt);
+                }
+            }
+            if (!isMetadataStruct) {
+                usedStruct.push_back(st);
+            }
+        } else {
+            usedStruct.push_back(st);
+        }
+    }
+    p->structType = usedStruct;
+    return p;
+}
+
 }  // namespace DPDK
--- a/backends/dpdk/dpdkAsmOpt.h
+++ b/backends/dpdk/dpdkAsmOpt.h
@@ -135,6 +135,30 @@ class RemoveLabelAfterLabel : public Transform {
     }
 };
 
+
+// This pass Collects all metadata struct member used in program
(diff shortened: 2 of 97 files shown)


backends/dpdk/dpdkAsmOpt.cpp
```diff
@@ -163,4 +163,40 @@ const IR::IndexedVector<IR::DpdkAsmStatement> *RemoveLabelAfterLabel::removeLabe
     }
     return new_l;
 }
+
+const IR::Node* RemoveUnusedMetadataFields::preorder(IR::DpdkAsmProgram *p) {
+    auto usedFieldsCollector = new CollectUsedMetadataField();
+    usedFieldsCollector->setCalledBy(this);
+    p->apply(*usedFieldsCollector);
```
mbudiu-vmw, line 170: you can probably call prune() too after this, this visitor does not need to traverse p.
This pattern of allocating the visitor in the preorder makes the program just a bit harder to read, why not just put CollectUsedMetadataFields in the list of backend passes and pass a reference to it to RemoveUnusedMetadataFields (or a reference to the used_fields structure)?
Frankly, it's not a big difference, it up to you, but the other pattern is the one used almost everywhere. We do allocate visitors locally too, but that's usually when we want to apply them to inner nodes.

backends/dpdk/dpdkAsmOpt.h
```diff
@@ -135,6 +135,27 @@ class RemoveLabelAfterLabel : public Transform {
     }
 };
 
+
+// This pass Collects all metadata struct member used in program
+class CollectUsedMetadataField : public Inspector {
```
mbudiu-vmw, line 140: Alternatively, if this pass is used only within dpdkAsmOpt.cpp, it does not need to have a declaration in the header, it can be local to the file (within an empty namespace to avoid name clashes).

backends/dpdk/dpdkAsmOpt.h
```diff
@@ -135,6 +135,27 @@ class RemoveLabelAfterLabel : public Transform {
...
+    ordered_set<cstring> used_fields;
+  public:
+    bool preorder(const IR::Member *m) override {
```
mbudiu-vmw, line 143: This is not necessarily a field of the metadata structure, but I guess it will include all the fields. So you may keep some extra fields here if they happen to show up in other structures.

mbudiu-vmw (approved):

backends/dpdk/dpdkAsmOpt.cpp
kamleshbhalui: Incorporated the suggestions.

backends/dpdk/dpdkAsmOpt.h
kamleshbhalui: Now it's used in backend.cpp

backends/dpdk/dpdkAsmOpt.h
kamleshbhalui: Now only insert if it's metadata's member.

backends/dpdk/dpdkAsmOpt.cpp
```diff
@@ -163,4 +163,35 @@ const IR::IndexedVector<IR::DpdkAsmStatement> *RemoveLabelAfterLabel::removeLabe
     }
     return new_l;
 }
+
+const IR::Node* RemoveUnusedMetadataFields::preorder(IR::DpdkAsmProgram *p) {
+    IR::IndexedVector<IR::DpdkStructType> usedStruct;
+    bool isMetadataStruct = false;
```
mbudiu-vmw, line 169: I would call this "foundMetadataStruct".
Perhaps you can change it in a future PR.

mbudiu-vmw (approved):

PR #3097, opened 2022-02-24
Add debug messages for generated eBPF programs

tatry:
This PR adds `--trace` flag to the p4c-ebpf compiler. With the option enabled, p4c-ebpf compiler will generate additional debug logs that will be printed by eBPF program during packet processing. These logs might be helpful when debugging compiler, P4 programs or some tools. This option should only be used for troubleshooting as it decreases performance.

With this option disabled (which is by default) performance should not be affected because debug messages are not emitted into C code.

To see debug messages from program read from /sys/kernel/debug/tracing/trace_pipe file or use `bpftool prog tracelog`.

### DIFF
--- a/backends/ebpf/ebpfOptions.cpp
+++ b/backends/ebpf/ebpfOptions.cpp
@@ -23,4 +23,7 @@ EbpfOptions::EbpfOptions() {
         registerOption("--emit-externs", nullptr,
                 [this](const char*) { emitExterns = true; return true; },
                 "[ebpf back-end] Allow for user-provided implementation of extern functions.");
+        registerOption("--trace", nullptr,
+                [this](const char*) { emitTraceMessages = true; return true; },
+                "Generate tracing messages of packet processing");
 }
--- a/backends/ebpf/target.cpp
+++ b/backends/ebpf/target.cpp
@@ -20,6 +20,21 @@ limitations under the License.
 namespace EBPF {
 
 
+void Target::emitPreamble(Util::SourceCodeBuilder* builder) const {
+    (void) builder;
+}
+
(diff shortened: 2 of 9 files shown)


backends/ebpf/ebpfOptions.cpp
```diff
@@ -23,4 +23,7 @@ EbpfOptions::EbpfOptions() {
         registerOption("--emit-externs", nullptr,
                 [this](const char*) { emitExterns = true; return true; },
                 "[ebpf back-end] Allow for user-provided implementation of extern functions.");
+        registerOption("--trace", nullptr,
```
mbudiu-vmw, line 26: another alternative is to always emit them but to guard them with a macro `#if TRACE` which can be enabled when compiling the resulting C. But this is fine.

backends/ebpf/target.cpp
```diff
@@ -20,6 +20,21 @@ limitations under the License.
 namespace EBPF {
 
 
+void Target::emitPreamble(Util::SourceCodeBuilder* builder) const {
+    (void) builder;
+}
+
+void Target::emitTraceMessage(Util::SourceCodeBuilder* builder, const char* format,
```
mbudiu-vmw, line 27: you can perhaps document this function explaining what the arguments represent.
Also, you can probably give (in comments) an example where you use runtime arguments; there are such examples in the code, but perhaps here they would be easier to find.

mbudiu-vmw (approved):
Very nice, thank you.
One other thing you can do that could be useful is to add support for the P4 log_msg function in the same way.

mbudiu-vmw:
Let me know if I can merge this.

tatry:
I've added documentation for `emitTraceMessage`.

@mbudiu-vmw could you merge this PR, please?

PR #3100, opened 2022-02-27
Add correct and error test cases for structure-values expressions

jafingerhut:
I did not find any behavior that I would consider a bug in writing these tests, but I also did not see any existing tests that exercise some of these compile-time error catching cases in the p4_16_errors_outputs directory, so it seems they are not exercised by any existing test cases before these.


rst0git (approved):
LGTM

PR #3104, opened 2022-03-02
Forbid egress pipeline in dpdk by default

kamleshbhalui:
As there are no egress pipeline use case, currently forbid it by default in dpdk.

### DIFF
--- a/backends/dpdk/dpdkProgram.cpp
+++ b/backends/dpdk/dpdkProgram.cpp
@@ -113,9 +113,10 @@ const IR::DpdkAsmProgram *ConvertToDpdkProgram::create(IR::P4Program *prog) {
     for (auto kv : structure->parsers) {
         if (kv.first == "IngressParser")
             kv.second->apply(*ingress_parser_converter);
-        else if (kv.first == "EgressParser")
-            kv.second->apply(*egress_parser_converter);
-        else if (kv.first == "MainParserT")
+        else if (kv.first == "EgressParser") {
+            if (options.enableEgress)
+                kv.second->apply(*egress_parser_converter);
+        } else if (kv.first == "MainParserT")
             kv.second->apply(*ingress_parser_converter);
         else
             BUG("Unknown parser %s", kv.second->name);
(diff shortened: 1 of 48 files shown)


backends/dpdk/dpdkProgram.cpp
```diff
@@ -113,9 +113,10 @@ const IR::DpdkAsmProgram *ConvertToDpdkProgram::create(IR::P4Program *prog) {
     for (auto kv : structure->parsers) {
         if (kv.first == "IngressParser")
             kv.second->apply(*ingress_parser_converter);
-        else if (kv.first == "EgressParser")
-            kv.second->apply(*egress_parser_converter);
-        else if (kv.first == "MainParserT")
+        else if (kv.first == "EgressParser") {
```
mbudiu-vmw, line 116: this is a bit strange.
First, because you compile for PSA, you should always have an egress.
Second, by giving a flag you are causing a large part of the P4 program to be ignored. I would at least check that the egress programs are "empty" and give a warning. Alternatively, you could dispense with the flag and not generate any code if all the egress components are empty. In other words, infer the value of this flag from the source program.

kamleshbhalui: Currently, code generated for ingress pipeline and egress pipeline are in single output file which is not right, because two should be independent.
Real future job is to generate two output file one for egress pipeline and other for egress pipeline from a single p4 program which has non empty ingress and egress block. Until this done we generate code for ingress only even when egress are present, but once above task done, we will get rid of this flag, that's why the flag is not exposed to user, it's hidden.

mbudiu-vmw: it does not look very hidden to me.
Anyway, if this is temporary this should also be documented.

mbudiu-vmw (approved):

kamleshbhalui:
please merge this

mbudiu-vmw:
You have some build failures

kamleshbhalui:
Failure does not seem to related
Error: Failed to download metadata for repo 'updates-modular': Cannot prepare internal mirrorlist: Status code: 503 for https://mirrors.fedoraproject.org/metalink?repo=updates-released-modular-f35&arch=x86_64 (IP: 152.19.134.198)
can we restart build?

mbudiu-vmw:
I will take a chance

PR #3106, opened 2022-03-02
loop_revisit method for dealing with recursive loops in the IR

ChrisDodd:
- by default will BUG("IR loop detected") (the old behavior), but allows
  writing visitor passes that can override and deal with it.

This is an experimental infrastructure extension that is most useful with the ResolutionContext/apply with context param stuff added awhile back -- using that code, a visitor can look up symbols and recursively visit the definition of the symbol, but doing so may lead to loops in the traversal.  This change allows dealing with those resulting loops rather than just exiting with a BUG (though that is still the default behavior) by overriding the loop_revisit method in the Visitor subclass.


mbudiu-vmw:
Even DAGs cause difficulties today, allowing loops will make reasoning about the flow of control much harder.

mbudiu-vmw (approved):

PR #3109, opened 2022-03-04
DPDK: Fix missing action definitions in spec file

usha1830:
DPDK backend eliminates unused actions as an optimization.

However, the names in table action list and the actual action names retrieved using toString() are sometimes different and thus some actions are eliminated even though they are used.

This PR fixes the same by using the .name field instead of toString(). 

Now the internal name will be used in all the places. NoAction is handled in special way because in that case unnecessarily multiple copies of NoAction will be created.


backends/dpdk/dpdkProgram.cpp
```diff
@@ -508,7 +508,10 @@ bool ConvertToDpdkControl::preorder(const IR::P4Action *a) {
     for (auto i : helper->get_instr())
         stmt_list->push_back(i);
 
-    auto action = new IR::DpdkAction(*stmt_list, a->name, *a->parameters);
+    auto actName = a->name.name;
+    if (a->name.toString() == "NoAction")
```
mbudiu-vmw, line 512: toString() or name?

mbudiu-vmw (changes requested):

backends/dpdk/dpdkProgram.cpp
usha1830: Here it is toString() only. 
 .name would have the name with **_declid** suffix. like NoAction_1,NoAction_2 etc

mbudiu-vmw (approved):
In general I prefer originalName to toString().
toString() can change. It is for human consumption.

usha1830:
> In general I prefer originalName to toString(). toString() can change. It is for human consumption.
 Updated to use originalName.

usha1830:
@mbudiu-vmw Please merge if changes look fine now.

PR #3115, opened 2022-03-08
DPDK: Fix core dump while generating bfrt json for PNA programs with action selector

usha1830:
This PR fixes bfrt json generation for PNA programs using action selector table implementation.

Since PNA is derived from PSA and shares and supports all the externs that PSA supports, we create a common P4RuntimeArchHandler class for PSA and PNA and then derive PSA and PSA from the common class. This common class is a template class and PSA and PNA specializes it based on the template parameter "Arch". Earlier implementation was using the


mbudiu-vmw (approved):

usha1830:
Thanks @mbudiu-vmw . Could you please merge this?

PR #3116, opened 2022-03-08
Print the current node type name when checkedTo fails

fruffy:
This way one can immediately see the actual type of the node.


mbudiu-vmw (approved):
I would rephrase "x with type y is not a z"

PR #3117, opened 2022-03-09
from1.0: fix signed register type

vhavel:
Signed p4_14 Register is currently incorrectly translated into unsigned p4_16 Register due to missing signedness handling.


mbudiu-vmw (approved):

mbudiu-vmw:
Can you add a test too?

PR #3119, opened 2022-03-10
Extend eBPF kernel target with support for additional BPF helpers and more types of BPF maps

osinstom:
This PR extends the eBPF kernel target with:
- support for defining other BPF map types
- support for BPF map-in-maps
- support for BPF maps using BPF spinlocks
- method to generate packet descriptor type (e.g. `__sk_buff`) depending on the eBPF hook type. It's required to support mixed deployments leveraging both TC and XDP programs.
- method to generate packet buffer's size adjustment
- BTF annotation of BPF maps, so that BPF types can be read at runtime

### DIFF
--- a/backends/ebpf/target.cpp
+++ b/backends/ebpf/target.cpp
@@ -76,13 +76,20 @@ void KernelSamplesTarget::emitTableDecl(Util::SourceCodeBuilder* builder,
 
     kind = getBPFMapType(tableKind);
 
-    if (tableKind == TablePerCPUArray || tableKind == TableArray) {
+    if (keyType != "u32" && (tableKind == TablePerCPUArray || tableKind == TableArray)) {
         // it's more safe to overwrite user-provided key type,
         // as array map must have u32 key type.
         keyType = "u32";
-    } else if (tableKind == TableProgArray) {
+        ::warning(ErrorType::WARN_INVALID,
+                  "Invalid key type (%1%) for table kind %2%, replacing with u32",
+                  keyType, kind);
+    } else if (tableKind == TableProgArray && (keyType != "u32" || valueType != "u32")) {
         keyType = "u32";
         valueType = "u32";
+        ::warning(ErrorType::WARN_INVALID,
+                  "Invalid key type (%1%) or value type (%2%) for table kind %3%, "
+                  "replacing with u32",
+                  keyType, valueType, kind);
     }
 
     if (tableKind == TableLPMTrie) {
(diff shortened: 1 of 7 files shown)


backends/ebpf/target.cpp
```diff
@@ -85,7 +153,7 @@ void KernelSamplesTarget::emitLicense(Util::SourceCodeBuilder* builder, cstring
 
 void KernelSamplesTarget::emitCodeSection(
     Util::SourceCodeBuilder* builder, cstring sectionName) const {
-    builder->appendFormat("SEC(\"prog\")\n", sectionName.c_str());
+    builder->appendFormat("SEC(\"%s\")\n", sectionName.c_str());
```
osinstom, line 156: `prog` was hardcoded here before. It's moved to `ebpfProgram.cpp` now.

backends/ebpf/target.cpp
```diff
@@ -62,19 +70,79 @@ void KernelSamplesTarget::emitTableDecl(Util::SourceCodeBuilder* builder,
...
-    builder->appendFormat("sizeof(%s), sizeof(%s), %d)",
-                          keyType.c_str(), valueType.c_str(), size);
+    cstring kind, flags;
+    cstring registerTable = "REGISTER_TABLE(%s, %s, %s, %s, %d)";
+    cstring registerTableWithFlags = "REGISTER_TABLE_FLAGS(%s, %s, %s, %s, %d, %s)";
+
+    kind = getBPFMapType(tableKind);
+
+    if (tableKind == TablePerCPUArray || tableKind == TableArray) {
+        // it's more safe to overwrite user-provided key type,
```
mbudiu-vmw, line 80: Perhaps you should give a warning if this is not what the user specified.

backends/ebpf/target.h
```diff
@@ -53,6 +59,34 @@ class Target {
     virtual void emitTableDecl(Util::SourceCodeBuilder* builder,
                                cstring tblName, TableKind tableKind,
                                cstring keyType, cstring valueType, unsigned size) const = 0;
+    virtual void emitTableDeclSpinlock(Util::SourceCodeBuilder* builder,
+                               cstring tblName, TableKind tableKind,
+                               cstring keyType, cstring valueType, unsigned size) const {
+        (void) builder;
```
mbudiu-vmw, line 65: how about actually giving a warning that on this architecture the table is not supported?
I suspect that this is what it amounts to.

mbudiu-vmw (approved):
I hope that you will soon submit a PR which tests some of these new table type too.
I am approving, but perhaps you can consider the small changes that I requested.

osinstom:
@mbudiu-vmw I've address your review, thanks! We have ~70 test cases based on the PTF framework that verify different mechanisms of PSA impl for eBPF backend. We'll contribute them too.

PR #3120, opened 2022-03-11
Solution for "Non Type_Bits type bool for expression" error

vladyslav-dubina:
Added a simple test verify_disjunction.p4 to show the need for this change, we can run this test with `--loopsUnroll` argument
Changes solve the `Non Type_Bits type bool for expression` error.
Since there are cases when the `expression ` in the `setNonConstant ` function is of type `Type_Boolean`. This is mainly possible with multiple `or` in the condition.

### DIFF
--- a/midend/interpreter.cpp
+++ b/midend/interpreter.cpp
@@ -750,7 +750,7 @@ void ExpressionEvaluator::checkResult(const IR::Expression* expression,
 
 void ExpressionEvaluator::setNonConstant(const IR::Expression* expression) {
     auto type = typeMap->getType(expression, true);
-    if (type->is<IR::BoolLiteral>()) {
+    if (type->is<IR::BoolLiteral>() || type->is<IR::Type_Boolean>()) {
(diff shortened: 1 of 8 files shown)


midend/interpreter.cpp
```diff
@@ -750,7 +750,7 @@ void ExpressionEvaluator::checkResult(const IR::Expression* expression,
 
 void ExpressionEvaluator::setNonConstant(const IR::Expression* expression) {
     auto type = typeMap->getType(expression, true);
-    if (type->is<IR::BoolLiteral>()) {
+    if (type->is<IR::BoolLiteral>() || type->is<IR::Type_Boolean>()) {
```
fruffy, line 753: I would remove `BoolLiteral `here, it does not seem to make sense if you are checking for a NonConstant?

mbudiu-vmw: Yes, a type cannot be a literal.

mbudiu-vmw (changes requested):

mbudiu-vmw (approved):

PR #3121, opened 2022-03-11
Fixing bug for ParserUnroll application to a p4 program with a header union

VolodymyrPeschanenkoIntel:
These small changes allow to use ParserUnroll class for the programs with header unions.


midend/parserUnroll.cpp
```diff
@@ -178,13 +178,22 @@ class ParserStateRewriter : public Transform {
...
  protected:
-    const IR::Type* getTypeArray(const IR::Node* element) {
+const IR::Type* getTypeArray(const IR::Node* element) {
         if (element->is<IR::ArrayIndex>()) {
             const IR::Expression* left = element->to<IR::ArrayIndex>()->left;
             if (left->type->is<IR::Type_Stack>())
                 return left->type->to<IR::Type_Stack>()->elementType;
         }
-        return typeMap->getType(element, true);
+        auto* currentType = typeMap->getType(element);
```
fruffy, line 187: How can it happen that type information is missing here? Is this a problem with type inference?

mbudiu-vmw:
The right solution is to fix the evaluation of expressions with type headerunion in the symbolic interpreter.
I think that #3063 attempts to do this, but it is not yet ready.

midend/parserUnroll.cpp
VolodymyrPeschanenkoIntel: The problem with right part of the [assignment](https://github.com/VolodymyrPeschanenkoIntel/p4c/blob/205467c5841284ee7af871bd0198b803f1e04c06/testdata/p4_16_samples/parser-unroll-test7.p4#L37). If element is hdr.h[0].i1.index then  typeMap->getType(element) returns nullptr.

VolodymyrPeschanenkoIntel:
> The right solution is to fix the evaluation of expressions with type header union in the symbolic interpreter. I think that #3063 attempts to do this, but it is not yet ready.

The getTypeArray function is not a part of the interpreter and It doesn't take part in a calculation process. #3063 contains changes for support header union in a evaluation process. I think that is a some particular case for support of header union.

fruffy:
Now that https://github.com/p4lang/p4c/pull/3063 has been merged what is the status of this pull request?

VolodymyrPeschanenkoIntel:
> Now that #3063 has been merged what is the status of this pull request?

These changes still actual, I wait for review. #3063 contains update for interpreter. This PR contains changes inside parserUnroll.

fruffy:
@mbudiu-vmw Do you have any comments on this pull request?

mbudiu-vmw:
Yes, I have requested changes which have not been carried out.
There are a couple of comments in the code which have not been addressed.

VolodymyrPeschanenkoIntel:
> Yes, I have requested changes which have not been carried out. There are a couple of comments in the code which have not been addressed.

Could, you, please, specify them?

VolodymyrPeschanenkoLitSoft:
As far as I understood we are waiting for solving the issue with the description of header union in p4lang [here](https://github.com/p4lang/p4-spec/issues/876). It was opened two years ago. It looks like we are stuck here.

mbudiu-vmw:
The compiler should usually be ahead of the spec - this helps us make sure that the spec is right.
I think that the intent was always to support stacks of header unions.

VolodymyrPeschanenkoIntel:
@mbudiu-vmw, Could you, please, make a review again?

mbudiu-vmw:
This PR contains only test cases, is this intended?

mbudiu-vmw (approved):

VolodymyrPeschanenkoLitSoft:
> This PR contains only test cases, is this intended?

I would like to add an additional test for header union stacks.

Issue #3057, opened 2022-01-27
Comparisons of list expressions, structure-valued expressions, tuples and structs


MollyDream:
In testing all the comparisons between list expressions, structure-valued expressions, tuples, and structs in the p4c compiler, we found some examples rejected by the compiler (relates to p4lang/p4-spec#960). Some are likely compiler bugs, while some may not be.

Note: the tests are reordered and renumbered for a cleaner view, so it looks different from  p4lang/p4-spec#960.
```
struct S {
    bit<32> a;
    bit<32> b;
}

control compute(inout hdr h) {
    apply {
        bool b1 = { 1, 2 } == { 1, 2 };

        bool b2 = { a = 8w1, b = 8w2 } == { a = 8w1, b = 8w2 };
         // bool b2_ = { a = 1, b = 2 } == { a = 1, b = 2 }; // p4c: Compiler Bug: Could not find type for <Constant>(11032) 1

         // bool b3 = { a = 1, b = 2 } == {1, 2};  // p4c: error: ==: Cannot unify type Tuple(2) with unknown struct
        bool b4 = (S) { a = 1, b = 2 } == {1, 2};

        bool b5 = (S) { a = 1, b = 2 } == { a = 1, b = 2 };
        // bool b5_ = { a = 1, b = 2 } == (S) { a = 1, b = 2 }; // p4c: error: ==: Cannot unify struct S to unknown struct
       
        bool b6 = (S) { a = 1, b = 2 } == (S) { a = 1, b = 2 };

        S s1 = { a = 1, b = 2 };
        S s2 = { a = 1, b = 2 };
        bool b7 = s1 == { a = 1, b = 2 };
        bool b8 = s1 == (S) { a = 1, b = 2 };
        bool b9 = s1 == s2;

        tuple<bit<8>, bit<8>> t1 = {1, 2};
        tuple<bit<8>, bit<8>> t2 = {1, 2};
        bool b10 = t1 == { 1, 2 };
        bool b11 = t1 == t2;
    }
}
```
The tests `b2_` and `b5_` are likely bugs.
- For `b2_`, it seems to be some other issue with the compiler since `b2` works well.
- For `b5_`, it seems weird that the compiler cannot infer the type of the LHS. Given that `b5` works well, and there is no rule in the spec enforcing the direction of type inference, I think it is more likely a problem with the compiler.

With the test `b3` aside, it seems tempting to summarize the allowed comparisons as:
[...]

PR #3122, opened 2022-03-11
Unifying struct and list expressions fix (#3057) 

milossyrmia:
Unifying struct and list expressions fix (#3057) 

* I corrected the code so that the compiler bug does not occur when comparing structures of unknown type, but to report an error

* I enabled a comparison between the structure and the cast structure so that cast can be put on either side

* Added tests


frontends/p4/typeChecking/typeChecker.cpp
```diff
@@ -1679,8 +1680,12 @@ bool TypeInference::compare(const IR::Node* errorPosition,
 
             bool lcst = isCompileTimeConstant(compare->left);
             bool rcst = isCompileTimeConstant(compare->right);
-
-            auto tvs = unify(errorPosition, ltype, rtype);
+            TypeVariableSubstitution* tvs;
```
mbudiu-vmw, line 1683: space between ) and {

testdata/p4_16_errors_outputs/issue3057-1.p4-stderr
```diff
@@ -0,0 +1,6 @@
+issue3057-1.p4(11): [--Werror=type-error] error: ==: cannot compare initializers with unknown types
```
mbudiu-vmw, line 1: this message isn't very good. I know you didn't change this, but perhaps you can fix it.
The message should say "cannot compare structure-valued expressions with unknown types".

mbudiu-vmw (changes requested):

milossyrmia:
Hi, @mbudiu-vmw, I implemented the small changes you requested.
I was referring to an issue (#3057) and I'm not sure what kind of an issue should I file.

testdata/p4_16_errors_outputs/issue3057-1.p4-stderr
mbudiu-vmw: I expect that you will need to change these reference files now that the error message has changed.

milossyrmia:
Hi @mbudiu-vmw, I implemented the changes you requested, but I forgot to change the output in the test, now I did it too.

PR #3124, opened 2022-03-12
Fix off-by-one bug in source file position calculation

### LINKED ISSUE #3112: InputSources::getSourceLine returns wrong file for extern method decl in last non-empty line of file
Consider the following set of input files:

foo_example.p4:
```
13: #include <foo_externs.p4>
14: #include <foo_parser.p4>
```

foo_externs.p4:
```
18: #ifndef _FOO_EXTERNS_P4_
19: #define _FOO_EXTERNS_P4_
20: 
21: #include "foo_core.p4"
22: 
23: extern void func1();
24: extern void func2();
25: 
26: #endif
```

This generates the following pre-processor output:
```
327: # 22 "/path/to/foo_externs.p4" 2
328: 
329: extern void func1();
330: extern void func2();
331: # 14 "foo_example.p4" 2
332: # 1 "/path/to/foo_parser.p4" 1
```

When parsed, this generates intra-p4c data structures:

Entries in InputSources::line_file_map:
```
[310] = {"foo_core.p4", 1}
[327] = {"foo_externs.p4", 22}
[331] = {"foo_example.p4", 14}
[332] = {"foo_parser.p4", 1}
```

SourceInfo for extern method definitions in IR:
```
func1: 329.13-329.17
func2: 330.13-330.17
```

Now, if I implement an IR pass that does this:

```
bool PassName::preorder(const IR::Method* m) {
    cstring srcName = m->srcInfo.getSourceFile();
```

... I get answer "foo_example.p4" for method func2, rather than the expected "foo_externs.p4". However, it works OK for func1...

I believe the issue arises in the following code:
```
SourceFileLine InputSources::getSourceLine(unsigned line) const {
    auto it = line_file_map.upper_bound(line+1);
```

... when called for line==330 (func2), it points at the [332] entry, so gets adjusted to the [331] entry, so returns filename "foo_example.p4". If I modify the code _not_ to add 1 to line:

```
SourceFileLine InputSources::getSourceLine(unsigned line) const {
    auto it = line_file_map.upper_bound(line);
```

... then for line==330 (func2), it points at the [331] entry, so gets adjusted to the [330] entry, so returns filename "foo_externs.p4" which is correct.

I _guess_ that the "+1" may have been a mistaken attempt at correcting for 0-based-vs-1-based line numbers, or the fact that #line directives describe numbering for the "physical" line after the line containing the directive? But the physical line numbers in line_file_map use the same numbering base/scheme as the line numbers in SourceInfo, so there's no need for any offset adjustment, and we want to search line_file_map for the first directive in any line after that which the SourceInfo refers to, so the searching process doesn't need to care that the #line directive describes the line after it; that aspect is dealt with a few lines later where realLine is calculated.

mbudiu-vmw:
Fixes #3112

### DIFF
--- a/lib/source_file.cpp
+++ b/lib/source_file.cpp
@@ -160,7 +160,7 @@ void InputSources::mapLine(cstring file, unsigned originalSourceLineNo) {
 }
 
 SourceFileLine InputSources::getSourceLine(unsigned line) const {
-    auto it = line_file_map.upper_bound(line+1);
+    auto it = line_file_map.upper_bound(line);
     if (it == line_file_map.begin())
         // There must be always something mapped to line 0
         BUG("No source information for line %1%", line);


mbudiu-vmw:
@swarren does this fix work for you?
It seems that your diagnosis was correct.

swarren:
Yes, that's the fix I have locally. Sorry I didn't get around to a PR myself; I'm still waiting on corporate approval to contribute code to this project.

mbudiu-vmw:
There is no attached test since this bug is normally not visible in any way in the output produced by the compiler.

fruffy (approved):

PR #3125, opened 2022-03-12
Struct expressions can be ConstantExpressions

### LINKED ISSUE #3098: Should structure-valued expressions with fields that are all compile-time known, itself be compile-time known?
Attached is an example program where that appears to not be the case.

```bash
$ p4test --version
p4test
Version 1.2.2.1 (SHA: 289431c0d BUILD: DEBUG)

$ git log -n 1 | head -n 3
commit 289431c0de84543adf26372fe6c157bdfcf76733
Author: Kamlesh Kumar <kamleshbhalui@gmail.com>
Date:   Sat Feb 19 07:33:53 2022 +0530
```

The following command gives the compile-time error shown.  It seems
that perhaps p4c consideres that a structure-valued expression is
never a compile-time known value, even if all field values are
compile-time known values.

Is that true?

If it is true now, is it reasonable to change things so that if all
field values are compile-time known values, then the structure-valued
expression as a whole is also considered a compile-time known value?

```bash
$ p4test v1model_valuelut_with_const_entries1.p4 
v1model_valuelut_with_const_entries1.p4(149): [--Werror=invalid] error: StructExpression: Cannot evaluate to a compile-time constant
        default_value = (my_result1_t) {val1=0, val2=0})
                                       ^^^^^^^^^^^^^^^^
```

The command below has a similar error message, probably for the same
root cause, but I am not sure since the error is less specific.  Note
that the first occurrence of 'const_entries' for extern instance named
lut1 gives no such compile-time error, so it seems that list
expressions that bottom out in all compile-time known values are
considered compile-time known values by p4c today.

```bash
$ p4test -DCONST_ENTRIES_WITH_STRUCT_TYPES v1model_valuelut_with_const_entries1.p4 
v1model_valuelut_with_const_entries1.p4(142): [--Werror=invalid] error: ListExpression: Cannot evaluate to a compile-time constant
        const_entries = {
                        ^
```
[v1model_valuelut_with_const_entries1.p4.txt](https://github.com/p4lang/p4c/files/8134051/v1model_valuelut_with_const_entries1.p4.txt)

mbudiu-vmw:
Fixes #3098
The typechecker was fine, but the "evaluator" https://p4.org/p4-spec/docs/P4-16-v1.2.2.html#sec-p4-abstract-mach wasn't.


jfingerh:
I see that some of the tests failed.  Do you think the changes are OK, or do you think you want to make changes that might make the tests pass?

jafingerhut (approved):
LGTM.  As is my usual process for p4c reviews, I have not attempted to review the C++ code changes for correctness.  I have compiled p4c with these changes on my local system and verified that the error messages I saw before with the two variations of the program attached to my issue https://github.com/p4lang/p4c/issues/3098 now both give 0 error messages with this modified version of the compiler, so that definitely looks like an improvement to me.  Thanks.

PR #3127, opened 2022-03-14
Problems with Cast type

vladyslav-dubina:
It helps to solve the error :
```
 In file: /p4c/frontends/p4/typeMap.cpp:126
Compiler Bug: Could not find type for <Cast>(44247) 
```
The error was due to incorrect handling of `Cast` in `interpreter.cpp`.
For example use  `default.p4` test with `--loopsUnroll` argument

### DIFF
--- a/midend/interpreter.cpp
+++ b/midend/interpreter.cpp
@@ -858,17 +858,19 @@ void ExpressionEvaluator::postorder(const IR::Operation_Unary* expression) {
         set(expression, result);
         return;
     }
+    if (auto cast = expression->to<IR::Cast>()) {
+        if (cast->destType->is<IR::Type_Boolean>() && l->is<SymbolicInteger>()) {
+            l = new SymbolicBool(sv->state);
+        } else if (cast->destType->is<IR::Type_Bits>() && l->is<SymbolicBool>()) {
+            l = new SymbolicInteger(sv->state, cast->destType->to<IR::Type_Bits>());
+        } else {
+            BUG_CHECK(!l->is<SymbolicInteger>() || !l->is<SymbolicBool>(),
+                      "%1% unexpected type %2% in cast", cast, cast->destType);
+         }
+        set(expression, l);
+        return;
+    }
     if (sv->isUnknown()) {
-        if (auto cast = expression->to<IR::Cast>()) {
-            if (cast->destType->is<IR::Type_Boolean>() && l->is<SymbolicInteger>()) {
-                l = new SymbolicBool(sv->state);
-            } else if (cast->destType->is<IR::Type_Bits>() && l->is<SymbolicBool>()) {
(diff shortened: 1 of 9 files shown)


OleksandrLetychevskyiIntel:
@fruffy , some of the tests have failed, but I can't achieve the same behavior on the local machine

fruffy:
These tests use a different compiler, `ebpf-kernel` for example. You need to compile the programs with that compiler.

midend/interpreter.cpp
```diff
@@ -858,17 +858,19 @@ void ExpressionEvaluator::postorder(const IR::Operation_Unary* expression) {
         set(expression, result);
         return;
     }
+    if (auto cast = expression->to<IR::Cast>()) {
```
mbudiu-vmw, line 861: this is not the correct fix, as written for casts constant folding will never happen.
The correct fix is to allow for result below after `auto result = clone->apply(cf)` to return a different type.

mbudiu-vmw (changes requested):

midend/interpreter.cpp
fruffy: The problem is that entering constant folding will throw this error. The type is missing despite type inference being applied.

mbudiu-vmw: this means that there is a second bug, but this is still not the right fix.

mbudiu-vmw: But really ExpressionEvaluator is an Inspector, there is no reason for the program to have nodes without a type in the tpyeMap.

fruffy: Yes, the bug with the missing type seems unrelated. However, moving cast out like this can probably solve some of the "xzy is not a bool/bit" errors without extra code. Unless `constantFolding` is required here. I do not exactly know the intention of this interpreter code. 

https://github.com/p4lang/p4c/blob/main/frontends/common/constantFolding.cpp#L815

This is the cast that failed, I am not sure how it can happen. Maybe the cast node is generated and not part of the type map?

mbudiu-vmw: The crash happens here:
```
auto result = clone->apply(cf);
```

and that is because `clone` does not have a type in the typemap, since it was just created. The solution for this is in the previous branch of the `if` statement, which does this:

```
auto type = typeMap->getType(getOriginal(), true);
typeMap->setType(clone, type);  // needed by the constant folding
```

That's why the type isn't there.

But there is another bug; this code assumes that the output type is the same as the input type, which is not true for a cast. I will fix this in a separate PR and we can close this one.

mbudiu-vmw:
The correct fix is in https://github.com/p4lang/p4c/pull/3135.
Next time please file first an issue describing the problem, and then a PR with the solution.

PR #3128, opened 2022-03-14
DPDK: Add support for "mirror_packet" PNA extern

usha1830:
This PR adds support for PNA extern "mirror_packet"

```
extern void mirror_packet(MirrorSlotId_t mirror_slot_id,
                          MirrorSessionId_t mirror_session_id);
```

This translates to mirror instruction in dpdk
`mirror <SLOT_ID> <SESSON_ID>`
If the arguments are constants, they should be moved to Metadata as the mirror instruction in dpdk does not support immediate values as operands.

Added a valid test and a few negative tests.
Negative tests are added for the following:
- Mirror session id 0 is reserved by the architecture, and must not be used by a P4 developer
- Invoking mirror_packet(x) is supported from control blocks other than Main control
- Passing incorrect number of arguments
- Passing non-constant arguments

### DIFF
--- a/backends/dpdk/dpdkHelpers.cpp
+++ b/backends/dpdk/dpdkHelpers.cpp
@@ -765,6 +765,33 @@ bool ConvertStatementToDpdk::preorder(const IR::MethodCallStatement *s) {
             } else {
                 ::error("%1%: unhandled function", s);
             }
+        } else if (a->method->name == "mirror_packet") {
+            auto slot_id = a->expr->arguments->at(0)->expression;
+            auto session_id = a->expr->arguments->at(1)->expression;
+            const IR::Expression *slot_idexpr = slot_id;
+            const IR::Expression *session_idexpr = session_id;
+            // mirror instruction supports only metadata as operands, move any constant arguments
+            // to metadata fields
+            if (slot_id->is<IR::Constant>()){
+                IR::ID tmpName(refmap->newName("mirrorSlot"));
+                BUG_CHECK(metadataStruct, "Metadata structure missing unexpectedly!");
+                metadataStruct->fields.push_back(new IR::StructField(tmpName, slot_id->type));
+                auto tmpMember = new IR::Member(new IR::PathExpression("m"), tmpName);
+                add_instr(new IR::DpdkMovStatement(tmpMember, slot_id));
+                slot_idexpr = tmpMember;
+            }
+            if (session_id->is<IR::Constant>()){
+                unsigned value =  session_id->to<IR::Constant>()->asUnsigned();
+                if (value == 0)
+                    ::error("Mirror session ID 0 is reserved for use by Architecture");
+                IR::ID tmpName(refmap->newName("mirrorSession"));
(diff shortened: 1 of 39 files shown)


backends/dpdk/dpdkHelpers.cpp
```diff
@@ -765,6 +765,33 @@ bool ConvertStatementToDpdk::preorder(const IR::MethodCallStatement *s) {
             } else {
                 ::error("%1%: unhandled function", s);
             }
+        } else if (a->method->name == "mirror_packet") {
+            auto slot_id = a->expr->arguments->at(0)->expression;
```
mbudiu-vmw, line 769: first you should check that there are exactly 2 arguments and give an error otherwise (probably a modelError).

backends/dpdk/dpdkHelpers.cpp
```diff
@@ -765,6 +765,33 @@ bool ConvertStatementToDpdk::preorder(const IR::MethodCallStatement *s) {
...
+            auto session_id = a->expr->arguments->at(1)->expression;
+            const IR::Expression *slot_idexpr = slot_id;
+            const IR::Expression *session_idexpr = session_id;
+            // mirror instruction supports only metadata as operands, move any constant arguments
+            // to metadata fields
+            if (slot_id->is<IR::Constant>()){
```
mbudiu-vmw, line 775: no space before {

backends/dpdk/dpdkHelpers.cpp
```diff
@@ -765,6 +765,33 @@ bool ConvertStatementToDpdk::preorder(const IR::MethodCallStatement *s) {
...
+                IR::ID tmpName(refmap->newName("mirrorSlot"));
+                BUG_CHECK(metadataStruct, "Metadata structure missing unexpectedly!");
+                metadataStruct->fields.push_back(new IR::StructField(tmpName, slot_id->type));
+                auto tmpMember = new IR::Member(new IR::PathExpression("m"), tmpName);
+                add_instr(new IR::DpdkMovStatement(tmpMember, slot_id));
+                slot_idexpr = tmpMember;
+            }
```
mbudiu-vmw, line 782: are you sure that anything that is not a constant will work?

mbudiu-vmw (changes requested):

backends/dpdk/dpdkHelpers.cpp
usha1830: No of arguments mismatch is caught by the frontend, I also added a test for that.
I can add the check to be on the safer side.

usha1830: Added a check.

backends/dpdk/dpdkHelpers.cpp
usha1830: Frontend rejects the program if the arguments to mirror_packet are not compile-time constants. Updated the code to assert on this condition.

usha1830:
@mbudiu-vmw I have updated the code as per you comments. Please let me know if anything else is required.

mbudiu-vmw (approved):

PR #3129, opened 2022-03-14
Use 0 as action ID for NoAction and refactor eBPF table implementation

osinstom:
This PR changes the way of handling `NoAction` in eBPF tables. With this change, `NoAction` will always be allocated with actionID = 0 regardless of its position in the action list. User-defined actions are enumerated starting from index=1. 

Additionally, this PR introduces modifications to the format of the ActionIDs list. We get rid of `enum` to store action IDs and implement new naming format: `<TABLE_NAME>_ACT_<ACTION_NAME>`. 

Also, this PR refactors eBPF table implementation, so that it's easier to inherit from the base eBPF table. 

Co-authored-by: Jan Palimąka <jan.palimaka@orange.com>

### DIFF
--- a/backends/ebpf/ebpfTable.cpp
+++ b/backends/ebpf/ebpfTable.cpp
@@ -374,7 +374,12 @@ void EBPFTable::emitKey(CodeBuilder* builder, cstring keyName) {
                 swap = "bpf_htonl";
             } else if (width <= 64) {
                 swap = "bpf_htonll";
-            }  // TODO: handle width > 64 bits
+            } else {
+                // TODO: handle width > 64 bits
+                ::error(ErrorType::ERR_UNSUPPORTED,
+                        "%1%: fields wider than 64 bits are not supported yet",
+                        fieldName);
+            }
         }
 
         bool isLPMKeyBigEndian = false;
--- a/backends/ebpf/ebpfTable.cpp
+++ b/backends/ebpf/ebpfTable.cpp
@@ -22,45 +22,38 @@ limitations under the License.
 
 namespace EBPF {
 
-namespace {
-class ActionTranslationVisitor : public CodeGenInspector {
- protected:
-    const EBPFProgram*  program;
-    const IR::P4Action* action;
-    cstring             valueName;
-
- public:
-    ActionTranslationVisitor(cstring valueName, const EBPFProgram* program):
-            CodeGenInspector(program->refMap, program->typeMap), program(program),
-            action(nullptr), valueName(valueName)
-    { CHECK_NULL(program); }
-
-    bool preorder(const IR::PathExpression* expression) {
-        auto decl = program->refMap->getDeclaration(expression->path, true);
-        if (decl->is<IR::Parameter>()) {
-            auto param = decl->to<IR::Parameter>();
-            bool isParam = action->parameters->getParameter(param->name) == param;
-            if (isParam) {
-                builder->append(valueName);
-                builder->append("->u.");
-                cstring name = EBPFObject::externalName(action);
-                builder->append(name);
-                builder->append(".");
-                builder->append(expression->path->toString());  // original name
-                return false;
-            }
-        }
-        visit(expression->path);
+bool ActionTranslationVisitor::preorder(const IR::PathExpression* expression) {
+    if (isActionParameter(expression)) {
+        cstring paramStr = getActionParamStr(expression);
+        builder->append(paramStr.c_str());
         return false;
     }
+    visit(expression->path);
+    return false;
+}
 
-    bool preorder(const IR::P4Action* act) {
-        action = act;
-        visit(action->body);
-        return false;
+bool ActionTranslationVisitor::isActionParameter(const IR::PathExpression *expression) const {
+    auto decl = program->refMap->getDeclaration(expression->path, true);
+    if (decl->is<IR::Parameter>()) {
+        auto param = decl->to<IR::Parameter>();
+        return action->parameters->getParameter(param->name) == param;
     }
-};  // ActionTranslationVisitor
-}  // namespace
+    return false;
+}
+
+cstring ActionTranslationVisitor::getActionParamStr(const IR::Expression *expression) const {
+    cstring actionName = EBPFObject::externalName(action);
+    auto paramStr = Util::printf_format("%s->u.%s.%s",
+                                        valueName, actionName,
+                                        expression->toString());
+    return paramStr;
+}
+
+bool ActionTranslationVisitor::preorder(const IR::P4Action* act) {
+    action = act;
+    visit(action->body);
+    return false;
+}
 
 ////////////////////////////////////////////////////////////////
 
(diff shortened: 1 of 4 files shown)


backends/ebpf/ebpfTable.h
```diff
@@ -38,44 +58,84 @@ class EBPFTableBase : public EBPFObject {
...
+    }
+    virtual void emitLookupDefault(CodeBuilder* builder, cstring key, cstring value) {
+        builder->target->emitTableLookup(builder, defaultActionMapName, key, value);
+        builder->endOfStatement(true);
+    }
+    virtual bool isMatchTypeSupported(const IR::Declaration_ID* matchType) {
+        return matchType->name.name == P4::P4CoreLibrary::instance.exactMatch.name ||
+               matchType->name.name == P4::P4CoreLibrary::instance.lpmMatch.name;
+    }
+    // Whether to drop packet if no match entry found.
```
mbudiu-vmw, line 117: this is not really standard P4 semantics.
In P4 a table apply can never cause a packet to be dropped; you have to code that up explicitly.

backends/ebpf/ebpfTable.h
```diff
@@ -38,44 +58,84 @@ class EBPFTableBase : public EBPFObject {
                   CodeGenInspector* codeGen) :
             program(program), instanceName(instanceName), codeGen(codeGen) {
         CHECK_NULL(codeGen); CHECK_NULL(program);
-        keyTypeName = program->refMap->newName(instanceName + "_key");
-        valueTypeName = program->refMap->newName(instanceName + "_value");
+        keyTypeName = instanceName + "_key";
```
mbudiu-vmw, line 61: The original code made sure that no such variables exist already in the code.
What would happen if they do exist?

backends/ebpf/ebpfTable.cpp
```diff
@@ -338,15 +435,20 @@ void EBPFTable::emitAction(CodeBuilder* builder, cstring valueName) {
         auto action = adecl->getNode()->to<IR::P4Action>();
         cstring name = EBPFObject::externalName(action), msgStr, convStr;
         builder->emitIndent();
-        builder->appendFormat("case %s: ", name.c_str());
+        if (action->name.originalName == P4::P4CoreLibrary::instance.noAction.name) {
```
mbudiu-vmw, line 438: couldn't you unify these two cases by having p4ActionToActionIDName return 0?

backends/ebpf/ebpfTable.h
```diff
@@ -38,44 +58,84 @@ class EBPFTableBase : public EBPFObject {
...
+        return matchType->name.name == P4::P4CoreLibrary::instance.exactMatch.name ||
+               matchType->name.name == P4::P4CoreLibrary::instance.lpmMatch.name;
+    }
+    // Whether to drop packet if no match entry found.
+    // Some table implementations may want to continue processing.
+    virtual bool dropOnNoMatchingEntryFound() const { return true; }
 };
 
-class EBPFCounterTable final : public EBPFTableBase {
+class EBPFCounterTable : public EBPFTableBase {
```
mbudiu-vmw, line 122: I wonder whether these API changes will impact the p4c-xdp project which also extends the ebpf backend.
Could you actually install that project github.com/vmware/p4c-xdp and run its tests as well?
It should be as easy as creating a symlink in the extensions directory.
In principle I can help with fixes there if you cannot do them.

backends/ebpf/ebpfTable.cpp
```diff
@@ -151,54 +186,70 @@ void EBPFTable::emitActionArguments(CodeBuilder* builder,
...
+void EBPFTable::emitValueActionIDNames(CodeBuilder* builder) {
+    // create type definition for action
+    builder->emitIndent();
+    unsigned int action_idx = 1;  // 0 is reserved for NoAction
     for (auto a : actionList->actionList) {
         auto adecl = program->refMap->getDeclaration(a->getPath(), true);
         auto action = adecl->getNode()->to<IR::P4Action>();
-        cstring name = EBPFObject::externalName(action);
+        // NoAction is from standard library
+        if (action->name.originalName == P4::P4CoreLibrary::instance.noAction.name) {
```
mbudiu-vmw, line 210: so why not define that as 0?

backends/ebpf/ebpfTable.cpp
```diff
@@ -289,29 +340,75 @@ void EBPFTable::emitInstance(CodeBuilder* builder) {
...
+
+            if (width <= 8) {
+                swap = "";  // single byte, nothing to swap
+            } else if (width <= 16) {
+                swap = "bpf_htons";
+            } else if (width <= 32) {
+                swap = "bpf_htonl";
+            } else if (width <= 64) {
+                swap = "bpf_htonll";
+            }  // TODO: handle width > 64 bits
```
mbudiu-vmw, line 376: can you generate some error here?

backends/ebpf/ebpfTable.cpp
```diff
@@ -289,29 +340,75 @@ void EBPFTable::emitInstance(CodeBuilder* builder) {
...
+                isLPMKeyBigEndian = true;
         }
 
         builder->emitIndent();
         if (memcpy) {
-            builder->appendFormat("memcpy(&%s.%s, &", keyName.c_str(), fieldName.c_str());
-            codeGen->visit(c->expression);
-            builder->appendFormat(", %d)", scalar->bytesRequired());
+            if (isLPMKeyBigEndian) {
+                // FIXME: will not work on big endian machines because byte swap
```
mbudiu-vmw, line 388: does this bug exist in ubpf too? I thought you fixed it there. If so, could you backport it?

backends/ebpf/ebpfTable.h
mbudiu-vmw: Shouldn't this be 'false' by default?

osinstom: We are fully aware of what the standard P4 semantics is, but the reason for this is how it has been implemented for eBPF model, see the logic before our modifications to this piece of code:

https://github.com/p4lang/p4c/blob/b9a2dcbc34f002cb6b6819dcc543f4caa5e4018a/backends/ebpf/ebpfTable.cpp#L345 

Do you want to change this? If so, this method isn't necessary anymore.

mbudiu-vmw: So you are saying that the original implementation was already wrong? That's not good. I wonder why I did that.
What does ubpf do?
I think it would be good to fix it.
Perhaps you can do this by having some compiler options that need to be set to trigger the non-standard semantics.
This just in case users depend on this behavior.

backends/ebpf/ebpfTable.cpp
osinstom: We don't need to define a constant for `NoAction`, because we explicitly generate `case 0:` here: https://github.com/p4lang/p4c/pull/3129/files#diff-218b38cacc617dd2223b207e565c9b04ace178cfc080f199e48832a3266f3ebdR439.

backends/ebpf/ebpfTable.h
osinstom: I think `newName()` is useful for objects' instances, because it protects from conflicting names if we have multiple lookups to a table. I don't see a reason why we would need a unique key and value **type** - since key and value types are specific to a P4 table. Does it make sense?

mbudiu-vmw: For any P4 program the user writes the generated C should be unambiguous. There should be no way the user program names a table (or some other construct) in such a way that your generated names collide.

mbudiu-vmw (approved):

backends/ebpf/ebpfTable.h
osinstom: ubpf does the same, I'm also not sure why.. perhaps we just followed the implementation for eBPF. I agree that it would be good to fix it. In the PSA impl for eBPF we override this and PSA tables don't drop if match isn't found. 

Should I fix it in a separate PR or this one?

mbudiu-vmw: you can file an issue for it for now so we can keep track.

osinstom: I've opened #3130.

PR #3133, opened 2022-03-15
Improve efficiency of range translation to mask 

vlstill:
This improves efficiency of matching of ranges of form 0..N when translated to masks. In some cases this can significantly decrease the number of transitions in parser.


jafingerhut:
Do you have an example P4 source program and its front/mid-end expected outputs that demonstrates a change in p4c behavior on a program from before this change, vs. after this change?  Thanks.

mbudiu-vmw (approved):

vlstill:
@jafingerhut, the failed test does demonstrate that -- the file in <https://github.com/p4lang/p4c/blob/main/testdata/p4_16_samples/issue2485.p4>compiles now using just one match for the range `0x0000 .. 0x0004`, compared to three before:

```
         transition select(hdr.h.a) {
-            16w0x0: parse;
-            16w0x1: parse;
-            16w0x2 &&& 16w0xfffe: parse;
+            16w0x0 &&& 16w0xfffc: parse;
             16w0x4: parse;
             default: accept;
         }
```

In general this will improve any non-trivial range starting with 0.

@mbudiu-vmw, thanks for your help. I have pushed fixed test result so hopefully that one will work now. Looking at the fedora build, this could be actually related since I do initialize `std::pair<unsigned, unsigned>` out of bigints, I will have a look at this.

vlstill:
Interestingly I cannot replicate the build error with gcc 11.2.0 (which is almost the same version as in the Fedora build, it has 11.2.1) on Ubuntu. Nevertheless, I am trying a work around by casting the bigint explictly in the test. Hopefully it will build now.

vlstill:
Everything passed, is it possible to merge this PR please?

PR #3134, opened 2022-03-15
Allow to use simple types inside out struct as a parser's parameter

VolodymyrPeschanenkoIntel:
The problem was in some additional checking which disable possibility to use struct with simple types as an output parameter for a parser. behavioral-model supports such possibility.

### DIFF
--- a/testdata/p4_16_samples/struct_with_st-bmv2_0.stf
+++ b/testdata/p4_16_samples/struct_with_st-bmv2_0.stf
@@ -1,0 +1,12 @@
+# STF test for /home/vpeschax/networking.switching.barefoot.bf-p4c-tools/res/struct_with_st-bmv2
+# p4testgen seed:  1000
+
+# Traces
+# MyParser: [Parser] MyParser
+# [State] start
+# [Event]: Extract PacketTooShort
+# [Event]: Extract Cond: p4t*zombie.const.0.*packetLen_bits >= 24;
+# [Event]: Prepending the emit buffer to the program packet.
+
+packet 0 00
+expect 0 00$


testdata/p4_16_samples/struct_with_st-bmv2.p4
```diff
@@ -0,0 +1,53 @@
...
+        packet.extract(hdr.bits);
+        hdr.bt = hdr.bits.bt;
+        hdr.it = hdr.bits.it;
+        hdr.b = hdr.bits.b;
+        hdr.x = hdr.bits.x;
+        transition accept;
+    }
+}
+
+control mau(inout headers hdr, inout metadata meta, inout standard_metadata_t sm) {
```
mbudiu-vmw, line 40: can you please use the fields in the hdr here to make sure that they are correctly passed from the parser to the ingress?
Do something with them which influences the visible output.

fruffy (commented):
This is missing an STF test that actually runs the behavioural model.

fruffy:
I will check the validation error. All the other tests pass.

jafingerhut:
In the test program for this PR, there are non-header fields of the struct type `headers`.  This is something that the v1model architecture is documented not to support.  Neither does the PSA architecture.  For some reason that I have forgotten, it has become fairly common that a user-defined struct containing only headers type is one output of a parser, and a separate user-defined struct often called "user-defined metadata" is a separate out or inout struct type from parser to control, and control to deparser, for containing members with any type.

Are your changes intentionally trying to generalize this restriction?  If so, I am curious why you want to do so?

VolodymyrPeschanenkoIntel:
> In the test program for this PR, there are non-header fields of the struct type `headers`. This is something that the v1model architecture is documented not to support. Neither does the PSA architecture. For some reason that I have forgotten, it has become fairly common that a user-defined struct containing only headers type is one output of a parser, and a separate user-defined struct often called "user-defined metadata" is a separate out or inout struct type from parser to control, and control to deparser, for containing members with any type.
> 
> Are your changes intentionally trying to generalize this restriction? If so, I am curious why you want to do so?
> 
> Here is where the restriction is documented for the v1model architecture: https://github.com/p4lang/behavioral-model/blob/main/docs/simple_switch.md#restrictions-on-type-h
> 
> Here is the PR for the PSA specification planning to document this restriction for the PSA architecture: [p4lang/p4-spec#722](https://github.com/p4lang/p4-spec/issues/722)

@jafingerhut , `headers` contains field `bits` which is a header. I understood that this is a restriction, but this example works fine on the behavioral-model and that makes this restriction too general. Do you have some other examples which could be checked without this restriction on the behavioral-model?

jafingerhut:
@VolodymyrPeschanenkoIntel Are you still interested in pursuing these changes?  Or perhaps OK to close this PR as abandoned?  Note: If you do still wish to pursue it, I am still curious what the motivation for the generalization is.  Perhaps it is simply "I know how to generalize it, so I wanted to?"

PR #3136, opened 2022-03-16
Implement EBPF deparser

osinstom:
This PR introduces a common, re-usable implementation of EBPF Deparser that can be used by different architecture models. The implementation is based on https://github.com/vmware/p4c-xdp. 

Since `ebpf_model.p4` doesn't define a deparser block, this PR doesn't affect its code. The main use case is PSA implementation for eBPF backend. It would be good to leverage the common deparser implementation in p4c-xdp at some point.

CC: @kmateuszssak @tatry

### DIFF
--- a/backends/ebpf/ebpfDeparser.cpp
+++ b/backends/ebpf/ebpfDeparser.cpp
@@ -1,0 +1,376 @@
+/*
+Copyright 2022-present Orange
+Copyright 2022-present Open Networking Foundation
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+#include "ebpfDeparser.h"
+
+namespace EBPF {
+
+DeparserBodyTranslator::DeparserBodyTranslator(const EBPFDeparser *deparser) :
+        ControlBodyTranslator(deparser), deparser(deparser) {
+    setName("DeparserBodyTranslator");
+}
+
+bool DeparserBodyTranslator::preorder(const IR::MethodCallExpression* expression) {
+    auto mi = P4::MethodInstance::resolve(expression,
+                                          control->program->refMap,
+                                          control->program->typeMap);
+    auto ext = mi->to<P4::ExternMethod>();
+    if (ext != nullptr) {
+        // We skip headers emit processing which is handled by DeparserHdrEmitTranslator
+        if (ext->method->name.name == p4lib.packetOut.emit.name)
+            return false;
+    }
+
+    return ControlBodyTranslator::preorder(expression);
+}
+
+DeparserPrepareBufferTranslator::DeparserPrepareBufferTranslator(const EBPFDeparser *deparser) :
+        ControlBodyTranslator(deparser), deparser(deparser) {
+    setName("DeparserPrepareBufferTranslator");
+}
+
+bool DeparserPrepareBufferTranslator::preorder(const IR::BlockStatement* s) {
+    for (auto a : s->components) {
+        if (auto method = a->to<IR::MethodCallStatement>()) {
+            if (auto expr = method->methodCall->method->to<IR::Member>()) {
+                if (expr->member.name == p4lib.packetOut.emit.name) {
+                    visit(a);
+                }
+            }
+        }
+    }
+
+    return false;
+}
+
+bool DeparserPrepareBufferTranslator::preorder(const IR::MethodCallExpression* expression) {
+    auto mi = P4::MethodInstance::resolve(expression,
+                                          control->program->refMap,
+                                          control->program->typeMap);
+    auto ext = mi->to<P4::ExternMethod>();
+    if (ext != nullptr) {
+        processMethod(ext);
+        return false;
+    }
+
+    return false;
+}
+
+void DeparserPrepareBufferTranslator::processMethod(const P4::ExternMethod *method) {
+    if (method->method->name.name == p4lib.packetOut.emit.name) {
+        auto decl = method->object;
+        if (decl == deparser->packet_out) {
+            auto expr = method->expr->arguments->at(0)->expression;
+            auto exprType = deparser->program->typeMap->getType(expr);
+            auto headerToEmit = exprType->to<IR::Type_Header>();
+            if (headerToEmit == nullptr) {
+                ::error(ErrorType::ERR_UNSUPPORTED_ON_TARGET,
+                        "Cannot emit a non-header type %1%", expr);
+            }
+
+            unsigned width = headerToEmit->width_bits();
+            builder->emitIndent();
+            builder->append("if (");
+            this->visit(expr);
+            builder->append(".ebpf_valid) ");
+            builder->blockStart();
+            builder->emitIndent();
+            builder->appendFormat("%s += %d;",
+                                  this->deparser->outerHdrLengthVar.c_str(), width);
+            builder->newline();
+            builder->blockEnd(true);
+        }
+    }
+}
+
+DeparserHdrEmitTranslator::DeparserHdrEmitTranslator(const EBPFDeparser *deparser) :
+        DeparserPrepareBufferTranslator(deparser), deparser(deparser) {
+    setName("DeparserHdrEmitTranslator");
+}
+
+void DeparserHdrEmitTranslator::processMethod(const P4::ExternMethod *method) {
+    if (method->method->name.name == p4lib.packetOut.emit.name) {
+        auto decl = method->object;
+        if (decl == deparser->packet_out) {
+            auto expr = method->expr->arguments->at(0)->expression;
+            auto exprType = deparser->program->typeMap->getType(expr);
+            auto headerToEmit = exprType->to<IR::Type_Header>();
+            if (headerToEmit == nullptr) {
+                ::error(ErrorType::ERR_UNSUPPORTED_ON_TARGET,
+                        "Cannot emit a non-header type %1%", expr);
+            }
+
+            cstring msgStr;
+            builder->emitIndent();
+            builder->append("if (");
+            this->visit(expr);
+            builder->append(".ebpf_valid) ");
+            builder->blockStart();
+            auto program = deparser->program;
+            unsigned width = headerToEmit->width_bits();
+            msgStr = Util::printf_format("Deparser: emitting header %s",
+                                         expr->toString().c_str());
+            builder->target->emitTraceMessage(builder, msgStr.c_str());
+
+            builder->emitIndent();
+            builder->appendFormat("if (%s < %s + BYTES(%s + %d)) ",
+                                  program->packetEndVar.c_str(),
+                                  program->packetStartVar.c_str(),
+                                  program->offsetVar.c_str(), width);
+            builder->blockStart();
+            builder->target->emitTraceMessage(builder,
+                                              "Deparser: invalid packet (packet too short)");
+            builder->emitIndent();
+            // We immediately return instead of jumping to reject state.
+            // It avoids reaching BPF_COMPLEXITY_LIMIT_JMP_SEQ.
+            builder->appendFormat("return %s;", builder->target->abortReturnCode().c_str());
+            builder->newline();
+            builder->blockEnd(true);
+            builder->emitIndent();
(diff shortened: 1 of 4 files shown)


osinstom:
@rst0git what kind of tests do you have in mind? 

@stuart-chu Please read the PR description carefully. This PR doesn't extend `ebpf_model.p4` with deparser. This PR just contributes a piece of code that could be used for an architecture model that defines a deparser block. The driving use case is PSA implementation for eBPF, which will be contributed soon.

rst0git:
> what kind of tests do you have in mind?

It would be good if we could add a P4 program with STF test (e.g. https://github.com/rst0git/p4c/commit/a5fdc50019cc449f892a8ac77e2fc08f6bfbd2a4) to validate the functionality of the output code. However, it looks like such test would currently fail because `p4c-ebpf` doesn't apply the deparser code.

osinstom:
@rst0git yes, the goal of this PR isn't to extend `ebpf_model.p4` with deparser. The driving use case is PSA implementation for eBPF. We have almost 70 test cases based on the PTF framework that verify different mechanisms of PSA impl. We'll contribute all of them.

osinstom:
@mbudiu-vmw could I get a review for this PR please?

osinstom:
@rst0git BTW is the test passing on your branch? I'm asking out of curiosity. If there is a decision from `ebpf_model` maintainers (@mbudiu-vmw ?) to extend the architecture model with deparser, I can surely add it, but I want to avoid introducing breaking modifications to the already-defined arch model.

rst0git (approved):

rst0git:
> is the test passing on your branch?

Yes, it is passing without STF test, i.e., the compiler works, but doesn't include a deparser code in the output C program.

> The driving use case is PSA implementation for eBPF.

I am happy to approve this PR but I am not sure I understand how this code will be used.
Would the PSA implementation use a different compiler, or `p4c-ebpf` would be compatible with both `ebpf_model` and PSA?

jonny-zzy:
@osinstom If I want to use this code in ebpf_model for testing, can I modify the ebpf_model.p4 only or any other code which need to be modified?

osinstom:
@rst0git `p4c-ebpf` will be compatible with both `ebpf_model` and PSA. Once this PR and #3137 are merged, we'll submit a PR with PSA impl based on the TC hook. You'll have a better understanding how this deparser can be used.

backends/ebpf/ebpfDeparser.cpp
```diff
@@ -0,0 +1,376 @@
...
+}
+
+DeparserPrepareBufferTranslator::DeparserPrepareBufferTranslator(const EBPFDeparser *deparser) :
+        ControlBodyTranslator(deparser), deparser(deparser) {
+    setName("DeparserPrepareBufferTranslator");
+}
+
+bool DeparserPrepareBufferTranslator::preorder(const IR::BlockStatement* s) {
+    for (auto a : s->components) {
+        if (auto method = a->to<IR::MethodCallStatement>()) {
```
mbudiu-vmw, line 48: You already check the function name below, so it may be enough to just call visit(a) for MethodCallStatements.

backends/ebpf/ebpfDeparser.cpp
```diff
@@ -0,0 +1,376 @@
...
+    }
+
+    return false;
+}
+
+void DeparserPrepareBufferTranslator::processMethod(const P4::ExternMethod *method) {
+    if (method->method->name.name == p4lib.packetOut.emit.name) {
+        auto decl = method->object;
+        if (decl == deparser->packet_out) {
+            auto expr = method->expr->arguments->at(0)->expression;
```
mbudiu-vmw, line 77: you should check that enough arguments exist.
The model seems to check that, but people can load the wrong model, and you don't want to crash.
Then you should call modelError.

backends/ebpf/ebpfDeparser.cpp
```diff
@@ -0,0 +1,376 @@
...
+                            "Only headers with fixed widths supported %1%", f);
+                    return;
+                }
+                emitField(builder, f->name, expr, alignment, etype);
+                alignment += et->widthInBits();
+                alignment %= 8;
+            }
+            builder->blockEnd(true);
+        }
+    }
```
mbudiu-vmw, line 161: Is doing nothing for other methods the right thing?
Should there be an `else error` here?

backends/ebpf/ebpfDeparser.cpp
```diff
@@ -0,0 +1,376 @@
...
+                    ::error(ErrorType::ERR_UNSUPPORTED_ON_TARGET,
+                            "Only headers with fixed widths supported %1%", f);
+                    return;
+                }
+                emitField(builder, f->name, expr, alignment, etype);
+                alignment += et->widthInBits();
+                alignment %= 8;
+            }
+            builder->blockEnd(true);
+        }
```
mbudiu-vmw, line 160: and what if this is not packet_out?

backends/ebpf/ebpfDeparser.cpp
```diff
@@ -0,0 +1,376 @@
...
+    auto program = deparser->program;
+
+    auto et = dynamic_cast<EBPF::IHasWidth *>(type);
+    if (et == nullptr) {
+        ::error(ErrorType::ERR_UNSUPPORTED_ON_TARGET,
+                "Only headers with fixed widths supported %1%", hdrExpr);
+        return;
+    }
+    unsigned widthToEmit = et->widthInBits();
+    unsigned loadSize = 0;
```
mbudiu-vmw, line 176: load?
perhaps this should be renamed

backends/ebpf/ebpfDeparser.cpp
```diff
@@ -0,0 +1,376 @@
...
+        builder->appendFormat(".%s))[%d]", field, i);
+        builder->endOfStatement(true);
+        unsigned freeBits = alignment != 0 ? (8 - alignment) : 8;
+        bitsInCurrentByte = left >= 8 ? 8 : left;
+        unsigned bitsToWrite =
+                bitsInCurrentByte > freeBits ? freeBits : bitsInCurrentByte;
+        BUG_CHECK((bitsToWrite > 0) && (bitsToWrite <= 8),
+                  "invalid bitsToWrite %d", bitsToWrite);
+        builder->emitIndent();
+        if (alignment == 0 && bitsToWrite == 8) {  // write whole byte
```
mbudiu-vmw, line 234: looking forward for tests that will validate this.
I got this kind of code several times wrong, so I don't trust myself to validate it by reading it.

mbudiu-vmw (approved):

backends/ebpf/ebpfDeparser.cpp
osinstom: yeah, this part is always tricky. We'll have tests for that.

backends/ebpf/ebpfDeparser.cpp
osinstom: does P4 language allow to call `emit()` for other fields than `packet_out`? Anyway, I'll produce a bug alert in such a case.

jafingerhut: `emit()` is a method of the extern object `packet_out`, so no: https://github.com/p4lang/p4c/blob/main/p4include/core.p4#L60

mbudiu-vmw: yes, but no one precludes a program writer from defining their own extern with an emit method.
So you should check that it is the right extern and method.

backends/ebpf/ebpfDeparser.cpp
```diff
@@ -0,0 +1,381 @@
...
+                            "Only headers with fixed widths supported %1%", f);
+                    return;
+                }
+                emitField(builder, f->name, expr, alignment, etype);
+                alignment += et->widthInBits();
+                alignment %= 8;
+            }
+            builder->blockEnd(true);
+        } else {
+            BUG("emit() should only be invoked for packet_out");
```
mbudiu-vmw, line 162: I think this should be an error, not a bug.

jonny-zzy:
When I fellow this mothod, I found another place not fixed, Is there any places to modify?
/usr/local/bin/bin/p4c-ebpf ./example.p4 -o example.c
/usr/local/bin/share/p4c/p4include/ebpf_model.p4(70): [--Werror=expected] error: Expected toplevel package package ebpfFilter to have 2 parameters
package ebpfFilter<H>(parse<H> prs,
        ^^^^^^^^^^

ebpf_model.p4 like this:
/* architectural model for EBPF packet filter target architecture */

parser parse<H>(packet_in packet, out H headers);
control filter<H>(inout H headers, out bool accept);
control deparse<H>(in H headers, out packet_out packet);

package ebpfFilter<H>(parse<H> prs,
                      filter<H> filt,
                      deparse<H> dprs);

jonny-zzy:
> @jonny-zzy You also need to extend `ebpfProgram.cpp` as follows:
> 
> * extend `build()` to create `EBPFDeparser` object and assign `IR::Parameter`'s based on the P4 package. Similarly as it is done for Parser and Control: https://github.com/p4lang/p4c/blob/main/backends/ebpf/ebpfProgram.cpp#L56
> * extend `emitC` with a line to call `deparser->emit()`, as it is done for parser and control: https://github.com/p4lang/p4c/blob/main/backends/ebpf/ebpfProgram.cpp#L98

@osinstom 
When I fellow this mothod, I found another place not fixed, Is there any places to modify?
/usr/local/bin/bin/p4c-ebpf ./example.p4 -o example.c
/usr/local/bin/share/p4c/p4include/ebpf_model.p4(70): [--Werror=expected] error: Expected toplevel package package ebpfFilter to have 2 parameters
package ebpfFilter(parse prs,
^^^^^^^^^^

ebpf_model.p4 like this:
/* architectural model for EBPF packet filter target architecture */

parser parse(packet_in packet, out H headers);
control filter(inout H headers, out bool accept);
control deparse(in H headers, out packet_out packet);

package ebpfFilter(parse prs,
filter filt,
deparse dprs);

rst0git:
@jonny-zzy you need this change in backends/ebpf/ebpfProgram.cpp as well:
https://github.com/rst0git/p4c/commit/a5fdc50019cc449f892a8ac77e2fc08f6bfbd2a4#diff-c02a3f37e5b2624700d003c83df7aa6e130b3d5c3150c6dc54360faccb39dcd7L36

jonny-zzy:
class EBPFDeparser has a member packet_out, but I don't see where the assignment is. and the judgment here:  if (decl == deparser->packet_out) ebpfDeparser.cpp line109 will cause bug exception. How to solve this problem?

/usr/local/bin/bin/p4c-ebpf ./test.p4 -o test.c
In file: /root/code/p4c/backends/ebpf/ebpfDeparser.cpp:162
Compiler Bug: emit() should only be invoked for packet_out

p4code like:
control dprs(packet_out packet, in Headers_t headers) {
    apply {
        packet.emit(headers.ethernet);
        packet.emit(headers.ipv4);
    }
}

jonny-zzy:
@osinstom @rst0git I modified the code to support ebpf_model.p4, and compiled .c file finally as expected. The p4 header only included ethernet, arp, ipv4,  as the input parameter of deparser at the same time. clang-14 complier according to the compile option of -o2 or -o3, would exceed the limit stack of 512 bytes kernel, and I couldn't understand the compile optimization options, have you any sugestions? tks all.
p4 header
/*************************************************************************
 ***********************  H E A D E R  ***********************************
 *************************************************************************/
header ethernet_t {
    macAddr_t dstAddr;
    macAddr_t srcAddr;
    bit<16> etherType;
}

header arp_t{
    bit<16>   hwType;
    bit<16>   protoType;
    bit<8>    hwAddrLen;
    bit<8>    protoAddrLen;
    bit<16>   opcode;
}

header arp_ipv4_t {
    bit<48>   hwSrcAddr;
    bit<32>   protoSrcAddr;
    bit<48>   hwDstAddr;
    bit<32>   protoDstAddr;
}

header ipv4_t {
    bit<4>  version;
    bit<4>  ihl;
    bit<8>  diffserv;
    bit<16> totalLen;
    bit<16> identification;
    bit<3>  flags;
    bit<13> fragOffset;
    bit<8>  ttl;
    bit<8>  protocol;
    bit<16> hdrChecksum;
    ip4Addr_t srcAddr;
    ip4Addr_t dstAddr;
}


struct headers {
    @name("ethernet")
    ethernet_t ethernet;
    @name("arp")
    arp_t arp;
    arp_ipv4_t arp_ipv4;
    @name("ipv4")
    ipv4_t     ipv4;
}

seems that if write_byte method is called multiple times, the origin stack of struct header will be enlarged for optimization.

mbudiu-vmw:
Writing comments on a merged pull request is not the best solution; you should probably open a new issue.
In the past there were problems with the stack because llvm would use 8 bytes for each byte of data that was used as a local variable. I don't know if this has been solved in the ebpf llvm back-end.

PR #3137, opened 2022-03-16
Add packet_in.length() and packet_in.advance() methods to a eBPF target

kmateuszssak:
This PR adds to the eBPF kernel target:
- support for packet_in `length()` method
- support for packet_in `advance(...)` method
- packet length variable

The driving use case for this change is PSA implementation for eBPF backend, but both `advance()` and `length()` primitives are part of core P4 language, so they will also be avaible for `ebpf_model.p4`.


testdata/p4_16_samples/advance_ebpf.stf
```diff
@@ -0,0 +1,14 @@
+packet 0 08 00 01
```
mbudiu-vmw, line 1: all these tests are positive, can you add a negative one, where a packet is dropped?

backends/ebpf/ebpfParser.cpp
```diff
@@ -73,6 +74,49 @@ StateTranslationVisitor::compileLookahead(const IR::Expression* destination) {
...
 }
 
+void
+StateTranslationVisitor::compileAdvance(const P4::ExternMethod* extMethod) {
+    auto argExpr = extMethod->expr->arguments->at(0)->expression;
+    cstring argStr;
+    if (auto cnst = argExpr->to<IR::Constant>()) {
+        argStr = cstring::to_cstring(cnst->asUnsigned());
+    } else {
+        argStr = argExpr->toString();
```
mbudiu-vmw, line 84: I don't think this can be correct for an arbitrary expression argExpr. This will generate a P4 representation of the expression, which has nothing to do with the C code you need to generate. You need to call the code generator for expressions here.

backends/ebpf/ebpfParser.cpp
```diff
@@ -73,6 +74,49 @@ StateTranslationVisitor::compileLookahead(const IR::Expression* destination) {
...
+    builder->target->emitTraceMessage(builder, "Parser (advance): check pkt_len=%%d < "
+                                               "last_read_byte=%%d", 2,
+                                      state->parser->program->lengthVar.c_str(),
+                                      offsetStr.c_str());
+    builder->emitIndent();
+    builder->appendFormat("if (%s < %s + BYTES(%s + ",
+                          state->parser->program->packetEndVar.c_str(),
+                          state->parser->program->packetStartVar.c_str(),
+                          state->parser->program->offsetVar.c_str());
+    visit(argExpr);
```
mbudiu-vmw, line 97: This is the right way to do it.
Moreover, you may want to save the result of the evaluation in a temporary so you don't evaluate the expression multiple times. There should be no side-effects, so I think the result will always be correct, but perhaps inefficient.

mbudiu-vmw (changes requested):

backends/ebpf/ebpfParser.cpp
mbudiu-vmw: it is not about the compile-time cost, it is about the runtime cost.
You will evaluate the expression twice.

backends/ebpf/ebpfParser.cpp
kmateuszssak: This was just for logging purposes. We don't have `emitTraceMessage` method that would receive expression as an argument.  
So the easiest way was to put there P4 representation (for logging I thougth it might be enough). 
My proposition is to emit a trace message only in case of Constant. Do you agree?

mbudiu-vmw (approved):

backends/ebpf/ebpfParser.cpp
kmateuszssak: Do you mean eBPF program performance? By the way we can implement advance with only one `visit`

mbudiu-vmw: yes, ebpf performance,
The number of 'visit' calls does not matter.

testdata/p4_16_samples/advance_ebpf.stf
kmateuszssak: Since `advance` works only with constant arguments I changed a testcase that checks only constant + added negative one

mbudiu-vmw:
Then you should just give an error for a non-constant advance.
Although there is a trick you can use to convert a non-constant advance into a (logarithmic number) of constant advances: something like:
```
if (advance_amount & 0x80) 
  advance(128);
if (advance_amount & 0x40) 
  advance(64);
...
```

osinstom:
> Then you should just give an error for a non-constant advance. Although there is a trick you can use to convert a non-constant advance into a (logarithmic number) of constant advances: something like:
> 
> ```
> if (advance_amount & 0x80) 
>   advance(128);
> if (advance_amount & 0x40) 
>   advance(64);
> ...
> ```

That's a nice trick indeed! However, I believe we should be able to satisfy verifier by doing more "smart" checks on memory. It just requires more investigation.

backends/ebpf/ebpfParser.cpp
kmateuszssak: Yes, you're right. I had in mind constant cases before. Assuming we will fix non-constant advance it will affect a performance

kmateuszssak:
@mbudiu-vmw I think I've addressed all of the comments. Thank you for the review and please merge if tests will pass (I'm pretty sure about this).

PR #3139, opened 2022-03-18
Implement PSA for eBPF backend

osinstom:
This PR implements the PSA model for eBPF backend. Please read the included README to learn about the design and implementation details. This PR adds the PSA model that leverages the eBFP/TC hook. In subsequent PRs we're planning to add PTF-based test cases, support for XDP-based design, support for all PSA externs and ternary matching. 

The basic usage model:

```
$ p4c-ebpf --arch psa --target kernel -o out.c program.p4
```

Note that current code contains many placeholders that are not used now, but they will make adding support for PSA externs easier.

This is a big milestone for the team that have been working on this project for over 1 year. Thanks to main developers @kmateuszssak @tatry and other folks that contributed to this project: @fdangtran and @elfadel.


backends/ebpf/CMakeLists.txt
```diff
@@ -35,7 +35,26 @@ set (P4C_EBPF_SRCS
...
+  ../bmv2/common/controlFlowGraph.cpp
+  ../bmv2/common/deparser.cpp
+  ../bmv2/common/expression.cpp
+  ../bmv2/common/extern.cpp
+  ../bmv2/common/globals.cpp
+  ../bmv2/common/header.cpp
+  ../bmv2/common/lower.cpp
+  ../bmv2/common/metermap.cpp
+  ../bmv2/common/parser.cpp
+  ../bmv2/common/programStructure.cpp
```
mbudiu-vmw, line 50: It looks to me that this is in fact the only file you need from bmv2, but you bring all the other ones because they are included. There must be a better way to do this.
I would start by moving psaSwitch.cpp outside of the bmv2 backend. That should rip out all dependencies that are not needed.
Do you want to talk about it before I review this PR?

osinstom: Yes, I'm fine with everything what can make the code cleaner and maintainable.

osinstom: I'd need some guidance from you/others, because I'm not so familiar with the BMv2 part.

mbudiu-vmw: I have the impression you only need PsaProgramStructure.
Can you try to move that to a separate header file and only use that?
Let me know if you have trouble doing that.

backends/ebpf/psa/README.md
```diff
@@ -0,0 +1,233 @@
...
+
+# Getting started
+
+## Installation 
+
+Follow standard steps for the P4 compiler to install the eBPF backend with the PSA support.
+
+## Using PSA-eBPF
+
+**Note!** The PSA implemented for eBPF backend is verified to work with the kernel version 5.8+.
```
rst0git, line 128: It might be useful to mention the CPU architecture as well (e.g., x86?)

backends/ebpf/psa/README.md
```diff
@@ -0,0 +1,233 @@
...
+# Prerequisites
+
+- Refer to the [Cilium docs](https://docs.cilium.io/en/latest/bpf/) to learn more about eBPF.
+- This guide assumes at least basic familiarity with [the PSA specification](https://p4.org/p4-spec/docs/PSA.html).
+- The PSA implementation inherits some mechanisms (e.g. generation of Parser and Control block) from `ebpf_model`. Please, get familiar with
+[the base eBPF backend](../README.md) first.
+
+# Design
+
+P4 packet processing is translated into a set of eBPF programs attached to the TC hook. The eBPF programs implements packet processing defined
```
rst0git, line 14: *The eBPF programs implement ...

backends/ebpf/CMakeLists.txt
osinstom: @mbudiu-vmw I've spent some time trying to refactor it a bit. The problem is that `PsaProgramStructure` contains `createX` functions (e.g., `createParsers`, etc.), which depends on other objects like `ActionConverter`, etc. I've not found a way to re-organize code structure to minimize dependencies. If you see a better way to make PSA objects re-usable, I'll be happy to apply it to PSA/eBPF backend, once ready. 

So far, it seems to be the only way to re-use PSA objects. In fact, we based on DPDK backend that used the same approach in the past: https://github.com/p4lang/p4c/blob/b8894efd1481fe83c03d61b674a596052e509244/backends/dpdk/CMakeLists.txt#L98.

mbudiu-vmw: Let me take a look at this.

rst0git: > The problem is that PsaProgramStructure contains createX functions (e.g., createParsers, etc.), which depends on other objects like ActionConverter, etc. 

@osinstom maybe we can move the common PSA code in `backends/common` to make it reusable?

osinstom:
@mbudiu-vmw @rst0git I've synced with master and minimized build dependencies.

mbudiu-vmw:
So I guess this is now ready for review?

osinstom:
> So I guess this is now ready for review?

I think so

backends/ebpf/psa/ebpfPipeline.cpp
```diff
@@ -0,0 +1,607 @@
...
+    builder->newline();
+    emitHeadersFromCPUMAP(builder);
+    builder->newline();
+    emitMetadataFromCPUMAP(builder);
+    builder->newline();
+
+    emitPSAControlOutputMetadata(builder);
+    emitPSAControlInputMetadata(builder);
+
+    msgStr = Util::printf_format("%s parser: parsing new packet, path=%%d, pkt_len=%%d",
```
mbudiu-vmw, line 395: shouldn't this code be in the parser code generator?

backends/ebpf/psa/ebpfPsaArch.cpp
```diff
@@ -0,0 +1,637 @@
...
+    builder->appendLine("#include <stdbool.h>");
+    builder->appendLine("#include <linux/if_ether.h>");
+    builder->appendLine("#include \"psa.h\"");
+}
+
+void PSAArch::emitPreamble(CodeBuilder *builder) const {
+    emitCommonPreamble(builder);
+    builder->newline();
+
+    builder->appendLine("#define CLONE_MAX_PORTS 64");
```
mbudiu-vmw, line 37: This kind of stuff could also be in a fixed header file. But perhaps it's better to have fewer of these hanging around?

backends/ebpf/psa/ebpfPsaArch.cpp
```diff
@@ -0,0 +1,637 @@
...
+
+    /*
+     * 5. BPF map definitions.
+     */
+    emitInstances(builder);
+
+    /*
+     * 6. BPF map initialization
+     */
+    emitInitializer(builder);
```
mbudiu-vmw, line 311: interesting, I expected that this is done at runtime. I wonder how bpf does it, since maps do not exist until the program is started.

backends/ebpf/psa/ebpfPsaArch.cpp
```diff
@@ -0,0 +1,637 @@
...
+    return true;
+}
+
+// =====================EBPFParser=============================
+bool ConvertToEBPFParserPSA::preorder(const IR::ParserBlock *prsr) {
+    auto pl = prsr->container->type->applyParams;
+
+    parser = new EBPFPsaParser(program, prsr, typemap);
+
+    auto it = pl->parameters.begin();
```
mbudiu-vmw, line 470: check?

backends/ebpf/psa/ebpfPsaArch.cpp
```diff
@@ -0,0 +1,637 @@
...
+                            "%1%: only one LPM field allowed", it->matchType);
+                    return false;
+                }
+                tableKind = TableLPMTrie;
+            }
+        }
+    }
+
+    // use 1024 by default
+    size_t size = 1024;
```
mbudiu-vmw, line 561: hidden like this it will be hard to find and change. Perhaps it should also be a compiler option.

backends/ebpf/psa/ebpfPsaArch.cpp
```diff
@@ -0,0 +1,637 @@
...
+
+    EBPFTablePSA *table = new EBPFTablePSA(program, tblblk, control->codeGen, name, size);
+
+    control->tables.emplace(tblblk->container->name, table);
+    return true;
+}
+
+bool ConvertToEBPFControlPSA::preorder(const IR::AssignmentStatement *a) {
+    // the condition covers both ingress and egress timestamp
+    if (a->right->toString().endsWith("timestamp")) {
```
mbudiu-vmw, line 579: This is somewhat ugly.

mbudiu-vmw (changes requested):

mbudiu-vmw:
Please make any changes as new commits, I can't read this much code again.

backends/ebpf/psa/README.md
```diff
@@ -0,0 +1,233 @@
...
+  The Ingress pipeline is composed of Parser, Control block and Deparser. The details of Parser, Control block and Deparser implementation
+  will be explained further in this document. The same eBPF program in TC contains also the Traffic Manager.
+  The role of Traffic Manager is to redirect traffic between the Ingress (TC) and Egress (TC).
+  It is also responsible for packet replication via clone sessions or multicast groups and sending packet to CPU.
+- `tc-egress` - The PSA Egress pipeline (composed of Parser, Control block and Deparser) is attached to the TC Egress hook. As there is no
+  XDP hook in the Egress path, the use of TC is mandatory for the egress processing.
+  
+## Packet paths
+
+## NFP (Normal Packet From Port)
```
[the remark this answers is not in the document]
osinstom: This comes from PSA specs.

backends/ebpf/psa/README.md
```diff
@@ -0,0 +1,233 @@
...
+
+## Planned features
+
+All the below features are already implemented and will be contributed to the P4 compiler in subsequent pull requests.
+
+- **XDP support.** The current version of P4-eBPF compiler leverages the BPF TC hook for P4-programmable packet processing. 
+The TC subsystem enables implementation of the full PSA specification, contrary to XDP, but offers lower throughput. We're going to 
+contribute the XDP-based version of P4-eBPF that is not fully spec-compliant, but provides higher throughput.
+- **XDP2TC mode.** the generated eBPF programs use `bpf_xdp_adjust_meta()` to transfer original EtherType from XDP to TC. This mode 
+may not be supported by some NIC drivers. We will add two other modes that can be used alternatively.
```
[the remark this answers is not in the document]
osinstom: Well.. auto-detection might be tricky, since the P4 compiler can be run on a different machine than the one where the eBPF program is actually deployed. The XDP2TC mode will be configured with compiler flags.

backends/ebpf/psa/ebpfPsaParser.cpp
```diff
@@ -0,0 +1,125 @@
...
+    builder->endOfStatement(true);
+    builder->blockEnd(true);
+}
+
+EBPFPsaParser::EBPFPsaParser(const EBPFProgram* program, const IR::ParserBlock* block,
+                             const P4::TypeMap* typeMap) : EBPFParser(program, block, typeMap) {
+    visitor = new PsaStateTranslationVisitor(program->refMap, program->typeMap, this);
+}
+
+void EBPFPsaParser::emitRejectState(CodeBuilder* builder) {
```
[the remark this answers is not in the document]
osinstom: [emitRejectState is actually in the base class.](https://github.com/p4lang/p4c/pull/3139/files/b9c6e7d28dded851058d6019dd358f1730b5d0ef#diff-9cff861a646f282e515c120ffd35927508787fca57d7f949afc2d8bf207cd2dcR609) But there is a difference between how reject state should be handled in PSA (according to PSA specs) vs. eBPF model.

backends/ebpf/psa/ebpfPsaDeparser.cpp
```diff
@@ -0,0 +1,121 @@
...
+    builder->appendFormat("if (%s->drop) ", istd->name.name);
+    builder->blockStart();
+    builder->target->emitTraceMessage(builder, "PreDeparser: dropping packet..");
+    builder->emitIndent();
+    builder->appendFormat("return %s;\n", builder->target->abortReturnCode().c_str());
+    builder->blockEnd(true);
+
+    // if packet should be resubmitted, we skip deparser
+    builder->emitIndent();
+    builder->appendFormat("if (%s->resubmit) ", istd->name.name);
```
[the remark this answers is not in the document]
osinstom: Yes, the ingress deparser logic is implemented according to https://p4.org/p4-spec/docs/PSA.html#sec-after-ingress.

backends/ebpf/psa/ebpfPsaArch.cpp
osinstom: Do you mean "loaded" by "started"? This initialization function is actually called by `psabpf pipeline load`. I'll add a section to documentation about this.

backends/ebpf/psa/README.md
```diff
@@ -0,0 +1,233 @@
...
+
+The purpose of `RECIRCULATE` is to transfer packet processing back from the Egress Deparser to the Ingress Parser.
+
+In order to implement `RECIRCULATE` we assume the existence of `PSA_PORT_RECIRCULATE` ports. Therefore, packet recirculation is simply performed by
+invoking `bpf_redirect()` to the `PSA_PORT_RECIRCULATE` port with `BPF_F_INGRESS` flag to enforce processing a packet by the Ingress pipeline.
+
+## Metadata
+
+There are some global metadata defined for the PSA architecture. For example, `packet_path` must be shared among different pipelines.
+To share a global metadata between pipelines we use `skb->cb` (control buffer), which gives us 20B that are free to use.
```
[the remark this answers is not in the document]
osinstom: We would need to design some other approach, but it's not likely that it becomes not enough. Actually, we only need to share `packet_path` between pipelines.

backends/ebpf/psa/README.md
```diff
@@ -0,0 +1,233 @@
...
+Refer to [the bpftool guide](https://manpages.ubuntu.com/manpages/focal/man8/bpftool-prog.8.html) for more examples how to use it.
+
+# TODO / Limitations
+
+We list the known bugs/limitations below. Refer to the Roadmap section for features planned in the near future.
+
+- Larger bit fields (e.g. IPv6 addresses) may not work properly
+- We noticed that `bpf_xdp_adjust_meta()` isn't implemented by some NIC drivers, so the PSA impl may not work 
+with some NICs. So far, we have verified the correct behavior with Intel 82599ES.
+- `lookahead()` with bit fields (e.g., `bit<16>`) doesn't work.
```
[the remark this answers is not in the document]
osinstom: We can verify that and fix in a separate PR.

backends/ebpf/psa/xdpHelpProgram.h
```diff
@@ -0,0 +1,76 @@
...
+limitations under the License.
+*/
+#ifndef BACKENDS_EBPF_PSA_XDPHELPPROGRAM_H_
+#define BACKENDS_EBPF_PSA_XDPHELPPROGRAM_H_
+
+#include "backends/ebpf/ebpfProgram.h"
+
+namespace EBPF {
+
+class XDPHelpProgram : public EBPFProgram {
```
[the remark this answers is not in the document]
osinstom: I'm not sure.. we would like to minimize C dependencies that a user needs to compile a generated C file to eBPF bytecode. Actually, we would like to follow the other direction with `psa.h` and auto-generate PSA standard fields. What's the better approach in your opinion?

backends/ebpf/psa/ebpfPipeline.cpp
osinstom: We need `lengthVar`, `sectionName` from `ebpfProgram` and `compilerGlobalMetadata` from `ebpfPipeline`. So, I suggest keeping it here.

backends/ebpf/psa/xdpHelpProgram.h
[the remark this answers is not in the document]
mbudiu-vmw: There's something to be said for having this inside the compiler, since you no longer worry about inconsistent versions and installation directories.

backends/ebpf/psa/ebpfPipeline.h
```diff
@@ -0,0 +1,186 @@
...
+class EBPFEgressPipeline : public EBPFPipeline {
+ public:
+    EBPFEgressPipeline(cstring name, const EbpfOptions& options, P4::ReferenceMap* refMap,
+                       P4::TypeMap* typeMap) : EBPFPipeline(name, options, refMap, typeMap) { }
+
+    virtual void emitGetSharedHeaders(CodeBuilder* builder) {
+        (void) builder;
+        // this method should never be called.
+        ::error(ErrorType::ERR_UNSUPPORTED,
+                "We only implement pipeline optimization for XDP so far.");
```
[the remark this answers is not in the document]
osinstom: It's a leftover, to be removed

backends/ebpf/psa/xdpHelpProgram.h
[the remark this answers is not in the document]
osinstom: If you don't mind I would keep this as is and auto-generate `psa.h` in future as well.

backends/ebpf/psa/ebpfPipeline.cpp
```diff
@@ -0,0 +1,607 @@
...
+        builder->emitIndent();
+        builder->appendFormat("struct psa_ingress_input_metadata_t %s = {\n"
+                              "            .ingress_port = %s,\n"
+                              "            .packet_path = %s,\n"
+                              "            .parser_error = %s,\n"
+                              "    };",
+                              control->inputStandardMetadata->name.name,
+                              ifindexVar.c_str(), packetPathVar.c_str(), errorVar.c_str());
+        builder->newline();
+        if (shouldEmitTimestamp()) {
```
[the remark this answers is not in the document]
osinstom: Not really, this can be auto-discovered by the compiler.

backends/ebpf/psa/ebpfPsaArch.cpp
osinstom: Exactly. This could probably be configurable with compiler options in future. I'll add a TODO.

backends/ebpf/psa/ebpfPsaArch.cpp
osinstom: Yeah, I agree it is. The goal is to find out if `ingress_timestamp` or `egress_timestamp` is used in a control block. We haven't found a better way to do this so far.

backends/ebpf/psa/README.md
```diff
@@ -0,0 +1,233 @@
...
+
+P4 packet processing is translated into a set of eBPF programs attached to the TC hook. The eBPF programs implement packet processing defined
+in a P4 program written according to the PSA model. The TC hook is used as a main engine, because it enables a full implementation of the PSA specification.
+We also plan to contribute the XDP-based version of the PSA implementation that does not implement the full specification, but provides better performance.
+
+The TC-based design of PSA for eBPF is depicted in Figure below.
+
+![TC-based PSA-eBPF design](psa-ebpf-design.png)
+
+- `xdp-helper` - the fixed, non-programmable "helper" program attached to the XDP hook. The role of the `xdp-helper` program is to prepare
```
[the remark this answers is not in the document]
osinstom: Why? It may be easily confused with the TC Ingress/Egress hooks coming from eBPF terminology.

backends/ebpf/ebpfParser.cpp
```diff
@@ -351,14 +331,33 @@ StateTranslationVisitor::compileExtract(const IR::Expression* destination) {
...
                                             program->offsetVar, cstring::to_cstring(width));
+
     builder->target->emitTraceMessage(builder, "Parser: check pkt_len=%d >= last_read_byte=%d",
                                       2, program->lengthVar.c_str(), offsetStr.c_str());
 
+    // to load some fields the compiler will use larger words
+    // than actual width of a field (e.g. 48-bit field loaded using load_dword())
+    // we must ensure that the larger word is not outside of packet buffer.
+    // FIXME: this can fail if a packet does not contain additional payload after header.
+    //  However, we don't have better solution in case of using load_X functions to parse packet.
```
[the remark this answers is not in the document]
osinstom: This would probably require a significant refactoring, but I've added TODO to consider using a collection of smaller widths.

backends/ebpf/ebpfParser.cpp
```diff
@@ -465,6 +497,24 @@ bool StateTranslationVisitor::preorder(const IR::MethodCallExpression* expressio
     return false;
 }
 
+bool StateTranslationVisitor::preorder(const IR::StructExpression *expr) {
```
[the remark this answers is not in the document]
osinstom: Yes, it can. I moved it there.

backends/ebpf/psa/ebpfPsaTable.cpp
```diff
@@ -0,0 +1,288 @@
...
+    // TODO: placeholder for handling psa_implementation
+    EBPFTable::emitAction(builder, valueName, actionRunVariable);
+}
+
+void EBPFTablePSA::emitInitializer(CodeBuilder *builder) {
+    this->emitDefaultActionInitializer(builder);
+    this->emitConstEntriesInitializer(builder);
+}
+
+void EBPFTablePSA::emitConstEntriesInitializer(CodeBuilder *builder) {
```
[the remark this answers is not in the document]
osinstom: Actually, I believe the classic eBPF backend has already support for const entries: 

https://github.com/p4lang/p4c/blob/main/backends/ebpf/ebpfTable.cpp#L566. 

However, there is a difference how the control-plane API look like for these two architectures (e.g., `ebpf_model` uses `BPF_OBJ_GET`), so that it cannot be re-used.

Yes, it's executed by the control plane. See Control-plane API section from the updated README.

backends/ebpf/psa/ebpfPsaDeparser.cpp
```diff
@@ -0,0 +1,121 @@
...
+
+DeparserBodyTranslatorPSA::DeparserBodyTranslatorPSA(const EBPFDeparserPSA *deparser) :
+        CodeGenInspector(deparser->program->refMap, deparser->program->typeMap),
+        DeparserBodyTranslator(deparser) {
+    setName("DeparserBodyTranslatorPSA");
+}
+
+void DeparserBodyTranslatorPSA::processFunction(const P4::ExternFunction *function) {
+    auto dprs = dynamic_cast<const EBPFDeparserPSA*>(deparser);
+    if (function->method->name.name == "psa_resubmit") {
```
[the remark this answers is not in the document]
osinstom: I don't get it, what expression? Please provide code suggestion to facilitate the process.

backends/ebpf/ebpfTable.h
```diff
@@ -117,6 +118,7 @@ class EBPFTable : public EBPFTableBase {
     // Whether to drop packet if no match entry found.
     // Some table implementations may want to continue processing.
     virtual bool dropOnNoMatchingEntryFound() const { return true; }
+    virtual bool singleActionRun() const { return true; }
```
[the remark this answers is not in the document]
osinstom: Actually, this is a leftover and is not needed anymore.

osinstom:
@mbudiu-vmw Thanks for review! I believe I've address all of your comments.

backends/ebpf/psa/ebpfPsaDeparser.cpp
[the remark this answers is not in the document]
mbudiu-vmw: "(!%s->drop && %s->resubmit)"

backends/ebpf/psa/ebpfPsaGen.cpp
```diff
@@ -0,0 +1,641 @@
...
+
+    EBPFTablePSA *table = new EBPFTablePSA(program, tblblk, control->codeGen);
+
+    control->tables.emplace(tblblk->container->name, table);
+    return true;
+}
+
+bool ConvertToEBPFControlPSA::preorder(const IR::AssignmentStatement *a) {
+    // the condition covers both ingress and egress timestamp
+    if (a->right->toString().endsWith("timestamp")) {
```
mbudiu-vmw, line 583: A more robust way is to look for a Member named "timestamp" from a type that is like the metadata. Or at least a member...

backends/ebpf/psa/README.md
```diff
@@ -0,0 +1,264 @@
...
+does another lookup to the inner map to find all members.
+
+## CE2E (Clone Egress to Egress)
+
+CE2E refers to process of copying a packet that was handled by the Egress pipeline and resubmitting the cloned packet to the Egress Parser.
+
+CE2E is implemented by invoking `bpf_clone_redirect()` helper in the Egress path. Output ports are determined based on the
+`clone_session_id` and lookup to "clone_session" BPF map, which is shared among TC ingress and egress (eBPF subsystem allows for map sharing between programs).
+
+## Sending packet to CPU
```
rst0git, line 92: The subtitle "Sending packet to CPU" is slightly confusing in the context of eBPF because everything is running on CPU. Perhaps it would be more appropriate to change the subtitle to "Sending packet to Control Plane"?

@jafingerhut What do you think?

jafingerhut: The current PSA specification might be contributing to some confusion here, since that spec does refer to a "CPU port" to send packets from the data plane to the controller.  That specification does have at least several places with a kind of built-in assumption that the P4 data plane is separate from a nearby control CPU, at least in how it names things.

I'm not sure if there are only a few small surgical changes in the PSA spec that would match your suggestion here, but unfortunately if we want to avoid changing names in the P4 include file for the architecture, there are already a few names like these in there (https://github.com/p4lang/p4c/blob/main/p4include/bmv2/psa.p4):

```
const PortId_t PSA_PORT_CPU = (PortId_t) 0xfffffffd;

const CloneSessionId_t PSA_CLONE_SESSION_TO_CPU = (CloneSessionId_t) 0;
```

osinstom:
@mbudiu-vmw I've addressed your last comments. Please resolve conversations/let me know what else I need to address.

backends/ebpf/psa/ebpfPsaGen.cpp
```diff
@@ -579,9 +579,11 @@ bool ConvertToEBPFControlPSA::preorder(const IR::TableBlock *tblblk) {
 }
 
 bool ConvertToEBPFControlPSA::preorder(const IR::AssignmentStatement *a) {
-    // the condition covers both ingress and egress timestamp
-    if (a->right->toString().endsWith("timestamp")) {
-        control->timestampIsUsed = true;
+    if (auto m = a->right->to<IR::Member>()) {
```
mbudiu-vmw, line 582: I think you can do even better: just use the IR::Member visitor function and check there.

mbudiu-vmw (approved):

mbudiu-vmw:
Will give you a chance to address the last two small comments.

PR #3148, opened 2022-03-24
Compile the P4 compiler tools in host configuration

chrisdolan:
There's no reason I'm aware of to compile the compilers for the target architecture.

As a side benefit, this may improve the build system's cache hit rate slightly because these tools don't need to be recompiled for Tsan/Asan/Msan/etc let alone x86 vs. ARM vs. etc.

Signed-off-by: Chris Dolan <chrisdolan@google.com>

### DIFF
--- a/bazel/p4_library.bzl
+++ b/bazel/p4_library.bzl
@@ -143,7 +143,7 @@ p4_library = rule(
         "_p4c": attr.label(
             default = Label("@com_github_p4lang_p4c//:p4c_bmv2"),
             executable = True,
-            cfg = "target",
+            cfg = "host",
         ),
         "_p4include": attr.label(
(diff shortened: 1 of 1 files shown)


mbudiu-vmw (approved):
I know nothing about Bazel but I see no problem approving this

chrisdolan:
@mbudiu-vmw I'm happy to explain it further if it helps. But yes, it's pretty trivial. I tested on our copy with Clang's undefined-behavior sanitizer configured for the build -- without this change, the compiler itself is compiled with UBsan (incorrectly) and crashes but with this change the compiler is compiled optimized and runs normally.

smolkaj (approved):
LGTM, thanks Chris!

PR #3150, opened 2022-03-24
Add old parser state in loopsUnroll if ExpressionEvaluator returns a error

VolodymyrPeschanenkoIntel:
These changes allow to save parser state for case where ExpressionEvaluator returns an error, otherwise this state could be eliminated from the result.
To repeat this problem, please, run the following command:
`./p4test -I . /p4include --target bmv2 --std p4-16 --loopsUnroll --arch v1model --dump 11 --top4 End ../testdata/p4_16_samples/issue314.p4`
The parser doesn't have start state after ParserUnroll without these changes.

### DIFF
--- a/midend/parserUnroll.cpp
+++ b/midend/parserUnroll.cpp
@@ -656,6 +656,7 @@ class ParserSymbolicInterpreter {
                 // don't evaluate successors anymore
                 continue;
             }
+            bool notAdded = newStates.count(getNewName(stateInfo)) == 0;
             auto nextStates = evaluateState(stateInfo, newStates);
             if (nextStates.first == nullptr) {
                 if (nextStates.second && stateInfo->predecessor &&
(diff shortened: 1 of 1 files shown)


midend/parserUnroll.cpp
```diff
@@ -666,6 +667,11 @@ class ParserSymbolicInterpreter {
                         IR::IndexedVector<IR::StatOrDecl>(),
                         new IR::PathExpression(new IR::Type_State(),
                             new IR::Path(outOfBoundsStateName, false)));
+                } else {
```
mbudiu-vmw, line 670: why is this "else" needed?
I could argue that the program for which this "bug" happens is certainly incorrect, because it will always produce a runtime error in the start state. Are there any "correct" programs where this bug is present?

VolodymyrPeschanenkoIntel: I think that this bug appear if program is incorrect (like issue314.p4), but even in that case parserUnroll can't eliminate states from a parser.

mbudiu-vmw: This is not really an explanation. My question is "why don't you do the addition in the "if" branch as well.

VolodymyrPeschanenkoIntel: What kind of addition you mean? evaluateState returns nullptr in the first argument if some error appeared during execution. `If body` corresponds to the generation of a `OutOfBound` states and `else body` corresponds to all other cases. `notAdded` flag is used to add old state once only.

mbudiu-vmw (approved):

PR #3152, opened 2022-03-25
typeChecker: make error message less confusing

mvido:
Same error message was used for all struct-like types:
* IR::Type_UnknownStruct
* IR::Type_Struct
* IR::Type_Header
* IR::Type_HeaderUnion

In case of a header or header_union it could result in a message
like this:
"""
Field my_field is not a member of structure header my_header_h
"""
which might have been confusing for a user (why is error message
mentioning structure and header at one place?).

The word "structure" is now omitted in case of IR::Type_Header
and IR::Type_HeaderUnion.
String representation of IR::Type_Header and IR::Type_HeaderUnion
nodes used in this error message already contains also header or
header_union keyword.


jnfoster (approved):
Nice! LGTM.

mbudiu-vmw (approved):

Issue #2623, opened 2020-12-01
Compilers allows instantiations with don't care types.


mbudiu-vmw:
The following program should be rejected:
```
#define V1MODEL_VERSION 20201000
#include <core.p4>
#include <v1model.p4>

struct Headers { }
struct Metadata { }
parser parse(packet_in p, out Headers h, inout Metadata m,
inout standard_metadata_t sm) {
    state start { transition accept; } }
control verifyChecksum(inout Headers h, inout Metadata m) { apply { } }
control egress(inout Headers h, inout Metadata m,
inout standard_metadata_t sm) { apply { } }
control computeChecksum(inout Headers h, inout Metadata m) { apply { } }
control deparse(packet_out p, in Headers h) { apply { } }
control ingress(inout Headers h, inout Metadata m,
inout standard_metadata_t sm) {
    bit<8> index = 0;
    bit<8> result = 0;
    bit<16> index1 = 10;
    meter<_>(1024, MeterType.bytes) mtr;  // Use `result` argument to infer type.
    apply {
        mtr.execute_meter(index, result);
        mtr.execute_meter(index1, result);
    }
}
V1Switch(parse(), verifyChecksum(), ingress(), egress(),
computeChecksum(), deparse()) main;
```

PR #3154, opened 2022-03-29
Compilers allows instantiations with don't care types. #2623

strahinjasyrmia:
I think that solution for this problem could be to make the compiler to restrict instantiation of extern template types that doesn't use template parameters in their constructor (I'm referencing on specification : "The “don't care” identifier (_) can only be used for an out function/method argument, when the value of returned in that argument is ignored by subsequent computations. **When used in generic functions or methods, the compiler may reject the program if it is unable to infer a type for the don't care argument. [pg.81]**").
If the template parameter is used somewhere between constructor parameters, than we can allow instantiation because we can infer type of the template parameter based on the type of the constructor argument.


frontends/p4/typeChecking/typeChecker.cpp
```diff
@@ -1264,6 +1264,37 @@ const IR::Node* TypeInference::postorder(IR::Type_Package* decl) {
...
+    // If there is no template_parameter in constructor parameters
+    // then compiler should reject instantiation
+    // because it cannot infer the type of template_parameter
+    for (auto arg : *(type->arguments)) {
+        if (arg->is<IR::Type_Dontcare>()) {
+            auto decl = refMap->getDeclaration(type->baseType->path);
+            if (auto extern_decl = decl->to<IR::Type_Extern>()) {
+                for (auto template_parameter : extern_decl->typeParameters->parameters) {
+                    for (auto meth : extern_decl->methods) {
+                        if (meth->type->returnType == nullptr) {
```
mbudiu-vmw, line 1276: An extern can have multiple constructors. You stop at the first one. It may not be the right one.

frontends/p4/typeChecking/typeChecker.cpp
```diff
@@ -1264,6 +1264,37 @@ const IR::Node* TypeInference::postorder(IR::Type_Package* decl) {
...
+                            bool found = false;
+                            for (auto meth_params : *(meth->getParameters())) {
+                                if (template_parameter->getP4Type()->equiv
+                                    (*(meth_params->type->getP4Type()))) {
```
mbudiu-vmw, line 1280: Why do the types need to be equivalent?
One of the types could have references to the other one, e.g., Generic<T>, where Generic is a generic type.

mbudiu-vmw (changes requested):

frontends/p4/typeChecking/typeChecker.cpp
strahinjasyrmia: Working on fix. Thanks!

frontends/p4/typeChecking/typeChecker.cpp
strahinjasyrmia: The idea was to check if the template parameter is used somewhere between constructor parameters. 

The following code should be accepted (type can be infered from constructor parameter - but now I see that this isn't supported in the compiler yet - I might have to add that too) : 
```
extern E<T> {
    E(T t);
    void f(in T arg);
}
control c() {
    E<_>(0) e; 
    apply {
        e.f(0);
    }
} 
```
And the next one shouldn't (because function _f_ could be called with parameter of any type, for the same object e (just like with meters in original issue).: 
```
extern E<T> {
    E();
    void f(in T arg);
}
control c() {
    E<_>() e;       //rejected here
    apply {
        e.f(0);
        e.f("a"); 
    }
}
```

jnfoster: For the first example, what type is used for the instantiation of `T`? If we knew the type of `0`, I'd agree, but the width has to be inferred too.

For the second example, I agree it shouldn't type check. But I'm not sure I agree with the sentence "function f could be called with parameter of any type" if you mean method `f` is polymorphic. It's not, because the type parameter is on the `extern` and not on the `f` method. So rather than thinking that "`f` could be called with parameter of any type," I would say it as "extern `E`'s type variable could be instantiated with any type, then method `f` would take an argument of the same type." Hope that makes sense?

strahinjasyrmia: I printed the type of constructor argument and it was `int` for all integer arguments. `bit<N>` when width is specified, `Type_String` for strings etc...

```
extern E<T> {
    E(T t);
    void f(in T arg);
}
control c() {
    E<_>(0) e;      // Type: int
 // E<_>(8w10) e;   // Type: bit<8>
 // E<_>("a") e;    // Type: Type_String
    apply {
        e.f(0);
    }
} 
control proto();
package top(proto p);

top(c()) main;
```

I might have chosen the bad words in the explanation for second example. But I think we are on the same page - the type of argument that is used in constructor must be the only type that is used in method call.

PR #3158, opened 2022-03-30
Add PTF test infra for eBPF backend

osinstom:
This PR introduces the PTF test infra for eBPF backend and adds a pre-merge job to run PTF tests. 

The initial set of PTF tests verifies the basic functionalities and packet paths implemented for PSA-eBPF. 

This is another PR from series introducing PSA support for eBPF backend. This PR depends on #3139.


backends/ebpf/tests/test.sh
```diff
@@ -0,0 +1,177 @@
...
+
+# Remove any temporary files from previous run. It might be useful to
+# preserve these files after test run for inspection
+rm -rf ptf_out/*
+echo "Removed old temporary files from previous tests"
+
+# Trace all command from this point
+set -x
+
+declare -a INTERFACES=("eth0" "eth1" "eth2" "eth3" "eth4" "eth5")
```
rst0git, line 89: @osinstom I'm not sure if all these interfaces would be available in the VMs used by GitHub actions.

Perhaps we could use something like test environment from the XDP Tutorial:
https://github.com/xdp-project/xdp-tutorial/tree/master/testenv

osinstom: I believe that our setup is very similar to the test env from the XDP tutorial, we just use different names.

backends/ebpf/tests/ptf/common.py
```diff
@@ -0,0 +1,229 @@
+# Copyright 2022-present Orange
```
mbudiu-vmw, line 1: I find it useful to have some documentation about what this file is doing.

backends/ebpf/tests/ptf/common.py
```diff
@@ -0,0 +1,229 @@
...
+        p4args = "--Wdisable=unused"
+        if self.is_trace_logs_enabled():
+            p4args += " --trace"
+
+        logger.info("P4ARGS=" + p4args)
+        self.exec_cmd("make -f ../runtime/kernel.mk BPFOBJ={output} P4FILE={p4file} "
+                      "ARGS=\"{cargs}\" P4C=p4c-ebpf P4ARGS=\"{p4args}\" psa".format(
+                            output=self.test_prog_image,
+                            p4file=self.p4_file_path,
+                            cargs="-DPSA_PORT_RECIRCULATE=2",
```
mbudiu-vmw, line 76: since this is defined here perhaps you can use the macro in the P4 code.

backends/ebpf/tests/test.sh
```diff
@@ -0,0 +1,177 @@
...
+      # unknown option
+      ;;
+  esac
+done
+
+cleanup
+trap cleanup EXIT
+
+# Remove any temporary files from previous run. It might be useful to
+# preserve these files after test run for inspection
```
mbudiu-vmw, line 82: yes, the p4c test scripts have an option -b to preserve the temporary files

mbudiu-vmw (approved):

mbudiu-vmw:
All your pdf tests failed with this error: `STDERR: make: *** No rule to make target 'psa'.  Stop.`

osinstom:
@mbudiu-vmw #3139 must be merged first.

backends/ebpf/tests/test.sh
osinstom: Yes, we preserve output files for failing tests.

backends/ebpf/tests/ptf/common.py
osinstom: I've check that, but if we define the macro in the P4 code it's not propagated down to the C code. We must pass it to `clang`. Also, hardcoding it in the P4 code would require modifying P4 programs each time the recirculate port number is changed (it might become dynamically allocated in future). 

Please also note that the packet recirculation support should be rather treated as an "experimental" feature. We'll create an issue to track this.

tools/ci-build.sh
```diff
@@ -67,6 +67,31 @@ cd /p4c
...
+                           libjansson-dev \
+                           linux-tools-`uname -r`"
+  # Package "linux-tools-generic-hwe-20.04" is not required because
+  # we test under current kernel, not the newest one
+  apt-get install -y --no-install-recommends ${P4C_PTF_PACKAGES}
+
+  git clone --recursive https://github.com/P4-Research/psabpf.git /tmp/psabpf
+  cd /tmp/psabpf
+  # FIXME: psabpf is under heavy development, later use git tags when it will be ready to use
+  git reset --hard 8fc9687
```
rst0git, line 82: Would it be possible to create a "stable" branch in the psabpf repository and use this branch to test p4c?

This would make it a lot easier to run the tests locally.

rst0git (approved):
LGTM

PR #3160, opened 2022-03-31
DPDK Backend: Insert table keys generated by compiler closer to the standard and user metadata

usha1830:
DPDK needs all the table keys to come from single structure and all the table keys should fit within 64 bytes inclusive of all holes in between.

For ex: The below program will fail to load onto DPDK pipeline because the gap between two table keys is more than 64 bytes.

```
struct metadata {
   bit<8> tmp1;
   ...
   ...
   // more fields summing up to >=64 bytes
   ...
   bit <8> tmp2;
}

control{
   table {
      key = {
              tmp1 : exact;
               tmp2 : exact;
         }
    …
   }
   …
}
```
This PR inserts the table keys generated by the compiler right after the standard and user metadata to keep them closer to the metadata fields which can possibly be the table keys in the user program.


backends/dpdk/dpdkAsmOpt.cpp
```diff
@@ -194,4 +208,65 @@ const IR::Node* RemoveUnusedMetadataFields::preorder(IR::DpdkAsmProgram *p) {
...
+            auto keyMem = key->expression->to<IR::Member>();
+            auto type = keyMem->expr->type;
+            if (type->is<IR::Type_Struct>() && isMetadataStruct(type->to<IR::Type_Struct>())) {
+                auto offset = metaStruct->getFieldBitOffset(keyMem->member.name);
+                if (min == -1 || min > offset)
+                    min = offset;
+                if (max == -1 || max < offset) {
+                    max = offset;
+                    auto field_type = key->expression->type;
+                    if (auto t = field_type->to<IR::Type_Bits>()) {
```
mbudiu-vmw, line 247: you may want to make this a utility function

mbudiu-vmw (approved):

backends/dpdk/dpdkAsmOpt.cpp
usha1830: Done.

usha1830:
@mbudiu-vmw Addressed the comments. Please merge if the changes look fine.

PR #3164, opened 2022-04-01
Add Digest extern to PSA/eBPF backend

osinstom:
This PR adds support for Digest extern to PSA/eBPF backend.

Digests are implemented using `BPF_MAP_TYPE_QUEUE`. An eBPF program generates a Digest by calling `bpf_map_push_elem`. A user space application can get a Digest by calling `psabpf-ctl digest get pipe` or `bpf_map_lookup_and_delete_elem` C API.

This is another PR from series implementing PSA support for eBPF backend.


backends/ebpf/psa/README.md
```diff
@@ -141,7 +141,17 @@ An inner map is a hash map storing clone session/multicast group members, accord
...
-a control plane must add new element to the inner map.
+a con1trol plane must add new element to the inner map.
+
+## PSA externs
+
+### Digest
+
+Digests are intended to carry a small piece of user-defined data from the data plane to a control plane.
+The PSA-eBPF compiler translates each Digest instance into `BPF_MAP_TYPE_QUEUE` that implements a FIFO queue.
+If a deparser triggers the `pack()` method, an eBPF program inserts data defined for a Digest into the BPF queue map using `bpf_map_push_elem`.
```
mbudiu-vmw, line 152: perhaps you can add a link to the relevant section in the psa specification

backends/ebpf/psa/README.md
```diff
@@ -141,7 +141,17 @@ An inner map is a hash map storing clone session/multicast group members, accord
...
 or `multicast_group` in a PSA program) and initialize an inner map. To add a new clone session/multicast group member,
-a control plane must add new element to the inner map.
+a con1trol plane must add new element to the inner map.
+
+## PSA externs
+
+### Digest
+
+Digests are intended to carry a small piece of user-defined data from the data plane to a control plane.
+The PSA-eBPF compiler translates each Digest instance into `BPF_MAP_TYPE_QUEUE` that implements a FIFO queue.
```
mbudiu-vmw, line 151: I guess you have a separate table to keep them strongly typed, instead of one table with a union?
This will lose the relative ordering of digests between different tables.

backends/ebpf/psa/ebpfPsaDeparser.cpp
```diff
@@ -33,6 +33,33 @@ void DeparserBodyTranslatorPSA::processFunction(const P4::ExternFunction *functi
...
+        builder->appendFormat("bpf_map_push_elem(&%s, &", digestMapName);
+        this->visit(arg);
+        builder->appendFormat(", BPF_EXIST)");
+        return;
+    }
+
+    DeparserBodyTranslator::processMethod(method);
+}
+
+void EBPFDeparserPSA::emitDigestInstances(CodeBuilder* builder) const {
```
mbudiu-vmw, line 52: I would name this function differently.
First, it is really about the control-plane API. Perhaps there should be a common prefix/suffix to indicate all functions that deal with control plane APIs. Second, it's really about creating the API which allows digest messages to be received. Perhaps something like "createDigestAPI"?

backends/ebpf/psa/ebpfPsaGen.cpp
```diff
@@ -632,11 +633,34 @@ bool ConvertToEBPFDeparserPSA::preorder(const IR::ControlBlock *ctrl) {
...
-        // TODO: placeholder for handling digests
+        findDigests(p4Control);
         this->visit(p4Control->body);
     }
 
     return false;
 }
 
+void ConvertToEBPFDeparserPSA::findDigests(const IR::P4Control *p4Control) {
+    // Digests are only at ingress
```
mbudiu-vmw, line 644: That's why you should probably give an error if you find one in egress.
I would check this later.
Also, perhaps this field should not be called "type"; we already have many types.

mbudiu-vmw (approved):

backends/ebpf/psa/README.md
osinstom: For a single Digest instance, there is a separate BPF map of type `BPF_MAP_TYPE_QUEUE`. No unions, etc. Just a straightforward mapping.

backends/ebpf/psa/ebpfPsaDeparser.cpp
osinstom: We use `emitTableInstances` for generating instances of BPF maps implementing P4 tables. Why would we need to change this convention for Digests?

mbudiu-vmw: I guess the way PSA implements digests makes this design choice reasonable.

backends/ebpf/psa/README.md
jafingerhut: There is no requirement in PSA that data plane -> controller ordering must be maintained across different instances of Digest externs.

osinstom:
@mbudiu-vmw what about this one? I've addressed/answered your review.

mbudiu-vmw:
As you can see, lately I have a lot of reviews to do, so I appreciate if you actively tell me when you are done. 
This way I don't have to poll everything myself, and also I don't review stuff which isn't ready.

osinstom:
Got it and understood, I'll be pinging you when I'm done with addressing a review!

PR #3165, opened 2022-04-01
Add support for Counter extern to PSA/eBPF backend

osinstom:
This PR adds the support for indirect Counter extern to PSA/eBPF backend.

We currently implement Counter as a BPF array map, so the index width must be less or equal to `bit<32>`. We're planning to use BPF hash maps for Counters with wider index type.

Similarly to `ebpf_model`, the PSA Counter implementation uses `__sync_fetch_and_add` to provide atomicity.

Co-authored-by: Jan Palimąka <jan.palimaka@orange.com>

### DIFF
--- a/backends/ebpf/psa/ebpfPsaTable.cpp
+++ b/backends/ebpf/psa/ebpfPsaTable.cpp
@@ -49,7 +49,7 @@ void ActionTranslationVisitorPSA::processMethod(const P4::ExternMethod* method) 
     auto declType = method->originalExternType;
     auto decl = method->object;
     BUG_CHECK(decl->is<IR::Declaration_Instance>(),
-              "Extern has not been declared");
+              "Extern has not been declared: %1%", decl);
     auto di = decl->to<IR::Declaration_Instance>();
     auto instanceName = EBPFObject::externalName(di);
 
--- a/backends/ebpf/psa/ebpfPsaTable.cpp
+++ b/backends/ebpf/psa/ebpfPsaTable.cpp
@@ -22,6 +22,52 @@ limitations under the License.
 
 namespace EBPF {
 
+// =====================ActionTranslationVisitorPSA=============================
+ActionTranslationVisitorPSA::ActionTranslationVisitorPSA(const EBPFProgram* program,
+                                                         cstring valueName) :
+        CodeGenInspector(program->refMap, program->typeMap),
+        ActionTranslationVisitor(valueName, program),
+        ControlBodyTranslatorPSA(program->to<EBPFPipeline>()->control) {}
+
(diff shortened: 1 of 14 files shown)


backends/ebpf/psa/ebpfPsaTable.cpp
```diff
@@ -22,6 +22,52 @@ limitations under the License.
...
+    else if (auto cast = expression->to<IR::Cast>())
+        return isActionParameter(cast->expr);
+    else
+        return false;
+}
+
+void ActionTranslationVisitorPSA::processMethod(const P4::ExternMethod* method) {
+    auto declType = method->originalExternType;
+    auto decl = method->object;
+    BUG_CHECK(decl->is<IR::Declaration_Instance>(),
```
mbudiu-vmw, line 51: I find it useful to print something about the IR node that causes the failure, otherwise you make debugging a bit harder. But I understand that probably this never failed so far.
`... declared: %1%", decl);`

backends/ebpf/psa/externs/ebpfPsaCounter.cpp
```diff
@@ -0,0 +1,255 @@
...
+    }
+
+    if (!di->type->is<IR::Type_Specialized>()) {
+        ::error(ErrorType::ERR_MODEL, "Missing specialization: %1%", di);
+        return;
+    }
+    auto ts = di->type->to<IR::Type_Specialized>();
+
+    if (ts->arguments->size() != 2) {
+        ::error(ErrorType::ERR_MODEL, "Expected 2 specialization types: %1%", ts);
```
mbudiu-vmw, line 39: I would say "expected a type specialized with two arguments"

mbudiu-vmw (approved):

jafingerhut:
Regarding this sentence in your comment: "We currently implement Counter as a BPF array map, so the index width must be less or equal to bit<32>. We're planning to use BPF hash maps for Counters with wider index type."

I won't stop you, of course, but It seems to me to have very low utility to support indirect _anything_ in P4 with an index wider than 32 bits.  I guess long term on a general purpose CPU someone might want to have more than 2^32 counters/meters/register-array-elements, but I would be surprised if that is an urgent desire.

jnfoster:
"32-bit indices ought to be enough for anybody." —@jafingerhut, March 2022.

osinstom:
> 

I agree, Andy. I would say that this is something to consider in future, but, definitely, it's not an urgent desire.

osinstom:
@mbudiu-vmw can we have it merged if approved? Other externs are in the queue waiting for this :)

mbudiu-vmw:
I will assume you have addressed the comments.

PR #3168, opened 2022-04-02
manage lifetime of Visitor::visited

vhavel:
All three `Visitor` implementations hold a data structure containing information about all visited IR nodes in the `visited` attribute. This attribute is a pointer to an object dynamically allocated in `init_apply`, which is later set to `nullptr` when the visitor no longer needs it. The garbare collector is then supposed to destroy these objects.

This memory management mechanism doesn't work perfectly, however. In our backend implementation, managing lifetime of `visited` objects by refcounting  (as submitted in this PR) leads to  3x (!) peak memory usage decrease on large P4 programs. It seems the garbage collector isn't able to release the `visited` hashtable properly, leaving piles of IR nodes in the memory. This should also bring the codebase closer to proper memory management in ENABLE_GC=OFF builds, although this is only a tiny improvement in this direction.

I'm not sure on the exact reason why libgc doesn't release these objects properly. I've tried various versions including the latest 8.2 with similar results.

This behavior can be also reproduced on small P4 programs from this repository, for example
```
/usr/bin/time -f "%M" ./p4c-dpdk --arch psa ../backends/dpdk/examples/vxlan.p4
```
shows peak memory usage of 46969 bytes before the change, and 40392 after this change.

### DIFF
--- a/ir/visitor.cpp
+++ b/ir/visitor.cpp
@@ -180,15 +180,15 @@ Visitor::profile_t Visitor::init_apply(const IR::Node *root, const Context *pare
 }
 Visitor::profile_t Modifier::init_apply(const IR::Node *root) {
     auto rv = Visitor::init_apply(root);
-    visited = new ChangeTracker();
+    visited = std::make_shared<ChangeTracker>();
     return rv; }
 Visitor::profile_t Inspector::init_apply(const IR::Node *root) {
     auto rv = Visitor::init_apply(root);
-    visited = new visited_t();
+    visited = std::make_shared<visited_t>();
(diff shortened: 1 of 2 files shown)


mbudiu-vmw (approved):

mbudiu-vmw:
I have approved this, but it would be good for @ChrisDodd to review too, since they are using visitors in more complicated ways than the open-source part of the compiler does.

ChrisDodd (approved):
Yes, this should be fine -- since we replace global ::operator new/delete to go through the garbage collector, it should all work.

Issue #3143, opened 2022-03-22
Empty SymBitMatrix underflow


Staon:
If the `SymBitMatrix` is empty the automatic conversion to `bitvec` underflows the zero size and then spends much CPU time iterating from 4 billions to almost zero: 

https://github.com/p4lang/p4c/blob/fe4b558350aebac0fe1b6b9b01cbd076cf14a801/lib/symbitmatrix.h#L62:

```c++
operator bitvec() const {
  auto rv = self.getslice((row*row+row)/2, row+1);
  for (auto c = self.size()-1; c > row; --c)
     if (self(row, c)) rv[c] = 1;
  return rv; }
```

PR #3169, opened 2022-04-04
Fixed underflow during row-access in the SymBitMatrix (#3143)

Staon:
Fix of the #3143.

### DIFF
--- a/lib/symbitmatrix.h
+++ b/lib/symbitmatrix.h
@@ -56,12 +56,14 @@ class SymBitMatrix : private bitvec {
                 if (self.getrange((row*row+row)/2, row+1) != 0) return true;
             } else {
                 if (self.getslice((row*row+row)/2, row+1)) return true; }
-            for (auto c = self.size()-1; c > row; --c)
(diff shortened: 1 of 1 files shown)


lib/symbitmatrix.h
```diff
@@ -56,12 +56,14 @@ class SymBitMatrix : private bitvec {
                 if (self.getrange((row*row+row)/2, row+1) != 0) return true;
             } else {
                 if (self.getslice((row*row+row)/2, row+1)) return true; }
-            for (auto c = self.size()-1; c > row; --c)
+            const auto size = self.size();
```
mbudiu-vmw, line 59: do you know why the original iteration was downwards?

mbudiu-vmw (approved):

lib/symbitmatrix.h
Staon: Honestly, I don't. I guess it was some sort of optimization - the underlying vector allocated just once when it met first non-zero bit. Iterating upwards triggers standard schema of buffer doubling.

Issue #3172, opened 2022-04-04
Error message for already specialized type is misleading


apinski-cavium:
Take:
```
header t<tt>
{
  bit y;
}

typedef t ty;
typedef ty<bit> tty;
typedef tty<bit> tty1;
```
Currently we get:
```
t2.p4(8): [--Werror=type-error] error: tty<bit<1>>: Type header t<...> is not generic and thus it cannot be specialized using type arguments
typedef tty<bit> tty1;
        ^^^^^^^^
t2.p4(7)
typedef ty<bit> tty;
        ^^^^^^^
```
But this is misleading and should be improved saying tty is already specialized so it cannot be specialized using type arguments or something to that effect.

PR #3174, opened 2022-04-04
typedef with generic types requires type arguments

### LINKED ISSUE #3170: Error message for checking of types inside header does not say what the specialization is location is
Take:
```
header yy
{
  bit tt;
}
struct y
{
  yy[2] tt;
}
header t<tt>
{
  tt y;
}

typedef t ty;
typedef ty<y> tty;
```
Right now the compiler outputs:
```
t2.p4(11): [--Werror=type-error] error: Field 'y' of 'header t' cannot have type 'struct y'
  tt y;
     ^
t2.p4(5)
struct y
       ^
```
But it does not reference the last typedef line for the subsitution that had happened so it is not obvious where the problem is located.
I suspect this is an issue with most places of subsitutions too include functions.

mbudiu-vmw:

Fixes #3170

### DIFF
--- a/frontends/p4/typeChecking/typeChecker.cpp
+++ b/frontends/p4/typeChecking/typeChecker.cpp
@@ -1474,6 +1474,14 @@ const IR::Node* TypeInference::postorder(IR::Type_Typedef* tdecl) {
     auto type = getType(tdecl->type);
     if (type == nullptr)
         return tdecl;
+    BUG_CHECK(type->is<IR::Type_Type>(), "%1%: expected a TypeType", type);
+    auto stype = type->to<IR::Type_Type>()->type;
+    if (auto gen = stype->to<IR::IMayBeGenericType>()) {
+        if (gen->getTypeParameters()->size() != 0) {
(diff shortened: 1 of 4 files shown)


mbudiu-vmw:
Fixes #3172 
... by just forbidding the example.

apinski-cavium:
Is this even invalid because the specification is not clear here.

apinski-cavium:
Actually I think it should be invalid, though the specifications should be clear too.

jnfoster:
I agree this was a compiler bug. Can you point out where the spec is unclear? In the example program that illustrated the bug, I'd argue a generic without its type arguments isn't really a type.

mbudiu-vmw:
That generic is something like a function that takes some type arguments and returns a type. Not really a type, indeed.
I think that both behaviors are reasonable (the previous one, and the current one), but since no one asked for this feature I think it's safe to not allow it.

mbudiu-vmw:
So, can anyone approve this PR?

jnfoster (approved):

jnfoster:
My pleasure.

PR #3175, opened 2022-04-05
Shorten the Identifer name, including dots(.) in Member expression

kamleshbhalui:
to be less than 64 because dpdk does not allow Identifer longer than 63 chars

### DIFF
--- a/BUILD.bazel
+++ b/BUILD.bazel
@@ -145,6 +145,7 @@ filegroup(
     srcs = [
         "frontends/p4-14/ir-v1.def",
         "backends/bmv2/bmv2.def",
+        "backends/dpdk/dpdk.def",
         # p4c extensions may set this target to a `filegroup` containing
         # additional .def files.
         "@com_github_p4lang_p4c_extension//:ir_extensions",
@@ -190,6 +191,9 @@ cc_library(
         "frontends/parsers/p4/p4lexer.cc",
         "frontends/parsers/v1/v1lexer.cc",
         "ir/ir-generated.cpp",
+        "backends/dpdk/spec.cpp",
+        "backends/dpdk/dbprint-dpdk.cpp",
+        "backends/dpdk/printUtils.cpp",
     ],
     textual_hdrs = glob([
         "ir/**/*.h",
@@ -197,6 +201,7 @@ cc_library(
         "frontends/**/*.hpp",
         "midend/**/*.h",
         "control-plane/**/*.h",
+        "backends/dpdk/*.h",
     ]) + [
         ":p4_parser_yacc",
         ":v1_parser_yacc",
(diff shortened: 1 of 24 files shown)


backends/dpdk/dpdkAsmOpt.h
```diff
@@ -173,6 +173,125 @@ class ValidateTableKeys : public Inspector {
     int getFieldSizeBits(const IR::Type *field_type);
 };
 
+// This pass shorten the Identifier length
+class ShortenTokenLength : public Transform {
+    ordered_map<cstring, cstring> newNameMap;
+    ordered_map<cstring, cstring> &origNameMap;
+    static size_t count;
+    // Currently Dpdk allows Identifier of 63 char long or less
```
mbudiu-vmw, line 181: this is just a silly restriction, since I suspect the identifiers do not have a runtime representation.
wouldn't it be easier to fix this in the DPDK spec/implementation?
we're no longer in the 1970s when every byte counted.

backends/dpdk/dpdkAsmOpt.h
```diff
@@ -173,6 +173,125 @@ class ValidateTableKeys : public Inspector {
...
+    // including dots(.) for member exp.
+    // worst case member expression will look like below(for headers)
+    // 1.30.30 => 63(including dot(.))
+    cstring shortenString(cstring str, size_t allowedLength = 60) {
```
mbudiu-vmw, line 185: We already have a class which does this, MinimalNameGenerator.
You should reuse that code.

mbudiu-vmw (changes requested):

backends/dpdk/dpdkAsmOpt.h
kamleshbhalui: This runs post codegen and applied on some dpdk specific types too.

ir/ir.cpp
```diff
@@ -34,6 +34,8 @@ const cstring Type_Error::error = "error";
 
 int IR::Declaration::nextId = 0;
 int IR::This::nextId = 0;
+// used in dpdk backend
```
mbudiu-vmw, line 37: Let's move this to a separate file, dpdk.cpp.

mbudiu-vmw (changes requested):

BUILD.bazel
```diff
@@ -275,8 +280,71 @@ cc_binary(
     data = [":p4include"],
 )
 
-# This builds the p4test backend.
+# dpdk backend
```
mbudiu-vmw, line 283: perhaps for the bazel build you want to request a review from @smolkaj

mbudiu-vmw (approved):

BUILD.bazel
kamleshbhalui: I am unable to add @smolkaj to reviewer list. may be due to permission issues. Can you please add him?

mbudiu-vmw: That mention is enough, he had been notified. If he doesn't reply in a couple of days I will merge it anyway

backends/dpdk/dpdk.def
```diff
@@ -1,5 +1,6 @@
 interface IDPDKNode {
     virtual std::ostream& toSpec(std::ostream& out) const = 0;
+    static ordered_map<cstring, cstring> origNameMap;
```
hanw, line 3: Do you have to use a name map on every backend node? Can you move it to the ShortenTokenLength pass and apply that pass on the backend IR tree before codegen?

kamleshbhalui: done

BUILD.bazel
smolkaj: Got it, taking a look :)

BUILD.bazel
```diff
@@ -275,8 +280,71 @@ cc_binary(
...
+
+cc_proto_library(
+    name = "p4info_dpdk_cc_proto",
+    deps = [":p4info_dpdk_proto"],
+)
+
+cc_library(
+    name = "p4c_dpdk_lib",
+    srcs = glob(
+        ["backends/dpdk/*.cpp", "backends/dpdk/control-plane/*.cpp", "backends/bmv2/common/lower.cpp"],
```
smolkaj, line 309: Nit: maybe put each file on a separate line for consistency / readability?

smolkaj: Here and below

smolkaj (approved):
LGTM for the BUILD file changes

mbudiu-vmw:
I will merge this now, but please consider the suggested changes in a future PR. Thank you all.

PR #3176, opened 2022-04-05
Improving the driver re: the checking of input pathnames, improving error messages, trying to prevent misleading error messages.

Abe149:


mbudiu-vmw (commented):
I am actually not familiar with this piece of code, never wrote or read any part of it.
But 22 commits sounds a lot.

tools/driver/p4c_src/main.py
```diff
@@ -45,25 +47,40 @@ def display_supported_targets(cfg):
...
+    parser.add_argument(JSON_input_flag, dest="json_source", default=None,
+                        help="Use IR from JSON representation dumped"
+                             " previously.  The compilation starts with a"
+                             " reduced midEnd.  Only allowed/recognized in"
+                             " developer builds.")
     parser.add_argument("--pp", dest="pretty_print", default=None,
                         help="Pretty-print the program in the specified file.")
 
+
+### a utility function for grammatical correctness [and DRY]
```
mbudiu-vmw, line 79: In general I think it's better to use """ instead of # for this kind of documentation.

tools/driver/p4c_src/main.py
```diff
@@ -136,69 +154,134 @@ def main():
...
 
+
+
     user_defined_version = os.environ.get('P4C_DEFAULT_VERSION')
-    if user_defined_version != None:
+    ### Note: the next line of real code _was_τhe clear and easy-to-understand
+    ###       “if user_defined_version != None:”, but “pycodestyle”
+    ###       complained about that idiom so I rewrote it to something
+    ###       semantically-equivalent but with different syntax.
+    if not (user_defined_version is None):
```
mbudiu-vmw, line 237: I think that the python style is "is not None"

tools/driver/p4c_src/main.py
```diff
@@ -136,69 +154,134 @@ def main():
...
+    #       opts.language =  "p4-16"
+    ###
+    ### The following should be more general, i.e. it should handle e.g.
+    ###   “p4_22” ⇒ “p4-22” just fine.
+    ###
+    ### Note: Abe dislikes the ASCII dash/minus in
+    ###   “p4-”<some decimal integer>, but is keeping it here so as to not
+    ###   break/change this part of the contract/interface between the driver
+    ###   and the rest of the compiler.  One of these days/years,
+    ###   Abe suggests switching to e.g. “P4₁₄” and “P4₁₆” instead... ;-)
```
mbudiu-vmw, line 258: perhaps, but I have no idea how to type these symbols on my keyboard

tools/driver/p4c_src/main.py
```diff
@@ -224,39 +307,152 @@ def main():
...
             backend = target
             break
-    if backend == None:
+    if backend is None:
         parser.error("Unknown backend: {}-{}".format(str(opts.target),
                                                      str(opts.arch)))
+    error_count = 0
+
+    JSON_input_specified = env_indicates_developer_build and opts.json_source
+### ^^^^^^^^^^^^^^^^^^^^: DRY
```
mbudiu-vmw, line 316: I have no idea what DRY stands for

tools/driver/p4c_src/main.py
```diff
@@ -224,39 +307,152 @@ def main():
...
+        ###       which tries to feed the driver both a JSON input pathname
+        ###       via the relevant flag _and_ another input pathname,
+        ###       that will now fail whereas before the JSON input pathname
+        ###       would silently override the other input pathname[s].
+        print("\n"
+              "ERROR: it seems that both (firstly) a JSON input pathname "
+              'was specified via "' + JSON_input_flag + '"'
+              " _and_ (secondly) " + str(len(opts.P4_source_files)) + ""
+              ###                                                     ^^
+              ### The above nonsense is to get rid of a W504 in “pycodestyle”
```
mbudiu-vmw, line 332: If pycodestyle is stupid you should probably just disable that rule. I assume it's possible via comments; for pylint it is.

tools/driver/p4c_src/main.py
```diff
@@ -224,39 +307,152 @@ def main():
...
+        if JSON_input_specified:
+            pathnames_to_check.append(opts.json_source)
+
+        for pathname in pathnames_to_check:
+            if not os.path.isfile(pathname):
+                print('\nERROR: the input file "{}" does not exist.'.
+                      format(pathname), file=sys.stderr)
+                error_count += 1
+
+    elif not any_input_specified:
```
mbudiu-vmw, line 391: Maybe you should call this "some_input_specified"

tools/driver/p4c_src/main.py
```diff
@@ -224,39 +307,152 @@ def main():
...
+        use_a_dummy_P4_pseudoPathname = True
+
```
mbudiu-vmw, line 393: so many empty lines?

tools/driver/p4c_src/main.py
```diff
@@ -224,39 +307,152 @@ def main():
...
+              format(error_count, s_and_were_or_just_was(error_count)),
+              file=sys.stderr)
+        sys.exit(min(255, error_count))
+        ### maximum of 255: being extra-careful here,
+        ###                 just in case “error_count” is
+        ###                 a positive integer multiple of 256
+
+
+
+    ### A reminder: by this point, “error_count” _must_ equal zero,
```
mbudiu-vmw, line 408: Replacing this comment with an assert `error_count == 0` would be more readable and succinct

mbudiu-vmw (changes requested):

tools/driver/p4c_src/main.py
Abe149: Quoting `https://en.wikipedia.org/wiki/Don%27t_repeat_yourself`:

     "Don't repeat yourself" (DRY) is a [principle](https://en.wikipedia.org/wiki/Principle#Principle_as_axiom_or_logical_fundament) of [software development](https://en.wikipedia.org/wiki/Software_development_process) aimed at reducing repetition of software patterns,[[1]](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself#cite_note-1) replacing it with abstractions or using [data normalization](https://en.wikipedia.org/wiki/Data_normalization) to avoid redundancy.


Do you think I should spell it out in the [driver] comment[s]?

tools/driver/p4c_src/main.py
Abe149: Yes, but I have observed a bug in the Python language: “is not (None)” and “is (not None)” mean completely different things, and IMO that is a language-design bug, so I am just avoiding that ambiguity/problem on purpose by writing the code in a way that is both unambiguous and obviously means what it really means.  I hope that is OK.  😉

I guess, if we were designing a Python-like language from scratch, and we didn`t mind non-letter characters in key“words”, we could create an “isn`t” operator...  but that is probably just yet another of my silly programming dreams. 😉

tools/driver/p4c_src/main.py
Abe149: Some good pages for copy+paste...

[English Wikipedia](https://en.wikipedia.org/wiki/Unicode_subscripts_and_superscripts)
[https://unicode-table.com/en/sets/superscript-and-subscript-letters/](https://unicode-table.com/en/sets/superscript-and-subscript-letters/)
[https://unicode-table.com/en/blocks/superscripts-and-subscripts/](https://unicode-table.com/en/blocks/superscripts-and-subscripts/)
[https://www.fileformat.info/info/unicode/block/superscripts_and_subscripts/list.htm](https://www.fileformat.info/info/unicode/block/superscripts_and_subscripts/list.htm)
[https://lingojam.com/SubscriptGenerator](https://lingojam.com/SubscriptGenerator)

        

A PDF from the Unicode standard itself (also usable as a source for copy+paste): [Superscripts and Subscripts](https://unicode.org/charts/PDF/U2070.pdf)

        

... and last but perhaps not least, here is a collection of free-of-charge keymap files for entering Unicode on recent-enough versions of MS Windows; with the one of these that I have been using recently, holding down the (Windows95) menu/context key and pressing the “1!” key gives me ‘₁’, and holding down the (Windows95) menu/context key and pressing the “6^” key gives me ‘₆’...

[http://ilyaz.org/keyboard/](http://ilyaz.org/keyboard/)  (currently redirects to “https://math.berkeley.edu/~serganov/ilyaz.org/keyboard/”.)

tools/driver/p4c_src/main.py
Abe149: I made some changes to the defaults for “pycodestyle”; please see the file “setup.cfg___for_pycodestyle” in the same subdirectory.  [“setup.cfg” is merely a symlink to that file; I did it that way on purpose because IMO the name “setup.cfg” is too vague, and I didn`t want anybody thinking that I had checked garbage into the repo.]

Do you think I should add W504 to the “setup.cfg___for_pycodestyle” file and remove the relevant idiom from the Python code?

tools/driver/p4c_src/main.py
Abe149: Agreed; changed in my latest [local] Git.  Should be included in a batch of changes I intend to upload soon-ish.

tools/driver/p4c_src/main.py
Abe149: I added expansions/explanations of “DRY” in the 2 places it appears [both in comments] in “main.py” [in my WIP Git], and left alone “DRYified”.  I guess that`s good enough, so I`m going to tell GitHub to “resolve” this one.

tools/driver/p4c_src/main.py
Abe149: Do you feel strongly that the variable must be renamed?

tools/driver/p4c_src/main.py
Abe149: I often use 3 blank/empty lines in a row to visually indicate something like “these two sections of code might be in the same scope, but they are _conceptually_ different/separate”.  I could add a comment line instead, if you think that`s much better, but the comment might just be something like “#############################”.  😉

tools/driver/p4c_src/main.py
Abe149: Thanks for the suggestion.  I added a similar assertion just now in my Git WIP.  I also added comments that explain the motivation/thinking.

tools/driver/p4c_src/main.py
Abe149: Is it OK to “resolve” this one in GitHub now?

tools/driver/p4c_src/main.py
Abe149: Oh, and “is_not” would work fine as a key“word” too, of course. 😉

mbudiu-vmw (approved):

PR #3193, opened 2022-04-08
Fixing parserUnroll bug for sequential next operator usage

VolodymyrPeschanenkoIntel:
These changes fix bug of parserUnroll which appears in issue561-6-bmv2.p4 example.
This example contains the following code:
```
state parseO1O1 {
        packet.extract(hdr.u.next.byte);
        packet.extract(hdr.u.next.byte);
        transition accept;
}
```
ParserUnroll generated the following state before these changes:
```
state parseO1O1 {
        packet.extract(hdr.u[0].byte);
        packet.extract(hdr.u[0].byte);  /* <= here should be hdr.u[1].byte, according to the semantics of a next operator. */
        transition accept;
}
```


fruffy:
Any way to write a test for this? It will take a while until the loops unrolling pull request for p4test is in.

fruffy (approved):
Looks fine from my point of view. @mbudiu-vmw  any comments?

midend/parserUnroll.cpp
```diff
@@ -147,19 +147,19 @@ class ParserStateRewriter : public Transform {
...
+                idx = state->statesIndexes.at(expression->expr->toString());
+                if (idx + 1 < array->size && expression->member.name != IR::Type_Stack::last) {
+                    offset = 1;
+                }
             }
             if (expression->member.name == IR::Type_Stack::lastIndex) {
                 return new IR::Constant(IR::Type_Bits::get(32), idx);
             } else {
-                state->statesIndexes[expression->expr->toString()] = idx;
+                state->statesIndexes[expression->expr->toString()] = idx + offset;
```
fruffy, line 160: `expression->expr->toString()` is there no better way to map this?

VolodymyrPeschanenkoIntel: clone_id can't be applied here. It is possible to implement operator< for IR::Member, but IR::Member::member still contains cstring. So, for now, I think that this is the best solution.

VolodymyrPeschanenkoIntel: clone_id can't be applied here. It is possible to implement operator< for IR::Member, but IR::Member::member still contains cstring. So, for now, I think that this is the best solution.

mbudiu-vmw:
The spec does actually not mandate this behavior. The spec says that `nextIndex` should only be incremented if extracting directly to `next`. You can argue that this is a bug in the spec, but we should first take this change to the language design committee.

mbudiu-vmw:
See https://p4.org/p4-spec/docs/P4-16-v1.2.2.html#sec-packet-extract-one

mbudiu-vmw:
Filed https://github.com/p4lang/p4-spec/issues/1048

midend/parserUnroll.cpp
mbudiu-vmw: node->toString() should only be used for reporting user messages.
Such usage in the compiler is unsafe, since the different expressions may generate identical toString outputs.

mbudiu-vmw (changes requested):

VolodymyrPeschanenkoIntel:
I added StackVariable class to support std::map<IR::Member, ...>

midend/parserUnroll.cpp
```diff
@@ -147,19 +225,19 @@ class ParserStateRewriter : public Transform {
             BUG_CHECK(l->is<SymbolicArray>(), "%1%: expected an array", l);
             auto array = l->to<SymbolicArray>();
             unsigned idx = 0;
-            for (size_t i = 0; i < array->size; i++) {
-                auto* v = array->get(expression, i);
-                if (v->hasUninitializedParts())
-                    break;
-                else
-                    idx = i;
+            unsigned offset = 0;
```
mbudiu-vmw, line 228: can you explain what is happening here?

mbudiu-vmw (changes requested):

midend/parserUnroll.cpp
```diff
@@ -5,13 +5,91 @@
...
+    if (member == nullptr) {
+        return false;
+    }
+
+    // A member is a stack variable if it is qualified by a PathExpression or if its qualifier is a
+    // stack variable.
+    return member->expr->is<IR::PathExpression>() || repOk(member->expr);
+}
+
+bool StackVariable::operator<(const StackVariable& other) const {
```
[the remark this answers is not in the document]
VolodymyrPeschanenkoIntel: I need comparisons to use this class as a key for std::map.

mbudiu-vmw:
You do not seem to have addressed the comments made in my previous review.

midend/parserUnroll.cpp
VolodymyrPeschanenkoIntel: if it is a next operator then we need to make offset = 1, otherwise it should be 0. I did some refactoring there.

VolodymyrPeschanenkoIntel:
> The spec does actually not mandate this behavior. The spec says that `nextIndex` should only be incremented if extracting directly to `next`. You can argue that this is a bug in the spec, but we should first take this change to the language design committee.

It looks very strange, because in the documentation of P4C we have:
`hs.next: produces a reference to the element with index hs.nextIndex in the stack. May only be used in a parser. If the stack's nextIndex counter is greater than or equal to size, then evaluating this expression results in a transition to reject and sets the error to error.StackOutOfBounds. If hs is an l-value, then hs.next is also an l-value.`
As I can see there are no restrictions for operator `next `usage.  So, I think that we need to increment index of a header stack for next operator independently of a place where we use it. Am I right?

midend/parserUnroll.h
```diff
@@ -33,6 +33,54 @@ const char outOfBoundsStateName[] = "stateOutOfBound";
...
+ public:
+    /// Determines whether @expr can represent a StateVariable.
+    static bool repOk(const IR::Expression* expr);
+
+    // Implicit conversions to allow implementations to be treated like a Node*.
+    operator const IR::Expression*() const { return expr; }
+    const IR::Expression& operator*() const { return *expr; }
+    const IR::Expression* operator->() const { return expr; }
+
+    // Implements comparisons so that StateVariables can be used as map keys.
```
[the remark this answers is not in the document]
VolodymyrPeschanenkoIntel: Does p4c have some hash for IR::Expression? I used std::map. So, I turned back to operator<

mbudiu-vmw:
> > The spec does actually not mandate this behavior. The spec says that `nextIndex` should only be incremented if extracting directly to `next`. You can argue that this is a bug in the spec, but we should first take this change to the language design committee.
> 
> It looks very strange, because in the documentation of P4C we have: `hs.next: produces a reference to the element with index hs.nextIndex in the stack. May only be used in a parser. If the stack's nextIndex counter is greater than or equal to size, then evaluating this expression results in a transition to reject and sets the error to error.StackOutOfBounds. If hs is an l-value, then hs.next is also an l-value.` As I can see there are no restrictions for operator `next `usage. So, I think that we need to increment index of a header stack for next operator independently of a place where we use it. Am I right?

See https://p4.org/p4-spec/docs/P4-16-v1.2.2.html#sec-packet-extract-one, in particular the pseudo-code for extract. It says nothing about header_unions. That is arguably wrong, but it should be changed if it is wrong.

midend/parserUnroll.cpp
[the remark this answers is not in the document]
mbudiu-vmw: The template declaration for unordered_map https://en.cppreference.com/w/cpp/container/unordered_map has two arguments for functions that you can implement for computing the Hash and the equality.

VolodymyrPeschanenkoIntel:
> > > The spec does actually not mandate this behavior. The spec says that `nextIndex` should only be incremented if extracting directly to `next`. You can argue that this is a bug in the spec, but we should first take this change to the language design committee.
> > 
> > 
> > It looks very strange, because in the documentation of P4C we have: `hs.next: produces a reference to the element with index hs.nextIndex in the stack. May only be used in a parser. If the stack's nextIndex counter is greater than or equal to size, then evaluating this expression results in a transition to reject and sets the error to error.StackOutOfBounds. If hs is an l-value, then hs.next is also an l-value.` As I can see there are no restrictions for operator `next `usage. So, I think that we need to increment index of a header stack for next operator independently of a place where we use it. Am I right?
> 
> See https://p4.org/p4-spec/docs/P4-16-v1.2.2.html#sec-packet-extract-one, in particular the pseudo-code for extract. It says nothing about header_unions. That is arguably wrong, but it should be changed if it is wrong.

I see. So, you said that "we should first take this change to the language design committee". How to do that?

midend/parserUnroll.cpp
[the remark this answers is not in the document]
mbudiu-vmw: This comment has not been addressed; I suggested the use of unordered_map.

midend/parserUnroll.cpp
```diff
@@ -23,75 +24,32 @@ bool StackVariable::repOk(const IR::Expression* expr) {
...
-int StackVariable::compare(const IR::PathExpression* p1, const IR::PathExpression* p2) {
-    if (p1->path->name.name < p2->path->name.name) {
-        return -1;
-    }
-    if (p1->path->name.name > p2->path->name.name) {
-        return 1;
+size_t StackVariableHash::operator()(const StackVariable& var) const {
+    const IR::Member* curMember = var.operator->();
+    std::vector<size_t> h;
+    while(curMember) {
```
mbudiu-vmw: missing space

midend/parserUnroll.cpp
```diff
@@ -23,75 +24,32 @@ bool StackVariable::repOk(const IR::Expression* expr) {
...
+        h.push_back(Util::Hash::fnv1a<const cstring>(curMember->member.name));
+        if (auto* path = curMember->expr->to<IR::PathExpression>()) {
+            h.push_back(Util::Hash::fnv1a<const cstring>(path->path->name));
+            break;
+        }
+        BUG_CHECK(curMember->expr->is<IR::Member>(), "Not a valid StateVariable: %1%", curMember);
```
mbudiu-vmw, line 41: these two lines are the same as a checkedTo.

mbudiu-vmw (approved):

midend/parserUnroll.cpp
fruffy: This needs to be fixed, then I can merge @VolodymyrPeschanenkoIntel.

VolodymyrPeschanenkoIntel:
@fruffy, @mbudiu-vmw Can we merge it?

PR #3195, opened 2022-04-08
Add support for DirectCounter extern to PSA/eBPF backend

tatry:
This PR adds the support for DirectCounter extern to PSA/eBPF backend.

Co-authored-by: Tomasz Osiński <tomasz@opennetworking.org>

### DIFF
--- a/backends/ebpf/ebpfTable.h
+++ b/backends/ebpf/ebpfTable.h
@@ -104,6 +104,7 @@ class EBPFTable : public EBPFTableBase {
     virtual void emitValueType(CodeBuilder* builder);
     virtual void emitValueActionIDNames(CodeBuilder* builder);
     virtual void emitValueStructStructure(CodeBuilder* builder);
+    virtual void emitDirectTypes(CodeBuilder* builder) { (void) builder; }
     virtual void emitAction(CodeBuilder* builder, cstring valueName, cstring actionRunVariable);
     virtual void emitInitializer(CodeBuilder* builder);
     virtual void emitLookup(CodeBuilder* builder, cstring key, cstring value) {
--- a/backends/ebpf/ebpfTable.h
+++ b/backends/ebpf/ebpfTable.h
@@ -104,7 +104,8 @@ class EBPFTable : public EBPFTableBase {
     virtual void emitValueType(CodeBuilder* builder);
     virtual void emitValueActionIDNames(CodeBuilder* builder);
     virtual void emitValueStructStructure(CodeBuilder* builder);
-    virtual void emitDirectTypes(CodeBuilder* builder) { (void) builder; }
+    // Emits value types used by direct externs.
+    virtual void emitDirectValueTypes(CodeBuilder* builder) { (void) builder; }
     virtual void emitAction(CodeBuilder* builder, cstring valueName, cstring actionRunVariable);
     virtual void emitInitializer(CodeBuilder* builder);
     virtual void emitLookup(CodeBuilder* builder, cstring key, cstring value) {
--- a/backends/ebpf/psa/ebpfPsaTable.cpp
+++ b/backends/ebpf/psa/ebpfPsaTable.cpp
@@ -223,7 +223,7 @@ void EBPFTablePSA::emitTypes(CodeBuilder* builder) {
  * Order of emitting counters and meters affects generated layout of BPF map value.
  * Do not change this order!
  */
-void EBPFTablePSA::emitDirectTypes(CodeBuilder* builder) {
+void EBPFTablePSA::emitDirectValueTypes(CodeBuilder* builder) {
     for (auto ctr : counters) {
         ctr.second->emitValueType(builder);
     }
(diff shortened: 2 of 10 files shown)


backends/ebpf/ebpfTable.h
```diff
@@ -104,6 +104,7 @@ class EBPFTable : public EBPFTableBase {
     virtual void emitValueType(CodeBuilder* builder);
     virtual void emitValueActionIDNames(CodeBuilder* builder);
     virtual void emitValueStructStructure(CodeBuilder* builder);
+    virtual void emitDirectTypes(CodeBuilder* builder) { (void) builder; }
```
mbudiu-vmw, line 107: can you document this function? I can't guess what it's doing from its name.

backends/ebpf/psa/ebpfPsaTable.cpp
```diff
@@ -23,12 +23,36 @@ limitations under the License.
 
 namespace EBPF {
 
+template<class F>
+class EBPFTablePsaPropertyVisitor : public Inspector {
+ protected:
+    F functor;  // called for every property entry
```
mbudiu-vmw, line 29: seems to be called for their values which are PathExpression only.

backends/ebpf/psa/ebpfPsaTable.cpp
```diff
@@ -54,16 +78,14 @@ void ActionTranslationVisitorPSA::processMethod(const P4::ExternMethod* method)
...
+    if (declType->name.name == "DirectCounter") {
+        auto ctr = table->getDirectCounter(instanceName);
+        if (ctr != nullptr)
+            ctr->emitDirectMethodInvocation(builder, method, valueName);
+        else
             ::error(ErrorType::ERR_NOT_FOUND,
-                    "%1%: Counter named %2% not found",
-                    method->expr, instanceName);
-        }
+                    "%1%: Table %2% do not own DirectCounter named %3%",
```
mbudiu-vmw, line 87: "does"

backends/ebpf/psa/ebpfPsaTable.cpp
```diff
@@ -175,6 +213,17 @@ void EBPFTablePSA::emitTypes(CodeBuilder* builder) {
     // TODO: placeholder for handling PSA-specific types
 }
 
+/**
+ * Remember that order of emitting counters and meters affects future access to BPF maps.
```
mbudiu-vmw, line 217: I find this comment mysterious. "future"? In which way does it affect the access?

backends/ebpf/psa/externs/ebpfPsaCounter.cpp
```diff
@@ -153,6 +187,35 @@ void EBPFCounterPSA::emitMethodInvocation(CodeBuilder* builder, const P4::Extern
...
 }
 
+void EBPFCounterPSA::emitDirectMethodInvocation(CodeBuilder* builder,
+                                                const P4::ExternMethod* method,
+                                                cstring valuePtr) {
+    if (method->method->name.name != "count") {
+        ::error(ErrorType::ERR_UNSUPPORTED, "Unexpected method %1%", method->expr);
+        return;
+    }
+    BUG_CHECK(isDirect, "Bad Counter invocation");
```
mbudiu-vmw, line 197: should these be ::error, or are you satisfied with BUG?

mbudiu-vmw (approved):
I am approving, but I have a few small comments.

backends/ebpf/psa/ebpfPsaTable.cpp
tatry: What do you mean? The comment should be clarified or something else?

backends/ebpf/psa/externs/ebpfPsaCounter.cpp
tatry: This situation should never occur, where for DirectCounter emitMethodInvocation is called and for Counter emitDirectMethodInvocation is called. This has more to do with compiler implementation than P4 code, so `BUG` is a better option in my opinion.

backends/ebpf/psa/ebpfPsaTable.cpp
mbudiu-vmw: Actually the right way to do this is to have this be a base class and define a new subclass for each different analysis, with just a preorder(PathExpression*) method. Please rewrite it that way.

tatry:
@mbudiu-vmw thank you for your review, I think that I've addressed all the comments.

backends/ebpf/psa/ebpfPsaTable.cpp
```diff
@@ -23,12 +23,106 @@ limitations under the License.
...
 
+class EBPFTablePsaPropertyVisitor : public Inspector {
+ protected:
+    EBPFTablePSA* table;
+
+ public:
+    explicit EBPFTablePsaPropertyVisitor(EBPFTablePSA* table) : table(table) {}
+
+    // Use these two preorders to print error when property contains something other than name of
+    // extern instance. ListExpression is required because without it Expression will take precede
```
mbudiu-vmw, line 34: precedence

mbudiu-vmw (approved):

PR #3196, opened 2022-04-08
Prototype implementation of static_assert

### LINKED ISSUE #3188: Add a static_assert "function" to P4
Which is evaluated at compilation time.

mbudiu-vmw:

Fixes #3188

We should discuss this feature in the language design working group. Seems very useful.
It came up during discussions about a module system for P4.


apinski-cavium:
Hmm, why not make static_assert a keyword and then have it not as a "function call" and have the parser parse it instead? Seems like a better than having an extern here. Maybe I am coming from a C/C++ point of view where adding keywords now is fine (C23 adds static_assert as a keyword).

mbudiu-vmw:
I try to keep backwards compatibility. I don't want to add new reserved keywords unless absolutely necessary.

mbudiu-vmw:
on the other hand, as it is implemented it may break programs that use this identifier.

mbudiu-vmw:
Can we get a review here to merge it?

mbudiu-vmw:
@jafingerhut @jnfoster can you please review this PR?

frontends/p4/staticAssert.h
```diff
@@ -0,0 +1,87 @@
...
+                            CHECK_NULL(msg);
+                            if (auto sl = msg->expression->to<IR::StringLiteral>()) {
+                                message = sl->value;
+                            }
+                        }
+                        ::error(ErrorType::ERR_EXPECTED, "%1%: %2%",
+                                method, message);
+                    }
+                } else {
+                    ::warning("Could not evaluate static_assert to a constant: %1%", arg);
```
jnfoster, line 68: This is the behavior we want? A static assert that cannot be statically resolved is just a warning and not an error.

jnfoster (approved):
LGTM

frontends/p4/staticAssert.h
mbudiu-vmw: I actually expect that this will fail in typechecking, since the parameters are directionless.
So this case should never arise in practice.

mbudiu-vmw: I guess this would signal a failure of constant-folding some expression.

mbudiu-vmw:
Discussed at https://github.com/p4lang/p4-spec/pull/1129

frontends/p4/staticAssert.h
jnfoster: Ahh right. Makes sense.

jnfoster: Should this be a `BUG` then?

mbudiu-vmw: It could be a BUG as well, if we are confident in the constant-folding code.
I say we leave it like this for now until someone complains about hitting the warning.

PR #3199, opened 2022-04-08
Remove incorrect test for recursion

### LINKED ISSUE #3197: Compiler incorrectly rejects some non-recursive calls as recursive
See https://github.com/p4lang/p4-spec/issues/1050

mbudiu-vmw:

Fixes #3197

It turns out that one cannot write recursive functions or actions in P4 due to the way name resolution is made: in the body of function `f` any reference to an identifier `f` will not resolve as the current function. So you will get an error like "declaration not found".


jnfoster (approved):
Looks good.

However, should we add test that recursive functions are still rejected?

mbudiu-vmw:
Hard to tell, since the error is very generic ("declaration not found"), but apparently not.

PR #3202, opened 2022-04-10
Add support for Register extern to PSA/eBPF backend

osinstom:


backends/ebpf/psa/externs/ebpfPsaRegister.cpp
```diff
@@ -0,0 +1,230 @@
...
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+#include "ebpfPsaRegister.h"
+#include "backends/ebpf/psa/ebpfPsaControl.h"
+
+namespace EBPF {
+
+EBPFRegisterPSA::EBPFRegisterPSA(const EBPFProgram *program,
```
mbudiu-vmw, line 23: This stuff could be useful even in the basic ebpf backend, but perhaps it's too late to move it there.

backends/ebpf/psa/externs/ebpfPsaRegister.cpp
```diff
@@ -0,0 +1,230 @@
...
+                                 cstring instanceName, const IR::Declaration_Instance *di,
+                                 CodeGenInspector *codeGen) : EBPFTableBase(program,
+                                                                            instanceName,
+                                                                            codeGen) {
+    CHECK_NULL(di);
+    if (!di->type->is<IR::Type_Specialized>()) {
+        ::error(ErrorType::ERR_MODEL, "Missing specialization: %1%", di);
+        return;
+    }
+    auto ts = di->type->to<IR::Type_Specialized>();
```
mbudiu-vmw, line 33: Didn't you create some helper functions to extract arguments from specialized types?

backends/ebpf/psa/externs/ebpfPsaRegister.cpp
```diff
@@ -0,0 +1,230 @@
...
+    }
+
+    this->keyArg = ts->arguments->at(1);
+    this->valueArg = ts->arguments->at(0);
+    this->keyType = EBPFTypeFactory::instance->create(keyArg);
+    this->valueType = EBPFTypeFactory::instance->create(valueArg);
+
+    if (auto wt = dynamic_cast<IHasWidth *>(this->keyType)) {
+        unsigned keyWidth = wt->widthInBits();
+        // For keys <= 32 bit register is based on array map,
```
mbudiu-vmw, line 47: Maybe you should put this into a separate policy function; annotations could be used to generalize this later.

backends/ebpf/psa/externs/ebpfPsaRegister.cpp
```diff
@@ -0,0 +1,230 @@
...
+
+    auto declaredSize = di->arguments->at(0)->expression->to<IR::Constant>();
+    if (!declaredSize->fitsUint()) {
+        ::error(ErrorType::ERR_OVERLIMIT, "%1%: size too large", declaredSize);
+        return;
+    }
+    size = declaredSize->asUnsigned();
+
+    if (di->arguments->size() == 2) {
+        this->initialValue = di->arguments->at(1)->expression->to<IR::Constant>();
```
mbudiu-vmw, line 65: Maybe you should document the fact that this can be nullptr.
But if the argument is present and not a constant you will also get a nullptr, which is probably wrong.

backends/ebpf/psa/externs/ebpfPsaRegister.cpp
```diff
@@ -0,0 +1,230 @@
...
+        useArrayMap = (keyWidth > 0 && keyWidth <= 32);
+    }
+
+    if (di->arguments->size() < 1) {
+        ::error(ErrorType::ERR_MODEL, "Expected at least 1 argument: %1%", di);
+        return;
+    }
+
+    auto declaredSize = di->arguments->at(0)->expression->to<IR::Constant>();
+    if (!declaredSize->fitsUint()) {
```
mbudiu-vmw, line 58: no CHECK_NULL?
Has this been validated someplace else?

backends/ebpf/psa/externs/ebpfPsaRegister.cpp
```diff
@@ -0,0 +1,230 @@
...
+    }
+
+    builder->target->emitTraceMessage(builder, "Register: Entry not found, using default value");
+    builder->blockEnd(true);
+}
+
+void EBPFRegisterPSA::emitRegisterWrite(CodeBuilder* builder, const P4::ExternMethod* method,
+                                        ControlBodyTranslatorPSA* translator) {
+    auto indexArgExpr = method->expr->arguments->at(0)->expression->to<IR::PathExpression>();
+    cstring indexParamName = translator->getActionParamName(indexArgExpr);
```
mbudiu-vmw, line 203: is this implying that within an action the index must be an action parameter?
So it cannot be some expression that involves the action parameter?

mbudiu-vmw (changes requested):

backends/ebpf/psa/externs/ebpfPsaRegister.cpp
osinstom: `EBPFObject::getSpecializedTypeName` only gets a type name, we need to extract arguments here.

backends/ebpf/psa/externs/ebpfPsaRegister.cpp
osinstom: We can think about moving it to EBPF model, but we would need to agree on the extern definition, which might be different from the one from PSA.

backends/ebpf/psa/externs/ebpfPsaRegister.cpp
osinstom: the function name is misleading, I'll change it

osinstom: I believe that the behavior that you mention is tested in https://github.com/p4lang/p4c/blob/1c179a11f52f2b38c7db5e3fda68e16893ed05b2/backends/ebpf/tests/p4testdata/register-action.p4.

mbudiu-vmw:
The ptf tests are currently failing.

osinstom:
@mbudiu-vmw I've fixed PTF tests. They are passing now.

mbudiu-vmw (approved):

Issue #3203, opened 2022-04-11
"Compiler bug" found with generic types and specialized and function argument


apinski-cavium:
This was found while getting a feel for generic types:
```
struct h<t>
{
  t f;
}
bit func(h<bit> a)
{
  return a.f;
}
```
Compiler output:
```
apinski@xeond:~/src/p4/p4c/build$ ./p4c generic.p4
In file: /home/apinski/src/p4/p4c/frontends/p4/specializeGenericTypes.cpp:119
Compiler Bug: /home/apinski/src/p4/p4c/frontends/p4/specializeGenericTypes.cpp:119: Null insert
```

PR #3207, opened 2022-04-12
Insert type specializations before functions

### LINKED ISSUE #3204: Compiler bug with generic types and decl inside a function
Take:
```
struct h<t>
{
  t f;
}
bit func()
{
  h<bit> a;
  a.f = 1;
  return a.f;
}
```
We get the following compiler bug:
```
In file: /home/apinski/src/p4/p4c/ir/ir-inline.h:159
Compiler Bug: visitor returned invalid type Vector<Node> for IndexedVector<StatOrDecl>
```

mbudiu-vmw:
Fixes #3204


mbudiu-vmw:
Fixes #3203

rst0git (approved):
LGTM

PR #3209, opened 2022-04-12
DPDK Backend: Fix binary operations with 1st operand is not same as dst operand

usha1830:
DPDK needs the binary operation to have 1st src operand same as dst operand.
Some cases with non commutative operations were not correctly handled by the existing transformation pass.
For example:
`a = b << a`
should be converted to

```
tmp =b
tmp = tmp << a
a = tmp
```


hanw:
You can save one more instruction with
```
a = a - 10
a = -a
```

usha1830:
> You can save one more instruction with
> 
> ```
> a = a - 10
> a = -a
> ```

Han, the transformation also applies to other non-commutative operations like shift
I have updated the PR description  to avoid confusion, will also update the code comment for this.

hanw:
Ok, but my comment is still valid.

backends/dpdk/dpdkUtils.cpp
```diff
@@ -0,0 +1,56 @@
...
+bool isCommutativeBinaryOperation(const IR::Operation_Binary *bin) {
+    auto right = bin->right;
+    if (right->is<IR::Add>() || right->is<IR::Equ>() || right->is<IR::LOr>() ||
+        right->is<IR::LAnd>() || right->is<IR::BOr>() || right->is<IR::BAnd>() ||
+        right->is<IR::BXor>())
+        return true;
+    return false;
+}
+
+bool isStandardMetadata(cstring name) {
```
mbudiu-vmw, line 41: This function is actually PSA-specific, not DPDK specific. So it should probably move to some common file eventually.

mbudiu-vmw (approved):

usha1830:
> Ok, but my comment is still valid.

Yes.  Do you think, this case should be specially handled? 

Also, I don't see any instruction for negation here 
https://doc.dpdk.org/api/rte__swx__pipeline_8h.html#acf7715fccb903f8eca8b000953b188cb

backends/dpdk/dpdkUtils.cpp
usha1830: How about creating a utils.h/cpp under backends/common/ ?

hanw:
> > Ok, but my comment is still valid.
> 
> Yes. Do you think, this case should be specially handled?
> 
> Also, I don't see any instruction for negation here https://doc.dpdk.org/api/rte__swx__pipeline_8h.html#acf7715fccb903f8eca8b000953b188cb

Thanks for the pointer. This looks fine.

backends/dpdk/dpdkUtils.cpp
mbudiu-vmw: We should probably move more stuff in such a place. But it should be in some directory named "psa".

usha1830: This specific function is for both psa and pna metadata. Of course, we can split it up if we decide to go with arch-specific common utils.
Can this PR be merged if this does not need to be an immediate action?

Issue #2800, opened 2021-06-14
Check whether the expansion of a range into ternary values is correct for signed numbers


mbudiu-vmw:

vlstill:
As far as I understand, the P4-16 specification allows only `bit` but not `int` for masks/ternary matches (https://p4.org/p4-spec/docs/P4-16-v1.2.2.html#sec-cubes) while it allows also `int` for ranges. That would suggest to me that masks should not be used with signed values. Is the `IR::Mask` actually intended to internally work also with signed values? Or am I missing something?

mbudiu-vmw:
Yes, this is probably a lurking bug. I will mark it as such.
I wonder whether we could handle it by converting a negative range to a positive one using two's complement.

mbudiu-vmw:
The solution may be to break a range containing 0 into two disjoint ranges.

vlstill:
That might be actually the only way (since going over 0 most of the bits change anyway). I will have a look at this, I have made some tweaks into the range conversion recently so it should be an easy one.

vlstill:
The solution merged in #3212 is the following

- if range is done over a signed type (i.e. the type of the key is signed, `int<N>`), the key is bitcast to corresponding `bit<N>` type so that we can match by mask which only works on unsigned according to P4 specs
- the range is converted to an unsigned range which matches the original range as reinterpreted from 2-complement to unsigned 
  - if the range goes over 0 it is split in two (min..-1, 0..max) as the original range is not continuous when reinterpreted to unsigned
- the unsigned range is converted to masks (no changes here)

I think this issue can be closed now.

PR #3212, opened 2022-04-13
Fix expansion of signed ranges to masks

vlstill:
This fixes "#2800 Check whether the expansion of a range into ternary values is correct for signed numbers".

### DIFF
--- a/midend/replaceSelectRange.cpp
+++ b/midend/replaceSelectRange.cpp
@@ -66,7 +66,7 @@ expandRange(const IR::Range *r, std::vector<const IR::Mask *> &masks, const IR::
 }
 
 std::vector<const IR::Mask *>
-DoReplaceSelectRange::rangeToMasks(const IR::Range *r, int keyIndex) {
+DoReplaceSelectRange::rangeToMasks(const IR::Range *r, size_t keyIndex) {
     std::vector<const IR::Mask *> masks;
 
     auto l = r->left->to<IR::Constant>();
@@ -155,7 +155,7 @@ const IR::Node *DoReplaceSelectRange::postorder(IR::SelectExpression *e) {
 
     if (!signedIndicesToReplace.empty()) {
         IR::Vector<IR::Expression> newSelectList;
-        int idx = 0;
+        size_t idx = 0;
         for (auto *expr : e->select->components) {
             if (signedIndicesToReplace.count(idx)) {
                 auto eType = expr->type->to<IR::Type_Bits>();
@@ -199,7 +199,7 @@ const IR::Node* DoReplaceSelectRange::postorder(IR::SelectCase* sc) {
 
         newVectors.push_back(first);
 
-        int idx = 0;
+        size_t idx = 0;
         for (auto key : oldList->components) {
             if (auto r = key->to<IR::Range>()) {
                 auto masks = rangeToMasks(r, idx);
--- a/midend/replaceSelectRange.cpp
+++ b/midend/replaceSelectRange.cpp
@@ -21,7 +21,7 @@
 namespace P4 {
 
 std::vector<const IR::Mask *>
-DoReplaceSelectRange::rangeToMasks(const IR::Range *r) {
+DoReplaceSelectRange::rangeToMasks(const IR::Range *r, int keyIndex) {
     std::vector<const IR::Mask *> masks;
 
     auto l = r->left->to<IR::Constant>();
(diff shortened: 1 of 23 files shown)


jafingerhut (approved):
I have reviewed the test programs and their -midend results output by the compiler, and they all look good to me.  I have not attempted to review the C++ code changes.

jafingerhut:
I am not familiar enough with the error output from the failing test to understand what it means, but perhaps that is output from the Gauntlet tests looking for non-equivalent programs output by the front or midend?  If so, it might actually be a bug in the Gauntlet equivalence checker for this new feature, but again, I do not know enough about its output or implementation to be able to tell.  @fruffy ?

midend/replaceSelectRange.cpp
```diff
@@ -21,7 +21,7 @@
 namespace P4 {
 
 std::vector<const IR::Mask *>
-DoReplaceSelectRange::rangeToMasks(const IR::Range *r) {
+DoReplaceSelectRange::rangeToMasks(const IR::Range *r, int keyIndex) {
```
mbudiu-vmw, line 24: I think it's actually nicer to return a list of vectors for the result.
one in the common case, two if the range contains 0.
the keyIndex and signedIndicesToReplace make the data flow much harder to understand.

midend/replaceSelectRange.cpp
```diff
@@ -46,38 +46,67 @@ DoReplaceSelectRange::rangeToMasks(const IR::Range *r) {
...
+        BUG_CHECK(min <= max, "range bounds inverted %1%..%2%", min, max);
 
-        auto valConst = new IR::Constant(constType, min, base, true);
-        auto maskConst = new IR::Constant(constType, mask, base, true);
-        masks.push_back(new IR::Mask(r->srcInfo, valConst, maskConst));
+        big_int range_size_remaining = max - min + 1;
 
-        range_size_remaining -= match_stride;
-        min += match_stride;
+        while (range_size_remaining > 0) {
```
mbudiu-vmw, line 89: Maybe this should become a separate function.

testdata/p4_16_samples/issue2800c.p4
```diff
@@ -0,0 +1,36 @@
...
+
+struct Headers {
+    H h;
+}
+
+parser p(packet_in packet, out Headers hdr) {
+    state start {
+        packet.extract(hdr.h);
+        transition select(hdr.h.a, hdr.h.b) {
+            (0 .. 7, -5 .. 4): parse;
```
mbudiu-vmw, line 20: How about a test that has 2 ranges containing 0?

mbudiu-vmw (changes requested):

fruffy:
@jafingerhut Let me check. Considering this is a new feature I am guessing the tool does not have support for it yet.

vlstill:
@jafingerhut, @fruffy, looking at the output (not knowing what is is really :-D, but masing it on my knowledge of bit-vector SMT)

- it seems to me this is having problems with frontend code (`issue2800b-FrontEnd_0_P4V1::getV1ModelVersion.p4`, `issue2800b-FrontEnd_12_TypeInference.p4`) which suggest the problem is with the input, not the code produced by midend
- for hdr.h.a (model `issue2800b`) one of the inferred bounds is  65531 (0xfffc) which is unsigned reinterpretation of -5 which is the lower bound in the code. This suggests (together with use of `bvule` instead of `bvsle` (bit-vector unsigned/signed less then)) that the verifier is not working correctly with signed values
- I think the condition is even more wrong (reformated & commented):

```lisp
(let ((a!1 (ite  ; if
      (and (bvule #xfffb ((_ extract 31 16) extract_hdr))        ; 0xfffb <= extract_hdr[31:16]    (a)
           (= ((_ extract 31 19) extract_hdr) #b0000000000000)   ; and extract_hdr[31:19] == 0     (b)
           (bvule ((_ extract 18 16) extract_hdr) #b100)         ; and extract_hdr[18:16] <= 4     (c)
           (= ((_ extract 15 3) extract_hdr) #b0000000000000))   ; and extract_hdr[15:3] == 0      (d)
      #xffff ; then the result is 0xffff (that is -1, goign through parse state)
      ((_ extract 31 16) extract_hdr)))) ; else it is the value extracted from the packet
    (ite extract_hdr_valid a!1 invalid))
```
  - obviously, the conditions (a) and (c) overlap (and (a) is wrong), but also (b) interacts weirdly with (a) in my opinion (if higher 13 bits are zero, then obviously the largest possible number would be 0x7 not 0xfffc) -> I would expect the zero check to be part of the other bound, but maybe this is just artefact of the mishandled constants again

testdata/p4_16_samples/issue2800c.p4
vlstill: Sure, I have added one more test with two signed fields (not both of them are going over 0 since I realized I did not have any test for purely negative range. I think the "going over zero" case is covered sufficiently and this test mainly covers the missing "multiple fields are signed" case.

midend/replaceSelectRange.cpp
vlstill: I agree that `signedIndicesToReplace` does make the flow harder to understand. But this is mainly because it is indeed complex -- the indices are detected when range is expanded, but they are used in the corresponding select through which we backtrack later (making use of `postorder`). This needs to be done this way as at the time we encounter the select first, we don't know which values will contain ranges and when working with ranges we cannot replace a parent node (at least I don't think so, I could be wrong since I am a beginner to p4c). The only alternative to indices I see would be to remember the actual corresponding select expressions, but that in my opinion improves nothing and could lead to danger in case of some weird select with duplicated keys:

```
select (foo.a, foo.b, foo.a) {
   -5..5, 42, _ : ...
   _, 16, -32000 : ...
}
```

In this (rather pathological) case, we need to insert the bitcast only to the first occurence of `foo.a` not to both instances of `foo.a` expression in select.

---

As for the vector-of-vectors. I don't think this is a good idea as returning a vector of vectors implies that there can be any number of vectors, including 0 and more than 2. Therefore I think that would be bad design of the function as the type would allow too many invalid possibilities. It could make sense to return pair (isSigned, vectors) and resolve the index one level up in the postorder of mask.

---

For now, I have just added a comment about the indices into the class.

midend/replaceSelectRange.cpp
vlstill: I have refactored the `rangeToMasks` function a bit since indeed it was quite long.

fruffy:
Thanks for looking into this? Yeah I believe it is a type conversion issue in the range matching. The validator uses its own type inference which may have flaws. I xfailed those couple programs until I have implemented the fix.

vlstill:
@fruffy, thanks. So if @mbudiu-vmw is happy with my changes this can be merged.

midend/replaceSelectRange.cpp
```diff
@@ -115,12 +143,47 @@ DoReplaceSelectRange::cartesianAppend(const std::vector<IR::Vector<IR::Expressio
     return newVecs;
 }
 
+const IR::Node *DoReplaceSelectRange::preorder(IR::SelectExpression *e) {
+    BUG_CHECK(!inSelect, "A select nested in select: %1%", e);
+    inSelect = true;
```
mbudiu-vmw, line 148: In general you don't need such a boolean flag.
It can be replaced with `findContext<IR::SelectExpression>() != nullptr`, called from the child node.

midend/replaceSelectRange.h
```diff
@@ -38,12 +38,21 @@ class DoReplaceSelectRange : public Transform {
     // number of new cases generated for ternary operations due to a range
     // Each case is a key set expression.
     const uint MAX_CASES;
+    bool inSelect = false;
+    // Collects select indices which will need to be replaced with bitcast of
+    // the original value to unsigned. This is needed if we encounter a range
+    // over a signed value at the given index.
+    std::set<int> signedIndicesToReplace;
```
mbudiu-vmw, line 45: why not a set of size_t or unsigned values?
I would change this comment to say "an index i is in this set if selectExpression->components[i] needs to be cast from int<N> to bit<N>". This is only needed if there is a label that has in the i-th position a range expression that contains 0 inside the range".

mbudiu-vmw (changes requested):

mbudiu-vmw:
These are minor changes requested.

midend/replaceSelectRange.h
vlstill: (Because old habits die hard.) Changed to size_t. 

As for the comment, sure, except that we need to replace it any time the value is signed, not only when it crosses over zero. Masks are only defined for `bit` values in P4. I changed it.

midend/replaceSelectRange.cpp
vlstill: This flag is there mostly to avoid introduction of more inconsistencies if AST is unexpected (or visitor is broken, which is probably significantly less likely). We could avoid it altogether but I believe a cheap assert is better then possibility on long debugging in future. If I replace it with `findContext` I don't think it would be that cheap.

mbudiu-vmw: The compiler code is not optimized for speed, but rather for correctness, readability, and maintainability. I think that findContext is more readable, and it is in only one place, so it is more maintainable.

vlstill: changed

mbudiu-vmw (approved):

PR #3220, opened 2022-04-15
Allow casts int to int

### LINKED ISSUE #3219: Cast int -> int is illegal
Filed originally here: https://github.com/p4lang/p4-spec/issues/1058

mbudiu-vmw:
Fixes #3219

### DIFF
--- a/testdata/p4_16_samples/issue3219.p4
+++ b/testdata/p4_16_samples/issue3219.p4
@@ -1,0 +1,9 @@
+bit<4> func1() {
+    bit<4> t = (int)1;
+    return t;
+}
+
+bit<4> func2() {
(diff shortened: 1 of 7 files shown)


testdata/p4_16_samples/issue3219.p4
```diff
@@ -0,0 +1,9 @@
+bit<4> func1() {
```
fruffy, line 1: Can we make this a test that produces an output? We'll see whether Gauntlet needs fixing..

fruffy (approved):

testdata/p4_16_samples/issue3219.p4
mbudiu-vmw: Done.

PR #3224, opened 2022-04-19
Replace pna direction with register read

kamleshbhalui:
Extract packet direction for PNA from an input port, By creating a Register extern network_port_mask, which will be initialized by P4SDE.
The direction metadata should be replaced by a register read from network_port_mask register and ANDing it with mask for input port

istd.direction => network_port_mask.read(0) & (32w0x1 << istd.input_port)


backends/dpdk/DpdkXfail.cmake
```diff
@@ -96,6 +92,11 @@ p4c_add_xfail_reason("dpdk"
   testdata/p4_16_samples/pna-example-mirror-packet-error3.p4
   )
 
+p4c_add_xfail_reason("dpdk"
+  "declaration not found"
```
mbudiu-vmw, line 96: why is this supposed to fail with p4c-dpdk only?

backends/dpdk/dpdkMetadata.cpp
```diff
@@ -0,0 +1,154 @@
...
+    }
+    // "network_port_mask" name is used in dpdk for initialzing direction port mask
+    // make sure no such decls exist with that name
+    if (mng1.newName(registerInstanceName) != registerInstanceName)
+        ::error(ErrorType::ERR_UNSUPPORTED_ON_TARGET,"decl name %s is reserved for dpdk pna",
+                registerInstanceName);
+}
+
+const IR::Node* DirectionToRegRead::preorder(IR::DpdkAsmProgram *p) {
+        uniqueNames(p);
```
mbudiu-vmw, line 59: too many spaces

backends/dpdk/dpdkMetadata.cpp
```diff
@@ -0,0 +1,154 @@
...
+    left_shift_tmp = mng.newName("left_shift_tmp");
+    registerInstanceName = "network_port_mask";
+
+    P4::MinimalNameGenerator mng1;
+    for (auto decl : p->externDeclarations) {
+        mng1.usedName(decl->name);
+    }
+    // "network_port_mask" name is used in dpdk for initialzing direction port mask
+    // make sure no such decls exist with that name
+    if (mng1.newName(registerInstanceName) != registerInstanceName)
```
mbudiu-vmw, line 53: Perhaps it's best if you add a new api to the name generator to check whether a name is in use.
This works too, but it is a bit ugly.

backends/dpdk/dpdkMetadata.h
```diff
@@ -0,0 +1,65 @@
...
+
+#ifndef BACKENDS_DPDK_DPDKMETADATA_H_
+#define BACKENDS_DPDK_DPDKMETADATA_H_
+
+#include "frontends/common/resolveReferences/referenceMap.h"
+#include "ir/ir.h"
+
+namespace DPDK {
+
+class AddNewMetadataFields : public Transform {
```
mbudiu-vmw, line 25: Do you want to add a comment somewhere describing what these passes collectively do (with a small example)?

mbudiu-vmw (changes requested):

backends/dpdk/DpdkXfail.cmake
kamleshbhalui: it is failing may be due to some bug in compiler frontend. I am just recording this test for fixing it later.

backends/dpdk/dpdkMetadata.cpp
```diff
@@ -42,23 +42,25 @@ void DirectionToRegRead::uniqueNames(IR::DpdkAsmProgram *p) {
...
-    }
     // "network_port_mask" name is used in dpdk for initialzing direction port mask
     // make sure no such decls exist with that name
-    if (mng1.newName(registerInstanceName) != registerInstanceName)
+    registerInstanceName = "network_port_mask";
+    for (auto decl : p->externDeclarations) {
+        usedNames.insert(decl->name);
+    }
+
+    if (usedNames.count(registerInstanceName))
```
mbudiu-vmw, line 53: This works if usedNames contains everything you need to check for. I was hoping that you'd add a new method `MinimalNameGenerator::nameIsUsed(cstring name)`

mbudiu-vmw (approved):

Issue #3225, opened 2022-04-19
Slicing from a cast bool does not compile


fruffy:
The following program does not compile: 

```p4
header H {
    bool  a;
    bit<1>  b;
}

struct Headers {
    ethernet_t eth_hdr;
    H    h;
}

struct Meta {}

control ingress(inout Headers h, inout Meta m, inout standard_metadata_t sm) {
    apply {
        h.h.b = ((bit<1>)h.h.a)[0:0];
    }
}
```
```bash
...
FrontEnd_25_SpecializeGenericFunctions
Writing program to dmp/gauntlet_action_return-bmv2-FrontEnd_25_SpecializeGenericFunctions.p4
P4::TypeChecking_0_ResolveReferences
P4::TypeChecking_1_TypeInference
TableKeyNames_0_TypeChecking
TableKeyNames_1_DoTableKeyNames
FrontEnd_26_TableKeyNames
Writing program to dmp/gauntlet_action_return-bmv2-FrontEnd_26_TableKeyNames.p4
P4::TypeChecking_0_ResolveReferences
P4::TypeChecking_1_TypeInference
ConstantFolding_0_TypeChecking
ConstantFolding_1_DoConstantFolding
ConstantFolding_2_ClearTypeMap
PassRepeated_0_ConstantFolding
P4::TypeChecking_0_ResolveReferences
P4::TypeChecking_1_TypeInference
P4::TypeChecking_2_ApplyTypesToExpressions
P4::TypeChecking_3_ResolveReferences
P4::StrengthReduction_0_TypeChecking
P4::StrengthReduction_1_StrengthReduction
PassRepeated_1_StrengthReduction
PassRepeated_2_Reassociation
P4::TypeChecking_0_ResolveReferences
../../p4c/testdata/p4_16_samples/gauntlet_action_return-bmv2.p4(24): [--Werror=type-error] error: AssignmentStatement
        h.h.b = ((bit<1>)h.h.a)[0:0];
              ^
  ---- Actual error:
  Cannot unify type 'bool' with type 'bit<1>'
  ---- Originating from:
  ../../p4c/testdata/p4_16_samples/gauntlet_action_return-bmv2.p4(24): Source expression 'h.h.a' produces a result of type 'bool' which cannot be assigned to a left-value with type 'bit<1>'
          h.h.b = ((bit<1>)h.h.a)[0:0];
```

Is this correct behavior? If I remove the slice, it does compile. 

[...]

PR #3226, opened 2022-04-19
Strength reduction should not remove some casts

mbudiu-vmw:

### DIFF
--- a/frontends/p4/strengthReduction.cpp
+++ b/frontends/p4/strengthReduction.cpp
@@ -444,10 +444,22 @@ const IR::Node* DoStrengthReduction::postorder(IR::Slice* expr) {
     }
 
     while (auto cast = expr->e0->to<IR::Cast>()) {
-        if (expr->getH() < size_t(cast->expr->type->width_bits())) {
-            expr->e0 = cast->expr;
+        if (auto tb = cast->expr->type->to<IR::Type_Bits>()) {
+            auto type = expr->type->to<IR::Type_Bits>();
+            CHECK_NULL(type);
+            if (tb->isSigned != type->isSigned)
+                // Cannot remove casts that change sign.
+                break;
+            if (expr->getH() < size_t(tb->width_bits())) {
+                expr->e0 = cast->expr;
+            } else {
+                break;
+            }
         } else {
-            break; } }
+            // Cast from a different type.
+            break;
+        }
+    }
 
     // out-of-bound error has been caught in type checking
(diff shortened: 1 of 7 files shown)


mbudiu-vmw:
Fixed #3225

frontends/p4/strengthReduction.cpp
```diff
@@ -444,10 +444,22 @@ const IR::Node* DoStrengthReduction::postorder(IR::Slice* expr) {
     }
 
     while (auto cast = expr->e0->to<IR::Cast>()) {
-        if (expr->getH() < size_t(cast->expr->type->width_bits())) {
-            expr->e0 = cast->expr;
+        if (auto tb = cast->expr->type->to<IR::Type_Bits>()) {
+            auto type = expr->type->to<IR::Type_Bits>();
```
fruffy, line 448: Why not use checkedTo here?

frontends/p4/strengthReduction.cpp
```diff
@@ -444,10 +444,22 @@ const IR::Node* DoStrengthReduction::postorder(IR::Slice* expr) {
...
+            CHECK_NULL(type);
+            if (tb->isSigned != type->isSigned)
```
fruffy, line 450: I always assumed that slices of signed bit strings will produce an unsigned bit string. This should probably ensure that a cast to `int<W>` is converted into a `bit<W>`. A cast from `int<W>` to bit<W> should be acceptable though? 

This is what the spec says: 
>The result is an unsigned bit-string of width m - l + 1, including the bits numbered from l (which becomes the least significant bit of the result) to m (the most significant bit of the result) from the source operand.

mbudiu-vmw: I added another test case which fails without this change; it's in the test file:
`h.h.b = ((bit<1>)h.h.c)[0:0];`

fruffy: But
`((bit<1>)h.h.c)[0:0];`
to
`((bit<1>)h.h.c);` 
should be acceptable, right?

fruffy (approved):

frontends/p4/strengthReduction.cpp
mbudiu-vmw: The previous result was just h.h.c, which failed to typecheck.

fruffy:
The only thing is to to add a new member d to the header and assign c to that one. Otherwise, the assignment is optimized away.

fruffy:
Is this ready to be merged?

PR #3228, opened 2022-04-20
Add check of uniqueness of control plane names

dmatousek:
[Control plane names](https://p4.org/p4-spec/docs/P4-16-v1.2.2.html#sec-control-plane-api-annotations) are not checked for uniqueness. It may lead to various issues in subsequent passes when uniqueness is assumed.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -243,6 +243,7 @@ include_directories (
   ${P4C_SOURCE_DIR}/extensions
   ${P4C_SOURCE_DIR}
   ${P4C_SOURCE_DIR}/test/frameworks/gtest/googletest/include
+  ${P4C_SOURCE_DIR}/test/frameworks/gtest/googlemock/include
   ${P4C_BINARY_DIR}
   )
 add_definitions (-DCONFIG_PREFIX="${CMAKE_INSTALL_PREFIX}")
--- a/frontends/p4/hierarchicalNames.cpp
+++ b/frontends/p4/hierarchicalNames.cpp
@@ -27,14 +27,25 @@ const IR::Node* HierarchicalNames::postorder(IR::Annotation* annotation) {
         return annotation;
 
     cstring name = annotation->getName();
-    if (name.startsWith("."))
-        return annotation;
-    cstring newName = "";
-    for (cstring s : stack)
-        newName += s + ".";
-    newName += name;
-    LOG2("Changing " << name << " to " << newName);
-    annotation = new IR::Annotation(annotation->name, newName);
+    if (!name.startsWith(".")) {
+        cstring newName = "";
+        for (cstring s : stack)
+            newName += s + ".";
+        newName += name;
+        LOG2("Changing " << name << " to " << newName);
+        annotation = new IR::Annotation(annotation->name, newName);
+        name = newName;
+    }
+    // The node the annotation belongs to
+    CHECK_NULL(getContext()->parent);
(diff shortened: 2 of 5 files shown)


jafingerhut:
By any chance, is this intended to address an issue like the one described in this issue?  https://github.com/p4lang/p4c/issues/2755

One of the example programs on that issue has identical `@name` annotations on two different actions, which seems to me like a bug to accept, but note that internal passes in open source p4c create this situation by design right now (which I personally also think is a mistake, and would be better to avoid, but it has been like this for years).

mbudiu-vmw:
See also #3037, which didn't work.

jafingerhut:
I tried building p4c with the proposed changes on this PR after 3 commits, and ran it locally on my system.

It fails for a very simple P4 program with 2 tables, both of which have `NoAction` in their actions list.  You should be able to reproduce this with the following commands:
```
$ git clone https://github.com/jafingerhut/p4-guide
$ cd p4-guide/name-annotations/v1model
$ p4c --target bmv2 --arch v1model --p4runtime-files actions-5-no-annot.p4info.txt actions-5-no-annot.p4 
/usr/local/share/p4c/p4include/core.p4(70): [--Werror=duplicate] error: .NoAction: conflicting control plane name: .NoAction
action NoAction() {}
       ^^^^^^^^
/usr/local/share/p4c/p4include/core.p4(70)
action NoAction() {}
       ^^^^^^^^
```
As far as the original source code given to the compiler, there are no name conflicts at all.

The error messages occur, I believe, because of the way that one or more p4c front end passes duplicates actions in the IR when the same action is an action of multiple P4 tables, and automatically creates identical `@name` annotations on these duplicated actions.

See #2755 for additional notes and details.  I would love to have a meeting of the minds of P4 Runtime API folks, compiler folks, etc. to see what can or should be done about this.  I have ideas described on that issue, but not sure what the best way to go here is.

dmatousek:
Thanks for your feedback and referencing related issues. I agree that there is no clear fix and it has to be further discussed.

jfingerh:
@dmatousek I spoke with some folks at Intel including @hanw about the underlying issue here, and he had a suggestion for an experiment to try: put the @name conflict checking somewhere in the front end _before_ the LocalizeAllActions pass, which is one that we know intentionally creates copies of actions with identical @name annotations.  If there is a @name conflict before that pass, it should be one that the P4 developer themselves wrote.

The only disadvantage I can think of to such an approach is that now P4_16 source code produced as the output of passes after LocalizeAllActions could cause errors when used as input to p4c.  I do not know how serious that is, but perhaps having a command line option that skips the new @name conflict check would help with that.

jafingerhut:
> @dmatousek I spoke with some folks at Intel including @hanw about the underlying issue here, and he had a suggestion for an experiment to try: put the @name conflict checking somewhere in the front end _before_ the LocalizeAllActions pass, which is one that we know intentionally creates copies of actions with identical @name annotations. If there is a @name conflict before that pass, it should be one that the P4 developer themselves wrote.
> 
> The only disadvantage I can think of to such an approach is that now P4_16 source code produced as the output of passes after LocalizeAllActions could cause errors when used as input to p4c. I do not know how serious that is, but perhaps having a command line option that skips the new @name conflict check would help with that.

I have created a PR that attempts to use the approach described in my earlier comment quoted above.  It can be found here: https://github.com/p4lang/p4c/pull/4951

PR #3232, opened 2022-04-20
Improve diagnostic messages for `AssignmentStatement`, `Operation_Binary`, `ListExpression`, and `StructExpression`

kfcripps:


testdata/p4_14_samples_outputs/parser_pragma2.p4-stderr
```diff
@@ -1,6 +1,6 @@
 parser_pragma2.p4(18): [--Wwarn=parser-transition] warning: SelectCase: unreachable
         0xAB00 : Cowles;
         ^^^^^^
-parser_pragma2.p4(16): [--Wwarn=parser-transition] warning: 'ListExpression': transition does not depend on select argument
+parser_pragma2.p4(16): [--Wwarn=parser-transition] warning: '{ tmp }': transition does not depend on select argument
```
mbudiu-vmw, line 4: This is one problem with this improvement: this error message no longer makes sense.

testdata/p4_16_errors_outputs/issue2317.p4-stderr
```diff
@@ -1,3 +1,3 @@
-issue2317.p4(16): [--Werror=type-error] error: ==: cannot be applied to action results
+issue2317.p4(16): [--Werror=type-error] error: f == f: cannot be applied to action results
```
mbudiu-vmw, line 1: this one is also confusing

mbudiu-vmw (commented):
This doesn't work so well when the printed expression is not one that appeared in the source program. This is not a problem related to this PR, but this PR makes the problem a bit worse.

testdata/p4_14_samples_outputs/parser_pragma2.p4-stderr
kfcripps: I noticed that too and I am not sure of a good way to fix this. Is this problem unique to `ListExpression` components, or is it a problem when printing compiler-generated `Expression`s in other places, too?

kfcripps: I noticed that too and I am not sure of a good way to fix this. Is this problem unique to `ListExpression` components, or is it a problem when printing compiler-generated `Expression`s in other places, too?

testdata/p4_16_errors_outputs/issue2317.p4-stderr
kfcripps: This one can be fixed by enhancing `MethodCallExpression`'s `toString()` method. This change would affect ~110 tests so I opted to leave that for a separate PR, but I can make that change here if you prefer.
This is not unique to printing `MethodCallExpression`s used in `==` operations, though. The `MethodCallExpression` will be printed as `f` in other places too.

kfcripps: This one can be fixed by enhancing `MethodCallExpression`'s `toString()` method. This change would affect ~110 tests so I opted to leave that for a separate PR, but I can make that change here if you prefer.
This is not unique to printing `MethodCallExpression`s used in `==` operations, though. The `MethodCallExpression` will be printed as `f` in other places too.

testdata/p4_14_samples_outputs/parser_pragma2.p4-stderr
mbudiu-vmw: No, it happens in other places as well.

kfcripps:
superseded by https://github.com/p4lang/p4c/pull/4353

PR #3234, opened 2022-04-20
Do not allow casts to struct types

### LINKED ISSUE #3233: Casting to struct is allowed? but not always?
Take:
```
struct s
{
  bit t;
  bit t1;
}

s func(bit t, bit t1)
{
  return (s)(s)(s){t, t1};
}
s func1(s t1)
{
  return (s)(s)(s)t1;
}
```
func is allowed but func1 is not allowed? I would assume both would be rejected. It seems like the multiple casts are being removed leaving only the last one before doing type checking.

mbudiu-vmw:
Fixes #3233


fruffy (approved):
So is there a spec change planned for this?

mbudiu-vmw:
I think this is actually in line with the current spec, which only allows casts for base types.

PR #3241, opened 2022-04-22
Fix error message for cast to 'type'

### LINKED ISSUE #3221: error message for casting with "type" defined types needs help
Take:
```
typedef bit<1> b1;
typedef bit<2> b2;
type b1 t1;
type b2 t2;

t1 func(b2 a)
{
  return (t1)a;
}
```
Currently we get:
```
casttype.p4(8): [--Werror=type-error] error: cast
  return (t1)a;
         ^^^^^
  ---- Actual error:
  Cannot unify type 'bit<2>' with type 'bit<1>'
  ---- Originating from:
  Cannot cast from 'bit<2>' to 'bit<1>'
```
But casting betweeen bit<2> and bit<1> is valid (according to 8.9.1) but what is not valid is cast between bit<2> and t1 (which is a type defined via type of bit<1>).

Either we make type act similar to enum with an underlying type (which is a specification change) or the error message needs to be improved.

mbudiu-vmw:

Fixes #3221

### DIFF
--- a/frontends/p4/typeChecking/typeChecker.cpp
+++ b/frontends/p4/typeChecking/typeChecker.cpp
@@ -2650,7 +2650,7 @@ const IR::Node* TypeInference::postorder(IR::Cast* expression) {
             destType = getTypeType(destType->to<IR::Type_Newtype>()->type);
 
         auto tvs = unify(expression, destType, sourceType,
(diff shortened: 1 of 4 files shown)


rst0git (approved):
LGTM

PR #3245, opened 2022-04-23
DPDK Backend: PNA Pass metadata support

usha1830:
DPDK implements pass metadata using "_recircid_" instruction.

All occurrences of pass metadata usage should be prepended with _recircid_ instruction to fetch the latest _pass_id_ from the target.

_"recircid"_ instruction takes a metadata type as argument to fetch the _pass_id_ into that metadata.

### DIFF
--- a/backends/dpdk/dpdkMetadata.h
+++ b/backends/dpdk/dpdkMetadata.h
@@ -69,5 +69,17 @@ class DirectionToRegRead : public Transform {
     replaceDirectionWithRegRead(IR::IndexedVector<IR::DpdkAsmStatement> stmts);
 };
 
+class PrependPassRecircId : public Transform {
+    IR::IndexedVector<IR::DpdkAsmStatement> newStmts;
+
+ public:
+    PrependPassRecircId() {}
+    bool isPass(const IR::Member *m);
+    const IR::Node *postorder(IR::DpdkAction *a);
+    const IR::Node *postorder(IR::DpdkListStatement *l) override;
+    IR::IndexedVector<IR::DpdkAsmStatement>
+    prependPassWithRecircid(IR::IndexedVector<IR::DpdkAsmStatement> stmts);
+};
+
 }  // namespace DPDK
 #endif  // BACKENDS_DPDK_DPDKMETADATA_H_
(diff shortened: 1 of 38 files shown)


backends/dpdk/dpdkMetadata.h
```diff
@@ -69,5 +69,17 @@ class DirectionToRegRead : public Transform {
     replaceDirectionWithRegRead(IR::IndexedVector<IR::DpdkAsmStatement> stmts);
 };
 
+class PrependPassRecircId : public Transform {
```
mbudiu-vmw, line 72: The comment that accompanies this PR should be present in the code too, here.

mbudiu-vmw: I am not familiar with PNA, can you also explain how this recirc works?

mbudiu-vmw (changes requested):

backends/dpdk/dpdkMetadata.h
usha1830: The `recirculate() `extern corresponds to "_recirculate_" instruction and marks the packet for recirculation. This means that when packet reaches _tx_ instruction, it is reinjected into the pipeline for another full pass instead of being sent to the output port.

The same packet can be recirculated multiple times, with the recirculation pass number retrieved by the "_recircid_" instruction which corresponds to "_pass_" metadata. The pass number can be used by the program to execute different code on the same packet in different pipeline passes. The packet meta-data is preserved between the pipeline passes.

usha1830: DPDK reference patch for packet recirculation 
https://patches.dpdk.org/project/dpdk/patch/20220406185535.73084-6-cristian.dumitrescu@intel.com/

mbudiu-vmw:
There is no "recirculate" extern in your P4 examples.
What triggers recirculation?

usha1830:
> There is no "recirculate" extern in your P4 examples. What triggers recirculation?

Updated the examples.

mbudiu-vmw (approved):

usha1830:
@mbudiu-vmw Can this PR be merged?.

PR #3257, opened 2022-04-27
[dpdk] Eliminate header copy

kamleshbhalui:
Dpdk does not support direct header copy.
So, Eliminate temporary copies of header, which generally populated after inlining, if it;s not temporary copy then transform direct header copy to element wise copy.

### DIFF
--- a/backends/dpdk/dpdkArch.cpp
+++ b/backends/dpdk/dpdkArch.cpp
@@ -2263,4 +2263,95 @@ void CollectAddOnMissTable::postorder(const IR::MethodCallStatement *mcs) {
     return;
 }
 
+bool ElimHeaderCopy::isHeader(const IR::Expression* e) {
+    auto type = typeMap->getType(e);
+    if (type)
+        return type->is<IR::Type_Header>() && !e->is<IR::MethodCallExpression>();
+    return false;
+}
+
+const IR::Node* ElimHeaderCopy::preorder(IR::AssignmentStatement* as) {
+    if (isHeader(as->left) && isHeader(as->right)) {
+        if (auto path = as->left->to<IR::PathExpression>()) {
+            replacementMap.insert(std::make_pair(path->path->name.name,
+                                   as->right->to<IR::Member>()));
+            return new IR::EmptyStatement();
+        } else if (auto path = as->right->to<IR::PathExpression>()) {
+            replacementMap.insert(std::make_pair(path->path->name.name,
+                                  as->left->to<IR::Member>()));
+            return  new IR::EmptyStatement();
+        } else {
+            IR::ID methodName;
+            IR::IndexedVector<IR::StatOrDecl> components;
+            // copy validity flag
+            auto isValid = new IR::Member(as->right->srcInfo, as->right,
+                              IR::ID(IR::Type_Header::isValid));
+            auto result = new IR::MethodCallExpression(as->right->srcInfo, IR::Type::Boolean::get(),
+                                                   isValid);
+            typeMap->setType(isValid, new IR::Type_Method(IR::Type::Boolean::get(),
+                                          new IR::ParameterList(), IR::Type_Header::isValid));
+            auto method = new IR::Member(as->left->srcInfo, as->left,
+                                         IR::ID(IR::Type_Header::setValid));
+            auto mc = new IR::MethodCallExpression(as->left->srcInfo, method,
+                          new IR::Vector<IR::Argument>());
+            typeMap->setType(method, new IR::Type_Method(IR::Type_Void::get(),
+                                new IR::ParameterList(), IR::Type_Header::setValid));
+            components.push_back(new IR::MethodCallStatement(mc->srcInfo, mc));
+            auto ifTrue = components;
+            components.clear();
+            auto method1 = new IR::Member(as->left->srcInfo, as->left,
+                                              IR::ID(IR::Type_Header::setInvalid));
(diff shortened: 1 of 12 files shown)


backends/dpdk/dpdkArch.cpp
```diff
@@ -2253,4 +2253,63 @@ void CollectAddOnMissTable::postorder(const IR::MethodCallStatement *mcs) {
...
+            replacementMap.insert(std::make_pair(path->path->name.name,
+                                   as->right->to<IR::Member>()));
+            return new IR::EmptyStatement();
+        } else if (auto path = as->right->to<IR::PathExpression>()) {
+            replacementMap.insert(std::make_pair(path->path->name.name,
+                                  as->left->to<IR::Member>()));
+            return  new IR::EmptyStatement();
+        } else {
+            // both are non temporary header instance do element wise copy
+            IR::IndexedVector<IR::StatOrDecl> components;
```
mbudiu-vmw, line 2275: This looks wrong, since it won't copy the valid flag.

backends/dpdk/dpdkArch.cpp
```diff
@@ -2253,4 +2253,63 @@ void CollectAddOnMissTable::postorder(const IR::MethodCallStatement *mcs) {
...
+    auto type = typeMap->getType(e);
+    if (type)
+        return type->is<IR::Type_Header>() && !e->is<IR::MethodCallExpression>();
+    return false;
+}
+
+const IR::Node* ElimHeaderCopy::preorder(IR::AssignmentStatement* as) {
+    if (isHeader(as->left) && isHeader(as->right)) {
+        if (auto path = as->left->to<IR::PathExpression>()) {
+            replacementMap.insert(std::make_pair(path->path->name.name,
```
mbudiu-vmw, line 2266: what is happening here?

backends/dpdk/dpdkArch.h
```diff
@@ -926,6 +926,20 @@ class CollectErrors : public Inspector {
     }
 };
 
+// Eliminate temporary copies of header, which generally populated after inlining,
```
mbudiu-vmw, line 929: We have a similar pass for copying structures, but it won't really apply here. It handles cases of nested structures.
Does this pass handle headers with structs inside?
This comment should use ///.
Also, please add a P4 code example showing the before and after.

backends/dpdk/dpdkArch.cpp
```diff
@@ -2253,4 +2253,63 @@ void CollectAddOnMissTable::postorder(const IR::MethodCallStatement *mcs) {
...
+}
+
+const IR::Node* ElimHeaderCopy::preorder(IR::MethodCallExpression *me) {
+    auto m = me->method->to<IR::Member>();
+    if (!m)
+        return me;
+    if (!isHeader(m->expr))
+        return me;
+    auto expr = m->expr->to<IR::PathExpression>();
+    if (expr && replacementMap.count(expr->path->name.name)) {
```
mbudiu-vmw, line 2294: I cannot guess what is happening here, can you please add comments explaining?

mbudiu-vmw (changes requested):

backends/dpdk/dpdkArch.h
kamleshbhalui: >Does this pass handle headers with structs inside?
Actually structs inside gets flatten by midend 
i.e.
```
struct foo {
    EthernetAddress dstAddr;
    EthernetAddress srcAddr;
    bit<16>         etherType;
}

header ethernet_t {
    foo f;
    EthernetAddress dstAddr;
    EthernetAddress srcAddr;
    bit<16>         etherType;
}
```

becomes
```
header ethernet_t {
    bit<48> _f_dstAddr0;
    bit<48> _f_srcAddr1;
    bit<16> _f_etherType2;
    bit<48> _dstAddr3;
    bit<48> _srcAddr4;
    bit<16> _etherType5;
}
```

backends/dpdk/dpdkArch.cpp
kamleshbhalui: it's for 
replacing eth_0.setInvalid()
with hdr.ethernet.setInvalid()

backends/dpdk/dpdkArch.cpp
kamleshbhalui: It's for populating replacement map by keeping temporary header name as key and source of assignment statement as value.
i.e.
for below assignment statement
 eth_0 = hdr.ethernet
replacement map will be like
replacementMap["eth_0"] = hdr.ethernet;

later on all uses of eth_0 will be replace with there value in replacementMap.

backends/dpdk/dpdkArch.cpp
kamleshbhalui: done

backends/dpdk/dpdkArch.cpp
kamleshbhalui: added

backends/dpdk/dpdkArch.cpp
mbudiu-vmw: I would add this information as a comment in the header file where the replacementMap is defined.

mbudiu-vmw (approved):

mbudiu-vmw (approved):

kamleshbhalui:
@mbudiu-vmw Please merge it. is there any policy regarding who can have write access to p4lang/p4c repo ?

PR #3261, opened 2022-04-28
Update used psabpf library for PTF tests (PSA/eBPF)

tatry:
It adds support, compared to earlier version, for:
- DirectMeter;
- DirectCounter;
- set default action;
- get table entry (limited to single entry for table without implementation).

CC: @osinstom @kmateuszssak

### DIFF
--- a/backends/ebpf/tests/ptf/common.py
+++ b/backends/ebpf/tests/ptf/common.py
@@ -42,6 +42,7 @@ TEST_PIPELINE_ID = 1
 TEST_PIPELINE_MOUNT_PATH = "/sys/fs/bpf/pipeline{}".format(TEST_PIPELINE_ID)
 PIPELINE_MAPS_MOUNT_PATH = "{}/maps".format(TEST_PIPELINE_MOUNT_PATH)
 
+
 def xdp2tc_head_not_supported(cls):
     if cls.xdp2tc_mode(cls) == 'head':
         cls.skip = True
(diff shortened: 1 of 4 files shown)


backends/ebpf/tests/ptf/common.py
```diff
@@ -208,33 +242,75 @@ def table_write(self, method, table, keys, action=0, data=None, priority=None, r
...
+    def table_get(self, table, keys, indirect=False):
+        cmd = "psabpf-ctl table get pipe {} {} ".format(TEST_PIPELINE_ID, table)
+        if indirect:
+            # TODO: cmd = cmd + "ref "
+            self.fail("support for indirect table not supported yet")
+        cmd = cmd + self._table_create_str_from_key(keys=keys)
+        _, stdout, _ = self.exec_ns_cmd(cmd, "Table get entry failed")
+        return json.loads(stdout)[table]
+
+    def table_verify(self, table, keys, action=0, priority=None, data=None, references=None,
```
mbudiu-vmw, line 281: Some of these functions could use a documentation """

backends/ebpf/tests/ptf/common.py
```diff
@@ -208,33 +242,75 @@ def table_write(self, method, table, keys, action=0, data=None, priority=None, r
...
-            for d in data:
-                cmd = cmd + "{} ".format(d)
+        cmd = cmd + self._table_create_str_from_key(keys=keys)
+        cmd = cmd + self._table_create_str_from_data(data=data, counters=counters, meters=meters)
         if priority:
             cmd = cmd + "priority {}".format(priority)
         self.exec_ns_cmd(cmd, "Table {} failed".format(method))
 
-    def table_add(self, table, keys, action=0, data=None, priority=None, references=None):
+    def table_add(self, table, keys, action=0, data=None, priority=None, references=None,
```
mbudiu-vmw, line 251: have you considered adding type annotation to your python? 
I find they help a lot with readability.
So actions are encoded as integers? I wonder whether p4info could be used to use action symbolic names.

tatry: > have you considered adding type annotation to your python? I find they help a lot with readability.

At some point of time I had an idea to add that annotations, but I abandoned it. I don't remember why. 

> So actions are encoded as integers? I wonder whether p4info could be used to use action symbolic names.

Yes, exactly, actions are encoded as integers. We decided that p4info can't be used at the level of `psabpf[-cli]`, so every think what we can comes from BTF side. DirectCounters and DirectMeters are recognized by name, so for actions it should be also possible, but is not implemented yet.

osinstom:
@mbudiu-vmw can we merge this?

mbudiu-vmw (approved):

PR #3262, opened 2022-04-28
prefer LLD over gold as linker

grg:
LLD is much faster than gold in testing, so prefer LLD if it is found.

Test results for a debug build with the Intel Tofino backend after
touching a single file:
 * gold: 313.410 s (averaged across 10 runs)
 * lld: 151.783 s (averaged across 10 runs)


grg:
Closing PR as @vlstill has a better PR.

PR #3263, opened 2022-04-28
Add support for LLD linker, cmake options for LLD and Gold

vlstill:
A somewhat different approach to the same problem as  in #3262, where @grg wrote:

> LLD is much faster than gold in testing, so prefer LLD if it is found.
>
> Test results for a debug build with the Intel Tofino backend after
> touching a single file:
> 
>     gold: 313.410 s (averaged across 10 runs)
>     lld: 151.783 s (averaged across 10 runs)
> 

The difference is mainly in the options to enable/disable lookup for linkers.

Tested with GCC & Clang. Since the detection fails if the `-fuse-ld=???` is not supported by the given gcc/clang version there is no need for version checks.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -217,25 +217,38 @@ add_cxx_compiler_option ("-Wno-deprecated-declarations")
 # Enable/disable runtime exceptions on nullpointer dereferencing
 # add_cxx_compiler_option ("-fsanitize=null")
 
-# If we're on GCC, use the gold linker if available.
-if(CMAKE_COMPILER_IS_GNUCC)
+# If we're on GCC or Clang, use the prefer LLD or Gold linker if available.
+set(BUILD_LINK_WITH_GOLD ON CACHE BOOL "Use Gold linker for build if available")
+set(BUILD_LINK_WITH_LLD ON CACHE BOOL "Use LLD linker for build if available (overrides BUILD_LINK_WITH_GOLD)")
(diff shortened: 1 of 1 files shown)


grg (approved):
Looks good to me.

It's slightly more verbose than my proposal, but it's better for at least two reasons:
1. It eliminates some duplication.
2. Total lines of code should be smaller, and with fewer levels of nested ifs, if we have 3+ linker options.

fruffy (approved):

CMakeLists.txt
```diff
@@ -217,25 +217,38 @@ add_cxx_compiler_option ("-Wno-deprecated-declarations")
...
-            COMMAND ${CMAKE_C_COMPILER} -fuse-ld=gold -Wl,--version
-            ERROR_QUIET OUTPUT_VARIABLE LD_VERSION)
-    if ("${LD_VERSION}" MATCHES "GNU gold")
-        set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -fuse-ld=gold")
-        set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fuse-ld=gold")
-        message(STATUS "Using the GNU gold linker.")
-    else ()
-        message(STATUS "Using the default system linker.")
+    if (BUILD_STATIC_RELEASE AND CMAKE_CXX_COMPILER_ID STREQUAL "GNU" AND CMAKE_CXX_COMPILER_VERSION VERSION_GREATER 8.0)
+        add_cxx_compiler_option ("-flto")
```
davidbolvansky, line 227: GCC LTO does not work with LLD.

PR #3265, opened 2022-04-29
Do not crash on illegal uses of 'this'

### LINKED ISSUE #3264: this used in a non method causes an compiler bug message to show up
Take:
```
void func()
{
  bit t;
  t = this;
}
```
We get:
```
enumcast.p4(4): [--Werror=invalid] error: This can only be used in the definition of an abstract method
  t = this;
      ^^^^
In file: /home/apinski/src/p4/p4c/frontends/common/resolveReferences/referenceMap.cpp:52
Compiler Bug: /home/apinski/src/p4/p4c/frontends/common/resolveReferences/referenceMap.cpp:52: Null decl

```
The Compiler bug message should not happen :).

mbudiu-vmw:
Fixes #3264


rst0git (approved):
LGTM

PR #3267, opened 2022-04-29
Add UPF sample program for PSA/eBPF backend

osinstom:


backends/ebpf/psa/examples/upf.p4
```diff
@@ -0,0 +1,609 @@
+/*
```
mbudiu-vmw, line 1: Is there anything in these programs that is specific to ebpf, or they should work in principle on any target that implements PSA? If the latter is true perhaps the programs should be in a "psa" directory instead.

mbudiu-vmw (approved):

PR #3268, opened 2022-04-29
Improvement of p4c-graphs backend

st22nestrel:
Class Controls - added visiting of nodes with key and actions, which extends table node processing.
Class Parsers rewritten, so it now uses boost library for generation of graphs.
Created class Graph_visitor, which implements functionality to:
- create fullGraph in 'dot' format from boost graphs of program blocks
- create fullGraph in predefined 'json' format
- output separate graphs for each program block in 'dot' format (old functionality)

Updated Readme accordingly


mbudiu-vmw:
Please check the build failures.

backends/graphs/README.md
```diff
@@ -1,7 +1,9 @@
 # Graphs Backend
 
 This backend produces visual representations of a P4 program as dot files. For
-now it only supports the generation of graphs for top-level control and parser blocks.
+now it supports the generation of graphs for top-level control and parser blocks,
+generation of fullGraph, which merges graphs for top-level program blocks and
```
mbudiu-vmw, line 5: what is really a "fullGraph"?

backends/graphs/README.md
```diff
@@ -23,6 +25,43 @@ cd out
 dot <name>.dot -Tpng > <name>.png
 ```
 
+For generation of dot fullGraph, use option `--fullGraph` and 
+for generation of fullGraph represented in json, use option `--jsonOut`.
+
+## Format of json output
```
mbudiu-vmw, line 31: Are there any tools that consume this json?

backends/graphs/README.md
```diff
@@ -31,13 +70,18 @@ test program:
 
 ![Flowlet switching ingress graph](resources/flowlet_switching-bmv2.ingress.png)
 
+Here is the graph generated for the egress control block of the
```
mbudiu-vmw, line 73: I find the old graph more readable; the new one wastes too much vertical space.

backends/graphs/resources/firewall.fullGraph.json
```diff
@@ -0,0 +1,496 @@
...
+          "name" : "reject",
+          "type" : "state",
+          "type_enum" : 8
+        }
+      ],
+      "transitions" : [
+        {
+          "from" : 0,
+          "to" : 1,
+          "cond" : ""
```
mbudiu-vmw, line 43: why have all these empty cond fields?

backends/graphs/README.md
```diff
@@ -31,13 +70,18 @@ test program:
...
+[flowlet_switching-bmv2.p4](../../testdata/p4_16_samples/flowlet_switching-bmv2.p4)
+test program:
+
+![Flowlet switching egress graph](resources/flowlet_switching-bmv2.egress.png)
```
mbudiu-vmw, line 77: some of the displayed nodes are very wide, is there a way to fix that?

mbudiu-vmw (changes requested):

backends/graphs/README.md
st22nestrel: There are no tools that consume this json. 
I just have a python script that generates the fullGraph in dot format from this json input. But it is the same fullGraph as the one generated from *p4c-graphs* backend.

The idea behind fullGraph in json was that it could be manipulated more easily than fullGraph in dot, and used for further processing of this graph. But no such tools were implemented yet.

backends/graphs/resources/firewall.fullGraph.json
st22nestrel: When transition between nodes has no label, 'cond' field is empty (has value "").

If that is a problem, I could remove addition of this entry when value is "".

backends/graphs/README.md
st22nestrel: I can set limit for length of string to some reasonable number, like 25 chars?

Was unsure whether too lengthy nodes are a problem, as they are also created by old solution.

backends/graphs/README.md
st22nestrel: It merges all generated boost graphs of parsers and control blocks into one graph.
This graph enables users to easily observe control flow of full program.

backends/graphs/README.md
st22nestrel: I uploaded the script here [https://gist.github.com/st22nestrel/4d23d028b0b211c91e86a194c1269e28]()

backends/graphs/resources/firewall.fullGraph.json
st22nestrel: Issue fixed

mbudiu-vmw (approved):

PR #3275, opened 2022-05-02
Test for isValid as a header field

mbudiu-vmw:
From https://github.com/p4lang/p4-spec/issues/1060


fruffy (approved):

fruffy:
Unrelated question. @rst0git do you have a hunch why the fedora tests run so much faster than the Ubuntu docker tests? 17 minutes vs 40+ minutes. It might actually make sense to ditch the container then and build the behavioral model from scratch...

PR #3278, opened 2022-05-02
Correctly handle don't care named arguments 

### LINKED ISSUE #3274: Compiler does not handle correctly don't care arguments
This was discovered when fixing https://github.com/p4lang/p4-spec/issues/1066
Here are two examples that trigger different bugs:
```
extern void f(in bit<32> x, out bit<16> y);

void x() {
    f(x = 1, y = _);
}

control c() {
    apply {
        x();
    }
}

control _c();
package top(_c _c);

top(c()) main;
```

The above one inserts the temporary argument in the wrong place.

```
extern void f(in bit<32> x, out bit<16> y);

void x() {
    f(x = 1, y = _);
}

control c() {
    apply {
        x();
    }
}

control _c();
package top(_c _c);

top(c()) main;
```

Here the def-use analysis deletes the function call although it may have side-effects.

mbudiu-vmw:

Fixes #3274 
This also fixes a bug where functions were considered to be side-effect-free if they had all parameters 'in'. That is wrong, such a function can still call extern functions that have side-effects.


fruffy (approved):
This is missing the ref files. Some of the ref file changes are hard to read, but Gauntlet is at least not complaining.

mbudiu-vmw:
I started to make some changes to improve the quality of the generated code, which turned into something bigger. I will split this into two PRs, a small one here with correct outputs and another one which improves code quality by running def-use after inlining again.

Issue #3273, opened 2022-05-02
Compiler bug when using int as return type for abstract function in extern object


apinski-cavium:
Take:
```
extern e {
  e();
  abstract int f();
}

e() t = {
  int f()
  {
    return 1;
  }
};
```
Currently we get:
```
In file: /home/apinski/src/p4/p4c/frontends/p4/typeChecking/typeChecker.cpp:1017
Compiler Bug: int/1 -> <Type_InfInt>(46): expected no type variables
```

mbudiu-vmw:
Fixed by #3280

PR #3280, opened 2022-05-02
Forbid methods to return int or string

mbudiu-vmw:

### DIFF
--- a/testdata/p4_16_errors_outputs/string-e1.p4-stderr
+++ b/testdata/p4_16_errors_outputs/string-e1.p4-stderr
@@ -1,9 +1,15 @@
 string-e1.p4(2): [--Werror=type-error] error: Field 's' of 'struct S' cannot have type 'string'
     string s;
            ^
+string-e1.p4(2)
+    string s;
+    ^^^^^^
 string-e1.p4(1): [--Werror=type-error] error: Error while analyzing struct S
 struct S {
        ^
 string-e1.p4(7): [--Werror=type-error] error: v: Cannot declare variables with type string
         string v;
         ^^^^^^^^^
+string-e1.p4(7)
+        string v;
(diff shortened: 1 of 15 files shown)


fruffy (approved):
What is the context of this PR? Is this a spec issue?

testdata/p4_16_errors_outputs/string-e1.p4-stderr
```diff
@@ -1,9 +1,15 @@
 string-e1.p4(2): [--Werror=type-error] error: Field 's' of 'struct S' cannot have type 'string'
     string s;
            ^
+string-e1.p4(2)
```
fruffy, line 4: These error messages make no sense to me.

mbudiu-vmw:
We only support compile-time constant values of type `int` or `string`, so even if a method typechecks you probably can't call it, since it won't be evaluated at compilation time.

testdata/p4_16_errors_outputs/string-e1.p4-stderr
mbudiu-vmw: 'string' is not an allowed type for a struct field. How would you write the error message?

fruffy: Why is the 
```
string-e1.p4(2)
    string s;
```
duplicated?

mbudiu-vmw: First the message points to the field name, then to the type name.

fruffy: I guess it would be nice to add these labels, if possible. E.g., 
Field:  string s
Type: string s

mbudiu-vmw:
Fixes #3273

PR #3290, opened 2022-05-07
DPDK Backend: Connection tracking support

usha1830:
Connection tracking support for PNA needs the following extern functions to be supported.

- `add_entry` extern with additional timeout parameter
- `restart_expire_timer`
- `set_entry_expire_timer`

DPDK provides the following corresponding instructions for connection tracking

- `learn` instruction with timeout parameter (The second parameter of this instruction "action_params" is optional in DPDK)
- `rearm` instruction
- `rearm` instruction with timeout parameter

The timeout parameter is the index in timeout structure defined in the learner table.

All of the above instruction expect all the arguments in metadata and hence if any argument is constant, it should be moved to metadata.

These externs have constraints on where these externs can be called from. Added a new pass to check all these constraints.
The existing convertToDpdk pass is enhanced to process these new externs and generate equivalent assembly spec.

DPDK reference patch describing the rearm and learn instructions:
https://patches.dpdk.org/project/dpdk/patch/20220421155951.31811-3-cristian.dumitrescu@intel.com/

@jafingerhut 's Connection tracking example from [pna](https://github.com/jfingerh/pna/blob/7da8a57c57262cc77c790e1610fea6a81d6af458/examples/pna-example-tcp-connection-tracking.p4) repo is added as a test here

### DIFF
--- a/backends/dpdk/dpdkArch.cpp
+++ b/backends/dpdk/dpdkArch.cpp
@@ -2239,6 +2239,12 @@ void CollectAddOnMissTable::postorder(const IR::P4Table* t) {
                     default_action);
         }
     }
+    if (use_add_on_miss) {
+        for (auto action : t->getActionList()->actionList) {
+            auto action_decl = refMap->getDeclaration(action->getPath())->to<IR::P4Action>();
+            structure->learner_action_table.emplace(action_decl->externalName(), t);
+        }
+    }
 }
 
 void CollectAddOnMissTable::postorder(const IR::MethodCallStatement *mcs) {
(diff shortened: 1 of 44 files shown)


backends/dpdk/dpdkArch.cpp
```diff
@@ -2255,14 +2261,84 @@ void CollectAddOnMissTable::postorder(const IR::MethodCallStatement *mcs) {
...
+                        return;
+                    } else {
+                        auto use_add_on_miss = expr->to<IR::BoolLiteral>()->value;
+                        if (use_add_on_miss)
+                            isValidExternCall = true;
+                    }
+                }
+            }
+        }
+        if (!isValidExternCall) {
```
mbudiu-vmw, line 2333: should this be after the next brace?

backends/dpdk/dpdkHelpers.cpp
```diff
@@ -768,26 +768,67 @@ bool ConvertStatementToDpdk::preorder(const IR::MethodCallStatement *s) {
...
+                IR::ID tmo(refmap->newName("timeout_id"));
+                auto timeout = new IR::Member(new IR::PathExpression("m"), tmo);
+                metadataStruct->fields.push_back(new IR::StructField(tmo, timeout_id->type));
+                add_instr(new IR::DpdkMovStatement(timeout, timeout_id));
+                timeout_id = timeout;
+            }
             if (param->is<IR::Member>()) {
                 auto argument = param->to<IR::Member>();
-                add_instr(new IR::DpdkLearnStatement(action_name, argument));
+                add_instr(new IR::DpdkLearnStatement(action_name, timeout_id, argument));
```
mbudiu-vmw, line 792: here the argument order is different. not a big deal, but consistency could make things simpler.

backends/dpdk/dpdkHelpers.cpp
```diff
@@ -768,26 +768,67 @@ bool ConvertStatementToDpdk::preorder(const IR::MethodCallStatement *s) {
...
+                add_instr(new IR::DpdkLearnStatement(action_name, timeout_id, learnMember));
             } else {
                 ::error(ErrorType::ERR_UNEXPECTED, "%1%: unhandled function", s);
             }
+        } else if (a->method->name == "restart_expire_timer") {
+            add_instr(new IR::DpdkRearmStatement());
+        } else if (a->method->name == "set_entry_expire_time") {
+            auto args = a->expr->arguments;
+            if (args->size() != 1) {
+                ::error(ErrorType::ERR_UNEXPECTED, "Expected 1 arguments for %1%",
```
mbudiu-vmw, line 816: 1 argument

mbudiu-vmw (approved):

backends/dpdk/dpdkArch.cpp
usha1830: Updated.

backends/p4test/CMakeLists.txt
```diff
@@ -108,6 +108,9 @@ set (P4_XFAIL_TESTS
   # These programs uses conditional execution inside action body which is currently
   # unsupported
   testdata/p4_16_samples/psa-dpdk-action-jumpOpt.p4
+  testdata/p4_16_samples/pna-example-tcp-connection-tracking.p4
```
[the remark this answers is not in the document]
usha1830: Done.

backends/dpdk/dpdkHelpers.cpp
usha1830: `add_entry` has 3 parameters in pna.p4. None of the parameters are optional.
Learn instruction has 2 variants. 
```
learn <action_name> <timeout>                     /// when action_params are empty
learn <action_name> <action_params> <timeout>     /// when action params are non-empty
```
<action_param> is optional in the learn instruction, hence kept it at the end.

midend/predication.h
```diff
@@ -113,10 +113,6 @@ class Predication final : public Transform {
     // false for statements that are not dependent.
     std::map<cstring, bool> isStatementDependent;
     const IR::Statement* error(const IR::Statement* statement) const {
-        if (inside_action && ifNestingLevel > 0)
```
mbudiu-vmw, line 116: If you want to delete this code you should delete the whole function and its calls; the function does nothing now.
But I think this should be guided by a policy flag: "report errors when a statement cannot be converted".
Then invoke it with the flag set to "no" in your compiler.

testdata/p4_16_samples/issue513.p4
```diff
@@ -55,8 +55,6 @@ control cIngress(inout Parsed_packet hdr,
     action foo() {
         meta.b = meta.b + 5;
         if (meta.b > 10) {
-            // Next line causes error for p4test: MethodCallStatement:
```
mbudiu-vmw, line 58: Why delete this comment?
Why is the test still failing after your change?

midend/predication.h
mbudiu-vmw: Finally, you are not even using this pass in your backend, why are you modifying it?

mbudiu-vmw (changes requested):

backends/p4test/CMakeLists.txt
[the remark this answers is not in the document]
usha1830: I think I misunderstood this comment and made the recent changes to midend/predication.h by removing the check for conditional execution inside action body.

Some of the tests added in this PR fail with p4test backend because of use of conditional execution within action. So, I added them to XFAIL list for p4test.

Shall I revert back to that, .i.e., no change to the predication.h and adding back the tests with conditional execution in action as XFAIL here?

midend/predication.h
usha1830: Yes. DPDK backend does not use this pass. Wondering if I should still modify to use the policy flag and call it appropriately from p4test backend?

mbudiu-vmw: As you wish

usha1830: On a closer look at predication pass, I think we can leave it as it is now because the error is  thrown for the statements which prevent the if-else to Mux conversion.
I reverted the changes made to predication pass.

mbudiu-vmw (approved):

PR #3294, opened 2022-05-09
ordered_(set|map): Increased exception safety via the new `AutoEraseGuard`.

DoctorNoobingstoneIPresume:


lib/util_container.h
```diff
@@ -0,0 +1,118 @@
...
+
+    void swap (AutoEraseGuard &other) noexcept
+    {
+        using std::swap;
+        swap (pContainer, other.pContainer);
+        swap (iterator  , other.iterator  );
+    }
+
+    AutoEraseGuard (AutoEraseGuard &&other) noexcept
+        : pContainer (boost::exchange (other.pContainer, nullptr))
```
ChrisDodd, line 79: Is there a reason to make this class movable, rather than just leaving these deleted?  For the expected use of having a auto AutoEraseGuard on the stack for cleanup, they are not needed.

PR #3295, opened 2022-05-09
Building with `cmake -D"ENABLE_UNIFIED_COMPILATION=0"` fixed.

DoctorNoobingstoneIPresume:


mbudiu-vmw:
cpplint does not like the order of your includes. You can check with 'make cpplint'.

DoctorNoobingstoneIPresume:
> cpplint does not like the order of your includes. You can check with 'make cpplint'.

I believe the current order (`#include`-ing system-wide headers before project-specific headers) is bad:

(1) We cannot, in project-specific headers (wrapper headers), define macros which affect the behaviour/meaning of (further-included) system-wide headers (e.g. `#define NOMINMAX` before `#include <windows.h>`) because other translation units might-and-actually-currently-should include the system-wide header without/before the project-specific wrapper header.

(2) It is difficult to test whether our project-specific headers are self-sufficient (because they are always included after system-wide headers).

Moreover, the current order is inconsistent in that "x.cpp" first includes "x.h" (a project-specific header), then it includes system-wide headers and then it includes other project-specific headers.

mbudiu-vmw:
Same as #3457

DoctorNoobingstoneIPresume:
> Same as #3457

Thank you !!  >:D<

PR #3298, opened 2022-05-09
lib/crash.cpp: `addr2line`: `fcntl` calls in `pipe` initializations fixed.

DoctorNoobingstoneIPresume:


mbudiu-vmw (approved):

fruffy:
Should we merge this? What is the reason for this change?

apinski-cavium:
> Should we merge this? What is the reason for this change?
The fix looks correct to me. we using F_SETFD and oring in FD_CLOEXEC  so that should be used with F_GETFD instead of the current F_GETFL which seems wrong. FD flags vs FD status flags really.
See https://man7.org/linux/man-pages/man2/fcntl.2.html .

PR #3300, opened 2022-05-09
Fixing bug for concat elimination in SimpleSwitch

VolodymyrPeschanenkoIntel:
These small changes fix error in type checking for `concat` operation.  The idea is very simple: the type of the `mask` should be the same as for the `cast1`, because in the next line of code contains creation of `BAnd` operation with them and if the `mask` and the `cast1` will have different types then `TypeInference` generates an error for following example:
`./p4c-bm2-ss -o  ./output/strength4.json ../testdata/p4_16_samples/strength4.p4`


mbudiu-vmw (approved):

testdata/p4_16_samples/strength4-bmv2.p4
```diff
@@ -0,0 +1,74 @@
+#include <core.p4>
```
fruffy, line 1: Is this the exact same test as strength4.p4? Can we just rename that file?

VolodymyrPeschanenkoIntel: I remove it.

mbudiu-vmw (approved):

PR #3303, opened 2022-05-09
Allow shifts with constant signed positive amounts

mbudiu-vmw:

I have deleted a test that was rejected originally but now should be fine.

### DIFF
--- a/testdata/p4_16_samples_outputs/issue3287-first.p4
+++ b/testdata/p4_16_samples_outputs/issue3287-first.p4
@@ -1,4 +1,17 @@
+#include <core.p4>
+
 bit<4> func(in bit<4> l) {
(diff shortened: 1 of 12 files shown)


testdata/p4_16_samples_outputs/issue3287-first.p4
```diff
@@ -0,0 +1,4 @@
+bit<4> func(in bit<4> l) {
```
fruffy, line 1: This function is optimized away, so it is difficult to tell how it is actually transformed.

fruffy: Can we make it so this test is not optimized away?

fruffy (approved):

PR #3310, opened 2022-05-11
Do not allow structure have specialized parsers as type arguments

### LINKED ISSUE #3304: specialized parser type accepted as a type argument
Take:
```
parser p<t>(in t a);

struct s<t>
{
 t field;
}

void f()
{
 s<p<bit>> v;
}
```
If p was not specialized, we get the following error message:
```
generictype4.p4(16): [--Werror=type-error] error: s<p>: Cannot use parser p as a type parameter
```

mbudiu-vmw:
Fixes #3304


jnfoster (approved):
LGTM.

PR #3313, opened 2022-05-12
Prevent strength reduction 'a - constant' => 'a + (-constant)' for unsigned expressions

usha1830:
Strength reduction pass in frontend converts` a - constant => a + (-constant) ` irrespective of the signed-ness of the expression.
 
Example:
`a - 1 => a + (-1)`
if a is bit<2> type, the constant -1 is represented as 2w3 making the expression` a + 2w3.`

This PR fixes this issue by preventing this optimization for unsigned bit type expressions.


fruffy:
Can you show an example or test where this is a problem?

usha1830:
> Why make this change?

Without the new check added, even unsigned values are negated and the type remains unchanged (unsigned) after negating. Using either the type or value, there doesn't seem to be a way to determine whether this value is a negative or positive.
The isSigned flag is also not set while the negation is done.

Please take a look at some of the many changed output files. The value is negated but the type remains unsigned. 
Please correct me if I am wrong.

mbudiu-vmw:
Yes, you are right, but why is this a problem?

usha1830:
> Yes, you are right, but why is this a problem?

I came across this issue while fixing another issue is dpdk backend.  
Since the constants are unchanged until the end, these expressions result in an add instruction with huge values (equivalent to negative value). Since these values are valid in the range for their types, they are emitted as it is in the assembly which is wrong. 

For example: -1 represented in 2 bits is equivalent to 3 which is again a valid 2 bit unsigned number.
So instead of `var -1`, assembly is generated for `var + 3`

mbudiu-vmw:
If your assembler does not produce the right result there is a bug either in the assembler or in the backend.
Whether the user writes a+3 or a-1, these two expressions should be equivalent for a 2-bit a.
Changing the front-end will not solve this bug.

apinski-cavium:
Maybe the problem is the backend does not do "reduction" after the add.
I know when GCC is converting gimple (high level IR) to RTL (low level IR), after all adds, it does a reduction (basically introduces an and) if the bitwidth of the original type is different from the bitwidth of the mode (modes are like QI(8bit), HI(16bit), SI(32bit) and DI(64bit) like the register sizes). Maybe that is missed in the DPDK backend when dealing with non-bit width modes.

Note it would be interesting really to use C23 feature of _Bitint in the backends if the compiler supports it (https://www.open-std.org/jtc1/sc22/wg14/www/docs/n2709.pdf) so you don't need to handle this specially.

usha1830:
> If your assembler does not produce the right result there is a bug either in the assembler or in the backend. Whether the user writes a+3 or a-1, these two expressions should be equivalent for a 2-bit a. Changing the front-end will not solve this bug.

Sorry, I am still confused about this.

In my example, "a-1", a is a bit<2> type. As per the language specification, bit<w> types are always unsigned, so 2w3 for bit<2> should be a valid positive number, right? 

The dpdk assembler doesn't work on the bit types. The constants are always extended to 32-bit or 64-bit. 

I can try fixing the backend. However, the backend should sign/zero extend the values. Should the backend always sign extend the values irrespective of whether it is signed (int<w>) or unsigned (bit<w>)?

usha1830:
> Maybe the problem is the backend does not do "reduction" after the add. I know when GCC is converting gimple (high level IR) to RTL (low level IR), after all adds, it does a reduction (basically introduces an and) if the bitwidth of the original type is different from the bitwidth of the mode (modes are like QI(8bit), HI(16bit), SI(32bit) and DI(64bit) like the register sizes). Maybe that is missed in the DPDK backend when dealing with non-bit width modes.
> 
> Note it would be interesting really to use C23 feature of _Bitint in the backends if the compiler supports it (https://www.open-std.org/jtc1/sc22/wg14/www/docs/n2709.pdf) so you don't need to handle this specially.

@apinski-cavium Could you please help me with some more pointers on "reduction" in GCC?

mbudiu-vmw:
If the user declares a value as 2-bit you should only guarantee that the last 2 bits are always correct. The rest are meaningless. For unsigned values what you need to do is just apply a mask of 0x3 after each operation.

If the *architecture* declares a 2 bit value (e.g., in some metadata), you also are only promising 2 correct bits for any result that computes that metadata. The architecture should not even look at the rest of the bits.

usha1830:
> If the user declares a value as 2-bit you should only guarantee that the last 2 bits are always correct. The rest are meaningless. For unsigned values what you need to do is just apply a mask of 0x3 after each operation.
> 
> If the _architecture_ declares a 2 bit value (e.g., in some metadata), you also are only promising 2 correct bits for any result that computes that metadata. The architecture should not even look at the rest of the bits.

Ok. I will close this PR now. Will try to fix in backend/assembler.

Thanks for the clarifications!

usha1830:
Reopening this PR.

Quick summary of the issue: The StrengthReduction pass creates issues in DPDK backend in differentiating between the two nodes: "a - 1" and "a + 3" where 'a' is 2 bits wide. Both nodes are converted to "a + 2w3" by this pass. And DPDK does not support sub-byte operands so it has to promote the constant to an 8b/32b value.

I tried to fix this in DPDK backend but not able to do that. This pass converts the unsigned values to signed values but does not change the signed-ness of the node i.e. it replaces "a - 1" with "a + (-1)" but '-1' is left unsigned. This prevents the backend to correctly decide whether to zero-extend or sign-extend.

Is there a way to get the original "Sub" node when the control reaches the backend?

Also, what is the benefit of this conversion "Sub" to "Add"? Are there any targets which do not support "Sub"?

If that is the case, should we think about moving this pass to midend and letting the backends choose the optimizations based on some policy.

Thanks!

mbudiu-vmw:
As I said, the two expressions should be fully equivalent. The user could write either a-1 or a+3 and the result produced should be the same. The fact that the frontend converts one into the other should really have no impact. If you cannot handle one variant you have a bug in your backend.

usha1830:
They are equivalent as long as 'a' is bit<2> type. When 'a' is up-converted to 8-bit (as DPDK does not support sub-byte operands), they are not equivalent anymore.

mbudiu-vmw:
You are not implementing correctly the conversion to 8 bits. You should mask after each computation if you do unsigned arithmetic.

usha1830:
Strength reduction pass is retaining the type of the original operand while negating the value. 

If the operand of original SUB node is bit<w>type (bit types are unsigned as per the language spec), then the value remains unsigned even after it is negated. 

Hence, backend is not able to correctly up-convert it.

mbudiu-vmw:
Type_bits has a is Signed field which should always be preserved.

usha1830:
I still don't understand how preserving the isSigned flag even after negating an unsigned value would work. The backend will always find it as unsigned and will never sign-extend the value.

Also, can you please tell why do we have this pass in frontend? Having this pass in the midend can help the backends choose the optimizations they want to enable.

mbudiu-vmw:
I think we can close this one.

Issue #3312, opened 2022-05-12
ControlGraphs are bugged for switch(table.apply().action_run)


MichalKekely:
When switch statement based on the action that was run is used the ControlGraphs fails on the following BUG_CHECK:
https://github.com/p4lang/p4c/blob/5285887bd4f6fcf6fd40b27e58291f647583f3b1/backends/graphs/controls.cpp#L170

This is due to the new changes that now generate vertexes for Key/Actions when visiting P4Table. This means that instead of generating switch cases that are attached to the table vertex (so a singular parent) the compiler tries to generate them after all of the P4Actions (so multiple parents).

Attached is a basic P4 source code that should manifest this issue:
[graph_error.p4.txt](https://github.com/p4lang/p4c/files/8677441/graph_error.p4.txt)

PR #3327, opened 2022-05-16
Fix handling of keys without name annotation in graphs

vlstill:
Currently if a key without a `@name` annotation is generated by midend the graph backend fails on SIGSEGV. Furthermore, if multiple annotations were to appear, the graph output would be wrong. This fixes both cases, using the `expression->toString()` as fallback to `@name` annotation.


vlstill:
... that is the result of the test before the changes, now it passes of course.

MichalKekely:
Added 1 more fix to the same files - using "const bool &" in the Graph_visitor class was pointless and caused problems when creating objects

mbudiu-vmw:
The test is never invoked in the CI scripts, would you consider adding that?
(same for the previous test you added.)

mbudiu-vmw:
I can help explain how that's done in other backends.

MichalKekely:
> The test is never invoked in the CI scripts, would you consider adding that? (same for the previous test you added.)

Would renaming it to "*-bmv2" suffice? (I guess then it should be picked up by CMakeLists for bmv2)

mbudiu-vmw:
Today the tests are only run by p4test. Renaming it as -bmv2 would also compile with p4c-bm2-ss.
But what you want is to compile with p4-graphs. 
How a test is run is described in the makefile by a "driver". Usually the driver calls a python script.
There is currently no python script that runs p4-graphs.
Perhaps the simplest example I can give you is from a different project I am working on:
https://github.com/vmware/nerpa/blob/ofp4/ofp4/CMakeLists.txt
These are the relevant lines in the makefile:
```

# Program used to run tests
set(OF_DRIVER ${P4C_SOURCE_DIR}/extensions/ofp4/run-of-test.py)

# Tests to run
set (OF_TEST_SUITES "${CMAKE_CURRENT_SOURCE_DIR}/tests/*.p4")
p4c_add_tests("of" ${OF_DRIVER} "${OF_TEST_SUITES}" "${OF_XFAIL_TESTS}")
```
In your case you would replace "extensions/ofp4/run-of-test.py" with another script that you write, which would just invoke p4c-graphs on the input file and return the status of that command.

vlstill:
I'll try to enable that, how do you suggest we match the graph tests, maybe just glob for `*graph*.p4` in `testdata/p4_16_samples/` & ` testdata/p4_14_samples/`?

mbudiu-vmw:
that could be a good way to do it

vlstill:
@mbudiu-vmw, I have added a very minimalist test driver for graph (it just runs p4c-graph with whatever arguments are passed to the driver) and activated it on the test added here & in #3312. I have checked that these tests pass & that the one introduced here fails if I revert 0d38b42eece1c32428876326ac915803236da79a.

mbudiu-vmw:
I assume you have run `make check-graph`.

mbudiu-vmw (approved):

vlstill:
I did run ` ctest -V -R graph` but I have just checked `make check-graph` runs the same tests and works the same.

vlstill:
@mbudiu-vmw, I am kind of clueless of what happened in the CI. Do you have any idea what could have caused the failure on `dpdk/testdata/p4_16_samples/pna-example-tcp-connection-tracking.p4` which was in no way changed and neither is it picked by the new graph target... Is there any other source the test could have came from?

mbudiu-vmw:
I think we have the same failure in the CI without this PR.
Can you please fix this by running the respective test with the -f flag and submit the modified files?
This is probably because we merged some PRs that weren't rebased on main.

vlstill:
OK, I have rebased & regenerated the p4.spec file

PR #3330, opened 2022-05-16
Check that parameter default values are compile-time known

### LINKED ISSUE #3316: default argument constraints are not checked except for wrong type when calling
Take:
```
bit h(in bit a)
{
  return a;
}
bit f(in bit a = h(1), in bit b)
{
  return b;
}
bit g(in bit a)
{
  return f(b = a);
}
```

The above is accepted even though the specifications has the following:
```
- Default parameter values are only allowed for 'in' or direction-less parameters;
  these values must evaluate to compile-time constants.
```
h(1) is not a compile time constant as far as I can tell.

Note the other thing is if you replace h(1) with 2w1, the code is rejected but only at the point where the call is made rather than the definition point.

mbudiu-vmw:
Fixes #3316

### DIFF
--- a/frontends/p4/typeChecking/typeChecker.cpp
+++ b/frontends/p4/typeChecking/typeChecker.cpp
@@ -1623,6 +1623,13 @@ const IR::Node* TypeInference::postorder(IR::Parameter* param) {
             return param;
         }
     }
+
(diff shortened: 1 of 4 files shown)


jnfoster (approved):
LGTM.

PR #3332, opened 2022-05-16
Add generic structs to P4Runtime digest example

antoninbas:
To show how they are represented in the P4Info.

### DIFF
--- a/testdata/p4_16_samples_outputs/p4rt_digest_complex-first.p4
+++ b/testdata/p4_16_samples_outputs/p4rt_digest_complex-first.p4
@@ -9,6 +9,15 @@ struct s_t {
     bit<16> f16;
 }
 
+struct s2_t<T> {
+    bit<16> f16;
+    T       f;
+}
+
+@name("s3_t") struct s3_t<T> {
+    T f;
+}
+
 header h_t {
     s_t     s;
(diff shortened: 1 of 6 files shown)


testdata/p4_16_samples_outputs/p4rt_digest_complex-first.p4
```diff
@@ -42,15 +51,26 @@ control MyEC(inout EMPTY a, inout EMPTY b, in psa_egress_input_metadata_t c, ino
     }
 }
 
+struct s2_t_0 {
+    bit<16> f16;
+    bit<32> f;
+}
+
+@name("s3_t") struct s3_t_0 {
```
antoninbas, line 59: @mbudiu-vmw I don't know if that was discussed at the time of adding generic structs, but this is a bit problematic. It's easy to shoot yourself in the foot as all generated structs will have the same annotation, and the P4Info will actually remove all of the conflicting definitions save for one (so the P4Info will be incorrect).

mbudiu-vmw: no, the interaction between generic structs and the control plane was never discussed.

mbudiu-vmw (approved):

PR #3337, opened 2022-05-18
Use psabpf-ctl register command in PTF tests

kmateuszssak:
- [X]  Add `register_get` and `register_set` commands to PTF tests.
- [X]  Use `key` param name in singular form (key is a list of fields).


mbudiu-vmw (approved):

kmateuszssak:
@mbudiu-vmw Yes, they look weird but we don't have a mechanism that allow us to share `const` values between Python PTF tests and P4 test programs. I added a comment. Maybe it will help a bit.

PR #3339, opened 2022-05-19
Require a start state in parsers

### LINKED ISSUE #3338: parser with one state but not named start is not rejected
Take:
```
parser MyParser1(){
    state not_start {
        transition accept;
    }
}
```
I would have assumed this would have been rejected based on the following in the specifications:
```
At least one state, named `start`, must be present in any `parser`.
```
But currently it is accepted without even a warning. The rest of the paragraph is implemented though, that is there is a check for duplicated states and a check for accept and reject states.

mbudiu-vmw:
Fixes #3338


fruffy (approved):

jnfoster (commented):
Nice.

jnfoster (approved):

PR #3341, opened 2022-05-19
Allow unification between serenum and underlying type

### LINKED ISSUE #3333: "Compiler Bug: At this point in the compilation typechecking should not infer new types anymore, but it did." with select statement and enum types
While trying to understand the type rules for select since I would have expected serializable enums to "decay" (implict cast) into their underlaying type (but it does not and it is partly inconsistent with the rest of the specifications really) I found this issue:
Take:
```
enum bit<8> myenum {  value = 0 }
parser MyParser1(in  myenum a) {
    state start {
        transition select(a) {
                1 .. 22: state1;
                _: accept;
        }
    }
    state state1 {
        transition accept;
    }
}
```

With p4c we get:
```
t.p4(8): [--Werror=type-error] error: SelectExpression
        transition select(a) {
                   ^
  ---- Actual error:
  t.p4(1): Cannot unify type 'bit<8>' with type 'myenum'
  enum bit<8> myenum
              ^^^^^^
  ---- Originating from:
  t.p4(1): 'match' case label type 'bit<8>' does not match expected type 'myenum'
  enum bit<8> myenum
              ^^^^^^
In file: /home/apinski/src/p4/p4c/frontends/p4/typeChecking/typeChecker.cpp:161
```

Note the error message also does not point to which case lable either, it just points to the enum type but that is a different issue.

mbudiu-vmw:
Fixes #3333

### DIFF
--- a/testdata/p4_16_samples_outputs/issue3333.p4
+++ b/testdata/p4_16_samples_outputs/issue3333.p4
@@ -1,0 +1,16 @@
+enum bit<8> myenum {
+    value = 0
+}
+
+parser MyParser1(in myenum a) {
+    state start {
+        transition select(a) {
+            1 .. 22: state1;
+            default: accept;
+        }
+    }
+    state state1 {
+        transition accept;
+    }
+}
+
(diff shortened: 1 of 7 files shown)


mbudiu-vmw:
This PR seems to have broken something. I will check.

testdata/p4_16_samples_outputs/issue3333.p4
```diff
@@ -0,0 +1,16 @@
+enum bit<8> myenum {
```
fruffy, line 1: I think a problem with these tests is that they are all optimized away in the front end. So you can not actually see how these programs are transformed or how the mid end passes handle this. Even though most of these changes just check for errors.

fruffy (approved):

PR #3348, opened 2022-05-23
Fixed start state renaming for p4-14 translation using @packet_entry

MichalKekely:
The issue was that if there was a transition into the start state (for example from one of the @packet_entry states) and @packet_entry used, the compiler renamed the original start state to replace it with a new one. However these transitions into the original start state were not renamed.

### DIFF
--- a/frontends/p4/fromv1.0/converters.h
+++ b/frontends/p4/fromv1.0/converters.h
@@ -769,7 +769,7 @@ class InsertCompilerGeneratedStartState: public Transform {
         if (path->name.name != IR::ParserState::start)
             return path;
         // Just to make sure we can also check it explicitly
-        auto pe = getContext()->node->to<IR::PathExpression>();;
+        auto pe = getContext()->node->to<IR::PathExpression>();
         auto sc = findContext<IR::SelectCase>();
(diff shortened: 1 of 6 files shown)


frontends/p4/fromv1.0/converters.h
```diff
@@ -759,6 +759,18 @@ class InsertCompilerGeneratedStartState: public Transform {
         return state;
     }
 
+    // Rename any path refering to original start state
+    const IR::Node* postorder(IR::Path* path) override {
```
mbudiu-vmw, line 763: won't this also rename a variable reference if the variable is named "start"?
To be safe this should check the parent context to be a SelectCase, and moreover this to be the child #1.

mbudiu-vmw (changes requested):

frontends/p4/fromv1.0/converters.h
MichalKekely: Updated to explicitly check if the path that is being changed is either within:
`ParserState->selectExpression<PathExpression>->path`
Or:
`SelectCase->state<PathExpression>->path`

PR #3365, opened 2022-05-26
accept PSA_MeterColor_t in action parameter

kamleshbhalui:


control-plane/bfruntime.cpp
```diff
@@ -601,9 +601,15 @@ BFRuntimeGenerator::makeActionSpecs(const p4configv1::Table& table,
         for (const auto& param : action->params()) {
             auto* annotations = transformAnnotations(
                 param.annotations().begin(), param.annotations().end());
-            addActionDataField(
-                dataJson, param.id(), param.name(), true /* mandatory */,
-                false /* read_only */, makeTypeBytes(param.bitwidth()), annotations);
+            if (param.type_name().name() == "PSA_MeterColor_t") {
```
mbudiu-vmw, line 604: it may be safer to check the type of the parameter as obtained from the typeMap (the canonical type).
This may not work if the type is a typedef (unless the midend actually eliminates typedefs).
You should probably add a test for this case.

mbudiu-vmw (approved):

control-plane/bfruntime.cpp
kamleshbhalui: bfrt generation happens before midend.

control-plane/p4RuntimeSerializer.cpp
```diff
@@ -973,20 +973,24 @@ class P4RuntimeAnalyzer {
             auto paramType = typeMap->getType(actionParam, true);
             if (!paramType->is<IR::Type_Bits>() && !paramType->is<IR::Type_Boolean>()
                 && !paramType->is<IR::Type_Newtype>() &&
-                !paramType->is<IR::Type_SerEnum>()) {
+                !paramType->is<IR::Type_SerEnum>() && !paramType->is<IR::Type_Enum>()) {
```
kamleshbhalui, line 976: type already checked while generating p4info here.

kamleshbhalui:
@mbudiu-vmw please review it.

mbudiu-vmw (approved):

PR #3366, opened 2022-05-27
String utility functions

mbudiu-vmw:

### DIFF
--- a/ir/node.cpp
+++ b/ir/node.cpp
@@ -18,11 +18,20 @@ limitations under the License.
 #include "ir/json_loader.h"
 
 #include "node.h"
+// #include <signal.h>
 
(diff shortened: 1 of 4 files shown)


ir/node.cpp
```diff
@@ -18,11 +18,20 @@ limitations under the License.
 #include "ir/json_loader.h"
 
 #include "node.h"
+// #include <signal.h>
```
fruffy, line 21: Left a commented include here.

fruffy (approved):

ir/node.cpp
mbudiu-vmw: yes, this is paired with the commented code a bit below too

fruffy: Ah I see, a comment about this could be useful.

PR #3369, opened 2022-05-27
Dpdk backend: Minor changes in context json and learner table timeout values

usha1830:
- target_action_name field in context json is renamed to target_name and is emitted for all table types now. Earlier it was only emiited for match tables (tables  from P4 program), now it is emiited for compiler generated tables for action profile and action selectors.
- Dpdk target has improved its API to provide more timeout ids which can be configures by control plane. The initial value for each timeout id is 120s.


mbudiu-vmw (approved):

usha1830:
@mbudiu-vmw Thanks for the review. Please merge.

usha1830:
I deleted the branch by mistake in an attempt to sync with upstream.

Created a new PR with same set of changes and a minor change in printing numbers 
https://github.com/p4lang/p4c/pull/3372

PR #3371, opened 2022-05-30
Dpdk Backend: Support dynamic array indexing for header stack elements

usha1830:
This PR is to support dynamic array indexing for accessing the elements of a header stack using subscript notation.
The array index expression within the following statements/expressions are supported

- Assignment statements (on both sides as well)
- Table keys
- If condition expression
- Switch expression
- Function param
- Binary operation
- Unary Operation
- Slice

`meta.left = hdrs.udp[meta.depth - 1].sport;`

should be translated as 

```
tmp = meta.depth - 1
if (tmp == 0)
    meta.left = hdrs.udp[0].sport;
if (tmp == 1)
    meta.left = hdrs.udp[1].sport;
```

If the array subscript is present on both sides with different expression in the [], it will be translated into multi level branches like below:

`hdrs.udp[meta.depth].sport = hdrs.udp[meta.depth - 1].sport;`  =>

```
tmp = meta.depth - 1
tmp1 = meta.depth

if (tmp1 == 0)
    if (tmp == 0)
        hdr.udp[0].sport = hdr.udp[0].sport
if (tmp == 1)
        hdr.udp[0].sport = hdr.udp[1].sport
if (tmp1 == 1)
    if (tmp == 0)
        hdr.udp[1].sport = hdr.udp[0].sport
if (tmp == 1)
        hdr.udp[1].sport = hdr.udp[1].sport		
```

### DIFF
--- a/backends/dpdk/dpdkArch.cpp
+++ b/backends/dpdk/dpdkArch.cpp
@@ -880,7 +880,13 @@ bool ExpressionUnroll::preorder(const IR::MethodCallExpression *m) {
     return false;
 }
 
-bool ExpressionUnroll::preorder(const IR::Member *) {
+bool ExpressionUnroll::preorder(const IR::Member *m) {
+    if (m->expr->is<IR::ArrayIndex>()) {
+        root = new IR::PathExpression(IR::ID(refMap->newName("tmp")));
+        decl.push_back(new IR::Declaration_Variable(root->path->name, m->type));
+        stmt.push_back(new IR::AssignmentStatement(root, m));
+        return false;
+    }
     root = nullptr;
     return false;
 }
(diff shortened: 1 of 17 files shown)


backends/dpdk/dpdkArch.cpp
```diff
@@ -1397,15 +1410,177 @@ const IR::Node* DismantleMuxExpressions::preorder(IR::Mux* expression) {
...
+            if (!ai->right->is<IR::Constant>())
+                return true;
+        }
+    }
+    return false;
+}
+
+// This function drives the replacement of statements with variable array index expressions
+// with a series of if statements based on the number of elements in the header stack.
+void SplitHSIndexExpression::replaceVarIndexWithIf(IR::AssignmentStatement *statement) {
```
mbudiu-vmw, line 1434: Did you check whether we don't already have a midend pass for this? I recall reviewing one.

usha1830: Oh! I do see a pass in midend,  _HSIndexTransform_
It is not used in dpdk backend. I will try invoking that and see it can be used here.
Thanks for the pointer!!

usha1830:
I have created another PR https://github.com/p4lang/p4c/pull/3375 to use the midend pass. Closing this.

PR #3372, opened 2022-05-30
Dpdk backend: Minor changes in context json and learner table timeout values. Supercedes PR #3369

usha1830:
This PR supersedes #3369 

> target_action_name field in context json is renamed to target_name and is emitted for all table types now. Earlier it was only emiited for match tables (tables from P4 program), now it is emiited for compiler generated tables for action profile and action selectors.
> Dpdk target has improved its API to provide more timeout ids which can be configures by control plane. The initial value for each timeout id is 120s.

Additional change in the print format for learner table timeout values.


mbudiu-vmw (approved):

usha1830:
@mbudiu-vmw Thanks for the review. Could you please merge this?

PR #3377, opened 2022-06-02
Fixed comparison between two Type_Bits

MichalKekely:
Due to the changes where new Type_Bits are created the "==" comparison no longer works, equiv is used instead.

### DIFF
--- a/frontends/p4/typeChecking/typeChecker.cpp
+++ b/frontends/p4/typeChecking/typeChecker.cpp
@@ -1837,7 +1837,8 @@ const IR::Node* TypeInference::postorder(IR::Operation_Relation* expression) {
         expression->left = c.left;
         expression->right = c.right;
     } else {
-        if (!ltype->is<IR::Type_Bits>() || !rtype->is<IR::Type_Bits>() || !(ltype == rtype)) {
+        if (!ltype->is<IR::Type_Bits>() || !rtype->is<IR::Type_Bits>() || !(ltype->equiv(*rtype))) {
+            std::cerr << "Here" << std::endl;
             typeError("%1%: not defined on %2% and %3%",
                       expression, ltype->toString(), rtype->toString());
             return expression;


frontends/p4/typeChecking/typeChecker.cpp
```diff
@@ -1837,7 +1837,8 @@ const IR::Node* TypeInference::postorder(IR::Operation_Relation* expression) {
         expression->left = c.left;
         expression->right = c.right;
     } else {
-        if (!ltype->is<IR::Type_Bits>() || !rtype->is<IR::Type_Bits>() || !(ltype == rtype)) {
+        if (!ltype->is<IR::Type_Bits>() || !rtype->is<IR::Type_Bits>() || !(ltype->equiv(*rtype))) {
+            std::cerr << "Here" << std::endl;
```
mbudiu-vmw, line 1841: this should not be here

MichalKekely: Yes, removed :-)

mbudiu-vmw (approved):

mbudiu-vmw:
If you can add a test that fails without this it's also useful.

MichalKekely:
> If you can add a test that fails without this it's also useful.

Just working on it

MichalKekely:
> If you can add a test that fails without this it's also useful.

Well, this seems to be quite specific issue with a non-open source midend. The issue is created by not properly clearing the `TypeMap` before `TypeChecking` after some pass did transformations that reintroduced the same `Type_Bits`, however with different id.

It seems to be fixable in the specific midend also by properly clearing the `TypeMap`, but I would still consider going forward with this change as well. Let's assume we have these two types:
`[1502] Type_Bits size=16 isSigned=0`
`[6970] Type_Bits size=16 isSigned=0`
It would make sense that these two should be comparable (since both of these `Type_Bits` objects clearly represent the same type). However `==` returns `false` for these (and therefore the original if statement is triggered), while `equiv` returns `true` (and therefore the changed if statement). And it also does not make much sense when the actual error message looks like:
`[--Werror=type-error] error: >=: not defined on bit<16> and bit<16>`

mbudiu-vmw:
The original design for the typeMap called for using "canonical type representations" which could be compared for equality. There is a bunch of code in the compiler trying to do that.  However, it's not clear that "canonical" is well-defined when dealing with generics, so equiv() or unification are the right way to check types for "equality". The spec does not really say when two types are equal either.

PR #3381, opened 2022-06-03
Bring commonly used builtin options under lib directory

kamleshbhalui:


lib/options.cpp
```diff
@@ -47,6 +47,18 @@ std::vector<const char*>* Util::Options::process(int argc,
     if (argc == 0 || argv == nullptr)
         throw std::logic_error("No arguments to process");
     binaryName = argv[0];
+    // collect command line args
+    for (int i = 0; i < argc; i++) {
+        compileCommand += argv[i];
+        compileCommand += " ";
+    }
+    compileCommand = compileCommand.replace("(from pragmas)", "");
```
mbudiu-vmw, line 55: Where is this coming from? Why does it apply to all backends?

kamleshbhalui: it is required here (dpdk backend) and other backend adopting from base code would need
and this information is available(first hand here).

mbudiu-vmw: Where is the string "from pragmas" coming from?

kamleshbhalui: frontends/common/applyOptionsPragmas.cpp:30

mbudiu-vmw: Wouldn't it be cleaner to just skip "(from pragmas)" only if it is argv[0]?

kamleshbhalui: yes skipped it.

mbudiu-vmw (approved):

PR #3385, opened 2022-06-07
DPDK Backend: Fix action name used in learn instruction, minor fix in context json

usha1830:
This PR fixes two issues:

1) Learn instruction corresponding to add_entry extern takes action name from user as input.
However, compiler modifies the action name for creating unique names. Since the names are different, dpdk pipeline fails to load.

Added a mapping between user visible name and compilers internal name used in spec file to emit the right compiler generated action name.

```
extern bool add_entry<T>(string action_name,
                         in T action_params,
                         in ExpireTimeProfileId_t expire_time_profile_id);
```
2) "name" field for actions in context json should match with action name in bfrt json. bfrt json uses controlPlaneName(). Updated the "name" field to use controlPlaneName.

### DIFF
--- a/backends/dpdk/dpdkArch.cpp
+++ b/backends/dpdk/dpdkArch.cpp
@@ -2249,6 +2249,13 @@ void CollectAddOnMissTable::postorder(const IR::P4Table* t) {
     if (use_add_on_miss) {
         for (auto action : t->getActionList()->actionList) {
             auto action_decl = refMap->getDeclaration(action->getPath())->to<IR::P4Action>();
+            // Map the compiler generated internal name of action (emitted in .spec file) with
+            // user visible name in P4 program.
+            // To get the user visible name, strip any prefixes from externalName.
+            cstring userVisibleName = action_decl->externalName();
+            userVisibleName = userVisibleName.findlast('.');
+            userVisibleName = userVisibleName.trim(".\t\n\r");
+            structure->learner_action_map.emplace(userVisibleName, action_decl->name.name);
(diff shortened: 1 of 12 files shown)


backends/dpdk/dpdkArch.cpp
```diff
@@ -2249,6 +2249,13 @@ void CollectAddOnMissTable::postorder(const IR::P4Table* t) {
     if (use_add_on_miss) {
         for (auto action : t->getActionList()->actionList) {
             auto action_decl = refMap->getDeclaration(action->getPath())->to<IR::P4Action>();
+            // Map the compiler generated internal name of action (emitted in .spec file) with
+            // user visible name in P4 program.
+            // To get the user visible name, strip any prefixes from externalName.
+            cstring userVisibleName = action_decl->externalName();
+            userVisibleName = userVisibleName.findlast('.');
+            userVisibleName = userVisibleName.trim(".\t\n\r");
```
mbudiu-vmw, line 2257: do we really expect spaces in this name?

mbudiu-vmw (approved):

backends/dpdk/dpdkArch.cpp
usha1830: Not really! I can keep just '.' in this list of characters to be trimmed.
I will take care of this along with other changes in a future PR.

usha1830: Not really! I can keep just '.' in this list of characters to be trimmed.
I will take care of this along with other changes in a future PR.

PR #3389, opened 2022-06-09
BUILD_USE_COLOR=OFF disables colors for the C++ compiler

davidbolvansky:
BUILD_USE_COLOR=OFF currently does not do what I would expect and what I need - it fails to disable colors (ansi control characters) - note, gcc default mode is "auto", not "never".

Either we should proceed with a patch like this one, or rename the original option to something like "BUILD_FORCE_COLORS". I prefer the former as I have an use case for BUILD_USE_COLOR=OFF.


mbudiu-vmw (approved):

davidbolvansky:
CLA signed. ready to be merged.

@mbudiu-vmw

PR #3393, opened 2022-06-13
Make bitvec helpers available as global, don't polute top-level namespace

vlstill:
- remove code duplication
- `count_{leading,trailing}_zeroes` + `popcount` are now available globally to use on built-in integral types
- builtin_* helpers + the aforementioned new functions are defined in `namespace bv` to avoid possible collisions

### DIFF
--- a/lib/bitvec.h
+++ b/lib/bitvec.h
@@ -60,6 +60,7 @@ static inline int builtin_popcount(unsigned long long x) { return __builtin_popc
 template<typename I>
 int count_trailing_zeroes(I x) {
     static_assert(std::is_integral<I>::value, "'I' has to be integral type");
+    assert(x != 0);
 #if defined(__GNUC__) || defined(__clang__)
     return bv::builtin_ctz(x);
 #else
@@ -73,6 +74,7 @@ int count_trailing_zeroes(I x) {
 template<typename I>
 int count_leading_zeroes(I x) {
     static_assert(std::is_integral<I>::value, "'I' has to be integral type");
+    assert(x != 0);
 #if defined(__GNUC__) || defined(__clang__)
     return bv::builtin_clz(x);
 #else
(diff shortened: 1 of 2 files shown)


lib/bitvec.h
```diff
@@ -55,6 +57,48 @@ static inline int builtin_popcount(unsigned long x) { return __builtin_popcountl
 static inline int builtin_popcount(unsigned long long x) { return __builtin_popcountll(x); }
 #endif
 
+template<typename I>
+int count_trailing_zeroes(I x) {
```
davidbolvansky, line 61: Any protection against zero?

lib/bitvec.h
```diff
@@ -55,6 +57,48 @@ static inline int builtin_popcount(unsigned long x) { return __builtin_popcountl
...
+    static_assert(std::is_integral<I>::value, "'I' has to be integral type");
```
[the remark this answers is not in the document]
vlstill: Added an `assert`. I did not want to add BUG_CHECK for such a low-level functionality & there are already asserts in bitvec.h.

lib/bitvec.h
vlstill: Added an `assert`. I did not want to add BUG_CHECK for such a low-level functionality & there are already asserts in bitvec.h.

vlstill:
@davidbolvansky, it appears I have accidentally removed your comment asking why I have used template. Sorry about that, I wanted to delete my misplaced comment and apparently remove the thread (why am I even allowed to do this I don't understand).

The reply: Template to avoid duplicated code... I could probably restrict it more (i.e. unsigned & <= 64 bits) to ensure the proper builtin wrappers should be available. To be even more safe we could do something like `template<typename I> builtin_ctz(I) = delete` to avoid any implicit casts (as *it appears* that things like `uint32_t` or `uintptr_t` have to be typedefs so it should not produce false errors).

lib/bitvec.h
davidbolvansky: Thanks

vlstill:
@davidbolvansky, I did make the static assert a bit more comprehensive.

mbudiu-vmw (approved):

PR #3397, opened 2022-06-16
Update BUILD.bazel to remove constat patterns from glob

lzhzero:
Constant patterns in glob can be error-prone,
moving the files outside the globs.

### DIFF
--- a/BUILD.bazel
+++ b/BUILD.bazel
@@ -306,10 +306,21 @@ cc_proto_library(
 cc_library(
     name = "p4c_dpdk_lib",
     srcs = glob(
-        ["backends/dpdk/*.cpp", "backends/dpdk/control-plane/*.cpp", "backends/bmv2/common/lower.cpp"],
-        exclude = ["backends/dpdk/main.cpp", "backends/dpdk/spec.cpp", "backends/dpdk/printUtils.cpp", "backends/dpdk/dbprint-dpdk.cpp"],
-    ),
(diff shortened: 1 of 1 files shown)


BUILD.bazel
```diff
@@ -306,10 +306,21 @@ cc_proto_library(
...
     srcs = glob(
-        ["backends/dpdk/*.cpp", "backends/dpdk/control-plane/*.cpp", "backends/bmv2/common/lower.cpp"],
-        exclude = ["backends/dpdk/main.cpp", "backends/dpdk/spec.cpp", "backends/dpdk/printUtils.cpp", "backends/dpdk/dbprint-dpdk.cpp"],
-    ),
-    hdrs = glob(["backends/dpdk/*.h", "backends/dpdk/control-plane/*.h", "backends/bmv2/common/lower.h"]),
+        [
+            "backends/dpdk/*.cpp",
+            "backends/dpdk/control-plane/*.cpp",
+        ],
+        exclude = [
```
mbudiu-vmw, line 313: so the patterns in exclude, which are constant, but inside glob, are not a problem?

mbudiu-vmw (approved):

BUILD.bazel
lzhzero: The issue with the default include=[] is that constant patterns such  as the `backends/bmv2/common/lower.cpp`, will be silently ignored if the file is not available.

PR #3398, opened 2022-06-20
Improve the way we wait for commands in driver

vlstill:
- avoid leaving orphaned command subprocesses
- don't let CTRL+C kill the driver, forward it instead

### DIFF
--- a/tools/driver/p4c_src/driver.py
+++ b/tools/driver/p4c_src/driver.py
@@ -12,8 +12,11 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 import os
-import shlex, subprocess
+import signal
+import shlex
+import subprocess
 import sys
+import traceback
 
 import p4c_src.util as util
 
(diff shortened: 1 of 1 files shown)


tools/driver/p4c_src/driver.py
```diff
@@ -251,7 +251,32 @@ def runCmd(self, step, cmd):
...
+        try:
+            while True:
+                try:
+                    p.communicate()
+                    break  # done waiting, process ended
+                except KeyboardInterrupt:
+                    import signal
+                    p.send_signal(signal.SIGINT)
+        except:
+            import traceback
```
mbudiu-vmw, line 270: are local imports considered good style?

mbudiu-vmw (approved):

tools/driver/p4c_src/driver.py
vlstill: On the other hand, I don't think there are downsides to moving them to the top (maybe the start could be a tiny bit slower, but likely these modules are loaded somewhere in background anyway). So I moved them to the top.

PR #3399, opened 2022-06-20
Fix incorrect copy propagation into table match keys

vhavel:
If table is applied at multiple places, local copy propagation must ensure expressions are propagated only if all control flow paths to all apply method calls assign the same expression to a variable. Currently, LocalCopyPropagation correctly inhibits copy propagation into match key `k` if two different paths to apply() call assign a different (not equivalent) expressions into `k`. But if one path does not assign any value to `k`, and all other assign `k=a`, e.g.:

```
if (cond) {
    k = a;
    t.apply();
} else {
    t.apply();
}
```

the current implementation incorrectly propagates `a` into the table's match key (rewriting `k` to `a`).

This PR fixes this problem by forbidding copy propagation of a match key in presence of a path without any remap. Before this change, the testcase `testdata/p4_16_samples/copyprop1.p4` fails because `copyprop1-midend.p4` incorrectly matches on `hdr.payload.y` instead of `x`.

### DIFF
--- a/midend/local_copyprop.cpp
+++ b/midend/local_copyprop.cpp
@@ -585,8 +585,10 @@ void DoLocalCopyPropagation::apply_function(DoLocalCopyPropagation::FuncInfo *ac
 
 void DoLocalCopyPropagation::apply_table(DoLocalCopyPropagation::TableInfo *tbl) {
     ++tbl->apply_count;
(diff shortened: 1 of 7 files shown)


mbudiu-vmw:
@ChrisDodd wrote this code, so he should review this PR.

PR #3400, opened 2022-06-20
Use width_bits() instead of size for constantFolding

fruffy:
`width_bits` fulfills the same functionality but can be overridden. This makes the pass a bit more flexible.


mbudiu-vmw (approved):
This will only help if you subclass Type_Bits.

fruffy:
Yes, part of the reason this was added!

PR #3401, opened 2022-06-22
Update rules that require a cpp toolchain to use a helper function th…

katre:
…at defines the toolchain type.


smolkaj:
Thanks, LGTM.

PR #3403, opened 2022-06-24
Fix getting element type from HeaderUnion

OleksandrLetychevskyiIntel:
Helps solve the error : `Could not find type for <Member>(584464) .end`
For example test `issue561-bmv2.p4` with `--loopsUnroll` flag

### DIFF
--- a/midend/parserUnroll.cpp
+++ b/midend/parserUnroll.cpp
@@ -179,7 +179,9 @@ class ParserStateRewriter : public Transform {
     IR::Node* postorder(IR::Member* expression) {
         if (!afterExec)
             return expression;
-        auto basetype = getTypeArray(expression->expr);
+        const IR::Type* basetype = getTypeArray(expression->expr);
+        if (!basetype)
+            basetype = expression->expr->type;
         if (basetype->is<IR::Type_Stack>()) {
             auto l = afterExec->get(expression->expr);
             BUG_CHECK(l->is<SymbolicArray>(), "%1%: expected an array", l);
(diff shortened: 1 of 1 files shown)


mbudiu-vmw (commented):
I have kindof lost context for this change, is expression->type properly set at this point?

OleksandrLetychevskyiIntel:
Since there is a H`eaderUnion` then `typeMap->getType(element, true);` for some reason does not find the type

mbudiu-vmw:
You should understand why that doesn't work and fix that problem.

OleksandrLetychevskyiIntel:
If I understand correctly, then the type is not set somewhere here
https://github.com/vladyslav-dubina/p4c/blob/a992d35292b5c65c20f9b0807e6371c260869290/frontends/p4/evaluator/evaluator.cpp#L225
This is due to the lack of `HeaderUnion ` support, the problem can be solved by introducing additional checks inside this pass, or we can use what is already presented in this PR

mbudiu-vmw:
The Evaluator pass does not insert types in the typemap.
The only pass which inserts types in the typeMap is typeInference.
It is also the only pass which sets the ->type field of an expression, if you invoke it with a 'true' flag.
So if you have an expression without a type in the typeMap, it was created after your last call to typeInference.
In general, the ->type field of an expression could be stale if you haven't updated it recently.
The way to diagnose this is to find the IR node which does not have a know type, read its id using a debugger, then put a breakpoint in the traceCreation function in node.cpp to see when the node is created. Then you can see why typeInference isn't run between the creation moment and the moment you try to get the type.

OleksandrLetychevskyiIntel:
My statement yesterday was a little wrong, because the desired member appears in the program after
https://github.com/vladyslav-dubina/p4c/blob/a992d35292b5c65c20f9b0807e6371c260869290/frontends/p4/inlining.cpp

mbudiu-vmw:
ResetHeaders is called early in the frontend, so typeInference is called many times between this point and unrolling.
Don't forget that a Transform or Modifier actually makes a clone of an IR::Node, so you have to use getOriginal() to look up something in a typeMap (this applies only to the currently visited node, not to its children).

mbudiu-vmw:
I have submitted #3420 which fixes this problem, but I haven't audited the entire code to see whether there aren't other similar problems elsewhere. I will close this PR, please approve the other one if it fixes the problem you are reporting. 
BTW: you should always file an issue first for such problems.

PR #3406, opened 2022-06-27
Implement Redundant Copy Elimination

kamleshbhalui:
algorithm:

1)collect all used, copied and defined variables in all statements in a block
2)iterate over list of statements in a block in reverse order
3)if a statement is mov/cast, check if destination variable is a single definition, has single use, and also is a single copy
4) then drop that statement and update all use point of destintation with source of mov/cast operation.
else copy statement and push to the to the list.
5)create the list in original order and return.

### DIFF
--- a/backends/dpdk/dpdkAsmOpt.cpp
+++ b/backends/dpdk/dpdkAsmOpt.cpp
@@ -264,5 +264,124 @@ bool ValidateTableKeys::preorder(const IR::DpdkAsmProgram *p) {
     return false;
 }
 
+const IR::Expression* CopyPropagationAndElimination::getIrreplaceableExpr(cstring str,
+                                                                        bool allowConst) {
+    if (collectUseDef->dontEliminate.count(str))
+        return nullptr;
+    if (!str.startsWith("m."))
+        return nullptr;
+    auto expr = collectUseDef->replacementMap[str];
+    const IR::Expression* prev = nullptr;
+    cstring exprStr;
+    if (expr) exprStr = expr->toString();
+    while (expr != nullptr
+        && (!allowConst ? expr->is<IR::Member>() : true)
+        && str != exprStr
+        && (!allowConst ? exprStr.startsWith("m.") : true)
+        && collectUseDef->dontEliminate.count(exprStr) == 0
+        && newUsesInfo[exprStr] == 0
+        && (expr->is<IR::Member>() ? collectUseDef->haveSingleUseDef(exprStr) : true)) {
+        prev = expr;
+        expr = collectUseDef->replacementMap[exprStr];
+        if (expr) exprStr = expr->toString();
+    }
+    if (expr != nullptr &&  allowConst)
+        return expr;
+    else if (expr != nullptr && expr->is<IR::Member>())
+        return expr;
+    else
+        return prev;
+}
+
+const IR::Expression* CopyPropagationAndElimination::replaceIfCopy(const IR::Expression *expr,
+                                                                    bool allowConst) {
+    if (!expr)
+        return expr;
+    auto str = expr->toString();
+    if (collectUseDef->haveSingleUseDef(str)) {
+        if (auto rexpr = getIrreplaceableExpr(str, allowConst)) {
+            return rexpr;
+        }
+    }
+     if (!expr->is<IR::Constant>())
+        newUsesInfo[expr->toString()]++;
+    return expr;
+}
+
+const IR::DpdkAsmStatement* CopyPropagationAndElimination::elimCastOrMov(
+    const IR::DpdkAsmStatement* stmt) {
+    const IR::Expression* srcExpr = nullptr, *dstExpr = nullptr;
+    if (stmt->is<IR::DpdkMovStatement>()) {
+        auto mv = stmt->to<IR::DpdkMovStatement>();
+        srcExpr = mv->src;
+        dstExpr = mv->dst;
+    } else {
+        auto mv = stmt->to<IR::DpdkCastStatement>();
+        srcExpr = mv->src;
+        dstExpr = mv->dst;
+    }
+    auto src = srcExpr->toString();
+    auto dst = dstExpr->toString();
+    bool dropCopy = false;
+    if (collectUseDef->dontEliminate.count(dst) == 0
+            && dst.startsWith("m.")
+            && newUsesInfo[dst] == 0
+            && collectUseDef->haveSingleUseDef(dst)) {
+        dropCopy = true;  // do nothing, drop the statement
+    }
+    if (!dropCopy) {
+        auto r = replaceIfCopy(srcExpr);
+        if (dst != r->toString() && dst.startsWith("m.")) {
+            newUsesInfo[r->toString()]++;
+            return new IR::DpdkMovStatement(dstExpr, r);
+        } else {
+            newUsesInfo[src]++;
+            return stmt;
+        }
+    } else {
+        return nullptr;
+    }
+}
+
+IR::IndexedVector<IR::DpdkAsmStatement>
+CopyPropagationAndElimination::copyPropAndDeadCodeElim(
+    IR::IndexedVector<IR::DpdkAsmStatement> stmts) {
+    IR::IndexedVector<IR::DpdkAsmStatement> instr;
+    for (auto stmt1 = stmts.rbegin(); stmt1 != stmts.rend(); stmt1++) {
+        auto stmt = *stmt1;
+        if (stmt->is<IR::DpdkMovStatement>() || stmt->is<IR::DpdkCastStatement>()) {
+            if (auto newMv = elimCastOrMov(stmt))
+                instr.push_back(newMv);
+        } else if (auto jc = stmt->to<IR::DpdkJmpLessStatement>()) {
+            instr.push_back(new IR::DpdkJmpLessStatement(jc->label,
+                        replaceIfCopy(jc->src1, false), replaceIfCopy(jc->src2)));
+        } else if (auto jc = stmt->to<IR::DpdkJmpGreaterStatement>()) {
+            instr.push_back(new IR::DpdkJmpGreaterStatement(jc->label,
+                        replaceIfCopy(jc->src1, false), replaceIfCopy(jc->src2)));
+        } else if (auto je = stmt->to<IR::DpdkJmpEqualStatement>()) {
+            instr.push_back(new IR::DpdkJmpEqualStatement(je->label,
+                        replaceIfCopy(je->src1, false), replaceIfCopy(je->src2)));
+        } else if (auto jne = stmt->to<IR::DpdkJmpNotEqualStatement>()) {
+            instr.push_back(new IR::DpdkJmpNotEqualStatement(jne->label,
+                        replaceIfCopy(jne->src1, false), replaceIfCopy(jne->src2)));
+       } else if (auto l = stmt->to<IR::DpdkLearnStatement>()) {
+            instr.push_back(new IR::DpdkLearnStatement(l->action,
+                        replaceIfCopy(l->timeout, false), replaceIfCopy(l->argument, false)));
+       } else if (auto m = stmt->to<IR::DpdkMeterExecuteStatement>()) {
+            instr.push_back(new IR::DpdkMeterExecuteStatement(m->meter,
+                    replaceIfCopy(m->index), replaceIfCopy(m->length),
+                    replaceIfCopy(m->color_in), replaceIfCopy(m->color_out)));
+       } else {
+            instr.push_back(stmt);
+        }
+    }
+
+    IR::IndexedVector<IR::DpdkAsmStatement> instrr;
+    for (auto stmt = instr.rbegin(); stmt != instr.rend(); stmt++) {
+        instrr.push_back(*stmt);
+    }
+    return instrr;
+}
+
 size_t ShortenTokenLength::count = 0;
 }  // namespace DPDK
--- a/backends/dpdk/dpdkAsmOpt.h
+++ b/backends/dpdk/dpdkAsmOpt.h
@@ -29,6 +29,7 @@ limitations under the License.
 #include "ir/ir.h"
 #include "lib/gmputil.h"
 #include "lib/json.h"
+#include "dpdkUtils.h"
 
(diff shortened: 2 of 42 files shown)


mbudiu-vmw:
Here you are essentially building new optimizations over your own IR. It would be probably more suitable for this code to be reviewed by people in your team. I will take a look, but it won't be very in-depth.

backends/dpdk/dpdkAsmOpt.h
```diff
@@ -290,6 +291,68 @@ class ShortenTokenLength : public Transform {
     }
 };
 
+// This pass identify redundant copies/moves and eliminate it
```
mbudiu-vmw, line 294: indentifies ... and eliminates them

backends/dpdk/dpdkAsmOpt.h
```diff
@@ -290,6 +291,68 @@ class ShortenTokenLength : public Transform {
...
+// and propogate original sources.
+class CopyPropagationAndElimination : public Transform {
+    std::unordered_map<cstring, int> usesInfo;
```
mbudiu-vmw, line 297: so what are all these cstrings?

mbudiu-vmw: you seem to depend on the fact that names are unique. you should probably document that.

backends/dpdk/dpdkAsmOpt.cpp
```diff
@@ -264,5 +264,236 @@ bool ValidateTableKeys::preorder(const IR::DpdkAsmProgram *p) {
...
+                                                                        bool allowConst) {
+    if (dontEliminate.count(str))
+        return nullptr;
+    if (!str.startsWith("m."))
+        return nullptr;
+    auto expr = replacementMap[str];
+    const IR::Expression* prev = nullptr;
+    while (expr != nullptr
+        && (!allowConst ? expr->is<IR::Member>() : true)
+        && str != expr->toString()
```
mbudiu-vmw, line 277: For the P4 IR toString is lossy, it does not capture the entire expression.
Perhaps this is different for your part of the IR.

mbudiu-vmw: you call expr->toString() many times, perhaps you can do it only once.

backends/dpdk/dpdkAsmOpt.cpp
```diff
@@ -264,5 +264,236 @@ bool ValidateTableKeys::preorder(const IR::DpdkAsmProgram *p) {
...
+
+void CopyPropagationAndElimination::markUseDef(const IR::DpdkUnaryStatement *u) {
+    usesInfo[u->src->toString()]++;
+    defInfo[u->dst->toString()]++;
+}
+
+void CopyPropagationAndElimination::markUseDef(const IR::DpdkCastStatement *c) {
+    usesInfo[c->src->toString()]++;
+    defInfo[c->dst->toString()]++;
+    replacementMap[c->dst->toString()] = c->src;
```
mbudiu-vmw, line 313: is the cast a noop?

backends/dpdk/dpdkAsmOpt.cpp
```diff
@@ -264,5 +264,236 @@ bool ValidateTableKeys::preorder(const IR::DpdkAsmProgram *p) {
...
+        } else {
+            instr.push_back(stmt);
+            newUsesInfo[src]++;
+        }
+    }
+}
+
+void CopyPropagationAndElimination::collectUseDef(IR::IndexedVector<IR::DpdkAsmStatement> stmts) {
+    for (auto stmt : stmts) {
+        if (auto mv = stmt->to<IR::DpdkMovStatement>()) {
```
mbudiu-vmw, line 377: usually a visitor does this kind of dispatch.

backends/dpdk/dpdkAsmOpt.cpp
```diff
@@ -264,5 +264,236 @@ bool ValidateTableKeys::preorder(const IR::DpdkAsmProgram *p) {
...
+                    cstring name = e->header->toString() + "." + f->name.toString();
+                    defInfo[name]++;
+                }
+            if (e->length)
+                usesInfo[e->length->toString()]++;
+        } else if (auto l = stmt->to<IR::DpdkLookaheadStatement>()) {
+            auto type = typeMap->getType(l->header)->to<IR::Type_Header>();
+            if (type)
+                for (auto f : type->fields) {
+                    cstring name = l->header->toString() + "." +f->name.toString();
```
mbudiu-vmw, line 409: so sometimes you use header fields, sometimes full headers?

mbudiu-vmw: have headers with structs been flattened?

backends/dpdk/dpdkAsmOpt.cpp
```diff
@@ -264,5 +264,236 @@ bool ValidateTableKeys::preorder(const IR::DpdkAsmProgram *p) {
...
+
+void CopyPropagationAndElimination::elimCastOrMov(const IR::DpdkAsmStatement* stmt,
+                                        IR::IndexedVector<IR::DpdkAsmStatement>& instr) {
+    const IR::Expression* srcExpr = nullptr, *dstExpr = nullptr;
+    if (stmt->is<IR::DpdkMovStatement>()) {
+        auto mv = stmt->to<IR::DpdkMovStatement>();
+        srcExpr = mv->src;
+        dstExpr = mv->dst;
+    } else {
+        auto mv = stmt->to<IR::DpdkCastStatement>();
```
mbudiu-vmw, line 350: this is suspicious to me, some casts should do work, e.g., sign-extension.

backends/dpdk/dpdkAsmOpt.cpp
kamleshbhalui: no, it's copy.

backends/dpdk/dpdkAsmOpt.cpp
kamleshbhalui: yes

kamleshbhalui: > so sometimes you use header fields, sometimes full headers?

It's not like that, because there won't be such uses.
Here I am trying to construct cstring for member expression combining header  name and it's fields.

i.e. consider below dpdk statement 
`emit h.ethernet`
above statement basically defines the all fields of ethernet header in the form like
`
h.ethernet.srcAddr
h.ethernet.dstAddr
...
`

backends/dpdk/dpdkAsmOpt.cpp
kamleshbhalui: yes, will clean it up.

backends/dpdk/dpdkAsmOpt.cpp
kamleshbhalui: same goes for lookahead.

backends/dpdk/dpdkAsmOpt.cpp
kamleshbhalui: dpdk cast just blindly convert it to plain mov/copy. https://github.com/p4lang/p4c/blob/main/backends/dpdk/spec.cpp#L536

backends/dpdk/dpdkAsmOpt.cpp
kamleshbhalui: yes but we need ordered statement list, which dpdk backend already provides.

backends/dpdk/dpdkAsmOpt.cpp
mbudiu-vmw: I mean, it does never change the bit pattern?
That's strange, I would expect some casts to perform masking and others to perform sign-extension.
Like casting int<8> to int<16>.

backends/dpdk/dpdkAsmOpt.cpp
mbudiu-vmw: visitors will traverse the list in the order of the statements inside.

backends/dpdk/dpdkAsmOpt.h
kamleshbhalui: done

backends/dpdk/dpdkAsmOpt.cpp
kamleshbhalui: is it possible to get statements in reverse order also?

kamleshbhalui:
Addressed all comments

backends/dpdk/dpdkAsmOpt.cpp
kamleshbhalui: I meant can we make visitor visit statement in reverse order ? Instead of collecting all statements in a vector and reversing it.

mbudiu-vmw: no

backends/dpdk/dpdkAsmOpt.h
```diff
@@ -290,6 +291,246 @@ class ShortenTokenLength : public Transform {
...
+    std::set<cstring> dontEliminate;
+
+    explicit CollectUseDefInfo(P4::TypeMap* typeMap) : typeMap(typeMap) {
+        dontEliminate.insert("m.pna_main_output_metadata_output_port");
+        dontEliminate.insert("m.psa_ingress_output_metadata_drop");
+        dontEliminate.insert("m.psa_ingress_output_metadata_egress_port");
+    }
+
+    bool preorder(const IR::DpdkJmpCondStatement *b) override {
+        usesInfo[b->src1->toString()]++;
```
mbudiu-vmw, line 315: so you are not really computing any reachability here, just assuming that all instructions are reachable and marking their operands as used.
This also won't discover dead writes (which are immediately overwritten.)

backends/dpdk/dpdkAsmOpt.h
```diff
@@ -290,6 +291,246 @@ class ShortenTokenLength : public Transform {
...
+        usesInfo[b->src2->toString()]++;
+        return false;
+    }
+
+    bool preorder(const IR::DpdkLearnStatement *b) override {
+        usesInfo[b->timeout->toString()]++;
```
mbudiu-vmw, line 321: Makes me wonder whether you could define some methods for the base statement class which returns a list of source and destination expressions. Then you need only one function instead of one preorder for each statement.

backends/dpdk/dpdkAsmOpt.h
```diff
@@ -290,6 +291,246 @@ class ShortenTokenLength : public Transform {
...
+
+    bool preorder(const IR::DpdkCounterCountStatement* c) override {
+        usesInfo[c->index->toString()]++;
+        if (c->incr)
+            usesInfo[c->incr->toString()]++;
+        return false;
+    }
+
+    bool preorder(const IR::DpdkRegisterDeclStatement* r) override {
+            usesInfo[r->size->toString()]++;
```
mbudiu-vmw, line 450: indentation

mbudiu-vmw: here and below

backends/dpdk/dpdkAsmOpt.h
```diff
@@ -290,6 +291,246 @@ class ShortenTokenLength : public Transform {
...
+    }
+
+    bool preorder(const IR::DpdkRegisterWriteStatement* r) override {
+            usesInfo[r->index->toString()]++;
+            return false;
+    }
+
+    bool preorder(const IR::DpdkTable *t) override {
+        auto keys = t->match_keys;
+        if (keys)
```
mbudiu-vmw, line 467: indentation is off here

backends/dpdk/dpdkAsmOpt.h
```diff
@@ -290,6 +291,246 @@ class ShortenTokenLength : public Transform {
...
+    }
+
+    const IR::Expression* getIrreplaceableExpr(cstring str, bool allowConst);
+    const IR::Expression* replaceIfCopy(const IR::Expression *expr, bool allowConst = true);
+    const IR::DpdkAsmStatement* elimCastOrMov(const IR::DpdkAsmStatement* stmt);
+    IR::IndexedVector<IR::DpdkAsmStatement> copyPropAndDeadCodeElim(
+                                                IR::IndexedVector<IR::DpdkAsmStatement> stmts);
+
+    void clear() {
+        if (collectUseDef)
```
mbudiu-vmw, line 503: another alternative is to allocate a new visitor for each visit instead of reusing it.

backends/dpdk/dpdkAsmOpt.h
```diff
@@ -290,6 +291,246 @@ class ShortenTokenLength : public Transform {
...
+    const IR::Node* preorder(IR::DpdkAction* a) override {
+        clear();
+        collectUseDef = new CollectUseDefInfo(typeMap);
+        collectUseDef->setCalledBy(this);
+        a->apply(*collectUseDef);
+        return a;
+    }
+
+    const IR::Node* preorder(IR::DpdkListStatement *l) override {
+        clear();
```
mbudiu-vmw, line 516: can you have nested lists? then this may be wrong.

mbudiu-vmw (approved):
I am approving, but perhaps this code could be improved.

backends/dpdk/dpdkAsmOpt.h
kamleshbhalui: no we can not have nested list.

backends/dpdk/dpdkAsmOpt.h
kamleshbhalui: we tried use base class as much possible, where source and destination are similar i.e. DpdkBinaryStatement DpdkJmpCondStatement etc..

backends/dpdk/dpdkAsmOpt.h
kamleshbhalui: yes we consider all statements are reachable currently.

backends/dpdk/dpdkAsmOpt.h
kamleshbhalui: done

backends/dpdk/dpdkAsmOpt.h
kamleshbhalui: done

kamleshbhalui:
@mbudiu-vmw can we merge it?

PR #3409, opened 2022-06-28
Add support for atomic annotation for registers

kmateuszssak:
This PR adds support for an atomic annotation for registers.
Due to constrained BPF environment a use of an atomic annotation is limited to cases where one `register.read()` and one `register.write()` operations are perfomed (on the same register index).

### DIFF
--- a/backends/ebpf/psa/ebpfPsaGen.cpp
+++ b/backends/ebpf/psa/ebpfPsaGen.cpp
@@ -690,6 +690,76 @@ bool ConvertToEBPFControlPSA::preorder(const IR::ExternBlock* instance) {
     return false;
 }
 
+void ConvertToEBPFControlPSA::processAtomicAnnotation(const IR::BlockStatement* b) {
+    auto ann = b->getAnnotation(IR::Annotation::atomicAnnotation);
+    cstring registerName = cstring::empty;
+
+    if (ann) {
+        int registerReadCounter = 0;
+        int registerWriteCounter = 0;
+        for (auto comp : b->components) {
+            const IR::MethodCallExpression *mExpr = nullptr;
+            if (auto assign = comp->to<IR::AssignmentStatement>()) {
+                mExpr = assign->right->to<IR::MethodCallExpression>();
+            } else if (comp->is<IR::MethodCallExpression>()) {
+                mExpr = comp->to<IR::MethodCallExpression>();
+            }
+
+            if (mExpr) {
+                auto mi = P4::MethodInstance::resolve(mExpr,
+                                                      control->program->refMap,
+                                                      control->program->typeMap);
+                auto ext = mi->to<P4::ExternMethod>();
+                if (ext != nullptr) {
+                    auto decl = ext->object;
+                    auto declType = ext->originalExternType;
+                    if (declType->name.name == "Hash")
+                        continue;
+                    else if (declType->name.name == "Register") {
+                        if (registerName.isNullOrEmpty())
+                            registerName = EBPFObject::externalName(decl);
+                        else if (registerName != EBPFObject::externalName(decl)) {
+                            ::error("You can handle only one register in the same atomic block");
+                            return;
+                        } else if (ext->method->name.name == "read" &&
+                                ++registerReadCounter > 1) {
+                            ::error("You can handle only one register read operation"
+                                    " in the same atomic block");
+                            return;
+                        } else if (ext->method->name.name == "write" &&
+                                ++registerWriteCounter > 1) {
+                            ::error("You can handle only one register write operation"
+                                    " in the same atomic block");
+                            return;
+                        }
+                    } else {
+                        ::error("You can handle only one Register, Hash "
+                                "externs in the same atomic block");
+                        return;
+                    }
+                } else {
+                    ::error("You can handle only one register in the same atomic block");
+                    return;
+                }
+            }
+        }
+
+        auto reg = control->getRegister(registerName);
+        if (reg)
+            reg->isAtomic = true;
+        else {
+            ::error("No register found: %s", registerName);
+            return;
+        }
+    }
+}
+
+bool ConvertToEBPFControlPSA::preorder(const IR::BlockStatement* b) {
+    auto ret = Inspector::preorder(b);
+    processAtomicAnnotation(b);
+    return ret;
+}
+
 // =====================EBPFDeparser=============================
 bool ConvertToEBPFDeparserPSA::preorder(const IR::ControlBlock *ctrl) {
     if (pipelineType == TC_INGRESS) {
--- a/backends/ebpf/psa/externs/ebpfPsaRegister.cpp
+++ b/backends/ebpf/psa/externs/ebpfPsaRegister.cpp
@@ -91,8 +91,34 @@ void EBPFRegisterPSA::emitKeyType(CodeBuilder* builder) {
 
 void EBPFRegisterPSA::emitValueType(CodeBuilder* builder) {
     builder->emitIndent();
-    builder->append("typedef ");
-    this->valueType->declare(builder, valueTypeName, false);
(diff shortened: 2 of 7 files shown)


mbudiu-vmw (changes requested):
One of the ptf tests is failing.

backends/ebpf/psa/externs/ebpfPsaRegister.cpp
```diff
@@ -91,8 +91,34 @@ void EBPFRegisterPSA::emitKeyType(CodeBuilder* builder) {
...
+                    builder->endOfStatement(true);
+                }
+            }
+        } else {
+            builder->emitIndent();
+            this->valueType->declare(builder, "value", false);
+            builder->endOfStatement(true);
+        }
+        builder->emitIndent();
+        builder->append("struct bpf_spin_lock lock");
```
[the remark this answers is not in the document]
kmateuszssak: It's not possible to do in BPF. Lock can be associated only with BPF map. Register in fact is a BPF map. In the [BPF helpers documentation](https://man7.org/linux/man-pages/man7/bpf-helpers.7.html) there are more details of what can be done (section bpf_spin_lock).

backends/ebpf/psa/externs/ebpfPsaRegister.cpp
```diff
@@ -161,27 +187,33 @@ void EBPFRegisterPSA::emitRegisterRead(CodeBuilder* builder, const P4::ExternMet
...
     builder->emitIndent();
-    builder->target->emitTableLookup(builder, dataMapName, indexParamName, valueName);
+    builder->target->emitTableLookup(builder, dataMapName, indexParamName, readValueName);
     builder->endOfStatement(true);
 
     builder->emitIndent();
-    builder->appendFormat("if (%s != NULL) ", valueName.c_str());
+    builder->appendFormat("if (%s != NULL) ", readValueName.c_str());
     builder->blockStart();
+    if (this->isAtomic) {
```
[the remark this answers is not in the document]
kmateuszssak: The spinlock is protecting a region between `register.read()` and `register.write()`. Other cases are not possible to protect because of spinlocks constraints.

kmateuszssak:
@mbudiu-vmw I've addressed comments. Please merge if it's okay.

backends/ebpf/psa/externs/ebpfPsaRegister.cpp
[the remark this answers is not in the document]
mbudiu-vmw: What if you have multiple reads followed by a write?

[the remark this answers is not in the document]
mbudiu-vmw: When you have a register is it shared across all cores?

backends/ebpf/psa/ebpfPsaGen.cpp
```diff
@@ -690,6 +690,107 @@ bool ConvertToEBPFControlPSA::preorder(const IR::ExternBlock* instance) {
...
+                        continue;
+                    else if (declType->name.name == "Register") {
+                        if (registerName.isNullOrEmpty())
+                            registerName = EBPFObject::externalName(decl);
+
+                        auto indexArgExpr = ext->expr->arguments->at(0)->
+                                expression->to<IR::PathExpression>();
+                        if (registerName != EBPFObject::externalName(decl)) {
+                            ::error(ErrorType::ERR_UNSUPPORTED_ON_TARGET,
+                                    "%1%: You can handle only one register "
```
mbudiu-vmw, line 736: I would probably remove "You" and say something "target supports only ..."

backends/ebpf/psa/ebpfPsaGen.cpp
```diff
@@ -690,6 +690,107 @@ bool ConvertToEBPFControlPSA::preorder(const IR::ExternBlock* instance) {
...
+                    return;
+                }
+            }
+        }
+
+        auto reg = control->getRegister(registerName);
+        if (reg)
+            reg->usedInAtomicBlock = true;
+        else {
+            ::error(ErrorType::ERR_UNEXPECTED,
```
mbudiu-vmw, line 781: Should this be a BUG?

backends/ebpf/psa/ebpfPsaGen.cpp
```diff
@@ -690,6 +690,107 @@ bool ConvertToEBPFControlPSA::preorder(const IR::ExternBlock* instance) {
...
+                            }
+                        } else if (ext->method->name.name == "write") {
+                            if (++registerWriteCounter > 1 && registerReadCounter > 0) {
+                                ::error(ErrorType::ERR_UNSUPPORTED_ON_TARGET,
+                                        "%1%: You can handle only one register write operation"
+                                        " after register read "
+                                        "in the same atomic block", mExpr);
+                                return;
+                            }
+                            if (indexName != indexArgExpr->path->name.name) {
```
mbudiu-vmw, line 755: the fact that you use the same expression does not guarantee that its value hasn't changed.
consider the example: `reg.read(i); i = i+1; reg.write(i, 0);`

mbudiu-vmw (changes requested):

backends/ebpf/psa/ebpfPsaGen.cpp
kmateuszssak: I've added more checks.

mbudiu-vmw:
cpplint is unhappy.

backends/ebpf/psa/externs/ebpfPsaRegister.cpp
```diff
@@ -180,11 +180,28 @@ void EBPFRegisterPSA::emitInstance(CodeBuilder *builder) {
                                    valueTypeName, size);
 }
 
+cstring EBPFRegisterPSA::getParamName(CodeBuilder* builder, const IR::Expression *expr,
```
mbudiu-vmw, line 183: I would document this function, the name is somewhat misleading.
It really makes up a variable that holds the value of the index or value used to access a register.

mbudiu-vmw: So has there been a pass before this one which makes sure that any complex expression used to index in a register has been substituted by a temporary?

backends/ebpf/psa/ebpfPsaGen.cpp
```diff
@@ -690,6 +690,31 @@ bool ConvertToEBPFControlPSA::preorder(const IR::ExternBlock* instance) {
...
+        readIndexExpr = expr;
+    }
+    bool preorder(const IR::AssignmentStatement* a) override;
+
+    bool readIndexChanged();
+};
+
+bool IndexVisitor::preorder(const IR::AssignmentStatement *a) {
+    if (auto expr = a->left->to<IR::PathExpression>()) {
+        if(expr->path->name.name == readIndexExpr->path->name.name)
```
mbudiu-vmw, line 707: no space

backends/ebpf/psa/ebpfPsaGen.cpp
```diff
@@ -690,6 +690,31 @@ bool ConvertToEBPFControlPSA::preorder(const IR::ExternBlock* instance) {
     return false;
 }
 
+class IndexVisitor : public Inspector {
```
mbudiu-vmw, line 693: I would give a more precise name to this visitor, like "PathOnLHS"; it really detects whether there is an assignment to a specific path expression.

backends/ebpf/psa/ebpfPsaGen.cpp
mbudiu-vmw: Also, for visitors that are used only locally I often embed them in an empty namespace to prevent them from polluting the C++ global namespace of names,

mbudiu-vmw:
There is another problem: an expression can also be mutated by being an out parameter to a call.

PR #3411, opened 2022-06-28
P4C-DPDK - Incorrect constant value when there is a substract operation involved with a constant

github-sajan:
This PR addresses the issue in dpdk backend when there is subtraction operation with constant.

If isSigned flag is set and is not 8-bit aligned, sign extension (32/64 bit) is done for the constant
If isSigned flag is not set and is not 8-bit aligned,
insert an "BAnd" node after the "Add" for masking the value to the original bitwidth.


backends/dpdk/dpdkHelpers.cpp
```diff
@@ -134,27 +135,58 @@ bool ConvertStatementToDpdk::preorder(const IR::AssignmentStatement *a) {
                 src1Op = src1Member;
             }
         }
+        /* If a constant is present in binary operation and it is not 8-bit
+           aligned and it is signed, then the constant value is sign extended
+           with a width of either 32 bit if its original bit-width is less than
+           32 bit or 64 bit if the bit-width of constant is greater than 32 bit
+        */
+
+        if (!isEightBitAligned(src2Op) && src2Op->is<IR::Constant>()) {
```
mbudiu-vmw, line 144: Shouldn't sign-extension be done for non-constant values too? Otherwise I think that overflow handling will be incorrect.

mbudiu-vmw (changes requested):

backends/dpdk/dpdkHelpers.cpp
github-sajan: Added shift operation for sign extension for non-constant value

backends/dpdk/dpdkHelpers.cpp
```diff
@@ -134,27 +135,64 @@ bool ConvertStatementToDpdk::preorder(const IR::AssignmentStatement *a) {
...
         } else {
             BUG("%1% not implemented.", right);
         }
+        /* If there is a binary operation present which involves non 8-bit aligned
+           fields , add BAnd operation with mask value of original bit-width to
+           avoid any result overflow */
+        if (!isEightBitAligned(left) && !isSignExtended) {
+            if (right->is<IR::Add>() || right->is<IR::Sub>() ||
+                right->is<IR::Shl>() || right->is<IR::Shr>()) {
+                auto width = left->type->width_bits();
```
mbudiu-vmw, line 191: Maybe you want to use the Constant class, this may overflow in C++.

backends/dpdk/dpdkHelpers.cpp
```diff
@@ -134,27 +135,64 @@ bool ConvertStatementToDpdk::preorder(const IR::AssignmentStatement *a) {
...
+                    auto consOrgBitwidth = src2Op->to<IR::Expression>()->type->width_bits();
+                    auto bitWidth = consOrgBitwidth < 32 ? 32 : 64;
+                    if (src2Op->is<IR::Constant>()) {
+                        auto consValue = src2Op->to<IR::Constant>()->value;
+                        auto val = consValue << (bitWidth - consOrgBitwidth);
+                        consValue = val >> (bitWidth - consOrgBitwidth);
+                        src2Op = new IR::Constant(IR::Type_Bits::get(bitWidth), consValue);
+                    } else {
+                        auto val = bitWidth - consOrgBitwidth;
+                        add_instr(new IR::DpdkShlStatement(src2Op, src2Op, new IR::Constant(val)));
```
mbudiu-vmw, line 157: so this is a sign-extension?
some languages have two kinds of shift right: arithmetic and logic. I guess you need arithmetic.
Is the dpdk shift right semantics based on the type?

mbudiu-vmw (changes requested):

backends/dpdk/dpdkHelpers.cpp
github-sajan: Shift operations are logical in DPDK.
For sign extension of non-constant variables, i am moving the msb of variable through left and right operation.
Adding snippet for the same ( sign extended 16 bit to 32 bit )
(Snip)
    #include <iostream>
#include <cstdint> 
using namespace std;

int main() {
    int16_t var;
    int b = 17;
    cin>>var;
    int32_t var2 = var << 16;
   cout<<"Size of var "<<sizeof(var)<<endl;
    cout<<hex <<var<< dec <<endl;
    

    var2 = var2 >> 16;
    cout<<"Size of var2 "<<sizeof(var2)<<endl;
    cout<<hex <<var2<< dec <<endl;
    return 0;
}

Input : -2
Output :
Size of var 2
fffe
Size of var2 4
fffffffe
(Snip)

backends/dpdk/dpdkHelpers.cpp
github-sajan: Sorry. I did not get the comment. Could you please add more to understand?

mbudiu-vmw (approved):

mbudiu-vmw:
Looks like several dpdk tests are failing with this change.

github-sajan:
> Looks like several dpdk tests are failing with this change.

Updated the outputs

PR #3412, opened 2022-06-29
Fix incorrect overload resolution for specialized (package) types

### LINKED ISSUE #3379: overloading of package constructors
Take:
```
parser mypt<t>(in t tt);

parser MyParser1(in bit tt) {
  state start {
    transition select(tt) {
     0: accept;
     _: reject;
   }
 }
}

typedef bit t;
package mypackaget(mypt<t> t2);
package mypackaget(mypt<t> t1, mypt<t> t2);
mypackaget(MyParser1()) t4;
```
This compiles which is shocking but if you change mypackaget to be a generic, it does not:
```
parser mypt<t>(in t tt);

parser MyParser1(in bit tt) {
  state start {
    transition select(tt) {
     0: accept;
     _: reject;
   }
 }
}

package mypackaget<t>(mypt<t> t2);
package mypackaget<t>(mypt<t> t1, mypt<t> t2);
mypackaget<bit>(MyParser1()) t4;
```
It is shocking that one is accepted but not the other. I would rather have both being rejected.

mbudiu-vmw:
Fixes #3379

### DIFF
--- a/frontends/common/resolveReferences/resolveReferences.cpp
+++ b/frontends/common/resolveReferences/resolveReferences.cpp
@@ -171,6 +171,9 @@ const IR::Vector<IR::Argument> *ResolutionContext::methodArguments(cstring name)
             if (auto type = decl->type->to<IR::Type_Name>()) {
                 if (type->path->name == name)
                     return decl->arguments; }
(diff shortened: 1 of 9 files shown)


rst0git (approved):
LGTM

PR #3419, opened 2022-07-01
Fix def-use analysis bug when handling virtual methods

### LINKED ISSUE #3417: Compiler bug in simplifyDefUse
When compiling this code in the following extern inside a parser, the compiler throws a bug saying that it reached an unreachable expression in FindUninitialized.
```c
    Register<bit<128>, bit<16>>(32w0, 128w0) reg;
    
    RegisterAction<bit<128>, bit<16>, bit<64>>(reg) regAct = {
        void apply(inout bit<128> val, out bit<64> rv) {
            return;
        }
    };
```
Full error:  (with -vvvv)
```
SimplifyDefUse_0_Cloner
PassRepeated_0_TypeChecking
/mnt/find_uninit_v1model.p4(41): [--Wwarn=uninitialized_out_param] warning: out parameter 'rv' may be uninitialized when 'apply' terminates
        void apply(inout bit<128> val, out bit<64> rv) {
                                                   ^^
/mnt/bf-p4c-tools/find_uninit_v1model.p4(41)
        void apply(inout bit<128> val, out bit<64> rv) {
             ^^^^^
In file: /mnt/compilers/p4c/frontends/p4/simplifyDefUse.cpp:402
Compiler Bug: reached an unreachable expression in FindUninitialized

```
I have attached the corresponding p4 below.
[find_uninit_v1model.p4.txt](https://github.com/p4lang/p4c/files/9021371/find_uninit_v1model.p4.txt)

---
These were the last few passes:

mbudiu-vmw:
Fixes #3417

### DIFF
--- a/frontends/p4/simplifyDefUse.cpp
+++ b/frontends/p4/simplifyDefUse.cpp
@@ -134,7 +134,7 @@ class DeclarationToExpression {
 
 DeclarationToExpression* DeclarationToExpression::instance = nullptr;
 
-class HeaderDefinitions {
+class HeaderDefinitions : public IHasDbPrint {
     ReferenceMap* refMap;
     TypeMap* typeMap;
     StorageMap* storageMap;
@@ -156,6 +156,13 @@ class HeaderDefinitions {
     HeaderDefinitions(ReferenceMap* refMap, TypeMap* typeMap, StorageMap* storageMap) :
         refMap(refMap), typeMap(typeMap), storageMap(storageMap)
         { CHECK_NULL(refMap); CHECK_NULL(typeMap); CHECK_NULL(storageMap); }
+
+
+
+    void dbprint(std::ostream& out) const {
+        for (auto it : defs)
+            out << *it.first << " -> " << toString(it.second) << std::endl;
+    }
 
     /// A helper function for getting a storage location from an expression.
     /// In case of accessing a header stack with non-constant index, it returns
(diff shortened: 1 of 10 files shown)


mbudiu-vmw:
The fix is just 1 line, but I added a few helper methods as well.

frontends/p4/simplifyDefUse.cpp
```diff
@@ -572,6 +580,7 @@ class FindUninitialized : public Inspector {
         headerDefs->clear();
         initHeaderParams(parser->getApplyMethodType()->parameters);
         visitVirtualMethods(parser->parserLocals);
+        unreachable = false;
```
mbudiu-vmw, line 583: this is the fix; this code is similar to the handling of controls.
If there are virtual methods they may leave "unreachable" on true.

frontends/p4/ternaryBool.h
```diff
@@ -17,15 +17,17 @@ limitations under the License.
 #ifndef _FRONTENDS_P4_TERNARYBOOL_H_
 #define _FRONTENDS_P4_TERNARYBOOL_H_
 
+#include "lib/cstring.h"
+
 namespace P4 {
+enum class TernaryBool {
```
fruffy, line 23: Might be missing a newline here?

fruffy (approved):

PR #3425, opened 2022-07-08
Added new P4C option

OleksandrLetychevskyiIntel:
Added option and function to set value for `noInclude` argument.
If true do not generate #include statements. This can help with debugging


mbudiu-vmw (approved):

mbudiu-vmw:
I would rename this flag, since this name suggests that the compiler will not process `#include` directives.
It could be named --doNotEmitIncludes

PR #3428, opened 2022-07-09
remove unused include

vhavel:
std::variant isn't actually used. This include also prevents gcc-6 based builds as variant is supported only by newer gccs.


mvido (approved):

mvido:
For reference, this unused include was added by https://github.com/p4lang/p4c/pull/3405

mbudiu-vmw (approved):

PR #3431, opened 2022-07-13
ordered_set: operator<: Fix for a case of strict inclusion.

DoctorNoobingstoneIPresume:
When all elements compared so far in sets `x` and `y` have been equivalent
and one of the sets (`x`) ends, but the other one (`y`) does not end
(i.e. `x` is a strict prefix of `y`),
we used to get `y < x`, but with this changeset we are going to get `x < y`.

Example (in pseudocode) (with case-insensitive comparison):
  we used to get `{'A', 'b'} > {'a', 'B', 'c'}`,
  now fixed to   `{'A', 'b'} < {'a', 'B', 'c'}`.
This is similar to how `std::lexicographical_compare` behaves.

### DIFF
--- a/lib/ordered_set.h
+++ b/lib/ordered_set.h
@@ -99,11 +99,11 @@ class ordered_set {
         // have a set of ordered_sets (or use ordered_set as a map key).
         auto it = a.data_map.begin();
         for (auto &el : data_map) {
-            if (it == a.data_map.end()) return true;
+            if (it == a.data_map.end()) return false;
             if (mapcmp()(el.first, it->first)) return true;
             if (mapcmp()(it->first, el.first)) return false;
             ++it; }
-        return false; }
+        return it != a.data_map.end(); }
 
     // FIXME add allocator and comparator ctors...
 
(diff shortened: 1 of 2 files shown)


DoctorNoobingstoneIPresume:
Perhaps we could replace the whole implementation (with the loop) with just:

```
return std::lexicographical_compare
(
    data_map.begin (), data_map.end (),
    a.data_map.begin (), a.data_map.end (),
    data_map.key_comp ()
);
```

mbudiu-vmw:
Can you please try to run `make cpplint` prior to submitting the PR?

DoctorNoobingstoneIPresume:
> Can you please try to run `make cpplint` prior to submitting the PR?

Thank you for review. I will try to do that. I dare suggest that maybe we remove cpplint ? :)

E.g.: Alternative implementation with `std::lexicographical_compare` (in comment above: https://github.com/p4lang/p4c/pull/3431#issuecomment-1183228462) is never going to pass cpplint as it is, but I think it is clearer -- the additional whitespace makes obvious what is similar and what is different between the two lines.

mbudiu-vmw:
Having tools makes it much easier to enforce coding standards. Without tools like cpplint conversations can devolve into religious arguments. For example, I hate the defaults of tools like rustfmt, but I save a lot of time avoiding arguing about the right way to format Rust. In the end, time is more valuable.

PR #3432, opened 2022-07-14
ordered_set: Leaner, faster, richer, safer.

DoctorNoobingstoneIPresume:
Previously, for each element, we used to allocate (and later deallocate)
two separate nodes:

  - a node inside a list (two link-pointers plus payload of two pointers);
  - a node inside a tree (three link-pointers plus one bool plus actual payload).

Now, we only allocate a single node:

  - a node inside a set (three link-pointers plus payload).

The payload of the node inside the set
contains the two link-pointers for list operations
and then the actual payload (an element of type T).

Benefits:
  - fewer allocations/deallocations;
  - less memory consumed (the payload of two pointers in the list was wasteful);
  - better locality of data;
  - superiour error-safety guarantees
    (the intrusive container operations offer the strongest guarantee: nofail)
    (=> merging a fix such as PR #3294 is no longer needed);
  - improved speed.

Side note:
  We have preferred boost::container::set to std::set
  because the former by default tries to merge the colour flag (a bool flag)
  with one of the three links, therefore further reducing memory consumption.
  TODO: Consider how this works in a World of Garbage Collection.

The current changeset is a draft -- not yet complete for the purposes of p4c,
but offering the essential features of ordered_set:
iteration in sorted order (now with backwards iteration as well)
and iteration in user-controlled (sequence) order.

It represents a proof of concept and a request for comments. (-:

### DIFF
--- a/lib/ordered_set.h
+++ b/lib/ordered_set.h
@@ -17,234 +17,540 @@ limitations under the License.
 #ifndef _LIB_ORDERED_SET_H_
 #define _LIB_ORDERED_SET_H_
 
+#include <boost/container/set.hpp>
+#include <boost/intrusive/list.hpp>
+#include <boost/tuple/tuple.hpp>
+
+#include <algorithm>
 #include <functional>
 #include <initializer_list>
-#include <list>
-#include <map>
-#include <set>
 #include <utility>
+#include <iterator>
 
-// Remembers items in insertion order
-template <class T, class COMP = std::less<T>, class ALLOC = std::allocator<T>>
-class ordered_set {
- public:
-    typedef T                   key_type;
-    typedef T                   value_type;
-    typedef COMP                key_compare;
-    typedef COMP                value_compare;
-    typedef ALLOC               allocator_type;
-    typedef const T             &reference;
-    typedef const T             &const_reference;
 
- private:
-    typedef std::list<T, ALLOC> list_type;
-    list_type                   data;
+//============================================================================
+// EnlistedNode <T>:
+//
+//   An object of this type holds:
+//     - two pointers (to the previous node and to the next node);
+//     - a payload of type T.
+//============================================================================
 
- public:
-    typedef typename list_type::iterator                iterator;
-    typedef typename list_type::const_iterator          const_iterator;
-    typedef std::reverse_iterator<iterator>             reverse_iterator;
-    typedef std::reverse_iterator<const_iterator>       const_reverse_iterator;
+struct MyTag1;
 
- private:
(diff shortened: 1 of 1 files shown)


mbudiu-vmw:
This code does not compile in CI.

lib/ordered_set.h
```diff
@@ -17,234 +17,540 @@ limitations under the License.
 #ifndef _LIB_ORDERED_SET_H_
 #define _LIB_ORDERED_SET_H_
 
+#include <boost/container/set.hpp>
```
ChrisDodd, line 20: Why use `boost::set` rather than `std::set` -- the latter is more portable and reliable

lib/ordered_set.h
```diff
@@ -17,234 +17,540 @@ limitations under the License.
...
+#include <boost/intrusive/list.hpp>
+#include <boost/tuple/tuple.hpp>
```
ChrisDodd, line 22: We should always use `std::tuple` rather than `boost::tuple` for better standards compliance.

lib/ordered_set.h
```diff
@@ -17,234 +17,540 @@ limitations under the License.
...
+    typedef
+        EnlistedNode <T>
+        MyEnlistedNode;
+
+    typedef
+        EnlistedComp <Comp>
+        MyEnlistedComp;
+
+    typedef
+        typename Alloc::template rebind <MyEnlistedNode>::other
```
ChrisDodd, line 166: rebind is deprecated and removed in some newer compilers -- need to use `rebind_alloc` instead.

ChrisDodd (commented):
Where performance is a concern, I would expect that an `hvec_set` (based on hashvector.h similar to hvec_map) would be much more efficient.

lib/ordered_set.h
DoctorNoobingstoneIPresume: Please accept apologies for my replying late.

As explained in my first comment above, the Boost version of `set` attempts to merge the colour flag with one of the pointers => each node in the tree demands smaller block on the heap.

DoctorNoobingstoneIPresume:
> Hi @DoctorNoobingstoneIPresume are you planning to continue work on this? It looks very useful.

Thank you so much... I have now remembered this -- and with joy, I dare say. I hope to, indeed, but need some time to get things back into my head.

Thank you for the kind message and the encouragement. It can make a difference in a community. ❤️

lib/ordered_set.h
fruffy: We are in the processing of getting rid of boost in the compiler. On the other hand, we now have support for Abseil.

PR #3433, opened 2022-07-14
DPDK Backend: Emit name information for externs in context json

usha1830:
The names of extern objects in spec file and bfrt json/p4info  differ in cases when the extern declarations are not in the global scope. Compiler adds some prefix and suffix to the extern object's name based on the scope of the the object.  New array of externs needs to be added into context json to handle this mismatch issue and to allow control plane software to access the name of the extern as in the spec file corresponding to the control plane name.

The schema changes are as follows:
```
"externs" : [
   {
    name : "ipv4_host_counter",
    target_name: "ipv4_host_counter_1",
    extern_type: "counter",    
    attributes {
        type: packets_and_bytes
    }    
    },
{
    name : "reg1",
    target_name: "reg1_0",
    extern_type: "register",    
    attributes {
    }    
},
....
]
```

 
[context_all.json.txt](https://github.com/p4lang/p4c/files/9108154/context_all.json.txt)
[psa-example-dpdk-externs.p4.txt](https://github.com/p4lang/p4c/files/9108157/psa-example-dpdk-externs.p4.txt)


backends/dpdk/dpdkContext.cpp
```diff
@@ -109,6 +109,45 @@ void DpdkContextGenerator::CollectTablesAndSetAttributes() {
         }
     }
 
+    for (auto ed :  structure->externDecls) {
+        if (auto type = ed->type->to<IR::Type_Specialized>()) {
+            auto externTypeName = type->baseType->path->name.name;
+            if (externTypeName == "Counter" ||
```
mbudiu-vmw, line 115: do you want to ignore silently all other externs?

mbudiu-vmw (approved):

PR #3435, opened 2022-07-14
Load table action from json without 'name' field for BMV2

VolodymyrPeschanenkoIntel:
This small changes in script allow to load table action from JSON file without `name` field.
Without this changes this example finished with the message:
` Exception  'name'`


backends/bmv2/bmv2stf.py
```diff
@@ -320,8 +320,10 @@ def __init__(self, jsonTable):
         self.key = TableKey()
         self.actions = {}
         for k in jsonTable["key"]:
-            name = k["name"]
-            if name is None:
+            name = ""
+            if "name" in k:
+                name = k["name"]
+            else:
```
jafingerhut, line 326: Given the documentation for the `name` and `target` keys of a table definition in the BMv2 JSON file that you can find here https://github.com/p4lang/behavioral-model/blob/main/docs/JSON_format.md, then search for the only occurrence of "name: this attribute", this change appears like a correct one to me, in order to handle a larger set of correct BMv2 JSON files.  Pinging @antoninbas to see if he has any comment on this change, but I will approve this change regardless, in case he is busy.

jafingerhut (approved):
Looks good to me.  It is a bit odd to have a test case that intentionally uses fields of a header that is known to be invalid, but I suspect it might be difficult to find another test case that triggers the same issue, and the compiler does give good warnings for the source program.

testdata/p4_16_samples/invalid-hdr-warnings3-bmv2.stf
```diff
@@ -0,0 +1,16 @@
+# STF test for /home/vpeschax/networking.switching.barefoot.bf-p4c-tools/compilers/p4c/build/testgen/testgen-p4c-bmv2/invalid-hdr-warnings3.out/invalid-hdr-warnings3
```
fruffy, line 1: I would remove these hardcoded paths and the comments are not quite necessary yet.

testdata/p4_16_samples/invalid-hdr-warnings3-bmv2.p4
```diff
@@ -0,0 +1,73 @@
+#include <core.p4>
```
fruffy, line 1: Instead of adding a new program, why not rename the old one?

antoninbas (commented):
I didn't look at the test programs, and I would expect p4c to always generate the "name" attribute, but I agree with the change to `bmv2stf.py`. The "name" attribute is indeed optional.

PR #3440, opened 2022-07-15
DPDK Backend: Add support for tdi.json

usha1830:
Tdi json file needs to be emitted from p4c-dpdk to interface with the [TDI](https://github.com/p4lang/tdi) .The contents of tdi json file is same as bfrt json with few additional program/environment related fields given below
```
{
"program_name" : "psa-example-mask-range",
  "build_date" : "Mon Jun 27 13:36:57 2022",
  "compile_command" : "./p4c-dpdk --arch psa -o test.spec ../testdata/p4_16_samples/psa-example-mask-range.p4 --bf-rt-schema bfrt.json --tdi tdi.json --p4runtime-files p4info.txt --context context.json",
  "compiler_version" : "0.1 (SHA: ad5b4a2d1 BUILD: RELEASE)",
  "schema_version" : "0.1",
  "target" : "DPDK",
<< bfrt json contents >>
}
```
Attached sample tdi json and corresponding bfrt json.
[bfrt.json.txt](https://github.com/p4lang/p4c/files/9120183/bfrt.json.txt)
[psa-example-mask-range.p4.txt](https://github.com/p4lang/p4c/files/9120184/psa-example-mask-range.p4.txt)
[tdi.json.txt](https://github.com/p4lang/p4c/files/9120185/tdi.json.txt)

### DIFF
--- a/backends/dpdk/control-plane/bfruntime_ext.cpp
+++ b/backends/dpdk/control-plane/bfruntime_ext.cpp
@@ -227,10 +227,10 @@ BFRuntimeSchemaGenerator::genSchema() const {
         json->emplace("build_date", cstring(options.getBuildDate()));
         json->emplace("compile_command", cstring(options.getCompileCommand()));
         json->emplace("compiler_version", cstring(options.compilerVersion));
-        json->emplace("schema_version", cstring("0.1"));
+        json->emplace("schema_version", tdiSchemaVersion);
         json->emplace("target", cstring("DPDK"));
     } else {
-        json->emplace("schema_version", cstring("1.0.0"));
+        json->emplace("schema_version", bfrtSchemaVersion);
     }
 
     auto* tablesJson = new Util::JsonArray();
--- a/backends/dpdk/control-plane/bfruntime_ext.cpp
+++ b/backends/dpdk/control-plane/bfruntime_ext.cpp
@@ -213,7 +213,25 @@ const Util::JsonObject*
 BFRuntimeSchemaGenerator::genSchema() const {
     auto* json = new Util::JsonObject();
 
-    json->emplace("schema_version", cstring("1.0.0"));
+    if (isTDI) {
+        cstring progName =  options.file;
+        auto fileName = progName.findlast('/');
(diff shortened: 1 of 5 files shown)


backends/dpdk/control-plane/bfruntime_ext.cpp
```diff
@@ -213,7 +213,25 @@ const Util::JsonObject*
...
+        if (fileName)
+            progName = fileName;
+        auto fileext = progName.find(".");
+        progName = progName.replace(fileext, "");
+        progName = progName.trim("/\t\n\r");
+        json->emplace("program_name", progName);
+        json->emplace("build_date", cstring(options.getBuildDate()));
+        json->emplace("compile_command", cstring(options.getCompileCommand()));
+        json->emplace("compiler_version", cstring(options.compilerVersion));
+        json->emplace("schema_version", cstring("0.1"));
```
mbudiu-vmw, line 230: should this be a constant instead of a literal?

backends/dpdk/control-plane/bfruntime_ext.cpp
```diff
@@ -213,7 +213,25 @@ const Util::JsonObject*
...
+        json->emplace("target", cstring("DPDK"));
+    } else {
+        json->emplace("schema_version", cstring("1.0.0"));
```
mbudiu-vmw, line 233: this one as well

mbudiu-vmw (approved):

backends/dpdk/control-plane/bfruntime_ext.cpp
usha1830: Version strings can have more than one decimal point. So I think a string literal should be ok here.
Do you mean instead of string literal, we should use a const string type variable to hold the version string?

mbudiu-vmw: Yes, a const string, in one place.

usha1830:
Some ptf-ebpf checks are failing. Not related to the changes in this PR.

Issue #3391, opened 2022-06-10
Can we use p4c instead of p4c-bmv2 in the Bazel build rule?


lzhzero:
https://github.com/p4lang/p4c/blob/f49d2282b61a55446b9c18f18bbe170f41fd9c96/bazel/p4_library.bzl#L143-L147

Currently the private attribute `_p4c` points to the `p4c-bmv2` binary.

We are interested in using Balze rules to compile for dpdk backend.
So can we set the rule to use `p4c` binary? Or maybe make the attribute non private so  people can choose it freely?

Thanks,
Zhang

mbudiu-vmw:
I don't know anything about bazel, but it's a build system, so it should be able to express what you need.
Feel free to submit any changes that may be needed as a PR.
Can anyone with Bazel experience answer this question?

lzhzero:
I will try to create a PR for that.

I think my question now is: what is the difference between `p4c` and `p4c-bm2-ss` or `p4c-dpdk` binaries generated by the default CMake?

I know that `p4c-bm2-ss` can use `--top4` to dump the P4 representations for different compiler stages. Are there other differences other than debugging support?

mbudiu-vmw:
p4c-bm2-ss generates code for the bmv2 backend, a json file.
p4c-dpdk generates dpdk assembly code.
All such compilers share a common front-end, but have different back-ends. See the compiler-design.pptx document in the docs folde.
p4c is, as far as I know, a driver which invokes other compiler based on flags passed.

PR #3441, opened 2022-07-15
Add bazel support for p4c-dpdk compiler

lzhzero:
1. Made `p4c` atribute of the p4_library rule configurable, so the
`p4c-dpdk` compiler can be selected if needed. The default
compiler is still `p4c-bm2-ss`.

2. Added DPDK backend specific bf-rt-schema and context output files.

3. Added everything under `p4include` folder into the default include.

5. updated blaze example with dpdk pna program vxlan.p4.
Note: This vxlan.p4 program  already exist in `backends/dpdk/examples/vxlan.p4`

### DIFF
--- a/BUILD.bazel
+++ b/BUILD.bazel
@@ -8,7 +8,7 @@ package(
 
 filegroup(
     name = "p4include",
-    srcs = glob(["p4include/*.p4"]),
+    srcs = glob(["p4include/**/*.p4"]),
 )
 
 genrule(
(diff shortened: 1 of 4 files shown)


smolkaj:
This is super cool, thanks for the extension!

I don't feel strongly, but I wonder if renaming the `p4c` parameter to `p4c_backend` makes sense -- perhaps that makes things a bit more readable from the user perspective? 

LGTM from my side either way.

mbudiu-vmw (approved):
I will wait for the changes suggested by @smolkaj to merge.

lzhzero:
Thanks for the suggestion. The PR is now updated.

I only changed the user facing blaze rule attribute name, and did not change the internal variable name `p4c`. Otherwise we might need to change other internal variable names to be consistent.

Please let me know what do you think.

Also this PR will close issue https://github.com/p4lang/p4c/issues/3391

PR #3444, opened 2022-07-15
Use Type_Dontcare for _ expressions in table entries

### LINKED ISSUE #3442: Underscore has wrong type
The following program is rejected by `p4test`,
```p4
control c();
package p(c c);

control myc () {
  action a() { }
  action b() { }
  table t {
    key = {}
    actions = { a; }
    const entries = {
      _ : a();
    }
  }
  apply { 
    t.apply();
  } 
}
p(myc ()) main;
```
with the following error:
```
foo.p4(10): [--Werror=type-error] error: Entry
      _ : a();
      ^^^^^^^
  ---- Actual error:
  Tuples with different sizes 0 vs 1
  ---- Originating from:
  foo.p4(10): Table entry has type 'tuple<>' which is not the expected type 'tuple<_>'
        _ : a();
        ^
```

I think the type of `_` should probably be inferred as `tuple<>` here...

mbudiu-vmw:
Fixes #3442
For this to work we need to remember whether the original expression was _ or { _ }.
That's why the IR class had to be augmented.
(We may need to do the same for select expressions, which suffer from a similar confusion.)

### DIFF
--- a/frontends/p4/typeChecking/typeChecker.cpp
+++ b/frontends/p4/typeChecking/typeChecker.cpp
@@ -1914,8 +1914,6 @@ const IR::Node* TypeInference::postorder(IR::Key* key) {
     }
     LOG2("Setting key type to " << dbp(keyTuple));
     setType(key, keyTuple);
-    // installing also for the original because we cannot tell which one will survive in the ir
-    LOG2("Setting key type to " << dbp(getOriginal()));
     setType(getOriginal(), keyTuple);
     return key;
 }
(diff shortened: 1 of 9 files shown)


vgurevich:
I would hold off merging this until we resolve https://github.com/p4lang/p4-spec/issues/1120.
I wonder if there is any other case, where we will need to unify `_` with `tuple<>`

mbudiu-vmw:
This is a bug, _ should unify with everything.
The solution should be different if you want such tables to be empty.

jnfoster:
Agree. This was a bug and is independent of the earlier issue about keys.

vgurevich:
No problem -- but then get ready to deal with the fallout :)

rst0git (approved):
LGTM

I can confirm that the patch in the pull request the P4 program fixes the error in https://github.com/p4lang/p4c/issues/3442

PR #3445, opened 2022-07-18
Fixed usage of yyFlexLexer with multiple parsers

davidbolvansky:
First, UBSan found issue here:

> [2022-04-27T21:17:41.988Z] /bfn/bf-p4c-compilers/p4c/frontends/parsers/v1/v1lexer.hpp:20:7: runtime error: member call on address 0x7ffeb47d6800 which does not point to an object of type 'p4FlexLexer' [2022-04-27T21:17:41.988Z] 0x7ffeb47d6800: note: object is of type 'V1::V1Lexer' [2022-04-27T21:17:41.988Z] fe 7f 00 00 78 44 cd 13 46 56 00 00 00 10 b9 26 46 56 00 00 01 00 00 00 01 00 00 00 00 00 00 00 [2022-04-27T21:17:41.988Z] ^~~~~~~~~~~~~~~~~~~~~~~ [2022-04-27T21:17:41.988Z] vptr for 'V1::V1Lexer'

I analyzed it and we should drop "yyFlexLexerOnce" hacks and use recommended ways to do it. Fixes bugs and follows bison guidelines.

Nicely explained here:
https://stackoverflow.com/questions/35606354/multiple-parsers-in-flex-bison-include-fails

While here, I also replaced parser_class_name with api.parser.class, as bison warns that parser_class_name  is deprecated.

### DIFF
--- a/frontends/parsers/p4/p4lexer.hpp
+++ b/frontends/parsers/p4/p4lexer.hpp
@@ -4,31 +4,12 @@
 // FlexLexer.h requires you to provide an external #include guard so that it can
 // be included multiple times, each time providing a different definition for
 // yyFlexLexer, to define multiple lexer base classes.
-#ifndef yyFlexLexerOnce
+
 #define yyFlexLexer p4FlexLexer
 #include <FlexLexer.h>
-#endif
+#undef yyFlexLexer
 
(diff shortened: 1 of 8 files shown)


mbudiu-vmw (approved):
I hope this will work with all versions of flex/bison that we are supposed to support.

Issue #3451, opened 2022-07-20
Do recent changes to file p4parser.ypp require newer version of bison?


jafingerhut:
Starting with this commit https://github.com/p4lang/p4c/commit/82d3208f8e2f86f59ff1632907eacd1ab162862c attempting to build p4c using cmake on an Ubuntu 18.04 system fails with these error messages:

```
Scanning dependencies of target mkv1dirs
[ 18%] Built target mkv1dirs
Scanning dependencies of target mkp4dirs
[ 18%] Built target mkp4dirs
[ 19%] [BISON][v1Parser] Building parser with bison 3.0.4
parsers/v1/v1parser.ypp:24.9-24: error: %define variable 'api.parser.class' is not used
 %define api.parser.class {V1Parser}
         ^^^^^^^^^^^^^^^^
frontends/CMakeFiles/frontend.dir/build.make:85: recipe for target 'frontends/parsers/v1/v1parser.cpp' failed
make[2]: *** [frontends/parsers/v1/v1parser.cpp] Error 1
CMakeFiles/Makefile2:1738: recipe for target 'frontends/CMakeFiles/frontend.dir/all' failed
make[1]: *** [frontends/CMakeFiles/frontend.dir/all] Error 2
Makefile:162: recipe for target 'all' failed
make: *** [all] Error 2
```

This does not occur on an Ubuntu 20.04 system.  In both cases, I am using the latest binary packages released by Ubuntu for flex, bison, gcc, and many other dependencies.  Here are versions of these tools installed on the two systems I have tried on today:

Ubuntu 18.04
```
$ flex --version
flex 2.6.4
$ bison --version
bison (GNU Bison) 3.0.4
$ gcc --version
gcc (Ubuntu 7.5.0-3ubuntu1~18.04) 7.5.0
```

Ubuntu 20.04
```
$ flex --version
flex 2.6.4
$ bison --version
bison (GNU Bison) 3.5.1
$ gcc --version
gcc (Ubuntu 9.4.0-1ubuntu1~20.04.1) 9.4.0
```

These are the minimum versions required according to p4c's build scripts:
```
-- Found FLEX: /usr/bin/flex (found suitable version "2.6.4", minimum required is "2.0") 
-- Found BISON: /usr/bin/bison (found suitable version "3.0.4", minimum required is "3.0") 
```

I have not determined whether installing a newer version of bison on Ubuntu 18.04 would improve on this behavior.  @davidbolvansky Do you happen to know?

[...]

PR #3454, opened 2022-07-20
Fixed build with older bison on Ubuntu 18

davidbolvansky:
Part of recent commit changed (deprecated) parser_class_name to api.parser.class but unfortnutely this broke build on Ubuntu 18, as bison version there does not support new form. So for now use parser_class_name..

Fixes https://github.com/p4lang/p4c/issues/3451


mbudiu-vmw (approved):

jafingerhut:
I tried out these changes on an Ubuntu 18.04 system with bison v3.0.4 installed, and there were no compile-time errors.  Thanks.

PR #3455, opened 2022-07-20
Add a compiler pass that adds missing IDs to control plane objects

### LINKED ISSUE #3448: Add compiler pass that attaches missing P4Runtime IDs to appropiate control plane objects
The P4RuntimeSerializer [tries to generate](https://github.com/p4lang/p4c/blob/main/control-plane/p4RuntimeSerializer.cpp#L464) deterministic IDs for control plane objects if they are missing the `@id` annotation. However, this is done in a fairly ad-hoc fashion when generating the P4Info file. It would be nice to instead have this step as a compiler pass in the simple_switch mid end. That way it is also possible to persist IDs for a given P4 program.

fruffy:
This pull request moves the `SymbolTable` and other objects out of the `RuntimeSerializer` so that it can be used by dedicated compiler passes. In this case, a new compiler pass (`MissingIdAssigner`), which uses the symbol table to attach IDs to control plane IR objects that are missing an ID. This pass can be used to transform a P4 program to have full P4Runtime information without needing to parse the P4Info file or serialized protobuf objects. 

These changes should not affect the normal control-plane behavior, but the computed IDs should be identical to the IDs computed by the serializer. To make the behavior identical, it is unfortunately necessary to use a dedicated evaluator pass within the ID assigner, which is redundant.

This pass can be used by any target back end right as first step after the front end. The only inputs needed are the ref and type map and the corresponding architecture handler.

Looking for some feedback on how to improve the structure of the pass. Because of the way the symbol table is constructed, it is necessary to compute it directly in the `IR::Program` visitor step, which is a little awkward. Also, I could try to refactor the entire serializer to expect IDs at all times and each back end has to run the `MissingIdAssigner` first, but this might be too invasive of a change. 

Fixes #3448.


control-plane/p4RuntimeSymbolTable.cpp
```diff
@@ -0,0 +1,400 @@
+/*
```
mbudiu-vmw, line 1: I will assume that definitions in this file are just moved from the other source.

control-plane/p4RuntimeSymbolTable.cpp
```diff
@@ -0,0 +1,400 @@
...
+    forAllMatching<IR::MethodCallExpression>(
+        control->body, [&](const IR::MethodCallExpression* call) {
+            auto* instance = P4::MethodInstance::resolve(call, refMap, typeMap);
+            if (instance->is<P4::ExternFunction>()) {
+                archHandler->collectExternFunction(
+                    &symbols, instance->to<P4::ExternFunction>());
+            }
+        });
+}
+void collectExternSymbols(P4RuntimeSymbolTable& symbols,
```
mbudiu-vmw, line 126: we usually leave an empty line between function definitions

control-plane/addMissingIds.h
```diff
@@ -0,0 +1,68 @@
...
+#define CONTROL_PLANE_ADDMISSINGIDS_H_
+
+#include "frontends/common/resolveReferences/resolveReferences.h"
+#include "frontends/p4/typeChecking/typeChecker.h"
+#include "ir/ir.h"
+#include "p4RuntimeSymbolTable.h"
+
+namespace P4 {
+
+/// Scans the P4 program, run the evaluator pass, and derives the P4Runtime Ids
```
mbudiu-vmw, line 12: Should this comment apply to the PassManager below?

control-plane/addMissingIds.h
```diff
@@ -0,0 +1,68 @@
...
+    /// The symbol table that contains all the ID information.
+    /// This ID assigner first computes this table,
+    /// then uses the results to assign missing IDs.
+    const ControlPlaneAPI::P4RuntimeSymbolTable* symbols = nullptr;
+
+    /// The arch builder is necessary to compute the correct symbol table for a
+    /// particular architecture.
+    const ControlPlaneAPI::P4RuntimeArchHandlerBuilderIface& archBuilder;
+
+    const IR::P4Program* preorder(IR::P4Program* program) override;
```
mbudiu-vmw, line 38: Here however empty lines are not useful.

control-plane/addMissingIds.h
```diff
@@ -0,0 +1,68 @@
...
+    const IR::P4Table* preorder(IR::P4Table* table) override;
+
+    const IR::Type_Header* preorder(IR::Type_Header* hdr) override;
+
+    const IR::P4ValueSet* preorder(IR::P4ValueSet* valueSet) override;
+
+    const IR::P4Action* preorder(IR::P4Action* action) override;
+
+ public:
+    explicit MissingIdAssigner(
```
mbudiu-vmw, line 49: I think explicit is required only if you have a single constructor argument - both here and below.

control-plane/addMissingIds.cpp
```diff
@@ -0,0 +1,214 @@
...
+
+AddMissingIdAnnotations::AddMissingIdAnnotations(
+    ReferenceMap* refMap, TypeMap* typeMap,
+    const ControlPlaneAPI::P4RuntimeArchHandlerBuilderIface* archBuilder) {
+    CHECK_NULL(refMap);
+    CHECK_NULL(typeMap);
+
+    passes.push_back(new ControlPlaneAPI::ParseP4RuntimeAnnotations());
+    passes.push_back(new MissingIdAssigner(refMap, typeMap, *archBuilder));
+    // Need to refresh types after we have updated some expressions with
```
mbudiu-vmw, line 208: the rule we use is the opposite: if a pass needs types, it should run first typeChecking.
So this should not be necessary - whatever comes after will do it if needed.

control-plane/addMissingIds.cpp
```diff
@@ -0,0 +1,214 @@
+#include "addMissingIds.h"
+
+#include "frontends/p4/evaluator/evaluator.h"
+#include "p4RuntimeAnnotations.h"
+#include "p4RuntimeArchStandard.h"
+
+namespace P4 {
+
+const IR::P4Program* MissingIdAssigner::preorder(IR::P4Program* program) {
+    auto evaluator =  P4::EvaluatorPass(refMap, typeMap);
```
mbudiu-vmw, line 10: why hide the evaluator in this pass instead of running it from the PassManager?

control-plane/addMissingIds.cpp
```diff
@@ -0,0 +1,214 @@
...
+    program->apply(evaluator);
+    auto* toplevel = evaluator.getToplevelBlock();
+    CHECK_NULL(toplevel);
+    symbols = ControlPlaneAPI::P4RuntimeSymbolTable::generateSymbols(
+        toplevel->getProgram(), toplevel, refMap, typeMap,
+        archBuilder(refMap, typeMap, toplevel));
+    auto objectVector = IR::Vector<IR::Node>();
+    for (const auto* obj : program->objects) {
+        objectVector.push_back(obj->apply(
```
mbudiu-vmw, line 19: this seems unnecessary - calling the visitor on the program should do exactly this.
You are also allocating a new visitor for each program object, that sounds wasteful.

control-plane/addMissingIds.cpp
```diff
@@ -0,0 +1,214 @@
...
+            ControlPlaneAPI::P4RuntimeSymbolType::TABLE(), table);
+        const auto* idConst =
+            new IR::Constant(new IR::Type_Bits(ID_BIT_WIDTH, false), symbolId);
+        annoExprs.push_back(idConst);
+        newAnnos->add(new IR::Annotation("id", annoExprs));
+        table->annotations = newAnnos;
+    }
+    size_t propertyIdx = 0;
+
+    const IR::Key* key = nullptr;
```
mbudiu-vmw, line 46: you are reconstructing here the logic of the visitor, why not just make a postorder function for a table property?

control-plane/addMissingIds.cpp
```diff
@@ -0,0 +1,214 @@
...
+        if (property->name == "key") {
+            key = property->value->checkedTo<IR::Key>();
+            break;
+        }
+    }
+    if (key == nullptr) {
+        return table;
+    }
+    auto* newKey = key->clone();
+    IR::Vector<IR::KeyElement> newKeys;
```
mbudiu-vmw, line 59: same here, you can make this as a postorder function for a key

mbudiu-vmw (changes requested):

mbudiu-vmw:
However, I would heed @antoninbas' comments carefully. I would really prefer if he approved the design.

control-plane/addMissingIds.cpp
fruffy: If I do not compute this in this specific order, the evaluator will return a nullptr for `getTopLevelBlock`. When I move the evaluator in the pass manager the reference to the top level block is not set yet. 

It is a little awkward, the other way to solve this is to make the Symboltable mutable and move a reference to it around.

control-plane/addMissingIds.cpp
fruffy: I was getting a segmentation fault for using *this. Let me try again.

control-plane/addMissingIds.cpp
fruffy: I think I can't since I need to keep track of the individual elements to assign the appropriate IDs.

control-plane/p4RuntimeSymbolTable.cpp
fruffy: Yes, I only added a new function call.

control-plane/addMissingIds.cpp
fruffy: I would like to not use the evaluator at all, the only reason I currently use it is to preserve the exact functionality as in `p4RuntimeSerializer.cpp`.

mbudiu-vmw (approved):

antoninbas (commented):
This is fine by me as it seems to just be about re-organizing some code, and the P4RuntimeSerializer functionality is not impacted.

However, I don't understand the changes to `backends/bmv2/`. You are adding 2 include statements to `backends/bmv2/simple_switch/midend.cpp`, but no other code, so I am not sure why this change is for.  I am not clear about the PR description either (`Add a BMv2 pass that adds missing IDs to control plane objects`). This change doesn't really seem related to bmv2.

fruffy:
This pass was first added to the `simple_switch` mid end to test it, It was also meant to replace the ID assignment in the runtime back end. I removed the leftover code. 

Add this pass to other mid ends and removing the ID assignment portion in the serializer is a further TODO.

PR #3458, opened 2022-07-22
Fixes related to type and function specialization

### LINKED ISSUE #3291: Compiler bug with struct<struct<bit>> where struct is the same struct but using a typedef
Take:
```
struct h<t>{
  t f;
}

typedef h<bit> tt;
typedef h<tt> t;
```
This produces:
```
In file: /home/apinski/src/p4/p4c/ir/visitor.h:351
Compiler Bug: IR loop detected
```

If instead of the two typedefs you do:
```
typedef h<h<bit>> t1;
```
You get:
```
header.p4(5): [--Werror=type-error] error: h: recursive type specialization
typedef h<h<bit>> t1;
          ^
```
But I don't see how this is not allowed even with the approved section in the last language design meeting.

mbudiu-vmw:
Fixes #3291 
The fix is in the typechecking code.
But the tests revealed some bugs in the specialization code: now we insert the specialized function code in a different place in the source code, to make sure it's after all type declarations that it may depend on. So the function specialization code is more similar to the existing type specialization code.

### DIFF
--- a/frontends/p4/def_use.cpp
+++ b/frontends/p4/def_use.cpp
@@ -103,6 +103,8 @@ StorageLocation* StorageFactory::create(const IR::Type* type, cstring name) cons
         result->setLastIndexField(create(IR::Type_Bits::get(32), name + "." + indexFieldName));
         return result;
     }
+    BUG_CHECK(!type->is<IR::Type_SpecializedCanonical>(),
+              "%1%: unspecialized type in def-use", type);
     return nullptr;
 }
 
--- a/frontends/p4/def_use.cpp
+++ b/frontends/p4/def_use.cpp
@@ -103,8 +103,6 @@ StorageLocation* StorageFactory::create(const IR::Type* type, cstring name) cons
         result->setLastIndexField(create(IR::Type_Bits::get(32), name + "." + indexFieldName));
         return result;
     }
-    BUG_CHECK(!type->is<IR::Type_SpecializedCanonical>(),
-              "%1%: unspecialized type in def-use", type);
     return nullptr;
 }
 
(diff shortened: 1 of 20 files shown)


mbudiu-vmw:
@fruffy It looks like gauntlet does not like one of the new tests.

fruffy:
Gauntlet complains about this transformation:

```p4
bool g_0(in h_0 a) {
    @name("v") h<h_0> v_0;
    v_0.f = a;
    return v_0.f == a;
}
bool gg() {
    bool tmp;
    @name("a") h_0 a_0;
    a_0 = (h_0){f = 1w0};
    tmp = g_0(a_0);
    return tmp;
}
control c(out bool x) {
    apply {
        x = gg();
    }
}
```
to

```p4
bool g_0(in h_0 a) {
    @name("v") h<h_0> v_0;
    ;
    return v_0.f == a;
}
bool gg() {
    bool tmp;
    @name("a") h_0 a_0;
    a_0 = (h_0){f = 1w0};
    tmp = g_0(a_0);
    return tmp;
}
control c(out bool x) {
    apply {
        x = gg();
    }
}
```

Remove the `v_0.f = a;` does not seem correct to me, unless I am missing something. Since v_0 is not initialized.

mbudiu-vmw:
Gauntlet seems right, this looks like a bug, but it's probably somewhere else.

mbudiu-vmw:
@fruffy indeed, Gauntlet was right. I have fixed that bug, and now this is ready for review.
There are in fact several bug fixes in this PR.
In particular, specialization of types and functions has to be iterated, because a generic function may use generic types to call other generic functions. This caused a bug in def-use analysis, which does not understand generic types, which cased the incorrect code removal that Gauntled complained about. There is now a sanity check about this.

backends/ebpf/targets/xdp_target.py
```diff
@@ -0,0 +1 @@
+/home/mbudiu/git/p4c/extensions/p4c-xdp/xdp_target.py
```
fruffy, line 1: There must be a way to ignore symlinks in specific folders. Let me check.

fruffy (approved):

PR #3463, opened 2022-07-29
Dpdk backend : Flatten header union individual elements

usha1830:
The header union should be flattened into its individual elements for DPDK.
For ex:

```
  header_union U {
      Hdr1 h1;
      Hdr2 h2;
  }
 
  struct Headers {
      Hdr1 h1;
      U u;
  }
  
  is replaced by 
  struct  Headers {
      Hdr1 h1;
      Hdr1 u_h1;
      Hdr2 u_h2;
  }
```

Some of the v1model tests involving header union are ported to PNA for testing the changes in this PR.
This PR does not handle header union stacks. I will work on it in a follow-up PR.


backends/dpdk/dpdkArch.h
```diff
@@ -987,6 +988,56 @@ class ElimHeaderCopy : public Transform {
...
+ *     Hdr2 h2;
+ * }
+ *
+ * struct Headers {
+ *     Hdr1 h1;
+ *     U u;
+ * }
+ *
+ * is replaced by
+ * struct  Headers {
```
mbudiu-vmw, line 1005: What is the union is a standalone variable?
Just doing this change won't preserve accurately the semantics of unions.
In particular, setting a union field to valid should set all other fields to invalid.

backends/dpdk/dpdkArch.cpp
```diff
@@ -2448,4 +2446,86 @@ const IR::Node* ElimHeaderCopy::postorder(IR::Member* m) {
...
+             toInsert.push_back(new IR::Declaration_Variable(IR::ID(uName), uType));
+        }
+    }
+    return dv;
+}
+
+const IR::Node* DoFlattenHeaderUnion::postorder(IR::Member* m) {
+    if (m->expr->type->to<IR::Type_HeaderUnion>()) {
+        if (auto huf = m->expr->to<IR::Member>()) {
+            return new IR::Member(huf->expr,IR::ID(huf->member.name+"_"+m->member.name));
```
mbudiu-vmw, line 2481: this is not guaranteed to generate a field name that does not exist.
the safer way would be to use the NameGenerator.
Then you will have to carry a map of the renamed fields around.

backends/dpdk/dpdkArch.cpp
```diff
@@ -2448,4 +2446,86 @@ const IR::Node* ElimHeaderCopy::postorder(IR::Member* m) {
...
+    return new IR::Type_Struct(s->name, s->annotations, fields);
+}
+
+const IR::Node* DoFlattenHeaderUnion::postorder(IR::Declaration_Variable *dv) {
+    auto ftype = typeMap->getTypeType(dv->type, true);
+    if (ftype->is<IR::Type_HeaderUnion>()) {
+        for (auto sfu : ftype->to<IR::Type_HeaderUnion>()->fields) {
+             cstring uName = dv->name.name + "_" + sfu->name.name;
+             auto uType =  sfu->type->getP4Type();
+             toInsert.push_back(new IR::Declaration_Variable(IR::ID(uName), uType));
```
mbudiu-vmw, line 2472: there could already be variables with such a name.

mbudiu-vmw (changes requested):

backends/dpdk/dpdkArch.cpp
usha1830: Ok. I will update the code to use name generator/refmap.

backends/dpdk/dpdkArch.h
usha1830: >What is the union is a standalone variable?

Standalone union variable should be handled by the "Declaration_Variable" visitor. Do you see any other possibilities?

Thanks for pointing the semantics w.r.t validity bit of header unions. So, every assignment to a header union element must be converted to a sequence of statements involving validity bit modifications.
Going over the specification again, I have found the following statements which need to be handled. Please let me know if anything is missing.

```
1) U u;
H1 my_h1 = { 8w0 }; 
u.h1 = my_h1;       
translates to
 if (my_h1.isValid()) {
     u_h1.setValid();
	 ///all other elements of union set to invalid
	 u_hi = my_h1;
 } else{
     u_h1.setInvalid();
 }
 
2) u.h1 = { 16w1}
translates to
u_h1.setValid();
  ///all other elements of union set to invalid
   u_h1 = {16w1};
   
3) u.h1.setValid()
translates to 
u_h1.setValid();
  ///all other elements of union set to invalid
  
4) U u1, u2;
   u1 = u2 
translates to
if (u2.isValid){
    for elements in u1{
       if element is valid{
	     copy element and set validity bit
		 set all other elements invalid and break;	   
	   }	      
	}
} else {
    ///all elements of union set to invalid
}

5) Comparison of header unions
if (u == u1)
...
translates to
tmp = u == u1;
if(u.isInvalid && u1.isInvalid)
    tmp = 1;
else if ((u.isValid && u1.isInvalid) || (u.isInvalid && u1.isValid))
    tmp = 0;
else {
//both are valid, check individual fields
//in both unions the same field is valid, and the values of the valid fields are equal as headers).
   tmp = 1;
}
if (tmp)
....
```

mbudiu-vmw:
This pass you are writing could be generally useful, I would recommend making it into a midend pass which converts all uses of unions to structs. Then it could be reused for other backends.

usha1830:
Opened a new PR for moving this to midend and handling the validity the semantics https://github.com/p4lang/p4c/pull/3524.
Closing this one.

PR #3464, opened 2022-07-29
Dpdk Backend: Move Mux handling after simplifying select expression

usha1830:
Comparison operations on RHS of assignment are translated into Mux expressions. Mux expressions are further simplified into if-else statement by the dpdk backend
Mux expression handling in backend works on Assignment statement assuming the Mux expressions are simplified into RHS of Assignment statements.

When statements like below are present in parser state, copy propagation moves the assignment statement inside Select expression which is not simplified when control reaches Mux handling. 
```
bool var = a == b;
select(var)
```

Moving DismantleMuxExpressions after RemoveComplexExpressions (which simplifies Select Expressions) solves this issue.


mbudiu-vmw (approved):

mbudiu-vmw:
I don't understand why the MacOS build error happened.

PR #3469, opened 2022-08-01
Implicit casts between InfInt and Bit are allowed

### LINKED ISSUE #3238: "const tuple<int> t = {1};" valid?
Take:
```
const tuple<int> t = {1};
bit func()
{
  return t[0];
}
```
According to the "Type nesting rules", tuple<int> is not a valid type but the reference compiler allows it for const declarations. Are the rules for const decl/types different from the normal types?

This ties in with https://github.com/p4lang/p4-spec/issues/1062 really.

mbudiu-vmw:
Fixes #3238


jnfoster (approved):
LGTM. And consistent with the spec.

PR #3473, opened 2022-08-03
Dpdk Backend: Allow non-constant arguments to mirror_packet extern

usha1830:
Removed the checks for allowing only constant arguments to mirror_packet extern.
The "mirror" instruction in DPDK assembly expects both the arguments in metadata. Hence, move instruction is inserted to move the argument to metadata if not already present in metadata


usha1830:
Mac OS checks are cancelled for unknown reason!!

mbudiu-vmw (approved):

usha1830:
> Mac OS checks are cancelled for unknown reason!!

@mbudiu-vmw Thanks for the review!. Now the Mac OS checks are passing. Could you please merge this?

PR #3474, opened 2022-08-03
Remove GMP dependency from BUILD.bazel

kheradmandG:


mbudiu-vmw (approved):
This looks fine to me, but this gmp.h stuff is really a choice for each "customer".
I don't have a problem if all customers who use bazel do not use gmp.

kheradmandG:
Thanks for taking a look. As you mentioned, the correct way of handling this is to make the dependency on GMP parametric but that requires quite some effort. It is fine to keep it like before, although we (Google) will remove the dependency internally and my guess is that we are the main consumer of that bazel build file.

smolkaj:
Agreed with all points.

In the meantime, flipping the default from GMP to boost seems okay?
It is arguably the better default, since it removes the dependency on GMP and we already depend on Boost.

mbudiu-vmw:
Sure, go ahead

kheradmandG:
Nice! Though I don't have a write access so somebody with a write access should merge this (and #3475).

mbudiu-vmw:
Something's weird with the MacOS build in CI. Once I triggered it manually it seems to work, but otherwise all MacOS jobs show up as cancelled. I will merge once these jobs finish, they only build the software, so they shouldn't take too long.

kheradmandG:
Thanks!

PR #3480, opened 2022-08-05
Do not reset header valid bits when inlining a subparser

bleibig:
This operation seems to be unnecessary and mainly just increases the number of fields that have to be dealt with in a backend. This PR removes the calls to the pass that inserts the `setInvalid()` calls, and modifies the expected test outputs accordingly.


mbudiu-vmw:
Why is this operation unnecessary?
According to the spec this should happen: https://p4.org/wp-content/uploads/2022/07/P4-16-spec.html#sec-calling-convention

jafingerhut:
It has been a few years, so I don't recall exactly, but I thought there was at least one test case with an STF test for the BMv2 software switch that was intended to catch this kind of omission in the compiler.  I didn't notice any tests failing for this PR, though.

mbudiu-vmw:
I will close this PR, feel free to reopen if you think it is actually correct.

bleibig:
Yes, after reexamining the spec, the compiler is doing the correct thing so this can be closed.

PR #3481, opened 2022-08-05
default default_action should be 'const'

mbudiu-vmw:
If we take https://github.com/p4lang/p4-spec/pull/1126 we have to take this one too.
Note that this may break user programs.


mbudiu-vmw:
We should change the spec instead of changing the compiler.

PR #3485, opened 2022-08-13
Remove GMP support. 

### LINKED ISSUE #3462: Consider dropping the GMP implementation?
Is there a particular reason why the GMP library is still included with the compiler? As far as I know, boost::multiprecision was added because of licensing issues (#1418) but it also seems to cover all necessary use cases. Removing GMP could simplify the build and also the CI runs.

fruffy:
This PR removes libgmp from the compiler as dependency. boost multiprecision handles all the use cases and is already included in the boost main dependencies. This also helps with reducing the number of CI runs. 

This might be a little to aggressive. Maybe just setting the default to boost and remove gmp from CI might be enough. Fixes #3462.


fruffy:
Any opinions on this PR? I would like to cut down on the number of tests that we run on each commit such that we do not spend as much time waiting for runners to be dequeued. 

GMP also does not appear to be used much in the compiler considering the licensing constraints. The boost library handles all the use cases. However, there may be private back ends that depend on it.

pkotikal (approved):

davidbolvansky:
+1

fruffy:
@smolkaj @kheradmandG Are there any constraints from your side here?

kheradmandG:
We don't use GMP anymore so this should be fine (we also cannot depend on Boost versions higher than 1.67 at the moment but this is not that problematic here).

ChrisDodd:
Making boost the default would certainly be ok (in fact, I thought i was the default), but I though at one point there was concern that boost::mulitprecision was significantly slower than GMP?  At Intel we always use boost rather than GMP for licensing, but others may prefer differently.

mbudiu-vmw (approved):

mbudiu-vmw:
Why not remove the enable gmp flags altogether?

fruffy:
I believe I did do this, am I missing any? Unfortunately, libgmp-dev still needs to be installed because the behavioral model is built with GMP. 

Also this PR can not be merged because it is blocked on expected CI runs (which have been removed). Are you able to untick them from the branch restrictions?

fruffy:
@jnfoster This PR can not be merged because some CI tests were removed but Github's CI still expect results from them. Are you able to remove these expected tests from the branch protection rules?

jnfoster:
I can force merge this. Shall I?

fruffy:
Yeah we can give this a try? https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/defining-the-mergeability-of-pull-requests/about-protected-branches#require-status-checks-before-merging 
Either way, after merging the required status checks _should_ update...

jnfoster (approved):

jnfoster:
I have overridden the branch protection rules for this one.

Issue #3486, opened 2022-08-16
Compiler bug: varbit<(8*anyIntegerHere)> causes error


kitlei-robert:
This code

```
header varbit_t {
    varbit<(8*12)> f1; // 12 bytes of varbit
}
```

worked fine previously, but now (tested with a p4c commit of 2022-08-01) causes an error like this:

```
test-varbit-2.p4(9): [--Werror=invalid] error: varbit<96>: invalid type size
    varbit<(8*12)> f1;
    ^^^^^^^^^^^^^^
```

Indeed, any natural number in `varbit<(8*???)>` causes an equivalent error.

The equivalent

```
header varbit_t {
    varbit<96> f1;
}
```

compiles without any issue.

Possible cause: in [this commit](https://github.com/p4lang/p4c/commit/f26b26e7c6bb4ba38d76fdf4db596e0ac54d2d1f), `type->width_bits()` is used to check the size of the varbit type, and I suspect that in case it is the result of an expression, that evaluates to `0` instead of the correct value of `96`.

PR #3487, opened 2022-08-16
Fix bug related to varbits width

mbudiu-vmw:


mbudiu-vmw:
Fixes #3486

fruffy (commented):
Will zero width bits not have the same problem? For example, with an expression such as `bit<1*0>`

mbudiu-vmw:
Perhaps I should change the test to < 0 from <= 0.

fruffy (approved):

PR #3489, opened 2022-08-17
Delete keys that are empty tuples

mbudiu-vmw:

This brings the implementation in line with the spec: https://github.com/p4lang/p4-spec/pull/1132


jafingerhut (approved):
LGTM.  I tested earlier and so no difference between explicit `key = {}` tables and those without a `key` property at all, but this change certainly looks like it should guarantee that it remains this way.  Thanks.

PR #3490, opened 2022-08-18
Fix incorrect CFG generated for BMV2 backend for switch statements

### LINKED ISSUE #3488: Compiler bug: Fall-through switch case statement broken
I am hitting what seems like a basic issue with a fall-through action in a P4 switch statement.
I have a switch statement with a fall-through case, like so with the fall-through action at line number 105.

```
100 control aaa {
101   apply {  
102        ... 
103        meta.xxx = hdr.yyy;
104        switch (table1.apply().action_run) {
105            actionA:                         // fall-through
106            actionB: {
107                table2.apply();
108             }
109         }
110        table3.apply();
111     }
112 }
```

the json generated skips table 1, table 2 and directly jumps to table3

```
        {
          "name" : "tbl_aaa103",
          "source_info" : {
            "filename" : "aaa.p4",
            "line" : 103,
            "column" : 10,
            "source_fragment" : "= hdr.yyy; ..."
          },
          ... 
          "base_default_next" : "table3",
          "next_tables" : {
            "aaa103" : "table3"
          },

```

whereas if I change the code to remove just the fall-through action without any other changes

```
100 control aaa {
101   apply {  
102        ... 
103        meta.xxx = hdr.yyy;
104        switch (table1.apply().action_run) {
106            actionB: {
107                table2.apply();
108             }
109         }
110        table3.apply();
111     }
112 }
```

now the generated json properly points to table1 as the next stage.  

```
        {
          "name" : "tbl_aaa103",
          "source_info" : {
            "filename" : "aaa.p4",
            "line" : 103,
            "column" : 10,
            "source_fragment" : "= hdr.yyy; ..."
          },
          ... 
          "base_default_next" : "table1",
          "next_tables" : {
            "aaa103" : "table1"
          },
```

I am invoking p4c from the p4lang/p4c:latest docker image.

mbudiu-vmw:
Fixes #3488
But we should also add @jafingerhut's STF test when available.


fruffy:
This was an interesting case so I tried to see what the tools can do. 

I was not able to reproduce this with Gauntlet because the bug is actually not in a compiler pass but in the binary generation phase. We do not have any means to validate back end binaries, but this is something that could be very useful. 

I was however able to reproduce with testgen. Attached is another failing test case on top of Andy's.
The trace expects `test_ingress.pre_tbl1` to be executed, but this does not happen. 

[issue3488-bmv2.stf.txt](https://github.com/p4lang/p4c/files/9369789/issue3488-bmv2.stf.txt)

jfingerh:
I have added a slightly different P4 program than the one in the original report (to add an easily observable side effect in the default action of one of the tables that was not being executed due to the bug), and an STF test for it that fails using the latest released p4c, but should pass if the bmv2 back end was correct, to the issue 3488.

mbudiu-vmw:
I have added the STF tests from @jafingerhut and @fruffy . They pass with this new change. Thank you.
This was really a "nondeterministic" problem: the CFG had an extra edge, but if it wasn't the first in a collection, it would not affect the emitted CFG... that's why we never saw the bug before.

jafingerhut (approved):
LGTM.  As usual, I read over the test cases and their expected output only.  I don't have the knowledge to comment on the code changes in the compiler.

PR #3491, opened 2022-08-18
Deprecate unified build in favor of unity build.

fruffy:
The custom unified build code might actually not be needed anymore. CMAKE has added [UNITY_BUILDS](https://cmake.org/cmake/help/latest/prop_tgt/UNITY_BUILD.html), which seem equivalent. 

I left the library in the cmake folder, in case downstream repositories still needed. There is only one build testing that the non-unity build compiles, which reduces the amount of CI runs. I can also make unity builds the default if requested.


fruffy:
This requires fixes on some downstream repositories but otherwise works well. We may even drop the CI option to simplify tests.

mbudiu-vmw:
Does this work with any version of cmake?
Is it faster or comparable to the old unified build?

fruffy:
It works with CMake 3.16+, wich is the default since Ubuntu 20.04. The build times are roughly the same, but can be tweaked with `UNITY_BUILD_BATCH_SIZE`.

mbudiu-vmw:
I personally don't have a problem with this, but I think this should work on as many platforms as possible.
If this precludes people running older versions of the software from building, then it could be problematic.

fruffy:
I guess it boils down how we want to maintain backwards compatibility. We dropped support for anything lower than Ubuntu 20.04 with [this](https://github.com/p4lang/third-party/issues/25) pull request. So it might make sense to use features that are provided by it. 

The original unified_build library is still there. I can try to create a version that uses it depending on the CMake version.

jfingerh:
"We dropped support for anything lower than Ubuntu 20.04 with https://github.com/p4lang/third-party/issues/25 pull request."

Techically, we stopped doing automated CI testing for anything lower than Ubuntu 20.04 with that change.

There was a commit in the last month or two that broke the build for Ubuntu 18.04, which when pointed out later via an issue then caused someone to create another PR that restored a working build for Ubuntu 18.04.

jfingerh:
I agree that at some point it definitely makes sense not to support Ubuntu 18.04.  A reasonable question is: when?  One choice is right now.  Another would be to wait until 5 years after Ubuntu 18.04 was released in April 2023, when Ubuntu discontinues free support for it.

fruffy:
The problem is that without CI we can not make any guarantees for 18.04. It will be quite hard to respect 18.04 restrictions going forward. This is related to the discussion in https://github.com/p4lang/p4c/issues/3451. Iirc that change worked because there was a solution that works for both 18.04 and future systems. This might not always be possible. 

For this particular PR, I can try to come up with a solution that can also work with older CMake systems.

davidbolvansky:
> It works with CMake 3.16+, wich is the default since Ubuntu 20.04. The build times are roughly the same, but can be tweaked with `UNITY_BUILD_BATCH_SIZE`.

What is the plan now? Land this PR after April 2023? Or just land it now and one (on U18) can install newer cmake easily for example with pip install cmake==3.16.3.

fruffy:
> What is the plan now? Land this PR after April 2023? Or just land it now and one (on U18) can install newer cmake easily for example with pip install cmake==3.16.3.

I will have to revive this PR but I am not quite sure what to do with `build_unified` yet. I will have to implement an equivalent version so that downstream projects can still use it. It might be nicer if that explicit function were not necessary.

jfingerh:
I have no objections to merging updates that require specific later versions of cmake on Ubuntu 18.04, as long as there are clear README directions on building on Ubuntu 18.04 somewhere.

davidbolvansky:
> build_unified

I would remove it. 

Could you create new cmake option so we could try new cmake native unified build downstream and do necessary changes?

CMakeLists.txt
```diff
@@ -83,13 +83,17 @@ else()
...
 endif()
 include(P4CUtils)
-include(UnifiedBuild)
 
 # # search in /usr/local first
 # set (CMAKE_FIND_ROOT_PATH "/usr/local/bin" "${CMAKE_FIND_ROOT_PATH}")
 set (P4C_CXX_FLAGS "")
 set (P4C_LIB_DEPS)
 
+
```
davidbolvansky, line 92: if (ENABLE_UNIFIED_COMPILATION)
set(CMAKE_UNITY_BUILD ON)

backends/p4tools/common/lib/timer.cpp
```diff
@@ -50,52 +50,50 @@ struct RootCounter {
...
-        auto duration = Clock::now() - start_time;
+        auto duration = Clock::now() - startTime;
         self->add(duration);
         // Restore previous counter as current.
         RootCounter::get().setCurrent(parent);
     }
 };
 
-ScopedTimer::ScopedTimer(const char* name) : ctx(new ScopedTimer::Ctx(name)) {}
+ScopedTimer::ScopedTimer(const char* name) : ctx(new ScopedTimerCtx(name)) {}
```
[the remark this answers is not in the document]
fruffy: https://github.com/p4lang/p4c/pull/3785

CMakeLists.txt
```diff
@@ -90,6 +92,11 @@ set (P4C_CXX_FLAGS "")
 set (P4C_LIB_DEPS)
 
 
+# Support a legacy option.
+if (ENABLE_UNIFIED_COMPILATION)
```
davidbolvansky, line 96: Feel free to emit an warning here that this option is deprecated and it will be removed after some time.

davidbolvansky (commented):
cmake_minimum_required / README  needs to be updated

.github/workflows/ci-p4tools.yml
```diff
@@ -27,12 +27,12 @@ jobs:
...
       uses: hendrikmuhs/ccache-action@v1
       with:
-        key: test-tools-${{ matrix.unified }}-${{ runner.os }}
+        key: test-tools-${{ runner.os }}
         max-size: 1000M
 
     - name: Build (Ubuntu Linux)
       run: |
-        docker build -t p4c --build-arg IMAGE_TYPE=test --build-arg ENABLE_UNIFIED_COMPILATION=ON --build-arg ENABLE_TEST_TOOLS=ON .
+        docker build -t p4c --build-arg IMAGE_TYPE=test --build-arg CMAKE_UNITY_BUILD=ON --build-arg ENABLE_TEST_TOOLS=ON .
```
rst0git, line 35: It looks like this pull request may have merge conflicts with https://github.com/p4lang/p4c/pull/3770
Should we merge the tests for Ubuntu 20.04 and 22.04 first?

davidbolvansky: Pushed. Please rebase.

davidbolvansky:
For now please pip install cmake 3.16.3 here:
https://github.com/p4lang/p4c/blob/main/tools/ci-build.sh#L98

tools/ci-build.sh
```diff
@@ -107,6 +107,12 @@ apt-get install -y --no-install-recommends \
   ${P4C_EBPF_DEPS} \
   ${P4C_RUNTIME_DEPS}
 
+# For Ubuntu 18.04, remove the Debian package and install the pypi-supplied version instead.
+if [[ "${DISTRIB_RELEASE}" == "18.04" ]] ; then
+apt remove --purge -y cmake
+pip3 install cmake==3.16.3
+fi
```
rst0git, line 114: By default the Ubuntu 18.04 container image does not have `cmake` installed. It might be better to remove `cmake` from `P4C_DEPS` above and use "if / else" statement like:
```bash
if [[ "${DISTRIB_RELEASE}" == "18.04" ]] ; then
    # For Ubuntu 18.04, remove the Debian package and install the pypi-supplied version instead.
    P4C_PIP_PACKAGES+=" cmake==3.16.3"
else
    P4C_DEPS+=" cmake"
fi
```

.github/workflows/ci-validation.yml
```diff
@@ -35,9 +35,14 @@ jobs:
...
-        (cd /p4c/ && sudo -E tools/ci-build.sh)
-        sudo cp -rf /p4c/.ccache .
+        (cd /p4c/ && sudo -E tools/ci-build.sh) || echo "BUILD_SUCCESS=false" >> $GITHUB_ENV
 
     - name: Validate
-      run: sudo -E ctest  -R toz3-validate-p4c --output-on-failure --schedule-random
-      working-directory: /p4c/build
+      if: env.BUILD_SUCCESS == 'true'
+      run: |
+          docker run --privileged -w /p4c/build -e $CTEST_PARALLEL_LEVEL p4c ctest  -R toz3-validate-p4c --output-on-failure --schedule-random
```
davidbolvansky, line 43: Are you sure? Why docker again?

fruffy: Something broke with this file while rebasing. Will fix.

.github/workflows/ci-validation.yml
```diff
@@ -35,9 +35,15 @@ jobs:
     - name: Build (Ubuntu 20.04)
       run: |
         sudo cp -rf . /p4c/
-        (cd /p4c/ && sudo -E tools/ci-build.sh)
+        (cd /p4c/ && sudo -E tools/ci-build.sh) || echo "BUILD_SUCCESS=false" >> $GITHUB_ENV
         sudo cp -rf /p4c/.ccache .
 
     - name: Validate
-      run: sudo -E ctest  -R toz3-validate-p4c --output-on-failure --schedule-random
-      working-directory: /p4c/build
```
davidbolvansky, line 44: You need this workdir

.github/workflows/ci-validation.yml
```diff
@@ -35,9 +35,15 @@ jobs:
...
+      if: env.BUILD_SUCCESS == 'true'
+      run: |
+          ctest  -R toz3-validate-p4c --output-on-failure --schedule-random
```
davidbolvansky, line 44: Will it run? dont you need sudo -E?

i am suprised why you need any changes at all in this file

davidbolvansky: Some hours ago when I was looking why validate fails it was due to some leftover of removed build_unified.

davidbolvansky: Try to avoid any changes here and see what CI says

fruffy: Yes it should run without sudo. `sudo` is only required for eBPF kernel tests.  These changes need to be preserved to avoid build failures of Gauntlet affecting PRs. This is because Gauntlet is a downstream dependency.

davidbolvansky: But this is still broken. Validate Step ends in Build failed (but build went fine, test image was created so strange…. what is worse, it is green, so easy to overlook any issues - I changed it in my previous PR but I am not sure why you reverted those changes).

.github/workflows/ci-test.yml
```diff
@@ -206,7 +202,7 @@ jobs:
 
     - name: Build (MacOS)
       run: |
-          ./bootstrap.sh -DENABLE_GC=ON -DCMAKE_BUILD_TYPE=RELEASE -DENABLE_UNIFIED_COMPILATION=ON && cd build && make -j2
+          ./bootstrap.sh -DENABLE_GC=ON -DCMAKE_BUILD_TYPE=RELEASE -DCMAKE_UNITY_BUILD=ON -DENABLE_GMP=${{ matrix.enable_gmp }} && cd build && make -j2
```
davidbolvansky, line 205: GMP is removed

.github/workflows/ci-validation.yml
fruffy: Yes, that is intentional. Failed builds should not show up in PRs since they can only be fixed in the downstream repository. 

I will fix the odd skipping.

.github/workflows/ci-validation.yml
```diff
@@ -35,9 +35,14 @@ jobs:
     - name: Build (Ubuntu 20.04)
       run: |
         sudo cp -rf . /p4c/
-        (cd /p4c/ && sudo -E tools/ci-build.sh)
-        sudo cp -rf /p4c/.ccache .
```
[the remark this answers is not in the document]
davidbolvansky: Removed subshell will break this line as you will not copy to itself (see cd above!)

.github/workflows/ci-validation.yml
```diff
@@ -19,8 +19,8 @@ jobs:
     env:
       CTEST_PARALLEL_LEVEL: 4
       IMAGE_TYPE: test
-      ENABLE_UNIFIED_COMPILATION: ON
       VALIDATION: ON
+      BUILD_SUCCESS: true
```
davidbolvansky, line 23: I would default to false

.github/workflows/ci-validation.yml
```diff
@@ -36,9 +36,16 @@ jobs:
     - name: Build (Ubuntu 20.04)
       run: |
         sudo cp -rf . /p4c/
-        (cd /p4c/ && sudo -E tools/ci-build.sh)
+        cd /p4c/ && sudo -E tools/ci-build.sh || echo "BUILD_SUCCESS=false" >> $GITHUB_ENV
```
davidbolvansky, line 39: (cd /p4c/ && sudo -E tools/ci-build.sh && echo "BUILD_SUCCESS=true" >> $GITHUB_ENV)

fruffy: Will wait for #3792.

davidbolvansky:
Please fix failing CI - otherwise I see no blockers to proceed with this PR, do you?

.github/workflows/ci-validation.yml
```diff
@@ -35,8 +35,15 @@ jobs:
 
     - name: Build (Ubuntu 20.04)
       run: |
-        tools/ci-build.sh
+        tools/ci-build.sh || echo "BUILD_SUCCESS=false" >> $GITHUB_ENV
```
vlstill, line 38: Why is it not sufficient to fail normally?

fruffy: This tool is not part of the P4C repository, if it does not build it should not block a PR. It must be fixed separately.

fruffy:
@mbudiu-vmw @vhavel This PR needs authorized approval because it touches some Testgen code.

mbudiu-vmw:
CI and cmake are not my speciality, so I'd rather have someone else comment.

davidbolvansky (approved):

fruffy:
> CI and cmake are not my speciality, so I'd rather have someone else comment.

The issue is that this PR can not be emerged unless someone of the code owners approves. It is a little unfortunate.

fruffy:
@vhavel @pkotikal This PR is stuck because of the reviewer lock we have on testgen files. If you have some time could you give this a review? This PR deprecates our homecooked unified build in favor of CMake's unity builds.

frontends/common/parser_options.cpp
```diff
@@ -490,8 +490,8 @@ bool ParserOptions::isAnnotationDisabled(const IR::Annotation *a) const {
 }
 
 DebugHook ParserOptions::getDebugHook() const {
-    using namespace std::placeholders;
-    auto dp = std::bind(&ParserOptions::dumpPass, this, _1, _2, _3, _4);
+    auto dp = std::bind(&ParserOptions::dumpPass, this, std::placeholders::_1,
+                        std::placeholders::_2, std::placeholders::_3, std::placeholders::_4);
     return dp;
```
vlstill, line 495: As far as I understand it, the original code was actually using placeholders the way they were intended. There is basically nothing wrong with *local* `using namespace`.

.github/workflows/ci-validation.yml
fruffy: I believe most of the tests being non-required is an accident. We do not have control over adjusting the settings which of the tests are required. We should fix this. 

I am okay with removing this check, I maintain the tool after all. 

@mbudiu-vmw Should we make this a non-required test?

frontends/common/parser_options.cpp
fruffy: For bf-p4c, I had to make the namespace explicit because the unified builds also ended up including boost placeholders. This will lead to a conflict. It seems to be an unfortunate side-effect of including boost.

backends/p4tools/modules/testgen/targets/bmv2/backend/stf/stf.cpp
```diff
@@ -235,7 +235,7 @@ inja::json::array_t STF::getClone(const std::map<cstring, const TestObject *> &c
     return cloneJson;
 }
 
-static std::string getTestCase() {
+std::string STF::getTestCaseTemplate() {
```
mbudiu-vmw, line 238: how are these changes related to unified builds?

mbudiu-vmw (commented):
This looks fine, but I really don't know whether the new requirements will prevent people with older software from building the compiler.

backends/p4tools/modules/testgen/targets/bmv2/backend/stf/stf.cpp
fruffy: When multiple `.cpp` files are compiled together this function violates the one-definition rule and the compiler complains.

fruffy:
> This looks fine, but I really don't know whether the new requirements will prevent people with older software from building the compiler.

We do have a CI run for Ubuntu 18, which already requires some workarounds. Ubuntu 16.04 we dropped support a long time ago. I do not know about other systems.

frontends/common/parser_options.cpp
vlstill: Maybe someone is using `using namespace boost::placeholders` globally / on .cpp level? That would be the actual bug... (or not exactly bug, but in my opinion that would be the wrong use of the using in presence of something like unity build that breaks file scopes).

fruffy: I believe [this](https://www.boost.org/doc/libs/1_71_0/boost/bind.hpp) file is getting pulled in somewhere. 

Related: 
https://stackoverflow.com/questions/53203970/why-boostbind-insists-pulling-boostplaceholders-into-global-namespace

vlstill: We have discussed this internally, the problem is some other boost headers pull the `using namespace boost::placeholders` into global namespace (at least in older boost versions). Therefore this cannot be fixed nicely. In order to allow inclusion of such boost headers (e.g. `<boost/graph/transitive_closure.hpp>` for boost versions < 1.77) we need to keep the uses of `std::placeholders::_1` fully qualified). Thanks @fruffy for tracking the source.

vlstill (approved):

fruffy:
@mbudiu-vmw  @vlstill  and @davidbolvansky  both approved but I can not merge this PR since it is missing an authoritative approval.

jnfoster (approved):

jnfoster:
I believe I have the power. Done.

fruffy:
Thank you!

PR #3495, opened 2022-08-26
Contribute the P4Tools platform and P4Testgen 

fruffy-bfn:
This is the initial commit to add P4Testgen ([Video](https://www.youtube.com/watch?v=yucl3azbh8Q), [Slides](https://opennetworking.org/wp-content/uploads/2022/05/Fabian-Ruffy-Final-Slide-Deck.pdf)) as a back end to P4C. This is rather large pull request, however all the parts included are necessary to generate proper tests using P4Testgen.

Description:

# P4Tools
P4Tools is a collection of tools that make testing P4 targets and programs a little easier. So far the platform supports the following tools and projects:

- [P4Testgen](https://github.com/fruffy-bfn/p4c/tree/p4tools/backends/p4tools/testgen): An input-output test case generator for P4.

### Building

P4Tools can be built using the following CMAKE configuration in the P4C repository.

```
mkdir build
cd build
cmake .. -DENABLE_GMP=OFF -DCMAKE_EXPORT_COMPILE_COMMANDS=ON -DENABLE_TOOLS=ON
make
```

### Dependencies
* [inja](https://github.com/pantor/inja) template engine for testcase generation.
* [gsl-lite](https://github.com/gsl-lite/gsl-lite) C++ Core Guidelines Support Library, used to
support and enforce C++ best practices.
* [z3](https://github.com/Z3Prover/z3) P4Tools requires a Z3 version that is at least 4.8.14. 

# P4Testgen - Generating Tests for P4 Targets

The P4Testgen tool uses symbolic execution to automatically generate input-output tests for a given P4 program. P4Testgen attempts to generate a set of tests of all reachable paths within the program. Each test consists of an input packet, the control-plane commands to populate match-action tables, and a predicate on the output packet. Each generated test also has a human-readable trace that indicates the path through the program being exercised by that test.

Currently, P4Testgen supports generating tests for the v1model architecture of the behavioral model. 

## Usage
After building, the main binary can be found in `build/p4check`.

To generate tests for a particular v1model program run
```
build/p4check testgen --target bmv2 --arch v1model --std p4-16 -Ibuild/p4include --test-backend STF --print-traces --seed 1000 --max-tests 10 --out-dir build/testgen/testgen-p4c-bmv2/bmv2_table_default_action.out backends/p4tools/testgen/targets/bmv2/test/p4-programs/bmv2_table_default_action.p4
```
`p4check` is an umbrella binary which delegates execution to all tools of the p4-tools repository.
`testgen` specifies that the testgen tool should be used. In the future, other tools may be supported, for example random program generation (smith, mutate) or translation validators.

Once P4Testgen has generated tests, the tests can be executed by either the P4Runtime or STF test back ends.

### Running ctest
Once compiled, run ctest in the build directory in the following way : `ctest -V -R testgen-` 

This will execute all the tests that are currently related to P4Testgen.


fruffy-bfn:
@mbudiu-vmw This pull request is basically ready to go. I tried to add p4testgen in a non-invasive way which should not affect the rest of the compiler. We require an additional CI run but this run takes roughly the same amount as the other test runs. The CI run compiles p4tools/p4testgen and then generates 10 tests for each v1model bmv2 P4 program. 

To reduce the burden on CI I introduced #3485  and #3491 which will substantially reduce the number of CI runs. However, until they are ready it might take a little while.

Please let us know if you have any questions and concerns related to this pull request. Unfortunately, I do not think we can break it into smaller parts since they will negatively impact the remaining development workflow. However, we will consolidate and contribute parts that are duplicated and/or redundant over the next couple of weeks, once the tool is merged. 

I will also try to get other compiler developers to take a look over this PR, considering how large it is.

jnfoster (approved):
🎉🎉🎉
🚀🚀🚀

I confess I have only skimmed this mega PR -- I haven't reviewed the code with a fine-tooth comb. However, I've been a part of the team developing this code for over a year, so I know it well. And I just confirmed that I can build it from a fresh clone on Mac OS. So I think this is good to go!

Congrats @fruffy-bfn! What a great milestone!

backends/p4tools/cmake/FindZ3.cmake
```diff
@@ -0,0 +1,68 @@
...
+find_path(
+  Z3_INCLUDE_DIR
+  NAMES z3++.h
+  HINTS ${PC_Z3_INCLUDEDIR} ${PC_Z3_INCLUDE_DIRS}
+  PATH_SUFFIXES z3 Z3
+)
+
+find_library(
+  Z3_LIBRARY
+  NAMES libz3.a
```
rst0git, line 13: The following change would allow to build with `libz3.so` (e.g., when installed from a package)
```Suggestion
  NAMES z3
```

rst0git (approved):

CMakeLists.txt
```diff
@@ -35,6 +35,7 @@ OPTION (ENABLE_EBPF "Build the EBPF backend (required for the full test suite)"
 OPTION (ENABLE_UBPF "Build the uBPF backend (required for the full test suite)" ON)
 OPTION (ENABLE_DPDK "Build the DPDK backend (required for the full test suite)" ON)
 OPTION (ENABLE_P4TEST "Build the P4Test backend (required for the full test suite)" ON)
+OPTION (ENABLE_TOOLS "Build the P4Tools development platform" OFF)
```
mbudiu-vmw, line 38: "TOOLS" is a very generic name. Should this be called TEST_TOOLS?

backends/p4tools/README.md
```diff
@@ -0,0 +1,110 @@
...
+
+```
+p4tools
+ ├─ cmake       ── common CMake modules
+ ├─ common      ── C++ source: common code for the various components of p4check
+ ├─ mutate      ── C++ source: p4mutate
+ ├─ p4check     ── C++ source: the main entry point for p4check
+ ├─ scripts
+ │   ├─ hooks   ── git hooks
+ │   └─ tools   ── development-support scripts (e.g., cpplint)
```
mbudiu-vmw, line 13: can't we reuse the existing cpplint tool?

backends/p4tools/README.md
```diff
@@ -0,0 +1,110 @@
...
+  environment or a symbol table), it's probably in `lib`.
+
+### C++ Coding style
+
+We generally follow the [Google C++ Style
+Guide](https://google.github.io/styleguide/cppguide.html). This is partially
+enforced by `cpplint` and `clang-format` and their respective configuration files. Both tools run in a git hook and as part of CI. To run `clang-format` on Ubuntu 16.04, install it with `pip3 install --user clang-format`.
+
+Some deviations from the Style Guide are highlighted below. While this
+repository depends heavily on the main P4 compiler ([P4C](https://github.com/p4lang/p4c/)), the code there doesn't always follow our style; don't follow its precedent!
```
mbudiu-vmw, line 58: that is unfortunate and too late to change now.

backends/p4tools/cmake/common.cmake
```diff
@@ -0,0 +1,49 @@
+# Common configuration across all P4Testgen projects.
```
mbudiu-vmw, line 1: so is this P4tools or p4testgen? I prefer the latter.

backends/p4tools/common/compiler/configuration.h
```diff
@@ -0,0 +1,30 @@
...
+#define COMMON_COMPILER_CONFIGURATION_H_
+
+#include <climits>
+#include <limits>
+
+#include "ir/configuration.h"
+
+namespace P4Tools {
+
+/// A P4CConfiguration implementation that doesn't limit the maximum width for a bit field or
```
mbudiu-vmw, line 11: it is a limit, but larger

backends/p4tools/testgen/testgen.h
```diff
@@ -0,0 +1,33 @@
...
+#include "lib/cstring.h"
+
+#include "backends/p4tools/testgen/lib/final_state.h"
+#include "backends/p4tools/testgen/options.h"
+
+namespace P4Tools {
+
+namespace P4Testgen {
+
+/// This is p4testgen.
```
mbudiu-vmw, line 15: not a particularly useful comment.

backends/p4tools/common/compiler/convert_varbits.h
```diff
@@ -0,0 +1,34 @@
...
+#define COMMON_COMPILER_CONVERT_VARBITS_H_
+
+#include "frontends/common/resolveReferences/referenceMap.h"
+#include "frontends/p4/typeMap.h"
+#include "ir/ir.h"
+#include "lib/null.h"
+
+namespace P4Tools {
+
+/// Converts all existing Type_Varbit types in the program into a custom Size_Type_Varbit type.
```
mbudiu-vmw, line 11: Nothing precludes the runtime implementation of varbits from doing this. In fact, it does in bmv2.

mbudiu-vmw (commented):
At the rate I am reading code I expect this will take a couple of months to review at least.
I hope that there will be an attempt to fix the issues found. Otherwise I am not sure that the review makes sense. Here are some comments on a few files I read.

I really think that it would have been better to do this incrementally.

backends/p4tools/README.md
fruffy-bfn: Yes, I would do this in a separate PR. We can move up a lot of the tools here, they are generally useful.

backends/p4tools/README.md
fruffy-bfn: Actually, I do not think the differences are too big. However, we have more automated tools to check for style, autoformatting, etc. For example, we run https://github.com/fruffy-bfn/p4c/blob/p4tools/backends/p4tools/.clang-format for the PRs to automatically check code formatting. We can do this for the entire compiler. It would just require one big commit that autoformats the code.

Similarly, we run [IWYU](https://github.com/include-what-you-use/include-what-you-use) to clean up all the includes.

backends/p4tools/common/compiler/convert_varbits.h
fruffy-bfn: We need a varbit type that retains the original max-width of the varbit and also contains information about the current assigned width. I do not think this is possible with the current varbit type? How does BMv2 do it? It would simplify the code a lot.

backends/p4tools/cmake/common.cmake
fruffy-bfn: It is P4Tools. P4Testgen is one tool contained within P4Tools. Currently, the only one but there will be more.

fruffy-bfn:
> At the rate I am reading code I expect this will take a couple of months to review at least. I hope that there will be an attempt to fix the issues found. Otherwise I am not sure that the review makes sense. Here are some comments on a few files I read.
> 
> I really think that it would have been better to do this incrementally.

Yes, it is unfortunate, but it is hard to break this down into smaller parts. The code will make little sense and we are blocked on further development until all parts are merged in. We could segment it into the `common`, `testgen`, and `testgen/target` repositories?

backends/p4tools/common/compiler/convert_varbits.h
mbudiu-vmw: Quoting from https://p4.org/wp-content/uploads/2022/07/P4-16-spec.html#sec-varbit-string

> Each occurrence of the type varbit has a statically-declared maximum width, as well as a dynamic width, which must not exceed the static bound. Prior to initialization a variable-size bit-string has an unknown dynamic width.
> 
> Variable-length bit-strings support a limited set of operations:
> 
>     Assignment to another variable-sized bit-string. The target of the assignment must have the same static width as the source. When executed, the assignment sets the dynamic width of the target to the dynamic width of the source.
>     Comparison for equality or inequality with another varbit field. Two varbit fields can be compared only if they have the same type. Two varbits are equal if they have the same dynamic width and all the bits up to the dynamic width are the same. 

Isn't this exactly what you need?

backends/p4tools/scripts/tools/cpplint.py
```diff
@@ -0,0 +1,6903 @@
+#!/usr/bin/env python3
```
mbudiu-vmw, line 1: do we really need two copies of cpplint?
I can understand we need two configurations, but we should be able to reuse the tool.

backends/p4tools/scripts/tools/iwyu_mappings/gcc.libc.imp
```diff
@@ -0,0 +1,211 @@
+# Mappings for GNU libc
```
mbudiu-vmw, line 1: what are these imp files? Do they need to be in the repository?

backends/p4tools/scripts/tools/run-clang-format
```diff
@@ -0,0 +1,62 @@
+#!/bin/bash
+
+
+args=$(getopt -l "write" -o "wh" -- "$@")
```
mbudiu-vmw, line 4: can we give a .sh suffix to this file to make it clear that it's a shell script?

backends/p4tools/testgen/targets/bmv2/register.h
```diff
@@ -0,0 +1,23 @@
...
+#include "backends/p4tools/testgen/options.h"
+#include "backends/p4tools/testgen/targets/bmv2/bmv2.h"
+#include "backends/p4tools/testgen/targets/bmv2/target.h"
+#include "backends/p4tools/testgen/testgen.h"
+
+namespace P4Tools {
+
+namespace P4Testgen {
+
+void bmv2_registerCompilerTarget() { Bmv2::BMv2_V1ModelCompilerTarget::make(); }
```
mbudiu-vmw, line 15: In general I would recommend a little comment in each file describing its purpose.
For example, here based on the file name I thought this handles v1model Registers. 
It seems to be something quite different.

backends/p4tools/testgen/targets/bmv2/test/BMV2ProtobufXfail.cmake
```diff
@@ -0,0 +1,155 @@
...
+p4tools_add_xfail_reason(
+  "testgen-p4c-bmv2-protobuf"
+  "Unknown or unimplemented extern method: fn_foo"
+  issue3091.p4  # user defined extern
+)
+
+p4tools_add_xfail_reason(
+  "testgen-p4c-bmv2-protobuf"
+  "The BMV2 architecture requires 6 pipes"
+  # All of these tests are not valid bmv2 programs but import v1model.
```
mbudiu-vmw, line 116: do they need it? if not, we can remove the include.

backends/p4tools/testgen/targets/bmv2/test/p4-programs/bmv2_action_run.p4
```diff
@@ -0,0 +1,102 @@
+#include <v1model.p4>
```
mbudiu-vmw, line 1: I bet it wouldn't have been too difficult to make all these test programs part of a separate PR.
But now I have started reviewing this one, let's not change this, or I will have to start from scratch.

mbudiu-vmw (commented):
I read a few more files.
I don't plan to read the .p4 files, I assume that they are validated enough by the tools.

backends/p4tools/common/compiler/convert_varbits.h
fruffy: Yes, afaik the `IR::Type_Varbits` does only have one size field. Even if we generated a new Type_Varbits, we would lose information about the original size and the types are not equivalent.

backends/p4tools/scripts/tools/iwyu_mappings/gcc.libc.imp
fruffy: They are mapping files that are used by IWYU. They are quite useful because they help with generating the right includes when using the tool. Without them, IWYU will not be usable.

backends/p4tools/testgen/targets/bmv2/test/BMV2ProtobufXfail.cmake
fruffy: I was thinking that we patch those up once the tool is merged in. We can also move all the sample programs we have into the test folder.

backends/p4tools/common/compiler/convert_varbits.h
mbudiu-vmw: The max size is compile-time information.
The compiler is not concerned with the dynamic size, that only exists at runtime.
I still fail to see why you need your encoding.

backends/p4tools/scripts/tools/iwyu_mappings/gcc.libc.imp
mbudiu-vmw: did you write them by hand?  if they are generated by a tool we should use the tool to generate the files.
or perhaps they come from a repository?

backends/p4tools/testgen/targets/bmv2/test/BMV2Xfail.cmake
```diff
@@ -0,0 +1,305 @@
+# XFAILS: tests that *temporarily* fail
+# ================================================
+# Xfails are _temporary_ failures: the tests should work but we haven't fixed p4testgen yet.
```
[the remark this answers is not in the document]
fruffy: `1. P4C Toolchain Issues ` describes issues we have found with either the compiler or the simple switch model. We have filed these. There are probably more, but we are still just running a limited amount of tests on the programs.

backends/p4tools/smith/smith.h
```diff
@@ -0,0 +1,23 @@
...
+
+#include "backends/p4tools/common/p4ctool.h"
+#include "backends/p4tools/smith/options.h"
+#include "ir/ir.h"
+
+namespace P4Tools {
+
+namespace P4Smith {
+
+class Smith : public AbstractP4cTool<SmithOptions> {
```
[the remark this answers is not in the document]
fruffy: I can remove this.

backends/p4tools/common/compiler/convert_varbits.h
fruffy: Our interpreter assigns a dynamic size [here](https://github.com/p4lang/p4c/pull/3495/files#diff-ed409c5ff044fe5d8cafe027e7ef5007f433def7e39bd28974d9f3253c7be1aeR36) when we call an extract. At this point the varbit will have both information about the original maximum size and the dynamic size. We use the maximum size as means of comparison to avoid type mismatch issues.  
https://github.com/p4lang/p4c/pull/3495/files#diff-bfbb8698949a4875ee7cd51ba9a17c7a58f9a7a6a1f4542455a3b551f483ba30R18
(I will need to update that comment)

backends/p4tools/scripts/tools/iwyu_mappings/gcc.libc.imp
fruffy: https://github.com/include-what-you-use/include-what-you-use
They are handpicked from here. We could potentially link against this module and run the tool from there. 
`p4c.imp` is written by us.

backends/p4tools/testgen/targets/bmv2/test/p4-programs/bmv2_action_run.p4
fruffy: Yes, but then we would have not known whether the tool executes these program correctly. I will create a PR that contributes them to the testdata repository.

backends/p4tools/scripts/tools/iwyu_mappings/gcc.libc.imp
rst0git: @fruffy It might be useful to add a `Makefile` that would make it easier to update these files in the future:

```Makefile
all: gcc.libc.imp libcxx.imp stl.c.headers.imp stl.c.headers.imp boost-1.75-all.imp

gcc.libc.imp:
	curl -s https://raw.githubusercontent.com/include-what-you-use/include-what-you-use/master/gcc.libc.imp -o $@

libcxx.imp:
	curl -s https://raw.githubusercontent.com/include-what-you-use/include-what-you-use/master/libcxx.imp -o $@

stl.c.headers.imp:
	curl -s https://raw.githubusercontent.com/include-what-you-use/include-what-you-use/master/stl.c.headers.imp -o $@

boost-all.imp:
	curl -s https://raw.githubusercontent.com/include-what-you-use/include-what-you-use/master/boost-all.imp -o $@

boost-1.75-all.imp:
	curl -s https://raw.githubusercontent.com/include-what-you-use/include-what-you-use/master/boost-1.75-all.imp -o $@
```

It would be useful to explain how these files are used in `README.md`.

backends/p4tools/testgen/targets/bmv2/test/p4-programs/bmv2_action_run.p4
fruffy-bfn: On second thought, I am not entirely sure whether it makes sense to include them. They might add unnecessary CI time. They are not testing very unique cases.

backends/p4tools/testgen/targets/bmv2/register.h
fruffy-bfn: In general, we try to add comments to each function definition but sometimes things slip through. I am hoping that we are able to refactor this whole structure very soon to make it a bit simpler. It has grown over time and is starting to become confusing.

backends/p4tools/scripts/tools/iwyu_mappings/Makefile
```diff
@@ -0,0 +1,23 @@
...
+boost-all.imp:
+	curl -s https://raw.githubusercontent.com/include-what-you-use/include-what-you-use/master/boost-all.imp -o $@
+
+boost-1.75-all.imp:
+	curl -s https://raw.githubusercontent.com/include-what-you-use/include-what-you-use/master/boost-1.75-all.imp -o $@
+
+clean:
+	rm -f $(iwyu_mappings)
+
+refresh: clean all
```
rst0git, line 23:
```Makefile
.PHONY: all clean refresh
```

backends/p4tools/testgen/test/lib/taint.cpp
```diff
@@ -0,0 +1,280 @@
...
+
+namespace Test {
+
+namespace {
+
+using P4Tools::IRUtils;
+using P4Tools::Taint;
+
+/// Test whether taint propagation works at all.
+/// Input: 8w2 + taint
```
mbudiu-vmw, line 24: since taint has a width, perhaps you should indicate it here. `taint<8>`?

backends/p4tools/testgen/backend/tf.h
```diff
@@ -0,0 +1,44 @@
...
+
+#include "lib/cstring.h"
+
+#include "backends/p4tools/testgen/lib/test_spec.h"
+
+namespace P4Tools {
+
+namespace P4Testgen {
+
+/// Extracts information from the @testSpec to emit a TF test case.
```
mbudiu-vmw, line 16: TF = Test Framework?

mbudiu-vmw: This comment seems wrong to me. Isn't this a base class that should be overridden? If that's correct the constructor should be protected.

backends/p4tools/scripts/tools/README.md
```diff
@@ -0,0 +1,60 @@
+# Tools
+Contains utility scripts for supporting the development of P4Tools.
+
+## IWYU
+
+[IWYU](https://github.com/include-what-you-use/include-what-you-use) checks ensures that every symbol used in a C/C++ file has a corresponding include. Vice-versa IWYU also ensures that unnecessary includes are removed.
```
mbudiu-vmw, line 6: remove "checks"

backends/p4tools/scripts/hooks/pre-commit
```diff
@@ -0,0 +1,98 @@
+#!/usr/bin/env python3
+
+"""pre-commit linter
+
+Catches some basic coding style by running `tools/cpplint.py` on code being
```
mbudiu-vmw, line 5: maybe we should do this for the compiler too

backends/p4tools/common/options.cpp
```diff
@@ -0,0 +1,226 @@
...
+    registerOption(
+        "--seed", "seed",
+        [this](const char* arg) {
+            seed = std::stoul(arg);
+            return true;
+        },
+        "Provides a randomization seed");
+
+    // Inherit some compiler options, setting them up to be forwarded to the compiler.
+    std::vector<InheritedCompilerOptionSpec> inheritedCompilerOptions = {
```
mbudiu-vmw, line 164: can't this be done by inheriting another Options class?

backends/p4tools/testgen/targets/bmv2/target.cpp
```diff
@@ -0,0 +1,117 @@
...
+void BMv2_V1ModelTestgenTarget::make() {
+    static BMv2_V1ModelTestgenTarget* INSTANCE = nullptr;
+    if (INSTANCE == nullptr) {
+        INSTANCE = new BMv2_V1ModelTestgenTarget();
+    }
+}
+
+const BMv2_V1ModelProgramInfo* BMv2_V1ModelTestgenTarget::initProgram_impl(
+    const IR::P4Program* program, const IR::Declaration_Instance* mainDecl) const {
+    // The blocks in the main declaration are just the arguments in the constructor call.
```
mbudiu-vmw, line 39: I actually would very much like to automate this step -- see https://github.com/p4lang/p4c/pull/3281
This should make it much easier to write tools specific to a model.

mbudiu-vmw: Maybe if you have no other project going on we could team up?

backends/p4tools/testgen/lib/exceptions.h
```diff
@@ -0,0 +1,41 @@
+#ifndef TESTGEN_LIB_EXCEPTIONS_H_
+#define TESTGEN_LIB_EXCEPTIONS_H_
+
+#include "lib/exceptions.h"
+
+namespace P4Tools {
+
+namespace P4Testgen {
+
+/// This class indicates a feature that is not implemented in P4Testgen.
```
mbudiu-vmw, line 10: I think this file should be called "unimplemented.h".

backends/p4tools/testgen/core/target.h
```diff
@@ -0,0 +1,164 @@
...
+#include "backends/p4tools/testgen/lib/execution_state.h"
+#include "backends/p4tools/testgen/lib/namespace_context.h"
+#include "backends/p4tools/testgen/lib/test_backend.h"
+
+namespace P4Tools {
+
+namespace P4Testgen {
+
+/// Specifies a canonical representation of the target pipeline as documented in P4 code.
+class ArchSpec {
```
mbudiu-vmw, line 32: this could be automated: https://github.com/p4lang/p4c/pull/3281, I have left another comment about this.

backends/p4tools/common/core/solver.h
```diff
@@ -0,0 +1,60 @@
...
+    /// Saves solver state to the given JSON generator.
+    virtual void toJSON(JSONGenerator&) const = 0;
+
+    /// @returns whether this solver is incremental.
+    virtual bool isInIncrementalMode() const = 0;
+
+    /// Cast an abstract solver to its specific sub type.
+    /// Casts involving "const" have to explicit in this case. This is necessary because abstract
+    /// solvers have no straightforward mechanism to support clone operations.
+    template <typename T>
```
mbudiu-vmw, line 52: was thinking to move this to a base class that everyone who wants to do "to" could inherit from. including in the compiler.

backends/p4tools/common/lib/coverage.cpp
```diff
@@ -0,0 +1,59 @@
...
+
+#include "lib/log.h"
+
+namespace P4Tools {
+
+namespace Coverage {
+
+CollectStatements::CollectStatements(CoverageSet& output) : statements(output) {}
+
+bool CollectStatements::preorder(const IR::AssignmentStatement* stmt) {
```
mbudiu-vmw, line 13: can't this be done for the Statement class?

mbudiu-vmw (commented):
So far the code looks great, I went through about 160/250 files, but most of the work is still aheda of me. This is another collection of small comments.

backends/p4tools/common/compiler/convert_errors.h
```diff
@@ -0,0 +1,73 @@
...
+#include "lib/null.h"
+#include "lib/safe_vector.h"
+#include "midend/convertEnums.h"
+
+namespace P4Tools {
+
+/**
+ * Policy function: given a number of error values should return
+ * the size of a Type_Bits type used to represent the values.
+ * This class is lefted from enum conversion path.
```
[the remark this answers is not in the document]
fruffy-bfn: It is a modification of the convert_enums pass, there is no convert_errors pass. We were not able to consolidate those two passes, sadly. Maybe there is a good way? We can move up all these passes in the midend, since they are generally useful.

backends/p4tools/common/compiler/copy_headers.h
```diff
@@ -0,0 +1,35 @@
...
+namespace P4Tools {
+
+/// Converts assignments of headers, header stacks, and structs into assignments of their
+/// individual fields.
+///
+/// Header assignments are converted into an `if` statement. If the source header is valid, then
+/// the validity bit in the destination header is set and the component fields are copied.
+/// Otherwise, if the source header is invalid, then the validity bit in the destination header is
+/// cleared.
+class DoCopyHeaders : public Transform {
```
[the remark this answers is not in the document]
fruffy-bfn: It extends the copyStructures pass. I think it might be possible to merge these two.

backends/p4tools/common/lib/coverage.cpp
fruffy-bfn: I believe this was done because not all statements are relevant. For example, switch or block statements are only wrappers for the assignment and method call statements contained within.

backends/p4tools/common/compiler/copy_headers.h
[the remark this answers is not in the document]
fruffy-bfn: I will add an issue for both of the passes, so we can make sure to add them to the midend folder later and potentially merge or fix them.

backends/p4tools/testgen/targets/bmv2/test/STFTestTemplate.cmake
```diff
@@ -0,0 +1,68 @@
+# This file defines how a test should be written for a particular target. This is used by testutils
```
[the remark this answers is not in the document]
fruffy-bfn: This script automatically runs STF tests on the tests generated by P4Testgen. The BMv2 back end also supports the generating of PTF and Protobuf tests, but we do not have a test runner for these. I think these CMake files can also 
be cleaned up.

backends/p4tools/common/options.cpp
fruffy-bfn: I am actually not sure why this options class was written this way. Let me see.

backends/p4tools/testgen/lib/exceptions.h
fruffy-bfn: There is a plan to add finer-grained exceptions. For example, exceptions that indicate that the tool has encountered a path that is not valid, exceptions for unimplemented features, or exceptions that are fatal. We are still using COMPILER_BUG, but those messages are confusing.

backends/p4tools/common/core/solver.h
fruffy-bfn: Yes, we can do this. But there are some includes such as model and formulae.h that make this class a little opinionated.

backends/p4tools/testgen/targets/bmv2/target.cpp
fruffy: Yes, something like this is sorely needed for testgen because we have to map the structure of the package to the nodes of the parsed P4 program.. But I am not entirely sure how we would consume this model yet. 

I am happy to contribute to this, but I unfortunately have little time.

backends/p4tools/testgen/targets/bmv2/constants.cpp
```diff
@@ -0,0 +1,11 @@
+#include "backends/p4tools/testgen/targets/bmv2/constants.h"
```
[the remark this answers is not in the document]
fruffy: Some constants were moved up and now the file is empty. It is probably useful to keep it when you have to deal with constants that can not be defined in a header-only file.

fruffy-bfn:
@VolodymyrPeschanenkoIntel  issue281 is failing because of a bug in loopsUnroll. The pass adds a OutofBounds transition although the parser simply contains an infinite loop.

backends/p4tools/common/options.cpp
fruffy-bfn: I tried simplifying this but unfortunately this requires a major rewrite. These options are written this way to make it possible to have every target extension define its own options without needing to produce a different binary. 
`p4check testgen`, `p4check mutate`,  `p4check verify` can all have their own options.

PR #3496, opened 2022-08-29
Dpdk Backend: Fix support for non-zero action arguments for default action in a table using action selector

usha1830:
When a table has psa/pna_implementation property set to Action selector, it is split into 3 tables, namely group table, member table and base table. 
- The action list of the base table after split contains a compiler generated action` <tblname>_set_group_id` and the default action from the original table in input P4 program.  While creating the action list, the default action was added along with its arguments causing typechecking error. 
```
"Werror=type-error] error: 32w0: parameter vport cannot be bound: it is set by the control plane
        const default_action = a1(vport);
                                             ^^^^^^
```

The fix in this PR handles this issue and also adds code to preserve the const-ness of default action.

### DIFF
--- a/backends/dpdk/dpdkArch.cpp
+++ b/backends/dpdk/dpdkArch.cpp
@@ -1692,7 +1692,14 @@ SplitP4TableCommon::create_match_table(const IR::P4Table *tbl) {
 
     auto actionCall = new IR::MethodCallExpression(new IR::PathExpression(actionName));
     actionsList.push_back(new IR::ActionListElement(actionCall));
-    actionsList.push_back(new IR::ActionListElement(tbl->getDefaultAction()));
+    auto default_action = tbl->getDefaultAction();
+    if (default_action) {
+        if (auto mc = default_action->to<IR::MethodCallExpression>()) {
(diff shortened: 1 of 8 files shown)


backends/dpdk/dpdkArch.cpp
```diff
@@ -1692,12 +1692,23 @@ SplitP4TableCommon::create_match_table(const IR::P4Table *tbl) {
...
+    auto default_action = tbl->getDefaultAction();
+    if (default_action) {
+        if (auto mc = default_action->to<IR::MethodCallExpression>()) {
+            // Ignore action params of default action by creating new MethodCallExpression
+            auto defAction = new IR::MethodCallExpression(mc->method->to<IR::PathExpression>());
+            actionsList.push_back(new IR::ActionListElement(defAction));
+        }
+    }
+
+    auto constDefAction = tbl->properties->getProperty("default_action");
```
mbudiu-vmw, line 1704: could this be null?

usha1830: Thanks for pointing out. I will add a check for that

PR #3498, opened 2022-08-29
Handle table.apply().hit correctly in ubpf backend

### LINKED ISSUE #3483: Error occur when compile C file generated by p4c-ubpf with clang
I compile p4 file which includes if statement using p4c-ubpf.  The control block of P4 file like follows:
```
control pipe(inout headers hdr, inout metadata meta, inout standard_metadata std_meta) {

    apply {
        if (port_mapping.apply().hit) {
            bd.apply();
        } else {
            port_assign.apply();
        }
    }

}
```

and the generated c file is as follows:
```
            if ({
                /* construct key */
                struct pipe_port_mapping_key key = {}; 
                key.std_meta_input_port = std_meta->input_port;
                /* value */
                struct pipe_port_mapping_value *value = NULL;
                /* perform lookup */
                value = ubpf_map_lookup(&pipe_port_mapping, &key);

                if (value == NULL) {
                    /* miss; find default action */
                    hit = 0;
                    value = ubpf_map_lookup(&pipe_port_mapping_defaultAction, &ebpf_zero);
                } else {
                    hit = 1;
                }   
                if (value != NULL) {
                    /* run action */
                    switch (value->action) {
                        case pipe_set_bd: 
                        {   
                            meta.ingress_metadata.bd = value->u.pipe_set_bd.bd;
                        }   
                        break;
                        case pipe_port_mapping_NoAction: 
                        {   
    
                        }   
                        break;
                        default: return 1;
                    }   
                }   
                else return 1;
            }   
.hit) {
```

now, the if statement is a block, and error occurs compile it with clang,  I don't know if I made a mistake.

mbudiu-vmw:
Fixes #3483

### DIFF
--- a/backends/ubpf/ubpfControl.cpp
+++ b/backends/ubpf/ubpfControl.cpp
@@ -429,8 +429,21 @@ namespace UBPF {
     }
 
     bool UBPFControlBodyTranslator::preorder(const IR::IfStatement *statement) {
+        bool isHit = P4::TableApplySolver::isHit(statement->condition, control->program->refMap,
+                                                 control->program->typeMap);
+        if (isHit) {
+            // visit first the table, and then the conditional
+            auto member = statement->condition->to<IR::Member>();
+            CHECK_NULL(member);
+            visit(member->expr);  // table application.  Sets 'hitVariable'
+            builder->emitIndent();
+        }
+
         builder->append("if (");
-        visit(statement->condition);
+        if (isHit)
+            builder->append(control->hitVariable);
+        else
+            visit(statement->condition);
         builder->append(") ");
         if (!statement->ifTrue->is<IR::BlockStatement>()) {
             builder->blockStart();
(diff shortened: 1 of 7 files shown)


mbudiu-vmw:
@osinstom can you please review this?
It should be fine, it parallels the ebpf backend which has fixed this a long time ago.

hesingh:
Now we have yet another place to fix such an issue. It is the ebpf psa code. This code does not even support IfStatement in translator.

osinstom (approved):

osinstom:
LGTM - sorry for ignoring #3483, I'm quite busy recently with other projects.

osinstom:
@hesingh ebpf psa code inherits from "legacy" ebpf code. In fact, `.hit` is used extensively in PTF tests/examples, for example: https://github.com/p4lang/p4c/blob/main/backends/ebpf/psa/examples/upf.p4.

rst0git (approved):

PR #3502, opened 2022-08-31
Add Toeplitz hash algorithm to v1model

qobilidop:
We'd like to make the Toeplitz hash algorithm available in v1model. The actual Toeplitz hash function has already been added to BMv2: https://github.com/p4lang/behavioral-model/pull/1134.

### DIFF
--- a/backends/bmv2/common/extern.cpp
+++ b/backends/bmv2/common/extern.cpp
@@ -259,6 +259,8 @@ ExternConverter::convertHashAlgorithm(cstring algorithm) {
         result = "csum16";
     else if (algorithm == P4V1::V1Model::instance.algorithm.xor16.name)
         result = "xor16";
+    else if (algorithm == P4V1::V1Model::instance.algorithm.toeplitz.name)
+        result = "toeplitz";
     else
         ::error(ErrorType::ERR_UNSUPPORTED, "Unsupported algorithm %1%", algorithm);
     return result;
--- a/backends/bmv2/simple_switch/simpleSwitch.cpp
+++ b/backends/bmv2/simple_switch/simpleSwitch.cpp
@@ -251,7 +251,8 @@ Util::IJson* ExternConverter_hash::convertExternFunction(
         v1model.algorithm.crc32.name, v1model.algorithm.crc32_custom.name,
         v1model.algorithm.crc16.name, v1model.algorithm.crc16_custom.name,
         v1model.algorithm.random.name, v1model.algorithm.identity.name,
-        v1model.algorithm.csum16.name, v1model.algorithm.xor16.name
+        v1model.algorithm.csum16.name, v1model.algorithm.xor16.name,
+        v1model.algorithm.toeplitz.name
     };
 
     if (mc->arguments->size() != 5) {
--- a/frontends/p4/fromv1.0/programStructure.cpp
+++ b/frontends/p4/fromv1.0/programStructure.cpp
@@ -1121,6 +1121,8 @@ const IR::Expression* ProgramStructure::convertHashAlgorithm(
         result = v1model.algorithm.csum16.Id();
     } else if (algorithm == "xor16") {
         result = v1model.algorithm.xor16.Id();
+    } else if (algorithm == "toeplitz") {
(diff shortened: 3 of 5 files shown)


jnfoster:
Thanks for the contribution.

I appreciate that having support in Bmv2 is sufficient for your intended use case. However, If we add an extern to `v1model.p4` then _any_ target that claims to support the V1Model architecture will have an additional requirement: it will also needs to implement the Toeplitz hash. And this may be difficult or impossible on some hardware targets.

So I think this needs a discussion before we can proceed. Perhaps the Architecture Working Group can take it on?

qobilidop:
Thanks for your prompt response and explanation! I totally understand the concerns. I'm willing to participate in any discussion needed.

qobilidop:
I have some questions about the v1model architecture. Quoting from https://github.com/p4lang/behavioral-model/blob/main/docs/simple_switch.md:

> The v1model architecture is included with the p4c compiler, and was designed to be identical to the P4_14 switch architecture, enabling straightforward auto-translation from P4_14 programs to P4_16 programs that use the v1model architecture.

So this pretty much means v1model follows the [P4_14 spec](https://p4.org/p4-spec/p4-14/v1.1.0/tex/p4.pdf). And since P4_16 is developing the new PSA spec, the P4_14 spec will stay as it is now. So we shouldn't modify v1model because that breaks the original intention to follow P4_14?

jnfoster:
V1model started out as an architecture for internal use by tools transforming P4_14 code into P4_16. But now it's kind of acquired a life of its own. (For example, your company uses it for validation tools :-)

In general, P4_16's architectures give you the flexibility to create your own pipelines with whatever structure and externs you want. And Bmv2 is *designed* to be a flexible software switch framework that gives you a bunch of features for implementing simple software targets those architectures. In practice, to the best of my knowledge, few people have developed their own architectures to date. But all of the pieces are there. It would take work, but you wouldn't have to revisit the essential elements of P4_16's design. 

For the question of whether we can take a new extern, like Toeplitz hash, in a standard (i.e., "blessed" by the community) architecutre like V1Model or PSA, that's a question for which we need community buy in. I personally don't mind if we add externs more aggressively and allow targets to "subset". But it may confusing if we have many such externs, targets only implement idiosyncratic subsets, and yet they claim to achieve "compliance" with the architecture. So we need to be a bit careful here and establish customs and norms for the community. 

By the way, these questions are timely as we are dealing with similar questions in the PNA effort (there, more related to the pipeline itself, but I wouldn't be surprised if there are soon similar discussions about externs.) @jafingerhut may want to weigh in.

mbudiu-vmw:
A partial simple solution would be to publish this hash function in a separate library which can be `#include`d for targets that support it. What is impossible to do is to patch the enum that holds the names of the legal values.

qobilidop:
> A partial simple solution would be to publish this hash function in a separate library which can be `#include`d for targets that support it. What is impossible to do is to patch the enum that holds the names of the legal values.

I'm interested in this solution. Could you share more details or existing examples on how to do this?

I also just found the [instructions on how to implement custom externs](https://github.com/p4lang/behavioral-model/issues/697). I'm still going through the details there, so I don't know if it's the same as your proposal or a different one.

mbudiu-vmw:
From the p4c compiler there isn't really any work to do. You just put the declaration of your extern/function in a p4 file.
You probably have to name it something like `hash_toeplitz` instead of `hash`. Then people who want to use it will just include the p4 file in their programs and call the function. If the backend supports it, they are fine, otherwise they should get an error. 
This isn't different from libraries in any other language really. The main question is whether we publish this one in p4include.

qobilidop:
Close this PR because I no longer need to modify v1model. The [alternative approach of using custom extern](https://github.com/p4lang/behavioral-model/pull/1140) works.

mbudiu-vmw:
I am actually planning to bring this up next Monday in the P4 language design working group.
It is a good question about how to evolve the model.

qobilidop:
> I am actually planning to bring this up next Monday in the P4 language design working group. It is a good question about how to evolve the model.

Ahh, sorry if I caused any inconvenience here. It's great to know there will be some related discussions!

PR #3510, opened 2022-09-03
Implement the eBPF extension for P4Tools' P4Testgen

fruffy:
This is a proof of concept implementation for P4Testgen to demonstrate its extensibility for other targets. The implementation is a little janky but works.

The eBPF test failures actually do expose flaws in the test back end. 

This PR will be cleaned up and rebased once the P4Tools PR is merged in.


hesingh:
@fruffy If I have new source code under bmv2 from which a library is built, how can I use your tool to test the library's code? I do have several P4 programs to test the library and I use Cisco trex for traffic testing. I guess I first create stf files to test the P4 code?

fruffy:
If you have a test script that can execute STF tests on your target than you can just use the tests generated by P4Testgen to test your target. For example, the p4c repository has [run-bmv2-test.py](https://github.com/p4lang/p4c/blob/main/backends/bmv2/run-bmv2-test.py) and [run-ebpf-test.py](https://github.com/p4lang/p4c/blob/main/backends/ebpf/run-ebpf-test.py).

However, P4Testgen only support `ebpf_model.p4` and `v1model.p4` for now.

backends/p4tools/common/lib/formulae.cpp
```diff
@@ -0,0 +1,130 @@
...
+    if (member == nullptr) {
+        return false;
+    }
+
+    // A member is a state variable if it is qualified by a PathExpression or if its qualifier is a
+    // state variable.
+    return member->expr->is<IR::PathExpression>() || repOk(member->expr);
+}
+
+bool StateVariable::operator<(const StateVariable& other) const {
```
mbudiu-vmw, line 26: this is nondeterministic isn't it?
maybe it's better to use the node id field?

backends/p4tools/common/lib/formulae.cpp
```diff
@@ -0,0 +1,130 @@
...
+        return compare(p1, e2);
+    }
+    BUG("Not a valid StateVariable: %1%", e1);
+}
+
+int StateVariable::compare(const IR::Member* m1, const IR::Expression* e2) {
+    if (const auto* m2 = e2->to<IR::Member>()) {
+        return compare(m1, m2);
+    }
+    if (e2->is<IR::PathExpression>()) {
```
mbudiu-vmw, line 49: could you just compare typeid values to be more general?

backends/p4tools/common/lib/formulae.cpp
```diff
@@ -0,0 +1,130 @@
...
+    // state variable.
+    return member->expr->is<IR::PathExpression>() || repOk(member->expr);
+}
+
+const IR::MethodCallExpression* ConcolicVariable::getCall() const { return call; }
+
+uint64_t ConcolicVariable::getInstanceId() const { return instanceId; }
+
+bool ConcolicVariable::operator<(const ConcolicVariable& other) const {
+    if (node->toString() < other.node->toString()) {
```
mbudiu-vmw, line 116: in general tostring does not serialize completely an expression, this may be a bug waiting to strike.

backends/p4tools/common/lib/ir.cpp
```diff
@@ -0,0 +1,257 @@
...
+    if (const auto* tb = t->to<IR::Type_Bits>()) {
+        return (tb->isSigned) ? -(big_int(1) << tb->width_bits() - 1) : big_int(0);
+    }
+    if (t->is<IR::Type_Boolean>()) {
+        return 0;
+    }
+    P4C_UNIMPLEMENTED("Maximum value calculation for type %1% not implemented.", t);
+}
+
+const IR::MethodCallExpression* IRUtils::generateInternalMethodCall(
```
mbudiu-vmw, line 201: this is an interesting function, this should be documented better somewhere

backends/p4tools/common/lib/trace_events.cpp
```diff
@@ -0,0 +1,310 @@
...
+}
+
+const TraceEvent::ListExpression* TraceEvent::ListExpression::apply(Transform& visitor) const {
+    return new ListExpression(value->apply(visitor)->checkedTo<IR::ListExpression>(), label);
+}
+
+void TraceEvent::ListExpression::complete(Model* model) const { model->complete(value); }
+
+const TraceEvent::ListExpression* TraceEvent::ListExpression::evaluate(const Model& model) const {
+    IR::Vector<IR::Expression> evalutedExpressions;
```
mbudiu-vmw, line 93: typo in variable name

backends/p4tools/common/lib/taint.cpp
```diff
@@ -0,0 +1,274 @@
...
+#include "backends/p4tools/common/lib/symbolic_env.h"
+#include "lib/cstring.h"
+#include "lib/exceptions.h"
+#include "lib/null.h"
+
+namespace P4Tools {
+
+const IR::StringLiteral Taint::taintedStringLiteral = IR::StringLiteral(cstring("Taint"));
+
+void getTaintPairs(const SymbolicMapType& varMap, const IR::Expression* expr,
```
mbudiu-vmw, line 17: this function deserves some documentation. Could this be a static function?

backends/p4tools/p4check/commands.cpp
```diff
@@ -0,0 +1,177 @@
...
+#include "lib/exceptions.h"
+#include "lib/map.h"
+
+namespace P4Tools {
+
+P4CheckCommands::P4CheckCommands() {
+    registerCommand(HELP, {}, {"help", "-h", "--help"}, "Shows this help message and exits");
+    registerCommand(VERSION, {}, {"version", "--version"}, "Prints version information and exits");
+    registerCommand(TESTGEN, {"p4testgen"}, {"testgen"}, "Generates packet tests for P4 programs");
+    registerCommand(SMITH, {"smith"}, {"smith"}, "Generates random p4 programs");
```
mbudiu-vmw, line 23: are these command implemented for any backend?

backends/p4tools/testgen/core/program_info.h
```diff
@@ -0,0 +1,109 @@
...
+    ProgramInfo& operator=(const ProgramInfo&) = default;
+
+    ProgramInfo& operator=(ProgramInfo&&) = default;
+
+    virtual ~ProgramInfo() = default;
+
+    /// The P4 program from which this object is derived.
+    const IR::P4Program* program;
+
+    /// Casts this object to T. Useful for casting to a concrete implementation of this class.
```
mbudiu-vmw, line 50: this is repeated multiple times, please move it to a base abstract class

mbudiu-vmw (commented):
I still have about 80 C++ files to read.
I think you could have made the effort to break this into several independent PRs.

fruffy-bfn:
This is the eBPF PR, I will try to address the comments in the other PR.

backends/p4tools/common/core/z3_solver.h
```diff
@@ -0,0 +1,138 @@
...
+    friend class Z3JSON;
+    friend class Z3SolverAccessor;
+
+ public:
+    virtual ~Z3Solver() = default;
+
+    explicit Z3Solver(bool isIncremental = true,
+                      boost::optional<std::istream&> inOpt = boost::none);
+
+    void comment(cstring comment) override;
```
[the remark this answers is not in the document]
fruffy-bfn: This is the general code style for this code base, every declaration in a header file should have at least one newline between them.

backends/p4tools/p4check/commands.cpp
fruffy-bfn: They will be added later as part of new pull requests. Currently these should throw an error, if called.

backends/p4tools/testgen/core/program_info.h
fruffy-bfn: Should this be a general abstract class for the compiler?

backends/p4tools/common/lib/formulae.cpp
fruffy-bfn: I believe this code was actually superseded by a definition in `common/testgen.def`. Thanks for catching this.

backends/p4tools/common/lib/formulae.cpp
fruffy-bfn: I would actually like to implement this in an IR def file, but this is a future TODO.

backends/p4tools/common/lib/saturation_elim.h
```diff
@@ -0,0 +1,24 @@
+#ifndef COMMON_LIB_SATURATION_ELIM_H_
+#define COMMON_LIB_SATURATION_ELIM_H_
+
+#include "ir/ir.h"
+
+namespace P4Tools {
+
+/// Contains utility functions for eliminating saturating arithmetic.
+class SaturationElim {
```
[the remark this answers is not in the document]
fruffy: Will do this after the merge? So that we can address the issues with the pass in a cleaner way?

backends/p4tools/common/lib/zombie.cpp
```diff
@@ -0,0 +1,64 @@
...
+    // and names.
+    return *mkZombie(type, isConst, incarnation, name);
+}
+
+const StateVariable* Zombie::mkZombie(const IR::Type* type, bool isConst, int incarnation,
+                                      cstring name) {
+    static IR::PathExpression zombieHdr(new IR::Path(P4tZombie));
+
+    using key_t = std::pair<bool, int>;
+    static std::map<key_t, const IR::Member*> incarnations;
```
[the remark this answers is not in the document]
fruffy-bfn: So we were actually inspired by this style here. https://github.com/p4lang/p4c/blob/main/ir/type.cpp#L67
Also not sure which is better.

backends/p4tools/common/lib/saturation_elim.h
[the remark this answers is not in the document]
mbudiu-vmw: that's the point of the review, to simplify the PR. This should be an easy one.

backends/p4tools/testgen/core/program_info.h
mbudiu-vmw: I think we could have only one for the whole compiler, probably in lib.

mbudiu-vmw (commented):
I have reviewed a few more file, I still have about 60.
The main question is about what part of stf is specific to a target.

backends/p4tools/testgen/core/externs.cpp
```diff
@@ -0,0 +1,114 @@
...
+
+namespace P4Testgen {
+
+bool ExternMethodImpls::exec(const IR::MethodCallExpression* call, const IR::Expression* receiver,
+                             IR::ID& name, const IR::Vector<IR::Argument>* args,
+                             const ExecutionState& state,
+                             SmallStepEvaluator::Result& result) const {
+    // We have to check the extern type here. We may receive a specialized canonical type, which we
+    // need to unpack.
+    // TODO: How do we handle the arguments of this type specialized?
```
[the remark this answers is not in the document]
fruffy: So this PR depends on #3495. All the changes there will be reflected here once the other one is merged in. This was mostly a test to see whether we can run the eBPF back end easily on CI, which seems to work well.

 I will pick up addressing the review comments next week

mbudiu-vmw:
I am not reviewing two PRs of 50K lines each. I am reviewing just this one.

fruffy:
Once the other PR is merged I will rebase it there should be considerably less lines of code. This PR is still a draft, it was not meant to be reviewed yet. Unfortunately, I can not change the base to fruffy-bfn/p4c.

backends/p4tools/testgen/core/small_step/cmd_stepper.cpp
```diff
@@ -0,0 +1,644 @@
...
+    // Get the initial constraints of the target. These constraints influence branch selection.
+    boost::optional<const Constraint*> cond = programInfo.getTargetConstraints();
+
+    // Have the target break apart the main declaration instance.
+    auto* nextState = new ExecutionState(state);
+
+    // Initialize all relevant environment variables for the respective target.
+    initializeTargetEnvironment(nextState);
+
+    // We mandate all packets to conform to a fixed size. This is a temporary change.
```
mbudiu-vmw, line 376: change from what?

backends/p4tools/testgen/core/small_step/cmd_stepper.cpp
```diff
@@ -0,0 +1,644 @@
...
+    auto pktSize = TestgenOptions::get().packetSize;
+    if (pktSize != 0) {
+        auto maxPktLength = ExecutionState::getMaxPacketLength_bits();
+        if (pktSize < 0 || pktSize > maxPktLength) {
+            ::error("Invalid input packet size. The valid range is from 1 to %s bits.",
```
mbudiu-vmw, line 381: is this %s correct?

backends/p4tools/testgen/core/small_step/cmd_stepper.cpp
```diff
@@ -0,0 +1,644 @@
...
+    nextState->pushContinuation(new ExecutionState::StackFrame(k, state.getNamespaceContext()));
+
+    // Enter the parser state's namespace
+    nextState->pushNamespace(parserState);
+
+    // Replace the parser state with its non-declaration components, followed by the select
+    // expression.
+    std::vector<Continuation::Command> replacements;
+    for (const auto* statOrDecl : parserState->components) {
+        if (statOrDecl->is<IR::IDeclaration>() && !statOrDecl->is<IR::Declaration_Variable>()) {
```
mbudiu-vmw, line 472: aren't local variables to states eliminated?

backends/p4tools/testgen/core/small_step/cmd_stepper.cpp
```diff
@@ -0,0 +1,644 @@
...
+bool CmdStepper::preorder(const IR::BlockStatement* block) {
+    logStep(block);
+
+    auto* nextState = new ExecutionState(state);
+    std::vector<Continuation::Command> replacements;
+    // Do not forget to add the namespace of the block statement.
+    nextState->pushNamespace(block);
+    // TODO (Fabian): Remove this? What is this for?
+    for (const auto* statOrDecl : block->components) {
+        if (statOrDecl->is<IR::IDeclaration>() && !statOrDecl->is<IR::Declaration_Variable>()) {
```
mbudiu-vmw, line 495: locals should have been eliminated by most midends

mbudiu-vmw: but if they aren't they should probably not be ignored

backends/p4tools/testgen/core/small_step/expr_stepper.h
```diff
@@ -0,0 +1,157 @@
...
+    /// We delegate evaluation to the TableStepper, which needs to access protected members.
+    friend class TableStepper;
+
+    /// Extract utils may access some protected members of the expression stepper.
+    friend class ExtractUtils;
+
+ protected:
+    /// Contains information that is useful for externs that advance the parser cursor.
+    /// For example, advance, extract, or lookahead.
+    struct AdvanceInfo {
```
mbudiu-vmw, line 32: Maybe this should be called PacketCursorAdvance

backends/p4tools/testgen/core/small_step/expr_stepper.cpp
```diff
@@ -0,0 +1,475 @@
...
+    }
+    nextState->replaceTopBody(&replacements);
+    result->emplace_back(nextState);
+}
+
+bool ExprStepper::preorder(const IR::Member* member) {
+    logStep(member);
+
+    if (member->expr->is<IR::MethodCallExpression>() &&
+        (member->member.name == IR::Type_Table::hit ||
```
mbudiu-vmw, line 66: I hope there is no confusion with other methods that can have the same name.
you are not checking that these are applied to a table.

backends/p4tools/testgen/core/small_step/expr_stepper.cpp
```diff
@@ -0,0 +1,475 @@
...
+        return stepSymbolicValue(member);
+    }
+    return stepSymbolicValue(state.get(member));
+}
+
+void ExprStepper::evalActionCall(const IR::P4Action* action, const IR::MethodCallExpression* call) {
+    const auto* actionNameSpace = action->to<IR::INamespace>();
+    BUG_CHECK(actionNameSpace, "Does not instantiate an INamespace: %1%", actionNameSpace);
+    auto* nextState = new ExecutionState(state);
+    if (action->body == nullptr) {
```
mbudiu-vmw, line 95: I don't think this can happen

backends/p4tools/testgen/core/small_step/expr_stepper.cpp
```diff
@@ -0,0 +1,475 @@
...
+        // The action has no body, so there is little to do here.
+        nextState->popBody();
+    } else {
+        // If the action has arguments, these are usually directionless control plane input.
+        // We introduce a zombie variable that takes the argument value. This value is either
+        // provided by a constant entry or synthesized by us.
+        for (size_t argIdx = 0; argIdx < call->arguments->size(); ++argIdx) {
+            const auto& parameters = action->parameters;
```
mbudiu-vmw, line 103: you could assert that they are directionless

backends/p4tools/testgen/core/small_step/expr_stepper.cpp
```diff
@@ -0,0 +1,475 @@
...
+        TESTGEN_UNIMPLEMENTED(
+            "The SelectExpression %1% is trying to match on a tainted key set."
+            " This means it is matching on uninitialized data."
+            " P4Testgen currently does not support this case.",
+            selectExpression);
+    }
+
+    // Handle case where the first select case matches: proceed to the next parser state,
+    // guarded by its path condition.
+    const auto* equality = GenEq::equate(selectExpression->select, selectCase->keyset);
```
mbudiu-vmw, line 392: keyset can represent a set, not just a value. I hope that equate handles this.
Otherwise you need something like "ismember"

backends/p4tools/testgen/core/small_step/expr_stepper.cpp
```diff
@@ -0,0 +1,475 @@
...
+
+    if (!SymbolicEnv::isSymbolicValue(slice->e0)) {
+        return stepToSubexpr(slice->e0, result, state, [slice](const Continuation::Parameter* v) {
+            auto* result = slice->clone();
+            result->e0 = v->param;
+            return Continuation::Return(result);
+        });
+    }
+
+    if (!SymbolicEnv::isSymbolicValue(slice->e1)) {
```
mbudiu-vmw, line 452: these should always be constants, otherwise we can't typecheck

backends/p4tools/testgen/core/small_step/table_stepper.h
```diff
@@ -0,0 +1,207 @@
...
+     * Table Variable Getter functions
+     * ========================================================================================= */
+    /// @returns the boolean-typed state variable that tracks whether given table is reached.
+    static const StateVariable& getTableReachedVar(const IR::P4Table* table);
+
+    /// @returns the state variable that tracks the value read by the key at the given index in the
+    /// given the table.
+    static const StateVariable& getTableKeyReadVar(const IR::P4Table* table, int keyIdx);
+
+    /// @returns the boolean-typed state variable that tracks whether the given table was hit. The
```
mbudiu-vmw, line 79: more precisely, if a table application results in a hit.

backends/p4tools/testgen/core/small_step/table_stepper.h
```diff
@@ -0,0 +1,207 @@
...
+    /// is Mask/Default then priority is given to constants.
+    /// For more detailed information see
+    /// https://github.com/p4lang/behavioral-model/blob/main/docs/simple_switch.md#longest-prefix-match-tables
+    int compareLPMEntries(const IR::Expression* left, const IR::Expression* right);
+
+    /// Sorts entries for lpm.
+    /// TODO: Check whether we still need this
+    IR::Vector<IR::Entry> sortLpmEntries(const IR::Vector<IR::Entry>& entries);
+
+    /// @returns whether the table is immutable, meaning control plane entries can not be added.
```
mbudiu-vmw, line 115: the default_action can also be const

backends/p4tools/testgen/core/small_step/table_stepper.h
```diff
@@ -0,0 +1,207 @@
...
+    bool checkTableIsImmutable();
+
+    /// Add the default action path to the expression stepper. If only hits if @param
+    /// tableMissCondition is true.
+    void addDefaultAction(boost::optional<const IR::Expression*> tableMissCondition);
+
+    /// Helper function that collects the list of actin elements contained in the table.
```
mbudiu-vmw, line 122: actions

backends/p4tools/testgen/core/small_step/table_stepper.h
```diff
@@ -0,0 +1,207 @@
...
+    /// immutable. This helper function unrolls all entries and aligns the table key with the match
+    /// provided by the constant entry.
+    /// @param table the table we invoke.
+    /// @returns tableMissCondition reference to the current constraints of the miss condition.
+    /// Any constant entry we hit implies that the table is hit, and the default action is not
+    /// executed.
+    const IR::Expression* evalTableConstEntries();
+
+    /// This helper function evaluates potential insertion from the control plane. We use zombie
+    /// variables to mimic an operator inserting entries. We only cover ONE entry per table for
```
mbudiu-vmw, line 174: Does the comment with ONE belong here?
So in your tests a table has always at most one entry?

backends/p4tools/testgen/core/small_step/table_stepper.h
```diff
@@ -0,0 +1,207 @@
...
+    /// now.
+    /// @param table the table we invoke.
+    void evalTableControlEntries(const std::vector<const IR::ActionListElement*>& tableActionList);
+
+    /// This is a special function that handles the case where the key of a table is tainted. This
+    /// means that the entire execution of the table is tainted. We can dramatically simplify
+    /// execution here and we also do not need control plane entries.
+    /// The tricky part is that we still need execute all possible actions and taint whatever they
+    /// may write, iff the table has constant entries.
```
mbudiu-vmw, line 183: why only if they are constant entries?

backends/p4tools/testgen/core/small_step/table_stepper.cpp
```diff
@@ -0,0 +1,662 @@
...
+                entryKey->is<IR::Range>() || entryKey->is<IR::Mask>()) {
+                entriesSorted.push_back(entry);
+            } else {
+                BUG("%1%: Unhandled entry key type: %2%", entryKey, entryKey->node_type_name());
+            }
+        }
+    }
+    const IR::Expression* left = nullptr;
+    const IR::Expression* right = nullptr;
+    for (size_t i = 0; i < entriesSorted.size(); i++) {
```
mbudiu-vmw, line 113: you can't use quicksort?

backends/p4tools/testgen/core/small_step/table_stepper.cpp
```diff
@@ -0,0 +1,662 @@
...
+        for (size_t idx = 0; idx < keys->keyElements.size(); ++idx) {
+            const auto* key = keys->keyElements.at(idx);
+            const auto* entryKey = entry->keys->components.at(idx);
+            BUG_CHECK(key->expression, "Null expression %1% for matching in the table %2%", entry,
+                      table);
+            // These always match, so do not even consider them in the equation.
+            if (entryKey->is<IR::DefaultExpression>()) {
+                continue;
+            }
+            // TODO(Fabian): This may not generalize. Find a way to handle this in z3_solver.cpp
```
mbudiu-vmw, line 272: what may not generalize?

backends/p4tools/testgen/core/small_step/table_stepper.cpp
```diff
@@ -0,0 +1,662 @@
...
+    nextState->replaceTopBody(&replacements);
+    stepper->result->emplace_back(nextState);
+}
+
+bool TableStepper::resolveTableKeys() {
+    auto propertyIdx = -1;
+    const IR::Key* key = nullptr;
+    for (propertyIdx = 0; propertyIdx < static_cast<int>(table->properties->properties.size());
+         ++propertyIdx) {
+        const auto* property = table->properties->properties.at(propertyIdx);
```
mbudiu-vmw, line 473: there is a getKey() method for tables.

backends/p4tools/testgen/core/small_step/table_stepper.cpp
```diff
@@ -0,0 +1,662 @@
...
+        // Initialize the standard keyProperties.
+        KeyProperties keyProperties(keyElement, fieldName, keyIdx, keyMatchType, keyHasTaint);
+        properties.resolvedKeys.emplace_back(keyProperties);
+    }
+    return false;
+}
+
+bool TableStepper::checkTableIsImmutable() {
+    bool isConstant = false;
+    if (table->properties->getProperty("entries") != nullptr) {
```
mbudiu-vmw, line 553: there is a getEntries method

backends/p4tools/testgen/core/small_step/table_stepper.cpp
```diff
@@ -0,0 +1,662 @@
...
+        return tableActionList;
+    }
+    for (size_t idx = 0; idx < actionList->size(); idx++) {
+        const auto* action = actionList->actionList.at(idx);
+        if (action->getAnnotation("defaultonly") != nullptr) {
+            continue;
+        }
+        CHECK_NULL(action->expression);
+        const auto* tableAction = action->expression->checkedTo<IR::MethodCallExpression>();
+        BUG_CHECK(tableAction, "Invalid action '%1%' in the table '%2%'", action, table);
```
mbudiu-vmw, line 574: this can never happen because of the checkedTo

backends/p4tools/testgen/core/small_step/extern_stepper.cpp
```diff
@@ -0,0 +1,949 @@
...
+#include "backends/p4tools/testgen/core/small_step/expr_stepper.h"
+#include "backends/p4tools/testgen/core/small_step/small_step.h"
+#include "backends/p4tools/testgen/lib/continuation.h"
+#include "backends/p4tools/testgen/lib/execution_state.h"
+
+namespace P4Tools {
+
+namespace P4Testgen {
+
+void ExprStepper::setFields(ExecutionState* nextState,
```
mbudiu-vmw, line 30: maybe this file should be called core_externs_stepper.

backends/p4tools/testgen/core/small_step/extern_stepper.cpp
```diff
@@ -0,0 +1,949 @@
...
+             } else {
+                 TESTGEN_UNIMPLEMENTED("Append input %1% of type %2% not supported", appendVar,
+                                       appendVar->type);
+             }
+             nextState->popBody();
+             result->emplace_back(nextState);
+         }},
+        /* ======================================================================================
+         *  prepend_emit_buffer
+         *  This internal extern appends the remaining slice of the input packet to the output of
```
mbudiu-vmw, line 242: you mean payload?

backends/p4tools/testgen/core/small_step/extern_stepper.cpp
```diff
@@ -0,0 +1,949 @@
...
+                IR::ID& /*methodName*/, const IR::Vector<IR::Argument>* args,
+                const ExecutionState& state, SmallStepEvaluator::Result& result) {
+             const auto* globalRef = args->at(0)->expression;
+             if (!(globalRef->is<IR::Member>() || globalRef->is<IR::PathExpression>())) {
+                 TESTGEN_UNIMPLEMENTED("Global input %1% of type %2% not supported", globalRef,
+                                       globalRef->type);
+             }
+
+             const auto* paramRef = args->at(1)->expression;
+             if (!(paramRef->is<IR::Member>() || paramRef->is<IR::PathExpression>())) {
```
mbudiu-vmw, line 305: this is a bit confusing, an argument is not a parameter. perhaps the error message needs to be changed. this is a problem below as well.

backends/p4tools/testgen/core/small_step/extern_stepper.cpp
```diff
@@ -0,0 +1,949 @@
...
+                         call);
+                 }
+                 // In the second case, where the advance amount is a runtime expression, we need to
+                 // invoke the solver.
+                 // The size of the advance expression should be smaller than the maximum packet
+                 // size.
+                 auto* sizeRestriction = new IR::Leq(
+                     advanceExpr, IRUtils::getConstant(advanceExpr->type,
+                                                       ExecutionState::getMaxPacketLength_bits()));
+                 // The advance expression should ideally fit into a multiple of 8 bits.
```
mbudiu-vmw, line 552: fit or have a size that is?

backends/p4tools/testgen/core/small_step/extern_stepper.cpp
```diff
@@ -0,0 +1,949 @@
...
+                 auto* bytesRestriction = new IR::Equ(
```
mbudiu-vmw, line 553: I don't see why that's true.

backends/p4tools/common/lib/saturation_elim.h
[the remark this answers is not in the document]
fruffy-bfn: This file has a couple dependencies on IRUtils. First we need to move IRUtils to somewhere in the compiler, then we can port the utility passes.

backends/p4tools/common/lib/taint.h
```diff
@@ -0,0 +1,38 @@
...
+#include "backends/p4tools/common/lib/model.h"
+#include "ir/ir.h"
+
+namespace P4Tools {
+
+class Taint {
+ public:
+    static const IR::StringLiteral taintedStringLiteral;
+
+    /// @returns the mask for the corresponding program packet. In the first step, the packet
```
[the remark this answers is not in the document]
fruffy-bfn: We have a PR coming up that will clean up these functions. It might merge it into this PR.

backends/p4tools/testgen/targets/bmv2/test_spec.h
```diff
@@ -0,0 +1,191 @@
...
+
+    /// Each element is an API name paired with a match rule.
+    std::vector<Bmv2RegisterCondition> registerConditions;
+
+ public:
+    explicit Bmv2RegisterValue(const IR::Expression* initialValue);
+
+    cstring getObjectName() const override;
+
+    /// Each element is an API name paired with a match rule.
```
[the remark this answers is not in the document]
fruffy-bfn: Yes. but this is not implemented yet.

backends/p4tools/testgen/targets/bmv2/test_spec.cpp
```diff
@@ -0,0 +1,175 @@
...
+void Bmv2RegisterValue::addRegisterCondition(Bmv2RegisterCondition cond) {
+    registerConditions.push_back(cond);
+}
+
+const IR::Expression* Bmv2RegisterValue::getInitialValue() const { return initialValue; }
+
+cstring Bmv2RegisterValue::getObjectName() const { return "Bmv2RegisterValue"; }
+
+const IR::Expression* Bmv2RegisterValue::getCurrentValue(const IR::Expression* index) const {
+    const IR::Expression* baseExpr = initialValue;
```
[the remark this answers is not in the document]
fruffy-bfn: We do not have the index we want to access when we push a new register condition. Is this what you mean?

backends/p4tools/testgen/targets/bmv2/test_spec.cpp
```diff
@@ -0,0 +1,175 @@
...
+        const auto* storedVal = bmv2registerValue.value;
+        baseExpr =
+            new IR::Mux(baseExpr->type, new IR::Equ(storedIndex, index), storedVal, baseExpr);
+    }
+    return baseExpr;
+}
+
+const IR::Constant* Bmv2RegisterValue::getEvaluatedValue() const {
+    const auto* constant = initialValue->to<IR::Constant>();
+    BUG_CHECK(constant, "Variable is not a constant, has the test object %1% been evaluated?",
```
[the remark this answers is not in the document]
fruffy-bfn: Yes, they can be tainted. This error message is rather generic and specialized per test  object. I can try to specialize them.

backends/p4tools/testgen/core/small_step/cmd_stepper.cpp
```diff
@@ -0,0 +1,644 @@
...
+    } else if (const auto* p4control = typeDecl->to<IR::P4Control>()) {
+        params = p4control->getApplyParameters();
+        nextState->pushNamespace(p4control);
+    } else {
+        TESTGEN_UNIMPLEMENTED("Constructed type %s of type %s not supported.", typeDecl,
+                              typeDecl->node_type_name());
+    }
+    for (size_t paramIdx = 0; paramIdx < params->size(); ++paramIdx) {
+        const auto* param = params->getParameter(paramIdx);
+        const auto* paramType = param->type;
```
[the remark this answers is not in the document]
fruffy-bfn: In this case, yes. The copy_in and copy_out helpers will reset the parameters appropriately, this is a helper function to initialize all the block parameters as symbolic variables.

backends/p4tools/testgen/core/small_step/cmd_stepper.cpp
```diff
@@ -0,0 +1,644 @@
...
+        replacements.emplace_back(decl);
+    }
+    // Add the body, if it is not empty
+    if (p4control->body != nullptr) {
+        replacements.emplace_back(p4control->body);
+    }
+    // Remove the invocation of the control from the current body and push the resulting
+    // continuation onto the continuation stack.
+    nextState->popBody();
+    // Exit terminates the entire control pipeline (only the control).
```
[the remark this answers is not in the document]
fruffy-bfn: The way we have modelled the exit continuation and exception is such that copy_out is still executed.

backends/p4tools/testgen/core/small_step/expr_stepper.cpp
fruffy-bfn: I believe we ran into this. If not I will remove it but I need to check.

backends/p4tools/testgen/core/small_step/expr_stepper.cpp
fruffy-bfn: There is a PR in-flight that rewrites this function, which should also include simplifications. There is some more issues with it. For example, default expressions and taint expressions are not handled quite correctly.

backends/p4tools/testgen/core/small_step/table_stepper.h
fruffy-bfn: But that just means that we can not override the default action, right?

backends/p4tools/testgen/core/small_step/table_stepper.h
fruffy-bfn: Yup, covering exactly one table path. We may change this in the future but for now this is how the interpreter is designed.

backends/p4tools/testgen/core/small_step/table_stepper.h
fruffy-bfn: If the table keys are tainted and the table does not have constant entries, we simply do not insert a table entry and execute the default action, which is a safe choice.

backends/p4tools/testgen/core/small_step/extern_stepper.cpp
fruffy-bfn: https://github.com/p4lang/behavioral-model/issues/727#issuecomment-470648582

It is the case for BMv2 only, but we have this as a general restriction until we specialize the advance for targets.

backends/p4tools/testgen/targets/bmv2/test_spec.cpp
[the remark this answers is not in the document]
mbudiu-vmw: You push the register condition on a stack, and when you try to access it you build a Mux expression. Couldn't you just directly build the Mux expression here?

backends/p4tools/testgen/core/small_step/table_stepper.h
mbudiu-vmw: yes, the same as for the entries

jfingerh: I believe that whether the entries are const (i.e. not changeable by the control plane), and whether the default_action is const (i.e. not changeable by the control plane) are independent properties of a P4 table.  All four combinations of (const entries, const default_action), (const entries, non-const default_action), (non-const entries, const default_action), and (non-const entries, non-const default_action) are supported by p4c and the language spec.

When I say "non-const entries" here, I simply mean that there is no 'entries' property of the table at all in the P4 source code.  Although for future reference, there are people who are encouraging me to help define "entries = { ... }" without `const` for P4 soon-ish.  There is no such language spec definition or p4c implementation for that today that I am aware of.

backends/p4tools/testgen/targets/bmv2/test_spec.cpp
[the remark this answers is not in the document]
fruffy-bfn: No because we use the index as the condition of the argument. The read index may be different every time, but the "writes" of the register were done to a particular register index.  

For example, you could push `( index: "hdr.h.idx", val: "hdr.h.a")` into the register and later you may want to read using index 1. With the model of the function this will return  `(hdr.h.idx == 1 ? : hdr.h.a : default)`. 

You could also have a symbolic index you read from (hdr.h.jdx). Then the return value will look like  `(hdr.h.idx == hdr.h.jdx ? : hdr.h.a : default)`. 

Essentially, what we are trying to return here is the latest symbolic write to the register given a specific index.

backends/p4tools/testgen/core/small_step/table_stepper.cpp
fruffy-bfn: This method returns and EntriesList, which does not have the `isConstant` member.

backends/p4tools/common/lib/symbolic_env.cpp
```diff
@@ -0,0 +1,146 @@
...
+     public:
+        explicit SubstVisitor(const SymbolicEnv& symbolicEnv) : symbolicEnv(symbolicEnv) {}
+    };
+
+    return expr->apply(SubstVisitor(*this));
+}
+
+const SymbolicMapType& SymbolicEnv::getInternalMap() const { return map; }
+
+bool SymbolicEnv::isSymbolicValue(const IR::Node* node) {
```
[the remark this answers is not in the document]
fruffy: I tried to write a visitor for this but it is not straightforward. The visitor will likely be as verbose as this function.

backends/p4tools/testgen/core/small_step/table_stepper.cpp
fruffy: @OleksandrLetychevskyiIntel can you please check this?

backends/p4tools/testgen/core/small_step/extern_stepper.cpp
fruffy: It also contains internal externs. Might make sense to split things up.

backends/p4tools/testgen/core/small_step/expr_stepper.cpp
fruffy-bfn: Is there another methodCallExpression that can reduce produce .hit? Otherwise we would need to check for the type.

mbudiu-vmw: it depends on the architectural model and whether it is extensible. In ebpf you can declare your own extern functions (but not methods). So I guess this is good enough for now.

backends/p4tools/testgen/core/small_step/cmd_stepper.cpp
```diff
@@ -0,0 +1,644 @@
...
+    if (leftType->is<IR::Type_Name>()) {
+        leftType = state.resolveType(leftType->to<IR::Type_Name>());
+    }
+    // Although we typically expand structure assignments into individual member assignments using
+    // the copyHeaders pass, some extern functions may return a list or struct expression. We can
+    // not always expand these return values as we do with the expandLookahead pass.
+    // Correspondingly, we need to retrieve the fields and set each member individually. This
+    // assumes that all headers and structures have been flattened and no nesting is left.
+    if (const auto* structType = leftType->to<IR::Type_StructLike>()) {
+        const auto* listExpr = assign->right->checkedTo<IR::ListExpression>();
```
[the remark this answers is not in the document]
fruffy: Yes, this check can not be removed because there are extern that return a ListExpression.

backends/p4tools/testgen/lib/continuation.h
```diff
@@ -0,0 +1,214 @@
...
+        /// Delegates to IR equality.
+        bool operator==(const Return& other) const;
+
+        Return() : expr(boost::none) {}
+        explicit Return(const IR::Node* expr);
+    };
+
+    /// Alias for various property types that can be set. We restrict these to keep the feature
+    /// simple.
+    using StateProperty = boost::variant<cstring, uint64_t, int64_t, bool, const IR::Expression*>;
```
mbudiu-vmw, line 88: why call this StateProperty and not PropertyValue?

backends/p4tools/testgen/lib/continuation.h
```diff
@@ -0,0 +1,214 @@
...
+        StateProperty property;
+
+        /// Delegates to the equality of the property.
+        bool operator==(const PropertyUpdate& other) const;
+
+        explicit PropertyUpdate(cstring propertyName, StateProperty property);
+    };
+
+    /// Guards are commands that can be inserted into the continuation stack to ensure that the
+    /// interpreter does not violate a given invariant. If the condition evaluates to false, the
```
mbudiu-vmw, line 103: which branch is discarded?

backends/p4tools/testgen/lib/test_spec.h
```diff
@@ -0,0 +1,384 @@
...
+    virtual const TestObject* evaluate(const Model& model) const = 0;
+};
+
+/* =========================================================================================
+ *  Packet
+ * ========================================================================================= */
+
+class Packet : public TestObject {
+ private:
+    /// The port associated with this packets.
```
mbudiu-vmw, line 47: input port I assume.

backends/p4tools/testgen/lib/test_spec.h
```diff
@@ -0,0 +1,384 @@
...
+    /// @returns the match value. It is expected to be a constant at this point.
+    /// A BUG is thrown otherwise.
+    const IR::Constant* getEvaluatedValue() const;
+};
+
+using FieldMatch = boost::variant<Exact, Ternary, LPM, Range>;
+
+class TableRule : public TestObject {
+ private:
+    /// Each element is an API name paired with a match rule.
```
mbudiu-vmw, line 250: what is an API name?

backends/p4tools/testgen/lib/test_spec.h
```diff
@@ -0,0 +1,384 @@
...
+/* =========================================================================================
+ * Test Specification
+ * ========================================================================================= */
+
+class TestSpec {
+ private:
+    /// The input packet of the test.
+    const Packet ingressPacket;
+
+    /// The expected output packet of this test.
```
mbudiu-vmw, line 327: I guess no multicast is possible

backends/p4tools/testgen/lib/test_spec.h
```diff
@@ -0,0 +1,384 @@
...
+    const boost::optional<Packet> egressPacket;
+
+    /// The traces that have been collected while the interpreter stepped through the program.
+    const std::vector<gsl::not_null<const TraceEvent*>> traces;
+
+    /// A map of table properties. For example, tables objects or action profiles.
```
mbudiu-vmw, line 333: key is table name?

backends/p4tools/testgen/lib/test_spec.h
```diff
@@ -0,0 +1,384 @@
...
+    std::map<cstring, std::map<cstring, const TestObject*>> testObjects;
+
+ public:
+    TestSpec(Packet ingressPacket, boost::optional<Packet> egressPacket,
+             std::vector<gsl::not_null<const TraceEvent*>> traces);
+
+    /// Add a test object to the test specification with @param category as the object category
+    /// ("tables", "registers") and objectLabel as the label
```
mbudiu-vmw, line 341: at this point I can't guess what a category is

backends/p4tools/testgen/lib/test_spec.cpp
```diff
@@ -0,0 +1,340 @@
...
+        BUG("Unable to find test object with the label %1%. ", propertyName);
+    }
+    return nullptr;
+}
+
+void TableConfig::addTableProperty(cstring propertyName, const TestObject* property) {
+    tableProperties[propertyName] = property;
+}
+
+const TableConfig* TableConfig::evaluate(const Model& model) const {
```
mbudiu-vmw, line 274: Based on the file name I would not have guessed that this file contains all these objects.

backends/p4tools/testgen/lib/execution_state.h
```diff
@@ -0,0 +1,476 @@
...
+    /// becomes the top of the stack.
+    ///
+    // Invariant: if the @body is empty, then so is this, and this state is terminal.
+    std::stack<gsl::not_null<const StackFrame*>> stack;
+
+    /// State properties are bools, integers, or strings that can be set and propagated across
+    /// execution state. They are used to influence execution along a particular continuation path.
+    /// For example, we can set the "inUndefinedState" property to indicate that all state being
+    /// written while this variable is active is tainted. This property must be unset manually to
+    /// resume normal operation. Usually, the unset call is inserted directly after the tainted
```
mbudiu-vmw, line 115: which unset call?

backends/p4tools/testgen/lib/execution_state.h
```diff
@@ -0,0 +1,476 @@
...
+
+    /// The parserErrorLabel is set by the parser to indicate the variable corresponding to the
+    /// parser error that is set by various built-in functions such as verify or extract.
+    const IR::Member* parserErrorLabel = nullptr;
+
+    /// This variable tracks how much of the input packet has been parsed.
+    /// Typically, this is equivalent to the size of the input packet. However, there may be
+    /// functions such as lookahead that peek into the program packet without actually advancing the
+    /// cursor. The cursor ensures that parsing remains consistent in these cases.
+    int inputPacketCursor = 0;
```
mbudiu-vmw, line 135: why not size_t?

mbudiu-vmw (commented):
I am down to a dozen files left to review

backends/p4tools/testgen/lib/execution_state.h
fruffy-bfn: Because this variable is also incremented by `width_bits()`, which has the return type `int`. It might make sense to return `size_t` for `width_bits` in general. This is a very common type mismatch in the compiler.

backends/p4tools/testgen/lib/exceptions.h
```diff
@@ -0,0 +1,41 @@
...
+namespace P4Testgen {
+
+/// This class indicates a feature that is not implemented in P4Testgen.
+/// Paths with this unimplemented feature should be skipped
+class TestgenUnimplemented final : public Util::P4CExceptionBase {
+ public:
+    template <typename... T>
+    explicit TestgenUnimplemented(const char* format, T... args)
+        : P4CExceptionBase(format, args...) {
+        // Check if output is redirected and if so, then don't color text so that
```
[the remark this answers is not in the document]
fruffy-bfn: This is actually from the original `/lib/exceptions.h`.

fruffy-bfn:
I left unresolved comments open. Do any of them still need to be addressed?

mbudiu-vmw:
> I left unresolved comments open. Do any of them still need to be addressed? I will work on the eBPF comments after #3495 is merged.

With so many commits it is impossible for me to keep track of which comments have been addressed.

mbudiu-vmw:
1. If a comment will require future work I think you should file a corresponding issue
2. I will try to finish reviewing the PR this week
3. I would suggest writing a short top-level readme explaining the core structure of the test generator: how it symbolically executes a program representing the results of evaluation as expressions using the IR itself, and how it uses Z3 to explore various paths, etc. This could also be a short guide to the code structure. Having that would have helped me a lot. I am still piecing my understanding as a puzzle.

jnfoster:
We can share a draft paper with you privately (would love your feedback!)

We are working on some public documentation -- more focused on the UI than the internals, but can include a bit of that too.

mbudiu-vmw:
Papers are great, but a short README is still necessary, especially if the paper does not talk about the code structure.

jnfoster:
Agree!

backends/p4tools/testgen/targets/bmv2/expr_stepper.cpp
```diff
@@ -0,0 +1,1742 @@
...
+         *  statement failed, and optionally other information about the
+         *  failure.
+         *  For example, on the simple_switch target, executing an assert
+         *  statement with a false argument causes a log message with the file
+         *  name and line number of the assert statement to be printed, and
+         *  then the simple_switch process exits.
+         *  If you provide the --ndebug command line option to p4c when
+         *  compiling, the compiled program behaves as if all assert statements
+         *  were not present in the source code.
+         *  We strongly recommend that you avoid using expressions as an
```
mbudiu-vmw, line 252: I would expect that the front-end pass which eliminates side-effects will actually pull them out in a separate call.

backends/p4tools/testgen/targets/bmv2/expr_stepper.cpp
```diff
@@ -0,0 +1,1742 @@
...
+                 nextState->set(fieldRef,
+                                programInfo.createTargetUninitialized(resultField->type, true));
+                 nextState->popBody();
+                 result->emplace_back(nextState);
+                 return false;
+             }
+             BUG("Not a Type_Bits: %1%", resultField->type);
+         }},
+        /* ======================================================================================
+         *  assume
```
mbudiu-vmw, line 208: this is great documentation, it should not be only in code, it should be in some documents that users can read.

backends/p4tools/testgen/targets/bmv2/expr_stepper.cpp
```diff
@@ -0,0 +1,1742 @@
...
+         *  v1model architecture documentation (Note 1) for more details.
+         * ====================================================================================== */
+        {"*method.clone_preserving_field_list",
+         {"type", "session", "data"},
+         [this](const IR::MethodCallExpression* call, const IR::Expression* /*receiver*/,
+                IR::ID& /*methodName*/, const IR::Vector<IR::Argument>* args,
+                const ExecutionState& state, SmallStepEvaluator::Result& result) {
+             uint64_t recirculateCount = 0;
+             // Grab the recirculate count. Stop after more than 1 circulation loop to avoid
+             // infinite recirculation loops.
```
mbudiu-vmw, line 793: Is this part of the bmv2 spec?

backends/p4tools/testgen/targets/bmv2/expr_stepper.cpp
```diff
@@ -0,0 +1,1742 @@
...
+                 result->emplace_back(nextState);
+                 return;
+             }
+
+             auto cloneType = args->at(0)->expression->checkedTo<IR::Constant>()->asUint64();
+             const auto* sessionIdExpr = args->at(1)->expression;
+             auto recirculateIndex = args->at(2)->expression->checkedTo<IR::Constant>()->asUint64();
+             boost::optional<const Constraint*> cond = boost::none;
+
+             if (cloneType == BMv2Constants::CLONE_TYPE_I2E) {
```
mbudiu-vmw, line 842: does this make sure that the effects of the clone happen at the end of the pipeline?

backends/p4tools/testgen/targets/bmv2/expr_stepper.cpp
```diff
@@ -0,0 +1,1742 @@
...
+                 // Pick a clone port var. For now, pick a random value from 0-511.
+                 const auto* rndConst =
```
mbudiu-vmw, line 844: In some sense this is an executable specification of BMv2.
I wonder whether one could take advantage of that in some way besides test generation. For example, why do you need bmv2 at all, can you replace it with this implementation?

backends/p4tools/testgen/targets/bmv2/expr_stepper.cpp
```diff
@@ -0,0 +1,1742 @@
...
+             // Set the appropriate instance type, which will be processed by "check_recirculate".
+             nextState->setProperty("recirculate_instance_type",
+                                    BMv2Constants::PKT_INSTANCE_TYPE_RECIRC);
+             nextState->popBody();
+             result->emplace_back(nextState);
+         }},
+        /* ======================================================================================
+         *  clone
+         *  clone is in most ways identical to the clone_preserving_field_list
+         *  operation, with the only difference being that it never preserves
```
mbudiu-vmw, line 1062: right, could the code above be reused instead?

mbudiu-vmw (approved):
I am done. This looks very nice.

backends/p4tools/testgen/targets/bmv2/expr_stepper.cpp
fruffy-bfn: This documentation is actually part of the v1model.p4 file, and most of this documentation was written by @jafingerhut. I do not want to take credit for that. :) 

The documentation for v1model was actually very helpful for implementing the semantics of the architecture.

lib/castable.h
```diff
@@ -17,14 +17,44 @@ limitations under the License.
 #ifndef _LIB_CASTABLE_H_
 #define _LIB_CASTABLE_H_
 
+#include <typeinfo>
```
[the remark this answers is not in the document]
fruffy-bfn: Yes, this was after the merge. The castable file was missing `checkedTo`.

backends/p4tools/testgen/targets/bmv2/table_stepper.cpp
```diff
@@ -0,0 +1,326 @@
...
+    // For now, we consider optional match types to be a no-op.
+    if (keyProperties.matchType == BMv2Constants::MATCH_KIND_OPT) {
+        return hitCondition;
+    }
+    // Action selector entries are not part of the match.
+    if (keyProperties.matchType == BMv2Constants::MATCH_KIND_SELECTOR) {
+        bmv2_V1ModelProperties.actionSelectorKeys.emplace_back(keyExpr);
+        return hitCondition;
+    }
+    // Ranges are not yet implemented for Tofino STF tests.
```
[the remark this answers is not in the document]
fruffy-bfn: The reason why we check the test framework here is so that we can fail early instead of stepping through the entire program. Also in some cases, we can have a failover mechanism if the test framework does not support a specific construct.

backends/p4tools/testgen/targets/bmv2/expr_stepper.cpp
fruffy-bfn: Well, we can only give you the expected output for a given input (packet and control plane config) but BMv2 also models all the other parts of the stack such as the control plane API etc. So are a little bit more limited here.

backends/p4tools/testgen/targets/bmv2/expr_stepper.cpp
fruffy-bfn: Practically speaking clone can only be called in ingress or egress. This is enforced by the compiler. So we have some leeway here since the egress block will immediately follow in the I2E case.

backends/p4tools/testgen/targets/bmv2/expr_stepper.cpp
fruffy-bfn: No, in fact BMv2 seems to circulate more rounds. I do not have the exact count. Will add this as a TODO.

backends/p4tools/testgen/targets/bmv2/concolic.cpp
```diff
@@ -0,0 +1,332 @@
...
+             (*resolvedConcolicVariables)[concolicMember] =
+                 IRUtils::getConstant(checksumVarType, computedResult);
+
+         } else {
+             TESTGEN_UNIMPLEMENTED("Checksum output %1% of type %2% not supported", checksumVar,
+                                   checksumVar->type);
+         }
+         // Generated equations for all the variables that were assigned a value in this iteration
+         // of concolic execution.
+         for (const auto& variable : resolvedExpressions) {
```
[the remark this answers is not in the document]
fruffy-bfn: There might be multiple expressions that have been resolved for a single concolic variable. 

For example, for hash functions, return parameter is a concolic variable, but the inputs {h.eth_hdr.dst_addr, h.eth_hdr.src_addr, h.eth_hdr.eth_type) are `resolvedExpressions` that need to be resolved and bound via an equation to compute a solution.

backends/p4tools/testgen/targets/bmv2/expr_stepper.cpp
jfingerh: I am pretty sure that bmv2 does not enforce any limit on the number of recirculations or clones of a packet.

backends/p4tools/testgen/targets/ebpf/backend/stf/stf.cpp
```diff
@@ -0,0 +1,296 @@
...
+    rulesJson["matches"] = inja::json::array();
+    rulesJson["act_args"] = inja::json::array();
+    rulesJson["needs_priority"] = false;
+
+    for (auto const& match : matches) {
+        auto fieldName = match.first;
+        auto const& fieldMatch = match.second;
+
+        // Replace header stack indices hdr[<index>] with hdr$<index>.
+        // TODO: This is a limitation of the stf parser. We should fix this.
```
fruffy, line 98: The STF parser can not handle header stacks as key names. We have to rewrite them. This is a strange limitation we should fix at some point.

backends/p4tools/testgen/targets/ebpf/test/EBPFXfail.cmake
```diff
@@ -0,0 +1,45 @@
+# XFAILS: tests that currently fail. Most of these are temporary.
+# ================================================
+
+# The code generated by the compiler is invalid.
+p4tools_add_xfail_reason(
+  "testgen-p4c-ebpf"
+  "Error 2: Failed to build the filter:"
+  lpm_ebpf.p4
```
fruffy, line 8: These are compilation failures triggered by P4Testgen. It seems like there are issue with the synthesized control plane interface.

fruffy (commented):
@mbudiu-vmw I cleaned and fixed up the eBPF back end for P4Testgen. Maybe if you have time could you give this a second check? Otherwise I will merge it. It seems to expose some problems with the test framework. I am thinking of fixing these in a separate PR.

mbudiu-vmw:
How about removing the "Draft" label?

backends/p4tools/testgen/core/small_step/extern_stepper.cpp
```diff
@@ -776,6 +786,28 @@ void ExprStepper::evalExternMethodCall(const IR::MethodCallExpression* call,
...
+         *  some target architectures.
+         * ======================================================================================
+         */
+        {"packet_in.length",
+         {},
+         [](const IR::MethodCallExpression* /*call*/, const IR::Expression* /*receiver*/,
+            IR::ID& /*name*/, const IR::Vector<IR::Argument>* /*args*/, const ExecutionState& state,
+            SmallStepEvaluator::Result& result) {
+             auto* nextState = new ExecutionState(state);
+             // TODO: Which packet length do we actually get here?
```
mbudiu-vmw, line 801: the length is not changed by the parsing.

backends/p4tools/testgen/targets/ebpf/test/EBPFXfail.cmake
mbudiu-vmw: Many

backends/p4tools/testgen/targets/ebpf/table_stepper.h
```diff
@@ -0,0 +1,57 @@
...
+
+namespace P4Tools {
+
+namespace P4Testgen {
+
+namespace EBPF {
+
+class EBPFTableStepper : public TableStepper {
+ private:
+    /// Specifies the type of the table implementation:
```
mbudiu-vmw, line 24: I don't understand what is the relationship between these enums and ebpf.
Ebpf has only 2 implementation properties, array or hash.

backends/p4tools/testgen/targets/ebpf/backend/stf/stf.h
```diff
@@ -0,0 +1,79 @@
...
+#include "backends/p4tools/testgen/targets/ebpf/test_spec.h"
+
+namespace P4Tools {
+
+namespace P4Testgen {
+
+namespace EBPF {
+
+/// Extracts information from the @testSpec to emit a STF test case.
+class STF : public TF {
```
mbudiu-vmw, line 28: I repeat the question about what in this class is specific to EBPF. It looks to me like STF should be the same for all backends.

mbudiu-vmw (approved):

backends/p4tools/testgen/targets/ebpf/backend/stf/stf.h
fruffy: There are subtle difference between how the STF tests are processed for each back end. For example, BMv2 does not need [this](https://github.com/p4lang/p4c/pull/3510/files/2fe95c34809d2682684aee35790cdff58e73a0c3#diff-792eb6879e16c8c09289fd83c718ffa32b0baa5ea1235157fa0d13dfde2703afR98) and does support `$` as a postfix to indicate exact packet length matching. Furthermore, BMv2 supports various externs such as clone, action profiles, registers, which are handled in this stf class.

backends/p4tools/testgen/targets/ebpf/table_stepper.h
fruffy: Yes, eBPF does not have a lot of additional table properties or implementations. We could add hash and array to this enum, but they are not changing the semantics of entry generation for tests if I recall correctly.

pkotikal (approved):

PR #3512, opened 2022-09-05
Fix: prevent copy elimination for statement which does not support constant values

kamleshbhalui:
dpdk does not expect operands of some of the statements as constant this pr prevent copy prop and elimination for those stmts.

### DIFF
--- a/backends/dpdk/dpdkAsmOpt.cpp
+++ b/backends/dpdk/dpdkAsmOpt.cpp
@@ -370,9 +370,6 @@ CopyPropagationAndElimination::copyPropAndDeadCodeElim(
         } else if (auto jne = stmt->to<IR::DpdkJmpNotEqualStatement>()) {
             instr.push_back(new IR::DpdkJmpNotEqualStatement(jne->label,
                         replaceIfCopy(jne->src1, false), replaceIfCopy(jne->src2)));
-       } else if (auto l = stmt->to<IR::DpdkLearnStatement>()) {
-            instr.push_back(new IR::DpdkLearnStatement(l->action,
-                        replaceIfCopy(l->timeout, false), replaceIfCopy(l->argument, false)));
        } else if (auto m = stmt->to<IR::DpdkMeterExecuteStatement>()) {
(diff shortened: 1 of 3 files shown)


backends/dpdk/dpdkAsmOpt.cpp
```diff
@@ -355,22 +355,100 @@ CopyPropagationAndElimination::copyPropAndDeadCodeElim(
         } else if (auto jc = stmt->to<IR::DpdkJmpLessStatement>()) {
             instr.push_back(new IR::DpdkJmpLessStatement(jc->label,
                         replaceIfCopy(jc->src1, false), replaceIfCopy(jc->src2)));
+        } else if (auto jc = stmt->to<IR::DpdkJmpLessOrEqualStatement>()) {
```
mbudiu-vmw, line 358: this implementation now has a lot of redundancy, I suspect you could do much better by using a visitor with a preorder rmethod on the base classes. But this will work.

backends/dpdk/dpdkAsmOpt.h
```diff
@@ -329,21 +329,32 @@ class CollectUseDefInfo : public Inspector {
...
         return false;
     }
 
     bool preorder(const IR::DpdkBinaryStatement *b) override {
         usesInfo[b->src1->toString()]++;
         usesInfo[b->src2->toString()]++;
         defInfo[b->dst->toString()]++;
+        // dst and src1 can not be eliminated, because both are same
+        // and dpdk does not allow src1 to be constant
+        dontEliminate[b->dst->toString()] = true;
```
mbudiu-vmw, line 356: it looks to me like this could be a helper function which takes an expression.
that could be useful particularly if later you decide to change the way you represent the "dontEliminate" information.

mbudiu-vmw (approved):

backends/dpdk/dpdkAsmOpt.cpp
kamleshbhalui: I would like to create a new PR for refactoring.

PR #3513, opened 2022-09-05
cmake: Fixed multi line comment

davidbolvansky:
Currently, this variable is broken - if you print BUILD_USE_COLOR you will see:
-- BUILD_USE_COLOR= OFF;CACHE;BOOL;Use color in C++ compiler output (even if ;the compiler does not detect terminal, e.g. when using ccache

So even if BUILD_USE_COLOR is OFF, inside if(...) this variable as evaluated as true ("OFF;CACHE;BOOL;Use color in C++ compiler output (even if ;the compiler does not detect terminal, e.g. when using ccache" is not empty")

Solution: Use \ for line continuation.


mbudiu-vmw (approved):

davidbolvansky:
Thanks

PR #3515, opened 2022-09-06
Silence -Warray-bounds for generated code for protobuf

davidbolvansky:
Reason: https://github.com/protocolbuffers/protobuf/issues/7140

As there is no easy way to workaround it (generated code), silence it, to avoid breakages eg. for builds with -Werror.


mbudiu-vmw (approved):

davidbolvansky:
Thanks!

PR #3521, opened 2022-09-08
Fix hdr align

kamleshbhalui:
Properly discard bits when casting to lower sizes
Fix typeMap for constants


backends/dpdk/dpdkArch.cpp
```diff
@@ -474,6 +474,29 @@ bool CollectMetadataHeaderInfo::preorder(const IR::Type_Struct *s) {
...
+        bool all_hdr_field_aligned = true;
+        for (auto field : st->fields) {
+            if ((*field).type->is<IR::Type_Bits>()) {
+                auto t = (*field).type->to<IR::Type_Bits>();
+                auto width = t->width_bits();
+                size_sum_so_far += width;
+                if ((width & 0x7) != 0) {
+                    all_hdr_field_aligned = false;
+                }
+            }
```
mbudiu-vmw, line 488: and if it's not bit<> you do nothing?

mbudiu-vmw (approved):

backends/dpdk/dpdkArch.cpp
kamleshbhalui: it's always bit<> at this point.

kamleshbhalui: all fields will be of bit<> type at this point.

mbudiu-vmw: that's what BUG is for.

kamleshbhalui: yes done

PR #3524, opened 2022-09-10
Dpdk backend: Flatten header union into individual fields

usha1830:
This PR supercedes https://github.com/p4lang/p4c/pull/3463
The header union should be flattened into its individual elements and the occurrences of header union variables should be replaced accordingly.
For ex:

 ```
 header_union U {
      Hdr1 h1;
      Hdr2 h2;
  }
 
  struct Headers {
      Hdr1 h1;
      U u;
  }
```
is replaced by

```
  struct  Headers {
      Hdr1 h1;
      Hdr1 u_h1;
      Hdr2 u_h2;
  }
```
The validity semantic should be also be handled. At any point in time, only one element of header union can be valid. So whenever there is an assignment to any element of header union, it should be marked valid using setValid and the other elements should be marked invalid by calling setInvalid method.
Added code comments for the details of the cases that are handled.

### CODE REVIEW

File: midend/flattenUnions.h [hunk 1]
```diff
@@ -0,0 +1,166 @@
...
+ * }
+ *
+ * is replaced by
+ * struct  Headers {
+ *     Hdr1 h1;
+ *     Hdr1 u_h1;
+ *     Hdr2 u_h2;
+ * }
+ */
+class DoFlattenHeaderUnion : public Transform {
```
mbudiu-vmw, line 48: this is a very useful pass, I will try to use it in the ebpf backend too

testdata/p4_16_samples_outputs/pna-dpdk-bvec_union-midend.p4
```diff
@@ -0,0 +1,105 @@
...
+
+header Hdr2 {
+    bit<16> _b0;
+    bit<1>  _row_alt0_valid1;
+    bit<7>  _row_alt0_port2;
+    bit<1>  _row_alt1_valid3;
+    bit<7>  _row_alt1_port4;
+}
+
+header_union U {
```
mbudiu-vmw, line 34: shouldn't this be gone?

mbudiu-vmw (commented):
I will continue the review after the answer to my question

testdata/p4_16_samples_outputs/pna-dpdk-bvec_union-midend.p4
usha1830: This output is coming from p4test backend. Currently, this pass is only invoked from DPDK midend. I quickly tried enabling this in p4test midend and see many regressions. I will fix those and update the PR.

backends/p4test/CMakeLists.txt
```diff
@@ -106,6 +106,9 @@ set (P4_XFAIL_TESTS
   testdata/p4_16_samples/pna-example-tcp-connection-tracking-err-1.p4
   testdata/p4_16_samples/pna-example-tcp-connection-tracking-err.p4
   testdata/p4_16_samples/pna-example-pass-2.p4
+  testdata/p4_16_samples/invalid-hdr-warnings6.p4
+  testdata/p4_16_samples/invalid-hdr-warnings7.p4
```
usha1830, line 110: Handling validity flags for header union instances introduces if condition. These case have started failing because the header union instance is present within action and conditional statements are not supported within action for p4test backend. Perhaps, we should enable the support of conditional statements within action for p4test as well.

mbudiu-vmw: the frontend has no problem with that, we need code generation in the back-end.
You could try the ebpf backend, I think that one supports already conditions in actions.
For bmv2 we never got around to implementing that, although there is support.

usha1830: DPDK also supports conditions in action and some of these tests are ported to PNA and are added to the DPDK regressions. 
We could try enabling this pass for other backends . However, I would prefer to limit this PR to dpdk and p4test backends.

usha1830:
@mbudiu-vmw 1 test is still failing with the p4test and I am looking into it. It would be great if you could provide your inputs/opinion on the overall structure of this pass. I have used some passes from frontend  before and after these new passes to optimize the generated output. Many output files have changed due to this.

mbudiu-vmw:
The error you are getting for issue982 suggests that your elimination pass does not remove all references to the eliminated unions.

backends/p4test/midend.cpp
```diff
@@ -103,6 +104,7 @@ MidEnd::MidEnd(CompilerOptions& options, std::ostream* outStream) {
         new P4::RemoveSelectBooleans(&refMap, &typeMap),
         new P4::FlattenHeaders(&refMap, &typeMap),
         new P4::FlattenInterfaceStructs(&refMap, &typeMap),
+        new P4::FlattenHeaderUnion(&refMap, &typeMap),
```
mbudiu-vmw, line 109: which pass fails? we can just remove the Predication pass if it's the one

usha1830:
> The error you are getting for issue982 suggests that your elimination pass does not remove all references to the eliminated unions.
It was occurring when the struct containing header_union type field is typedef-ed and the new type is used in the program. Invoking EliminateTypedef before this header union elimination fixes this error. Updated PR with fix

backends/p4test/midend.cpp
usha1830: After invoking typedef elimination before header union flattening, all tests are passing on my local machine

usha1830: > which pass fails? we can just remove the Predication pass if it's the one

Removing Predication  pass the tests which are added to p4test's XFAIL list due to conditional statement within action.  Regressions also pass if reference output files are updated. 

Are you suggesting we can safely remove Predication pass? Or should we do in another PR to avoid any side-effect that it might cause.

mbudiu-vmw: yes, we can remove it. this PR or another one isn't much of a difference really
we could also add a flag so it's applied only to controls. or perhaps a flag so that it only gives a warning and an error...

usha1830:
validate-p4c is failing. I don't understand the failure log from validate-p4c. Could somebody point me to any documentation on how to read this output?

usha1830:
> validate-p4c is failing. I don't understand the failure log from validate-p4c. Could somebody point me to any documentation on how to read this output?

@mbudiu-vmw 
Attaching the diff report of --top4 output before and after header union flattening pass for some of the tests that are failing in validate-p4c check. 
Could you please take a look and help me find the incorrect transformation which is causing these failures. 
[Report1.pdf](https://github.com/p4lang/p4c/files/9648178/Report1.pdf)

mbudiu-vmw:
@fruffy is the right person to answer this question. The validation pass sometimes has false positives.

fruffy:
I will take a look at this. This is a fairly tricky case.

usha1830:
> I will take a look at this. This is a fairly tricky case.

Please take a look at the comments mentioned here for the intended transformations 
https://github.com/p4lang/p4c/pull/3524/files#diff-0e52a75383e54b94bc4d93f70c6ccb8eba703f44751c595d23062ec97576c446R68

Please also look at the diff report attached [here](https://github.com/p4lang/p4c/pull/3524#issuecomment-1258342483)

fruffy:
The problem right now is that the semantic model for the individual programmable blocks does not quite match this pass' transformation. I need to adjust it and see whether the failures still happen. I disabled checking this pass for now in Gauntlet so this issue can be unblocked, but I can not guarantee that the pass does not have bugs.

mbudiu-vmw:
Beware that the simplifyDefUse pass has some bugs. I started working on it, but never managed to complete the work.
It works fine where it is now, but if you invoke it later it may produce incorrect results.

midend/flattenUnions.h
```diff
@@ -0,0 +1,163 @@
...
+ *                                    |     u_h1.setInvalid()
+ *                                    | }
+ *   b) u.h1 = {16W1}                 | This is already transformed into individual element copy
+ *                                    | by earlier passes and individual assignment is translated
+ *                                    | as same as (a)
+ *   c) u1.h1 = u2.h1                 | same as a) with my_h1 replaced with u2_h1
+ *   d) u1 = u2                       | This is already transformed into individual element copy
+ *                                    | by earlier passes and individual assignment is translated
+ *                                    | same as (c)
+ *   e) a = u.isValid()               | tmp = 0;
```
mbudiu-vmw, line 85: this should probably be a || of booleans instead. No cast will be needed.

midend/flattenUnions.cpp
```diff
@@ -0,0 +1,359 @@
...
+
+namespace P4 {
+const IR::MethodCallStatement*
+HandleValidityHeaderUnion::processValidityForStr(const IR::Statement *s, const IR::Member *m,
+                                                 cstring headerElement, cstring setValid) {
+    auto method = new IR::Member(s->srcInfo, new IR::Member(m->expr,IR::ID(headerElement)),
+                                 IR::ID(setValid));
+    auto mc = new IR::MethodCallExpression(s->srcInfo, method,
+                                           new IR::Vector<IR::Argument>());
+    typeMap->setType(method, new IR::Type_Method(IR::Type_Void::get(),
```
mbudiu-vmw, line 27: is this necessary? usually we only set types if they are needed right away, otherwise the next typeChecking pass can do it.

midend/flattenUnions.cpp
```diff
@@ -0,0 +1,359 @@
...
+                     new IR::ParameterList(), setValid));
+    return new IR::MethodCallStatement(mc->srcInfo, mc);
+}
+
+const IR::Node* HandleValidityHeaderUnion::setInvalidforRest(const IR::Statement *s,
+                                           const IR::Member *m, const IR::Type_HeaderUnion *hu,
+                                           cstring exclude, bool setValidforCurrMem) {
+    auto code_block = new IR::IndexedVector<IR::StatOrDecl>;
```
mbudiu-vmw, line 35: How does your backend handle this?
If the validity bits are all in a single word this can be done with a single assignment.
I wonder whether it wouldn't be more efficient to write a back-end specific replacement.

midend/flattenUnions.cpp
```diff
@@ -0,0 +1,359 @@
...
+
+const IR::Node * HandleValidityHeaderUnion::expandIsValid(
+                 const IR::Statement *a, const IR::MethodCallExpression *mce) {
+    auto mi = P4::MethodInstance::resolve(mce, refMap, typeMap);
+    if (auto bim = mi->to<P4::BuiltInMethod>()) {
+        if (bim->name == "isValid") {   // hdr.u.isValid() or u.isValid
+            if (auto huType = bim->appliedTo->type->to<IR::Type_HeaderUnion>()) {  // u or hdr.u
+                cstring tmp = refMap->newName("tmp");
+                IR::PathExpression *tmpVar = new IR::PathExpression(IR::ID(tmp));
+                toInsert.push_back(new IR::Declaration_Variable(
```
mbudiu-vmw, line 59: why not a boolean instead?

midend/flattenUnions.cpp
```diff
@@ -0,0 +1,359 @@
...
+                    typeMap->setType(method, new IR::Type_Method(IR::Type_Void::get(),
+                                     new IR::ParameterList(), IR::Type_Header::isValid));
+                    auto addOp = new IR::Add(a->srcInfo, tmpVar,
+                                             new IR::Constant(IR::Type_Bits::get(32), 1));
+                    auto assn = new IR::AssignmentStatement(a->srcInfo, tmpVar, addOp);
+                    code_block->push_back(new IR::IfStatement(a->srcInfo, mc, assn, nullptr));
+                }
+                auto cond = new IR::Equ(a->srcInfo, tmpVar,
+                                        new IR::Constant(IR::Type_Bits::get(32), 1));
+                if (auto  assn = a->to<IR::AssignmentStatement>())
```
mbudiu-vmw, line 79: can you return a value and let the postorder caller do this part?

midend/flattenUnions.cpp
```diff
@@ -0,0 +1,359 @@
...
+            if (auto m = a->appliedTo->to<IR::Member>()) {
+                if (auto huType = m->expr->type->to<IR::Type_HeaderUnion>())  // u.h1  or hdr.u.h1
+                    return setInvalidforRest(mcs, m, huType, m->member.name, false);
+            }
+        }
+    }
+    return mcs;
+}
+
+const IR::Node* HandleValidityHeaderUnion::postorder(IR::P4Action* action) {
```
mbudiu-vmw, line 171: This pattern shows up a lot in the compiler, we should think whether it can be somehow automated...
But not in this PR.

midend/flattenUnions.cpp
```diff
@@ -0,0 +1,359 @@
...
+    }
+    auto body = new IR::BlockStatement(action->body->srcInfo);
+    for (auto a : actiondecls)
+        body->push_back(a);
+    action->body = body;
+    return action;
+}
+
+const IR::Node* DoFlattenHeaderUnion::postorder(IR::Function* function) {
+    auto functiondecls = function->body->components;
```
mbudiu-vmw, line 309: aren't the functions inlined by now?

mbudiu-vmw (approved):
In general this looks good.

midend/flattenUnions.cpp
usha1830: Removed

midend/flattenUnions.cpp
usha1830: Done.

midend/flattenUnions.cpp
usha1830: Yes they are. I removed the visitor and also added a test for this.

usha1830 (commented):
Update as per review

mbudiu-vmw:
I will merge this PR since the tests pass. I assume that it's ready.

PR #3525, opened 2022-09-12
Add note to DPDK README that PSA egress is not implemented

jafingerhut:


jafingerhut:
@usha1830 For your consideration on whether to approve.  I can change it to be more accurate, if I have made a mistake.

usha1830 (approved):
This looks fine to me

backends/dpdk/README.md
```diff
@@ -35,6 +35,8 @@ To load the 'spec' file in dpdk follow the instructions in the
 - Parser Value Sets
 
 ### Unsupported PSA externs and features
+- egress parser, control, and deparser are not implemented, only
+  ingress parser, control, and deparser.  There is no traffic manager.
```
rst0git, line 39: If I understand correctly, "traffic manager" is described in the PSA specification as "Packet buffer and Replication Engine (PRE)". Is that correct?

jafingerhut: I have added another commit to this PR that mentions packet buffer and packet replication engine, to make the terms easier to find in the PSA spec.

rst0git (approved):
LGTM

PR #3528, opened 2022-09-15
Since dpdk now Support large struct fields(more than 64 bits) as key

kamleshbhalui:
Removing all checks


mbudiu-vmw:
I love doing this kind of reviews.

mbudiu-vmw (approved):

Issue #3531, opened 2022-09-21
Question about recently added static_assert() externs


mvido:
StaticAssert pass in frontend evaluates static_assert invocations. After this pass, the nodes representing the invocations of static_assert remain in the IR tree and possibly can cause problems for midends or backends. Does it make sense to eliminate these invocation nodes after they are evaluated in frontend or is there any use case why they could be needed after they are evaluated (and thus the midends/backends should deal with them)?

mbudiu-vmw:
We should in fact constant-fold it, since it should always evaluate to 'true'.

PR #3533, opened 2022-09-21
Constant-fold static_assert calls

mbudiu-vmw:


mbudiu-vmw:
Fixes #3531

frontends/p4/staticAssert.h
```diff
@@ -63,13 +65,28 @@ class DoStaticAssert : public Inspector {
...
                 } else {
-                    ::warning("Could not evaluate static_assert to a constant: %1%", arg);
+                    ::error(ErrorType::ERR_UNEXPECTED,
+                            "Could not evaluate static_assert to a constant: %1%", arg);
+                    return method;
                 }
             }
         }
-        return false;
+        if (getContext()->node->is<IR::MethodCallStatement>()) {
```
apinski-cavium, line 77: It might be useful to add a testcase for the above case.
Something like:
```
#include <core.p4>
void f()
{
  static_assert(static_assert(true, ""));
}

control c()
{
  apply {
  f();
  }
}
control ct();
package pt(ct c);
pt(c()) main;
```
Currently this produces the following warning I think:
```
static_assert5.p4:4: warning: Could not evaluate static_assert to a constant: static_assert
```
Note I have a `hack` in the code to change the warning/error message format to be more like the GNU message style (I might submit a patch later on to change to that style based on a command line option/environment variable and an option to disable printing out the carrot; this is useful for testing using dejagnu).

mbudiu-vmw: This test passes with the new changes. I have updated the test case in the PR.
The patch will be much appreciated.

fruffy (approved):

Issue #3453, opened 2022-07-20
eBPF PTF tests are failing


osinstom:
The eBPF PTF CI job is failing due to the new kernel (5.15) that has been released for Ubuntu 20.04 runner machine: https://github.com/actions/virtual-environments/commit/154099992ecb2fe92d55f51dc12827f47ee160da. 

@tatry and I are already looking into it. 

CC: @mbudiu-vmw

fruffy:
Can this issue be closed or is there another fix coming?

tatry:
Yes, it can be closed. Now tests should works for kernel 5.15.

PR #3535, opened 2022-09-23
Run PTF tests under different kernel versions

tatry:
This PR allows to run PTF-ebpf tests on different kernel versions using public GitHub runners.

Why this PR is needed:
- Ensure some sort of backward compatibility for ebpf backend.
- Test new features with older kernels (see below).

Because GitHub runners do not support KVM acceleration and tests last long (2-3h, see [example workflow run](https://github.com/tatry/p4c/actions/runs/3096196743) in my fork), with @osinstom we made following decisions:
- Do not run on pull requests or pushes, but in weekly manner using `cron`. When tests fails this would require more investigation to debug, but current work flow is not changed nor extended in time.
- Each kernel is tested in a separate job, which should ensure that timeout 6h is not met. Testing more kernels in one may not be successful due to that timeout.
- Cache is used to speed up jobs, but images might not always be available (i.e. approximately there is place for 6 images/kernels and is preserved for 7 days). The cache used is different from used in P4C already (or at least it looks like they are different).
- Preserve existing PTF tests to test ebpf backend.

This PR doesn't provide script to run these tests locally, because they heavily use virtual machines. If anybody wants run test locally on a specified kernel it is recommended to just install that kernel on his/her machine and boot into it to run tests.

Example use would be during resolving issue https://github.com/p4lang/p4c/issues/3453, where changes in the latest kernel cause problems for `psabpf-ctl` (we will rename it into `nikss-ctl`) with adding a new inner maps to an outer map.

### DIFF
--- a/.github/workflows/ci-ptf-kernels-weekly.yml
+++ b/.github/workflows/ci-ptf-kernels-weekly.yml
@@ -1,0 +1,90 @@
+# Copyright 2022-present Orange
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+name: "ptf-ebpf-kernels-weekly"
+
+on:
+  schedule:
+    # Every sunday at 1:17 am UTC
+    - cron: '17 1 * * 0'
+
+jobs:
+  ptf-linux:
+    # Prevent triggering by a fork because this workflow is resource intensive
+    if: ${{ github.repository == 'p4lang/p4c' && github.ref == 'refs/heads/main' }}
+    strategy:
+      fail-fast: false
+      matrix:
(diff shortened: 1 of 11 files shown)


mbudiu-vmw:
From my point of view this looks fine.
I don't know who pays for our CI and how much quota we have.
Running once a week sounds reasonable.
Anyone else wants to comment on this?

fruffy:
I do not have a problem with this, considering it runs once a week. 
To simplify CI a little I would also suggest to do the same with the ptf-ebpf action. Or, considering the repository is not part of the p4c back ends,  move testing for it to a downstream repository (https://github.com/P4-Research/psabpf).

osinstom:
> I do not have a problem with this, considering it runs once a week. To simplify CI a little I would also suggest to do the same with the ptf-ebpf action. Or, considering the repository is not part of the p4c back ends, move testing for it to a downstream repository (https://github.com/P4-Research/psabpf).

@fruffy I'm not getting your point:
1) ptf-ebpf action's purpose is to be run as pre-merge job to validate if a new PR doesn't break PSA/eBPF backend functionality on the latest kernel, while the goal of weekly job running on different kernels is to validate backward compatibility/operability on different (older) kernels. The purpose of these 2 jobs is different.
2) Which repository is not part of the p4c back ends? If you mean psabpf, the CI jobs are not testing psabpf - they test PSA/eBPF compiler but use psabpf to populate P4 table entries for PTF tests

osinstom (approved):

fruffy:
I see, I assumed psabpf to be a compiler extension but it is completely independent. I introduced changes in the past that were breaking it, but they seem related to dependencies instead. Nevermind then!

Issue #3544, opened 2022-10-01
"Parser cycle without extracting any bytes" warning should not be printed for parser cycle that advances the packet pointer


kfcripps:
I noticed that when I build test testdata/p4_16_samples/parser-unroll-issue3537-1.p4, the following warning is emitted:
```
[--Wwarn=invalid] warning: Parser cycle without extracting any bytes:
Parser ParserI state chain: start, s1, s2, s2
```
Why is this warning printed here? Why do I care that I have a parser cycle that doesn't extract any bytes? In the case of this example, I only want to advance the packet cursor without extracting any headers, so I'm not sure what I can do to get rid of this warning.

_Originally posted by @kfcripps in https://github.com/p4lang/p4c/issues/3538#issuecomment-1264031764_

kfcripps:
My main concern is not that the warning is printed, but rather that
> the "advance" call is not accounted in the code that checks whether the packet pointer has been advanced

which could result in other (currently unknown) unintended behavior in the `ParsersUnroll` pass

mbudiu-vmw:
Fixed by #3548

PR #3548, opened 2022-10-03
Change warrning messages in ParserUnroll

VolodymyrPeschanenkoLitSoft:
These changes were done according to #3544. ParserUnroll hasn't any deep analysis for the real reason of non-unrolled loops. I think that such analysis will need the same efforts for implementation as ParserUnroll. So, I think that the warning messages should be changed:  ParserUnroll can't detect the number of loop iterations without an explanation of the real reason.

### DIFF
--- a/midend/parserUnroll.cpp
+++ b/midend/parserUnroll.cpp
@@ -638,7 +638,7 @@ class ParserSymbolicInterpreter {
                     if (equStackVariableMap(crt->statesIndexes, state->statesIndexes)) {
                         ::warning(ErrorType::ERR_INVALID,
                                   "Parser cycle can't be unrolled, because ParserUnroll can't \
-                                   detect the number of loops iterations:\n%1%",
+                                   detect the number of loop iterations:\n%1%",
                                   stateChain(state));
                         wasError = true;
                     }
--- a/midend/parserUnroll.cpp
+++ b/midend/parserUnroll.cpp
@@ -635,16 +635,11 @@ class ParserSymbolicInterpreter {
                             }
                         }
                     }
-
-                    if (conservative)
-                        ::warning(ErrorType::WARN_PARSER_TRANSITION,
-                                    "Potential parser cycle without extracting any bytes:\n%1%",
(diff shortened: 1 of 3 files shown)


kfcripps:
I wonder if the warning would go away if we also handled `packet_in::advance` calls in `ExpressionEvaluator::postorder(const IR::MethodCallExpression* expression)` (similar to how `packet_in::extract` calls are handled there)..

midend/parserUnroll.cpp
```diff
@@ -625,26 +625,19 @@ class ParserSymbolicInterpreter {
...
                     }
-
-                    if (conservative)
-                        ::warning(ErrorType::WARN_PARSER_TRANSITION,
-                                    "Potential parser cycle without extracting any bytes:\n%1%",
-                                    stateChain(state));
-                    else
+                    if (equStackVariableMap(crt->statesIndexes, state->statesIndexes)) {
                         ::warning(ErrorType::ERR_INVALID,
-                                  "Parser cycle without extracting any bytes:\n%1%",
```
mbudiu-vmw, line 645: this is ok, but is advance considered?

mbudiu-vmw (approved):

midend/parserUnroll.cpp
VolodymyrPeschanenkoLitSoft: The warning message will be printed for an "advance" without an explanation of the real reason:
```
[--Wwarn=invalid] warning: Parser cycle can't be unrolled, because ParserUnroll can't detect the number of loop iterations:
Parser ParserI state chain: start, s1, s2, s2
```
I don't think that catching an "advance" here is a good solution because it is considered in Interpreter class only. ParserUnroll uses this class for the calculation of the values of header stack indexes only.  The real reason why this loop can't be unrolled with ParserUnroll is that the states of the header stack indexes are the same for the last iteration of the loop. In general, it could be any reason for that, not just "advance". I can't build a complete list of the reasons for that. So, I decide to make the warning message more general.

PR #3549, opened 2022-10-03
Add a policy to control predication pass errors

mbudiu-vmw:


mbudiu-vmw:
This change adds a policy to the predication pass which controls whether the inability to remove statements is fatal or not.
It is conceivable that predication can be applied on targets that still support conditionals in actions, for example.
With this change the p4test compiler no longer will report errors when predication cannot eliminate all if statements in actions.

testdata/p4_16_samples_outputs/issue513-midend.p4
```diff
@@ -0,0 +1,68 @@
...
+    state start {
+        pkt.extract<Ethernet_h>(hdr.ethernet);
+        transition accept;
+    }
+}
+
+control cIngress(inout Parsed_packet hdr, inout mystruct1 meta, inout standard_metadata_t stdmeta) {
+    @name("cIngress.foo") action foo() {
+        meta.b = meta.b + 4w5;
+        mark_to_drop();
```
rst0git, line 36: @mbudiu-vmw Should we use `mark_to_drop(stdmeta)` here to avoid the warning about deprecated feature?

jafingerhut, line 36: There is no 'if' statement in this action, which appears to be incorrect, as it does not match the behavior of the input program.

mbudiu-vmw: maybe it's not that easy, I guess the pass was not designed for this case.
I will mark this as draft for now.

mbudiu-vmw: I just moved this test from err, without any other changes.

jafingerhut: If you expect this might not happen quickly, Mihai, would you (or anyone reading this) be able to provide the p4test command line option that I asked about in this p4-arch email? https://groups.google.com/a/lists.p4.org/g/p4-arch/c/FD-IxwekCUI/m/9hSFxPSWAwAJ

rst0git: > can you recommend a full p4test command line that would skip the fewest passes possible, and would not give warnings for 'if' conditions inside of action bodies?

Hi Andy, as far as I understand, the warning for `if` conditions inside of action bodies happens at the _Predication_ midend pass.
https://github.com/p4lang/p4c/blob/ceeb043ac19208ea34ec94fd1a1ddae7ee9e7ed1/midend/predication.h#L117-L119

Thus, `p4test --excludeMidendPasses Predication <program>.p4` command should not show this warning. For example:
```sh
p4test --excludeMidendPasses Predication testdata/p4_16_samples/pna-example-tcp-connection-tracking.p4
```

jfingerh: @rst0git That worked for me.  Thanks, Radostin!

mbudiu-vmw:
This PR is wrong, so I will close it.

PR #3568, opened 2022-10-12
Converting constraints from a string to IR expressions

OleksandrLetychevskyiIntel:
Converts string constraints to IR expressions and generates restrictions for the program

### DIFF
--- a/backends/p4tools/testgen/targets/bmv2/p4_asserts_parser.cpp
+++ b/backends/p4tools/testgen/targets/bmv2/p4_asserts_parser.cpp
@@ -1,0 +1,624 @@
+#include "backends/p4tools/testgen/targets/bmv2/p4_asserts_parser.h"
+
+#include <boost/multiprecision/cpp_int.hpp>
+#include <boost/optional/optional_io.hpp>
+
+#include "lib/error.h"
+
+#include "backends/p4tools/common/core/z3_solver.h"
+#include "backends/p4tools/common/lib/ir.h"
+
+namespace P4Tools {
+
+namespace AssertsParser {
+
+AssertsParser::AssertsParser(std::vector<std::vector<const IR::Expression*>>& output)
+    : restrictionsVec(output) {
+    setName("Restrictions");
+}
+
+bool is_space(char c) noexcept {
+    switch (c) {
+        case ' ':
+        case '\t':
+        case '\r':
+            return true;
+        default:
+            return false;
+    }
+}
+
+std::ostream& operator<<(std::ostream& os, const Token::Kind& kind) {
+    static const char* const names[]{
+        "Text",         "True",       "False",          "LineStatementClose",
+        "Id",           "Number",     "Minus",          "Plus",
+        "Dot",          "FieldAcces", "MetadataAccess", "LeftParen",
+        "RightParen",   "Equal",      "NotEqual",       "GreaterThan",
+        "GreaterEqual", "LessThan",   "LessEqual",      "LNot",
+        "Colon",        "Semicolon",  "Conjunction",    "Disjunction",
+        "Implication",  "Slash",      "Percent",        "Shr",
+        "Shl",          "Mul",        "Comment",        "Unknown",
+        "EndString",    "End",
+    };
+    return os << names[static_cast<int>(kind)];
+}
+
+const IR::Expression* makeExpr(std::vector<const IR::Expression*> input) {
+    const IR::Expression* expr = nullptr;
+    for (int i = 0; i < input.size(); i++) {
+        if (auto lOr = input[i]->to<IR::LOr>()) {
+            if (expr == nullptr) {
+                if (lOr->right->toString() == "(tmp") {
+                    std::vector<const IR::Expression*> tmp;
+                    int idx = i;
+                    i++;
+                    if (i + 1 == input.size()) {
+                        expr = new IR::LOr(input[idx - 1], input[i]);
+                        break;
+                    }
+                    while (i < input.size()) {
+                        tmp.push_back(input[i]);
+                        i++;
+                    }
+                    expr = new IR::LOr(input[idx - 1], makeExpr(tmp));
+                    break;
+                }
+                expr = new IR::LOr(input[i - 1], input[i + 1]);
+                continue;
+            } else if (lOr->right->toString() == "(tmp") {
+                if (i + 1 != input.size()) {
+                    std::vector<const IR::Expression*> tmp;
+                    i++;
+                    if (i + 1 == input.size()) {
+                        expr = new IR::LOr(input[i - 2], input[i]);
+                    }
+                    while (i < input.size()) {
+                        tmp.push_back(input[i]);
+                        i++;
+                    }
+                    expr = new IR::LOr(expr, makeExpr(tmp));
+                }
+
+            } else {
+                if (i + 1 != input.size()) {
+                    expr = new IR::LOr(expr, input[i + 1]);
+                }
+            }
+        }
+        if (input[i]->is<IR::LAnd>()) {
+            if (expr == nullptr) {
+                expr = new IR::LAnd(input[i - 1], input[i + 1]);
+                continue;
+            } else {
+                if (i + 1 != input.size()) expr = new IR::LAnd(expr, input[i + 1]);
+            }
+        }
+    }
+
+    return expr;
+}
+const IR::Expression* getIR(std::vector<Token> tokens, IR::Vector<IR::KeyElement> keyElements) {
+    const IR::Expression* expr = nullptr;
+    std::vector<const IR::Expression*> exprVec;
+    const IR::Type_Base* type = nullptr;
+
+    for (int i = 0; i < tokens.size(); i++) {
+        if (tokens[i].kind() == Token::Kind::Minus || tokens[i].kind() == Token::Kind::Plus ||
+            tokens[i].kind() == Token::Kind::Equal || tokens[i].kind() == Token::Kind::NotEqual ||
+            tokens[i].kind() == Token::Kind::GreaterThan ||
+            tokens[i].kind() == Token::Kind::GreaterEqual ||
+            tokens[i].kind() == Token::Kind::LessThan ||
+            tokens[i].kind() == Token::Kind::LessEqual || tokens[i].kind() == Token::Kind::Slash ||
+            tokens[i].kind() == Token::Kind::Percent || tokens[i].kind() == Token::Kind::Shr ||
+            tokens[i].kind() == Token::Kind::Shl || tokens[i].kind() == Token::Kind::Mul ||
+            tokens[i].kind() == Token::Kind::NotEqual) {
+            const IR::Expression* leftL = nullptr;
+            const IR::Expression* rightL = nullptr;
+            if (i != 0) {
+                i--;
+                i--;
+            }
+            if (i <= 0 || tokens[i].kind() == Token::Kind::Conjunction ||
+                tokens[i].kind() == Token::Kind::Disjunction ||
+                tokens[i].kind() == Token::Kind::LeftParen ||
+                tokens[i].kind() == Token::Kind::Implication) {
+                i++;
+
+                if (tokens[i].kind() == Token::Kind::Text) {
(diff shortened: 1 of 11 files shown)


backends/p4tools/testgen/targets/bmv2/p4_refers_to_parser.h
```diff
@@ -0,0 +1,35 @@
...
+
+#include "ir/ir.h"
+
+namespace P4Tools {
+namespace RefersToParser {
+
+class RefersToParser : public Transform {
+    std::vector<std::vector<const IR::Expression*>>& restrictionsVec;
+    std::vector<const IR::P4Action*> actionVector;
+    const IR::P4Action* findAction(const IR::ActionListElement* input);
```
fruffy, line 19: These functions are all missing /// comments.

backends/p4tools/testgen/targets/bmv2/program_info.cpp
```diff
@@ -52,9 +54,17 @@ BMv2_V1ModelProgramInfo::BMv2_V1ModelProgramInfo(
     }
     // Sending a too short packet in BMV2 produces nonsense, so we require the packet size to be
     // larger than 32 bits
-    targetConstraints =
+    const IR::Operation_Binary* constraint =
         new IR::Grt(IR::Type::Boolean::get(), ExecutionState::getInputPacketSizeVar(),
                     IRUtils::getConstant(ExecutionState::getPacketSizeVarType(), 32));
+    program->apply(AssertsParser::AssertsParser(restrictionsVec));
```
fruffy, line 60: Please explain the purpose of these passes with a comment.

backends/p4tools/testgen/targets/bmv2/program_info.h
```diff
@@ -31,6 +31,9 @@ class BMv2_V1ModelProgramInfo : public ProgramInfo {
     /// The bit width of standard_metadata.parser_error.
     static const IR::Type_Bits parserErrBits;
 
+    /// Vector containing pairs of restrictions and nodes to which these restrictions apply.
+    std::vector<std::vector<const IR::Expression*>> restrictionsVec;
```
fruffy, line 35: Is this actually still needed? Is this access anywhere else?

backends/p4tools/testgen/targets/bmv2/p4_asserts_parser.cpp
```diff
@@ -0,0 +1,623 @@
...
+        "Colon",        "Semicolon",  "Conjunction",    "Disjunction",
+        "Implication",  "Slash",      "Percent",        "Shr",
+        "Shl",          "Mul",        "Comment",        "Unknown",
+        "EndString",    "End",
+    };
+    return os << names[static_cast<int>(kind)];
+}
+
+const IR::Expression* makeExpr(std::vector<const IR::Expression*> input) {
+    const IR::Expression* expr = nullptr;
```
fruffy, line 46: Can you simplify this function and add comments? It is hard to understand and comprehend. Rule of thumb: Segment this code into independent logical parts and move these parts into a helper function.

backends/p4tools/testgen/targets/bmv2/p4_asserts_parser.cpp
```diff
@@ -0,0 +1,623 @@
...
+        return expr;
+    }
+
+    expr = makeExpr(tmp);
+
+    return expr;
+}
+
+std::vector<const IR::Expression*> AssertsParser::genIRStructs(cstring str) {
+    Lexer lex(str);
```
fruffy, line 338: Similar comments apply to all these other functions. They are huge and very hard to read.

backends/p4tools/testgen/targets/bmv2/p4_asserts_parser.h
```diff
@@ -0,0 +1,113 @@
...
+
+    Token(Kind kind, const char* beg, std::size_t len) noexcept
+        : m_kind{kind}, m_lexeme(beg, len) {}
+
+    Token(Kind kind, const char* beg, const char* end) noexcept
+        : m_kind{kind}, m_lexeme(beg, std::distance(beg, end)) {}
+
+    Kind kind() const noexcept { return m_kind; }
+
+    void kind(Kind kind) noexcept { m_kind = kind; }
```
fruffy, line 78: Move all of these bodies to the  cpp file. Also add comments please.

backends/p4tools/testgen/targets/bmv2/p4_refers_to_parser.cpp
```diff
@@ -0,0 +1,152 @@
...
+    for (long unsigned int i = 2; i < input.size(); i++) {
+        result += input[i]->text;
+    }
+    return result;
+}
+
+void RefersToParser::createRefersToConstraint(IR::Vector<IR::Annotation> annotations,
+                                              const IR::Type* inputType, cstring controlPlaneName,
+                                              int id, bool isParameter, cstring inputName) {
+    cstring destTableName = "";
```
fruffy, line 90: This function can also be simplified.

backends/p4tools/testgen/targets/bmv2/program_info.h
```diff
@@ -1,5 +1,5 @@
-#ifndef BACKENDS_P4TOOLS_TESTGEN_TARGETS_BMV2_PROGRAM_INFO_H_
-#define BACKENDS_P4TOOLS_TESTGEN_TARGETS_BMV2_PROGRAM_INFO_H_
+#ifndef TESTGEN_TARGETS_BMV2_PROGRAM_INFO_H_
```
fruffy, line 1: This an incorrect header guard definition. I wonder why cpplint does not catch this.

fruffy (changes requested):

backends/p4tools/testgen/targets/bmv2/p4_refers_to_parser.cpp
OleksandrLetychevskyiIntel: It is already quite simple, of course, complicated by the fact that I tried to unify it for two cases, so as not to duplicate the code. After all, `refers_to ` happens both in the key and in the parameter, because of this, we had to make a function that can support both options. What exactly are you trying to simplify here?

fruffy: For example, you could reduce the nesting by using `continue`. The function does not need to iterate further if the annotation name is not `refers_to`. You can also reset destTableName and destKeyName right at the beginning.

fruffy:
I will run this code tomorrow to see whether it is complete.

fruffy:
I tried to clean up some of the code but there is lots of work left. This code has a lot of problems. I attached a patch, please apply it. 

[p4_asserts_parser.patch.txt](https://github.com/p4lang/p4c/files/9855413/p4_asserts_parser.patch.txt)

backends/p4tools/testgen/targets/bmv2/p4_asserts_parser.cpp
```diff
@@ -0,0 +1,673 @@
...
+            if (i <= 0 || tokens[i].is_one_of(Token::Kind::Conjunction, Token::Kind::Disjunction,
+                                              Token::Kind::LeftParen, Token::Kind::Implication)) {
+                i++;
+                leftL = makeConstant(tokens[i], keyElements);
+            } else {
+                leftL = exprVec[exprVec.size() - 1];
+                exprVec.pop_back();
+                i++;
+            }
+            i++;
```
fruffy, line 195: Please do not use index manipulation like this. It is confusing and error-prone. There is no need to have a single i++ or i-- in this entire file. I tried to refactor this but I do not quite understand the intention behind some of these statements. Furthermore, `i` must never be negative. That is considered a bug.
Also you are adding and removing expressions in this for loop from the vector you are iterating on at the same time. Please break up this code such that these two operations are cleanly separated.

fruffy (changes requested):

backends/p4tools/testgen/targets/bmv2/p4_asserts_parser.cpp
OleksandrLetychevskyiIntel: That is, to do something without changing the index?

backends/p4tools/testgen/targets/bmv2/test/BMV2PTFXfail.cmake
```diff
@@ -15,3 +15,9 @@ p4tools_add_xfail_reason(
   # Assert/Assume error: assert/assume(false).
   bmv2_assert.p4
 )
+
+p4tools_add_xfail_reason(
+  "testgen-p4c-bmv2-ptf"
+  "Key does not have a name annotation."
+  bmv2_restrictions.p4
```
fruffy, line 22: I would fix up this program instead of adding it to xfails. Why does the key here not have an annotation?

OleksandrLetychevskyiIntel: I don't quite understand, is the name annotation required for table keys?

fruffy: If it is not present it generally is always added, I am curious why that does not happen here.

backends/p4tools/testgen/targets/bmv2/p4_asserts_parser.cpp
```diff
@@ -0,0 +1,663 @@
...
+    const IR::Expression* expr = nullptr;
+    for (uint64_t i = 0; i < input.size(); i++) {
+        if (const auto* lOr = input[i]->to<IR::LOr>()) {
+            if (lOr->right->toString() == "(tmp") {
+                if (i + 1 == input.size()) {
+                    break;
+                }
+                std::vector<const IR::Expression*> tmp;
+                uint64_t idx = i;
+                i++;
```
fruffy, line 89: There is still a bunch of i++ in this code. They should not be necessary.

backends/p4tools/testgen/targets/bmv2/p4_asserts_parser.cpp
```diff
@@ -0,0 +1,663 @@
...
+            cstring annotationName;
+            if (const auto* annotation = key->getAnnotation("name")) {
+                if (!annotation->body.empty()) {
+                    annotationName = annotation->body[0]->text;
+                } else if (!annotation->expr.empty()) {
+                    annotationName = annotation->expr[0]->toString();
+                }
+            }
+            BUG_CHECK(annotationName.size() > 0, "Key does not have a name annotation.");
+            cstring separator = "\"";
```
fruffy, line 148: Can you please add a comment explaining why this is needed or what it is for?

fruffy:
I went through the code again, there are still items left to do. I made some comments. 

Here is a new patch that should be applied. 

[p4_asserts_parser.patch.txt](https://github.com/p4lang/p4c/files/9881386/p4_asserts_parser.patch.txt)

backends/p4tools/testgen/targets/bmv2/p4_asserts_parser.cpp
OleksandrLetychevskyiIntel: I don't know for what reason, but when I take `toString` from expr in annotation , it returns the string `"keyname"` to me and not `keyname`. Due to the fact that `"` remains in the string, the algorithm cannot compare the names

fruffy:
Another patch. 
[p4_asserts_parser.patch.txt](https://github.com/p4lang/p4c/files/9890707/p4_asserts_parser.patch.txt)

backends/p4tools/testgen/targets/bmv2/p4_asserts_parser.cpp
fruffy: I do not think this has been fixed yet.

OleksandrLetychevskyiIntel: Look, I changed the algorithm now removed this part

fruffy:
Another patch set to apply. 
[p4_asserts_parser.patch.txt](https://github.com/p4lang/p4c/files/9933153/p4_asserts_parser.patch.txt)

fruffy:
Can you update this to the latest changes in master? We should run the gtests to check before merging.

PR #3569, opened 2022-10-12
Reachability feature

VolodymyrPeschanenkoLitSoft:
These changes contain a path for building a control flow-directed cycle graph with corresponding unit tests. This path could be used for checking the existence of a path between two IR::Node vertexes inside a program.

### DIFF
--- a/backends/p4tools/common/testgen.def
+++ b/backends/p4tools/common/testgen.def
@@ -12,7 +12,7 @@ class Extracted_Varbits : Type_Bits {
     int assignedSize;
 
     /// Returns the current width this variable bit type has.
-    int width_bits() const { return assignedSize; }
+    int width_bits() const override { return assignedSize; }
 
     equiv {
         // Ignore the current size and only check for equivalence of the assigned size.
@@ -25,7 +25,6 @@ class Extracted_Varbits : Type_Bits {
     }
 
     cstring baseName() const { return "SizedVarbit"; }
-
     toString {
         return baseName() + "<" + Util::toString(size) + ">" + "<" + Util::toString(assignedSize) +
                ">";
--- a/backends/p4tools/testgen/targets/bmv2/test/p4-programs/bmv2_if.p4
+++ b/backends/p4tools/testgen/targets/bmv2/test/p4-programs/bmv2_if.p4
@@ -1,0 +1,69 @@
+#include <v1model.p4>
+
+header ethernet_t {
+    bit<48> dst_addr;
+    bit<48> src_addr;
+    bit<16> eth_type;
+}
+
+header H {
+    bit<8> a;
+    bit<8> b;
+}
+
+struct Headers {
+    ethernet_t eth_hdr;
+    H          h;
+}
+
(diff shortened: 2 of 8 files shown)


fruffy-bfn:
@mbudiu-vmw This pass is actually a statement reachability analysis, which allows you to ask questions such as "Is table xyz reachable from statement a"? The tests show how this is intended to work. 

I was wondering, do we have something similar in the compiler already or are there utilities we can reuse for this pass? For now, we can just use this for P4Tools and then eventually contribute it to the mid end once it has matured.

fruffy:
Can you move this pass to the P4Tools directory for now? We can contribute this pass to top-level p4c later. Also I believe this is missing the programInfo part.

mbudiu-vmw:
There are several passes that compute this kind of information. There is a CFG data structure which is custom to BMV2. The ControlFlowVisitor keeps track of reachability too. The def-use pass as well. And the loop unrolling pass does that for parsers too. The graphs backend is yet another one.

fruffy:
> There are several passes that compute this kind of information. There is a CFG data structure which is custom to BMV2. The ControlFlowVisitor keeps track of reachability too. The def-use pass as well. And the loop unrolling pass does that for parsers too. The graphs backend is yet another one.

Is any of these general enough to apply to all statements? This is what we have been using this particular pass for. For now, we will just use it in our back end, but I wonder if it may make sense to merge some of this.

fruffy (approved):

VolodymyrPeschanenkoLitSoft:
> What reachability definition are you computing? The ControlFlowVisitor is very general.

It defines the reachability of some IR::Node from another IR::Node by control flow ad it can translate the names of the annotations into corresponded IR::Nodes which were synchronized with the p4tools control flow traversal procedure. 

Regarding the usage of the ControlFlowVisitor, I need to experiment with it. I let you know soon.

VolodymyrPeschanenkoLitSoft:
> What reachability definition are you computing? The ControlFlowVisitor is very general.
Regarding the usage of the ControlFlowVisitor: 
We decide to use Inspector instead of ControlFlowVisitor, because we didn't want to make copy the object before the call the apply. The apply function is crushed if it has the same object in an argument. In this case, we need to call the visit function from the Visitor and take care of the joins by ourselves. It means that algorithms for joins detection will not be called from ControlFlowVisitor and in this case, there are no reasons to use it.

VolodymyrPeschanenkoLitSoft:
> > What reachability definition are you computing? The ControlFlowVisitor is very general.

Regarding the usage of the ControlFlowVisitor:
We decide to use Inspector instead of ControlFlowVisitor, because we didn't want to make copy the object before the call the apply. The apply function is crushed if it has the same object in an argument. In this case, we need to call the visit function from the Visitor and take care of the joins by ourselves. It means that algorithms for joins detection will not be called from ControlFlowVisitor and in this case, there are no reasons to use it.

fruffy:
I will merge this to unblock PRs that depend on it. We should move the discussion to an issue.

PR #3577, opened 2022-10-14
Fix and improve ordered_map & ordered_set

vlstill:
* Fix erase in ordered_{set,map}

  - The original implementation used non-const iterator which makes the interface
  different from the standard set/map.
  - ordered_map::{insert,emplace}(iterator hint, ...) overloads have the same
  problem, but I kept them. These functions don't really make sense for
  insertion-order-preserving containers so it would be best to remove them. But
  that is not possible as they are used. The {insert,emplace} for a
  iterator-range with hint is very broken as it inserts values in the reverse
  order to whay they are in the input range.

* Do not expose modifying iterator from ordered_set

  - It is very unsafe -- if the value was changed using the iterator there is a
    risk of introducing duplicates in the set.
  - Therefore both iterator and const_iterator are now the same -- this matches
    what std::set does in libstdc++.

* Allow construction of ordered_{map,set} from iterator range
* Fix initalizer_list constructor in ordered_map 
  Initializer list is a reference-like object, pass it by value (as std:: does).

### DIFF
--- a/lib/ordered_map.h
+++ b/lib/ordered_map.h
@@ -84,6 +84,8 @@ class ordered_map {
  public:
     ordered_map() {}
     ordered_map(const ordered_map &a) : data(a.data) { init_data_map(); }
+    template<typename InputIt>
+    ordered_map(InputIt first, InputIt last) : data(first, last) { init_data_map(); }
     ordered_map(ordered_map &&a) = default; /* move is ok? */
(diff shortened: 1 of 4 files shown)


mbudiu-vmw:
@ChrisDodd wrote this code, so I requested a review from him

mbudiu-vmw (approved):

PR #3579, opened 2022-10-17
Remove header field size constraints, now

kamleshbhalui:
dpdk target accepts larger fields size

### DIFF
--- a/backends/dpdk/dpdkArch.h
+++ b/backends/dpdk/dpdkArch.h
@@ -233,6 +233,9 @@ class InjectOutputPortMetadataField : public Transform {
     const IR::Node *preorder(IR::Type_Struct *s) override;
 };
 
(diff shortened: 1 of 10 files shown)


backends/dpdk/dpdkArch.h
```diff
@@ -232,7 +232,9 @@ class InjectOutputPortMetadataField : public Transform {
         structure(structure) {}
     const IR::Node *preorder(IR::Type_Struct *s) override;
 };
-
+/// This pass replace unaligned header fields with aligned header fields
```
mbudiu-vmw, line 235: replace -> replaces (twice)

mbudiu-vmw (approved):

backends/dpdk/dpdkArch.h
kamleshbhalui: fixed

PR #3582, opened 2022-10-18
made "errorCount", "warningCount", and "maxErrorCount" [all "unsigned int"] be "protected" rather than "private" in "class ErrorReporter" in "error_reporter.h"

Abe149:
This small change-set will enable the developers of in-house P4 compilers [e.g. Intel] to elegantly [i.e. without nasty kludges] make local changes to the errors/warnings/both behavior in subclasses of "`class ErrorReporter`".

### DIFF
--- a/lib/error_reporter.h
+++ b/lib/error_reporter.h
@@ -36,6 +36,10 @@ enum class DiagnosticAction {
 // Some compatibility for printf-style arguments is also supported.
 class ErrorReporter {
  protected:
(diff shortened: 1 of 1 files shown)


davidbolvansky (approved):
Makes sense. Thanks.

Abe149:
> Makes sense. Thanks.

You are welcome.  :-)

amreshk (approved):

Issue #3584, opened 2022-10-19
Exclusion ranges overlap error when Clang is used as C++ compiler


davidbolvansky:
On Linux Ubuntu 20 LTS with clang 12 and newer and GC 7.6.4 or  8.2.2 (two version I tried) and simple "cmake ../" I always results in error "Exclusion ranges overlap" which is coming from Boehm GC.

asl:
@davidbolvansky on aarch64 linux (Ubuntu 20.04) with default clang 10 I'm getting hang of irgenerator when GC is enabled

asl:
(feel free to assign to me ;) )

asl:
The hang reason is similar to #3415. A bit more sophisticated workaround is necessary. I cannot reproduce "Exclusion ranges overlap" on aarch64/linux. Let's see if the new fix will fix your problem as well.

asl:
Now it seems we can reproduce the issue on CI: https://github.com/p4lang/p4c/actions/runs/3284630342/jobs/5410814947

PR #3585, opened 2022-10-19
Restructure malloc / realloc implementations to inhibit libcall optimizations done by compilers even more

asl:


asl:
Tagging @davidbolvansky

asl:
This may fix https://github.com/p4lang/p4c/issues/3584

davidbolvansky:
Possibly easier approach

`void* malloc(size_t x) __attribute__((no_builtin("malloc"))) {
    return realloc(nullptr, x);
}`

https://godbolt.org/z/cseMGqb6G

davidbolvansky:
Alternative appoach (works even with clang < 10 which has no no_builtin attribute) https://godbolt.org/z/T86vs8dEW

asl:
> void* malloc(size_t x) __attribute__((no_builtin("malloc"))) { return realloc(nullptr, x); }

Unfortunately, it's clang-only attribute (https://gcc.gnu.org/bugzilla/show_bug.cgi?id=96628). So, we'd need to provide some gcc shim. And nothing would prevent from, say, future versions of gcc to optimize in the similar way.

davidbolvansky:
Okay, I that case, please add -fno-builtin to compilation flags of this file or for all p4c.

https://godbolt.org/z/Gxfar6cvq

Disables libcall optimizations.

davidbolvansky:
could you try to add 
set_target_properties(p4ctoolkit PROPERTIES COMPILE_FLAGS -fno-builtin)

as last line of https://github.com/p4lang/p4c/blob/main/lib/CMakeLists.txt ?

davidbolvansky:
Alternative approach: https://github.com/p4lang/p4c/pull/3586

asl:
@davidbolvansky Does it fix the issue with clang on linux for you?

davidbolvansky:
> @davidbolvansky Does it fix the issue with clang on linux for you?

https://github.com/p4lang/p4c/pull/3586 fixes https://github.com/p4lang/p4c/issues/3584. I think it is a simple fix and more future-proof. Does my path fix issue for you? (Compiler hangs)

asl:
It should work IMO. I can imagine it could be broken with LTO and / or unity builds though...

davidbolvansky:
> It should work IMO. I can imagine it could be broken with LTO and / or unity builds though...

I think it should work fine but I will leave a note for future readers in that PR.

PR #3588, opened 2022-10-21
Dpdk Backend: Minor changes to add error check for mask operands and adding optional match kind to pna.p4

usha1830:


p4include/pna.p4
```diff
@@ -263,8 +263,9 @@ enum PNA_IdleTimeout_t {
 
 // BEGIN:Match_kinds
 match_kind {
-    range,   /// Used to represent min..max intervals
-    selector /// Used for dynamic action selection via the ActionSelector extern
+    range,    /// Used to represent min..max intervals
+    selector, /// Used for dynamic action selection via the ActionSelector extern
+    optional  /// Either an exact match, or a wildcard matching any value for the entire field
```
apinski-cavium, line 268: Right now the PNA spec does not even specify the match kinds. It just says:
```
TBD: Consider simply referencing the corresponding section of the PSA specification for this, unless we want to have something different in PNA.
```

It seems like it is best if we decide on the spec of match kinds here before adding to things to DPDK pna implementation.

jafingerhut: There is one existing PR, and another that I believe Nate Foster is working on, that would add `optional` match_kind to PNA.  It has been in v1model and PSA for over a year now, I believe, and I thought it was already in PNA, but must have neglected to add it when it was added to PSA.  I can add this as a discussion topic for the 2022-Oct-24 public PNA meeting.

usha1830:
Since optional mathckind PR https://github.com/p4lang/pna/pull/62 is now merged, can I get a review on this PR.

mbudiu-vmw (approved):

Issue #3508, opened 2022-07-13
Problem with building expression for dereference_header_stack


VolodymyrPeschanenkoIntel:
Model behavioral finished with the message [here](https://github.com/p4lang/behavioral-model/blob/f16d0de3486aa7fb2e1fe554aac7d237cc1adc33/src/bm_sim/P4Objects.cpp#L289) for attached example:
Invalid 'type' in expression: ''
bad json:
{
  "left" : {
     "type" : "header_stack",
     "value" : "h_0_h2"
   },
   "op" : "dereference_header_stack",
   "right" : {
   "type" : "field",
   "value" : [ "scalars", "tmp" ]
   }
}

This error message appears on the [invalid-hdr-warnings4.p4](https://github.com/p4lang/p4c/blob/main/testdata/p4_16_samples/invalid-hdr-warnings4.p4).
[invalid-hdr-warnings4.stf.txt](https://github.com/p4lang/behavioral-model/files/9103237/invalid-hdr-warnings4.stf.txt)

PR #3591, opened 2022-10-22
Run def-use analysis again after inlining 

mbudiu-vmw:
Inlining leads to many optimization opportunities.
Doing this has uncovered a few bugs in the analysis itself, which should be fixed in this PR.
Unfortunately many reference outputs change.


mbudiu-vmw:
test-tools has failed, but I don't know how to figure out what the problem is.

fruffy:
This actually fixes a bug in the generation of the JSON for the behavioral model (#3508). Not entirely sure why, but only the XFails [here](https://github.com/p4lang/p4c/blob/ffcfb76542e3b2eb705effc6ba7e59c270913592/backends/p4tools/testgen/targets/bmv2/test/BMV2Xfail.cmake#L46) need to be updated.

jafingerhut:
Some of the expected `-frontend` and `-midend` output files appear to get quite a lot shorter.  Is it worth checking whether they contain enough that the test program and expected output file still "cover" the compiler functionality we expect?  I know that might be a tedious task, and I can help with some of them in that regard if it is deemed important.

For example, look at the changes in this PR for the file decl2-frontend.p4.  Is there enough left to verify the compiler is not mis-transforming the input program?

testdata/p4_16_samples_outputs/issue3274-1-FrontEnd_41_SimplifyDefUse-first.p4
```diff
@@ -0,0 +1,12 @@
+void x() {
+}
```
jafingerhut, line 2: Should these expected output files with "FrontEnd_41_SimplifyDefUse" be in this PR?  I don't see any input P4 program in the p4c repo that corresponds to them.  Perhaps added into this PR by mistake due to some local new test P4 program you have?

mbudiu-vmw: I have deleted these.
In principle gauntlet checks equivalence of the code, so it should flag any file where the semantics changes.
But it does not check all files, and not between all compilation points.
@fruffy what are the representations checked? We should document this somewhere.
I have reviewed some -frontend files, but not all.

fruffy: Gauntlet currently only checks p4test's front and mid end. I have also disabled parser checking for Gauntlet (the missing implementation for extract causes issues). The only pass that is skipped is [FlattenHeaderUnion](https://github.com/p4gauntlet/toz3/blob/master/compare/compare.cpp#L14). I can try to add a README on this.

testdata/p4_16_samples_outputs/issue1642-bmv2-midend.p4
```diff
@@ -21,13 +21,13 @@ struct parsed_packet_t {
 
 struct local_metadata_t {
     short  _s0;
-    @field_list(0)
+    @field_list(0)
```
fruffy, line 24: Any reason why there is a space being added here?

testdata/p4_16_samples_outputs/invalid-hdr-warnings4-frontend.p4
```diff
@@ -39,51 +39,29 @@ control IngressI(inout H hdr, inout M meta, inout standard_metadata_t smeta) {
...
     }
     @name("IngressI.validateHeader") action validateHeader_1() {
-        hd_11 = h_0.h2[tmp];
         hd_11.setValid();
-        h_0.h2[tmp] = hd_11;
     }
     @name("IngressI.invalidateHeader") action invalidateHeader() {
-        hd_12 = h_0.h2[0];
         hd_12.setInvalid();
-        h_0.h2[0] = hd_12;
```
fruffy, line 66: Some of these seem quite aggressive. Gauntlet does not complain but I will try to check these manually.

fruffy (approved):

rst0git (approved):
LGTM

PR #3596, opened 2022-10-23
Remove expected output for non-existent action_param_serenum.p4 test program

jafingerhut:


jafingerhut:
@hesingh As far as I can tell these expected output files were added by you in 2019 when you were adding a feature, but there is no action_param_serenum.p4 test program that corresponds to them, and never was (at least not ever checked in).

If you have such a test program to add, feel free to add it, but if not, it seems best to delete these unused files.

mbudiu-vmw (approved):

hesingh:
Fine to remove them, thanks.

PR #3599, opened 2022-10-23
Add missing P4 test program issue2495-bmv2.p4

jafingerhut:


jafingerhut:
The expected output files have been checked in for some time, but there was no test program in p4_16_samples directory that ever used them for anything.  It looks like a good test case for an actual issue that was filed, so I copied the .p4 file from the samples_outputs directory to the samples directory.

mbudiu-vmw (approved):

PR #3606, opened 2022-10-23
Fix extra toP4 whitespace

### LINKED ISSUE #3517: Fix up broken BMv2 programs
This list of programs should compile with v1model, but does not because the programs are ill-formed in some way. 
```
  logging.p4
  issue584-1.p4
  issue1806.p4
  issue2104.p4
  issue2105.p4
  issue2890.p4
  noMatch.p4
  pred.p4
  pred1.p4
  pred2.p4
  register-serenum.p4
  table-key-serenum.p4
```

fruffy:
Fixes `toP4`'s extra whitespace that is added at the end of the program and after annotations. This PR changes lots of files, spot checking will be necessary.


jafingerhut:
I looked over the changed test programs in the p4_16_samples directory of this PR.  It appears that basically these test programs included v1model.p4, but were not written to use the v1model architecture, and your changes modify them so that they do use the v1model architecture.

I did not carefully check the original reason that these test programs were written.  It would be good if we did not unintionally change the reason they were created, and what cases within the compiler's implementation that they cover.  If these changes _did_ happen to do that, another possible change to the test program would be to no longer include the v1model.p4 file.

fruffy:
I chose to make them complete v1model programs to add more test cases for the behavioral model. However, just removing the include also works. I will also move those files to a separate PR since the diff has blown up.

fruffy:
Migrated the test file cleanup to #3608.

mbudiu-vmw (approved):

PR #3608, opened 2022-10-24
Remove unncessary v1model include in P4 test programs - fix up programs that use v1model externs.

### LINKED ISSUE #3517: Fix up broken BMv2 programs

fruffy:
This PR depends on #3606 for ref files. Fixes #3517. Instead of converting all of the programs to v1model this PR only converts the programs that use v1model externs.


mbudiu-vmw (approved):
This PR does not include the reference outputs

fruffy:
Yes, will update after #3606. #3591 should probably be merged before #3606, otherwise ref files will need to be updated, too.

pkotikal (approved):

testdata/p4_16_samples_outputs/issue1806-first.p4
```diff
@@ -1,6 +1,7 @@
 #include <core.p4>
-#define V1MODEL_VERSION 20180101
```
jafingerhut, line 2: I admit I am a bit fuzzy on this V1MODEL_VERSION variable that was used here, and do not know how to check whether removing it is a good idea or a bad idea in the test programs where you have removed it.   Any advice you might provide on helping guide my thinking about when this V1MODEL_VERSION is supposed to be used, vs. not?

jafingerhut (approved):
LGTM, with a minor question in there that I do not know how to answer, but I would guess you have looked into it more closely than I have.

testdata/p4_16_samples_outputs/issue1806-first.p4
fruffy: I believe this variable is added automatically by toP4 [here](https://github.com/p4lang/p4c/blob/main/frontends/p4/toP4/toP4.cpp#L178) whenever `v1model.p4` is included.

PR #3615, opened 2022-10-25
switch labels must be path expressions

### LINKED ISSUE #3611: Compiler bug with switch on action_run and not actions
Testcase:
```
action NoAction() {}
action a() {}
action b() {}
action d() {}
control c()
{
  table t {
    actions = {a; b;}
  }
  apply {
    switch(t.apply().action_run) {
      1: {} // { dg-error "" }
      default: {}
    }
  }
}
```
We hit the assert on frontends/p4/typeChecking/typeChecker.cpp:3807 :
```
            auto pe = c->label->to<IR::PathExpression>();
            CHECK_NULL(pe);
```
the label is not a path expression but rather a constant.

mbudiu-vmw:
Fixes #3611


jafingerhut (approved):
LGTM.  As usual, I am reviewing the test programs and expected outputs, not changes to the p4c internals.

frontends/p4/typeChecking/typeChecker.cpp
```diff
@@ -3803,14 +3803,17 @@ const IR::Node* TypeInference::postorder(IR::SwitchStatement* stat) {
         for (auto c : stat->cases) {
             if (c->label->is<IR::DefaultExpression>())
                 continue;
-            auto pe = c->label->to<IR::PathExpression>();
-            CHECK_NULL(pe);
```
apinski-cavium, line 3807: Wouldn't it be easier to just do:
```
if (pe == nullptr) {
                typeError("%1%: 'switch' label must be an action name or 'default'",
                          c->label);
  continue;
}
```
Also I think the check for default needs something extra because you can have duplicated default still.

mbudiu-vmw: I have improved the error messages too

PR #3621, opened 2022-10-25
Remove unused expected output files -- their test programs were renamed

jafingerhut:
Sigh.  I reviewed https://github.com/p4lang/p4c/pull/3608 myself, but failed to notice that these renamed tests would lead to unused expected output files.  Thankfully I've got a Python program now that catches them, which I plan to run monthly or so from now on: https://github.com/jafingerhut/p4-guide/blob/master/bin/sanity-check-p4c-testdata.py


fruffy (approved):

fruffy:
We can also just add this to CI, similar to cpplint.

PR #3624, opened 2022-10-26
Handle correctly general switch statements when simplifying

### LINKED ISSUE #3619: Compiler bug with switch fall through at the end (and no other statements)
Take:
```
control c(in bit<4> v)
{
  apply{
    switch(v) {
      default:
    }
  }
}
```

We get a warning like we should but then a compiler bug happens:
```
testsuite/p4c.language/valid/control/bitswitch-1.p4(6): [--Wwarn=missing] warning: SwitchCase: fallthrough with no statement
      default:
      ^^^^^^^
In file: /home/apinski/src/p4/p4c/frontends/p4/simplify.cpp:99
Compiler Bug: testsuite/p4c.language/valid/control/bitswitch-1.p4(4): v;: expected a Member
    switch(v) {
           ^
```
Yes the above code should not really appear in real code, I was just writing testcases for my front-end to make sure I did not mess it up and happened upon the crash here.

mbudiu-vmw:
Fixes #3619

### DIFF
--- a/frontends/p4/simplify.cpp
+++ b/frontends/p4/simplify.cpp
@@ -16,6 +16,7 @@ limitations under the License.
 
 #include "simplify.h"
 #include "sideEffects.h"
+#include "tableApply.h"
 
 namespace P4 {
 
@@ -94,12 +95,15 @@ const IR::Node* DoSimplifyControlFlow::postorder(IR::SwitchStatement* statement)
     if (statement->cases.empty()) {
         // If this is a table application remove the switch altogether but keep
         // the table application.  Otherwise remove the switch altogether.
-        if (auto mem = statement->expression->to<IR::Member>()) {
-            if (auto mce = mem->expr->to<IR::MethodCallExpression>()) {
-                LOG2("Removing switch statement " << statement << " keeping " << mce);
-                return new IR::MethodCallStatement(mce->srcInfo, mce);
-            }
+        if (TableApplySolver::isActionRun(statement->expression, refMap, typeMap)) {
+            auto mce = statement->expression->checkedTo<IR::Member>()->
+                    expr->checkedTo<IR::MethodCallExpression>();
+            LOG2("Removing switch statement " << statement << " keeping " << mce);
+            return new IR::MethodCallStatement(statement->srcInfo, mce);
         }
+        if (SideEffects::check(statement->expression, this, refMap, typeMap))
+            // This can happen if this pass is run before SideEffectOrdering.
+            return statement;
         LOG2("Removing switch statement " << statement);
         return nullptr;
     }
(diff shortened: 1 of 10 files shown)


apinski-cavium:
Does this handle side effects in the switch expression that is:
```
action NoAction() {}
control c(in bit<4> v)
{
  table t { actions = {} }
  apply{
    switch(t.apply().hit) {
      default:
    }
  }
}
```
And still have the table apply there?

mbudiu-vmw:
It should, if there is a method call in the switch statement it is preserved.
I was assuming this pass is run after the pass which moves side effects in separate statements, but I have to check.

mbudiu-vmw:
I don't think this is conservative enough, since it is also run before the side-effect-ordering pass.
I will improve it.

frontends/p4/simplify.cpp
```diff
@@ -92,16 +93,19 @@ const IR::Node* DoSimplifyControlFlow::postorder(IR::EmptyStatement* statement)
...
-        // switch conditions.  The parser should have rejected programs for
-        // which this is not the case.
-        BUG_CHECK(statement->expression->is<IR::Member>(),
-                  "%1%: expected a Member", statement->expression);
-        auto expr = statement->expression->to<IR::Member>();
-        BUG_CHECK(expr->expr->is<IR::MethodCallExpression>(),
-                  "%1%: expected a table invocation", expr->expr);
-        auto mce = expr->expr->to<IR::MethodCallExpression>();
-        return new IR::MethodCallStatement(mce->srcInfo, mce);
+        // If this is a table application remove the switch altogether but keep
```
fruffy, line 96: Do we have a test for this?

fruffy (approved):

frontends/p4/simplify.cpp
fruffy: Also another test for a function call with side-effects could be useful.

rst0git (approved):
LGTM

frontends/p4/simplify.cpp
mbudiu-vmw: issue3619-1.p4 is testing the side-effect.

fruffy: But that function call is side-effect free. Also this does not include a table.apply(), right?

mbudiu-vmw: This pass does not know that the function has no side effects. All tests with table applications would qualify for the second case.

fruffy:
```
if (SideEffects::check(statement->expression, this, refMap, typeMap))
    // This can happen if this pass is run before SideEffectOrdering.
    return statement;
```
I am trying to understand this segment. If the function call has side effects, ideally only the method call statement should remain, right? I do not think we have a test for this. 

Similarly, 
`if (TableApplySolver::isActionRun(statement->expression, refMap, typeMap)) {`
should check whether there is an action_run call and then only leave the table apply statement. 

I assume we have tests already for this case? Since this was the default behavior.

mbudiu-vmw: If the argument of the switch may have side-effects then we leave it unchaged. This pass is invoked again later, after SideEffectOrdering has removed calls with side effects from switch statements (all except table.apply().action_run), and that case will never occur again.

PR #3625, opened 2022-10-26
CI: add sanitizers

davidbolvansky:
Enable undefined behaviour sanitizer and address sanitizer.


apinski-cavium:
Note GCC has a similar (same) sanitizers that can be run there too.
I found GCC's address sanitizer finding more use out of scope issues than clang's does find.

Also I noticed you had to disable the leak detection. I could find you can only disable it at runtime via the `ASAN_OPTIONS=detect_leaks=0 ` env variable settings.

davidbolvansky:
And clang somehow crashes in CI.
So yes, I will switch to gcc build.

davidbolvansky:
Ready for review.

fruffy:
Do you want this as a nightly test by any chance? Or should we merge the sanitizer test with the normal test runner?

davidbolvansky:
As the normal one.

fruffy:
Then let's merge `build-linux` and `build-linux-sanitizers`.  `build-linux-sanitizers` seems like a superset to the other test to me and the additional sanitizer checks do not increase the compile time noticeably.

fruffy:
Are we actually looking at the “maybe-uninitialized” warnings? If not, I would go with the second option. 

Otherwise, we can keep the duplicate CI run. I am guessing we have to run the tests to trigger some sanitizer assertions, right?

davidbolvansky:
>> Are we actually looking at the “maybe-uninitialized” warnings? If not, I would go with the second option.

I look on them and they are highly unlikely to happen in practice. What is worse, sometimes innocent code change triggers for example gcc inliner and then gcc emits (from middle-ene!) warnings randomly..

Second option sounds fine to me, so I will update this patch. 

And yes to the last question.

fruffy (approved):

.github/workflows/ci-test.yml
```diff
@@ -32,9 +32,9 @@ jobs:
         key: test-${{ matrix.unified }}-${{ runner.os }}-gcc
         max-size: 1000M
 
-    - name: Build (Ubuntu Linux, GCC)
+    - name: Build (Ubuntu Linux, GCC, Sanitizers)
```
fruffy, line 35: Just a nit, maybe also rename the top-level test to include this name. Then the label immediately shows in the action panel.

davidbolvansky:
Feel free to merge it if you are fine with the current version of this PR, @fruffy

davidbolvansky:
Thanks for the review and feedback, @fruffy

PR #3626, opened 2022-10-27
An action profile implementation for v1model protobuf.

fruffy:
Implements [one-shot action selector/profile](https://p4.org/p4-spec/p4runtime/v1.3.0/P4Runtime-Spec.html#sec-oneshot) programming for the v1model protobuf back end.

### DIFF
--- a/backends/p4tools/testgen/targets/bmv2/table_stepper.cpp
+++ b/backends/p4tools/testgen/targets/bmv2/table_stepper.cpp
@@ -154,6 +154,105 @@ void BMv2_V1ModelTableStepper::evalTableActionProfile(
     }
 }
 
+void BMv2_V1ModelTableStepper::evalTableActionSelector(
+    const std::vector<const IR::ActionListElement*>& tableActionList) {
+    const auto* keys = table->getKey();
+    // If we have no keys, there is nothing to match.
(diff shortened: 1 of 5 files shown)


backends/p4tools/testgen/targets/bmv2/table_stepper.cpp
```diff
@@ -154,6 +154,105 @@ void BMv2_V1ModelTableStepper::evalTableActionProfile(
     }
 }
 
+void BMv2_V1ModelTableStepper::evalTableActionSelector(
+    const std::vector<const IR::ActionListElement*>& tableActionList) {
+    const auto* keys = table->getKey();
+    // If we have no keys, there is nothing to match.
+    if (keys == nullptr) {
```
pkotikal, line 161: Do we need to handle the default action, even in this case?

pkotikal (approved):
Are we removing the commented out code?

backends/p4tools/testgen/targets/bmv2/table_stepper.cpp
fruffy: Thanks for catching that, fixed. 

We have no support for default action override yet, which is why we return here. We should consolidate this code once we implement default action overrides.

PR #3630, opened 2022-10-27
Migrate some P4Testgen utilities to lib folder.

fruffy:
This is in preparation to migrate some files from the P4Tools folder to the compiler library folders.


fruffy:
Not sure what to do about Extract_Varbits node. It can not be cached or checked since it is part of the Tools IR. This means getConstant either has to moved to the tools back end or the Extract_Varbits IR type has to be moved to the core IR. Both solutions are not ideal.

ir/irutils.h
```diff
@@ -0,0 +1,98 @@
...
+#include "lib/error.h"
+#include "lib/exceptions.h"
+
+namespace IR {
+
+/// Utility functions for generating IR nodes.
+//
+// Some of these are just thin wrappers around functions in the IR, but it's nice having everything
+// in one place.
+class IRUtils {
```
ChrisDodd, line 23: since this class has no instances and no non-static members, maybe it should be a namespace?

or maybe even put all these functions in namespace IR directly.

backends/p4tools/common/lib/util.cpp
```diff
@@ -42,27 +47,126 @@ std::string TestgenUtils::getTimeStamp() {
...
+    return IR::getConstant(type, randInt);
+}
+
+/* =========================================================================================
+ *  Variables and symbolic constants.
+ * ========================================================================================= */
+
+const cstring Utils::Valid = "*valid";
+
+const StateVariable& Utils::getZombieTableVar(const IR::Type* type, const IR::P4Table* table,
```
mbudiu-vmw, line 92: this part is not really migration

mbudiu-vmw (approved):

PR #3631, opened 2022-10-28
DPDK Backend: Direct counter and direct meter support

usha1830:
This PR adds support for DirectCounter and DirectMeter extern in .spec file and context json output files for dpdk compiler

In the spec file
- For direct counter , a new regarray is declared for direct counter extern with size equal to the owner table size + 1 for regular tables and table_size *4 + 1 for learner tables.
- Similarly, for direct meter,  a new metarray is declared for direct counter extern with size equal to the owner table size + 1 for regular tables and table_size *4 + 1 for learner tables.

For the count method of direct counter, the following instructions should be emitted.
```
entryid m.table_entry_index
regadd per_prefix_pkt_bytes_count_0_packets m.table_entry_index 1
```

In case of PACKETS_AND_BYTES counter, two regarrays should be emitted with the counter name suffixed with "_packets" and "_bytes"
Similarly, the count method also should have "regadd" instruction for packets and bytes counter.

For the execute method of direct meter, the following instructions should be emitted.
```
entryid m.table_entry_index
meter <metername> m.table_entry_index <packet_length> <color_in> <color_out>
```

Following tests are added:
- Valid case for regular table using Direct Counter
- Valid case for regular table using Direct Meter
- Valid case for learner table using Direct Counter
- Valid case for learner table using Direct Meter
- Error cases for coexistence of action selector and direct counter/direct meter
- Error for default action not having count/execute method
- Error for calling count/execute of multiple extern instances in same action
- Error for calling count/execute from non-owner table action
- Error for using Direct Meter/Direct Counter extern with tables having wildcard/lpm match kind
- Error for calling count/execute from apply block


jfingerh:
@usha1830  Out of curiosity, why "table_size *4 + 1 for learner tables" ?  Why not "table_size + 1" for learner tables?

usha1830:
> @usha1830 Out of curiosity, why "table_size *4 + 1 for learner tables" ? Why not "table_size + 1" for learner tables?

This is based on internal implementation in dpdk target. There are 4 buckets for each index. 
Reference : https://git.dpdk.org/dpdk/tree/lib/table/rte_swx_table_learner.c#n243

testdata/p4_16_samples_outputs/pna-add-on-miss.p4.spec
```diff
@@ -61,7 +61,7 @@ action next_hop2 args instanceof next_hop2_arg_t {
 
 action add_on_miss_action2 args none {
 	mov m.MainControlT_tmp 0x0
-	mov m.MainControlT_tmp_0 0x4d2
+	mov m.MainControlT_tmp_0 0x4D2
```
usha1830, line 64: I don't know why this change is occurring.
After syncing with main, I regenerated the output file and see this. All hex constants are changed to uppercase.

backends/dpdk/dpdkArch.cpp
```diff
@@ -2374,6 +2402,191 @@ const IR::Node* SplitP4TableCommon::postorder(IR::SwitchStatement* statement) {
...
+    cstring externName = a->originalExternType->getName().name;
+    if (externName == "DirectMeter" || externName == "DirectCounter") {
+        auto di = a->object->to<IR::Declaration_Instance>();
+        cstring instanceName = di->name.name;
+        if (a->method->getName().name == method) {
+            if (instancename != "" && instanceName == instancename) {
+                found = true;
+            }
+            if (oneInstance == "")
+            oneInstance = instanceName;
```
mbudiu-vmw, line 2428: indentation

backends/dpdk/dpdkArch.cpp
```diff
@@ -2374,6 +2402,191 @@ const IR::Node* SplitP4TableCommon::postorder(IR::SwitchStatement* statement) {
...
+        }
+    }
+    return found;
+}
+
+bool CollectDirectCounterMeter::ifMethodFound(const IR::P4Action *a,
+                        cstring method, cstring instancename) {
+    cstring oneInstance = "";
+    bool found = false;
+    for (auto stmt : a->body->components) {
```
mbudiu-vmw, line 2445: you could do this by using a small visitor. it would be more robust and handle nested blocks, etc.

backends/dpdk/dpdkArch.cpp
```diff
@@ -2374,6 +2402,191 @@ const IR::Node* SplitP4TableCommon::postorder(IR::SwitchStatement* statement) {
...
+                    found = checkMethodCallInAction(a, method, oneInstance, instancename);
+                }
+            }
+        }
+    }
+    return found;
+}
+
+void CollectDirectCounterMeter::postorder(const IR::P4Action* a) {
+    ifMethodFound(a, "count");
```
mbudiu-vmw, line 2465: you ignore the result, that's strange

backends/dpdk/dpdkArch.cpp
```diff
@@ -2374,6 +2402,191 @@ const IR::Node* SplitP4TableCommon::postorder(IR::SwitchStatement* statement) {
...
+    auto table_type = ::get(structure->table_type_map, tbl->name.name);
+
+    // Direct Counter and Meter are not supported with Wildcard match tables
+    if (table_type == InternalTableType::WILDCARD && (counterInstance || meterInstance)) {
+        ::error(ErrorType::ERR_UNSUPPORTED_ON_TARGET, "Direct counters and direct meters are"
+                " unsupported for wildcard match table %1%", tbl->name);
+        return;
+    }
+    if (table_type == InternalTableType::LEARNER)
+            table_size *= 4;
```
mbudiu-vmw, line 2499: inconsistent indentation

backends/dpdk/dpdkArch.cpp
```diff
@@ -2374,6 +2402,191 @@ const IR::Node* SplitP4TableCommon::postorder(IR::SwitchStatement* statement) {
...
+        directMeterCounterSizeMap.emplace(meterExternName, table_size + 1);
+        structure->direct_resource_map.emplace(meterExternName, tbl);
+    }
+}
+
+void ValidateDirectCounterMeter::validateMethodInvocation(P4::ExternMethod *a) {
+    if ((a->originalExternType->getName().name == "DirectCounter" &&
+        a->method->getName().name == "count") ||
+        (a->originalExternType->getName().name == "DirectMeter" &&
+        a->method->getName().name == "execute")) {
```
mbudiu-vmw, line 2545: what if the method does not have this name?

backends/dpdk/dpdkArch.cpp
```diff
@@ -2374,6 +2402,191 @@ const IR::Node* SplitP4TableCommon::postorder(IR::SwitchStatement* statement) {
...
+            if (!act) {
+                ::error(ErrorType::ERR_UNEXPECTED," %1% method of %2% extern "
+                "must only be called from within an action", a->method->getName().name,
+                di->name.originalName);
+                return;
+            }
+            for (auto action : ownerTable->getActionList()->actionList) {
+                auto action_decl = refMap->getDeclaration(action->getPath())->to<IR::P4Action>();
+                    if (act->name.name == action_decl->name.name) {
+                    invokedFromOwnerTable = true;
```
mbudiu-vmw, line 2559: indentation

backends/dpdk/dpdkHelpers.cpp
```diff
@@ -368,6 +368,31 @@ bool ConvertStatementToDpdk::preorder(const IR::AssignmentStatement *a) {
                     i = new IR::DpdkMeterExecuteStatement(
                          e->object->getName(), index, length, color_in, left);
                 }
+            } else if (e->originalExternType->getName().name == "DirectMeter") {
+                if (e->method->getName().name == "execute") {
+                    auto argSize = e->expr->arguments->size();
+
+                    // DPDK target needs packet length as mandatory parameters
+                    if (argSize < 1) {
+                        ::error(ErrorType::ERR_UNEXPECTED, "Expected atleast 1 argument "
```
mbudiu-vmw, line 377: missing space

backends/dpdk/dpdkHelpers.cpp
```diff
@@ -1030,6 +1055,52 @@ bool ConvertStatementToDpdk::preorder(const IR::MethodCallStatement *s) {
...
+        } else if (a->originalExternType->getName().name == "DirectMeter") {
+            if (a->method->getName().name != "execute") {
+                BUG("Direct Meter function %1% not implemented.", a->method->getName().name);
+            }
+        } else if (a->originalExternType->getName().name == "DirectCounter") {
+            auto di = a->object->to<IR::Declaration_Instance>();
+            auto declArgs = di->arguments;
+            unsigned value = 0;
+            auto counter_type = declArgs->at(0)->expression;
+            if (counter_type->is<IR::Constant>())
```
mbudiu-vmw, line 1067: usually you write if (auto c = counter_type->to<IR::Constant>()) value = c->asUnsigned().

backends/dpdk/dpdkHelpers.cpp
```diff
@@ -1030,6 +1055,52 @@ bool ConvertStatementToDpdk::preorder(const IR::MethodCallStatement *s) {
...
+                value = counter_type->to<IR::Constant>()->asUnsigned();
+            if (a->method->getName().name == "count") {
+                auto args = a->expr->arguments;
+                if (args->size() > 1){
```
mbudiu-vmw, line 1071: we usually use spaces before {

backends/dpdk/spec.cpp
```diff
@@ -117,6 +153,22 @@ std::ostream &IR::DpdkExternDeclaration::toSpec(std::ostream &out) const {
...
+    } else if (DPDK::toStr(this->getType()) == "DirectMeter") {
+        auto args = this->arguments;
+        if (args->size() < 1) {
+            ::error(ErrorType::ERR_INVALID,
+                    "Meter extern declaration %1% must have "
+                    "meter type parameter", this->Name());
+        } else {
+            IR::Expression *n_meters =nullptr;
+            if (directMeterCounterSizeMap.count(this->Name())) {
+                n_meters = new IR::Constant(directMeterCounterSizeMap.at(this->Name()));
```
mbudiu-vmw, line 165: we usually omit this-> if it is not ambiguous

mbudiu-vmw (approved):

backends/dpdk/dpdkArch.cpp
jafingerhut: I suspect the answer is "this is implementing the PSA/PNA architecture definitions of DirectCounter and DirectMeter, and the method names are defined in the spec".  So the methods won't have any other names.

mbudiu-vmw: that's why an error should be given

jafingerhut: Sorry, I misunderstood the reason for your question.  Carry on :-)

backends/dpdk/dpdkArch.cpp
usha1830: I didn't get this comment correctly.
A visitor for BlockStatement and then calling visit(a->body) here?
Visitors for MethodCallStatement and AssignmentStatement  would also be needed I guess.

backends/dpdk/dpdkArch.cpp
usha1830: ifMethodFound function is called from 2 places. 
One for making sure that the default action of owner table of the Direct Meter/Counter extern instance has a execute/count method. In this case, the return value is checked to issue error if specified method call doesn't exist in default action.
Second, for making sure that an action only contains count/execute method calls for only one Direct counter/meter instance. In this case, the error is emitted in the ifMethodFound function itself and hence we are not using the return value here.

backends/dpdk/dpdkArch.cpp
usha1830: Yes. I will add error here

backends/dpdk/dpdkArch.cpp
mbudiu-vmw: You should probably create a new visitor instead of a loop

backends/dpdk/dpdkArch.cpp
mbudiu-vmw: This would be a good comment

mbudiu-vmw:
You should rebase this

usha1830:
@mbudiu-vmw Addressed all review comments. Please check.
Last commit https://github.com/p4lang/p4c/pull/3631/commits/ea75b39084fbe8c222f7de3d5f26c3cbb97463e5  is a minor change on top the existing PR. It adds table id for the table with which the direct meter or counter is associated. Please review.

mbudiu-vmw (approved):

PR #3632, opened 2022-10-28
[P4Testgen] Implementation for "Counter.count" extern function 

OleksandrLetychevskyiIntel:
Implementation of the external function `Counter.count`. An attempt to use the `register` implementation for `counter`, since according to the `v1model` documentation, `counter` and `register` are quite similar objects


backends/p4tools/testgen/targets/bmv2/expr_stepper.cpp
```diff
@@ -585,13 +585,77 @@ void BMv2_V1ModelExprStepper::evalExternMethodCall(const IR::MethodCallExpressio
...
+             // index. If the counter has not been added had, create a new counter object.
+             // "Count" increment an entry by one. If index >= size, no counter state will be
+             // updated.
+             const auto* counterState =
+                 state.getTestObject("countervalues", externInstance->toString(), false);
+
+             Bmv2RegisterValue* counterValue = nullptr;
+             if (counterState != nullptr) {
+                 counterValue =
+                     new Bmv2RegisterValue(*counterState->checkedTo<Bmv2RegisterValue>());
```
fruffy, line 637: If you use a register object, you will have no means to differentiate them in the test back end. There is a lot of code overlap, but you should be able to tell when you have a counter or a register.

fruffy (changes requested):

fruffy:
Can you add an example test that exercises these counters appropriately?

PR #3637, opened 2022-10-29
Update comments in issue2201-bmv2.p4 to match current behavior

jafingerhut:
This makes the example program much more useful to show to people who want to know how BMv2 simple_switch implements the log_msg extern function, as of 2022-Oct (it has not changed in quite some time before that).

Remove issue2201-1-bmv2.p4 since it had strictly less test coverage than issue2201-bmv2.p4.


mbudiu-vmw (approved):
no one will find this file with this name

apinski-cavium:
> no one will find this file with this name

Funny you named testcase that way: https://github.com/p4lang/p4c/pull/2462 .

I do think we should have some coding style on how test cases are named. naming them after the issue # is not always useful as obvious from this issue and such.

jafingerhut:
"no one will find this file with this name" Agreed.  If we want to document the behaviors here at some point in time, e.g. because we want to add the log_msg extern to the PSA and/or PNA specs, then at least this test case will contain accurate information to start that from, as opposed to obsolete information.

PR #3646, opened 2022-10-31
Restructure P4Tools to easily add new modules.

fruffy:
This PR restructures P4Tools such that it is easier to add new modules to the framework. The PR removes the p4check binary and instead adds a binary per module. This means, instead of `p4check testgen` one can now call `p4testgen` directly. 

To add a new module, a folder inside `p4tools/modules` with a corresponding `CMakelists` file needs to be created. The CMake script should pick this up directly.


backends/p4tools/modules/testgen/CMakeLists.txt
```diff
@@ -96,13 +96,29 @@ endforeach()
...
+
+target_link_libraries(
+  testgen
+  ${TESTGEN_LIBS}
+)
+
+add_p4tools_executable(p4testgen main.cpp)
 
 target_link_libraries(
   p4testgen
```
pkotikal, line 109: Should `p4testgen` be removed?

backends/p4tools/modules/testgen/main.cpp
```diff
@@ -0,0 +1,34 @@
...
+    setup_signals();
+
+    std::vector<const char*> args;
+    for (int i = 0; i < argc; ++i) {
+        args.push_back(argv[i]);
+    }
+
+    try {
+        P4Tools::P4Testgen::Testgen().main(args);
+    } catch (const Util::CompilerBug& e) {
```
pkotikal, line 16: We should add a util function and replace `CompilerBug` to `ToolsBug`, `TestgenBug` or something (not in this PR).

backends/p4tools/modules/testgen/targets/bmv2/test/P4Tests.cmake
```diff
@@ -3,7 +3,7 @@ include(${CMAKE_CURRENT_LIST_DIR}/../../../cmake/TestUtils.cmake)
 # Add v1model tests from the p4c submodule
 set(P4TOOLS_BINARY_DIR ${CMAKE_SOURCE_DIR}/build)
 set(P4TESTGEN_DIR ${CMAKE_SOURCE_DIR}/build/testgen)
-set(P4TESTGEN_DRIVER "${P4TOOLS_BINARY_DIR}/p4check testgen")
```
pkotikal, line 6: So `p4check` doesn't exist at all now. Any module needs to be invoked via it's name like `p4testgen`?

pkotikal (approved):
Looks good, thank you!

backends/p4tools/modules/testgen/targets/bmv2/test/P4Tests.cmake
fruffy: Yes, after this PR `p4check` does not exist any more. All tests are run with the `p4testgen` binary.

Issue #3636, opened 2022-10-29
Test programs that would be nice to fix


jfingerh:
I happened to pick one of these to copy and paste an IPv4 definition from for my own testing purposes, and was very confused when they did not parse correctly.  It turns out they are not correct definitions of IPv4 headers.  Unless the test programs are intended to exercise something related to an incorrectly defined IPv4 header, I'd recommend correcting them.

The IPv4 total length field is 16 bits wide, not 32 bits wide.

@usha1830 Do you want me to fix these?  Or know someone (perhaps you) who wants to?

```
psa-example-dpdk-byte-alignment_1.p4:    bit<32> totalLen;
psa-example-dpdk-byte-alignment_2.p4:    bit<32> totalLen;
psa-example-dpdk-byte-alignment_3.p4:    bit<32> totalLen;
psa-example-dpdk-byte-alignment_5.p4:    bit<32> totalLen;
psa-example-dpdk-byte-alignment_6.p4:    bit<32> totalLen;
psa-example-dpdk-byte-alignment_7.p4:    bit<32> totalLen;
psa-example-dpdk-byte-alignment_8.p4:    bit<32> totalLen;
psa-example-dpdk-byte-alignment_9.p4:    bit<32> totalLen;
psa-example-dpdk-externs.p4:    bit<32> totalLen;
psa-example-dpdk-meter1.p4:    bit<32> totalLen;
psa-hash-03.p4:    bit<32> totalLen;
psa-hash-04.p4:    bit<32> totalLen;
psa-hash-07.p4:    bit<32> totalLen;
psa-hash-08.p4:    bit<32> totalLen;
psa-hash-09.p4:    bit<32> totalLen;
psa-hash-10.p4:    bit<32> totalLen;
```

PR #3647, opened 2022-11-01
Fix incorrect tests : totalLen field of ipv4 header, action name in add_entry call

usha1830:
Fixes https://github.com/p4lang/p4c/issues/3636. 
Also fixes the incorrect action name in add_entry call for few tests


usha1830:
Some tests have some user- defined struct field within the header (testdata/p4_16_samples/psa-example-dpdk-byte-alignment_6.p4, testdata/p4_16_samples/psa-example-dpdk-byte-alignment_8.p4 etc) . This was added for testing byte alignment feature for struct field within header which is initialized using {}.

I have not removed such fields, only modified the totalLen field and its usage in the program.

jafingerhut (approved):
LGTM.

PR #3651, opened 2022-11-02
Tighten checks for NoAction

### LINKED ISSUE #3644: Compiler bug with invalid definition of NoAction
Take:
```
const bit NoAction = 1;
control c()
{
  table t {
    actions = {}
  }
  apply {}
}
```
This gives a compiler bug error message instead of saying NoAction is defined incorrectly.
Found this while designing my front-end to make sure NoAction is defined correctly when adding the default_action property if it does not exist.

mbudiu-vmw:
Fixes #3644

### DIFF
--- a/testdata/p4_16_errors_outputs/missing_match.p4-stderr
+++ b/testdata/p4_16_errors_outputs/missing_match.p4-stderr
@@ -1,3 +1,1 @@
-missing_match.p4(20): [--Werror=not-found] error: noSuchMatch: declaration not found
-        key = { b : noSuchMatch; }
-                    ^^^^^^^^^^^
+[--Werror=Target model error] error: Declaration of the action NoAction not found; did you include core.p4?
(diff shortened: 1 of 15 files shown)


testdata/p4_16_errors_outputs/missing_match.p4-stderr
```diff
@@ -1,3 +1 @@
-missing_match.p4(20): [--Werror=not-found] error: noSuchMatch: declaration not found
-        key = { b : noSuchMatch; }
-                    ^^^^^^^^^^^
+[--Werror=Target model error] error: Declaration of the action NoAction not found; did you include core.p4?
```
jfingerh, line 1: Should we modify the test program missing_match.p4 so that it gives a similar error message as it did before?  If we do not do that, I worry slightly that we are losing test coverage for the error that this program was exercising before.

mbudiu-vmw: Done.

jfingerh (approved):
LGTM.  As usual, I have reviewed the test programs and their expected outputs only, not the internal compiler implementation detials.

mbudiu-vmw:
@jafingerhut you left the review with an ID which cannot approve PRs

jafingerhut (approved):

PR #3664, opened 2022-11-04
Remove useless grammar rule

mbudiu-vmw:

Turns out that `type` declarations are restricted to base types, so the grammar rule that allows derived types is useless.
This could technically be a backwards-incompatible change. To be discussed by the LDWG together with https://github.com/p4lang/p4-spec/issues/1044


apinski-cavium:
In my front-end I didn't implement this part of the spec because of the spec I had raised
so +1 from me.

jafingerhut (approved):
Looks reasonable to me.

PR #3667, opened 2022-11-05
Experimental: literal for invalid header

mbudiu-vmw:
This is a prerequisite for finalizing the implementation of the ... default initialization operator.


mbudiu-vmw:
I have chosen the syntax (H){#} to denote an invalid header with type H.
We should debate this during the next LDWG meeting.

vgurevich:
So, it is true that these two will be equivalent:
```
header.setInvalid();`
```
and 
```
header = (header_t){#}
```

mbudiu-vmw:
They are equivalent

jafingerhut (approved):
The syntax looks reasonable to me.  I say go for it!

mbudiu-vmw:
In the spirit of past experiments I will actually merge this PR.

Issue #3258, opened 2022-04-27
Addition assignment (+=) operator


rst0git:
Many programming languages have support for operators such as addition assignment (`+=`). However, the P4 compiler currently does not support this operator. For example, running `p4test` on the following code would fail:
```P4
action foo() {
    bit<32> a = 1;
    a += 1;
}
```
```bash
$ p4test test.p4
test.p4(3):syntax error, unexpected +
    a +
      ^
[--Werror=overlimit] error: 1 errors encountered, aborting compilation
```
But the following works as expected:
```P4
action foo() {
    bit<32> a = 1;
    a = a + 1;
}
```
Is this intentionally not supported or it has not been implemented yet?

mbudiu-vmw:
This should be discussed in the design group first. Do you want to write a pr against the spec?

rst0git:
> Do you want to write a pr against the spec?

Yes, I would be happy to do that.

mbudiu-vmw:
@rst0git I have filed this LDWG design issue to track this feature: https://github.com/p4lang/p4-spec/issues/1139

PR #3669, opened 2022-11-06
Add support for compound assignment operators

rst0git:
This pull request adds support for compound assignment expressions of the form `E1 op= E2` as discussed in https://github.com/p4lang/p4-spec/pull/1144 and https://github.com/p4lang/p4-spec/issues/1139.

Fixes https://github.com/p4lang/p4c/issues/3258

### DIFF
--- a/frontends/parsers/p4/p4parser.ypp
+++ b/frontends/parsers/p4/p4parser.ypp
@@ -206,6 +206,18 @@ inline std::ostream& operator<<(std::ostream& out, const P4::Token& t) {
 %token<Token>      QUESTION "?"
 %token<Token>      DOT "."
 %token<Token>      ASSIGN "="
+%token<Token>      ASSIGN_PLUS "+="
+%token<Token>      ASSIGN_MINUS "-="
+%token<Token>      ASSIGN_MUL "*="
+%token<Token>      ASSIGN_DIV "/="
+%token<Token>      ASSIGN_MOD "%="
+%token<Token>      ASSIGN_BIT_AND "&="
+%token<Token>      ASSIGN_BIT_OR "|="
+%token<Token>      ASSIGN_XOR "^="
+%token<Token>      ASSIGN_SHL "<<="
+%token<Token>      ASSIGN_SHR ">>="
+%token<Token>      ASSIGN_PLUS_SAT "|+|="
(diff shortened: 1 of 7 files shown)


frontends/parsers/p4/p4parser.ypp
```diff
@@ -1161,6 +1173,54 @@ assignmentOrMethodCallStatement
                                                  $1, $3, $6);
           $$ = new IR::MethodCallStatement(@1 + @7, mc); }
 
+    | lvalue "+=" expression ";"
```
mbudiu-vmw, line 1176: I am not sure this is the right semantics. This implementation will repeat the side effects on the lhs twice.

apinski-cavium: Note most lvalue by definition could not have a side effect.
The only one which I can think of is: `headerstack[f(a)].b`  where f is defined to take 1 arguments which is inout (you could do something similar with externs too).

jnfoster:
We will need to be careful if we want to ensure that l-values never have side effects. I am not even sure that's true of the current language design and it would be fragile to ensure it going forward.

jnfoster:
@rst0git are you still interested in completing this feature? Do you need help with the `p4c` implementation?

jnfoster:
@rst0git, can you let us know if you are still working on this? We'd be glad to see it added, but need to fix the side-effect issue...

rst0git:
@jnfoster I'm sorry for the delayed response. I will update the pull request in the next few days.

frontends/parsers/p4/p4parser.ypp
ChrisDodd: regardless of the inout arguments, a method call may have arbitrary side effects within in, and the way this code duplicates the reference to the lhs, a method call inside an index would end up being called twice.  We have a pass (DoSimplifyExpressions) that exists to deal with this kind of program, but there are a number of other passes that currently must run before it (such as typeChecking) that would need to understand what is going on.  So this almost certainly needs a new IR node type (some kind of op-assignment statement) to represent this until SimplifyExpressions can turn it into a normal AssignmentStatement, and all the passes that run before need to know about it.

ChrisDodd:
Superceeded by #5108

PR #3673, opened 2022-11-07
Build DCG if pattern flag was enabled

VolodymyrPeschanenkoIntel:
These changes allow to build DCG for p4tools if pattern flag was enabled. Otherwise, reachability engine featured will not work.

### DIFF
--- a/backends/p4tools/testgen/core/program_info.cpp
+++ b/backends/p4tools/testgen/core/program_info.cpp
@@ -15,7 +15,7 @@ ProgramInfo::ProgramInfo(const IR::P4Program* program)
       concolicMethodImpls({}),
       program(program) {
     concolicMethodImpls.add(*Concolic::getCoreConcolicMethodImpls());
-    if (TestgenOptions::get().dcg) {
+    if (TestgenOptions::get().dcg || TestgenOptions::get().pattern.length()) {
         // Create DCG.
         P4::ReferenceMap refMap;
         P4::TypeMap typeMap;


fruffy:
We should be running gtest, I wonder why these do not fail on CI.

vhavel (approved):

fruffy:
> > We should be running gtest, I wonder why these do not fail on CI.
> 
> I cannot find any place where these would be run, in p4tools CI workflow we're only adding `testgen-` ctests
> 
> https://github.com/p4lang/p4c/blob/e4369e08fa97a3327a992fc76fdec1cd5a60bf76/.github/workflows/ci-p4tools.yml#L42

It looks like the gtest are named poorly and this is why they never ran. 
https://github.com/p4lang/p4c/blob/main/backends/p4tools/testgen/CMakeLists.txt#L112
The name should be testgen-gtest. 

@VolodymyrPeschanenkoIntel  Can you change the label in this PR, too?

.github/workflows/ci-p4tools.yml
```diff
@@ -39,4 +39,4 @@ jobs:
       # this is needed to create network namespaces for the ebpf tests.
     - name: Run tests (Ubuntu Linux)
       run: |
-        sudo docker run --privileged -w /p4c/build -e $CTEST_PARALLEL_LEVEL p4c ctest -R testgen- -j2 --output-on-failure --schedule-random
+        sudo docker run --privileged -w /p4c/build -e $CTEST_PARALLEL_LEVEL p4c ctest -R gtest-p4testgen|testgen- -j2 --output-on-failure --schedule-random
```
fruffy, line 42: It might make more sense to just rename the gtest label directly in the CMake file. Is there a reason why this does not work?

VolodymyrPeschanenkoIntel: > It might make more sense to just rename the gtest label directly in the CMake file. Is there a reason why this does not work?

This is make sense, I'll do that.

fruffy: Thanks!

backends/p4tools/testgen/test/small-step/reachability.cpp
```diff
@@ -373,6 +373,7 @@ void callTestgen(const char* inputFile, const char* behavior, const char* path,
     cmdTestgen << "--pattern \"" << behavior << "\" ";
     cmdTestgen << "--out-dir \"" << buildPath << path << "\" \"" << sourcePath << prefix;
     cmdTestgen << inputFile << "\"";
+    std::cout << cmdTestgen.str() << std::endl;
```
fruffy, line 376: Spurious print.

fruffy (changes requested):

fruffy (approved):

PR #3677, opened 2022-11-08
Rename psabpf repository into NIKSS

tatry:
With the latest version of `psabpf` we have changed its name to `NIKSS` (Native In-Kernel SDN Switch) as well the location of repository to https://github.com/NIKSS-vSwitch/nikss (old link should still work, but it is recommended to use new one).


rst0git (approved):
LGTM

backends/ebpf/tests/ptf/common.py
```diff
@@ -124,15 +124,15 @@ def setUp(self):
         self.dataplane = ptf.dataplane_instance
         self.dataplane.flush()
 
-        self.exec_ns_cmd("psabpf-ctl pipeline load id {} {}".format(TEST_PIPELINE_ID, self.test_prog_image), "Can't load programs into eBPF subsystem")
+        self.exec_ns_cmd("nikss-ctl pipeline load id {} {}".format(TEST_PIPELINE_ID, self.test_prog_image), "Can't load programs into eBPF subsystem")
```
mbudiu-vmw, line 127: Maybe the name should be in a variable

mbudiu-vmw (approved):

PR #3678, opened 2022-11-08
Run IWYU on P4Tools.

fruffy:
Just cleaning up includes.


rst0git:
@fruffy I was wondering if the changes in this pull request are auto-generated and if it would be possible to run the IWYU tool in our CI?

fruffy:
The changes are autogenerated but IWYU itself is not quite reliable enough to run it in CI. First, it builds quite slowly and, second, the CMake version does not seem to throw errors as expected.

It will also take a while to clean up the includes for p4c. I started with the ir folder, which is a big mess. I actually have an upcoming PR for this.

rst0git (approved):

pkotikal (approved):

fruffy:
Thanks!

PR #3683, opened 2022-11-09
Added collapsedChain to IR::ParserState

y-lavoie:
Knowing what original states made up a new merged state created by CollapseChains in SimplifyParsers allows more flexibility in the backend.

### DIFF
--- a/ir/ir.def
+++ b/ir/ir.def
@@ -60,6 +60,7 @@ class ParserState : ISimpleNamespace, Declaration, IAnnotated {
     optional inline IndexedVector<StatOrDecl>   components;
     // selectExpression can be a SelectExpression, or a PathExpression representing a state
     NullOK Expression                   selectExpression;
+    optional std::vector<ParserState> collapsedChain;
 
     Annotations getAnnotations() const override { return annotations; }
     Util::Enumerator<IDeclaration>* getDeclarations() const override {
--- a/frontends/p4/simplifyParsers.cpp
+++ b/frontends/p4/simplifyParsers.cpp
@@ -161,11 +161,13 @@ class CollapseChains : public Transform {
             if (chainStart.find(s) != chainStart.end()) {
                 // collapse chain
                 auto components = new IR::IndexedVector<IR::StatOrDecl>();
+                std::vector<const IR::ParserState*> collapsedChain;
                 auto crt = s;
                 LOG1("Chaining states into " << dbp(crt));
                 const IR::Expression *select = nullptr;
                 while (true) {
                     components->append(crt->components);
+                    collapsedChain.push_back(crt);
                     select = crt->selectExpression;
                     crt = ::get(chain, crt);
                     if (crt == nullptr)
@@ -173,7 +175,7 @@ class CollapseChains : public Transform {
                     LOG1("Adding " << dbp(crt) << " to chain");
                 }
                 s = new IR::ParserState(s->srcInfo, s->name, s->annotations,
-                                        *components, select);
+                                        *components, select, collapsedChain);
             }
             states->push_back(s);
         }


ir/ir.def
```diff
@@ -60,6 +60,7 @@ class ParserState : ISimpleNamespace, Declaration, IAnnotated {
     optional inline IndexedVector<StatOrDecl>   components;
     // selectExpression can be a SelectExpression, or a PathExpression representing a state
     NullOK Expression                   selectExpression;
+    optional std::vector<ParserState> collapsedChain;
```
mbudiu-vmw, line 63: I think this kind of information belongs in an annotation.

y-lavoie: It's not easy to put the original IR into an Annotation. One benefit of this approach is that it allows you to easily uncollapse the chain back into its original states or have better visibility to what the IR originally looked like.

mbudiu-vmw: The state will carry all the original states in its representation, all the visitors will continue to analyze and transform these states, although they no longer technically exist in the program. I expect this will cause some serious maintenance problems in the future. Who needs better visibility into the original IR? Is this for debugging by humans? Or you want to write some visitor that undoes the tranformation later? We have backtracking for this kind of stuff.

y-lavoie: If the IR was stored in an `IR::Vector`, then yes, those states would be visited and cause the issue you are describing. However, IR in a `std::vector` will not get visited by visitors, so I don't think that problem applies here.

I do want to write a pass that can undo this transformation, yes.

y-lavoie: I observed that the states stored in my proposed changes no longer appear in the P4 program after the SimplifyParsers pass.

mbudiu-vmw:
Maybe you can describe your design and we can see what the best way to do it is. It's true that vector is not visited.

apinski-cavium:
Seems like you could split the parser state into two in some cases.
E.g. if you have:
```
state a {
  statement1;
  statement2;
  transition select ...
}
```
Couldn't you transform this into:
```
state a {
  statement1;
  transition a_part2;
}
state a_part2 {
  statement2;
  transition select ...
}
```

As long as you promote the variables used across statement 1 and statement 2.

y-lavoie:
Current behaviour in the frontend is this:

Before CollapseChains:
```
state a {
  statement1;
  transition b;
}

state b {
  statement2;
  transition select ...
}
```

After CollapseChains:
```
state a {
  statement1;
  statement2;
  transition select ...
}
```

Logically, I believe that simply restoring the IR back to have it is the best approach. You could of course do the transformation @apinski-cavium that you suggested, but then the name of `state b` is lost.

mbudiu-vmw:
Often if a compiler isn't smart enough programmers can help by writing the code in a way they know will work.
So I understand the desire to prevent the compiler from obfuscating programmer intentions.
I can see several solutions:
- move the pass to the midend. Then an architecture can just not run it if it is considered detrimental.
- use a flag to skip that particular pass. This can be done with `--excludeFrontendPasses`. 
- add an Ox flag to indicate optimization level. Some passes are skipped when compiling with O0.
- add an annotation to states/parsers saying: "don't ever merge this state"

But I think that @apinski-cavium is fundamentally right: Writing a pass which decomposes a state which is too large into pieces is really the right solution long-term. It will prove beneficial to all users, not just power users.

I generally prefer a good long-term solution, so I would suggest combining the --exclude flag with an effort to develop the state-splitting pass. I think that adding data to the IR for very narrow uses is a trap with long-term maintenance costs.

y-lavoie:
What about splitting up `SimplifyParsers` and running the passes it contains (`RemoveUnreachableStates` and `CollapseChains`) separately? 

- `RemoveUnreachableStates` is something that should stay in the frontend, as it essentially eliminates dead code. I believe such a pass is beneficial to all architectures. 
- As @mbudiu-vmw suggested, `CollapseChains` could be moved to the midend so that an architecture can choose not to run it if it is considered detrimental.

y-lavoie:
As for splitting up parser states that are too large, situations where that would be required are likely to be specific to each architecture or target (hardware limitations, etc.), so passes performing this transformation should be architecture or target specific as well.

mbudiu-vmw:
The easiest way to do it is to add a boolean argument to SimplifyParsers indicating whether collapsing is to be done.
Then in the frontend we call it without this flag.
Do you want to submit a PR for that?
Beware that some reference outputs may change.
Certainly, the splitting pass should be architecture-dependent, it should be guided by knowledge of available resources.

PR #3685, opened 2022-11-10
P4Tools: Implement default action override for BMv2 STF.

### LINKED ISSUE #3557: Create a utils library for the P4Tools test back ends.
There is a lot of redundancy in the test back ends. Create a utility library that all test back ends can use.

fruffy:
This PR implements default action overrides for P4Testgen's BMv2 extension. Only for the STF back end so far. 

If a table does not have a key, it can not match and pick an action. However, it might still be possible to override the default action and choose a different path. This PR implements the logic to override a table's default action.

Also fixes #3557.

### DIFF
--- a/backends/bmv2/bmv2stf.py
+++ b/backends/bmv2/bmv2stf.py
@@ -452,7 +452,7 @@ class RunBMV2(object):
             actionArgs.set(k, v)
         command = "table_set_default " + tableName + " " + actionName
         if actionArgs.size():
-            command += " => " + str(actionArgs)
+            command += " " + str(actionArgs)
         return command
     def parse_table_add(self, cmd):
         tableName, cmd = nextWord(cmd)
--- a/backends/p4tools/testgen/core/small_step/table_stepper.cpp
+++ b/backends/p4tools/testgen/core/small_step/table_stepper.cpp
@@ -99,10 +99,7 @@ bool TableStepper::compareLPMEntries(const IR::Entry* leftIn, const IR::Entry* r
         if (const auto* rightMask = right->checkedTo<IR::Mask>()) {
             rightMaskVal = rightMask->right->checkedTo<IR::Constant>();
         }
-        if ((leftMaskVal->value) < (rightMaskVal->value)) {
-            return false;
-        }
-        return true;
+        return (leftMaskVal->value) >= (rightMaskVal->value);
     }
 
     BUG("Unhandled sort elements type: left: %1% - %2% \n right: %3% - %4% ", left,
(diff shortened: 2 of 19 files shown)


backends/bmv2/bmv2stf.py
```diff
@@ -452,7 +452,7 @@ def parse_table_set_default(self, cmd):
             actionArgs.set(k, v)
         command = "table_set_default " + tableName + " " + actionName
         if actionArgs.size():
-            command += " => " + str(actionArgs)
+            command += " " + str(actionArgs)
```
fruffy, line 455: There was a bug in the STF test framework. Default action overrides were not using the correct syntax.

backends/p4tools/testgen/core/small_step/table_stepper.cpp
```diff
@@ -509,14 +563,18 @@ bool TableStepper::resolveTableKeys() {
...
+    if (entriesAnnotation != nullptr) {
+        isConstant = entriesAnnotation->isConstant;
     }
     // Also check if the table is invisible to the control plane.
     // This also implies that it cannot be modified.
-    return isConstant || table->getAnnotation("hidden") != nullptr;
+    properties.tableIsImmutable = isConstant || table->getAnnotation("hidden") != nullptr;
+    const auto* defaultAction = table->properties->getProperty("default_action");
+    CHECK_NULL(defaultAction);
+    properties.defaultIsImmutable = defaultAction->isConstant;
```
pkotikal, line 577: Nice!

backends/p4tools/testgen/lib/tf.h
```diff
@@ -14,7 +16,7 @@ namespace P4Tools {
 namespace P4Testgen {
 
 /// THe default base class for the various test frameworks (TF). Every test framework has a test
-/// name and a seed associated with it.
+/// name and a seed associated with it. Also contains a variety of common utility functions.
```
pkotikal, line 19: Do all test frameworks support `ActionProfiles` and `ActionSelectors` - do we want to move it to common tf?

backends/p4tools/testgen/targets/bmv2/backend/stf/stf.cpp
```diff
@@ -302,9 +245,14 @@ static std::string getTestCase() {
 ## if control_plane
 ## for table in control_plane.tables
 # Table {{table.table_name}}
+## if existsIn(table, "default_override")
+setdefault {{table.table_name}} {{table.default_override.action_name}}({% for a in table.default_override.act_args %}{{a.param}}:{{a.value}}{% if not loop.is_last %},{% endif %}{% endfor %})
```
pkotikal, line 249: Very nice, now it's very convenient, thank you!

pkotikal (approved):

backends/p4tools/testgen/lib/tf.h
fruffy: I think it is shared between enough back ends that we can add it to common.

PR #3688, opened 2022-11-11
PSA/eBPF: use slice-by-8 algorithm for CRC-32

majchwo:
Previously CRC-32 was calculated using bitwise operations (bit-by-bit). This PR implements more efficient algorithm for CRC-32 using slice-by-8 algorithm. The performance gain depends on the length of an input data. For example, with 40 bytes of input data we observed about 35% reduction of CPU cycles per packet.

This work has been done in cooperation with @osinstom.


mbudiu-vmw (approved):

osinstom:
A few PTF tests are failing, @majchwo and I will fix them soon.

osinstom:
@mbudiu-vmw @majchwo done.

PR #3689, opened 2022-11-11
Unroll finite loops in the programs with infinite ones

VolodymyrPeschanenkoIntel:
These changes allow to unroll all finite loops in the programs even if it contain infinite once. Be fore these changes it leave a program as is for that cases.


backends/p4tools/testgen/targets/bmv2/test/BMV2Xfail.cmake
```diff
@@ -90,6 +90,7 @@ p4tools_add_xfail_reason(
 p4tools_add_xfail_reason(
   "testgen-p4c-bmv2"
   "differs|Expected ([0-9]+) packets on port ([0-9]+) got ([0-9]+)"
+  parser-unroll-test9.p4
```
fruffy, line 93: This does not look right to me. This indicates that the implementation of unrolling is incorrect?

VolodymyrPeschanenkoIntel: I don't think that it is a bug of parser unroll. Test was failed with attached STF file. 
[parser-unroll-test9_8.stf.txt](https://github.com/p4lang/p4c/files/10014139/parser-unroll-test9_8.stf.txt)
It contains the `stateOutOfBound`:
```
    state stateOutOfBound {
        verify(false, error.StackOutOfBounds);
        transition reject;
    }]
```
I think that this packed should be dropped, but generated STF doesn't represent that. So, I think that it's a bug of p4tools. @fruffy, What do you think?

fruffy: https://github.com/p4lang/behavioral-model/blob/d52ac6257bb3a58606383d03b31ed89671504791/docs/simple_switch.md#restrictions-on-parser-code
In BMv2, packets that enter the reject state are not dropped.

fruffy: Is the packet dropped or is there just a value different?

VolodymyrPeschanenkoIntel: > Is the packet dropped or is there just a value different?

The value is different, but I don't think that it is a bug of the `ParserUnroll`. In this example, we have some bug in the loop: `start`, `start_loops`, `finite_loop` (when the index is 0), `finite_loop1` (when the index is 1), `finite_loop2` (when the index is out of bound), `stateOutOfBound`. This path and result after the `ParserUnroll` are correct. 

This problem appears when program has some memory which should be emitted after `OutOfBound`.
To demonstrate that I modified parser-unroll-test2 example with one new `index` variable which should be emitted. Also I increased  the tests numbers to 20 and `p4tool` finished with the same problem (`Packet 0 on port 0 differs` in case for OutOfBound).
[parser-unroll-test2-1.p4.txt](https://github.com/p4lang/p4c/files/10021631/parser-unroll-test2-1.p4.txt)

@fruffy, What do you think?

fruffy: So the issue is that an assignment is removed?

mbudiu-vmw:
I no longer can claim to understand the parser unrolling code, so I am leaning heavily on the testing infrastructure.
Can we make sure that there are automatically generated tests that run using unrolling?

mbudiu-vmw:
One way to treat loops that the compiler cannot unroll is to give a warning or an error and leave the parser unchanged.

VolodymyrPeschanenkoIntel:
@mbudiu-vmw, Do we have something that I need to change in this PR?

mbudiu-vmw:
No, do you want me to merge it?

PR #3691, opened 2022-11-12
PSA/eBPF: Enable passing packets up to the kernel stack

osinstom:
This PR introduces an implicit custom packet path to the PSA model used for eBPF backend. The new packet path enables integrating PSA pipelines for eBPF with the standard Linux kernel network stack. The main use case is to delegate ARP/ICMP handling to the kernel, instead of re-inventing the wheel by implementing own ARP/ICMP handlers in userspace. The new packet path also enables sending packets to the userspace application listening on a socket. 

To send packets to the kernel, a P4 programmer should set `drop` to `false` and leave `egress_port` unspecified (or set to `0`) in the ingress pipeline.

Note that reply packets (ICMP Reply, ARP Response, etc.) traverse the egress pipeline, but the PSA Egress pipeline is unable to distinguish between packets from PSA Ingress and from kernel. Therefore, we disallow handling packets from kernel in the egress pipeline. Such packets are ignored by the egress eBPF program and directly send out to the output port.


backends/ebpf/psa/ebpfPipeline.cpp
```diff
@@ -568,6 +583,28 @@ void TCIngressPipeline::emitTrafficManager(CodeBuilder *builder) {
                           control->outputStandardMetadata->name.name);
     builder->newline();
 
+    builder->appendFormat("if (%s.drop == false && %s.egress_port == 0) ",
```
mbudiu-vmw, line 586: usually you don't write 'x==false' but '!x'

backends/ebpf/psa/README.md
```diff
@@ -37,6 +37,20 @@ The TC-based design of PSA for eBPF is depicted in Figure below.
...
 ## Packet paths
 
+### NTK (Normal Packet To Kernel) 
+
+**WARNING!** The NTK packet path is a custom packet path used for the PSA-eBPF only! It is not a standardized PSA packet path. 
+
+The NTK packet path allows integrating P4/PSA programs for eBPF with the standard Linux kernel stack. The main use case is handling 
+ICMP/ARP requests and sending packet to the userspace process listening on a socket.
+
+The NTK path is enforced if `drop` is set to `false` and `egress_port` is left unspecified or set to 0 (it's a special implicit port number that forwards packets to the kernel stack).
```
mbudiu-vmw, line 47: I would not say "unspecified", but "unchanged". Since it is an inout value, it must be initialized properly.

mbudiu-vmw (approved):

osinstom:
@mbudiu-vmw I addressed your comments, thanks! Can we merge this?

PR #3692, opened 2022-11-13
Fix name shortening

kamleshbhalui:
- Fix JumpAction operands name
- Remove suffixes from NoAction, only when originalname is NoAction
- Fix Action param names

### DIFF
--- a/testdata/p4_16_dpdk_errors_outputs/pna-dpdk-direct-meter-err-1.p4-error
+++ b/testdata/p4_16_dpdk_errors_outputs/pna-dpdk-direct-meter-err-1.p4-error
@@ -1,6 +1,6 @@
-[--Werror=unexpected] error: execute method for different DirectMeter instances (meter0_0 and meter1_0) called within same action
-[--Werror=unexpected] error: execute method for different DirectMeter instances (meter0_0 and meter1_0) called within same action
+pna-dpdk-direct-meter-err-1.p4(99): [--Werror=expected] error: Expected default action default_route_drop to have 'dpdk_execute' method call for DirectMeter extern instance MainControlImpl.meter0
+    action default_route_drop() {
+           ^^^^^^^^^^^^^^^^^^
 pna-dpdk-direct-meter-err-1.p4(116): [--Werror=target-error] error: Direct counters and direct meters are unsupported for wildcard match table ipv4_da_lpm
     table ipv4_da_lpm {
           ^^^^^^^^^^^
-[--Werror=unexpected] error: execute method for different DirectMeter instances (meter0_0 and meter1_0) called within same action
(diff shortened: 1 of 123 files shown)


testdata/p4_16_samples_outputs/psa-dpdk-table-key-consolidation-mixed-keys-6.p4.spec
```diff
@@ -69,6 +69,7 @@ header ethernet instanceof ethernet_t
 header ipv4 instanceof ipv4_t
 header tcp instanceof tcp_t
 
+;oldname:NoAction_1
```
jfingerh, line 72: What happens if a P4 developer writes an action like this in their code, which is perfectly legal, although perhaps unusual:
```
action NoAction_5 () {
     // code here that does something that is NOT a no-op
}
```

kamleshbhalui: thanks, fixed it now, it will be preserved now.

testdata/p4_16_dpdk_errors_outputs/pna-dpdk-direct-meter-err-1.p4-error
```diff
@@ -1,6 +1,6 @@
-[--Werror=unexpected] error: execute method for different DirectMeter instances (meter0_0 and meter1_0) called within same action
-[--Werror=unexpected] error: execute method for different DirectMeter instances (meter0_0 and meter1_0) called within same action
```
jfingerh, line 2: Is it correct that these error messages are no longer present?  Were they only present before because of some bug that you fixed with this PR?

testdata/p4_16_pna_errors_outputs/pna-dpdk-direct-counter-err-3.p4-error
```diff
@@ -1,3 +1,3 @@
-pna.p4(385): [--Werror=unexpected] error: count method of per_prefix_pkt_bytes_count extern can only be invoked from within action of ownertable
-  void count(@optional in bit<32> increment);
```
jfingerh, line 2: Same question for the error message above.

testdata/p4_16_dpdk_errors_outputs/pna-dpdk-direct-meter-err-1.p4-error
usha1830: Yes. The PR does not seem to change anything that would trigger this error output change.

usha1830 (approved):

jafingerhut:
Merging since Usha and I have both approved, and changes limited to DPDK back end.

jafingerhut (approved):

PR #3698, opened 2022-11-15
Migrate away from deprecated Bazel option

smolkaj:
Migrate from cfg = "host" to cfg = "exec".
Host is being deprecated and replaced by exec, see details here: https://docs.bazel.build/versions/4.1.0/skylark/rules.html#configurations


smolkaj:
@kheradmandG could you do a sanity check please?

mbudiu-vmw (approved):

smolkaj:
Thanks @mbudiu-vmw, I think we can merge this.

PR #3701, opened 2022-11-16
Clean up the includes in the IR folder

fruffy:
Major rework of how the ir libraries are structured. WiP. Depends on #3679 and #3678.


frontends/p4/symbol_table.cpp
```diff
@@ -176,23 +176,23 @@ void ProgramStructure::pop() {
 void ProgramStructure::declareType(IR::ID id) {
     if (debug) fprintf(debugStream, "ProgramStructure: adding type %s\n", id.name.c_str());
 
-    LOG3("ProgramStructure: adding type " << id);
+    LOG3("ProgramStructure: adding type " << id.toString());
```
fruffy, line 179: I had to add a `toString` method here. I do not know how this compiled before. I do not think `ID` had the `<<` operator implemented.

frontends/parsers/v1/v1parser.ypp
```diff
@@ -73,8 +75,19 @@ inline std::ostream& operator<<(std::ostream& out, const V1::HeaderType& headerT
 }
 
 inline std::ostream& operator<<(std::ostream& out, const V1::Attributes& a) {
-    return out << "Attributes(" << (a.signed_ ? "signed," : "") <<
-                  (a.saturating ? "saturating," : "") << a.annotations << ')'; }
+    out << "Attributes(" << (a.signed_ ? "signed," : "") << (a.saturating ? "saturating," : "");
```
fruffy, line 78: Similarly, `<< a.annotations` was never implemented, so I print these values manually.

ir/CMakeLists.txt
```diff
@@ -66,3 +66,7 @@ set (IR_DEF_FILES ${IR_DEF_FILES} ${BASE_IR_DEF_FILES} PARENT_SCOPE)
 build_unified(IR_SRCS)
 add_library (ir STATIC ${IR_SRCS})
 add_dependencies(ir genIR)
+
+if(ENABLE_IWYU)
```
fruffy, line 70: Only adding IWYU for the `ir` folder for now.

ir/ir.h
```diff
@@ -17,41 +17,8 @@ limitations under the License.
 #ifndef _IR_IR_H_
 #define _IR_IR_H_
 
-#include <strings.h>
```
fruffy, line 20: Most of the IR includes have been removed and only `ir-inline.h` and `ir-generated.h` remain. Hopefully, this reduces the amount of megaincludes in the compiler.

ir/v1.def
```diff
@@ -337,7 +337,18 @@ class NameList {
     safe_vector<ID>  names = {};
     NameList(Util::SourceInfo si, cstring n) { names.emplace_back(si, n); }
     NameList(Util::SourceInfo si, ID n) { names.emplace_back(si, n); }
-    dump_fields { out << "names=" << names; }
+    dump_fields {
```
fruffy, line 340: Another def file where I had to add manual output printing for a vector.

lib/source_file.h
```diff
@@ -311,6 +315,11 @@ class InputSources final {
     std::vector<Comment*> comments;
 };
 
+inline std::ostream& operator<<(std::ostream& out, const SourceInfo& obj) {
```
fruffy, line 318: Add a friend printing function here.

testdata/p4_16_samples_outputs/parser-inline/parser-inline-test1-frontend.p4
```diff
@@ -70,7 +70,29 @@ parser ParserImpl(packet_in packet, out headers hdr, inout metadata meta, inout
         hdr.h2.setInvalid();
         hdr.h3.setInvalid();
         hdr.h4.setInvalid();
-        transition Subparser_start;
+        transition Subparser_start_0;
```
fruffy, line 73: I do not understand why this happens, but this seems like correct behavior to me? The parser is actually inlined. But this is something we should properly check.

frontends/p4/symbol_table.cpp
mbudiu-vmw: We use IDs in many places in LOG calls.
There must be something else going on.

mbudiu-vmw:
You should not mix reformatting with this PR, because it makes it very difficult to review.
The ir.h including everything was by design.
You should talk to @ChrisDodd about this change.

frontends/p4/optimizeExpressions.h
```diff
@@ -0,0 +1,24 @@
+#ifndef FRONTENDS_P4_OPTIMIZEEXPRESSIONS_H_
+#define FRONTENDS_P4_OPTIMIZEEXPRESSIONS_H_
+
+#include "frontends/common/constantFolding.h"
+#include "frontends/p4/strengthReduction.h"
+
+namespace IR {
```
vhavel, line 7: Headers in frontends/p4 usually put names into `namespace P4`, should we do the same here?

vhavel (approved):
Thanks for this, I agree it's better in long-run to maintain proper includes instead of having one "super-header". It's surprising that the amount of includes this PR needs to add is relatively low - headers are likely pulled in anyway - on the other hand it's great backends won't need many changes after this PR.

fruffy:
Yeah, most of the back ends and the mid end just use IR nodes and visitors. The [generated IR file](https://github.com/p4lang/p4c/pull/3701/files#diff-792e3e275e19235316be8e986bc73d77d24a5219710bcbd990febfe13fe5a480R124) also still pulls in a lot of includes. I think this can also be improved, but at least the file is self-contained now.

PR #3707, opened 2022-11-19
Equal types are always equivalent

### LINKED ISSUE #3699: Compiler bug with header stacks and constant folding and saturation add
I thought this was working previously but it is no longer working and also gives a compiler bug.
I was using header stacks to get a static assert before I had implemented static assert in my front-end (similar trick was used in C and C++ with arrays).

Full testcase:
```
const bit<4> one = 1;
const bit<4> max = 0xf;
const bit<4> value = max |+| one;
const bit<4> value1 = max |+| 2;

header h{}


// Pre static_assert static assert until static_assert support is added
void f(in h[value == max ? 1 : -1] a){}
void f1(in h[(max |+| max) == max ? 1 : -1] a){}
void f1(in h[(max |+| 0) == max ? 1 : -1] a){}
void f1(in h[value1 == max ? 1 : -1] a){}


const int<5> ones = 1;
const int<5> maxs = 0xf;
const int<5> values = maxs |+| ones;
const int<5> values1 = maxs |+| 2;
void f(in h[values == maxs ? 1 : -1] a){}
void f1(in h[(maxs |+| maxs) == maxs ? 1 : -1] a){}
void f1(in h[(maxs |+| 0) == maxs ? 1 : -1] a){}
void f1(in h[values1 == maxs ? 1 : -1] a){}
// { dg-warning "main" "" { target *-*-* } 0 }
```

The error message that P4C gives is:
```
testsuite/p4c.language/valid/sat_add.p4(10): [--Werror=type-error] error: Width must be specified for at least one of 1 or -1
void f(in h[value == max ? 1 : -1] a){}
                           ^
testsuite/p4c.language/valid/sat_add.p4(10)
void f(in h[value == max ? 1 : -1] a){}
                                ^
testsuite/p4c.language/valid/sat_add.p4(10): [--Werror=type-error] error: header h[?]: Size of header stack type should be a constant
void f(in h[value == max ? 1 : -1] a){}
          ^^^^^^^^^^^^^^^^^^^^^^^^
In file: /home/apinski/src/p4/p4c/frontends/p4/typeMap.cpp:115
Compiler Bug: Changing type of <Parameter>(1502) a/36 in type map from <Type_Stack>(1491) to <Type_Stack>(1491)
```
The constant folding is part of the issue but the compiler bug is definitely a bug.

mbudiu-vmw:
Fixes #3699

### DIFF
--- a/frontends/p4/typeMap.cpp
+++ b/frontends/p4/typeMap.cpp
@@ -142,6 +142,7 @@ void TypeMap::addSubstitutions(const TypeVariableSubstitution* tvs) {
 bool TypeMap::equivalent(const IR::Type* left, const IR::Type* right, bool strict) const {
     if (!strict) strict = strictStruct;
     LOG3("Checking equivalence of " << left << " and " << right);
+    if (left == right) return true;
     if (left == nullptr) return right == nullptr;
     if (right == nullptr) return false;
     if (left->node_type_name() != right->node_type_name()) return false;
--- a/testdata/p4_16_errors/issue3699.p4
+++ b/testdata/p4_16_errors/issue3699.p4
@@ -1,0 +1,23 @@
+const bit<4> one = 1;
+const bit<4> max = 0xf;
+const bit<4> value = max |+| one;
+const bit<4> value1 = max |+| 2;
+
+header h{}
+
+
+// Pre static_assert static assert until static_assert support is added
+void f(in h[value == max ? 1 : -1] a){}
+void f1(in h[(max |+| max) == max ? 1 : -1] a){}
+void f1(in h[(max |+| 0) == max ? 1 : -1] a){}
+void f1(in h[value1 == max ? 1 : -1] a){}
+
+
+const int<5> ones = 1;
+const int<5> maxs = 0xf;
+const int<5> values = maxs |+| ones;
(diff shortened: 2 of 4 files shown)


apinski-cavium:
Hmm, isn't the code in the testcase valid? Or is it just the way constant folding interacts with other passes that causes p4c to reject it too? At one point in the last 6 months, I remembered the testcase working fully too.

jfingerh:
If the goal is simply to eliminate the CompilerBug, then it looks like this PR should do that.

@apinski-cavium Are you hoping for a change to the compiler that makes the added test program have no compile-time errors?

mbudiu-vmw:
If the expression using the mux was used to initialize a const, then this could be a bug.
As it is, I think it's an acceptable behavior of the compiler, although it's not the best user experience really.

mbudiu-vmw:
@apinski-cavium what do you think, is this a legal behavior?

mbudiu-vmw:
I think we should merge this anyway, and we can debate about the evaluation order later.

jnfoster (approved):
It would be insane for this PR not to be correct. 

Now, there may be bugs elsewhere in the compiler, but I agree we should merge this now and deal with any other bugs later.

Issue #3705, opened 2022-11-19
clang-format is not listed as a dependency


mbudiu-vmw:
So far clang was one of the possible supported compilers, but was not required.
When the clang format check fails there are no indications given to the user how to remedy the problem.

mbudiu-vmw:
clang-format has the same problem as cpplint: it does not run on the extension directories and has no configurable exclusions list.

mbudiu-vmw:
On my ubuntu installation apt-get installs clang-format 6.0. This program gives the following errors when invoked:
```
clang-format: Unknown command line argument '--Werror'.  Try: 'clang-format -help'
clang-format: Did you mean '-cursor'?
clang-format: Unknown command line argument '--dry-run'.  Try: 'clang-format -help'
clang-format: Did you mean '-icp-lto'?
```

mbudiu-vmw:
Ideally clang-format should be run as a pre-commit hook so that files that will be rejected cannot even be pushed. Otherwise we'll generate lots of failed jobs needlessly.

fruffy:
clang-format can be installed using `pip3 install --user --upgrade clang-format`. This should install the right and latest version. 

Regarding the commit hook, unfortunately these have to be installed manually by each person. You can use the `tools/install_git_hooks.sh`. I will update the README instructions.

mbudiu-vmw:
```
$ pip3 install "clang-format>=15.0.4"
Collecting clang-format>=15.0.4
  Downloading https://files.pythonhosted.org/packages/31/3d/83a35768932c398476595509812aae58b352a7052909dab1e2a941fb89f2/clang-format-15.0.4.tar.gz
    Complete output from command python setup.py egg_info:
    Traceback (most recent call last):
      File "<string>", line 1, in <module>
      File "/tmp/pip-build-y7wag41v/clang-format/setup.py", line 3, in <module>
        from skbuild import setup
    ModuleNotFoundError: No module named 'skbuild'
```

clang-format is written in Python?

fruffy:
The pip wheel version should be a wrapper around a precompiled clang-format binary, which is periodically updated. It is more convenient than having to rely on a specific Ubuntu package. 

I am not quite sure why this installation does not work in your case. I guess `pip3 install --upgrade pip` can help?

usha1830:
> ```
> $ pip3 install "clang-format>=15.0.4"
> Collecting clang-format>=15.0.4
[...]

PR #3709, opened 2022-11-20
Linter usability improvements.

### LINKED ISSUE #3704: cpplint list of files is not configurable
Currently run-cpplint.sh does not follow symlinks, so it does not check extensions.
This can be solved by adding the -L flag to the find command.
However, a more important problem is that the list of exclusions is global in this file, so people who add a new extension cannot add exclusions. This makes it impossible to run cpplint using this tool.
The previous way of doing it using CMakeList was more modular.

fruffy:
Fixes #3704 and #3705.


fruffy:
@mbudiu-vmw Making the linters part of CI is a little tricky because you have to run CMake. CMake requires a series of dependencies that need to be installed first. So the CI linter check will not be as light-weight anymore.

kamleshbhalui:
I think we also have to strictly say which version of clang-format CI uses, and everyone has to do the formatting with the same version, as one version can break the formatting done by another version.

fruffy:
> I think we also have to strictly say which version of clang-format CI uses, and everyone has to do the formatting with the same version, as one version can break the formatting done by another version.

I do prefer always running the latest clang-format on CI because upgrading with pip3 is easy and the project moves fairly quickly. However, we can also fix it at the current version, which is `15.0.4`.

CMakeLists.txt
```diff
@@ -464,19 +489,40 @@ if(DEFINED CPPLINT_FILES)
...
+
+find_program(CLANG_FORMAT_CMD clang-format)
+
+if(NOT ${CLANG_FORMAT_CMD})
+  if(DEFINED CLANG_FORMAT_FILES)
+    list(SORT CLANG_FORMAT_FILES)
+    set(CLANG_FORMAT_CMD clang-format)
+    add_custom_target(
+      clang-format
+      COMMAND ${CLANG_FORMAT_CMD} --Werror --dry-run -i ${CLANG_FORMAT_FILES}
```
mbudiu-vmw, line 508: my version of clang-format does not understand these flags

CMakeLists.txt
```diff
@@ -451,7 +451,32 @@ endif ()
...
+  frontends/*.h
+  extensions/*.cpp
+  extensions/*.h
+  ir/*.cpp
+  ir/*.h
+  tools/*.cpp
+  tools/*.h
+)
+# Filter some folders from the CI checks.
+list(FILTER P4C_LINT_LIST EXCLUDE REGEX "backends/p4tools/submodules")
```
mbudiu-vmw, line 472: Is this how extensions should declare exclusions too?
Maybe you can add a comment if that's correct.

mbudiu-vmw (approved):

CMakeLists.txt
fruffy: I would install the latest version (15.0.4) with `pip` and not with `apt`. I do not quite know how to bridge this otherwise.

CMakeLists.txt
fruffy: This is done specifically for the compiler files. Extensions are free to add files how they see fit.

PR #3713, opened 2022-11-21
Dags needs to be revisited when renaming metadata

### LINKED ISSUE #3620: BMv2 back end: Duplicate objects of type 'header' with name 'h'
Triggered by #3605.

`gauntlet_action_mux-bmv2.p4` fails because of an issue in BMv2 compiler back end (?). Running an STF test produces the output: 

```
Duplicate objects of type 'header' with name 'h' 
```
It looks like the compiler produces an incorrect JSON: 

```json
"headers" : [
  {
    "name" : "scalars",
    "id" : 0,
    "header_type" : "scalars_0",
    "metadata" : true,
    "pi_omit" : true
  },
  {
    "name" : "standard_metadata",
    "id" : 1,
    "header_type" : "standard_metadata",
    "metadata" : true,
    "pi_omit" : true
  },
  {
    "name" : "eth_hdr",
    "id" : 2,
    "header_type" : "ethernet_t",
    "metadata" : false,
    "pi_omit" : true
  },
  {
    "name" : "h",
    "id" : 3,
    "header_type" : "H",
    "metadata" : false,
    "pi_omit" : true
  },
  {
    "name" : "h",
    "id" : 4,
    "header_type" : "H",
    "metadata" : false,
    "pi_omit" : true
  }
],
```

mbudiu-vmw:
Fixes #3620 
We actually don't have a test to reproduce the bug at this point, but this change cannot be wrong...

### DIFF
--- a/backends/bmv2/common/backend.h
+++ b/backends/bmv2/common/backend.h
@@ -140,6 +140,7 @@ class RenameUserMetadata : public Transform {
         : refMap(refMap), userMetaType(userMetaType), namePrefix(namePrefix) {
         setName("RenameUserMetadata");
         CHECK_NULL(refMap);
+        visitDagOnce = false;
     }
 
     const IR::Node* postorder(IR::Type_Struct* type) override {


jfingerh:
I can rubber-stamp this with a yes, but there is about zero chance of me learning enough about the compiler internals to actually review it the way you wish someone would.

jafingerhut (approved):
I approve that Mihai believes this is not wrong :-)

PR #3715, opened 2022-11-21
Off by one warning error in constant-folding for large shifts

### LINKED ISSUE #3289: warning for unsigned fixed width shift left is twice
Take:
```
bit<5> f(in int<3> a)
{
  return (5w1) << 6;
}
```
p4c produces:
```
shiftoverflow.p4(3): [--Wwarn=overflow] warning: <<: Shifting 5-bit value with 6
  return (5w1) << 6;
         ^^^^^^^^^^
shiftoverflow.p4(3): [--Wwarn=mismatch] warning: 5w64: value does not fit in 5 bits
  return (5w1) << 6;
         ^^^^^^^^^^
```

If we change 6 to 5, the first warning is gone but maybe it should not be; I think there is an off by one error.

The second warning is another interesting one as I thought shifts have a well defined behavior always. How can there be a mismatch?
Also the specification only talks about overflow for the signed case and no overflow for the unsigned case so both of these warnings should not be there.

mbudiu-vmw:
Fixes #3289

### DIFF
--- a/frontends/common/constantFolding.cpp
+++ b/frontends/common/constantFolding.cpp
@@ -306,7 +306,8 @@ const IR::Node* DoConstantFolding::postorder(IR::Add* e) {
 }
 
 const IR::Node* DoConstantFolding::postorder(IR::AddSat* e) {
-    return binary(e, [](big_int a, big_int b) -> big_int { return a + b; }, true);
+    return binary(
+        e, [](big_int a, big_int b) -> big_int { return a + b; }, true);
 }
 
 const IR::Node* DoConstantFolding::postorder(IR::Sub* e) {
@@ -314,7 +315,8 @@ const IR::Node* DoConstantFolding::postorder(IR::Sub* e) {
 }
 
 const IR::Node* DoConstantFolding::postorder(IR::SubSat* e) {
-    return binary(e, [](big_int a, big_int b) -> big_int { return a - b; }, true);
+    return binary(
+        e, [](big_int a, big_int b) -> big_int { return a - b; }, true);
 }
 
 const IR::Node* DoConstantFolding::postorder(IR::Mul* e) {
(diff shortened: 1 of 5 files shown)


frontends/common/constantFolding.cpp
```diff
@@ -761,7 +761,7 @@ const IR::Node* DoConstantFolding::shift(const IR::Operation_Binary* e) {
 
     auto tb = left->type->to<IR::Type_Bits>();
     if (tb != nullptr) {
-        if (((unsigned)tb->width_bits() < shift) && warnings)
+        if (((unsigned)tb->width_bits() <= shift) && warnings)
             ::warning(ErrorType::WARN_OVERFLOW, "%1%: Shifting %2%-bit value with %3%", e,
```
rst0git, line 765: From user's perspective, I'm not sure if printing the value here is useful. For example, the following code snippet
```P4
bit<5> f(in bit<16> a)
{
  return (5w1) << 2048;
}
```
would have warning:
> shiftoverflow.p4(3): [--Wwarn=mismatch] warning: 5w32317006071311007300714876688669951960444102669715484032130345427524655138867890893197201411522913463688717960921898019494119559150490921095088152386448283120630877367300996091750197750389652106796057638384067568276792218642619756161838094338476170470581645852036305042887575891541065808607552399123930385521914333389668342420684974786564569494856176035326322058077805659331026192708460314150258592864177116725943603718461857357598351152301645904403697613233287231227125684710820209725157101726931323469678542580656697935045997268352998638215525166389437335543602135433229604645318478604952148193555853611059596230656: value does not fit in 5 bits

Perhaps we could use more a succinct warning like `Left shift count >= width of type`?

mbudiu-vmw: The error message you are quoting comes from a different place in the code, which checks for overflow in assignments (or in general, casting a large value to a narrower bitwidth).
If you think that the message can be improved you should probably file a new issue.

PR #3720, opened 2022-11-22
Fix bug caused by uninitialized variable in typechecker

### LINKED ISSUE #3719: Compiler Bug: nested action list?
After upgrading from p4c as of Nov 1
```
4a74084c675a3e8bdf7a318e468cbd5bb27c6914 Do not constant fold mux before it has been typechecked (... by Mihai Budiu <[mbudiu@vmware.com](mailto:mbudiu@vmware.com)>
```
to p4c as of Nov 22
```
- aa0a24caece8a3c30ac69076e08f74ec35d81007 P4Tools: Unify packet sizing options. (#3695) by fruffy <[5960321+fruffy@users.noreply.github.com](mailto:5960321+fruffy@users.noreply.github.com)>
```
I'm seeing the following regression in one of our programs:
```
In file: third_party/p4lang_p4c/frontends/p4/typeChecking/typeChecker.cpp:3902
Compiler Bug: { @proto_id acl_copy();
  @proto_id acl_trap();
  @proto_id acl_forward();
  @proto_id acl_mirror();
  @proto_id acl_drop(standard_metadata);
  @proto_id acl_experimental_trap();
  @defaultonly NoAction(); }: nested action list?
```

mbudiu-vmw:
Fixes #3719 
Also fixed a couple of warnings in gc.cpp
There is no simple test, but hopefully @smolkaj can confirm that the bug is fixed.

### DIFF
--- a/frontends/p4/typeChecking/typeChecker.cpp
+++ b/frontends/p4/typeChecking/typeChecker.cpp
@@ -134,7 +134,8 @@ TypeInference::TypeInference(ReferenceMap* refMap, TypeMap* typeMap, bool readOn
       typeMap(typeMap),
       initialNode(nullptr),
       readOnly(readOnly),
-      checkArrays(checkArrays) {
+      checkArrays(checkArrays),
+      currentActionList(nullptr) {
     CHECK_NULL(typeMap);
     CHECK_NULL(refMap);
     visitDagOnce = false;  // the done() method will take care of this
--- a/lib/gc.cpp
+++ b/lib/gc.cpp
@@ -37,6 +37,7 @@ limitations under the License.
 #define _GLIBCXX_USE_NOEXCEPT _NOEXCEPT
 #endif
 
+#if HAVE_LIBGC
 static bool done_init, started_init;
 // emergency pool to allow a few extra allocations after a bad_alloc is thrown so we
 // can generate reasonable errors, a stack trace, etc
@@ -44,7 +45,6 @@ static char emergency_pool[16 * 1024];
 static char* emergency_ptr;
 
 // One can disable the GC, e.g., to run under Valgrind, by editing config.h
-#if HAVE_LIBGC
 void* operator new(std::size_t size) {
     /* DANGER -- on OSX, can't safely call the garbage collector allocation
      * routines from a static global constructor without manually initializing


smolkaj:
I tried manually and can confirm that it fixes the issue.

Thanks so much Mihai for the incredibly fast diagnosis & fix!

smolkaj:
Any ideas why https://github.com/p4lang/p4c/blob/main/testdata/p4_16_samples/pins/pins_middleblock.p4 didn't catch this, @mbudiu-vmw ?

jafingerhut:
@smolkaj I do not know why your tests were failing, but when I built with latest p4c source code, _without_ the changes in this PR, using the following commands on an Ubuntu 20.04 system:
```
cd p4c
mkdir build
cd build
cmake .. -DCMAKE_BUILD_TYPE=DEBUG
make -j2
```
I got no errors or crashes at all when compiling the version of pins_middleblock.p4 that is in the p4c repo today, nor when attempting to compile the version of middleblock.p4 in this repo: https://github.com/p4lang/p4c/blob/main/testdata/p4_16_samples/pins/pins_middleblock.p4

I don't know why you did, but we could have been using different versions of OS, glibc, C++ compiler, etc. as I mentioned above.  I can give you the versions I ran with, if it is of interest to you, but I don't know how else to explain it than whatever is different about our build and/or run-time environments.

jafingerhut (approved):
I can approve of this if you want, Mihai, but I don't know the code, so not sure why you would ask me for review unless it is that I respond relatively quickly :-)

mbudiu-vmw:
Uninitialized variables are not deterministic. If it happens to be 0 all works well.

smolkaj:
Makes sense (well, sort of, but this is C++ after all) :) 

Andy, on the other thread we discussed that both Mihai and I were able to reproduce with Bazel, but not with cmake.
If we had latent undefined behavior here I suppose that's not surprising.

By the way, Mihai, props for getting your hands dirty with Bazel 👍 Thanks a lot for fixing this, once again!

rst0git (approved):
LGTM

cppcheck is also able to detect this problem:
```console
$ cppcheck frontends/p4/typeChecking/typeChecker.cpp --inconclusive --enable=warning
Checking frontends/p4/typeChecking/typeChecker.cpp ...
frontends/p4/typeChecking/typeChecker.cpp:131:16: warning: Member variable 'TypeInference::currentActionList' is not initialized in the constructor. [uninitMemberVar]
TypeInference::TypeInference(ReferenceMap* refMap, TypeMap* typeMap, bool readOnly,
               ^
```

mbudiu-vmw:
@fruffy this fails with the following report: `testgen-p4c-bmv2/issue3702-bmv2.p4`
Is this a true positive?
Even if it is a true positive, I assume it is not really related to this PR.
If it is a true positive, can you please file a reproduction for it?
I will merge this PR anyway, since it fixes a real bug.

fruffy:
This is an issue that has already been fixed.

davidbolvansky:
> Uninitialized variables are not deterministic. If it happens to be 0 all works well.

If think we could enhance our Clang CI build with additional flag -ftrivial-auto-var-init=pattern - I can try.
https://clang.llvm.org/docs/ClangCommandLineReference.html

See it in action:
https://godbolt.org/z/fPMdf3Gsx

Real bugs would be more exposed now.

apinski-cavium:
>If think we could enhance our Clang CI build with additional flag -ftrivial-auto-var-init=pattern - I can try.
If you are using GCC 12 or above, you can use the same option too.

davidbolvansky:
> > If think we could enhance our Clang CI build with additional flag -ftrivial-auto-var-init=pattern - I can try.
> 
> If you are using GCC 12 or above, you can use the same option too.

Yeah :)

PR #3733, opened 2022-11-24
Remove unused files in p4_16_samples_outputs directory

jafingerhut:
Several test programs that used to be in the p4_16_samples directory were moved to other directories such as p4_16_dpdk_errors, and thus their new expected output files are in other directories such as p4_16_dpdk_errors_outputs.


jafingerhut:
@usha1830 For your review.  I think it might have been you that moved many of these test programs from p4_16_samples to a different directory, or maybe someone else working on the DPDK back end.

mbudiu-vmw:
If no test fails this must be good

mbudiu-vmw (approved):

usha1830:
Thanks for this cleanup, Andy!!

PR #3737, opened 2022-11-28
 Correction of PTF tests for BMV2

OleksandrLetychevskyiIntel:
Attempt to fix generation of PTF tests for BMV2.


backends/p4tools/modules/testgen/targets/bmv2/test/TestTemplate.cmake
```diff
@@ -18,6 +18,51 @@ macro(check_with_bmv2 testfile testfolder p4test)
   file(APPEND ${testfile} "done\n")
 endmacro(check_with_bmv2)
 
+
```
fruffy-bfn, line 21: Can you make this a shell script? It will be easier to debug and read. You could also make this a Python script.

fruffy:
@antoninbas  @OleksandrLetychevskyiIntel is trying to get PTF tests with simple switch grpc to run on our CI, but for some reason the runner can not find `google.rpc`. Do you know an easy fix to this? Is there a repository that is actually running PTF tests in CI?

antoninbas:
> @antoninbas
> 
> @OleksandrLetychevskyiIntel is trying to get PTF tests with simple switch grpc to run on our CI, but for some reason the runner can not find `google.rpc`. Do you know an easy fix to this? Is there a repository that is actually running PTF tests in CI?

No, I don't think any p4lang repo runs P4Runtime PTF tests in CI.
I remember running into a similar issue for https://github.com/p4lang/p4runtime-shell/tree/main recently, and I had to make some changes to the installation scripts there.

tools/ci-build.sh
```diff
@@ -80,7 +90,8 @@ function install_ptf_ebpf_test_deps() (
   export P4C_PTF_PACKAGES="gcc-multilib \
                            python3-six \
                            libgmp-dev \
-                           libjansson-dev"
+                           libjansson-dev\
+                           protobuf-compiler"
```
fruffy, line 94: This installation will not work. These are only for the EBPF-PTF version of the compiler, which is not enabled here.

fruffy:
I found a way to build this. https://github.com/fruffy/p4c/blob/2d151e411d5a9c84edbe119fabb33899281e1527/tools/ci-build.sh

Please try to use this file configuration. 





> No, I don't think any p4lang repo runs P4Runtime PTF tests in CI. I remember running into a similar issue for https://github.com/p4lang/p4runtime-shell/tree/main recently, and I had to make some changes to the installation scripts there.

I had to install `google-api-python-client` then the error was resolved. I think it might make sense to simplify the third-party installation. It seems unnecessarily complex, partly because the workarounds were intended for Ubuntu 16.04.

tools/ci-build.sh
```diff
@@ -124,9 +124,17 @@ if [[ "${DISTRIB_RELEASE}" == "18.04" ]] ; then
 fi
 ccache --set-config max_size=1G
 
+# we want to use Python as the default so change the symlinks
```
fruffy-bfn, line 127: Remove these commands. You should not need to add them.

backends/p4tools/modules/testgen/targets/bmv2/program_info.cpp
```diff
@@ -129,6 +131,7 @@ std::vector<Continuation::Command> BMv2_V1ModelProgramInfo::processDeclaration(
         auto* egressPortVar =
             new IR::Member(IR::getBitType(TestgenTarget::getPortNumWidth_bits()),
                            new IR::PathExpression("*standard_metadata"), "egress_port");
+        cmds.emplace_back(Continuation::Guard(getPortConstraint(getTargetOutputPortVar())));
```
fruffy, line 134: Do not use a guard here. Just a normal comparison for the ingress port variable. It needs to be smaller than 7.

backends/p4tools/modules/testgen/targets/bmv2/program_info.cpp
```diff
@@ -159,6 +162,12 @@ const IR::Member* BMv2_V1ModelProgramInfo::getTargetInputPortVar() const {
     return new IR::Member(IR::getBitType(TestgenTarget::getPortNumWidth_bits()),
                           new IR::PathExpression("*standard_metadata"), "ingress_port");
 }
+const IR::Expression* BMv2_V1ModelProgramInfo::getPortConstraint(const IR::Member* portVar) const {
+    const IR::Operation_Binary* portConstraint =
+        new IR::LAnd(new IR::Geq(portVar, new IR::Constant(portVar->type, 0)),
```
fruffy, line 167: You should not need the Geq since these values can not be less than 0.

backends/p4tools/modules/testgen/targets/bmv2/program_info.cpp
```diff
@@ -57,6 +57,7 @@ BMv2_V1ModelProgramInfo::BMv2_V1ModelProgramInfo(
     /// the deparser. This sequence also includes nodes that handle transitions between the
     /// individual component instantiations.
     int pipeIdx = 0;
+    pipelineSequence.emplace_back(Continuation::Guard(getPortConstraint(getTargetInputPortVar())));
```
fruffy, line 60: Do not use a guard here. Just a normal comparison for the ingress port variable. It needs to be smaller than 7.  Add this to the target constraints in the program info.

tools/ci-build.sh
```diff
@@ -103,6 +103,7 @@ P4C_PIP_PACKAGES="ipaddr \
                   clang-format>=15.0.4"
 
 
+
```
fruffy, line 106: This change is unnecessary.

backends/p4tools/modules/testgen/targets/bmv2/test/TestTemplate.cmake
```diff
@@ -18,6 +18,24 @@ macro(check_with_bmv2 testfile testfolder p4test)
...
+# Write the script to check BMv2 STF tests to the designated test file.
```
fruffy, line 22: Please update these comments.

fruffy (changes requested):
Small change requests.

fruffy:
@mbudiu-vmw @jafingerhut 

This is almost ready, the only thing that is left is to update Xfails. 

This PR adds the capability to run PTF tests for BMv2 in the P4C CI. This PR adds a runner, a virtual test environment, and also an installation procedure to get everything for the dependencies. P4Testgen generates PTF tests for all the BMV2 programs now. 

There is some refactoring that should be done for the Python scripts (ebpf_env, testutils, run-bmv2-test, run-bmv2-ptf-test, etc).  There should also be documentation and references to the sources of some of the code (base_test is lifted from [here](https://github.com/jafingerhut/p4-guide/blob/master/testlib/base_test.py for example).

I think it might be better to do this in a separate PR to keep things simpler.

Many thanks for the scripts in https://github.com/jafingerhut/p4-guide/tree/master/ptf-tests, Andy. They have been very useful in bootstrapping this.

mbudiu-vmw:
Our CI already takes very long, we should probably schedule these rather infrequently.

fruffy:
> Our CI already takes very long, we should probably schedule these rather infrequently.

Currently, the tools tests run faster than the normal compiler tests. I have some ideas how to speed up the STF tests, too.

jafingerhut:
Will these changes enable PTF tests to be checked in and run by CI, similar to the way that checking .stf files into the p4_16_samples directory causes STF tests to run during CI?

If yes, it would be nice to see one example of that.  It is perfectly fine if it is checked in as part of some future PR, rather than this one.

What control plane API do your PTF tests use to add table entries?  The simple_switch Thrift API?  P4Runtime API?

fruffy:
> Will these changes enable PTF tests to be checked in and run by CI, similar to the way that checking .stf files into the p4_16_samples directory causes STF tests to run during CI?

Yes, it is also much faster than STF checking

> If yes, it would be nice to see one example of that. It is perfectly fine if it is checked in as part of some future PR, rather than this one.

I can generate some sample tests, any particular program? 

> What control plane API do your PTF tests use to add table entries? The simple_switch Thrift API? P4Runtime API?
Currently, we use PTF with the P4Runtime API and simple_switch GRPC.

jafingerhut:
> > If yes, it would be nice to see one example of that. It is perfectly fine if it is checked in as part of some future PR, rather than this one.
> 
> I can generate some sample tests, any particular program?

ternary2-bmv2.stf has some table additions to tables with ternary fields, and some packets sent in with specific expected packets out.  And its STF test is quite short, so it should not be a lot of work.

fruffy:
> > > If yes, it would be nice to see one example of that. It is perfectly fine if it is checked in as part of some future PR, rather than this one.
> > 
> > 
> > I can generate some sample tests, any particular program?
> 
> ternary2-bmv2.stf has some table additions to tables with ternary fields, and some packets sent in with specific expected packets out. And its STF test is quite short, so it should not be a lot of work.

Okay, I will clean up the XFails and merge this PR. After I will contribute some tests for ternary2 which use the new framework. That program is quite dense actually, it produces ~300 unique tests.

fruffy (approved):

PR #3738, opened 2022-11-28
eBPF/PSA: Implement caching for ActionSelector, LPM and ternary tables

tatry:
This PR introduces optimization for the PSA/eBPF backend called `table caching` which adds cache table with only `exact` match keys for time consuming look-ups. We identified three situations where this optimization can improve performance:

- Table with `ternary` (and/or `lpm`, `exact`) key - skip slow TSS algorithm if the key was earlier matched.
- Table with `lpm` (and/or `exact`) key - skip slow `LPM_TRIE` map (especially when there is many entries) if the key was earlier matched.
- `ActionSelector` member selection from group - skip slow checksum calculation for `selector` key if it was earlier calculated.

This optimization may not improve performance in every case, so it must be explicitly enabled by compiler option `--table-caching`.

Limitations/to do:

- `DirectCounter` and `DirectMeter` are not supported by this optimization (tables with these externs will not have enabled cache).
- Cache size is a half of table size, but it would be more configurable or smart during compilation.
- Updates to tables with cache or `ActionSelector` require cache invalidation (`nikss` library will remove all cached entries).

### DIFF
--- a/backends/ebpf/psa/README.md
+++ b/backends/ebpf/psa/README.md
@@ -535,6 +535,23 @@ $ bpftool help
 
 Refer to [the bpftool guide](https://manpages.ubuntu.com/manpages/focal/man8/bpftool-prog.8.html) for more examples how to use it.
 
+# Performance optimizations
+
+## Table caching
+
+Table caching optimizes P4 table lookups by adding a cache with all `exact` matches for time-consuming lookups including:
+- table with `ternary` (and/or lpm, exact) key - skip slow TSS algorithm if the key was matched earlier.
+- table with `lpm` (and/or exact) key - skip slow `LPM_TRIE` map (especially when there is many entries) if the key was matched earlier.
+- `ActionSelector` member selection from group - skip slow checksum calculation for `selector` key if it was calculated earlier.
+
+The fast exact-match map is added in front of each instance of a table that contains a `lpm`, `ternary` or `selector` match
+key. The table cache is implemented with `BPF_MAP_TYPE_LRU_HASH`, which shares its implementation with the BPF hash map.
+The LRU map provides a good lookup performance, but lower performance on map updates due to a maintenance process. Thus,
+this optimization fits into use cases, where a value of table key changes infrequently between packets.
+
+This optimization may not improve performance in every case, so it must be explicitly enabled by compiler option. To enable
+table caching pass `--table-caching` to the compiler.
+
 # TODO / Limitations
 
 We list the known bugs/limitations below. Refer to the Roadmap section for features planned in the near future.
@@ -545,7 +562,7 @@ with some NICs. So far, we have verified the correct behavior with Intel 82599ES
 - `lookahead()` with bit fields (e.g., `bit<16>`) doesn't work.
 - `@atomic` operation is not supported yet.
 - `psa_idle_timeout` is not supported yet.
-- DirectCounter and DirectMeter externs are not supported for P4 tables with implementation (ActionProfile).
+- DirectCounter and DirectMeter externs are not supported for P4 tables with implementation (ActionProfile or ActionSelector).
 - The `xdp2tc=head` mode works only for packets larger than 34 bytes (the size of Ethernet and IPv4 header).
 - `value_set` only supports the exact match type and can only match on a single field in the `select()` expression.
 - The number of entries in ternary tables are limited by the number of unique ternary masks. If a P4 program uses many ternary tables and the `--max-ternary-masks` (default: 128) is set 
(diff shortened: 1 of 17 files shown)


backends/ebpf/psa/README.md
```diff
@@ -535,6 +535,23 @@ $ bpftool help
...
+# Performance optimizations
+
+## Table caching
+
+Table caching optimizes P4 table lookups by adding a cache with all `exact` matches for time-consuming lookups including:
+- table with `ternary` (and/or lpm, exact) key - skip slow TSS algorithm if the key was matched earlier.
+- table with `lpm` (and/or exact) key - skip slow `LPM_TRIE` map (especially when there is many entries) if the key was matched earlier.
+- `ActionSelector` member selection from group - skip slow checksum calculation for `selector` key if it was calculated earlier.
+
+The fast exact-match map is added in front of each instance of a table that contains a `lpm`, `ternary` or `selector` match
```
mbudiu-vmw, line 547: I also expect that modifying the original table may invalidate the cache.

backends/ebpf/psa/README.md
```diff
@@ -545,14 +562,20 @@ with some NICs. So far, we have verified the correct behavior with Intel 82599ES
...
   requires iteration over BPF maps. Note that the recent kernel introduced the [bpf_for_each_map_elem()](https://lwn.net/Articles/846504/) helper that should simplify the iteration process and help to overcome the current limitation.
 - Setting a size of ternary tables does not currently work. 
 - DirectMeter cannot be used if a table defines `ternary` match fields, as [BPF spinlocks are not allowed in inner maps of map-in-map](https://patchwork.ozlabs.org/project/netdev/patch/20190124041403.2100609-2-ast@kernel.org/).
+- Table cache optimization can't be enabled on tables with DirectCounter or DirectMeter due to two different states of a
+  table entry. Tables with these externs will not have enabled cache optimization even when enabled by compiler option.
+- When table cache optimization is enabled for a table, the number of cached entries is determined as a half of table size.
+  This would be more configurable or smart during compilation.
+- Updates to tables or ActionSelector with enabled table cache optimization require cache invalidation. `nikss` library
```
mbudiu-vmw, line 577: You may be able to do better if you keep enough state to figure out what to invalidate.
Invalidating the whole cache may be very expensive.

backends/ebpf/psa/ebpfPsaTable.cpp
```diff
@@ -899,4 +903,130 @@ cstring EBPFTablePSA::addPrefixFunc(bool trace) {
...
+
+    // TODO: make cache size calculation more smart. Consider using annotation or compiler option.
+    size_t cacheSize = std::max((size_t)1, size / 2);
+    builder->target->emitTableDecl(builder, cacheTableName, TableHashLRU,
+                                   "struct " + cacheKeyTypeName, "struct " + cacheValueTypeName,
+                                   cacheSize);
+}
+
+void EBPFTablePSA::emitCacheLookup(CodeBuilder* builder, cstring key, cstring value) {
+    cstring cacheVal = "cached_value";
```
mbudiu-vmw, line 963: if this behaves like a table could you have reused the code generation for a table by creating a fake table?

backends/ebpf/psa/externs/ebpfPsaTableImplementation.cpp
```diff
@@ -636,4 +654,137 @@ void EBPFActionSelectorPSA::verifyTableEmptyGroupAction(const EBPFTablePSA* inst
...
+        auto scalar = ebpfType->to<EBPFScalarType>();
+        unsigned width = 0;
+        if (scalar != nullptr) {
+            width = scalar->implementationWidthInBits();
+            memcpy = !EBPFScalarType::generatesScalar(width);
+        }
+
+        builder->emitIndent();
+        if (memcpy) {
+            builder->appendFormat("memcpy(&%s.%s, &", cacheKeyVar.c_str(), fieldName.c_str());
```
mbudiu-vmw, line 733: so ebpf supports memcpy... or perhaps this is just an intrinsic that the compiler knows about

mbudiu-vmw (approved):

backends/ebpf/psa/README.md
tatry: Cache invalidation is up to control plane. I think it never occur from eBPF program side.

backends/ebpf/psa/README.md
tatry: I agree, it might be expensive. I think that key and mask is enough to implement selective cache invalidation when updating table. Whole cache is invalidated because it is a simple algorithm and always work, but might be changed in the future.

backends/ebpf/psa/externs/ebpfPsaTableImplementation.cpp
tatry: Probably it maps to an intrinsic that the compiler knows about. I forgot to change it to `__builtin_memcpy`.

backends/ebpf/psa/ebpfPsaTable.cpp
tatry: I can say cache behaves like a table but not in the same way. Both have similarly named methods, but internals of these are different. So table's method have to be overridden by these methods for cache (this code will not disappear). And the main part of table lookup is done by control block (not a table itself), so it is not easily accessible from table.

I think it will be rather hard to reuse code for table, so I would keep as is.

PR #3726, opened 2022-11-23
Move some P4Tool compiler passes to the midend folder.

fruffy:
Fixes #3518. 

I merged the `CopyHeaders` pass into the `CopyStructures` pass. This should be checked since I am not sure whether I preserved its semantics correctly. 

Also there is a question whether `coverage.cpp` should become a part of the midend or remain part of tools. I think it can be generally useful.

PR #3739, opened 2022-11-29
Try to merge copyHeaders into copyStructures.

fruffy:
Follow up on #3726.


pkotikal (approved):

midend/copyStructures.cpp
```diff
@@ -62,54 +62,109 @@ const IR::Node* RemoveAliases::postorder(IR::P4Control* control) {
...
+                    statement->right);
         }
-        return new IR::BlockStatement(statement->srcInfo, *retval);
+        return statement;
     }
 
-    return statement;
+    const auto& srcInfo = statement->srcInfo;
+    auto* retval = new IR::IndexedVector<IR::StatOrDecl>();
+    if (const auto* list = statement->right->to<IR::ListExpression>()) {
```
mbudiu-vmw, line 84: I think this cannot happen anymore, these should have been replaced with StructExpressions.
This is a leftover from when we didn't have StructExpressions.

mbudiu-vmw (approved):

midend/copyStructures.cpp
fruffy: So where in the compiler do ListExpressions still exist? They are useful in that StructExpressions require NamedExpressions, a specific type of Expression.

mbudiu-vmw: They may still exist before the type checker replaces them with struct expressions. They show up as initializers for various data structures.

PR #3740, opened 2022-11-29
File filter workaround for CMake versions lower that 3.6.

fruffy-bfn:
Apparently `list(FILTER)` is only supported with CMake versions after 3.6. However, the compiler still supports CMake 3.0.2. This is a workaround to be able to filter files from list with a lower version. 

It might make sense to raise the minimum required CMake version. At least to the version Ubuntu 18.04 supplies, which is  3.10.2.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -459,7 +459,7 @@ install (DIRECTORY ${P4C_SOURCE_DIR}/p4include
 ################################ Linters Begin ################################
 
 file(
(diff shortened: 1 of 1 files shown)


mbudiu-vmw (approved):

davidbolvansky:
Not needed for us as we moved to newer cmake.

fruffy:
Closing in favour of #3745.

PR #3745, opened 2022-11-30
Raise minimum required CMake version to Ubuntu 18.04 version.

fruffy:
This PR is a proposal to raise the minimum required CMake version to 3.10.2, which is the default Ubuntu 18.04 version. Our CI only tests that version and support for older CMake versions can not be guaranteed. 

The number is tentative, it could also be 3.6, since the current CMake has code that requires at least 3.6.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -12,7 +12,8 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-cmake_minimum_required (VERSION 3.16.3 FATAL_ERROR)
+# This is the CMake version supplied by Ubuntu 18.04.
+cmake_minimum_required (VERSION 3.10.2 FATAL_ERROR)
 
 find_program(CCACHE_PROGRAM ccache)
 if(CCACHE_PROGRAM)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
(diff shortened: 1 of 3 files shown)


apinski-cavium:
3.10.2 is the cmake version which is included in Ubuntu 1804 and it works just fine.  Can you just increase the min version to 3.10.2 ?

davidbolvansky:
p4c does not promise support for 18.04, nothing is tested on 18.

Or something changed and officially yes?

it would be great to make a decision - either fully support it or officially drop it.

jafingerhut:
This project is run by volunteers, pretty much, so it seems the word "official" might be a bit overstated.

That said, it seems worth not going out of our way to intentionally break builds on Ubuntu 18.04 until 2023-Apr when Ubuntu 18.04 reaches end of free long term support.

apinski-cavium:
Note I am still use ubuntu 1804 and compile p4c and test just fine but will be moving next year to a newer ubuntu.  The machine I use for main development is one I got in late 2018 and has been stable which is why I don't move as fast as some folks move to newer OS. I don't want to interrupt my flow with distro upgrades.

davidbolvansky:
> This project is run by volunteers, pretty much, so it seems the word "official" might be a bit overstated.

This is how it is written in README.

> That said, it seems worth not going out of our way to intentionally break builds on Ubuntu 18.04 until 2023-Apr when Ubuntu 18.04 reaches end of free long term support.

This seems like a reasonable approach. It would be good to document it in README somehow that after Apr 2023 things would change.

davidbolvansky (approved):

PR #3748, opened 2022-12-01
Implement support for non-const table entries

mihaibudiu:

This validates the design in https://github.com/p4lang/p4-spec/pull/1180


mbudiu-vmw:
This implementation does not have the code to generate the priorities yet, but validates the grammar

mbudiu-vmw:
there are a bunch of failures due to the inability to install the software in CI.
I don't understand the bazel complaints:
```
ERROR: /home/runner/work/p4c/p4c/BUILD.bazel:180:11: Compiling frontends/common/constantFolding.cpp failed: undeclared inclusion(s) in rule '//:ir_frontend_midend_control_plane':
this rule is missing dependency declarations for the following files included by 'frontends/common/constantFolding.cpp':
  '/usr/lib/gcc/x86_64-linux-gnu/11/include/stddef.h'
  '/usr/lib/gcc/x86_64-linux-gnu/11/include/stdarg.h'
```
I haven't touched this file.

mbudiu-vmw:
@smolkaj any clues why bazel is unhappy?

smolkaj:
This is the error message:
```
ERROR: /home/runner/work/p4c/p4c/BUILD.bazel:180:11: Compiling frontends/common/constantFolding.cpp failed: undeclared inclusion(s) in rule '//:ir_frontend_midend_control_plane':
this rule is missing dependency declarations for the following files included by 'frontends/common/constantFolding.cpp':
  '/usr/lib/gcc/x86_64-linux-gnu/11/include/stddef.h'
  '/usr/lib/gcc/x86_64-linux-gnu/11/include/stdarg.h'
  '/usr/lib/gcc/x86_64-linux-gnu/11/include/stdint.h'
  '/usr/lib/gcc/x86_64-linux-gnu/11/include/limits.h'
  '/usr/lib/gcc/x86_64-linux-gnu/11/include/syslimits.h'
  '/usr/lib/gcc/x86_64-linux-gnu/11/include/float.h'
...
```
Did you make any changes to `frontends/common/constantFolding.cpp`? Not as far as I can tell?

mbudiu-vmw:
I didn't that's why I find this puzzling.

mbudiu-vmw:
@smolkaj Some people say it's a caching bug: https://stackoverflow.com/questions/53527134/bazel-missing-dependency-declarations-for-stddef-h-and-others-ubuntu-18-04

smolkaj:
You could try clearing the cache here? https://github.com/p4lang/p4c/actions/caches?query=key%3ALinux-build+

smolkaj:
I tried on my laptop and couldn't reproduce the issue. So could indeed be a caching issue. If we can't figure out how to clear the cache, here are some random things we might try...:

- add a `.bazelrc` with the following contents to the root of the repo:
  ```
  # Use C++17.
  build: --cxxopt=-std=c++17 --host_cxxopt=-std=c++17
  ```
- replace `ubuntu-latest` with `ubuntu-22.04` in `ci-bazel.yml`

fruffy:
I cleared the cache. Looks like it is fixed.

mbudiu-vmw:
@jafingerhut can you please review this?

mbudiu-vmw:
If we settle on this syntax I will implement the additional features in the spec PR, e.g., priority assignment.

mihaibudiu:
@jafingerhut the last commit is meant to implement the spec completely with regards to non-const entries

jafingerhut:
@mihaibudiu Thanks!  I will try to review at least the test cases, and perhaps parts of the implementation regarding generation of warnings, soon-ish.

mihaibudiu:
I have tested the warning generation too, but was too lazy to add the tests.

jafingerhut:
@mihaibudiu Fortunately adding test cases and checking their outputs is my p4c specialty! :-)

jnfoster:
Human p4testgen ;-)

jafingerhut:
Thanks for this implementation.  I noticed that it is a bit different in the way it assigns priorities to entries than the language spec proposes.

The spec says that if some priorities are explicitly given by the P4 developer in the source code, but not all, that

(1) the first entry must have a priority explicitly specified.
(2) pseudocode in the spec explains how all other entry priorities are determined, from earliest entries in the source to the last entry in the source.

This PR as of commit 4 has a different implementation, where if largest_priority_wins is true, it assigns priorities from last entry in the source back to the first.

I think it would be good for the spec and implementation to match in how priorities are assigned to entries.  Either we should review a PR that updates the spec to match this implementation, or we should make this implementation match the spec.

jafingerhut:
Another thing that would be good to have: in the code that determines the entry priorities, save the results of the determined priority values in such a way that the control plane API generation code can read and output those priority values.

I understand that if you want to emit P4_16 source code from the IR after the priorities are determined, in such a way that only the entries that the P4 developer explicitly specified priority values will have them in that printed P4_16 source code later, you would need a way in the IR to remember which entries have priority values from the source code, versus which ones have priority values determined by the compiler.  Either that, you the P4_16 code emitted from the IR should simply have explicit priority values for every entry, which I cannot think of any harm if we do that.  Not only would there be no harm, but it would be a way to verify that the compiler-determined priority values were correct, by diff'ing them against the priority values in the `-frontend.p4` and/or `-midend.p4` files output by the compiler, and compared against files stored in this repo.

frontends/p4/entryPriorities.cpp
```diff
@@ -0,0 +1,85 @@
...
+#include "entryPriorities.h"
+
+namespace P4 {
+
+const IR::Node *EntryPriorities::preorder(IR::EntriesList *entries) {
+    auto table = findContext<IR::P4Table>();
+    CHECK_NULL(table);
+    auto ep = table->properties->getProperty(IR::TableProperties::entriesPropertyName);
+    CHECK_NULL(ep);
+    if (ep->isConstant) return entries;
```
jafingerhut, line 26: I think that if method tableNeedsPriority() returns false for this table, that we should _not_ execute the code below that calculates entry priorities.  Instead, we should issue a warning or error if the source code contains explicit priority values for any entry.

If the P4 developer tries to specify priority values for such a table, I think they have a fundamental misunderstanding, and should get an error that says something like "This table cannot support entry priorities.  For example, tables with only exact and lpm match kinds cannot support entry priorities".

jafingerhut:
I have attached a zip file to this comment containing 4 new test programs for this issue.  Each has one or more table definitions in it, with comments explaining what warnings or errors I would expect the compiler to issue for it.  The ones in p4_16_errors currently cause no errors to be issued with the changes in this PR as of commit 4.

[p4c-pull-3748-test-progs.zip](https://github.com/p4lang/p4c/files/11523484/p4c-pull-3748-test-progs.zip)

frontends/p4/entryPriorities.cpp
jafingerhut: Inside the zip file I have attached to this comment: https://github.com/p4lang/p4c/pull/3748#issuecomment-1556059443 the test program init-entries-error1-bmv2.p4 contains 3 different but similar table definitions that I believe should all give an error message at compile time for the reason mentioned in my comment above.

mihaibudiu:
> Another thing that would be good to have: in the code that determines the entry priorities, save the results of the determined priority values in such a way that the control plane API generation code can read and output those priority values.

This is what we are doing: after this code every entry in a non-const entries table will have a priority.

mihaibudiu:
I hope that now I've got it right.

mihaibudiu:
Let me add the test cases in the zip file too

mihaibudiu:
A more important problem is that we get this error:
`init-entries-bmv2.p4(249): [--Werror=unsupported] error: entries: P4Runtime only supports constant table initializers`
I am not sure how to fix this.

jafingerhut:
> A more important problem is that we get this error: `init-entries-bmv2.p4(249): [--Werror=unsupported] error: entries: P4Runtime only supports constant table initializers` I am not sure how to fix this.

This comes from the code that writes P4Runtime API output files, two of them:

+ The P4Info file
+ The 'entries' file, triggered by the command line option `--p4runtime-entries-files`

Currently, the writing of the entries file generates its own priority values independently of the pass that you added, always assigning priority values from N for the first entry down to 1 for the last entry.  It should be updated to use the priority values calculated earlier, if largest_priority_wins is true, or something like (M-P) for each entry if largest_priority_wins is false, where P is the calculated priority of the entry, and M is the maximum priority value supported, perhaps 2^32 or 2^32-1.

mihaibudiu:
By the time the API generation is reached the priorities will be all set.
Perhaps we can just remove the const check

jafingerhut:
The priority values might be all set in the pass that you added, but the p4RuntimeSerializer code currently ignores them.  It would be good to update that code to use the priority values your new code calculates (if largest_priority_wins is true, or the modification I suggested in my previous comment if it is false).

frontends/p4/entryPriorities.cpp
```diff
@@ -0,0 +1,160 @@
...
+                                          priority, entry->keys, entry->action, entry->singleton);
+            entries->entries[i] = newEntry;
+        }
+        auto previous = usedPriorities.find(currentPriority);
+        if (previous != usedPriorities.end()) {
+            warn(ErrorType::WARN_DUPLICATE_PRIORITIES, "%1% and %2% have the same priority %3%",
+                 previous->second, entries->entries[i], currentPriority);
+        }
+        usedPriorities.emplace(currentPriority, entries->entries[i]);
+        if (i > 0 && currentPriority < previousPriority) {
```
jafingerhut, line 134: I think this condition needs to be something like this:

```
i > 0 && ((largestWins && currentPriority > previousPriority) || (!largestWins && currentPriority < previousPriority))
```
As written now, it produces warnings when largestWins is true, in cases where it should not.

jafingerhut:
FYI here is a link to the code that writes out a P4Runtime 'entries' file: https://github.com/p4lang/p4c/blob/main/control-plane/p4RuntimeSerializer.cpp#L1009-L1035

Note that it generates priority values perfectly correctly for `const entries` before the enhancement to the language spec was written, by always assigning largest_priority_wins priority values from N for the first entry in the source code, down to 1 for the last entry in the source code.

mihaibudiu:
I hope I have fixed the warning.
The last commit also modifies P4Runtime to use the priorities if they are specified.

mihaibudiu:
@fruffy all these are testgen failures, can you take a look at them?
Thanks,

jafingerhut:
@mihaibudiu It seems that perhaps the expected output files for 2 of the new test programs need to be updated as well?

jafingerhut (approved):
I have checked all of the test programs, and their expected output files, and they look good to me, except for some things that have not yet been decided in the P4 API work group, so it seems best to save any such changes for later after that group has decided them.  I will create a separate issue to track that.  They should be very minor changes on top of this PR.

fruffy:
> @fruffy all these are testgen failures, can you take a look at them? Thanks,

There is only one testgen failure and it looks like it does not support the current non-const table setup. 
This is the error:
`[02:33:30.992] [bmv2] [E] [thread 61472] [P4Runtime] Error when adding match entry to target`
Could you add it to the Xfails?

mihaibudiu:
> > @fruffy all these are testgen failures, can you take a look at them? Thanks,
> 
> There is only one testgen failure and it looks like it does not support the current non-const table setup. This is the error: `[02:33:30.992] [bmv2] [E] [thread 61472] [P4Runtime] Error when adding match entry to target` Could you add it to the Xfails?

Hope I did it in the right way, the error report has a few tens of pages

fruffy:
I would just add

```

p4tools_add_xfail_reason(
   "testgen-p4c-bmv2-ptf"
   "Error when adding match entry to target"
   init-entries-bmv2.p4
)

```

to the BMV2PTFXfail.cmake file.

backends/p4tools/modules/testgen/targets/bmv2/test/''
```diff
@@ -0,0 +1,290 @@
+# XFAILS: tests that currently fail. Most of these are temporary.
```
fruffy, line 1: I think something went wrong with the file here.

fruffy (approved):

jafingerhut:
Is it clear to anyone what might be causing the macOS test failures?

fruffy:
```
2023-05-26T18:21:40.7270690Z /Users/runner/work/p4c/p4c/frontends/p4/entryPriorities.cpp:77:33: error: call to constructor of 'IR::Constant' is ambiguous
2023-05-26T18:21:40.7276490Z             auto priority = new IR::Constant(currentPriority);
```

fruffy:
Also recommend adding this: 
https://github.com/p4lang/p4c/pull/3970/files#diff-1e7de1ae2d059d21e1dd75d5812d5a34b0222cef273b7c3a2af62eb747f9d20aR241
to get rid of the excessive warnings, it should help.

fruffy:
@mihaibudiu I merged a PR that cleans up the MacOS error message. I believe there is still an error of the same kind left.

jafingerhut:
Yay!  All tests passing, 2 positive reviews.  Time to merge?  I am happy to do it.

PR #3749, opened 2022-12-01
CI: also run on Ubuntu 20 and 18

davidbolvansky:
_Ubuntu-latest label will now use Ubuntu-22.04_

https://github.blog/changelog/2022-12-01-github-actions-larger-runners-using-ubuntu-latest-label-will-now-use-ubuntu-22-04/

-> Preserve testing on Ubuntu 20.04.

Also I added build & test job on Ubuntu 18.04 as it seems (see link below) that some vendors / developers need to support p4c-based compilers (so working p4c is crucial) on Ubuntu 18.04, possibly for longer time than till official EOL date.
It makes no sense to intentionally break builds or just intentionally drop support for Ubuntu 18.04.

https://github.com/p4lang/p4c/pull/3745

### DIFF
--- a/.github/workflows/ci-test.yml
+++ b/.github/workflows/ci-test.yml
@@ -12,8 +12,8 @@ concurrency:
   cancel-in-progress: ${{ github.ref != 'refs/heads/main' }}
 
 jobs:
-  # Build with gcc and test p4c on Ubuntu 20.04.
-  test-linux-sanitizers:
+  # Build with gcc and test p4c on Ubuntu 22.04.
+  test-ubuntu22-gcc-sanitizers:
     strategy:
       fail-fast: false
       matrix:
@@ -33,7 +33,7 @@ jobs:
         key: test-${{ matrix.unified }}-${{ runner.os }}-gcc
         max-size: 1000M
 
-    - name: Build (Ubuntu Linux, GCC, Sanitizers)
+    - name: Build (Ubuntu 22.04, GCC, Sanitizers)
       run: |
         docker build -t p4c --build-arg IMAGE_TYPE=test --build-arg ENABLE_UNIFIED_COMPILATION=${{ matrix.unified }} --build-arg ENABLE_SANITIZERS=ON .
         ./tools/export_ccache.sh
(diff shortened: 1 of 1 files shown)


fruffy:
I think one of the issues is that we run everything in docker, which is built on Ubuntu 20.04. Changing the action Ubuntu version will not have any impact.

davidbolvansky:
> I think one of the issues is that we run everything in docker, which is built on Ubuntu 20.04. Changing the action Ubuntu version will not have any impact.

Hm right.

fruffy:
If we find a way to prepackage BMv2 we could simplify this entire process quite a lot. That is the only dependency that takes time to install.

fruffy:
How often is this updated? Are able to fix the version?

rst0git:
Yes, I'm maintaining these packages. We are also using these packages to build the P4 Tutorials VM (https://github.com/p4lang/tutorials/pull/439).

fruffy:
This could be an option. We could install dependencies the same way the Fedora test does. Then we could freely test different Ubuntu versions.

fruffy:
> > If we find a way to prepackage BMv2 we could simplify this entire process quite a lot.
> 
> BMv2 can be installed from package as follows:
> 
> ```
> . /etc/os-release
> echo "deb http://download.opensuse.org/repositories/home:/p4lang/xUbuntu_${VERSION_ID}/ /" | sudo tee /etc/apt/sources.list.d/home:p4lang.list
> curl -L "http://download.opensuse.org/repositories/home:/p4lang/xUbuntu_${VERSION_ID}/Release.key" | sudo apt-key add -
> sudo apt-get update
> sudo apt install p4lang-bmv2
> ```

Can you add this to the behavioral model installation? This seem like useful information.

davidbolvansky:
> > If we find a way to prepackage BMv2 we could simplify this entire process quite a lot.
> 
> BMv2 can be installed from package as follows:
> 
> ```
> . /etc/os-release
> echo "deb http://download.opensuse.org/repositories/home:/p4lang/xUbuntu_${VERSION_ID}/ /" | sudo tee /etc/apt/sources.list.d/home:p4lang.list
> curl -L "http://download.opensuse.org/repositories/home:/p4lang/xUbuntu_${VERSION_ID}/Release.key" | sudo apt-key add -
> sudo apt-get update
> sudo apt install p4lang-bmv2
> ```

Nice, so this effort have restarted with your steps:
https://github.com/p4lang/p4c/pull/3770

It seems you have no packages for ubuntu 18, right?

rst0git:
> It seems you have no packages for ubuntu 18, right?

No. The versions of some of the dependencies of BMv2 available as packages are too old in Ubuntu 18.04. Creating a BMv2 package for Ubuntu 18.04 would also require creating packages for these dependencies.

PR #3752, opened 2022-12-02
PSA/ebpf backend: add support for generating programs for XDP hook

tatry:
This PR adds functionality to generate programs for the XDP layer for the PSA/ebpf (part of NIKSS) backend.

Update of `libbpf` (new version v0.3) is required for definition of `bpf_devmap_val` struct which is available starting from this version.

CC: @osinstom @kmateuszssak @fdangtran @elfadel


backends/ebpf/psa/README.md
```diff
@@ -11,9 +11,16 @@ This directory implements PSA (Portable Switch Architecture) for the eBPF backen
 
 # Design
 
+The PSA to eBPF compiler provides two flavors of generated eBPF code: TC-based design and XDP-based design.
+The TC-based design leverages eBPF TC (Traffic Control) hook and is able to implement any PSA program at the cost of performance.
```
mbudiu-vmw, line 15: you can probably remove "at the cost of performance".

backends/ebpf/psa/README.md
```diff
@@ -11,9 +11,16 @@ This directory implements PSA (Portable Switch Architecture) for the eBPF backen
...
+The TC-based design leverages eBPF TC (Traffic Control) hook and is able to implement any PSA program at the cost of performance. 
+The XDP-based design offloads packet processing to eBPF XDP (eXpress Data Path) hook and provides better performance than the TC-based flavor.
+However, the XDP-based design lacks support for packet recirculation, QoS (no integration with TC qdisc) and CLONE_E2E packet path.
+
+## TC-based design (default)
+
 P4 packet processing is translated into a set of eBPF programs attached to the TC hook. The eBPF programs implement packet processing defined
 in a P4 program written according to the PSA model. The TC hook is used as a main engine, because it enables a full implementation of the PSA specification.
-We also plan to contribute the XDP-based version of the PSA implementation that does not implement the full specification, but provides better performance.
+The XDP-based version of the PSA implementation does not implement the full specification, but provides better performance.
```
mbudiu-vmw, line 23: can you provide a ballpark estimate of the improvement so people can decide which one to use?

backends/ebpf/psa/ebpfPipeline.cpp
```diff
@@ -672,4 +672,220 @@ void TCEgressPipeline::emitTrafficManager(CodeBuilder* builder) {
...
+}
+
+// =====================TCTrafficManagerForXDP=============================
+
+void TCTrafficManagerForXDP::emitGlobalMetadataInitializer(CodeBuilder* builder) {
+    EBPFPipeline::emitGlobalMetadataInitializer(builder);
+
+    // if Traffic Manager decided to pass packet to the kernel stack earlier, send it up immediately
+    builder->emitIndent();
+    builder->appendFormat("if (%s->pass_to_kernel == true) return %s;", compilerGlobalMetadata,
```
mbudiu-vmw, line 761: usually == true does not improve readability

backends/ebpf/psa/ebpfPipeline.cpp
```diff
@@ -672,4 +672,220 @@ void TCEgressPipeline::emitTrafficManager(CodeBuilder* builder) {
...
+
+    builder->blockEnd(true);
+}
+
+void TCTrafficManagerForXDP::emitReadXDP2TCMetadataFromHead(CodeBuilder* builder) {
+    builder->emitIndent();
+    builder->append(
+        "    void *data = (void *)(long)skb->data;\n"
+        "    void *data_end = (void *)(long)skb->data_end;\n"
+        "    if (((char *) data + 14 + sizeof(struct xdp2tc_metadata)) > (char *) data_end) {\n"
```
mbudiu-vmw, line 803: 14?

backends/ebpf/psa/ebpfPsaDeparser.cpp
```diff
@@ -179,4 +179,125 @@ void TCIngressDeparserPSA::emitPreDeparser(CodeBuilder* builder) {
...
+        builder->emitIndent();
+        builder->append("if (ret) ");
+        builder->blockStart();
+        builder->target->emitTraceMessage(builder, "Deparser: failed to push XDP2TC metadata");
+        builder->emitIndent();
+        builder->appendFormat("return %s;", builder->target->abortReturnCode().c_str());
+        builder->newline();
+        builder->blockEnd(true);
+        builder->append(
+            "    data = (void *)(long)skb->data;\n"
```
mbudiu-vmw, line 246: this pattern shows up a lot, is it worthwhile to factor into a macro?

backends/ebpf/psa/ebpfPsaDeparser.cpp
```diff
@@ -179,4 +179,125 @@ void TCIngressDeparserPSA::emitPreDeparser(CodeBuilder* builder) {
...
+            "    data_end = (void *)(long)skb->data_end;\n"
+            "    if (((char *) data + 14 + sizeof(struct xdp2tc_metadata)) > (char *) data_end) {\n"
+            "        return XDP_ABORTED;\n"
+            "    }\n");
+        builder->emitIndent();
+        builder->appendLine("__builtin_memmove(data, data + sizeof(struct xdp2tc_metadata), 14);");
+        builder->emitIndent();
+        builder->appendLine(
+            "__builtin_memcpy(data + 14, "
```
mbudiu-vmw, line 255: in general I think it's good to document these magic constants

backends/ebpf/psa/ebpfPsaGen.cpp
```diff
@@ -423,93 +511,107 @@ void PSAArchTC::emitInstances(CodeBuilder* builder) const {
...
+        "    __u16 pkt_ether_type;\n"
+        "} __attribute__((aligned(4)));",
+        tcIngressForXDP->parser->headerType->to<EBPFStructType>()->name);
+    builder->newline();
 }
 
-void PSAArchTC::emitInitializerSection(CodeBuilder* builder) const {
-    builder->appendLine("SEC(\"classifier/map-initializer\")");
+void PSAArchXDP::emitDummyProgram(CodeBuilder* builder) const {
+    // this is static program, so we can just paste a piece of code.
```
mbudiu-vmw, line 605: I don't understand this comment

mbudiu-vmw (approved):

backends/ebpf/ebpfBackend.cpp
```diff
@@ -73,7 +73,10 @@ void run_ebpf_backend(const EbpfOptions& options, const IR::ToplevelBlock* tople
 
     Target* target;
     if (options.target.isNullOrEmpty() || options.target == "kernel") {
-        target = new KernelSamplesTarget(options.emitTraceMessages);
+        if (!options.generateToXDP)
```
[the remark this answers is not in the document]
osinstom: I think the main reason is that we treat XDP as kind of optimization of the default TC-based design, since our design in XDP doesn't implement full PSA spec (see README - no CoS, no CLONE_E2E, no packet recirculation). 

But to be honest, I don't have a strong opinion here. We could probably make it a new target pretty easily but then we would require some refactoring of other targets (e.g., `kernel` target is not a good name anymore since we have more than 1 kernel-based target, also how to resolve conflicts between PSA-eBPF and p4c-xdp?).

backends/ebpf/psa/README.md
osinstom: Yes, I think we could @tatry, but the improvement depends on the P4 program. A ball park estimate is over 2x improvement in throughput per CPU core measured in MPPS.

You can find an extensive performance evaluation in our recent paper: https://dl.acm.org/doi/10.1145/3555050.3569117. We could also reference this paper in the README?

backends/ebpf/psa/ebpfPsaDeparser.cpp
tatry: I agree that it requires something like that.

I think rather about a function that takes a block of C code. It would split code into lines and for each line emits indent, line of code, and makes some string replace if required. Then it would be applied to a lot of places of PSA/eBPF backend. In such case I think this should be separate PR. What do you think about it?

backends/ebpf/psa/ebpfPipeline.cpp
tatry: The size of Ethernet header. It could be replaced by a proper `sizeof`.

PR #3754, opened 2022-12-02
Check that value sets have a constant constructor argument

### LINKED ISSUE #3751: value_set with a size that is not a compiler known constant
Take:
```
parser p(in bit<32> t)
{
  value_set<bit>(t) vs;
  state start { transition accept; }
}

parser pt(in bit<32> t);
package pp(pt t);
pp(p()) main;
```
I would have assumed that size of the value_set needs to be a compiler known constant as the specifications says:
```
The semantics of the size argument is similar to the size property of a table.
```
And then for size property has:
```
When present, its value must always be an integer compile-time known value.
```

Maybe a specifications improvement is still needed for value_set to be more explicit about what the size argument could be.

mbudiu-vmw:
Fixes #3751

### DIFF
--- a/frontends/p4/validateValueSets.h
+++ b/frontends/p4/validateValueSets.h
@@ -28,9 +28,7 @@ namespace P4 {
  */
 class ValidateValueSets final : public Inspector {
  public:
-    ValidateValueSets() {
-        setName("ValidateValueSets");
-    }
+    ValidateValueSets() { setName("ValidateValueSets"); }
     void postorder(const IR::P4ValueSet* valueSet) override {
         if (!valueSet->size->is<IR::Constant>()) {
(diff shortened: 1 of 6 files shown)


frontends/p4/validateValueSets.h
```diff
@@ -0,0 +1,44 @@
...
+
+namespace P4 {
+
+/**
+ * Checks that match annotations only have 1 argument which is of type match_kind.
+ */
+class ValidateValueSets final : public Inspector {
+ public:
+    ValidateValueSets() {
+        setName("ValidateValueSets");
```
ChrisDodd, line 32: This seems (mostly) pointless as the default name for the pass will be the (qualified) name for the pass `P4::ValidateValueSets`

ChrisDodd (approved):

PR #3756, opened 2022-12-04
Testgen: Switch cases can fall through.

fruffy-bfn:
Testgen did not model the case where a switch case matches, but falls through to another case. This PR keeps track of a match boolean causes any downstream switch cases to be executed.

### DIFF
--- a/backends/p4tools/modules/testgen/core/small_step/cmd_stepper.cpp
+++ b/backends/p4tools/modules/testgen/core/small_step/cmd_stepper.cpp
@@ -596,19 +596,24 @@ bool CmdStepper::preorder(const IR::SwitchStatement* switchStatement) {
         cmds.emplace_back(Continuation::PropertyUpdate("inUndefinedState", currentTaint));
     } else {
         // Otherwise, we pick the switch statement case in a normal fashion.
+
+        bool hasMatched = false;
         for (const auto* switchCase : switchStatement->cases) {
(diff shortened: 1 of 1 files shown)


backends/p4tools/modules/testgen/core/small_step/cmd_stepper.cpp
```diff
@@ -596,19 +596,24 @@ bool CmdStepper::preorder(const IR::SwitchStatement* switchStatement) {
         cmds.emplace_back(Continuation::PropertyUpdate("inUndefinedState", currentTaint));
     } else {
         // Otherwise, we pick the switch statement case in a normal fashion.
+
+        bool hasMatched = false;
         for (const auto* switchCase : switchStatement->cases) {
-            // Nothing to do with this statement, fall through to the next case.
+            // We have either matched already, or still need to match.
+            hasMatched = hasMatched || switchStatement->expression->equiv(*switchCase->label);
```
hannelita, line 603: Sorry, I don't understand when exactly we'd be changing this flag, i.e. you are doing a reassignment on the LHS, but when we'd be turning that into true?

fruffy-bfn: If isMatched is `false` but the expression is equivalent to the label isMatched is set to true. After, all subsequent cases should also be executed until we hit a case that has a block statement..

vhavel (approved):

PR #3758, opened 2022-12-06
Dpdk Backend: Fix action selector and action profile to apply the group/member tables only on base table hit

usha1830:
A table wit action selector property is split into base table, group table and member table.

Prior to this PR,  group and member tables were unconditionally applied
```
table.apply() ==> 
table0.apply()
group_table.apply()
member_table.apply()
```
This PR adds conditions for applying the group and member tables
The table apply can occur as independent MethodCallStatement, within IfStatement and SwitchStatement. Code comments explain how each of these cases are converted


backends/dpdk/dpdkArch.h
```diff
@@ -842,33 +842,40 @@ class CopyMatchKeysToSingleStruct : public P4::KeySideEffect {
...
 class SplitP4TableCommon : public Transform {
+    cstring switchExprTmp;
+    DeclarationInjector injector;
+
  public:
     enum class TableImplementation { DEFAULT, ACTION_PROFILE, ACTION_SELECTOR };
     P4::ReferenceMap* refMap;
     P4::TypeMap* typeMap;
     DpdkProgramStructure* structure;
+    std::map<cstring, std::vector<std::tuple<cstring, IR::Constant*>>>& actionCaseMap;
```
mbudiu-vmw, line 853: in general I try to document such complicated maps, explaining what the keys and values are.

mbudiu-vmw: an alternative is to wrap this into a separate class

backends/dpdk/dpdkArch.h
```diff
@@ -894,26 +902,59 @@ class SplitActionSelectorTable : public SplitP4TableCommon {
...
+    SplitActionProfileTable(
+        P4::ReferenceMap* refMap, P4::TypeMap* typeMap, DpdkProgramStructure* structure,
+        std::map<cstring, std::vector<std::tuple<cstring, IR::Constant*>>>& actionCaseMap)
+        : SplitP4TableCommon(refMap, typeMap, structure, actionCaseMap) {
         implementation = TableImplementation::ACTION_PROFILE;
     }
     const IR::Node* postorder(IR::P4Table* tbl) override;
 };
 
+class UpdateActionForSwitch : public Transform {
```
mbudiu-vmw, line 914: I also try to document the passes I am building.

backends/dpdk/dpdkArch.h
```diff
@@ -894,26 +902,59 @@ class SplitActionSelectorTable : public SplitP4TableCommon {
...
+        return action;
+    }
+};
+
 /**
- * Handle ActionSelector and ActionProfile extern in PSA
- */
+ * Handle ActionSelector and ActionProfile extern */
 class ConvertActionSelectorAndProfile : public PassManager {
+    // Map which holds the switch expression variable and constant tuple per switch statement for
```
mbudiu-vmw, line 943: I see the documentation here.
But if you define a separate class, that's the natural place to put it.

backends/dpdk/dpdkArch.cpp
```diff
@@ -2116,137 +2127,182 @@ const IR::Node* SplitP4TableCommon::postorder(IR::MethodCallStatement* statement
 
         if (group_ids.count(tableName) != 0) {
             auto group_id = group_ids.at(tableName);
+            group_id_expr = new IR::PathExpression(group_id);
             decls->push_back(new IR::AssignmentStatement(
-                new IR::PathExpression(group_id), new IR::Constant(IR::Type_Bits::get(32), 0)));
+                group_id_expr, new IR::Constant(IR::Type_Bits::get(32), 0xFFFFFFFF)));
```
mbudiu-vmw, line 2132: any explanation about what this constant means?

mbudiu-vmw: maybe you can just make it a utility function, I see it's used many times.

backends/dpdk/dpdkArch.cpp
```diff
@@ -2116,137 +2127,182 @@ const IR::Node* SplitP4TableCommon::postorder(IR::MethodCallStatement* statement
...
-        //   foo(); } } }
         //
         // if (t.apply().hit) {
         //   foo();
         // }
         // is equivalent to
         // if (t0.apply().hit) {
-        //   if (t1.apply().hit) {
-        //      if (t2.apply().hit) {
+        //    if (group_id != initial_groupid) {
```
mbudiu-vmw, line 2204: maybe you could add an else branch to the top if

mbudiu-vmw (approved):

backends/dpdk/dpdkArch.h
usha1830: Just to clarify, are you suggesting to put this actionCaseMap into a separate class?

backends/dpdk/dpdkArch.h
usha1830: Sure, will do.

backends/dpdk/dpdkArch.h
mbudiu-vmw: Yes

backends/dpdk/dpdkArch.cpp
usha1830: For the "if" in line:2203? we are only replacing the condition in top if and adding some additional statements within the ifTrue block. If there is an else in the input P4, it will still remain there.

mbudiu-vmw (approved):

PR #3759, opened 2022-12-06
Apply clang-format to ir and lib files.

fruffy:


mbudiu-vmw:
Many of my PRs now fail because I forgot to reformat the code.
This is a big waste of CI resources.
I really think you should not be allowed to push of the code is not formatted.

fruffy:
clang-format is part of the [git hook](https://github.com/p4lang/p4c/blob/main/tools/hooks/pre-commit#L73) script, but it is optional and has to be installed manually. I do not think there is a straightforward way to make this automatically part of P4C's commit flow.

fruffy:
The issue is largely the same as the ref files, but for linting you will get feedback early now.

We could make it so that linting has to pass first before tests are run, but sometimes you do not care about linting.

rst0git:
> clang-format is part of the git hook script, but it is optional and has to be installed manually. I do not think there is a straightforward way to make this automatically part of P4C's commit flow.

It should be possible to run clang-format in our CI. Here is an example of how it is used in another open-source project:
https://github.com/checkpoint-restore/criu/blob/criu-dev/.github/workflows/lint.yml

mbudiu-vmw (approved):

PR #3762, opened 2022-12-07
Refactor add_cpplint_files and add_clang_format files to support downstream projects, large file numbers.

fruffy-bfn:
The linters were unable to support large lists of files because of a [Make/Unix limitation](https://unix.stackexchange.com/questions/480685/argument-list-too-long-error-with-makefile).

This PR does two things. 
1) It refactors how linter files are collected by setting a global property. Now downstream projects can directly write to this global property, which is then consumed by top-level P4C.
2) It writes the collected linter files to a file, which is then piped into the appropriate linter. This way we avoid the "too many arguments" error that may occur.


davidbolvansky (approved):
add_cpplint_files changes LGTM.

CMakeLists.txt
```diff
@@ -507,18 +509,31 @@ list(FILTER P4C_LINT_LIST EXCLUDE REGEX "control-plane/p4runtime")
...
+
 if(DEFINED CPPLINT_FILES)
-  list ( SORT CPPLINT_FILES )
-  set (CPPLINT_CMD ${P4C_SOURCE_DIR}/tools/cpplint.py)
-  set (CPPLINT_ARGS --root=${P4C_SOURCE_DIR} --extensions=h,hpp,cpp,ypp,l)
-  add_custom_target(cpplint
-    COMMAND ${CPPLINT_CMD} ${CPPLINT_ARGS} ${CPPLINT_FILES}
+  # Replace the semicolon-separators with space separators.
+  string(REPLACE ";" " " CPPLINT_FILES "${CPPLINT_FILES}")
+  # Write the list to a file.
```
davidbolvansky, line 518: I would mention why need need this workaround.

cmake/P4CUtils.cmake
```diff
@@ -49,7 +49,8 @@ macro (p4c_add_library name symbol var)
 endmacro(p4c_add_library)
 
 # Add files with the appropriate path to the list of linted files
-function(add_cpplint_files dir filelist)
```
davidbolvansky, line 52: Probably some warning comment is needed that these macros could be used by downstream projects so it is important to leave them as macros, etc..

davidbolvansky (approved):

CMakeLists.txt
davidbolvansky: And probably you do not need this if you use xargs -d ';'

cmake/P4CUtils.cmake
fruffy: Now that we are adding cpplint files to a global property this might actually not be necessary any longer.

davidbolvansky:
I will try it on our side again to ensure everything is fine.

davidbolvansky:
cpplint
Skipping input '': Can't open for reading

please add '-r' to xargs


Otherwise OK, LGTM.

PR #3764, opened 2022-12-07
Add missing semicolon at end of Bison rule in p4parser.ypp

jafingerhut:
This omission does not seem to have caused any functional problems.  I only noticed it while doing diffs between an updated version of the P4 language spec repo's grammar.mdk file and this one.


mbudiu-vmw:
I don't really understand how that's possible.

jfingerh:
Perhaps bison is forgiving of such things, i.e. it automatically infers the end of one rule and the beginning of another in at least some cases like this?  I do not know.

hesingh:
An Action has no termination semicolon - this why the code works. Adding a semicolon on a new line is terminating the rule and what this PR has done.

https://fog.misty.com/perry/osp/references/bison/by-section/bison_24.html

mbudiu-vmw (approved):

PR #3769, opened 2022-12-10
Remove backends from CMake include paths. 

### LINKED ISSUE #3768: [p4testgen] Use consistent header import paths
We use a different build system to build p4testgen (bazel). The inconsistent pattern of imports in `backends/p4tools` is causing problems for us. 

Particularly, we have problems with the following patterns:

1. E.g.  in `backends/p4tools/common/lib/taint.cpp` there is 

```
#include "backends/p4tools/common/lib/symbolic_env.h"
...
#include "p4tools/common/lib/model.h"
```

This pattern is repeated in many files in `backends/p4tools`. Would be great if we could replace all "#include "p4tools/..." with "#include "backends/p4toos/...".

2. in `backends/p4tools/modules/testgen/targets/bmv2/bmv2.cpp` there is `#include "bmv2/common/annotations.h`. Similarly, would be nice to replace it with `#include "backends/bmv2/common/annotations.h`. Fortunately, only this file has this problem.

@fruffy for visibility.

fruffy:
Fixes #3768. Make include paths more rigorous by removing ambiguity. Every backend include should contain the full path.

### DIFF
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -32,6 +32,6 @@ CheckOptions:
   - { key: readability-identifier-naming.GlobalConstantCase,  value: UPPER_CASE }
   - { key: readability-identifier-naming.StaticConstantCase,  value: UPPER_CASE }
   - { key: readability-identifier-naming.StaticVariableCase,  value: UPPER_CASE }
-...
 ExtraArgs:
-  - '--std=c++11'
+  - '--std=c++17'
+...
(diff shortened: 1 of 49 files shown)


kheradmandG (approved):

mbudiu-vmw (approved):

fruffy:
Thanks!

PR #3770, opened 2022-12-10
Run build and tests directly on Ubuntu 20.04, 22.04 without docker

davidbolvansky:


fruffy:
https://github.com/p4lang/third-party/blob/main/Dockerfile contains most of the dependencies. You will also have to install protobuf, thrift, grpc, nanomsg etc. I am not sure if the bmv2 package provides these. But with newer versions they can easily be installed using debian.

davidbolvansky:
> > https://github.com/p4lang/third-party/blob/main/Dockerfile contains most of the dependencies. You will also have to install protobuf, thrift, grpc, nanomsg etc. I am not sure if the bmv2 package provides these. But with newer versions they can easily be installed using debian.
> 
> probably they are installed as building works fine on Ubuntu 22.

but ptf is missing. So lets switch to ubuntu only for normal tests, dafault to existing base image.

fruffy:
> but ptf is missing. So lets switch to ubuntu only for normal tests, default to existing base image.

Isn't this install just pip3 install --user ptf?

fruffy:
I think it is also fine to just test the build for Ubuntu 18.04 and not run tests.

davidbolvansky:
> I think it is also fine to just test the build for Ubuntu 18.04 and not run tests since there are limited guarantees.

I was able to run them so it would be great to keep them for some time. If p4c project after April 2023 decides to drop this CI support for U18 with reasonable reasons, I am fine with that.

rst0git:
> You will also have to install protobuf, thrift, grpc, nanomsg etc. I am not sure if the bmv2 package provides these.

Yes. These packages are listed as [dependencies](https://github.com/rst0git/p4lang-packages/blob/482b5f274fa4e372951ca305a8eaf96a941138ce/p4lang-bmv2/control#L5) for the bmv2 package.

.github/workflows/ci-test.yml
```diff
@@ -12,13 +12,43 @@ concurrency:
...
+
+    - name: ccache
+      uses: hendrikmuhs/ccache-action@v1
+      with:
+        key: test-${{ runner.os }}-gcc
+        max-size: 1000M
+
+    - name: Build (Ubuntu 22.04, GCC)
+      run: |
+        docker build -t p4c --build-arg UBUNTU_VERSION=22.04 --build-arg IMAGE_TYPE=test --build-arg ENABLE_UNIFIED_COMPILATION=ON .
```
fruffy, line 36: Do you even need the docker image at this point?

tools/ci-build.sh
```diff
@@ -53,15 +63,24 @@ apt install -y --no-install-recommends \
...
+
+ccache --set-config cache_dir=/p4c/.ccache
+ccache --set-config max_size=1G
 
 # we want to use Python as the default so change the symlinks
 ln -sf /usr/bin/python3 /usr/bin/python
 ln -sf /usr/bin/pip3 /usr/bin/pip
 
+pip3 install --upgrade pip
 pip3 install wheel
```
fruffy, line 84: `wheel` might not be needed anymore.

ir/ir-inline.h
```diff
@@ -269,7 +269,7 @@ void IR::NameMap<T, MAP, COMP, ALLOC>::visit_children(Visitor& v) {
                 namemap_insert_helper(i, cstring(obj_name(s)), std::move(s), symbols, new_symbols);
                 i = symbols.erase(i);
             }
-        } else {
+        } else if (n) {
```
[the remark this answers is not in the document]
fruffy: What is the warning?

Dockerfile
```diff
@@ -1,4 +1,6 @@
-FROM p4lang/behavioral-model:latest
+ARG UBUNTU_VERSION=20.04
+ARG BASE_IMAGE=ubuntu:${UBUNTU_VERSION}
+FROM ${BASE_IMAGE}
```
rst0git, line 3: I think we need to keep `p4lang/behavioral-model:latest` as default base image.
This Dockerfile is used to automatically [build and publish](https://github.com/p4lang/p4c/blob/3ae0821aaff1b8eafb4caf4485cc602de2910c26/.github/workflows/ci-container-image.yml) the p4c container image in [Docker Hub](https://hub.docker.com/r/p4lang/p4c). This container image is being used in other projects.

fruffy: We could keep this docker file for the release build. With David's new setup we might not need a docker file at all for the p4c image.

.github/workflows/ci-test.yml
davidbolvansky: Well, good point, probably not.

ir/ir-inline.h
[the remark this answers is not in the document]
davidbolvansky: that n maybe be null pointer (not really, sanitizers would catch it)

.github/workflows/ci-test.yml
```diff
@@ -63,15 +102,44 @@ jobs:
...
+      uses: hendrikmuhs/ccache-action@v1
+      with:
+        key: test-ubuntu-18.04-gcc
+        max-size: 1000M
+
+    - name: Build (Ubuntu 18.04, GCC)
       run: |
-        docker build -t p4c --build-arg IMAGE_TYPE=test --build-arg ENABLE_UNIFIED_COMPILATION=ON --build-arg COMPILE_WITH_CLANG=ON \
-        --build-arg BUILD_AUTO_VAR_INIT_PATTERN=ON --build-arg ENABLE_SANITIZERS=ON .
+        docker build -t p4c --build-arg BASE_IMAGE=ubuntu:18.04 --build-arg IMAGE_TYPE=test --build-arg ENABLE_UNIFIED_COMPILATION=ON .
```
fruffy, line 137: This also needs to be updated to use Ubuntu 18.04 directly. Also other images such as the tools test should not use docker anymore then.

davidbolvansky: There is no Ubuntu 18.04 image for github actions. So nope, not possible.

davidbolvansky: >> Also other images such as the tools test should not use docker anymore then.

Sorry but for now I am not gonna do everything. We can move it incrementally.

fruffy: But we do not modify the base image right now so this will test on the Ubuntu 20.04 docker image as far as I can see.

fruffy: > > > Also other images such as the tools test should not use docker anymore then.
> 
> Sorry but for now I am not gonna do everything. We can move it incrementally.

The issue is that the new ci-build configuration seems to break the ccache that has been set up for the docker build process. As far as I can tell, the builds are taking longer now. We can check this by completing one build and rerunning to see whether the build is cached.

davidbolvansky: Build on Ubuntu 22 - 6m 14s.

Seems pretty OK to me.

davidbolvansky: Comparing it with for example https://github.com/p4lang/p4c/pull/3778 I see nothing problematic.

fruffy: The problem is the other tests, `static-build-test-p4c`,  `test-tools`,  `ptf-ebpf`, `validate` etc.

davidbolvansky: >> the builds are taking longer now. 

Sure?

test-p4c / test-ubuntu20 (OFF) (pull_request) Successful in 5m
test-p4c / test-ubuntu20 (ON) (pull_request) Successful in 37m

-------------------------------------------

https://github.com/p4lang/p4c/pull/3778

test-p4c / test-linux (OFF) (pull_request) Successful in 5m
test-p4c / test-linux (ON) (pull_request) Successful in 69m

fruffy: ptf-ebpf / ptf-linux (pull_request) Successful in 62m 

https://github.com/p4lang/p4c/pull/3778
 ptf-ebpf / ptf-linux (pull_request) Successful in 18m

fruffy: Yes, but this PR is breaking the ccache setup for all other tests. 

The reason `validate` is faster is that the tests are not running for some reason, that seems like an unrelated bug. 

In general, `test-tools` will now be the slowest test because the build alone takes 55 minutes.

davidbolvansky: >> The problem is the other tests, static-build-test-p4c, test-tools, ptf-ebpf, validate etc.

I can convert them but then we would stop testing with Dockerfile so it can slowly bitrot.

davidbolvansky: > Yes, but this PR is breaking the ccache setup for all other tests.

How? Could you explain it?

Docker builds uses export_ccache.sh which runs
docker cp $(docker create --rm p4c):/p4c/.ccache .
, non-docker runs 
sudo cp -rf /p4c/.ccache .

I pushed empty commit to see if regressions are real.

fruffy: Could be [this](https://github.com/p4lang/p4c/pull/3770/files#diff-76f06548e16be4f20ba423b3768e29bfca82a8ecd274407c02b37b3b5816f0c8L57) change.

fruffy: > > > The problem is the other tests, static-build-test-p4c, test-tools, ptf-ebpf, validate etc.
> 
> I can convert them but then we would stop testing with Dockerfile so it can slowly bitrot.

We can either convert them or have a workaround. As Radostin [mentioned](https://github.com/p4lang/p4c/pull/3770#discussion_r1045273096) we need to create one docker build either way.

davidbolvansky: >> In general, test-tools will now be the slowest test because the build alone takes 55 minutes.

ccache is working fine, rebuild time: 7m 25s

https://github.com/p4lang/p4c/actions/runs/3697344554/jobs/6262220551

davidbolvansky:
Based on new runs, I believe ccache is working fine, rebuild times are short.

davidbolvansky:
CI also significatly faster.

for example sanitizers 120min -> 71 min.

My assumption is that  CTEST_PARALLEL_LEVEL was not handled correctly (when you run something with sudo you lost env vars, you need to use -E switch)

davidbolvansky:
Ok, done

ir/ir-inline.h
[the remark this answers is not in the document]
fruffy: Instead of the if checks, we should use a BUG_CHECK here.

fruffy:
Looks good. The question is what we do with the docker build process that publishes builds to [dockerhub](https://hub.docker.com/r/p4lang/p4c/). Have a nightly test run?

davidbolvansky:
Or we can preserve testing with docker on U20. I dont know..

fruffy:
@rst0git any thoughts on this?

ir/ir-inline.h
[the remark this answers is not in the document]
fruffy: [inline.patch.txt](https://github.com/p4lang/p4c/files/10237706/inline.patch.txt)

How about this patch?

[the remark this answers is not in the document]
davidbolvansky: Feel free to commit it to my branch.

I was unable to apply this patch.

(Sad how our time is wasted due to false positive gcc warnings enabled by default :/)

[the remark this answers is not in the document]
fruffy: I do not think this is a false positive, n->node_type_name() could be interpreted as nullptr access.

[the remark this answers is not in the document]
davidbolvansky: But does not happen in practice as sanitizers would catch for sure.

rst0git:
> The question is what we do with the docker build process that publishes builds to dockerhub.

With the following change, when `BASE_IMAGE` is not specified, the container image should be the same.

```diff
-FROM p4lang/behavioral-model:latest
+ARG BASE_IMAGE=p4lang/behavioral-model:latest
+FROM ${BASE_IMAGE}
```
I think with these changes the [CI workflow](https://github.com/p4lang/p4c/blob/bd040325b95a0951d2c6c543abc0499458048e04/.github/workflows/ci-container-image.yml) updating the container image in Dockerhub should continue to work as expected.

ir/ir-inline.h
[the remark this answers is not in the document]
rst0git: Perhaps we should open a separate pull request with these changes?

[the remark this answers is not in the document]
davidbolvansky: I dont think we need another PR as this PR enables more testing, especially on Ubuntu 22, where we have newer GCC which emits more warnings. Seems OK to me to have such fixes in this PR.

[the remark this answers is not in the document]
davidbolvansky: Tests now fail - expected? 

Sorry but I would really prefer first version of these changes (my) as your changes seems to break existing tests and I am not going to debug them.

fruffy:
> > The question is what we do with the docker build process that publishes builds to dockerhub.
> 
> With the following change, when `BASE_IMAGE` is not specified, the container image should be the same.
> 
> ```diff
> -FROM p4lang/behavioral-model:latest
> +ARG BASE_IMAGE=p4lang/behavioral-model:latest
> +FROM ${BASE_IMAGE}
> ```
> 
> I think with these changes the [CI workflow](https://github.com/p4lang/p4c/blob/bd040325b95a0951d2c6c543abc0499458048e04/.github/workflows/ci-container-image.yml) updating the container image in Dockerhub should continue to work as expected.

It'd be nice if we ran tests on this release build. This way we can make sure the docker build does not rot.

fruffy:
@antoninbas @mbudiu-vmw

This PR would remove the docker third-party build from most of the tests. Any further concerns?

antoninbas:
no concern from me, but then I am not the one maintaining this

rst0git:
> It'd be nice if we ran tests on this release build. This way we can make sure the docker build does not rot.

If something with the container image ever breaks, we can introduce the necessary tests.

rst0git (approved):

mbudiu-vmw:
I am not maintaining this either

davidbolvansky:
> > It'd be nice if we ran tests on this release build. This way we can make sure the docker build does not rot.
> 
> If something with the container image ever breaks, we can introduce the necessary tests.

Yes, exactly.

davidbolvansky:
@fruffy is it ok for you?

rst0git:
> I am not maintaining this either

I would happy to be maintainer.

rst0git:
> I am not maintaining this either

I would be happy to be a maintainer.

PR #3774, opened 2022-12-12
Convert BUG macros to constexpr variadic functions for better typing.

fruffy-bfn:
WiP.


lib/exceptions.h
```diff
@@ -125,25 +125,29 @@ class CompilationError : public P4CExceptionBase {
...
-#define P4C_UNIMPLEMENTED(...)                                              \
-    do {                                                                    \
-        throw Util::CompilerUnimplemented(__LINE__, __FILE__, __VA_ARGS__); \
-    } while (0)
-
 }  // namespace Util
 
+template <class... Args>
+[[noreturn]] inline auto BUG(Args&&... args) {
+    throw Util::CompilerBug(__LINE__, __FILE__, std::forward<Args>(args)...);
```
vhavel, line 132: Wouldn't this alter the meaning of __LINE__ and __FILE__?

apinski-cavium: Yes and that is why C++20 adds std::source_location:
https://en.cppreference.com/w/cpp/utility/source_location

That is only there for C++20 and not in C++17 mode for GCC.

fruffy-bfn: Yes, this will not work with C++20 I believe. I was just getting annoyed with warnings about the BUG macros, so I wrote up this draft. Once we switch to C++20, this might be something worth considering. 

This new syntax is horrid, though.

PR #3775, opened 2022-12-12
Add assert and assume mode for testgen.

fruffy-bfn:
This PR adds assert and assume mode to P4Testgen. These modes can be enabled via the `--assumption-mode` and `--assertion-mode` flags.

- If only `--assumption-mode` is active, P4Testgen will only generate tests that satisfy all assert and assume conditions.
- If also `--assertion-mode` is active, P4Testgen will try to only produce tests that violate assertions also including assumptions in assume calls.
- If only `--assertion-mode` is active, P4Testgen will not add assumptions to the path and only generate tests that violate asserts.

### DIFF
--- a/backends/p4tools/modules/testgen/options.h
+++ b/backends/p4tools/modules/testgen/options.h
@@ -74,6 +74,14 @@ class TestgenOptions : public AbstractP4cToolOptions {
 
     /// Enforces the test generation of tests with mandatory output packet.
     bool withOutputPacket = false;
+    
+    /// Add conditions defined in assert/assume to the path conditions.
+    /// Only tests which satisfy these conditions can be generated.
+    bool enforceAssumptions = false;
+
+    /// Produce only tests that violate the condition defined in assert calls.
+    /// This will either produce no tests or only tests that contain counter examples.
+    bool assertionModeEnabled = false;
 
     const char *getIncludePath() override;
 
--- a/backends/p4tools/modules/testgen/options.h
+++ b/backends/p4tools/modules/testgen/options.h
@@ -74,7 +74,7 @@ class TestgenOptions : public AbstractP4cToolOptions {
 
     /// Enforces the test generation of tests with mandatory output packet.
     bool withOutputPacket = false;
-    
+
     /// Add conditions defined in assert/assume to the path conditions.
     /// Only tests which satisfy these conditions can be generated.
     bool enforceAssumptions = false;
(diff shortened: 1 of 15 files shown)


backends/p4tools/modules/testgen/options.h
```diff
@@ -65,6 +65,14 @@ class TestgenOptions : public AbstractP4cToolOptions {
     /// for statement reachability analysis.
     bool dcg = false;
 
+    /// Add conditions defined in assert/assume to the path conditions.
+    /// Only tests which satisfy these conditions can be generated.
+    bool enforceAssumptions = false;
```
jafingerhut, line 70: Do you plan on there being a way to only add conditions for `assume`, but not for `assert`?

I could imagine one might want to enforce `assume` conditions in tests created, but at the same time try to find tests that violate `assert` conditions.

Why?  `assume` would be conditions of "I am telling you I will be operating in an environment where these conditions are true, for reasons that are not necessarily possible to infer from the P4 source code, e.g. because I know I am operating in an environment where I will never see these protocol/NextHeader values.", but `assert` statements would be conditions of "I am hoping these conditions are always true, but want the tools to catch me if I am wrong in this hope."

fruffy-bfn: Yes, this is still in draft mode. Currently, we have two flags. 

- If only `enforceAssumptions` is active, P4Testgen will only generate tests that satisfy all assert and assume conditions.
- If also `genAssertionViolations` is active, P4Testgen will try to only produce tests that violate assertions also including assumptions in assume calls.
- If only  `genAssertionViolations` is active, P4Testgen will not add assumptions to the path and only generate tests that violate asserts. 

The tricky part is generating only violating tests while exploring the full program(i.e., not stopping at the first violating assert).

fruffy: This feature is now ready and can be co-developed with https://github.com/p4lang/p4-spec/issues/1207. Testgen can implement the specified functionality with assumption and assertion mode. The question is how to write tests here.

vhavel (commented):
I suggest we make `--assumption-mode` always active, or at least default.

Imagine a situation, where P4 programmer adds `assume()` calls to instruct a P4 compiler that only certain match key combinations are valid, and the P4 compiler uses this information to optimize, but generates "incorrect" code for control plane setup where the assumption doesn't hold. I think it would be surprising if testgen generated tests which trigger this "undefined behavior".

Or do you expect use cases where testcases which do not satisfy all assume conditions are wanted?

vhavel (approved):

PR #3776, opened 2022-12-14
Incorporate reachability in all possible exploration strategies

hannelita:
Reachability can lead to the scenario where we don't produce any tests (due to a bad pattern, for example). 
This PR lets the user know when no tests are produced. 
Generalized for all exploration strategies.


backends/p4tools/modules/testgen/core/exploration_strategy/exploration_strategy.cpp
```diff
@@ -70,8 +70,16 @@ uint64_t ExplorationStrategy::selectBranch(const std::vector<Branch>& branches)
     return Utils::getRandInt(branches.size() - 1);
 }
 
+void ExplorationStrategy::printTestGenerationStatus(bool initialTest) {
```
fruffy, line 73: Why make this an extra function and not just print a ::warning for each strategy?

hannelita: I'm trying to encapsulate the message logic in a function. At the moment it's indeed only about the dcg. I'd rather have a method call in each strategy than a hard-coded warning in each class.

fruffy: You could also check `testBackend`'s `testCount` in the `testgen.cpp` and emit a warning if it is 0. This should handle more than just one scenario.

hannelita: Right, but I'd also like to handle the case where the issue is particularly coming from the dcg/pattern.

fruffy: The main reason why I think it could be better to have this in the `testgen.cpp` is twofold. 
1. There is already quite a lot of logic in the exploration strategy classes and this would add one more check and flag. This is a core part of the interpreter and we should keep it simple.
2. Emitting a warning with 0 tests in the testgen classes catches all scenarios, not just this particular one. I think it can be generally useful to emit this warning.

hannelita: All right.
It's not clear to me where in TestBackend you'd like to have it. 
If we never reach any FinalState, the run method is not a candidate.

fruffy: So in `testgen.cpp`, after the try-catch, we can check whether the test count is 0. If it is 0, no callback has been executed, meaning no tests have been produced. Then we should emit a warning. We can introduce a getter for the `testCount` variable in the TestBackEnd class. I think this should achieve the same behavior.

PR #3783, opened 2022-12-15
Option to produce tests exclusively with output packets

hannelita:
Enforces the test generation of tests with an output packet.

### DIFF
--- a/backends/p4tools/modules/testgen/lib/test_backend.cpp
+++ b/backends/p4tools/modules/testgen/lib/test_backend.cpp
@@ -144,7 +144,7 @@ bool TestBackEnd::run(const FinalState& state) {
         // selectBranches, trackBranches and others continue to work.
         if (TestgenOptions::get().withOutputPacket && testCount > 0) {
             auto outputPacketSize = testInfo.outputPacket->type->width_bits();
-            if (outputPacketSize <= 0) {
+            if (outputPacketSize <= 0 || testInfo.packetIsDropped) {
                 return testCount > maxTests - 1;
             }
         }
--- a/backends/p4tools/modules/testgen/lib/test_backend.cpp
+++ b/backends/p4tools/modules/testgen/lib/test_backend.cpp
@@ -139,6 +139,13 @@ bool TestBackEnd::run(const FinalState& state) {
             return testCount > maxTests - 1;
         }
 
+        if (TestgenOptions::get().withOutputPacket && testCount > 0) {
+            auto outputPacketSize = testInfo.outputPacket->type->width_bits();
+            if (outputPacketSize <= 0) {
+                return testCount > maxTests - 1;
+            }
+        }
+
         const auto* testSpec = createTestSpec(executionState, completedModel, testInfo);
 
         float coverage = static_cast<float>(visitedStatements.size()) / allStatements.size();
--- a/backends/p4tools/modules/testgen/lib/test_backend.cpp
+++ b/backends/p4tools/modules/testgen/lib/test_backend.cpp
@@ -139,6 +139,9 @@ bool TestBackEnd::run(const FinalState& state) {
             return testCount > maxTests - 1;
         }
 
+        // Don't increase the test count if --with-output-packet is enabled and we don't
+        // produce a test with an output packet. Excludes the first test, so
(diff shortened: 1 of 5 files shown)


backends/p4tools/modules/testgen/lib/test_backend.cpp
```diff
@@ -139,6 +139,16 @@ bool TestBackEnd::run(const FinalState& state) {
             return testCount > maxTests - 1;
         }
 
+        // Don't increase the test count if --with-output-packet is enabled and we don't
+        // produce a test with an output packet. Excludes the first test, so
+        // selectBranches, trackBranches and others continue to work.
+        if (TestgenOptions::get().withOutputPacket && testCount > 0) {
```
fruffy, line 145: This is missing `testInfo.packetIsDropped`. 

Trying to understand, why is the first test necessary for trackBranches etc? What if that test does not have an output packet?

hannelita: I'm not sure if we're considering removing dropped packets or not. I decided to keep them.
If we have trackBranches and the `--with-output-packet` , I'm basically ignoring the logic and generating the test anyway. `--with-output-packet` will very likely be an accident if used with trackBranches, I believe

hannelita: All right, I can include the pkt dropped.
Tracking the options combination is something I won't address in this PR. 
Having an option to produce tests with an output pkt only was a feature request. Nothing to do with reachability/exploration strategies.

hannelita:
WIP - treat concolic case.

backends/p4tools/modules/testgen/lib/test_backend.cpp
fruffy: Okay, in that case I would move the check as early as possible in the test back end and check only for packet.dropped/0-width. This is the only case where we do not generate as second packet iirc. 

> Tracking the options combination is something I won't address in this PR.

But this is a bug, isn't it? We should handle this, otherwise it could be confusing. Similarly, this option will do the wrong thing if max-tests is set to 1. It does not seem like we need the `testCount > 0` check otherwise.

hannelita: So what are all the options that should ignore this flag?

fruffy: `--track-branches` seems like the only option from what I can see. If you enter a branch that generates output this will conflict with this option.

hannelita: It looks like those are assumptions on our end. We probably should address these options combinations and properly inform the user about these choices.

fruffy: I mixed up `--input-branches` and `--track-branches`, sorry about that. `--track-branches` is fine, `--input-branches`  needs a warning or error. 

`--maxTests 1` will work as intended if t`estCount > 0` is removed. So I do not think we need any further handling there.

backends/p4tools/modules/testgen/options.cpp
```diff
@@ -104,6 +104,20 @@ TestgenOptions::TestgenOptions()
         "Track the branches that are chosen in the symbolic executor. This can be used for "
         "deterministic replay.");
 
+    registerOption(
+        "--with-output-packet", nullptr,
+        [this](const char*) {
+            withOutputPacket = true;
+            if (trackBranches) {
```
fruffy, line 111:
```suggestion
            if (!selectedBranches.empty()) {
```
My mistake.

backends/p4tools/modules/testgen/options.cpp
```diff
@@ -104,6 +104,20 @@ TestgenOptions::TestgenOptions()
...
+                std::cerr << "--track-branches cannot guarantee --with-output-packet."
```
fruffy, line 112:
```suggestion
                std::cerr << "--input-branches can conflict with --with-output-packet."
```

fruffy (approved):
LGTM after suggestions are applied. One heads up. This could loop quite a while with large programs that have only dropped packets. But there is little we can do about that.

fruffy (approved):

PR #3785, opened 2022-12-16
Move timer from testgen to lib folder.

fruffy:
This moves the ScopedTimer class of P4Tools into the general lib folder. A scoped timer measures and logs the lifetime of the timer object from creation to destruction. This is a handy class as it only needs to be instantiated to record the time some functions take.


davidbolvansky (approved):
Useful, thanks

fruffy:
@mbudiu-vmw  Could you give this an approval/review?

mbudiu-vmw (approved):

PR #3786, opened 2022-12-17
Fix container build

davidbolvansky:


tools/ci-build.sh
```diff
@@ -95,15 +93,18 @@ P4C_PIP_PACKAGES="ipaddr \
...
 
+apt update
+apt install -y curl gnupg
+
 if [[ "${DISTRIB_RELEASE}" == "18.04" ]] ; then
   P4C_DEPS+=" libprotobuf-dev protobuf-compiler"
 else
   echo "deb http://download.opensuse.org/repositories/home:/p4lang/xUbuntu_${DISTRIB_RELEASE}/ /" | tee /etc/apt/sources.list.d/home:p4lang.list
   curl -L "http://download.opensuse.org/repositories/home:/p4lang/xUbuntu_${DISTRIB_RELEASE}/Release.key" | apt-key add -
+  apt update
```
rst0git, line 104: we only need to install `curl` `gnupg` in the "else" branch:
```bash
apt-get update && apt-get install -y curl gnupg
echo "deb http://download.opensuse.org/repositories/home:/p4lang/xUbuntu_${DISTRIB_RELEASE}/ /" | tee /etc/apt/sources.list.d/home:p4lang.list
curl -L "http://download.opensuse.org/repositories/home:/p4lang/xUbuntu_${DISTRIB_RELEASE}/Release.key" | apt-key add -
apt-get update
```

tools/ci-build.sh
```diff
@@ -95,15 +93,18 @@ P4C_PIP_PACKAGES="ipaddr \
...
   P4C_DEPS+=" libprotobuf-dev protobuf-compiler"
 else
+  apt install -y curl gnupg
   echo "deb http://download.opensuse.org/repositories/home:/p4lang/xUbuntu_${DISTRIB_RELEASE}/ /" | tee /etc/apt/sources.list.d/home:p4lang.list
   curl -L "http://download.opensuse.org/repositories/home:/p4lang/xUbuntu_${DISTRIB_RELEASE}/Release.key" | apt-key add -
+  apt update
   P4C_DEPS+=" p4lang-bmv2"
 fi
 
-apt update
```
rst0git, line 106: I think we can keep a single `apt-get update` here as it will be executed after the "if" condition above.

tools/ci-build.sh
```diff
@@ -94,17 +92,18 @@ P4C_PIP_PACKAGES="ipaddr \
                   scapy==2.4.5 \
                   clang-format>=15.0.4"
 
-
 if [[ "${DISTRIB_RELEASE}" == "18.04" ]] ; then
```
rst0git, line 95:
```Suggestion
if [[ "${DISTRIB_RELEASE}" == "18.04" ]] || [[ "$(which simple_switch 2> /dev/null)" != "" ]] ; then
```
With this change, the `p4lang-bmv2` package would be installed only if BMv2 is not already installed.
For  example, BMv2 should already be available in the `p4lang/behavioral-model:latest` container image.

tools/ci-build.sh
```diff
@@ -94,17 +92,18 @@ P4C_PIP_PACKAGES="ipaddr \
...
 
-
-if [[ "${DISTRIB_RELEASE}" == "18.04" ]] ; then
+if [[ "${DISTRIB_RELEASE}" == "18.04" ]] || [[ "$(which simple_switch 2> /dev/null)" != "" ]] ; then
   P4C_DEPS+=" libprotobuf-dev protobuf-compiler"
 else
+  apt-get update && apt-get install -y curl gnupg
   echo "deb http://download.opensuse.org/repositories/home:/p4lang/xUbuntu_${DISTRIB_RELEASE}/ /" | tee /etc/apt/sources.list.d/home:p4lang.list
   curl -L "http://download.opensuse.org/repositories/home:/p4lang/xUbuntu_${DISTRIB_RELEASE}/Release.key" | apt-key add -
+  apt-get update
```
rst0git, line 101:
```Suggestion
```
` apt-get update` is no longer necessary here because there is one just after the "if" statement.

rst0git (approved):
LGTM

I tested the changes from this pull request in my fork to confirm that the container image build & push works as expected:
- https://github.com/rst0git/p4c/actions/runs/3724144571/jobs/6316110809
- https://hub.docker.com/r/rstoyanov/p4c/tags
- `docker run -it rstoyanov/p4c bash`

davidbolvansky:
Thanks!

PR #3788, opened 2022-12-18
Build (only) container images for new PRs

davidbolvansky:
Detect problems sooner than after commit.

### DIFF
--- a/.github/workflows/ci-container-image.yml
+++ b/.github/workflows/ci-container-image.yml
@@ -1,9 +1,10 @@
-name: Build and push latest image to container registry
+name: container-image
 
 on:
   push:
-    branches:
-      - main
+    branches: [ main ]
+  pull_request:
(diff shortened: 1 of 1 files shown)


.github/workflows/ci-container-image.yml
```diff
@@ -1,9 +1,10 @@
-name: Build and push latest image to container registry
+name: container-image
 
 on:
   push:
-    branches:
-      - main
+    branches: [ main ]
+  pull_request:
+    branches: [ main ]
```
fruffy, line 7: Trying to understand this? Doesn't this run on every pull request to main then?

davidbolvansky: This is goal of this patch, to run it tor every new PR.

Currently container build runs only when patch is commited to main. (or via schedule)

fruffy (approved):

PR #3790, opened 2022-12-19
ebpf/PSA: improve support for wide fields

tatry:
This PR improves support for fields wider than 64 bits. This is done mostly by fixing bugs and some refactors.

The assumption for fields wider than 64 bits is that byte order of them is network byte order, which means that byte order is preserved during parser or deparser.

To achieve initial support for wide fields, the following changes has been made:

- Fixed byte order of bytes emitted in deparser. Parser had proper implementation of this.
- Allow to assign one field to another.
- Improved code for generating table key during lookup. `exact`, `lpm` or `ternary` match type can used.
- Improved generation of table const entries (now mostly using Visitors).
- Improved prefix length calculation for LPM table.

The current PR allows to parse, deparse, match in tables and assign wide fields. Subsequent PR(s) will extend wide fields support for externs.

CC: @osinstom

### DIFF
--- a/backends/ebpf/ebpfTable.cpp
+++ b/backends/ebpf/ebpfTable.cpp
@@ -478,9 +478,14 @@ void EBPFTable::emitKey(CodeBuilder* builder, cstring keyName) {
             } else if (width <= 64) {
                 swap = "bpf_htonll";
             } else {
-                // TODO: handle width > 64 bits
-                ::error(ErrorType::ERR_UNSUPPORTED,
-                        "%1%: fields wider than 64 bits are not supported yet", fieldName);
+                // The code works with fields wider than 64 bits for PSA architecture. It is shred
+                // with filter model, so should works but has not been tested. Error message is
+                // preserved for filter model because existing tests expect it.
+                // TODO: handle width > 64 bits for filter model
+                if (program->options.arch.isNullOrEmpty() || program->options.arch == "filter") {
+                    ::error(ErrorType::ERR_UNSUPPORTED,
+                            "%1%: fields wider than 64 bits are not supported yet", fieldName);
+                }
             }
         }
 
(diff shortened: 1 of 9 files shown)


backends/ebpf/ebpfTable.cpp
```diff
@@ -478,9 +478,14 @@ void EBPFTable::emitKey(CodeBuilder* builder, cstring keyName) {
             } else if (width <= 64) {
                 swap = "bpf_htonll";
             } else {
-                // TODO: handle width > 64 bits
-                ::error(ErrorType::ERR_UNSUPPORTED,
-                        "%1%: fields wider than 64 bits are not supported yet", fieldName);
+                // The code works with fields wider than 64 bits for PSA architecture. It is shred
```
mbudiu-vmw, line 481: some typos in this comment.

backends/ebpf/ebpfTable.cpp
```diff
@@ -478,9 +478,14 @@ void EBPFTable::emitKey(CodeBuilder* builder, cstring keyName) {
...
+                // with filter model, so should works but has not been tested. Error message is
```
mbudiu-vmw, line 482: since you are doing this you should also consider inserting the midend pass saturationElim which will enable saturated arithmetic for ebpf.

mbudiu-vmw (approved):

backends/ebpf/ebpfTable.cpp
tatry: I'm not sure if the midend pass saturationElim is related to this PR. In this PR I enable basic operations on wide fields (wider than 64 bits), like assignment, but I didn't touch the arithmetic operations.

tatry:
I've force-pushed branch in hope to fix bazel checks

PR #3791, opened 2022-12-19
Remove permissions from nonexecutable files.

fruffy:


mbudiu-vmw (approved):

fruffy:
https://github.com/bazelbuild/bazel/issues/8622 breaks Bazel builds. I tried a couple things but it looks like this comes from upstream dependencies. Not sure how to fix yet.

@smolkaj any ideas?

smolkaj:
Thanks for the ping, we had the same issue in several other repos.
The regression is due to the release of Bazel 6, which made some breaking changes.
As a workaround, we can fix Bazel to version < 5 by adding a `.bazelversion` file, see https://github.com/bazelbuild/bazel/issues/8622#issuecomment-1358123020

PR #3792, opened 2022-12-19
Refactor ci-build.sh script.

fruffy:
Refactor the `ci-build` script.
- Remove unnecessary `.ccache` copy calls from Github action scripts.
- Only use sudo calls in `ci-build` script where necessary. Gets the script one step closer to be run as a normal dependency install script.
- Add a check for building with Docker in `ci-build` script.

### DIFF
--- a/tools/ci-build.sh
+++ b/tools/ci-build.sh
@@ -5,6 +5,9 @@
 set -e  # Exit on error.
 set -x  # Make command execution verbose
 
+THIS_DIR=$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )
+P4C_DIR=$(readlink -f ${THIS_DIR}/..)
+
 # Default to using 2 make jobs, which is a good default for CI. If you're
 # building locally or you know there are more cores available, you may want to
 # override this.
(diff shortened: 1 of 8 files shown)


davidbolvansky:
You can wrap these three lines with new script. This new approach (which I am not sure if does not break ccache at all) looks like a downgrade to me, ci-build.sh becomes uselessly more complex.

Plus please preserve apt-get. I am not sure why you reverted these changes. apt-get should be used in the scripts, not apt.

tools/ci-build.sh
```diff
@@ -5,6 +5,9 @@
 set -e  # Exit on error.
 set -x  # Make command execution verbose
 
+THIS_DIR=$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )
+P4C_DIR=$(readlink -f ${THIS_DIR}/..)
```
davidbolvansky, line 9: readlink -f $0/..

rst0git: IIRC `readlink -f` is not available in macos

fruffy: I think the other version is more readable, any reason why `$0`?

davidbolvansky: $0 is part of the POSIX shell specification, whereas BASH_SOURCE, as the name suggests, is Bash-specific.

davidbolvansky (approved):

PR #3797, opened 2022-12-20
Use https with opensuse repositories

rst0git:
This pull request updates the links to download.opensuse.org to use https instead of http.


davidbolvansky (approved):
LGTM

Issue #3612, opened 2022-10-24
switch allowing duplicated cases for switch on bit<4>


apinski-cavium:
Take:
```
#include <core.p4>

control c(in bit<4> a)
{
  apply {
    switch (a) {
      4w0: {}
      4w1: {} // { dg-note "" }
      4w1: {} // { dg-error "" }
    }
  }
}

control ct(in bit<4> a);
package pt(ct t);
pt(c()) main;
```
This should fail as 4w1 is repeated.
The langauge spec is clear here even:
```
12.7.3. Notes common to all switch statements
It is a compile-time error if two labels of a switch statement equal each other. 
```

PR #3799, opened 2022-12-22
Fix duplicate switch label for general switch statement

usha1830:
Fixes https://github.com/p4lang/p4c/issues/3612

### DIFF
--- a/frontends/p4/typeChecking/typeChecker.cpp
+++ b/frontends/p4/typeChecking/typeChecker.cpp
@@ -3782,9 +3782,17 @@ const IR::Node* TypeInference::postorder(IR::SwitchStatement* stat) {
         if (isCompileTimeConstant(stat->expression))
             warn(ErrorType::WARN_MISMATCH, "%1%: constant expression in switch", stat->expression);
 
+        std::map<big_int, const IR::Node*> caselabels;
         for (auto& c : stat->cases) {
             if (!isCompileTimeConstant(c->label))
                 typeError("%1%: must be a compile-time constant", c->label);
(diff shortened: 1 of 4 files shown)


frontends/p4/typeChecking/typeChecker.cpp
```diff
@@ -3782,9 +3782,17 @@ const IR::Node* TypeInference::postorder(IR::SwitchStatement* stat) {
         if (isCompileTimeConstant(stat->expression))
             warn(ErrorType::WARN_MISMATCH, "%1%: constant expression in switch", stat->expression);
 
+        std::map<big_int, const IR::Node*> caselabels;
```
mbudiu-vmw, line 3785: The type may not be integer: https://p4.org/p4-spec/docs/P4-16-v-1.2.3.html#sec-switch-statement-with-integer-or-enumerated-type-expression

usha1830: I see. I will check and update. Thanks for pointing out!

mbudiu-vmw: First, I would move this out of typeChecking, it needs to be done only once. Especially since this is a quadratic algorithm and the number of labels can be very large. 
The simplest way would probably be to use the constantFolding pass to generate equality expressions label1 == label2 and see whether any of them evaluates to true.

usha1830: Thanks for the pointers, Mihai. I will explore the constantFolding pass and other possible approaches for detecting this.

PR #3803, opened 2022-12-23
Introduce stack initializers

### LINKED ISSUE #3802: Compiler bug with cast of double type'd type
Take:
```
type bit<32> N32;
type N32 NN32;
void f()
{
  NN32 ttt = (NN32)1;
}
```
Produces:
```
In file: /home/apinski/src/p4/p4c/frontends/p4/typeChecking/typeChecker.cpp:158
```

mbudiu-vmw:

This also fixes two other bugs:
- invalid headers didn't work properly with generic header types: `(H<bit<32>>){#}`
- Fixes #3802


mbudiu-vmw:
This is experimental, and should be discussed at the next LDWG.

mbudiu-vmw:
Here is the example test case:
```
header H<T> {
    bit<32> b;
    T t;
}

control c(out bit<32> r) {
    apply {
        H<bit<32>>[3] s;
        s = (H<bit<32>>[3]){ {0, 1}, {2, 3}, (H<bit<32>>){#} };
        r = s[0].b + s[0].t + s[1].b + s[1].t + s[2].b;
    }
}
```

mbudiu-vmw:
You always need to specify all elements of the stack, otherwise the expression is rejected.

testdata/p4_16_samples_outputs/array-copy-bmv2-midend.p4
```diff
@@ -18,7 +18,8 @@ parser P(packet_in p, out Headers h, inout Meta m, inout standard_metadata_t sm)
     state start {
         p.extract<Hdr>(h.h1.next);
         p.extract<Hdr>(h.h1.next);
-        h.h2 = h.h1;
+        h.h2[0] = h.h1[0];
+        h.h2[1] = h.h1[1];
```
jafingerhut, line 22: There are cases where this kind of replacement is incorrect, because it does not preserve the nextIndex value of the header stack when one is assigned to the other.  We have at least one test case for that.  Is this incomplete/incorrect replacement being done in that case where it is incorrect, and caught as a failing test?  That test name is subparser-with-header-stack-bmv2.p4

jafingerhut: Perhaps your compiler changes are somehow checking, before making such a replacement, that the nextIndex value of the header stack on the LHS of the assignment statement is never used later?

mbudiu-vmw: The pass already existed, I just enabled it. Assignment copies this invisible field?

jafingerhut: Assignment between header stacks is defined to copy the nextIndex field from one header stack to another, yes.

I checked, and this PR has no changes to the expected output files for test program subparser-with-header-stack-bmv2.p4, so perhaps these code changes are somehow only making the replacement when it is correct?  I do not know.

jafingerhut:
What value of nextIndex is assigned to the left-hand side when the right-hand side is one of these new header stack expressions?  Always the max value?

mbudiu-vmw:
@fruffy this crashes the testgen backend.

fruffy:
> @fruffy this crashes the testgen backend.

The PTF-eBPF tests also fail, it seems this pass some side-effect. Let me check for tools.

mbudiu-vmw:
ptf-ebpf tests have been failing for weeks

fruffy:
It might be caused  by the libbpf update. The tests pass on other PRs.

fruffy:
There seems to be a bug in the CopyStructures pass, it does not unroll the stack assignment any more.

fruffy:
I see, just saw the comment and the corresponding issue. This is not easy to fix, we can add that test to Xfails then.

backends/ebpf/targets/xdp_target.py
```diff
@@ -0,0 +1 @@
+/home/mbudiu/git/p4c/extensions/p4c-xdp/xdp_target.py
```
rst0git, line 1: It appears that this has been added accidentally.

mbudiu-vmw:
@fruffy I guess I didn't put the xfails in the right place... where should the other 2 be?

fruffy (approved):

mbudiu-vmw:
Fixes https://github.com/p4lang/p4c/issues/3836

rst0git (approved):
LGTM

PR #3804, opened 2022-12-23
Experimental: support for trailing commas

mbudiu-vmw:
Fixes https://github.com/p4lang/p4-spec/issues/1205


mbudiu-vmw:
To be discussed in the LDWG.
Notice that this is a grammar-only change, so it should not affect any backend.

jafingerhut (approved):
LGTM, including the grammar changes.  I did not think carefully about whether there are other grammar rules that were not changed, that could be similarly modified.

mbudiu-vmw:
I will merge this change, since the LDWG seems to have been in approval.

PR #3794, opened 2022-12-19
Constant-fold isValid calls when possible

mbudiu-vmw:
Signed-off-by: Mihai Budiu <mbudiu@vmware.com>
Fixes #3779

PR #3805, opened 2022-12-23
Simplify constant-fold method

rst0git:
This pull request aims to follow up on https://github.com/p4lang/p4c/pull/3794 to simplify `DoConstantFolding::postorder(IR::Member* e)` by removing unnecessary if-else statements.

### DIFF
--- a/frontends/common/constantFolding.cpp
+++ b/frontends/common/constantFolding.cpp
@@ -623,42 +623,45 @@ const IR::Node* DoConstantFolding::postorder(IR::Member* e) {
     auto type = typeMap->getType(orig->expr, true);
     auto origtype = typeMap->getType(orig);
 
-    const IR::Expression* result = e;
     if (type->is<IR::Type_Stack>() && e->member == IR::Type_Stack::arraySize) {
         auto st = type->to<IR::Type_Stack>();
         auto size = st->getSize();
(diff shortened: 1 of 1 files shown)


frontends/common/constantFolding.cpp
```diff
@@ -623,42 +623,45 @@ const IR::Node* DoConstantFolding::postorder(IR::Member* e) {
...
-            return e;
-        } else {
-            BUG("Unexpected initializer: %1%", expr);
+    auto expr = getConstant(e->expr);
+    if (expr == nullptr) return e;
+
+    auto structType = type->to<IR::Type_StructLike>();
+    if (structType == nullptr) BUG("Expected a struct type, got %1%", type);
+
+    if (auto list = expr->to<IR::ListExpression>()) {
```
mbudiu-vmw, line 638: I wonder whether this is still necessary; the type checker will convert list expressions to structexpressions when needed.

mbudiu-vmw (approved):

frontends/common/constantFolding.cpp
rst0git: Thank you Mihai for the code review! To avoid the risk of breaking something, I would leave it as is for the time being.

PR #3809, opened 2022-12-23
Use big_int for match_t so can match wider that 64 bits

ChrisDodd:


mbudiu-vmw (approved):

mbudiu-vmw:
You need to run clang-format

ChrisDodd:
> You need to run clang-format

Ugh, looks like it is now requiring references to be mis-formatted?  We should fix that.

PR #3816, opened 2023-01-02
Fix for STF tests with BMv2 when Scapy 2.5.0 is installed

jafingerhut:


jafingerhut:
I was running the p4c tests on an Ubuntu 20.04 system today.  Last week when I tried this, Scapy 2.4.x was the version installed, but today it was Scapy 2.5.0, which seems like it may have removed the method `convert_to`.  Whatever the reason, attempting to call `convert_to(Raw)` on many packets in STF tests was causing an exception to be thrown.  The tests pass with the proposed changes in this PR, and I believe should work fine with older versions of Scapy on Python3, too.

mbudiu-vmw (approved):

PR #3817, opened 2023-01-02
initialize direction metadata at begining of pipeline

kamleshbhalui:
Read direction only once in a pipeline and input port and direction is same throughout


usha1830 (approved):
LGTM

Issue #3772, opened 2022-12-11
p4c "-DT=1 + 1" is broken


apinski-cavium:
```
apinski@xeond:~/src/p4/octeontxkpu$ ../p4c/build/p4c t6.p4 -D"T=1 + 1"   -v
running cc -E -DT=1 + 1 -C -undef -nostdinc -x assembler-with-cpp -I /bajas/pinskia/src/p4/p4c/build/p4include -o './t6.p4i' 't6.p4'
cc: error: +: No such file or directory
cc: error: 1: No such file or directory
```
Looks like quoting is missed ...

davidbolvansky:
https://github.com/p4lang/p4c/blob/main/tools/driver/p4c_src/driver.py#L120

mbudiu-vmw:
This is a Python bug. Anyone wants to tackle it?

apinski-cavium:
No this is not a bug in the python code but rather it is a bug in common/parser_options.cpp dealing with spaces.
The way cpp/cc is invoked is using popen and you just pass the string directly without re-encoding the spaces and such.
The same will happen if you have an include directory with a space in it.
Or with use -MF with a file with a space.
Most likely you should use posix_spawnp (https://pubs.opengroup.org/onlinepubs/9699919799/functions/posix_spawn.html) instead of popen. And have an array of arguments instead of popen doing the tokenizing.

PR #3820, opened 2023-01-03
Handle spaces in preprocessor options

rst0git:
This pull request encloses the value specified with the `-D` option into quotes to allow the use of white-spaces.

Fixes: #3772


apinski-cavium:
This does not fix: -I"a b" or even '-Dhelp="; breakme"' .

PR #3821, opened 2023-01-04
Fix IndexedVector methods and constructor

vlstill:
This fixes `IndexedVector::pop_back` and a constructor from `safe_vector`. These do not compile if instantiated, therefore it is not possible to call `pop_back` on `IndexedVector`. This is required for a bugfix in the Intel Tofino P4 compiler.


mbudiu-vmw (approved):

mbudiu-vmw:
How about a test using these methods? Probably in gtest.

vlstill:
I've added some basic tests (mostly covering what I have fixed) and it uncovered one more bug in the original code which I have also fixed.

PR #3825, opened 2023-01-06
Make ports an int32 type for the Protobuf BMv2 back end.

fruffy:
Make the protobuf format consistent with the output. It is strange that protoc has not complained. Likely because all output port values are single digits which do not require formatting.

@kheradmandG In general, ports are a 9-bit type but they are represented as int32. If they ever expand to larger bit-widths we can change the format.

### DIFF
--- a/backends/p4tools/modules/testgen/targets/bmv2/backend/protobuf/protobuf.cpp
+++ b/backends/p4tools/modules/testgen/targets/bmv2/backend/protobuf/protobuf.cpp
@@ -293,13 +293,13 @@ traces: "{{trace_item}}"
 
 input_packet {
   packet: "{{send.pkt}}"
-  port: "{{send.ig_port}}"
+  port: {{send.ig_port}}
 }
 
 ## if verify
 expected_output_packet {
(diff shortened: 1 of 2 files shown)


backends/p4tools/modules/testgen/targets/bmv2/backend/protobuf/protobuf.cpp
```diff
@@ -293,13 +293,13 @@ traces: "{{trace_item}}"
 
 input_packet {
   packet: "{{send.pkt}}"
-  port: "{{send.ig_port}}"
+  port: {{send.ig_port}}
```
mbudiu-vmw, line 296: I don't know protobuf, but this change seems different.

mbudiu-vmw (approved):

backends/p4tools/modules/testgen/targets/bmv2/backend/protobuf/protobuf.cpp
fruffy: The protobuf spec now expects an integer instead of a string, this is why this change was necessary.

PR #3826, opened 2023-01-06
Simplify bazel cache key.

fruffy:
The Bazel CI build quickly fills up the available cache because it creates an entry for any single commit. However, the cache is loaded from the latest entry. Maybe we only need two keys?

We simplify follow the guide here: https://github.com/actions/cache/blob/main/examples.md#---bazel

The caches already work on a branch-basis as documented here:  https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows#matching-a-cache-key

### DIFF
--- a/backends/p4fmt/CMakeLists.txt
+++ b/backends/p4fmt/CMakeLists.txt
@@ -3,6 +3,7 @@ set(FMT_SRCS
     options.cpp
     main.cpp
     p4formatter.cpp
+    attach.cpp
 )
 
 set(REFCHECK_SRCS
@@ -10,6 +11,7 @@ set(REFCHECK_SRCS
     options.cpp
     p4fmt.cpp
     p4formatter.cpp
+    attach.cpp
 )
 
 # p4fmt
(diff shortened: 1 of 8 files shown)


smolkaj:
Cache keys are immutable. So with this strategy, our main branch cache entry will be created once and then go more and more stale over time.

fruffy:
> Cache keys are immutable. So with this strategy, our main branch cache entry will be created once and then go more and more stale over time.

Are you sure? That would be surprising behavior considering this is the recommended way. There is an option to always save:
https://github.com/actions/cache/tree/main/save#always-save-cache

fruffy:
We could adopt the strategy gRPC uses: https://github.com/grpc/grpc/blob/master/.github/workflows/pr-auto-fix.yaml#L24

smolkaj:
> Are you sure? That would be surprising behavior considering this is the recommended way.

I know, right? :)

You can confirm by looking at the test run of this PR: 
image

> There is an option to always save:
https://github.com/actions/cache/tree/main/save#always-save-cache

Overriding cache keys with the save action may work, you could give that a try.

Having experimented with this quite a bit, you might want to take a look at https://github.com/pins/pins-infra/blob/master/.github/workflows/ci.yml for a strategy that has stood the test of time.

PR #3829, opened 2023-01-08
Add stop metrics for P4Testgen. Sanitize some P4Testgen options.

fruffy:
- Introduce a stop metric for P4Testgen en. If a stop metric argument is provided, P4Testgen will stop producing test if that particular metric is satisfied. 
- max-tests inputs with values less than 0 will cause P4Testgen to exhaustively generate tests.
- Sanitize inputs for various options and make sure to reject invalid inputs.

@hannelita I also modified the input valuation and arguments for exploration strategies since the options behave similar to stop metrics. Happy to hear suggestions on how to improve this.

### DIFF
--- a/backends/p4tools/modules/testgen/lib/test_backend.h
+++ b/backends/p4tools/modules/testgen/lib/test_backend.h
@@ -23,6 +23,10 @@ namespace P4Tools {
 namespace P4Testgen {
 
 class TestBackEnd {
+ private:
+    /// The current test count. If it exceeds @var maxTests, the symbolic executor will stop.
+    uint64_t testCount = 0;
+
  protected:
     /// ProgramInfo is used to access some target specific information for test generation.
     const ProgramInfo& programInfo;
@@ -35,10 +39,7 @@ class TestBackEnd {
     ExplorationStrategy& symbex;
 
     /// Test maximum number of tests that are to be produced.
-    int maxTests;
-
-    /// The current test count. If it exceeds @var maxTests, the symbolic executor will stop.
-    int testCount = 0;
+    uint64_t maxTests;
 
     explicit TestBackEnd(const ProgramInfo& programInfo, ExplorationStrategy& symbex)
         : programInfo(programInfo), symbex(symbex), maxTests(TestgenOptions::get().maxTests) {
@@ -47,6 +48,8 @@ class TestBackEnd {
             maxTests = 1;
         }
     }
+
+    [[nodiscard]] bool needsToTerminate(uint64_t testCount) const { return testCount == maxTests; }
 
  public:
     TestBackEnd(const TestBackEnd&) = default;
@@ -92,7 +95,7 @@ class TestBackEnd {
     /// @returns false if the test generation is to be aborted (for example when the port is
     /// tainted.)
     virtual bool printTestInfo(const ExecutionState* executionState, const TestInfo& testInfo,
-                               int testCount, const IR::Expression* outputPortExpr);
+                               const IR::Expression* outputPortExpr);
 
     /// @returns a new modules with all concolic variables in the program resolved.
     const Model* computeConcolicVariables(const ExecutionState* executionState,
@@ -114,7 +117,7 @@ class TestBackEnd {
     static void printPerformanceReport();
 
     /// Accessors.
-    int getTestCount();
+    [[nodiscard]] uint64_t getTestCount() const;
 };
 
 }  // namespace P4Testgen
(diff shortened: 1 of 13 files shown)


backends/p4tools/modules/testgen/options.cpp
```diff
@@ -36,32 +41,79 @@ TestgenOptions::TestgenOptions()
...
+                maxTests = std::stoull(arg);
+            } catch (std::invalid_argument&) {
+                ::error("Invalid input value %1% for --max-tests. Expected integer.", arg);
+                return false;
+            }
             return true;
         },
-        "Sets the maximum number of tests to be generated");
+        "Sets the maximum number of tests to be generated [default: 1]. Setting the value to <=0 "
+        "will generate tests until no more paths can be found.");
```
hannelita, line 58: I am not sure if a negative number of tests makes sense. It's a bit counterintuitive, I think. It could be a mistake. If we want to exhaustively generate tests, I think it should be a specific stop metric, along with `MAX_STATEMENT_COVERAGE`.

More generally, here's a suggestion: every time the user inputs an arg that does not make much sense, I believe we should default it to something that won't take too long to terminate and produce some tests.

backends/p4tools/modules/testgen/options.cpp
```diff
@@ -121,27 +173,88 @@ TestgenOptions::TestgenOptions()
...
+    registerOption(
+        "--pop-level", "popLevel",
+        [this](const char* arg) {
+            int64_t popLevelTmp = 0;
+            try {
+                // Unfortunately, we can not use std::stoul because negative inputs are okay
+                // according to the C++ standard.
+                popLevelTmp = std::stoll(arg);
+                if (popLevelTmp < 0) {
+                    throw std::invalid_argument("Negative input.");
```
hannelita, line 201: If we're throwing invalid arg here, would it make sense to use the right ranges already? i.e. `popLevel` should not be negative, but it also needs to be greater than 1. 

Maybe an elegant improvement would be creating a function in `Utils`, i.e. `Utils::positiveGreaterThan(BOUND)` ? Then it could be reused in `popLevel`, `linearEnumeration` and `saddlePoint`.

hannelita:
We do have this README `https://github.com/p4lang/p4c/blob/main/backends/p4tools/modules/testgen/README.md` and it looks like it's outdated (it'll be a lot more outdated after these changes). Shall we update it, or erase the contents of it and suggest checking instructions in the command line?

Also, I believe we should have a full command suggestion in the README.

backends/p4tools/modules/testgen/options.cpp
fruffy: Fixed. Regarding the second point, I considered this but we still have to catch the exception and use the appropriate type so the gain is relatively small.

backends/p4tools/modules/testgen/options.cpp
fruffy: I made it such that an error is thrown if the input is negative. I still think having 0 as value to generate infinite tests is an intuitive input since it is used in a fair number of applications. It can also be easily combined with other stop metrics.

fruffy:
> Also, I believe we should have a full command suggestion in the README.

Do you mean the output of  `./p4testgen --help`? I updated it.

backends/p4tools/modules/testgen/options.cpp
hannelita: The only issue I see with that is non-termination. I'm not sure if it's explicit to the user it can happen. That's why I was suggesting enforcing a bounded model; if not, then use an explicit flag.

fruffy: In this case, non-termination is an explicit user choice (0 tests, which is not a default) and the stop metric is exhaustion of all available branches. P4Testgen will generate all the unique tests for that particular program, then stop. 

The only case where non-termination is possible is a unconditional parser-loop. There the program graph is not a DAG anymore. P4Testgen currently does not handle those correctly. Ideally, we would bound these in some form.

backends/p4tools/modules/testgen/lib/test_backend.h
```diff
@@ -48,6 +49,8 @@ class TestBackEnd {
         }
     }
 
+    [[nodiscard]] bool needsToTerminate(int64_t testCount) const { return testCount == maxTests; }
```
vhavel, line 52: We can be more defensive here

```suggestion
    [[nodiscard]] bool needsToTerminate(int64_t testCount) const { return testCount >= maxTests; }
```

vhavel (approved):

backends/p4tools/modules/testgen/lib/test_backend.h
fruffy: This will cause test generation to terminate if 0 is chosen as input instead of looping until all branches have been explored.

vhavel: Ah, ok then.

fruffy: I am not sure if there is a better solution here but I do like 0 as input for unbounded test case generation.

PR #3810, opened 2022-12-24
Fix clang formatting

ChrisDodd:
Someone checked in a broken .clang-format that broke all the formatting on pointer/reference declarations.  This fixes that.

PR #3831, opened 2023-01-09
Fix clang_format mismatch.

fruffy:
Also fix pip package versions using a requirements.txt file.

### DIFF
--- a/README.md
+++ b/README.md
@@ -264,7 +264,7 @@ libfl-dev libboost-dev libboost-iostreams-dev \
 libboost-graph-dev llvm pkg-config python3 python3-pip \
 tcpdump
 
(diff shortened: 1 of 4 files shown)


fruffy:
@mbudiu-vmw There are unrelated PRs failing because this check is missing. I also fixed all Python dependencies. Could you give this a review?

mbudiu-vmw:
What should we do about #3810?

mbudiu-vmw (approved):

fruffy:
Thanks! Regarding #3810 the clang-format file is not broken, it's just a different pointer alignment. I am indifferent about the style as long as it is consistent.

mbudiu-vmw:
We always used the other pointer style.

PR #3833, opened 2023-01-10
Unify ctxt json's id with other runtime files

kamleshbhalui:
Emit same handle id in context json as in p4info and bfrt.


kamleshbhalui:
failures are not related to changes

usha1830 (approved):

PR #3839, opened 2023-01-12
Remove nondeterminism in P4Testgen.

fruffy:
Use ordered_map instead of std::map to iterate based on insertion order, not key order. This is needed in some cases where pointers are map keys.


vhavel (approved):

davidbolvansky:
@vlstill was fighting with nondeterminism in bf-p4c, maybe he can help you in this area with some tips :)

PR #3844, opened 2023-01-15
Avoiding choosing the same port as egress port for clone.

fruffy:
Choosing the same clone port as the current egress port will throw an error. This is an attempt at a workaround.

### DIFF
--- a/backends/p4tools/modules/testgen/targets/bmv2/expr_stepper.cpp
+++ b/backends/p4tools/modules/testgen/targets/bmv2/expr_stepper.cpp
@@ -859,9 +859,10 @@ void BMv2_V1ModelExprStepper::evalExternMethodCall(const IR::MethodCallExpressio
 
              if (cloneType == BMv2Constants::CLONE_TYPE_I2E) {
                  // Pick a clone port var. For now, pick a random value from 0-511.
-                 const auto* rndConst =
-                     Utils::getRandConstantForWidth(TestgenTarget::getPortNumWidth_bits());
-                 const auto* clonePortVar = rndConst;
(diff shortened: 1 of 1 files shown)


jnfoster (approved):
Looks like a reasonable solution to this problem.

backends/p4tools/modules/testgen/targets/bmv2/expr_stepper.cpp
```diff
@@ -1143,9 +1144,10 @@ void BMv2_V1ModelExprStepper::evalExternMethodCall(const IR::MethodCallExpressio
 
              if (cloneType == BMv2Constants::CLONE_TYPE_I2E) {
                  // Pick a clone port var. For now, pick a random value from 0-511.
-                 const auto* rndConst =
-                     Utils::getRandConstantForWidth(TestgenTarget::getPortNumWidth_bits());
-                 const auto* clonePortVar = rndConst;
+                 const auto* egressPortVar = programInfo.getTargetOutputPortVar();
+                 const auto& clonePortVar = Utils::getZombieConst(
+                     egressPortVar->type, 0, "clone_port_var" + std::to_string(call->clone_id));
+                 cond = new IR::Neq(egressPortVar, clonePortVar);
```
jnfoster, line 1150: Pity we had to copy-pasta the code above :-(

fruffy: clone3, clone, clone_preserving_field_list, and recirculate_preserving_field_list are all quite similar but slightly different. It might be worthwhile to find out the common behavior and factor it out but it can be tricky here.

PR #3845, opened 2023-01-17
Fix ckadd argument

kamleshbhalui:
This patch consider header also as valid argument to Checksum methods add/sum.


usha1830 (approved):
Looks fine

PR #3850, opened 2023-01-18
Require action_run to be within a switch statement

### LINKED ISSUE #3847: is table.apply().action_run usage outside of switch statement valid?
Take:
```
#include <core.p4>

control c(in bit t)
{
  table tt {
    actions = {}
  }
  apply {
    if (tt.apply().action_run == tt.apply().action_run) {
    }
  }
}

control ct(in bit t);
package pt(ct c);

pt(c()) main;
```
This is accepted by p4test without any error message.
The spec has the following:
```
table.apply().action_run, which can only appear as the expression of a switch statement 
```
So I had assumed you could not use action_run outside of a switch statement but it seems like you can in the end ...

mbudiu-vmw:
Fixes #3847

### DIFF
--- a/frontends/p4/sideEffects.cpp
+++ b/frontends/p4/sideEffects.cpp
@@ -24,7 +24,7 @@ namespace P4 {
 
 cstring DoSimplifyExpressions::createTemporary(const IR::Type *type) {
     type = type->getP4Type();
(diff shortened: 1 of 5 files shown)


jafingerhut (approved):
LGTM.  As usual, reviewing the P4 test program and expected outputs only.

PR #3854, opened 2023-01-20
ebpf/PSA: Support for wide fields in Register and Meter

tatry:
This commit allows to use Registers and Meters with fields wider than 64 bits.

CC: @osinstom

### DIFF
--- a/backends/ebpf/codeGen.cpp
+++ b/backends/ebpf/codeGen.cpp
@@ -29,8 +29,20 @@ void CodeGenInspector::substitute(const IR::Parameter *p, const IR::Parameter *w
 }
 
 bool CodeGenInspector::preorder(const IR::Constant *expression) {
-    builder->append(Util::toString(expression->value, 0, false, expression->base));
-    return true;
+    unsigned width = EBPFInitializerUtils::ebpfTypeWidth(typeMap, expression);
+
+    if (EBPFScalarType::generatesScalar(width)) {
+        builder->append(Util::toString(expression->value, 0, false, expression->base));
+        return true;
+    }
+
+    cstring str = EBPFInitializerUtils::genHexStr(expression->value, width, expression);
+    builder->append("{ ");
+    for (size_t i = 0; i < str.size() / 2; ++i)
+        builder->appendFormat("0x%s, ", str.substr(2 * i, 2));
+    builder->append("}");
+
+    return false;
 }
 
 bool CodeGenInspector::preorder(const IR::StringLiteral *expression) {
@@ -329,9 +341,12 @@ bool CodeGenInspector::preorder(const IR::AssignmentStatement *a) {
     }
 
     if (memcpy) {
-        builder->append("memcpy(&");
+        builder->append("__builtin_memcpy(&");
         visit(a->left);
         builder->append(", &");
+        if (a->right->is<IR::Constant>()) {
+            builder->appendFormat("(u8[%u])", scalar->bytesRequired());
+        }
         visit(a->right);
         builder->appendFormat(", %d)", scalar->bytesRequired());
     } else {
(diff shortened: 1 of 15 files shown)


mbudiu-vmw:
We could promote some of you to committers if you can review each other's code.

backends/ebpf/codeGen.cpp
```diff
@@ -429,4 +444,28 @@ void CodeGenInspector::widthCheck(const IR::Node *node) const {
             node, tb->size);
 }
 
+unsigned EBPFInitializerUtils::ebpfTypeWidth(P4::TypeMap *typeMap, const IR::Expression *expr) {
+    auto type = typeMap->getType(expr);
+    if (type == nullptr) type = expr->type;
+    if (type->is<IR::Type_InfInt>()) return 32;  // let's assume 32 bit for int type
```
mbudiu-vmw, line 450: the only place where this can happen as far I know is in rhs of shifts.

backends/ebpf/ebpfParser.cpp
```diff
@@ -293,6 +295,23 @@ void StateTranslationVisitor::compileExtractField(const IR::Expression *expr, cs
...
+            // To explain the problem in error lets assume that we have bit<68> field with value:
+            //   0x11223344556677889
+            //                     ^ this digit will be parsed into a half of byte
+            // Such fields are parsed into a table of bytes in network byte order, so possible
+            // values in dataplane are (note the position of additional '0' at the end):
+            //   0x112233445566778809
+            //   0x112233445566778890
+            // To correctly insert that padding, the length of field must be known, but tools like
+            // nikss-ctl (and the nikss library) don't consume P4info.txt to have such knowledge.
+            // There is also a bug in (de)parser causing that such fields aren't deparsed correctly.
```
mbudiu-vmw, line 308: causing such fields to be deparsed incorrectly.

backends/ebpf/ebpfParser.cpp
```diff
@@ -293,6 +295,23 @@ void StateTranslationVisitor::compileExtractField(const IR::Expression *expr, cs
...
+            ::error(ErrorType::ERR_UNSUPPORTED_ON_TARGET,
+                    "%1%: fields wider than 64 bits must have size in multiply of 8 bits (1 byte) "
```
mbudiu-vmw, line 310: a size multiple of 8 bits

backends/ebpf/ebpfParser.cpp
```diff
@@ -293,6 +295,23 @@ void StateTranslationVisitor::compileExtractField(const IR::Expression *expr, cs
         builder->append(")");
         builder->endOfStatement(true);
     } else {
+        if (program->options.arch == "psa" && widthToExtract % 8 != 0) {
```
mbudiu-vmw, line 298: yes, in my implementation I never figured out how all this stuff should be handled.

backends/ebpf/psa/README.md
```diff
@@ -619,7 +619,8 @@ table caching pass `--table-caching` to the compiler.
 
 We list the known bugs/limitations below. Refer to the Roadmap section for features planned in the near future.
 
-- Larger bit fields (e.g. IPv6 addresses) may not work properly.
+- Fields wider than 64 bits must have size multiply of 8 bits, otherwise they may have unexpected value in the LSB byte.
```
mbudiu-vmw, line 622: multiple

backends/ebpf/psa/externs/ebpfPsaRegister.cpp
```diff
@@ -186,17 +189,24 @@ void EBPFRegisterPSA::emitRegisterRead(CodeBuilder *builder, const P4::ExternMet
...
-            builder->append(this->initialValue->value.str());
-        } else {
-            builder->append("(");
+        if (initialValue != nullptr || leftExpression->type->is<IR::Type_Bits>()) {
+            // let's create fake assigment statement and use it to generate valid code
+            const IR::Expression *right =
+                initialValue != nullptr ? initialValue : new IR::Constant(leftExpression->type, 0);
+            const auto *assigment = new IR::AssignmentStatement(leftExpression, right);
+            ControlBodyTranslatorPSA cg(*translator);
+            assigment->apply(cg);
```
mbudiu-vmw, line 198: does this work?
For example, the typemap won't always have a type for the right.
In general, it would be nice to do this in a separate pass before code generation.

mbudiu-vmw (approved):

mbudiu-vmw:
Approved, but you may want to make a few changes.

backends/ebpf/psa/externs/ebpfPsaRegister.cpp
tatry: I think it should works (PTF tests are passing, especially [`register.RegisterDefaultPSATest`](https://github.com/p4lang/p4c/blob/21105c8ea809ee255464784d0a671dd58f567868/backends/ebpf/tests/ptf/register.py#L84-L98)).
If the initial value is defined for register, then type map contains type for right - because of expression generated by frontend.
If the initial value is not defined, type map does not contain type for right, but type is copied from left. For this case exist this line: https://github.com/p4lang/p4c/blob/7554cb932dfae7db0b72a08c3747f1008b27bfa6/backends/ebpf/codeGen.cpp#L449

Maybe it is not the best way, but is there any reason to not to use `expression->type`?

backends/ebpf/codeGen.cpp
tatry: I can't recall where exactly hit the problem with this type.
For now I am able to make this statement true with slice, like `field[127:64]` (which is translated using bit shifts).

backends/ebpf/psa/externs/ebpfPsaRegister.cpp
mbudiu-vmw: Expression->type is only populated if you run the type inference with a flag indicating that you want this.
Some simple expressions like constants may have the type correct, though.
If it works, let's leave it like this.

osinstom:
> We could promote some of you to committers if you can review each other's code.

hey @mbudiu-vmw, I think it'd make sense to promote me and @tatry, so that we can review the code - especially related to eBPF backend.

mbudiu-vmw:
@jnfoster I have tried to add them to the list of committers, but I don't seem to be able to.
First they seem to need to be part of the p4lang org, and I can't edit that list.

PR #3855, opened 2023-01-23
Refactor Python testing scripts and utilities - Part 1 

fruffy:
- Replace the custom defined "outputs" with an actual logger that prints and logs to a file as needed 
- Use modern Python3 code (f-strings, pathlib,  etc) 
- Use subprocess.run instead of opening a process than running it.


fruffy:
@vlstill This is a big PR, but among others, it refactors the test utilities to use `subprocess.run` instead of `check_output `etc.  After this PR, I am planning to generalize ebpfenv, such that it can be used as a general virtual namespace environment. Finally, add typing information to all the Python utilities. You had some comments on the PTF PR, are those resolved or do you have any other suggestions?

fruffy:
@mbudiu-vmw This PR makes a lot of changes in the testutils library and migrates to actual logging for the test scripts from the custom logging that we had. It does not touch the Python scripts which do not depend on testutils, such as `run-bmv2-test.py` etc. 

@tatry @osinstom This PR touches ubpf scripts, too. Any dependencies I should keep in mind there?

backends/bmv2/run-bmv2-ptf-test.py
```diff
@@ -1,16 +1,16 @@
 #!/usr/bin/env python3
 
 import argparse
-import subprocess
 import os
 import sys
-import socket
 import random
+import logging
```
mbudiu-vmw, line 7: I will let the owners of this file review the changes

backends/bmv2/run-bmv2-test.py
```diff
@@ -320,39 +320,11 @@ def process_file(options, argv):
...
-            base, ext = os.path.splitext(basename)
-            dirname = os.path.dirname(options.p4filename)
-            options.observationLog = os.path.join(dirname, '%s.p4.obs' % base)
-
     try:
         result = process_file(options, argv)
     except Exception as e:
-        print("Exception ", e)
-        sys.exit(FAILURE)
+        print(f"There was a problem parsing STF file {options.testFile}:")
```
mbudiu-vmw, line 326: why is this always a parsing problem?

backends/ebpf/run-ebpf-test.py
```diff
@@ -26,126 +26,152 @@
...
-        self.compiler = ""              # Path to the P4 compiler binary
-        self.p4Filename = ""            # File that is being compiled
-        self.testfile = ""              # path to stf test file that is used
-        self.verbose = False            # Enable verbose output
-        self.replace = False            # Replace previous outputs
-        self.target = "test"            # The name of the target compiler
-        # Actual location of the test framework
-        self.testdir = os.path.dirname(os.path.realpath(__file__))
-        self.extern = ""                # Path to C file with extern definition
+        self.binary = ""  # This program's name.
```
mbudiu-vmw, line 117: I find the original version easier to read. Why spend time reformatting whitespace?

backends/ebpf/targets/bcc_target.py
```diff
@@ -14,26 +14,27 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-import os
 import sys
+from pathlib import Path
```
mbudiu-vmw, line 18: is this file even used?

backends/ebpf/targets/ebpfenv.py
```diff
@@ -16,155 +16,155 @@
     interfaces. The bridge runs in a completely isolated namespace.
     Allows the loading and testing of eBPF programs. """
 
-import os
 import sys
-from subprocess import Popen, PIPE
+import logging
```
mbudiu-vmw, line 20: I will let the owners review this file

backends/p4tools/modules/testgen/targets/bmv2/test/BMV2PTFXfail.cmake
```diff
@@ -40,7 +40,8 @@ p4tools_add_xfail_reason(
 
 p4tools_add_xfail_reason(
   "testgen-p4c-bmv2-ptf"
-  "Invalid reference to object of type"
+  "Segmentation fault"
```
mbudiu-vmw, line 43: This is not a good sign.
I see no reason for the changes in the test scripts to lead to a segfault, something must be wrong somewhere.

fruffy: In both cases the segmentation fault is thrown.

The underlying cause is a bit tricky. Previously, simple_switch_grpc was invoked with --log-console, which directly printed output to stdout/err. However, this occasionally overwhelmed the process pipe and the process got stuck. 

Now, we use `--log-file`, but simple switch crashes before the statement can be logged/generated. So it does not appear anymore. 

The completely solution is to either fix simple_switch's logging or to fix the Popen call so that it does not get stuck.

backends/bmv2/run-bmv2-ptf-test.py
fruffy: That's use! We approve :)

backends/p4tools/modules/testgen/targets/bmv2/test/BMV2PTFXfail.cmake
mbudiu-vmw: If it's the same bug please file an issue and add a comment to the makefile.

fruffy: I believe it is this bug https://github.com/p4lang/behavioral-model/issues/1124

mbudiu-vmw: I am not aware of a corresponding bug for p4c, we should have one. Can you please add one?

backends/bmv2/bmv2stf.py
```diff
@@ -668,8 +735,8 @@ def openInterface(ifname):
                 raise e
```
vlstill, line 735: This code (above, not edited) is suspicious:
- sleeping for a fixed time as a normal procedure (e.g. not timeout) seems like bad idea since it increases chances of spurious failures. Would one of the executables finish normally after it runs the test? Can we demote the the `terminate` only to timeout case and try to wait with a timeout (there is `.wait(timeout=...)`)? 
- `wait` after `termninate` might still block indefinetly if the executable ignores `SIGTERM`. In case of timeout, there should probably be attempt to `SIGTERM` it, than wait for some time and then `SIGKILL` it (`.kill()`)
- this would better be done with terminate/kill in `finally` probably

backends/ebpf/run-ebpf-test.py
```diff
@@ -154,42 +184,50 @@ def run_test(options, argv):
...
-                      options.verbose, "Removing", tmpdir)
-        shutil.rmtree(tmpdir)
+        testutils.del_dir(options.testdir)
     return result
 
 
-if __name__ == '__main__':
-    """ main """
+if __name__ == "__main__":
+    """main"""
```
vlstill, line 206:
```suggestion
```

not a very useful documentation... (and probably not even working for code blocks)

backends/ebpf/targets/ebpfenv.py
```diff
@@ -16,155 +16,157 @@
...
-from testutils import *
+# Append tools to the import path.
+FILE_DIR = Path(__file__).resolve().parent
+# Append tools to the import path.
+sys.path.append(str(FILE_DIR.joinpath("../../../tools")))
+import testutils
 
 
-class Bridge(object):
+class Bridge:
```
vlstill, line 30: I think there is some difference in behaviour of classes that are derived explicitly from `object`. You should probably check that, maybe there is a reason for that to be like it was.

backends/ebpf/targets/ebpfenv.py
```diff
@@ -16,155 +16,157 @@
...
 
     def ns_init(self):
-        """ Initialize the namespace. """
-        cmd = "ip netns add %s" % self.ns_name
-        errmsg = "Failed to create namespace %s :" % self.ns_name
-        result = run_timeout(self.verbose, cmd, TIMEOUT, self.outputs, errmsg)
+        """Initialize the namespace."""
+        cmd = f"ip netns add {self.ns_name}"
+        errmsg = f"Failed to create namespace {self.ns_name} :"
+        result = testutils.exec_process(cmd, errmsg).returncode
```
vlstill, line 42: is the timeout now implicit in `exec_process`?

backends/ebpf/targets/target.py
```diff
@@ -90,31 +89,30 @@ def compile_p4(self, argv):
         args += "P4FILE=" + self.options.p4filename + " "
         # location of the P4 compiler
         args += "P4C=" + self.compiler
-        p4_args = ' '.join(map(str, argv))
-        if (p4_args):
+        p4_args = " ".join(map(str, argv))
+        if p4_args:
             # Remaining arguments
-            args += " P4ARGS=\"" + p4_args + "\" "
+            args += ' P4ARGS="' + p4_args + '" '
```
vlstill, line 95:
```suggestion
            args += f' P4ARGS="{p4_args}" '

File: backends/ebpf/targets/target.py [hunk 12]
```diff
@@ -90,31 +89,30 @@ def compile_p4(self, argv):
...
         errmsg = "Failed to compile P4:"
-        result = run_timeout(self.options.verbose, args, TIMEOUT,
-                             self.outputs, errmsg)
-        if result != SUCCESS:
+        out, returncode = testutils.exec_process(args, errmsg)
+        if returncode != testutils.SUCCESS:
             # If the compiler crashed fail the test
-            if 'Compiler Bug' in open(self.outputs["stderr"]).readlines():
-                sys.exit(FAILURE)
+            if "Compiler Bug" in str(out):
```
vlstill, line 100: is `out` binary?

tools/testutils.py
```diff
@@ -17,133 +17,160 @@
...
 
 
-def run_process(verbose, proc, timeout, outputs, errmsg):
-
+def run_process(proc, timeout, errmsg):
     def kill(process):
         process.kill()
 
-    timer = Timer(TIMEOUT, kill, [proc])
+    timer = Timer(timeout, kill, [proc])
```
vlstill, line 125:
```suggestion
    timer = Timer(timeout, lambda p: p.kill(), [proc])
```

vlstill: But probably not needed anyway: there is a `timeout` in communicate, so you should probably use that: https://docs.python.org/3/library/subprocess.html#subprocess.Popen.communicate

tools/testutils.py
```diff
@@ -17,133 +17,160 @@
...
     if proc is None:
         # Never even started
-        report_err(outputs["stderr"], "Process failed to start")
+        log.error("Process failed to start")
     return proc
 
 
-def run_process(verbose, proc, timeout, outputs, errmsg):
-
+def run_process(proc, timeout, errmsg):
```
vlstill, line 121: The name is a bit misleading, it suggest the process is executed, but in fact only its outputs are consumed.

tools/testutils.py
```diff
@@ -17,133 +17,160 @@
...
+        result = subprocess.run(
+            args, shell=True, timeout=timeout, universal_newlines=True, check=True, **output_args
+        )
+    except subprocess.CalledProcessError as e:
+        out = e.stdout
+        returncode = e.returncode
+        log.error(f"Error {returncode} when executing {e.cmd}:\n{errmsg}\n{out}")
+    else:
+        out = result.stdout
+        returncode = result.returncode
```
vlstill, line 160: I'd suggest putting this into the `try` block, it is more readable and natural and this lines of code cannot raise `CalledProcessError `.

vlstill (commented):
I was trying primarily to look at the Python side of the change, I did not check overall design. I think this goes in the right direction and I have a few suggestions in the code.

I'd suggest running the code through some python formatting checker such as `pycodestyle` and possibly through `flake8` (linter). I think that for languages that have specified code style, it generally makes sense to adhere to it (even though I might personally disagree with some python style decisions).

osinstom:
> 

@fruffy it's been a while since we worked on the uBPF backend, but there should not be any specific dependencies to be aware of.

backends/bmv2/base_test.py
```diff
@@ -33,8 +39,7 @@ class partialmethod(partial):
     def __get__(self, instance, owner):
         if instance is None:
             return self
-        return partial(self.func, instance, *(self.args or ()),
-                       **(self.keywords or {}))
+        return partial(self.func, instance, *(self.args or ()), **(self.keywords or {}))
```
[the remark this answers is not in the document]
fruffy: I have been using a `.yapf.style` and a `.pylintrc` file with a character limit set to 100. I used 100 because our C code also uses 100 as character limit. I can add this file and its formatting in a separate PR?

I have been not using black since it is a little _too_ opinionated. There are so many options for style tools, I am actually not sure which we should use.

backends/ebpf/targets/ebpfenv.py
fruffy: A lot of this code was written for Python2, which at some point required object iirc.

backends/ebpf/targets/ebpfenv.py
fruffy: Yes, the default value is 10 minutes.

backends/bmv2/base_test.py
```diff
@@ -246,22 +248,19 @@ def __exit__(self, exc_type, exc_value, tb):
         if not issubclass(exc_type, P4RuntimeWriteException):
             # let unexpected exceptions pass through
             return False
-        self.exception = exc_value  # store for later retrieval
+        self.exception = exc_value # store for later retrieval
```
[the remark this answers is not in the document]
fruffy: Having formatting that complies with Mihai's request and the two space rule is unfortunately quite difficult. I have not been able to find a configuration that does both (Aligning comments as well as maintaining a minimum distance). 

 Personally, I prefer to have no inline comments at all since they often run over the maximum character length.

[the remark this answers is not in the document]
fruffy: https://github.com/google/yapf/issues/675

This is the issue I am facing. But other formatters are also not better at this. It might be less headache to forbid inline comments in general.

backends/bmv2/base_test.py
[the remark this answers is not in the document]
vlstill: I see. Is there any reason to not add the style config files in this PR?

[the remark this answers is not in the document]
fruffy: This PR is already big, so I was hoping to do the formatting in a separate PR. Then we can also agree on a style there.

backends/p4tools/modules/testgen/targets/bmv2/test/BMV2PTFXfail.cmake
fruffy: I asked Antonin to migrate the issue.

backends/bmv2/base_test.py
```diff
@@ -25,6 +26,11 @@
 from p4.config.v1 import p4info_pb2
 import google.protobuf.text_format
 
+FILE_DIR = Path(__file__).resolve().parent
+TOOLS_PATH = FILE_DIR.joinpath("../../tools")
+sys.path.append(str(TOOLS_PATH))
+import testutils
```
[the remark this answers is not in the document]
fruffy: Really dislike this boilerplate code actually. Do you have a better idea how to handle relative parent-level imports? Without requiring the user to input the path every time.

fruffy:
@mbudiu-vmw Any concerns on your side that have not been resolved yet?

mbudiu-vmw:
This is not my favorite part of the compiler, so I try to avoid it. If you promise to take care of it, it's yours.

mbudiu-vmw (approved):

fruffy:
> This is not my favorite part of the compiler, so I try to avoid it. If you promise to take care of it, it's yours.

Thanks! A lot of the code that was touched in this PR I have written at some point, so I am on the hook anyway.

PR #3857, opened 2023-01-23
Fix flawed code in up4.p4 and the pins switch models.

fruffy:
`up.p4` was not checking whether the IPv4 header is valid before applying the routing control, which reads from the IPv4 header.

The PINS programs uses the `p4runtime_translation` annotation, which is currently not supported by PI (https://github.com/p4lang/behavioral-model/issues/1176).

### DIFF
--- a/backends/p4tools/modules/testgen/targets/bmv2/test/p4-programs/up4.p4
+++ b/backends/p4tools/modules/testgen/targets/bmv2/test/p4-programs/up4.p4
@@ -736,7 +736,9 @@ control PreQosPipe(inout parsed_headers_t hdr, inout local_metadata_t local_meta
                 }
             }
         }
-        Routing.apply(hdr, local_meta, std_meta);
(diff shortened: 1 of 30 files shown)


backends/p4tools/modules/testgen/targets/bmv2/test/p4-programs/up4.p4
```diff
@@ -736,7 +736,9 @@ control PreQosPipe(inout parsed_headers_t hdr, inout local_metadata_t local_meta
                 }
             }
         }
-        Routing.apply(hdr, local_meta, std_meta);
+        if (hdr.ipv4.isValid()) {
```
apinski-cavium, line 739: Won't it be better if Routing control checks if the header is valid?
I would assume Acl control would need the same check too ...

pkotikal (approved):

PR #3858, opened 2023-01-24
Dpdk Backend: Add IPSec support 

usha1830:
```
PNA extern for IPsec:
	// IPsec accelerator result for the current packet
	enum ipsec_status {
	    IPSEC_SUCCESS,
	    IPSEC_ERROR
	}
	 
	// IPsec accelerator extern definition
	extern ipsec_accelerator {
	                // IPsec constructor.
	                ipsec_accelerator();
	 
	                // Set the security association (SA) index for the current packet.
	                //
	                // When not invoked, the SA index to be used is undefined, leading to target dependent
	                // behavior.
	                void set_sa_index<T>(in T sa_index);
	 
	                // Set the offset to the IPv4/IPv6 header for the current packet.
	                //
	                // When not invoked, the default value set for the IPv4/IPv6 header offset is 0.
	                void set_ip_header_offset<T>(in T offset);
	 
	                // Enable the IPsec operation to be performed on the current packet.
	                //
	                // The type of the operation (i.e. encrypt/decrypt) and its associated parameters (such as
	                // the crypto cipher and/or authentication parameters, the tunnel/transport mode headers,
	                // etc) are completely specified by the SA that was set for the current packet.
	                //
	                // When enabled, this operation takes place once the deparser operation is completed.
	                void enable();
	 
	                // Disable any IPsec operation that might have been previously enabled for the current
	                // packet.
	                void disable();
	 
	                // Returns true when the current packet has been processed by the IPsec accelerator and
	                // reinjected back into the pipeline, and false otherwise.
	                bool from_ipsec(out ipsec_status status);
	}

```

This extern is to be included in dpdk specific pna.p4

2. Pseudo Header for IPsec accelerator (platform_hdr_t)
This a pseudo header inserted by the compiler with one field for security association index. This header needs to be instantiated in the main encapsulating header struct which is passed as argument to MainParser and MainControl

```
			T is a bit type 
			header platform_hdr_t {
			    T sa_id;
			}
			
			struct headers_t {
			    ... existing headers
				platform_hdr_t ipsec_hdr;
			}

```

Insert the pseudo header only if ipsec_accelerator is instantiated and enable is called.
3. IPSec needs 4 location registers. Size of these registers is 1 and these are set by the control plane.
-    ipsec_port_out_inbound
-    ipsec_port_out_outbound
-    ipsec_port_in_inbound
-    ipsec_port_in_outbound

4.  Method “set_sa_index”: sets the **ipsec_hdr.sa_id** field.

5. Method “enable”:
a. Needs to validate the ipsec_instrisic header: **ipsec_hdr.setValid()**
b. Needs to set the output port to the IPsec output port. This is done in the beginning of deparser, so that if there are multiple calls to enable and disable, the setting of output port for ipsec is done just once.
```
	1. if (ipsec_hdr.isValid()) 
		port_out = (direction == NET_TO_HOST) ? ipsec_inbound_port_out : ipsec_outbound_port_out ;
	2. //Deparser original start
	3. emit(ipsec_hdr)
	4. emit(rest of the headers)
```
6. Method “disable”:
a. Needs to invalidate the ipsec_hdr: **ipsec_hdr.setInvalid()**
7. Method “from_ipsec”:
a. Needs to test if the input port is equal to the IPsec accelerator port
b. Currently, for P4-DPDK, the status is always set to “success”.
```
status = SUCCESS
from_ipsec = (port_in == ipsec_port_in_inbound) || (port_in == ipsec_port_in_outbound)
```


backends/dpdk/examples/ipsec.p4
```diff
@@ -0,0 +1,290 @@
...
+	packet_in pkt,
+	out   headers_t hdrs,
+	inout metadata_t meta,
+	in    pna_main_parser_input_metadata_t istd)
+{
+	bool from_ipsec;
+	ipsec_status status;
+
+	state start {
+		from_ipsec = ipsec.from_ipsec(status);
```
hanw, line 90: Can you show how 'status' might be used in the parser?

backends/dpdk/dpdkArch.cpp
```diff
@@ -3035,4 +3035,153 @@ std::vector<std::pair<cstring, const IR::Type *>>
...
+    auto decls = new IR::IndexedVector<IR::StatOrDecl>;
+    for (auto reg : portRegs)
+        decls->push_back(getRegDeclarationInstance(reg, IPSEC_PORT_REG_SIZE,
+                                                   IPSEC_PORT_REG_INDEX_BITWIDTH,
+                                                   IPSEC_PORT_REG_INITVAL_BITWIDTH));
+    return decls;
+}
+
+const IR::Node *InsertReqDeclForIPSec::preorder(IR::P4Program *program) {
+    auto ipsecInfo = new CollectIPSecInfo(is_ipsec_used, sa_id_width, refMap, typeMap, structure);
```
mbudiu-vmw, line 3050: why not just insert this is a pass in the parent pass manager?

backends/dpdk/dpdkArch.cpp
```diff
@@ -3035,4 +3035,153 @@ std::vector<std::pair<cstring, const IR::Type *>>
...
+    ipsecInfo->setCalledBy(this);
+    program->apply(*ipsecInfo);
+    if (!is_ipsec_used) return program;
+    cstring resName = "";
+    if (!uniqueNames(refMap, registerInstanceNames, resName))
+        ::error(ErrorType::ERR_RESERVED, "%1% name is reserved for DPDK IPSec port register",
```
mbudiu-vmw, line 3056: Do you still want to continue processing here?

backends/dpdk/dpdkArch.cpp
```diff
@@ -3035,4 +3035,153 @@ std::vector<std::pair<cstring, const IR::Type *>>
...
+                newHeaderName);
+        return program;
+    }
+    ipsecHeader = new IR::Type_Header(IR::ID(newHeaderName), newHeaderFields);
+
+    // Programs using ipsec accelerator must contain ethernet and ipv4 headers, hence a struct
+    // containing headerinstances for all headers must be present
+    BUG_CHECK(structure->header_type != "",
+              "Encapsulating struct containing all headers used in the program is missing");
+    for (auto obj : program->objects) {
```
mbudiu-vmw, line 3078: why not use the visitor methods for this?

backends/dpdk/dpdkArch.cpp
```diff
@@ -3035,4 +3035,153 @@ std::vector<std::pair<cstring, const IR::Type *>>
...
+            regRdArgs);
+        auto regRdOutOutbound = new IR::MethodCallExpression(
+            new IR::Member(new IR::PathExpression(IR::ID("ipsec_port_out_outbound")),
+                           IR::ID("read")),
+            regRdArgs);
+
+        auto portOutInBound = new IR::AssignmentStatement(
+            c->body->srcInfo,
+            new IR::Member(new IR::PathExpression(IR::ID("m")),
+                           IR::ID("pna_main_output_metadata_output_port")),
```
mbudiu-vmw, line 3146: one of these two assignments must be wrong, since they write to the same field

backends/dpdk/dpdkArch.cpp
```diff
@@ -3035,4 +3035,153 @@ std::vector<std::pair<cstring, const IR::Type *>>
...
+            new IR::IfStatement(c->body->srcInfo, isNetToHost, portOutInBound, portOutOutBound);
+        IR::Statement *fillIpsecPortOut = new IR::IfStatement(ipsecIsValid, ifTrue, nullptr);
+        IR::Statement *ipsecEmit = new IR::MethodCallStatement(new IR::MethodCallExpression(
+            new IR::Member(new IR::PathExpression(pkt_out->name), IR::ID("emit")), typeArgs, args));
+        stmt.push_back(fillIpsecPortOut);
+        stmt.push_back(ipsecEmit);
+        return stmt;
+    };
+
+    for (auto kv : structure->deparsers) {
```
mbudiu-vmw, line 3174: this could be also done in a visitor

backends/dpdk/dpdkArch.h
```diff
@@ -1320,5 +1320,100 @@ class CollectProgramStructure : public PassManager {
...
+          sa_id_width(sa_id_width),
+          refMap(refMap),
+          typeMap(typeMap),
+          structure(structure) {}
+    bool preorder(const IR::MethodCallStatement *mcs) override {
+        auto mi = P4::MethodInstance::resolve(mcs->methodCall, refMap, typeMap);
+        if (auto a = mi->to<P4::ExternMethod>()) {
+            if (a->originalExternType->getName().name == "ipsec_accelerator") {
+                if (structure->isPSA()) {
+                    ::error(ErrorType::ERR_MODEL, "%1% is not available for PSA programs",
```
mbudiu-vmw, line 1345: this can only happen if someone added a declaration of the extern by hand...

backends/dpdk/dpdkArch.h
```diff
@@ -1320,5 +1320,100 @@ class CollectProgramStructure : public PassManager {
...
+                            a->originalExternType->getName().name);
+                    return false;
+                }
+                if (a->method->getName().name == "enable") {
+                    is_ipsec_used = true;
+                } else if (a->method->getName().name == "set_sa_index") {
+                    auto typeArgs = a->expr->typeArguments;
+                    if (typeArgs->size() != 1) {
+                        ::error(ErrorType::ERR_UNEXPECTED,
```
mbudiu-vmw, line 1354: this would indicate that something is wrong with the model.p4 file, we usually use a modelError error. "unexpected" won't be as clear to the user.

backends/dpdk/dpdkArch.h
```diff
@@ -1320,5 +1320,100 @@ class CollectProgramStructure : public PassManager {
...
+    DpdkProgramStructure *structure;
+
+ public:
+    DpdkHandleIPSec(P4::ReferenceMap *refMap, P4::TypeMap *typeMap, DpdkProgramStructure *structure)
+        : refMap(refMap), typeMap(typeMap), structure(structure) {
+        passes.push_back(new InsertReqDeclForIPSec(refMap, typeMap, structure));
+        passes.push_back(new P4::ClearTypeMap(typeMap));
+        passes.push_back(new P4::ResolveReferences(refMap));
+        passes.push_back(new P4::TypeInference(refMap, typeMap, false));
+        passes.push_back(new P4::TypeChecking(refMap, typeMap));
```
mbudiu-vmw, line 1414: We usually prefer to call typechecking *before* a pass if needed.
This way a pass will build whatever data structures it needs.
Here you assume that the next pass needs a fresh typemap.
This is perhaps more reasonable in a backend - but probably the backend should document the fact that every pass expects a correct typemap.
I wonder whether this can mess up with the structure you have built earlier.

backends/dpdk/dpdkHelpers.cpp
```diff
@@ -416,6 +416,57 @@ bool ConvertStatementToDpdk::preorder(const IR::AssignmentStatement *a) {
                 if (e->method->getName().name == "lookahead") {
                     i = new IR::DpdkLookaheadStatement(left);
                 }
+            } else if (e->originalExternType->getName().name == "ipsec_accelerator") {
+                if (e->method->getName().name == "from_ipsec") {
+                    auto argSize = e->expr->arguments->size();
+                    if (argSize != 1) {
+                        ::error(ErrorType::ERR_UNEXPECTED,
```
mbudiu-vmw, line 423: again a model error

backends/dpdk/dpdkUtils.h
```diff
@@ -29,5 +30,11 @@ bool isEightBitAligned(const IR::Expression *e);
 bool isDirection(const IR::Member *m);
 bool isHeadersStruct(const IR::Type_Struct *st);
 const IR::Type_Bits *getEightBitAlignedType(const IR::Type_Bits *tb);
+
+// Check for reserved names for DPDK target
```
mbudiu-vmw, line 34: either the comment or the function name can be improved

backends/dpdk/dpdkUtils.h
```diff
@@ -29,5 +30,11 @@ bool isEightBitAligned(const IR::Expression *e);
...
+bool uniqueNames(P4::ReferenceMap *refMap, std::vector<cstring> names, cstring &resName);
+// Creates Register extern declaration for holding persistent information
+IR::Declaration_Instance *getRegDeclarationInstance(cstring instanceName, int regSize,
```
mbudiu-vmw, line 37: why "get" and not "create" then?

mbudiu-vmw (approved):

backends/dpdk/dpdkArch.cpp
```diff
@@ -3035,4 +3035,153 @@ std::vector<std::pair<cstring, const IR::Type *>>
...
+    IR::ID newHeaderFieldName("sa_id");
+    IR::IndexedVector<IR::StructField> newHeaderFields;
+    newHeaderFields.push_back(
+        new IR::StructField(newHeaderFieldName, IR::Type_Bits::get(sa_id_width)));
+    if (!uniqueNames(refMap, {newHeaderName}, resName)) {
+        ::error(ErrorType::ERR_RESERVED, "%1% type name is reserved for DPDK platform header",
+                newHeaderName);
+        return program;
+    }
+    ipsecHeader = new IR::Type_Header(IR::ID(newHeaderName), newHeaderFields);
```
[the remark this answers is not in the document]
usha1830: This will be only added if ipsec_accelerators enable method is called.

backends/dpdk/examples/ipsec.p4
usha1830: It can be tested in "if" or "verify" statements. 
The from_ipsec method can well be called outside the parser and the status can tested in if.

usha1830:
Thanks for the review @hanw @mbudiu-vmw
I will update the code as per comments

backends/dpdk/dpdkArch.h
usha1830: Since we are using PNA direction and PNA output port in the code. It is safer to reject any usage of this extern for PSA. I would prefer to keep this check.

backends/dpdk/dpdkArch.cpp
usha1830: Here, we need to insert this set of statements (from lambda function) in deparser. We have the information about which control blocks is/are deparser(s) stored in the "structure". This information is collected in the initial passes while inspecting the program as per specified architecture.  Hence, I am checking that while visiting P4Control. I am not able to think of any other way to achieve this here. Please suggest if I am missing something.

backends/dpdk/dpdkArch.cpp
usha1830: These assignments are placed in true and false blocks of "if-else"
corresponds to 
   ```
  if (direction == NET_TO_HOST)
	     m.pna_main_output_metadata_output_port = ipsec_inbound_port_out
     else
	     m.pna_main_output_metadata_output_port = ipsec_outbound_port_out
```

backends/dpdk/dpdkArch.cpp
usha1830: Visitors for Type_Struct and P4Parser? This is for adding the new  header and register declarations to the program. Only for placing these new declarations better in the program, I have added check for Type_Struct and Parser.

backends/dpdk/dpdkArch.cpp
mbudiu-vmw: it would be clearer to factor out the common subexpression. But it makes sense.

usha1830: Updated comments and refactored common subexpression.

mbudiu-vmw:
I have already approved

PR #3860, opened 2023-01-24
Replace invalid clone id 0 with 1 for selected BMv2 programs

fruffy:
The P4Runtime specification [reserves](https://p4.org/p4-spec/p4runtime/main/P4Runtime-Spec.html#sec-clonesessionentry) session ID 0. Correspondingly, some of the test programs in the test suite do not configure the session ID correctly. It does not look like any of the tests was intended to validate this feature. 

 This PR changes the ID to 1 so that the tests can run using P4Runtime.

### DIFF
--- a/backends/p4tools/modules/testgen/targets/bmv2/test/BMV2PTFXfail.cmake
+++ b/backends/p4tools/modules/testgen/targets/bmv2/test/BMV2PTFXfail.cmake
@@ -246,7 +246,7 @@ p4tools_add_xfail_reason(
   "testgen-p4c-bmv2-ptf"
   "At index"
   # At index 0: UNKNOWN, 'Error when adding match entry to target'
-  # enums are not supported in P4Runtime yey https://github.com/p4lang/behavioral-model/issues/1178
+  # enums are not supported in P4Runtime yet https://github.com/p4lang/behavioral-model/issues/1178
   issue1062-1-bmv2.p4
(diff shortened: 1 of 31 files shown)


jafingerhut (approved):
LGTM.  Don't be surprised if you find other issues like this with P4 test programs that have never been tested with control plane API configuration in these kinds of areas.  Most of the P4 programs were only written with testing some aspect of p4c in mind.  Regarding the particular things you are changing with this PR, I would not be surprised if BMv2's Thrift API allows configuration of mirror session with id 0.  In any case, good to make modest changes like this if it improves the ability to test these programs using P4Runtime API.

fruffy:
> LGTM. Don't be surprised if you find other issues like this with P4 test programs that have never been tested with control plane API configuration in these kinds of areas. Most of the P4 programs were only written with testing some aspect of p4c in mind. Regarding the particular things you are changing with this PR, I would not be surprised if BMv2's Thrift API allows configuration of mirror session with id 0. In any case, good to make modest changes like this if it improves the ability to test these programs using P4Runtime API.

Yes, in fact the STF tests for these programs were passing because they use the Thrift API. 
@mbudiu-vmw  could you approve? This code touches some code in the tools back end and needs extra approval.

vhavel (approved):
LGTM, all these testcases seem to work with arbitrary session id, allowing for using these with p4runtime is an improvement

PR #3870, opened 2023-01-31
Add style and formating files for Python. Format Python code.

fruffy:
@vlstill @davidbolvansky 

Adding a .pylintrc and a pytoml for style recommendations and formatting. Line-width is 100 because the majority of the files in this repository are written for a width of 100 characters. ~~Yapf was chosen as formatter because it is configurable, as opposed to black, which enforces a particular style.~~ [Black](https://github.com/psf/black) is the formatter of choice because it is well maintained, by official Python maintainers, and requires little configuration.


.pylintrc
```diff
@@ -0,0 +1,618 @@
...
+# List of plugins (as comma separated values of python module names) to load,
+# usually to register additional checkers.
+load-plugins=
+
+# Pickle collected data for later comparisons.
+persistent=yes
+
+# Minimum Python version to use for version dependent checks. Will default to
+# the version used to run pylint.
+py-version=3.10
```
vlstill, line 87: Should this be specified? It seems like something that will be hard to find once the build images are updated and newer minimal python is used.

fruffy: We can probably set it to 3.8 instead, the standard Ubuntu 20.04 version.

fruffy:
@vlstill Any other comments?

tools/cpplint.py
```diff
@@ -329,8 +327,8 @@
     'runtime/memset',
     'runtime/indentation_namespace',
     'runtime/operator',
-    # This check currently does not work correctly. So it is disabled
-    # 'runtime/override',
+                                            # This check currently does not work correctly. So it is disabled
+                                            # 'runtime/override',
```
vlstill, line 331: This is suspiciously formatted

tools/cpplint.py
```diff
@@ -46,7 +45,7 @@
 import getopt
```
vlstill, line 45: This file probably should be excluded from formatting, it is an external tool just included here for convenience I presume.

tools/driver/p4c_src/main.py
```diff
@@ -369,17 +473,17 @@ def main():
...
               'was specified via "' + JSON_input_flag + '"'
               " _and_ (secondly) " + str(len(opts.P4_source_files)) + ""
-              ###                                                     ^^
-              ### The above nonsense is to get rid of a W504 in “pycodestyle”
-              ###   [“W504 line break after binary operator”]
-              ### The same goes for “"" +” two lines down.
+                                                                               ###                                                     ^^
+                                                                               ### The above nonsense is to get rid of a W504 in “pycodestyle”
+                                                                               ###   [“W504 line break after binary operator”]
+                                                                               ### The same goes for “"" +” two lines down.
```
vlstill, line 479: This is not nice formatting :-(.

backends/ebpf/tests/ptf/common.py
```diff
@@ -309,8 +318,16 @@ def _table_create_str_from_action(self, action):
...
+    def table_write(self,
+                    method,
+                    table,
+                    key,
+                    action=0,
+                    data=None,
+                    priority=None,
+                    references=None,
+                    counters=None,
+                    meters=None):
```
vlstill, line 330: Is this always a good formatting? To me it spreads the code over too many lines, but it might still be worth the consistency.

backends/ebpf/tests/ptf/test.py
```diff
@@ -60,8 +60,8 @@ class SimpleTunnelingPSATest(P4EbpfTest):
     def runTest(self):
         pkt = Ether(dst="11:11:11:11:11:11") / testutils.simple_ip_only_packet(ip_dst="192.168.1.1")
 
-        exp_pkt = Ether(dst="11:11:11:11:11:11") / MPLS(label=20, cos=5, s=1, ttl=64) / testutils.simple_ip_only_packet(
-            ip_dst="192.168.1.1")
+        exp_pkt = Ether(dst="11:11:11:11:11:11") / MPLS(
+            label=20, cos=5, s=1, ttl=64) / testutils.simple_ip_only_packet(ip_dst="192.168.1.1")
```
vlstill, line 64: This one is the most readable due to the header name being split from its content, but I don't see a good alternative, as both

```python
        exp_pkt = Ether(dst="11:11:11:11:11:11") / \
            MPLS(label=20, cos=5, s=1, ttl=64) / \
            testutils.simple_ip_only_packet(ip_dst="192.168.1.1")
```

and

```python
        exp_pkt = (Ether(dst="11:11:11:11:11:11") /
            MPLS(label=20, cos=5, s=1, ttl=64) /
            testutils.simple_ip_only_packet(ip_dst="192.168.1.1"))
```

look somewhat weird. For me, the last one might be acceptable and so is the one you have chosen. I'd probably avoid the `\` version due to weird combination with `/`.

vlstill (commented):
Thank you for this, having a consistent formatting is definitely step in the right direction and I think the formatting is generally good. However, I can see the following problems (all also marked in code):

- The comments inside multi-line lists are pushed right beyond the longest list element even if they are on otherwise-empty lines. This looks wrong and makes these comments hard to read.
- Formatting of long calls spreads the code over too many lines (especially if the arguments are short), but it might still be worth the consistency.
- The formatting of infix operator chains is suboptimal, but I'm afraid there is no good way to do this in Python.

backends/ebpf/tests/ptf/common.py
fruffy: yapf does this in case you have named assignments. I am not sure the alternative is better. I can try to see whether there is a better configuration.

tools/driver/p4c_src/main.py
fruffy: Yes, will fix that. In general there is no good way to fix inline comments. I will try to move all comments that are inline to a separate line.

vlstill: It would be best if that could be specified in the style.

tools/driver/p4c_src/main.py
```diff
@@ -96,157 +95,144 @@ def __call__(self, parser, namespace, values, option_string=None):
...
+        "When the optimization is disabled, "
+        "for each invocation of the subparser "
+        "all states of the subparser are inlined, "
+        "which means that the subparser "
+        "might be inlined multiple times even if "
+        "it is the same instance "
+        "which is invoked multiple times. "
+        "When the optimization is enabled, the compiler "
+        "tries to identify the cases where it can inline"
+        " the subparser's states only once for multiple "
```
vlstill, line 226: Please reflow the help manually.

fruffy: Done.

fruffy:
@vlstill Mind giving this another review? I switched to black with some tweaks and the formatting seems sound.

qobilidop:
I happen to see this PR while working on updating one of the Python scripts, and have one minor suggestion. I wonder if [isort](https://pycqa.github.io/isort/) could be used together with black to sort imports besides formatting? isort [integrates](https://black.readthedocs.io/en/stable/guides/using_black_with_other_tools.html#isort) with black [easily](https://pycqa.github.io/isort/docs/configuration/black_compatibility.html) so this should be straightforward.

fruffy:
> I happen to see this PR while working on updating one of the Python scripts, and have one minor suggestion. I wonder if [isort](https://pycqa.github.io/isort/) could be used together with black to sort imports besides formatting? isort [integrates](https://black.readthedocs.io/en/stable/guides/using_black_with_other_tools.html#isort) with black [easily](https://pycqa.github.io/isort/docs/configuration/black_compatibility.html) so this should be straightforward.

Done.

fruffy:
@mbudiu-vmw Mind giving this a review?

mbudiu-vmw (approved):
I don't really see any improvement in the readability of the Python code.
I am not sure what these tools bring. If they found some bugs I would be more convinced.

fruffy:
> I don't really see any improvement in the readability of the Python code. I am not sure what these tools bring. If they found some bugs I would be more convinced.

This PR came up because of a request. These are the benefits I see:
1. The configurations essentially enforce the code preferences described in human prose in guides like [this](https://github.com/p4lang/p4c/tree/main/docs#coding-conventions) one. 
2. Applying autoformatting becomes easier because, once the style is enforced across the entire project, one does not have to worry about using a formatter when writing code. With that you do not produce diffs that are unnecessary. This also means you do not have to deal with manual code formatting. The configuration also makes sure everyone uses the same formatter for this particular project.
3. PR reviews become easier because the linters take care of arguments around style. 
4. Code style is consistent across the entire project, which does help readability (even though what can be considered readable code may be up to debate).

We could also try to apply linters that find bugs. For example, mypy likely would find many inconsistencies in our Python code. But instrumenting those is laborious and the linters are more opinionated.

PR #3873, opened 2023-02-02
Unbounded random strategy to pick a branch for test generation

hannelita:
Run with `--exploration-strategy UNBOUNDED_RANDOM_ACCESS_STACK` . No heuristics. Not bounding how deep we'll go into the stack to pick the next vector of branches to produce a test on `multipop`.


fruffy (approved):

fruffy:
Could you fix the linting failure? Then I can merge.

hannelita:
Is there any reason to cap the lint in 104 and at the same time enforce the comment in `#endif` headers? It will force the use if abbreviated files (which I think is not ideal); deeply nested directories won't help much in this case.
Can we increase the lint limit to more characters?

fruffy:
I think it is better just to shorten the names a bit, we can name this file unbounded_random_access and the other file also random access for example.

hannelita:
Just pointing out that the deep directory nesting is not helping in this case. 
I'll go ahead and change the names.

fruffy:
I think you could also disable clang-format for these sections https://stackoverflow.com/a/33258592 but this will not fix including this file etc.

PR #3874, opened 2023-02-03
Fix bug in parserUnroll

VolodymyrPeschanenkoLitSoft:
These small changes allow getting the same result for TLV_parser.p4 after the second application of the parserUnroll.

### DIFF
--- a/midend/parserUnroll.cpp
+++ b/midend/parserUnroll.cpp
@@ -168,7 +168,9 @@ class ParserStateRewriter : public Transform {
         IR::ArrayIndex *newExpression = expression->clone();
         ExpressionEvaluator ev(refMap, typeMap, valueMap);
         auto *value = ev.evaluate(expression->right, false);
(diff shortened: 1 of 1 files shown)


midend/parserUnroll.cpp
```diff
@@ -168,7 +168,9 @@ class ParserStateRewriter : public Transform {
         IR::ArrayIndex *newExpression = expression->clone();
         ExpressionEvaluator ev(refMap, typeMap, valueMap);
         auto *value = ev.evaluate(expression->right, false);
-        if (!value->is<SymbolicInteger>()) return expression;
+        if (!value->is<SymbolicInteger>() || expression->right->is<IR::Constant>()) {
```
mbudiu-vmw, line 171: maybe this case can be pushed a bit up, to avoid cloning the expression and invoking the evaluator?

mbudiu-vmw (approved):

PR #3877, opened 2023-02-06
ebpf/PSA: Support for wide fields in Digest

tatry:
This PR adds support for wide fields in Digest extern.

Also small refactor has been done - Digest implementation was placed in a separate file, like any other PSA externs.

CC: @osinstom

### DIFF
--- a/backends/ebpf/psa/externs/ebpfPsaDigest.cpp
+++ b/backends/ebpf/psa/externs/ebpfPsaDigest.cpp
@@ -81,25 +81,10 @@ class EBPFDigestPSAValueVisitor : public CodeGenInspector {
         builder->endOfStatement(true);
 
         for (const auto *f : se->components) {
-            unsigned width = EBPFInitializerUtils::ebpfTypeWidth(typeMap, f->expression);
-            builder->emitIndent();
-
-            if (EBPFScalarType::generatesScalar(width)) {
-                builder->appendFormat("%s.", tmpVar);
-                builder->append(f->name);
-                builder->append(" = ");
-                codegen->visit(f->expression);
-            } else {
-                builder->appendFormat("__builtin_memcpy(%s.", tmpVar);
-                builder->append(f->name);
-                builder->append(", ");
-                if (f->expression->is<IR::Constant>()) {
-                    builder->appendFormat("&(u8[%u])", width / 8);
-                }
-                codegen->visit(f->expression);
-                builder->appendFormat(", %u)", width / 8);
-            }
-            builder->endOfStatement(true);
+            auto type = typeMap->getType(f->expression);
+            cstring path = Util::printf_format("%s.%s", tmpVar, f->name.name);
+            codegen->emitAssignStatement(type, nullptr, path, f->expression);
+            builder->newline();
         }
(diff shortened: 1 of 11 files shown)


backends/ebpf/psa/externs/ebpfPsaDigest.cpp
```diff
@@ -0,0 +1,177 @@
...
+#include "ebpfPsaDigest.h"
+
+#include "backends/ebpf/ebpfType.h"
+#include "backends/ebpf/psa/ebpfPsaDeparser.h"
+
+namespace EBPF {
+
+class EBPFDigestPSAValueVisitor : public CodeGenInspector {
+ protected:
+    cstring digestMapInstance;
```
mbudiu-vmw, line 26: can't guess what this does based on the name.

backends/ebpf/psa/externs/ebpfPsaDigest.cpp
```diff
@@ -0,0 +1,177 @@
...
+        return false;
+    }
+
+    // handle expression like: "digest.pack(metadata.msg)"
+    bool preorder(const IR::Member *member) override {
+        pushExistingElem(member);
+        return false;
+    }
+
+    // handle expression like: "digest.pack(value)"
```
mbudiu-vmw, line 59: Is this the 'value' part in digest.pack?

backends/ebpf/psa/externs/ebpfPsaDigest.cpp
```diff
@@ -0,0 +1,177 @@
...
+
+        builder->emitIndent();
+        valueType->declare(builder, tmpVar, false);
+        builder->endOfStatement(true);
+
+        for (const auto *f : se->components) {
+            unsigned width = EBPFInitializerUtils::ebpfTypeWidth(typeMap, f->expression);
+            builder->emitIndent();
+
+            if (EBPFScalarType::generatesScalar(width)) {
```
mbudiu-vmw, line 95: I wonder whether this code pattern shouldn't be some utility function.

mbudiu-vmw (approved):

backends/ebpf/psa/externs/ebpfPsaDigest.cpp
tatry: I've clarified this comment and other about these `preorder`s

backends/ebpf/psa/externs/ebpfPsaDigest.cpp
tatry: I've done small refactor and now instance of `EBPFDigestPSA` is used to push element to the queue instead of using this variable.

backends/ebpf/psa/externs/ebpfPsaDigest.cpp
tatry: I've added such function in `CodeGenInspector` which is used also in `preorder(const IR::AssignmentStatement)`

tatry:
I would like to remind you to merge PR if it is looks good to you.

Only Mac OS tests fails, but they were broken at the time PR creation.

PR #3878, opened 2023-02-06
Refactor P4Testgen strategies - Part 1. Add a new strategy - greedy search.

fruffy-bfn:
- Add a lookahead visitor to P4Testgen. This visitor collects the statements that could potentially be visited for a particular statement. These latent statements are added to the branch state and are consumed by some strategies. The difference of this visitor to the normal coverage visitor is that it fully traverses the parser DAG.
- The lookahead visitor needs to be applied manually in crucial spots. For example, when a table adds an action, or when an if statements adds a branch. I have not found a better solution for this particular lookahead yet. 
- Only add statements if their source info is valid. This implies that they can be found in the original P4 program. 
- Introduce a new strategy - greedy potential (name is tentative). This strategy greedily selects the first execution state that may cover a new statement in the program. 
- Improve the way performance reports are generated - use an inja template.
- Add a plotting script to the benchmarking folder. Refactor the coverage script. 

There will be a second part, which brings some of the improvements of the new strategy to existing strategies. The second part will also rename the strategies appropriately.

### DIFF
--- a/backends/bmv2/run-bmv2-ptf-test.py
+++ b/backends/bmv2/run-bmv2-ptf-test.py
@@ -173,8 +173,9 @@ def run_test(options: Options) -> int:
     if result != testutils.SUCCESS:
         # Terminate the switch process and emit its output in case of failure.
         testutils.kill_proc_group(switch_proc)
-        switchout = switchlog.with_suffix(".txt").read_text()
-        testutils.log.error("######## Switch log ########\n%s", switchout)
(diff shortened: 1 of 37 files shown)


jnfoster:
Did you intend to include plotting scripts?

fruffy:
> Did you intend to include plotting scripts?

Yes

jnfoster (approved):
LGTM

PR #3879, opened 2023-02-07
SplitFlowVisitor to better capture split/join in ControlFlowVisitor

ChrisDodd:
- SplitFlowVisitor manages the flow_clone/flow_merge calls for nodes that want to visit children "in parallel".
- join_flows can visit other parallel branches on the stack in order to (try to) visit all parents before visiting the child.
- flow_copy method needed for ControlFlowVisitors to update(replace) the state rather than merging.
- BackwardCompatibleBroken flag to preserve/restore old behavior, for incremental update

### DIFF
--- a/ir/visitor.h
+++ b/ir/visitor.h
@@ -53,6 +53,10 @@ struct Visitor_Context {
     int depth;
 };
 
+class SplitFlowVisit_base;
+
+class Inspector;
+
 class Visitor {
  public:
     typedef Visitor_Context Context;
@@ -181,6 +185,7 @@ class Visitor {
 
     // Functions for IR visit_children to call for ControlFlowVisitors.
     virtual Visitor &flow_clone() { return *this; }
+    SplitFlowVisit_base *split_link = nullptr;
 
     /** Merge the given visitor into this visitor at a joint point in the
      * control flow graph.  Should update @this and leave the other unchanged.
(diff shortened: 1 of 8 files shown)


midend/local_copyprop.cpp
```diff
@@ -185,6 +185,14 @@ void DoLocalCopyPropagation::flow_merge(Visitor &a_) {
     }
     need_key_rewrite |= a.need_key_rewrite;
 }
+void DoLocalCopyPropagation::flow_copy(ControlFlowVisitor &a_) {
+    auto &a = dynamic_cast<DoLocalCopyPropagation &>(a_);
+    BUG_CHECK(working == a.working, "inconsitent DoLocalCopyPropagation state on copy");
```
mbudiu-vmw, line 190: typos in error message

ir/visitor.h
```diff
@@ -479,6 +498,121 @@ class ControlFlowVisitor : public virtual Visitor {
...
+};
+
+template<class N> class SplitFlowVisit : public SplitFlowVisit_base {
+    std::vector<const N **> nodes;
+    std::vector<const N *const *> const_nodes;
+
+ public:
+    explicit SplitFlowVisit(Visitor &v) : SplitFlowVisit_base(v) {}
+    void addNode(const N *&node) {
+        BUG_CHECK(const_nodes.empty(), "Mixing const and non-const in SplitFlowVisit");
```
mbudiu-vmw, line 537: maybe these 2 common checks could be moved to a "validate" method?

mbudiu-vmw (approved):

ir/visitor.h
ChrisDodd: The checks are subtly different (dealing with `const_nodes` and `nodes` in the const and non-const versions of the functions.)

midend/local_copyprop.cpp
ChrisDodd: good catch :+1:

PR #3882, opened 2023-02-09
add a feature to control logging via an annotation on annotated IR ob…

hanw:
…ject only

To use this feature, backend can set the enableLoggingGlobally to false based on a command line flag or a custom annotation, and set the enableLoggingInContext to true on the annotated IR class. 

This helps to filter excessive logging in our backend to focus on certain instance of IR class that is touched by a pass.

### DIFF
--- a/backends/p4test/p4test.cpp
+++ b/backends/p4test/p4test.cpp
@@ -74,7 +74,7 @@ class P4TestOptions : public CompilerOptions {
             "read previously dumped json instead of P4 source code");
         registerOption(
             "--turn-off-logn", nullptr,
-            [this](const char *) {
+            [](const char *) {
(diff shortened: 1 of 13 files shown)


mbudiu-vmw (approved):

mbudiu-vmw:
Don't forget to clang-format

fruffy:
If you install pip3 install --user "clang-format==15.0.4" and then run `make clang-format-fix-errors` you should get the same result as what CI checks for. 

The MacOS tests are failing for an unrelated reason, which needs to be fixed.

hanw:
Looks like the macos issue is gating a few other PRs as well?

fruffy:
I think right now PRs can still be merged despite the test failing.

PR #3884, opened 2023-02-12
Avoid hang in typechecking

ChrisDodd:
Ran into this issue into a program and stripped it down to a smaller testcase.  The problem is inferring a type of `int` for a constructor param that can't (yet) be inferred to a concrete type.  This avoids the hang, but it is not clear if something more is needed

### DIFF
--- a/frontends/p4/typeChecking/typeChecker.cpp
+++ b/frontends/p4/typeChecking/typeChecker.cpp
@@ -53,12 +53,12 @@ class ConstantTypeSubstitution : public Transform {
         auto repl = cstType;
         while (repl->is<IR::ITypeVar>()) {
             auto next = subst->get(repl->to<IR::ITypeVar>());
-            if (!next) break;
+            if (!next || next == repl) break;
             repl = next;
         }
         if (repl != cstType) {
             // We may replace a type variable with another one
-            LOG2("Inferred type " << repl << " for " << cst);
+            LOG2("Inferred type " << repl << " for " << cst << " (was " << cstType << ")");
             cst = new IR::Constant(cst->srcInfo, repl, cst->value, cst->base);
         } else {
             LOG2("No type inferred for " << cst << " repl is " << repl);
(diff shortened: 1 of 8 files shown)


mbudiu-vmw:
Trying to figure out why a type is being substituted with self. Will either propose an alternative fix or approve this one.

ChrisDodd:
The issues I'm seeing are all related to Type_InfInt.  Unfortunately as soon as I hack around one problem, I seem to run into another -- updated test case hits a further bug ("At this point in the compilation typechecking should not infer new types anymore, but it did.").

mbudiu-vmw:
The core problem is that the chain of type variable substitutions creates a cycle. This should normally not happen. Still have to figure out the root cause.

jnfoster:
I'm a bit paged out from the code, but I have a hunch for what the problem might be. 

When you implement unification, you need to *compose* substitutions. Here is the code as given by @bcpierce00 in his textbook TAPL:
tapl-unification

If substitutions are being represented as finite maps (a natural choice!) a common mistake is to think that you can implement composition of two substitutions by taking the union of the maps. But this is not correct. For example, if the first substitution maps `X` to `Y -> Z` and the second substitution maps `Y` to `int`, then applying the composition of the two substitutions should produce `int -> Z`. But applying the substitution denoted by the union of the maps will produce `Y -> Z`. At this point, a common mistake is to think that the substitution needs to be applied repeatedly to a fixed point. And on the example, this produces the right answer. But it is wrong in other cases.

When I say "common" above, that is based on my past experiences assigning ML-style type inference to undergrads. I saw these two mistakes many times. 

It's been a while since I've had my head deep in p4c's C++ code, but I recall seeing something that looked like this bug many years ago. Could that possibly be the issue?

I'd be happy to dig in if it's helpful (but not until later this week!)

mbudiu-vmw:
Yes, we have two forms of composition, one which is just map union and was supposed to be invoked when the domains and codomains are disjoint, but perhaps they aren't always.

mbudiu-vmw:
The problem stems from map composition `[X -> T]C`, where C already contains `T -> X` (X and T are both variables).

ChrisDodd:
superceded by #3889

PR #3885, opened 2023-02-12
Add the DASH BMv2 model to the tested programs.

fruffy:
Another large BMv2 program to test on.

### DIFF
--- a/.git-blame-ignore-revs
+++ b/.git-blame-ignore-revs
@@ -5,3 +5,5 @@
 e26a49dbf5749ee92585cec9e95ca781d377ebfd
 # clang-format ir and lib folder.
 7a722a3fb8205c57d7e9554fc7c62c2629720d5a
+# Switched from left to right pointer alignment.
+f1181a071844f759b618ed3c894cf56e2df1dd05
--- a/backends/bmv2/CMakeLists.txt
+++ b/backends/bmv2/CMakeLists.txt
@@ -129,9 +129,10 @@ p4c_find_tests("${P4TESTS_FOR_V1MODEL}" v1tests INCLUDE "${V1_INCLUDE_PATTERNS}"
 
 set (BMV2_V1MODEL_TEST_SUITES
   "${P4C_SOURCE_DIR}/testdata/p4_16_bmv_errors/*-bmv2.p4"
+  "${P4C_SOURCE_DIR}/testdata/p4_16_samples/dash/*.p4"
   "${P4C_SOURCE_DIR}/testdata/p4_16_samples/fabric_*/fabric.p4"
-  "${P4C_SOURCE_DIR}/testdata/p4_16_samples/pins/*.p4"
   "${P4C_SOURCE_DIR}/testdata/p4_16_samples/omec/*.p4"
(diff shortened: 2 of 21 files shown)


mbudiu-vmw (approved):

fruffy:
I believe there is a bug in the way `Inline` handles table names. It converts the table name into `outbound_acl_stage1:dash_acl_rule|dash_acl` instead of `outbound_acl_stage1`.

fruffy:
@jafingerhut You mentioned the dash programs were slightly unusual in the way they were written. Were you referring to the table names, for example? Are names like `stage1:dash_acl_rule|dash_acl` legal P4Runtime names?

jafingerhut:
@fruffy I had forgotten about the unusual characters in @name annotations, which are a bit odd, true.

More likely what I was referring to is that there are new custom match_kinds and syntax of proposed non-standard P4 language constructs in some of the original source files (e.g. search for a match_kind 'list' and I think also 'list_port', and a keyword 'struct_graph' in some of the files), that give errors if you attempt to compile them with the latest published p4c.  Those are only enabled with certain preprocessor symbol #define settings that you probably did not use.

jafingerhut:
@fruffy By the way, are the unusual characters in @name annotations causing problems within p4c itself?  e.g. causing intermediate P4_16 files to be generated that cannot be parsed as P4_16 source code by p4c?  Or some other issues?

It seems worth mentioning this fact to the authors of the DASH P4 program, in case they are unaware of it.

fruffy:
Rerequesting to make sure that the changes to `inline.cpp` are acceptable.

jafingerhut:
I have created this issue on the DASH repo pointing out the possible issue with `@name` annotations having characters that are not legal in P4 identifiers.

@fruffy I am guessing that if the string contents of a `@name` annotation were a legal P4 identifier name, there should be no problems encountered?

apinski-cavium:
I remember asking about `@name` specifically on keys as the p4c will definitely create `@name` annotations like `a & 15`.

mbudiu-vmw:
https://p4.org/p4-spec/docs/P4-16-v-1.2.3.html#sec-cp-names describes how control-plane names are calculated.
I think this change is not conformant to the spec.

apinski-cavium:
> https://p4.org/p4-spec/docs/P4-16-v-1.2.3.html#sec-cp-names describes how control-plane names are calculated. I think this change is not conformant to the spec.

https://github.com/p4lang/p4-spec/pull/1141 ...

fruffy:
> https://p4.org/p4-spec/docs/P4-16-v-1.2.3.html#sec-cp-names describes how control-plane names are calculated. I think this change is not conformant to the spec.

In what way is this not conformant?  I believe the control plane will still just refer to the `@name`, which is unchanged.

mbudiu-vmw:
Is there a test case exercising the new inline code?
Then we can tell whether the result is wrong.

fruffy:
> Is there a test case exercising the new inline code? Then we can tell whether the result is wrong.

Done, the compiler seems to be handling the characters well. And so does the PTF framework, surprisingly. The STF framework does not.

fruffy:
Any remaining concerns with this PR?

mbudiu-vmw:
I have approved

PR #3893, opened 2023-02-15
Install & use GCC 9 in Ubuntu 18 build

vlstill:
Ubuntu 18 comes with GCC 7, but we don't officianlly support it, our support start with GCC 9. Therefore, it is expected that anyone using Ubuntu 18 will have to use GCC from elsewhere. The https://launchpad.net/~ubuntu-toolchain-r/+archive/ubuntu/test PPA seems like the most likely source.


mbudiu-vmw (approved):

vlstill:
It seems to work:
```
-- The C compiler identification is GNU 9.4.0
-- The CXX compiler identification is GNU 9.4.0
```

PR #3895, opened 2023-02-15
ci: Fix build on macos

rst0git:

PR #3896, opened 2023-02-16
CI: Find protobuf and pass it to cmake on macOS

vlstill:


.github/workflows/ci-test.yml
```diff
@@ -205,8 +205,16 @@ jobs:
...
+      # for some reason, brew does not install protobuf in a way cmake would be
+      # able to find it. Also, the libarary needs to be specified manually in
+      # order for the compiler to be correctly linked with protobuf.
       run: |
-          ./bootstrap.sh -DENABLE_GC=ON -DCMAKE_BUILD_TYPE=RELEASE -DENABLE_UNIFIED_COMPILATION=ON && cd build && make -j2
+          PROTOBUF=`brew --prefix --installed protobuf`
+          ./bootstrap.sh -DENABLE_GC=ON -DCMAKE_BUILD_TYPE=RELEASE \
+              -DENABLE_UNIFIED_COMPILATION=ON \
+              -DProtobuf_INCLUDE_DIR=$PROTOBUF/include \
+              -DProtobuf_LIBRARY=$PROTOBUF/lib/libprotobuf.dylib
```
vlstill, line 216: For some reason, the cmake is not able to correctly link to the library even if given path to the right directory with libraries. This is the problem also present in #3895. I've used this trick which I have learned on Archlinux where the detection is also broken.

davidbolvansky: https://stackoverflow.com/questions/34795816/hinting-findname-cmake-files-with-a-custom-directory

Did you try playing with cmake prefix path?

rst0git: This problem is caused by a recent change with in the protobuf formula (https://github.com/Homebrew/homebrew-core/pull/122277) and suspect that it might be fixed in the future.

rst0git (approved):

.github/workflows/ci-test.yml
```diff
@@ -205,8 +205,18 @@ jobs:
            echo "/usr/local/opt/bison/bin" >> $GITHUB_PATH
 
     - name: Build (MacOS)
+      # for some reason, brew does not install protobuf in a way cmake would be
+      # able to find it. Also, the libarary needs to be specified manually in
```
davidbolvansky, line 209: libarary -> library

.github/workflows/ci-test.yml
```diff
@@ -205,8 +205,18 @@ jobs:
...
+      # order for the compiler to be correctly linked with protobuf.
+      # Also, there are no static protobuf libs in homebrew.
```
davidbolvansky, line 211: Add link to https://github.com/Homebrew/homebrew-core/pull/122277 ?

davidbolvansky (approved):

fruffy:
Thanks for fixing this!

PR #3901, opened 2023-02-20
Do not allow arguments with type table, control, etc

### LINKED ISSUE #3808: Compiler bug with don'tcare type argument and table
Take this obvious invalid code:
```
action NoAction(){}

extern void f<T>(in T a);
control eq0()
{
  table t {
    const size = 100;
    actions = {NoAction;}
  }
  apply {
    f<_>(t);
  }
}
```
The compiler produces:
```
In file: /home/apinski/src/p4/p4c/frontends/p4/typeChecking/typeChecker.cpp:189
Compiler Bug: Could not find type of t
```

mbudiu-vmw:
Fixes #3808


frontends/p4/typeChecking/typeChecker.cpp
```diff
@@ -3590,6 +3596,17 @@ const IR::Node *TypeInference::postorder(IR::MethodCallExpression *expression) {
                 param = *paramIt;
             }
 
+            if (param->type->is<IR::Type_Dontcare>())
+                typeError("%1%: Could not infer a type for parameter %2%", arg, param);
```
fruffy, line 3600:
```suggestion
                typeError("%1%: Could not infer a type for parameter %2% because the parameter's type is \"Type_Dontcare\"", arg, param);
```

fruffy (approved):

rst0git (approved):

frontends/p4/typeChecking/typeChecker.cpp
mbudiu-vmw: Type_Dontcare is really an internal class name of the compiler which does not mean much to the user.
But we can say "_".

fruffy: Maybe instead of Type_Dontcare calling it a don't-care type can work? `_` also looks confusing.

frontends/p4/typeChecking/typeChecker.cpp
```diff
@@ -3590,6 +3596,20 @@ const IR::Node *TypeInference::postorder(IR::MethodCallExpression *expression) {
                 param = *paramIt;
             }
 
+            if (param->type->is<IR::Type_Dontcare>())
+                typeError(
+                    "%1%: Could not infer a type for parameter %2% "
+                    "(inferred type is don't care '_')",
```
ChrisDodd, line 3602: Shouldn't this test here be for `arg->type` rather than `param->type`?  `arg->type` is the type from the argument expression (what we're inferring) while `param->type` will be the declared type of the parameter.  As is, this would seem to flag any time we instantiate something a with a don't care and then actually infer something for it, which is very much the opposite of what the message is saying.

frontends/p4/typeChecking/typeChecker.cpp
```diff
@@ -3590,6 +3596,20 @@ const IR::Node *TypeInference::postorder(IR::MethodCallExpression *expression) {
...
+                    arg, param);
+
+            // By calling generic functions with don't care parameters
+            // we can force parameters to have illegal types.  Check here for this case.
+            // e.g., void f<T>(in T arg); table t { }; f(t);
+            if (param->type->is<IR::Type_Table>() || param->type->is<IR::Type_Action>() ||
+                param->type->is<IR::Type_Control>() || param->type->is<IR::Type_Package>() ||
+                param->type->is<IR::Type_Parser>())
+                typeError("%1%: argument cannot have type %2%", arg, param->type);
```
ChrisDodd, line 3611: Similarly here, should test `arg->type` instead of (as well as?) `param->type`

frontends/p4/typeChecking/typeChecker.cpp
mbudiu-vmw: This is really about the instantiation of a generic extern, where the generic types are used to specify parameters to some of the methods. You are allowed to use _ (don't care) for a type argument when you instantiate an extern, but then you are not supposed to call any of the methods that depend on that type. Here we detect the case when after type unification we still don't know what the parameter type is (it only unifies with _).

ChrisDodd: It should be ok to use `_` as a type argument if type inferencing can infer what the type should be based on the use.  As it is, it will ALWAYS flag an error using `_` as a type argument.

PR #3910, opened 2023-03-02
Dpdk Backend: Add support for assignments to and from large (>64-bit) operand

usha1830:
To support assignments between variables <= 64-bit(small) and >64-bit (large) variables, the smaller operand must be a header field.

For ex:

```
bit<128> var1;
bit<32> var2;
var 1 = (bit<128>) var2  
```

var2 should be moved to a header to make this assignment work on the DPDK target.
Compiler creates a pseudo header for holding the variables that need to be a header field and inserts additional assignments to move the smaller operand in a pseudo header field

```
header pseudo_hdr {
bit<32> pseudo;
}

struct mainHeader{
....
pseudo_hdr p;
}

h.p.pseudo = var2
var1 = (bit<128>)h.p.pseudo;
```

In case, large operand is being assigned with a constant, the constant is capped to 64-bits and is moved in a header field. The new header field then replaces the constant in the assignment.

### DIFF
--- a/backends/dpdk/dpdkArch.cpp
+++ b/backends/dpdk/dpdkArch.cpp
@@ -2939,10 +2939,18 @@ const IR::Node *MoveNonHeaderFieldsToPseudoHeader::postorder(IR::AssignmentState
     if (is_all_args_header) return assn;
     auto result = new IR::IndexedVector<IR::StatOrDecl>();
     if ((isLargeFieldOperand(assn->left) && !isLargeFieldOperand(assn->right) &&
-         !isHeader(assn->right)) ||
+         !isInsideHeader(assn->right)) ||
         (isLargeFieldOperand(assn->left) && assn->right->is<IR::Constant>())) {
         auto expr = assn->right;
         if (auto base = assn->right->to<IR::Cast>()) expr = base->expr;
+        if (auto cst = assn->right->to<IR::Constant>()) {
+            if (!cst->fitsUint64()) {
+                ::error(ErrorType::ERR_OVERLIMIT,
+                        "DPDK target supports up-to 64-bit immediate values, %1% exceeds the limit",
+                        cst);
+                return assn;
+            }
+        }
         auto stm = addAssignmentStmt(expr);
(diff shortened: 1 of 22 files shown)


backends/dpdk/dpdkArch.cpp
```diff
@@ -2917,16 +2917,66 @@ MoveNonHeaderFieldsToPseudoHeader::addAssignmentStmt(const IR::Expression *e) {
     } else {
         type = typeMap->getType(e, true)->to<IR::Type_Bits>();
     }
+    if (e->is<IR::Constant>() && (type->width_bits() > 64)) {
+        type = IR::Type_Bits::get(64);
```
mbudiu-vmw, line 2921: this looks strange to me

backends/dpdk/dpdkUtils.cpp
```diff
@@ -75,6 +75,24 @@ bool isEightBitAligned(const IR::Expression *e) {
...
+    auto expr = e;
+    if (auto base = e->to<IR::Cast>()) expr = base->expr;
+    if (auto type = expr->type->to<IR::Type_Bits>()) {
+        auto size = type->width_bits();
+        if (size > 64) return true;
+    }
+    return false;
+}
+
+bool isHeader(const IR::Expression *expr) {
```
mbudiu-vmw, line 88: Maybe it should be called "isInsideHeader"

mbudiu-vmw (approved):

backends/dpdk/dpdkArch.cpp
usha1830: Target supports constants only up-to 64-bits. Perhaps, adding a check and issuing error in case the constant value doesn't fit in 64-bit would be better .I will update the code for that.

backends/dpdk/dpdkUtils.cpp
usha1830: Will update

usha1830:
@mbudiu-vmw Ok to merge?

PR #3912, opened 2023-03-03
ebpf/PSA: Add support for wide fields in parser value_set

tatry:
This PR adds ability to match fields wider than 64 bits in parser value set.

CC: @osinstom


backends/ebpf/codeGen.cpp
```diff
@@ -462,6 +462,11 @@ unsigned EBPFInitializerUtils::ebpfTypeWidth(P4::TypeMap *typeMap, const IR::Exp
     auto type = typeMap->getType(expr);
     if (type == nullptr) type = expr->type;
     if (type->is<IR::Type_InfInt>()) return 32;  // let's assume 32 bit for int type
+    if (type->is<IR::Type_Set>()) type = type->to<IR::Type_Set>()->elementType;
+    // FIXME: When select expression from parser supports more than one field then only the first
+    //        component will taken into account.
```
osinstom, line 467: What exactly do you want to fix here in the future? Please be explicit, like "remove below if" or sth. 

```suggestion
    // FIXME: When a select expression from the parser supports more than one field, only the first
    //        component will be taken into account.
```

osinstom (approved):

PR #3916, opened 2023-03-05
Remove predication pass from p4test

mbudiu-vmw:
The main problem with the predication pass is that it either converts all ifs in actions to conditionals, or it emits a hard error.
The ideal solution would have the predication pass leave expressions it cannot handle untouched in the hope a subsequent pass can fix them.
A less ideal solution would be to have the predication pass leave the program unchanged if it fails to convert everything.
For now I have disabled completely the predication pass in p4test.
The downside is that the tests for predication which relied on p4test no longer do anything.
But hopefully other backends exercise predication enough.
The upside is that we remove many xfails.


fruffy:
Any particular reason the pass can not handle the conditionals in actions? Is that just because of BMv2? Or is it the analysis too complex? Otherwise we could just parametrize.

fruffy (approved):

jafingerhut:
I checked the first 10 to 15 of the updated `-midend.p4` expected output files in p4_16_samples_outputs, and they looked correct to me, but there are a lot more of them than that.

Does gauntlet run on the contents of all PR's?

If not, when does gauntlet run?

fruffy:
> Does gauntlet run on the contents of all PR's?
> 
> If not, when does gauntlet run?

It does run on every PR on all the programs in the P4-16 testdata folder. However, it only runs on the output of `p4test`.

jafingerhut (approved):
LGTM.  I checked about the first 10 to 15 of the updated midend expected output files by hand, and found no errors.  Fabian says that Gauntlet is run on all of the p4test expected output files, so hopefully it should catch any errors in any others that I did not check by hand.

PR #3920, opened 2023-03-07
Add an explicit optional match key type.

fruffy:
As per discussion in https://github.com/p4lang/PI/issues/587 P4Testgen should handle optional matches explicitly instead of ignoring them. If the optional match is tainted, ignore it, but still add the priority. 

This PR also makes match keys target-independent. Now every target extension can add their own match key types without affecting other back ends.


jnfoster (approved):
LGTM

PR #3924, opened 2023-03-15
Third party libraries are now included as SYSTEM

MichalKekely:
* This is done to suppress the -pedantic warnings in some versions of google protobuf and boost


vlstill (approved):

davidbolvansky (approved):

fruffy (approved):
Nit: Maybe add a comment why these directories are included with SYSTEM?

PR #3926, opened 2023-03-15
P4Testgen: Try out a simpler advance expression calculation

fruffy:
We do not need the restriction that the advance size must be larger than the buffer. In fact, it can lead to incorrect tests.


jnfoster (approved):
Cool!

PR #3930, opened 2023-03-19
Link against the Boehm-Demers-Weiser Garbage Collector using FetchContent.

fruffy:
To have more control over linking and setup of the Boehm-Demers-Weiser garbage collector P4C uses we add it as a FetchContent dependency. For compatibility users can still choose to link with an external dependency. Several other changes were necessary to make this work: 
- We build and link the static libraries of the garbage collection library. This way we do not run into missing library problems. To be able to do so we had to disable some definitions in `gc_cpp.cpp`.
- We enable threads and fork support by default to ensure parallelism support.
- We do not link bdwgc's include directories globally. Instead we use the Abseil approach and mandate that every target links to library that pulls in the BDWGC headers.
- Clean up the top-level CMakelists a little and move all the module includes to the top.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -109,11 +109,11 @@ include(P4CUtils)
 include(FetchContent)
 set(CMAKE_POLICY_DEFAULT_CMP0077 NEW)
 # More utilities.
-include (CheckIncludeFile)
-include (CheckIncludeFileCXX)
-include (CheckFunctionExists)
-include (FindPythonModule)
-INCLUDE(CPack)
+include(CheckIncludeFile)
+include(CheckIncludeFileCXX)
+include(CheckFunctionExists)
+include(FindPythonModule)
+include(CPack)
 # TODO: Remove this deprecated include eventually.
 include(UnifiedBuild)
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -44,8 +44,7 @@ OPTION (ENABLE_P4TEST "Build the P4Test backend (required for the full test suit
 OPTION (ENABLE_TEST_TOOLS "Build the P4Tools development platform" OFF)
 OPTION (ENABLE_P4C_GRAPHS "Build the p4c-graphs backend" ON)
 OPTION (ENABLE_PROTOBUF_STATIC "Link against Protobuf statically" ON)
-OPTION (ENABLE_GC "Use libgc" ON)
-OPTION (ENABLE_MULTITHREAD "Use multithreading" OFF)
+OPTION (ENABLE_GC "Compile with the Boehm-Demers-Weiser garbage collector." ON)
 OPTION (ENABLE_WERROR "Treat warnings as errors" OFF)
 OPTION (ENABLE_SANITIZERS "Enable sanitizers" OFF)
 OPTION (STATIC_BUILD_WITH_DYNAMIC_GLIBC "Build a (mostly) statically linked release binary. \
@@ -103,11 +102,18 @@ else()
     WORKING_DIRECTORY ${PROJECT_SOURCE_DIR})
   set (P4C_VERSION "${P4C_SEM_VERSION_STRING} (SHA: ${P4C_GIT_SHA} BUILD: ${CMAKE_BUILD_TYPE})")
 endif()
+
 # P4 General Utilities
 include(P4CUtils)
 # CMake Utilities to fetch dependencies.
 include(FetchContent)
 set(CMAKE_POLICY_DEFAULT_CMP0077 NEW)
+# More utilities.
+include (CheckIncludeFile)
+include (CheckIncludeFileCXX)
+include (CheckFunctionExists)
+include (FindPythonModule)
+INCLUDE(CPack)
 # TODO: Remove this deprecated include eventually.
 include(UnifiedBuild)
 
@@ -198,6 +204,10 @@ p4c_obtain_abseil()
 include(Protobuf)
 p4c_obtain_protobuf()
 
+# We require -pthread to make std::call_once work, even if we're not using threads...
+set(THREADS_PREFER_PTHREAD_FLAG ON)
+find_package(Threads REQUIRED)
+
 # The boost graph headers are optional and only required by the graphs back end.
 find_package (Boost QUIET COMPONENTS graph)
 if (Boost_FOUND)
@@ -207,20 +217,18 @@ else ()
 endif ()
 find_package (Boost REQUIRED COMPONENTS iostreams)
 
+# Compile with the Boehm garbage collector (https://github.com/ivmai/bdwgc), if requested.
+# One can disable the GC, e.g., to run under Valgrind, by editing config.h.
 if (ENABLE_GC)
-  find_package (LibGc 7.4.2 REQUIRED)
-  set (HAVE_LIBGC 1)
+  include(BDWGC)
+  p4c_obtain_bdwgc()
 endif ()
 if (ENABLE_MULTITHREAD)
   add_definitions(-DMULTITHREAD)
 endif()
-# we require -pthread to make std::call_once work, even if we're not using threads...
-set(THREADS_PREFER_PTHREAD_FLAG ON)
-find_package(Threads REQUIRED)
 list (APPEND P4C_LIB_DEPS ${CMAKE_THREAD_LIBS_INIT})
 include_directories(SYSTEM ${Boost_INCLUDE_DIRS})
 include_directories(SYSTEM ${PROTOBUF_INCLUDE_DIRS})
-include_directories(SYSTEM ${LIBGC_INCLUDE_DIR})
 set (HAVE_LIBBOOST_IOSTREAMS 1)
 list (APPEND P4C_LIB_DEPS ${Boost_LIBRARIES})
 if (ENABLE_GC)
@@ -229,19 +237,18 @@ endif ()
 set (P4C_ABSL_LIBRARIES absl::flat_hash_map absl::flat_hash_set)
 list (APPEND P4C_LIB_DEPS ${P4C_ABSL_LIBRARIES})
 
-# other required libraries
+# Other required libraries.
 p4c_add_library (rt clock_gettime HAVE_CLOCK_GETTIME)
 
-# check includes
-include (CheckIncludeFile)
-include (CheckIncludeFileCXX)
+# Check includes.
 check_include_file(execinfo.h HAVE_EXECINFO_H)
 check_include_file(ucontext.h HAVE_UCONTEXT_H)
 check_include_file(backtrace-supported.h HAVE_LIBBACKTRACE)
 check_include_file_cxx(mm_malloc.h HAVE_MM_MALLOC_H)
 check_include_file_cxx(cxxabi.h HAVE_CXXABI_H)
 if (HAVE_LIBBACKTRACE)
-set (P4C_LIB_DEPS "${P4C_LIB_DEPS};-lbacktrace")
+  message(STATUS "Linking libbacktrace.")
+  set (P4C_LIB_DEPS "${P4C_LIB_DEPS};-lbacktrace")
 endif ()
 
 # check functions
@@ -252,16 +259,13 @@ if (ENABLE_GC)
   set (CMAKE_REQUIRED_LIBRARIES ${CMAKE_REQUIRED_LIBRARIES} ${LIBGC_LIBRARIES})
 endif ()
 
-include (CheckFunctionExists)
 check_function_exists (memchr HAVE_MEMCHR)
 check_function_exists (pipe2 HAVE_PIPE2)
-check_function_exists (GC_print_stats HAVE_GC_PRINT_STATS)
 
 # restore CMAKE_REQUIRED_LIBRARIES
 set (CMAKE_REQUIRED_LIBRARIES ${CMAKE_REQUIRED_LIBRARIES_PRECHECK})
 
 # python modules
-include (FindPythonModule)
 find_python_module (difflib REQUIRED)
 find_python_module (shutil REQUIRED)
 find_python_module (tempfile REQUIRED)
@@ -421,7 +425,7 @@ FetchContent_Declare(
 )
 FetchContent_MakeAvailable(p4runtime)
 set(FETCHCONTENT_QUIET ${FETCHCONTENT_QUIET_PREV})
-message("Done with setting up P4Runtime for P4C.")
+message(STATUS "Done with setting up P4Runtime for P4C.")
 # The standard P4Runtime protocol buffers message definitions live in the PI
 # repo, which is included in this repo as a git module.
 set(P4RUNTIME_STD_DIR ${p4runtime_SOURCE_DIR}/proto CACHE INTERNAL
@@ -626,6 +630,5 @@ SET(CPACK_SOURCE_PACKAGE_FILE_NAME
    "p4c-${P4C_SEM_VERSION_STRING}")
 SET(CPACK_SOURCE_IGNORE_FILES
    "${PROJECT_SOURCE_DIR}/${CMAKE_PROJECT_NAME}-*;${PROJECT_SOURCE_DIR}/${CMAKE_PROJECT_NAME}_*;/build/;/.git/;/config.log;/CMakeFiles/;CMakeCache.txt$;.tar.gz$;/_CPack_Packages;/Makefile$;~$;/build-deb;/clean-deb;/filter-empty-entries;/make-symbols;/make-ppa;/make-deb;/debian.conf;/make-rpm;/rpm.conf;${CPACK_SOURCE_IGNORE_FILES}")
-INCLUDE(CPack)
 
 ADD_CUSTOM_TARGET(dist COMMAND ${CMAKE_MAKE_PROGRAM} clean package_source)
--- a/cmake/BDWGC.cmake
+++ b/cmake/BDWGC.cmake
@@ -1,0 +1,74 @@
+macro(p4c_obtain_bdwgc)
+  include(CheckSymbolExists)
+
+  option(
+    P4C_USE_PREINSTALLED_BDWGC
+    "Look for a preinstalled version of BDWGC in the system instead of installing a prebuilt binary using FetchContent."
+    OFF
+  )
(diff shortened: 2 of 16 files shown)


vlstill:
Hi @fruffy! What is the motivation to turn libGC into a submodule? Do we need something from a specific new version? Even if so, is submodule really a good solution? Unless we want to track upstream closely I think we should stick either to using installed libgc, or if we can't we can use cmake's fetch content. A submodule adds extra overhead for checkout (and both submodule and fetch content add build overhead, but I guess that is less of a concern).

There are concerns that this would impact downstream compilers which now use a fixed version of libGC in their build/release environments (especially if p4c was to track upstream libCG closely, as opposed to tracking releases).

fruffy:
> Hi @fruffy! What is the motivation to turn libGC into a submodule? Do we need something from a specific new version? Even if so, is submodule really a good solution? Unless we want to track upstream closely I think we should stick either to using installed libgc, or if we can't we can use cmake's fetch content. A submodule adds extra overhead for checkout (and both submodule and fetch content add build overhead, but I guess that is less of a concern).
> 
> There are concerns that this would impact downstream compilers which now use a fixed version of libGC in their build/release environments (especially if p4c was to track upstream libCG closely, as opposed to tracking releases).

libGC interacts badly with the current version of the Z3 solver (for example, https://github.com/p4lang/p4c/pull/3927) and more recent version of Z3 (4.12.0) do not work at all. A newer version of the garbage collector and the [appropriate configuration]([4.12.0](https://github.com/Z3Prover/z3/releases/tag/z3-4.12.0)) appears to fix these problems and makes the garbage collection more stable (for example, the crash you see every time when running gdb disappears). Increased performance is a nice side-effect.

It seems prudent to include something as essential as the garbage collection directly as a submodule. It allows us to more easily configure the collector and fix bugs like this one. Thankfully, there are not many source files and it compiles very quickly. 

There is still a lot more testing to do, of course. I am not planning to merge this any time soon, but it is good to have this PR as scratch pad to weed out issues with P4Testgen and Z3 etc.

vlstill:
I see. I've noticed the Z3 problem with libGC but did not investigate it further. It is good that it can be fixed on libGC side.

The improvements in stability & speed would be very nice :-).

> It seems prudent to include something as essential as the garbage collection directly as a submodule. It allows us to more easily configure the collector and fix bugs. Thankfully, there are not many source files and it compiles very quickly.

If we are dependent on a particular (minimal) version that is not widely available and on custom configuration than I agree this would be simpler or even required.

I still wonder if we can:

- use a release tag/branch for our submodule
- use [fetch-content](https://cmake.org/cmake/help/latest/module/FetchContent.html) instead to submodule to avoid having the sources checked out in the repo (but still have a fixed version from git, just using cmake instead of submodules).

fruffy:
> I see. I've noticed the Z3 problem with libGC but did not investigate it further. It is good that it can be fixed on libGC side.
> 
> The improvements in stability & speed would be very nice :-).
> 
> > It seems prudent to include something as essential as the garbage collection directly as a submodule. It allows us to more easily configure the collector and fix bugs. Thankfully, there are not many source files and it compiles very quickly.
> 
> If we are dependent on a particular (minimal) version that is not widely available and on custom configuration than I agree this would be simpler or even required.
> 
> I still wonder if we can:
> 
>     * use a release tag/branch for our submodule
> 
>     * use [fetch-content](https://cmake.org/cmake/help/latest/module/FetchContent.html) instead to submodule to avoid having the sources checked out in the repo (but still have a fixed version from git, just using cmake instead of submodules).

FetchContent works great, thanks for pointing me to this. It took a little tweaking but now I got it integrated with the CMake build. The one downside is that this will not work with bazel etc, but those build scripts do not use GC anyway... 

The only thing left to fix now is the MacOS build.

CMakeLists.txt
```diff
@@ -143,23 +141,54 @@ else ()
...
+  set(enable_large_config ON CACHE BOOL "Optimize for large heap or root set.")
+  add_definitions("-DREDIRECT_REALLOC=GC_REALLOC")
+  add_definitions("-DREDIRECT_FREE=GC_FREE")
+  if (APPLE)
+    append("-Wno-deprecated-declarations" CMAKE_C_FLAGS CMAKE_CXX_FLAGS)
+    # add_definitions("-DGC_NO_PTHREAD_SIGMASK")
+    # add_definitions("-DNO_MARKER_SPECIAL_SIGMASK")
+  endif()
+  if(BUILD_STATIC_RELEASE)
+    set(enable_threads OFF CACHE BOOL "Support threads.")
```
davidbolvansky, line 166: why off?

davidbolvansky:
>> libGC interacts badly with the current version of the Z3 solver (for example, https://github.com/p4lang/p4c/pull/3927). A more recent version of Z3 (4.12.0) does not work at all.

Oh, that is quite serious issue for the future. Is there a bug report for this problem on z3 issue tracker? Could we bisect problematic commits?

CMakeLists.txt
fruffy: I have not been able to successfully compile it with the thread library.

fruffy:
> > > libGC interacts badly with the current version of the Z3 solver (for example, #3927). A more recent version of Z3 (4.12.0) does not work at all.
> 
> Oh, that is a quite serious issue for the future. Is there a bug report for this problem on z3 issue tracker? Could we bisect problematic commits?

I have not done a deep dive why Z3 is causing these issues but it could be related to https://github.com/Z3Prover/z3/issues/6572. glibc 2.34 removes pthread which could have some effect. This might get fixed.

CMakeLists.txt
davidbolvansky: Then we can expect performance regressions in static builds (ones we give to customers :/)

fruffy: Not quite sure, if I remember correctly the Ubuntu packages do not have `THREADS` defined by default.

fruffy: E.g., if you try to use `GC_allow_register_threads` it will complain.

fruffy: https://stackoverflow.com/questions/35116327/when-g-static-link-pthread-cause-segmentation-fault-why

I am running into this problem. Somehow pthread is not linked correctly. I have not found an order to make this work. 
This change is possibly related, which makes things quite confusing. 
https://developers.redhat.com/articles/2021/12/17/why-glibc-234-removed-libpthread#the_developer_view
Dropping `-static-libgcc -static-libstdc++` does not seem to help either.

vlstill: What does this option mean? Is it enabling support for running multi-threaded programs with libGC, or is it enabling parallel runs of garbage collecting threads?

CMakeLists.txt
```diff
@@ -149,39 +157,78 @@ endif ()
...
+  # We may have to compile very large programs which allocate a lot on the heap.
+  set(enable_large_config ON CACHE BOOL "Optimize for large heap or root set.")
+  set(without_libatomic_ops OFF CACHE BOOL "Use atomic_ops.h in libatomic_ops/src" )
+
+  # MacOS has problems with threads and malloc redirect. Disable it for now.
+  # We still use the C++ new/delete etc garbage collection.
+  if (APPLE)
+    append("-Wno-deprecated-declarations" CMAKE_C_FLAGS CMAKE_CXX_FLAGS)
+    add_definitions("-DNO_MARKER_SPECIAL_SIGMASK")
+    set(enable_redirect_malloc OFF CACHE BOOL "Redirect malloc and friends to GC routines.")
```
vlstill, line 179: So this affects only malloc, not `new`/`delete`?

vlstill (commented):
Does this have impact on the compiler's performance?

cmake/BDWGc.cmake
```diff
@@ -0,0 +1,40 @@
+macro(p4c_obtain_bdwgc)
+  include(CheckSymbolExists)
+  # Turn on C++ support. We do not need the headers and documentation.
+  set(enable_cplusplus ON CACHE BOOL "C++ support")
+  set(install_headers ON CACHE BOOL "Install header and pkg-config metadata files.")
```
vlstill, line 5: Your comment says "we do **not** need the headers", but then you enable them.

lib/gc.cpp
```diff
@@ -30,18 +32,13 @@ limitations under the License.
...
 #include "log.h"
 #include "n4.h"
 
-/* glibc++ requires defining global delete with this exception spec to avoid warnings.
- * If it's not defined, probably not using glibc++ and don't need anything */
-#ifndef _GLIBCXX_USE_NOEXCEPT
-#define _GLIBCXX_USE_NOEXCEPT _NOEXCEPT
-#endif
-
+// One can disable the GC, e.g., to run under Valgrind, by editing config.h
```
vlstill, line 38: probably better to hint at disabling it in CMake

cmake/BDWGc.cmake
```diff
@@ -0,0 +1,40 @@
...
+
+  # MacOS has problems with threads and malloc redirect. Disable it for now.
+  # We still use the C++ new/delete etc garbage collection.
+  if (APPLE)
+    append("-Wno-deprecated-declarations" CMAKE_C_FLAGS CMAKE_CXX_FLAGS)
+    add_definitions("-DNO_MARKER_SPECIAL_SIGMASK")
+    set(enable_redirect_malloc OFF CACHE BOOL "Redirect malloc and friends to GC routines.")
+  else()
+    # For any other distribution we redirect all malloc calls in the program.
+    set(enable_redirect_malloc ON CACHE BOOL "Redirect malloc and friends to GC routines.")
```
vlstill, line 19: What does this option do? In `gc.cpp` there is our own `malloc` override, but this options suggests that libgc has its own way of doing so.

fruffy: Currently, it is necessary to prevent segmentation faults if configured with thread support. I have not found a configuration yet that does not require it.

asl:
> I see. I've noticed the Z3 problem with libGC but did not investigate it further. It is good that it can be fixed on libGC side.

As far as I can see, there are double frees there. So the problem might be with Z3. Just exposed / hidden by GC.

fruffy:
> > I see. I've noticed the Z3 problem with libGC but did not investigate it further. It is good that it can be fixed on libGC side.
> 
> As far as I can see, there are double frees there. So the problem might be with Z3. Just exposed / hidden by GC.

I also had that issue with linking P4C against other frameworks, for example gRPC. Unclear what causes it but bumping the GC version does seem to fix it.

asl:
> I also had that issue with linking P4C against other frameworks, for example gRPC. Unclear what causes it but bumping the GC version does seem to fix it.

On the other hand, it might be result of GC freeing pointer too early :) I am seeing that we are not forwarding aligned allocation functions via GC, it might cause some problems. We do have this patch downstream (otherwise one cannot use libgc with LLVM with C++17 enabled). I will submit PR shortly

cmake/bdwgc.patch
```diff
@@ -0,0 +1,13 @@
+diff --git a/gc_cpp.cc b/gc_cpp.cc
+index 341805a6..fb6b0a01 100644
+--- a/gc_cpp.cc
++++ b/gc_cpp.cc
+@@ -38,7 +38,7 @@ built-in "new" and "delete".
+ #endif
+ #include "gc_cpp.h"
+ 
+-#if !(defined(_MSC_VER) || defined(__DMC__)) || defined(GC_NO_INLINE_STD_NEW)
++#if !defined(SKIP_CPP_DEFINITIONS) && (!(defined(_MSC_VER) || defined(__DMC__)) || defined(GC_NO_INLINE_STD_NEW))
```
fruffy, line 10: It might be worthwhile to upstream this. This is the only modification need to be able to link BDWGC statically.

cmake/BDWGC.cmake
```diff
@@ -0,0 +1,74 @@
...
+
+    # Turn on C++ support.
+    set(enable_cplusplus ON CACHE BOOL "C++ support")
+    set(install_headers OFF CACHE BOOL "Install header and pkg-config metadata files.")
+    set(enable_docs ${ENABLE_DOCS} CACHE BOOL "Build and install documentation.")
+    # We may have to compile very large programs which allocate a lot on the heap.
+    set(enable_large_config ON CACHE BOOL "Optimize for large heap or root set.")
+    set(without_libatomic_ops OFF CACHE BOOL "Use atomic_ops.h in libatomic_ops/src")
+    # Redirect all malloc calls in the program.
+    set(enable_redirect_malloc OFF CACHE BOOL "Redirect malloc and friends to GC routines.")
```
ChrisDodd, line 31: We do this manually (overriding malloc to call GC_malloc) in our lib/gc.cpp -- perhaps by turning this on we could avoid that?

cmake/BDWGC.cmake
```diff
@@ -0,0 +1,74 @@
...
+    # Try to enable thread-local-storage for better performance.
+    set(enable_thread_local_alloc ON CACHE BOOL "Turn on thread-local allocation optimization")
+
+    fetchcontent_declare(
+      bdwgc
+      GIT_REPOSITORY https://github.com/ivmai/bdwgc.git
+      GIT_TAG e340b2e869e02718de9c9d7fa440ef4b35785388 # 8.2.6
```
ChrisDodd, line 38: Is there an easy way for someone to override this if they want to use their own repo/copy of bdwgc instead of the public one?  This would apply to other things we're pulling with fetchcontent (protobuf, abseil, etc) -- it would be nice if there was one place all the references (repo URLs and tags) were collected rather than being scattered across different CMakefiles for each dependency.

cmake/BDWGC.cmake
fruffy: I tried that previously but I saw some performance regressions. Maybe we can experiment with this once this PR has been merged.

cmake/BDWGC.cmake
fruffy: Are you looking for a central CMake file that contains all the Github tags of dependencies we pull? We could define that and use it for all core FetchContent dependencies. So far that would be Abseil, Protobuf, Gtest, and BDWGC. 

I think where this gets complicated is with modularity. For example, the GC dependency is only pulled in when `ENABLE_GC` is toggled. `Protobuf` is also only meant to be installed when we enable the control plane.

fruffy: I think a central dependency repository might be tricky since every back end manages their own dependencies. There is [string(JSON...)](https://cmake.org/cmake/help/latest/command/string.html#json) which we could use to parse a JSON file with all the necessary information on dependencies. That file could also be user-provided.

We could also just create a helper `fetchcontent_declare_with_option` which adds CMake options to set the URL, URL_HASH and GIT TAG for each `fetchcontent_declare` invocation.

ChrisDodd (approved):
Looks like this is still getting hung up on the fedora and old Ubuntu issues, but hopefully doesn't interoduce any more hard-to-debug dependencies.

fruffy:
> Looks like this is still getting hung up on the fedora and old Ubuntu issues, but hopefully doesn't interoduce any more hard-to-debug dependencies.

https://github.com/p4lang/p4c/pull/4623 introduced the Ubuntu 18.04 failure. Not sure why... for some reason this version of boost pulls in hash_combine for the `cpp_int`.

The Fedora failure should be fixed in #4644.

vlstill:
> What is the reason to use submodule instead of FetchContent?

I see that is just an out-of-date top of the PR, could you please update it?

vlstill (commented):
A few little things.

cmake/BDWGC.cmake
```diff
@@ -0,0 +1,74 @@
...
+    fetchcontent_makeavailable_but_exclude_install(bdwgc)
+
+    # Bdwgc source code may trigger warnings which we need to ignore.
+    set_target_properties(gc PROPERTIES COMPILE_FLAGS "-Wno-error -w")
+    set_target_properties(gccpp PROPERTIES COMPILE_FLAGS "-Wno-error -w")
+    set_target_properties(cord PROPERTIES COMPILE_FLAGS "-Wno-error -w")
+
+    # Add some extra compile definitions which allow us to use our customizations.
+    target_compile_definitions(gc PUBLIC
+      HANDLE_FORK ENABLE_DISCLAIM GC_USE_DLOPEN_WRAP SKIP_CPP_DEFINITIONS
```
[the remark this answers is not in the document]
fruffy: Will do. 
For context on `DLOPEN_WRAP` https://github.com/ivmai/bdwgc/blob/master/docs/README.macros#L416 and https://github.com/ivmai/bdwgc/issues/561#issuecomment-1610830189
It seems to help in multi-threaded environments. 

For an explanation of `SKIP_CPP_DEFINTIONS` see https://github.com/p4lang/p4c/pull/3930#discussion_r1554197842

fruffy:
@asl Planning to merge this soon. Lmk if you have any concerns here.

cmake/BDWGC.cmake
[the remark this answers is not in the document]
ChrisDodd: > > GC_USE_DLOPEN_WRAP      Causes the collector to redefine malloc and
> > intercepted pthread routines with their real names, and causes it to use
> > dlopen and dlsym to refer to the original versions.  This makes it possible
> > to build an LD_PRELOADable malloc replacement library.
> 
> Doesn't this interfere with our own re-declaration of `malloc`? Or maybe the BDWGC one is not linked because there is already malloc defined by our binary...

I think it would be ok, as malloc ends up redirected to the gc regardless, but the main reason we have redeclarations of malloc is that the default libgc builds on Ubuntu and others DON'T use GC_USE_DLOPEN_WRAP.  With it, we should be able to remove our malloc replacements.

vlstill (approved):

cmake/BDWGC.cmake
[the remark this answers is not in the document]
vlstill: yes, I was wondering if we could actually get rid of our `gc.cpp` (we would have to be willing to drop support for distribution version of libgc in most distros).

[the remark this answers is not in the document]
asl: @vlstill I'm not sure we can drop on Darwin. Though, it's worth investigating.

cmake/bdwgc.patch
fruffy: https://github.com/ivmai/bdwgc/pull/643

PR #3937, opened 2023-03-23
Implement a PNA-DPDK extension with a metadata test back end for P4Testgen

fruffy-bfn:
This pull request implements a [PNA](https://p4.org/p4-spec/docs/PNA.html) extension for P4Testgen. The current back end is specialized towards [DPDK](https://github.com/p4lang/p4c/tree/main/backends/dpdk), but the common code is generic enough to support other back ends. 

Overall, the back end is still rather simple and only implements a few externs. There is no software target to execute tests on yet. 

However, the extension does contain a test back end. The test back end is the metadata back end, the same back end as v1model supports.


jnfoster (approved):
LGTM. Thanks!

PR #3938, opened 2023-03-24
[experimental] Allow actions to instantiate externs

mihaibudiu:
This shows an example that would allow the effect of Direct registers in P4.

### DIFF
--- a/testdata/p4_16_samples/action-register.p4
+++ b/testdata/p4_16_samples/action-register.p4
@@ -1,0 +1,26 @@
+#include <core.p4>
+
+extern Local<D> {
+  Local();
+  D read();
+  void write(in D data);
+}
+
+control C() {
+  action RW() {
+    Local<bit<32>>() local;
+    local.write(local.read() + 1);
+  }
+  table t {
+    actions = { RW; }
+  }
+  apply {
+    t.apply();
+  }
+}
+
+control CTRL();
+package top(CTRL _c);
+
+top(C()) main;
+
--- a/testdata/p4_16_samples_outputs/action-register-midend.p4
+++ b/testdata/p4_16_samples_outputs/action-register-midend.p4
@@ -1,0 +1,32 @@
+#include <core.p4>
+
+extern Local<D> {
+    Local();
+    D read();
+    void write(in D data);
+}
+
+control C() {
+    @name("C.tmp") bit<32> tmp;
+    @noWarn("unused") @name(".NoAction") action NoAction_1() {
+    }
+    @name("C.RW") action RW() {
+        @name("C.local") Local<bit<32>>() local_0;
+        tmp = local_0.read();
+        local_0.write(tmp + 32w1);
+    }
+    @name("C.t") table t_0 {
+        actions = {
+            RW();
+            @defaultonly NoAction_1();
+        }
+        default_action = NoAction_1();
+    }
+    apply {
+        t_0.apply();
+    }
+}
+
+control CTRL();
+package top(CTRL _c);
+top(C()) main;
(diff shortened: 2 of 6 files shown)


testdata/p4_16_samples_outputs/action-register-midend.p4
```diff
@@ -0,0 +1,32 @@
...
+    D read();
+    void write(in D data);
+}
+
+control C() {
+    @name("C.tmp") bit<32> tmp;
+    @noWarn("unused") @name(".NoAction") action NoAction_1() {
+    }
+    @name("C.RW") action RW() {
+        @name("C.local") Local<bit<32>>() local_0;
```
jafingerhut, line 14: Does this change permit multiple actions in the same control C that each declare a local extern instance with the same name `local`?  Local declarations inside of action bodies definitely seems to imply that this should be possible.

Note that adding the action name to the auto-generated `@name` annotation IS NOT ENOUGH to make that possible right now, because we also allow multiple actions with the same name to be defined in the same control.

mbudiu-vmw: two externs declarations in separate actions will be distinct objects, even if they have the same local name.

jafingerhut: The `@name` annotation "C.local" is "enclosing_control_name.local_instance_name_inside_action", yes?

I guess I will try out this PR's changes myself and find out, but what if there was another action that also had an instance with the same name `local` inside of control `C`?  Would their `@name` annotations be identical?  That should be considered a bug, shouldn't it?

jafingerhut: See the attached test program action-register2.p4, which is just a minor variation of your action-register.p4.  Look at the output program from the last FrontEnd pass.  It has two DirectRegister instances with `@name("C.local")`.
 It seems like a bug to me that p4c is intentionally generating identical `@name` annotations for two different instances.

If you do not consider that a bug, why not?

What can a control plane API generation pass do with this that is possibly useful?

jafingerhut: Attaching zip file this time.
[p4c-pr-3938-maybe-bug.zip](https://github.com/p4lang/p4c/files/11070682/p4c-pr-3938-maybe-bug.zip)

jafingerhut: Note: I am not saying that I think if this is considered a bug, that it needs to be fixed right away.  It would be best to worry about it only if we do move forward with enhancing the language with the ability to instantiate extern objects within action bodies.

mbudiu-vmw: The right solution is to also use the action name as part of the generated name.

jafingerhut: But don't we allow more than one action with the same name, too?

mbudiu-vmw: I am not sure we do. Even if we do, they can use annotations to specify different names.

testdata/p4_16_samples/action-register.p4
```diff
@@ -0,0 +1,26 @@
...
+
+extern Local<D> {
+  Local();
+  D read();
+  void write(in D data);
+}
+
+control C() {
+  action RW() {
+    Local<bit<32>>() local;
```
nvswarren, line 11: How many instantiations does this create? I assume from the following comment on the merge request:

> This shows an example that would allow the effect of Direct registers in P4.

... that there is a separate instance per P4 table entry that invokes this action, and also per each apply block call-site that directly invokes this action, rather than there being e.g. one instance or one per P4 table or anything like that. And further, there's no kind of connection between any of the instances.

mbudiu-vmw: We probably would restrict such actions to being called from tables only. At least that is the use case that is requested.
Your interpretation is the intended one otherwise.

PR #3944, opened 2023-03-27
Remove boost::filesystem from P4Testgen.

fruffy:


backends/p4tools/modules/testgen/targets/bmv2/target.cpp
```diff
@@ -77,7 +77,7 @@ const BMv2_V1ModelProgramInfo *BMv2_V1ModelTestgenTarget::initProgram_impl(
 
 Bmv2TestBackend *BMv2_V1ModelTestgenTarget::getTestBackend_impl(
     const ProgramInfo &programInfo, SymbolicExecutor &symbex,
-    const boost::filesystem::path &testPath, boost::optional<uint32_t> seed) const {
+    const std::filesystem::path &testPath, boost::optional<uint32_t> seed) const {
```
pkotikal, line 80: Can we change `boost::optional` to `std::optional` too?

pkotikal (approved):

backends/p4tools/modules/testgen/targets/bmv2/target.cpp
fruffy: Yes, the corresponding PR can be found here: https://github.com/p4lang/p4c/pull/3942

fruffy: I will also try to do this for std::variant, but that is a bit more tricky.

pkotikal: got it, is there a reason we are moving away from boost - is it to avoid extra dependencies?

fruffy: Yes, boost adds compile time and binary size overhead. It is also one of the major sources for build issues (next to protobuf). So I am trying to at least fix some low-hanging fruits.

PR #3950, opened 2023-03-30
Clean up the trace event class of P4Tools. 

fruffy:


backends/p4tools/common/lib/trace_events.cpp
pkotikal, line 1: Is this empty file? Do we still need it?

backends/p4tools/common/lib/trace_event_types.cpp
```diff
@@ -0,0 +1,303 @@
...
+#include "backends/p4tools/common/lib/model.h"
+#include "backends/p4tools/common/lib/symbolic_env.h"
+#include "backends/p4tools/common/lib/taint.h"
+#include "ir/vector.h"
+#include "lib/exceptions.h"
+#include "lib/log.h"
+#include "lib/null.h"
+#include "lib/source_file.h"
+
+namespace P4Tools::TraceEvents {
```
pkotikal, line 18: TraceEvent?

pkotikal (approved):

backends/p4tools/common/lib/trace_events.cpp
fruffy: Nice eye, thanks for spotting that. Removed.

PR #3951, opened 2023-03-30
Use a version of simple switch that uses nanomsg instead of virtual interfaces for PTF tests.

fruffy:
This PR adds support for nanomsg to the `run-bmv2-ptf-test.py` script. It does not yet add support to create a large number of ports. The behavior of the script is the same as the veth interfaces for now.


jafingerhut:
What is the motivation for such a change?  Is it that since veth interfaces are no longer required for packets into and out of simple_switch, that PTF tests can be run without root privileges?

fruffy:
> What is the motivation for such a change? Is it that since veth interfaces are no longer required for packets into and out of simple_switch, that PTF tests can be run without root privileges?

There are three advantages to this approach.
1. No root privileges required since we just create files.
2. Full support of the 511 BMv2 ports. PTF is tied to veth interfaces and creating 511 interfaces per test is overwhelming.
3. No minimum packet size. The Linux veth interfaces have strange behavior (either dropping the packet or producing garbage) if the packet is too short (do not quite remember the size, 40 bits?). This is not the case with nanomsg. 

The downside is that we are not "emulating" real network conditions anymore. However, for testing the tool chain I would say nanomsg is sufficient. And we can make this a toggle option for the script (either run veth or nanomsg).

jnfoster (approved):
Approve!

PR #3953, opened 2023-03-30
Eliminate typedefs when resolving constructor type declaration

qobilidop:
This fixes #3952.

PR #3954, opened 2023-03-31
Add documentation on adding test data

qobilidop:
As a follow-up on #3953.

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -218,6 +218,19 @@ tests are using at the link below:
 
 + [https://hub.docker.com/r/p4lang/behavioral-model/builds](https://hub.docker.com/r/p4lang/behavioral-model/builds)
 
+### Adding new test data
+
+To add a new input test with a sample P4 code file (under `testdata/p4_16_samples/` for example), one needs to:
+
+* Add the `*.p4` file to the `testdata/p4_16_samples/` directory. The file name might determine which test suite this test belongs to. Those are determined by cmake commands.
+  * For example, [any P4 file under `testdata/p4_16_samples/`](https://github.com/p4lang/p4c/blob/de2eaa085152abd0690660fbe561eaf5db7b2bf7/backends/p4test/CMakeLists.txt#L63) belongs to the [`p4` test suite](https://github.com/p4lang/p4c/blob/de2eaa085152abd0690660fbe561eaf5db7b2bf7/backends/p4test/CMakeLists.txt#L113) (meaning it will run with `make check-p4`).
+  * [File ending with `*-bmv2.p4`](https://github.com/p4lang/p4c/blob/de2eaa085152abd0690660fbe561eaf5db7b2bf7/backends/bmv2/CMakeLists.txt#L127) also belongs to the [`bmv2` test suite](https://github.com/p4lang/p4c/blob/de2eaa085152abd0690660fbe561eaf5db7b2bf7/backends/bmv2/CMakeLists.txt#L200) (meaning it will run with `make check-bmv2`).
+* Then generate reference outputs:
+  * For a frontend-only test, you can do `../backends/p4test/run-p4-sample.py . -f ../testdata/p4_16_samples/some_name.p4`. Note that this command needs to run under the `build/` directory.
+  * For a test targeting bmv2 backend, the corresponding command is `../backends/bmv2/run-bmv2-test.py`.
+  * If you have many reference outputs to add/update, you could also do `P4TEST_REPLACE=True make check` (or `make check-*`) to update all tests.
+* Now when you do testing, the new test case will be included.
+
 ## Coding conventions
 
 * Coding style is guided by the [following rules](CodingStandardPhilosophy.md).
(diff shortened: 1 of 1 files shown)


docs/README.md
```diff
@@ -218,6 +218,16 @@ tests are using at the link below:
...
 
+### Adding new test data
+
+To add a new input test with a sample P4 code file (under `testdata/p4_16_samples/` for example), one needs to:
+
+* Add the `*.p4` file to the `testdata/p4_16_samples/` directory. The file name might determine which test suite this test belongs to. Those are determined by cmake commands.
+  * For example, [any P4 file under `testdata/p4_16_samples/`](https://github.com/p4lang/p4c/blob/de2eaa085152abd0690660fbe561eaf5db7b2bf7/backends/p4test/CMakeLists.txt#L63) belongs to the [`p4` test suite](https://github.com/p4lang/p4c/blob/de2eaa085152abd0690660fbe561eaf5db7b2bf7/backends/p4test/CMakeLists.txt#L113) (meaning it will run with `make check-p4`).
+  * [File ending with `*-bmv2.p4`](https://github.com/p4lang/p4c/blob/de2eaa085152abd0690660fbe561eaf5db7b2bf7/backends/bmv2/CMakeLists.txt#L127) also belongs to the [`bmv2` test suite](https://github.com/p4lang/p4c/blob/de2eaa085152abd0690660fbe561eaf5db7b2bf7/backends/bmv2/CMakeLists.txt#L200) (meaning it will run with `make check-bmv2`).
+* Then generate expected outputs by `P4TEST_REPLACE=True make check` (or a more specific test command to save time).
```
fruffy, line 228: There is also the `-f` flag for `run-bmv2-test.py` and `run-p4-sample.py`.

qobilidop: Got it! I didn't know them before. Let me try them out and include them in the doc.

fruffy: P4TEST_REPLACE=TRUE is basically equivalent, but sometimes you have an environment where that command is awkward to use.

qobilidop: I have updated the doc.

docs/README.md
```diff
@@ -218,6 +218,19 @@ tests are using at the link below:
...
+
+* Add the `*.p4` file to the `testdata/p4_16_samples/` directory. The file name might determine which test suite this test belongs to. Those are determined by cmake commands.
+  * For example, [any P4 file under `testdata/p4_16_samples/`](https://github.com/p4lang/p4c/blob/de2eaa085152abd0690660fbe561eaf5db7b2bf7/backends/p4test/CMakeLists.txt#L63) belongs to the [`p4` test suite](https://github.com/p4lang/p4c/blob/de2eaa085152abd0690660fbe561eaf5db7b2bf7/backends/p4test/CMakeLists.txt#L113) (meaning it will run with `make check-p4`).
+  * [File ending with `*-bmv2.p4`](https://github.com/p4lang/p4c/blob/de2eaa085152abd0690660fbe561eaf5db7b2bf7/backends/bmv2/CMakeLists.txt#L127) also belongs to the [`bmv2` test suite](https://github.com/p4lang/p4c/blob/de2eaa085152abd0690660fbe561eaf5db7b2bf7/backends/bmv2/CMakeLists.txt#L200) (meaning it will run with `make check-bmv2`).
+* Then generate reference outputs:
+  * For a frontend-only test, you can do `../backends/p4test/run-p4-sample.py . -f ../testdata/p4_16_samples/some_name.p4`. Note that this command needs to run under the `build/` directory.
```
qobilidop, line 229: I'm not sure if there is a more convenient way to use this script.

fruffy: Unfortunately not for this script. `run-bmv2-test.py` has the option 
```
parser.add_argument("-bd",
                    "--buildir",
                    dest="builddir",
                    help="The path to the compiler build directory, "
                    "default is \"build\".")
```
which you can use. 
We should also add this to run-p4-sample for convenience.

docs/README.md
```diff
@@ -218,6 +218,19 @@ tests are using at the link below:
...
+  * For a test targeting bmv2 backend, the corresponding command is `../backends/bmv2/run-bmv2-test.py`.
+  * If you have many reference outputs to add/update, you could also do `P4TEST_REPLACE=True make check` (or `make check-*`) to update all tests.
+* Now when you do testing, the new test case will be included.
```
fruffy, line 232:
```suggestion
* The reference files for each test will be updated after running the tests.
```

docs/README.md
fruffy, line 229:
```suggestion
  * For a frontend-only test, you can run `../backends/p4test/run-p4-sample.py . -f ../testdata/p4_16_samples/some_name.p4`. Note that this command needs to run under the `build/` directory. The test will fail if the test output is missing or does not match with the existing reference outputs. Toggling the `-f` flag will force the script to produce new reference outputs, which can, and should be committed, along with the changes that caused the output change. 
```

fruffy (approved):

docs/README.md
qobilidop: Got it. And it seems `run-p4-sample.py` could be updated to use argparse as a first step. I can take a look and update `run-p4-sample.py` this weekend.

fruffy: That would be quite useful. However, there is no pressing need for it, so do not feel obliged!

PR #3962, opened 2023-04-03
Always insert instantiation statement at the end of the local statement list

### LINKED ISSUE #3955: Problematic inline code ordering causes use before declaration error
Update: Turns out this has nothing to do with type checking, but an issue with inline code ordering.

## Test code to reproduce the issue

When passing an extern to an inner control block, and using it in a table action, the frontend type checking would fail.

I think this should work, because according to the language spec:
- > For anything other than an action, e.g. a **control**, parser, or function, a **directionless parameter** means that the value supplied as an argument in a call must be a compile-time known value (see Section 18.1).
- 18.1
  - > The following are compile-time known values:
    - > Instances constructed by instance declarations (Section 11.3) and constructor invocations.
- 11.3
  - > Instantiations are similar to variable declarations, but are reserved for the types with constructors (**extern objects**, control blocks, parsers, and packages)

Here's an example that fails to compile (`test.p4`):
```p4
#include <core.p4>

extern MyCounter<I> {
    MyCounter(bit<32> size);
    void count(in I index);
}

typedef bit<10> my_counter_index_t;
typedef MyCounter<my_counter_index_t> my_counter_t;

control Inner(my_counter_t counter_set) {
    action count(my_counter_index_t index) {
        counter_set.count(index);
    }

    table counter_table {
        actions = {
            count;
        }
    }

    apply {
        counter_table.apply();
    }
}

control Test() {
    my_counter_t(1024) counter_set;

    apply {
        Inner.apply(counter_set);
    }
}

control C();
package P(C _c);
P(Test()) main;
```

The error message is:
```
In file: /path/to/p4c/frontends/p4/typeChecking/typeChecker.cpp:186
Compiler Bug: test.p4(28): Could not find type of <Declaration_Instance>(35172) counter_set/39 @name("counter_set") my_counter_t counter_set(1024)
    my_counter_t(1024) counter_set;
                       ^^^^^^^^^^^
```

To explore the conditions when this error would occur, I have done some experiments.

## Experiment 1

When the extern instance declaration is in the same control block as its usage, there is no issue. For example, this compiles:
```p4
#include <core.p4>

extern MyCounter<I> {
    MyCounter(bit<32> size);
    void count(in I index);
}

typedef bit<10> my_counter_index_t;
typedef MyCounter<my_counter_index_t> my_counter_t;

control Test() {
    my_counter_t(1024) counter_set;

    action count(my_counter_index_t index) {
        counter_set.count(index);
    }

    table counter_table {
        actions = {
            count;
        }
    }

    apply {
        counter_table.apply();
    }
}

control C();
package P(C _c);
P(Test()) main;
```

## Experiment 2

When the extern usage is not in a table action, there is no issue. For example, this compiles:
```p4
#include <core.p4>

extern MyCounter<I> {
    MyCounter(bit<32> size);
    void count(in I index);
}

typedef bit<10> my_counter_index_t;
typedef MyCounter<my_counter_index_t> my_counter_t;

control Inner(my_counter_t counter_set) {
    apply {
        counter_set.count(42);
    }
}

control Test() {
    my_counter_t(1024) counter_set;

    apply {
        Inner.apply(counter_set);
    }
}

control C();
package P(C _c);
P(Test()) main;
```

I'll keep debugging and try to come up with a fix. But maybe someone can spot the real issue faster than me.

qobilidop:
This fixes #3955. Many test outputs need to be updated after this change.

### DIFF
--- a/frontends/p4/directCalls.cpp
+++ b/frontends/p4/directCalls.cpp
@@ -3,15 +3,13 @@
 namespace P4 {
 
 const IR::Node *DoInstantiateCalls::postorder(IR::P4Parser *parser) {
-    insert.append(parser->parserLocals);
-    parser->parserLocals = insert;
+    parser->parserLocals.append(insert);
     insert.clear();
     return parser;
 }
 
 const IR::Node *DoInstantiateCalls::postorder(IR::P4Control *control) {
-    insert.append(control->controlLocals);
-    control->controlLocals = insert;
+    control->controlLocals.append(insert);
     insert.clear();
     return control;
 }
--- a/testdata/p4_16_errors_outputs/list-error.p4
+++ b/testdata/p4_16_errors_outputs/list-error.p4
@@ -15,4 +15,3 @@ control c() {
 control C();
 package top(C _c);
 top(c()) main;
-
--- a/testdata/p4_16_errors_outputs/psa-meter2.p4
+++ b/testdata/p4_16_errors_outputs/psa-meter2.p4
@@ -67,8 +67,5 @@ control MyED(packet_out buffer, out EMPTY a, out EMPTY b, inout EMPTY c, in EMPT
 }
 
 IngressPipeline(MyIP(), MyIC(), MyID()) ip;
-
 EgressPipeline(MyEP(), MyEC(), MyED()) ep;
-
 PSA_Switch(ip, PacketReplicationEngine(), ep, BufferingQueueingEngine()) main;
-
--- a/testdata/p4_16_samples/extern-inst-as-param.p4
+++ b/testdata/p4_16_samples/extern-inst-as-param.p4
@@ -1,0 +1,37 @@
+#include <core.p4>
+
+extern MyCounter<I> {
+    MyCounter(bit<32> size);
+    void count(in I index);
+}
+
+typedef bit<10> my_counter_index_t;
+typedef MyCounter<my_counter_index_t> my_counter_t;
+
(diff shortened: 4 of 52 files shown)


qobilidop:
I used `P4TEST_REPLACE=True make check` to update all test outputs. I'm still in the process of manually reading and checking these changes.

qobilidop:
Ahh, I just see #821 that does the opposite change.

mbudiu-vmw:
> I used `P4TEST_REPLACE=True make check` to update all test outputs. I'm still in the process of manually reading and checking these changes.

The gauntlet tool does some of that, by I am not sure whether it will check such early passes. 
(It performs translation validation, making sure that the program at two specific program points has the same semantics.)

mbudiu-vmw (approved):

fruffy:
> The gauntlet tool does some of that, by I am not sure whether it will check such early passes.

Gauntlet checks all passes of the front and mid end of p4test. That also includes type checking etc. So the inlining pass is checked. We actually have an issue open for that https://github.com/p4lang/p4c/issues/2470. 

The only thing it relies on is the parser. An unfortunate side effect of that is that it is hard to maintain :disappointed:.

qobilidop:
Just wonder if this PR needs more reviews or is it ready to be merged? I would appreciate it a lot if it could be merged soon, because I have some P4 code at work pending on this to compile. Thanks a lot for all reviewers' time!

PR #3968, opened 2023-04-06
Implement support for default initializers (...)

mbudiu-vmw:
This finally implements support for default initializers in the compiler. These have been in the spec for a while: https://p4.org/p4-spec/docs/P4-16-v-1.2.3.html#sec-initializing-with-default-values

The algorithm is in principle very simple: when you see an expression like `{ a, ... }` you use type unification to figure out what the type of "..." should be. Then the DefaultValues pass generates literals for each of these possible types. 

Note that even this PR does not yet support generic structs/headers/unions/stacks. It may be easy to do that, but we'll do it in a separate PR. It may be doable by just moving the DefaultValues pass after type specialization.

We have introduced a new type Type_Any which can unify with anything. It is the type of the "..." expression.
We have also introduced a new Type_Partial, which wraps fragments of another type. For example, the `...` used in a StructExpression has a type Type_Partial(Type_Struct()), where the struct has exactly the fields that `...` should cover.


frontends/p4/defaultValues.cpp
```diff
@@ -0,0 +1,166 @@
...
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+#include "defaultValues.h"
+
+namespace P4 {
+
+const IR::Expression *DoDefaultValues::defaultValue(const IR::Expression *expression,
```
fruffy, line 21: Possible to make this a general utility function? There is also some overlap with https://github.com/p4lang/p4c/blob/main/ir/irutils.h. We can merge some of this.

mbudiu-vmw: This is tied to the https://p4.org/p4-spec/docs/P4-16-v-1.2.3.html#sec-default-values default values section of the spec. Where else is this function useful?

ir/type.def
```diff
@@ -62,6 +62,42 @@ inline bool operator>>(cstring s, IR::Direction &d) {
 }
 #end
 
+/// This represents a type that can unify with any other type.
+/// It is the original type of Dots (...) expressions.
+/// There is no syntax to represent this type.
+/// Treated like a type variable so that unification can assign it a value.
+class Type_Any : Type, ITypeVar {
```
fruffy, line 69: What is the difference between this and Type_DontCare?

mbudiu-vmw: There is a literal for the don't care type, written as _. The unification rules are also different, don't care doesn't need to be unified, whereas ANY should always be resolved to a different, concrete type.

frontends/p4/defaultValues.cpp
mbudiu-vmw: I have factored out the defaultValue function as a static method of this class.
Hopefully it is what you need.

ir/type.def
mbudiu-vmw: Yes, the unification rules for Don't care are different. I thought about it.
But I think that having an extra class with a clear meaning is much simpler than overloading the meaning of the existing class and then reasoning about the different kinds of uses.
This way we are also sure we don't break anything that worked so far.

fruffy (approved):

frontends/p4/defaultValues.cpp
```diff
@@ -0,0 +1,171 @@
...
+        ::error(ErrorType::ERR_UNSUPPORTED, "%1% default values for varbit types", srcInfo);
+        return nullptr;
+    } else if (auto ht = type->to<IR::Type_Header>()) {
+        return new IR::InvalidHeader(ht->getP4Type());
+    } else if (auto hu = type->to<IR::Type_HeaderUnion>()) {
+        return new IR::InvalidHeaderUnion(hu->getP4Type());
+    } else if (auto st = type->to<IR::Type_StructLike>()) {
+        auto vec = new IR::IndexedVector<IR::NamedExpression>();
+        for (auto field : st->fields) {
+            auto value = defaultValue(srcInfo, field->type);
```
fruffy, line 49: What happens here if `field->type` is a Type_Name? Is this resolved beforehand somehow? If so, how?

mbudiu-vmw: It really depends whether that is the canonical type (from the typeMap) or not.
If it's the canonical type there should never be a Type_Name.
But it does make the point that the previous version of this function is not equivalent to the current one.
This will crash rather than be wrong, so let's leave it like this for now.

fruffy: How do you retrieve the canonical type? `getTypeType` or `getType`? 
I would like to use this function so I have to make sure the input type is canonical.

mbudiu-vmw: When you get a type from the typeMap it is canonical, and so are all it's subfields.
if you get a type from the program it is probably not canonical.
A canonical type can be converted to a program type with getP4Type().

PR #3970, opened 2023-04-06
Pull in libbpf only when the eBPF back end is active. Pull libbpf using FetchContent instead of a custom script. 

fruffy:
We have libbpf listed as a submodule dependency but it is only really necessary when the eBPF back end is active. Only pull the dependency conditionally when the eBPF back end is active.


jafingerhut (approved):
I am not the ideal reviewer for in-depth details of p4c's build system, but I am generally in favor of the transition to using CMake FetchContent that has been developed recently, and the tests are all passing, so I will throw my approval in the ring.

PR #3974, opened 2023-04-07
[P4Testgen] Implement meter support for the BMv2 V1model PTF test back end

fruffy:
Implement P4Testgen BMv2 support for meters and direct meters.

- Add an interface to the `base_test.py` file to configure meters.
- Also add a helper function to the PTF test back end to make sure we are able to set the color we want. For green, we set the thresholds to the maximum, for red we set the thresholds to the minimum, for yellow we use a mix.
- direct meters can only be used if they are attached to a table and the table has an entry. Implement a new pass that creates a direct meter <-> table mapping. Look up this mapping when executing the extern. If there is no table entry in the table associated with the meter, skip meter execution since we can not configure it.

### DIFF
--- a/backends/bmv2/base_test.py
+++ b/backends/bmv2/base_test.py
@@ -3,8 +3,6 @@
 # repository. This file in turn was a modified version of
 # https://github.com/p4lang/PI/blob/ec6865edc770b42f22fea15e6da17ca58a83d3a6/proto/ptf/base_test.py.
 
-from collections import Counter
-from functools import wraps, partialmethod
 import logging
 import queue
 import re
@@ -12,7 +10,7 @@ import sys
 import threading
 import time
 from collections import Counter
-from functools import partial, wraps
+from functools import partialmethod, wraps
 from pathlib import Path
 
 import google.protobuf.text_format
@@ -29,6 +27,7 @@ FILE_DIR = Path(__file__).resolve().parent
 TOOLS_PATH = FILE_DIR.joinpath("../../tools")
 sys.path.append(str(TOOLS_PATH))
 import testutils
+
 
 def stringify(n, length=0):
     """Take a non-negative integer 'n' as the first parameter, and a
@@ -52,6 +51,7 @@ def stringify(n, length=0):
             length = n_size_bytes
     s = n.to_bytes(length, byteorder="big")
     return s
+
 
 # Used to indicate that the gRPC error Status object returned by the server has
 # an incorrect format.
@@ -267,8 +267,14 @@ class P4RuntimeTest(BaseTest):
     def import_p4info_names(self):
         suffix_count = Counter()
         for obj_type in [
-                "tables", "action_profiles", "actions", "counters", "direct_counters",
-                "controller_packet_metadata", "meters", "direct_meters",
+            "tables",
+            "action_profiles",
+            "actions",
+            "counters",
+            "direct_counters",
+            "controller_packet_metadata",
+            "meters",
+            "direct_meters",
         ]:
             for obj in getattr(self.p4info, obj_type):
                 pre = obj.preamble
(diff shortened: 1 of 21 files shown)


backends/bmv2/base_test.py
```diff
@@ -982,6 +899,25 @@ def counter_dump_data(self, counter_name, direct=False):
...
 
+    def meter_write(self, meter_name, index, meter_config, direct):
+        req = self.get_new_write_request()
+        update = req.updates.add()
+        update.type = p4runtime_pb2.Update.MODIFY
+        entity = update.entity
+        if direct:
+            meter_write = entity.direct_meter_entry
+            meter_obj = self.get_obj("direct_meters", meter_name)
+            meter_write.table_entry.table_id = meter_obj.direct_table_id
```
fruffy, line 910: @antoninbas I have a question about direct_meter writes. I am trying to use the p4runtime API to craft an update message, but I am unable to get this  particular write to work. The message I am seeing is 
`* At index 0: INVALID_ARGUMENT, 'Missing non-ternary field in match key'`.

I am not exactly sure how to craft a match key here that makes sense. I tried various things but keep getting the same message. Have not found a similar example, other operations on direct externs seem to be reads.   Any advice?

antoninbas: You are only setting the `table_id` field in `meter_write.table_entry`. This is not enough, every table entry has its own meter entry. You need to provide the full match key, like you would if you were doing a `MODIFY` on a `TableEntry`.

You need to be able to call `set_match_key` here, for the direct case: the index is replaced by a match key.

fruffy: I see, so the direct_meter match key needs to match exactly with an existing, already installed match key?

antoninbas: For a write operation, yes.
The direct meter entry can also be configured at the time of inserting the table entry, to avoid 2 RPCs: https://github.com/p4lang/p4runtime/blob/90553b90a12ead5c19700e7fef21164dea5b6d22/proto/p4/v1/p4runtime.proto#L191

fruffy: Thanks! I got something to work. 
There is one edge case that is tricky. I am guessing the direct meter always requires an entry? For example, if you have this construct:
```
table meter_table {
    actions = {
        meter_assign;
    }
    key = {
        h.eth_hdr.dst_addr: exact;
    }
    size = 16384;
    default_action = meter_assign();
    meters = table_attached_meter;
}
```
where meter_assign is executed as default action, one might have no entry on the table. I am guessing this case is not supported? I tried abusing the API with `is_default_action` but it did not work.

antoninbas: P4Runtime & PI support direct resources for the default table entry, but bmv2 does not.
When attempting this, an error should be returned and the simple_switch_grpc logs should read: `[PI] bmv2 does not support direct meters for default entries`.

fruffy: Thanks, that was quite helpful.

backends/p4tools/modules/testgen/targets/bmv2/test_spec.cpp
```diff
@@ -8,73 +8,123 @@
...
+cstring Bmv2V1ModelMeterValue::getObjectName() const { return "Bmv2V1ModelMeterValue"; }
 
-const Bmv2RegisterCondition *Bmv2RegisterCondition::evaluate(const Model &model) const {
-    const auto *evaluatedIndex = model.evaluate(index);
-    const auto *evaluatedValue = model.evaluate(value);
-    return new Bmv2RegisterCondition(evaluatedIndex, evaluatedValue);
+const Bmv2V1ModelMeterValue *Bmv2V1ModelMeterValue::evaluate(const Model &model) const {
+    const auto *evaluatedValue = model.evaluate(getInitialValue());
+    auto *evaluatedMeterValue = new Bmv2V1ModelMeterValue(evaluatedValue, isDirect);
+    const std::vector<ActionArg> evaluatedConditions;
```
pkotikal, line 118: Is `evaluatedConditions` unused?

backends/p4tools/modules/testgen/targets/bmv2/test_spec.cpp
```diff
@@ -8,73 +8,123 @@
...
+
+Bmv2V1ModelRegisterValue::Bmv2V1ModelRegisterValue(const IR::Expression *initialValue)
+    : IndexMap(initialValue) {}
+
+cstring Bmv2V1ModelRegisterValue::getObjectName() const { return "Bmv2V1ModelRegisterValue"; }
+
+const Bmv2V1ModelRegisterValue *Bmv2V1ModelRegisterValue::evaluate(const Model &model) const {
+    const auto *evaluatedValue = model.evaluate(getInitialValue());
+    auto *evaluatedRegisterValue = new Bmv2V1ModelRegisterValue(evaluatedValue);
     const std::vector<ActionArg> evaluatedConditions;
```
pkotikal, line 97: Same here.

backends/p4tools/modules/testgen/targets/bmv2/backend/ptf/ptf.h
```diff
@@ -47,9 +47,12 @@ class PTF : public TF {
                     float currentCoverage) override;
 
  private:
+    /// @returns the static preamble string used by inja to serialize the test preamble.
+    static std::string getPreamble();
```
pkotikal, line 51: Is this still needed?

pkotikal (approved):
Looks good, thank you!

PR #3976, opened 2023-04-10
[P4Testgen] Clean up the implementation of the BMv2 clone externs. 

### LINKED ISSUE #3863: P4Testgen does not support sequential clone calls in BMv2.
Subsequent BMv2 clone extern call override the session ID. Only the last clone call is considered valid. P4Testgen does not model this correctly and instead assumes that each call creates an independent output packet.

fruffy:
Fixes  #3863.

- Calling clone again in the ingress control will override any previous clone calls. We were not modelling this behavior. Now, we use a clone info test object which contains the execution state at the time the clone call was made. This clone info object is converted into a clone info spec after the deparser. 
- We were also not respecting the session ID. If we are able to control the session ID, we have to limit ourselves to a particular range of values. 
- Introduce a weak notion of a "traffic manager" extern, which handles calls such as recirculate or clone.
- Refactor the clone class in BMV2's expression stepper, add more comments, and clean up the code.
- Factor the test object class out of the test objects file and introduce a test object map.
- Various little fixes here and there. 

With this pull request, `pins_middleblock.p4` should be testable without any errors.

### DIFF
--- a/testdata/p4_16_samples/pins/pins_fabric.p4
+++ b/testdata/p4_16_samples/pins/pins_fabric.p4
@@ -372,7 +372,7 @@ control routing(in headers_t headers, inout local_metadata_t local_metadata, ino
         local_metadata.route_metadata = route_metadata;
     }
     @id(0x0100000F) action trap() {
-        clone(CloneType.I2E, 511);
+        clone(CloneType.I2E, 1024);
         mark_to_drop(standard_metadata);
     }
     @id(0x01000015) action set_metadata_and_drop(@id(1) route_metadata_t route_metadata) {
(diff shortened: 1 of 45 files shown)


testdata/p4_16_samples/pins/pins_fabric.p4
```diff
@@ -372,7 +372,7 @@ control routing(in headers_t headers, inout local_metadata_t local_metadata, ino
         local_metadata.route_metadata = route_metadata;
     }
     @id(0x0100000F) action trap() {
-        clone(CloneType.I2E, 1024);
```
fruffy, line 375: @smolkaj  Just a quick clarification question. Is the session ID value here intentional? I had to change it to 511 because the open-source P4Runtime server [only allows session IDs numbering lower than 512](https://github.com/p4lang/PI/blob/4961fb9fb7a03b8fe7f1511f05fc7a0238b1d88c/proto/frontend/src/pre_clone_mgr.h#L86).

fruffy: Ah great, that makes things easier. Do you know whether or when there is a refresh coming to the files at 
https://github.com/sonic-net/sonic-pins/tree/main/sai_p4/instantiations/google 
?

smolkaj: @rhalstea do you know the answer?

smolkaj: Generally we would be happy to have our up-to-date programs upstreamed to p4c, we should just keep a copyright notice.

rhalstea: We usually only update the programs in https://github.com/sonic-net/sonic-pins/tree/main/sai_p4/instantiations/google when we're upstreaming new features into a SONiC release which have around a 6 month cadence. However, we don't always upstream new P4 features each release.

If we want to have up-to-date programs I like @smolkaj's suggestion of p4c. The copies in sonic-pins need to be kept in sync with the P4RT App so as to not break the component tests, and that dependency would be the bottleneck.

rhalstea: I guess I didn't explicitly answer @fruffy's question about the next refresh ... It won't we anytime soon. The feature's we're targeting for the May SONiC release don't require a P4 program updates, and we haven't decided on the features to upstream for the next November release.

fruffy: Thanks, that makes sense. In the meantime we can fix issues manually until the next refresh. Or we could update out-of-band (directly from the private repo to p4c's versions) , if necessary.  Currently, there is no pressing need.

fruffy:
@nikil21 @abhinav2sri

This will make the implementation of recirculation functions easier.

pkotikal (approved):

PR #3796, opened 2022-12-19
Fix bazel version to 5.4.0

smolkaj:
This is to work around breaking changes introduced by Bazel 6: bazelbuild/bazel#8622 (comment)

Fixes https://github.com/p4lang/p4c/issues/3795

@fruffy

PR #3979, opened 2023-04-11
Bump Bazel dependency versions to enable using latest Bazel

qobilidop:
The context is in #3796. I did some further debugging and found the compatibility issues come from dependencies. So updating dependencies solves the issue without having to constrain the Bazel version.


smolkaj:
Nice! Thanks for figuring this out.
LGTM provided that CI passes.

bazel/p4c_deps.bzl
```diff
@@ -43,9 +43,9 @@ filegroup(
         git_repository(
             name = "com_github_p4lang_p4runtime",
             remote = "https://github.com/p4lang/p4runtime",
-            # Newest commit on main branch as of Jan 22, 2021.
-            commit = "0d40261b67283999bf0f03bd6b40b5374c7aebd0",
-            shallow_since = "1611340571 -0800",
+            # Newest commit on main branch as of April 11, 2023.
+            commit = "90553b90a12ead5c19700e7fef21164dea5b6d22",
```
fruffy, line 47: This P4Runtime version is now very different from the one the compiler actually uses. 
https://github.com/p4lang/p4runtime/tree/7a322f35f0c80bf20bc7fcc96f9d1ab77e5fd07a
So the open-source framework will not test that version.

fruffy (approved):

bazel/p4c_deps.bzl
smolkaj: Just to clarify -- are you saying p4c uses different versions of P4Runtime in the cbuild vs bazel build? Or something else?

fruffy: Yes, if the version in the `p4c_deps` file is what is used to generate the protos etc.

qobilidop: For the bazel build alone, I guess the major user is Google. And internally at Google, we do build p4c together with the latest version of its dependencies including p4runtime. So I think using the latest version here makes sense.

But I do worry there could be potential issues, maybe uncaught yet, that stems from this version difference.

Also ping @lzhzero for reference.

fruffy: You can try to update the P4Runtime version for the compiler, too. But I think there might be some complications because of the protobuf version.

qobilidop: Sure, I can try that in the future. Good to know the complications.

smolkaj: Thanks for raising that, Fabian.

Thus far we haven't been super careful about keeping these versions aligned with the cbuild, which is the source of truth. Agreed that there is a danger that this could cause issues, but thus far we haven't seen any. So personally, I am okay with winging things for now -- we can refine our approach once we hit an issue. Wdyt?

fruffy: Yeah I have no problem with that. Let me know when you want this merged. 

The only reason why the open-source P4C version is not updated is because of the Protobuf dependency. It seems, to get a newer version of Protobuf, we need to build it from scratch. That adds an unacceptable overhead. I have not found a good alternative to the Ubuntu packaged version.

smolkaj: Ah, yes, Google likes building its libraries from scratch, so there may not be a good alternative.

We can go ahead and merge this from my side. @qobilidop ?

qobilidop: I have no issues with merging this.

PR #3990, opened 2023-04-25
[P4Testgen] More StateVariable refactoring. Allow PathExpression to be state variables. 

fruffy:
- StateVariables can now be both PathExpressions or Members. This gets rid of unnecessary conversions we had to do. This change is also a setup for more simplifications to the way we handle references in the interpreter. The goal is to get rid of IR::Member or IR::PathExpression completely. 
- Create a `table_utils` file in `common`. This file contains general utility functions for table processing, which are not tied to the interpreter functionality.
- Move the `constants.h` to `common`. It broadly defines constants from the P4 specification. 
- Remove some unnecessary variable declarations from the table stepper. They were never actually used.

### DIFF
--- a/backends/p4tools/common/lib/table_utils.cpp
+++ b/backends/p4tools/common/lib/table_utils.cpp
@@ -1,0 +1,82 @@
+#include "backends/p4tools/common/lib/table_utils.h"
+
+namespace P4Tools::TableUtils {
+
+void checkTableImmutability(const IR::P4Table *table, TableProperties &properties) {
+    bool isConstant = false;
+    const auto *entriesAnnotation = table->properties->getProperty("entries");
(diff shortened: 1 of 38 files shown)


backends/p4tools/common/lib/table_utils.cpp
```diff
@@ -0,0 +1,82 @@
+#include "backends/p4tools/common/lib/table_utils.h"
```
mbudiu-vmw, line 1: no copyright?

mbudiu-vmw (approved):

backends/p4tools/common/lib/table_utils.cpp
fruffy: None of the p4tools file have a copyright, but it is a good point. Something we should add.

Issue #3989, opened 2023-04-25
irgenerator crash during build


outpaddling:
This is not a high priority, as I'm just tinkering with the idea of creating a FreeBSD port out of sheer curiosity.  A colleague is working with P4 so I decided to check out the build process.  Reporting this in the hopes that it might help you improve the code.  If you have any ideas what might be causing the seg fault, I'd be happy to test fixes.  I tried building with gcc12 instead of clang, and tweaked optimization levels as a shot in the dark.  No difference.

Tail of the build output.

`[ 13% 49/353] : && /usr/bin/c++ -pipe -g -fstack-protector-strong -fno-strict-aliasing -fdiagnostics-color=never  -Wall -Wextra -Wno-overloaded-virtual -Wno-deprecated -Wno-deprecated-declarations -pedantic -Wno-gnu-zero-variadic-macro-arguments -pipe -g -fstack-protector-strong -fno-strict-aliasing  -Wall -Wextra -Wno-overloaded-virtual -Wno-deprecated -Wno-deprecated-declarations -pedantic -Wno-gnu-zero-variadic-macro-arguments -lexecinfo -fstack-protector-strong -fuse-ld=lld tools/ir-generator/CMakeFiles/irgenerator.dir/generator.cpp.o tools/ir-generator/CMakeFiles/irgenerator.dir/irclass.cpp.o tools/ir-generator/CMakeFiles/irgenerator.dir/methods.cpp.o tools/ir-generator/CMakeFiles/irgenerator.dir/type.cpp.o tools/ir-generator/CMakeFiles/irgenerator.dir/ir-generator.cpp.o -o tools/ir-generator/irgenerator  -Wl,-rpath,/usr/local/lib  lib/libp4ctoolkit.a  -pthread  /usr/local/lib/libboost_iostreams.so  /usr/local/lib/libboost_regex.so  /usr/local/lib/libgc.so  /usr/local/lib/libgccpp.so  -lrt  -pthread && :
[...]

PR #3991, opened 2023-04-26
FreeBSD patches

outpaddling:
Minor patches for building on FreeBSD.

### DIFF
--- a/frontends/common/parser_options.cpp
+++ b/frontends/common/parser_options.cpp
@@ -21,6 +21,7 @@ limitations under the License.
 
 #include <sys/stat.h>
 #include <sys/types.h>
+#include <sys/wait.h>
 
 #include <regex>
 #include <unordered_set>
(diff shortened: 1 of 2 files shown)


frontends/common/parser_options.cpp
```diff
@@ -21,6 +21,7 @@ limitations under the License.
 
 #include <sys/stat.h>
 #include <sys/types.h>
+#include <sys/wait.h>
```
mbudiu-vmw, line 24: what do we need from sys/wait.h?

mbudiu-vmw (approved):

frontends/common/parser_options.cpp
apinski-cavium: The include added to test/gtest/helpers.cpp is for the same reason too.

fruffy:
Prompted by #3989.

PR #3993, opened 2023-04-27
Add some more type information to the strength reduction pass.

fruffy:
StrengthReduction was missing some type information, which could lead to problems (e.g., instances of `Type_Unknown`) when using [optimizeExpression](https://github.com/p4lang/p4c/blob/main/frontends/p4/optimizeExpressions.h). 

Try to add some more type information, which can be derived from the input expressions.

### DIFF
--- a/frontends/p4/strengthReduction.cpp
+++ b/frontends/p4/strengthReduction.cpp
@@ -230,14 +230,14 @@ const IR::Node *DoStrengthReduction::postorder(IR::Mul *expr) {
     if (isOne(expr->right)) return expr->left;
     auto exp = isPowerOf2(expr->left);
     if (exp >= 0) {
-        auto amt = new IR::Constant(expr->left->type, exp);
-        auto sh = new IR::Shl(expr->srcInfo, expr->right->type, expr->right, amt);
+        auto amt = new IR::Constant(exp);
+        auto sh = new IR::Shl(expr->srcInfo, expr->right, amt);
         return sh;
     }
     exp = isPowerOf2(expr->right);
     if (exp >= 0) {
-        auto amt = new IR::Constant(expr->right->type, exp);
-        auto sh = new IR::Shl(expr->srcInfo, expr->left->type, expr->left, amt);
+        auto amt = new IR::Constant(exp);
+        auto sh = new IR::Shl(expr->srcInfo, expr->left, amt);
         return sh;
     }
     if (hasSideEffects(expr)) return expr;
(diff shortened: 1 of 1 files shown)


frontends/p4/strengthReduction.cpp
```diff
@@ -80,7 +80,8 @@ const IR::Node *DoStrengthReduction::postorder(IR::BAnd *expr) {
     if (isAllOnes(expr->right)) return expr->left;
     auto l = expr->left->to<IR::Cmpl>();
     auto r = expr->right->to<IR::Cmpl>();
-    if (l && r) return new IR::Cmpl(new IR::BOr(expr->srcInfo, l->expr, r->expr));
+    if (l && r)
```
mbudiu-vmw, line 83: The field `type` of expressions very rarely has the correct value.
But I guess there is no harm copying a wrong value.

frontends/p4/strengthReduction.cpp
```diff
@@ -299,9 +302,13 @@ const IR::Node *DoStrengthReduction::postorder(IR::Mux *expr) {
         return expr->e0;
     else if (isFalse(expr->e1) && isTrue(expr->e2))
         return new IR::LNot(expr->e0);
-    else if (auto lnot = expr->e0->to<IR::LNot>())
-        return new IR::Mux(lnot->expr, expr->e2, expr->e1);
-    else if (!hasSideEffects(expr) && expr->e1->equiv(*expr->e2))
+    else if (const auto *lnot = expr->e0->to<IR::LNot>()) {
+        expr->e0 = lnot->expr;
```
mbudiu-vmw, line 306: I find the original much more readable

mbudiu-vmw (approved):

frontends/p4/strengthReduction.cpp
fruffy: Assume you ran TypeChecking before applying this pass, the type should be correct, no?

frontends/p4/strengthReduction.cpp
fruffy: But it does incur another copy and does not preserve source information and type.

frontends/p4/strengthReduction.cpp
mbudiu-vmw: Only if you pass a specific flag to the typeChecking pass.
The problem is that adding types to the IR nodes changes the nodes (which are immutable), and thus invalidates any maps you may have built.

PR #3997, opened 2023-04-28
[P4Testgen] Fix include paths

### LINKED ISSUE #3996: Testgen does not find "buildin" include path (<core.p4>, etc.)
Hi! While `p4c` itself does use the default p4include without specifying path to it, p4testgen apparently does not:

```bash
$ ./p4c --target bmv2 --std p4-16 --arch v1model ../testdata/p4_16_bmv_errors/issue1167-bmv2.p4
# ok, compilation fail, probably expected for this test
$ ./p4testgen --target bmv2 --std p4-16 --arch v1model ../testdata/p4_16_bmv_er
rors/issue1167-bmv2.p4  --test-backend stf
../testdata/p4_16_bmv_errors/issue1167-bmv2.p4:17:19: error: no include path in which to search for core.p4
   17 | #include <core.p4>
      |                   ^
../testdata/p4_16_bmv_errors/issue1167-bmv2.p4:18:22: error: no include path in which to search for v1model.p4
   18 | #include <v1model.p4>
      |
$ ./p4testgen --target bmv2 --std p4-16 --arch v1model ../testdata/p4_16_bmv_errors/issue1167-bmv2.p4  --test-backend stf -I../p4include/
# OK, testgen runs
```

... all of this was running in a clean build of p4c with `-DENABLE_TEST_TOOLS` and no other options flipped (compared to bootstrap).

This is broken in the current f20d87ae9af1508dd8e781dd48d6553cc3c6d568, but it worked without the `-I` option in 574ea01ddbfc5c5684f03d6b01bc98da743a5385. Apparently this is a problem only in a new clean build. So this looks like a configuration error. Building 574ea01ddbfc5c5684f03d6b01bc98da743a5385 first and than updating to f20d87ae9af1508dd8e781dd48d6553cc3c6d568 the problem does not occur.

Fabian, @fruffy, could you please have a look?

fruffy:
Fixes #3996.

### DIFF
--- a/backends/p4tools/modules/testgen/CMakeLists.txt
+++ b/backends/p4tools/modules/testgen/CMakeLists.txt
@@ -116,8 +116,8 @@ add_custom_target(
   linkp4testgen
   # Add some convenience links for invoking p4testgen.
   # TODO: How to best handle includes?
-  COMMAND ${CMAKE_COMMAND} -E create_symlink ${CMAKE_CURRENT_BINARY_DIR}/p4testgen ${CMAKE_BINARY_DIR}/p4testgen
-  COMMAND ${CMAKE_COMMAND} -E create_symlink ${P4C_BINARY_DIR_PATH_REL}/p4include ${CMAKE_CURRENT_BINARY_DIR}/p4include
+  COMMAND ${CMAKE_COMMAND} -E create_symlink ${P4C_BINARY_DIR}/p4testgen ${CMAKE_BINARY_DIR}/p4testgen
+  COMMAND ${CMAKE_COMMAND} -E create_symlink ${P4C_BINARY_DIR}/p4include ${CMAKE_CURRENT_BINARY_DIR}/p4include
   DEPENDS update_includes
 )
 
--- a/backends/p4tools/modules/testgen/CMakeLists.txt
+++ b/backends/p4tools/modules/testgen/CMakeLists.txt
@@ -116,7 +116,7 @@ add_custom_target(
   linkp4testgen
   # Add some convenience links for invoking p4testgen.
   # TODO: How to best handle includes?
-  COMMAND ${CMAKE_COMMAND} -E create_symlink ${P4C_BINARY_DIR}/p4testgen ${CMAKE_BINARY_DIR}/p4testgen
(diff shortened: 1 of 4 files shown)


jnfoster (approved):
A bit of a bandaid, but OK for now.

fruffy:
@vlstill Does this fix your problem?

backends/p4tools/modules/testgen/CMakeLists.txt
```diff
@@ -114,7 +114,9 @@ target_link_libraries(
 add_custom_target(
   linkp4testgen
   # Add some convenience links for invoking p4testgen.
+  # TODO: How to best handle includes?
   COMMAND ${CMAKE_COMMAND} -E create_symlink ${CMAKE_CURRENT_BINARY_DIR}/p4testgen ${CMAKE_BINARY_DIR}/p4testgen
+  COMMAND ${CMAKE_COMMAND} -E create_symlink ${P4C_BINARY_DIR_PATH_REL}/p4include ${CMAKE_CURRENT_BINARY_DIR}/p4include
```
vlstill, line 119: Interestingly, this does not work for me (maybe because I have p4c embedded in other project, although it worked the recent testgen changes...). The creates symlink leads to `/p4include`, suggesting that `${P4C_BINARY_DIR_PATH_REL}` is not set. As far as I can see, the variable is set in `ebfp`, `bmv2` and `p4test` backends, but not in `p4tools`. It would probably need to be set in this file to work correctly.

backends/p4tools/modules/testgen/CMakeLists.txt
```diff
@@ -114,7 +114,9 @@ target_link_libraries(
 add_custom_target(
   linkp4testgen
   # Add some convenience links for invoking p4testgen.
+  # TODO: How to best handle includes?
   COMMAND ${CMAKE_COMMAND} -E create_symlink ${CMAKE_CURRENT_BINARY_DIR}/p4testgen ${CMAKE_BINARY_DIR}/p4testgen
```
vlstill, line 118: This is not related to the original issue, but I've just noticed you are linking `p4testgen` to top-level project, even in `p4c` is embedded into a downstream project using it.

I would suggest symlinking `p4testgen` into `${P4C_BINARY_DIR}` instead is a better choice (even though it can break some downstream workflow). This will leave the downstream projects to do the symlinking on their level if they so choose.

This is just a thought for future cleanup.

PR #3999, opened 2023-05-01
[P4Testgen] Also include table entries in the coverage set.

fruffy:
- Generalize coverage in P4Testgen. Now we can not only cover statements, but all other IR nodes. In this case, we add table entry coverage. 
- We introduce an option to specify the type of coverage. This way we can precisely control which IR nodes to target with our test case generation. These coverage options can be stacked (i.e., we can track both statement and table entry coverage). By default, we do not track coverage at all.
- We improve the forward-looking coverage visitor to also resolve method calls, which makes the coverage analysis more accurate.
 
##### TODO

The last thing to add is caching for the `CoverableNodesScanner`. The coverable nodes remain constant for many IR nodes we look up. We should just save the computation we have done for a particular node to save computation cost.
I will add this cache in a subsequent PR.


jnfoster (approved):
LGTM

PR #4002, opened 2023-05-03
Make P4-16 the default parser option.

fruffy:
P4-14 is effectively deprecated. Reflect this in the default settings for the options parser.


jnfoster (approved):

mbudiu-vmw (approved):
Will this break build systems that customers may have?
I personally have no problem with this, but I am not the user.

fruffy:
> Will this break build systems that customers may have? I personally have no problem with this, but I am not the user.

I believe most back ends already set the version like this: https://github.com/p4lang/p4c/blob/main/backends/p4test/p4test.cpp#L109
In that case, there will not be any impact. 

If things break, they will hopefully notify us. :)

PR #4003, opened 2023-05-03
Allowing local_copyprop to propagate MethodCallExpression into table …

hanw:
…key, for example:
```
bool is_valid = hdr.ethernet.isValid();
table t {
   key = { is_valid : exact };
   ...
}
```
is transformed to 
```
table t {
    keys = { hdr.ethernet.isValid() : exact };
    ...
}
```
which saves resource reserved for the boolean variable.

### DIFF
--- a/midend/local_copyprop.cpp
+++ b/midend/local_copyprop.cpp
@@ -655,11 +655,30 @@ void DoLocalCopyPropagation::apply_table(DoLocalCopyPropagation::TableInfo *tbl)
                                                                                   << var->val);
                     var->live = true;
                 }
+            } else if (var->val && lvalue_out(var->val)->is<IR::MethodCallExpression>()) {
+                if (hasSideEffects(lvalue_out(var->val))) {
+                    LOG3("  cannot propagate expression with side effect into table key "
+                         << vname << ": " << var->val);
+                    var->live = true;
+                } else if (tbl->apply_count > 1 && (!tbl->key_remap.count(vname) ||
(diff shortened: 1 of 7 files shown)


midend/local_copyprop.cpp
```diff
@@ -640,7 +640,8 @@ void DoLocalCopyPropagation::apply_table(DoLocalCopyPropagation::TableInfo *tbl)
     for (auto key : tbl->keyreads) {
         forOverlapAvail(key, [&remaps_seen, key, tbl, this](cstring vname, VarInfo *var) {
             remaps_seen.insert(vname);
-            if (var->val && lvalue_out(var->val)->is<IR::PathExpression>()) {
+            if (var->val && (lvalue_out(var->val)->is<IR::PathExpression>() ||
+                             lvalue_out(var->val)->is<IR::MethodCallExpression>())) {
```
mbudiu-vmw, line 644: This will propagate method calls that can have side-effects.
That may be problematic.
isValid() is really a special case.

hanw: I modified the PR to prevent method calls with side effect to propagate into table keys.

mbudiu-vmw:
No tests?

hanw:
> No tests?

Added a test, thanks for the reminder.

mbudiu-vmw (approved):
There may be changes in some reference outputs.

hanw:
@mbudiu-vmw How should I handle header union in this PR? Does header union isValid() have side effect?

mbudiu-vmw:
No side effects

PR #4004, opened 2023-05-04
[P4Testgen] Fix small issues with metadata test backend. Add more functions to IR utilities. Merge default value function.

fruffy-bfn:
- Merge the `defaultValue` from `defaultValues.h` with `getDefaultValue` from `irutils.h`
- Fix up small issues with the metadata collection test back ends. Also collect the parser states that were traversed.
- Clean up some incorrect naming after the expanding coverage from IR statements to IR nodes.
- Do not use source info for coverage sets, instead use the clone id.
- Move the inja submodule initialization into the testgen extension. 
- Fix incorrect variable setting in the CMake test backend files.


jnfoster (approved):
Thanks!

PR #4008, opened 2023-05-16
Properly track indices when removing parser value sets from the select cases

fruffy-bfn:
Additional changes:
- Improve taint.cpp error messages.
- Create a p4cgraphs library, which we can link against.
- Get rid of the `register.h` snake-casing and use camel case consistent with the rest of the framework.


backends/p4tools/common/lib/taint.cpp
```diff
@@ -29,10 +29,20 @@ static bitvec computeTaintedBits(const SymbolicMapType &varMap, const IR::Expres
     CHECK_NULL(expr);
     // TODO: Replace these two with IR::StateVariable.
     if (const auto *member = expr->to<IR::Member>()) {
```
pkotikal, line 31: Nit: Move this into a small helper function?

pkotikal (approved):

backends/p4tools/common/lib/taint.cpp
fruffy-bfn: Yes, will try to clean up once I get rid of IR::Member in Testgen

PR #4014, opened 2023-05-25
[P4Testgen] Open up P4Testgen interface. 

fruffy-bfn:
- Add PRIVATE/PUBLIC labels to CMake library dependencies in the P4Tools framework.
- Remove `getPortNumWidthBits` and replace it with the constants in the respective back ends.
- Remove the seed as parameter from the back end implementation. That should be inferred from the options class.
- Make FinalState a movable class by replacing`const Model &` with `std::reference_wrapper<const Model>`.
- Introduce a helper function in the `testgen.cpp` class. 

@vlstill This PR makes some API changes, test backends will not have a seed argument and instead grab it from the static options class and `getPortNumWidthBits` is replaced by referencing a constant instead.


.github/CODEOWNERS
```diff
@@ -1,3 +1,3 @@
 # The P4Tools repository is owned by the following maintainers.
-backends/p4tools @fruffy @pkotikal @jnfoster @mbudiu-vmw
```
jnfoster, line 2: 😢

jnfoster (approved):
A quick review, but looks good!

PR #4021, opened 2023-06-05
[P4Testgen] Simplify the model

fruffy:
- The model was also saving state variables. This was not necessary since they are never really accessed. Remove this feature to make model usage less confusing. 
- The only instance where access to StateVariables was required in the model was the computation of the payload. We make the payload a symbolic variable instead, which is saved in the model.
- Taint computation required a map to look up members or state variables. However, these variables should have been resolved before that. Clean up the interface and require all state variables to be resolved before computing taint. 
- Add StringLiterals to the list of accepted Z3 expression. They can be used to represent the chosen action after a table has executed and in turn matched in switch statements. 
- Clean up the `test_coverage.py` script for artifact evaluation. Also update the testgen README.

@vlstill The only change impacting the extensions is that `state.hasTaint(cond)` will have to be replaced with `Taint::hasTaint(cond)`.


backends/p4tools/modules/testgen/README.md
```diff
@@ -26,44 +27,46 @@ These are the current usage flags:
...
+                                       "MAX_STATEMENT_COVERAGE".
+--packet-size-range packetSizeRange    Specify the possible range of the input packet size in bits. The format is [min]:[max]. The default values are "0:72000". The maximum is set to jumbo frame size (9000 bytes).
+--out-dir outputDir                    The output directory for the generated tests. The directory will be created, if it does not exist.
+--test-backend testBackend             Select the test back end. P4Testgen will produce tests that correspond to the input format of this test back end.
+--input-branches selectedBranches      List of the selected branches which should be chosen for selection.
+--track-branches                       Track the branches that are chosen in the symbolic executor. This can be used for deterministic replay.
+--with-output-packet                   Produced tests must have an output packet.
+--path-selection pathSelectionPolicy   Selects a specific path selection strategy for test generation. Options are: DEPTH_FIRST, RANDOM_BACKTRACK, GREEDY_STATEMENT_SEARCH, and RANDOM_STATEMENT_SEARCH. Defaults to DEPTH_FIRST.
```
pkotikal, line 58: Update this list to reflect the strategies above?

backends/p4tools/modules/testgen/README.md
```diff
@@ -26,44 +27,46 @@ These are the current usage flags:
...
+--track-coverage coverageItem          Specifies, which IR nodes to track for coverage in the targeted P4 program. Multiple options are possible: Currently supported: STATEMENTS, TABLE_ENTRIES Defaults to no coverage.
+--saddle-point saddlePoint             Threshold to invoke multiPop on RANDOM_ACCESS_MAX_COVERAGE.
```
pkotikal, line 60: Do we still support this?

backends/p4tools/modules/testgen/lib/final_state.h
```diff
@@ -39,8 +39,8 @@ class FinalState {
     /// Complete the model according to target-specific completion criteria.
```
pkotikal, line 39: Update comment?

pkotikal (approved):

backends/p4tools/modules/testgen/README.md
fruffy: Technically yes, the name is now RANDOM_STATEMENT_SEARCH. I will update it. We can also drop it since I do not know how functional it is.

backends/p4tools/benchmarks/test_coverage.py
```diff
@@ -298,17 +301,14 @@ def main(args, extra_args):
     print(f"Chosen seeds: {seeds}")
 
     strategies = [
```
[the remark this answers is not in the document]
fruffy: We have more but for the paper we only benchmarked these three strategies.

backends/p4tools/common/lib/symbolic_env.h
```diff
@@ -26,12 +26,6 @@ class SymbolicEnv {
     /// done on the given value before updating the symbolic state.
     void set(const IR::StateVariable &var, const IR::Expression *value);
 
-    /// Evaluates this symbolic environment in the context of the given model.
-    ///
-    /// A BUG occurs if any symbolic value in this environment refers to a variable that is not
-    /// bound by the given model.
-    [[nodiscard]] Model *evaluate(const Model &model) const;
-
     /// Substitutes state variables in @expr for their symbolic value in this environment.
```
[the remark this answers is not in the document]
fruffy: I believe it is used in some tests and the execution state class uses it to for debugging.

PR #4023, opened 2023-06-06
Also suppress no-gnu-zero-variadic-macro-arguments warnings for MacOs.

fruffy:
Get rid of the excessive printing of warnings for MacOS. Also add notes for some other spurious warnings. Unfortunately, we can not get rid of the Bison .ypp warning until we deprecate Ubuntu 18.04. 

This PR also coincidentally fixes a breaking change introduced by Protobuf in version 22.x. Protobuf now also pulls in Abseil, which we currently do not require and thus we run into undefined symbol errors. So far, this issue only occurs with MacOS builds, but may become a problem for all other builds in the future. 

Protobuf is a really problematic, huge dependency, I wish we could get rid of it.


jafingerhut (approved):
Caveat: I have never built p4c on macOS, nor looked at these scripts before.  However, I trust Fabian and the macOS CI tests are passing :-)

PR #4024, opened 2023-06-09
[P4Testgen] Resolve array indices in the interpreter instead of converting them in a compiler pass. 

fruffy-bfn:
This PR adds support for ArrayIndices within P4Testgen. We extend the compare function of state variables to also support indices and add an additional visitor function.

With this change, we can remove some of the conversion passes that caused the P4 IR to not type check after the mid end.


jnfoster (approved):
LGTM

PR #4025, opened 2023-06-12
Dpdk Backend: Fixed target_name and action parameter bitwidth

usha1830:
Fixes the following:
1) The "target_name" attribute in context json is incorrectly prefixed with control block name. This is not required as the external name of the table is captured in "name" attribute"
2) Fixed the bitwidth of action parameters as DPDK target needs all metadata to be byte aligned. The bitwidth is updated to 32 if the original bitwidth is <32, 64 otherwise.

### DIFF
--- a/backends/dpdk/dpdkContext.cpp
+++ b/backends/dpdk/dpdkContext.cpp
@@ -18,6 +18,7 @@ limitations under the License.
 
 #include "backend.h"
 #include "control-plane/bfruntime_ext.h"
+#include "dpdkUtils.h"
 #include "printUtils.h"
 namespace DPDK {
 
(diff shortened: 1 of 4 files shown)


backends/dpdk/dpdkContext.cpp
```diff
@@ -411,6 +417,12 @@ Util::JsonArray *DpdkContextGenerator::addActions(const IR::P4Table *table,
                 for (auto param : *(attr.params)) {
                     if (param->type->is<IR::Type_Bits>()) {
                         param_width = param->type->width_bits();
+                        if (param_width % 8 != 0) {
```
davidbolvansky, line 420: Avoid code duplication, create new function like get_param_width?

usha1830: Done

mihaibudiu (approved):

PR #4027, opened 2023-06-14
Hotfix for failing MacOS tests. Pin boost. 

fruffy:


fruffy:
> Haven't run this, but I'll trust you :-)

Thanks :). The MacOS is a bit brittle now because it depends on particular versions of dependencies. Hope that is fixable eventually.

Issue #4026, opened 2023-06-12
Unexpected PRIORITY error message


fruffy:
After #3748 `priority` has become a key word. This breaks existing programs using this name as a parameter. Was this intended? 
Example: 

```
control c(inout bit<32> x, in bit<32> priority) {

    apply {
        x = priority;
    }
}

control proto(inout bit<32> x);
package top(proto p);

top(c()) main;
```
Error:
```
../testdata/p4_16_samples/priority-key.p4(20):syntax error, unexpected PRIORITY
        x = priority
            ^^^^^^^^
[--Werror=overlimit] error: 1 errors encountered, aborting compilation
```

jnfoster:
Alternatively, we should allow using `priority` as an ordinary identifier, as we do with several other keywords.

mihaibudiu:
Let's first see if it can be fixed.

mihaibudiu:
I suspect this can be fixed, since I don't think the grammar is inherently ambiguous. 
I actually attempted to write the proper grammar, but by mistake I added PRIORITY to the production for `name` instead of `nonTypeName`. Unfortunately, adding it to `nonTypeName` causes a shift-reduce conflict. I suspect that's because on the discovery of a PRIORITY token it could be either the start of an expression for a keyset that starts with a PRIORITY identifier, or it could be a proper priority.

I am afraid I made a similar mistake with the 'list' identifier. Maybe @ChrisDodd can help, he is a bison wizard?

I think one solution would copy all the expression productions, similar to the nonBraceExpression rule, but that's ugly.

ChrisDodd:
The basic problem is the prefix-ambiguity of a `priority =` really needs two tokens of lookahead to differentiate it from the beginning of an expression that begins with `priority`.  This is easy enough to fix by duplicating the `entry` rule with and without an `entryPriority` so there is no need for an empty reduction of `opetEntryPriority` before seeing the `=`.  #4053

jafingerhut:
Now that #4053 is merged, perhaps there should also be a test program added that verifies `priority` can be used as an identifier name now?

PR #4029, opened 2023-06-14
Add the source file to the src info comparison operator.

fruffy:
The `SrcInfo` class was not taking the file name into account for equal or less operators. It can happen that the line and column number for a particular object are the same, but the source file is completely different.

### DIFF
--- a/lib/source_file.h
+++ b/lib/source_file.h
@@ -168,7 +168,9 @@ class SourceInfo final {
         return *this;
     }
 
-    bool operator==(const SourceInfo &rhs) const { return start == rhs.start && end == rhs.end; }
+    bool operator==(const SourceInfo &rhs) const {
+        return start == rhs.start && end == rhs.end && getSourceFile() == rhs.getSourceFile();
+    }
 
     cstring toDebugString() const;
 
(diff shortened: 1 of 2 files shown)


lib/source_file.h
```diff
@@ -168,7 +168,9 @@ class SourceInfo final {
         return *this;
     }
 
-    bool operator==(const SourceInfo &rhs) const { return start == rhs.start && end == rhs.end; }
+    bool operator==(const SourceInfo &rhs) const {
+        return start == rhs.start && end == rhs.end && getSourceFile() == rhs.getSourceFile();
```
ChrisDodd, line 172: The source file depends (solely) on the (start) line number -- so start == rhs.start, then the source files by definition must be identical.  So this extra comparison can never add anything?

ChrisDodd:
> The `SrcInfo` class was not taking the file name into account for equal or less operators. It can happen that the line and column number for a particular object are the same, but the source file is completely different.

This seems impossible -- the file name is determined from the line number, so if the line numbers are the same the file name by definition must be the same.

lib/source_file.h
mihaibudiu: Hey, @ChrisDodd, can you please take a look at https://github.com/p4lang/p4c/issues/4026, maybe you have a suggestion.

fruffy: > The source file depends (solely) on the (start) line number -- so if start == rhs.start, then the source files by definition must be identical. So this extra comparison can never add anything?

How is sourceInfo constructed? Or rather how is the file name stored?

I have been running into issues with #include files and their source info. 

I am maintaining a map which compares the sourceInfo of IR nodes here: 
https://github.com/p4lang/p4c/pull/4029/files#diff-0c31e90ea5ad25031d5734ed00bc8dee477c6c54a3208d1476b26d1eb5407c33R10
This works reasonably well, but in some cases the sourceLine line and column number are the same, but the originating file is different, which seems to lead to clashes.

fruffy:
Closing until I find a better solution.

PR #4030, opened 2023-06-15
[P4Testgen] Reduce overtainting by short-circuiting some expressions and extern invocations

### LINKED ISSUE #3887: Testgen taint analysis overtaints in some cases
For example, the expression `p.ipv4.isValid() && p.tcp.isValid() && p.ipv4.ttl == 1` is considered tainted since the last value `p.ipv4.ttl == 1` is uninitialized when p.ipv4 is invalid.

However, the expression short-circuits and `p.ipv4.ttl` is never evaluated if `p.ipv4` is invalid. The expression should not be tainted.

fruffy-bfn:
Sometimes we can overtaint because an expression that technically short-circuits is treated as taint. Implement an optimization for this behavior in the taint checker. Also try to eliminate sources of taint in some externs by exiting early (verify and update checksum).

Fixes #3887.

### DIFF
--- a/backends/p4tools/common/lib/taint.cpp
+++ b/backends/p4tools/common/lib/taint.cpp
@@ -112,6 +112,22 @@ bool Taint::hasTaint(const IR::Expression *expr) {
         return false;
     }
     if (const auto *binaryExpr = expr->to<IR::Operation_Binary>()) {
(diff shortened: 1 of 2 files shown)


jnfoster (approved):
Nice!

PR #4031, opened 2023-06-15
[P4Testgen] Remove execution state from the symbolic executor, make it a parameter.

fruffy-bfn:
Execution state should only live as long as the `run` function of the symbolic executor is active. This also makes sure we do not access some stale execution state via the symbolic executor. 

This impacts the SelectedBranches implementation, which is now simplified.


jnfoster (approved):
Nice. This is a good change.

PR #4034, opened 2023-06-16
[P4Testgen] Control the Z3 dependency in CMake using fetchcontent, generalize FindZ3.cmake

fruffy-bfn:
One of the barriers of installing P4Tools for many people is that we require a particular version of Z3. This PR introduces CMake's FetchContent to pull the version of Z3 that we currently support. This way we can avoid version mismatches and users do not need to install a custom version since we build P4Tools using the local Z3 version and not the global one.

There is a risk to this and that is that FetchContent is somewhat immature in usability. We have to make sure it is stable enough to be usable.


jafingerhut (approved):
I have tested out these changes on a freshly installed Ubuntu 20.04 system in a slightly tweaked version of my install bash scripts that use this version of the p4c repo, and it installs p4testgen just fine.

backends/p4tools/CMakeLists.txt
```diff
@@ -34,18 +34,20 @@ project(P4TOOLS VERSION ${P4C_SEM_VERSION_STRING})
...
+set (P4TOOLS_Z3_VERSION "4.11.2")
+message("Fetching Z3 version ${P4TOOLS_Z3_VERSION} for P4Tools...")
+set(FETCHCONTENT_QUIET OFF CACHE BOOL "Silence FetchContent output" FORCE)
+FetchContent_Declare(z3
+    URL https://github.com/Z3Prover/z3/releases/download/z3-${P4TOOLS_Z3_VERSION}/z3-${P4TOOLS_Z3_VERSION}-x64-glibc-2.31.zip
+    URL_HASH SHA256=9d0f70e61e82b321f35e6cad1343615d2dead6f2c54337a24293725de2900fb6
+    USES_TERMINAL_DOWNLOAD      TRUE
+)
+FetchContent_MakeAvailable(z3)
+set(FETCHCONTENT_QUIET ON CACHE BOOL "Silence FetchContent output" FORCE)
```
vlstill, line 47: Please reset it to the previous state, it might have been `OFF` before.

backends/p4tools/CMakeLists.txt
```diff
@@ -34,18 +34,20 @@ project(P4TOOLS VERSION ${P4C_SEM_VERSION_STRING})
...
+include_directories(BEFORE ${z3_SOURCE_DIR}/include)
+set (P4TOOLS_Z3_LIB ${z3_SOURCE_DIR}/bin/libz3.a)
```
vlstill, line 49: Shouldn't the include and lib dirs be set by `FetchContent_MakeAvailable`? I think they should as it will just run normal cmake build. If you just call `MakeAvailable` and not explicitly set the paths I think that would allow higher-level projects to override Z3 with system one, I think that should be allowed (at their own risk of breaking things by using different version, of course :-) ).

vlstill (changes requested):
Why are there so many code changes if the main change is making Z3 a submodule? Why would that require change outside CMake?

Also, I have two points about the CMake that I think should be resolved before merging this.

backends/p4tools/CMakeLists.txt
fruffy-bfn: We are just fetching a release `.zip` without a CMakeLists file, so FetchContent does not know the include or library folders.  Building z3 from scratch unfortunately takes quite some time. 

The reason I used a custom variable for the lib here was to not conflict with bf-p4c's Z3 library. Let me think about a good alternative here.

jnfoster (approved):
LGTM

backends/p4tools/common/CMakeLists.txt
```diff
@@ -5,16 +5,24 @@ include(common)
...
+# Pull in a specific version of Z3 and link against it.
+set(P4TOOLS_Z3_VERSION "4.11.2")
+message("Fetching Z3 version ${P4TOOLS_Z3_VERSION} for P4Tools...")
+
+# Print out download state while setting up Z3.
+set(FETCHCONTENT_QUIET_PREV ${FETCHCONTENT_QUIET})
+set(FETCHCONTENT_QUIET OFF)
+fetchcontent_declare(
+  z3
+  URL https://github.com/Z3Prover/z3/releases/download/z3-${P4TOOLS_Z3_VERSION}/z3-${P4TOOLS_Z3_VERSION}-x64-glibc-2.31.zip
```
vlstill, line 17: Oh, I have not noticed before that you are using a binary release of Z3, not building from source. In that case, I strongly suggest to use it only as default to simplify build on Linux + x64, but allow the user to override with their own Z3. The reason is this is actually quite fragile:

- The build uses GLIBC 2.31, while this is OK for Ubuntu 20/Debian 10 but it will not run on RHEL 8 that has 2.28 (even though `p4c` will build there if newer GCC is used).
- It is build with GCC 9.4 (`objdump -s --section .comment libz3.so`), while we support GCC >= 9.1 (which is probably compatible), the problem is this will break builds with Clang + libc++ -- while not usual on Linux, someone could be using this.
- It only works on Linux + x86. I presume `p4c` can be also build on macOS and it is quite likely it would build on *BSD, but neither of these will work with this Z3 library.

---

Therefore what I suggest is to allow the user to specify `Z3_INCLUDE_DIR` and `Z3_LIBRARY` the way they could before. If these are not specified, than some rudimentary check should be done and than Z3 fetched:
- check that [`CMAKE_SYSTEM_NAME`](https://cmake.org/cmake/help/v3.0/variable/CMAKE_SYSTEM_NAME.html#cmake-system-name) is `Linux`
- check that [`CMAKE_SYSTEM_PROCESSOR`](https://cmake.org/cmake/help/latest/variable/CMAKE_SYSTEM_PROCESSOR.html) is `x86_64`.

Without setting the paths, the install should proceed as you suggested here. This way most users (who use Linux with usual setup) will benefit from this while the other will not need to modify CMakeLists or do some hack to override Z3.

vlstill (changes requested):

vlstill:
Also, thanks for cleaning the rest of the changes away from this PR.

vlstill (approved):
Thank you. I've checked that using installed Z3 works.

PR #4036, opened 2023-06-16
[P4Testgen] Miscellaneous fixes for P4testgen

### LINKED ISSUE #4033: Issues with p4testgen back end
Sample command line here: 
is:
```
p4testgen –target [TARGET] –arch [ARCH] –max-tests 10 –out-dir [OUT] prog.p4
```
but current p4testgen as of 2023-Jun-16 gives error unless one also supplies `--test-backend ptf` or one of the other allowed test backends.

Is it expected that stf and metadata test-backends produce only 1 test case, even if the p4testgen output says it is generating more?

Very minor: Seems slightly weird that test cases are numbered [0, N-1] in protobuf back end, but [1,N] in PTF test Python program.  Given that numbers of test cases start with 0 in stdout output of p4testgen, seems better to number them starting from 0 in Python PTF code, too.

In the PTF back end, the Python program written by p4testgen has a `@bt.autocleanup` decorator on the `setUp` method of class `AbstractTest`.  It does _nothing_ there.  I believe it should instead be on the `runTestImpl` method, where I have found that it actually can remove table entries at the end of a test case execution.

fruffy:
- Unify test numbering and the name of the variable in the test back ends. They all start from 0 now. Also make sure the test name does not have a duplicate `.`. 
- Provide hints on available test back ends. 
- Verify that no other packets are present if the packet is dropped in PTF.
- Add some hints on the support Z3 versions. #4034 will fix this, too.
- Support array indices when converting references.

Fixes #4033.

### DIFF
--- a/backends/p4tools/README.md
+++ b/backends/p4tools/README.md
@@ -30,6 +30,7 @@ make
 ## Dependencies
 * [inja](https://github.com/pantor/inja) template engine for testcase generation.
 * [z3](https://github.com/Z3Prover/z3) SMT solver to compute path constraints.
+    * Important: We currently only support Z3 versions 4.8.14 to 4.12.0.
 
 
 ## Development Style
--- a/backends/p4tools/common/lib/variables.cpp
+++ b/backends/p4tools/common/lib/variables.cpp
@@ -54,6 +54,9 @@ IR::StateVariable convertReference(const IR::Expression *ref) {
     if (const auto *member = ref->to<IR::Member>()) {
         return member;
     }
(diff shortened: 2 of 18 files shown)


jafingerhut (approved):
I have tried out these changes and all the changes look like improvements to me.  The PTF test generated still has classes named starting at 1, e.g. `class Test1` up to `class Test10` when generating 10 tests, instead of `class Test0` up to `class Test9`.

jnfoster (approved):
LGTM

PR #4039, opened 2023-06-19
[P4Testgen] Fix incorrect boolean cast in the Z3 solver implementation.

### LINKED ISSUE #4038: Testgen does not properly generate inputs when parser decision depends on `bool` variable
Attached program [testgen-bmv2.p4](https://github.com/p4lang/p4c/files/11788214/testgen-bmv2.p4.txt) is mainly interesting for its parser. If I run testgen like this I get 4 tests (and complete coverage):

```bash
$ ./p4testgen --target bmv2 --std p4-16 --arch v1model --test-backend PTF --track-coverage STATEMENTS testgen-bmv2.p4  --max-tests 100
```

However, if I cause the program to use `bool` instead of `bit<1>` for `Foo::b`, I instead get only 2 tests and incomplete coverage:

```bash
$ ./p4testgen --target bmv2 --std p4-16 --arch v1model --test-backend PTF --track-coverage STATEMENTS testgen-bmv2.p4  --max-tests 100 -DUSE_BOOL
```

This seems to me like a bug in testgen where for some reason it does not consider both options of what `var_parser_a.b` could have (the value seems to be `false` in both tests).

@fruffy, could you please have a look at this weird behavior?

fruffy:
Fixes #4038.

Casts from bit<1> to boolean were not handled correctly in the solver. They always evaluated to false, producing fewer branches than actually possible.

Thanks for bringing attention to this @vlstill, this is a nasty one.

### DIFF
--- a/backends/p4tools/common/core/z3_solver.cpp
+++ b/backends/p4tools/common/core/z3_solver.cpp
@@ -477,7 +477,7 @@ bool Z3Translator::preorder(const IR::Cast *cast) {
     if (cast->destType->is<IR::Type_Boolean>()) {
         if (const auto *exprType = castExtrType->to<IR::Type_Bits>()) {
             if (exprType->width_bits() == 1) {
-                castExpr = solver.z3context.bool_val(castExpr.bool_value() == Z3_L_TRUE);
+                castExpr = z3::operator==(castExpr, solver.z3context.bv_val(1, 1));
             } else {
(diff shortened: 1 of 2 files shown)


mihaibudiu (approved):

jnfoster (approved):
Whew!

vlstill:
@fruffy, thank you!

PR #4042, opened 2023-06-20
Fix race condition for parallel PTF nanomsg tests by using network namespaces.

fruffy:
The P4Tools PTF tests using nanomsg are failing because the device ID and TCP ports are conflicting during parallel execution. 

Use network namespaces as a bandaid to avoid the spurious CI failures until I find a better solution.

### DIFF
--- a/backends/bmv2/run-bmv2-ptf-test.py
+++ b/backends/bmv2/run-bmv2-ptf-test.py
@@ -6,6 +6,7 @@ import os
 import random
 import sys
 import tempfile
+import uuid
 from datetime import datetime
 from pathlib import Path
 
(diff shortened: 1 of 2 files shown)


mihaibudiu:
I am not the right person to review ptf programs, you may ask contributors to the ebpf backend, they wrote many

fruffy:
> I am not the right person to review ptf programs, you may ask contributors to the ebpf backend, they wrote many

This is a problem I introduced with #3951 unrelated to the PTF programs in the eBPF back end. The issue is that race conditions can occur when two tests bind to the same port. Since the tests are not contained in namespaces one of the programs will fail. 

This PR just hotfixes this regression so that other PRs can pass.

backends/bmv2/run-bmv2-ptf-test.py
```diff
@@ -97,6 +98,25 @@ def __del__(self):
             # Terminate the switch process and emit its output in case of failure.
             testutils.kill_proc_group(self.switch_proc)
 
+    def create_bridge(self, num_ifaces: int) -> Bridge:
+        """Create a network namespace environment."""
+        testutils.log.info(
+            "---------------------- Creating a namespace ----------------------",
+        )
+        random.seed(datetime.now().timestamp())
+        bridge = Bridge(str(random.randint(0, sys.maxsize)))
```
mihaibudiu, line 107: wouldn't a guid be safer?

mihaibudiu (approved):

PR #4044, opened 2023-06-21
Cleaner dbprint of if-else if chains

ChrisDodd:


mihaibudiu (approved):

fruffy (approved):
Btw, clang-format has recently added the [insertBraces](https://clang.llvm.org/docs/ClangFormatStyleOptions.html#insertbraces) option, which can automatize newline bracket formatting, too. #4042 will fix the failing tools test, so this can be merged.

ChrisDodd:
> Btw, clang-format has recently added the [insertBraces](https://clang.llvm.org/docs/ClangFormatStyleOptions.html#insertbraces) option, which can automatize newline bracket formatting, too. #4042 will fix the failing tools test, so this can be merged.

I'm not sure if insertBraces doesn't make things even worse.  I already strongly dislike putting closing braces on their own lines, as that tends to mess up the vertical formatting and make the code harder to read and scan, but clang-format does not appear to have an option to put the `}` on the end of the line as I generally prefer.

PR #4045, opened 2023-06-22
Fixed tc table type

### LINKED ISSUE #4041: p4tc backend table type seems define wrong
Hi,
p4c tc backend seems to be a problem with the command to create the table entry.

I have tried compiling the p4 file in p4tc-v2-examples (https://github.com/p4tc-dev/p4tc-v2-examples/blob/main/model2/1table/exact/redirect_l2_pna_1t_exact.p4) with p4tc backend, But the generated runtime file doesn't seem to match the current p4tc code.
The shell command to create a table table in the generated runtime file is as follows,
```
$TC p4template create table/redirect_l2_pna_1t_exact/ingress/nh_table \
tblid 1\
type EXACT_TYPE \
keysz 32 nummasks 8 tentries 262144 \
table_acts act name redirect_l2_pna_1t_exact/ingress/send_nh \
act name redirect_l2_pna_1t_exact/ingress/drop
```
The type type is EXACT_TYPE, but the type defined by p4tc is exact, so an error occurs.
In the backends/tc/tc.def file, should the strings `EXACT_TYPE`, `LPM_TYPE`, `TERNARY_TYPE` be `exact`, `lpm`, `ternary`?
```
        switch(matchType) {
                case EXACT_TYPE :
                    matchTypeString = "EXACT_TYPE";
                    break;
                case LPM_TYPE :
                    matchTypeString = "LPM_TYPE";
                    break;
                case TERNARY_TYPE :
                    matchTypeString = "TERNARY_TYPE";
                    break;
        }
```

Sosutha:
This PR fixes the table type in template file for tc backend.
Fixes #4041

### DIFF
--- a/backends/tc/tc.def
+++ b/backends/tc/tc.def
@@ -204,13 +204,13 @@ class TCTable {
         cstring matchTypeString = "";
         switch(matchType) {
                 case EXACT_TYPE :
-                    matchTypeString = "EXACT_TYPE";
+                    matchTypeString = "exact";
                     break;
                 case LPM_TYPE :
-                    matchTypeString = "LPM_TYPE";
+                    matchTypeString = "lpm";
                     break;
                 case TERNARY_TYPE :
-                    matchTypeString = "TERNARY_TYPE";
+                    matchTypeString = "ternary";
                     break;
         }
         return matchTypeString;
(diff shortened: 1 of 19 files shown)


mihaibudiu (commented):
Can you add a test?

Sosutha:
> Can you add a test?

There are changes in the outputs of existing testcases. Hence I did not add new test.

mihaibudiu:
The changes should be part of the commit. But according to the issue the tests didn't catch this problem

Sosutha:
> The changes should be part of the commit. But according to the issue the tests didn't catch this problem

Tests did not catch this problem since the expected data was with table types as EXACT_TYPE, LPM_TYPE and TERNARY_TYPE.
Now since the table type that the p4tc supports is exact, lpm, ternary, Compiler code and expected data has been updated in this commit.

usha1830 (approved):
LGTM.
@mihaibudiu This looks like a mismatch issue between what is recognized by target and what is generated by compiler.

PR #4046, opened 2023-06-22
[P4Testgen] BMv2 test generation improvements

fruffy:
- Use source information as the comparison operator again, not clone ID.
- Clean up the Z3 solver. Do not keep context around and reset memory usage periodically to avoid running out of memory.
- Reduce the PTF wait time to accelerate PTF testing.
- Clean up the `entry_restriction` parser and print a warning if a restriction is not satisfiable. Also ensure argument and parameter names are consistent. 
- Change the optional implementation once again. Convert it into a ternary match that either matches exactly or is wildcarded.

### DIFF
--- a/backends/p4tools/modules/testgen/targets/bmv2/backend/ptf/ptf.cpp
+++ b/backends/p4tools/modules/testgen/targets/bmv2/backend/ptf/ptf.cpp
@@ -404,10 +404,10 @@ class Test{{test_id}}(AbstractTest):
 ## else 
         ptfutils.verify_packet(self, exp_pkt, eg_port)
         bt.testutils.log.info("Verifying no other packets ...")
-        ptfutils.verify_no_other_packets(self, self.device_id, timeout=2)
+        ptfutils.verify_no_other_packets(self, self.device_id, timeout=0.1)
 ## endif
 ## else
-        ptfutils.verify_no_other_packets(self, self.device_id, timeout=2)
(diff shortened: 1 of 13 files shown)


jnfoster (approved):
LGTM

backends/p4tools/modules/testgen/targets/bmv2/backend/ptf/ptf.cpp
```diff
@@ -404,10 +404,10 @@ class Test{{test_id}}(AbstractTest):
 ## else 
         ptfutils.verify_packet(self, exp_pkt, eg_port)
         bt.testutils.log.info("Verifying no other packets ...")
-        ptfutils.verify_no_other_packets(self, self.device_id, timeout=2)
+        ptfutils.verify_no_other_packets(self, self.device_id, timeout=0.1)
```
jafingerhut, line 407: I would recommend maybe using a global variable name or a return value from a global function defined with `def` at the top level to get this timeout value, so that after a PTF test is generated, a user can replace this number in one place in the PTF source file, rather than many places.

Similarly if there are other timeouts to wait for a different reason.

jafingerhut: Rationale: 0.1 sec might work right now for BMv2, but I've been testing with other environments where it is definitely not long enough.

fruffy: What are the environments where issues are coming up?

PR #4047, opened 2023-06-22
Include annotations for new_type messages.

rhalstea:


mihaibudiu (approved):

rhalstea:
Thanks @mihaibudiu, should I write a test for this somewhere? Up to this point I've only ran it manually.

mihaibudiu:
Tests are always good. I don't really know how the control-plane API is tested, though.

rhalstea:
What would be my next steps? I don't have merge permissions.

mihaibudiu:
If anyone who knows this code could comment it would be great. Otherwise I can merge it.

jafingerhut:
If your changes should cause the contents of the P4Info file created by p4c for at least one P4 test program you can write to change, from before your code changes to after your code changes, you can add the .p4 source file in the testdata/p4_16_samples directory, and then when you compile your local copy of p4c you can `cd p4c/build`, run cmake, and that should create a test bash script with a name containing <foo> if the P4 source file name is <foo>.p4.  If you run that test script inside the p4c/build directory with `-f` option, it creates the P4Info and several other output files and writes them into the testdata/p4_16_samples_output directory.  Check them manually to ensure they contain what you expect, and if they do, add the P4 source file and the p4_16_samples_outputs/ files to this PR.

rhalstea:
@antoninbas seems to have written most of the code, or @smolkaj ... any thoughts?

antoninbas:
LGTM, but Andy is right: one of the existing test case should be updated with an annotation, to validate this change

rhalstea:
Thanks @jafingerhut and @antoninbas ... updated p416-type-use.p4 which seemed relevant.

jafingerhut (approved):
LGTM

smolkaj (approved):

PR #4048, opened 2023-06-26
Wrap the TC defines in a namespace, use constexpr.

### LINKED ISSUE #4028: p4tc pollutes IR-generated.h with generic #defines
@Sosutha 

https://github.com/p4lang/p4c/blob/main/backends/tc/tc_defines.h is included in the `.def` file and declares very generic `#defines`. These #defines can potentially clash with other back ends. Please make sure these #defines are constant expressions and wrap them in a namespace.

fruffy:
Fixes #4028.

### DIFF
--- a/backends/tc/tc_defines.h
+++ b/backends/tc/tc_defines.h
@@ -17,64 +17,67 @@ and limitations under the License.
 #ifndef BACKENDS_TC_TC_DEFINES_H_
 #define BACKENDS_TC_TC_DEFINES_H_
 
-#define DEFAULT_TABLE_ENTRIES 2048
-#define DEFAULT_KEY_MASK 8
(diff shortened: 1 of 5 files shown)


backends/tc/tc_defines.h
```diff
@@ -17,64 +17,66 @@ and limitations under the License.
...
-#define MAX_PNA_OUTPUT_META 1
-
-#define TABLEDEFAULT 0
-#define TABLEONLY 1
-#define DEFAULTONLY 2
-
-#define EXACT_TYPE 0
-#define LPM_TYPE 1
-#define TERNARY_TYPE 2
+namespace TcConstants {
```
mihaibudiu, line 20: why not just TC?

mihaibudiu (approved):

Issue #3986, opened 2023-04-20
When to "officially" drop support for Ubuntu 18.04 building of p4c?


jafingerhut:
https://ubuntu.com/about/release-cycle

As you can see at the link above, the period for "hardware and maintenance updates" ends 5 years after a LTS release, so for Ubuntu 18.04 that ends during April 2023.

As far as I know, the extra 5 years of Extended Security Maintenance requires paying Ubuntu/Canonical money.

For brevity, I will call an Ubuntu LTS release "expired" if it is over 5 years since its initial release date.  So Ubuntu 16.04 is expired now, and Ubuntu 18.04 is expired some time this month, but definitely by May 1, 2023.

It seems reasonable to me that for _building_ p4c from source code (and perhaps for any other open source p4lang project), it is reasonable to no longer go to significant extra work to support that for expired Ubuntu LTS releases.  By that time, there are always 2 additional LTS releases available to use.

Do we want to make this "official p4c developer/maintainer policy"?  That is, if p4c fails to compile from source on an expired Ubuntu release, it is less likely that anyone will devote any effort to fixing such isses?  And in fact we might update some build scripts to take advantage of later versions of the build tools available as the default installable packages on later Ubuntu releases, even if that is known to break the build for expired Ubuntu versions?

Fabian Ruffy and others are currently dealing with p4c build system issues that are known to cause the build to fail for Ubuntu 18.04.  If some kind of majority of p4c developers agree that it is OK for builds of p4c on expired Ubuntu releases to no longer work, his changes can be made easier.

jafingerhut:
@mbudiu-vmw @jnfoster @fruffy @chrisdodd @hanw @usha1830 @vgurevich Just pinging what I hope is a representative set of people on the questions raised in this issue.  I am pretty sure that the p4c build no longer works on an Ubuntu 16.04 system today, but the question of exactly how long to support expired Ubuntu releases might be one we could agree upon.

fruffy:
I am always happy to drop older Ubuntu versions. However, Ubuntu 18.04 is the only CI check for the docker image. We have one other CI check, but that only builds and does not run tests. Regardless of the decision here, I will try to fix the issue I am encountering in #3985.

One thing we could do is to tag the commit where we officially dropped support for a particular version. That way, we can still package P4C for a particular debian/Ubuntu distribution, if needed.

mbudiu-vmw:
[...]

PR #4049, opened 2023-06-27
Change Dockerfile to build p4tools by default.

chrispsommers:
Build p4tools into p4c docker image by default. In particular, this will make p4testgen easily available in the prebuilt docker image on dockerhub/p4lang/p4c.

Impact:
* On my x86 platform, the p4c docker image build time increased from 29 min to 37 min. Since the main intent is to simplify developers' lives by using a prebuilt image, this seems like a good tradeoff. Building is done in the background in a CI pipeline.
* The docker p4c image size increased from 1.59GB to 1.70GB.

### DIFF
--- a/.github/workflows/ci-test.yml
+++ b/.github/workflows/ci-test.yml
@@ -129,7 +129,8 @@ jobs:
 
     - name: Build (Ubuntu 18.04, GCC)
       run: |
-        docker build -t p4c --build-arg BASE_IMAGE=ubuntu:18.04 --build-arg IMAGE_TYPE=test .
+        docker build -t p4c --build-arg BASE_IMAGE=ubuntu:18.04 --build-arg IMAGE_TYPE=test \
+        --build-arg ENABLE_TEST_TOOLS=OFF .
         ./tools/export_ccache.sh
 
       # run with sudo (...) --privileged
--- a/Dockerfile
+++ b/Dockerfile
@@ -25,7 +25,7 @@ ARG DEBIAN_FRONTEND=noninteractive
 # Whether to install dependencies required to run PTF-ebpf tests
 ARG INSTALL_PTF_EBPF_DEPENDENCIES=OFF
 # Whether to build the P4Tools back end and platform.
-ARG ENABLE_TEST_TOOLS=OFF
+ARG ENABLE_TEST_TOOLS=ON
 # Whether to treat warnings as errors.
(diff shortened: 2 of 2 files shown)


fruffy (approved):
Great! The size increase makes sense. The P4Tools framework is rather large and pulls in a dependencies such as Z3.

fruffy:
The nanomsg installation is failing for Ubuntu 18. We may consider disable P4Testgen support for this particular CI run.

chrispsommers:
I noticed that docker build under Ubuntu 18.04 fails after adding p4toolks to the build. It looks like lack of Python support in older distros. https://github.com/p4lang/p4c/actions/runs/5384634043/jobs/9772783270?pr=4049

It would be a shame to suppress routine docker build/publish of p4tools because it fails in an older OS. Any opinions on how to proceed? I could experiment with docker build-args to suppress building p4tools for older images which don't support it.

fruffy:
> I noticed that docker build under Ubuntu 18.04 fails after adding p4toolks to the build. It looks like lack of Python support in older distros. https://github.com/p4lang/p4c/actions/runs/5384634043/jobs/9772783270?pr=4049
> 
> It would be a shame to suppress routine docker build/publish of p4tools because it fails in an older OS. Any opinions on how to proceed? I could experiment with docker build-args to suppress building p4tools for older images which don't support it.

Yes, this is part of a discussion here #3986. Technically, we do not need nanomsg, which is the installation that is failing. But P4Tools has never been tested on Ubuntu 18.04 and I believe that is not the only problem. What we can do is to disable building P4Tools for Ubuntu 18.04 here:
https://github.com/p4lang/p4c/blob/main/tools/ci-build.sh#L82

chrispsommers:
> What we can do is to disable building P4Tools for Ubuntu 18.04 here:
> https://github.com/p4lang/p4c/blob/main/tools/ci-build.sh#L82

Thanks! I came to the same conclusion about where to disable p4tools build for 18.04 and am testing a fix now. Fortunately, the Dockerfile already parameterized everything with `ARG` so it's a simple thing to change.

chrispsommers:
> Great! The size increase makes sense. The P4Tools framework is rather large and pulls in a dependencies such as Z3.

@fruffy Thanks for the quick review! Hopefully, this lowers the barrier to experimentation and adoption of p4testgen. Thanks @jafingerhut for encouraging me to submit this.

PR #4050, opened 2023-06-28
Add Bazel build rules for P4TestGen

kheradmandG:


BUILD.bazel
```diff
@@ -145,6 +145,7 @@ filegroup(
         "frontends/p4-14/ir-v1.def",
         "backends/bmv2/bmv2.def",
         "backends/dpdk/dpdk.def",
+        "//backends/p4tools:p4tools.def",
```
smolkaj, line 148: Nit: use a relative path, for consistency?

backends/p4tools/BUILD.bazel
```diff
@@ -0,0 +1,138 @@
+# The BUILD file for p4testgen, inspired by the tool's cmake build rules.
+
+exports_files(["p4tools.def"])
+
+package(
+    #default_applicable_licenses = ["//:license"],
```
smolkaj, line 6: Remove?

backends/p4tools/BUILD.bazel
```diff
@@ -0,0 +1,138 @@
...
+    licenses = ["notice"],
+)
+
+# Extend the list with the name of the desired targets (in "modules/testgen/targets/).
+TESTGEN_TARGETS = ["bmv2"]
```
smolkaj, line 11: Nit: since we expect this to be extended, add a trailing comma to the list and make it multi-line?
```
TESTGEN_TARGETS = [
  "bmv2",
]
```

bazel/BUILD.inja.bzl
```diff
@@ -0,0 +1,7 @@
+cc_library(
```
smolkaj, line 1: Here and for the other BUILD files, the extension should be `.bazel`.

bazel/p4c_deps.bzl
```diff
@@ -75,3 +75,35 @@ filegroup(
             strip_prefix = "protobuf-21.10",
             sha256 = "6fc9b6efc18acb2fd5fb3bcf981572539c3432600042b662a162c1226b362426",
         )
+    if not native.existing_rule("rules_foreign_cc"):
+        http_archive(
+            name = "rules_foreign_cc",
+            sha256 = "d54742ffbdc6924f222d2179f0e10e911c5c659c4ae74158e9fe827aad862ac6",
+            strip_prefix = "rules_foreign_cc-0.2.0",
+            url = "https://github.com/bazelbuild/rules_foreign_cc/archive/0.2.0.tar.gz",
```
smolkaj, line 83: Nit: shall we try bumping this up to the latest version?

smolkaj (changes requested):
Heroic work! Thanks, Ali!

A few nits

BUILD.bazel
kheradmandG: Since  'backends/p4tools' is a subpackage (while the other files are not), this should be like this

bazel/p4c_deps.bzl
kheradmandG: The commands have changed in the new version. Get this errors:
```
ERROR: com_github_z3prover_z3/BUILD.bazel:8:15: @com_github_z3prover_z3//:api: no such attribute 'configure_env_vars' in 'configure_make' rule
ERROR: com_github_z3prover_z3/BUILD.bazel:8:15: @com_github_z3prover_z3//:api: no such attribute 'make_commands' in 'configure_make' rule
```
Could not figure out what needs to be changed how with quick search.

fruffy (approved):
Great, thanks doing this. I do not know Bazel that well so I can not comment on the actual code.

Is there anything in the current setup of P4Tools that made it difficult to write this build script? 

Also  `testgen` is just a P4Tools module. In the future, there will be more modules, not just `testgen`. For example, a random code generator or other fuzzing tools. 

I would consider splitting the Bazel code into a `p4tools-common` part and the testgen-specific code, if that is possible.

kheradmandG:
Thanks Fabian.

> Is there anything in the current setup of P4Tools that made it difficult to write this build script?

- In the past, some includes were using inconsistent paths, but this seems to be mostly addressed now. 
- There was some challenge with auto-generated source files as well. See [version](https://github.com/p4lang/p4c/pull/4050/files#diff-333ed5f42a318714f702961606bb4ee97e00ebc705cf08b2bdf8219181fd33c8R14-R29) and [register targets](https://github.com/p4lang/p4c/pull/4050/files#diff-333ed5f42a318714f702961606bb4ee97e00ebc705cf08b2bdf8219181fd33c8R50-R77). 

> I would consider splitting the Bazel code into a p4tools-common part and the testgen-specific code, if that is possible.

It is already somewhat like that I think. There is [common](https://github.com/p4lang/p4c/pull/4050/files#diff-333ed5f42a318714f702961606bb4ee97e00ebc705cf08b2bdf8219181fd33c8R31-R48) library and then [testgen](https://github.com/p4lang/p4c/pull/4050/files#diff-333ed5f42a318714f702961606bb4ee97e00ebc705cf08b2bdf8219181fd33c8R79-R114) library that uses it.

bazel/p4c_deps.bzl
smolkaj: Yes, +1, no need to dig deeper.

fruffy:
>     * In the past, some includes were using inconsistent paths, but this seems to be mostly addressed now.
> 
>     * There was some challenge with auto-generated source files as well. See [version](https://github.com/p4lang/p4c/pull/4050/files#diff-333ed5f42a318714f702961606bb4ee97e00ebc705cf08b2bdf8219181fd33c8R14-R29) and [register targets](https://github.com/p4lang/p4c/pull/4050/files#diff-333ed5f42a318714f702961606bb4ee97e00ebc705cf08b2bdf8219181fd33c8R50-R77).

I guess this seems unavoidable, at least I do not know an approach that generalizes across build frameworks. 

> > I would consider splitting the Bazel code into a p4tools-common part and the testgen-specific code, if that is possible.
> 
> It is already somewhat like that I think. There is [common](https://github.com/p4lang/p4c/pull/4050/files#diff-333ed5f42a318714f702961606bb4ee97e00ebc705cf08b2bdf8219181fd33c8R31-R48) library and then [testgen](https://github.com/p4lang/p4c/pull/4050/files#diff-333ed5f42a318714f702961606bb4ee97e00ebc705cf08b2bdf8219181fd33c8R79-R114) library that uses it.

The CMake build picks up all modules present in the `module` folder automatically. The top-level code has no mention of `testgen`, which is not the case here. In CMake, you just scan a specific folder and it picks up the submodules automatically. This allows you to extend the codebase with external repositories or keep some code private. Is there a Bazel equivalent?

kheradmandG:
> The CMake build picks up all modules present in the `module` folder automatically. The top-level code has no mention of `testgen`, which is not the case here. In CMake, you just scan a specific folder and it picks up the submodules automatically. This allows you to extend the codebase with external repositories or keep some code private. Is there a Bazel equivalent?

Probably there is an equivalent but for now I've only focused on testgen with bmv2 target. Supporting everything probably requires more trial and error to figure out the dependencies and so on. The current form is enough for our usecase. We can extend this in the future. 

> Btw, let me know when this is ready to merge.

Let's merge it :)

PR #4051, opened 2023-06-29
[P4Testgen] More fixes to the refers_to and p4_asserts parser

fruffy:
- Clean up table reference lookups for the refers_to parser. The refers_to parser was never checking the table references in the annotation. 
- Make sure string conversions to `boost::multiprecision:cpp_int` are safe by converting the `std::string_view` to a std::string first.  This is an additional copy, but older versions of boost do not safely convert `std::string_view`
- General cleanup and edits. 
- Fix a bug with path selection strategies (they did not update the execution state).

We should really use a proper lexer/parser for the P4Constraints implementation here.


jnfoster (approved):
LGTM 👍

PR #4053, opened 2023-07-07
Fix parser so that `priority` can be used in expressions

ChrisDodd:
We could refactor things so that `keysetExpression` always returns a `ListExpression` to simplify the duplication a bit (then the `if` in the `entry` actions goes away.)  This likely also means changing the `keyset` field in `SelectExpression` to be a `ListExpression` as well, which could simplify many places in the code where those are handled using an explicit check for `ListExpression`.


mihaibudiu (approved):

mihaibudiu:
Fixes #4026

fruffy:
Great, thanks for fixing this!

Issue #4055, opened 2023-07-08
Build errors with Protobuf 23


jkhsjdhjs:
Hey,

I recently created a package of the p4 compiler for the Arch Linux User Repository: https://aur.archlinux.org/packages/p4lang-p4c
It was already a bit complicated to build it, but I eventually got it working by using `CMAKE_UNITY_BUILD=ON` (for some reason the compilation error-ed without it) and by specifying `PROTOBUF_LIBRARY=protobuf`, because it would complain about the Protobuf version or not find it while linking (see https://aur.archlinux.org/cgit/aur.git/tree/PKGBUILD?h=p4lang-p4c#n50).

Recently, the Arch Linux developers updated protobuf from 21.12 to 23.3 (don't ask me about these version numbers, but that seems to correspond to library version 4.23.3, while the previous one was 3.21.12). Anyway, this protobuf version depends on the abseil cpp libraries, and since CMakeLists.txt currently only links against Protobuf directly, this leaves undefined references that cause the build to fail. I found a possible solution to that problem here: https://github.com/protocolbuffers/protobuf/issues/12292#issuecomment-1529680040
By adding `CONFIG` to the `find_package` call, CMake should load the respective protobuf cmake files, which would then also link the program correctly against the abseil libraries and other possible dependencies of protobuf.

However, since adding this (and changing the version number from 3.0.0 to 23.3.0, because it complained), the build fails at the `Genering protobuf dependencies` step with:
```
[ 47%] Generating protobuf files
/bin/sh: line 1: -I: command not found
make[2]: *** [control-plane/CMakeFiles/controlplane-gen.dir/build.make:78: control-plane/google/rpc/status.pb.cc] Error 127
make[1]: *** [CMakeFiles/Makefile2:1871: control-plane/CMakeFiles/controlplane-gen.dir/all] Error 2
make: *** [Makefile:166: all] Error 2
```

Running make with `-d` for debug mode shows the following:
```
[ 47%] Generating protobuf files
Reaping winning child 0x561bb9b00500 PID 192167 
[...]

PR #4056, opened 2023-07-09
[Wip] Install Protobuf using FetchContent.

fruffy:


CMakeLists.txt
```diff
@@ -126,6 +130,8 @@ if (BUILD_STATIC_RELEASE)
   # Do not bring in dynamic libstdcc and libgcc
   set(CMAKE_EXE_LINKER_FLAGS "-static -static-libgcc -static-libstdc++ -Wl,-z,muldefs")
   add_definitions(-DP4C_STATIC_BUILD)
+else ()
+  set(BUILD_SHARED_LIBS ON)
```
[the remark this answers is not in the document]
fruffy: Internal libraries are already linked statically since we only build them this way, e.g. 
https://github.com/p4lang/p4c/blob/main/ir/CMakeLists.txt#L66

The intention was to allow Protobuf to be linked dynamically to reduce binary size as per the discussion in #4055. But it is probably more prudent to always package it in to avoid lookup issues even though that may mean we get a huge binary.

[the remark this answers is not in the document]
vlstill: I see. How much does the protobuf increase binary size? But I would still suggest using static protobuf.

cmake/Protobuf.cmake
```diff
@@ -0,0 +1,76 @@
...
+      fetchcontent_declare(
+        protoc
+        URL https://github.com/protocolbuffers/protobuf/releases/download/v${P4C_PROTOBUF_VERSION}/protoc-${P4C_PROTOBUF_VERSION}-linux-x86_64.zip
+        URL_HASH SHA256=3a4c1e5f2516c639d3079b1586e703fc7bcfa2136d58bda24d1d54f949c315e8
+        USES_TERMINAL_DOWNLOAD TRUE
+        GIT_PROGRESS TRUE
+      )
+    endif()
+
+    fetchcontent_makeavailable(protoc)
```
vlstill, line 61: should protoc be installed?

tools/install_mac_deps.sh
```diff
@@ -6,20 +6,15 @@ if [[ ! -x $BREW ]]; then
...
-      libtool openssl pkg-config python coreutils grep ${PROTOBUF_LIB}
+      libtool openssl pkg-config python coreutils grep
 
-# Prefer Homebrew's bison, grep, and protobuf over the macOS-provided version
-$BREW link --force bison grep ${PROTOBUF_LIB} ${BOOST_LIB}
+# Prefer Homebrew's bison and grep over the macOS-provided version
+$BREW link --force bison grep ${BOOST_LIB}
 echo 'export PATH="/usr/local/opt/bison/bin:$PATH"' >> ~/.bash_profile
-echo 'export PATH="/usr/local/opt/${PROTOBUF_LIB}/bin:$PATH"' >> ~/.bash_profile
 eecho 'export PATH="/usr/local/opt/grep/libexec/gnubin:$PATH"' >> ~/.bash_profile
```
vlstill, line 18: I know you did not change this code, but unless there is some nonstandard `eecho` on macOS, this is a typo and therefore does not work.

vlstill (approved):

cmake/Protobuf.cmake
fruffy: This just downloads the binary packages, the installation of protoc takes a bit. `make install` will not include protoc since it is not a CMake package.

tools/install_mac_deps.sh
fruffy: Will do this in a separate PR.

pkotikal (approved):
Thank you!

cmake/Protobuf.cmake
```diff
@@ -0,0 +1,76 @@
...
+      fetchcontent_declare(
+        protoc
+        URL https://github.com/protocolbuffers/protobuf/releases/download/v${P4C_PROTOBUF_VERSION}/protoc-${P4C_PROTOBUF_VERSION}-osx-x86_64.zip
+        USES_TERMINAL_DOWNLOAD TRUE
+        GIT_PROGRESS TRUE
+      )
+    else()
+      fetchcontent_declare(
+        protoc
+        URL https://github.com/protocolbuffers/protobuf/releases/download/v${P4C_PROTOBUF_VERSION}/protoc-${P4C_PROTOBUF_VERSION}-linux-x86_64.zip
```
rst0git, line 54: @fruffy This seems to break the build for architectures other than x86_64 (e.g., aarch64).

fruffy: Afaik we do not support or make guarantees for any architecture other than `x86_64`. But if there is a need this should be easily patchable.

rst0git: p4c is also used on ARM64 platforms like [P4EDGE](https://p4edge.net/) and [P4Pi](https://github.com/p4lang/p4pi). It would be great if we could also enable support for the `aarch64` architecture.

fruffy: Using the option `P4C_USE_PREINSTALLED_PROTOBUF=ON` will use the system's Protobuf version. In that case things should compile as usual until we have a patch for these types of architectures in place.

rst0git: Thank Fabian!

Some platforms are not connected to the internet and FetchContent would fail when building p4c. As a workaround, we can run `cmake` on a system with internet access and include `build/_deps` with the source code.

Do you have any advice on how to enable building p4c offline?

davidbolvansky: I dont think that building without internet is a supported usecase.

fruffy: There are ways to run FetchContent fully disconnected https://cmake.org/cmake/help/latest/module/FetchContent.html#command:fetchcontent_populate
At least I have been in scenarios where I did not have internet and was able to build as usual.

Now, populating `build/deps` in the beginning is a different story. But that is no different that running `git submodule init --update`. I do not have a good suggestion there yet other than using the PREINSTALLED options.

PR #4060, opened 2023-07-11
Update the dash-pipeline programs.

fruffy:
Add the PNA version of the dash program, too.


jafingerhut:
This PR has conflicts that must be resolved.  I don't know the root cause of the failing CI builds, but at least resolving the conflicts would trigger running them again, and we can see what fails with latest version of repo rather than the last run which appears to be over a month ago.

fruffy:
> This PR has conflicts that must be resolved. I don't know the root cause of the failing CI builds, but at least resolving the conflicts would trigger running them again, and we can see what fails with latest version of repo rather than the last run which appears to be over a month ago.

The failures happen because I did not update the reference files for the updated dash program. Once the dash program is fixed, I will update this PR again.

jafingerhut:
Maybe this is already there and I missed it, but it might be nice to have a README file in the p4_16_samples/dash directory that describes exact steps that can be used to create the version of the DASH P4 code in this commit, e.g. something like:
```
git clone https://github.com/sonic-net/DASH
cd DASH
git checkout some_commit_SHA_value
```
plus any `p4test` command with `-DDEFINE_ME` options that I am guessing you might have used?

jafingerhut (approved):
LGTM

jafingerhut:
Does p4testgen give some kind of error, or cause CI tests to fail, if the "dead code" is left in?  I see you have it commented out.

fruffy:
No, no in this case. It just does not achieve 100% coverage for the DPDK program.

PR #4061, opened 2023-07-12
[P4Testgen] Update the plotting script.

fruffy-bfn:


jnfoster (approved):
LGTM. My only quuestion: does this need to be in the p4c repo? Is it generally useful?

PR #4063, opened 2023-07-13
Avoid using reserved identifiers as include guards

vlstill:
Batch-processed by

```bash
$ git ls-tree --full-tree --name-only -r HEAD | grep '[.]h$' | while read filename; do  sed -Ei -e 's/^(#(define|ifndef)\s+)_([A-Z0-9_]+_H_)$/\1\3/' -e 's,^(#endif\s+)(/\*\s+_([A-Z0-9_]+_H_)\s+\*/|//\s+_([A-Z0-9_]+_H_))\s*$,\1/* \3\4 */,' $filename; done
```

Identifiers of the following forms are reserved:

- identifiers with a double underscore anywhere;
- identifiers that begin with an underscore followed by an uppercase letter;
- in the global namespace, identifiers that begin with an underscore.

We should not use reserved identifiers. More e.g. at https://en.cppreference.com/w/cpp/language/identifiers

### DIFF
--- a/backends/ebpf/codeGen.h
+++ b/backends/ebpf/codeGen.h
@@ -14,8 +14,8 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-#ifndef _BACKENDS_EBPF_CODEGEN_H_
-#define _BACKENDS_EBPF_CODEGEN_H_
+#ifndef BACKENDS_EBPF_CODEGEN_H_
+#define BACKENDS_EBPF_CODEGEN_H_
 
 #include "frontends/p4/typeMap.h"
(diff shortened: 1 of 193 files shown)


vlstill:
Considering there are 192 changed files I did not check all of them one-by-one. I did scroll through the changes though to hopefully detect anything that does not look like guard change and I think such change would very likely break compilation anyway.

fruffy (approved):

fruffy (commented):
There is also this code
https://github.com/p4lang/p4c/blob/main/tools/ir-generator/irclass.cpp#L107

PR #4068, opened 2023-07-17
Add an option to the DPDK P4C back end to generate TDI builder configurations.

fruffy-bfn:
This PR adds the parameter `--tdi-builder-conf` to the p4c-dpdk compiler. This parameter emits the configuration necessary for the `tdi-pipeline_builder`. I am not a direct user, so I simply guessed the parameters necessary. Please let me know if there is anything else to add or remove.

Sidenote:

The `p4c-dpdk` compiler now has about four to five different parameters to create a configuration and output file.
```
--bf-rt-schema file        Generate and write BF-RT JSON schema to the specified file
-o outfile                 Write output to outfile
--tdi file                 Generate and write TDI JSON to the specified file
--context file             Generate and write context JSON to the specified file
--tdi-builder-conf file    Generate and write the TDI builder configuration to the specified file
```

 It might make sense to consolidate some of these?

### DIFF
--- a/backends/dpdk/tdiConf.cpp
+++ b/backends/dpdk/tdiConf.cpp
@@ -22,20 +22,25 @@ limitations under the License.
 namespace DPDK {
 
 void TdiBfrtConf::generate(DPDK::DpdkOptions &options) {
+    if (options.outputFile.isNullOrEmpty()) {
+        ::error(ErrorType::ERR_UNEXPECTED,
+                "No output file provided. Unable to generate correct TDI builder config file.");
+        return;
+    }
+    auto inputFile = std::filesystem::path(options.file.c_str());
     auto outFile = std::filesystem::path(options.outputFile.c_str());
-    auto tdiFile = std::filesystem::path(options.tdiFile.c_str());
-    auto programName = outFile.stem();
     auto outDir = outFile.parent_path();
-    auto configFile = outFile.replace_filename("spec");
+    auto tdiFile = std::filesystem::path(options.tdiBuilderConf.c_str());
+    auto programName = inputFile.stem();
 
-    if (options.bfRtSchema == nullptr) {
+    if (options.bfRtSchema.isNullOrEmpty()) {
         options.bfRtSchema = (outDir / programName).replace_filename("json").c_str();
         ::warning(
             "BF-Runtime Schema file name not provided, but is required for the TDI builder "
             "configuration. Generating file %1%",
             options.bfRtSchema);
     }
-    if (options.ctxtFile == nullptr) {
+    if (options.ctxtFile.isNullOrEmpty()) {
         options.ctxtFile = (outDir / "context.json").c_str();
         ::warning(
             "DPDK context file name not provided, but is required for the TDI builder "
@@ -48,8 +53,7 @@ void TdiBfrtConf::generate(DPDK::DpdkOptions &options) {
     // TODO: Ideally, this should be a template. We could use Inja, but this adds another
     // dependency.
     std::stringstream ss;
-    ss << R"""(
-{
+    ss << R"""({
     "chip_list": [
         {
             "chip_family": "dpdk"
@@ -63,16 +67,18 @@ void TdiBfrtConf::generate(DPDK::DpdkOptions &options) {
                 {
 )""";
     ss << R"""(                    "program-name" : )""";
-    ss << "\"" << programName << "\",";
+    ss << "" << programName << ",\n";
     ss << R"""(                    "bfrt-config" : )""";
-    ss << "\"" << bfRtSchema << "\",";
+    ss << "\"" << bfRtSchema << "\",\n";
     ss << R"""(                    "p4_pipelines": [
-                        {)""";
+                        {
+)""";
     ss << R"""(                            "p4_pipeline_name": "pipe",
                             "context": )""";
     ss << "\"" << contextFile << "\",";
-    ss << R"""(                            "config": )""";
-    ss << "\"" << configFile << "\",";
+    ss << R"""(
+                            "config": )""";
+    ss << "" << outFile << ",";
     ss << R"""(
                             "pipe_scope": [
                                 0,
--- a/backends/dpdk/tdiConf.cpp
+++ b/backends/dpdk/tdiConf.cpp
@@ -1,0 +1,98 @@
+/* Copyright 2023 Intel Corporation
+
+Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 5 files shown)


jafingerhut:
Out of curiosity, would this new generated conf file contain full absolute path names to the other files that it references?  Or if relative, relative to what directory?

fruffy:
> Out of curiosity, would this new generated conf file contain full absolute path names to the other files that it references? Or if relative, relative to what directory?

Currently, the conf file simply contains the path names which were provided as arguments to `p4c-dpdk`. With the assumption that the user provided them correctly. However, I agree that some sanitizing and absolute paths would be better. 

Btw, let me know whether this file captures the expected workflow. I also noticed the compiler has a `--tdi` option but it is unclear what it is used for. There seems to be some redundancy regarding the parameters that are available.

backends/dpdk/tdiConf.cpp
```diff
@@ -0,0 +1,104 @@
...
+)""";
+    ss << R"""(                            "p4_pipeline_name": "pipe",
+                            "context": )""";
+    ss << contextFile << ",";
+    ss << R"""(
+                            "config": )""";
+    ss << outFile << ",";
+    ss << R"""(
+                            "pipe_scope": [
+                                0,
```
Hoooao, line 84: the `tdi_pipeline_builder` cannot parse the ending `,` in json.

usha1830:
> Btw, let me know whether this file captures the expected workflow. I also noticed the compiler has a `--tdi` option but it is unclear what it is used for. There seems to be some redundancy regarding the parameters that are available.
--tdi option generates the tdi.json file. This is same as bfrt.json with some additional program level information such as program name, build date, compile command, compiler version etc. This is different from the .conf file.

backends/dpdk/tdiConf.cpp
```diff
@@ -0,0 +1,104 @@
...
+            "configuration. Generating file %1%",
+            options.bfRtSchema);
+    }
+    if (options.ctxtFile.isNullOrEmpty()) {
+        options.ctxtFile = (outDir / "context.json").c_str();
+        ::warning(
+            "DPDK context file name not provided, but is required for the TDI builder "
+            "configuration. Generating file %1%",
+            options.ctxtFile);
+    }
```
jafingerhut, line 49: So the behavior here is that if you give the command line option to generate the TDI pipeline builder conf file, and also the command line options to specify file names for bfrt scheme and context json files, it will use those file names in the generated conf file?

And if you do not specify the options on the command line to generate those files, but do give the option to generate the TDI pipeline builder conf file, then it will issue warnings, and make up its own default file names for those other output files, and generate them anyway?

If so, that seems reasonable to me.

fruffy-bfn: > So the behavior here is that if you give the command line option to generate the TDI pipeline builder conf file, and also the command line options to specify file names for bfrt scheme and context json files, it will use those file names in the generated conf file?
> 
> And if you do not specify the options on the command line to generate those files, but do give the option to generate the TDI pipeline builder conf file, then it will issue warnings, and make up its own default file names for those other output files, and generate them anyway?
> 
> If so, that seems reasonable to me.

Yes. I considered throwing an error, but it seemed more convenient to generate all the files in one go with one parameter.

jafingerhut (approved):
LGTM.  I haven't tested it yet, but can give it a go after it is merged and file an issue if anything seems amiss.  Nothing seems wrong from a code inspection.

usha1830:
> > > --tdi option generates the tdi.json file. This is same as bfrt.json with some additional program level information such as program name, build date, compile command, compiler version etc. This is different from the .conf file.
> 
> What is `tdi.json` file used for?

I am not sure if it is being used. This was added to be used by TDI replacing the bfrt.json. But I believe bfrt.json is the which is being used by p4-dpdk-target as I have never received any query related to tdi.json.

fruffy:
> I am not sure if it is being used. This was added to be used by TDI replacing the bfrt.json. But I believe bfrt.json is the which is being used by p4-dpdk-target as I have never received any query related to tdi.json.

Should we remove it then? We can easily add the option back if a request comes up. In the meantime, it might help simplify the possible options for the compiler.

backends/dpdk/tdiConf.cpp
```diff
@@ -0,0 +1,104 @@
...
+                {
+)""";
+    ss << R"""(                    "program-name" : )""";
+    ss << programName << ",\n";
+    ss << R"""(                    "bfrt-config" : )""";
+    ss << bfRtSchema << ",\n";
+    ss << R"""(                    "p4_pipelines": [
+                        {
+)""";
+    ss << R"""(                            "p4_pipeline_name": "pipe",
```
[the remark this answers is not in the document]
fruffy: There doesn't seem a way to infer this information for PNA, only for PSA. Let me think what to do about this.

usha1830:
> > I am not sure if it is being used. This was added to be used by TDI replacing the bfrt.json. But I believe bfrt.json is the which is being used by p4-dpdk-target as I have never received any query related to tdi.json.
> 
> Should we remove it then? We can easily add the option back if a request comes up. In the meantime, it might help simplify the possible options for the compiler.

@jamescchoi Does TDI needs tdi.json to be emitted from compiler?

usha1830 (approved):

jafingerhut:
@fruffy @usha1830 See previous comment, in case you are curious.

saynb:
> I guess that means eventually we can remove `--bf-rt-schema`? Or even now?

Removing it will result in backward incompatibility in a lot of places. Can we come up with a phase out plan? Does the `--tdi` option already exist today and it generates a new file? We want to separate out BFRT and TDI for the reasons @jafingerhut  mentioned. Today we use the bf-rt-schema everywhere even for TDI in most targets but we would prefer to using `--tdi`. We would also want to move to using a schema validator present in p4lang/tdi rather than one which resides here but that can be taken as a later task.

fruffy-bfn:
>Removing it will result in backward incompatibility in a lot of places.

What are the places? We can probably just mark it deprecated, issue a warning and eventually remove it. We really should get rid of bf-rt-schema because the name is confusing. 

In the meantime we can unify the options to do the same thing.  The tdi option exists already, but it does not seem to be used.

>We would also want to move to using a schema validator present in p4lang/tdi rather than one which resides here but that can be taken as a later task.

Which schema validator are you referring to?

saynb:
> Which schema validator are you referring to?

It is a WIP as a PR. That's why I mentioned it as a separate task

PR #4084, opened 2023-07-26
Fix missing header for intmax_t typedef

felipealmeida:
Add include directive for cstdint which is necessary to use std::intmax_t, also qualify intmax_t with std namespace because of the use of C++ header instead of C.


fruffy:
Can you fix the linter errors?

felipealmeida:
Yes. I'll do it.

felipealmeida:
Done

fruffy (approved):

PR #4088, opened 2023-08-01
Use libbacktrace for stack dumps if available

ChrisDodd:
- Memory allocation tracing hooks to track where memory is being allocated and print summary with stack trace of each allocation site.  Used when logging in ComputeWriteSet to understand how/why it is using so much memory.

### DIFF
--- a/BUILD.bazel
+++ b/BUILD.bazel
@@ -17,7 +17,8 @@ genrule(
     outs = ["config.h"],
     cmd = " | ".join([
         "sed 's|cmakedefine|define|g' < $(SRCS)",
-        "sed 's|define HAVE_LIBGC 1|undef HAVE_LIBGC|g' > $(OUTS)",
+        "sed 's|define HAVE_LIBGC 1|undef HAVE_LIBGC|g'",
+        "sed 's|define HAVE_LIBBACKTRACE 1|undef HAVE_LIBBACKTRACE|g' > $(OUTS)",
     ]),
     visibility = ["//visibility:private"],
 )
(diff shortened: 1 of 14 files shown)


ChrisDodd:
Turns out with PIE executables (which Ubuntu 20.04 uses by default), addr2line is completely broken and can't produce a useful backtrace.  The libbacktrace library from Github is a good alternative -- produces nice dumps and even deals with inlined calls, and is more efficient to boot.  We don't really want a dependency on it, but using it automatically when it is installed is a good compromise.

mihaibudiu (approved):

jafingerhut:
Has anyone analyzed the reasons for the CI test failures?

fruffy:
> Has anyone analyzed the reasons for the CI test failures?

Likely a missing include, which is why the unity builds pass. The sanitizer failure looks like a `clang` compilation issue.

fruffy:
Test failures are because I added `GIT_SHALLOW` to the p4runtime submodule pull. This will cause issues when trying to checkout an older commit. #4096 will fix this.

jafingerhut:
@chrisdodd Looks like 3 of the CI builds are still failing even after https://github.com/p4lang/p4c/pull/4092 was merged in last week., 2 of them related to Bazel builds.  Is this something that someone at Google might be able to help with?

fruffy:
> @ChrisDodd Looks like 3 of the CI builds are still failing even after #4092 was merged in last week., 2 of them related to Bazel builds. Is this something that someone at Google might be able to help with?

I believe the failure is a Clang problem, since the clang build also fails. Bazel uses Clang by default iirc.

Issue #1122, opened 2017-12-22
EBPF back-end byte order is inconsistent


mbudiu-vmw:
The C translation of the P4 code does not use a consistent byte-order.

Note: there is a plan to rework the ebpf back-end to change the C representation of data to be packed by default and to unpack data and convert to host order only when necessary.

PR #4089, opened 2023-08-01
Fix MAC Address endianness issue

komaljai:
"dstAddr" and "srcAddr" fields are of type "macaddr", which is stored in sk_buff in little endian form.
load_dword will convert big_endian to little_endian and load it to cpu. The issue is instead of calling load_dword (which does endianess conversion), we are calling __builtin_memcpy when there is annotation 'macaddr'


backends/tc/ebpfCodeGen.cpp
```diff
@@ -407,14 +407,170 @@ void TCIngressPipelinePNA::emitLocalVariables(EBPF::CodeBuilder *builder) {
...
 
+void PnaStateTranslationVisitor::compileExtractField(const IR::Expression *expr,
+                                                  const IR::StructField *field, unsigned alignment,
+                                                  EBPF::EBPFType *type) {
+    unsigned widthToExtract = dynamic_cast<EBPF::IHasWidth *>(type)->widthInBits();
+    auto program = state->parser->program;
+    cstring msgStr;
+    cstring fieldName = field->name.name;
+
+    bool checkIfMAC = false;
```
fruffy, line 433: The eBPF back end has no knowledge of "MAC" . Rather it looks like you are trying to match on a specific bit-width?

backends/tc/ebpfCodeGen.cpp
```diff
@@ -407,14 +407,170 @@ void TCIngressPipelinePNA::emitLocalVariables(EBPF::CodeBuilder *builder) {
...
+    }
+
+    builder->emitIndent();
+    builder->appendFormat("%s += %d", program->offsetVar.c_str(), widthToExtract);
+    builder->endOfStatement(true);
+
+    // eBPF can pass 64 bits of data as one argument passed in 64 bit register,
+    // so value of the field is printed only when it fits into that register
+    if (widthToExtract <= 64) {
+        cstring exprStr = expr->is<IR::PathExpression>()
```
fruffy, line 552: I believe you can use `toString` in both cases here. The outcome is the same.

backends/tc/ebpfCodeGen.cpp
```diff
@@ -407,14 +407,170 @@ void TCIngressPipelinePNA::emitLocalVariables(EBPF::CodeBuilder *builder) {
...
+                              ? expr->to<IR::PathExpression>()->path->name.name
+                              : expr->toString();
+
+        if (expr->is<IR::Member>() && expr->to<IR::Member>()->expr->is<IR::PathExpression>() &&
+            isPointerVariable(
+                expr->to<IR::Member>()->expr->to<IR::PathExpression>()->path->name.name)) {
```
fruffy, line 558: Use `checkedTo` instead of `to.

fruffy (commented):
This could fix #1122 but it looks like it only applies to the parser extraction right now. I am not quite sure about this implementation. What motivated this change?

Can this be implemented in a dedicated compiler pass?

backends/tc/ebpfCodeGen.cpp
komaljai: This change is not in eBPF compiler, but in P4TC. If user gives annotation "macaddr", the expectation is not to convert the endianness. This is not on specific bit width, but annotation "macaddr".

backends/tc/ebpfCodeGen.cpp
komaljai: Addressed.

backends/tc/ebpfCodeGen.cpp
komaljai: Addressed.

backends/tc/ebpfCodeGen.cpp
```diff
@@ -408,13 +408,175 @@ void TCIngressPipelinePNA::emitLocalVariables(EBPF::CodeBuilder *builder) {
...
+        cstring exprStr = expr->is<IR::PathExpression>()
+                              ? expr->to<IR::PathExpression>()->path->name.name
+                              : expr->toString();
+
+        if (expr->is<IR::Member>() &&
+            expr->checkedTo<IR::Member>()->expr->is<IR::PathExpression>() &&
+            isPointerVariable(
+                expr->to<IR::Member>()->expr->checkedTo<IR::PathExpression>()->path->name.name)) {
+            exprStr = exprStr.replace(".", "->");
+        }
```
fruffy, line 566:
```suggestion
        if (auto member = expr->to<IR>Member>()) {
            if (auto pathExpr = member->expr->to<IR::PathExpression()) {
                if (isPointerVariable(pathExpr->name.name)) {
                    exprStr = exprStr.replace(".", "->");
                }
            }
        }
```

backends/tc/ebpfCodeGen.cpp
```diff
@@ -408,13 +408,175 @@ void TCIngressPipelinePNA::emitLocalVariables(EBPF::CodeBuilder *builder) {
...
+    builder->emitIndent();
+    builder->appendFormat("%s += %d", program->offsetVar.c_str(), widthToExtract);
+    builder->endOfStatement(true);
+
+    // eBPF can pass 64 bits of data as one argument passed in 64 bit register,
+    // so value of the field is printed only when it fits into that register
+    if (widthToExtract <= 64) {
+        cstring exprStr = expr->is<IR::PathExpression>()
+                              ? expr->to<IR::PathExpression>()->path->name.name
+                              : expr->toString();
```
fruffy, line 559:
```suggestion
        cstring exprStr = expr->toString();
```

backends/tc/ebpfCodeGen.cpp
fruffy: In general, I would recommend checking early that `expr` is either a path expression or a member and then resolve variables as needed. Do some checking that it needs to be either of these values then you can avoid these casts.

testdata/p4tc_samples_outputs/global_action_example_02.c
```diff
@@ -208,10 +208,10 @@ static __always_inline int process(struct __sk_buff *skb, struct my_ingress_head
                 goto reject;
             }
 
-            hdr->ethernet.dstAddr = (u64)((load_dword(pkt, BYTES(ebpf_packetOffsetInBits)) >> 16) & EBPF_MASK(u64, 48));
+            __builtin_memcpy(&hdr->ethernet.dstAddr, pkt + BYTES(ebpf_packetOffsetInBits), 6);
             ebpf_packetOffsetInBits += 48;
 
-            hdr->ethernet.srcAddr = (u64)((load_dword(pkt, BYTES(ebpf_packetOffsetInBits)) >> 16) & EBPF_MASK(u64, 48));
+            __builtin_memcpy(&hdr->ethernet.srcAddr, pkt + BYTES(ebpf_packetOffsetInBits), 6);
```
fruffy, line 214: Is this is a portable function? I believe clang requires language extensions to support this particular memcpy version.

fruffy: Would be nice to use a less compiler-specific version here.

usha1830 (approved):
LGTM

testdata/p4tc_samples_outputs/global_action_example_02.c
komaljai: __builtin_memcpy is commonly used in eBPF code (not only P4-ebpf). There are many instances in same code where __builtin_memcpy is used.

backends/tc/ebpfCodeGen.cpp
komaljai: Addressed

fruffy (approved):

jafingerhut:
@komaljai Have you been able to get your status with the ONF CLA checker to pass yet?

usha1830:
> @jafingerhut Not yet, I am yet to be added in CLA.

Merging this as the CLA check is passing now.

PR #4092, opened 2023-08-02
Update GoogleTest and install it via FetchContent

fruffy:
- Update the GoogleTest version we are using, it was almost 6 years old which makes using newer features difficult. 
- Also use FetchContent to pull it in.
- Remove some BMv2 code from the top-level gtest folder. It should be in the back end. 
- Fix up some includes since GTEST is not always enabled.


jnfoster (approved):
LGTM

PR #4096, opened 2023-08-05
[Hotfix] Remove git shallow from p4runtime FetchContent

fruffy:
The problem with `GIT_SHALLOW` is that whenever the submodule updates, pulling will fail because the commit has not been stored locally. This can break CI.


fruffy:
Closing in favour of #4092.

PR #4103, opened 2023-08-09
[P4Testgen] Encode more P4Runtime constraints for the behavioral model

fruffy:
See also https://github.com/p4lang/p4-constraints/pull/101

### DIFF
--- a/backends/p4tools/modules/testgen/targets/bmv2/table_stepper.cpp
+++ b/backends/p4tools/modules/testgen/targets/bmv2/table_stepper.cpp
@@ -8,6 +8,7 @@
 
 #include <boost/multiprecision/cpp_int.hpp>
 
+#include "backends/p4tools/common/lib/constants.h"
 #include "backends/p4tools/common/lib/trace_event_types.h"
 #include "backends/p4tools/common/lib/variables.h"
 #include "ir/declaration.h"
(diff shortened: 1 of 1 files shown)


jafingerhut:
Is there a brief description of this change you can give?  From scanning the most dense part of the code changes, it appears that you are adding support for ternary fields, and perhaps also adding the ability of optional match_kind keys to be all-wildcard, whereas before this change you only ever created all-exact-match table entries for optional match_kind?

fruffy:
This change has been prompted by https://github.com/p4lang/p4-constraints/pull/100. We are trying to align more with the p4runtime specification in terms of how we generate our table entries. This required some changes to the optional matches we are making and how we produce ternary matches. This PR encodes this in the form of additional constraints.

jafingerhut:
@fruffy So I do not have the necessary context to do what I would think of as a proper code review for this kind of change in p4testgen, and I am not sure I have the time required to gain that context any time soon.  By proper code review, I mean thinking about the logic of the changes, and understanding whether they seem correct, or have some bug in them.  I know the P4Runtime API constraints pretty well, but I don't think that is enough.

Would you prefer a rubber-stamp LGTM that means "yeah, I see CI tests did not break", or someone who can give a proper review?

pkotikal (approved):

PR #4105, opened 2023-08-13
Run Gauntlet validation, sanitizers, and Ubuntu 18 build/test as nightly actions.

fruffy:
Save some CI resources on PRs.

### DIFF
--- a/.github/workflows/ci-test.yml
+++ b/.github/workflows/ci-test.yml
@@ -75,39 +75,6 @@ jobs:
       working-directory: ./build
       if: matrix.unity == 'ON'
 
-  # Build with clang and test p4c on Ubuntu 20.04.
-  test-ubuntu20-clang-sanitizers:
-    strategy:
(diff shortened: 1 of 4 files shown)


jafingerhut (approved):
LGTM.  I guess the resource you are most concerned with reducing here is the time spent waiting while the diffs in a PR are tested?

fruffy:
@jnfoster The PR is blocked because `validate` is a required run. Can you disable it? In general, we should update the required CI runs.

jafingerhut:
@jnfoster Do you know who the set of people is that has authorization to make the required steps for this to pass CI checks?

If it does not include @fruffy, I would be in favor of adding him to that set.  Also anyone else you think would make sense to add to that set.

fruffy:
> After discussing with the (informal) p4c maintainers, I've bumped up @fruffy's privileges on this repository. He should be able to modify CI configuration settings now.

Thank you!

PR #4108, opened 2023-08-14
Allow visitor ref to process IR nodes

vlstill:
This small change allows `PassManager::VisitorRef` to be applied without putting it into a pass manager first. This is motivated by the following pattern of hiding pass implementations yet allowing them to be tested separately.

file `some_feature.h`:

```cpp
PassManager::VisitorRef makeFoo(...);
PassManager::VisitorRef makeBar(...);

struct SomeFeature : Passmanager {
    SomeFeature() : PassManager({makeFoo(), makeBar()}) { }
};
```

file `some_feature.cpp`

```cpp
struct Foo : Transform {
...
};

PassManager::VisitorRef makeFoo(...) { return new Foo(...); }

...
```

As you can see, this allows completely hiding visitor implementation in a `.cpp` file, not exposing its private members and overload list in the header. At the same time the `make*` functions, together with this change, allow testing of the constituent visitors.


ir/pass_manager.h
```diff
@@ -89,6 +89,13 @@ class PassManager : virtual public Visitor, virtual public Backtrack {
                       t();
                       return n;
                   })) {}
+
+        /// Apply the visitor held in this VisitorRef on a given IR root node. It returns the
+        /// modified node.
+        const IR::Node *process(const IR::Node *root) {
+            CHECK_NULL(visitor);
+            return root->apply(*visitor);
+        }
```
ChrisDodd, line 98: Better to define

    Visitor *operator->() const { return visitor; }

then you can just call `->apply` (or anything else) on a VisitorRef

Maybe also define

    explicit operator bool() const { return visitor != nullptr; }

ChrisDodd (commented):
I don't really see a need/use for this?  Just have your `makeFoo`/`makeBar` functions return a `Visitor *` rather than a `VisitorRef`.  The point of `VisitorRef` is to allow lambdas to be used in the `initializer_list` for a `PassManager` (by automatically sticking them in a `VisitFunctor` instance)

PR #4111, opened 2023-08-15
Use namespace for dpdk PTF tests

Hoooao:
Previously the test driver `run-dpdk-ptf-test.py` modified the host's network config which caused side-effects like disabling IPv6. Using namespace, which would be cleaned up after the test, resolved this. Code is mostly from `run-bmv2-ptf-test.py`.

### DIFF
--- a/backends/dpdk/run-dpdk-ptf-test.py
+++ b/backends/dpdk/run-dpdk-ptf-test.py
@@ -20,6 +20,10 @@ TOOLS_PATH = FILE_DIR.joinpath("../../tools")
 sys.path.append(str(TOOLS_PATH))
 import testutils
 
+BRIDGE_PATH = FILE_DIR.joinpath("../ebpf/targets")
+sys.path.append(str(BRIDGE_PATH))
+from ebpfenv import Bridge
+
 PARSER = argparse.ArgumentParser()
 PARSER.add_argument("p4c_dir", help="The location of the the compiler source directory")
 PARSER.add_argument("p4_file", help="the p4 file to process")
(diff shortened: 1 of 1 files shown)


backends/dpdk/run-dpdk-ptf-test.py
```diff
@@ -81,21 +85,12 @@
...
-
-
-# We want to disable IPv6 to avoid ICMPv6 spam
-def diable_IPv6():
-    testutils.exec_process("sysctl -w net.ipv6.conf.all.disable_ipv6=1")
-    testutils.exec_process("sysctl -w net.ipv6.conf.default.disable_ipv6=1")
-    # Also filter igmp packets, -w is necessary because of a race condition
-    testutils.exec_process("iptables -w -A OUTPUT -p 2 -j DROP")
-    return testutils.SUCCESS
+def is_port_alive(ns, port) -> bool:
```
fruffy, line 88: Is there a more pythonic way to do this? Something where we do not have to call netstat?

For example, like this check:
https://github.com/p4lang/p4c/blob/main/tools/testutils.py#L134

fruffy (approved):

backends/dpdk/run-dpdk-ptf-test.py
fruffy: We have utilities for this. 
`bridge.ns_exec` should take care of executing in the right namespace.

Hoooao: `ns_exec` simply appends `ip netns exec <ns>` before the cmd. It cannot handle calling python functions in an easy way.

fruffy: Oh I see what you mean, okay let's leave it like this then. Just use `bridge.ns_exec` instead.

jafingerhut:
Merging this.

PR #4113, opened 2023-08-22
eBPF backend: generate P4Runtime files if required

tatry:
Allow to generate P4Runtime files by ebpf backend, because options related to them are ignored.

Supports only PSA architecture. For ebpf filer model options for generating P4info files are (still) ignored.

### DIFF
--- a/backends/ebpf/p4c-ebpf.cpp
+++ b/backends/ebpf/p4c-ebpf.cpp
@@ -73,8 +73,10 @@ void compile(EbpfOptions &options) {
         if (::errorCount() > 0) return;
     }
 
-    P4::serializeP4RuntimeIfRequired(program, options);
-    if (::errorCount() > 0) return;
+    if (!options.arch.isNullOrEmpty() && options.arch != "filter") {
+        P4::serializeP4RuntimeIfRequired(program, options);
+        if (::errorCount() > 0) return;
+    }
 
     EBPF::MidEnd midend;
     midend.addDebugHook(hook);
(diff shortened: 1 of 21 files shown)


fruffy (approved):
One thing to also do is add golden tests here. Keep around reference P4Runtime files to check whether the output format is correct.

osinstom (approved):

jafingerhut:
@tatry Are there existing automated CI tests that check for eBPF whether the output P4Info files contain the expected contents?  As far as I know, the P4Info file generation is currently the same regardless of the target, and covers the v1model, PSA, and PNA architectures equally well.

tatry:
@fruffy 
> Are there existing automated CI tests that check for eBPF whether the output P4Info files contain the expected contents?

It depends on architecture:
- eBPF filter model use *standard* CI tests, so it should not a big problem to add them (at least I think that). I verified here one thing: this model is not supported by `P4RuntimeSerializer`. I didn't tested with option `--arch filter`, so I should disable generation P4info for eBPF filter model. But without that option P4info for `v1model` is generated, that's why I thought it is supported.
- PSA architecture use custom PTF tests. There will be more work to add such functionality, because these tests have no support to compare P4info files - they are focused on packets rather on files.

I think that I can manage with these issues within this PR.

tatry:
I've added some P4info files as examples and reference for tests. I think it is ready for re-review now.

jafingerhut:
This has 2 approving reviews, and passing all CI tests, so will merge it in as is.  If this was too early, sorry for the confusion, and I guess we can create additional issues/PRs to improve things further.

PR #4116, opened 2023-08-23
Migrate Automatic Exec Groups by adding a toolchain parameter to the affected actions

kotlaja:
This is a step forward for the full migration of Automatic Exec Groups (AEGs). This change will be effective once AEGs are enabled.

Goal of this migration:
* Each toolchain type should have its own execution platform. Before this change, the execution platform was selected on a rule base, not on a toolchain type base.
* Each action which uses a tool coming from a toolchain should specify the toolchain type inside ctx.action.{run, run_shell} toolchain attribute.

The affected action uses tools from cpp toolchain. Hence, cpp toolchain is added to the `toolchain` param.

### DIFF
--- a/bazel/p4_library.bzl
+++ b/bazel/p4_library.bzl
@@ -1,6 +1,6 @@
 """P4 compilation rule."""
 
-load("@bazel_tools//tools/cpp:toolchain_utils.bzl", "find_cpp_toolchain", "use_cpp_toolchain")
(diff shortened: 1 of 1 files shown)


fruffy (approved):
@smolkaj @kheradmandG

smolkaj:
Thanks for approving, LGTM.

PR #4129, opened 2023-09-03
Allow unknown fields when parsing P4Info files

jafingerhut:
For example, recently the field has_initial_entries was added to P4Info messages in the p4c implementation of generating P4Info files. However, if someone is testing with a version of behavioral-model built from a p4lang/PI version that has not been updated yet, then without the changes in this PR, the Merge() calls will fail if they see such an unknown message field.  With these changes, they will silently allow such unknown fields.

### DIFF
--- a/backends/bmv2/base_test.py
+++ b/backends/bmv2/base_test.py
@@ -219,8 +219,7 @@ class P4RuntimeTest(BaseTest):
         testutils.log.info(f"Reading p4info from {proto_txt_path}")
         self.p4info = p4info_pb2.P4Info()
         with open(proto_txt_path, "rb") as fin:
-            google.protobuf.text_format.Merge(fin.read(), self.p4info,
-                                              allow_unknown_field=True)
+            google.protobuf.text_format.Merge(fin.read(), self.p4info, allow_unknown_field=True)
 
         self.import_p4info_names()
 
@@ -248,8 +247,7 @@ class P4RuntimeTest(BaseTest):
         # config.p4info = self.p4info
         proto_txt_path = ptfutils.test_param_get("p4info")
         with open(proto_txt_path, "r", encoding="utf-8") as fin:
-            google.protobuf.text_format.Merge(fin.read(), config.p4info,
-                                              allow_unknown_field=True)
+            google.protobuf.text_format.Merge(fin.read(), config.p4info, allow_unknown_field=True)
         config_path = ptfutils.test_param_get("config")
         testutils.log.info(f"Reading config (compiled P4 program) from {config_path}")
         with open(config_path, "rb") as config_f:
@@ -1121,8 +1119,7 @@ def update_config(config_path, p4info_path, grpc_addr, device_id):
     request.device_id = device_id
     config = request.config
     with open(p4info_path, "r", encoding="utf-8") as p4info_f:
-        google.protobuf.text_format.Merge(p4info_f.read(), config.p4info,
-                                          allow_unknown_field=True)
+        google.protobuf.text_format.Merge(p4info_f.read(), config.p4info, allow_unknown_field=True)
     with open(config_path, "rb") as config_f:
         config.p4_device_config = config_f.read()
     request.action = p4runtime_pb2.SetForwardingPipelineConfigRequest.VERIFY_AND_COMMIT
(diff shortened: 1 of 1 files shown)


jafingerhut:
@smolkaj @fruffy @antoninbas @chrispsommers Not sure if you have an opinion on a change like this that enables more compatibility when P4Info file contents are generalized in backwards-compatible ways by adding new fields.  On a system that I have tested building open source P4 dev tools on, this change enables many of the bmv2-ptf tests to pass, that fail without these changes.

I know that I could also update the version of p4lang/PI repo that I am building to include the new P4Runtime API additions made recently, but this change seems like it might be useful, too?

mihaibudiu (approved):

jafingerhut:
Only 2 out of all of the CI tests are failing.  They are called:

+ test-p4c / test-ubuntu20-clang-sanitizers
+ test-p4c / test-ubuntu20 (Unity ON)

Both of those run more than 2000 automated tests for p4c that are enabled, and only one of them fails, the one called `bmv2-ptf/testdata/p4_16_samples/ternary2-bmv2.p4`.  There are many WARNING messages in the log showing details of the failure, but it is not clear to me why that one particular test is failing, while all of the others are passing.

Does anyone know?  Are there perhaps multiple versions of the Python protobuf bindings involved in different CI tests, and this one does not support the `allow_unknown_field` optional argument to `google.protobuf.text_format.Merge`, but the other CI tests do allow it?

fruffy:
The error is 
```
 WARNING: ----------------------------------------------------------------------
WARNING: Traceback (most recent call last):
WARNING:   File "/home/runner/work/p4c/p4c/backends/bmv2/base_test.py", line 1098, in handle
WARNING:     return f(*args, **kwargs)
WARNING:   File "/home/runner/work/p4c/p4c/build/tmp2wef0upb/ternary2-bmv2.py", line 32, in setUp
WARNING:     bt.P4RuntimeTest.setUp(self)
WARNING:   File "/home/runner/work/p4c/p4c/backends/bmv2/base_test.py", line 222, in setUp
WARNING:     google.protobuf.text_format.Merge(fin.read(), self.p4info, allow_unknown_field=True)
WARNING: TypeError: Merge() got an unexpected keyword argument 'allow_unknown_field'
WARNING: 
WARNING: During handling of the above exception, another exception occurred:
WARNING: 
WARNING: Traceback (most recent call last):
WARNING:   File "/home/runner/work/p4c/p4c/backends/bmv2/base_test.py", line 1100, in handle
WARNING:     test.undo_write_requests(test._reqs)
WARNING: AttributeError: 'Test1' object has no attribute '_reqs'
```

It is the only BMv2 PTF test we are running across the entire suite. What surprises me is that none of the P4Testgen PTF tests is failing.

jafingerhut:
Don't the Ubuntu 18.04 and 22.04 tests also run this BMv2 PTF test?  And perhaps also macOS?

fruffy:
I think 18.04 and MacOS do not run these tests, but Ubuntu 22.04 does. That might also explain why the tools CI test does not fail. So it must be a difference in Protobuf? We install different bindings for Ubuntu 22.04 here:
https://github.com/p4lang/p4c/blob/main/tools/ci-build.sh#L135

jafingerhut:
@jhavrane If you are now on the approved list for the ONF CLA check, can you make some kind of inconsequential change in another commit on this PR to force the ONF CLA check to be performed again?  If you cannot find an inconsequential change eaisly, then a commit that makes a change, then another that undoes it, would also "poke the server" and make it redo this check.  Sorry, I know how to make Github CI redo all of the other CI tests, but not that one.

fruffy:
@jafingerhut What is the status of this PR? What is the current blocker?

jafingerhut:
@fruffy Several CI tests were failing for reasons that were not obvious to me why, and then I forgot about the PR until you reminder.  I have created essentially the same PR here, since I was not sure how to resolve the conflicts in this one: https://github.com/p4lang/p4c/pull/4341

jafingerhut:
Closing this PR in favor of a more recently created one based on later version of this repo: https://github.com/p4lang/p4c/pull/4341

smolkaj:
Sorry for the late comment on this.

> @smolkaj @fruffy @antoninbas @chrispsommers Not sure if you have an opinion on a change like this that enables more compatibility when P4Info file contents are generalized in backwards-compatible ways by adding new fields.

Ignoring unknown fields is the standard approach in proto, so I think this is the way to go :+1: 
This happens by default with protos binary format, but is opt-in for the text format, which is typically not assumed to be used for IPC where schema versions may differ.

PR #4134, opened 2023-09-06
backends/ebpf: Fix #4098 by renaming conflicting write_partial macro.

### LINKED ISSUE #4098: backends/ebpf: conflicting declarations for `write_partial`
In `ebpfProgram.cpp`, the emitted `write_partial` macro takes 3 arguments, but in `ebpfDeparser.cpp` it is used with 4 arguments, as defined in `psa/ebpfPsaGen.cpp`.

Ideally there should only be a single definition which all call sites can agree on. Alternatively, we could have 2 different versions but they should not share the same name. I think that the 4-arg version is a generalisation of the 3-arg one (but decoding the macro logic is painful!), if so we should probably standardise on that.

CC @osinstom who seems to be the author of the 2nd version.

thomascalvert-xlnx:
As mentioned in the issue, we have conflicting versions of this macro. This patch avoids the issue by renaming the second version.

In an ideal world, all call sites would make use of the second, more general version. But that felt like too invasive a change, so instead I went for a minimalist approach.

The issue never arose before, because the only flow which made use of `ebpfDeparser` was within the PSA target.

### DIFF
--- a/backends/ebpf/ebpfProgram.cpp
+++ b/backends/ebpf/ebpfProgram.cpp
@@ -180,6 +180,24 @@ class ErrorCodesVisitor : public Inspector {
 };
 }  // namespace
 
+void EBPFProgram::emitCommonPreamble(CodeBuilder *builder) {
+    builder->newline();
+    builder->appendLine("#define EBPF_MASK(t, w) ((((t)(1)) << (w)) - (t)1)");
+    builder->appendLine("#define BYTES(w) ((w) / 8)");
+    builder->appendLine(
+        "#define write_partial(a, s, v) do "
+        "{ u8 mask = EBPF_MASK(u8, s); "
+        "*((u8*)a) = ((*((u8*)a)) & ~mask) | (((v) >> (8 - (s))) & mask); "
+        "} while (0)");
+    builder->appendLine(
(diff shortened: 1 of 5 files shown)


backends/ebpf/ebpfProgram.cpp
```diff
@@ -198,6 +199,9 @@ void EBPFProgram::emitPreamble(CodeBuilder *builder) {
         "{ u8 mask = EBPF_MASK(u8, s); "
         "*((u8*)a) = ((*((u8*)a)) & ~mask) | (((v) >> (8 - (s))) & mask); "
         "} while (0)");
+    builder->appendLine(
+        "#define write_partial_ex(a, w, s, v) do { *((u8*)a) = ((*((u8*)a)) "
+        "& ~(EBPF_MASK(u8, w) << s)) | (v << s) ; } while (0)");
```
tatry, line 204: Excluding definition of `memcpy` definitions here looks the same as for PSA arch. I think it would be better to move (or create) `emitCommonPreamble` to `EBPFPrgram` class with this common code. From `PSAEbpfGenerator` it could be called via `ingress` field.

thomascalvert-xlnx: Thanks for the pointers - that is indeed a cleaner way to do things. Rewrote the PR accordingly.

tatry (approved):

PR #4149, opened 2023-09-09
Allow lookahead of structs with size of 0

### LINKED ISSUE #4143: Compiler crash on lookahead call with bit<0> in structure
### Issue
```
  what():  In file: ./p4c/midend/expandLookahead.cpp:46
Compiler Bug: bit<0>: unexpected size 0
```

### P4 program
```
#include <core.p4>

parser Parser(packet_in pkt);
package SimpleArch(Parser p);

struct Foo {
    bit<0> f0;
}

parser MyParser(packet_in packet) {
   state start {
       transition select(packet.lookahead<Foo>().f0) {
           0: accept;
           default: reject;
       }
   }
}

SimpleArch(MyParser()) main;
```

mihaibudiu:
Fixes #4143

### DIFF
--- a/midend/expandLookahead.cpp
+++ b/midend/expandLookahead.cpp
@@ -43,7 +43,7 @@ void DoExpandLookahead::expand(
         }
     } else if (type->is<IR::Type_Bits>() || type->is<IR::Type_Boolean>()) {
         unsigned size = type->width_bits();
-        BUG_CHECK(size > 0, "%1%: unexpected size %2%", type, size);
+        if (size == 0) return;
         const IR::Expression *expression =
             new IR::Slice(bitvector->clone(), *offset - 1, *offset - size);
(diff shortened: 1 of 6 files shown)


jafingerhut (approved):
LGTM.  Such a lookahead is pretty much a no-op, yes?

fruffy:
> LGTM. Such a lookahead is pretty much a no-op, yes?

You could probably write/add a pass to simplify this code, but not sure whether that is worth it.

mihaibudiu:
I think that in the end that's what you'll get

jnfoster:
Can't wait until we announce this new feature at the 2024 P4 Workshop...
<Don Steve Jobs-esque turtleneck>
Announcing... 0-bit lookahead. Your packet parsers have never been so Zen.

PR #4150, opened 2023-09-11
Hashvec-based maps to save memory

ChrisDodd:
The impetus for this comes from a customer case that is blowing up (memory-use-wise) in the frontend ComputeWriteSets code -- it turns out that the overhead for an ordered_map where both the key and value are just pointers is quite extreme.

This code is old C code that I've repurposed to be the base class of an improve C++ map class -- it keeps the map elements in a simple `vector<pair<KEY, VALUE>>` and has a separate hash table of indexes into the vector for fast lookup.  So the overhead is just a few bytes per entry, compared to the ~48 bytes (6 pointers) per entry used by ordered_map.  This reduces the peak memory use for our case from 12GB to 5.1GB.  Speed is pretty much unchanged.

This could trivially be extended to support more efficient ordered_sets as well (and multimaps and multisets)

### DIFF
--- a/lib/hvec_map.h
+++ b/lib/hvec_map.h
@@ -1,0 +1,342 @@
+/*
+Copyright 2023-present Intel
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+#ifndef LIB_HVEC_MAP_H_
+#define LIB_HVEC_MAP_H_
+
+#include <tuple>
+#include <vector>
+
+#include "exceptions.h"
+#include "hashvec.h"
+
+template <class KEY, class VAL, class HASH = std::hash<KEY>, class PRED = std::equal_to<KEY>,
+          class ALLOC = std::allocator<std::pair<const KEY, VAL>>>
+class hvec_map : hash_vector_base {
+    HASH hf;  // FIXME -- use empty base optimization for these?
+    PRED eql;
+
+ public:
+    typedef KEY key_type;
+    typedef std::pair<const KEY, VAL> value_type;
+    typedef VAL mapped_type;
+    typedef HASH hasher;
+    typedef PRED key_equal;
+    typedef ALLOC allocator_type;
+
+    typedef typename std::vector<value_type>::pointer pointer;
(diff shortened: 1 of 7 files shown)


lib/hvec_map.h
```diff
@@ -0,0 +1,335 @@
...
+    uint32_t limit() override { return data.size(); }
+    void resizedata(size_t sz) override { data.resize(sz); }
+    void moveentry(size_t to, size_t from) override {
+        // can't just assign, as the keys are const -- need to destruct & construct
+        data[to].~value_type();
+        new (&data[to]) value_type(std::move(data[from]));
+    }
+};
+
+// XXX(seth): We use this namespace to hide our get() overloads from ADL. GCC
```
fruffy, line 289: Stale comments?

fruffy (commented):
Couple questions: 
It looks like we are rolling our own implementation here, but we do not have tests etc. Is there maybe an implementation from boost/abseil etc that could serve the same purpose?

Is this also reliably deterministic, like ordered_map?

lib/hvec_map.h
ChrisDodd: Not really stale -- the comment is correct as to why the weird namespace stuff is done (this is just copied from the ordered_map.h code), which Seth originally put in.  There should be a better way of doing this, but I'm not sure what it is.

ChrisDodd:
> Couple questions: It looks like we are rolling our own implementation here, but we do not have tests etc. Is there maybe an implementation from boost/abseil etc that could serve the same purpose?
> 
> Is this also reliably deterministic, like ordered_map?
> 
> Any other places in the compiler where we could use this? What are the tradeoffs?

I can try to translate some of the original C unit tests for this code so we can have some -- this is 40 year old C code, so that was pretty solid, but the translating to C++ might be fragile.  I'm not aware of any decent (space and time efficient) hash table implementations from boost or elsewhere (they tend to trade-off wasting lots of space to maximize speed, or be very slow to minimize space).

This code also has the same ordering guarantees as ordered_map, so it is a drop-in replacement for that as well -- just faster and uses less memory.

fruffy (approved):

fruffy:
Should `ordered_map` even be used at this point? Or are there any advantages to it? 

I am aware of this benchmark of recent hash map implementations, but it does not track memory efficiency. Or rather, just comments on it briefly:

https://martin.ankerl.com/2022/08/27/hashmap-bench-01/

ChrisDodd:
> This finally fixes #151. I don't know why that issue was closed.

This reduces the peak memory use for DefUse by a bit over half, so its a nice improvement

> Should `ordered_map` even be used at this point? Or are there any advantages to it?

ordered_map supports `lower_bound` and `upper_bound` since it is based on trees rather than a hash function.  It also allows insert at a specific position in the order (rather than just at the end), which could be added with hvec_maps, but would be non-trivial and require O(n) time (to move possibly everything in the vector and then rewrite the hash table).  Not clear if we ever use either feature.

Issue #4158, opened 2023-09-14
ebpf-kernel/length_ebpf.p4 test fails to pass eBPF verifier


thomascalvert-xlnx:
With p4c `555029d03`, running Ubuntu 20.04 with kernel 5.4.0-144, I see the above test fail:

```
0: (61) r2 = *(u32 *)(r1 +80)
1: (61) r1 = *(u32 *)(r1 +76)
2: (bc) w3 = w2
3: (1c) w3 -= w1
last_idx 3 first_idx 0
regs=8 stack=0 before 2: (bc) w3 = w2
regs=4 stack=0 before 1: (61) r1 = *(u32 *)(r1 +76)
regs=4 stack=0 before 0: (61) r2 = *(u32 *)(r1 +80)
R3 32-bit pointer arithmetic prohibited
```

Note that kernel tests need to be run by root, and are skipped otherwise. The simplest way to reproduce is, from your `build` directory:
```
sudo ctest --output-on-failure -R 'ebpf-kernel/testdata/p4_16_samples/length_ebpf.p4'
```

The actual C instruction which causes the issue is:
```
u32 ebpf_pkt_len = ebpf_packetEnd - ebpf_packetStart;
```
which I believe is used to support the `packet.length()` call used in the program's parser. I suspect that, since no other test program uses this method, the variable is usually optimised out. (The only other uses are for trace messages which again are generally disabled.)

fruffy:
I believe the eBPF maintain has no dedicated maintainer. It might take some time until this bug is fixed.

mihaibudiu:
The ebpf backend contributors have been usually very responsive.

fruffy:
> The ebpf backend contributors have been usually very responsive.

This bug is in the classic eBPF kernel tests, not sure whether it extends to the PSA eBPF portion. 

@tatry @osinstom

Is this a bug in parts of the code that you depend on?

tatry:
I believe the problem is related to classic eBPF only (filter model). Packet length is calculated here:
https://github.com/p4lang/p4c/blob/27b7171569fd9fa0942ea6a18dfc0727491d0663/backends/ebpf/ebpfProgram.cpp#L245-L246
which generates aforementioned problematic code. This is not used for PSA architecture.

As a side note, PSA architecture has its own method to calculate packet length here:
[...]

PR #4159, opened 2023-09-15
backends/ebpf: Change packet length calc to pass kernel verifier (#4158)

thomascalvert-xlnx:
This basically copies the strategy used in the PSA sub-backend.

Edit: as discovered in a comment below, this PR is only needed for passing the verifier in kernel versions prior to 5.9.

### DIFF
--- a/backends/ebpf/ebpfProgram.cpp
+++ b/backends/ebpf/ebpfProgram.cpp
@@ -242,8 +242,8 @@ void EBPFProgram::emitLocalVariables(CodeBuilder *builder) {
     builder->newline();
 
     builder->emitIndent();
-    builder->appendFormat("u32 %s = %s - %s", lengthVar.c_str(), packetEndVar.c_str(),
-                          packetStartVar.c_str());
+    builder->appendFormat("u32 %s = %s", lengthVar.c_str(),
+                          builder->target->dataLength(model.CPacketName.str()).c_str());
     builder->endOfStatement(true);
 }
 
--- a/backends/ebpf/target.h
+++ b/backends/ebpf/target.h
@@ -96,6 +96,7 @@ class Target {
                           cstring argName) const = 0;
     virtual cstring dataOffset(cstring base) const = 0;
     virtual cstring dataEnd(cstring base) const = 0;
+    virtual cstring dataLength(cstring base) const = 0;
     virtual cstring forwardReturnCode() const = 0;
     virtual cstring dropReturnCode() const = 0;
     virtual cstring abortReturnCode() const = 0;
@@ -183,6 +184,7 @@ class KernelSamplesTarget : public Target {
     cstring dataEnd(cstring base) const override {
         return cstring("((void*)(long)") + base + "->data_end)";
     }
+    cstring dataLength(cstring base) const override { return cstring(base) + "->len"; }
     cstring forwardReturnCode() const override { return "TC_ACT_OK"; }
     cstring dropReturnCode() const override { return "TC_ACT_SHOT"; }
     cstring abortReturnCode() const override { return "TC_ACT_SHOT"; }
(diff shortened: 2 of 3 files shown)


osinstom (approved):

mihaibudiu:
Can you add a test as well? Thank you

thomascalvert-xlnx:
@mihaibudiu could you please elaborate? As mentioned in the linked issue there is already a test, the failure of which prompted this fix. Are you looking for something else?
```
sudo ctest --output-on-failure -R 'ebpf-kernel/testdata/p4_16_samples/length_ebpf.p4'
```

mihaibudiu:
We usually want a test that would fail without this patch and will pass with the patch.
It's just regression testing, to make sure this bug is not reintroduced in later commits.
The fact that nothing has changed under testdata (or no xfail was removed) indicates that there is no such test part of the PR.

jafingerhut:
Perhaps the failing test that Thomas is pointing out that he fixed is not being run in any CI test currently?

mihaibudiu:
Making it be part of the CI would be the right approach

mihaibudiu:
I also think you need to run clang-format on your code

thomascalvert-xlnx:
> Making it be part of the CI would be the right approach

@mihaibudiu @jafingerhut I think the reason this test (and others) are not run as part of CI is because they require `root` access in order to configure the network namespaces and load the eBPF program into the kernel. https://github.com/p4lang/p4c/blob/6e7cc248043469740de79d78a788424303879ed4/backends/ebpf/targets/kernel_target.py#L182-L184  From [this stackoverflow](https://stackoverflow.com/a/65956131) answer it appears that it may be possible to load eBPF programs as a user but I think this would require significant rework of the test harness (and doesn't handle the namespaces part).

> I also think you need to run clang-format on your code

Yes apologies I'm still ramping up with this part of the process. Will fix the formatting issues here & in the other PR.

fruffy:
> > Making it be part of the CI would be the right approach
> 
> @mihaibudiu @jafingerhut I think the reason this test (and others) are not run as part of CI is because they require `root` access in order to configure the network namespaces and load the eBPF program into the kernel.
> 
> https://github.com/p4lang/p4c/blob/6e7cc248043469740de79d78a788424303879ed4/backends/ebpf/targets/kernel_target.py#L182-L184
> From [this stackoverflow](https://stackoverflow.com/a/65956131) answer it appears that it may be possible to load eBPF programs as a user but I think this would require significant rework of the test harness (and doesn't handle the namespaces part).
> 
> > I also think you need to run clang-format on your code
> 
> Yes apologies I'm still ramping up with this part of the process. Will fix the formatting issues here & in the other PR.

Just looked at this. Actually, the reason is that the libbpf dependency is built after we check for its existence... Let me fix this.

thomascalvert-xlnx:
Fixed clang-format issues and rebased onto top of main.

thomascalvert-xlnx:
Thanks @fruffy for providing a fix. Indeed I was wrong and CI tests are already run as root.

@jafingerhut the failures in #4166 seem to be different to this one - the errors are about a relocation without PIE. Surprisingly, the `length_ebpf` test seems to have passed in that regression run. My only guess at this time is that it could be due to different kernel versions, `5.15.0-1046-azure` vs `5.4.0-144-generic`.

After some more digging I found [this patch](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/kernel/bpf/verifier.c?id=6c6935419e2fd8ef1fcde71c4e50b9095e520900), which seems to suggest that the kernel verifier accepts this construct as of kernel 5.9. This is consistent with the observed behaviour.

In summary: this patch only applies to kernels versions prior to 5.9. At a glance, the CI runs all seem to be running more recent kernels (Ubuntu20: `5.15.0-1046-azure`; all others: `6.2.0-1011-azure`)

thomascalvert-xlnx:
This got included in https://github.com/p4lang/p4c/commit/aefe41fae5c381644262ba318df03124615c04be - closing.

PR #4161, opened 2023-09-15
[P4Testgen] Match switch labels directly using string literal expressions

fruffy:
This PR simplifies the way we handle `action_return`. Instead of returning an index and then trying to reconstruction the action from the index and the table, we simply return a string expression. This string expression contains the action name, which we can directly match.


fruffy:
@pkotikal Could you give this a review?

pkotikal (approved):

PR #4164, opened 2023-09-16
Increase the timeout of the switch.p4 p4_14 program

fruffy:
Maybe this can help with the sporadic test failures. I wonder whether we should even still bother with testing p4-14.


jafingerhut:
I do not have any strong opinion on whether to continue testing this program, but it does have the distinction of being one of the largest P4 programs in the test suite.

jafingerhut (approved):
LGTM

PR #4167, opened 2023-09-18
[clang-format] Enforce on the entire codebase.

thomascalvert-xlnx:
This is simply the result of running 'make clang-format-fix-errors' on the top of the main branch.


fruffy:
Which version of clang-format are you using? 

We usually use `pip3 install clang-format==15.0.6` but we can upgrade with this PR. 

https://github.com/p4lang/p4c/blob/main/requirements.txt#L7C1-L7C1

thomascalvert-xlnx:
Oh sorry for the noise -- it appears that the distro-installed one (v10) was being picked up instead of the pip installed one. After removing the former thing are in order.

PR #4168, opened 2023-09-18
Simplify the DPDK PTF build and test scripts. 

fruffy:
- Refactor the `build-ci.sh` into optional segments (BMv2, eBPF, DPDK, etc) and support the installation of some DPDK dependencies. Remove the `dpdk-ci-build.sh` in favour of the refactored scripts. 
- Make the DPDK PTF test and installation scripts less dependent on environment variables. Install all DPDK dependencies in one shared directory. 
- Clean up the DPDK PTF Python test script and remove unneeded configuration options. 

@Hoooao


fruffy:
@jafingerhut These changes simplify the way we can PTF DPDK tests. The goal is to remove as much friction as possible. The first thing this gets rid of is the use of environment variables. The installation is also simplified. Could you give this PR a pass? #4173 depends on it.

tools/ci-build.sh
```diff
@@ -63,120 +76,144 @@ P4C_DEPS="bison \
...
+  if [[ "${INSTALL_PTF_EBPF_DEPENDENCIES}" == "ON" ]] ; then
+    install_ptf_ebpf_test_deps
+  fi
 fi
+# ! ------  END EBPF -----------------------------------------------
+
+# ! ------  BEGIN DPDK -----------------------------------------------
+function build_dpdk() {
+  # Replace existing Protobuf to match the one specified in the runtime shell.
+  # TODO: Do we really need the shell?
```
jafingerhut, line 201: p4runtime-shell package is needed if you want to run this PTF test without modifying it: https://github.com/p4lang/p4c/blob/main/testdata/p4_16_samples/pna-dpdk-add_on_miss0.py#L27-L29

It could be modified so that it does not use p4runtime-shell, of course, but not clear to me that is worth any effort to make such a change.

jafingerhut (approved):
I have looked at all of the changes, but I cannot claim to have weighed every line of change as carefully as someone responsible for maintaining it :-).  It passes all tests.  I verified that the DPDK test does run the one currently-included automated DPDK PTF test, and it passes, so that is good.  These changes do appear to simplify the build and test scripts, so that is good.

tools/ci-build.sh
fruffy: I believe #4173 will fix this.

Issue #4156, opened 2023-09-13
p4c-dpdk compiling 2023-Sep-13 version of DASH P4 code produces spec file that fails to load


jafingerhut:
Note: This issue was discovered using the latest version of p4c-dpdk as of 2023-Sep-13 (commit SHA mentioned below).  It is true that this version of p4c-dpdk DOES FIX these earlier issues:
+ https://github.com/p4lang/p4c/issues/3965
+ https://github.com/p4lang/p4c/issues/3966

However, there is a newer version of DASH P4 code since 2023-Apr when those issues were filed, and when testing with the latest version, p4c-dpdk produces a .spec file that fails to load into the P4-DPDK software switch.  Hence the creation of this issue to track this.

Version of p4c source used to build p4c-dpdk:
```
$ git log -n 1 | cat
commit 5b76a26e4afbeb041d4eb47faa32e589521a4a50
Author: Usha Gupta <usha.gupta@intel.com>
Date:   Wed Sep 13 01:39:56 2023 +0530

    Fix whole header move instructions for DPDK (#4153)
```

I compiled p4c-dpdk binary from that source code.

Version of DASH P4 source code tested with:

```
$ git clone https://github.com/sonic-net/DASH
$ cd DASH
$ git log -n 1 | head -n 3
commit 3bbe99f19cb047770e43d63204ff1e94ffadca27
Author: Kamil Cudnik <kcudnik@gmail.com>
Date:   Thu Sep 7 19:34:04 2023 +0200
```

Compile the DASH P4 code, without changes from the version at the commit SHA shown above, for DPDK:

```
$ cd DASH/dash-pipeline/bmv2
$ mkdir -p pipe
$ p4c-dpdk -DTARGET_DPDK_PNA -DPNA_CONNTRACK \
    --arch pna \
    --p4runtime-files p4Info.txt \
    --bf-rt-schema bf-rt.json \
    --context pipe/context.json \
    -o pipe/dash_pipeline.spec \
    dash_pipeline.p4
```
This produced the following output files, included in the attached zip file:

+ p4Info.txt
+ bf-rt.json
+ pipe/context.json
+ pipe/dash_pipeline.spec

In earlier issues, the root cause was something that P4-DPDK did not expect to find in the .spec file and/or the context.json file.  If that is the case here, I do not yet know what portion of those files might be the root cause of the loading failure.  Hopefully someone can add a comment to this issue with more details when they find out.
[...]

PR #4171, opened 2023-09-20
Dpdk backend: Remove illegal instruction generation, reject operations on >64bit operands

usha1830:
Some issues were found while loading the compiler generated spec file for DASH pipeline. mentioned in this https://github.com/p4lang/p4c/issues/4156
This PR removes emission of all instructions which are not available in the target and restricts all operations (arithmetic, logical etc) on >64-bits operands.


backends/dpdk/dpdk.def
```diff
@@ -304,27 +304,6 @@ class DpdkMovStatement : DpdkUnaryStatement {
 #noconstructor
 }
 
-class DpdkNegStatement: DpdkUnaryStatement {
```
mihaibudiu, line 307: how come all these are no longer necessary?

mihaibudiu (approved):

backends/dpdk/dpdk.def
usha1830: These were added in the initial commit for p4c-dpdk. I presume these were just placeholder instructions since the development of p4-dpdk  software pipeline and compiler was going in parallel. The functionality of all these (now removed) operations are achieved using other existing instructions in the p4-dpdk pipeline.

PR #4072, opened 2023-07-18
Add a PTF test CI pipeline for p4c-dpdk on the DPDK SoftNIC

Hoooao:
# Summary

This PR sets up a PTF test CI pipeline for p4c-dpdk and the dpdk-target. 

- [PTF(Packet Testing Framework)](https://github.com/p4lang/ptf) is a Python based dataplane test framework.
- [p4c-dpdk](https://github.com/p4lang/p4c/blob/main/backends/dpdk) is the p4 dpdk compiler backend that translates the P4-16 programs to DPDK API to configure the DPDK software switch (SWX) pipeline.
- [DPDK-target](https://github.com/p4lang/p4-dpdk-target/tree/main) implements the P4 pipeline,driven by artifacts generated by the P4 compiler.

# Background

A PTF test CI pipeline for p4c-dpdk on the DPDK SWX is currently missing. Adding such tests is beneficial for both the development of the compiler backend and DPDK SWX pipeline.

# Demo

- Make sure `infrap4d` and `p4sde/dpdk-target` are properly [setup](https://github.com/ipdk-io/networking-recipe/blob/main/docs/guides/dpdk-guide.md). 
- Define `$P4C_DIR` as the root directory for p4c. 
- Run `sudo $P4C_DIR/backends/dpdk/run-dpdk-ptf-test.py $P4C_DIR $P4C_DIR/testdata/p4_16_samples/pna-dpdk-add_on_miss0.p4 --ipdk-recipe $IPDK_RECIPE --sde-install $SDE_INSTALL --ld-library-path $LD_LIBRARY_PATH --testfile $P4C_DIR/testdata/p4_16_samples/pna-dpdk-add_on_miss0.py -ll DEBUG`

Output:
```
Updated Environment Variables ...
// Skip

1024
1024
INFO: ---------------------- Build conf file ----------------------
[...]

PR #4173, opened 2023-09-23
[P4Testgen] Add a DPDK-PTF P4Testgen back end and the corresponding test runner

Hoooao:
Improved based on #4072 to use Testgen to generate PTF tests for pna/dpdk p4 programs.


.github/workflows/PTF-TEST.yml
```diff
@@ -0,0 +1,100 @@
+name: "PTF-backend"
+# This is just a temporary workflow to test the PTF backend
+# Will be merged to ci-dpdk-ptf-tests.yml once it is ready
+on:
+  push:
+    branches:
+      - ptf-test-pr
```
fruffy, line 7: This test needs to be updated. I would recommend replacing the existing Github action workflow.

testdata/p4_16_samples/pna-dpdk-small_sample.p4
```diff
@@ -0,0 +1,180 @@
+/*
```
fruffy, line 1: This example does not have a corresponding reference file. They need to be generated.

.github/workflows/PTF-TEST.yml
Hoooao: yes, it will replace the existing dpdk ptf ci.

jafingerhut:
@Hoooao @fruffy Question about the goal of this PR: When I compile the latest version of p4testgen, install it, and then try running a command like this:
```
$ cd p4c/testdata/p4_16_samples
$ p4testgen --target dpdk --arch pna --max-tests 1000 --out-dir tmp --test-backend ptf pna-dpdk-header-stack-assignment.p4 
In file: /home/andy/forks/p4c/backends/p4tools/modules/testgen/targets/pna/test_backend.cpp:49
Unimplemented compiler support: Test back end PTF not implemented for this target. Supported back ends are [METADATA].
```
Is the purpose of this PR to somehow use p4testgen to auto-generate PTF tests and execute them?  What p4testgen command line do you use to accomplish the generation of the PTF tests?

fruffy:
> @Hoooao @fruffy Question about the goal of this PR: When I compile the latest version of p4testgen, install it, and then try running a command like this:
> 
> ```
> $ cd p4c/testdata/p4_16_samples
> $ p4testgen --target dpdk --arch pna --max-tests 1000 --out-dir tmp --test-backend ptf pna-dpdk-header-stack-assignment.p4 
> In file: /home/andy/forks/p4c/backends/p4tools/modules/testgen/targets/pna/test_backend.cpp:49
> Unimplemented compiler support: Test back end PTF not implemented for this target. Supported back ends are [METADATA].
> ```
> 
> Is the purpose of this PR to somehow use p4testgen to auto-generate PTF tests and execute them? What p4testgen command line do you use to accomplish the generation of the PTF tests?

Yes, the goal is be able to generate PTF tests for the DPDK PNA architecture using the command you described.

jafingerhut:
@Hoooao It is obviously up to you when you want to consider this feedback relative to other changes you are making on this PR, but note that there are currently unresolved conflicts in a couple of files between your changes, and some files that have had commits merged in that change them.  You can see the "This branch has conflicts that must be resolved" messages on the Github PR page for details.

Hoooao:
> @Hoooao It is obviously up to you when you want to consider this feedback relative to other changes you are making on this PR, but note that there are currently unresolved conflicts in a couple of files between your changes, and some files that have had commits merged in that change them. You can see the "This branch has conflicts that must be resolved" messages on the Github PR page for details.

Thanks for the heads up

jafingerhut:
@Hoooao Do you consider this read for review now, and then merging if reviews are OK?  Or you expect to make more commits before review?

jafingerhut:
FYI, in case it might be related to the reason you are seeing P4-DPDK fail to start, if you have a number of ports that is not a power of 2, it fails when you try to load a P4 program into it (not when it starts up, but when you try to do the SetForwardingPipelineConfig to load the P4 program).

If you are having issues where infrap4d fails completely to start up at all, not even to the point that it should be waiting for a SetForwardingPipelineConfig request, then I do not have any guesses why that may be.

fruffy:
> If you are having issues where infrap4d fails completely to start up at all, not even to the point that it should be waiting for a SetForwardingPipelineConfig request, then I do not have any guesses why that may be.

It appears to be crashing when trying to launch `bf_switchd`. Unclear why unfortunately.

jafingerhut:
My belief is that there should be no reason to run both infrap4d and bf_switchd.

bf_switchd gives TDI access to table add/modify/delete operations in the P4-DPDK data plane.

infrap4d contains a P4Runtime API server based on the Stratum code that listens for incoming TCP connection requests from P4Runtime API clients, and translates request messages into the appropriate TDI add/modify/delete operations.

Running both at the same time would likely only lead to problems, would be my guess.

jafingerhut:
> `infrap4d` launches `bf_switchd` [here](https://github.com/ipdk-io/stratum-dev/blob/339022ac39df6678261ed13ecb4c35528087d1ed/stratum/hal/lib/tdi/dpdk/dpdk_sde_target.cc#L82). Would love to avoid calling it but that seems to be baked in.

I was unaware of that.  Thanks for the info.

My statement is still correct, I think, from the perspective that there is no reason if you are manually running commands in sequence, to start both infrap4d and bf_switchd _from the command line_.  You pick one.

fruffy:
> My statement is still correct, I think, from the perspective that there is no reason if you are manually running commands in sequence, to start both infrap4d and bf_switchd _from the command line_. You pick one.

In our case we are calling `infrap4d`, which in turns calls into the `bf_switchd` library, which then crashes. We can only reproduce this in CI, but it is unfortunately a show stopper to develop a functional testing pipeline for the PTF DPDK target.

fruffy:
@jafingerhut In this run https://github.com/p4lang/p4c/actions/runs/6501241470/job/17658197303?pr=4173 infrap4d fails to start (and appears to crash). This seemingly only happens in CI. Do you know anyone we could poke about this?

jafingerhut:
@bharticemk @5abeel @satish153 @swaroopsarma Would one of you know why @fruffy (Fabian) might be experiencing the problems described in this comment: https://github.com/p4lang/p4c/pull/4173#issuecomment-1761700086 ?

I have run infrap4d within an IPDK Networking docker container before, but I have not yet attempted to do so outside of a Docker container, which I believe Fabian is trying to do here.  I do not have a good summary of exactly how he has built and runs these tests, but I believe all of the details are available in this PR and the code it is based upon.

backends/dpdk/base_test.py
```diff
@@ -0,0 +1,1142 @@
+# This file was copied and then modified from the file
```
fruffy, line 1: There is a lot of redundancy here. We can likely make this a shared library for both BMv2 and DPDK.

backends/dpdk/run-dpdk-ptf-test.py
```diff
@@ -233,19 +248,21 @@ def build_and_load_pipeline(
...
-            f"{self.options.ipdk_install_dir}/bin/p4rt-ctl "
-            "set-pipe br0 "
-            f"{conf_bin} "
-            f"{info_name} "
-        )
-        returncode = self.bridge.ns_exec(command, timeout=30)
-        if returncode != testutils.SUCCESS:
-            testutils.log.error("Failed to load pipeline")
-            return returncode
+        # NOTE: in generated PTF tests, the pipelines are loaded in individual test cases.
```
fruffy, line 251: This comment should be removed. Fewer moving parts is good.

jafingerhut:
> It also adds a PTF DPDK testing library which does not rely on p4runtime shell and instead uses Protobuf messages directly. @Hoooao implemented all of this.

Is there any difference between the Python code in this PR for creating/parsing Protobuf messages, and that used for any target, e.g. BMv2, other than DPDK?  It seems beneficial to have as few different Python libraries for this as we can get down to.

fruffy:
> > It also adds a PTF DPDK testing library which does not rely on p4runtime shell and instead uses Protobuf messages directly. @Hoooao implemented all of this.
> 
> Is there any difference between the Python code in this PR for creating/parsing Protobuf messages, and that used for any target, e.g. BMv2, other than DPDK? It seems beneficial to have as few different Python libraries for this as we can get down to.

There are subtle differences in the two libraries, largely revolving around election id and device id. We could fix this in a separate PR and add a common P4Runtime communication library?

fruffy:
> > It also adds a PTF DPDK testing library which does not rely on p4runtime shell and instead uses Protobuf messages directly. @Hoooao implemented all of this.
> 
> Is there any difference between the Python code in this PR for creating/parsing Protobuf messages, and that used for any target, e.g. BMv2, other than DPDK? It seems beneficial to have as few different Python libraries for this as we can get down to.

The changes were easy enough to do them here. We now have one single P4Runtime Python library that we can use for both BMv2 and DPDK. Will do some more cleanup for this PR. We will also have to fix the deprecated [add-on-miss](https://github.com/p4lang/p4c/blob/main/testdata/p4_16_samples/pna-dpdk-add_on_miss0.py) test.

backends/dpdk/run-dpdk-ptf-test.py
```diff
@@ -72,15 +72,17 @@
     help="The log level to choose.",
 )
 
-# 9559 is the default P4Runtime API server port
-GRPC_PORT: int = 9559
+# 28000 is the default P4Runtime API server port
+GRPC_PORT: int = 28000
```
jafingerhut, line 76: I have no personal issue if you want to use a non-default port number for the P4Runtime API connection, but this change to the comment is definitely incorrect.  9559 is the default port, as mentioned in the P4Runtime API spec.

jafingerhut (approved):
Except for my latest minor comment about a comment that was changed to now be factually incorrect, this looks good to me.

fruffy:
> Except for my latest minor comment about a comment that was changed to now be factually incorrect, this looks good to me.

Fixed. I believe 28000 was the old port.  I have created an issue (#4204) for the remaining TODOs. This will be done in a new PR.

PR #4179, opened 2023-09-29
Get host architecture dynamically while downloading `protoc` binary

Skazza94:
In `cmake/Protobuf.cmake`, when `P4C_USE_PREINSTALLED_PROTOBUF` is enabled (which is the default case), the `protoc` binary is directly fetched from GitHub releases. However, there was an issue with the hardcoded assumption that the downloaded `.zip` archive is for the `x86_64` architecture, which resulted in compilation errors on `arm64` systems such as Apple M1/M2 silicons or when building multi-platform Docker containers.

This pull request addresses the problem by dynamically selecting the appropriate `.zip` archive based on the host architecture. It achieves this by utilizing the `CMAKE_HOST_SYSTEM_PROCESSOR` variable in CMake, which leverages `uname -r` to identify the target architecture.

Currently, this selection logic has been implemented for the `arm64` architecture, but it can be easily extended to support other architectures as needed.

Furthermore, in order to accommodate this dynamic selection, the `URL_HASH` directive in the `fetchcontent_declare` has been removed since it was not feasible to compute the SHA1 hash of the file dynamically.

### DIFF
--- a/cmake/Protobuf.cmake
+++ b/cmake/Protobuf.cmake
@@ -44,10 +44,13 @@ macro(p4c_obtain_protobuf)
       USES_TERMINAL_DOWNLOAD TRUE
       GIT_PROGRESS TRUE
     )
+
     # Derive the target architecture in order to download the right zip.
     set(protobuf_ARCH "x86_64")
-    if(CMAKE_HOST_SYSTEM_PROCESSOR STREQUAL "aarch64")
+    if(CMAKE_HOST_SYSTEM_PROCESSOR STREQUAL "aarch64" OR CMAKE_HOST_SYSTEM_PROCESSOR STREQUAL "arm64")
       set(protobuf_ARCH "aarch_64")
+    else()
+      MESSAGE(FATAL_ERROR "Unsupported host architecture `${CMAKE_HOST_SYSTEM_PROCESSOR}`")
     endif()
 
     # Pull a different protoc binary for MacOS.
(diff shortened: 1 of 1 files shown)


cmake/Protobuf.cmake
```diff
@@ -44,20 +44,25 @@ macro(p4c_obtain_protobuf)
       USES_TERMINAL_DOWNLOAD TRUE
       GIT_PROGRESS TRUE
     )
+    # Derive the target architecture in order to download the right zip.
+    set(protobuf_ARCH "x86_64")
```
fruffy, line 48: Is it possible to set the value to just `CMAKE_HOST_SYSTEM_PROCESSOR` and throw an error if we encounter an unsupported arch? We should be able to support `x86_64` and `aarch_64` and `arm64`.

fruffy (approved):

cmake/Protobuf.cmake
Skazza94: Hi @fruffy, the problem is that the protoc binary in [Protobuf releases](https://github.com/protocolbuffers/protobuf/releases/tag/v21.11) is called `aarch_64`, that is why I have to remap the `arm64` value coming from `uname -r` to that one.

However, I agree in adding an error when the architecture is not supported.

Skazza94: Hi @fruffy,
I added an additional check on the `arm64` value and also the error handling.

Let me know if you need additional changes. 

Thanks.

fruffy (approved):

PR #4182, opened 2023-10-02
Add P4Smith, a random program generator to the P4Tools framework

fruffy:
This is a port of [Bludgeon](https://github.com/p4gauntlet/bludgeon), which has initially been written for Gauntlet. 
Bludgeon is a P4 random program generator in the spirit of CSmith. It can generate random programs for p4test, p4c-bm2-ss, p4c-bm2-psa, and for the Tofino compiler. 

Porting Bludgeon to the P4Tools framework will make it more extensible and improve its support for P4 targets. The new name of this tool is now P4Smith. This version is simple but can already generate random programs for:
- v1model1 and `psa` on BMv2. 
- `tna` on Tofino 1
- `pna` on DPDK
- Generic programs for `p4test`

There is currently a lot of code duplication etc happening. That is an artifact of the port. Also this is still a port of graduate student code, so there is a lot of constructs that could be improved. I am hoping that this initial version is enough to build a common platform for a P4 fuzzer.


vlstill:
Hi! This looks like quite an interesting idea. Would this allow easily adding target platforms similarly to P4 testgen? What is currently missing for this to be non-WIP?

fruffy:
> Hi! This looks like quite an interesting idea. Would this allow easily adding target platforms similarly to P4 testgen? What is currently missing for this to be non-WIP?

There is not much work left, you can already generate random programs. I just need to finish the target-specific instantiation to make it possible to produce custom code per target.

fruffy:
Depends on #4650.

fruffy:
@asl @ChrisDodd We might be able to use this to fuzz #4562.

asl:
@fruffy Does it support loop?

fruffy:
> @fruffy Does it support loop?

Not yet, this random program generator uses the P4C IR to generate elements. But it could be extended to support loops.

fruffy:
@asl @vlstill @jafingerhut Requesting a review on this contribution to the P4Tools framework. 

This is getting increasingly hard for me to maintain with the compiler changes. I think it makes sense to get the initial version of this fuzzer in, then we can iteratively improve it. There are still many things that should be done here, but a rewrite in bulk is currently impractical for me.

jafingerhut (approved):
I have only reviewed the two README.md files, and they look good.  As far as the C++ code, I am happy to try to build it when this is merged in, and try it out a little bit to see whether it generates some random programs for me.

PR #4185, opened 2023-10-09
backends/ebpf: Output error if a header is not byte-aligned (#4176).

### LINKED ISSUE #4176: Should we support headers which aren't byte-aligned?
This question arose whilst working on the eBPF backend, however it applies more generally to the rest of the compiler.

Consider a `header` which has a size in bits which isn't a multiple of 8. I have never encountered any protocol which would define something like this, but it seems that the P4 language permits it - please correct me if I'm wrong.
```
header foo {
  bit<3> bar;
}
```

Currently I believe that the eBPF backend won't deal with this case properly: although it tracks the packet offset in bits there is an assumption that headers start on a byte boundary:

https://github.com/p4lang/p4c/blob/aefe41fae5c381644262ba318df03124615c04be/backends/ebpf/ebpfParser.cpp#L462-L475

No errors or warnings will be generated, but the emitted code won't work properly. I could not find any tests which define non-byte-sized headers either. I have not checked other backends but wouldn't be surprised if there exists similar conditions.

The compiler should either handle such headers correctly, or print an error message when they are encountered. I am in favour of the latter approach since there are no real-world scenarios for such headers, and doing so will allow for future performance optimisations.

thomascalvert-xlnx:
Also adds support for testing error messages in the eBPF backend (separate commit).

Previously, declaring such a header would not have caused any compile-time errors but generated functionally incorrect code. With this patch, the user is made aware and compilation fails.

Fixes #4176

### DIFF
--- a/backends/ebpf/CMakeLists.txt
+++ b/backends/ebpf/CMakeLists.txt
@@ -182,6 +182,9 @@ set (EBPF_TEST_SUITES
 set (EBPF_KERNEL_TEST_SUITES
   "${P4C_SOURCE_DIR}/testdata/p4_16_samples/*_ebpf-kernel.p4"
   )
+set (EBPF_ERRORS_SUITES
+  "${P4C_SOURCE_DIR}/testdata/p4_16_ebpf_errors/*.p4"
+  )
 
 # determine the kernel version
 execute_process(COMMAND uname -r
@@ -247,6 +250,7 @@ endif()
 # Ideally, this is done via check for the python package
 p4c_add_tests("ebpf-bcc" ${EBPF_DRIVER_BCC} ${EBPF_TEST_SUITES} "${XFAIL_TESTS_BCC}")
 p4c_add_tests("ebpf" ${EBPF_DRIVER_TEST} ${EBPF_TEST_SUITES} "${XFAIL_TESTS_TEST}")
+p4c_add_tests("ebpf-errors" ${EBPF_DRIVER_TEST} ${EBPF_ERRORS_SUITES} "${XFAIL_TESTS_TEST}")
 
 # These are special tests with args that are not included in the default ebpf tests
 p4c_add_test_with_args("ebpf" ${EBPF_DRIVER_TEST} FALSE "testdata/p4_16_samples/ebpf_checksum_extern.p4" "testdata/p4_16_samples/ebpf_checksum_extern.p4" "--extern-file ${P4C_SOURCE_DIR}/testdata/extern_modules/extern-checksum-ebpf.c" "")
--- a/backends/ebpf/ebpfParser.cpp
+++ b/backends/ebpf/ebpfParser.cpp
@@ -407,7 +407,14 @@ void StateTranslationVisitor::compileExtract(const IR::Expression *destination) 
         return;
     }
 
+    // We expect all headers to start on a byte boundary.
     unsigned width = ht->width_bits();
+    if ((width % 8) != 0) {
+        ::error(ErrorType::ERR_UNSUPPORTED_ON_TARGET,
+                "Header %1% size %2% is not a multiple of 8 bits.", destination, width);
+        return;
+    }
+
     auto program = state->parser->program;
 
     cstring offsetStr =
--- a/backends/ebpf/run-ebpf-test.py
+++ b/backends/ebpf/run-ebpf-test.py
@@ -178,17 +178,38 @@ def run_test(options, argv):
     if options.extern:
         argv.append("--emit-externs")
     # Compile the p4 file to the specified target
-    result, expected_error = ebpf.compile_p4(argv)
+    result, expected_error, out = ebpf.compile_p4(argv)
 
     # Compile and run the generated output
-    if result == testutils.SUCCESS and not expected_error:
-        # If no empty.stf present, don't try to run the model at all
-        if not os.path.isfile(testfile):
-            msg = "No stf file present!"
-            testutils.log.info(msg)
-            result = testutils.SUCCESS
-        else:
-            result = run_model(ebpf, testfile)
+    if result == testutils.SUCCESS:
+        if not expected_error:
+            # If no empty.stf present, don't try to run the model at all
+            if not os.path.isfile(testfile):
+                msg = "No stf file present!"
+                testutils.log.info(msg)
+                result = testutils.SUCCESS
+            else:
+                result = run_model(ebpf, testfile)
+        else:  # Expected error, check output matches
+            exp_file = options.p4filename.replace("_errors/", "_errors_outputs/", 1) + "-stderr"
+            with open(exp_file, 'r') as f:
+                exp_stderr = f.read().strip()
+            # Output will have full path which differs
+            out = out.replace(options.p4filename, basename).strip()
+            # Output will have one extra line with make error, trim that
+            out = out[: len(exp_stderr)]
+            if exp_stderr != out:
+                testutils.log.error("Error output mismatch!")
+                testutils.log.error(f"Expected output({len(exp_stderr)}):\n{exp_stderr}")
+                testutils.log.error(f"Actual output({len(out)}):\n{out}")
+                for i in range(0, min(len(exp_stderr), len(out))):
+                    if exp_stderr[i] != out[i]:
+                        testutils.log.error(
+                            f"First mismatch at index {i}: {exp_stderr[i]} vs {out[i]}"
+                        )
+                        break
+                testutils.log.error("Failing test.")
+                result = testutils.FAILURE
     # Only if we did not expect it to fail
     if result != testutils.SUCCESS:
         return result
--- a/backends/ebpf/targets/target.py
+++ b/backends/ebpf/targets/target.py
@@ -103,7 +103,7 @@ class EBPFTarget:
         if returncode != testutils.SUCCESS:
             testutils.log.error("Failed to compile the P4 program.")
             # If the compiler crashed fail the test
-            if "Compiler Bug" in out:
+            if out is not None and "Compiler Bug" in out:
                 sys.exit(testutils.FAILURE)
 
         # Check if we expect the p4 compilation of the p4 file to fail
@@ -114,7 +114,7 @@ class EBPFTarget:
                 returncode = testutils.FAILURE
             else:
                 returncode = testutils.SUCCESS
-        return returncode, expected_error
+        return returncode, expected_error, out
 
     def _write_pcap_files(self, iface_pkts_map):
         """Writes the collected packets to their respective interfaces.
--- a/testdata/p4_16_ebpf_errors/header_unaligned.p4
+++ b/testdata/p4_16_ebpf_errors/header_unaligned.p4
@@ -1,0 +1,28 @@
+#include <ebpf_model.p4>
+#include <core.p4>
+
+header ThreeBit_h {
+    bit<3> x;
+}
+
+struct Headers_t {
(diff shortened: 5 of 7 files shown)


fruffy (commented):
There is an empty file in the errors directory. Is that intentional?

fruffy (approved):

fruffy:
Testgen fails because the newly added program does not compile. Could you add it to the XFails [here](https://github.com/p4lang/p4c/blob/main/backends/p4tools/modules/testgen/targets/ebpf/test/EBPFXfail.cmake)?

thomascalvert-xlnx:
@fruffy thanks for the explanation. I'm happy to add the test to xfails. Should we consider excluding the entire `p4_16_ebpf_errors` directory though? Any program in this directory is expected to fail compilation.

fruffy:
> @fruffy thanks for the explanation. I'm happy to add the test to xfails. Should we consider excluding the entire `p4_16_ebpf_errors` directory though? Any program in this directory is expected to fail compilation.

Sure, if it is not to difficult to implement in CMake for you? Otherwise, just adding to Xfails is fine.

thomascalvert-xlnx:
I have reproduced the CI failure in my local dev environment. The solution is not to add the new test to Xfails, because it's not included: the p4tools testgen CMake script already doesn't look for tests in the `_errors` directories:
https://github.com/p4lang/p4c/blob/main/backends/p4tools/modules/testgen/targets/ebpf/test/P4Tests.cmake#L10-L12

The program which is failing is `op_bin.p4` with the error message `fields wider than 64 bits are not supported yet`, which is exactly what we expect, but the test framework says "required regular expression not found!
https://github.com/p4lang/p4c/blob/main/backends/p4tools/modules/testgen/targets/ebpf/test/EBPFXfail.cmake#L23-L27

The issue is the addition of `capture_output=True` argument to `testutils.exec_process()` call in `backends/ebpf/targets/target.py` which is included in this patch - from memory this was required for the stderr output to be captured correctly for checking. I will try and find a compromise...

fruffy:
I think the most simple solution might be to print the output in case of a failure. I am not sure why it is swallowed but `out` should contain all the information we need.

thomascalvert-xlnx:
@fruffy Ok with the latest fixup I can see both tests passing locally. Had to add a new cmdline argument - it's not super elegant but it works. Hopefully the CI machinery will concur :)

@jafingerhut yes that matches my understanding, and is what the first commit in this PR implements for the eBPF backend. But the p4tools/testgen tests seem to perform this kind of checking in CMake (rather than in the python runner script), hence the need for an alternative. (NB: this is mostly educated guessing from my part from skimming the source code)

fruffy:
I am not sure about this solution. It shouldn't be necessary to capture the error since the default behavior is to emit the output. Let me take a look.

fruffy:
@thomascalvert-xlnx 
Could you try this patch and see whether it works? It undoes the last commit and fixes the LogPipe in the testutils library to also record whatever the pipe logs. This way we can easily return output. 
```
diff --git a/backends/ebpf/run-ebpf-test.py b/backends/ebpf/run-ebpf-test.py
index 1bf404bfb..af5ed308f 100755
--- a/backends/ebpf/run-ebpf-test.py
+++ b/backends/ebpf/run-ebpf-test.py
@@ -87,12 +87,6 @@ PARSER.add_argument(
     choices=["CRITICAL", "ERROR", "WARNING", "INFO", "DEBUG", "NOTSET"],
     help="The log level to choose.",
 )
-PARSER.add_argument(
-    "-o",
-    "--no-capture-output",
-    action="store_true",
-    help="Don't capture output (i.e. emit stdout/stderr).",
-)
 
 
 def import_from(module, name):
@@ -133,8 +127,6 @@ class Options:
         # The location of the eBPF runtime, some targets may overwrite this.
         self.runtimedir = str(FILE_DIR.joinpath("runtime"))
         self.extern = ""  # Path to C file with extern definition.
-        # Capture stderr output (used by testgen)
-        self.capture_output = True
 
 
 def run_model(ebpf, testfile):
@@ -186,8 +178,7 @@ def run_test(options, argv):
     if options.extern:
         argv.append("--emit-externs")
     # Compile the p4 file to the specified target
-    result, expected_error, out = ebpf.compile_p4(argv, capture_output=options.capture_output)
-
+    result, expected_error, out = ebpf.compile_p4(argv)
     # Compile and run the generated output
     if result == testutils.SUCCESS:
         if not expected_error:
@@ -242,7 +233,6 @@ if __name__ == "__main__":
     options.target = args.target
     options.extern = args.extern
     options.testdir = tempfile.mkdtemp(dir=os.path.abspath("./"))
-    options.capture_output = not args.no_capture_output
     os.chmod(options.testdir, 0o755)
     # Configure logging.
     logging.basicConfig(
diff --git a/backends/ebpf/targets/target.py b/backends/ebpf/targets/target.py
index 9a0a65a64..08d87d373 100644
--- a/backends/ebpf/targets/target.py
+++ b/backends/ebpf/targets/target.py
@@ -79,7 +79,7 @@ class EBPFTarget:
         """Extracts the interface name out of a pcap filename"""
         return int(os.path.basename(f).rstrip(PCAP_SUFFIX).lstrip(PCAP_PREFIX).rsplit("_", 1)[0])
 
-    def compile_p4(self, argv, capture_output=False):
+    def compile_p4(self, argv):
         # To override
         """Compile the p4 target"""
         if not os.path.isfile(self.options.p4filename):
@@ -99,7 +99,7 @@ class EBPFTarget:
         if p4_args:
             # Remaining arguments
             args += f' P4ARGS="{p4_args}" '
-        out, returncode = testutils.exec_process(args, capture_output=capture_output)
+        out, returncode = testutils.exec_process(args)
         if returncode != testutils.SUCCESS:
             testutils.log.error("Failed to compile the P4 program.")
             # If the compiler crashed fail the test
diff --git a/backends/p4tools/modules/testgen/targets/ebpf/test/TestTemplate.cmake b/backends/p4tools/modules/testgen/targets/ebpf/test/TestTemplate.cmake
index 4c9e058b5..4b5749843 100644
--- a/backends/p4tools/modules/testgen/targets/ebpf/test/TestTemplate.cmake
+++ b/backends/p4tools/modules/testgen/targets/ebpf/test/TestTemplate.cmake
@@ -13,7 +13,7 @@ macro(check_with_kernel testfile testfolder p4test runner_args)
   file(APPEND ${testfile} "for item in \${stffiles[@]}\n")
   file(APPEND ${testfile} "do\n")
   file(APPEND ${testfile} "\techo \"Found \${item}\"\n")
-  file(APPEND ${__testfile} "\tpython3 ${__ebpfrunner} ${runner_args} -o -t kernel -c ${__p4cebpfpath} -tf \${item} ${P4C_SOURCE_DIR} ${p4test}\n")
+  file(APPEND ${__testfile} "\tpython3 ${__ebpfrunner} ${runner_args} -t kernel -c ${__p4cebpfpath} -tf \${item} ${P4C_SOURCE_DIR} ${p4test}\n")
   file(APPEND ${testfile} "done\n")
 endmacro(check_with_kernel)
 
diff --git a/tools/testutils.py b/tools/testutils.py
index 19b7f5b38..50576d92a 100644
--- a/tools/testutils.py
+++ b/tools/testutils.py
@@ -49,10 +49,12 @@ class LogPipe(threading.Thread):
         and start the thread
         """
         threading.Thread.__init__(self)
-        self.daemon = False
-        self.level = level
+        self.daemon: bool = False
+        self.level:int = level
         self.fd_read, self.fd_write = os.pipe()
         self.pipe_reader = os.fdopen(self.fd_read)
+        # We capture what we log to this string.
+        self.out: str = ""
         self.start()
 
     def fileno(self) -> int:
@@ -60,10 +62,10 @@ class LogPipe(threading.Thread):
         return self.fd_write
 
     def run(self) -> None:
-        """Run the thread, logging everything."""
+        """Run the thread, logging and record everything."""
         for line in iter(self.pipe_reader.readline, ""):
             log.log(self.level, line.strip("\n"))
-
+            self.out += line
         self.pipe_reader.close()
 
     def close(self) -> None:
@@ -216,6 +218,8 @@ def exec_process(args: str, **extra_args) -> ProcessResult:
         args = list(filter(None, args))
 
     # Set up log pipes for both stdout and stderr.
+    outpipe: Optional[LogPipe] = None
+    errpipe: Optional[LogPipe] = None
     if "capture_output" not in extra_args:
         if "stdout" not in extra_args:
             outpipe = LogPipe(logging.INFO)
@@ -225,10 +229,14 @@ def exec_process(args: str, **extra_args) -> ProcessResult:
             output_args["stderr"] = errpipe
     try:
         result = subprocess.run(args, check=True, **output_args)
-        out = result.stdout
+        if outpipe:
+            out = outpipe.out
+        else:
+            out = result.stdout
         returncode = result.returncode
     except subprocess.CalledProcessError as exception:
-        out = exception.stderr
+        if errpipe:
+            out = errpipe.out
         returncode = exception.returncode
         cmd = exception.cmd
         # Rejoin the list for better readability.
@@ -236,7 +244,8 @@ def exec_process(args: str, **extra_args) -> ProcessResult:
             cmd = " ".join(cmd)
         log.error('Error %s when executing "%s".', returncode, cmd)
     except subprocess.TimeoutExpired as exception:
-        out = exception.stderr
+        if errpipe:
+            out = errpipe.out
         returncode = FAILURE
         cmd = exception.cmd
         # Rejoin the list for better readability.
@@ -245,9 +254,9 @@ def exec_process(args: str, **extra_args) -> ProcessResult:
         log.error("Timed out when executing %s.", cmd)
     finally:
         if "capture_output" not in extra_args:
-            if "stdout" not in extra_args:
+            if outpipe:
                 outpipe.close()
-            if "stderr" not in extra_args:
+            if errpipe:
                 errpipe.close()
     return ProcessResult(out, returncode)
 ```

**Comment by thomascalvert-xlnx** [comment]
@fruffy many thanks for the patch, much cleaner solution and it works for me on the 2 test cases (op_bin and header_unaligned) :)

Before merging my suggestion would be to squash the [fixup] patch into [67b4f6d](https://github.com/p4lang/p4c/pull/4185/commits/67b4f6d90ef2cbe4b7507c6523274dabc4922060), and rebase the whole thing onto main. Alternatively we could also just squash the entire PR into one commit :shrug:

**Comment by fruffy** [comment]
We usually always do squash commits when merging into main. Is that what you mean?

**Comment by thomascalvert-xlnx** [comment]
The changes in this PR are fairly orthogonal so I thought it logical to keep them as separate commits - happy to help rewrite history to that aim. That being said, I don't feel strongly so equally happy to go with this repo's convention.

**Comment by fruffy** [comment]
Oh I see what you mean. We could make a separate PR. But that would prolong merging this commit at least for 1-2 days if that is okay with you.

**Comment by thomascalvert-xlnx** [comment]
I'm not in any hurry. I have rebased the PR onto top of tree and squashed the fixup as mentioned above. The overall diff is identical to [9b7e9ab37](https://github.com/p4lang/p4c/commit/9b7e9ab3738c6f2423a67c7a9afe182898480cac).
```
* 491b78c28 (HEAD -> ebpf-offsets, tcalvert/ebpf-offsets) testutils: Fix LogPipe to always record output.
* 65fe205a7 backends/ebpf: Output an error if a header is not byte-aligned (#4176).
* d9bc60c7e backends/ebpf: Support errors testing.
* fdf58f16a (tcalvert/main, origin/main, origin/HEAD, main) Update BUILD.bazel (#4192)
```

fruffy:
Merged as rebase commit to preserve the structure.

PR #4191, opened 2023-10-12
[P4Testgen] Only look ahead for the else branch when it is not null.

fruffy-g:
Also check for null in node coverage.


jonathan-dilorenzo (approved):
LGTM

smolkaj (approved):
LGTM

fruffy:
@jafingerhut @jnfoster 
Are we clear to add @smolkaj and @jonathan-dilorenzo as approved submitters to the compiler, so that they are able to review pull requests etc? Do we have a process for that?

PR #4193, opened 2023-10-13
Stop endianess conversion in parser

komaljai:
Stop endianess conversion from big endian to little endian in parser.


usha1830 (approved):
LGTM

PR #4196, opened 2023-10-16
Fix BMv2 backend to support non-field-access assignment LHS in parser block

### LINKED ISSUE #4194: Non-field-access assignment in parser is not supported on BMv2 target
This is a sub-issue of #3429. Fixing this is a necessary but not sufficient step towards fixing the original `hs.next` assignment issue. So I'm creating a new issue to track this incremental step.

Below is a minimal example to reproduce this issue:

`test.p4`:
```p4
#include <core.p4>
#include <v1model.p4>

header hs_item_t {
    bit<8> field;
}

struct metadata_t {
    hs_item_t[4] hs;
    bit<2> hs_next_index;
}

header data_t {
    bit<8> f;
}

struct header_t {}

parser TestParser(packet_in pkt,
                  out header_t hdr,
                  inout metadata_t meta,
                  inout standard_metadata_t std_meta) {
    state start {
        meta.hs[meta.hs_next_index].setValid();
        meta.hs[meta.hs_next_index] = {8w0};
        meta.hs_next_index = meta.hs_next_index + 1;
        transition accept;
    }
}

control DefaultDeparser(packet_out pkt, in header_t hdr) {
    apply {
        pkt.emit(hdr);
    }
}

control EmptyControl(inout header_t hdr, inout metadata_t meta, inout standard_metadata_t std_meta) {
    apply {}
}

control EmptyChecksum(inout header_t hdr, inout metadata_t meta) {
    apply {}
}

V1Switch(
    TestParser(),
    /*VerifyChecksum=*/EmptyChecksum(),
    /*Ingress=*/EmptyControl(),
    /*Egress=*/EmptyControl(),
    /*ComputeChecksum=*/EmptyChecksum(),
    DefaultDeparser()
) main;
```

```shell
$ p4c --target bmv2 --arch v1model --std p4-16 test.p4
$ sudo simple_switch test.json
Parser 'set' operation has invalid destination type 'expression', expected 'field'
bad json:
{
   "op" : "set",
   "parameters" : [
      {
         "type" : "expression",
         "value" : {
            "left" : {
               "type" : "expression",
               "value" : {
                  "left" : {
                     "type" : "header_stack",
                     "value" : "userMetadata.hs"
                  },
                  "op" : "dereference_header_stack",
                  "right" : {
                     "type" : "field",
                     "value" : [ "scalars", "metadata_t.hs_next_index" ]
                  }
               }
            },
            "op" : "access_field",
            "right" : 0
         }
      },
      {
         "type" : "hexstr",
         "value" : "0x00"
      }
   ]
}
```

qobilidop:
Fixes #4194.


qobilidop:
@fruffy The failures seem to be P4Testgen related. Do you have an idea what might be the issue?

fruffy:
This looks like a problem with the [parserUnroll](https://github.com/p4lang/p4c/blob/main/midend/parserUnroll.h#L298) pass. P4Testgen uses it. That pass is unfortunately quite complex and I do not know it very well. This might be difficult to debug. 
You could add these failures to Xfails.

mihaibudiu:
These are SIGSEGV, should be easy to diagnose and fix.

fruffy:
> These are SIGSEGV, should be easy to diagnose and fix.

Well the crash happens at [here](https://github.com/p4lang/p4c/blob/main/midend/parserUnroll.cpp#L175) because it is trying to clone the constant of an uninitialized `SymbolicInteger`. Why that integer is uninitialized I am not sure.

mihaibudiu:
A SymbolicInteger does not have to have a constant value (otherwise it would be just a constant).
If the constant value is nullptr then this function should probably just return, similar to the test two lines above.

VolodymyrPeschanenkoLitSoft:
> This looks like a problem with the [parserUnroll](https://github.com/p4lang/p4c/blob/main/midend/parserUnroll.h#L298) pass. P4Testgen uses it. That pass is unfortunately quite complex and I do not know it very well. This might be difficult to debug. You could add these failures to Xfails.
> 
> @VolodymyrPeschanenko Could you take a look if you have time? Maybe it is an easy fix.

I'll see on it tomorrow.

backends/p4tools/modules/testgen/targets/bmv2/test/BMV2MetadataXfail.cmake
```diff
@@ -65,6 +65,14 @@ p4tools_add_xfail_reason(
   # and more specifically when member is passed to getTypeArray
 )
 
+p4tools_add_xfail_reason(
+  "testgen-p4c-bmv2-metadata"
+  "Compiler Bug: Exiting with SIGSEGV"
```
mihaibudiu, line 70: I think you can easily fix this one

testdata/p4_16_samples/parser-non-field-lhs-bmv2.p4
```diff
@@ -0,0 +1,73 @@
...
+    bit<2> hs_next_index;
+}
+
+parser TestParser(packet_in pkt,
+                  out header_t hdr,
+                  inout metadata_t meta,
+                  inout standard_metadata_t std_meta) {
+    state start {
+        hdr.hs[meta.hs_next_index].setValid();
+        hdr.hs[meta.hs_next_index] = {8w0};
```
jafingerhut, line 24: The initial value of meta.hs_next_index is unspecified in this program.  Is that what you intended?  If not, it would be better to explicitly initialize meta.hs_next_index before its first use, e.g. by assigning it 0.

testdata/p4_16_samples/parser-non-field-lhs-bmv2.stf
```diff
@@ -0,0 +1,2 @@
+packet 0  DEADBEEF DEADBEEF
+expect 0  00010203 DEADBEEF
```
jafingerhut, line 2: The output packet should be `00010203 DEADBEEF DEADBEEF`, yes?

If so, this test would be more strict and well-documented if you changed the `expect` line to this:
```
expect 0 00010203 DEADBEEF DEADBEEF $
```
The `$` at the end of an `expect` line in an STF test means "and that should be the end of the output packet".  Without it, `expect` lines only check that the packet begins as specified, but can have any additional data following that.

testdata/p4_16_samples/parser-non-field-lhs-bmv2.p4
VolodymyrPeschanenkoLitSoft: This is a good solution for this PR. In the general case, ParserUnroll can't generate an exception for such examples. So, I'll try to fix it.

jafingerhut: I am actually a bit surprised that there is no warning from p4c that `meta.hs_next_index` is used without being initialized.

qobilidop: Thanks for catching this! I've added the initialization as you suggested.

testdata/p4_16_samples/parser-non-field-lhs-bmv2.stf
qobilidop: You are right! My understanding of how this program should behave was wrong. And I learned a bit more STF here.

qobilidop:
The failing DPDK PTF test `dpdk-ptf/testdata/p4_16_samples/pna-dpdk-add_on_miss0.p4` seems irrelevant to this PR.

fruffy:
The failure is caused by instability in `infrap4d`, you can ignore it.

fruffy:
The encountered bug is covered in #4200.

fruffy:
Is this ready to be merged?

testdata/p4_16_samples_outputs/parser-non-field-lhs-bmv2-first.p4
```diff
@@ -0,0 +1,62 @@
...
+        hdr.hs[meta.hs_next_index].setValid();
+        hdr.hs[meta.hs_next_index].field = 8w0;
+        hdr.hs[meta.hs_next_index].field[3:0] = 4w2;
+        meta.hs_next_index = meta.hs_next_index + 2w1;
+        transition hs_3;
+    }
+    state hs_3 {
+        hdr.hs[hdr.hs[0].field + 8w3].setValid();
+        hdr.hs[hdr.hs[1].field + 8w2].field = 8w0;
+        hdr.hs[hdr.hs[2].field + 8w1].field[3:0] = 4w3;
```
antoninbas, line 41: could we use a `lookahead` expression here for the RHS to make the test more comprehensive?

backends/bmv2/common/parser.cpp
```diff
@@ -106,11 +143,17 @@ Util::IJson *ParserConverter::convertParserStatement(const IR::StatOrDecl *stat)
...
-        auto type = ctxt->typeMap->getType(assign->left, true);
-        cstring operation = jsonAssignment(type, true);
+        auto expr = assign->left;
+        auto type = ctxt->typeMap->getType(expr, true);
+        cstring operation = jsonAssignment(type, expr, true);
         result->emplace("op", operation);
-        auto l = ctxt->conv->convertLeftValue(assign->left);
         bool convertBool = type->is<IR::Type_Boolean>();
+        Util::IJson *l;
+        if (ctxt->conv->isArrayIndexRuntime(assign->left)) {
```
antoninbas, line 152: I am not very familiar with the bmv2 backend code, maybe it would be nice to have a comment to explain why this is handled as a special case?

Shouldn't `convertLeftValue` also be able to handle this case gracefully?

backends/bmv2/common/parser.cpp
```diff
@@ -25,7 +25,43 @@ limitations under the License.
...
+
+field_expr
+    : prefixedNonTypeName
+    | field_expr dot_name
+    | field_expr "[" constant "]"
+*/
+bool ParserConverter::isFieldExpr(const IR::Expression *expr) {
+    if (expr->is<IR::PathExpression>()) return true;
+    if (expr->is<IR::Member>()) {
+        auto member = expr->to<IR::Member>();
```
antoninbas, line 53: nit: I think you can keep the code less verbose by using `if (auto member = expr->to<IR::Member>())`. I think the rest of the p4c codebase uses that a lot.

backends/bmv2/common/parser.cpp
```diff
@@ -36,9 +72,10 @@ cstring ParserConverter::jsonAssignment(const IR::Type *type, bool inParser) {
         else
             return "assign_header_stack";
     }
-    if (inParser)
+    if (inParser && isFieldExpr(expr))
         // Unfortunately set can do some things that assign cannot,
         // e.g., handle lookahead on the RHS.
+        // One limitation of set is that its LHS has to be a field expr.
```
antoninbas, line 78: my understanding is that at this stage, we are guaranteed that if there is a `lookahead` on the RHS, the LHS has to be a field expression. @mihaibudiu mentioned that there was a pass to extract all `lookahead`s and assign their value to a temporary metadata field.

Maybe the comment should be updated a bit:
> Unlike "assign", "set" is implemented more efficiently in bmv2, and can handle a lookahead expression as the RHS. However, "set" requires its LHS to be a field expression.
> The RemoveComplexExpressions pass guarantees that all lookahead<T>() expressions are lifted into tmp = lookahead<T>(). Therefore, we are guaranteed that anytime the RHS includes a lookahead, it will be a bare lookahead call and the LHS will be a simple field expression. These assignments will be handled by this if branch, using "set".

backends/bmv2/common/parser.cpp
qobilidop: Sure, will do. Just want to give a short explanation here before I have time to address other comments. This part was added to address the issue described in https://github.com/p4lang/p4c/issues/4194#issuecomment-1763706371. One observation I had was that the issue only occurs in parser block, but not control block. So I looked at the corresponding code for control block (see below), and basically copied the diff here without fully understanding why it works.

https://github.com/p4lang/p4c/blob/ac40e1f0812fac8e5d0744bc7febb4c1fb0e6a1e/backends/bmv2/common/action.cpp#L122-L126

testdata/p4_16_samples_outputs/parser-non-field-lhs-bmv2-first.p4
qobilidop: Done. I've also identified a bug in the `ExpandLookahead` pass and fixed it.

backends/bmv2/common/parser.cpp
qobilidop: Done.

backends/bmv2/common/parser.cpp
qobilidop: Updated.

backends/bmv2/common/parser.cpp
qobilidop: I've dug up where this logic came from: https://github.com/p4lang/p4c/pull/2139/files#r369298969.

To quote @hesingh:
> It is needed to deal with the case when a runtime index is a left value and is the only expression in JSON (without any other arithmetic operations). The JSON format needs an extra wrapping `"type": "expression"`.
> 
> Regarding convertLeftValue, note, the right value is also using `convert` because simple_switch supports runtime index in right value with simple_switch. See this code few lines below the left code.
> 
> `auto right = ctxt->conv->convert(r, true, true, convertBool);`

qobilidop: I've added this explanation as comments in the code.

midend/expandLookahead.cpp
```diff
@@ -78,7 +79,7 @@ DoExpandLookahead::ExpansionInfo *DoExpandLookahead::convertLookahead(
               em->method);
     auto targ = expression->typeArguments->at(0);
     auto typearg = typeMap->getTypeType(targ, true);
-    if (!typearg->is<IR::Type_StructLike>() && !typearg->is<IR::Type_Tuple>()) return nullptr;
+    if (!SymbolicValueFactory(typeMap).isFixedWidth(typearg)) return nullptr;
```
qobilidop, line 82: Tests are failing because this is changing many test outputs. Will move this to a separate PR.

qobilidop: Moved to #4264. Will continue the testing of this PR once that PR is merged.

mihaibudiu: It looks really weird to depend on SymbolicValueFactory in this pass.
Why can't the isFixedWidth function be a method of the typeMap, for example?

qobilidop: Makes sense. I'll refactor it to be a method of typeMap.

qobilidop: Done in #4264.

PR #4197, opened 2023-10-16
Ensured that error messages cannot be demoted to info messages

dmatousek:
The PR ensures that error messages cannot be demoted to info messages.

With `p4c --Winfo=invalid ...`, the following error is printed out:

`[--Werror=invalid] error: Error invalid cannot be demoted`

### DIFF
--- a/frontends/common/parser_options.cpp
+++ b/frontends/common/parser_options.cpp
@@ -206,6 +206,10 @@ ParserOptions::ParserOptions() : Util::Options(defaultMessage) {
         "--Winfo", "diagnostic",
         [](const char *diagnostic) {
             if (diagnostic) {
+                if (ErrorCatalog::getCatalog().isError(diagnostic)) {
+                    ::error(ErrorType::ERR_INVALID, "Error %1% cannot be demoted", diagnostic);
+                    return false;
+                }
                 P4CContext::get().setDiagnosticAction(diagnostic, DiagnosticAction::Info);
             }
             return true;
--- a/lib/error_catalog.h
+++ b/lib/error_catalog.h
@@ -112,6 +112,23 @@ class ErrorCatalog {
         return "--unknown--";
     }
 
+    /// retrieve the code for name
+    int getCode(cstring name) {
+        auto it =
+            std::find_if(errorCatalog.begin(), errorCatalog.end(),
+                         [name](const std::pair<int, cstring> &p) { return p.second == name; });
+        if (it != errorCatalog.end()) return it->first;
+        return -1;
+    }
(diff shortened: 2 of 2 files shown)


fruffy:
I have a clarification question related to the previous pull request #4184. What does `INFO_INFERRED` actually mean? When is that type of message meant to be used? 

The reason I am asking is because I believe info messages can be generally useful and replace the use of `printf` and `std::cout`  in the compiler. It would be nice to just have a generic info type.

dmatousek:
> I have a clarification question related to the previous pull request #4184. What does `INFO_INFERRED` actually mean? When is that type of message meant to be used?

It is meant to be used in cases when some piece of information is inferred by the compiler. An example can be inferring a default value for an uninitialized variable. Another example can be inferring connectivity of architecture-owned externs if the user does not specify it explicitly.

> The reason I am asking is because I believe info messages can be generally useful and replace the use of `printf` and `std::cout` in the compiler. It would be nice to just have a generic info type.

We discussed this internally. However, we came to the conclusion that it's beneficial for the info messages to be categorized in the same way as warning and error messages are so that you can work with them in the same way.

I expect that the list of info types will be extending. Are you aware of any `printf`/`std::cout` that could be replaced with `::info`?

fruffy (approved):

fruffy:
> > I have a clarification question related to the previous pull request #4184. What does `INFO_INFERRED` actually mean? When is that type of message meant to be used?
> 
> It is meant to be used in cases when some piece of information is inferred by the compiler. An example can be inferring a default value for an uninitialized variable. Another example can be inferring connectivity of architecture-owned externs if the user does not specify it explicitly.
> 
> > The reason I am asking is because I believe info messages can be generally useful and replace the use of `printf` and `std::cout` in the compiler. It would be nice to just have a generic info type.
> 
> We discussed this internally. However, we came to the conclusion that it's beneficial for the info messages to be categorized in the same way as warning and error messages are so that you can work with them in the same way.
> 
> I expect that the list of info types will be extending. Are you aware of any `printf`/`std::cout` that could be replaced with `::info`?

Thanks for clarifying, makes sense. 

`std::cout` is [scattered](https://github.com/search?q=repo%3Ap4lang%2Fp4c%20std%3A%3Acout&type=code) throughout the compiler back ends. P4Tools simply abuses the logging framework to achieve [a similar effect](https://github.com/p4lang/p4c/blob/main/backends/p4tools/modules/testgen/lib/logging.h#L28C28-L28C28) as info messages.

vlstill (approved):

PR #4200, opened 2023-10-18
ParserUnroll: fix seg. fault for uninitialized variables inside header stack indexes

VolodymyrPeschanenkoLitSoft:
These changes fix the segmentation fault inside ParserUnroll for uninitialized variables that are used as a header stack index. In this case, ParserUnroll generates a warning message and leaves the program as is.


fruffy:
Thanks for fixing this!

midend/parserUnroll.cpp
```diff
@@ -172,6 +173,12 @@ class ParserStateRewriter : public Transform {
         ExpressionEvaluator ev(refMap, typeMap, valueMap);
         auto *value = ev.evaluate(expression->right, false);
         if (!value->is<SymbolicInteger>()) return expression;
+        if (!value->to<SymbolicInteger>()->isKnown()) {
+            ::warning(ErrorType::ERR_INVALID,
+                      "ParserUnroll can't conretize uninitialized value:\n%1%", expression->right);
```
jafingerhut, line 178: typo: conretize -> concretize ?

midend/parserUnroll.cpp
```diff
@@ -172,6 +173,12 @@ class ParserStateRewriter : public Transform {
         ExpressionEvaluator ev(refMap, typeMap, valueMap);
         auto *value = ev.evaluate(expression->right, false);
         if (!value->is<SymbolicInteger>()) return expression;
+        if (!value->to<SymbolicInteger>()->isKnown()) {
+            ::warning(ErrorType::ERR_INVALID,
+                      "ParserUnroll can't concretize uninitialized value:\n%1%", expression->right);
```
mihaibudiu, line 178: I think this error message could be more user-friendly.
Most users won't know what 'concretize' means here.

VolodymyrPeschanenkoLitSoft:
> Thanks for fixing this!

You're welcome ;). It was nice to remember this code.

midend/parserUnroll.cpp
mihaibudiu: Perhaps "uninitialized value prevents loop unrolling". That's not great, but it's a bit better.

jafingerhut:
Are the files in the testdata/p4_16_samples_output/parser-unroll directory used for anything during CI testing? If yes, what for?

VolodymyrPeschanenkoLitSoft:
> Are the files in the testdata/p4_16_samples_output/parser-unroll directory used for anything during CI testing? If yes, what for?
These files are used to store results for p4test when it runs with the `--loopsUnroll` parameter.

fruffy (approved):

backends/p4tools/modules/testgen/targets/bmv2/test/BMV2STFXfail.cmake
```diff
@@ -235,6 +235,12 @@ p4tools_add_xfail_reason(
   issue584-1-bmv2.p4
 )
 
+p4tools_add_xfail_reason(
+  "testgen-p4c-bmv2-stf"
+  "is not a constant"
```
fruffy, line 240: Nitpick: This could be more precise with an comment on the actual error message. 
Like
https://github.com/fruffy/flay/blob/master/targets/v1model/test/BMv2V1ModelXfail.cmake#L8
for example.

jafingerhut (approved):
LGTM

Issue #4066, opened 2023-07-14
Support `.textproto` extension for output files in Protobuf Text Format


smolkaj:
The official file extension for Protobuf Text Format files is `.textproto`: https://protobuf.dev/reference/protobuf/textformat-spec/#text-format-files

But IIUC, currently p4c  requires `.txt` instead in a bunch of places, e.g. to generate P4infos or table entry files.

Today:
```
    p4test --arch v1model \
        --p4runtime-files p4info.txt \
        --p4runtime-entries-files entries.txt \
        table-entries-ternary-bmv2.p4
```

Proposed:
```
    p4test --arch v1model \
        --p4runtime-files p4info.textproto \
        --p4runtime-entries-files entries.textproto \
        table-entries-ternary-bmv2.p4
```

The proposal is to start supporting `.textproto` and, optionally, to slowly phase out support for `.txt`, e.g. by initially giving a warning and eventually failing with an error.

@jafingerhut for visibility, as this came up in my review of https://github.com/p4lang/p4runtime/pull/432.

qobilidop:
Just one minor update. It seems the official recommendation has become `.txtpb` (from the same source: https://protobuf.dev/reference/protobuf/textformat-spec/#text-format-files).

smolkaj:
Thanks for the update!

(I do wonder how many times more we will decide to change this recommendation.....)

PR #4201, opened 2023-10-19
[P4Testgen] Change Protobuf backend test case file extension to `.txtpb`

qobilidop:
With the same reasoning as #4066. `.proto` is usually used for the schema definition file, while `.txtpb` is the current recommendation for the [text format file](https://protobuf.dev/reference/protobuf/textformat-spec/#text-format-files).


fruffy (approved):

fruffy (changes requested):
Can you please also update 
https://github.com/p4lang/p4c/blob/main/backends/p4tools/modules/testgen/benchmarks/test_coverage.py#L203
?

qobilidop:
Done.

fruffy (approved):

PR #4206, opened 2023-10-23
generic -O option to set optimization level/options

ChrisDodd:
- disable inlining in frontend with -O0

### DIFF
--- a/frontends/p4/frontend.cpp
+++ b/frontends/p4/frontend.cpp
@@ -187,22 +187,26 @@ const IR::P4Program *FrontEnd::run(const CompilerOptions &options, const IR::P4P
         new DefaultValues(&refMap, &typeMap),
         new BindTypeVariables(&refMap, &typeMap),
         new EntryPriorities(&refMap),
-        new PassRepeated(
-            {new SpecializeGenericTypes(&refMap, &typeMap),
-             new DefaultArguments(&refMap, &typeMap),  // add default argument values to parameters
-             new ResolveReferences(&refMap),
-             new SetStrictStruct(&typeMap, true),          // Next pass uses strict struct checking
-             new TypeInference(&refMap, &typeMap, false),  // more casts may be needed
-             new SetStrictStruct(&typeMap, false),
-             new SpecializeGenericFunctions(&refMap, &typeMap)}),
+        new PassRepeated({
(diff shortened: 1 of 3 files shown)


frontends/p4/frontend.cpp
```diff
@@ -227,30 +230,37 @@ const IR::P4Program *FrontEnd::run(const CompilerOptions &options, const IR::P4P
...
+            new Inline(&refMap, &typeMap, evaluator, options.optimizeParserInlining),
+            new InlineActions(&refMap, &typeMap),
+            new LocalizeAllActions(&refMap),
+            new UniqueNames(&refMap),
+            new UniqueParameters(&refMap, &typeMap),
+            // Must be done before inlining functions, to allow
+            // function calls used as action arguments to be inlined
+            // in the proper place.
+            new RemoveActionParameters(&refMap, &typeMap),
+            new InlineFunctions(&refMap, &typeMap),
```
mihaibudiu, line 246: Some later stages that don't know about functions or action calls may have problems. I guess you will discover quickly.

mihaibudiu (approved):

frontends/p4/frontend.cpp
ChrisDodd: This defaults to `-O1`, so these backends will only have problems if given an explicit `-O0`.  However, the inlining should probably be moved to the midend as inlining being required (or not) is really target specific.

frontends/p4/frontend.cpp
```diff
@@ -158,6 +158,7 @@ const IR::P4Program *FrontEnd::run(const CompilerOptions &options, const IR::P4P
     refMap.setIsV1(isv1);
 
     auto evaluator = new P4::EvaluatorPass(&refMap, &typeMap);
+    // clang-format off  -- clang-format scrambles this into unreadable mess
```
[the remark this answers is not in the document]
ChrisDodd: Yes, adding a trailing comma seems to fix the strange clang-format formatting that occurs -- I'll use that instead.

frontends/p4/frontend.cpp
mihaibudiu: Ideally it should be in the midend, but it does eliminate some constructs that later passes do not expect, so it may not be easy to do that.

PR #4207, opened 2023-10-23
[P4Testgen] Fix textproto generation and escape traces properly.

### LINKED ISSUE #4202: [P4Testgen] How to capture traces in generated test cases?
I see there is `traces` field in the `TestCase` proto:

https://github.com/p4lang/p4c/blob/112b734b5b23500c53c199b30bb2d914c52c146c/backends/p4tools/modules/testgen/targets/bmv2/proto/p4testgen.proto#L24-L36

But when I use `p4testgen` with the `--print-traces` option (for a v1model/bmv2 P4 program), I only see traces printed out in terminal, but not recorded in the generated test cases.

I wonder if this is the expected behavior? And how can I capture traces in generated test cases?

fruffy:
Cheap workaround to deal with double quote characters within a trace object. Also fix up the Protobuf format for P4Testgen test messages. We are currently not testing these, so breakages were not detected.

Fixes #4202. @qobilidop


smolkaj (approved):
LGTM

PR #4209, opened 2023-10-26
Fix P4TC issues

komaljai:
Issues addressed -

1) In the parser generation remove an extraneous semicolon from file include code. 
2) In the parser generated code declare compiler_meta struct:
struct pna_global_metadata *compiler_meta__ = (struct pna_global_metadata *) skb->cb;
3) Make generated template file to have executable permissions
4) Generate template file by default (should have same name as pipeline)
5) Generate the introspection file by default with the SAME name as the pipeline.
6) Rename kfunc bpf_skb_p4tc_tbl_read to bpf_skb_p4tc_tbl_read
7) Move declaration of 'struct hdr_md' from post_parser.cpp to header.h file
8) Change post_parser.c file name to control_block.c
9) Remove '-i' and -'c' options.


usha1830 (approved):
Looks good to me

PR #4210, opened 2023-10-29
Repair the static DPDK PTF tests.

### LINKED ISSUE #4204: Convert existing DPDK PTF tests in testdata to the new testing format
Existing tests still use the runtime shell and are not function. 
Migrate these new tests to the new testing library which was introduced with #4173.

fruffy:
Fixes #4204.


testdata/p4_16_samples/pna-dpdk-add_on_miss0.py
```diff
@@ -16,170 +16,168 @@
...
         #                          tcp_sport=sport, tcp_dport=dport)
 
-        
-        tu.verify_packets(self, exp_pkt_for_miss, [eg_port])
-        logging.info("    packet experienced a miss in ct_tcp_table as expected")
+        ptfutils.verify_packets(self, exp_pkt_for_miss, [eg_port])
+        bt.testutils.log.info("packet experienced a miss in ct_tcp_table as expected")
+
+        bt.testutils.log.info("Attempting to delete all entries in ipv4_host")
+        # TODO: This code does not seem functional?
```
fruffy, line 156: @jafingerhut I am unable to delete table entries on the P4Runtime interface for the DPDK SoftNIC. Even when I read the entries from the table, then try to delete those exact same entries I am getting a "not found" error. Is this a known issue?

The P4Runtme shell code also has the same issue, but it was never uncovered because it was only ever used on tables with no entries.

jafingerhut: I can double-check, but if I recall correctly, the current DPDK implementation does not support adding or deleting entries from the control plane specifically for add-on-miss tables.  For add-on-miss tables, it only supports adding and deleting entries from the data plane.

fruffy: Ah good to know. Strangely enough I am able to add entries, but I can not remove them.

fruffy:
@jafingerhut I would like to merge this code before further bitrot sets in. This PR simply cleans up the non-functional PTF DPDK code in the compiler and makes it possible to run it again.

The CI tests remain disabled since we have not been able to solve the nondeterministic failures.

PR #4212, opened 2023-10-30
[WIP] Use IPDK-container to Spped-up DPDK-PTF CI

Hoooao:
Now the DPDK-PTF takes ~20min to finish with ccache and ~1h to finish without ccache. Saved ~40min from building IPDK/infrap4d and related dependencies. 

Issue:
    We have 2 non-deterministic issues seemingly related to memory and gRPC. See https://github.com/Hoooao/p4c-dpdk-ci-pipeline/actions/runs/6663670233/job/18109927830


.github/workflows/CI-DOCKER-DPDK.yml
```diff
@@ -0,0 +1,52 @@
...
+
+      - name: Build IPDK
+        run: |
+             cp p4c/tools/dpdk_ptf.Dockerfile .
+             docker build -t dpdk_ptf -f dpdk_ptf.Dockerfile .
+
+      # hugepage has to be set in the same shell session
+      - name: Run DPDK PTF tests using P4Testgen
+        run: |
+             sudo -E docker run --privileged -w /root/p4c/build dpdk_ptf bash -c "../../scripts/set_hugepages.sh && ctest -j1 --output-on-failure -R testgen-p4c-pna-ptf"
```
fruffy, line 48: Can we just map hugepages instead?

https://superuser.com/questions/1309438/how-to-enable-huge-page-support-in-a-dockerfile

.github/workflows/CI-DOCKER-DPDK.yml
```diff
@@ -0,0 +1,52 @@
...
+      - name: Checkout DPDK-target
+        uses: actions/checkout@v3
+        with:
+             repository: p4lang/p4-dpdk-target
+             path: p4sde
+             submodules: recursive
+
+      - name: Build IPDK
+        run: |
+             cp p4c/tools/dpdk_ptf.Dockerfile .
```
fruffy, line 42: Why copy the file instead of just specifying the path? Is there a particular problem with the setup?

tools/dpdk_ptf.Dockerfile
```diff
@@ -0,0 +1,54 @@
+FROM ghcr.io/ipdk-io/ipdk-ubuntu2004-x86_64:ipdk_v23.07
```
fruffy, line 1: `tools` might be a strange folder for this file. Maybe create a `docker` folder?

fruffy (changes requested):
First round of comments. These seems decently reliable already.

tools/dpdk_ptf.Dockerfile
```diff
@@ -0,0 +1,54 @@
...
+WORKDIR /root
+RUN rm -rf p4*
+COPY ./p4c /root/p4c
+COPY ./p4sde /root/p4sde
+COPY ./.ccache /root/.ccache
+RUN apt-get update
+RUN apt-get install rsync git ccache libhugetlbfs-bin -y
+RUN ccache -s
+
+# Update GCC to 11 -> GCC-9 cause config problem, remove after base OS is updated to 22.04
```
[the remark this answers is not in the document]
jafingerhut: I am guessing you are asking whether someone at Intel might publish a docker image with IPDK software already compiled and installed, rather than following instructions like these?  https://github.com/ipdk-io/ipdk/blob/main/build/networking/README_DOCKER.md

If so, let me know and I can try pinging a couple of people to find out if they have any intention of supporting such a thing.

.github/workflows/CI-DOCKER-DPDK.yml
Hoooao: It is for simplification in the Dockerfile where the folders, p4c and p4sde, can be copied from the current directory directly without having to go to parent directory for them.

.github/workflows/CI-DOCKER-DPDK.yml
Hoooao: But we want to use the script, in order to run the `set_hugepages.sh`, the ipdk repo has to be pulled, which is not used later.

fruffy: There is no `set_hugepages` in the install dir? In `sbin`?

Hoooao: it is in the container

fruffy: Oh duh, right. Seems like there is no easy solution then.

tools/dpdk_ptf.Dockerfile
```diff
@@ -0,0 +1,54 @@
...
+ENV INSTALL_EBPF="OFF"
+ENV IMAGE_TYPE="test"
+ENV CTEST_PARALLEL_LEVEL="4"
+ENV CMAKE_FLAGS="-DENABLE_P4TC=OFF -DENABLE_BMV2=OFF -DENABLE_EBPF=OFF -DENABLE_UBPF=OFF -DENABLE_GTESTS=OFF -DENABLE_P4TEST=OFF -DENABLE_P4C_GRAPHS=OFF -DIPDK_INSTALL_DIR=/root/ipdk_install "
+WORKDIR /root/p4c
+RUN apt-get install python3-dev -y # Required by packges in ci-build, remove after base OS is updated to 22.04
+RUN tools/ci-build.sh 
+
+# Disable github aciton default THP setting
+CMD ["sudo", "hugeadm", "--thp-madvise "]
```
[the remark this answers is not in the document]
Hoooao: I just tried to move it out. I think they are the same.

.github/workflows/CI-DOCKER-DPDK.yml
```diff
@@ -0,0 +1,52 @@
+name: "p4c-dpdk-ptf-p4testgen-tests"
```
fruffy, line 1: Can you merge this with the existing file?

Hoooao: sure

tools/dpdk_ptf.Dockerfile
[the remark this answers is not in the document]
fruffy: The docker image is built using a base OS that is Ubuntu 22.04, but the docker image itself is 20.04. That is normally fine, but the [p4-dpdk-target](https://github.com/p4lang/p4-dpdk-target) itself is built with Ubuntu 22.04. To get it to work on Ubuntu 20.04 requires a bunch of workarounds. It would be nice to remove this confounding factor. 

We thought we fixed the nondeterministic failures but that is sadly still not the case.

PR #4213, opened 2023-10-30
Fix some issues found by static analysis

vlstill:


fruffy (dismissed):
Which linter is this? Publicly available?

vlstill:
> Which linter is this? Publicly available?

Coverity, so no. My feeling is a lot of this could actually be found by clang-tidy too though, but I had not checked that yet.

backends/p4tools/modules/testgen/targets/ebpf/test_backend.cpp
```diff
@@ -66,12 +66,14 @@ TestBackEnd::TestInfo EBPFTestBackend::produceTestInfo(
             IR::getConstant(IR::getBitType(outPktSize), EBPFTestBackend::ZERO_PKT_VAL);
         testInfo.packetTaintMask =
             IR::getConstant(IR::getBitType(outPktSize), EBPFTestBackend::ZERO_PKT_MAX);
+    } else {
```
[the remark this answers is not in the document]
fruffy: This seems good. I am not even sure whether the condition can ever be true. The eBPF back end can not reduce packet size.

ir/v1.def
```diff
@@ -230,7 +230,7 @@ class CalculatedField : IAnnotated {
         bool                    update;
         ID                      name;
         Expression              cond;
-        update_or_verify() {}  // FIXME -- needed by umpack_json(safe_vector) -- should not be
+        update_or_verify() { update = false; }  // FIXME -- needed by umpack_json(safe_vector) -- should not be
```
fruffy, line 233: Should this be in the function call or in the initializer of `update`?

vlstill: I would like to put it into the initializer, but for some reason that breaks the generated code -- the parametric constructor has just 3 arguments. I don't know why that is, it seems quite weird to me. Therefore I've added it to the constructor. I'm quite surprised nested classes can be generated by IRGen.

fruffy: `optional bool           update = false;`
maybe?

vlstill:
@fruffy is this test working normally?
image
I've seen in failed also in

fruffy:
> @fruffy is this test working normally? image I've seen in failed also in

It is failing about 30% of the time. We are unsure what causes this failure but are trying to fix it. It might be caused by a particular Github runner being chosen.

ir/v1.def
vlstill: That worked, thanks.

PR #4215, opened 2023-10-31
[P4Testgen] Add a compiler pass to resolve Type_Name in StructExpressions.

fruffy:
~~In some cases, one may want to create a struct expression, which contains the relevant type directly, without having to use a `Type_Name` as indirection. This PR relaxes the constraints around the creation of struct expressions to make this possible. We simply add a new parameter that allows us to toggle whether or not to accept `IR::Type_StructLike` types.~~

When resolving struct expressions, simply preserve Type_Name for structType but assign the resolved type to type. This works for all dependent PRs and preserves functionality of passes such as toP4 etc.

### DIFF
--- a/backends/p4tools/common/CMakeLists.txt
+++ b/backends/p4tools/common/CMakeLists.txt
@@ -12,6 +12,7 @@ set(
 
   compiler/compiler_target.cpp
   compiler/convert_hs_index.cpp
+  compiler/convert_struct_expr.cpp
   compiler/convert_varbits.cpp
   compiler/midend.cpp
   compiler/reachability.cpp
--- a/backends/p4tools/common/compiler/convert_struct_expr.cpp
+++ b/backends/p4tools/common/compiler/convert_struct_expr.cpp
@@ -1,0 +1,25 @@
+#include "backends/p4tools/common/compiler/convert_struct_expr.h"
+
+#include "ir/irutils.h"
+
+namespace P4Tools {
+
+const IR::Node *ConvertStructExpr::postorder(IR::StructExpression *structExpr) {
+    auto structType = structExpr->type;
+    bool resolved = false;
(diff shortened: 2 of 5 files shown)


fruffy:
@mihaibudiu I updated the constructor to include an optional `structOkay` boolean.

mihaibudiu:
this can't be the whole PR, there is no use of this flag.

fruffy:
The flag is used in https://github.com/p4lang/p4c/pull/4231 and https://github.com/p4lang/p4c/pull/4255. And an extension which is not part of the main repository here: https://github.com/fruffy/flay

I factored out the change to make the PR more readable. 

I could add a compiler pass which resolves `Type_Name` for StructExpressions. It would make use of this flag. It might be useful for this PR, too: https://github.com/p4lang/p4c/pull/4225

mihaibudiu:
The pass which resolves the name is called ResolveReferences, it's perhaps the oldest pass in the compiler.
If a struct has no name there is nothing to resolve.

mihaibudiu (approved):

fruffy:
> The pass which resolves the name is called ResolveReferences, it's perhaps the oldest pass in the compiler.
> If a struct has no name there is nothing to resolve.

The idea is to resolve all `Type_Name` past the mid end since they serve no purpose after the passes have completed. This way we can access any type of any expression directly without requiring a map access. Any `IR::Expression` will have a fully usable type.

mihaibudiu:
Various backends execute additional transformation passes after the mid-end.
Moreover, this may break to toP4 pass.

fruffy:
@mihaibudiu I believe I found a more palatable solution which does not affect the ir or the rest of the compiler. When resolving struct expressions, I simply preserve Type_Name for `structType` but assign the resolved type to `type`. This works for all dependent PRs and preserves functionality of passes such as toP4 etc.

PR #4220, opened 2023-11-01
[P4Testgen] Move common code of BMv2 test back ends into a common class.

fruffy-g:
This is the second PR of a series of three to refactor the BMv2 test back ends. 

This PR moves code that is common to the increasing amount of  test back ends into a `common` class. This makes it easier to add new test back ends to the BMv2 P4Testgen extension. 

It also renables the BMv2 Protobuf back end and updates its Xfails.

### DIFF
--- a/backends/p4tools/modules/testgen/targets/bmv2/test/BMV2ProtobufXfail.cmake
+++ b/backends/p4tools/modules/testgen/targets/bmv2/test/BMV2ProtobufXfail.cmake
@@ -9,6 +9,13 @@
 ####################################################################################################
 # These are failures in P4Testgen that need to be fixed.
 
+p4tools_add_xfail_reason(
+  "testgen-p4c-bmv2-protobuf"
+  "Non-numeric, non-boolean member expression: .* Type: Type_Stack"
+  # We can not expand stacks in parsers because information about .next is lost.
+  # P4Testgen needs to maintain its own internal .next variable for stacks.
+  array-copy-bmv2.p4
+)
 
 p4tools_add_xfail_reason(
   "testgen-p4c-bmv2-protobuf"
@@ -115,6 +122,13 @@ p4tools_add_xfail_reason(
   "with type Type_Specialized is not a Type_Declaration"
   # Pipeline as a parameter of a switch, not a valid v1model program
   issue1304.p4
+)
+
+p4tools_add_xfail_reason(
+  "testgen-p4c-bmv2-protobuf"
+  "is not a constant"
+  # Using an uninitialized variable as a header stack index in the parser.
+  parser-unroll-test10.p4
 )
 
 ####################################################################################################
--- a/backends/p4tools/modules/testgen/targets/bmv2/test_backend/common.cpp
+++ b/backends/p4tools/modules/testgen/targets/bmv2/test_backend/common.cpp
@@ -1,0 +1,196 @@
+#include "backends/p4tools/modules/testgen/targets/bmv2/test_backend/common.h"
+
+#include <filesystem>
+#include <map>
+#include <optional>
+#include <string>
+#include <utility>
+#include <vector>
+
+#include <inja/inja.hpp>
+
+#include "backends/p4tools/common/lib/format_int.h"
+#include "ir/ir.h"
+#include "nlohmann/json.hpp"
+
+#include "backends/p4tools/modules/testgen/lib/exceptions.h"
+#include "backends/p4tools/modules/testgen/lib/tf.h"
+#include "backends/p4tools/modules/testgen/targets/bmv2/test_spec.h"
+
+namespace P4Tools::P4Testgen::Bmv2 {
+
+Bmv2TF::Bmv2TF(std::filesystem::path basePath, std::optional<unsigned int> seed)
+    : TF(std::move(basePath), seed) {}
+
+inja::json Bmv2TF::getClone(const TestObjectMap &cloneSpecs) const {
+    auto cloneSpec = inja::json::object();
+    auto cloneJsons = inja::json::array_t();
+    auto hasClone = false;
+    for (auto cloneSpecTuple : cloneSpecs) {
+        inja::json cloneSpecJson;
+        const auto *cloneSpec = cloneSpecTuple.second->checkedTo<Bmv2V1ModelCloneSpec>();
+        cloneSpecJson["session_id"] = cloneSpec->getEvaluatedSessionId()->asUint64();
+        cloneSpecJson["clone_port"] = cloneSpec->getEvaluatedClonePort()->asInt();
+        cloneSpecJson["cloned"] = cloneSpec->isClonedPacket();
+        hasClone = hasClone || cloneSpec->isClonedPacket();
+        cloneJsons.push_back(cloneSpecJson);
+    }
+    cloneSpec["clone_pkts"] = cloneJsons;
+    cloneSpec["has_clone"] = hasClone;
+    return cloneSpec;
+}
+
+inja::json::array_t Bmv2TF::getMeter(const TestObjectMap &meterValues) const {
+    auto meterJson = inja::json::array_t();
+    for (auto meterValueInfoTuple : meterValues) {
+        const auto *meterValue = meterValueInfoTuple.second->checkedTo<Bmv2V1ModelMeterValue>();
+
+        const auto meterEntries = meterValue->unravelMap();
+        for (const auto &meterEntry : meterEntries) {
+            inja::json meterInfoJson;
+            meterInfoJson["name"] = meterValueInfoTuple.first;
+            meterInfoJson["value"] = formatHexExpr(meterEntry.second.second);
+            meterInfoJson["index"] = formatHex(meterEntry.first, meterEntry.second.first);
+            if (meterValue->isDirectMeter()) {
+                meterInfoJson["is_direct"] = "True";
+            } else {
+                meterInfoJson["is_direct"] = "False";
+            }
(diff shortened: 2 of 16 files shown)


backends/p4tools/modules/testgen/targets/bmv2/test/BMV2ProtobufXfail.cmake
```diff
@@ -9,6 +9,13 @@
 ####################################################################################################
 # These are failures in P4Testgen that need to be fixed.
 
+p4tools_add_xfail_reason(
```
jonathan-dilorenzo, line 12: These are new errors that arose due to this refactoring?

jonathan-dilorenzo: Because you turned ON running the tests below I assume?

backends/p4tools/modules/testgen/targets/bmv2/test_backend/common.h
```diff
@@ -0,0 +1,43 @@
...
+
+#include <inja/inja.hpp>
+
+#include "backends/p4tools/modules/testgen/lib/test_object.h"
+#include "backends/p4tools/modules/testgen/lib/test_spec.h"
+#include "backends/p4tools/modules/testgen/lib/tf.h"
+
+namespace P4Tools::P4Testgen::Bmv2 {
+
+class Bmv2TF : public TF {
```
jonathan-dilorenzo, line 16: I'd write some comment clarifying that TF is TestFramework here... You could even spell it out in the name (which would be the usual google approach).

backends/p4tools/modules/testgen/targets/bmv2/test_backend/common.cpp
jonathan-dilorenzo, line 1: Assuming this is basically moved and slightly generalized?

backends/p4tools/modules/testgen/targets/bmv2/test_backend/metadata.cpp
```diff
@@ -19,62 +19,12 @@
...
 #include "backends/p4tools/modules/testgen/lib/test_object.h"
-#include "backends/p4tools/modules/testgen/lib/tf.h"
 #include "backends/p4tools/modules/testgen/targets/bmv2/test_spec.h"
 
 namespace P4Tools::P4Testgen::Bmv2 {
 
-Metadata::Metadata(std::filesystem::path basePath, std::optional<unsigned int> seed = std::nullopt)
-    : TF(std::move(basePath), seed) {}
-
-std::vector<std::pair<size_t, size_t>> Metadata::getIgnoreMasks(const IR::Constant *mask) {
```
jonathan-dilorenzo, line 30: I didn't see where this got moved to? Maybe it's just removed?

backends/p4tools/modules/testgen/targets/bmv2/test_backend/protobuf.cpp
```diff
@@ -228,7 +200,7 @@ inja::json Protobuf::getControlPlaneForTable(const TableMatchMap &matches,
     return rulesJson;
 }
 
-inja::json Protobuf::getSend(const TestSpec *testSpec) {
```
jonathan-dilorenzo, line 231: It's a bit odd to me that you're not also using the 'common' version here? It seems like it has to do with formatting, but perhaps that could come from another input to the function or something?

Same below.

backends/p4tools/modules/testgen/targets/bmv2/test_backend/ptf.cpp
```diff
@@ -91,125 +44,7 @@ std::vector<std::pair<size_t, size_t>> PTF::getIgnoreMasks(const IR::Constant *m
...
-    inja::json sendJson;
-    sendJson["ig_port"] = iPacket->getPort();
-    auto dataStr = formatHexExpr(payload, false, true, false);
-    sendJson["pkt"] = insertHexSeparators(dataStr);
-    sendJson["pkt_size"] = payload->type->width_bits();
-    return sendJson;
-}
-
-inja::json PTF::getVerify(const TestSpec *testSpec) {
+inja::json PTF::getExpectedPacket(const TestSpec *testSpec) const {
```
jonathan-dilorenzo, line 47: Again, this looks very similar except for the ignoring masks part and what I assume is minor formatting?

backends/p4tools/modules/testgen/targets/bmv2/test_backend/common.h
```diff
@@ -0,0 +1,43 @@
...
+ public:
+    explicit Bmv2TF(std::filesystem::path basePath,
+                    std::optional<unsigned int> seed = std::nullopt);
+
```
jonathan-dilorenzo, line 20: I'm quite surprised to not see a virtual OutputTest here. Can you maybe say a bit why that doesn't belong?

jonathan-dilorenzo (changes requested):

backends/p4tools/modules/testgen/targets/bmv2/test/BMV2ProtobufXfail.cmake
fruffy-g: Yes, exactly. These were errors that existed before but were not captured yet.

backends/p4tools/modules/testgen/targets/bmv2/test_backend/metadata.cpp
fruffy-g: Yes, turns out this code was not used at all.

backends/p4tools/modules/testgen/targets/bmv2/test_backend/common.h
fruffy-g: I can file an issue for this and expand `TF` in a separate PR across all open-source back ends? Should be a quick change.

backends/p4tools/modules/testgen/targets/bmv2/test_backend/protobuf.cpp
fruffy-g: Yes I would like to use the common version, but the way hex is interpreted in the other frameworks (STF for some reason uses raw hex without a prefix) versus how it is interpreted in text protobuf (using text separators) seems incompatible. Maybe I can simplify this.

backends/p4tools/modules/testgen/targets/bmv2/test_backend/common.h
fruffy-g: `outputTest` is already purely virtual in `TF`, but every test back end instantiation is slightly different. This makes it not worthwhile to provide a common, shared function.

backends/p4tools/modules/testgen/targets/bmv2/test_backend/ptf.cpp
fruffy-g: `getVerify` was not very descriptive, so I renamed the function.

backends/p4tools/modules/testgen/targets/bmv2/test_backend/common.h
jonathan-dilorenzo: SGTM

backends/p4tools/modules/testgen/targets/bmv2/test_backend/metadata.cpp
jonathan-dilorenzo: Haha, ok.

backends/p4tools/modules/testgen/targets/bmv2/test_backend/protobuf.cpp
jonathan-dilorenzo: I see. I could imagine using that as an input to the function somehow. Or having each subclass provide a 'InterpretHex' function of their own (e.g.).

I would ideally fix that in this CL, but I can approve it if you think it's better to do this as a second pass later.

backends/p4tools/modules/testgen/targets/bmv2/test_backend/ptf.cpp
jonathan-dilorenzo: Sorry, I meant that it looks very similar to the 'common' function, making me think they could be combined.

backends/p4tools/modules/testgen/targets/bmv2/test_backend/common.h
jonathan-dilorenzo: Ah, I see! That makes sense. Forgot that Bmv2TF inherits from TF! Thanks!

jonathan-dilorenzo (approved):

backends/p4tools/modules/testgen/targets/bmv2/test_backend/common.h
fruffy-g: https://github.com/p4lang/p4c/issues/4229

backends/p4tools/modules/testgen/targets/bmv2/test_backend/ptf.cpp
fruffy-g: There is a subtle difference: `getIgnoreMasks(payloadMask)` is called, which is not the case in the other frameworks. I will think about how to unify this.

backends/p4tools/modules/testgen/targets/bmv2/test_backend/protobuf.cpp
fruffy-g: https://github.com/p4lang/p4c/issues/4230

PR #4222, opened 2023-11-02
Still parse STF files, even when BMv2 model is not present in system

fruffy-g:
Small quality-of-life improvement. Still parse STF files even though the BMv2 model cannot be run.

### DIFF
--- a/backends/bmv2/run-bmv2-test.py
+++ b/backends/bmv2/run-bmv2-test.py
@@ -214,9 +214,6 @@ timeout = 10 * 60
 
 
 def run_model(options, tmpdir, jsonfile):
-    if not options.hasBMv2:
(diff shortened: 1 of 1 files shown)


jafingerhut:
Is the reason for this so that even if a developer has not installed BMv2, that running p4c CI still does syntax checks on the STF files?  Or is there some other benefit you have in mind here?

fruffy-g:
> Is the reason for this so that even if a developer has not installed BMv2, that running p4c CI still does syntax checks on the STF files? Or is there some other benefit you have in mind here?

Yes,  I ran into this exact problem. :) I erroneously assumed a test was passing even though the STF syntax was incorrect.

jafingerhut (approved):
LGTM

PR #4226, opened 2023-11-06
Fix build order for extensions

ChrisDodd:
According to the cmake docs, the order of `add_subdirectory` commands should not matter, but it seems like it does -- if anything in extensions wants to include the p4parser.hpp (bison output) generated as part of the `frontends`, it must be after `frontends` to get built in the right order.  This might just be a bug in some version(s) of cmake we're tripping over as it doesn't always fail with the wrong order, but seems to (reliably?) work with this order.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -396,6 +396,10 @@ set(P4RUNTIME_STD_DIR ${p4runtime_SOURCE_DIR}/proto CACHE INTERNAL
 )
 ########################################## P4Runtime End ##########################################
 
+add_subdirectory (frontends)
+add_subdirectory (midend)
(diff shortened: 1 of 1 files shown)


fruffy:
In my experience this might be caused by a target that does not depend on `frontend`  or `frontend-parser-gen`. When calling a parallel build with `make -jX` this target is built before frontend then.

ChrisDodd:
> In my experience this might be caused by a target that does not depend on `frontend` or `frontend-parser-gen`. When calling a parallel build with `make -jX` this target is built before frontend then.

`frontend-parser-gen` seems to be the magic incantation.  I'd tried to make it dependent on `p4Parser`, but that didn't work.

PR #4227, opened 2023-11-06
improved bootstrap.sh

ChrisDodd:


bootstrap.sh
```diff
@@ -13,12 +13,96 @@
...
+make_relative_link ()
+{
+    local target="${1:?make_relative_link: Argument 1 should be the target.}"
+    local link_name="${2:?make_relative_link: Argument 2 should be the link name.}"
+    local link_name_dir="$(dirname "${link_name}")"
+    mkdir -p "${link_name_dir}/"
+    ln -sf "$(realpath --relative-to "${link_name_dir}/" "${target}")" "${link_name}"
+}
+
+make_relative_link "${mydir}/.gdbinit" "backends/bmv2/p4c-bm2-psa-gdb.gdb"
```
fruffy, line 105: Maybe this is something to instantiate within CMakelists for each back end instead?

fruffy (approved):

fruffy:
The MacOS test is failing because of a substitution issue.

bootstrap.sh
ChrisDodd: If there's a way to do it in cmake, maybe.  I've had little luck with getting cmake to set up things like this, and it is not strictly needed for the build.  Just nice to have if you run gdb

ChrisDodd:
> The MacOS test is failing because of a substitution issue.

Ugh -- I guess MacOS's bash is not really bash?  They don't like GPL'd things.

ChrisDodd:
There seem to be some intermittent failures in the dpdk CI tests that this change is running into.

fruffy:
> There seem to be some intermittent failures in the dpdk CI tests that this change is running into.

Yes, unfortunately the P4-DPDK target is not stable. Trying to disable the tests here: #4228.

PR #4231, opened 2023-11-08
[P4Testgen]  Add struct and header expressions as a first-class expression to P4Testgen. 

fruffy:
This PR adds support for struct expressions in P4Testgen. Concretely, it adds utility functions to handle assignments and pass struct expressions as full argument. It also generalizes some utility functions to support BaseListExpressions, StructExpressions, and HeaderStackExpressions.

The goal of these changes is to enable two features:
- Support of meta information of complex P4 objects (such as validity and the `next` index of header stacks)
- Support of struct expressions as method call arguments. This will make implementation of externs significantly easier because we do not have to resolve the arguments within the extern any longer. Instead, we are able to access all necessary arguments directly. 

It also introduces a new IR type. HeaderExpressions. HeaderExpressions track the validity of the header as an expression, which makes it possible to track validity of a header symbolically and not just as a boolean constant.

### DIFF
--- a/backends/p4tools/common/core/abstract_execution_state.cpp
+++ b/backends/p4tools/common/core/abstract_execution_state.cpp
@@ -122,8 +122,8 @@ std::vector<const IR::Expression *> AbstractExecutionState::flattenComplexExpres
             auto subList = flattenComplexExpression(listElem->expression, flatValids);
             exprList.insert(exprList.end(), subList.begin(), subList.end());
         }
-        if (auto richStructExpr = structExpr->to<IR::HeaderExpression>()) {
-            flatValids.emplace_back(richStructExpr->validity);
+        if (auto headerExpr = structExpr->to<IR::HeaderExpression>()) {
+            flatValids.emplace_back(headerExpr->validity);
         }
     } else if (const auto *headerStackExpr = inputExpression->to<IR::HeaderStackExpression>()) {
         for (const auto *headerStackElem : headerStackExpr->components) {
@@ -189,19 +189,15 @@ void AbstractExecutionState::assignStructLike(const IR::StateVariable &left,
         std::vector<const IR::Expression *> flatRightValids;
         auto flatTargetFields = getFlatFields(left, &flatLeftValids);
         auto flatStructFields = flattenComplexExpression(structExpr, flatRightValids);
-        auto flatTargetSize = flatTargetFields.size();
-        auto flatStructSize = flatStructFields.size();
-        BUG_CHECK(flatTargetSize == flatStructSize,
+        BUG_CHECK(flatTargetFields.size() == flatStructFields.size(),
                   "The size of target fields (%1%) and the size of source fields (%2%) are "
                   "different.",
-                  flatTargetSize, flatStructSize);
-        auto flatLeftValidSize = flatLeftValids.size();
-        auto flatRightValidSize = flatRightValids.size();
+                  flatTargetFields.size(), flatStructFields.size());
         BUG_CHECK(
-            flatLeftValidSize == flatRightValidSize,
+            flatLeftValids.size() == flatRightValids.size(),
             "The size of target valid fields (%1%) and the size of source valid fields (%2%) are "
             "different.",
-            flatLeftValidSize, flatRightValidSize);
+            flatLeftValids.size(), flatRightValids.size());
         // First, complete the validity assignments for the data structure.
         for (size_t idx = 0; idx < flatLeftValids.size(); ++idx) {
             const auto &flatLeftValidRef = flatLeftValids[idx];
--- a/backends/p4tools/common/core/abstract_execution_state.cpp
+++ b/backends/p4tools/common/core/abstract_execution_state.cpp
@@ -74,6 +74,68 @@ void AbstractExecutionState::popNamespace() { namespaces = namespaces->pop(); }
  *  General utilities involving execution state.
  * ========================================================================================= */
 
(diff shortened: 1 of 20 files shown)


backends/p4tools/common/core/abstract_execution_state.h
```diff
@@ -87,11 +87,20 @@ class AbstractExecutionState {
...
+    /// StructExpression, or HeaderStackExpression. @returns nullptr if the expression is not
+    /// complex.
+    const IR::Expression *convertToComplexExpression(const IR::StateVariable &parent) const;
+
+    /// Takes in a complex expression as a StructExpression, ListExpression, or
+    /// HeaderStackExpression, flattens it into a vector and @returns this vector. In parallel this
+    /// function fills a vector of the validity variables of any Type_Header encountered in this
+    /// expression.
+    static std::vector<const IR::Expression *> flattenComplexExpression(
+        const IR::Expression *inputExpression, std::vector<const IR::Expression *> &flatValids);
```
vlstill, line 101: I suggest returning a pair of vectors (or named struct) instead of using out-parameter. Out-params are often considered code smell in modern C++.

backends/p4tools/common/core/abstract_execution_state.cpp
```diff
@@ -286,4 +383,19 @@ const IR::P4Action *AbstractExecutionState::getP4Action(
...
 
+const IR::P4Table *AbstractExecutionState::findTable(const IR::Member *member) const {
+    if (member->member != IR::IApply::applyMethodName) {
+        return nullptr;
+    }
+    if (member->expr->is<IR::PathExpression>()) {
+        const auto *declaration = findDecl(member->expr->to<IR::PathExpression>());
+        return declaration->to<IR::P4Table>();
+    }
+    const auto *type = member->expr->type;
```
vlstill, line 394: If which cases should the `findDecl` fail and lookup-by-type be required?

backends/p4tools/common/core/abstract_execution_state.cpp
```diff
@@ -74,6 +74,68 @@ void AbstractExecutionState::popNamespace() { namespaces = namespaces->pop(); }
...
+
+std::vector<const IR::Expression *> AbstractExecutionState::flattenComplexExpression(
+    const IR::Expression *inputExpression, std::vector<const IR::Expression *> &flatValids) {
+    std::vector<const IR::Expression *> exprList;
+    if (const auto *structExpr = inputExpression->to<IR::StructExpression>()) {
+        for (const auto *listElem : structExpr->components) {
+            auto subList = flattenComplexExpression(listElem->expression, flatValids);
+            exprList.insert(exprList.end(), subList.begin(), subList.end());
+        }
+        if (auto richStructExpr = structExpr->to<IR::HeaderExpression>()) {
```
vlstill, line 125: why not name it `headerExpr`?

backends/p4tools/common/core/abstract_execution_state.cpp
```diff
@@ -120,19 +182,54 @@ void AbstractExecutionState::initializeStructLike(const Target &target,
...
-        return tableType->table;
+void AbstractExecutionState::assignStructLike(const IR::StateVariable &left,
+                                              const IR::Expression *right) {
+    if (const auto *structExpr = right->to<IR::StructExpression>()) {
+        std::vector<IR::StateVariable> flatLeftValids;
+        std::vector<const IR::Expression *> flatRightValids;
+        auto flatTargetFields = getFlatFields(left, &flatLeftValids);
+        auto flatStructFields = flattenComplexExpression(structExpr, flatRightValids);
+        auto flatTargetSize = flatTargetFields.size();
+        auto flatStructSize = flatStructFields.size();
```
vlstill, line 193: nitpick: I don't see much reason for naming the .size() of a vector, which is calculated in O(1) and is not a very long expression

backends/p4tools/p4tools.def
```diff
@@ -241,3 +241,20 @@ public:
           concolicId(concolicId),
           concolicMethodName(methodName) {}
 }
+
+/// An extension of a struct expression. Also tracks validity as an expression.
+class HeaderExpression : StructExpression {
```
vlstill, line 246: I am wondering if this could clash with the need for something similar in the core IR (i.e. not just for p4tools). But I guess if that happens this can be moved or changed.

fruffy: Actually, the compiler already has a similar expression in the form of an [InvalidHeader](https://github.com/p4lang/p4c/blob/main/ir/expression.def#L492) and StructExpression (valid header). What we need, which is not yet captures is a HeaderExpression that can model validity symbolically.

I thought about this, but there does not seem to be a use case for this in the general compiler.

backends/p4tools/common/core/abstract_execution_state.cpp
fruffy: Part of the reason for this naming is just for readability since these variables are used multiple times.

backends/p4tools/common/core/abstract_execution_state.h
fruffy: Also would do this in a separate issue, fixing this for not just this function is a little more involved. 
#4266

backends/p4tools/p4tools.def
vlstill: OK, let's keep it this way. It is unlikely the compiler will need arbitrary expressions as validity bits (unless it starts doing SMT-based optimization :-D).

vlstill (approved):

PR #4232, opened 2023-11-09
Simplify jsonAssignment in ParserConverter and ActionConverter

antoninbas:
The inParser parameter is not actually used. I think it was introduced back when both converters shared the same implementation.


antoninbas:
@qobilidop FYI

jafingerhut (approved):
LGTM, as in it appears that these changes appear like they should not change any behavior of the code.

mihaibudiu:
Can't say this is fresh in my mind, but it looks reasonable.

PR #4233, opened 2023-11-09
Fixed ebpf_packetOffsetInBits field in parser and control_block C file

komaljai:
1. In the control block C file, the code is checking to see if the parsed packet is smaller than the packet that will be generated by the deparser. The problem is that ebpf_packetOffsetInBits is set by the parser C file, but is not being passed on to the control block C file. The solution is to make ebpf_packetOffsetInBits part of the eBPF map struct in header file, which is included in both parser and control block C file
2. Change permisson for default action.
3. Convert header fields to host endian in parser.c file


usha1830 (approved):
Looks good to me

komaljai:
I see the fedora and ptf tests are failing, and it is not related to this PR, Are others also facing the same issue?

usha1830 (approved):

PR #4237, opened 2023-11-09
Fix some problems and suggestions found by clang-tidy

vlstill:
All these are found by the clang-tidy static analyzer. At this point, I've been only running in on the headers as this is mainly a byproduct of linting a downstream tool. Some of these changes are in the readability category rather than possible bugs, but I think they are all worth applying.

### DIFF
--- a/lib/stringref.h
+++ b/lib/stringref.h
@@ -51,11 +51,14 @@ struct StringRef {
         len = 0;
     }
     StringRef(const StringRef &a) : p(a.p), len(a.len) {}
-    StringRef &operator=(const StringRef &a) {  // NOLINT(bugprone-unhandled-self-assignment)
+    // avoid clang-tidy complaining
+    // NOLINTBEGIN(bugprone-unhandled-self-assignment)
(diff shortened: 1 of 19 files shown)


lib/stringref.h
```diff
@@ -51,11 +51,14 @@ struct StringRef {
         len = 0;
     }
     StringRef(const StringRef &a) : p(a.p), len(a.len) {}
+    // avoid clang-tidy complaining
+    // NOLINTBEGIN(bugprone-unhandled-self-assignment)
```
fruffy, line 55:
```
    if(this == &a)
      return *this;
```
instead?

fruffy (approved):
Any chance we can make this a part of our CMake setup? Ideally, we can run clang-tidy checks as part of CI/ or just apply fix-its automatically. We can tweak the `.clang-tidy` file as needed.

lib/stringref.h
vlstill: This one is safe to be performed, I'm not sure if the if is worth having in such cases. If the if is there, one would have to wonder why it is needed.

PR #4238, opened 2023-11-10
Report offending node in the type-checker immutability check

vlstill:
Currently, the compiler would produce only a very vague error message if type checking runs in immutable mode and the IR tree is actually modified. This is extremely annoying to debug. Therefore, I've added the check after each node is processed and the check now reports the offending node.

I've also modified the `end_apply` function (which still prints the generic check, although it is probably redundant):
- calling parent `end_apply`
- the nodes should be comparend for exact pointer equality, not for internal semantic equality -- it is not OK if the transform produces new tree that is equivalent in the readOnly mode. It must keep the original tree.


vlstill:
The failed tests look like some problems with dependencies rather than anything related to this change :-/.

mihaibudiu (approved):

fruffy (approved):

PR #4239, opened 2023-11-10
[P4Testgen] Add EliminateInvalidHeaders midend pass

vlstill:
Previously, testgen would fail will errors like this (when #4238 is applied to make the error readable):

```
Internal error: In file: ./p4c/frontends/p4/typeChecking/typeChecker.cpp:172
Compiler Bug: build/init.p4(78): At this point in the compilation typechecking should not infer new types anymore, but it did: node (hdr_0_t){#}.isValid(); changed to 0;
        hdr.hdr_0 = (hdr_0_t){#};
                  ^
build/init.p4(78)
        hdr.hdr_0 = (hdr_0_t){#};
                  ^
```

### DIFF
--- a/backends/p4tools/common/compiler/midend.cpp
+++ b/backends/p4tools/common/compiler/midend.cpp
@@ -15,6 +15,7 @@
 #include "midend/convertEnums.h"
 #include "midend/convertErrors.h"
 #include "midend/copyStructures.h"
+#include "midend/eliminateInvalidHeaders.h"
 #include "midend/eliminateNewtype.h"
 #include "midend/eliminateSerEnums.h"
 #include "midend/eliminateSwitch.h"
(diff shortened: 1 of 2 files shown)


fruffy:
Could you add a test for this failure? Or fix existing tests to work with bmv2 etc

vlstill:
> Could you add a test for this failure? Or fix existing tests to work with bmv2 etc

@fruffy, can you give me pointers where the p4c testgen tests live? I've never interacted with them before except for running them with ctest.

mihaibudiu (approved):

fruffy:
> > Could you add a test for this failure? Or fix existing tests to work with bmv2 etc
> 
> @fruffy, can you give me pointers where the p4c testgen tests live? I've never interacted with them before except for running them with ctest.

P4Testgen will pick up all the programs in the `testdata` folder which include `v1model.p4`. So, if https://github.com/p4lang/p4c/blob/main/testdata/p4_16_samples/invalid-header.p4 is transformed into a v1model program, the test should automatically be included.

Alternatively, custom tests for BMv2 are placed in this folder:

https://github.com/p4lang/p4c/tree/main/backends/p4tools/modules/testgen/targets/bmv2/test/p4-programs

PR #4241, opened 2023-11-10
Update gRPC python version in install_fedora_deps.sh

antoninbas:
The fedora:latest container now comes with Python 3.12. Older gRPC versions do not support this Python version. See
https://github.com/grpc/grpc/issues/33063.

An alternative would be to pin the fedora container version, but then someone has to remember to update it periodically...

### DIFF
--- a/backends/bmv2/bmv2stf.py
+++ b/backends/bmv2/bmv2stf.py
@@ -748,7 +748,7 @@ class RunBMV2(object):
                     str(rand),
                 ]
                 + self.interfaceArgs()
-                + ["../" + self.jsonfile]
+                + [self.jsonfile]
             )
             if self.cmd_line_args:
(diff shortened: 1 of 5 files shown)


antoninbas:
This will unstuck pending PRs which have received approval

mihaibudiu (approved):

rst0git (approved):

jafingerhut (approved):
The changes look reasonable to me, but after commit 2 it appears that perhaps the BMv2 tests are failing for a different reason than before?

antoninbas:
> The changes look reasonable to me, but after commit 2 it appears that perhaps the BMv2 tests are failing for a different reason than before?

Yes, I'm on commit 3 now. I'm moving from issue to issue, all because of a more recent Python version. But hopefully, the 3rd issue was the last one.

PR #4244, opened 2023-11-13
Allow overriding of 'usage' method in the Options class

jhavrane:
This change allows to override default usage() method in Options.


fruffy:
Only thing I would suggest is to change the title to something like "Allow overriding of 'usage' method in the Options class.".

PR #4249, opened 2023-11-15
Added range check of error codes

dmatousek:
The PR adds named constants for min and max values of error, warning, and info message codes to be used in backends. Also, the `ErrorCatalog::add` method now expects a code type so that it can perform range check. E.g.:

`ErrorCatalog::getCatalog().add<ErrorMessage::MessageType::Info, INFO_USER_CODE>("user-code");`

If the range check fails, a bug is triggered.

In addition, a new info code INFO_PROGRESS is added.

### DIFF
--- a/lib/error_catalog.h
+++ b/lib/error_catalog.h
@@ -21,10 +21,12 @@ limitations under the License.
 #include <string>
 
 #include "cstring.h"
+#include "lib/error_message.h"
+#include "lib/exceptions.h"
+
+using MessageType = ErrorMessage::MessageType;
 
 /// enumerate supported errors
-/// Backends should extend this class with additional errors in the range 500-999 and
-/// warnings in the range 1500-2141.
 /// It is a class and not an enum class because in C++11 you can't extend an enum class
 class ErrorType {
  public:
@@ -49,8 +51,8 @@ class ErrorType {
     static const int ERR_MODEL;                  // something is wrong with the target model
     static const int ERR_RESERVED;               // Reserved for target use
 
-    // If we specialize for 1000 error types we're good!
-    static const int ERR_MAX_ERRORS;
+    static const int ERR_MIN_BACKEND;  // first allowed backend error code
+    static const int ERR_MAX;          // last allowed error code
 
     // -------- Warnings -----------
     // warnings as initially defined with a format string
@@ -78,13 +80,15 @@ class ErrorType {
     static const int WARN_DUPLICATE_PRIORITIES;     // two entries with the same priority
     static const int WARN_ENTRIES_OUT_OF_ORDER;     // entries with priorities out of order
 
-    static const int WARN_MAX_WARNINGS;
+    static const int WARN_MIN_BACKEND;  // first allowed backend warning code
+    static const int WARN_MAX;          // last allowed warning code
 
     // -------- Info messages -------------
     // info messages as initially defined with a format string
     static const int INFO_INFERRED;  // information inferred by compiler
 
-    static const int INFO_MAX_INFOS;
+    static const int INFO_MIN_BACKEND;  // first allowed backend info code
+    static const int INFO_MAX;          // last allowed info code
 };
 
 class ErrorCatalog {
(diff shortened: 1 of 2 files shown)


lib/error_catalog.h
```diff
@@ -97,10 +102,24 @@ class ErrorCatalog {
 
     /// add to the catalog
     /// returns false if the code already exists and forceReplace was not set to true
+    /// @param type      - error/warning/info message
     /// @param errorCode - integer value for the error/warning
     /// @param name      - name for the error. Used to enable/disable all errors of that type
     /// @param forceReplace - override an existing error type in the catalog
-    bool add(int errorCode, const char *name, bool forceReplace = false) {
+    bool add(MessageType type, int errorCode, const char *name, bool forceReplace = false) {
```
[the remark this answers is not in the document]
vlstill: It would require actually calling `add` in compile-time context. However map::emplace is not constexpr (not even in C++23). Or it would need to be a template, somewhat like this:

```cpp
template<MessageType type, int errCode>
bool add(const char *name, bool forceReplace = flase) {
   static_assert(type !=  MessageType::Error ||
            (errorCode >= ErrorType::ERR_MIN_BACKEND && errorCode < ErrorType::ERR_MAX),
            "Error code out of range");
   ...
}
```

That would work too (I believe). It would be called like this:

```cpp
errorCatalog.add<MessageType::Error, Err::ERR_FOO>("foo");
```

Which is not super-nice, but I'd say it is worth the added safety.

For this to work, the constants would need to be declared in the `.h` file. Probably as

```cpp
static constexpr int ERR_MIN_BACKEND = ...;
```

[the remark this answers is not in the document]
fruffy: I'll let you decide whether you consider it worth implementing. I think it'd be nice, since runtime `BUG`s are a persistence source of annoyance.

fruffy (approved):

lib/error_catalog.h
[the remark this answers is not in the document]
dmatousek: Thanks for your suggestions, they really make sense!

lib/error_catalog.h
```diff
@@ -97,10 +103,19 @@ class ErrorCatalog {
...
     /// add to the catalog
     /// returns false if the code already exists and forceReplace was not set to true
+    /// @param type      - error/warning/info message
     /// @param errorCode - integer value for the error/warning
     /// @param name      - name for the error. Used to enable/disable all errors of that type
     /// @param forceReplace - override an existing error type in the catalog
-    bool add(int errorCode, const char *name, bool forceReplace = false) {
+    template <MessageType type, int errorCode>
+    bool add(const char *name, bool forceReplace = false) {
+        static_assert(type != MessageType::Error ||
```
fruffy, line 112: Neat, does this work `errorcode` or should we make explicit that it is a `constexpr`? Same for all the error codes.

dmatousek: Sorry, what do you mean?

fruffy: All the integers are `const` not `constexpr`. Does this work with `static_assert`? Or does the compiler convert them automatically to compile time constants?

dmatousek: Yes, it works. The compilation fails with:
```
lib/error_catalog.h:112:50: error: static assertion failed
  112 |         static_assert(type != MessageType::Error ||
      |                       ~~~~~~~~~~~~~~~~~~~~~~~~~~~^~
  113 |             (errorCode >= ErrorType::ERR_MIN_BACKEND && errorCode <= ErrorType::ERR_MAX));
      |             ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
I believe that for an integral type `const` implies `consexpr`.

fruffy (approved):

dmatousek:
What do I have to do to be able to merge p4c PRs?

fruffy:
I think we need to add you as a committer. Working on that. I can merge it for you.

dmatousek:
Thank you!

fruffy:
There should be a person at Intel that can add you to p4lang, is that not the case anymore?

PR #4252, opened 2023-11-16
[P4Testgen] Change crash message

p-sawicki:
Changed messages emitted when Testgen crashes to differentiate them from the ones reported by the compiler. Example:
```
Internal error: In file: ./p4c/backends/p4tools/modules/testgen/lib/execution_state.cpp:301
Testgen Bug: Did not find exception handler for Exception::NoMatch.

Please submit a bug report with your code.
```

I think this will be useful for example for CI flows where first Testgen is used to generate test cases and then a compiler is used on the same input. Currently it's not obvious if Testgen crashes because the message contains `Compiler Bug`.

### DIFF
--- a/backends/p4tools/modules/testgen/main.cpp
+++ b/backends/p4tools/modules/testgen/main.cpp
@@ -10,7 +10,7 @@
 std::string updateErrorMsg(std::string errorMsg) {
     for (const std::string_view toReplace : {"Compiler", "compiler"}) {
         if (const auto pos = errorMsg.find(toReplace); pos != std::string::npos) {
-            errorMsg.replace(pos, toReplace.size(), "Testgen");
+            errorMsg.replace(pos, toReplace.size(), "P4Testgen");
             break;
         }
     }
--- a/backends/p4tools/modules/testgen/main.cpp
+++ b/backends/p4tools/modules/testgen/main.cpp
@@ -7,6 +7,17 @@
 
 #include "backends/p4tools/modules/testgen/testgen.h"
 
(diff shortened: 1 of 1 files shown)


backends/p4tools/modules/testgen/main.cpp
```diff
@@ -7,6 +7,17 @@
 
 #include "backends/p4tools/modules/testgen/testgen.h"
 
+std::string updateErrorMsg(std::string errorMsg) {
+    for (const std::string_view toReplace : {"Compiler", "compiler"}) {
+        if (const auto pos = errorMsg.find(toReplace); pos != std::string::npos) {
+            errorMsg.replace(pos, toReplace.size(), "Testgen");
```
fruffy, line 13:
```suggestion
            errorMsg.replace(pos, toReplace.size(), "P4Testgen");
```

fruffy (approved):
P4Testgen also has an exception class which we can customize
https://github.com/p4lang/p4c/blob/main/backends/p4tools/modules/testgen/lib/exceptions.h

We could replace BUG and CHECK_NULL etc with P4Testgen exceptions. 

However, your method also captures `BUGs` caused by utility functions in the compiler, so it should be more complete.

p-sawicki:
Thanks for reviewing, @fruffy. I don't have write access so I won't be able to merge. Can you do merge the PR for me or should I apply for write access somewhere?

fruffy:
I can merge it. Not sure how write access is managed these days.

@jnfoster @jafingerhut Do we have a concrete policy?

vlstill (approved):

PR #4253, opened 2023-11-17
Fix clang-tidy config file.

fruffy:
`AnalyzeTemporaryDtors` is deprecated and throws and error. 

`--std=c++17` does not have any effect, it seems. It only throws an error.


fruffy:
This will fix clangd while #4254 is blocked.

vlstill (approved):

PR #4256, opened 2023-11-20
Fixed checking of nested tuple fields

dmatousek:
According to the [spec](https://staging.p4.org/p4-spec/docs/P4-16-v1.2.4.html#sec-type-nesting), tuple fields should adhere to the same rules as struct fields.

### DIFF
--- a/frontends/p4/typeChecking/typeChecker.cpp
+++ b/frontends/p4/typeChecking/typeChecker.cpp
@@ -1478,20 +1478,6 @@ const IR::Node *TypeInference::postorder(IR::Type_List *type) {
     return type;
 }
 
-const IR::Node *TypeInference::postorder(IR::Type_Tuple *type) {
-    for (auto field : type->components) {
-        auto fieldType = getTypeType(field);
-        if (auto spec = fieldType->to<IR::Type_SpecializedCanonical>()) fieldType = spec->baseType;
-        if (fieldType->is<IR::IContainer>() || fieldType->is<IR::Type_ArchBlock>() ||
-            fieldType->is<IR::Type_Extern>()) {
-            typeError("%1%: not supported as a tuple field", field);
-            return type;
-        }
-    }
-    (void)setTypeType(type);
-    return type;
-}
-
 const IR::Node *TypeInference::postorder(IR::Type_P4List *type) {
     (void)setTypeType(type);
     return type;
@@ -1642,8 +1628,8 @@ const IR::Node *TypeInference::postorder(IR::Type_Stack *type) {
 /// Validate the fields of a struct type using the supplied checker.
 /// The checker returns "false" when a field is invalid.
 /// Return true on success
-bool TypeInference::validateFields(const IR::Type *type,
-                                   std::function<bool(const IR::Type *)> checker) const {
+bool TypeInference::validateStructLikeFields(const IR::Type *type,
+                                             std::function<bool(const IR::Type *)> checker) const {
     if (type == nullptr) return false;
     BUG_CHECK(type->is<IR::Type_StructLike>(), "%1%; expected a Struct-like", type);
     auto strct = type->to<IR::Type_StructLike>();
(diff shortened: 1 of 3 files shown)


testdata/p4_16_samples/issue3238.p4
```diff
@@ -1,17 +1,14 @@
-bit func(bit t1)
+bit func(int t1)
```
mihaibudiu, line 1: why modify this test?

testdata/p4_16_samples/list3.p4
```diff
@@ -1,6 +1,6 @@
 #include <core.p4>
 
-typedef int EntryPriority_t;
+typedef int<8> EntryPriority_t;
```
mihaibudiu, line 3: why modify this test?
These are all regression tests, if you modify them they no longer serve their purpose.
You should add new tests if you want to check new properties.
If this test becomes invalid there should be a very good reason for invalidating it.

dmatousek: This test is incorrect from the perspective of the (current) specification (a tuple cannot hold `int`). However, I found the test useful otherwise so I wanted to keep it with a minimal change.

testdata/p4_16_samples/issue3238.p4
dmatousek: This test is incorrect from the perspective of the (current) specification (a tuple cannot hold `int`). I tried to modify the test so that it tests the original intended behavior (implicit cast `int` -> `bit`, #3469).

testdata/p4_16_samples/list3.p4
mihaibudiu: In this case perhaps the spec is wrong. A tuple should be able to hold an int if it is a compile-time constant.

testdata/p4_16_errors_outputs/issue3750-1.p4-stderr
```diff
@@ -1,6 +1,14 @@
...
-             ^^
-issue3750-1.p4(14): [--Werror=type-error] error: E<bit<32>>: not supported as a tuple field
-void g(tuple<E<bit<32>>> t)
-             ^^^^^^^^^^
+       ^^^^^^^^^
+issue3750-1.p4(1): [--Werror=type-error] error: Field '0' of 'tuple<ct>' cannot have type 'ct'
+extern ct {
+       ^^
+In file: /home/runner/work/p4c/p4c/frontends/p4/methodInstance.cpp:86
+issue3750-1.p4(11): unexpected expression t[0]; resolving method instance
```
dmatousek, line 11: Just realized that this is the result of the `BUG` call. It has to be fixed.

testdata/p4_16_errors_outputs/issue3750.p4-stderr
```diff
@@ -1,3 +1,14 @@
...
+        ^^
+issue3750.p4(2): [--Werror=type-error] error: Error while analyzing tuple<control ct>
 void f(tuple<ct> t)
-             ^^
+       ^^^^^^^^^
+issue3750.p4(1): [--Werror=type-error] error: Field '0' of 'tuple<control ct>' cannot have type 'control ct'
+control ct();
+        ^^
+In file: /home/runner/work/p4c/p4c/frontends/p4/methodInstance.cpp:86
+issue3750.p4(4): unexpected expression t[0]; resolving method instance
```
dmatousek, line 11: The same as above.

testdata/p4_16_samples/list3.p4
dmatousek: I didn't get it. Could you elaborate a bit more (ideally wrt to the [spec](https://staging.p4.org/p4-spec/docs/P4-16-v1.2.4.html#sec-type-nesting)) please?

mihaibudiu: It seems useful to allow int as a type for a tuple element.
If the spec says that this is not possible, perhaps the right thing is to change the spec.

PR #4258, opened 2023-11-20
More code improvements based on static analysis

vlstill:


backends/dpdk/dpdkArch.cpp
```diff
@@ -2640,7 +2640,7 @@ void CollectAddOnMissTable::postorder(const IR::P4Table *t) {
         ::error(ErrorType::ERR_UNEXPECTED,
                 "%1%: add_on_miss property is defined, "
                 "but default_action not specificed for table %2%",
-                default_action, t->name);
+                add_on_miss, t->name);
```
vlstill, line 2643: Note that this is guarded by `if (use_add_on_miss && default_action == nullptr)` so using `default_action` in the error makes no sense.

fruffy (approved):

PR #4262, opened 2023-11-26
Fix some problems with the Protobuf CMake file

fruffy:
A P4C extension I was working on needed gRPC. However, the way we currently build Protobuf does not combine well with the gRPC FetchContent module. This PR adds a couple fixes for the Protobuf FetchContent script. 

 - Build protoc from scratch.
 - Ensure that we always build with position-independent code locally but not across the entire compiler.
 - Update how we set variables, align it to be the same as how the Protobuf CMake sets it.


CMakeLists.txt
```diff
@@ -128,7 +128,7 @@ if (BUILD_STATIC_RELEASE)
   # Set the static variable
   set(P4C_STATIC_BUILD STATIC)
   # Do not bring in dynamic libstdcc and libgcc
-  set(CMAKE_EXE_LINKER_FLAGS "-static -static-libgcc -static-libstdc++ -Wl,-z,muldefs")
+  set(CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} -static-libgcc -static-libstdc++ -Wl,-z,muldefs")
```
fruffy, line 131: Linking protoc when using -static causes a series of problems. It seems to be a bad idea to use -static in general, so let's just fix this call.

backends/dpdk/CMakeLists.txt
```diff
@@ -28,7 +28,7 @@ add_custom_target (dpdk_runtime_dir
 # placed in the build directory inside `control-plane` directory
 add_custom_command(
     OUTPUT ${DPDK_P4RUNTIME_INFO_GEN_SRCS} ${DPDK_P4RUNTIME_INFO_GEN_HDRS}
-  COMMAND ${PROTOBUF_PROTOC_EXECUTABLE}
+  COMMAND ${Protobuf_PROTOC_EXECUTABLE}
```
fruffy, line 31: The canonical Protobuf variable prefix is now `Protobuf_` instead of `PROTOBUF_`.

https://cmake.org/cmake/help/latest/module/FindProtobuf.html

cmake/Protobuf.cmake
```diff
@@ -13,30 +13,75 @@ macro(p4c_obtain_protobuf)
...
     # Print out download state while setting up Protobuf.
     set(FETCHCONTENT_QUIET_PREV ${FETCHCONTENT_QUIET})
     set(FETCHCONTENT_QUIET OFF)
+    # Build Protobuf with position-independent code.
+    set(CMAKE_POSITION_INDEPENDENT_CODE_PREV ${CMAKE_POSITION_INDEPENDENT_CODE})
+    set(CMAKE_POSITION_INDEPENDENT_CODE ON)
+
     set(protobuf_BUILD_TESTS OFF CACHE BOOL "Build tests.")
-    set(protobuf_BUILD_PROTOC_BINARIES OFF CACHE BOOL "Build libprotoc and protoc compiler.")
+    set(protobuf_BUILD_PROTOC_BINARIES ON CACHE BOOL "Build libprotoc and protoc compiler.")
```
fruffy, line 67: We build protoc from scratch, because can interact badly with protoc plugins such as `grpc_cpp_plugin`.

CMakeLists.txt
vlstill: But this option is under `BUILD_STATIC_RELEASE`. It seems wrong to me to have this build non-static binary. If `BUILD_STATIC_RELEASE` is broken, we should just remove it. Arguably it is broken on most linux distros as glibc should not be linked statically (and often there is no corresponding `.a`). Therefore, `BUILD_STATIC_RELASE` would be useful only on distros like Alpine (which uses MUSL instead of GLIBC).

I suggest you just don't enable `BUILD_STATIC_RELEASE`. If you have need for semi-static (i.e. glibc dynamic, rest static) build, that should be handled by a new CMake option in a new PR.

cmake/Protobuf.cmake
```diff
@@ -13,30 +13,75 @@ macro(p4c_obtain_protobuf)
...
+        PROTOBUF_LITE_LIBRARY
+        PROTOBUF_LITE_LIBRARY_DEBUG
+        )
+        if (DEFINED ${UPPER})
+            string(REPLACE "PROTOBUF_" "Protobuf_" Camel ${UPPER})
+            if (NOT DEFINED ${Camel})
+                set(${Camel} ${${UPPER}})
+            endif()
+        endif()
+    endforeach()
```
vlstill, line 39: Why is this needed? Looking at https://cmake.org/cmake/help/v3.6/module/FindProtobuf.html it seems the "new" versions are defined since CMake 3.6, we require at least 3.16 for p4c. Or is this coming from some other source? How did it work before?

cmake/Protobuf.cmake
```diff
@@ -13,30 +13,75 @@ macro(p4c_obtain_protobuf)
...
+                                                         "arm64"
+    )
+      set(protobuf_ARCH "aarch_64")
+    elseif(CMAKE_HOST_SYSTEM_PROCESSOR STREQUAL "x86_64" OR CMAKE_HOST_SYSTEM_PROCESSOR STREQUAL
+                                                            "amd64"
+    )
+      set(protobuf_ARCH "x86_64")
+    else()
+      message(FATAL_ERROR "Unsupported host architecture `${CMAKE_HOST_SYSTEM_PROCESSOR}`")
+    endif()
```
vlstill, line 83: Seems to be unused now.

control-plane/CMakeLists.txt
```diff
@@ -51,8 +51,8 @@ set (P4RUNTIME_GEN_PYTHON "control-plane")
 add_custom_target (mkP4configdir
   ${CMAKE_COMMAND} -E make_directory ${CMAKE_BINARY_DIR}/p4/config)
 add_custom_command(OUTPUT ${P4RUNTIME_GEN_SRCS} ${P4RUNTIME_GEN_HDRS}
-  COMMAND ${PROTOBUF_PROTOC_EXECUTABLE}
-          -I ${P4RUNTIME_STD_DIR} -I${CMAKE_CURRENT_SOURCE_DIR}
+  COMMAND ${Protobuf_PROTOC_EXECUTABLE}
+          -I${P4RUNTIME_STD_DIR} -I${CMAKE_CURRENT_SOURCE_DIR} -I${Protobuf_INCLUDE_DIR}
```
vlstill, line 55: Why not `PROTOC_BINARY`?

cmake/Protobuf.cmake
fruffy: We are now using the CamelCase version of variables. It seemed prudent to add this conversion to be backwards compatible. But I happily remove it if we can assume that builds with users' existing systems.

control-plane/CMakeLists.txt
fruffy: Using the generator expression is fine for commands in the build stage but not when writing to a file in the configure stage. There are some scripts (primarily in P4Testgen), which require access to the binary in the configure stage to write a file. I wish I had a good way to resolve this.

CMakeLists.txt
fruffy: We a CI test that causes problems, so it is a little tricky to make this work. `-static` not only causes problems for Protobuf but also the [libgc](https://github.com/p4lang/p4c/pull/3930) build.

Let me experiment a bit with this. I am able to build locally with clang+lld. So maybe that is one way to ensure we can at least generate some form of static release. 

With bf-p4c, we did not even use the `-static` iirc.

vlstill: Yep, that is why I suggested actually removing static as a possibility. If we add semi-static instead, it would not be hard if anyone needed full-static to set it from it (basically just setting `CMAKE_EXE_LINKER_FLAGS` to `-static`). We probably have no way of knowing if static build is used by anyone (unless they show up here).

cmake/Protobuf.cmake
vlstill: It seems the `FindPackge` module for Protobuf does define the "old-style" names itself (which is why the used until now). So I don't think we should not it ourselves.

And since we don't need to keep compatibility with CMake < 3.6 the worst thing that could happen is that someone will have to find this change and update their cmake when updating p4c.

CMakeLists.txt
fruffy: So what is the best course of action here? Throw a warning when `BUILD_STATIC_RELEASE` is enabled? Or rename BUILD_STATIC_RELEASE?

control-plane/CMakeLists.txt
vlstill: I see. I've just noticed this is not consistent among all the usages, but I guess it actually makes sense to use the "canonical" way (provided by FindPackage) wherever possible.

CMakeLists.txt
vlstill: I would suggest renaming it. This seems the cleanest to me and least risk of breaking someones workflow. Maybe something like `BUILD_SEMI_STATIC_RELEASE`, although it does not explain what it means to be semi-static. But I cannot think of a short way to express "static except for basic OS dependencies like glibc".

.github/workflows/ci-static-build-test.yml
```diff
@@ -37,3 +37,12 @@ jobs:
...
         sudo -E tools/ci-build.sh
+        # Disabling checks until we find a better way to build a fully static binary.
+        ldd ./build/p4c-bm2-psa 2>&1 # | grep -E -qi "(not a dynamic executable)|(statically linked)"
+        ldd ./build/p4c-bm2-ss 2>&1 # | grep -E -qi "(not a dynamic executable)|(statically linked)"
+        ldd ./build/p4c-dpdk 2>&1 # | grep -E -qi "(not a dynamic executable)|(statically linked)"
+        ldd ./build/p4c-ebpf 2>&1 # | grep -E -qi "(not a dynamic executable)|(statically linked)"
+        ldd ./build/p4c-pna-p4tc 2>&1 # | grep -E -qi "(not a dynamic executable)|(statically linked)"
+        ldd ./build/p4c-ubpf 2>&1 # | grep -E -qi "(not a dynamic executable)|(statically linked)"
+        ldd ./build/p4test 2>&1 # | grep -E -qi "(not a dynamic executable)|(statically linked)"
+        ldd ./build/p4testgen 2>&1 # | grep -E -qi "(not a dynamic executable)|(statically linked)"
```
vlstill, line 48: We could check that there is no boost, no libstdc++, etc. It would be a partial check only, but it would be at least something.

vlstill (commented):
I think the changes now make sense. But I strongly suggest the renaming of static build is a separate PR. It is not directly related to Protobuf and it would be good if it is more visible in the commit messages for the benefit of downstream tools etc.

vlstill (approved):

PR #4267, opened 2023-11-29
[P4Testgen] Move arch spec to ProgramInfo to allow it to depend on P4 program

vlstill:
Before this change, the `ArchSpec` was defined in each target. However, this approach is too restrictive for the targets that allow variability in the exact signature of the pipeline. In these cases, it is necessary to have the ArchSpec depend not only on the architecture itself, but also on the actual P4 code (on the instances used in the `main` package). Therefore, I propose moving the ArchSpec definition to program info.

This should not pose any problem as:
* the places that use it either have access to ProgramInfo
* or it is accessed when program info is being created (i.e. `TestgenTarget::initProgramImpl`) 
  * For existing targets I make the arch spec also available as static const value member of the `*ProgramInfo` class so it can be accessed -- this also demostrates clearly that it is not dependent on the instance of Program Info.
  * For targets where ArchSpec is dependend on the P4 program presumably it can be built first and then it will be passed to the program info.

### DIFF
--- a/backends/p4tools/modules/testgen/core/program_info.h
+++ b/backends/p4tools/modules/testgen/core/program_info.h
@@ -55,6 +55,13 @@ class ProgramInfo : public ICastable {
     /// The generated dcg.
     const NodesCallGraph *dcg = nullptr;
 
+    /// A vector that maps the architecture parameters of each pipe to the corresponding
+    /// global architecture variables. For example, this map specifies which parameter of each pipe
+    /// refers to the input header.
+    // The arch map needs to be public to be subclassed.
+    /// @returns a reference to the architecture map defined in this target
+    [[nodiscard]] virtual const ArchSpec &getArchSpec() const = 0;
+
     /// @returns the series of nodes that has been computed by this particular target.
     [[nodiscard]] const std::vector<Continuation::Command> *getPipelineSequence() const;
 
--- a/backends/p4tools/modules/testgen/core/small_step/extern_stepper.cpp
+++ b/backends/p4tools/modules/testgen/core/small_step/extern_stepper.cpp
@@ -292,10 +292,10 @@ void ExprStepper::evalInternalExternMethodCall(const IR::MethodCallExpression *c
                 const ExecutionState &state, SmallStepEvaluator::Result &result) {
              const auto *blockRef = args->at(0)->expression->checkedTo<IR::StringLiteral>();
              const auto *block = state.findDecl(new IR::Path(blockRef->value));
-             const auto *archSpec = TestgenTarget::getArchSpec();
              auto blockName = block->getName().name;
              auto &nextState = state.clone();
              auto canonicalName = getProgramInfo().getCanonicalBlockName(blockName);
+             const auto &archSpec = getProgramInfo().getArchSpec();
              const auto *blockApply = block->to<IR::IApply>();
              CHECK_NULL(blockApply);
              const auto *blockParams = blockApply->getApplyParameters();
@@ -306,7 +306,7 @@ void ExprStepper::evalInternalExternMethodCall(const IR::MethodCallExpression *c
              nextState.setProperty("inUndefinedState", false);
              for (size_t paramIdx = 0; paramIdx < blockParams->size(); ++paramIdx) {
                  const auto *internalParam = blockParams->getParameter(paramIdx);
-                 auto externalParamName = archSpec->getParamName(canonicalName, paramIdx);
+                 auto externalParamName = archSpec.getParamName(canonicalName, paramIdx);
                  nextState.copyIn(TestgenTarget::get(), internalParam, externalParamName);
              }
              nextState.setProperty("inUndefinedState", currentTaint);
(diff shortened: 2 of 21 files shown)


fruffy:
Conceptually this is problematic. The arch spec is meant to be a static representation of the target we are generating tests for, independent of the input program. Maybe there is an other way to solve this? What problem is this trying to address?

For reference, ideally we would like to implement this proposal:
https://github.com/jafingerhut/p4-guide/tree/master/specifying-p4-architectures#goals-of-a-new-language-for-specifying-p4-architectures

Which defines the structure of the architecture in pure P4 and is associated with one particular target.

vlstill:
> Conceptually this is problematic. The arch spec is meant to be a static representation of the target we are generating tests for, independent of the input program.

I can see why that would be conceptually simpler, but that does limit the target quite significantly. Also, the ArchSpec used quite lightly, there is basically one place in the testgen core that uses the ArchSpec at all (`copy_in` implementation).

> Maybe there is an other way to solve this? What problem is this trying to address?

I have an architecture which has some parameters open for the implementer to choose. For example, the number and direction of metadata parameters. I don't see a way to represent this using `ArchSpec` so the only alternative for generalization would be to actually replace the `copy_in` implementation (and risk breakages if more parts of testgen start using ArchSpec).
 
> For reference, ideally we would like to implement this proposal: https://github.com/jafingerhut/p4-guide/tree/master/specifying-p4-architectures#goals-of-a-new-language-for-specifying-p4-architectures
> 
> Which defines the structure of the architecture in pure P4 and is associated with one particular target.

That does sound interesting from the specification/verification interface standpoint. However, it also seems like a big undertaking and therefore does not help with my problem.

fruffy:
> I can see why that would be conceptually simpler, but that does limit the target quite significantly. Also, the ArchSpec used quite lightly, there is basically one place in the testgen core that uses the ArchSpec at all (`copy_in` implementation).

Part of the reason I am cautious here is that the Tofino architecture is more complex (multi-pipe, ghost thread, etc) than what is being tested in open-source. I can not verify that this approach will work seamlessly. @pkotikal could you check whether we can port the Tofino back end to these changes?

>I have an architecture which has some parameters open for the implementer to choose. For example, the number and direction of metadata parameters. I don't see a way to represent this using ArchSpec so the only alternative for generalization would be to actually replace the copy_in implementation (and risk breakages if more parts of testgen start using ArchSpec).

We do have support of optional parameters if they are at the end. Do they have canonical names or a particular mapping across pipes?

I have been meaning to simplify and merge `copy-in `into the P4Parser/P4Control visitor functions and make copy-out a first-class primitive. How would you extend it?

vlstill:
> > I can see why that would be conceptually simpler, but that does limit the target quite significantly. Also, the ArchSpec used quite lightly, there is basically one place in the testgen core that uses the ArchSpec at all (`copy_in` implementation).
> 
> Part of the reason I am cautious here is that the Tofino architecture is more complex (multi-pipe, ghost thread, etc) than what is being tested in open-source. I can not verify that this approach will work seamlessly. @pkotikal could you check whether we can port the Tofino back end to these changes?

I understand. I think my change should only allow more to be done. At the same time, I don't see anything in the testgen that would allow existence of multiple pipes expressed in the `ArchSpec`, that would probably require a different generalization.

> 
> > I have an architecture which has some parameters open for the implementer to choose. For example, the number and direction of metadata parameters. I don't see a way to represent this using ArchSpec so the only alternative for generalization would be to actually replace the copy_in implementation (and risk breakages if more parts of testgen start using ArchSpec).
> 
> We do have support of optional parameters if they are at the end. Do they have canonical names or a particular mapping across pipes?

That is not enough for my use case.

> 
> I have been meaning to simplify and merge `copy-in `into the P4Parser/P4Control visitor functions and make copy-out a first-class primitive. How would you extend it?

I would need to change the source of the `externalParamName` mapping which is currently coming from `ArchSpec`.

If they are first-class in the sense they would work based on the actual arguments in the P4 program, that would not be a problem. But I guess there would still need to be something binding the arguments in one block to the appropriate arguments of the next block and that I think requires some source of knowledge that is not directly described in P4. It is a pity this binding exists completely outside the P4 language.

vlstill:
(the force push is only a rebase to the latest main)

fruffy:
> I don't see anything in the testgen that would allow existence of multiple pipes expressed in the `ArchSpec`

The mapping in Tofino is rather complicated, but funnily enough the arch spec is expressive enough for this to work. All the plumbing logic is implemented in ProgramInfo and Target.

Recently, the lines between what goes into `target.cpp`, `bmv2/psa/tofino/etc.cpp`, and `program_info.cpp` have become quite blurry.  It makes the code a bit harder to understand for newcomers, unfortunately. 

Let's wait for Prathima to report back but we could migrate the arch spec to the program info. It makes sense as this information is derived from the input program.

vlstill:
@pkotikal, is it OK to merge this? It had already collected a bunch of conflicts and I have the work on my side that waits for this done.

I'll resolve the conflict once I have a green light on the "conceptual" merit of this move.

fruffy (approved):
Approving this.

At some point it might make sense to get rid of the target object entirely since ProgramInfo covers most of its use cases.

pkotikal:
> @pkotikal, is it OK to merge this? It had already collected a bunch of conflicts and I have the work on my side that waits for this done.
> 
> I'll resolve the conflict once I have a green light on the "conceptual" merit of this move.

Yes, please merge. Thanks!

PR #4268, opened 2023-11-29
[P4Testgen] Move argumentsToTypeDeclarations out of TestgenTarget

vlstill:
The utility function `TestgenTarget::argumentsToTypeDeclarations` can be also useful outside of the `TestgenTarget`. This makes it available. I did not see a good place to put it, therefore I've create a new file for it with intention it could be used for other utility functions.


fruffy:
Could you move this to `common/lib/util.h` instead?

vlstill:
I've tried that originally. However this depends on a static method from programInfo. Maybe that one could be extracted too, but otherwise this is testgen-specific.

vlstill:
yep. I think moving the helpers now makes sense

PR #4275, opened 2023-12-05
[P4Testgen] Fix behavior of coverage on edge cases with no nodes to cover

vlstill:
This PR changes behavior of testgen on various edge cases with regard to node coverage:

* If there are no coverable nodes, then set coverage to 100 %.
  * This could be a bit controversial, but I'd say it makes sense to have an empty program be considered fully covered. It does get a bit more tricky if `--track-coverage TABLE_ENTRIES` is enabled (without `STATEMENTS`) and there are no tables.
* the `--assert-min-coverage` can no longer be enabled unless `--track-coverage` is enabled as it does not make sense without anything to cover.
* The test "banner" is now different for the case the coverage is disabled and for the case coverage is enabled and there are no coverable nodes:
  * `============ Test 0 ============` (coverage disabled)
  * `============ Test 0: No coverable nodes ============`

### DIFF
--- a/backends/p4tools/modules/testgen/lib/test_backend.cpp
+++ b/backends/p4tools/modules/testgen/lib/test_backend.cpp
@@ -127,7 +127,8 @@ bool TestBackEnd::run(const FinalState &state) {
         } else if (coverableNodes.empty()) {
             printFeature("test_info", 4,
                          "============ Test %1%: No coverable nodes ============", testCount);
-            coverage = 1;  // all 0 nodes covered
+            // All 0 nodes covered.
+            coverage = 1.0;
         } else {
             coverage =
                 static_cast<float>(visitedNodes.size()) / static_cast<float>(coverableNodes.size());
--- a/backends/p4tools/modules/testgen/lib/test_backend.cpp
+++ b/backends/p4tools/modules/testgen/lib/test_backend.cpp
@@ -122,9 +122,12 @@ bool TestBackEnd::run(const FinalState &state) {
         }
 
         const P4::Coverage::CoverageSet &visitedNodes = symbex.getVisitedNodes();
-        if (coverableNodes.empty()) {
-            printFeature("test_info", 4, "============ Test %1%: No nodes ============", testCount);
-            coverage = 1; // all 0 nodes covered
+        if (!testgenOptions.coverageOptions.coverageEnabled()) {
+            printFeature("test_info", 4, "============ Test %1% ============", testCount);
+        } else if (coverableNodes.empty()) {
+            printFeature("test_info", 4,
+                         "============ Test %1%: No coverable nodes ============", testCount);
+            coverage = 1;  // all 0 nodes covered
         } else {
             coverage =
                 static_cast<float>(visitedNodes.size()) / static_cast<float>(coverableNodes.size());
--- a/backends/p4tools/modules/testgen/lib/test_backend.cpp
+++ b/backends/p4tools/modules/testgen/lib/test_backend.cpp
@@ -122,7 +122,7 @@ bool TestBackEnd::run(const FinalState &state) {
         }
 
         const P4::Coverage::CoverageSet &visitedNodes = symbex.getVisitedNodes();
-        if (!testgenOptions.coverageOptions.coverageEnabled()) {
+        if (!testgenOptions.hasCoverageTracking) {
             printFeature("test_info", 4, "============ Test %1% ============", testCount);
         } else if (coverableNodes.empty()) {
             printFeature("test_info", 4,
--- a/backends/p4tools/modules/testgen/lib/test_backend.cpp
+++ b/backends/p4tools/modules/testgen/lib/test_backend.cpp
@@ -123,7 +123,8 @@ bool TestBackEnd::run(const FinalState &state) {
 
         const P4::Coverage::CoverageSet &visitedNodes = symbex.getVisitedNodes();
         if (coverableNodes.empty()) {
-            printFeature("test_info", 4, "============ Test %1% ============", testCount);
+            printFeature("test_info", 4, "============ Test %1%: No nodes ============", testCount);
+            coverage = 1; // all 0 nodes covered
         } else {
             coverage =
                 static_cast<float>(visitedNodes.size()) / static_cast<float>(coverableNodes.size());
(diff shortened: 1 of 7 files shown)


midend/coverage.h
```diff
@@ -29,6 +29,10 @@ struct CoverageOptions {
     bool coverTableEntries = false;
     /// Skip tests which do not increase coverage.
     bool onlyCoveringTests = false;
+
+    /// @returns true if and only if at least one of the coverage-collection options (cover*) is
+    /// enabled
+    bool coverageEnabled() const { return coverStatements || coverTableEntries; }
```
vlstill, line 35: I've opted to add this getter to make it more likely this will be properly updated when new coverage types are added.

backends/p4tools/modules/testgen/testgen.cpp
```diff
@@ -115,6 +115,14 @@ int Testgen::mainImpl(const IR::P4Program *program) {
     if (seed) {
         printFeature("test_info", 4, "============ Program seed %1% =============\n", *seed);
     }
+    if (testgenOptions.minCoverage > 0 && !testgenOptions.coverageOptions.coverageEnabled()) {
```
fruffy, line 118: This is something I would catch in the options early, if possible.

backends/p4tools/modules/testgen/lib/test_backend.cpp
```diff
@@ -123,8 +123,12 @@ bool TestBackEnd::run(const FinalState &state) {
         }
 
         const P4::Coverage::CoverageSet &visitedNodes = symbex.getVisitedNodes();
-        if (coverableNodes.empty()) {
+        if (!testgenOptions.coverageOptions.coverageEnabled()) {
             printFeature("test_info", 4, "============ Test %1% ============", testCount);
+        } else if (coverableNodes.empty()) {
+            printFeature("test_info", 4,
+                         "============ Test %1%: No coverable nodes ============", testCount);
+            coverage = 1;  // all 0 nodes covered
```
fruffy, line 131:
```suggestion
            // All 0 nodes covered.
            coverage = 1.0;
```

midend/coverage.h
fruffy: Instead, I would add an option to `CoverageOptions` that sets this value true, whenever one of the values is set.

fruffy (changes requested):

midend/coverage.h
vlstill: I am not sure I understand what you mean. Do you suggest to have a derived boolean argument in the struct? That would seem like needless duplication of information at risk of inconsistency to me.

fruffy: Yes, set the value here: https://github.com/p4lang/p4c/blob/main/backends/p4tools/modules/testgen/options.cpp#L256

You can easily precompute the boolean value here. It will also automatically pick up any new coverage option that might be added. This way you do not need to hard-code all the possible coverage options in the helper.

backends/p4tools/modules/testgen/testgen.cpp
fruffy: Options should be processed as early as possible to get immediate feedback. Otherwise, you run your tests for a couple of minutes then get this error. 

We typically validate options already for each lambda we add. For example, here you could throw an error if you want to enable minCoverage without previously enable coverage in some way. This way you also do not require an extra validate call. This requires ordering, but that seems fine to me. This is a limitation of the way our options work.

vlstill: I don't like requiring one to be set before another, that is why I did not want to put the validation into the lambdas.

midend/coverage.h
vlstill: I don't like that all that much. The `CoverageOptions` struct is not strictly bound to be filled by options only, it is not even defined in the headers that process options. Therefore, having a derived attribute adds an invariant to the class that needs to hold when the class values are modified/set. Having the getter needs an invariant too (that it ORs all the `cover*` values), but this one is at least localized to the class itself.

fruffy: You could also add this flag to the testgen options class instead of the options structure.

vlstill: OK, let's do it that way. Although I still think having the member function did make sense.

backends/p4tools/modules/testgen/options.cpp
```diff
@@ -382,4 +383,15 @@ TestgenOptions::TestgenOptions()
...
 }
 
+void TestgenOptions::validate() const {
+    if (minCoverage > 0 && !hasCoverageTracking) {
+        ::error(
+            ErrorType::ERR_INVALID,
+            "It is not allowed to have --assert-min-coverage set to non-zero without a coverage "
+            "tracking enabled with --track-coverage option. Without coverage tracking, the "
+            "--assert-min-coverage is meaningless.");
+        FATAL_ERROR("Incompatible options given.");
```
fruffy, line 393: I approved but I do not think this is necessary. We will exist if we encounter an error during processing. If not, that is a bug.

vlstill:
Rebased + using the same name as is used in the `ParserOptions` (predecessor of `CompilerOptions`) for this purpose (it is not a base class, so the defintions are independent).

backends/p4tools/modules/testgen/options.cpp
fruffy: Hmm that is a little odd, we have been using `::error` just fine in the options classes and the compiler context is set up before the validate call.

The entire way options are handled in P4Testgen is a little funky (there are a bunch of hacks to set up context). So there  might be something awry there.

fruffy: Maybe the exception thrown with FATAL_ERROR causes the problem? We have not tested any exception handling that early.

vlstill: I think the handler for errors in the option parsing works differently, it also prints the usage if the argument parsing fails. What I was testing (and what produced this weird error) was just having the `error` there and no `FATAL_ERROR`, so there were no exceptions thrown.

fruffy: Did this get resolved?

vlstill: Unless we change error handling during process, the FATAL_ERROR is the best bet to cleanly end the processing (in the compiler, there are only warnings, so this does not matter).

fruffy: I looked into this, basically we need to do this 
```
if (::errorCount() > 0) {
    return std::nullopt;
}
``` 
after the validate call to not return a broken compile context. We could also return  true/false in validate.

vlstill: Made it return. `::errorCount()` is not available before the context is set (it could be extracted from the context, but this seems cleaner to me).

PR #4276, opened 2023-12-07
[P4Testgen] Add support for @format annotations in P4 programs for protobuf-ir tests

### LINKED ISSUE #4273: [P4Testgen] Respect SAI annotations when generating Protobuf tests
Exact matches can be `hex_str`, `ipv4`, `ipv6`, `mac`, or `str`, and PDPI currently enforces that you use the format that is specified in the program using annotations of the form `@format(IPV4_ADDRESS)` etc (default is hex). So we would need to either branch on the format here (which you can obtain easily from the IrP4Info). Or we would need to add a PDPI post-processing step that corrects the format, or perhaps extend PDPI with an option to be more lenient.

_Originally posted by @smolkaj in https://github.com/p4lang/p4c/pull/4221#discussion_r1383887618_

fruffy-g:
P4 SAI annotations such as `@format` dictate a specific format, which must be respected by P4Testgen when generating Protobuf PDPI tests. Hence, we add a couple utility functions to the Protobuf PDPI test back end to produce the right format in the case that a key has such an annotation. 

Fixes #4273.

### DIFF
--- a/backends/p4tools/common/lib/format_int.cpp
+++ b/backends/p4tools/common/lib/format_int.cpp
@@ -269,7 +269,8 @@ std::string insertHexSeparators(const std::string &dataStr) {
     return insertSeparators(dataStr, "\\x", 2, false);
 }
 
-std::vector<uint8_t> convertBigIntToBytes(const big_int &dataInt, int targetWidthBits) {
+std::vector<uint8_t> convertBigIntToBytes(const big_int &dataInt, int targetWidthBits,
+                                          bool padLeft) {
     /// Chunk size is 8 bits, i.e., a byte.
     constexpr uint8_t chunkSize = 8U;
 
@@ -282,7 +283,11 @@ std::vector<uint8_t> convertBigIntToBytes(const big_int &dataInt, int targetWidt
     auto diff = targetWidthBytes - bytes.size();
     if (targetWidthBytes > bytes.size() && diff > 0UL) {
         for (size_t i = 0; i < diff; ++i) {
-            bytes.push_back(0);
+            if (padLeft) {
+                bytes.insert(bytes.begin(), 0);
+            } else {
+                bytes.push_back(0);
+            }
         }
     }
     return bytes;
--- a/backends/p4tools/common/lib/format_int.cpp
+++ b/backends/p4tools/common/lib/format_int.cpp
@@ -17,12 +17,6 @@
 
 namespace P4Tools {
 
-static constexpr uint8_t IPV4_BYTE_SIZE = 4U;
-static constexpr uint8_t IPV6_BYTE_SIZE = 16U;
-static constexpr uint8_t MAC_BYTE_SIZE = 6U;
-/// Chunk size is 8 bits, i.e., a byte.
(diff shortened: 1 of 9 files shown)


backends/p4tools/common/lib/format_int.cpp
```diff
@@ -16,6 +17,12 @@
 
 namespace P4Tools {
 
+static constexpr uint8_t IPV4_BYTE_SIZE = 4U;
+static constexpr uint8_t IPV6_BYTE_SIZE = 16U;
+static constexpr uint8_t MAC_BYTE_SIZE = 6U;
+/// Chunk size is 8 bits, i.e., a byte.
+static constexpr uint8_t CHUNK_SIZE = 8U;
```
smolkaj, line 24: Would it be better to inline this number?
We are using uint8_t in several places, and several functions use "Bytes" in function and variable names. So it's not like this is a parameter that can be changed? I guess I don't see the motivation for declaring a global variable for this?

backends/p4tools/common/lib/format_int.cpp
```diff
@@ -268,4 +275,69 @@ std::string insertHexSeparators(const std::string &dataStr) {
     return insertSeparators(dataStr, "\\x", 2, false);
 }
 
+std::vector<uint8_t> convertBigIntToBytes(const big_int &dataInt, int targetWidthBits) {
```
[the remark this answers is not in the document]
smolkaj: Never mind, this is documented in the header file already.

backends/p4tools/modules/testgen/targets/bmv2/test_backend/protobuf_ir.cpp
```diff
@@ -9,14 +9,38 @@
...
+        BUG_CHECK(formatAnnotation->body.size() == 1,
+                  "@format annotation can only have one member.");
+        auto formatString = formatAnnotation->body.at(0)->text;
+        if (formatString == "IPV4_ADDRESS") {
+            formatString = "ipv4";
+        } else if (formatString == "IPV6_ADDRESS") {
+            formatString = "ipv6";
+        } else if (formatString == "MAC_ADDRESS") {
+            formatString = "mac";
+        } else {
```
smolkaj, line 37: Also support explicit HEX_STR annotation?

backends/p4tools/modules/testgen/targets/bmv2/test_backend/protobuf_ir.cpp
```diff
@@ -140,6 +164,105 @@ entities {
...
+        return formatHexExpr(value);
+    }
+    // At this point, any value must be a constant.
+    const auto *constant = value->checkedTo<IR::Constant>();
+    if (type == "ipv4") {
+        auto convertedString = convertToIpv4String(convertBigIntToBytes(constant->value, 32));
+        if (convertedString.has_value()) {
+            return convertedString.value();
+        }
+        BUG("Failed to convert IPv4 string.");
```
smolkaj, line 178: can we include the value that failed to be converted here? Similarly a few more times below.

smolkaj (approved):
Some nits, otherwise LGTM.

PR #4277, opened 2023-12-08
Fix problem with bpftool in PTF tests

tatry:
PTF tests recently fail due to failing build of bpftool used for some test cases. Solution is to use latest release instead of HEAD of main branch.

At least should work with bpftool v7.3.0

### DIFF
--- a/tools/ci-build.sh
+++ b/tools/ci-build.sh
@@ -180,9 +180,7 @@ function install_ptf_ebpf_test_deps() (
     sudo make install
 
     # install bpftool
-    # get tag for latest release, see https://docs.github.com/en/rest/releases/releases?apiVersion=2022-11-28#get-the-latest-release
-    BPFTOOL_RELEASE_TAG=$(curl -s https://api.github.com/repos/libbpf/bpftool/releases/latest | jq '.tag_name' | sed 's/\"//g')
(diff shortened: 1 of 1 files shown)


tools/ci-build.sh
```diff
@@ -180,7 +180,9 @@ function install_ptf_ebpf_test_deps() (
     sudo make install
 
     # install bpftool
-    git clone --recurse-submodules https://github.com/libbpf/bpftool.git /tmp/bpftool
+    # get tag for latest release, see https://docs.github.com/en/rest/releases/releases?apiVersion=2022-11-28#get-the-latest-release
+    BPFTOOL_RELEASE_TAG=$(curl -s https://api.github.com/repos/libbpf/bpftool/releases/latest | jq '.tag_name' | sed 's/\"//g')
+    git clone --recurse-submodules --branch "$BPFTOOL_RELEASE_TAG" https://github.com/libbpf/bpftool.git /tmp/bpftool
```
fruffy, line 185: I would just fix this to https://github.com/libbpf/bpftool/releases/tag/v7.3.0 instead? 

Building with the latest release could also easily lead to breakage.

tatry: > Building with the latest release could also easily lead to breakage.
Yes, you are right. After some additional research (made after this PR) I can say that fix to v7.3.0 will be the best solution.

fruffy (approved):

PR #4278, opened 2023-12-08
[P4Testgen] Fix accidental system file categorization.

fruffy:
Some files were accidentally classified as system files, although they were local. This causes problems with such as tools as include-fixer.

### DIFF
--- a/backends/p4tools/common/CMakeLists.txt
+++ b/backends/p4tools/common/CMakeLists.txt
@@ -44,10 +44,10 @@ target_link_libraries(
 
 target_include_directories(
   p4tools-common
-  SYSTEM BEFORE PUBLIC ${P4TOOLS_Z3_INCLUDE_DIR}
   PUBLIC "${CMAKE_BINARY_DIR}/common"
   PUBLIC "${P4C_SOURCE_DIR}"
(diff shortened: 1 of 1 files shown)


backends/p4tools/common/CMakeLists.txt
```diff
@@ -44,10 +44,10 @@ target_link_libraries(
 
 target_include_directories(
   p4tools-common
-  SYSTEM BEFORE PUBLIC ${P4TOOLS_Z3_INCLUDE_DIR}
   PUBLIC "${CMAKE_BINARY_DIR}/common"
   PUBLIC "${P4C_SOURCE_DIR}"
   PUBLIC "${P4C_BINARY_DIR}"
+  SYSTEM BEFORE PUBLIC ${P4TOOLS_Z3_INCLUDE_DIR}
```
vlstill, line 50: From https://cmake.org/cmake/help/latest/command/target_include_directories.html it seems that the `SYSTEM` should be actually specified at the beginning, I'm a bit surprised this works. I think it would be best to split the system and non-system libraries to a separate `target_include_directories`.

vlstill (approved):

PR #4284, opened 2023-12-11
Remove -static and rename static build appropriately.

fruffy:
Turns out we can not really build a clean static binary on Linux because of limitations of the glibc library. 

https://stackoverflow.com/questions/3430400/linux-static-linking-is-dead

Also, linking with `-static` causes problems when building Protobuf and [bdwgc](https://github.com/ivmai/bdwgc) from scratch. Hence, we remove `-static` from our build process and rename the CMake parameters appropriately.

### DIFF
--- a/.github/workflows/ci-static-build-test.yml
+++ b/.github/workflows/ci-static-build-test.yml
@@ -34,7 +34,8 @@ jobs:
         key: test-static-${{ runner.os }}
         max-size: 1000M
 
-    - name: Build (Ubuntu 20.04)
+    - shell: bash
+      name: Build (Ubuntu 20.04)
       run: |
         sudo -E tools/ci-build.sh
         # Disabling these checks until we find a better way to build a fully static binary.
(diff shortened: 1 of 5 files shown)


.github/workflows/ci-static-build-test.yml
```diff
@@ -37,3 +37,12 @@ jobs:
     - name: Build (Ubuntu 20.04)
       run: |
         sudo -E tools/ci-build.sh
+        # Disabling checks until we find a better way to build a fully static binary.
```
vlstill, line 40: There is a way to build fully static binaries on Linux. It basically requires using different libc implementation than glibc as glibc will do `dlopen` for some features even if build statically and therefore it is never fully static ([SO thread](https://stackoverflow.com/questions/57476533/why-is-statically-linking-glibc-discouraged)). One option that I have seen successfully used in using the [Alpine Linux](https://www.alpinelinux.org/) that uses musl libc. I can even try making that work for p4c if desirable. But there is a counter-argument to that and that is that linking libc statically moves the critical compatibility interface from glibc, which is widely used and attempts to be forward-compatible to Linux kernel interface, which does not need to be so stable. I'm not sure how much concern this actually is as I don't have enough knowledge about Linux kernel interlace.

For most uses it should be sufficient to use static-sans-glibc and to build on a distro with the oldest glibc that can possibly be used. Note that using musl instead of glibc will likely bring some performance differences (and a small potential for different bugs).

fruffy:
> @fruffy, what is the status of this? The changes look reasonable to me, but it is still marked WIP.

I was thinking about a solution similar to what you described in your comment. And I also need to implement the CI check. It is a little finicky since we need to grep for a lack of specific libraries.

vlstill:
> I was thinking about a solution similar to what you described in your comment.

I still think the static-sans-glibc is actually a more useful build type. There would be niche cases where fully-static would be desired, but that would basically require validating the whole compiler in that mode quite thoroughly.

CMakeLists.txt
```diff
@@ -127,10 +129,13 @@ if (BUILD_STATIC_RELEASE)
   set(Boost_USE_STATIC_RUNTIME OFF)
   # Set the static variable
   set(P4C_STATIC_BUILD STATIC)
+  # TODO: We can not use -static here because of compilation and portability problems:
+  # TODO: Keep track of https://stackoverflow.com/questions/3430400/linux-static-linking-is-dead
+  # and see whether static linking improves on Linux.
   # Do not bring in dynamic libstdcc and libgcc
-  set(CMAKE_EXE_LINKER_FLAGS "-static -static-libgcc -static-libstdc++ -Wl,-z,muldefs")
+  set(CMAKE_EXE_LINKER_FLAGS "-static-libgcc -static-libstdc++ -Wl,-z,muldefs")
```
vlstill, line 136: It would still make sense to include the previous value of `CMAKE_EXE_LINKER_FLAGS` I think. The user should have ability to add to them.

vlstill (commented):
Just one comment. Otherwise, I agree with this. It should probably be also reviewed by someone of the core p4c like @mihaibudiu.

vlstill (approved):

PR #4287, opened 2023-12-13
Add validation of constant enum values fitting

vlstill:
According to P4 spec v1.2.4 section 7.2.1 Enumeration types, the serialized enums should raise error if a value is not representable in the underlying type. This was not the case for p4c. I opted to add this check to type checking as it cannot run after it (as type checking cuts the constants, emitting just a warning), but running it before is more complicated as the type would need to be resolved manually. Therefore, the implementation has to resolve type names manually.

**User-visble changes**

* Compiler now rejects enumeration constants in serialized enums that are not representable in the underlying type. This now matches with the P4 spec.
* If for some reason a constant that does not fit is desired, it can be defined with an explicit cast.


vlstill:
Some example errors (from the unit tests):

```
(string)(5): [--Werror=type-error] error: unrepresentable: Serialized enum constant value -1 is out of bounds of the underlying type bit<8>; the value -1 is negative, but the underlying type bit<8> is unsigned
            unrepresentable = -1
            ^^^^^^^^^^^^^^^^^^^^

(string)(5): [--Werror=type-error] error: unrepresentable_n: Serialized enum constant value -129 is out of bounds of the underlying type int<8>; the value -129 requires 9 bits but the underlying signed type int<8> only contains 8 bits
            unrepresentable_n = -129
            ^^^^^^^^^^^^^^^^^^^^^^^^
(string)(2): [--Werror=type-error] error: int: Illegal type for enum; only bit<> and int<> are allowed; note that the used type is unsized integral type
        enum int FailingExample {
             ^^^

(string)(3): [--Werror=type-error] error: MyBit4: Illegal type for enum; only bit<> and int<> are allowed; note that a type-declared types are not allowed even if they are fixed-size
        enum MyBit4 FailingExample {
             ^^^^^^
```

Note that explicit cast can be used when the larger values is needed for some reason, in this case, you will just get a warning:

```
(string)(3): [--Wwarn=overflow] warning: -8s1000: signed value does not fit in 8 bits
            explicit_cast = (int<8>)-1000
                                     ^^^^

**Comment by jafingerhut** [comment]
@vlstill I like the new error checking proposed here.  Would you please also add new test P4 programs in the p4_16_errors directory that exercise all new error messages you are introducing?

File: frontends/p4/typeChecking/typeChecker.cpp [hunk 1]
```diff
@@ -1510,11 +1524,47 @@ const IR::Node *TypeInference::postorder(IR::SerEnumMember *member) {
...
+            note = "; note that the used type is unsized integral type";
+        }
+        if (resolvedType->is<IR::Type_Newtype>()) {
+            note = "; note that type-declared types are not allowed even if they are fixed-size";
+        }
+        typeError("%1%: Illegal type for enum; only bit<> and int<> are allowed%2%", serEnum->type,
+                  note);
         return member;
     }
+    // validate the constant fits -- non-fitting enum constants should pruce error
```
mihaibudiu, line 1541: produce

frontends/p4/typeChecking/typeChecker.cpp
```diff
@@ -1502,6 +1504,18 @@ const IR::Node *TypeInference::postorder(IR::Type_Set *type) {
     return type;
 }
 
+static int getReqSize(big_int val) {
```
mihaibudiu, line 1507: This function needs either a better name or a comment.

frontends/p4/typeChecking/typeChecker.cpp
```diff
@@ -1510,11 +1524,47 @@ const IR::Node *TypeInference::postorder(IR::SerEnumMember *member) {
...
+        }
+        if (resolvedType->is<IR::Type_Newtype>()) {
+            note = "; note that type-declared types are not allowed even if they are fixed-size";
+        }
+        typeError("%1%: Illegal type for enum; only bit<> and int<> are allowed%2%", serEnum->type,
+                  note);
         return member;
     }
+    // validate the constant fits -- non-fitting enum constants should pruce error
+    if (const auto *constant = member->value->to<IR::Constant>()) {
```
mihaibudiu, line 1542: I would have implemented this by casting the value to the underlying type, that should give a warning already.
But I guess this works.

mihaibudiu (approved):
I like better `testdata` tests than gtest tests, but this is fine too.

frontends/p4/typeChecking/typeChecker.cpp
vlstill: I think emitting the warning and then somehow processing it and turning it into error would be more confusing.

PR #4289, opened 2023-12-13
Add error on 0-size header stack

p-sawicki:
[P4 spec section on header stacks](https://p4.org/p4-spec/docs/P4-16-v-1.2.3.html#sec-header-stacks) states that their size must be a positive integer, but that is currently not checked and stacks with size 0 are allowed.

This PR adds the check.

### DIFF
--- a/frontends/p4/typeMap.cpp
+++ b/frontends/p4/typeMap.cpp
@@ -160,15 +160,23 @@ bool TypeMap::equivalent(const IR::Type *left, const IR::Type *right, bool stric
         return lv->getVarName() == rv->getVarName() && lv->getDeclId() == rv->getDeclId();
     }
     if (auto ls = left->to<IR::Type_Stack>()) {
+        const auto isLegal = [](const IR::Type_Stack *stack) {
+            if (!stack->sizeKnown()) {
+                ::error(ErrorType::ERR_TYPE_ERROR,
+                        "%1%: Size of header stack type should be a constant", stack);
+                return false;
+            }
+            if (stack->getSize() < 1) {
+                ::error(ErrorType::ERR_TYPE_ERROR,
+                        "Invalid size of header stack %1%. The size must be a positive integer.",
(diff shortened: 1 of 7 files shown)


mihaibudiu:
If we take the spec as correct, this PR is good.
The question is whether there are reasonable uses for 0-sized stacks, and the spec should be fixed instead.

vlstill:
> If we take the spec as correct, this PR is good. The question is whether there are reasonable uses for 0-sized stacks, and the spec should be fixed instead.

Is there any good reason to allow it? I could imagine if we had some strong meta-programming features in the language, one might want the ability to consider zero-sized stack to be just omitted. That said, I don't see how I would express a struct parametrized in this way in P4. If the only way of parametrization that can do it is preprocessor, then it is better to just have it omit the field entirely in my opinion.

It seems to me better to forbid this than to risk some passes will fail on this because they expect this is forbidden.

We can always revisit this if some new language feature comes that would make it reasonable to not have the limitation.

frontends/p4/typeMap.cpp
```diff
@@ -160,15 +160,23 @@ bool TypeMap::equivalent(const IR::Type *left, const IR::Type *right, bool stric
         return lv->getVarName() == rv->getVarName() && lv->getDeclId() == rv->getDeclId();
     }
     if (auto ls = left->to<IR::Type_Stack>()) {
+        const auto isLegal = [](const IR::Type_Stack *stack) {
```
vlstill, line 163: I'm surprised the checks are in this place, this does not seem the right place to put them to me. I don't even see why this placement guarantees the check will run during compilation.

Maybe this could be placed into type checking (although that one is very long already :-/), or into some other validation pass (it should run after constant folding though).

mihaibudiu:
Maybe this should first be discussed in the language design working group.
It has tackled the problem of 0-sized types before.

Issue #4288, opened 2023-12-13
Compiler Bug: At this point in the compilation typechecking should not infer new types anymore


p-sawicki:
The program

```
#include <core.p4>

control Deparser(packet_out packet);
package SimpleArch(Deparser dep);

header hdr_t {
   bit<8> f0;
}

control MyDeparser(packet_out packet) (hdr_t h) {
    apply {
      packet.emit(h);
    }
}

SimpleArch(MyDeparser({0})) main;
```

crashes in frontend with the message
```
Compiler bug triggered, please report this error: 
exception type: Util::CompilerBug
error: In file: ./p4c/frontends/p4/typeChecking/typeChecker.cpp:172
Compiler Bug: main.p4(12): At this point in the compilation typechecking should not infer new types anymore, but it did: node packet.emit<hdr_t>({0}); changed to packet.emit<hdr_t>({
  f0:0; });
      packet.emit(h);
      ^^^^^^^^^^^^^^
main.p4(12)
      packet.emit(h);
      ^^^^^^^^^^^^^^

jafingerhut:
I agree that improving the compiler to no longer experience a CompilerBug for such a program would be nice.

I only wanted to mention that probably one of the reasons that this has not been reported before is that it is extremely common for a deparser to take a parameter that is a struct type containing the header variables to be emitted, whereas this program has a deparser that takes exactly one header-type parameter to be emitted.  Thus it seems that while enhancing the compiler to avoid this CompilerBug is a nice-to-have, it might not be critical for actual P4 implementations to have.

mihaibudiu:
The typechecker makes no assumptions on the number of controls or their arguments, this must be something else.
It's probably related to the constructor argument which is instantiated with a header, there aren't many tests with constructor arguments.

mihaibudiu:
Yes, in fact the error message says that the error is that type inference changes the constructor argument to `{f0: 0}`. That is right, but it probably happens too late.

mihaibudiu:
Sorry, not constructor argument, but the function argument.

PR #4291, opened 2023-12-13
Allow one more round of type inference after specialization

mihaibudiu:
This is not the most principled fix, but hopefully it will work.
A principled fix would require more work.
I can describe the principled solution if anyone is interested to do it.
It basically boils down to improving `TypeChecking::containerInstantiation` to treat arguments like the `assign` function does.

### DIFF
--- a/frontends/p4/specialize.cpp
+++ b/frontends/p4/specialize.cpp
@@ -284,6 +284,8 @@ SpecializeAll::SpecializeAll(ReferenceMap *refMap, TypeMap *typeMap) : PassRepea
     passes.emplace_back(new TypeChecking(refMap, typeMap));
     passes.emplace_back(new FindSpecializations(&specMap));
     passes.emplace_back(new Specialize(&specMap));
+    passes.emplace_back(new ResolveReferences(refMap));
+    passes.emplace_back(new TypeInference(refMap, typeMap, false));  // more casts may be needed
     passes.emplace_back(new RemoveAllUnusedDeclarations(refMap));
     specMap.refMap = refMap;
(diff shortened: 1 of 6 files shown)


mihaibudiu:
Fixes #4288

fruffy (approved):
What is a better fix? Maybe we can create an issue for it. The passes can be quite expensive for larger programs, so keep the amount low is always a good thing.

mihaibudiu:
The last line of my comment describes the better fix, but it requires more work.

fruffy:
> The last line of my comment describes the better fix, but it requires more work.

How much work would be necessary here? I am not familiar with the typechecker, for example.

mihaibudiu:
No one is really familiar with the typechecker except me. So perhaps we should just merge this.
I don't plan to spend a couple of days on it.

vlstill:
> I can describe the principled solution if anyone is interested to do it.
> It basically boils down to improving TypeChecking::containerInstantiation to treat arguments like the assign function does.

Would that mean this particular change in the IR done by type checker would not happen? If so, why does it happen now?

mihaibudiu:
If you dump the program at various point during compilation you will notice that the constructor argument `{0}` is converted by the type checker to `{ 8w0 }` in the first pass. It should be converted to `(hdr_t){ f0 = 8w0; }`. This is what the latter pass does, after the constructor argument is "inlined" by the specialization pass. This PR merely allows this change to be made this late in the compilation process.

vlstill:
Thanks for the explanation.

@mihaibudiu, could you please merge this, or is there a reason this was not done yet?

PR #4296, opened 2023-12-17
Fix initialization order in gtest helper

vlstill:
In #4287 I've accidentally create use of unitialized value in test because of initialization order of the `RedirectStderr`. This fixes it by making sure the stringstream is initialized first.

### DIFF
--- a/test/gtest/helpers.h
+++ b/test/gtest/helpers.h
@@ -149,8 +149,8 @@ struct RedirectStderr {
     }
 
  private:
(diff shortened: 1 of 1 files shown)


fruffy (approved):
Shouldn't our CI have caught this? Why didn't the -Werror work?

vlstill:
This ways found by clang... maybe gcc ignored it for some reason... we should probably have a build with clang too (I can add that), my experience is that each of them has some warnings the other is missing.

PR #4297, opened 2023-12-17
[TESTING] Remove use of deprecated C++ features, allow compilation with C++20

vlstill:


fruffy:
@vlstill Should we rebase this PR. Things might be easier after recent updates.

vlstill:
@fruffy I've tried this few weeks ago after the recent changes like removal of StringRef. Sadly it still did not work easily. The main problem is that there is a lot of implicit casts between string-like types (`cstring` and `IR::ID` being the main culprits I believe). With the new rules for comparsion operators in C++20 (mainly the symmetry -- ability of the compiler to flip arguments of == and !=) this produces a lot of ambiguous overloads. I think the best way to get rid of these is actually to force lot of implicit casts to be explicit. At least the `IR::ID` shouldn't really be implicitly constructible from string types. This will be a breaking change of course, but it should not be too hard to apply to the downstream projects.

I did not get much further as sadly I don't have much time to spare for the open source p4c lately :-(.

fruffy:
> @fruffy I've tried this few weeks ago after the recent changes like removal of StringRef. Sadly it still did not work easily. The main problem is that there is a lot of implicit casts between string-like types (`cstring` and `IR::ID` being the main culprits I believe). With the new rules for comparsion operators in C++20 (mainly the symmetry -- ability of the compiler to flip arguments of == and !=) this produces a lot of ambiguous overloads. I think the best way to get rid of these is actually to force lot of implicit casts to be explicit. At least the `IR::ID` shouldn't really be implicitly constructible from string types. This will be a breaking change of course, but it should not be too hard to apply to the downstream projects.
> 
> I did not get much further as sadly I don't have much time to spare for the open source p4c lately :-(.

Let me check how much work needs to be done with the current compiler.

vlstill:
@fruffy I've pushed what I had in my local branch here. The old one is https://github.com/vlstill/p4c/tree/forward-cpp-friendly-old.

vlstill:
stale, closing

PR #4298, opened 2023-12-18
[P4Testgen] Fix stringliteral conversion. Value must be a literal, NOT a constant

fruffy:
Quite the subtle issue I ran into. In the Z3 solver, string_constants are not constants but functions. We need to assign a value instead to get the correct semantic behavior. Otherwise, the solver can freely assign a value to the function.

### DIFF
--- a/backends/p4tools/modules/testgen/test/z3-solver/constraints.cpp
+++ b/backends/p4tools/modules/testgen/test/z3-solver/constraints.cpp
@@ -1,0 +1,191 @@
+#include <gtest/gtest.h>
+
+#include <optional>
+#include <vector>
+
+#include "backends/p4tools/common/core/z3_solver.h"
+#include "backends/p4tools/common/lib/variables.h"
+#include "ir/ir-generated.h"
+#include "ir/ir.h"
+#include "ir/irutils.h"
+#include "lib/cstring.h"
+
+namespace Test {
+using ConstraintVector = const std::vector<const Constraint *>;
+
+class Z3SolverSatisfiabilityChecks : public ::testing::Test {
+    P4Tools::Z3Solver solver;
+
+ public:
+    /// Checks whether the result of the solver calculating @param expression matches @param
+    /// expectedResult.
+    void testCheckSat(const ConstraintVector &expression, std::optional<bool> expectedResult) {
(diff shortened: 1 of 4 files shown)


backends/p4tools/modules/testgen/test/z3-solver/constraints.cpp
```diff
@@ -0,0 +1,191 @@
...
+        testCheckSat(inputExpression, false);
+    }
+    {
+        auto *expression = new IR::Equ(fooVar, IR::getConstant(eightBitType, 1));
+        ConstraintVector inputExpression = {expression};
+        testCheckSat(inputExpression, true);
+    }
+    {
+        auto *expression = new IR::Equ(fooVar, IR::getConstant(eightBitType, 1));
+        auto *constraint1 = new IR::Equ(fooVar, IR::getConstant(eightBitType, 1));
```
vlstill, line 51: Is this deliberately the same as `expression`?

backends/p4tools/modules/testgen/test/z3-solver/constraints.cpp
```diff
@@ -0,0 +1,191 @@
...
+    {
+        auto *expression = new IR::Equ(fooVar, barVar);
+        auto *constraint1 = new IR::Equ(fooVar, IR::getConstant(eightBitType, 1));
+        ConstraintVector inputExpression = {expression, constraint1};
+        testCheckSat(inputExpression, true);
+    }
+    {
+        auto *expression = new IR::Equ(fooVar, barVar);
+        auto *constraint1 = new IR::Equ(fooVar, IR::getConstant(eightBitType, 1));
+        auto *constraint2 = new IR::Equ(fooVar, IR::getConstant(eightBitType, 1));
```
vlstill, line 64: Is this deliberately the same as `constraint2`?

vlstill (commented):
I agree with the change, good catch. I just have a question about the tests.

backends/p4tools/modules/testgen/test/z3-solver/constraints.cpp
fruffy: Yes, the other one was not. Fixed.

vlstill (approved):

PR #4299, opened 2023-12-18
Remove uses of `incompatible_use_toolchain_transition`.

katre:
Now that Bazel 7.0 has been released, it's time to remove this tech debt.

This has been a no-op since Bazel 5.0, I've been waiting to remove the code for two years, it's time.

Part of https://github.com/bazelbuild/bazel/issues/14127.

### DIFF
--- a/bazel/p4_library.bzl
+++ b/bazel/p4_library.bzl
@@ -172,7 +172,6 @@ p4_library = rule(
         ),
         "_cc_toolchain": attr.label(default = Label("@bazel_tools//tools/cpp:current_cc_toolchain")),
     },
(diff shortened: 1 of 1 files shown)


fruffy:
Fyi, Bazel is pinned to 6.4.0 [here](https://github.com/p4lang/p4c/blob/main/.bazelversion) and [here](https://github.com/p4lang/p4c/blob/main/bazel/example/.bazelversion#L1) because of a (small) breaking change in 7.0. Happy to approve if that is not an issue.

katre:
This is fine with Bazel 6.40, please merge.

smolkaj (approved):

PR #4300, opened 2023-12-19
Make sure P4 expression optimization does not strip away types

vlstill:
Fixes this error:

```
Internal error: In file: ./p4c/backends/p4tools/modules/testgen/lib/continuation.cpp:124
P4Testgen Bug: Continuation v; has parameter of type bit<32>, but was given an argument (|*packetLen_bits(bit<32>)| >> 3) + 4294967238; of type <Type_Unknown>(2)
```

Which is caused by `P4::optimizeExpression` not preserving type when changing `x - CONST` to `x + (-CONST)`.

The change is mainly in two passes: ConstantFolding and StrengthReduction -- I've tried to make sure the types of expressions are preserved by these passes. I've also added propagation of source info to a few places.

### DIFF
--- a/backends/p4tools/common/lib/symbolic_env.cpp
+++ b/backends/p4tools/common/lib/symbolic_env.cpp
@@ -27,6 +27,8 @@ const IR::Expression *SymbolicEnv::get(const IR::StateVariable &var) const {
 bool SymbolicEnv::exists(const IR::StateVariable &var) const { return map.find(var) != map.end(); }
 
 void SymbolicEnv::set(const IR::StateVariable &var, const IR::Expression *value) {
+    BUG_CHECK(value->type && !value->type->is<IR::Type_Unknown>(),
+              "Cannot set value with unspecified type: %1%", value);
     map[var] = value;
 }
 
--- a/backends/p4tools/common/lib/symbolic_env.cpp
+++ b/backends/p4tools/common/lib/symbolic_env.cpp
@@ -8,7 +8,6 @@
 #include <boost/container/vector.hpp>
 
 #include "backends/p4tools/common/lib/model.h"
-#include "frontends/p4/optimizeExpressions.h"
 #include "ir/indexed_vector.h"
 #include "ir/vector.h"
 #include "ir/visitor.h"
@@ -27,33 +26,7 @@ const IR::Expression *SymbolicEnv::get(const IR::StateVariable &var) const {
 
 bool SymbolicEnv::exists(const IR::StateVariable &var) const { return map.find(var) != map.end(); }
 
-/// Compare types, considering Extracted_Varbit and bits equal if the (real/extracted) sizes are
-/// equal. This is because the packet expression can be something like 0 ++
-/// (Extracted_Varbit<N>)pkt_var. This expression is typed as bit<N>, but the optimizer removes the
-/// 0 ++ and makes it into Extracted_Varbit type.
-/// TODO: Maybe we can just get rid of Extracted_Varbit and use bit it its place everywhere.
-static bool typeEquivSansVarbit(const IR::Type *a, const IR::Type *b) {
-    if (a->equiv(*b)) {
-        return true;
-    }
-    const auto *abit = a->to<IR::Type_Bits>();
-    const auto *avar = a->to<IR::Extracted_Varbits>();
-    const auto *bbit = b->to<IR::Type_Bits>();
-    const auto *bvar = b->to<IR::Extracted_Varbits>();
-    return (abit && bvar && abit->width_bits() == bvar->width_bits()) ||
-           (avar && bbit && avar->width_bits() == bbit->width_bits());
-}
-
 void SymbolicEnv::set(const IR::StateVariable &var, const IR::Expression *value) {
-    const auto *type = value->type;
-    BUG_CHECK(type && !type->is<IR::Type_Unknown>(), "Cannot set value with unspecified type: %1%",
-              value);
-    value = P4::optimizeExpression(value);
-    BUG_CHECK(value->type && !value->type->is<IR::Type_Unknown>(),
-              "The P4 expression optimizer stripped a type of %1% (was %2%)", value, type);
-    BUG_CHECK(typeEquivSansVarbit(type, value->type),
-              "The P4 expression optimizer had changed type of %1% (%2% -> %3%)", value, type,
-              value->type);
     map[var] = value;
 }
 
--- a/backends/p4tools/common/lib/symbolic_env.cpp
+++ b/backends/p4tools/common/lib/symbolic_env.cpp
@@ -28,7 +28,18 @@ const IR::Expression *SymbolicEnv::get(const IR::StateVariable &var) const {
 bool SymbolicEnv::exists(const IR::StateVariable &var) const { return map.find(var) != map.end(); }
 
 void SymbolicEnv::set(const IR::StateVariable &var, const IR::Expression *value) {
-    map[var] = P4::optimizeExpression(value);
+    const auto *type = value->type;
+    BUG_CHECK(type && !type->is<IR::Type_Unknown>(), "Cannot set value with unspecified type: %1%",
+              value);
+    value = P4::optimizeExpression(value);
+    // The P4 passes used in optimizeExpression can strip types which then breaks P4Tools, make sure
+    // at least the top-level type is set.
+    if (value->type->is<IR::Type_Unknown>()) {
+        auto *clone = value->clone();
+        clone->type = type;
+        value = clone;
+    }
+    map[var] = value;
 }
 
 const IR::Expression *SymbolicEnv::subst(const IR::Expression *expr) const {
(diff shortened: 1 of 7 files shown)


backends/p4tools/common/lib/symbolic_env.cpp
```diff
@@ -28,7 +28,18 @@ const IR::Expression *SymbolicEnv::get(const IR::StateVariable &var) const {
...
 
 void SymbolicEnv::set(const IR::StateVariable &var, const IR::Expression *value) {
-    map[var] = P4::optimizeExpression(value);
+    const auto *type = value->type;
+    BUG_CHECK(type && !type->is<IR::Type_Unknown>(), "Cannot set value with unspecified type: %1%",
+              value);
+    value = P4::optimizeExpression(value);
+    // The P4 passes used in optimizeExpression can strip types which then breaks P4Tools, make sure
+    // at least the top-level type is set.
+    if (value->type->is<IR::Type_Unknown>()) {
```
fruffy, line 37: I'd prefer to fix the optimization pass and have a BUG_CHECK here instead. We are using StrengthReduction and ConstantFolding. They definitely should preserve types.

vlstill: I will have a look into those two passes. But in general my impression from many of the P4 passes in the compiler is that they don't bother with types and just expect type checker to run after :-/. I think that is conceptually a wrong decision, but it could be hard to change it.

fruffy: True, but we can still fix up individual passes that only handle expressions etc. I think I stumbled across this problem before and did some work to fix it. The changes necessary to the passes were not major but I unfortunately discarded them..

backends/p4tools/common/lib/symbolic_env.cpp
```diff
@@ -28,7 +28,16 @@ const IR::Expression *SymbolicEnv::get(const IR::StateVariable &var) const {
...
 
 void SymbolicEnv::set(const IR::StateVariable &var, const IR::Expression *value) {
-    map[var] = P4::optimizeExpression(value);
+    const auto *type = value->type;
+    BUG_CHECK(type && !type->is<IR::Type_Unknown>(), "Cannot set value with unspecified type: %1%",
+              value);
+    value = P4::optimizeExpression(value);
+    BUG_CHECK(value->type && !value->type->is<IR::Type_Unknown>(),
+              "The P4 expression optimizer stripped a type of %1% (was %2%)", value, type);
+    BUG_CHECK(type->equiv(*value->type),
```
fruffy, line 37: I think it is fair to change the underlying type of the expression. You could imagine something like `bit<8>(0)` ++ `bit<8>(4)` being optimized to `bit<16>(4)`.

vlstill: In that case the `++` would still be `bit<16>`. If an optimizer changed type I think that would have to break type checking.

fruffy: Duh. My bad. Can't think of any other example. Even casts preserve the type.

backends/p4tools/common/lib/symbolic_env.cpp
vlstill: @fruffy, I came across one point where the optimizer does change the type. When there is just a varbit extracted (e.g. `testgen-p4c-bmv2-protobuf-ir/issue447-bmv2.p4`), the packet expression is something line `0 ++ (SizedVarbit<32><16>)|pktvar_0(bit<16>)|` which is of type `bit<16>` but the optimizer turns it into `(SizedVarbit<32><16>)|pktvar_0(bit<16>)|` which is `SizedVarbit<32><16>`. For now I've commited a change that considers `bit<N>` and `SizedVarbit<M><N>` the same for the validity check.

Looking at the other code that relates to varbits, it seems like these are mostly special cases that treat `SizedVarbit<M><N>` as `bit<N>`. So I think we might be just fine leaving the `varbit<M>` in the IR and handling the special case in `extract` only (with the extracted type being `bit<N>`. This would probably simplify the overall handling. The only cost I can see would be that the trace would not contain info that the type is a varbit. What do you think? This should be also OK for the targets here in p4c, I'm not sure about Tofino (@pkotikal) and of course other targets that I don't know about. But I think that should be a separate PR.

backends/p4tools/modules/testgen/core/small_step/extern_stepper.cpp
```diff
@@ -37,7 +37,9 @@ std::vector<std::pair<IR::StateVariable, const IR::Expression *>> ExprStepper::s
     ExecutionState &nextState, const std::vector<IR::StateVariable> &flatFields,
     int varBitFieldSize) {
     std::vector<std::pair<IR::StateVariable, const IR::Expression *>> fields;
-    for (const auto &fieldRef : flatFields) {
+    // Make a copy of the StateVariable so it can be modified in the varbit case (and it is just a
```
fruffy, line 40: Hmm, in this case we are making a copy for the common case (no Varbit). Is that really better?

vlstill: This is quite small class (one pointer + vptr) so I'd say this does not matter either way. Could be even better since there will be no dereference, but could be also exactly the same if the optimizer is smart enough (I haven't checked).

backends/p4tools/common/lib/symbolic_env.cpp
```diff
@@ -27,8 +27,34 @@ const IR::Expression *SymbolicEnv::get(const IR::StateVariable &var) const {
...
+    const auto *avar = a->to<IR::Extracted_Varbits>();
+    const auto *bbit = b->to<IR::Type_Bits>();
+    const auto *bvar = b->to<IR::Extracted_Varbits>();
+    return (abit && bvar && abit->width_bits() == bvar->width_bits()) ||
+           (avar && bbit && avar->width_bits() == bbit->width_bits());
+}
+
 void SymbolicEnv::set(const IR::StateVariable &var, const IR::Expression *value) {
-    map[var] = P4::optimizeExpression(value);
+    const auto *type = value->type;
```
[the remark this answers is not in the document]
vlstill: You mean having the optimization and checks in ExecutionState::set instead of here?

backends/p4tools/common/lib/symbolic_env.cpp
```diff
@@ -27,8 +27,34 @@ const IR::Expression *SymbolicEnv::get(const IR::StateVariable &var) const {
 
 bool SymbolicEnv::exists(const IR::StateVariable &var) const { return map.find(var) != map.end(); }
 
+/// Compare types, considering Extracted_Varbit and bits equal if the (real/extracted) sizes are
+/// equal. This is because the packet expression can be something like 0 ++
+/// (Extracted_Varbit<N>)pkt_var. This expression is typed as bit<N>, but the optimizer removes the
+/// 0 ++ and makes it into Extracted_Varbit type.
+/// TODO: Maybe we can just get rid of Extracted_Varbit and use bit it its place everywhere.
```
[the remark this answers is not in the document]
vlstill: Oh, you mean in assignment of headers that contain varbits and in direct varbit field assignment? I did not think about that, but of course it should work too. In that case we need to know something was varbit and can't replace it with just `bit<N>`.

backends/p4tools/common/lib/symbolic_env.cpp
```diff
@@ -28,7 +27,7 @@ const IR::Expression *SymbolicEnv::get(const IR::StateVariable &var) const {
 bool SymbolicEnv::exists(const IR::StateVariable &var) const { return map.find(var) != map.end(); }
 
 void SymbolicEnv::set(const IR::StateVariable &var, const IR::Expression *value) {
-    map[var] = P4::optimizeExpression(value);
```
vlstill, line 31: If we remove the optimization, I'm not sure if the check if type is set should be here too? May not be necessary for all tools that use `SymbolicEnv`.

backends/p4tools/common/lib/symbolic_env.cpp
vlstill: Let's raise an issue for this so it is not forgotten but leave it for later (I'll raise the issue).

backends/p4tools/modules/testgen/core/small_step/extern_stepper.cpp
fruffy: Yeah I think that is a check we should keep in this particular setter. The other code can be moved out.

fruffy (approved):

backends/p4tools/modules/testgen/core/small_step/extern_stepper.cpp
vlstill: I'm confused, you have presubmably replied to a different thread, did you mean that the check for type being set in the expression should be (also) in `SymbolicEnv::set`?

fruffy: Yes, the comment was meant for https://github.com/p4lang/p4c/pull/4300#discussion_r1442624167. Not sure what happened.

PR #4301, opened 2023-12-19
Dpdk Backend: Fix incorrect field extraction arithmetic

usha1830:
when multiple non-byte aligned header fields are combined into single byte aligned field. The references of these fields are updated with incorrect slice operation.

For ex: when 
```
 header ipv4_t {
                            ...
                            bit<3>  flags;
                            bit<13> fragOffset;
                            ...
                        }
```

 is converted into

 ```
                   header ipv4_t {
                            ...
                            bit<16>  flags_fragOffset;
                            ...
                        }
```

_flags_ field should be converted to _flags_fragOffset[15:13]_
_fragOffset_ should be converted to _flags_fragOffset[12:0]_

But currently the conversion happens like this
_flags_ field is converted to  _flags_fragOffset[2:0]_
_fragOffset_ is converted to  _flags_fragOffset[15:3]_

Fixes https://github.com/p4lang/p4c/issues/4077

### DIFF
--- a/testdata/p4_16_samples_outputs/pna-dpdk-header-stack-assignment.p4.spec
+++ b/testdata/p4_16_samples_outputs/pna-dpdk-header-stack-assignment.p4.spec
@@ -122,38 +122,38 @@ action encap_one_tunnel_layer_ipv4 args instanceof encap_one_tunnel_layer_ipv4_a
 	mov h.mac.type 0x800
 	validate h.ipv4_0
 	mov m.MainControlT_tmp h.ipv4_0.version_ihl
-	and m.MainControlT_tmp 0xF0
+	and m.MainControlT_tmp 0xF
 	mov h.ipv4_0.version_ihl m.MainControlT_tmp
-	or h.ipv4_0.version_ihl 0x4
+	or h.ipv4_0.version_ihl 0x40
 	mov m.MainControlT_tmp_0 h.ipv4_0.version_ihl
-	and m.MainControlT_tmp_0 0xF
+	and m.MainControlT_tmp_0 0xF0
 	mov h.ipv4_0.version_ihl m.MainControlT_tmp_0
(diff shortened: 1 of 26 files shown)

### CODE REVIEW

File: testdata/p4_16_samples_outputs/pna-dpdk-header-stack-assignment.p4.spec [hunk 1]
```diff
@@ -122,38 +122,38 @@ action encap_one_tunnel_layer_ipv4 args instanceof encap_one_tunnel_layer_ipv4_a
 	mov h.mac.type 0x800
 	validate h.ipv4_0
 	mov m.MainControlT_tmp h.ipv4_0.version_ihl
-	and m.MainControlT_tmp 0xF0
+	and m.MainControlT_tmp 0xF
```
mihaibudiu, line 125: how come this bug wasn't caught?

mihaibudiu (approved):

testdata/p4_16_samples_outputs/pna-dpdk-header-stack-assignment.p4.spec
jafingerhut: It was caught.  That is why there was a bug report about it.

Do you mean, why wasn't it caught earlier than Sep 2023 when I reported it?

Probably at least in part because there are no automated CI tests that pass data packets through the P4 DPDK that would have caught this.  Fabian Ruffy has tried to create some such tests, but he has had difficulty making them work consistently enough to use for CI purposes on this repository.

usha1830: And Perhaps no user tried accessing the individual header fields of the combined fields. Some of the early users may be combining these fields by themselves in the P4 program and thereby handling these extractions also in the program itself.
Thanks to Andy who caught this!!

PR #4306, opened 2023-12-20
[P4Testgen] Fix append and prepend invocation.

fruffy:
@pkotikal This should help with the problems you have encountered in https://github.com/p4lang/p4c/commit/689e013d3d870fa906852d93acd76cfc5bbe5634#r135134661

Since we are recording an expression, fix also the way we print expressions. I should implement a more disciplined way to format these inputs without introducing linebreaks.

### DIFF
--- a/backends/p4tools/common/lib/trace_event_types.cpp
+++ b/backends/p4tools/common/lib/trace_event_types.cpp
@@ -50,15 +50,28 @@ const Expression *Expression::apply(Transform &visitor) const {
 }
 
 const Expression *Expression::evaluate(const Model &model, bool doComplete) const {
-    if (Taint::hasTaint(value)) {
-        return new Expression(&Taint::TAINTED_STRING_LITERAL, label);
+    const IR::Expression *evaluatedValue = nullptr;
+    if (const auto *structExpr = value->to<IR::StructExpression>()) {
+        evaluatedValue = model.evaluateStructExpr(structExpr, doComplete);
+    } else if (const auto *listExpr = value->to<IR::BaseListExpression>()) {
+        evaluatedValue = model.evaluateListExpr(listExpr, doComplete);
+    } else if (Taint::hasTaint(value)) {
+        evaluatedValue = &Taint::TAINTED_STRING_LITERAL;
+    } else {
+        evaluatedValue = model.evaluate(value, doComplete);
     }
-    return new Expression(model.evaluate(value, doComplete), label);
+    return new Expression(evaluatedValue, label);
 }
 
 void Expression::print(std::ostream &os) const {
+    // Convert the assignment to a string and strip any new lines.
+    // TODO: Maybe there is a better way to format newlines?
+    std::stringstream exprStream;
+    value->dbprint(exprStream);
+    auto expString = exprStream.str();
+    expString.erase(std::remove(expString.begin(), expString.end(), '\n'), expString.cend());
     Generic::print(os);
-    os << ": " << formatHexExpr(value, {true});
+    os << ": " << expString;
 }
 
 /* =============================================================================================
@@ -154,9 +167,9 @@ const AssignmentStatement *AssignmentStatement::apply(Transform &visitor) const 
 const AssignmentStatement *AssignmentStatement::evaluate(const Model &model,
                                                          bool doComplete) const {
     const IR::Expression *right = nullptr;
-    if (auto structExpr = stmt.right->to<IR::StructExpression>()) {
+    if (const auto *structExpr = stmt.right->to<IR::StructExpression>()) {
         right = model.evaluateStructExpr(structExpr, doComplete);
-    } else if (auto listExpr = stmt.right->to<IR::BaseListExpression>()) {
+    } else if (const auto *listExpr = stmt.right->to<IR::BaseListExpression>()) {
         right = model.evaluateListExpr(listExpr, doComplete);
     } else if (Taint::hasTaint(stmt.right)) {
         right = &Taint::TAINTED_STRING_LITERAL;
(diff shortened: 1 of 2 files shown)


fruffy:
@vlstill Don't want to burden you with too many requests but this one has been stuck for a while.

backends/p4tools/modules/testgen/core/small_step/extern_stepper.cpp
```diff
@@ -164,29 +164,15 @@ void ExprStepper::evalInternalExternMethodCall(const IR::MethodCallExpression *c
              auto &nextState = state.clone();
              const auto *prependType = state.resolveType(prependVar->type);
 
-             if (prependType->is<IR::Type_StructLike>()) {
-                 auto prependRef = ToolsVariables::convertReference(prependVar);
-                 // We only support flat assignments, so retrieve all fields from the input
-                 // argument.
-                 const auto flatFields = nextState.getFlatFields(prependRef);
-                 // Iterate through the fields in reverse order.
```
vlstill, line 172: Why we don't have to reverse any more?

backends/p4tools/modules/testgen/core/small_step/extern_stepper.cpp
```diff
@@ -164,29 +164,15 @@ void ExprStepper::evalInternalExternMethodCall(const IR::MethodCallExpression *c
...
+                     nextState.prependToPacketBuffer(expr);
                  }
              } else if (prependType->is<IR::Type_Bits>()) {
-                 if (!(prependVar->is<IR::Member>() || prependVar->is<IR::PathExpression>() ||
-                       prependVar->is<IR::TaintExpression>() || prependVar->is<IR::Constant>())) {
-                     TESTGEN_UNIMPLEMENTED("Prepend input %1% of type %2% not supported",
-                                           prependVar, prependVar->type);
-                 }
-                 // Prepend the field to the packet buffer.
-                 nextState.add(*new TraceEvents::Expression(prependVar, "PrependToProgramHeader"));
```
vlstill, line 187: A trace got lost in this case? Is that intentional?

backends/p4tools/common/lib/trace_event_types.cpp
```diff
@@ -50,15 +50,28 @@ const Expression *Expression::apply(Transform &visitor) const {
 }
 
 const Expression *Expression::evaluate(const Model &model, bool doComplete) const {
-    if (Taint::hasTaint(value)) {
-        return new Expression(&Taint::TAINTED_STRING_LITERAL, label);
+    const IR::Expression *evaluatedValue = nullptr;
+    if (const auto *structExpr = value->to<IR::StructExpression>()) {
+        evaluatedValue = model.evaluateStructExpr(structExpr, doComplete);
+    } else if (const auto *listExpr = value->to<IR::BaseListExpression>()) {
+        evaluatedValue = model.evaluateListExpr(listExpr, doComplete);
```
vlstill, line 57: I'm curious, why this has to be done here and not in the model itself?

backends/p4tools/modules/testgen/core/small_step/extern_stepper.cpp
```diff
@@ -210,21 +196,14 @@ void ExprStepper::evalInternalExternMethodCall(const IR::MethodCallExpression *c
...
+                 for (const auto *expr : exprList) {
+                     nextState.appendToPacketBuffer(expr);
                  }
              } else if (appendType->is<IR::Type_Bits>()) {
-                 if (!(appendVar->is<IR::Member>() || appendVar->is<IR::PathExpression>() ||
-                       appendVar->is<IR::TaintExpression>() || appendVar->is<IR::Constant>())) {
-                     TESTGEN_UNIMPLEMENTED("append input %1% of type %2% not supported", appendVar,
-                                           appendVar->type);
-                 }
-                 nextState.add(*new TraceEvents::Expression(appendVar, "AppendToProgramHeader"));
```
vlstill, line 227: This piece of code is very similar to the one in prepend, could you please factor it out?

backends/p4tools/modules/testgen/core/small_step/extern_stepper.cpp
fruffy: Previously we were iterating over the individual members when prepending. If you started from the beginning and prepend you would invert the order of fields. Now we just prepend the entire struct expression and order is preserved.

backends/p4tools/modules/testgen/core/small_step/extern_stepper.cpp
fruffy: I was kind of following the rule of three here. It did not seem worthwhile to factor it out because I would need to parametrize the append/prepend including the trace event. You end up saving not much and just gain more indirection.

backends/p4tools/common/lib/trace_event_types.cpp
fruffy: The problem is simply typing. `evaluate` returns an `IR::Literal`. The expectation of calling evaluate is that it will give you a `IR::Literal` or a list/struct of literals. 

To simplify I would need to add another function that returns `IR::Expression` to the model or simplify the entire thing to return `IR::Expression`.

vlstill: I see, OK. It is probably better to keep the more concrete types unless we need to go over all the options in many places.

backends/p4tools/modules/testgen/core/small_step/extern_stepper.cpp
vlstill: I understand why the reverse was there. What confuses me is that there is still an iteration over the flattened expression, so I would expect the reverse is still needed.

fruffy: You are right, the reverse iterator should still be there. I was looking at the wrong piece of code...

We are passing a struct expression but once we unroll that struct expression we still need to do that in reverse.

fruffy:
@vlstill Can you give this another look?

vlstill (approved):

Issue #4304, opened 2023-12-19
[P4Testgen] Add a p4test-like target to testgen (?)


vlstill:
@fruffy, I think you have mentioned this already at some point and I think this would make sense. Currently, all the testgen P4 testing happens on one of the targets which are actually not too simple (e.g. BMV2) and one might not actually want them build, especially if one is working on another testgen target. I think it would make sense to have a testgen target with a very minimal architecture and simple, testing-oriented output format that would be used basically just for testing and bug reproduction. Then we could have tests similar to the frontend/midend compiler tests that just record the expected outputs (from testgen in this case).

I think an architecture with minimal parser/control/deparser and no externs defined by the architecture would be in order for this. The output could be either something like STF, or (I thing this would be even better) a JSON-based format that would be easier to inspect on failures.

@fruffy, and others interested in testgen, what do you think about this? Is this something we want to pursue.

fruffy:
That would definitely be nice. The biggest hurdle is patching together the package, which can be in arbitrary order. With https://github.com/p4lang/p4c/pull/4267 it should be easier, since we can just relax the arch specification and take the package as is for generating tests. 

However, I have no plans to pursue this for now since I am working on a couple other rewrites.

jnfoster:
Just to make sure I understand: the idea is to use `p4testgen` as a _concrete_ interpreter? Then we can run tests natively without having to start up a separate software model. 

This is obviously not useful for testing a real target, but could be useful for tracking bugs and such in the compiler.

If that's right, sounds cool!

vlstill:
@jnfoster I'm not sure I understand your comment, so I will try to explain myself a bit more.

The idea is to have a testgen target that does not correspond to any real target, but is kind of a minimal example and it useful for testing. My original ideas were about testgen testing, but actually it could be very useful also for compiler testing (since the testgen uses a lot of the common passes).

My idea was to define a very simple arch (and a fixed one, so I was not thinking about having flexibility there as @fruffy's comment seems to suggest). Maybe something like this:

```p4
#include <core.p4>

struct metadata_t {
    bit<8> ingress_port;
    bit<8> egress_port;
    bit<1> dropped;
}

extern void assume(in bool check);
extern void assert(in bool check);

parser ParserT<H, M>(packet_in pkt, out H hdrs, out M md_user, inout metadata_t meta);
// I don't know where to put parser_error so I put it in the main control as an extra argument,
// similar to how PNA has it in intrinsic md_user in pre-control
[...]

PR #4307, opened 2023-12-21
[P4Testgen] Implement coverage tracking of actions

### LINKED ISSUE #4305: [P4Testgen] Coverage ignores empty actions
This is probably an unforeseen artifact of the way testgen counts coverage. If there is an empty action, there are no IR nodes it in and therefore (presumably) it does not add to coverage metric. This means that in certain invocations (e.g. `--path-selection GREEDY_STATEMENT_SEARCH --track-coverage STATEMENTS --track-coverage TABLE_ENTRIES --only-covering-tests --stop-metric MAX_NODE_COVERAGE`), the execution that leads to empty action is ignored.

For example, with this code snippet only `a0` and `a3` were hit using these options (and I suspect `a0` was only hit because it was first).
```p4
    action t0_a0() { NoAction(); }
    action t0_a1() { }
    action t0_a2() { ; }
    action t0_a3() { hdr.h0.v0 = 42; }

    table t0 {
        key = {
            hdr.h0.v0 : exact;
            hdr.h0.v1 : exact;
        }
        actions = {
            t0_a0;
            t0_a1;
            t0_a2;
            t0_a3;
        }
        default_action = t0_a0;
        size = 32;
    }

    apply{
        if (hdr.h0.isValid())
            t0.apply();
    }
```

What would be a good solution here? One thing I can think about is adding an empty statement to each empty action in the testgen's midend. Then the tracking could be left as is and the cost would be presumably negligible. I think this could be used in all cases (both for `STATEMENTS` and for `TABLE_ENTRIES` coverage). @fruffy, what do you think?

vlstill:
Resolves #4305.

This PR adds a new coverage tracking option for P4Testgen: `--track-coverage ACTIONS`. This makes it possible to track coverage of action invocation, including empty actions which are not included in the statement coverage because they don't contain any statements.

I've added a test that exercises this feature and that I have used to test it manually. However, I don't see a way we can currently automatically test that this testgen will generate all the right tests for this test case. This could be tested better if/when #4304 is implemented.

### DIFF
--- a/backends/p4tools/modules/testgen/core/small_step/table_stepper.cpp
+++ b/backends/p4tools/modules/testgen/core/small_step/table_stepper.cpp
@@ -184,7 +184,6 @@ const IR::Expression *TableStepper::evalTableConstEntries() {
         const auto *actionType = stepper->state.getP4Action(tableAction);
         auto &nextState = stepper->state.clone();
         nextState.markVisited(entry);
-        nextState.markVisited(actionType);
         // Compute the table key for a constant entry
         const auto *hitCondition = TableUtils::computeEntryMatch(*table, *entry, *key);
 
--- a/backends/p4tools/modules/testgen/core/small_step/table_stepper.cpp
+++ b/backends/p4tools/modules/testgen/core/small_step/table_stepper.cpp
@@ -249,7 +249,6 @@ void TableStepper::setTableDefaultEntries(
 
         auto &nextState = stepper->state.clone();
 
-        nextState.markVisited(actionType);
         // We get the control plane name of the action we are calling.
         cstring actionName = actionType->controlPlaneName();
         // Synthesize arguments for the call based on the action parameters.
@@ -321,7 +320,6 @@ void TableStepper::evalTableControlEntries(
         // Try to find the action declaration corresponding to the path reference in the table.
         const auto *actionType = stepper->state.getP4Action(tableAction);
 
-        nextState.markVisited(actionType);
         // We get the control plane name of the action we are calling.
         cstring actionName = actionType->controlPlaneName();
         // Synthesize arguments for the call based on the action parameters.
@@ -511,7 +509,6 @@ void TableStepper::addDefaultAction(std::optional<const IR::Expression *> tableM
     std::stringstream tableStream;
     tableStream << "Table Branch: " << properties.tableName;
     tableStream << " Choosing default action: " << actionPath;
-    nextState.markVisited(actionType);
     nextState.add(*new TraceEvents::Generic(tableStream.str()));
     replacements.emplace_back(new IR::MethodCallStatement(Util::SourceInfo(), tableAction));
     // Some path selection strategies depend on looking ahead and collecting potential
--- a/backends/p4tools/modules/testgen/core/small_step/table_stepper.cpp
+++ b/backends/p4tools/modules/testgen/core/small_step/table_stepper.cpp
@@ -184,6 +184,7 @@ const IR::Expression *TableStepper::evalTableConstEntries() {
         const auto *actionType = stepper->state.getP4Action(tableAction);
         auto &nextState = stepper->state.clone();
         nextState.markVisited(entry);
+        nextState.markVisited(actionType);
         // Compute the table key for a constant entry
         const auto *hitCondition = TableUtils::computeEntryMatch(*table, *entry, *key);
 
--- a/backends/p4tools/modules/testgen/core/small_step/table_stepper.cpp
+++ b/backends/p4tools/modules/testgen/core/small_step/table_stepper.cpp
@@ -248,6 +248,7 @@ void TableStepper::setTableDefaultEntries(
 
         auto &nextState = stepper->state.clone();
 
+        nextState.markVisited(actionType);
         // We get the control plane name of the action we are calling.
         cstring actionName = actionType->controlPlaneName();
         // Synthesize arguments for the call based on the action parameters.
@@ -319,6 +320,7 @@ void TableStepper::evalTableControlEntries(
         // Try to find the action declaration corresponding to the path reference in the table.
         const auto *actionType = stepper->state.getP4Action(tableAction);
 
+        nextState.markVisited(actionType);
         // We get the control plane name of the action we are calling.
         cstring actionName = actionType->controlPlaneName();
         // Synthesize arguments for the call based on the action parameters.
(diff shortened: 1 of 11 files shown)


vlstill:
```bash
# now with --track-coverage STATEMENTS --track-coverage ACTIONS
$ ./p4testgen --target bmv2 --test-backend STF ../backends/p4tools/modules/testgen/targets/bmv2/test/p4-programs/bmv2_table_actions_coverage.p4 --max-tests 256  --path-selection GREEDY_STATEMENT_SEARCH --only-covering-tests --stop-metric MAX_NODE_COVERAGE --track-coverage STATEMENTS --track-coverage ACTIONS
warning: Ingress parser exception handler not fully implemented
============ Test 0: Nodes covered: 0.526316 (10/19) ============
============ Test 1: Nodes covered: 0.684211 (13/19) ============
============ Test 2: Nodes covered: 0.736842 (14/19) ============
============ Test 3: Nodes covered: 0.842105 (16/19) ============
============ Test 4: Nodes covered: 0.894737 (17/19) ============
============ Test 5: Nodes covered: 0.947368 (18/19) ============
============ Test 6: Nodes covered: 1 (19/19) ============

# and without (with just STATEMENTS)
$ ./p4testgen --target bmv2 --test-backend STF ../backends/p4tools/modules/testgen/targets/bmv2/test/p4-programs/bmv2_table_actions_coverage.p4 --max-tests 256  --path-selection GREEDY_STATEMENT_SEARCH --only-covering-tests --stop-metric MAX_NODE_COVERAGE --track-coverage STATEMENTS 
warning: Ingress parser exception handler not fully implemented
============ Test 0: Nodes covered: 0.777778 (7/9) ============
============ Test 1: Nodes covered: 0.888889 (8/9) ============
============ Test 2: Nodes covered: 1 (9/9) ============

File: backends/p4tools/modules/testgen/core/small_step/table_stepper.cpp [hunk 1]
```diff
@@ -184,6 +184,7 @@ const IR::Expression *TableStepper::evalTableConstEntries() {
         const auto *actionType = stepper->state.getP4Action(tableAction);
         auto &nextState = stepper->state.clone();
         nextState.markVisited(entry);
+        nextState.markVisited(actionType);
```
fruffy, line 187: Hmm, this is actually a very specific type of coverage. This does not cover when an action is called directly and not via a table, for example. Is this intended? 

Otherwise, I would just move the markVisited call into the P4Action visitor.

backends/p4tools/modules/testgen/lib/collect_coverable_nodes.cpp
```diff
@@ -122,6 +122,16 @@ bool CoverableNodesScanner::preorder(const IR::ExitStatement *stmt) {
     return true;
 }
 
+bool CoverableNodesScanner::preorder(const IR::P4Action *act) {
+    // Only track actions, which have a valid source position in the P4 program.
+    if (coverageOptions.coverActions && act->getSourceInfo().isValid()) {
+        coverableNodes.insert(act);
+    }
+    // visit body only
```
fruffy, line 130:
```suggestion
    // Visit body only.
```

backends/p4tools/modules/testgen/lib/collect_coverable_nodes.h
```diff
@@ -39,6 +39,9 @@ class CoverableNodesScanner : public Inspector {
     bool preorder(const IR::ExitStatement *stmt) override;
     bool preorder(const IR::MethodCallExpression *call) override;
 
+    /// Actions coverage
```
fruffy, line 42:
```suggestion
    /// Actions coverage.
```

backends/p4tools/modules/testgen/core/small_step/table_stepper.cpp
vlstill: Hmm. I'm not sure how the testgen's midend is configured and if there will be directly-called actions left, I should check it (I presumed there will be none). If there can be actions called directly, it would make sense to handle those too probably (presumably they would be still included in statements coverage because of the call, but it would still make sense to have them in actions coverage too).

A slight problem is that I actually have to mark the action as visited before the step into it is generate as in many places the table stepper creates a new, cloned action out of the original, with the arguments given by the control plane config; https://github.com/p4lang/p4c/blob/47862844a1599e0ca3e8cb31c84be894c083d48f/backends/p4tools/modules/testgen/core/small_step/table_stepper.cpp#L346-L348 In these cases marking the action as visited only when the call is interpreted would yield wrong results I believe.

vlstill: OK, `MoveActionsToTables` is not running in testgen, let me think about a way to handle directly called actions too.

fruffy: >A slight problem is that I actually have to mark the action as visited before the step into it is generate as in many places the table stepper creates a new, cloned action out of the original, with the arguments given by the control plane config; 

That should not be a problem for the markVisited function since we are comparing entries based on the valid source information? If we mismatch or add duplicate entries that looks like a bug.

vlstill:
I've changed this to track all actions calls (I've also changed the test).

```bash
# now with --track-coverage STATEMENTS --track-coverage ACTIONS
$ ./p4testgen  --target bmv2 --test-backend STF ../backends/p4tools/modules/testgen/targets/bmv2/test/p4-programs/bmv2_table_actions_coverage.p4 --max-tests 256  --path-selection GREEDY_STATEMENT_SEARCH --only-covering-tests --stop-metric MAX_NODE_COVERAGE --track-coverage STATEMENTS --track-coverage ACTIONS
warning: Ingress parser exception handler not fully implemented
============ Test 0: Nodes covered: 0.461538 (12/26) ============
============ Test 1: Nodes covered: 0.653846 (17/26) ============
============ Test 2: Nodes covered: 0.692308 (18/26) ============
============ Test 3: Nodes covered: 0.807692 (21/26) ============
============ Test 4: Nodes covered: 0.846154 (22/26) ============
============ Test 5: Nodes covered: 0.923077 (24/26) ============
============ Test 6: Nodes covered: 0.961538 (25/26) ============
============ Test 7: Nodes covered: 1 (26/26) ============

# and without (with just STATEMENTS)
$ ./p4testgen  --target bmv2 --test-backend STF ../backends/p4tools/modules/testgen/targets/bmv2/test/p4-programs/bmv2_table_actions_coverage.p4 --max-tests 256  --
path-selection GREEDY_STATEMENT_SEARCH --only-covering-tests --stop-metric MAX_NODE_COVERAGE --track-coverage STATEMENTS
warning: Ingress parser exception handler not fully implemented
============ Test 0: Nodes covered: 0.615385 (8/13) ============
============ Test 1: Nodes covered: 0.692308 (9/13) ============
============ Test 2: Nodes covered: 0.846154 (11/13) ============
============ Test 3: Nodes covered: 0.923077 (12/13) ============
============ Test 4: Nodes covered: 1 (13/13) ============
```

Note that for freestanding actions, the action calls are statements so they are actually included in the `STATEMENTS` tracking, while the actions called indirectly through tables are not included.

backends/p4tools/modules/testgen/core/small_step/table_stepper.cpp
vlstill: I see. I did not realize we use the source info for identity. In that case this is actually quite simple.

fruffy (approved):

Issue #1270, opened 2018-05-04
Use a struct type as T in value_set<T>


hanw:
Is the following P4 code correct syntax when the type T in PVS is a struct?

```
struct my_packet {
  bit<32> data;
}

struct value_set_t {
  bit<32> field;
}

parser MyParser(packet_in b, out my_packet p, inout my_metadata m, inout standard_metadata_t s) {

    value_set<value_set_t>(4) pvs;

    state start {
        b.extract(p.data);
        transition select({ p.data }) {
            pvs: accept;
            {0x810} : foo;
        }
    }

    state foo {
        transition accept;
    }
}
```

When I compile with p4test at commit d50facb5, type unification failed at the constant key expression {0x810}.

```
error: SelectExpression: Cannot unify tuple type Tuple(1) with non tuple-type int
        transition select({ p.data.da }) {
```

Complete test program below:
[pvs.p4.gz](https://github.com/p4lang/p4c/files/1975923/pvs.p4.gz)

jafingerhut:
Do you get a similar error message if the struct contains 2 fields, and the transition select expression is a list of 2 expressions with type `bit<W>`?

Or does the error only occur for a struct with 1 field, and a list of 1 value?

mbudiu-vmw:
We should support the second variant: the most natural one for the user, even if the compiler has to do a bit more work. I will tag these as enhancements, because they are in line with the current spec; we should update the spec to allow struct types for value sets.

hanw:
Updated text and example in spec to allow struct type for the value set as part of https://github.com/p4lang/p4-spec/pull/602 https://github.com/p4lang/p4-spec/pull/602/commits/278afae9a3db7f495c474768ed9bf6575679e6e9

PR #4309, opened 2023-12-22
Fix warnings on signed integer initialization

AndrzejSawula:
Remove misleading overflow warning for correct-width signed integer literals with MSBit set, i.e. representing negative numbers. This is actually the only way integer literals of explicitly-declared width can be specified (as per language specs, Section 6.3.3.2).

As these warnings are mentioned in the language specs, Section 7.1.6.6, this change is accompanied with one to the specs content, PR [#1270](https://github.com/p4lang/p4-spec/pull/1270).

### DIFF
--- a/ir/expression.cpp
+++ b/ir/expression.cpp
@@ -90,17 +90,32 @@ void IR::Constant::handleOverflow(bool noWarning) {
     big_int mask = Util::mask(width);
 
     if (tb->isSigned) {
+        auto masked_value = value;
+        bool is_overflow = false;
         big_int max = (one << (width - 1)) - 1;
         big_int min = -(one << (width - 1));
-        if (value < min || value > max) {
+
+        if (value >= 0) {
+            if (value != (value & mask)) {
+                is_overflow = true;
+                masked_value = value & mask;
+            }
+            if (masked_value > max) masked_value -= (one << width);
+        } else {
+            if (value < min) {
+                is_overflow = true;
+                masked_value = value & mask;
+            }
+        }
+        if (is_overflow) {
             if (!noWarning)
                 ::warning(ErrorType::WARN_OVERFLOW, "%1%: signed value does not fit in %2% bits",
                           this, width);
             LOG2("value=" << value << ", min=" << min << ", max=" << max << ", masked="
-                          << (value & mask) << ", adj=" << ((value & mask) - (one << width)));
-            value = value & mask;
-            if (value > max) value -= (one << width);
+                          << (value & mask) << ", adj=" << ((value & mask) - (one << width))
+                          << ", processed value=" << masked_value);
         }
+        value = masked_value;
     } else {
         if (value < 0) {
             if (!noWarning)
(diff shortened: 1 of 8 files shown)


ir/expression.cpp
```diff
@@ -90,17 +90,32 @@ void IR::Constant::handleOverflow(bool noWarning) {
     big_int mask = Util::mask(width);
 
     if (tb->isSigned) {
+        auto masked_value = value;
```
fruffy, line 93: The logic here looks fairly complex. Possible to add some tests to https://github.com/p4lang/p4c/blob/main/test/gtest/constant_expr_test.cpp which exercise this code?

ir/expression.cpp
```diff
@@ -90,17 +90,32 @@ void IR::Constant::handleOverflow(bool noWarning) {
...
+                masked_value = value & mask;
+            }
+            if (masked_value > max) masked_value -= (one << width);
+        } else {
+            if (value < min) {
+                is_overflow = true;
+                masked_value = value & mask;
+            }
+        }
+        if (is_overflow) {
```
ChrisDodd, line 110: This big block of code seems to be equivalent to just changing one line

    big_int max = (one << width) -1;

fruffy: Wouldn't this need to be adjusted based on signedness (`value >= 0`, `value <0`)?

@asawulaINTC Can you comment?

test/gtest/constant_expr_test.cpp
```diff
@@ -141,4 +141,43 @@ TEST_F(ConstantExpr, TestIntegerFuncs) {
...
+    auto c2 = IR::Constant(&t7s, 0xff);
+    EXPECT_EQ(c2.asInt(), -1);
+
+    // overflow, but this should not even generate a warning under regular settings
+    auto c3 = IR::Constant(&t7u, 0xff);
+    EXPECT_EQ(c3.asUnsigned(), 127);
+
+    auto c4 = IR::Constant(&t13s, 0x1555);  // 0b1_0101_0101_0101
+    EXPECT_EQ(c4.asInt(), -2731);
+    EXPECT_ANY_THROW(c4.asUnsigned());  // will also print error message
```
fruffy, line 170: Does this work? Alternative, it should be possible to get the number of warnings triggered with `diagnosticCount()`.

fruffy (commented):
Thanks for the tests.

jonathan-dilorenzo:
It looks like this didn't come up yet in the corresponding spec change, but, based on discussions in the LDWG, I think we'd prefer to keep the warnings on signed integer initialization unless you have a use-case that doesn't fit our imagination in mind. Let me add a comment to the specification issue: github.com/p4lang/p4-spec/issues/1269

ChrisDodd:
I've been experimenting a bit with this and another implementation issue I see is that it does not appear to be possible to stop specific warnings with explicit casts -- that is, something like `(bit<8>)10w256` will give a warning about 256 being out of range for a `bit<8>`.  This seems wrong (and will cause problems when using `-Werror` if you have code that wants the expected modulo behavior for the code).  This should probably be a separate bug filed against p4c.

PR #4314, opened 2024-01-04
Rename --with-output-packet to --output-packet-only.

fruffy-g:
Communicate the intent of the option more clearly.


smolkaj (approved):
Nice, LGTM :)

Issue #4315, opened 2024-01-04
MoveInitializers drops the srcInfo when moving to the top of a P4Control body


grg:
The MoveInitializers pass loses the P4Control->body->srcInfo when moving initializers.

PR #4317, opened 2024-01-04
keep P4Control->body->srcInfo during init move

grg:
Ensure that MoveInitializers keeps the srcInfo when creating a new body for a P4Control.

Fix for issue #4315

### DIFF
--- a/test/gtest/frontend_test.cpp
+++ b/test/gtest/frontend_test.cpp
@@ -172,8 +172,7 @@ TEST_F(P4CFrontendEnumValidation, InvalidType) {
 // Tests for MoveInitializers
 struct P4CFrontendMoveInitializers : P4CFrontend {
     P4CFrontendMoveInitializers() {
-        addPasses({new P4::ResolveReferences(&refMap),
-                   new P4::MoveInitializers(&refMap)});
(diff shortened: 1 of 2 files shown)


test/gtest/frontend_test.cpp
```diff
@@ -168,4 +169,39 @@ TEST_F(P4CFrontendEnumValidation, InvalidType) {
...
+TEST_F(P4CFrontendMoveInitializers, P4ControlSrcInfo) {
+    std::string program = P4_SOURCE(R"(
+        control m() {
+            bool blah = false;
+            apply{}
+        }
+    )");
+    RedirectStderr errors;
+    const auto *prog = parseAndProcess(program);
+    errors.dumpAndReset();
```
vlstill, line 191: You don't need to redirect stderr since it is not inspeced.
```suggestion
    const auto *prog = parseAndProcess(program);

vlstill (approved):

amreshk (approved):

hanw (approved):

test/gtest/frontend_test.cpp
grg: Good catch. I had copy/pasted code from the other tests.

PR #4319, opened 2024-01-05
Allow constructing PassRepeated from a PassManager

vlstill:
... and add `explicit` to the constructors that could be accidentally invoced by implicit casts otherwise


grg (approved):
LGTM.

Is the explicit to prevent accidental casts on the repeats argument? Or are there cases where an accidental cast could occur on the first argument? (Can something be cast to a PassManager?)

PR #4320, opened 2024-01-05
Respect ENABLE_LTO even for static build, but keep LTO default for static

vlstill:
Up until now, the ENABLE_LTO option was ignored for static builds. This can be quite confusing. With this change, the static now defaults to using LTO, but it is possible to override it explicitly.

Also, the check for GCC 8+ for LTO is no longer necessary as p4c supports only GCC 9+.


CMakeLists.txt
```diff
@@ -282,19 +289,15 @@ if (ENABLE_LTO AND CMAKE_CXX_COMPILER_ID STREQUAL "GNU")
     message(WARNING "LLD does not work with GCC's LTO object format, switching to Gold.")
     set(BUILD_LINK_WITH_LLD OFF)
   endif ()
-  add_cxx_compiler_option  ("-flto")
+  if (CMAKE_CXX_COMPILER_VERSION VERSION_LESS 8.0)
```
davidbolvansky, line 292: No need to worry about as GCC 9 or newer is required by p4c anyway

davidbolvansky (approved):
Thanks

CMakeLists.txt
vlstill: True, I may as well remove this check...

PR #4323, opened 2024-01-10
[P4Testgen] Extend the CompilerTarget runProgram function with data structures which can pass on more information.

fruffy:
Major rewrite of the ProgramInfo and CompilerTarget infrastructure. The basic idea is to compute all necessary static analysis information in the compilation phase, then pass this information forward to ProgramInfo.

For this, we introduce a CompilerResult object. This object is produced in the compilation phase. Any P4Tools module can add arbitrary information to this class. The results object is then passed to ProgramInfo.

This PR is part 1. It introduces the `CompilerResult` data structure, which records information produces by the compiler which runs before P4Testgen is invoked. This includes the P4 program after the mid end for now.

### DIFF
--- a/backends/p4tools/common/compiler/compiler_target.h
+++ b/backends/p4tools/common/compiler/compiler_target.h
@@ -31,6 +31,10 @@ class CompilerResult : public ICastable {
     [[nodiscard]] const IR::P4Program &getProgram() const;
 };
 
+/// P4Tools compilers may return an error instead of a compiler result.
+/// This is a convenience definition for the return value.
+using CompilerResultOrError = std::optional<std::reference_wrapper<const CompilerResult>>;
+
 /// Encapsulates the details of invoking the P4 compiler for a target device and architecture.
 class CompilerTarget : public Target {
  public:
@@ -46,25 +50,25 @@ class CompilerTarget : public Target {
     /// program.
     ///
     /// @returns std::nullopt if an error occurs during compilation.
-    static std::optional<const CompilerResult *> runCompiler();
+    static CompilerResultOrError runCompiler();
 
     /// Runs the P4 compiler to produce an IR and other information for the given source code.
     ///
     /// @returns std::nullopt if an error occurs during compilation.
-    static std::optional<const CompilerResult *> runCompiler(const std::string &source);
+    static CompilerResultOrError runCompiler(const std::string &source);
 
  private:
     /// Runs the front and mid ends on the given parsed program.
     ///
     /// @returns std::nullopt if an error occurs during compilation.
-    static std::optional<const CompilerResult *> runCompiler(const IR::P4Program *);
+    static CompilerResultOrError runCompiler(const IR::P4Program *);
 
  protected:
     /// @see @makeContext.
     [[nodiscard]] virtual ICompileContext *makeContextImpl() const;
 
     /// @see runCompiler.
-    virtual std::optional<const CompilerResult *> runCompilerImpl(const IR::P4Program *) const;
+    virtual CompilerResultOrError runCompilerImpl(const IR::P4Program *) const;
 
     /// This implementation just forwards the given arguments to the compiler.
     ///
--- a/backends/p4tools/common/compiler/compiler_target.h
+++ b/backends/p4tools/common/compiler/compiler_target.h
@@ -1,6 +1,7 @@
 #ifndef BACKENDS_P4TOOLS_COMMON_COMPILER_COMPILER_TARGET_H_
 #define BACKENDS_P4TOOLS_COMMON_COMPILER_COMPILER_TARGET_H_
 
+#include <functional>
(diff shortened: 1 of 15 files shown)


backends/p4tools/common/compiler/compiler_target.h
```diff
@@ -25,28 +42,29 @@ class CompilerTarget : public Target {
     /// @returns any unprocessed arguments, or nullptr if there was an error.
     static std::vector<const char *> *initCompiler(int argc, char **argv);
 
-    /// Runs the P4 compiler to produce an IR.
+    /// Runs the P4 compiler to produce an IR and various other kinds of information on the input
+    /// program.
     ///
     /// @returns std::nullopt if an error occurs during compilation.
-    static std::optional<const IR::P4Program *> runCompiler();
+    static std::optional<const CompilerResult *> runCompiler();
```
vlstill, line 49: Is there any reason to return optional of a pointer? I don't see `std::nullopt` meaning something different than optional with `nullptr` (and if it did it would be quite confusing I think). I suggest using `std::optional<CompilerResult>` as the struct will likely mainly contain pointers/references and it will not be copied much around so this will matter even if it become larger.

backends/p4tools/modules/testgen/testgen.cpp
```diff
@@ -91,12 +90,12 @@ int generateAbstractTests(const TestgenOptions &testgenOptions, const ProgramInf
...
 }
 
-int Testgen::mainImpl(const IR::P4Program *program) {
+int Testgen::mainImpl(const CompilerResult &compilerResult) {
     // Register all available testgen targets.
     // These are discovered by CMAKE, which fills out the register.h.in file.
     registerTestgenTargets();
 
-    const auto *programInfo = TestgenTarget::initProgram(program);
+    const auto *programInfo = TestgenTarget::initProgram(&compilerResult.getProgram());
```
vlstill, line 98: I presume the next step would be to actually pass the `compilerResult` to `TestgenTarget::initProgram`. Would it make sense to do this change already now? Without that, this does not actually change anything...

vlstill (commented):
It would be really good to have more info about how this would be used and what would be the next steps. At this point, this just introduces a data structure that is not used in the end. I presume #4324 is the next step -- I'll have a look at that too as I can't really evaluate this one in isolation.

backends/p4tools/modules/testgen/testgen.cpp
fruffy: I initially had all changes bundled up in one PR, but decided to split the PR because the amount files touched grew quite large. #4324 alone touches ~30 files.

backends/p4tools/common/compiler/compiler_target.h
fruffy: The pointer is necessary here because CompilerResult will be subclassed by the target back ends, e.g., BMv2CompilerResult etc. It could also be a reference, but that would require a reference_wrapper. 

I much prefer using `std::optional` over nullptr these days because it makes it very explicit that an error has happened. Going forward it will also be easier to replace this with std::expect or absl:StatusOr.

vlstill: hmm, I did not realize that `CompilerResult` will be subclassed, that does make a lot of sense. I don't like that `std::optional<pointer>` solution much because it has the ugly state where the optional contains `nullptr`. You are right that optional is more explicit about having to resolve the error though :-/ (and `std::optional<std::reference_wrapper<X>>` looks ugly and has the additional disadvantage that `auto x = opt.value()` behaves weirdly.

vlstill: I'm starting to think that `std::optional<std::reference_wrapper<const CompilerResult>>` is the right option here if you want the added safety of optional. This is a bit uglier to use (the type is long, and one needs to explicitly call `.get()` in many cases). But it is unlikely the `auto` problem I mentioned above will cause problems exactly because one has to call `get()` to get the actual instance. And having `std::optional<const CompilerResult *>(nullptr)` would be really confusing. Too bad std does not allow `std::optional<const CompilerResult &>` :-/.

backends/p4tools/modules/testgen/test/gtest_utils.cpp
```diff
@@ -31,7 +31,7 @@ std::optional<const P4ToolsTestCase> P4ToolsTestCase::create(
     if (!compilerResults.has_value()) {
         return std::nullopt;
     }
-    return P4ToolsTestCase{compilerResults.value()->getProgram()};
+    return P4ToolsTestCase(compilerResults.value().get().getProgram());
```
vlstill, line 34: Since you have already checked for nullopt, you can use `compilerResults->get()` if you want (`operator->` and `operator*` are unchecked, unlike `value()`).

vlstill (approved):

PR #4324, opened 2024-01-10
[P4Testgen] Refactor ProgramInfo to be initialized with CompilerResult instead of a IR::P4Program

fruffy:
Builds on #4323 and introduces the CompilerResult data structure to the ProgramInfo class.

It also introduces the BMv2CompilerResult, which also contains the P4Rutime API for the program. This API is computed in the mid end for this target.

### DIFF
--- a/backends/p4tools/modules/testgen/targets/bmv2/bmv2.h
+++ b/backends/p4tools/modules/testgen/targets/bmv2/bmv2.h
@@ -7,13 +7,27 @@
 
 namespace P4Tools::P4Testgen::Bmv2 {
 
+/// Extends the CompilerResult with the associated P4RuntimeApi
+class BMv2V1ModelCompilerResult : public CompilerResult {
+ private:
+    /// The runtimeAPI inferred from this particular BMv2 V1Model P4 program.
+    P4::P4RuntimeAPI p4runtimeApi;
+
+ public:
+    explicit BMv2V1ModelCompilerResult(const IR::P4Program &program, P4::P4RuntimeAPI p4runtimeApi);
+
+    [[nodiscard]] const P4::P4RuntimeAPI &getP4RuntimeApi() const;
+};
+
 class Bmv2V1ModelCompilerTarget : public CompilerTarget {
  public:
     /// Registers this target.
     static void make();
 
  private:
-    MidEnd mkMidEnd(const CompilerOptions &options) const override;
+    [[nodiscard]] MidEnd mkMidEnd(const CompilerOptions &options) const override;
+
+    CompilerResultOrError runCompilerImpl(const IR::P4Program *program) const override;
 
     Bmv2V1ModelCompilerTarget();
 };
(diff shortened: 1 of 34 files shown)


backends/p4tools/modules/testgen/targets/bmv2/bmv2.h
```diff
@@ -7,13 +7,31 @@
 
 namespace P4Tools::P4Testgen::Bmv2 {
 
+/// Extends the CompilerResult with the associated P4RuntimeApi
+class BMv2V1ModelCompilerResult : public CompilerResult {
+ private:
+    /// The runtimeAPI inferred from this particular BMv2 V1Model P4 program.
+    P4::P4RuntimeAPI p4runtimeApi;
```
vlstill, line 14: Is this not used at this point?

vlstill (commented):
Looks good, one question...

backends/p4tools/modules/testgen/targets/bmv2/bmv2.h
fruffy: Yes, I think this is an artifact of the split. I could try to move this into the children PRs.

vlstill: While it would be better not leaving this here, I could also live with it merged in this PR.

backends/p4tools/modules/testgen/targets/bmv2/bmv2.h
```diff
@@ -7,13 +7,30 @@
...
+    [[nodiscard]] const P4::P4RuntimeAPI &getP4RuntimeApi() const;
+};
+
 class Bmv2V1ModelCompilerTarget : public CompilerTarget {
  public:
     /// Registers this target.
     static void make();
 
+    /// @returns the singleton instance for the current target.
+    static const Bmv2V1ModelCompilerTarget &get();
```
vlstill, line 28: The singleton was not there before?

backends/p4tools/modules/testgen/targets/bmv2/test/BMV2MetadataXfail.cmake
```diff
@@ -69,6 +69,12 @@ p4tools_add_xfail_reason(
   bmv2_assert.p4
 )
 
+p4tools_add_xfail_reason(
+  "testgen-p4c-bmv2-metadata"
+  "Unsupported type argument for Value Set"
```
vlstill, line 74: What happend to these tests? Are they broken by this PR somehow, or were they not really working before too and just now started actually failing? Is this something part 3 will resolve?

fruffy: `p4c-bm2-ss` throws the same error when invoked on the program. It means that this P4 program does not produce a valid P4Runtime API:
```
build/p4c-bm2-ss --target bmv2 --arch v1model --p4runtime-files out.txt testdata/p4_16_samples/pvs-nested-struct.p4
testdata/p4_16_samples/pvs-nested-struct.p4(22): [--Werror=unsupported] error: Unsupported type argument for Value Set; this version of P4Runtime requires that when the type parameter of a Value Set is a struct, all the fields of the struct must be of type bit<W>, but inner is not
  inner_value_set_t inner;
  ```
We could avoid the error by not invoking the P4RuntimeSerializer for certain test back ends, but you will still run into a similar problem when trying to run the program. The Xfail already existed for the PTF tests, for instance.

vlstill (approved):

PR #4325, opened 2024-01-10
Allow constant-folding of arbitrary-precision integer casts

vlstill:
This showed as bug in testgen, but the core of the issue is that the constant folder was not handling `(int)` cast. That is fixed by this PR.


mihaibudiu:
No tests?

vlstill:
> No tests?

Covered by existing tests (54f51b0).

mihaibudiu (approved):

PR #4328, opened 2024-01-10
Documentation & testfix follow-up to #4160

thomascalvert-xlnx:
Split this out of #4327 since it's really not part of those changes.

@fruffy

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,8 +24,9 @@ The code contains seven sample backends:
 * p4c-dpdk: can be used to target the DPDK software switch (SWX) pipeline
   https://doc.dpdk.org/guides/rel_notes/release_20_11.html,
 * p4c-ebpf: can be used to generate C code which can be compiled to [eBPF](https://en.wikipedia.org/wiki/Berkeley_Packet_Filter)
-  and then loaded in the Linux kernel. The eBPF backend currently implements two architecture models:
-  [ebpf_model.p4 for packet filtering](./backends/ebpf/README.md) and [the fully-featured PSA (Portable Switch Architecture) model](./backends/ebpf/psa/README.md).
+  and then loaded in the Linux kernel. The eBPF backend currently implements three architecture models:
+  [ebpf_model.p4 for packet filtering, xdp_model.p4 for XDP](./backends/ebpf/README.md) and
+  [the fully-featured PSA (Portable Switch Architecture) model](./backends/ebpf/psa/README.md).
 * p4test: a source-to-source P4 translator which can be used for
   testing, learning compiler internals and debugging,
 * p4c-graphs: can be used to generate visual representations of a P4 program;
--- a/backends/ebpf/README.md
+++ b/backends/ebpf/README.md
@@ -1,16 +1,26 @@
 # eBPF Backend
 
-The back-end accepts only P4_16 code written for the `ebpf_model.p4`
-filter model.  It generates C code that can be afterwards compiled
-into eBPF (extended Berkeley Packet Filters
-https://en.wikipedia.org/wiki/Berkeley_Packet_Filter) using clang/llvm
-or bcc (https://github.com/iovisor/bcc.git).
+The back-end accepts only P4_16 code written for the `ebpf_model.p4` or
+`xdp_model.p4` filter models.  It generates C code that can be afterwards
+compiled into eBPF (extended Berkeley Packet Filters
+https://en.wikipedia.org/wiki/Berkeley_Packet_Filter) using clang/llvm or
+bcc (https://github.com/iovisor/bcc.git).
 
 An older version of this compiler for compiling P4_14 is available at
 https://github.com/iovisor/bcc/tree/master/src/cc/frontends/p4
 
 Identifiers starting with ebpf_ are reserved in P4 programs, including
 for structure field names.
+
+## Target architectures
+
+The `ebpf_model.p4` target is a classifier-only: the program returns a
+boolean which controls whether the packet is passed or dropped. In P4
+terms, this means there is no deparser.
+
+The `xdp_model.p4` target adds packet editing support, and is meant to
+replicate the capabilities of the Linux kernel's XDP environment. It
+can be viewed as an extension of the previous model which adds a deparser.
 
 ## Background
 
@@ -229,10 +239,6 @@ that the compiler's capabilities will improve gradually.
 
 Here are some limitations imposed on the P4 programs:
 
-* this architecture only supports packet filters: the control block
-  returns a boolean value which indicates whether a packet is
-  forwarded or dropped
-
 * arbitrary parsers can be compiled, but the BCC compiler will reject
   parsers that contain cycles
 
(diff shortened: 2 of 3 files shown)


backends/ebpf/README.md
```diff
@@ -1,17 +1,27 @@
 # eBPF Backend
 
-The back-end accepts only P4_16 code written for the `ebpf_model.p4`
-filter model.  It generates C code that can be afterwards compiled
-into eBPF (extended Berkeley Packet Filters
-https://en.wikipedia.org/wiki/Berkeley_Packet_Filter) using clang/llvm
-or bcc (https://github.com/iovisor/bcc.git).
+The back-end accepts only P4_16 code written for the `ebpf_model.p4` or
+`xdp_model.p4` filter models.  It generates C code that can be afterwards
```
fruffy, line 4: Is this accurate? I do not think this works unless we integrate https://github.com/vmware-archive/p4c-xdp?

README.md
```diff
@@ -23,8 +23,9 @@ The code contains seven sample backends:
 * p4c-dpdk: can be used to target the DPDK software switch (SWX) pipeline
   https://doc.dpdk.org/guides/rel_notes/release_20_11.html,
 * p4c-ebpf: can be used to generate C code which can be compiled to [eBPF](https://en.wikipedia.org/wiki/Berkeley_Packet_Filter)
-  and then loaded in the Linux kernel. The eBPF backend currently implements two architecture models:
-  [ebpf_model.p4 for packet filtering](./backends/ebpf/README.md) and [the fully-featured PSA (Portable Switch Architecture) model](./backends/ebpf/psa/README.md).
+  and then loaded in the Linux kernel. The eBPF backend currently implements three architecture models:
+  [ebpf_model.p4 for packet filtering, xdp_model.p4 for XDP](./backends/ebpf/README.md) and
```
fruffy, line 27: `xdp_model.p4`, which is an extension, is not part of the compiler yet. 
https://github.com/vmware-archive/p4c-xdp

It is sensible to merge it in at this point.

backends/ebpf/run-ebpf-test.py
```diff
@@ -196,7 +196,10 @@ def run_test(options, argv):
                 exp_stderr = f.read().strip()
             # Output will have full path which differs
             out = out.replace(options.p4filename, basename).strip()
-            # Output will have one extra line with make error, trim that
+            # Output may start with annoying "jobserver unavailable" make warning, trim that
+            if out[0:5] == "make[":
```
fruffy, line 200: This looks strange to me, what sort of failure produces this error? Can we fix the source of the problem?

backends/ebpf/README.md
thomascalvert-xlnx: The PR #4160 effectively did that. See for example this test case: https://github.com/p4lang/p4c/blob/main/testdata/p4_16_samples/xdp_dec_ttl_ebpf-kernel.p4

README.md
thomascalvert-xlnx: As above, it's supported now. The [`xdp_model.p4`](https://github.com/p4lang/p4c/blob/main/p4include/xdp_model.p4) file is part of this repo, and the eBPF backend autodetects to target here:

https://github.com/p4lang/p4c/blob/main/backends/ebpf/ebpfProgram.cpp#L33C37-L55

backends/ebpf/run-ebpf-test.py
thomascalvert-xlnx: I tried quite hard to fix the source but it seems like there is no way to force CMake to do the right thing, which is why we end up with this kludge.

The issue is that, when running `make -j4 check` (notice the `-j`) argument, make emits a "jobserver unavailable" warning which thus causes the output to not match.

https://stackoverflow.com/questions/33171335/jobserver-unavailable-when-building-external-projects-with-cmake

README.md
fruffy: Uff, I completely missed that, even though I even commented on the PR. I guess because I checked for the `xdp_target` and extension instead. https://github.com/vmware-archive/p4c-xdp/blob/master/xdp_target.py

fruffy: It would be nice if we could also port all the existing tests from the old repository. But this we can do in a different PR.

backends/ebpf/run-ebpf-test.py
fruffy: If I understand this right, CI might fail because in the error output there is a spurious `jobserver unavailable` in the output, which mismatches with the golden tests?

And we get this spurious output because we compile the generated C code with `make -jX`? Can we add this as a comment to clarify?

fruffy: We do use `-jX` (https://github.com/p4lang/p4c/blob/main/.github/workflows/ci-test.yml#L21) but CI nodes only have 2 cores iirc. 

Even so, the output that is checked here comes from the eBPF test and make, which is triggered in functions like this one:
https://github.com/p4lang/p4c/blob/main/backends/ebpf/targets/target.py#L63
https://github.com/p4lang/p4c/blob/main/backends/ebpf/targets/kernel_target.py#L43

thomascalvert-xlnx: I was talking about running `make -j4 check` to run the test suite -- it seems like the CI calls `ctest` directly, which may be why we're seeing different behaviours (because we no longer have `make` as a parent process).

The make calls in the python target files are the ones which emit the warning, yes.

fruffy: Looks like `make -j4 check` causes subprocesses that use make to also run in parallel. Might be worthwhile to simply fix the make calls of the script to be single-threaded. The parallelism is not intended.

thomascalvert-xlnx: What you're saying is a simple `make check` runs in parallel already, so running `make -j4 check` is invalid? I guess we could just drop this change then.

thomascalvert-xlnx: I did try adding `-j1` there and it didn't suppress the warning.

Honestly I'm now of the opinion that it's on me for running `make check` with `-j4` and I should just stop doing that. We have 3 options:

1. Keep this patch as it is (doesn't hurt as far as I can tell) but add some more comments to clarify the issue.
2. Try to fix the root cause in the CMake-Make-Python-Make chain.
3. Drop this patch since it doesn't fix anything in CI anyway.

fruffy: Naturally, I prefer 2 (I believe we need to add `-j1` to all make calls in the script?) but I am also okay with option 1 as long as we clearly document the cause, as you also pointed out.

README.md
fruffy: Can we create an issue to port the old XDP tests to the compiler, so that we do not forget to  track this?

thomascalvert-xlnx: Done: #4337

backends/ebpf/run-ebpf-test.py
thomascalvert-xlnx: I can't figure it out - just adding `-j1` does not solve it. In the meanwhile I've improved the comment here.

fruffy (approved):

fruffy:
I can merge this once the PR is rebased.

PR #4329, opened 2024-01-11
P4TC - Support const entries

komaljai:


backends/tc/backend.cpp
```diff
@@ -271,6 +271,53 @@ void ConvertToBackendIR::postorder(const IR::P4Action *action) {
     }
 }
 
+void ConvertToBackendIR::updateConstEntries(const IR::P4Table *t, IR::TCTable *tabledef) {
+    // Check if there are const entries.
+    auto entriesList = t->getEntries();
+    if (entriesList == nullptr) return;
+    auto keys = t->getKey();
+    for (auto e : entriesList->entries) {
+        if (keys == nullptr) {
```
usha1830, line 280: This if can be moved outside for as it is obtained outside the loop only

backends/tc/backend.cpp
```diff
@@ -271,6 +271,53 @@ void ConvertToBackendIR::postorder(const IR::P4Action *action) {
...
+            return;
+        }
+        auto keyset = e->getKeys();
+        if (keyset->components.size() != keys->keyElements.size()) {
+            ::error("No of keys in const_entries should be same as no of keys in the table.");
```
usha1830, line 285: Please add error type.

usha1830 (approved):
LGTM!

backends/tc/backend.cpp
komaljai: addressed

backends/tc/backend.cpp
komaljai: addressed

PR #4332, opened 2024-01-12
Ignore clang-tidy complaints about macro do-while loops.

fruffy:
Fix a minor annoyance. We can't change the macros we currently use, which expand to do-while loops.


vlstill (approved):
Ok. Although I don't see much reason to avoid do-while even in the general case.

PR #4334, opened 2024-01-15
Remove P4C options from gtest binary.

fruffy:
The compiler gtest were always hard to use because the compiler options class shadows the actual gtest parameters. The compiler options are never used, so remove them. 

Now we can use flags such as `--gtest_filter=` to select for a particular gtest.

### DIFF
--- a/test/gtest/gtestp4c.cpp
+++ b/test/gtest/gtestp4c.cpp
@@ -14,17 +14,12 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-#include "frontends/common/options.h"
 #include "gtest/gtest.h"
 #include "helpers.h"
-#include "lib/log.h"
 
 int main(int argc, char **argv) {
     ::testing::InitGoogleTest(&argc, argv);
     AutoCompileContext autoGTestContext(new GTestContext);
-
-    GTestContext::get().options().process(argc, argv);
-    if (diagnosticCount() > 0) return -1;
 
     // Initialize the global test environment.
     (void)P4CTestEnvironment::get();


test/gtest/gtestp4c.cpp
```diff
@@ -14,18 +14,13 @@ See the License for the specific language governing permissions and
...
-#include "frontends/common/options.h"
 #include "gtest/gtest.h"
 #include "helpers.h"
-#include "lib/log.h"
 
 int main(int argc, char **argv) {
     ::testing::InitGoogleTest(&argc, argv);
     AutoCompileContext autoGTestContext(new GTestContext);
 
-    GTestContext::get().options().process(argc, argv);
```
vlstill, line 26: Can we also remove GTestContext now? Or is a context needed for some of the tests to work?

fruffy: Without GTestContext you will get something like `CompileContextStack has an empty CompileContext stack`. 

Would love to get rid of the dependency on compile context at some point. It makes it more difficult to develop a self-contained library API for the compiler.

vlstill: I see, OK.

> It makes it more difficult to develop a self-contained library API for the compiler.

I guess there the problem is really that the context is a singleton, not that there is a context.

fruffy: Yup, there are a bunch of these singletons scattered around and several utility functions depend on it. I believe they are largely used to keep track of errors and warnings.

vlstill (approved):

PR #4336, opened 2024-01-15
Add missing equiv implementation to Type_Any.

fruffy:
`Type_Any` has a `declid` member but does not have a corresponding equiv function that ignores. All other IR nodes with `declid` have this implementation, it seems it was forgotten here. Fix it.


vlstill (approved):
Looks consistent to the rest of the IR to me.

BTW. is it documented somewhere what is the meaning behind `equiv` and `==` since they are clearly different?

PR #4338, opened 2024-01-19
Convert deparser header to 'inout' and address casting for functions with control block parameters 

komaljai:
Functions that are receiving parameters in the control block signature are not being casted correctly.
```
void ipip_push(inout headers_t hdr, in metadata_t meta) { ... }
```

For such functions the compiler is assuming the BPF generated code are plain structs.
For these two parameters we are actually using pointers.


usha1830 (approved):
LGTM

PR #4340, opened 2024-01-19
[P4Testgen] Introduce the option to produce lowercase hex for the formatting library.

fruffy-g:
Also add an options structure to the library

### DIFF
--- a/backends/p4tools/common/lib/format_int.h
+++ b/backends/p4tools/common/lib/format_int.h
@@ -10,6 +10,14 @@
 
 namespace P4Tools {
 
+/// Defines common formatting options.
+struct FormatOptions {
+    bool useSep = false;
+    bool pad = true;
+    bool usePrefix = true;
+    bool writeUpperCaseHex = true;
+};
+
 /// Format @arg value as either binary or hexadecimal string.
 /// Hexadecimal strings are prefixed with "0x", whereas binary strings are prefixed with "0b". If
 /// @arg useSep is true, the output is formatted in groups of four digits separated by the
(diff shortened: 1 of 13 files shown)


backends/p4tools/common/lib/format_int.h
```diff
@@ -10,6 +10,14 @@
 
 namespace P4Tools {
 
+/// Defines common formatting options.
+struct FormatOptions {
+    bool useSep = false;
+    bool pad = true;
+    bool usePrefix = true;
+    bool writeUpperCaseHex = true;
```
smolkaj, line 18: Nit: any reason why we say "write" here, but "use" for some of the other options? If not, shall we be consistent?

backends/p4tools/common/lib/trace_event_types.cpp
```diff
@@ -58,7 +58,7 @@ const Expression *Expression::evaluate(const Model &model, bool doComplete) cons
 
 void Expression::print(std::ostream &os) const {
     Generic::print(os);
-    os << ": " << formatHexExpr(value, true);
+    os << ": " << formatHexExpr(value, {true});
```
smolkaj, line 61: `{true}`: This isn't great for code readability, use designated initializers instead?
```
{.useSep = true}
```

It is technically a C++20 feature, so may not be supported?

Ditto a couple of more times, throughout.

smolkaj (approved):
Very nice :)

backends/p4tools/common/lib/trace_event_types.cpp
fruffy-g: Yes, the compiler still uses C++17. :/  But I do not think there is any reason why we can't move to C++20. Would unlock some nice things.

vlstill: My main concern is if this requires bumping the minimal required version of GCC. If so, that could be problematic for downstream tools, or even for users of p4c directly. Not much of C++20 is supported in GCC 9. Designated initializers are, because they were GNU extension for a long time... But this should be probably discussed elsewhere.

PR #4341, opened 2024-01-20
Allow unknown fields when parsing P4Info files

jafingerhut:


jafingerhut:
This PR is the same as https://github.com/p4lang/p4c/pull/4129 except it is up to date with latest main version of p4c.

jafingerhut:
I have checked the protobuf source repo here https://github.com/protocolbuffers/protobuf looking through several versions of it to see which versions support the `allow_unknown_field` optional parameter to the Python method `google.protobuf.text_format.Merge`, and which do not.

I believe that this optional parameter was added in v3.8.0 of Protobuf.  I would guess that the reason this p4c CI test:

+ test-p4c / test-ubuntu20 (Unity ON)

is failing, is because it is using a version of Protobuf older than that, but I am not sure where to verify that guess.

fruffy:
> I have checked the protobuf source repo here https://github.com/protocolbuffers/protobuf looking through several versions of it to see which versions support the `allow_unknown_field` optional parameter to the Python method `google.protobuf.text_format.Merge`, and which do not.
> 
> I believe that this optional parameter was added in v3.8.0 of Protobuf. I would guess that the reason this p4c CI test:
> 
>     * test-p4c / test-ubuntu20 (Unity ON)
> 
> 
> is failing, is because it is using a version of Protobuf older than that, but I am not sure where to verify that guess.

What happens if you remove this check [here](https://github.com/p4lang/p4c/blob/main/tools/ci-build.sh#L135)? It might be that the preinstalled Protobuf version is too old.

fruffy (approved):
Seems like it works. Would be great to add these dependencies to the [requirements.txt](https://github.com/p4lang/p4c/blob/main/requirements.txt) file. But this might break Ubuntu 18.04, which  we still run as a nightly build.

jafingerhut:
> Seems like it works. Would be great to add these dependencies to the [requirements.txt](https://github.com/p4lang/p4c/blob/main/requirements.txt) file. But this might break Ubuntu 18.04, which we still run as a nightly build.

Commit 4 of this PR I added the versions of protobuf, grpcio and googleapis-common-protos packages to requirements.txt.  We will see soon from CI tests what happens.

jafingerhut:
@fruffy Where can one find the results of nightly Ubuntu 18.04 builds?

jafingerhut:
Or if we want to take a chance, I can merge this PR and then we'll see tomorrow what the nightly results are, and backo out the requirements.txt changes if they mess those up?  I think I will try that now.

fruffy:
https://github.com/p4lang/p4c/actions/runs/7619068359/job/20753493581

It looks like Ubuntu 18 does not support anything over Protobuf 3.19.

jafingerhut:
@fruffy Any preference on what to do next?  e.g. back out the changes to the requirements.txt file?

PR #4342, opened 2024-01-22
cstring: Add string literal suffix, add toLower

vlstill:
Two small additions to the `cstring` utility class:

- allow creating of `cstring`-typed literals thanks to the new `_cs` suffix:
  ```cpp
  using namespace P4::literals;
  foo("abc"_cs + charVal);
  ```

  This is useful for disambiguating overloads or for forcing use of `cstring` operations.

- Add `toLower`. There was already `toUpper`, but `toLower` was missing until now.


fruffy (approved):
Sounds good.  I would add a small unit test for both for sanity checking.

lib/cstring.h
```diff
@@ -356,6 +358,16 @@ inline std::ostream &operator<<(std::ostream &out, cstring s) {
...
 }
 
+/// Let's prevent literal clashes. A user wishing to use the literal can do using namespace
+/// P4::literals, similarly as they can do using namespace std::literals for the standard once.
+namespace P4::literals {
+
+/// A user-provided literal suffix to allow creation of cstring from literals: "foo"_cs.
+/// Note that the C++ standard mandates that all user-defined literal suffixes defined outside of
+/// the standard library must start with underscore.
+inline cstring operator""_cs(const char *str, std::size_t len) { return cstring(str, len); }
```
asl, line 368: By the way, why it essentially makes copy of a literal? Shouldn't this be semantically equivalent to `cstring::literal`?

vlstill: I just didn't notice `cstring::literal`, feel free to change the `_cs`.

asl: Ok, great! So, there is no particular reason that I'm missing here :)

PR #4347, opened 2024-01-22
Switch from C++17 to C++20.

fruffy:
Let's check what breaks.

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -1,1 +1,1 @@
-build --action_env=BAZEL_CXXOPTS="-std=c++17"
+build --action_env=BAZEL_CXXOPTS="-std=c++20"
--- a/.clang-format
+++ b/.clang-format
@@ -20,7 +20,7 @@ IncludeBlocks: Regroup
 # Force pointers to the type for C++.
 DerivePointerAlignment: false
 PointerAlignment: Right
-Standard: c++17
+Standard: c++20
 IncludeCategories:
 # Matches common headers first and sorts them before project includes
 - Regex: '^<.+/.*\.h>'
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -242,8 +242,8 @@ enable_testing ()
 #   set (CPACK_PACKAGE_VERSION_PATCH ${CPACK_PACKAGE_VERSION_PATCH}-${__P4C_VERSION_RC})
 # endif ()
 
-# set (CMAKE_CXX_EXTENSIONS OFF) # prefer using -std=c++17 rather than -std=gnu++17
-set (CMAKE_CXX_STANDARD 17)
+# set (CMAKE_CXX_EXTENSIONS OFF) # prefer using -std=c++20 rather than -std=gnu++20
+set (CMAKE_CXX_STANDARD 20)
 set (CMAKE_CXX_STANDARD_REQUIRED ON)
 
 add_cxx_compiler_option ("-Wall")
--- a/bazel/example/.bazelrc
+++ b/bazel/example/.bazelrc
@@ -1,1 +1,1 @@
-build --action_env=BAZEL_CXXOPTS="-std=c++17"
+build --action_env=BAZEL_CXXOPTS="-std=c++20"


vlstill:
I started looking at this some time ago in https://github.com/p4lang/p4c/pull/4297. I was probably too sneaky though. Still not complete, there is quite lot of things that break and I don't have much time for it now. The main problem is clashing operators.

vlstill:
> > I started looking at this some time ago in #4297. I was probably too sneaky though. Still not complete, there is quite lot of things that break and I don't have much time for it now. The main problem is clashing operators.
> 
> I can close this PR in favor of yours. I just wanted to see what needs to be done and it seems moving up is non-trivial.

OK. Yes, it does not seem as easy as was the move to C++17.

fruffy:
There is not much missing for C++20 support. It looks like there is just one small failure left.

vlstill:
Hmm, the error is in boost, that might be hard to fix unless it is our bad usage of boost of course :-). Seems to be related to a particular version as it fails just in the sanitizers build.

What are your thoughts on switching? Do you think it is time for that? We don't get many goodies until we bump minimal required GCC, but even some things from GCC 9 are nice. It would also unlock C++20 for downstreams that have newer compilers. On the other hand, we would be having a more experimental setup, with GCC that is far from stable on C++20.

fruffy:
> Hmm, the error is in boost, that might be hard to fix unless it is our bad usage of boost of course :-). Seems to be related to a particular version as it fails just in the sanitizers build.

The boost version should be the same for Ubuntu 20.04 but this compiler is Clang. Maybe that is the reason. 

>What are your thoughts on switching? Do you think it is time for that? We don't get many goodies until we bump minimal required GCC, but even some things from GCC 9 are nice. It would also unlock C++20 for downstreams that have newer compilers. On the other hand, we would be having a more experimental setup, with GCC that is far from stable on C++20.

I am happy to switch, there are some folks that have been waiting for us to upgrade to C++20 already. We can try to be conservative with the features we introduce and keep Ubuntu 20.04 as the minimum supported version until it hits EOL.

fruffy:
> Hmm, the error is in boost, that might be hard to fix unless it is our bad usage of boost of course :-). Seems to be related to a particular version as it fails just in the sanitizers build.

Cherrypicking #4663 fixes the issue. There were two more small fixes required which I can fork into a separate PR.

vlstill:
> > Hmm, the error is in boost, that might be hard to fix unless it is our bad usage of boost of course :-). Seems to be related to a particular version as it fails just in the sanitizers build.
> 
> Cherrypicking #4663 fixes the issue. There were two more small fixes required which I can fork into a separate PR.

Personally I'm OK with the operator fixes to be there, but the boost should be separate.

vlstill:
I've just notice that `docs/C++.md` still mentions C++11. No point fixing it to 17 now, but should be fixed when we switch to C++20.

fruffy:
> Personally I'm OK with the operator fixes to be there, but the boost should be separate.

Yes, the changes were for testing purposes, I will move them into separate PRs.

asl:
I tend to agree with @vlstill as it seems to be some bad interaction with standard library. Anyway, if boost uses things that were deprecated in C++17 and removed in C++20, then there is no way to workaround rather than patch / upgrade the sources.

Or we can just get rid `boost::format` so the problem will go away

PS: There are reports about overheads (runtime and compile time) of `std::format`, so we might evaluate the things here...

fruffy:
> Or we can just get rid `boost::format` so the problem will go away

Would be nice, but quite tough because of the difference in formatting... not even sure how you could start here. 

> PS: There are reports about overheads (runtime and compile time) of `std::format`, so we might evaluate the things here...

What about [fmt](https://github.com/fmtlib/fmt)? Will also be much easier to replace with `std::format` in the future.

asl:
> What about [fmt](https://github.com/fmtlib/fmt)? Will also be much easier to replace with `std::format` in the future.

it is nice, yes. But probably we'd need to check / benchmark different solutions

fruffy:
> > What about [fmt](https://github.com/fmtlib/fmt)? Will also be much easier to replace with `std::format` in the future.
> 
> it is nice, yes. But probably we'd need to check / benchmark different solutions

Hmm, I think at least for the usage of boost::format in the compiler most of the calls are used for debugging/warnings/errors, which are not that performance-critical, no? Going with the standard solution is likely fine here.

asl:
> Hmm, I think at least for the usage of boost::format in the compiler most of the calls are used for debugging/warnings/errors, which are not that performance-critical, no? Going with the standard solution is likely fine here.

The complains were both for compile time and for runtime IIRC. This is why I said we need to check :)

vlstill:
Apparently boost 1.74 is the one to fix this: https://www.boost.org/users/history/version_1_74_0.html

> [Format](https://www.boost.org/libs/format/):
> Correct allocator usage (fixes C++20 compilation). (Glen Fernandes)

So that means the one in Ubuntu 22.04 is OK (that is 1.74), but the one in 20.04 (1.71) is bad.

vlstill:
The reason this fails in the sanitizer build has actually nothing to do with clang and everything to do with GCC 10, which is installed in that image, but not in the one used for normal Ubuntu 20.04 build. It is GCC 10 that first removes the overload that `boost::format` requires. So that is why it failed only there.

vlstill:
Are the TC tests supposed to run on 18.04?

Seems like 22.04 is hitting some bug in `hoist_common_match_operations.cpp:333`, we would probably need to rewrite the `std::accumulate` to a plain range for. I may try it tomorrow unless you do in the meantime. That said, we should probably merge #5121 first.

fruffy:
> That said, we should probably merge #5121 before this one.

Kinda scared of that one and C++20, ha.

vlstill:
We will also need to run this on tofino, currently I am getting problems in `bf-p4c/mau/stateful_alu.cpp` so I did not yet get to the problem with Ubuntu 22.

fruffy:
> FYI, locally I am still hitting the boost bug from #4874.

Personally, I still think the best way to solve this is to build with our own boost version.

asl:
> Personally, I still think the best way to solve this is to build with our own boost version.

Or at least require boost version w/o the bug? Do we know what is the minimum Ubuntu version that works here?

fruffy:
> > Personally, I still think the best way to solve this is to build with our own boost version.
> 
> Or at least require boost version w/o the bug? Do we know what is the minimum Ubuntu version that works here?

iirc it was Boost 1.76+ https://stackoverflow.com/a/67702145, which is supplied with Ubuntu 22.04 and above. Since we are deprecating Ubuntu20.04 we _should_ not run into this anymore.

backends/tofino/bf-p4c/mau/gateway.cpp
```diff
@@ -160,7 +160,7 @@ const IR::Expression *CanonGatewayExpr::postorder(IR::Operation::Relation *e) {
     Pattern::Match<IR::Expression> e1, e2;
     Pattern::Match<IR::Constant> k1, k2;
     int width = safe_width_bits(e->left->type);
-    if (((e1 & k1) == k2).match(e) || ((e1 & k1) != k2).match(e)) {
+    if (((e1 & k1) == k2).match(e) || ((e1 & k1) == k2).match(e)) {
```
ChrisDodd, line 163: This changes what is being matched (from an IR::Neq to an IR::Equ) to completely change the behavior of the code.  It seems just wrong

backends/tofino/bf-p4c/mau/instruction_selection.cpp
```diff
@@ -3111,7 +3111,7 @@ const IR::Node *RemoveUnnecessaryActionArgSlice::preorder(IR::Slice *sl) {
 // Simplify "actionArg != 0 ? f0 : f1" to "actionArg ? f0 : f1"
 const IR::Node *SimplifyConditionalActionArg::postorder(IR::Mux *mux) {
     Pattern::Match<IR::MAU::ActionArg> aa;
-    if ((0 != aa).match(mux->e0)) mux->e0 = aa;
+    if ((aa.operator!=(0)).match(mux->e0)) mux->e0 = aa;
```
ChrisDodd, line 3114: This should not be necessary.

backends/tofino/bf-p4c/mau/stateful_alu.cpp
```diff
@@ -1328,8 +1328,8 @@ bool CreateSaluInstruction::preorder(const IR::Operation::Relation *rel, cstring
     if (etype == IF) {
         Pattern::Match<IR::Expression> e1, e2;
         Pattern::Match<IR::Constant> k;
-        if (Device::statefulAluSpec().CmpMask && ((e1 & k) == e2).match(rel) && !k->fitsUint() &&
-            !k->fitsInt()) {
+        if (Device::statefulAluSpec().CmpMask && ((e1 & k).operator==(e2)).match(rel) &&
+            !k->fitsUint() && !k->fitsInt()) {
```
ChrisDodd, line 1332: This should not be necessary

ChrisDodd (approved):

smolkaj (approved):
LGTM 🚀

asl (approved):
Looks trivial enough!

fruffy:
@jafingerhut @vlstill Any concerns or objections? I am planning to pull the trigger and switch the next week.

jafingerhut:
> @jafingerhut @vlstill Any concerns or objections? I am planning to pull the trigger and switch the next week.

No concerns or objections from me.  It looks like you have a good variety of reviews from other folks, which is good.

PR #4348, opened 2024-01-22
Trigger CI workflows based on assigned label.

fruffy:
This PR does two things. 

First, it adds a mechanism to optionally trigger certain CI runs (sanitizers, ubuntu18, validation) by adding the corresponding label to the PR. 

Second, it makes the installation of Python packages for Protobuf subject to the Python version. This way we can install the correct version of a Python package depending on the version being used. This fixes the Ubuntu 18 build issues.


jafingerhut (approved):
So the intent is that ubuntu-18 CI runs will run nightly, _and_ on any PRs with the 'run-buntu18' label?  And similarly for a couple of other nightly CI tests?  If so, looks good to me.

PR #4349, opened 2024-01-22
Remove XDP code generated for TC

vbnogueira:
Remove XDP section that is always generated for the TC backend. This XDP code generated for TC was responsible for setting the skb->protocol field which is not writable in TC. We have a found better way to do that with an extern, hence this XDP section isn't needed anymore.

### DIFF
--- a/backends/tc/ebpfCodeGen.cpp
+++ b/backends/tc/ebpfCodeGen.cpp
@@ -127,9 +127,7 @@ void PNAArchTC::emit(EBPF::CodeBuilder *builder) const {
      * 1. Automatically generated comment
      * 2. Includes
      * 3. Headers, structs
-     * 4. BPF map definitions.
-     * 5. XDP helper program.
-     * 6. TC Pipeline program for post-parser.
+     * 4. TC Pipeline program for post-parser.
      */
 
     // 1. Automatically generated comment.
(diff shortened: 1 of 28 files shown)


fruffy:
Who maintains the TC back end currently? 
@komaljai  @Sosutha ?

Sosutha:
> Who maintains the TC back end currently? @komaljai @Sosutha ?

Yes. Also @usha1830

komaljai:
@vbnogueira There are some tests failing, Could you check?

usha1830 (approved):
LGTM!

vbnogueira:
> @vbnogueira There are some tests failing, Could you check?

It seems like the eBPF backend's test are always assuming this section is there.
I think the solution may be to add another condition so that this section is only removed for the tc backend
We can discuss this in more detail at today's meeting

usha1830:
@vbnogueira the tests are still failing. Looks like the reference output files are not updated. 
I also see missing header inclusion statement in the test output for many files. Could you please check?

vbnogueira:
> @vbnogueira the tests are still failing. Looks like the reference output files are not updated. I also see missing header inclusion statement in the test output for many files. Could you please check?

So, I updated the output files with compiler generated ones. The PR diff better illustrates this.
I'm trying to understand why the output generated by the test is missing the first lines.
Probably some different parameter the test gives to `p4c-pna-p4tc`
Am debugging

vbnogueira:
> > @vbnogueira the tests are still failing. Looks like the reference output files are not updated. I also see missing header inclusion statement in the test output for many files. Could you please check?
> 
> So, I updated the output files with compiler generated ones. The PR diff better illustrates this. I'm trying to understand why the output generated by the test is missing the first lines. Probably some different parameter the test gives to `p4c-pna-p4tc` Am debugging

Figured out the issues, tests are passing

usha1830:
@vbnogueira Could you please rebase this with main

vbnogueira:
> @vbnogueira Could you please rebase this with main

Rebased and all tests are passing

PR #4352, opened 2024-01-23
 Deprecate .txt, support in favour of .txtpb. 

### LINKED ISSUE #4066: Support `.textproto` extension for output files in Protobuf Text Format

fruffy:
Ran into this so might as well fix it. 

Fixes #4066. 

First commit has the actual changes, second commit updates the reference files, third commit updates the DPDK and PSA back ends, too.


fruffy:
> Is it intentional to delete all *.p4 file in the directory testdata/p4_16_bmv_errors, and a few other directories?

Considering the tests pass, yes. It looks like these are stale error files which have never been cleaned up.

jafingerhut:
> > Is it intentional to delete all *.p4 file in the directory testdata/p4_16_bmv_errors, and a few other directories?
> 
> Considering the tests pass, yes. It looks like these are stale error files which have never been cleaned up.

The reason I ask is that some files with suffix `.p4` are source files for unit tests, so deleting those files would eliminate those unit tests, and not cause any test failures.

There are other files with suffix `.p4` that are expected output files to be checked against actual output files during tests, and deleting those (without also deleting the corresponding source `.p4` files) _would_ case test failures.

I am not sure which category the ones in directory `testdata/p4_16_bmv_errors` fall under.

smolkaj:
This is a really nice quality of life improvement, thanks for the change!

fruffy:
> > > Is it intentional to delete all *.p4 file in the directory testdata/p4_16_bmv_errors, and a few other directories?
> > 
> > 
> > Considering the tests pass, yes. It looks like these are stale error files which have never been cleaned up.
> 
> The reason I ask is that some files with suffix `.p4` are source files for unit tests, so deleting those files would eliminate those unit tests, and not cause any test failures.
> 
> There are other files with suffix `.p4` that are expected output files to be checked against actual output files during tests, and deleting those (without also deleting the corresponding source `.p4` files) _would_ case test failures.
> 
> I am not sure which category the ones in directory `testdata/p4_16_bmv_errors` fall under.

Oh I see what you mean. Long day, will fix this.

jafingerhut (approved):
After commit 3 on this PR I see many changes files in `testdata` subdirectories with `_outputs` in their names, and only in those directories, which makes me much more confident that no test cases have been changed or removed.  Looks good to me.  Given the number of places you found that `.txt` needed updating to `.txtpb`, I wouldn't be surprised if there might be another one in there lurking somewhere, but if so, it probably isn't being tested in CI.

PR #4360, opened 2024-01-24
Improve diagnostics for `AssignmentStatement`

kfcripps:
I surrounded several places that can print `AssignmentStatement`s with '', but now other types of nodes get surrounded with '' when they don't necessarily need to be. @fruffy Let me know if you prefer it this way or without quotes in these places.


fruffy (approved):

jafingerhut (approved):
I have not reviewed the C++ code changes, only the changes to error messages in the expected output files.  Those changes all look reasonable to me.

kfcripps:
Rebased

PR #4363, opened 2024-01-25
[P4Testgen] Basic support for `@p4runtime_translation` and `@p4runtime_translation_mappings`.

fruffy-g:
This PR adds support for `@p4runtime_translation` and `@p4runtime_translation_mappings` in the Protobuf IR back end of P4Testgen.  There are two phases.

1. Since the P4Runtime translation annotations are only attached to the basic type definitions, we add a pass which propagates the annotations to any control plane element which uses the type. We can now reference the annotations when generating tests. 

2. In the Protobuf IR back end, we add two checks for `@p4runtime_translation` and `@p4runtime_translation_mappings`. For @p4runtime_translation, we simply override the behavior of the `getFormatOfNode` function. If a `@p4runtime_translation` annotation is present, we return the value mapped in the annotation. Otherwise we default for normal behavior. 
3. `@p4runtime_translation_mappings` are experimental and not well defined, so we simply brute-force inference here. We generate a <cstring, cstring> map which maps the second element in the pair to the first. We also parse the annotation as a list expression.  We add a wrapper to `formatNetworkValue` which overrides the returned value if it is contained in the map. 
4. Tests are missing. These properties are difficult to test with the current infrastructure. We could try to parse some programs which have these annotations.

### DIFF
--- a/backends/p4tools/modules/testgen/targets/bmv2/test_backend/protobuf_ir.h
+++ b/backends/p4tools/modules/testgen/targets/bmv2/test_backend/protobuf_ir.h
@@ -72,12 +72,25 @@ class ProtobufIr : public Bmv2TestFramework {
     /// @returns the inja test case template as a string.
     static std::string getTestCaseTemplate();
 
-    /// Tries to find the @format annotation of a node and, if present, returns the format specified
-    /// in this annotation. Returns "hex" by default.
+    /// Checks whether the node has a `@p4runtime_translation` attached to it. If that is the case,
+    /// returns the name of the translated type contained within the annotation.
+    static std::optional<std::string> checkForP4RuntimeTranslationAnnotation(
+        const IR::IAnnotated *node);
+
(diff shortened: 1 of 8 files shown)


backends/p4tools/modules/testgen/targets/bmv2/test_backend/protobuf_ir.h
```diff
@@ -72,13 +72,26 @@ class ProtobufIr : public Bmv2TestFramework {
     /// @returns the inja test case template as a string.
     static std::string getTestCaseTemplate();
 
-    /// Tries to find the @format annotation of a node and, if present, returns the format specified
-    /// in this annotation. Returns "hex" by default.
+    /// Checks whether the node has a `@p4runtime_translation` attached to it. If that is the case,
+    /// returns the translated type contained within the annotation.
```
smolkaj, line 76: The name of the translated type?

smolkaj (approved):

PR #4364, opened 2024-01-26
Remove deprecated `std::allocator` members

smolkaj:
The change replaces `std::allocator` members that were deprecated in C++17 and removed in C++20, see https://en.cppreference.com/w/cpp/memory/allocator for a detailed list.

### DIFF
--- a/lib/ordered_map.h
+++ b/lib/ordered_map.h
@@ -66,8 +66,8 @@ class ordered_map {
         COMP comp;
         bool operator()(const K *a, const K *b) const { return comp(*a, *b); }
     };
-    using map_alloc =
-        typename ALLOC::template rebind<std::pair<const K *const, list_iterator>>::other;
+    using map_alloc = typename std::allocator_traits<
+        ALLOC>::template rebind_alloc<std::pair<const K *const, list_iterator>>;
     using map_type = std::map<const K *, list_iterator, mapcmp, map_alloc>;
     map_type data_map;
     void init_data_map() {
--- a/lib/ordered_set.h
+++ b/lib/ordered_set.h
@@ -57,8 +57,8 @@ class ordered_set {
         COMP comp;
         bool operator()(const T *a, const T *b) const { return comp(*a, *b); }
     };
-    using map_alloc =
-        typename ALLOC::template rebind<std::pair<const T *const, list_iterator>>::other;
+    using map_alloc = typename std::allocator_traits<
+        ALLOC>::template rebind_alloc<std::pair<const T *const, list_iterator>>;
     using map_type = std::map<const T *, list_iterator, mapcmp, map_alloc>;
     map_type data_map;
     void init_data_map() {


smolkaj:
Clang format is complaining. Stupid question, how can I run it? I'm using Codespaces.

smolkaj:
I tried `make clang-format` like in the CI script, but I suppose I need to generate the Makefile first? What's the quickest way to get there?

vlstill:
> I tried `make clang-format` like in the CI script, but I suppose I need to generate the Makefile first? What's the quickest way to get there?

Just follow the README: https://github.com/p4lang/p4c#installing-p4c-from-source

vlstill:
Btw if you are trying to get C++20 to work, there is much more to do... I've started few weeks ago in https://github.com/p4lang/p4c/pull/4297 and hit multiple problems. Sadly I don't have much time for that right now.

fruffy:
> Clang format is complaining. Stupid question, how can I run it? I'm using Codespaces.

I am not sure how codespaces works, but `make clang-format-fix-errors` only works with CMake currently. You also need to use `pip3 install clang-format==15.0.6`

fruffy (approved):

vlstill:
> Well, it can't hurt to start in small increments.

For sure it does. The point of the comment was making sure we don't both end up independently doing very similar changes (this one is quite simple, some of the other may not be...). I will try to separate what can be used from my original PR so it can be merged.

fruffy:
I just tried Codespaces (really neat stuff) and to get it to work you need to run `pip3 install --upgrade clang-format==15.0.6` and force reinstallation. The initial codespace environment doesn't ship the binary for some reason.

smolkaj:
Yes, Codespaces is great!

Did you find out a shortcut to fixing the formatting issues, or do I really have to install all p4c dependencies just to format some files? I no nothing about cmake, so would be very grateful about a concrete list of commands I can start from a vanilla Codespaces instance. (Will share the list if I figure it out.)

fruffy:
> Yes, Codespaces is great!
> 
> Did you find out a shortcut to fixing the formatting issues, or do I really have to install all p4c dependencies just to format some files? I no nothing about cmake, so would be very grateful about a concrete list of commands I can start from a vanilla Codespaces instance. (Will share the list if I figure it out.)

```
sudo apt update
sudo apt-get install cmake g++ git automake libtool libgc-dev bison flex \
libfl-dev libboost-dev libboost-iostreams-dev \
libboost-graph-dev llvm pkg-config python3 python3-pip \
tcpdump

pip3 install --user -r requirements.txt
mkdir build
cd build
cmake ..
pip3 install --upgrade clang-format==15.0.6
make clang-format-fix-errors
```
should do the trick. We should really provide some simple install scripts. Everything we have now is built for CI.

vlstill:
> Unfortunately, some checks are required. Maybe we can refactor CMake in a way to run the linters without requiring installing some dependencies. But that is tricky.

I would be quite weary of that. If one needs to format .cpp/.h files, it means they have modified them and therefore they should also check it compiles. What *might* be a good idea is to improve the linter CI output to actually indicate what is wrong/how it should be formatted, so that for very simple cases it would actually make sense to just look at the output from CI and format accordingly, without re-running clang-format locally (I personally tend to run the compiler, but not format, but that is probably my mistake :-D).

fruffy:
>I would be quite weary of that. If one needs to format .cpp/.h files, it means they have modified them and therefore they should also check it compiles. 

Well, technically CI takes care of that for you. Sometimes you just want to make a quick change. Even in CI we currently need to install all the dependencies just to run a linter check. That doesn't seem right. 

>What might be a good idea is to improve the linter CI output to actually indicate what is wrong/how it should be formatted

We do have this, clang-format is just a bit hard to read sometimes. This is the current output:
```
Formatting [706/855] /home/runner/work/p4c/p4c/lib/ordered_map.h
/home/runner/work/p4c/p4c/lib/ordered_map.h:69:54: error: code should be clang-formatted [-Wclang-format-violations]
    using map_alloc = typename std::allocator_traits<
                                                     ^
/home/runner/work/p4c/p4c/lib/ordered_map.h:70:39: error: code should be clang-formatted [-Wclang-format-violations]
        ALLOC>::template rebind_alloc<std::pair<const K *const, list_iterator>>;
                                      ^
Formatting [707/855] /home/runner/work/p4c/p4c/lib/ordered_set.h
/home/runner/work/p4c/p4c/lib/ordered_set.h:60:54: error: code should be clang-formatted [-Wclang-format-violations]
    using map_alloc = typename std::allocator_traits<
                                                     ^
/home/runner/work/p4c/p4c/lib/ordered_set.h:61:39: error: code should be clang-formatted [-Wclang-format-violations]
        ALLOC>::template rebind_alloc<std::pair<const T *const, list_iterator>>;
```

Although I am sure we can make this prettier.

smolkaj:
Thanks a lot, Fabian! Looks like https://github.com/p4lang/p4c/pull/4373 is taking care of things -- thanks @vistill -- but this will be very helpful for future reference.

I also didn't realize that clang-format actually showed me how it wants things formatted, as you noted the output is quite subtle.

+1 that making sure that code compiles is the job of CI.

PR #4367, opened 2024-01-26
Fix the option to pipe stdin to the compiler.

fruffy:
This option was never working because `-` was interpreted as a flag parameter. This little change fixes that. Now it is possible to directly pipe a program to the compiler instead of reading a file. 

Open to suggestions on how to improve this option. The parser options in general are old and quite messy.

### DIFF
--- a/lib/options.cpp
+++ b/lib/options.cpp
@@ -67,7 +67,7 @@ std::vector<const char *> *Util::Options::process(int argc, char *const argv[]) 
                 usage();
                 return nullptr;
             }
-        } else if (opt.startsWith("-")) {
+        } else if (opt.startsWith("-") && opt.size() > 1) {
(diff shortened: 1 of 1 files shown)


smolkaj:
What would be an example of a command that didn't work before but works now?

fruffy:
> What would be a concrete example of a command that didn't work before but works now?

```
echo "control dummy();" | p4c/build/p4test -
[--Wwarn=missing] warning: Program does not contain a `main' module
```

smolkaj:
Thanks, I understand better now.
Perhaps worth adding a quick comment in the code about it?

smolkaj (approved):
Sweet!

jafingerhut (approved):
LGTM

PR #4369, opened 2024-01-29
[P4Testgen] Add a silent mode to P4Tools. Add performance counting at the top level.

fruffy:
Add an option to disable printing any information to standard output. This does not disable warnings, errors, or outputs which may be enabled by other options. 

Also add a timer to the top-level and print metrics there for better collection of statics.

### DIFF
--- a/backends/p4tools/common/options.cpp
+++ b/backends/p4tools/common/options.cpp
@@ -181,6 +181,14 @@ AbstractP4cToolOptions::AbstractP4cToolOptions(cstring message) : Options(messag
         },
         "Provides a randomization seed");
 
+    registerOption(
+        "--disable-info-logging", nullptr,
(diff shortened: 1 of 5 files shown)


backends/p4tools/common/options.cpp
```diff
@@ -181,6 +181,14 @@ AbstractP4cToolOptions::AbstractP4cToolOptions(cstring message) : Options(messag
         },
         "Provides a randomization seed");
 
+    registerOption(
+        "--disable-info", nullptr,
```
smolkaj, line 185: Would it make sense to include the keyword "log" or "logging" in the name of the flag somehow?

smolkaj: --disable-info-logging?

smolkaj (approved):

PR #4371, opened 2024-01-29
Add test P4 programs and STF files related to p4c issue 4370

jafingerhut:


jafingerhut:
@fruffy You mentioned earlier "Could you add a PR with this example program and predication enabled in P4Test?"  This PR has the example program added, and updated expected output files, but I am pretty sure predication is disabled in p4test for these expected output files.  Do you have a recommendation on how I can change this PR so that predication is enabled?

fruffy:
You need to add pass somewhere [here](https://github.com/p4lang/p4c/blob/main/backends/p4test/midend.cpp#L83). 

Although the simple switch mid end does use it, so the bug should be caught by the STF tests. 
https://github.com/p4lang/p4c/blob/main/backends/bmv2/simple_switch/midend.cpp#L105

I can try to run validation on it in case you add the pass to the p4test mid end.

jafingerhut:
I did add an STF test that is failing, by catching the misbehavior.  The exact same STF test passes with the version of the program that "inlines" the action body into the ingress control (manually in source code, not via some compiler pass).

I will try making the change you suggest to the p4test source code in this PR, although I would guess that this will cause many other tests to fail besides my new one, because of mismatches between expected contents of -midend.p4 output files vs. the updated actual midend output.

fruffy:
You are right, there will be a series of failures. For reference, this is the PR that disabled it: 
https://github.com/p4lang/p4c/pull/3916

I was curious whether validation will pick it up. 

https://github.com/p4lang/p4c/actions/runs/7697736211/job/20975378374#step:5:1793
seems to be an instance of a BMv2 failure using your provided tests. 

Does the test need to be this complex or is there a simpler example reproducing the issue?

jafingerhut:
I can reduce the STF test to a single packet.

I have tried making the P4 code simpler than what is included, but I believe that if I remove any of the if branches, the output of the predication pass is correct.  The original code that I reduced this from is way longer.

PR #4372, opened 2024-01-29
[P4Testgen] Introduce a configuration structure for the test back ends. Add the option to customize test names.

fruffy:


backends/p4tools/modules/testgen/targets/bmv2/test_backend/protobuf.cpp
```diff
@@ -323,7 +324,10 @@ void Protobuf::emitTestcase(const TestSpec *testSpec, cstring selectedBranches,
     dataJson["coverage"] = coverageStr.str();
 
     LOG5("Protobuf test back end: emitting testcase:" << std::setw(4) << dataJson);
-    auto incrementedbasePath = basePath;
+
+    auto optBasePath = getTestBackendConfiguration().fileBasePath;
+    BUG_CHECK(optBasePath.has_value(), "Base path is not set.");
+    auto incrementedbasePath = optBasePath.value();
     incrementedbasePath.concat("_" + std::to_string(testId));
     incrementedbasePath.replace_extension(".txtpb");
```
vlstill, line 332: It would be great to lift this into a lib function (of the config?), parametrized by the test id and extension.

vlstill (approved):

backends/p4tools/modules/testgen/targets/bmv2/test_backend/protobuf.cpp
fruffy: It's an annoyingly common pattern with `std::optional` and `std::expected`. Makes you wish C++ had a `?` operator. 

I have been thinking about introducing https://github.com/TartanLlama/optional at some point, since it fills the gap `std::optional` left with optional references and unpacking. 
Or simply use some custom macro like this 
```cpp
#define ASSIGN_OR_RETURN_WITH_MESSAGE_IMPL(temporary, targetVariable, inputFunction, returnValue, \
                                           errorFunction)                                         \
    auto(temporary) = inputFunction;                                                              \
    if (!(temporary)) {                                                                           \
        errorFunction;                                                                            \
        return returnValue;                                                                       \
    }                                                                                             \
    targetVariable = *(temporary);  // NOLINT

#define ASSIGN_OR_RETURN_WITH_MESSAGE(targetVariable, inputFunction, returnValue, errorFunction) \
    ASSIGN_OR_RETURN_WITH_MESSAGE_IMPL(STATUS_MACROS_CONCAT_NAME(temporary, __LINE__),           \
                                       targetVariable, inputFunction, returnValue, errorFunction)
```



There is also https://github.com/abseil/abseil-cpp/blob/master/absl/status/statusor.h.

fruffy: I have created an issue to track this #4403.

vlstill: Yes, the general case is interesting too, but here I meant the function to derive test file name from configuration, which is almost exactly the same in most testgen backend.

fruffy: Yes, there are more patterns that are similar in the test back ends. Any back end that writes to files and does not use a preamble (PTF) has a similar pattern. I could move the entire sequence into a helper. 

However, I would like to clean up the error handling first before introducing a new utility function.

PR #3547, opened 2022-10-02
Switch to C++17 standard

davidbolvansky:

PR #4373, opened 2024-01-29
Use std::allocator_traits::rebind_alloc instead of std::allocator::rebind

vlstill:
The latter is deprecated as of C++17 and removed as of C++20.


vlstill:
The is the first part of https://github.com/p4lang/p4c/pull/4297 which is too large to be done in one go.

I am fully aware this is the same change as https://github.com/p4lang/p4c/pull/4364, but since that one does not pass and I had this patch already done, I've raised this separate PR.

fruffy (approved):

smolkaj (approved):
Sweet, thanks!

ChrisDodd:
Do we need an `#if __cplusplus >` ... here as rebind_alloc does not exist prior to C++14?

fruffy:
> Do we need an `#if __cplusplus >= 20170101` ... here as rebind_alloc does not exist in earlier versions?

Both of  P4C's build tools (CMake and Bazel) require C++17 and the corresponding compilers as minimum version. It was introduced with this PR: https://github.com/p4lang/p4c/pull/3547. Over the past year no one has raised any concerns.

PR #4303, opened 2023-12-19
[P4Testgen] Use P4Runtime in the protobuf test back end. 

fruffy:
With #4292 merged we can now introduce P4Runtime in some target back ends. This is the first step to serializing Protobuf directly instead of using Inja as middleman. 

Instead of computing the P4Runtime IDs ourselves, we can now simply look them up in the P4Info file. This uses the newly added P4Info api.

PR #4374, opened 2024-01-29
[P4Testgen] Add a test case generation API to P4Testgen.

fruffy:
This PR adds the functionality to directly call into P4Testgen to generate tests, instead of writing the tests out to a file. This is useful for calling into P4Testgen directly in case you do not have memory constraints or consider the generated tests temporary objects. There is much more to be done to make this pleasant to use but this PR adds basic support. The newly added test `api_test.cpp` shows how this API could be used. 

The changes mostly involve refactoring to the test framework infrastructure. Currently, this functionality is only supported for Protobuf tests in BMv2, but can be extended to others.

This PR depends on #4372, #4369, and #4303.


smolkaj (approved):
Nice, this is a welcome addition!

PR #4375, opened 2024-01-29
Fixes for MacOS/macports

ChrisDodd:
Some fixes I needed to make things build on my MacOS system

### DIFF
--- a/backends/bmv2/run-bmv2-test.py
+++ b/backends/bmv2/run-bmv2-test.py
@@ -48,7 +48,7 @@ def parse_args():
         "-bd",
         "--buildir",
         dest="builddir",
-        help="The path to the compiler build directory, default is \"build\".",
+        help="The path to the compiler build directory, default is current directory.",
     )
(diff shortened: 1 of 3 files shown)


backends/bmv2/run-bmv2-test.py
```diff
@@ -48,7 +48,7 @@ def parse_args():
         "-bd",
         "--buildir",
         dest="builddir",
-        help="The path to the compiler build directory, default is \"build\".",
+        help="The path to the compiler build directory, default is current directory.",
```
fruffy, line 51: I have no problems with these changes, but I do not know whether they may break things for others.

fruffy (approved):

backends/bmv2/run-bmv2-test.py
ChrisDodd: The point is to allow building in a build directory other than `build`, which used to work, but sometime in the past broke.

Issue #3898, opened 2023-02-17
Replace boost constructs with their C++17 STL equivalents


fruffy:
Inspired by https://github.com/ethereum/solidity/issues/7259.
Now that we have C++17 support there is some boost code we can remove. The less boost we have, the better. 
For now, this can apply:

- [ ] changing `boost::regex` to `std::regex`
- [ ] changing `boost::optional` to `std::optional`
- [ ] changing `boost::variant` to `std::variant`
- [ ] changing `boost::filesystem` to `std::filesystem`

apinski-cavium:
> changing boost::filesystem to std::filesystem
I assume we already require GCC 9 ...

fruffy:
> I assume we already require GCC 9 ...

Not quite sure what you mean by that? GCC 9 should support std::filesystem, no?

apinski-cavium:
> > I assume we already require GCC 9 ...
> 
> Not quite sure what you mean by that? GCC 9 should support std::filesystem, no?

Before GCC 9, std::filesystem was not included in libstdc++ but rather its own library.

fruffy:
> Before GCC 9, std::filesystem was not included in libstdc++ but rather its own library.

I see, but here we should be good, right? The only problem I see is clang and MacOS, I do not know about the level of support there.

vlstill:
> changing boost::regex to std::regex

I'm wondering if this is a good idea. The `std::regex` implementation in libstdc++ is quite widely considered to be bad and slow with no chances of fixing it without an ABI break which is opposed by many in the C++ community. I wonder how does `boost::regex` compare to `std::regex` and if it is even worth replacing it. If we could get rid of all of `boost` then yes, but we can't do that.

For the other cases I agree, the `std::` versions are probably even better than boost once. I just hope there are no parts in boost that were not standardised and we use.

fruffy:
It looks like the open-source code does not even use boost::regex. So we may not have to worry about this particular replacement.

asl:
Some additional bits here sans `boost::format` and `boost::multiprecision`:
  - control-plane/p4RuntimeSymbolTable.cpp uses `boost::split` and `boost::adaptors::reverse`. We can readily replace the latter with `iterator_range`. And the first... well, if abseil would be a dependency, we can use `StrSplit` from there.
  - graph backend uses boost::graph. Looks like a hard dependency to me
  - backends/p4tools/common/lib/model.h uses `boost::container::flat_map`. Need to re-examine if this is needed.
[...]

PR #4376, opened 2024-01-30
Memoize & improve name lookup in `ResolveReferences` 

kfcripps:
Improves performance of the `ResolveReferences` pass by using memoization.

For one of our backend apps, this improved compile-time performance by ~3x.

`time ninja check` on this branch:
```
real    1m14.940s
user    42m25.675s
sys     5m39.234s
```

`time ninja check` on `main` branch:
```
real    1m23.792s
user    49m4.392s
sys     5m39.192s
```

I can also run other benchmarks if requested.

### DIFF
--- a/frontends/common/resolveReferences/resolveReferences.cpp
+++ b/frontends/common/resolveReferences/resolveReferences.cpp
@@ -52,9 +52,7 @@ Util::Enumerator<const IR::IDeclaration *> *ResolutionContext::getDeclsByName(
     auto nsIt = declNames.find(ns);
     std::function<const IR::IDeclaration *(
         const std::pair<const cstring, const IR::IDeclaration *> &)>
-        second = [](const std::pair<const cstring, const IR::IDeclaration *> &entry) {
-            return entry.second;
-        };
+        second = [](const auto &entry) { return entry.second; };
 
     if (nsIt != declNames.end()) {
         auto range = nsIt->second.equal_range(name);
--- a/frontends/common/resolveReferences/resolveReferences.cpp
+++ b/frontends/common/resolveReferences/resolveReferences.cpp
@@ -28,32 +28,31 @@ static const std::vector<const IR::IDeclaration *> empty;
 
 ResolutionContext::ResolutionContext() { anyOrder = P4CContext::get().options().isv1(); }
 
-Util::Enumerator<const IR::IDeclaration*>*
-ResolutionContext::getDeclarations(const IR::INamespace *ns) const {
+Util::Enumerator<const IR::IDeclaration *> *ResolutionContext::getDeclarations(
+    const IR::INamespace *ns) const {
     auto nsIt = namespaceDecls.find(ns);
     auto *declVec = nsIt != namespaceDecls.end() ? nsIt->second : memoizeDeclarations(ns);
-    return Util::Enumerator<const IR::IDeclaration*>::createEnumerator(*declVec);
+    return Util::Enumerator<const IR::IDeclaration *>::createEnumerator(*declVec);
 }
 
-std::vector<const IR::IDeclaration*>*
-ResolutionContext::memoizeDeclarations(const IR::INamespace* ns) const {
+std::vector<const IR::IDeclaration *> *ResolutionContext::memoizeDeclarations(
+    const IR::INamespace *ns) const {
     auto decls = ns->getDeclarations();
     if (auto nest = ns->to<IR::INestedNamespace>()) {
         // boost bug -- trying to iterate with an adaptor over an unnamed temp crashes
         auto temp = nest->getNestedNamespaces();
-        for (auto nn : boost::adaptors::reverse(temp))
-            decls = nn->getDeclarations()->concat(decls);
+        for (auto nn : boost::adaptors::reverse(temp)) decls = nn->getDeclarations()->concat(decls);
     }
 
     return (namespaceDecls[ns] = decls->toVector());
 }
 
-Util::Enumerator<const IR::IDeclaration*>*
-ResolutionContext::getDeclsByName(const IR::INamespace *ns, cstring name) const {
+Util::Enumerator<const IR::IDeclaration *> *ResolutionContext::getDeclsByName(
+    const IR::INamespace *ns, cstring name) const {
     auto nsIt = declNames.find(ns);
-    std::function<const IR::IDeclaration*(const std::pair<const cstring,
-                                          const IR::IDeclaration *> &)> second =
-        [](const std::pair<const cstring, const IR::IDeclaration *> &entry){
+    std::function<const IR::IDeclaration *(
+        const std::pair<const cstring, const IR::IDeclaration *> &)>
+        second = [](const std::pair<const cstring, const IR::IDeclaration *> &entry) {
             return entry.second;
         };
 
@@ -62,8 +61,7 @@ ResolutionContext::getDeclsByName(const IR::INamespace *ns, cstring name) const 
         return Util::enumerate(range.first, range.second)->map(second);
     }
 
-    for (auto *d : *getDeclarations(ns))
-        declNames[ns].emplace(d->getName().name, d);
+    for (auto *d : *getDeclarations(ns)) declNames[ns].emplace(d->getName().name, d);
 
     auto range = declNames[ns].equal_range(name);
     return Util::enumerate(range.first, range.second)->map(second);
--- a/frontends/common/resolveReferences/resolveReferences.cpp
+++ b/frontends/common/resolveReferences/resolveReferences.cpp
@@ -28,18 +28,11 @@ static const std::vector<const IR::IDeclaration *> empty;
 
 ResolutionContext::ResolutionContext() { anyOrder = P4CContext::get().options().isv1(); }
 
-Util::Enumerator<const IR::IDeclaration *> *ResolutionContext::getDeclarations(
-    const IR::INamespace *ns) const {
-    auto nsIt = namespaceDecls.find(ns);
-    auto *declVec = nsIt != namespaceDecls.end() ? nsIt->second : memoizeDeclarations(ns);
-    return Util::Enumerator<const IR::IDeclaration *>::createEnumerator(*declVec);
-}
-
 std::vector<const IR::IDeclaration *> *ResolutionContext::memoizeDeclarations(
     const IR::INamespace *ns) const {
     auto decls = ns->getDeclarations();
     if (auto nest = ns->to<IR::INestedNamespace>()) {
-        // boost bug -- trying to iterate with an adaptor over an unnamed temp crashes
+        // boost::adaptors::reverse expects a reference
         auto temp = nest->getNestedNamespaces();
         for (auto nn : boost::adaptors::reverse(temp)) decls = nn->getDeclarations()->concat(decls);
     }
@@ -47,21 +40,30 @@ std::vector<const IR::IDeclaration *> *ResolutionContext::memoizeDeclarations(
     return (namespaceDecls[ns] = decls->toVector());
 }
 
+Util::Enumerator<const IR::IDeclaration *> *ResolutionContext::getDeclarations(
+    const IR::INamespace *ns) const {
+    auto nsIt = namespaceDecls.find(ns);
+    auto *declVec = nsIt != namespaceDecls.end() ? nsIt->second : memoizeDeclarations(ns);
+    return Util::Enumerator<const IR::IDeclaration *>::createEnumerator(*declVec);
+}
+
+std::unordered_multimap<cstring, const IR::IDeclaration *> &ResolutionContext::memoizeDeclsByName(
+    const IR::INamespace *ns) const {
+    auto &namesToDecls = namespaceDeclNames[ns];
+    for (auto *d : *getDeclarations(ns)) namesToDecls.emplace(d->getName().name, d);
+    return namesToDecls;
+}
+
 Util::Enumerator<const IR::IDeclaration *> *ResolutionContext::getDeclsByName(
     const IR::INamespace *ns, cstring name) const {
-    auto nsIt = declNames.find(ns);
     std::function<const IR::IDeclaration *(
         const std::pair<const cstring, const IR::IDeclaration *> &)>
         second = [](const auto &entry) { return entry.second; };
 
-    if (nsIt != declNames.end()) {
-        auto range = nsIt->second.equal_range(name);
-        return Util::enumerate(range.first, range.second)->map(second);
-    }
+    auto nsIt = namespaceDeclNames.find(ns);
+    auto &namesToDecls = nsIt != namespaceDeclNames.end() ? nsIt->second : memoizeDeclsByName(ns);
 
-    for (auto *d : *getDeclarations(ns)) declNames[ns].emplace(d->getName().name, d);
-
-    auto range = declNames[ns].equal_range(name);
+    auto range = namesToDecls.equal_range(name);
     return Util::enumerate(range.first, range.second)->map(second);
 }
 
@@ -185,7 +187,7 @@ const std::vector<const IR::IDeclaration *> *ResolutionContext::lookup(
                   current->node_type_name());
     }
     if (auto nested = current->to<IR::INestedNamespace>()) {
-        // boost bug -- trying to iterate with an adaptor over an unnamed temp crashes
+        // boost::adaptors::reverse expects a reference
         auto temp = nested->getNestedNamespaces();
         for (auto nn : boost::adaptors::reverse(temp)) {
             auto rv = lookup(nn, name, type);
--- a/frontends/common/resolveReferences/resolveReferences.cpp
+++ b/frontends/common/resolveReferences/resolveReferences.cpp
@@ -28,6 +28,47 @@ static const std::vector<const IR::IDeclaration *> empty;
 
 ResolutionContext::ResolutionContext() { anyOrder = P4CContext::get().options().isv1(); }
 
+Util::Enumerator<const IR::IDeclaration*>*
+ResolutionContext::getDeclarations(const IR::INamespace *ns) const {
+    auto nsIt = namespaceDecls.find(ns);
+    auto *declVec = nsIt != namespaceDecls.end() ? nsIt->second : memoizeDeclarations(ns);
+    return Util::Enumerator<const IR::IDeclaration*>::createEnumerator(*declVec);
+}
+
+std::vector<const IR::IDeclaration*>*
+ResolutionContext::memoizeDeclarations(const IR::INamespace* ns) const {
+    auto decls = ns->getDeclarations();
+    if (auto nest = ns->to<IR::INestedNamespace>()) {
+        // boost bug -- trying to iterate with an adaptor over an unnamed temp crashes
+        auto temp = nest->getNestedNamespaces();
+        for (auto nn : boost::adaptors::reverse(temp))
+            decls = nn->getDeclarations()->concat(decls);
+    }
+
+    return (namespaceDecls[ns] = decls->toVector());
+}
+
+Util::Enumerator<const IR::IDeclaration*>*
+ResolutionContext::getDeclsByName(const IR::INamespace *ns, cstring name) const {
+    auto nsIt = declNames.find(ns);
+    std::function<const IR::IDeclaration*(const std::pair<const cstring,
+                                          const IR::IDeclaration *> &)> second =
+        [](const std::pair<const cstring, const IR::IDeclaration *> &entry){
+            return entry.second;
+        };
+
+    if (nsIt != declNames.end()) {
+        auto range = nsIt->second.equal_range(name);
+        return Util::enumerate(range.first, range.second)->map(second);
+    }
+
+    for (auto *d : *getDeclarations(ns))
+        declNames[ns].emplace(d->getName().name, d);
+
+    auto range = declNames[ns].equal_range(name);
+    return Util::enumerate(range.first, range.second)->map(second);
+}
+
 const std::vector<const IR::IDeclaration *> *ResolutionContext::resolve(
     IR::ID name, P4::ResolutionType type) const {
     const Context *ctxt = nullptr;
@@ -44,7 +85,7 @@ const std::vector<const IR::IDeclaration *> *ResolutionContext::lookup(
     LOG2("Trying to resolve in " << current->toString());
 
     if (auto gen = current->to<IR::IGeneralNamespace>()) {
-        Util::Enumerator<const IR::IDeclaration *> *decls = gen->getDeclsByName(name);
+        Util::Enumerator<const IR::IDeclaration *> *decls = getDeclsByName(gen, name);
         switch (type) {
             case P4::ResolutionType::Any:
                 break;
@@ -297,12 +338,7 @@ const IR::IDeclaration *ResolveReferences::resolvePath(const IR::Path *path, boo
 void ResolveReferences::checkShadowing(const IR::INamespace *ns) const {
     if (!checkShadow) return;
     std::map<cstring, const IR::Node *> prev_in_scope;  // check for shadowing within a scope
-    auto decls = ns->getDeclarations();
-    if (auto nest = ns->to<IR::INestedNamespace>()) {
-        // boost bug -- trying to iterate with an adaptor over an unnamed temp crashes
-        auto temp = nest->getNestedNamespaces();
-        for (auto nn : boost::adaptors::reverse(temp)) decls = nn->getDeclarations()->concat(decls);
-    }
+    auto decls = getDeclarations(ns);
     for (auto *decl : *decls) {
         const IR::Node *node = decl->getNode();
         if (node->is<IR::StructField>()) continue;
--- a/frontends/common/resolveReferences/resolveReferences.h
+++ b/frontends/common/resolveReferences/resolveReferences.h
@@ -29,6 +29,12 @@ enum class ResolutionType { Any, Type, TypeVariable };
 
 /// Visitor mixin for looking up names in enclosing scopes from the Visitor::Context
 class ResolutionContext : virtual public Visitor, public DeclarationLookup {
+    mutable std::unordered_map<const IR::INamespace*,
+                               std::vector<const IR::IDeclaration*>*> namespaceDecls;
+    mutable std::unordered_map<const IR::INamespace*,
+                               std::unordered_multimap<cstring,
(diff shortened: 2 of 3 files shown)


asl:
To add more context here. `ResolveReferences` complexity is quadratic in number of declarations. And it's expensive as it relies on `dynamic_cast` for typechecking and downcasting. So, we end with # of `dynamic_cast` calls that is proportional to square of number of declarations (I have some solution for `dynamic_cast` itself, but it's a bit different story – stay tuned).

While I did not want to rewrite it so it won't be quadratic, what we can do is to reduce quadratic amount of `dynamic_cast` to linear one via some memoization.

frontends/common/resolveReferences/resolveReferences.cpp
```diff
@@ -28,6 +28,43 @@ static const std::vector<const IR::IDeclaration *> empty;
...
+    auto nsIt = namespaceDecls.find(ns);
+    auto *declVec = nsIt != namespaceDecls.end() ? nsIt->second : memoizeDeclarations(ns);
+    return Util::Enumerator<const IR::IDeclaration *>::createEnumerator(*declVec);
+}
+
+std::vector<const IR::IDeclaration *> *ResolutionContext::memoizeDeclarations(
+    const IR::INamespace *ns) const {
+    auto decls = ns->getDeclarations();
+    if (auto nest = ns->to<IR::INestedNamespace>()) {
+        // boost bug -- trying to iterate with an adaptor over an unnamed temp crashes
```
vlstill, line 42: It is not a bug... it would just be a dangling reference. I know this comment existed before, but please update it to indicate it is to prevent dangling reference.

frontends/common/resolveReferences/resolveReferences.cpp
```diff
@@ -28,6 +28,43 @@ static const std::vector<const IR::IDeclaration *> empty;
...
+
+    if (nsIt != declNames.end()) {
+        auto range = nsIt->second.equal_range(name);
+        return Util::enumerate(range.first, range.second)->map(second);
+    }
+
+    for (auto *d : *getDeclarations(ns)) declNames[ns].emplace(d->getName().name, d);
+
+    auto range = declNames[ns].equal_range(name);
+    return Util::enumerate(range.first, range.second)->map(second);
```
vlstill, line 65: There is `Values` in `lib/map.h` that should do a similar thing and might be better. You might need to create an enumerator from it since the rest of code uses enumerators.

frontends/common/resolveReferences/resolveReferences.cpp
```diff
@@ -28,6 +28,43 @@ static const std::vector<const IR::IDeclaration *> empty;
...
+
+    return (namespaceDecls[ns] = decls->toVector());
+}
+
+Util::Enumerator<const IR::IDeclaration *> *ResolutionContext::getDeclsByName(
+    const IR::INamespace *ns, cstring name) const {
+    auto nsIt = declNames.find(ns);
+    std::function<const IR::IDeclaration *(
+        const std::pair<const cstring, const IR::IDeclaration *> &)>
+        second = [](const auto &entry) { return entry.second; };
```
vlstill, line 55: I don't see a reason to explicitly write the type of the function you can declare it as `auto second = ...`.

vlstill (commented):
I am not supper happy about the `mutable`, but I guess this is a safe use of it and probably better than making these semantically-const methods non-const.

frontends/common/resolveReferences/resolveReferences.cpp
fruffy: I would also prefer to not use any boost in the first place here.

frontends/common/resolveReferences/resolveReferences.h
```diff
@@ -29,6 +29,12 @@ enum class ResolutionType { Any, Type, TypeVariable };
 
 /// Visitor mixin for looking up names in enclosing scopes from the Visitor::Context
 class ResolutionContext : virtual public Visitor, public DeclarationLookup {
+    mutable std::unordered_map<const IR::INamespace *, std::vector<const IR::IDeclaration *> *>
```
fruffy, line 32: What requires the addition of mutable? This doesn't seem right to me.

frontends/common/resolveReferences/resolveReferences.h
```diff
@@ -70,6 +76,12 @@ class ResolutionContext : virtual public Visitor, public DeclarationLookup {
 
     const IR::IDeclaration *getDeclaration(const IR::Path *path, bool notNull = false) const;
     const IR::IDeclaration *getDeclaration(const IR::This *, bool notNull = false) const;
+
+    Util::Enumerator<const IR::IDeclaration *> *getDeclarations(const IR::INamespace *ns) const;
+    std::vector<const IR::IDeclaration *> *memoizeDeclarations(const IR::INamespace *ns) const;
```
fruffy, line 81: If these functions modify the maps they shouldn't be const, no?

fruffy (commented):
Thanks for this contribution. This looks interesting.

> While I did not want to rewrite it so it won't be quadratic, what we can do is to reduce quadratic amount of `dynamic_cast` to linear one via some memoization.

What does it take to rewrite it, or what did you have in mind? It might be better to develop a clean solution which we can maintain over the long-term over implementing short-term optimizations. They will become harder to maintain over time. 

Maybe we can factor out memoization into a pre-apply step in the reference map, instead of mixing it in, for example.

frontends/common/resolveReferences/resolveReferences.cpp
kfcripps: What would be the preferred replacement for `boost::adaptors::reverse`?

vlstill: We could just add that adaptor ourselves, if we don't have it somewhere already. I actually happen to have it ready if you want to just take this code. Or I can contribute it in a separate PR


```cpp
/// A wrapper over a container that allows iterating it in reverse in a range for:
/// for (auto &x : reversed(someContainer)) { ... }
/// Note the name is deliberatly lower-case to look like a function as it is used like a function.
template<typename Container>
class reversed {
 public:
    static constexpr const bool is_const = std::is_const_v<std::remove_reference_t<Container>>;
    using iterator = std::conditional_t<is_const, typename Container::const_reverse_iterator,
                                                  typename Container::reverse_iterator>;
    using const_iterator = typename Container::const_reverse_iterator;
    using value_type = typename Container::value_type;
    using reference = std::conditional_t<is_const, typename Container::const_reference,
                                                   typename Container::reference>;
    using const_reference = typename Container::const_reference;

    /// Does NOT take owenership of the container, keeps reference only.
    explicit reversed(Container &container) : m_container(container) { }

    /// Prohibit copy and move, this is not intended to ever be stored.
    reversed(reversed &&) = delete;

    iterator begin() { return m_container.rbegin(); }
    iterator end() { return m_container.rend(); }
    const_iterator begin() const { return m_container.rbegin(); }
    const_iterator end() const { return m_container.rend(); }
    const_iterator cbegin() const { return m_container.crbegin(); }
    const_iterator cend() const { return m_container.crend(); }
 private:
    Container &m_container;
};
```

frontends/common/resolveReferences/resolveReferences.cpp
kfcripps: This will not work, because `Enumerator::map(...)` expects a `std::function`, and not a lambda:
```
/local/kfcripps/repos/p4c/frontends/common/resolveReferences/resolveReferences.cpp: In member function 'Util::Enumerator<const IR::IDeclaration*>* P4::ResolutionContext::getDeclsByName(const IR::INamespace*, cstring) const':
/local/kfcripps/repos/p4c/frontends/common/resolveReferences/resolveReferences.cpp:57:70: error: no matching function for call to 'Util::Enumerator<std::pair<const cstring, const IR::IDeclaration*> >::map(P4::ResolutionContext::getDeclsByName(const IR::INamespace*, cstring) const::<lambda(const auto:1&)>&)'
   57 |         return Util::enumerate(range.first, range.second)->map(second);
      |                                                                      ^
In file included from /local/kfcripps/repos/p4c/ir/vector.h:22,
                 from /local/kfcripps/repos/p4c/ir/indexed_vector.h:23,
                 from /local/kfcripps/repos/p4c/build/ninja/release/ir/ir-generated.h:10,
                 from /local/kfcripps/repos/p4c/ir/ir.h:21,
                 from /local/kfcripps/repos/p4c/frontends/common/resolveReferences/resolveReferences.h:20,
                 from /local/kfcripps/repos/p4c/frontends/common/resolveReferences/resolveReferences.cpp:17:
/local/kfcripps/repos/p4c/lib/enumerator.h:110:20: note: candidate: 'template<class S> Util::Enumerator<S>* Util::Enumerator<T>::map(std::function<S(const T&)>) [with S = S; T = std::pair<const cstring, const IR::IDeclaration*>]'
  110 |     Enumerator<S> *map(std::function<S(const T &)> map);
      |                    ^~~
/local/kfcripps/repos/p4c/lib/enumerator.h:110:20: note:   template argument deduction/substitution failed:
/local/kfcripps/repos/p4c/frontends/common/resolveReferences/resolveReferences.cpp:57:70: note:   'P4::ResolutionContext::getDeclsByName(const IR::INamespace*, cstring) const::<lambda(const auto:1&)>' is not derived from 'std::function<S(const std::pair<const cstring, const IR::IDeclaration*>&)>'
   57 |         return Util::enumerate(range.first, range.second)->map(second);
      |                                                                      ^
/local/kfcripps/repos/p4c/frontends/common/resolveReferences/resolveReferences.cpp:63:66: error: no matching function for call to 'Util::Enumerator<std::pair<const cstring, const IR::IDeclaration*> >::map(P4::ResolutionContext::getDeclsByName(const IR::INamespace*, cstring) const::<lambda(const auto:1&)>&)'
   63 |     return Util::enumerate(range.first, range.second)->map(second);
      |                                                                  ^
In file included from /local/kfcripps/repos/p4c/ir/vector.h:22,
                 from /local/kfcripps/repos/p4c/ir/indexed_vector.h:23,
                 from /local/kfcripps/repos/p4c/build/ninja/release/ir/ir-generated.h:10,
                 from /local/kfcripps/repos/p4c/ir/ir.h:21,
                 from /local/kfcripps/repos/p4c/frontends/common/resolveReferences/resolveReferences.h:20,
                 from /local/kfcripps/repos/p4c/frontends/common/resolveReferences/resolveReferences.cpp:17:
/local/kfcripps/repos/p4c/lib/enumerator.h:110:20: note: candidate: 'template<class S> Util::Enumerator<S>* Util::Enumerator<T>::map(std::function<S(const T&)>) [with S = S; T = std::pair<const cstring, const IR::IDeclaration*>]'
  110 |     Enumerator<S> *map(std::function<S(const T &)> map);
      |                    ^~~
/local/kfcripps/repos/p4c/lib/enumerator.h:110:20: note:   template argument deduction/substitution failed:
/local/kfcripps/repos/p4c/frontends/common/resolveReferences/resolveReferences.cpp:63:66: note:   'P4::ResolutionContext::getDeclsByName(const IR::INamespace*, cstring) const::<lambda(const auto:1&)>' is not derived from 'std::function<S(const std::pair<const cstring, const IR::IDeclaration*>&)>'
   63 |     return Util::enumerate(range.first, range.second)->map(second);
      |                                                                  ^
[58/264] Building CXX object frontends/CMakeFiles/frontend.dir/p4/actionsInlining.cpp.o
ninja: build stopped: subcommand failed.
```

frontends/common/resolveReferences/resolveReferences.cpp
kfcripps: I would prefer that you add it and replace this boost usage in a different PR, as this PR is about something else. But I can update the comment in this PR. Btw, `boost::adapter::reverse` is also used below in this file:
```
        // boost bug -- trying to iterate with an adaptor over an unnamed temp crashes
        auto temp = nested->getNestedNamespaces();
        for (auto nn : boost::adaptors::reverse(temp)) {
            auto rv = lookup(nn, name, type);
            if (!rv->empty()) return rv;
        }
```

fruffy: For now you could also just use a reverse_iterator `rbegin` `rend()` which are native.

frontends/common/resolveReferences/resolveReferences.h
kfcripps: As @vlstill mentioned, `mutable` members are often used to avoid making semantically `const` methods non-const. But @asl (the original author) may add additional comment here.

frontends/common/resolveReferences/resolveReferences.cpp
kfcripps: I can do that, but if @vlstill is going to replace this anyway with his `reversed` adapter in a different PR, then I'd rather just leave it as is.

frontends/common/resolveReferences/resolveReferences.h
fruffy: Well, the methods are `const` for a reason. Adding mutable defeats the purpose. I understand using mutable for threaded implementations or counters, but here it appears like flawed design to me.

frontends/common/resolveReferences/resolveReferences.cpp
kfcripps: I'm not sure I understand, are you saying to try:
```
return Util::enumerate(Values(range).begin(), Values(range).end())->map(second);
```
?

frontends/common/resolveReferences/resolveReferences.h
asl: This is a pretty standard approach to use `mutable` for the purposes of memoization. Essentially we are capturing the return value of `const` method and next time it is called with the same arguments we used the captured result.

So, capturing result is a stateful method and therefore we'd either need to remove const (that is not semantically correct) or make the cache `mutable`. The common design patter to use the latter approach.

frontends/common/resolveReferences/resolveReferences.cpp
asl: > I would also prefer to not use any boost in the first place here. We have been slowly trying to get rid of the boost dependencies in the compiler. #3898

This is a pre-exisinting code. Just moved into a helper. I would not add such change in the same commit for the sake of clarity. Boost could be removed elsewhere.

frontends/common/resolveReferences/resolveReferences.cpp
asl: > I don't see a reason to explicitly write the type of the function you can declare it as `auto second = ...`.

You cannot due to way stuff in `enumerators.h` is written. It cannot deduce type.

asl:
> Thanks for this contribution. This looks interesting.
> 
> > While I did not want to rewrite it so it won't be quadratic, what we can do is to reduce quadratic amount of `dynamic_cast` to linear one via some memoization.
> 
> What does it take to rewrite it, or what did you have in mind? It might be better to develop a clean solution which we can maintain over the long-term over implementing short-term optimizations. They will become harder to maintain over time.

I do not recall off-hand. This patch was written more than 1.5 years ago. I remember I decided to go this way because the existing implementation is quite complex with some corner cases here and there.

Note that the impact is quite large on almost any ibput that is longer than 3 lines of P4 code

frontends/common/resolveReferences/resolveReferences.cpp
vlstill: The point is that you would need `second` because that is exactly what the `Values` adaptor does. But since we need the enumerator anyway, this does not help much in the end.

frontends/common/resolveReferences/resolveReferences.cpp
vlstill: > > I don't see a reason to explicitly write the type of the function you can declare it as `auto second = ...`.
> 
> You cannot due to way stuff in `enumerators.h` is written. It cannot deduce type.

You are right, I can see it now.

frontends/common/resolveReferences/resolveReferences.h
vlstill: Yep, this is pretty standard. The function is still semantically const (calling just const functions does not change their results for the same inputs), the only thing that can be observed is timing. If we disregard threads that is :-). But luckily there are no threads except for GC, which is no-issue here (if someone were to compute something in parallel they would not be able to share `ReferenceMap`).

I believe the standard library internally uses a stronger contract for `const` that mandates that `const` functions are safe to be called in parallel, but I am not 100 % sure about that.

asl: > I believe the standard library internally uses a stronger contract for const that mandates that const functions are safe to be called in parallel, but I am not 100 % sure about that.

Yes, we do have this at least in libc++ (have no idea about libstdc++, did not touch it for ages). Introduction of threads would require lots of changes here and there as there lots of data structures that are shared inside p4c. It would require careful review and mutexes here and there. Should this be necessary we can introduce some generic "memoize" (similar to what's in Folly for example) wrapper, but I think it should be done separately.

frontends/common/resolveReferences/resolveReferences.h
asl: No, due to `mutable` used on maps. Semantically these are const functions

frontends/common/resolveReferences/resolveReferences.h
vlstill: Yep, I've mentioned threads just as a side note :-). That is a whole new level of complexity that I hope we can get by without.

fruffy: Personally, I think being forced to use `mutable` is not a good sign. There is typically an alternative, phased solution that involves pre-computation available. While usage of `mutable` is standard there are many things standard in C++, which is a big mess of a language. 

For P4C itself, maintainability and correctness used to take precedence over performance. With that the compiler's  C++philosophy has been to implement a less elegant/performant, but easily understandable solution. In cases, where that was not the case we are often paying the prize. For instance there are several complex passes in the compiler such as the Predication or LoopsUnrolling pass, which are broken in parts but no one has the time or energy to fix them. Ultimately, all of this becomes maintenance burden.

If we go with this option it would be great to add some more documentation to ReferenceMap/ResolveReferences and highlight that the appropriate functions are using caching which depends on `mutable`.

asl: > For P4C itself, maintainability and correctness used to take precedence over performance. With that the compiler's C++philosophy has been to implement a less elegant/performant, but easily understandable solution. In cases, where that was not the case we are often paying the prize. 

These are all important trade-offs that should be made. Wearing my LLVM hat on, I am all for maintainability and clearness of implementation. Note, however, that compiler itself is a tool that is used by many people all the day. And as far as I can see, there are some areas where p4c performance could definitely be better. I am seeing lots of cases where not very complex P4 input causes massive compiler slowdown. And here we are talking not about seconds of compilation time, but minutes and tens of minutes.

For example, in the original downstream testcase this change dropped the compilation time from 129 seconds down to 34. These days we are trying to tackle cases when p4c itself takes 10-15 minutes to run effectively causing very poor developer experience.

There are lots of factors that contribute to this poor performance. Some due to easy to fix corner cases, but others are due to more widespread overall design inefficiencies. And if we are talking about 3-4x compiler speedup as compared to some additional implementation complexity, then for me the tradeoff is clear. Hope the p4c contributors share my view on the things here.

Certainly, one could try to rewrite the entire implementation to get rid of quadratic behavior in this particular case. Maybe we will have time for this one day. For now we are upstreaming the downstream patches that we know to work reasonable well. We are certainly trying to cleanup them as much as possible.

fruffy: I agree, the performance of the compiler definitely needs work. It has been a long-standing problem and everyone benefits from the optimizations, so they are definitely welcome! 

My concern was mostly regarding maintenance on the open-source side, as there are few to none people claiming ownership on the involved code these days. If things go wrong the person responsible may not be reachable. And things such as `mutable` can make debugging very tricky as const-qualified functions may not behave as one would expect. 

Thankfully, we do have fairly expansive testing available. However, issues can still slip by. It is good to know that this code was used extensively downstream and these are patches. I was missing this context, for example.

mihaibudiu: None of the tests in the repository tests the multithreaded compiler, which is used extensively in the production environment for backtracking. Ideally @ChrisDodd would weigh on this.

asl: @mihaibudiu This looks like a problem. Can the tests be added somehow?

mihaibudiu: I personally don't have any multithreaded tests to add. Maybe @ChrisDodd has.

komaljai:
> > @fruffy I think this `test-p4c-ubuntu-...-sanitizers` job might be broken on `main`:
> > ```
> > /home/runner/work/p4c/p4c/backends/tc/ebpfCodeGen.h:273:10: error: 'preorder' overrides a member function but is not marked 'override' [-Werror,-Winconsistent-missing-override]
> >     bool preorder(const IR::ExternBlock *instance);
> >          ^
> > /home/runner/work/p4c/p4c/ir/visitor.h:413:27: note: overridden virtual function is here
> >     IRNODE_ALL_SUBCLASSES(DECLARE_VISIT_FUNCTIONS)
> >                           ^
> > In file included from /home/runner/work/p4c/p4c/build/backends/tc/CMakeFiles/p4c-pna-p4tc.dir/Unity/unity_0_cxx.cxx:4:
> > In file included from /home/runner/work/p4c/p4c/backends/tc/backend.cpp:17:
> > In file included from /home/runner/work/p4c/p4c/backends/tc/backend.h:23:
> > /home/runner/work/p4c/p4c/backends/tc/ebpfCodeGen.h:308:10: error: 'processFunction' overrides a member function but is not marked 'override' [-Werror,-Winconsistent-missing-override]
> >     void processFunction(const P4::ExternFunction *function);
> >          ^
> > /home/runner/work/p4c/p4c/backends/ebpf/ebpfControl.h:44:18: note: overridden virtual function is here
> >     virtual void processFunction(const P4::ExternFunction *function);
> >                  ^
> > In file included from /home/runner/work/p4c/p4c/build/backends/tc/CMakeFiles/p4c-pna-p4tc.dir/Unity/unity_0_cxx.cxx:4:
> > In file included from /home/runner/work/p4c/p4c/backends/tc/backend.cpp:17:
> > In file included from /home/runner/work/p4c/p4c/backends/tc/backend.h:23:
> > /home/runner/work/p4c/p4c/backends/tc/ebpfCodeGen.h:309:10: error: 'processApply' overrides a member function but is not marked 'override' [-Werror,-Winconsistent-missing-override]
> >     void processApply(const P4::ApplyMethod *method);
> >          ^
> > /home/runner/work/p4c/p4c/backends/ebpf/ebpfControl.h:43:18: note: overridden virtual function is here
> >     virtual void processApply(const P4::ApplyMethod *method);
> >                  ^
> > 3 errors generated.
> > make[2]: *** [backends/tc/CMakeFiles/p4c-pna-p4tc.dir/build.make:118: backends/tc/CMakeFiles/p4c-pna-p4tc.dir/Unity/unity_0_cxx.cxx.o] Error 1
> > make[1]: *** [CMakeFiles/Makefile2:2683: backends/tc/CMakeFiles/p4c-pna-p4tc.dir/all] Error 2
> > make: *** [Makefile:166: all] Error 2
> > Error: Process completed with exit code 2.
> > ```
> 
> It's an optional pass which I toggled for this particular contribution. It looks like another PR broke it. @komaljai it looks like either #4378 or #4329 broke something. Could you take a look at https://github.com/p4lang/p4c/actions/runs/7713913116/job/21024954410?pr=4376?

I see the issue has been fixed by @fruffy in PR #4380  Rebasing the branch will resolve the issue @kfcripps

kfcripps:
Just an additional data point: Without these changes, one of our apps takes 9-10x as long to compile:
```
real    16m25.376s
user    16m13.839s
sys     0m14.537s
```
vs
```
real    1m41.843s
user    1m31.168s
sys     0m13.158s
```

frontends/common/resolveReferences/resolveReferences.h
kfcripps: Are we sure that this will cause a problem with the multithreaded compiler, or are we just waiting for @ChrisDodd to comment?

mihaibudiu: We don't know

asl: This might happen if two different threads would share a common `ResolutionContext`. Given that the latter is `Visitor` itself, I would consider it highly unlikely...

ChrisDodd: The idea behind supporting multithreading in the compiler is mostly by having the IR always be read-only, so multiple threads can traverse and visit it simultaneously.  Visitors that don't need to store anything can also be read-only, but any visitor that keeps around traversal state will require a separate instance of the visitor for each thread.  So similarly limiting a ResolutionContext to a single thread is also not too bad.

frontends/common/resolveReferences/resolveReferences.cpp
```diff
@@ -28,6 +28,45 @@ static const std::vector<const IR::IDeclaration *> empty;
...
+    }
+
+    return (namespaceDecls[ns] = decls->toVector());
+}
+
+Util::Enumerator<const IR::IDeclaration *> *ResolutionContext::getDeclarations(
+    const IR::INamespace *ns) const {
+    auto nsIt = namespaceDecls.find(ns);
+    auto *declVec = nsIt != namespaceDecls.end() ? nsIt->second : memoizeDeclarations(ns);
+    return Util::Enumerator<const IR::IDeclaration *>::createEnumerator(*declVec);
```
ChrisDodd, line 47: I suspect you could get a much bigger gain here by getting rid of the java-style `Util::Enumerator` (which does a bunch of dynamic allocation and generates much garbage to be collected by the GC) and using C++ style enumeration instead, but that qould require rewriting all the code that uses this (many passes).

One could think about having a second interface in a ResolutionContext that allows for thatm and then incrementally change other code to use the new interface.  But that is a bigger project,

ChrisDodd (approved):

kfcripps:
Thanks @ChrisDodd for looking at this.

@fruffy @vlstill @mihaibudiu I will merge this soon, unless you have any other comments or want to spend more time looking at this later.

fruffy (approved):

frontends/common/resolveReferences/resolveReferences.cpp
fruffy: Created an issue #4412.

PR #4377, opened 2024-01-30
Implement lightweight RTTI for (not only) IR::Node class hierarchies

asl:
C++ RTTI is heavily used in p4c for type checking and downcasting from IR::Node to a particular descendant. However, generic RTTI implementation as provided by C++ runtime is very heavy and results in many overheads. Typically RTTI support routines (e.g. `dynamic_cast` and corresponding typeid checks) are among top 5 in p4c execution profile.

Unfortunately, p4c cannot re-use static RTTI as implemented by e.g. LLVM as IR::Node class hierarchy contains multiple inheritance with abstract and virtual bases and cross-casting to virtual base is normal in p4c.

This PR implements a bit more sophisticated (as compared by LLVM) RTTI implementation for semi-open class hierarchies. It requires some boilerplate code, but the majority of it is autogenerated by `ir-generator`. Still, manual implementation is also possible as shown by other class hierarchies.

The major features of this implementation are:
  - typeid for a given class could be either specified manually or auto-generated by a compiler
  - typeid for `IR::Node` and its descendants are generated by `ir-generator` and therefore it is possible (in theory) to switch over different node types in addition to other type of polymorphism. Switching over node types in `Vector<T>` and `IndexedVector<T>` is possible as well.
  - `->typeId()` method is fast, it's just a single virtual function call
  - `->is<T>()` method is usually compiled down to several typeid checks by a compiler and is fast as well
  - `->to<T>()` is slightly more elaborated, but still quite fast as we rely on compiler to generate necessary this-adjustment thunks, the implementation fast-paths the case when `->to<T>()` returns nullptr.

Overall the implementation provides a solid 20-25% compile time speedup on various inputs. For example, compare `gtestp4c` runtime (benchmarking using hyperfine, 10 iterations plus a warmup run) on my laptop:

|:---|---:|---:|---:|---:|
| `gtestp4c-main` | 22.671 ± 0.191 | 22.367 | 22.988 | 1.31 ± 0.01 |
| `gtestp4c-rtti` | 17.245 ± 0.095 | 17.132 | 17.417 | 1.00 |

The bulk implementation is in `lib/rtti.h` and relies a bit on constexpr trickery. I checked that everything is resolved down to compile-time constants for gcc >= 9 and recent versions of clang.

Overall, nothing should be changed for the users provided that they use methods from `ICastable`. Direct use of C++ RTTI is supported as well, but should be discouraged due to slowness and overheads.

I also made few cleanups while there:
 - Unchecked `->to<T>()` are converted either to `->checkedTo<T>()` or `->as<T>()` depending on the usage
 - `dynamic_cast` was systematically replaced by methods of `ICastable`
 - Some missed `override` and `const` on methods were added to silence compiler warnings
 - json loader was refreshed a bit while porting to `Json` hierarchy to new RTTI


vlstill:
Speeding up RTTI/`to`/`as`/`is` is definitely a good idea (and it would also be a good idea to go over the code and replace very common `is`+`to` combo by just `to`). That said, I am wondering if this is just too complex, especially for the node hierarchy (BTW. I did not yet decode how this works, this will definitely need a documentation if we want to take this approach).

I am wondering whether it would be possible to make use of existing functions in the `Node` hierarchy, namely the mechanism used for visitors. After all, visitors do dynamic type resolution and it is much cheaper there. I think something like this should be doable (take this as a pseudo code, it is not tested):

```cpp
template<typename T>
const T *to() const {
    struct CastVisitor : Inspector {
        const T *out = nullptr;
        bool preorder(const T *n) override {
            out = n;
            return false;
        }
        bool preorder(const IR::Node *n) overrinde { return false; } // or whatever is root of the hierarchy
    };
    CastVisitor vis;
    this->apply_visitor_preorder(vis);
    return vis.out;
}
```

The idea is to let the generated IR hierarchy/visitor pattern resolved the types for us. With a bit of optimization (a specific light-weight apply for this) it should need just two virtual calls -- this optimization would probably be needed if especially if we wanted to make the non-const `to` fast too. A care must be taken this works for non-generated classes as well, ideally with minimal boilerplate, a fallback to `dynamic_cast` for them might go 99.9% the way we need.


> I also made few cleanups while there:
>
>*    Unchecked ->to<T>() are converted either to ->checkedTo<T>() or ->as<T>() depending on the usage
>*    dynamic_cast was systematically replaced by methods of ICastable
>*    Some missed override and const on methods were added to silence compiler warnings
>*    json loader was refreshed a bit while porting to Json hierarchy to new RTTI

I must say I'm not very happy with such a big change containing not-directly-related changes.

docs/C++.md
```diff
@@ -15,3 +15,17 @@ for multiple dispatch in the IR and Visitor classes.
...
+
+#### Using lightweight RTTI for other class hierarchies
+Other class hierarchies besides `IR::Node` descendants could also benefit from lightweight RTTI functionality. In order to use it one needs to annotate the intended class hierarchy as follows:
+  * Derive from `IR::Castable` (or lower-level `RTTI::Base`)
+  * Use `DECLARE_TYPEINFO(Class, Bases..)` variadic macro to define necessary boilerplate. Here typeid will generated during the compile time. Should necessary, explicit typeid could be specified by `DECLARE_TYPEINFO(Class, TypeId, Bases...)`
```
fruffy, line 25: Generally not a fan of forced macro usage. It makes the compiler much harder to use or extend. Any way we can avoid this using C++17/20? I will try to take a closer look later.

fruffy (commented):
Again, thanks for these contributions, great to see some performance work being done. 

Two questions/requests: 

1) How does this affect compile time?  The compiler nowadays has slow compile times, partially because of macro usage that got out of hand. 

2) It is possible to split out the replacements of `dynamic_cast` with ICastable into a separate PR? That seems like a safe change and is easier to review. Ditto for all the smallish fixes you implemented on the way.

asl:
> I am wondering whether it would be possible to make use of existing functions in the `Node` hierarchy, namely the mechanism used for visitors. After all, visitors do dynamic type resolution and it is much cheaper there. I think something like this should be doable (take this as a pseudo code, it is not tested):

Please don't. The existing Visitors have very large overhead and GC traffic. I would probably say that they have even more overhead than existing C++ RTTI usage.

>  it should need just two virtual calls -- this optimization would probably be needed if especially if we wanted to make the non-const to fast too.

The RTTI implementation is just a single virtual function call (using this for `this` adjustment due to virtual bases use).

> I must say I'm not very happy with such a big change containing not-directly-related changes.

The changes are directly related, sadly. Sorry, I should make it clearer:
  - The code used a mixture of `ICastable` and `dynamic_cast` calls. So, I had to unify it
  - The code contained 3 or 4 duplicates of `ICastable` functionality, again this should be cleaned
  - In few places `dynamic_cast` was used to cast away constness. With `ICastable` interface this was a compiler error, so had to be fixed

docs/C++.md
asl: The current code relies on C++17. Macro is used to intrusively inject the code to the class hierarchy. An alternative would be a CRTP-like mixin. However, IMO this would both negatively affect the compile time. Please see the particular implementation at https://github.com/p4lang/p4c/pull/4377/files#diff-d0caf68249685b4c35c4e2aa7e95735b438987527ecdaf4db8f29a8158c8a405R167

asl:
> You mean runtime here?

Compiler runtime, yes. We are improving the compilation time here which is quite.. long even for small inputs. Here is the example profile of `P4CParserUnroll.switch_20160512` from unit test:

Screenshot 2024-01-30 at 08 58 48

Red arrow indicates C++ RTTI calls that totals to 31% of overall runtime (most of GC traffic is from Visitors boilerplate, ignore it for now).

asl:
> 1. How does this affect compile time?  The compiler nowadays has slow compile times, partially because of macro usage that got out of hand.

I checked the compiler trace profile and I do not see preprocessor being an issue here. Part of the problem is that `ir.h` is just one huge file that is included everywhere, we can probably discuss this in the separate issue. So far I do not see anywhere where `rtti.h` would pop up in the compiler profile. Note that typeid's are statically generated by `ir-generator` for the whole `IR::Node` hierarchy and therefore the possible compile impact of constexpr fallbacks are negligible there.

vlstill:
> > I am wondering whether it would be possible to make use of existing functions in the `Node` hierarchy, namely the mechanism used for visitors. After all, visitors do dynamic type resolution and it is much cheaper there. I think something like this should be doable (take this as a pseudo code, it is not tested):
> 
> Please don't. The existing Visitors have very large overhead and GC traffic. I would probably say that they have even more overhead than existing C++ RTTI usage.

The existing visitors have, but the visitor pattern in general should not. That is what I had in mind with the additional apply overload.

---

> > I must say I'm not very happy with such a big change containing not-directly-related changes.
> 
> The changes are directly related, sadly. Sorry, I should make it clearer:
> 
>   * The code used a mixture of `ICastable` and `dynamic_cast` calls. So, I had to unify it
> 
>   * The code contained 3 or 4 duplicates of `ICastable` functionality, again this should be cleaned
> 
>   * In few places `dynamic_cast` was used to cast away constness. With `ICastable` interface this was a compiler error, so had to be fixed

Hmm, I see. These fixes could still be done before the RTTI but I can definitely see why they ended in one PR.

---

I will try to have a look at your RTTI mechanism in more detail.

asl:
> Hmm, I see. These fixes could still be done before the RTTI but I can definitely see why they ended in one PR.

I can factor out this into a separate PR just in case, if this would make review easier.

fruffy:
> I can factor out this into a separate PR just in case, if this would make review easier.

Yes please! Also because it might take longer to get the other things in.

asl:
> I will try to have a look at your RTTI mechanism in more detail.

Thanks! It's just `lib/rtti.h`.

lib/rtti.h
```diff
@@ -0,0 +1,181 @@
...
+struct TypeNameHolder {
+    template <size_t N>
+    constexpr TypeNameHolder(const char (&str)[N]) : str(&str[0]), length(N - 1) {}  // NOLINT
+
+    const char *str;
+    size_t length;
+};
+
+template <typename T>
+constexpr TypeNameHolder TypeNameProxy() {
```
vlstill, line 40: the p4c mostly has function names start with lowercase

lib/rtti.h
```diff
@@ -0,0 +1,181 @@
...
+#error "Unsupported compiler"
+#endif
+}
+
+static constexpr uint64_t FNV1a(const char *str, size_t n,
+                                uint64_t hash = UINT64_C(0xcbf29ce484222325)) {
+    return n == 0 ? hash : FNV1a(str + 1, n - 1, UINT64_C(0x100000001b3) * (hash ^ str[0]));
+}
+
+static constexpr uint64_t FNV1a(TypeNameHolder str) { return FNV1a(str.str, str.length); }
```
vlstill, line 55: Nitpick: it would be cleaner to write this as a `for` loop and not have the `const char *` version at all.

lib/rtti.h
```diff
@@ -0,0 +1,181 @@
...
+    template <typename T>
+    [[nodiscard]] static const void *dyn_cast(TypeId typeId, const T *ptr) noexcept {
+        if (id() == typeId) return static_cast<const This *>(ptr);
+
+        std::array<const void *, sizeof...(Parents)> ptrs = {
+            Parents::TypeInfo::dyn_cast(typeId, ptr)...};
+
+        auto it =
+            std::find_if(ptrs.begin(), ptrs.end(), [](const void *ptr) { return ptr != nullptr; });
+        return (it != ptrs.end()) ? *it : nullptr;
```
vlstill, line 121: Why would this not work:

```cpp
if (isA(typeId)) return static_cast<const This *>(ptr);
```

I don't see why you would need to defer to the parent's cast function. Also I think the `Parents::TypeInfo` would cast it to the parent pointer, which might need adjusting in presence of multiple inheritance.

Also, this is definitely one of the places that could use comments what exactly is the meaning of the arguments and the return type.

lib/rtti.h
```diff
@@ -0,0 +1,181 @@
...
+    [[nodiscard]] virtual bool isA(TypeId typeId) const noexcept = 0;
+
+    template <typename T>
+    [[nodiscard]] bool is() const noexcept {
+        return isA(TypeInfo<T>::id());
+    }
+
+    template <typename T>
+    [[nodiscard]] T *to() noexcept {
+        return reinterpret_cast<T *>(const_cast<void *>(toImpl(TypeInfo<T>::id())));
```
vlstill, line 138: even simpler, would it be possible to do just `return is<T>() ? static_cast<T *>(this) : nullptr`? The compiler is fully willing to create the static cast to a child, even in presence of multiple inheritance. It will just create a nonsence pointer if the child is not really of that type.

ir/indexed_vector.h
```diff
@@ -212,6 +212,9 @@ class IndexedVector : public Vector<T> {
                       "invalid element %1%", el);
         }
     }
+
+    DECLARE_TYPEINFO_WITH_NESTED_TYPEID(IndexedVector<T>, NK_IndexedVectorT, T::static_typeId(),
```
vlstill, line 216: Is it necessary to use the nested type ID? The name used for the ID derivation does contain the template argument anyway, right?

lib/rtti.h
```diff
@@ -0,0 +1,181 @@
...
+#ifdef __clang__
+    return __PRETTY_FUNCTION__;
+#elif defined(__GNUC__)
+    return __PRETTY_FUNCTION__;
+#else
+#error "Unsupported compiler"
+#endif
+}
+
+static constexpr uint64_t FNV1a(const char *str, size_t n,
```
vlstill, line 50: Would it be possible to just compare the strings instead of calculating a hash? The comparison happens all at compile-time anyway if I'm looking correctly. If we ever get a collision that will be a terrible mess to debug.

vlstill (commented):
I've taken a first look at `rtti.h`, I did not play with it to check if some of my suggestions work as I don't have time that today. I have a few points where I could see some simplifications, but I am not sure if they can be all done.

I believe overall this would be a meaningful change with a good promise. Most of the boilerplate will be generated, so I am not overly concerned about that.

docs/C++.md
```diff
@@ -15,3 +15,17 @@ for multiple dispatch in the IR and Visitor classes.
...
+
+One example of custom hierarchy using this RTTI implementation could be seen in `frontends/p4/typeChecking/typeConstraints.h`.
+
+Important notes:
+  * In order for RTTI to work all class hierarchy should be annotated using `DECLARE_TYPEINFO` macro.
```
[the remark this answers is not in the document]
asl: One of `static_assert`'s in `TypeInfo<T>` will fire. There are several checks there:
  - Class itself is derived from RTTI::Base
  - All its parents are derived from RTTI::Base
  - Typeinfo requested across class hierarchy (so for T->to<S>() T is indeed a parent of S)
  - DECLARE_TYPEINFO is specified for the class (checking the presence of `rttiEnabledMarker`)

lib/rtti.h
asl: The comparison happens in runtime certainly as typeid value is used to determine the dynamic type of object pointer. Let's check `node->is<T>()` implementation for example:

```c++
    template <typename T>
    [[nodiscard]] bool is() const noexcept {
        return isA(TypeInfo<T>::id());
    }
```

Here we are deriving typeid of `T` and checking if `this` is-a given a typeid. Now, `isA` implementation is as follows (note that the function is virtual as we need to know the exact type:
```c++
    [[nodiscard]] bool isA(RTTI::TypeId typeId) const noexcept override {
        return TypeInfo::isA(typeId);
    }
```

Here `TypeInfo = RTTI::TypeInfo<T, AllBasesOfT...>`, so we can check if typeid of input type `T` is compatible with typeid of `this` and all its immediate bases. Then we perform recursion up hierarchy:
```
    [[nodiscard]] static constexpr bool isA(TypeId typeId) noexcept {
        return (id() == typeId) || (... || (Parents::TypeInfo::isA(typeId)));
    }
```

So, the check is performed in runtime as we need to know the exact child via virtual dispatch.

Side note: fold in `isA` is inlined by all the compilers I checked, so it results in a series of fine-grained comparisons (that are typically optimized and combined quite nice).

ir/indexed_vector.h
asl: I want to develop the foundation of possible improved usage in the future. In particular, static polymorphism. With this PR it is possible to do things like this:

```c++
switch (node->typeId()) {
  case NK_Type:
    ...
  case NK_SomethingElse:
    ...
  default:
    ...
}
```
without relying on double-dispatch in Visitors and lots of overheads involved. But what can we do with `Vector<T>` and `IndexedVector<T>`? Can we do something similar here? This is nested typeid is for. Essentially it is split into two pieces: 8 bit discriminator and 56-bits of "real" typeid. `0x00` discriminator is currently reserved for `IR::Node` hierarchy, `0xFF` is for hash-produced values. Correspondingly `0x01` is for `Vector` and `0x02` is for `IndexedVector`.

Now it is possible to:
  - Determine if we are having a particular `Vector<SomeClass>` via checking the full typeid
  - Determine if this is a generic `Vector<T>` via checking discriminator
  - Perform the switch as above via `T`'s typeid from `Vector<T>` or `IndexedVector<T>`.

Hope it makes the usage case of nested typeid's clear

lib/rtti.h
asl: Well, it's not that simple sadly. We need to perform downcast here (from `Node*` to a particular `T*`). `static_cast` would work if we'd cast to the first base (of a dynamic type) as it is essentially a no-op. Things a bit more elaborated when we'd need to cast to non-first base and this is quite typical with p4c as there are lots of casts to e.g. `IDeclaration`.

Let me show you an example. Consider the following class hierarchy:
```c++
class A;
class B;
class C: public A, public B;
```

Now consider the following:
```c++
 A *a = new C;
```

For `a` downcasting to `C*` is free as it's no-op cast. In order to downcast `a` to type `B*` (or, it would be better to name this cross-cast) we need to know the exact class layout here as we need to perform so-called _this-adjustment_ (to put simple, perform something like `a + sizeof(A)`.  And this is essentially what `dynamic_cast` does: it checks the emitted typeinfo to determine offsets of bases and perform the necessary adjustment.

How can we determine the class layout and base offsets without `dynamic_cast`? Well, actually we do not need this. Compiler does this for us when virtual functions are called. Let's extend our example:
```c++
class Base {
protected:
 virtual void foo() = 0;
};
class A : public virtual Base {
protected:
  void foo() override;
};
class B : public virtual Base {
protected:
  void foo() override;
};
class C: public A, public B {
protected:
  void foo() override;
};
```

In order for everything like this to work:
```c++
A *a = new C;
B *b = new C;
C *c = new C;
a->foo(); b->foo(); c->foo();
```
the compiler would need to route the calls via so-called `this-adjustment` thunks, so inside `C::foo()` we will always get the proper `this`. And this is the trick we are using here. We are using a virtual function call to `toImpl`, so compiler would do necessary `this-adjustment` and we'll end with proper dynamic type implementation. Then we'd potentially need to perform cast to one of our bases (doing again adjustment if necessary, note that inside `dyn_cast` `T` and `This` are in direct relationship, so we ensure that `static_cast` will never produce an UB).

p4c case is slightly more elaborated. For example, we are having (this is one of the simpliest cases):
```c++
class Node : public virtual INode;
class StatOrDecl : public Node;
class IDeclaration : public virtual INode;
class Declaration : public StatOrDecl, public virtual IDeclaration;
```

If we're having code like this:
```c++
Node *n = new Declaration();
n->to<IDeclaration>();
```
then the execution flow is as follows:
 - Inside `to<T>` we determine the typeid of `IDeclaration`
 - Call virtually `toImpl`, it lands with proper `this` inside `Declaration::toImpl`; we know all immediate bases of `Declaration` via `RTTI::TypeInfo<>`.
 - From there we do upcast via `TypeInfo::dyn_cast`

Note that `TypeInfo::dyn_cast` is not virtual and all the recursion and `find` there is nicely inlined by a compiler to quite tight implementation. I tried more lazy approach (iterate over bases, return on first match) and it seems to optimize much worse provided notable slowdown.

Hope this helps understanding the flow.

I'd really like to have just plain `static_cast` like we are having in LLVM, but this is not quite possibly here :(

lib/rtti.h
asl: See below

lib/rtti.h
asl: Ah, and string comparison in runtime is not quite an option here. `__platform_strcmp` on the profile in https://github.com/p4lang/p4c/pull/4377#issuecomment-1917495763 actually comes from typeid comparison

ir/indexed_vector.h
asl: And note that the implementation supports two ways to specify typeid:
  - Explicit. And this is used for IR::Node and friends. We rely on `ir-generator` to generate `NodeKind` enum for us.
  - Implicit. Autogenerated from the hash of type name

lib/rtti.h
fruffy: What about using `std::type_index`? I believe that is guaranteed to be unique, no?

asl: > What about using std::type_index? I believe that is guaranteed to be unique, no?

Well, it depends on how you'd define "unique". `std::type_index` is just a thin wrapper over the `type_info` pointer. It has various explicit instantiations defined (e.g. `std::hash<type_index>`), so you can use it in associate containers. Everything else (e.g. `==`, etc.) is delegated to underlying type_info. Here is one implementation to be precise: https://github.com/llvm/llvm-project/blob/main/libcxx/include/typeindex#L64 So it's unique in terms of `operator==` yes, but there are interesting details here.

We are returning essentially to the same set of problems we are trying to solve:
  - We'd need to use `typeid()` that we're trying to avoid
  - type_info comparison (`type_info::operator==` and corresponding `type_index::operator==` that just delegates to it) could be slow on some platforms and in some cases (Apple ARM64 is one of the most notable example among the others), so we'd need to resort to string comparison. Addresses of `type_info`'s are not unique.

You can check the details of Itanium C++ ABI for the description and variants of `type_info` layout, but p4c case is essentially the most complex one due to multiple virtual bases and diamond-shaped inheritance.

Another IMO important point is that this implementation allows for plug-in manual typeid that could be used for various static polymorphism as shown in https://github.com/p4lang/p4c/pull/4377#discussion_r1473533876 I do not think there is sane way of collecting typeid's from the class hierarchy via ordinary C++ RTTI.

PS: I must probably say that having custom RTTI for closed or semi-open class hierarchies is quite common thing:
  - LLVM has it. And things that are using it down the stack (Swift compiler, MLIR-based things, etc.), see https://llvm.org/docs/HowToSetUpLLVMStyleRTTI.html and https://github.com/llvm/llvm-project/blob/main/llvm/include/llvm/Support/Casting.h
  - Game engines have RTTI for game objects. Here is an example RTTI implementation in Lumberyard: https://github.com/aws/lumberyard/blob/master/dev/Code/Framework/AzCore/AzCore/RTTI/RTTI.h

asl:
I rebased the PR, so now it includes only RTTI-related changes as ICastable pre-requisite was merged in.

lib/rtti.h
vlstill: I see. I did not know the `constexpr` implementation according to the C++17 rules is still so buggy.

asl: Yeah. `constexpr`'s are quite hard from the compiler perspective. Also, they put some additional pressure on standard library and they allowed by a standard to fallback to runtime implementation. So yes, we need to ensure that the generated code is what we'd expect. C+20 will simplify lots of things here.

lib/rtti.h
vlstill: Also, please add a comment that we require the T's real type to be in the string so that when someone tries to implement this for MSVC or something they are not too surprised.

lib/rtti.h
vlstill, line 50: Note: There is already `fnv1a` in `lib/hash.h`, but looking at it, its interface is quite a bit different so adding this version alongside it could be more confusing then helpful (here we do need `uint64_t` output, while the one in `hash.h` uses `size_t` and works also in cases where `size_t` is 32 bits).

backends/bmv2/common/controlFlowGraph.h
```diff
@@ -72,10 +72,12 @@ class CFG final : public IHasDbPrint {
         const cstring name;
         EdgeSet successors;
 
-        void dbprint(std::ostream &out) const;
+        void dbprint(std::ostream &out) const override;
         void addPredecessors(const EdgeSet *set);
         void computeSuccessors();
         cstring toString() const { return name; }
+
+        DECLARE_TYPEINFO(Node);
```
vlstill, line 80: Missing `ICastable` base? While I don't see many reasons to cast to `ICastable`, I think the typeid declarations should match the hierarchy (restricted to what is `RTTI::Base`-derived). Otherwise we risk very weird bugs.

vlstill (commented):
I think I now understand how this works and I think this is good. Thank you both for the implementation and the explanations.

While I understand that the idea behind p4c was to make the compiler conceptually relatively simple at some performance cost, I believe we should strive to make the compiler faster if the cost of doing so is reasonable. If we want P4 to be popular and the P4 compilers to be based on p4c it needs to be usable and that requires speed too. As @asl said, custom RTTI is used in other places, LLVM being one of the prime compiler examples. 

I have few more comment & questions.

(I did not go over all the changes through the code base, but I've glanced at parts of them).

lib/rtti.h
```diff
@@ -0,0 +1,181 @@
...
+
+ protected:
+    [[nodiscard]] virtual const void *toImpl(TypeId typeId) const noexcept = 0;
+};
+
+}  // namespace RTTI
+
+#define DECLARE_TYPEINFO(T, ...)                                                  \
+ private:                                                                         \
+    static constexpr RTTI::TypeId static_typeId() { return RTTI::InvalidTypeId; } \
```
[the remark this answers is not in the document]
asl: Ah, it's an interesting question! Consider we are having the following class hierarchy:
```c++
class A {
  DECLARE_TYPEINFO_WITH_TYPEID(A, 42);
}; 

class B : public A {
  DECLARE_TYPEINFO(A, B);
};
```

So, `A` has explicit typeid and it is hash-based for `B`. Now we might have a problem, as `B` will have `static_typeId()` inherited from `A`. And therefore things will go very wrong. So we'd need to hide `static_typeId()` in `B`. And this is what this private declaration is doing.

We do have an examples of such usage in P4C, namely `NameMap<V>` and `NodeMap<K, V>`. Both derives from `IR::Node`, but has hash-generated typeids because their value types are not within `IR::Node` hierarchy, so we cannot do nested typeids.

lib/rtti.h
```diff
@@ -0,0 +1,181 @@
...
+    DECLARE_TYPEINFO_COMMON(T, ##__VA_ARGS__)
+
+#define DECLARE_TYPEINFO_WITH_TYPEID(T, Id, ...)                 \
+ public:                                                         \
+    static constexpr RTTI::TypeId static_typeId() { return Id; } \
+    DECLARE_TYPEINFO_COMMON(T, ##__VA_ARGS__)
+
+#define DECLARE_TYPEINFO_WITH_NESTED_TYPEID(T, OuterId, Id, ...)            \
```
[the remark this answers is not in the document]
asl: Yes, we might make this work with hashes as well. But I'd certainly like to have proper usecase first.

backends/bmv2/common/controlFlowGraph.h
asl: Actually not, exactly the opposite. Here is how we are declaring that `Node` is the base of our class hierarchy. We do not have bare `ICastable` pointers, we are not supposed to cast to `ICastable`. Even if we'd try to do this, we'll end with compile errors (as `ICastable` does not have RTTI).

So, the typeid lookup for `is<T>` will end with `Node` as well as all casts

lib/rtti.h
asl: Sure. I believe for MSVC we'll end with `FUNCSIG` (but I do not have MSVC on hand to check it out). With C++20 we can use something like `std::source_location::function_name`

lib/rtti.h
asl: Yes, this is an explicit 64-bit FNV1a implementation, not 32-bit one.

fruffy: This is something we should keep track off. I think the use of that hash function is limited to the cstring implementation.

lib/rtti.h
```diff
@@ -0,0 +1,181 @@
...
+        return reinterpret_cast<const T *>(toImpl(TypeInfo<T>::id()));
+    }
+
+ protected:
+    [[nodiscard]] virtual const void *toImpl(TypeId typeId) const noexcept = 0;
+};
+
+}  // namespace RTTI
+
+#define DECLARE_TYPEINFO(T, ...)                                                  \
```
fruffy, line 152: Can we add a comment to all these defines explaining when they need to be used?

lib/castable.h
```diff
@@ -20,40 +20,23 @@ limitations under the License.
 #include <typeinfo>
 
 #include "lib/exceptions.h"
+#include "lib/rtti.h"
 
 /// Handy type conversion methods that can be inherited by various base classes.
-class ICastable {
+class ICastable : public virtual RTTI::Base {
```
fruffy, line 26: Please add a comment here that whoever extends this class will also have to declare type info using macros.

lib/rtti.h
```diff
@@ -0,0 +1,181 @@
...
+        std::array<const void *, sizeof...(Parents)> ptrs = {
+            Parents::TypeInfo::dyn_cast(typeId, ptr)...};
+
+        auto it =
+            std::find_if(ptrs.begin(), ptrs.end(), [](const void *ptr) { return ptr != nullptr; });
+        return (it != ptrs.end()) ? *it : nullptr;
+    }
+};
+
+struct Base {
```
fruffy, line 125: Could you add comments (`///`) to all the public functions and interfaces? Similar in description to the `.md` you added and the comments you gave, that was quite helpful in context The docs are often quickly outdated and this seems like an extension that deserve a lot of documentation for clarity.

fruffy (commented):
@grg @hanw @usha1830 This will heavily affect the Tofino compiler, so probably worth taking a look. 

@smolkaj @jonathan-dilorenzo @qobilidop Tagging in case this may impact you.

lib/rtti.h
asl: @fruffy Funny enough, the situation is a bit more complicated. `lib/hash.cpp` implements fnv1a and (old) murmur2. murmur2 is used in `cstring` (for unification, as string hash) and fnv1 is used to hash cstring' contents insider parser unroll. Both cases do not make much sense to me:

1. murmur2 is old, slow and known to have flaws
2. fnv1a is quite slow, has mediocre quality. And actually its usage in parser unroll is questionable: since cstrings are guaranteed unique, then the address of string is enough as 64-bit hash

In my opinion:
 - murmur2 should be replaced with something a bit more modern and performant. I'd stick with xxhash3.
 - fnv1a usage from parser unroll should be removed

The only reason why fnv1a was chosen here is that it's small enough to be evaluated in constexpr context without slowing down compilation time. And for this its quality is good enough here: note that we are hashing few dozen of strings that are known during compile time, so our collision requirements are much lower than generic purpose string hash.

PS: Certainly, hash refactoring is a separate task :)

fruffy: I will create an issue for this. 

`ParsersUnroll` is a troubled compiler pass. We shouldn't shy away from making improvements to it and replacing the hash function, if needed.

backends/bmv2/common/controlFlowGraph.h
vlstill: I see, so basically this separates the various RTTI-enabled hierarchies from each other, right? We don't want to have pointers to `RTTI::Base` floating around. Could you please add a comment in into `rtti.h` about this.

asl: Yes, exactly.

ir/indexed_vector.h
asl: Yes, though we know which classes are abstract and which are concrete via that tree macro. Still, it's something that one could consider.

fruffy:
@asl Thanks for adding the comments, will take a look. I will also ping some other stakeholders to see whether they have any concerns.

asl:
@fruffy @vlstill I believe I handled all the comments. Please take a look.

grg:
@asl Great work on this change. Always good to see meaningful performance improvements.

When generating the NodeKind enum elements, could you include the namespace hierarchy as part of the name please? I tested your branch against the Tofino compiler and received a small number of redefinition errors for `NK_*` identifiers. The Tofino compiler has a backend IR in which a small number of classes are named identically to classes in the open source IR, but with a different namespace. (For example, we have a `BFN::ParserState` class that causes a redefinition of NK_ParserState -- the other definition comes from `IR::ParserState`.)

The code right now only looks at the class name from the `IrClass` objects, and doesn't do anything with the `IrNamespace` objects:
```
for (auto cls : *getClasses()) t << "  NK_" << cls->name << " = " << nkId++ << ",\n";
```

Unfortunately `IrNamespace` doesn't currently have a method to produce a flattened cstring of the hierarchy.

tools/ir-generator/irclass.cpp
```diff
@@ -193,11 +196,38 @@ void IrDefinitions::generate(std::ostream &t, std::ostream &out, std::ostream &i
...
+    // Emit node kinds
+    // TODO: Probably it would make sense to topo-sort the IDs to optimize the
+    // comparison trees generated by a compiler
+    t << "enum class NodeKind : uint64_t {\n"
+      << "  Auto = 0,\n"
+      << "  INode = 1,\n"
+      << "  Node = 2,\n";
+
+    unsigned nkId = 3;
+    for (auto cls : *getClasses()) t << "  " << cls->name << " = " << nkId++ << ",\n";
```
grg, line 218: Please see: https://github.com/p4lang/p4c/pull/4377#issuecomment-1928778499

I'd like to extend the NK_* identifiers to include the namespace hierarchy to avoid redefinition errors when there are two classes with the same name but different namespaces.

asl: How does this look like in .def?

asl: Nevermind, was able to reproduce this.

asl:
@grg I believe https://github.com/p4lang/p4c/pull/4377/commits/5d2d991f61cda217ddf13f867a443e34c6397e50 handles namespaces as well. Will you please give it a try?

lib/rtti.h
```diff
@@ -0,0 +1,297 @@
...
+/// P4C IR::Node class herarchy relies a lot on virtual bases and at the same
+/// time does a lot of dynamic type checks and downcasts /
+/// upcasts. Unfortunately, `static_cast` cannot be used to cast to / via
+/// virtual bases and use of `dynamic_cast` involves a non-negligible overhead.
+/// The typical IR::Node class hierarchy looks like this:
+///   class Node : public virtual INode;
+///   class StatOrDecl : public Node;
+///   class IDeclaration : public virtual INode;
+///   class Declaration : public StatOrDecl, public virtual IDeclaration;
+/// and it is pretty common code pattern that does something like this:
```
vlstill, line 35: maybe
```suggestion
/// and it is pretty common code pattern to do something like this:

File: lib/rtti.h [hunk 17]
```diff
@@ -0,0 +1,297 @@
...
+///   IR:: Node *n = new Declaration();
```
vlstill, line 36: Inconsistent use of `IR::`

```suggestion
///   Node *n = new Declaration();

File: lib/rtti.h [hunk 18]
```diff
@@ -0,0 +1,297 @@
...
+}
+
+static constexpr uint64_t FNV1a(const char *str, size_t n,
+                                uint64_t hash = UINT64_C(0xcbf29ce484222325)) {
+    return n == 0 ? hash : FNV1a(str + 1, n - 1, UINT64_C(0x100000001b3) * (hash ^ str[0]));
+}
+
+static constexpr uint64_t FNV1a(TypeNameHolder str) { return FNV1a(str.str, str.length); }
+
+// TypeIdResolver provides a unified way to specified type name `T`. Typeid
```
vlstill, line 104: "... provides a unified way to getting type id of the type `T`, or something like that might be better

lib/rtti.h
```diff
@@ -0,0 +1,297 @@
...
+//  Also, the explicit specification could fallback to an implicit one, of the
+//  returned typeid is `InvalidTypeId`
+//
+//  The top 8 bits of typeid are reserved as discriminator and could be used to
+//  delineate different typeids. Currently the following discriminators are reserved:
+//   - 0xFF is used to mark all hash-derived typeids
+//   - 0x00 is used to mark all explicitly-specified typeids
+//  In addition, IR::Node hierarchy uses the following discriminators:
+//   - 0x01 for Vector<T> with low 56 bits contain typeid of `T`.
+//   - 0x02 for IndexedVector<T> with lot 56 bits contain typeid of `T`.
```
vlstill, line 118:
```suggestion
//   - 0x02 for IndexedVector<T> with low 56 bits contain typeid of `T`.

File: lib/rtti.h [hunk 20]
```diff
@@ -0,0 +1,297 @@
...
+// };
+// class B : public A {
+//   DECLARE_TYPEINFO(A, B);
+// };
+// We need to ensure static_typeId is not visible inside `B`, otherwise
+// TypeIdResolver will be confused.
+
+/// Same as DECLARE_TYPEINFO but allows for explicit typeid to be specified.
+/// No further checks for typeid collision, etc. are performed. Example:
+/// struct Base {
```
vlstill, line 262: There either has to be a check that Id is not equal to `InvalidTypeId`, or (and I would prefer that just for ease of use), `static_typeId` should never be used directly in a place that expect type id.

vlstill (commented):
Thank you for the fixes. Most of my comments are just text tweaks or minor improvement suggestions. 

My only big concern is with interaction of the `(Indexed)Vector`'s nested type ID with classes that would explicitly define type ID to `RTTI::InvalidTypeId` (e.g. to allow non-generated IR hierarchy class to be put into `Vector`). I think that would break (possibly quite catastrophically) -- see below for details.

tools/ir-generator/irclass.cpp
```diff
@@ -193,11 +196,41 @@ void IrDefinitions::generate(std::ostream &t, std::ostream &out, std::ostream &i
...
+    t << "  IDeclaration = " << nkId++ << ",\n";
+    t << "  VectorBase = " << nkId++ << ",\n";
+    t << "  VectorT = UINT64_C(1) << 56,\n"
+      << "  IndexedVectorT = UINT64_C(2) << 56,\n"
+      << "  FirstAuto = UINT64_C(0xFF) << 56\n"
+      << "};\n"
+      << " inline bool operator==(uint64_t lhs, NodeKind rhs) { return lhs == uint64_t(rhs); }\n"
+      << " inline bool operator==(NodeKind lhs, uint64_t rhs) { return uint64_t(lhs) == rhs; }\n"
+      << " inline bool operator!=(uint64_t lhs, NodeKind rhs) { return lhs != uint64_t(rhs); }\n"
+      << " inline bool operator!=(NodeKind lhs, uint64_t rhs) { return uint64_t(lhs) != rhs; }\n";
```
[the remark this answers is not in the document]
asl: We cannot do it here. As "lib/rtti.h" is not included by this time. I am not sure it's a good idea to include "lib/rtti.h" from inside gen-tree.h

tools/ir-generator/irclass.cpp
```diff
@@ -193,11 +196,41 @@ void IrDefinitions::generate(std::ostream &t, std::ostream &out, std::ostream &i
...
         cls->declare(t);
         exit_namespace(t, cls->containedIn);
     }
+
+    t << std::endl;
+
+    // Emit node kinds
+    // TODO: Probably it would make sense to topo-sort the IDs to optimize the
+    // comparison trees generated by a compiler
+    t << "enum class NodeKind : uint64_t {\n"
```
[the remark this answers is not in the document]
asl: No, see above

lib/rtti.h
asl: I think the best way to proceed is just to allow combining of hash-based Id's here? What do you think?

grg:
> 5d2d991

Yes, this gets past the issue I was running into. Thanks!

ir/indexed_vector.h
```diff
@@ -212,6 +212,9 @@ class IndexedVector : public Vector<T> {
                       "invalid element %1%", el);
         }
     }
+
+    DECLARE_TYPEINFO_WITH_NESTED_TYPEID(IndexedVector<T>, NodeKind::IndexedVectorT,
+                                        T::static_typeId(), Vector<T>);
```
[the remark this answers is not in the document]
asl: I added some plumbing that makes `static_typeId` an implementation detail. Now things could be seamlessly combined as we essentially go down to `TypeIdResolver<InnerT>`.

asl:
@vlstill I believe I addressed all review comments:
 - Now it is possible to combine arbitrary typeid with discriminator
 - Additional tests were added for comparison with `dynamic_cast`, etc.
 - Few more wording and typos fixes

asl:
@fruffy Looks like tests started to fail after https://github.com/p4lang/p4c/pull/4387

fruffy:
> @fruffy Looks like tests started to fail after #4387

Yes, you can ignore it since it is not a required test. There is a fix for it already https://github.com/p4lang/p4c/pull/4397

tools/ir-generator/irclass.cpp
```diff
@@ -222,15 +223,30 @@ void IrDefinitions::generate(std::ostream &t, std::ostream &out, std::ostream &i
...
-    t << "  VectorT = UINT64_C(1) << 56,\n"
-      << "  IndexedVectorT = UINT64_C(2) << 56,\n"
-      << "  FirstAuto = UINT64_C(0xFF) << 56\n"
+    t << "  VectorBase = " << nkId++ << "\n"
+      << "};\n";
+    t << "enum class NodeDiscriminator : RTTI::TypeId {\n"
+      << "  NodeT = UINT64_C(1),\n"
+      << "  VectorT = UINT64_C(1),\n"
+      << "  IndexedVectorT = UINT64_C(2),\n"
+      << "  FirstAuto = UINT64_C(0xFF)\n"
```
vlstill, line 232: Maybe just `Auto` as this is no longer first, all the auto-values have this discriminator.

lib/rtti.h
```diff
@@ -67,6 +67,9 @@ namespace RTTI {
 
 using TypeId = std::uint64_t;
 static constexpr TypeId InvalidTypeId = UINT64_C(0);
+static constexpr unsigned kDiscriminatorBits = 56;
+static constexpr uint64_t kInnerTypeIdMask = (UINT64_C(1) << kDiscriminatorBits) - 1;
+static constexpr uint64_t kHashDiscriminator = UINT64_C(0xFF);
```
vlstill, line 72: What does the `k` prefix mean? I don't think we use this naming convention anywhere in the code base.

vlstill (commented):
Thank you for the changes, this way of defining nested type IDs is nicer indeed. I have just two last nitpicks.

tools/ir-generator/irclass.cpp
asl: Done

asl:
@vlstill @fruffy @grg Any further comments? Thanks!

fruffy:
> @vlstill @fruffy @grg Any further comments? Thanks!

Will try to give a final look tomorrow.

lib/rtti.h
fruffy: This is probably something we should encode in the `.clang-tidy` file. We have some style suggestions there already but they are fairly arbitrary.

vlstill (approved):

lib/rtti.h
```diff
@@ -0,0 +1,313 @@
...
+}
+
+static constexpr uint64_t FNV1a(const char *str, size_t n,
+                                uint64_t hash = UINT64_C(0xcbf29ce484222325)) {
+    return n == 0 ? hash : FNV1a(str + 1, n - 1, UINT64_C(0x100000001b3) * (hash ^ str[0]));
+}
+
+static constexpr uint64_t FNV1a(TypeNameHolder str) { return FNV1a(str.str, str.length); }
+
+// TypeIdResolver provides a unified way to getting typeid of the type `T`. Typeid
```
fruffy, line 107: Nit: Inconsistent use of `//` and `///` here and in other headers. Or I can't tell the system used here. From what I can tell, we use /// nowadays for function/class documentation and `//` for inline comments.

fruffy (approved):
Thanks! Let's simmer this until Monday before merging?
@ChrisDodd any comments?

lib/rtti.h
asl: This was intentional. `TypeIdResolver` is an internal class, so its description should not be a part of documentation. Still, I wanted to keep some kind of comments.

ChrisDodd:
It is disturbing to me that this is more efficient than dyamic_cast -- dynamic_cast *should* be faster than a virtual function call.  I would have thought that C++ compilers would have fixed their poor implementations by now.

I'll try to look this over in more detail,  buy overall it seems acceptable.

asl:
> It is disturbing to me that this is more efficient than dyamic_cast -- dynamic_cast _should_ be faster than a virtual function call. I would have thought that C++ compilers would have fixed their poor implementations by now.

Well... not quite given all the requirements standard imposes and given that p4c uses the most complex case (multiple inheritance with virtual bases). It's in the same way as C+ standard essentially demands for one particular implementation of `std::unordered_map` banning open addressing, etc. See https://github.com/llvm/llvm-project/blob/main/libcxxabi/src/private_typeinfo.cpp#L897 as an example of current implementation

> I'll try to look this over in more detail, buy overall it seems acceptable.

Thanks!

lib/rtti.h
ChrisDodd: The difference between `///` and `//` is how it affects Doxygen -- which I can never keep straight

lib/rtti.h
```diff
@@ -0,0 +1,313 @@
...
+
+    template <typename T>
+    [[nodiscard]] static const void *dyn_cast(TypeId typeId, const T *ptr) noexcept {
+        if (id() == typeId) return static_cast<const This *>(ptr);
+
+        std::array<const void *, sizeof...(Parents)> ptrs = {
+            Parents::TypeInfo::dyn_cast(typeId, ptr)...};
+
+        auto it =
+            std::find_if(ptrs.begin(), ptrs.end(), [](const void *ptr) { return ptr != nullptr; });
```
ChrisDodd, line 198: Won't this effectively call the parent `dyn_cast` for all the parents recursively before looking to see if any are non-null?  As soon as a single parent returns a non-null value, you can return that -- no need to check other parents.

lib/rtti.h
```diff
@@ -0,0 +1,313 @@
...
+        return (it != ptrs.end()) ? *it : nullptr;
```
ChrisDodd, line 199: I'm struggling to find a good way of adding a cache here to avoid the recursive parent calls at least some time, making this more in line of what the cost of a `dynamic_cast` *should* be (one indirection through the vtable + a conditional branch or move + an add), but it would require a bunch of profiling to see if it would be effective.  Something like:

    static std::pair<TypeId, intptr_t> cache[kCacheSize];
    if (cache[typeId % kCacheSize].first == typeId])
        return reinterpret_cast<const This *>(reinterpret_cast<intptr_t>(ptr) + cache[typeId % kCacheSize].second);

lib/rtti.h
```diff
@@ -0,0 +1,313 @@
...
+    /// Same as `to`, but returns const pointer to T.
+    template <typename T>
+    [[nodiscard]] const T *to() const noexcept {
+        return reinterpret_cast<const T *>(toImpl(TypeInfo<T>::id()));
+    }
+
+ protected:
+    [[nodiscard]] virtual const void *toImpl(TypeId typeId) const noexcept = 0;
+};
+
```
ChrisDodd, line 239: Might be useful to add

    template <typename T>
    static inline const T *to(const Base *ptr) noexcept {
        return ptr ? ptr->to<T>() : nullptr;
    }

That way, most `dynamic_cast` calls (eg, those in all the `Visitor::visit` functions) could be replaced with `RTTT::to`

lib/rtti.h
```diff
@@ -0,0 +1,313 @@
...
+///   DECLARE_TYPEINFO(Base2);
+/// };
+///
+/// struct Derived : public Base1, public virtual Base2 {
+///   ...
+///   DECLARE_TYPEINFO(Derived, Base1, Base2);
+/// };
+#define DECLARE_TYPEINFO(T, ...)                                                  \
+ private:                                                                         \
+    static constexpr RTTI::TypeId static_typeId() { return RTTI::InvalidTypeId; } \
```
ChrisDodd, line 261: Rather than doing all the complex hasing stuff, why not just something like

    static constexpr RTTI::TypeId static_typeId() { return reinterpret_cast<uintptr_t>(&static_typeId); }

That way, the linker will provide a distinct value (address) for each class automatically.

ChrisDodd (commented):
Looks reasonable overall.

It still bothers me that the C++ compiler uses more than a single indirection through the vtable plus a conditional move or branch and a single add instruction for `dynamic_cast`

lib/rtti.h
asl: I tried this and it seems to yield much worse performance. What happens here is that compiler is happily inlines everything (through parents) and we end with quite nice and tight implementation. Adding lazy resolution / early returns here prevents this – too much conditions are added.

lib/rtti.h
asl: See above: since we're in `constexpr` context the code is decently inlined into a tight linear code, so there is no recursion left. Adding some additional stuff like caching or early returns seem to yield worse recursive code in the end :)

lib/rtti.h
asl: We already replaced these in `Visitors`, maybe I missed something?

asl:
> It still bothers me that the C++ compiler uses more than a single indirection through the vtable plus a conditional move or branch and a single add instruction for `dynamic_cast`

It is indeed so. The key issue is that we need to compute the offset dynamically. And here is the problem as RTTI only holds information about immediate bases. Therefore we need to traverse RTTI for class hierarchy to compute the offset given source and target classes. Virtual bases makes whole picture more complicated.

lib/rtti.h
asl: This inhibits many optimizations for `is<>` or where typeid is involved. Having typeid as a compile-time constant (instead of link-time) allows for some nice simplifications in the set of comparisons.

asl:
I am going to land this provided that will be no last minute objections :)

fruffy:
> I am going to land this provided that there will be no last minute objections :)

Go for it! I will disable merging for https://github.com/p4lang/p4c/pull/4374 because it introduces another ICastable object. That will break.

asl:
@fruffy Thanks!

PR #4382, opened 2024-01-30
[NFC] Eliminate the majority of `dynamic_cast` usage in the codebase in favor of ICastable interface

asl:
This is a spin-off from #4377


asl:
Tagging @fruffy @vlstill

frontends/p4/typeChecking/typeUnification.cpp
```diff
@@ -450,9 +450,8 @@ bool TypeUnification::unify(const BinaryConstraint *constraint) {
         }
         return constraint->reportError(constraints->getCurrentSubstitution());
     } else if (dest->is<IR::Type_Declaration>() && src->is<IR::Type_Declaration>()) {
-        bool canUnify = typeid(dest) == typeid(src) && dest->to<IR::Type_Declaration>()->name ==
```
asl, line 453: This is a pre-existing p4c bug. `typeid(dest) == typeid(src)` is always true since both `dest` and `src` are of `IR::Type*` types and typeid of these expressions matches. Changing to `typeid(*dest) == typeid(*src)` would yield an error as it will consider generic substituted type and correspondent substitution to be separate types, when they are not. This change just removes always true condition. No test changes / failures.

ir/node.h
```diff
@@ -60,12 +68,19 @@ class INode : public Util::IHasSourceInfo, public IHasDbPrint, public ICastable
     virtual cstring node_type_name() const = 0;
     virtual void validate() const {}
     virtual const Annotation *getAnnotation(cstring) const { return nullptr; }
-    /// A checked version of INode::to. A BUG occurs if the cast fails.
-    ///
-    /// A similar effect can be achieved with `&as<T>()`, but this method
-    /// produces a message that is easier to debug.
+
+    // default checkedTo implementation for nodes: just fallback to generic ICastable method
+    template <typename T>
```
asl, line 73: Note that `static_type_name` is defined only on concrete classes, not on interfaces. So, without this change use of `checkedCastTo<T>` where `T` is interface would produce compile error. As a result, one was able to use `as<T>()` with `T` being interface, but not `checkedCastTo<T>()`.

backends/p4tools/common/core/abstract_execution_state.cpp
```diff
@@ -101,7 +101,7 @@ const IR::Expression *AbstractExecutionState::convertToComplexExpression(
         IR::Vector<IR::Expression> components;
         const auto *elementType = resolveType(ts->elementType);
         for (size_t idx = 0; idx < ts->getSize(); idx++) {
-            auto ref = new IR::ArrayIndex(elementType, parent, new IR::Constant(idx));
+            auto ref = new IR::ArrayIndex(elementType, parent, new IR::Constant(uint64_t(idx)));
```
vlstill, line 104: Out of curiosity, was the original code resolved to a wrong overload, if so, what platform is it?

frontends/p4/reassociation.cpp
```diff
@@ -25,7 +25,7 @@ const IR::Node *Reassociation::reassociate(IR::Operation_Binary *root) {
     if (!leftBin) return root;
     if (leftBin->getStringOp() != root->getStringOp()) return root;
     if (!leftBin->right->is<IR::Constant>()) return root;
-    auto c = dynamic_cast<IR::Operation_Binary *>(root->clone());
+    auto c = root->clone()->to<IR::Operation_Binary>();
```
vlstill, line 28: Might as well do the check to prevent null-deref. And I think doing the check before clone is minorly better.

```suggestion
    auto c = root->checkedTo<IR::Operation_Binary>()->clone();

File: ir/ir-inline.h [hunk 5]
```diff
@@ -88,20 +88,20 @@ void IR::Vector<T>::visit_children(Visitor &v) {
             i++;
             continue;
         }
-        if (auto l = dynamic_cast<const Vector *>(n)) {
+        if (auto l = n->template to<Vector<T>>()) {
```
vlstill, line 91: I suggest you replace `auto n = v.apply_visitor(*i)` above with `const IR::Node *n = v.apply_visitor(*i)`. That will resolve the ambiguity and there will be no reason for `template` here.

ir/node.h
```diff
@@ -50,6 +50,14 @@ class IndexedVector;  // IWYU pragma: keep
 
 // node interface
 class INode : public Util::IHasSourceInfo, public IHasDbPrint, public ICastable {
+ private:
+    template <class, class = void>
+    struct has_static_type_name : std::false_type {};
+
+    template <class T>
+    struct has_static_type_name<T, std::void_t<decltype(T::static_type_name())>> : std::true_type {
+    };
```
vlstill, line 59: This could be possibly useful elsewhere. I would at least make it a standalone class in `node.h`, not a nested one. Please also add a short docstring & possibly the `_v` templated variable equivalent like the std:: traits have.

ir/node.h
vlstill: Hmm, maybe we should have `static_type_name` generated also for interfaces. That is a separate task though.

ir/node.h
```diff
@@ -60,12 +68,19 @@ class INode : public Util::IHasSourceInfo, public IHasDbPrint, public ICastable
...
     virtual const Annotation *getAnnotation(cstring) const { return nullptr; }
-    /// A checked version of INode::to. A BUG occurs if the cast fails.
-    ///
-    /// A similar effect can be achieved with `&as<T>()`, but this method
-    /// produces a message that is easier to debug.
+
+    // default checkedTo implementation for nodes: just fallback to generic ICastable method
+    template <typename T>
+    const T *checkedTo(
+        typename std::enable_if_t<!has_static_type_name<T>::value> * = nullptr) const {
```
vlstill, line 75: Hmm, could you place the enable_if on the return type? I don't like the default argument much as it can confuse someone (how much easier this would be with concepts!).

lib/castable.h
```diff
@@ -38,6 +38,12 @@ class ICastable {
         return dynamic_cast<const T &>(*this);
     }
 
+    /// Tries to convert the class to type T. Returns a nullptr if the cast fails.
```
vlstill, line 41: The comment is wrong, as is the original one -- as mentioned above, this throws `std::bad_cast` on failure.

frontends/p4/typeChecking/typeUnification.cpp
vlstill: I definitely agree this was always-true conjunct. I am just wondering if this is a bug that we are not checking that the "kinds" of types match, but I would say it is probably safe to compare names only. It doubt there can be two entities with the same name and one declaring different kind of thing then the other (e.g. extern vs. parser), as you already have an example where the types does not match correctly, I agree with just dropping the conjunct.

vlstill (commented):
Thank you for this changes and for splitting them off, this is a great cleanup. I have just a few minor points and questions.

backends/p4tools/common/core/abstract_execution_state.cpp
asl: This is the typical Mac problem. `size_t` is `unsigned long` there, not `unsigned long long` as e.g. on Linux.

backends/ebpf/ebpfControl.cpp
```diff
@@ -154,7 +154,7 @@ bool ControlBodyTranslator::preorder(const IR::MethodCallExpression *expression)
 
 void ControlBodyTranslator::compileEmitField(const IR::Expression *expr, cstring field,
                                              unsigned alignment, EBPFType *type) {
-    unsigned widthToEmit = dynamic_cast<IHasWidth *>(type)->widthInBits();
+    unsigned widthToEmit = type->as<IHasWidth>().widthInBits();
```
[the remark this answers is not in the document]
asl: Yes, this is intentional. There are couple of places with unchecked `->to<T>()` calls or `dynamic_cast` calls. So, given that `as / checkedTo` throws an error, I replaced such cases with either `as<T>()` when the result was directly used or `checkedTo` when we needed pointer.

frontends/p4/typeChecking/typeUnification.cpp
asl: If we take the exiting testcase for this code, then checking the typeids yields to very confusing error message. As I noted – the ordinary type and the substituted generic type are different nodes, so the typeid check would fail. However, the diagnostics produces is confusing, IIRC – cannot unify "int" with "int" or something like this.

ir/node.h
asl: Likely there was some rationale for this behavior. But maybe not. Though, I tend to agree that having a static type name everywhere would be be better.

lib/castable.h
asl: Oh, nice catch. Indeed!

asl:
@vlstill Added suggestions from the review, please check

lib/castable.h
```diff
@@ -32,12 +32,18 @@ class ICastable {
         return to<T>() != nullptr;
     }
 
-    /// Tries to convert the class to type T. Returns a nullptr if the cast fails.
+    /// Performs a checked cast. A BUG occurs if the cast fails.
```
fruffy, line 35: `std::bad_cast` instead?

ir/node.h
fruffy: >Likely there was some rationale for this behavior. 

I added the ICastable class at some point because I noticed we repeated this pattern a lot. At the time `node_type_name` and `static_type_name` were simply not required. There is no reason not to add them, but it will likely take some effort across the entire compiler.

asl: @fruffy Sorry, I do not follow. What are you suggesting? Indeed we can get rid of this implementation, however we'll end with less friendly error message. Some tests checks this btw.

lib/castable.h
asl: Let's leave it as-is. As I said, `custom-rtti` branch implements it via `checkedCast`. So we won't change the comment all the time.

ir/node.h
fruffy: Just clarifying on 
>Likely there was some rationale for this behavior.

There was not a concrete reason why things were done this way. No need to get rid of the implementation.

lib/castable.h
fruffy: Mostly asking because it might take a while longer until the other PR is merged. But the comment was wrong in the first place and this is an improvement... 

BUG and std::bad_cast are close enough in behavior so it is fine.

ir/node.h
fruffy: Yes, there were a few nodes that implemented their own `checkedTo` and `to` but did not add the other functions. This class just unified things, the mangled typename was, well, "good enough".

backends/p4tools/common/core/abstract_execution_state.cpp
fruffy, line 104:
```suggestion
            auto ref = new IR::ArrayIndex(elementType, parent, new IR::Constant(static_cast<uint64_t>(idx)));
```

fruffy (approved):
I have no further comments from my side on this PR.

asl:
@fruffy @vlstill Thanks for the reviews. I addressed the last comments. Will you please merge when you will have a chance? I will rebase `custom-rtti` branch afterwards.

asl:
@fruffy Thanks for restarting! Could you please merge it for me? Auto-merge does not work since I do not have write access.

fruffy:
> @fruffy Thanks for restarting! Could you please merge it for me? Auto-merge does not work since I do not have write access.

If you plan on contributing more to p4c going forward we could make Kyle and you contributors with merge rights (https://github.com/orgs/p4lang/teams/p4c-committers).

asl:
> > @fruffy Thanks for restarting! Could you please merge it for me? Auto-merge does not work since I do not have write access.
> 
> If you plan on contributing more to p4c going forward we could make Kyle and you contributors with merge rights (https://github.com/orgs/p4lang/teams/p4c-committers).

I do have other patches in my queue, yes. Just one step at a time :)

kfcripps:
> If you plan on contributing more to p4c going forward we could make Kyle and you contributors with merge rights (https://github.com/orgs/p4lang/teams/p4c-committers).

Yes, that will be great. As Anton mentioned earlier, we have already made some of these changes to our compiler over a year ago, but haven't been able to upstream them yet for a number of reasons. But moving forward, we should be able to contribute more, I think. :)

fruffy:
> > If you plan on contributing more to p4c going forward we could make Kyle and you contributors with merge rights (https://github.com/orgs/p4lang/teams/p4c-committers).
> 
> Yes, that will be great. As Anton mentioned earlier, we have already made some of these changes to our compiler over a year ago, but haven't been able to upstream them yet for a number of reasons. But moving forward, we should be able to contribute more, I think. :)

Could you send me an email (on my website)  with your email addresses and some context? I can then start the discussion.

PR #4383, opened 2024-01-31
Dpdk backend:Explicitly use method toString() while moving isValid() Methodcall used as table key, to metadata.

usha1830:
This is required for https://github.com/p4lang/p4c/pull/4354#issuecomment-1914141950
No changes in test output currently, however once above PR is merged in main, some dpdk tests will be broken. Hence, fixing it before merging @kfcripps  PR #4354 .

### DIFF
--- a/backends/dpdk/dpdkArch.cpp
+++ b/backends/dpdk/dpdkArch.cpp
@@ -1542,6 +1542,39 @@ struct keyInfo *CopyMatchKeysToSingleStruct::getKeyInfo(IR::Key *keys) {
     return oneKey;
 }
 
+cstring CopyMatchKeysToSingleStruct::getTableKeyName(const IR::Expression *e) {
+    cstring keyName;
+    if (auto pe = e->to<IR::PathExpression>()) {
+        keyName = pe->path->name;
+    } else if (auto mem = e->to<IR::Member>()) {
+        /* We expect all nested structures are flattened and only
+         * expected expressions are 'h.<hdrname>.<fieldname> or m.<fieldname> */
+        if (isValidMemberField(mem)) {
+            if (auto mexpr = mem->expr->to<IR::Member>()) {
+                keyName = "h." + mexpr->member + "." + mem->member;
+            } else if (mem->expr->is<IR::PathExpression>()) {
+                keyName = "m." + mem->member;
+            }
+        } else {
+            ::error(ErrorType::ERR_INVALID, "Invalid member expression '%1%' used as table key",
+                    mem);
+        }
+    } else if (auto m = e->to<IR::MethodCallExpression>()) {
+        BUG_CHECK(isValidCall(m),
+                  "Method calls except isValid must be simplified before"
+                  " reaching here, found %1%",
+                  m->method);
+        /* Moving h.<hdr>.isValid() used as table key to Metadata */
+        auto mem = m->method->to<IR::Member>();
+        CHECK_NULL(mem);
+        keyName = "h." + mem->expr->to<IR::Member>()->member + "." + IR::Type_Header::isValid;
+    } else {
+        ::error(ErrorType::ERR_UNSUPPORTED_ON_TARGET, "Unsupported key expression %1%",
+                e->toString());
+    }
+    return keyName;
+}
+
 const IR::Node *CopyMatchKeysToSingleStruct::preorder(IR::Key *keys) {
     // If any key field is from different structure, put all keys in metadata
     LOG3("Visiting " << keys);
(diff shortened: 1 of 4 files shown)


kfcripps:
Thank you @usha1830. I temporarily added your fix to #4354, just to check whether this fixes the issue.

usha1830:
> Thank you @usha1830. I temporarily added your fix to #4354, just to check whether this fixes the issue.

@kfcripps I can close this PR if it is included in your PR and if it fixes the issue.

fruffy:
> > Thank you @usha1830. I temporarily added your fix to #4354, just to check whether this fixes the issue.
> 
> @kfcripps I can close this PR if it is included in your PR and if it fixes the issue.

No, let's merge this PR so we can keep track of separate contributions. Thanks!

fruffy (approved):

backends/dpdk/dpdkArch.cpp
```diff
@@ -1666,6 +1666,16 @@ const IR::Node *CopyMatchKeysToSingleStruct::postorder(IR::KeyElement *element)
     }
 
     auto keyName = element->expression->toString();
+    if (auto m = element->expression->to<IR::MethodCallExpression>()) {
+        /* Moving <hdr>.isValid() used as table key to Metadata */
```
fruffy, line 1670: Why not just extract the path expression or member from the method call and then convert it into a string? For me that is the safest option because you are directly accessing the path.

fruffy: Also check that the expression is a method call before calling `toString`.

backends/dpdk/dpdkArch.cpp
```diff
@@ -1666,6 +1666,16 @@ const IR::Node *CopyMatchKeysToSingleStruct::postorder(IR::KeyElement *element)
...
+        auto mi = P4::MethodInstance::resolve(m, refMap, typeMap);
+        if (auto b = mi->to<P4::BuiltInMethod>()) {
+            if (b->name == "isValid") {
+                keyName = m->method->toString();
+            }
+        }
+    }
+
     bool isHeader = false;
```
kfcripps, line 1679: Why only do this for `isValid()` calls, and not all `MethodCallExpression`s?

kfcripps:
It does seem to fix the dpdk failures in my PR though. Thanks!

backends/dpdk/dpdkArch.cpp
usha1830: The table keys at this point could be PathExpression , Member or MethodCallExpression (only isValid). 
The first toString() call on element->expression handles all cases and later isValid is handled. I added a check for confirming that the method call is indeed an isValid call and based on that the keyName is obtained now.

fruffy: What I mean to say is that you can implement this without ever needing to call `toString`. Instead, you can check that each type is either a member, a path expression, or a valid method call. Then resolve the expression appropriately into its subparts. That is how it is done in other parts of the compiler. This way you safely enforce typing, too. 

Largely because `toString` is not meant for the purpose it is used here. The semantics of the function can change, like it happened before.

backends/dpdk/dpdkArch.cpp
```diff
@@ -1665,22 +1662,56 @@ const IR::Node *CopyMatchKeysToSingleStruct::postorder(IR::KeyElement *element)
         insertions = it->second;
     }
 
-    auto keyName = element->expression->toString();
+    cstring keyName;
+    if (auto pe = element->expression->to<IR::PathExpression>()) {
+        keyName = pe->path->name;
+    } else if (auto mem = element->expression->to<IR::Member>()) {
+        /* We expect all nested structures are flattened and only
+         * expected expressions are 'h.<hdrname>.<fieldname> or m.<fieldname> */
```
fruffy, line 1670: Must every header really start with `h`? This seems like an odd restriction. If this restriction is not there you can likely resolve everything recursively.

backends/dpdk/dpdkArch.cpp
```diff
@@ -1665,22 +1662,56 @@ const IR::Node *CopyMatchKeysToSingleStruct::postorder(IR::KeyElement *element)
         insertions = it->second;
     }
 
-    auto keyName = element->expression->toString();
+    cstring keyName;
```
fruffy, line 1665: I would factor this out into a helper.

backends/dpdk/dpdkArch.cpp
```diff
@@ -1665,22 +1662,56 @@ const IR::Node *CopyMatchKeysToSingleStruct::postorder(IR::KeyElement *element)
...
+    if (keyName.isNullOrEmpty()) return element;
     bool isHeader = false;
     /* All header fields are prefixed with "h." and metadata fields are prefixed with "m."
      * Prefix the match field with control and table name */
     if (keyName.startsWith("h.")) {
         isHeader = true;
         keyName = keyName.replace('.', '_');
-        keyName =
-            keyName.replace("h_", control->name.toString() + "_" + table->name.toString() + "_");
+        keyName = keyName.replace(
```
fruffy, line 1706: `originalName` can be null in specific scenarios. You can either check for that or use `toString()` here, which falls back to `name`. It depends on your expectations.

fruffy (commented):
Nice, thanks! Just a couple nitpicks.

backends/dpdk/dpdkArch.cpp
usha1830: The earlier code was using toString() which returns originalName if it is not null, name otherwise. In my attempt to remove toString(), I changed this as well :). In most cases, originalName for control and tables is not Null, I will add a check anyways. 
Thanks!

backends/dpdk/dpdkArch.cpp
usha1830: Yes. Dpdk backend converts all headers into this format in the process of moving closer to the target spec file format.

backends/dpdk/dpdkArch.cpp
fruffy: Yes, just wanted to point out that the original usage was fine. For IR nodes `toString` can be unreliable but for `ID` `toString` should be stable. Although I have run into bugs because I expected `originalName` to be used but got `name`. Very subtle and unpleasant.

backends/dpdk/dpdkArch.cpp
fruffy: Ah, that enforcement looks good to me then! This could also be a helper which you could use in other parts of the code for validation, if necessary.

usha1830: Added a helper

backends/dpdk/dpdkArch.cpp
usha1830: Added

backends/dpdk/dpdkArch.cpp
usha1830: Reverted to original implementation.

fruffy (approved):
Thanks! You can ignore the failing tools test.

usha1830:
@fruffy testgen gtest failed for some reason. Could you please check and suggest what could be the issue?

usha1830:
> Thanks! You can ignore the failing tools test.

Thanks! I rebased again,

PR #4391, opened 2024-02-05
Fixes from static analysis

vlstill:
Fixes for issues found by coverity.

### DIFF
--- a/frontends/p4/alias.h
+++ b/frontends/p4/alias.h
@@ -123,6 +123,8 @@ class ReadsWrites : public Inspector {
     void postorder(const IR::Operation_Binary *expression) override {
         auto left = ::get(rw, expression->left);
         auto right = ::get(rw, expression->right);
+        CHECK_NULL(left);
+        CHECK_NULL(right);
         rw.emplace(expression, left->join(right));
     }
 
@@ -135,11 +137,13 @@ class ReadsWrites : public Inspector {
 
     void postorder(const IR::Operation_Unary *expression) override {
         auto e = ::get(rw, expression->expr);
+        CHECK_NULL(e);
         rw.emplace(expression, e);
     }
 
     void postorder(const IR::Member *expression) override {
         auto e = ::get(rw, expression->expr);
+        CHECK_NULL(e);
         auto result = e->append(expression->member);
         rw.emplace(expression, result);
     }
@@ -182,11 +186,15 @@ class ReadsWrites : public Inspector {
         auto e0 = ::get(rw, expression->e0);
         auto e1 = ::get(rw, expression->e1);
         auto e2 = ::get(rw, expression->e2);
+        CHECK_NULL(e0);
+        CHECK_NULL(e1);
+        CHECK_NULL(e2);
         rw.emplace(expression, e0->join(e1)->join(e2));
     }
 
     void postorder(const IR::Slice *expression) override {
         auto e = ::get(rw, expression->e0);
+        CHECK_NULL(e);
         rw.emplace(expression, e);
     }
 
@@ -194,6 +202,7 @@ class ReadsWrites : public Inspector {
         auto e = ::get(rw, expression->method);
         for (auto a : *expression->arguments) {
             auto s = ::get(rw, a->expression);
+            CHECK_NULL(s);
             e = e->join(s);
         }
         rw.emplace(expression, e);
@@ -203,6 +212,7 @@ class ReadsWrites : public Inspector {
         const SetOfLocations *result = new SetOfLocations();
         for (auto e : *expression->arguments) {
             auto s = ::get(rw, e->expression);
+            CHECK_NULL(s);
             result = result->join(s);
         }
         rw.emplace(expression, result);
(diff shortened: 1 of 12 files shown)


backends/p4tools/modules/testgen/targets/bmv2/p4_asserts_parser.cpp
```diff
@@ -530,7 +530,7 @@ const IR::Node *AssertsParser::postorder(IR::P4Table *node) {
         auto *cloneProperties = node->properties->clone();
         IR::IndexedVector<IR::Property> properties;
         for (const auto *property : cloneProperties->properties) {
-            if (property->name.name != "key" || property->name.name != "entries") {
+            if (property->name.name != "key" && property->name.name != "entries") {
```
vlstill, line 533: @fruffy, note that the original condition was a tautology. I've changed it to match the comment above.

midend/removeComplexExpressions.cpp
```diff
@@ -104,7 +104,7 @@ const IR::Vector<IR::Expression> *RemoveComplexExpressions::simplifyExpressions(
     // of the list.  Otherwise we simplify the argument itself.
     // This is mostly for functions that take FieldLists - these
     // should still take a list as argument.
-    bool changes = true;
+    bool changes = false;
```
vlstill, line 107: Note the for cycle below.

vlstill:
It might make more sense to go over the PR commit-by-commit as each commit is a separate problem fix.

backends/p4tools/modules/testgen/targets/bmv2/p4_asserts_parser.cpp
fruffy: Thanks for catching this. Hope I can get rid of this pass soon.

frontends/p4/alias.h
```diff
@@ -123,6 +123,8 @@ class ReadsWrites : public Inspector {
     void postorder(const IR::Operation_Binary *expression) override {
         auto left = ::get(rw, expression->left);
         auto right = ::get(rw, expression->right);
+        CHECK_NULL(left);
```
fruffy, line 126: We use this a lot. Maybe we should implement some form of checked `::get`?

lib/options.cpp
```diff
@@ -15,7 +15,8 @@ limitations under the License.
 */
 
 #include "options.h"
-#include <lib/null.h>
+
+#include "null.h"
```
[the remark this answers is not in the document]
vlstill: I was mainly surprised this failed in `bazel`, but not everywhere else. This seems like some sort of inconsistency in the build system, but since I don't know bazel I didn't want to investigate it.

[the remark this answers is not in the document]
fruffy: Ah, the quotes in my comment got swallowed.

The lack of support for angled brackets is by design: https://bazel.build/docs/bazel-and-cpp#include-paths

frontends/p4/alias.h
vlstill: Maybe... or just use `.at()` it does produce a standard exception instead of p4c one, but that is not a big problem in my opinion. Maybe we should have some sort of guidelines. I'd guess many new programmers don't even know there is `::get` until they find some in the code.

vlstill:
@fruffy, I've fixed the include. The `::get` seems to me like something that should be done separately (or I could use `at` in this file now, instead of the added `CHECK_NULL`, but I don't want to go over all the `::get` uses now as that seems like quite a lot of work).

fruffy (approved):
Maybe we can add a form of these static analysis checks to the open-source infrastructure. We do have the Github marketplace available but never used it.

vlstill:
> Maybe we can add a form of these static analysis checks to the open-source infrastructure. We do have the Github marketplace available but never used it.

Yeah, that would be great if there is a good free analyzer. It seems that clang-tidy is quite a bit more limited then coverity, and it is pretty slow as you have pointed out. Maybe there would be a way to run in in scheduled runs (nightly, weekly), or in some sort of differential mode though. I don't know about other options sadly.

fruffy:
> > Maybe we can add a form of these static analysis checks to the open-source infrastructure. We do have the Github marketplace available but never used it.
> 
> Yeah, that would be great if there is a good free analyzer. It seems that clang-tidy is quite a bit more limited then coverity, and it is pretty slow as you have pointed out. Maybe there would be a way to run in in scheduled runs (nightly, weekly), or in some sort of differential mode though. I don't know about other options sadly.

I will take a look. 

P4Testgen will fail because of a failing test that slipped past CI. #4397 should fix it, feel free to merge that one, patch the fix, or just ignore the failure.

vlstill:
It looks like the auto merger is willing to auto-merge with failed optional tests... Is this deliberate?

fruffy:
> It looks like the auto merger is willing to auto-merge with failed optional tests... Is this deliberate?

Well, I also did not expect that. But it makes sense, there might be optional tests that fail frequently and you do not want to make the auto-merge useless because of that. The tools tests are still optional.

PR #4393, opened 2024-02-05
Update MacOS runner to Ventura, add MacOS Sonoma (M1) runner

### LINKED ISSUE #4390: P4C arm64/aarch64 support
This increasingly comes up as a topic. For example, users with M1 MacBook would like to try out the compiler but can not get it to work. It might not be too much effort to fix the issues causing compiler errors.

fruffy:
Fixes #4390. Build P4C for MacOS Sonoma (M1) and Ventura. 
https://github.blog/changelog/2024-01-30-github-actions-macos-14-sonoma-is-now-available/

TODO: Add support for the behavioral model and P4Testgen. I might do that in a separate PR.

### DIFF
--- a/.github/workflows/ci-test.yml
+++ b/.github/workflows/ci-test.yml
@@ -16,8 +16,6 @@ concurrency:
 jobs:
   # Build with gcc and test p4c on Ubuntu 22.04.
   test-ubuntu22:
-    strategy:
-      fail-fast: false
     runs-on: ubuntu-22.04
     env:
       CTEST_PARALLEL_LEVEL: 4
@@ -79,8 +77,6 @@ jobs:
 
   # Build and test p4c on Fedora.
   test-fedora-linux:
-    strategy:
-      fail-fast: false
     # This job runs in Fedora container that runs in Ubuntu VM.
     runs-on: ubuntu-latest
     container:
@@ -116,11 +112,42 @@ jobs:
       run: sudo -E ctest --output-on-failure --schedule-random
       working-directory: ./build
 
-  # Build and test p4c on MacOS 12
+  # Build and test p4c on MacOS for M1 Macs.
+  test-mac-os-m1:
+    runs-on: macos-14
+    env:
+      CTEST_PARALLEL_LEVEL: 4
+    steps:
+    - uses: actions/checkout@v4
+      with:
+        submodules: recursive
+
+    - name: ccache
+      uses: hendrikmuhs/ccache-action@v1
+      with:
+        key: test-${{ runner.os }}
+        max-size: 1000M
+
+    - name: Install dependencies (MacOS)
+      run: |
+           tools/install_mac_deps.sh
+
+    - name: Build (MacOS)
+      run: |
+          source ~/.bash_profile
+          ./bootstrap.sh -DENABLE_GC=ON -DCMAKE_BUILD_TYPE=RELEASE \
+              -DCMAKE_UNITY_BUILD=ON
+          make -Cbuild -j$((`nproc`+1))
+
+    - name: Run tests (MacOS)
+      run: |
+        source ~/.bash_profile
+        ctest --output-on-failure --schedule-random -LE "bpf|ubpf"
+      working-directory: ./build
+
+  # Build and test p4c on MacOS 13 on x86.
   test-mac-os:
-    strategy:
-      fail-fast: false
-    runs-on: macos-13-xlarge
+    runs-on: macos-13
     env:
       CTEST_PARALLEL_LEVEL: 4
     steps:
--- a/.github/workflows/ci-test.yml
+++ b/.github/workflows/ci-test.yml
@@ -128,6 +128,15 @@ jobs:
         key: test-${{ runner.os }}
         max-size: 1000M
 
+    - name: Cache Homebrew Packages
+      id: cache-homebrew-packages
+      uses: actions/cache@v4
+      env:
+        cache-name: homebrew-packages
+      with:
+        path: $(brew --prefix)
+        key: ${{ runner.os }}-${{ hashFiles('tools/install_mac_deps.sh') }}
+
     - name: Install dependencies (MacOS)
       run: |
            tools/install_mac_deps.sh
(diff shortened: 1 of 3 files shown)


tools/install_mac_deps.sh
```diff
@@ -5,30 +5,38 @@
...
-$BREW link ${BOOST_LIB}
-# Prefer Homebrew's bison and grep over the macOS-provided version
-$BREW install bison
-echo 'export PATH="/usr/local/opt/bison/bin:$PATH"' >> ~/.bash_profile
-$BREW install grep
-echo 'export PATH="/usr/local/opt/grep/libexec/gnubin:$PATH"' >> ~/.bash_profile
+# Install some custom requirements on OS X using brew
+BOOST_LIB="boost@1.84"
+brew install autoconf automake bdw-gc ccache cmake \
+      libtool openssl pkg-config coreutils bison grep \
```
asl, line 26: What's about flex? I remember macos shipped some ancient version by default, I do not recall the present status (esp. on builders)

asl: On Ventura laptop I'm having:
```
/usr/bin/flex --version
flex 2.6.4 Apple(flex-34)
```

This seems to be enough, ok.

tools/install_mac_deps.sh
```diff
@@ -5,30 +5,38 @@
...
-# Install some custom requirements on OS X using brew
-BREW=/usr/local/bin/brew
-if [[ ! -x $BREW ]]; then
-    /bin/bash -c "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/HEAD/install.sh)"
-fi
 
-BOOST_LIB="boost@1.76"
+# Set up Homebrew differently for arm64.
+if [[ $(uname -m) == 'arm64' ]]; then
+    (echo; echo 'eval "$(/opt/homebrew/bin/brew shellenv)"') >> ~/.zprofile
```
grg, line 11: Is it a clean environment each time CI runs? I assume it is, but asking in case we need to check for the eval line before adding it to .zprofile.

grg (approved):

.github/workflows/ci-test.yml
```diff
@@ -116,11 +112,42 @@ jobs:
...
+      with:
+        submodules: recursive
+
+    - name: ccache
+      uses: hendrikmuhs/ccache-action@v1
+      with:
+        key: test-${{ runner.os }}
+        max-size: 1000M
+
+    - name: Install dependencies (MacOS)
```
asl, line 131: How about caching installed packages? Something like this:
```
      - name: Cache Homebrew Packages
        id: cache-homebrew-packages
        uses: actions/cache@v3
        env:
          cache-name: homebrew-packages
        with:
          path: $(brew --prefix)
          key: ${{ matrix.os }}-${{ env.cache-name }}-${{ hashFiles('tools/install_mac_deps.sh') }}
```

tools/install_mac_deps.sh
grg: Huh, I have an older version on Sonoma (14.3.1):
```
flex 2.5.35 Apple(flex-32)
```

asl: Probably depends on installed xcode as well

tools/install_mac_deps.sh
fruffy: Do you have a suggestion for a cleaner command to set up `brew`? I do not have a Mac so I am just pasting what people suggest.

fruffy: Ideally, one should be able to run this command locally, too.

asl:
Just side one: in order to test the action before commit the proper way is:
 - Ensure the workflow yaml does exist in `main` branch (it is now)
 - Add `workflow_dispatch` event
 - Now you can trigger it manually from the branch and see if it works.

tools/install_mac_deps.sh
grg: This seems reasonable to me.

tools/install_mac_deps.sh
```diff
@@ -5,30 +5,38 @@
...
 
-# Install some custom requirements on OS X using brew
-BREW=/usr/local/bin/brew
-if [[ ! -x $BREW ]]; then
-    /bin/bash -c "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/HEAD/install.sh)"
-fi
 
-BOOST_LIB="boost@1.76"
+# Set up Homebrew differently for arm64.
+if [[ $(uname -m) == 'arm64' ]]; then
```
asl, line 10: This probably should be guarded to check if brew is not installed (if we're going to suggest that this script could be run standalone and not only as a part of CI). I do not want the line to added each time I am running the script. Though I do not care as I am not using `zsh` :)

fruffy:
> Just side one: in order to test the action before commit the proper way is:
> 
>     * Ensure the workflow yaml does exist in `main` branch (it is)
> 
>     * Add `workflow_dispatch` event
> 
>     * Now you can trigger it manually from the branch and see if it works.

Interesting, never tried this. Although you still need to have your changes uploaded, no? This is a workflow to test without creating a PR or a fork?

asl:
> Interesting, never tried this. Although you still need to have your changes uploaded, no? This is a workflow to test without creating a PR or a fork?

Yes. exactly. Just push the branch and trigger from branch by hands. Might be better and not waste CI cycles for other actions

rst0git (approved):
LGTM

tools/install_mac_deps.sh
fruffy: Used ChatGPT to "enhance" the script with some guards, please take another look.

tools/install_mac_deps.sh
fruffy: So, should I add a custom version? I believe in any case flex should be new enough.

grg: I think the default version is fine for recent versions of MacOS. I've built it locally on my system in the past with 2.5.35.

asl: Yes, default is fine now.

asl (approved):

PR #4394, opened 2024-02-05
Try to fix Github actions syntax.

fruffy:
The condition was always true. Maybe this syntax works.


fruffy:
@jafingerhut Once again request your review on these CI tweaks. Hopefully, this is the final change, this syntax should work now.

jafingerhut (approved):

PR #4396, opened 2024-02-05
Do not check for shadowing repeatedly in inlining passes.

kfcripps:
We found that checking shadowing in `ResolveReferences` was a big bottleneck for one of our apps, and it is done each time an inlining pass is repeated. We turned this off and added a new `ResolveReferences` pass to check shadowing after all inlining is completed. This greatly helped to speed up these passes, which were taking the longest.

Compilation times for one of our apps without these changes:
```
real    22m7.669s
user    21m55.887s
sys     0m14.823s
```

and with them:
```
real    16m36.850s
user    16m24.559s
sys     0m15.304s
```

### DIFF
--- a/frontends/p4/actionsInlining.h
+++ b/frontends/p4/actionsInlining.h
@@ -51,9 +51,7 @@ class ActionsInliner : public AbstractInliner<ActionsInlineList, AInlineWorkList
     std::map<const IR::MethodCallStatement *, const IR::P4Action *> *replMap;
 
  public:
-    explicit ActionsInliner(bool isv1) : refMap(new P4::ReferenceMap()), replMap(nullptr) {
-        refMap->setIsV1(isv1);
-    }
+    explicit ActionsInliner(P4::ReferenceMap *refMap) : refMap(refMap), replMap(nullptr) {}
     Visitor::profile_t init_apply(const IR::Node *node) override;
     const IR::Node *preorder(IR::P4Parser *cont) override {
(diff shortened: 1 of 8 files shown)


kfcripps:
tagging @asl @fruffy

frontends/p4/actionsInlining.h
```diff
@@ -51,9 +51,7 @@ class ActionsInliner : public AbstractInliner<ActionsInlineList, AInlineWorkList
     std::map<const IR::MethodCallStatement *, const IR::P4Action *> *replMap;
 
  public:
-    explicit ActionsInliner(bool isv1) : refMap(new P4::ReferenceMap()), replMap(nullptr) {
-        refMap->setIsV1(isv1);
```
fruffy, line 55: Any reason why you are removing the isv1 checks? I think we should deprecated that strange check eventually but it seems like you may change some semantics here.

grg: I think this is because the ReferenceMap being passed in already has the isv1 flag set appropriately.

fruffy: Ah yeah, makes sense.

kfcripps: Yes, that's correct.

fruffy (approved):
@grg Any concerns on this PR?

ChrisDodd (approved):

PR #4387, opened 2024-02-01
[P4Testgen] Hotfixes and improvements to the P4Constraints parsers.

fruffy:
- Add a small builtin map to the refers_to parser. Builtin objects are prefixed by `builtin::`
- Improve the readability and error handling of the refers_to parser. 
- Fix a crash in the `entry_restriction` parser when comparing to a negative constant. 
- Add support for `referenced_by` and `action_restriction`. 
- Add some type aliases.

PR #4397, opened 2024-02-06
[P4Testgen] Fix failing P4Testgen gtest.

fruffy:
I missed fixing a check in #4387.

std:vector<std::vector>> was flattened into std::vector<>, which is why this check is failing. We have a structure of `[a,b,c]` instead of `[[a,b,c]]` now.


vlstill:
What is the cause of the change? With just the change, I am not able to check this is a correct fix.

fruffy:
IN #4387 I flattened `std:vector<std::vector>>` into `std::vector<>`. There was no reason for it to be 2-dimensional, but that required fixing some checks in the gtests.

kfcripps (approved):

PR #4399, opened 2024-02-07
Implict cast fix

ChrisDodd:
- Allow implicit casts between serializable enums and their underlying type in action calls

### DIFF
--- a/frontends/p4/typeChecking/typeChecker.cpp
+++ b/frontends/p4/typeChecking/typeChecker.cpp
@@ -3493,7 +3493,7 @@ const IR::Expression *TypeInference::actionCall(bool inActionList,
         if (paramType == nullptr || argType == nullptr)
             // type checking failed before
             return actionCall;
-        constraints.addEqualityConstraint(actionCall, paramType, argType);
+        constraints.addImplicitCastConstraint(actionCall, paramType, argType);
         if (param->direction == IR::Direction::None) {
             if (inActionList) {
                 typeError("%1%: parameter %2% cannot be bound: it is set by the control plane", arg,
@@ -3508,6 +3508,15 @@ const IR::Expression *TypeInference::actionCall(bool inActionList,
                 if (!isCompileTimeConstant(arg->expression))
                     typeError("%1%: action argument must be a compile-time constant",
                               arg->expression);
+            }
+            // This is like an assignment; may make additional conversions.
+            newExpr = assignment(arg, param->type, arg->expression);
+            if (readOnly) {
+                // FIXME -- if we're in readonly mode, we should not have introduced any mods
+                // here, but there's a bug in the DPDK backend where it generates a ListExpression
+                // that would be converted to a StructExpression, and other problems where it
+                // can't deal with that StructExpressions, so we hack to avoid breaking those tests
+                newExpr = arg->expression;
             }
         } else if (param->direction == IR::Direction::Out ||
                    param->direction == IR::Direction::InOut) {
--- a/frontends/p4/typeChecking/typeConstraints.cpp
+++ b/frontends/p4/typeChecking/typeConstraints.cpp
@@ -31,6 +31,12 @@ void TypeConstraints::addEqualityConstraint(const IR::Node *source, const IR::Ty
     add(c);
 }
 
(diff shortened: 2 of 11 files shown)


vlstill:
What part of the spec is this implementing? As far as I see [8.11.2](https://p4.org/p4-spec/docs/P4-16-v1.2.4.html#sec-implicit-casts) of the v1.2.4 spec allows only implicit casts of serializable enum -> underlying type, but not the other way around. Your test seems to imply you have now allowed both.

ChrisDodd:
Hmm -- it seems weird to allow implicit casts one way and not the other.  I was originally just looking at the implicit cast from serializable enum to underlying type, but the way typechecking works, it ends up being symmetric.

vlstill:
> Hmm -- it seems weird to allow implicit casts one way and not the other. I was originally just looking at the implicit cast from serializable enum to underlying type, but the way typechecking works, it ends up being symmetric.

I would argue that for implicit casts the spec makes sense -- a seralizable-enum-typed variable is supposed to contain one of the named values, but an implicit cast from the underlying type would make it very easy to accidentally create a value that is not one of the named once. I think such a thing should be only for explicit casts.

ChrisDodd:
Ok, I've managed to figure out how to get this to allow the implicit casts one way and not the other -- there's some confusion here between type unification (which is symmetric) and type checking (which is not), so it needs an extra type check to disallow the 'wrong' conversion.

frontends/p4/typeChecking/typeChecker.cpp
```diff
@@ -3512,6 +3512,15 @@ const IR::Expression *TypeInference::actionCall(bool inActionList,
                     typeError("%1%: action argument must be a compile-time constant",
                               arg->expression);
             }
+            // This is like an assignment; may make additional conversions.
+            newExpr = assignment(arg, param->type, arg->expression);
+            if (readOnly) {
```
fruffy, line 3517: Can you elaborate on that bug? Maybe we can fix the DPDK back end.

ChrisDodd: Basically, the DPDK backend creates new code that is not type-correct, so a following TypeChecking pass tries to introduce new implicit casts, but fails as the code is `readOnly`.  I'm not sure why it only fails after this change, as it seems like it should fail regardless.

I tried adding a additional TypeInferencing pass in the DPDK backend to introduce the implicit casts correctly, but that caused some other problems.  This feels like a bit of a hack, leaving the code as semi-type-incorrect (not making the implicit cast explicit) when the readOnly flag is set.

fruffy: Which pass is this? Maybe we can fix the generation of this action call in the first place?

I think this is something the DPDK back end needs to fix before we add more workarounds to the type checker. The class is already complicated.

ChrisDodd: The problem is in DPDK::PrependPDotToActionArgs (line ~1318) where it is creating a ListExpression rather than a StructExpression to initialize a struct.  The typechecker wants this to be a StructExpression, so will introduce a "cast" (actually rewrite the ListExpression as a StructExpression), which then causes a problem as it is running with `readOnly = true` (backend.cpp line 107).  If I change that call to have readOnly = false, then it doesn't assert and just makes the change.  However, it then has problems later in the backend with code that doesn't know what to do with a StructExpression.

Without my change, the typeChecker does not correctly type check action args, so it doesn't try to insert the conversion.

The easiest fix is probably just to add a call to P4::TypeInference with readOnly = false somewhere, and fix the backend code to understand a StructExpression.

fruffy: @usha1830 Could you take a look at this?

fruffy: #4553

fruffy (approved):
Tentatively approving. @vlstill any follow-up?

vlstill (approved):

frontends/p4/typeChecking/typeChecker.cpp
usha1830: Sure. I will check this out.

PR #4400, opened 2024-02-07
[P4Testgen] Fix generation of specific fields in the Protobuf-IR test backend.

fruffy-g:
Also move  the hex separator utility to the common class.


backends/p4tools/modules/testgen/targets/bmv2/test_backend/protobuf_ir.cpp
```diff
@@ -168,7 +171,21 @@ entities {
 
 std::string ProtobufIr::formatNetworkValue(const std::string &type, const IR::Expression *value) {
     if (type == "hex_str") {
-        return formatHexExpr(value, {false, true, false, false});
+        return formatHexExpr(value, {false, true, true, false});
+    }
+    // Assume that any string format can be converted from a string literal, bool
+    // literal, or constant.
+    if (type == "str") {
```
smolkaj, line 178: Eventually we would want to apply some sort of translation here. Should we leave a TODO in the meantime?

smolkaj (approved):

PR #4401, opened 2024-02-08
[P4Testgen] Add an option to override the test name. 

fruffy:


backends/p4tools/modules/testgen/options.h
```diff
@@ -90,6 +91,10 @@ class TestgenOptions : public AbstractP4cToolOptions {
     /// Specifies minimum coverage that needs to be achieved for P4Testgen to exit successfully.
     float minCoverage = 0;
 
+    /// The base name of the tests which are generated.
+    /// Defaults to the name of the input program, if provided.
+    std::optional<cstring> testName;
```
smolkaj, line 96: Nit: If it is the base name, perhaps it would be good for the variable to be named accordingly? `testBaseName`?

smolkaj (approved):
Nice, LGTM

PR #4402, opened 2024-02-08
ConstantFolding hook to allow backends to do special stuff

grg:
Hook allows skipping of PathExpression processing. Allows backends to do custom processing.

### DIFF
--- a/frontends/common/constantFolding.cpp
+++ b/frontends/common/constantFolding.cpp
@@ -23,6 +23,8 @@ limitations under the License.
 
 namespace P4 {
 
+ConstantFolding::filter_hook_t ConstantFolding::filter_hook;
+
 class CloneConstants : public Transform {
  public:
     CloneConstants() = default;
@@ -87,6 +89,9 @@ const IR::Expression *DoConstantFolding::getConstant(const IR::Expression *expr)
 }
 
 const IR::Node *DoConstantFolding::postorder(IR::PathExpression *e) {
+    if (ConstantFolding::filter_hook) {
+        if (auto *rv = ConstantFolding::filter_hook(this, e)) return rv;
+    }
     if (refMap == nullptr || assignmentTarget) return e;
     auto decl = refMap->getDeclaration(e->path);
     if (decl == nullptr) return e;
--- a/frontends/common/constantFolding.h
+++ b/frontends/common/constantFolding.h
@@ -163,10 +163,21 @@ class ConstantFolding : public PassManager {
         setName("ConstantFolding");
     }
 
+    /** Function definition for hook to filter which Expression objects are processed.
+     *
+     * Function should return nullptr to process the Expression, or a non-null pointer
+     * to bypass processing. Returned value will be returned by the postorder function.
+     */
     typedef std::function<const IR::Expression *(Visitor *, const IR::Expression *)>
         filter_hook_t;
+
+    /** Active hook to filter which PathExpression objects are processed */
     static filter_hook_t filter_hook;
+
+    /** Set the filter hook to programatically skip processing of select expressions */
     static void setFilterHook(filter_hook_t filter) { filter_hook = filter; }
+
+    /** Clear the filter hook (process all expressions) */
     static void unsetFilterHook() { filter_hook = filter_hook_t(); }
 };
 
--- a/frontends/common/constantFolding.h
+++ b/frontends/common/constantFolding.h
@@ -163,20 +163,19 @@ class ConstantFolding : public PassManager {
         setName("ConstantFolding");
     }
 
-    /** Function definition for hook to filter which Expression objects are processed.
-     *
-     * Function should return nullptr to process the Expression, or a non-null pointer
-     * to bypass processing. Returned value will be returned by the postorder function.
-     */
+    /// Function definition for hook to filter which Expression objects are processed.
+    ///
+    /// Function should return nullptr to process the Expression, or a non-null pointer
+    /// to bypass processing. Returned value will be returned by the postorder function.
     typedef std::function<const IR::Expression *(Visitor *, const IR::Expression *)> filter_hook_t;
 
-    /** Active hook to filter which PathExpression objects are processed */
+    /// Active hook to filter which PathExpression objects are processed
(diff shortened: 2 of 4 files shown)


frontends/common/constantFolding.h
```diff
@@ -162,6 +162,23 @@ class ConstantFolding : public PassManager {
         if (typeMap != nullptr) passes.push_back(new ClearTypeMap(typeMap));
         setName("ConstantFolding");
     }
+
+    /** Function definition for hook to filter which Expression objects are processed.
```
fruffy, line 166: Optional nit: Even in this class the comments are mixed but at least on the open-source side we have been increasingly using three slashes `///`.

frontends/common/constantFolding.h
```diff
@@ -162,6 +162,23 @@ class ConstantFolding : public PassManager {
...
+     *
+     * Function should return nullptr to process the Expression, or a non-null pointer
+     * to bypass processing. Returned value will be returned by the postorder function.
+     */
+    typedef std::function<const IR::Expression *(Visitor *, const IR::Expression *)>
```
fruffy, line 171: `using`?

fruffy (approved):
A test or usage example would be great.

frontends/common/constantFolding.h
grg: I'm happy to switch to `///`

grg:
> A test or usage example would be great.

Done. (It's not enough to say we use it internally within Intel? 😜 )

grg:
@fruffy Could you take a brief look at the test please?

frontends/common/constantFolding.h
```diff
@@ -162,6 +162,21 @@ class ConstantFolding : public PassManager {
         if (typeMap != nullptr) passes.push_back(new ClearTypeMap(typeMap));
         setName("ConstantFolding");
     }
+
+    /// Function definition for hook to filter which Expression objects are processed.
+    ///
+    /// Function should return nullptr to process the Expression, or a non-null pointer
+    /// to bypass processing. Returned value will be returned by the postorder function.
+    using filter_hook_t = std::function<const IR::Expression *(Visitor *, const IR::Expression *)>;
```
vlstill, line 170: I think it would be better to name the type `FilterHook` to match the naming conventions.

frontends/common/constantFolding.h
```diff
@@ -162,6 +162,21 @@ class ConstantFolding : public PassManager {
         if (typeMap != nullptr) passes.push_back(new ClearTypeMap(typeMap));
         setName("ConstantFolding");
     }
+
+    /// Function definition for hook to filter which Expression objects are processed.
+    ///
+    /// Function should return nullptr to process the Expression, or a non-null pointer
+    /// to bypass processing. Returned value will be returned by the postorder function.
```
vlstill, line 169: Can you give an example how this would be used? Are you trying to modify a behavior in some of the standard pass managers? Or are you modifying it in your midend/backend? For the second case, it would make significantly more sense to derive a backend-specific version of constant folding from the generic one.

frontends/common/constantFolding.cpp
```diff
@@ -87,6 +89,9 @@ const IR::Expression *DoConstantFolding::getConstant(const IR::Expression *expr)
 }
 
 const IR::Node *DoConstantFolding::postorder(IR::PathExpression *e) {
+    if (ConstantFolding::filter_hook) {
+        if (auto *rv = ConstantFolding::filter_hook(this, e)) return rv;
```
vlstill, line 93: I don't like at all that the filter hook is taking `IR::Expression`, but hooking only to the `IR::PathExpression` case.

vlstill (changes requested):
@grg, could you please provide more details on your use case? I'll try to think about a better solution. I don't think this is a good change. Using a static hook the is set at completely different place then the pass is actually used seems quite dangerous and messy to me.

vlstill:
Hmm, this is used in frontend, which is shared by all p4c-based tools. I thin a possible solution would be to implement a `FrontendPolicy` class, that contains sub-polies for selected passes and some other frontend settings. I propose it would contain:

- newly introduced policy for `ConstantFolding`
- the `skipSideEffectOrdering` flag
- a policy for optilizations (whether to disable inlining if `-O0` is given)

I'll try to draft it if I have some time after work today, it should be a relatively small change and would clean up the frontend overall I believe.

test/gtest/constant_folding.cpp
```diff
@@ -0,0 +1,132 @@
...
+
+// Constant folding with filtering active
+TEST_F(P4CConstantFoldingValidation, filter) {
+    // Add a filter to skip processing of some structs
+    P4::ConstantFolding::setFilterHook(ConstantFoldFilter());
+
+    auto *program = parseAndProcess(constantFoldFilterCode)->to<IR::P4Program>();;
+    ASSERT_TRUE(program);
+
+    // my_hdr_stack_1_t.hdr should be a Constant
```
fruffy, line 116:
```suggestion
    // my_hdr_stack_1_t.hdr should not be a Constant
```

fruffy (commented):
> @fruffy Could you take a brief look at the test please?

Looks good to me, thanks! You should be able to run `make clang-format-fix-errors` to fix the linting errors.

frontends/common/constantFolding.cpp
ChrisDodd: Yes, I was of mixed mind about this.  Hooking everything would be more expensive, having every method call the hook and most of the time returning nullptr.  The compromise of only calling for PathExpression should probably also change to having the hook function take a PathExpression rather than an Expression.

ChrisDodd (approved):
While I wrote this code originally, I was never really happy with it.

I particularly don't like making this a "global" hook rather than a per visitor instance hook, but that would require passing the hook to P4::FrontEnd::FrontEnd to get it applied there.  But #4406 would allow that, so may be the way to go.

grg:
Closing as this has been superseded by #4406

grg:
Closing as this has been superseded by https://github.com/p4lang/p4c/pull/4406

PR #4405, opened 2024-02-09
Create a dependabot.yml for Github actions

fruffy:
No one keeps track of Github actions versioning so try to refresh them automatically using Dependabot. There is little harm in automating this because the bot effectively checks itself using CI when updating a dependency.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,11 @@
+# To get started with Dependabot version updates, you'll need to specify which
+# package ecosystems to update and where the package manifests are located.
+# Please see the documentation for all configuration options:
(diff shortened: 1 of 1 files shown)


fruffy:
@jafingerhut Pinging you here because I have requested reviews on other Github Action related items from you recently. 

I am adding this because we have a couple warnings in the action workflows regarding actions that are out-of-date. 

image

fruffy:
@jafingerhut Could you give this a look? It is the same thing running in the behavioral model repo
https://github.com/p4lang/behavioral-model/blob/main/.github/dependabot.yml

jafingerhut (approved):
LGTM.  Confirmed that this is the same configuration we are currently using for p4lang/behavioral-model repo.

PR #4407, opened 2024-02-09
allow json output format to be modified

grg:
Expose JsonPrintOptions to allow JSON output format to be modified.


grg:
Many of the tests are failing because dependencies can't be fixed:
```
E: Failed to fetch http://download.opensuse.org/repositories/home:/p4lang/xUbuntu_20.04/amd64/p4lang-pi_0.1.0-15_amd64.deb  500  Internal Server Error [IP: 195.135.223.226 80]
Fetched 14.3 MB in 2s (7397 kB/s)
E: Failed to fetch http://download.opensuse.org/repositories/home:/p4lang/xUbuntu_20.04/amd64/p4lang-bmv2_1.15.0-7_amd64.deb  500  Internal Server Error [IP: 195.135.223.226 80]
```

What do we need to do to resolve this? Or is it just a case of wait for something on the source to update?

fruffy:
> Many of the tests are failing because dependencies can't be fixed:
> 
> ```
> E: Failed to fetch http://download.opensuse.org/repositories/home:/p4lang/xUbuntu_20.04/amd64/p4lang-pi_0.1.0-15_amd64.deb  500  Internal Server Error [IP: 195.135.223.226 80]
> Fetched 14.3 MB in 2s (7397 kB/s)
> E: Failed to fetch http://download.opensuse.org/repositories/home:/p4lang/xUbuntu_20.04/amd64/p4lang-bmv2_1.15.0-7_amd64.deb  500  Internal Server Error [IP: 195.135.223.226 80]
> ```
> 
> What do we need to do to resolve this? Or is it just a case of wait for something on the source to update?

@rst0git The opensuse repository is quite flaky and we run into this a lot in CI. Do you know a more stable repository we could migrate to? What is the process like?

rst0git:
> Do you know a more stable repository we could migrate to? What is the process like?

Sure, we can use [Launchpad](https://launchpad.net/~p4lang) instead. I'll update the packages in Launchpad and open a pull request for the installation process in CI.

fruffy:
> > Do you know a more stable repository we could migrate to? What is the process like?
> 
> Sure, we can use [Launchpad](https://launchpad.net/~p4lang) instead. I'll update the packages in Launchpad and open a pull request for the installation process in CI.

Great, thank you!

ChrisDodd (approved):

grg:
The Mac build in test suite fails as follows:
```
In file included from /Users/runner/work/p4c/p4c/frontends/common/options.h:24:
/Users/runner/work/p4c/p4c/control-plane/p4RuntimeSerializer.h:23:10: fatal error: 'google/protobuf/util/json_util.h' file not found
#include <google/protobuf/util/json_util.h>
         ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
I'm going to change a class instance to a pointer to avoid this. But for my own education, how would I fix this if I wasn't going to switch to a pointer.

(The pointer allows me to just write `class ClassINeedFromHeaderFile;` instead of importing the whole header file.)

fruffy:
> I'm going to change a class instance to a pointer to avoid this. But for my own education, how would I fix this if I wasn't going to switch to a pointer.

Honestly, I am not sure why this one is failing. It could be a CLang vs GCC issue. I enabled the sanitizer, which uses CLang instead of GCC to check that.

grg:
> > I'm going to change a class instance to a pointer to avoid this. But for my own education, how would I fix this if I wasn't going to switch to a pointer.
> 
> Honestly, I am not sure why this one is failing. It could be a CLang vs GCC issue. I enabled the sanitizer, which uses CLang instead of GCC to check that.

I pushed a change that switches to a pointer instead. Don't entirely know if I like it (I need to create a JsonPrintOptions object in one of the constructors for P4Runtime). It's the most recent commit.

BTW, if you want to test my PR without this change, feel free to temporary back out the last commit.

grg:
> > I'm going to change a class instance to a pointer to avoid this. But for my own education, how would I fix this if I wasn't going to switch to a pointer.
> 
> Honestly, I am not sure why this one is failing. It could be a CLang vs GCC issue. I enabled the sanitizer, which uses CLang instead of GCC to check that.

Following up, I think that we need to add a `target_include_directories` call in cmake files for any cpp files that directly or indirectly need the Protobuf includes. Below is a sample from backends/p4tools/common/CMakeLists.txt. I'm guessing that the Mac OS protobuf directory is outside of the "default" include directories for the build, whereas the Linux location is.
```
target_include_directories(
  p4tools-control-plane
  SYSTEM BEFORE PUBLIC ${Protobuf_INCLUDE_DIRS}
)
```

I'd prefer not to have to extend the set of sources that require the protobuf includes, unless we chose to include it for some large chunk of files (e.g., entire build, all frontend files, ...).

grg:
> What confuses me is that this include should be exported by the controlplane dependency. I will try to take a look this week.

Any opinion on whether it's better with or without the 3rd commit?

fruffy:
> > What confuses me is that this include should be exported by the controlplane dependency. I will try to take a look this week.
> 
> Any opinion on whether it's better with or without the 3rd commit?

I think it is better without. So far I am unable to reproduce this. I do not have a Mac.

grg:
> > > What confuses me is that this include should be exported by the controlplane dependency. I will try to take a look this week.
> > 
> > 
> > Any opinion on whether it's better with or without the 3rd commit?
> 
> I think it is better without. So far I am unable to reproduce this. I do not have a Mac.

Here's what I had to change on a Mac to get it to build without the last commit:
```
diff --git a/backends/bmv2/CMakeLists.txt b/backends/bmv2/CMakeLists.txt
index 034e2a095..30c3cc492 100644
--- a/backends/bmv2/CMakeLists.txt
+++ b/backends/bmv2/CMakeLists.txt
@@ -84,6 +84,9 @@ set (IR_DEF_FILES ${IR_DEF_FILES} ${CMAKE_CURRENT_SOURCE_DIR}/bmv2.def PARENT_SC
 
 add_library(bmv2backend STATIC ${BMV2_BACKEND_COMMON_SRCS})
 add_dependencies(bmv2backend ir-generated frontend)
+target_include_directories(bmv2backend
+  SYSTEM BEFORE PUBLIC ${Protobuf_INCLUDE_DIRS}
+)
 
 add_executable(p4c-bm2-ss ${BMV2_SIMPLE_SWITCH_SRCS})
 target_link_libraries (p4c-bm2-ss bmv2backend ${P4C_LIBRARIES} ${P4C_LIB_DEPS})
diff --git a/frontends/CMakeLists.txt b/frontends/CMakeLists.txt
index c2704d246..1c505c76b 100644
--- a/frontends/CMakeLists.txt
+++ b/frontends/CMakeLists.txt
@@ -240,6 +240,9 @@ add_parser(p4)
 
 add_library (frontend-parser-gen OBJECT ${p4PARSER_GEN_SRCS} ${v1PARSER_GEN_SRCS})
 add_dependencies (frontend-parser-gen mkp4dirs mkv1dirs ir-generated)
+target_include_directories(frontend-parser-gen
+  SYSTEM BEFORE PUBLIC ${Protobuf_INCLUDE_DIRS}
+)
 
 set (FRONTEND_SOURCES
   ${COMMON_FRONTEND_SRCS}
```

backends/p4tools/modules/testgen/targets/bmv2/bmv2.h
```diff
@@ -4,6 +4,7 @@
 #include "backends/p4tools/common/compiler/compiler_target.h"
 #include "backends/p4tools/common/compiler/midend.h"
 #include "backends/p4tools/common/lib/variables.h"
+#include "control-plane/P4RuntimeSerializer.h"
```
fruffy, line 7:
```suggestion
#include "control-plane/p4RuntimeSerializer.h"
```

PR #4409, opened 2024-02-11
Propagate source info correctly in compiler passes

fruffy:
The way SideEffectOrdering and other passes propagates source information causes it to duplicate the source information in some cases. Ideally, we want to point to either the left- or right-hand side of an expression once we have factored it out into a statement.


frontends/p4/sideEffects.cpp
```diff
@@ -299,8 +300,9 @@ const IR::Node *DoSimplifyExpressions::preorder(IR::Mux *expression) {
     auto ifFalse = statements;
     statements = save;
 
-    auto ifStatement = new IR::IfStatement(expression->e0, new IR::BlockStatement(ifTrue),
-                                           new IR::BlockStatement(ifFalse));
+    auto ifStatement =
+        new IR::IfStatement(expression->e0->srcInfo, expression->e0, new IR::BlockStatement(ifTrue),
```
vlstill, line 304: This one looks questionable to me, but I guess both options work reasonably well.

vlstill (commented):
Looks good to me, but why did the tests fail?

fruffy:
It looks like the problem is the predication pass, which does not preserve source information well. Thankfully, we do not use it in P4Testgen, because it has [problems](https://github.com/p4lang/p4c/issues?q=is%3Aissue+predication+label%3A%22Predication+pass%22).

frontends/p4/functionsInlining.cpp
```diff
@@ -262,20 +262,22 @@ const IR::Statement *FunctionsInliner::inlineBefore(const IR::Node *calleeNode,
     for (auto param : callee->type->parameters->parameters) {
         auto argument = substitution.lookup(param);
         cstring newName = nameGen->newName(param->name.name.string_view());
+        auto name = IR::ID(argument->expression->srcInfo, newName);
```
ChrisDodd: Shouldn't this use `param->srcInfo` rather than `argument->expression->srcInfo` (the source location of the parameter in the function being inlined, rather than the source location of the argument in the call site?)

A number of places in this inlining code look they're confusing the source location(s) the function being inlined with the source locations of the call site.  Need to carefully consider in each case which is more appropriate.  This seems to be a problem independent of this change.

frontends/p4/sideEffects.cpp
```diff
@@ -15,30 +15,35 @@
 
 namespace P4 {
 
-cstring DoSimplifyExpressions::createTemporary(const IR::Type *type) {
+cstring DoSimplifyExpressions::createTemporary(Util::SourceInfo srcInfo, const IR::Type *type) {
```
ChrisDodd: If you're adding a srcInfo for the new tempvar, then this function should probably return the `IR::ID` created (including the srcInfo) rather than just a `cstring`.  Not sure how that interacts with what you've added to `addExpression` however.

testdata/p4_16_samples_outputs/gauntlet_infinite_loop.p4-stderr
```diff
@@ -22,13 +22,12 @@ gauntlet_infinite_loop.p4(29): [--Wwarn=uninitialized-use] warning: padding.p ma
...
-[--Wwarn=uninitialized-use] warning: tmp_0.p may be uninitialized
+gauntlet_infinite_loop.p4(43): [--Wwarn=uninitialized-use] warning: tmp.nop may be uninitialized
+        sub_parser.apply(packet, hdr, hdr.p);
+                                 ^^^
+gauntlet_infinite_loop.p4(43): [--Wwarn=uninitialized-use] warning: tmp_0.p may be uninitialized
+        sub_parser.apply(packet, hdr, hdr.p);
+                                      ^^^^^
+gauntlet_infinite_loop.p4(43): [--Wwarn=invalid-header] warning: accessing a field of an invalid header tmp_0
+        sub_parser.apply(packet, hdr, hdr.p);
+                                      ^^^^^
```
ChrisDodd, line 33: while having line numbers in these error messages is definitely better than not having them, I'm not convinced that these are the *right* linenumbers.  The last one defintely seems off -- the only correct line for this warning is 29 (which is already being output above), so perhaps these are spurious warnings for stuff that has already been warned about.  Having the names `tmp` and `tmp_0` in the messages (when these names appear nowhere in the source) is definitely bad,

ChrisDodd (commented):
While having better srcInfo for messages is definitely good, it is very hard to test (and we need more testcases).  So this seems like it might be an improvement, it is hard to tell.

testdata/p4_16_samples_outputs/gauntlet_infinite_loop.p4-stderr
fruffy: There seems to be an issue with the uninitialized reads in simplifyDefUse, planning to address that in a follow-up.

frontends/p4/functionsInlining.cpp
fruffy: Reverted those changes for a separate PR, to make it easier to review.

testdata/p4_16_samples_outputs/complex10-midend.p4
```diff
@@ -11,13 +11,13 @@ control c(inout bit<32> r) {
     @hidden action complex10l12_0() {
         tmp_1 = false;
     }
-    @hidden action act() {
+    @hidden action complex10l12_1() {
```
fruffy, line 14: These changes seem to happen because the `actionSynthesis` pass now has access to the correct source info.

ChrisDodd: Yes, `actionSynthesis` names the actions based on the srcInfo of the statement being turned into an action, FileBaseName + "l" + LineNumber;  the generic name `act` is used when the statement has no srcInfo.  This name is then passed to `makeUnique`, which will add a `_N` suffix if the name already appears elsewhere in the program.

testdata/p4_16_samples_outputs/gauntlet_infinite_loop.p4-stderr
fruffy: #5779

frontends/p4/functionsInlining.cpp
fruffy: #5781 

I created stacked PRs that can be folded.

PR #4413, opened 2024-02-12
Add buildifier to Bazel workspace and run it on the bazel build files.

fruffy:
Maybe we can make this part of CI, but I am not quite sure yet how to do this.

Also the indirect test currently fails because it tries to pull in buildifier also. Is there a way to make this check optional?

### DIFF
--- a/.github/workflows/ci-bazel.yml
+++ b/.github/workflows/ci-bazel.yml
@@ -38,6 +38,9 @@ jobs:
         chmod +x $BAZEL
         sudo mv $BAZEL /usr/local/bin/bazel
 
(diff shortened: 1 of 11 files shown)


smolkaj (approved):
Very nice!

Ideally we would also add a CI script to enforce formatting.
But this is a strict incremental improvement.

fruffy:
> Very nice!
> 
> Ideally we would also add a CI script to enforce formatting. But this is a strict incremental improvement.

Done, please give the changes another look.

smolkaj (approved):
Awesome, LGTM.

PR #4414, opened 2024-02-12
Add IR utility that helps with return of multiple statements from Transform

vlstill:
Example of use is in the test: https://github.com/p4lang/p4c/pull/4414/files#diff-c05e1bbc752f9a1299cec11e3e39374bcd48de0eb2939bdca8b4fca6c0aa65d8R51.

### DIFF
--- a/ir/irutils.cpp
+++ b/ir/irutils.cpp
@@ -282,7 +282,7 @@ const IR::Node *inlineBlockImpl(const Transform &t, Stmts &&stmts) {
             return stmt;
         }
     }
-    if (t.getParent<const IR::Node>()->is<IR::BlockStatement>()) {
+    if (t.getParent<IR::BlockStatement>()) {
         return new IR::IndexedVector<IR::StatOrDecl>(std::forward<Stmts>(stmts));
     }
     Util::SourceInfo srcInfo;
--- a/ir/irutils.cpp
+++ b/ir/irutils.cpp
@@ -277,7 +277,10 @@ std::vector<const Expression *> flattenListOrStructExpression(const Expression *
 template <typename Stmts>
 const IR::Node *inlineBlockImpl(const Transform &t, Stmts &&stmts) {
     if (stmts.size() == 1) {
-        return *stmts.begin();
+        // it could also be a declaration, and it that case, we need to wrap it in a block anyway
+        if (auto *stmt = (*stmts.begin())->template to<IR::Statement>()) {
+            return stmt;
+        }
     }
     if (t.getParent<const IR::Node>()->is<IR::BlockStatement>()) {
         return new IR::IndexedVector<IR::StatOrDecl>(std::forward<Stmts>(stmts));
--- a/ir/irutils.cpp
+++ b/ir/irutils.cpp
@@ -16,6 +16,7 @@
 #include "ir/indexed_vector.h"
 #include "ir/ir.h"
 #include "ir/vector.h"
+#include "ir/visitor.h"
 #include "lib/exceptions.h"
 #include "lib/rtti.h"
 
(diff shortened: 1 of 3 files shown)


test/gtest/transforms.cpp
```diff
@@ -43,4 +44,38 @@ TEST_F(P4C_IR, Transform) {
...
+        }
+
+        const IR::Constant *c;
+    };
+
+    const auto *c2 = new IR::Constant(2);
+    const auto *c4 = new IR::Constant(4);
+    const auto *pe = new IR::PathExpression(new IR::Path("foo"));
+    const auto *stmt = new IR::BlockStatement(IR::IndexedVector<IR::StatOrDecl>(
+        {new IR::AssignmentStatement(pe, c2),
```
fruffy, line 63: You can get nicer formatting by using a trailing comma.

https://lists.llvm.org/pipermail/cfe-commits/Week-of-Mon-20140317/101748.html

ir/irutils.h
```diff
@@ -108,6 +115,27 @@ big_int getMaxBvVal(int bitWidth);
 // This is 0 for unsigned and -(2^(t->size - 1)) for signed.
 big_int getMinBvVal(const Type *t);
 
+/// A helper useful when returning multiple statmenets out of a Transform's function that should
```
asl, line 118: Typo, `statements`

test/gtest/transforms.cpp
vlstill: Interesting, I did not even know that the trailing comma is accepted there.

ir/irutils.h
```diff
@@ -108,6 +115,27 @@ big_int getMaxBvVal(int bitWidth);
 // This is 0 for unsigned and -(2^(t->size - 1)) for signed.
 big_int getMinBvVal(const Type *t);
 
+/// A helper useful when returning multiple statemenets out of a Transform's function that should
```
asl, line 118: still a typo – `statemenets` :)

vlstill: oh my, sorry

ir/irutils.cpp
```diff
@@ -272,4 +273,36 @@ std::vector<const Expression *> flattenListOrStructExpression(const Expression *
...
 
+template <typename Stmts>
+const IR::Node *inlineBlockImpl(const Transform &t, Stmts &&stmts) {
+    if (stmts.size() == 1) {
+        // it could also be a declaration, and it that case, we need to wrap it in a block anyway
+        if (auto *stmt = (*stmts.begin())->template to<IR::Statement>()) {
+            return stmt;
+        }
+    }
+    if (t.getParent<const IR::Node>()->is<IR::BlockStatement>()) {
```
ChrisDodd, line 284: could be just `if (t.getParent<IR::BlockStatement>())`

ChrisDodd:
The way I've dealt with this in other passes is to just return a BlockStatement then use SimplifyControlFlow, but this is more efficient.

ir/irutils.cpp
vlstill: Right, for some reason I thought `getParent` fails if the type is not right, I probably confused it with `getOriginal`.

ir/irutils.cpp
```diff
@@ -272,4 +273,36 @@ std::vector<const Expression *> flattenListOrStructExpression(const Expression *
...
+        // it could also be a declaration, and it that case, we need to wrap it in a block anyway
+        if (auto *stmt = (*stmts.begin())->template to<IR::Statement>()) {
+            return stmt;
+        }
+    }
+    if (t.getParent<IR::BlockStatement>()) {
+        return new IR::IndexedVector<IR::StatOrDecl>(std::forward<Stmts>(stmts));
+    }
+    Util::SourceInfo srcInfo;
+    if (stmts.size() > 0) {  // no .empty in initializer_list!
```
fruffy, line 288: `std::empty`? I guess that won't work for the other two...

fruffy (approved):

ir/irutils.cpp
vlstill: It should work for anything that has .empty on c-style array and on initializer_list. But I could not decide if it is nicer or uglier than size() > 0.

PR #4416, opened 2024-02-12
[P4Testgen] Add an option to only generate tests with dropped packets. 

fruffy:
Add an option to P4Testgen to only generate tests where the packet is dropped. This PR also shows the benefits of the new API. We can now write targeted gtests to exercise a feature. Very useful! 

I added tests for both `--dropped-packet-only` and `--output-packet-only`.


smolkaj (approved):
Very nice!

PR #4417, opened 2024-02-12
[P4Testgen] Add an option to selectively ignore control plane elements. 

fruffy:


backends/p4tools/modules/testgen/options.h
```diff
@@ -69,6 +69,9 @@ class TestgenOptions : public AbstractP4cToolOptions {
     /// TestGen will consider these when choosing input and output ports.
     std::vector<std::pair<int, int>> permittedPortRanges;
 
+    /// Skip generating a control plane entry for the entities in this list.
+    std::set<cstring> skippedControlPlaneEntities;
```
jonathan-dilorenzo, line 73: Can you say a bit about why you chose this type rather than just a set of normal strings e.g.?

backends/p4tools/modules/testgen/core/small_step/table_stepper.cpp
```diff
@@ -578,6 +576,15 @@ TableStepper::TableStepper(ExprStepper *stepper, const IR::P4Table *table)
         const auto *action = table->getActionList()->actionList.at(index);
         properties.actionIdMap.emplace(action->controlPlaneName(), index);
     }
+
+    // Set the appropriate properties when the table is immutable, meaning it has constant entries.
+    TableUtils::checkTableImmutability(*table, properties);
+
+    // If the table is in the set of entities to skip, we set it immutable.
+    // P4Testgen will not add a control plane entry for this table.
+    if (TestgenOptions::get().skippedControlPlaneEntities.count(properties.tableName) != 0U) {
```
jonathan-dilorenzo, line 585: Do standard sets not have a contains?! It looks like it does, but maybe p4c doesn't support C++20 yet? 

If so, would still much prefer find... This is weird.

backends/p4tools/modules/testgen/core/small_step/table_stepper.cpp
```diff
@@ -578,6 +576,15 @@ TableStepper::TableStepper(ExprStepper *stepper, const IR::P4Table *table)
...
+        properties.tableIsImmutable = true;
```
jonathan-dilorenzo, line 586: checkTableImmutability is setting this property based on what's passed in instead? I guess not technically part of this PR, but not a great way to go... Can you just make it 'IsTableImmutable' and then set the property in the same way you do here?

backends/p4tools/modules/testgen/options.cpp
```diff
@@ -154,6 +153,22 @@ TestgenOptions::TestgenOptions()
...
+            // Convert the input into a StringStream and split by comma (',').
+            std::stringstream argStream(arg);
+            while (argStream.good()) {
+                std::string substr;
+                std::getline(argStream, substr, ',');
+                skippedControlPlaneEntities.emplace(substr);
+            }
+            return true;
+        },
+        "Specify the set of control plane entities where P4Testgen should not generate a "
```
jonathan-dilorenzo, line 168: nit: technically "where" -> "for which" and you would remove the 'for' at the end of the sentence.

backends/p4tools/modules/testgen/targets/bmv2/test/testgen_api/control_plane_filter_test.cpp
```diff
@@ -0,0 +1,124 @@
...
+#include "backends/p4tools/modules/testgen/targets/bmv2/test_backend/protobuf_ir.h"
+#include "backends/p4tools/modules/testgen/testgen.h"
+
+namespace Test {
+
+namespace {
+
+TEST(P4TestgenControlPlaneFilterTest, FiltersControlPlaneEntities) {
+    std::stringstream streamTest;
+    streamTest << R"p4(
```
jonathan-dilorenzo, line 16: Maybe you make these manually every time... But would it be worth just having a function that returns something like this for all of your tests?

Regardless, you're probably better off putting this in a function here and then calling it in multiple tests. It would make it more natural to test e.g. what happens if you're given a non-existent control plane entity. Or one that isn't supported yet?

It looks like it would just be ignored effectively. Is that what you want to have happen? Do you want to do greater validation and give an Unimplemented error or an InvalidArgument error depending on the situation?

backends/p4tools/modules/testgen/options.h
fruffy: [controlPlaneName()](https://github.com/p4lang/p4c/blob/main/ir/declaration.h#L47) returns a `cstring` so it does seem like a more natural fit. `cstring` is fairly pervasive in the compiler, I am torn whether we should eventually get rid of it.

backends/p4tools/modules/testgen/core/small_step/table_stepper.cpp
fruffy: No, this is a C++20 feature. It is common to use `count` for this purpose because it is a bit more readable than find, so purely a force of habit on my side. I have never quite figured out whether it is slower or not, but both ostensibly have logarithmic complexity. Might be better to use find for consistency.

backends/p4tools/modules/testgen/core/small_step/table_stepper.cpp
fruffy: The current implementation for `checkTableImmutability` is this one: 
```cpp
void checkTableImmutability(const IR::P4Table &table, TableProperties &properties) {
    bool isConstant = false;
    const auto *entriesAnnotation = table.properties->getProperty("entries");
    if (entriesAnnotation != nullptr) {
        isConstant = entriesAnnotation->isConstant;
    }
    // Also check if the table is invisible to the control plane.
    // This also implies that it cannot be modified.
    properties.tableIsImmutable = isConstant || table.getAnnotation("hidden") != nullptr;
    const auto *defaultAction = table.properties->getProperty("default_action");
    CHECK_NULL(defaultAction);
    properties.defaultIsImmutable = defaultAction->isConstant;
}
```
It also sets `defaultIsImmutable`. The main reason the function was constructed this way is because other extensions may pass in an extended `TableProperties` class. For example, `TofinoTableProperties`.  

The implementation can definitely be improved, I can try to refactor this in a separate PR.

backends/p4tools/modules/testgen/targets/bmv2/test/testgen_api/control_plane_filter_test.cpp
fruffy: >Maybe you make these manually every time... But would it be worth just having a function that returns something like this for all of your tests?

Yes! I thought about this, but I figured I should wait until have more clarity on the test patterns, e.g., what elements and headers I keep reusing. 

>It looks like it would just be ignored effectively. Is that what you want to have happen? 

So this was intentional. Mostly because P4Testgen currently does not have a collection of valid control-plane elements. I first would need to create that collection in a target-independent manner (since BMv2 has different control-plane elements to Tofino or the DPDK-target), then perform the lookup. Presumably I could do this with a P4Info, but again not every extension uses that either.

It would be nice to have that validation, but it will require some work.

backends/p4tools/modules/testgen/options.h
jonathan-dilorenzo: SGTM

backends/p4tools/modules/testgen/core/small_step/table_stepper.cpp
jonathan-dilorenzo: That would be my take, but I'll allow whichever you prefer.

backends/p4tools/modules/testgen/core/small_step/table_stepper.cpp
jonathan-dilorenzo: Ah, I see. Ok. I would defer to you on how clear the implementation is versus the breadth that it needs to support. Looks pretty fine to me tbh. It just threw me off that they were setting the same value and looked totally different.

backends/p4tools/modules/testgen/targets/bmv2/test/testgen_api/control_plane_filter_test.cpp
jonathan-dilorenzo: I see. It seems fine to skip the validation (I think that's fairly typical), but I would prefer to have some tests with those cases showing that nothing catastrophic happens at least. I think that's the only change I really would want to see.

And again, I would recommend doing it (at least locally in this file) by factoring out the P4 program creating to a function that you call in ~4 different test cases (showing that it works with 1 or multiple tables, and the behavior with a not-yet-supported and non-existent entity).

backends/p4tools/modules/testgen/targets/bmv2/test/testgen_api/control_plane_filter_test.cpp
```diff
@@ -0,0 +1,264 @@
...
+
+    // First, we ensure that tests are generated correctly. We expect two tests.
+    // One which exercises action acl_drop and one which exercises the default action, NoAction.
+    auto testListOpt =
+        P4Tools::P4Testgen::Testgen::generateTests(source, compilerOptions, testgenOptions);
+
+    ASSERT_TRUE(testListOpt.has_value());
+    auto testList = testListOpt.value();
+    ASSERT_EQ(testList.size(), 2);
+    const auto *testWithDrop = testList[0]->checkedTo<P4Tools::P4Testgen::Bmv2::ProtobufIrTest>();
```
jonathan-dilorenzo, line 111: Should it be clear why you always expect the drop one to be first? Seems like an implementation detail? Probably better to loop through them and check that at least one is a drop and at least one exercises NoAction?

backends/p4tools/modules/testgen/targets/bmv2/test/testgen_api/control_plane_filter_test.cpp
```diff
@@ -0,0 +1,264 @@
...
+    EXPECT_THAT(testWithDrop->getFormattedTest(),
+                testing::Not(testing::HasSubstr(R"(expected_packet)")));
```
jonathan-dilorenzo, line 113: nit: Ideally, use `using` to make this `Not(HasSubstr(...`. We try to make it sorta english readable.

jonathan-dilorenzo (approved):

backends/p4tools/modules/testgen/targets/bmv2/test/testgen_api/control_plane_filter_test.cpp
```diff
@@ -0,0 +1,264 @@
...
+            drop_table.apply();
+        }
+    })");
+    auto compilerOptions = generateDefaultApiTestCompilerOptions();
+    auto &testgenOptions = generateDefaultApiTestTestgenOptions();
+
+    // First, we ensure that tests are generated correctly. We expect two tests.
+    // One which exercises action acl_drop and one which exercises the default action, NoAction.
+    auto testListOpt =
+        P4Tools::P4Testgen::Testgen::generateTests(source, compilerOptions, testgenOptions);
```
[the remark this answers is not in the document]
fruffy: Currently no, there is no interface for that because the P4 Program is parsed and then traverses the front and mid end. It would be nice to just pass in a `IR::P4Program` class. Is that what you are referring to?

backends/p4tools/modules/testgen/targets/bmv2/test/testgen_api/control_plane_filter_test.cpp
```diff
@@ -0,0 +1,264 @@
...
+    testgenOptions.maxTests = 0;
+    // Create a bespoke packet for the Ethernet extract call.
+    constexpr int kEthHdrBitSize = 112;
+    testgenOptions.minPktSize = kEthHdrBitSize;
+    testgenOptions.maxPktSize = kEthHdrBitSize;
+    return testgenOptions;
+}
+
+TEST(P4TestgenControlPlaneFilterTest, GeneratesCorrectTests) {
+    auto source = generateTestProgram(R"(
```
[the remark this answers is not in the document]
fruffy: Definitely a recovering auto user here, hard to drop that habit :).

backends/p4tools/modules/testgen/targets/bmv2/test/testgen_api/control_plane_filter_test.cpp
fruffy: Tried to implement something here. Any way to use a lambda in Protobuf matchers? Currently I need to convert the entire vector to strings before passing them to the matchers.

PR #4419, opened 2024-02-13
Error casting bool const to other than bit<1>

ChrisDodd:
Trivial corner case that should not be allowed according to the spec

### DIFF
--- a/frontends/common/constantFolding.cpp
+++ b/frontends/common/constantFolding.cpp
@@ -798,6 +798,9 @@ const IR::Node *DoConstantFolding::postorder(IR::Cast *e) {
         if (auto arg = expr->to<IR::Constant>()) {
             return cast(arg, arg->base, type);
         } else if (auto arg = expr->to<IR::BoolLiteral>()) {
+            if (type->isSigned || type->size != 1)
+                error(ErrorType::ERR_INVALID, "%1%: Cannot cast %1% directly to %2% (use bit<1>)",
+                      arg, type);
             int v = arg->value ? 1 : 0;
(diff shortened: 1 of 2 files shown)


vlstill (approved):

vlstill:
I am quite surprised this is not checked in type checking though. That would seem like a better place to do it (especially if we want to boot out the constant folding out of frontend in future!). The type checker checks it for variable access, why not for constant?

ChrisDodd:
> I am quite surprised this is not checked in type checking though. That would seem like a better place to do it (especially if we want to boot out the constant folding out of frontend in future!). The type checker checks it for variable access, why not for constant?

The basic issue is that we do constant folding before type checking so we can deal with `bit<N>` where `N` is an arbitrary constant expression.  It might be better to rearrange things to only do constant folding (and copy propagation, and inlining) into type args (and things that end up in type args) prior to type checking, but that would be trickier.

frontends/common/constantFolding.cpp
```diff
@@ -798,6 +798,9 @@ const IR::Node *DoConstantFolding::postorder(IR::Cast *e) {
         if (auto arg = expr->to<IR::Constant>()) {
             return cast(arg, arg->base, type);
         } else if (auto arg = expr->to<IR::BoolLiteral>()) {
+            if (type->isSigned || type->size != 1)
+                error(ErrorType::ERR_INVALID, "%1%: Cannot cast %1% driectly to %2% (use bit<1>)",
```
fruffy, line 802:
```suggestion
                error(ErrorType::ERR_INVALID, "%1%: Cannot cast %1% directly to %2% (use bit<1>)",
```

PR #4421, opened 2024-02-13
test for ContantFolding + policy

grg:
Add a test for ConstantFolding and ConstantFoldingPolicy.

Ported test from abandoned #4402.

### DIFF
--- a/test/gtest/constant_folding.cpp
+++ b/test/gtest/constant_folding.cpp
@@ -1,8 +1,9 @@
+#include <gtest/gtest.h>
+
 #include "frontends/common/constantFolding.h"
 #include "frontends/common/parseInput.h"
 #include "frontends/common/resolveReferences/referenceMap.h"
 #include "frontends/common/resolveReferences/resolveReferences.h"
-#include "gtest/gtest.h"
 #include "helpers.h"
 #include "ir/ir.h"
 #include "ir/pass_manager.h"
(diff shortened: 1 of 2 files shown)


fruffy (approved):

test/gtest/constant_folding.cpp
```diff
@@ -0,0 +1,135 @@
+#include "frontends/common/constantFolding.h"
+#include "frontends/common/parseInput.h"
+#include "frontends/common/resolveReferences/referenceMap.h"
+#include "frontends/common/resolveReferences/resolveReferences.h"
+#include "gtest/gtest.h"
```
fruffy, line 5:
```suggestion
#include <gtest/gtest.h>
```

+ clang-format on it.

grg: clang-format doesn't change this (at least not on my test systems).

If angled brackets are preferred, then we should change _all_ of the files in that directory to use them. Currently only `helpers.h` and `p4runtime.cpp` use angled-brackets. Let me know if this is preferred and I'll create a separate PR.

fruffy: I think it is a good idea to make all gtest includes angled, since it is a third-party dependency not in our control. It matters for some analysis frameworks. 

We can leave it as is here in this PR.

grg: Created a new PR for the other files: #4422

PR #4422, opened 2024-02-14
gtest include: quotes -> angled brackets

grg:
Convert quotes to angle brackets in `#include "gtest/gtest.h"`. Suggested by @fruffy in #4421.


fruffy (approved):
Thanks!

PR #4424, opened 2024-02-14
Provide better hash functions and boilerplate

### LINKED ISSUE #4392: Update the hash functions used in `hash.cpp` and `cstring.cpp`
`lib/hash.cpp` has been contributed several years back to support string hashes in the compiler's `cstring` implementation. However, the hash function has flaws. We should consider replacing it with more robust hashes such as `xxhash3` if we decide to making hash functions part of the public compiler library. 

Context from https://github.com/p4lang/p4c/pull/4377#discussion_r1477781295

>  `lib/hash.cpp` implements fnv1a and (old) murmur2. murmur2 is used in `cstring` (for unification, as string hash) and fnv1 is used to hash cstring' contents insider parser unroll. Both cases do not make much sense to me:
> 
> 1. murmur2 is old, slow and known to have flaws
> 2. fnv1a is quite slow, has mediocre quality. And actually its usage in parser unroll is questionable: since cstrings are guaranteed unique, then the address of string is enough as 64-bit hash
> 
> In my opinion:
>  - murmur2 should be replaced with something a bit more modern and performant. I'd stick with xxhash3.
>  - fnv1a usage from parser unroll should be removed
> 
> The only reason why fnv1a was chosen here is that it's small enough to be evaluated in constexpr context without slowing down compilation time. And for this its quality is good enough here: note that we are hashing few dozen of strings that are known during compile time, so our collision requirements are much lower than generic purpose string hash.
> 
> PS: Certainly, hash refactoring is a separate task :)
> PPS: Frankly speaking, just the address of string buffer is quite poor `hash<cstring>` for hash map implementation collision-wise. However, I believe we can easily make it much stronger just passing it through xxh3.

asl:
This PR provides a bit more comprehensive implementation of hashes that could be used within compiler. In particular:
 - FNV1A and Murmur2 are removed
 - 64-bit xxHash3 was imported with few simplifications from xxhash_clean project and without workarounds for old compilers
 - Particular hash algorithm is considered implementation detail, one should not bother about this. Few low-level functions are provided: hash itself, hash avalanche and hash combining
 - On top of that some useful hash boilerplate was added. Since C++ standard committee is still bikeshedding the particular better hashing interface implementation (see https://gist.github.com/dietmarkuehl/b0767e76bc85e29e74f61e994f105ad9?permalink_comment_id=3923250 for some summary), we do not have anything that is even currently under consideration. The present implementation is inspired by N3333, P0029 and P0814 and provides std-compatible implementations of decent quality (on top of low-level functions) that 'just work' with lots of useful sugar on top of that.

This improves runtime of P4CParserUnroll.switch_20160512 by ~5% since we combine hashes directly without filling and hashing an `std::vector` each time inside `StackVariableHash`.

Fixes #4392

### DIFF
--- a/lib/hash.cpp
+++ b/lib/hash.cpp
@@ -6,6 +6,8 @@
 namespace Util {
 
 namespace {
+using namespace Detail;
+
 constexpr size_t XXH3_SECRETSIZE_MIN = 136;
 constexpr size_t XXH_SECRET_DEFAULT_SIZE = 192;
 
@@ -26,9 +28,6 @@ constexpr uint8_t kSecret[XXH_SECRET_DEFAULT_SIZE] = {
     0x45, 0xcb, 0x3a, 0x8f, 0x95, 0x16, 0x04, 0x28, 0xaf, 0xd7, 0xfb, 0xca, 0xbb, 0x4b, 0x40, 0x7e,
 };
 // clang-format on
-
-constexpr uint64_t PRIME_MX1 = UINT64_C(0x165667919E3779F9);
-constexpr uint64_t PRIME_MX2 = UINT64_C(0x9FB21C651E98DF25);
 
 #if defined(_WIN32) /* Windows is always little endian */ \
     || defined(__LITTLE_ENDIAN__) ||                      \
@@ -126,40 +125,15 @@ constexpr size_t XXH_STRIPE_LEN = 64;
 constexpr size_t XXH_SECRET_CONSUME_RATE = 8;
 constexpr size_t XXH_ACC_NB = XXH_STRIPE_LEN / sizeof(uint64_t);
 
-static uint64_t rotl64(uint64_t X, size_t R) { return (X << R) | (X >> (64 - R)); }
-
 constexpr uint32_t PRIME32_1 = UINT32_C(0x9E3779B1);
 constexpr uint32_t PRIME32_2 = UINT32_C(0x85EBCA77);
 constexpr uint32_t PRIME32_3 = UINT32_C(0xC2B2AE3D);
-
-static const uint64_t PRIME64_1 = UINT64_C(11400714785074694791);
-static const uint64_t PRIME64_2 = UINT64_C(14029467366897019727);
-static const uint64_t PRIME64_3 = UINT64_C(1609587929392839161);
-static const uint64_t PRIME64_4 = UINT64_C(9650029242287828579);
-static const uint64_t PRIME64_5 = UINT64_C(2870177450012600261);
-
-static uint64_t XXH64_avalanche(uint64_t hash) {
-    hash ^= hash >> 33;
-    hash *= PRIME64_2;
-    hash ^= hash >> 29;
-    hash *= PRIME64_3;
-    hash ^= hash >> 32;
-    return hash;
-}
 
 static uint64_t XXH3_avalanche(uint64_t hash) {
     hash ^= hash >> 37;
     hash *= PRIME_MX1;
     hash ^= hash >> 32;
     return hash;
-}
-
-static uint64_t XXH3_rrmxmx(uint64_t acc, uint64_t len) {
-    acc ^= rotl64(acc, 49) ^ rotl64(acc, 24);
-    acc *= PRIME_MX2;
-    acc ^= (acc >> 35) + len;
-    acc *= PRIME_MX2;
-    return acc ^ (acc >> 28);
 }
 
 static uint64_t XXH3_len_1to3_64b(const uint8_t *input, size_t len, const uint8_t *secret,
--- a/lib/hash.cpp
+++ b/lib/hash.cpp
@@ -36,6 +36,7 @@
 
 #include <cstdint>
 #include <cstring>
+#include <cassert>
 
 namespace Util {
 
--- a/lib/hash.cpp
+++ b/lib/hash.cpp
@@ -4,105 +4,338 @@
 #include <cstring>
 
 namespace Util {
-namespace Hash {
 
-namespace Detail {
-std::uint32_t murmur32(const void *data, std::uint32_t size) {
-    const std::uint32_t m = 0x5bd1e995;
-    const std::uint64_t seed = 0xc70f6907UL;
-    std::uint32_t result = seed ^ size;
-    const char *raw = static_cast<const char *>(data);
+namespace {
+constexpr size_t XXH3_SECRETSIZE_MIN = 136;
+constexpr size_t XXH_SECRET_DEFAULT_SIZE = 192;
 
-    while (size >= sizeof(std::uint32_t)) {
-        std::uint32_t k;
-        std::memcpy(&k, raw, sizeof(k));
-        k *= m;
-        k ^= k >> 24;
-        k *= m;
-        result *= m;
-        result ^= k;
-        raw += 4;
-        size -= 4;
-    }
+/* Pseudorandom data taken directly from FARSH */
+// clang-format off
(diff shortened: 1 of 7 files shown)


lib/hash.h
```diff
@@ -2,42 +2,282 @@
...
+};
+template <typename... Types>
+struct Hasher<std::tuple<Types...>> {
+    size_t operator()(const std::tuple<Types...> &val) const { return apply(Hash(), val); }
+};
+
+// In general, pointers are bad hashes: their low 2-3 bits are zero, likewise
+// for the upper bits depending on the ABI. Also, the middle bits might not have
+// enough entropy as addresses come from some common pool. To solve this problem
+// we just use a single iteration of hash_avalanche to improve mixing.
```
vlstill, line 200: Please turn this into Doxygen comment

lib/hash.h
```diff
@@ -2,42 +2,282 @@
...
+template <typename... Types>
+struct hash<std::tuple<Types...>> {
+ public:
+    size_t operator()(std::tuple<Types...> const &key) const {
+        Util::Detail::TupleHasher<sizeof...(Types) - 1, Types...> hasher;
+
+        return hasher(key);
+    }
+};
+}  // namespace std
```
vlstill, line 281: I'm a bit worried about this. Isn't it forbidden by the standard to specialize standard classes for other standard (and builtin) types (i.e. it is all fine with specializing `std::hash<>` for our `cstring`, but we should not specialize it for anything that the std defines)?

I would suggest we are just explicit and use the `Util::Hash` as hasher when we need to hash tuples. Possible problem is that in that case we could be that we would not be able to hash nested things that have `std::hash` but not `Util::Hash`... maybe `Util::Hash` can fallback to `std::hash`? The template specialization rules _might_ be actually doing the right think if you define the non-specialized struct to defer to `std::hash`.

vlstill (commented):
Thank you for this. Could you please add doxygen comment on Hash and Hasher (the declaration) so that people know which one to use in their collections and what is Hasher for?

lib/hash.h
asl: > I'm a bit worried about this. Isn't it forbidden by the standard to specialize standard classes for other standard (and builtin) types (i.e. it is all fine with specializing std::hash<> for our cstring, but we should not specialize it for anything that the std defines)?

That's exactly one of the points of aforementioned proposals. It was proposed in N3983 that the standard library would provide such specializations, but ... nothing happened. One of the main reasons is that there is no standard way to combine hashes and therefore everything stalled and became grey zone. I can certainly remove those specializations, it's not a big deal. Overall, specializations in `std` namespace are considered as mistakes in standard as rules what is permitted and what is not are confusing at least, but it's too late to fix it :) Maybe in C++42.

I'd rather not fallback `Util::Hash` to `std::hash`. Otherwise we might lose possibility to define latter in terms of former due to cyclic dependencies.

asl: Note that removing specializations for pairs and tuples will make impossible `hash_combine` as well:
```c++
    auto a = std::make_pair(1, 2);
    auto b = std::make_pair(3, 4);
    auto c = std::make_pair(1, 2);
    auto d = std::make_pair(2, 1);
    EXPECT_EQ(Util::hash_combine(a), Util::hash_combine(c));
    EXPECT_NE(Util::hash_combine(b), Util::hash_combine(c));
    EXPECT_NE(Util::hash_combine(d), Util::hash_combine(c));

    EXPECT_EQ(Util::hash_combine(a, b), Util::hash_combine(c, b));
    EXPECT_NE(Util::hash_combine(a, b), Util::hash_combine(b, a));
```
will stop working. One would need either use `hash_combine_generic` or disassemble pair / tuple.

lib/hash.cpp
```diff
@@ -1,139 +1,346 @@
...
+    0x45, 0xcb, 0x3a, 0x8f, 0x95, 0x16, 0x04, 0x28, 0xaf, 0xd7, 0xfb, 0xca, 0xbb, 0x4b, 0x40, 0x7e,
 };
+// clang-format on
 
-template <>
-struct fnv1a_traits<sizeof(std::uint64_t)> {
-    static const std::size_t offset_basis = static_cast<std::size_t>(14695981039346656037ULL);
-    static const std::size_t prime = static_cast<std::size_t>(1099511628211ULL);
-};
+#if defined(_WIN32) /* Windows is always little endian */ \
```
fruffy, line 67: One day we may support windows :).

lib/hash.cpp
```diff
@@ -1,139 +1,346 @@
+/*
+ *  xxHash - Fast Hash algorithm
+ *  Copyright (C) 2012-2021, Yann Collet
+ *
+ *  BSD 2-Clause License (http://www.opensource.org/licenses/bsd-license.php)
```
fruffy, line 5: BSD 2 is compatible with Apache, so this should not cause license problems.

test/gtest/hash.cpp
```diff
@@ -0,0 +1,292 @@
...
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+#include "lib/hash.h"
+
+#include <unordered_map>
+#include <unordered_set>
+
+#include "gtest/gtest.h"
```
fruffy, line 20: Can we use `#include <gtest/gtest.h>` since this is an external third-party dependency? This can be helpful for some tools like `clang-tidy` and `bazel`. I believe Bazel even throws an error if a system include is in quoted brackets. It just currently does not build the test folder.

lib/hash.h
```diff
@@ -2,42 +2,301 @@
...
+};
+
+/// In general, pointers are bad hashes: their low 2-3 bits are zero, likewise
+/// for the upper bits depending on the ABI. Also, the middle bits might not
+/// have enough entropy as addresses come from some common pool. To solve this
+/// problem we just use a single iteration of hash_avalanche to improve mixing
+/// (see Detail::IntegerHasher).
+template <typename T>
+struct Hasher<T *> {
+    // FIXME: better use std::bit_cast from C++20
```
fruffy, line 215: I am thinking we should create an umbrella issue that tracks this as part of a todo list. We are not far from C++20.

fruffy (approved):
I trust you on the hash implementation, I do not know much about it. 

Can we also replace the hash function in `rtti.h` with this? Or move it here? Unless you want to keep it private.

asl:
> Can we also replace the hash function in `rtti.h` with this? Or move it here? Unless you want to keep it private.

No. It is not constexpr :) And I do not want to make it constexpr for the sake of compile time

lib/hash.cpp
asl: Yes. I tried to keep windows things if they are isolated and simple enough.

lib/hash.cpp
asl: xxhash is also used in LLVM. LLVM is under Apache2 (+ LLVM exception), so there is a precedent just in case.

test/gtest/hash.cpp
asl: Sure, done!

asl:
@vlstill Please let me know if there are any outstanding comments, or if this ready to be merged

Thanks!

lib/hash.h
vlstill: OK, I guess we can remove them when the standard finally defines them. The ambiguity should be quite clear at that point so hopefully it will not be a big mess.

vlstill (approved):

lib/hash.h
asl: Yes, we can just get rid of them, it's easy :)

Issue #4412, opened 2024-02-12
Replace Util::Enumerator with a C++-native enumerator in resolveReferences


fruffy:
Context: https://github.com/p4lang/p4c/pull/4376#discussion_r1485685582

vlstill:
I think that ideally, we would replace a lot of these with C++ ranges (although they are not a direct replacement in many cases I guess -- partially because Util::Enumerator does type erasure). But that would depend raising compiler requirement for building p4c. I'm not sure who are the bounding users (and if we even know). Ubuntu 20.04 LTS has 10.5, which *should* have working ranges (I haven't tested/don't know if there are any bugs). 18.04LTS has 7.4, far older then the required 9. So I am not sure where is the GCC 9 requirement coming from. Might be RHEL 8, which has GCC 9, but maybe if that is the case we could ask users to use https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/8/html/developing_c_and_cpp_applications_in_rhel_8/additional-toolsets-for-development_developing-applications#gcc-toolset_assembly_additional-toolsets-for-development there (yes I know RHEL 8 is not officially supported by p4c but presumably it would be quite common in enterprise).

The other important question is if the C++20 support is stable enough on the given compiler.

An alternative would be to use a 3rd party library that is similar to C++ ranges and support GCC 9, but I am not sure if there is a good candidate.

asl:
> (although they are not a direct replacement in many cases I guess -- partially because Util::Enumerator does type erasure).

Yes. I think the modelling example here is `getDeclarations`.  The bad thing about `Enumerator` itself is that various adapters (`where` / `concat`) does lots of allocations for no particular reason. This would be replaced with views ideally (though I remember there were some performance issues with `concat` in ranges-v3).

For type erasure we might consider replacing with value semantics (Model / Concept-based polymorphism), though It's likely will be similar wrt allocations and overheads...

> The other important question is if the C++20 support is stable enough on the given compiler.

Well... that's another story, yes. There are lots of small issues and corner cases. E.g. in order to build LLVM with C++20 enabled one needs to have clang 17.0.6.

PR #4432, opened 2024-02-16
Refactor ReferenceResolver to use native C++-enumerators in some places

asl:
Refactor ReferenceResolver to get rid of majority of unnecessary heap allocations due to obsessive `Enumerator<>` usage:
  - Use iterators and iterator ranges where possible
  - Ensure we do not allocate on heap temporary objects and exploit move semantics as much as possible returning vectors by value
  - Do some refactoring around `Enumerator<>->toVector()`: make it return result by value. It seems that the use in the tree is always about creating some temporaries. And it does not make any sense to allocate vectors on heap then
  - Few small improvements here and there, e.g. removing dependence from boost adapters here
 
Partially addresses https://github.com/p4lang/p4c/issues/4412

Ideally, we'd switch from memory allocation at `getDeclarations` / `getDeclsByName`, but it's a bit long way to go for now.

### DIFF
--- a/backends/p4tools/common/lib/namespace_context.cpp
+++ b/backends/p4tools/common/lib/namespace_context.cpp
@@ -37,13 +37,8 @@ const IR::IDeclaration *NamespaceContext::findNestedDecl(
         // Handle case where current namespace is an IGeneralNamespace.
         // If there is no match, we fall through.
         if (const auto *ns = subNamespace->to<IR::IGeneralNamespace>()) {
-            auto decls = ns->getDeclsByName(name)->toVector();
-            if (!decls.empty()) {
-                // TODO: Figure out what to do with multiple results. Maybe return all of them and
-                // let the caller sort it out?
-                BUG_CHECK(decls.size() == 1, "Handling of overloaded names not implemented");
-                return decls.front();
-            }
+            auto *decl = ns->getDeclsByName(name)->singleOrDefault();
+            if (decl != nullptr) return decl;
         }
 
         // As last resort, fall through to a NestedNamespace check.
@@ -82,13 +77,8 @@ const IR::IDeclaration *NamespaceContext::findDecl(const IR::Path *path) const {
     // Handle case where current namespace is an IGeneralNamespace.
     // If there is no match, we fall through.
     if (const auto *ns = curNamespace->to<IR::IGeneralNamespace>()) {
-        auto decls = ns->getDeclsByName(name)->toVector();
-        if (!decls.empty()) {
-            // TODO: Figure out what to do with multiple results. Maybe return all of them and let
-            // the caller sort it out?
-            BUG_CHECK(decls.size() == 1, "Handling of overloaded names not implemented");
-            return decls.front();
-        }
+        auto *decl = ns->getDeclsByName(name)->singleOrDefault();
+        if (decl != nullptr) return decl;
     }
     // As last resort, check if the NestedNamespace contains the declaration.
     if (const auto *ns = curNamespace->to<IR::INestedNamespace>()) {
--- a/backends/p4tools/common/lib/namespace_context.cpp
+++ b/backends/p4tools/common/lib/namespace_context.cpp
@@ -37,12 +37,12 @@ const IR::IDeclaration *NamespaceContext::findNestedDecl(
         // Handle case where current namespace is an IGeneralNamespace.
         // If there is no match, we fall through.
         if (const auto *ns = subNamespace->to<IR::IGeneralNamespace>()) {
-            const auto *decls = ns->getDeclsByName(name)->toVector();
-            if (!decls->empty()) {
+            auto decls = ns->getDeclsByName(name)->toVector();
+            if (!decls.empty()) {
                 // TODO: Figure out what to do with multiple results. Maybe return all of them and
                 // let the caller sort it out?
-                BUG_CHECK(decls->size() == 1, "Handling of overloaded names not implemented");
-                return decls->at(0);
+                BUG_CHECK(decls.size() == 1, "Handling of overloaded names not implemented");
+                return decls.front();
             }
         }
 
@@ -82,12 +82,12 @@ const IR::IDeclaration *NamespaceContext::findDecl(const IR::Path *path) const {
     // Handle case where current namespace is an IGeneralNamespace.
     // If there is no match, we fall through.
     if (const auto *ns = curNamespace->to<IR::IGeneralNamespace>()) {
-        const auto *decls = ns->getDeclsByName(name)->toVector();
-        if (!decls->empty()) {
+        auto decls = ns->getDeclsByName(name)->toVector();
+        if (!decls.empty()) {
             // TODO: Figure out what to do with multiple results. Maybe return all of them and let
             // the caller sort it out?
-            BUG_CHECK(decls->size() == 1, "Handling of overloaded names not implemented");
-            return decls->at(0);
+            BUG_CHECK(decls.size() == 1, "Handling of overloaded names not implemented");
+            return decls.front();
         }
     }
     // As last resort, check if the NestedNamespace contains the declaration.
--- a/frontends/common/resolveReferences/resolveReferences.cpp
+++ b/frontends/common/resolveReferences/resolveReferences.cpp
@@ -314,7 +314,8 @@ const IR::IDeclaration *ResolveReferences::resolvePath(const IR::Path *path, boo
 
 void ResolveReferences::checkShadowing(const IR::INamespace *ns) const {
     if (!checkShadow) return;
-    std::map<cstring, const IR::Node *> prev_in_scope;  // check for shadowing within a scope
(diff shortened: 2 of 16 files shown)


asl:
Maybe it would make sense to wait until C++20 adoption and replace `Enumerator` with ranges and views over them.

frontends/common/resolveReferences/resolveReferences.cpp
```diff
@@ -178,35 +161,32 @@ const std::vector<const IR::IDeclaration *> *ResolutionContext::lookup(
...
                   current->node_type_name());
     }
-    if (auto nested = current->to<IR::INestedNamespace>()) {
-        // boost::adaptors::reverse expects a reference
+    if (const auto *nested = current->to<IR::INestedNamespace>()) {
         auto temp = nested->getNestedNamespaces();
-        for (auto nn : boost::adaptors::reverse(temp)) {
-            auto rv = lookup(nn, name, type);
-            if (!rv->empty()) return rv;
+        for (auto it = temp.rbegin(); it != temp.rend(); ++it) {
```
vlstill, line 172: We might add reverse adaptor to the `iterator_range`.

vlstill (commented):
Generally I think this is a good change, thank you for your contribution. You may consider special-casing the common task of taking the first element of an enumerator in the enumerator itself. Do you have any performance measurements? I wonder what is the visible impact.

frontends/common/resolveReferences/resolveReferences.h
```diff
@@ -48,14 +50,13 @@ class ResolutionContext : virtual public Visitor, public DeclarationLookup {
...
-    const std::vector<const IR::IDeclaration *> *lookup(const IR::INamespace *ns, IR::ID name,
-                                                        ResolutionType type) const;
+    std::vector<const IR::IDeclaration *> lookup(const IR::INamespace *ns, const IR::ID &name,
+                                                 ResolutionType type) const;
 
     // match kinds exist in their own special namespace, made from all the match_kind
     // declarations in the global scope.  Unlike errors, we don't merge those scopes in
     // the parser, so we have to find them and scan them here.
-    const std::vector<const IR::IDeclaration *> *lookupMatchKind(IR::ID name) const;
+    std::vector<const IR::IDeclaration *> lookupMatchKind(const IR::ID &name) const;
```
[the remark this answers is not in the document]
asl: We are having:
```
struct ID : Util::IHasSourceInfo {
    Util::SourceInfo srcInfo;
    cstring name = nullptr;
    cstring originalName = nullptr;
}
```

And `Util::SourceInfo` is:
```
class SourceInfo final {
 public:
    cstring filename = "";
    int line = -1;
    int column = -1;
    cstring srcBrief = "";
```

So, definitely should be passed by reference.

frontends/common/resolveReferences/resolveReferences.cpp
```diff
@@ -16,74 +16,57 @@ limitations under the License.
...
-    auto decls = ns->getDeclarations();
+    std::vector<const IR::IDeclaration *> decls;
     if (auto nest = ns->to<IR::INestedNamespace>()) {
-        // boost::adaptors::reverse expects a reference
-        auto temp = nest->getNestedNamespaces();
-        for (auto nn : boost::adaptors::reverse(temp)) decls = nn->getDeclarations()->concat(decls);
+        for (const auto *nn : nest->getNestedNamespaces()) {
+            auto nnDecls = nn->getDeclarations();
+            // FIXME: should just insert()
+            for (const auto *decl : *nnDecls) decls.push_back(decl);
```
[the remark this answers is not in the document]
asl: Because the way how `Enumerator` is implemented.... `Enumerator::begin() / Enumerator::end()` return so-called `EnumeratorHandle` which does not satisfy iterator requirements (no trait types like `reference` / `value_type`, etc.)

backends/p4tools/common/lib/namespace_context.cpp
```diff
@@ -82,12 +82,12 @@ const IR::IDeclaration *NamespaceContext::findDecl(const IR::Path *path) const {
     // Handle case where current namespace is an IGeneralNamespace.
     // If there is no match, we fall through.
     if (const auto *ns = curNamespace->to<IR::IGeneralNamespace>()) {
-        const auto *decls = ns->getDeclsByName(name)->toVector();
-        if (!decls->empty()) {
+        auto decls = ns->getDeclsByName(name)->toVector();
+        if (!decls.empty()) {
```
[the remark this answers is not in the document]
asl: Yes, I thought about this, but somehow decided to not to add... Will check it out.

asl:
> I wonder what is the visible impact.

We do not have much visible impact after memoization patch :) If we'd get rid of memoization, then there will be quite visible impact for large apps.

lib/iterator_range.h
```diff
@@ -0,0 +1,57 @@
...
+
+template <class T>
+iterator_range<T> make_range(T x, T y) {
+    return iterator_range<T>(std::move(x), std::move(y));
+}
+
+template <typename T>
+iterator_range<T> make_range(std::pair<T, T> p) {
+    return iterator_range<T>(std::move(p.first), std::move(p.second));
+}
```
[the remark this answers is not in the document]
asl: Likely not. I added deduction guide after I had those :)

lib/iterator_range.h
```diff
@@ -0,0 +1,57 @@
...
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+#ifndef LIB_ITERATOR_RANGE_H_
+#define LIB_ITERATOR_RANGE_H_
+
+#include <utility>
+
+namespace Util {
```
fruffy, line 20: Is this a partial substitute for the lack of C++20 ranges? Can you add a comment describing its purpose? I wonder how transient this utility library will be. 

Alternatively, we could add a drop-in substitute like https://github.com/ericniebler/range-v3 to the compiler and then replace it once we migrate to C++20.

frontends/common/resolveReferences/resolveReferences.cpp
```diff
@@ -16,74 +16,57 @@ limitations under the License.
 
 #include "resolveReferences.h"
 
-#include <sstream>
-
-#include <boost/range/adaptor/reversed.hpp>
```
fruffy, line 21: Always good to see boost dependencies being removed :).

lib/iterator_range.h
fruffy: I guess what I am getting at is that these little libraries end up hanging around and over the years people lose information about what they are for. With the vendored dependencies it is a little more explicit because you know they are placeholders until the standards progress. 

Can we add an explanation why this was added, what it is being used for and how it could be replaced, if at all?

asl: Yes, sure. And remember: "every large project should have lisp interpreter inside" :)

frontends/common/resolveReferences/resolveReferences.h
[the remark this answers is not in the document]
vlstill: I forgot there is a SourceInfo inside, I should have checked.

frontends/common/resolveReferences/resolveReferences.cpp
asl: We just replace with out own boost ;)

backends/p4tools/common/lib/namespace_context.cpp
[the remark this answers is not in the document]
asl: Actually, there is already `single()` method that does exactly this. It's just `Enumerator` interface is quite alien for C++ so noone seem to use is :)

test/gtest/parser_unroll.cpp
```diff
@@ -178,12 +178,12 @@ class MidEnd : public PassManager {
 #endif
 
 const IR::P4Parser *getParser(const IR::P4Program *program) {
```
fruffy, line 180: Ideally, this should return an error/nullpointer/optional if there is no parser in the program but that is not your responsibility to fix in this PR.

lib/iterator_range.h
```diff
@@ -0,0 +1,71 @@
...
+}
+
+template <typename Range>
+constexpr auto end_impl(Range &&range) {
+    return end(std::forward<Range>(range));
+}
+
+}  // namespace Detail
+
+// This is a lightweight alternative for C++20 std::range. Should be replaced
```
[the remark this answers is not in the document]
asl: Well, certainly these little pieces here and there should be revisit each time the standard is bumped :)

test/gtest/parser_unroll.cpp
asl: I tried to keep the original semantics of the code. Certainly there are lots of BUG_CHECK's here instead of real diagnostics. I think we'd probably need to use something like Expected idiom so the errors could be detected more easily and propagated.

asl:
I am going to merge this unless noone would object as this would allow us to get rid additional boost bits.

PR #4433, opened 2024-02-16
Pass FrontEndPolicy to unit tests

vlstill:
**Breaking change (dev/test)**: Side-effects ordering is enabled in frontend tests (by #4406).

I've accidentally **enabled side-effects ordering in all frontend tests** throughout the code base in https://github.com/p4lang/p4c/pull/4406 😢. I am sorry about this.

Arguably this could be a good thing (that is the indented use of the frotnend), but it could be surprising to downstream projects. Also, as the policy itself was not exposed, it was not possible to change this behaviour in downstream tests. This PR mitigates the second problem, it exposes frontend policy in tests so that tools can set it in any way they desire. **The side-effects ordering is still enabled by default**, which I think is a more reasonable default (it is the default used in frontend after all).

I don't see a way of ensuring downstream action on this without needlessly breaking interface of the tests that don't depend on either side effects ordering or annotation parsing. Therefore I just intend to ping people who I guess could be working on downstream projects here.

### DIFF
--- a/frontends/p4/frontend.cpp
+++ b/frontends/p4/frontend.cpp
@@ -147,7 +147,6 @@ class SetStrictStruct : public NoVisit {
 
 }  // namespace
 
-// TODO: remove skipSideEffectOrdering flag
 const IR::P4Program *FrontEnd::run(const CompilerOptions &options, const IR::P4Program *program,
                                    std::ostream *outStream) {
     if (program == nullptr && options.listFrontendPasses == 0) return nullptr;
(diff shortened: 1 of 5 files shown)


fruffy (approved):

fruffy:
I didn't see any breakage fwiw but this may affect the Tofino tests.

vlstill:
> I didn't see any breakage fwiw but this may affect the Tofino tests.

It was also really easy with the old setting to accidentally test with side effects ordering disabled which I think is a mistake as presumably most of the backends will want it enabled.

grg (approved):
LGTM

PR #4434, opened 2024-02-19
Don't apply defaultArguments in table action lists

ChrisDodd:
When actions have default args, they should not be applied in the action list of a table, as the control plane may supply non-default values for these args.


ChrisDodd:
I added a testcase that triggers the BUG_CHECK I was seeing before with the previous code, but doesn't cover all the cases that should work with default args (because they still don't work).

fruffy (approved):

Issue #4439, opened 2024-02-20
RemoveUnusedDeclarations: inconsistency in when warnings are produced


grg:
RemovedUnusedDeclarations will remove unused P4Control, P4Parser, and P4Table objects.

When it removes a P4Table object, it produces a warning with:
```
        if (giveWarning(getOriginal()))
            warn(ErrorType::WARN_UNUSED, "Table %1% is not used; removing", table);
```

But when it removes P4Control/P4Parser objects, it silently removes them without reporting a warning.

Is there a reason why warnings are not reported for P4Control/P4Parser objects? If not, I would like to add similar warnings for those two objet types.

grg:
Closing as fixed. Merged #4440. (No objections here or in 4440.)

PR #4440, opened 2024-02-20
warn on unused controls/parsers

grg:
Add warnings for unused P4Control and P4Parser objects to match P4Table warnings. Addresses issue #4439.

### DIFF
--- a/testdata/p4_16_samples_outputs/issue3379-1.p4-stderr
+++ b/testdata/p4_16_samples_outputs/issue3379-1.p4-stderr
@@ -1,4 +1,7 @@
 issue3379-1.p4(14): [--Wwarn=unused] warning: t4: unused instance
 mypackaget<bit>(MyParser1()) t4;
                              ^^
+issue3379-1.p4(3): [--Wwarn=unused] warning: Parser MyParser1 is not used; removing
+parser MyParser1(in bit tt) {
+       ^^^^^^^^^
 [--Wwarn=missing] warning: Program does not contain a `main' module
--- a/frontends/p4/unusedDeclarations.cpp
+++ b/frontends/p4/unusedDeclarations.cpp
@@ -54,6 +54,8 @@ const IR::Node *RemoveUnusedDeclarations::preorder(IR::Type_SerEnum *type) {
 const IR::Node *RemoveUnusedDeclarations::preorder(IR::P4Control *cont) {
     auto orig = getOriginal<IR::P4Control>();
     if (!refMap->isUsed(orig)) {
+        if (giveWarning(orig))
+            warn(ErrorType::WARN_UNUSED, "Control %1% is not used; removing", cont);
         LOG3("Removing " << cont << dbp(orig));
         prune();
         return nullptr;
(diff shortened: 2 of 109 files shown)


testdata/p4_16_samples_outputs/issue3379-1.p4-stderr
```diff
@@ -1,4 +1,7 @@
 issue3379-1.p4(14): [--Wwarn=unused] warning: t4: unused instance
 mypackaget<bit>(MyParser1()) t4;
                              ^^
+issue3379-1.p4(3): [--Wwarn=unused] warning: Parser MyParser1 is not used; removing
+parser MyParser1(in bit tt) {
+       ^^^^^^^^^
```
jafingerhut, line 6: @grg Do you know why Myparser1 is considered unused in this test program?

jafingerhut (approved):
I left one question about a warning message for one of the test programs, but if there is a good explanation for that one, the rest all looks good to me.

testdata/p4_16_samples_outputs/issue3379-1.p4-stderr
fruffy: I believe that is because there is no `main` declaration.

grg:
@jafingerhut / @fruffy: Any thoughts on whether these warnings should be added? I opened a ticket (#4439) to get feedback on this. My opinion is that the extra warnings should be added for consistency.

fruffy:
> @jafingerhut / @fruffy: Any thoughts on whether these warnings should be added? I opened a ticket (#4439) to get feedback on this. My opinion is that the extra warnings should be added for consistency.

I do not have any strong opinions but I think your change makes sense.

But what about multi-pipe programs? Do the warnings work correctly here? I guess they should since the parser/control is also removed.

grg:
> > @jafingerhut / @fruffy: Any thoughts on whether these warnings should be added? I opened a ticket (#4439) to get feedback on this. My opinion is that the extra warnings should be added for consistency.
> 
> I do not have any strong opinions but I think your change makes sense.
> 
> But what about multi-pipe programs? Do the warnings work correctly here? I guess they should since the parser/control is also removed.

It works as expected with multi-pipe programs: a warning is emitted only if a control/parser is _never_ used in any of the pipes. If it's used in at least one pipe then no warning is emitted.

fruffy (approved):

grg:
@asl / @ChrisDodd Is this proposed change likely to impact either of you negatively?

asl:
> @asl / @ChrisDodd Is this proposed change likely to impact either of you negatively?

I believe we are ok. Cross-checking with @kfcripps

jafingerhut:
I see 3 approvals on this.  Are we waiting for explicit approval from @ChrisDodd ?

grg:
> I see 3 approvals on this. Are we waiting for explicit approval from @ChrisDodd ?

I've added it to the merge queue. This was put on the back burner last week while I was working on upstreaming the midend def-use pass from Tofino.

PR #4441, opened 2024-02-21
Avoid to-after-is double calls to RTTI in the type checker

vlstill:
These calls are redundant and cost extra. While this probably has little impact of the compiler speed with the new RTTI, this way of using `to` also makes it easier to see the cast is properly checker.

### DIFF
--- a/frontends/p4/typeChecking/typeChecker.cpp
+++ b/frontends/p4/typeChecking/typeChecker.cpp
@@ -335,7 +335,7 @@ const IR::Type *TypeInference::canonicalize(const IR::Type *type) {
 
     auto exists = typeMap->getType(type);
     if (exists != nullptr) {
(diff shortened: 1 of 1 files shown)


fruffy (approved):
A lot of these can probably be `const auto*`?

vlstill:
> A lot of these can probably be `const auto*`?

probably... one day I will try to configure `clang-tidy` or something to add these automatically. Most of the type checker does not even use the `auto *` convention for pointers/IR nodes.

fruffy:
> > A lot of these can probably be `const auto*`?
> 
> probably... one day I will try to configure `clang-tidy` or something to add these automatically. Most of the type checker does not even use the `auto *` convention for pointers/IR nodes.

Well... there is a PR open which adds clang-tidy support to our CMake. 😉  
We can run it selectively. Maybe it is even possible to parametrize it.

asl (approved):
Lots of similar code everywhere, yes!

PR #4442, opened 2024-02-21
Fixes from static analysis

vlstill:
- Add a BUG_CHECK
- constantFolding: Be a bit more defensive
- ir-generator: Make sure we don't dereference null when exploring class hierarchy
- Permit AutoCompileContext to throw in destructor if context is inconsistent
- typeChecker: Hanlde possible nullptr
- parserUnroll: Account for SymbolicValueFactory::create returning nullptr
- typeChecker: Avoid dereferencing null type after getType failure

### DIFF
--- a/frontends/p4/typeChecking/typeChecker.cpp
+++ b/frontends/p4/typeChecking/typeChecker.cpp
@@ -3074,7 +3074,7 @@ const IR::Node *TypeInference::postorder(IR::Slice *expression) {
     }
 
     auto e1type = getType(expression->e1);
-    if (e1type->is<IR::Type_SerEnum>()) {
+    if (e1type && e1type->is<IR::Type_SerEnum>()) {
         auto ei = EnumInstance::resolve(expression->e1, typeMap);
         CHECK_NULL(ei);
         auto sei = ei->to<SerEnumInstance>();
@@ -3085,7 +3085,7 @@ const IR::Node *TypeInference::postorder(IR::Slice *expression) {
         expression->e1 = sei->value;
     }
     auto e2type = getType(expression->e2);
-    if (e2type->is<IR::Type_SerEnum>()) {
+    if (e2type && e2type->is<IR::Type_SerEnum>()) {
         auto ei = EnumInstance::resolve(expression->e2, typeMap);
         CHECK_NULL(ei);
         auto sei = ei->to<SerEnumInstance>();
--- a/frontends/p4/typeChecking/typeChecker.cpp
+++ b/frontends/p4/typeChecking/typeChecker.cpp
@@ -3430,7 +3430,7 @@ const IR::Expression *TypeInference::actionCall(bool inActionList,
     }
     auto method = actionCall->method;
     auto methodType = getType(method);
-    if (!methodType) return actionCall; // error emitted in getType
+    if (!methodType) return actionCall;  // error emitted in getType
     auto baseType = methodType->to<IR::Type_Action>();
     if (!baseType) {
         typeError("%1%: must be an action", method);
--- a/frontends/p4/typeChecking/typeChecker.cpp
+++ b/frontends/p4/typeChecking/typeChecker.cpp
@@ -3326,8 +3326,10 @@ const IR::Node *TypeInference::postorder(IR::Member *expression) {
     }
 
     if (auto *apply = type->to<IR::IApply>(); apply && member == IR::IApply::applyMethodName) {
-        auto methodType = apply->getApplyMethodType();
-        methodType = canonicalize(methodType)->to<IR::Type_Method>();
+        auto *methodType = apply->getApplyMethodType();
+        auto *canon = canonicalize(methodType);
+        if (!canon) return expression;
+        methodType = canon->to<IR::Type_Method>();
         if (methodType == nullptr) return expression;
         learn(methodType, this);
         setType(getOriginal(), methodType);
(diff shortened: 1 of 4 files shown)


asl (approved):

tools/ir-generator/irclass.cpp
```diff
@@ -251,7 +251,7 @@ void IrDefinitions::generate(std::ostream &t, std::ostream &out, std::ostream &i
 }
 
 void IrClass::generateTreeMacro(std::ostream &out) const {
-    for (auto p = this; p != nodeClass(); p = p->getParent()) out << "  ";
+    for (auto p = this; p && p != nodeClass(); p = p->getParent()) out << "  ";
```
fruffy, line 254: This is a bit odd, I am guessing it doesn't like the nullptr comparison.

fruffy (approved):

frontends/p4/typeChecking/typeChecker.cpp
```diff
@@ -3073,7 +3074,7 @@ const IR::Node *TypeInference::postorder(IR::Slice *expression) {
     }
 
     auto e1type = getType(expression->e1);
```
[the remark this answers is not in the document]
vlstill: This was done the same as `getType` is used in most of other places. Based on the comment in `getType` itself, the `nullptr` can only be returned if an `error` was already emitted. If it was not emitted, there is a `BUG_CHECK`. So this is just a path to get multiple type errors out of the code.

https://github.com/p4lang/p4c/blob/16b9f07d83292645c2ac10adcc6d44e79fb570c8/frontends/p4/typeChecking/typeChecker.cpp#L192-L200

midend/parserUnroll.cpp
```diff
@@ -391,7 +391,7 @@ class ParserSymbolicInterpreter {
             }
 
             if (value == nullptr) value = factory->create(type, true);
-            if (value->is<SymbolicError>()) {
```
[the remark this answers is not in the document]
vlstill: Well the code below still checks if it is nullptr only only then sets it. That said, if it was nullptr it would crash until now, so presumably it never is :-). It seems to me based on reading the code that the factory could return `nullptr` if this pass was applied to a parser with non-inlined subparser. So that would suggest the pass probably should not be called.

Sadly, this is a complex code that is not well documented. If there was a documented prerequisite that the parsers has to be inlined I'd be happy adding an assert there. But it seems to that at least for the factory, someone was thinking about the case where they are not. I believe this modification is safe, adding the error would be a little bit more risky.

tools/ir-generator/irclass.cpp
fruffy: Comment was truncated. I meant to say this piece of code is a little funky in that it assumes that when p is nullptr it can terminate. The static analysis is right to add this check here but the code itself seems prone to errors.

midend/parserUnroll.cpp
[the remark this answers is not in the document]
fruffy: Makes sense, the parser unroll pass unfortunately does not have good error checking. I think this error should be handled correctly but this is not the point of this PR. The changes here just make this erroneous behavior explicit.

tools/ir-generator/irclass.cpp
vlstill: Yep, you are right in that, if it comes to nullptr it is a bug (elsewhere), which will now be reported.

tools/ir-generator/irclass.cpp
```diff
@@ -251,10 +251,14 @@ void IrDefinitions::generate(std::ostream &t, std::ostream &out, std::ostream &i
 }
 
 void IrClass::generateTreeMacro(std::ostream &out) const {
-    for (auto p = this; p != nodeClass(); p = p->getParent()) out << "  ";
+    {
```
fruffy, line 254: Any particular reason for the bracketing? You could just set p in the second loop again, no? Other than that approving early.

fruffy: Also would add brackets to the for loop, I completely missed the `out << " "`; on the first read.

fruffy (approved):

tools/ir-generator/irclass.cpp
vlstill: No particular reason, just limiting scope of the `p`. But I guess it looks too weird, changed.

vlstill: > Also would add brackets to the for loop, I completely missed the `out << " "`; on the first read.

Yes, this is one of my least favorite features of the code format we use :-(. Personally I tend to use brackets around single line blocks in p4c because of it, even though I never used them in other codebases.

But in general I tend to not change formatting of existing code that I just bug-fix.

vlstill:
Little tangential, but:

@fruffy, I am quite confused from how the merge queues work. I thought it was this PR in which the macOS test failed (https://github.com/p4lang/p4c/commit/ac6d9d532b264982fc0bd50740b77259b3a476d4 / https://github.com/p4lang/p4c/actions/runs/8100632186) but apparently it got still merged, without that test ever running. And your #4487 does not seem like it is using the queue at all. The merge queue surely does look very interesting, and I hope we can continue using it, but this is a bit weird.

fruffy:
Actually, this is working as intended. The merge queue will only block on required tests. We can make more tests required, but we have kept stuff like P4Testgen and ptf-linux optional.

fruffy:
The M1 test is not required yet, for example.

vlstill:
Interesting, it seems the merge queue is just running the required checks. That kind of makes sense, although I am not super keen of it hiding the rest. Do you know why some of the PRs are just in the "auto-merge" state, not in a queue?

PR #4443, opened 2024-02-21
Add support for parser errors in tc backend

vbnogueira:
Add support for parser errors in tc backend and also try to handle pna_main_input_metadata_t fields better by rejecting programs that use unsupported fields


vbnogueira:
@komaljai @usha1830 please take a look when you have some time

vbnogueira:
Seems like this is breaking the ebpf backend in some cases.
Will regroup and push a new version once we fix this

vbnogueira:
> Seems like this is breaking the ebpf backend in some cases. Will regroup and push a new version once we fix this

Fixed and pushed

backends/tc/ebpfCodeGen.cpp
```diff
@@ -1323,6 +1341,26 @@ ControlBodyTranslatorPNA::ControlBodyTranslatorPNA(const EBPF::EBPFControlPSA *c
...
+                    builder->append("compiler_meta__->parser_error");
+                    return false;
+                } else {
+                    ::error(ErrorType::ERR_UNSUPPORTED_ON_TARGET,
+                            "%1%: this metadata field is not supported", m);
+                }
+            }
+        }
+    }
+    return CodeGenInspector::preorder(m);
```
usha1830, line 1361: Is there any other way to achieve this? I have not seen visitor of other class being called like this in other parts of the code.

vbnogueira: @mouse-mt  says:

> Is there any other way to achieve this?

I don't know.  I haven't found any, but, given my lack of grokkage of
the compiler as a whole, that means little.  If the same effect can be
achieved some other way they like better, I have no problem with that.

> I have not seen visitor of other class being called like this in other parts of the code.

I'm not quite sure what this is trying to say.  In particular, I don't
see any visitor of any class being called in that code.

But I think of my code as just a slight adaptation of code that was
already there.  I'm not sure I recall everything.  But, based on a
quick look, I think I wrote that based on pieces such as

backends/tc/ebpfCodeGen.cpp, where a particular member name gets mapped
to a string not textually related to the member:

```
bool ControlBodyTranslatorPNA::checkPnaPortMem(const IR::Member *m) {
    if (m->expr != nullptr && m->expr->type != nullptr) {
        if (auto str_type = m->expr->type->to<IR::Type_Struct>()) {
            if (str_type->name == "pna_main_input_metadata_t" && m->member.name == "input_port") {
                return true;
            }
        }
    }
    return false;
}
[...]
            auto mem = c->expression->to<IR::Member>();
            builder->emitIndent();
            if (memcpy) {
[...]
            } else if (mem && checkPnaPortMem(mem)) {
                builder->appendFormat("%s.%s = ", keyname.c_str(), fieldName.c_str());
                builder->append("skb->ifindex");
            } else {
[...]
            }
```
backends/ebpf/codeGen.cpp, where references to member names are
converted to text:

```
bool CodeGenInspector::preorder(const IR::Member *expression) {
    bool isErrorAccess = false;
    if (auto tne = expression->expr->to<IR::TypeNameExpression>()) {
        if (auto tn = tne->typeName->to<IR::Type_Name>()) {
            if (tn->path->name.name == IR::Type_Error::error) isErrorAccess = true;
        }
    }

    int prec = expressionPrecedence;
    expressionPrecedence = expression->getPrecedence();
    auto ei = P4::EnumInstance::resolve(expression, typeMap);
    if (ei == nullptr && !isErrorAccess) {
        visit(expression->expr);
        auto pe = expression->expr->to<IR::PathExpression>();
        if (pe != nullptr && isPointerVariable(pe->path->name.name)) {
            builder->append("->");
        } else {
            builder->append(".");
        }
    }
    builder->append(expression->member);
    expressionPrecedence = prec;
    return false;
}
```

backends/dpdk/printUtils.cpp, which has a much smaller example of
converting a emember reference to text:

```
bool ConvertToString::preorder(const IR::Member *e) {
    out << toStr(e->expr) << "." << e->member.toString();
    return false;
}
```

I have the feeling I had a much more similar example of converting a
particular member name to special-case text, but I can't seem to find
it now with just a few greps.

usha1830: I wasn't asking about the way Member expression is translated, but was only concerned about the return statement with a preorder call for another unrelated class.
I just grepped for similar instances and found some in ebpf backend code. So, it looks fine to me as is.

usha1830 (approved):

Issue #4418, opened 2024-02-13
Reduce Visitor overhead


asl:
Visitors are central objects in P4C architecture. However, the current implementation imposes lots of overheads and some inefficiencies.

I would name some of those that draw my attention

1. They create lots of GC/malloc traffic. One of main sources of this is tracking visited nodes. For example, on P4CParserUnroll.switch_20160512 testcase we are having 50k calls to `Inspector::init_apply` and 138k calls to `Transform::init_apply`. On the real-world sources I had cases 10x larger. Each of such calls creates a new map to track visited nodes (just visited plus few bits for `Inspector` and `ChangeTracker` for `Modifier` and `Transform`). For some reasons these maps are wrapped into shared_ptr's, though I do not see where they are copied. `ForwardChildren` takes a reference to it.

What does matter is that majority of these maps are small. For that 50k calls to `Inspector` we are having 10k maps of size 1 (essentially node with no children), another 10k maps of size 2 (so, single children and nothing else) and approx 45k (so 90% of total) of maps of size <= 16. Rehashing and corresponding memory allocation take quite some visible amount of time in overall profile. Note that `std::unordered_map` is quite malloc-heavy due to restrictions and guarantees required by C++ standard. Situation with `Transforms` is similar – 85% of nodes are contained in small maps (< 16 entries).

We definitely need to optimize for small map case, e.g. having space pre-allocated and implementing map adapter on top of plain vector (aka `flat_map`). The linear scan over few vector elements won't be much slower than hash table lookup as evidence shows us.

The amount of map lookups is also quite large. We'd probably switch from `std::unordered_map` into some more performant solution.

2. Transforms are cloning always. Regardless whether node is changed or not. Likely this should be changed to some lazy implementation. However, it seems that some passes rely heavily on the way how nodes are currently traversed and what are the implementation details there.

3. Visitors are written in recursive manner. This generates huge call stacks and results in lots of virtual function calls for almost every action.

4. Visitors use textbook double-dispatch. So, every `apply` essentially costs 2 virtual function calls for each node.

Now this could be solved with static dispatch. With pre-defined typeids we can just switch over them, eliminating one virtual function call. The second virtual function call could be eliminating via CRTP.

That said, these are not something easy or isolated changes as Visitors are everywhere :) I will be contributing patches for some of these issues. Spoiler: for some downstream code we saw 4x reduction in the compile time via removing of various overheads here and there.

vlstill:
[...]

Issue #4389, opened 2024-02-04
hvec_map::insert has a non-standard behavior


asl:
`hvec_map::insert` looks like this:
```c++
    std::pair<iterator, bool> insert(const value_type &v) {
        bool new_key = false;
        size_t idx = hv_insert(&v.first);
        if (idx >= data.size()) {
            idx = data.size();
            data.push_back(v);
            new_key = true;
        } else {
            if ((new_key = erased[idx])) {
                erased[idx] = 0;
                const_cast<KEY &>(data[idx].first) = v.first;
                data[idx].second = v.second;
            } else {
                data[idx].second = v.second;
            }
        }
        return std::make_pair(iterator(*this, idx), new_key);
    }
```

Note that regardless whether element existed in the map or not, value is replaced. The `bool` result indicates whether the key was replaced and this is non-standard as well. Normally we expect from map inserts: a pair consisting of an iterator to the inserted element (or to the element that prevented the insertion) and a bool value set to true if and only if the insertion took place.

Such behavior could lead to very subtle bugs in common pattern `insert-or-merge`, e.g.:
```c++
hvec_map<key, value> m;
auto [it, inserted] = m.insert({something, something_else});
if (!inserted)
 merge(it->second, something_else);
```

`hvec_map::emplace` has the same problem.

asl:
Tagging @ChrisDodd 

I think the `else` in `if ((new_key = erased[idx])) {` should be removed to fix the issue. We cannot override value for non-erased entries.

ChrisDodd:
Fixed by #4458

PR #4447, opened 2024-02-22
Refactor some visitor internals

asl:
This is a first step towards #4418

The particular set of changes includes:
 - Refactor `Inspector`'s visiting tracking to be similar to `ChangeTracker` of `Modifier` / `Transform`. Albeit very similar, there are important differences here and there
 - Fixed subtle iterator invalidation bug in `Inspector::apply_visitor`. It was not exposed due to way `std::unordered_map` is implemented in libc++ / libstdc++. Long story short: iterator was kept across possible insertion. It will be invalidated in case of rehash
 - Ensured no internal guts of visiting process is exposed for the purposes of `visitOnce` / `visitAgain`. Previously we had a pointer to a field inside map's value stored in `Visitor`. So we risked a lot to obtain some dangling pointer to freed memory. It worked due to guarantees we had from `std::unordered_map`, but would stop to work with maps with less strong guarantees. And in general, it looks like an escaping pointer complicating code clarity.
 - Reduced size of `Visitor::Context` via rearranging fields to reduce alignment padding waste
 - Removed extraneous map lookups from hot code paths

As a result we are having ~5% speedup on `P4CParserUnroll.switch_20160512` testcase:

|:---|---:|---:|---:|---:|
| `gtestp4c-baseline --gtest_filter=P4CParserUnroll.switch_20160512` | 8.620 ± 0.113 | 8.413 | 8.749 | 1.05 ± 0.02 |
| `gtestp4c --gtest_filter=P4CParserUnroll.switch_20160512` | 8.200 ± 0.098 | 8.059 | 8.362 | 1.00 |

(benchmarking by hyperfine with 10 iterations and warm-up)

### DIFF
--- a/ir/visitor.cpp
+++ b/ir/visitor.cpp
@@ -36,6 +36,8 @@ limitations under the License.
 #include "lib/log.h"
 #include "lib/map.h"
 
+enum class VisitStatus : unsigned { New, Revisit, Busy, Done };
+
 /** @class Visitor::ChangeTracker
  *  @brief Assists visitors in traversing the IR.
 
@@ -58,15 +60,23 @@ class Visitor::ChangeTracker {
  public:
     /** Begin tracking @n during a visiting pass.  Use `finish(@n)` to mark @n as
      * visited once the pass completes.
+     *
+     * @return Status of the node @n wrt visit: whether node is currently being
+     * visited (`VisitStatus::Busy`), or if node was already visited
+     * (`VisitStatus::Done`), node was never seen before (`VisitStatus::New`) or
+     * seen, but should be revisited (`VisitStatus::Revisit`).
      */
-    [[nodiscard]] std::pair<bool, bool> start(const IR::Node *n, bool defaultVisitOnce) {
+    [[nodiscard]] VisitStatus try_start(const IR::Node *n, bool defaultVisitOnce) {
         // Initialization
         auto [it, inserted] = visited.emplace(n, visit_info_t{true, defaultVisitOnce, n});
 
-        bool busy = !inserted && it->second.visit_in_progress;
-        bool done = !inserted && !it->second.visit_in_progress && it->second.visitOnce;
+        if (!inserted) {  // We already seen this node, determine its status
+            if (it->second.visit_in_progress) return VisitStatus::Busy;
+            if (it->second.visitOnce) return VisitStatus::Done;
+            return VisitStatus::Revisit;
+        }
 
-        return {busy, done};
+        return VisitStatus::New;
     }
 
     /** Mark the process of visiting @orig as finished, with @final being the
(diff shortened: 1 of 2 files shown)


ir/visitor.cpp
```diff
@@ -311,25 +429,23 @@ const IR::Node *Modifier::apply_visitor(const IR::Node *n, const char *name) {
     if (ctxt) ctxt->child_name = name;
     if (n) {
         PushContext local(ctxt, n);
-        if (visited->busy(n)) {
+        auto [busy, done] = visited->start(n, visitDagOnce);
```
grg, line 432: In the original code, `start` really was only used when starting processing on a node. Now it has the concept of "get status and start if not already started". Perhaps a name like `update` / `update_status` / `start_or_status`?

ir/visitor.cpp
```diff
@@ -158,8 +155,121 @@ class Visitor::ChangeTracker {
...
+};
+
+/** @class Visitor::Tracker
+ *  @brief Assists visitors in traversing the IR.
+
+ *  A Tracker object assists visitors traversing the IR by tracking each
+ *  node.  The `start` method begins tracking, and `finish` ends it.  The
+ *  `done` method determines whether the node has been visited.
+ */
+class Visitor::Tracker {
```
grg, line 183: Can we just use ChangeTracker instead of introducing a new class just for Inspector passes? ChangeTracker does track the result object that this doesn't -- I haven't carefully compared every function to see what other differences there are.

The downside though to reusing ChangeTracker is that it is heaver weight -- I think there are far more Inspector passes than Modifier/Transform passes, so perhaps it would be observable (e.g., memory use/GC impact)?

grg (commented):
No major issues: just a couple of minor comments.

ir/visitor.cpp
asl: It was mixed thing actually. There was "implicit" start in `Inspector` and slightly different thing in `Modifier`/ `Transform`. I thought about `try_start`. But if there are other options, then I'm all for it :)

ir/visitor.cpp
asl: It's more heavier. And also it expects that there is something as `result` doing few extra lookups.

The impact of `Inspectors` is quite large, so I'd keep the code separate, the difference is quite visible.

asl:
@grg Renamed the function. Also it returns more clear status now, not just pair of bools

grg (approved):
LGTM. And the new VisitStatus enum is clearer than the pair of bools.

ir/visitor.h
```diff
@@ -351,16 +351,17 @@ class Visitor {
...
     void visit_children(const IR::Node *, std::function<void()> fn) { fn(); }
+    class Tracker;        // used by Inspector -- private to it
     class ChangeTracker;  // used by Modifier and Transform -- private to them
     // This overrides visitDagOnce for a single node -- can only be called from
     // preorder and postorder functions
-    void visitOnce() const { *visitCurrentOnce = true; }
-    void visitAgain() const { *visitCurrentOnce = false; }
+    // FIXME: It would be better named visitCurrentOnce() / visitCurrenAgain()
+    virtual void visitOnce() const { BUG("do not know how to handle request"); }
+    virtual void visitAgain() const { BUG("do not know how to handle request"); }
```
ChrisDodd, line 360: Maybe delete these here -- make them only in Inspector/Modifier/Transform, and final s well (no need for virtual)

ir/visitor.h
```diff
@@ -388,14 +389,12 @@ class Modifier : public virtual Visitor {
...
+    void visitAgain() const override;
 };
 
 class Inspector : public virtual Visitor {
-    struct info_t {
-        bool done, visitOnce;
-    };
-    typedef std::unordered_map<const IR::Node *, info_t> visited_t;
-    std::shared_ptr<visited_t> visited;
+    std::shared_ptr<Tracker> visited;
```
ChrisDodd, line 397: The point of having this be an info_t instead of a trscker is tha an Inspector can't change nodes, so only needs to store 2 bits per node visited and not an additional pointer.  If you change it to an hvec_map instead of an unordered_map, it would might be possible to make it even more compact.

asl: See next PRs for further improvements here :)

asl: And for now I cannot use it due to https://github.com/p4lang/p4c/issues/4389

Issue #1274, opened 2018-05-09
Fail to install the latest p4c


Wasdns:
Hi, all.

Today I try to install the latest version of p4c on my PC (running Ubuntu 14.04). I have installed all the dependencies required by p4c on my workspace. However, when I execute the command `make -j4`, I got the following error:

```
[  0%] Built target linkgraphs
[  0%] Built target update_includes
[  0%] Built target linkbmv2
[  0%] Built target linkp4cebpf
[  1%] Building CXX object lib/CMakeFiles/p4ctoolkit.dir/unified_libp4ctoolkit_srcs_1.cpp.o
[  1%] Built target mkv1dirs
[  1%] Built target linkp4test
[  1%] Built target mkp4dirs
[  1%] Built target mkP4configdir
[  3%] Building CXX object test/CMakeFiles/gtest.dir/frameworks/gtest/googletest/src/gtest-all.cc.o
[  5%] Copying p4c driver
[  5%] Built target p4c_driver
[  7%] Linking CXX static library libp4ctoolkit.a
[  7%] Built target p4ctoolkit
[  8%] [BISON][IRgenParser] Building parser with bison 3.0.2
[ 10%] [FLEX][IRgenLexer] Building scanner with flex 2.5.35
Scanning dependencies of target irgenerator
[ 12%] Linking CXX static library libgtest.a
[ 14%] Building CXX object tools/ir-generator/CMakeFiles/irgenerator.dir/unified_irgenerator_srcs_1.cpp.o
[ 16%] Building CXX object tools/ir-generator/CMakeFiles/irgenerator.dir/ir-generator.cpp.o
[ 16%] Built target gtest
In file included from ir-generator.ypp:126:0:
/home/sdn/p4c/build/tools/ir-generator/ir-generator-lex.c:577:5: warning: ‘{anonymous}::yy_flex_debug’ defined but not used [-Wunused-variable]
 int yy_flex_debug = 0;
     ^~~~~~~~~~~~~
/home/sdn/p4c/build/tools/ir-generator/ir-generator-lex.c:355:5: warning: ‘{anonymous}::yylineno’ defined but not used [-Wunused-variable]
 int yylineno = 1;
     ^~~~~~~~
[ 17%] Linking CXX executable irgenerator
[ 17%] Built target irgenerator
[ 19%] Generating IR class files
[...]

PR #4452, opened 2024-02-22
Allow annotations on functions

ChrisDodd:
Probably an oversight that this wasn't allowed before.  Use case is to specify which hardware resource(s) should be used to implement a specific function

### DIFF
--- a/frontends/parsers/p4/p4parser.ypp
+++ b/frontends/parsers/p4/p4parser.ypp
@@ -1413,7 +1413,11 @@ initializer
 /**************** Expressions ****************/
 
 functionDeclaration
-    : functionPrototype blockStatement   {
+    : annotations functionPrototype blockStatement   {
+            auto ann = new IR::Annotations(@1, *$1);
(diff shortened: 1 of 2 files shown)


jafingerhut:
Good addition.  Can you add a simple test program using this feature, too?

jafingerhut:
Do you want to create an issue and/or PR for the language spec, too?  If you would prefer somone else do that, I can do it.

ChrisDodd:
> Good addition. Can you add a simple test program using this feature, too?

I can add a testcase defining a trivial function with a random annotation, but that really only tests that the perser accepts it (which is pretty trivial -- not sure if it is worth the extra 1 second of CI time)

ChrisDodd:
I added a spec (grammar) change to match [#1274](https://github.com/p4lang/p4-spec/pull/1274)

jafingerhut (approved):
LGTM, although I am not the best person to review the C++ parts of the changes, which is most of this :-)

PR #4458, opened 2024-02-23
Fix hvec_map insert/emplace

ChrisDodd:
- should not replace the value if the key is already present


asl:
Can we also add `count` while here?

ChrisDodd:
> Can we also add `count` while here?

sure

asl (approved):
LGTM, thanks!

PR #4459, opened 2024-02-23
Use better maps to store visitor state

asl:
This PR imports modern hash map implementation from https://github.com/greg7mdp/parallel-hashmap Essentially it's a hash map extracted from Abseil plus some additional stuff on top of that. Note that while it is a drop-in replacement of `std::unordered_map` it has different guarantees wrt iterator invalidation as it uses open-addressing scheme.

Advantages of this map is:
 - Much faster lookups
 - Much less malloc traffic
 - Better memory usage

On our downstream codebase just this switch alone yields ~2.1x compile time reduction on some large P4C apps.

The PR contains several commits to measure effects:
 - Use of map itself (`gtestp4c-phmap` below)
 - Preallocation of 16 slots (we know that in majority of cases this would be enough) (`gtestp4c-phmap-16` below)
 - Switch from `std::hash` to `Utils::Hash` (`gtestp4c-phmap-16-hash` below)

To compare with `hvec_map` I also benchmarked wrt it (`gtestp4c-hmap`). The benchmarking results are:

|:---|---:|---:|---:|---:|
| `gtestp4c-baseline --gtest_filter=P4CParserUnroll.switch_20160512` | 8.375 ± 0.177 | 8.118 | 8.648 | 1.25 ± 0.04 |
| `gtestp4c-phmap --gtest_filter=P4CParserUnroll.switch_20160512` | 6.886 ± 0.173 | 6.563 | 7.061 | 1.03 ± 0.03 |
| `gtestp4c-phmap-16 --gtest_filter=P4CParserUnroll.switch_20160512` | 6.722 ± 0.118 | 6.539 | 6.900 | 1.00 ± 0.03 |
| `gtestp4c-phmap-16-hash --gtest_filter=P4CParserUnroll.switch_20160512` | 6.705 ± 0.147 | 6.546 | 7.038 | 1.00 |
| `gtestp4c-hmap --gtest_filter=P4CParserUnroll.switch_20160512` | 7.547 ± 0.099 | 7.356 | 7.689 | 1.13 ± 0.03 |

To summarize: it yields 25% improvements against the baseline `std::unordered_map` and 13% improvements towards `hvec_map`. As I already said, for some of our downstream apps the difference is more than two-fold, the compilation time reduced from 5m 14s down to 2m 29s (`hvec_map` was not that attractive there sadly). I also saw some marginal reduction of peak RSS from 15.3 GiB down to 14.87 GiB.

To give more benchmarking results, here are the whole gtest times (here the proportion of `Visitor` code is much less):

|:---|---:|---:|---:|---:|
| `gtestp4c-baseline` | 14.398 ± 0.310 | 13.920 | 14.923 | 1.17 ± 0.03 |
| `gtestp4c-hmap` | 13.083 ± 0.280 | 12.683 | 13.496 | 1.06 ± 0.03 |
| `gtestp4c-phmap-16-hash` | 12.338 ± 0.256 | 12.003 | 12.877 | 1.00 |

### DIFF
--- a/ir/CMakeLists.txt
+++ b/ir/CMakeLists.txt
@@ -64,7 +64,8 @@ set (BASE_IR_DEF_FILES
 set (IR_DEF_FILES ${IR_DEF_FILES} ${BASE_IR_DEF_FILES} PARENT_SCOPE)
 
 add_library (ir STATIC ${IR_SRCS})
-target_link_libraries(ir absl::flat_hash_map)
+target_link_libraries(ir
+  PRIVATE absl::flat_hash_map)
 add_dependencies(ir genIR)
 
 
--- a/ir/CMakeLists.txt
+++ b/ir/CMakeLists.txt
@@ -64,6 +64,7 @@ set (BASE_IR_DEF_FILES
 set (IR_DEF_FILES ${IR_DEF_FILES} ${BASE_IR_DEF_FILES} PARENT_SCOPE)
 
 add_library (ir STATIC ${IR_SRCS})
+target_link_libraries(ir absl::flat_hash_map)
 add_dependencies(ir genIR)
 
 
--- a/ir/visitor.cpp
+++ b/ir/visitor.cpp
@@ -19,8 +19,8 @@ limitations under the License.
 #include <stdlib.h>
 #include <time.h>
 
+#include "absl/container/flat_hash_map.h"
 #include "ir/ir-generated.h"
-#include "lib/source_file.h"
 
 #if HAVE_LIBGC
 #include <gc/gc.h>
@@ -54,7 +54,7 @@ class Visitor::ChangeTracker {
         bool visitOnce;
         const IR::Node *result;
     };
-    typedef std::unordered_map<const IR::Node *, visit_info_t> visited_t;
+    using visited_t = absl::flat_hash_map<const IR::Node *, visit_info_t>;
     visited_t visited;
 
  public:
@@ -130,7 +130,7 @@ class Visitor::ChangeTracker {
     void revisit_visited() {
         for (auto it = visited.begin(); it != visited.end();) {
             if (!it->second.visit_in_progress)
-                it = visited.erase(it);
+                visited.erase(it++);
             else
                 ++it;
         }
@@ -195,7 +195,7 @@ class Visitor::Tracker {
     struct info_t {
         bool done, visitOnce;
     };
-    typedef std::unordered_map<const IR::Node *, info_t> visited_t;
+    using visited_t = absl::flat_hash_map<const IR::Node *, info_t>;
     visited_t visited;
 
  public:
@@ -204,7 +204,7 @@ class Visitor::Tracker {
     void revisit_visited() {
         for (auto it = visited.begin(); it != visited.end();) {
             if (it->second.done)
-                it = visited.erase(it);
+                visited.erase(it++);
             else
                 ++it;
         }
--- a/ir/visitor.cpp
+++ b/ir/visitor.cpp
@@ -179,13 +179,13 @@ class Visitor::ChangeTracker {
 
     /** Produce the final result of visiting @n.
      *
-     * @return The ultimate result of visiting @n, or @n if `finish(@n)` has not
+     * @return The ultimate result of visiting @n, or `nullptr` if `finish(@n)` has not
      * been invoked.
      */
     const IR::Node *finalResult(const IR::Node *n) const {
         auto it = visited.find(n);
         bool done = it != visited.end() && !it->second.visit_in_progress && it->second.visitOnce;
-        return done ? it->second.result : n;
+        return done ? it->second.result : nullptr;
     }
 
     void visitOnce(const IR::Node *n) {
@@ -459,7 +459,8 @@ namespace {
 class ForwardChildren : public Visitor {
     const ChangeTracker &visited;
     const IR::Node *apply_visitor(const IR::Node *n, const char * = 0) {
-        return visited.finalResult(n);
+        if (const auto *result = visited.finalResult(n)) return result;
+        return n;
     }
 
  public:
--- a/ir/visitor.cpp
+++ b/ir/visitor.cpp
@@ -173,6 +173,17 @@ class Visitor::ChangeTracker {
         return it->second.result;
     }
 
+    /** Produce the final result of visiting @n.
+     *
+     * @return The ultimate result of visiting @n, or @n `finish(@n)` has not
+     * been invoked.
+     */
+    const IR::Node *finalResult(const IR::Node *n) const {
+        auto it = visited.find(n);
+        bool done = it != visited.end() && !it->second.visit_in_progress && it->second.visitOnce;
+        return done ? it->second.result : n;
+    }
+
     void visitOnce(const IR::Node *n) {
         auto it = visited.find(n);
         if (it == visited.end()) BUG("visitor state tracker corrupted");
@@ -440,8 +451,7 @@ namespace {
 class ForwardChildren : public Visitor {
     const ChangeTracker &visited;
     const IR::Node *apply_visitor(const IR::Node *n, const char * = 0) {
-        if (visited.done(n)) return visited.result(n);
-        return n;
+        return visited.finalResult(n);
     }
 
  public:
--- a/ir/visitor.cpp
+++ b/ir/visitor.cpp
@@ -58,6 +58,8 @@ class Visitor::ChangeTracker {
     visited_t visited;
 
  public:
+    ChangeTracker() : visited(16) {}
+
     /** Begin tracking @n during a visiting pass.  Use `finish(@n)` to mark @n as
      * visited once the pass completes.
      *
@@ -199,6 +201,8 @@ class Visitor::Tracker {
     visited_t visited;
 
  public:
+    Tracker() : visited(16) {}
+
     /** Forget nodes that have already been visited, allowing them to be visited
      * again. */
     void revisit_visited() {
--- a/ir/visitor.cpp
+++ b/ir/visitor.cpp
@@ -135,6 +135,8 @@ class Visitor::ChangeTracker {
     void revisit_visited() {
         for (auto it = visited.begin(); it != visited.end();) {
             if (!it->second.visit_in_progress)
+                // `visited` is abseil map, therefore erase does not return iterator, use
+                // post-increment
                 visited.erase(it++);
             else
                 ++it;
@@ -224,6 +226,8 @@ class Visitor::Tracker {
     void revisit_visited() {
         for (auto it = visited.begin(); it != visited.end();) {
             if (it->second.done)
+                // `visited` is abseil map, therefore erase does not return iterator, use
+                // post-increment
                 visited.erase(it++);
             else
                 ++it;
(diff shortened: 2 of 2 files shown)


asl:
For the record, I also tried https://github.com/Tessil/robin-map – it was slower than phmap. And it is not a drop-in replacement, so code changed are required.

fruffy:
Just two notes: 

- We may soon get Abseil "for free" because the compiler's Protobuf dependency will pull it in with Protobuf 22.x: https://protobuf.dev/news/2022-08-03/#abseil-dep. We simply haven't pulled the trigger on an upgrade. 

- P4Testgen heavily depends on the visitors' multiple dispatch so it should be affected by these changes. I will try to add a benchmark. We could also write up independent visitor microbenchmarks.

grg:
> * We may soon get Abseil "for free" because the compiler's Protobuf dependency will pull it in with Protobuf 22.x: https://protobuf.dev/news/2022-08-03/#abseil-dep. We simply haven't pulled the trigger on an upgrade.

Is there an ETA for when Protobuf 22.x upgrade is likely to occur? Wondering whether it makes sense for this change to wait so that it can avoid adding and later removing the Abseil files?

fruffy:
Another thing to do is to break down our boost dependencies into the independent modules which are now available. For example:
https://github.com/boostorg/format
https://github.com/boostorg/multiprecision

That would simplify things instead of pulling in `libboost-all-dev` always.

asl:
> * We may soon get Abseil "for free" because the compiler's Protobuf dependency will pull it in with Protobuf 22.x: https://protobuf.dev/news/2022-08-03/#abseil-dep. We simply haven't pulled the trigger on an upgrade.
> *

Note that it started as a library from there. But since that time it had some changes here and there and also there are some notable differences. Certainly it would worth checking and either keeping it or replacing with abseil maps.

> P4Testgen heavily depends on the visitors' multiple dispatch so it should be affected by these changes. I will try to add a benchmark. We could also write up independent visitor microbenchmarks.

Nothing wrt multiple dispatch here :) We are reducing malloc / GC traffic here and speeding up basic operations (insert / find).

ir/visitor.cpp
```diff
@@ -195,10 +209,12 @@ class Visitor::Tracker {
     struct info_t {
         bool done, visitOnce;
     };
-    typedef std::unordered_map<const IR::Node *, info_t> visited_t;
+    using visited_t = phmap::parallel_flat_hash_map<const IR::Node *, info_t, Util::Hash>;
     visited_t visited;
 
  public:
+    Tracker() : visited(16) {}
```
grg, line 216: Please add a very brief comment explaining the number 16 🙂 E.g., "pre-allocating 16 slots yields a performance improvement and is sufficient for most cases".

I want to avoid a "huh?" response when people see this code later on.

asl: Yeah! Likely we will change with inlined pre-allocated storage in the future, but one step at a time.

asl:
> Doesn't this change speed up traversal for Transform and Inspector, too? I'd assume it would help with interpreters like P4Testgen.

Yes, it speeds-up the traverse a lot. But not the multiple-dispatch itself.

> Well one thing we could try is to bump the Protobuf version and see what happens. If nothing breaks we could add a PR for it, then rebase this one top and link against the absl map. In parallel, we can try to clean up the other dependencies.

Let me give it a try and see what will happen :)

fruffy:
> If abseil is a dependency, then you could use `absl::format` as well :)

Iirc the problem is the difference in format specifiers. Boost uses `%1%, %2%,...` which may not be supported in the other frameworks. The migration is finicky.

asl:
> Iirc the problem is the difference in format specifiers. Boost uses `%1%, %2%,...` which may not be supported in the other frameworks. The migration is finicky.

Yes. I do not recall how well `std::format` is supported across the compilers off-hand. Migration of specifiers is a pain, I confirm (I once migrated large codebase to https://github.com/fmtlib/fmt)

ChrisDodd:
I'd be interested in seeing why hvec_map is so much slower than this, given it is using essentially the same algorithms.  You could try giving it an initial capacity of 16 as well, and maybe a different hash function (what is the default has function for pointers -- just identity?  That would seem to be pretty good.)

asl:
> You could try giving it an initial capacity of 16 as well, and maybe a different hash function (what is the default has function for pointers -- just identity? That would seem to be pretty good.)

The benchmarking above already do this (pre-allocation of capacity plus better hash function, we do compare apples-to-apples). I tried `hvec_map` in other cases as well (PRs pending) and the overall outcome is the same: `hvec_map` might be faster than `unordered_map`, but always slower than these "swiss hash maps".

asl:
Rebased on top of abseil. Here are the updated results (the differences with tables above is the non-unity build here):

|:---|---:|---:|---:|---:|
| `gtestp4c-baseline --gtest_filter=P4CParserUnroll.switch_20160512` | 8.467 ± 0.164 | 8.150 | 8.682 | 1.38 ± 0.06 |
| `gtestp4c --gtest_filter=P4CParserUnroll.switch_20160512` | 6.134 ± 0.214 | 5.861 | 6.446 | 1.00 |

ir/visitor.cpp
```diff
@@ -130,7 +135,7 @@ class Visitor::ChangeTracker {
     void revisit_visited() {
         for (auto it = visited.begin(); it != visited.end();) {
             if (!it->second.visit_in_progress)
-                it = visited.erase(it);
+                visited.erase(it++);
```
grg, line 138: I had to look up why you make this change: I see that Abseil returns void on erase instead of an iterator.

ir/visitor.cpp
```diff
@@ -170,6 +175,17 @@ class Visitor::ChangeTracker {
         return it->second.result;
     }
 
+    /** Produce the final result of visiting @n.
+     *
+     * @return The ultimate result of visiting @n, or @n `finish(@n)` has not
```
grg, line 180:
```suggestion
     * @return The ultimate result of visiting @n, or @n if `finish(@n)` has not
```

grg (approved):
LGTM, but please please give the other reviewers a chance to provide feedback before merging.

(The speedups sound great!)

ir/visitor.cpp
asl: Yes. Unfortunately it's something they are not compatible with std maps. Though it is explicitly specified that post-incrementing iterator would work.

asl:
> LGTM, but please please give the other reviewers a chance to provide feedback before merging.
> 
> (The speedups sound great!)

Absolutely!

grg:
One follow-up: it would be good to include a small section in the developer docs (e.g., docs/README.md -- maybe a "performance considerations" section?) as to why some collections are now Abseil. You could also consider adding guidance on when code might benefit from Abseil collections and how to switch.

Feel free to do this as a separate PR or part of #4473.

asl:
> One follow-up: it would be good to include a small section in the developer docs (e.g., docs/README.md -- maybe a "performance considerations" section?) as to why some collections are now Abseil. You could also consider adding guidance on when code might benefit from Abseil collections and how to switch.

Good idea. I think I'd also mention `InlinedVector`

ir/visitor.cpp
vlstill: Please add a comment about this (the post-increment). This is non-intuitive for anyone not very familiar with abseil maps.

ir/visitor.cpp
```diff
@@ -170,6 +175,17 @@ class Visitor::ChangeTracker {
         return it->second.result;
     }
 
+    /** Produce the final result of visiting @n.
+     *
+     * @return The ultimate result of visiting @n, or @n if `finish(@n)` has not
+     * been invoked.
+     */
+    const IR::Node *finalResult(const IR::Node *n) const {
```
vlstill, line 183: I think the interface could be a bit confusing, maybe it would be more straight-forward if it returned a final result, or nullptr if the final result does not exits?

vlstill (commented):
The speedup does look very interesting, thank you.

We could probably get even more by replacing standard maps sprinkled around the code base, but that would be significantly more work (and we would also need a better `ordered_map`/`ordered_set`).

asl:
> We could probably get even more by replacing standard maps sprinkled around the code base, but that would be significantly more work (and we would also need a better `ordered_map`/`ordered_set`

See https://github.com/p4lang/p4c/pull/4473 just in case :) Some care should be taken here as these maps have a bi different semantics about iterator invalidation and pointers stability, so we cannot just automatically replaces ones with others. However, it is definitely beneficial in few hot code paths

ir/CMakeLists.txt
```diff
@@ -64,6 +64,7 @@ set (BASE_IR_DEF_FILES
 set (IR_DEF_FILES ${IR_DEF_FILES} ${BASE_IR_DEF_FILES} PARENT_SCOPE)
 
 add_library (ir STATIC ${IR_SRCS})
+target_link_libraries(ir absl::flat_hash_map)
```
fruffy, line 67: Not sure how we want to manage linking... Should each dependent target link absl::flat_hash_map, should `ir` export it? Also unclear to me what the best approach is to keep binary size small. 
I am guess we should make `PUBLIC` explicit here.

ir/visitor.cpp
grg: @vlstill I also thought about asking for a comment but I'm in two minds about it:
* It is different to the STL interface.
* If the Abseil collections spread to more places, are we going to add a comment every time the Abseil interface differs from the STL interface?

I'm happy to accept the argument that Visitor requires extra documentation because it is a core class, and so more people will look at the code to better understand how the compiler works/improve its performance.

ir/CMakeLists.txt
asl: I'll revise this PR wrt dependencies

fruffy: `add_dependency` was used because historically all the libs were built independently and linked together by each back end. However, to ensure proper build order in parallel builds, `add_dependency` was needed. 

The fine-grained approach works for me. #4474 implements some of this as I ran into some issues downstream.

asl: > `add_dependency` was used because historically all the libs were built independently and linked together by each back end. However, to ensure proper build order in parallel builds, `add_dependency` was needed.

Yes. But this is not how the dependencies should be done. Things are pulling headers, so this is a build-time dependency, not a link-time...

fruffy: With binary size I meant the final compiler binaries, which can be huge. I believe this is primarily caused by all the templating and ir-generated header, but also by the chaotic linking, which also seems to affect linking time. It can be excruciatingly slow.

fruffy: > Yes. But this is not how the dependencies should be done. Things are pulling headers, so this is a build-time dependency, not a link-time... #4473 also has some... we'd need to unify all somehow

Yeah I agree with you, the compiler infrastructure has been going through various phases of structuring with each back end doing their own thing.

ir/visitor.cpp
asl: Let's add the comment for now. I believe this is the only place where abseil maps are different interface-wise (sans heterogeneous lookup, but we will have it for `unordered_map` with C++20).

The behavior wrt pointer / iterators stability is different, but I will take care about documentation separately

fruffy (approved):

fruffy:
@asl I removed this from the merge queue because @vlstill and @grg should still give their final OK.

asl:
> @asl I removed this from the merge queue because @vlstill and @grg should still give their final OK.

Ok, no problem

vlstill (approved):

PR #4460, opened 2024-02-23
P4TC - Change header fields to network order in C Program

komaljai:
Convert all header fields to network order in EBPF C Program.
Handle corresponding arithmetic operations with respect to endianness in parser and control block.


fruffy:
Byte order is a long-standing problem in the eBPF back end. https://github.com/p4lang/p4c/issues/1122

Does this PR address the issue?

komaljai:
> Byte order is a long-standing problem in the eBPF back end. #1122
> 
> Does this PR address the issue? We may not even need to make the distinction between "TC" and eBPF then.

Thanks @fruffy for pointing the EBPF back-end open issue.
This PR addresses the same issue, however due to several opens, have currently marked it as Draft.

komaljai:
This is no longer required in TC backend.

PR #4463, opened 2024-02-23
Bump protobuf version and add Abseil as compiler dependency. 

asl:
Fix some missed dependencies here and there

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -151,6 +151,16 @@ if(BUILD_STATIC_RELEASE_SANS_GLIBC)
   add_definitions(-DP4C_STATIC_BUILD)
 endif()
 
+# Ensure we enable sanitizers before fetching dependencies so the correct flags
+# are propagated below
+if (ENABLE_SANITIZERS)
+  append("-fsanitize=undefined,address" CMAKE_C_FLAGS CMAKE_CXX_FLAGS)
+endif ()
+
+if (BUILD_AUTO_VAR_INIT_PATTERN)
+  add_cxx_compiler_option  ("-ftrivial-auto-var-init=pattern")
+endif ()
+
 # Required tools and libraries.
 find_package (PythonInterp 3 REQUIRED)
 find_package (FLEX 2.0 REQUIRED)
@@ -268,14 +278,6 @@ if (CMAKE_CXX_COMPILER_ID STREQUAL "Clang" OR CMAKE_CXX_COMPILER_ID STREQUAL "Ap
   # FIXME: with C++20 we would be able to use standard __VA_OPT__
   add_cxx_compiler_option ("-Wno-gnu-zero-variadic-macro-arguments")
 endif()
-
-if (ENABLE_SANITIZERS)
-  append("-fsanitize=undefined,address" CMAKE_C_FLAGS CMAKE_CXX_FLAGS)
-endif ()
-
-if (BUILD_AUTO_VAR_INIT_PATTERN)
-  add_cxx_compiler_option  ("-ftrivial-auto-var-init=pattern")
-endif ()
 
 if (ENABLE_WERROR)
   add_cxx_compiler_option  ("-Werror")
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -23,6 +23,10 @@ endif()
 
 project (P4C)
 
+# set (CMAKE_CXX_EXTENSIONS OFF) # prefer using -std=c++17 rather than -std=gnu++17
+set (CMAKE_CXX_STANDARD 17)
+set (CMAKE_CXX_STANDARD_REQUIRED ON)
+
 set (CMAKE_MODULE_PATH ${PROJECT_SOURCE_DIR}/cmake)
 set (P4C_RUNTIME_OUTPUT_DIRECTORY bin)
 set (P4C_LIBRARY_OUTPUT_DIRECTORY lib)
@@ -151,7 +155,13 @@ endif()
 find_package (PythonInterp 3 REQUIRED)
 find_package (FLEX 2.0 REQUIRED)
 find_package (BISON 3.0 REQUIRED)
-find_package (absl REQUIRED)
+# If we enable GTest, install GoogleTest with the appropriate options.
+if (ENABLE_GTESTS)
+  include(GoogleTest)
+  p4c_obtain_googletest()
+endif ()
+include(Abseil)
+p4c_obtain_abseil()
 include(Protobuf)
 p4c_obtain_protobuf()
 
@@ -242,10 +252,6 @@ enable_testing ()
 # if (__P4C_VERSION_RC)
 #   set (CPACK_PACKAGE_VERSION_PATCH ${CPACK_PACKAGE_VERSION_PATCH}-${__P4C_VERSION_RC})
 # endif ()
-
-# set (CMAKE_CXX_EXTENSIONS OFF) # prefer using -std=c++17 rather than -std=gnu++17
-set (CMAKE_CXX_STANDARD 17)
-set (CMAKE_CXX_STANDARD_REQUIRED ON)
 
 add_cxx_compiler_option ("-Wall")
 add_cxx_compiler_option ("-Wextra")
@@ -358,12 +364,6 @@ set (IR_GENERATOR ${IR_GENERATOR_DIRECTORY}/irgenerator)
 
 # Component libraries: must be defined before being used in the back end executables.
 set (P4C_LIBRARIES controlplane midend frontend ir p4ctoolkit ir-generated)
-
-# If we enable GTest, install GoogleTest with the appropriate options.
-if (ENABLE_GTESTS)
-  include(GoogleTest)
-  p4c_obtain_googletest()
-endif ()
 
 # The order of adding subdirectories matters because of target dependencies.
 add_subdirectory (tools/driver)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -151,6 +151,7 @@ endif()
 find_package (PythonInterp 3 REQUIRED)
 find_package (FLEX 2.0 REQUIRED)
 find_package (BISON 3.0 REQUIRED)
+find_package (absl REQUIRED)
 include(Protobuf)
 p4c_obtain_protobuf()
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -155,6 +155,9 @@ endif()
 # are propagated below
 if (ENABLE_SANITIZERS)
   append("-fsanitize=undefined,address" CMAKE_C_FLAGS CMAKE_CXX_FLAGS)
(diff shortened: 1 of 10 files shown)


asl:
@fruffy This just uses system abseil. One would likely need to update bazel and what's on builder.

We can use `FetchContent` for abseil. Or... we can requre it globally as boost. What are the preferences?

fruffy:
I see you changed the option to use the system-provided Abseil. So, using the Abseil provided by Protobuf does not work?

fruffy:
I have been slowly migrating all our dependencies to FetchContent for the reasons you have listed. It is otherwise too difficult to keep track of all the different system-provided versions. If we add Abseil it should be via FetchContent. 

In general we try to give users the option to use a system-provided version.

asl:
> I see you changed the option to use the system-provided Abseil. So, using the Abseil provided by Protobuf does not work?

Not the way we're fetching it :) We just downloading a release tarball that does not include abseil sources. So the options are:
 - Pull from git tag
 - Pull abseil separately

fruffy:
I could take a crack at this myself but I do not know whether I can get around to it the next couple of days.

asl:
> I could take a crack at this myself but I do not know whether I can get around to it the next couple of days.

I'll try to do this and then we will see where we will be.

asl:
I had to bump to the latest protobuf (25.3) as otherwise it won't work with latest LTS Abseil. Apparently protobuf relies on some abseil internals and therefore there is version lock :)

cmake/Abseil.cmake
```diff
@@ -0,0 +1,56 @@
+macro(p4c_obtain_abseil)
+  option(
+    P4C_USE_PREINSTALLED_ABSEIL
+    "Look for a preinstalled version of Protobuf in the system instead of installing a prebuilt binary using FetchContent."
```
fruffy, line 4:
```suggestion
    "Look for a preinstalled version of Abseil in the system instead of installing the library using FetchContent."
```

cmake/Abseil.cmake
```diff
@@ -0,0 +1,56 @@
...
+  )
+
+  # If P4C_USE_PREINSTALLED_ABSEIL is ON just try to find a preinstalled version of Abseil.
+  if(P4C_USE_PREINSTALLED_ABSEIL)
+    if(ENABLE_ABSEIL_STATIC)
+      set(SAVED_CMAKE_FIND_LIBRARY_SUFFIXES ${CMAKE_FIND_LIBRARY_SUFFIXES})
+      set(CMAKE_FIND_LIBRARY_SUFFIXES .a)
+    endif()
+    find_package(absl CONFIG)
+    if(NOT absl_FOUND)
```
fruffy, line 15: This was a hack required for bad versions of Protobuf. We might not need it for Abseil.

cmake/Abseil.cmake
```diff
@@ -0,0 +1,56 @@
...
+    endif()
+
+    if(ENABLE_ABSEIL_STATIC)
+      set(CMAKE_FIND_LIBRARY_SUFFIXES ${SAVED_CMAKE_FIND_LIBRARY_SUFFIXES})
+    endif()
+  else()
+    set(P4C_ABSEIL_VERSION "20240116.1")
+    message("Fetching Abseil version ${P4C_ABSEIL_VERSION} for P4C...")
+
+    # Unity builds do not work for Abseil...
```
fruffy, line 26: Is this still true?

fruffy (approved):
Wonderful, thanks for getting this done!

fruffy:
@smolkaj @jonathan-dilorenzo Please take a look.

cmake/Abseil.cmake
asl: Yes. I had to add this explicitly. There are some global constants

cmake/Abseil.cmake
asl: Yeah. Actually we might just use policy instead

asl:
Note that bazel build of p4runtime pulls abseil as well. However, rather old version – protobuf cannot work with it. So, include order does matter :)

fruffy:
We should update the README (https://github.com/p4lang/p4c/edit/main/README.md) since we are adding a new dependency. There are quite a few Protobuf-specific segments. I think that README is the only place though.

cmake/Abseil.cmake
asl: https://cmake.org/cmake/help/latest/variable/CMAKE_FIND_PACKAGE_PREFER_CONFIG.html Looks like this is what is needed :)

asl:
@fruffy I think I updated readme

asl:
Looks like we are having some asan / ubsan fails
1. Ubsan is an old UB in JsonParser. I will fix it.
2. Asan is somewhere in protobuf, but I do not see it in MacOS. Will check it out.

asl:
ASAN issue: https://github.com/protocolbuffers/protobuf/issues/13115

asl:
Now we're having a known issue where ubsan with 128-bit integers tries to pull symbols from compiler-rt (that does not exist in libgcc):

```
/usr/bin/ld: ../abseil-build/absl/strings/libabsl_strings.a(numbers.cc.o): in function `bool absl::lts_20240116::(anonymous namespace)::safe_parse_positive_int<absl::lts_20240116::int128>(std::basic_string_view<char, std::char_traits<char> >, int, absl::lts_20240116::int128*)':
numbers.cc:(.text+0x9d69): undefined reference to `__muloti4'
```

See https://bugs.llvm.org/show_bug.cgi?id=16404 for more information. I'll try to see what might be the most viable workaround here

fruffy:
Thanks for looking into this! Working with Protobuf can be a pain.

CMakeLists.txt
```diff
@@ -147,10 +151,30 @@ if(BUILD_STATIC_RELEASE_SANS_GLIBC)
   add_definitions(-DP4C_STATIC_BUILD)
 endif()
 
+# Ensure we enable sanitizers before fetching dependencies so the correct flags
+# are propagated below
+if (ENABLE_SANITIZERS)
+  append("-fsanitize=undefined,address" CMAKE_C_FLAGS CMAKE_CXX_FLAGS)
+  if (CMAKE_CXX_COMPILER_ID STREQUAL "Clang" AND UNIX AND NOT APPLE)
+    append("--rtlib=compiler-rt --unwindlib=libgcc" CMAKE_EXE_LINKER_FLAGS)
```
fruffy, line 159: Seems like this works, could you add an explanation with the link why this was necessary?

asl: Yes, just wanted to double check it will fix the CI

README.md
```diff
@@ -193,13 +193,15 @@ sudo dpkg -i /path/to/package.deb
        library. Default is ON.
      - `-DENABLE_GTESTS=ON|OFF`. Enable building and running GTest unit tests.
        Default is ON.
-     - `-DP4C_USE_PREINSTALLED_PROTOBUF=ON|OFF`. Try to find a system version of Protobuf instead of a CMake version.
+     - `-DP4C_USE_PREINSTALLED_ABSEIL=ON|OFF`. Try to find a system version of Abseil instead of a fetched one. Default is OFF.
+     - `-DP4C_USE_PREINSTALLED_PROTOBUF=ON|OFF`. Try to find a system version of Protobuf instead of a CMake version. Default is OFF.
+     - `-DENABLE_ABSEIL_STATIC=ON|OFF`. Enable the use of static abseil libraries. Default is ON. Only has an effect when `P4C_USE_PREINSTALLED_BASEIL` is enabled.
```
vlstill, line 198:
```suggestion
     - `-DENABLE_ABSEIL_STATIC=ON|OFF`. Enable the use of static abseil libraries. Default is ON. Only has an effect when `P4C_USE_PREINSTALLED_ABSEIL` is enabled.

**Review by vlstill — COMMENTED** [review]
Is Abseil used only as dependency for protobuf, or is it used by something else too? Was this dependency added lately?

**Comment by asl** [comment]
> Is Abseil used only as dependency for protobuf, or is it used by something else too? Was this dependency added lately?

Abseil is a hard dependency of protobuf > 3.21. There are some decent data structured that we can use from there (see https://github.com/p4lang/p4c/pull/4459#issuecomment-1961692760 as an example) as well, so it could be win-win scenario.

**Comment by ChrisDodd** [comment]
This PR seems to break on MacOS for me (errors in the google/protobuf code).  Is there some other new dependency needed to make this version of protobuf build?

**Comment by asl** [comment]
@ChrisDodd It works for me on my Mac laptop (and on CI as well). What are the errors? Are you using system-installed protobuf?

**Comment by fruffy** [comment]
> This PR seems to break on MacOS for me (errors in the google/protobuf code). Is there some other new dependency needed to make this version of protobuf build?

What are the exact errors? Maybe nuking the build folder helps? We have been testing these changes on Sonoma and Ventura.

**Comment by ChrisDodd** [comment]
It gives errors like

```
[ 60%] Building CXX object _deps/protobuf-build/CMakeFiles/libprotoc.dir/src/google/protobuf/compiler/code_generator.cc.o
In file included from /Users/cdodd/p4org-p4c-alt/build/_deps/protobuf-src/src/google/protobuf/compiler/code_generator.cc:24:
/Users/cdodd/p4org-p4c-alt/build/_deps/protobuf-src/src/google/protobuf/feature_resolver.h:41:25: error: use of undeclared identifier 'FeatureSetDefaults'
  static absl::StatusOr<FeatureSetDefaults> CompileDefaults(
                        ^
/Users/cdodd/p4org-p4c-alt/build/_deps/protobuf-src/src/google/protobuf/feature_resolver.h:44:7: error: unknown type name 'Edition'
      Edition minimum_edition, Edition maximum_edition);
```

with a fresh build folder.  The problem seems to be this `FeatureSetDefaults` type and related things which are used in the protobuf code, but are never defined anywhere?  So I would guess there much be some config header file which is supposed to be generated that contains it, but something goes awry with that.  The build on Ubuntu works fine even though it similarly has no definition for this anywhere -- I'm not sure how.

ChrisDodd:
Yes, looks like the build is picking up old protobuf headers before its own headers -- the `-I` options on the build command line are in the wrong order.  Not sure how to best fix that, but can at least avoid it by removing the installed older protobuf headers.

asl:
Interesting, it should not be like this.... Could you post what's on cmdline so we can check / fix. I was planning to do some cmake cleaning one day...

fruffy:
> Yes, looks like the build is picking up old protobuf headers before its own headers -- the `-I` options on the build command line are in the wrong order. Not sure how to best fix that, but can at least avoid it by removing the installed older protobuf headers.


So just to clarify, this happens only in the protobuf code but not ours, right? 

Protobuf pulling in the wrong includes is a common issue and I do not know a good way to fix this for the entire compiler. 
https://github.com/p4lang/p4c/issues/4477 is related. We may be able to clean this up now that we are mandating a newer Protobuf version without broken CMake support.

PR #4465, opened 2024-02-25
Add some more global operator new / delete overrides

asl:
- Aligned new / delete
 - Nothrow variants

Note that the former is required if e.g. LLVM is used with P4C (for backend).

Some cleanups here as well

### DIFF
--- a/lib/gc.cpp
+++ b/lib/gc.cpp
@@ -81,6 +81,22 @@ void *operator new(std::size_t size) {
     return rv;
 }
 
+void *operator new(std::size_t size, const std::nothrow_t &) noexcept {
+    TRACE_ALLOC(size)
+
+    maybe_initialize_gc();
+    // FIXME: Call nothrow operator new from libgc with suitable new libgc
+    // versions
+    auto *rv = ::operator new(size, UseGC, 0, 0);
+    if (!rv && emergency_ptr && emergency_ptr + size < emergency_pool + sizeof(emergency_pool)) {
+        rv = emergency_ptr;
+        size += -size & 0xf;  // align to 16 bytes
+        emergency_ptr += size;
+    }
+    if (!rv && !emergency_ptr) emergency_ptr = emergency_pool;
+    return rv;
+}
+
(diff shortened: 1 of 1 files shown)


fruffy (commented):
Tentative approval, will run P4Testgen with this and check whether I see any difference in memory usage.

fruffy:
> > Tentative approval, will run P4Testgen with this and check whether I see any difference in memory usage.
> 
> It shouldn't be. It's just a correctness thing.

Yes, not expecting any differences, but it is sensitive to gc change since it is one of the few tools using the garbage collector over longer stretches of time.

asl:
> > > Tentative approval, will run P4Testgen with this and check whether I see any difference in memory usage.
> > 
> > 
> > It shouldn't be. It's just a correctness thing.
> 
> Yes, not expecting any differences, but it is sensitive to gc change since it is one of the few tools using the garbage collector over longer stretches of time.

Right. You can just throw `abort()` there to see if there any use of these. Or maybe just `printf` :)

lib/gc.cpp
```diff
@@ -86,6 +81,51 @@ void *operator new(std::size_t size) {
...
+void *operator new(std::size_t size, const std::nothrow_t &) noexcept {
+    TRACE_ALLOC(size)
+
+    maybe_initialize_gc();
+    // FIXME: Call nothrow operator new from libgc with suitable new libgc
+    // versions
+    auto *rv = ::operator new(size, UseGC, 0, 0);
+    if (!rv && emergency_ptr && emergency_ptr + size < emergency_pool + sizeof(emergency_pool)) {
+        rv = emergency_ptr;
+        size += -size & 0xf;  // align to 16 bytes
```
fruffy, line 93: Not your code but this particular piece looks a little funky to me.  I had to pause a bit to think what negating an unsigned means and what this is trying to do. Testing for the right behavior in case of a bad alloc seems tricky.

asl: You'd just need to get used to calculate masks on-fly :) I changed this thing, is it more clear now?

PR #4470, opened 2024-02-26
Try to speed up the CI build process

### LINKED ISSUE #4469: `test-p4c / test-ubuntu20 (Unity OFF)` is very slow
Unclear what exactly the cause is but this particular build is slowing down merging and all other CI runs. It would be nice to speed it up. Part of the problem is that the free Github runners only parallelize to 2 cores, which can become unpleasantly slow for non-unified builds.

fruffy:
Runners have 4 cores now: Add ```-j$((`nproc`+1))``` where possible. 

Fixes #4469.

### DIFF
--- a/tools/ci-build.sh
+++ b/tools/ci-build.sh
@@ -252,8 +252,8 @@ cmake ${CMAKE_FLAGS} ..
 
 # If CMAKE_ONLY is active, only run CMake. Do not build.
 if [ "$CMAKE_ONLY" == "OFF" ]; then
-  make
-  sudo make install
+  make -j$((`nproc`+1))
+  sudo make -j$((`nproc`+1)) install
   # Print ccache statistics after building
   ccache -p -s
 fi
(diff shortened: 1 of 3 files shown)


tools/ci-build.sh
```diff
@@ -252,8 +252,8 @@ cmake ${CMAKE_FLAGS} ..
 
 # If CMAKE_ONLY is active, only run CMake. Do not build.
 if [ "$CMAKE_ONLY" == "OFF" ]; then
-  make
-  sudo make install
+  make -j$((`nproc`+1))
```
asl, line 255: Why don't use ninja (via `-G Ninja`)? And you can just do `cmake --build <dir>` so it could invoke whatever tool was used to generate build files. Ninja scales much better and multi-threaded by default.

asl: I think I could say that I never used `make` with p4c. Over the last ~5 years or so :) But YMMV

asl: You can do same for `cmake --install` as well, btw.

asl: Yes, absolutely! Yeah, I might not have all backends enabled, so this might be  an issue.

fruffy: #4471

asl (approved):

jafingerhut (approved):
LGTM

PR #4473, opened 2024-02-27
Use abseil maps even more

asl:
This is PR on top of https://github.com/p4lang/p4c/pull/4459 to show how we can improve the things even more with better data structures.

Essentially, this uses more efficient data structures on some hot code paths:
 - abseil hash maps / sets in reference resolver and type map
 - abseil maps / sets & pre-allocated vector in use-def analisys
 - remove some useless allocations and memory traffic in use def

Here are results:
|:---|---:|---:|---:|---:|
| `gtestp4c-abseil --gtest_filter=P4CParserUnroll.switch_20160512` | 6.056 ± 0.232 | 5.773 | 6.465 | 1.29 ± 0.05 |
| `gtestp4c --gtest_filter=P4CParserUnroll.switch_20160512` | 4.701 ± 0.061 | 4.600 | 4.783 | 1.00 |

`gtestp4c-abseil` is https://github.com/p4lang/p4c/pull/4459 (so yes 78% faster a total compared to `main`).

### DIFF
--- a/BUILD.bazel
+++ b/BUILD.bazel
@@ -239,6 +239,8 @@ cc_library(
         "@com_github_p4lang_p4runtime//:p4runtime_cc_proto",
         "@com_github_p4lang_p4runtime//:p4types_cc_proto",
         "@com_google_absl//absl/container:flat_hash_map",
+        "@com_google_absl//absl/container:flat_hash_set",
+        "@com_google_absl//absl/container:inlined_vector",
         "@com_google_protobuf//:protobuf",
     ],
 )
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -203,15 +203,17 @@ endif()
 # we require -pthread to make std::call_once work, even if we're not using threads...
 set(THREADS_PREFER_PTHREAD_FLAG ON)
 find_package(Threads REQUIRED)
-set (P4C_LIB_DEPS "${P4C_LIB_DEPS};${CMAKE_THREAD_LIBS_INIT}")
+list (APPEND P4C_LIB_DEPS ${CMAKE_THREAD_LIBS_INIT})
 include_directories(SYSTEM ${Boost_INCLUDE_DIRS})
 include_directories(SYSTEM ${PROTOBUF_INCLUDE_DIRS})
 include_directories(SYSTEM ${LIBGC_INCLUDE_DIR})
 set (HAVE_LIBBOOST_IOSTREAMS 1)
-set (P4C_LIB_DEPS "${P4C_LIB_DEPS};${Boost_LIBRARIES}")
+list (APPEND P4C_LIB_DEPS ${Boost_LIBRARIES})
 if (ENABLE_GC)
-  set (P4C_LIB_DEPS "${P4C_LIB_DEPS};${LIBGC_LIBRARIES}")
+  list (APPEND P4C_LIB_DEPS ${LIBGC_LIBRARIES})
 endif ()
+set (P4C_ABSL_LIBRARIES absl::flat_hash_map absl::flat_hash_set)
+list (APPEND P4C_LIB_DEPS ${P4C_ABSL_LIBRARIES})
 
 # other required libraries
 p4c_add_library (rt clock_gettime HAVE_CLOCK_GETTIME)
@@ -507,6 +509,7 @@ add_custom_target(genIR DEPENDS ${IR_GENERATED_SRCS})
 set_source_files_properties(${IR_GENERATOR} PROPERTIES GENERATED TRUE)
 add_library(ir-generated OBJECT ${IR_GENERATED_SRCS} ${EXTENSION_IR_SOURCES})
 add_dependencies(ir-generated ir genIR)
+target_link_libraries(ir-generated ${P4C_LIB_DEPS})
 
 ######################################## IR Generation End ########################################
 
--- a/backends/p4tools/common/CMakeLists.txt
+++ b/backends/p4tools/common/CMakeLists.txt
@@ -41,6 +41,8 @@ add_p4tools_library(p4tools-common ${P4C_TOOLS_COMMON_SOURCES})
 target_link_libraries(
   p4tools-common
   PUBLIC ${P4TOOLS_Z3_LIB}
+  PRIVATE frontend
+  PRIVATE ir-generated
 )
 
 target_include_directories(
@@ -54,9 +56,5 @@ target_include_directories(
   PUBLIC "${P4C_BINARY_DIR}"
 )
 
-add_dependencies(p4tools-common ir-generated frontend)
-
 # Add control-plane-specific extensions.
 add_subdirectory(${CMAKE_CURRENT_LIST_DIR}/control_plane)
-
-
--- a/backends/p4tools/common/CMakeLists.txt
+++ b/backends/p4tools/common/CMakeLists.txt
@@ -42,7 +42,6 @@ target_link_libraries(
   p4tools-common
   PUBLIC ${P4TOOLS_Z3_LIB}
   PRIVATE frontend
-  PRIVATE ir-generated
 )
 
 target_include_directories(
--- a/frontends/common/resolveReferences/referenceMap.cpp
+++ b/frontends/common/resolveReferences/referenceMap.cpp
@@ -43,8 +43,7 @@ void ReferenceMap::setDeclaration(const IR::Path *path, const IR::IDeclaration *
     CHECK_NULL(path);
     CHECK_NULL(decl);
     LOG3("Resolved " << dbp(path) << " to " << dbp(decl));
-    auto it = pathToDeclaration.find(path);
-    auto previous = it != pathToDeclaration.end() ? it->second : nullptr;
+    const auto *previous = get(pathToDeclaration, path);
     if (previous != nullptr && previous != decl)
         BUG("%1% already resolved to %2% instead of %3%", dbp(path), dbp(previous),
             dbp(decl->getNode()));
@@ -57,8 +56,7 @@ void ReferenceMap::setDeclaration(const IR::This *pointer, const IR::IDeclaratio
     CHECK_NULL(pointer);
     CHECK_NULL(decl);
     LOG3("Resolved " << dbp(pointer) << " to " << dbp(decl));
-    auto it = thisToDeclaration.find(pointer);
-    auto previous = it != thisToDeclaration.end() ? it->second : nullptr;
+    const auto *previous = get(thisToDeclaration, pointer);
     if (previous != nullptr && previous != decl)
         BUG("%1% already resolved to %2% instead of %3%", dbp(pointer), dbp(previous), dbp(decl));
     thisToDeclaration.emplace(pointer, decl);
@@ -66,8 +64,7 @@ void ReferenceMap::setDeclaration(const IR::This *pointer, const IR::IDeclaratio
 
 const IR::IDeclaration *ReferenceMap::getDeclaration(const IR::This *pointer, bool notNull) const {
     CHECK_NULL(pointer);
-    auto it = thisToDeclaration.find(pointer);
-    auto result = it != thisToDeclaration.end() ? it->second : nullptr;
+    const auto *result = get(thisToDeclaration, pointer);
 
     if (result)
         LOG3("Looking up " << dbp(pointer) << " found " << dbp(result));
@@ -80,8 +77,7 @@ const IR::IDeclaration *ReferenceMap::getDeclaration(const IR::This *pointer, bo
 
 const IR::IDeclaration *ReferenceMap::getDeclaration(const IR::Path *path, bool notNull) const {
     CHECK_NULL(path);
-    auto it = pathToDeclaration.find(path);
-    auto result = it != pathToDeclaration.end() ? it->second : nullptr;
+    const auto *result = get(pathToDeclaration, path);
 
     if (result)
         LOG3("Looking up " << dbp(path) << " found " << dbp(result));
--- a/frontends/common/resolveReferences/referenceMap.cpp
+++ b/frontends/common/resolveReferences/referenceMap.cpp
@@ -35,7 +35,7 @@ void ReferenceMap::clear() {
     program = nullptr;
     used.clear();
     thisToDeclaration.clear();
-    for (auto &reserved : P4::reservedWords) usedNames.insert({reserved, 0});
+    for (auto &reserved : P4::reservedWords) usedNames.emplace(reserved, 0);
     ProgramMap::clear();
 }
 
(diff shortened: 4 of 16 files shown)


asl:
I think some other benchmarks are also required. Unfortunately, we do not have much that could be ported from downstream to "stock" p4c.

frontends/common/resolveReferences/referenceMap.h
```diff
@@ -67,17 +68,17 @@ class ReferenceMap final : public ProgramMap, public NameGenerator, public Decla
     bool isv1;
 
     /// Maps paths in the program to declarations.
-    ordered_map<const IR::Path *, const IR::IDeclaration *> pathToDeclaration;
+    absl::flat_hash_map<const IR::Path *, const IR::IDeclaration *, Util::Hash> pathToDeclaration;
```
vlstill, line 71: Note: In general we should be quite careful when replacing `ordered_map` with something that does not keep order as `ordered_map` could be used to make sure the compiler is deterministic (and nondeterminism bugs are extremely hard to pinpoint :-/).

In this particular case though, I don't see any risk of nondeterminism as nothing iterates over the containers (expect possibly the debug print).

frontends/common/resolveReferences/referenceMap.cpp
```diff
@@ -77,7 +80,8 @@ const IR::IDeclaration *ReferenceMap::getDeclaration(const IR::This *pointer, bo
 
 const IR::IDeclaration *ReferenceMap::getDeclaration(const IR::Path *path, bool notNull) const {
     CHECK_NULL(path);
-    auto result = get(pathToDeclaration, path);
+    auto it = pathToDeclaration.find(path);
+    auto result = it != pathToDeclaration.end() ? it->second : nullptr;
```
vlstill, line 84: We could and probably should add a `get` overload for the abseil maps to `lib/map.h`. These can be quite useful and especially for maps where the value type is a pointer but the map should never contain `nullptr` they make it much easier to use.

frontends/p4/def_use.h
```diff
@@ -129,7 +130,7 @@ class StructLocation : public WithFieldsLocation {
 /// Interface for locations that support an index operation
 class IndexedLocation : public StorageLocation {
  protected:
-    std::vector<const StorageLocation *> elements;
+    absl::InlinedVector<const StorageLocation *, 8> elements;
```
vlstill, line 133: nice, a vector with small-vector optimization will be useful to have. Any reason for the 8 elements in particular?

frontends/p4/def_use.h
```diff
@@ -332,16 +336,24 @@ struct hash<P4::ProgramPoint> {
...
+template <>
+struct Hasher<P4::ProgramPoint> {
+    size_t operator()(const P4::ProgramPoint &p) const { return p.hash(); }
+};
+}  // namespace Util
+
 namespace P4 {
 class ProgramPoints : public IHasDbPrint {
-    typedef std::unordered_set<ProgramPoint> Points;
+    typedef absl::flat_hash_set<ProgramPoint, Util::Hash> Points;
```
vlstill, line 348: If we want to do this in most places where we can, it might be worth defining a using in `lib/map.h`, something like `template<typename T> using FlatHashSet = absl::flat_hash_set<T, Util::Hash>`.

frontends/p4/simplifyDefUse.cpp
```diff
@@ -30,9 +31,10 @@ namespace {
 class HasUses {
     // Set of program points whose left-hand sides are used elsewhere
     // in the program together with their use count
-    std::set<const IR::Node *> used;
+    absl::flat_hash_set<const IR::Node *> used;
```
vlstill, line 34: Any reason to not use `Util::Hash`?

frontends/p4/simplifyDefUse.cpp
```diff
@@ -30,9 +31,10 @@ namespace {
...
 
     class SliceTracker {
+        // FIXME: Can squeeze this bool into lower bit of pointer
```
vlstill, line 37: Personally, I wouldn't go there unless we have a very good reason to. I do realize the `bool` makes the struct 8 bits larger, but unless we know that significantly limits us I think the cost of tainting a pointer is too much.

vlstill (commented):
I went over the C++ changes. This definitely goes in the right direction, thank you. But I have some points that I believe needs to be addressed. You seem to be using `Util::Hash` only sometimes, and it is not clear to me if this is deliberate, or if it is an omission. Also, I suggest we make the `get` availble for the abseil maps to simplify both migration and the code itself.

frontends/common/resolveReferences/referenceMap.h
asl: I double checked all `ordered_map` differences when made this changes. In many cases there is no iteration at all besides the debug priting. So, just having stable debug output does not justify the generic slowdowns.

frontends/p4/def_use.h
asl: For now it was more or less wild guess just to have the same memory footprint for the object itself.

frontends/p4/def_use.h
asl: This would require some massive changes to all dependencies and increased include size. So I'd rather keep it fine-grained for now. Otherwise we may end with the same proliferation as we have now with `ordered_map` (in many cases w/o particular reason)

frontends/p4/simplifyDefUse.cpp
asl: No. Just missed here

frontends/p4/simplifyDefUse.cpp
asl: Yes, this is why it's just some observation. I do not know the particular impact here.

frontends/common/resolveReferences/referenceMap.cpp
asl: Are we ok to have all abseil maps and corresponding includes to be exposed via `lib/map.h`?

frontends/p4/def_use.h
asl: All right, here is the distribution of sizes on switch testcase:
```
     0      1      2      3      4
387633 321734 174704  94070  11821
```

I will check downstream large app as well

asl: One downstream app:
```
     0      1      2      3      4     5     6     7     8     9
8932648 6993790 12263 1793251 1684828 1295764 1105095 1213573 394486 34967 
```

asl: And sorry, the sizes are for `ProgramPoints` below, not `IndexedLocation`.

frontends/p4/def_use.h
```diff
@@ -278,14 +279,14 @@ class ProgramPoint : public IHasDbPrint {
     /// the previous context.  E.g., a stack [Function] is the context before
     /// the function, while [Function, nullptr] is the context after the
     /// function terminates.
-    std::vector<const IR::Node *> stack;
+    absl::InlinedVector<const IR::Node *, 8> stack;  // Has inline space for 8 nodes
```
asl, line 282: See sizes above ^^^

BUILD.bazel
```diff
@@ -43,7 +43,10 @@ cc_library(
 cc_library(
     name = "lib",
     srcs = glob(["lib/*.cpp"]),
-    hdrs = glob(["lib/*.h"]),
+    hdrs = glob([
+        "lib/*.h",
+        "lib/parallel_hashmap/*.h",
```
fruffy, line 48: I believe this can be removed?

CMakeLists.txt
```diff
@@ -507,6 +509,7 @@ add_custom_target(genIR DEPENDS ${IR_GENERATED_SRCS})
 set_source_files_properties(${IR_GENERATOR} PROPERTIES GENERATED TRUE)
 add_library(ir-generated OBJECT ${IR_GENERATED_SRCS} ${EXTENSION_IR_SOURCES})
 add_dependencies(ir-generated ir genIR)
+target_link_libraries(ir-generated ${P4C_LIB_DEPS})
```
fruffy, line 512: Hmm, ir-generated is a special library that should be encapsulated by `ir.h`. There should be no include like `ir/ir-generated.h` anywhere. include-fixer sometimes includes it.

cmake/Linters.cmake
```diff
@@ -24,6 +24,7 @@ list(FILTER P4C_LINT_LIST EXCLUDE REGEX "backends/ubpf/runtime")
 list(FILTER P4C_LINT_LIST EXCLUDE REGEX "control-plane/p4runtime")
 list(FILTER P4C_LINT_LIST EXCLUDE REGEX "backends/tc/runtime")
 list(FILTER P4C_LINT_LIST EXCLUDE REGEX "test/frameworks")
+list(FILTER P4C_LINT_LIST EXCLUDE REGEX "lib/parallel_hashmap")
```
fruffy, line 27: Remove?

backends/p4tools/common/CMakeLists.txt
```diff
@@ -41,6 +41,8 @@ add_p4tools_library(p4tools-common ${P4C_TOOLS_COMMON_SOURCES})
 target_link_libraries(
   p4tools-common
   PUBLIC ${P4TOOLS_Z3_LIB}
+  PRIVATE frontend
+  PRIVATE ir-generated
```
fruffy, line 45: Hmm I am only realizing this now, but `ir-generated` should never show up outside the `ir` folder. It should be encapsulated by the `ir` library, which effectively exports all ir-related utilities. If this is necessary we might be doing something wrong. This might also explain why we sometimes get the `ir/ir-generated.h` includes. This should not happen.

cmake/Linters.cmake
asl: Oh, I thought these were removed... strange...

backends/p4tools/common/CMakeLists.txt
asl: Let me check what is going on. Here I was essentially transitioning from existing `add_dependency` to proper library

asl: This looks like a common pattern that was Ctrl+C / Ctrl-V all around the backends:
```
ubpf/CMakeLists.txt
64:add_dependencies(p4c-ubpf ir-generated frontend)

p4tools/modules/testgen/test/z3-solver/constraints.cpp
8:#include "ir/ir-generated.h"

p4tools/modules/testgen/targets/bmv2/expr_stepper.cpp
21:#include "ir/ir-generated.h"

p4test/CMakeLists.txt
31:add_dependencies(p4test ir-generated frontend)

graphs/CMakeLists.txt
33:target_link_libraries(p4cgraphs ir-generated frontend)
36:target_link_libraries (p4c-graphs p4cgraphs ir-generated frontend ${P4C_LIBRARIES} ${P4C_LIB_DEPS})

ebpf/CMakeLists.txt
126:add_dependencies(p4c-ebpf ir-generated frontend)

bmv2/CMakeLists.txt
86:target_link_libraries(bmv2backend ir-generated frontend)
```

asl: As far as I can see, `ir-generated` is not a part of `ir`. It is setup in top-level CMakeLists:
```
# Component libraries: must be defined before being used in the back end executables.
set (P4C_LIBRARIES controlplane midend frontend ir p4ctoolkit ir-generated)
```
and more rules further on.

Probably something should be refactored. I tend to agree that these `-gen` targets should be internal to corresponding libraries.

frontends/common/resolveReferences/referenceMap.cpp
asl: https://github.com/p4lang/p4c/pull/4483 generalizes `lib/map.h` to allow different maps.

backends/p4tools/common/CMakeLists.txt
fruffy: No you are right, I misremembered. There are two parts here. `ir-generated.h` is always meant to be encapsulated by `ir.h`. It used to be the case that one caused compilation errors when including the `ir-generated` header, but I cleaned that up at some point. Now it doesn't cause issues anymore but one should still include `ir.h`. I will clean up the `ir-generated.h` includes, some of them are caused by me and clang-tidy's include fixer :disappointed:. I need to configure that one right. 

The CMake target itself is actually separate and each back end links to it separately. Not a big fan of it either, but that is partially necessary because each back end can add their own IR sources to the target, like the DPDK target does: 
https://github.com/p4lang/p4c/blob/main/backends/dpdk/CMakeLists.txt#L99

We can keep thing as is for now.

frontends/common/resolveReferences/resolveReferences.h
```diff
@@ -17,9 +17,9 @@ limitations under the License.
 #ifndef COMMON_RESOLVEREFERENCES_RESOLVEREFERENCES_H_
 #define COMMON_RESOLVEREFERENCES_RESOLVEREFERENCES_H_
 
+#include "absl/container/flat_hash_map.h"
```
fruffy, line 20: Actually, can we make absl a bracketed system include to be consistent?

asl: I think someone asked somewhere exactly the opposite for fetched dependencies :)

asl: Need to find. But I'm happy to change to angled brackets. This makes much more sense to me :)

asl: Ok, apparently abseil expects to use quoted includes. And bazel build even enforces this (the path is exported as `-iquoted`).

fruffy: Ugh, seems to be an opinionated maintainer. This might cause issues with Werror and clang-tidy. Not a big deal but annoying.

We can just patch this, I can get this done in a separate PR.

fruffy (approved):
Approving from my side.

frontends/common/resolveReferences/referenceMap.cpp
vlstill: Thank you. I think it would make sense to merge #4483 first and reverse the `get` expansions in this PR.

asl: Yes, I will refine after #4483

frontends/p4/simplifyDefUse.cpp
```diff
@@ -369,7 +371,7 @@ class FindUninitialized : public Inspector {
     ProgramPoint currentPoint;  // context of the current expression/statement
     /// For some simple expresssions keep here the read location sets.
     /// This does not include location sets read by subexpressions.
-    std::map<const IR::Expression *, const LocationSet *> readLocations;
+    absl::flat_hash_map<const IR::Expression *, const LocationSet *> readLocations;
```
vlstill, line 374: no `Util::Hash` here?

frontends/common/resolveReferences/referenceMap.cpp
asl: @vlstill This was refined, please take a look, thanks!

vlstill (approved):

Issue #4420, opened 2024-02-13
Make the control plane optional


fruffy:
The control plane pulls in a bunch of dependencies (Protobuf and soon also Abseil). Not every back end needs these dependencies. Add a CMake option to enable/disable building the control-plane. If a back end does require the control-plane make it required and throw an error if the control-plane is disabled.

PR #4474, opened 2024-02-27
Try to clean up the Protobuf includes.

fruffy:
The control plane target depends on Protobuf and Abseil. At least for Protobuf, it should export this dependency such that other downstream targets do not need to include the Protobuf include directories all the time. 

Unclear what to do about Abseil. It is more or less a core compiler dependency and libraries should be linked by each target individually. 

This is tangentially related to https://github.com/p4lang/p4c/issues/4420. Because the control plane pulls in all this stuff, it should be optional.

### DIFF
--- a/control-plane/CMakeLists.txt
+++ b/control-plane/CMakeLists.txt
@@ -79,7 +79,7 @@ set (CONTROLPLANE_SRCS
   p4RuntimeSymbolTable.cpp
   typeSpecConverter.cpp
   bfruntime.cpp
-  )
+)
 
 set (CONTROLPLANE_HDRS
   addMissingIds.h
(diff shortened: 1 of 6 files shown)


fruffy:
@asl This may conflict with the changes you made in the Abseil Map PRs.

asl:
@fruffy Yes, I saw this. Feel free to merge this first. And I will resolve the conflicts

control-plane/CMakeLists.txt
```diff
@@ -93,22 +93,28 @@ set (CONTROLPLANE_HDRS
...
   bfruntime.h
-  )
+)
 
 add_library(controlplane-gen OBJECT ${P4RUNTIME_GEN_SRCS})
-target_include_directories(controlplane-gen
-  SYSTEM BEFORE PUBLIC ${Protobuf_INCLUDE_DIRS}
+target_include_directories(
+  controlplane-gen SYSTEM BEFORE PUBLIC ${Protobuf_INCLUDE_DIRS} SYSTEM BEFORE
+  PUBLIC ${ABSL_COMMON_INCLUDE_DIRS}
```
asl, line 101: Probably worth a fixme, that if we'd include proper protobuf targets, then we can get rid of `Protobuf_INCLUDE_DIRS` as include dirs will be exposed via interface of protobuf link target

asl (approved):

PR #4480, opened 2024-02-28
Fix for pipeline id for tc backend

Sosutha:
* Removed pipeline id from template for tc backend
* Removed pipeline id from introspection.json for tc backend
* Updated the testcases outputs


usha1830 (approved):
LGTM

Issue #4481, opened 2024-02-28
Reduce proliferation of cstrings


asl:
cstrings are commonly used as "the string type" in p4c. However, they should be used with more care. The reason is that they are internalized, so each new string is stored forever in some global map.

This makes them great for various containers, both keys and values, for IR strings, etc. However, in many cases `cstring`s are used just for random transient strings. As a result:
 - Copies are created every time `cstring` is constructed
 - These copies are stuck forever

As a result, we might observe some elevated memory usage especially for long-lasting processes. I definitely saw that ~1% of runtime of benchmark from https://github.com/p4lang/p4c/pull/4475 is spent on string hash used for `cstring` internalization. I have not measured how many strings are in `cstring` hash, but I'd assume a lot.

Instead, one should resort to `std::string` and `std::string_view` for transient values that are not required to be saved forever. Also, we might consider adding some kind of `StringMap` with string keys that would own keys and not pollute the common cstring cache.

asl:
Tagging @fruffy

vlstill:
> Instead, one should resort to std::string and std::string_view for transient values that are not required to be saved forever.

I agree there 100 %. We might need to provide some utility functions to make the migration easier. Some of the uses might be there just because some things are easier with cstring then with std::string (or because we relatively recently switched do C++17).

> Also, we might consider adding some kind of StringMap with string keys that would own keys and not pollute the common cstring cache.

Could you please give a use-case for that?

asl:
> Could you please give a use-case for that?

Just loud thinking – if one would need to have internalized string keys (e.g. for faster lookup), but definitely would worth starting from the proper usecase in the code

asl:
I collected some statistics on downstream project. And the results are overwhelming. Essentially cstring cache works as sink, collecting everyone and their brother: all format strings, pieces of input, json values, ...

Essentially after frontend the cstring cache contained 300k strings (!), midend raised this up to 311k (though we do not run much midend passes). Number of map lookups was great as well totalling 86M after frontend (each cstring ctor involves at least 1 map lookup, or 2 if the value should be stored).

So, lots of hash table lookups, lots of string hashes, etc.

I will try to see if there is some low-hanging fruit here... Likely cstring should be "storage-only" type for node members. But not a replacement for generic string type...

fruffy:
>I will try to see if there is some low-hanging fruit here.. 

[...]

PR #4482, opened 2024-02-28
Replace boost string algorithms with their abseil counterparts

asl:
Some refactoring and cleanup while there:
 - Code moved from headers down to cpp
 - `StringRef` replaced with `std::string_view` (header is not removed as I do not know how this might affect downstream)

### DIFF
--- a/control-plane/p4RuntimeSymbolTable.cpp
+++ b/control-plane/p4RuntimeSymbolTable.cpp
@@ -15,8 +15,7 @@ limitations under the License.
 */
 #include "p4RuntimeSymbolTable.h"
 
-#include <boost/algorithm/string/split.hpp>
-
+#include "absl/strings/str_split.h"
 #include "lib/cstring.h"
 #include "lib/iterator_range.h"
 #include "p4RuntimeArchHandler.h"
@@ -292,9 +291,8 @@ uint32_t P4::ControlPlaneAPI::P4RuntimeSymbolTable::jenkinsOneAtATimeHash(const 
 
 cstring P4::ControlPlaneAPI::P4SymbolSuffixSet::shortestUniqueSuffix(const cstring &symbol) const {
     BUG_CHECK(!symbol.isNullOrEmpty(), "Null or empty symbol name?");
-    std::vector<cstring> components;
     const char *cSymbol = symbol.c_str();
-    boost::split(components, cSymbol, [](char c) { return c == '.'; });
+    std::vector<cstring> components = absl::StrSplit(cSymbol, '.');
 
     // Determine how many suffix components we need to uniquely identify
     // this symbol. For example, if we have the symbols "d.a.c" and "e.b.c",
@@ -303,11 +301,10 @@ cstring P4::ControlPlaneAPI::P4SymbolSuffixSet::shortestUniqueSuffix(const cstri
     unsigned neededComponents = 0;
     auto *node = suffixesRoot;
     for (auto &component : Util::iterator_range(components).reverse()) {
-        if (node->edges.find(component) == node->edges.end()) {
-            BUG("Symbol is not in suffix set: %1%", symbol);
-        }
+        auto it = node->edges.find(component);
+        BUG_CHECK(it != node->edges.end(), "Symbol is not in suffix set: %1%", symbol);
 
-        node = node->edges[component];
+        node = it->second;
         neededComponents++;
 
         // If there's only one suffix that passes through this node, we have
@@ -323,7 +320,7 @@ cstring P4::ControlPlaneAPI::P4SymbolSuffixSet::shortestUniqueSuffix(const cstri
     BUG_CHECK(neededComponents <= components.size(), "Too many components?");
     std::string uniqueSuffix;
     std::for_each(components.end() - neededComponents, components.end(),
-                  [&](const cstring &component) {
+                  [&](const auto &component) {
                       if (!uniqueSuffix.empty()) {
                           uniqueSuffix.append(".");
                       }
@@ -346,16 +343,15 @@ void P4::ControlPlaneAPI::P4SymbolSuffixSet::addSymbol(const cstring &symbol) {
     // strictly and have tests for them, it's safest to ensure this
     // precondition here.
     {
-        auto result = symbols.insert(symbol);
-        if (!result.second) {
+        auto [_, inserted] = symbols.insert(symbol);
+        if (!inserted) {
             return;  // It was already present.
         }
     }
 
     // Split the symbol name into dot-separated components.
-    std::vector<cstring> components;
     const char *cSymbol = symbol.c_str();
-    boost::split(components, cSymbol, [](char c) { return c == '.'; });
+    std::vector<cstring> components = absl::StrSplit(cSymbol, '.');
 
     // Insert the components into our tree of suffixes. We work
     // right-to-left through the symbol name, since we're concerned with
@@ -367,10 +363,9 @@ void P4::ControlPlaneAPI::P4SymbolSuffixSet::addSymbol(const cstring &symbol) {
     // (Nodes are in parentheses, and edge labels are in quotes.)
     auto *node = suffixesRoot;
     for (auto &component : Util::iterator_range(components).reverse()) {
-        if (node->edges.find(component) == node->edges.end()) {
-            node->edges[component] = new SuffixNode;
-        }
-        node = node->edges[component];
+        auto [it, inserted] = node->edges.emplace(component, nullptr);
+        if (inserted) it->second = new SuffixNode;
+        node = it->second;
         node->instances++;
     }
 }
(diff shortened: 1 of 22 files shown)


control-plane/p4RuntimeSymbolTable.cpp
```diff
@@ -254,7 +254,7 @@ void P4::ControlPlaneAPI::P4RuntimeSymbolTable::computeIdsForSymbols(P4RuntimeSy
     for (const auto &mapping : nameToIteratorMap) {
         const cstring name = mapping.first;
         const auto iterator = mapping.second;
-        const uint32_t nameId = jenkinsOneAtATimeHash(name.c_str(), name.size());
+        const uint32_t nameId = Util::Hash{}(name);
```
fruffy, line 257: I think it is a good idea to clean this up but a lot of infrastructure relies on the computation of the P4Runtime IDs. The test failures stem from this change because the reference files have changed. I am not sure whether we can update this hash. If we do, we should do it in a separate PR. 

@smolkaj @jonathan-dilorenzo

asl: Let me move this to a separate PR

vlstill (commented):
Thank you for cleaning away the `StringRef`. Could you also put `[[deprecated(...)]]` on the `StringRef` struct so the downstream tools are kind reminded to upgrade and we have a path to drop it completely some time in future? I think that would make sense :-).

On a more general note: is there a strong reason to replace boost algos with their abseil counterparts? It abseil significantly better? Is it easier to use? Do we strive to get rid of boost altogether? Since we have abseil for other reasons, I am not against this, but for example in case of the substitutions the boost syntax had some advantages (named paceholders).

asl:
> On a more general note: is there a strong reason to replace boost algos with their abseil counterparts? It abseil significantly better? Is it easier to use? Do we strive to get rid of boost altogether? Since we have abseil for other reasons, I am not against this, but for example in case of the substitutions the boost syntax had some advantages (named paceholders).

I like abseil string algorithms more. For example `StrSplit` could return a vector of `string_view`'s, essentially skipping the copies of sting pieces. It also returns vector by value, `boost::split` takes output by reference.

lib/cstring.h
```diff
@@ -103,6 +103,12 @@ class cstring {
         construct_from_shared(string.data(), string.length());
     }
 
+    // construct cstring from std::string_view. Do not use if possible, this is linear
+    // time operation if string not exists in table, because the underlying string must be copied.
+    cstring(std::string_view string) {  // NOLINT(runtime/explicit)
```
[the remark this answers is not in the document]
asl: abseil string algorithms are quite centered around `string_views` with automatic conversions. So, this constructor allows us to write, e.g.:
```
    std::vector<cstring> components = absl::StrSplit(cSymbol, '.');
```

See "adaptation to return types" at https://abseil.io/docs/cpp/guides/strings#abslstrsplit-for-splitting-strings

Need to check if it allows explicit conversion.

fruffy:
> On a more general note: is there a strong reason to replace boost algos with their abseil counterparts? It abseil significantly better? Is it easier to use? Do we strive to get rid of boost altogether? Since we have abseil for other reasons, I am not against this, but for example in case of the substitutions the boost syntax had some advantages (named paceholders).

My personal opinion is that it is nice to keep our dependencies simple. Since we are basically forced to use Abseil with Protobuf we might as well make use of it. Since we do not want to keep two utility frameworks around we should slowly deprecated widespread usage of boost. Another reason I am in favor of slowly phasing out boost is that is generally a huge framework and unlike Abseil including one boost header can pull in many others. We do not need to get rid of it entirely but should be judicious of what we include. https://github.com/p4lang/p4c/issues/3898 is an attempt to simplify things a little.

asl:
> My personal opinion is that it is nice to keep our dependencies simple. Since we are basically forced to use Abseil with Protobuf we might as well make use of it. Since we do not want to keep two utility frameworks around we should slowly deprecated widespread usage of boost. Another reason I am in favor of slowly phasing out boost is that is generally a huge framework and unlike Abseil including one boost header can pull in many others. We do not need to get rid of it entirely but should be judicious of what we include. #3898 is an attempt to simplify things a little.

I think after this PR we'll end with:
 - boost::graph for graph backend
 - boost::multiprecision
 - boost::random
 - boost::format
 - few containers here and there, that could be replaced with some header-only implementation dropped to `lib`.

The latter could be replaced by `fmt` or Abseil's format routines, or we can wait for next standard bump. It's just huge pile of work :)

boost::random is used for Mersenne-Twister – need to double check if these days normal standard library is mature enough to instantiate rng with bigint type.

vlstill:
> boost::random is used for Mersenne-Twister – need to double check if these days normal standard library is mature enough to instantiate rng with bigint type.

It seems it is not :-/

> UIntType -- The result type generated by the generator. The effect is undefined if this is not one of unsigned short, unsigned int, unsigned long, or unsigned long long. (https://en.cppreference.com/w/cpp/numeric/random/mersenne_twister_engine)

vlstill (approved):
This looks good to me.

asl:
@fruffy Will you please take a look when you have a chance? Thanks!

fruffy:
> @fruffy Will you please take a look when you have a chance? Thanks!

Will try to give a look tomorrow, things have been a bit busy lately..

asl:
> Will try to give a look tomorrow, things have been a bit busy lately..

Sure, no worries! Just wanted to clear my patch queue a bit :)

frontends/p4/typeChecking/typeConstraints.cpp
```diff
@@ -111,4 +114,69 @@ void TypeConstraints::dbprint(std::ostream &out) const {
     }
 }
 
+std::string TypeConstraint::localError(Explain *explainer) const {
+    if (errFormat.isNullOrEmpty()) return "";
+
+    std::string message, explanation;
+    boost::format fmt = boost::format(errFormat);
```
fruffy, line 121: I am guessing we can not get rid of this particular format call because the input is boost-formatted?

frontends/parsers/v1/v1lexer.ll
```diff
@@ -69,7 +68,7 @@ using Parser = V1::V1Parser;
 
 "@pragma"[ \t]*[A-Za-z_][A-Za-z0-9_]* {
                   BEGIN((driver.saveState = PRAGMA_LINE));
-                  return Parser::make_PRAGMA(StringRef(yytext+7).trim(), driver.yylloc); }
+                  return Parser::make_PRAGMA(cstring(V1Lexer::trim(yytext+7)), driver.yylloc); }
```
fruffy, line 71: I am guessing this is needed because you are converting a `string_view` to cstring?Unable to check the interface here right now, maybe the input to `make_PRAGMA` could be string_view itself or just a char pointer?

lib/source_file.h
```diff
@@ -319,6 +319,7 @@ class InputSources final {
 
 }  // namespace Util
 
+// FIXME: Do we really need to have this in header? This pulls the whole iostream header
```
fruffy, line 322: Are things breaking if you move it out of the header?

lib/cstring.h
```diff
@@ -103,6 +103,12 @@ class cstring {
         construct_from_shared(string.data(), string.length());
     }
 
+    // construct cstring from std::string_view. Do not use if possible, this is linear
```
fruffy, line 106: Our comment style really is all over the place. The one for this file is a lowercase double slashes one. One day I will set up doxygen to warn about inconsistencies.

fruffy (approved):

frontends/p4/typeChecking/typeConstraints.cpp
asl: Let me check

frontends/parsers/v1/v1lexer.ll
asl: Yes. I've made constructor explicit, so, explicit conversion is needed. `make_PRAGMA` is autogenerated, so changing its interface would require some decent amount of changes here and there. So, maybe next time as a part of https://github.com/p4lang/p4c/issues/4481

lib/cstring.h
asl: yes :(

frontends/p4/typeChecking/typeConstraints.cpp
asl: So, yes, the input is just string, so we need to wrap, etc :(

However, I looked around and did some light refactoring removing unnecessary `cstring`'s.

lib/source_file.cpp
```diff
@@ -311,3 +311,7 @@ cstring SourceFileLine::toString() const {
 }
 
 }  // namespace Util
+
+////////////////////////////////////////////////////////
+
+[[gnu::used]] void dbprint(const IHasDbPrint *o) { o->dbprint(std::cout); }
```
vlstill, line 317: Could you please add a short comment that the `gnu::used` is required to make sure linker does not omit it?

vlstill (approved):

frontends/p4/typeChecking/typeConstraints.cpp
fruffy: Thanks, looks good to me!

PR #4483, opened 2024-02-28
Generalize get in map utils

asl:


lib/map.h
```diff
@@ -17,42 +17,66 @@ limitations under the License.
...
-    if (it != m.end()) return it->second;
-    return def;
+    return it != m.end() ? it->second : typename Map::mapped_type{};
+}
+template <class Map, typename Key = typename Map::key_type,
+          typename Value = typename Map::mapped_type>
+typename Map::mapped_type get(const Map &m, const Key &key, Value &&def) {
+    using M = typename Map::mapped_type;
+    auto it = m.find(key);
+    return it != m.end() ? it->second : static_cast<M>(static_cast<Value &&>(def));
```
vlstill, line 34: std::move instead of the inner static cast?

lib/map.h
```diff
@@ -17,42 +17,66 @@ limitations under the License.
...
+/// Given a map and a key, return the value corresponding to the key in the map,
+/// or a given default value if the key doesn't exist in the map.
+template <typename Map, typename Key>
+typename Map::mapped_type get(const Map &m, const Key &key) {
     auto it = m.find(key);
-    if (it != m.end()) return it->second;
-    return def;
+    return it != m.end() ? it->second : typename Map::mapped_type{};
+}
+template <class Map, typename Key = typename Map::key_type,
```
vlstill, line 29: Why has the Key default type here, but not in the previous overload?

lib/map.h
```diff
@@ -17,42 +17,66 @@ limitations under the License.
...
-    return 0;
+    return it != m.end() ? &it->second : nullptr;
 }
 
-template <class K, class T, class V, class Comp, class Alloc>
-inline const V *getref(const std::map<K, V, Comp, Alloc> &m, T key) {
+// Given a map and a key, return a optional<V> if the key exists and None if the
+// key does not exist in the map.
+template <class Map, typename Key = typename Map::key_type>
+std::optional<typename Map::mapped_type> get_optional(const Map &m, const Key &key) {
```
vlstill, line 54: `getopt` would be more consistent with `getref`. I'm not sure myself which one is better, we could go with the more explicit name you have chosen.

lib/map.h
```diff
@@ -17,42 +17,66 @@ limitations under the License.
...
+template <typename Map, typename Key>
+typename Map::mapped_type get(const Map &m, const Key &key) {
     auto it = m.find(key);
-    if (it != m.end()) return it->second;
-    return def;
+    return it != m.end() ? it->second : typename Map::mapped_type{};
+}
+template <class Map, typename Key = typename Map::key_type,
+          typename Value = typename Map::mapped_type>
+typename Map::mapped_type get(const Map &m, const Key &key, Value &&def) {
```
vlstill, line 31: I think we could use a version that takes `const Value &def` for cases where the default is an lvalue and we want it copied when used.

vlstill (commented):
Thank you. Just a few things:

lib/map.h
asl: Actually an oversight. I was playing with heterogenous lookup and this is just a leftover. Can be removed, we have `mapped_type` to distinguish map types from everything else.

lib/map.h
asl: I though about this. But `getopt` in a global scope is already taken by...getopt :) https://man7.org/linux/man-pages/man3/getopt.3.html So, less name clashes. It has different signature, but since it's a C function I'm afraid of macroses. We can give `getopt` a try and see how it will go.

lib/map.h
asl: Oh, no. Then lvalues will be moved. I would change it to `std::forward`, it will be more clear that we're doing a "universal reference" stuff here.

lib/map.h
asl: See above. I believe the universal reference version with `std::forward` should take care of this.

lib/map.h
vlstill: Oh sorry about that, I did not realize it is universal reference.

asl: This part is always confusing...

lib/map.h
vlstill: I see. Now I'm slightly leaning towards `get_optional` to avoid possible confusion. The error messages could get weird with `getopt` too.

lib/map.h
```diff
@@ -17,42 +17,65 @@ limitations under the License.
...
-    if (it != m.end()) return &it->second;
-    return 0;
+    return it != m.end() ? &it->second : nullptr;
 }
 
-template <class K, class T, class V, class Comp, class Alloc>
-inline const V *getref(const std::map<K, V, Comp, Alloc> &m, T key) {
+// Given a map and a key, return a optional<V> if the key exists and None if the
+// key does not exist in the map.
+template <class Map, typename Key = typename Map::key_type>
```
vlstill, line 52:
```suggestion
template <class Map, typename Key>

File: lib/map.h [hunk 6]
```diff
@@ -17,42 +17,65 @@ limitations under the License.
...
+
+template <typename Map, typename Key>
+typename Map::mapped_type get(const Map *m, const Key &key) {
+    return m ? get(*m, key) : typename Map::mapped_type{};
 }
 
-template <class K, class T, class V, class Comp, class Alloc>
-inline V get(const std::map<K, V, Comp, Alloc> *m, T key, V def = V()) {
-    return m ? get(*m, key, def) : def;
+template <class Map, typename Key = typename Map::key_type,
```
vlstill, line 65:
```suggestion
template <class Map, typename Key,

File: lib/map.h [hunk 7]
```diff
@@ -17,42 +17,65 @@ limitations under the License.
...
+/// Given a map and a key, return the value corresponding to the key in the map,
+/// or a given default value if the key doesn't exist in the map.
+template <typename Map, typename Key>
+typename Map::mapped_type get(const Map &m, const Key &key) {
     auto it = m.find(key);
-    if (it != m.end()) return it->second;
-    return def;
+    return it != m.end() ? it->second : typename Map::mapped_type{};
+}
+template <class Map, typename Key, typename Value = typename Map::mapped_type>
```
vlstill, line 29: Now thinking about this once again, is there any reason for the default for the `Value`? I don't see when it would be used -- it seems to me the compiler would always derive `Value` from the expression passed in for `def` anyway.

asl: yes and if necessary it could be overriden

vlstill (approved):

PR #4485, opened 2024-02-29
Fix for default_action in tc backend

Sosutha:
* tc template code change for default miss action
 * NoAction case generation in c code for tc backend
 * default case deletion in c code for tc backend
 * p4tc testcases output updated


usha1830 (approved):
LGTM

PR #4488, opened 2024-02-29
Fetch latest brew formulae

asl:
Let's see if this would help M1


fruffy (approved):
Approving, but iirc this can be very slow. We should observe the CI times.

asl:
> Approving, but iirc this can be very slow. We should observe the CI times.

First time, yes. Then it should be cached. It's ~30 seconds locally to me

asl:
@fruffy Now it seems to work. Though, the dir for brew cache is wrong. I've added another step to capture it, testing

PR #4489, opened 2024-02-29
Midend def-use pass

grg:
Contributing a midend def-use analysis pass.

The pass is an inspector named ComputeDefUse. The pass populates a private attribute named `defuse`. This defuse information is queried via one of two functions: `getDefs` and `getUses`. Both functions take a `const IR::Node*` parameter and return the defs or uses for that node.

There is also a ostream operator that takes a ComputeDefUse object and prints the defuse information in human-readable form. For example:
```
defs:
  h.h1.f1<146088>(control:4:17): { <93152>(control:0:47) }
  inout ParsedHeaders h<93152>(control:0:47): { <93152>(control:0:47), <146097>(control:5:17), <146080>(control:2:13) }
  inout Metadata m<93156>(control:0:65): { <93156>(control:0:65) }
uses:
  inout ParsedHeaders h<93152>(control:0:47): { <146088>(control:4:17), <93152>(control:0:47) }
  h.h1.f1<146097>(control:5:17): { <93152>(control:0:47) }
  h.h1.f2<146080>(control:2:13): { <93152>(control:0:47) }
  inout Metadata m<93156>(control:0:65): { <93156>(control:0:65) }
```
The `defs` sections show for each element what the corresponding defs are (e.g., the use `h.h1.f1` on line 4 of control has a corresponding def on line 0 of control). Similarly, the `uses` section shows for each element what the corresponding uses are.

This pass has _not_ been extensively tested.

@ChrisDodd The second, third, and fourth commits are changes on top of the original def-use code: they fix a few issues that were uncovered while testing:
* The second fixes an issue where split_slice is passed a range that doesn't correspond to any existing slices.
* The third sets liveness from the input of a control block, and propagates liveness in do_write while drilling down into structlike types.
* The fourth attempts to distinguish between control blocks that are "type declarations" vs actual block instantiations and doesn't attempt to set field liveness for type declarations.

The test issue1914-1.p4 provides an example of a control block type declaration:
```
control EmptyVerifyChecksum<H, M>(inout H hdr, inout M meta) {
    apply { }
}
```
The midend def-use pass sees this as:
```
  [96958] P4Control name=EmptyVerifyChecksum declid=17
    type: [250] Type_Control name=EmptyVerifyChecksum declid=16
      annotations: [5] Annotations
      typeParameters: [238] TypeParameters
        [235] Type_Var name=H declid=14
        [236] Type_Var name=M declid=15
      applyParams: [248] ParameterList
        [242] Parameter name=hdr declid=37 direction=inout
          annotations: [5] Annotations
          type: [241] Type_Name
            path: [240] Path name=H absolute=0
        [246] Parameter name=meta declid=38 direction=inout
          annotations: [5] Annotations
          type: [245] Type_Name
            path: [244] Path name=M absolute=0
    constructorParams: [258] ParameterList
    body: [96979] BlockStatement
      annotations: [5] Annotations
```
Without the protection in the fourth commit, the pass would call resolveUnique on H and M which both fail.

### DIFF
--- a/midend/def_use.h
+++ b/midend/def_use.h
@@ -101,6 +101,7 @@ class ComputeDefUse : public Inspector,
     };
     ordered_map<const IR::IDeclaration *, def_info_t> def_info;
     void add_uses(const loc_t *, def_info_t &);
+    void set_live_from_type(def_info_t &di, const IR::Type *type);
 
     // computed defuse info for all uses and defs in the program
     struct defuse_t {
@@ -162,6 +163,7 @@ class ComputeDefUse : public Inspector,
     }
 
     // for debugging
+    friend std::ostream &operator<<(std::ostream &, const loc_t &);
     friend std::ostream &operator<<(std::ostream &, const defuse_t &);
     friend std::ostream &operator<<(std::ostream &out, const ComputeDefUse &cdu) {
         return out << cdu.defuse;
--- a/midend/def_use.h
+++ b/midend/def_use.h
@@ -1,0 +1,173 @@
+/*
+Copyright 2024 Intel Corp.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+#ifndef MIDEND_DEF_USE_H_
+#define MIDEND_DEF_USE_H_
+
+#include "frontends/common/resolveReferences/resolveReferences.h"
+#include "ir/ir.h"
+#include "lib/bitrange.h"
+
+namespace P4 {
+
+/**
+ * @brief Compute defuse info within P4Parser and P4Control blocks in the midend.
+ *
+ * This pass finds all uses and definitions of field/slice values in all controls
+ * and parsers in the program and stores maps of which defs reach which uses.
+ * After the pass runs, getDefs(use) will return all the definitions in the program
+ * that reach the argument use while getUses(def) will return all the uses of the
+ * given def.  The returned values are sets of ComputeUseDef::loc_t objects which
+ * contain both the node that is def or use as well as the context path from the root
+ * of the IR to that node -- actions that are used by mulitple tables or parser states
+ * reachable via multiple paths may have mulitple entries as a result
+ *
+ * @pre Currently the code does not consider calls between controls or parsers, as it is
+ * expected to run after inlining when all such calls have been flattened.
+ * It could be extended to deal with the before inlining case.
+ */
+class ComputeDefUse : public Inspector,
+                      public ControlFlowVisitor,
+                      public P4WriteContext,
+                      public P4::ResolutionContext {
+    ComputeDefUse *clone() const { return new ComputeDefUse(*this); }
+    void flow_merge(Visitor &) override;
+    void flow_copy(ControlFlowVisitor &) override;
+    enum { SKIPPING, NORMAL, READ_ONLY, WRITE_ONLY } state = SKIPPING;
+
+ public:
+    // a location in the program.  Includes the context from the visitor, which needs to
+    // be copied out of the Visitor::Context objects, as they are allocated on the stack and
+    // will become invalid as the IR traversal continues
+    struct loc_t {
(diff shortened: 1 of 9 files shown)


ir/CMakeLists.txt
```diff
@@ -14,6 +14,7 @@
 
 set (IR_SRCS
   base.cpp
+  bitrange.cpp
```
fruffy, line 17: This seems like its own contribution? Maybe we can do it in a separate PR?

fruffy: Also it looks like there is no IR usage at all, can we make part of `lib`?

fruffy (commented):
Awesome, thanks a lot for contributing this! Quick initial comments.

ir/CMakeLists.txt
grg: Sure, I can make it a separate PR. It'll need to go in first since the def-use pass uses `le_bitrange`.

I initially made it part of lib but then switched to ir. There are a pair of functions for JSON serialization/deserialization that require the JSONGenerator/JSONLoader classes from ir, but this causes compilation failure in clean builds because lib is built before the ir generator runs. (The JSONGenerator/JSONLoader source files pull in ir.h.)

The Tofino backend IR uses le_bitrange/nw_bitrange in a handful of places, so it requires the ability to serialize to JSON. Here are a few ways I could deal with this:
* Strip JSON serialization code from the open-source bitrange code, and subclass in Tofino to add JSON serialization back in. This would require work to ensure the Tofino code always uses the subclass.
* Only two functions in bitrange.cpp use JSONGenerator/JSONLoader. I could place these two functions in ir/bitrange.cpp and have the rest of the code in lib. But this means that using p4ctoolkit library would require also linking in the ir library.
* Leave all the code in ir, even though the open source IR doesn't use the new bitrange classes.
Any better suggestions that these options?

grg: Done: #4496

fruffy (approved):
I am surprised this code does not have any impact on existing ref files.

grg:
> I am surprised this code does not have any impact on existing ref files.

Why do you say that? The information collected by this pass in the p4test backend is not currently being used by any subsequent passes.

fruffy:
> > I am surprised this code does not have any impact on existing ref files.
> 
> Why do you say that? The information collected by this pass in the p4test backend is not currently being used by any subsequent passes.

Nevermind, for some reason I thought this would be used by a latter pass. Ignore my drive-by comment.

grg:
> > > I am surprised this code does not have any impact on existing ref files.
> > 
> > 
> > Why do you say that? The information collected by this pass in the p4test backend is not currently being used by any subsequent passes.
> 
> Nevermind, for some reason I thought this would be used by a latter pass. Ignore my drive-by comment.

Ideally it will be used by a latter pass. We've just got to wait for someone to write that latter pass 🙂

ChrisDodd (approved):
Overall I think this is not quite fully-baked and needs more testing, but getting it in will make that easier to do.

asl:
@grg Just to understand the place of this pass as compared to frontend use-def pass. Is it possible to provide some explanation, etc. about the intended usage?

midend/def_use.h
```diff
@@ -143,6 +144,7 @@ class ComputeDefUse : public Inspector,
  public:
     ComputeDefUse()
         : ResolutionContext(true), cached_locs(*new std::set<loc_t>), defuse(*new defuse_t) {
+        visitDagOnce = false;
```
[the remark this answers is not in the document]
grg: I've taken this out and all the tests still pass.

There was an issue that this worked around, but I don't recall the details. Looking at the commit that introduced that change, I think it was one of two things:
- Either this gave "better" (but still not correct) information for headers passed into/out of a control block before I added the code to set liveness for all in headers.
- Something else I was experimenting with needed this, but I removed the rest of the experiment but accidentally left this in.

grg:
> @grg Just to understand the place of this pass as compared to frontend use-def pass. Is it possible to provide some explanation, etc. about the intended usage?

I believe the front-end def-use passes identify all the defs of fields (ComputeWriteSet in def_use.h/cpp), and identify _which_ writes have uses (FindUnitialized in simplifyDefUse.cpp), but I don't think these record _where_ the defs are used.

This new pass computes the defs and uses for each field, allowing for queries of the form "give me all the defs for node A" or "give me all the uses for node B".

Similar defuse information is used in the Tofino compiler for things like live range analysis, which is part of the PHV allocation (roughly register allocation), identifying uninitialized reads coming from P4 code, and additional unused field elimination. The Tofino code uses a backend def-use pass to do this, but Chris has implemented this midend version before switching roles to allow some of this functionality/additional functionality to be placed in the midend.

@ChrisDodd Feel free to correct anything I've got wrong or elaborate where you feel it might be useful.

PR #4491, opened 2024-03-01
Add ninja as a default build system for some runners.

### LINKED ISSUE #4471: Introduce modern CMake build style and Ninja to P4C

fruffy:
Fixes #4471.

Needs a bit more work. There are lot of different build systems and options scattered around.

### DIFF
--- a/.github/workflows/ci-test-fedora.yml
+++ b/.github/workflows/ci-test-fedora.yml
@@ -43,8 +43,8 @@ jobs:
 
       - name: Build p4c (Fedora Linux)
         run: |
-          ./bootstrap.sh -DCMAKE_BUILD_TYPE=RELEASE -DCMAKE_UNITY_BUILD=ON
-          make -j$((`nproc`+1)) -C build
+          ./bootstrap.sh -DCMAKE_BUILD_TYPE=RELEASE -DCMAKE_UNITY_BUILD=ON --build-generator "Ninja"
+          cmake --build build -- -j $(nproc)
 
       - name: Run p4c tests (Fedora Linux)
         # Need to use sudo for the eBPF kernel tests.
--- a/tools/ci-build.sh
+++ b/tools/ci-build.sh
@@ -38,6 +38,8 @@ P4C_DIR=$(readlink -f ${THIS_DIR}/..)
 : "${ENABLE_SANITIZERS:=OFF}"
 # Only execute the steps necessary to successfully run CMake.
 : "${CMAKE_ONLY:=OFF}"
+# The build generator to use. Defaults to Make.
+: "${BUILD_GENERATOR:="Unix Makefiles"}"
 # Build with -ftrivial-auto-var-init=pattern to catch more bugs caused by
 # uninitialized variables.
 : "${BUILD_AUTO_VAR_INIT_PATTERN:=OFF}"
@@ -62,6 +64,7 @@ P4C_DEPS="bison \
           build-essential \
           ccache \
           flex \
+          ninja-build \
           g++ \
           git \
           lld \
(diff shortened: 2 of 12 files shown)


tools/ci-build.sh
```diff
@@ -248,17 +249,16 @@ fi
...
 mkdir -p ${P4C_DIR}/build
 cd ${P4C_DIR}/build
-cmake ${CMAKE_FLAGS} ..
+cmake ${CMAKE_FLAGS} -G Ninja ..
 
 # If CMAKE_ONLY is active, only run CMake. Do not build.
 if [ "$CMAKE_ONLY" == "OFF" ]; then
-  make -j$((`nproc`+1))
-  sudo make -j$((`nproc`+1)) install
+  cmake --build . -- -j $(nproc)
```
asl, line 256: Note that you do not need `-j` with ninja. It defaults to all cpu cores available.

fruffy: In case you use make or something else as generator we may want to preserve the parallelism.

tools/ci-build.sh
```diff
@@ -85,8 +88,13 @@ sudo apt-get update
 sudo apt-get install -y --no-install-recommends ${P4C_DEPS}
 sudo pip3 install --upgrade pip
 sudo pip3 install -r ${P4C_DIR}/requirements.txt
-# ! ------  END CORE -----------------------------------------------
 
+if ! command -v ninja &> /dev/null
+then
+    echo "ninja could not be found, using 'make' instead"
+    BUILD_GENERATOR="Unix Makefiles"
+fi
```
vlstill, line 96: I wonder if this is a good idea. I suggest we fail in CI if Ninja is not available. That will make sure we are really building with Ninja and not accidentally building with make because Ninja got lost somewhere. The check would be there only if Ninja is selected.

.github/workflows/ci-test-fedora.yml
vlstill, line 1: Please explicitly select makefiles in the configure so we don't end with Ninja accidentally if/when we change the default. Also please add a comment that this one deliberately uses make to test that make works. I would maybe even mention it in the name of the build.

vlstill (commented):
Great to have Ninja officially supported/tested! Just two suggestions to make the CI a bit more robust.

tools/ci-build.sh
```diff
@@ -39,7 +39,7 @@ P4C_DIR=$(readlink -f ${THIS_DIR}/..)
 # Only execute the steps necessary to successfully run CMake.
 : "${CMAKE_ONLY:=OFF}"
 # The build generator to use. Defaults to Ninja. If that is not available "make" is used.
-: "${BUILD_GENERATOR:=Ninja}"
+: "${BUILD_GENERATOR:="Unix Makefiles"}"
```
vlstill, line 42: You probably need to enclose the inner string in apostrophes:

```suggestion
: "${BUILD_GENERATOR:='Unix Makefiles'}"

File: tools/ci-build.sh [hunk 4]
```diff
@@ -39,7 +39,7 @@ P4C_DIR=$(readlink -f ${THIS_DIR}/..)
 # Only execute the steps necessary to successfully run CMake.
 : "${CMAKE_ONLY:=OFF}"
 # The build generator to use. Defaults to Ninja. If that is not available "make" is used.
```
vlstill, line 41: The comment does not match code.

vlstill (commented):
Is there a build that uses make now? I don't see any and I think it is a good idea to have one. Ubuntu unity could be a good candidate as it should be relatively fast anyway as a unity build. The non-unity could be an alternative if the speed is of concern as it does not run tests (it is hard, but not impossible, to imagine a different build system to cause bugs in the build).

fruffy:
> Is there a build that uses make now? I don't see any and I think it is a good idea to have one. Ubuntu unity could be a good candidate as it should be relatively fast anyway as a unity build. The non-unity could be an alternative if the speed is of concern as it does not run tests (it is hard, but not impossible, to imagine a different build system to cause bugs in the build).

I figured we could use the nightly sanitizer build for this: https://github.com/p4lang/p4c/blob/main/.github/workflows/ci-ubuntu-20-sanitizer-nightly.yml

I can also change the unity build to use make again.

tools/ci-build.sh
fruffy: This does seem to break the script.

vlstill:
OK, I think having make just in nightly would be sufficient, I am OK with either of these choices.

vlstill (approved):

PR #4492, opened 2024-03-01
Build P4Testgen on MacOS.

fruffy:
Make sure we can also build P4Testgen on MacOS. This does not run tests. We still need to enable BMv2 support for that.


asl (approved):

vlstill (approved):
Let's have testgen everywhere! :-)

PR #4496, opened 2024-03-01
HalfOpenRange/ClosedRange classes

grg:
New classes provide useful abstractions, such as le_bitrange/nw_bitrange.

Splitting out from #4489 at @fruffy's request.

@fruffy I've introduced a new helper struct `RangeJsonHelpers` that contains a pair std::function instances for the rangeToJSON/rangeFromJSON functions. The placeholder implementation of these functions call BUG with a message that the functionality is unimplemented. There's a function `ir/bitrange.h:registerRangeJsonHelpers()` that replaces the placeholders with functions that invoke the appropriate JSONGenerator/JSONLoader functions.

~~@ChrisDodd / @vlstill / @fruffy Do any of you have a better suggestion for how to handle JSON serialization/de-serialization. Here's the backstory/summary:~~
* ~~I'm porting the bitrange classes from Tofino to use le_bitrange in the midend def-use pass (#4489).~~
* ~~The open source compiler doesn't use le_bitrange and other HalfOpenRange/ClosedRange derivatives in the IR, so @fruffy suggested we put the code in lib/bitrange.h + cpp.~~
* ~~Tofino _does_ use these classes in backend IR, and so requires the ability to serialize/deserialize the objects.~~
* ~~lib builds before the IR generator runs, and so lib code can't invoke JSONGenerator/JSONLoader functions since their header files indirectly use ir-generated.h.~~
* ~~I've introduced a helper struct with function instances for serialization/deserialization. The default functions are placeholders that call BUG. I've added a `registerRangeJsonHelpers()` in ir that registers real functions that correctly invoke JSONGenerator/JSONLoader if serialization/deserialization is required.~~

Refactored to avoid this ugliness after @vlstil's [suggestion](https://github.com/p4lang/p4c/pull/4496#issuecomment-1976559045).

### DIFF
--- a/lib/bitrange.h
+++ b/lib/bitrange.h
@@ -17,7 +17,15 @@ limitations under the License.
 #ifndef LIB_BITRANGE_H_
 #define LIB_BITRANGE_H_
 
+#include <algorithm>
+#include <iosfwd>
+#include <limits>
+#include <optional>
+#include <utility>
+
 #include "bitvec.h"
+#include "exceptions.h"
+#include "hash.h"
 
 /* iterate over ranges of contiguous bits in a bitvector */
 class bitranges {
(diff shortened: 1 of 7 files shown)


grg:
> @grg I think what you could do is:
> 
> * Have a non-templated, freestanding `rangeToJson`/`rangeFromJson` just declared in `bitrange.h`.
> * Use these to implement the member functions in the bitranges.
> * Define these functions in some .cpp in the `ir/` component.
> * I believe the linker should not complain for the `lib` component itself, and not even for something that uses it as the declared-only function is not used (and not `virtual`). The `fromJson`/`toJson` members are instantiated only when called anyway (as the bitrange is a template).
> * The linker would pull out the implementation from the `ir` normally unless someone links just `lib` and not `ir` and uses the `fromJson`/`toJson` (which is basically impossible as the loader/generator is also defined in `ir`).
> 
> Alternatively, you could even remove the member functions completely from bitrange and replace them by freestanding functions declared and defined in `ir` completely but that would probably be uglier.

Thanks @vlstill. I should have tried this from the beginning. I incorrectly assumed that the linker would complain because it couldn't find implementations for rangeToJSON/rangeFromJSON. But you're correct that it only complains if those functions are called. (I tested this before updating the PR.)

fruffy (commented):
Very nice, thank you! Any chance there are tests we get for "free", i.e., they are already downstream? If not, that's fine.

fruffy (approved):

lib/bitrange.h
```diff
@@ -51,4 +58,686 @@ class bitranges {
...
+class JSONGenerator;
+class JSONLoader;
+
+/**
+ * @return the result of dividing @dividend by @divisor, rounded towards
+ * negative infinity. This is different than normal C++ integer division, which
+ * rounds towards zero. For example, `-7 / 8 == 0`, but
+ * `divideFloor(-7, 8) == -1`.
+ */
+inline int divideFloor(int dividend, int divisor) {
```
asl, line 70: Can these be moved to some separate header? E.g. `math_extras.h` or so? And / or put into some `detail` namespace instead?

lib/bitrange.h
```diff
@@ -51,4 +58,686 @@ class bitranges {
...
+/**
+ * A helper type used to construct a range, specified as the closed interval
+ * between two bit indices. For example, `FromTo(6, 8)` denotes a range
+ * containing three bits: bit 6, bit 7, and bit 8.
+ *
+ * FromTo is only intended to be used as a constructor parameter, to allow
+ * creation of ranges in a more readable manner; use it to construct a
+ * HalfOpenRange or ClosedRange.
+ */
+struct FromTo {
```
asl, line 128: Can these be put into some namespace? So the usage from the source code would be e.g. `bitrange::FromTo(6, 8)`?

lib/bitrange.h
```diff
@@ -51,4 +58,686 @@ class bitranges {
...
+/**
+ * A helper type used to construct a range, specified as the region of a certain
+ * length beginning at a certain bit index. For example, `StartLen(6, 8)`
+ * denotes a range containing eight bits: bit 6, bit 7, bit 8, bit 9, bit 10,
+ * bit 11, bit 12, and bit 13.
+ *
+ * As with FromTo, StartLen is only meant to be used when constructing a
+ * HalfOpenRange or ClosedRange.
+ */
+struct StartLen {
```
asl, line 145: Ditto

asl:
@grg Lots of types are exposed to a default namespace here. It seems some are either implementation details and could be hidden or generic enough to be moved to a separate header. Some use of additional namespaces might provide some usage benefits IMO :)

grg:
> Very nice, thank you! Any chance there are tests we get for "free", i.e., they are already downstream? If not, that's fine.

I've pulled in the Tofino bitrange tests. I had to make a few small changes, but mostly came for free.

fruffy:
> The work I'm creating for myself while porting the Tofino midend def-use pass 🤣

Well, your contributions are very much appreciated! 😄

lib/bitrange.h
grg: I've kept them in bitrange.h for now but put them in a BitRange::Detail namespace.

lib/bitrange.h
grg: Moved to BitRange namespace.

grg:
@asl Please take a look at the namespace updates.

lib/bitrange.h
```diff
@@ -51,4 +58,714 @@ class bitranges {
...
+ * @return the result of dividing @dividend by @divisor, rounded towards
+ * negative infinity. This is different than normal C++ integer division, which
+ * rounds towards zero. For example, `-7 / 8 == 0`, but
+ * `divideFloor(-7, 8) == -1`.
+ */
+inline int divideFloor(int dividend, int divisor) {
+    const int quotient = dividend / divisor;
+    const int remainder = dividend % divisor;
+    if ((remainder != 0) && ((remainder < 0) != (divisor < 0))) return quotient - 1;
+    return quotient;
```
ChrisDodd, line 77: Does this get properly folded to a shift when divisor is a constant power of 2?  Similarly, moduloFloor should get folded to an and

ChrisDodd (approved):

lib/bitrange.h
```diff
@@ -51,4 +58,714 @@ class bitranges {
...
+        return HalfOpenRange(BitRange::rangeFromJSON(json));
+    }
+
+    /// Total ordering, first by lo, then by hi.
+    bool operator<(const HalfOpenRange &other) const {
+        if (lo != other.lo) return lo < other.lo;
+        return hi < other.hi;
+    }
+
+    friend size_t hash_value(const HalfOpenRange &r) {
```
asl, line 456: Sorry, I missed this previously... While we are here, can we add `std::hash` / `Util::Hasher` specializations? Also, `Utils::hash_combine` combines `std::hash<T>`'s by default with `T` being `int` here, so I think it would be better to use `return Utils::Hash{}(r.lo, r.hi)` here, it would mix better (not sure it matters, but just in case).

Same for `ClosedRange` below.

lib/bitrange.h
grg: Testing in the [Compiler Explorer](https://godbolt.org/z/Ks3rf7e3T):
* clang: folds it to a right shift
* gcc: does _not_ fold to a right-shift
(I tested several different -O levels with gcc, but didn't get it to fold.)

Any suggestions how to restructure the code to get gcc to fold it to a shift?

grg: And neither compiler does the fold to and for moduloFloor 🙁

lib/bitrange.h
grg: I think I've implemented what you've requested with the specializations. Please take a look and let me know if anything is missing.

asl (approved):
LGTM, Thanks!

lib/bitrange.h
grg: Updated Compiler Explorer with both divideFloor and moduloFloor: https://godbolt.org/z/8a1KeYGon

grg: I've got a solution to this, but I'm concerned about its portability. Will open a separate PR for that to get feedback after this has merged.

PR #4505, opened 2024-03-06
Link Bazel badge in the Bazel README section to the actual CI test

Sanket-0510:
The link to the bazel build is broken or rather it is not provided, I have added the right link for the same.

### DIFF
--- a/README.md
+++ b/README.md
@@ -430,8 +430,7 @@ at least 4GB via Docker preferences or you are likely to see "internal compiler
 errors" from gcc which are caused by low memory.
 
 # Bazel
-
-![bazel build](https://github.com/p4lang/p4c/workflows/bazel/badge.svg)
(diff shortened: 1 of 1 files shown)


fruffy:
That's a funny one, thanks for spotting it. Can you please make the pull request title more precise? Describe more clearly what broken link is fixed here. 

Then we can merge this one.

Sanket-0510:
> Thanks! On naming your pull requests and commits I would recommend following conventions such as the imperative style. I edited the title with an example of how to do this.
> 
> There are some guides out there, for example: https://cbea.ms/git-commit/ https://tbaggery.com/2008/04/19/a-note-about-git-commit-messages.html
> 
> These guidelines are not mandatory, for example I prefer to use periods, but are good to keep in mind.

thanks for this @fruffy I will make sure to follow these from next time

PR #4508, opened 2024-03-06
gc: implement posix_memalign

ChrisDodd:
In case any backend wants to use this function.


fruffy (approved):

vlstill (approved):

lib/gc.cpp
```diff
@@ -262,6 +262,12 @@ void *calloc(size_t size, size_t elsize) {
     if (rv) memset(rv, 0, size);
     return rv;
 }
+int posix_memalign(void **memptr, size_t alignment, size_t size) throw() {
```
vlstill, line 265: `throw()` should be equivalent to `noexcept` in C++17 and will be dropped in C++20.
But sadly it seems like the exception spec will need to be os-dependent :-(.

PR #4509, opened 2024-03-07
Add 4 digit uid to dump file names

ChrisDodd:
I noticed an issue with dumping passes that get called more than once, generally in a `PassRepeated` -- the filename generated is the same, so the dumps overwrite and you end up with only the last iteration of the pass.  This change adds a unique index number so you get mulitple files, and as a bonus, the files will be sorted in order in `ls`

### DIFF
--- a/frontends/common/parser_options.cpp
+++ b/frontends/common/parser_options.cpp
@@ -483,7 +483,9 @@ void ParserOptions::dumpPass(const char *manager, unsigned seq, const char *pass
             exit(1);
         }
         if (match) {
-            cstring suffix = cstring("-") + name;
+            char buf[16];
+            snprintf(buf, sizeof(buf), "-%04zu-", ++dump_uid);
+            cstring suffix = cstring(buf) + name;
             cstring filename = file;
             if (filename == "-") filename = "tmp.p4";
 
--- a/frontends/common/parser_options.h
+++ b/frontends/common/parser_options.h
@@ -43,6 +43,9 @@ class ParserOptions : public Util::Options {
     // annotation names that are to be ignored by the compiler
     std::set<cstring> disabledAnnotations;
 
+    // used to generate dump file names
+    mutable size_t dump_uid = 0;
+
  protected:
     // Function that is returned by getDebugHook.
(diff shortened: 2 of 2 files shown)


frontends/common/parser_options.cpp
```diff
@@ -483,7 +483,10 @@ void ParserOptions::dumpPass(const char *manager, unsigned seq, const char *pass
             exit(1);
         }
         if (match) {
-            cstring suffix = cstring("-") + name;
+            char buf[8];
+            static int uid;
```
vlstill, line 487: Can you please make this a member variable of `ParserOptions` rather than a static one? I think that starting dump count from 0 for new instance of options makes sense.

fruffy: Agree, we use `static` a lot and is causing quite a few problems these days.

ChrisDodd: Needs to be a mutable member, as this is a const function, but trivially done.

vlstill: I did not notice the const... it is not the nicest use of mutable, but I prefer it over `static`.

frontends/common/parser_options.h
```diff
@@ -43,6 +43,9 @@ class ParserOptions : public Util::Options {
     // annotation names that are to be ignored by the compiler
     std::set<cstring> disabledAnnotations;
 
+    // used to generate dump file names
+    mutable size_t dump_uid;
```
vlstill, line 47:
```suggestion
    mutable size_t dump_uid = 0;
```

... and drop the init in the constructor.

frontends/common/parser_options.cpp
```diff
@@ -45,6 +45,7 @@ const char *p4_14includePath = CONFIG_PKGDATADIR "/p4_14include";
 const char *ParserOptions::defaultMessage = "Compile a P4 program";
 
 ParserOptions::ParserOptions() : Util::Options(defaultMessage) {
+    dump_uid = 0;
```
vlstill, line 48:
```suggestion

vlstill (approved):
Just a nitpick with a more modern style.

PR #4511, opened 2024-03-07
Use Abseil system includes

fruffy:
Partially addresses #4523. Try to patch Abseil.

### DIFF
--- a/BUILD.bazel
+++ b/BUILD.bazel
@@ -210,10 +210,12 @@ cc_library(
     ],
     copts = [
         # Where p4c should look for p4include at runtime.
-        ("-DCONFIG_PKGDATADIR=\\\"external/%s\\\"" % repository_name()),
         # This will work only if the binary is executed by Bazel. For a general
         # solution, we would need to make p4c aware of Bazel, specifically:
         # https://github.com/bazelbuild/bazel/blob/master/tools/cpp/runfiles/runfiles_src.h
+        ("-DCONFIG_PKGDATADIR=\\\"external/%s\\\"" % repository_name()),
+        # Abseil does not export its includes, we add them manually.
+        "-Iexternal/com_google_absl",
     ],
     textual_hdrs = glob([
         "ir/**/*.h",
--- a/BUILD.bazel
+++ b/BUILD.bazel
@@ -210,12 +210,10 @@ cc_library(
     ],
     copts = [
         # Where p4c should look for p4include at runtime.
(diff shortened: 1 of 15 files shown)


asl:
@fruffy maybe you're having abseil installed globally? Bazel explicitly does on CI:
```
-iquote external/com_google_absl -iquote bazel-out/k8-fastbuild/bin/external/com_google_absl
```

So, you can only use `#include "absl/foo/bar.h"`

fruffy:
> @fruffy maybe you're having abseil installed globally? Bazel explicitly does on CI:

I do not have libabseil-dev installed, but it turns out to be some old caching. I do not exactly understand what Abseil does differently from Protobuf that prohibits angle-bracketed includes. 

Will use this PR as a scratchpad to see whether I can patch Abseil appropriately to support angle-bracketed includes.

smolkaj (approved):
Just for my understanding, what's the motivation here? Are angled brackets what's generally preferred in OSS for depending on other libraries?

fruffy:
> Just for my understanding, what's the motivation here? Are angled brackets what's generally preferred in OSS for depending on other libraries?

https://github.com/abseil/abseil-cpp/issues/740 provides some rationale. There are also arch or nix users who want to supply their own abseil/protobuf system packages.

smolkaj:
@fruffy, this PR is causing some trouble for upstream users.

Specifically, imagine an upstream user X depends on both `absl` and `p4c`. Then X would declare an absl version in their `x_deps.bzl` file. And now X need to also patch their `absl` version, or otherwise will trigger a compilation error in p4c.

Do we really want to require all users (and their transitive dependents) to patch their abseil versions?

I didn't realize this had such a big blast radius when I approved.

Would it be possible to revert this PR and think about a better solution?

fruffy:
This is for open-source Bazel users, right? Which repository is affected by this change? I do not really see a good way to solve this.  

I think it might be better to lobby the Abseil maintainers to fix the way they export their includes.

smolkaj:
Yes, correct. Off the top of my head, at least http://github.com/p4lang/p4-constraints and https://github.com/sonic-net/sonic-pins, though I suspect there are others.

> I think it might be better to lobby the Abseil maintainers to fix the way they export their includes.

Agreed, let's definitely do this.

However, we probablyu also need some short term solution here, because we don't if and when the above will converge. Could this PR be rolled back in the meantime?

fruffy:
This also seems like a big limitation in Bazel, how do you consolidate dependencies there if two repositories require a different version of the same dependency?

PR #4512, opened 2024-03-07
help compiler folding for divideFloor/moduloFloor

grg:
Add additional test to assist compiler folding to perform these replacements when the divisor is a power-of-two compile-time constant:
 - divideFloor -> right-shift
 - moduloFloor -> bitwise and

This is in response to this [PR comment](https://github.com/p4lang/p4c/pull/4496#discussion_r1515132560) from @ChrisDodd:
> Does this get properly folded to a shift when divisor is a constant power of 2? Similarly, moduloFloor should get folded to an and

The proposed solution adds a test that helps the compiler with optimization for a power-of-two compile-time constant:
```
if (__builtin_constant_p(divisor) && isPowerOfTwo(divisor))
```

`__builtin_constant_p()` and `__builtin_clz` are not supported by all compilers and so are protected by #if guards for gcc/clang.

This [Compiler Explorer test](https://godbolt.org/z/EExYboa5b) shows the impact of the extra test on both clang 15.0.0 and gcc 12.3. (These versions match those available on my test system.) The code contains multiple divideFloor/moduloFloor functions:
* divisor = 4 without optimization
* divisor = 4 with optimization above
* divisor = 5 without optimization
* divisor = 5 with optimization above

### DIFF
--- a/lib/CMakeLists.txt
+++ b/lib/CMakeLists.txt
@@ -93,3 +93,4 @@ add_library(p4ctoolkit STATIC ${LIBP4CTOOLKIT_SRCS})
 
 # Disable libcall (realloc, malloc) optimizations which may cause infinite loops.
 set_target_properties(p4ctoolkit PROPERTIES COMPILE_FLAGS -fno-builtin)
+target_link_libraries(p4ctoolkit absl::bits)
--- a/lib/bitrange.h
+++ b/lib/bitrange.h
@@ -23,6 +23,7 @@ limitations under the License.
 #include <optional>
 #include <utility>
 
+#include "absl/numeric/bits.h"
 #include "bitvec.h"
 #include "exceptions.h"
 #include "hash.h"
@@ -71,7 +72,16 @@ namespace Detail {
  * rounds towards zero. For example, `-7 / 8 == 0`, but
  * `divideFloor(-7, 8) == -1`.
  */
-inline int divideFloor(int dividend, int divisor) {
+constexpr inline int divideFloor(int dividend, int divisor) {
+#if defined(__GNUC__) || defined(__clang__)
+    // Code to enable compiler folding when the divisor is a power-of-two compile-time constant
+    // In this case, compiler should fold to a right-shift
+    // FIXME: Replace absl with std after moving to C++20
+    unsigned u_divisor = static_cast<unsigned>(divisor);
+    if (__builtin_constant_p(u_divisor) && absl::has_single_bit(u_divisor))
+        return dividend >> (absl::bit_width(u_divisor) - 1);
+#endif  // defined(__GNUC__) || defined(__clang__)
+
     const int quotient = dividend / divisor;
     const int remainder = dividend % divisor;
     if ((remainder != 0) && ((remainder < 0) != (divisor < 0))) return quotient - 1;
@@ -102,7 +112,15 @@ constexpr int modulo(int dividend, int divisor) {
  *     dividends than for negative dividends.
  * To make this concrete, `-7 % 8 == -7`, but `moduloFloor(-7, 8) == 1`.
  */
-inline int moduloFloor(const int dividend, const int divisor) {
+constexpr inline int moduloFloor(const int dividend, const int divisor) {
+#if defined(__GNUC__) || defined(__clang__)
+    // Code to enable compiler folding when the divisor is a power-of-two compile-time constant
+    // In this case, compiler should fold to a bitwise-and
+    // FIXME: Replace absl with std after moving to C++20
+    if (__builtin_constant_p(divisor) && absl::has_single_bit(static_cast<unsigned>(divisor)))
+        return dividend & (divisor - 1);
+#endif  // defined(__GNUC__) || defined(__clang__)
+
     const int remainder = modulo(dividend, divisor);
     if (remainder == 0 || dividend >= 0) return remainder;
     return divisor - remainder;
--- a/BUILD.bazel
+++ b/BUILD.bazel
@@ -48,6 +48,7 @@ cc_library(
         ":config_h",
         "@boost//:format",
         "@boost//:multiprecision",
+        "@com_google_absl//absl/numeric:bits",
         "@com_google_googletest//:gtest",
     ],
 )


lib/bitrange.h
```diff
@@ -65,13 +65,31 @@ class JSONLoader;
 namespace BitRange {
 
 namespace Detail {
+#if defined(__GNUC__) || defined(__clang__)
+/**
+ * @return boolean indicating if @p value is a power of two.
+ *
+ * This function is declared constexpr to allow it to be used in tests to optimize code generation.
+ */
+constexpr bool isPowerOfTwo(int value) { return value && !(value & (value - 1)); }
```
asl, line 74: What is the intended behavior wrt negative numbers? This currently returns `false` for e.g. `-2`, but it takes `int` as an argument. If negatives are not expected, I'd go with explicit `unsigned` here, or maybe even something like this?

```c++
template <typename T, typename = std::enable_if_t<std::is_unsigned_v<T>>>
[[nodiscard]] constexpr inline bool isPowerOfTwo(T value) noexcept {
 return (value != 0) && ((value & (value - 1)) == 0);
}
```

lib/bitrange.h
```diff
@@ -65,13 +65,31 @@ class JSONLoader;
...
+
+constexpr int log2ForPowerOfTwo(int value) { return sizeof(value) * 8 - __builtin_clz(value) - 1; }
```
asl, line 76: This explicitly assumes that `sizeof(value) == 4` as `__builtin_clz` accepts unsigned argument. Also for the clarity I'd change the argument to `unsigned`.

In fact, we can use abseil implementations of C++20 `<bit`> routines: `absl::countl_zero` and their friends. With FIXME to replace the implementation with standard ones.

Instead of `isPowerOfTwo ` above we could use `absl::has_single_bit`.

Even if we'd use our own implementations I'd use C++20 names to ensure we can just drop them in the future.

lib/bitrange.h
```diff
@@ -65,13 +65,31 @@ class JSONLoader;
...
+#endif  // defined(__GNUC__) || defined(__clang__)
+
 /**
  * @return the result of dividing @dividend by @divisor, rounded towards
  * negative infinity. This is different than normal C++ integer division, which
  * rounds towards zero. For example, `-7 / 8 == 0`, but
  * `divideFloor(-7, 8) == -1`.
  */
 inline int divideFloor(int dividend, int divisor) {
```
asl, line 85: it could be `constexpr`, right?

asl (commented):
See comments about possible simplifications and C++20-related usage

lib/bitrange.h
grg: I wish I'd known about the Abseil alternatives earlier: would have saved my attempting to roll my own equivalents to the STL functions that exist in C++20 and beyond.

lib/bitrange.h
grg: For divideFloor/moduloFloor, I think it makes sense to restrict the divisor to unsigned numbers. The dividend _does_ need to be signed, but I don't call isPowerOfTwo or log2ForPowerOfTwo on the dividend.

lib/bitrange.h
asl: Yeah, sorry. I would probably need to add some short summary to docs of what is available....

grg: BTW, do you know of a more portable alternative to `__builtin_constant_p`?

grg:
@asl Do you happen to know what I need to adjust in cmake to get it to build p4ctoolkit (lib directory) correctly? It currently fails with:
```
In file included from /home/runner/work/p4c/p4c/lib/bitrange.cpp:17,
                 from /home/runner/work/p4c/p4c/build/lib/CMakeFiles/p4ctoolkit.dir/Unity/unity_0_cxx.cxx:10:
/home/runner/work/p4c/p4c/lib/bitrange.h:26:10: fatal error: absl/numeric/bits.h: No such file or directory
   26 | #include "absl/numeric/bits.h"
      |          ^~~~~~~~~~~~~~~~~~~~~
compilation terminated.
make[2]: *** [lib/CMakeFiles/p4ctoolkit.dir/build.make:104: lib/CMakeFiles/p4ctoolkit.dir/Unity/unity_0_cxx.cxx.o] Error 1
```

asl:
> @asl Do you happen to know what I need to adjust in cmake to get it to build p4ctoolkit (lib directory) correctly? It currently fails with:

I think you'd need to do `target_link_library(lib absl::bits)` to have includes / paths propagated properly.

lib/bitrange.h
asl: > BTW, do you know of a more portable alternative to `__builtin_constant_p`?

Unfortunately, not. I believe the question was raised few times, but there is no direct alternative. Maybe `if consteval` from C++23 is the closest alternative.

lib/bitrange.h
```diff
@@ -65,13 +66,29 @@ class JSONLoader;
...
 
 namespace Detail {
+/**
+ * @return find the bit position of the first one bit
+ *
+ * Will return all-ones if no bit is set
+ *
+ * This function is declared constexpr to allow it to be used in tests to optimize code generation.
+ */
+constexpr unsigned ffs(unsigned value) { return sizeof(value) * 8 - absl::countl_zero(value) - 1; }
```
grg, line 76: @asl I think this comment was on this line (I force-pushed a few times and it seemed to result in losing the comment in github):
> Isn't it countr_zero essentially?

It is as long as value is a power of 2 🙂 
I'm only using it in the case where value is a power of 2, so maybe I will switch. But I'll wait until I've got my build issues sorted out.

grg:
> > @asl Do you happen to know what I need to adjust in cmake to get it to build p4ctoolkit (lib directory) correctly? It currently fails with:
> 
> I think you'd need to do `target_link_library(lib absl::bits)` to have includes / paths propagated properly.

That seems to do the trick.

lib/CMakeLists.txt
```diff
@@ -93,3 +93,4 @@ add_library(p4ctoolkit STATIC ${LIBP4CTOOLKIT_SRCS})
 
 # Disable libcall (realloc, malloc) optimizations which may cause infinite loops.
 set_target_properties(p4ctoolkit PROPERTIES COMPILE_FLAGS -fno-builtin)
+target_link_libraries(p4ctoolkit absl::bits)
```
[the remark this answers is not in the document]
asl: Yeah, right. In few other places we only depend inside .cpp, so `PRIVATE` is a way to reduce visibility.

grg:
@asl I think I've got it down to a minimal set of changes using Abseil functionality as much as possible.

asl (approved):
Thanks!

PR #4515, opened 2024-03-08
Update formatters (clang-format, isort, black)

fruffy:
It's been a while since we have last updated the formatters. clang-format in particular has implemented a variety of changes.

### DIFF
--- a/lib/hashvec.cpp
+++ b/lib/hashvec.cpp
@@ -314,11 +314,12 @@ void hash_vector_base::redo_hash() {
         if (erased[i]) continue;
         const void *k = getkey(i);
         if (find(k, &cache)) {
-            if (info->ismulti)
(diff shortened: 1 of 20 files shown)


lib/hashvec.cpp
```diff
@@ -314,11 +314,12 @@ void hash_vector_base::redo_hash() {
         if (erased[i]) continue;
         const void *k = getkey(i);
         if (find(k, &cache)) {
-            if (info->ismulti)
-                while (find_next(k, &cache))
-                    ;
-            else
+            if (info->ismulti) {
```
fruffy, line 317: These changes are a little funky, I believe explicit brackets make the intent clearer.

asl: Yes!

asl (approved):

PR #4517, opened 2024-03-10
Fix broken MacOS and Ubuntu18.04 runs. 

fruffy:


tools/install_mac_deps.sh
```diff
@@ -44,7 +44,9 @@ REQUIRED_PACKAGES=(
     ${BOOST_LIB}
 )
 for package in "${REQUIRED_PACKAGES[@]}"; do
-  brew_install ${package}
+    # FIXME: || true should not be needed but because boost@1.84 pulls in python@3.12, which fails
+    # to install we need to ignore the error.
+  brew_install ${package} || true
```
asl, line 49: Do we know which package has `/usr/local/bin/2to3`?

fruffy: I think it is the Python package.

fruffy: I have tried a couple things, including removing `2to3` but there are more files it conflicts with. Even when you install python@3.12 beforehand, the boost installation tries to pull it in again and fails. Maybe there is a convenient flag to deal with that.

asl: Looks like known issue: https://github.com/actions/runner-images/issues/4020 Some solutions are quite harsh: https://github.com/Gelmo/warfork-qfusion/commit/79f3dd021b5b74d19647a4bf225d488c9b3dd44d

asl (approved):
Sadly we have to do this...

PR #4520, opened 2024-03-11
Clean up P4Tools and control-plane linking.

fruffy:
Make sure we only export what is necessary and link most libraries privately. Also remove the `add_p4tools_library` function, it is too inflexible and `add_library` works just fine.

### DIFF
--- a/control-plane/CMakeLists.txt
+++ b/control-plane/CMakeLists.txt
@@ -96,19 +96,17 @@ set (CONTROLPLANE_HDRS
 )
 
 add_library(controlplane-gen OBJECT ${P4RUNTIME_GEN_SRCS})
-# TODO(https://github.com/p4lang/p4c/issues/4477):
-# We should not need this, but because of the way managed includes and
-# libraries for Protobuf we need to explicitly add the include directories here.
-target_include_directories(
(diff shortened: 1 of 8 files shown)


control-plane/CMakeLists.txt
```diff
@@ -96,19 +96,17 @@ set (CONTROLPLANE_HDRS
...
-# libraries for Protobuf we need to explicitly add the include directories here.
-target_include_directories(
-  controlplane-gen SYSTEM BEFORE PUBLIC ${Protobuf_INCLUDE_DIRS} SYSTEM BEFORE
-  PUBLIC ${ABSL_COMMON_INCLUDE_DIRS}
+
+target_include_directories(controlplane-gen
+  # TODO(https://github.com/p4lang/p4c/issues/4477):
+  # We should not need this, but because of the way managed includes and
+  # libraries for Protobuf we need to explicitly add the include directories here.
+  SYSTEM BEFORE PUBLIC ${Protobuf_INCLUDE_DIRS}
```
asl, line 104: While we are here, maybe solve protobuf include madness? Can we just link protobuf as library?

fruffy: Since we pushed the minimum Protobuf version to 3.25 we should be able to clean this up yeah. Was planning on doing it once I have some more cycles.

asl (approved):

PR #4521, opened 2024-03-12
 YAML formatting from Redhead YAML language server.

fruffy:
https://github.com/redhat-developer/yaml-language-server includes an auto-formatter.

### DIFF
--- a/.github/workflows/ci-auto-release.yml
+++ b/.github/workflows/ci-auto-release.yml
@@ -2,7 +2,7 @@ name: Automated Release
 
 on:
   schedule:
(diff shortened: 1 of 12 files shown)


vlstill (commented):
looks good. Is there any plan to automatically check the formatting, or is that not worth the work? My guess is we can do it later if we have more modifations of these files breaking the format.

fruffy:
> Is there any plan to automatically check the formatting, or is that not worth the work?

Did not plan for it right now. This was just a spontaneous fixup because I was editing the YAML files and noticed the indentation/formatting was off. Our linting run is fast, it should not be too difficult to add.

vlstill (approved):

PR #4524, opened 2024-03-12
tc struct fields addition and memory initialization

Sosutha:
* Addition of fields in table key value structure in tc backend

* Memory initialization of key structure

* Updated testcase outputs


usha1830 (approved):
LGTM. Please rebase.

PR #4525, opened 2024-03-12
Try yet another MacOS-14 fix.

fruffy:
Seems like Github is tweaking their runner image a lot. Fix another breakage.


vlstill (approved):
While I don't understand `brew` this one looks much nicer then the previous workaround :-).

asl (approved):
Looks reasonable alternative to me

PR #4526, opened 2024-03-12
Move the network hashes implementation from Bmv2 testgen to lib/

vlstill:
There are hash functions from various network protocols hidden deep in the BMv2 testgen target implementation. This moves then to the `lib/` so they can be used elsewhere (mainly I would presume in other testgen targets, but I don't see anything testgen/p4tools specific about them, therefore I'm moving them to top-level `lib/`).

I'm open to a better name then `nethash` if someone has a suggestion :-).


vlstill:
I plan to do follow-up PRs with more library functions to allow code sharing with the BMv2 concolic functions implementation.

vlstill:
Seems like the macOS failure may be related to #4525.

fruffy:
Fyi, these are lifted from https://github.com/p4lang/behavioral-model/blob/5f1c590c7bdb32ababb6d6fe18977cf13ae3b043/include/bm/bm_sim/calculations.h but at this point we can freely modify them. The good thing is changes are differentially tested against BMv2's execution of the same hashes.

lib/nethash.cpp
```diff
@@ -1,13 +1,30 @@
-#include "backends/p4tools/modules/testgen/targets/bmv2/contrib/bmv2_hash/calculations.h"
+#include "lib/nethash.h"
 
 #include <arpa/inet.h>
 
 #include <algorithm>
 
-namespace P4Tools::P4Testgen::Bmv2 {
+namespace nethash {
```
fruffy, line 7: Namespaces are typically CamelCase here.

backends/p4tools/modules/testgen/targets/bmv2/concolic.cpp
```diff
@@ -278,4 +269,24 @@ const ConcolicMethodImpls::ImplList *Bmv2Concolic::getBmv2ConcolicMethodImpls()
...
+    switch (algo) {
+        ALGO_CASE(crc32);
+        ALGO_CASE(crc32_custom);
+        ALGO_CASE(crc16);
+        ALGO_CASE(crc16_custom);
+        ALGO_CASE(random);
+        ALGO_CASE(identity);
+        ALGO_CASE(csum16);
+        ALGO_CASE(xor16);
+        // no default: let the compiler produce a warning if some defined enum value is not covered
```
fruffy, line 285:
```suggestion
        // No default: let the compiler produce a warning if some defined enum value is not covered.
```

backends/p4tools/modules/testgen/targets/bmv2/concolic.cpp
```diff
@@ -278,4 +269,24 @@ const ConcolicMethodImpls::ImplList *Bmv2Concolic::getBmv2ConcolicMethodImpls()
...
+    }
+#undef ALGO_CASE
+    // for values other then the declared ones
```
fruffy, line 288:
```suggestion
    // For values other then the declared ones.
```

lib/nethash.cpp
```diff
@@ -1,13 +1,30 @@
...
 
-/* generating from my Python script gen_crc_tables inspired from the C code at:
-   http://www.barrgroup.com/Embedded-Systems/How-To/CRC-Calculation-C-Code */
+/* This code was adapted from:
```
fruffy, line 9: Make this comment part of the header?

lib/nethash.h
```diff
@@ -0,0 +1,40 @@
...
+/*
+ * Antonin Bas (antonin@barefootnetworks.com)
+ *
+ */
+
+#include <cstddef>
+#include <cstdint>
+#include <vector>
+
+/// A collection of hashing functions commonly used in network protocols.
```
fruffy, line 28: Also would add a comment where this was adopted from:
https://github.com/p4lang/behavioral-model/blob/main/src/bm_sim/calculations.cpp

lib/nethash.cpp
vlstill: Actually I think that both this, and the license snipped with (c) Barefoot should be in the .cpp. The reason is that is the code that has that source. The declarations are kind of "on top of that". Or we can have the disclaimer at both, but I think it should be at the implementation and so should be the license/copyright.

vlstill:
@fruffy, I've done a little refactoring on to of the move and I've added basic tests for the CRC sums I've refactored. If you feel like this should be done separately, I can end this at 2f2161c and split the rest.

vlstill:
> The good thing is changes are differentially tested against BMv2's execution of the same hashes.

Sadly it seems at least the `crcCCITT` variant is never used in existing tests. I've accidentally broken it and only realized it after all the tests passes.

fruffy:
> @fruffy, I've done a little refactoring on to of the move and I've added basic tests for the CRC sums I've refactored. If you feel like this should be done separately, I can end this at [2f2161c](https://github.com/p4lang/p4c/commit/2f2161caf023679709e56cd513e485b8d89edaab) and split the rest.

Good idea, didn't even consider that we need tests. I think it makes sense to add the tests here: https://github.com/p4lang/behavioral-model/blob/main/tests/test_calculations.cpp

We can do this in this PR.

vlstill:
@fruffy, do you have any code suggestion? Do you think someone else should review this? I would really like to get this one in reasonably soon.

As for the tests in BMv2, I've tried adding them (p4lang/behavioral-model#1235), but I can't compile BMv2 on my work machine and sadly the CI does not show me what failed. I will re-try later.

fruffy:
Oh I mean the other way round. Since this is now a lib class we should copy the BMv2 tests to the compiler. Apologize for the confusion. 

Other than that I have no further comments, code

vlstill:
> Oh I mean the other way round. Since this is now a lib class we should copy the BMv2 tests to the compiler. Apologize for the confusion.

Oh! There are not many in BMv2. But I've added them and wrote some more for csum16 & identity.

vlstill:
The validate does not build, but this is unrelated -- some problem with `RemoveUnusedDeclarations` pass.

vlstill:
Changed the PR name to indicate that this is not only about testgen.

fruffy:
> The validate does not build, but this is unrelated -- some problem with `RemoveUnusedDeclarations` pass.

Will fix that.

fruffy (approved):

lib/nethash.h
```diff
@@ -0,0 +1,29 @@
...
+/// CRC16-CCITT (parameters: NO bit-reflection, polynomial 0x1021, init = ~0 and xor_out = 0).
+uint16_t crcCCITT(const uint8_t *buf, size_t len);
+
+/// CRC-32 (parameters: bit-reflection, polynomial 0x04C11DB7, init = ~0 and xor_out = ~0).
+uint32_t crc32(const uint8_t *buf, size_t len);
+
+/// 16-bit ones' complement checksum (used in IP, TCP, UDP, ...).
+uint16_t csum16(const uint8_t *buf, size_t len);
+
+uint16_t xor16(const uint8_t *buf, size_t len);
```
fruffy, line 24: Missing comments?

fruffy (approved):

PR #4528, opened 2024-03-12
Add policy controlling RemoveUnusedDeclarations

ChrisDodd:
This allows the target to define an override of RemoveUnusedDeclarations that changes the what gets removed and what doesn't, by defining a simple subclass of `P4::RemoveUnusedDeclarations` that only overrides one or two methods and is otherwise the same.

- allow removing unused externs when that is apropriate for the target
- allow not removing unused functions when that is appropriate

I tried just defining a hook function that is called for each IDeclaration, but using that ends up doing a whole series or dynamic_casts, which seems ugly and duplicates what Visitor is already doing for the preorder/postorder methods, so this ends up being more efficient and clearer.


grg (approved):
LGTM

fruffy (approved):

PR #4533, opened 2024-03-14
Support new ActionProfile annotations for P4Runtime.

jonathan-dilorenzo:
Support new annotations introduced to the P4Runtime standard in https://github.com/p4lang/p4runtime/pull/399.

### DIFF
--- a/control-plane/p4RuntimeArchStandard.h
+++ b/control-plane/p4RuntimeArchStandard.h
@@ -768,8 +768,12 @@ class P4RuntimeArchHandlerCommon : public P4RuntimeArchHandlerIface {
         auto id = symbols.getId(SymbolType::P4RT_ACTION_PROFILE(), actionProfile.name);
         setPreamble(profile->mutable_preamble(), id, actionProfile.name,
                     symbols.getAlias(actionProfile.name), actionProfile.annotations,
-                    // exclude @max_group_size if present
-                    [](cstring name) { return name == "max_group_size"; });
+                    // exclude @max_group_size, @selector_size_semantics, and
+                    // @max_member_weight if present
+                    [](cstring name) {
+                        return name == "max_group_size" || name == "selector_size_semantics" ||
+                               name == "max_member_weight";
+                    });
         profile->set_with_selector(actionProfile.type == ActionProfileType::INDIRECT_WITH_SELECTOR);
         profile->set_size(actionProfile.size);
         auto maxGroupSizeAnnotation = actionProfile.annotations->getAnnotation("max_group_size");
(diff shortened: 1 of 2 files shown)


control-plane/p4RuntimeArchStandard.h
```diff
@@ -786,6 +790,57 @@ class P4RuntimeArchHandlerCommon : public P4RuntimeArchHandlerIface {
...
         }
 
+        // By default, an action profile uses the SumOfWeights semantics.
+        auto selectorSizeSemanticsAnnotation =
+            actionProfile.annotations->getAnnotation("selector_size_semantics");
+        if (selectorSizeSemanticsAnnotation) {
+            if (actionProfile.type == ActionProfileType::INDIRECT_WITH_SELECTOR) {
+                auto selectorSizeSemantics = selectorSizeSemanticsAnnotation->expr[0]->toString();
+                CHECK_NULL(selectorSizeSemantics);
+                // The expression may only contain 'sum_of_weights' or 'sum_of_members'
```
fruffy, line 800: Nit: This is actually "sum_of_weights" and "sum_of_members"? Same applies to the error message a couple lines down.

control-plane/p4RuntimeArchStandard.h
```diff
@@ -786,6 +790,57 @@ class P4RuntimeArchHandlerCommon : public P4RuntimeArchHandlerIface {
...
+        }
+
+        // By default, an action profile uses the SumOfWeights semantics.
+        auto maxMemberWeightAnnotation =
+            actionProfile.annotations->getAnnotation("max_member_weight");
+        if (maxMemberWeightAnnotation) {
+            if (actionProfile.type == ActionProfileType::INDIRECT_WITH_SELECTOR &&
+                profile->has_sum_of_members()) {
+                auto maxMemberWeightConstant =
+                    maxMemberWeightAnnotation->expr[0]->to<IR::Constant>();
```
fruffy, line 827: You can use `checkedTo` here.

control-plane/p4RuntimeArchStandard.h
```diff
@@ -786,6 +790,57 @@ class P4RuntimeArchHandlerCommon : public P4RuntimeArchHandlerIface {
             }
         }
 
+        // By default, an action profile uses the SumOfWeights semantics.
+        auto selectorSizeSemanticsAnnotation =
+            actionProfile.annotations->getAnnotation("selector_size_semantics");
+        if (selectorSizeSemanticsAnnotation) {
+            if (actionProfile.type == ActionProfileType::INDIRECT_WITH_SELECTOR) {
+                auto selectorSizeSemantics = selectorSizeSemanticsAnnotation->expr[0]->toString();
```
fruffy, line 798: You can use `checkedTo` here and cast to an IR::StringLiteral, which has a cstring as value member. It looks like the annotation is parsed as that.

fruffy (approved):
Approving, some small suggestions.

control-plane/p4RuntimeArchStandard.h
jonathan-dilorenzo: Ah, good point. It's a little confusing for users though, because if they actually put `@selector_size_semantics(sum_of_weights)` P4 would get confused and think it was an identifier (at least it does if I try it). So they need to put `@selector_size_semantics("sum_of_weights")`. But if I say the expression needs to be `\"sum_of_weights\"`, then I worry they'll put `@selector_size_semantics("\"sum_of_weights\"")`... Wdyt?

fruffy: At least for the error message `\"sum_of_weights\"` should turn into `"sum_of_weights"` no?

Ha we can leave it as is. It looks like no matter what we do here there will be a user that could be confused.

jonathan-dilorenzo: Yeah :'(. Fixed the error message though. I think that's a good point.

fruffy (approved):

PR #4541, opened 2024-03-16
Remove obsolete `output_to_genfiles = True`.

tjgq:
This is a no-op for Bazel, which enables `--incompatible_merge_genfiles_directory` by default.

### DIFF
--- a/bazel/bison.bzl
+++ b/bazel/bison.bzl
@@ -61,5 +61,4 @@ genyacc = rule(
                   "subject to $(location ...) expansion.",
         ),
     },
-    output_to_genfiles = True,
 )


fruffy (approved):

fruffy:
Approving, but can you give some more detail on the rationale here? It can't hurt to make this explicit, even if it is the default value.

tjgq:
> Approving, but can you give some more detail on the rationale here? It can't hurt to make this explicit, even if it is the default value.

When `--incompatible_merge_genfiles_directory` is enabled (which is the case for Bazel, but not Blaze at Google), `output_to_genfiles` has no effect - it doesn't matter what you set it to. I'm making this change because we want to switch the default at Google, but we can't flip the flag globally - we have to do it rule by rule (and we import this rule for use internally).

Once we're done, `output_to_genfiles` will be removed entirely, so there's no advantage to leaving it here explicitly set to `False`. 99% of Bazel users don't encounter the `genfiles` directory at all, so leaving it is in fact more confusing.

fruffy:
Thanks!

PR #4543, opened 2024-03-16
Aggressively clean up the Protobuf CMake dependency.

### LINKED ISSUE #4477: Revisit Protobuf include directory policy
Now that we have increased the minimum Protobuf version we may actually be able to request a Protobuf version that includes the fixes to the CMake config file. This way we can just link to Protobuf normally instead of manually patching libraries and includes.

fruffy:
Now that we are mandating Protobuf 3.25 try to simplify the use of extra variables and workarounds for our Protobuf dependency.

- Use `protobuf_generate` instead our own custom command. This way we can rely on the canonical definitions instead of having to maintain our own generators.
- Do not use ${Protobuf_LIBRARY}, instead use the aliased versions `protobuf::`
- Do not use `PROTOBUF_PROTOC_INCLUDES`, instead use `Protobuf_INCLUDE_DIRS` only.
- Also clean up some unnecessary include directories in the top-level CMakeLists. Try to simplify things. 

With CMake 3.24 we may also be able to override `find_package` with FetchContent: https://cmake.org/cmake/help/latest/module/FetchContent.html

Fixes #4477.


fruffy:
@jkhsjdhjs We are trying to refresh our Protobuf dependency and simplify it but I know you had some trouble with setting the right variables in Arch Protobuf. If we mandate a recent version do these build settings work for you? Or do we still need to keep the compatibility variables?

jkhsjdhjs:
@fruffy Yep, it doesn't build for me with the changes of this PR:
```
CMake Error at cmake/Protobuf.cmake:14 (find_package):
  Could not find a configuration file for package "Protobuf" that is
  compatible with requested version "25.3".

  The following configuration files were considered but not accepted:

    /usr/lib/cmake/protobuf/protobuf-config.cmake, version: 25.3.0
    /lib/cmake/protobuf/protobuf-config.cmake, version: 25.3.0

  CMakeLists.txt:185 (p4c_obtain_protobuf)
```
It works if I remove the `25.3` from the `find_package()` call.

However, I then get these errors:
```
CMake Error at control-plane/CMakeLists.txt:53 (add_custom_command):
  Error evaluating generator expression:

    $<TARGET_FILE:protoc>

  No target "protoc"


CMake Error at backends/dpdk/CMakeLists.txt:29 (add_custom_command):
  Error evaluating generator expression:

    $<TARGET_FILE:protoc>

  No target "protoc"


CMake Error at control-plane/CMakeLists.txt:53 (add_custom_command):
  Error evaluating generator expression:

    $<TARGET_FILE:protoc>

  No target "protoc"


CMake Error at backends/dpdk/CMakeLists.txt:29 (add_custom_command):
  Error evaluating generator expression:

    $<TARGET_FILE:protoc>

  No target "protoc"
```

fruffy:
Thanks for checking this! Maybe using `25.3.0` works?

fruffy:
@jkhsjdhjs I tested this locally and on MacOS, and I believe things should now work. Took me a lot of trial and error.

jkhsjdhjs (approved):
> Took me a lot of trial and error.

Yes, I can see that :D

fruffy force-pushed a lot

I just tested a build with the changes on latest Arch Linux using system protobuf and abseil and it works fine. LGTM :+1:

fruffy:
> > Took me a lot of trial and error.
> 
> Yes, I can see that :D
> 
> fruffy force-pushed a lot
> 
> I just tested a build with the changes on latest Arch Linux using system protobuf and abseil and it works fine. LGTM 👍

Great! Yeah, a lot of this was me banging my head against the wall trying to figure out why the MacOS build can not find Protobuf. Only to realize that it requires path hints.

cmake/Protobuf.cmake
```diff
@@ -1,4 +1,6 @@
 macro(p4c_obtain_protobuf)
+  set(P4C_PROTOBUF_VERSION "25.3")
+  set(P4C_PROTOC_VERSION "4.25")
```
vlstill, line 3: is `P4C_PROTOC_VERSION` used at all?

vlstill (commented):
Just two things I've noticed. (I got little lost in all the changes so I did not fully review it, but it looks reasonable overall).

cmake/Protobuf.cmake
fruffy: No, I was hoping to use it for `https://github.com/p4lang/p4c/pull/4543/files#diff-08181e83a64a6a99c0661f941d7b457ae42e4313384e857565f4d9545fd82903R22` but that code still seems to break. Let me try it one more time.

fruffy: I just tried, and it looks like adding the version simply breaks the find_package command. I removed the version specifier for now.

vlstill (approved):
I am no familiar with protobuf and how it should be used in CMake. Nor am I familiar with the P4C control plane. But I don't see any problems with the code.

fruffy:
> I am no familiar with protobuf and how it should be used in CMake. Nor am I familiar with the P4C control plane. But I don't see any problems with the code.

fwiw I have been testing this PR in downstream dependencies for months now.

PR #4546, opened 2024-03-18
[P4Testgen] Clean up the direct extern map implementation for BMv2.

fruffy-g:
- Use cstring instead of the declaration pointer in the direct extern map.
- Check for all implementations.
- Create an explicit static array of implementations.

### DIFF
--- a/backends/p4tools/modules/testgen/targets/bmv2/map_direct_externs.cpp
+++ b/backends/p4tools/modules/testgen/targets/bmv2/map_direct_externs.cpp
@@ -54,10 +54,9 @@ bool MapDirectExterns::preorder(const IR::P4Table *table) {
         const auto *impl = table->properties->getProperty(tableProperty);
         if (impl != nullptr) {
             auto declInstanceOpt = getExternFromTableImplementation(impl);
-            if (!declInstanceOpt.has_value()) {
-                return false;
+            if (declInstanceOpt.has_value()) {
+                directExternMap.emplace(declInstanceOpt.value()->controlPlaneName(), table);
             }
-            directExternMap.emplace(declInstanceOpt.value()->controlPlaneName(), table);
         }
     }
     return false;
--- a/backends/p4tools/modules/testgen/targets/bmv2/map_direct_externs.cpp
+++ b/backends/p4tools/modules/testgen/targets/bmv2/map_direct_externs.cpp
@@ -1,5 +1,11 @@
 #include "backends/p4tools/modules/testgen/targets/bmv2/map_direct_externs.h"
 
+#include <cstdio>
+#include <optional>
+
+#include "ir/ir.h"
+#include "lib/error.h"
+
 namespace P4Tools::P4Testgen::Bmv2 {
 
 bool MapDirectExterns::preorder(const IR::Declaration_Instance *declInstance) {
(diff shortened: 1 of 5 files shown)


backends/p4tools/modules/testgen/targets/bmv2/map_direct_externs.h
```diff
@@ -1,34 +1,44 @@
...
 #include <map>
+#include <optional>
 
 #include "ir/ir.h"
 #include "ir/visitor.h"
+#include "lib/cstring.h"
 
 namespace P4Tools::P4Testgen::Bmv2 {
 
 /// A map of direct extern declarations which are attached to a table.
```
jonathan-dilorenzo, line 14: nit: It's worth adding a quick comment about what the map is. E.g. it's not immediately clear to me why this changed from an IDeclaration to  a cstring and what that cstring represents. Maybe: `// A map from extern names to tables they are attached to.` (if true)?

backends/p4tools/modules/testgen/targets/bmv2/map_direct_externs.cpp
```diff
@@ -1,46 +1,68 @@
...
+}
+
+bool MapDirectExterns::preorder(const IR::P4Table *table) {
+    // Try to get extern implementation properties from the table.
+    for (const auto *tableProperty : kTableExternProperties) {
+        const auto *impl = table->properties->getProperty(tableProperty);
+        if (impl != nullptr) {
+            auto declInstanceOpt = getExternFromTableImplementation(impl);
+            if (!declInstanceOpt.has_value()) {
+                return false;
```
jonathan-dilorenzo, line 58: This is a bit strange to me. I would've thought you'd just want to continue so that you can test all extern properties? Or, if having an implementation defined, yet not having that extern not be defined is super illegal, then maybe better to allow this function to fail?

This seems like a strange limbo state.

backends/p4tools/modules/testgen/targets/bmv2/map_direct_externs.cpp
```diff
@@ -1,46 +1,68 @@
...
 }
 
-bool MapDirectExterns::preorder(const IR::P4Table *table) {
-    // Try to either get counters or meters from the table.
-    const auto *impl = table->properties->getProperty("meters");
-    if (impl == nullptr) {
-        impl = table->properties->getProperty("counters");
+std::optional<const IR::Declaration_Instance *> MapDirectExterns::getExternFromTableImplementation(
+    const IR::Property *tableImplementation) {
+    const auto *selectorExpr = tableImplementation->value->to<IR::ExpressionValue>();
```
jonathan-dilorenzo, line 18: nit: Why move from `checkedTo` to just `to`? Isn't that less safe?

jonathan-dilorenzo: Maybe this is what the nullptr case below covers?

jonathan-dilorenzo (approved):
Minor points. LGTM.

backends/p4tools/modules/testgen/targets/bmv2/map_direct_externs.cpp
fruffy-g: It's an exception vs error resolution thing. The exception will throw once and then terminate entirely. With the error we can continue execution and potentially emit other errors in this particular pass. Compilation will still stop once this compiler pass has completed.

backends/p4tools/modules/testgen/targets/bmv2/map_direct_externs.cpp
fruffy-g: Good catch, that was not intentional.

fruffy-g: There are indeed some cases where not returning a declaration instance has a benign cause where we do not throw an error. Iteration should not stop there.

PR #4556, opened 2024-03-24
Suppress Abseil warnings.

### LINKED ISSUE #4523: Protobuf & Abseil compilation pollutes build with a lot of warnings
Compiling with `FetchContent` Abseil and Protobuf, there is *a lot* of warnings. In parallel build they are also interleaved with proper p4c/extensions build which makes readability even worse.

* We should supress warnings in the builds of dependencies, they are not relevant.
* We should make sure the abseil/protobuf include search paths are included as `-isystem` -- some of the warnings are coming from abseil headers included into our code.

fruffy:
Fixes #4523. To be able to capture all Abseil libraries and ignore warnings on them we had to introduce a helper function.


fruffy:
Protobuf warnings will be fixed in #4543.

vlstill (approved):

PR #4557, opened 2024-03-25
Add width to parameters of entries in template file.

komaljai:
Add width to parameters of const entries in template file


backends/tc/tc.def
```diff
@@ -379,9 +379,9 @@ class TCTable {
                 tcTable += "\n$TC p4template create table/" + pipelineName
                             + "/" + controlName + "/" + tableName
                             + " entry" + entry->toString();
+                tcTable += " permissions 0x1024";
```
usha1830, line 382: Are the permissions fixed as 0x1024 for all tables?

usha1830 (approved):
LGTM overall.

backends/tc/tc.def
komaljai: yes

PR #4120, opened 2023-08-27
P4 spec issue 1261 add bounded loops

jafingerhut:

PR #4558, opened 2024-03-25
[WIP] Add example implementation for bounded foreach-style loops

asl:
For #4120 and https://github.com/p4lang/p4-spec/issues/1261

This is an example implementation for foreach-style bounded loops. It contains no tests (as they are written for dowstream backend) and some sema checks are certainly lacking, but still shows the intention. One thing that is currently missed is the use-def changes, I have not ported them to upstream, but will re-add shortly. Certainly changes to other passes might be required (e.g. copyprop).

The PR has some tricks how to handle loop index variable in order to play nicely with the existing code.

Examples of supported code:
```p4
action a0(bit<8> x, bit<8> y) {
        // Reduce
        bit<64> idx = -1;
        foreach (bit<8> i in 0 .. x) {
            foreach (bit<8> j in i .. y) {
                idx = foo(j);
                if (idx == -1)
                    break;
            }
        }
    }

    action a1(bit<8> x, bit<8> y) {
        // Reduce
        bit<64> idx = -1;
        foreach (bit<8> i in 0 .. x) {
            foreach (bit<8> j in 0 .. y) {
                idx = foo(j);
                if (idx == -1) {
                    continue;
                }
                idx = foo(i);
            }
        }
    }
```

### DIFF
--- a/frontends/parsers/p4/p4parser.ypp
+++ b/frontends/parsers/p4/p4parser.ypp
@@ -317,7 +317,7 @@ inline std::ostream& operator<<(std::ostream& out, const P4::Token& t) {
 %type<IR::Vector<IR::Expression>*> strList
 %type<IR::Expression*> intOrStr
 %type<IR::Declaration*> loopIndexDeclaration
-%type<IR::Expression*> rangeExpression
+%type<IR::Range*> rangeExpression
 %type<IR::Statement*> foreachStatement breakStatement continueStatement
 
 // %precedence COMMA
--- a/frontends/parsers/p4/p4parser.ypp
+++ b/frontends/parsers/p4/p4parser.ypp
@@ -231,6 +231,7 @@ inline std::ostream& operator<<(std::ostream& out, const P4::Token& t) {
                    ELSE ENTRIES ENUM ERROR EXIT EXTERN HEADER HEADER_UNION IF IN INOUT
                    INT KEY  LIST SELECT MATCH_KIND TYPE OUT PACKAGE PARSER PRAGMA PRIORITY RETURN
                    STATE STRING STRUCT SWITCH TABLE TRANSITION TUPLE TYPEDEF VARBIT VALUESET VOID
+%token<Token>      FOREACH
 
 %token<cstring> IDENTIFIER TYPE_IDENTIFIER STRING_LITERAL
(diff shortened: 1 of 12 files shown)


asl:
Tagging @jafingerhut and @ChrisDodd

ir/write_context.cpp
```diff
@@ -93,6 +93,9 @@ bool P4WriteContext::isRead(bool root_value) {
         }
     }
     if (ctxt->node->is<IR::IndexedVector<IR::StatOrDecl>>()) return false;
-    if (ctxt->node->is<IR::IfStatement>()) return ctxt->child_index == 0;
+    if (ctxt->node->is<IR::IfStatement>())
+        if (ctxt->node->is<IR::IfStatement>() || ctxt->node->is<IR::ForEachStatement>())
+            return ctxt->child_index == 0;
```
ChrisDodd, line 98: Line 96 here should be removed -- it prevents the `||` in line 97 from doing anything

asl: Nice catch! Yes, merge error.

frontends/parsers/p4/p4parser.ypp
```diff
@@ -231,6 +231,7 @@ inline std::ostream& operator<<(std::ostream& out, const P4::Token& t) {
                    ELSE ENTRIES ENUM ERROR EXIT EXTERN HEADER HEADER_UNION IF IN INOUT
                    INT KEY  LIST SELECT MATCH_KIND TYPE OUT PACKAGE PARSER PRAGMA PRIORITY RETURN
                    STATE STRING STRUCT SWITCH TABLE TRANSITION TUPLE TYPEDEF VARBIT VALUESET VOID
+%token<Token>      FOREACH BREAK CONTINUE
```
asl, line 234: This is structured in a way to simplify review. It could be merged into the token list above

asl:
Another part missed here from downstream: type inference that checks that types of loop index variable is compatible with the range.

fruffy:
Are the changes here merged into #4562? Can this be closed?

ChrisDodd:
> Are the changes here merged into #4562? Can this be closed?

There were minor syntax changes (`for` instead of a new `foreach` keyword), but I believe this can be closed.

PR #4561, opened 2024-03-26
Added default action code for tc backend

Sosutha:
* Added default action code in control block for tc backend

*  Updated testcase outputs


usha1830 (approved):
LGTM

PR #4562, opened 2024-03-26
Changes for for loops

ChrisDodd:
Pulling together changes from #4120 and #4558 for for loops

### DIFF
--- a/backends/bmv2/CMakeLists.txt
+++ b/backends/bmv2/CMakeLists.txt
@@ -63,6 +63,7 @@ set (BMV2_BACKEND_COMMON_HDRS
   common/action.h
   common/annotations.h
   common/backend.h
+  common/check_unsupported.h
   common/control.h
   common/controlFlowGraph.h
   common/deparser.h
--- a/frontends/p4/def_use.cpp
+++ b/frontends/p4/def_use.cpp
@@ -469,8 +469,8 @@ bool ComputeWriteSet::setDefinitions(Definitions *defs, const IR::Node *node, bo
     // overwriting always in parser states.  In this case we actually expect
     // that the definitions are monotonically increasing.
     if (findContext<IR::ParserState>()) overwrite = true;
-    // loop bodies get visited repeatedly until a fixed point, so we likewise
-    // expect monotonically increasing write sets
+    // Loop bodies get visited repeatedly until a fixed point, so we likewise
+    // expect monotonically increasing write sets.
     if (continueDefinitions != nullptr) overwrite = true;  // in a loop
     allDefinitions->setDefinitionsAt(point, currentDefinitions, overwrite);
     if (LOGGING(5))
--- a/frontends/p4/def_use.cpp
+++ b/frontends/p4/def_use.cpp
@@ -257,7 +257,8 @@ bool LocationSet::operator==(const LocationSet &other) const {
     auto it = other.begin();
     for (auto s : locations) {
         if (it == other.end() || *it != s) return false;
-        ++it; }
+        ++it;
+    }
     return it == other.end();
 }
 
--- a/frontends/p4/def_use.cpp
+++ b/frontends/p4/def_use.cpp
@@ -854,6 +854,7 @@ bool ComputeWriteSet::preorder(const IR::ForStatement *statement) {
         visit(statement->body, "body");
         currentDefinitions = currentDefinitions->joinDefinitions(continueDefinitions);
         visit(statement->updates, "updates");
+        currentDefinitions = currentDefinitions->joinDefinitions(startDefs);
     } while (!(*startDefs == *currentDefinitions));
 
     exitDefs = exitDefs->joinDefinitions(breakDefinitions);
@@ -881,12 +882,13 @@ bool ComputeWriteSet::preorder(const IR::ForInStatement *statement) {
         lhs = false;
         auto cond = getWrites(statement->ref);
         auto defs = currentDefinitions->writes(getProgramPoint(), cond);
-        exitDefs = exitDefs->joinDefinitions(defs);
         (void)setDefinitions(defs, statement->ref, true);
         visit(statement->body, "body");
         currentDefinitions = currentDefinitions->joinDefinitions(continueDefinitions);
+        currentDefinitions = currentDefinitions->joinDefinitions(startDefs);
     } while (!(*startDefs == *currentDefinitions));
 
+    exitDefs = exitDefs->joinDefinitions(currentDefinitions);
     exitDefs = exitDefs->joinDefinitions(breakDefinitions);
     breakDefinitions = saveBreak;
     continueDefinitions = saveContinue;
--- a/frontends/p4/def_use.cpp
+++ b/frontends/p4/def_use.cpp
@@ -253,6 +253,14 @@ bool LocationSet::overlaps(const LocationSet *other) const {
     return false;
 }
 
+bool LocationSet::operator==(const LocationSet &other) const {
+    auto it = other.begin();
+    for (auto s : locations) {
+        if (it == other.end() || *it != s) return false;
+        ++it; }
+    return it == other.end();
+}
+
 void ProgramPoints::add(const ProgramPoints *from) {
     points.insert(from->points.begin(), from->points.end());
 }
@@ -406,7 +414,7 @@ void ComputeWriteSet::enterScope(const IR::ParameterList *parameters,
             }
         }
     }
-    allDefinitions->setDefinitionsAt(entryPoint, defs, false);
+    allDefinitions->setDefinitionsAt(entryPoint, defs, true);
     currentDefinitions = defs;
     if (LOGGING(5))
         LOG5("CWS Entered scope " << entryPoint << " definitions are " << Log::endl << defs);
@@ -460,6 +468,9 @@ bool ComputeWriteSet::setDefinitions(Definitions *defs, const IR::Node *node, bo
     // overwriting always in parser states.  In this case we actually expect
     // that the definitions are monotonically increasing.
     if (findContext<IR::ParserState>()) overwrite = true;
+    // loop bodies get visited repeatedly until a fixed point, so we likewise
+    // expect monotonically increasing write sets
+    if (continueDefinitions != nullptr) overwrite = true;  // in a loop
     allDefinitions->setDefinitionsAt(point, currentDefinitions, overwrite);
     if (LOGGING(5))
         LOG5("CWS Definitions at " << point << " are " << Log::endl << defs);
@@ -820,6 +831,67 @@ bool ComputeWriteSet::preorder(const IR::IfStatement *statement) {
     return setDefinitions(result);
 }
 
+bool ComputeWriteSet::preorder(const IR::ForStatement *statement) {
+    LOG3("CWS Visiting " << dbp(statement));
+    if (currentDefinitions->isUnreachable()) return setDefinitions(currentDefinitions);
+    visit(statement->init, "init");
+
+    auto saveBreak = breakDefinitions;
+    auto saveContinue = continueDefinitions;
+    breakDefinitions = new Definitions();
+    continueDefinitions = new Definitions();
+    Definitions *startDefs = nullptr;
+    Definitions *exitDefs = nullptr;
+
+    do {
+        startDefs = currentDefinitions;
+        visit(statement->condition, "condition");
+        auto cond = getWrites(statement->condition);
+        // exitDefs are the definitions after evaluating the condition
+        exitDefs = currentDefinitions->writes(getProgramPoint(), cond);
+        (void)setDefinitions(exitDefs, statement->condition, true);
+        visit(statement->body, "body");
+        currentDefinitions = currentDefinitions->joinDefinitions(continueDefinitions);
+        visit(statement->updates, "updates");
+    } while (!(*startDefs == *currentDefinitions));
+
+    exitDefs = exitDefs->joinDefinitions(breakDefinitions);
+    breakDefinitions = saveBreak;
+    continueDefinitions = saveContinue;
+    return setDefinitions(exitDefs);
+}
+
+bool ComputeWriteSet::preorder(const IR::ForInStatement *statement) {
+    LOG3("CWS Visiting " << dbp(statement));
+    if (currentDefinitions->isUnreachable()) return setDefinitions(currentDefinitions);
+    visit(statement->collection, "collection");
+
+    auto saveBreak = breakDefinitions;
+    auto saveContinue = continueDefinitions;
+    breakDefinitions = new Definitions();
+    continueDefinitions = new Definitions();
+    Definitions *startDefs = nullptr;
+    Definitions *exitDefs = currentDefinitions;  // in case collection is empty;
+
+    do {
+        startDefs = currentDefinitions;
+        lhs = true;
+        visit(statement->ref, "ref");
+        lhs = false;
+        auto cond = getWrites(statement->ref);
+        auto defs = currentDefinitions->writes(getProgramPoint(), cond);
+        exitDefs = exitDefs->joinDefinitions(defs);
+        (void)setDefinitions(defs, statement->ref, true);
+        visit(statement->body, "body");
+        currentDefinitions = currentDefinitions->joinDefinitions(continueDefinitions);
+    } while (!(*startDefs == *currentDefinitions));
+
+    exitDefs = exitDefs->joinDefinitions(breakDefinitions);
+    breakDefinitions = saveBreak;
+    continueDefinitions = saveContinue;
+    return setDefinitions(exitDefs);
+}
+
 bool ComputeWriteSet::preorder(const IR::BlockStatement *statement) {
     if (currentDefinitions->isUnreachable()) return setDefinitions(currentDefinitions);
     visit(statement->components, "components");
@@ -827,27 +899,35 @@ bool ComputeWriteSet::preorder(const IR::BlockStatement *statement) {
 }
 
 bool ComputeWriteSet::preorder(const IR::ReturnStatement *statement) {
+    if (currentDefinitions->isUnreachable()) return setDefinitions(currentDefinitions);
     if (statement->expression != nullptr) visit(statement->expression);
-    returnedDefinitions = returnedDefinitions->joinDefinitions(currentDefinitions);
-    if (LOGGING(5))
-        LOG5("Return definitions " << returnedDefinitions);
-    else
-        LOG3("Return " << returnedDefinitions->size() << " definitions");
-    auto defs = currentDefinitions->cloneDefinitions();
-    defs->setUnreachable();
-    return setDefinitions(defs);
+    return handleJump("Return", returnedDefinitions);
 }
 
 bool ComputeWriteSet::preorder(const IR::ExitStatement *) {
     if (currentDefinitions->isUnreachable()) return setDefinitions(currentDefinitions);
-    exitDefinitions = exitDefinitions->joinDefinitions(currentDefinitions);
+    return handleJump("Exit", exitDefinitions);
+}
+
+bool ComputeWriteSet::preorder(const IR::BreakStatement *) {
+    if (currentDefinitions->isUnreachable()) return setDefinitions(currentDefinitions);
+    return handleJump("Break", breakDefinitions);
+}
+
+bool ComputeWriteSet::preorder(const IR::ContinueStatement *) {
+    if (currentDefinitions->isUnreachable()) return setDefinitions(currentDefinitions);
+    return handleJump("Continue", continueDefinitions);
+}
+
+bool ComputeWriteSet::handleJump(const char *tok, Definitions *&defs) {
+    defs = defs->joinDefinitions(currentDefinitions);
     if (LOGGING(5))
-        LOG5("Exit definitions " << exitDefinitions);
+        LOG5(tok << " definitions " << defs);
     else
-        LOG3("Exit with " << exitDefinitions->size() << " definitions");
-    auto defs = currentDefinitions->cloneDefinitions();
-    defs->setUnreachable();
-    return setDefinitions(defs);
+        LOG3(tok << " with " << defs->size() << " definitions");
+    auto after = currentDefinitions->cloneDefinitions();
+    after->setUnreachable();
+    return setDefinitions(after);
 }
 
 bool ComputeWriteSet::preorder(const IR::EmptyStatement *) {
--- a/frontends/p4/def_use.h
+++ b/frontends/p4/def_use.h
@@ -477,12 +477,12 @@ class AllDefinitions : public IHasDbPrint {
 
 class ComputeWriteSet : public Inspector, public IHasDbPrint {
  protected:
-    AllDefinitions *allDefinitions;    /// Result computed by this pass.
-    Definitions *currentDefinitions;   /// Before statement currently processed.
-    Definitions *returnedDefinitions;  /// Definitions after return statements.
-    Definitions *exitDefinitions;      /// Definitions after exit statements.
-    Definitions *breakDefinitions = nullptr;
-    Definitions *continueDefinitions = nullptr;
+    AllDefinitions *allDefinitions;              /// Result computed by this pass.
+    Definitions *currentDefinitions;             /// Before statement currently processed.
+    Definitions *returnedDefinitions;            /// Definitions after return statements.
+    Definitions *exitDefinitions;                /// Definitions after exit statements.
+    Definitions *breakDefinitions = nullptr;     /// Definitions at break statements.
+    Definitions *continueDefinitions = nullptr;  /// Definitions at continue statements.
     ProgramPoint callingContext;
     const StorageMap *storageMap;
     /// if true we are processing an expression on the lhs of an assignment
--- a/frontends/p4/def_use.h
+++ b/frontends/p4/def_use.h
@@ -231,6 +231,7 @@ class LocationSet : public IHasDbPrint {
     }
     // only defined for canonical representations
     bool overlaps(const LocationSet *other) const;
+    bool operator==(const LocationSet &other) const;
     bool isEmpty() const { return locations.empty(); }
 };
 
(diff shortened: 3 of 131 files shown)


ir/ir.def
```diff
@@ -500,6 +510,40 @@ class SwitchStatement : Statement {
...
+        v.visit(condition, "condition");
+        // FIXME -- need visit closure over the body and updates + condition
+        v.visit(body, "body");
+        v.visit(updates, "updates"); }
+}
+
+// FIXME -- should we try to directly convert this to a ForStatement in the parser?
+class ForInStatement : Statement, ISimpleNamespace {
+    NullOK Declaration_Variable         decl;
+    // the declaration will be moved to the enclosing block by MoveDeclarations, but
```
asl, line 533: I do not recall all the details, but iirc the change in MoveDeclarations was necessary for nested loops like + some shadowing:
```
action a0(bit<8> x, bit<8> y) {
        // Reduce
        bit<64> idx = -1;
        foreach (bit<8> i in 0 .. x) {
            foreach (bit<8> j in i .. y) {
                idx = foo(j);
                if (idx == -1)
                    break;
            }
        }
    }
```

asl:
@ChrisDodd Consider the following P4 code:

```p4
bit<64> _popcount(in bit<64> val) {
    bit<64> n = 0;
    bit<64> v = val;
    for (bit<64> popcnti in 1 .. 64w63) {
        if (v == 0)
            return n;
        n = n + 1;
        v = v & (v - 1);
    }
    return n;
}
```

There are two issues with existing PR:
 1. Warning about `popcnti` being uninitialized.
 2. `v = v & (v - 1);` is removed by simplifyUseDef (`n = n + 1` is not removed as `return n` is considered as a valid use of that def).

My solution for 1. was factoring out loop index variable and loop container into a separate `ForEachHeader` node. So we can explicitly make it read and write loop index variable and has read index variable defined after header node.

For 2. I marked all defs in the loop body to be used by loop itself.

But maybe you might have some different solution for this issue as you know the underlying code much better than myself :)

ChrisDodd:
I've added support for loops for frontend DefUse, midend DefUse, midend localCopyprop and general ControlFlowVisitor use.  Midend Defuse uses that generic ControlFlowVisitor, but it turned out localCopyprop didn't work properly for that, so it needed an additional stuff -- basically due ti the fact that it did analysis and transform in one pass, it could not use an iterative loop closure, so instead needed one analysis pass of the loop body to remove potential copies that are dependent on the loop body, followed by the normal analysis+rewrite pass to to copyprop into and within the loop body.

At this point, the simple test cases I've looked at look good, though it is still giving the spurious "uninitialized variable" for the loop var in for..in loops.  There are a number of hacky things in a coupld of places that probably could be done better too.

ChrisDodd:
> There are two issues with existing PR:
> 
> 1. Warning about `popcnti` being uninitialized.
> 2. `v = v & (v - 1);` is removed by simplifyUseDef (`n = n + 1` is not removed as `return n` is considered as a valid use of that def).

I think issue 2 was likely coming from local copyprop -- I've fixed that so it doesn't incorrectly propagate into the loop the first time through.  The warning is still an issue I need to look into, though.

asl:
> I think issue 2 was likely coming from local copyprop -- I've fixed that so it doesn't incorrectly propagate into the loop the first time through. The warning is still an issue I need to look into, though.

@ChrisDodd Both issues are from `simplifyUseDef`. The warning appears because there is `use` of loop index variable before real def (it's in `beforeStart` state). As for the second – there is no `use` of `v` after the loop and therefore the last def (assignment) is considered redundant and therefore is removed.

ChrisDodd:
> > I think issue 2 was likely coming from local copyprop -- I've fixed that so it doesn't incorrectly propagate into the loop the first time through. The warning is still an issue I need to look into, though.
> 
> @ChrisDodd Both issues are from `simplifyUseDef`. The warning appears because there is `use` of loop index variable before real def (it's in `beforeStart` state). As for the second – there is no `use` of `v` after the loop and therefore the last def (assignment) is considered redundant and therefore is removed.

The warning is coming from FindUninitialized, which does its own flow analysis independent of DefUse (and likewise not using the builting visitor infrastructure for flow analysis), so needs to be updated for loops.  But I don't see a problem with the `v =`... being removed -- it is still there after SimplifyDefUse, and also still there after LocalCopyprop (now).  Updating frontend DefUse for loops (which SimplifyDefUse uses for removals) liekly fixed that.

asl:
> But I don't see a problem with the `v =`... being removed -- it is still there after SimplifyDefUse, and also still there after LocalCopyprop (now). Updating frontend DefUse for loops (which SimplifyDefUse uses for removals) liekly fixed that.

Oh, indeed, you are right. I was looking into an old set of dumps. With your latest changes to frontend SimplifyDefUse indeed it is not removed anymore. So, this is a viable solution that could subsume what we are having downstream.

> The warning is coming from FindUninitialized, which does its own flow analysis independent of DefUse (

I believe it does use DefUse, e.g. in `registerUses`: https://github.com/p4lang/p4c/blob/main/frontends/p4/simplifyDefUse.cpp#L1011:
```c++
        LOG3("LocationSet for '" << expression << "' is <<" << read << ">>");

        auto points = currentDefinitions->getPoints(read);

        if (reportUninitialized && !lhs && points->containsBeforeStart() &&
            hasUninitializedHeaderUnion(expression, currentDefinitions, read)) {
...
```

Here the points for loop index variable is `BeforeStart` (while recording the uses for loop index variable PathExpression):
```
  FU Registering uses for 'popcnti_0/popcnti;'
  FU Current point is (after) <AssignmentStatement>(600624)[[v_0/v = val;]] definitions are
    56 val/val=>{<Function>(614279) _popcount/2455 }
    57 n_0/n=>{<AssignmentStatement>(600618)[[n_0/n = 0;]] }
    58 v_0/v=>{<AssignmentStatement>(600624)[[v_0/v = val;]] }
    59 popcnti_0/popcnti=>{<BeforeStart> }
  LocationSet for 'popcnti_0/popcnti;' is <<59 popcnti_0/popcnti >>
```
hence the warning.

My solution was to introduce `ForEachHeader` node that does both def and use of loop index variable. So after this node we have proper def for loop index variable (and have to "silence" a warning only only in that single case).

asl:
@ChrisDodd I am seeing null checks failed at https://github.com/p4lang/p4c/pull/4562/files#diff-6eaf37d31585aaaa76257fcc3f96a3d523ad39cd3adb88a225389df8b60814c8R959:
```
Compiler Bug: p4c/frontends/p4/def_use.h:204: Null location
```
(line 204 downstream is `LocationSet::add`).

So, apparently, `storageMap` contains empty location for the loop index declaration. The code in question is:
```p4
    action a0(bit<8> x, bit<8> y) {
        bit<64> idx = -1;
        foreach (bit<8> i1 in 0 .. x) {
            foreach (bit<8> j in i1 .. y) {
                idx = foo(j);
                if (idx == -1)
                    break;
            }
        }
    }
```

ChrisDodd:
> @ChrisDodd I am seeing null checks failed at https://github.com/p4lang/p4c/pull/4562/files#diff-6eaf37d31585aaaa76257fcc3f96a3d523ad39cd3adb88a225389df8b60814c8R959:
> 
> ```
> Compiler Bug: p4c/frontends/p4/def_use.h:204: Null location
> ```

I don't see any Compiler Bugs with this code but I do see spurious "maybe uninitialized" warnings for i1 and j -- turns out there was a small bug in def_use for ForInStatements setting reachable defs for the body of the loop (was unioning in the state from before the loop, which should only be done for after the loop, not the loop body itself), and I'd missed a place in FindUninitialized setting the current point after the condition in ForIn loops.   Things would be much simpler if these passes used the ControlFlowVisitor infrastructure which handles all this in one place, rather than having to reimplement it for every pass.

I've added your two testcase here, but more testcases would be nice

asl:
> I've added your two testcase here, but more testcases would be nice

Let me see what I'm having downstream. I cannot submit them directly as they use private packages, etc. So, I will extract one-by-one

asl:
Here are the tests:

```p4
bit<8> foo(in bit<8> aa, in bit<8> bb) {
   // Do some magic to ensure aa / bb will not be simplified

    return aa - bb;
}

    action a0(bit<8> x, bit<8> y) {
        // Simple loops: with and w/o use of index variable.
        foreach (bit<8> i in 1 .. (x+y)) {
            foo(x, y);
        }

        foreach (bit<8> i in 1 .. (x+y)) {
            foo(x, y + i);
        }

        // And with constant range as well
        foreach (bit<8> i in 8w1 .. 42) {
            foo(x, y + i);
        }
    }

    action a1(bit<8> x, bit<8> y) {
        // Nested loops
        foreach (bit<8> i in 1 .. x) {
          foreach (bit<8> j in i .. y)
            foo(x, y + i);
        }
    }

     // Scope test    
    action a2(bit<8> x, bit<8> y) {
        bit<8> i = 10;
        foreach (bit<8> i in 1 .. x) {
            foo(x, y + i);
        }

        foreach (bit<8> k in i .. (x+y)) {
          foo(i, 2*i+x);
        }
    }
```

Same as `break` testcase above but with continue: 
```p4
    action a1(bit<8> x, bit<8> y) {
        bit<64> idx = -1;
        foreach (bit<8> i in 0 .. x) {
            foreach (bit<8> j in 0 .. y) {
                idx = foo(j);
                if (idx == -1) {
                    continue;
                }
                idx = foo(i);
            }
        }
    }
```

I am afraid all other testcases are much tighter integrated downstream

asl:
> turns out there was a small bug in def_use for ForInStatements setting reachable defs for the body of the loop (was unioning in the

Have you pushed these fixes? I do see only tests were added...

ChrisDodd:
Ok, there were some more issues with the flow analysis over loops that wer breaking the forloop2 test, so I've fixed that too.

Flow analysis over loops is non-trivial with a  bunch of corner cases, so it is particularly good to have it in one place in the infrastructure rather than having to reimplement it for every visitor.

ChrisDodd:
I'm still working on code to unroll loops to support targets that don't support loops natively, but this is already a pretty large change and has all the necessary support for targets that do suppoert loops.

ChrisDodd:
So loops in an action like this:

>     action a1(bit<8> x, bit<8> y) {
>         // Nested loops
>         foreach (bit<8> i in 1 .. x) {
>           foreach (bit<8> j in i .. y)
>             foo(x, y + i);
>         }
>     }

where the loop count depends on an action data parameter (rather than a constant) can't be easily bounded (other than by the possible range of the type, so 255 here), so don't seem to fit in with the intent of https://github.com/p4lang/p4-spec/issues/1261 or https://github.com/p4lang/p4c/pull/4120.  However, they do seem fine with general loops, which this PR supports.  Unrolling this loop seems impractical, and would really require a different action for each possible loop count, deciding on which one to use based on the actual parameter value of any particular entry.

testdata/p4_16_samples/forloop1.p4
```diff
@@ -0,0 +1,32 @@
...
+}
+
+struct headers_t {
+    t1          head;
+    t2[8]       stack;
+}
+
+control c(inout headers_t hdrs) {
+    apply {
+        for (t2 v in hdrs.stack) {
```
fruffy, line 23: @asl Any reason for `in` as opposed to C++ style foreach `:`? This way we can avoid introducing a new keyword. Is this emulating Rust-style? Looks a little odd with the type specifier.

frontends/p4/def_use.cpp
```diff
@@ -460,6 +469,9 @@ bool ComputeWriteSet::setDefinitions(Definitions *defs, const IR::Node *node, bo
     // overwriting always in parser states.  In this case we actually expect
     // that the definitions are monotonically increasing.
     if (findContext<IR::ParserState>()) overwrite = true;
+    // loop bodies get visited repeatedly until a fixed point, so we likewise
+    // expect monotonically increasing write sets
```
fruffy, line 473: Nit: 
```suggestion
    // Loop bodies get visited repeatedly until a fixed point, so we likewise
    // expect monotonically increasing write sets.
```

frontends/p4/def_use.h
```diff
@@ -480,6 +481,8 @@ class ComputeWriteSet : public Inspector, public IHasDbPrint {
     Definitions *currentDefinitions;   /// Before statement currently processed.
     Definitions *returnedDefinitions;  /// Definitions after return statements.
     Definitions *exitDefinitions;      /// Definitions after exit statements.
+    Definitions *breakDefinitions = nullptr;
```
fruffy, line 484: Missing comments for these definitions.

frontends/p4/validateParsedProgram.cpp
```diff
@@ -251,4 +251,17 @@ void ValidateParsedProgram::postorder(const IR::Dots *dots) {
     }
 }
 
+/// Check that continue and break statements are only used in the context of a for statement
```
fruffy, line 254:
```suggestion
/// Check that continue and break statements are only used in the context of a for statement.
```

ir/visitor.h
```diff
@@ -479,6 +481,8 @@ class ControlFlowVisitor : public virtual Visitor {
     friend void dump(const flow_join_points_t &);
     friend void dump(const flow_join_points_t *);
 
+    bool unreachable = false;
```
fruffy, line 484: Clarifying comment? What is unreachable?

ir/visitor.h
```diff
@@ -65,8 +65,8 @@ struct Visitor_Context {
     }
 };
 
+class ControlFlowVisitor;
```
fruffy, line 68: Why not declare a new visitor class that is specialized to Loops instead of extending IR::Visitor? Seems like some dirty abstractions here. Are the extensions only relevant to ControlflowVisitors or do all visitors need to support them?

fruffy (commented):
Totally out of the loop on this, giving just some high-level comments. 

Are these loops meant to be bounded or unbounded? If bounded we should definitely have a pass unrolling them.

This is definitely something that should be fuzzed/translation-validated. Wish I had the time to extend the validator..

testdata/p4_16_samples/forloop1.p4
asl: Downstream we used `foreach (type foo in range)` syntax, but @ChrisDodd changed to `for` :) 

My idea was to use entirely new keyword to explicitly deviate from C / C++ syntax to outline semantic differences. I'm fine with C++-style range for syntax just in case.

asl:
> Are these loops meant to be bounded or unbounded? If bounded we should definitely have a pass unrolling them.

I guess the key question is what is the definition of "bounded". There might be different possibilities:
 - The loop range could be fully evaluated to a compile-time constant (therefore things like action parameters cannot be used in loop bounds)
 - Loop range cannot be fully evaluated to a compile-time constant, but could be formed into a well-defined polyhedron with boundaries possibly depending on things like action parameters. This could be seen as "bounded" in some sense as we can always provide a useful bound (e.g. from `0` to `x`)
 - Something else

Polyhedral representation is more or less standard in various loop optimization techniques as it could allow different transformations like partial unrolling, peeling, etc.

It is much easier to enforce any sane kind on boundness for `foreach`-style loops. And much harder for C / C++-style `for` loops as condition and loop increment expressions could be quite arbitrary...

testdata/p4_16_samples/forloop1.p4
ChrisDodd: `in` is already keyword in P4, so does not constitute a new keyword, just repurposing an existing one.  Using `for` instead of `foreach` is avoiding a new keyword, and is what was in Andy's original change.

ir/visitor.h
ChrisDodd: ControlFlowVisitor is a mixin for doing control flow analysis over the IR, so it needs to be aware of all nodes that a relevant to control flow (loops and and otherwise).  This is a forward declaration so we can have a virtual method Visitor::controlFlowVisitor() that does a dynamic_cast<ControlFlowVisitor>, but more efficiently (Avoid the gcc/clang optimization bug that dynamic_cast is much slower than a virtual function call when it should not be).

ChrisDodd:
@asl @fruffy @thantry and anyone else with interest -- there will likely be some discussion of this in the next P4 Language working group meeting on May 6 (https://p4.org/working-groups/)

fruffy:
> @asl @fruffy @thantry and anyone else with interest -- there will likely be some discussion of this in the next P4 Language working group meeting on May 6 (https://p4.org/working-groups/)

That will be helpful to understand how the current specification and this PR correlate.

ir/visitor.h
```diff
@@ -518,12 +529,19 @@ class ControlFlowVisitor : public virtual Visitor {
      * edge are never join points.
      */
     virtual bool filter_join_point(const IR::Node *) { return false; }
-    ControlFlowVisitor &flow_clone() override;
-    void flow_merge(Visitor &) override = 0;
-    virtual void flow_copy(ControlFlowVisitor &) = 0;
     ControlFlowVisitor() : globals(*new std::map<cstring, ControlFlowVisitor &>) {}
 
  public:
+    ControlFlowVisitor *controlFlowVisitor() { return this; }
```
asl, line 535:
```suggestion
    ControlFlowVisitor *controlFlowVisitor() override { return this; }
```

testdata/p4_16_samples_outputs/forloop2-frontend.p4
```diff
@@ -0,0 +1,47 @@
...
+                ;
+            } else {
+                n_0 = n_0 + 64w1;
+                v_0 = v_0 & v_0 + 64w18446744073709551615;
+            }
+        }
+        if (hasReturned) {
+            ;
+        } else {
+            retval = n_0;
```
asl, line 41: it is strange we do not have `hasReturned = true` here. Is this expected?

testdata/p4_16_samples_outputs/forloop2.p4-stderr
```diff
@@ -0,0 +1,2 @@
+[--Wwarn=uninitialized_use] warning: retval may be uninitialized
```
asl, line 1: The warning is misleading. First of all, there is already a value returned. Second, there is no `retval` in the source code. `retval` is just internal symbol name inserted during lowering and therefore should not be exposed.

testdata/p4_16_samples_outputs/forloop2-frontend.p4
ChrisDodd: This is because the frontend does RemoveReturns (a prereq for inlining), but does not do anything with loops.  Would be better to not do all that in the frontend.  See #4634

testdata/p4_16_samples_outputs/forloop2.p4-stderr
ChrisDodd: Yes, I'm not sure where the warning is coming from -- I'll look into it.

asl: I think these are connected (the previous `hasReturned` thing). Downstream we do not have these, however, we are having some decent amount of fixes (and we do not rely on `hasReturn` in general). Though, I have not yet found what was the fix and why we have not upstreamed it...

ChrisDodd: The warning comes from a combination of the code produced by RemoveReturns and the flow analysis code -- the latter does not do value correlation, so it finds an impossible path through the code where it thinks the `retval` introduced by RemoveReturns might be uninitialized.  Improving flow analysis to also do value analysis (maybe just for `bool` vars) would be major work.

We could have RemoveReturns always initialize `retval`.  Or we could have it mark retval with an annotation that would supress the warning.

ChrisDodd: This problem doesn't have anything specific to do with loops -- it can happen for any function that contains multiple `return` statements that gets inlined.

lib/log.h
```diff
@@ -113,6 +113,7 @@ void increaseVerbosity();
     (LOGGING(N) ? ::Log::Detail::fileLogOutput(__FILE__)                  \
                       << ::Log::Detail::OutputLogPrefix(__FILE__, N) << X \
                       << ::Log::Detail::clearPrefix << std::endl          \
+                      << std::flush                                       \
```
fruffy, line 116: Doesn't `std::endl` already flush?
https://en.cppreference.com/w/cpp/io/manip/endl

fruffy: https://stackoverflow.com/a/50707073

ChrisDodd: Not sure how that crept in there -- I think at one point I was having an issue with log messages not coming out as I expected and was tweaking things to try to make it work, but that should not have been part of this change

frontends/p4/removeReturns.cpp
```diff
@@ -127,14 +127,15 @@ const IR::Node *DoRemoveReturns::preorder(IR::ReturnStatement *statement) {
...
     vec->push_back(
         new IR::AssignmentStatement(statement->srcInfo, left, new IR::BoolLiteral(true)));
     if (statement->expression != nullptr) {
-        left = new IR::PathExpression(returnedValue);
+        left = new IR::PathExpression(statement->expression->type, returnedValue);
         vec->push_back(
             new IR::AssignmentStatement(statement->srcInfo, left, statement->expression));
     }
+    if (findContext<IR::LoopStatement>()) vec->push_back(new IR::BreakStatement);
     return new IR::BlockStatement(*vec);
```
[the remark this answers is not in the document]
ChrisDodd: Hmm, yes a nested loop might require an extra check and break to skip additional iterations of the outer loop.  There's also some potential issues with break/continue messing up the flow analysis done in removeReturns, causing it to miss potential places it should be checking the hasReturned flag.  Defintely need more test cases.

midend/removeExits.cpp
```diff
@@ -55,8 +55,11 @@ void DoRemoveExits::callExit(const IR::Node *node) {
 
 const IR::Node *DoRemoveExits::preorder(IR::ExitStatement *statement) {
     set(TernaryBool::Yes);
-    auto left = new IR::PathExpression(returnVar);
-    return new IR::AssignmentStatement(statement->srcInfo, left, new IR::BoolLiteral(true));
+    auto left = new IR::PathExpression(IR::Type::Boolean::get(), returnVar);
+    auto trueVal = new IR::BoolLiteral(true);
+    IR::Statement *rv = new IR::AssignmentStatement(statement->srcInfo, left, trueVal);
+    if (findContext<IR::LoopStatement>()) rv = new IR::BlockStatement({rv, new IR::BreakStatement});
+    return rv;
```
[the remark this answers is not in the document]
ChrisDodd: removeExits mostly reuses removeReturns (inherits from it), so as long as that deals with this case it should be fine.

frontends/p4/validateParsedProgram.cpp
```diff
@@ -251,4 +251,17 @@ void ValidateParsedProgram::postorder(const IR::Dots *dots) {
...
+    if (!findContext<IR::ForStatement>() && !findContext<IR::ForInStatement>())
+        ::error(ErrorType::ERR_INVALID,
+                "%1%: break statement must be used in the context of a for statement.", s);
+}
+
+void ValidateParsedProgram::postorder(const IR::ContinueStatement *s) {
+    if (!findContext<IR::ForStatement>() && !findContext<IR::ForInStatement>())
+        ::error(ErrorType::ERR_INVALID,
+                "%1%: continue statement must be used in the context of a for statement.", s);
+}
```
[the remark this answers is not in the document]
ChrisDodd: Added a simple test

frontends/p4/removeReturns.cpp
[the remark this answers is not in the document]
ChrisDodd: For nested loops with a return/exit in the inner loop, it needs to add an additional conditional break in the outer loop.  I added that and added a testcase (forloop7.p4)

jafingerhut (approved):
My review is based primarily upon the contents of the new test P4 programs and their expected output files, as I am not a good person for reviewing changes to the internals of p4c.  I am glad to see others reviewing this PR, and please do not take my approval as any kind of dismissal of other outstanding comments on this PR.

testdata/p4_16_samples/forloop2.p4
```diff
@@ -0,0 +1,31 @@
...
+
+struct headers_t {
+    t1          t1;
+}
+
+bit<64> _popcount(in bit<64> val) {
+    bit<64> n = 0;
+    bit<64> v = val;
+    for (bit<64> popcnti in 1 .. 64w63) {
+        if (v == 0)
```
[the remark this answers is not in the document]
ChrisDodd: Yes -- I added a change to TypeInference that gets this (and an error test to go with it).

frontends/p4/typeChecking/typeChecker.cpp
```diff
@@ -4149,6 +4153,28 @@ const IR::Node *TypeInference::postorder(IR::AssignmentStatement *assign) {
...
+    LOG3("TI Visiting " << dbp(getOriginal()));
+    auto ltype = getType(forin->ref);
+    if (ltype == nullptr) return forin;
+
+    if (!isLeftValue(forin->ref)) {
+        typeError("Expression %1% cannot be the target of an assignment", forin->ref);
+        LOG2(forin->ref);
+        return forin;
+    }
+    if (auto range = forin->collection->to<IR::Range>()) {
```
kfcripps, line 4166: Question: Is it possible for `collection` to not be `IR::Range`?

asl: It seems so: https://github.com/p4lang/p4c/pull/4562/files#diff-85fbe45d422156081940f1eb7e38cd4ff1bd03574da101de1cf7021a166851c9R1349

kfcripps: Then is it possible to also check the type for non-range `collection`s?

ChrisDodd: Yes, there should be some other checks here for that case, but I'm still unsure as to exactly the full scope of iterations over 'collections'.  Need support for `list<T>` and `header_stack` at least; maybe other things as well?

asl: maybe we can do the opposite and explicitly require a `Range` here? lists and header stacks are so different conceptually here

backends/bmv2/CMakeLists.txt
zzmic, line 1: Hi, @ChrisDodd. I was wondering what compiler you've used to compile the examples you provided (e.g., ./testdata/p4_16_samples/forloop1.p4) since it's apparent that "BMV2 does not support loops." Thanks!

ChrisDodd: The p4test compiler compiles these loops

zzmic: Thanks for clarifying!

PR #4568, opened 2024-03-27
Implement Consistent Commenting Style in backend

### LINKED ISSUE #4567: Format Comments Consistently in P4C.
This issue aims to address the lack of consistency in comment formatting across the core parts of the P4C compiler. Following a discussion in #4563, it has been suggested to adopt a unified commenting style: using triple slashes (///) for documenting functions and classes, and double slashes (//) for comments within code. This approach aligns with the recommendation provided in the Doxygen documentation.

sumit7754:
Fixes #4567 

This pull request addresses the issue #4567, which discusses the need for a consistent commenting style in the core parts of the compiler. Following the discussions in the issue, I have made the following changes:

Used triple slashes (///) for commenting functions and classes.
Used double slashes (//) for comments within the code.


AdarshRawat1:
I am not sure if this was decided for backends as well.

backends/tc/pnaProgramStructure.cpp
```diff
@@ -217,7 +217,7 @@ void InspectPnaProgram::postorder(const IR::P4Control *c) {
 }
 
 bool ParsePnaArchitecture::preorder(const IR::ToplevelBlock *block) {
-    /// Blocks are not in IR tree, use a custom visitor to traverse
+    // Blocks are not in IR tree, use a custom visitor to traverse
```
AdarshRawat1, line 220: Unintended change (I guess)

backends/bmv2/common/JsonObjects.cpp
```diff
@@ -105,14 +104,14 @@ void JsonObjects::add_meta_info() {
...
- * @param fields a JsonArray for the fields in the header
- */
+
+ /// Create a header type in json.
+ /// @param name header name
+ /// @param type header type
+ /// @param max_length  maximum length for a header with varbit fields;
+ ///                    if 0 header does not contain varbit fields
+ /// @param fields a JsonArray for the fields in the header
+
```
fruffy, line 114: You added an additional space here between the comment and the C++ code. Please remove it for all instances.

backends/bmv2/common/controlFlowGraph.cpp
```diff
@@ -166,7 +165,7 @@ bool CFG::checkImplementable() const {
 namespace {
 class CFGBuilder : public Inspector {
     CFG *cfg;
-    /// predecessors of current CFG node
```
fruffy, line 169: This is not correct. Members in header files also will have triple slashes. We only use double slashes for "internal" comments within functions, for example.

fruffy: Please fix this for all instances.

fruffy (changes requested):
Thank you for your effort, this is a great start! Some initial comments.

backends/bmv2/common/controlFlowGraph.cpp
sumit7754: /// Create a header type in json.
 /// @param name header name
 /// @param type header type
 /// @param max_length  maximum length for a header with varbit fields;
 ///                    if 0 header does not contain varbit fields
 /// @param fields a JsonArray for the fields in the header

Are you referring to a gap in these comments?

fruffy: You can format Github comments using triple '`', it will make your comment more readable: https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/creating-and-highlighting-code-blocks

Yes, in all the comments where there is a newline between the comments and the actual function.

backends/bmv2/common/action.cpp
```diff
@@ -213,4 +212,4 @@ void ActionConverter::postorder(const IR::P4Action *action) {
     ctxt->structure->ids.emplace(action, id);
 }
 
-}  // namespace BMV2
```
sumit7754, line 216: for the namespace should I use /// or // ?

fruffy: They are internal, use two slashes here.

fruffy: Actually, another helpful contribution is to write this all down. Very good to have! We should have done this much earlier.

sumit7754: Sure we can do this.

sumit7754: If you elaborate more then that will be helpful

fruffy: Basically, all these details we have discussed (comment style for namespaces, triple slashes vs star comments, newlines before functions) should be written down in text. This way we can refer to our "style document" whenever these questions come up.

sumit7754: /**
     * Convert an expression into JSON
     * @param e   expression to convert
     * @param doFixup  Insert masking operations for operands to ensure that the result
     *                 matches the specification.  BMv2 does arithmetic using unbounded
     *                 precision, but the spec requires fixed precision, specified by the types.
     * @param wrap     Wrap the result into an additiona JSON expression block.
     *                 See the BMv2 JSON spec.
     * @param convertBool  Wrap the result into a cast from boolean to data (b2d JSON).
     */  For this type of comments, what kind of formatting should we use?

fruffy: Triple slashes.

fruffy:
Closing this in favor of smaller pull requests per back end folder. Like #4586.

PR #4575, opened 2024-03-29
[Added] P4 Naming Conventions

Abhinavcode13:
This PR introduces and documents the naming conventions for various elements in P4, including :

- built-in types
- user-defined types
- type variables
- variables
- constants
- errors

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -293,6 +293,20 @@ To add a new input test with a sample P4 code file (under `testdata/p4_16_sample
   * Use `ordered_map` and `ordered_set` when you need to iterate;
     they provide deterministic iterators.
 
+## Naming conventions
+
+P4 provides a rich assortment of types. Base types include bit-strings, numbers, and errors. There are also built-in types for representing constructs such as parsers, pipelines, actions, and tables. Users can construct new types based on these: structures, enumerations, headers, header stacks, header unions, etc.
+
+In this document we adopt the following conventions:
+
+- Built-in types are written with lowercase characters—e.g., `int<20>`,
+- User-defined types are capitalized—e.g., `IPv4Address`,
+- Type variables are always uppercase—e.g., parser `P<H, IH>(...)`,
+- Variables are uncapitalized— e.g., `ipv4header`,
+- Constants are written with uppercase characters—e.g., `CPU_PORT`, and
+- Errors and enumerations are written in camel-case— e.g. `PacketTooShort`.
+    
+
 ## Compiler Driver
 
 **p4c** is a compiler driver. The goal is to provide a consistent user interface


docs/README.md
```diff
@@ -293,6 +293,19 @@ To add a new input test with a sample P4 code file (under `testdata/p4_16_sample
...
+## Naming conventions
+
+P4 provides a rich assortment of types. Base types include bit-strings, numbers, and errors. There are also built-in types for representing constructs such as parsers, pipelines, actions, and tables. Users can construct new types based on these: structures, enumerations, headers, header stacks, header unions, etc.
+
+In this document we adopt the following conventions:
+
+- Built-in types are written with lowercase characters—e.g., `int<20>`,
+- User-defined types are capitalized—e.g., `IPv4Address`,
+- Type variables are always uppercase—e.g., parser `P<H, IH>(...)`,
+- Variables are uncapitalized— e.g., `ipv4header`,
```
AdarshRawat1, line 305: This convention can make reading variable names difficult 

example - 

`maxshiftwidth`   for  `maxShiftWidth` 
https://github.com/p4lang/p4c/blob/f8df27f677a2eedac6357139f8d40828763af75f/backends/bmv2/common/lower.h#L33

`typeargnum`  for `typeArgNum`
https://github.com/p4lang/p4c/blob/f8df27f677a2eedac6357139f8d40828763af75f/backends/dpdk/dpdkContext.cpp#L128

Abhinavcode13: Reference docs for this P4 naming conventions https://p4.org/p4-spec/docs/P4-16-v1.0.0-spec.html#sec-naming-conventions , if that's the case than this info has to be updated. @AdarshRawat1

Abhinavcode13: @fruffy Does this variable naming convention need to be updated?

Abhinavcode13: Example for first criteria :
https://github.com/p4lang/p4c/blob/f8df27f677a2eedac6357139f8d40828763af75f/backends/dpdk/dpdkArch.cpp#L1502
Since the variable used here is not capitalized. Accordingly, the rule for naming variables would be that we could only have one word with an uncapitalized first character, and the other criterion would be meeting the first condition and beginning another meaningful term with a capital character.

Example for second criteria :
https://github.com/p4lang/p4c/blob/f8df27f677a2eedac6357139f8d40828763af75f/backends/dpdk/dpdkArch.cpp#L1503

fruffy: There are two things going on here. 

1) The naming conventions defined in the P4 specifications apply to P4 programs, not the C++ code of the compiler.

2) For the compiler, we have not decided on a consistent style yet. Currently, each back end has their own style because they are used by different teams. Google style is different from LLVM style is different from Intel or Java style. There is actually an issue which we need to resolve: https://github.com/p4lang/p4c/issues/4404
I have a PR open that defines the style but it is not yet complete: https://github.com/p4lang/p4c/pull/4410
I might work on this next week. 

FYI, there are certain names for these casing styles:
- CamelCase
- camelBack
- snake_case
- kebab-case

AdarshRawat1: > There are two things going on here.
> ....... complete: [Implement the Google variable naming style more faithfully. #4410](https://github.com/p4lang/p4c/pull/4410)

Thank you for explanation sir.

>    I might work on this next week.

Great 😁

> FYI, there are certain names for these casing styles:
> 
> * CamelCase
> * camelBack
> * snake_case
> * kebab-case

Exactly that's my point, instead of writing 

>  _Variables are uncapitalized— e.g., `ipv4header`_

We might have to rephrase it as - `Snake-case` based on [Google Comment style guide](https://google.github.io/styleguide/cppguide.html#Variable_Names) or  `CamelCase` based on [this PR](https://github.com/p4lang/p4c/pull/4410#discussion_r1486393364)

Abhinavcode13: Given you clarified things, it should be okay to use snake_case for the time being? @fruffy

Abhinavcode13: I observe that camelCase is primarily used for the front end, and different naming formats as you mentioned above are used for the back end.

fruffy: This is not something I decide on my own. We first have to make sure other people agree before we do a refactoring. I will check this next week.

Abhinavcode13: Absolutely

Abhinavcode13:
Any update on this @fruffy ?

ChrisDodd:
The language does not define or enforce any naming conventions for P4, other than reserving names containing `__` for compiler/architecture internals.  If someone wants a convention for P4 programs, IMO the right way to do it would be with a linter, not in the compiler.  Perhaps a cpplint or clang-format config could be made to work for P4?

In the compiler's source code, there are a mix of conventions depending on where the code came from and how it evolved.  Some things not mentioned above:

- CamelCase for public/global types and snake_case_t for internal/private types
- snake_case for private fields

fruffy:
> The language does not define or enforce any naming conventions for P4, other than reserving names containing `__` for compiler/architecture internals. If someone wants a convention for P4 programs, IMO the right way to do it would be with a linter, not in the compiler. Perhaps a cpplint or clang-format config could be made to work for P4?

There is a treesitter implementation for P4, which could probably be used for Linting: https://github.com/ace-design/tree-sitter-p4

But it was never completed. 

>CamelCase for public/global types and snake_case_t for internal/private types
>snake_case for private fields

Will try to encode this in #4410. With type you mean a typedef?

ChrisDodd:
> With type you mean a typedef?

typedef or nested struct

vlstill:
> The language does not define or enforce any naming conventions for P4, other than reserving names containing `__` for compiler/architecture internals. If someone wants a convention for P4 programs, IMO the right way to do it would be with a linter, not in the compiler.

I agree.

Also, any recommended code style should be consistent with the stuff defined in specification. E.g. there are things like `hdr.setValid()` but also `hs.lastIndex` and `hs.pop_front(int count)` built-in for header stacks and the naming of struct variables and methods (which are the closest in-language concepts to these builtins) should be consistent with these. Which is a problem since `setValid()` and `pop_front()` are not consistent with each other :-(. Maybe we could add `hs.popFront()` to the language too (and keep `pop_front` for compatibility), but this is deep within the language design area, not the compiler area.

Even if we don't need changes in the spec, I think the recommended code style should be be an attachment of the spec, rather then part of the compiler.

fruffy:
>Maybe we could add hs.popFront() to the language too (and keep pop_front for compatibility), but this is deep within the language design area, not the compiler area.

I think this might be a good idea. Simply to keep style consistent and not confusing.

jafingerhut:
> > Maybe we could add hs.popFront() to the language too (and keep pop_front for compatibility), but this is deep within the language design area, not the compiler area.
> 
> I think this might be a good idea. Simply to keep style consistent and not confusing.

Here is the result of a bit of digging I did in the past on names in the language spec, and their styles: https://github.com/p4lang/p4-spec/issues/1004#issuecomment-1008338434

vlstill:
> Here is the result of a bit of digging I did in the past on names in the language spec, and their styles: [p4lang/p4-spec#1004 (comment)](https://github.com/p4lang/p4-spec/issues/1004#issuecomment-1008338434)

Oh wow, this is much more tangled then I realized. In that case I suggest the best we can do is to define code style + linter and adhere to it when adding new names to the spec. We would have to treat the existing names that do not match the style as exceptions and acknowledge this in the linter.

camelCase seems to be more prevalent for methods/members so I suggest going with that one as default. Probably also for actions/functions/blocks (with `NoAction` being the special one).

The user-defined types and externs could probably be derived from PSA/PNA and their examples, unless it is also inconsistent :-).

PR #4584, opened 2024-03-31
Update rules_boost which pulled in the compromised xz dependency.

fruffy:
Wild story: https://www.openwall.com/lists/oss-security/2024/03/29/4

rules_boost issue: https://github.com/nelhage/rules_boost/issues/555


asl:
Looks good to me in general, though I would suggest someone who really uses bazel build to double confirm :)

jafingerhut (approved):
LGTM, with the same recommendation as someone else made for someone who actually uses Bazel to try this out, but understood the desire to get it in soon.

PR #4588, opened 2024-04-01
Force black to use the pytoml file as configuration.

fruffy:
Some downstream repositories do not seem to pick up the right black configuration. This can happen with out-of-tree builds. Unclear why for me. 

Forcing the configuration path fixes this problem.


fruffy:
@jafingerhut @vlstill Could you give this a quick review? We have some downstream repositories' lint CI failing because of this problem.

vlstill (approved):
While I am not very familiar with black this seems innocent enough to approve.

PR #4591, opened 2024-04-01
Fix spurious failure when two tests try to remove `ptf.log` at the same time.

fruffy:
It can happen that test A deletes `ptf.log` and then test B tries to delete the log a second later. This will cause the following error: 
```
WARNING: Traceback (most recent call last):
WARNING:   File "/usr/local/bin/ptf", line 684, in <module>
WARNING:     logging_setup(config)
WARNING:   File "/usr/local/bin/ptf", line 449, in logging_setup
WARNING:     os.remove(config["log_file"])
WARNING: FileNotFoundError: [Errno 2] No such file or directory: 'ptf.log'
```
To avoid that from happening we make sure the log is saved in the unique temporary directory instead. We already did this for the actual PTF test, but not for the initial PTF check. 

Ideally, we would fix this error from occurring at all in PTF.


jafingerhut (approved):
LGTM

PR #4595, opened 2024-04-02
Fix a compilation error with unity builds and clang.

fruffy:
This is an esoteric issue that only happens when you compile a unity build with clang.
For the following code
```cpp
int posix_memalign(void **memptr, size_t alignment, size_t size)
#ifndef __APPLE__
    noexcept
#endif
{
    maybe_initialize_gc();

    TRACE_ALLOC(size)
    return GC_posix_memalign(memptr, alignment, size);
}
```
you get this exception:

```
In file included from p4c/lib/big_int_util.cpp:17:                                                                                   
In file included from p4c/lib/big_int_util.h:20:                                                                                     
In file included from /usr/include/boost/multiprecision/cpp_int.hpp:12:               
In file included from /usr/include/boost/multiprecision/number.hpp:26:                
In file included from /usr/include/boost/multiprecision/detail/generic_interconvert.hpp:9:                                                                                  
In file included from /usr/include/boost/multiprecision/detail/default_ops.hpp:12:    
In file included from /usr/include/boost/math/special_functions/next.hpp:23:          
In file included from /usr/lib/llvm-19/lib/clang/19/include/xmmintrin.h:31:           
/usr/lib/llvm-19/lib/clang/19/include/mm_malloc.h:25:16: error: 'posix_memalign' is missing exception specification 'throw()'
```

The reason is that the definition in `mm_malloc.h`:
`extern int posix_memalign(void **__memptr, size_t __alignment, size_t __size);`
clashes with the definition in `stdlib.h`
`
extern int posix_memalign (void **__memptr, size_t __alignment, size_t __size)
     __THROW __nonnull ((1)) __wur;
`

@asl Do you maybe have a better approach here?

### DIFF
--- a/.github/workflows/ci-test-debian.yml
+++ b/.github/workflows/ci-test-debian.yml
@@ -20,6 +20,7 @@ jobs:
     env:
       CTEST_PARALLEL_LEVEL: 4
       IMAGE_TYPE: test
+      COMPILE_WITH_CLANG: ON
       BUILD_GENERATOR: Ninja
     steps:
       - uses: actions/checkout@v4
@@ -51,6 +52,7 @@ jobs:
         unity: [ON, OFF]
     runs-on: ubuntu-20.04
     env:
+      COMPILE_WITH_CLANG: ON
       CTEST_PARALLEL_LEVEL: 4
       IMAGE_TYPE: test
       CMAKE_UNITY_BUILD: ${{ matrix.unity }}
--- a/.github/workflows/ci-test-debian.yml
+++ b/.github/workflows/ci-test-debian.yml
@@ -20,7 +20,6 @@ jobs:
     env:
       CTEST_PARALLEL_LEVEL: 4
       IMAGE_TYPE: test
-      COMPILE_WITH_CLANG: ON
       BUILD_GENERATOR: Ninja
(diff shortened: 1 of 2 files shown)


asl:
Interesting, I definitely did a unity build using clang. I will check.

asl:
@fruffy I think better fix should look like this:

```c++
int posix_memalign(void **memptr, size_t alignment, size_t size)
#ifdef __GLIBC__
 __THROW
#endif
```

As far as I can see, it will be compatible with glibc, musl, bionic and libc on darwin platforms.

Will you be able to test it (on the system that reproduces the original issue)?

fruffy:
@asl The fix I have is currently the only one that works. I was able to reproduce it in CI, but I do not know a good way to solve this problem other than importing the extern declaration earlier.

fruffy:
> @fruffy From https://github.com/p4lang/p4c/actions/runs/8539037927/job/23392909174?pr=4595 it looks like it is not related to unity builds at all, as it reproduced on non-unity build as well.

Yeah I was able to reproduce this by including `mm_malloc` in the bottom of the compilation unit. It may be hard to run into this setup without unity builds though.

PR #4597, opened 2024-04-04
Add option to build with dynamic libc and c++, but static other dependencies; change static option name

vlstill:
**Breaking change:** CMake option `BUILD_STATIC_SANS_GLIBC` was renamed to `STATIC_BUILD_WITH_DYNAMIC_GLIBC`.

The semi-static build that we support now (`BUILD_STATIC_SANS_GLIBC`) is useful for building binaries that can be redistributed without adding runtime dependencies on dynamic libraries. However, it is only applicable if all the C++-std-library-using dependencies can be linked statically. If a downstream project has some C++ dependencies (that use the standard library) that need to be linked dynamically this leads to inconsistencies in libstd++ initialization which can render the library unusable.

One can of course fall back to fully-dynamic build, but in my opinion it is useful to be able to link dependencies such as Boost, Abseil and Z3 statically even if there are some dependencies that need to be linked dynamically.

This PR introduces a new build option `BUILD_STATIC_WITH_DYNAMIC_STDLIB` that links all the libraries it can find statically, but the libc and libstdc++ (or other C++ stdlib) are linked dynamically.


vlstill:
I've also added a CI job for this option. I'm not sure if we want it running all the time, but unless we are quite constrained by the amount of CI runner hours we can use I suggest we keep it.

fruffy:
> I've also added a CI job for this option. I'm not sure if we want it running all the time, but unless we are quite constrained by the amount of CI runner hours we can use I suggest we keep it.

I would make it nightly or optional. We are indeed quite constrained on CI hours these days.

.github/workflows/ci-static-sans-stdlib-build-test.yml
```diff
@@ -0,0 +1,58 @@
...
+  cancel-in-progress: ${{ github.ref != 'refs/heads/main' }}
+
+jobs:
+  # Build a p4c release on Ubuntu 20.04.
+  build-linux:
+    runs-on: ubuntu-20.04
+    env:
+      IMAGE_TYPE: test
+      CMAKE_UNITY_BUILD: ON
+      BUILD_STATIC_RELEASE_SANS_STDLIB: ON
```
fruffy, line 23: Maybe even make this part of the build matrix of the static tests?

vlstill: OK, I can try to do that.

vlstill:
> > I've also added a CI job for this option. I'm not sure if we want it running all the time, but unless we are quite constrained by the amount of CI runner hours we can use I suggest we keep it.
> 
> I would make it nightly or optional. We are indeed quite constrained on CI hours these days.

In that case I suggest making both static variants nightly with option to enable them by a label (probably one label for both), similar to validate or sanitizers. @fruffy If you agree, please disable the "required" on the `static-build-test-p4c`.

fruffy:
> > > I've also added a CI job for this option. I'm not sure if we want it running all the time, but unless we are quite constrained by the amount of CI runner hours we can use I suggest we keep it.
> > 
> > 
> > I would make it nightly or optional. We are indeed quite constrained on CI hours these days.
> 
> In that case I suggest making both static variants nightly with option to enable them by a label (probably one label for both), similar to validate or sanitizers. @fruffy If you agree, please disable the "required" on the `static-build-test-p4c`.

Sounds like a good idea. I think one label is good enough. I will disable the required once this PR is ready to be merged.

.github/workflows/ci-static-build-test.yml
```diff
@@ -16,11 +17,18 @@ concurrency:
 jobs:
   # Build a p4c release on Ubuntu 20.04.
   build-linux:
+    # Only run on pull requests with the "run-sanitizer" label.
```
vlstill, line 20:
```suggested
    # Only run on pull requests with the "run-static" label.

File: .github/workflows/ci-static-build-test.yml [hunk 3]
```diff
@@ -1,12 +1,13 @@
...
 on:
+  schedule:
+    # Every day on midnight UTC
+    - cron: "0 0 * * *"
   push:
     branches: [main]
   pull_request:
     branches: [main]
-  merge_group:
-    branches: [main]
```
vlstill, line 9: @fruffy Why is this not present in the scheduled builds? I've checked the existing and none contains it.

fruffy: What do you mean?

vlstill: The `on`->`merge_group` key is not present in any of the CI jobs that have `schedule`.

CMakeLists.txt
```diff
@@ -48,10 +48,15 @@ OPTION (ENABLE_GC "Use libgc" ON)
 OPTION (ENABLE_MULTITHREAD "Use multithreading" OFF)
 OPTION (ENABLE_WERROR "Treat warnings as errors" OFF)
 OPTION (ENABLE_SANITIZERS "Enable sanitizers" OFF)
-OPTION (BUILD_STATIC_RELEASE_SANS_GLIBC
-       "Build a (mostly) statically linked release binary. Glibc is linked dynamically." OFF
-)
-if (BUILD_STATIC_RELEASE_SANS_GLIBC)
+OPTION (BUILD_STATIC_RELEASE_SANS_GLIBC "Build a (mostly) statically linked release binary. Glibc \
```
fruffy, line 51: It might make sense to drop the sans and just use "dynamic". It's a little longer but gets the point across better?

tools/ci-build.sh
```diff
@@ -58,6 +61,15 @@ if [ "$IN_DOCKER" == "TRUE" ]; then
   function sudo() { command "$@"; }
 fi
 
+# Resolve the matrix in the static builds to the right cmake settings.
```
fruffy, line 64: Can we move this check into the YAML? I have been hoping to simplify this build script to a point where we can use it like a normal installation script.

https://stackoverflow.com/questions/66025220/paired-values-in-github-actions-matrix

tools/ci-check-static.sh
```diff
@@ -0,0 +1,51 @@
+#!/bin/bash
```
fruffy, line 1: Neat!

CMakeLists.txt
vlstill: what about `BUILD_PREFER_STATIC_DYNAMIC_GLIBC`/`...DYNAMIC_STDLIB`? That I think is better aligned with the meaning. "...WITH_DYNAMIC..." would read even better, but the option is quite long already :-/.

Do we want a "deprecated" catch for the old spelling? Or do we expect the downstreams to notice/have tests in place :-). I have no idea who uses this besides the Altera compiler :-D.

fruffy: Yeah let's just drop it, iirc the Tofino compiler also had its own static build.

fruffy: I know we are bikeshedding here but I feel like `STATIC_BUILD_PREFER_DYNAMIC_GLIBC` is easier to read?

fruffy (approved):
Early approving to unblock things.

CMakeLists.txt
vlstill: I think the thing that is preferred is the static, except for glibc/stdlib. Or we could just drop the `PREFER` and name this `BUILD_STATIC_WITH_DYNAMIC_GLIBC`, or even `STATIC_BUILD_WITH_DYNAMIC_GLIBC` (I feel like the `BUILD` is a prefix for build options, e.g. also `BUILD_LINK_WITH_GOLD` etc., but I don't have too strong feelings about this).

vlstill: OK. I've renamed it.

vlstill:
@fruffy, did you already remove the static check from the required once? I'm little surprised I don't see it as missing since it was renamed.

fruffy:
> @fruffy, did you already remove the static check from the required once? I'm little surprised I don't see it as missing since it was renamed.

Yes I did

PR #4598, opened 2024-04-04
Deparser: write only (possibly-)changed header fields

vbnogueira:
Leverage compiler to write only (possibly-)changed header fields in deparser instead of always rewriting all of them

For example, in the simple program below:

```
control ingress(
    inout my_ingress_headers_t  hdr,
    inout my_ingress_metadata_t meta,
    in    pna_main_input_metadata_t  istd,
    inout pna_main_output_metadata_t ostd
)
{
   action send_nh(@tc_type("dev") PortId_t port, @tc_type("macaddr") bit<48> srcMac, @tc_type("macaddr") bit<48> dstMac) {
        hdr.ethernet.srcAddr = srcMac;
        hdr.ethernet.dstAddr = dstMac;
        send_to_port(port);
   }

   action drop() {
        drop_packet();
   }

   table nh_table {
        key = {
            hdr.ipv4.dstAddr : exact @tc_type("ipv4") @name("dstAddr");
        }
        actions = {
            send_nh;
            drop;
        }
        size = L3_TABLE_SIZE;
        const default_action = drop;
    }

    apply {
            nh_table.apply();
    }
   }
}

    /*********************  D E P A R S E R  ************************/

control Ingress_Deparser(
    packet_out pkt,
    inout    my_ingress_headers_t hdr,
    in    my_ingress_metadata_t meta,
    in    pna_main_output_metadata_t ostd)
{
    apply {
        pkt.emit(hdr.ethernet);
        pkt.emit(hdr.ipv4);
    }
}
```

The compiler should not rewrite every header in the deparser stage, just the ones that are overwritten by the program itself, in the above example those would be `source and destination mac addresses`

This version has some debugging output, but we have a second patch that removes those


osinstom (commented):
I have some comments about the code itself, but I'd like to clarify the proposed feature first. 

In general, I agree that we perform some redundant/useless write operations in the deparser when a field is not touched. The current approach is generalized for simplicity (at the cost of additional eBPF operations). The proposed solution can work perfectly fine when there is no encap/decap in a P4 program. Have you considered the encap/decap scenario? If the entire header changes its position in a packet, we must write all fields.

PR #4600, opened 2024-04-05
restore missing parentheses

grg:
Restore parentheses that were accidentally dropped from begin call as part of recent code improvements (#4513).

### DIFF
--- a/lib/map.h
+++ b/lib/map.h
@@ -207,7 +207,7 @@ class MapForKey {
     M &map;
     typename M::key_type key;
     class iterator {
-        using MapIt = decltype(map.begin);
+        using MapIt = decltype(map.begin());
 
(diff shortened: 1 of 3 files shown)


fruffy (approved):

asl (approved):
Looks like this case is not covered by anything :(

grg:
> Looks like this case is not covered by anything :(

The test case is the Tofino compiler 🤣 (It's been a couple of months since we last synced with this repo, hence only hitting this issue now.)

I'm going to add a test case, either as part of this PR or a separate PR.

grg:
I've added a single test for ValuesToKeys (which uses MapForKey). It fails to compile without my fix, but compiles and passes with my fix.

PR #4605, opened 2024-04-08
Fix a warning emerging from a testsuite in GTest on newer versions of Clang. 

fruffy:
Context: https://github.com/google/googletest/issues/2271#issuecomment-665742471

### DIFF
--- a/frontends/p4/fromv1.0/programStructure.cpp
+++ b/frontends/p4/fromv1.0/programStructure.cpp
@@ -1225,7 +1225,7 @@ const IR::Statement *ProgramStructure::convertPrimitive(const IR::Primitive *pri
 #define OPS_CK(primitive, n) \
     BUG_CHECK((primitive)->operands.size() == n, "Expected " #n " operands for %1%", primitive)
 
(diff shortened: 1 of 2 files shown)


lib/gc.cpp
```diff
@@ -14,6 +14,14 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
+// This include is necessary to prevent this error from happening:
+// "error: 'posix_memalign' is missing exception specification 'noexcept'"
+// This error is possible when compiling with clang and including "mm_malloc.h" after the definition
+// of posix_memalign. This can easily happen in unity builds.
+#if defined(__clang__)
+#include <mm_malloc.h>
```
vlstill, line 22: Is this header clang-specific?

vlstill (approved):

lib/gc.cpp
fruffy: Actually, this is what I am using to fix build failures (https://github.com/p4lang/p4c/pull/4595). Shouldn't be in this PR.

grg (approved):
LGTM

PR #4608, opened 2024-04-11
Inspector: clear the done flag when revisiting node

grg:
Set the done flag to false when revisiting a node in Tracker::try_start. Fixes a small regression introduced in #4447.

Without this fix, a Tofino pass errors out with an assertion failure from the Visitor infrastructure: `Assertion `current.depth < 10000' failed.` Without resetting the done flag in the tracker, the Inspector::apply_visitor method will always follow the "New or Revisit" path instead of the "Busy" path after a node has been visited once.

The associated test program mimics the Tofino pass that shows the problem.

### DIFF
--- a/ir/visitor.cpp
+++ b/ir/visitor.cpp
@@ -78,6 +78,7 @@ class Visitor::ChangeTracker {
         if (!inserted) {  // We already seen this node, determine its status
             if (it->second.visit_in_progress) return VisitStatus::Busy;
             if (it->second.visitOnce) return VisitStatus::Done;
+            it->second.visit_in_progress = true;
             return VisitStatus::Revisit;
         }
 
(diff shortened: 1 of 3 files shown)


asl:
@grg I think the code is same here as with `Visitor::ChangeTracker`. Does it also have the same problem? Looks like `Visitor::ChangeTracker::try_start` is pretty much the same.

grg:
> @grg I think the code is similar here as with `Visitor::ChangeTracker`. Does it also have the same problem? Looks like `Visitor::ChangeTracker::try_start` is pretty much the same. However, it seems that it uses slightly different set of flags...

I've now applied the fix to `Visitor::ChangeTracker::try_start` after creating a second test case for `Modifier` and seeing the same issue. The issue existed with Modifier/Transform even before the refactoring in #4447, but no one must have encountered it.

asl (approved):
Awesome, thanks!

PR #4609, opened 2024-04-11
Ensure we take compiler-provided declaration of posix_memalign

asl:
Some systems (e.g. those with GNU libc) declare posix_memalign with an exception specifier. Compilers (clang including) might have special handling of this to allow posix_memalign redeclaration with / without exception specifier. As we override posix_memalign for GC purposes we really need to ensure the proper include order to workaround this weirdness.

### DIFF
--- a/lib/gc.cpp
+++ b/lib/gc.cpp
@@ -263,7 +263,7 @@ void *calloc(size_t size, size_t elsize) {
     return rv;
 }
 int posix_memalign(void **memptr, size_t alignment, size_t size)
-#ifndef __APPLE__
+#ifdef __GLIBC__
     noexcept
 #endif
 {
--- a/lib/gc.cpp
+++ b/lib/gc.cpp
@@ -14,6 +14,13 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
+// Some systems (e.g. those with GNU libc) declare posix_memalign with an
+// exception specifier. Compilers (clang including) might have special handling
+// of this to allow posix_memalign redeclaration with / without exception
+// specifier. As we define posix_memalign below in this file we really need to
+// ensure the proper include order to workaround this this weirdness.
+#include <mm_malloc.h>  // NOLINT(build/include_order)
+
 #include "config.h"
 #if HAVE_LIBGC
 #include <gc/gc_cpp.h>
@@ -264,7 +271,7 @@ void *calloc(size_t size, size_t elsize) {
 }
 int posix_memalign(void **memptr, size_t alignment, size_t size)
 #ifdef __GLIBC__
-    noexcept
+    __THROW
 #endif
 {
     maybe_initialize_gc();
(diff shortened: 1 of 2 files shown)


fruffy:
Thanks, we can merge this instead of https://github.com/p4lang/p4c/pull/4595. They are doing similar things.

asl:
@fruffy Yes, I am checking different approaches and how the problem does manifest.

fruffy (approved):
Conditionally approving assuming `COMPILE_WITH_CLANG: ON` is removed.

asl:
@fruffy How about adding builder that would build p4c using clang on Ubuntu?

vlstill:
> @fruffy How about adding builder that would build p4c using clang on Ubuntu?

Considering the discussion I had with @fruffy regarding the static builds, this should be probably part of nightly builds not PR suite (the risk of breakage would be relatively small I guess, especially breakage just under clang and not on macOS). I think having a clang build in CI would be useful.

fruffy:
We do already have a Clang nightly build, it is the sanitizer one. It just does not use unity builds.

asl:
> We do already have a Clang nightly build, it is the sanitizer one. It just does not use unity builds.

Ah, great. How does it do notifications?

fruffy:
> > We do already have a Clang nightly build, it is the sanitizer one. It just does not use unity builds.
> 
> Ah, great. How does it do notifications?

Notifications are kind of strange here: https://docs.github.com/en/actions/monitoring-and-troubleshooting-workflows/notifications-for-workflow-runs

The last user that modified the workflow (me) gets notified of failures.

lib/gc.cpp
```diff
@@ -14,6 +14,13 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
+// Some systems (e.g. those with GNU libc) declare posix_memalign with an
+// exception specifier. Compilers (clang including) might have special handling
+// of this to allow posix_memalign redeclaration with / without exception
+// specifier. As we define posix_memalign below in this file we really need to
+// ensure the proper include order to workaround this this weirdness.
```
asl, line 21:
```suggestion
// ensure the proper include order to workaround this weirdness.
```

asl:
@fruffy Ok, I think this is the least bad approach, still not perfect, but we'd have to live with this as-is. Merging when ready.

fruffy:
> @fruffy Looks like ptf-ebpf fails due to some network issue?

Looks like a spurious failure, I restarted the test. It's optional so it should not block merging anyway.

asl:
> Looks like a spurious failure, I restarted the test. It's optional so it should not block merging anyway.

I already restarted it 3 or 4 times...

fruffy:
> > Looks like a spurious failure, I restarted the test. It's optional so it should not block merging anyway.
> 
> I already restarted it 3 or 4 times...

Ah just saw. Hmm, there is nothing we can really do about it for now. I would ignore it. If the issue persists we can start to look into it. This particular test is not super important.

PR #4606, opened 2024-04-10
Added initial PNA support into p4c.

rupesh-chiluka-marvell:
To add PNA support into the bmv2 backend. Leveraging the _psa_switch_ as the foundation, the _pna_nic_ has been developed.

A basic L2 forward PNA p4 program has been compiled with the updated P4 compiler (p4c-bm2-pna). Basic L2 Forward test has been carried out.

PR #4615, opened 2024-04-12
[#4614] Unconditionally copy `dpdk` p4include files to the binary directory

### LINKED ISSUE #4614: PNA and PSA DPDK tests should not be run by `check-p4` target
With the following `cmake` options:
```
cmake ../../../ -G Ninja -DCMAKE_BUILD_TYPE=RELEASE -DCMAKE_INSTALL_PREFIX=. -DENABLE_BMV2=OFF -DENABLE_EBPF=OFF -DENABLE_UBPF=OFF -DENABLE_DPDK=OFF -DENABLE_P4TEST=ON -DENABLE_GC=ON -DENABLE_P4TC=OFF
ninja -j4
```
`check-p4` fails because `check-p4` still runs the `pna-*` and `psa-*` DPDK tests in `testdata/p4_16_samples/`. `backend/p4test/CMakeLists.txt`:
```
set (P4TEST_SUITES
  "${P4C_SOURCE_DIR}/testdata/p4_16_samples/*.p4"
  ...
```

and in `backend/dpdk/CMakeLists.txnt`:
```
set (P4_16_SUITES
  "${P4C_SOURCE_DIR}/testdata/p4_16_samples/pna-*.p4"
  ...
 p4c_add_tests("dpdk" ${DPDK_COMPILER_DRIVER} "${P4_16_SUITES}" "" "--bfrt")
```

The below tests under `check-p4` are failing:
```
The following tests FAILED:
        823 - p4/testdata/p4_16_samples/pna-dpdk-bvec_union.p4 (Failed)
        824 - p4/testdata/p4_16_samples/pna-dpdk-bvec_union_dpdk.p4 (Failed)
        827 - p4/testdata/p4_16_samples/pna-dpdk-direct-meter-learner.p4 (Failed)
        828 - p4/testdata/p4_16_samples/pna-dpdk-flatten-local-struct-decl.p4 (Failed)
        830 - p4/testdata/p4_16_samples/pna-dpdk-header-union-stack.p4 (Failed)
        831 - p4/testdata/p4_16_samples/pna-dpdk-header-union-stack1.p4 (Failed)
        832 - p4/testdata/p4_16_samples/pna-dpdk-header-union-stack2.p4 (Failed)
        833 - p4/testdata/p4_16_samples/pna-dpdk-invalid-hdr-warnings5.p4 (Failed)
        834 - p4/testdata/p4_16_samples/pna-dpdk-invalid-hdr-warnings6.p4 (Failed)
        835 - p4/testdata/p4_16_samples/pna-dpdk-parser-state-err.p4 (Failed)
        836 - p4/testdata/p4_16_samples/pna-dpdk-parser-wrong-arith.p4 (Failed)
        837 - p4/testdata/p4_16_samples/pna-dpdk-small_sample.p4 (Failed)
        844 - p4/testdata/p4_16_samples/pna-dpdk-table-key-use-annon.p4 (Failed)
        845 - p4/testdata/p4_16_samples/pna-dpdk-toeplitz-hash-1.p4 (Failed)
        846 - p4/testdata/p4_16_samples/pna-dpdk-toeplitz-hash.p4 (Failed)
        847 - p4/testdata/p4_16_samples/pna-dpdk-union-bmv2.p4 (Failed)
        848 - p4/testdata/p4_16_samples/pna-dpdk-wrong-warning.p4 (Failed)
        849 - p4/testdata/p4_16_samples/pna-elim-hdr-copy-dpdk.p4 (Failed)
        855 - p4/testdata/p4_16_samples/pna-example-dpdk-varbit-1.p4 (Failed)
        856 - p4/testdata/p4_16_samples/pna-example-dpdk-varbit.p4 (Failed)
        859 - p4/testdata/p4_16_samples/pna-example-ipsec.p4 (Failed)
        881 - p4/testdata/p4_16_samples/pna-too-big-label-name-dpdk.p4 (Failed)
        918 - p4/testdata/p4_16_samples/psa-dpdk-checksum-arg-header.p4 (Failed)
        925 - p4/testdata/p4_16_samples/psa-dpdk-large-header-fields.p4 (Failed)
        926 - p4/testdata/p4_16_samples/psa-dpdk-large-struct-fields.p4 (Failed)
        941 - p4/testdata/p4_16_samples/psa-dpdk-table-entries-exact-ternary.p4 (Failed)
        965 - p4/testdata/p4_16_samples/psa-dpdk-tmp-mask-align.p4 (Failed)
        973 - p4/testdata/p4_16_samples/psa-example-dpdk-byte-alignment_1.p4 (Failed)
        974 - p4/testdata/p4_16_samples/psa-example-dpdk-byte-alignment_2.p4 (Failed)
        975 - p4/testdata/p4_16_samples/psa-example-dpdk-byte-alignment_3.p4 (Failed)
        977 - p4/testdata/p4_16_samples/psa-example-dpdk-byte-alignment_5.p4 (Failed)
        978 - p4/testdata/p4_16_samples/psa-example-dpdk-byte-alignment_6.p4 (Failed)
        979 - p4/testdata/p4_16_samples/psa-example-dpdk-byte-alignment_7.p4 (Failed)
        980 - p4/testdata/p4_16_samples/psa-example-dpdk-byte-alignment_8.p4 (Failed)
        981 - p4/testdata/p4_16_samples/psa-example-dpdk-byte-alignment_9.p4 (Failed)
        982 - p4/testdata/p4_16_samples/psa-example-dpdk-counter.p4 (Failed)
        983 - p4/testdata/p4_16_samples/psa-example-dpdk-directmeter.p4 (Failed)
        984 - p4/testdata/p4_16_samples/psa-example-dpdk-externs.p4 (Failed)
        985 - p4/testdata/p4_16_samples/psa-example-dpdk-local-variable.p4 (Failed)
        986 - p4/testdata/p4_16_samples/psa-example-dpdk-meter-execute-err.p4 (Failed)
        987 - p4/testdata/p4_16_samples/psa-example-dpdk-meter.p4 (Failed)
        988 - p4/testdata/p4_16_samples/psa-example-dpdk-meter1.p4 (Failed)
```

kfcripps:
~~Moves all psa and pna dpdk tests to `p4_16_pna_samples/` and `p4_16_psa_samples/` so that the `check-p4` testing target does not run them.~~
Unconditionally copy `dpdk` p4include files to the binary directory so that they will be available for the `check-p4` testing target even when the dpdk backend is disabled.

Fixes #4614


CMakeLists.txt
```diff
@@ -539,18 +539,13 @@ add_custom_target(update_includes ALL
   COMMAND ${CMAKE_COMMAND} -E copy_if_different ${P4C_SOURCE_DIR}/p4include/*.p4 ${P4C_BINARY_DIR}/p4include
   COMMAND ${CMAKE_COMMAND} -E make_directory ${P4C_BINARY_DIR}/p4include/bmv2
   COMMAND ${CMAKE_COMMAND} -E copy_if_different ${P4C_SOURCE_DIR}/p4include/bmv2/psa.p4 ${P4C_BINARY_DIR}/p4include/bmv2
+  COMMAND ${CMAKE_COMMAND} -E make_directory ${P4C_BINARY_DIR}/p4include/dpdk
```
fruffy, line 542: Add a TODO pointing to #4614 that we should clean this up?

fruffy (approved):

CMakeLists.txt
kfcripps: By "clean this up," do you mean changing the tests' <pna.p4> includes to either <bmv2/pna.p4> or <dpdk/pna.p4>? Or do you mean that check-p4 should not run these dpdk or bmv2 tests at all?

fruffy: I think both cases will end up being the result of resolving #4614. We can also put the TODO as deciding whether to fully disable these tests.

kfcripps: Is there any downside to running them under check-p4? My understanding now is that for all of the bmv2 and dpdk tests, check-p4 is just checking the frontend-generated and midend-generated IR and/or diagnostics, so they would still be useful to run. However if we are checking what is generated by e.g. the dpdk pna backend, then it wouldn't make sense to run the dpdk pna tests under check-p4

kfcripps: But it definitely makes sense to at least make the includes more verbose. I can try doing that in this PR later.

kfcripps:
Hmm, I just noticed that we also have [p4include/pna.p4](https://github.com/p4lang/p4c/blob/main/p4include/pna.p4) in addition to the dpdk-specific [p4include/dpdk/pna.p4](https://github.com/p4lang/p4c/blob/main/p4include/dpdk/pna.p4). As far as I understand, it doesn't make sense to include the dpdk-specific `pna.p4` unless the test requires one of the dpdk-specific extern functions not present in the generic pna.p4 file, so I have also replaced all `#include <dpdk/pna.p4>` with `#include <pna.p4>` wherever dpdk/pna.p4 is not needed.

testdata/p4_16_samples_outputs/psa-example-select_tuple-wc.p4
```diff
@@ -1,5 +1,5 @@
 #include <core.p4>
-#include <bmv2/psa.p4>
```
fruffy, line 2: Are these changes because the file is DPDK-related?

fruffy (commented):
Thanks a lot! This a great cleanup. @jafingerhut could you also take a look?

testdata/p4_16_samples_outputs/psa-example-select_tuple-wc.p4
kfcripps: testdata/p4_16_samples/psa-example-select_tuple-wc.p4 was added in a dpdk-related PR (#2836), so the intent was likely to include dpdk/psa.p4. However, testdata/p4_16_samples/psa-example-select_tuple-wc.p4 contained `#include <psa.p4>`, so bmv2/psa.p4 was actually included.

jafingerhut:
FYI, in case this isn't obvious to everyone:

It is _expected_ that different target devices that implement PSA, or that implement PNA, architectures can customize the bit widths of several types, e.g. PortId_t, for that particular target device.  Most P4 programs written for such an architecture will likely compile and run tests in this repo fine regardless of the particular value for those bit widths.

Some implementations have chosen to make additional customizations to pna.p4, e.g. dpdk/pna.p4 has additional changes to some parameters of some extern methods.  As @kfcripps pointed out earlier, many particular test programs might not be using such methods, and those will likely work identically regardless of which version of the pna.p4 include file they use.

While we could consider trying to divide up PNA architecture test programs into these 3 categories:

(a) ones that must use a BMv2-specific pna.p4 include file (assuming one exists in the near future, as appears likely given other PRs in the works)
(b) ones that must use a DPDK-specific pna.p4 include file
(c) ones that can use "a more generic pna.p4 include file"

Note that if in the future the DPDK-specific pna.p4 include file were to change (as an example, but this is not specific to DPDK), it might require moving test programs around that were formerly in group (c) into group (b).

I am not personally strongly against such an approach, but would like to suggest that it might be less busy-work for everyone involved if we DO NOT attempt to have category (c) tests _at all_.  That is, _every_ test program for the PNA architecture would be focused solely on testing one particular target.

If we take that approach, yes, then each target-implementer of an architecture gets complete control over what their test programs are, and all of them would include their target-specific version of the pna.p4 include file, and it would be up to that team of people to decide exactly what their test programs test, vs. what they do not.

That approach also makes it easier to manage test programs in the currently-common situation that different targets implement different subsets of an architecture specification.

kfcripps:
> Note that if in the future the DPDK-specific pna.p4 include file were to change (as an example, but this is not specific to DPDK), it might require moving test programs around that were formerly in group (c) into group (b).

If a test is already in group (c) then why would changing the dpdk-specific pna.p4 file require moving the test to group (b)?

> I am not personally strongly against such an approach, but would like to suggest that it might be less busy-work for everyone involved if we DO NOT attempt to have category (c) tests at all. That is, every test program for the PNA architecture would be focused solely on testing one particular target.

> If we take that approach, yes, then each target-implementer of an architecture gets complete control over what their test programs are, and all of them would include their target-specific version of the pna.p4 include file, and it would be up to that team of people to decide exactly what their test programs test, vs. what they do not.

> That approach also makes it easier to manage test programs in the currently-common situation that different targets implement different subsets of an architecture specification.

I don't really have a preference between always including the target-specific pna.p4 vs. sometimes including the generic pna.p4 file so I'll let you two decide @fruffy @jafingerhut and then I can update the includes touched by this PR accordingly.

jafingerhut:
@kfcripps asked: "If a test is already in group (c) then why would changing the dpdk-specific pna.p4 file require moving the test to group (b)?"

Because the P4 program being tested uses some feature in the dpdk-specific pna.p4 file that the DPDK developers decided to modify, similarly to how at some point in the past they decided to modify the signatures of some extern methods to take additional parameters.  They might decide to do so again in the future (or another target might), and if any test P4 programs used that method, it would need to be updated to continue compiling and running for the target.

kfcripps:
> @kfcripps asked: "If a test is already in group (c) then why would changing the dpdk-specific pna.p4 file require moving the test to group (b)?"
> 
> Because the P4 program being tested uses some feature in the dpdk-specific pna.p4 file that the DPDK developers decided to modify, similarly to how at some point in the past they decided to modify the signatures of some extern methods to take additional parameters. They might decide to do so again in the future (or another target might), and if any test P4 programs used that method, it would need to be updated to continue compiling and running for the target.

@jafingerhut But if a test is in group (c), that means it's not including the dpdk-specific pna.p4 file, and therefore not using some dpdk-specific feature, right?

kfcripps:
@jafingerhut Do you mean that a test might need to be moved from group (c) to (b) if:

- DPDK makes a dpdk-specific change to some existing feature that is currently exposed in the generic pna.p4 file, and
- The test uses this feature in some way, and
- The test belongs to the set of tests added to in `P4_16_SUITES` in https://github.com/p4lang/p4c/blob/main/backends/dpdk/CMakeLists.txt, and
- We want the test to continue to exist among `P4_16_SUITES` for some reason

?

kfcripps:
@fruffy @jafingerhut Should I merge this as is? Or shall I remove all `#include <pna.p4>` first?

fruffy:
I lost track of the discussion.

I do like the shared approach solely for the reason that we are looking into developing a BMv2 PNA target https://github.com/p4lang/p4c/pull/4606 and can reuse all the programs with generic includes as test programs.

The alternative to this approach, without duplicating test programs, is to use `#define` directives which include the right header depending on the compiler. But that requires even more rewriting work.

kfcripps:
> I do like the shared approach solely for the reason that we are looking into developing a BMv2 PNA target https://github.com/p4lang/p4c/pull/4606 and can reuse all the programs with generic includes as test programs.

Ok, I will merge this as-is if the changes look ok to you @jafingerhut?

jafingerhut:
@kfcripps I will approve this PR in a moment.  Please go ahead and we can work out later if there are more changes we want to make.  Thanks.

kfcripps:
@jafingerhut @fruffy Thanks!

PR #4617, opened 2024-04-13
added the option to make control-plane build optional also handled Errors for other backends

### LINKED ISSUE #4420: Make the control plane optional

Sanket-0510:
fixes #4420 
I have configured option for control_plane_build. By default its ON. I have also handled errors for the backends which requires control-plane as build dependency.


CMakeLists.txt
```diff
@@ -434,7 +435,20 @@ endforeach(ext)
 
 add_subdirectory (frontends)
 add_subdirectory (midend)
-add_subdirectory (control-plane)
+
+if(NOT BUILD_CONTROL_PLANE)
```
fruffy, line 439: This is a good first step, but I would move these checks into CMakelists of the respective back end folder. 

In all likelihood things will break if you try to build without control-plane now, because there are many CMake targets that depend on it. There might even be some code that accidentally includes it. 

To fix this for CMAKE you can at least create a variable (e.g., P4C_CONTROLPLANE) which is only set with the value `controlplane` when the control plane is active.

CMakeLists.txt
```diff
@@ -63,6 +63,7 @@ OPTION (BUILD_AUTO_VAR_INIT_PATTERN "Initialize variables with pattern during bu
 OPTION (ENABLE_IWYU "Enable checking includes with IWYU" OFF)
 # Support a legacy option. TODO: Remove?
 OPTION (ENABLE_UNIFIED_COMPILATION "Enable CMAKE_UNITY_BUILD" OFF)
+OPTION (BUILD_CONTROL_PLANE "Build the control plane" ON)
```
fruffy, line 66:
```suggestion
OPTION (ENABLE_CONTROLPLANE "Build the control plane" ON)
```

CMakeLists.txt
Sanket-0510: But that won't solve our problem at the moment. I looked all over Backends, if am not wrong it's p4tools, dpdk and Tests which requires control-plane as dependency.

fruffy: What do you mean? These error checks should be handled in each individual back end.

Sanket-0510: Ohhh I get it I should have handled this in the individual backend. Also at the moment 3 backends are there having Control-plane as dependency if I am not wrong?

fruffy: There might be more, but we will see this once we try to start to compile code without the control plane enabled.

Sanket-0510: Okay I'll try to move these checks to the individual backend ✅

PR #4618, opened 2024-04-13
Fix: Outdated links

AdarshRawat1:
## Changes 
Fixing outdated links throughout the repository.


backends/ebpf/README.md
```diff
@@ -37,7 +37,7 @@ Communications Review in 2014:
 P4 itself is protocol-independent but allows programmers to express a
 rich set of data plane behaviors and protocols.  This back-end only
 supports the newest version of the P4 programming language,
-[P4_16](http://p4.org/wp-content/uploads/2016/12/P4_16-prerelease-Dec_16.html).
+[P4_16](https://p4.org/p4-spec/docs/P4-16-v-1.2.3.html).
```
jafingerhut, line 40: I am not sure whether the purpose of this link is to point at one specific revision of the language specification.

However, if the goal is to point at the latest version, _and_ older versions, they are all available here: https://p4.org/specs

so we may want to consider using that as the URL here rather than pointing at a specific version.

AdarshRawat1: Sure, I'll update the link. 
> Definitely, it is supposed to point to the latest version of P4 and the link provided by me can get obsolete soon.

fruffy (approved):

jafingerhut (approved):
LGTM

fruffy:
This PR is still a draft, is it ready?

AdarshRawat1:
I'm looking for any other outdated links, that needs to be updated.

AdarshRawat1:
# Help 

The link here points to an older Version, but bcc has discontinued support for compiling P4 and have removed it [here](https://github.com/iovisor/bcc/commit/e17c4f7324d8fc5cc24ba8ee1db451666cd7ced3) .  So should I remove this reference ? 
https://github.com/p4lang/p4c/blob/8d21a8b4f2dce71c51ede5a23f891a9f6b8cff2d/backends/ebpf/README.md?plain=1#L8

AdarshRawat1:
Note 
> I tried to correct the link to the best of my knowledge, but there might be errors in them. Kindly re-review the links once.

jafingerhut:
For the outdated link to `bcc`, it might be nice to explicitly mark it as a "historical reference only".  I only suggest that in case someone comes across it in the future and decides to investigate further to make changes that cause it to be supported again, or decides to create a system with older versions of software where they can try it out.

AdarshRawat1:
> For the outdated link to `bcc`, it might be nice to explicitly mark it as a "historical reference only".

Sure ! So should leave a link to the point of 'bcc' repository before the support for P4 was removed 
or remove the link?

fruffy:
> > For the outdated link to `bcc`, it might be nice to explicitly mark it as a "historical reference only".
> 
> Sure ! So should leave a link to the point of 'bcc' repository before the support for P4 was removed or remove the link?

I would add an annotation saying that this link is for historical reference only.

AdarshRawat1:
Any suggestions on this outdated link?

> ## 2nd outdated link
> This might not be that significant. [current link](https://mail.python.org/pipermail/python-list/2012-August/628809.html)
> 
> https://github.com/p4lang/p4c/blob/8d21a8b4f2dce71c51ede5a23f891a9f6b8cff2d/tools/cpplint.py#L4834-L4835

fruffy:
> Any suggestions on this outdated link?
> 
> > ## 2nd outdated link
> > This might not be that significant. [current link](https://mail.python.org/pipermail/python-list/2012-August/628809.html)
> > https://github.com/p4lang/p4c/blob/8d21a8b4f2dce71c51ede5a23f891a9f6b8cff2d/tools/cpplint.py#L4834-L4835

You can ignore it. We should deprecated cpplint at some point.

PR #4623, opened 2024-04-17
Move IRUtils Literal get functions to the respective IR members. Add stringliteral get function. 

fruffy:
I have been doing this a lot lately, so might as well contribute a utility function. Maybe all of these "get" functions should be static class members?

### DIFF
--- a/backends/p4tools/common/compiler/convert_hs_index.cpp
+++ b/backends/p4tools/common/compiler/convert_hs_index.cpp
@@ -18,7 +18,7 @@ const IR::Node *HSIndexToMember::postorder(IR::ArrayIndex *curArrayIndex) {
 const IR::ArrayIndex *HSIndexToMember::produceStackIndex(const IR::Type *type,
                                                          const IR::Expression *expression,
                                                          size_t arrayIndex) {
-    return new IR::ArrayIndex(type, expression, IR::getConstant(IR::getBitType(32), arrayIndex));
+    return new IR::ArrayIndex(type, expression, IR::Constant::get(IR::getBitType(32), arrayIndex));
 }
 
 }  // namespace P4Tools
--- a/backends/p4tools/modules/testgen/core/small_step/table_stepper.h
+++ b/backends/p4tools/modules/testgen/core/small_step/table_stepper.h
@@ -73,7 +73,8 @@ class TableStepper {
 
     /// Sets the action taken by the given table. The arguments in the given MethodCallExpression
     /// are assumed to be symbolic values.
-    const IR::StringLiteral *getTableActionString(const IR::MethodCallExpression *actionCall);
+    static const IR::StringLiteral *getTableActionString(
+        const IR::MethodCallExpression *actionCall);
 
     /// A helper function to iteratively resolve table keys into symbolic values.
     /// This function returns false, if no key needs to be resolved.
--- a/ir/expression.cpp
+++ b/ir/expression.cpp
@@ -168,7 +168,7 @@ const IR::BoolLiteral *IR::BoolLiteral::get(bool value, const Util::SourceInfo &
     return value ? &TRUE_BOOLLITERAL : &FALSE_BOOLLITERAL;
 }
 
-const IR::StringLiteral *IR::StringLiteral::get(const IR::Type *t, cstring value,
+const IR::StringLiteral *IR::StringLiteral::get(cstring value, const IR::Type *t,
                                                 const Util::SourceInfo &si) {
     // Do not cache values with a non-empty source info (yet).
     if (si.isValid()) {
--- a/ir/expression.cpp
+++ b/ir/expression.cpp
@@ -16,6 +16,7 @@ limitations under the License.
 
 #include <ostream>
 
+#include "absl/container/flat_hash_map.h"
 #include "ir/id.h"
 #include "ir/indexed_vector.h"
 #include "ir/ir.h"
@@ -147,7 +148,7 @@ const IR::Constant *IR::Constant::get(const IR::Type *t, big_int v, Util::Source
     }
     // Constants are interned. Keys in the intern map are pairs of types and values.
     using key_t = std::tuple<int, RTTI::TypeId, bool, big_int>;
-    static std::map<key_t, const Constant *> CONSTANTS;
+    static absl::flat_hash_map<key_t, const Constant *, Util::Hash> CONSTANTS;
 
     auto *&result = CONSTANTS[{tb->width_bits(), t->typeId(), tb->isSigned, v}];
     if (result == nullptr) {
@@ -176,7 +177,7 @@ const IR::StringLiteral *IR::StringLiteral::get(cstring value, const IR::Type *t
     }
     // String literals are interned.
     using key_t = std::pair<cstring, const IR::Type *>;
-    static std::map<key_t, const IR::StringLiteral *> STRINGS;
+    static absl::flat_hash_map<key_t, const IR::StringLiteral *, Util::Hash> STRINGS;
 
     auto *&result = STRINGS[{value, t}];
     if (result == nullptr) {
--- a/ir/expression.cpp
+++ b/ir/expression.cpp
@@ -137,3 +137,50 @@ IR::Constant IR::Constant::operator-() const { return IR::Constant(-value); }
 IR::Constant IR::Constant::GetMask(unsigned width) {
     return (IR::Constant(1) << width) - IR::Constant(1);
 }
+
+const IR::Constant *IR::Constant::get(const IR::Type *t, big_int v, Util::SourceInfo si) {
+    // Only cache bits with width lower than 16 bit to restrict the size of the cache.
+    // Do not cache values with a non-empty source info (yet).
+    const auto *tb = t->to<Type_Bits>();
+    if (t->width_bits() > 16 || tb == nullptr || si.isValid()) {
+        return new IR::Constant(si, t, v);
+    }
+    // Constants are interned. Keys in the intern map are pairs of types and values.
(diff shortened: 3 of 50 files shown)


vlstill:
> . Maybe all of these "get" functions should be static class members?

Do you mean e.g. `IR::StringLiteral::get`?

fruffy:
> > Maybe all of these "get" functions should be static class members?
> 
> Do you mean e.g. `IR::StringLiteral::get`?

Yes, something in that form.

fruffy:
> > > > Maybe all of these "get" functions should be static class members?
> > > 
> > > 
> > > Do you mean e.g. `IR::StringLiteral::get`?
> > 
> > 
> > Yes, something in that form.
> 
> OK, I think that would make more sense than the freestanding functions.

Alright, gave this a shot.

ir/expression.def
```diff
@@ -257,6 +261,11 @@ class Constant : Literal {
 class BoolLiteral : Literal {
     bool value;
     toString{ return value ? "true" : "false"; }
+
+#emit
+    /// @return a bool literal. Both booleans are interned.
+    static const BoolLiteral *get(bool value, const Util::SourceInfo &si = {});
+#end
```
ChrisDodd, line 268: I don't think these functions need `#emit`/`#end` wrapped around them -- should work fine without them

fruffy: Seems like this worked.

backends/p4tools/common/compiler/convert_hs_index.cpp
```diff
@@ -18,7 +18,7 @@ const IR::Node *HSIndexToMember::postorder(IR::ArrayIndex *curArrayIndex) {
 const IR::ArrayIndex *HSIndexToMember::produceStackIndex(const IR::Type *type,
                                                          const IR::Expression *expression,
                                                          size_t arrayIndex) {
-    return new IR::ArrayIndex(type, expression, IR::getConstant(IR::getBitType(32), arrayIndex));
+    return new IR::ArrayIndex(type, expression, IR::Constant::get(IR::getBitType(32), arrayIndex));
```
vlstill, line 21: why not `IR::Type_Bits::get(32)`?

ir/expression.def
```diff
@@ -267,6 +274,9 @@ class StringLiteral : Literal {
 #emit
     operator IR::ID() const { return IR::ID(srcInfo, value); }
 #end
+
+    /// @returns a string literal. The value is cached.
+    static const StringLiteral *get(cstring value, const Type *t = Type_String::get(), const Util::SourceInfo &si = {});
```
vlstill, line 279: The literals with non-string type seem quite niche (or even hacky). I wonder if these should be created by a different overload (or just use the constructor for them). I am not too happy with the `get` functions having different order of arguments then the constructors, that could be quite confusing :-(.

ir/expression.cpp
```diff
@@ -137,3 +137,45 @@ IR::Constant IR::Constant::operator-() const { return IR::Constant(-value); }
...
+
+const IR::Constant *IR::Constant::get(const IR::Type *t, big_int v, Util::SourceInfo si) {
+    // Only cache bits with width lower than 16 bit to restrict the size of the cache.
+    const auto *tb = t->to<Type_Bits>();
+    if (t->width_bits() > 16 || tb == nullptr) {
+        return new IR::Constant(si, t, v);
+    }
+    // Constants are interned. Keys in the intern map are pairs of types and values.
+    using key_t = std::tuple<int, RTTI::TypeId, bool, big_int>;
+    static std::map<key_t, const Constant *> CONSTANTS;
```
vlstill, line 149: suggestion: Use an abseil map with the new hasher?

ir/expression.cpp
```diff
@@ -137,3 +137,45 @@ IR::Constant IR::Constant::operator-() const { return IR::Constant(-value); }
...
+    if (result == nullptr) {
+        result = new Constant(si, tb, v);
+    }
+
+    return result;
+}
+
+const IR::BoolLiteral *IR::BoolLiteral::get(bool value, const Util::SourceInfo &si) {
+    // Boolean literals are interned.
+    static std::map<bool, const IR::BoolLiteral *> LITERALS;
```
vlstill, line 161: this does not look like a job for a map :-). There are just 2 possible values.

ir/expression.cpp
```diff
@@ -137,3 +137,45 @@ IR::Constant IR::Constant::operator-() const { return IR::Constant(-value); }
 IR::Constant IR::Constant::GetMask(unsigned width) {
     return (IR::Constant(1) << width) - IR::Constant(1);
 }
+
+const IR::Constant *IR::Constant::get(const IR::Type *t, big_int v, Util::SourceInfo si) {
```
vlstill, line 141: Please be careful we can only cache the values if the source info is empty. Otherwise we would need to take the source info into account. This kind of defeats the purpose of the cache. But we don't want to break error messages by the cache. Same applies below.

vlstill (dismissed):
I think there is a problem when caching values that can have valid source info. Either we cannot cache these at all, or we would need the cache to take the source info into account (but then it is unlikely it would hit). Otherwise, we risk breaking error messages because the programmer will believe that the source info is there, but it won't (or it will be wrong one!). I wonder if the cache makes sense at all. Maybe in testgen it could...

I do realize this is not new in the PR. But since you have changed this code and likely intend to use it more widely, I suggest this is fixed now.

ir/expression.cpp
fruffy: I think only caching in case of an empty source info is a good approach. #4302 could solve this problem.

ir/expression.def
fruffy: There is `Type_MatchKind` which could be use here. I did not want to restrict the type choice too much.

backends/p4tools/common/compiler/convert_hs_index.cpp
fruffy: Good point, may introduce a separate PR for this.

ChrisDodd: Better (more consistent) to use `IR::Type::Bits::get(32);`

ir/expression.cpp
fruffy: Once we are taking SourceInfo into account we may have to resort to a map again, but for now I am just declaring two static values.

backends/p4tools/common/compiler/convert_hs_index.cpp
fruffy: I remember now why. The implementation of `IR::getBitType` and `IR::Type::Bits::get` differs slightly. `IR::Type::Bits::get` imposes a maximum size for the bit width which can cause problems with P4Testgen. I will try to untangle this in a separate PR.

ir/expression.cpp
fruffy: `big_int` does not have an implementation for hashing it looks like? It works for the string literals at least.

backends/p4tools/common/compiler/convert_hs_index.cpp
ChrisDodd: The limit is `maximumWidthSupported()` which defaults to 2048, but can be set larger (backends/p4tools/common sets it to `INT_MAX`)

backends/p4tools/modules/testgen/core/small_step/table_stepper.h
```diff
@@ -73,7 +73,8 @@ class TableStepper {
 
     /// Sets the action taken by the given table. The arguments in the given MethodCallExpression
     /// are assumed to be symbolic values.
-    const IR::StringLiteral *getTableActionString(const IR::MethodCallExpression *actionCall);
+    static const IR::StringLiteral *getTableActionString(
+        const IR::MethodCallExpression *actionCall);
```
ChrisDodd, line 77: This function is defined in small_step/table_stepper.cpp (and used there) and also used in targets/pna/shared_table_stepper.cpp (and not defined there).  So I'm suprised it still compiles if you make the declaration `static` -- does it depend on unified build pulling them into the same compilation unit?

ChrisDodd (approved):

backends/p4tools/common/compiler/convert_hs_index.cpp
fruffy: Ah right, this was resolved already. I will create a PR that removes `IR::getBitType`.

backends/p4tools/modules/testgen/core/small_step/table_stepper.h
fruffy: This is a static class member function so the derived classes should be able to access it? This also works without a unified build.

ChrisDodd: Ah, my bad -- I thought it was a file-scopt function, not a member.

lib/hash.h
```diff
@@ -11,6 +11,8 @@
 #include <tuple>
 #include <type_traits>
 
+#include "lib/big_int_util.h"
```
asl, line 14: maybe it would make sense to reverse the dependency and instantiate hasher inside `lib/big_int_util.h`? Right now it seems that everyone who'd like to use `lib/hash.h` would pull `big_int_util.h`, this seems to be not quite a good thing to me.

fruffy: `big_int_util` is mostly used for the declaration of `big_int`, which is used fairly pervasively in the compiler. We can split that out?

asl: Right. But why does it become the hash dependency? It seems it should be the opposite: `big_int_utils` should depend on `hash` and not vice versa.

fruffy: Yes, I have a fix for that in #4655.

PR #4626, opened 2024-04-17
Call PassManager's debug hook even after a failing pass

vlstill:
**breaking change**: Debug hooks in `PassManager` can now be called with the IR::Node being `nullptr`. This indicates that the last executed pass failed and did not return any value.

This allows running a debug hook also after a failing pass too. This can be useful e.g. to check diagnostic counts for the pass in the debug hook.

### DIFF
--- a/frontends/common/parser_options.cpp
+++ b/frontends/common/parser_options.cpp
@@ -490,15 +490,18 @@ void ParserOptions::dumpPass(const char *manager, unsigned seq, const char *pass
             if (filename == "-") filename = "tmp.p4";
 
             cstring fileName = makeFileName(dumpFolder, filename, suffix);
-            auto stream = openFile(fileName, true);
+            std::unique_ptr<std::ostream> stream{openFile(fileName, true)};
             if (stream != nullptr) {
                 if (Log::verbose()) std::cerr << "Writing program to " << fileName << std::endl;
-                P4::ToP4 toP4(stream, Log::verbose(), file);
+                P4::ToP4 toP4(stream.get(), Log::verbose(), file);
                 if (noIncludes) {
                     toP4.setnoIncludesArg(true);
                 }
-                node->apply(toP4);
-                delete stream;  // close the file
+                if (node) {
+                    node->apply(toP4);
+                } else {
+                    *stream << "No P4 program returned by the pass" << std::endl;
+                }
             }
             break;
         }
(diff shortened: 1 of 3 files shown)


frontends/common/parser_options.cpp
```diff
@@ -490,15 +490,18 @@ void ParserOptions::dumpPass(const char *manager, unsigned seq, const char *pass
...
+                P4::ToP4 toP4(stream.get(), Log::verbose(), file);
                 if (noIncludes) {
                     toP4.setnoIncludesArg(true);
                 }
-                node->apply(toP4);
-                delete stream;  // close the file
+                if (node) {
+                    node->apply(toP4);
+                } else {
+                    *stream << "No P4 program returned by the pass" << std::endl;
```
fruffy, line 503: So this only happens when there is an error right? I would clarify this in the message.

ir/pass_manager.cpp
```diff
@@ -85,8 +85,8 @@ const IR::Node *PassManager::apply_visitor(const IR::Node *program, const char *
                     LOG3(log_indent << "heap after " << v->name() << ": in use " << n4(mem)
                                     << "B, max " << n4(maxmem) << "B");
                 }
-                if (stop_on_error && ::errorCount() > initial_error_count) break;
-                if ((program = after) == nullptr) break;
+                if (stop_on_error && ::errorCount() > initial_error_count) early_exit_flag = true;
+                if ((program = after) == nullptr) early_exit_flag = true;
```
fruffy, line 89: What does it mean when `(program = after) == nullptr`? Unclear to me.

jafingerhut: I think this is C/C++'s fun "you can embed assignment statements inside of expressions" thing, semantically equivalent to doing:

```
program = after;
if (program == nullptr) early_exit_flag = true;
```
Or maybe I don't understand your question.

fruffy: Oh yes, it was more meant as a clarification question under which circumstances the `after` is nullptr. 
But that expression should also be simplified while we are at it.

vlstill: At least the Transform's code suggests it is possible to return `nullptr` in the pre/postorder and if you do that for root the entire `apply` would yield `nullptr`. This would be quite weird I guess, but it is allowed.

I've pushed a commit that eliminates the `after` variable altogether.

fruffy (approved):
Approving, although I can not say for sure there won't be any unintended side-effects.

vlstill:
> Approving, although I can not say for sure there won't be any unintended side-effects.

One effect I can see is that the compiler will attempt to dump P4 also form the pass that triggered an error. That would of course happen only if dumping is enabled, so in develompment. The rest will depend on the debug hooks used by downstream projects, which is hard to know... If this turns out to be problematic, the hook can always check `::errorCount()`.

PR #4628, opened 2024-04-18
Use `IR::Annotation::nameAnnotation` everywhere instead of hard-coding `@name` annotation name

kfcripps:


testdata/p4_16_samples_outputs/annotation-inline-propagate-frontend.p4
```diff
@@ -32,7 +32,7 @@ control MyVerifyChecksum(inout headers hdr, inout metadata meta) {
     }
 }
 
-@my_anno_5 @my_anno_3 control MyIngress(inout headers hdr, inout metadata meta, inout standard_metadata_t standard_metadata) {
+@my_anno_5 @my_anno_3 @name("MyIngress.inner_ctrl") control MyIngress(inout headers hdr, inout metadata meta, inout standard_metadata_t standard_metadata) {
```
asl, line 35: Why do we have test output change here?

kfcripps: @asl Apparently, we can't reference `IR::Annotations::nameAnnotation` in a static container initializer list:
```
// set of annotations to _not_ propagate during inlining
std::set<cstring> Inline::noPropagateAnnotations = {"name"};
```
so it must remain hard-coded here.

asl: Yes, order of static initializers in different TUs is not quite defined.

fruffy (approved):

PR #4629, opened 2024-04-18
Contribute DiagnosticCountInPassHook

vlstill:
This is the reason behind #4626. A pass manager hook that is useful for testing (p4test-like) the compiler output or just for the developers to see what passes emit diagnostic messages.

Example output (with the hook in driver added and running with `-TdiagnosticCountInPass:1`):

```
main.p4(65): [--Wwarn=parser-transition] warning: sub_parser: implicit transition to `reject'
           state sub_parser{
                 ^^^^^^^^^^
PASS FrontEnd ~> CreateBuiltins emitted 1 warning
[--Wwarn=parser-transition] warning: SelectCase: unreachable case
PASS MidEnd ~> PassRepeated emitted 1 warning
```

### DIFF
--- a/ir/pass_utils.cpp
+++ b/ir/pass_utils.cpp
@@ -44,7 +44,7 @@ DebugHook getDiagnosticCountInPassHook(BaseCompileContext &ctxt) {
                 }
             }
 
-            // if the log level is high enough, emit also pass count
+            // If the log level is high enough, emit also pass count.
             if (Log::fileLogLevelIsAtLeast(DIAGNOSTIC_COUNT_IN_PASS_TAG, 4)) {
                 LOG_FEATURE(DIAGNOSTIC_COUNT_IN_PASS_TAG, 4,
                             "PASS " << manager << "[" << seqNo << "] ~> " << pass << " emitted "
--- a/ir/pass_utils.cpp
+++ b/ir/pass_utils.cpp
@@ -1,0 +1,64 @@
+#include "ir/pass_utils.h"
+
+#include <initializer_list>
+#include <tuple>
+
+#include "lib/log.h"
+
+/// @file
+/// @authors Vladimír Štill
+/// @brief Utilities for passes and pass managers.
+
+namespace P4 {
+
+// The state needs to be shared between all hook copies as the hook is copied in the pass manager.
+struct DiagnosticCountInPassHookState {
+    explicit DiagnosticCountInPassHookState(BaseCompileContext &ctxt)
+        : lastDiagCount(ctxt.errorReporter().getDiagnosticCount()),
+          lastErrorCount(ctxt.errorReporter().getErrorCount()),
+          lastWarningCount(ctxt.errorReporter().getWarningCount()),
+          lastInfoCount(ctxt.errorReporter().getInfoCount()) {}
+
+    unsigned lastDiagCount, lastErrorCount, lastWarningCount, lastInfoCount;
+};
+
+DebugHook getDiagnosticCountInPassHook(BaseCompileContext &ctxt) {
+    return [state = std::make_shared<DiagnosticCountInPassHookState>(ctxt), &ctxt](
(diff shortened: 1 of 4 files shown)


ir/pass_utils.cpp
```diff
@@ -0,0 +1,64 @@
+#include "ir/pass_utils.h"
+
+#include <initializer_list>
+#include <tuple>
+
+#include "lib/log.h"
+
+/// @file
+/// @authors Vladimír Štill
```
fruffy, line 9: Nit: A little non-standard, should we use a copyright notice instead?

ir/pass_utils.h
```diff
@@ -0,0 +1,35 @@
+#ifndef IR_PASS_UTILS_H_
+#define IR_PASS_UTILS_H_
+
+#include "ir/pass_manager.h"
+#include "lib/compile_context.h"
+
+/// @file
```
fruffy, line 7: Same here.

ir/pass_utils.cpp
```diff
@@ -0,0 +1,64 @@
...
+                  std::tuple(warns, state->lastWarningCount, "warning"),
+                  std::tuple(infos, state->lastInfoCount, "info")}) {
+                if (newCnt > oldCnt) {
+                    auto diff = newCnt - oldCnt;
+                    summary << sep << diff << " " << kind << (diff > 1 ? "s" : "");
+                    sep = ", ";
+                }
+            }
+
+            // if the log level is high enough, emit also pass count
```
fruffy, line 47:
```suggestion
            // If the log level is high enough, emit also pass count.
```

fruffy (approved):
Could be nice to have an example exercising this hook.

ir/pass_utils.cpp
vlstill: Is there a "standard" copyright notice in modern "P4C"? I'm mostly finding old barefoot notices...

Or I can just skip this.

vlstill:
> Could be nice to have an example exercising this hook.

Let me at it to P4test, the output has to be enabled anyway so it should not break anything.

vlstill:
```bash
$ ./p4c/p4test -TdiagnosticCountInPass:1 <(printf 'const bit<42> x = -4;\nconst bit<16> y = (bit<16>)x;\nvoid foo() { return 4; }')
/dev/fd/63(1): [--Wwarn=mismatch] warning: -42w4: negative value with unsigned type
const bit<42> x = -4;
                   ^
/dev/fd/63(1): [--Wwarn=mismatch] warning: 16w4398046511100: value does not fit in 16 bits
const bit<42> x = -4;
                   ^
PASS ConstantFolding ~> DoConstantFolding emitted 2 warnings
/dev/fd/63(3): [--Werror=type-error] error: return 4: return expression in function with void return
void foo() { return 4; }
             ^^^^^^
PASS FrontEnd ~> P4::TypeInference emitted 1 error

fruffy (approved):

ir/pass_utils.cpp
fruffy: I do not think there is a standard, but the Apache template we use is the most common one. Recent example here: https://github.com/p4lang/p4c/blob/main/midend/def_use.h#L2 

But we also have https://github.com/p4lang/p4c/blob/e30416337cc0e3bf046909324aa7ae0b160f5224/lib/nethash.cpp#L23 here now, so I think we can go with either version.

vlstill:
Any idea what is wrong with the Fedora build?

>  /__w/p4c/p4c/build/_deps/abseil-src/absl/numeric/internal/bits.h:309:24: error: '__builtin_ctzs' needs isa option -mbmi

fruffy:
> Any idea what is wrong with the Fedora build?
> 
> > /__w/p4c/p4c/build/_deps/abseil-src/absl/numeric/internal/bits.h:309:24: error: '__builtin_ctzs' needs isa option -mbmi

It looks like Fedora GCC switched from 13 to 14 and this instruction now needs an option.

PR #4636, opened 2024-04-25
Rename reused-counter.p4 as reused-counter-bmv2.p4 so a test is created for it

jafingerhut:
The README.md file in directory testdata/p4_16_bmv_errors says that all test programs in that directory must have a name ending with `-bmv2.p4`, but `reused-counter.p4` did not, and I have confirmed that no `.test.p4` bash script is being generate for it by cmake.


fruffy:
I have made the fedora test optional until we find a good solution for it.

fruffy (approved):

PR #4638, opened 2024-04-25
Add -mbmi flag to Abseil for GCC versions greater than 14. 

fruffy:


vlstill:
This is quite weird. The BMI (bit manipulation instructions) seems to have been introduced in Intel Haswell. While that is 11 years old now, there are definitely computers running older HW still present around. I would expect GCC to be able lower the builtin to something else without the `-mbmi`. If we compiler with `-mbmi` we risk users hitting invalid instruction.

vlstill:
This is quite confusing. There is a GCC issue saying this builtin (`__builtin_ctzs`) will not be added: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=112789. Yet apparently the compiler used on fedore somehow reports it has it, even when it does not actually: https://github.com/abseil/abseil-cpp/blob/master/absl/numeric/internal/bits.h#L306C23-L306C37.

vlstill:
Yep, fedora 40 is shipping unreleased GCC version. Too bad abseil does not have any way to override the check.

ChrisDodd:
Not clear why only fedora gives this error -- the default target for gcc on most machines is generic x86_64 (ie, no additional extensions over what was in the original), so it would not be using requiring bmi instructions before.  Has something changed with abseil such that it does thing with shorts that weren't being done before?  We use `__builtin_ctz` in out bitvec code, but only for int/long/long long, not for short

fruffy:
> This is quite weird. The BMI (bit manipulation instructions) seems to have been introduced in Intel Haswell. While that is 11 years old now, there are definitely computers running older HW still present around. I would expect GCC to be able lower the builtin to something else without the `-mbmi`. If we compiler with `-mbmi` we risk users hitting invalid instruction.

I can add a flag that this is only added with GCC >14.

fruffy:
I do not understand how changing the flags for Abseil can cause `-werror` to be set for ubpf program compilation.

fruffy:
This PR is ready now. The fix to the failing uBPF programs is in #4644.

fruffy:
> @fruffy This smells like a mainline bug. Not sure where, actually. But maybe start with Abseil?

https://github.com/abseil/abseil-cpp/issues/1664 but unclear whether things will be resolved there.

asl:
> > @fruffy This smells like a mainline bug. Not sure where, actually. But maybe start with Abseil?
> 
> [abseil/abseil-cpp#1664](https://github.com/abseil/abseil-cpp/issues/1664) but unclear whether things will be resolved there.

Thanks! Yes, let's see what is there. IIRC `-march=native` was not quite compatible with arm64/clang, but it's only for GCC, so I'd guess we are fine for now.

asl (approved):

cmake/Abseil.cmake
```diff
@@ -53,7 +53,10 @@ macro(p4c_obtain_abseil)
         # Do not suppress warnings for Abseil library targets that are aliased.
         get_target_property(target_type ${target} TYPE)
         if (NOT ${target_type} STREQUAL "INTERFACE_LIBRARY")
-          set_target_properties(${target} PROPERTIES COMPILE_FLAGS "-Wno-error -w")
+          if (CMAKE_COMPILER_IS_GNUCC AND CMAKE_CXX_COMPILER_VERSION VERSION_GREATER 14)
+            target_compile_options(${target} PUBLIC "-march=native")
```
vlstill, line 57: Maybe add a cmake warning there? This could produce surprise incompatibilities in the binary. And it might not be that far in future before someone tries to build a release binary with GCC 14.

vlstill: Also it might be good to reference the abseil issue in a comment here. Hopefully we will be able to remove this hack in future.

vlstill:
#4643 make me think about this a bit more. I suggest enabling just the minimal set of flags (i.e. `-mbmi`) for the GCC 14 case. We should still try to preserve as much compatibility as we can.

cmake/Abseil.cmake
```diff
@@ -53,15 +53,24 @@ macro(p4c_obtain_abseil)
...
+          if (CMAKE_COMPILER_IS_GNUCC AND CMAKE_CXX_COMPILER_VERSION VERSION_GREATER 14)
+            target_compile_options(${target} PUBLIC "-mbmi")
+          endif()
+          target_compile_options(${target} PRIVATE "-Wno-error" "-w")
         endif()
       endif()
     endforeach()
+    # TODO: Remove once the Abseil compilation issue is fixed.
+    if (CMAKE_COMPILER_IS_GNUCC AND CMAKE_CXX_COMPILER_VERSION VERSION_GREATER 14)
+      message(WARNING "Compiling with GCC>14. Adding -mbmi to Abseil targets.")
```
vlstill, line 67:
```suggestion
      message(WARNING "Compiling with GCC > 14. Adding -mbmi to Abseil targets, this may cause incompatibility with old CPUs.")

vlstill (approved):
Looks good to me now. I just suggest a minor extension of the warning.

Issue #2407, opened 2020-06-01
Use libbpf instead of iproute2 to load eBPF programs


fruffy:
The dependency on iproute2 and tc causes some issues when loading eBPF programs. Different versions of Ubuntu package different versions of iproute2, which offers different levels of support for eBPF. In addition, there seems to be some problems trying to load more advanced map types such as LPM_TRIE maps.

[libbpf](https://github.com/libbpf/libbpf) offers an API to load and manage eBPF programs. It might make sense to write a small loader that takes care of attaching an eBPF program to an interface.

jafingerhut:
I do not know the intricacies involved here, but depending upon what the major EBPF users want, you could consider limiting 'official', i.e. checked-into-repository, support to only long term Ubuntu releases, and even carefully delimit a subset of those.

For example, one could consider no longer supporting Ubuntu 16.04 effective immediately, or "soon", for this purpose, although that would require updating other things which are currently tested on Travis on Ubuntu 16.04 images, I believe.

fruffy:
That could also be an option, I think for sanity's sake, Ubuntu 16.04 needs to be eventually deprecated to keep up with all the eBPF/XDP changes. Another reason to get rid of iproute2 is simply the lack of flexibility. There are data structures such as the `bpf_elf_map` I have to maintain just for loading with `tc`.

PR #4644, opened 2024-04-29
Fix missing declaration of custom externs in the generated eBPF/uBPF header file.

fruffy:
Fixes a warning that occurs when compiling eBPF/uBPF programs with a custom, user-defined extern. The extern has not been declared, which causes GCC to spit out `warning: implicit declaration of function`.

Because there are also method declarations in `core.p4` and the corresponding include file I need to add a filter based on source information. I do not know whether there is a better approach to this. 

I also added the error type, which is the same as the enum type. I also could add `Type_String`, but it has unclear semantics on width. 

@thomascalvert-xlnx


fruffy:
The failing tests are here: https://github.com/p4lang/p4c/actions/runs/8862986990/job/24336526254

The problem is that, while the `-include` enables compilation. The function definition is missing when compiling individual objects. That leads to this unnecessary warning, which is treated as error for some systems.

My fix simply adds a declaration in the header file to prevent the warning from occurring. The definition is still in the `.c` file of the extern. 

It would be great if we had golden files which show the change.

thomascalvert-xlnx (commented):
Thanks for clarifying @fruffy. I notice in that test output that only the ubpf extern tests are failing, not ebpf - do we know why?

I agree that emitting the declarations is a good step forwards, the suggested `-include` method feels like a bit of a hack (imho).

As mentioned in my previous comment, I don't think that discriminating based on the source filename is robust. We're already missing `ebpf_model.p4` and `xdp_model.p4` here. If we don't want to emit declarations for "standard" externs which are already defined, perhaps a better solution would be to filter based on a list of such standard externs? Any other suggestions?

fruffy:
>If we don't want to emit declarations for "standard" externs which are already defined, perhaps a better solution would be to filter based on a list of such standard externs? Any other suggestions?

I will give this a shot.

fruffy:
> I notice in that test output that only the ubpf extern tests are failing, not ebpf - do we know why?

Yeah, I wondered about that too. It must be related to the compile order, but I can't identify the difference.

fruffy:
@thomascalvert-xlnx I implemented a technique to filter based on method names instead. I also created #4645 because I ran into a bunch of issues. For example, `progTarget` is a member class of the EBPFProgram but is only really initialized by the XDP back end. In all other cases it is a nullptr.

thomascalvert-xlnx (approved):
LGTM thanks for looking at this Fabian!

Agree it's hard to work on the eBPF backend due to the tangled dependencies & inheritance with other backends (ubpf, tc, psa). Regarding the `progTarget` variable you refer to, as far as I can tell it is in fact only really useful to the PSA backend (sub-backend?).

fruffy:
I do not understand the Ubuntu 20.04 failure. The tests passes on Ubuntu 18.04 and Ubuntu 22.04 but not here. The output doesn't help either. @thomascalvert-xlnx any intuition. Do we really need to declare every function `static inline`?

thomascalvert-xlnx:
It looks like the verifier is rejecting the program with the following error:
```
BTF debug data section '.BTF' rejected: Invalid argument (22)!
```
There seems to be a similar issue reported here: https://github.com/xdp-project/xdp-tutorial/issues/38

As far as I can tell it's an issue with specific tool versions - could be a mismatch between libbpf versions of the OS vs this repo? Or could be some problem with the debug info, try dumping it as suggested in the link above?

fruffy:
https://github.com/p4lang/p4c/issues/2407 could help here, but I never implemented it. Unfortunate.

fruffy:
I give up. I do not know what else to try. I will just ignore this test on Ubuntu 20.04.

fruffy:
@thomascalvert-xlnx I disabled method signature generation for eBPF and only enabled it for uBPF. Until we fix the way we load eBPF programs (or move on from Ubuntu 20.04) I do not see a better way to do this. Please take a second look.

fruffy:
@thomascalvert-xlnx I will merge this tomorrow if you have no further comments.

thomascalvert-xlnx (approved):
Sorry for the delay, we had a bank holiday here and this thread fell out of my cache!

LGTM now, thank you @fruffy :)

PR #4647, opened 2024-05-01
Introduce guard for aarch64 GCC compilation.

### LINKED ISSUE #4639: include of file mm_malloc.h breaks build for Ubuntu aarch64 systems
This commit on 2024-Apr-18: https://github.com/p4lang/p4c/commit/c7d688a4ab857c684c06c5b6f6aaa5c698765360

added these lines to file `lib/gc.cpp`:

```
// Some systems (e.g. those with GNU libc) declare posix_memalign with an
// exception specifier. Compilers (clang including) might have special handling
// of this to allow posix_memalign redeclaration with / without exception
// specifier. As we define posix_memalign below in this file we really need to
// ensure the proper include order to workaround this weirdness.
#include <mm_malloc.h>  // NOLINT(build/include_order)
```

This file exists in Ubuntu package libgcc-11-dev on Ubuntu 22.04 for amd64 arch processors, but there is no such file for aarch64 processors on Ubuntu 22.04.

It would be nice if p4c could compile on recent Ubuntu versions for aarch64 processor architecture, too.  It was building a month ago, before the commit linked above, but I haven't done exhaustive testing on every commit to see if there are any blocking issues to achieve that other than the commit above.

fruffy:
Fixes #4639.

### DIFF
--- a/BUILD.bazel
+++ b/BUILD.bazel
@@ -29,7 +29,10 @@ genrule(
     cmd = " | ".join([
         "sed 's|cmakedefine|define|g' < $(SRCS)",
         "sed 's|define HAVE_LIBGC 1|undef HAVE_LIBGC|g'",
+        # TODO: We should actually check these properly instead of just #undefing them.
+        # Maybe select() can help here?
         "sed 's|define HAVE_LIBBACKTRACE 1|undef HAVE_LIBBACKTRACE|g' > $(OUTS)",
+        "sed 's|define HAVE_MM_MALLOC_H 1|undef HAVE_MM_MALLOC_H|g' > $(OUTS)",
     ]),
     visibility = ["//visibility:private"],
 )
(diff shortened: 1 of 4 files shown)


jafingerhut (approved):
LGTM.  With these changes, I verified I can build p4c and pass several tests on an Ubuntu 20.04 aarch64 system.

ChrisDodd:
Should this be done with a `check_include_file(mm_malloc.h, HAVE_MM_MALLOC_H)` in CMakeLists.txt and `#cmakedefine HAVE_MM_MALLOC_H 1` in cmake/config.h.cmake?

fruffy:
> Should this be done with a `check_include_file(mm_malloc.h, HAVE_MM_MALLOC_H)` in CMakeLists.txt and `#cmakedefine HAVE_MM_MALLOC_H 1` in cmake/config.h.cmake?

I can add this instead, just a little tricky to test whether it is working right.

jafingerhut:
I retested this PR locally after it had 2 commits on an Ubuntu 20.04 aarch64 Linux system, and it compiled and tested fine.

PR #4649, opened 2024-05-02
Use check_include_file_cxx instead of check_include_file to find mm_malloc.h

fruffy:
It looks like the other function never finds `mm_malloc.h`

### DIFF
--- a/.github/workflows/ci-static-build-test.yml
+++ b/.github/workflows/ci-static-build-test.yml
@@ -21,7 +21,7 @@ jobs:
     if: ${{ github.event_name == 'schedule' || contains(github.event.pull_request.labels.*.name, 'run-static') }}
     # Note: Actions have no real ternary condition operator, but this hack works because non-empty
     # strings are true.
-    name: "Dynamic ${{ matrix.dynamic.glibc == 'ON' && 'glibc' || 'stdlib' }}"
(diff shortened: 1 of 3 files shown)


ChrisDodd (approved):

jafingerhut:
I know this is already merged, but as a confirmation, I built on an Ubuntu 20.04 aarch64 system with these changes.  The build succeeded, and resulting binaries passed at least a few tests.

ChrisDodd:
I think @asl originally put in this include for some particular variation (gcc vs clang perhaps, from the comment).  It might be good if he can recall under what condition the include was required originally.

asl:
@ChrisDodd The particular change was added in https://github.com/p4lang/p4c/pull/4609 in order to fix `noexcept` / no-`noexcept` problem of `posix_memalign` declaration. It's really a mess that depends on combination of compilers and libc flavours. Unfortunately, we have to do something like this as we re-define `posix_memalign` override later.

Thinking about this more, maybe we'd resort to cmake-stage check whether `posix_memalign` is defined with `noexcept` specifier or not. And use this for redefine.

fruffy:
Yes, I ran into this trying to compile a unity build with clang (https://github.com/p4lang/p4c/pull/4595) and this was the fix we came up with. 

The CMake check sounds good, but I am not sure whether we can check the noexcept?

PR #4650, opened 2024-05-02
Merge TestgenCompilerTarget into TestgenTarget.

fruffy:
Some `#codehealth`. We currently have two "Targets", the CompilerTarget and the ToolTarget. However, in practice we do not make any use of this distinction and the two targets end up being a constant source of maintenance burden and confusion. This PR consolidates the two into one single target. This also reduces the amount of code you need to copy when you create a new P4Tools tool or target. 

This PR is also a first step into reducing the amount of static objects. Now, instead of two static targets we only have one. To deal with the hardcoded tool initialization I had to parametrize some functions, but imho this just exposes some hidden ugliness. Technically, we do not even really need `toolName` as a distinction in the registry. We never have more than one tool in use since the tools are generally different binaries. But I have decided to keep it for now. 

@vlstill the amount of changes in the P4Testgen targets should be minimal. Mostly changing includes to `compiler_result` and merging the `CompilerTarget` into the `TestgenTarget` object.


backends/p4tools/common/options.h
```diff
@@ -29,6 +34,11 @@ class AbstractP4cToolOptions : protected Util::Options {
     /// @returns a compilation context on success, std::nullopt on error.
     std::optional<ICompileContext *> process(const std::vector<const char *> &args);
 
+    // No copy constructor and no self-assignments.
+    AbstractP4cToolOptions(const AbstractP4cToolOptions &) = delete;
+
+    AbstractP4cToolOptions &operator=(const AbstractP4cToolOptions &) = delete;
```
vlstill, line 40: Is move acceptable?

backends/p4tools/modules/testgen/core/target.h
```diff
@@ -64,7 +61,10 @@ class TestgenTarget : public Target {
     virtual ExprStepper *getExprStepperImpl(ExecutionState &state, AbstractSolver &solver,
                                             const ProgramInfo &programInfo) const = 0;
 
-    explicit TestgenTarget(std::string deviceName, std::string archName);
+    explicit TestgenTarget(const std::string &deviceName, const std::string &archName);
+
+ private:
+    CompilerResultOrError runCompilerImpl(const IR::P4Program *program) const override;
```
vlstill, line 67: Should this be `private` if it is a virtual function? Maybe make it `protected` so that it can be called from a derived class.

backends/p4tools/modules/testgen/targets/ebpf/target.cpp
```diff
@@ -86,4 +86,12 @@ EBPFExprStepper *EBPFTestgenTarget::getExprStepperImpl(ExecutionState &state,
     return new EBPFExprStepper(state, solver, programInfo);
 }
 
+MidEnd EBPFTestgenTarget::mkMidEnd(const CompilerOptions &options) const {
+    MidEnd midEnd(options);
+    midEnd.addPasses({});
```
vlstill, line 91: Is there any point in adding empty list of passes?

vlstill (commented):
Overall looks good to me, with a few points below:

backends/p4tools/common/options.h
fruffy: That was suggested by clang-tidy. https://clang.llvm.org/extra/clang-tidy/checks/modernize/use-equals-delete.html

backends/p4tools/common/core/target.cpp
```diff
@@ -76,12 +76,13 @@ const IR::Expression *Target::createTargetUninitialized(const IR::Type *type,
     return IR::getDefaultValue(type);
 }
 
-Target::Target(std::string toolName, std::string deviceName, std::string archName)
+Target::Target(std::string_view toolName, const std::string &deviceName,
+               const std::string &archName)
     : toolName(toolName), spec(deviceName, archName) {
     // Register this instance.
-    BUG_CHECK(!registry[spec].count(toolName), "Already registered %1%/%2% instance for %3%",
+    BUG_CHECK(!registry[spec].count(toolName.data()), "Already registered %1%/%2% instance for %3%",
```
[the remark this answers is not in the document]
fruffy: Doesn't seem to work for the [] operator sadly.

backends/p4tools/common/compiler/compiler_target.cpp
```diff
@@ -119,8 +118,12 @@ const IR::P4Program *CompilerTarget::runMidEnd(const IR::P4Program *program) con
...
 }
 
-CompilerTarget::CompilerTarget(std::string deviceName, std::string archName)
-    : Target("compiler", std::move(deviceName), std::move(archName)) {}
+CompilerTarget::CompilerTarget(std::string_view toolName, const std::string &deviceName,
+                               const std::string &archName)
+    : Target(toolName, deviceName, archName) {}
+
+const CompilerTarget &CompilerTarget::get(std::string_view toolName) {
+    return Target::get<CompilerTarget>(toolName.data());
```
vlstill, line 126:
```suggestion
    return Target::get<CompilerTarget>(toolName);

File: backends/p4tools/common/core/target.cpp [hunk 6]
```diff
@@ -76,12 +77,13 @@ const IR::Expression *Target::createTargetUninitialized(const IR::Type *type,
...
 
-Target::Target(std::string toolName, std::string deviceName, std::string archName)
+Target::Target(std::string_view toolName, const std::string &deviceName,
+               const std::string &archName)
     : toolName(toolName), spec(deviceName, archName) {
     // Register this instance.
     BUG_CHECK(!registry[spec].count(toolName), "Already registered %1%/%2% instance for %3%",
               deviceName, archName, toolName);
-    registry[spec][toolName] = this;
+    registry[spec][toolName.data()] = this;
```
vlstill, line 86:
```suggestion
    registry[spec][toolName] = this;

File: backends/p4tools/common/core/target.h [hunk 7]
```diff
@@ -70,23 +70,24 @@ class Target {
...
 
         const auto &instances = registry.at(*curTarget);
-        BUG_CHECK(instances.count(toolName), "Architecture %1% on device %2% not supported for %3%",
-                  curTarget->archName, curTarget->deviceName, toolName);
+        BUG_CHECK(instances.count(toolName.data()),
+                  "Architecture %1% on device %2% not supported for %3%", curTarget->archName,
+                  curTarget->deviceName, toolName);
 
-        const auto *instance = instances.at(toolName);
+        const auto *instance = instances.at(toolName.data());
```
vlstill, line 90:
```suggestion
        const auto *instance = instances.at(toolName);

File: backends/p4tools/modules/testgen/targets/pna/target.cpp [hunk 8]
```diff
@@ -78,4 +78,12 @@ PnaDpdkExprStepper *PnaDpdkTestgenTarget::getExprStepperImpl(ExecutionState &sta
     return new PnaDpdkExprStepper(state, solver, programInfo);
 }
 
+MidEnd PnaDpdkTestgenTarget::mkMidEnd(const CompilerOptions &options) const {
+    MidEnd midEnd(options);
+    midEnd.addPasses({});
```
vlstill, line 83:
```suggestion

backends/p4tools/common/core/target.cpp
fruffy: Lookups do not seem to work for transparent comparators after all.

vlstill (approved):

backends/p4tools/common/core/target.cpp
vlstill: Oh, I accidentally looked at the first level map in `registry` where the key is `Target::Spec`. Then this is weird, it should work if I understand C++ documentation correctly :-/.

vlstill:
@fruffy, do we want to make breaking changes in P4Tools too? It might be confusing to most backend developers if most don't use it. Do you know if there is anyone besides the two of us and the Tofino testgen?

fruffy:
> @fruffy, do we want to make breaking changes in P4Tools too? It might be confusing to most backend developers if most don't use it. Do you know if there is anyone besides the two of us and the Tofino testgen?

For now I figured it might be better to not mark the breaking changes here and instead ping folks. There is also the Mount Evans team and the Google team but I do not know of other teams actively developing their own back ends otherwise.

vlstill:
Yep, OK, if someone asks for it maybe we can bake a special tag for breaking changes in p4tools.

PR #4653, opened 2024-05-06
Improve error message when shifting int by non-const

### LINKED ISSUE #4652: Wrong error message when arbitrary-precision int is shifted by runtime value
Using [shift.p4](https://github.com/p4lang/p4c/files/15218626/shift.p4.txt) as a test program.

```
$ ./p4c/p4test shift.p4
shift.p4(8): [--Werror=type-error] error: 5 >> b: width of left operand of shift needs to be specified
    hdr.v = (bit<8>)(a >> b);
                     ^^^^^^
```

This error message is not correct, as [spec v1.2.4 sec 8.9.2](https://p4.org/p4-spec/docs/P4-16-v1.2.4.html#sec-a-note-about-shifts) says:

> The left operand of shifts can be any one out of unsigned bit-strings, signed bit-strings, and arbitrary-precision integers, and the right operand of shifts must be either an expression of type `bit<S>` or a non-negative integer compile-time known value. The result has the same type as the left operand.

So the problem is not that the left operand is `int`. The actual problem is that this would imply the result is also `int`, which is a problem. It is a problem because `b` is a runtime value and therefore the result is a runtime value too. And runtime values cannot have type `int` ([sec 8.8 f spec](https://p4.org/p4-spec/docs/P4-16-v1.2.4.html#sec-varint-ops)):

> The type int denotes arbitrary-precision integers. In P4, all expressions of type int must be compile-time known values. [...]

Note that this works if `b` is turned into a constant or if `a` is turned into fixed-width value.

---

I think this is just a case of unclear message in the end.

vlstill:
Fixes #4652. Note that besides changing the error I've also allowed the case when the right operand is `bit<N>`/`int<N>` but it is also constant. This case should be allowed and it does actually work in most cases in practice, it is just covered by constant folding.

---

BTW: I did not realize until now that the type checker actually *does constant folding* too! This will make any "small frontend" quite hard to do if it was to not do constant folding.

### DIFF
--- a/frontends/p4/typeChecking/typeChecker.cpp
+++ b/frontends/p4/typeChecking/typeChecker.cpp
@@ -2684,8 +2684,13 @@ const IR::Node *TypeInference::shift(const IR::Operation_Binary *expression) {
         return expression;
     }
 
-    if (ltype->is<IR::Type_InfInt>() && !rtype->is<IR::Type_InfInt>()) {
-        typeError("%1%: width of left operand of shift needs to be specified", expression);
+    if (ltype->is<IR::Type_InfInt>() && !rtype->is<IR::Type_InfInt>() &&
+        !isCompileTimeConstant(expression->right)) {
+        typeError(
+            "%1%: shift result type is arbitrary-precision int, but right operand is not constant; "
+            "width of left operand of shift needs to be specified or both operands need to be "
+            "constant",
+            expression);
         return expression;
     }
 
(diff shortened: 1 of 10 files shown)


ChrisDodd (approved):

fruffy:
> BTW: I did not realize until now that the type checker actually _does constant folding_ too! This will make any "small frontend" quite hard to do if it was to not do constant folding.

Yes, I believe this is necessary to fold compile-time arithmetic in `bit<>` types. At least the `constantFolding` pass is relatively self-contained and can be applied without a type or ref-map. We could fork out a smaller version that only works on compile-time constants and is used for the purpose of type inference.

It could then be extended by the full ConstantFolding pass.

PR #4655, opened 2024-05-08
Fixes to lib/hash and lib/big_int_util.

fruffy:
A couple fixes after #4623 has been merged.

- Move the boost definition of `boost::multiprecision::cpp_int` into its own header file.
- Fix missing of typing for BoolLiterals with SourceInfo.
- Move `big_int` hashing to `big_int_utils`.

### DIFF
--- a/BUILD.bazel
+++ b/BUILD.bazel
@@ -41,6 +41,7 @@ cc_library(
     name = "config_h",
     hdrs = ["config.h"],
     include_prefix = ".",
+    deps = [":sed_config_h"],
 )
 
 cc_library(
--- a/lib/alloc_trace.cpp
+++ b/lib/alloc_trace.cpp
@@ -64,6 +64,9 @@ void AllocTrace::count(void **bt, size_t sz) {
 }
 
 std::ostream &operator<<(std::ostream &out, const AllocTrace &at) {
+#if HAVE_LIBGC
+    PauseTrace temp_pause;
+#endif
     typedef decltype(at.data)::value_type data_t;
     std::vector<std::pair<size_t, const data_t *>> sorted;
     size_t total_total = 0;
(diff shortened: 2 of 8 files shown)


fruffy:
There is a new spurious failure in the Bazel build... it looks like it is missing the `HAVE_LIBBOOST_GRAPH` definition.

BUILD.bazel
vlstill, line 1: I have no ability to check the bazel config.

lib/alloc_trace.cpp
```diff
@@ -64,9 +64,6 @@ void AllocTrace::count(void **bt, size_t sz) {
 }
 
 std::ostream &operator<<(std::ostream &out, const AllocTrace &at) {
-#if HAVE_LIBGC
-    PauseTrace temp_pause;
-#endif
```
vlstill, line 69: Why is this removed? I don't know what it does, but it does not seem very related.

vlstill (commented):
Just one question. The rest look OK to me.

BUILD.bazel
fruffy: I am just guessing here but it does seem to fix the spurious failure.

lib/alloc_trace.cpp
fruffy: A warning popped out https://github.com/p4lang/p4c/actions/runs/9003590407/job/24734561146. Not sure what triggered it to appear only now but this piece of code does seem unused.

vlstill: Weird. I would expect it to be a guard that does something in its constructor and destructor (based on the name).

fruffy: I can alternatively just add a NOLINT.

vlstill: Indeed, see https://github.com/p4lang/p4c/blob/main/lib/alloc_trace.h#L67. I don't think this should be removed. Is it possible this build (which produced the warning) is without GC? I think the "right" solution to suppress the warning might be to put and explicit destructor there even for the case where the `HAVE_LIBGC` is false.

fruffy: `alloc_trace.h` did not include config and removing `config.h` from `big_int_util.h` caused the `#define` to be missing.

vlstill (approved):

